PR #2925, opened 2019-10-25
Implement multi-touch input handling

iiSaLMaN:
- Implements touch input handling properly with multi-touch support but haven't been used yet. (platform-specific usages will come in separate PRs)
- [x] Depends on ppy/osuTK#55

First step onto resolving #1514

PR #3017, opened 2019-11-20
Refine some generic constraints

huoyaoyuan:
Found it's really annoying to work with generics allowing both struct and class when applying nullable annotations.

Some explanations:
- ResourceStore is already implying reference type by using null for not found.
- interfaces inherits IDrawable should all imply reference types.
- Configuration manager is already implying enum by the `(int)(object)lookup` cast in framework side.
- The enum constraints are better to exist at my side. Not strongly required.

**This is a breaking change that requires game-side generic stuff to update their constraints.**

A cheat sheet for generic constraints:
##### with nullable disabled:
Constraint|SomeClass|SomeStruct|SomeStruct?
-|-|-|-
-|√|√|√
class|√||
struct||√|
SomeInterface|√|√|
SomeBaseClass|√||

##### with nullable enabled:
Constraint|SomeClass|SomeClass?|SomeStruct|SomeStruct?
-|-|-|-|-
-|√|√|√|√
class|√||||
class?|√|√||
struct|||√|
notnull|√||√|
SomeInterface|√||√|
SomeInterface?|√|√|√|
SomeBaseClass|√||
SomeBaseClass?|√|√|


osu.Framework/IO/Stores/IGlyphStore.cs
```diff
@@ -6,7 +6,7 @@
 namespace osu.Framework.IO.Stores
 {
     /// <summary>
-    /// Interface for a <see cref="IResourceStore{CharacterGlyph}"/>.
+    /// Interface for a <see cref="IResourceStore{T}"/> of <see cref="CharacterGlyph"/>.
```
huoyaoyuan: Xmldoc only refers open generics, not constructed ones.

peppy:
I haven't completely gone through the reasoning for this yet, but is there a way we can enforce these rules if we're going to use them?

huoyaoyuan:
> but is there a way we can enforce these rules if we're going to use them?

Do you mean that certain interfaces must be constrained together with class or struct? I think there's no way except custom analyzers now.

Constraints on generic methods is much less critical on generic classes, when input-only or copied from input to output.

osu.Framework/Graphics/OpenGL/Vertices/VertexUtils.cs
```diff
@@ -15,12 +15,12 @@ namespace osu.Framework.Graphics.OpenGL.Vertices
...
     /// </summary>
     internal static class VertexUtils<T>
-        where T : IVertex
+        where T : struct, IVertex
     {
         /// <summary>
         /// The stride of the vertex of type <typeparamref name="T"/>.
         /// </summary>
-        public static readonly int STRIDE = Marshal.SizeOf(default(T));
+        public static readonly int STRIDE = Marshal.SizeOf<T>();
```
smoogipoo: This needs to be tested on Xamarin.iOS. This was an intentional change because `Marshal.SizeOf<T>` was too deep for AOT.

osu.Framework/IO/Stores/IResourceStore.cs
```diff
@@ -14,6 +14,7 @@
 namespace osu.Framework.IO.Stores
 {
     public interface IResourceStore<T> : IDisposable
+        where T : class
```
smoogipoo: Debatable. I foresee usages where we may want non-class returns.

osu.Framework/Input/StateChanges/ButtonInput.cs
```diff
@@ -13,7 +14,7 @@ namespace osu.Framework.Input.StateChanges
     /// </summary>
     /// <typeparam name="TButton">The type of button.</typeparam>
     public abstract class ButtonInput<TButton> : IInput
-        where TButton : struct
+        where TButton : struct, Enum
```
smoogipoo: There will be cases where this won't be an enum, see: https://github.com/ppy/osu-framework/pull/2925/files#diff-930a86981796d62f4cea9e06cb8016fbR10

osu.Framework/Input/StateChanges/ButtonInputEntry.cs
```diff
@@ -1,14 +1,16 @@
...
+
 namespace osu.Framework.Input.StateChanges
 {
     /// <summary>
     /// Denotes a button state.
     /// </summary>
     /// <typeparam name="TButton">Type of button.</typeparam>
     public struct ButtonInputEntry<TButton>
-        where TButton : struct
+        where TButton : struct, Enum
```
smoogipoo: Same here re: enum inputs.

osu.Framework/Input/States/ButtonStates.cs
```diff
@@ -13,7 +14,7 @@ namespace osu.Framework.Input.States
     /// </summary>
     /// <typeparam name="TButton">The type of button.</typeparam>
     public class ButtonStates<TButton> : IEnumerable<TButton>
-        where TButton : struct
+        where TButton : struct, Enum
```
smoogipoo: Same here re: enum inputs.

osu.Framework/Input/StateChanges/Events/ButtonStateChangeEvent.cs
```diff
@@ -1,12 +1,13 @@
...
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using osu.Framework.Input.States;
 
 namespace osu.Framework.Input.StateChanges.Events
 {
     public class ButtonStateChangeEvent<TButton> : InputStateChangeEvent
-        where TButton : struct
+        where TButton : struct, Enum
```
smoogipoo: Same here re: enum inputs.

osu.Framework/Graphics/OpenGL/Vertices/VertexUtils.cs
smoogipoo: I guess since this succeeded on CI, it's probably safe. Marking as resolved.

osu.Framework/IO/Stores/IResourceStore.cs
huoyaoyuan: - Current implementation cannot work with value type. It checks with null for not found.
- Annotation on the `Task<T> GetAsync` will only be possible if constrained.

osu.Framework/Graphics/OpenGL/Vertices/VertexUtils.cs
huoyaoyuan: Well I didn't mention iOS. The Windows CI never covers iOS AOT.

huoyaoyuan: Hmmm, `SizeOf(default(T))` binds to `SizeOf<T>(T)`, not `SizeOf(object)`. I think `SizeOf<T>()` gives same information to AOT.

PR #3019, opened 2019-11-21
Improve video performance and memory consumption

Pizzabelly:
This uses a GLSL shader to display the video frames in planar format without converting them to RGBA. This should remove some CPU overhead. It also changes they way frames are uploaded so it doesn't need to copy every frame.

The shader is from this post https://nervous.io/ffmpeg/opengl/2017/05/15/ffmpeg-pbo-yuv/

In cases where a video isn't the correct format (YUV420P) there is a libav filtergraph that will convert the frames. 

This should be tested on mobile platforms, I personally don't have a way to test them.


peppy:
Got any benchmarks?

osu.Framework/Graphics/Video/VideoSprite.cs
```diff
@@ -126,6 +132,18 @@ protected override void Update()
         {
             base.Update();
 
+            if (!setUniforms)
+            {
+                int y = 0, u = 1, v = 2;
+                TextureShader.GetUniform<int>("tex_y").UpdateValue(ref y);
+                TextureShader.GetUniform<int>("tex_u").UpdateValue(ref u);
```
smoogipoo: This should be done in a `VideoSpriteDrawNode`. Something like:

```csharp
class VideoSpriteDrawNode:
    Draw():
        Shader.GetUniform(...).Value = ...
        base.Draw();
```

osu.Framework/Graphics/Video/VideoSprite.cs
```diff
@@ -126,6 +132,18 @@ protected override void Update()
         {
             base.Update();
 
+            if (!setUniforms)
```
smoogipoo: When moving it to the drawnode, don't keep this. It's unsafe in case these values ever change, and value updates are cached/optimised at a uniform level.

osu.Framework/Resources/Shaders/sh_VideoRounded.fs
```diff
@@ -0,0 +1,135 @@
...
+const vec3 offsets     = vec3(-0.0625, -0.5, -0.5);
+
+vec3 sampleRgb(vec2 loc) {
+  float y = texture2D(tex_y, loc).r;
+  float u = texture2D(tex_u, loc).r;
+  float v = texture2D(tex_v, loc).r;
+  return bt601_coeff * (vec3(y, u, v) + offsets);
+}
+
+highp float distanceFromRoundedRect(highp vec2 offset, highp float radius)
```
smoogipoo: can't have the duplication of this entire file. need to consider either integrating this into TextureRounded or providing a way to extend existing shaders with new functionality.

we can already kinda do this with imports, it would just require moving these methods + the `main()` method into a separate `.h` file and renaming `main()` to something else - `getRoundedColour(vec4 texColour)` or similar.

Pizzabelly:
devenv_7SpWd7ueR2
devenv_sTlMCpIQOR

This is an example of 25 second period on the same map (https://osu.ppy.sh/beatmapsets/345263#osu/762518)

I'll get to work on those changes tomorrow.

osu.Framework/Graphics/OpenGL/GLWrapper.cs
```diff
@@ -291,6 +292,29 @@ internal static void SetActiveBatch(IVertexBatch batch)
         internal static int GetTextureUnitId(TextureUnit unit) => (int)unit - (int)TextureUnit.Texture0;
         internal static bool AtlasTextureIsBound(TextureUnit unit) => last_bound_texture[GetTextureUnitId(unit)] is TextureGLAtlas;
 
+        /// <summary>
+        /// Binds a video texture to darw with.
```
turbedi: Typo, should be _draw_

Pizzabelly:
I did a lot more testing on this and found a couple issues that should be fixed now.

Firstly I got an Android build working and the performance improvement is very noticeable on my phone (Pixel 2 XL). Though there was an issue were the video would show an old frame if the new frame uploaded to the TextureGL hadn't been processed in time. The IsReady flag in VideoTexture should fix this, it's set after the call to DoUpload finishes.

Then when testing various maps with videos I found a couple that would not play even in master.

https://osu.ppy.sh/beatmapsets/499953#osu/1063867 
The video in this map has some invalid data in the beginning which causes an error in avcodec_send_packet. This can be safely ignored to normally play the video. So instead of throwing an exception on an avcodec_send_packet error I figured we can just log it.

https://osu.ppy.sh/beatmapsets/40499#osu/499713
This video doesn't have duration data in the stream, so its safer to rely on the AVFormatContex duration (after calling avformat_find_stream_info).

osu.Framework/Graphics/Video/VideoSprite.cs
Pizzabelly: Does my VideoSpriteDrawNode look ok?

osu.Framework/Resources/Shaders/sh_VideoRounded.fs
Pizzabelly: I moved the methods to sh_Masking.h and also made sh_yuv2rgb.h so that code wasn't copied as well.

peppy:
@Pizzabelly sorry for the long delay without review on this one. Are you able to check the merge conflicts and resolve them if you can? Will try and get this reviewed this week.

Pizzabelly:
@peppy no problem, everything sounds good. Also, I like the idea of #3149 as I think the scope of this pr could be expanded (shaders for other pixel formats, colorspace constants based on [yuv->rgb coeffs](https://github.com/FFmpeg/FFmpeg/blob/master/libswscale/yuv2rgb.c#L49)).

xbmc has a good example of what that would look like:
https://github.com/xbmc/xbmc/blob/master/system/shaders/GL/1.5/gl_yuv2rgb_basic.glsl
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/VideoPlayer/VideoRenderers/VideoShaders/YUV2RGBShaderGL.cpp

osu.Framework/Graphics/Video/VideoDecoder.cs
```diff
@@ -286,6 +284,69 @@ private static long seek(void* opaque, long offset, int whence)
             return decoder.videoStream.Position;
         }
 
+        private void prepareFilters()
+        {
+            var filterStr = "";
```
peppy: @Pizzabelly is this code copied/referenced from somewhere specific?

peppy:
Tested against world cup 2019 & 2020 videos for compatibility, looks good 👍 .

osu.Framework/Resources/Shaders/sh_yuv2rgb.h
```diff
@@ -0,0 +1,17 @@
+uniform sampler2D m_SamplerY;
+uniform sampler2D m_SamplerU;
+uniform sampler2D m_SamplerV;
```
Tom94: Why 3 single-channel samplers (and therefore 3 textures) rather than a single RGB texture, where R->Y, G->U, and B->V?

Tom94: Nevermind, silly question after looking at the code some more. YUV are stored at different resolutions.

osu.Framework/Graphics/Video/VideoTexture.cs
```diff
@@ -0,0 +1,114 @@
...
+                memoryLease = NativeMemoryTracker.AddMemory(this, Width * Height * 3 / 2);
+
+                textureIds = new int[3];
+                GL.GenTextures(textureIds.Length, textureIds);
+
+                for (int i = 0; i < textureIds.Length; i++)
+                {
+                    GL.BindTexture(TextureTarget.Texture2D, textureIds[i]);
+                    GL.TexImage2D(TextureTarget2d.Texture2D, 0, TextureComponentCount.R8,
+                        videoUpload.Frame->width / (i > 0 ? 2 : 1), videoUpload.Frame->height / (i > 0 ? 2 : 1), 0, PixelFormat.Red, PixelType.UnsignedByte, IntPtr.Zero);
```
Tom94: Just for sanity: did you verify correct behavior on non-even-resolution videos?

peppy: will test, good call

osu.Framework/Graphics/Video/VideoDecoder.cs
Pizzabelly: yes it comes from here https://ffmpeg.org/doxygen/2.1/doc_2examples_2filtering_video_8c-example.html

osu.Framework/Graphics/Video/VideoTexture.cs
Pizzabelly: It seems the correct way to handle this is to round up the UV dimensions

osu.Framework/Graphics/Video/VideoDecoder.cs
Pizzabelly: I changed this part in my most recent commit to use swscale instead of a libavfilter. It performs better on mobile and I think the code is cleaner.

osu.Framework/Resources/Shaders/sh_yuv2rgb.h
```diff
@@ -2,16 +2,15 @@ uniform sampler2D m_SamplerY;
 uniform sampler2D m_SamplerU;
 uniform sampler2D m_SamplerV;
 
-const mediump mat3 bt601_coeff = mat3(1.164,  1.164, 1.164,
-                                      0.0,   -0.392, 2.017,
-                                      1.596, -0.813, 0.0);
+uniform mat3 yuvCoeff;
```
smoogipoo:
```suggestion
uniform mediump mat3 yuvCoeff;
```

osu.Framework/Resources/Shaders/sh_yuv2rgb.h
```diff
@@ -0,0 +1,16 @@
...
+uniform sampler2D m_SamplerV;
+
+uniform mat3 yuvCoeff;
+
+// Y - 16, Cb - 128, Cr - 128
+const mediump vec3 offsets = vec3(-0.0625, -0.5, -0.5);
+
+lowp vec3 sampleRgb(vec2 loc) 
+{
+  float y = texture2D(m_SamplerY, loc).r;
```
smoogipoo:
```suggestion
  lowp float y = texture2D(m_SamplerY, loc).r;
```

osu.Framework/Resources/Shaders/sh_yuv2rgb.h
```diff
@@ -0,0 +1,16 @@
...
+  float u = texture2D(m_SamplerU, loc).r;
```
smoogipoo:
```suggestion
  lowp float u = texture2D(m_SamplerU, loc).r;
```

osu.Framework/Resources/Shaders/sh_yuv2rgb.h
```diff
@@ -0,0 +1,16 @@
...
+  float v = texture2D(m_SamplerV, loc).r;
```
smoogipoo:
```suggestion
  lowp float v = texture2D(m_SamplerV, loc).r;
```

peppy:
@Pizzabelly planning on getting this merged in within the next few hours – do you intend to make further changes? If so, may be best adding them on a separate follow-up PR.

Pizzabelly:
@peppy sounds good. Everything for this PR is complete so if anything comes up I'll be sure to make a follow-up PR :)

peppy:
@Pizzabelly I would like to award a bounty for your work here if that's something you're interested in. Please fill in the form [here](https://docs.google.com/forms/d/e/1FAIpQLSet_8iFAgPMG526pBZ2Kic6HSh7XPM3fE8xPcnWNkMzINDdYg/viewform) if so.

Pizzabelly:
@peppy Thank you so much, this is an awesome thing you are doing here <3

PR #3020, opened 2019-11-21
Enforce CornerExponent of 2 for circles unless explicitly set

swoolcock:
There was a regression in rendering of circles that are not `CircularContainer`s (`RingPiece`, `CirclePiece` etc.), since they are not explicitly set to use a `CornerExponent` of 2.

This change will assume a `CornerExponent` of 2 if the `CornerRadius` is half of either of the `DrawSize` dimensions, unless the user has explicitly set a value for `CornerExponent`.

Are you ok with this change @Tom94 ?

Example of the pointy circles:
Screen Shot 2019-11-21 at 7 00 30 pm

### DIFF
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -1340,7 +1340,16 @@ namespace osu.Framework.Graphics.Containers
             }
         }
 
-        private float cornerExponent = 2.5f;
+        /// <summary>
+        /// Ensures that the exponent will always be 2 if the corner radius is half the smallest edge,
+        /// unless the user has explicitly set CornerExponent.
+        /// </summary>
+        private float effectiveCornerExponent =>
+            cornerExponent ?? (Precision.AlmostEquals(cornerRadius, Math.Min(DrawSize.X, DrawSize.Y) / 2f) ? 2f : default_corner_exponent);
+
+        private const float default_corner_exponent = 2.5f;
+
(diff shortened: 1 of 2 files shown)


swoolcock:
@Tom94 What I was thinking as a potential solution is to add a `bool CornerContinuous`. Then when rendering:

1) If `CornerContinuous` is false, always render with a `CornerExponent` of 2, regardless of the user's setting.
2) If `CornerContinuous` is true, and the user has not explicitly set a `CornerExponent`, always use 2.5f.
3) If `CornerContinuous` is true, and the user has explicitly set a `CornerExponent`, use that.

A `CircularContainer` should always force `CornerContinuous` to false, but I am unsure as to what the default should be.

Defaulting to false means the user needs to know about the property get the benefits.

Defaulting to true means that if the user makes circular components without using a `CircularContainer` (`RingPiece`, etc.) they will look bad unless the user explicitly sets `CornerContinuous` to false.

Tom94:
I don't really like that option as it has the same default-dilemma as `CornerExponent` 2 vs 2.5 has, only with the added complexity of the extra bool.

smoogipoo:
My $0.02 on this: corner exponent should be 2 by default, and 2.5 only if the user wants it. I don't think `CornerExponent` is the correct API to expose for this (in which situations would you use something that isn't 2 or 2.5?).

Tom94:
Larger values can look more pleasing on smaller buttons, whereas for larger buttons 2.5 is probably pretty good.

PR #3021, opened 2019-11-21
Replace all MathHelper.Clamp usages with Math.Clamp

turbedi:
.NET Standard 2.1 implements Math.Clamp , use it instead of MathHelper.Clamp from osuTK.

During quick testing I noticed that now an exception is thrown during the DelayedLoadWrapper visual test. 
[Exception](https://pastebin.com/14a2abPn)
I've added a commit that fixes the test, but I'm not completely sure if this is the correct fix.


osu.Framework/Graphics/Containers/ScrollContainer.cs
```diff
@@ -474,7 +474,7 @@ private void updatePosition()
             // then we should handle the clamping force. Note, that if the target is _within_
             // acceptable bounds, then we do not need special handling of the clamping force, as
             // we will naturally scroll back into acceptable bounds.
-            if (!IsDragging && Current != Clamp(Current) && target != Clamp(target, -0.01f))
+            if (!IsDragging && Current != Clamp(Current) && target != Clamp(target, 0.01f))
```
Tom94: Pretty confident this is not the correct fix. Probably best to check what the actual behavior was beforehand (i.e., how `MathHelper.Clamp` handled min>max) and manually implementing that behavior in `ScrollContainer.Clamp`.

peppy: yep, needs proper testing and understanding before this is merged.

turbedi: from osuTK:
`        public static float Clamp(float n, float min, float max)
        {
            return Math.Max(Math.Min(n, max), min);
        }`
So when min>max, then min is returned in every case.
Should I change it to how osuTK implemented it?

peppy: @Tom94 i'm having trouble understanding where the -0.01f comes from in the first place (and its purpose). the comment doesn't seem to touch on it and removing it has no discernible visual change.

Tom94: On quick inspection: removing the -0.01 seems to disable the strong clamping force (setting `localDistanceDecay` within the if-block) once `target` approached its clamped position. Pretty sure that'd be unintentional.

turbedi:
Reverted the behavior of ScrollContainer.Clamp, it now uses the old Clamp implementation from osuTK again.

osu.Framework/Graphics/Containers/ScrollContainer.cs
```diff
@@ -474,7 +474,7 @@ private void updatePosition()
             // then we should handle the clamping force. Note, that if the target is _within_
             // acceptable bounds, then we do not need special handling of the clamping force, as
             // we will naturally scroll back into acceptable bounds.
-            if (!IsDragging && Current != Clamp(Current) && target != Clamp(target, 0.01f))
+            if (!IsDragging && Current != Clamp(Current) && target != Clamp(target, -0.01f))
```
swoolcock: You should not be passing a negative value here as it makes no sense.  Fix `Clamp` such that it doesn't require a negative.

turbedi: I have just changed the value back to how it was before and I can't wrap my head around how to fix it correctly without breaking the strong clamp force

PR #1165, opened 2017-11-16
Fallback to no sound device if no output device is available

MattiaPette:
The game shouldn't crash when no audio devices are enabled on the system.

Resolves https://github.com/ppy/osu/issues/1302

PR #3025, opened 2019-11-24
Fix audio not working when no audio devices are present

phosphene47:
Fixes https://github.com/ppy/osu/issues/1302
Fixes https://github.com/ppy/osu/issues/2598
Fixes https://github.com/ppy/osu/issues/5866

Supercedes https://github.com/ppy/osu-framework/pull/1165

This is my attempt at fixing this long standing issue.

- Prioritizes the preferred audio device when initializing devices, falling back gracefully to system default on fail, and then "No sound" device provided by Bass itself as last resort.
- Resumes track automatically when falling back to "No sound", because Bass will stop the track for some reason.

I only have tested this by disabling all audio devices, unplugging them and using service manager to stop the audio service.

There are no unit tests included because I don't know how to reproduce an environment with no or broken audio devices in a unit test.

- [x] Ensure getting the current device every frame is not going to be a performance issue.

### DIFF
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -134,7 +134,7 @@ namespace osu.Framework.Audio
             {
                 updateAvailableAudioDevices();
                 checkAudioDeviceChanged();
-            }, 1000, true);
+            }, 100, true);
         }
 
         protected override void Dispose(bool disposing)
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -197,6 +197,16 @@ namespace osu.Framework.Audio
                 yield return Bass.GetDeviceInfo(i);
         }
 
+        private string noSoundDeviceNameCache;
+
+        /// <summary>
+        /// Gets the name of Bass "No sound" device.
+        /// </summary>
+        /// <remarks>
+        /// Although we can refer to this device by the string "No sound", this property can be used to avoid hardcoding that string.
+        /// </remarks>
+        private string noSoundDevice => noSoundDeviceNameCache ?? (noSoundDeviceNameCache = Bass.GetDeviceInfo(0).Name);
+
         private bool setAudioDevice(string preferredDevice = null)
         {
             updateAvailableAudioDevices();
@@ -206,7 +216,7 @@ namespace osu.Framework.Audio
 
             // use in the order: preferred device, default device, fallback "No sound" device (Bass ID: 0)
             // this allows us to initialize and continue using Bass without failing every subsequent Bass calls
-            newDevice ??= audioDevices.Find(df => df.IsDefault).Name ?? "No sound";
+            newDevice ??= audioDevices.Find(df => df.IsDefault).Name ?? noSoundDevice;
 
             bool oldDeviceValid = Bass.CurrentDevice >= 0;
 
@@ -244,7 +254,7 @@ namespace osu.Framework.Audio
             {
                 //the new device didn't go as planned. we need another option.
 
-                if (preferredDevice == "No sound")
+                if (preferredDevice == noSoundDevice)
                 {
                     //we're fucked. even "No sound" device won't initialise.
                     currentAudioDevice = null;
@@ -256,7 +266,7 @@ namespace osu.Framework.Audio
                     return setAudioDevice();
 
                 // default device failed: let's try using "No sound" device.
-                return setAudioDevice("No sound");
+                return setAudioDevice(noSoundDevice);
             }
 
             if (Bass.LastError == Errors.Already)
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -204,7 +204,7 @@ namespace osu.Framework.Audio
             string oldDevice = currentAudioDevice;
             string newDevice = preferredDevice;
 
-            // fall back to "No sound" device (Bass ID: 0) if no actual audio devices are found
+            // use in the order: preferred device, default device, fallback "No sound" device (Bass ID: 0)
             // this allows us to initialize and continue using Bass without failing every subsequent Bass calls
             newDevice ??= audioDevices.Find(df => df.IsDefault).Name ?? "No sound";
 
@@ -244,15 +244,19 @@ namespace osu.Framework.Audio
             {
                 //the new device didn't go as planned. we need another option.
 
-                if (preferredDevice == null)
+                if (preferredDevice == "No sound")
                 {
                     //we're fucked. even "No sound" device won't initialise.
                     currentAudioDevice = null;
                     return false;
                 }
 
-                //let's try again using the default device.
-                return setAudioDevice();
+                // preferred device failed: let's try using the default device.
+                if (preferredDevice != null)
+                    return setAudioDevice();
+
+                // default device failed: let's try using "No sound" device.
+                return setAudioDevice("No sound");
             }
 
             if (Bass.LastError == Errors.Already)
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -218,27 +218,22 @@ namespace osu.Framework.Audio
 
             if (oldDeviceValid)
             {
-                DeviceInfo oldDeviceInfo = Bass.GetDeviceInfo(Bass.CurrentDevice);
+                var oldDeviceInfo = Bass.GetDeviceInfo(Bass.CurrentDevice);
                 oldDeviceValid &= oldDeviceInfo.IsEnabled && oldDeviceInfo.IsInitialized;
             }
 
+            // same device
             if (newDevice == oldDevice && oldDeviceValid)
                 return true;
 
             int newDeviceIndex = audioDevices.FindIndex(df => df.Name == newDevice);
 
-            DeviceInfo newDeviceInfo = new DeviceInfo();
+            // preferred device might be unavailable
+            // in that case, continue using the old device if it is working, or fall back to default device
+            if (newDeviceIndex == -1)
+                return oldDeviceValid || setAudioDevice();
 
-            try
-            {
-                if (newDeviceIndex >= 0)
-                    newDeviceInfo = Bass.GetDeviceInfo(newDeviceIndex);
-                //we may have previously initialised this device.
-            }
-            catch
-            {
-            }
-
+            // initialize new device
             if (!InitBass(newDeviceIndex) && Bass.LastError != Errors.Already)
             {
                 //the new device didn't go as planned. we need another option.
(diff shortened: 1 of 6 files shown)


bdach:
The `No sound` device is an interesting find - what platforms did you test this on, though? BASS documentation doesn't really say a lot about it, except for the fact that it should be present on win32 and linux - no mentions about mobile or macOS.

Another thing to keep in mind is that per the docs some BASS operations will fail if the current device is the `no sound` one. From what I can see, it's mostly the `BASS_{format}_StreamCreate*` families of functions, which I haven't looked at in detail, but there's also `BASS_GetVolume` and `BASS_SetVolume`, which will return `BASS_ERROR_NOTAVAIL` when using device 0, and are imported in ManagedBass via P/Invokes and [used directly in `Bass.Volume`](https://github.com/ManagedBass/ManagedBass/blob/996a97706d63ac00a52e6e1be618c7c19cc985bd/src/Bass/Shared/Bass/PInvoke/Devices.cs#L214-L222). I haven't checked if we're using them, however.

osu.Framework/Audio/AudioManager.cs
```diff
@@ -204,8 +204,9 @@ private bool setAudioDevice(string preferredDevice = null)
             string oldDevice = currentAudioDevice;
             string newDevice = preferredDevice;
 
-            if (string.IsNullOrEmpty(newDevice))
-                newDevice = audioDevices.Find(df => df.IsDefault).Name;
+            // use in the order: preferred device, default device, fallback "No sound" device (Bass ID: 0)
+            // this allows us to initialize and continue using Bass without failing every subsequent Bass calls
+            newDevice ??= audioDevices.Find(df => df.IsDefault).Name ?? "No sound";
```
bdach: How about using the 0 ID in these checks instead? Feels slightly safer than a magic string.

phosphene47:
I've tested this on Windows 10.0.18362 Build 18362. I didn't realize this might be platform-specific, so I'll test on my mac soon.

I'm not sure about `...StreamCreate`, but it works on Windows for all I know. All unit tests are passing as well.

`BASS_GetVolume` and `BASS_SetVolume` is not used anywhere in the codebase. Volumes are managed using adjustments to the channels directly.

osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -141,13 +141,26 @@ void IBassAudio.UpdateDevice(int deviceIndex)
...
+
+            Interlocked.Exchange(ref currentTime, Bass.ChannelBytes2Seconds(activeStream, bytePosition) * 1000);
+
+            // update running state
+            var running = Bass.ChannelIsActive(activeStream) == PlaybackState.Playing;
 
-            Interlocked.Exchange(ref currentTime, Bass.ChannelBytes2Seconds(activeStream, Bass.ChannelGetPosition(activeStream)) * 1000);
+            // if we have been played, always say/lie that we are running
+            // this handles the case where Bass reports ChannelIsActive != Playing after output device changes
+            isRunning = running || (isPlayed && bytePosition != byteLength);
```
bdach: This sort of check slightly worries me. There's nothing in BASS docs that would say anything about this - what's the actual returned playback state when/after the device changes to `No sound`? `Stopped`, I presume?

osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -141,13 +141,26 @@ void IBassAudio.UpdateDevice(int deviceIndex)
         {
             Bass.ChannelSetDevice(activeStream, deviceIndex);
             Trace.Assert(Bass.LastError == Errors.OK);
+
+            // Bass may stop us if the output device changes (this is true for "No sound" device)
+            // so resume playing if we were playing previously
+            if (isPlayed && Bass.ChannelIsActive(activeStream) != PlaybackState.Playing)
+                Start();
```
bdach: These checks are sort of inconsistent. In `UpdateState()` the channel is considered to be running only in the `Playing` state, here it's considered to be running in `Playing | Stalled | Paused`. Should probably be `Playing | Stalled` in both places?

bdach:
If this goes in, we should probably also ban usage of `Bass.Volume` through `Microsoft.CodeAnalysis.BannedApiAnalyzers` for extra safety.

osu.Framework/Audio/Track/TrackBass.cs
phosphene47: > what's the actual returned playback state when/after the device changes to `No sound`?

From my observation, Bass has always stopped a track after changing the device to `No sound`. I do understand the concern from lack of reliable documentation from Bass however.

> Does it stay `Stopped` even after resuming in `UpdateState()`?

I think you meant `UpdateDevice()` because there's no resumption in `UpdateState()`. No, it will stay `Stopped` **until** resuming in `UpdateDevice()`. This is because device update is scheduled to run every 1 second (see ctor in `AudioManager`), and the track will stay stopped until that happens.

This check won't be necessary if device change detection is run every frame, but calling `Bass.GetDeviceInfo` every frame sounds expensive (needs bench.)

osu.Framework/Audio/AudioManager.cs
phosphene47: I think `"No sound"` is a reliable constant rather than a magic string. Bass will always return the same string, and changing that means they're breaking compatibility. I don't think it's any less safer than `0`.

osu.Framework/Audio/Track/TrackBass.cs
huoyaoyuan: Honestly, when representing closed-range things, a number is always safer than string.

bdach: I'm pretty sure we shouldn't be calling `GetDeviceInfo` every frame, yeah. I think this is as good as it might get since we seem to be forced to poll audio devices.

phosphene47: [This video](https://streamable.com/8x3cu) demonstates what happens without this condition in place.

phosphene47:
As for unit testing, I devised a basic approach.

Instead of creating a broken audio device or removing them at OS level which probably isn't feasible, I will extract out `Bass.Init` calls in `AudioManager` into a virtual method and create a derived class that overrides this call to return mock results.

phosphene47:
Unit tests have been added for playing tracks without devices, and losing a device while playing.

peppy:
In some quick testing, this works great for me. Will do a full code review soon!

peppy:
I've made some changes on top of your contributions:

## Test quality

- Share code where possible (via `AudioThreadTest` base class)
- Centralise wait-or-assert logic (and use it in place of `Thread.Sleep`)
- Remove all `Thread.Sleep`s. These are unreliable and potentially slow down tests.
- Improve reliability of is-progressing checks. This actually false-negatived for me due to the way it was being run.

## macOS support

In order to make this work correctly on macOS I changed the start-after-device-change code to perform a `Stop()` before the `Start()`. This is required because the track's state is not correctly updated to `Stopped`/`Stalled` after device loss.

I chose to remove the check conditional and perform this on all device changes, just for simplicity (if needed we could add back the conditional with the extra `deviceIndex` check: `if (isPlayed && (deviceIndex == 0 || !isRunningState(Bass.ChannelIsActive(activeStream))))`).

osu.Framework/Audio/AudioManager.cs
```diff
@@ -316,64 +320,26 @@ private void checkAudioDeviceChanged()
...
-                                    continue;
-
-                                if (setAudioDevice(d.Name))
-                                    break;
-                            }
-                        }
-                    }
-                }
-                else
+                updateAvailableAudioDevices();
```
peppy: this is now running 10x as often as before. is this required? if we want to keep it running this often, we should benchmark this to make sure it's within acceptable performance.

osu.Framework/Audio/AudioManager.cs
```diff
@@ -190,74 +189,73 @@ public ISampleStore GetSampleStore(IResourceStore<byte[]> store = null)
...
 
+        private string noSoundDeviceNameCache;
+
+        /// <summary>
+        /// Gets the name of Bass "No sound" device.
+        /// </summary>
+        /// <remarks>
+        /// Although we can refer to this device by the string "No sound", this property can be used to avoid hardcoding that string.
+        /// </remarks>
+        private string noSoundDevice => noSoundDeviceNameCache ??= Bass.GetDeviceInfo(Bass.NoSoundDevice).Name;
```
peppy: have you looked at using the fact that "No sound" is always at index zero, rather than relying on strings?

osu.Framework/Audio/AudioManager.cs
phosphene47: I'll post a benchmark of `updateAvailableAudioDevices` later.

FYI the reasoning behind this change is explained in https://github.com/ppy/osu-framework/pull/3025#discussion_r349923436.

e.g. if all audio devices were disconnected during osu! gameplay, the game clock would completely **stop** until device change is detected. Making detection more frequent can alleviate this to some degree. There could be a better way around this however.

peppy: can we not check validity of the current device like the old code, then trigger a full enumeration when this changes?

phosphene47: I'm not sure what you mean. The old code still triggered a full enumeration every second. Also, full enumeration is needed to detect new devices regardless of the validity of the current device.

peppy: The old code had two separate method calls. I'm proposing doing the "check valid" one more often (potentially every frame, even) and running enumeration at a less frequent interval – unless current device becomes invalid when it is run immediately.

osu.Framework/Audio/Track/TrackBass.cs
phosphene47: This isn't necessary anymore because device check will be run every frame https://github.com/ppy/osu-framework/pull/3025#discussion_r350570319.

peppy:
This looks to be in a very good state now. My only concern is the per-frame device validity check, so I will be benchmarking that before merging.

peppy:
image

Seems safe, on windows at least (1 sample over 100 seconds so ~0.032% usage). Let's go with it for now.

peppy:
Actually, on second check (switching profiling to be based on CPU instruction time) it is actually a considerable overhead:

image

Let's reduce the polling interval down to 100ms for safety.

phosphene47:
> Let's reduce the polling interval down to 100ms for safety.

Shall we bring this logic https://github.com/ppy/osu-framework/pull/3025#discussion_r350574351 back as well?

peppy:
How do things behave without that logic present?

phosphene47:
Because device validity check is done less frequently, when switching to "No sound" device, there will be a brief time where the track will be stopped, before we resume it manually.

See https://streamable.com/8x3cu

The logic is essentially lying that the track is still playing.

peppy:
I think that should be fine, as long as there's an inline comment explaining this edge case.

peppy:
Just a heads up that there was a [bounty offered](https://github.com/ppy/osu/issues/1302#issuecomment-534870714) on this issue, so please apply if that interests you.

PR #3026, opened 2019-11-24
Fix interpolation extremities on exponential and elastic curves

### LINKED ISSUE #3015: Exponential and Elastic easing types don't reach correct extremities
Two of the easing types (`Expo` and `Elastic`) provided by the framework don't return correct extreme values when the input (time) is 0 or 1. This can cause some weird visual jumping to occur.

https://www.desmos.com/calculator/h7nhxrctlu

Checking some [other](https://github.com/d3/d3-ease#easeExpIn) [implementations](https://github.com/tweenjs/tween.js/blob/master/src/Tween.js) [of](https://github.com/danro/jquery-easing/blob/master/jquery.easing.js#L74) [easing](http://hackage.haskell.org/package/ease-0.1.0.1/docs/src/Ease.html#expoOut) [curves](http://www.kiwijs.org/api/files/src_animations_tweens_easing_Exponential.ts.html#l43) all seem to use the same equations (or [equivalently broken](https://gillcup.readthedocs.io/en/latest/_modules/gillcup/easing.html#exponential) ones).

Tests available [here](https://github.com/peppy/osu-framework/commit/c009d9b73857fa84dea19883f80f9da7d3de6d6c).

bdach:
Resolves #3015.

# Summary

`Elastic` and `Expo` curves due to the formulae used would not stop exactly at 0 or 1 at time 0 or 1, as pointed out in tests in c84b164. Resolve this discrepancy by introducing small linear factors to the appropriate easing curves.

The linear factors offset the differences at the extremities that were wrong, and then fade to 0 at the other ends which were correct. A special case are the `InOut` easings, in which the factor will be fully included at both ends and then fade out linearly to 0 at `time = 0.5`.

Since the adjustments involved are minuscule (less than 4 significant decimal places), they should not be noticeable during transitions, but fix the jumping at extreme points seen before.

# Remarks

Aside from the unit tests originally written by @peppy (sorry for taking co-credit on that commit, I don't quite know why that happened - might have been due to me rebasing a bit to clean up history before push) I also included a test scene that allows for previewing all available curves for comparison reasons.

To check the visual effects you can either run the test scene or see the attached videos: [before](https://streamable.com/94t0e), [after](https://streamable.com/ct70d). The difference is quite subtle and will mostly be apparent on large sizes, and at the very ends of the transitions.

### DIFF
--- a/osu.Framework.Tests/MathUtils/TestInterpolation.cs
+++ b/osu.Framework.Tests/MathUtils/TestInterpolation.cs
@@ -2,6 +2,8 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Graphics;
 using osu.Framework.MathUtils;
@@ -12,6 +14,14 @@ namespace osu.Framework.Tests.MathUtils
     [TestFixture]
     public class TestInterpolation
     {
+        [TestCaseSource(nameof(getEasings))]
+        public void TestEasingStartsAtZero(Easing easing) => Assert.That(Interpolation.ApplyEasing(easing, 0), Is.EqualTo(0).Within(Precision.DOUBLE_EPSILON));
+
+        [TestCaseSource(nameof(getEasings))]
+        public void TestEasingEndsAtOne(Easing easing) => Assert.That(Interpolation.ApplyEasing(easing, 1), Is.EqualTo(1).Within(Precision.DOUBLE_EPSILON));
+
+        private static IEnumerable<Easing> getEasings() => Enum.GetValues(typeof(Easing)).OfType<Easing>();
+
         [Test]
         public void TestLerp()
         {
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneEasingCurves.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneEasingCurves.cs
@@ -41,7 +41,7 @@ namespace osu.Framework.Tests.Visual.Drawables
 
             AddSliderStep("resize easings", default_size, 3 * default_size, default_size, size =>
             {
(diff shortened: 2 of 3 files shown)


Tom94:
I'm not certain how good the JIT optimizations are, but it might make sense to bake all the constants into static variables to make sure they're not re-computed all the time. Pow is relatively expensive, after all. Perhaps worth a test.

bdach:
Yeah, I figured it might be worth it. Will benchmark and report back.

bdach:
Benchmarking results ([see benchmark code](https://gist.github.com/bdach/0edda878ee2664faf2bbd0056fdfe9ca)):

### Before extractions to constants

|      Method |            Easing |     Mean |   Error |   StdDev |
|------------ |------------------ |---------:|--------:|---------:|
| ApplyEasing |            InExpo | 101.4 us | 2.00 us |  3.71 us |
| ApplyEasing |         InElastic | 114.5 us | 2.19 us |  2.34 us |
| ApplyEasing |    OutElasticHalf | 128.5 us | 2.55 us |  4.05 us |
| ApplyEasing | OutElasticQuarter | 127.6 us | 2.55 us |  3.22 us |
| ApplyEasing |      InOutElastic | 134.7 us | 3.95 us | 11.65 us |

### After extractions to constants

|      Method |            Easing |     Mean |    Error |   StdDev |
|------------ |------------------ |---------:|---------:|---------:|
| ApplyEasing |            InExpo | 46.87 us | 0.635 us | 0.594 us |
| ApplyEasing |         InElastic | 64.19 us | 1.146 us | 0.957 us |
| ApplyEasing |    OutElasticHalf | 62.81 us | 0.950 us | 0.889 us |
| ApplyEasing | OutElasticQuarter | 59.69 us | 0.533 us | 0.499 us |
| ApplyEasing |      InOutElastic | 65.64 us | 0.658 us | 0.616 us |

Yup, seems like a definite win performance-wise; just pushed that in.

PR #3037, opened 2019-11-27
Move common logic to base implementation of GlyphStore

peppy:
- [x] Depends on https://github.com/ppy/osu-framework/pull/3035
- [x] Depends on https://github.com/ppy/osu-framework/pull/3036

Allowing for better testing of different caching / parsing algorithms. `GlyphStore` can also now be used to load with zero caching, as a baseline.

This also does some clean-up on `RawCachingGlyphStore` for readability purposes.

### DIFF
--- a/osu.Framework/IO/Stores/GlyphStore.cs
+++ b/osu.Framework/IO/Stores/GlyphStore.cs
@@ -83,9 +83,12 @@ namespace osu.Framework.IO.Stores
 
         protected virtual Image<Rgba32> GetPageImageForCharacter(Character character)
         {
-            using (var stream = Store.GetStream($@"{AssetName}_{character.Page.ToString().PadLeft((Font.Pages.Count - 1).ToString().Length, '0')}.png"))
+            using (var stream = Store.GetStream(GetFilenameForPage(character.Page)))
                 return TextureUpload.LoadFromStream<Rgba32>(stream);
         }
+
+        protected string GetFilenameForPage(int page)
+            => $@"{AssetName}_{page.ToString().PadLeft((Font.Pages.Count - 1).ToString().Length, '0')}.png";
 
         public CharacterGlyph Get(char character)
         {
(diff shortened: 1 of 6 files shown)


osu.Framework/IO/Stores/RawCachingGlyphStore.cs
```diff
@@ -0,0 +1,135 @@
...
+                    return pageLookup[character.Page] = new PageInfo
+                    {
+                        Size = new Size(int.Parse(split[2]), int.Parse(split[3])),
+                        Filename = existing
+                    };
+                }
+
+                using (var convert = GetPageImageForCharacter(character))
+                {
+                    // todo: use i# memoryallocator once netstandard supports stream operations
```
peppy: Removed in https://github.com/ppy/osu-framework/pull/3042.

osu.Framework/IO/Stores/GlyphStore.cs
```diff
@@ -1,67 +1,66 @@
-﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
```
smoogipoo: Double license header

osu.Framework/IO/Stores/RawCachingGlyphStore.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        protected override TextureUpload LoadCharacter(Character character)
+        {
+            if (!pageLookup.TryGetValue(character.Page, out var pageInfo))
+                pageInfo = createCachedPageInfo(character);
+
+            return createTextureUpload(character, pageInfo);
+        }
+
+        private PageInfo createCachedPageInfo(Character character)
```
smoogipoo: I was confused as to how this was actually working because this seems like it caches per character, whereas that's not actually what's happening.

Pass in the page instead.

PR #3038, opened 2019-11-27
Add benchmarks project

peppy:
Provides a permanent home for benchmarks so we stop throwing them away and are forced to keep the up-to-date with the current codebase. Can be run via nUnit (and will probably require exclude rules to be added so as to not run on CI for now).

- [x] Depends on https://github.com/ppy/osu-framework/pull/3037
- [x] Exclude from CI runs
- Adds benchmarks for font loading, including adding back `TimedExpiryGlyphStore` for comparison purposes.


huoyaoyuan:
There should be documentation about how to run one benchmark with seeing console output. Running them via IDE may not show console.

peppy:
Good point, will add.

huoyaoyuan:
To view console output from things executed by nunit in VS:
Go to Test Explorer, expand the "test", and select one in the inner-most level(should be a method-parameter group). Then click "Open additional output of this result" in the Test Detail Summary section.

huoyaoyuan:
Another consideration: use launch profiles of IDE, using console argument to specify which benchmark to run.
It increases maintenance work when adding new benchmark, but provides perfect availability to both console and IDE.

https://github.com/dotnet/performance/blob/17a1a8fd550733c1e0eea7e410cb503f8f41f5e4/src/benchmarks/micro/Program.cs#L35-42

Seems benchmarkdotnet provides some support of that.

huoyaoyuan:
Remarks for VS launch settings:
`<project dir>/Properties/launchSettings.json`
```json
{
  "profiles": {
    "osu.Desktop": {
      "commandName": "Project"
    },
    "newProfile1": {
      "commandName": "Project",
      "commandLineArgs": "--arg1"
    }
  }
}
```

huoyaoyuan:
Updates pushed at https://github.com/huoyaoyuan/osu-framework/tree/benchmark-project .

image

Since Rider knows VS launch config now, can we convert to use VS ones? It looks much simpler.
Needs confirmation about Linux/Mac Rider.

peppy:
Can they be defined as a tree?

image

I like our rider layout currently.

huoyaoyuan:
No, VS launch settings are associated with projects, not solutions.
Let's keep them and add VS ones for necessary.

PR #2395, opened 2019-05-11
Implement initial OML support

holly-hacker:
Add an initial implementation for osu! markup language, which reads yaml files and creates Drawables from them.

---

Adds a partial implementation for the osu! markup language as described in #30. It deviates from the spec in the following ways:
- Transitions are not implemented. I've been thinking about how to implement it for a few days and I can't come up with a way to combine an Action with an arbitrary Delegate. Any help would be greatly appreciated.
- The Type field in Properties is not used, since it's derived from the found field/property.
- The Default state is not yet applied.

OmlReader's constructor accepts a Dictionary<string,Type> for resolving Drawable types and Dictionary<string,Delegate> which can be assigned to events. The only properties that can be set are the types supported by TypeConverter (primitives, Enums), Color4/ColourInfo (in RGB/RGBA hex format or as static field of Color4) and Vector2.

This probably isn't ready for merging yet, I've opened this PR to get some feedback. This implementation may throw a bit eagerly and there are no unit tests yet.

Issue #30, opened 2016-09-05
osu! markup language


smoogipoo:
Initial draft spec of OML. It's changed a little bit since the last iteration in the transitional repository so you'd best go over this again and point out any inconsistencies/points of clarification or just general comments. Topic for the next pair programming session.

---
# OML - osu! markup language
## What is OML

OML is a markup language very similar to [QML](http://doc.qt.io/qt-5/qmlreference.html), oriented towards designing layouts for games that are built upon the osu!framework. The language is YAML-Validated and generates appropriate C# code upon compilation.
## Basic syntax

The following is an _object_ that represents a 100x100 red rectangle:

```
Rectangle:
  Width: 100
  Height: 100
  Colour: Color.Red
```

`Width`, `Height` and `Colour` are _general_ properties of the Rectangle. These are applied when the object is constructed.
### Special Properties

Objects have _special_ properties: 
`Extends`
`States`
`Transitions`
`Events`
`Properties`
`Children`
#### The `Extends` Property

Objects extend `Drawable` by default and inherit all of `Drawable`'s properties. It is possible to change this behaviour where this is not intended by specifying the `Extends` property of the object:

```
Rectangle:
  Extends: Box  # Rectangle will now inherit all of Box's properties
```
#### The `States` Property

A state defines how an object looks at a point in time by modifying the value of the general properties of the object. This special property may be used to define _named_ states which may be referred to by multiple places in further code.  
The property named "Default" is optional but will be applied when the object is first constructed.

```
Rectangle:
  States:
    - Name: Hovered
      Colour: Color.Red
    - Name: Default       # Optional state to be applied when the object is first constructed
      Colour: Color.Blue
```
#### The `Transitions` Property

A transition defines how an object should transition towards a state. This special property may be used to define _named_ transitions which may be referred to by multiple places in further code. These can be used to transition to either named or anonymous states.

```
Rectangle:
  States:
    - Name: Hovered
      Colour: Color.Red
  Transitions:
    - Name: HoveredTransition
      State: Hovered           # Named state
      Duration: 500
    - Name: ScaleTransition
      State:                   # Anonymous state
        Scale: 1.5
```
#### The `Events` Property

An event defines the transitions that should be applied to an object when the event is called. Events are exposed in the generated C# code. These can be used to apply either named or anonymous transitions.

```
Rectangle:
  States:
    - Name: Hovered
      Colour: Color.Red
  Transitions:
    - Name: HoveredTransition
      State: Hovered
[...]

PR #3041, opened 2019-11-27
[WIP] Initial implementation of OML (osu! Markup Language)

Mempler:
# Initial implementation of an Markup Language for osu.Framework
#### inspired of #2395 and #30 but with a Completely different design.

It is very feature incomplete and still requires a lot of work until it's mergeable.
OML is HTML Styled but not the same, i hope to Improve the HTML Style in the future.
I've chosen to not work with #2395 because the Initial Design was way too different to be merged into one.

Example:
```xml
<oml>
    <container width="250" height="250"> <!-- a Container, also known as div in html. -->
        <sprite texture="https://a.ppy.sh/10291354" width="25" height="25" margin="25" />
        <box width="250" height="250" color="orange" />
    </container>
</oml>
```

## TODO:
* [x] Images
* [x] Containers
* [x] Box
* [x] Circle
* [x] Nested Child Containers
* [x] Buffered Containers
* [x] Video Sprites
* [ ] Style Sheets
* [ ] Transitions
* [ ] States, Events, Callbacks, all the Scripting things.
* [ ]  (CANCELED) ~C# Scripting~
* [ ] Properly written visual tests.

want more core features ? write it down. i'll take a look at them.


phosphene47:
For reference, this is quite similar to a [project](https://github.com/chiyadev/osu-design) I created before, but never worked on further.

swoolcock:
You can't make `BufferedContainer` a default.  It's far too heavyweight and should only be used when you specifically want a renderbuffer for some reason.

swoolcock:
Some suggestions:

* The markup syntax should match actual class and property names where possible, as it will make it easier for developers when jumping between code and markup.  Case-insensitive is fine.
* Would prefer not to have more than one tag per object as it will pollute the list of available tags and confuse developers.
* `` would be better as `<sprite texture="" />` for reasons noted above.

Mempler:
> * Would prefer not to have more than one tag per object as it will pollute the list of available tags and confuse developers.

you mean like, Removing the aliases ? E.G ` == <sprite>` and i used `src=""` instead of `texture=""` because i wanted it to make it as HTML as possible since most people have at least a little knowledge of HTML.

swoolcock:
> i used `src=""` instead of `texture=""` because i wanted to make it as HTML as possible since most people have at least a little knowledge of HTML.

This is osu!framework, not HTML.  The less ambiguity the better.

osu.Framework.Tests/Visual/OML/OmlBox.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace osu.Framework.Tests.Visual.OML
 {
-    public class BoxTest : TestScene
+    public class OmlBox : TestScene
```
swoolcock: I don't think these class renames were intentional.

Mempler: Nope, i actually renamed them to fit the Normal Test cases.

Mempler: I'll rename them again because i missed `Test` in front of them. whoops

osu.Framework/MathUtils/UnitConverter.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// Helper function for Unit Conversation.
+    /// </summary>
+    public static class UnitConverter
+    {
+        // https://www.w3schools.com/cssref/css_units.asp
+        /// <summary>
+        /// Computes Centimetre to Pixels
+        /// </summary>
+        /// <returns>Pixels</returns>
+        public static double Cm2Pixel(double cm) => cm * 37.795275591d; // this could be actually wrong. as i calculated it in my head.
```
swoolcock: This is completely device dependent.  You can't convert physical distances to/from pixels without knowing the DPI of the display.  I guarantee you this will be wildly inaccurate on any kind of retina device.

Mempler: Yea, i thought so. any idea how i may get the DPI ?

swoolcock: I think this is an XY problem.  What are you actually trying to do here?

peppy:
Not sure about this being a pull request on osu-framework. I don't see this being merged as it is not the direction we are looking to go in (for one our implementation would not be xml).

You might be best continuing developing this on your own repository, but not calling it OML (don't hesitate to reach out for help or advice via discord or issues here, though).

peppy:
Actually, keeping it here may be okay if you ensure that the parser is  modular so the source does not need be the xml format.

Mempler:
Well that would be bad since i spend multiple hours on this. would be a waste if it gets thrown away as i'm working on Style Sheets atm.
```
# @ = EventFunction

var backgroundColor = rgba(50, 50, 50, .2);

root {
    colour: $backgroundColor
}

CoolName {
    width: 50;
    height: 50;

    colour: rgba(40, 40, 40, .2);
}

# Implements OnClick Event
CoolName : @OnClick {
    FadeTo(.24 * width, 400ms) # s = second, ms = millisecond, ns = nanosecond
}
```
as an example

Mempler:
The parser is modular, it can be easily reworked.

Mempler:
Lemme clarify why i closed this PR.
I've closed this PR due to Peppy don't want shit like this.. Cya

PR #3044, opened 2019-11-27
Update iOS native packaging

### LINKED ISSUE #3030: iOS framework package duplicates native libraries
image

It's eating a lot of disk space if you have iOS project loaded at game side. **408MB when the nupkg is expanded and preserved**.
At Android side, the native libraries are compressed as embedded resources of the assembly.
The two packaging directories are different flavors of old and new project system. I'm not sure which one takes effect.

I'm not sure if we can really solve this now. Maybe waited until .NET 5 supports better on iOS.

huoyaoyuan:
Fixes #3030 .

I'm essentially following https://github.com/ericsink/SQLitePCL.raw/tree/master/src/SQLitePCLRaw.lib.e_sqlite3.ios , since sqlite should be the most used package with native asset in .NET world.

Please do a manual test because I cannot, neither CI does.

The change packs native libraries with assembly instead of nupkg. Test results on project reference should be safely applicable on package reference at game side.

- [x] Test framework tests on iOS
- [x] Test osu! on iOS via project reference (local framework)
- [x] Test osu! on iOS via nuget package


swoolcock:
I know there was a reason I didn't do it this way, but I can't remember what it was.  It might be worth moving the `LinkerFlags` in there too.

huoyaoyuan:
Another downside of using `NativeReference` is that Windows build will try to explain them as something else, and then provide warnings. The name conflicts with some COM stuff.

frenzibyte (dismissed):
Size dropped from 150.3MB to 75.2MB, native libraries don't get duplicated anymore. 👍🏼

Thoroughly tested with clean builds to ensure no leftover libraries in mtouch cache interfere. Can confirm all libraries have been linked properly, along with the specified linker flags and frameworks.

frenzibyte:
Blocking momentarily while I figure out if building on Release configuration has actually been busted with this PR, or has always been the case.

frenzibyte:
Unblocking as the issue appeared to be unrelated from this PR, will resolve in a separate one.

frenzibyte:
Actually templates can't be updated without a framework release with this PR included first.

frenzibyte (approved):

peppy (approved):

PR #3045, opened 2019-11-27
Fix one remaining case of incorrect audio testing

peppy:
Had a feeling I missed one..

https://ci.appveyor.com/project/peppy/osu-framework/builds/29163136/tests


phosphene47:
Sorry I should have caught that 😓

peppy:
@phosphene47 also for future reference, the `expected` value goes first in an `Assert.AreEqual`. or use `Assert.That(1, Is.EqualTo(1))` instead for clarity.

phosphene47:
Noted.

PR #3046, opened 2019-11-28
Refactor decoupled clock to better handle negative seeks

smoogipoo:
`DecoupledInterpolatingFramedClock` couldn't handle negative seeks at all.

During `Seek()`, it stops the source clock and calls `ProcessFrame` which attempts to start the source clock. This only happens if the seek on the source clock to `CurrentTime` succeeds.  

The problem is that the source clock is seeked to the `CurrentTime` from the _previous_ frame (i.e. before the `Seek()`), which would always succeed unless `CurrentTime` was negative in the first place.  

It would then follow through and use the interpolated source time, which would be the time from the last frame.

### DIFF
--- a/osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs
+++ b/osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs
@@ -1,8 +1,6 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System;
-
 namespace osu.Framework.Timing
 {
     /// <summary>
@@ -106,7 +104,7 @@ namespace osu.Framework.Timing
             }
 
             elapsedFrameTime = elapsedTime;
-            currentTime = elapsedTime < 0 ? Math.Min(currentTime, proposedTime) : Math.Max(currentTime, proposedTime);
+            currentTime = proposedTime;
         }
 
         public override void ChangeSource(IClock source)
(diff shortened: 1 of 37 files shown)


osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs
```diff
@@ -62,40 +64,46 @@ public override void ProcessFrame()
...
 
-            double proposedTime = useInterpolatedSourceTime ? base.CurrentTime : decoupledClock.CurrentTime;
+                // since the decoupled clock was started, update it once more.
+                decoupledClock.ProcessFrame();
+            }
 
             elapsedFrameTime = useInterpolatedSourceTime ? base.ElapsedFrameTime : decoupledClock.ElapsedFrameTime;
-
-            currentTime = elapsedFrameTime < 0 ? proposedTime : Math.Max(currentTime, proposedTime);
+            currentTime = elapsedFrameTime < 0 ? ProposedCurrentTime : Math.Max(currentTime, proposedTime);
```
peppy: the only thing i'm not sure about is the reasoning behind using `ProposedCurrentTime` when negative here

peppy: All tests pass even after changing this to the local `proposedTime`, so gonna leave this PR until it's clarified.

smoogipoo: Ah yeah my eyes glossed over that, should be the local one, but good point there needs to be a test for this.

smoogipoo:
Whoops, that was a bad rebase. My bad - not gonna force-push to fix it.

smoogipoo:
I've refactored this a little bit further - haven't added any additional tests or anything, just consolidating the logic and making sense of it. All existing test cases passed and from `TestSceneClocks` it all seemed ok.

Issue #3047, opened 2019-11-28
Non-2.5 CornerExponent bounds are not accurate for rotated ellipsoids


smoogipoo:
Here's a comparison of the various rotational degrees:

| angle | Exponent=0 | Exponent=1 | Exponent=2 | Exponent=2.5 | Exponent=6 |
| ------ | ------------ | ------------ | ------------ | -------------- | ------------ |
| 45 | ![image](https://user-images.githubusercontent.com/1329837/69788879-bbe80100-1202-11ea-930f-a87ff92bb71f.png) | ![image](https://user-images.githubusercontent.com/1329837/69788894-c0acb500-1202-11ea-91d6-22691d77f0a1.png) | ![image](https://user-images.githubusercontent.com/1329837/69788901-c5716900-1202-11ea-93a8-ef6bdcbc59f3.png) | ![image](https://user-images.githubusercontent.com/1329837/69788913-ca361d00-1202-11ea-912d-804331019070.png) | ![image](https://user-images.githubusercontent.com/1329837/69788932-d15d2b00-1202-11ea-92f3-2cd20b6450bd.png) |
| 20 | ![image](https://user-images.githubusercontent.com/1329837/69788940-d8843900-1202-11ea-959c-0da927b0c5d9.png) | ![image](https://user-images.githubusercontent.com/1329837/69788948-dcb05680-1202-11ea-9dbb-0a8cd0cb158c.png) | ![image](https://user-images.githubusercontent.com/1329837/69788952-e043dd80-1202-11ea-8ca1-755db776578c.png) | ![image](https://user-images.githubusercontent.com/1329837/69788962-e639be80-1202-11ea-9723-819fd7a66685.png) | ![image](https://user-images.githubusercontent.com/1329837/69788965-e934af00-1202-11ea-936d-b17d0f6be301.png) |

Tom94:
This is known/intentional. Turns out, there's no closed-form solution of bounds for general super-ellipsoids out there, so the code uses a cheap approximation.

PR #3048, opened 2019-11-28
Change CornerExponent to default to 2

smoogipoo:
Prevents regressions. See: https://github.com/ppy/osu-framework/pull/3020, https://github.com/ppy/osu-framework/issues/3047

### DIFF
--- a/osu.Framework/Graphics/Containers/CircularContainer.cs
+++ b/osu.Framework/Graphics/Containers/CircularContainer.cs
@@ -15,7 +15,6 @@ namespace osu.Framework.Graphics.Containers
             // this shouldn't have to be done here, but it's the only place it works correctly.
             // see https://github.com/ppy/osu-framework/pull/1666
             CornerRadius = Math.Min(DrawSize.X, DrawSize.Y) / 2f;
-            CornerExponent = 2;
 
             return base.GenerateDrawNodeSubtree(frame, treeIndex, forceNewDrawNode);
         }
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -1340,11 +1340,11 @@ namespace osu.Framework.Graphics.Containers
             }
         }
 
-        private float cornerExponent = 2.5f;
+        private float cornerExponent = 2f;
 
         /// <summary>
-        /// Determines how gentle the curve of the corner straightens. A value of 2 results in
-        /// circular arcs, a value of 2.5 (default) results in something closer to apple's "continuous corner".
+        /// Determines how gentle the curve of the corner straightens. A value of 2 (default) results in
+        /// circular arcs, a value of 2.5 results in something closer to apple's "continuous corner".
         /// Values between 2 and 10 result in varying degrees of "continuousness", where larger values are smoother.
         /// Values between 1 and 2 result in a "flatter" appearance than round corners.
         /// Values between 0 and 1 result in a concave, round corner as opposed to a convex round corner,


Tom94:
I _really_ wouldn't call the bounds thing a regression for `CornerExponent = 2.5`. The inaccuracy is pretty much imperceptible.

smoogipoo:
If it's not mathematically accurate, I don't believe in it being the default.

Tom94:
I disagree with that statement on so many levels.

Our bezier approximations of sliders are inaccurate in the same bounded way as these corner exponents of 2.5 are. The point is for corners to look accurate and for the worst-case behavior to be bounded to be imperceptible.

Tom94:
I'm much in favor of a solution similar to @swoolcock 's PR: https://github.com/ppy/osu-framework/pull/3020

Only that instead of a hard-decision to use 2.0 for circles and 2.5 for everthing else, I'd make it a smooth transition.

smoogipoo:
It is actually _very_ on retina. Put it this way, I didn't just make these changes after carefully examining the pixels, I noticed it while going through test cases.

Here's how it looks on retina:

image

I assume it looks _even worse_ at higher resolutions.

Tom94:
You only run into that situation for near-circular things, where I agree a CornerExponent of 2 is more appropriate. See my reference of Samah's PR above.

smoogipoo:
Want to explain to me what "near-circular" means to you?

image

image

smoogipoo:
Also

> Our bezier approximations of sliders are inaccurate in the same bounded way

As far as I'm aware, they are not. Maybe the actual tracing of the path around the bezier curve is not accurate, but the bounding box _is_. Also, sliders are not a UI component so far as boxes.

Tom94:
Without zooming in, I don't actually see the gaps on my retina screen. I honestly couldn't tell the difference between the perfectly-touching 2.0 exponent and the 2.5 exponent in this case.

Regarding the Bezier: pretty sure the bounding box is based on the path and not on the actual bezier equation. If we used the bezier equation directly, we'd run into numerical problems due to all the exponentials with paths longer than 10-20 vertices.

My point is that we use approximations everywhere. The goal should be visual accuracy, not numerical accuracy. Whether that invalidates your concerns here, I do not claim, but I object the hard requirement of mathematical correctness. (And I am saying this as the guy who loves to make things mathematically accurate.)

Tom94:
... and the whole discussion completely disregards the question of whether the desire for pretty continuous corners happens more often in the wild than the desire for 100% accurate bounds of rotated containers with rounded corners. Do we have any such use case in osu!? I can't think of one of the top of my hat.

Imho, when there is a tradeoff, the default should cater to the more frequent use case.

peppy:
Ignoring mathematics or anything, let's consider a user approaching the framework:

They want to make a rounded corner button and discover `CornerExponent`. As they set it higher, it starts to look awesome and they choose to leave it at `14` because of this.

Then later in development they start to get reports that users can click outside the button, due to the (easily visible) inaccuracies at the higher exponent.

I don't think this is a great thing to expose if we can't guarantee at least <1px input accuracy. At 2.5 it seems to be roughly within expectation (around 1-1.5 retina pixels of inaccuracy) so I am indifferent there, but going higher seems to be an issue.

Tom94:
Users won't be able to click into the button. Hover is 100% accurate. Only the autosize bounds aren't.

peppy:
In that case, adding a remark in xmldoc mentioning the inaccuracy is likely enough.

peppy:
I'm going to make a call here and revert this one as a first step forward (to avoid the current status-quo which is causing weird behaviour in framework consumers that aren't using `CircularContainer`). I'd rather have things in a "working as before" state while we hone the behaviour into something we can all agree on.

A next step would be making (or updating @swoolcock's [PR](https://github.com/ppy/osu-framework/pull/3020)) to add the mentioned changes to make it ease from 2.5 to 2 rather than make a sudden jump.

PR #3049, opened 2019-11-28
Add methods for converting radians to degrees and vice versa and rename MathUtils namespace to Utils

turbedi:
Replaces the MathHelper-functions from osuTK.
Removes all usages of osuTK.MathHelper for osu-framework together with PR #3021.
Renames the MathUtils namespace to Utils, otherwise it will create conflicts


huoyaoyuan:
Should be placed in some class under `MathUtils`? `Extensions` should contains only extension methods, not "normal" methods.

turbedi:
Oh, you're right. How should I name the class? `NumericMethods`?

huoyaoyuan:
Maybe just `Angles`? Let English speaker decide if there's better word.

turbedi:
Moved the class to MathUtils and renamed it to Angles. Is it fine now?

swoolcock:
To be honest I would just put all the ex-`MathHelper` functions in a partial `MathUtils` class.  I don't see the need for a separate `Angles` class. etc.

turbedi:
So far we don't use any other functions from MathHelper except `Swap` which is used once, not sure if it is worth to add this function just for one usage and rename the class to `MathHelper`. Any other opinions/suggestions?

peppy:
I also disagree with the "Angles" class. It's a hard name to find/remember. `MathUtils` sounds best.

smoogipoo:
`MathUtils` conflicts with the namespace. Perhaps we should rename the namespace to `Utils`? Though it's a pretty big change.

Otherwise I'd suggest `Trigonometry` or the original `Angles`, or even `Angle` with renames of methods to `ToRadians()` and `ToDegrees()`.

swoolcock:
> `MathUtils` conflicts with the namespace. Perhaps we should rename the namespace to `Utils`? Though it's a pretty big change.

I agree with renaming the namespace.

turbedi:
Renamed the namespace from `MathUtils` to `Utils`, build now passes.

PR #3050, opened 2019-11-28
Revert assembly load logic change to attempt to fix tests

peppy:


huoyaoyuan:
The game side usages can get better fix by switching to `typeof(XXX).Assembly` to get rid of loading logic.
Despite of that, agree for reverting the exact-same behaviour.

peppy:
Can you make a PR showing how that should be done? It's not immediately obvious to me how that would work in `RulesetStore`.

huoyaoyuan:
Looks so weird. `RulesetStore` shouldn't have any path calling this. No idea.

May get resolved if `public virtual IResourceStore<byte[]> CreateResourceStore() => new NamespacedResourceStore<byte[]>(new DllResourceStore(GetType().Assembly.Location), @"Resources");` gets updated.

peppy:
Without looking to deep, I'd argue it's something to do with the way assemblies are loaded via the different methods, causing them to be incorrectly reloaded.

I remember now that when writing the `DllResourceStore` code, I had to do it in a *very* specific way to make NUnit happy.

huoyaoyuan:
At my sight, we should avoid loading assembly ourselves anywhere, unless extremely necessary. `typeof` guarantees a correct result when the type is available.

peppy:
That sounds fair as a direction forward. DllResourceStore may be best replaced/renamed to `AssemblyResourceStore` and take a type/generic if it turns out to be the culprit.

PR #1645, opened 2018-06-24
Add library loading and load libbass.so globally on linux

Criminalllz:
Because libbass_fx does not specify libbass as a dependency, we need  to load libbass globally so that libbass_fx knows where to look for it. This patch also adds a way to load libraries with some of the flags of dlopen.

- Closes https://github.com/ppy/osu/issues/2852

---

Fix beatmaps not playing on linux.

PR #3057, opened 2019-11-30
Fix startup crash regression on Linux

fishfishfish2104:
This reverts commit 6b8ff8a1809286ada21b541defafe74049c478c0. which fixes https://github.com/ppy/osu/issues/7018


swoolcock:
For future reference:

* When making a pull request, work from a new branch (not master).
* Putting fixes/closes/resolves etc. in the pull request comments (as you have) will automatically close the related issues.  You don't need to manually do it.

fishfishfish2104:
thanks for the tips

huoyaoyuan:
Surprised that this does cause regression. The native package change did influence Linux bass.
The problem seems not "can't find bass", but "bass library not loaded when required" instead.

bdach:
See original discussion under ppy/osu#2852, #1645 and linked dotnet issues.

Issue #2448, opened 2019-05-27
Decouple GameWindow from osuTK


swoolcock:
Proposal to decouple `GameWindow` from osuTK, as a first step toward removing osuTK entirely.

Currently `GameWindow` already abstracts away the native window to a degree, but it still implements the osuTK `IGameWindow` interface, which we will want to replace.

I propose that we make our own `IGameWindow` interface (better name pending) that has nothing to do with osuTK, but will provide a similar functionality such that we can avoid referencing the `GameWindow` class (or its subclasses) entirely.  This also gives us the opportunity to decide which window management features we actually want/need in the interface.  osuTK's `IGameWindow` is currently rather bloated.

We will be able to keep the current osuTK-coupled concrete `GameWindow` class for now, but the next step would be to potentially provide an SDL2 implementation using SDLSharp or similar. https://github.com/flibitijibibo/SDL2-CS

Related to #2119

peppy:
We haven't yet decided a path forward as for what we are going to do about the backend, but SDL (+veldrid), OpenTK 4.0 etc. are possibilities.

While this isn't an imminent focus for us (would rather get android pushed out and continue game implementation) you are more than welcome to continuing abstracting the remaining opentk links we have, as I see this as a good move regardless of the path forward.

swoolcock:
As part of reorganising the exposed events and bindables, I’m thinking that key and mouse events should not be part of the `IWindow` interface, as we expose those through input handlers.  The concrete window implementation would delegate OS input events through to the handlers automatically.

swoolcock:
Pushed WIP branch for discussion:
https://github.com/swoolcock/osu-framework/tree/gamehost-refactor

swoolcock:
After some more work on this I've successfully split the window/audio/input/video initialisations and encapsulated them mostly within the backend implementations.

As for graphics, most of the current OpenGL references are within the `osu.Framework.Graphics.OpenGL` namespace, and already wrapped by the framework.  I'm hesitant to redirect it through an abstraction layer, as we would essentially be cloning what Veldrid does already.

Replacing the graphics API is already going to require rewriting shaders to a degree, so I feel it would be better to just make a decision on which path we are taking and replace osuTK in one go.

swoolcock:
Note that Veldrid uses `System.Numerics` for vector primitives, and has its own `RgbaFloat` and `RgbaByte` colour structs.

huoyaoyuan:
[...]

PR #3059, opened 2019-12-01
Replace macOS native osuTK reflection

swoolcock:
Replaces the reflected osuTK calls in `MacOS.Native` with real implementations.  Required as part of dropping osuTK.  Implementations loosely borrowed from osuTK.

Note that the `FoundationLibrary` reference is not required and has been removed.

Related to #2448

### DIFF
--- a/osu.Framework/Platform/MacOS/Native/Cocoa.cs
+++ b/osu.Framework/Platform/MacOS/Native/Cocoa.cs
@@ -2,16 +2,18 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Linq;
-using System.Reflection;
 using System.Runtime.InteropServices;
 
 namespace osu.Framework.Platform.MacOS.Native
 {
     internal static class Cocoa
     {
+        internal const string LIB_DL = "libdl.dylib";
+        internal const string LIB_APPKIT = "/System/Library/Frameworks/AppKit.framework/AppKit";
         internal const string LIB_OBJ_C = "/usr/lib/libobjc.dylib";
         internal const string LIB_CORE_GRAPHICS = "/System/Library/Frameworks/CoreGraphics.framework/Versions/Current/CoreGraphics";
+
+        internal const int RTLD_NOW = 2;
 
         [DllImport(LIB_OBJ_C, EntryPoint = "objc_msgSend")]
         public static extern IntPtr SendIntPtr(IntPtr receiver, IntPtr selector);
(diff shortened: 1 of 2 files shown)


osu.Framework/Platform/MacOS/Native/Cocoa.cs
```diff
@@ -61,30 +69,45 @@ internal static class Cocoa
...
         public static extern bool CGCursorIsVisible();
 
         [DllImport(LIB_CORE_GRAPHICS, EntryPoint = "CGEventSourceFlagsState")]
         public static extern ulong CGEventSourceFlagsState(int stateID);
 
+        [DllImport(LIB_DL)]
+        private static extern IntPtr dlsym(IntPtr handle, string name);
+
+        [DllImport(LIB_DL)]
+        private static extern IntPtr dlopen(string fileName, int flags);
```
huoyaoyuan: Don't P/Invoke on this - use `NativeLibrary.Load` instead.

osu.Framework/Platform/MacOS/Native/Cocoa.cs
```diff
@@ -61,30 +69,45 @@ internal static class Cocoa
...
-        public static IntPtr FoundationLibrary;
 
         [DllImport(LIB_CORE_GRAPHICS, EntryPoint = "CGCursorIsVisible")]
         public static extern bool CGCursorIsVisible();
 
         [DllImport(LIB_CORE_GRAPHICS, EntryPoint = "CGEventSourceFlagsState")]
         public static extern ulong CGEventSourceFlagsState(int stateID);
 
+        [DllImport(LIB_DL)]
+        private static extern IntPtr dlsym(IntPtr handle, string name);
```
huoyaoyuan: Use `NativeLibrary.GetExport/TryGetExport`.

swoolcock:
Reason?

huoyaoyuan:
The reason is direct: if there's framework provided functionality, always use it instead of directly calling into the OS.
Although this is not portable code, use `NativeLibrary` on every platform still looks cleaner.

huoyaoyuan:
Maybe I've misunderstood something. `TryGetExport` is mainly for loading function exports, not sure about string constants. But it at least worth a try.

swoolcock:
I’m planning on replacing most of this native code with `NativeLibraryLoader` anyway.

https://github.com/mellinoe/nativelibraryloader

huoyaoyuan:
I didn't qualify it, but I meant `System.Runtime.InteropServices.NativeLibrary`. Please prefer it when it's suitable. .NET Core has worked a lot for cross-platform interop.

smoogipoo:
I'm not interested in increasing the complexity of implementation by using a new tool which may or may not have its own problems. This is just intended to take us further away from osuTK.

PR #3070, opened 2019-12-04
Fix OnEntering being called before drawable is loaded

peppy:
Can lead to unintended failure if doing initialisation in `LoadComplete` that is expected to be completed before `OnEntering` is hit.

### DIFF
--- a/osu.Framework/Screens/ScreenStack.cs
+++ b/osu.Framework/Screens/ScreenStack.cs
@@ -101,6 +101,12 @@ namespace osu.Framework.Screens
 
             var newScreenDrawable = newScreen.AsDrawable();
 
+            if (newScreenDrawable.IsLoaded)
+                throw new InvalidOperationException("A screen should not be loaded before being pushed.");
+
+            // this needs to be queued here before the load is begun so it preceed any potential OnSuspending event (also attached to OnLoadComplete).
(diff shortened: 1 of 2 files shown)


osu.Framework/Screens/ScreenStack.cs
```diff
@@ -101,6 +101,12 @@ internal void Push(IScreen source, IScreen newScreen)
 
             var newScreenDrawable = newScreen.AsDrawable();
 
+            if (newScreenDrawable.IsLoaded)
+                throw new InvalidOperationException("A screen should not be loaded before being pushed.");
```
smoogipoo: We definitely don't want this. Make this call OnEntering() instead?

peppy: If this is actually a possible workflow, it needs a test written. As discussed, it's likely not, since screens are disposed on exiting.

smoogipoo: If you're going to be defining this behaviour _now_, then the `Push()` xmldoc needs to state this.

Also there needs to be a test for this.

peppy: Previously it would silently fail. Now it more correctly throws. Documentation and test added.

PR #3072, opened 2019-12-04
Make TextBox abstract and add custom carets

jorolf:
# Breaking changes
## `TextBox` and all related components are now abstract

`BasicTextBox` is provided as a drop-in replacement that comes with the framework design language.

## `PasswordTextBox` has been renamed 

Renamed to `BasicPasswordTextBox` and comes with the framework design language.

`BasicTextBox` still works as expected

This also adds the ability to custom skin and animate the carets that a textbox uses. (see TestScene for example)


osu.Framework/Graphics/UserInterface/BasicTextBox.cs
```diff
@@ -1,20 +1,21 @@
...
 
 namespace osu.Framework.Graphics.UserInterface
 {
     public class BasicTextBox : TextBox
     {
-        protected override float CaretWidth => 2;
+        protected virtual float CaretWidth => 2;
 
-        protected override Color4 SelectionColour => FrameworkColour.YellowGreen;
+        protected virtual Color4 SelectionColour => FrameworkColour.YellowGreen;
```
bdach: I'm not sure about this design. If you override `CreateCaret` so that it doesn't return a `BasicDrawableCaret`, aren't these just going to be unused? They should probably be properties of the caret itself.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -338,16 +330,15 @@ private void updateCursorAndLayout()
 
             textUpdateScheduler.Update();
 
-            float caretWidth = CaretWidth;
-
-            Vector2 cursorPos = Vector2.Zero;
+            float cursorPos = 0;
             if (text.Length > 0)
-                cursorPos.X = getPositionAt(selectionLeft) - CaretWidth / 2;
+                cursorPos = getPositionAt(selectionLeft);
```
bdach: The adjustment for the width has gone - is there anything to replace it? I don't see that horizontal anchoring is guaranteed to be to centre for the caret either.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -998,26 +980,23 @@ private void onImeComposition(string s)
...
-                Colour = Color4.Transparent;
-                Anchor = Anchor.CentreLeft;
-                Origin = Anchor.CentreLeft;
+                RelativeSizeAxes = Axes.Y;
+            }
 
-                Masking = true;
-                CornerRadius = 1;
+            /// <summary>
+            /// Absolute selection width or null for default width
```
bdach: This xmldoc/API design seems strange/misleading. Where is the `null` for default width logic enforced? It's not in this abstract class, so are implementors expected to check for `null` themselves and implement the constraint?

IMO instead of exposing a half-well-defined `SelectionWidth` the abstract caret should be able to accept an adjustment for the caret width (through a protected abstract property, maybe) and then use it for determining the width of the selection. Gets rid of the nullable and also of having implementors uphold the contract themselves.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -561,8 +539,9 @@ protected void InsertString(string text)
                     continue;
                 }
 
+                const float fade_time = 120;
                 var col = (Color4)ch.Colour;
-                ch.FadeColour(col.Opacity(0)).FadeColour(col, caret_move_time * 2, Easing.Out);
+                ch.FadeColour(col.Opacity(0)).FadeColour(col, fade_time, Easing.Out);
```
bdach: If we want to make this textbox abstract I'd advocate for getting rid of all effects entirely and just leave the abstract one raw and brutalist if you will. This is now weird since the base textbox enforces text fade which was related to the caret and now they've been decoupled entirely, so this becomes entirely a magic number.

bdach:
Was `TextBox` usable by end-users before (as in, was it fully functional and could someone have used it in normal game code)? If it was, then this might end up being a breaking change for consumers even if we provide a replacement since it won't be drop-in due to naming.

osu.Framework/Graphics/UserInterface/BasicTextBox.cs
jorolf: They already are properties of the caret. I just didn't want to break backwards compatibility.

osu.Framework/Graphics/UserInterface/TextBox.cs
jorolf: I should probably change "null for default width" to "null if nothing is currently selected". And yes, I'd expect implementors to check for null and implement the differing behavior. (the compiler even enforces that as you can't just implicitly cast ``float?`` to ``float``)

jorolf:
Yes this is a breaking change (read the description)

osu.Framework/Graphics/UserInterface/TextBox.cs
bdach: Does it really make sense for us to allow users to mess with the width of the text selected? Even if we want to allow them to add some padding that should probably be done on the actual drawable. I don't see a use case that would warrant exposing this like that.

osu.Framework/Graphics/UserInterface/TextBox.cs
jorolf: We shouldn't enforce this offset imho. If it's needed for a specific caret then it should be implemented there. (Although I'm actually not quite sure how this worked or why it was needed exactly)

bdach:
I've reformatted the description to use the standard breaking change description format used in the releases and [on the wiki](https://github.com/ppy/osu-framework/wiki/Breaking-changes). I'd appreciate it if you had a look that I haven't messed something up in the process.

osu.Framework/Graphics/UserInterface/TextBox.cs
jorolf: I'm not quite sure if I understand what you want me to do. Should I add a ``DefaultCaretWidth`` that is being set instead of ``null`` to remove the nullability? Should I just set the ``Drawables`` width directly?

bdach: I just don't agree with consumers even seeing the selection width, it's not that declarative. The base caret should make sure that the width of the selection is such that all the text that needs to be is inside of the selection - now implementors have to check for `null` and in 99% of cases replicate the obvious logic.

The only thing it makes sense for me to expose is the width of the actual textbox caret, really.

osu.Framework/Graphics/UserInterface/BasicTextBox.cs
```diff
@@ -29,6 +30,66 @@ public BasicTextBox()
...
+                            .FadeTo(0.5f, 200, Easing.Out)
+                            .FadeColour(SelectionColour, 200, Easing.Out);
+                    }
+                    else
+                    {
+                        this.ResizeWidthTo(CaretWidth, caret_move_time, Easing.Out);
+                        this
+                            .FadeColour(Color4.White, 200, Easing.Out)
+                            .Loop(c => c.FadeTo(0.7f).FadeTo(0.4f, 500, Easing.InOutSine));
+                    }
```
bdach: Now that I think about it I'm not too hot on animation logic sitting inside of this setter, either. Case for a bindable maybe?

jorolf: A bindable is probably overkill. We don't need to bind it to anything or disabled it or do other bindable stuff with it.

osu.Framework/Graphics/UserInterface/TextBox.cs
jorolf: Sorry that I only reply now but I still have no clue how I should implement this? How should the ``TextBox`` change the ``SelectionWidth``?

osu.Framework/Graphics/UserInterface/BasicTextBox.cs
peppy: I think this is fine since it's only in the Basic implementation.

osu.Framework/Graphics/UserInterface/TextBox.cs
peppy: I've changed this to be a documented method for clarity.

peppy:
@smoogipoo requesting a semi-thorough review on this.

PR #3089, opened 2019-12-09
Combine BindableNumber implementations into base class

peppy:
Generally makes working with `BindableNumber<>` a lot better.

PR #3090, opened 2019-12-09
Fix SliderBar's Current implementation

### LINKED ISSUE #3088: SliderBar doesn't use BindableWithCurrent
Not correctly holding a reference.

peppy:
Closes #3088.

- [x] Depends on https://github.com/ppy/osu-framework/pull/3089


peppy:
Tested against osu!, can confirm fixes the related issue.

smoogipoo:
Shouldn't there be a framework-side test for this?

peppy:
The behaviour is tested in `AggregateBindableTest` already, so I thought it was pretty redundant to add one specifically for this.

smoogipoo:
I meant a test for SliderBar, but alright.

PR #3093, opened 2019-12-10
Make BindableNumber able to handle more types

### LINKED ISSUE #3092: Have BindableNumber<T> support more types
Up until now, `BindableNumber<T>` was abstract and only implementations of it could be used, of which there were 4 default choices provided by the framework - `BindableDouble`, `BindableFloat`, `BindableLong` and `BindableInt`.

As of https://github.com/ppy/osu-framework/pull/3089, `BindableNumber<T>` can now be instantiated, however it only supports the types `float`, `double`, `int` and `long`. All other types will throw `InvalidOperationException`s: https://github.com/ppy/osu-framework/blob/61cbe268a6656455bbe6ff69515721f7121c5732/osu.Framework/Bindables/BindableNumber.cs#L119-L129

All methods in this class should be updated to support the types defined by the ctor: https://github.com/ppy/osu-framework/blob/61cbe268a6656455bbe6ff69515721f7121c5732/osu.Framework/Bindables/BindableNumber.cs#L24-L33

The exceptions in such methods can then also be removed.

smoogipoo:
Resolves #3092

### DIFF
--- a/osu.Framework/Bindables/BindableNumber.cs
+++ b/osu.Framework/Bindables/BindableNumber.cs
@@ -116,17 +116,28 @@ namespace osu.Framework.Bindables
         {
             get
             {
-                if (typeof(T) == typeof(double))
-                    return (T)(object)double.MinValue;
-                if (typeof(T) == typeof(float))
-                    return (T)(object)float.MinValue;
-                if (typeof(T) == typeof(int))
+                if (typeof(T) == typeof(sbyte))
+                    return (T)(object)sbyte.MinValue;
+                else if (typeof(T) == typeof(byte))
+                    return (T)(object)byte.MinValue;
+                else if (typeof(T) == typeof(short))
+                    return (T)(object)short.MinValue;
+                else if (typeof(T) == typeof(ushort))
+                    return (T)(object)ushort.MinValue;
+                else if (typeof(T) == typeof(int))
                     return (T)(object)int.MinValue;
-                if (typeof(T) == typeof(long))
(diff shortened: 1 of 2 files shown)


osu.Framework/Bindables/BindableNumber.cs
```diff
@@ -116,17 +116,28 @@ protected virtual T DefaultMinValue
...
+                if (typeof(T) == typeof(ulong))
+                    return (T)(object)ulong.MinValue;
+                if (typeof(T) == typeof(float))
+                    return (T)(object)float.MinValue;
+                if (typeof(T) == typeof(double))
+                    return (T)(object)double.MinValue;
 
-                throw new NotSupportedException(
-                    $"{nameof(BindableNumber<T>)} needs to override {nameof(DefaultMinValue)} to provide a sane default.");
+                throw new NotSupportedException("How do you get here?");
```
peppy: Probably should be a `Trace.Assert` (already checked in `ctor`). A good time opportunity to remove this silly exception message at very least.

osu.Framework/Bindables/BindableNumber.cs
```diff
@@ -377,5 +401,18 @@ private static T min(T value1, T value2)
 
         private static T clamp(T value, T minValue, T maxValue)
             => max(minValue, min(maxValue, value));
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
smoogipoo: Have tested that the JIT will not inline this by default.

PR #3104, opened 2019-12-12
Fix finalizer potentially polluting disposal test results

peppy:


smoogipoo:
Probably want to do the same in `TestUnbindOnClearInternal`, since it should always be testing unbind on `ClearInternal` not unbind on dispose.

PR #3115, opened 2019-12-13
Fix joystick-related binding issues

galister:
* Joystick AxisNegative & AxisPositive buttons to respect deadzones
* Prevent deadzones from being less than FLOAT_EPSILON as this hurts the math
* Add support for up to 128 joystick buttons

Currently, if you try to bind your keys (even keyboard / mouse) while a joystick is plugged in, Osu will record the ever-present wiggle of the axes due to there not being a deadzone for the virtual joy buttons. Another issue is that when a deadzone is not set, a value of 0 is used, which will cause issues since we're working with floats.

The button limit increase for https://github.com/ppy/osuTK/pull/59 also made it in here.


peppy:
For future contributions, please make sure to make a branch called something related to the change (ie. don't make pull requests from your "master" branch).

PR #3116, opened 2019-12-14
Fix runaway scenario in video decoding loop

peppy:
If a video was not present, scheduled events were added even if the state hadn't changed due to missing early return (and non-matching delegate in `AddOnce`).


osu.Framework/Graphics/Video/VideoDecoder.cs
```diff
@@ -62,20 +62,23 @@ public unsafe class VideoDecoder : IDisposable
...
-                if (bindableState.Value != value)
-                    scheduler?.AddOnce(() => bindableState.Value = value);
+                if (volatileState == value) return;
 
+                scheduler?.Add(() => bindableState.Value = value);
                 volatileState = value;
             }
         }
 
+        private void updateState() => bindableState.Value = volatileState;
```
smoogipoo: can be removed (or used)

PR #3013, opened 2019-11-20
Initial SDL2 window implementation

swoolcock:
This PR adds a WIP SDL2 window implementation.  It is primarily for incremental testing and development, so that the final implementation does not become a huge PR.

Only minor changes are currently required in osu! as all osuTK OpenGL bindings are being forcefully loaded with `SDL_GL_GetProcAddress`.

Audio and video are unaffected by this change as they are unrelated to osuTK and SDL2.

SDL2 can be enabled in visual tests by passing the `--sdl` command line argument.  If SDL2 is not enabled via command line, this PR should continue to use existing codepaths and have no effect on existing framework usage.

Successfully tested with osu! on macOS and Windows.  This PR is **not** a breaking change for framework consumers.  Initially for this PR I would like to ensure that there are no problems sans-SDL2 on all platforms, and we can check SDL2 on all platforms as a matter of course.

This PR uses the SDL2 bindings provided by Veldrid, as it is likely we will be moving to that eventually for the graphics portion.

Included in this PR:

* Window creation and display
* OpenGL passthrough for graphics
* Key input
* Text input
* Mouse input

To be implemented in future PRs:

* Raw mouse input
* Joystick input
* Resolution switching
* Frame limiting

Related to #2448 
Dependency of https://github.com/ppy/osu/pull/6991

PR #3117, opened 2019-12-15
Set framesync based on the refresh rate of the current display

swoolcock:
Previously the frame sync rate was always based on the refresh rate of the primary display.

The only caveat is that moving the window between displays will not automatically update the frame sync to the correct refresh rate until a change is triggered.

Needs testing on a system with multiple refresh rates.  I don't have access to one.

Note that this inadvertently fixes the broken macOS menu focus on the SDL branch as osuTK's `DisplayDevice` is no longer being initialised.

Related to #3013

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -721,7 +721,7 @@ namespace osu.Framework.Platform
                 if (Window == null)
                     return;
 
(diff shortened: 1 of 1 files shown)


peppy:
Works as expected with multiple displays (doesn't update when crossing displays but is correct refresh rate for the display the window opened on).

PR #3119, opened 2019-12-16
Optimize some struct performance by using readonly methods and in modifier

huoyaoyuan:
Just updated the 2 large structs that be compared *directly* at *evert frame*.
The vertex types are compared via generic constraint. Adding a new interface that constrains `Equals(in T)` increases the complexity.

I'm looking forward to the record feature that can make most structs immutable.

Breaking changes:

# vNext

`Quad` and `Triangle` have been made readonly

Code that previously mutated the structs should be converted to create a new instance every time.


huoyaoyuan:
The way treating colour spaces is a bit weird. I'm reverting change on it.
Tested with osu to get the correct color drawn.

smoogipoo:
Seems good to me. @peppy also requesting your second review on this.

peppy:
My only issues is there's no inspection to let us know when we can use this for future additions/changes.

PR #3128, opened 2019-12-19
Fix framesync regression on Android and iOS

swoolcock:
Resolves https://github.com/ppy/osu/issues/7282
Regression from #3117 

The `CurrentDisplay` property in osuTK's Android and iOS implementations throws a `NotImplementedException`.  This change ensures we return null instead, which will then default to 60hz as per #3117 

The better solution would be to correctly implement `CurrentDisplay` and `DisplayDevice` in osuTK for iOS and Android, but this is non-trivial and an extreme edge case.  This problem can be easily solved with the SDL2 migration.  I would also like to avoid any and all large changes to osuTK where possible.

Edit: Would appreciate if someone could please test this on Android for me, as I do not have an Android device to test with.

### DIFF
--- a/osu.Framework.Android/AndroidGameWindow.cs
+++ b/osu.Framework.Android/AndroidGameWindow.cs
@@ -2,9 +2,10 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using osu.Framework.Configuration;
-using osu.Framework.Platform;
 using osuTK.Graphics;
 using System.Collections.Generic;
+using osuTK;
+using GameWindow = osu.Framework.Platform.GameWindow;
 
 namespace osu.Framework.Android
 {
@@ -48,6 +49,12 @@ namespace osu.Framework.Android
         public override void Run(double updateRate)
         {
             View.Run(updateRate);
+        }
+
+        public override DisplayDevice CurrentDisplay
+        {
+            get => null;
+            set { }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


bdach:
I'll add testing this on Android to my shortlist of things to do in the near future. Should report back by the end of the week.

peppy:
What will this do on non-60hz devices (especially iOS, which is also changed in this PR)?

swoolcock:
Nice catch, have changed to make sure they work the same way as before the PR.

osu.Framework.iOS/IOSGameWindow.cs
```diff
@@ -36,6 +37,12 @@ public override osuTK.WindowState WindowState
             set { }
         }
 
+        public override DisplayDevice CurrentDisplay
```
peppy: i'm so confused as to why this is necessary

osu.Framework.Android/AndroidGameWindow.cs
```diff
@@ -49,5 +50,11 @@ public override void Run(double updateRate)
         {
             View.Run(updateRate);
         }
+
+        public override DisplayDevice CurrentDisplay
+        {
+            get => DisplayDevice.Default;
+            set { }
```
peppy: throw `InvalidOperationException` instead

osu.Framework.iOS/IOSGameWindow.cs
swoolcock: Calling `GameWindow.CurrentDisplay` by default delegates to the osuTK implementation, which throws an exception on iOS and Android.  Since the previous PR replaced `DisplayDevice.Default` with a call to `GameWindow.CurrentDisplay`, I have changed these implementations in `IOSGameWindow` and `AndroidGameWindow` to ensure it doesn't call the osuTK versions and thus regress.

PR #3132, opened 2019-12-20
.NET Core 3.1 at framework side

huoyaoyuan:
ppy/osu#7208 and ppy/osu#7304 .
Not updating the libraries because 3.1 doesn't provide any related API.


huoyaoyuan:
I forgot readme here.

PR #3133, opened 2019-12-21
Change TabControl.IsSwitchable to property

Joehuu:
### `TabControl.IsSwitchable` is changed to auto-implemented property with default `false`

As there can be many similar tab controls on any one screen, the one with the highest depth will absorb all input. This change makes it more manageable by manually setting it where it's needed, and avoids nested classes.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs
@@ -69,7 +69,7 @@ namespace osu.Framework.Graphics.UserInterface
         /// <summary>
         /// When true, tabs can be switched back and forth using PlatformAction.DocumentPrevious and PlatformAction.DocumentNext respectively.
         /// </summary>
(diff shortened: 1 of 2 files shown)


smoogipoo:
I'm fine with this. @peppy

osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -69,7 +69,7 @@ public IEnumerable<T> Items
         /// <summary>
         /// When true, tabs can be switched back and forth using PlatformAction.DocumentPrevious and PlatformAction.DocumentNext respectively.
         /// </summary>
-        public virtual bool IsSwitchable => true;
```
peppy: please revert the naming to be `IsSwitchable`. it's not guaranteed to be the 'tab' key and may be confused with tabbing between multiple UI controls.

PR #3141, opened 2019-12-26
Fix GetCommonPrefix extension method failing on empty enumerable source

peppy:
Closes https://github.com/ppy/osu/issues/7349.


huoyaoyuan:
I did a span-based implementation sevaral days ago, and should be able to handle empty sequence:
```csharp
        public static string GetCommonPrefix(this IEnumerable<string> collection)
        {
            ReadOnlySpan<char> prefix = default;
            foreach (var str in collection)
            {
                if (prefix.IsEmpty) // the first string
                {
                    prefix = str;
                    continue;
                }

                while (!prefix.IsEmpty)
                {
                    if (str.AsSpan().StartsWith(prefix))
                        break;
                    else
                        prefix = prefix[0..^1];
                }

                if (prefix.IsEmpty)
                    return string.Empty;
            }

            return new string(prefix);
        }
```

PR #3143, opened 2019-12-27
Add Drawable.Empty static helper for empty construction

smoogipoo:
For patterns such as:

```
protected virtual Drawable CreateXYZ() => new Container();
```

which can now be written as:

```
protected virtual Drawable CreateXYZ() => Drawable.Empty;
```

Where: returning `null` feels wrong in hierarchical construction, and returning a `Container` feels arbitrary.

Note: Empty Drawable instances aren't shared.

### DIFF
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -2294,6 +2294,15 @@ namespace osu.Framework.Graphics
             else
                 return shortClass;
         }
+
+        /// <summary>
+        /// Represents a new instance of an empty <see cref="Drawable"/>.
+        /// </summary>
+        public static Drawable Empty => new EmptyDrawable();
+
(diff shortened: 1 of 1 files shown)


swoolcock:
I disagree with this being a property if it’s not a shared value.  I would expect consecutively reading from a property would give the same value/object every time.  I think `Drawable.Empty()` is clearer.

smoogipoo:
Yeah I'm not sure... If we wanted to follow o!f conventions, neither Empty nor Empty() are appropriate, and it should be named `CreateEmpty()`.

swoolcock:
`CreateXXX` have historically been `protected virtual` though, not `public static`.

smoogipoo:
The original intention of `CreateXXX()` was just to let the user know an allocation was occurring, but maybe that should be the initial assumption for methods in the first place. Will name it `Empty()` for now.

PR #1962, opened 2018-10-21
Make user agent configurable

UselessToucan:
At the moment UserAgent is hardcoded and there is no way to override it.
`client.DefaultRequestHeaders.UserAgent.ParseAdd("osu!");`
https://github.com/ppy/osu-framework/blob/master/osu.Framework/IO/Network/WebRequest.cs#L137

It is not correct since osu! is not the only game based on the framework.
This PR allows you to set UserAgent using framework config.

Known framework users besides osu!: @default0, Lovewing(@sr229, @noud02)

PR #3145, opened 2019-12-28
Set UserAgent on WebRequest level

UselessToucan:
alternative to https://github.com/ppy/osu-framework/pull/1962

This PR moves UserAgent assigning from `DefaultRequestHeaders` to `internalPerform` and makes it customizable via `WebRequest.UserAgent` virtual property.


peppy:
I think this looks fine. Can you please open a matching PR on osu!-side?

Issue #3148, opened 2019-12-31
Splitting the NuGet Package


default0:
The main `osu.Framework` NuGet Package includes the very large native ffmpeg libraries as well as the not-very-small `Microsoft.CodeAnalysis` (~7 MB for Microsoft.CodeAnalysis.dll and Microsoft.CodeAnalysis.CSharp.dll).

Digging through the code, it seems that ffmpeg is only used inside the `osu.Framework.Graphics.Video` namespace and the `Microsoft.CodeAnalysis` dependency seems to only be used inside the `osu.Framework.Testing` namespace.

To me both of those are obvious candidates for separating them into standalone NuGet packages and removing them from the main `osu.Framework` package.
This would allow users to depend only on the parts of the framework (and their transitive, relatively heavy dependencies) that are necessary for the specific project.

In addition to this, if the NuGet package does become split, there is an option to create a shell/main package (fe `osu.Framework.Full`) that pulls everything for convenience incase every feature of the framework should be available/utilized.

Is this a path forward that this framework should go towards? If so, what should be the organization of the effort - especially with regards to the platform-specific projects (`osu.Framework.Android` etc)?

peppy:
Sounds fair. The goal was to (at very least) move the ffmpeg libs out for desktop, simlar to the mobile counterparts.

Not sure about codeanalysis. I believe that is only used by tests, so maybe we can split that out in some fashion such that it is not consumed by actual framework applications.

huoyaoyuan:
Note: including testing suites in main package has another side effect that makes Visual Studio considers osu.Framework and osu.Game to be unit test projects.

CodeAnalysis adds a large amouont/size of shipped assemblies to the end user, and causes `System.Memory` problem of Xamarin (I'm not sure if it's the only culprit). Vote for splitting it too.

PR #3149, opened 2020-01-01
Splitting Video Decoding into its own NuGet Package

default0:
This is to start with the effort described in #3148.

This PR is in its current state incomplete and needs discussion around a few design issues/decisions that were necessitated by the change in project structure - I also need a bit of insight/feedback into the current deployment/packaging process, especially around native libraries, as I am unfamiliar with it.

The following issues were encountered by moving the Video decoding into its own `osu.Framework.Video` project:

`osu.Framework.Video` is *not* the name of the namespace for video decoding as of today (it is `osu.Framework.Graphics.Video`) - however `osu.Framework.Graphics.Video` does not feel like an appropriate project name. For now I have kept the namespace of the classes declared by the project the same to minimize hassle when migrating.
Given that framework consumers who need video functionality will be broken anyways (because they will need to reference a new package) it might be better to change the namespace to `osu.Framework.Video` right away though.

The video decoding process depended on `TextureGL.FlushUploads()` which is an internal method. The options to deal with this are to either rewrite the logic to not rely on calling `FlushUploads()`, using an `InternalsVisibleTo`-Attribute to expose it to `osu.Framework.Video` or simply make the method public.
I chose to make the method public for now.

The visual tests displayed debug information about the decoding process exposed by `VideoSprite`, specifically the `CurrentFrameTime` and `AvailableFrames` properties, both of which were marked internal. I chose to make those public as well and added documentation to them.

The `GameHost` base class exposed a method `CreateVideoDecoder` which hard-depended on the `VideoDecoder` class being present. Moving this to its own project meant that a factory of this type could no longer exist. I chose to replace this method with a generic `Create<T>` method and enable subclasses of `GameHost` to call a protected helper method called `AddPlatformFactory` to enable the `AndroidGameHost` and `IOSGameHost` classes to override the default implementation (which simply forwards to `Activator.CreateInstance`). Another option to circumvent this would be to keep the special case added for video decoding by adding an `IVideoDecoder` interface in the `osu.Framework` and implementing/depending on that inside the `osu.Framework.Video` package, though having video-decoding/playback-specific code inside the `osu.Framework` even after this change rubs me the wrong way.

As I mentioned at the start, I do not understand the deployment/packaging process currently. My assumption right now is that `osu.Framework.NativeLibs` is what actually causes the large dlls for video decoding etc. to be pulled and as of the creation of this PR that project remains untouched. I would think that the next step would be to either create a separate `osu.Framework.NativeLibs.Video` project that includes everything required for video decoding (and referencing that inside `osu.Framework.Video` project the way `osu.Framework` currently references `osu.Framework.NativeLibs`) would be the way to go. Please advise on this point.


osu.Framework/Platform/GameHost.cs
```diff
@@ -921,13 +922,26 @@ public virtual IResourceStore<TextureUpload> CreateTextureLoaderStore(IResourceS
...
-        public virtual VideoDecoder CreateVideoDecoder(Stream stream, Scheduler scheduler) => new VideoDecoder(stream, scheduler);
+        /// <param name="args">The arguments passed to the constructor of the object.</param>
+        /// <returns>An instance of the specified type or its platform-specific subtype.</returns>
+        public T Create<T>(params object[] args)
+        {
+            if (platformFactories.TryGetValue(typeof(T), out var factory))
+                return (T)factory(args);
+            else
+                return (T)Activator.CreateInstance(typeof(T), args);
+        }
```
bdach: This API surface is far far *far* too generic for my liking. This can accept and create anything and everything. Args are entirely too untyped as well. Maybe extension methods can help with this?
```csharp
public static class GameHostVideoExtensions
{
    public static VideoDecoder CreateVideoDecoder(this GameHost host, Stream stream, Scheduler scheduler)
    {
        switch (host)
        {
            case IOSGameHost _:
                return new IOSVideoDecoder(stream, scheduler);
            case AndroidGameHost _:
                return new AndroidVideoDecoder(stream, scheduler);
            default:
                return new VideoDecoder(stream, scheduler);
        }
    }
}
```

I guess doing it the above way would harm extensibility, but if extensibility is deemed important we could also create a way for `GameHost` implementations to register themselves in `GameHostVideoExtensions` as requiring a different implementation.

osu.Framework.Video/VideoSprite.cs
```diff
@@ -77,9 +77,15 @@ public double PlaybackPosition
...
+        /// <summary>
+        /// The time in milliseconds that the currently displayed frame in the video represents.
+        /// </summary>
+        public double CurrentFrameTime => lastFrame?.Time ?? 0;
 
-        internal int AvailableFrames => availableFrames.Count;
+        /// <summary>
+        /// The number of decoded frames available in the buffer.
+        /// </summary>
+        public int AvailableFrames => availableFrames.Count;
```
bdach: Kind of going back and forth on this. `CurrentFrameTime` being public might be *sort* of ok but `AvailableFrames` shouldn't be since it's an implementation detail. Would rather it be protected and subclass `VideoSprite` in the test scene to get back the debug info.

osu.Framework/Graphics/OpenGL/Textures/TextureGL.cs
```diff
@@ -99,7 +99,7 @@ public void Dispose()
         /// <summary>
         /// Flush any unprocessed uploads without actually uploading.
         /// </summary>
-        internal abstract void FlushUploads();
+        public abstract void FlushUploads();
```
bdach: I think this is ok being public. It's not pretty but I don't see a good way to avoid it either.

peppy:
Before completing this, we may want to consider the optimisation PR (https://github.com/ppy/osu-framework/pull/3019) since that touches the code quite a lot.

osu.Framework/Platform/GameHost.cs
default0: This cannot be done.
If GameHostVideoExtensions lives in osu.Framework.Video, osu.Framework.Video would have to depend on osu.Framework.Android / osu.Framework.iOS which should not be a thing at all, ever.
The only way you can get something like this with an extension method is if you create separate extension methods in osu.Framework.Android and osu.Framework.iOS - but this also severely limits discoverability of the API since you have to have a `using` of the appropriate namespace around to see the API even exists.

I understand the reservations about it being untyped and I agree that it's an issue, but you cannot provide a typed API for types and signatures that are unavailable at compile-time.

default0:
Sounds fair, I consider this to be "on hold" until the path forward is clear anyways :-)

osu.Framework/Platform/GameHost.cs
bdach: Right, you're correct with the dependency chain. In that case I'd maybe consider hoisting out platform-specific video decoder implementations to `osu.Framework.Video` too. I don't think that'd be too much of a code smell and it shouldn't be a lot of work either eyeballing it (they're just P/Invokes).

For me the existence of this `Create<T>` API alone is big enough that I'm questioning if this change makes sense if it cannot go away. Kind of feels like trying to decouple things that need to be coupled anyways. I'd rather have type safety in a monolith than a weird runtime-checked API with two packages.

osu-framework.sln
```diff
@@ -37,7 +37,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		osu.Framework.iOS.props = osu.Framework.iOS.props
 	EndProjectSection
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "osu.Framework.Benchmarks", "osu.Framework.Benchmarks\osu.Framework.Benchmarks.csproj", "{F294C804-8AE2-4020-841A-AF0D97FBE80C}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "osu.Framework.Benchmarks", "osu.Framework.Benchmarks\osu.Framework.Benchmarks.csproj", "{F294C804-8AE2-4020-841A-AF0D97FBE80C}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "osu.Framework.Video", "osu.Framework.Video\osu.Framework.Video.csproj", "{55613B00-6E8D-4DDD-8367-2D64672E24C8}"
```
huoyaoyuan: Not sure which version of IDE you are using, but latest Visual Studio stores all .NET Core SDK-style projects with the `9A` type GUID.

osu.Framework.Video/osu.Framework.Video.csproj
```diff
@@ -0,0 +1,34 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Library</OutputType>
+    <TargetFrameworks>netcoreapp3.0;netstandard2.1</TargetFrameworks>
+    <ApplicationIcon />
+    <StartupObject />
+    <Win32Resource />
+    <Description>Adds Video decoding and playback support for the osu!framework</Description>
+    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
```
huoyaoyuan: Redundant default properties (all of these except description)

osu.Framework.Video/osu.Framework.Video.csproj
```diff
@@ -0,0 +1,34 @@
...
+    <DocumentationFile>obj\Release\netcoreapp3.0\osu.Framework.Video.xml</DocumentationFile>
+    <DefineConstants>TRACE;JETBRAINS_ANNOTATIONS</DefineConstants>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Optimize>false</Optimize>
+    <DocumentationFile>obj\Debug\netcoreapp3.0\osu.Framework.Video.xml</DocumentationFile>
+  </PropertyGroup>
```
huoyaoyuan: Redundant conditional groups. See the framework project for combining them.

huoyaoyuan:
You are also missing other project metadata, like package id (should be `ppy.osu.Framework.Video`.
I suggest to edit project file by hand, following the format of main framework.

smoogipoo:
Closing this for now. It makes some sense but isn't immediately on our radar and has significant merge conflicts.

PR #3150, opened 2020-01-01
Fix possible null dereference

UselessToucan:


osu.Framework.Benchmarks/BenchmarkFontLoading.cs
```diff
@@ -85,7 +85,7 @@ private void runFor(GlyphStore store)
                 {
                     var icon = (IconUsage)p.GetValue(null);
                     using (var upload = store.Get(icon.Icon.ToString()))
-                        Trace.Assert(upload.Data != null);
+                        Trace.Assert(upload != null && upload.Data != null);
```
bdach: Might as well null propagate here?

bdach:
Also note that adding nullchecks will trigger some null-reference tracking analyzer inspections that maybe don't really make sense (see CI).

osu.Framework/Platform/GameHost.cs
```diff
@@ -798,7 +798,7 @@ protected virtual void SetupConfig(IDictionary<FrameworkSetting, object> gameDef
             bypassFrontToBackPass = DebugConfig.GetBindable<bool>(DebugSetting.BypassFrontToBackPass);
 
             threadLocale = Config.GetBindable<string>(FrameworkSetting.Locale);
-            threadLocale.BindValueChanged(locale =>
+            threadLocale?.BindValueChanged(locale =>
```
smoogipoo: Config bindables should never be null. This should probably return an exception instead on the `GetBindable()` call.

osu.Framework/IO/Stores/RawCachingGlyphStore.cs
```diff
@@ -48,7 +48,7 @@ private PageInfo createCachedPageInfo(int page)
 
             using (var stream = Store.GetStream(filename))
             {
-                string streamMd5 = stream.ComputeMD5Hash();
+                string streamMd5 = stream?.ComputeMD5Hash();
```
smoogipoo: Should either fail completely or return null characters to match the general implementation of resource stores.

osu.Framework/Graphics/Textures/TextureAtlas.cs
```diff
@@ -64,7 +64,7 @@ public void Reset()
             AtlasTexture = new TextureGLAtlas(atlasWidth, atlasHeight, manualMipmaps, filteringMode);
 
             using (var whiteTex = Add(WHITE_PIXEL_SIZE, WHITE_PIXEL_SIZE))
-                whiteTex.SetData(new TextureUpload(new Image<Rgba32>(SixLabors.ImageSharp.Configuration.Default, whiteTex.Width, whiteTex.Height, Rgba32.White)));
+                whiteTex?.SetData(new TextureUpload(new Image<Rgba32>(SixLabors.ImageSharp.Configuration.Default, whiteTex.Width, whiteTex.Height, Rgba32.White)));
```
smoogipoo: Should fail completely, in the ctor.

osu.Framework/Graphics/Shaders/ShaderPart.cs
```diff
@@ -114,7 +114,7 @@ private string loadFile(byte[] bytes, bool mainFile)
 
                         string backbufferCode = loadFile(manager.LoadRaw("sh_Backbuffer_Internal.h"), false);
 
-                        backbufferCode = backbufferCode.Replace("{{ real_main }}", realMainName);
+                        backbufferCode = backbufferCode?.Replace("{{ real_main }}", realMainName);
```
smoogipoo: Should be asserted.

smoogipoo:
Let's update these classes/methods to support NRT and revisit later.

PR #3155, opened 2020-01-05
Update hover before other mouse event propagates

### LINKED ISSUE #3153: Hovering should be updated before any mouse event propagations
Hovering should be updated before any mouse event propagations to avoid incorrect `IsHovered` values inside event as can be seen in the test.

```csharp
    public class TestSceneHoverUpdates : ManualInputManagerTestScene
    {
        private TestDrawable hoveredDrawable;

        [SetUp]
        public override void SetUp() => Schedule(() =>
        {
            base.SetUp();

            Child = hoveredDrawable = new TestDrawable();
        });

        [Test]
        public void TestCheckHoverFromMouseMove()
        {
            AddStep("move mouse inside drawable", () => InputManager.MoveMouseTo(hoveredDrawable));
            AddAssert("is hover correct", () => !hoveredDrawable.IncorrectHover);
        }

        private class TestDrawable : Box
        {
            public bool IncorrectHover = false;

            public TestDrawable()
            {
                Anchor = Anchor.Centre;
                Origin = Anchor.Centre;
                Size = new Vector2(128);
            }

            protected override bool OnMouseMove(MouseMoveEvent e)
            {
                if (!IncorrectHover && IsHovered != ReceivePositionalInputAt(e.ScreenSpaceMousePosition))
                    IncorrectHover = true;

                return base.OnMouseMove(e);
            }
        }
    }
```

mcendu:
Fixes #3153

### DIFF
--- a/osu.Framework.Tests/Visual/Input/TestSceneHoverUpdates.cs
+++ b/osu.Framework.Tests/Visual/Input/TestSceneHoverUpdates.cs
@@ -1,0 +1,51 @@
+﻿using NUnit.Framework;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Shapes;
+using osu.Framework.Input.Events;
+using osu.Framework.Testing;
+using osuTK;
+
+namespace osu.Framework.Tests.Visual.Input
+{
+    public class TestSceneHoverUpdates : ManualInputManagerTestScene
+    {
+        private TestDrawable hoveredDrawable;
+
+        [SetUp]
+        public override void SetUp() => Schedule(() =>
+        {
+            base.SetUp();
+
+            Child = hoveredDrawable = new TestDrawable();
+
+            InputManager.MoveMouseTo(Vector2.Zero);
+        });
+
+        [Test]
+        public void TestCheckHoverFromMouseMove()
+        {
+            AddStep("move mouse inside drawable", () => InputManager.MoveMouseTo(hoveredDrawable));
+            AddAssert("is hover correct", () => !hoveredDrawable.IncorrectHover);
+        }
+
+        private class TestDrawable : Box
+        {
+            public bool IncorrectHover = false;
+
+            public TestDrawable()
+            {
+                Anchor = Anchor.Centre;
+                Origin = Anchor.Centre;
+                Size = new Vector2(128);
+            }
+
+            protected override bool OnMouseMove(MouseMoveEvent e)
+            {
+                if (!IncorrectHover && IsHovered != ReceivePositionalInputAt(e.ScreenSpaceMousePosition))
+                    IncorrectHover = true;
+
+                return base.OnMouseMove(e);
+            }
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


osu.Framework.Tests/Visual/Input/TestSceneHoverUpdates.cs
```diff
@@ -0,0 +1,51 @@
...
+        private TestDrawable hoveredDrawable;
+
+        [SetUp]
+        public override void SetUp() => Schedule(() =>
+        {
+            base.SetUp();
+
+            Child = hoveredDrawable = new TestDrawable();
+
+            InputManager.MoveMouseTo(Vector2.Zero);
```
iiSaLMaN: you can just override `InitialMousePosition` to `Vector2.Zero`

smoogipoo:
This is changing the ordering of events. You've changed:

```
1 -> OnMouseMove
2 -> OnHover
```

into

```
1 -> OnHover
2 -> OnMouseMove
```

This can't happen. As for the underlying issue, I'd need more time to convince myself that `IsHovered` should be set before any movement event because I'm not sure that's correct.
Is there a scenario which you are running into that's dependent on this being resolved? @iiSaLMaN

iiSaLMaN:
> Is there a scenario which you are running into that's dependent on this being resolved? @iiSaLMaN

yes, the osu! ruleset seems to be entirely broken with my touch input implementation due to tapping in a circle while the hover is not updated in here: https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableHitCircle.cs#L196, `OnPressed(action)` called from a `MouseDownEvent`.

smoogipoo:
That sounds like you should separate your `IInput` into a position and a button change. Mouse handlers already do the same thing.

iiSaLMaN:
I already have `TouchButtonInput` and `TouchPositionInput`, but there is a case where `TouchButtonInput` can be applied with a pointer at position X:
`new TouchButtonInput(new TouchPointer(source, pos), true)`

iiSaLMaN:
@smoogipoo If you meant separating the input in a way that a `TouchPositionInput` updates the hover then `TouchButtonInput` presses the object, then i'm not sure how that'll work unless you meant to make `TouchPositionInput` act completely as a `MousePositionAbsoluteInput`, then that would be too complicated to make work.

I suggest that all the events that are to be propagated gets saved up in a list, and after the manager finishes from applying handler's pending inputs, the events will be propagated starting from the most important (mouse event, key event, or joystick event) to the least important. (drag event, hover event and focus event), sounds fair?

iiSaLMaN:
After trying to implement events sorting by most important to least important, I've found out that `MouseButtonEventManager` is too complicated to implement in, will see what i can do with your suggestion.

iiSaLMaN:
@mcendu Are you planning on resolving conversations and fixing the event order?

Issue #3154, opened 2020-01-05
Reconsider LocalisedString


huoyaoyuan:
There are 2 concepts of "displaying language" in osu:
- Game metadata. The call site provides 2 variations of text, and the text sprite selects one to display.
- UI text. The call site provides a lookup key and some optional formattable elements. The text sprite displays in the user's language.

Current implementation is written with case 1 in mind. For case 2, I'm strongly against using English text as lookup key due to polysemous words and phases. This means that the 2 cases can have nearly nothing to share at implementation site:
- Case 1, requires Romanized text, Unicode text, and the PreferUnicodeMetadata option.
- Case 2, requires lookup key, format arguments, and the UserLanguage option.

Tagging @smoogipoo .

### My supposed implementation

Rename `LocalisedString` to something like `StringDescriptor` or `LocalisableString` (this struct exists **before** actual localsation), which is accepted by `SpriteText`. It can be constructed by several cases: plain text, Romanized/Unicode, key/arguments.

`SpriteText` doesn't care about it, and passes it directly to localisation manager to ask for a `Bindable<string>`.

Localisation manager decides what to return based on it's type, like current implementation. I suggest to use different classes for different cases, to reduce user option binding overhead.

huoyaoyuan:
Clearifications/highlights for what I suggest to change:
Class rename. It really confuses me.
Don't mix lookup key and Romanized metadata. The lookup key can be provided together with an English fallback.
Replace `LocalisedBindableString` with 3 classes.
Remove the settable interface `ILocalisedBindableString` because the returned `Bindable<string>` may not handle all cases.

smoogipoo:
Seems like a fair direction. I also disliked the name `LocalisedString`. Crowd-in will give us key-value mappings to fulfil case 2.
Keep in mind the 3rd case, which is plain text. Localisation should be transparent to a user of the framework, unless they specifically request it via constructing a `LocalisableString`.

UselessToucan:
By the way, why osu! does not use [GNU/Gettext](https://github.com/VitaliiTsilnyk/NGettext)? It looks quite easy to [use](https://github.com/VitaliiTsilnyk/NGettext#installation-and-usage).

bdach:
As someone who's used gettext before when working with django I advise against using it for our purposes, for the following reasons:

* It uses two formats, one (`.po`) human-editable and one (`.mo`) not. The latter one is actually used by consumer applications, therefore introducing a new tooling step.
[...]

PR #3157, opened 2020-01-06
Localisation overhaul

huoyaoyuan:
Changes proposed in #3154 . Requires game side update.

Highlights:
- The name `LocalisableStringDescriptor` may be too long. Remove the term "descriptor"?
- I'll write new code with NRT enabled.
- For different constructing scenarios, use `FromXXX` instead of `new` to point out.
- Keep the implicit conversion **from** string for not localisable text. Remove the implicit conversion **to** string to reduce misusages (For example, `==` was bound to `(string)left == (string)right` previously). `ToString()` is expected to give a good result.
- I don't expect a great amount of not localisable text in the game that's ready for ship, thus not optimizing for them.
- Interpolated string that's converted into `System.FormattableString` does have more overhead, but the language team is planning to improve it. Tolerate it since text formatting shouldn't be so hot in game.

# Breaking Changes

## LocalisedString no longer exists

For strings which have romanisable content, `RomanisableString` should be used instead. Note that the order of arguments has been reversed.

## Multiple UI Components now use `LocalisableString` in place of `string`

If you have custom implementations, you will need to update the overridden signatures.

## Explicit `ToString` may be required where previous not

Getting the current value of some strings (ie. on UI Components) will now require an explicit call to `.ToString()`.


osu.Framework/Localisation/LocalisableStringDescriptor.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// <summary>
+    /// A descriptor representing text that can be localised and formatted.
+    /// </summary>
+    public readonly struct LocalisableStringDescriptor : IEquatable<LocalisableStringDescriptor>
+    {
+        // Reuse string1 and string2 in different types to save size.
+        // TODO: this is a good candidate of discriminated union (https://github.com/dotnet/csharplang/issues/113).
+        private readonly StringLocalisationType type;
+        private readonly string string1;
+        private readonly string? string2;
```
bdach: Not a fan of these. Quite confusing to read and nebulous meaning, might end up being bug-prone. I'd rather have separate subclasses for separate localisation types.

osu.Framework/Localisation/LocalisableStringDescriptor.cs
```diff
@@ -0,0 +1,151 @@
...
+        private readonly object?[]? args;
```
bdach: Even less a fan of this type. IMO this should be less permissive (should be `object?[]` at least and coerce to an empty collection if passed null, or even arguably `object[]` and disallow passing null args at all).

If allocations are deemed an issue then maybe `IEnumerable.Empty<object?>()` or our own array-based analogue could be considered.

osu.Framework/Localisation/LocalisableStringDescriptor.cs
```diff
@@ -0,0 +1,151 @@
...
+        {
+            /// <summary>
+            /// The <see cref="LocalisableStringDescriptor"/> is constructed from plain text and not localisable.
+            /// </summary>
+            None,
+
+            /// <summary>
+            /// The <see cref="LocalisableStringDescriptor"/> is constructed from Romanization/Unicode pair.
+            /// </summary>
+            Romanization,
```
bdach: Annoying nitpick: should be
```suggestion
            Romanisation,
```
for consistency (American vs British spelling). Goes for every other place too.

osu.Framework/Localisation/LocalisationManager_LocalisedBindableString.cs
```diff
@@ -29,27 +29,43 @@ public LocalisedBindableString(LocalisedString text, IBindable<IResourceStore<st
 
             private void updateValue()
             {
-                string newText = preferUnicode.Value ? text.Text.Original : text.Text.Fallback;
-
-                if (text.ShouldLocalise && storage.Value != null)
-                    newText = storage.Value.Get(newText);
-
-                if (text.Args?.Length > 0 && !string.IsNullOrEmpty(newText))
+                if (text.TryGetPlainText(out string plain))
```
bdach: This huge `if;else if;else if;else` makes me think subtypes would be preferable even more by hiding the logic into a virtual call. Possibly worth benchmarking.

osu.Framework/Localisation/LocalisableStringDescriptor.cs
huoyaoyuan: Subclassing union is like how F# doing it. It will increase GC overhead.

> might end up being bug-prone.

I'm limiting the uses of the shared fields within this type.

osu.Framework/Localisation/LocalisableStringDescriptor.cs
```diff
@@ -0,0 +1,151 @@
...
+using osu.Framework.Configuration;
+
+#nullable enable
+
+namespace osu.Framework.Localisation
+{
+    /// <summary>
+    /// A descriptor representing text that can be localised and formatted.
+    /// </summary>
+    public readonly struct LocalisableStringDescriptor : IEquatable<LocalisableStringDescriptor>
```
iiSaLMaN: seems too long, I think `LocalisableString` is enough for this.

osu.Framework/Localisation/LocalisableStringDescriptor.cs
huoyaoyuan: Top level nullable because it won't be passed for non-formattable scenarios.
If allocations are deemed an issue, `IEnumerable.Empty<object?>()` will definitely bring more overhead. The WIP `System.ValueFormattableString` and `System.Variant` should be the future.

osu.Framework/Localisation/LocalisationManager_LocalisedBindableString.cs
huoyaoyuan: I chose not to subclassing because it has updatable text source. Subclassing will force the updater to set a correct type of descriptor, which is fine but hard to guarantee.

huoyaoyuan: Well, got an idea immediately, which will increase complexity of `SpriteText` slightly, but not increasing GC overhead.

bdach:
I'll readily admit I have little to no sense of the target scale of localisation work that will have to be made in the end, so I might very well be wrong, but all this feels too aggressively and prematurely optimised. I'll defer to the opinions of the core team since they have experience with localising stable.

huoyaoyuan:
> prematurely optimised

I admit this.

However, given the architecture, localisation changes will also involve plain text, especially quickly updated numbers. We should be careful to not introduce significant overhead to those.

osu.Framework/Localisation/LocalisableString.cs
```diff
@@ -0,0 +1,151 @@
...
+
+namespace osu.Framework.Localisation
+{
+    /// <summary>
+    /// A descriptor representing text that can be localised and formatted.
+    /// </summary>
+    public readonly struct LocalisableString : IEquatable<LocalisableString>
+    {
+        // Reuse string1 and string2 in different types to save size.
+        // TODO: this is a good candidate of discriminated union (https://github.com/dotnet/csharplang/issues/113).
```
smoogipoo: Any reason for not using explicit struct layouts instead?

smoogipoo:
On the topic of cleaning up `LocalisableString`... I'm not against pre-optimising it. However I wonder if the following structure could work instead:

```csharp
LocalisableString:
    LocalisationType Type;
    string Contents;
    ILocalisationData Data;

TranslatableString : ILocalisationData
    string Key;
    string Fallback;

    implicit operator LocalisableString(TranslatableString str) => new LocalisableString
    {
        Type = LocalisationType.Translation,
        Content = Fallback,
        Data = this
    }

LocalisationEngine
    GetBindable(LocalisableString str)
        switch (str.Type)
            case LocalisationType.Translation:
                data = (TranslatableString)str.Data
                // Use data for key lookup

Usage:
new SpriteText
{
    Text = new TranslatableString("my_key", "fallback string goes here")
}

SpriteText.Text == "fallback string goes here" (via "Content")
```

huoyaoyuan:
My current implementation:
28 bytes struct on every scenario

smoogipoo's idea:
plain text: 8 bytes struct + pre-allocated string
formattable: 8 bytes struct + allocated 24 bytes
unicode: 8 bytes struct + allocated 16 bytes

Yeah don't be afraid of allocation. Big structs can cause other problems.

smoogipoo:
Not sure how to interpret that finding. I'm not optimising for allocation, I'm optimising for readability of code while an improvement to the underlying mechanism hasn't been implemented.

As far as allocations are concerned, the data will always be heap-allocated. Essentially the struct is going to be 8 (str-ptr) + 8 (data-ptr) = 16 bytes in all cases.
And the heap allocation is fine because in the cases where data is used it's not going to be changed often. One exception is maybe localisation + formatting but at the same time you also have to be conscious of formatting too many times too, so I don't think it's a point of concern for us.

The only thing that needs optimisation is the plain text scenario, for which two pointers is the absolute best you can do.

huoyaoyuan:
> I'm not optimising for allocation, I'm optimising for readability of code while an improvement to the underlying mechanism hasn't been implemented in the language.

I was manually simulating the discriminated union feature. Complex simulation should only be used when it's getting a significant benefit, which is not the case here. The DU simulation's overhead is more on plain texts.

osu.Framework.Tests/Localisation/LocalisationTest.cs
```diff
@@ -81,6 +81,21 @@ public void TestFormatted()
...
+        public void TestFormattedInterpolation()
+        {
+            const string arg_0 = "formatted";
+
+            manager.AddLanguage("ja-JP", new FakeStorage("ja"));
+            config.Set(FrameworkSetting.Locale, "ja-JP");
+
+            string expectedResult = string.Format(FakeStorage.LOCALISABLE_FORMAT_STRING_JA, arg_0);
+
+            var formattedText = manager.GetLocalisedString(new TranslatableString(FakeStorage.LOCALISABLE_FORMAT_STRING_EN, interpolation: $"The {arg_0} fallback should only matches argument count"));
```
huoyaoyuan: `string` has higher overload priority than `FormattableString`, so the interpolation overload can only be selected by using named argument.
It's not easy to discover, but I really like this usage.

smoogipoo: What's the use of having the `FormattableString` overload in the first place?

osu.Framework/Testing/Drawables/Steps/StepSlider.cs
```diff
@@ -99,6 +99,6 @@ protected void Success()
             spriteText.Alpha = 0.8f;
         }
 
-        public override string ToString() => spriteText.Text;
```
smoogipoo: Not sure I like having to attach `ToString()` everywhere... o!f is a very simple use-case so it's not used all that much, could you update osu!-side with the same required changes, just to get an idea of how it feels?

osu.Framework/Localisation/RomanisableUnicodeString.cs
```diff
@@ -0,0 +1,32 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+#nullable enable
+
+namespace osu.Framework.Localisation
+{
+    /// <summary>
+    /// A string that has variation in Unicode and Romanised form.
+    /// </summary>
+    public class RomanisableUnicodeString
```
smoogipoo: Maybe just `RomanisableString`?

osu.Framework/Localisation/RomanisableUnicodeString.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+#nullable enable
+
+namespace osu.Framework.Localisation
+{
+    /// <summary>
+    /// A string that has variation in Unicode and Romanised form.
```
smoogipoo:
```suggestion
    /// A string that has variations in Unicode and Romanised form.
```

osu.Framework/Localisation/LocalisationManager_LocalisedBindableString.cs
```diff
@@ -29,27 +27,16 @@ public LocalisedBindableString(LocalisedString text, IBindable<IResourceStore<st
 
             private void updateValue()
             {
-                string newText = preferUnicode.Value ? text.Text.Original : text.Text.Fallback;
-
-                if (text.ShouldLocalise && storage.Value != null)
-                    newText = storage.Value.Get(newText);
-
-                if (text.Args?.Length > 0 && !string.IsNullOrEmpty(newText))
+                Value = text.Data switch
```
smoogipoo: I'm fine with this personally. But given that we haven't used switch-exprs anywhere else, I'd also want the opinion of @peppy .

osu.Framework.Tests/Localisation/LocalisationTest.cs
huoyaoyuan: Just for the interpolation syntax to combine fallback and arguments (personally I don't like writing `{0}`).

smoogipoo: Hmm I see, this is fine.

osu.Framework/Localisation/TranslatableString.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            if (localisedFormat == null) return ToString();
+
+            try
+            {
+                return string.Format(store.EffectiveCulture, localisedFormat, Args);
+            }
+            catch (FormatException)
+            {
+                // The formatting has failed
```
bdach: What's with this exception silencing? Should at least be logged at low severity IMO.

osu.Framework/Testing/Drawables/Steps/StepSlider.cs
huoyaoyuan: Currently game side doesn't use the implicit conversion to string. There are only 1 class and 2 cases to change (OsuDropdown).

When investigating, I think `IHasText` should be updated to use `LocalisableString`. Is this expected?

smoogipoo: Yeah, `IHasText` should be changed.

huoyaoyuan:
Well TextFlowContainer is blocking OsuCheckBox... Most other cases are just of changing the exposed string type.

huoyaoyuan:
I considered about TextFlowContainer:
- Let TextFlowContainer operate on characters, instead of SpriteText? This may add more duplication with SpriteText, and cannot support `AddText(SpriteText)` well.
- Store what's added into TextFlowContainer, and recreate SpriteTexts when locale changes. This adds a lot of overhead, and the implementation will change a lot.
- Move line wrapping capability to SpriteText. **This is my preferred option.**
  - OsuCheckBox and others that just require line wrapping can simply use SpriteText, to reduce complexity.
  - It's unlikely to see localisable long articles in game. SpriteText can provide the localisation capability paragraph by paragraph.
  - But the arranging mechanism of FlowContainer cannot be used then. TextFlowContainer is already using its own `computeLayout()`.
  - How about adding an internal Offset to SpriteText that's used by TextFlowContainer? Then a paragraph can use multiple cross-line SpriteText for each sentence.

@smoogipoo asking for your consideration here.

smoogipoo:
Can you describe the issue you're having with `TextFlowContainer` further?

huoyaoyuan:
Currently, the downside of `TextFlowContainer`:
- It split sentence into words, and create `SpriteText` for each word, to implement line wrapping. This is not friendly to localisation beacause different language won't have same words. Moreover, Chinese and Japanese don't use spaces in text, thus cannot be handled by current mechanism.

It has several usages:
- Used by `OsuCheckBox`. This should definitely localisable, but it's the simplest. It's only using the line wrapping capability. If the capability is moved to `SpriteText`, `OsuCheckBox` can use it directly.
- `MarkdownTextFlowContainer`. Markdown usually are returned from non-localisable source, so keeping it is acceptable.
- `OsuTextFlowContainer` are used by several potentially localisable texts that requires line wrapping.
- `LinkFlowContainer` is the most complex scenario. It's not composed sentence by sentence. It should be surely localisable.

I'm finding a solution that can handle **localisation** and **line wrapping/arrangement** at the same time. This is mostly for `LinkFlowContainer`.

smoogipoo:
I'm not against having such functionality present for `SpriteText` (or rather within `MultilineTextBuilder`), but the functionality should also be present within `TextFlowContainer`. For example `MarkdownTextBuilder` will need support for per-character line splitting in the case of CJ texts.

It would be nice if `TextBuilder` could be refactored/abstracted further to support common implementation between `SpriteText` and `TextFlowContainer`.

smoogipoo:
Yep, I think it's best to leave it to a separate PR.

huoyaoyuan:
OK, then this one should be ready to review, together with the game side PR.

*I'll be inactive on Discord during winter vacation. Tag me on GitHub please.*

osu.Framework/Localisation/ILocalisationStore.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System.Globalization;
+using osu.Framework.IO.Stores;
+
+namespace osu.Framework.Localisation
+{
+    public interface ILocalisationStore : IResourceStore<string>
+    {
+        /// <summary>
+        /// The <see cref="CultureInfo"/> that's same with content of this <see cref="ILocalisationStore"/> and can be used for formatting number etc.
```
smoogipoo, line 12: If I'm understanding this correct, I think it'd read better as `The <see cref="CultureInfo"/> corresponding to the content of this <see cref="ILocalisationStore"/>.`

osu.Framework/Localisation/ILocalisationStore.cs
```diff
@@ -0,0 +1,16 @@
...
+        /// </summary>
+        public CultureInfo EffectiveCulture { get; }
```
smoogipoo, line 14:
```suggestion
        CultureInfo EffectiveCulture { get; }
```

`public` is the default access modifier for interface methods.

osu.Framework/Localisation/LocalisableString.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// <summary>
+    /// A descriptor representing text that can be localised and formatted.
+    /// </summary>
+    public readonly struct LocalisableString : IEquatable<LocalisableString>
+    {
+        /// <summary>
+        /// The underlying data, can be <see cref="string"/>, <see cref="TranslatableString"/>, or <see cref="RomanisableString"/>.
+        /// </summary>
+        internal readonly object? Data;
+        // it's somehow common to call default(LocalisableString), and we should return empty string then.
```
smoogipoo, line 19: Probably should reside somewhere close to `ToString()`.

osu.Framework/Localisation/LocalisationManager_LocalisedBindableString.cs
smoogipoo: Let's not use switch-exprs for now.

peppy:
@smoogipoo you're still okay with this direction, yeah? seems like this was basically in a ready-state last time you reviewed.

peppy (approved):

smoogipoo:
Yep I was fine with this last time. I think my only remaining consideration was breaking changes.

huoyaoyuan:
I've been busy for a long time and won't be available until later this year. You can make changes on this based on your decision.

peppy:
Thanks for the heads up! Luckily you left this PR in a very good state.

peppy:
While applying changes osu-side for this from scratch (just to get a feel for how large the breaking changes are) I noticed that the order of parameters for `RomanisableString` is opposite of what we used to have for `LocalisedString`. I'm not hugely against this, but was this an intentional change?

peppy:
I've pushed some fixes to actually make `SpriteText` work (not sure if this regressed somewhere along the way. @smoogipoo need a review on this since it seems like it would be quite easy to break something.

bdach:
> I noticed that the order of parameters for `RomanisableString` is opposite of what we used to have for `LocalisedString`.

Having the original "source/unromanised" text go first and the romanised variant go second feels slightly more correct to me, but I wouldn't have personally cared much if it was the other way around. Seeing that you already seem to have gone through game-side usages and flipped the order I'd probably leave that as it is now.

smoogipoo:
@peppy I don't know the reasoning for the change in order, but I'm fine with it being this way.

smoogipoo:
The `SpriteText` changes look fine to me. I'll leave an approval with your merge.

smoogipoo (approved):

PR #3161, opened 2020-01-07
Provide current mouse position with each mouse event

iiSaLMaN:
This PR will allow the mouse events to be shared across multiple positional-affecting input as the event name will eventually be changed to `PositionalEvent` or similar.

### `UIEvent.MousePosition` should not be used unless targeting the mouse position directly, use `MouseEvent.CurrentMousePosition` instead.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
@@ -7,7 +7,6 @@ using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics.UserInterface;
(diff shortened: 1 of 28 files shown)


smoogipoo:
I still don't understand what this is really achieving, and your breaking change comment isn't helping. 

Beyond the rename, are you eventually hoping to pass a different "current" position in for each touch point? So:
```
MousePosition: The current cursor position from a dominating pointing device.
CurrentMousePosition: The current cursor position from any pointing device.
```
If so, this feels exceedingly convoluted, you can't use the terminology `CurrentMousePosition` for this.

iiSaLMaN:
Yes! that's pretty much the goal of this, sorry for not including it in the PR.

iiSaLMaN:
Closing as not required for now.

PR #3164, opened 2020-01-08
Disallow binding to already bound object

UselessToucan:
I guess we should not allow binding to the same object multiple times.

### DIFF
--- a/osu.Framework/Bindables/Bindable.cs
+++ b/osu.Framework/Bindables/Bindable.cs
@@ -182,6 +182,7 @@ namespace osu.Framework.Bindables
         /// This will adopt any values and value limitations of the bindable bound to.
         /// </summary>
         /// <param name="them">The foreign bindable. This should always be the most permanent end of the bind (ie. a ConfigManager).</param>
+        /// <exception cref="InvalidOperationException">Thrown when attempting to bind to an already bound object.</exception>
         public virtual void BindTo(Bindable<T> them)
         {
             if (Bindings != null && Bindings.Contains(them))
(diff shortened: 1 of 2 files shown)


osu.Framework/Bindables/Bindable.cs
```diff
@@ -184,6 +184,9 @@ public Bindable<T> BindTarget
         /// <param name="them">The foreign bindable. This should always be the most permanent end of the bind (ie. a ConfigManager).</param>
         public virtual void BindTo(Bindable<T> them)
```
iiSaLMaN: since this is public, i think adding an exception xmldoc would be better.

osu.Framework/Bindables/Bindable.cs
```diff
@@ -184,6 +184,9 @@ public Bindable<T> BindTarget
...
         {
+            if (Bindings != null && Bindings.Contains(them))
+                throw new InvalidOperationException($"The bindable is already bound to {them}");
```
iiSaLMaN:
```suggestion
                throw new InvalidOperationException($"This bindable is already bound to {them}");
```

smoogipoo:
Keep in mind that this makes `BindTo()` now effectively N^2 complexity, which means complex hierarchies such as hitobjects that bind many times to higher levels, are going to become noticeably slower.

Is this actually an issue somewhere? Because it seems like bindables would behave fine regardless of being bound to multiple times.

peppy:
It would have fired events multiple times, at very least.

Wouldn't have thought that this was an issue, but I guess we did (or are?)
 do pretty massive bindings on `DrawableHitObject`s at one point...

smoogipoo:
Upon further discussion, this is unlikely to be a problem with proper use of bindables (bindings happening bottom-up).

PR #3165, opened 2020-01-09
Fix weak list incorrectly containing removed objects

smoogipoo:
Pathway wasn't tested :/

Would now cause unbind + re-bind to bindables to fail.

### DIFF
--- a/osu.Framework/Lists/WeakList.cs
+++ b/osu.Framework/Lists/WeakList.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 
 namespace osu.Framework.Lists
 {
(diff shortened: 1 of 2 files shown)


osu.Framework/Lists/WeakList.cs
```diff
@@ -49,9 +49,27 @@ public bool Remove(WeakReference<T> weakReference)
             return found;
         }
 
-        public bool Contains(T item) => list.Any(t => t.Reference.TryGetTarget(out var obj) && obj == item);
+        public bool Contains(T item)
+        {
+            foreach (var t in list)
```
peppy: was this changed from linq as a performance/alloc consideration?

smoogipoo: Yeah, do you want benchmarks? Though I think by now we're pretty certain that linq is more expensive in 99.9% of cases :p

peppy: just checking, since the change could have been made within the original linq operation.

PR #3170, opened 2020-01-11
Fix tab switching to include items in dropdown

peppy:
Closes https://github.com/ppy/osu-framework/issues/3169.


peppy:
Is this still required with https://github.com/ppy/osu-framework/pull/3183?

smoogipoo:
Yeah, since it doesn’t use `Items`

TheWildTree:
Is there a chance of getting this merged anytime soon?

PR #3171, opened 2020-01-11
Fix text boxes crashing upon trying to delete words when empty

EVAST9919:
Fixes https://github.com/ppy/osu/issues/7505


bdach:
Needs a test.

bdach:
I checked out the code locally to fix the test case attached, but while doing so and reading neighbouring code I realised that this was going to be an issue for deleting next word too due to clamping being used in the exact same way, so I went ahead and I pushed a test case in, as well as adjusted the fix to a more proper one in my opinion.

In general the `MathHelper.Clamp` -> `Math.Clamp` has changed semantics somewhat - the old version was playing fast & loose with the bounds and now we might start seeing those types of exceptions more. Which is not necessarily a bad thing - the old way could have introduced subtle bugs as the order of the operations taken there (min/max first) would influence the clamp result if the bounds were improper.

Issue #2527, opened 2019-06-12
OptimisingContainer needs to schedule unloads even when hidden


peppy:
`DelayedLoadUnloadWrapper`s are currently not unloaded if their `OptimisingContainer` is not being updated.

Case in point: fullscreen overlays (in osu!) may be hidden themselves, and therefore their contained `ScrollContainer`s are no longer able to unload their optimising children. They still need to.

We may consider scheduling the unloads to a higher level component.

PR #3180, opened 2020-01-15
Rework DLW/DLUW to support general unloads and intersections

smoogipoo:
Resolves https://github.com/ppy/osu-framework/issues/2527

Also allows unloads to happen with non-scrolling containers (e.g. if hidden by movement inside a masking area).

The following tests now pass:
image

Also tested with social overlay osu!-side.

### DIFF
--- a/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
+++ b/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
@@ -136,9 +136,11 @@ namespace osu.Framework.Graphics.Containers
 
         public override bool Invalidate(Invalidation invalidation = Invalidation.All, Drawable source = null, bool shallPropagate = true)
         {
+            bool result = base.Invalidate(invalidation, source, shallPropagate);
+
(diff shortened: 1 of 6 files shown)


smoogipoo:
Blocking until intermittent test case is fixed: https://ci.appveyor.com/project/peppy/osu-framework/builds/30125576

smoogipoo:
Fixed.

The issue was `timeToUnload = 0` causing instantaneous unloads. Don't know whether this is correct but it is the previous functionality.

osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
```diff
@@ -125,78 +126,66 @@ protected virtual void CancelTasks()
 
         private readonly Cached optimisingContainerCache = new Cached();
         private readonly Cached isIntersectingCache = new Cached();
-
-        private ScheduledDelegate loadScheduledDelegate;
+        private ScheduledDelegate isIntersectingReset;
```
peppy: not sure how well this reads without Delegate in the variable name

PR #3181, opened 2020-01-16
Expose alive children from Container

smoogipoo:
Useful for cases where you want to iterate over only alive children for performance reasons, e.g.:

https://github.com/ppy/osu/blob/ef735f106e0899fc11554b7d290d4c5bd1262323/osu.Game/Rulesets/UI/HitObjectContainer.cs#L16

https://github.com/ppy/osu/blob/ef735f106e0899fc11554b7d290d4c5bd1262323/osu.Game/Screens/Edit/Compose/Components/BlueprintContainer.cs#L484

### DIFF
--- a/osu.Framework/Graphics/Containers/Container.cs
+++ b/osu.Framework/Graphics/Containers/Container.cs
@@ -43,6 +43,11 @@ namespace osu.Framework.Graphics.Containers
                 internalChildrenAsT = (IReadOnlyList<T>)InternalChildren;
             else
                 internalChildrenAsT = new LazyList<Drawable, T>(InternalChildren, c => (T)c);
+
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Containers/Container.cs
```diff
@@ -75,6 +80,21 @@ public IReadOnlyList<T> Children
             set => ChildrenEnumerable = value;
         }
 
+        /// <summary>
+        /// The publicly accessibly list of alive children. Forwards to the alive children of <see cref="Content"/>.
```
peppy:
```suggestion
        /// The publicly accessible list of alive children. Forwards to the alive children of <see cref="Content"/>.
```

PR #3184, opened 2020-01-16
Implement and handle visibility of Windows Touch Keyboard

iiSaLMaN:
- Also tidied up windows native code while I was implementing this.

Closes https://github.com/ppy/osu-framework/issues/3175


bdach:
Unless I'm missing something, the "tidy-up" change, which basically seems to boil down to changing access modifiers, doesn't have any practical effect, and IMO should be split off into a separate PR to stop obscuring the diff of the actual change here, if not completely reverted.

iiSaLMaN:
> Unless I'm missing something, the "tidy-up" change, which basically seems to boil down to changing access modifiers, doesn't have any practical effect, and IMO should be split off into a separate PR to stop obscuring the diff of the actual change here, if not completely reverted.

It was actually a simple change, but since the file is too big it looked like the change is big, will remove.

osu.Framework/Platform/Windows/Native/Input.cs
```diff
@@ -2660,4 +2660,110 @@ public enum WindowsMessages : uint
...
+
+        /// <summary>
+        /// Changes the cursor to a question mark with a pointer. If the user then clicks a control in the dialog box, the control receives a <see cref="WindowsMessages.HELP"/> message.
+        /// </summary>
+        CONTEXTHELP = 0xF180,
+
+        /// <summary>
+        /// Selects the default item; the user double-clicked the window menu.
+        /// </summary>
+        DEFAUKT = 0xF160,
```
bdach:
```suggestion
        DEFAULT = 0xF160,
```

osu.Framework/Platform/Windows/WindowsTextInput.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace osu.Framework.Platform.Windows
+{
+    public class WindowsTextInput : GameWindowTextInput
+    {
+        public WindowsTextInput(IWindow window)
+            : base(window)
+        {
+        }
+
+        private void startTabTip() => Task.Run(() => Process.Start(new ProcessStartInfo
```
bdach: Is there *no saner* way to do this, rather than fire-and-forget launching a subprocess per call?

osu.Framework/Platform/Windows/WindowsTextInput.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            FileName = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.CommonProgramFiles), @"microsoft shared\ink\TabTip.exe"),
+            UseShellExecute = true,
+        }));
+
+        private void closeTouchKeyboard()
+        {
+            var kbdWnd = Methods.FindWindow("IPTIP_Main_Window", null);
```
bdach: This is also very WTF, what with scraping all open windows for that particular one.

osu.Framework/Platform/Windows/WindowsTextInput.cs
iiSaLMaN: so executing this program called `TabTip` will show the touch keyboard if Windows is in tablet mode and no physical keyboard attached (tested), this is the most simplest and also *sanist way* to show the touch keyboard, there are other methods which require very deep and complicated code.

[There was one easy method that didn't require starting the TabTip](https://stackoverflow.com/a/39385492), but it can be broken by simply hiding the touch keyboard icon from the task bar, which i'm pretty sure will be rejected with no doubts.

osu.Framework/Platform/Windows/WindowsTextInput.cs
iiSaLMaN: That's seemingly the only way to close the touch keyboard window. (looking for `TabTip` and trying to close/kill it doesn't seem to work)

iiSaLMaN:
> This seems very crudely hacked together. I'd say we need to use a saner API if Windows provides one, or arguably not do it at all.

Unfortunately there is no API to use for showing the touch keyboard on Windows, this is seemingly the only sane way.

> The current method needs work as-is anyway (there's some frivolous WinAPI and process usage here, I'd say)

I would like to hear if there's anything I can do to make this better.

bdach:
There might be an API, but undocumented - see [this Stack Overflow post](https://stackoverflow.com/a/40921638).

In general the way you're invoking it now is unacceptable for me, because you're (1) managing something *the OS* should be managing for you and (2) doing it in a very fragile, non-resilient way (what if the subprocess doesn't spawn? why is the keyboard dismissed by killing the process? what if a future OS update changes the window title and your hack to kill it stops working?)

If the best Windows 10 can do is a reverse-engineered undocumented interface, then this feature isn't worth implementing for me. At least until something *resembling* a stable API shows up.

smoogipoo:
You may be able to use https://docs.microsoft.com/ja-jp/windows/win32/api/peninputpanel/nn-peninputpanel-itextinputpanel , which seems to reside in `Tiptsf.dll`.
Otherwise as @bdach said, I’d rather not implement this.

jai-x:
I was under the impression that this functionality was handled automatically with the use of the SDL2 window (https://github.com/ppy/osu-framework/pull/3013)? I think you can test this by running lazer with the flag `--sdl` as a launch option.

osu.Framework/Platform/Windows/Native/Methods.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace osu.Framework.Platform.Windows.Native
+{
+    internal static class Methods
```
huoyaoyuan: If P/Invokes are defined in a separated class, the class should be named `NativeMethods` according to .NET guideline.

But given current code structure, I think it should have a descriptive name, in contrast to `Input` and `Execution`.

swoolcock:
The SDL2 change is unrelated.

osu.Framework/Platform/Windows/Native/Methods.cs
bdach: Out of scope for this PR. Should be a separate cleanup one if deemed necessary.

iiSaLMaN:
@smoogipoo I've looked up on that, with help from @bdach we've found out that there's a [managed library] for these interfaces and classes and also found out that it's .NET Framework-specific, i've tried playing with this but it doesn't seem to work at all, `COMException` is always thrown when trying to access any function with description telling that the COM call returned `E_FAIL`.

swoolcock:
Would appreciate if someone could test this function on a Windows 10 tablet PC and see if it triggers the software keyboard.  I’m thinking the “on some platforms” probably applies to operating systems that don’t traditionally have hardware keyboards.  It would still be beneficial to test.

https://wiki.libsdl.org/SDL_StartTextInput

bdach:
FWIW, I don't think it's actually necessary to have a tablet PC, I have a tablet mode setting in my notification sidebar menu on Windows 10 and it does seem to be able to call the keyboard (albeit not unprompted).

iiSaLMaN:
Closing as bad implementation.

PR #3123, opened 2019-12-18
Add DefaultChanged event to bindables

peppy:
Please review stringently.

PR #3188, opened 2020-01-19
Fix binding leased to bindable potentially throwing exceptions

### LINKED ISSUE #3108: Binding to a leased bindable provides unexpected behaviour
Binding to a leased bindable doesn't allow changing the value. It probably should.

```csharp
var copy = leased.GetBoundCopy();
copy.Value = "test"; // works
```

```csharp
var copy = new BindableString();
leased.BindTo(copy);
copy.Value = "test"; // doesn't work
```

iiSaLMaN:
- Forgotten override after addition of modifiable `Default` in #3123

Closes #3108

### DIFF
--- a/osu.Framework.Tests/Bindables/BindableLeasingTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableLeasingTest.cs
@@ -263,5 +263,28 @@ namespace osu.Framework.Tests.Bindables
             leased.UnbindAll();
             leased.UnbindAll();
         }
+
+        [Test]
+        public void TestBindingToLeased()
+        {
+            var leased = original.BeginLease(false);
+
+            var copy1 = leased.GetBoundCopy();
+            copy1.Value = 2;
+            Assert.AreEqual(original.Value, 2);
+            Assert.AreEqual(original.Value, leased.Value);
+            Assert.AreEqual(original.Value, copy1.Value);
+
+            var copy2 = new Bindable<int>();
+            leased.BindTo(copy2);
+            copy2.Value = 3;
+            Assert.AreEqual(copy2.Value, 3);
+            Assert.AreEqual(copy2.Value, leased.Value);
+
+            leased.Value = 4;
+            Assert.AreEqual(original.Value, 4);
+            Assert.AreEqual(original.Value, copy2.Value);
+            Assert.AreEqual(original.Value, leased.Value);
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


osu.Framework/Bindables/LeasedBindable.cs
```diff
@@ -73,6 +73,20 @@ public override T Value
...
+        {
+            get => base.Default;
+            set
+            {
+                if (source != null)
+                    checkValid();
+
+                if (EqualityComparer<T>.Default.Equals(Default, value)) return;
+
+                SetValue(base.Default, value, true);
```
smoogipoo: SetValue?

osu.Framework.Tests/Bindables/BindableLeasingTest.cs
```diff
@@ -263,5 +263,28 @@ public void TestUnbindAllReturnsLease()
             leased.UnbindAll();
             leased.UnbindAll();
         }
+
+        [Test]
+        public void TestBindingToLeased()
+        {
+            var leased = original.BeginLease(false);
+
+            var copy1 = leased.GetBoundCopy();
```
smoogipoo: So “copy” that is a copy of the leased here, but “copy2” is a completely separate bindable below?

Please split into multiple testcases and name the methods based on what they test (look at all other methods in this class), and rename the variables to be more descriptive.

osu.Framework/Bindables/LeasedBindable.cs
iiSaLMaN: oops

osu.Framework/Bindables/Bindable.cs
```diff
@@ -118,11 +118,11 @@ public virtual T Default
 
                 if (EqualityComparer<T>.Default.Equals(defaultValue, value)) return;
 
-                setDefault(defaultValue, value);
+                SetDefault(defaultValue, value);
             }
         }
 
-        private void setDefault(T previousValue, T value, bool bypassChecks = false, Bindable<T> source = null)
+        protected void SetDefault(T previousValue, T value, bool bypassChecks = false, Bindable<T> source = null)
```
smoogipoo:
```suggestion
        internal void SetDefault(T previousValue, T value, bool bypassChecks = false, Bindable<T> source = null)
```

smoogipoo:
seems ok.

PR #3190, opened 2020-01-20
Make OnDrag, OnDragEnd, and OnMouseUp return void instead of bool

smoogipoo:
While working on a related fix, I found an issue relating to mouse input whereby a hierarchy that received mouse-down events may not receive mouse-up events if a drawable above them blocked it.

This isn't limited to `OnMouseUp`, but all mouse-release events (`OnMouseUp`, `OnDrag`, `OnDragEnd`), for lack of better naming.

I'm doing this only for mouse events for now since they're the only input events that are correctly. Key, joystick, and keybinding events are all broken in other ways (I'm working on fixing them). Those other events will be changed after said fixes have been applied.

Here's an example illustrating this, although this isn't really testable anymore (hence why I haven't added a test):

```csharp
[Test]
public void DoTest()
{
    var receptors = new InputReceptor[3];

    AddStep("setup", () =>
    {
        Children = new Drawable[]
        {
            receptors[0] = new InputReceptor
            {
                Size = new Vector2(100),
                MouseDown = () => true,
            },
            receptors[1] = new InputReceptor { Size = new Vector2(100) },
            receptors[2] = new InputReceptor
            {
                Size = new Vector2(100),
                MouseUp = () => true,
            }
        };
    });

    AddStep("move mouse to receptors", () => InputManager.MoveMouseTo(receptors[0]));
    AddStep("mousedown", () => InputManager.PressButton(MouseButton.Left));
    AddStep("mouseup", () => InputManager.ReleaseButton(MouseButton.Left));

    AddAssert("all received down", () => receptors.All(r => r.DownReceived)); // Passes
    AddAssert("0 received up", () => receptors[0].UpReceived); // Fails
    AddAssert("1 received up", () => receptors[1].UpReceived); // Fails
    AddAssert("2 received up", () => receptors[2].UpReceived); // Passes
}

private class InputReceptor : Box
{
    public bool DownReceived;
    public bool UpReceived;

    public Func<bool> MouseDown;
    public Func<bool> MouseUp;

    {
        DownReceived = true;
        return MouseDown?.Invoke() ?? false;
    }

    protected override bool OnMouseUp(MouseUpEvent e)
    {
        UpReceived = true;
        return MouseUp?.Invoke() ?? false;
    }
}
```

# Breaking changes:

# vNext

## `OnDrag`, `OnDragEnd`, and `OnMouseUp` now return `void`

These events can no longer be blocked by other `Drawable`s in the hierarchy.

This guarantees that `Drawable`s that have received the respective `OnDragStart` and `OnMouseDown` events will receive accompanying `OnDrag`, `OnDragEnd` and `OnMouseUp` events

### DIFF
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -1870,7 +1870,8 @@ namespace osu.Framework.Graphics
                     return OnMouseDown(mouseDown);
 
                 case MouseUpEvent mouseUp:
-                    return OnMouseUp(mouseUp);
+                    OnMouseUp(mouseUp);
+                    return false;
 
                 case ClickEvent click:
                     return OnClick(click);
@@ -1882,10 +1883,12 @@ namespace osu.Framework.Graphics
                     return OnDragStart(dragStart);
 
                 case DragEvent drag:
-                    return OnDrag(drag);
+                    OnDrag(drag);
+                    return false;
 
                 case DragEndEvent dragEnd:
-                    return OnDragEnd(dragEnd);
+                    OnDragEnd(dragEnd);
+                    return false;
 
                 case ScrollEvent scroll:
                     return OnScroll(scroll);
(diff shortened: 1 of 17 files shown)


osu.Framework/Graphics/Drawable.cs
```diff
@@ -1932,12 +1935,12 @@ protected virtual void OnHoverLost(HoverLostEvent e)
         }
 
         protected virtual bool OnMouseDown(MouseDownEvent e) => Handle(e);
-        protected virtual bool OnMouseUp(MouseUpEvent e) => Handle(e);
+        protected virtual void OnMouseUp(MouseUpEvent e) => Handle(e);
```
iiSaLMaN: Small nitpick, but for the sake of consistency, this should have brackets like the other void functions
```diff
- protected virtual void OnMouseUp(MouseUpEvent e) => Handle(e);
+ protected virtual void OnMouseUp(MouseUpEvent e)
+ {
+     Handle(e);
+ }

File: osu.Framework/Graphics/Drawable.cs [hunk 2]
```diff
@@ -1932,12 +1935,12 @@ protected virtual void OnHoverLost(HoverLostEvent e)
...
         protected virtual bool OnClick(ClickEvent e) => Handle(e);
         protected virtual bool OnDoubleClick(DoubleClickEvent e) => Handle(e);
         protected virtual bool OnDragStart(DragStartEvent e) => Handle(e);
-        protected virtual bool OnDrag(DragEvent e) => Handle(e);
-        protected virtual bool OnDragEnd(DragEndEvent e) => Handle(e);
+        protected virtual void OnDrag(DragEvent e) => Handle(e);
+        protected virtual void OnDragEnd(DragEndEvent e) => Handle(e);
```
iiSaLMaN: here as well.

osu.Framework/Graphics/Drawable.cs
smoogipoo: not refactoring this in this PR.

smoogipoo: not refactoring this in this PR.

PR #3196, opened 2020-01-21
Reorder text box characters depth after removal action

### LINKED ISSUE #2904: Characters in TextBox get added in incorrect order
I seem to have found a consistent case in which the characters in a TextBox are added incorrectly.

Steps to reproduce
1) Type a word that is at least a few characters long
2) Delete any 2 characters but keep the last one untouched
3) Go back to the end and type a single character
4) Repeat 2 and 3

The character now appears to the right of the Charet and going through the characters the Charet jumps around. It also can't be (visually) selected.
This is repeatable if the text is long enough as only one character can be added this way each time before it unscrambles itself.
Here a little [video](https://streamable.com/vkkz3) for the quick bamboozle.

iiSaLMaN:
- This would fix cases where newly added text would conflict depths with existing ones which causes ordering and incorrect caret positioning issues.

Before reordering on removal:
LhiDfdVxfD
After reordering on removal:
7BL0OdBYjP

Closes #2904

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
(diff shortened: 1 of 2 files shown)


osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
```diff
@@ -323,11 +324,114 @@ public void TestNextWordDeletion()
...
+            });
+
+            AddStep("focus textbox", () =>
+            {
+                InputManager.MoveMouseTo(textBox);
+                InputManager.Click(MouseButton.Left);
+            });
+
+            AddStep("insert word", () => textBox.InsertString("eventext"));
+            AddStep("remove 2 letters", remove2FirstLetters);
```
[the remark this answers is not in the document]
bdach: Number of characters to delete should be a parameter, too. Don't need to change it, it just looks better IMO.

osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
```diff
@@ -323,11 +324,116 @@ public void TestNextWordDeletion()
             AddAssert("text is empty", () => textBox.Text.Length == 0);
         }
 
+        /// <summary>
+        /// Removes first 2 characters and append them, this tests layout positioning of the characters in the text box.
+        /// </summary>
```
bdach: Not sure xmldocs on test cases do that much good, the steps are pretty self-explanatory.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -450,6 +451,10 @@ private bool removeCharacterOrSelection(bool sound = true)
 
             text = text.Remove(start, count);
 
+            // Reorder characters depth after removal to avoid ordering issues with newly added characters.
+            for (int i = 0; i < TextFlow.Count; i++)
```
bdach: I think
```suggestion
            for (int i = start; i < TextFlow.Count; i++)
```
is a valid optimisation to make here, since the first `start` characters should have their depth set appropriately. Seems to pass tests, at least. It's not intensely important though; maybe even better left as-is to be absolutely sure of correctness.

osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
iiSaLMaN: Added it to mention that this is test is necessary for the sake of testing layout positioning.

osu.Framework/Graphics/UserInterface/TextBox.cs
iiSaLMaN: Would need input on this one, I'm aware of this causing the bug to show up again somehow and causing a twisted issue to solve in the future.

peppy: If it's passing your tests and you're still cautious of it, this means your tests are not covering the issue.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -484,6 +489,9 @@ protected virtual Drawable AddCharacterToFlow(char c)
             Drawable ch = GetDrawableCharacter(c);
             ch.Depth = -selectionLeft;
 
+            // Assert no existing character has same depth as this to avoid ordering issues on a text box.
+            Trace.Assert(TextFlow.All(d => d.Depth != ch.Depth), $"The {nameof(TextFlow)} has more than one character with the same depth.");
```
peppy: This will be running every character insertion on release builds and will throw a hard error to the user. Doesn't sound like something we'd want.

Are you sure we want to be doing this? Did you mean `Debug.Assert`?

iiSaLMaN: Actually, I think it may not be necessary since there are tests for this, will remove.

peppy:
I'm slightly uncomfortable with the depth being set in three different places. If we're going to continue doing this, then I think there should be a `getDepthForCharacterIndex(int index)` or similar.

@smoogipoo interested to hear your thoughts in general on this change

smoogipoo:
Only two different places, so it's not super bad. Wouldn't mind a helper method for this though that's also used in addition.

In the long term, it seems like what we really want is a custom `Container` that contains children in an absolutely-ordered fashion (a-la `List<T>`). As for how to achieve this, I'm not really sure right now - there are potential pathways such as a virtual function to create the children list.
We probably will run into this at some point again in the future, especially after supporting multi-line `TextBox`es, for performance reasons.

peppy:
Okay, let's go with this for now.

PR #3201, opened 2020-01-22
Make OnDoubleClick return void instead of bool

smoogipoo:
It's a singular target event which has no blocking behaviour.


smoogipoo:
Turns out this is actually needed, as it _does_ block the next click from happening.

Issue #2909, opened 2019-10-20
Ensure consistent initial state in manual input visual test cases


bdach:
Running test cases which use manual input inside the visual test browser is susceptible to state pollution. This is especially pertinent in the case of mouse input, wherein the pre-existing mouse position within the test area can affect test execution, if the test author doesn't set the position explicitly themselves.

[Here's a video with an example.](https://streamable.com/d732k) I'm running the same test twice; once I purposefully (but might as well have accidentally) move the mouse over the test area, which then stays the same throughout the test, and once I make extra care not to do that, which makes the mouse stay at (0,0). This affects the outcome of the test - first case is a pass, the second is a fail.

I think this is an asterisk that doesn't need to be there - these sort of issues lead to devs running tests locally, seeing them pass, but then fail at CI time for not entirely clear reasons. Making sure there's a set initial state when using manual test input ensures consistent testcase execution and reduces chances of test interference.

I think mouse position is going to be the major and possibly only case here, as all of the other cases are probably less likely, but I haven't tested for sure.

peppy:
This looks really weird. Tests are supposed to isolate input (see the "local" "test" buttons top-right) and it looks to be set to the correct isolation level here. I think the capturing/isolating may not be occurring due to the game context in the test you're testing with? ie. because that test is making an `OsuGame` which has its own `UserInputManager`

iiSaLMaN:
I believe this is not really an issue anymore, unless i'm missing something?

peppy:
I believe it should still be reproducible, but it would be nice to have a solid reproduction case.. @bdach happen to know what was causing the issue in your case? it looks like it could be that the window was considered non-active in the last run of your video?

bdach:
I'm pretty sure it is still reproducible, yes, and it's not exclusive to scenes with nested input managers either. However I think it's (somewhat) easily fixable.

The default behaviour for manual input test scenes is to preserve any prior mouse position:

https://github.com/ppy/osu-framework/blob/77488e39f9dbf66a35f4161305a2e42c6937963a/osu.Framework/Testing/ManualInputManagerTestScene.cs#L24-L28

It's not a great default in my opinion and something better should be used instead (obvious candidate is `Vector2(0)`). I'd even advocate for entirely removing the `null` case for preserving prior position to guarantee independence and consistency of test case execution.

[...]

PR #3204, opened 2020-01-23
Fix ManualInputManager's initial cursor position not being deterministic

peppy:
Closes https://github.com/ppy/osu-framework/issues/2909.

### DIFF
--- a/osu.Framework/Testing/ManualInputManagerTestScene.cs
+++ b/osu.Framework/Testing/ManualInputManagerTestScene.cs
@@ -35,7 +35,13 @@ namespace osu.Framework.Testing
         private readonly BasicButton buttonLocal;
 
         [SetUp]
-        public virtual void SetUp() => ResetInput();
+        public void SetUpBase()
+        {
+            ResetInput();
+            SetUp();
+        }
+
+        public virtual void SetUp() { }
 
         protected ManualInputManagerTestScene()
         {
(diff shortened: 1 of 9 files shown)


iiSaLMaN:
Is there gonna be a follow-up PR of calling `base.SetUp()` for all `ManualInputManagerTestScene` derivers? (`TestSceneButton` is not the only one) If not, I will do.

peppy:
It should be called in all cases.

iiSaLMaN:
Actually there's just one test scene `TestSceneSliderBall`

osu.Framework/Testing/ManualInputManagerTestScene.cs
```diff
@@ -36,7 +35,13 @@ public abstract class ManualInputManagerTestScene : TestScene
         private readonly BasicButton buttonLocal;
 
         [SetUp]
-        public virtual void SetUp() => ResetInput();
+        public void SetUpBase()
```
smoogipoo: Hmm... Typically in `nunit` you're able to apply `[SetUp]` to methods in inheritance hierarchies. Does this currently work with our testing system? If not we should probably make that work instead and remove the `virtual` altogether.

peppy: Only reason I did it this way is to avoid it being a breaking change, but I'm fine to change it if you're okay with that.

smoogipoo: One blocker for this is that nunit specifies the base SetUp to run before the derived one, whereas we don't ensure that ourselves: https://nunit.org/docs/2.5/setup.html

Versus:

image

Otherwise, how about obsoleting? I don't think there'll be any changes required osu!-side since this is a `FrameworkTestScene` anyway (whereas osu! we generally require deriving `OsuTestScene` instead).

PR #3209, opened 2020-01-27
Make OnKeyDown() repeats only occur on still-valid Drawables

smoogipoo:
Similar to the old behaviour. This is probably more expected behaviour from the framework considering that mouse buttons do the same thing on click.

It would also solve the issue presented in https://github.com/ppy/osu-framework/pull/3206, but that's still a valid PR.

Adds further test cases for these scenarios.

### DIFF
--- a/osu.Framework/Input/KeyEventManager.cs
+++ b/osu.Framework/Input/KeyEventManager.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Collections.Generic;
+using System.Linq;
 using osu.Framework.Graphics;
 using osu.Framework.Input.Events;
 using osu.Framework.Input.States;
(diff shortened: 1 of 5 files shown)


osu.Framework/Input/KeyEventManager.cs
```diff
@@ -19,7 +20,13 @@ public KeyEventManager(Key key)
         {
         }
 
-        public void HandleRepeat(InputState state) => PropagateButtonEvent(ButtonDownInputQueue, new KeyDownEvent(state, Button, true));
+        public void HandleRepeat(InputState state)
+        {
+            // Only drawables that can still handle input should handle the repeat
+            var drawables = ButtonDownInputQueue.Intersect(InputQueue).Where(t => t.IsAlive && t.IsPresent);
```
UselessToucan: What about performance? `InputQueue` creates a new list and walks the whole scene graph on each call.

peppy: Input events (even repeats) should be seldom enough that this shouldn't be an issue.

smoogipoo: Furthermore, excluding the `Intersect()` this is no worse than it was before any of this abstraction. And `Intersect` is quite well optimised in LINQ (this all happens within a few microseconds for thousands of drawables).

PR #3220, opened 2020-01-31
Add the ability to add teardown steps to test scenes

peppy:


smoogipoo:
@peppy see if you agree with my changes, I had to refactor it to convince myself it was correct. otherwise g2g.

PR #3226, opened 2020-01-31
Consider grid cell content presence and lifetime when auto-sizing

### LINKED ISSUE #3214: Auto-sizing grid cells don't consider presence
Calculations should be functionally equivalent to the standard auto-size computation of `CompositeDrawable`, but it currently doesn't consider `IsPresent`.

```
new GridContainer
{
    Size = new Vector2(200),
    Content = new[]
    {
        new Box { RelativeSizeAxes = Axes.Both },
        new Box { RelativeSizeAxes = Axes.Both },
        new Box { Size = new Vector2(50), Alpha = 0 }
    },
    ColumnDimensions = new[]
    {
        new Dimension(),
        new Dimension(),
        new Dimension(GridSizeMode.AutoSize),
    }
}
```

Expected: cell column sizes are `(100, 100, 0)`
Actual: cell column sizes are `(75, 75, 50)`

bdach:
Resolves #3214.

# Summary

If a `GridContainer` cell's content is not present and alive at the time of update, do not consider its bounding box dimensions and assume 0 instead.

# Remarks

Various test cases attached, also including setting `IsPresent = true` in addition to `Alpha = 0` to make sure that case still behaves as desired. Can expand if coverage is deemed insufficient.

### DIFF
--- a/osu.Framework.Tests/Visual/Layout/TestSceneGridContainer.cs
+++ b/osu.Framework.Tests/Visual/Layout/TestSceneGridContainer.cs
@@ -652,6 +652,86 @@ namespace osu.Framework.Tests.Visual.Layout
             AddAssert("content spans grid size", () => Precision.AlmostEquals(grid.DrawWidth, grid.Content[0].Sum(d => d.DrawWidth)));
         }
 
+        [TestCase(true)]
+        [TestCase(false)]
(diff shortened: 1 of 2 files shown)


smoogipoo:
Does this already also work for lifetime? I.e. `Drawable.IsAlive`?

bdach:
It didn't, but it should now.

PR #3228, opened 2020-02-01
Add BindTarget to IBindable and implementations

### LINKED ISSUE #3212: BindableList<T> does not have BindTarget
It is provided for convenience by `Bindable<T>`:

https://github.com/ppy/osu-framework/blob/200907be11d839cad7e245ff55b18b329d92386d/osu.Framework/Bindables/Bindable.cs#L175

### LINKED ISSUE #3213: IBindable doesn't support BindTarget
It's only provided by `Bindable<T>`, but could be provided by `IBindable<T>`.

https://github.com/ppy/osu-framework/blob/200907be11d839cad7e245ff55b18b329d92386d/osu.Framework/Bindables/Bindable.cs#L175

LastExceed:
quick and easy.

* Closes #3212.
* Closes #3213.

# Breaking changes

## `IBindable.BindTarget` has been added

`BindTarget` has been added as a convenience setter-only property, which is intended to be an alias for `BindTo` in object initializer scenarios.

### DIFF
--- a/osu.Framework.Tests/Bindables/BindableListTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableListTest.cs
@@ -49,6 +49,30 @@ namespace osu.Framework.Tests.Bindables
 
         #endregion
 
+        #region BindTarget
+
+        /// <summary>
+        /// Tests binding via the various <see cref="BindableList{T}.BindTarget"/> methods.
+        /// </summary>
+        [Test]
+        public void TestBindViaBindTarget()
+        {
+            BindableList<int> parentBindable = new BindableList<int>();
+
+            BindableList<int> bindable1 = new BindableList<int>();
+            IBindableList<int> bindable2 = new BindableList<int>();
+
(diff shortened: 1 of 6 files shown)


peppy:
We may want to add this to the interface

peppy:
Added interface implementations. @smoogipoo please confirm the direction of cast I've used is correct for the underlying interface implementations.

Probably should add test coverage for this before merging.

LastExceed:
now also closes #3213

bdach:
As an aside, the interface addition will technically be a breaking change, unless we decide to provide a default implementation in the interface leveraging C# 8.0's default member implementations. I'm unconvinced if we should however, since default members have non-trivial gotchas and maybe it'd be better to make potential consumers wary of this change.

smoogipoo:
Yeah these are the perfect candidate for default implementations, will make the changes myself.

peppy:
I originally did this as default interfaces but it felt pretty weird. Either or I guess.

Issue #3234, opened 2020-02-03
SetUpSteps / TearDownSteps potentially run multiple times


peppy:
```csharp
public class DerivedTest : BaseTest
{
    [SetUpSteps] // <- causes a double execution.
    public override void SetUpSteps() => base.SetUpSteps();
}

public class BaseTest : TestScene
{
    [SetUpSteps]
    public virtual void SetUpSteps() => AddStep("setup step", () => {});
}

```

PR #3235, opened 2020-02-03
Fix test methods potentially being duplicated

smoogipoo:
Fixes https://github.com/ppy/osu-framework/issues/3234

Uses internal nunit method to determine the methods to invoke, which is also used by their own code:

https://github.com/nunit/nunit/blob/c7df8ecb5fe8cbcbe26eb2c34e906452bedb77a4/src/NUnitFramework/framework/Internal/Tests/TestFixture.cs#L43-L46

### DIFF
--- a/osu.Framework.Tests/Visual/Testing/TestSceneDerivedTest.cs
+++ b/osu.Framework.Tests/Visual/Testing/TestSceneDerivedTest.cs
@@ -1,17 +1,20 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using NUnit.Framework;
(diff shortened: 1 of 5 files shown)


osu.Framework.Tests/Visual/Testing/TestSceneDerivedTest.cs
```diff
@@ -1,17 +1,20 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using NUnit.Framework;
+using osu.Framework.Testing;
+
 namespace osu.Framework.Tests.Visual.Testing
 {
     public class TestSceneDerivedTest : TestSceneTest
```
peppy: probably want to keep this test as well

PR #3236, opened 2020-02-03
Add missing thread safety to ScheduledDelegate

peppy:
Fixes a rare cross-thread failure seen at https://ci.appveyor.com/project/peppy/osu/builds/30531069/tests.


osu.Framework/Threading/ScheduledDelegate.cs
```diff
@@ -0,0 +1,142 @@
...
+                    throw new InvalidOperationException($"Can not run a {nameof(ScheduledDelegate)} that has been already {nameof(Completed)}.");
+
+                RunTaskInternal();
+            }
+        }
+
+        /// <summary>
+        /// Invokes the scheduled task without throwing on incorrect state.
+        /// </summary>
+        internal void RunTaskInternal()
```
UselessToucan: I suggest to make this method `private` or `protected`

osu.Framework/Threading/Scheduler.cs
```diff
@@ -104,11 +104,8 @@ public virtual int Update()
 
             while (getNextTask(out ScheduledDelegate sd))
             {
-                if (!sd.Cancelled && !sd.Completed)
-                {
-                    //todo: error handling
-                    sd.RunTask();
-                }
+                //todo: error handling
+                sd.RunTaskInternal();
```
UselessToucan: Calling an `Internal` method from the outside looks wrong to me :(

peppy: i'm not sure i follow. what would you prefer here naming/structure wise?

osu.Framework/Threading/ScheduledDelegate.cs
peppy: it's intentionally internal so it can be called from `Scheduler`.

osu.Framework/Threading/Scheduler.cs
UselessToucan: I would prefer a `RunTask` call here.

peppy: i think you're missing the whole reasoning for this. it bypasses the exceptions.

iiSaLMaN: this seems to be changing behaviour, what if that task has been executed before `Scheduler.Update()` has been called.

peppy: please read the contents of the `RunTaskInternal` method.

iiSaLMaN: oh right, it early returns; seems OK then.

osu.Framework/Threading/ScheduledDelegate.cs
```diff
@@ -0,0 +1,142 @@
...
+
+                    if (ExecutionTime < currentTime && !PerformRepeatCatchUpExecutions)
+                        ExecutionTime = currentTime.Value + RepeatInterval;
+                }
+            }
+        }
+
+        public enum RunState
+        {
+            Waiting,
```
iiSaLMaN: `Pending`

smoogipoo: Add XMLDocs to all of these, even if they're simple.

PR #3238, opened 2020-02-04
Don't log spammy input events to debug log

peppy:
Fixes on-screen log spam that is mostly useless.

2020-02-04 at 18 37 15

### DIFF
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -566,6 +566,7 @@ namespace osu.Framework.Input
                     return !k.Repeat;
 
                 case DragEvent _:
+                case ScrollEvent _:
                 case MouseMoveEvent _:
                     return false;
 
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -549,13 +549,29 @@ namespace osu.Framework.Input
         {
             var handledBy = drawables.FirstOrDefault(target => target.TriggerEvent(e));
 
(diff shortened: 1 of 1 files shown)


osu.Framework/Input/InputManager.cs
```diff
@@ -558,6 +558,22 @@ protected virtual bool PropagateBlockableEvent(IEnumerable<Drawable> drawables,
...
         }
 
+        private bool shouldLog(UIEvent eventType)
+        {
+            switch (eventType)
+            {
+                case KeyDownEvent k:
+                    return !k.Repeat;
+
+                case DragEvent _:
```
smoogipoo: Allow at the very least `DragStartEvent`

osu.Framework/Input/InputManager.cs
```diff
@@ -558,6 +558,22 @@ protected virtual bool PropagateBlockableEvent(IEnumerable<Drawable> drawables,
...
+                case MouseMoveEvent _:
```
smoogipoo: What about `ScrollEvent`? Using any sort of smooth scrolling is guaranteed to blast you with logs.

osu.Framework/Input/InputManager.cs
peppy: DragStart doesn't derive DragEvent, so should be good

osu.Framework/Input/InputManager.cs
peppy: oops, missed this one when i inversed the function.

PR #3239, opened 2020-02-05
Improve template game

peppy:
The existing one was flawed in many ways and provided a shocking onboarding experience.

- Adds a `DrawSizePreservingFillContainer` and `GameBase` class
- Adds more boilerplate samples, including screen and screen stack
- Adds a sample showing sprite/texture loading
- Adds a resource project
- Adds inspectcode defaults


peppy:
Yes, this was agreed upon. Else you are presented with a solution that has errors (since the default code stylings don't match ours).

PR #3246, opened 2020-02-06
Check for cancellation during load components iteration

peppy:


osu.Framework/Graphics/Containers/CompositeDrawable.cs
```diff
@@ -202,10 +202,15 @@ protected internal Task LoadComponentAsync<TLoadable>(TLoadable component, Actio
             loadComponents(components, Dependencies, false);
         }
 
-        private void loadComponents<TLoadable>(IEnumerable<TLoadable> components, IReadOnlyDependencyContainer dependencies, bool isAsyncContext) where TLoadable : Drawable
+        private void loadComponents<TLoadable>(IEnumerable<TLoadable> components, IReadOnlyDependencyContainer dependencies, bool isDirectAsyncContext, CancellationToken? cancellation = null) where TLoadable : Drawable
```
smoogipoo: Can the cancellation token be non-nullable? In general they should never be null,

Issue #3245, opened 2020-02-06
ScrollContainer handles drag and scroll events when empty


smoogipoo:
There's no reason for it to do this, since nothing is displayed in the container.

PR #3248, opened 2020-02-06
Do not handle input in empty ScrollContainer

UselessToucan:
resolves https://github.com/ppy/osu-framework/issues/3245


osu.Framework/Graphics/Containers/Container.cs
```diff
@@ -50,6 +51,12 @@ public Container()
                 aliveInternalChildrenAsT = new LazyList<Drawable, T>(AliveInternalChildren, c => (T)c);
         }
 
+        internal override bool BuildNonPositionalInputQueue(List<Drawable> queue, bool allowBlocking = true) =>
+            AliveInternalChildren.Any() && base.BuildNonPositionalInputQueue(queue, allowBlocking);
+
+        internal override bool BuildPositionalInputQueue(Vector2 screenSpacePos, List<Drawable> queue) =>
+            AliveInternalChildren.Any() && base.BuildPositionalInputQueue(screenSpacePos, queue);
```
iiSaLMaN: I think you're looking for overriding `PropagateNonPositionalInputSubTree` and `PropagatePositionalInputSubTree` instead.

UselessToucan: Never used these methods. Will check them out tomorrow.

UselessToucan:
>     * Why not on `CompositeDrawable` instead?

Because the issue is about `ScrollContainer`. I decided that `CompositeDrawable` level is too wide for this case.

>     * This should probably have tests to validate this solution and not regress on the issue further on.

Indeed. I will add them tomorrow.

smoogipoo:
I disagree that this is what we want. It’s possible that we want composites to handle input when there’s no alive children. Something like moving around the screen (e.g. maybe a “blueprint” that overlays the drawable but needs to know about the drawable hover state, or the drawable itself is handling this), or perhaps adding a child on mouse/key events.

The issue was specific to ScrollContainer, not intended as a general issue.

peppy:
unsure if it should be done at the input queue building, as opposed to the event overrides, where we would usually do this kind of thing.

PR #3252, opened 2020-02-08
Bump bass libraries for macOS

peppy:
Apple started enforcing a minimum SDK version for compiled libraries to be notarized. Bumping bass.dylib to latest official release; bass_fx.dylib to version [available here](http://www.un4seen.com/forum/?topic=18517.0) (latest official release is not updated yet).

Don't merge yet, pending integration testing


peppy:
- Tested to be working as expected (runtime)
- Tested to silence notarization errors

PR #3254, opened 2020-02-11
Android pen hover support

voidedWarranties:
Mouse position events are sent while the pen is hovering.
Additionally, pressing the pen's side button while hovering will send a right click.

Closes ppy/osu#6401


smoogipoo:
Let's get this in and test on production, since none of us have a way of test this right now.

ojh050118:
I tested and it works very well. But there is a slight delay.

PR #3256, opened 2020-02-12
Apply correct current value on rewind between disparate transforms

peppy:
While this has not been noticed until now due to how transform methods like FadeInFromZero are implemented (they use two transforms), rewinding between transforms where the `EndValue` was different to the `StartValue` of the next transform could leave us in a bad state. Reverse playback would actually look different to forward playback.

Example (opposite fade direction to implemented test):

PNG image-A892C2269293-1

### DIFF
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
@@ -237,7 +237,7 @@ namespace osu.Framework.Tests.Visual.Drawables
                     box.FadeOutFromOne(interval);
                     box.Delay(interval * 3).FadeOutFromOne(interval);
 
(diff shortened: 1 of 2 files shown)


osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
```diff
@@ -219,6 +219,38 @@ public void StartInMiddleOfSequence()
...
+            checkAtTime(interval * 4, _ => true);
+
+            AddStep("add transforms", () =>
+            {
+                using (box.BeginAbsoluteSequence(0))
+                {
+                    box.FadeInFromZero(interval);
+                    box.Delay(interval * 3).FadeInFromZero(interval);
+
+                    box.RemoveTransform(box.Transforms[2]);
```
smoogipoo: Perhaps add a comment here to indicate that this isn't actually related to the failure case, but related to an implementation  detail of `FadeInFromZero`.

PR #3257, opened 2020-02-12
Fix adding a transform in the past from a future vantage point potentially getting incorrect StartValue

peppy:
- [x] Depends on https://github.com/ppy/osu-framework/pull/3256

PNG image-54D560E1EB1F-1

### DIFF
--- a/osu.Framework/Graphics/Transforms/Transformable.cs
+++ b/osu.Framework/Graphics/Transforms/Transformable.cs
@@ -467,7 +467,7 @@ namespace osu.Framework.Graphics.Transforms
             // If our newly added transform could have an immediate effect, then let's
             // make this effect happen immediately.
             if (transform.StartTime < Time.Current || transform.EndTime <= Time.Current)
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Transforms/Transformable.cs
```diff
@@ -462,7 +467,7 @@ public void AddTransform(Transform transform, ulong? customTransformID = null)
             // If our newly added transform could have an immediate effect, then let's
             // make this effect happen immediately.
             if (transform.StartTime < Time.Current || transform.EndTime <= Time.Current)
-                updateTransforms(Time.Current);
+                updateTransforms(Time.Current, transform.Rewindable && transform.EndTime <= Time.Current);
```
smoogipoo: Shouldn't this be `transform.StartTime <= Time.Current`?

Also `transform.Rewindable` has a different meaning to whether time is actually rewindable (`RemoveCompletedTransforms`).

PR #3262, opened 2020-02-15
Expose API for current display and display mode

swoolcock:
Introduces `Display` and `DisplayMode` classes that bridge the gap between osuTK's `DisplayDevice`/`DisplayResolution` classes and SDL2's display/display mode functions.

`IWindow.Display` and `IWindow.DisplayMode` properties have been introduced that represent the physical display containing the window, and resolution of that display respectively.  osuTK is implemented via the existing `CurrentDisplay` property, and SDL2 is implemented via native calls.

An `IWindow.Displays` property has been introduced that returns a snapshot enumerable of all currently available display devices, including their supported display modes.  osuTK is implemented using `DisplayDevice.AvailableResolutions`, and SDL2 is implemented via native calls.

Steps have been added to `TestSceneFullscreen` visual test to query and log the available displays on the system.  These steps will not be run on CI due to the lack of an available window.

This PR is a necessary step in bringing proper support for resolution selection in the SDL2 migration and will make it cleaner to manage window position and sizing as we aim for parity with osuTK.

This is not a breaking change, but eventually the existing `CurrentDisplay` property can be dropped from `IWindow` since it is osuTK-specific.


osu.Framework/Platform/DisplayMode.cs
```diff
@@ -0,0 +1,35 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Drawing;
+
+namespace osu.Framework.Platform
+{
+    /// <summary>
+    /// Represents a display mode on a given <see cref="Display"/>.
+    /// </summary>
+    public class DisplayMode
```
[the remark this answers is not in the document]
swoolcock: Was originally a struct but I changed to class.  Undecided as to what `Display` and `DisplayMode` should be.  If it's a struct it should probably be `readonly` too.

osu.Framework/Platform/Display.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// The available <see cref="DisplayMode"/>s on this display.
+        /// </summary>
+        public DisplayMode[] DisplayModes { get; }
+
+        /// <summary>
+        /// The zero-based index of the <see cref="Display"/>.
+        /// </summary>
+        public int Index { get; }
```
iiSaLMaN: What's the usage of this?

swoolcock: Used when setting the display from osuTK or SDL2.  Need to know the original index to set it.

PR #3264, opened 2020-02-16
Update inspectcode and fix new issues

peppy:


peppy:
Am I missing something? The inspections mentioned on CI look to be incorrect (for instance the `MethodInfo` one looks to still be able to return null). Is there a `NotNull` specification somewhere?

bdach:
The errors reported do seem bogus. All of them concern null checks after calling linq's `OrDefault` methods, and I'm seeing [other](https://youtrack.jetbrains.com/issue/RSRP-478172) [people](https://youtrack.jetbrains.com/issue/RSRP-478134) complain about it in JetBrains' issue tracker.

peppy:
I guess we wait this one out...

bdach:
Was a bit worried this wasn't going to be resolved until 2020.1, but apparently [a bug fix is planned](https://youtrack.jetbrains.com/issue/RSRP-478172#focus=streamItem-27-3948193.0-0).

PR #3265, opened 2020-02-16
Add test attempting to align adjacent sprites

peppy:
Continuing discussion on discord, this is probably something we want to show/be able to do. Currently not so easily possibly due to FP error and inability to perform alignement / snapping to screen pixels.

Current test scene looks like this, but I think we can do better:


peppy:
Closing in favour of issue

PR #3266, opened 2020-02-16
Implement touch input methods

iiSaLMaN:
Split from #2925 

Implements `IInput`s for applying to the touch state:
 - `TouchActivityInput`: an `IInput` that changes whether a provided touch source is active or not.
 - `TouchPositionInput`: an `IInput` that changes the position of a provided touch.


osu.Framework/Input/StateChanges/ButtonInput.cs
```diff
@@ -72,8 +72,21 @@ public void Apply(InputState state, IInputStateChangeHandler handler)
                 {
                     var buttonStateChange = CreateEvent(state, entry.Button, entry.IsPressed ? ButtonStateChangeKind.Pressed : ButtonStateChangeKind.Released);
                     handler.HandleInputStateChange(buttonStateChange);
+
+                    OnButtonStateChanged(state, entry.Button, entry.IsPressed);
```
smoogipoo: Isn't this occurring too late? As in if you query the `TouchPosition` at the point where the state is applied, you're going to get a non-null position even though it's released?

osu.Framework/Input/StateChanges/Events/TouchPositionChangeEvent.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// The touch source of this change event.
+        /// </summary>
+        public readonly MouseButton Source;
+
+        /// <summary>
+        /// The last position of <see cref="Touch"/>.
+        /// </summary>
+        public readonly Vector2 LastPosition;
+
+        public TouchPositionChangeEvent(InputState state, IInput input, MouseButton source, Vector2 lastPosition)
```
smoogipoo: Maybe accept the `Touch` struct?

osu.Framework/Input/StateChanges/Events/TouchPositionChangeEvent.cs
```diff
@@ -0,0 +1,29 @@
...
+            : base(state, input)
+        {
```
smoogipoo: Should throw exceptions, like the other places.

osu.Framework/Input/StateChanges/TouchActivityInput.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// <summary>
+    /// Denotes a change of the touch activity state (finger down, up).
+    /// Any provided touch source should always be in the range <see cref="MouseButton.Touch1"/>-<see cref="MouseButton.Touch10"/>.
+    /// </summary>
+    public class TouchActivityInput : ButtonInput<MouseButton>
+    {
+        public TouchActivityInput(IEnumerable<ButtonInputEntry<MouseButton>> entries)
+            : base(entries)
+        {
+            Trace.Assert(Entries.All(e => e.Button >= MouseButton.Touch1));
```
smoogipoo: Since this is public and a user can construct this, I'd like these asserts to be exceptions instead. They indicate programmer error rather than logical error.

osu.Framework/Input/StateChanges/ButtonInput.cs
iiSaLMaN: I was quite unsure where to place the event invocation (initially I came up with virtualizing `Apply()`). 

The current position of the touch passed to the `TouchEvent`s is retrieved with `state.Touch.GetTouchPosition(source)` right before the propagation happens as seen below:
https://github.com/ppy/osu-framework/blob/f3865fbabe62b79c6fcf6f42a5927a4b2cbde7e9/osu.Framework/Input/TouchEventManager.cs#L52-L53

osu.Framework/Input/StateChanges/TouchPositionInput.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// The touch source to be modified.
+        /// </summary>
+        public readonly MouseButton Source;
+
+        /// <summary>
+        /// The new position to move to.
+        /// </summary>
+        public readonly Vector2 Position;
+
+        public TouchPositionInput(MouseButton source, Vector2 newPosition)
```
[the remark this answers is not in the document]
iiSaLMaN: In another constructor, sure. That structure was originally needed in `TouchEvent`s.

osu.Framework/Input/StateChanges/Events/TouchPositionChangeEvent.cs
iiSaLMaN: that would only replace `MouseButton source`:
```csharp
public TouchPositionChangeEvent(InputState state, IInput input, Touch touch, Vector2 lastPosition)
{
    Source = touch.Source;
    LastPosition = lastPosition;
}

Comment (smoogipoo) [reply osu.Framework/Input/StateChanges/Events/TouchPositionChangeEvent.cs]: I'd like to see `Touch` become the wrapper for "A MouseButton between Touch1 and Touch10 combined with a position".

That is to say, this could have the signature:

```csharp
public TouchPositionChangeEvent(InputState state, IInput input, Touch lastTouch)
{
    Source = touch.Source;
    LastPosition = touch.Position;
}
```

And you could potentially remove the local `Source` and just store the touch directly (e.g. as `LastTouch`).

iiSaLMaN: If this is for ensuring that `Touch.Source` is in the touch range then that wouldn't change anything as it, unfortunately, can be `default`-initialized. Otherwise it looks pretty much redundant to me constructing `Touch` structures all the way to the events, may even rename it to `TouchArgs` as it only meant to ease up passing touch event arguments (`Source`, `Position`):
https://github.com/ppy/osu-framework/blob/f3865fbabe62b79c6fcf6f42a5927a4b2cbde7e9/osu.Framework/Input/Events/TouchEvent.cs#L46-L51

smoogipoo: It's more for consistency than enforcing that, because the concept of a `Touch` may change. Right now the concept of a touch is defined in two ways, either:
1. A `Touch`.
2. A `MouseButton` + `Vector2`.

And it should be one of them, probably just `Touch` is fine. Name it however you want but there should be some consistency with how it's used.

iiSaLMaN:
> 
> 
> Would like to see `TouchInputTest` headless test for this, otherwise looks good.

Alright, requires #3269 though so need to write the test in it.

swoolcock:
> > Would like to see `TouchInputTest` headless test for this, otherwise looks good.
> 
> Alright, requires #3269 though so need to write the test in it.

Ok, sounds good.

swoolcock:
@smoogipoo requesting re-review please when you have time, so we can start to get these touch PRs merged.

osu.Framework/Input/Touch.cs
```diff
@@ -0,0 +1,47 @@
...
+        public readonly MouseButton Source;
+
+        /// <summary>
+        /// The current position of this touch.
+        /// </summary>
+        public readonly Vector2 Position;
+
+        public Touch(MouseButton source, Vector2 position)
+        {
+            if (source < MouseButton.Touch1 || source > MouseButton.Touch10)
```
smoogipoo, line 27: I've been thinking about it for a while, but I don't like using `MouseButton` here, as it's going to 100% result in a breaking change when this is changed.

I suggest a separate `TouchSource` enum, with values matching the ones in `MouseButton`, which is to be passed into the `OnTouch` methods (in #3269).

osu.Framework/Input/StateChanges/ButtonInput.cs
smoogipoo: Does setting the touch position to null actually resolve any issue? I'm not sure we even need to consider that scenario. May make more sense to keep stale positions, and instead check whether the source is active if that is a concern.

osu.Framework/Input/StateChanges/Events/TouchActivityChangeEvent.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using osu.Framework.Input.States;
+using osuTK.Input;
+
+namespace osu.Framework.Input.StateChanges.Events
+{
+    public class TouchActivityChangeEvent : ButtonStateChangeEvent<MouseButton>
```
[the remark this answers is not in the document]
iiSaLMaN: Surprisingly, merging both `IInput`s made sense for everything, now the new `TouchStateChangeEvent` can be handled simply in `InputManager` by prioritizing mouse position input application before button input and removed the case of null touch position elements. 

although slightly different from your suggestion as the use of `ButtonsInput` became redundant for what the code looks like now.

smoogipoo:
I think this looks pretty good to me, but I'd like to see the dependant PRs updated before this is merged, as this is untestable by itself.

In the future, please avoid implementations without testability. I would've been fine with the big original PR.

iiSaLMaN:
Closing as packed up to one PR due to no possible way to test this at its separation: #3269

PR #3269, opened 2020-02-16
Implement touch input & handling support

frenzibyte:
Updated from #2925


osu.Framework.Tests/Input/TouchInputTest.cs
```diff
@@ -0,0 +1,301 @@
...
+                                         .Select(s => new InputReceptor(s)),
+            };
+        });
+
+        private static Vector2 getDownPos(InputReceptor r) => r.ScreenSpaceDrawQuad.TopLeft + Vector2.One; // +1 to avoid precision errors causing receptor not receiving input.
+        private static Vector2 getMovePos(InputReceptor r) => r.ScreenSpaceDrawQuad.Centre;
+        private static Vector2 getUpPos(InputReceptor r) => r.ScreenSpaceDrawQuad.BottomLeft + new Vector2(10f);
+
+        [Test]
+        public void TestTouchInputHandling()
```
iiSaLMaN, line 42: I'm a bit aware about the complexity in here but that's currently the best I can do.

osu.Framework/Input/InputManager.cs
```diff
@@ -398,7 +420,7 @@ private IEnumerable<Drawable> buildNonPositionalInputQueue()
 
         private readonly List<Drawable> positionalInputQueue = new List<Drawable>();
 
-        private IEnumerable<Drawable> buildPositionalInputQueue(InputState state)
+        private IEnumerable<Drawable> buildPositionalInputQueue(InputState state, Vector2? screenSpacePos = null)
```
peppy, line 423: the InputState is only being passed in to read the mouse. why? i'll change it to be sensible.

frenzibyte: ah, so you're suggesting to use the `CurrentState` now that the method is allowed to read a certain position if needed? yeah that would be better.

peppy: i'm suggesting just having Vector2 as a param. see pushed commits.

peppy:
How have you tested this PR? Through headless tests alone? Am I supposed to be able to use `TestSceneTouchVisualiser.cs ` somehow?

frenzibyte:
> How have you tested this PR? Through headless tests alone? Am I supposed to be able to use `TestSceneTouchVisualiser.cs ` somehow?

Headless tests for ensuring that events are invoked correctly and with right information, the touch visualiser test scene listens to touch input and visualises them with a circle moving to the finger position. 

And yeah forgot to mention that, you'll have to write a touch input handler for the device you're using this on in order to use the touch visualiser, I have an android one stashed waiting for this PR to be merged, can push here if needed.

peppy:
Please include as a gist at least.

osu.Framework.Tests/Input/TouchInputTest.cs
```diff
@@ -0,0 +1,301 @@
...
+            {
+                RelativeSizeAxes = Axes.Both,
+                Position = new Vector2(15f),
+                Spacing = new Vector2(15f),
+                ChildrenEnumerable = Enum.GetValues(typeof(TouchSource)).Cast<TouchSource>()
+                                         .Select(s => new InputReceptor(s)),
+            };
+        });
+
+        private static Vector2 getDownPos(InputReceptor r) => r.ScreenSpaceDrawQuad.TopLeft + Vector2.One; // +1 to avoid precision errors causing receptor not receiving input.
```
peppy, line 37: Pretty sure this is not correct test coverage. If this test was visual, you'd probably have noticed this, so I'd recommend making this non-headless as a starting point.

You're testing at line 65, what I presume, is that the receptor only gets events for the correct touch. But each receptor is at a unique screen position with no overlap, so this isn't actually testing the case where a touch arrives and may be routed to the wrong receptor.

The receptors should likely all be overlapping each other and take up the full size of the container.

osu.Framework.Tests/Input/TouchInputTest.cs
```diff
@@ -0,0 +1,301 @@
...
+        [Test]
+        public void TestMouseInputAppliedFromPrimaryTouch()
+        {
+            AddStep("activate touches", () =>
+            {
+                foreach (var r in receptors)
+                    InputManager.BeginTouch(new Touch(r.AssociatedSource, getDownPos(r)));
+            });
+
+            AddAssert("received correct mouse-down event", () =>
```
peppy, line 145: spilt this up into two asserts, one checking the Touch1 receptor specifically, and another checking the other receptors were not activated.

frenzibyte:
> Please include as a gist at least.

Sure, https://gist.github.com/frenzibyte/17b0185ddc31ca96720ea57639134955

osu.Framework.Tests/Input/TouchInputTest.cs
frenzibyte: Oh, and let each receptor only handle the touch source associated to them, that's a great idea!

osu.Framework.Tests/Input/TouchInputTest.cs
frenzibyte: Actually, unless I misunderstood, with the above suggestions, all receptors will be precisely stacked on top of each other, which means mouse events will propagate into all of them (unless one absorbs it), so this will technically be already resolved.

peppy: That should not be the case. You're returning `true` in handle.

peppy:
Just to check, are you planning on updating the tests for this PR?

frenzibyte:
Yes will get it today.

osu.Framework.Tests/Input/TouchInputTest.cs
frenzibyte: After confirmation that it is intended for a handling drawable's siblings to receive opposing events to the handler, checking if other receptors did not receive mouse events is unnecessary, because of:
```csharp
protected override bool Handle(UIEvent e)
{
    switch (e)
    {
        case MouseDownEvent _:
        case MouseMoveEvent _:
        case DragEvent _:
        case MouseUpEvent _:
            if (AssociatedSource == TouchSource.Touch1)
            {
                MouseEvents.Enqueue((MouseEvent)e);
                return true;
            }

        case DragStartEvent _:
             return AssociatedSource == TouchSource.Touch1;
    }

    return false;
}

Comment (peppy) [reply osu.Framework.Tests/Input/TouchInputTest.cs:145]: because of what? that code alone doesn't really much anything.

Comment (frenzibyte) [reply osu.Framework.Tests/Input/TouchInputTest.cs:145]: the receptor for `TouchSource.Touch1` will only be the one having mouse events enqueued, I'm gonna push my commits and let you decide.

Comment (peppy) [reply osu.Framework.Tests/Input/TouchInputTest.cs:145]: then what's the point of even having different receptors? why not just have one handle everything?

Comment (frenzibyte) [reply osu.Framework.Tests/Input/TouchInputTest.cs:145]: > then what's the point of even having different receptors? why not just have one handle everything?

OK right that's a good question, seems my code was written wrong from what I had in mind about what this test covers... 

I have separated each receptor to handle a separate touch source to test that the `TouchEventManager` is working properly for each source, but what I've written now (and what I believe is what you suggested) is to stack all receptors on top of each other, but that's incorrect since touch receptors in front of the currently tested one will also receive events that receptor is waiting for to enqueue.

I fixed all of this weirdness by moving each receptor to a specific X position, and each source is in front of the other in ascending order, so it'll look like this:
image

Now each touch event will be absorbed by the front-most receptor and the test will ensure that the front-most receptor is the expected one to receive that specific event.

Comment (peppy) [reply osu.Framework.Tests/Input/TouchInputTest.cs]: Looks more sensible. Making the visual tests easy to understand helps a lot when reviewing and diagnosing future issues.

**Comment by peppy** [comment]
In the updated test scenes, the mouse cursor seems to get an undefined position after the touch end events are fired. Is this expected? It doesn't visually look correct and doesn't look to be being tested, so feels quite undefined to me.

**Comment by frenzibyte** [comment]
> In the updated test scenes, the mouse cursor seems to get an undefined position after the touch end events are fired. Is this expected? It doesn't visually look correct and doesn't look to be being tested, so feels quite undefined to me.
> 
> 

this is actually intended, it is to ensure that the touch end event would still be received even if its outside the drawable's boundaries, which I believe is the correct behaviour (just like dragging on mouse).
image

maybe adding a parent drawable with the text "parent" at the bottom would help understanding this.

**Comment by peppy** [comment]
Make it a separate step then ("move touch outside of area"), and potential its own test.

**Comment by frenzibyte** [comment]
> Make it a separate step then ("move touch outside of area"), and potential its own test.

Done, also fixed a small unnoticed issue about move events potentially raised when no movement occurred.

File: osu.Framework/Input/InputManager.cs [hunk 5]
```diff
@@ -549,6 +590,18 @@ public virtual void HandleInputStateChange(InputStateChangeEvent inputStateChang
                     HandleKeyboardKeyStateChange(keyboardKeyStateChange);
                     return;
 
+                case TouchStateChangeEvent touchChange:
+                    HandleTouchStateChange(touchChange);
+
+                    // primary touch mapped to mouse.
+                    if (touchChange.Touch.Source == TouchSource.Touch1)
```
peppy, line 597: I'm 99% sure this is going to need to be exposed as a toggle (defaulting to on) on `InputManager`, else it's going to be very awkward to handle touches (they will fire two events) right?

In the osu! usage scenario, the gameplay (`RulsetInputManager`) would be disabling this and HitObjects / input overlays would need to handle mouse and touch as separate events.

peppy:
I think this is looking pretty good, so let's get this in as-is. We can add the mentioned toggle in a follow-up pull request (after we have touch handlers implemented for both supported mobile platforms) along with tests covering it.

frenzibyte:
> I think this is looking pretty good, so let's get this in as-is. We can add the mentioned toggle in a follow-up pull request (after we have touch handlers implemented for both supported mobile platforms) along with tests covering it.

Alright, I'll ready up the android touch handler.

osu.Framework/Input/InputManager.cs
frenzibyte: Yeah that's probably the reason osu! ruleset is not behaving correctly, will do it now.

PR #3270, opened 2020-02-16
Add touch event cases to PassThroughInputManager

frenzibyte:
Split from #2925 

- [x] Depends on #3269

Adds touch event cases to the `PassThroughInputManager` for propagating touch input to the hierarchy.


frenzibyte:
I believe the mouse button synchronization is done as-is (release stale but don't press new) for the input to not potentially be mistaken as a click, not the case for touches so decided it to follow the rest.

bdach:
While that makes sense to me (clicks are transient, discrete events, while a touch is persistent), it may be something  worth documenting in some way.

frenzibyte:
Yeah definitely, will push another branch for that.

Issue #3081, opened 2019-12-06
BindableList<T> events should provide the original list


smoogipoo:
When you need to run some post-processing on the list after an insertion or removal has occurred, there's currently no way to do this and retain access to the original list without storing an intermediate list, since if you access the bindable list during `ItemsRemoved` it will contain the post-mutated collection.

As for how to do this, there's a few things we have to consider:

1. Performance - raw performance + allocations. Do we want to incur the hit of allocating a new list with every event, and every propagation of said event?
2. The API - we probably want to return a custom class that we can adjust later on, similar to `ValueChangedEvent<T>`.

iiSaLMaN:
Regarding of perfomance, there is a way to have an `IEnumerable` with custom enumerator to represent a list before an item was removed or after an item is added.

I can go through implementing this if this enumerable idea has been accepted, though I'm not sure what to name it, maybe `MutatedList`?

peppy:
A better direction may be to standardise our events to follow .NET standards like https://docs.microsoft.com/en-au/dotnet/api/system.collections.specialized.notifycollectionchangedeventargs?view=netframework-4.8

smoogipoo:
Bumping this up in priority, since it'll help with solidifying the implementation of things like `RearrangeableListContainer`.

PR #3273, opened 2020-02-17
Add BindableList.CollectionChanged event implementing NotifyCollectionChangedEventHandler

smoogipoo:
Resolves https://github.com/ppy/osu-framework/issues/3081

Follows the model of https://source.dot.net/#System.ObjectModel/System/Collections/ObjectModel/ObservableCollection.cs .

We're going to try out providing the complete event handler with this to:
1. Implement the interface.
2. Follow other implementations such as realm-dotnet.
3. Look into how this affects us in the long term complexity wise. I discovered early on one potential advantage of it, which is as a potential performance improvement to remove a variable capture.

Tests cover both scenarios simultaneously, pending removal in 20200817.

Breaking changes:

# vNext

## `BindableList<T>.ItemsAdded` is now obsolete

The `ItemsAdded` event failed to provide enough context when items are moved around or inserted into the list.

It has been replaced by the `CollectionChanged` [NotifyCollectionChangedEventHandler](https://docs.microsoft.com/en-au/dotnet/api/system.collections.specialized.notifycollectionchangedeventhandler?view=netframework-4.8), which provides context such as the newly-added items and the indices at which the addition took place.

This event is triggered by the following methods with `Action = NotifyCollectionChangedAction.Add`:
```csharp
BindableList<T>.Add(T item)
BindableList<T>.Insert(int index, T item)
BindableList<T>.AddRange(IEnumerable<T> items)
```

The following is an example migration utilising the new `CollectionChanged` event:
```diff
BindableList<int> list = new BindableList<int>();

- list.ItemsAdded += items =>
- {
-     foreach (var item in items)
-         Console.WriteLine($"Added: {item}");
- }

+ list.CollectionChanged += (_, args) =>
+ {
+     switch (args.Action)
+     {
+         case NotifyCollectionChangedAction.Add:
+             foreach (var item in args.NewItems.Cast<int>())
+                 Console.WriteLine($"Added: {item});
+             break;
+     }
+ }
```

Note that `CollectionChanged` should _not_ be used in conjunction with the `ItemsAdded` event.

## `BindableList<T>.ItemsRemoved` is now obsolete

As above, the `ItemsRemoved` revent has also been replaced by the `CollectionChanged` [NotifyCollectionChangedEventHandler](https://docs.microsoft.com/en-au/dotnet/api/system.collections.specialized.notifycollectionchangedeventhandler?view=netframework-4.8).

This event is triggered by the following methods with `Action = NotifyCollectionChangedAction.Remove`:
```csharp
BindableList<T>.Clear()
BindableList<T>.Remove(T item)
BindableList<T>.RemoveRange(int index, int count)
BindableList<T>.RemoveAt(int index)
BindableList<T>.RemoveAll(Predicate<T> match)
```

The following is an example migration utilising the new `CollectionChanged` event:
```diff
BindableList<int> list = new BindableList<int>();

- list.ItemsRemoved += items =>
- {
-     foreach (var item in items)
-         Console.WriteLine($"Removed: {item}");
- }

+ list.CollectionChanged += (_, args) =>
+ {
+     switch (args.Action)
+     {
+         // Handles both clear and remove events
+         case NotifyCollectionChangedAction.Remove:
+             foreach (var item in args.OldItems.Cast<int>())
+                 Console.WriteLine($"Removed: {item});
+             break;
+     }
+ }
```

Note that `CollectionChanged` should _not_ be used in conjunction with the `ItemsRemoved` event.

### DIFF
--- a/osu.Framework/Bindables/BindableList.cs
+++ b/osu.Framework/Bindables/BindableList.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.Linq;
 using osu.Framework.Caching;
 using osu.Framework.Lists;
@@ -15,12 +16,19 @@ namespace osu.Framework.Bindables
         /// <summary>
         /// An event which is raised when any items are added to this <see cref="BindableList{T}"/>.
         /// </summary>
+        [Obsolete("Use CollectionChanged instead.")]
         public event Action<IEnumerable<T>> ItemsAdded;
 
         /// <summary>
         /// An event which is raised when any items are removed from this <see cref="BindableList{T}"/>.
         /// </summary>
+        [Obsolete("Use CollectionChanged instead.")]
         public event Action<IEnumerable<T>> ItemsRemoved;
+
+        /// <summary>
+        /// An event which is raised when this <see cref="BindableList{T}"/> changes.
+        /// </summary>
+        public event NotifyCollectionChangedEventHandler CollectionChanged;
 
         /// <summary>
         /// An event which is raised when <see cref="Disabled"/>'s state has changed (or manually via <see cref="triggerDisabledChange(bool)"/>).
@@ -77,8 +85,7 @@ namespace osu.Framework.Bindables
                 }
             }
 
-            ItemsRemoved?.Invoke(new[] { lastItem });
-            ItemsAdded?.Invoke(new[] { item });
+            notifyCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Replace, item, lastItem, index));
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


osu.Framework/Bindables/BindableList.cs
```diff
@@ -77,8 +85,7 @@ private void setIndex(int index, T item, BindableList<T> caller)
                 }
             }
 
-            ItemsRemoved?.Invoke(new[] { lastItem });
-            ItemsAdded?.Invoke(new[] { item });
+            notifyCollectionChanged(new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Replace, item, lastItem, index));
```
iiSaLMaN: to simplify lines a bit, how about passing arguments of the event parameters into the method parameters instead.
`notifyCollectionChanged(NotifyCollectionChangedAction.Replace, item, lastItem, index);`

smoogipoo: They’re all different signatures, though. Sure can have 4 overloads but I’m not sure it’s helping much at that poInt.

iiSaLMaN: Oh, didn't realize that. Nevermind me then.

smoogipoo:
PR description updated.

PR #3274, opened 2020-02-17
Add method to move items within a BindableList

smoogipoo:
Fixes https://github.com/ppy/osu/issues/7860

Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3273

Helps with implementing / solidifying the implementation of `RearrangeableListContainer`.


smoogipoo:
This also fixes the crash scenario in https://github.com/ppy/osu/issues/7860 (so I've marked the issue as being fixed), though there's still an osu-side issue regarding the ordering.

PR #3278, opened 2020-02-18
Reorganize Templates Project to Allow For Multiple Templates

TimOliver:
Greetings! 👋

I've been helping to improve osu!framework's onboarding and documentation processes.

As part of this, I've created a sample game powered by osu!framework, and we've agreed to make it a part of the .NET templates collection for osu!framework.

Up until now, osu!framework has only had one template (The basic spinning box template), and so before I can make a PR to merge my game code, we need to re-organize the layout of templates in this repo so that it can properly support multiple entries.

This PR moves all of the contents of `template-game` into a new folder called `osu.Framework.Templates` and reconfigures `osu.Framework.Templates.csproj` to represent all child template projects.

For any instances needed for  differentiation, the original template is now referred to 'empty', but still maintains the default `osu-framework-game` identifier.

Let me know if there's anything I need to change! Thanks! :)


peppy:
License header inspections are a fix at `CodeFileSanity`; will look into that.

template-game/.idea/.idea.TemplateGame.Desktop/.idea/.name
```diff
@@ -1 +0,0 @@
-TemplateGame.Desktop
```
peppy: don't think you wanted to nuke these files

PR #3283, opened 2020-02-18
Minor optimizations/cleanups

turbedi:
- Replace string.Substring(...).Trim() (or siimlar) calls with string.AsSpan(...).Trim().ToString()
- Replace remaining string.IndexOf(string, StringComparison.Ordinal) call with string.IndexOf(char)

No changes in behaviour.

### DIFF
--- a/osu.Framework/Configuration/IniConfigManager.cs
+++ b/osu.Framework/Configuration/IniConfigManager.cs
@@ -48,8 +48,8 @@ namespace osu.Framework.Configuration
 
                         if (line.Length == 0 || line[0] == '#' || equalsIndex < 0) continue;
 
-                        string key = line.Substring(0, equalsIndex).Trim();
-                        string val = line.Remove(0, equalsIndex + 1).Trim();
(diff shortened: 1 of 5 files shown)


osu.Framework/Input/Bindings/KeyBindingContainer.cs
```diff
@@ -279,8 +279,7 @@ public void TriggerPressed(T pressed)
 
         private List<Drawable> getInputQueue(KeyBinding binding, bool rebuildIfEmpty = false)
         {
-            if (!keyBindingQueues.ContainsKey(binding))
-                keyBindingQueues.Add(binding, new List<Drawable>());
+            keyBindingQueues.TryAdd(binding, new List<Drawable>());
```
smoogipoo: This constructs a new list every invocation, just to be thrown away.

turbedi: Sorry, I missed that. I have removed the related commit.

peppy:
Please don’t force push pull requests

PR #3289, opened 2020-02-20
Add TextBox event handlers

frenzibyte:
- [x] Depends on #3640 
- [x] Depends on #3641

Allows for adding custom logic when something happens to a `TextBox`, will come in handy for moving audio samples from the abstract `TextBox` to *osu!lazer's*.

Also has other changes:
- Changes `Commit()` return type to `bool` for indicating that a commit actually occurred. (to avoid potentially executing custom logic on `Commit()` 2 times when `ReleaseFocusOnCommit` and `CommitOnFocusLost` are both set)
- Adds a bonus feature: whether the caret is currently selecting text while moving, may be useful for playing different sounds based on what state the caret is in.


osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -553,6 +558,30 @@ private Drawable addCharacter(char c)
...
+        /// <param name="added">The inserted text string.</param>
+        protected virtual void OnTextAdd(string added)
+        {
+        }
+
+        /// <summary>
+        /// Invoked whenever a text string has been removed from <see cref="Text"/>.
+        /// </summary>
+        /// <param name="removed">The removed text string.</param>
+        protected virtual void OnTextRemove(string removed)
```
smoogipoo: `OnTextRemoved`

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -553,6 +558,30 @@ private Drawable addCharacter(char c)
...
+        {
+        }
+
+        /// <summary>
+        /// Invoked whenever the caret has moved from its position.
+        /// </summary>
+        /// <param name="selecting">Whether the caret is selecting text while moving.</param>
+        protected virtual void OnCaretMove(bool selecting)
```
smoogipoo: `OnCaretMoved`

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -914,16 +951,20 @@ private void onImeComposition(string s)
                 {
                     //if we are no longer matching, we want to remove all further characters.
                     removeCharacterOrSelection(false);
+                    deletedString += imeDrawables[matchCount].Item1;
```
bdach: I'm not sure how to actually get this code to trigger (attempted with Chinese IME), but this somewhat worries me. `+=` operations on strings mean an allocation every time. Could avoid it by having a `stringBuilder`.

iiSaLMaN:
> Should it even be valid to have a text box that has both `ReleaseFocusOnCommit` and `CommitOnFocusLost` set? That just sounds broken by design, I'm not sure why someone would want to shoot themselves in the foot with that gun.

out-of-scope for this PR but definitely something to open an issue for, in fact, some of the `TextBox` component logic are outdated and may need to be revised and refactored at some point.

bdach:
It is out of scope here but I'd argue it needs to be a prereq PR as the `bool Commit()` is a little strange in my opinion, and it'd probably end up being a simple exception throw on improper usage or something.

iiSaLMaN:
> It is out of scope here but I'd argue it needs to be a prereq PR as the `bool Commit()` is a little strange in my opinion

Why though? this could be useful to block committing on certain conditions with its sub-classes not required to duplicate the checks and just early-return if `false` is returned in the base method.

Might as well force that on all overridable functions.

bdach:
> this could be useful to block committing on certain conditions with its sub-classes not required to duplicate the checks and just early-return if `false` is returned in the base method.

I'd have to see a concrete convincing example that justifies this. Just the ability to do so doesn't convince me as it is added complexity, and I'm not sure that it's necessary.

iiSaLMaN:
> Should it even be valid to have a text box that has both `ReleaseFocusOnCommit` and `CommitOnFocusLost` set?

After rethinking this a bit, this should definitely be valid, this type of `TextBox` would commit current text when focus has been lost naturally (not by `ReleaseFocusOnCommit`) and release focus from the `TextBox` when committing naturally (not by `CommitOnFocusLost`).

maybe rewriting the check to commit on focus lost only if there is new committable text and remove the early-return check on `Commit()`.

bdach:
OK, I agree, if those behaviours are separate they do make more sense, but the double commit still does not. Still a matter for a separate PR, with proper tests and documentation.

iiSaLMaN:
> OK, I agree, if those behaviours are separate they do make more sense, but the double commit still does not. Still a matter for a separate PR, with proper tests and documentation.

I don't think tests are needed, it's a matter of code quality. I'm pretty sure documentation isn't even needed after separating the checks. (you'll see)

bdach:
The publicly-exposed flags steer behaviour the stability of which is important to framework consumers, therefore I'd say it needs to be covered with tests to avoid regressions (which, FWIW, goes for as much public framework API as possible).

frenzibyte:
@bdach I've resolved the double-commit fuzz with an internal `OnTextCommitted(...)` event handler for simplicity.

I've also merged `addCharacter()` to `InsertString()` and made `removeCharacterOrSelection()` accept removing more than one character for simplifying IME logic and centralize event invocations of text changes.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -167,7 +167,7 @@ public virtual bool OnPressed(PlatformAction action)
 
                     clipboard?.SetText(SelectedText);
                     if (action.ActionType == PlatformActionType.Cut)
-                        removeCharacterOrSelection();
+                        removeCharactersOrSelection();
```
peppy, line 170: specify a count of zero here for safety. or even split the function up into `removeCharacters` and `removeSelection` (calling a common function if required).

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -498,56 +503,85 @@ protected virtual Drawable AddCharacterToFlow(char c)
         protected float CalculatedTextSize => TextFlow.DrawSize.Y - (TextFlow.Padding.Top + TextFlow.Padding.Bottom);
 
         /// <summary>
-        /// Insert an arbitrary string into the text at the current position.
+        /// Insert an arbitrary string into this <see cref="TextBox"/> at the current position.
```
peppy, line 506:
```suggestion
        /// Insert an arbitrary string at the current position.
```

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -498,56 +503,85 @@ protected virtual Drawable AddCharacterToFlow(char c)
...
+        protected void InsertString(string value, Action<char, Drawable> onDrawableCreated = null)
         {
-            if (string.IsNullOrEmpty(text)) return;
+            if (string.IsNullOrEmpty(value)) return;
 
-            foreach (char c in text)
-            {
-                var ch = addCharacter(c);
+            if (Current.Disabled)
+                return;
```
peppy, line 515: hasn't this changed the behaviour? it used to notify of an error and doesn't now.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -412,25 +412,29 @@ private void moveSelection(int offset, bool expand)
             if (oldStart != selectionStart || oldEnd != selectionEnd)
             {
                 audio.Samples.Get(@"Keyboard/key-movement")?.Play();
+                OnCaretMoved(expand);
                 cursorAndLayout.Invalidate();
             }
         }
 
-        private bool removeCharacterOrSelection(bool sound = true)
```
peppy, line 419: i get that there's a follow-up PR, but this one alone will change behaviour right? now it's always playing the sound.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -904,38 +945,24 @@ private void onImeComposition(string s)
...
-
-                if (dr != null)
-                {
-                    dr.Colour = Color4.Aqua;
-                    dr.Alpha = 0.6f;
-                    imeDrawables.Add(dr);
-                }
-            }
-
-            audio.Samples.Get($@"Keyboard/key-press-{RNG.Next(1, 5)}")?.Play();
```
peppy, line 938: this is also removed, changing behaviour.

frenzibyte: Whoops right, was kinda hard to follow up with these behaviours...

peppy: Well this can be avoided. You're making many changes in this PR which are unrelated to your goal. Try to minimise the changes and propose unnecessary ones in a separate pass.

frenzibyte: Thing is from what I'm seeing, these changes won't make any sense without their purposes in (centralize textbox event calls and fix ime awkwardness), that's why I chose to include it in here, but will take this into consideration in future PRs. :+1:

peppy: Adding test coverage to assert you haven't missed edge cases (like you have) would be appropriate, as least.

osu.Framework/Graphics/UserInterface/TextBox.cs
frenzibyte: Actually forgot to pointed that out, yeah it'll always play the sound, but now the IME logic has been rewritten to play the sound once while it removes all of the non-matching characters, so no behaviour changes here.

osu.Framework/Graphics/UserInterface/TextBox.cs
frenzibyte: Right, I actually find it kinda weird that `ReadOnly` and `Current.Disabled` are different, will return back behaviour.

osu.Framework/Graphics/UserInterface/TextBox.cs
frenzibyte: See, that was the tough part of this, but now after rethinking this, a sane behaviour for this will be that `removeSelection()` will be a helper method to remove exactly the selected text, while `removeCharacters` will do the actual removing logic, will give that a shot.

frenzibyte: Done in here: https://github.com/ppy/osu-framework/pull/3289/commits/87967ff9c08ad00e95b8f8659aaa2eadccbe146f

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -498,56 +511,88 @@ protected virtual Drawable AddCharacterToFlow(char c)
...
-                ch.Show();
-            }
-        }
+                if (text.Length + 1 > LengthLimit)
+                {
+                    NotifyInputError();
+                    break;
+                }
 
-        private Drawable addCharacter(char c)
```
bdach, line 522: Was there a particular reason `addCharacter` was merged into `InsertString`? I think I preferred the old version, the new one is a bit more complex and involved.

If it's due to parameter passing then I'd maybe even make it a local function inside of `InsertString`.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -888,14 +941,10 @@ private void onImeComposition(string s)
             //search for unchanged characters..
             int matchCount = 0;
             bool matching = true;
-            bool didDelete = false;
 
             int searchStart = text.Length - imeDrawables.Count;
 
-            //we want to keep processing to the end of the longest string (the current displayed or the new composition).
-            int maxLength = Math.Max(imeDrawables.Count, s.Length);
```
bdach, line 896: Mind explaining this? This behaviour change (which I'm judging by the message of the commit that introduced it) seems to be some sort of fix for IME input, but if it is one then arguably it needs to be split out of here.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -412,27 +415,33 @@ private void moveSelection(int offset, bool expand)
...
+        /// Removes the selected text if a selection persists.
+        /// </summary>
+        private void removeSelection() => removeCharacters(selectionLength);
 
-            if (text.Length == 0) return false;
-            if (selectionLength == 0 && selectionLeft == 0) return false;
+        /// <summary>
+        /// Removes a specified amount of characters left side of the current position.
+        /// </summary>
+        private void removeCharacters(int amount = 1)
```
bdach, line 431: nit: it's technically `number` (amount is uncountable)

frenzibyte:
> This PR overall is really difficult in reviewing. There are changes all over and it's hard for me to confidently arrive at the conclusion that this won't break things.

OK I will split all changes into separate PRs, although I put them here just for them to make sense where they belong, but sure if it helps reviewing.

osu.Framework/Graphics/UserInterface/TextBox.cs
frenzibyte: I've removed this because the loop doesn't delete characters anymore, I simply: 
1. check what's the matching count between the new composed string `s` and what's shown in the textbox `text`, 
2. if the matching count is less than the `imeDrawables` count, then there are stale substring of characters we need to remove.
3. if the new composed string is greater than the matching count, then there are new characters we need to add.

frenzibyte:
Will leave this unchanged until the prerequisites are merged, then I'm gonna resolve conflicts.

peppy:
Could do with some conflict resolution now.

peppy:
Would have been nice to have some very simple test coverage of these events, but I guess it's fine as-is.

PR #3290, opened 2020-02-20
Remove all audio-related logic from TextBox

frenzibyte:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3289

Will be moved to *osu!lazer*.


osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -416,13 +411,12 @@ private void moveSelection(int offset, bool expand)
...
             if (oldStart != selectionStart || oldEnd != selectionEnd)
             {
-                audio.Samples.Get(@"Keyboard/key-movement")?.Play();
                 OnCaretMove(expand);
                 cursorAndLayout.Invalidate();
             }
         }
 
-        private bool removeCharacterOrSelection(bool sound = true)
+        private bool removeCharacterOrSelection(bool invokeEvent = true)
```
bdach: Again, I don't know how to get the code path calling this with `false` to run so I don't know what the precise case is (aside from being IME-related), but it seems a little weird to suppress text box events for a potential consumer.

Also, while we're touching this, might as well get rid of the unused return value maybe?

peppy:
Added breaking change note for this.

PR #1842, opened 2018-09-01
Allow controlling behaviour of expensive_operations_queue

default0:
This adds an int-property to the GameHost that allows configuring how many
expensive operations (defined by the expensive_operations_queue of GLWrapper)
are allowed per frame.

This came up when checking memory usage of a preloader in my game.
The way my preloader works is by loading all textures the game needs at once, so that no loading lag happens during the actual gameplay. I found that due to the way the texture uploads get scheduled, the game would take up ~1 GB of RAM for quite a while, until all the uploads are flushed, which due to the nature of the preloader, usually happens around twenty to thirty seconds after the preloader is finished. And even then, memory usage only goes down when the GC decides to do a full collection.

Changing the behaviour of the expensive_operations_queue so it aggressively finished all its operations while the preloader runs (and then resetting it back to its normal mode of operation after the preloader is done) reduces memory usage from ~1GB to 200 MB in my case, and I don't think my use-case of messing with the way texture uploads are processed during a loading screen is all that rare of a pattern.

PR #3291, opened 2020-02-20
Improve texture queueing mechanism

peppy:
Until now, the mechanism we had in place for background texture uploading had a few issues:

- It would only upload one item per frame
- This wouldn't be so bad, but it would also only process one item, even if it was already uploaded

This was particularly bad when the `FrameStatisticsDisplay` was visible, as it would queue four texture uploads per frame (even though it was only performing one actual upload), causing a runaway scenario where the queue would never recover.

This PR makes the queue aware of the texture's upload and queued state to ensure the queue stays under control. It also allows more than one texture to be uploaded per frame if the queue grows too large.

Should hopefully resolve the issues experienced on https://github.com/ppy/osu-framework/pull/1842/files (I think this PR may be able to be closed as a result).


smoogipoo:
I've refactored `GLWrapper` a bit - changed naming, added one more stat, split out complex condition.

Feel free to merge if you're fine with my changes.

PR #3292, opened 2020-02-21
Restore video texture buffer count to 3 for now

peppy:
In testing this seems fine. Setting back to sane value pending further feedback from @Pizzabelly


smoogipoo:
@Pizzabelly can you provide some insight as to why this was raised to 9? The reasoning was not properly understood at the point of merging :/

Pizzabelly:
Oh, I changed that when trying to figure out the issue of old frames being shown again. I didn't change it back after I fixed that issue, but I think its good now to change it back to 3.

PR #3294, opened 2020-02-21
Improve video decoder error messages

bdach:
# Summary

Add delegate for calling into ffmpeg's `av_strerror` function to retrieve human-readable errors instead of error codes and call it within exception messages instead of directly embedding the numerical code
itself.

# Remarks

While ffmpeg's error messages can sometimes be sort of obtuse (i.e. `Invalid data found when processing input`), they did help me with diagnosing the issues with decoding on linux more than the random error code number.

Tested on Linux and Windows.

### DIFF
--- a/osu.Framework/Graphics/Video/VideoDecoder.cs
+++ b/osu.Framework/Graphics/Video/VideoDecoder.cs
@@ -549,6 +549,7 @@ namespace osu.Framework.Graphics.Video
                 av_frame_unref = AGffmpeg.av_frame_unref,
                 av_frame_get_buffer = AGffmpeg.av_frame_get_buffer,
                 av_strdup = AGffmpeg.av_strdup,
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Video/VideoDecoder.cs
```diff
@@ -546,7 +566,8 @@ protected virtual FFmpegFuncs CreateFuncs()
                 avio_alloc_context = AGffmpeg.avio_alloc_context,
                 sws_freeContext = AGffmpeg.sws_freeContext,
                 sws_getContext = AGffmpeg.sws_getContext,
-                sws_scale = AGffmpeg.sws_scale
+                sws_scale = AGffmpeg.sws_scale,
+                av_strerror = AGffmpeg.av_strerror
```
iiSaLMaN: nitpick but this should be within the `av_` functions above.

PR #3296, opened 2020-02-22
Add update CPU time profiling to global statistics

peppy:
Generally we rely on existing profilers for tracking down performance issues, but as profilers do not show local class names, we end up having to work with this kind of mess:

dotTraceView64_2020-02-23_02-43-23

This solves the issue and gives a quick overview of what components are using the most CPU during their update operations.

image

- Updates at most once every second for readability purposes.
- Aims to be lowest overhead possible. Split the update loop into two versions to avoid the function call overhead when disabled (it does make a difference).

### DIFF
--- a/osu.Framework/Statistics/TypePerformanceMonitor.cs
+++ b/osu.Framework/Statistics/TypePerformanceMonitor.cs
@@ -14,7 +14,7 @@ namespace osu.Framework.Statistics
 
         private static readonly Stack<Type> current_collection_type_stack = new Stack<Type>();
 
-        private static readonly Dictionary<Type, double> collected_times = new Dictionary<Type, double>();
+        private static readonly Dictionary<Type, Collection> collected_times = new Dictionary<Type, Collection>();
 
         private static double consumptionTime;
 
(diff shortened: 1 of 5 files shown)


osu.Framework/Statistics/TypePerformanceMonitor.cs
```diff
@@ -0,0 +1,128 @@
...
+            current_collection_type_stack.Clear();
+            consumeStopwatchElapsedTime();
+
+            if (framesSinceLastReport > 0 && clock.CurrentTime - lastReport > 1000)
+            {
+                GlobalStatistics.Clear(group_name);
+
+                int i = 0;
+
+                foreach (var t in collected_times.OrderByDescending(t => t.Value.TotalMilliseconds).Take(5))
```
bdach: Just as a side note, you can deconstruct `KeyValuePair`s like value tuples and do i.e.
```csharp
foreach (var (type, collectedTime) in collected_times.OrderByDescending(t => t.Value.TotalMilliseconds).Take(5))
```

peppy: i'm aware of this, but still unsure if i like it as a pattern (unless the source is also a tuple)

PR #3301, opened 2020-02-24
Add ability to toggle between single and multi-threaded execution

peppy:
Toggle with Ctrl+Alt+F7

Note that the execution rate becomes dependent on the window loop at this point, which is generally locked to whatever the WM or OS is providing us with. Our local threads still attempt to keep their relative execution rates, meaning there is potential for sub-optimal frame times depending on frame limiter setting (using unlimited or vsync is recommended for consistency).

**This is intended only for profiling/benchmarking purposes.** It is a barebones implementation with missing safeties (for example, the thread checks are bypassed, rather than checking for input thread). I mainly just want this on appveyor for easier testability.

That said, I won't decline reviews or discussion on the direction of implementation.

Road forward:
- Test performance / heat / energy metrics on mobile devices
- Short term: if a huge improvement is seen on mobile, consider hardcoding enabled (and remove toggle logic if it seen as too complex)
- Compare performance on single core desktop (for fair comparisons I intend on creating a BenchmarkDotNet game loop for this)

Update:

- Initial benchmarks show a 60-70% reduction in CPU usage (iPhone SE / SampleGame).
- Blocking audio operations will cause a hard freeze due to waiting on a threaded operation that will never be reached.

.

- [x] Depends on https://github.com/ppy/osu-framework/pull/3313
- [x] Depends on https://github.com/ppy/osu-framework/pull/3314
- [x] Depends on https://github.com/ppy/osu-framework/pull/3317

### DIFF
--- a/osu.Framework/Platform/ThreadRunner.cs
+++ b/osu.Framework/Platform/ThreadRunner.cs
@@ -105,13 +105,7 @@ namespace osu.Framework.Platform
                     lock (threads)
                     {
                         foreach (var t in threads)
-                        {
-                            // required as we are constantly changing thread contexts while remaining on the same logical thread
-                            ThreadSafety.ResetAllForCurrentThread();
-
-                            t.MakeCurrent();
                             t.ProcessFrame();
-                        }
                     }
 
                     break;
--- a/osu.Framework/Platform/ThreadRunner.cs
+++ b/osu.Framework/Platform/ThreadRunner.cs
@@ -8,7 +8,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-using System.Threading;
 using osu.Framework.Development;
 using osu.Framework.Extensions.IEnumerableExtensions;
 using osu.Framework.Logging;
@@ -146,7 +145,9 @@ namespace osu.Framework.Platform
             {
                 case ExecutionMode.MultiThreaded:
                 {
-                    ThreadSafety.SingleThreadThread = null;
+                    ThreadSafety.IsAudioThread = false;
+                    ThreadSafety.IsDrawThread = false;
+                    ThreadSafety.IsInputThread = false;
 
                     // switch to multi-threaded
                     foreach (var t in Threads)
@@ -160,8 +161,6 @@ namespace osu.Framework.Platform
 
                 case ExecutionMode.SingleThreaded:
                 {
-                    ThreadSafety.SingleThreadThread = Thread.CurrentThread;
-
                     // switch to single-threaded.
                     foreach (var t in Threads)
                         t.Pause();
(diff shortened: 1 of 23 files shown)


peppy:
This is ready for review. Now dependent on https://github.com/ppy/osu-framework/pull/3313.

osu.Framework/Platform/ThreadRunner.cs
```diff
@@ -0,0 +1,205 @@
...
+                    break;
+                }
+
+                case ExecutionMode.SingleThread:
+                {
+                    // switch to single-threaded.
+                    foreach (var t in Threads)
+                    {
+                        t.MakeCurrent();
+
```
smoogipoo:
```suggestion
```

osu.Framework/Threading/UpdateThread.cs
```diff
@@ -5,21 +5,35 @@
...
-        public UpdateThread(Action onNewFrame)
+        private readonly DrawThread drawThread;
+
+        public UpdateThread(Action onNewFrame, DrawThread drawThread)
             : base(onNewFrame, "Update")
         {
+            this.drawThread = drawThread;
+        }
+
+        protected override void OnInitialize()
```
smoogipoo:
```suggestion
        protected sealed override void OnInitialize()
```

osu.Framework/Threading/UpdateThread.cs
```diff
@@ -5,21 +5,35 @@
...
+            if (ThreadSafety.ExecutionMode != ExecutionMode.SingleThread)
+            {
+                //this was added due to the dependency on GLWrapper.MaxTextureSize begin initialised.
+                drawThread?.WaitUntilInitialized();
+            }
         }
 
         public override bool IsCurrent => ThreadSafety.IsUpdateThread;
 
         internal override void MakeCurrent()
```
smoogipoo:
```suggestion
        internal sealed override void MakeCurrent()
```

osu.Framework/Threading/DrawThread.cs
```diff
@@ -5,24 +5,54 @@
...
+        public DrawThread(Action onNewFrame, GameHost host)
             : base(onNewFrame, "Draw")
         {
+            this.host = host;
         }
 
         public override bool IsCurrent => ThreadSafety.IsDrawThread;
 
-        internal override void MakeCurrent()
+        protected override void OnInitialize()
```
smoogipoo:
```suggestion
        protected sealed override void OnInitialize()
```

osu.Framework/Threading/DrawThread.cs
```diff
@@ -5,24 +5,54 @@
...
+        }
+
+        internal sealed override void MakeCurrent()
         {
             base.MakeCurrent();
+
             ThreadSafety.IsDrawThread = true;
         }
 
+        protected override void Cleanup()
```
smoogipoo:
```suggestion
        protected sealed override void Cleanup()
```

osu.Framework/Threading/AudioThread.cs
```diff
@@ -23,6 +23,7 @@ public AudioThread()
         internal override void MakeCurrent()
```
smoogipoo:
```suggestion
        internal sealed override void MakeCurrent()
```

PR #3302, opened 2020-02-26
Expose some ScrollContainer members as public

aemino:
Exposing `ScrollContainer.scrollableExtent` is a prerequisite for ppy/osu#7407. I added a few other properties that seemed reasonable to expose, but if I took too heavy-handed of an approach here, I can reduce the changes to just what's necessary.


osu.Framework/Graphics/Containers/ScrollContainer.cs
```diff
@@ -120,25 +120,25 @@ public bool ScrollbarOverlapsContent
         /// </summary>
         private float target;
 
-        private float scrollableExtent => Math.Max(availableContent - displayableContent, 0);
+        public float ScrollableExtent => Math.Max(AvailableContent - DisplayableContent, 0);
```
peppy: I don't think exposing these are a problem, but if we are to do so they need xmldoc.

aemino: I drafted summaries for the two undocumented properties that this PR exposes, but I wasn't quite sure what to write, so feedback will be appreciated.

Issue #223, opened 2016-10-24
Invalidation does not always propagate to children when it needs to.


Tom94:
This is best explained by an example scenario.

`Drawable drawable;`
`Container container = new Container { Children = new[] { drawable = new Drawable(); } };`
`container.Invalidate(Invalidation.All);`
`DoStuff(drawable.ScreenSpaceDrawQuad);`
`container.Invalidate(Invalidation.Colour);`

Now future invalidations of `container` will never propagate to `drawable` because everything within `container` is already invalidated, even though `drawable.ScreenSpaceDrawQuad` is still valid.

I fixed this particular scenario in https://github.com/ppy/osu-framework/pull/220/commits/fd135a3b01e69c134838d93f891eab039fef5ca6 by making `Invalidation.Color` invalidate both `DrawInfo` and `ScreenSpaceDrawQuad` but the fix is specific and does not address this class of problem in general.

- Fixes https://github.com/ppy/osu/issues/7273.

peppy:
@Tom94 is this still an issue?

Tom94:
Yup, it still is. Note, that currently this is not a correctness problem, but rather makes invalidations slightly more inefficient than they could be. More specifically: if something like `Color` is invalidated, then the entire `DrawInfo` gets invalidated; potentially for the subtree if it depends on the `DrawInfo` of its parent.

peppy:
Just for documentation, there are a few special cases coded in to handle the fallout effects of this:

https://github.com/ppy/osu-framework/pull/1416/commits/baba2699d52b5ff5919f7dea29b2a7a186d6a75d#diff-74fa55675af99d2ab8d5ed0f726caf41R332

and

https://github.com/ppy/osu-framework/blob/fa6b8c6a172cc4b3645fb9b07f2227e8135fdbe4/osu.Framework/Graphics/Containers/FlowContainer.cs#L90

ekrctb:
An another case is illustrated in #1803 `FillFlow2` and `FillFlow2Simplified`.

ekrctb:
I think this invariant should be maintained:
- If a `Cached` (Say `A`) is invalid then other `Cached` (Say `B`) which depends on `A` should be invalid.
```csharp
Parent:
  Cached<TypeA> Abacking;
  TypeA A => Abacking.IsValid ? Abacking.Value : (Abacking.Value = ComputeA());
  TypeA ComputeA() => ...;
Child:
  Cached<TypeB> Bbacking;
[...]

PR #3307, opened 2020-02-26
Rework invalidations game-wide to fix subtle layout issues

smoogipoo:
Fixes https://github.com/ppy/osu-framework/issues/223
Fixes https://github.com/ppy/osu/issues/7273
Fixes https://github.com/ppy/osu/issues/3798
Also fixes invalidation issues mentioned in https://github.com/ppy/osu/pull/7707
Also fixes any other invalidation issues game-wide (changelog builds not reflowing on ui scale changes, user panels in social overlay not reflowing on ui scale changes, etc...)

Sorry for the large changes. These are unavoidable since it's such a radical departure from what was there previously. I've tried to keep user-side changes quite minimal (see: https://github.com/ppy/osu/pull/8006), with potential for iterating on this further in the future (described below).

## Explanation

There are two important concepts that must be maintained:
1. A drawable that is invalidated should not re-invalidate with the same flags.
2. The validation of a drawable must propagate all the way to the root.

These can both be optimised: for an invalidation we only need to propagate when the drawable hasn't already been invalidated with the same flags. Similarly for validation, we can stop propagating when drawables have already been optimised.

Tho achieve all of this along with the optimisation mentioned, it is imperative that:
1. `Drawable` store a list of triggered invalidation flags. When an invalidation request is received, it first checks to see if the flags have previously been triggered, and if not continues along with the invalidation.
2. Layout is explicitly defined.

(2) is done through the new `LayoutValue` and `LayoutValue<T>` classes.

`LayoutValue` is similar to `Cached` in that it can stores some value (whether an invalidation state or a provided `T`). It can also be invalidated and validated similar to `Cached` whenever the user determines this is required.

However additionally, `LayoutValue` responds to a specific set of user-determined `Invalidation` flags. This is done automatically whenever `Drawable.Invalidate()` is invoked, and replaces the previous manual invalidation of `Cached` values inside `Invalidate()`.

It also responds to validation (either manually via `LayoutValue.Validate()` or via setting `LayoutValue<T>.Value`). This is important as it allows us to propagate the validation request up towards the root, dismissing any invalidation flags that were previously set.

## Run-time complexity

Both `Invalidate()` and `ValidateSuperTree()` run in O(n) time:
1. On invalidation, only flags that were previously invalidated propagate forward. So for a `Drawable` that has both its `Position` and `Scale` changed in one frame (both affecting `MiscGeometry`), only the first of the two will propagate through the hierarchy (along with invalidating the local layout).
2. On invalidation, only the **immediate parent** of the invalidated `Drawable` is also invalidated. This deals with cases such as auto-size, where the parent needs to know about invalidations occurring on the child, however it is expected that changes in the parent propagate upwards only when necessary beyond this point.
3. On validation, only the flags that are still in an invalidated state propagate upwards. So for a position that has both `Position` and `Scale` validated, only the first of the two will propagate upwards. Other children of the same parenting container will likewise not propagate their related changes since the parenting composite is marked as valid in the given invalidation flags.

As pathological test cases, I have tested the following scenarios:
1. Storyboard: https://osu.ppy.sh/beatmapsets/470977#osu/1006822
    * Performance is on-par with current master.
2. Gameplay: https://osu.ppy.sh/beatmapsets/150945#osu/372245
    * Performance is ~10ms worse than current master in the most extreme areas, but is generally on-par with master.
3. Hiding/showing the toolbar in song select
    * Performance is slightly (~0.7ms) better than master.
4. Editor playback: https://osu.ppy.sh/beatmapsets/332532#osu/736215
    * Performance is ~3ms worse than current master.

Any tradeoffs here should mostly be unnoticeable, but in the cases they are it's to achieve correctness. Further optimisations can be done on the layout system, some are mentioned below.

## Future goals

* Removing various invalidation flags from layout values that don't require them. The `Presence` and `Parent` invalidations were mostly added as hacks previously, and may now not be required anymore due to per-flag invalidations. For example `BufferedContainer.screenSpaceSizeBacking`.
* Consider automating the parameters of `LayoutValue` via attributes. I tried this but it's quite a lot of work and performance has to be considered. It would help provide safety against both forgetting to invoke `AddLayout()` and invoking it too late (for which an exception is thrown).
* Consider removal of `OnInvalidate()` in some fashion. It's currently pretty unsafe in the context of `CompositeDrawable`, where the propagation can be stopped if a malicious consumer decides to do so.
    * One of the things I worked on was a [`LayoutDelegate`](https://github.com/ppy/osu-framework/compare/master...smoogipoo:layout-delegate) concept, which would make this safe. Micro-benchmark performance needs careful consideration however, and delegates prove to be about 4x slower than virtual methods.
* With the removal of `OnInvalidate()`, consider de-recursifying invalidation. I've tried to de-recursify `ValidateSuperTree` and it actually makes it perform worse (implying the JIT does some tail-call optimisations), so this will need micro-benchmarks.

In general I think the layout API is a little bit fragile as it is, but it's very seldom used by a framework consumer that I think breaking it further down the road to improve the API isn't too big of an issue.  
Other issues can be addressed via improvements such as `Drawable` pooling.

# vNext

## Layout validation and invalidation has been reworked

To better cover edge-case scenarios where layout wasn't properly refreshed, the process of layout invalidation has changed.

### In cases where `Invalidate()` affected a `Cached` member, the following adjustment is necessary:

```diff
public class MyDrawable : Drawable
{
-    private readonly Cached myLayoutValue = new Cached();
-
-    protected override bool OnInvalidate(Invalidation invalidation, Drawable source, bool shallPropagate)
-    {
-        var result = base.OnInvalidate(invalidation, source, shallPropagate);
-
-        if ((invalidation & (Invalidation.DrawSize | Invalidation.Presence)) > 0) /* any invalidation type of your choice */
-            result &= !myLayoutValue.Invalidate();
-
-        return result;
-    }

+    /* can also be a LayoutValue<T> */
+    private readonly LayoutValue myLayoutValue = new LayoutValue(Invalidation.DrawSize | Invalidation.Presence);
+
+    public MyDrawable()
+    {
+        AddLayout(myLayoutValue);
+    }
}
```

### In cases where `InvalidateFromChild()` affected a `Cached` member, the following adjustment is necessary:

```diff
public class MyDrawable : Drawable
{
-    private readonly Cached myLayoutValue = new Cached();
-
-    protected override void InvalidateFromChild(Invalidation invalidation, Drawable source)
-    {
-        if ((invalidation & (Invalidation.DrawSize | Invalidation.Presence)) > 0) /* any invalidation type of your choice */
-            myLayoutValue.Invalidate();
-    }

+    /* can also be a LayoutValue<T> */
+    private readonly LayoutValue myLayoutValue = new LayoutValue(Invalidation.DrawSize | Invalidation.Presence, InvalidationSource.Child);
+
+    public MyDrawable()
+    {
+        AddLayout(myLayoutValue);
+    }
}
```

### In cases where _both_ `Invalidate()` and `InvalidateFromChild()` affected the same `Cached` member, the following adjustment is required:

```diff
public class MyDrawable : Drawable
{
-    private readonly Cached myLayoutValue = new Cached();
-
-    protected override bool OnInvalidate(Invalidation invalidation, Drawable source, bool shallPropagate)
-    {
-        var result = base.OnInvalidate(invalidation, source, shallPropagate);
-
-        if ((invalidation & Invalidation.DrawSize) > 0) /* any invalidation type of your choice */
-            result &= !myLayoutValue.Invalidate();
-
-        return result;
-    }
-
-    protected override void InvalidateFromChild(Invalidation invalidation, Drawable source)
-    {
-        if ((invalidation & Invalidation.Presence) > 0) /* any invalidation type of your choice */
-            myLayoutValue.Invalidate();
-    }

+    /* can also be a LayoutValue<T> */
+    private readonly LayoutValue myLocalLayoutValue = new LayoutValue(Invalidation.DrawSize);
+    private readonly LayoutValue myChildLayoutValue = new LayoutValue(Invalidation.Presence, InvalidationSource.Child);
+
+    public MyDrawable()
+    {
+        AddLayout(myLocalLayoutValue);
+        AddLayout(myChildLayoutValue);
+    }
+
+    protected override void Update()
+    {
+        base.Update();
+
+        /* wherever your re-validation was previously done */
+        if (!myChildLayoutValue.IsValid)
+        {
+            myLocalLayoutValue.Invalidate();
+            myChildLayoutValue.Validate();
+        }
+
+        if (!myLocalLayoutValue.IsValid)
+        {
+            /* your custom validation */
+            myLocalLayoutValue.Validate();
+        }
+    }
}
```

### In cases where custom logic that can't be described as layout was done in `Invalidate()`, the following adjustment is necessary:

```diff
public class MyDrawable : Drawable
{
-    protected override bool OnInvalidate(Invalidation invalidation, Drawable source, bool shallPropagate)
-    {
-        var result = base.OnInvalidate(invalidation, source, shallPropagate);
-
-        if ((invalidation & (Invalidation.DrawSize | Invalidation.Presence)) > 0) /* any invalidation type of your choice */
-        {
-            performCustomAction();
-            result = true;
-        }
-
-        return result;
-    }

+    protected override bool OnInvalidate(Invalidation invalidation, InvalidationSource source)
+    {
+        var result = base.OnInvalidate(invalidation, source);
+
+        if ((invalidation & (Invalidation.DrawSize | Invalidation.Presence)) > 0) /* any invalidation type of your choice */
+        {
+            performCustomAction();
+            result = true;
+        }
+
+        return result;
+    }

    private void performCustomAction()
    {
        /* custom invalidation logic here */
    }
}
```


smoogipoo:
cc @Tom94 hoping you can take a quick look through this if you ever have time.

Also @bdach would like your pass over this too if possible!

osu.Framework/Graphics/Containers/CompositeDrawable.cs
```diff
@@ -750,6 +750,8 @@ protected void MakeChildAlive(Drawable child)
             aliveInternalChildren.Add(child);
             child.IsAlive = true;
 
+            Invalidate(Invalidation.Presence, InvalidationSource.Child);
```
smoogipoo: This also needs to happen in make dead.

bdach:
I'd be willing to give this a look, but as this looks like a PR that I'll need to get some context for and do some mental processing in order to provide a meaningful review, it might take me a day or two - so if this gets enough LGTMs before that time then go ahead with the merge.

bdach:
I did a first pass over this today, mostly to review the overall problem statement and look at the overarching architecture of the new invalidation system. I think I got up to speed on the terms and the framework laid down seems pretty good so far. I've been taking my own notes along the way as well, so I may be able to contribute some of that back as documentation if it's salvageable.

One nit that seems not quite clear to me with the invalidation flags is how does `MiscGeometry` differ from `DrawInfo`. I assume `DrawInfo` is the superset (since `All` does not currently include `MiscGeometry`), but it may become clearer when I see the usage patterns.

I've also had a cursory look at the overarching idea for the (in)validation algorithm you laid down, and the description at least seems good. I also like how the framework consumer change templates look so far, but this is not so much a definite judgement, as a first impression. Next pass I'm planning for is to look at the actual propagation algorithm & usage patterns, as well as the possible pain points outlined in the PR description.

smoogipoo:
`MiscGeometry` is included into `All` via `ParentSizeToFit`.

`DrawInfo` is a very legacy thing. Once upon a time it was basically telling us to invalidate `Drawable.drawInfoBacking` because _something_ happened in the hierarchy that required re-computation of it. For example, a change in `Margin` of a parent doesn't affect a child's position nor size, but the child's `DrawInfo` still needs to be recomputed to update the matrices.

These days `Margin` does, in fact, invalidate `MiscGeometry`, but that didn't always used to be the case. `DrawInfo` could very well be removed, as it's pretty much guaranteed that every usage has at least one other relevant flag, or is implemented incorrectly and causing unnecessary invalidations.

For example, `RelativeChildOffset` and `RelativeChildSize` are implemented incorrectly because they cause an unnecessary implicit operation of `DrawInfo` through `InvalidationFromParentSize` even when `DrawInfo` doesn't actually need to be re-computed, for children in completely absolute coordinate space.

bdach:
Ah, apparently I can't read now, that's great. Thanks for the correction.

I guess `DrawInfo` is also something to revisit/remove in the future then too.

smoogipoo:
There’s one case that I missed, which is that children never actually see `DrawSize` / `MiscGeometry` invalidations from their parent unless their size has actually changed, because `Invalidate()` coalesces those two flags into `DrawInfo`

This is actually an optimisation in the current code but really serves no purpose for us here since we can respond to `DrawSize` from the parent source, and may in-fact result in anti-optimisations for consumers if they’re not careful. There’s quite a lot of optimisations that can be done here, which is why the API is not final. The current code maintains the coalescing of flags, but I foresee responding to multiple invalidation sources as becoming a more common occurrence. So feel free to propose adjustments to `LayoutValue` :D

bdach:
So I think I'm through the second pass of this. It's a slow process, maybe because I'm actually looking at code this far down framework for the first time (either that, or I'm just a little dim). I'm learning quite a bit along the way, though, so maybe not all's lost?

This time I was looking at the fundamental structure in `Drawable` & co., both by looking at code and having a peek at a simple example in the debugger, and it looks to be matching the PR description exactly. I see no fault with the arguments made about the invariants of the tree traversal while invalidating and re-validating, or with the runtime complexity argument.

I guess one small remark I'd make is that upwards invalidations should be discouraged/avoided as much as possible, since as far as I understand even though they're still linear, they'll be less performant since they go against the overall flow of the code (invalidations propagating downwards, re-validations propagating upwards), so there'll be some redundant checks and early-returns, which might size up to a non-negligible constant factor.

The core ideas all seem very sound. I think the third and final pass I'd like to do tomorrow is to review the updated usages one-by-one, to see all the corner-cases we have so far.

As for adjustments for `LayoutValue` - I'm not really sure I can provide any meaningful feedback, since I don't have much experience with designing frameworks so it's hard to me to come up with possible generalisations that would meet the needs of possible consumers. Maybe something will arise when I look through the remaining parts, but for now I'd honestly just take this PR as-is and iterate on this base to eliminate the warts (the invalidation flag set being a little arcane is probably the most important one if you ask me).

osu.Framework/Graphics/Drawable.cs
```diff
@@ -1668,67 +1687,120 @@ private Vector2 computeRequiredParentSizeToFit()
...
 
-            if ((invalidation & Invalidation.Parent) > 0)
-                alreadyInvalidated = false;
-
-            if (!alreadyInvalidated || (invalidation & Invalidation.DrawNode) > 0)
+            if (anyInvalidated)
                 InvalidationID = invalidation_counter.Increment();
 
-            OnInvalidate?.Invoke(this);
+            Invalidated?.Invoke(this);
```
bdach: This is the only nit I couldn't find any good justification for; is there a reason this is raised without checking `anyInvalidated`? As far as I can tell the only consumer of this event in framework & lazer is the draw visualiser, and for that case it might be a smidge misleading to have this fire even if nothing was actually invalidated. Maybe we want to add info whether invalidation was or wasn't performed in the delegate?

peppy: I think it's fine to leave it this way for now. This event is `internal` and only ever intended to be used for visualisation purposes. While it's probably safe to move inside the `anyInvalidated` check, I'm fine with it staying where it is as that's how we had it until now.

Worth case it'll just run `FadeOutFromOne` more than once unnecessarily.

peppy:
I think this is in a really good state for now. Further iterations on code quality can happen post-merge – let's get this in and start (hopefully not) finding remaining issues with invalidation.

PR #3313, opened 2020-02-27
Calculate elapsed time values using FrameStatistics source instead of clock

peppy:
Until now, we were reading `ElapsedFrameTime` from threads' clock, but only 10 times per second. This means it could be inaccurate if frame times were different between these polls.

- Now considers every frame
- Uses `FrameStatistics.CollectedTimes` instead of `ElapsedFrameTime`. This is a prerequisite for making the elapsed time display actually work with single threaded execution. In practice, this means a small framework overhead may be excluded from the calculation, but if anything this makes the display more correct in its context.


smoogipoo:
After showing and hiding the display:

image

smoogipoo:
After showing and hiding the display:

image

PR #3314, opened 2020-02-27
Move frame sync toggle to platform action

peppy:
Unsure why it was not, but it meant it didn't work on iOS etc.


swoolcock:
LGTM, I'm glad that lingering TODO is now gone.

Issue #3308, opened 2020-02-27
There should be an option to prevent screen suspension


smoogipoo:
It may be desirable to prevent the screen from suspending. For example, if a replay is being watched or if a long break is entered.

Should be provided at a GameHost level as a `BindableBool`, which is referenced in both iOS and Android projects.

TimOliver:
Merged. :)

peppy:
If you use `Closes #issuenumber` in your PRs it will automatically close the issues, for what it's worth.

Also this function isn't yet implemented osu!-side, we may want to look at that.

PR #3315, opened 2020-02-27
Add game host setting for disabling mobile system idle timers

TimOliver:
Addresses https://github.com/ppy/osu-framework/issues/3308 in adding a new property to `GameHost` that allows for explicitly disabling any system-level idle timers temporarily. 

The use-case behind this is to set it at times when the user might not interact with the game for an extended period of time (eg, watching a game replay etc).

## Changes

1. Add `IsIdleTimerEnabled` to `GameHost` class.
2. Added iOS system integration in `IOSGameHost`
3. Added Android system integration in `AndroidGameActivity`.

## Remaining Tasks

* I'm still unfamiliar with the code/feature best practices, so these may need refining.
* ~It should be relatively easy to unit test these too, but I might need some help on how to add tests.~


osu.Framework/Platform/GameHost.cs
```diff
@@ -60,6 +60,13 @@ public abstract class GameHost : IIpcHost, IDisposable
         /// </summary>
         public readonly IBindable<bool> IsActive = new Bindable<bool>(true);
 
+        /// <summary>
+        /// For cases where the user isn't interacting with the game (eg, watching replays),
+        /// disable any system level timers that might dim or turn off the screen.
+        /// For preserving battery life on mobile devices, this should be left on as much as possible.
+        /// </summary>
+        public readonly IBindable<bool> IsIdleTimerEnabled = new Bindable<bool>(true);
```
iiSaLMaN: Name seems to be a bit long and too technical, how about `AllowScreenSuspension` or similar?

osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -53,6 +54,8 @@ protected override void OnCreate(Bundle savedInstanceState)
                     UIVisibilityFlags = systemUiFlags;
                 }
             };
+
+            gameView.Host.IsIdleTimerEnabled.ValueChanged += val => handleIdleTimerEnabledChange(val.NewValue);
```
peppy: this won't correctly work if the value is set before this point. Use `BindaValueChanged(del, true)` (see documentation for `true` parameter).

osu.Framework.iOS/IOSGameHost.cs
```diff
@@ -31,8 +32,11 @@ public class IOSGameHost : GameHost
         public IOSGameHost(IOSGameView gameView)
         {
             this.gameView = gameView;
+
             NSNotificationCenter.DefaultCenter.AddObserver(UIKeyboard.WillShowNotification, handleKeyboardNotification);
             NSNotificationCenter.DefaultCenter.AddObserver(UIKeyboard.DidHideNotification, handleKeyboardNotification);
+
+            IsIdleTimerEnabled.ValueChanged += val => handleIdleTimerEnabledChange(val.NewValue);
```
peppy: as above

osu.Framework.Android/AndroidGameActivity.cs
peppy: also is there a reason this is in `AndroidGameActivity` specifically?

osu.Framework.iOS/IOSGameHost.cs
```diff
@@ -59,6 +63,17 @@ private void handleKeyboardNotification(NSNotification notification)
...
 
+        /// <summary>
+        /// When <see cref="GameHost.IsIdleTimerEnabled"/> changes values, the value
+        /// is forwarded to the appropriate UIKit APIs to control
+        /// the timer that is responsible for dimming/locking the screen after
+        /// a certain amount of inactivity.
+        /// </summary>
+        private void handleIdleTimerEnabledChange(bool isEnabled)
+        {
+            UIApplication.SharedApplication.IdleTimerDisabled = !isEnabled;
```
peppy: either inline this or change the method signature to match the `ValueChanged` spec to avoid the second delegate construction above (`val => ...`)

osu.Framework/Platform/GameHost.cs
peppy: agree

osu.Framework/Platform/GameHost.cs
```diff
@@ -60,6 +60,13 @@ public abstract class GameHost : IIpcHost, IDisposable
         /// </summary>
         public readonly IBindable<bool> IsActive = new Bindable<bool>(true);
 
+        /// <summary>
+        /// For cases where the user isn't interacting with the game (eg, watching replays),
+        /// disable any system level timers that might dim or turn off the screen.
+        /// For preserving battery life on mobile devices, this should be left on as much as possible.
+        /// </summary>
```
peppy:
```suggestion
        /// <summary>
        /// Disable any system level timers that might dim or turn off the screen.
        /// </summary>
        /// <remarks>
        /// To preserve battery life on mobile devices, this should be left on whenever possible.
        /// </remarks>
```

osu.Framework.Android/AndroidGameActivity.cs
TimOliver: It's in Android game activity because apparently the disable screen flag is a property of the `Window` object the activity is hosting, which seems to be completely separate from the Game Host object.

peppy:
Keep in mind this actually needs to be used at osu!'s end to be useful.

osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -53,6 +55,14 @@ protected override void OnCreate(Bundle savedInstanceState)
                     UIVisibilityFlags = systemUiFlags;
                 }
             };
+
+            gameView.Host.AllowScreenSuspension.BindValueChanged(allow =>
+            {
+                if (allow.NewValue)
```
peppy: i get a strong feeling this may also need to be main thread, like iOS. will test on android on monday, should hold off merging until that's checked (else this could cause a hard crash on startup).

bdach: It does cause a hard crash on startup due to `gameView.Host` not being initialised at this point yet. Exposing a `HostStarted` event at `AndroidGameView`, subscribing to it and binding only then resolves the crash:
```csharp
            gameView.HostStarted += host =>
            {
                host.AllowScreenSuspension.BindValueChanged(allow =>
                {
                    if (allow.NewValue)
                        Window.AddFlags(WindowManagerFlags.KeepScreenOn);
                    else
                        Window.ClearFlags(WindowManagerFlags.KeepScreenOn);
                });
            };
```
No other alterations seem to be necessary & the toggle seems to be working fine.

TimOliver: Awesome! Thanks so much for that @bdach!

I've just integrated those changes, and while I was able to confirm they build, I haven't got any way to test Android builds in my current setup. If you get a chance, please have a look at my code and tell me if that looks okay.

Issue #3312, opened 2020-02-27
Revisit iOS input polling


swoolcock:
Following on from https://github.com/ppy/osu/issues/8008

The original OpenTK implementation for iOS locks the `Update` and `Render` events at the same rate.  It's up to the developer as to whether they want to use `CADisplayLink` by calling `RunWithFrameInterval(int frameInterval)` or to use an `NSTimer` by calling `Run(double updatesPerSecond)`.  A `CADisplayLink` will attempt to synchronise with a fraction of the display refresh rate, whereas an `NSTimer` can run at whichever rate the developer specifies.

We currently use `public void Run() => RunWithFrameInterval(1);`, which will attempt to call the `RunIteration` function at the display's refresh rate.  This (previously) would call `Update` and `Render` exactly once per loop.

Since we use our own draw thread, I added the `ShouldCallOnRender` and `ShouldCallOnUpdate` properties to avoid the extra `glBindFramebuffer` that was causing flickering and crashes.

https://github.com/ppy/osuTK/blob/master/src/osuTK.iOS/iOSGameView.cs#L929

On iOS we currently only use the `Update` event to queue pending inputs.  Since we're not using the `Render` event at all, I don't see any reason to keep it as a `CADisplayLink` when we could request a specific rate for input polling.

I suggest changing `gameView.Run();` to `gameView.Run(120);` in `GameAppDelegate` to use 120hz as a baseline for input polling as it will cover the current maximum sample report rate for iOS devices.

https://github.com/ppy/osu-framework/blob/master/osu.Framework.iOS/GameAppDelegate.cs#L33

peppy:
@swoolcock see https://github.com/ppy/osu-framework/pull/3301, which also requires changes here. that PR wants the "input" thread (ie. main window thread) to run at the refresh rate, at very least.

Currently it doesn't, and seems to be locked to 60hz even on an ipad pro.

peppy:
Closing this for now as we are now in a pretty good state (via #3316).

PR #3316, opened 2020-02-28
Attached Screen FPS to game update loop on iOS

TimOliver:
As discussed in https://github.com/ppy/osu-framework/issues/3312, here's my branch/PR ~in progress~ for making this work better on high-FPS devices. It looks like previously, the game view was relying on an osuTK API that set up an `NSTimer` under the hood, but this one instead uses `CADisplayLink`.

`UIScreen` actually exposes the number of FPS the device is capable of, which is what `CADisplayLink` is then configured against.

### DIFF
--- a/osu.Framework.iOS/GameAppDelegate.cs
+++ b/osu.Framework.iOS/GameAppDelegate.cs
@@ -30,7 +30,7 @@ namespace osu.Framework.iOS
             Window.MakeKeyAndVisible();
 
             // required to trigger the osuTK update loop, which is used for input handling.
-            gameView.Run();
+            gameView.Run(UIScreen.MainScreen.MaximumFramesPerSecond);
 
             host.Run(CreateGame());
 
--- a/osu.Framework.iOS/GameAppDelegate.cs
+++ b/osu.Framework.iOS/GameAppDelegate.cs
@@ -30,7 +30,7 @@ namespace osu.Framework.iOS
             Window.MakeKeyAndVisible();
 
             // required to trigger the osuTK update loop, which is used for input handling.
-            gameView.Run(UIScreen.MainScreen.MaximumFramesPerSecond);
+            gameView.RunWithFrameInterval((int)UIScreen.MainScreen.MaximumFramesPerSecond);
 
             host.Run(CreateGame());
 


peppy:
`NSTimer` should not be used. We want `CADisplayLink`.

I'm actually not sure how this PR is working, since @TimOliver seemed to state the opposite of what you said above, but looking at the source your statements are correct.

The settings we are using for `CADisplayLink` seem to be correct and should be returning the full refresh rate, so further investigation is required.

peppy:
The real issue is likely a missing `CADisableMinimumFrameDuration`: https://developer.apple.com/library/archive/technotes/tn2460/_index.html#//apple_ref/doc/uid/DTS40017657

TimOliver:
Ohhh. I see. `Run()` is overloaded to take either an int or a double, and depending on that type, it will forward it to a `NSTimer` or to a `CADisplayLink` implementation. Okay, at the very least, I've explicitly pointed it to the `CADisplayLink` one and set the maximum number of frames.

My understanding is that `NSTimer` should never be used to drive a render queue. For performance reasons, the CPU can decide to coalesce timer events for battery reasons, which means they might not consistently line up with screen refreshes. Whereas `CADsiplayLink` is explicitly tied to the physical refresh rate of the display.

@peppy Hmm. Okay. I'll try adding `CADisableMinimumFrameDuration` as well. That was supposed to be a temporary stopgap back in iOS 10 times when Pro Motion was still new, but it's possible Apple still hasn't made it the default.

TimOliver:
Hmm. So apparently `CADisplayLink.frameInterval` was deprecated in iOS 10, so it's possible setting it in iOS 13 is just `NO-OP`ing, which would explain why me setting it to 120 didn't freeze the app. 😅

https://developer.apple.com/documentation/quartzcore/cadisplaylink/1621231-frameinterval

It might be worth updating the iOS osuTK code to stop using frame intervals, and optionally just set `preferredFramesPerSecond`.

osu.Framework.iOS/GameAppDelegate.cs
```diff
@@ -30,7 +30,7 @@ public override bool FinishedLaunching(UIApplication application, NSDictionary l
             Window.MakeKeyAndVisible();
 
             // required to trigger the osuTK update loop, which is used for input handling.
-            gameView.Run();
+            gameView.RunWithFrameInterval((int)UIScreen.MainScreen.MaximumFramesPerSecond);
```
peppy:
```suggestion
            gameView.RunWithFrameInterval(0); // parameter of zero uses device refresh rate.
```

Proposing as a safe alternative without potential `osuTK` changes.

TimOliver:
From Paul Hudson's [Hacking with Swift article about NSTimers](https://www.hackingwithswift.com/articles/117/the-ultimate-guide-to-timer)

> Synchronizing your timer with screen updates
> Some people, particularly those making games, try to use timers to have some work done before every frame is drawn – i.e., 60 or 120 frames per second, depending on your device.
> 
> This is a mistake: timers are not designed for that level of accuracy, and you have no way of knowing how much time has elapsed since the last frame was drawn. So, you might think you have 1/60th or 1/120th of a second to run your code, but in practice half of that might already have passed before your timer was triggered.
> 
> So, if you want to have some code run immediately after the previous display update, you should use CADisplayLink instead. I’ve written some example code for that already (see How to synchronize code to drawing using CADisplayLink), but here’s a quick snippet:
> 
> let displayLink = CADisplayLink(target: self, selector: #selector(fireTimer))
> displayLink.add(to: .current, forMode: .defaultRunLoopMode)
> As with firing timers, if you want your display link method to trigger even when the UI is currently being used, make sure you specify .commonModes rather than .defaultRunLoopMode.

TimOliver:
Alright. I was able to contain all of my changes to the osuTK library, so this PR isn't needed anymore.

Here's the link to the osuTK PR: https://github.com/ppy/osuTK/pull/61

PR #3332, opened 2020-03-04
Extend color extension

andy840119:
Split from #3138 

Added : 
- `Color4 FromHex(string hex)` : Copied from osu's [OsuColour](https://github.com/ppy/osu/blob/master/osu.Game/Graphics/OsuColour.cs)
- `string ToHex(this Color4 color, bool forceOutputAlpha = false)`
- `Color4 ToRGB(float h, float s, float v)`
- `ToHsv(this Color4 color, out float h, out float s, out float v)`

### DIFF
--- a/osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
+++ b/osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
@@ -152,14 +152,17 @@ namespace osu.Framework.Extensions.Color4Extensions
         }
 
         /// <summary>
-        /// Convert HSV to Color4
+        /// Convert HSV to <see cref="Color4"/>
         /// </summary>
         /// <param name="h">Hue value, between 0 to 360</param>
         /// <param name="s">Saturation value, between 0 to 1</param>
         /// <param name="v">Value value, between 0 to 1</param>
         /// <returns></returns>
-        public static Color4 ToRGB(float h, float s, float v)
+        public static Color4 FromHSV(float h, float s, float v)
         {
+            if (h < 0 || h > 360)
+                throw new ArgumentOutOfRangeException(nameof(h), "Hue must be between 0 and 360.");
+
             int hi = ((int)(h / 60.0f)) % 6;
             float f = h / 60.0f - (int)(h / 60.0);
             float p = v * (1 - s);
@@ -185,10 +188,10 @@ namespace osu.Framework.Extensions.Color4Extensions
 
                 case 5:
                     return toColor4(v, p, q);
-            }
 
-            // Should not goes to here
-            throw new ArgumentOutOfRangeException($"{nameof(hi)} is not valid.");
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(h), "Hue is out of range.");
+            }
 
             static Color4 toColor4(float fr, float fg, float fb)
             {
--- a/osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
+++ b/osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
@@ -3,6 +3,7 @@
 
 using osuTK.Graphics;
 using System;
+using System.Globalization;
 
 namespace osu.Framework.Extensions.Color4Extensions
 {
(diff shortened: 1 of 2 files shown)


osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
```diff
@@ -96,5 +97,139 @@ public static Color4 Multiply(this Color4 colour, float scalar)
...
+            {
+                byte r = (byte)Math.Clamp(fr * 255, 0, 255);
+                byte g = (byte)Math.Clamp(fg * 255, 0, 255);
+                byte b = (byte)Math.Clamp(fb * 255, 0, 255);
+                return new Color4(r, g, b, 255);
+            }
+        }
+
+        /// <summary>
+        /// Convert color4 to HSV
```
smoogipoo:
```suggestion
        /// Convert <see cref="Color4"/> to HSV.
```

osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
```diff
@@ -96,5 +97,139 @@ public static Color4 Multiply(this Color4 colour, float scalar)
...
+        /// </summary>
+        /// <param name="color">Color4</param>
```
smoogipoo:
```suggestion
        /// <param name="colour">The <see cref="Color4"/> to convert.</param>
```

osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
```diff
@@ -96,5 +97,139 @@ public static Color4 Multiply(this Color4 colour, float scalar)
...
+        /// <param name="h">Hue value, between 0 to 360</param>
+        /// <param name="s">Saturation value, between 0 to 1</param>
+        /// <param name="v">Value value, between 0 to 1</param>
+        public static void ToHsv(this Color4 color, out float h, out float s, out float v)
```
smoogipoo: As mentioned in the original PR, I would also prefer that this returned a named tuple, since there may be situations where a user doesn't care about using all the returned values.

`(float h, float s, float v)`

smoogipoo: This should also be called `ToHSV` to match the other methods. This'll require a change to the .DotSettings (see: https://github.com/ppy/osu-framework/blob/master/osu-framework.sln.DotSettings#L318)

osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
```diff
@@ -96,5 +97,139 @@ public static Color4 Multiply(this Color4 colour, float scalar)
...
+
+                case 4:
+                    return toColor4(t, p, v);
+
+                case 5:
+                    return toColor4(v, p, q);
+            }
+
+            // Should not goes to here
+            throw new ArgumentOutOfRangeException($"{nameof(hi)} is not valid.");
```
smoogipoo: Change this to an exception at the start of the method, something like:

```cs
if (h < 0 || h > 360) throw new ArgumentOutOfRangeException(nameof(h), "Hue must be between 0 and 360.");
```

osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
```diff
@@ -96,5 +97,139 @@ public static Color4 Multiply(this Color4 colour, float scalar)
...
+        }
+
+        /// <summary>
+        /// Convert HSV to Color4
+        /// </summary>
+        /// <param name="h">Hue value, between 0 to 360</param>
+        /// <param name="s">Saturation value, between 0 to 1</param>
+        /// <param name="v">Value value, between 0 to 1</param>
+        /// <returns></returns>
+        public static Color4 ToRGB(float h, float s, float v)
```
smoogipoo: I'm imagining a future where we'll have HSL too, or HSP, other methods of generating colours. For that reason, this may be better suited as `FromHSV()` instead.

Also fits with `FromHex()`.

osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
```diff
@@ -96,5 +97,139 @@ public static Color4 Multiply(this Color4 colour, float scalar)
...
+            }
+        }
+
+        public static string ToHex(this Color4 color, bool forceOutputAlpha = false)
+        {
+            var argb = color.ToArgb();
+            byte a = (byte)(argb >> 24);
+            byte r = (byte)(argb >> 16);
+            byte g = (byte)(argb >> 8);
+            byte b = (byte)argb;
```
smoogipoo: As mentioned in the original PR, would like this to be changed to referencing the Color4 values directly.

osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
andy840119: Oops, sorry

osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
andy840119: Like `byte a = (byte)(color.ToArgb()>> 24);` ?

smoogipoo: Ah, I see the issue. The components are floats, not bytes.

Fine to leave as is.

osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
andy840119: image
Found duplicated setting.
Should it be removed in this pr?

smoogipoo: I see no harm in doing so.

smoogipoo:
Made some quick refactorings.

@peppy quick pass pls.

PR #3337, opened 2020-03-05
Explicitly define type interpolations via interfaces

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3336

Interested on people's thoughts on this. It's not really required but it may ease implementation overhead, especially as this API is changed. Future changes will add an override for a general easing function.

It's a little bit hacky in that it's creating an uninitialised type and then calling a method on it, but maybe it's not too bad?

This is also one of those functions that is unlikely for consumers to have used just yet, since we have a pretty wide array of transforms/interpolations already.
In the case where a consumer did use it, they'll be provided with the helpful exception - "implement this interface on this type" where the interface has the correct function definition.

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkInterpolation.cs
+++ b/osu.Framework.Benchmarks/BenchmarkInterpolation.cs
@@ -1,0 +1,39 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using BenchmarkDotNet.Attributes;
+using osu.Framework.Graphics;
+using osu.Framework.Utils;
+
+namespace osu.Framework.Benchmarks
+{
+    public class BenchmarkInterpolation : BenchmarkTest
+    {
+        [Benchmark]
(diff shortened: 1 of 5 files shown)


smoogipoo:
Just for completeness:

Old:

|                   Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
| InterpolateMarginPadding | 63.17 ns | 0.359 ns | 0.335 ns |     - |     - |     - |         - |
|         InterpolateFloat | 10.76 ns | 0.026 ns | 0.023 ns |     - |     - |     - |         - |

New:

|                   Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
| InterpolateMarginPadding | 64.11 ns | 0.115 ns | 0.102 ns |     - |     - |     - |         - |
|         InterpolateFloat | 10.96 ns | 0.029 ns | 0.027 ns |     - |     - |     - |         - |

PR #3344, opened 2020-03-09
Fix stack traces not being logged when performance profiling is turned on

peppy:
Was about to use this in https://github.com/ppy/osu/issues/8185 and found that the threading changes meant we were sending a null thread to the `BackgroundStackTraceCollector`


peppy:
#3347 instead.

Issue #3310, opened 2020-02-27
The wrong audio class is used on iOS


smoogipoo:
When running the game while music is playing (e.g. via Apple Music), the music is expected to be paused but instead keeps running in the background.

TimOliver:
This has now been merged. :)

PR #3351, opened 2020-03-09
Add Dynamic Audio Mode Switching on iOS

TimOliver:
Addresses #3310.

This PR attempts to introduce a compromise between the folks who always prefer the game audio to be on, and those who think the mute switch should be respected.

It changes Bass' audio policy to one that observes the mute switch on launch. And then, if the user does anything to change the volume (ie, clicks either volume buttons), it will then change the audio policy to one that disregards the mute switch, and play regardless.

It's been tested and confirmed to work in the osu!framework audio test suite on my iPad Pro.

I'm not 100% happy with the code layout. I'm not sure if putting this in the app delegate is polluting that class, or if it's enough code to warrant being in its own controller class. 

In any case, feedback welcome. :)


osu.Framework.iOS/GameAppDelegate.cs
```diff
@@ -34,6 +36,10 @@ public override bool FinishedLaunching(UIApplication application, NSDictionary l
 
             host.Run(CreateGame());
 
+            // Watch for the volume button changing in order to change audio policy
+            AVAudioSession audioSession = AVAudioSession.SharedInstance();
+            audioSession.AddObserver(this, "outputVolume", NSKeyValueObservingOptions.New, IntPtr.Zero);
```
swoolcock: Is this triggered by toggling the mute switch, too?  If not you might want to catch that and do the same.

TimOliver: Ooh. That's a good point! I didn't bother looking into it, because in theory, the mute switch being off, even without changing the audio mode would yield the audio playing.

I was hoping there was an official API for actually detecting the mute switch changing so I could explicitly control the audio modes, but there doesn't seem to be any way of explicitly seeing the mute switch.

TimOliver: Welp. I just spent the last few hours trying to work out if there is *any* way to access the state of the mute switch (without resorting to hacky ways like playing silent audio files), and it doesn't look like there is.

This should do for now, and we can change it down the line if need be.

swoolcock: Toggling the switch while an app is playing with background audio permissions (such as Music and YouTube) will only put the ringer in silent mode, it doesn't stop all sounds from playing.  To be honest I think the framework should respect this functionality and entirely ignore the side switch.

peppy: I agree with this, but some people seem to disagree, that "all games should respect the mute switch". Maybe needs further user testing or a poll.

TimOliver: I'm one of those people. 😅 Mainly because I'm used to all apps being silent on launch, and only playing noise when I explicit tap on an item in the app itself. (The fact that the game opens with a very loud opening intro is probably what makes that obvious).

What if we considered making this an option in the settings page? 🤔

peppy: Food for thought:

image

This happens more than I'd like (3 times since last osu!stream update)

osu.Framework.iOS/GameAppDelegate.cs
```diff
@@ -50,5 +56,14 @@ private void aotImageSharp()
...
             }
         }
+
+        public override void ObserveValue(NSString keyPath, NSObject ofObject, NSDictionary change, IntPtr context)
+        {
+            if (keyPath != "outputVolume")
+                return;
+
+            AVAudioSession audioSession = AVAudioSession.SharedInstance();
+            audioSession.SetCategory(AVAudioSessionCategory.Playback);
```
peppy:
```suggestion
            switch (keyPath)
            {
                case "outputVolume":
                    AVAudioSession.SharedInstance().SetCategory(AVAudioSessionCategory.Playback);
                    break;
            }
```

osu.Framework.iOS/GameAppDelegate.cs
```diff
@@ -51,6 +57,16 @@ private void aotImageSharp()
             }
         }
 
+        public override void ObserveValue(NSString keyPath, NSObject ofObject, NSDictionary change, IntPtr context)
+        {
+            switch (keyPath)
+            {
+                case "outputVolume":
+                    AVAudioSession.SharedInstance().SetCategory(AVAudioSessionCategory.Playback);
```
peppy: should we only be setting the category when a change occurs the first time? is there any repercussion from calling this potentially many times per second?

osu.Framework.iOS/GameAppDelegate.cs
```diff
@@ -51,6 +57,16 @@ private void aotImageSharp()
             }
         }
 
+        public override void ObserveValue(NSString keyPath, NSObject ofObject, NSDictionary change, IntPtr context)
+        {
+            switch (keyPath)
+            {
+                case "outputVolume":
```
peppy: move "outputVolume" to a const at top of file, since it's used in two places

```csharp
private const string output_volume = "outputVolume";
```

osu.Framework.iOS/GameAppDelegate.cs
TimOliver: It should be fine. The default behaviour of setting properties on `NSObject`s is that internally, it will check to see if the value already matches, and if it does, it exits out.

The reason I didn't add a check in here was because I wasn't sure how to check what the current category is. For some reason, setting the category is an `int`, but retrieving the category has been mapped to a string in Xamarin land.

osu.Framework.iOS/GameAppDelegate.cs
TimOliver: Done!

PR #3353, opened 2020-03-11
Add interface for leased bindables

frenzibyte:
Allows for spotting leased bindable instances without needing to know its value type. (e.g. `bindable is ILeasedBindable`)


peppy (dismissed):

osu.Framework/Bindables/Bindable.cs
```diff
@@ -421,7 +421,7 @@ void ISerializableBindable.DeserializeFrom(JsonReader reader, JsonSerializer ser
             Value = serializer.Deserialize<T>(reader);
         }
 
-        private LeasedBindable<T> leasedBindable;
+        private ILeasedBindable<T> leasedBindable;
```
bdach, line 424: not sure there is much point to this change given it's a private reference... maybe would make more sense if this was `protected internal`, but as-is, it's pretty useless. would revert for now.

in fact this probably *should* be protected, because if you want to add leasing to `BindableNumber`, then this will be essentially a dead unused field just taking up stack space in that class, and you'll have another one duplicating this one in `BindableNumber` itself. that's in scope of the dependent PR, though.

osu.Framework/Bindables/ILeasedBindable.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework.Bindables
+{
+    /// <summary>
+    /// An interface that represents a read-only leased bindable.
```
bdach, line 7: would probably drop the "read-only" qualifier from this xmldoc. if you want a read-only bindable, you might as well scope up to `IBindable`. the leased part (and, being able to return, as a consequence) is the important part.

bdach (changes requested):

osu.Framework/Bindables/Bindable.cs
frenzibyte: I've done support for `BindableNumber` this way:
https://github.com/ppy/osu-framework/blob/cb98257fdf4dd2c1b8027166e702daea86d841ef/osu.Framework/Bindables/Bindable.cs#L435-L445
https://github.com/ppy/osu-framework/blob/cb98257fdf4dd2c1b8027166e702daea86d841ef/osu.Framework/Bindables/BindableNumber.cs#L444

So this is required as-is (`LeasedBindableNumber<T>` doesn't inherit `LeasedBindable<T>`), unless you want me to move it to the dependent PR, can do.

bdach: as I said above, this change does not make much sense or stand on its own in this PR, so I'd move it back where it does make sense.

frenzibyte: Ah, that's what you meant, alright.

bdach (approved):
looks inoffensive enough to me

osu.Framework.Tests/Bindables/BindableLeasingTest.cs
```diff
@@ -89,7 +89,7 @@ public void TestIncorrectEndLease()
 
             // end a lease with an incorrect bindable
             original.BeginLease(true);
-            Assert.Throws<InvalidOperationException>(() => original.EndLease(original));
+            Assert.Throws<InvalidOperationException>(() => original.EndLease(new Bindable<int>().BeginLease(true)));
```
peppy, line 92: what's the reasoning behind this change?

frenzibyte: EndLease now only accepts `ILeasedBindable<T>` as an argument, so had to make another leased instance for this.

peppy (approved):

osu.Framework.Tests/Bindables/BindableLeasingTest.cs
peppy: Completely missed that. Seems fine.

PR #3354, opened 2020-03-11
Fix stalls in audio playback after silence

peppy:
- Closes https://github.com/ppy/osu/issues/7344

The addition of `UpdatePeriod` is not required but I feel better having it there for now.


bdach:
I'll test this on Android within the next day or two. Nice work :+1:

peppy:
As a heads up I forgot to update the framework in the latest release. Will do another deploy on friday.

bdach:
I can confirm (using my repro case and some of the ones supplied in the original report) that this also seems to resolve total gameplay stalls on beatmap start on Android! 🎉

PR #3355, opened 2020-03-11
Expose Animation's frame index and frame data

peppy:


osu.Framework/Graphics/Animations/Animation.cs
```diff
@@ -136,11 +139,11 @@ public void AddFrames(IEnumerable<FrameData<T>> frames)
 
         private void displayFrame(int index)
```
smoogipoo: Index should be removed from this method if it's just going to use the current frame. Also make sure that any places where this is called explicitly set `CurrentFrameIndex` (e.g. `GotoFrame()`). Or otherwise revert the relevant change in this method.

PR #3358, opened 2020-03-11
Fix multiple issues with TextureAtlasVisualiser

peppy:
- Fixes references being held indefinitely (even when not open)
- Fixed bind count being incorrect (was counting usages, not binds)
- Shows bind count as a number
- Changes colour range to show even low usages
- Removes unnecessary `TextureRemoved` logic (relies on `WeakReference` now)
- Renames `TextureAdded` to `TextureCreated`

### DIFF
--- a/osu.Framework/Graphics/OpenGL/GLWrapper.cs
+++ b/osu.Framework/Graphics/OpenGL/GLWrapper.cs
@@ -359,10 +359,12 @@ namespace osu.Framework.Graphics.OpenGL
         /// </summary>
         /// <param name="texture">The texture to bind.</param>
         /// <param name="unit">The texture unit to bind it to.</param>
-        public static void BindTexture(TextureGL texture, TextureUnit unit = TextureUnit.Texture0)
+        public static bool BindTexture(TextureGL texture, TextureUnit unit = TextureUnit.Texture0)
         {
-            BindTexture(texture?.TextureId ?? 0, unit);
+            bool didBind = BindTexture(texture?.TextureId ?? 0, unit);
             last_bound_texture_is_atlas[GetTextureUnitId(unit)] = texture is TextureGLAtlas;
+
+            return didBind;
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


osu.Framework/Graphics/OpenGL/GLWrapper.cs
```diff
@@ -359,33 +359,36 @@ internal static void SetActiveBatch(IVertexBatch batch)
...
             last_bound_texture_is_atlas[GetTextureUnitId(unit)] = texture is TextureGLAtlas;
+
+            return didBind;
         }
 
         /// <summary>
         /// Binds a texture to draw with.
         /// </summary>
         /// <param name="textureId">The texture to bind.</param>
         /// <param name="unit">The texture unit to bind it to.</param>
```
smoogipoo: Needs an xmldoc for the return value. Something along the lines of:

`<c>true</c> if the texture is newly-bound. <c>false</c> if the texture was already bound.`

osu.Framework/Graphics/OpenGL/Textures/TextureGLAtlas.cs
```diff
@@ -15,48 +16,23 @@ internal class TextureGLAtlas : TextureGLSingle
...
-        /// <summary>
-        /// The total amount of times this <see cref="TextureGLAtlas"/> was bound.
-        /// </summary>
-        public ulong BindCount { get; private set; }
+        public static event Action<TextureGLAtlas> TextureCreated;
 
         public TextureGLAtlas(int width, int height, bool manualMipmaps, All filteringMode = All.Linear)
             : base(width, height, manualMipmaps, filteringMode)
         {
             lock (all_atlases)
```
smoogipoo: can remove this lock now

osu.Framework/Graphics/OpenGL/Textures/TextureGLAtlas.cs
```diff
@@ -74,8 +50,6 @@ protected override void Dispose(bool isDisposing)
 
             lock (all_atlases)
```
smoogipoo: And this. Maybe just remove this method altogether?

peppy: Leaving method for safety (we don't want to expose the raw list).

PR #3360, opened 2020-03-11
Add ability to get bindings of an IBindable

frenzibyte:
Adds an enumerable to get all bindings of an `IBindable` as read-only bindables (of type `IBindable` / `IBindable<T>`).

Required for #3361 to access the bindable's bindings for value change propagation.

# vNext
## Bindable bindings are now a member of `IBindable` and `IBindable<T>`
Any `IBindable`/`IBindable<T>` implementations must implement these members:
```csharp
IEnumerable<IBindable> Bindings { get; }
IEnumerable<IBindable<T>> Bindings { get; }
```

### DIFF
--- a/osu.Framework/Bindables/Bindable.cs
+++ b/osu.Framework/Bindables/Bindable.cs
@@ -132,6 +132,11 @@ namespace osu.Framework.Bindables
 
         private WeakReference<Bindable<T>> weakReference => weakReferenceCache.IsValid ? weakReferenceCache.Value : weakReferenceCache.Value = new WeakReference<Bindable<T>>(this);
 
+        IEnumerable<IBindable> IBindable.Bindings => InternalBindings;
+        public IEnumerable<IBindable<T>> Bindings => InternalBindings;
+
+        protected LockedWeakList<Bindable<T>> InternalBindings { get; private set; }
+
         /// <summary>
         /// Creates a new bindable instance. This is used for deserialization of bindables.
         /// </summary>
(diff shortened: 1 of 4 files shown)


bdach:
I take issue with exposing bindings publicly, especially in the way done here. You're leaking a reference to the `InternalBindings` list, which means you're now one cast away from messing with internal state.

iiSaLMaN:
> 
> 
> I take issue with exposing bindings publicly, especially in the way done here. You're leaking a reference to the `InternalBindings` list, which means you're now one cast away from messing with internal state.

Yeah, even though casting sometimes gives you what you want, say you've got an `IBindable<string>` that you're meant to read only, you can cast that to `Bindable<string>` and change its value. anyways, the dependent PR only requires it to be at the very least `protected internal`, will change to that if this still bothers you.

peppy:
Why would this ever be required? When making a change which exposes something that used to be private, please provide reasoning.

iiSaLMaN:
https://github.com/ppy/osu-framework/pull/3361 in `BindableProperty.cs`, it has to propagate the value to the source bindable's `Bindings`, so I was forced to change the accessibility of `Bindings`, and also move it to the interface.

frenzibyte:
I'll close this, I don't believe this is right at all, and the direction of `BindableProperty` doesn't sound right.

I think what should be done is making `BindableProperty`s bind to each other and have their own bindings list, instead of exposing the `Bindable`'s bindings list and some weird action to get the equivalent property in another `Bindable` instance.

PR #3361, opened 2020-03-12
Introduce BindableProperty<T>

frenzibyte:
## Introducing `BindableProperty<T>`
`BindableProperty` is a class that wraps up value propagation logic of a certain property inside, all that is required from the consumer is to create an instance of it and provide the following:
 - `initialValue` (ctor param): The initial value of the property. (`OnValueChange` is not invoked on this)
- `OnValueChange`: An event to be invoked whenever a value has changed after it has been propagated to bindings.

Here's an example of adding a string bindable property:
```csharp
class MyBindable<T> : Bindable<T>
{
    public event Action<string, string> StringChanged;

    private BindableProperty<string> stringProperty;

    public MyBindable(string initialString)
    {
        stringProperty = new BindableProperty<string>(initialString, this, b => (b as MyBindable<T>)?.stringProperty)
        {
            OnValueChange = (oldValue, newValue) => StringChanged?.Invoke(oldValue, newValue)
        }
    }

    public string String
    {
        get => stringProperty.Value;
        set => stringProperty.Value = value;
    }

    public override void BindTo(Bindable<T> them)
    {
        if (them is MyBindable<T> other)
            String = other.String;

        base.BindTo(them);
    }
}
```

### DIFF
--- a/osu.Framework/Bindables/Bindable.cs
+++ b/osu.Framework/Bindables/Bindable.cs
@@ -132,6 +132,11 @@ namespace osu.Framework.Bindables
 
         private WeakReference<Bindable<T>> weakReference => weakReferenceCache.IsValid ? weakReferenceCache.Value : weakReferenceCache.Value = new WeakReference<Bindable<T>>(this);
 
+        IEnumerable<IBindable> IBindable.Bindings => InternalBindings;
+        public IEnumerable<IBindable<T>> Bindings => InternalBindings;
+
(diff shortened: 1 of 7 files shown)


frenzibyte:
I will close this and attempt this again without the `IBindable.Bindings` exposure and that `IHasPropertyGuards` thing, and PR again once I reach a sensible state with this.

PR #3362, opened 2020-03-12
Use BindableProperty<T> in Bindable<T>

frenzibyte:
- [ ] Depends on #3361 
- [ ] Depends on #3353


frenzibyte:

PR #3363, opened 2020-03-12
Use BindableProperty<T> in BindableNumber<T>

frenzibyte:
- [ ] Depends on #3362 

Decided to move this one out of the previous PR as it seem to be a bit longer in terms of LoC.


frenzibyte:

PR #3364, opened 2020-03-12
Use BindableProperty<T> in BindableList<T>

frenzibyte:
- [ ] Depends on #3361


frenzibyte:

PR #3366, opened 2020-03-12
Throw when binding a BindableNumber<T> to a bindable with incompatible value

frenzibyte:
- Covers binding a `BindableNumber` with defined properties to a normal `Bindable` with a value that isn't compatible with the `BindableNumber` properties. (precision / min-value / max-value)

Found this edge-case while fixing the bindable properties transferring behaviour and realized it didn't have a test, went ahead and wrote one.

### DIFF
--- a/osu.Framework.Tests/Bindables/BindableBindingTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableBindingTest.cs
@@ -23,6 +23,15 @@ namespace osu.Framework.Tests.Bindables
         }
 
         [Test]
+        public void TestBindToIncompatibleValueBindable()
+        {
+            BindableInt bindable1 = new BindableInt { MinValue = 100, MaxValue = 200, Value = 150 };
+            Bindable<int> bindable2 = new Bindable<int>(50);
+
+            Assert.Throws<InvalidOperationException>(() => bindable1.BindTo(bindable2));
+        }
+
+        [Test]
         public void TestPropagation()
         {
             Bindable<string> bindable1 = new Bindable<string>("default");
--- a/osu.Framework.Tests/Bindables/BindableNumberTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableNumberTest.cs
@@ -200,23 +200,6 @@ namespace osu.Framework.Tests.Bindables
             Assert.That(bindable.Value, Is.EqualTo(3));
         }
 
-        /// <summary>
-        /// Ensures that the value is updated when you bind to a bindable with a value that doesn't respect the number properties.
-        /// </summary>
-        [Test]
-        public void TestValueUpdatedOnBindingToInvalidValueBindable()
-        {
-            var bindable = new BindableInt { MinValue = 50, MaxValue = 200, Value = 55 };
-            var remote = new Bindable<int>(15);
-
-            // The value on bindable should change in this sequence: 55 (initial) -> 15 ('remote' value on bind) -> 50 (value updated with number props)
-            // The value of remote should change in this sequence: 15 (initial) -> 50 (updated from 'bindable')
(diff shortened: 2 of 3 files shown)


peppy:
Is this supposed to pass?

iiSaLMaN:
> Is this supposed to pass?

Sorry, I wasn't able to work on this, should be good now. (long story short my laptop can't handle the amount of work I'm putting it in due to how slow it is.., will upgrade it soon)

frenzibyte:
Upon coming back to this (in attempt to get rid of my old PRs), I've got confused for a bit on what this PR is about, but I understand it now and I think it's not quite correct/clear in its current form.

Initially, the PR was supposed to add a test case covering a potential edge case when binding a `BindableNumber` to a regular `Bindable` with incompatible values ([`e08561b` (#3366)](https://github.com/ppy/osu-framework/pull/3366/commits/e08561b0aef3e9ae713e5e1a21e53f910bfb7650)), but turns out that `BindableNumber` actually fails to handle this edge case.

Need to know whether binding `BindableNumber` to a regular `Bindable` should be valid to begin with, so I can proceed to either attempting to fix the above edge case or just throw an exception when binding to a `Bindable` and call it a day.

peppy:
Have you checked for existing usages in osu/framework where we are binding a `Bindable` to `BindableNumber`? That'd be a good starting point in making a decision on how we handle this going forward.

frenzibyte:
In the `BindableNumber.BindTo(Bindable)` case, found only 1 existing usage osu!-side (`SkinnableComboCounter#L26`), and it's a valid argument that one could mix up between binding with `BindableNumber<T>` or `Bindable<T>` (which has been done a lot in the opposite case `Bindable.BindTo(BindableNumber`), therefore I would say this shouldn't really throw an exception out there.

After thinking about the issue for a bit, this entire "BindableNumber<T> not handling binding to incompatible value" edge-case is directly coming from `Bindable` not handling value rejection on `BindTo()` properly, causing desynchronisation between it and the other bindable.
Here's a test I wrote that shows the underlying issue that I'm talking about:
<details>
<summary>"Bind with upstream rejection" test</summary>

```csharp
        [Test]
        public void TestBindWithUpstreamRejection()
        {
            Bindable<string> bindable1 = new Bindable<string>("rejected value");
            Bindable<string> bindable2 = new Bindable<string>();

            int changed1 = 0, changed2 = 0;

            bindable1.ValueChanged += v => changed1++;
            bindable2.ValueChanged += v =>
            {
                bindable2.Value = "accepted value";
                changed2++;
            };

            bindable2.BindTo(bindable1);

            Assert.AreEqual("accepted value", bindable1.Value);
            Assert.AreEqual(bindable1.Value, bindable2.Value);

            // bindable1 should only receive the final value changed, skipping the intermediary (overidden) one.
            Assert.AreEqual(1, changed1);
            Assert.AreEqual(2, changed2);
        }
```

</details>

The problem is, handling the above case properly means allowing the other bindable value to be changed just by a `BindTo()` call, therefore this could be abused to change `IBindable<T>` instances without casting or anything explicit:
```csharp
IBindable<double> abuser = new BindableDouble { MinValue = 50, MaxValue = 50 }
abuser.BindTo(readOnlyBindable);
// readOnlyBindable.Value will change to 50.
```

But I believe we don't care that much about `IBindable<T>` instances getting modified, so I would assume it's fine fixing this issue, what do you think?

peppy:
What is the result of that test method in your comment? does that fail right now? do you have a fix in mind for that?

Really having trouble parsing this issue and your train of thought. Maybe just do things the way you expect them to be and allow me to review the final changes.

frenzibyte:
Check #4069

Issue #3311, opened 2020-02-27
The game shouldn't keep running while backgrounded


smoogipoo:
Causes battery death / excessive heat production on mobile devices, even though nothing is being displayed.

peppy:
May be specific to xamarin debug scenarios, but should investigate regardless that we are doing suspension correctly.

TimOliver:
I'm exploring this now.

I tested in iOS land that upon locking the device, or dismissing the app, all of the appropriate `UIApplicationDelegate` callbacks (eg, "device will suspend") are being called properly without issue.

That being said, I'm using Activity Monitor in Instruments to detect how much CPU usage osu!framework tests takes up.

At any point when open, osu!framework consumes 70% of the CPU load of my iPad Pro. If I lock the device, usage drops to 1%.

But. If I dismiss the app and return to the Home Screen, without actually closing the app in the background, the CPU usage doesn't drop and continues at 70% *while the app is completely not in the foreground*.

![Screen Shot 2020-03-10 at 16 56 39](https://user-images.githubusercontent.com/429119/76296607-5de2b380-62f1-11ea-914d-f3ea56fb2ba3.png)

This screenshot was taken while the iPad was unlocked, but osu!framework was in the background.

FWIW, this is a Debug build of osu!framework.

TimOliver:
Did a stack trace in Instruments Time Profiler.

![Screen Shot 2020-03-10 at 17 33 14](https://user-images.githubusercontent.com/429119/76301849-efeeba00-62f9-11ea-95b7-de074a9b068c.png)

peppy:
Closed via #3367 (to some extent). There are still some background threads we aren't yielding, but at very least it should be greatly improved to the point of not affecting battery life.

PR #3367, opened 2020-03-12
Suspend/Resume all threads when app is backgrounded in iOS

TimOliver:
Addresses #3311.

At present, even when the app is backgrounded in iOS, the process will still constantly consume the same amount of CPU resources.

This PR solves that by explicitly pausing all of the threads when iOS reports the app has been backgrounded, and subsequently starting them up again when resumed.

I had an issue where even after suspending all the threads, the next `windowUpdate()` would automatically resume them all, so I added a `suspended` flag to stop that. But if there is a way to make `windowUpdate()` pause as well, that would be better.


peppy:
Logic looks quite correct. Added a test because no test = no merge.

peppy:
@smoogipoo request a quick review of the test coverage

PR #3371, opened 2020-03-15
Refresh audio DeviceInfo async from AudioThread

2yangk23:
My attempt to fix: https://github.com/ppy/osu/issues/8185

This change starts up a new thread in the AudioManager that continuously polls for new audio DeviceInfo. Whenever new changes are detected, a task is scheduled to check to current device and whether it should be switched to a better one. (e.g. No Sound -> actual device)

Don't know if other people are actually affected by this latency problem, but I figured this device polling is non-critical and doesn't really need to be part of the main game threads.

### DIFF
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -86,7 +86,7 @@ namespace osu.Framework.Audio
 
         private Scheduler eventScheduler => EventScheduler ?? scheduler;
 
-        private readonly Thread deviceSyncThread;
+        private readonly CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
 
         /// <summary>
         /// The scheduler used for invoking publicly exposed delegate events.
@@ -126,27 +126,28 @@ namespace osu.Framework.Audio
                 return store;
             });
 
-            // sync audioDevices every 200ms
-            deviceSyncThread = new Thread(() =>
+            // sync audioDevices every 1000ms
+            CancellationToken token = cancellationTokenSource.Token;
+            var thread = new Thread(() =>
             {
-                while (true)
+                while (!token.IsCancellationRequested)
                 {
                     try
                     {
                         syncAudioDevices();
-                        System.Threading.Thread.Sleep(200);
+                        System.Threading.Thread.Sleep(1000);
                     }
                     catch
                     {
                     }
                 }
-            });
-            deviceSyncThread.Start();
+            }) { IsBackground = true };
+            thread.Start();
         }
 
         protected override void Dispose(bool disposing)
         {
-            deviceSyncThread.Abort();
+            cancellationTokenSource.Cancel();
             Thread.UnregisterManager(this);
 
             OnNewDevice = null;
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -246,9 +246,6 @@ namespace osu.Framework.Audio
                 InitBass(deviceIndex);
             }
 
-            // Sync newly initialized device
-            syncAudioDevices();
-
             if (Bass.LastError != Errors.OK)
             {
                 Logger.Log($@"BASS failed to initialize with error code {Bass.LastError:D}: {Bass.LastError}.", LoggingTarget.Runtime, LogLevel.Important);
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -3,16 +3,19 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using ManagedBass;
 using osu.Framework.Audio.Sample;
 using osu.Framework.Audio.Track;
-using osu.Framework.IO.Stores;
-using osu.Framework.Threading;
-using System.Linq;
-using System.Diagnostics;
 using osu.Framework.Bindables;
 using osu.Framework.Extensions.TypeExtensions;
+using osu.Framework.IO.Stores;
 using osu.Framework.Logging;
+using osu.Framework.Threading;
 
 namespace osu.Framework.Audio
 {
@@ -32,9 +35,6 @@ namespace osu.Framework.Audio
         /// The thread audio operations (mainly Bass calls) are ran on.
         /// </summary>
         internal readonly AudioThread Thread;
-
-        private List<DeviceInfo> audioDevices = new List<DeviceInfo>();
-        private List<string> audioDeviceNames = new List<string>();
 
         /// <summary>
         /// The names of all available audio devices.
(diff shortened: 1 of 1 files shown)


osu.Framework/Audio/AudioManager.cs
```diff
@@ -326,12 +326,14 @@ protected virtual IEnumerable<DeviceInfo> EnumerateAllDevices()
 
         protected virtual bool IsCurrentDeviceValid()
         {
-            var deviceIndex = Bass.CurrentDevice;
-            var device = deviceIndex == Bass.DefaultDevice ? default : Bass.GetDeviceInfo(deviceIndex);
-
+            var device = audioDevices.ElementAtOrDefault(Bass.CurrentDevice);
```
peppy: could the index potentially change now that the querying is on another thread?

osu.Framework/Audio/AudioManager.cs
```diff
@@ -120,34 +126,28 @@ public AudioManager(AudioThread audioThread, ResourceStore<byte[]> trackStore, R
...
+                    }
+                    catch
+                    {
+                    }
                 }
-                catch
-                {
-                }
-            }, 1000, true);
+            }, token, TaskCreationOptions.None, TaskScheduler.Default);
```
peppy: mark `LongRunning` or use `new Thread` instead. don't want to put this on the threadpool.

osu.Framework/Audio/AudioManager.cs
```diff
@@ -120,34 +126,28 @@ public AudioManager(AudioThread audioThread, ResourceStore<byte[]> trackStore, R
...
+            CancellationToken token = cancellationTokenSource.Token;
+            Task.Factory.StartNew(() =>
             {
-                try
+                while (!token.IsCancellationRequested)
                 {
-                    setAudioDevice(AudioDevice.Value);
+                    try
+                    {
+                        var task = Task.Delay(200, token);
```
peppy: you changed the rate 5x what it was before. not sure about this. the device enumeration is doing considerable allocations.

peppy: also, use `Thread.Sleep` to avoid task allocations.

osu.Framework/Audio/AudioManager.cs
2yangk23: From this [thread](http://www.un4seen.com/forum/?topic=18835.0), my understanding is that devices will not move around after they have been added.

osu.Framework/Audio/AudioManager.cs
2yangk23: Hm, I could separately check just the current device more frequently (~100-200ms) and change this to every 1000ms like it is currently, but then it requires the audioDevices list to be mutable. (which requires a bunch of manual locking) I had done it this way at first, but switched to immutable list to keep things simpler. Would you prefer manual locking of the list to reduce allocations?

Alternatively, I could just change this to 1000ms, and let device changes be slower. So if you are using a device that becomes disabled, it will take up to 1000ms for the game to find a new device to use.

osu.Framework/Audio/AudioManager.cs
```diff
@@ -120,34 +126,27 @@ public AudioManager(AudioThread audioThread, ResourceStore<byte[]> trackStore, R
...
-                {
-                }
-            }, 1000, true);
+            });
+            deviceSyncThread.Start();
         }
 
         protected override void Dispose(bool disposing)
         {
+            deviceSyncThread.Abort();
```
peppy: pretty sure this is deprecated.

osu.Framework/Audio/AudioManager.cs
```diff
@@ -120,34 +126,27 @@ public AudioManager(AudioThread audioThread, ResourceStore<byte[]> trackStore, R
...
-            {
-                try
-                {
-                    setAudioDevice(AudioDevice.Value);
-                }
-                catch
-                {
-                }
-            }, 1000, true);
+            });
```
peppy: need to mark `IsBackground`

osu.Framework/Audio/AudioManager.cs
peppy: I think per second is fine if we're going to go with the single method

2yangk23:
Are the tests supposed to pass? It's passing locally but it seems like they are timing out of AppVeyor. The tests require there to be at least 1 audio device so maybe that's why?

peppy:
Yes, tests should pass. You can likely test locally by disabling all output devices.

osu.Framework/Audio/AudioManager.cs
```diff
@@ -326,12 +336,14 @@ protected virtual IEnumerable<DeviceInfo> EnumerateAllDevices()
 
         protected virtual bool IsCurrentDeviceValid()
         {
-            var deviceIndex = Bass.CurrentDevice;
-            var device = deviceIndex == Bass.DefaultDevice ? default : Bass.GetDeviceInfo(deviceIndex);
-
+            var device = audioDevices.ElementAtOrDefault(Bass.CurrentDevice);
```
smoogipoo: Potential nullref here.

2yangk23: What case are you thinking of? I don't see how this can be null.

bdach: The actual access here looks fine, the subsequent accesses to fields of `device` next line less so due to the `OrDefault` part.

2yangk23: Actually it should be fine because DeviceInfo is a struct, and IsEnabled and IsInitialized are both bit flags which would be 0 by default.

2yangk23:
I noticed a slight behavior change when you rearranged the thread which now causes the audio device to initialized twice at startup. This is my best explanation of why it happens:

There are 2 ways `setAudioDevice` can be triggered:
- Loading config (`AudioDevice` Bindable) at startup schedules `onDeviceChanged`
- When `syncAudioDevices` detects a change in `audioDeviceNames` it schedules `onDevicesChanged`

When the game starts up, it both loads the config and updates the initial list, which causes 2 calls to `setAudioDevice` to be scheduled. The problem is that `audioDevices` becomes stale after an init because the device that has just been initialized still looks uninitialized in the list. (I had done this purposely to try and avoid running `GetDeviceInfo` in the `AudioThread`) This is fine under the **assumption** that `audioDevices`/`audioDeviceNames` will always be updated right before a call to `setAudioDevice` which is broken in this case.

Potential solutions:
1. Fix the assumption and update `audioDevices` and `audioDeviceNames` before `setAudioDevice` when loading from config. This makes a call to `GetDeviceInfo` in game thread, but loading from config is rare and won't happen at a critical time.
2. Update `audioDevices` and `audioDeviceNames` always after device initialization. Also makes a call to `GetDeviceInfo` in game thread, and this may cause game lag if the device changes and gets initialized during game play.
3. Move the initial sync back into the thread. (How it was before) This is kind of janky, but the reason it works is that the config value is loaded before the first sync, so the config value loading is not able to set an audio device, and the audio device ends up being set during the first sync.

Which one do you think is appropriate? (My preference 1>3>2)

peppy:
We have to ensure that all initialisation is done on the audio thread, I think. If the audio thread isn't started, we shouldn't be making bass calls.

Is it feasible to make the config call scheduled and debounced (using `scheduler.AddOnce()`)?

2yangk23:
One call is `setAudioDevice(args.NewValue)` and the other is `setAudioDevice() == setAudioDevice(AudioDevice.Value)` would these two be considered the same? `args.NewValue` should be equivalent to `AudioDevice.Value`

peppy:
for the first part, you can just change the method to be setAudioDevice and always reference the bindable itself, rather than using the args.

not sure about second part, would need to investigate further.

just to be certain: the current `master` may initialise twice but does it cause any adverse effects?

2yangk23:
Doesn't cause any problems, just undesired behavior. I'll try a few things to resolve it.

Issue #2697, opened 2019-07-25
Assertion hit in ScrollContainer.OnDrag


peppy:
```

Assertion Failed

   at osu.Framework.Graphics.Containers.ScrollContainer`1.OnDrag(DragEvent e) in C:\Users\Dean\Projects\osu-framework\osu.Framework\Graphics\Containers\ScrollContainer.cs:line 304
   at osu.Game.Graphics.Containers.OsuScrollContainer.OnDrag(DragEvent e) in C:\Users\Dean\Projects\osu\osu.Game\Graphics\Containers\OsuScrollContainer.cs:line 60
   at osu.Framework.Graphics.Drawable.TriggerEvent(UIEvent e) in C:\Users\Dean\Projects\osu-framework\osu.Framework\Graphics\Drawable.cs:line 1877
   at osu.Framework.Input.MouseButtonEventManager.<>c__DisplayClass38_0.<PropagateMouseButtonEvent>b__0(Drawable target) in C:\Users\Dean\Projects\osu-framework\osu.Framework\Input\MouseButtonEv
entManager.cs:line 269
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
   at osu.Framework.Input.MouseButtonEventManager.PropagateMouseButtonEvent(IEnumerable`1 drawables, MouseButtonEvent e) in C:\Users\Dean\Projects\osu-framework\osu.Framework\Input\MouseButtonEv
entManager.cs:line 269
   at osu.Framework.Input.MouseButtonEventManager.HandleMouseDrag(InputState state, Vector2 lastPosition) in C:\Users\Dean\Projects\osu-framework\osu.Framework\Input\MouseButtonEventManager.cs:l
ine 225
   at osu.Framework.Input.MouseButtonEventManager.HandlePositionChange(InputState state, Vector2 lastPosition) in C:\Users\Dean\Projects\osu-framework\osu.Framework\Input\MouseButtonEventManager
.cs:line 116
   at osu.Framework.Input.InputManager.HandleMousePositionChange(MousePositionChangeEvent e) in C:\Users\Dean\Projects\osu-framework\osu.Framework\Input\InputManager.cs:line 470
[...]

PR #3373, opened 2020-03-16
Only compute clampedScrollOffset if necessary

2yangk23:
Fixes: https://github.com/ppy/osu-framework/issues/2697


2yangk23:
I tried adding a test, but no matter what I do, the ScrollContainer can only scroll in integer value offsets, which won't cause this problem since it's caused by precision.

2yangk23:
Turns out the offset is dependent on the size of the window. Gonna need to think of a more reliable way to test this.

Issue #3377, opened 2020-03-17
Default audio device no longer updates


Joppe27:
Since the update a few moments ago, the default update device will not update anymore and will keep outputting to the audio device that was default on lazer startup. Probably a mistake in [osu-framework#3371](https://github.com/ppy/osu-framework/pull/3371) .
version: latest
hw: realtek analog audio

iiSaLMaN:
I think this should've been closed since the merge of #3378?

PR #3378, opened 2020-03-17
Fix change detection in audioDevices list

2yangk23:
This seems to fix the double init, as well as https://github.com/ppy/osu-framework/issues/3377

Let me know if you find any problems.

### DIFF
--- a/osu.Framework.Tests/Audio/AudioThreadTest.cs
+++ b/osu.Framework.Tests/Audio/AudioThreadTest.cs
@@ -36,6 +36,8 @@ namespace osu.Framework.Tests.Audio
 
             thread.Exit();
 
+            Manager?.Dispose();
+
             WaitForOrAssert(() => thread.Exited, "Audio thread did not exit in time");
         }
 
--- a/osu.Framework.Tests/Audio/AudioThreadTest.cs
+++ b/osu.Framework.Tests/Audio/AudioThreadTest.cs
@@ -38,8 +38,6 @@ namespace osu.Framework.Tests.Audio
             thread.Exit();
 
             WaitForOrAssert(() => thread.Exited, "Audio thread did not exit in time");
-
-            Bass.Free();
         }
 
         public void CheckTrackIsProgressing(Track track)
--- a/osu.Framework.Tests/Audio/AudioThreadTest.cs
+++ b/osu.Framework.Tests/Audio/AudioThreadTest.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using ManagedBass;
 using NUnit.Framework;
 using osu.Framework.Audio.Track;
 using osu.Framework.IO.Stores;
@@ -37,6 +38,8 @@ namespace osu.Framework.Tests.Audio
             thread.Exit();
 
             WaitForOrAssert(() => thread.Exited, "Audio thread did not exit in time");
+
+            Bass.Free();
         }
 
         public void CheckTrackIsProgressing(Track track)
--- a/osu.Framework.Tests/Audio/AudioThreadTest.cs
+++ b/osu.Framework.Tests/Audio/AudioThreadTest.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
-using ManagedBass;
 using NUnit.Framework;
 using osu.Framework.Audio.Track;
 using osu.Framework.IO.Stores;
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -129,11 +129,6 @@ namespace osu.Framework.Audio
 
             scheduler.Add(() =>
             {
-                syncAudioDevices();
-                // If user has no audio devices at initialization, force one to be set.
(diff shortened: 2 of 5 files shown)


osu.Framework.Tests/Audio/AudioThreadTest.cs
```diff
@@ -37,6 +38,8 @@ public void TearDown()
             thread.Exit();
 
             WaitForOrAssert(() => thread.Exited, "Audio thread did not exit in time");
+
+            Bass.Free();
```
peppy: why is this required if it's happening in the thread shutdown logic?

image

osu.Framework/Audio/AudioManager.cs
```diff
@@ -77,6 +77,8 @@ public class AudioManager : AudioCollectionManager<AdjustableAudioComponent>
             MaxValue = 1
         };
 
+        public override bool IsLoaded => Bass.CurrentDevice != Bass.DefaultDevice;
```
peppy: Is there not a case where `CurrentDevice` can be `DefaultDevice`?

2yangk23: From documentation of CurrentDevice:
> Gets or sets the device setting of the current thread... 0 = no sound, 1 = first real output device.
> A return value of -1 indicates error.

DefaultDevice = -1, but I could use -1 here instead if you think that's more clear. You can't actually initialize DefaultDevice so you could consider it invalid.

osu.Framework.Tests/Audio/AudioThreadTest.cs
2yangk23: You're right it is calling Free(), but I guess Free() does not reset CurrentDevice back to -1. This causes the next test to be unable to initialize properly and it gets stuck waiting on preconditions. For whatever reason, calling Free() twice set CurrentDevice back to -1, but I'll find a better way to solve this.

osu.Framework/Audio/AudioManager.cs
2yangk23: Very strange behavior:
>[AudioThread] Free() device:0
[AudioThread] result:True error:OK device:2
[DeviceLosingAudioTest] Free() device:2
[DeviceLosingAudioTest] result:True error:OK device:-1

peppy: Might be worth isolating this to ensure nothing is calling `Init` or another function between the two free calls

osu.Framework.Tests/Audio/AudioThreadTest.cs
peppy: Might be worth isolating this to ensure nothing is calling Init or another function between the two free calls

osu.Framework/Audio/AudioManager.cs
2yangk23: Actually I think I understand it now. Each time you call `Free()` it just frees a single device. But it is possible that we have multiple devices initialized and you need to keep calling `Free()` until `CurrentDevice` is back to `DefaultDevice`. Note that this is observed behavior and I'm not sure why it isn't documented...

osu.Framework.Tests/Audio/AudioThreadTest.cs
2yangk23: Actually I think I understand it now. Each time you call Free() it just frees a single device. But it is possible that we have multiple devices initialized and you need to keep calling Free() until CurrentDevice is back to DefaultDevice. Note that this is observed behavior and I'm not sure why it isn't documented...

peppy: http://www.un4seen.com/forum/?topic=16664.msg116540#msg116540 this is the only reference i can find, which suggests that the above may only happen if we are calling `Init` multiple times.

2yangk23: Yea we are calling `Init` multiples times. For example the test `TestPlaybackWithDeviceLoss` starts by initializing some device, the "loses" that device, which causes the NoSound device to get initialized. (2 devices are now initialized). When the audio thread was exiting, `Free` was only called once which only frees one of the devices. Then when the next test started, there was already an initialized device which caused problems.

peppy: I see.

So in teardown, let's do a `while` over freeing i guess. add a comment inline mentioning why we are doing this.

peppy:
@2yangk23 sorry for the back and forth, but could you please check my latest commit? I think it should resolve the issues without resorting to a `while` loop.

2yangk23:
Functionally it's working fine, but the test is still broken so you may not be calling Free() for all devices.

peppy:
Will take another look into this today (I can't repro the failure on a macOS host, seems windows specific).

2yangk23:
I looked at your code and it looks like you added the `Free` call when we get `Errors.Already`. The problem is you only get `Errors.Already` when you initialize the same device twice, not if you initialize two different devices. The problem that causes the test to fail is when you initialize two different devices at the same time (e.g. NoSound device + Real device).

peppy:
Right, I see. That makes sense as to why it's still failing then. Il'l add back the loop on top of my changes with a comment. Maybe we'll be able to resolve it better one day.

PR #3383, opened 2020-03-18
Allow handling drawable input event externally

luaneko:
Adds a public `OnInput` event that is raised by `Handle(UIEvent)` and can be registered from outside. This is useful when extending the input handling logic of some drawable without having to derive it.

### DIFF
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -276,8 +276,8 @@ namespace osu.Framework.Graphics
 
             double timeBefore = DebugUtils.LogPerformanceIssues ? perf_clock.CurrentTime : 0;
 
-            RequestsNonPositionalInput = HandleInputCache.RequestsNonPositionalInput(this);
-            RequestsPositionalInput = HandleInputCache.RequestsPositionalInput(this);
+            RequestsNonPositionalInput = onInput != null || HandleInputCache.RequestsNonPositionalInput(this);
+            RequestsPositionalInput = onInput != null || HandleInputCache.RequestsPositionalInput(this);
 
             RequestsNonPositionalInputSubTree = RequestsNonPositionalInput;
             RequestsPositionalInputSubTree = RequestsPositionalInput;
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -109,7 +109,7 @@ namespace osu.Framework.Graphics
                 Parent = null;
 
                 OnUpdate = null;
-                OnInput = null;
+                onInput = null;
                 Invalidated = null;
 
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Drawable.cs
```diff
@@ -2195,15 +2201,26 @@ public bool TriggerEvent(UIEvent e)
...
+        private bool requestsNonPositionalCached;
+        private bool requestsPositionalCached;
+
         /// <summary>
         /// Whether this drawable should receive non-positional input. This does not mean that the drawable will immediately handle the received input, but that it may handle it at some point.
         /// </summary>
-        internal bool RequestsNonPositionalInput { get; private set; }
+        internal bool RequestsNonPositionalInput
+        {
+            get => requestsNonPositionalCached || OnInput != null;
```
smoogipoo: What happens if the delegate is set after it’s constructed, like say, inside BDL `load()`?

peppy:
Not sure about this. We intentionally limit publicly exposing things like this to (hopefully) lead to more well-thought design. Can you give a use case where you would be seeing this as useful?

phosphene47:
In my specific case, I'm writing some drawable wrappers for my [ofreact](https://github.com/phosphene47/ofreact) project where consumers need to define input handlers declaratively rather than by deriving drawables.

This is because composition is preferred over inheritance in declarative programming. There is a set of built-in drawable-wrapping components like `Text`, `Box`, `Container`, etc. that form the building blocks of a more complex component; they are akin to sealed classes. And they need to receive input handling delegates from the rendering function of a more complex component, to *set* on the drawable.

Example pseudocode: This is a normal osu!framework drawable that uses inheritance to add input logic.

```csharp
class mytext : SpriteText {
  ctor() { Text = "my text"; }

  override OnMouseDown() => Text = "another text";
}
```

This would be equivalent to:

```csharp
class mytext : declarative Drawable {
  override Render() {
    var (text, setText) = UseState("my text");

    return new SpriteTextWrapper(
      text: text,
      onMouseDown: () => setText("another text"));
  }
}
```

`SpriteTextWrapper` class represents a `SpriteText` drawable in the declarative component tree. Notice that it is taking a delegate via constructor parameter: this delegate would be assigned to the `OnInput` event of `SpriteText`.

smoogipoo:
Since you're creating Drawables dynamically, maybe a better choice is to dynamically override `Handle()` to invoke your event. Going forward you could use this to apply hooks other than just to `Handle()`.

Something like https://docs.microsoft.com/en-us/dotnet/api/system.reflection.emit.typebuilder.definemethodoverride?view=netstandard-2.1 may be what I'm imagining.

smoogipoo:
Let's come back to this one when implementing OML (again).

PR #3385, opened 2020-03-19
Add game benchmark

smoogipoo:
Exposes:

1. Ability to run a single frame via invoking `RunSingleFrame()` (see: `BenchmarkManySpinningBoxes`).
2. Ability to create a custom game and run any logic on it - e.g. if you want to test adding a transform or querying a screenstack.


osu.Framework.Benchmarks/GameBenchmark.cs
```diff
@@ -0,0 +1,106 @@
...
+
+            public void RunSingleFrame()
+            {
+                ExecutionMode = ExecutionMode.SingleThread;
+                base.RunMainLoop();
+            }
+
+            public override void RunMainLoop()
+            {
+                if (HasRunOnce.IsSet)
```
peppy: probably want to stop this from running untethered (will take a full core)

PR #3387, opened 2020-03-19
Add some sanity checks for audio management

kOchirasu:
Sorry for so many PRs, but it should be good now. Not really fixing anything here just adding sanity checks.

Some takeaways:
- Do not call `Free` on the **active** device (Calling `Free` on an inactive device is fine). Active is defined as the device assigned by `Bass.ChannelSetDevice`.
- If a device was disabled and enabled again, it needs to be initialized again even though it will look to be already initialized from flags.
- If you switch from device A to B, then back to A again, you actually do not need to reinitialize A. Setting CurrentDevice to A is sufficient. (We don't differentiate this in the code and just re-init, but it seems to work fine)

### DIFF
--- a/osu.Framework/Threading/AudioThread.cs
+++ b/osu.Framework/Threading/AudioThread.cs
@@ -75,12 +75,16 @@ namespace osu.Framework.Threading
         {
             base.PerformExit();
 
+            List<AudioManager> managersCopy;
+
             lock (managers)
             {
-                foreach (var manager in managers)
-                    manager.Dispose();
+                managersCopy = new List<AudioManager>(managers);
                 managers.Clear();
             }
+
+            foreach (var manager in managersCopy)
+                manager.Dispose();
         }
     }
 }
--- a/osu.Framework/Threading/AudioThread.cs
+++ b/osu.Framework/Threading/AudioThread.cs
@@ -81,11 +81,6 @@ namespace osu.Framework.Threading
                     manager.Dispose();
                 managers.Clear();
             }
-
-            // Safety net to ensure we have freed all devices before exiting.
-            // This is mainly required for device-lost scenarios.
-            // See https://github.com/ppy/osu-framework/pull/3378 for further discussion.
-            while (Bass.Free()) { }
         }
     }
 }
--- a/osu.Framework.Tests/Audio/AudioManagerWithDeviceLoss.cs
+++ b/osu.Framework.Tests/Audio/AudioManagerWithDeviceLoss.cs
@@ -3,7 +3,6 @@
 
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
 using ManagedBass;
 using osu.Framework.Audio;
 using osu.Framework.IO.Stores;
@@ -93,7 +92,6 @@ namespace osu.Framework.Tests.Audio
         public override void Dispose()
         {
             Dispose(true);
-            Thread.Sleep(1000); // Ensure sync thread has been cancelled
             base.Dispose();
         }
     }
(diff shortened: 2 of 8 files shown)


2yangk23:
For some reason tests failed because I moved the Free call into AudioManager. Looking in to it...

2yangk23:
Don't know why this keeps failing. Passes locally every time with and without audio devices enabled.

osu.Framework/Threading/AudioThread.cs
```diff
@@ -75,17 +75,16 @@ protected override void PerformExit()
...
-                    manager.Dispose();
+                managersCopy = new List<AudioManager>(managers);
                 managers.Clear();
             }
 
-            // Safety net to ensure we have freed all devices before exiting.
-            // This is mainly required for device-lost scenarios.
-            // See https://github.com/ppy/osu-framework/pull/3378 for further discussion.
-            while (Bass.Free()) { }
+            foreach (var manager in managersCopy)
```
peppy: what's the reason for making a copy here?

2yangk23: It is a deadlock risk (during testing) since Dispose is called synchronously rather than schedules on AudioThread.

case:
1. AudioThread.Dispose is called
2. AudioThread acquires lock on managers
3. AudioThread calls dispose for all manager in managers
4. manager.Dipose is called
5. manager calls AudioThread.UnregisterManager
6. manager tries to acquire lock on managers (deadlock)

peppy: I see, due to the schedule of the dispose. We are 100% sure we do want to schedule the disposal calls still?

2yangk23: I don't know that scheduling the dispose call makes sense because it gets scheduled and AudioThread is immediately killed, so it will never be run. I remember that you had mentioned that it's scheduled because we need it to run on the AudioThread, so possibly what needs to happen is something like:

1. AudioThread.Dispose is called
2. AudioThread calls dispose for all managers, and unregisters them
3. AudioThread waits until all scheduled work is complete (how can this be guaranteed?)
4. AudioThread.Dispose completes

peppy: But this code here is guaranteed to be on the audio thread, right? So we can potentially call the non-queueing method I think?

Trying to understand where `manager.Dispose` is being called from, is it a finalizer or otherwise?

2yangk23: `manager.Dispose()` is normally called by `AudioThread.Dispose` and it schedules `manager.Dispose(bool)` in the AudioThread scheduler. I'm not sure where `AudioThread.Dispose` is called from.

During these I overrode `manager.Dispose()` in `AudioManagerWithDeviceLoss` to directly call `manager.Dispose(bool)` otherwise `Bass.Free` is never called and causes tests to interfere with each other.

peppy: Hmm, this is all very confusing and hard to follow. I'll have to debug and figure out the best path forward, since we really shouldn't be able to daedlock at this point, all going well.

2yangk23: With the current implementation, the deadlock should not happen normally (it's only a risk during testing)

Call stack would be something like this during testing:
```
AudioThread.Dispose() (lock)
>manager.Dispose()
>>manager.Dispose(bool)
>>>AudioThread.UnregisterManager() (deadlock)
```
what happens otherwise is:
```
AudioThread.Dispose() (lock)
>manager.Dispose()
>>schedule(manager.Dispose(bool))
(lock released)
[scheduler] manager.Dispose(bool) (lock)
```

peppy:
Looks like this failed on CI :(.

Issue #3389, opened 2020-03-19
Always use non-capturing lambda instead of method groups


smoogipoo:
Currently in C#, method groups always allocate a new delegate instance.

Our current R# inspections require us to convert non-capturing lambdas into method groups. We should reverse this rule if possible, or otherwise disable it.

See: https://github.com/dotnet/roslyn/issues/5835

I've submitted a request to R# to add an inspection for this: https://youtrack.jetbrains.com/issue/RIDER-41342

PR #3390, opened 2020-03-19
Remove some minor delegate allocations

smoogipoo:
This is the result of a custom analyser I've written to look for these cases. It's not ready for the public yet.

This is not a complete list of issues which would close https://github.com/ppy/osu-framework/issues/3389. Other cases involve restructuring how we're doing delegates to support a sender model or are otherwise more complex. Examples include:
1. `BufferedDrawNode` + `BufferedContainerDrawNode` (via the `InvokeOnDisposal`)
2. `AdjustableAudioComponent` (via `InvalidateState()`)

Master:
|                          Method |     Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------- |---------:|----------:|----------:|-------:|------:|------:|----------:|
| CreateSequenceWithDefaultEasing | 2.128 us | 0.0146 us | 0.0136 us | 0.4997 |     - |     - |   2.56 KB |

This PR:
|                          Method |     Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------- |---------:|----------:|----------:|-------:|------:|------:|----------:|
| CreateSequenceWithDefaultEasing | 2.061 us | 0.0122 us | 0.0108 us | 0.4616 |     - |     - |   2.38 KB |

(Reduction from the `TransformCustom` change)


peppy:
Think it's worth putting inline comments on these changes to mention the method group is being avoided for allocation reasons? I feel like someone is going to come along and PR it back to the original code because it looks cleaner.

PR #3392, opened 2020-03-20
Expose Target field in ScrollContainer

EVAST9919:


iiSaLMaN:
`Target` field is the target value for `Current` to approach to, which sounds weird to be exposed publicly. Can you please provide the reason for exposing such a field?

EVAST9919:
It will help me with https://github.com/ppy/osu/pull/7972. I need to know at which point the target will be changed.

iiSaLMaN:
Assuming this is about hiding the scroll-to-top button if it's already pressed, you probably want to add a `IsScrolledToStart` method instead.

EVAST9919:
No. My example: 
Screenshot_1

osu.Framework/Graphics/Containers/ScrollContainer.cs
```diff
@@ -118,7 +118,7 @@ public bool ScrollbarOverlapsContent
         /// <summary>
         /// The target scroll position which is exponentially approached by current via a rate of distanceDecay.
         /// </summary>
-        private float target;
+        public float Target { get; private set; }
```
iiSaLMaN:
```suggestion
        protected float Target { get; private set; }
```

EVAST9919: Sounds fair

PR #3393, opened 2020-03-22
Push screen before OnExiting instead of after

mcendu:
Fixes ppy/osu#8381

Can break some stuff as `next` is no longer the current screen during OnExiting.


osu.Framework/Screens/ScreenStack.cs
```diff
@@ -257,16 +257,21 @@ private bool exitFrom([CanBeNull] IScreen source, bool shouldFireExitEvent = tru
...
+                // Add the screen back on the stack to allow pushing screens in OnExiting.
+                stack.Push(toExit);
+
                 // if a screen is !ValidForResume, it should not be allowed to block unless it is the current screen (source == null)
                 // OnExiting should still be called regardless.
-                bool blockRequested = toExit.OnExiting(CurrentScreen);
+                bool blockRequested = toExit.OnExiting(next);
 
                 if ((source == null || toExit.ValidForResume) && blockRequested)
                 {
```
iiSaLMaN:
```suggestion
```

osu.Framework/Screens/ScreenStack.cs
```diff
@@ -257,16 +257,21 @@ private bool exitFrom([CanBeNull] IScreen source, bool shouldFireExitEvent = tru
...
-                    // If the exit event gets cancelled, add the screen back on the stack.
-                    stack.Push(toExit);
                     return true;
                 }
```
iiSaLMaN:
```suggestion
```

peppy:
tests would be great!

mcendu:
I am not sure about how to refer to the newly created object inside the object initializer itself, without inheritance or triggering CI error.

smoogipoo:
Can't push to your branch but you can rewrite it as:
```csharp
pushAndEnsureCurrent(() =>
{
    screen1 = new TestScreen(id: 1);
    screen1.Exiting = () => 
    {
        screen1.Push(new TestScreen(id: 2));
        return true;
    };

    return screen1;
}
```

PR #3394, opened 2020-03-22
Update .gitignore to support JetBrain Rider

EdrowsLuo:
https://github.com/JetBrains/resharper-rider-samples/blob/master/.gitignore


peppy:
Yep, please de-dupe (ie. only add new lines which were potentially missing).

PR #3396, opened 2020-03-25
Handle TextureAtlas findPosition() edge cases

ThereGoesMySanity:
There are 2 bugs fixed by this PR:

 * `Add` didn't account for the padding. This would either cause an infinite loop or create an invalid texture when the texture width or height was between 1008 and 1024.

 * If a new textureatlas was created inside `findPosition` with `Reset`, the position checks would be skipped after the atlas is created. Since the 24x24 white pixel has been added, textures with a width between 984 and 1008 would skip the X position check and would be offset outside the texture by the white pixel. To fix this, I made sure that the position checks were always performed.

### DIFF
--- a/osu.Framework.Tests/Graphics/TextureAtlasTest.cs
+++ b/osu.Framework.Tests/Graphics/TextureAtlasTest.cs
@@ -16,7 +16,7 @@ namespace osu.Framework.Tests.Graphics
         {
             TextureAtlas atlas = new TextureAtlas(1024, 1024);
             TextureGL texture = atlas.Add(width, height);
-            if(texture != null)
+            if (texture != null)
             {
                 Assert.AreEqual(texture.Width, width, message: $"Width: {texture.Width} != {width} for texture {width}x{height}");
                 Assert.AreEqual(texture.Height, height, message: $"Height: {texture.Height} != {height} for texture {width}x{height}");
--- a/osu.Framework.Tests/Graphics/TextureAtlasTest.cs
+++ b/osu.Framework.Tests/Graphics/TextureAtlasTest.cs
@@ -11,6 +11,28 @@ namespace osu.Framework.Tests.Graphics
     [TestFixture]
     public class TextureAtlasTest
     {
+        [TestCase(0, 0)]
+        [TestCase(1, 1)]
+        [TestCase(984, 20)]
+        [TestCase(985, 20)]
+        [TestCase(1020, 20)]
+        [TestCase(1024, 20)]
+        [TestCase(1025, 20)]
+        [TestCase(20, 984)]
+        [TestCase(20, 985)]
+        [TestCase(20, 1020)]
+        [TestCase(20, 1024)]
+        [TestCase(20, 1500)]
+        [TestCase(985, 985)]
+        [TestCase(1020, 985)]
+        [TestCase(1500, 985)]
+        [TestCase(1020, 1020)]
+        [TestCase(1500, 1500)]
+        public void TestAtlasAdd(int width, int height)
+        {
+            Assert.DoesNotThrow(() => testWithSize(width, height));
+        }
+
         private void testWithSize(int width, int height)
         {
             TextureAtlas atlas = new TextureAtlas(1024, 1024);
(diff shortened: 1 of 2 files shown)


peppy:
Thanks for the contribution!

This may be a good time to add unit tests for the TextureAtlas class. Is this something you're willing to look at? I think it should be quite easy to test it isolated from all other components.

ThereGoesMySanity:
Sure - do you want me to make a new PR for that or just use this one?

peppy:
In this PR is fine, covering the specific fixes you made and potentially any other edge cases you can think of.

ThereGoesMySanity:
Ok - the fix I just committed for 985x985 textures does not work because it prevents the white square from being added in the "white square only" TextureAtlas. I had to make more extensive modifications to the code to fix that, because there was no way to detect whether or not the white square was present from within `Add`.

peppy:
Please check the failing code style inspections https://ci.appveyor.com/project/peppy/osu-framework/builds/31701439/messages

peppy:
I've updated your tests to use `TestCase` for better visibility when failures occur:

image

osu.Framework.Tests/Graphics/TextureAtlasTest.cs
```diff
@@ -0,0 +1,59 @@
...
+                Assert.AreEqual(texture.Width, width, message: $"Width: {texture.Width} != {width} for texture {width}x{height}");
+                Assert.AreEqual(texture.Height, height, message: $"Height: {texture.Height} != {height} for texture {width}x{height}");
+
+                RectangleF rect = texture.GetTextureRect(null);
+                Assert.LessOrEqual(rect.X + rect.Width, 1, message: $"Returned texture is wider than TextureAtlas for texture {width}x{height}");
+                Assert.LessOrEqual(rect.Y + rect.Height, 1, message: $"Returned texture is taller than TextureAtlas for texture {width}x{height}");
+            }
+            else
+            {
+                Assert.True(width > 1024 - TextureAtlas.PADDING || height > 1008 - TextureAtlas.PADDING ||
```
smoogipoo: Why is this comparing against a height of 1008, and not 1024?

osu.Framework/Graphics/Textures/TextureAtlas.cs
```diff
@@ -67,19 +67,27 @@ public void Reset()
...
             {
                 Logger.Log($"TextureAtlas initialised ({atlasWidth}x{atlasHeight})", LoggingTarget.Performance);
                 Reset();
             }
-            else if (currentPosition.Y + height > atlasHeight - PADDING)
+
+            if (currentPosition.Y + height > atlasHeight - PADDING)
             {
+                if (height > atlasHeight - PADDING - WHITE_PIXEL_SIZE
+                    && width > atlasWidth - PADDING - WHITE_PIXEL_SIZE)
```
smoogipoo: Shouldn't this be an `||`? I.e. what happens if you have a texture that is 1px wide or 1px high, but exceeds the padding otherwise?
Specifically, with a testcase of `20x1008`, this creates a new texture atlas and makes the subtexture bite into the area reserved for the white pixel.

Also, after the bounds of the whitepixel are exceeded, the texture is allowed to be up to the padded size.

smoogipoo:
Failing case:

```
[TestCase(1008, 20)]
```

Evidently, `TextureAtlas` needs an overhaul or rewrite because this is so messy I can't observe these edge cases from a quick pass.

osu.Framework/Graphics/Textures/TextureAtlas.cs
```diff
@@ -106,12 +114,16 @@ private Vector2I findPosition(int width, int height)
         /// <returns>A texture, or null if the requested size exceeds the atlas' bounds.</returns>
         internal TextureGL Add(int width, int height)
         {
-            if (width > atlasWidth || height > atlasHeight)
+            if (width > atlasWidth - PADDING || height > atlasHeight - PADDING)
```
smoogipoo: This staggering of the conditions between the two methods needs to go.

osu.Framework/Graphics/Textures/TextureAtlas.cs
```diff
@@ -67,19 +67,27 @@ public void Reset()
...
             if (AtlasTexture == null)
             {
                 Logger.Log($"TextureAtlas initialised ({atlasWidth}x{atlasHeight})", LoggingTarget.Performance);
                 Reset();
             }
-            else if (currentPosition.Y + height > atlasHeight - PADDING)
+
+            if (currentPosition.Y + height > atlasHeight - PADDING)
             {
+                if (height > atlasHeight - PADDING - WHITE_PIXEL_SIZE
```
smoogipoo: Would prefer these as pre-checks outside the parenting `if`.

ThereGoesMySanity:
The edge cases are difficult to describe in words so I made a diagram to better demonstrate.

diagram

osu.Framework/Graphics/Textures/TextureAtlas.cs
ThereGoesMySanity: see diagram

osu.Framework/Graphics/Textures/TextureAtlas.cs
ThereGoesMySanity: see diagram

smoogipoo:
This needs better commenting, because it's not immediately obvious that it's done in the Y check to bypass whitepixel.

Though I would rather this maybe split out into two methods or via a `force` parameter or something, for the whitepixel case such that the checks could exist at a top-level instead of inside the y-check.

ThereGoesMySanity:
Should I make a new pull request for these changes?

smoogipoo:
yep!

PR #3401, opened 2020-03-26
Expose less backend-specific types in IWindow

swoolcock:
Partially cleans up `IWindow` and `Window` to expose osu!framework and .NET standard types rather than osuTK and/or Veldrid types.

Breaking changes for consumers (but not osu!):
* `IWindow.WindowState` now uses a framework enum rather than osuTK.
* `IWindow.VerticalSync` is now a bool rather than an osuTK enum.

A future PR will also remove/replace `CurrentDisplay` and `AvailableResolutions` as they currently use osuTK types, but that's a larger job and a breaking change for osu!.

### DIFF
--- a/osu.Framework/Platform/Sdl2WindowBackend.cs
+++ b/osu.Framework/Platform/Sdl2WindowBackend.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
-using System.Numerics;
 using osu.Framework.Caching;
 using osu.Framework.Extensions;
 using osu.Framework.Input.StateChanges;
@@ -15,7 +14,7 @@ using Veldrid.Sdl2;
 using Key = osuTK.Input.Key;
 using MouseButton = osuTK.Input.MouseButton;
 using MouseEvent = Veldrid.MouseEvent;
-using Point = Veldrid.Point;
+using Point = System.Drawing.Point;
 using TKVector2 = osuTK.Vector2;
 
 namespace osu.Framework.Platform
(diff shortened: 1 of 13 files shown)


osu.Framework/Platform/Sdl2WindowBackend.cs
```diff
@@ -68,36 +67,36 @@ public bool Visible
 
         public bool Exists => implementation?.Exists ?? false;
 
-        private Vector2 position = Vector2.Zero;
+        private Point position = Point.Empty;
```
smoogipoo: Is there a reason why you chose the Point/Size abstraction rather than Vector2 / Vector2I ?

swoolcock: The naming of `X`/`Y` felt wrong for a `Size` property, so I changed to `System.Drawing` instead.  Also changed to `Point` for consistency.  If you'd prefer we stick with `System.Numerics` that's no problem.

smoogipoo: I was thinking osuTK + our own `Vector2I` to be consistent with the rest of the codebase. But since numerics is also weird, let's go with this for now...

swoolcock: I'm trying to avoid all usage of osuTK types from now on.  Eventually we're going to have to go through and change all `Vector2` and `Color4` usages throughout the codebase if we ever plan on dropping osuTK entirely.

smoogipoo: Blanket ban on vector2 and color4 isn't going to achieve anything. `Color4` will be moved to our codebase, and `Vector2` can also or the translation to System.Numerics is going to be very simple for it.

I'm not interested in destroying interop between components of the framework for the purpose of not using osuTK.

smoogipoo:
Seems fine to me for now.

PR #3405, opened 2020-03-27
Add animation rewinding support

peppy:
- Implemented in a similar way to `VideoSprite`.
- Adds proper testing.
- Improves efficiency (no longer sets frame every update, unless required).
- Tested and confirmed to close https://github.com/ppy/osu/issues/8112.


osu.Framework.Tests/Visual/Sprites/TestSceneAnimation.cs
```diff
@@ -2,140 +2,184 @@
...
-                    if ((contentRelativeAxes & Axes.Y) == 0)
-                        c.Height = 100;
-
-                    AddFrame(c);
-                }
+                    Repeat = false,
+                };
+            });
+
+            AddUntilStep("Wait for video to load", () => animation.IsLoaded);
```
bdach:
```suggestion
            AddUntilStep("Wait for animation to load", () => animation.IsLoaded);
```

osu.Framework.Tests/Visual/Sprites/TestSceneAnimation.cs
```diff
@@ -2,140 +2,184 @@
...
-            }
-        };
+                    timeText = new SpriteText { Text = "Animation is loading..." }
+                };
+            });
 
-        private class TestDrawableAnimation : DrawableAnimation
+            loadNewAnimation();
+
+            AddStep("Reset clock", () => clock.CurrentTime = 0);
```
bdach: These steps are behaving strangely. Try running `JumpForward` and then `JumpBack` in the test browser - the second should time out on one of the seek steps. (Also noticeable that `playback` doesn't always go back to the same value)

As far as I can tell this is due to them being reversed (clock should be reset before the animation is loaded, as animation construction reads the clock's current time to determine offset, and therefore in this current way each test reads the current time at the *end* of the previous test).

osu.Framework/Graphics/Animations/Animation.cs
```diff
@@ -41,13 +62,43 @@ public abstract class Animation<T> : CompositeDrawable, IAnimation
...
+                base.Clock = offsetClock = new FramedOffsetClock(Clock) { Offset = -Clock.CurrentTime };
+        }
+
+        public override IFrameBasedClock Clock
+        {
+            get => base.Clock;
+            set
+            {
+                if (startAtCurrentTime)
+                    throw new InvalidOperationException($"A {nameof(Animation<T>)} with {startAtCurrentTime} = true cannot receive a custom {nameof(Clock)}.");
```
bdach: I assume you meant

```suggestion
                    throw new InvalidOperationException($"A {nameof(Animation<T>)} with {nameof(startAtCurrentTime)} = true cannot receive a custom {nameof(Clock)}.");
```

osu.Framework/Graphics/Animations/Animation.cs
```diff
@@ -90,8 +141,11 @@ public void GotoFrame(int frameIndex)
             else if (frameIndex >= frameData.Count)
                 frameIndex = frameData.Count - 1;
 
-            CurrentFrameIndex = frameIndex;
-            updateCurrentFrame();
+            if (!startAtCurrentTime)
+                throw new InvalidOperationException($"A {nameof(Animation<T>)} with {startAtCurrentTime} = false cannot seek as it is dependent on an external clock.");
```
bdach: same here (adding suggestion to speed things up & make sure nothing is missed)
```suggestion
                throw new InvalidOperationException($"A {nameof(Animation<T>)} with {nameof(startAtCurrentTime)} = false cannot seek as it is dependent on an external clock.");
```

osu.Framework/Graphics/Animations/Animation.cs
```diff
@@ -15,12 +19,29 @@ public abstract class Animation<T> : CompositeDrawable, IAnimation
...
 
+        /// <summary>
+        /// The current playback position of the animation, in milliseconds.
+        /// </summary>
+        public double PlaybackPosition
+        {
+            get
+            {
+                if (Repeat)
+                    return Clock.CurrentTime % Duration;
```
bdach: This has one gotcha - because of how modulo division is implemented, if the offset clock has a negative position a looped animation will not loop. It kind of makes sense, but it should probably be documented (because someone might as well expect it to always loop, as in take the absolute value of the modulo op).

peppy: good call. for what it's worth, this is just a copied implementation from VideoSprite so it will affect that too. if one is to change, both should.

osu.Framework.Tests/Visual/Sprites/TestSceneAnimation.cs
peppy: Fixed, the clock needed to be constructed each time.

osu.Framework/Graphics/Animations/Animation.cs
peppy: Going to leave it this way for now. I *think* it makes sense.

osu.Framework/Graphics/Animations/Animation.cs
```diff
@@ -41,13 +62,43 @@ public abstract class Animation<T> : CompositeDrawable, IAnimation
...
 
-        protected Animation()
+        private readonly Cached currentFrameCache = new Cached();
+
+        private FramedOffsetClock offsetClock;
+
+        /// <summary>
+        /// Construct a new animation.
+        /// </summary>
+        /// <param name="startAtCurrentTime">Whether the current clock time should be assumed as the 0th video frame.</param>
```
smoogipoo:
```suggestion
        /// <param name="startAtCurrentTime">Whether the current clock time should be assumed as the 0th animation frame.</param>
```

PR #3408, opened 2020-03-30
Fix more edge cases in TextureAtlas

ThereGoesMySanity:
Changes:
 * Added new test cases, covering a couple of missed edge cases
 * Refactored `Add()` to have a `whitePixel` boolean argument. This means that `findPosition()` will never be called for the white pixel, and therefore can assume that the white pixel exists when calculating whether a texture will fit.
 * Added comments

Assumptions:
    I changed a couple checks from `width - PADDING` to `width - PADDING * 2` since edge cases were failing due to this. In making this change, I assumed that the intention was for images to have padding on all 4 sides, even at the edges of the texture atlas. If images are allowed go all the way to the bottom/right edges of the texture atlas, then I'd need to make some changes.


peppy:
Still think there's much improvement to be had here in terms of readability.

- `findPosition` should probably be renamed to `allocatePosition` (and moved down below public/internal methods). It should also have xmldoc explaining what it does and what a `null` return means.
- not a huge fan of the `whitePixel` argument to `Add`. If you want to ensure it's added then make the implementation local to `Reset` (probably make `Reset` call an `addWhitePixel` private method)
- The pre-checks in `findPosition` make the logic a bit hard to follow. I think the check for dual-direction overflow should probably be nested inside the part where the atlas is about to be reset, since that's the only place it will be executed.

smoogipoo:
Part of the reason I found the code hard to follow was because it was nested where it was about to be reset.

Highly disagree with nesting it - it's interpreted as iterating through the atlas causing whoever knows what other processes to occur such as position offsetting, only to _then_ figure out, after doing so, that the texture couldn't be fit in the atlas at all.

Can you explain to me why prechecks are hard to parse? The point of prechecks is so that they're easy to parse and exclude from your mind - it's the reason why I invert conditions and reduce nesting, and we've had that discussion many times before.

peppy:
The specific one I have issue with is the one I mentioned above – it's prechecking for something that can only happen in one code path, which makes me read over the code many more times to try and understand why it's done that way. For me it causes more confusion than otherwise.

peppy:
For reference, I was hoping for something more like this:

image

Does that read OK to you?

smoogipoo:
I don't understand why you have to "read over the code many more times". This is the point of a pre-check - so you _dont_ have to do that. It's saying that this condition _must_ hold true for the further code to execute.
You should be reading the pre-check, understanding what it means (i.e. that the texture is able to fit in an empty atlas), and read the rest of the code while knowing that condition always holds true.

The condition you've pointed out is the exact one I had problems with.

peppy:
Specifically, 

image

By moving the sub condition out it becomes harder for me to read because I need to understand why that's been done (and because the logic isn't easy to parse it takes a while to confirm it's correct).

smoogipoo:
The reasoning behind "this can't happen without this" is incorrect. Because the first is checking if it can fit from the current position (which it can), and the second is checking if it can fit _at all_, which it can't. The two aren't related.

peppy:
Maybe splitting out the "can fit at all" check into a private method would work better? `bool checkWouldOverflowEmptyAtlasTexture(int width, int height)` or something?

osu.Framework/Graphics/Textures/TextureAtlas.cs
```diff
@@ -27,6 +27,9 @@ public class TextureAtlas
         private readonly int atlasWidth;
         private readonly int atlasHeight;
 
+        private int MaxWidth => atlasWidth - PADDING;
```
smoogipoo, line 30: Might as well factor in a second `PADDING` here.

ThereGoesMySanity: I didn't do that so that I could also use that property inside of findPosition, where the second padding is accounted for in currentPosition

smoogipoo: Given that a texture outright can't be added that exceeds 2*PADDING, I'd say it makes sense for the ones inside `findPosition` to be written as `height + PADDING`.

I wouldn't be against have a rectangle `bounds` and use `rectangle.Bottom` / `rectangle.Right`.

Also privates should be named `maxWidth` :p

osu.Framework/Graphics/Textures/TextureAtlas.cs
```diff
@@ -27,6 +27,9 @@ public class TextureAtlas
         private readonly int atlasWidth;
         private readonly int atlasHeight;
 
+        private int maxWidth => atlasWidth - PADDING * 2;
```
peppy, line 30: I'd rename to `maxFittableWidth` / `maxFittableHeight` for clarity.

osu.Framework/Graphics/Textures/TextureAtlas.cs
```diff
@@ -127,13 +127,13 @@ private Vector2I findPosition(int width, int height)
                 Reset();
             }
 
-            if (currentPosition.Y + height > MaxHeight)
+            if (currentPosition.Y + height + PADDING > atlasHeight)
```
smoogipoo, line 130: Shouldn't this be `- PADDING`? Or in other words `> atlasHeight + PADDING`.

osu.Framework/Graphics/Textures/TextureAtlas.cs
```diff
@@ -127,13 +127,13 @@ private Vector2I findPosition(int width, int height)
...
             {
                 Logger.Log($"TextureAtlas size exceeded {++exceedCount} time(s); generating new texture ({atlasWidth}x{atlasHeight})", LoggingTarget.Performance);
                 Reset();
             }
 
-            if (currentPosition.X + width > MaxWidth)
+            if (currentPosition.X + width + PADDING > atlasWidth)
```
smoogipoo, line 136: Same here

osu.Framework/Graphics/Textures/TextureAtlas.cs
peppy: we want the padding to be present, event against the edges. the current code seems correct?

smoogipoo: Ah I didn't realise this was using atlasHeight rather than `MaxHeight`, nevermind.

PR #3410, opened 2020-03-30
Use char overloads for string methods when possible

turbedi:
And some other minor string optimizations/cleanups


peppy:
check CI logs, your indentation is wrong.

osu.Framework/Extensions/TypeExtensions/TypeExtensions.cs
```diff
@@ -30,7 +30,7 @@ private static string readableName(Type t, HashSet<Type> usedTypes)
             {
                 var typeArgs = t.GetGenericArguments().Except(usedTypes);
                 if (typeArgs.Any())
-                    result += "<" + string.Join(",", typeArgs.Select(genType => readableName(genType, usedTypes))) + ">";
+                    result += '<' + string.Join(',', typeArgs.Select(genType => readableName(genType, usedTypes))) + '>';
```
peppy: i can only assume this is actually worse off.

image

PR #3411, opened 2020-03-31
Better support custom clocks in VideoSprite and Animation

peppy:
With the [recent change](https://github.com/ppy/osu-framework/pull/3405) to allow rewinding animations, support for `GotoFrame` was limited to when a custom clock was not in use.

This change allows `GotoFrame` and `startAtCurrentTime` to be used in combination with custom clocks. It also tidies up implementation to be (mostly, bar `GotoFrame`) shared between `Animation` and `VideoSprite`


smoogipoo:
Looks fine. Tested to make sure it works for catch?

peppy:
Yep, works as expected.

PR #3413, opened 2020-03-31
Fix warnings during pack

huoyaoyuan:
These warnings are not reported by AppVeyor, however visible when looking at logs.


peppy:
To make this easier to merge, can you explain each of the changes you made?

- Why NU5128 is being ignored (at very least a link to it)
- Why NU5104 is being ignored
- Why the metadata is removed from the csproj (is it being read from elsewhere?)
- Why the app.manifest was removed (wasn't being used by the template project itself nor included in the subprojects?)

huoyaoyuan:
Comments added for warnings.

> * Why the metadata is removed from the csproj (is it being read from elsewhere?)
> * Why the app.manifest was removed (wasn't being used by the template project itself nor included in the subprojects?)

Because the `app.manifest` and `Directory.Build.props` are placed wrongly. The templated projects carries there own copies of those. They shouldn't be used for the template packaging project.
However, the wrong `Directory.Build.props` covers the one from repository root. After removing it, shared metadata will come from the root one.

PR #3419, opened 2020-04-02
Ensure animation starts at zero if stopped before load complete

peppy:
Clock not being processed meant it had an incorrect time reference.


smoogipoo:
Ah I forgot that VideoSprite existed, which needs the same change applied.

peppy:
You can't stop a video, so shouldn't apply?

peppy:
I guess there's no harm in adding it.

PR #3422, opened 2020-04-03
Fix animations incorrectly processing a parent clock

peppy:
- Closes https://github.com/ppy/osu/issues/8581

Turns out that Animation was processing the "source" clock a second time. Because in some cases this was the game's thread clock, this cause it to try and sleep as it thought it was running too fast, causing a frame rate decrease when frame limiters were enabled.

Because the animation class' clock setup has caused so much trouble, I've decided to refactor this whole thing a bit.

- New class `AnimationClockComposite` manages clock logic. It now uses a `ManualClock` rather than `FramedOffsetClock`, which tidies up the code and makes the time incrementing feel less hacky.
- Split `IAnimation` into two interfaces (`IAnimation` and `IFramedAnimation`) to allow code share between `Animation` and `VideoSprite`.
- `Animation` / `VideoSprite` no longer have any custom clock logic.
- `VideoSprite` is now named `Video` (the "sprite" is nested within now). This is so it can share clock logic with `AnimationClockComposite`
- As a safety measure, added a `ctor` parameter to `FramedOffsetClock` to make future consumers aware of the processes-by-default behaviour. This is not required by this PR but I think it will prevent future trip-ups on the same behaviour.

I'm hoping this is the final refactor of these classes; they are feeling pretty nailed in now.

I'd like to add some kind of protection on (at least) `GameThread`'s clocks to avoid multiple processes in a single frame, but that can come in a separate PR.

# Breaking Changes

## VideoSprite is now Video

This was renamed for clarity, as it is now not a sprite itself. As a result, videos can now be seeked without using a custom clock.

## Animation.Repeat is now Animation.Loop

Now shares the same interface as VideoSprite

### DIFF
--- a/osu.Framework/Graphics/Animations/AnimationClockComposite.cs
+++ b/osu.Framework/Graphics/Animations/AnimationClockComposite.cs
@@ -1,0 +1,97 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using osu.Framework.Allocation;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Timing;
+
+namespace osu.Framework.Graphics.Animations
+{
+    public abstract class AnimationClockComposite : CompositeDrawable, IAnimation
+    {
+        private readonly bool startAtCurrentTime;
+
+        private ManualClock manualClock;
+
+        /// <summary>
+        /// Construct a new animation.
+        /// </summary>
+        /// <param name="startAtCurrentTime">Whether the current clock time should be assumed as the 0th animation frame.</param>
+        protected AnimationClockComposite(bool startAtCurrentTime = true)
+        {
+            this.startAtCurrentTime = startAtCurrentTime;
+        }
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            base.AddInternal(new Container
+            {
+                RelativeSizeAxes = Axes.Both,
+                Clock = new FramedClock(manualClock = new ManualClock()),
+                Child = CreateContent()
+            });
+        }
+
+        public override IFrameBasedClock Clock
+        {
+            get => base.Clock;
+            set
+            {
+                base.Clock = value;
+
+                manualClock.CurrentTime = !startAtCurrentTime ? Clock.CurrentTime : 0;
+            }
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            if (!startAtCurrentTime)
+                manualClock.CurrentTime = Clock.CurrentTime;
+        }
+
+        protected internal override void AddInternal(Drawable drawable) => throw new InvalidOperationException($"Use {nameof(CreateContent)} instead.");
+
+        /// <summary>
+        /// Create the content container for animation display.
+        /// </summary>
+        /// <returns></returns>
+        public abstract Drawable CreateContent();
+
+        protected override void Update()
+        {
+            base.Update();
+
+            if (IsPlaying)
+                manualClock.CurrentTime += Time.Elapsed;
+        }
+
+        /// <summary>
+        /// The current playback position of the animation, in milliseconds.
+        /// </summary>
+        public double PlaybackPosition
+        {
+            get
+            {
+                if (Loop)
+                    return manualClock.CurrentTime % Duration;
+
+                return Math.Min(manualClock.CurrentTime, Duration);
+            }
+        }
+
+        public double Duration { get; protected set; }
+
+        public bool IsPlaying { get; set; } = true;
+
+        public bool Loop { get; set; }
+
+        public void Seek(double time) => manualClock.CurrentTime = time;
+    }
+}
--- a/osu.Framework/Graphics/Animations/AnimationClockComposite.cs
+++ b/osu.Framework/Graphics/Animations/AnimationClockComposite.cs
@@ -1,9 +1,6 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
-// See the LICENCE file in the repository root for full licence text.
-
 using System;
 using osu.Framework.Allocation;
 using osu.Framework.Graphics.Containers;
@@ -65,7 +62,7 @@ namespace osu.Framework.Graphics.Animations
         /// <summary>
         /// Create the content container for animation display.
         /// </summary>
-        /// <returns></returns>
+        /// <returns>The container providing the content to be added into this <see cref="AnimationClockComposite"/>'s hierarchy.</returns>
         public abstract Drawable CreateContent();
 
         private double lastConsumedTime;
--- a/osu.Framework/Graphics/Animations/AnimationClockComposite.cs
+++ b/osu.Framework/Graphics/Animations/AnimationClockComposite.cs
@@ -15,7 +15,7 @@ namespace osu.Framework.Graphics.Animations
     {
         private readonly bool startAtCurrentTime;
 
-        private ManualClock manualClock;
+        private readonly ManualClock manualClock = new ManualClock();
 
         /// <summary>
         /// Construct a new animation.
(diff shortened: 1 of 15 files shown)


peppy:
Don't merge just yet, issues with initial clock seek in some untested flows.

peppy:
This is good to go now. Note that behaviour on changing clock post-`LoadComplete` has changed now (to continue playback ignoring the clock change) in line with discord discussion.

osu.Framework/Graphics/Animations/IAnimation.cs
```diff
@@ -4,34 +4,39 @@
...
         /// <summary>
-        /// The number of frames this animation has.
+        /// The duration of the video that is being played. Can only be queried after the decoder has started decoding has loaded. This value may be an estimate by FFmpeg, depending on the video loaded.
         /// </summary>
-        int FrameCount { get; }
+        public double Duration { get; }
 
         /// <summary>
-        /// The currently visible frame's index.
+        /// True if the video has finished playing, false otherwise.
```
smoogipoo, line 17:
```suggestion
        /// True if the animation has finished playing, false otherwise.
```

osu.Framework/Graphics/Animations/AnimationClockComposite.cs
```diff
@@ -0,0 +1,115 @@
...
+            if (!startAtCurrentTime && !hasSeeked)
+                manualClock.CurrentTime += elapsed;
+        }
+
+        protected internal override void AddInternal(Drawable drawable) => throw new InvalidOperationException($"Use {nameof(CreateContent)} instead.");
+
+        /// <summary>
+        /// Create the content container for animation display.
+        /// </summary>
+        /// <returns></returns>
```
iiSaLMaN, line 68:
```suggestion
        /// <returns>The container providing the content to be added into this <see cref="AnimationClockComposite"/>'s hierarchy.</returns>
```

osu.Framework/Graphics/Animations/IAnimation.cs
```diff
@@ -4,34 +4,39 @@
...
         /// </summary>
-        int CurrentFrameIndex { get; }
+        public bool FinishedPlaying => !Loop && PlaybackPosition > Duration;
 
         /// <summary>
-        /// True if the animation is playing, false otherwise.
+        /// True if the animation is playing, false otherwise. Starts true.
```
smoogipoo, line 22:
```suggestion
        /// True if the animation is playing, false otherwise. <c>true</c> by default.
```

osu.Framework/Graphics/Animations/AnimationExtensions.cs
```diff
@@ -4,7 +4,7 @@
 namespace osu.Framework.Graphics.Animations
 {
     /// <summary>
-    /// This class holds various extension methods for the <see cref="IAnimation"/> interface.
+    /// This class holds various extension methods for the <see cref="IFramedAnimation"/> interface.
     /// </summary>
     public static class AnimationExtensions
```
iiSaLMaN, line 9: The changes in this file technically result in a breaking change

iiSaLMaN, line 9: Can be renamed to `IFramedAnimationExtensions` now.

osu.Framework/Graphics/Animations/IAnimation.cs
```diff
@@ -4,34 +4,39 @@
...
 {
     /// <summary>
-    /// Represents a generic, frame-based animation.
+    /// An animation / playback sequence.
     /// </summary>
     public interface IAnimation
     {
         /// <summary>
-        /// The number of frames this animation has.
+        /// The duration of the video that is being played. Can only be queried after the decoder has started decoding has loaded. This value may be an estimate by FFmpeg, depending on the video loaded.
```
smoogipoo, line 12:
```suggestion
        /// The duration of the animation. Can only be queried after the decoder has started decoding has loaded. This value may be an estimate by FFmpeg, depending on the video loaded.
```

Also the ffmpeg blurb is a little weird to me in an interface like this, because it's not true for `Animation`.

How about making it virtual and override in Video to add the comment/possibly an exception later down the track?

osu.Framework/Graphics/Animations/IFramedAnimation.cs
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
```
smoogipoo, line 5:
```suggestion
```

osu.Framework/Graphics/Animations/AnimationClockComposite.cs
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
```
smoogipoo, line 5:
```suggestion
```

osu.Framework/Graphics/Video/VideoSprite.cs
```diff
@@ -1,254 +1,33 @@
...
-using osu.Framework.Timing;
+using osuTK;
 
 namespace osu.Framework.Graphics.Video
 {
     /// <summary>
-    /// Represents a sprite that plays back a video from a stream or a file.
+    /// A sprite which holds a video with a custom conversion matrix. Use <see cref="Video"/> for loading and displaying a video.
     /// </summary>
     public class VideoSprite : Sprite
```
smoogipoo, line 15: This isn't actually usable by itself now right? Is there a reason why you added the blurb to the xmldoc rather than making this an internal, or even private class within `Video`?

osu.Framework/Graphics/Video/Video.cs
```diff
@@ -0,0 +1,178 @@
...
+            {
+                if (decoder != null)
+                    decoder.Looping = value;
+
+                base.Loop = value;
+            }
+        }
+
+        /// <summary>
+        /// True if this VideoSprites decoding process has faulted.
```
smoogipoo, line 43:
```suggestion
        /// True if this <see cref="Video"/>'s decoding process has faulted.
```

osu.Framework/Graphics/Video/Video.cs
```diff
@@ -0,0 +1,178 @@
...
+using osu.Framework.Bindables;
+using osu.Framework.Graphics.Animations;
+using osu.Framework.Logging;
+using osu.Framework.Platform;
+using osu.Framework.Graphics.Shaders;
+
+namespace osu.Framework.Graphics.Video
+{
+    /// <summary>
+    /// Represents a sprite that plays back a video from a stream or a file.
```
smoogipoo, line 18:
```suggestion
    /// Represents a <see cref="Drawable"/> that plays back a video from a stream or a file.
```

osu.Framework/Graphics/Animations/AnimationClockComposite.cs
smoogipoo, line 68:
```suggestion
```

osu.Framework/Graphics/Animations/IAnimation.cs
peppy: Oh, the ffmpeg part should really be removed. Now the duration is always set at a specific point in time (bdl) regardless.

osu.Framework/Graphics/Animations/AnimationExtensions.cs
peppy: yes, this PR introduced multiple breaking changes, including renaming a class too.

osu.Framework/Graphics/Video/VideoSprite.cs
peppy: I tried `internal` but it didn't work due to the `DrawNode` prerequisite and permission differences. Same reason for not nesting (it will get large). Maybe there's a way around this?

smoogipoo: I suggest making the drawnode take in a `Video` instead, and move the conversion matrix all the way to `Video`. At that point `VideoSprite` becomes a very dumb class that just has the bdl for the shaders, so it could be nested.

peppy: Actually that was enough to make it `internal` without nesting, check pushed changes.

smoogipoo:
This seems like it has somehow broken hit explosions osu!-side, which are supposed to apply additively but aren't anymore.

smoogipoo:
Looks like blending for us has been completely broken beyond one drawable level...

PR #3427, opened 2020-04-05
Expose displays as bindables through IWindow

swoolcock:
The primary purpose of this PR is to expose the current display as a bindable through `IWindow`.  It will provide us with the ability to add a display selection `Dropdown` in the osu! settings if desired, and is a required step in providing proper display mode selection in the SDL backend.

#### Breaking changes for osu!
* `LayoutSettings` will need to change to read from `DisplayModes` rather than `AvailableResolutions` https://github.com/ppy/osu/pull/8623

#### Potential breaking changes for other framework consumers
* `GameWindow.CurrentDisplay` renamed to `CurrentDisplayDevice` to avoid conflicts with the new bindable.

#### Non-breaking changes (since these were not being used outside of the SDL backend)
* `IWindow.Display` renamed to `CurrentDisplay` as a mutable bindable.
* `IWindow.DisplayMode` renamed to `CurrentDisplayMode`.

#### Notes
`GameWindow` now has an internal `Scheduler` on `UpdateFrame` so that it can safely change the current display from the bindable's `ValueChanged` event.  Doing this directly would cause the window to snap to the wrong position and wreak havoc with the cursor offset.  This event is subscribed to in `DesktopGameWindow`, since the `GameWindow.CurrentDisplayDevice` setter throws an exception.

Needs testing on Windows and Linux, since exclusive fullscreen is not currently implemented on macOS.


peppy:
The rename of `CurrentDisplay` to `CurrentDisplayDevice` feels quite arbitrary. Is there a reason for now just using the bindable everywhere?

swoolcock:
> The rename of `CurrentDisplay` to `CurrentDisplayDevice` feels quite arbitrary. Is there a reason for now just using the bindable everywhere?

There are quite a few places that refer to the osuTK `DisplayDevice` type directly, and I wanted to avoid frequent conversion back and forth from `Display` if possible.  I was hoping to avoid changing the inner workings of `GameWindow` too much if we can, but I can certainly use the bindable everywhere if you'd prefer.

peppy:
Oh, they are completely different types..

osu.Framework/Platform/GameWindow.cs
```diff
@@ -225,10 +241,17 @@ public bool CursorGrabbed
...
             get => DisplayDevice.FromRectangle(Bounds) ?? DisplayDevice.Default;
-            set => throw new InvalidOperationException($@"{GetType().Name}.{nameof(CurrentDisplay)} cannot be set.");
+            set => throw new InvalidOperationException($@"{GetType().Name}.{nameof(CurrentDisplayDevice)} cannot be set.");
+        }
+
+        private void checkCurrentDisplay()
+        {
+            int index = (int)CurrentDisplayDevice.GetIndex();
+            if (index != CurrentDisplay.Value?.Index)
+                CurrentDisplay.Value = AvailableDisplays.Skip(index).FirstOrDefault();
```
turbedi, line 254:
```suggestion
                CurrentDisplay.Value = AvailableDisplays.ElementAtOrDefault(index);
```

osu.Framework/Platform/Window.cs
```diff
@@ -277,6 +291,8 @@ public Window(IWindowBackend windowBackend, IGraphicsBackend graphicsBackend)
 
             windowBackend.Create();
 
+            displays.AddRange(windowBackend.Displays);
```
smoogipoo, line 294: Yeah same here, I'd really just expect this to reference the window backend directly rather than storing them, which seems like it could work if displays just return an IEnumerable rather than a bindable list.

osu.Framework/Platform/Window.cs
```diff
@@ -300,7 +316,12 @@ public Window(IWindowBackend windowBackend, IGraphicsBackend graphicsBackend)
             windowBackend.MouseWheel += OnMouseWheel;
             windowBackend.DragDrop += OnDragDrop;
 
+            windowBackend.DisplayChanged += d => CurrentDisplay.Value = d;
+
             graphicsBackend.Initialise(windowBackend);
+
+            CurrentDisplay.Value = windowBackend.CurrentDisplay;
+            CurrentDisplay.ValueChanged += evt => this.windowBackend.CurrentDisplay = evt.NewValue;
```
smoogipoo, line 324:
```suggestion
            CurrentDisplay.ValueChanged += evt => windowBackend.CurrentDisplay = evt.NewValue;
```

Can probably be removed?

osu.Framework/Platform/GameWindow.cs
```diff
@@ -71,21 +74,34 @@ public abstract class GameWindow : IWindow
         /// </summary>
         public IBindable<bool> IsActive => isActive;
 
-        public virtual IEnumerable<Display> Displays => new[] { Display };
+        private readonly BindableList<Display> displays = new BindableList<Display>();
+
+        public virtual IBindableList<Display> Displays => displays;
+
+        private readonly Display[] availableDisplays;
```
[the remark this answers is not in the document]
swoolcock: SDL (and I believe OpenTK) will only ever report the displays and display modes that existed when the app started, regardless of adding/removing at runtime.  I thought it was safe enough to just cache.

osu.Framework/Platform/GameWindow.cs
```diff
@@ -71,21 +74,34 @@ public abstract class GameWindow : IWindow
         /// </summary>
         public IBindable<bool> IsActive => isActive;
 
-        public virtual IEnumerable<Display> Displays => new[] { Display };
+        private readonly BindableList<Display> displays = new BindableList<Display>();
+
+        public virtual IBindableList<Display> Displays => displays;
```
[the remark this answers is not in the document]
swoolcock: `BindableList` so that it can be used as the `ItemSource` in a `Dropdown`.  Since the values never change it'd be fine to replace this with an enumerable and just `AddRange` it.

osu.Framework/Platform/GameWindow.cs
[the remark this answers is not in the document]
swoolcock: Note that `Display` is a class, not a struct.  I didn't want to be creating new instances on every call.

osu.Framework/Platform/GameWindow.cs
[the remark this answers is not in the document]
smoogipoo: `Dropdown` also supports `Items`.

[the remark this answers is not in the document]
swoolcock: Yep, will change to `IEnumerable<Display> Displays`.

osu.Framework/Platform/GameWindow.cs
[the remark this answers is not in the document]
smoogipoo: We can reassess memory usage when there's a reason for this to be invoked very frequently (i.e. every frame) outside of tests.
It's a small class as is, and if it cleans things up by not caching then I'd rather do that.

osu.Framework/Platform/GameWindow.cs
```diff
@@ -71,21 +74,34 @@ public abstract class GameWindow : IWindow
...
+
+        private readonly Display[] availableDisplays;
+
+        /// <summary>
+        /// Provides an <see cref="IEnumerable{Display}"/> that will initialise the cached array
+        /// of available displays.
+        /// Defaults to an empty enumerable.
+        /// </summary>
+        protected virtual IEnumerable<Display> AvailableDisplays => Enumerable.Empty<Display>();
```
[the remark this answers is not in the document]
swoolcock: Removed, since displays are no longer cached.

osu.Framework/Platform/GameWindow.cs
```diff
@@ -71,21 +74,23 @@ public abstract class GameWindow : IWindow
         /// </summary>
         public IBindable<bool> IsActive => isActive;
 
-        public virtual IEnumerable<Display> Displays => new[] { Display };
+        public virtual IEnumerable<Display> Displays => new[] { DisplayDevice.GetDisplay(DisplayIndex.Primary).ToDisplay() };
```
smoogipoo, line 77: What happens if osu! is started on a secondary display? It seems like previously this would return the screen the window is currently on, whereas this has changed that?

Is this ever used osu!-side, where that could be an issue?

osu.Framework/Platform/GameWindow.cs
```diff
@@ -98,6 +103,11 @@ protected GameWindow([NotNull] IGameWindow implementation)
             Implementation = implementation;
             Implementation.KeyDown += OnKeyDown;
 
+            CurrentDisplay.Value = PrimaryDisplay;
```
smoogipoo, line 106: Are you sure we don't want to `checkCurrentDisplay` instead? This also seems to have changed compared to previous.

osu.Framework/Platform/DesktopGameWindow.cs
```diff
@@ -73,14 +73,16 @@ public override DisplayDevice CurrentDisplay
...
+                      .Select(ExtensionMethods.ToDisplay);
 
         protected DesktopGameWindow()
             : base(default_width, default_height)
         {
             Resize += OnResize;
             Move += OnMove;
+
+            CurrentDisplay.ValueChanged += evt =>
+                UpdateFrameScheduler.Add(() => CurrentDisplayDevice = DisplayDevice.GetDisplay((DisplayIndex)evt.NewValue.Index));
```
smoogipoo, line 85: Relating to the previous comments, I believe this has also changed since it's going to be set to Primary on startup? I.e. the window will always start on the primary display?

osu.Framework/Platform/GameWindow.cs
swoolcock: This is overridden by `DesktopGameWindow`.  It would actually be more correct for this to either return nothing, or to be abstract and then implement it in the mobile subclasses.

I thought it would be better to return something that's guaranteed to exist regardless of platform.

osu.Framework/Platform/GameWindow.cs
swoolcock: The current display is changed in `SetupWindow` to be correct for the current configuration.

osu.Framework/Platform/DesktopGameWindow.cs
swoolcock: See previous comment re: `SetupWindow`.

PR #3428, opened 2020-04-06
Add temporary blending "fix" for BufferedContainer

smoogipoo:
"Regressed" from https://github.com/ppy/osu-framework/pull/3423, because blending is now handled correctly through multiple parents.

This is temporary because `BufferedContainer` blending is 100% messed up. It's incorrect because it's blending a pre-multiplied alpha texture as if it isn't pre-multiplied.
Over the last 3 days I've tried to get it to blend correctly, but it isn't an easy feat - premultiplying itself is theoretically very easy, but there's some weirdness going on inside the hierarchy of `BufferedContainer` that I need to investigate more.

For now, this restores the original behaviour prior to the PR mentioned above, in the hope that it'll be removed in a short while. This fixes the results screen.

Haven't written a test for this because there's really nothing of value to test here - it's completely incorrect.

### DIFF
--- a/osu.Framework/Graphics/Containers/BufferedContainer.cs
+++ b/osu.Framework/Graphics/Containers/BufferedContainer.cs
@@ -347,7 +347,17 @@ namespace osu.Framework.Graphics.Containers
         public DrawColourInfo? FrameBufferDrawColour => base.DrawColourInfo;
 
         // Children should not receive the true colour to avoid colour doubling when the frame-buffers are rendered to the back-buffer.
-        public override DrawColourInfo DrawColourInfo => new DrawColourInfo(Color4.White, base.DrawColourInfo.Blending);
+        public override DrawColourInfo DrawColourInfo
+        {
+            get
+            {
+                // Todo: This is incorrect.
+                var blending = Blending;
+                blending.ApplyDefaultToInherited();
+
+                return new DrawColourInfo(Color4.White, blending);
+            }
+        }
 
         protected override void Dispose(bool isDisposing)
         {


smoogipoo:
Explanation of the OP.

The results screen is broken because it applies a subtractive blending at a parent level. Due to the previous fix, children of the BCs receive the subtractive mode and subtract from the BC's framebuffer.
This would be compounded if `EffectBlending` actually respected the parent's blending mode, but it only changes if you set `Blending` or `EffectBlending` on the BC itself. This is a bug:

https://github.com/ppy/osu-framework/blob/fb601acc4d8dcc2ff5b60584fe64e440ac9cb6f2/osu.Framework/Graphics/Containers/BufferedContainer.cs#L328-L334

If this were fixed, there would be another subtraction blend between the BC and the backbuffer, which I believe is what we want.

As an initial idea, I thought that we'll want BC to always pass a `Mixture` blending mode to its children (which is essentially what this PR achieves, assuming the blending is set on a parent), however that breaks additive blending.
My idea for `BufferedContainer` and `Container` is that they should be completely interchangeable with no visual differences, meaning since children in a `Container` additively blend between themselves, so should children of a `BufferedContainer`, if the blending is set from outside.

As for why BC blending is completely broken, here's a few examples (left: `Container`, the expected output, right: `BufferedContainer`)

| blending mode / framebuffer background | mixture | additive |
| --- | --- | --- |
| transparent black (default) | image | image |
| transparent white | image | image |
| transparent yellow | image | image |

As you can see, it is impossible to achieve the same blending as `Container`, _unless_ you set the background of the BC to 100% opaque yellow, at which point the background box itself may as well not exist. This is undesirable in pretty much all occasions for us - we never have a BC on a single-colour background like this.

With pre-multiplied blending we can achieve the following results:

| mixture | additive |
| --- | --- |
| image | image |

Here's a further example with a gradient:
| blending mode / type | mixture | additive |
| --- | --- | --- |
| current | image | image |
| pre-multiplied | image | image |

As you can see, we can achieve _perfect_ blending using pre-multiplied alpha, and best of all this doesn't break the API between `Container` and `BufferedContainer` - children are still drawn in additive blending, unlike what this PR practically does.

Code for this:
```cs
[BackgroundDependencyLoader]
private void load()
{
    Children = new Drawable[]
    {
        new Container
        {
            Anchor = Anchor.Centre,
            Origin = Anchor.Centre,
            Size = new Vector2(200),
            X = -100,
            Children = new Drawable[]
            {
                new Box
                {
                    RelativeSizeAxes = Axes.Both,
                    Colour = ColourInfo.GradientVertical(Color4.Yellow, Color4.Blue)
                },
                new Container
                {
                    RelativeSizeAxes = Axes.Both,
                    Blending = BlendingParameters.Additive,
                    Children = new Drawable[]
                    {
                        new Box
                        {
                            Anchor = Anchor.Centre,
                            Origin = Anchor.Centre,
                            RelativeSizeAxes = Axes.Both,
                            Size = new Vector2(0.5f, 1),
                            Colour = Color4.Cyan,
                            Alpha = 0.5f,
                        },
                        new Box
                        {
                            Anchor = Anchor.Centre,
                            Origin = Anchor.Centre,
                            RelativeSizeAxes = Axes.Both,
                            Size = new Vector2(0.25f, 1),
                            Colour = Color4.Magenta,
                            Alpha = 0.5f
                        }
                    }
                }
            }
        },
        new Container
        {
            Anchor = Anchor.Centre,
            Origin = Anchor.Centre,
            Size = new Vector2(200),
            X = 100,
            Children = new Drawable[]
            {
                new Box
                {
                    RelativeSizeAxes = Axes.Both,
                    Colour = ColourInfo.GradientVertical(Color4.Yellow, Color4.Blue)
                },
                new BufferedContainer
                {
                    RelativeSizeAxes = Axes.Both,
                    // Required if not doing the blending tricks below.
                    // BackgroundColour = Color4.Yellow.Opacity(0),
                    // Blending = BlendingParameters.Additive,

                    // // Disable the blending tricks if using the above.
                    // // Use premultiplied alpha to draw to the screen
                    EffectBlending = new BlendingParameters
                    {
                        Source = BlendingType.One,

                        // Additive
                        Destination = BlendingType.One,
                        // Mixture
                        // Destination = BlendingType.OneMinusSrcAlpha,
                    },
                    // Set non-premultiplied alpha for children
                    Blending = new BlendingParameters
                    {
                        Source = BlendingType.SrcAlpha,

                        // Additive
                        Destination = BlendingType.One,
                        // Mixture
                        // Destination = BlendingType.OneMinusSrcAlpha

                        // The defaults are incorrect here
                        SourceAlpha = BlendingType.One,
                        DestinationAlpha = BlendingType.OneMinusSrcAlpha
                    },
                    Children = new Drawable[]
                    {
                        new Box
                        {
                            Anchor = Anchor.Centre,
                            Origin = Anchor.Centre,
                            RelativeSizeAxes = Axes.Both,
                            Size = new Vector2(0.5f, 1),
                            Colour = Color4.Cyan,
                            Alpha = 0.5f,
                        },
                        new Box
                        {
                            Anchor = Anchor.Centre,
                            Origin = Anchor.Centre,
                            RelativeSizeAxes = Axes.Both,
                            Size = new Vector2(0.25f, 1),
                            Colour = Color4.Magenta,
                            Alpha = 0.5f,
                        }
                    }
                }
            }
        }
    };
}
```

PR #3434, opened 2020-04-07
Include base test scene types in dynamic compilation

peppy:


peppy:
Tested randomly against osu! and framework tests w/dynamic-recompilation to ensure no regression. Also tested against all osu!-side tests.

PR #3435, opened 2020-04-07
Transfer FillAspectRatio outwards to fix Animation/Video fill modes scaling incorrectly

peppy:


osu.Framework/Graphics/Video/Video.cs
```diff
@@ -182,13 +182,7 @@ protected override void Dispose(bool isDisposing)
                 f.Texture.Dispose();
         }
 
-        protected override void UpdateSizing()
-        {
-            base.UpdateSizing();
-
-            // transfer fill mode to handle as one would expect.
-            Sprite.FillMode = FillMode;
```
smoogipoo, line 190: Also removable from TextureAnimation.

PR #3436, opened 2020-04-07
Expose original instance for dynamic compilation

peppy:
For scenarios like needing access to the original assembly for accessing resources.


peppy:
Open to better variable naming and potentially not including in interface.

PR #3450, opened 2020-04-14
Extract bindable propagations

UselessToucan:
As a XAML developer, I expect binding to be more powerful than it is now. For example, in UWP DataBinding supports multiple modes(one way, two way, one time) and converters(custom value processing during its propagation).

This PR extracts `Value`/`Default`/`Disabled` propagation to separate class hierarchy(`Binding<T>`). It allows us to implement various propagation algorithms without messing up the `Bindable<T>` itself.
In terms of design patterns, `Binding<T>` is a strategy. It encapsulates a propagation algorithm and theoretically can be passed to a `Bindable` from the outside.

### DIFF
--- a/osu.Framework.Tests/Bindables/BindableBindingTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableBindingTest.cs
@@ -5,6 +5,7 @@ using System;
 using NUnit.Framework;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
+using osu.Framework.Bindables.Bindings;
 using osu.Framework.Graphics;
 using osu.Framework.Timing;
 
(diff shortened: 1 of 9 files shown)


bdach:
I've done some WPF, and from what I remember `TwoWay` I would mostly use for synchronising view models with the actual controls with the whole `BindValueChanged` shebang, which isn't required in our case as our control properties usually don't need synchronising like that. Honestly I'd call the system WPF had a little byzantine and overly verbose in places (implementing a `DependencyProperty` was *bad*).

I think over the entire time I've been reviewing PRs in framework and lazer I've seen maybe one instance where a two-way bindable would be handy, and even then you could do fine with just one-way ones (it was #3138). I'm not sure the added complexity of having two-way bindables is worth it, especially with the danger of cyclical updates/conflicts (what if both ends of the bindable change value in the same frame? what's the reconciliation algorithm there? or is it just going to race?). I'd be much more comfortable with acyclical bindable flow as it's much easier to follow.

Could you give an example of a scenario in which you'd want a two-way bindable? Also, I would be very interested in seeing some tests for the two-way bindings.

jorolf:
Aren't Bindables already 2-way? https://github.com/ppy/osu-framework/blob/master/osu.Framework/Bindables/Bindable.cs#L196

bdach:
Huh, I suppose they are. That's kind of embarrassing.

PR #3451, opened 2020-04-14
Change ManualThreadRunner.HasRunOnce type from bool to ManualResetEventSlim

UselessToucan:


osu.Framework.Benchmarks/GameBenchmark.cs
```diff
@@ -64,8 +64,7 @@ public ManualGameHost(Game runnableGame)
                 });
 
                 // wait for the game to initialise before continuing with the benchmark process.
-                while (threadRunner?.HasRunOnce != true)
-                    Thread.Sleep(10);
+                threadRunner?.HasRunOnce.Wait();
```
smoogipoo, line 67: What about the case where the thread runner is null?

smoogipoo:
Closing as nothing is gained from this PR.

Issue #3243, opened 2020-02-06
Move template game into main solution


peppy:
Will allow for better maintenance of it.

PR #3458, opened 2020-04-17
Add template projects to sln

smoogipoo:
An interim solution so that we can keep the template projects somewhat up to date - dependabot will catch out of date packages for us. Seems to work on my fork: https://github.com/smoogipoo/osu-framework/pulls

I've put the templates into separate sln folders since they're 5 projects worth each, with the downside being the filtered solutions now get a few empty folders...


smoogipoo:
Not sure if I should mark https://github.com/ppy/osu-framework/issues/3243 as being resolved by this. I'd still like to see these projects compiled along with the working set...

PR #3480, opened 2020-04-23
Disable blending in front-to-back pass

smoogipoo:
Could lead GPU heuristics to disable early-z if blending is enabled.

I gain no performance benefit on FirePro D300 (although FTB does seem to give performance benefits in framework-side tests), but I'd be interested to see if this helps on iOS which is definitely a tiling renderer.


peppy:
Is front-to-back not hard-disabled on iOS currently?

Tom94:
IIRC it is... because it's slow (presumably, because iOS is not smart enough to do early z-testing). As far as I understand, smoogi's change is an attempt to encourage iOS to actually _do_ early z-testing, thereby making front-to-back worthwhile.

peppy:
Ah, I forgot there's still a settings toggle which can be used to test this. I guess that's the intention here?

smoogipoo:
Yeah, will enable when/if it provides any benefit but not preemptively.

PR #3483, opened 2020-04-24
Rework dynamic recompilation to use PE/PDB files

smoogipoo:
PRing for visibility for now as a possible path forward. This should lead us to lead to exact matches even in the case of partials / generics / etc.

The PE file is used to search for method signatures. The PDB then matches those signatures with documents.

Issues:
- [ ] Classes (including test scenes) must have at least one method. A property counts as a method, and so does an override - the class just cannot be effectively useless. Potentially have to fall-back to by-filename matching if this is the case, although it may be possible to get more information out of the PDB file...
- [ ] The above shows why `compileFinished` can nullref and crash the whole browser. Should be fixed.


smoogipoo:
Going to close this one for now. While it's an interesting path, I have already made plans for a different system and will likely not continue working on this.

PR #3484, opened 2020-04-24
Fix dynamic compilation auto-include not working

smoogipoo:


smoogipoo:
Ah, namespace is incorrect :/

PR #3486, opened 2020-04-25
Add tree connection lines to make draw visualiser easier to visually parse

### LINKED ISSUE #3462: Draw visualiser should draw a line from child to parent
Right now the only way to see what is the child's parent is to hover over it in the visualizer. This means you can't at a glance understand a complicated scene hierarchy.

Current:
current

 
Proposed (there are probably lines that would extend outside to the bottom, I just wouldn't know where to draw them):
proposal

peppy:
- Closes #3462 

image

Best effort for minimal implementation.


smoogipoo:
Refactored because topLevel wasn't the correct way of going about it and broke if `up one parent` was used.

PR #3494, opened 2020-04-29
SDL2# Backend Refactor

swoolcock:
Initial implementation of SDL2 backend refactor using SDL2#.

Implemented:
- [x] Mouse
- [x] Keyboard
- [x] Window events
- [x] Drag and drop

Not implemented:
- Joystick
- Raw input

Get sdl2-cs from NuGet source: https://ci.appveyor.com/nuget/sdl2-cs-kcu6ggyov55m

From what I'm reading, `SDL_SetRelativeMouseMode` *should* handle mouse raw input by default on Windows and Linux, but we will need to do our own thing for macOS.

`OnCloseRequested` is currently unimplemented as it causes a deadlock in `GameHost.OnExitRequested`.  I'll look into this in due course.

Currently the only framework setting implemented is `FrameworkSetting.WindowedSize`.  Next PR will likely contain implementations of `WindowMode`, `SizeFullscreen`, etc.  I wanted to keep the PR as small as possible to start with.

One notable difference between SDL and osuTK is that the window size on macOS is measured in logical dimensions (hence the `Scale` property).  This needs to be calculated per display after the window has been created (using `SDL_GL_GetDrawableSize`).  There's somewhat of a catch-22 here in that we need to create the window (on the correct display) before we know how large to make it.

As such, `WindowedSize` now refers to the logical window size rather than the internal renderable size.  On Windows this shouldn't make a difference as `SDL_GL_GetDrawableSize` seems to return 1 even with window scaling, but on macOS it means that switching between the osuTK and SDL backends will double/halve the window size each time.  This is something that we can look at in a near future PR.

Mutating functions/properties are deferred to the windowing thread via a `Scheduler` to prevent crashes and nasty stack traces like we currently get on macOS/osuTK when changing window state.

Needs testing on Windows and Linux.

@iiSaLMaN @bdach @Xyene
Would appreciate your input/testing in this.

Note: may need some TLC, mostly pushed for testing and visibility.


osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -0,0 +1,638 @@
...
+
+        private bool cursorConfined;
+
+        public bool CursorConfined
+        {
+            get => SdlWindowHandle == IntPtr.Zero ? cursorConfined : windowFlags.HasFlag(SDL.SDL_WindowFlags.SDL_WINDOW_MOUSE_CAPTURE);
+            set
+            {
+                cursorConfined = value;
+                scheduler.Add(() => SDL.SDL_CaptureMouse(value ? SDL.SDL_bool.SDL_TRUE : SDL.SDL_bool.SDL_FALSE));
```
Xyene, line 147: I think `SDL_CaptureMouse` is the wrong call here. From https://wiki.libsdl.org/SDL_CaptureMouse:

> When capturing is enabled, the current window will get all mouse events, but unlike relative mode, **no change is made to the cursor and it is not restrained to your window.** 

(emphasis mine)

I think https://wiki.libsdl.org/SDL_SetWindowGrab is wanted instead. The naming is pretty unfortunate, OpenTK fell for the same trap: https://github.com/ppy/osuTK/pull/64

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -0,0 +1,638 @@
...
+                cursorVisible = value;
+                scheduler.Add(() => SDL.SDL_ShowCursor(value ? SDL.SDL_ENABLE : SDL.SDL_DISABLE));
+            }
+        }
+
+        private bool cursorConfined;
+
+        public bool CursorConfined
+        {
+            get => SdlWindowHandle == IntPtr.Zero ? cursorConfined : windowFlags.HasFlag(SDL.SDL_WindowFlags.SDL_WINDOW_MOUSE_CAPTURE);
```
Xyene, line 143: Due to the same reason as below, I think this should be `SDL_WINDOW_INPUT_GRABBED`.

Xyene:
(This is likely an SDL2# issue, but I'll leave it here so it's in one place...)

The search path for `libSDL2.so` is "weird", in that it doesn't match whatever osuTK was doing. Out-of-the-box, it fails with:

```
Unhandled exception. System.DllNotFoundException: Unable to load shared library 'SDL2' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: libSDL2: cannot open shared object file: No such file or directory
   at SDL2.SDL.SDL_Init(UInt32 flags)
   at osu.Framework.Platform.Sdl.Sdl2WindowBackend..ctor() in /code/osu-framework/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs:line 284
   at osu.Framework.Platform.DesktopWindow..ctor() in /code/osu-framework/osu.Framework/Platform/DesktopWindow.cs:line 19
   at osu.Framework.Platform.Linux.LinuxGameHost.CreateWindow() in /code/osu-framework/osu.Framework/Platform/Linux/LinuxGameHost.cs:line 26
   at osu.Framework.Platform.GameHost.Run(Game game) in /code/osu-framework/osu.Framework/Platform/GameHost.cs:line 520
   at osu.Desktop.Program.Main(String[] args) in /code/osu/osu.Desktop/Program.cs:line 57
```

Running with `LD_DEBUG=files` shows:

```
      2825:	find library=libSDL2 [0]; searching
      2825:	search cache=/etc/ld.so.cache
      2825:	search path=/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/lib:/usr/lib		(system search path)
      2825:	 trying file=/lib/x86_64-linux-gnu/libSDL2
      2825:	 trying file=/usr/lib/x86_64-linux-gnu/libSDL2
      2825:	 trying file=/lib/libSDL2
      2825:	 trying file=/usr/lib/libSDL2
```

In particular, it's not searching `/usr/local/lib`, where my `libSDL2.so` is installed. On my system (a standard Debian install), the search paths should be:

```
$ cat /etc/ld.so.conf.d/*
/usr/lib/x86_64-linux-gnu/libfakeroot
# Multiarch support
/usr/local/lib/i386-linux-gnu
/lib/i386-linux-gnu
/usr/lib/i386-linux-gnu
/usr/local/lib/i686-linux-gnu
/lib/i686-linux-gnu
/usr/lib/i686-linux-gnu
# libc default configuration
/usr/local/lib
# Multiarch support
/usr/local/lib/x86_64-linux-gnu
/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu
# Legacy biarch compatibility support
/lib32
/usr/lib32
```

Workaround:

```shell
$ sudo ln -s /usr/local/lib/libSDL2.so /usr/lib/libSDL2.so
```

swoolcock:
> In particular, it's not searching `/usr/local/lib`, where my `libSDL2.so` is installed. On my system (a standard Debian install), the search paths should be:

Would prefer that SDL2 be bundled with our fork of SDL2# as it is with macOS and Windows.  If you could make a PR here adding the native lib(s) and updating app.config, it would be much appreciated.  :)

https://github.com/ppy/SDL2-CS

osu.Framework/Platform/Sdl/Sdl2Extensions.cs
```diff
@@ -0,0 +1,479 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using osuTK.Input;
+using SDL2;
+
+namespace osu.Framework.Platform.Sdl
+{
+    public static class Sdl2Extensions
+    {
+        public static Key ToKey(this SDL.SDL_Keysym sdlKeysym)
```
iiSaLMaN, line 11: deserves a todo comment removing this after SDL migration

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
swoolcock: Done, however the macOS implementation of `SDL_SetWindowGrab` (and osuTK) is less than perfect.  It warps the mouse cursor back inside the window if it attempts to exit, leaving a noticeable "bouncing" effect.

Additionally, part of the window frame is considered "in-bounds", so it's possible to resize or move the window while confined.

I don't think there's anything we can do here, but it's never worked properly with osuTK either.

swoolcock:
Mmmmm I'll look into the draw rate, it's going to be due to the buffer swap being scheduled to the windowing thread.  I'll verify whether this is actually required, and if not it can be executed immediately.

Xyene:
I haven't looked into the cause yet, but the OS cursor isn't being hidden, and shows up under/over the game-drawn one.

swoolcock:
Cursor hiding is known issue, requires a minor change in osu.Desktop.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
Xyene: [This ticket](https://bugzilla.libsdl.org/show_bug.cgi?id=3836) seems relevant, don't know if you've come across it before.

> If you don't plan on distributing your app in the Mac App Store you can compile SDL2 with SDL_MAC_NO_SANDBOX defined, which allows SDL to use a better method of making sure the cursor doesn't leave the window.
>
> It's also worth mentioning that SDL_SetRelativeMouseMode should be used when possible instead of SDL_SetWindowGrab, when feasible (e.g. when doing mouse movement tracking for a FPS game). The former avoids most of the issues SetWindowGrab may have.

(...or if the plan is to distribute osu! through the app store.)

smoogipoo: We won't be distributing osu! through the mac app store, no.

smoogipoo:
Looks fine up to this point on a quick code pass. Only tested on macOS but it seems there's others that have tested on windows.

bdach:
I can probably give this a test drive on Windows sometime this weekend if it doesn't get merged anyway before I get to it.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
Xyene: :+1: in that case, if it was desired, the GH action infra from the Linux build PR could be used to produce Mac OS builds with `SDL_MAC_NO_SANDBOX` defined, too.

swoolcock: Need to test that that define actually helps, though.

Xyene: Some tests (thanks to @WallE256 for running them):
* [Confine with default SDL on macOS](https://drive.google.com/file/d/1Pe14op9AQxdfDN4vbK0GS0SMsnex0o4x/view?usp=sharing)
* [Confine with SDL compiled with `SDL_MAC_NO_SANDBOX` on macOS](https://drive.google.com/file/d/1A1Iaxv_dmpPKMoiT5fjlmXOsnCGalVuY/view?usp=sharing)
* [`libSDL.dylib` compiled with `SDL_MAC_NO_SANDBOX`](https://drive.google.com/open?id=1jVEW7jggJlC_6nYMTmPvrhdK5v7rbTkk)

It looks like it's not *perfect*, but much better.

Xyene:
Another thing I noticed (don't know if it's expected at this point, though), is that the game no longer pauses on focus loss, regardless of whether it's windowed/borderless/fullscreen.

bdach:
> Another thing I noticed (don't know if it's expected at this point, though), is that the game no longer pauses on focus loss, regardless of whether it's windowed/borderless/fullscreen.

For what it's worth, that was the case for the [initial veldrid implementation](https://github.com/ppy/osu-framework/pull/3013#issuecomment-558846403), too. Probably just some stuff not hooked up yet.

Xyene:
Got it, I was using SDL through osuTK previously, so that would explain the discrepancy :+1:

swoolcock:
> Probably just some stuff not hooked up yet.

Correct, there's a few things that haven't been hooked up that will come in future PRs.

peppy:
Is this supposed to be passing CI? Should there be a dependency on some kind of SDL-CS publishing mechanism?

swoolcock:
> Is this supposed to be passing CI? Should there be a dependency on some kind of SDL-CS publishing mechanism?

Yeah, since SDL2-CS is not published, it's currently relying on the developer adding our AppVeyor feed as a source.  Will need to publish that before merging.

osu.Framework/Platform/Sdl/Sdl2Extensions.cs
```diff
@@ -0,0 +1,479 @@
...
+                case SDL.SDL_Keycode.SDLK_8:
+                    return Key.Number8;
+
+                case SDL.SDL_Keycode.SDLK_9:
+                    return Key.Number9;
+
+                case SDL.SDL_Keycode.SDLK_SEMICOLON:
+                    return Key.Semicolon;
+
+                case SDL.SDL_Keycode.SDLK_EQUALS:
```
turbedi, line 115: Is this intended? Or should it be `case SDL.SDL_Keycode.SDLK_PLUS` ?

swoolcock: According to the SDL wiki, `SDLK_PLUS` has no corresponding scancode, and the various key mappers in osuTK all map it the way I have.  Note that there is no enum case for `Equals` other than for the keypad.

https://wiki.libsdl.org/SDL_Keycode

swoolcock:
@peppy were you going to manually publish this, and/or does there need to be a script to do it?

Once ppy.SDL2-CS is published I can go through and make sure this passes CI.

peppy:
Will just go with manual this time. We can automate if it becomes a regular thing.

peppy:
https://www.nuget.org/packages/ppy.SDL2-CS/

smoogipoo:
@peppy if you could test this on Windows that'd be great. I've tested on linux and macOS and it works fine.

peppy:
Can do. What's the scope of testing? Just list any and every issue I come across?

smoogipoo:
At this point I've only tested to the point it runs and I can use keyboard/mouse + special key combos (e.g. Ctrl+F).

peppy:
Okay, so just testing things roughly work, not testing for actually being a feasible replacement for opentk?

smoogipoo:
Yeah. There's fixes in here as mentioned in the OP but I haven't done a pass from the perspective of replacing osuTK yet.

bdach:
It only took me a month but I playtested this a bit on Windows. Aside from the issues previously known when the veldrid implementation came in looks good.

peppy:
Only real issue I can notice (which has always been the case, I think) is that input does not handle correctly when exiting the window. We want to still receive input so the cursor can move out of the screen in such cases

PR #3497, opened 2020-04-30
Rework dynamic compilation to use Roslyn-backed type resolution

smoogipoo:
This is a complete rework of the type performed during dynamic compilation. It follows the following process with some optimisations in-between, which I hope to have documented in a somewhat readable way in code:

1. Map all types to the types which they reference.
2. Build a directed graph from the changed file towards the test scene type by following the map generated in (1).
3. Find all files corresponding to the types by following the graph in (2) by-parent.

Performance wise this is slightly slower on first initialisation (building the map in step (1)), but further changes are about as quick as they were previously. Step (1) could could probably be done in the background rather than on change, but I haven't done that in this PR.

Scenarios I've tested:
1. Changing `AnimationClockComposite` while in `TestSceneAnimation`.
2. Changing `Video` while in `TestSceneAnimation`.
3. Changing `Video` while in `TestSceneVideo`.
4. Changing `Menu` while in `TestSceneNestedMenus`.
5. Changing `TabItem<T>` while in `TestSceneTabControl`.
6. osu!-side: Changing `LegacyDrumRoll` while in `TestSceneDrawableDrumRoll`.

All with `RequiredTypes` removed.

I haven't removed `RequiredTypes` in this PR since it touches quite a lot of files. That can be done in a separate PR.

### DIFF
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -25,6 +25,7 @@ using osu.Framework.Extensions.ExceptionExtensions;
 using osu.Framework.Graphics.Effects;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Layout;
+using osu.Framework.Testing;
 using osu.Framework.Utils;
 
 namespace osu.Framework.Graphics.Containers
@@ -36,6 +37,7 @@ namespace osu.Framework.Graphics.Containers
     /// Additionally, <see cref="CompositeDrawable"/>s support various effects, such as masking, edge effect,
     /// padding, and automatic sizing depending on their children.
     /// </summary>
+    [ExcludeFromDynamicCompile]
     public abstract partial class CompositeDrawable : Drawable
     {
         #region Contruction and disposal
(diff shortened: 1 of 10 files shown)


smoogipoo:
Regarding the newly-added `Todo:`, I've kept the original assembly resolution code, but plan to update that in the future too. Roslyn can be queried for the metadata references to so, but it's not causing _huge_ problems right now.

smoogipoo:
Just noticed a slight issue with this one, documenting for now.

Suppose the type hierarchy:
```
A
  -> B
    -> C
```

Recompiling C generates `A'B'C'`, but then recompiling B generates `A'B'C` (where `'` indicates types existing in the dynamic assembly).

Likely path of resolution is to store the files that were changed through an entire testscene's lifetime and always recompile those.

peppy:
Does `ExcludeFromDynamicCompile` break the inclusion chain?

smoogipoo:
From only that type, yes. So when attached to TestBrowser it’ll stop inclusion of TestBrowser and the toolbars it references, but if the toolbars are referenced by the test or another component not marked for exclusion they’ll still get be included.

PR #3506, opened 2020-05-01
Add texture cropping

Tom94:


peppy:
Would be nice to have a simple test scene for this!

peppy:
Seems like `sample-texture` is always null via nunit runs for some reason..

bdach:
The texture is `null` in nunit runs because when running through nunit, the DLL resource store from `osu.Framework.Tests` is not picked up (only the one from `osu.Framework` is). That in turn is caused because:

* `GridTestScene` does not inherit from `FrameworkTestScene` which loads the test project resource store, as they reside in separate assemblies (grid is in framework itself, FTS in test project),
* and even if it did, the setup code for the test is done in `loadComplete()` and not in a  `[Test]` method, which means the `FrameworkTestSceneTestRunner` wouldn't have the chance to initialise the resource store anyway as the call stack is:

```
[OneTimeSetUp] SetupGameHost()
    CreateRunner()
        new FrameworkTestSceneTestRunner()
```

and the ctor/`loadComplete()` run before/independent to that.

That said I don't think that's necessarily really important in this case as the test has no asserts anyway (and doesn't really have any hope of having any asserts besides "does not crash"), so I'd say we let this one slide.

peppy:
That's an unfortunate chain, but agree that it's fine as is for this case.

Issue #1291, opened 2017-12-26
Lowres texture have a dark border


jorolf:
This is especially visible when having many sprites next to each other.

Example: 
A GridContainer which contains Sprites.
![grafik](https://user-images.githubusercontent.com/12570757/34364331-08341e1a-ea85-11e7-8235-b5be9b387afb.png)

Original texture:
![grasstile](https://user-images.githubusercontent.com/12570757/34364353-6f81ddc8-ea85-11e7-96a9-1c4bce290c8a.png)

This is probably due to the interpolation algorithm as it tries to find the adjacent colors which don't exist.

Same scenario but instead of linear interpolation the nearest color is used:
![grafik](https://user-images.githubusercontent.com/12570757/34364390-d45d723e-ea85-11e7-97ec-60ce1254bfe9.png)

Tom94:
The texture class supports passing a filtering mode for this exact reason. If custom edge behavior is also desired, then we could additionally (trivially) add an edge behavior parameter.

jorolf:
But how would I use it together with a TextureStore?

Tom94:
Looks like `TextureStore` would need to be extended to pass these parameters through.

peppy:
the example provided should really be done using gl-level wrap/repeat mode rather than creating many sprites

jorolf:
Probably, though it should be able to display different tiles at any cell

PR #3511, opened 2020-05-03
Shader-level texture wrap modes in conjunction with mipmapping and texture atlasses

Tom94:
This PR addresses several things (from pair programming with @smoogipoo ):
- Implements shader-level texture wrap modes to allow texture wrapping in conjunction with our use of texture atlasses.
- Implements edge-padding of atlassed textures which fixes texture filtering (as well as mipmapping) near the edges of textures. Not only does this address a number of existing problems (e.g. darkening of mania and notes in osu!lazer), but it is also an enabler for the wrap modes.
- Implements an API in the `Sprite` class for specifying a _sub-rectangle_ of the sprite to which the sprite's texture should be stretched and around which the texture is wrapped according to its wrap modes in the S (horizontal) and T (vertical) directions. See the provided test case for an implementation example.

Here is the effect of different wrap modes on a texture. The red-bordered region indicates the sub-rectangle of the sprite around which the wrap modes are applied.
Screen Shot 2020-05-03 at 10 06 54


jorolf:
Does this fix https://github.com/ppy/osu-framework/issues/1291 ?

Tom94:
Good chance it does. Let me know if you try and you find that it does work. :)

Tom94:
Needed to make the PR quite a bit bigger just now to fix bleeding problems with fonts. Turns out, certain textures (such as characters) rely on being surrounded by _black_ padding as opposed to repeated padding, so now the way padding is generated depends on the wrap mode of the texture.

This change necessitated making wrap modes `readonly` per texture and adding their specification to `TextureStore.Get`.

bdach:
So I just ran this on my box and textures look kind of... crispy?

image

Note the fonts looking off, and also the weird moiré-like patterns on the sprite. Not quite sure what that's about, I'll have a read of the source now, but I figured I'd just raise that as a concern once I saw it.

osu.Framework/Graphics/Textures/TextureAtlas.cs
```diff
@@ -47,7 +47,7 @@ internal TextureWhitePixel WhitePixel
         private readonly All filteringMode;
         private readonly object textureRetrievalLock = new object();
 
-        public TextureAtlas(int width, int height, bool manualMipmaps = false, All filteringMode = All.Linear)
+        public TextureAtlas(int width, int height, bool manualMipmaps = true, All filteringMode = All.Linear)
```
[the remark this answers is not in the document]
Tom94: Thanks for the feedback!

Yeah, mipmaps were completely borked and I failed to notice; my bad. I overhauled some portions of the code and the crispiness as well as wrapping issues should be fixed now. :)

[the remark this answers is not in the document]
bdach: Yep, visually looks good here now! :)

image

bdach:
Not a 100% on some of the CI inspections that are reported here. [This patch](https://gist.github.com/bdach/1adc82ab71fe29cf09781dfa472d7aa7) would fix up all of them as far as I can see, but some are kind of weird (the worst one by far is the *Call to base member with implicit default parameters* inspection, since the cases where I put `null, null` either report the warning if the params are missing, or show a *The parameter ... has the same value* message, which is not a warning but still mildly annoying), so I refrained from pushing that in directly for now.

peppy:
i think previously the solution was using `default` instead

peppy:
@smoogipoo can you quickly check over my refactors (should be nothing too offensive)? good to merge otherwise

osu.Framework/Graphics/OpenGL/Textures/TextureGLSingle.cs
```diff
@@ -171,14 +175,27 @@ public override RectangleF GetTextureRect(RectangleF? textureRect)
...
             RectangleF texRect = GetTextureRect(textureRect);
             Vector2 inflationAmount = inflationPercentage.HasValue ? new Vector2(inflationPercentage.Value.X * texRect.Width, inflationPercentage.Value.Y * texRect.Height) : Vector2.Zero;
-            RectangleF inflatedTexRect = texRect.Inflate(inflationAmount);
+
+            // If clamp to edge is active, allow the texture coordinates to penetrate by half the repeated atlas margin width
+            if (GLWrapper.CurrentWrapModeS == WrapMode.ClampToEdge || GLWrapper.CurrentWrapModeT == WrapMode.ClampToEdge)
+            {
+                Vector2 inflationVector = Vector2.Zero;
+                if (GLWrapper.CurrentWrapModeS == WrapMode.ClampToEdge)
+                    inflationVector.X = TextureAtlas.PADDING / 4f / width;
```
smoogipoo, line 191: Shouldn't this be `(1 << MAX_MIPMAP_LEVELS) / 2 / width`, as per `DrawQuad`?

osu.Framework/Graphics/Sprites/Sprite.cs
```diff
@@ -20,14 +20,50 @@ public class Sprite : Drawable, ITexturedShaderDrawable
...
+
+                return result;
+            }
+        }
 
         /// <summary>
         /// Maximum value that can be set for <see cref="EdgeSmoothness"/> on either axis.
         /// </summary>
-        public const int MAX_EDGE_SMOOTHNESS = 2;
+        public const int MAX_EDGE_SMOOTHNESS = 3;
```
smoogipoo, line 66: Did you mean to increment this? Not sure 3 would ever be used, and it'd probably be best to keep atlas padding as low as possible otherwise.

osu.Framework/Graphics/OpenGL/Textures/TextureGLSingle.cs
Tom94: Nope, it's `/4` instead of `/2`, because filtering has a half-texel radius. In other words: to obtain 100% of the padded color, the texture coordinate must be right in the middle of the padded region, and the padded region is `(1 << MAX_MIPMAP_LEVELS) / 2` wide (the atlas' padding is shared half-half among neighboring textures).

osu.Framework/Graphics/Sprites/Sprite.cs
Tom94: I did mean to increment this, because I did want the atlas' padding to become bigger. A padding of x2 is unfortunately not quite enough to completely prevent bleeding. The truly 100% correct approach would actually be x4, but the difference to 3x is barely noticeable.

osu.Framework/Graphics/OpenGL/Textures/TextureGLSingle.cs
smoogipoo: So is the one in `DrawQuad` incorrect then? Or why is it `/2` there?

Tom94: They're actually both more-or-less correct. I made a small mistake above: the padded region is not `(1 << MAX_MIPMAP_LEVELS) / 2` but `(1 << MAX_MIPMAP_LEVELS) / 2 * MAX_EDGE_SMOOTHNESS`... so with a `MAX_EDGE_SMOOTHNESS` of 2, the two are equivalent.

But since `MAX_EDGE_SMOOTHNESS` is 3 now, I'd stay on the safe side and use the smaller value of `DrawQuad` also in `DrawTriangle`. Good catch!

osu.Framework/Graphics/Sprites/Sprite.cs
smoogipoo: Having a hard time understanding why this is 3 still, according to all my calculations it should work fine as 2 but that doesn't seem to be the case on visual inspection... Maybe we can voice chat to show me through getting to this value (or x4), but I'll go with it for now.

peppy: Probably deserves a well-written inline comment on resolution so we don't need to figure this out each time.

smoogipoo:
Gonna leave it there for today. The only remaining issue is that `None` should not be the default wrap mode, since there is very few cases where you want the edge-bleed by default especially when the bleed is the transparent black background.

osu.Framework/Graphics/Sprites/Sprite.cs
Tom94: I think the reason is that textures aren't positioned in the atlas at multiples of `1 << MAX_MIPMAP_LEVEL`. This adds an additional level of blur into the created mipmaps. It may well be that this additional blur is of size ~`(1 << MAX_MIPMAP_LEVEL) / 2`, which is why adding twice that to the padding (by increasing `MAX_EDGE_SMOOTHNESS` to 3) is sufficient to completely cancel the effect.

(I.e. my previous statement that 4x is correct may be incorrect and 3x may be truly fine. I am increasingly convinced of this.)

peppy:
@smoogipoo is this okay to get in as-is? would be nice to get it merged.

smoogipoo:
Yeah I think this is fine as is for now. We can consider changing the default after-the-fact.

smoogipoo:
Added one more obsoleted member that's currently used in a ruleset `Sprite.WrapTexture`.

Considered adding `TextureGL.WrapMode` back as well, but it's not used throughout either osu! or any rulesets, so I think it's pretty safe...

smoogipoo:
@peppy please add a comment on the max smoothness variable if you can understand the reasoning there.

peppy:
I've just added a link to the above discussion for now.

PR #3512, opened 2020-05-04
Fix build warnings

smoogipoo:
New inspections as a result of updates to roslyn analyzers (fxcop, bannedsymbols).

Also following https://youtrack.jetbrains.com/issue/RIDER-43956 re: inspections, so not applying fixes there just yet.


peppy:
We're probably going to need Input.cs in the future but seems fine for now.

PR #3516, opened 2020-05-04
Fix static log targets not being purged when a custom log storage location is specified

peppy:


osu.Framework/Logging/Logger.cs
```diff
@@ -60,7 +60,13 @@ public class Logger
         public static Storage Storage
         {
             private get => storage;
-            set => storage = value ?? throw new ArgumentNullException(nameof(value));
+            set
+            {
+                storage = value ?? throw new ArgumentNullException(nameof(value));
+
+                // clear static loggers so they are correctly purged at the new storage location.
+                static_loggers.Clear();
```
smoogipoo, line 68: What if someone keeps a reference to the logger from a previous call to `Logger.GetLogger()`?

This is done in `RoslynTypeReferenceBuilder`, `APIAccess`, and `SquirrelUpdateManager`, for quick examples.

peppy: The intention isn't to fix those cases. It's just to recycle the static references so they get deleted correctly on first write to a new location.

bdach:
From the source it looks like this'll do what it intends to but I couldn't get this tested with the game-side PR today. Maybe tomorrow if this is still here unmerged.

PR #3517, opened 2020-05-04
Propagate BasePath to subdirectory storages for consistency

peppy:


osu.Framework/Platform/Storage.cs
```diff
@@ -4,14 +4,32 @@
...
+
+                if (dependentStorages != null)
+                {
+                    foreach (var s in dependentStorages)
+                        s.BasePath = basePath;
+                }
+            }
+        }
+
+        private WeakList<Storage> dependentStorages;
```
iiSaLMaN, line 32: private fields should probably be at the top

peppy: such minor discrepancies in ordering are pretty pointless to point out at this point.

peppy:
Hold off merging this one as I reassess https://github.com/ppy/osu/pull/8935 please. Looking at whether we can potentially seal `Storage` or whether a completely different solution is required.

smoogipoo:
Closing as this was superseded in the referencing PR.

PR #3522, opened 2020-05-05
Refactor storage to allow more flexibility in root location

peppy:
Tentatively PRing this for comment. May need a few further changes.

Alternative to https://github.com/ppy/osu-framework/pull/3517.


osu.Framework/Platform/GameHost.cs
```diff
@@ -125,7 +125,14 @@ public abstract class GameHost : IIpcHost, IDisposable
 
         public virtual Clipboard GetClipboard() => null;
 
-        protected abstract Storage GetStorage(string baseName);
+        /// <summary>
+        /// Retrieve a storage for the specified location.
+        /// </summary>
+        /// <param name="path">The absolute path to be used as a root for the storage.</param>
+        /// <returns></returns>
```
smoogipoo, line 132:
```suggestion
```

osu.Framework/Platform/NativeStorage.cs
```diff
@@ -93,5 +91,16 @@ public override Stream GetStream(string path, FileAccess access = FileAccess.Rea
         public override string GetDatabaseConnectionString(string name) => string.Concat("Data Source=", GetFullPath($@"{name}.db", true));
 
         public override void DeleteDatabase(string name) => Delete($@"{name}.db");
+
+        public override Storage GetStorageForDirectory(string path)
```
smoogipoo, line 95: Can this be removed? It looks identical to the `Storage` implementation as far as I see...

peppy: the activator line is passing in the host (ctor differs for all `NativeStorage`s)

PR #3523, opened 2020-05-05
Add front-to-back support for textured sprites

### LINKED ISSUE #3129: Front-to-back should be enabled for Sprites
FtB currently isn't enabled for `Sprite`, since the drawn `Texture` is not guaranteed to be fully opaque.

`TextureGLSingle` should compute the opaqueness of its texture, initially just as "fully opaque" or "not fully opaque".
Then, `SpriteDrawNode` needs to be updated to draw the opaque interior (in this case, the full sprite). See `BoxDrawNode` as an implementation reference.

Tom94:
Collab w/ @smoogipoo 

- [x] Depends on #3511 

Closes #3129

Implements load-time checking of textures for opacity. Then, if a `Sprite` uses a fully opaque texture, performs a depth pre-pass on the opaque region of the `Sprite`. This opaque region depends on `Sprite.Texture.WrapMode`, the mip to be used, and `Sprite.TextureRectangle`. See `Sprite.cs` for details on this computation.

More implementation details:
- changes existing front-to-back rendering to become a depth pre-pass. All color is now rendered in the back-to-front pass. This facilitated easier integration with textured sprites.
- shader overhead is reduced by masking all color writes in the front-to-back pass. This avoids the need for a specialized low-overhead depth-only shader. (GPU drivers are smart enough to omit the shader computation when color values are masked, since this is a common use case for shadow maps.)


Tom94:
Note: this came from pair-programming with @smoogipoo :)

peppy:
Cool stuff! Any benchmarks on how this affects performance of existing tests?

Tom94:
The performance within VisualTests is pretty much unaffected. Hopefully the benefits'll show up in situations where large sprites are layered in front of each other, e.g. menus that are built from textures, storyboards, and similar.

... and hopefully the benefit will outweigh scanning every new texture for whether it's opaque and/or transparent. Although the overhead of scanning for transparency is likely worthwhile anyway, since, in the future, drawing of fully transparent sprites, such as in minimalistic skins imported from stable, can be optimized away.

iiSaLMaN:
Sounds like this nicely resolves https://github.com/ppy/osu-framework/issues/3129! perhaps you've missed to add a `Closes #3129` there :P

Tom94:
Good point!

osu.Framework/Graphics/OpenGL/Textures/TextureGLSingle.cs
```diff
@@ -345,9 +343,6 @@ public override bool Bind(TextureUnit unit = TextureUnit.Texture0, WrapMode? wra
             if (textureId <= 0)
                 return false;
 
-            if (IsTransparent)
-                return false;
```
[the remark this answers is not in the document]
Tom94: I think this optimization is better put into different places. Transparent textures shouldn't fail to bind. It should be perfectly legal to bind a transparent texture. The places that use the texture to draw should instead be responsible to optimize away the drawing if the texture is transparent.

osu.Framework/Graphics/OpenGL/Textures/TextureGLSub.cs
```diff
@@ -122,7 +120,8 @@ public override void SetData(ITextureUpload upload, WrapMode? wrapModeS = null,
                 upload.Bounds = adjustedBounds;
             }
 
-            parent?.SetData(upload, wrapModeS ?? this.wrapModeS, wrapModeT ?? this.wrapModeT);
+            opacity = UpdateOpacity(upload, opacity);
+            parent?.SetData(upload, wrapModeS ?? WrapModeS, wrapModeT ?? WrapModeT, opacity);
```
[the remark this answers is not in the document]
Tom94: There are a bunch of properties we'd like to have. Perhaps the code will be more understandable with them in mind:
1. we want to compute a `ITextureUpload`'s opacity only once. Hence the initial value of `null` and the property `opacity == Opacity` otherwise.
2. if the `ITextureUpload` covers a subset of the texture's bounds, then the texture's opacity status can only (i) stay the same, if it agrees with the `ITextureUpload`'s opacity status or (ii) become `Opacity.Mixed` if it doesn't agree.
3. parent textures have larger bounds than their subtextures, so `UpdateOpacity` has to be called for each of them.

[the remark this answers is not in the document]
Tom94: Admittedly, the locals could have better names. I'll rename `opacity` to `uploadOpacity`.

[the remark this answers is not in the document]
bdach: Okay, that does indeed help in understanding - especially the part noting that it's possible for textures to be uploaded in parts which necessitates the fallback to `Mixed` as the slowest non-optimised path if something doesn't agree. Thanks for clarifying! 🙂

peppy:
Requires some non-trivial conflict resolution.

smoogipoo:
PR updated, this seems fine to me. @peppy want your review though.

| before | after |
| --- | --- |
| Screenshot_2020-06-11_16-08-17 | Screenshot_2020-06-11_16-08-27 |

You can notice it in the avatars on the left. Other cases (background, wedge, beatmap panels) don't utilise this because they're inside buffered containers.

Tom94:
This makes me wonder... can we make a class `OpaqueBufferedContainer` that always produces opaque textures by construction?

Tom94:
Or probably easier: a simple member variable of `BufferedContainer`

smoogipoo:
Yeah certainly, but unfortunately all that logic is inside `SpriteDrawNode`, which `BufferedDrawNode` and by extension `BufferedContainerDrawNode` isn't.

Maybe it'd be good to make a static helper for some of this stuff - it's also not just the DrawNode but the conservative quad too.

Tom94:
Absolutely!

Anothing thing I've been wondering about: does rendering to the framebuffer within a buffered container already make use of the depth prepass? This could also be a (potentially huge) optimization opportunity, given that we have quite a lot of rendering within buffered containers that are occluded by something in the foreground.

smoogipoo:
1. BufferedContainer itself doesn't do any internal FTB (i.e. within its own contents). It could but that needs extra considerations because BCs are basically the way to escape FTB in the first place, so for example if the BC has no user-provided depth buffer than it could safely create an implicit one that enables FTB. Would also need support at `GLWrapper` level since FTB is hardcoded to only apply when rendering to the backbuffer: https://github.com/ppy/osu-framework/blob/8f88198f2f5e2c9693ac2967e7dd607524405ee3/osu.Framework/Resources/Shaders/sh_Backbuffer_Internal.h#L3-L6
2. Or are you saying the BC should use the backbuffer's depth buffer for its own contents? I don't think that's possible without blitting from the backbuffer's depth to the FBO, which may flush the pipeline? Maybe if we were to render the entire game to an FBO (and blit to the backbuffer for swap), then we could keep a "backbuffer depth" attachment and apply that to all BCs (and would also satisfy (1)).

Tom94:
I was thinking about both 1 and 2. But primarily 2. I think a flush isn't all that bad, because switching to drawing into a buffered container requires a flush anyway, doesn't it?

smoogipoo:
I meant a GPU pipeline flush not a flush of our batches (which yes, does happen), because we'd require the rendering to complete. I could be wrong though - maybe since it'd be blitting between buffers on the GPU, OpenGL wouldn't require a pipeline flush for that case.

peppy:
before I start reviewing this, can you confirm it is ready for review (regardless of above discussion)?

Tom94:
Not sure if smoogi has anything to add, but nothing from my side.

smoogipoo:
Same here, it’s good to go.

osu.Framework/Graphics/Video/VideoTexture.cs
```diff
@@ -27,11 +27,14 @@ public VideoTexture(int width, int height, WrapMode wrapModeS = WrapMode.None, W
...
-        internal override void SetData(ITextureUpload upload, WrapMode wrapModeS, WrapMode wrapModeT)
+        internal override void SetData(ITextureUpload upload, WrapMode wrapModeS, WrapMode wrapModeT, Opacity? uploadOpacity)
         {
             UploadComplete = false;
 
-            base.SetData(upload, wrapModeS, wrapModeT);
+            // We do not support videos with transparency at this point,
+            // so the upload's opacity as well as the texture's opacity
+            // is always opaque.
+            base.SetData(upload, wrapModeS, wrapModeT, Opacity = Opacity.Opaque);
```
peppy, line 37: should we assert that the incoming opacity is null or opaque, just for safety?

osu.Framework/Graphics/OpenGL/Textures/TextureGL.cs
```diff
@@ -140,15 +143,58 @@ public void Dispose()
...
         /// <param name="upload">The <see cref="ITextureUpload"/> containing the data.</param>
         /// <param name="wrapModeS">The texture wrap mode in horizontal direction.</param>
         /// <param name="wrapModeT">The texture wrap mode in vertical direction.</param>
-        internal abstract void SetData(ITextureUpload upload, WrapMode wrapModeS, WrapMode wrapModeT);
+        /// <param name="uploadOpacity">Whether the upload is opaque, transparent, or a mix of both..</param>
+        internal abstract void SetData(ITextureUpload upload, WrapMode wrapModeS, WrapMode wrapModeT, Opacity? uploadOpacity);
+
+        protected static Opacity ComputeOpacity(ITextureUpload upload)
+        {
+            if (upload is VideoTextureUpload)
```
peppy, line 159: Curious if this is required, considering it is being forced in the `VideoTexture` class itself. maybe we should remove that local one (and make `VideoTextureUpload` internal if we do)?

osu.Framework/Graphics/Video/VideoTexture.cs
Tom94: That's gonna be very expensive to compute, so I don't think it's a good idea.

peppy: i just mean
```
if (uploadOpacity != null && uploadOpacity != Opacity.Opaque)
    throw new InvalidOperationException("Video texture uploads must always be opaque");
```

to avoid a user attempting to use this and provide the argument which is being silently ignored.

Tom94: That's fair.

osu.Framework/Graphics/OpenGL/Textures/TextureGL.cs
peppy: @smoogipoo can you comment on this one? it seems unnecessary (and because of that, a bit confusing when you come across it).

smoogipoo: Yeah it does look to be redundant for that reason. Would be fine with removing with just some quick manual testing to make sure that is the case.

peppy: Looks sane on a manual check.

PR #3525, opened 2020-05-07
Prefer memory overloads for Stream.ReadAsync/WriteAsync

turbedi:
Benefits from the returning type being `ValueTask<int>`.


bdach:
Have you benchmarked this? What are the measurable performance gains from these changes?

turbedi:
I haven't done any benchmarks yet, I should be able to do some this weekend.
[
As far as I understood, it should be a free optimization in cases where where `Stream.Read/WriteAsync  `methods are used.](https://github.com/dotnet/runtime/issues/33790)

peppy:
These are seldom used paths and the operation performed inside the task is magnitudes more intensive than creating a `Task` instance. I'd be wary of making this change just because-we-can in one location – I may think differently if this was applied along with a performance rule on inspectcode or roslyn to ensure these changes are made anywhere/everywhere.

PR #3531, opened 2020-05-08
Allow ExcludeFromDynamicCompile on enums

smoogipoo:


peppy:
Where's the required usage of this?

smoogipoo:
I want to add `OsuSetting` to the exclusion.

peppy:
Do you foresee this to continue being something we need to add in many places? Just wondering if we can heuristically decide excludes somehow..

smoogipoo:
I didn't intend for this to change whether this is used going forward, nor that I'd submit an osu!-side PR for this. I just had a very specific purpose for which I didn't see a reason not to have a target for.

For my specific purpose I believe moving `OsuSettings` out of the same file as `OsuConfigManager` would achieve the same.

I've been thinking about ways in which the hierarchy could be trimmed and have a few ideas, but need to think about the implementation and it's not super easy. Also a task parallel to this PR.

PR #3537, opened 2020-05-11
Introduce Colour4 type

swoolcock:
Intended as a replacement for osuTK `Color4` as part of the decoupling.  Prerequisite for future PRs doing the actual replacement.

It contains the bare minimum `Color4` functionality we require, as well as adding a few of our existing `Color4Extensions` extension functions.  More functionality can be added as required.

Note that it is readonly, unlike `Color4`.  While this is up for discussion, I think we should treat colours as immutable data types.  It may even be worth making `SRGBColour` and `ColourInfo` readonly in future.

Related to #2448


osu.Framework/Graphics/Colour/Colour4.cs
```diff
@@ -0,0 +1,909 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+
+namespace osu.Framework.Graphics.Colour
```
smoogipoo, line 6: `osu.Framework.Graphics` is a better location (too namespace-y).

osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
```diff
@@ -259,5 +260,19 @@ public static (float h, float s, float v) ToHSV(this Color4 colour)
 
             return (h, s, v);
         }
+
+        /// <summary>
+        /// Converts an osuTK <see cref="Color4"/> to an osu!framework <see cref="Colour4"/>.
+        /// </summary>
+        /// <param name="colour">The osuTK <see cref="Color4"/> to convert.</param>
+        public static Colour4 ToColour4(this Color4 colour) =>
```
smoogipoo, line 268: Ehhh.. I would rather that this was changed already and `[Obsolete]`d implicit operators were placed on `Colour4`.

swoolcock: I was hoping to not have any osuTK references in the `Colour4` type itself, but we can add and `[Obsolete]` them for sure.

smoogipoo: We may be able to fast-forward the obsoletion process if anything. This is just to leave the potential for maybe keeping osuTK around for normal 6-month obsoletion timeframe.

KoziLord:
I'd consider implementing `Colour4` as a wrapper for `Vector4` which uses vector instructions to speed up its math.
I wrote a small benchmark comparing the two approaches and these are the results I got, `Clr` being the wrapper version:
```
|         Method |      Mean |     Error |    StdDev |
|--------------- |----------:|----------:|----------:|
|      AddColour | 784.45 μs | 14.311 μs | 13.387 μs |
| MultiplyColour | 372.59 μs |  6.683 μs |  5.924 μs |
|    ClampColour | 504.91 μs |  7.860 μs |  7.352 μs |
|         AddClr |  67.38 μs |  1.339 μs |  2.380 μs |
|    MultiplyClr |  66.91 μs |  1.333 μs |  1.686 μs |
|       ClampClr | 373.06 μs |  7.445 μs | 14.164 μs |
```
([Benchmark code](https://gist.github.com/KoziLord/d4443eab5188c4344815131461c27923))

swoolcock:
@KoziLord yeah might be worth a look.  @smoogipoo any opinion on that?

smoogipoo:
I don't disagree.

osu.Framework/Graphics/Colour4.cs
```diff
@@ -0,0 +1,954 @@
...
+        #endregion
+
+        #region Chaining Functions
+
+        /// <summary>
+        /// Returns a new <see cref="Colour4"/> with the same RGB components, but multiplying the current alpha component by a scalar value.
+        /// The final alpha is clamped to the 0-1 range.
+        /// </summary>
+        /// <param name="scalar">The value that the existing alpha will be multiplied by.</param>
+        /// <returns></returns>
```
turbedi, line 91:
```suggestion
        /// <param name="scalar">The value that the existing alpha will be multiplied by.</param>
        /// <exception cref="ArgumentOutOfRangeException">Thrown if <see cref="scalar"/> is negative.</exception>
        /// <returns></returns>
```

osu.Framework/Graphics/Colour4.cs
```diff
@@ -57,10 +62,21 @@ public Colour4(float r, float g, float b, float a)
         /// <param name="a">The alpha component, in the 0-255 range.</param>
         public Colour4(byte r, byte g, byte b, byte a)
         {
-            R = r / (float)byte.MaxValue;
-            G = g / (float)byte.MaxValue;
-            B = b / (float)byte.MaxValue;
-            A = a / (float)byte.MaxValue;
+            Vector = new Vector4(
```
KoziLord, line 65: Change to `Vector = (new Vector4(r, g, b, a) / new Vector4(byte.MaxValue))`, this will perform one vector division instead of four scalar divisons

swoolcock:
To be fair, I don’t think colours are something we need to preoptimise right now.  This is a drop in replacement for the osuTK `Color4` so we have one less dependency on it.

KoziLord:
That's fair, just throwing out some ideas that could improve performance if need be.

osu.Framework/Graphics/Colour4.cs
```diff
@@ -0,0 +1,954 @@
...
+        /// <summary>
+        /// Linearly divides a colour by a scalar value. The final value is clamped to the 0-1 range.
+        /// </summary>
+        /// <param name="colour">The original colour.</param>
+        /// <param name="scalar">The scalar value to divide by. Must be positive.</param>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown if <see cref="scalar"/> is zero or negative.</exception>
+        public static Colour4 operator /(Colour4 colour, float scalar)
+        {
+            if (scalar <= 0)
+                throw new ArgumentOutOfRangeException(nameof(scalar), scalar, "Cannot divide colours by non-positive values.");
```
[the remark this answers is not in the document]
peppy: is that only in the case of actually throwing?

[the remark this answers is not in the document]
KoziLord: No, the benchmark never throws in this example.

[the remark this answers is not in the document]
bdach: Off of a quick look at [sharplab](https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcDMACR2DC2A3utmdgA4BOAlgG5jACm2ASkxBQPYB2AJkyrYqHbv0HYAvNh5MA7m1G8BVABQBKANylyOsllwAWAr2BgasqgHUawABYARJgBsmAc0ZM+AFTtUucgD6DlwcAHJcwL7+chrEeuRkIpzKggB0+Kbmljb2Tq4ezD5+AVLYAKzl2miJAL4JCQZwxpk8ZhaCuY4u7p7FMYHRARBxJDWJ5MliKhlZHda23QV9QwrSSAAMSNV1DeP6OM0mbdmdiw40IgDGwMGhEBFRJbHq8fsTU6lUsyfzXRfXJ4xMqVHbkervRqHFpzHLnS5MG6DZ4jV5jCaTJTib6tdpwvIIm6rMqbbYJCGJBLUeieXAIRQpbEJdEYyi0BjMbAWYDYK6wwTVBKJAzc454s55HqFLyrIXkFms8huJg8yQAPl5/KoYMVZAgKrlGIVusSNAAZthVAxnABXFgAHmwG2wMBg2GtduwGs2G3UhpN5AEyyKqw0gveAc1v0sZQ9TB1uoprIh/pFbTFpwWBMBsoj8v9iWVqo1fOjAoL5H1wArZGNkfNlrj2Edztd7rAtpY3o2vpruvswNkCgAkjxrTQ+AB5CiCRg0XgAUQAHlcmBRgPOeGH0H3EqXxUJpHGE4qkxiU3myNSOSw4EgAGxGbBB3oh55xdXYAelIfYUfjqcZyoOdFxXNcN14bd3ghWogA=) the answer seems to be no, this improves performance in both cases due to enabling inlining.

Unfortunately the example is not too clear to read as it's hard to tell what registers hold what and there's a lack of method offsets so it's hard to tell what's getting called, but there's clearly something going on in the setter calls of `ContainerWithDelegatedThrow` (when it shouldn't throw, it seems to do a direct `mov` and `ret`, when it should throw, it does an `int3` and doesn't even emit the constant value). Calls to `ContainerWithDirectThrow` however emit the same asm and just differ in param value.

PR #3541, opened 2020-05-13
Fix hard crash on MIDI initialisation failure

peppy:
- Closes https://github.com/ppy/osu/issues/9008

### DIFF
--- a/osu.Framework/Input/Handlers/Midi/MidiInputHandler.cs
+++ b/osu.Framework/Input/Handlers/Midi/MidiInputHandler.cs
@@ -32,23 +32,11 @@ namespace osu.Framework.Input.Handlers.Midi
 
         public override bool Initialize(GameHost host)
         {
-            try
-            {
-                refreshDevices();
-            }
-            catch (Exception e)
-            {
-                Logger.Error(e, RuntimeInfo.OS == RuntimeInfo.Platform.Linux
-                    ? "Couldn't list input devices. Is libasound2-dev installed?"
-                    : "Couldn't list input devices. There may be another application already using MIDI.");
-                return false;
-            }
-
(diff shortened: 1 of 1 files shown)


iiSaLMaN:
While this fixes the linked issue specifically, it'll still potentially fail if a MIDI device got plugged in post-initialization while another program looking for MIDIs "opens" the device before the framework. I guess the only solution here is to move the try-catch to inside `refreshDevices`.

peppy:
Have you reproduced this scenario? Do you have a specific application which has this kind of "looking for MIDIs" behaviour?

iiSaLMaN:
Not really, just envisioned as a potential scenario. Looks good now.

Stedoss:
Putting the whole method in a try-catch means that MIDI devices that are lower in the input list will get ignored I think; something like this might be more appropriate?
```csharp
foreach (IMidiPortDetails input in inputs)
{
    if (openedDevices.All(x => x.Key != input.Id))
    {
        IMidiInput newInput;
        try
        {
            newInput = MidiAccessManager.Default.OpenInputAsync(input.Id).Result; //This is the line that is throwing the exception.
        }
        catch
        {
            Logger.Log($"Unable to open MIDI input stream on device {input.Name} (is another application using this MIDI device?)");
            continue;
        }
        newInput.MessageReceived += onMidiMessageReceived;
        openedDevices[input.Id] = newInput;

        Logger.Log($"Connected MIDI device: {newInput.Details.Name}");
    }
}
```

This also allows for the MIDI device to be detected once the application that is using it stops (from my testing).

peppy:
That just gets complicated and is out of scope in my opinion. You would be getting that log output *every* 500ms. Restarting osu! after closing other apps is probably what we want for now.

PR #3545, opened 2020-05-13
Add joystick axis value change processing to InputManager

### LINKED ISSUE #3544: Joystick axis values are not visible from the InputManager's CurrentState
In the Joystick test scene, this piece of code retrieves the current value for each axis and displays it in its corresponding square: 

https://github.com/ppy/osu-framework/blob/755247a749c5414d034be8e07be71091557dedd7/osu.Framework.Tests/Visual/Input/TestSceneJoystick.cs#L238-L244

The problem is, no axis values are present in the joystick state, causing the values to stay as 0.00 as visible on the following GIF: 

Axis values stay zero, while off-center movement is properly detected by the square color

Analog input is properly seen and processed as the associated button events fired by the framework are registered.

I used a DualShock 4 for this GIF, but other joysticks behave the same.

Jeremiidesu:
Trying to fix #3544.

Currently using a Dualshock 4.

Before the implementation of the fix: 

Axis buttons get pressed, axis values stay 0

After implementation:

Axis buttons get pressed, axis values vary

Main issues I can see for now is that Rider reports a lot of memory allocations around my `JoystickAxisEvent` class, not too sure how to go about that. These memory allocations seem to be much less of a problem when running in Release mode, but I can't be sure as I don't know if Rider collects memory allocation information in Release mode.


osu.Framework/Input/StateChanges/Events/JoystickAxisChangeEvent.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace osu.Framework.Input.StateChanges.Events
+{
+    public class JoystickAxisChangeEvent : InputStateChangeEvent
+    {
+        public readonly int Axis;
+
+        public readonly float Value;
+
+        public JoystickAxisChangeEvent(InputState state, IInput input, int axis, float value)
```
iiSaLMaN, line 14: Always use `JoystickAxis` to refer to (`int axis, float value`)
```suggestion
        public JoystickAxisChangeEvent(InputState state, IInput input, JoystickAxis axis)
```

osu.Framework/Input/InputManager.cs
```diff
@@ -529,12 +529,28 @@ public virtual void HandleInputStateChange(InputStateChangeEvent inputStateChang
...
+        protected virtual void HandleJoystickAxisChange(JoystickAxisChangeEvent e)
+        {
+            var currentAxes = CurrentState.Joystick.Axes;
+
+            if (currentAxes.Any(a => a.Axis == e.Axis))
+            {
+                currentAxes.Remove(currentAxes.First(a => a.Axis == e.Axis));
+            }
+
+            currentAxes.Add(new JoystickAxis(e.Axis, e.Value));
```
iiSaLMaN, line 551: should be inside `JoystickEventManager` in an own handle method `HandleAxisChange()`

osu.Framework/Input/InputManager.cs
```diff
@@ -529,12 +529,28 @@ public virtual void HandleInputStateChange(InputStateChangeEvent inputStateChang
...
             }
         }
 
+        protected virtual void HandleJoystickAxisChange(JoystickAxisChangeEvent e)
+        {
+            var currentAxes = CurrentState.Joystick.Axes;
+
+            if (currentAxes.Any(a => a.Axis == e.Axis))
+            {
+                currentAxes.Remove(currentAxes.First(a => a.Axis == e.Axis));
```
iiSaLMaN, line 548: I think you should change the type of `JoystickState.Axes` into:
```csharp
public readonly float[] Axis = new float[MAX_AXES];
```
to simplify this into a one-liner `currentAxes[axis] = value`, with the constant declared as `public const int MAX_AXES = 64;` as `64` seems safe for maximum axes.

osu.Framework/Input/StateChanges/JoystickAxisInput.cs
```diff
@@ -0,0 +1,52 @@
...
+                {
+                    if (Precision.AlmostEquals(axes[i].Value, previous.First(a => a.Axis == axes[i].Axis).Value))
+                        continue;
+                }
+
+                Axes.Add(axes[i]);
+            }
+        }
+
+        protected List<JoystickAxis> GetJoystickAxes(InputState state) => state.Joystick?.Axes;
```
iiSaLMaN, line 39: Currently unused methods with no purpose aren't really necessary to be added.

osu.Framework/Input/StateChanges/JoystickAxisInput.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        protected JoystickAxisChangeEvent CreateEvent(InputState state, int axis, float value) => new JoystickAxisChangeEvent(state, this, axis, value);
```
iiSaLMaN, line 41: Can be in-lined instead.

osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
```diff
@@ -120,6 +121,11 @@ public OsuTKJoystickState(JoystickDevice device)
                         else if (value < deadzone * -1)
                             Buttons.SetPressed(JoystickButton.FirstAxisNegative + i, true);
                     }
+                    else
+                    {
+                        // Recenter the axis value if centered
+                        Axes.Add(new JoystickAxis(i, 0));
```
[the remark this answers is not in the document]
bdach: Sort of disagree on this one. This isn't one of those cases that I would say benefits from an early-return guard, as there's code immediately preceding it - so the `continue` would be obscuring the branching.

osu.Framework/Input/StateChanges/JoystickAxisInput.cs
```diff
@@ -0,0 +1,52 @@
...
+using osu.Framework.Input.States;
+using osu.Framework.Utils;
+
+namespace osu.Framework.Input.StateChanges
+{
+    public class JoystickAxisInput : IInput
+    {
+        private List<JoystickAxis> Axes;
+
+        public JoystickAxisInput(List<JoystickAxis> axes, List<JoystickAxis> previous)
```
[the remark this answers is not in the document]
bdach: Disagree with making a base class just for constructors. Use whatever's convenient wherever.

osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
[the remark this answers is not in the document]
iiSaLMaN: Right, I think rounding the axis value directly to zero is more better:
```csharp
for (int i = 0; i < JoystickDevice.MAX_AXES; i++)
{
    var value = device.RawState.GetAxis(i);
    
    // do not allow a deadzone below float_epsilon
    var deadzone = MathF.Max(device.DefaultDeadzones?[i] ?? 0, Precision.FLOAT_EPSILON);
    
    if (Precision.AlmostEquals(value, 0, deadzone))
        // Round values in the deadzone to zero.
        value = 0;

    Axes.Add(new JoystickAxis(i, value));

    if (value > deadzone)
         Buttons.SetPressed(JoystickButton.FirstAxisPositive + i, true);
    else if (value < -deadzone)
         Buttons.SetPressed(JoystickButton.FirstAxisNegative + i, true);
}

File: osu.Framework/Input/StateChanges/JoystickAxisInput.cs (hunk 7 above)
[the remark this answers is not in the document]
Comment (Jeremiidesu) [reply osu.Framework/Input/StateChanges/JoystickAxisInput.cs:16]: Still not sure sure how to go regarding List vs IEnumerable considering the last constructor's current code, but looks good to me for the set of three. Will see what works the best as I implement the other suggested changes.

File: osu.Framework.Tests/Visual/Input/TestSceneJoystick.cs [hunk 8]
```diff
@@ -239,8 +239,8 @@ protected override void Update()
             {
                 base.Update();
 
-                var joy = GetContainingInputManager().CurrentState.Joystick;
-                rawValue.Text = joy.Axes.Find(a => a.Axis == axisIndex).Value.ToString("0.00");
+                // var joy = GetContainingInputManager().CurrentState.Joystick;
+                // rawValue.Text = joy.Axes[axisIndex].ToString("0.00");
```
frenzibyte, line 243: Stale code doesn't need to be commented (unless has purpose), can be removed altogether.

Jeremiidesu: Wasn't too sure on what to do about that part as it's still worthy of a test IMO, but I'll do that for now.

osu.Framework.Tests/Visual/Input/TestSceneJoystick.cs
```diff
@@ -259,7 +259,8 @@ protected override bool OnJoystickAxisMove(JoystickAxisMoveEvent e)
             {
                 if (e.Axis.Axis == axisIndex)
                     rawValue.Text = e.Axis.Value.ToString("0.00");
-                return true;
+
+                return false;
```
Jeremiidesu, line 263: To be honest, I'm not too sure whether OnJoystickAxisMove should return a boolean considering returning `true` here prevented the values to display in the test scene. Chances are my implementation may be flawed too ; would love some input on this.

bdach: `OnJoystickAxisMove` returning a boolean is correct and consistent with all other input methods in the framework. Returning false here is also correct - the return value of input handlers indicates whether or not an input event has been handled.

If an input event has been handled (`true` is returned), it does not propagate further to other drawables - in this case likely the wrong `JoystickAxisButtonHandler` ended up marking another axis's event as handled.

If you *really* wanted to, you could rewrite this as

```csharp
protected override bool OnJoystickAxisMove(JoystickAxisMoveEvent e)
{
    if (e.Axis.Axis != axisIndex)
        return false;

    rawValue.Text = e.Axis.Value.ToString("0.00");
    return true;
}
```

but it doesn't really matter and the current version is fine.

osu.Framework/Input/JoystickAxisEventManager.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// Handles axis value changing.
+        /// </summary>
+        /// <param name="state">The current <see cref="InputState"/>.</param>
+        /// <param name="newValue">The new value for this axis.</param>
+        public void HandleAxisChange(InputState state, float newValue)
+        {
+            // Not enough movement, drop event.
+            if (Precision.AlmostEquals(value, newValue)) return;
+
+            value = newValue;
```
frenzibyte, line 48: This should be moved to the `IInput` instead (with the old value being retrieved from `state.Joystick.Axes[source]`).

osu.Framework.Tests/Visual/Input/TestSceneJoystick.cs
Jeremiidesu: I'll keep it as it is then.

frenzibyte: Going to say it here instead of in that review since you mentioned it, what @bdach said is correct, but I was thinking about approaching this slightly different:

You would have an `JoystickAxisStartEvent` event that would be called when an axis has been moved from the center position, handling it (returning true) would let the event manager start propagating movement events to the handling drawable, and an `JoystickAxisEndEvent` event which is called when an axis has been moved back to the center position. **(exactly like how dragging events are propagated)**

That would let you listen for specific axes by a simple:
```csharp
protected override bool OnJoystickAxisStart(JoystickAxisStartEvent e) => e.Axis.Axis == 2; // handle third axis.
```
while also being able to ready up some objects before working with them on new axis values.

---

Would like to receive input on whether this is worth adding.

bdach: I don't see how gamepad axes are analogous to drag events. Keep in mind that xinput doesn't always map controller sticks to axes, things like trigger buttons with analog output are also mapped to axes.

osu.Framework/Input/Events/JoystickAxisMoveEvent.cs
```diff
@@ -0,0 +1,25 @@
...
+        public readonly JoystickAxis Axis;
+
+        public JoystickAxisMoveEvent([NotNull] InputState state, JoystickAxis axis)
+            : base(state)
+        {
+            Axis = axis;
+        }
+
+        public override string ToString() => $"{GetType().ReadableName()}({Axis})";
+    }
```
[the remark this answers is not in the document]
Jeremiidesu: Would something like that be what you have in mind for the `JoystickEvent` base class?

```cs
public abstract class JoystickEvent : UIEvent
{
    protected JoystickEvent([NotNull] InputState state)
        : base(state)
    {
    }

    /// <summary>
    /// List of currently pressed joystick buttons.
    /// </summary>
    public IEnumerable<JoystickButton> PressedButtons => CurrentState.Joystick.Buttons;

    /// <summary>
    /// List of joystick axes. Axes which have zero value may be omitted.
    /// </summary>
    public IReadOnlyList<float> Axes => CurrentState.Joystick.Axes;
}
```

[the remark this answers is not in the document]
bdach: Disagree with the inheritance. What would `PressedButtons` mean on an *axis move* event? If anything axes should be deprecated/removed on `JoystickButtonEvent` to entirely separate button and axis inputs IMO.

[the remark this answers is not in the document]
frenzibyte: @bdach We're exposing relevant information useful on this event here (joystick-related info), just like `MouseEvent` exposing `HasAnyButtonsPressed`, or how `UIEvent` exposes some keyboard key presses or current mouse position.

Say:
```csharp
protected override bool OnJoystickAxisMove(JoystickAxisMoveEvent e)
{
    if (e.PressedButtons.First(b => b == JoystickButton.Button2))
    {
        // do some logic on current axis value while Button2 is pressed. 
        return true;
    }

    return false;
}
```

osu.Framework.Tests/Visual/Input/TestSceneJoystick.cs
frenzibyte: The only relation I'm specifiying here between gamepad axes and drag events are just propagation, I'm saying that you can filter what axes you would like to listen to their movements with `OnJoystickAxisStart(... e) => e.Axis.Axis == xyz;`

osu.Framework/Input/Events/JoystickAxisMoveEvent.cs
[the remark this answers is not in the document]
bdach: If someone really needs to correlate joystick movements with button presses or get the full raw joystick state they can handle `UIEvent` instead or access it through `CurrentState` anyway. I find exposing things unrelated to moving an axis in an axis move event very weird. (And for the record I find `HasAnyButtonsPressed` a little weird too but I'm guessing it was useful enough times to warrant inclusion.)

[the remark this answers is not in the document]
frenzibyte: Yeah the second suggestion only saves a couple of *accesses*, but **also** makes things consistent with how the other events are written and the inheritance in them (actually `MouseEvent` exposes `PressedButtons` and `IsPressed(MouseButton)` as well), that's why I suggested it.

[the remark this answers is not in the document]
Jeremiidesu: Considering the points made, it would make sense for consistency to go with a `JoystickEvent` class, unless it's planned to eventually (and consistently) phase out common event classes like `MouseEvent`. For an initial implementation without definite plants for the rest, I'd rather go for consistency than what makes the most sense.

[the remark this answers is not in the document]
frenzibyte: *sorry missed your question*.

> Would something like that be what you have in mind for the `JoystickEvent` base class?
> 
> ```cs
> public abstract class JoystickEvent : UIEvent
> {
>     protected JoystickEvent([NotNull] InputState state)
>         : base(state)
>     {
>     }
> 
>     /// <summary>
>     /// List of currently pressed joystick buttons.
>     /// </summary>
>     public IEnumerable<JoystickButton> PressedButtons => CurrentState.Joystick.Buttons;
> 
>     /// <summary>
>     /// List of joystick axes. Axes which have zero value may be omitted.
>     /// </summary>
>     public IReadOnlyList<float> Axes => CurrentState.Joystick.Axes;
> }
> ```
> 
> For the additional event info, I'm not sure whether I should use two `JoystickAxis` or two floats to avoid redundant information (the axis is going to be the same)

Yeah pretty much, as for the additional event info, I was thinking of something like (as we're in a `JoystickAxisMoveEvent`):
```csharp
/// <summary>
/// The current joystick axis data.
/// </summary>
public readonly JoystickAxis Axis;

/// <summary>
/// The last value of this joystick axis.
/// </summary>
public readonly float LastValue;

/// <summary>
/// The difference of axis value from last value to current value.
/// </summary>
public float Delta => Axis.Value - LastValue;
```

osu.Framework/Input/StateChanges/JoystickAxisInput.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using osu.Framework.Extensions.IEnumerableExtensions;
+using osu.Framework.Input.StateChanges.Events;
+using osu.Framework.Input.States;
+
+namespace osu.Framework.Input.StateChanges
+{
+    public class JoystickAxisInput : IInput
+    {
+        private readonly float[] axes = new float[JoystickState.MAX_AXES];
```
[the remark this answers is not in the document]
Jeremiidesu: > add a guard against axes collections that violate the maximum count
Not too sure what you mean by this, do you have an example?

osu.Framework/Input/Events/JoystickButtonEvent.cs
```diff
@@ -28,7 +28,7 @@ protected JoystickButtonEvent(InputState state, JoystickButton button)
         /// <summary>
         /// List of joystick axes. Axes which have zero value may be omitted.
         /// </summary>
-        public IEnumerable<JoystickAxis> Axes => CurrentState.Joystick.Axes;
+        public float[] Axes => CurrentState.Joystick.Axes;
```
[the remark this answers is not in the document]
Jeremiidesu: From what you're suggesting, that looks like to be blinding converting a `IEnumerable<float>` into a `IEnumerable<JoystickAxis>` without any regards for their value. Is this what you have in mind for "active axes" or should additional processing be done regarding what axes are considered active?

Jeremiidesu:
> `enum`s are better-suited here for identifying the source of the event (axis ID)

Regarding this, I added an `InputAxis` enum considering JoystickAxis is already the (axis, value) pair, and found myself replacing a lot of the `int`s referring to axes by this new `InputAxis` enum. Not too sure if that was the best way to proceed, but it makes sense to me.

osu.Framework/Input/Events/JoystickButtonEvent.cs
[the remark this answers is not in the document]
frenzibyte: Oh, I meant to only provide the non-zero-value ones but I forgot a `.Where` in there, my bad:
```csharp
public IEnumerable<JoystickAxis> GetActiveAxes() => Axes.Where(v => v != 0).Select((v, i) => new JoystickAxes(i, v);
```

osu.Framework.Tests/Visual/Input/TestSceneJoystick.cs
frenzibyte: My proposal can be dropped for now, seems like it'll add a lot without much benefits.

osu.Framework/Input/InputManager.cs
```diff
@@ -120,7 +120,8 @@ public Drawable DraggedDrawable
         private readonly Dictionary<JoystickButton, JoystickButtonEventManager> joystickButtonEventManagers = new Dictionary<JoystickButton, JoystickButtonEventManager>();
         private readonly Dictionary<MidiKey, MidiKeyEventManager> midiKeyEventManagers = new Dictionary<MidiKey, MidiKeyEventManager>();
 
-        private readonly JoystickAxisEventManager[] joystickAxisEventManagers = new JoystickAxisEventManager[JoystickState.MAX_AXES];
+        // private readonly JoystickAxisEventManager[] joystickAxisEventManagers = new JoystickAxisEventManager[JoystickState.MAX_AXES];
```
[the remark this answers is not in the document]
Jeremiidesu: Yup, that one's been removed in the last commit (unless I'm very stupid and missed it again)

osu.Framework/Input/JoystickAxisEventManager.cs
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        /// <summary>
+        /// Handles axis value changing.
+        /// </summary>
+        /// <param name="state">The current <see cref="InputState"/>.</param>
+        /// <param name="newValue">The new value for this axis.</param>
+        public void HandleAxisChange(InputState state, float newValue)
+        {
+            PropagateEvent(InputQueue, new JoystickAxisMoveEvent(state, new JoystickAxis(source, newValue), state.Joystick.Axes[source - InputAxis.FirstJoystickAxis]));
```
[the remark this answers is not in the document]
Jeremiidesu: I'll have to take a closer look for that one, but I think I'm passing at least two different values (the `newValue` and the one from the state). It may very well be possible that I swapped some things while testing and forgot to revert thinking everything looked good

Jeremiidesu:
My main workflow is to get features done and document later ; considering this code base I'll admit it's not the easiest of tasks to find what I worked on and properly document it. I'll make sure everything gets documented, accepting changes in wording like you suggested here as I'm clearly not the best as describing what is/does what.

osu.Framework/Input/JoystickAxisEventManager.cs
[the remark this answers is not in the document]
frenzibyte: Well, the new value is set before the event reached here, save it and pass it through as I said above:
image

frenzibyte:
Upon testing on my Switch Pro Controller, it seems to be working fine, but *rarely* it gets stuck on axes value `0.06f` and sometimes jumps between `0.06f` and dead-zone values.

Seems to be that the dead-zone threshold is a bit lower than it should be:
https://github.com/ppy/osu-framework/blob/258e538ca1c9925417daf56f507a3c4a60ba9cde/osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs#L178

Changing it to `0.075f` seems to fix the issues, showing axes values from `0.10f` and above. @Jeremiidesu does it seems like a sane value or is it too much?

Jeremiidesu:
That makes sense to me. In the event in which my DS4 sticks were amazingly centered when I first connected it, I encountered a similar issue. I'll change the value to 0.10f.

There's also a whole other issue with deadzones and axis origins which prevent triggers from being used (trigger axes start at -1, causing deadzones to be >1 every time, rendering the axis unusable) but I'd consider that to be out of the scope of this PR

osu.Framework/Input/JoystickAxisEventManager.cs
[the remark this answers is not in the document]
Jeremiidesu: Good catch! Added a `lastValue` passed through everything like so: 

```cs
var lastValue = state.Joystick.Axes[(int)a.Source];
state.Joystick.Axes[(int)a.Source] = a.Value;
handler.HandleInputStateChange(new JoystickAxisChangeEvent(state, this, a, lastValue));

File: osu.Framework.Tests/Visual/Input/TestSceneJoystick.cs [hunk 16]
```diff
@@ -40,7 +40,7 @@ public TestSceneJoystick()
             for (int i = 0; i < 4; i++)
                 hatFlow.Add(new JoystickHatHandler(i));
 
-            for (int i = 0; i < 64; i++)
+            for (int i = 0; i < 16; i++)
```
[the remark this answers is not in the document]
Jeremiidesu: `JoystickDevice` being a private class, `MAX_AXES` cannot be accessed from here. Instead, I'll use `JoystickState.MAX_AXES` as both the class and the constant are public

Jeremiidesu:
Aside from the way I decided to pass the last value through and possible missed documentation spots, I feel like this should be good to go.

osu.Framework/Input/Events/JoystickEvent.cs
```diff
@@ -20,8 +22,9 @@ protected JoystickEvent([NotNull] InputState state)
         public IEnumerable<JoystickButton> PressedButtons => CurrentState.Joystick.Buttons;
 
         /// <summary>
-        /// List of joystick axes.
+        /// List of joystick axes. Axes which have zero value may be omitted.
         /// </summary>
-        public IReadOnlyList<float> Axes => CurrentState.Joystick.Axes;
+        public IEnumerable<JoystickAxis> Axes =>
+            CurrentState.Joystick.GetAxes().Where(j => !Precision.AlmostEquals(0, j.Value));
```
frenzibyte, line 28: With the deadzone threshold we applied, precision equality shouldn't be necessary?

Jeremiidesu: Ah, yeah I guess. Too used to using these kinds of comparisons against zero due to Rider suggesting to "fix" float comparisons that I didn't really think much about it.

frenzibyte:
> There's also a whole other issue with deadzones and axis origins which prevent triggers from being used (trigger axes start at -1, causing deadzones to be >1 every time, rendering the axis unusable) but I'd consider that to be out of the scope of this PR

Yeah as we'll eventually move from osuTK to SDL, I've looked up for methods and there seem to be ones possible for fixing this: https://wiki.libsdl.org/SDL_GameControllerAxis. 

Can be opened as a seperate issue (and possibly marked as `fixed-in-sdl`?) after this is merged in.

osu.Framework/Input/StateChanges/Events/JoystickAxisChangeEvent.cs
```diff
@@ -0,0 +1,27 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Input.States;
+
+namespace osu.Framework.Input.StateChanges.Events
+{
+    public class JoystickAxisChangeEvent : InputStateChangeEvent
+    {
+        /// <summary>
+        /// The current joystick axis data
```
frenzibyte, line 11:
```suggestion
        /// The current joystick axis data.
```

osu.Framework/Input/StateChanges/JoystickAxisInput.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        public JoystickAxisInput(JoystickAxis axis)
+            : this(axis.Yield())
+        {
+        }
+
+        public JoystickAxisInput(IEnumerable<JoystickAxis> axes)
+        {
+            if (axes.Count() > JoystickState.MAX_AXES)
+                throw new ArgumentException("Too many axes in the passed collection", nameof(axes));
```
frenzibyte, line 29:
```suggestion
                throw new ArgumentException($"The length of the provided axes collection ({axes.Count()}) exceeds the maximum length ({JoystickState.MAX_AXES})", nameof(axes));
```

osu.Framework/Input/JoystickAxisEventManager.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Linq;
+using JetBrains.Annotations;
+using osu.Framework.Graphics;
+using osu.Framework.Input.Events;
+using osu.Framework.Input.States;
+using osu.Framework.Logging;
+
+namespace osu.Framework.Input
+{
+    public class JoystickAxisEventManager
```
frenzibyte, line 15: This is fine for now, don't want this PR to be complicated here but I would abstract `ButtonEventManager` to `EventManager` to avoid duplicating `InputQueue` and `PropagateEvent`, can be done at some point where per-joystick-device event managers is work-in-progress.

peppy:
The deadzone value will likely need to be a user configuration variable in the future but that's out of scope of this PR. Otherwise seems quite sane so let's get this in for the time being.

PR #3548, opened 2020-05-15
Fix libbass_fx load errors during tests on linux

smoogipoo:
Anything that touched bass would fail with:
```
Reason: Test host process crashed : /usr/share/dotnet/dotnet: symbol lookup error: /home/smgi/Repos/osu/osu.Game.Rulesets.Mania.Tests/bin/Debug/netcoreapp3.1/runtimes/linux-x64/native/libbass_fx.so: undefined symbol: BASS_GetVersion
```
This also affected nunit tests since they use `HeadlessGameHost` (not `LinuxGameHost`).

### DIFF
--- a/osu.Framework/Platform/Linux/LinuxGameHost.cs
+++ b/osu.Framework/Platform/Linux/LinuxGameHost.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.IO;
-using osu.Framework.Platform.Linux.Native;
 using osu.Framework.Platform.Linux.Sdl;
 using osuTK;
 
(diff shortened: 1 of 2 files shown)


peppy:
Does this mean no tests have ever been run on linux by anyone?

smoogipoo:
No unit tests, at least. Visual tests work fine.

peppy:
Concerning, considering how many linux contributors we have..

bdach:
Can't say I've ran CLI tests on my Linux install anytime recently either. I hardly ever use it anymore at this point.

peppy:
It should also occur from nUnit runs inside of rider / vs though, right?

smoogipoo:
Yeah, in rider it'll show up as "test not run" and give a very unhelpful log file.

PR #3555, opened 2020-05-17
Replace all BASS fail assertions with describing-exceptions

frenzibyte:
Helps investigating ppy/osu#9045.

Not quite sure about the method naming though.

### DIFF
--- a/osu.Framework/Audio/BassUtils.cs
+++ b/osu.Framework/Audio/BassUtils.cs
@@ -1,0 +1,31 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using ManagedBass;
+using osu.Framework.Logging;
+
+namespace osu.Framework.Audio
+{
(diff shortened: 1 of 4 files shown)


osu.Framework/Audio/BassUtils.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System;
+using ManagedBass;
+using osu.Framework.Logging;
+
+namespace osu.Framework.Audio
+{
+    internal static class BassUtils
+    {
+        private static string getFailString() => $@"BASS failed to initialize with error code {Bass.LastError:D}: {Bass.LastError}.";
```
peppy, line 12: any reason this is split out? seems like it would fit in fine inside `CheckFaulted`.

frenzibyte: it is used two times in there, probably should store in a local instead.

peppy:
@frenzibyte made a minor change to better handle initialisation errors. Please check it looks okay.

frenzibyte:
Seems good 👍 , just corrected the fail message I wrote as it was misleading.

PR #3560, opened 2020-05-18
Fix project build warnings & unresolved xmldoc references

frenzibyte:
*warnings*

### DIFF
--- a/osu.Framework/Graphics/Colour/SRGBColour.cs
+++ b/osu.Framework/Graphics/Colour/SRGBColour.cs
@@ -21,8 +21,10 @@ namespace osu.Framework.Graphics.Colour
         public static implicit operator SRGBColour(Color4 value) => new SRGBColour { Linear = value.ToLinear() };
         public static implicit operator Color4(SRGBColour value) => value.Linear.ToSRGB();
 
+#pragma warning disable 618 // can be removed alongside osuTK (Color4/Colour4 implicit conversions)
         public static implicit operator SRGBColour(Colour4 value) => new SRGBColour { Linear = value.ToLinear() };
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Colour/SRGBColour.cs
```diff
@@ -21,8 +21,10 @@ public struct SRGBColour : IEquatable<SRGBColour>
         public static implicit operator SRGBColour(Color4 value) => new SRGBColour { Linear = value.ToLinear() };
         public static implicit operator Color4(SRGBColour value) => value.Linear.ToSRGB();
 
+#pragma warning disable 618 // can be removed alongside osuTK (Color4/Colour4 implicit conversions)
         public static implicit operator SRGBColour(Colour4 value) => new SRGBColour { Linear = value.ToLinear() };
         public static implicit operator Colour4(SRGBColour value) => value.Linear.ToSRGB();
+#pragma warning restore 618
```
bdach, line 27: Not sure about this one - I'd argue we might want this warning to be visible in the transition period.

frenzibyte: Yeah was hesitant there, but that is good point to be raised 👌. I agree

swoolcock: Irrelevant, the next PR will remove the warnings on these operators since `SRGBColour` will use `Colour4` for `Linear`.

swoolcock:
Can we do these changes in the `Colour4` replacement PR instead?

frenzibyte:
Sure, I'll close this PR then.

PR #3564, opened 2020-05-20
Fix audio devices potentially getting initialized more than once

frenzibyte:
As found by the CI https://ci.appveyor.com/project/peppy/osu-framework/builds/32962217/tests
image

## The problem

The problem found by the CI is because calling `setAudioDevice()` twice in a row without `syncAudioDevices()` being invoked (`audioDevices` list must've been filled up first) will cause the audio device to be initialized again and will call `UpdateDevice(index)` with the same index, thus a `Bass.ChannelSetDevice(handle, index)` operation will fail with `Errors.Already`.

This PR fixes the issue by enforcing a list synchronization before performing any initialization steps and checks, also removes unnecessary `device.IsInitialized` check as `deviceIndex == Bass.CurrentDevice` pretty much does its job.
```diff
- if (device.IsInitialized && deviceIndex == Bass.CurrentDevice)
+ if (deviceIndex == Bass.CurrentDevice)
```

I've left the synchronization schedule another `setAudioDevice()` call even though this is called from `setAudioDevice(int)` itself, intended to fix a possible case of `Bass.CurrentDevice` and the requested device to set from the root `setAudioDevice(int)` both becoming disabled after synchronization. (by returning on the root `setAudioDevice(int)` and let the scheduled call from synchronization method change the current device to the default one or no-sound)

## The cause
Currently happening because of a race between an internal thread inside `AudioManager` where `syncAudioDevices()` method is invoked from and the `AudioThread` where a `setAudioDevice()` call is scheduled to from binding with the config bindable.

Here's a diagram illustrating how the race is happening:
setAudioDevicesCalledTwiceBeforeNextSyncAudioDevice()

This race will only happen if `syncAudioDevices()` is called first and updated the `audioDevices` list, making [the `setAudioDevice()` call from config-binding](https://github.com/ppy/osu-framework/blob/1a98edd0cb96bfdfc7da2a3853a369eea8c74898/osu.Framework/Audio/AudioManager.cs#L175) a successful one, [after that](https://github.com/ppy/osu-framework/blob/1a98edd0cb96bfdfc7da2a3853a369eea8c74898/osu.Framework/Audio/AudioManager.cs#L343) another [`setAudioDevice()` call will be scheduled](https://github.com/ppy/osu-framework/blob/1a98edd0cb96bfdfc7da2a3853a369eea8c74898/osu.Framework/Audio/AudioManager.cs#L180-L184) and [its thread will sleep for a whole second](https://github.com/ppy/osu-framework/blob/1a98edd0cb96bfdfc7da2a3853a369eea8c74898/osu.Framework/Audio/AudioManager.cs#L147) before invoking `syncAudioDevices()` again, therefore invoking the scheduled `setAudioDevice()` call before the next sync call and fails with the problem mentioned above.

### DIFF
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -246,7 +246,7 @@ namespace osu.Framework.Audio
                 return false;
 
             // same device
-            if (device.IsInitialized && deviceIndex == Bass.CurrentDevice)
+            if (deviceIndex == Bass.CurrentDevice)
                 return true;
 
             // initialize new device
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -239,6 +239,10 @@ namespace osu.Framework.Audio
 
         private bool setAudioDevice(int deviceIndex)
         {
+            // synchronise audio devices before attempting to
+            // set the requested device, our current list may became stale.
+            syncAudioDevices();
+
             var device = audioDevices.ElementAtOrDefault(deviceIndex);
 
             // device is invalid
@@ -326,37 +330,42 @@ namespace osu.Framework.Audio
             didInitialise = false;
         }
 
+        private readonly object syncDevicesLock = new object();
+
         private void syncAudioDevices()
         {
-            // audioDevices are updated if:
-            // - A new device is added
-            // - An existing device is Enabled/Disabled or set as Default
-            var updatedAudioDevices = EnumerateAllDevices().ToImmutableList();
-            if (audioDevices.SequenceEqual(updatedAudioDevices, updateComparer))
-                return;
-
-            audioDevices = updatedAudioDevices;
-
-            // Bass should always be providing "No sound" device
-            Trace.Assert(audioDevices.Count > 0, "Bass did not provide any audio devices.");
-
-            onDevicesChanged();
-
-            var oldDeviceNames = audioDeviceNames;
-            var newDeviceNames = audioDeviceNames = audioDevices.Skip(1).Where(d => d.IsEnabled).Select(d => d.Name).ToImmutableList();
-
-            var newDevices = newDeviceNames.Except(oldDeviceNames).ToList();
-            var lostDevices = oldDeviceNames.Except(newDeviceNames).ToList();
-
-            if (newDevices.Count > 0 || lostDevices.Count > 0)
+            lock (syncDevicesLock)
             {
-                eventScheduler.Add(delegate
+                // audioDevices are updated if:
+                // - A new device is added / A stale device is removed
+                // - An existing device is Enabled/Disabled or set as Default
+                var updatedAudioDevices = EnumerateAllDevices().ToImmutableList();
+                if (audioDevices.SequenceEqual(updatedAudioDevices, updateComparer))
+                    return;
+
+                audioDevices = updatedAudioDevices;
+
+                // Bass should always be providing "No sound" device
+                Trace.Assert(audioDevices.Count > 0, "Bass did not provide any audio devices.");
+
+                onDevicesChanged();
+
+                var oldDeviceNames = audioDeviceNames;
+                var newDeviceNames = audioDeviceNames = audioDevices.Skip(1).Where(d => d.IsEnabled).Select(d => d.Name).ToImmutableList();
+
+                var newDevices = newDeviceNames.Except(oldDeviceNames).ToList();
+                var lostDevices = oldDeviceNames.Except(newDeviceNames).ToList();
+
+                if (newDevices.Count > 0 || lostDevices.Count > 0)
                 {
-                    foreach (var d in newDevices)
-                        OnNewDevice?.Invoke(d);
-                    foreach (var d in lostDevices)
-                        OnLostDevice?.Invoke(d);
-                });
+                    eventScheduler.Add(delegate
+                    {
+                        foreach (var d in newDevices)
+                            OnNewDevice?.Invoke(d);
+                        foreach (var d in lostDevices)
+                            OnLostDevice?.Invoke(d);
+                    });
+                }
             }
         }
 


peppy:
Nothing in your explanation explains why the fix fixes the logic. From what I can tell, your PR will cause the initialisation to happen twice, the second time `Free`ing the `Init`ing a device in a potentially bad state (or maybe not)?

I can't understand how this relates to the issue you mention.

frenzibyte:
> Nothing in your explanation explains why the fix fixes the logic. From what I can tell, your PR will cause the initialisation to happen twice.

Figured this will happen (probably just another bad explanation there), to simplify things out, I've found out that it is possible for audio devices to be initialized more than once, and that is when `setAudioDevice()` is called more than once without the `audioDevices` list being updated, that is due to the `device.IsInitialized` check not becoming `true` on the other calls of `setAudioDevice()` until another `syncAudioDevices()` is called updating the `audioDevices` list.

This scenario did happen and i've illustrated how it happened in the diagram above (`setAudioDevice()` getting called two times without the list updated).

What this PR does is it fixes the logic to properly early-return on the described scenario above. and as I've said in the description, `device.IsInitialized` check is pretty much redundant so I just removed it.

> the second time `Free`ing the `Init`ing a device in a potentially bad state (or maybe not)?

I can't understand what you mean by that?

peppy:
The attached exception shows the second run hitting "Already". By removing the IsInitialized check, under the same scenario, by normal logic, it would Free then Initi rather than early returning. Is that not the case?

frenzibyte:
I don't think so, it should not bypass the `deviceIndex == Bass.CurrentDevice` check, as soon as the device is first-time initialized `Bass.CurrentDevice` will directly point to it. if the device was previously initialized then [`Bass.CurrentDevice` will be manually set to it](https://github.com/ppy/osu-framework/blob/1a98edd0cb96bfdfc7da2a3853a369eea8c74898/osu.Framework/Audio/AudioManager.cs#L260), so that check should be enough.

bdach:
I think I can understand what this is aiming to fix, but I don't think the fix is correct. It seems like removing the `isInitialized` check altogether will in fact free and re-init a device again, if it is initialised already but has a different index than `Bass.CurrentDevice` (see logic immediately succeeding the change in this PR). I'm not sure how possible that scenario is, but the current fix feels avoidably unsafe.

I think better alternatives to this would be:

* relaxing the `&&` check currently in `master` to a `||`, or
* explicitly calling `syncAudioDevices()` before trying to set.

I have not tested any of the above suggestions in any way, so buyer beware.

frenzibyte:
@bdach I mean, if the device is already initiailized, we shouldn't be early-returning right? we should set it as the current device and continue?

as for calling `syncAudioDevices` I initially planned on that, but I was afraid of making things convoluted, so I decided to just remove the `device.IsInitialized` check and call it a day.

bdach:
> I mean, if the device is already initiailized, we shouldn't be early-returning right? we should set it as the current device and continue?

Yeah that first suggestion wasn't a really good one. You could maybe add a code path that checks `IsInitialized` after the check edited here, which always only sets the device and early-returns if it is initialised, but that seems a little redundant when looking at surrounding code.

I'm not sure how adding an explicit sync is more convoluted than the other options, though. It definitely would make things *read* more obvious.

frenzibyte:
Ok, maybe it won't be that much of a problem for explicit syncing, I'll use that instead.

frenzibyte:
@bdach I've pushed changes, please check through them if they're correct now.

osu.Framework/Audio/AudioManager.cs
```diff
@@ -326,37 +330,42 @@ protected void FreeBass()
...
+
         private void syncAudioDevices()
         {
-            // audioDevices are updated if:
-            // - A new device is added
-            // - An existing device is Enabled/Disabled or set as Default
-            var updatedAudioDevices = EnumerateAllDevices().ToImmutableList();
-            if (audioDevices.SequenceEqual(updatedAudioDevices, updateComparer))
-                return;
+            lock (syncDevicesLock)
```
bdach, line 337: I'm pretty sure this entire method doesn't need to be (or even shouldn't be) under a lock. The actual critical section is the read & write of `audioDevices` and the write to `audioDeviceNames`.

osu.Framework/Audio/AudioManager.cs
```diff
@@ -326,37 +330,42 @@ protected void FreeBass()
...
-            audioDevices = updatedAudioDevices;
+                audioDevices = updatedAudioDevices;
 
-            // Bass should always be providing "No sound" device
-            Trace.Assert(audioDevices.Count > 0, "Bass did not provide any audio devices.");
+                // Bass should always be providing "No sound" device
+                Trace.Assert(audioDevices.Count > 0, "Bass did not provide any audio devices.");
 
-            onDevicesChanged();
+                onDevicesChanged();
```
bdach, line 351: can't be called here just like that. you have a closed loop now (`syncAudioDevices()` -> `onDevicesChanged()` -> potential scheduled `setAudioDevice()` -> `syncAudioDevices()` again...)

frenzibyte: pointed that in the description:
> I've left the synchronization schedule another `setAudioDevice()` call even though this is called from `setAudioDevice(int)` itself, intended to fix a possible case of `Bass.CurrentDevice` and the requested device to set from the root `setAudioDevice(int)` both becoming disabled after synchronization. (by returning on the root `setAudioDevice(int)` and let the scheduled call from synchronization method change the current device to the default one or no-sound)

If I don't call `onDevicesChanged` then we may potentially end up with no device set until the list changes again.

osu.Framework/Audio/AudioManager.cs
frenzibyte: what about scheduling event invocations, I was aware of that potentially happening (`OnNewDevice` invoked 2 times instead of once), unless the lock is expensive here I had no reason not to lock.

frenzibyte:
Closing this PR as https://github.com/ppy/osu-framework/pull/3700 is reworking the whole thing anyways, rendering this pointless.

PR #3573, opened 2020-05-22
Replace usage of Color4 with Colour4

swoolcock:
Required as part of osuTK removal.

### DIFF
--- a/osu.Framework.Tests/Visual/Containers/TestSceneMasking.cs
+++ b/osu.Framework.Tests/Visual/Containers/TestSceneMasking.cs
@@ -58,7 +58,7 @@ namespace osu.Framework.Tests.Visual.Containers
         {
             Add(new Box
             {
-                Colour = Color4.Black,
+                Colour = Colour4.Black,
                 Size = new Vector2(22, 4),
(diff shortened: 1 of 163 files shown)


osu.Framework.Tests/Visual/Containers/TestSceneMasking.cs
```diff
@@ -469,7 +467,7 @@ static Drawable createMaskingBox(float scale, float testCornerExponent)
                             Hollow = true,
                             Radius = 50,
                             Roundness = 50,
-                            Colour = new Color4(0, 255, 255, 255),
+                            Colour = new Colour4(0, 255, 255, 255),
```
turbedi, line 470: While you are at it, this can be changed to `Colour = Colour4.Aqua`

smoogipoo:
`Interpolation` requires `Color4` variants to be defined, otherwise osu! crashes at startup without migrating to `Colour4` (you can test by adding `CS0618` to `NoWarn` in `Directory.Build.props`).

swoolcock:
Closing for now, will likely do this in small chunks when we are ready.

PR #3574, opened 2020-05-24
Add pdb to DynamicClassCompiler to show more info in stack trace

Fuewburvpoa:
Currently after your code have been dynamically compiled and if it throws an exception then you will see stack trace but without file and line where this exception happened.

Adding pdb stream to compilation will show that information even after dynamic assembly been created.

You can see the difference in image below.

image

### DIFF
--- a/osu.Framework/Testing/DynamicClassCompiler.cs
+++ b/osu.Framework/Testing/DynamicClassCompiler.cs
@@ -155,18 +155,18 @@ namespace osu.Framework.Testing
                     options
                 );
 
-                using (var symbols = new MemoryStream())
-                using (var ms = new MemoryStream())
+                using (var pdbStream = new MemoryStream())
+                using (var peStream = new MemoryStream())
                 {
-                    var compilationResult = compilation.Emit(ms, symbols);
+                    var compilationResult = compilation.Emit(peStream, pdbStream);
 
                     if (compilationResult.Success)
                     {
-                        ms.Seek(0, SeekOrigin.Begin);
-                        symbols.Seek(0, SeekOrigin.Begin);
+                        peStream.Seek(0, SeekOrigin.Begin);
+                        pdbStream.Seek(0, SeekOrigin.Begin);
 
                         CompilationFinished?.Invoke(
-                            Assembly.Load(ms.ToArray(), symbols.ToArray()).GetModules()[0].GetTypes().LastOrDefault(t => t.FullName == targetType.FullName)
+                            Assembly.Load(peStream.ToArray(), pdbStream.ToArray()).GetModules()[0].GetTypes().LastOrDefault(t => t.FullName == targetType.FullName)
                         );
                     }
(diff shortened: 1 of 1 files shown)


osu.Framework/Testing/DynamicClassCompiler.cs
```diff
@@ -147,22 +148,25 @@ private async Task recompileAsync(Type targetType, string changedFile)
 
                 var compilation = CSharpCompilation.Create(
                     dynamicNamespace,
-                    requiredFiles.Select(file => CSharpSyntaxTree.ParseText(File.ReadAllText(file), parseOptions, file))
+                    requiredFiles.Select(file => CSharpSyntaxTree.ParseText(File.ReadAllText(file, Encoding.UTF8), parseOptions, file, encoding: Encoding.UTF8))
```
bdach, line 151: Is there a reason the encodings are hard-specified now? The previous version would attempt to auto-detect (at least [when reading the source](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.readalltext?view=netcore-3.1#System_IO_File_ReadAllText_System_String_)).

osu.Framework/Testing/DynamicClassCompiler.cs
```diff
@@ -147,22 +148,25 @@ private async Task recompileAsync(Type targetType, string changedFile)
...
                                  // Compile the assembly with a new version so that it replaces the existing one
                                  .Append(CSharpSyntaxTree.ParseText($"using System.Reflection; [assembly: AssemblyVersion(\"{assemblyVersion}\")]", parseOptions)),
                     references,
                     options
                 );
 
+                using (var symbols = new MemoryStream())
                 using (var ms = new MemoryStream())
```
bdach, line 159: I'd probably rename the other stream to make it clearer that it contains the actual compilation output.

osu.Framework/Testing/DynamicClassCompiler.cs
Fuewburvpoa: Not sure about File.ReadAllText.
But CSharpSyntaxTree.ParseText require encoding to be explicitly specified, or symbol stream will fail to read it, and since i've added it to CSharpSyntaxTree.ParseText, i thought that making File.ReadAllText with explicit encoding would be good to make sure both of them are on the same encoding.

bdach: I see. In that case I agree this is probably the best way to go forward as I'd be hoping nobody is actually doing UTF-32 for source files.

osu.Framework/Testing/DynamicClassCompiler.cs
Fuewburvpoa: using (var pdbStream = new MemoryStream())
using (var peStream = new MemoryStream())
Parameter names from compilation.Emit() method will be fine?

PR #3575, opened 2020-05-25
Add method to open an arbitrary subdirectory in native explorer

peppy:


osu.Framework/Platform/Storage.cs
```diff
@@ -123,6 +123,11 @@ public virtual Storage GetStorageForDirectory(string path)
         /// <summary>
         /// Opens a native file browser window to the root path of this storage.
         /// </summary>
-        public abstract void OpenInNativeExplorer();
+        public virtual void OpenInNativeExplorer() => OpenPathInNativeExplorer(string.Empty);
```
frenzibyte, line 126: Doesn't need to be virtual IMO, `OpenPathInNativeExplorer` should be enough to handle that.
```suggestion
        public void OpenInNativeExplorer() => OpenPathInNativeExplorer(string.Empty);
```

peppy: See the osu!-side PR.

smoogipoo:
Why do we need two virtual methods? Couldn't you expose only the `PathIn` version and override that?

peppy:
Ah, my bad. Should be fine without.

PR #3587, opened 2020-05-28
Avoid Activator.CreateInstance when getting bound copy of BindableList

peppy:
Why didn't we do this from the start? Did something change?


smoogipoo:
It was likely done this way to support inheritance, similar to `Bindable<T>`, but I think this is fine for `BindableList<T>`.

smoogipoo:
One thing we may consider is doing the same for `Bindable<T>` and making it virtual (maybe even in both cases). Can be a consideration for when bindables are separated into their own package.

PR #2821, opened 2019-09-04
Fix incorrect working directory when a nuget framework consumer is consumed by another project

peppy:
osu.Game.Rulesets.Custom.Tests -> osu.Game -> osu.Framework would cause the working directory to become the location of the `osu.Game.dll` in nuget cache. This was due to an incorrect lookup order for `GetEntryAssembly`.

Full testing:

```

osu.Framework.Tests (rider)

HostAssembly C:\Users\Dean\Projects\osu-framework\osu.Framework.Tests\bin\Debug\netcoreapp2.2\osu.Framework.Tests.dll
Assembly.GetEntryAssembly() C:\Users\Dean\Projects\osu-framework\osu.Framework.Tests\bin\Debug\netcoreapp2.2\osu.Framework.Tests.dll
Assembly.GetCallingAssembly() C:\Users\Dean\Projects\osu-framework\osu.Framework.Tests\bin\Debug\netcoreapp2.2\osu.Framework.dll

osu.Framework.Tests (nUnit rider)

HostAssembly C:\Users\Dean\Projects\osu-framework\osu.Framework.Tests\bin\Debug\netcoreapp2.2\osu.Framework.dll
Assembly.GetEntryAssembly() C:\Users\Dean\.nuget\packages\microsoft.testplatform.testhost\16.2.0\lib\netstandard1.5\testhost.dll
Assembly.GetCallingAssembly() C:\Users\Dean\Projects\osu-framework\osu.Framework.Tests\bin\Debug\netcoreapp2.2\osu.Framework.dll

osu.Game (rider)

HostAssembly C:\Users\Dean\Projects\osu\osu.Desktop\bin\Debug\netcoreapp2.2\osu!.dll
Assembly.GetEntryAssembly() C:\Users\Dean\Projects\osu\osu.Desktop\bin\Debug\netcoreapp2.2\osu!.dll
Assembly.GetCallingAssembly() C:\Users\Dean\.nuget\packages\ppy.osu.framework\2019.830.1\lib\netstandard2.0\osu.Framework.dll

osu.Game (nUnit rider)

HostAssembly C:\Users\Dean\Projects\osu\osu.Game.Tests\bin\Debug\netcoreapp2.2\osu.Game.dll
Assembly.GetEntryAssembly() C:\Users\Dean\.nuget\packages\microsoft.testplatform.testhost\16.2.0\lib\netstandard1.5\testhost.dll
[...]

PR #3589, opened 2020-05-29
Replace returning host assembly with NUnit current test assembly

frenzibyte:
Makes more sense to return the assembly containing the current running test as the entry assembly, instead of the game assembly. 

Also an upside with this change is that `DebugUtils.GetEntryAssembly()` will now return the exact same assembly value for normal runs on nunit runs as can be seen in the below results.

---

Only slightly improves code quality by avoiding use of mutable static members, not needed for any other PR but worth submitting this change for the sake of code quality.

Tested assembly values manually on different projects:
```
osu.Framework.Tests (normal run)
- Assembly.GetEntryAssembly(): /Users/salman/Desktop/osu-framework/osu.Framework.Tests/bin/Debug/netcoreapp3.1/osu.Framework.Tests.dll
- Assembly.GetCallingAssembly(): /Users/salman/Desktop/osu-framework/osu.Framework.Tests/bin/Debug/netcoreapp3.1/osu.Framework.dll

osu.Framework.Tests (NUnit run)
- nunit_test_assembly: /Users/salman/Desktop/osu-framework/osu.Framework.Tests/bin/Debug/netcoreapp3.1/osu.Framework.Tests.dll
- Assembly.GetEntryAssembly(): /Users/salman/Desktop/osu-framework/osu.Framework.Tests/bin/Debug/netcoreapp3.1/testhost.dll
- Assembly.GetCallingAssembly(): /Users/salman/Desktop/osu-framework/osu.Framework.Tests/bin/Debug/netcoreapp3.1/osu.Framework.dll

osu.Game.Tests (normal run)
- Assembly.GetEntryAssembly(): /Users/salman/Desktop/osu/osu.Game.Tests/bin/Debug/netcoreapp3.1/osu.Game.Tests.dll
- Assembly.GetCallingAssembly(): /Users/salman/Desktop/osu/osu.Game.Tests/bin/Debug/netcoreapp3.1/osu.Framework.dll

osu.Game.Tests (NUnit run)
- nunit_test_assembly: /Users/salman/Desktop/osu/osu.Game.Tests/bin/Debug/netcoreapp3.1/osu.Game.Tests.dll
- Assembly.GetEntryAssembly(): /Users/salman/Desktop/osu/osu.Game.Tests/bin/Debug/netcoreapp3.1/testhost.dll
- Assembly.GetCallingAssembly(): /Users/salman/Desktop/osu/osu.Game.Tests/bin/Debug/netcoreapp3.1/osu.Framework.dll

osu.Game.Rulesets.Tau.Tests (normal run)
- Assembly.GetEntryAssembly(): /Users/salman/Desktop/tau/osu.Game.Rulesets.tau.Tests/bin/Debug/netcoreapp3.0/osu.Game.Rulesets.tau.Tests.dll
- Assembly.GetCallingAssembly(): /Users/salman/Desktop/tau/osu.Game.Rulesets.tau.Tests/bin/Debug/netcoreapp3.0/osu.Framework.dll

osu.Game.Rulesets.Tau.Tests (NUnit run)
- nunit_test_assembly: /Users/salman/Desktop/tau/osu.Game.Rulesets.tau.Tests/bin/Debug/netcoreapp3.0/osu.Game.Rulesets.tau.Tests.dll
- Assembly.GetEntryAssembly(): /Users/salman/Desktop/tau/osu.Game.Rulesets.tau.Tests/bin/Debug/netcoreapp3.0/testhost.dll
- Assembly.GetCallingAssembly(): /Users/salman/Desktop/tau/osu.Game.Rulesets.tau.Tests/bin/Debug/netcoreapp3.0/osu.Framework.dll
```

### DIFF
--- a/osu.Framework/Development/DebugUtils.cs
+++ b/osu.Framework/Development/DebugUtils.cs
@@ -27,7 +27,7 @@ namespace osu.Framework.Development
             {
                 Debug.Assert(IsNUnitRunning);
 
-                var testName = TestContext.CurrentContext.Test.FullName;
+                var testName = TestContext.CurrentContext.Test.ClassName;
                 return AppDomain.CurrentDomain.GetAssemblies().First(asm => asm.GetType(testName) != null);
             }
         );
--- a/osu.Framework/Development/DebugUtils.cs
+++ b/osu.Framework/Development/DebugUtils.cs
@@ -12,8 +12,6 @@ namespace osu.Framework.Development
 {
     public static class DebugUtils
     {
-        internal static Assembly HostAssembly { get; set; }
-
         public static bool IsNUnitRunning => is_nunit_running.Value;
 
         private static readonly Lazy<bool> is_nunit_running = new Lazy<bool>(() =>
@@ -22,6 +20,15 @@ namespace osu.Framework.Development
 
                 // when running under nunit + netcore, entry assembly becomes nunit itself (testhost, Version=15.0.0.0), which isn't what we want.
                 return entry == null || entry.Location.Contains("testhost");
+            }
+        );
+
+        private static readonly Lazy<Assembly> nunit_test_assembly = new Lazy<Assembly>(() =>
+            {
+                Debug.Assert(IsNUnitRunning);
+
+                var testName = TestContext.CurrentContext.Test.FullName;
+                return AppDomain.CurrentDomain.GetAssemblies().First(asm => asm.GetType(testName) != null);
             }
         );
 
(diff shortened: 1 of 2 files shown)


peppy:
have you tested via `dotnet test`, visual studio unit tests pane, rider unit test pane? i believe these also caused differences.

frenzibyte:
Good point, the NUnit cases from the above result are from Rider NUnit test pane, results of `dotnet test` and Visual Studio look to be equal to rider test:
- `dotnet test`
image

- VS test pane
image

---

- Rider test pane

image

osu.Framework/Development/DebugUtils.cs
```diff
@@ -25,6 +23,15 @@ public static class DebugUtils
             }
         );
 
+        private static readonly Lazy<Assembly> nunit_test_assembly = new Lazy<Assembly>(() =>
+            {
+                Debug.Assert(IsNUnitRunning);
+
+                var testName = TestContext.CurrentContext.Test.ClassName;
```
peppy, line 30: have you tried using `TestContext.CurrentContext.TestDirectory` instead?

frenzibyte: Not sure what you mean, this method is for returning the assembly of the current NUnit test, we need the class type name to find its assembly?

bdach: But you're looking up the assembly to find its location...?

Actually, better yet, considering these lines:

https://github.com/ppy/osu-framework/blob/550468e0f91801af741ee8e123a260abd50e8e9a/osu.Framework/Development/DebugUtils.cs#L63-L69

isn't this path effectively dead? It seems like this would never actually get read...

frenzibyte: Both `GetEntryAssembly()` and `GetEntryPath()` may be used for different purposes. 

I wouldn't say no to remove the `GetEntryAssembly()` method altogether. but currently, returning `testhost` on `GetEntryAssembly()` and returning the test directory on `GetEntryPath()` sounds kind of weird as they may stay at different locations.

bdach: Looking at the usages it does make more sense to return the assembly under test here, rather than `testhost` - but I'd be in favour of eliminating the `IsDebugBuild` branch in `GetEntryPath()` (as long as it is equivalent, which I haven't checked).

frenzibyte: Yeah but I was aware of looking up for the NUnit test assembly by checking whether the test class is in any of the loaded assemblies when I can easily return the test directory from the test context, though it shouldn't really be expensive to do such things for once, will remove.

peppy:
Not sure if this is an improvement or even expected behaviour here. @smoogipoo thoughts?

osu.Framework/Development/DebugUtils.cs
smoogipoo, line 30: It's not a blocked but I'm not sure about using the `ClassName`. Technically it's safe but I wonder if using `FullName` is more correct.

smoogipoo:
This makes sense to me, what's your concern?

peppy:
I guess the rationale is that the "entry assembly" when running via nunit should be considered as always the same as the active test fixture?

Just checking because this is changing logic (a previous case where it would return `osu.Game` would now return `osu.Game.Tests` if the `Game` type was used from `osu.Game`, for isntance).

smoogipoo:
Going through history, I'd already documented that as an issue (https://github.com/ppy/osu-framework/pull/2821#issuecomment-528154480), so I think this is giving the expected results.

peppy:
No worries then, let's go with this.

osu.Framework/Development/DebugUtils.cs
frenzibyte: `FullName` will only conclude the method name if running a specific test case method, I can't find any way to make a class method be of another assembly from the class itself, so this seems safe unless C# changes itself.

PR #3590, opened 2020-05-30
Replace any potential usage of Environment.CurrentDirectory with an immutable static property

frenzibyte:
Using `Environment.CurrentDirectory` for storing / reading files is dangerous as the current directory is mutable and can be changed when performing certain operations (like opening solutions in roslyn type reference builder for example).

This fixes the issue by introducing a new immutable static `StartupDirectory` property (and `StartupStorage` for components that perform disk mutation operations in the startup directory) to use in place of `Environment.CurrentDirectory` uses.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -510,13 +510,9 @@ namespace osu.Framework.Platform
                 Trace.Listeners.Add(new ThrowingTraceListener());
 
                 var assembly = DebugUtils.GetEntryAssembly();
-                string assemblyPath = DebugUtils.GetEntryPath();
 
                 Logger.GameIdentifier = Name;
                 Logger.VersionIdentifier = assembly.GetName().Version.ToString();
-
-                if (assemblyPath != null)
-                    Environment.CurrentDirectory = assemblyPath;
 
                 Dependencies.CacheAs(this);
 
--- a/osu.Framework.Tests/Platform/PortableInstallationTest.cs
+++ b/osu.Framework.Tests/Platform/PortableInstallationTest.cs
@@ -11,7 +11,7 @@ namespace osu.Framework.Tests.Platform
     [TestFixture]
     public class PortableInstallationTest
     {
-        private Storage startupStorage => RuntimeInfo.StartupStorage;
+        private readonly Storage startupStorage = new NativeStorage(RuntimeInfo.StartupDirectory);
 
         [Test]
         public void TestPortableInstall()
--- a/osu.Framework.Tests/Platform/PortableInstallationTest.cs
+++ b/osu.Framework.Tests/Platform/PortableInstallationTest.cs
@@ -11,49 +11,51 @@ namespace osu.Framework.Tests.Platform
     [TestFixture]
     public class PortableInstallationTest
     {
+        private Storage startupStorage => RuntimeInfo.StartupStorage;
+
         [Test]
         public void TestPortableInstall()
         {
-            Assert.IsFalse(File.Exists(FrameworkConfigManager.FILENAME));
+            Assert.IsFalse(startupStorage.Exists(FrameworkConfigManager.FILENAME));
 
             using (var portable = new HeadlessGameHost(@"portable", portableInstallation: true))
             {
                 portable.Run(new TestGame());
-                Assert.AreEqual(Path.GetFullPath(FrameworkConfigManager.FILENAME), portable.Storage.GetFullPath(FrameworkConfigManager.FILENAME));
+                Assert.AreEqual(startupStorage.GetFullPath(FrameworkConfigManager.FILENAME), portable.Storage.GetFullPath(FrameworkConfigManager.FILENAME));
             }
 
             // portable mode should write the configuration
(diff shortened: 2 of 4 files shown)


peppy:
I don't understand why `StartupStorage` is required. None of the usages actually need it. It's just increasing the number of code changes made for seemingly no reason. Just keep the path?

frenzibyte:
> I don't understand why `StartupStorage` is required. None of the usages actually need it. It's just increasing the number of code changes made for seemingly no reason. Just keep the path?

I've added `StartupStorage` to serve the purpose of performing `File`/`Directory` operations on relative paths to avoid lengthen code lines with combining startup directory everywhere, by removing `StartupStorage`, existing usages will change into:

## osu!framework
```diff
DesktopGameHost.cs#L39:
- if (IsPortableInstallation || RuntimeInfo.StartupStorage.Exists(FrameworkConfigManager.FILENAME))
+ if (IsPortableInstallation || File.Exists(Path.Combine(RuntimeInfo.StartupDirectory, FrameworkConfigManager.FILENAME)))
```

## osu
```diff
RulesetStore.cs#L157:
- var files = RuntimeInfo.StartupStorage.GetFiles($"{ruleset_library_prefix}.*.dll");
+ var files = Directory.GetFiles(Path.Combine(RuntimeInfo.StartupDirectory, $"{ruleset_library_prefix}.*.dll"));

OsuTestScene.cs#L122:
- localStorage = new Lazy<Storage>(() => RuntimeInfo.StartupStorage.GetStorageForDirectory($"{GetType().Name}-{Guid.NewGuid()}"));
+ localStorage = new Lazy<Storage>(() => new NativeStorage(Path.Combine(RuntimeInfo.StartupDirectory, $"{GetType().Name}-{Guid.NewGuid()}")));

CustomDataDirectoryTest.cs#L39:
- string defaultStorageLocation = RuntimeInfo.StartupStorage.GetFullPath(Path.Combine("headless", nameof(TestDefaultDirectory)));
+ string defaultStorageLocation = Path.Combine(RuntimeInfo.StartupDirectory, "headless", nameof(TestDefaultDirectory));

CustomDataDirectoryTest.cs#L49:
- private string customPath { get; } = RuntimeInfo.StartupStorage.GetFullPath("custom-path");
+ private string customPath => Path.Combine(RuntimeInfo.StartupDirectory, "custom-path");

CustomDataDirectoryTest.cs#L56:
- string defaultStorageLocation = RuntimeInfo.StartupStorage.GetFullPath(Path.Combine("headless", nameof(TestCustomDirectory)));
+ string defaultStorageLocation = Path.Combine(RuntimeInfo.StartupDirectory, "headless", nameof(TestCustomDirectory));

CustomDataDirectoryTest.cs#L87:
- string defaultStorageLocation = RuntimeInfo.StartupStorage.GetFullPath(Path.Combine("headless", nameof(TestSubDirectoryLookup)))
+ string defaultStorageLocation = Path.Combine(RuntimeInfo.StartupDirectory, "headless", nameof(TestSubDirectoryLookup));

CustomDataDirectoryTest.cs#L139:
- string defaultStorageLocation = RuntimeInfo.StartupStorage.GetFullPath(Path.Combine("headless", nameof(TestMigration)));
+ string defaultStorageLocation = Path.Combine(RuntimeInfo.StartupDirectory, "headless", nameof(TestMigration));
```

*`CustomDataDirectoryTest.cs` seems to have been shortened with combining though...*

bdach:
It is not the main purpose of storages to perform path operations. It does not need to be exposed in this way, much less as a static instance. I am also in favour of removing it.

peppy:
Those changes look good.

frenzibyte:
> Those changes look good.

Alright as that's fine for you, will keep this in mind for consideration in future PRs.

peppy:
Still not sure about the storage usage in the test scene, but it's local enough that it's not really an issue.

@smoogipoo requesting a quick check on this to make sure you can't think/find any fail scenarios.

osu.Framework/Platform/GameHost.cs
```diff
@@ -510,14 +510,10 @@ public void Run(Game game)
                 Trace.Listeners.Add(new ThrowingTraceListener());
 
                 var assembly = DebugUtils.GetEntryAssembly();
-                string assemblyPath = DebugUtils.GetEntryPath();
 
                 Logger.GameIdentifier = Name;
                 Logger.VersionIdentifier = assembly.GetName().Version.ToString();
 
-                if (assemblyPath != null)
-                    Environment.CurrentDirectory = assemblyPath;
```
smoogipoo, line 519: I'm thinking maybe we should leave this here, not sure if any other projects are relying on the working directory currently...

peppy: Yeah, wasn't sure about this. I think it may actually be better that we *don't* change it going forward (I think that may be better expected behaviour for consumers). Hard to say what the correct direction is... maybe remove and see if anyone complains?

frenzibyte: I've removed it because it's not something we need anymore, also to enforce displaying bugs when there are still somehow usages to `Environment.CurrentDirectory` in osu!, would agree with @peppy and wait for complaints.

PR #3594, opened 2020-06-01
Block possible ways of resizing a DrawSizePreservingFillContainer

frenzibyte:
Avoids mistaking that resizing a DSPFC doesn't work as a bug, while it is intended for the container to always fill itself to its parent.

Can be bypassed by casting to `Container` or otherwise, but it does its job for normal usages so (I think it's fine?) to at least have it in its current shape as a warning of misusage for consumers.

### DIFF
--- a/osu.Framework/Graphics/Containers/DrawSizePreservingFillContainer.cs
+++ b/osu.Framework/Graphics/Containers/DrawSizePreservingFillContainer.cs
@@ -37,7 +37,7 @@ namespace osu.Framework.Graphics.Containers
                 RelativeSizeAxes = Axes.Both,
             });
 
-            RelativeSizeAxes = Axes.Both;
+            base.RelativeSizeAxes = Axes.Both;
         }
 
         protected override void Update()
@@ -67,6 +67,40 @@ namespace osu.Framework.Graphics.Containers
 
             content.Size = Vector2.Divide(Vector2.One, content.Scale);
         }
+
+        #region Size modification guards
+
+        public new Axes RelativeSizeAxes
+        {
+            get => base.RelativeSizeAxes;
+            set => throw new MustFillToParentException();
+        }
+
+        public new Axes AutoSizeAxes
+        {
+            get => base.AutoSizeAxes;
+            set => throw new MustFillToParentException();
+        }
+
+        public new Vector2 Size
+        {
+            get => base.Size;
+            set => throw new MustFillToParentException();
+        }
+
+        public new float Width
+        {
+            get => base.Width;
+            set => throw new MustFillToParentException();
+        }
+
+        public new float Height
+        {
+            get => base.Height;
+            set => throw new MustFillToParentException();
+        }
+
+        #endregion
     }
 
     /// <summary>
@@ -101,5 +135,14 @@ namespace osu.Framework.Graphics.Containers
         /// matched while aspect ratio of children is disregarded.
         /// </summary>
         Separate,
+    }
+
+    public class MustFillToParentException : InvalidOperationException
+    {
+        public MustFillToParentException()
+            : base($"This must fill up to the parent's determined {nameof(CompositeDrawable.ChildSize)}, "
+                   + "please modify the size of the parenting container instead.")
+        {
+        }
     }
 }


osu.Framework/Graphics/Containers/DrawSizePreservingFillContainer.cs
```diff
@@ -67,6 +67,40 @@ protected override void Update()
 
             content.Size = Vector2.Divide(Vector2.One, content.Scale);
         }
+
+        #region Size modification guards
+
+        public new Axes RelativeSizeAxes
```
jorolf, line 73: Why didn't you ``override`` these properties instead of creating new ones?

frenzibyte: Setting `base.RelativeSizeAxes` to `Axes.Both` on the constructor will automatically update the `Width` and `Height` components, if we override those properties, it'll crash as if `Width` / `Height` are being set.

The above can be fixed by having a check that can be set to true after we're done with setting up the size of the container, but then, it's kind-of unexpected that this component as a `Drawable` will crash when changing its size.

What I would see as a more correct solution to this is adding a marker interface / boolean that would be used as a check against whether it is allowed to change this `Drawable`'s size, but would currently look over-engineered if done just for this.

jorolf: Couldn't you call ``base.Height = 1; base.Width = 1`` to supress ``RelativeSizeAxis`` doing it?

frenzibyte: That's another workaround to the above case, but as I said it's kind-of unexpected that the component will crash when changing its size.
> The above can be fixed by having a check that can be set to true after we're done with setting up the size of the container, but then, it's kind-of unexpected that this component as a Drawable will crash when changing its size.
>
> What I would see as a more correct solution to this is adding a marker interface / boolean that would be used as a check against whether it is allowed to change this Drawable's size, but would currently look over-engineered if done just for this.

jorolf: I don't like how were differentiating between a "user" setting the size and another component settings the size. One is crashing and the other is undefined behavior?

frenzibyte: That's what I said in the description:
> Can be bypassed by casting to Container or otherwise, but it does its job for normal usages so (I think it's fine?) to at least have it in its current shape as a warning of misusage for consumers.

Currently only notifies as a warning of misusage, if it really is necessary to fix that then a marker interface / boolean property needs to be introduced for it.

osu.Framework/Graphics/Containers/DrawSizePreservingFillContainer.cs
```diff
@@ -102,4 +136,13 @@ public enum DrawSizePreservationStrategy
         /// </summary>
         Separate,
     }
+
+    public class MustFillToParentException : InvalidOperationException
```
bdach, line 140:
```suggestion
    public class MustFillToParentException : NotSupportedException
```

fits better

osu.Framework/Graphics/Containers/DrawSizePreservingFillContainer.cs
```diff
@@ -102,4 +136,13 @@ public enum DrawSizePreservationStrategy
...
+    {
+        public MustFillToParentException()
+            : base($"This must fill up to the parent's determined {nameof(CompositeDrawable.ChildSize)}, "
+                   + "please modify the size of the parenting container instead.")
```
bdach, line 144: this doesn't really tell a consumer all that well what is going on

```suggestion
            : base($"A {nameof(DrawSizePreservingFillContainer)} will automatically fill its parent's {nameof(CompositeDrawable.ChildSize)}). "
                   + "Modifying its dimensions is not supported.")
```

osu.Framework/Graphics/Containers/DrawSizePreservingFillContainer.cs
bdach: I'm wondering if only allowing these size members to be set to "sane" values (i.e. not throwing when `RelativeSizeAxes = Axes.Both`, `Width = 1`, `Height = 1`, etc. are set, but throwing for any other value) wouldn't make this better and allow eliminating the `new` members.

jorolf: I think this is just going to lead to more confusion (Especially if undocumented). What's wrong with just throwing on any usages outside of this class?

bdach: If you try to do this the obvious way (via overrides), this happens:

```
Unhandled exception. osu.Framework.Graphics.Containers.MustFillToParentException: This must fill up to the parent's determined ChildSize, please modify the size of the parenting container instead.
   at osu.Framework.Graphics.Containers.DrawSizePreservingFillContainer.set_Width(Single value) in D:\Open Source\osu-framework\osu.Framework\Graphics\Containers\DrawSizePreservingFillContainer.cs:line 94
   at osu.Framework.Graphics.Drawable.set_RelativeSizeAxes(Axes value) in D:\Open Source\osu-framework\osu.Framework\Graphics\Drawable.cs:line 768
   at osu.Framework.Graphics.Containers.CompositeDrawable.set_RelativeSizeAxes(Axes value) in D:\Open Source\osu-framework\osu.Framework\Graphics\Containers\CompositeDrawable.cs:line 1641
   at osu.Framework.Graphics.Containers.DrawSizePreservingFillContainer..ctor() in D:\Open Source\osu-framework\osu.Framework\Graphics\Containers\DrawSizePreservingFillContainer.cs:line 40
```

This is because the base `RelativeSizeAxes` setter _in `Drawable` itself_ uses the virtual `Width`/`Height` that throw now.

frenzibyte: @bdach Very good suggestion, I just remembered there are projects that set properties to their default values (including size properties).

osu.Framework/Graphics/Containers/DrawSizePreservingFillContainer.cs
frenzibyte: Classic english issues from me :P

Though slight change in your documentation, shouldn't 
 - "A DrawSizePreservingFillContainer will automatically **fill** its parent's ChildSize" 

become 
 - "A DrawSizePreservingFillContainer will automatically **fill up to** its parent's **determined** ChildSize"?

bdach: Either's clear enough to me I suppose.

osu.Framework/Graphics/Containers/DrawSizePreservingFillContainer.cs
```diff
@@ -67,6 +67,98 @@ protected override void Update()
...
+
+        public override Axes RelativeSizeAxes
+        {
+            get => base.RelativeSizeAxes;
+            set
+            {
+                if (RelativeSizeAxes == value) return;
+
+                // Allow filling-to-parent values.
+                if (value == Axes.Both)
```
jorolf, line 81: Can you add xml-doc that describes this special behavior?

frenzibyte: XMLDocs usually state what a property does, I don't think this really needs to have one. (if a consumer would actually read it from a `DrawSizePreservingFillContainer`)

Issue #3591, opened 2020-05-30
Log overlay sometimes displays messages out of order


Jeremiidesu:
![image](https://user-images.githubusercontent.com/20820425/83341298-1d884580-a2e2-11ea-884a-c2da7a528434.png)

The log in the Rider console never gets out of order, but the log overlay sometimes displays messages out of order. (Notice how the first two lines are swapped here in the log overlay)

peppy:
Just to confirm, you're comparing to the console output not log file, right?

Jeremiidesu:
Yes, I'm comparing the console output to the log overlay.

PR #3597, opened 2020-06-02
Fix LogOverlay potentially showing entries out of order

peppy:
Closes https://github.com/ppy/osu-framework/issues/3591.


osu.Framework/Graphics/Visualisation/LogOverlay.cs
```diff
@@ -73,18 +74,22 @@ protected override void LoadComplete()
             });
         }
 
+        private static int logPosition;
```
smoogipoo, line 77: Is there a reason for this being static?

peppy: no

PR #3600, opened 2020-06-03
Update inspectcode version used in CI

smoogipoo:


smoogipoo:
So I've changed everything to use asserts, just going to say ahead of time that I wholeheartedly disagree with this. Asserts like this rub me the long way, and I don't consider resharper ignores as an alternative.

peppy:
is `if (value is NonNullable resolvedValue)` a valid alternative? or just explicitly re-throwing a nullref in place of the assert `if (value == null) throw NullReferenceException(value)`?

I'm also fine with either of those (though the first one likely isn't so correct since we are implying guaranteed non-null.

smoogipoo:
Please let me know if you're fine with the changes.

osu.Framework/Platform/GameHost.cs
```diff
@@ -510,7 +512,7 @@ public void Run(Game game)
                 var assembly = DebugUtils.GetEntryAssembly();
 
                 Logger.GameIdentifier = Name;
-                Logger.VersionIdentifier = assembly.GetName().Version.ToString();
+                Logger.VersionIdentifier = assembly.GetName().Version?.ToString() ?? Logger.VersionIdentifier;
```
peppy, line 515: not a huge fan of this one but it'll do

osu.Framework/Statistics/DotNetRuntimeListener.cs
```diff
@@ -22,22 +22,25 @@ protected override void OnEventWritten(EventWrittenEventArgs data)
         {
             switch ((EventType)data.EventId)
             {
-                case EventType.GCStart_V1:
+                case EventType.GCStart_V1 when data.Payload != null:
```
peppy, line 25: another thing i'm really not sure about is `when`, but given it's only in this one file i guess we can leave it for now.

Issue #3223, opened 2020-01-31
GlyphStore should not crash for files that can't be loaded


smoogipoo:
This could happen if:
1. The file isn't included in resources, or is misproperly named by the user.
2. The file is not of the correct file type.

Every other resource store fails safely by generally returning nulls.

PR #3603, opened 2020-06-04
Fix font store lookups potentially crashing instead of returning null

peppy:
Closes https://github.com/ppy/osu-framework/issues/3223


osu.Framework/IO/Stores/GlyphStore.cs
```diff
@@ -72,7 +73,7 @@ public GlyphStore(ResourceStore<byte[]> store, string assetName = null, IResourc
             }
         }, TaskCreationOptions.PreferFairness);
 
-        public bool HasGlyph(char c) => Font.Characters.ContainsKey(c);
+        public bool HasGlyph(char c) => Font?.Characters.ContainsKey(c) == true;
 
         public int GetBaseHeight() => Font.Common.Base;
```
smoogipoo, line 78: What about all the other areas in this file where `Font` is used? Here, the overload, `GetKerning()`, `Get(string)`, `GetAvailableResources()`...

peppy: What would you expect those to return? I thought exception was most correct since they are undefined. `Get(string)` is an important one though, to meet rough expectations.

smoogipoo: Was thinking along the lines of `?? 0`... If the point is for this to not crash the app then I think it's warranted?
For example if a change to SpriteText was made that queried `GetBaseHeight()` before first checking if it can retrieve a non-null character, would you expect that to crash?

peppy: Fair enough, will change the remaining.

PR #3612, opened 2020-06-11
Implement mouse polling for SDL2

swoolcock:
Rather than using a `ScheduledDelegate` as in `OsuTKMouseHandler`, it instead directly polls the mouse position in the window loop in `Sdl2WindowBackend`.

An `EnableMousePolling` bindable is provided that controls whether polling should occur, and polling is skipped if the mouse is within the window.

If a new mouse position has been detected outside the window, it will fire a `MouseMove` event with coordinates relative to the window frame.

### DIFF
--- a/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
+++ b/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
@@ -254,8 +254,6 @@ namespace osu.Framework.Platform.Sdl
 
         private SDL.SDL_WindowFlags windowFlags => SdlWindowHandle == IntPtr.Zero ? 0 : (SDL.SDL_WindowFlags)SDL.SDL_GetWindowFlags(SdlWindowHandle);
 
-        public bool EnableMousePolling { get; set; } = true;
-
         #endregion
 
         #region IWindowBackend.Events
@@ -339,7 +337,7 @@ namespace osu.Framework.Platform.Sdl
 
                 processEvents();
 
-                if (EnableMousePolling && !mouseInWindow)
+                if (!mouseInWindow)
                     pollMouse();
 
                 OnUpdate();
--- a/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
+++ b/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
@@ -24,6 +24,9 @@ namespace osu.Framework.Platform.Sdl
 
         private readonly Scheduler scheduler = new Scheduler();
 
+        private bool mouseInWindow;
+        private Point previousPolledPoint = Point.Empty;
+
         #region Internal Properties
 
         internal IntPtr SdlWindowHandle { get; private set; } = IntPtr.Zero;
@@ -251,6 +254,8 @@ namespace osu.Framework.Platform.Sdl
 
         private SDL.SDL_WindowFlags windowFlags => SdlWindowHandle == IntPtr.Zero ? 0 : (SDL.SDL_WindowFlags)SDL.SDL_GetWindowFlags(SdlWindowHandle);
 
+        public bool EnableMousePolling { get; set; } = true;
+
         #endregion
 
         #region IWindowBackend.Events
@@ -334,6 +339,9 @@ namespace osu.Framework.Platform.Sdl
 
                 processEvents();
 
+                if (EnableMousePolling && !mouseInWindow)
+                    pollMouse();
+
                 OnUpdate();
             }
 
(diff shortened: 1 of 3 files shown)


osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -251,6 +254,8 @@ private void checkCurrentDisplay()
 
         private SDL.SDL_WindowFlags windowFlags => SdlWindowHandle == IntPtr.Zero ? 0 : (SDL.SDL_WindowFlags)SDL.SDL_GetWindowFlags(SdlWindowHandle);
 
+        public bool EnableMousePolling { get; set; } = true;
```
smoogipoo, line 257: This seems kinda weird to me. In the first place we won't immediately be using it, but if we did need it, why couldn't it be a flag in `MousePositionAbsoluteInput`? That way it becomes the `InputManager`'s responsibility to determined whether the events should be discarded or not or even apply some custom handling.

swoolcock: We are immediately using it though.  It's what allows the mouse cursor to be tracked outside the window.

smoogipoo: I mean a `false` value. I don't think we currently do that?

swoolcock: Pretty sure it's default.  Start up osu! and move the mouse cursor slightly outside the window bounds and watch the rendered cursor follow it offscreen.

smoogipoo: As in yes, I don't think we ever set it `false`. This is providing the ability to do that, but I'm saying maybe going through `MousePositionAbsoluteInput` is a better option if we even want to do this in the first place.
I'd remove this for now (always keep it effectively "true"), and consider adding into `MousePositionAbsoluteInput` when there's a case that needs it.

Also just thinking about it but consumers could already do this by blocking their own cursor from moving / blocking input.

swoolcock: Adding a `Polled` bool to `MousePositionAbsoluteInput` sound alright?

smoogipoo: Maybe `IsInWindow`? Would need to be in both `MousePositionAbsoluteInput` and `MousePositionChangeEvent`.

swoolcock: Default to `true`, or `false`?

smoogipoo: Does it need to be either? I'd expect the input handlers to be able to set the correct values.

swoolcock: As it stands, the SDL2 window backend directly instantiates `MousePositionAbsoluteInput` instances and fires events with them.  `MouseHandler` then passes it into `PendingInputs.Enqueue`.  This was done to avoid using osuTK types.

smoogipoo:
```diff
diff --git a/osu.Framework/Input/StateChanges/MousePositionAbsoluteInput.cs b/osu.Framework/Input/StateChanges/MousePositionAbsoluteInput.cs
index d89633421..8525025a4 100644
--- a/osu.Framework/Input/StateChanges/MousePositionAbsoluteInput.cs
+++ b/osu.Framework/Input/StateChanges/MousePositionAbsoluteInput.cs
@@ -19,7 +19,15 @@ public class MousePositionAbsoluteInput : IInput
         /// <summary>
         /// The position which will be assigned to the current position.
         /// </summary>
-        public Vector2 Position;
+        public readonly Vector2 Position;
+
+        public readonly bool IsInWindow;
+
+        public MousePositionAbsoluteInput(Vector2 position, bool isInWindow)
+        {
+            Position = position;
+            IsInWindow = isInWindow;
+        }
 
         public void Apply(InputState state, IInputStateChangeHandler handler)
         {
@@ -30,7 +38,7 @@ public void Apply(InputState state, IInputStateChangeHandler handler)
                 var lastPosition = mouse.IsPositionValid ? mouse.Position : Position;
                 mouse.IsPositionValid = true;
                 mouse.Position = Position;
-                handler.HandleInputStateChange(new MousePositionChangeEvent(state, this, lastPosition));
+                handler.HandleInputStateChange(new MousePositionChangeEvent(state, this, lastPosition.Yx, IsInWindow));
             }
         }
     }
```
Is all I mean.

swoolcock: Yes, I've already done this locally.  However, there are a couple of other places that instantiate it, so they would also need to pass through `false` then.

Additionally, since `MousePositionChangeEvent` will accept the new bool, `MousePositionRelativeInput` will need to also store the flag.

smoogipoo: Ah for other cases (`PassthroughInputManager`?) just pass `true`. And yeah relative input needs it too.

swoolcock: Note that osu! also instantiates `MousePositionAbsoluteInput` for the autopilot mod.  It might be worth just defaulting it to true and only explicitly set false in the places we know it will be.

smoogipoo: Hmm that's a good point actually, it would require an overloaded constructor to not be a breaking change then. But okay that's fine.

smoogipoo: And in fact a constructor is going to be a breaking change anyway, probably has to be a public field unlike my changes.

PR #3614, opened 2020-06-11
Switch TabControl tab if current tab is closed

### LINKED ISSUE #3152: Tab control should select next tab when current tab is closed

crabn3bula:
#3152

### DIFF
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs
@@ -242,8 +242,12 @@ namespace osu.Framework.Graphics.UserInterface
 
             if (tab == SelectedTab)
             {
-                tabMap.TryGetValue(Items.LastOrDefault(), out TabItem<T> lastTab);
-                SwitchTab(tab == lastTab ? -1 : 1);
+                if (Items.Count == 1)
+                    SelectedTab = null;
+                else {
+                    tabMap.TryGetValue(Items.LastOrDefault(), out TabItem<T> lastTab);
+                    SwitchTab(tab == lastTab ? -1 : 1);
+                }
             }
 
             items.Remove(tab.Value);
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs
@@ -231,9 +231,17 @@ namespace osu.Framework.Graphics.UserInterface
 
         /// <summary>
         /// Removes a <see cref="TabItem{T}"/> from this <see cref="TabControl{T}"/>.
-        /// If this <see cref="TabItem{T}"/> is currently selected and is not the only one then
-        /// switches tab to the next or previous one according to the tabs order, otherwise clears <see cref="SelectedTab"/>.
         /// </summary>
+        /// <remarks>
+        /// <para>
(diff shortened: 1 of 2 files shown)


smoogipoo:
Please add a test for this, `TestSceneTabControl` is a good place for it.

smoogipoo:
Seems like CI didn't like your formatting. Please run `dotnet format` inside `TabControl.cs` and `TestSceneTabControl.cs`.

osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -231,6 +231,7 @@ protected virtual void AddTabItem(TabItem<T> tab, bool addToDropdown = true)
 
         /// <summary>
         /// Removes a <see cref="TabItem{T}"/> from this <see cref="TabControl{T}"/>.
+        /// If the tab to be removed is currently selected, then the next one is selected, otherwise the previous one is selected
```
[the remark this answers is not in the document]
crabn3bula: Whoops. Sorry, This is my first attempt to contribution.

osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -240,7 +242,15 @@ protected virtual void RemoveTabItem(TabItem<T> tab, bool removeFromDropdown = t
                 throw new InvalidOperationException($"Cannot remove non-removable tab {tab}. Ensure {nameof(TabItem.IsRemovable)} is set appropriately.");
 
             if (tab == SelectedTab)
-                SelectedTab = null;
+            {
+                if (Items.Count == 1)
+                    SelectedTab = null;
+                else
+                {
+                    tabMap.TryGetValue(Items.LastOrDefault(), out TabItem<T> lastTab);
```
bdach, line 250: I'm internally debating whether this shouldn't be
```suggestion
                    tabMap.TryGetValue(Items.Last(), out TabItem<T> lastTab);
```
Reasoning is:

* Right now rider complains (rightly) that `TryGetValue` could throw due to trying to pass a null key.
* `RemoveTabItem` should never get called if `Items.Count == 0`.

This should never fail, but I think dropping the `OrDefault` would make for a saner (and easier to debug) failure.

crabn3bula:
I've applied the suggestions. Thank you for your patience and time spent, I will try not to make all these mistakes in the future.

bdach:
I've rewritten the xmldoc slightly. Other than that I think this is alright.

osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -240,7 +250,15 @@ protected virtual void RemoveTabItem(TabItem<T> tab, bool removeFromDropdown = t
                 throw new InvalidOperationException($"Cannot remove non-removable tab {tab}. Ensure {nameof(TabItem.IsRemovable)} is set appropriately.");
 
             if (tab == SelectedTab)
-                SelectedTab = null;
+            {
+                if (Items.Count == 1)
+                    SelectedTab = null;
+                else
+                {
+                    tabMap.TryGetValue(Items.Last(), out TabItem<T> lastTab);
```
peppy, line 258: ugh this completely falls over with how `ChannelTabControl` is implemented

peppy: there's no safety to check if the same tab is selected, and the change is being performed before the tab is removed from the `Items` being read here.

PR #3618, opened 2020-06-15
Fix TabControl's updated switching-on-removal logic not considering IsSwitchable

peppy:
Allow's osu!'s `ChannelTabControl` to work again. Without this check, it was possible that the switch call would re-select the tab to be removed.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs
@@ -251,7 +251,7 @@ namespace osu.Framework.Graphics.UserInterface
 
             if (tab == SelectedTab)
             {
-                if (Items.Count == 1)
+                if (TabMap.Values.Count(t => t.IsSwitchable) == 1)
                     SelectedTab = null;
                 else
                 {
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs
@@ -38,7 +38,14 @@ namespace osu.Framework.Graphics.UserInterface
         /// <summary>
         /// A collection of all tabs which are valid switch targets.
         /// </summary>
-        public TabItem<T>[] SwitchableTabs => TabContainer.AllTabItems.Where(tab => tab.IsSwitchable).ToArray();
+        protected internal IEnumerable<TabItem<T>> SwitchableTabs => AllTabs.Where(tab => tab.IsSwitchable);
+
+        protected internal IEnumerable<TabItem<T>> AllTabs => TabContainer.AllTabItems;
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -251,12 +256,13 @@ protected virtual void RemoveTabItem(TabItem<T> tab, bool removeFromDropdown = t
...
             {
-                if (Items.Count == 1)
+                if (SwitchableTabs.Length < 2)
                     SelectedTab = null;
                 else
                 {
-                    tabMap.TryGetValue(Items.Last(), out TabItem<T> lastTab);
-                    SwitchTab(tab == lastTab ? -1 : 1);
+                    // check all tabs as to include self (in correct iteration order)
+                    bool anySwitchableTabsToRight = tabMap.Values.SkipWhile(t => t != tab).Skip(1).Any(t => t.IsSwitchable);
```
smoogipoo, line 264: The order of `.Values` [isn't specified](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.values?view=netcore-3.1).

I also don't like that neither this code nor the original considers the order of the tabs, as defined by `TabContainer.AllTabItems` (see previous implementation of `SwitchTab()`).

osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -312,31 +318,20 @@ public virtual void SwitchTab(int direction, bool wrap = true)
             if (Math.Abs(direction) != 1)
                 throw new ArgumentException("value must be -1 or 1", nameof(direction));
 
-            TabItem<T>[] switchableTabs = TabContainer.AllTabItems.Where(tab => tab.IsSwitchable).ToArray();
-            int tabCount = switchableTabs.Length;
+            // the current selected tab may be an non-switchable tab, so search all tabs for a candidate.
+            // this is done to ensure ordering (ie. if an non-switchable tab is in the middle).
+            var allTabs = tabMap.Values.Where(t => t.IsSwitchable || t == SelectedTab);
```
smoogipoo, line 323: Same here - this should use the order of appearance as the original code did. The fact this doesn't cause a test failure is concerning, to say the least.

peppy: The fact that I can't understand that there even *is* an order is more concerning. There's too many variables and too many ways to access tabs.

smoogipoo: Yeah, agreed. I missed it in the previous changes too. We did discuss removing it, but nothing came of it.

peppy: Added a comment about this, should be resolved now.

PR #3627, opened 2020-06-18
Add trackpad scrolling support for iPadOS

### LINKED ISSUE #3621: Add 2-finger scrolling gesture support on iPadOS
As an extension to #3524, historically it has also been possible on most Apple trackpads to be able to scroll vertical content up and down by using a 2-finger sliding gesture.

In order to align the experiences of osu! on the iPad with the normal system expectations, this scrolling gesture should also be added. :)

While this functionality is given for free via certain UI controls in iPadOS, I might need to do some research to see what the most efficient way of hooking and passing these events (or even just recreating the effect from scratch) through to osu!framework is.

TimOliver:
Closes #3621 

This PR extends the trackpad mouse support on iPadOS by adding a `UIPanGestureRecognizer` object to the main game view, which then converts any scrolling offsets into `MouseScrollRelativeInput` events. Tested and confirmed working in the Test Browser.

It doesn't behave like a native iOS scroll view in that when you over scroll, the content will snap back on its own. If this ends up not being desirable, an alternative might be to make this code simulate a soft mouse click and drag event (Depending on how easy that is to do).

Additionally, the scrolling is quite fast, so it might be necessary to add some damping to it.

### DIFF
--- a/osu.Framework.iOS/Input/IOSMouseHandler.cs
+++ b/osu.Framework.iOS/Input/IOSMouseHandler.cs
@@ -47,7 +47,7 @@ namespace osu.Framework.iOS.Input
                 MaximumNumberOfTouches = 0 // Only enables this for 2 finger swipe, disabling clicking and dragging.
             };
             view.AddGestureRecognizer(panGestureRecognizer);
-            
+
             return true;
         }
 
@@ -75,9 +75,9 @@ namespace osu.Framework.iOS.Input
                 previousPanTranslation = translation;
 
             CGPoint delta = new CGPoint(
-                    translation.X - previousPanTranslation.X,
-                    translation.Y - previousPanTranslation.Y
-                );
+                translation.X - previousPanTranslation.X,
+                translation.Y - previousPanTranslation.Y
+            );
 
             PendingInputs.Enqueue(new MouseScrollRelativeInput
             {
--- a/osu.Framework.iOS/Input/IOSMouseHandler.cs
+++ b/osu.Framework.iOS/Input/IOSMouseHandler.cs
@@ -67,6 +67,8 @@ namespace osu.Framework.iOS.Input
             });
         }
 
+        private const float scroll_rate_adjust = 0.1f;
+
         private void panOffsetUpdated()
         {
             CGPoint translation = panGestureRecognizer.TranslationInView(view);
@@ -89,7 +91,7 @@ namespace osu.Framework.iOS.Input
             PendingInputs.Enqueue(new MouseScrollRelativeInput
             {
                 IsPrecise = true,
-                Delta = delta
+                Delta = delta * scroll_rate_adjust
             });
         }
     }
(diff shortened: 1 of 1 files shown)


peppy:
Right now we don't have scrollBegan / scrollEnded events so it's not possible to do the "snap back" correctly, but for future reference are there gesture recognizers exposed for this that we can use to make it work correctly?

TimOliver:
Okay fair enough. We can explore that later.

Yeah, it should be possible! All gesture recognisers have an internal state system so it's very easy to capture when one starts, changes, and then ends, and apply starting and ending logic as needed.

osu.Framework.iOS/Input/IOSMouseHandler.cs
```diff
@@ -37,15 +39,22 @@ public override bool Initialize(GameHost host)
 
             pointerInteraction = new UIPointerInteraction(mouseDelegate = new IOSMouseDelegate());
             mouseDelegate.LocationUpdated += locationUpdated;
-
             view.AddInteraction(pointerInteraction);
 
+            panGestureRecognizer = new UIPanGestureRecognizer(panOffsetUpdated)
+            {
+                AllowedScrollTypesMask = UIScrollTypeMask.Continuous,
+                MaximumNumberOfTouches = 0 // Only enables this for 2 finger swipe, disabling clicking and dragging.
```
peppy, line 47: do you have a link to documentation explaining this? having it at 0 is a bit counter-intuitive and the comment you added doesn't really clear things up.

swoolcock:
Hopefully a lot of this should be solved for us with `SDL_MultiGestureEvent`, when we move to SDL on iOS.

osu.Framework.iOS/Input/IOSMouseHandler.cs
TimOliver: Sadly not. The API Is so new there's very little documentation and info about its behaviour.

I discovered during testing that this panning gesture will be activated in two cases: if the user drags two fingers without clicking, or clicks down with one and drags with the other. The latter interferes with the drag logic already in osu!

In the iOS terminology, a "touch" is when the user touches the screen, or explicitly clicks the trackpad (Not sure what merely touching the trackpad is called.). By setting a breakpoint on the gesture object, I could see that when clicking the mouse and dragging, internally the number of `UITouch` objects active was 1, and when dragging just two fingers, the number is 0.

So, while the terminology is inconsistent now due to there being different types of touches, basically `MaximumNumberOfTouches = 0` means that only gestures that don't explicitly click the trackpad are recognised.

TimOliver:
> Hopefully a lot of this should be solved for us with `SDL_MultiGestureEvent`, when we move to SDL on iOS.

Oh yeah? Is that open source somewhere? I'd be curious to see how they're trying to solve these same problems. 🤣

peppy:
Your code was incorrectly handling the first arriving event (ignoring the value it provided). The iOS [sample code](https://developer.apple.com/documentation/uikit/touches_presses_and_gestures/handling_uikit_gestures/handling_pan_gestures) backs this up.

I added a div 10 multiplier, using our existing [macOS multiplier](https://github.com/ppy/osuTK/blob/master/src/osuTK/Platform/MacOS/MacOSFactory.cs#L37) for reference and comparing with the native Settings app for confirmation.

TimOliver:
Oh whoops. Good catch. I was assuming the first event wouldn't matter since the delta value would have been zero.

Nice! Thanks for doing the rest! :)

peppy:
A single debug run shows that it's always non-zero, and skipping it causes a lag before the scroll begins.

PR #3628, opened 2020-06-19
Disable opacity checks on mobile platforms

peppy:
Has caused a *huge* degradation in performance. Likely affects desktop too, just in a less noticeable way. I know it did on osu-stable (we use `unsafe` to get around this). I'm open to disabling this across desktop as well, until a better solution is applied with benchmarking.

Closes https://github.com/ppy/osu/issues/9307

### DIFF
--- a/osu.Framework/Graphics/OpenGL/Textures/TextureGL.cs
+++ b/osu.Framework/Graphics/OpenGL/Textures/TextureGL.cs
@@ -157,7 +157,7 @@ namespace osu.Framework.Graphics.OpenGL.Textures
         {
             // TODO: Investigate performance issues and remove this once we are certain there's no overhead.
             if (RuntimeInfo.IsMobile)
-                return Opacity.Transparent;
+                return Opacity.Mixed;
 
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/OpenGL/Textures/TextureGL.cs
```diff
@@ -155,6 +155,10 @@ public void Dispose()
 
         protected static Opacity ComputeOpacity(ITextureUpload upload)
         {
+            // TODO: Investigate performance issues and remove this once we are certain there's no overhead.
+            if (RuntimeInfo.IsMobile)
+                return Opacity.Transparent;
```
smoogipoo, line 160: `Mixed` is safest.

peppy: Thoughts on also blocking for desktop platforms for the time being?

smoogipoo: May be best to.

PR #3629, opened 2020-06-19
Fix clashing GUIDs in solution file

peppy:
Fixes the following:

image

### DIFF
--- a/osu-framework.sln
+++ b/osu-framework.sln
@@ -63,7 +63,9 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "FlappyDon.Game.Tests", "osu
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "FlappyDon.Desktop", "osu.Framework.Templates\templates\template-flappy\FlappyDon.Desktop\FlappyDon.Desktop.csproj", "{0309CF11-621A-4F23-8FBA-A583303A8531}"
 EndProject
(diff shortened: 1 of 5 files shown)


peppy:
Yeah, it's a bit like that. I just removed all the conflicting projects and added them back. One exception was that I had to regenerate one of the GUIDs manually because visual studio wasn't willing to do it for me.

osu-framework.sln
```diff
@@ -221,6 +221,12 @@ Global
 		{A95175BB-95D0-44B4-8B82-EABD166943DA}.Release|Any CPU.Build.0 = Release|Any CPU
 		{A95175BB-95D0-44B4-8B82-EABD166943DA}.Release|iPhone.ActiveCfg = Release|Any CPU
 		{A95175BB-95D0-44B4-8B82-EABD166943DA}.Release|iPhoneSimulator.ActiveCfg = Release|Any CPU
+		{7AA1DB5D-78DB-4693-AE50-D6078F5A0CAB}.Debug|Any CPU.ActiveCfg = Debug|iPhoneSimulator
+		{7AA1DB5D-78DB-4693-AE50-D6078F5A0CAB}.Debug|iPhone.ActiveCfg = Debug|iPhone
+		{7AA1DB5D-78DB-4693-AE50-D6078F5A0CAB}.Debug|iPhoneSimulator.ActiveCfg = Debug|iPhoneSimulator
+		{7AA1DB5D-78DB-4693-AE50-D6078F5A0CAB}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{7AA1DB5D-78DB-4693-AE50-D6078F5A0CAB}.Release|iPhone.ActiveCfg = Release|iPhone
+		{7AA1DB5D-78DB-4693-AE50-D6078F5A0CAB}.Release|iPhoneSimulator.ActiveCfg = Release|iPhoneSimulator
```
smoogipoo, line 229: Just as a note, this build config has only been added for one template. It's not going to stop me from merging this but would be good to note in case any weirdness arises.

peppy: They aren't actually enabled anywhere, so I don't think this has any effect

PR #3630, opened 2020-06-19
Fix test cursor not receiving press event when masked away

frenzibyte:
Confused me while testing my touch PR, since a MIMTS starts its mouse position at the actual screen-space zero position, the cursor will be masked away and it will not be built into the positional input queue (`PropagatePos... => ... && !IsMaskedAway`).

But to be honest, I don't see a reason why `ManualInputManager` take screen space positions instead of local ones, most of the usages I'm seeing do `InputManager.MoveMouseTo(someDrawable.ToScreenSpace(...))`, might consider opening an issue for that.


peppy:
For clarification what is MIMTS? `ManualInputManagerTestScene`?

frenzibyte:
yes

peppy:
- Could you please provide reproduction steps as to where this is visible? You have not given any test coverage nor any kind of hint as to what/where this is fixing.
- It takes screen positions because what if the target drawable is not a direct child?

frenzibyte:
> * Could you please provide reproduction steps as to where this is visible? You have not given any test coverage nor any kind of hint as to what/where this is fixing.

I've added a manual test case for reproducing this issue, moving a mouse from masked position while holding mouse left.

> * It takes screen positions because what if the target drawable is not a direct child?

A child out of the content boundaries? that shouldn't be allowed, otherwise not sure what you mean, a snippet might explain.

peppy:
Can an assert not be added to that test?

To re-explains screen coordinates: if it took local, any non-direct-child usages would need to `manualInputManager.MoveTo(manualInputManager.ToLocalSpace(child.Child.ScreenSpaceDrawQuad.TopLeft))`

frenzibyte:
I see, and yeah can do a quick one with checking whether propagation is allowed, will write that.

osu.Framework.Tests/Visual/Testing/TestSceneManualInputManagerTestScene.cs
```diff
@@ -30,6 +33,18 @@ public void TestResetInput()
...
+        [Test]
+        public void TestCursorReceivesEventsWhenMaskedAway()
+        {
+            Drawable cursor = null;
+
+            AddStep("retrieve cursor", () => cursor = InputManager.ChildrenOfType<CursorContainer>().Single().ActiveCursor);
+
+            AddStep("move mouse to screen zero", () => InputManager.MoveMouseTo(Vector2.Zero));
+            AddAssert("ensure cursor masked away", () => cursor.IsMaskedAway);
+            AddAssert("cursor allows event propagation", () => cursor.PropagatePositionalInputSubTree);
```
bdach, line 45: This test doesn't seem to make much sense. Of course the two asserts will be true, but that doesn't actually say anything about what kind of *behaviour* the test wants to enforce.

The previous one made more sense (it tested actual behaviour). Modifying it to something like this "works" and has an assert without having to expose too much, but it borders on hacky:

```csharp
        [Test]
        public void TestHoldLeftFromMaskedPosition()
        {
            AddStep("move mouse to screen zero", () => InputManager.MoveMouseTo(Vector2.Zero));
            AddStep("hold both mouse buttons", () =>
            {
                InputManager.PressButton(MouseButton.Left);
                InputManager.PressButton(MouseButton.Right);
            });
            AddStep("move mouse to content center", () => InputManager.MoveMouseTo(Content.ScreenSpaceDrawQuad.Centre));
            AddAssert("all cursor parts visible",
                () => InputManager.ChildrenOfType<CursorContainer>().Single()
                                  .Child
                                  .ChildrenOfType<Container>().All(container => container.Alpha > 0));
        }
```

Not sure if something like this can be done more neatly without exposing things...

frenzibyte: Well, yeah your snippet works but its testing something very specific in the design, that's why I was not sure about writing a test case at all, but IMO testing if the cursor itself allows event propagation tests essentially the same thing, because there shouldn't be any other way for the events not to be received while masked away.

bdach: Then I don't know, some kind of subclassing could be done? I feel like asserting that a property hard-coded to `true` is `true` misses the mark entirely.

frenzibyte: I mean, the test is meant to assert that the cursor will be built into the positional queue (i.e. it should be allowed to receive events), which consists of checking whether `PropagatePositionalInputSubTree` is true and `HandlePositionalInput` is true, should add the other boolean into the check though.

Although since you've said it, I'm not quite sure anymore, may need more input on how this should be tested.

frenzibyte: Now that I look back at this. @bdach pardon me, your suggested test case makes much sense than mine, test cases should be asserting things logically, instead of going straight into `Drawable`'s internals to check if it should receive events, will push.

frenzibyte:
Any concerns left on this PR? think its ready for merge.

peppy (approved):
Seems non-offensive enough.

PR #3634, opened 2020-06-19
Remove GDEdit

Altenhh:
As the project has been recently disbanded due to lack of motivation, it would be best to remove this now as the project no longer exists.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,6 @@ The BASS audio library (a dependency of this framework) is a commercial product.
 
 [osu!](https://github.com/ppy/osu) – rhythm is just a *click* away!
 
-[GDEdit](https://github.com/gd-edit/GDE) - A third-party Geometry Dash editor.
-
 <!--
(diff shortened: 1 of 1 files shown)


peppy:
curious why you'd delete it rather than just let it remain? it was in a pretty nice state, and in the spirit of open source you could have left it for someone else to potentially pick up.

Altenhh:
Fair point. I'll re-upload the code and leave it up to someone else to potentially pick it up. Though i heavily disagree that the code was in a nice state, it was fairly jumbled up and 90% of the code was copied from osu!lazer's code. However i do understand that lazer's code was meant to be an example on how you would use the framework and parts of it could be copied off from.

peppy:
I didn't mean to imply the code was great (I haven't looked into it that much) but it worked well as an application.

I know I lost some projects I worked on in my earlier days and always regretted doing so.

PR #3637, opened 2020-06-21
Fix Sprite.EdgeSmoothness not working

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9326

### DIFF
--- a/osu.Framework/Graphics/Sprites/Sprite.cs
+++ b/osu.Framework/Graphics/Sprites/Sprite.cs
@@ -132,7 +132,7 @@ namespace osu.Framework.Graphics.Sprites
                 if (value.X > MAX_EDGE_SMOOTHNESS || value.Y > MAX_EDGE_SMOOTHNESS)
                 {
                     throw new InvalidOperationException(
-                        $"May not smooth more than {MAX_EDGE_SMOOTHNESS} or will leak neighboring textures in atlas. Tried to smooth by ({EdgeSmoothness.X}, {EdgeSmoothness.Y}).");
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/Sprites/Sprite.cs
```diff
@@ -194,7 +194,8 @@ protected override Quad ComputeScreenSpaceDrawQuad()
 
             Vector3 scale = DrawInfo.MatrixInverse.ExtractScale();
 
-            return ToScreenSpace(DrawRectangle.Inflate(scale.Xy * EdgeSmoothness));
+            InflationAmount = scale.Xy * EdgeSmoothness;
+            return ToScreenSpace(DrawRectangle.Inflate(InflationAmount));
```
frenzibyte, line 198: Correct me If I'm wrong, but shouldn't `InflationAmount` have its own `ComputeInflationAmount` method? right now if just the order of setting properties in `SpriteDrawNode.ApplyState()` changes, it will break?
```diff
-            Texture = Source.Texture;
-            ScreenSpaceDrawQuad = Source.ScreenSpaceDrawQuad;
-            DrawRectangle = Source.DrawRectangle;
-            InflationAmount = Source.InflationAmount;
+            Texture = Source.Texture;
+            InflationAmount = Source.InflationAmount;
+            ScreenSpaceDrawQuad = Source.ScreenSpaceDrawQuad;
+            DrawRectangle = Source.DrawRectangle;
```

smoogipoo: Yeah. The `EdgeSmoothness` property also didn't have an invalidation. Fixed both.

PR #3638, opened 2020-06-21
Add virtual sample channel

bdach:
Step 1 towards addressing ppy/osu#8595.

# Summary

This PR adds a `SampleChannelVirtual`, which analogously to `TrackVirtual` fulfills the code contract of being a sample channel, but in reality does nothing in terms of audio playback. Consumers can use it as an alternative to a null sample channel.

Adding it to framework was suggested by @frenzibyte and I would tend to agree with that suggestion as it keeps things consistent.

# Remarks

I haven't added any tests, but this seems pointless to test as it's a do-nothing class.

I opted for this solution so as not to have to change the game-side skin lookup fallback logic which is complex enough as-is without adding another layer on top.

Not 100% on the `UpdateState()` override. It made sense to me at the time.

### DIFF
--- a/osu.Framework/Audio/Sample/SampleChannelVirtual.cs
+++ b/osu.Framework/Audio/Sample/SampleChannelVirtual.cs
@@ -1,0 +1,28 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework.Audio.Sample
+{
+    /// <summary>
+    /// A <see cref="SampleChannel"/> which explicitly plays no audio.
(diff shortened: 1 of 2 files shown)


osu.Framework/Audio/Sample/SampleChannelVirtual.cs
```diff
@@ -0,0 +1,28 @@
...
+        public SampleChannelVirtual()
+            : base(new SampleVirtual(), _ => { })
+        {
+        }
+
+        public override bool Playing => false;
+
+        protected override void UpdateState()
+        {
+            // empty override to avoid affecting sample channel count in frame statistics
```
peppy, line 21: I think we should let this count, just like virtual tracks are counting.

PR #3640, opened 2020-06-22
Merge text addition logic into one single method

frenzibyte:
Separated from #3289 

Merges `addCharacter(char)` into being inside `InsertString(string)`. Centralizes adding character to text logic into one method.


osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -500,47 +500,39 @@ protected virtual Drawable AddCharacterToFlow(char c)
...
-            text = text.Insert(selectionLeft, c.ToString());
-            selectionStart = selectionEnd = selectionLeft + 1;
+                onDrawableCreated?.Invoke(c, dc);
 
-            cursorAndLayout.Invalidate();
+                text = text.Insert(selectionLeft, c.ToString());
+                selectionStart = selectionEnd = selectionLeft + 1;
 
-            return ch;
+                cursorAndLayout.Invalidate();
```
frenzibyte, line 534: @bdach I've tried wrapping this in a local, but doesn't seem to have helped anything, I think it's fine as-is.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -500,47 +500,39 @@ protected virtual Drawable AddCharacterToFlow(char c)
         /// <summary>
         /// Insert an arbitrary string into the text at the current position.
         /// </summary>
-        /// <param name="text">The new text to insert.</param>
-        protected void InsertString(string text)
+        /// <param name="value">The string of text to insert.</param>
+        /// <param name="onDrawableCreated">An action invoked whenever a new character drawable has been created.</param>
+        protected void InsertString(string value, Action<char, Drawable> onDrawableCreated = null)
```
peppy, line 505: is `onDrawableCreated` going to ever be used anywhere except for the ime thing?

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -500,47 +500,39 @@ protected virtual Drawable AddCharacterToFlow(char c)
...
-            if (selectionLength > 0)
-                removeCharacterOrSelection();
-
-            if (text.Length + 1 > LengthLimit)
-            {
-                NotifyInputError();
-                return null;
-            }
+                if (selectionLength > 0)
+                    removeCharacterOrSelection();
```
peppy, line 524: isn't this in the wrong place now? it used to be run before checking the length limit. this looks like it would heavily break logic. please add tests when you fix, as well.

osu.Framework/Graphics/UserInterface/TextBox.cs
frenzibyte: don't think so, right sounds like it's exposing a callback that shouldn't exist as-is, will privatize.

osu.Framework/Graphics/UserInterface/TextBox.cs
frenzibyte: oh good catch, will do.

PR #3641, opened 2020-06-22
Allow deleting multiple text characters at once

frenzibyte:
Separated from #3289 

Allows deleting multiple characters at one method call, this will remove the unnecessary `sound` parameter used in the IME composition logic.


osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -445,19 +452,21 @@ private bool removeCharacterOrSelection(bool sound = true)
...
+            text = text.Remove(removeStart, removeCount);
 
             // Reorder characters depth after removal to avoid ordering issues with newly added characters.
-            for (int i = start; i < TextFlow.Count; i++)
+            for (int i = removeStart; i < TextFlow.Count; i++)
                 TextFlow.ChangeChildDepth(TextFlow[i], getDepthForCharacterIndex(i));
 
-            if (selectionLength > 0)
-                selectionStart = selectionEnd = selectionLeft;
+            if (removeCount >= selectionLength)
```
bdach, line 461: This is weird to me. You should either be able to delete any amount of characters without a selection, or the entire selection. It shouldn't be legal to call `removeCharacters` with `removeCount != selectionLength` to begin with.

I'd say this needs to be an assertion:

```csharp
Debug.Assert(selectionLength == 0 || removeCount == selectionLength);
```

frenzibyte: Yeah wasn't sure at first, because whenever I thought that initially I get to a conclusion that the two methods should be one, but I think it's fine.

peppy: @frenzibyte planning to address this?

peppy:
Requires non-trivial conflict resolution.

PR #3642, opened 2020-06-22
Add right click support to iPadOS

### LINKED ISSUE #3624: Right click support for mouse cursor on iPadOS

TimOliver:
Closes #3624 

This PR adds the ability to right-click via the system gesture (two fingers at once) on iPadOS. It adds support via the `IOSTouchHandler` class as the `UIView` touch APIs (eg `TouchesBegan`) are the easiest way to capture both the "click down" and "click up" events (The standard tap gesture recognisers only fire after the user has released their finger).

One somewhat inconvenient quirk of UIKit is that while it's possible to check if the click-down event was meant to be a right-click (via checking if the `ButtonMask` value of the supplied `UIEvent` is `Secondary`), the subsequent click-up event don't contain any of this information, and so (As far as I was able to tell), it's not possible to tell if a given click-up event was for a left click or a right click.

To mitigate this, when a right-click-down event is detected, a flag is set to remember this, and when the next click-up event occurs, it assumes this is the previous right click, and triggers a right-click mouse up event. I'm not 100% happy at the reliability of this, but I played with this a lot to see if there were any strange cases that might break it (eg, clicking and tapping the device at the same time), and it seemed fine.


osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -15,25 +15,38 @@ public class IOSTouchHandler : InputHandler
...
         }
 
-        private void handleUITouch(UITouch touch)
+        private void handleUITouch(UITouch touch, UIEvent evt)
         {
+            // Right-click on iOS 13.4 and up
+            MouseButton buttonValue = MouseButton.Left;
+            if (UIDevice.CurrentDevice.CheckSystemVersion(13, 4))
+            {
+                if (evt.ButtonMask == UIEventButtonMask.Secondary)
```
peppy, line 43: what happens if this isn't present? do previous versions fire secondary events?

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -42,11 +55,14 @@ private void handleUITouch(UITouch touch)
             {
                 case UITouchPhase.Moved:
                 case UITouchPhase.Began:
-                    PendingInputs.Enqueue(new MouseButtonInput(MouseButton.Left, true));
+                    PendingInputs.Enqueue(new MouseButtonInput(buttonValue, true));
                     break;
 
                 case UITouchPhase.Cancelled:
                 case UITouchPhase.Ended:
+                    if (rightClickPressed)
```
peppy, line 63: this seems *very* dodgy. what happens if the user does:

leftDown
rightDown
leftUp
rightUp

osu.Framework.iOS/Input/IOSTouchHandler.cs
TimOliver: No, that's an API that was explicitly added for handling right-clicks in iPadOS 13.4
https://developer.apple.com/documentation/uikit/uieventbuttonmask?language=objc

osu.Framework.iOS/Input/IOSTouchHandler.cs
TimOliver: Yeah. Agreed. I actually have a physical mouse here. I can test that and see exactly what happens.

I'll also research if there's a way to associate a "touch-up" event with the initial "touch-down" event so we can guarantee which is which.

peppy: If I recall correctly from many years ago, the incoming `UITouch` reference should match?

osu.Framework.iOS/Input/IOSTouchHandler.cs
peppy: The version check should probably not be required then? Since this is just an enum comparison.

TimOliver: Ahh, I tested that to confirm, and it seems `evt.ButtonMask` is an iOS 13.4 API, so it's definitely needed.

osu.Framework.iOS/Input/IOSTouchHandler.cs
TimOliver: I had a play with that today and you were right! It's possible to compare the two separate `UITouch` objects, and they'll match if they both came from the same click-down and click-up events.

I've refined it now by having an `NSMutableSet` that stores any incoming right-click events, and then checks the set during the "release" phase of the event. `NSMutableSet` is a bit more flexible in that it internally uses the `NSObject` hashing mechanisms for checking equality, so it'll match two touch objects the represent the same event, even if the memory addresses are different.

It's working properly on my Magic Trackpad now, but on my Logitech mouse, clicking both buttons at the same time causes left click to get stuck. I'll try and track that down, and then this should be good to go.

PR #3643, opened 2020-06-22
Add [Values] attribute support in test browser

peppy:
Note that the test is only really useful for visually confirming this behaviour. Hard to really do any other testing.


osu.Framework/Testing/TestBrowser.cs
```diff
@@ -651,4 +682,53 @@ public enum TestBrowserAction
         Reload,
         Search
     }
+
+    public class CartesianProduct<T>
```
frenzibyte, line 686: Leftover class?

peppy: Yep, removed.

osu.Framework/Testing/TestBrowser.cs
```diff
@@ -504,12 +504,43 @@ private void finishLoad(TestScene newTest, Action onCompletion)
                 {
                     string repeatSuffix = i > 0 ? $" ({i + 1})" : string.Empty;
 
-                    if (m.GetCustomAttribute(typeof(TestAttribute), false) != null)
+                    var methodWrapper = new MethodWrapper(m.GetType(), m);
+
+                    if (methodWrapper.GetCustomAttributes<TestAttribute>(false).FirstOrDefault() != null)
```
bdach, line 509: typical nit - I'd either use

```suggestion
                    if (methodWrapper.GetCustomAttributes<TestAttribute>(false).SingleOrDefault() != null)
```

(`AllowMultiple = false` on `TestAttribute` allows this), or use `Any()`.

osu.Framework/Testing/TestBrowser.cs
```diff
@@ -504,12 +504,43 @@ private void finishLoad(TestScene newTest, Action onCompletion)
...
-                        CurrentTest.AddLabel($"{name}{repeatSuffix}");
+                        var parameters = m.GetParameters();
+
+                        if (parameters.Length > 0)
+                        {
+                            var valueMatrix = new List<List<object>>();
+
+                            foreach (var p in methodWrapper.GetParameters())
+                            {
+                                var valueAttrib = p.GetCustomAttributes<ValuesAttribute>(false).FirstOrDefault();
```
bdach, line 519: same here

```suggestion
                                var valueAttrib = p.GetCustomAttributes<ValuesAttribute>(false).SingleOrDefault();
```

PR #3645, opened 2020-06-23
Don't activate run tool window on rider run

peppy:
Been meaning to PR this for a while so I don't need to keep the changes locally. The run window popping up is really annoying in my opinion, so as long as there is consensus I'd like to push this change out across all non-CLI project configurations.

### DIFF
--- a/.idea/.idea.osu-framework.Desktop/.idea/runConfigurations/SampleGame.xml
+++ b/.idea/.idea.osu-framework.Desktop/.idea/runConfigurations/SampleGame.xml
@@ -1,5 +1,5 @@
 <component name="ProjectRunConfigurationManager">
-  <configuration default="false" name="SampleGame" type="DotNetProject" factoryName=".NET Project">
+  <configuration default="false" name="SampleGame" type="DotNetProject" factoryName=".NET Project" activateToolWindowBeforeRun="false">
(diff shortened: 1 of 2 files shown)


smoogipoo:
I don't use the tool window so maybe @bdach ?

peppy:
It appears every run. There's no way to disable it so you must be seeing it.

This, for reference:

smoogipoo:
Ah. I actually like seeing that window, because I don't use keybinds there and/or hide my tabs.

peppy:
For what purpose? Are you actually reading the output? It's pretty useless to me in all cases.

smoogipoo:
The one I have more use for is the build window, but it looks like there's a rider setting to enable that one separately.
So I think I'm fine with this, still want @bdach 's opinion, though.

frenzibyte:
Yeah that window is really annoying with useless information most of the time, I agree with not popping it up at run.

bdach:
I sometimes will read the output of the run window - usually when I'm debugging something that runs in `Update` and I can't breakpoint as it'll affect frame time I'll log things and watch the run output. 

But I wouldn't say I actively use it every time so I wouldn't want to make a big deal out of it. I can always pull it up myself when I need it (and I've been doing dev on Linux recently, so I have `tail -f` as a replacement anyway)

smoogipoo:
Let's go with this for now then.

PR #3647, opened 2020-06-23
Split out bass amplitude computation to separate class

peppy:
Feels better from a code quality perspective, but may also pave the way to using this in `SampleChannel` (something I am currently looking in to.

- [x] Depends on https://github.com/ppy/osu-framework/pull/3646


peppy:
Adding a test scene for this, hold off on checking just yet (there's no way to test amplitudes are working framework side currently).

peppy:
Okay, good to go now.

PR #3648, opened 2020-06-23
Allow retrieving SampleChannel amplitudes

peppy:
- [x] Depends on #3647 

Not sure about the renaming breaking backwards compatibility. Can't add a temporary obsoleted class due to it being a struct.

### DIFF
--- a/osu.Framework/Audio/Sample/SampleChannel.cs
+++ b/osu.Framework/Audio/Sample/SampleChannel.cs
@@ -54,6 +54,6 @@ namespace osu.Framework.Audio.Sample
 
         public override bool IsAlive => base.IsAlive && !Played;
 
-        public virtual ChannelAmplitudes CurrentAmplitudes { get; } = new ChannelAmplitudes();
+        public virtual ChannelAmplitudes CurrentAmplitudes { get; } = ChannelAmplitudes.Empty;
     }
 }
--- a/osu.Framework/Audio/Sample/SampleChannel.cs
+++ b/osu.Framework/Audio/Sample/SampleChannel.cs
@@ -3,6 +3,7 @@
 
 using osu.Framework.Statistics;
 using System;
+using osu.Framework.Audio.Track;
 
 namespace osu.Framework.Audio.Sample
 {
@@ -52,5 +53,12 @@ namespace osu.Framework.Audio.Sample
         public virtual bool Played => WasStarted && !Playing;
 
         public override bool IsAlive => base.IsAlive && !Played;
+
+        /// <summary>
+        /// Current amplitude of stereo channels where 1 is full volume and 0 is silent.
+        /// LeftChannel and RightChannel represent the maximum current amplitude of all of the left and right channels respectively.
+        /// The most recent values are returned. Synchronisation between channels should not be expected.
+        /// </summary>
+        public virtual TrackAmplitudes CurrentAmplitudes { get; } = new TrackAmplitudes();
     }
 }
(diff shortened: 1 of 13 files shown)


smoogipoo:
Fyi this is going to break many current rulesets:
- https://github.com/EVAST9919/bosu
- https://github.com/EVAST9919/lazer-m-vis
- https://github.com/swoolcock/rush
- https://github.com/Altenhh/tau
- https://github.com/LumpBloom7/sentakki

peppy:
Yep. Open to suggestion on how to resolve that as per PR description.

smoogipoo:
Guess it's fine to break things, so long as others know. Don't see an easier way around this.

cc @EVAST9919 @swoolcock @Altenhh @LumpBloom7 you guys will need to re-build your rulesets to target the next osu! release (see the osu!-side PR).

peppy:
At this stage in development users should probably be expecting such issues to occasionally arise. As long as it gracefully fails to load the rulesets (which I think we have handled well these days) it shouldn't be a huge deal.

smoogipoo:
We actually have no error handling for rulesets. This is actually a 100% crash in those rulesets, of the form:

```
[runtime] 2020-06-23 07:24:43 [error]: System.MissingMethodException: Method not found: 'osu.Framework.Audio.Track.TrackAmplitudes osu.Framework.Audio.Track.Track.get_CurrentAmplitudes()'.
[runtime] 2020-06-23 07:24:43 [error]: at osu.Game.Rulesets.Bosu.MusicHelpers.MusicAmplitudesProvider.Update()
[runtime] 2020-06-23 07:24:43 [error]: at osu.Framework.Graphics.Drawable.UpdateSubTree() in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Drawable.cs:line 481
```

peppy:
Might be a good time to add handling of such cases..

peppy:
Shall we skip the rename for the time being, until we have better error handling? I'm fine with adding a TODO for it for a future cleanup effort.

smoogipoo:
Interestingly it seems like only the rulesets that directly reference the amplitudes crash, the others that override a method that doesn't exist just seem to not care.
So that only makes it bosu, m-vis, and sentakki that crash.

I'm fine with breaking things here.

osu.Framework/Audio/Sample/SampleChannel.cs
```diff
@@ -52,5 +53,7 @@ protected override void UpdateState()
         public virtual bool Played => WasStarted && !Playing;
 
         public override bool IsAlive => base.IsAlive && !Played;
+
+        public virtual ChannelAmplitudes CurrentAmplitudes { get; } = new ChannelAmplitudes();
```
smoogipoo, line 57:
```suggestion
        public virtual ChannelAmplitudes CurrentAmplitudes { get; } = ChannelAmplitudes.Empty;
```

osu.Framework/Audio/Track/Track.cs
```diff
@@ -112,12 +112,7 @@ public virtual double Rate
 
         public override bool HasCompleted => IsLoaded && !IsRunning && CurrentTime >= Length;
 
-        /// <summary>
-        /// Current amplitude of stereo channels where 1 is full volume and 0 is silent.
-        /// LeftChannel and RightChannel represent the maximum current amplitude of all of the left and right channels respectively.
-        /// The most recent values are returned. Synchronisation between channels should not be expected.
-        /// </summary>
-        public virtual TrackAmplitudes CurrentAmplitudes { get; } = new TrackAmplitudes();
+        public virtual ChannelAmplitudes CurrentAmplitudes { get; } = new ChannelAmplitudes();
```
smoogipoo, line 115:
```suggestion
        public virtual ChannelAmplitudes CurrentAmplitudes { get; } = ChannelAmplitudes.Empty;
```

PR #3649, opened 2020-06-23
Fix black square around triangle sprites due to front-to-back

smoogipoo:
* Closes ppy/osu#9351

Not a super huge fan of this but at least it's consistent with `Draw()`/`Blit()`...

Also adds the texture availability check, both to be consistent with `Draw()` and to prevent the potential error there.


Joehuu:
This fixes https://github.com/ppy/osu/issues/9351.

bdach:
Looks perfectly fine to me - not sure where's the worry coming from? Is it the virtual dispatch? It seems unavoidable as far as I can tell.

smoogipoo:
Less of a worry and more that I'm not particularly a fan of having `base.DrawOpaqueInterior()` be required to be called. It's consistent but moving the setting of the depth from there is one thing I'm keeping in the back of my mind.

Issue #3306, opened 2020-02-26
Test failure (disposed drawable attempted to load)


peppy:
https://ci.appveyor.com/project/peppy/osu-framework/builds/31063994/tests

PR #3653, opened 2020-06-24
Fix potential ObjectDisposedException in DelayedLoadUnloadWrapper

smoogipoo:
Fixes https://github.com/ppy/osu-framework/issues/3306

Couldn't actually replicate this one (hence no test). It's actually quite hard to do so because you have to juggle 3 schedules: the one inside `DLW.OnInvalidate()`, the one for `checkForUnload()`, and then another one to actually trigger the disposal at the correct time. Then you have to deal with `perUpdateTasks` vs non-`perUpdateTasks` in the `Scheduler` and it gets very implementation-specific.

But I can imagine why this could occur.

### DIFF
--- a/osu.Framework/Graphics/Containers/DelayedLoadUnloadWrapper.cs
+++ b/osu.Framework/Graphics/Containers/DelayedLoadUnloadWrapper.cs
@@ -84,6 +84,14 @@ namespace osu.Framework.Graphics.Containers
             });
         }
 
+        private readonly object unloadLock = new object();
+
+        protected override void Dispose(bool isDisposing)
+        {
+            lock (unloadClock)
+                base.Dispose(isDisposing);
+        }
+
         protected override void CancelTasks()
         {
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Containers/DelayedLoadUnloadWrapper.cs
```diff
@@ -84,6 +84,14 @@ protected override void EndDelayedLoad(Drawable content)
             });
         }
 
+        private readonly object unloadLock = new object();
+
+        protected override void Dispose(bool isDisposing)
+        {
+            lock (unloadClock)
```
peppy, line 91:
```suggestion
            lock (unloadLock)
```

i am very scared by this

peppy:
the lock on a `base` call is pretty ugly. can't think of a better solution without reworking disposal completely (something for another year).

PR #3655, opened 2020-06-24
Add improved state handling for right-click on iPadOS

TimOliver:
This PR is a refinement on top of #3642. 

In my testing, I noticed that if a user had a physical mouse and clicked both buttons down at the same time, the "Unclick" mouse event for the left button wouldn't be called, and the cursor was consistently getting stuck in a "left button still clicked" state.

I discovered today that this was because I had assumed clicking the left and right buttons generated separate streams of  `UITouch` objects, but I found that actually only one `UITouch` object is ever used when a user is clicking a physical mouse, and that single touch object simply transitions between "left-click" and "right-click" states on demand.

This PR does two things:
1) Properly silos all of the custom right-click logic, so regular screen taps aren't affected by it.
2) Detects when the state of the `UITouch` object has changed mid-flight, and performs the necessary "unclick" and "click" events to transition the mouse state in the framework to match the physical mouse.

Tested with both the Magic Keyboard and a Logitech gaming mouse and confirmed working on both.

### DIFF
--- a/osu.Framework.iOS/Input/IOSTouchHandler.cs
+++ b/osu.Framework.iOS/Input/IOSTouchHandler.cs
@@ -14,9 +14,9 @@ namespace osu.Framework.iOS.Input
     public class IOSTouchHandler : InputHandler
     {
         private readonly IOSGameView view;
-        private NSMutableSet<UITouch> pendingRightClickTouches = new NSMutableSet<UITouch>();
 
-        private bool rightClickSupport = UIDevice.CurrentDevice.CheckSystemVersion(13, 4);
+        private UIEventButtonMask lastButtonMask = UIEventButtonMask.Primary;
+        private readonly bool rightClickSupport = UIDevice.CurrentDevice.CheckSystemVersion(13, 4);
 
         public IOSTouchHandler(IOSGameView view)
         {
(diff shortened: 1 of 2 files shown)


osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -32,15 +32,17 @@ private void handleTouches(NSSet obj, UIEvent evt)
 
         private void handleUITouch(UITouch touch, UIEvent evt)
         {
-            var location = touch.LocationInView(null);
+            // Indirect pointer means the touch came from a mouse cursor, and wasn't a physcial touch on the screen
+            bool indirectTouch = (rightClickSupport && touch.Type == UITouchType.IndirectPointer);
```
peppy, line 36: this governs the whole functionality of this method. it should be the top level if-else.

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -32,15 +32,17 @@ private void handleTouches(NSSet obj, UIEvent evt)
...
+            bool rightClickEvent = (rightClickSupport && evt.ButtonMask == UIEventButtonMask.Secondary);
```
peppy, line 37: should be inside the `IndirectPointer` portion of the if-else

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -49,10 +51,46 @@ private void handleUITouch(UITouch touch, UIEvent evt)
...
+                        // A single UITouch object represents the mouse cursor on iPadOS 13.4.
+                        // If the user clicks both left and right buttons on a physical mouse, this doesn't generate more
+                        // touch objects; it just changes the button mask value for the one touch object without calling "Began" or "Ended".
+                        // Without accounting for this, the mouse button input can sometimes be left in a "stuck" state.
+
+                        if (rightClickEvent)
+                        {
+                            // If a right-click event occurred, and the touch wasn't already saved in the right-click set,
+                            // the user has transitioned from left-click to right-click.
+                            if (!pendingRightClickTouches.Contains(touch))
```
peppy, line 68: make function `transitionRightClick(event)`. if it's certain only one buttonmask is active at once, store the last one and perform the (left unclick) transition within that function.

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -49,10 +51,46 @@ private void handleUITouch(UITouch touch, UIEvent evt)
...
+                                pendingRightClickTouches.Add(touch);
+                            }
+
+                            PendingInputs.Enqueue(new MouseButtonInput(MouseButton.Right, true));
+                        }
+                        else
+                        {
+                            // If a left-click event has occurred, but the touch event was already saved in the right-click set,
+                            // the user has transitioned from a right-click event to a left-click.
+                            if (pendingRightClickTouches.Contains(touch))
```
peppy, line 80: we don't have any instances of copy paste coding. please don't copy paste code (same code exists below).

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -49,10 +51,46 @@ private void handleUITouch(UITouch touch, UIEvent evt)
                         PendingInputs.Enqueue(new MouseButtonInput(MouseButton.Left, true));
 
                     break;
+                case UITouchPhase.Moved:
+                    if (!indirectTouch)
+                        PendingInputs.Enqueue(new MouseButtonInput(MouseButton.Left, true));
+                    else
+                    {
+                        // A single UITouch object represents the mouse cursor on iPadOS 13.4.
```
peppy, line 59: is there documentation on this?

does this mean:
- left down
- right down
- right up

leaves the mouse in a state with no presses? the remaining left down is ignored?

TimOliver: Sadly no. There's no documentation on how the `UITouch` events map to the mouse pointer interactions. I worked this out by writing a separate app, and then getting each of the touch events to print a line to the console to see how the life-time of the interaction worked.

`touchesBegan` is only called at the beginning when either of the buttons is pressed down, and `touchesEnded` is only called when both buttons are released. When clicking between the buttons, it triggers `touchesMoved` to indicate the button mask changed, but neither `touchesBegan` or `touchesEnded` is called if one of the buttons is already down.

Okay! I've refactored the code to hopefully be a little more elegant. Please let me know if there's anything else you need for this. :)

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -34,45 +34,87 @@ private void handleUITouch(UITouch touch, UIEvent evt)
...
                     else
                         PendingInputs.Enqueue(new MouseButtonInput(MouseButton.Left, true));
 
                     break;
 
+                case UITouchPhase.Moved:
+                    if (indirectPointerTouch)
+                        transitionRightClick(evt);
+                    else
+                        PendingInputs.Enqueue(new MouseButtonInput(MouseButton.Left, true));
```
peppy, line 60: this is weird. it's triggering a mouse down event every move event.

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -34,45 +34,87 @@ private void handleUITouch(UITouch touch, UIEvent evt)
...
+            bool indirectPointerTouch = (rightClickSupport && touch.Type == UITouchType.IndirectPointer);
+
             PendingInputs.Enqueue(new MousePositionAbsoluteInput { Position = new Vector2((float)location.X * view.Scale, (float)location.Y * view.Scale) });
 
             switch (touch.Phase)
             {
-                case UITouchPhase.Moved:
                 case UITouchPhase.Began:
-                    if (rightClickSupport && evt.ButtonMask == UIEventButtonMask.Secondary)
+                    if (indirectPointerTouch)
```
peppy, line 45: you know you have a code smell if you have if if if if if of the same thing everywhere.

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -14,9 +14,9 @@ namespace osu.Framework.iOS.Input
     public class IOSTouchHandler : InputHandler
     {
         private readonly IOSGameView view;
-        private NSMutableSet<UITouch> pendingRightClickTouches = new NSMutableSet<UITouch>();
 
-        private bool rightClickSupport = UIDevice.CurrentDevice.CheckSystemVersion(13, 4);
+        private UIEventButtonMask lastButtonMask = UIEventButtonMask.Primary;
+        private readonly bool rightClickSupport = UIDevice.CurrentDevice.CheckSystemVersion(13, 4);
```
peppy, line 19: indirectPointerSupport? it's being used for more than just rightClick now (see line 38).

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -14,9 +14,9 @@ namespace osu.Framework.iOS.Input
     public class IOSTouchHandler : InputHandler
     {
         private readonly IOSGameView view;
-        private NSMutableSet<UITouch> pendingRightClickTouches = new NSMutableSet<UITouch>();
 
-        private bool rightClickSupport = UIDevice.CurrentDevice.CheckSystemVersion(13, 4);
+        private UIEventButtonMask lastButtonMask = UIEventButtonMask.Primary;
```
peppy, line 18: this shouldn't start as primary. it's potentially missing a state change.

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -34,45 +34,87 @@ private void handleUITouch(UITouch touch, UIEvent evt)
...
+            if (lastButtonMask != evt.ButtonMask)
+            {
+                PendingInputs.Enqueue(new MouseButtonInput(inactiveButton, false));
+                lastButtonMask = evt.ButtonMask;
+            }
+
+            PendingInputs.Enqueue(new MouseButtonInput(activeButton, true));
+        }
+
+        private bool isRightClick(UIEventButtonMask buttonMask)
```
peppy, line 100: should return a mousebutton so you don't have `? MouseButton.Right : MouseButton.Left;` in 4 places

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -34,45 +34,87 @@ private void handleUITouch(UITouch touch, UIEvent evt)
...
+            MouseButton inactiveButton = activeButton == MouseButton.Right ? MouseButton.Left : MouseButton.Right;
+
+            // A single UITouch object represents the mouse cursor on iPadOS 13.4.
+            // If the user clicks both left and right buttons on a physical mouse, this doesn't generate more
+            // touch objects; it just changes the button mask value for the one touch object without calling "Began" or "Ended".
+            // If the stored mask value doesn't match the active one, this means the user alternated buttons, so unclick
+            // the previous button, and click the new button
+            if (lastButtonMask != evt.ButtonMask)
+            {
+                PendingInputs.Enqueue(new MouseButtonInput(inactiveButton, false));
```
peppy, line 93: might fire an up event even though there was no down event. the lastButtonMask isn't being cleared so the previous touch pollutes the next one.

peppy:
I'm refactoring this at my end. Left a review pointing out the main issues though.

osu.Framework.iOS/Input/IOSTouchHandler.cs
TimOliver: Yep, that's how it was when I found this file (https://github.com/ppy/osu-framework/blob/d45d84b80de366c1028d72efed2a74683d334ca9/osu.Framework.iOS/Input/IOSTouchHandler.cs#L44). I deliberately kept it as is in order to not risk potentially changing the expected behaviour.

peppy:
Regression testing on a mouse would be appreciated, haven't done that yet.

ThatsNiceGuy:
I’ve tested the new iPadOS cursor features with a standard USB mouse (Deathadder Elite) and found that scrolling with the scroll wheel doesn’t work, neither does right clicking with the mouse buttons (It just gets handled as a left click.) 
I’ve also tested with a Magic Trackpad 2 over Bluetooth and while scrolling does work, right clicking with a two finger click doesn’t work for me
lazer version 2020.629.0

peppy:
you understand this isn't merged yet, right?

peppy:
Going to merge this, in testing it seems solid enough.

PR #1682, opened 2018-07-03
Make PassThroughInputManager sync with parent input state

ekrctb:
When `UseParentInput` is true, a `PassThroughInputManager` syncs its keyboard and joystick state to parent `InputManager`'s state.
For mouse buttons, only input that reached event handlers (`OnMouseDown` and `OnMouseUp`) are applied. Click initiated at outside of the `PassThroughInputManager` and blocked inputs are not synced. However, always sync `MouseUp` to maintain an invariant (for a mouse button) `this.pressed <= parent.pressed`. This is because a mouse inputs are positional and this behavior is matching to regular `Drawable`s.
- Closes #1672 
- When `KeyBindingContainer` handles `OnKeyDown` before `KeyBindings` are set there was a null reference. I fixed this. However, the new `KeyBindingContainer` doesn't fire actions that is bind to keys pressed while the new `KeyBindingContainer` is initialized. For example, when entering osu!catch play while pressing an arrow key, the catcher isn't moving initially. I think this is allowable but maybe should fix if people need.

PR #3656, opened 2020-06-25
Clarify properly why mouse buttons are not fully synchronised

frenzibyte:
As per #1682's description.

### DIFF
--- a/osu.Framework/Input/PassThroughInputManager.cs
+++ b/osu.Framework/Input/PassThroughInputManager.cs
@@ -154,8 +154,9 @@ namespace osu.Framework.Input
         /// <param name="parentState">Parent's state. If this is null, it is regarded as an empty state.</param>
         protected virtual void SyncInputState(InputState parentState)
         {
-            // Only release buttons not pressed in parent, synchronizing what parent has already pressed
-            // as well may potentially lead to false mouse click input propagated to children.
+            // invariant: if mouse button is currently pressed, then it has been pressed in parent (but not the converse)
+            // therefore, mouse up events are always synced from parent
+            // mouse down events are not synced to prevent false clicks
             var mouseButtonDifference = (parentState?.Mouse?.Buttons ?? new ButtonStates<MouseButton>()).EnumerateDifference(CurrentState.Mouse.Buttons);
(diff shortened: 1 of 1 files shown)


osu.Framework/Input/PassThroughInputManager.cs
```diff
@@ -154,7 +154,8 @@ public void Sync(bool useCachedParentInputManager = false)
         /// <param name="parentState">Parent's state. If this is null, it is regarded as an empty state.</param>
         protected virtual void SyncInputState(InputState parentState)
         {
-            // release all buttons that is not pressed on parent state
+            // Only release buttons not pressed in parent, synchronizing what parent has already pressed
+            // as well may potentially lead to false mouse click input propagated to children.
```
bdach, line 158: I can't parse this sentence.

How about:

```suggestion
            // invariant: if mouse button is pressed in child, then it is pressed in parent (but not the converse)
            // therefore, mouse up events are always synced from parent
            // mouse down events are not synced to prevent fake delayed clicks
```

frenzibyte: I don't get what it means by `if mouse button is pressed in child, then it is pressed in parent (but not the converse)`, perhaps you got it swapped?

bdach: I meant it precisely as written. Logically speaking we have the following possibilities:

| parent | child | possible |
| :-: | :-: | :-: |
| pressed | pressed | :heavy_check_mark: yes |
| pressed | not pressed | :heavy_check_mark: yes |
| not pressed | pressed | ❌ no |
| not pressed | not pressed | :heavy_check_mark: yes |

Therefore if the button is seen as pressed in the child then it must be pressed in the parent, but when it is not pressed, it's no longer ensured that it isn't pressed in the parent.

frenzibyte: oh so you're referring `child` as the current state and `parent` as the new state, that thing got me confused there, maybe reword to:
```suggestion
            // invariant: if mouse button is currently pressed, then it has been pressed in parent (but not the converse)
            // therefore, mouse up events are always synced from parent
            // mouse down events are not synced to prevent false clicks
```

bdach: sure

PR #3660, opened 2020-06-28
Fix PassThroughInputManager potentially not receiving release events

frenzibyte:
All event managers propagate release events to the handler drawable and all of what its in front of it built in to the queue. And for `TouchEventManager` which propagates touch move events to the queue of `TouchDownEvent` (test coverage about that coming in dependent PR), the pass-through input manager must always stick itself to the queue.

The test case provided for this will not produce different results based on the fix due to the manager actually tries to synchronize itself with the parent manager per-frame because of "some non-positional events" being blocked, which I think this fix would render that redundant, but decided to separate that change into another PR as it also kinda changes behaviour.

### DIFF
--- a/osu.Framework.Tests/Visual/Input/TestScenePassThroughInputManager.cs
+++ b/osu.Framework.Tests/Visual/Input/TestScenePassThroughInputManager.cs
@@ -89,6 +89,21 @@ namespace osu.Framework.Tests.Visual.Input
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


smoogipoo:
I think this is fine... @peppy any second opinion?

peppy:
I also... *think* this is okay. Cannot see any regressions in usage.

PR #3661, opened 2020-06-28
Remove PassThroughInputManager per-update synchronization

frenzibyte:
- [ ] Depends on #3660 (to replace this)

Seems to be unnecessary now after the aforementioned PR is merged, although there's a behaviour change here which is that some input may not be synchronized when there's a drawable in front of the manager absorbing their corresponding event (excluding `MouseMoveEvent` because of using `IRequireHighFrequencyMousePosition`).

Not quite sure if that's the only behaviour change this will result in though, but testing with osu! in both desktop and mobile it seems to be working correctly, especially on gameplay.

### DIFF
--- a/osu.Framework.Tests/Visual/Input/TestScenePassThroughInputManager.cs
+++ b/osu.Framework.Tests/Visual/Input/TestScenePassThroughInputManager.cs
@@ -89,6 +89,21 @@ namespace osu.Framework.Tests.Visual.Input
         }
 
         [Test]
+        public void TestUpReceivedOnDownFromSync()
+        {
+            addTestInputManagerStep();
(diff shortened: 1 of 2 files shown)


frenzibyte:
OK turns out this is required for receiving proper initial states and releasing correctly, will close this for now.

peppy:
If you found this was required, but it wasn't already tests, it would be hugely appreciated if you could add tests covering (and explaining) the behaviour.

frenzibyte:
> If you found this was required, but it wasn't already tests, it would be hugely appreciated if you could add tests covering (and explaining) the behaviour.

It is already tested as shown in https://ci.appveyor.com/project/peppy/osu-framework/builds/33778674, the test `ReceiveInitialState`.

PR #3663, opened 2020-06-28
Disallow mouse input from primary touch on PassThroughInputManagers

frenzibyte:
Actually doesn't seem testable as there's no way to distinguish between whether a redundant input has been applied and whether not, as mouse input from touch will be produced in order from bottom-child up to top-parent.

But makes sense either way to put this toggle for things to be as expected.

### DIFF
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -128,6 +128,11 @@ namespace osu.Framework.Input
 
         private readonly Dictionary<JoystickAxisSource, JoystickAxisEventManager> joystickAxisEventManagers = new Dictionary<JoystickAxisSource, JoystickAxisEventManager>();
 
+        /// <summary>
+        /// Whether to produce mouse input on any primary touch input.
+        /// </summary>
+        protected virtual bool MapMouseToPrimaryTouch => true;
+
         protected InputManager()
         {
             CurrentState = CreateInitialState();
@@ -593,8 +598,7 @@ namespace osu.Framework.Input
                 case TouchStateChangeEvent touchChange:
                     HandleTouchStateChange(touchChange);
 
-                    // primary touch mapped to mouse.
-                    if (touchChange.Touch.Source == TouchSource.Touch1)
+                    if (MapMouseToPrimaryTouch && touchChange.Touch.Source == TouchSource.Touch1)
                     {
                         new MousePositionAbsoluteInput { Position = touchChange.Touch.Position }.Apply(CurrentState, this);
                         new MouseButtonInput(MouseButton.Left, touchChange.State.Touch.IsActive(touchChange.Touch.Source)).Apply(CurrentState, this);
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -569,6 +569,21 @@ namespace osu.Framework.Input
                 manager.HandleButtonStateChange(e.State, active ? ButtonStateChangeKind.Pressed : ButtonStateChangeKind.Released);
         }
 
+        /// <summary>
+        /// Handles primary touch state change event to produce mouse input from.
+        /// </summary>
+        /// <param name="e">The primary touch state change event.</param>
+        /// <returns>Whether mouse input has been performed accordingly.</returns>
+        protected virtual bool HandleMouseTouchStateChange(TouchStateChangeEvent e)
+        {
+            if (!MapMouseToPrimaryTouch)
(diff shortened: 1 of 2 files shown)


osu.Framework/Input/PassThroughInputManager.cs
```diff
@@ -25,6 +25,8 @@ namespace osu.Framework.Input
     /// </remarks>
     public class PassThroughInputManager : CustomInputManager, IRequireHighFrequencyMousePosition
     {
+        protected override bool MapMouseToPrimaryTouch => !UseParentInput;
```
smoogipoo, line 28: This looks very weird to me. So someone can derive PTIM themselves and set this to true and break things?

Maybe a better way is to expose a `HandlePrimaryTouchStateChange` method that is derived by PTIM to do this.

peppy: Wouldn't that still allow breakage? Isn't `sealed` what you're looking for?

smoogipoo: I don't think `sealed` is appropriate, because you're pretty much always going to be using PTIM in code. Therefore it could be desirable to _also_ have this as an option that the user can override.

The way I see it is:
```
IM:
    virtual void HandlePrimaryTouchStateChange():
        if (!MapMouseToPrimaryTouch)
            return;

PTIM:
    override void HandlePrimaryTouchStateChange():
        return;
```

I don't see how that allows breakage. If the user's then overriding it themselves and doing some crazy custom stuff then they take full responsibility for what they're doing - the important part is that it's not PTIM or IM that's breaking things.

peppy: i'm not a huge fan of base blocking, prefer sealed bool myself. i can't think of a case you'd want to toggle it back on a PassThroughInputManager..

smoogipoo: Actually now I'm not sure I agree with the implementation in this PR at all. It seems like this would completely stop touch events from propagating as primary mouse button, but what about `ManualInputManager`? We probably do want it to propagate appropriately there, so I feel like this needs to be tied specifically to when input is received from a parent?

peppy: I haven't checked the content of the PR, but your explanation is how it should behave, correct.

frenzibyte: > I don't think sealed is appropriate, because you're pretty much always going to be using PTIM in code. Therefore it could be desirable to also have this as an option that the user can override.

Yeah good thinking with that, will switch to using that (with slight renaming in `HandlePrimaryTouchStateChange` to being `HandleMouseTouchStateChange` along with xmldoc explaining, and also returning `bool`)

> It seems like this would completely stop touch events from propagating as primary mouse button, but what about ManualInputManager? We probably do want it to propagate appropriately there, so I feel like this needs to be tied specifically to when input is received from a parent?

@smoogipoo I can't get what you mean exactly by that, can you please explain further? what do you mean by "this would completely stop touch event from propagating as primary mouse button", `ManualInputManager` should propagate correctly, when manually enqueuing input, it would automatically switch `UseParentInput` to `false`.

smoogipoo: > it would automatically switch `UseParentInput` to `false`

Gotcha, that's fine then.

osu.Framework/Input/InputManager.cs
```diff
@@ -128,6 +128,11 @@ public Drawable DraggedDrawable
 
         private readonly Dictionary<JoystickAxisSource, JoystickAxisEventManager> joystickAxisEventManagers = new Dictionary<JoystickAxisSource, JoystickAxisEventManager>();
 
+        /// <summary>
+        /// Whether to produce mouse input on any primary touch input.
+        /// </summary>
+        protected virtual bool MapMouseToPrimaryTouch => true;
```
smoogipoo, line 134: I'm not sure this is useful. I mentioned it as option (2) in https://github.com/ppy/osu-framework/issues/3667#issuecomment-651014453, but it isn't really going to allow that since it'd have to be applied to `UserInputManager` to make any difference, right? Which is too large of a scope.

peppy: once a marker interface is present, it could be blocked at pass through managers too via this flag, right?

smoogipoo: Yeah that's true.

PR #3665, opened 2020-06-28
Expose application of adjustements to an AdjustableAudioComponent to user code

Game4all:
# Summary

This aims to set the access modifier of the AdjustableAudioComponent.OnStateChanged() method to protected to allow using the same codepath as the framework when applying adjustments on a custom AdjustableAudioComponent in user code (eg: custom BASS stream).

As such, the method has been aswell renamed to `ApplyAdjustements` to show that it is called to apply adjustments on the audio component from the aggregate bindable values.


bdach:
What is the end goal of this, i.e. where and how would `ApplyAdjustments` be used?

Game4all:
Basically, this would get called internally to apply adjustements to the audio component when one of its bindable adjustments are modified (see: https://github.com/Game4all/osu-framework/blob/1a3395dc78514037f423a59357b740301ddba3c7/osu.Framework/Audio/AdjustableAudioComponent.cs#L37 as well as https://github.com/Game4all/osu-framework/blob/1a3395dc78514037f423a59357b740301ddba3c7/osu.Framework/Audio/Track/TrackBass.cs#L290 for an example of what would be done in it). 

Goal is really to be able to reuse the code path used to apply adjustments in the framework with custom AdjustableAudioComponents in usercode (eg: a custom `Track` class using an underlying BASS custom stream)

smoogipoo:
The part that concerns me is, and probably why a real-world example was requested is: don't you want the stream to be passed in at the very least too?
Something like a `int CreateStream(int stream)` method, which takes in a stream, applies adjustments, potentially creates new channels (e.g. bassfx reverse channels, I'm not 100% knowledgeable about all the capabilities of bass), and then returns the final channel.

peppy:
Also would you not need to somehow inject a new adjustment type into `AdjustableProperty` for this to be useful?

Game4all:
I feel like my point hasn't been understood so I'll attach a code example. Let's consider the very basic case here: 
```csharp
public class TrackBassStream : Track
{
    private int streamHandle;

    private readonly StreamProcedure proc;

    public override void Start() => Bass.ChannelPlay(streamHandle);

    public override void Stop() => Bass.ChannelStop(streamHandle);

    public TrackBassStream(StreamProcedure proc) 
    {
        streamHandle = Bass.CreateStream(44100, 2, BassFlags.Default | BassFlags.Float, (this.proc = proc));

        AggregateVolume.BindValueChanged(v => updateAdjustments());
        AggregateFrequency.BindValueChanged(f => updateAdjustments());
        //and so on for every audio adjustment that the framework supports
    }

    private void updateAdjustments() 
    {
        Bass.ChannelAttribute(streamHandle, ChannelAttribute.Frequency, AggregateFrequency.Value);
        Bass.ChannelSetAttribute(streamHandle, ChannelAttribute.Volume, AggregateVolume.Value);
    }
}
```
Adding support for frameworks' supported adjustments requires binding the adjustment bindable to a method containing code to apply them to the BASS channel. My point is that exposing `ApplyAdjustments` woud allow adding support for a framework's supported adjustment just by adding its corresponding `BASS.ChannelAttribute` call to the method, as all adjustment bindables are bound by default to `ApplyAdjustments()` in `AdjustableAudioComponent`.

smoogipoo:
I'm still unsure... It seems like you're trying to create custom track/sample component that adds functionality that `TrackBass` doesn't currently have? I'm not sure whether that's a good thing in principle - handling low-level components is one of the core features of o!f, such that you should never really be doing `TrackBassStream : Track` but rather `TrackBassStream : TrackBass`.

Maybe even things like DSP effects are things that I could see happening in o!f in one way or another.

One exception is using a custom audio framework altogether, at which point I could probably foresee this being useful - is that the case here?

Game4all:
Main reason I didn't derive `TrackBass` was because it used a file procedure whereas I needed to use a bass stream with a 
stream procedure to buffer PCM data generated in realtime.

Actually, I now guess (after reading your comment above) that making an abstract class `TrackBass` housing all calls to BASS for adjustments as well as for the track logic and adding a `int CreateStream()` method responsible of creating the underlying BASS stream may work aswell. 

It would just require moving current stream logic of `TrackBass` to another class.

peppy:
Sounds like you could potentially add a new protected constructor to TrackBass which takes in the bass stream location, then route the existing `Stream`-based `ctor` through that.

peppy:
@Game4all can you give https://github.com/ppy/osu-framework/pull/3696 a try and see if it works for your intended usage? Note that there may need to be further changes to allow more custom implementation (and potentially handling streams with no set length?)

Game4all:
I got to toy with a local checkout of your branch and got my custom BASS stream to work through `TrackBass`, altough I had to comment out this line: https://github.com/ppy/osu-framework/blob/4f362ceb11a3d316cfa2daba89476edd67e7e537/osu.Framework/Audio/Track/TrackBass.cs#L125

But I guess after looking at the returned error codes that this is very likely an issue on my end.

peppy:
I'm not sure it's possible to reverse a live stream. If you can share what errors you are seeing (or the code you're using) I can look to either catch the error or conditional the application of effect streams and get that approach merged.

Game4all:
I will post the code snippet aswell as the BASS errors encountered in the aforementionned PR and will continue potential further discussion there. I'm closing this as it isn't going in the envisioned direction and is being superseded by the aforementionned PR

PR #3666, opened 2020-06-29
Add a copy constructor to InputState

smoogipoo:
Noticed while updating osu!-side dependencies that:
1. `touch` was injected before `joystick` - build-time error.
2. `midi` wasn't passed through.

It's not a huge deal, but it's probably safer for the user to be able to specify the `InputState` to be copied from.

### DIFF
--- a/osu.Framework/Input/States/InputState.cs
+++ b/osu.Framework/Input/States/InputState.cs
@@ -3,14 +3,53 @@
 
 namespace osu.Framework.Input.States
 {
+    /// <summary>
+    /// An object that stores all input states.
+    /// </summary>
     public class InputState
     {
+        /// <summary>
+        /// The mouse state.
+        /// </summary>
         public readonly MouseState Mouse;
+
+        /// <summary>
+        /// The keyboard state.
+        /// </summary>
         public readonly KeyboardState Keyboard;
+
+        /// <summary>
+        /// The touch state.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


osu.Framework/Input/States/InputState.cs
```diff
@@ -3,14 +3,53 @@
...
+        /// The midi state.
+        /// </summary>
         public readonly MidiState Midi;
 
+        /// <summary>
+        /// Creates a new <see cref="InputState"/> using the individual input states from another <see cref="InputState"/>.
+        /// </summary>
+        /// <param name="other">The <see cref="InputState"/> to take the individual input states from.</param>
+        public InputState(InputState other)
+            : this(other.Mouse, other.Keyboard, other.Touch, other.Joystick, other.Midi)
```
frenzibyte, line 41: This would references those input states from `other` as they all are reference types (classes), is that intended?

smoogipoo: Yeah it is intended, since the use case is something like:

https://github.com/ppy/osu/blob/db1d02cb0f16abca8f516f4c450d667b28b226f3/osu.Game/Rulesets/UI/RulesetInputManager.cs#L45-L49

peppy: i'd say it's worth explicitly mentioning this in the xmldoc for clarity.

"Note that input states are referenced, not copied."

PR #3669, opened 2020-06-30
Schedule lifetime change callbacks after children

bdach:
Resolves ppy/osu#8145 (*finally*).

# Summary

As outlined in the issue thread, a storyboard usage of `LifetimeManagementContainer` revealed that setting `LifetimeStart` and `LifetimeEnd` inside a child's `Update()` was unsafe, as it was mutating `AliveInternalChildren` while they were being enumerated inside of `UpdateSubTree()` through the `childLifetimeChanged` callback.

Resolve by deferring all `childLifetimeChanged` executions until all alive children are processed during an update frame.

Here are two videos, demonstrating the manifestation of the bug on the map affected game-side and its resolution with this PR:

* [before](https://streamable.com/xxk4cq) - note the zombie samples highlighted in the draw visualiser and the lack of kick drum playback when they appear,
* [after](https://streamable.com/wa3qc0) - note the presence of the kick drums and no left-over alive children in the storyboard layer.

# Remarks

Mini-rant: in retrospect the signs this was going to be this sort of issue were there all along, but for some reason I could not understand the presentation of this bug for the longest time (I stared at it on and off for like two weeks). I was thinking this was a BASS sample concurrency issue several times and almost gave up on debugging this like three times along the way. Not sure if this was just me being bird-brained or just a nasty bug.

(That piano intro song is burned in my brain now from how many times I replayed it.)

Glad I caught it, though.

### DIFF
--- a/osu.Framework.Tests/Visual/Containers/TestSceneLifetimeManagementContainer.cs
+++ b/osu.Framework.Tests/Visual/Containers/TestSceneLifetimeManagementContainer.cs
@@ -193,6 +193,26 @@ namespace osu.Framework.Tests.Visual.Containers
             validate(1);
         }
 
+        [Test]
+        public void TestLifetimeMutatingChildren()
+        {
+            AddStep("detach container", () => Remove(container));
+
+            TestLifetimeMutatingChild first = null, second = null;
+            AddStep("add children", () =>
+            {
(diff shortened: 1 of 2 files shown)


peppy:
Nice job. Always good when the resulting fix makes the original problem feel obvious.

PR #3670, opened 2020-07-01
Make OnTouchMove() return void instead of bool

frenzibyte:
Should've been like this since we're propagating the move event to the `ButtonDownInputQueue`, same as how in dragging `DragEvent` is not handleable/blockable.

Also slightly fixes incorrect provided XMLDoc.


osu.Framework/Graphics/Drawable.cs
```diff
@@ -2222,17 +2223,16 @@ public bool TriggerEvent(UIEvent e)
         /// An event that occurs every time an active <see cref="Touch"/> has moved while hovering this <see cref="Drawable"/>.
         /// </summary>
         /// <remarks>
-        /// This will only be invoked on the <see cref="Drawable"/> that returned <code>true</code> from a previous <see cref="OnTouchDown"/> invocation.
+        /// This is guaranteed to be invoked if <see cref="OnTouchDown"/> was invoked.
```
peppy, line 2226: Does this imply it will *not* be invoked if `OnTouchDown` was not handled?

frenzibyte: The new XMLDoc matches up with the rest of the event handlers that have the same behaviour as this one (check `OnDrag`, and `OnKeyUp` as an example), I believe there is a task added to the documentation issue thread tracking the clarity of the XMLDocs attached here, so I believe to at least fix this rather than keeping it even wronger (yes if you read the old one, it's clearly incorrect).

peppy: Matching incorrect documentation should not be something we do. Let's make it understandable here as a starting point.

smoogipoo: Rather than incorrect documentation, it's documentation that could be worded better but is currently consistent.
As it is I'm fine with this documentation because it is consistent, so if you're suggesting something like that I'd really hope you provide an example of how the two documentations should read. I've said it before and I'll say it here again, the two types of documentation are of the form:
1. "guaranteed to be invoked if X was invoked"
2. "only invoked on the drawable that returned true from a previous X"

smoogipoo: With that being said, is this really "guaranteed"? If a user touches but doesn't move, will an `OnTouchMove` event be propagated? If not this should be worded more like "can only be invoked ...", similar to `OnClick()`.

peppy: Exactly, which is *almost* what the old documentation already said, right?

smoogipoo: Almost except for the returned true part, sure.

frenzibyte: Yeah just like any up events, they all have the "guaranteed" documentation while they're not really guaranteed.

smoogipoo: Do you have an example where up events are not guaranteed? That sounds like a huge bug if so.

frenzibyte: Oh, I actually misunderstood what you meant above, nevermind.

frenzibyte: @peppy I didn't get that you were referring to that case in particular, I've pushed an improvement, please check if it's correct now.

Issue #3667, opened 2020-06-29
Primary touch and mouse down events can trigger on different drawables


smoogipoo:
The primary touch is always passed through as mouse down and doesn't respect the blocking behaviour of the hierarchy. This can cause the touch down and mouse down input to be received by different Drawables in the same frame.

```csharp
public class TestSceneTesting : ManualInputManagerTestScene
{
    [Test]
    public void TestPrimaryTouchNotForwardedAsMouseIfHandledAsTouch()
    {
        MouseTarget mouseTarget = null;
        TouchTarget touchTarget = null;

        AddStep("create hierarchy", () =>
        {
            Children = new Drawable[]
            {
                mouseTarget = new MouseTarget { RelativeSizeAxes = Axes.Both },
                touchTarget = new TouchTarget { RelativeSizeAxes = Axes.Both }
            };
        });

        AddStep("touch down", () => InputManager.BeginTouch(new Touch(TouchSource.Touch1, mouseTarget.ScreenSpaceDrawQuad.Centre)));
        AddStep("touch up", () => InputManager.EndTouch(new Touch(TouchSource.Touch1, mouseTarget.ScreenSpaceDrawQuad.Centre)));

        AddAssert("touch target received input", () => touchTarget.ReceivedOnTouchDown);
        AddAssert("mouse target not received input", () => !mouseTarget.ReceivedOnMouseDown);
    }

    private class MouseTarget : Drawable
    {
        public bool ReceivedOnMouseDown;

        protected override bool OnMouseDown(MouseDownEvent e)
        {
            ReceivedOnMouseDown = true;
            return true;
        }
    }

    private class TouchTarget : Drawable
    {
        public bool ReceivedOnTouchDown;

        protected override bool OnTouchDown(TouchDownEvent e)
        {
            ReceivedOnTouchDown = true;
            return true;
        }
    }
}
```

peppy:
Just to better understand: what's a case where this would lead to bad/unwanted behaviour? I'm not actually sure we need to worry about this. Basically any case where an application is handling touches, it would outright want to disable the mouse cursor propagation part, right? Otherwise you'd be handling two events (at the same drawable or otherwise) and things would get weird.

smoogipoo:
[...]

PR #3675, opened 2020-07-01
Distinguish mouse events from touch input via bool property

frenzibyte:
Addresses a discussion in #3667, around combining touch and mouse input handling like cursor containers, magically required for dependent PR which made me work on it first.

The way implemented here is likely the best way to get proper values on this, there were other ways like directly passing the `IInput` to all button event managers and reading the value from there, or passing the bool to the `InputStateChangeEvent` and pass that class into the button event managers. But I'm confident with the way I wrote it now, consistent with the rest of current mouse state properties in the event, easy to pass, and simplifies the changes.

EDIT: Forgot to note, but I've decided to put a bool property instead of marker interface as that one would really pollute the `MouseButtonEventManager` class with checks through whether `XXXFromTouchEvent` should be propagated or the original one.

### DIFF
--- a/osu.Framework/Input/StateChanges/IMouseInput.cs
+++ b/osu.Framework/Input/StateChanges/IMouseInput.cs
@@ -1,0 +1,13 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework.Input.StateChanges
+{
+    public interface IMouseInput : IInput
+    {
+        /// <summary>
+        /// Whether this input is performed from a primary touch input source.
+        /// </summary>
+        public bool FromTouchSource { get; set; }
+    }
+}
--- a/osu.Framework/Input/StateChanges/IMouseInput.cs
+++ b/osu.Framework/Input/StateChanges/IMouseInput.cs
@@ -8,6 +8,6 @@ namespace osu.Framework.Input.StateChanges
         /// <summary>
         /// Whether this input is performed from a primary touch input source.
         /// </summary>
-        public bool FromTouchSource { get; set; }
+        bool FromTouchSource { get; set; }
     }
 }
--- a/osu.Framework/Input/StateChanges/MousePositionAbsoluteInput.cs
+++ b/osu.Framework/Input/StateChanges/MousePositionAbsoluteInput.cs
@@ -14,8 +14,10 @@ namespace osu.Framework.Input.StateChanges
     /// <remarks>
     /// This is the first input received from any pointing device.
     /// </remarks>
-    public class MousePositionAbsoluteInput : IInput
+    public class MousePositionAbsoluteInput : IMouseInput
     {
+        public bool FromTouchSource { get; set; }
+
         /// <summary>
         /// The position which will be assigned to the current position.
         /// </summary>
--- a/osu.Framework/Input/StateChanges/MousePositionAbsoluteInput.cs
+++ b/osu.Framework/Input/StateChanges/MousePositionAbsoluteInput.cs
@@ -29,6 +29,8 @@ namespace osu.Framework.Input.StateChanges
 
             if (!mouse.IsPositionValid || mouse.Position != Position)
             {
+                mouse.FromTouchSource = FromTouchSource;
+
                 var lastPosition = mouse.IsPositionValid ? mouse.Position : Position;
                 mouse.IsPositionValid = true;
                 mouse.Position = Position;
--- a/osu.Framework.Tests/Visual/Input/TestSceneTouchInput.cs
+++ b/osu.Framework.Tests/Visual/Input/TestSceneTouchInput.cs
@@ -289,12 +289,37 @@ namespace osu.Framework.Tests.Visual.Input
             });
         }
 
+        [Test]
+        public void TestMouseEventFromTouchIndication()
+        {
+            InputReceptor primaryReceptor = null;
+
+            AddStep("retrieve primary receptor", () => primaryReceptor = receptors[(int)TouchSource.Touch1]);
+            AddStep("setup handlers to avoid mouse-from-touch events", () =>
+            {
+                primaryReceptor.HandleMouse = me => !me.FromTouchSource;
+            });
+
+            AddStep("perform input on primary touch", () =>
+            {
+                InputManager.BeginTouch(new Touch(TouchSource.Touch1, getTouchDownPos(TouchSource.Touch1)));
+                InputManager.MoveTouchTo(new Touch(TouchSource.Touch1, getTouchMovePos(TouchSource.Touch1)));
+                InputManager.EndTouch(new Touch(TouchSource.Touch1, getTouchUpPos(TouchSource.Touch1)));
+            });
+            AddAssert("no mouse event received", () => primaryReceptor.MouseEvents.Count == 0);
+
+            AddStep("perform mouse move input", () => InputManager.MoveMouseTo(getTouchDownPos(TouchSource.Touch1)));
+            AddAssert("mouse event received", () => primaryReceptor.MouseEvents.Single() is MouseMoveEvent);
+        }
+
         private class InputReceptor : CompositeDrawable
         {
             public readonly TouchSource AssociatedSource;
 
             public readonly Queue<TouchEvent> TouchEvents = new Queue<TouchEvent>();
             public readonly Queue<MouseEvent> MouseEvents = new Queue<MouseEvent>();
+
+            public Func<MouseEvent, bool> HandleMouse;
 
             public InputReceptor(TouchSource source)
             {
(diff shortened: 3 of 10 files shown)


osu.Framework/Input/StateChanges/IMouseInput.cs
```diff
@@ -0,0 +1,13 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework.Input.StateChanges
+{
+    public interface IMouseInput : IInput
+    {
+        /// <summary>
+        /// Whether this input is performed from a primary touch input source.
+        /// </summary>
+        public bool FromTouchSource { get; set; }
```
bdach, line 11: I catch myself doing this often since C# 8 landed too, but the `public` modifier on the interface member is implied:

```suggestion
        bool FromTouchSource { get; set; }
```

osu.Framework/Input/StateChanges/MousePositionAbsoluteInput.cs
```diff
@@ -27,6 +29,8 @@ public void Apply(InputState state, IInputStateChangeHandler handler)
 
             if (!mouse.IsPositionValid || mouse.Position != Position)
             {
+                mouse.FromTouchSource = FromTouchSource;
```
bdach, line 32: Honestly not a huge fan of all this copying of this value in multiple places. Can this not be done in a more centralised way (via an abstract class or otherwise) or using a different approach (like the marker interface suggested in the discussion)?

frenzibyte: Yeah tbh, I hate it as well, but really can't do anything.
 - Can't do it in an abstract class as the current input classes abstractions are not compatible with doing something like this (`MouseButtonInput` inherits `ButtonsInput`)
 - Using a marker interface would really pollute the `MouseButtonEventManager` and not be of any use.

Would be nice if there was a way to do some-type of mixin or something, but can't find anything.

osu.Framework/Input/StateChanges/IMouseInput.cs
frenzibyte: Wait..., when did I put this o_O

peppy: This should not be in the mouse state, as discussed elsewhere.

frenzibyte: It is required to be in the mouse state, such that `IRequireHighFrequencyMousePosition` would work correctly and also useful for consumers to know whether the current mouse state is as-is because of touch input.

peppy: Sounds like we should be fixing the clone logic rather than hacking around it. this would imply that *any* deriving classes would be lost in the flow, so by not fixing but working around the issue, it is being left open for future failure (the classes should be sealed if you go down the path you are, but I don't believe this is what we want).

frenzibyte: I'm not sure what you mean by "the clone logic", and "this would imply that any deriving classes would be lost in the flow", can you elaborate further?

peppy: `MouseMoveEvent` is not sealed. It could be derived and cause the same loss-of-information you are implying above.

The change in this PR is hacking around an issue without resolving the underlying issue.

Fixing the clone logic implies fixing the case where this loss-of-information occurs, which you suggest is in the IRequireHighFrequency code path. Construction of the base class should be replaced with clone logic that respects the original type and content.

smoogipoo: I'd also like to see this implemented as a property rather than an interface. It may be slightly harder to implement but I think it's worthwhile... If we want to add additional properties (such as perhaps the touch source, if we maybe allow multiple touches to do this in the future, etc), I wouldn't want to pollute MouseEvent with that sort of information.

If the only issue is the high-frequency path, I think may be fine to store the property in the state (considering syncing...), and then switch on that state for the event to propagate in the hf path.

peppy: Did you mean a property in an interface?

smoogipoo: No. I meant keep the property inside `MouseState` as it currently is, and have the interface for UI events.

peppy: To clarify, you're saying you believe it *should* be in `MouseState` as per this PR?

smoogipoo: I think it makes sense from a syncing perspective, yes.

peppy: Are you okay with it being `private` or `internal`, in that case?

peppy: Also can you explain why you're against keeping it isolated to an interface? This would still allow further properties to be added in the future for the use cases you mention.

To build on my last comment, the user should not be aware of this bool unless they are very specifically wanting to use it (in an already-touch-supporting application). By exposing it in a mouse event, it's not only polluting cross-polluting separate input systems, but I think it's quite misleading – and badly guiding – consumers.

frenzibyte: Oh so this would mean even if it's gonna pollute the framework implementation internally, it's fine as long as it's not polluted to what the user sees, didn't really think you would be fine but talking about it from this perspective, that makes absolute sense, will change it.

smoogipoo: > No. I meant keep the property inside MouseState as it currently is, and have the interface for UI events.

I already mentioned having an interface for UI events.

As for `MouseState`, another alternative may be to not store the bool inside `MouseState` but inside `InputState` itself, as a "LastMovementSource" enum value (`Touch`/`Mouse`).

peppy: Please wait until the discussion has concluded, as a direction forward has not been attained yet.

peppy: @smoogipoo my criteria for implementation:

- should not be visible to the user via a `Mouse(Anything)` class to consumers
- should be accessible by the user via an interface (for special cases where the user is already supporting touch)

can you agree with that direction?

smoogipoo: I agree with that in the context of UI events (`Mouse(Anything)`). Not sure about `InputState`, I think it'd be best to expose it in some way - there are still cases where we handle input manually via `InputManager.CurrentState`, and some of those cases _may_ want to know which device the current position comes from.

E.g. if a user wants to handle _mouse_ input only, but accesses `CurrentState.Mouse.Position`, how can they be sure that that position comes from a mouse and not a touch?

peppy: "handle mouse only" should only be a scenario if the user is also handling touch, at which point they should be using the interface (it would be documented under the touch wiki page or otherwise).

there should be no scenario a consumer supports mouse and only mouse, disabling touch, unless they are aware of touch already.

smoogipoo: And what if you can't use the interface? For whatever reason - maybe you haven't even received a movement event because your drawable's too new to the hierarchy, or hasn't been moused over.

Example: A cursor that doesn't reside in a 100% screen area, which is shown only during mouse input, that wants to show the current mouse position when it's out of bounds, or an accurate position when no movement event has otherwise been received.

peppy: Ah, maybe I misunderstood – the goal here was not just to interface the events, but also the state, and have a `TouchBasedMouseState : MouseState`, along with updating the clone/transfer logic to allow for this.

Therefore the following would be possible:

`if (GetContainingInputManager().CurrentState.MouseState is ITouchSourcedMouseState touchSpecificData) { ... }`

smoogipoo: I see. That sounds difficult to handle, since a lot of that part of code is exposed to the user (see `InputManager.CreateInitialState()`). Beyond the difficulty of transferring states, just deciding when it's actually safe to switch the states around sounds non-trivial. If the user couldn't specify their own mouse state at construction time, then it'd be pretty simple.

I don't know how I feel about the entire `MouseState` member changing, conceptually, and potentially flip-flopping between two types for what is essentially almost a singleton. For what it's worth, I'm also not a fan of my idea of `InputState` having a mutable property either...

peppy: What do you think about sealing all state classes (see original comment https://github.com/ppy/osu-framework/pull/3675#discussion_r448670134) then working out a solution with that restriction in mind? As I mentioned, it's open to future error if we don't fix this here.

smoogipoo: I agree with sealing.

Am also thinking if there's any chance that maybe we _don't_ mutate the `MouseState` on a touch-to-mouse event? It would give the user the guarantee that `MouseState` always corresponds to a mouse device, and `TouchState` always corresponds to a touch device, but there'd need to be more extra considerations for `MouseEvent`, like perhaps instead of `IsPressed()` checking `CurrentState.Mouse` it'd check the union of Mouse+Touch (or even another state like an `internal` `InputState.TouchSourcedMouseState`, `internal` is fine in this case because a user can always reference `Mouse` and `Touch` independently to retrieve true device states).

frenzibyte: Sounds like something that could work, but would require a good amount of time and effort. going the interface way sounds easier for now to fix the issues this PR is for.

frenzibyte: @peppy About adding interfaces for mouse events to indicate they are from a touch source, there does not seem to be a way to properly organize the logic of whether to construct `TouchBasedMouseXXXEvent` or the regular `MouseXXXEvent`. 

The best I can do is adding a `static MouseXXXEvent Create(InputState, ...)` method to each constructible `MouseEvent`, does that seem fine or do you have a better plan?

peppy:
Closing (#3729 replaced this).

PR #3676, opened 2020-07-01
Perform mouse-from-touch input only when no drawable handled the touch

### LINKED ISSUE #3667: Primary touch and mouse down events can trigger on different drawables

frenzibyte:
- [x] Depends on #3675 (for fixing pass-through input manager)
- [x] Depends on #3670 (for addition of `HeldDrawable` to make sense)
- [x] Depends on #3755 

Refrain from producing mouse-from-touch input if there was actually a drawable handling the touch, this will resolve the issue where mouse events can potentially be handled by another drawable if the original touch event was handled. Sounds more logical not to produce mouse input as it is useless for the entire hierarchy anyways.

Now each input manager would start producing their own mouse-from-touch input instead of propagating from parent, this would allow each to propagate their touch events, and determine properly whether their hierarchy should have mouse input produced for them or not. (as the `TestScenePassThroughInputManager.TestMouseTouchProductionOnPassThrough` shows)

Closes #3667


smoogipoo:
As mentioned previously, is there a reason this was implemented as a property rather than an interface? It may be slightly harder to implement but I think it's worthwhile... If we want to add additional properties (such as perhaps the touch source, if we maybe allow multiple touches to do this in the future, etc), I wouldn't want to pollute `MouseEvent` with that sort of information.

peppy:
@frenzibyte wanna bring this up-to-date? not sure about the "passthroughinputmanager fix" you mention and if it is relevant.

frenzibyte:
Conflicts resolved, the "`PassThroughInputManager` fix" is that each pass-through input manager should now produce its own mouse-from-touch input to its input queue, this seems to be the correct way for each pass-through to propagate mouse-from-touch events (pass touch event to input queue, if no one handled touch, produce mouse-from-touch input for them); https://github.com/ppy/osu-framework/pull/3676/commits/aa8daca93b094af214c10149f2ab4fb1f9e86811 adds test coverage to that fix.

osu.Framework/Input/InputManager.cs
```diff
@@ -638,8 +638,20 @@ public virtual void HandleInputStateChange(InputStateChangeEvent inputStateChang
...
                     HandleTouchStateChange(touchChange);
-                    HandleMouseTouchStateChange(touchChange);
+
+                    touchHandled |= manager.HeldDrawable != null;
+
+                    // Produce mouse input if no drawable in the input queue has handled this touch event,
+                    // done for compatibility with components that do not handle multi-touching.
+                    if (!touchHandled)
+                        HandleMouseTouchStateChange(touchChange);
+
```
peppy, line 654: This is *really* weirdly structured. `GetButtonEventManagerFor` is called outside of and inside of `HandleTouchStateChange`, which is only ever used here.

osu.Framework/Input/PassThroughInputManager.cs
```diff
@@ -78,19 +77,15 @@ protected override List<IInput> GetPendingInputs()
             return pendingInputs;
         }
 
-        protected override bool HandleMouseTouchStateChange(TouchStateChangeEvent e)
-        {
-            // The parent manager will propagate mouse events from latest moved touch input if we are using it.
```
peppy, line 83: reasoning for removing this? every `PassThroughInputManager` is now doing the mapping logic?

osu.Framework/Input/TouchEventManager.cs
```diff
@@ -17,6 +17,12 @@ public class TouchEventManager : ButtonEventManager<TouchSource>
     {
         protected Vector2? TouchDownPosition;
 
+        /// <summary>
+        /// The drawable from the input queue that handled a <see cref="TouchEvent"/> corresponding to this touch source.
+        /// Null when no drawable has handled a touch event or the touch is not yet active / has been deactivated.
+        /// </summary>
+        public Drawable HeldDrawable { get; protected set; }
```
peppy, line 24: if we do choose to keep this (i'm hoping we don't need to) it should be private set and have "last" in its name, since it's only storing the last.

if it's guaranteed to only be one ever, that should be asserted.

osu.Framework/Input/PassThroughInputManager.cs
```diff
@@ -78,19 +77,15 @@ protected override List<IInput> GetPendingInputs()
...
-
-            return base.HandleMouseTouchStateChange(e);
-        }
-
         protected override bool Handle(UIEvent e)
         {
             if (!UseParentInput) return false;
 
+            // Don't handle mouse events sourced from touches, we may have a
+            // child drawable handling actual touches, we will produce one ourselves.
```
peppy, line 85: i don't understand this line (the "may have a child" part)

frenzibyte: I keep mixing myself between saying `children` and `input queue`, I guess saying `we may have a drawable in our input queue handling touch events` would make it understandable?

osu.Framework/Input/PassThroughInputManager.cs
frenzibyte: Yes, as the comment added in blocking mouse-from-touch events states, every `PassThroughInputManager` should do the mapping to ensure that each input queue of a `PassThroughInputManager` would receive mouse when none of them handle touch events.

osu.Framework/Input/InputManager.cs
frenzibyte: Not sure I see why this is weird, I would need to check if the touch event manager corresponding to the source of the changed touch has propagated a touch event and a drawable handled it, if the manager couldn't find a drawable handling the touch, then a mouse input would be propagated.

I'm listening to any suggestion that would make this logic more nicer as this is the best I could do.

osu.Framework/Input/TouchEventManager.cs
frenzibyte: I tried to expose a variable that would be legible to expose and would benefit in finding whether a `TouchEventManager` has propagated a down touch and a drawable has handled it. (just like `MouseButtonEventManager.ClickedDrawable` is exposed and the click event is propagated exactly the same as how touch downs are propagated)

I'm fine with it being `private set;` (although `ClickedDrawable` and `DraggedDrawable` are settable from `protected` access).

Not sure about making it having "last" as the variable is immediately set to `null` when the touch is up.

osu.Framework/Input/PassThroughInputManager.cs
peppy: we generally avoid using "we" in osu-framework xmldoc/comments as it is confusing

frenzibyte: If that's the case then sure

osu.Framework/Input/InputManager.cs
peppy: @smoogipoo can you take a look at this? not sure if there's a better way to structure this without polluting the call stack.

at very least need confirmation whether you're happy with this method of touch detection.

smoogipoo: I think it's fine, but I find the readability to be a bit lacking... Maybe it should be split into two bools instead?
```csharp
bool wasHolding = manager.HeldDrawable != null;

HandleTouchStateChange();

bool isHolding = manager.HeldDrawable != null;

if (wasHolding != isHolding)
    HandleMouseTuchStateChange();
```

peppy: That's changing the logic, is that intentional?

smoogipoo: Not intentional. I think `!wasHolding && !isHolding` is what's intended here.

PR #3679, opened 2020-07-03
Add iOS multitouch support

peppy:


osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -32,18 +36,18 @@ private void handleTouches(NSSet obj, UIEvent evt)
                 handleUITouch((UITouch)t, evt);
         }
 
+        private UITouch[] activeTouches = new UITouch[10];
```
smoogipoo, line 39:
```suggestion
        private UITouch[] activeTouches = new UITouch[max_touches];
```

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -14,6 +16,8 @@ namespace osu.Framework.iOS.Input
 {
     public class IOSTouchHandler : InputHandler
     {
+        private const int max_touches = 10; // should match TouchSource enum count.
```
smoogipoo, line 19: `TouchState.MAX_TOUCH_COUNT`

osu.Framework.iOS/Input/IOSTouchHandler.cs
```diff
@@ -69,22 +73,62 @@ private void handleUITouch(UITouch touch, UIEvent e)
...
+
+            for (int i = 0; i < activeTouches.Length; i++)
+            {
+                if (activeTouches[i] != null) continue;
+
+                activeTouches[i] = touch;
+                return (TouchSource)i;
+            }
+
+            throw new InvalidOperationException("Too many touches");
```
smoogipoo, line 116: I'm not sure about this. Is there any documentation that iOS only supports/only every will support up to 10 touches? Some rudimentary research shows people claiming 11 for iPad's, some claim up to 17...

https://developer.apple.com/forums/thread/17606 https://mattgemmell.com/ipad-multi-touch/

Maybe the touch should be discarded instead of hard throwing?

peppy: Wow, that's new. It used to be hard-capped at 10. Will discard for now.

peppy: Have tested with >10 touches. It will ignore the extra touches until one becomes available, at which point it will assign and track it from that point as a fresh touch. Does that sound OK?

smoogipoo: Sounds fine

PR #3680, opened 2020-07-03
Add option to disable TabControl switch-on-remove behaviour

smoogipoo:
In investigating https://github.com/ppy/osu/issues/9324 I initially took the path of making `Remove` return a `bool`. While that could still be done it isn't enough to resolve the issue, since the issue occurs _while_ we're in the removal process (i.e. `bindable callback -> remove -> switch -> bindable callback -> remove`), and would also require a schedule of the callback. I think that's pretty hacky.

So probably the easiest way to do this is just adding an option to disable this behaviour altogether for cases like `BreadcrumbControl` that are `TabControl`s only in appearance and don't need such functionality.

Have also used this opportunity to document the rest of the class, as one of the goals I'm hoping to achieve is documenting any classes I touch framework-side.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs
@@ -75,7 +75,7 @@ namespace osu.Framework.Graphics.UserInterface
         public bool AutoSort { set; get; }
 
         /// <summary>
-        /// The <see cref="Dropdown"/> which is displayed when tabs overflow the visible bounds of this <see cref="TabControl{T}"/>.
+        /// The <see cref="Dropdown{T}"/> which is displayed when tabs overflow the visible bounds of this <see cref="TabControl{T}"/>.
         /// </summary>
         [CanBeNull]
         protected readonly Dropdown<T> Dropdown;
@@ -268,7 +268,7 @@ namespace osu.Framework.Graphics.UserInterface
         /// Adds an arbitrary <see cref="TabItem{T}"/> to the <see cref="TabControl{T}"/>.
         /// </summary>
         /// <param name="tab">The tab to add.</param>
-        /// <param name="addToDropdown">Whether the tab should be added to the dropdown if supported by the <see cref="TabControl{T}"/> implementation.</param>
+        /// <param name="addToDropdown">Whether the tab should be added to the <see cref="Dropdown"/> if supported by the <see cref="TabControl{T}"/> implementation.</param>
         protected virtual void AddTabItem(TabItem<T> tab, bool addToDropdown = true)
         {
             tab.PinnedChanged += performTabSort;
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -71,27 +74,59 @@ public IReadOnlyList<T> Items
         /// </summary>
         public bool AutoSort { set; get; }
 
-        protected Dropdown<T> Dropdown;
+        /// <summary>
+        /// The <see cref="Dropdown"/> which is displayed when tabs overflow the visible bounds of this <see cref="TabControl{T}"/>.
```
bdach, line 78: nit: this is referencing the field and not the generic type, which feels slightly more useful from a documentation perspective

```suggestion
        /// The <see cref="Dropdown{T}"/> which is displayed when tabs overflow the visible bounds of this <see cref="TabControl{T}"/>.
```

osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -224,10 +265,10 @@ private void removeTab(T value, bool removeFromDropdown = true)
         }
 
         /// <summary>
-        /// Adds an arbitrary <see cref="TabItem{T}"/> to the control.
+        /// Adds an arbitrary <see cref="TabItem{T}"/> to the <see cref="TabControl{T}"/>.
         /// </summary>
         /// <param name="tab">The tab to add.</param>
-        /// <param name="addToDropdown">Whether the tab should be added to the Dropdown if supported by the <see cref="TabControl{T}"/> implementation.</param>
+        /// <param name="addToDropdown">Whether the tab should be added to the dropdown if supported by the <see cref="TabControl{T}"/> implementation.</param>
```
bdach, line 271: nit: maybe `cref` it?

```suggestion
        /// <param name="addToDropdown">Whether the tab should be added to the <see cref="Dropdown"/> if supported by the <see cref="TabControl{T}"/> implementation.</param>
```

osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -383,6 +421,10 @@ public void OnReleased(PlatformAction action)
         {
         }
 
+        /// <summary>
+        /// Creates the <see cref="TabFillFlowContainer"/> to contain the <see cref="TabItem{T}"/>s.
+        /// </summary>
+        /// <returns></returns>
```
bdach, line 427: nit: would just remove as it does nothing

```suggestion
```

osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -246,23 +287,16 @@ protected virtual void AddTabItem(TabItem<T> tab, bool addToDropdown = true)
...
-        /// </para>
-        /// <para>
-        /// If <paramref name="tab"/> is currently selected and is the only tab in the <see cref="TabControl{T}"/>,
-        /// then selection will be cleared.
-        /// </para>
+        /// If the current tab is removed and <see cref="SwitchTabOnRemove"/> is <c>true</c>, then selection will change to a new tab if possible or be cleared if there are no tabs remaining in the <see cref="TabControl{T}"/>.
         /// </remarks>
         /// <param name="tab">The tab to remove.</param>
-        /// <param name="removeFromDropdown">Whether the tab should be removed from the Dropdown if supported by the <see cref="TabControl{T}"/> implementation.</param>
+        /// <param name="removeFromDropdown">Whether the tab should be removed from the dropdown if supported by the <see cref="TabControl{T}"/> implementation.</param>
```
bdach, line 293: gonna add the suggestion here too to speed things up, if you think it's a good idea

```suggestion
        /// <param name="removeFromDropdown">Whether the tab should be removed from the <see cref="Dropdown"/> if supported by the <see cref="TabControl{T}"/> implementation.</param>
```

bdach:
Oh, and what do you think about adding a test for this just for completeness's sake? It's probably a weird thing to test a little bit but it should be relatively free, and it *is* part of the exposed API, so would be good to have coverage on it.

smoogipoo:
Thanks, test added. Revealed that the property wasn't public as I intended it to be.

Issue #3519, opened 2020-05-04
Video decoding falling behind can cause large memory consumption


MiraiSubject:
**Describe the bug:** 
The tournament showcase screen I recently PR'ed https://github.com/ppy/osu/pull/8859 is leaking memory. I first encountered this issue when livestreaming a showcase myself using this screen. 

So I decided to take a look in the debugger and checking all the scenes in lazer's tournament mode. Apparently only the showcase screen showed symptoms of memory leaking. In the graph shown you can see it linearly increasing and then flattening. The flattening is me switching to the other scenes and keeping it on for a few seconds. After that I switched back to the showcase screen and the memory usage started increasing again. 

**Screenshots or videos showing encountered issue:** 
![Taskmgr_2020-05-04_15-23-13](https://user-images.githubusercontent.com/803255/80977438-04948d80-8e25-11ea-82cf-0790821ff9fa.png)

![image](https://user-images.githubusercontent.com/803255/80977279-cbf4b400-8e24-11ea-9d65-f514b1c538d6.png)

**osu!lazer version:** 
2020.429.0 and https://github.com/ppy/osu/commit/3bf58b7e3e4554e20854e380a10f1568cf8983ea

**Logs:**
[network.log](https://github.com/ppy/osu/files/4575295/network.log)
[performance.log](https://github.com/ppy/osu/files/4575297/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4575298/runtime.log)

huoyaoyuan:
I detected memory leak when trying to refactor video decoding half a year ago, with 1080p video in visual test.

PR #3684, opened 2020-07-04
Fix video decoding memory leaks

smoogipoo:
Fixes https://github.com/ppy/osu-framework/issues/3519

Unmanaged memory leaks are quite tough to debug >_<


bdach:
On a quick look this does seem to resolve the problem, but having three `av_frame_free` calls, nested deep in conditionals, is making me uneasy. I'd be much more comfortable with this if the free was called at the end of the scope in which `frame` is declared.

I've briefly tested a variant that hoists `outFrame` out to that same scope and does this:

```csharp
                                    if (frame != outFrame)
                                        ffmpeg.av_frame_free(&frame);
```

(to avoid a double-free if the path that does `outFrame = frame` executes). It seemed to not die and not leak but I haven't tested it super well.

PR #3692, opened 2020-07-06
Fix setting GridContainer content by index

### LINKED ISSUE #3657: GridContainer doesn't support setting content indices
`gridContainer.Content[r][c] = ...` won't refresh the content of the grid.

UselessToucan:
resolves #3657 

The problem is that an array does not notify about changing its elements.
This PR adds a `GridContainerContent` class which wraps a jagged array and provides notifications for its elements changes.


smoogipoo:
Awesome, but this is a pretty large breaking change. Is it possible to have an implicit operator conversion from `Drawable[][]` to `GridContainerContent`?

osu.Framework/Graphics/Containers/GridContainerContent.cs
```diff
@@ -0,0 +1,77 @@
...
+                }
+            }
+        }
+
+        private void onArrayElementChanged()
+        {
+            ContentChanged?.Invoke();
+        }
+
+        public ArrayWrapper<ArrayWrapper<Drawable>> _ { get; }
```
smoogipoo, line 48: This is not a convention we have used anywhere else, please don't use `_` as a non-discard variable name.

UselessToucan:
> Is it possible to have an implicit operator conversion from `Drawable[][]` to `GridContainerContent`?

AFAIK it is not possible to convert an array to a custom object.  However, I can write an extension method in order to make code a little cleaner.

> Vice-versa would also be good to have (`GridContainerContent` -> `Drawable[][]`, even if it requires allocs).

We can expose the original array via method or property. See https://github.com/UselessToucan/osu-framework/commit/f5b3f9205a8beaa847e392563fc4342703eb2e3d.

bdach:
> AFAIK it is not possible to convert an array to a custom object. However, I can write an extension method in order to make code a little cleaner.

I think the suggestion was to add operators like

```csharp
public static implicit operator Drawable[][](GridContainerContent content);
public static implicit operator GridContainerContent(Drawable[][] drawables);
```

and it seems to at least not throw syntax errors and build? This way the changes to `Content` should no longer be breaking, or am I missing something?

osu.Framework/Graphics/Containers/GridContainerContent.cs
```diff
@@ -0,0 +1,79 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+
+namespace osu.Framework.Graphics.Containers
+{
+    /// <summary>
+    /// Implements a jagged array behavior with element change notifications
+    /// </summary>
+    public class GridContainerContent
```
frenzibyte, line 11: This needs member ordering, I get lost reading this, move fields, properties and indexer above the ctor

osu.Framework/Graphics/Containers/GridContainerContent.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        private readonly Drawable[][] source;
+
+        private void onArrayElementChanged()
+        {
+            ContentChanged?.Invoke();
+        }
+
+        public ArrayWrapper<ArrayWrapper<Drawable>> WrappedArray { get; }
```
frenzibyte, line 46: Why is this `publicly` exposed, the whole array wrapper class should be `private`, only collection interfaces should be exposed to lead to the `ArrayWrapper`'s indexer.

osu.Framework/Graphics/Containers/GridContainerContent.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            public T[] WrappedArray { get; set; }
+
+            public T this[int index]
+            {
+                get => WrappedArray[index];
+                set
+                {
+                    WrappedArray[index] = value;
+                    ArrayElementChanged?.Invoke();
```
frenzibyte, line 74: No equality check

osu.Framework/Graphics/Containers/GridContainerContent.cs
frenzibyte, line 11: Should probably inherit some collection interfaces for LINQ extensions

frenzibyte, line 11: Useful internally, only.
```suggestion
    internal class GridContainerContent
```

UselessToucan: This means that `GridContainer.Content` have to be internal which is not possible since it is used in some game classes.
image

smoogipoo: `public` class with a `private` or `internal` ctor works (i.e. always use the implicit operator).

bdach:
Unfortunately this is still a breaking change, albeit a slightly less breaking one - game-side there's [this usage](https://github.com/ppy/osu/blob/c21b1e358c55a03f74e7e02a74042ec7a25e2712/osu.Game.Rulesets.Osu/Statistics/AccuracyHeatmap.cs#L195) that needs adjusting for.

osu.Framework/Graphics/Containers/GridContainer.cs
```diff
@@ -131,8 +141,8 @@ private void layoutContent()
             if (cellContent.IsValid)
                 return;
 
-            int requiredRows = Content?.Length ?? 0;
-            int requiredColumns = requiredRows == 0 ? 0 : Content.Max(c => c?.Length ?? 0);
+            int requiredRows = Content?.Count() ?? 0;
+            int requiredColumns = requiredRows == 0 ? 0 : Content?.Max(c => c?.WrappedArray.Length ?? 0) ?? 0;
```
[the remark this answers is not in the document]
UselessToucan: Here is the `Count` method implementation:
https://github.com/dotnet/runtime/blob/master/src/libraries/System.Linq/src/System/Linq/Count.cs#L12

As you can see on the [line 29](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Linq/src/System/Linq/Count.cs#L29), it is using `ICollection.Count` property which [is implemented](https://github.com/dotnet/runtime/blob/master/src/coreclr/src/System.Private.CoreLib/src/System/Array.CoreCLR.cs#L16) in arrays.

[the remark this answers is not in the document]
bdach: Yes, that's what I meant by proposing to *not* change this to `Count()`.

osu.Framework/Graphics/Containers/GridContainerContent.cs
```diff
@@ -0,0 +1,104 @@
...
+        private void onArrayElementChanged()
+        {
+            ContentChanged?.Invoke();
+        }
+
+        /// <summary>
+        /// Wraps an array and provides a custom indexer with element change notification
+        /// </summary>
+        /// <typeparam name="T">An array data type</typeparam>
+        public class ArrayWrapper<T> : IEnumerable<T>
```
frenzibyte, line 64: I'm also talking about this to not be exposed, implement `IReadOnlyList<T>` and map all properties directly to the `WrappedArray`, now you don't need to expose this *internal* class.

osu.Framework/Graphics/Containers/GridContainerContent.cs
```diff
@@ -0,0 +1,104 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace osu.Framework.Graphics.Containers
+{
+    /// <summary>
+    /// Implements a jagged array behavior with element change notifications
+    /// </summary>
+    public class GridContainerContent : IEnumerable<GridContainerContent.ArrayWrapper<Drawable>>
```
frenzibyte, line 13: This should also use `IReadOnlyList<T>` instead to get to a very similar implementation to the `Content`'s original type.

osu.Framework/Graphics/Containers/GridContainer.cs
```diff
@@ -37,20 +37,30 @@ private void load()
         /// Null elements are allowed to represent blank rows/cells.
         /// </para>
         /// </summary>
-        public Drawable[][] Content
+        public GridContainerContent Content
         {
             get => content;
             set
             {
                 if (content == value)
```
frenzibyte, line 45: This is not correct now, one array can instantiate more than one `GridContainerContent`, use `IEquatable<T>` instead.

osu.Framework/Graphics/Containers/GridContainerContent.cs
frenzibyte: In response to the confusion reaction above (guess I suck at being concise), `ArrayWrapper` should not be exposed to `public` as it is very internal and should not be used anywhere outside the `GridContainerContent` class.

Note: my bad on picking `IReadOnlyList<T>`, misunderstood the interface, turns out it doesn't have a setter indexer, please replace that with `IList<T>` instead.

So, I'm suggesting to implement `IList<T>` such that the indexer implemented here can be called from the interface without exposing the whole class (picked `IList<T>` as there's no direct interface for arrays):
```csharp
public class GridContainerContent : IList< IList<Drawable>>
{
    private readonly ArrayWrapper<ArrayWrapper<string>> sourceWrapper;

    public IList<string> this[int index]
    {
        get => sourceWrapper[index];
        set => sourceWrapper[index] = new ArrayWrapper(list); // maybe change the wrapped list instead, saving immediate construction (e.g. sourceWrapper[index].ChangeWrappedArray(list))
    }

    private class ArrayWrapper<T> : IList<T>
    {
        public T this[int index]
        {
            get => wrappedList[index];
            set
            {
                  // check equality.

                  wrappedList[index] = value;
                  // Invoke event
             }
        }
    }
}

File: osu.Framework/Graphics/Containers/GridContainerContent.cs [hunk 9]
```diff
@@ -0,0 +1,229 @@
...
+
+            public ArrayWrapper(T[] arrayToWrap)
+            {
+                wrappedArray = arrayToWrap;
+            }
+
+            public int IndexOf(T item)
+            {
+                throw new NotSupportedException();
+            }
```
bdach, line 101: while I can understand mutating operations from `IList<T>` throwing exceptions for simplicity, however inelegant that might be, I don't understand why these non-mutating methods also throw while they could delegate to the inner array instead.

on another note, you could also probably implement all of these if you leveraged `BindableList`'s `CollectionChanged` events, although that would probably be fairly complex and possibly not worth the effort.

osu.Framework/Graphics/Containers/Markdown/MarkdownTable.cs
```diff
@@ -93,12 +93,12 @@ private void computeColumnDefinitions()
...
+            Span<float> columnWidths = stackalloc float[tableContainer.Content[0].Count];
 
             // Compute the maximum width of each column
-            for (int r = 0; r < tableContainer.Content.Length; r++)
+            for (int r = 0; r < tableContainer.Content.Count; r++)
             {
-                for (int c = 0; c < tableContainer.Content[r].Length; c++)
+                for (int c = 0; c < tableContainer.Content[r].Count; c++)
                     columnWidths[c] = Math.Max(columnWidths[c], ((MarkdownTableCell)tableContainer.Content[r][c]).ContentWidth);
             }
```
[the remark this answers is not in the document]
smoogipoo: I think it's fine. There's only one such usage in osu!, so I don't think it's a common thing.

smoogipoo (dismissed):
@peppy would like your quick check on this

osu.Framework/Graphics/Containers/GridContainerContent.cs
```diff
@@ -0,0 +1,36 @@
...
+        private GridContainerContent([NotNull] Drawable[][] drawables)
+            : base(new ObservableArray<Drawable>[drawables.Length])
+        {
+            for (int i = 0; i < drawables.Length; i++)
+            {
+                if (drawables[i] != null)
+                {
+                    var observableArray = new ObservableArray<Drawable>(drawables[i]);
+                    this[i] = observableArray;
+                    observableArray.ArrayElementChanged += OnArrayElementChanged;
```
frenzibyte-alt, line 23: this should not be here, but inside `ObservableArray` as I have shown in my suggestion commit:
image
[source](https://github.com/ppy/osu-framework/commit/472f144d2d01b65688bb6fbe7f94c3ce74f6dde9#diff-a777f284af1e2a423ca3ed3a4b4590e6R32-R38)

because on the case where you do `Content[i] = new[] { ... }`, the `ArrayElementChanged` event of the `ObservableArray` created implicitly from inside will not be propagated to the main one at here, thus elements of the created array will not be observable to `GridContainer`.

should probably add test case covering this scenario to ensure it doesn't regress next time.

(also that `INotifyElementChanged` interface is meant to be `private` and only used privately, not to clash out with the existing System.Collections.Specialized.INotifyCollectionChanged`)

osu.Framework/Graphics/Containers/GridContainer.cs
```diff
@@ -37,20 +37,31 @@ private void load()
...
         /// </para>
         /// </summary>
-        public Drawable[][] Content
+        public GridContainerContent Content
         {
             get => content;
             set
             {
-                if (content == value)
+                if (content?.Equals(value) ?? false)
```
frenzibyte-alt, line 45: nit: what if both values are null? probably minor and not affecting currently, but worth fixing rather than staying silent and potentially hitting a bug at some point, and that's why I added a `==` operator in my suggestion commit:
image
[source](https://github.com/ppy/osu-framework/commit/472f144d2d01b65688bb6fbe7f94c3ce74f6dde9#diff-a777f284af1e2a423ca3ed3a4b4590e6R74-R87)

though probably worth coupling `Equals(ObservableArray<T>)` to `==`:
```csharp
public bool Equals(ObservableArray<T> other) => this == other;
```

frenzibyte-alt (changes requested):
One thing that has been missed to push from my suggestion and must be fixed before merging, and a minor suggestion.

osu.Framework/Graphics/Containers/GridContainer.cs
peppy, line 45:
```suggestion
                if (content?.Equals(value) == true)
```

reads better imho and is how we do this elsewhere.

peppy (dismissed):
i think this looks okay apart from the minor code quality issue pointed out. haven't tested thoroughly though.

osu.Framework/Lists/INotifyArrayChanged.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+
+namespace osu.Framework.Lists
+{
+    internal interface INotifyArrayChanged
```
smoogipoo, line 8: Please make this `public` and xmldoc it. You can move the xmldoc of `ArrayElementChanged` from `ObservableArray<T>`.

smoogipoo (changes requested):

smoogipoo (approved):

PR #3696, opened 2020-07-07
Allow constructing TrackBass using a raw bass stream

peppy:


Game4all:
While toying around with this I ran into some problems with the following code which is rather basic BASS code:

```cs

public class MessingWTrackBass : Component 
{

    private int bassHandle;
    private TrackBass track;

    [BackgroundDependencyLoader]
    private void load(AudioManager mng) 
    {
        bassHandle = Bass.CreateStream(SampleRate, 2, BassFlags.Decode | BassFlags.Float, fetchBassData); //creating a BASS stream with a custom stream procedure.
        mng.AddItem(track = new TrackBass(bassHandle));
        mng.Play(); //won't work as the handle returned in prepareStream(int handle) is 0x00000 (so the effect channel creation failed).
    }

    private int fetchBassData(int _handle, IntPtr Buffer, int Length, IntPtr User)
    {
        var length = Length / sizeof(float);
        var sData = new float[length];

        //Removed unrelated code.

        sData.AsSpan().Fill(0);
        Marshal.Copy(sData, 0, Buffer, length);

        return Length;
    }

}

```

Setting up breakpoints at the level of the creation of the effet channels shows that channel creation fails on this line
https://github.com/ppy/osu-framework/blob/4f362ceb11a3d316cfa2daba89476edd67e7e537/osu.Framework/Audio/Track/TrackBass.cs#L125

with error code `NotAvailable`. 

So I guess that you were right when you said that streams can't be reversed. Fortunately this can  be circumvented in my case by setting the preview param in TrackBass ctor to true (I don't require tempo adjustment capabilities). 

Not sure it is worth adding any guard for now.

peppy:
@Game4all just to confirm, you would like to see this PR merged so you can make use of it yes?

Game4all:
yep, I definitely would love to see this getting merged to make use of it with BASS streams

PR #3706, opened 2020-07-10
Add the ability to disable the ManualInputManager cursor guide

peppy:
Sometimes it can block what you're trying to test.


frenzibyte:
I think it would be better if this gets a visual checkbox instead, as sometimes it may be useful to know where the mouse is currently located in. so I wouldn't need to go right back to the test scene code, re-enable it and recompile again.

peppy:
If you're willing to implement as such go for it (shouldn't save, just be remembered between tests). It can be a follow-up to this. I still want the ability to change the default state for tests where seeing this is completely irrelevant (`OsuRuleset.GameplayCursor` test).

PR #3711, opened 2020-07-12
Add property to allow sliderbar to handle input outside of its own hover area

peppy:
Can be used to implement https://github.com/ppy/osu/issues/9479.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/SliderBar.cs
+++ b/osu.Framework/Graphics/UserInterface/SliderBar.cs
@@ -160,7 +160,11 @@ namespace osu.Framework.Graphics.UserInterface
 
         protected override bool OnKeyDown(KeyDownEvent e)
         {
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/UserInterface/SliderBar.cs
```diff
@@ -155,7 +160,7 @@ protected override void OnDragEnd(DragEndEvent e)
 
         protected override bool OnKeyDown(KeyDownEvent e)
         {
-            if (!IsHovered || currentNumberInstantaneous.Disabled)
+            if ((!AllowKeyboardInputWhenNotHovered && !IsHovered) || currentNumberInstantaneous.Disabled)
```
bdach, line 163: This seems correct but a little difficult in reading. What would your thoughts be on something like this?

```csharp
if (currentNumberInstantaneous.Disabled)
    return false;

var shouldHandle = IsHovered || AllowKeyboardInputWhenNotHovered;
if (!shouldHandle)
    return false;
```

PR #3715, opened 2020-07-13
Fix seeking animation not consuming clock

smoogipoo:
The sequence:
```
Hide()
{wait some time}
GotoFrame(0)
Show()
```
Would cause the clock to consume the `{wait some time}` period the first frame it becomes visible.

Unsure whether it's better for the consumption to be ignored during `Update()`. Thoughts? May be tricky to implement...


bdach:
I [also got bit](https://discord.com/channels/188630481301012481/589331078574112768/705521002347626529) by a similar thing once before. [It was a conscious decision from what I've been told](https://discord.com/channels/188630481301012481/589331078574112768/705554740838400013), so I didn't follow up on it and opted for a local work-around instead, but I also think the new behaviour is better.

peppy:
Regressed storyboard video. How many times are we going to regress animation clock code :(

PR #3726, opened 2020-07-15
Add function in ManualInputManager to press and release a key

peppy:
Seems like something we keep locally implemented. Unsure of method name, just trying to follow example (`Click`).

### DIFF
--- a/osu.Framework/Testing/Input/ManualInputManager.cs
+++ b/osu.Framework/Testing/Input/ManualInputManager.cs
@@ -67,6 +67,15 @@ namespace osu.Framework.Testing.Input
         public void PressKey(Key key) => Input(new KeyboardKeyInput(key, true));
         public void ReleaseKey(Key key) => Input(new KeyboardKeyInput(key, false));
 
+        /// <summary>
+        /// Press and release the specified key.
+        /// </summary>
+        public void Key(Key key)
+        {
+            PressKey(key);
+            ReleaseKey(key);
+        }
+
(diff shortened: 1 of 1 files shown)


osu.Framework/Testing/Input/ManualInputManager.cs
```diff
@@ -67,6 +67,15 @@ public void Input(IInput input)
         public void PressKey(Key key) => Input(new KeyboardKeyInput(key, true));
         public void ReleaseKey(Key key) => Input(new KeyboardKeyInput(key, false));
 
+        /// <summary>
+        /// Press and release the specified key.
+        /// </summary>
+        public void Key(Key key)
```
smoogipoo, line 73: Thinking this should be called `Press`, with overloads for `Key`, `MouseButton`, `JoystickButton`, `MidiKey`...

peppy: feels like we'd want to rename `PressKey` if we use `Press` in such a way. Maybe there's a better word than press? Since "press then release" kinda makes sense.

smoogipoo: Hmmm... Okay let's go with this for now then.

frenzibyte: Sharing my opinion here: when we talk about "pressing", we usually indicate to press the button and release immediately, unlike the word "hold" where we would mean to hold the button without releasing until we say so.

So for me, it kinda make sense if we rename this method to `Press`, and rename `PressKey` to `HoldKey`, what do you all think?

PR #3728, opened 2020-07-16
Disable dynamic compilation with debugger attached

smoogipoo:
Lags a lot and takes a long time to compile. It's probably either many exceptions being thrown internally and/or general performance issues with async that we've had for years by this point (the whole reference building thing is completely async'd). 

Perhaps in the future we can add a very slim way of doing this, maybe by bringing back the old type resolver.


peppy:
Was thinking maybe we should show a notice when this is the case, but not as straightforward as I'd hoped. Should be fine I guess (updating the wiki is a must though).

smoogipoo:
Added to the wiki.

PR #3729, opened 2020-07-16
Add marker interface and store source of last change to MouseState

peppy:
Alternative to https://github.com/ppy/osu-framework/pull/3675/files using interface rather than polluting non-touch classes with touch-specific booleans. Took same test from other PR to make it easy to compare implementation / usage.

### DIFF
--- a/osu.Framework/Input/StateChanges/ISourcedFromTouch.cs
+++ b/osu.Framework/Input/StateChanges/ISourcedFromTouch.cs
@@ -1,0 +1,12 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Input.StateChanges.Events;
+
+namespace osu.Framework.Input.StateChanges
+{
+    public interface ISourcedFromTouch
+    {
+        TouchStateChangeEvent TouchEvent { get; set; }
+    }
+}
--- a/osu.Framework/Input/StateChanges/ISourcedFromTouch.cs
+++ b/osu.Framework/Input/StateChanges/ISourcedFromTouch.cs
@@ -5,8 +5,15 @@ using osu.Framework.Input.StateChanges.Events;
 
 namespace osu.Framework.Input.StateChanges
 {
+    /// <summary>
+    /// Denotes an input which was sourced from a touch event.
+    /// Generally used to mark when an alternate input was triggered from a touch source (ie. touch being emulated as a mouse).
+    /// </summary>
     public interface ISourcedFromTouch
     {
+        /// <summary>
+        /// The source touch event.
+        /// </summary>
         TouchStateChangeEvent TouchEvent { get; set; }
     }
 }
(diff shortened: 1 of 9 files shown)


osu.Framework/Input/States/MouseState.cs
```diff
@@ -17,6 +18,11 @@ public class MouseState
 
         public bool IsPositionValid { get; set; } = true;
 
+        /// <summary>
+        /// The last input source to make a change to state.
```
frenzibyte, line 22:
```suggestion
        /// The last input source making changes to this state.
```

osu.Framework/Input/StateChanges/ISourcedFromTouch.cs
```diff
@@ -0,0 +1,12 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Input.StateChanges.Events;
+
+namespace osu.Framework.Input.StateChanges
+{
+    public interface ISourcedFromTouch
+    {
+        TouchStateChangeEvent TouchEvent { get; set; }
+    }
```
bdach, line 11: I'd say this should have xmldoc.

UselessToucan:
`ISourcedFromTouch`, `MouseButtonInputFromTouch`, `MouseButtonInputFromTouch` are probably unnecessary

see https://github.com/UselessToucan/osu-framework/commit/5c07165c246acb89abdc52372407deff3f5a69f1

bdach:
Kind of out of the loop on the touch stuff so I'd rather abstain from reviewing this if you don't mind (I guess I can also try and get caught up but unsure if I will be able to pick up all the nuances).

osu.Framework/Input/StateChanges/ISourcedFromTouch.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <summary>
+    /// Denotes an input which was sourced from a touch event.
+    /// Generally used to mark when an alternate input was triggered from a touch source (ie. touch being emulated as a mouse).
+    /// </summary>
+    public interface ISourcedFromTouch : IInput
+    {
+        /// <summary>
+        /// The source touch event.
+        /// </summary>
+        TouchStateChangeEvent TouchEvent { get; set; }
```
smoogipoo, line 17: Maybe `get;` only and add to ctor? This shouldn't be settable from the outside. This is also true for all other inputs, but this is more likely the one that's going to be used if any.

peppy:
@smoogipoo Yes I looked at the comment and disagree with it. I think you may agree if you check the required changes (it looks to add complexity and reduce readability).

smoogipoo:
I'd already looked at it by the time I posted that comment. What part increases complexity and reduces readability? I see it as reducing complexity because there's not some arbitrary marker interface somewhere that is pretty new to the structure here.

At least state change events are [documented](https://github.com/ppy/osu-framework/wiki/Diagram:-From-input-handlers-to-events), minus touch and midi.

peppy:
Reading that diagram makes the proposed change even less understandable. It's associating a `ChangeEvent` with an `IInput`, but `ChangeEvents` are the result of an `IInput`? "An event that caused this input to be created" I don't think anyone would be able to understand this.

smoogipoo:
I see, thanks.

PR #3732, opened 2020-07-18
Remove array allocations in BassAmplitudeProcessor by removing consistency guarantee

peppy:
I've left the `readonly struct` data structure in place as it seems to still work correctly.


bdach:
Thoughts about having this throw if read access is attempted on a non-audio thread?

peppy:
No, since all of our use cases do that. The point is we don't *care* about consistency (and most up-to-date data is probably what we want fo visualisation purposes for the `OsuLogo` at least)

PR #3736, opened 2020-07-18
Fix excessive list allocation in GetPendingInputs paths

peppy:
Before:

image

After:

dotMemory-201 0 20200703 51832_2020-07-18_22-09-59


bdach:
Hmm, seeing that this code path gets used osu!-side for replay playback, this seems like a pretty wide breaking change across all rulesets that provide replay playback, and as such probably warrants a breaking change note. (Not that it should require very much migration work, but it would still be more than no work.)

peppy:
Yep, will need a notice.

bdach:
While I was looking at the osu!-side update PR it kind of hit me that the method name feels a bit off and something like `CollectPendingInputs` would be slightly better IMO. If we're breaking things then we might as well go the distance and rename; thoughts?

PR #3737, opened 2020-07-18
Reduce allocations caused by DepthValue

peppy:
Before:
image

After:
dotMemory-201 0 20200703 51832_2020-07-18_22-33-29


bdach:
Looks like a valid optimisation (although the assignments at `DrawOpaqueInteriorSubTree()` almost made me think they weren't - they are, however, thanks to the implicit conversion operator).

If there was one thing I'd suggest here however, it would be to remove the initial value specification of `depth` at the field and instead declare a ctor for `DepthValue` that calls `Reset()`. Thinking it would be a slight code quality improvement due to centralising initialisation/reset-to-default logic.

PR #3739, opened 2020-07-19
Fix CursorEffectContainer allocating excessively

peppy:
Before:

image

After:

image

- [x] Test allocations are still correct with `ReadOnlyCollection` change.

### DIFF
--- a/osu.Framework/Graphics/Cursor/CursorEffectContainer.cs
+++ b/osu.Framework/Graphics/Cursor/CursorEffectContainer.cs
@@ -47,7 +47,7 @@ namespace osu.Framework.Graphics.Cursor
             // keep track of all hovered drawables below this and nested effect containers
             // so we can decide which ones are valid candidates for receiving our effect and so
             // we know when we can abort our search.
-            for (int i = targetCandidates.IndexOf(this) - 1; i >= 0; i--)
+            for (int i = selfIndex - 1; i >= 0; i--)
             {
                 var candidate = targetCandidates[i] as TTarget;
 
--- a/osu.Framework/Graphics/Cursor/CursorEffectContainer.cs
+++ b/osu.Framework/Graphics/Cursor/CursorEffectContainer.cs
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using osu.Framework.Graphics.Containers;
-using osu.Framework.Input;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Input;
 
 namespace osu.Framework.Graphics.Cursor
 {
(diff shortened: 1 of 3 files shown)


osu.Framework/Input/InputManager.cs
```diff
@@ -112,13 +112,13 @@ public Drawable DraggedDrawable
...
         /// </summary>
-        public IEnumerable<Drawable> PositionalInputQueue => buildPositionalInputQueue(CurrentState.Mouse.Position);
+        public List<Drawable> PositionalInputQueue => buildPositionalInputQueue(CurrentState.Mouse.Position);
 
         /// <summary>
         /// Contains all <see cref="Drawable"/>s in top-down order which are considered
         /// for non-positional input.
         /// </summary>
-        public IEnumerable<Drawable> NonPositionalInputQueue => buildNonPositionalInputQueue();
+        public List<Drawable> NonPositionalInputQueue => buildNonPositionalInputQueue();
```
bdach, line 121: Do we want to be this permissive with these return types? As it stands they are leaking mutable references to private lists which feels like a major violation of encapsulation. `IReadOnlyList<T>` would work, except it doesn't implement `IndexOf(T)` for some reason...

I guess they were this bad before already if someone added a cast, but that is more of a "shooting yourself in the foot willingly by violating conventions" scenario rather than "accidental foot shot because the API let you".

osu.Framework/Graphics/Cursor/CursorEffectContainer.cs
```diff
@@ -36,15 +36,27 @@ private void findTargetChildren()
...
+
+            if (selfIndex < 0) return;
 
             childDrawables.Add(this);
 
             // keep track of all hovered drawables below this and nested effect containers
             // so we can decide which ones are valid candidates for receiving our effect and so
             // we know when we can abort our search.
-            foreach (var candidate in targetCandidates)
+            for (int i = targetCandidates.IndexOf(this) - 1; i >= 0; i--)
```
bdach, line 50: Does this purposely not reuse `selfIndex`? Is there a risk of cross-thread mutation since the previous `IndexOf` call?

osu.Framework/Input/InputManager.cs
peppy: In these cases, it is building the list each call, so I do not believe it is unsafe.

osu.Framework/Graphics/Cursor/CursorEffectContainer.cs
```diff
@@ -36,15 +36,27 @@ private void findTargetChildren()
...
             {
+                var candidate = targetCandidates[i] as TTarget;
+
+                if (candidate == null)
+                    continue;
+
+                if (!candidate.IsHovered)
```
smoogipoo, line 57: I don't think you can move this here. `IsHovered` is only set if this drawable handles positional input, meaning it will stop at containers and skip their children.

osu.Framework/Graphics/Cursor/CursorEffectContainer.cs
```diff
@@ -36,15 +36,27 @@ private void findTargetChildren()
...
 
             // keep track of all hovered drawables below this and nested effect containers
             // so we can decide which ones are valid candidates for receiving our effect and so
             // we know when we can abort our search.
-            foreach (var candidate in targetCandidates)
+            for (int i = targetCandidates.IndexOf(this) - 1; i >= 0; i--)
             {
+                var candidate = targetCandidates[i] as TTarget;
+
+                if (candidate == null)
```
smoogipoo, line 54: This is also very suspicious, similar to my other comment.

osu.Framework/Graphics/Cursor/CursorEffectContainer.cs
peppy: it's still checking the same drawables as far as i can tell. the logic is a bit confusing to parse, but please read over it once more as i think this has not changed behaviour at all.

of course i may have just read the method wrong, but i did check this about 5 times over due to the convoluted nature of the code.

osu.Framework/Graphics/Cursor/CursorEffectContainer.cs
peppy: no, that's an oversight

osu.Framework/Graphics/Cursor/CursorEffectContainer.cs
smoogipoo: I think this may be fine on a secondary look...

osu.Framework/Input/InputManager.cs
peppy: As discussed in discord, this should be changed to a `ReadOnlyCollection` (`.AsReadOnly`) to avoid potential bad consumer polluting future uses of the list.

peppy:
`ReadOnlyCollection` works well from an implementation perspective. Blocking pending profiling confirmation.

peppy:
Confirming `ReadOnlyCollection` does not work.

image

Any other ideas?

bdach:
Best I can think of is some sort of custom alloc-less shim for `List` right now... There really ought to be a way to do this in the standard library. I'll poke around with a profiler today.

PR #3740, opened 2020-07-19
Fix TooltipContainer allocating excessively

peppy:
- [x] Depends on #3739

Before:

image

After:

image

### DIFF
--- a/osu.Framework/Graphics/Cursor/TooltipContainer.cs
+++ b/osu.Framework/Graphics/Cursor/TooltipContainer.cs
@@ -1,16 +1,16 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using osuTK;
-using osuTK.Graphics;
+using System;
+using System.Collections.Generic;
+using System.Linq;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
 using osu.Framework.Input;
-using System;
-using System.Collections.Generic;
-using System.Linq;
+using osuTK;
+using osuTK.Graphics;
 
 namespace osu.Framework.Graphics.Cursor
 {
@@ -201,7 +201,16 @@ namespace osu.Framework.Graphics.Cursor
             if (inputManager.DraggedDrawable is IHasCustomTooltip customDraggedTarget)
                 return hasValidTooltip(customDraggedTarget) ? customDraggedTarget : null;
 
-            ITooltipContentProvider targetCandidate = FindTargets().FirstOrDefault(hasValidTooltip);
+            ITooltipContentProvider targetCandidate = null;
+
+            foreach (var target in FindTargets())
+            {
+                if (hasValidTooltip(target))
+                {
+                    targetCandidate = target;
+                    break;
+                }
+            }
 
             // check this first - if we find no target candidate we still want to clear the recorded positions and update the lastCandidate.
             if (targetCandidate != lastCandidate)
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Cursor/TooltipContainer.cs
```diff
@@ -213,6 +222,15 @@ private ITooltipContentProvider findTooltipTarget()
...
                 return null;
 
+            return handlePotentialTarget(targetCandidate);
+        }
+
+        private ITooltipContentProvider handlePotentialTarget(ITooltipContentProvider targetCandidate)
+        {
+            // this method is intentionally split out from the main lookup above as it has several expensive delegate (LINQ) allocations.
+            // this allows the case where no tooltip is displayed to run with no allocations.
+            // further optimisation work can be done here to reduce allocations while a tooltip is being displayed.
```
smoogipoo, line 232: Not that I disagree with splitting out such a huge method, and it's probably good from a JIT perspective too, but I don't think this reasoning is valid? LINQ allocations don't occur until the code is run, unless there should've been more code inside this method?

peppy: it's the delegates inside the linq (i tried to mention that in the comment).

delegates are allocated on entering the method.

smoogipoo: Makes no sense for that to happen. I've benchmarked it to confirm:

|                                 Method |        Mean |     Error |    StdDev | Ratio |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------------------- |------------:|----------:|----------:|------:|-------:|------:|------:|----------:|
|               MethodWithInlineDelegate | 16,789.2 ns | 326.93 ns | 335.73 ns |  1.00 | 1.8921 |     - |     - |   64000 B |
|    MethodWithUnreachableInlineDelegate |    547.2 ns |   0.87 ns |   0.77 ns |  0.03 |      - |     - |     - |         - |
| MethodWithUnreachableExtractedDelegate |    545.8 ns |   0.34 ns |   0.30 ns |  0.03 |      - |     - |     - |         - |

```csharp
[MemoryDiagnoser]
public class BenchmarkMethodDelegateAllocations
{
    [Benchmark(Baseline = true)]
    public int MethodWithInlineDelegate()
    {
        int res = 0;

        for (int i = 0; i < 1000; i++)
            res += methodWithInlineDelegate();

        return res;
    }

    [Benchmark]
    public int MethodWithUnreachableInlineDelegate()
    {
        int res = 0;

        for (int i = 0; i < 1000; i++)
            res += methodWithUnreachableInlineDelegate(false);

        return res;
    }

    [Benchmark]
    public int MethodWithUnreachableExtractedDelegate()
    {
        int res = 0;

        for (int i = 0; i < 1000; i++)
            res += methodWithUnreachableExtractedDelegate(false);

        return res;
    }

    private int methodWithInlineDelegate()
    {
        return testMethod(() => this);
    }

    private int methodWithUnreachableInlineDelegate(bool cond)
    {
        if (!cond)
            return 0;

        return testMethod(() => this);
    }

    private int methodWithUnreachableExtractedDelegate(bool cond)
    {
        if (!cond)
            return 0;

        return extractedMethod();
    }

    [MethodImpl(MethodImplOptions.NoInlining)]
    private int extractedMethod() => testMethod(() => this);

    private int testMethod(Func<object> obj) => obj().GetHashCode();
}
```

peppy: sure it's not getting somehow optimised away somehow? i'll re-test, but there's no other changes i made between the two profile runs here and it's apparent it helped.

smoogipoo: I get similar results in DEBUG, so I'm pretty sure nothing's getting optimised away.

peppy: i didn't mean debug/release, more the size/complexity of the method. all my testing for all PRs was done on release.

smoogipoo: The JIT is not allowed to do anything to debug builds, regardless of size/complexity. That would negate the entire point of debug builds.

smoogipoo: Are you sure it isn't just the allocation that you removed above? Keep in mind that method lamdas always allocate.

peppy: I haven't done anything in debug builds anywhere because it can change allocations.

I'll re-test this scenario tonight.

smoogipoo: TIL referencing locals will _always_ allocate the lambda classes, whereas referencing fields doesn't.

Fields:
unknown

Locals:
unknown

Good to... know? I guess?

I'll still report to MS and see if this is intentional or not. It seems like even the compiler has the knowledge to understand the code is unreachable in the `fields` case.

PR #3742, opened 2020-07-19
Fix LINQ allocations in focusTopMostRequestingDrawable

peppy:
Before:
dotMemory-201 0 20200703 51832_2020-07-19_19-22-15

After:
image


osu.Framework/Input/InputManager.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
```
bdach, line 1: something ate the byte order mark here

PR #3743, opened 2020-07-19
Fix LINQ / List alllocations in PropagateBlockableEvent

peppy:
Two small changes:
- Don't alloc lists for `IRequireHighFrequency`
- Don't alloc enumerator in `PropagateBlockingEvent` (due to `IEnumerable` param type)
- [x] Depends on https://github.com/ppy/osu-framework/pull/3739.

As one requires the other I did this in a single PR.

Before:
image

After:
image

### DIFF
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -387,7 +387,7 @@ namespace osu.Framework.Input
 
             if (CurrentState.Mouse.IsPositionValid)
             {
-                highFrequencyDrawables.Clear();
+                Debug.Assert(highFrequencyDrawables.Count == 0);
 
                 foreach (var d in PositionalInputQueue)
                 {
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -1,4 +1,4 @@
-// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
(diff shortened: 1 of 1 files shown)


osu.Framework/Input/InputManager.cs
```diff
@@ -385,7 +387,17 @@ protected override void Update()
 
             if (CurrentState.Mouse.IsPositionValid)
             {
-                PropagateBlockableEvent(PositionalInputQueue.Where(d => d is IRequireHighFrequencyMousePosition), new MouseMoveEvent(CurrentState));
+                highFrequencyDrawables.Clear();
```
bdach, line 390: The double clear is a little bit weird to me. Is there a particular reason for this? I'd probably either just clear once at the start of the method (although that could hold references for a frame) or clear at the end and make this an assert.

peppy: Assert sounds good. I added this here to drop references immediately, so would probably go with that direction (assert before, clear after).

osu.Framework/Input/InputManager.cs
```diff
@@ -687,7 +699,7 @@ protected virtual void HandleMouseButtonStateChange(ButtonStateChangeEvent<Mouse
         /// <param name="drawables">The drawables in the queue.</param>
         /// <param name="e">The event.</param>
         /// <returns>Whether the event was handled.</returns>
-        protected virtual bool PropagateBlockableEvent(IEnumerable<Drawable> drawables, UIEvent e)
+        protected virtual bool PropagateBlockableEvent(List<Drawable> drawables, UIEvent e)
```
smoogipoo, line 702: See CI failure. `IList<Drawable>` is probably good here to keep the interface simple (and iterate via index).

peppy: I've made is `Slim`, not `IList` because I don't think there will be any other usages.

PR #3744, opened 2020-07-19
Fix Enumerable allocations from AvailableInputHandlers

peppy:
Before:
image

After:
image

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -655,7 +655,7 @@ namespace osu.Framework.Platform
                     h.Dispose();
             }
 
-            AvailableInputHandlers = CreateAvailableInputHandlers();
+            AvailableInputHandlers = CreateAvailableInputHandlers().ToArray();
 
             foreach (var handler in AvailableInputHandlers)
             {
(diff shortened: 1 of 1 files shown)


osu.Framework/Platform/GameHost.cs
```diff
@@ -846,7 +846,7 @@ private void setVSyncMode()
 
         protected abstract IEnumerable<InputHandler> CreateAvailableInputHandlers();
 
-        public IEnumerable<InputHandler> AvailableInputHandlers { get; private set; }
+        public InputHandler[] AvailableInputHandlers { get; private set; }
```
bdach, line 849: Unsure of this API surface again. This makes elements mutable which is probably not something we want.

`IReadOnly{List,Collection}<InputHandler>` would both be better but haven't checked how they perform alloc-wise.

peppy: I went for `IReadOnlyList` initially but as you suspect, it still causes mass allocs on the enumerator (due to the `interface` type).

smoogipoo: Is `ImmutableArray` possible to use?

peppy: I'll give that a try tonight. Should work, I think.

peppy:
Blocking pending re-profiling.

peppy:
`ImmutableArray` works fine, as long as we are okay with also changing the next PR to also be `ImmutableArray` (#3745). Have tested with both and allocs look good.

PR #3747, opened 2020-07-19
Avoid comparing buttons when there is guaranteed to be no difference

peppy:
Reduces allocations when moving the mouse with no buttons pressed.

Before:
image

After:
image

### DIFF
--- a/osu.Framework/Input/States/ButtonStates.cs
+++ b/osu.Framework/Input/States/ButtonStates.cs
@@ -1,6 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
@@ -48,13 +49,22 @@ namespace osu.Framework.Input.States
             return true;
         }
 
-        public bool HasAnyButtonPressed => pressedButtons.Any();
+        public bool HasAnyButtonPressed => pressedButtons.Count > 0;
 
         /// <summary>
         /// Enumerates the differences between ourselves and a previous <see cref="ButtonStates{TButton}"/>.
         /// </summary>
         /// <param name="lastButtons">The previous <see cref="ButtonStates{TButton}"/>.</param>
-        public ButtonStateDifference EnumerateDifference(ButtonStates<TButton> lastButtons) => new ButtonStateDifference(lastButtons.Except(this).ToArray(), this.Except(lastButtons).ToArray());
+        public ButtonStateDifference EnumerateDifference(ButtonStates<TButton> lastButtons)
+        {
+            if (!lastButtons.HasAnyButtonPressed)
+                return new ButtonStateDifference(Array.Empty<TButton>(), this);
+
+            if (!HasAnyButtonPressed)
+                return new ButtonStateDifference(lastButtons, Array.Empty<TButton>());
+
+            return new ButtonStateDifference(lastButtons.Except(this).ToArray(), this.Except(lastButtons).ToArray());
+        }
 
         /// <summary>
         /// Copies the state of another <see cref="ButtonStates{TButton}"/> to ourselves.
--- a/osu.Framework/Input/States/ButtonStates.cs
+++ b/osu.Framework/Input/States/ButtonStates.cs
@@ -58,7 +58,7 @@ namespace osu.Framework.Input.States
         public ButtonStateDifference EnumerateDifference(ButtonStates<TButton> lastButtons)
         {
             if (!lastButtons.HasAnyButtonPressed)
-                return new ButtonStateDifference(Array.Empty<TButton>(), this.pressedButtons);
+                return new ButtonStateDifference(Array.Empty<TButton>(), pressedButtons);
 
             if (!HasAnyButtonPressed)
                 return new ButtonStateDifference(lastButtons.pressedButtons, Array.Empty<TButton>());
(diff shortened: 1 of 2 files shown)


osu.Framework/Input/States/ButtonStates.cs
```diff
@@ -48,13 +49,22 @@ public bool SetPressed(TButton button, bool pressed)
...
         /// </summary>
         /// <param name="lastButtons">The previous <see cref="ButtonStates{TButton}"/>.</param>
-        public ButtonStateDifference EnumerateDifference(ButtonStates<TButton> lastButtons) => new ButtonStateDifference(lastButtons.Except(this).ToArray(), this.Except(lastButtons).ToArray());
+        public ButtonStateDifference EnumerateDifference(ButtonStates<TButton> lastButtons)
+        {
+            if (!lastButtons.HasAnyButtonPressed)
+                return new ButtonStateDifference(Array.Empty<TButton>(), pressedButtons);
+
+            if (!HasAnyButtonPressed)
+                return new ButtonStateDifference(lastButtons.pressedButtons, Array.Empty<TButton>());
```
bdach, line 64: These changes are causing the test failures themselves - a copy somewhere is necessary. The scenario is as follows:

* `pressedButtons` is a list, so it doesn't get copied; `ButtonStateDifference` just receives a reference to the list.
* `ButtonInput` doesn't actually operate on the list, but does linq magic and defers execution:

https://github.com/ppy/osu-framework/blob/b4e6608a8d22b734b14536cb66844c3e61795591/osu.Framework/Input/StateChanges/ButtonInput.cs#L44-L50

* `Apply()` enumerates over these, and calls `SetPressed`:

https://github.com/ppy/osu-framework/blob/b4e6608a8d22b734b14536cb66844c3e61795591/osu.Framework/Input/StateChanges/ButtonInput.cs#L69-L76

...therefore modifying `Entries`, as it's not a proper list, but a concat iterator which has `pressedButtons` inside, also passed by reference.

Copying `pressedButtons` passed to `ButtonStateDifference` and enumerating the concat iterator at `ButtonInput` both seem to resolve the test failures. I think making a copy here (by calling `ToArray()` explicitly) is safer as it preserves previous semantics.

peppy: Yep. Probably another instance we will need to pool the arrays to avoid the alloc, or re-think at a different level.

peppy:
I've fixed this via copy in the `ButtonInput` ctor (and removed the linq in there for safety) but will block to double-check allocations. Pretty sure we're in the clear, though.

peppy:
Unblocking, still looks fine.

peppy:
Note that I didn't check allocs while pressing buttons, but if that's an issue (potential `ImmutableArray` allocs) it's better addressed in a later PR. My focus so far has only been on no input or move-only.

smoogipoo:
I've changed one code path to use `ImmutableArray.Create()`:

```
[Benchmark]
public ImmutableArray<int> ArrayToImmutable() => new[] { 1, 2, 3 }.ToImmutableArray();

[Benchmark]
public ImmutableArray<int> CreateImmutable() => ImmutableArray.Create(1, 2, 3);

|           Method |       Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------- |-----------:|----------:|----------:|-------:|------:|------:|----------:|
| ArrayToImmutable | 126.481 ns | 1.0339 ns | 0.9671 ns | 0.0608 |     - |     - |      80 B |
|  CreateImmutable |   6.272 ns | 0.0284 ns | 0.0252 ns | 0.0305 |     - |     - |      40 B |
```

We need to be really careful about this stuff, because it's super easy to say we're saving allocations but causing overall performance regressions. `ImmutableArray` is still very new to me so I've got a while to go in figuring out best practices / performance characteristics.

For example. I think we'll eventually want to have `ButtonStateDifference` store `ImmutableArray`s so that `ButtonInput` can then construct a builder using the exact length of differences and significantly improve performance even more:

```
[Benchmark]
public ImmutableArray<int> ListThenToImmutable()
{
    var list = new List<int>();

    for (int i = 0; i < 10; i++)
        list.Add(i);

    return list.ToImmutableArray();
}

[Benchmark]
public ImmutableArray<int> ImmutableBuilder()
{
    var builder = ImmutableArray.CreateBuilder<int>(10);

    for (int i = 0; i < 10; i++)
        builder.Add(i);

    return builder.MoveToImmutable();
}

|              Method |      Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------- |----------:|---------:|---------:|-------:|------:|------:|----------:|
| ListThenToImmutable | 172.33 ns | 1.538 ns | 1.438 ns | 0.2196 |     - |     - |     288 B |
|    ImmutableBuilder |  32.80 ns | 0.139 ns | 0.123 ns | 0.0732 |     - |     - |      96 B |
```

peppy:
Yep, definitely. My goal was to get allocs down to close-to-zero then do a performance pass afterwards. It's hard to cover all bases at once.

smoogipoo:
I don't think you're going to get much from a post-mortem sampling, just because the amount we're seeing here is in the range of nanoseconds. But when you add 100ns to 20 different places, suddenly you've got a very diverse call stack that increases runtime by 0.2ms, which by itself is going to be in the 0% range of an entire game.

That's why I'd advocate for profiling performance (even if just via benchmarks) as these changes get made instead.

peppy:
I'll be testing on an empty game, should still be easy to notice such tiny issues by isolating everything else away.

PR #3748, opened 2020-07-19
Avoid using LINQ to calculate button differences

peppy:
This reduces allocations when buttons are pressed, such as a drag operation.

- [x] Depends on https://github.com/ppy/osu-framework/pull/3747

Before:
image

After
dotMemory-201 0 20200703 51832_2020-07-19_23-07-36

### DIFF
--- a/osu.Framework/Input/States/ButtonStates.cs
+++ b/osu.Framework/Input/States/ButtonStates.cs
@@ -68,7 +68,7 @@ namespace osu.Framework.Input.States
 
             foreach (var b in pressedButtons)
             {
-                if (!lastButtons.Contains(b))
+                if (!lastButtons.pressedButtons.Contains(b))
                     (pressed ??= new List<TButton>()).Add(b);
             }
 
--- a/osu.Framework/Input/States/ButtonStates.cs
+++ b/osu.Framework/Input/States/ButtonStates.cs
@@ -16,12 +16,12 @@ namespace osu.Framework.Input.States
     public class ButtonStates<TButton> : IEnumerable<TButton>
         where TButton : struct
     {
-        private List<TButton> pressedButtons = new List<TButton>();
+        private HashSet<TButton> pressedButtons = new HashSet<TButton>();
 
         public ButtonStates<TButton> Clone()
         {
             var clone = (ButtonStates<TButton>)MemberwiseClone();
-            clone.pressedButtons = new List<TButton>(pressedButtons);
+            clone.pressedButtons = new HashSet<TButton>(pressedButtons);
             return clone;
         }
 
@@ -63,7 +63,22 @@ namespace osu.Framework.Input.States
             if (!HasAnyButtonPressed)
                 return new ButtonStateDifference(lastButtons.pressedButtons, Array.Empty<TButton>());
 
-            return new ButtonStateDifference(lastButtons.Except(this).ToArray(), this.Except(lastButtons).ToArray());
+            List<TButton> released = null;
+            List<TButton> pressed = null;
+
+            foreach (var b in pressedButtons)
+            {
+                if (!lastButtons.Contains(b))
+                    (pressed ??= new List<TButton>()).Add(b);
+            }
+
+            foreach (var b in lastButtons.pressedButtons)
+            {
+                if (!pressedButtons.Contains(b))
+                    (released ??= new List<TButton>()).Add(b);
+            }
+
+            return new ButtonStateDifference(released ?? Enumerable.Empty<TButton>(), pressed ?? Enumerable.Empty<TButton>());
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


bdach:
Code-wise looks sane but the issue in the dependency needs addressing here too.

osu.Framework/Input/States/ButtonStates.cs
```diff
@@ -16,12 +16,12 @@ namespace osu.Framework.Input.States
     public class ButtonStates<TButton> : IEnumerable<TButton>
         where TButton : struct
     {
-        private List<TButton> pressedButtons = new List<TButton>();
+        private HashSet<TButton> pressedButtons = new HashSet<TButton>();
```
smoogipoo, line 19: Part of me wonders whether we shouldn't just keep this a `List`. I know it basically makes enumeration O(N^2) but (including the fix I mentioned in that other comment) I've generated the following benchmarks:

```
[Benchmark]
public int EnumerateBothDifferences()
{
    var state1 = new ButtonStates<Key>();
    var state2 = new ButtonStates<Key>();

    for (int i = 0; i < N; i++)
        state1.SetPressed((Key)i, true);

    for (int i = N; i < 2 * N; i++)
        state2.SetPressed((Key)i, true);

    return state1.EnumerateDifference(state2).Pressed.Length;
}

List:
|                   Method |   N |       Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------- |---- |-----------:|---------:|---------:|-------:|------:|------:|----------:|
| EnumerateBothDifferences |   1 |   196.8 ns |  0.49 ns |  0.43 ns | 0.0119 |     - |     - |     400 B |
| EnumerateBothDifferences |   5 |   482.8 ns |  2.55 ns |  2.39 ns | 0.0191 |     - |     - |     656 B |
| EnumerateBothDifferences |  10 |   843.9 ns |  3.31 ns |  2.76 ns | 0.0305 |     - |     - |    1040 B |
| EnumerateBothDifferences |  50 | 3,944.6 ns | 14.56 ns | 12.91 ns | 0.0916 |     - |     - |    3088 B |
| EnumerateBothDifferences | 100 | 9,741.7 ns | 62.08 ns | 55.03 ns | 0.1678 |     - |     - |    5632 B |

Hashset:
|                   Method |   N |       Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------- |---- |-----------:|---------:|---------:|-------:|------:|------:|----------:|
| EnumerateBothDifferences |   1 |   262.9 ns |  0.65 ns |  0.51 ns | 0.0176 |     - |     - |     592 B |
| EnumerateBothDifferences |   5 |   645.0 ns |  3.90 ns |  3.45 ns | 0.0315 |     - |     - |    1072 B |
| EnumerateBothDifferences |  10 | 1,180.4 ns |  4.93 ns |  4.37 ns | 0.0572 |     - |     - |    1936 B |
| EnumerateBothDifferences |  50 | 4,624.1 ns | 15.98 ns | 14.95 ns | 0.2136 |     - |     - |    7376 B |
| EnumerateBothDifferences | 100 | 8,812.9 ns | 32.03 ns | 28.40 ns | 0.4425 |     - |     - |   15264 B |
```

Admittedly, this result does look a little weird, but it's probably being overweighed by all the other operations (`SetPressed()`, `ToArray()`, adding to two `Lists`, etc)... So maybe keep it for now but definitely something that should be considered.

bdach: Time-wise it makes sense. I recall reading that for small N `List.Contains` is faster.

Not sure why allocations also grew, though.

smoogipoo: The weird part is I have a separate benchmarks project that has the following results:

```
|  Method |    N |          Mean |        Error |       StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------- |----- |--------------:|-------------:|-------------:|------:|------:|------:|----------:|
|    List |    1 |      16.77 ns |     0.016 ns |     0.015 ns |     - |     - |     - |         - |
| HashSet |    1 |      18.62 ns |     0.065 ns |     0.054 ns |     - |     - |     - |         - |
|    List |   10 |     115.80 ns |     0.602 ns |     0.563 ns |     - |     - |     - |         - |
| HashSet |   10 |      98.62 ns |     0.446 ns |     0.396 ns |     - |     - |     - |         - |
|    List |  100 |   4,455.30 ns |     8.634 ns |     7.210 ns |     - |     - |     - |         - |
| HashSet |  100 |     862.24 ns |     1.493 ns |     1.323 ns |     - |     - |     - |         - |
|    List | 1000 | 290,684.85 ns | 1,512.307 ns | 1,340.621 ns |     - |     - |     - |         - |
| HashSet | 1000 |   8,508.37 ns |    10.891 ns |     9.094 ns |     - |     - |     - |         - |
```
Which goes contrary to the results presented here.

peppy: I'm okay with either, take your pick.

smoogipoo: Let's keep it as it is for now because performance is going to be better anyway with the other thing I commented on. Need to investigate during gameplay because it's hard to assess from synthetics here.

peppy: Just to reiterate, I'm only testing for no-buttons-pressed right now. The rest will be optimised in another pass.

PR #3749, opened 2020-07-19
Use ConcurrentQueue.Clear instead of TryDequeue loop in FrameStatisticsDisplay

turbedi:


bdach:
[Source, for reference](https://github.com/dotnet/runtime/blob/6072e4d3a7a2a1493f514cdf4be75a3d56580e84/src/libraries/System.Private.CoreLib/src/System/Collections/Concurrent/ConcurrentQueue.cs#L790-L808).

Definitely looks safer than dequeueing in a loop to me. Don't think we care too much about reads in-flight as long as the collection is cleared.

peppy:
I don't believe `.Clear` existed when this was implemented. Looks good.

PR #3753, opened 2020-07-21
Add wrapper to create a slim read-only list

smoogipoo:
This should help with https://github.com/ppy/osu-framework/pull/3739 (it's a drop-in replacement). Supports `List<T>` for now, though in the future it could be upgraded to support more lists with minimal performance cost (~5us in the benchmarks).

|             Method |      Mean |    Error |   StdDev | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------- |----------:|---------:|---------:|------:|--------:|-------:|------:|------:|----------:|
|               List |  29.03 us | 0.051 us | 0.045 us |  1.00 |    0.00 |      - |     - |     - |         - |
|     ListAsReadOnly | 124.87 us | 0.520 us | 0.461 us |  4.30 |    0.02 | 1.7090 |     - |     - |   64000 B |
| ListAsSlimReadOnly |  28.89 us | 0.051 us | 0.045 us |  0.99 |    0.00 |      - |     - |     - |         - |

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkSlimReadOnlyCollection.cs
+++ b/osu.Framework.Benchmarks/BenchmarkSlimReadOnlyCollection.cs
@@ -1,0 +1,57 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Collections.Generic;
+using BenchmarkDotNet.Attributes;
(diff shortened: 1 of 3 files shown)


smoogipoo:
Also I've made it a struct so you can `.AsSlimReadOnly()` inline when required. My benchmarks can't discern a cost involved in doing so, and it can always be upgraded into a reference if required.

peppy:
Works well

Issue #3639, opened 2020-06-22
Add configurable joystick axis deadzones


Aarc-Noa:
Current deadzones are not configurable and can lead to either very small (most axes) or very large deadzones (triggers) when the controller is connected.

maromalo:
Closed by https://github.com/ppy/osu-framework/pull/5122

Issue #2854, opened 2019-09-23
Framework can't handle left and right trigger (LT and RT) on joystick


GSculerlor:
https://streamable.com/760vg

**Framework ver**: 2019.921.0
**Joystick used**: Xbox 360 Controller (Wireless) and Rexus Gladius GX2 (Wired)

Jeremiidesu:
I think I found a way (on the latest master) to work around this problem.

The Xbox 360 controller I use (Wired), has LT and RT on axes 2 and 5 (AX3 and AX6 on the visual test scene). Plugging the controller or running the visual tests with the triggered **half-pressed** (not completely pressed, nor completely released) allows the visual tests to properly display pressed or release states. 

The main issue is that the values for the triggers go from -1 (completely released) to 1 (completely pressed down), causing the deadzone calculation to erroneously set the trigger deadzone to ~1.05 at line 223 here:

https://github.com/ppy/osu-framework/blob/b169c45c13e7cdc45b93f165d42ce23be4c028d5/osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs#L210-L225

Temporarily removing `axisValue +` from line 223 also solves the problem, with triggers being properly detected as axes as shown in the following screenshot: 

![LT and RT correctly displayed in AX3 and AX6 while the HTML5 Gamepad Tester reports B6 and B7, the triggers, as completely pressed](https://cdn.platea.moe/Jeremii/2020/05/brave_IbsRDutbXg.png)

Releasing the triggers turns the AX3 and AX6 squares red.

peppy:
This should now be resolved via #3900 (with `--sdl` enabled). Going to tentatively close - please comment if you have an issue with a specific controller.

PR #3759, opened 2020-07-22
Rescale joystick axis values according to deadzones

### LINKED ISSUE #3712: Scale joystick axis values according to deadzones
## Current situation

Currently, the way axis deadzones work in the framework is:

- If the current axis value is within the deadzone, the provided axis value is 0
- Otherwise, the provided axis value is the raw value

In the situation where the deadzone is large (say, 0.5), reported axis values would be as follows:

|Raw value|Provided value after deadzone application|
|---|---|
|0|0|
|0.25|0|
|0.5|0.5|
|0.75|0.75|
|1|1|

image

## My suggestion

Scaling them down as such would mitigate the sudden jumps:

|Raw value|Provided value after deadzone and scale|
|---|---|
|0|0|
|0.25|0|
|0.5|0|
|0.75|0.5|
|1|1|

image

Here's an [interactive GeoGebra file](https://www.geogebra.org/graphing/qjuhtxnv) to help visualize the idea.

Jeremiidesu:
Closes #3712.

Only issue left seems to be rounding errors that are particularly visible on large deadzones, which is fairly easy to replicate using triggers.

On my Dualshock 4, extreme values on triggers (deadzone `0.999f`) without clamping are `-1.0152590` and `0.9847410`.

### DIFF
--- a/osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
+++ b/osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
@@ -224,8 +224,8 @@ namespace osu.Framework.Input.Handlers.Joystick
                         if (Precision.AlmostEquals(0, axisValue))
                             continue;
 
-                        // Cap deadzones at 0.999f to avoid division by zero when rescaling
-                        defaultDeadZones.Value[i] = Math.Min(0.999f, axisValue + deadzone_threshold);
+                        // Cap deadzone at 0.5f to avoid division by zero and catastrophic cancellation when rescaling
+                        defaultDeadZones.Value[i] = Math.Min(0.5f, axisValue + deadzone_threshold);
                     }
                 }
             }
--- a/osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
+++ b/osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
@@ -114,12 +114,15 @@ namespace osu.Framework.Input.Handlers.Joystick
                     if (Precision.AlmostEquals(value, 0, deadzone))
                         // Round values in the deadzone to zero.
                         value = 0;
+                    else
+                        // Rescale values then clamp back into [-1;1] range.
+                        value = Math.Clamp((value - deadzone * Math.Sign(value)) / (1 - deadzone), -1, 1);
 
                     AxesValues[i] = value;
 
-                    if (value > deadzone)
+                    if (value > 0)
                         Buttons.SetPressed(JoystickButton.FirstAxisPositive + i, true);
-                    else if (value < -deadzone)
+                    else if (value < 0)
                         Buttons.SetPressed(JoystickButton.FirstAxisNegative + i, true);
                 }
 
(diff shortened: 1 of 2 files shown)


osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
```diff
@@ -221,7 +224,8 @@ public void UpdateRawState(osuTK.Input.JoystickState rawState)
                         if (Precision.AlmostEquals(0, axisValue))
                             continue;
 
-                        defaultDeadZones.Value[i] = axisValue + deadzone_threshold;
+                        // Cap deadzones at 0.999f to avoid division by zero when rescaling
+                        defaultDeadZones.Value[i] = Math.Min(0.999f, axisValue + deadzone_threshold);
```
bdach, line 228: With the aforementioned rounding anomalies (which are due to [catastrophic cancellation](https://en.wikipedia.org/wiki/Loss_of_significance) and which I can't see a way out of) when calculating `1 - deadzone`) I'd be in favour of capping this even lower (`0.95f`? `0.9f`? maybe even `0.5f`?)

Also if we're sanity-checking this, this should probably be clamped from the bottom (to 0, at least).

Jeremiidesu: I guess realistically deadzones shouldn't exceed `0.5f` so I'll go for that.

As for clamping from the bottom, I don't understand what you have in mind by that.

bdach: Well, the deadzone should be within the range of [0,1], right? I'd feel safer with this change if that was explicitly clamped from both sides here (although not sure whether that's a plausible scenario to happen).

Jeremiidesu: Given that just a few lines above, the `axisValue` is set to be the absolute value of the raw axis value, unless cosmic rays decide to intervene it should be impossible to have a negative deadzone. That said, I'll switch to a clamp for clarity.

bdach: Fair enough, wouldn't bother with clamping then.

bdach:
Seems like this PR is doing what it intends to.

As a vaguely related discussion, not sure whether using initial axis value readouts to seed the default deadzones is a good choice (consider launching a game with the stick/triggers held to their extreme values - unless the game specifically implements deadzone adjustment/calibration, it will be stuck with mostly non-functional input), but that is not really in scope here and a separate thing entirely.

frenzibyte:
Yeah, and there was an [old issue](https://github.com/ppy/osu-framework/issues/2854) opened about triggers in the Xbox 360 Controller not being handled properly, which turned out to be because of this exact logic.

I believe [the ability to configure joystick axis deadzones](https://github.com/ppy/osu-framework/issues/3639) should leverage this behaviour to at least always have usable joysticks.

peppy:
Tested on the only controller I have on hand and seems to work as expected (doesn't break). Let's get this in.

PR #3767, opened 2020-07-28
Fix android screen suspension logic

Game4all:
It turns out the logic wasn't even right for android screen suspension as it was missing a negation. Until now, the `KeepScreenOn` flag was set when `AllowScreenSuspension.Value = true`, which should actually be the other way around.


peppy:
Weird, because I did actually test this on an android device..

Game4all:
I tested both `master` and this PR on one of my devices to be sure I didn't regress behavior

bdach:
I recall testing this manually too but might have not paid enough attention.

PR #3779, opened 2020-07-31
Improve performance of transform logic when many transforms are present

peppy:
Focuses on decreasing the iteration overhead of per-frame transform processing when many transforms are present in the transforms list (and cannot be removed due to rewind support or otherwise).

# Basic idea

Store the last fully applied transform index and only continue processing from that point onwards. To keep things simple, invalidate this cache on any add/remove/rewind operation. These scenarios can be optimised in the future if/when required.

# Difficulties

To make this work, transforms had to be split into per-`TargetMember` lists. If this was not the case, it becomes hard to choose a starting index (originally I was using the "minimum" of any `TargetMember` but this can lead to situations with no optimisation is possible).

As such, there is now one new `TargetMemberTransformTracker` class per `TargetMember` being tracked for a `Transformable`. These are only constructed when used and the memory / allocation overhead does not seem any worse than it was previously.

# Benchmarks

## UpdateTransformsWithManyPresent 

|                          Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
| (master)  | 821.0 ms | 22.08 ms | 65.09 ms |     - |     - |     - |      48 B |
| (this PR) | 0.2574 ms | 3.84 us  | 3.40 us  |     - |     - |     - |       2 B |

- Closes https://github.com/ppy/osu/issues/9605

- [x] Add benchmarks (not sure if required; osu!-side storyboard shows an improvement from 40ms frame times to <0.1ms)
- [x] Add lazy initialisation of dictionary (similar to `transformsLazy`)
- [x] Assess whether dictionary should only be initialised/used over a certain list length (to save on memory @smoogipoo thoughts?)
- [x] Move transform add / remove / removeAt into dedicated methods to avoid needing to do the cache clear at each usage?

I've implemented this in the simplest way possible. Further optimisations (adjusting indices rather than completing flushing the cache) are obviously possible but come with added complexity. Mainly PRing early for feedback (cases which may need to be tested and aren't already covered; general direction etc.) and considerations.

### DIFF
--- a/osu.Framework/Graphics/Transforms/TargetMemberTransformTracker.cs
+++ b/osu.Framework/Graphics/Transforms/TargetMemberTransformTracker.cs
@@ -5,8 +5,8 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using osu.Framework.Caching;
 using osu.Framework.Lists;
-using osu.Framework.Timing;
 
 namespace osu.Framework.Graphics.Transforms
 {
@@ -29,9 +29,12 @@ namespace osu.Framework.Graphics.Transforms
 
         private readonly Transformable transformable;
 
-        private int? lastAppliedTransformIndex;
-
         private readonly Lazy<List<Action>> removalActions = new Lazy<List<Action>>(() => new List<Action>());
+
+        /// <summary>
+        /// The index of the last transform in <see cref="transforms"/> to be applied to completion.
+        /// </summary>
+        private readonly Cached<int> lastAppliedIndex = new Cached<int>();
 
         public TargetMemberTransformTracker(Transformable transformable, string targetMember)
         {
@@ -88,13 +91,13 @@ namespace osu.Framework.Graphics.Transforms
                 }
             }
 
-            for (int i = getLastAppliedIndex() ?? 0; i < transforms.Count; ++i)
+            for (int i = lastAppliedIndex.IsValid ? lastAppliedIndex.Value : 0; i < transforms.Count; ++i)
             {
                 var t = transforms[i];
 
                 var tCanRewind = !transformable.RemoveCompletedTransforms && t.Rewindable;
 
-                bool shouldFlushLastApplicationCache = false;
+                bool flushAppliedCache = false;
 
                 if (time < t.StartTime)
                     break;
@@ -106,7 +109,7 @@ namespace osu.Framework.Graphics.Transforms
                     // Since following transforms acting on the same target member are immediately removed when a
                     // new one is added, we can be sure that previous transforms were added before this one and can
                     // be safely removed.
-                    for (int j = getLastAppliedIndex() ?? 0; j < i; ++j)
+                    for (int j = lastAppliedIndex.IsValid ? lastAppliedIndex.Value : 0; j < i; ++j)
                     {
                         var u = transforms[j];
                         if (u.TargetMember != t.TargetMember) continue;
@@ -120,7 +123,7 @@ namespace osu.Framework.Graphics.Transforms
                         if (!tCanRewind)
                         {
                             transforms.RemoveAt(j--);
-                            shouldFlushLastApplicationCache = true;
+                            flushAppliedCache = true;
                             i--;
 
                             if (u.OnAbort != null)
@@ -148,7 +151,7 @@ namespace osu.Framework.Graphics.Transforms
                         if (!tCanRewind)
                         {
                             transforms.RemoveAt(i--);
-                            shouldFlushLastApplicationCache = true;
+                            flushAppliedCache = true;
                         }
 
                         if (t.IsLooping)
@@ -171,18 +174,18 @@ namespace osu.Framework.Graphics.Transforms
                             // this could be added back at a lower index than where we are currently iterating, but
                             // running the same transform twice isn't a huge deal.
                             transforms.Add(t);
-                            shouldFlushLastApplicationCache = true;
+                            flushAppliedCache = true;
                         }
                         else if (t.OnComplete != null)
                             removalActions.Value.Add(t.OnComplete);
                     }
                 }
 
-                if (shouldFlushLastApplicationCache)
+                if (flushAppliedCache)
                     resetLastAppliedCache();
                 // if this transform is applied to end, we can be sure that all previous transforms have been completed.
                 else if (t.AppliedToEnd)
-                    setLastAppliedIndex(i + 1);
+                    lastAppliedIndex.Value = i + 1;
             }
 
             invokePendingRemovalActions();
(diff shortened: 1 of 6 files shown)


peppy:
For verbosity, the test failure is from a very specific case as demonstrated by the test:

```
i=0  t=100 --------------- t = 1000
i=1       t=500  ------ t=750
```

Because I was using the `Max` last applied time, i=0 would be skipped in some cases.

I've fixed this issue by using `Min` and tracking `Applied` state as well.

As an aside, if this change is considered too complex / illegible in `Transformable`, the direction (which I believe we have discussed before) would be to split out transform logic into one processor class per TargetMember, maintaining its own pure list.

Of course, there are extra considerations if we go in this direction (ie. how do we apply transforms in chronological sequence, rather than per-member.. if this is even required) which I have not properly thought over.

peppy:
Blocking for now, will re-visit with per-TargetMember implementation as the fix for that test failure has regressed the effectiveness of the optimisation. While still possible, it'd likely require two dictionaries in Transformable (tracking `Applied` and `AppliedtoEnd`) at which point the complexity seems too high.

peppy:
Good to go.

Before:
```
|                          Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
| UpdateTransformsWithManyPresent | 821.0 ms | 22.08 ms | 65.09 ms |     - |     - |     - |      48 B |
```

After:
```
|                          Method |     Mean |   Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------- |---------:|--------:|--------:|------:|------:|------:|----------:|
| UpdateTransformsWithManyPresent | 257.4 us | 3.84 us | 3.40 us |     - |     - |     - |       2 B |
```

osu.Framework/Graphics/Transforms/Transformable.cs
```diff
@@ -81,10 +79,29 @@ protected void UpdateTransforms()
             updateTransforms(Time.Current);
         }
 
-        private readonly Lazy<List<Action>> removalActions = new Lazy<List<Action>>(() => new List<Action>());
-
         private double lastUpdateTransformsTime;
 
+        private readonly Lazy<List<TargetMemberTransformTracker>> targetMemberTrackers =
+            new Lazy<List<TargetMemberTransformTracker>>(() => new List<TargetMemberTransformTracker>());
```
smoogipoo, line 85: I know this is not the purpose of this PR, but how do you feel about removing this `Lazy`?

```
|              Method |       Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------- |-----------:|----------:|----------:|-------:|------:|------:|----------:|
|          CreateLazy |  16.971 ns | 0.0799 ns | 0.0708 ns | 0.0549 |     - |     - |      72 B |
|          CreateList |   9.055 ns | 0.1176 ns | 0.1100 ns | 0.0305 |     - |     - |      40 B |
| GetListCountViaLazy | 691.485 ns | 1.4367 ns | 1.3439 ns | 0.0849 |     - |     - |     112 B |
|  GetListCountDirect | 558.191 ns | 2.9287 ns | 2.5962 ns | 0.0296 |     - |     - |      40 B |


[MemoryDiagnoser]
public class LazyVsListAllocation
{
    [Benchmark]
    public object CreateLazy() => new Lazy<List<object>>();

    [Benchmark]
    public object CreateList() => new List<object>();

    [Benchmark]
    public int GetListCountViaLazy()
    {
        var lazy = new Lazy<List<object>>();

        int c = 0;

        for (int i = 0; i < 1000; i++)
            c += lazy.Value.Count;

        return c;
    }

    [Benchmark]
    public int GetListCountDirect()
    {
        var list = new List<object>();

        int c = 0;

        for (int i = 0; i < 1000; i++)
            c += list.Count;

        return c;
    }
}
```

peppy: Damn, that overhead. So we never should use lazy for lists.

But where's the lazy-in-lazy you mention?

smoogipoo: This is the result for sortedlist:
```
|           Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
| CreateSortedList | 17.96 ns | 0.150 ns | 0.133 ns | 0.0019 |     - |     - |      64 B |
```
Lazy isn't that that much worse but still doesn't give a noticeable improvement.

> But where's the lazy-in-lazy you mention?

Every tracker has a lazy list of its own.

peppy: Oh wow, that's not necessary either. I guess it's safe to remove both, since the list construction already allocates less than the lazy version, right?

smoogipoo: That's my thinking, yep.

peppy: Both removed.

osu.Framework/Graphics/Transforms/TargetMemberTransformTracker.cs
```diff
@@ -0,0 +1,300 @@
...
+        /// <summary>
+        /// Used to assign a monotonically increasing ID to <see cref="Transform"/>s as they are added. This member is
+        /// incremented whenever a <see cref="Transform"/> is added.
+        /// </summary>
+        private ulong currentTransformID;
+
+        /// <summary>
+        /// The index of the last transform in <see cref="transforms"/> to be applied to completion.
+        /// </summary>
+        private readonly Cached<int> lastAppliedIndex = new Cached<int>();
```
smoogipoo, line 43: Are you against changing this to an `int?`? This feels a bit excessive for the use-cases readability-wise. There's a slight performance improvement due to avoiding the non-inlineable method call but that's an aside to the readability:

```
Cached:

|                          Method |     Mean |   Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------- |---------:|--------:|--------:|------:|------:|------:|----------:|
| UpdateTransformsWithManyPresent | 261.2 us | 5.12 us | 6.09 us |     - |     - |     - |         - |

int?:

|                          Method |     Mean |   Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------- |---------:|--------:|--------:|------:|------:|------:|----------:|
| UpdateTransformsWithManyPresent | 229.0 us | 0.47 us | 0.40 us |     - |     - |     - |         - |
```

peppy: it used to be an int and i changed it to a cached. can also roll this back if you want. see which you prefer https://github.com/ppy/osu-framework/pull/3779/commits/460eac95a0999200f9c68625026fd93cd30d775f

smoogipoo: I prefer the int? version

PR #3781, opened 2020-08-02
SDL fullscreen implementation

swoolcock:
Implements:
* Borderless and exclusive fullscreen.
* Cursor confinement.
* Dynamic resizing.

Notes:
* Needs testing in Linux.
* SDL's default implementation of cursor confinement on macOS is not perfect, and the cursor is able to temporarily exit the window until the event loop warps it back.  I believe there is a way of telling SDL to use a different method, will investigate.

SDL event loop process:
1. Execute any queued commands from property setters (`Size`, `Position`, etc.).
2. Handle all pending SDL events, which may update properties and/or queue `On...` callbacks.
3. Poll the mouse position (if the cursor is outside the window) and queue `OnMouseMove` callbacks.
4. Execute all `On...` callbacks.


bdach:
Ran this on Linux (ubuntu 18.04 + i3wm).

* Fullscreen looks to be passing the basic "smoke test" of just trying to turn it on.
* Mouse confinement sort of works if I don't change resolution. Getting a weird failure case of:
  - change resolution to smaller (i.e. 1600x900)
  - change resolution back to native (but _not_ default, which is 2560x1080 on my setup)
  - confinement applies for the previous windows size and not the new one; switching to _default_ instantly fixes it
* Remembering screen position doesn't really apply on my particular setup (tiling wm) so I haven't tested that; will re-log onto gnome, check and edit this.

Will also give this a quick spin on Windows (hopefully even tonight). Disclaimer - I have not read any source yet.

bdach:
Windows unfortunately immediately dies with

```
[runtime] 2020-08-02 20:55:35 [error]: An unhandled error has occurred.
[runtime] 2020-08-02 20:55:35 [error]: osuTK.Graphics.GraphicsContextMissingException: No context is current in the calling thread (ThreadId: 1).
[runtime] 2020-08-02 20:55:35 [error]: at osuTK.Graphics.ES30.ErrorHelper..ctor(IGraphicsContext context) in D:\Open Source\osuTK\src\osuTK\Graphics\ES30\ErrorHelper.cs:line 59
[runtime] 2020-08-02 20:55:35 [error]: at osuTK.Graphics.ES30.GL.GetString(StringName name) in D:\Open Source\osuTK\src\osuTK\Graphics\ES30\ES30.cs:line 16290
[runtime] 2020-08-02 20:55:35 [error]: at osu.Framework.Platform.Sdl.PassthroughGraphicsBackend.Initialise(IWindowBackend windowBackend) in D:\Open Source\osu-framework\osu.Framework\Platform\Sdl\PassthroughGraphicsBackend.cs:line 58
[runtime] 2020-08-02 20:55:35 [error]: at osu.Framework.Platform.Window.Create() in D:\Open Source\osu-framework\osu.Framework\Platform\Window.cs:line 340
[runtime] 2020-08-02 20:55:35 [error]: at osu.Framework.Platform.GameHost.Run(Game game) in D:\Open Source\osu-framework\osu.Framework\Platform\GameHost.cs:line
 538
[runtime] 2020-08-02 20:55:35 [error]: at osu.Desktop.Program.Main(String[] args) in D:\Open Source\osu\osu.Desktop\Program.cs:line 55
```

which is especially weird seeing that it (a) works on `master` and (b) there have been no changes in or around context initialisation as far as I can see... Most likely not going to be able to investigate in detail today, however.

osu.Framework/Platform/Sdl/Sdl2GraphicsBackend.cs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using SDL2;
+
+namespace osu.Framework.Platform.Sdl
+{
+    internal class Sdl2GraphicsBackend : PassthroughGraphicsBackend
```
bdach, line 9: what's with the class structure here? I'm not getting why this stuff was moved out of the passthrough backend, and why this is the only inheritor

osu.Framework/Platform/Sdl/Sdl2GraphicsBackend.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        internal IntPtr SdlWindowHandle;
```
bdach, line 11: probably should be `private`?

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -69,111 +74,112 @@ public bool Visible
...
         {
-            get
-            {
-                if (SdlWindowHandle == IntPtr.Zero)
-                    return position;
-
-                SDL.SDL_GetWindowPosition(SdlWindowHandle, out int x, out int y);
-                return new Point(x, y);
-            }
+            get => SdlWindowHandle == IntPtr.Zero ? position : windowPosition;
```
bdach, line 79: this is very confusing to me. `position` is a plain local, and `windowPosition` is a private property that returns `position` if the handle is zero, otherwise it asks SDL, but *doesn't* store the retrieved value back to `position`. can the private `windowPosition` getter not just be moved here wholesale?

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -69,111 +74,112 @@ public bool Visible
...
+                case WindowState.Fullscreen:
+                    value = w / (float)windowDisplayMode.w;
+                    break;
 
-            cachedScale.Value = w / (float)Size.Width;
-            return cachedScale.Value;
+                case WindowState.FullscreenBorderless:
+                    SDL.SDL_GetDesktopDisplayMode(windowDisplayIndex, out var mode);
+                    value = w / (float)mode.w;
+                    break;
```
bdach, line 129: either SDL API is atrocious, or something is not right here. aren't the checks backwards? shouldn't `SDL_GetDesktopDisplayMode` be used for `Fullscreen`, and `SDL_GetWindowDisplayMode` for `FullscreenBorderless`?

as an aside, [SDL docs](https://wiki.libsdl.org/SDL_GetDesktopDisplayMode) say:

> There's a difference between this function and `SDL_GetCurrentDisplayMode()` when SDL runs fullscreen and has changed the resolution. In that case this function will return the previous native display mode, and not the current display mode.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -69,111 +74,112 @@ public bool Visible
...
-        public WindowState WindowState
+        public override WindowState WindowState
         {
-            get => SdlWindowHandle == IntPtr.Zero ? windowState : windowFlags.ToWindowState();
+            get => SdlWindowHandle == IntPtr.Zero ? initialWindowState : windowFlags.ToWindowState();
             set
             {
-                windowState = value;
-                scheduler.Add(() =>
+                initialWindowState = value;
```
bdach, line 172: it's a minor nit, but I think this set should be within the zero handle guard:

```csharp
if (SdlWindowHandle == IntPtr.Zero)
{
    initialWindowState = value;
    return;
}
```

just reads slightly less weird

osu.Framework/Platform/WindowBackend.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Linq;
+using osu.Framework.Input.StateChanges;
+
+namespace osu.Framework.Platform
+{
+    /// <summary>
+    /// Abstract implementation of <see cref="IWindowBackend"/> that provides default functionality
+    /// for window backend subclasses.
+    /// </summary>
+    public abstract class WindowBackend : IWindowBackend
```
bdach, line 16: I'm not sure I like this too much. If we have an abstract class, what's the point of the interface? I don't think they should both exist; I'd advocate for either removing the interface and staying with the abstract class, or for reverting this and just doing the mundane stuff every time. It's not a huge amount of duplication, but a huge decrease in complexity in my eyes.

osu.Framework/Platform/WindowBackend.cs
```diff
@@ -0,0 +1,90 @@
...
+        protected virtual void OnDragDrop(string file) => DragDrop?.Invoke(file);
+        protected virtual void OnDisplayChanged(Display display) => DisplayChanged?.Invoke(display);
+
+        #endregion
+
+        public virtual void Create() => throw new NotImplementedException();
+
+        public virtual void Run() => throw new NotImplementedException();
+
+        public virtual void Close() => throw new NotImplementedException();
```
bdach, line 88: `abstract` is better if this stays

bdach (changes requested):
only a source read-through, haven't tested personally with the latest changes yet.

@swoolcock have you checked windows after my first try? if not then I will, but I think I saw you mention somewhere that it should work now (although I can't find it now...)

swoolcock:
> @swoolcock have you checked windows after my first try? if not then I will, but I think I saw you mention somewhere that it should work now (although I can't find it now...)

Yes I've tested Windows, worked fine for me.

osu.Framework/Platform/DisplayMode.cs
```diff
@@ -30,14 +31,22 @@ namespace osu.Framework.Platform
...
             Size = size;
             BitsPerPixel = bitsPerPixel;
             RefreshRate = refreshRate;
+            Index = index;
+            DisplayIndex = displayIndex;
         }
 
         public override string ToString() => $"Size: {Size}, BitsPerPixel: {BitsPerPixel}, RefreshRate: {RefreshRate}, Format: {Format}";
+
+        public bool Equals(DisplayMode other) => Index == other.Index && DisplayIndex == other.DisplayIndex;
```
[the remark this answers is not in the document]
swoolcock: This may be left over from a previous work-in-progress that should have been removed with the most recent master merge.  The decision to include just those two fields was discussed and ok'd in Discord a while ago.  I'll remove it for now.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
swoolcock: Yes, the SDL API and documentation is atrocious.  `SDL_GetDesktopDisplayMode` gets the OS-level non-fullscreen resolution of the display, which is what we use for borderless.  `SDL_GetWindowDisplayMode` gets the resolution currently assigned to the window for exclusive fullscreen.

bdach: Gotcha, this definitely warrants some commenting as someone _will_ think this is broken at some point.

osu.Framework/Platform/Sdl/Sdl2GraphicsBackend.cs
swoolcock: I wanted to keep all SDL usage within the `osu.Framework.Platform.Sdl` namespace, as it was still undecided as to whether we were sticking with it.  I have also made window/graphics backends of GLFW and Silk.NET for testing purposes, since this modified structure allows it fairly easily.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
swoolcock: Will do.

osu.Framework/Platform/WindowBackend.cs
swoolcock: In hindsight the interfaces could now be removed, but I would prefer we not lock ourselves to a base abstract class where it determines which members are exposed rather than the interface.  This is what happened with `GameWindow` and I'd rather not repeat that.

osu.Framework/Platform/Sdl/Sdl2GraphicsBackend.cs
bdach: That's fair enough to me.

osu.Framework/Platform/WindowBackend.cs
bdach: Definitely would like more opinions on this one.

osu.Framework/Platform/PassthroughGraphicsBackend.cs
```diff
@@ -10,19 +10,15 @@
...
 
-namespace osu.Framework.Platform.Sdl
+namespace osu.Framework.Platform
 {
     /// <summary>
     /// Implementation of <see cref="IGraphicsBackend"/> that force-loads OpenGL
     /// endpoints into osuTK's bindings.
     /// </summary>
-    public class PassthroughGraphicsBackend : IGraphicsBackend
+    internal abstract class PassthroughGraphicsBackend : IGraphicsBackend
```
smoogipoo, line 20: Is there a reason for this to be `internal`? Avoid `internal` unless something should absolutely never be used outside of o!f - I could foresee someone wanting to implement their own derived `PassthroughGraphicsBackend`.

osu.Framework/Platform/Sdl/Sdl2GraphicsBackend.cs
smoogipoo, line 9: Same here, remove the `internal` - someone may want to implement their own gamehost.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -35,29 +39,30 @@ public class Sdl2WindowBackend : IWindowBackend
...
-        public string Title
+        public override string Title
         {
             get => SdlWindowHandle == IntPtr.Zero ? title : SDL.SDL_GetWindowTitle(SdlWindowHandle);
             set
             {
                 title = value;
-                scheduler.Add(() => SDL.SDL_SetWindowTitle(SdlWindowHandle, $"{value} (SDL)"));
+                if (SdlWindowHandle != IntPtr.Zero)
+                    commandScheduler.Add(() => SDL.SDL_SetWindowTitle(SdlWindowHandle, $"{value} (SDL)"));
```
smoogipoo, line 53: Shouldn't this be more like:
```
commandScheduler.Add(() =>
{
    if (...)
        SDL.SDL_SetWindowTitle(...)
}
```
Also, why's this check only done here and not in some other props like `Visible`, `CursorVisible`, `CursorConfined`, etc...? Is it even required? Shouldn't we be ensuring that `SdlWindowHandle != IntPtr.Zero` inside `Run()`?

osu.Framework/Platform/WindowBackend.cs
smoogipoo: I'm fine with this class existing, so long as it's only used for implementations and not for passing around.

osu.Framework/Platform/WindowBackend.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// for window backend subclasses.
+    /// </summary>
+    public abstract class WindowBackend : IWindowBackend
+    {
+        public virtual string Title { get; set; }
+        public virtual bool Visible { get; set; }
+        public virtual Point Position { get; set; }
+        public virtual Size Size { get; set; }
+        public virtual Size ClientSize { get; set; }
+        public virtual bool CursorVisible { get; set; }
```
smoogipoo, line 23: I'd rather these were abstract also.

smoogipoo (changes requested):

osu.Framework/Platform/PassthroughGraphicsBackend.cs
swoolcock: Will change to public.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
swoolcock: `Title`, `Size`, and `Position` are passed into `SDL_CreateWindow`, so there is no need to queue up a call to `SDL_SetWindowTitle/Size/Position` unless the window already exists and we wish to change them.  This also prevents potential unwanted window events.

Properties such as `Visible`, `CursorVisible`, `CursorConfined` can be queued even if the window does not yet exist, as the commands will not be executed until the `Run()` loop.

osu.Framework/Platform/WindowBackend.cs
swoolcock: Just those four, or all of the properties?

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
smoogipoo: How big is the concern for unwanted window events? Shouldn't they basically be no-ops? I'm just trying to keep things simple.

smoogipoo (approved):
Have tested on Linux and Windows. Gonna assume macOS works since it's your primary platform :p

smoogipoo:
@bdach when you have the opportunity, please test to make sure your previous issue has been resolved :)

bdach:
I've tested and it doesn't crash, but I have a couple of gripes. The minor one (in the context of this PR) is that cycling modes using <kbd>Alt</kbd>+<kbd>Enter</kbd> / <kbd>F11</kbd> seems to not be hooked up yet (might be coming later? not sure) which is annoying for testing.

The major gripes are two scenarios I found with my two-monitor multi-resolution setup. I have two displays (1920x1080 laptop built-in, and 2560x1080 widescreen).

If I change to full screen on the smaller monitor (with the default resolution on), then go back to windowed, move the window to the wider display and go to full screen, it will use the smaller monitor's resolution of 1920x1080, which is not terrible, but could be better.

However, if I change to full screen on the wider monitor, then go back to windowed, move window to smaller display and go to full screen, I get a crash, which is probably not good.

swoolcock:
> If I change to full screen on the smaller monitor (with the default resolution on), then go back to windowed, move the window to the wider display and go to full screen, it will use the smaller monitor's resolution of 1920x1080, which is not terrible, but could be better.
> 
> However, if I change to full screen on the wider monitor, then go back to windowed, move window to smaller display and go to full screen, I get a crash, which is probably not good.
> 
> Cursor confinement also works funky. Sometimes it doesn't stop working until an alt-tab if set to fullscreen.

Do these work as expected using the osuTK backend?

bdach:
Going windowed <-> full screen back and forth on two monitors works fine on osuTK.

Here's a sketch of a potential fix:

```diff
diff --git a/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs b/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
index f77d9a988..449c267ff 100644
--- a/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
+++ b/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
@@ -196,6 +196,9 @@ public override WindowState WindowState
                             break;

                         case WindowState.Fullscreen:
+                            // set window display mode again, just in case if it changed from the last time we were fullscreen.
+                            var fullscreenMode = windowDisplayMode;
+                            SDL.SDL_SetWindowDisplayMode(SdlWindowHandle, ref fullscreenMode);
+
                             SDL.SDL_SetWindowFullscreen(SdlWindowHandle, (uint)SDL.SDL_WindowFlags.SDL_WINDOW_FULLSCREEN);
                             break;

@@ -299,7 +302,12 @@ private SDL.SDL_DisplayMode windowDisplayMode
                     return default;

                 // SDL_GetWindowDisplayMode gets the resolution currently assigned to the window for *exclusive* fullscreen
-                SDL.SDL_GetWindowDisplayMode(SdlWindowHandle, out var mode);
+                if (SDL.SDL_GetWindowDisplayMode(SdlWindowHandle, out var mode) > 0)
+                    return mode;
+
+                // SDL_GetWindowDisplayMode can fail if the window was shown full-screen on a different (especially larger) window before.
+                // if that happens, fall back to first mode for the current display.
+                SDL.SDL_GetDisplayMode(CurrentDisplay.Index, 0, out mode);
                 return mode;
             }
         }
```

Not sure whether that change is acceptable or not, which is why I didn't push that outright. The window mode set might be a better fit when the current display changes.

The code is a little confusing to modify - there's a lot of similar terms (`CurrentDisplayMode`, `windowDisplayMode`) and admittedly it makes my head spin a little. 

Can't repro the confining brokenness. Probably was a knock-on effect from the crash that the above patch is proposing to fix.

swoolcock:
> +                if (SDL.SDL_GetWindowDisplayMode(SdlWindowHandle, out var mode) > 0)

This change looks good, except according to the SDL docs, this should be `>= 0`

> The code is a little confusing to modify - there's a lot of similar terms (`CurrentDisplayMode`, `windowDisplayMode`) and admittedly it makes my head spin a little.

`CurrentDisplayMode` is the framework-side `DisplayMode` defined in `IWindowBackend`.
`windowDisplayMode` returns an `SDL_DisplayMode` retrieved directly from the window handle (hence the `window` in the property names of `windowDisplayIndex`, `windowDisplayBounds`, `windowFlags` and `windowDisplayMode`).

I can't really test the multi-display-resolution thing on macOS as both my external monitors have the same resolutions as my primary display.

bdach:
> This change looks good, except according to the SDL docs, this should be `>= 0`

Hmmm... so this *was* a bug on my part, yeah, but now if the check is done properly, one of the scenario I described regresses - going fullscreen on small display -> windowed -> fullscreen on big display preserves the display mode from the small display leading to letterboxing. (Doesn't happen if you go through borderless, needs to be direct.)

What do you think about invalidating the stored display mode entirely when switching screens? As in doing the `SDL_GetDisplayMode(..., 0, ...)` path every time upon a screen switch, not only when the other call fails?

Tested cycling, works fine here.

swoolcock:
Sounds good, will do in a bit.

swoolcock:
@bdach It's already setting the best possible display mode match for the display when switching to fullscreen.

```csharp
var fullscreenMode = closestDisplayMode(currentDisplayMode);
SDL.SDL_SetWindowDisplayMode(SdlWindowHandle, ref fullscreenMode);

SDL.SDL_SetWindowFullscreen(SdlWindowHandle, (uint)SDL.SDL_WindowFlags.SDL_WINDOW_FULLSCREEN);
```

`currentDisplayMode` is our own type and should only ever be set by the configuration bindable.  It should always be valid.

I'm not sure what else needs to be done in that case.

bdach (approved):
Seems to behave more like I'd expect it to now 👍 Nice work!

smoogipoo (approved):

Issue #3780, opened 2020-08-01
ComputeIsMaskedAway should use Precision methods for floating-point comparisons


frenzibyte:
A drawable that has no size at all and stays at the absolute top-left position of its parent can potentially have `IsMaskedAway` set, while it is expected not to be. (encountered osu!-side where a `SkinnableSound` component added to `PauseOverlay` with zero size gets masked away, occurs in `TestScenePause` only)

<img width="1016" alt="image" src="https://user-images.githubusercontent.com/22781491/89098483-b590d100-d3f0-11ea-8f85-3b1ad2b3b082.png">

The values precisely:

| | `ScreenSpaceDrawQuad.AABBFloat` | `maskingBounds` |
|:--------:|:------------------------------------:|:------------------:|
| X | 846.093689 | 846.09375 |
| Y | 111.328125 | 111.328125 |

PR #3783, opened 2020-08-03
Approximate IsMaskedAway with Precision

smoogipoo:
Resolves https://github.com/ppy/osu/issues/9200
Resolves https://github.com/ppy/osu-framework/issues/3780
Should resolve https://github.com/ppy/osu/issues/9460

I can't seem to be able to write a test that exhibits the issue, so I haven't included one in here.

I'm not sure how to feel about this conceptually. It seems like it may be a bit of a hit-or-miss sort of scenario where the acceptable difference may still be too low in complex hierarchies with autosize/etc.

### DIFF
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -507,7 +507,7 @@ namespace osu.Framework.Graphics
         /// </summary>
         /// <param name="maskingBounds">The <see cref="RectangleF"/> that defines the masking bounds.</param>
         /// <returns>Whether this <see cref="Drawable"/> is currently masked away.</returns>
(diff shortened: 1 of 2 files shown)


bdach:
There's also ppy/osu#9460, which might end up being fixed by this. Haven't checked yet.

smoogipoo:
I also cannot replicate that one (different screen res/osu! res), so I'd appreciate a confirmation.

frenzibyte:
Here's an osu!-side case of #3780 I came up with while I reported the issue, and can confirm with that this resolved it!

framework master: https://streamable.com/hjhbwy
after this PR: https://streamable.com/1d2nhh

PR #3790, opened 2020-08-05
Add more DrawableAudioWrapper transform support

smoogipoo:
Implements:
- `TransformSequence` extensions, allowing things like `drawableTrack.VolumeTo(0).Then().VolumeTo(1, 500);`.
- Generic easings.


frenzibyte:
While at it, could add an interface (`IAudioTransformable` or similar) to allow all the transform extension methods you added applicable to classes like `SkinnableSound`.

smoogipoo:
I've done it through the `IAdjustableAudioComponent` interface, which slots in nicely with my upcoming changes anyway.

PR #3792, opened 2020-08-06
Fix previous sample channels not immediately freed on new playbacks

frenzibyte:
Not sure why it isn't done already, it is a major problem for long samples, as the previous one would not get freed and stay playing without the game being aware of, which is a resource leak to my knowledge.

Untestable as it is an audio code, but I believe it should be very clear.

### DIFF
--- a/osu.Framework/Audio/Sample/SampleChannelBass.cs
+++ b/osu.Framework/Audio/Sample/SampleChannelBass.cs
@@ -66,11 +66,11 @@ namespace osu.Framework.Audio.Sample
                     return;
                 }
 
-                // Remove looping flag from previous channel to prevent endless playback
-                setLoopFlag(false);
+                // Free previous channels as we're creating a new channel for every playback, since old channels
+                // may be overriden when too many other channels are created from the same sample.
+                if (Bass.ChannelIsActive(channel) != PlaybackState.Stopped)
+                    Bass.ChannelStop(channel);
 
-                // We are creating a new channel for every playback, since old channels may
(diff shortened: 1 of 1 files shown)


peppy:
How were you made aware of this? What was the scenario you used to confirm this change is what we want? Also this *could* be tested – we already have good coverage of audio track/sample.

frenzibyte:
@peppy I realized this while investigating through why the tests ensuring correct pause loop playback behaviour didn't fail lazer-side while we hear another pause loop playing while we shouldn't, and the reason was because the loop got played again which causes another channel to be created, and the previous one gets discarded away, but the previous one doesn't get stopped and freed, this is what the PR is aiming to fix.

> Also this could be tested – we already have good coverage of audio track/sample.

Hmm, there is in fact a way to check how many channels does a sample currently have, but I can't get the sample handle as it is and as it should be private, exposing a method to get the current channels from `Sample` would do if that's fine with you?

peppy:
The previous playback *does* already get de-looped though, so it would eventually stop even without your change. Have you confirmed that?

frenzibyte:
Yes but since the sample is long, it would take another play through its duration until it finishes then stops and gets freed

peppy:
I think this is fine for now.

peppy:
This is wrong and needs to be rolled back. It was correctly implemented before. It is not a leak as samples are automatically freed when the channel is reused, or the sample is freed. Channels themselves have a very low overhead so it should not be a huge concern in the first place.

Issue #3803, opened 2020-08-08
Add multi selection TabControl


ReflexLevel0:
Currently, only 1 item in the TabControl can be selected at a time. I am working on solving an issue in ppy/osu project ( https://github.com/ppy/osu/issues/9030) which requires a tab control for which multiple items can be selected at a time. For example, if there are 3 items in the tab control: a, b and c, user should be able to select any combination of those items (see image below for reference).

![image](https://user-images.githubusercontent.com/47285986/89714178-31fd5400-d99d-11ea-81e1-0981162aff8f.png)

bdach:
Are you going to be working on this?

frenzibyte:
In addition to the above, if you're going to work on this, `TabControl` shouldn't support something like this, I would say add a custom implementation displaying an `OsuTabItem` with a boolean result affecting the overall flags enum value (via a current value change or-ing / and-ing), and adding it up to a horizontally-flowing `FillFlowContainer`, as for any other details come ask at the [osu!dev discord server](https://discord.com/invite/ppy).

bdach:
Definitely should be a separate control, not something that's jammed into the base `TabControl`.

ReflexLevel0:
I don't know if I have enough skills to solve this, but I will try (I will notify you if I fail to solve it).

Issue #2423, opened 2019-05-20
Make more user interface components tabbable


HoutarouOreki:
It would be a great quality-of-life improvement if these components would support tab controls:
- Buttons
- Checkboxes
- Dropdowns
- Sliderbars

Keyboard controls are much better than mouse controls in terms of quickly navigating interfaces. Especially when you navigate them a few times per minute, moving your hand back and forth between mouse and keyboard may get tiring.

Example:
![tabsNeeded](https://user-images.githubusercontent.com/16261733/58038466-536fe700-7b30-11e9-873b-e0515948287a.gif)

That's a lot of back and forth movements in a timespan of about 10 seconds. In a perfect scenario a mouse should not be required at all.

UselessToucan:
@HoutarouOreki It would be nice if you provide the code for your example.

HoutarouOreki:
Updated the original post with it @UselessToucan.

nyquillerium:
Going to split this out into a new issue for each component so we can work on them in pieces

peppy:
Moving out of milestone.

PR #3804, opened 2020-08-09
Multi selection TabControl

ReflexLevel0:
Added a field to TabControl which lets the user select multiple tabs if it is set to true. It was suggested that a separate control should be made and that changes should be made to that control and not the original one, but the changes are so small that I don't think that is needed (+the new control acts in a very similar fashion compared to the original TabControl). If you disagree and think that this should be a separate control, please let me know and I will create a separate control.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs
@@ -112,6 +112,11 @@ namespace osu.Framework.Graphics.UserInterface
         public bool SwitchTabOnRemove { get; set; } = true;
 
         /// <summary>
+        /// When true, it allows multiple tabs to be selected at the same moment.
+        /// </summary>
+        public bool CanSelectMultipleTabs = false;
+
+        /// <summary>
         /// Creates an optional overflow dropdown.
         /// When implementing this dropdown make sure:
         /// <list type="bullet">
@@ -345,11 +350,19 @@ namespace osu.Framework.Graphics.UserInterface
             if (AutoSort && tab != null && !tab.IsPresent && !tab.Pinned)
                 performTabSort(tab);
 
-            // Deactivate previously selected tab
-            if (SelectedTab != null && SelectedTab != tab) SelectedTab.Active.Value = false;
+            // Deactivate previously selected tab if multi tab selection is disabled
+            if (SelectedTab != null && SelectedTab != tab && CanSelectMultipleTabs == false)
+                SelectedTab.Active.Value = false;
 
+            // Deactivating previously selected tab and not selecting the new tab if the clicked tab if the same as the last one, and CanSelectMultipleTabs is true
+            if (SelectedTab != null && SelectedTab == tab && CanSelectMultipleTabs)
+            {
+                SelectedTab.Active.Value = false;
+                return;
+            }
+
+            // Selecting new tab
             SelectedTab = tab;
-
             if (SelectedTab != null)
                 SelectedTab.Active.Value = true;
         }
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneTabControl.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneTabControl.cs
@@ -349,6 +349,77 @@ namespace osu.Framework.Tests.Visual.UserInterface
             AddAssert("contained items match added items", () => tabControl.Items.SequenceEqual(items));
         }
 
+        /// <summary>
+        /// Testing to see if <see cref="TabControl{T}.CanSelectMultipleTabs"/> field is working when selecting two tabs
+        /// </summary>
+        [TestCase(true)]
+        [TestCase(false)]
+        public void TestClickTwoTabs(bool canSelectMultipleTabs)
+        {
+            StyledTabControl tabControl = null;
+
+            AddStep("tab control setup", () =>
+            {
+                tabControl = createTestTabControl();
+                tabControl.CanSelectMultipleTabs = canSelectMultipleTabs;
+            });
+
+            AddStep("select first and then second tab", () =>
+            {
+                tabControl.TabMap[TestEnum.Test0].Click();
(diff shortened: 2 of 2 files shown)


osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -111,6 +111,11 @@ public IReadOnlyList<T> Items
         /// </remarks>
         public bool SwitchTabOnRemove { get; set; } = true;
 
+        /// <summary>
+        /// When true, it allows multiple tabs to be selected at the same moment.
```
tybug, line 115:
```suggestion
        /// When true, multiple tabs can be selected at the same time.
```
reads better to me

bdach:
Even on the surface level this approach seems invalid as you haven't changed `SelectedTab` at all. Which tab will `SelectedTab` return if you select multiple?

That is one of the reasons why I suggested to keep this to a separate control.

ReflexLevel0:
SelectedTab should return the last selected tab. If a tab was unselected then null would be returned (or I could set SelectedTab to the first tab whose active value is true when a tab is unselected if you think that would be better). I could also update the comment above the SelectedTab field to make the function of that field more clear. Would that be fine or not?

bdach:
The point I'm trying to make is, a multi-selection tab control should not be exposing a single `SelectedTab` as it's bad API design. It should be exposing all currently selected tabs at once. Doing that in `TabControl` would be a huge breaking change to make in itself and would require major adjustments game-side.

ReflexLevel0:
Yeah, I agree with that. I'll make a separate control.

ReflexLevel0:
Are the changes I made good or not? I have extracted common code in one class and derived 2 classes from that base class. One of those classes has a SelectedTab field which stores the selected tab, while the other class has SelectedTabs field which stores all selected tabs (please ignore unit tests for now, I will fix them later)

osu.Framework/Graphics/UserInterface/MultiSelectTabControl.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <summary>
+        /// All selected tabs
+        /// </summary>
+        public IEnumerable<TabItem<T>> SelectedTabs => TabMap.Values.Where(v => v.Active.Value == true);
+
+        public MultiSelectTabControl()
+        {
+            CanSelectMultipleTabs = true;
+        }
+    }
```
bdach, line 26: this is essentially taking the path of least resistance. I do not agree to having the logic of handling both the single-select and multi-select tab controls both in `BaseTabControl`.

in other words, `SingleSelectTabControl` should implement the entirety of the single selection behaviour, and `MultiSelectTabControl` should implement the entirety of the multiple selection behaviour.

ReflexLevel0: So if I understand correctly, TabControl shouldn't contain the CanSelectMultipleTabs field and the logic that is used for selecting a tab should be in the derived classes?

bdach: that's correct.

smoogipoo:
What's the use-case for a multi-selection tab control? This sounds like a type of control that is completely separate from a tab control.

Joehuu:
The use case is in https://github.com/ppy/osu-framework/issues/3803. Should be linked with this PR.

ReflexLevel0:
In the original TabControl, only 1 item could be selected at a time. In MultiSelectionTabControl multiple tabs can be selected at a time. They share most of the functionality so I saw no purpose in creating an entirely new control. The purpose is to solve issue 9030 in osu project (not osu-framework, but osu - https://github.com/ppy/osu/issues/9030). A TabControl was used for filtering beatmaps there, but to add more beatmap filters I needed a multi selection tab control.

smoogipoo:
Thing is, I'm not sure that should be a tab control. @peppy do you agree with a multi selection tabcontrol?

ReflexLevel0:
#2423 is marked as epic so I think it might be a good idea to make it tabbable (especially if more controls are going to become tabbable in the future, as the issue suggests). But I can make a completely separate control if you think it would be better.

smoogipoo:
#2423 is a completely separate issue. That's about about tabbing between components.

ReflexLevel0:
But couldn't we implement tabbing for MultiSelectTabControl too? For example, if user selected "English" as filter language and he then presses tab, the filter switches to "Chinese" (the next tab), and if he presses space (or some other button) in MultiSelectionTabControl, the tab would get selected (activated).

smoogipoo:
Yes it can be, but again that's a separate issue from tabcontrol / this PR. As I said `MultiSelectTabControl` shouldn't exist - it doesn't make sense as a "tab control".

What the issue requires is a completely separate component, maybe something that wraps a list of checkboxes, that allows both single and multi- selection.

ReflexLevel0:
MultiSelectionTabControl does allow for a single selection. You can select only English for filter language or you can select multiple languages (maybe I didn't understand what you meant to say).

smoogipoo:
Current:
```csharp
class TabControl { }

class MultiSelectionTabControl : TabControl { }
```

New:
```csharp
class TabControl { }

class CheckBoxList { }
```

ReflexLevel0:
Ahh, I think I get it. So the new component would let the user select multiple items in a list, but wouldn't let them tab through those items?

smoogipoo:
What you're describing is focusing items via tab, which is different from the functionality of a tabcontrol (performing actions on ctrl+tab). It can implement that, but will lack tabcontrol functions.

ReflexLevel0:
Ahh, now I get it. Sorry for the misunderstanding. I'll create a separate control.

bdach:
The interface looks sane, but the generic type argument shouldn't be `Checkbox`, it should be the actual model which the checkbox represents (in the desired end use case for example, you'd want to have a `CheckBoxList<ScoreRank>`.

PR #3805, opened 2020-08-10
Prevent menus from consuming Escape keypresses when fading out

voidedWarranties:
Prevents `Menu` from blocking Escape keypresses before the menu is finished fading/animating out (after `updateState` is first called)

For example, opening the osu! song select menu and pressing escape while it is still animating will produce a log message like this:
image
and it won't close as expected

Also caused [this test failure](https://ci.appveyor.com/project/peppy/osu/builds/34566073/tests)


peppy:
Couple of things:
- How are you determining it's caused that test failure? I can't understand how from reading the logs.
- Please add a test for this

Also I'm leaning towards potentially wanting to use `Handle*Input => isOpen` rather than doing it in the event handler itself.

voidedWarranties:
In that PR, I added a context menu to the Multi `RoomsContainer`
The test failed because pressing escape failed to close the screen
I was able to narrow down the problem-causing change to the `OsuContextMenuContainer`, and from there I traced it back to here
I since found that the menu is able to block these presses not due to some "Container loading" but from the menu fading out (as it does in the case of osu!)

I tried using `HandleNonPositionalInput` like this:
image
but the keypress still gets blocked when the menu is fading out, so I kept it as-is for now

osu.Framework.Tests/Visual/UserInterface/TestSceneClosableMenu.cs
```diff
@@ -52,5 +54,38 @@ public void TestClickItemClosesMenus()
...
+        [Test]
+        public void TestMenuIgnoresEscapeWhenClosed()
+        {
+            AnimatedMenu menu = null;
+
+            AddStep("find menu", () => menu = (AnimatedMenu)Menus.GetSubMenu(0));
+            AddStep("press escape", () => InputManager.Key(Key.Escape));
+            AddAssert("press handled", () => menu.PressBlocked);
+            AddStep("press escape again", () => InputManager.Key(Key.Escape));
+            AddAssert("press not handled", () => !menu.PressBlocked);
```
bdach, line 67: `HandleNonPositionalInput` would work if you reset the value of `PressBlocked` to `false` between these checks. as-is the test fails on the leftover `true` from the first assertion, as having `HandleNonPositionalInput => false` prevents the input from ever reaching the `OnKeyDown` handler.

voidedWarranties: `OnKeyDown` is still able to process these events even when `HandleNonPositionalInput` is false, am I missing something?

bdach: I have checked this PR earlier today and applied the following patch:

```difff
diff --git a/osu.Framework.Tests/Visual/UserInterface/TestSceneClosableMenu.cs b/osu.Framework.Tests/Visual/UserInterface/TestSceneClosableMenu.cs
index 0e57d99ee..47a2f8f12 100644
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneClosableMenu.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneClosableMenu.cs
@@ -63,6 +63,7 @@ public void TestMenuIgnoresEscapeWhenClosed()
             AddStep("find menu", () => menu = (AnimatedMenu)Menus.GetSubMenu(0));
             AddStep("press escape", () => InputManager.Key(Key.Escape));
             AddAssert("press handled", () => menu.PressBlocked);
+            AddStep("reset flag", () => menu.PressBlocked = false);
             AddStep("press escape again", () => InputManager.Key(Key.Escape));
             AddAssert("press not handled", () => !menu.PressBlocked);
         }
diff --git a/osu.Framework/Graphics/UserInterface/Menu.cs b/osu.Framework/Graphics/UserInterface/Menu.cs
index 333cff8a1..457cfb48a 100644
--- a/osu.Framework/Graphics/UserInterface/Menu.cs
+++ b/osu.Framework/Graphics/UserInterface/Menu.cs
@@ -67,6 +67,8 @@ public abstract class Menu : CompositeDrawable, IStateful<MenuState>
 
         private readonly Container<Menu> submenuContainer;
 
+        public override bool HandleNonPositionalInput => State == MenuState.Open;
+
         /// <summary>
         /// Constructs a menu.
         /// </summary>
@@ -480,7 +482,7 @@ private void menuItemHovered(DrawableMenuItem item)
 
         protected override bool OnKeyDown(KeyDownEvent e)
         {
-            if (state == MenuState.Open && e.Key == Key.Escape && !TopLevelMenu)
+            if (e.Key == Key.Escape && !TopLevelMenu)
             {
                 Close();
                 return true;
```

the test passes on my machine, and when I breakpoint on `AnimatedMenu.OnKeyDown` after running the test with debug, I get one hit before the flag is reset and none after. please check whether that is also the case for you.

voidedWarranties: The test passes on my machine, but slowing down the test and inspecting the menu in the draw visualizer shows the menu has already faded out completely by the second press
If you set the fade duration to a slightly larger number to compensate for the extra step, the test fails and the handler consumes both events

bdach: then the test case is badly written. it should not flake due to timing issues like that.

voidedWarranties: A solution to this appears to be fading to translucent so that the menu is logically "closed" but can still receive keypresses (if it is working incorrectly)
But, I feel something about the unexplained `0.5f` in `AnimateClose` would make the test somewhat confusing, is foregoing the `AnimateOpen` and `AnimateClose` overrides and instead forcing it to fade externally via `menu.FadeTo(0.5f)` better?

bdach: I think making the fade last long enough (a few secs or so should be enough) to the point at which it is no longer possible for the tempo of test step execution to affect whether or not the assert fires at the correct moment is a simpler resolution, possibly with an inline comment.

voidedWarranties: Alright, but is there a reason `HandleNonPositionalInput` appears to have no effect?

bdach: guessing it might be necessary to also override `PropagateNonPositionalInputSubTree`. haven't checked, however.

voidedWarranties: don't think `PropagateNonPositionalInputSubTree` is needed here, as it didn't have much effect in my testing, not completely sure though
another menu, `OsuDropdownMenu`, also has the same `HandleNonPositionalInput` check [here](https://github.com/ppy/osu/blob/fa14141bd4a97b72a522bd9dca31fd803a273167/osu.Game/Graphics/UserInterface/OsuDropdown.cs#L55), but pressing escape twice in quick succession in a place like a settings dropdown makes 2 log messages indicating the press was consumed twice, looks like the same thing is happening there too

bdach:
so I looked at this myself, again, and I'm pretty sure that if we don't want the state check inside `OnKeyDown`, this seems to be the proper solution:

```diff
diff --git a/osu.Framework/Graphics/UserInterface/Menu.cs b/osu.Framework/Graphics/UserInterface/Menu.cs
index 333cff8a1..ccff7d96b 100644
--- a/osu.Framework/Graphics/UserInterface/Menu.cs
+++ b/osu.Framework/Graphics/UserInterface/Menu.cs
@@ -67,6 +67,9 @@ public abstract class Menu : CompositeDrawable, IStateful<MenuState>
 
         private readonly Container<Menu> submenuContainer;
 
+        protected override bool ShouldBeConsideredForInput(Drawable child) => State == MenuState.Open;
+        public override bool PropagateNonPositionalInputSubTree => State == MenuState.Open;
+
         /// <summary>
         /// Constructs a menu.
         /// </summary>
@@ -238,6 +241,7 @@ private void updateState()
             {
                 case MenuState.Closed:
                     AnimateClose();
+                    GetContainingInputManager()?.ChangeFocus(null);
                     break;
 
                 case MenuState.Open:
@@ -480,7 +484,7 @@ private void menuItemHovered(DrawableMenuItem item)
 
         protected override bool OnKeyDown(KeyDownEvent e)
         {
-            if (state == MenuState.Open && e.Key == Key.Escape && !TopLevelMenu)
+            if (e.Key == Key.Escape && !TopLevelMenu)
             {
                 Close();
                 return true;
```

`PropagateNonPositionalInputSubtree` was partly correct, in that it prevents all children of the menu to be included in the input queue. 

`ShouldBeConsideredForInput` is caveat number 1. context menus are positioned inside a `ContextMenuContainer`, which is a `CursorEffectContainer`, which is a `Container`, which is a `CompositeDrawable`. this is what `CompositeDrawable.BuildNonPositionalInputQueue` looks like:

https://github.com/ppy/osu-framework/blob/14ee5802cb1e3d6ced37d1511ea55766f2a59e05/osu.Framework/Graphics/Containers/CompositeDrawable.cs#L1291-L1303

therefore, even if the menu itself returns `false` from its `BuildNonPositionalInputQueue` (due to returning `PropagateNonPositionalInputSubtree = false`), not overriding `ShouldBeConsideredForInput` will mark it back again for inclusion in the queue in its parent. by default `ShouldBeConsideredForInput` is true.

even after dispatching that, there is caveat number 2, which is the focus change. below is a fragment of `InputManager.buildNonPositionalInputQueue()`:

https://github.com/ppy/osu-framework/blob/c79e4ddf8bdb03783f6b5ec1d2fa23b399e0f90b/osu.Framework/Input/InputManager.cs#L475-L479

turns out the menu doesn't release focus when it's marked as closed, because it's still alive, present and in the hierarchy (at least until the fade out is complete). releasing focus immediately upon close fixes that part, although I'm unsure it will be correct from a UI standpoint.

voidedWarranties:
thanks for looking further into it,
the only necessary changes that I found are:
1. focus changes to `parentMenu` rather than `null`, as with `null` context menus close immediately after opening and submenus close when hovering, and
2. adding a check before attempting to change focus, as the game will crash if `updateState` is called during `OnFocusLost` (but I don't know whether I'm checking for the right thing, so feedback is needed there)

osu.Framework/Graphics/UserInterface/Menu.cs
```diff
@@ -238,6 +238,9 @@ private void updateState()
             {
                 case MenuState.Closed:
                     AnimateClose();
+
+                    if (GetContainingInputManager()?.FocusedDrawable == this)
```
bdach, line 242: `Drawable.HasFocus` exists

voidedWarranties: oops, missed that

peppy:
Test still passes without the two suggestions from @bdach, but only:

```
        public override bool HandleNonPositionalInput => State == MenuState.Open;
```

and the `ChangeFocus` you added.

This means the test is probably not correctly covering the changes.

peppy:
I actually don't follow the logic of why `PropagateNonPositionalInputSubtree` or otherwise is required, since we are just blocking the local key handling in the class itself, no?

bdach:
that's a fair point, I think I actually removed that one and had it still work but then got confused.

osu.Framework/Graphics/UserInterface/Menu.cs
```diff
@@ -478,6 +481,10 @@ private void menuItemHovered(DrawableMenuItem item)
             }
         }
 
+        protected override bool ShouldBeConsideredForInput(Drawable child) => base.ShouldBeConsideredForInput(child) && State == MenuState.Open;
```
smoogipoo, line 484: This isn't required if `PropagateNonPositionalInputSubTree` is overridden with the same condition.

PR #2870, opened 2019-09-30
Adjust vertex batching for modern-day use cases

smoogipoo:
1. From my experimentation, o!f almost never goes above 50 vertices per draw, and every single draw uses a new VBO. osu! is even lower at <30 vertices per draw. So 50 is a good value but I picked 100 vertices for a "safe" value.
2. I found that using 8 as the minimum amount of children resulted in a slightly reduced number of VBO binds:

| old | new |
| --- | ---- |
| image | image |

The next step I believe is to implement mapped buffers which we can write to unsynchronised, and re-use the same VBO for subsequent `Draw()` calls.

This improves performance a tad (centipede is now mostly update limited for me), and reduces memory usage by at least 2x. Here's some comparisons:
Left = master, right = this branch

image

image

image

image

image

image

image

PR #3813, opened 2020-08-12
Implement a mechanism for freeing unused static VBOs

smoogipoo:
I'm half-reluctantly opening this up since the mechanism has issues, so I'm interested in people's opinions.

The osu!lazer game-wide overlays will permanently hold VBOs. Scrolling down the changelog overlay can see QuadVertexBuffer memory usage skyrocket with no collection.
This PR "registers" the VBOs with the `GLWrapper`, which periodically checks whether any VBOs have remained unused for some number of frames, and frees them. Right now, this check happens every 300 frames, and any VBOs unused in the last 300 frames will be freed (i.e. at worst, VBOs can remain unused for up to 599 frames before they're freed).

It's a bit of a primitive and greedy solution which I see as a stop-gap measure for implementing a gc-type-system to do more fine-tuned per-VBO tracking.

Caveats:
- Instantaneous Gen2 pressure has increased. This is mainly due to using the default Imagesharp allocator which isn't cut out for this sort of task with the default pool size, but it's also partly also due to the size of our batches.
    - Our VBOs generally have a fixed size of 240KB. Just by reducing this to 24KB (10 quads per composite), Centipede goes from 2.3-2.7GB to 1.7GB. Changelog overlay goes from 1.5GB to 500MB. As I mentioned above, doing these sort of tweaks is probably more the task of a gc-type-system that can do more analysis at runtime.
    - Gen2 recollects, unlike current master. It's probably better that this happens rather than holding onto a slightly smaller chunk of memory endlessly.
- Slight performance decrease due to actually running the check. I've tried to make this as optimal as possible but I've measured up to 0.02ms lower performance.
- Performance decrease when scrolling up/down changelog, due to constantly re-creating VBOs. I'd say this is to be expected, but can also probably benefit in the future by rolling our own pools/gc-type-system.

Comparisons:

| description | master | this PR |
| --- | --- | --- |
| world.execute(me) results* | before-1 before-2 | after-1 after-2 |
| world.execute(me) after exit to menu + clear caches | before-3 before-4 | after-3 after-4 |
| centipede results** | before1 | after1 |
| changelog*** | before-7 before-8 | after-7 after-8 |

\* Gameplay was allowed to complete in full (no fast forward). Taken with a custom skin which results in less VBO allocs.
\*\* Taken with the osu!lazer default skin.
\*\*\* Taken after scrolling up-and-down multiple times until Gen2 stabilised.

Results may differ slightly since I've bumped up the checking interval from 300 to effectively 600.


peppy:
This is a bit of a tangent from this PR, but just to make sure we're on the same page, the end-goal should be to avoid needing so having so many separate batches in the first place by reducing shader/texture state changes, right? And therefore not having so many per-composite-VBOs. The way we're handling VBOs is really shoddy right now due to the amount of draw call interruptions (and the very arbitrary 8-child-drawable-makes-new-VBO rule) and that should be an ongoing focus, IMO.

Another common technique aren't doing right now but should probably be is decoupling VBOs from drawables. ie. rather than having fixed size VBOs per `CompositeDrawable`, the VBOs would be pooled themselves (not constantly alloc'd dealloc'd) at a central manager and multiple drawables would share the same VBOs and write to different ranges in them. though maybe that's what you mean by making a GC? more like an allocation manager?

I think right now we're giving far too much priority to not wanting to potentially rewrite VBO content (due to vertex count / structure changes) which is why we're hitting memory issues in the first place where they shouldn't exist.

peppy:
- Centipede screenshots are showing different stats, is that intended?
- Changelog is using double the memory now in the stats display, is that correct?

smoogipoo:
> the end-goal should be to avoid needing so having so many separate batches

Yeah, that is the eventual goal.

> decoupling VBOs from drawables

In terms of wording, yes, it would be an "allocation manager". Though what I envision is a tiered + promotion approach similar to a GC which depends on how static the vertices are. This is because we also have to minimise vertex uploads along with draw calls which can be done by grouping together vertices that change at similar rates.

> too much priority to not wanting to potentially rewrite VBO content

It's very easy to say this, but it's very hard to prove. While 8 drawables triggering VBO creation isn't a fundamental constant, it's based on empirical evidence I arrived at in https://github.com/ppy/osu-framework/pull/2870 when optimising for both performance and memory usage. You can rewrite too much and cause too many VBO uploads or rewrite too little and cause too many draw calls.

> Changelog is using double the memory now in the stats display

I mentioned that, yes. Please read the part of my post when I mentioned using Imagesharp's allocator (and the sub-points).

> Centipede screenshots are showing different stats

Whoops, screenshots got fumbled up, will fix.

peppy:
I did read your OP, but it's a bit unclear on the gen2 part (it mentions numbers which I'm not sure are before or after this change, none of which align with the screenshots).

For changelog / other edge cases, does that means that on a system with low memory, it will  potentially OOM where it didn't used to? Is that something we need to fix locally along with this change, if so?

smoogipoo:
If by OOM you mean result in a crash, then no. But it will cause a more immediate and productive gen2 collection.

With these changes, when changelog uses up 1.4GB gen2, a clear caches or otherwise a GC (both achieve the same thing) brings it back down to 200MB, far lower than the 700MB lifetime usage on master. A system with low memory will encounter an early gen2 and won't have their game constantly trying to gen2 forever more.

As for the mentioning of the numbers, that's a step above these changes. That's if we were to change the VBO size from 100 to 10 in https://github.com/ppy/osu-framework/blob/11b8f113e38aca8e6b170692c58e7d02ac79b318/osu.Framework/Graphics/Containers/CompositeDrawable_DrawNode.cs#L174

peppy:
Right. Are you planning to make that change in a separate PR? I agree that 100 is likely far too high given the current usage pattern.

smoogipoo:
Wasn't immediately planning to since I also don't have a clear picture of all of our usage patterns, and that value was also touched in https://github.com/ppy/osu-framework/pull/2870...

peppy:
I'd argue given our current memory situation we'd probably see better overall performance just by reducing memory overhead (most of the users reporting issues are 99% GC related, while draw performance is already pretty-good), and maybe we should give the reduction a roughly-blind try.

smoogipoo:
I'm not so hasty. I'd argue our draw memory usage is quite good already - the issues people have aren't because we're allocating some 1GB extra of collectible (after this PR) gen2 during gameplay, they're because we're allocating a ton of drawables. Pooling will resolve that. In-fact pooling will resolve much the same issue as the proposed change, at least for gameplay.

I'm not going to make such a low-level change that has so much potential to regress in hidden ways when there are other pressing issues.

peppy:
Hesitant to hinge everything on pooling. I would hope that the gl memory usage is lean enough to not be an issue in *any* usage scenario. At least the "waste" portion, where VBOs are much larger than they are needed to be. Undershooting is much better than overshooting (as it seems we may be doing currently).

smoogipoo:
How is undershooting better than overshooting? What you're suggesting is to cripple the performance of _every single system_ out there regardless of RAM, for the sole purpose of _maybe_ some low tier mobile device from 4 years ago or a PC with <=4GB RAM not experiencing as many gen2 stutters. That's it. Those devices aren't going to crash because of "ephemeral" gen2 allocs.

I want to stress that what you're gaining here is at most going from 2.5GB RAM (pre-GC) to 1.5GB (pre-GC) during gameplay. After GC they both equalise (or close to equalise). If you think that is in any way reasonable for gameplay, then I don't know what to say.

I'm not interested in going anywhere near that sort of change without either:
1. Multi-day work investigating the performance impact in multiple scenarios and ensuring that performance is not impacted, which the original PR would suggest would happen - osu! regularly averaged 30 vertices per batch, framework 50
2. A more general approach to resolving these issues, like the memory allocator/gc/whatever that was previously proposed.

peppy:
Depends on if that GC is happening during gameplay or not, of course. I'm specifically focused on reducing RAM not to save on RAM, but to reduce GC load, since I've seen that always show up when profiling (dottrace timeline).

For another day, for sure.

peppy:
Going to merge this, in testing it solve a lot of the edge case scenarios of bloated unmanaged memory and I can't notice the overhead.

PR #3818, opened 2020-08-13
Remember window position between sessions (SDL)

swoolcock:
Adds some minor missing features to the `DesktopWindow` class without making significant changes to the SDL window backend.  The work in this PR will be required regardless of the windowing toolkit we end up using.

* Last window display and relative position are now remembered between sessions
* Window mode is now remembered between sessions

Alternative to part of #3781.  If we end up sticking with SDL, much of the code in that PR can be rebased/reused.

Needs testing on Linux.

### DIFF
--- a/osu.Framework/Platform/DesktopWindow.cs
+++ b/osu.Framework/Platform/DesktopWindow.cs
@@ -141,12 +141,9 @@ namespace osu.Framework.Platform
 
         private void updateWindowPositionConfig()
         {
-            if (WindowState.Value == Platform.WindowState.Normal)
-            {
-                var relativePosition = RelativePosition;
-                windowPositionX.Value = relativePosition.X;
-                windowPositionY.Value = relativePosition.Y;
-            }
+            var relativePosition = RelativePosition;
+            windowPositionX.Value = relativePosition.X;
+            windowPositionY.Value = relativePosition.Y;
         }
 
         private void confineMouseModeChanged(ValueChangedEvent<ConfineMouseMode> args)
--- a/osu.Framework/Platform/DesktopWindow.cs
+++ b/osu.Framework/Platform/DesktopWindow.cs
@@ -1,15 +1,20 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
+using System.Drawing;
+using System.Linq;
 using osu.Framework.Bindables;
 using osu.Framework.Configuration;
 using osu.Framework.Input;
 using osu.Framework.Platform.Sdl;
+using osuTK;
 
 namespace osu.Framework.Platform
 {
     public class DesktopWindow : Window
     {
+        private readonly BindableSize sizeFullscreen = new BindableSize();
         private readonly BindableSize sizeWindowed = new BindableSize();
         private readonly BindableDouble windowPositionX = new BindableDouble();
         private readonly BindableDouble windowPositionY = new BindableDouble();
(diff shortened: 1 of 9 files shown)


osu.Framework/Platform/DesktopWindow.cs
```diff
@@ -1,19 +1,54 @@
...
     {
+        private readonly BindableSize sizeFullscreen = new BindableSize();
         private readonly BindableSize sizeWindowed = new BindableSize();
+        private readonly BindableDouble windowPositionX = new BindableDouble();
+        private readonly BindableDouble windowPositionY = new BindableDouble();
+        private readonly Bindable<DisplayIndex> windowDisplayIndex = new Bindable<DisplayIndex>();
 
         public readonly Bindable<ConfineMouseMode> ConfineMouseMode = new Bindable<ConfineMouseMode>();
 
+        protected Vector2 RelativePosition
```
bdach, line 25: I'm not sure I understand what this vec2 is supposed to mean. Why does the window size factor in at all during the calculation of this?

From what I'm understanding it's supposed to be a [0,1]-normalised value, but why not then just convert the absolute x,y coords of the top left corner of the window to that by dividing by width/height respectively? Or is this supposed to be something like percentage-based padding for the window (so setting to `(0.5, 0.5)` centers the window on the screen)?

In general xmldoc for this would be good.

osu.Framework/Platform/DesktopWindow.cs
```diff
@@ -27,24 +62,84 @@ public override void SetupWindow(FrameworkConfigManager config)
...
+
+        private void onMoved(Point point)
+        {
+            if (WindowState.Value == Platform.WindowState.Normal)
+                updateWindowPositionConfig();
+        }
+
+        private void updateWindowPositionConfig()
+        {
+            if (WindowState.Value == Platform.WindowState.Normal)
```
bdach, line 137: A bit weird to me how this is checked twice in both the callers of this and locally here.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -219,6 +228,7 @@ public DisplayMode CurrentDisplayMode
                 SDL.SDL_GetCurrentDisplayMode(currentDisplayIndex, out var mode);
                 return displayModeFromSDL(mode);
             }
+            set { }
```
bdach, line 231: I know this is a transitionary period but I'd still drop in a `// TODO` at least.

osu.Framework/Platform/Window.cs
```diff
@@ -485,32 +523,32 @@ public int Height
...
             {
-                Position.Value = value.Location;
-                Size.Value = new Size((int)(value.Width / Scale), (int)(value.Height / Scale));
             }
         }
 
         Size INativeWindow.ClientSize
         {
-            get => new Size((int)(Size.Value.Width * Scale), (int)(Size.Value.Height * Scale));
-            set => Size.Value = new Size((int)(value.Width / Scale), (int)(value.Height / Scale));
```
bdach, line 499: Why are these setter implementations going away?

osu.Framework/Platform/DesktopWindow.cs
swoolcock: You are correct in that `(0.5, 0.5)` centres the window; this is how the current osuTK implementation works.  I'll add some xmldoc for clarity..

osu.Framework/Platform/Window.cs
swoolcock: 1) We don't currently use them.
2) The properties only exist for backward compatibility.
3) The implementation of `IWindowBackend` now maps `ClientSize` directly to the framebuffer size, so "setting" it first requires a calculation of the scale, which I'm trying to avoid.

osu.Framework/Platform/DesktopWindow.cs
swoolcock: Must be left over from testing, will remove.

bdach (approved):
Looks ok to me. Probably wise have a second pair of eyes look this through, though.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -111,6 +111,15 @@ public Size Size
...
+        public Size ClientSize
+        {
+            get
+            {
+                SDL.SDL_GL_GetDrawableSize(SdlWindowHandle, out var w, out var h);
+                return new Size(w, h);
+            }
+        }
+
         public float Scale => validateScale();
```
smoogipoo, line 123: Is there a reason for not removing Scale, now that it's baked into ClientSize?

smoogipoo: I guess because it's still included in input events?

swoolcock: Yeah, it's still required for converting positional events from the windowing system to framebuffer space.  It's probably no longer necessary to be exposed in `Window` though, only `IWindowBackend`.

smoogipoo: Maybe see if that can be done? Would feel much better imo.

smoogipoo (approved):

Issue #3817, opened 2020-08-12
Transforms affecting overlapping target members do not combine correctly


bdach:
Originally reported by @LumpBloom7 [on discord](https://discord.com/channels/188630481301012481/188630652340404224/743152124757803049).

To explain the title: applying transforms simultaneously on i.e. `Y` and `Position` does not work properly due to the per-member tracker separation done in #3779.

Minimal reproducer example is:

```csharp
        [Test]
        public void TestMultipleTransformTargets()
        {
            boxTest(box =>
            {
                box.Delay(500).MoveToOffset(new Vector2(0, 0.25f), 500);
                box.MoveToY(0.5f, 250);
            });

            checkAtTime(double.MinValue, box => box.Y == 0);
            checkAtTime(0, box => box.Y == 0);
            checkAtTime(250, box => box.Y == 0.5f);
            checkAtTime(750, box => box.Y == 0.375f);
            checkAtTime(1000, box => box.Y == 0.25f);
            checkAtTime(1500, box => box.Y == 0.25f);
            checkAtTime(750, box => box.Y == 0.375f);
        }
```

(stick it in `TestSceneTransformRewinding` to see the failure - it should fail at the last step; you can also run the first two steps manually in test browser to see it break in real time on rewind). Debug shows the following:

![2020-08-12-221106_705x478_scrot](https://user-images.githubusercontent.com/20418176/90062966-ddcdd900-dce8-11ea-9c99-e20912a93386.png)

<hr>

Opening mostly for paper trail. I might circle back to look at how to fix this properly but as I'm reasonably certain this will not be related to the game-side follow point regression that I want to look into first and foremost so I don't want this to get lost.

peppy:
I knew there was going to be *something* which relied on transforms all being in one list. Now it's time to brainstorm how to resolve this one. The simplest solution would be to combine, reorder, then reapply transforms to be rewinded (only changes performance characteristics during rewind) so I might go with that initially unless anyone has a better idea?

bdach:
I'm not sure that any solution that doesn't combine `Position`, `X` and `Y` trackers entirely is valid. Don't have a confirmed counterexample to prove it but I think forward application of transforms can also suffer from bugs resulting from processing the separated trackers in creation order.

peppy:
Yeah, may be the case.

[...]

PR #3822, opened 2020-08-14
Fix transforms operating on same properties being applied out of order

peppy:
This basically brings back logic that was removed in the previous PR (#3779) as it is still required. The tracker class will now potentially be tracking more than one target member in the case they touch the same underlying property. Less optimal, but this is the best we can do (and should still be fine for most use cases).

For reference, I brought the `xAt(targetMember)` logic back from https://raw.githubusercontent.com/ppy/osu-framework/d0fc93793e7d4aac79d34dbdec7190a0951bdc16/osu.Framework/Graphics/Transforms/Transformable.cs, which can be cross-checked during review.

- Closes https://github.com/ppy/osu-framework/issues/3817

### DIFF
--- a/osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
+++ b/osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
@@ -60,7 +60,7 @@ namespace osu.Framework.Graphics.Transforms
             {
                 resetLastAppliedIndex();
 
-                var appliedToEndReverts = new List<string>();
+                bool appliedToEndRevert = false;
 
                 // Under the case that completed transforms are not removed, reversing the clock is permitted.
                 // We need to first look back through all the transforms and apply the start values of the ones that were previously
@@ -82,14 +82,14 @@ namespace osu.Framework.Graphics.Transforms
                         // we are in the middle of this transform, so we want to mark as not-completely-applied.
                         // note that we should only do this for the last transform to avoid incorrect application order.
                         // the actual application will be in the main loop below now that AppliedToEnd is false.
-                        if (!appliedToEndReverts.Contains(t.TargetMember))
+                        if (!appliedToEndRevert)
                         {
                             if (time < t.EndTime)
                                 t.AppliedToEnd = false;
                             else
                                 t.Apply(t.EndTime);
 
-                            appliedToEndReverts.Add(t.TargetMember);
+                            appliedToEndRevert = true;
                         }
                     }
                     else
--- a/osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
+++ b/osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
@@ -12,17 +12,17 @@ namespace osu.Framework.Graphics.Transforms
     /// <summary>
     /// Tracks the lifetime of transforms for one specified target member.
     /// </summary>
-    internal class TargetMemberTransformTracker
+    internal class TargetGroupingTransformTracker
     {
         /// <summary>
-        /// A list of <see cref="Transform"/>s associated with the <see cref="TargetMember"/>.
+        /// A list of <see cref="Transform"/>s associated with the <see cref="TargetGrouping"/>.
         /// </summary>
         public IEnumerable<Transform> Transforms => transforms;
 
         /// <summary>
         /// The member this instance is tracking.
         /// </summary>
-        public readonly string TargetMember;
+        public readonly string TargetGrouping;
 
         private readonly SortedList<Transform> transforms = new SortedList<Transform>(Transform.COMPARER);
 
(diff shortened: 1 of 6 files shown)


osu.Framework/Graphics/Transforms/Transform.cs
```diff
@@ -36,6 +36,11 @@ public abstract class Transform
 
         public abstract string TargetMember { get; }
 
+        /// <summary>
+        /// A grouping specification that should match any other <see cref="Transform"/>s which affect the same properties.
```
bdach, line 40: I'm not sure how to interpret this xmldoc. How about something like

```csharp
/// <summary>
/// The name of the grouping this <see cref="Transform"/> belongs to.
/// Defaults to <see cref="TargetMember"/>.
/// </summary>
/// <remarks>
/// Transforms in a single group affect the same property (or properties) of a <see cref="Transformable"/>.
/// It is assumed that transforms in different groups are independent from each other
/// in that they affect different properties, and therefore they can be applied independently
/// in any order without affecting the end result.
/// </remarks>
```

osu.Framework/Graphics/Transforms/Transformable.cs
```diff
@@ -79,20 +79,29 @@ protected void UpdateTransforms()
...
-                if (t.TargetMember == targetMember)
+                if (t.TargetMembers.Contains(targetMember))
                     return t;
             }
 
-            if (!createIfNotExisting) return null;
+            return null;
+        }
+
+        private TargetGroupingTransformTracker getTrackerForGrouping(string targetGrouping)
```
bdach, line 95: This will get or create a tracker, contrary to `getTrackerFor()`, which can only get; would probably reflect that in method name.

```suggestion
        private TargetGroupingTransformTracker getOrCreateTrackerForGrouping(string targetGrouping)
```
(+ usages)

osu.Framework/Graphics/Transforms/Transformable.cs
```diff
@@ -118,7 +127,7 @@ private void updateTransforms(double time, bool forceRewindReprocess = false)
         /// <param name="toRemove">The <see cref="Transform"/> to remove.</param>
         public void RemoveTransform(Transform toRemove)
         {
-            getTrackerFor(toRemove.TargetMember)?.RemoveTransform(toRemove);
+            getTrackerForGrouping(toRemove.TargetGrouping)?.RemoveTransform(toRemove);
```
bdach, line 130: This can allocate a tracker if there wasn't one already (previous code had `getTrackerFor` call with `createIfNotExisting = false` by default)

osu.Framework/Graphics/Transforms/Transformable.cs
peppy: yeah, was considering this but private method so didn't think much of it. sounds good.

osu.Framework/Graphics/Transforms/Transformable.cs
peppy: don't see this as a huge deal, do you think it needs to be protected? very rare case that you'd be trying to remove a transform that doesn't exist.

peppy: I've restored the parameter back to resolve this and the method naming in one go.

bdach: Suppose not, mostly just spotted it as a potential gotcha, but this shouldn't ever really be relevant save for egregious programmer error.

bdach:
I just ran the test scene again to be sure, and in the real-time view (achieved by just running first two steps of the new test here) something looks off - [see video](https://drive.google.com/file/d/1HEgoYNDNYJYACF5z90jwxWAOldhP5Hq3/view). There's a rapid jump right when the clock changes direction to backward.

Unsure whether this is caused by changes here, I could see a possibility it isn't.

osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
```diff
@@ -75,14 +82,14 @@ public void UpdateTransforms(in double time, bool rewinding)
                         // we are in the middle of this transform, so we want to mark as not-completely-applied.
                         // note that we should only do this for the last transform to avoid incorrect application order.
                         // the actual application will be in the main loop below now that AppliedToEnd is false.
-                        if (!appliedToEndRevert)
+                        if (!appliedToEndReverts.Contains(t.TargetMember))
```
bdach, line 85: It seems that the previous version without the per-member split is correct here (it fixes the problem from the [comment above](https://github.com/ppy/osu-framework/pull/3822#issuecomment-673927511)). The reason for the jump on the video is the `Position` and `Y` transforms both reverting to their `EndTime`, in backward temporal order.

I think the other `TargetMember` checks are still fine, although now I'm not all that sure...

peppy: I do wonder if that was the case before the original PR too, then?

bdach: I just checked 2020.730.1 (last release before the original PR), and yep, so it was.

bdach:
Think this should be ready to go now but maybe better to triple-check (hi @smoogipoo) this as I haven't touched transform code at all up until this week.

PR #3823, opened 2020-08-14
Reconfigure bass for lower latency on linux

smoogipoo:
Resolves https://github.com/ppy/osu/issues/3800

This should hopefully not cause any issues, because the buffer length remains unchanged from the previous value. This value is also ignored on Windows.
With that being said this needs confirmation on macOS and iOS, and possibly Android. On macOS/iOS specifically, buffer length is [documented](http://www.un4seen.com/doc/#bass/BASS_CONFIG_DEV_BUFFER.html) as being twice the device period, ignoring our custom value and potentially behaving badly after this change.

The other effect is CPU% increase from ~0.5% to ~1.5%.

Also I'm not sure what the removed lines were doing outside of the main `InitBass()` function, so I've moved them in there. I'm also further confused as to what `PlaybackBufferLength` does since I can't seem to find docs for it, but I've left it in for the time being (doesn't seem to have any effect with or without on my linux box).

### DIFF
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -280,9 +280,6 @@ namespace osu.Framework.Audio
             //we have successfully initialised a new device.
             UpdateDevice(deviceIndex);
 
-            Bass.PlaybackBufferLength = 100;
-            Bass.UpdatePeriod = 5;
-
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


peppy:
Tested on macOS, iOS (2016 iPhone SE and iPad Pro) with no discernible difference. All tests done on debug builds (framework also debug) to ensure no issues at higher than normal loads.

PR #3824, opened 2020-08-14
Fix LargeTextureStore never purging textures

peppy:
Regressed with the introduction of wrap modes (changing the key but not using it in all locations).

Also takes the load off the finalizer for actually disposing the gl-side texture.


bdach:
The effects of this are visible game-side in song select for one example, right?

The change looks good in code and behaviour to me, but unfortunately as far as I can tell not too testable without exposing the texture cache at least, which is a bit of a shame.

peppy:
It should be visible by watching `TextureGLSingle` memory consumption with osu!, yep. I believe tests can be made to cover this, but not sure it's worth the effort right now.

bdach:
Fair point, given now that we have tools to inspect all allocated textures spotting this if this should happen again should be obvious enough.

PR #3825, opened 2020-08-14
Fix DLUW not unloading when masked away

smoogipoo:
Fixes the issue found in https://github.com/ppy/osu/pull/9849

The case here is: begin an async load of the beatmap panel, and immediately scroll down to mask it away. Since the `EndDelayedLoad` callback was scheduled locally, DLUW won't have a chance to schedule the unload.

### DIFF
--- a/osu.Framework/Graphics/Containers/DelayedLoadUnloadWrapper.cs
+++ b/osu.Framework/Graphics/Containers/DelayedLoadUnloadWrapper.cs
@@ -70,7 +70,7 @@ namespace osu.Framework.Graphics.Containers
             content.LifetimeEnd = lifetimeEnd;
 
             // Scheduled for another frame since Update() may not have run yet and thus OptimisingContainer may not be up-to-date
-            Schedule(() =>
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/Containers/DelayedLoadUnloadWrapper.cs
```diff
@@ -70,7 +70,7 @@ protected override void EndDelayedLoad(Drawable content)
             content.LifetimeEnd = lifetimeEnd;
```
bdach, line 70: Is setting `Lifetime{Start,End}` safe to do on the game scheduler? There is a callback hooked up if the value changes (`LifetimeChanged`); this might also need to be enqueued onto the local scheduler if I understand correctly. Not sure which semantics we're after here, but this is a potential change in them.

smoogipoo: You're probably correct in thinking it's unsafe to do, I've added a schedule for it :+1:

PR #3829, opened 2020-08-15
Maintain internal pause on tracks with zero frequencies

frenzibyte:
Definitely was against copying all logic from `SampleChannelBass` to `TrackBass`, and wasn't sure about having a very small `BassZeroFrequencyPauser` helper class, so I decided to introduce a class that does all the relevant shared logic, which is taking in relative frequency values, converting them to actual channel frequencies, and handling the case where the relative frequency value may be `0`, by requesting the component to maintain an internal pause of their channel.

Test cases were added inside `TrackBassTest` instead as tracks show their current time, thus making it easier to test headlessly. (and saves up from having another track test scene)


osu.Framework/Audio/BassRelativeFrequencyHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        private int channel;
+        private float initialFrequency;
+
+        public Action RequestZeroFrequencyPause;
+        public Action RequestZeroFrequencyResume;
+
+        public bool ZeroFrequencyPauseRequested { get; private set; }
+
+        public void SetChannel(int c)
```
bdach, line 23: This should ideally not be a method, but a constructor. The class has zero purpose without receiving a channel.

osu.Framework/Audio/BassRelativeFrequencyHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            // http://bass.radio42.com/help/html/ff7623f0-6e9f-6be8-c8a7-17d3a6dc6d51.htm (BASS_ATTRIB_FREQ's description)
+            const int min_bass_freq = 100;
+            const int max_bass_freq = 100000;
+
+            int channelFrequency = (int)Math.Clamp(Math.Abs(initialFrequency * relativeFrequency), min_bass_freq, max_bass_freq);
+            Bass.ChannelSetAttribute(channel, ChannelAttribute.Frequency, channelFrequency);
+
+            // Maintain internal pause on zero frequency due to BASS not supporting them (0 is took for original rate in BASS API)
+            // Above documentation shows the frequency limits which the constants (min_bass_freq, max_bass_freq) came from.
```
bdach, line 41: not sure why this comment is here and not near the link

osu.Framework/Audio/BassRelativeFrequencyHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public void SetChannel(int c)
+        {
+            channel = c;
+            ZeroFrequencyPauseRequested = false;
+
+            Bass.ChannelGetAttribute(channel, ChannelAttribute.Frequency, out initialFrequency);
+        }
+
+        public void UpdateChannelFrequency(double relativeFrequency)
```
bdach, line 31: this is more of a `SetChannelFrequency` (and also needs xmldoc that `relativeFrequency` is relative to the track's initial frequency, IMO)

osu.Framework.Tests/Audio/TrackBassTest.cs
```diff
@@ -268,6 +268,84 @@ public void TestRateWithAggregateAdjustments()
...
+            // set back to one and update track.
+            track.Frequency.Value = 1;
+            takeEffectsAndUpdateAfter(50);
+
+            // ensure time didn't jump away, and is progressing normally.
+            Assert.IsTrue(track.IsRunning);
+            Assert.Greater(track.CurrentTime, currentTime);
+            Assert.Less(track.CurrentTime, currentTime + 1000.0);
+
+            void takeEffectsAndUpdateAfter(int after)
```
bdach, line 304: what is this name? what "effects"?

also why is this local function duplicated in both tests?

osu.Framework/Audio/BassRelativeFrequencyHandler.cs
frenzibyte: Thought same, but `SampleChannelBass` creates channels every new playback. one could argue this should then be constructed every playback, but `BassAmplitudesProcessor` does the exact same as what this is doing.

bdach: not entirely true, I looked at that. it has both a ctor and a `SetChannel` method.

frenzibyte: what's the point, `SetChannel` is required either way. unless you're suggesting for this to have a constructor as well, then sure.

bdach: yes, I'd rather this class was always guaranteed to have a channel associated with it, to avoid potential misuse due to forgetting to set the channel and then setting frequency and accidentally changing parameters of channel 0.

osu.Framework.Tests/Audio/TrackBassTest.cs
frenzibyte: "take effects" from what has changed before this (e.g. `track.Frequency.Value = x`, `track.Stop()`), named it like that because it's local and thus its scope is limited to these two tests only, didn't want to fill the entire tests with `updateTrack(); Thread.Sleep(x); updateTrack();`

> also why is this local function duplicated in both tests?

I don't really think it's something that should be declared as a private method if only these two tests use it, didn't want to bother improving the quality of this and waste more time on it. (which results in no benefits whatsoever)

osu.Framework/Audio/BassRelativeFrequencyHandler.cs
frenzibyte: wasn't sure about the naming yeah, `SetChannel` is a thing though, having them both together with `SetChannel` at their start seems kinda weird, unless you think it's a better thing, then sure will rename it.

having a xmldoc sounds fine, will add it.

bdach: `SetFrequency` then maybe

osu.Framework.Tests/Audio/TrackBassTest.cs
bdach: "waste more time"? it's a 30-second refactoring, at *most*. other private methods in the class aren't used by absolutely all tests, either.

osu.Framework/Audio/BassRelativeFrequencyHandler.cs
frenzibyte: the internal pause maintenance is here, therefore the comment placement is right, the link is placed above, to show the documentation of the frequency limits, wasn't sure about duplicating the link here and there, it's a couple of lines above, will add it in here I guess.

bdach: i meant only the second line

frenzibyte: it was meant to tell that there are frequency limits back when it was at `SampleChannelBass`, still had that thinking stuck in my mind, will move it above.

osu.Framework/Audio/BassRelativeFrequencyHandler.cs
frenzibyte: either way `BassRelativeFrequencyHandler` would need to be created right when the channel is created, how about just adding an assertion that `channel != 0` on any `SetFrequency` call since that's known to be an invalid handle.

bdach: an assert could be ok, *as long as* you have a source for the claim that a zero channel is invalid.

frenzibyte: fair enough

osu.Framework.Tests/Audio/TrackBassTest.cs
frenzibyte: the thing is, the name of the method tells it shouldn't really be a private method, but sure.

osu.Framework/Audio/Sample/SampleChannelBass.cs
```diff
@@ -92,17 +73,27 @@ public override void Play(bool restart = true)
...
+                if (relativeFrequencyHandler == null)
+                {
+                    relativeFrequencyHandler = new BassRelativeFrequencyHandler(channel)
+                    {
+                        RequestZeroFrequencyPause = () => Bass.ChannelPause(channel),
+                        RequestZeroFrequencyResume = () => Bass.ChannelPlay(channel),
+                    };
+                }
+                else
+                    relativeFrequencyHandler.SetChannel(channel);
```
bdach, line 87: yeah as I said I'd probably rather have the assert here

frenzibyte: What assert?

bdach: [the one you suggested](https://github.com/ppy/osu-framework/pull/3829#discussion_r470955985)

frenzibyte: Didn't do it as I couldn't find any proof that a zero handle is invalid, as you said above, so I didn't want to bother with this even more
> an assert could be ok, as long as you have a source for the claim that a zero channel is invalid.

Alright will change.

osu.Framework/Audio/BassRelativeFrequencyHandler.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        public void SetChannel(int c)
+        {
+            channel = c;
+            ZeroFrequencyPauseRequested = false;
+
+            Bass.ChannelGetAttribute(channel, ChannelAttribute.Frequency, out initialFrequency);
+        }
+
+        public void SetFrequency(double relativeFrequency)
```
bdach, line 32: this still doesn't have xmldoc

bdach (dismissed):

osu.Framework/Audio/BassRelativeFrequencyHandler.cs
frenzibyte: Whoops, didn't pay much attention to the full comment at the time I applied, maybe better limiting off each suggestion to its own conversation, or moving that other relevant suggestion to a new line.

Anyways filled the entire class with documentation, now waiting for your documentation review :p

bdach:
Made naming changes here, reworded xmldocs. Think the diff reads way simpler (and therefore better) with the new names.

peppy (approved):
Looks sane. Let's get this in and see how it behaves.

PR #3830, opened 2020-08-15
Fix build warnings

bdach:
# Summary

Somehow, somewhere, several build warnings have popped up yet again; this PR is supposed to resolve them in the framework. Taking it from the top:

## [CA1805: Do not initialise unnecessarily](https://docs.microsoft.com/sl-si/visualstudio/code-quality/ca1805?view=vs-2017)

This one is actually split into three commits; e3563b9 is supposed to deal with the most obvious cases where you can either eliminate the default or even straight up convert to a get-only constant property with no additional cost.

d2f4ada removes `AverageFrameTime` from `InterpolatingFramedClock`, which was a remnant from an old `IFramedClock` member [removed in this commit](https://github.com/ppy/osu-framework/commit/9ad78a1335d7e764b59bb25654f363b225a14f5c).

6dfd270 contains the most interesting cases. One is a declaration of a virtual property that is supposed to default to `null` (so I left the initialiser in and suppressed the warning to make it 1000% clear along with the xmldoc that it does default to `null`), and the other is... a mystery to me - it's a `Chameleon` property in `TestSceneSizing`, which is referenced in commented lines throughout the scene. It's been stuck in there since 2016, so just in case I left it in with a suppression (as if I didn't disable that warning, another warning shows that the field is unused).

## [CA2248: Provide correct enum argument to `Enum.HasFlag`](https://docs.microsoft.com/gl-es/visualstudio/code-quality/ca2248?view=vs-2019)

Straight and simple, one case fixed in af924b5. Not much to talk about there.

## CA2249: Consider using `string.Contains` instead of `string.IndexOf` for readability

(no docs link, must be fresh!) Also a no-brainer. Runtime sources [show that the two versions are equivalent](https://github.com/dotnet/runtime/blob/6072e4d3a7a2a1493f514cdf4be75a3d56580e84/src/libraries/System.Private.CoreLib/src/System/String.Searching.cs#L24-L29). (Also includes a hilarious comment suppressing this same warning.)

# Remarks

This PR is dedicated to [this discord message](https://discord.com/channels/188630481301012481/589331078574112768/744091307152244768).

### DIFF
--- a/osu.Framework.Tests/Clocks/DecoupleableClockTest.cs
+++ b/osu.Framework.Tests/Clocks/DecoupleableClockTest.cs
@@ -378,7 +378,7 @@ namespace osu.Framework.Tests.Clocks
 
         private class TestClockWithRange : TestClock
         {
-            public double MinTime { get; set; } = 0;
+            public double MinTime => 0;
             public double MaxTime { get; set; } = double.PositiveInfinity;
 
             public override bool Seek(double position)
--- a/osu.Framework.Tests/Visual/Containers/TestSceneSizing.cs
+++ b/osu.Framework.Tests/Visual/Containers/TestSceneSizing.cs
@@ -1153,9 +1153,6 @@ namespace osu.Framework.Tests.Visual.Containers
 
     internal class InfofulBox : Container
     {
-#pragma warning disable CA1805
-        public bool Chameleon = false;
-#pragma warning restore CA1805
         public bool AllowDrag = true;
 
         protected override void OnDrag(DragEvent e)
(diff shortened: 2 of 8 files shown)


frenzibyte:
About warning `CA1805`, that just sounds like that warning should be entirely disabled IMHO. if a redundant value was intentionally assigned to it, then there's probably something behind it, unless I'm missing a case this warning would get rid of.

peppy:
I've pushed two changes to remove the necessity for the disable rules. Please check.

peppy:
@frenzibyte i think you may be misunderstanding the rule. redundant means redundant, as in static analysis concludes it's 100% redundant.

frenzibyte:
@peppy Actually, yeah right, as that would also cause inconsistencies codewise between fields being initialized explicitly and others implicitly. makes sense now.

bdach:
Looks good here too. Weird brain lapse on the `null` case.

PR #3832, opened 2020-08-17
Revert device period to fix linux audio crackling

smoogipoo:
Resolves https://github.com/ppy/osu/issues/9865

Needs more testing, especially with existing user configs (I tested with default PA config).

### DIFF
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -293,7 +293,6 @@ namespace osu.Framework.Audio
                 return true;
 
             // reduce latency to a known sane minimum.
-            Bass.Configure(ManagedBass.Configuration.DevicePeriod, 1);
             Bass.Configure(ManagedBass.Configuration.DeviceBufferLength, 10);
             Bass.Configure(ManagedBass.Configuration.PlaybackBufferLength, 100);
 


peppy:
Did you not say the default was 200ms or something for you? If so, shall we try 10?

smoogipoo:
100ms is the reported PA latency. 10 is already the [BASS default](http://www.un4seen.com/doc/#bass/BASS_CONFIG_DEV_PERIOD.html).

smoogipoo:
@bdach When possible, can you try increasing the device buffer length to 20 or 30 and see if it still causes crackling (with dev period = 1)?

bdach:
I've tested a bit and it seems that at least on my system fiddling with device buffer length (even to ridiculous degrees, like setting 100) doesn't do much and it's all about the device period. Anything less than ~5-ish is crackly.

That said there seems to be something else going on there because the audio glitches don't exactly always appear - to get a reliable reproduction I need to `$ pulseaudio -k && pulseaudio --start` before every launch to ensure I get the crackles. I have precisely zero idea why that might be, though...

smoogipoo:
One thing that I've been able to replicate is when osu! is started along with _other_ sound (e.g. from youtube), the full system audio goes crackly and distorted and... weird.

Probably best to just revert this and let people set the settings they deem fit for their system.

Snowmander:
I tested the values out and the result doesn't make sense to me.

```
DevicePeriod 1 / DeviceBufferLength 1 / PlaybackBufferLength 1
-> stable

DevicePeriod 1 / DeviceBufferLength 10 / PlaybackBufferLength 100
-> stable

DevicePeriod commented out / DeviceBufferLength 1 /PlaybackBufferLength 1
-> audio completely broken

DevicePeriod commented out / DeviceBufferLength 1 / PlaybackBufferLength 35
-> stable (lowest I can go without audio breaks)

DevicePeriod commented out / DeviceBufferLength 10 / PlaybackBufferLength 100
-> stable
```

Tested on ALSA (exclusive). To see which setting is actually the best, latency needs to be tested.

Can't wait for these values to be adjustable ingame as it'd be nice if I didn't have to recompile every time to test or play osu.

bdach:
How many times did you repeat each try? Just once? Also, did you kill pulseaudio before every go?

Otherwise the audio is majorly inconsisten to me too. Sometimes it'll even be good with 1 device period but sometimes not. Probably just best to point linux users at pulseaudio configurations as I don't believe we should be making this adjustable in game.

Snowmander:
I'm not using pulseaudio because it's pain in the butt when it comes to low latency audio. Just plain alsa.

The testing is consistent for me. Back when I used pulseaudio with thepoon's modified wine for stable, the results weren't consistent for me as well. With pulse, CPU usage and general weirdness with pulseaudio can affect the test run a lot. Even in a "safe" setting, the game will perform differently depending on which programs (such as baloo file indexer or chrome) is running in the background.

Snowmander:
> I don't believe we should be making this adjustable in game.

It'd be really nice to be able to adjust these values ingame because it would allow the users to squeeze every bit of latency reduction they can get on their own risks. The reduction in latency is substantial (in stable, it goes from ~100ms down to 10ms) and the demand is definitely there, because there are many linux users who use patched wine to accomplish this on stable. I have 9ms latency on stable and the hitsounds sound so responsive compared to standard setups.

ArniDagur:
* Would it be a welcome PR to implement the ability to change these values in-game? You'd have to expose an API here, and then use said API in `osu!lazer`, if I'm not mistaken. I'm guessing I would add another file in https://github.com/ppy/osu/tree/master/osu.Game/Overlays/Settings/Sections/Audio?
* I'd always have to re-initialize BASS for every change, right?

peppy:
- It would need to be restricted to a platform specific implementation (ie. only visible for linux users)
- It would need to be done in a way that applies immediately. If this means a bass reinit is required, framework support for doing that automatically should likely be added.

Honestly I would hope for a better answer than exposing this to the user, but can understand if that's not feasible due to linux being linux.

ArniDagur:
The period setting applies to all platforms, not just Linux, right?

* Is your thinking that the hardware/software variability between different Windows and Mac computers is relatively smaller than for Linux, so one can justify the same value for everyone?
* Is it a goal to expose as few settings to the users as possible, for the sake of simplicity?

My intuition tells me that even for Windows systems, different users may like different settings for these values. Faster hardware may be able to meet stricter latency targets, and audio subsystem configuration (e.g. https://old.reddit.com/r/osugame/comments/hqoakj/tutorial_removing_perceivable_latency_for_good/) may play a part as well. Furthermore, some players may be fine with a little audio crackling if it means that they get lower latency, while others may be more sensitive to it (e.g. Twitch streamers who are concerned with audio quality). Thus, **my** preference would be to expose the settings on every platform.

peppy:
- I can tell you 100% that it does not affect windows as the minimum buffer length is managed by the OS itself.
- macOS during testing these PRs I could not notice any effect. Would require further confirmation but I'm guessing it's similar to windows.

ArniDagur:
For those interested, I just submitted https://github.com/ppy/osu-framework/pull/4008 and  https://github.com/ppy/osu/pull/10892.

PR #3835, opened 2020-08-17
Move IAggregateAudioAdjustment.GetAggregate() to extensions method

frenzibyte:
# Breaking changes

## `IAggregateAudioAdjustment.GetAggregate()` has been made an extension method

To avoid aggregate adjustment implementations from having to implement both `Aggregate{Volume,Balance,Frequency,Tempo}` and `GetAggregate()`, the latter has been made an extension method returning one of the aforementioned four values for the appropriate property of the adjustment.

Possible compilation failures related to `GetAggregate()` should be resolved by adding a using statement for the `osu.Framework.Audio` namespace to the files affected. 

---

Probably incorrect to have both `Aggregate{Volume/Balance/etc..}` and `GetAggregate()` require implementation from the class when one should redirect to the other.

Could use default interface implementation for this, but would cause changing every usage to do an explicit cast to the interface (`((IAggregateAudioAdjustment)adjustment).GetAggregate()`), which doesn't sound like something that would be allowed anywhere.

### DIFF
--- a/osu.Framework/Audio/AdjustableAudioComponent.cs
+++ b/osu.Framework/Audio/AdjustableAudioComponent.cs
@@ -74,8 +74,6 @@ namespace osu.Framework.Audio
 
         public IBindable<double> AggregateTempo => adjustments.AggregateTempo;
 
-        public IBindable<double> GetAggregate(AdjustableProperty type) => adjustments.GetAggregate(type);
-
         protected override void Dispose(bool disposing)
         {
             base.Dispose(disposing);
--- a/osu.Framework/Audio/AggregateAdjustmentExtensions.cs
+++ b/osu.Framework/Audio/AggregateAdjustmentExtensions.cs
@@ -1,0 +1,38 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using osu.Framework.Bindables;
+
+namespace osu.Framework.Audio
+{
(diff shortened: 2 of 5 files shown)


bdach:
This might actually be a slightly breaking change. I can imagine that in some scenarios adding an using statement for `osu.Framework.Audio` might be required if there wasn't one present previously in the file.

frenzibyte:
Oh good thinking, but still that's for usages outside framework (and outside subclasses of audio components as well), which is really narrows it down to near-zero, if not just zero.

peppy:
can you explain "something which wouldn't be allowed anywhere"?

frenzibyte:
It would be kind of a downside for readability, especially when it's an argument inside a method call, like one of `AudioAdjustments`'s usage for example:
```diff
     foreach (AdjustableProperty type in Enum.GetValues(typeof(AdjustableProperty)))
-        getAggregate(type).AddSource(component.GetAggregate(type));
+        getAggregate(type).AddSource(((IAggregateAudioAdjustments)component).GetAggregate(type));
```
but I can change it to a default interface implementation if you think it's better that way.

bdach:
I think this is fine and a default interface implementation would be uglier, but probably a second opinion would be good.

I've added an appropriate breaking change note to the OP.

peppy (approved):

peppy (approved):

peppy:
@smoogipoo double check you're okay with this change

smoogipoo (approved):

smoogipoo:
lgtm

PR #3836, opened 2020-08-17
Make drawable-wrapped audio components respect clock running state

frenzibyte:
- [x] Depends on #3829 
- [x] Depends on #3835

Makes components-backed `DrawableAudioWrapper`s respect the `IClock.IsRunning` state that's assigned to them, if stopped then a frequency adjustment to `0` would be applied internally to the underlying component itself, not visible by `DrawableAudioWrapper.AggregateFrequency`.

## Introducing `AudioWrapperAdjustments`

This introduces a private `AudioWrapperAdjustments` class, which has extra adjustments ("clock state frequency adjustment" currently) mixed internally with the public adjustments (`PublicAdjustments`) all to the `IAggregateAudioAdjustment` interface explicit implementation. 

This is done as a way to make "internal adjustments to the underlying component" possible without interfering with what the consumer adjusts or what children would bind to. 

Also chose against creating another `AudioAdjustments` for the internal adjustments as it has a useless amount of bindables in it, if it's possible to wrap some aggregates to their public ones and create aggregate bindables for others then why wasting.

---

Also adds a headless test scene for `DrawableAudioWrapper` ensuring the adjustments for the wrapper and the underlying component are as expected and are isolated from each other, and thus they should behave correctly in, mostly all cases.

---

### Note:

This alone doesn't entirely replace the existing "pause sample on gameplay pause logic" osu!-side, after testing with this logic turns out `FrameStabilityContainer` doesn't stop its `StabilityGameplayClock` when pausing, which is probably a bug as the parent gameplay clock stops, will look at fixing it as soon as possible.

### DIFF
--- a/osu.Framework.Tests/TestSceneDrawableAudioWrapper.cs
+++ b/osu.Framework.Tests/TestSceneDrawableAudioWrapper.cs
@@ -1,0 +1,148 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Linq;
+using NUnit.Framework;
+using osu.Framework.Audio;
+using osu.Framework.Audio.Sample;
+using osu.Framework.Audio.Track;
+using osu.Framework.Bindables;
+using osu.Framework.Graphics.Audio;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Testing;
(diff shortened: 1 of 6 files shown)


frenzibyte:
Closing as deemed unnecessary + conflicting with [osu!lazer's direction](https://github.com/ppy/osu/pull/10282/files), leaving it up to the consumer to pause their samples properly seems to be the best way to achieve this.

Issue #3837, opened 2020-08-17
Transforms can be skipped entirely


bdach:
Posting this as an issue because I've tried two approaches and while they fix the issue I'm about to describe, they fail other tests and I'm very worried this will either result in a revert of the recent transform optimisations or necessitate a rewrite of transform code entirely.  I can't describe this in the title any better, really.

Given this test case:

```csharp
        [Test]
        public void TestSimultaneousTransformsOutOfOrder()
        {
            boxTest(box =>
            {
                using (box.BeginAbsoluteSequence(0))
                {
                    box.MoveToX(0.5f, 4 * interval);
                    box.Delay(interval).MoveToY(0.5f, 2 * interval);
                }
            });

            checkAtTime(0, box => Precision.AlmostEquals(box.Position, new Vector2(0)));
            checkAtTime(interval, box => Precision.AlmostEquals(box.Position, new Vector2(0.125f, 0)));
            checkAtTime(2 * interval, box => Precision.AlmostEquals(box.Position, new Vector2(0.25f, 0.25f)));
            checkAtTime(3 * interval, box => Precision.AlmostEquals(box.Position, new Vector2(0.375f, 0.5f)));
            checkAtTime(4 * interval, box => Precision.AlmostEquals(box.Position, new Vector2(0.5f)));
            checkAtTime(3 * interval, box => Precision.AlmostEquals(box.Position, new Vector2(0.375f, 0.5f)));
            checkAtTime(2 * interval, box => Precision.AlmostEquals(box.Position, new Vector2(0.25f, 0.25f)));
            checkAtTime(interval, box => Precision.AlmostEquals(box.Position, new Vector2(0.125f, 0)));
            checkAtTime(0, box => Precision.AlmostEquals(box.Position, new Vector2(0)));
        }
```

`TargetGroupingTransformTracker` will skip over the longer transform when it's done processing the shorter one. This is because the tracker presumes the following:

https://github.com/ppy/osu-framework/blob/98fe709459a76488f5d338ff6ee5c11ec64583ac/osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs#L198-L200

That is not necessarily always the case. What *is* true is that the transforms are sorted by start time:

https://github.com/ppy/osu-framework/blob/98fe709459a76488f5d338ff6ee5c11ec64583ac/osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs#L27

[...]

PR #3838, opened 2020-08-17
Fix transforms added during load being potentially not processed

bdach:
Aims to resolve ppy/osu#9832, although I'm not sure if wasted effort in light of #3837.

# Summary

Most of the paper trail as to why this arose is [in the issue](https://github.com/ppy/osu/issues/9832#issuecomment-674459266). In short: adding a transform during load or outside hierarchy would wrongly update a single tracker's last applied transform index, while the transformable itself kept a last transform time which was not updated in line, leading to a desync.

Resolve as proposed [here](https://github.com/ppy/osu/issues/9832#issuecomment-674482116) by updating the entire transformable.

# Remarks

This resolution actually let me uncover #3837, as I went to check a storyboard and saw that it wasn't processing `X` transforms properly (but `Y` transforms worked). That was due to them having an equal `StartTime`.

I also tried [moving the last transform time to the tracker](https://github.com/ppy/osu-framework/compare/master...bdach:add-transform-in-the-past-v3) but I'm not sure I like that better over this. It's a CPU-memory tradeoff; that version reads a little nicer maybe but duplicating the same state/having trackers have their own separate last transform update time sounds like trouble down the line.

### DIFF
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
@@ -254,6 +254,37 @@ namespace osu.Framework.Tests.Visual.Drawables
         }
 
         [Test]
+        public void AddPastTransformFromFutureWhenNotInHierarchy()
+        {
+            AddStep("seek clock to 1000", () => manualClock.CurrentTime = interval * 4);
+
+            AddStep("create box", () =>
+            {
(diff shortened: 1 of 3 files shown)


peppy:
Going to tentatively merge this so there's at least one remaining issue to tackle. This looks sane. I'm also not sure where I prefer this being, but let's make minimal changes for now until we're sure the tracker optimisation can actually exist.

PR #3839, opened 2020-08-17
Invoke text mutation method events only when done by user input

frenzibyte:
Closes ppy/osu#9882
Regression from #3289

Replaces text mutation events to only being called when the mutation is done by user input.

## Remarks
- There is a commented piece of the pushed test case, which is about pressing a key and checking if the event handler has been invoked, but currently, the text input consumption logic is limited at a host level from manipulating or creating a test class for it, making this not straightforward to test, didn't want to bother with it so I added a TODO comment on top of it for now.
- First time using regions inside methods, but felt better than splitting each in an own test case method as it requires setting up the same textbox for each, lengthening the test scene even more.
- There is also a side-issue with the textbox text property that became visible on the written test (setting `Text` a second time), will report it asap.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System.Collections.Generic;
 using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Graphics;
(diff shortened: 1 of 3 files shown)


osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
```diff
@@ -412,6 +413,103 @@ public void TestReplaceSelectionWhileLimited()
...
+                });
+            });
+
+            AddStep("focus textbox", () =>
+            {
+                InputManager.MoveMouseTo(textBox);
+                InputManager.Click(MouseButton.Left);
+            });
+
+            #region Text mutation
```
peppy, line 438: regions inside methods? i don't like it. please split the test up into multiple methods if you want regions.

frenzibyte: Could replace regions with inline comments if that's something you like, the goal is to make the test method more readable than just a bunch of pieces each testing something different than the other. 

Splitting the test pieces means creating the custom textbox and focusing on it for each as well if that's fine for you.

peppy: If you aren't willing to fix this whole test scene, maybe just make a new one called `TestSceneTextBoxEvents` and use `SetUpSteps` / `SetUp` as we would normally.

frenzibyte: creating a new test scene sounds like a pretty good idea, pushed it. don't really have time to look at improving the whole test scene honestly, maybe will look at some point.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -557,18 +559,18 @@ private void insertString(string value, Action<Drawable> drawableCreationParamet
         protected abstract void NotifyInputError();
 
         /// <summary>
-        /// Invoked whenever a text string has been inserted to <see cref="Text"/>.
+        /// Invoked whenever pending text input from user has been consumed, this is not necessarily what has been inserted to <see cref="Text"/> (i.e. some may have been disallowed by <see cref="CanAddCharacter"/>)
```
peppy, line 562:
```suggestion
        /// Invoked when new text is added via user input```

Rename event to `OnUserTextAdded`. no need to call it consumed.

File: osu.Framework/Graphics/UserInterface/TextBox.cs [hunk 3]
```diff
@@ -557,18 +559,18 @@ private void insertString(string value, Action<Drawable> drawableCreationParamet
...
         /// </summary>
-        /// <param name="added">The inserted text string.</param>
-        protected virtual void OnTextAdded(string added)
+        /// <param name="consumed">The consumed text string.</param>
```
peppy, line 564:
```suggestion
        /// <param name="added">The text which was added.</param>
```

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -557,18 +559,18 @@ private void insertString(string value, Action<Drawable> drawableCreationParamet
...
+        protected virtual void OnUserTextConsumed(string consumed)
```
peppy, line 565:
```suggestion
        protected virtual void OnUserTextConsumed(string added)
```

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -557,18 +559,18 @@ private void insertString(string value, Action<Drawable> drawableCreationParamet
...
         {
         }
 
         /// <summary>
-        /// Invoked whenever a text string has been removed from <see cref="Text"/>.
+        /// Invoked whenever a text string has been removed by user input.
```
peppy, line 570:
```suggestion
        /// Invoked when text is removed via user input.
```

PR #3840, opened 2020-08-18
Expose PerformTabSort method in TabControl

EVAST9919:


bdach:
Would you mind to give context on how you're intending to use this so that we can ascertain whether this is the correct direction?

I'm guessing this has to do with ppy/osu#9896, but why does the pinning not work as expected in that scenario?

EVAST9919:
> Would you mind to give context on how you're intending to use this so that we can ascertain whether this is the correct direction?
> 
> I'm guessing this has to do with [ppy/osu#9896](https://github.com/ppy/osu/pull/9896), but why does the pinning not work as expected in that scenario?

Autosort works as expected but I want to manually perform sort when selected item moved from main content to dropdown (there's an updated video how it looks like)

bdach:
Holding off on this until we know whether the desired behaviour in the game-side PR can be achieved (might require more changes framework-side, or maybe none at all).

EVAST9919:
Closing as not-needed for now (and probably not needed at all)

PR #3841, opened 2020-08-18
Update MarkDig package

smoogipoo:
Not sure why the PR was closed.


peppy:
Because it requires updates at our end https://ci.appveyor.com/project/peppy/osu-framework/builds/34702255. In an effort to reduce noise I closed this rather than wasting time/effort on something that won't benefit us.

peppy (approved):

PR #3842, opened 2020-08-18
Fix transforms potentially being skipped due to incorrect order of application

peppy:
- Closes https://github.com/ppy/osu-framework/issues/3837.

Best understood by going through commits one by one and looking at test failures in `TestSceneTransformRewinding` in visual tests, but basically:

- We don't want to be applying transforms in the rewind code. This is iterating in the wrong direction through the transforms list and will not lead to correct application order between overlapping TargetMembers.
- With the removal of the application just mentioned, a regression shows up when adding a transform in the past which will immediately complete in the next frame – rewind logic will not be run to reset `AppliedToEnd` so the end value may not be updated correctly. This is resolved by marking as *not* `AppliedToEnd` in `AddTransform` when required.
- All dictionary / cache logic is taken from previous iteration I had written (https://raw.githubusercontent.com/ppy/osu-framework/aa8eaf127060cc4a0be82e5ab616c09bc44a5fff/osu.Framework/Graphics/Transforms/Transformable.cs)

### DIFF
--- a/osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
+++ b/osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
@@ -39,7 +39,7 @@ namespace osu.Framework.Graphics.Transforms
         /// <summary>
         /// The index of the last transform in <see cref="transforms"/> to be applied to completion.
         /// </summary>
-        private readonly Dictionary<string, int> lastAppliedIndex = new Dictionary<string, int>();
+        private readonly Dictionary<string, int> lastAppliedTransformIndices = new Dictionary<string, int>();
 
         /// <summary>
         /// All <see cref="Transform.TargetMember"/>s which are handled by this tracker.
@@ -99,7 +99,7 @@ namespace osu.Framework.Graphics.Transforms
                 }
             }
 
-            for (int i = getLastAppliedIndex() ?? 0; i < transforms.Count; ++i)
+            for (int i = getLastAppliedIndex(); i < transforms.Count; ++i)
             {
                 var t = transforms[i];
 
@@ -117,7 +117,7 @@ namespace osu.Framework.Graphics.Transforms
                     // Since following transforms acting on the same target member are immediately removed when a
                     // new one is added, we can be sure that previous transforms were added before this one and can
                     // be safely removed.
-                    for (int j = getLastAppliedIndex(t.TargetMember) ?? 0; j < i; ++j)
+                    for (int j = getLastAppliedIndex(t.TargetMember); j < i; ++j)
                     {
                         var u = transforms[j];
                         if (u.TargetMember != t.TargetMember) continue;
@@ -320,21 +320,19 @@ namespace osu.Framework.Graphics.Transforms
             }
         }
 
-        private readonly Dictionary<string, int?> lastAppliedTransformIndices = new Dictionary<string, int?>();
-
         /// <summary>
         /// Retrieve the last transform index that was <see cref="Transform.AppliedToEnd"/>.
         /// </summary>
         /// <param name="targetMember">An optional target member. If null, the lowest common last application is returned.</param>
-        private int? getLastAppliedIndex(string targetMember = null)
+        private int getLastAppliedIndex(string targetMember = null)
         {
             if (targetMember == null)
                 return lastAppliedTransformIndices.Values.Min();
 
-            if (lastAppliedTransformIndices.TryGetValue(targetMember, out int? val))
+            if (lastAppliedTransformIndices.TryGetValue(targetMember, out int val))
                 return val;
 
-            return null;
+            return 0;
         }
 
         /// <summary>
@@ -342,7 +340,7 @@ namespace osu.Framework.Graphics.Transforms
         /// </summary>
         /// <param name="targetMember">The target member to set the index of.</param>
         /// <param name="index">The index of the transform in <see cref="transforms"/>.</param>
-        private void setLastAppliedIndex(string targetMember, int? index = null)
+        private void setLastAppliedIndex(string targetMember, int index)
         {
             lastAppliedTransformIndices[targetMember] = index;
         }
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
```diff
@@ -323,9 +328,40 @@ private void invokePendingRemovalActions()
             }
         }
 
+        private readonly Dictionary<string, int?> lastAppliedTransformIndices = new Dictionary<string, int?>();
+
+        /// <summary>
+        /// Retrieve the last transform index that was <see cref="Transform.AppliedToEnd"/>.
+        /// </summary>
+        /// <param name="targetMember">An optional target member. If null, the highest common last application is returned.</param>
```
bdach, line 336:
```suggestion
        /// <param name="targetMember">An optional target member. If null, the lowest common last application is returned.</param>
```

osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
```diff
@@ -227,14 +223,23 @@ public void AddTransform(Transform transform, ulong? customTransformID = null)
...
+            for (int i = 0; i < transforms.Count; ++i)
             {
                 var t = transforms[i];
 
-                if (t.TargetMember == transform.TargetMember)
+                if (!insertionWasAtEndOfList && i < insertionIndex)
+                {
+                    // as we are removing all transforms after the newly added one, the TargetMember may not be in a good state.
+                    // this could mean that we get an incorrect value read into our StartValue on next update.
+                    // To avoid this, we ensure all transforms before the insertion point on next update.
```
bdach, line 239: accidentally a word, I think

```suggestion
                    // To avoid this, we ensure all transforms before the insertion point run on next update.
```

osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
```diff
@@ -227,14 +223,23 @@ public void AddTransform(Transform transform, ulong? customTransformID = null)
...
+                    t.AppliedToEnd = false;
+                }
+                else if (i > insertionIndex && t.TargetMember == transform.TargetMember)
```
bdach, line 242: While this seems correct something's off here. I checked out 6159b332 (so the commit before this change) and got no test fails across the framework test suite. Which one is supposed to fail?

peppy: I believe `AddPastTransformFromFuture` should fail..

peppy: The logic being "transforms in the past were removed because of this new transform being added. the transforms before the ones that were removed need to replay their final states so the new transform we added (also in the past but will be run to completion) gets the correct starting state".

How are you checking for test failure? nunit?

bdach: Both nunit and test scene. I've tried multiple times to make it fail but it doesn't seem to...

peppy: Reverted this commit for now.

osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
```diff
@@ -80,16 +80,12 @@ public void UpdateTransforms(in double time, bool rewinding)
...
-                        // note that we should only do this for the last transform to avoid incorrect application order.
+                        // note that we should only do this for the last transform of each TargetMember to avoid incorrect application order.
                         // the actual application will be in the main loop below now that AppliedToEnd is false.
-                        if (!appliedToEndRevert)
+                        if (!appliedToEndReverts.Contains(t.TargetMember))
                         {
-                            if (time < t.EndTime)
-                                t.AppliedToEnd = false;
-                            else
-                                t.Apply(t.EndTime);
```
smoogipoo, line 90: So this is removed because of "We don't want to be applying transforms in the rewind code", but what about the code in the `else` block just below? Is that fine? How is correct ordering enforced there?

osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
```diff
@@ -39,7 +39,7 @@ internal class TargetGroupingTransformTracker
         /// <summary>
         /// The index of the last transform in <see cref="transforms"/> to be applied to completion.
         /// </summary>
-        private int? lastAppliedIndex;
+        private readonly Dictionary<string, int> lastAppliedIndex = new Dictionary<string, int>();
```
smoogipoo, line 42: This is unused, either remove this dictionary or the one below. Probably the one below since it's an `int?` where it doesn't seem necessary to be.

smoogipoo (changes requested):

osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
bdach: I think the difference is that in the case of transforms that are in progress while the rewind runs, there is no need to apply them as the forward pass will apply them (and in correct order, too). Transforms yet to play need to be eagerly rewound to their start time as they won't be processed below. Removing the application in the `else` branch fails a test (`TestMultipleTransformTargets`) that seeks to `double.MinValue`.

osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
peppy: good catch

osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
peppy: For what it's worth, I haven't considered the else case. I'm looking to bring things back to a known state. There may still be remaining issues but those issues were present before any of these optimisations, and we can deal with them when they come up.

smoogipoo (approved):

bdach (approved):

Issue #3215, opened 2020-01-28
GridContainer computes bounding box too many times


smoogipoo:
A simple optimisation `GridContainer` can do is use `RequiredParentSizeToFit` instead of `BoundingBox`.

PR #3843, opened 2020-08-18
Switch BoundingBox with RequiredParentSizeToFit

lanpai:
Resolves https://github.com/ppy/osu-framework/issues/3215

### DIFF
--- a/osu.Framework/Graphics/Containers/GridContainer.cs
+++ b/osu.Framework/Graphics/Containers/GridContainer.cs
@@ -276,8 +276,8 @@ namespace osu.Framework.Graphics.Containers
         }
 
         private static bool shouldConsiderCell(Drawable cell) => cell != null && cell.IsAlive && cell.IsPresent;
-        private static float getCellWidth(Drawable cell) => shouldConsiderCell(cell) ? cell.BoundingBox.Width : 0;
-        private static float getCellHeight(Drawable cell) => shouldConsiderCell(cell) ? cell.BoundingBox.Height : 0;
+        private static float getCellWidth(Drawable cell) => shouldConsiderCell(cell) ? cell.RequiredParentSizeToFit.X : 0;
+        private static float getCellHeight(Drawable cell) => shouldConsiderCell(cell) ? cell.RequiredParentSizeToFit.Y : 0;
 
         /// <summary>
         /// Distributes any available length along all distributed dimensions, if required.


bdach:
I'm not sure this is an entirely free optimisation. While it's true that `RequiredParentSizeToFit` is based on `BoundingBox`, it also takes anchor into account. Therefore, the following test case passes on `master` but fails with this change:

```csharp
        [Test]
        public void TestSingleCellWithCustomAnchor()
        {
            FillBox box = null;
            AddStep("set content", () =>
            {
                grid.RelativeSizeAxes = Axes.None;
                grid.AutoSizeAxes = Axes.Both;
                grid.Content = new[]
                {
                    new Drawable[]
                    {
                        box = new FillBox
                        {
                            RelativeAnchorPosition = new Vector2(0.3f, 0.7f),
                            RelativeSizeAxes = Axes.None,
                            Width = 200,
                            Height = 150
                        }
                    },
                };
                grid.RowDimensions = new[] { new Dimension(GridSizeMode.AutoSize), };
                grid.ColumnDimensions = new[] { new Dimension(GridSizeMode.AutoSize), };
            });
            AddAssert("box is same size as grid", () => Precision.AlmostEquals(box.DrawSize, grid.DrawSize));
        }
```

Unsure if the above is the behaviour we actually want, but this will likely be a breaking change in some scenarios.

smoogipoo:
I'm actually not sure there is any optimisation here, beyond the caching of `ParentSizeToFit`...? Maybe a better optimisation is to cache `BoundingBox`, but I'd really like to know how expensive that really is in the first place because the grid cells only invalidate when the content has been invalidated with `ParentSizeToFit` (let's ignore `Presence` invalidations for simplicity), at which point any cached `BoundingBox` is likely to be invalidated anyway.

Perhaps a better optimisation is to not call `BoundingBox` multiple times by way of combining the two passes of `getCellSizesAlongAxes` into one?

bdach:
Yeah, this needs actual benchmarking for sure, regardless of the final direction.

smoogipoo:
While cleaning up pull requests, we found that your contributions here were forgotten about either due to dangling reviews or going in a direction we did not fully agree with. Given its age and inactivity, we're going to close this for now.

If you're still interested in seeing this change happen, either by bringing this pull request up-to-date or re-opening as a fresh one, please let us know in this thread before doing so and we will elaborate on what needs to be done (and whether it's a change we see being mergeable in the future).

PR #3844, opened 2020-08-19
Add origin detection and compensation for joystick axes

### LINKED ISSUE #3802: Auto-detect joystick axis origins
Currently, each axis' origin is its center which makes sense for analog sticks, but not for triggers.

Trigger origins are often on one end or the other of the axis currently, which, along with the current automatic deadzone system, renders these triggers unusable. 

My suggestion: detect the axis origin at the same time as the deadzone, by taking the origin to which it is the closest to (whether `-1`, `0`, or `1`)

Examples:

Analog sticks' resting positions are centered, auto-detection should therefore use an origin of `0`.
Triggers' resting position are on the `-1` end of the axis, therefore, their origin should be auto-detected to `-1`.

That being said, SDL would remove the need for such detection given that triggers seem to be handled directly at SDL level if we look at #3708 as well as this piece of documentation:
https://wiki.libsdl.org/SDL_GameControllerAxis // https://wiki.libsdl.org/SDL_GameControllerGetAxis

Jeremiidesu:
Fixes #3802, possibly (resolves) #2854 

The selected origin depends on the initial axis values, as explained on this diagram:
image

Deadzone detection is then adjusted accordingly be measured around the detected origin rather than around the `0` mark.

### DIFF
--- a/osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
+++ b/osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
@@ -243,8 +243,6 @@ namespace osu.Framework.Input.Handlers.Joystick
                             origin = 1;
                         else if (axisValue < 0)
                             origin = -1;
-                        else
-                            Logger.Log("No origin value was somehow determined", level: LogLevel.Important);
 
                         axisOrigins.Value[i] = origin;
 
--- a/osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
+++ b/osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
@@ -107,6 +107,11 @@ namespace osu.Framework.Input.Handlers.Joystick
                 for (int i = 0; i < JoystickDevice.MAX_AXES; i++)
                 {
                     var value = device.RawState.GetAxis(i);
+                    var origin = device.AxisOrigins?[i] ?? 0;
+
+                    // Rescale and flip for off-center origin
+                    if (origin != 0)
+                        value = (origin - value) / (2 * Math.Sign(origin));
 
                     // do not allow a deadzone below float_epsilon
                     var deadzone = MathF.Max(device.DefaultDeadzones?[i] ?? 0, Precision.FLOAT_EPSILON);
@@ -194,11 +199,17 @@ namespace osu.Framework.Input.Handlers.Joystick
             public osuTK.Input.JoystickState RawState { get; private set; }
 
             private readonly Lazy<float[]> defaultDeadZones = new Lazy<float[]>(() => new float[MAX_AXES]);
+            private readonly Lazy<float[]> axisOrigins = new Lazy<float[]>(() => new float[MAX_AXES]);
 
             /// <summary>
             /// Default deadzones for each axis. Can be null if deadzones have not been found.
             /// </summary>
             public float[] DefaultDeadzones => defaultDeadZones.IsValueCreated ? defaultDeadZones.Value : null;
+
+            /// <summary>
+            /// Axis origins for each axis. Can be null if origins have not been found.
+            /// </summary>
+            public float[] AxisOrigins => axisOrigins.IsValueCreated ? axisOrigins.Value : null;
 
             /// <summary>
             /// The <see cref="Guid"/> for this <see cref="JoystickDevice"/>.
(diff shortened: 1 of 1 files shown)


osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
```diff
@@ -220,12 +231,25 @@ public void UpdateRawState(osuTK.Input.JoystickState rawState)
...
+                        float origin = 0;
+
+                        if (Precision.AlmostEquals(axisValue, 0, 0.5f))
+                            origin = 0;
+                        else if (axisValue > 0)
+                            origin = 1;
+                        else if (axisValue < 0)
+                            origin = -1;
+                        else
+                            Logger.Log("No origin value was somehow determined", level: LogLevel.Important);
```
frenzibyte, line 247: that's never the case?

osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
```diff
@@ -107,6 +107,11 @@ public OsuTKJoystickState(JoystickDevice device)
...
                 {
                     var value = device.RawState.GetAxis(i);
+                    var origin = device.AxisOrigins?[i] ?? 0;
+
+                    // Rescale and flip for off-center origin
+                    if (origin != 0)
+                        value = (origin - value) / (2 * Math.Sign(origin));
 
                     // do not allow a deadzone below float_epsilon
                     var deadzone = MathF.Max(device.DefaultDeadzones?[i] ?? 0, Precision.FLOAT_EPSILON);
```
frenzibyte, line 117: Sounds like the axis value retrieval logic (from line 109 till 124) should be moved to its own private method `getAxisValue(JoystickDevice device, int axis)`.

frenzibyte (changes requested):
Not entirely sure about this, code suggestions aside. I think it's fine to have given that it'll fix the whole issue with axis triggers having different origins (@GSculerlor please check if this fixes your aforementioned Xbox 360 controller issue).

osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
Jeremiidesu: Oh - my mistake, I intended to remove it before pushing but seems like I forgot. Thanks for pointing it out!

osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
Jeremiidesu: Not too sure if it's worth moving to a separate method given that it's the only place where that logic is used, and isn't too long either - as far as I can tell, the adjusted value is used everywhere else.

frenzibyte: `getHatButtons()` already exists and is even shorter than this; when you read the whole logic as "save current values to state", which is what it is doing in general:
image 

it would make more sense to put this in its own method instead.

Jeremiidesu: Ah, fair. I'll do that then.

osu.Framework/Input/Handlers/Joystick/OsuTKJoystickHandler.cs
```diff
@@ -155,6 +143,28 @@ private IEnumerable<JoystickButton> getHatButtons(JoystickDevice device, int hat
...
+                var origin = device.AxisOrigins?[axis] ?? 0;
+
+                // Rescale and flip for off-center origin
+                if (origin != 0)
+                    value = (origin - value) / (2 * Math.Sign(origin));
+
+                // do not allow a deadzone below float_epsilon
+                var deadzone = MathF.Max(device.DefaultDeadzones?[axis] ?? 0, Precision.FLOAT_EPSILON);
+
+                if (Precision.AlmostEquals(value, 0, deadzone))
```
peppy, line 159: this should be removed (see https://github.com/ppy/osu-framework/pull/3869)

peppy (changes requested):

peppy:
I really don't know if this is the correct direction. Sure it may work for gamepads, but it's not a universal solution and is adding more local assumptions.

If we are to do this, though, it 100% needs tests written since the math logic here is hard to parse, and is not enough to let me know the assumptions and expectations of this change. I'd hope unit tests (not visual tests) can cover each case, showing what the raw readouts should be giving for each of the three origin decisions.

Also, we should probably rename this from "raw" to "normalised" or "transposed" and also expose the raw values for consumers that don't want this post-processing.

Jeremiidesu:
Another idea I had would have been to implement per-controller-type joystick handler but there doesn't seem to be a way to accurately detect what controller has been connected across OSes.

Currently, the first 8 characters of the GUID are vendor and product code concatenated, so that should help, but as far as I can remember, at least for the Dualshock 4, that part of the GUID is different between Windows and Manjaro Linux, and I don't believe that a lot of specific handlers is something that would be accepted. 

That could separate some of the assumptions this PR makes as it is a valid concern (might close and start from scratch depending on where this ends up going).

peppy:
Honestly I would prefer to keep things simple where possible. If this change is going to cover the average cases in a way joysticks can be consumed in the way you/other framework consumers need I think it's fine.

But we do need test coverage.

peppy:
Replaced by #3844 for the time being. We can reconsider this if the need comes up in the future.

PR #3848, opened 2020-08-26
Add support for seeking a non-loaded animation

### LINKED ISSUE #3847: Seeking a non-loaded animation throws a NullReferenceException
```
[runtime] 2020-08-26 11:05:26 [error]: System.NullReferenceException: Object reference not set to an instance of an object.
[runtime] 2020-08-26 11:05:26 [error]: at osu.Framework.Graphics.Transforms.Transformable.get_Time()
[runtime] 2020-08-26 11:05:26 [error]: at osu.Framework.Graphics.Animations.AnimationClockComposite.consumeClockTime()
[runtime] 2020-08-26 11:05:26 [error]: at osu.Framework.Graphics.Animations.AnimationClockComposite.set_PlaybackPosition(Double value)
[runtime] 2020-08-26 11:05:26 [error]: at osu.Framework.Graphics.Animations.AnimationClockComposite.Seek(Double time)
[runtime] 2020-08-26 11:05:26 [error]: at osu.Framework.Graphics.Animations.Animation`1.GotoFrame(Int32 frameIndex)
```

The code that throws this looks pretty inconspicuous:
```csharp
if (lightContainer != null)
{
    if (isHitting.NewValue)
        Column.TopLevelContainer.Add(lightContainer);
    else
        Column.TopLevelContainer.Remove(lightContainer);

    if (light is TextureAnimation lightAnimation)
        lightAnimation.GotoFrame(0);
}
```
The issue is that `GotoFrame()` throws the nullref in the case that the animation is not loaded, which requires the above code to be transformed into:
```csharp
if (lightContainer != null)
{
    if (isHitting.NewValue)
    {
        Column.TopLevelContainer.Add(lightContainer);

        if (light is TextureAnimation lightAnimation)
            lightAnimation.GotoFrame(0);
    }
    else
        Column.TopLevelContainer.Remove(lightContainer);
}
```

bdach:
Resolves #3847 / usable with ppy/osu#9985 (although not necessarily a hard dependency)

# Summary

Setting `PlaybackPosition` before the animation was loaded could result in a `NullReferenceException` due to not necessarily having received a clock in `Time` yet.

As proposed in the issue, resolve this by not consuming time in the setter of `PlaybackPosition` if the animation isn't yet loaded -`LoadComplete()` will do it anyway before the update loop starts to run.

# Remarks

I opted not to add the guard to `consumeClockTime` itself as other call sites should be safe and should actually throw if the operation doesn't succeed. The other call sites are:

* `Clock.set` - should have access to `Time` by the virtue of actually setting the animation's clock,
* `LoadComplete()` - should have correct clock in place already,
* `Update()` - has no right to run if the drawable isn't loaded.

Test cases added (both for seeking to start, as well as arbitrary frame).

### DIFF
--- a/osu.Framework.Tests/Visual/Sprites/TestSceneAnimation.cs
+++ b/osu.Framework.Tests/Visual/Sprites/TestSceneAnimation.cs
@@ -24,6 +24,9 @@ namespace osu.Framework.Tests.Visual.Sprites
         private TestAnimation animation;
         private Container animationContainer;
 
+        [Resolved]
+        private FontStore fontStore { get; set; }
+
         [SetUpSteps]
         public void SetUpSteps()
         {
@@ -232,7 +235,7 @@ namespace osu.Framework.Tests.Visual.Sprites
         {
             AddStep("load animation", () =>
             {
-                animationContainer.Child = animation = new TestAnimation(startFromCurrent)
+                animationContainer.Child = animation = new TestAnimation(startFromCurrent, fontStore)
                 {
                     Loop = false,
                 };
@@ -260,27 +263,16 @@ namespace osu.Framework.Tests.Visual.Sprites
         {
             public const int LOADABLE_FRAMES = 72;
 
-            [Resolved]
-            private FontStore fontStore { get; set; }
-
             public int FramesProcessed;
 
-            public TestAnimation(bool startFromCurrent)
+            // fontStore passed in via ctor to be able to test scenarios where an animation
+            // already has frames before load
+            public TestAnimation(bool startFromCurrent, FontStore fontStore)
                 : base(startFromCurrent)
             {
                 Anchor = Anchor.Centre;
                 Origin = Anchor.Centre;
-            }
 
-            protected override void DisplayFrame(Texture content)
-            {
-                FramesProcessed++;
-                base.DisplayFrame(content);
-            }
-
-            [BackgroundDependencyLoader]
-            private void load()
-            {
                 for (int i = 0; i < LOADABLE_FRAMES; i++)
                 {
                     AddFrame(new Texture(fontStore.Get(null, (char)('0' + i))?.Texture.TextureGL)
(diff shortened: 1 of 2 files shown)


peppy (approved):
Looks logical. I think there's likely many remaining cases we haven't covered in terms of animation usability. One test + fix at a time!

osu.Framework.Tests/Visual/Sprites/TestSceneAnimation.cs
```diff
@@ -228,11 +231,26 @@ public void TestGotoZeroOnFirstFrameVisible()
             AddAssert("animation restarted from 0", () => animation.PlaybackPosition < 1000);
         }
 
+        [TestCase(0)]
+        [TestCase(48)]
+        public void TestGotoFrameBeforeLoaded(int frame)
+        {
+            AddStep("create new animation", () => animation = new TestAnimation(true, fontStore)
```
frenzibyte, line 238: Too late here but probably worth testing `startAtCurrentTime = false`, which is same behaviour actually.
```suggestion
        [Test]
        public void TestGotoFrameBeforeLoaded([Values(0, 48)] int frame, [Values(true, false)] bool startAtCurrentTime)
        {
            AddStep("create new animation", () => animation = new TestAnimation(startAtCurrentTime, fontStore)
```

frenzibyte (changes requested):

osu.Framework.Tests/Visual/Sprites/TestSceneAnimation.cs
bdach: It doesn't really impact the fix in any capacity. `startAtCurrentTime` just determines whether to add the time consumed in `loadComplete()` to the animation's manual clock, so I'd say this would be superfluous.

frenzibyte: It kinda does, it's a scenario that hasn't been tested yet as far as I know (check `hasSeeked` bool), and this fix is allowing the scenario to be tested easily (previously you'll need to async load then seek), so probably worth covering it.

PR #3850, opened 2020-08-29
Remove obsolete members

frenzibyte-alt:
Would appreciate a double check on `BindableListTest` changes as there are actually a good amount of asserts on the obsolete events there along with the new one. (I've removed what's under the obsolete warning suppress and any assertion that uses a variable declared inside).

Also looks like `IBindableList` hasn't marked `ItemsAdded` and `ItemsRemoved` as obsolete, thus `GlobalStatisticsDisplay` and `Dropdown` not updated to use `CollectionChanged` instead. I think it's fine to remove the obsolete events? not sure... (currently removed them)


smoogipoo (approved):
Seems fine

PR #3860, opened 2020-09-02
Fix Game.Dispose potentially ending before children have been (async) disposed

peppy:
Wait for global async disposal queue to empty before disposing the game itself, as the components within may be relying on game/host access in their ongoing background functions (ie. if they are still in an async load cancelling state).

Fixes test failures seen in https://github.com/ppy/osu/pull/10026, that started to show up from various usage of `DelayedLoadUnloadWrapper` (which is using the async disposal queue in a slightly different way since https://github.com/ppy/osu-framework/pull/3825/files).

I also fixed an edge case in `AsyncDisposalQueue` which can't be easily tested but should be easy enough to understand:

- Enqueue an item
- Processing begins on that item
- New item enqueued before processing ends. Task isn't started because one is already running, but that task will not process any newly added items.

I've fixed this by making the task loop until the queue empties, and also locking over these checks to ensure safety.

### DIFF
--- a/osu.Framework.Tests/Platform/ComponentAsyncDisposalTest.cs
+++ b/osu.Framework.Tests/Platform/ComponentAsyncDisposalTest.cs
@@ -1,0 +1,116 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Threading;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using osu.Framework.Allocation;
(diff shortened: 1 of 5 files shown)


peppy:
Actually I can probably make a test to cover that failure case of the queue itself if that's seen as important. Involves smashing the queue with new items for a good few seconds then ensuring the processed count is 100% of queued items. I'd guess a few would be missed with the old logic.

peppy:
On writing a test I realise the original logic actually works correctly. Misread the task state check condition, which is explicitly checking *Not Running*. I'll revert the changes for now I think.

smoogipoo (approved):
lgtm! :+1:

PR #3864, opened 2020-09-04
Change disposal order to better protect against async disposal queue failures

peppy:
Previous change (#3860) was not enough to always fix the issue (happened again [here](https://ci.appveyor.com/project/peppy/osu/builds/35023352/tests)).

I was able to reproduce the failure locally, and this more solid fix has not failed over 1000+ test runs.

### DIFF
--- a/osu.Framework/Game.cs
+++ b/osu.Framework/Game.cs
@@ -352,7 +352,7 @@ namespace osu.Framework
         protected override void Dispose(bool isDisposing)
         {
             // ensure any async disposals are completed before we begin to rip components out.
-            // if we were to not wait, async disposals may through unexpected exceptions.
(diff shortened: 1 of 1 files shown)


osu.Framework/Game.cs
```diff
@@ -351,6 +351,13 @@ public void Exit()
 
         protected override void Dispose(bool isDisposing)
         {
+            // ensure any async disposals are completed before we begin to rip components out.
+            // if we were to not wait, async disposals may through unexpected exceptions.
```
smoogipoo, line 355:
```suggestion
            // if we were to not wait, async disposals may throw unexpected exceptions.
```

smoogipoo (changes requested):

smoogipoo (approved):

Issue #3644, opened 2020-06-22
[ExcludeFromDynamicCompile] doesn't apply to other types in file


smoogipoo:
`RoslynTypeReferenceBuilder` has two main modes of excluding types:
1. All types inheriting from `Game`.
2. All types with the `[ExcludeFromDynamicCompile]` attribute attached.

If either of the two conditions is trigger, the entire file itself should not be recompiled along with any partial files (`typeReference.Symbol.DeclaringSyntaxReferences`).

PR #3865, opened 2020-09-04
Exclude entire files from dynamic compilation

smoogipoo:
Fixes https://github.com/ppy/osu-framework/issues/3644

`[ExcludeFromDynamicCompile]` should affect the entire rest of the file.


frenzibyte:
Sounds like it closes https://github.com/ppy/osu-framework/issues/3644

bdach (approved):
seems to work as expected :+1:

PR #3866, opened 2020-09-05
Added more Game.cs documentation.

rubiksmaster02:


peppy:
This looks to be done using automated tooling. The changes are wrong and the additions are mostly misleading or outright incorrect.

rubiksmaster02:
what?

rubiksmaster02:
I literally did this by hand, hold up i can get proof

osu.Framework/Game.cs
```diff
@@ -25,27 +25,45 @@ namespace osu.Framework
 {
     public abstract class Game : Container, IKeyBindingHandler<FrameworkAction>, IHandleGlobalKeyboardInput
     {
+        /// <summary>
+        /// Gets this <see cref="Host"/>'s window.
+        /// </summary>
         public IWindow Window => Host?.Window;
 
+        /// <summary>
+        /// Gets the current ResourceStore{T}.
```
peppy, line 34: What makes this "current"?

osu.Framework/Game.cs
```diff
@@ -25,27 +25,45 @@ namespace osu.Framework
...
+        /// </summary>
         public ResourceStore<byte[]> Resources { get; private set; }
 
+        /// <summary>
+        /// Gets this <see cref="TextureStore"/>.
```
peppy, line 39: This is pointless. If we're adding xmldoc it
- doesn't need to say "gets this"
- should probably explain what the property is used for.

osu.Framework/Game.cs
```diff
@@ -25,27 +25,45 @@ namespace osu.Framework
...
+        /// <summary>
+        /// Gets this <see cref="GameHost"/>.
+        /// </summary>
         protected GameHost Host { get; private set; }
 
         private readonly Bindable<bool> isActive = new Bindable<bool>(true);
 
         /// <summary>
-        /// Whether the game is active (in the foreground).
+        /// Gets whether the game is active (in the foreground).
```
peppy, line 51: Unnecessary

osu.Framework/Game.cs
```diff
@@ -25,27 +25,45 @@ namespace osu.Framework
...
         /// </summary>
         public IBindable<bool> IsActive => isActive;
 
+        /// <summary>
+        /// Get this <see cref="AudioManager"/>.
```
peppy, line 56: what is "this"? it doesn't make sense.

osu.Framework/Game.cs
```diff
@@ -25,27 +25,45 @@ namespace osu.Framework
...
         public AudioManager Audio { get; private set; }
 
+        /// <summary>
+        /// Get this <see cref="ShaderManager"/>.
+        /// </summary>
         public ShaderManager Shaders { get; private set; }
 
         /// <summary>
-        /// A store containing fonts accessible game-wide.
+        /// Gets a store containing fonts accessible game-wide.
```
peppy, line 66: unnecessary

osu.Framework/Game.cs
```diff
@@ -64,8 +85,15 @@ public abstract class Game : Container, IKeyBindingHandler<FrameworkAction>, IHa
 
         private LogOverlay logOverlay;
 
+        /// <summary>
+        /// Gets the <see cref="content"/>.
```
peppy, line 89: content is private. this is incorrect as it should not reference a private variable. also it's pointless.

osu.Framework/Game.cs
```diff
@@ -64,8 +85,15 @@ public abstract class Game : Container, IKeyBindingHandler<FrameworkAction>, IHa
...
+        /// </summary>
         protected override Container<Drawable> Content => content;
 
+        /// <summary>
+        /// Creates a new <see cref="UserInputManager"/>.
```
peppy, line 94: this should never be called by the user. it's only for overriding purposes. taht's the kind of thing xmldoc should contain.

osu.Framework/Game.cs
```diff
@@ -113,6 +141,11 @@ public virtual void SetHost(GameHost host)
 
         private DependencyContainer dependencies;
 
+        /// <summary>
+        /// Creates a <see cref="DependencyContainer"/> with children.
```
peppy, line 145: it doesn't "create" more than once. misleading and incorrect (what children?)

peppy (changes requested):

peppy:
If this was not a troll pull request hopefully the above is enough to explain.

PR #3868, opened 2020-09-05
Typo fixes & Math.Clamp

Poliwrath:
Judging by https://github.com/ppy/osu/commit/da41b70d438108d09b058bdc05b8f6f93c8d1f19, replacing MathHelper with Math seems fine.

### DIFF
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -200,7 +200,7 @@ namespace osu.Framework.Graphics
 
         /// <summary>
         /// Whether this Drawable is fully loaded.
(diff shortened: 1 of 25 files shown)


osu.Framework/Graphics/Drawable.cs
```diff
@@ -1531,12 +1531,12 @@ internal set
         internal virtual Drawable Original => this;
 
         /// <summary>
-        /// True iff <see cref="CreateProxy"/> has been called before.
+        /// True if <see cref="CreateProxy"/> has been called before.
```
frenzibyte, line 1534: https://en.wikipedia.org/wiki/If_and_only_if, the same applies to the rest.

frenzibyte (changes requested):

smoogipoo (approved):

PR #3870, opened 2020-09-06
Ensure ThreadedTaskScheduler eventually disposes

peppy:


osu.Framework/Threading/ThreadedTaskScheduler.cs
```diff
@@ -87,7 +94,7 @@ public void Dispose()
             tasks.CompleteAdding();
 
             foreach (var thread in threads)
-                thread.Join();
+                thread.Join(TimeSpan.FromSeconds(10));
```
bdach, line 97: I'm not actually sure this is what we want, if we want to stop the thread after 10 seconds. I'm pretty sure `Thread.Join` does not actually kill at the timeout, it just bails if 10 seconds pass and returns `false`. What we probably *would* want is this:

```csharp
            foreach (var thread in threads)
            {
                bool shutDownGracefully = thread.Join(TimeSpan.FromSeconds(10));
                if (!shutDownGracefully)
                    thread.Abort();
            }
```

...except for the fact that Linux doesn't support it...

bdach: Although now that I look at it the change in 9227e13c sorta works that way, although in a not entirely intuitive manner (at least to me). What it does is disposes the task collection from under the running threads, so they catch the exception while trying to consume the blocking collection and then bail. As long as they are guaranteed to receive the exception while waiting on an item, I suppose this is an acceptable solution.

peppy: We don't care about the thread stopping. Also, thread.Abort is not a safe or supported operation. The idea here is to allow the game process to stop, which will bring all background threads to a halt at an OS level.

peppy: See https://docs.microsoft.com/en-us/dotnet/api/system.threading.thread.abort?view=netcore-3.1#:~:text=and%20later%20and%20.-,NET%20Core.,block%20unless%20ResetAbort%20is%20called. for pretty good explanation of why `Thread.Abort` is no longer a thing.

bdach: Okay, yeah, that does sound like a mess. I've never done much multithreading with raw `Thread`s contrary to TPL, so TIL.

bdach:
Stressed the test by sticking a `[Repeat(100)]` on it, got no fails; tested with my hang scenario on a framework checkout - seems like this alleviates the most pressing issue of the game not closing. :+1: 

Still probably more work required game-side to perhaps allow cancellation of beatmap conversion and fix the actual conversion bug itself.

bdach (approved):

smoogipoo (approved):

PR #3873, opened 2020-09-07
Add support for approximating B-splines

Karoo13:
generalized ApproximateBezier to clamped uniform B-splines


Karoo13:
regarding the change of `count` to `n+1`, most mathematical discussion of splines and related algorithms index this way, with n+1 control points indexed 0 to n. in this context, count is equally valid, but the number n is typically more convenient when doing other spline evaluation and manipulation algorithms. if the physicality of the number of points is of concern, think of n as the number of segments, which is equally physical and often advantageous when multiple splines are chained together.

smoogipoo (approved):
I think this can go in pretty harmlessly. I've tested that it doesn't break osu!-side diffcalc.

PR #3874, opened 2020-09-08
Update Windows BASS libraries

bdach:
Should eventually resolve ppy/osu#9952, after `NativeLibs` and framework bumps.

# Summary

This PR bumps the version of the Windows BASS binaries from 2.4.15.2 to 2.4.15.25. The binaries were provided by un4seen [in this forum message](https://www.un4seen.com/forum/?topic=19062.msg133542#msg133542). The expected sha256 checksums are:

```
791b30b2f5694447757ef48e178cdab53fdbca02e94ec165abe8e582a8da1a2c  win-x64/native/bass.dll
4d2138911d9b47d35494c1d696e655bb7b55e7fadee62b1ffd11c73de8cef54b  win-x86/native/bass.dll
```

The bumped binaries contain a fix for BASS's thread-global mutation of SSE instruction flags, which caused the failure in the aforementioned issue game-side. (See linked issue and forum thread for full paper trail.)

I've tested this twice (today and yesterday, full setup procedure from scratch from game-side `master`) against the game and registered no further mutation of MXCSR after any of the BASS P/Invokes we do through ManagedBass, but as the procedure to test locally is quite involved due to having to pack `NativeLibs` manually, add a local nuget source, then reference it against framework and *only then* do a local checkout, it might be good to have a second pair of eyes test this with the editor reproduction scenario.

# Remarks

Not 100% sure I should be PRing this, but the release schedule for an actual full minor release seems pretty long (2.4.15.0 came out back in December last year), so given the linked issue is milestoned I figured I might as well, at least for visibility.

Unix shared objects are unchanged due to (a) not being provided, and (b) not reproducing the issue in the first place.

I've also updated the x86 binary, because why not at this point if I'm doing this. We're not using it anyway, and from what I gather it wasn't affected by the issue in the first place.

Because the FP flags were being changed globally, this might actually silently fix a whole lot of bugs we might not even have been aware of, especially in single-thread mode.


peppy:
As an aside, you can test this by just dropping the bass.dll in place after building using a file operation, rather than bothering with nuget.

peppy (approved):
Resolves issue at my end 👍

PercyDan54:
Issue also resolved here after replacing bass.dll.
Thanks.
图片

PR #3885, opened 2020-09-14
Fix ChildrenOfType<> early exit for matching composite types

### LINKED ISSUE #3871: ChildrenOfType<> test extension exits too early for matching types
The early break causes children to not be traversed
https://github.com/ppy/osu-framework/blob/07a103d6c0897237235041cec00230f0f6e07210/osu.Framework/Testing/TestingExtensions.cs#L19-L22

```csharp
[Test]
public void Test()
{
    Container container = null;

    AddStep("create children", () =>
    {
        Child = container = new Container { Child = new Box() };
    });

    AddAssert("ChildrenOfType returns 2 children", () => container.ChildrenOfType<Drawable>().Count() == 2);
}
```

UselessToucan:
resolves #3871

### DIFF
--- a/osu.Framework.Tests/Visual/Testing/TestSceneTestingExtensions.cs
+++ b/osu.Framework.Tests/Visual/Testing/TestSceneTestingExtensions.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Linq;
+using NUnit.Framework;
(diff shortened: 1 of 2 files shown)


frenzibyte:
Can just be shortened to:
```csharp
if (drawable is T)
    yield return drawable;

if (drawable is CompositeDrawable composite)
{
     foreach (var child in composite.InternalChildren)
     {
           foreach (var found in child.ChildrenOfType<T>())
               yield return found;
     }
}
```

frenzibyte (approved):

smoogipoo (approved):

PR #3886, opened 2020-09-15
Workaround a BASS issue with parsing WAV files

Jan4V:
This PR tries to propose a workaround to a BASS issue that can be observed when trying to play WAV files that contain no audio data (eg. are silent). This can be observed on [this map (https://osu.ppy.sh/beatmapsets/855677)](https://osu.ppy.sh/beatmapsets/855677) for example. It uses such sounds as the "sliderslide" hitsounds.

When parsing these files, BASS violates the specification by handling zero sized sections as chunks spanning the rest of the file. For this reason, it tries to parse following chunks (in this case the ID3 tags for example) as audio, generating a severely unpleasant sound (which in this case is made even more unpleasant by the fact that the sliding hitsound is looped).

There are 2 good ways to solve this in my opinion:

1. Get this fixed in BASS - Unfortunately, it seems to me that it hasn't been maintained/updated in a while (as the latest version is from almost a year ago). Additionally, I can't seem to find how I would even go about reporting this to them, perhaps commercial customers such as osu! have a way to do that? It also seems as though this is a regression of some sort, as stable doesn't seem to exhibit this issue, and it may be possible that this is actually intended for some other user of the library dealing with invalid WAV files which require this quirk.
2. Use the workaround I proposed in this PR - It is pretty simple, and not very likely to ever break, as the specification hasn't changed since 1991, and is very unlikely to ever change. I made sure to make it as clear as possible to see what is going on.

### DIFF
--- a/osu.Framework/Audio/Sample/SampleStore.cs
+++ b/osu.Framework/Audio/Sample/SampleStore.cs
@@ -44,7 +44,43 @@ namespace osu.Framework.Audio.Sample
                 if (!sampleCache.TryGetValue(name, out Sample sample))
                 {
                     byte[] data = store.Get(name);
-                    sample = sampleCache[name] = data == null ? null : new SampleBass(data, PendingActions, PlaybackConcurrency);
+
+                    // Parse WAV files to workaround BASS issue with empty WAV files
+                    // (WAV header magic == 0x46464952)
+                    if (data != null && data.Length > 0 && BitConverter.ToInt32(data, 0x00) == 0x46464952)
+                    {
+                        int position = 0x0C;
+                        // iterate over the WAV sections
+                        while (position < data.Length - 0x08)
+                        {
+                            // parse the section header
+                            var magic = BitConverter.ToInt32(data, position + 0x00);
+                            var size = BitConverter.ToInt32(data, position + 0x04);
+
+                            // look for the "data" section ("data" header magic == 0x61746164)
+                            if (magic == 0x61746164)
+                            {
+                                // if the size of the section is zero, pass no data to BASS, so the sample is empty
+                                if (size == 0)
+                                    sample = sampleCache[name] = new SampleBass(Array.Empty<byte>(), PendingActions, PlaybackConcurrency);
+
+                                break;
+                            }
+
+                            // skip the current section, as it's not "data"
+                            position += size;
+                            // in the case of odd-sized sections, there is an additional unaccounted byte at the end of the data that also needs to be skipped
+                            position += size % 2;
+                            // the size field doesn't take the header into account, skip that as well
+                            position += 0x08;
+                        }
+
+                        // in case we haven't set the sample as empty, let BASS handle it as usual
+                        if(sample == null)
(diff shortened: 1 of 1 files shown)


bdach:
We were already aware of this issue; see https://github.com/ppy/osu/issues/9939 for paper trail.

The workaround we intended to apply is to not load samples that are too small (less than 1024 bytes in size). While that might not be the prettiest resolution it's what stable does, so it's the ideal resolution from a backwards compatibility standpoint.

It is not true that BASS is unmaintained; see the forum at https://www.un4seen.com/. We've literally just reported and got a resolution to another bug in BASS over the last month (see #3874 and associated issue). I am strongly against rolling any manual parsing of WAV files framework-side and therefore don't think this is the correct direction to take.

Jan4V:
That's fair, wasn't aware of that issue. Do you believe the best would be to try to get the underlying issue I found in BASS fixed there, or would it be preferred to just replicate stables workaround?

bdach:
Both would probably be best, actually, as long as BASS really does violate the spec.

Jan4V:
Perhaps it's worth noting that it is actually possible to reproduce this bug in stable as well (for example, using this file: [https://gofile.io/d/Ao5i7U](https://gofile.io/d/Ao5i7U)). So I'm not sure if it's worth implementing that as a solution to this issue. My workaround handles this fine, although it does have the obvious issue of adding "WAV parsing" into the framework. It seems like the only desirable solution would be to in fact get this fixed in BASS.

bdach:
> So I'm not sure if it's worth implementing that as a solution to this issue. 

It wouldn't be a solution to the issue of BASS not adhering to spec. It would be bringing back legacy code paths for backwards compatibility reasons, as stated above. That's why I explicitly said that having both resolutions would be ideal.

> My workaround handles this fine, although it does have the obvious issue of adding "WAV parsing" into the framework. It seems like the only desirable solution would be to in fact get this fixed in BASS.

Well, yes. If we have a library for doing audio, why would we do our own custom audio on top of that? Not to mention that if a fix does make it upstream, all consumers of the library will benefit. That is the proper operating procedure in open source.

Jan4V:
> Well, yes. If we have a library for doing audio, why would we do our own custom audio on top of that? Not to mention that if a fix does make it upstream, all consumers of the library will benefit. That is the proper operating procedure in open source.

True, when working on this I wasn't aware BASS development was actually active, and it is not in fact open source, so I thought this would be the best place to fix this. I have raised the issue to them on their forum, so we will see if it's intended, or if they are going to fix it.

peppy:
For reference, can you link where this was reported?

Joehuu:
https://github.com/ppy/osu-framework/pull/3886#issuecomment-692817599:
> https://www.un4seen.com/forum/?topic=19075.0

bdach:
@Jan4V are you planning to PR the updated binaries supplied by un4seen, or should I?

Jan4V:
Sorry, I wasn't home, glad it got resolved.

PR #3887, opened 2020-09-16
Add constraint-based Assert model support

UselessToucan:
This PR allows us to write asserts in our Visual tests using [NUnit's constraint model](https://docs.nunit.org/articles/nunit/writing-tests/assertions/assertion-models/constraint.html).

### DIFF
--- a/osu.Framework.Tests/Visual/Bindables/TestSceneBindableAutoUnbinding.cs
+++ b/osu.Framework.Tests/Visual/Bindables/TestSceneBindableAutoUnbinding.cs
@@ -39,24 +39,24 @@ namespace osu.Framework.Tests.Visual.Bindables
 
             AddStep("attempt value transfer", () => drawable1.Bindable.Value = 10);
 
-            AddAssert("transfer 1-2 completed", () => drawable1.Bindable.Value == drawable2.Bindable.Value);
-            AddAssert("transfer 1-3 completed", () => drawable1.Bindable.Value == drawable3.Bindable.Value);
-            AddAssert("transfer 1-4 completed", () => drawable1.Bindable.Value == drawable4.Bindable.Value);
+            AddAssert("transfer 1-2 completed", () => drawable1.Bindable.Value, Is.EqualTo(drawable2.Bindable.Value));
+            AddAssert("transfer 1-3 completed", () => drawable1.Bindable.Value, Is.EqualTo(drawable3.Bindable.Value));
+            AddAssert("transfer 1-4 completed", () => drawable1.Bindable.Value, Is.EqualTo(drawable4.Bindable.Value));
 
             AddStep("expire child 4", () => drawable4.Expire());
 
             AddStep("attempt value transfer", () => drawable1.Bindable.Value = 20);
 
-            AddAssert("transfer 1-2 completed", () => drawable1.Bindable.Value == drawable2.Bindable.Value);
-            AddAssert("transfer 1-3 completed", () => drawable1.Bindable.Value == drawable3.Bindable.Value);
-            AddAssert("transfer 1-4 skipped", () => drawable1.Bindable.Value != drawable4.Bindable.Value);
+            AddAssert("transfer 1-2 completed", () => drawable1.Bindable.Value, Is.EqualTo(drawable2.Bindable.Value));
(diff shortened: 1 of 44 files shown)


bdach:
I'm going to be brutally honest, I don't think I see the point. Some of the usages read actively worse or overly verbose to me compared to the existing code.

osu.Framework.Tests/Visual/Bindables/TestSceneBindableAutoUnbinding.cs
```diff
@@ -39,24 +39,24 @@ public void TestBindableAutoUnbindingAssign()
 
             AddStep("attempt value transfer", () => drawable1.Bindable.Value = 10);
 
-            AddAssert("transfer 1-2 completed", () => drawable1.Bindable.Value == drawable2.Bindable.Value);
-            AddAssert("transfer 1-3 completed", () => drawable1.Bindable.Value == drawable3.Bindable.Value);
-            AddAssert("transfer 1-4 completed", () => drawable1.Bindable.Value == drawable4.Bindable.Value);
+            AddAssert("transfer 1-2 completed", () => drawable1.Bindable.Value, () => Is.EqualTo(drawable2.Bindable.Value));
```
bdach, line 42: To expound on the above, this sort of thing shouldn't be a thing. The `() => Is.EqualTo()` is not only more verbose, it also *adds* complexity because the constraint model is not suited for deferred execution. What this is, is essentially a brand new rake to step on.

bdach (changes requested):

osu.Framework.Tests/Visual/Bindables/TestSceneBindableAutoUnbinding.cs
UselessToucan: I agree that it does not work well with deferred execution. What about cases where it is not required like `Is.EqualTo(2)` or `Is.Null`?

peppy: Coming from a different angle here: the main issues with the visual testing setup which need fixing are:

- when debugging, exceptions are thrown away from the condition (making it hard to track)
- for things like equality comparisons, a failure does not list the values, meaning you need to manually investigate.

I'd be more inclined to take on a new system if this can be improved upon.

bdach: Still opposed to those, but I suppose it's a matter of personal preference as it reads awkwardly to me.

That said I consider this particular issue a blocker. It's way too easy to misuse this API and have code that *looks* reasonable but actually does the wrong thing. *Especially in a test.* When reviewing a PR, I don't want to think about if this given variable in this test requires deferred execution or not.

peppy: I probably shouldn't have replied in this particular conversation thread. I haven't actually checked the code or the issue pointed out by you (which seems 100% valid), just stating my issues with our current assertion logic, which should be prioritised as being fixed, rather than just changing syntactical sugar to match the latest craze.

bdach: Yeah, for the record I agree wholeheartedly. This might have come out awkwardly as we both happened to be typing out our earlier comments at roughly the same time.

bdach:
As a part of ongoing efforts to reduce the number of stale open pull requests, I am going to close this pull request for the following reasons:

* As pointed out above, the constraint model does not play well with deferred execution, and therefore makes test code more confusing to make sense of.
* Lack of general support for the direction of changes suggested here.
* Lack of updates for almost a year.
* Large amount of merge conflicts remaining.

If you still wish to push for this change to be made, the deferred execution blocker would have to be resolved, most likely via refactoring the testing setup to not require it.

I would however recommend looking at other pain points of our current testing setup that have been listed [in this comment](https://github.com/ppy/osu-framework/pull/3887#discussion_r489389316), if you're so inclined.

bdach:
Closing this also sparked [the following discord conversation](https://discord.com/channels/188630481301012481/589331078574112768/850331875377545256), which might be useful to read through if you still wish to continue with this direction.

PR #3889, opened 2020-09-19
Handle exit shortcut via platform action

bdach:
Resolves ppy/osu#3483.

# Breaking changes

## `Game` now implements `IKeyBindingHandler<PlatformAction>` by default

To resolve a bug causing invalid keypress handling of <kbd>Alt</kbd>+<kbd>F4</kbd> on Windows, exiting the game has been made a `PlatformAction`. All subclasses of `Game` that previously implemented `IKeyBindingHandler<PlatformAction>` should now override `On{Pressed,Released}(PlatformAction)` and delegate to `Game`'s base implementation to allow the game to close via that shortcut.

# Summary

The apparent binding conflict described in the issue was only reproducible under certain conditions. To get the clash, the following sequence of key events must fire:

1. <kbd>Alt</kbd> down, <kbd>F4</kbd> down (in any order)
2. <kbd>Alt</kbd> up
3. <kbd>F4</kbd> up

Swapping steps 2 and 3 will not cause the issue to reproduce any more. The reason why this particular order is necessary is that `WindowsGameWindow` was doing a local implementation of the shortcut:

https://github.com/ppy/osu-framework/blob/a68e97bc4d191ca64f5252894dd3254d035c5c45/osu.Framework/Platform/Windows/WindowsGameWindow.cs#L22-L26

Because the game swallows the shortcut, the framework input handlers never see it, and therefore the key binding handler's `Modifiers` flag has no effect, as it essentially sees nothing followed by a <kbd>F4</kbd> keypress.

Resolve by moving the local implementation to the base game, leveraging a newly added global action. The reason why `Host.Window?.Close()` is used instead of `Host.Exit()` is that the latter is an unconditional exit that cannot be blocked by the game.

# Remarks

Not sure how you'll see this resolution. I suppose this could be also fixed by just removing the `return` to allow the game to see the input. The upside to this resolution is that it's cleaner from a code standpoint in my view, but the downside is that a clumsy user that wants to handle a platform action in `Game` could forget to call `base` and therefore break their game closing.

This resolution also opens a way of potentially adding a similar key binding for other platforms, although not sure what that would be (<kbd>Ctrl</kbd>+<kbd>Q</kbd>?), so I only moved the Windows one locally for now.

I thought about adding a test for this, but it's a little weird. A headless test does not work, because it has no window, so it would rightly do nothing. I could either add a path to unconditionally exit the host in that case (seems weird to do just for a test), or run a normal game host (has not been done before from what I can see, would require calling `Host.GetSuitableHost()` outright to work portably). I'll add one on request.


frenzibyte (dismissed):

peppy (approved):
Works well for me. Let's get this in.

peppy:
Interesting to note: this doesn't work with SDL (never gets hit). Probably something we need to fix, as we won't be able to intercept exit with SDL enabled. It doesn't work on master either, so this is not a regression.

bdach:
I'll have a look at what SDL is doing in this regard. It might be something like SDL intercepting it and making an `SDL_QUIT` event out of it, or a window close event.

PR #3894, opened 2020-09-23
Make Dispose() synchronous again

smoogipoo:
Continuing off from https://github.com/ppy/osu/pull/10200, disposal doesn't seem to be a huge issue - most of the time is taken up by `UnbindAllBindings` which runs synchronously anyway.


bdach:
Not sure I'll be able to meaningfully comment of the validity of (a)syncness of disposal when it comes to the overarching design, but as for the failing test:

- The failing test is setting up a drawable to be during load when it's being removed from the hierarchy:

  https://github.com/ppy/osu-framework/blob/8a60e3fa3d4cd4561f3307a678c2cfcca44387a2/osu.Framework.Tests/Containers/TestSceneLoadComponentAsync.cs#L45

- The test is then supposed to wait some and then check that `composite` is not disposed:

  https://github.com/ppy/osu-framework/blob/8a60e3fa3d4cd4561f3307a678c2cfcca44387a2/osu.Framework.Tests/Containers/TestSceneLoadComponentAsync.cs#L47-L51

  The test fails on the first assert, but that's actually a red herring. The `Clear` call actually blocks until `composite` is disposed (which it can't be, as it's blocked on a load). The async-loading `composite` throws an exception after 10 secs:

  https://github.com/ppy/osu-framework/blob/8a60e3fa3d4cd4561f3307a678c2cfcca44387a2/osu.Framework.Tests/Containers/TestSceneLoadComponentAsync.cs#L135-L140

  which makes the load terminate, therefore allowing disposal to continue and finish up, failing the asserts.

The fact that there was a test for this is unnerving, as things could get potentially deadlock-y if we have consumers (the game, in particular) relying on this behaviour.

bdach (changes requested):
as above

smoogipoo:
Even without deadlocking, it could manifest by disposing a drawable that's in the process of loading a large online resource...

Not sure about this change anymore. May be best to expose something synchronous (`UnbindAllBindables`) for specific cases when you need to drop resources immediately, and call it a day.

peppy:
Bumping this in priority (and because I just replicated this precise branch at https://github.com/ppy/osu-framework/compare/master...peppy:remove-async-disposal?expand=1).

This is the cause of https://github.com/ppy/osu/issues/12426, which is not easily fixable locally.

The proposed direction is to also make all `IResourceStores` use `async` flow, add `CancellationToken` to their parameters and [turn on SynchronizationContexts](https://github.com/ppy/osu-framework/pull/4149).

I'll look at this later this week.

PR #3895, opened 2020-09-23
Reduce TextureStore lock contention

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/10229 (otherwise osu! will crash after exiting from gameplay)

This is pretty visible by scrolling through beatmap listing with https://github.com/ppy/osu-framework/pull/3894.

The contention occurs between `Get()` and `Purge()`, especially when online resources are involved.

Another level of locking has been added to ensure that only one retrieval is happening concurrently (preserves current behaviour) and that the lock on `textureCache` is taken for as short of a period as possible.

### DIFF
--- a/osu.Framework/Graphics/Textures/TextureStore.cs
+++ b/osu.Framework/Graphics/Textures/TextureStore.cs
@@ -15,6 +15,7 @@ namespace osu.Framework.Graphics.Textures
     public class TextureStore : ResourceStore<TextureUpload>
     {
         private readonly Dictionary<string, Texture> textureCache = new Dictionary<string, Texture>();
+        private readonly object retrievalLock = new object();
 
         private readonly All filteringMode;
         private readonly bool manualMipmaps;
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Textures/TextureStore.cs
```diff
@@ -109,26 +110,41 @@ public virtual Texture Get(string name, WrapMode wrapModeS, WrapMode wrapModeT)
...
-                if (!textureCache.TryGetValue(key, out var tex))
+                if (textureCache.TryGetValue(key, out var tex))
+                    return tex;
+            }
+
+            // Take an exclusive lock on retrieval of the texture.
+            lock (retrievalLock)
+            {
+                // If another retrieval of the texture happened before us, we should check if the texture exists in the cache again.
+                lock (textureCache)
```
bdach, line 124: I think the contention here could be reduced further here if this used a `ReaderWriterLockSlim` (although I'm not sure if this was done this way on purpose). Something like

```csharp
lock.EnterReadMode();
// query cache
lock.ExitReadMode();

// on cache miss
lock.EnterWriteMode();
// re-query cache, early return if found
// otherwise fetch and store texture
lock.ExitWriteMode();
```

This way readers shouldn't block each other on the cache query the way they do right now, but they'll get blocked during a write, and writers won't write until all reads have concluded (but they'll get priority on the lock over new readers). I'd suggest taking an upgradeable read lock, but apparently that one is exclusive so it'd essentially become a mutex again...

peppy: i'm all for using the `ReaderWriterLockSlim` pattern if only to follow a known standard (if we can)

smoogipoo: Not sure if overengineering, but I've made the requested change. Please check because it's also my first time using RWLS.

osu.Framework/Graphics/Textures/TextureStore.cs
```diff
@@ -109,26 +118,61 @@ public virtual Texture Get(string name, WrapMode wrapModeS, WrapMode wrapModeT)
...
+                if (textureCache.TryGetValue(key, out var tex))
+                    return tex;
+            }
+            finally
+            {
+                textureLock.ExitReadLock();
+            }
+
+            // Take an exclusive lock on retrieval of the texture.
+            lock (retrievalLock)
```
peppy, line 135: is this still required?

smoogipoo: Yes, we still want only one thread to retrieve textures at the same time.

bdach: Entering writer mode is blocked if the lock is taken by anyone else in any other mode, so I think this should just take the write lock instead of using `retrievalLock`. Unless there's an issue with that that I'm missing...?

smoogipoo: The point is to not take a huge lock around everything during the actual texture retrieval process, which can take seconds at a time.
The contention point is specifically with `Purge()`.

bdach: Purge is taking the upgradeable lock which as I mentioned before is a straight mutex, but I can see why this would be required - we don't want possibility of races around disposal to avoid some unlucky reader getting a disposed texture... Now I'm not convinced whether this was worth it anymore.

smoogipoo: Well `Purge` can just as well use read + write lock rather than upgradeable. Makes no difference as far as it's concerned, but I thought the "mutex" part of upgradeable was only mutex as long as it's both read-write, and not just read?

Idk, as I mentioned before, I would consider this overengineering. This is not a high-concurrency scenario.

bdach: Although then again could try taking write lock to check cache in purge, and only take the write lock if the texture to be disposed is in there. Should be safe and maybe keep the contention down a little, but then again I should have probably checked the characteristics of this before suggesting.

peppy:
Looks sane 👍 .

Weird because I thought we already addressed this in a similar way before...

peppy (approved):

PR #3896, opened 2020-09-24
Allow all types of steps to be considered setup steps

peppy:
Previously, the ability to mark steps as setup was tied to the `SetUpStepButton` class. This meant that `UntilStep`s were not considered setup steps even when contained within `[SetUpSteps]`, causing the visual test browser to not automatically run them.

Fixes the most annoying case on osu!-side with "exit all screens" would be run and not run any of the actual test code.

This also removes the public `AddSetupStep` method because we aren't using it (and I don't think anyone should be, anyway).

Tested to not regress any osu!-side tests.


osu.Framework/Testing/TestScene.cs
```diff
@@ -292,27 +292,11 @@ private void runNextStep(Action onCompletion, Action<Exception> onError, Func<St
 
         public void AddStep(StepButton step) => schedule(() => StepsContainer.Add(step));
 
-        public StepButton AddSetupStep(string description, Action action)
```
smoogipoo, line 295: Maybe keep this an obsolete it? Seems like there's at least a few usages: https://github.com/search?l=C%23&q=AddSetupStep&type=Code

smoogipoo (changes requested):

osu.Framework/Testing/TestScene.cs
peppy: only one of those is valid. if you really insist i'll do it but just seems unnecessary

peppy: Added back

smoogipoo (approved):

PR #3900, opened 2020-09-27
SDL Joystick Handling

swoolcock:
Implements joystick handling with the SDL window backend.  Buttons and axes are mapped as close to osuTK as possible.

This PR is primarily for parity with osuTK.  In the future we will want to identify individual controllers, and potentially identify buttons based on logical meaning rather than button number (for example, show `Button1` as `X` if an Xbox controller is connected, or `Square` if a DualShock is connected).

Differences to osuTK:
* Triggers range from 0 to 1 rather than -1 to 1.  This is how SDL reports it, and I see the -1 to 1 range as a design flaw in osuTK.
* ~~Unmapped button presses will be ignored.  This is to avoid problems with controllers that send multiple button events for a single press, or use buttons as flags.  For example, the Xbox Elite Series 2 sends two different button indices with each of the ABXY buttons, and it uses buttons 35 and 36 to indicate the current profile selection and whether the trigger locks are enabled.  The downside is that extra buttons (such as the Xbox Elite paddles or the DualShock touch pad) cannot be used.~~

- [x] Dependent on #3911


osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -339,11 +345,53 @@ private static DisplayMode displayModeFromSDL(SDL.SDL_DisplayMode mode, int disp
             return new DisplayMode(SDL.SDL_GetPixelFormatName(mode.format), new Size(mode.w, mode.h), bpp, mode.refresh_rate, modeIndex, displayIndex);
         }
 
+        private void enqueueJoystickAxisInput(int instanceID, SDL.SDL_GameControllerAxis gcAxis, JoystickAxisSource axisSource, short axisValue)
+        {
+            var clamped = Math.Clamp((float)axisValue / short.MaxValue, -1f, 1f);
```
peppy, line 350: according to the PR description, should this one 0..1?

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -525,22 +573,94 @@ private void handleTouchFingerEvent(SDL.SDL_TouchFingerEvent evtTfinger)
 
         private void handleControllerDeviceEvent(SDL.SDL_ControllerDeviceEvent evtCdevice)
         {
+            switch (evtCdevice.type)
+            {
+                case SDL.SDL_EventType.SDL_CONTROLLERDEVICEADDED:
+                    var controller = SDL.SDL_GameControllerOpen(evtCdevice.which);
+                    var instanceID = SDL.SDL_JoystickGetDeviceInstanceID(evtCdevice.which);
+                    axisDirectionButtons[instanceID] = new JoystickButton[(int)JoystickAxisSource.AxisCount];
```
peppy, line 581: i'm confused why there's three dictionaries here. if it's all related data, why not make a class for it?

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -339,11 +345,53 @@ private static DisplayMode displayModeFromSDL(SDL.SDL_DisplayMode mode, int disp
             return new DisplayMode(SDL.SDL_GetPixelFormatName(mode.format), new Size(mode.w, mode.h), bpp, mode.refresh_rate, modeIndex, displayIndex);
         }
 
+        private void enqueueJoystickAxisInput(int instanceID, SDL.SDL_GameControllerAxis gcAxis, JoystickAxisSource axisSource, short axisValue)
+        {
+            var clamped = Math.Clamp((float)axisValue / short.MaxValue, -1f, 1f);
+            var value = Math.Abs(clamped) < deadzone_threshold ? 0 : Math.Sign(clamped) * (Math.Abs(clamped) - deadzone_threshold) / (1f - deadzone_threshold);
```
peppy, line 351: is this doing a mapping to -1..1? i can't see where this would be happening, am i missing something?

peppy (changes requested):

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
swoolcock: SDL reports an axis value between 0 and 32767 for triggers, so it doesn't matter.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
swoolcock: I can do that.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
swoolcock: This does similar to osuTK, where the `deadzone_threshold` is used as the new zero, rather than just dropping any values less than the threshold.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
peppy: i'm confused.  is there a reason to have it clamp between -1..1 if it never goes below 0?

bdach: [From `SDL_GameControllerGetAxis`](https://wiki.libsdl.org/SDL_GameControllerGetAxis):

> The state is a value ranging from -32768 to 32767. Triggers, however, range from 0 to 32767 (they never return a negative value). 

So it sounds like actual positional axes still go -1..1, but "button-like" axes (triggers, etc.) go 0..1.

swoolcock: If `axisValue` is -32768, `(float)axisValue / short.MaxValue` will be less than -1.

peppy: Aha, that's what I was missing. I'm just trying to figure out how to see this 0..1 thing now.

peppy: Tested with switch pro controller, xbox one and joystick with 64 buttons 3 hats 6 axes. all work as expected  👍 .

swoolcock: Note that Xbox Elite Series 2 does not 100% work (on macOS, at least), as SDL does not have correct mappings for it.  This is fixed in a PR I made to the SDL_GameControllerDB project, so it's something we may want to look at later.

peppy: As already discussed, we shouldn't need to provide support for proprietary stuff. As long as generic HID works there's no issue.

swoolcock: The problem is that SDL *does* detect it as a "controller", but as an Elite Series 1.  The mapping is therefore incorrect.
I think we can safely assume it will be fixed as the SDL team adds more controller mappings.

peppy: The point is we're trying to get SDL implemented. Not add support for one controller that likely only you care about.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -339,11 +343,53 @@ private static DisplayMode displayModeFromSDL(SDL.SDL_DisplayMode mode, int disp
...
+
+            if (!controllers.TryGetValue(instanceID, out var state))
+                return;
+
+            var directionButton = state.AxisDirectionButtons[(int)axisSource];
```
peppy, line 354: split `(int)axisSource` out into a local `int index`, it's used 5+ times

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -339,11 +343,53 @@ private static DisplayMode displayModeFromSDL(SDL.SDL_DisplayMode mode, int disp
...
+            var negativeButton = JoystickButton.FirstAxisNegative + (int)axisSource;
+            var positiveButton = JoystickButton.FirstAxisPositive + (int)axisSource;
+
+            if (value == 0 && directionButton != 0 || value < 0 && directionButton == positiveButton || value > 0 && directionButton == negativeButton)
```
peppy, line 358: this is basically impossible to read

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -339,11 +343,53 @@ private static DisplayMode displayModeFromSDL(SDL.SDL_DisplayMode mode, int disp
...
+                if (gcAxis == SDL.SDL_GameControllerAxis.SDL_CONTROLLER_AXIS_TRIGGERLEFT)
+                    eventScheduler.Add(() => OnJoystickButtonUp(new JoystickButtonInput(JoystickButton.GamePadLeftTrigger, false)));
+                else if (gcAxis == SDL.SDL_GameControllerAxis.SDL_CONTROLLER_AXIS_TRIGGERRIGHT)
+                    eventScheduler.Add(() => OnJoystickButtonUp(new JoystickButtonInput(JoystickButton.GamePadRightTrigger, false)));
+
+                eventScheduler.Add(() => OnJoystickButtonUp(new JoystickButtonInput(directionButton, false)));
+                state.AxisDirectionButtons[(int)axisSource] = 0;
+            }
+
+            if (value < 0 && directionButton != negativeButton)
```
peppy, line 369: as above, can't really follow this logic

peppy (changes requested):

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -717,6 +848,154 @@ private MouseButton mouseButtonFromEvent(byte button)
...
+                    return JoystickAxisSource.Axis5;
+
+                case SDL.SDL_GameControllerAxis.SDL_CONTROLLER_AXIS_TRIGGERRIGHT:
+                    return JoystickAxisSource.Axis6;
+            }
+        }
+
+        protected class ControllerState
+        {
+            public readonly int InstanceID;
```
bdach, line 936: This seems to be never read - was there intent behind it? Also not sure if this was intended to be public API, but maybe not too relevant for now.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -717,6 +848,154 @@ private MouseButton mouseButtonFromEvent(byte button)
...
+                    return;
+
+                ButtonBindings = Enumerable.Range(0, (int)SDL.SDL_GameControllerButton.SDL_CONTROLLER_BUTTON_MAX)
+                                           .Select(i => SDL.SDL_GameControllerGetBindForButton(ControllerHandle, (SDL.SDL_GameControllerButton)i)).ToArray();
+
+                AxisBindings = Enumerable.Range(0, (int)SDL.SDL_GameControllerAxis.SDL_CONTROLLER_AXIS_MAX)
+                                         .Select(i => SDL.SDL_GameControllerGetBindForAxis(ControllerHandle, (SDL.SDL_GameControllerAxis)i)).ToArray();
+            }
+
+            public SDL.SDL_GameControllerButton GetButtonForIndex(byte index)
```
bdach, line 976: also unused, unsure of purpose

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -339,11 +343,64 @@ private static DisplayMode displayModeFromSDL(SDL.SDL_DisplayMode mode, int disp
             return new DisplayMode(SDL.SDL_GetPixelFormatName(mode.format), new Size(mode.w, mode.h), bpp, mode.refresh_rate, modeIndex, displayIndex);
         }
 
+        private void enqueueJoystickAxisInput(int instanceID, SDL.SDL_GameControllerAxis gcAxis, JoystickAxisSource axisSource, short axisValue)
+        {
+            var clamped = Math.Clamp((float)axisValue / short.MaxValue, -1f, 1f);
+            var value = Math.Abs(clamped) < deadzone_threshold ? 0 : Math.Sign(clamped) * (Math.Abs(clamped) - deadzone_threshold) / (1f - deadzone_threshold);
```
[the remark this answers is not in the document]
swoolcock: I disagree with pulling a single line of code out into its own method when it’s used exactly once.  Maybe just a one line comment explaining what the code does?

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -339,11 +343,64 @@ private static DisplayMode displayModeFromSDL(SDL.SDL_DisplayMode mode, int disp
...
+            if (!controllers.TryGetValue(instanceID, out var state))
+                return;
+
+            var index = (int)axisSource;
+
+            // determine which directional buttons are used for this axis
+            var negativeButton = JoystickButton.FirstAxisNegative + index;
+            var positiveButton = JoystickButton.FirstAxisPositive + index;
+            var currentButton = state.AxisDirectionButtons[index];
+            var expectedButton = value < 0 ? negativeButton : value > 0 ? positiveButton : 0;
```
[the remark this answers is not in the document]
swoolcock: As above, I think this could just be clarified with a one line comment rather than making a short “only used once” helper method.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
swoolcock: This is SDL’s internal ID for joystick.  It’s what’s used as the key for the dictionary.  Technically this could be removed, but only because we (currently) already have the instance ID at the places we need to use it.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
swoolcock: Originally used in the same fashion as `GetAxisForIndex`, currently unused due to the second note in the PR description regarding “flag” button presses being detected.  In retrospect, I’m thinking it may be better to detect those too (using this method) which will be equivalent to osuTK.

peppy (changes requested):

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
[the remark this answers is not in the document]
peppy: what's the reasoning against splitting out into a method? i agree it would make the code more legible.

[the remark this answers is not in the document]
swoolcock: I can change it to a helper method then, no problem.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
swoolcock: Changed to no longer drop unmapped buttons (as such, this method is now used).

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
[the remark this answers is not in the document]
swoolcock: Changed, as below.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -717,6 +869,154 @@ private MouseButton mouseButtonFromEvent(byte button)
...
+
+                case SDL.SDL_GameControllerAxis.SDL_CONTROLLER_AXIS_RIGHTY:
+                    return JoystickAxisSource.Axis5;
+
+                case SDL.SDL_GameControllerAxis.SDL_CONTROLLER_AXIS_TRIGGERRIGHT:
+                    return JoystickAxisSource.Axis6;
+            }
+        }
+
+        protected class ControllerState
```
peppy, line 955: should be in its own file

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -24,13 +25,16 @@ public class Sdl2WindowBackend : WindowBackend
     {
         private const int default_width = 1366;
         private const int default_height = 768;
+        private const float deadzone_threshold = 0.075f;
```
peppy, line 28: I don't know where to start with reviewing this class.

All joystick / controller logic should be moved either into another class. If making a new class for this is too hard to manage, then make it a partial class `Sdl2WindowBackend_Joystick` or something.

peppy (changes requested):

swoolcock:
Note: Currently blocked by #3911 as it will make it easier to centralise the joystick logic.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -598,22 +616,91 @@ private void handleTouchFingerEvent(SDL.SDL_TouchFingerEvent evtTfinger)
...
         private void handleJoyDeviceEvent(SDL.SDL_JoyDeviceEvent evtJdevice)
         {
+            switch (evtJdevice.type)
+            {
+                case SDL.SDL_EventType.SDL_JOYDEVICEADDED:
+                    var instanceID = SDL.SDL_JoystickGetDeviceInstanceID(evtJdevice.which);
+
+                    // if the joystick is already opened, ignore it
+                    if (controllers.ContainsKey(instanceID))
+                        break;
```
bdach, line 669: Out of curiosity, is there a reason why the joystick function calls have guards for double open/close but the controller ones don't? I can't see anything in the docs about that.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -789,6 +881,89 @@ private MouseButton mouseButtonFromEvent(byte button)
             }
         }
 
+        private JoystickButton joystickButtonFromGameController(SDL.SDL_GameControllerButton button)
```
bdach, line 884: Maybe move these mappings out to an extension method or something? Might keep the class a little leaner (already nearing on 1k lines as is).

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
swoolcock: SDL will only fire controller events if the joystick is recognised as a mapped controller. It will fire joystick events in both cases, so we need to make sure it doesn’t handle the events twice.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
swoolcock: I had thought of moving this to `Sdl2Extensions`, actually. Will do so today.

peppy:
Code looks sane. Will do a final check with my army of controllers tonight and merge pending success.

peppy:
I've noticed a couple of issues which need further investigation before this is merged:

- Starting the game with a non-zeroed axis causes the deadzone to become insanely huge. This is an issue for controllers which don't reset to zero when not interacted with (ie. throttles)



- Positive axis maxes at 0.9999835 for me. Need to confirm whether this is an issue with this PR or the controller's output:



Note how negative values always reach -1 as expected, but positive never do.

bdach:
For what it's worth I can hit +/-1 on axes (testing on ubuntu, with an xbox 360 gamepad). Can't seem to get any funny thing to happen with the deadzones, either, but that controller has no stationary axes - tried to emulate by holding down trigger during game open but it also looks correct as far as I can tell.

swoolcock:
@peppy which controller are you testing with?  I'm guessing it's one that doesn't show up in SDL's controller mappings, thus it will be handled solely by the joystick API.  It's likely that the deadzone/positive axis problems are sourced directly from SDL, so I might need to put some debug-mode logging for the raw values in the SDL_JoyAxisEvent.

peppy:
Okay never mind, I forgot the SDL switch. Now I'm getting hard crashes.

```csharp
Unhandled exception. System.NullReferenceException: Object reference not set to
an instance of an object.
   at osu.Framework.Platform.Sdl.Sdl2ControllerBindings.GetAxisForIndex(Byte ind
ex) in C:\Users\Dean\Projects\osu-framework\osu.Framework\Platform\Sdl\Sdl2Contr
ollerBindings.cs:line 64
   at osu.Framework.Platform.Sdl.Sdl2WindowBackend.handleJoyAxisEvent(SDL_JoyAxi
sEvent evtJaxis) in C:\Users\Dean\Projects\osu-framework\osu.Framework\Platform\
Sdl\Sdl2WindowBackend.cs:line 723
   at osu.Framework.Platform.Sdl.Sdl2WindowBackend.processEvents() in C:\Users\D
ean\Projects\osu-framework\osu.Framework\Platform\Sdl\Sdl2WindowBackend.cs:line
552
   at osu.Framework.Platform.Sdl.Sdl2WindowBackend.Run() in C:\Users\Dean\Projec
ts\osu-framework\osu.Framework\Platform\Sdl\Sdl2WindowBackend.cs:line 447
   at osu.Framework.Platform.Window.Run() in C:\Users\Dean\Projects\osu-framewor
k\osu.Framework\Platform\Window.cs:line 321
```

peppy:
image

peppy (approved):

PR #3901, opened 2020-09-27
Extract numeric type check to utils class from BindableNumber

frenzibyte:
Required in another PR involving arithmetic operation between generic types, therefore I need to use this method there to shorten stuff and not copy-paste.

### DIFF
--- a/osu.Framework/Utils/Validation.cs
+++ b/osu.Framework/Utils/Validation.cs
@@ -26,19 +26,7 @@ namespace osu.Framework.Utils
         public static bool IsFinite(MarginPadding toCheck) => float.IsFinite(toCheck.Top) && float.IsFinite(toCheck.Bottom) && float.IsFinite(toCheck.Left) && float.IsFinite(toCheck.Right);
 
         /// <summary>
-        /// Returns whether the provided generic type <typeparamref name="T"/> is a numeric type, i.e. one of the following types:
-        /// <list type="bullet">
-        /// <item><description><see cref="sbyte"/></description></item>
-        /// <item><description><see cref="byte"/></description></item>
-        /// <item><description><see cref="short"/></description></item>
-        /// <item><description><see cref="ushort"/></description></item>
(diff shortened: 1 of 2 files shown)


osu.Framework/Utils/Validation.cs
```diff
@@ -23,5 +24,34 @@ public static class Validation
...
+        /// <item><description><see cref="short"/></description></item>
+        /// <item><description><see cref="ushort"/></description></item>
+        /// <item><description><see cref="int"/></description></item>
+        /// <item><description><see cref="uint"/></description></item>
+        /// <item><description><see cref="long"/></description></item>
+        /// <item><description><see cref="ulong"/></description></item>
+        /// <item><description><see cref="float"/></description></item>
+        /// <item><description><see cref="double"/></description></item>
+        /// </list>
+        /// </summary>
```
bdach, line 42: why?

look at:

* the method name
* the method implementation

and rethink whether this xmldoc is necessary

bdach (changes requested):

osu.Framework/Utils/Validation.cs
frenzibyte: the class is public for framework users, therefore xmldocs must be 100% filled I believe, and mostly for the part that there are types not in the check like `decimal` for example, better show it to the user by just hovering on the method instead of making them disassemble the framework package or otherwise.

bdach: then maybe list the exception instead of literally *everything else*, making the xmldoc longer than the method implementation itself?

frenzibyte: sure I guess

bdach: also this should probably *not* be public, while we're at it

frenzibyte: yeah got concerned right after I said it, alright will make it `internal`.

frenzibyte:
No longer needed.

PR #3902, opened 2020-09-27
Fix offset-based position transform not respecting previous transforms

### LINKED ISSUE #3821: MoveToOffset() does not respect previous position transforms on all cases
Realized it on #3817, probably unnoticed due to its lack of usage.

Since `MoveToOffset()` is an "offset transformation", the end value depends on the current one, the way it's currently done is by checking the last existing `Position` transform:
https://github.com/ppy/osu-framework/blob/a949b10f85ade1137b63e61cbb48bb32aa781834/osu.Framework/Graphics/TransformableExtensions.cs#L666

which would not work on all cases, such as using `MoveTo[X,Y]()` transforms before instead:
```csharp
            boxTest(box =>
            {
                box.MoveToY(0.25f, 250);
                box.Delay(500).MoveToOffset(new Vector2(0, 0.25f), 250);
            });

            checkAtTime(0, box => box.Y == 0);
            checkAtTime(250, box => box.Y == 0.25f);
            checkAtTime(500, box => box.Y == 0.25f);
            checkAtTime(750, box => box.Y == 0.5f);
```
or not adding the transforms in order:
```csharp
            boxTest(box =>
            {
                box.Delay(500).MoveToOffset(new Vector2(0, 0.25f), 250);
                box.MoveToOffset(new Vector2(0, 0.25f), 250);
            });

            checkAtTime(0, box => box.Y == 0);
            checkAtTime(250, box => box.Y == 0.25f);
            checkAtTime(500, box => box.Y == 0.25f);
            checkAtTime(750, box => box.Y == 0.5f);
```

(above cases can be testable my inserting them into `TestSceneTransformRewinding` each in a `Test` method).

---

Reporting this for tracking, didn't wanna miss this one as well.

frenzibyte:
Closes #3821 

- [ ] Depends on #3901

Basically `MoveToOffset` previously calculates the end value based off the current `Position` value **when the transform is first added** (i.e. when `MoveToOffset(...)` call is first executed), which is quite incorrect for cases mentioned in the linked issue (which is added to the test).

Therefore I fixed it by introducing a `TransformOffset` class that calculates the end value by the start value **when the start time of the transform has reached**.

---

# vNext

## `Transform.ReadIntoStartValue()` has been renamed to `ReadValues()`

Renamed as the latest changes were depending on not just reading into `StartValue`, but into `EndValue` as well. See https://github.com/ppy/osu-framework/pull/3902

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkTransform.cs
+++ b/osu.Framework.Benchmarks/BenchmarkTransform.cs
@@ -100,7 +100,7 @@ namespace osu.Framework.Benchmarks
                 throw new NotImplementedException();
             }
 
-            protected override void ReadIntoStartValue(Box d)
+            protected override void ReadValues(Box d)
             {
                 throw new NotImplementedException();
             }
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
@@ -408,6 +408,36 @@ namespace osu.Framework.Tests.Visual.Drawables
             checkAtTime(750, box => box.Y == 0.375f);
         }
 
+        [Test]
+        public void TestMoveToOffsetRespectsRelevantTransforms()
+        {
+            boxTest(box =>
+            {
+                box.MoveToY(0.25f, 250);
+                box.Delay(500).MoveToOffset(new Vector2(0, 0.25f), 250);
+            });
+
+            checkAtTime(0, box => box.Y == 0);
+            checkAtTime(250, box => box.Y == 0.25f);
+            checkAtTime(500, box => box.Y == 0.25f);
+            checkAtTime(750, box => box.Y == 0.5f);
+        }
+
+        [Test]
+        public void TestMoveToOffsetRespectsTransformsOrder()
+        {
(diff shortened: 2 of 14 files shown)


frenzibyte:
Looks like this will be a breaking change for whoever implements the `Transform` class, can't figure out any way to avoid it...

smoogipoo:
This is completely overengineered to the point where I don't know where to start. Why not make a custom transform for this single purpose? Derive `Transform` and set `Position = StartValue + offset` without all the other changes in this PR. Remove the whole `MakeOffsetTransform` thing. Remove the `sum()` thing.

smoogipoo (changes requested):
As mentioned

peppy:
That is what I suggested in discord, for what it's worth..

frenzibyte:
The point of all these changes is about adding `MakeOffsetTransform`, as I thought it would be worth it to allow adding offset transforms to any property / field I want, but actually looking back now, not so sure that it would be a graspable change anymore with the `sum()` method hanging around like this...

Alright will supersede this PR with a local custom position offset transform.

smoogipoo:
Offsetting is not a common thing in general. If it's needed somewhere it can be implemented locally, it's maybe 10 LOC.

PR #3903, opened 2020-09-28
Fix offset-based position transform not respecting previous transforms

### LINKED ISSUE #3821: MoveToOffset() does not respect previous position transforms on all cases

frenzibyte:
Closes #3821
Supersedes #3902 

See https://github.com/ppy/osu-framework/pull/3902#issuecomment-699742964 for reason why superseded.

---

`ReadIntoStartValue()` is added before `Apply()` to fix cases where the position offset transform would be applied without its value being read, therefore leading to incorrect results. (as the commit says https://github.com/ppy/osu-framework/commit/1589926137f8e295bad6923bd939ea8af96d9c43)

### DIFF
--- a/osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
+++ b/osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
@@ -303,6 +303,7 @@ namespace osu.Framework.Graphics.Transforms
 
             foreach (Transform t in toFlush)
             {
+                t.ReadIntoStartValue();
                 t.Apply(t.EndTime);
                 t.OnComplete?.Invoke();
             }
--- a/osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
+++ b/osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
@@ -303,7 +303,12 @@ namespace osu.Framework.Graphics.Transforms
 
             foreach (Transform t in toFlush)
             {
-                t.ReadIntoStartValue();
+                if (!t.HasStartValue)
+                {
+                    t.ReadIntoStartValue();
+                    t.HasStartValue = true;
+                }
+
                 t.Apply(t.EndTime);
                 t.OnComplete?.Invoke();
             }
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
@@ -408,6 +408,36 @@ namespace osu.Framework.Tests.Visual.Drawables
             checkAtTime(750, box => box.Y == 0.375f);
         }
 
+        [Test]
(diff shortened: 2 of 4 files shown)


osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
```diff
@@ -303,6 +303,7 @@ public virtual void FinishTransforms(string targetMember = null)
 
             foreach (Transform t in toFlush)
             {
+                t.ReadIntoStartValue();
```
smoogipoo, line 306: Surround with `if (!t.HasStartValue) { ...; t.HasStartValue = true; }`

smoogipoo (changes requested):

osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
frenzibyte: oh right, I thought the transform would be removed, but that would not be the case for rewinding

smoogipoo (approved):

smoogipoo:
@peppy This seems sane to me but I want your review too since you're more knowledgeable about the inner workings of transforms w.r.t reading of values/applying.

peppy:
I think it seems sane. @smoogipoo have you cross-tested with osu! (especially gameplay/storyboard rewinding/ffwding)?

peppy (approved):

smoogipoo:
Just did, seems fine.

PR #3904, opened 2020-09-28
Remove unnecessary interpolation delegate members

frenzibyte:
Deemed unnecessary after asking what it is for.


smoogipoo (approved):
For context, this is from the time that these methods used to be direct interpolation methods, but these days it goes through an extra level of generic with its own caching.

PR #3905, opened 2020-09-30
Implement SetIconFromStream for SDL

swoolcock:
Implements `DesktopWindow.SetIconFromStream` in a similar fashion to `DesktopGameWindow`.

Windows .ico files can contain multiple sizes, some of which may be stored as raw PNG rather than custom bitmap planes.  On non-Windows platforms, we pass that PNG directly to SDL to use in `SDL_SetWindowIcon`.  Needs testing on Linux.

Due to the way SDL implements `SDL_SetWindowIcon` on Windows, it can look bad when scaling down large images for the "small icon" that Windows expects for the taskbar, etc.  For this reason, we use the Windows API directly as in the existing `DesktopGameWindow` to select both a large and a small size.  I believe the "large" icon is generally used for things like task switching and the volume control panel.

Windows API using 24x24 for the "small icon" (identical to the old implementation):
Windows-Win-API

`SDL_SetWindowIcon` on Windows using the closest PNG to 256x256 in `lazer.ico`:
Windows-SDL_SetWindowIcon

`SDL_SetWindowIcon` on macOS using the closest PNG to 256x256 in `lazer.ico`:
macOS-SDL_SetWindowIcon


osu.Framework/Platform/Windows/WindowsGameWindow.cs
```diff
@@ -26,8 +26,11 @@ public override void SetIconFromStream(Stream stream)
             smallIcon = iconGroup.CreateIcon(24, 24);
             largeIcon = iconGroup.CreateIcon(256, 256);
 
-            SendMessage(WindowInfo.Handle, seticon_message, (IntPtr)0, smallIcon.Handle);
-            SendMessage(WindowInfo.Handle, seticon_message, (IntPtr)1, largeIcon.Handle);
+            if (smallIcon != null)
+                SendMessage(WindowInfo.Handle, seticon_message, (IntPtr)0, smallIcon.Handle);
+
+            if (largeIcon != null)
+                SendMessage(WindowInfo.Handle, seticon_message, (IntPtr)1, largeIcon.Handle);
```
bdach, line 33: Not sure why this logic has the icons null-checked separately while `WindowsDesktopWindow` bails if any one icon is null.

osu.Framework/Platform/DesktopWindow.cs
```diff
@@ -151,6 +157,18 @@ protected override void UpdateWindowMode(WindowMode mode)
...
+
+        internal virtual void SetIconFromGroup(IconGroup iconGroup)
+        {
+            // LoadRawIcon returns raw PNG data if available, which avoids any Windows-specific pinvokes
+            var bytes = iconGroup.LoadRawIcon(default_icon_size, default_icon_size);
+            if (bytes == null)
+                return;
+
+            WindowBackend.SetIcon(Image.Load<Rgba32>(bytes));
+        }
```
bdach, line 170: I'm not really feeling this too much, to be honest. Relying on technical implementation details of a Windows-specific icon format feels a bit too... arcane? Say, a Linux or Mac dev comes in and wants to set their game icon - now they have to download a Windows `.ico` tool and fiddle with it enough to get it to work, making sure the PNG data makes it into the `.ico` somehow.

I'd feel much better about this if it was an either-or thing - if someone wants to go through the effort of supplying Windows `.ico`s for better AA/size-specific font variants, then sure, but if not, then we fall back to an *actual* `.png` file. Other platforms would just use the `.png` directly.

osu.Framework/Platform/WindowBackend.cs
```diff
@@ -29,6 +33,9 @@ public abstract class WindowBackend : IWindowBackend
 
         public virtual IEnumerable<Display> Displays => Enumerable.Empty<Display>();
         public virtual Display PrimaryDisplay => Displays.First();
+        public virtual IntPtr WindowHandle { get; } = IntPtr.Zero;
```
bdach, line 36: Not sure if a nullptr is a sane default. Would probably just make `abstract`.

bdach:
FWIW, I can confirm this works on linux.

osu.Framework/Platform/DesktopWindow.cs
```diff
@@ -151,6 +157,18 @@ protected override void UpdateWindowMode(WindowMode mode)
             ConfineMouseMode.TriggerChange();
         }
 
+        public virtual void SetIconFromStream(Stream stream) => SetIconFromGroup(new IconGroup(stream));
```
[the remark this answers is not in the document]
swoolcock: It’s called from the framework consumer.  Check the osu!-side PR.

osu.Framework/Platform/Windows/WindowsGameWindow.cs
swoolcock: Because in the legacy case, this is the sole implementation of setting the icon (there is no implementation to fall back on).  The null checks are there because `CreateIcon` is now `[CanBeNull]` rather than throwing if a size cannot be found.

osu.Framework/Platform/DesktopWindow.cs
swoolcock: Will change to also accept an image directly.  In our case we will not need to change anything osu!-side, though, since we have a .ico file.

osu.Framework/Platform/DesktopWindow.cs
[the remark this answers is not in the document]
bdach: Ah, fair. Not sure how I glossed over that one...

osu.Framework/Platform/Windows/WindowsGameWindow.cs
bdach: Inline comment to clarify might be good there.

osu.Framework/Platform/DesktopWindow.cs
peppy: Optimally this would be automatically run on startup for windows backends, fetching the icon from the app executable itself, rather than requiring a manual implementation.

swoolcock: @peppy Does osu! stable currently do this?  It would be helpful if I could see the current implementation (unless it's using `System.Drawing.Icon`, in which case we'll have to use the winapi directly).

peppy: Stable uses `Icon.ExtractAssociatedIcon(exeName)`

swoolcock: Is this actually required for standalone executables, though?  Running `osu!.exe --sdl` straight from the latest release has the correct window icon anyway.

peppy: If it already has the correct icon then is this PR needed? Is SDL doing this for us? Does it get the sizing correct (ie. not use a completely incorrect icon size/colour choice from the .ico)?

swoolcock: From what I can see, a default icon is assigned by setting the `hIcon` field in the `WNDCLASS` struct passed to `RegisterClass`.  SDL does not assign this, and according to Microsoft's documentation https://docs.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-wndclassa

> If this member is NULL, the system provides a default icon.

I can only think that Windows is using `SHGetFileInfo` or similar to get a default associated icon for osu!.exe.

If you're fine with relying on that, we can close this PR.  However, this PR will guarantee an icon is assigned on all desktop platforms regardless of how the application is bundled/not bundled.

peppy: Yeah, we may still want this PR.

I guess the next step would be checking on windows, with and without this PR each of the following:

- Displays correctly in taskbar?
- Displays correctly as desktop icon, windows explorer (check the smallest and largest size). The icon should scale and not stop at a certain point like stable currently does (broken):

image

- Displays correctly in start menu pane (check each size)?
- Displays correctly in volume mixer (stretch: fixing windows 7, where lazer [currently balloons out in volume mixer](https://github.com/ppy/osu/issues/2424))

image

- Displays correctly in title bar (24 or 16px whatever the size is, should use the actual matching size in the ico, not scale down) and task manager list view. I think current lazer may be incorrect currently (scaling down:

image

swoolcock: > * Displays correctly in taskbar?
> * Displays correctly in volume mixer (stretch: fixing windows 7, where lazer [currently balloons out in volume mixer](https://github.com/ppy/osu/issues/2424))

Without the PR, these display correctly with either backend when the icon is bundled in the .exe as a Windows resource.  It also displays correctly when running via `dotnet` for osuTK, but not for SDL.  With the PR, these display correctly in all cases.  Only tested in Windows 10, I do not have a Windows 7 or Windows 8 computer handy.

> * Displays correctly as desktop icon, windows explorer (check the smallest and largest size). The icon should scale 
and not stop at a certain point like stable currently does (broken):
> * Displays correctly in start menu pane (check each size)?

Dependent on the icon being bundled in the .exe, unrelated to runtime code.

> * Displays correctly in title bar (24 or 16px whatever the size is, should use the actual matching size in the ico, not scale down) and task manager list view. I think current lazer may be incorrect currently (scaling down:

It was currently using 24px (which gets scaled to 16px automatically).  I have changed this to 16px, but it looks no different.  It appears that this is the actual image in lazer.ico (checked using Icon Explorer, selection background included to contrast the white outline):

Screen Shot 2020-10-02 at 12 17 00 pm

We may need to improve the legibility of the small icons in the resource.

I have also changed it to ensure the it always the takes the highest bit depth available, since it was initially incorrectly choosing 4-bit when requesting 16px.

osu.Framework/Platform/Windows/Native/IconGroup.cs
```diff
@@ -36,75 +38,115 @@ internal struct IconDir
...
+        /// <returns>The index of the icon in the icon directory, or -1 if a valid icon could not be found.</returns>
+        private int findClosestEntry(int width, int height, int bpp, bool requireRawData)
+        {
+            var results = Enumerable.Range(0, iconDir.Count)
+                                    .Where(i => iconDir.Entries[i].Width <= width && iconDir.Entries[i].Height <= height && iconDir.Entries[i].BitCount <= bpp)
+                                    .Where(i => iconDir.Entries[i].HasRawData || !requireRawData)
+                                    .OrderByDescending(i => iconDir.Entries[i].Width)
+                                    .ThenByDescending(i => iconDir.Entries[i].Height)
+                                    .ThenByDescending(i => iconDir.Entries[i].BitCount)
+                                    .Take(1).ToArray();
```
bdach, line 130: a little strange - could just `FirstOrDefault() ?? -1` and save the array alloc

bdach (commented):
seems relatively fine otherwise - not too keen on the actual image -> ICO fallback in `DesktopWindow` but it's probably more hassle than required to completely split them apart

osu.Framework/Platform/Windows/Native/IconGroup.cs
swoolcock: `FirstOrDefault` on int is 0 though, right?

bdach: Ah dammit I forget every time. I guess this is fine then, to prevent boxing every single item in an `int?`...

swoolcock: Could alternatively do something like this, which removes the need for `Enumerable.Range` and frequent array indexing:

```csharp
return iconDir.Entries
              .Where(e => e.Width <= width && e.Height <= height && e.BitCount <= bpp)
              .Where(e => e.HasRawData || !requireRawData)
              .OrderByDescending(e => e.Width)
              .ThenByDescending(e => e.Height)
              .ThenByDescending(e => e.BitCount)
              .Select((_, i) => (int?)i)
              .FirstOrDefault() ?? -1;
```

bdach: I just remembered that `DefaultWithEmpty(-1).First()` is also a possibility.

swoolcock: Will do that instead then.

bdach (dismissed):
looks about as good as can get to me, given that we're trying to handle `.ico` files in a platform-agnostic manner

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -393,6 +449,22 @@ public override void Run()
 
         public override void RequestClose() => eventScheduler.Add(OnCloseRequested);
 
+        public override unsafe void SetIcon(Image<Rgba32> image)
+        {
+            var data = image.GetPixelSpan().ToArray();
```
peppy, line 454: pretty sure the `ToArray` can be optimised out but probably not worth it considering how often this is called.

osu.Framework/Platform/WindowBackend.cs
```diff
@@ -15,6 +17,8 @@ namespace osu.Framework.Platform
     /// </summary>
     public abstract class WindowBackend : IWindowBackend
     {
+        #region Properties
```
peppy, line 20: we generally don't use regions like this. make your IDE do this for you.

peppy (approved):

PR #3906, opened 2020-09-30
Add BindCollectionChanged to IBindableList interface

peppy:
Looks like an oversight.


smoogipoo (approved):

smoogipoo:
What crashed?

peppy:
Nothing, just wanted to use this after `GetBoundCopy()` without having to cast from `IBindableList` to `BindableList`.

peppy:
what the fuck is that ... title

Issue #3518, opened 2020-05-04
Cannot use more than one controller


Jeremiidesu:
When connecting two or more controllers, only the first one will respond to inputs.

smoogipoo:
Joystick or midi controller?

Jeremiidesu:
Joysticks, not MIDI controllers ; updated the title and first comment to reflect this

Jeremiidesu:
Additional stuff I just noticed: 

- The joystick test scene only handles the first connected joystick
- A game project of mine handles input for all connected joysticks as the same joystick.

In any case, not what could be expected from proper multiple joystick support

peppy:
Just to confirm, all your controllers are correctly firing events?

Jeremiidesu:
Yes, they all do fire events, with no way to determine the controller that fired the event.

PR #3911, opened 2020-10-02
Move IInput instantiation responsibility to InputHandlers

swoolcock:
Changes input event arguments to use value types and defers instantiation of `IInput` classes to the currently active `InputHandler`.

This has two main advantages:

1) It decouples the `IWindowBackend` and `Window` classes from the `osu.Framework.Input` namespace and types, and
2) It allows individual `InputHandler`s to configure the `IInput` classes before adding them to `PendingInputs`.  This means that joystick axis deadzone calculations can be done within the `InputHandler` and not be reimplemented per-backend.

~~The only concern I have is that we will need to change the event signature if more or different data is required in future.  It may be beneficial to wrap these in a struct so that only the struct fields need change.  This would be similar to how osuTK currently uses `KeyboardKeyEventArgs` to provide more data than just the `Key`.~~

~~For example, rather than:~~

```csharp
event Action<Key> KeyDown;
```

~~We could provide modifier data if desired by defining:~~

```csharp
public readonly struct KeyInputArgs
{
    public readonly Key Key;
    public readonly int Modifiers;
}

event Action<KeyInputArgs> KeyDown;
```

~~This could however be done at a later date as part of migrating `Window` events and methods to the `IWindow` interface, and not part of this PR.~~

**Edit:** I have implemented this as part of this PR so as to make centralising joystick logic cleaner in #3900


osu.Framework/Platform/InputArgs.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            Button = button;
+            Pressed = pressed;
+        }
+    }
+
+    /// <summary>
+    /// Aggregates the data returned from a MouseMove event.
+    /// </summary>
+    public struct MouseMoveInputArgs
```
smoogipoo, line 34: Separate all of these out into separate files.

I'm also wondering if this is overengineering... What's the purpose of these? The events are all very internal to me.

smoogipoo (changes requested):

osu.Framework/Platform/InputArgs.cs
swoolcock: I plan on introducing a `JoystickAxisInputArgs` etc. type with axis information (including which controller it comes from), but I wanted to make it consistent with the other input-triggered events.

It's quite possible that these types will need to be used in the `osu.Framework.iOS`/`Android` projects too, so they need to be public.

peppy: I'm not really sure how this makes things more consistent. Do you mean in naming? Because as structs, there's no real consistency being added (no inheritance or otherwise).

Looking at things another way: is there a reason you could add a `JoystickAxisInput` or similar (following the existing flow)?

swoolcock: > I'm not really sure how this makes things more consistent. Do you mean in naming? Because as structs, there's no real consistency being added (no inheritance or otherwise).

Partly in naming, partly in the fact that they are structs that can have more data added without needing to changing the event signature.

> Looking at things another way: is there a reason you could add a JoystickAxisInput or similar (following the existing flow)?

`JoystickAxisInput` needs to have deadzone scaling applied before being passed to `PendingInputs`, which means it is better to calculate and instantiate it in the `InputHandler` rather than doing it per-backend (which is what you have requested to change in #3900).

swoolcock: I can revert the struct stuff for now if you'd prefer, I was just trying to avoid having more than one argument in the event's `Action`.  The joystick events will eventually need to include at least two arguments each once we add the joystick number as part of resolving #3518

bdach: To bring up a concrete example, an SDL mouse move event with this PR will result in a conversion chain of

```
SDL.SDLMouseMotionEvent
MouseMoveInputArgs
MousePositionAbsoluteInput
```

For comparison, osuTK does

```
osuTK.Input.MouseEventArgs
OsuTKEventMouseState
MousePositionAbsoluteInput
```

which I guess is not strictly *worse* in the number of conversions but I feel like we should be aiming for less overhead wherever possible.

It's hard for me to recommend something for sure without seeing the end result you'd like to arrive at. Is the deadzone logic shared, or distinct for each backend? If it's the former, then maybe a pure processing component that does the necessary calculations/conversions could save on the extra indirection layer?

(As an aside, if this sticks, these all really ought to be readonly structs.)

swoolcock: > It's hard for me to recommend something for sure without seeing the end result you'd like to arrive at. Is the deadzone logic shared, or distinct for each backend? If it's the former, then maybe a pure processing component that does the necessary calculations/conversions could save on the extra indirection layer?

Backends should preferably not perform any event data massaging and instead provide raw data to the input handler.  This is why I was planning on moving the deadzone logic from `Sdl2WindowBackend` to the upcoming `JoystickHandler` input handler.

> (As an aside, if this sticks, these all really ought to be readonly structs.)

I agree.  If you check the conversion chains above, you'll see that `SDL.SDLMouseMotionEvent` and `MouseMoveInputArgs` are both structs, however their osuTK counterparts are both classes.  Changing the `*Input` classes to structs is a larger change that I would prefer to keep to its own future PR.

smoogipoo: Personally I just don't see a reason for the structs. As I said before, I don't consider these to be public APIs so signatures changes aren't a big deal.

Beyond that you're not gaining anything other than the events looking arguably nicer, but I also question that at the point where you have:
```
struct KeyPressEventArgs
    bool Pressed

event Action<KeyPressEventArgs> KeyDown
event Action<KeyPressEventArgs> KeyUp
```

Would prefer a revert.

peppy (approved):

peppy (approved):

PR #3916, opened 2020-10-04
Expose sample length

bdach:
PRing for potential use with ppy/osu#10358, albeit I'm not 100% convinced it will prove useful just yet.

# Summary

Expose the length of `Sample`s, `SampleChannel`s and `DrawableSample`s to framework consumers. The unit chosen to expose at is milliseconds, for consistency with other members (`Time.Current`, `Track.Length`).

# Remarks

`VirtualSample`s and `SampleChannelVirtual` instances will report a zero length by default.

### DIFF
--- a/osu.Framework/Audio/Sample/SampleChannel.cs
+++ b/osu.Framework/Audio/Sample/SampleChannel.cs
@@ -56,6 +56,11 @@ namespace osu.Framework.Audio.Sample
 
         public override bool IsAlive => base.IsAlive && !Played;
 
+        /// <summary>
+        /// The length of the underlying <see cref="Sample"/>, in milliseconds.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


osu.Framework/Audio/Sample/SampleChannel.cs
```diff
@@ -56,6 +56,11 @@ protected override void UpdateState()
 
         public override bool IsAlive => base.IsAlive && !Played;
 
+        /// <summary>
+        /// The length of the underlying <see cref="Sample"/>, in milliseconds.
+        /// </summary>
+        public double Length => Sample.Length;
```
frenzibyte, line 62: Should also be added to `ISampleChannel`.

peppy (approved):

PR #3917, opened 2020-10-05
Always clear context in DrawThread.Cleanup

### LINKED ISSUE #3879: Changing thread mode while in sdl freezes the game
**Describe the crash:** dont know if this should be a framework issue or not. when osu is run with the --sdl parameters, and changing the thread mode (either single to multi and vise versa) freezes the game while the sound and game still somewhat responds (evident by pressing ctrl + o for settings and the pop up open sound can be heard and pressing p key to go to song select and the song plays for like 3 seconds and crashes immediately). i dont know if this is hardware since when osu is restarted, the game works and even switches the thread to the one you switched (e.g. pre-crash: multi Switched to single, when restarted post-crash: single

**Screenshots or videos showing encountered issue:** 
https://drive.google.com/file/d/1UMt-dDiMl-j5JZsHvddLXp0zrF-joXjm/view?usp=sharing

**osu!lazer version:** 2020.903.0

**Logs:**

[runtime.log](https://github.com/ppy/osu/files/5193965/runtime.log)

**Computer Specifications:** 
Device name	PC
Processor	Intel(R) Pentium(R) CPU        P6200  @ 2.13GHz   2.13 GHz
Installed RAM	6.00 GB (5.80 GB usable)
GPU         First gen intel hd graphics
System type	64-bit operating system, x64-based processor

swoolcock:
Resolves #3879

Ensures that we always clear the context in `DrawThread.Cleanup`.  It was previously skipping this step when running with the SDL backend, causing rendering to stop on Windows.

`MakeCurrent` has been added to `IWindow` to pair with `ClearCurrent`.  Note that it is currently defined as `new` since it conflicts with osuTK's `IGameWindow.MakeCurrent`, which will eventually be removed.


osu.Framework/Platform/GameWindow.cs
```diff
@@ -309,6 +309,10 @@ public virtual void CycleMode()
             WindowMode.Value = currentValue;
         }
 
+        void IWindow.MakeCurrent() => ((IGameWindow)this).MakeCurrent();
+
+        public void ClearCurrent() => GraphicsContext.CurrentContext.MakeCurrent(null);
```
smoogipoo, line 314: This is missing the null check from the old code, is this really ok?

smoogipoo (changes requested):

osu.Framework/Platform/GameWindow.cs
swoolcock: I mean it should be fine, since that null check was only there to test whether we were running the osuTK or SDL backend.  I can put it back in if you'd prefer.

smoogipoo (approved):
Works fine/looks fine

PR #3920, opened 2020-10-09
Add Directory and File Selector Components

LeNitrous:
From the lazer repository which is then abstracted to be available of use to any projects that will need it.


osu.Framework/Graphics/UserInterface/DirectoryCurrentDisplay.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.IO;
+using System.Linq;
+using osu.Framework.Allocation;
+using osu.Framework.Bindables;
+using osu.Framework.Graphics.Containers;
+using osuTK;
+
+namespace osu.Framework.Graphics.UserInterface
+{
+    public abstract class DirectoryCurrentDisplay : CompositeDrawable
```
peppy, line 14: These class names need improvement.

osu.Framework/Graphics/UserInterface/DirectoryDisplayPiece.cs
```diff
@@ -0,0 +1,58 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Allocation;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Sprites;
+using osuTK;
+
+namespace osu.Framework.Graphics.UserInterface
+{
+    public abstract class DirectoryDisplayPiece : CompositeDrawable
```
peppy, line 11: These class names need improvement. Having `DirectoryDisplayPiece` and `DirectoryPiece` is weird. Also I don't believe we use the "Piece" terminology in framework, nor should we.

osu.Framework/Graphics/UserInterface/DirectorySelector.cs
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    directoryFlow.Add(CreateParentDirectoryPiece(CurrentPath.Value.Parent));
+
+                    directoryFlow.AddRange(GetEntriesForPath(CurrentPath.Value));
+                }
+            }
+            catch (Exception)
+            {
+                CurrentPath.Value = directory.OldValue;
+                this.FlashColour(Colour4.Red, 300);
```
peppy, line 92: Probably should be moved to a virtual method with no default effect here.

peppy (changes requested):

peppy:
All public classes, methods and properties need full xmldoc.

osu.Framework/Graphics/UserInterface/BasicFileSelector.cs
```diff
@@ -0,0 +1,63 @@
...
+        protected override ScrollContainer<Drawable> CreateScrollContainer() => new BasicScrollContainer();
+
+        protected override DirectoryListingFile CreateFileItem(FileInfo file) => new BasicFilePiece(file);
+
+        protected override void NotifySelectionError()
+        {
+            this.FlashColour(Colour4.Red, 300);
+        }
+
+        private class BasicFilePiece : DirectoryListingFile
```
peppy, line 27: Did you forget about renaming this one? Also not sure the basic drawables need to be in their own files. Nesting with the related listing component like it's done here should be fine.

peppy (changes requested):

peppy:
Still not 100% sure on the split of abstract vs basic, but also not wanting to spend more time until this sees usage in the real world, at which point we can refactor further.

peppy (approved):

PR #3921, opened 2020-10-09
Implement OnCloseRequested for SDL

swoolcock:
Implements the missing `OnExitRequested` flow.

### DIFF
--- a/osu.Framework.Tests/TestGame.cs
+++ b/osu.Framework.Tests/TestGame.cs
@@ -2,16 +2,22 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using osu.Framework.Allocation;
+using osu.Framework.Bindables;
 using osu.Framework.IO.Stores;
 
 namespace osu.Framework.Tests
 {
+    [Cached]
     internal class TestGame : Game
     {
+        public readonly Bindable<bool> BlockExit = new Bindable<bool>();
+
         [BackgroundDependencyLoader]
         private void load()
         {
             Resources.AddStore(new NamespacedResourceStore<byte[]>(new DllResourceStore(typeof(TestGame).Assembly), "Resources"));
         }
+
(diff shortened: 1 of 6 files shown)


peppy:
This doesn't catch alt-f4 on windows, while our old logic did.

swoolcock:
Ah, I see SDL does not general `SDL_QUIT` events for Alt+F4/Cmd+Q etc.  Will look into it.

swoolcock:
What is the expected functionality of pressing Alt+F4?  Instant quit/window dismissal?

peppy:
the ability to intercept as you've implemented. we need to save changes etc.

swoolcock:
`Window.Close` => `Sdl2WindowBackend.Close` was performing a forceful close rather than requesting one as per osuTK.

Normally Alt+F4 and Cmd+Q both send an `SDL_QUIT` message and safely handle the request flow.  However, recent changes in `WindowGameHost` catch Alt+F4 as a `PlatformAction` and redirect it to `Window.Close`.  On osuTK this was not a problem due to `Window.Close` following the request flow.  I have updated the SDL backend to work in the same way.

swoolcock:
Setting the `SDL_HINT_WINDOWS_NO_CLOSE_ON_ALT_F4` hint prevents SDL from redirecting Alt+F4 to `SDL_QUIT`, solving the problem of a double `OnExitRequested` being called on Windows.

Given we only define the Alt+F4 `PlatformAction.Exit` in `WindowsGameHost`, there is currently no need to use a similar hint or workaround on other desktop platforms.

Tested in Windows 10.

peppy:
Added test coverage. Looks to work as expected.

peppy (dismissed):

peppy (approved):

bdach (approved):

PR #3923, opened 2020-10-10
Avoid using Lazy<> for drawable scheduler to reduce alloc overhead

peppy:
Due to the reference to `this.Clock`, the delegate was being constructed per drawable. Not a huge overhead but I'd rather not see it while profiling (as such large numbers of `Func` usually point to an issue).

image

Benchmarked rebased on #3927:

before:
|              Method |     Mean |   Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------- |---------:|--------:|--------:|------:|------:|------:|----------:|
|   TestEmptyGameLoop | 994.5 us | 2.52 us | 2.36 us |     - |     - |     - |      41 B |
|  TestAllocContainer | 993.4 us | 2.03 us | 1.90 us |     - |     - |     - |    1996 B |
| TestLoadedContainer | 993.0 us | 1.57 us | 1.47 us |     - |     - |     - |    2404 B |
|     TestAllocSprite | 992.3 us | 7.51 us | 7.02 us |     - |     - |     - |    1601 B |
|    TestLoadedSprite | 993.8 us | 1.66 us | 1.56 us |     - |     - |     - |    1853 B |

after:
|              Method |     Mean |   Error |  StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------- |---------:|--------:|--------:|------:|------:|------:|----------:|
|   TestEmptyGameLoop | 992.4 us | 1.41 us | 1.25 us |     - |     - |     - |      41 B |
|  TestAllocContainer | 992.4 us | 2.38 us | 2.23 us |     - |     - |     - |    1706 B |
| TestLoadedContainer | 992.6 us | 1.72 us | 1.61 us |     - |     - |     - |    2132 B |
|     TestAllocSprite | 992.9 us | 2.42 us | 2.26 us |     - |     - |     - |    1484 B |
|    TestLoadedSprite | 992.4 us | 3.00 us | 2.80 us |     - |     - |     - |    1717 B |


osu.Framework/Graphics/Drawable.cs
```diff
@@ -439,13 +438,23 @@ public float Depth
...
         /// </summary>
-        protected internal Scheduler Scheduler => scheduler.Value;
+        protected internal Scheduler Scheduler
+        {
+            get
+            {
+                if (scheduler != null)
+                    return scheduler;
+
+                lock (loadLock)
```
bdach, line 454: Everything else seems fine, but gonna tag in @smoogipoo for an opinion on this - how do you feel about this sharing the load lock? There are cases where this could be bad but @peppy convinced me that we shouldn't bother about it too much - not sure whether you'll see it the same way.

peppy: The only edge case I can think of is the `internal` access to `Scheduler` by the framework, but on a quick check of all internal usages I can't see any of importance.

peppy:
Added benchmark coverage to OP.

osu.Framework/Graphics/Drawable.cs
smoogipoo: The only concerning case is a deadlock where something in BDL awaits on an async task that somehow schedules to the current supertree. I don't know how this could happen in practice though.

Seems fine by me.

smoogipoo (approved):

PR #3924, opened 2020-10-10
Clear reference to dependencies aggressively on CompositeDrawable disposal

peppy:
While we'd like to say that reference leaks will never occur, when they do, having a reference to dependencies still live can cause a runaway scenario where the single reference leak can be holding references to a whole subtree for good.

Adding this as a failsafe avoids such scenarios.

### DIFF
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -296,6 +296,7 @@ namespace osu.Framework.Graphics.Containers
             }
 
             OnAutoSize = null;
+            Dependencies = null;
             schedulerAfterChildren = null;
 
             base.Dispose(isDisposing);


bdach (approved):
seems straightforward enough :+1:

smoogipoo:
Keep in mind that this only delays the problem, and doesn't fix the fact that the original reference is still being held.

I.e. in the case of something like:
```
A     D
|    /
|   /
B  /
| / 
|/
C
```
only the branch A-B is going to be collected, but D-C (the original culprit) is not.

peppy:
The aim here is to fix cases where all drawables holding the reference have been disposed. In the case you mention, something has not been disposed right?

smoogipoo:
This is the case where A disposes, B then C also dispose, but D doesn't dispose because it's at a higher level.

D still holds a reference to C which blocks it from being cleaned up by the GC.

peppy:
.. sure

but this wasn't intended to handle that situation (it's kinda impossible to fix that). this is explicitly the case where drawables are disposed but referenced (the one in your memory dump).

PR #3925, opened 2020-10-10
Implement Window.IsActive for new backend

swoolcock:
Just links `IsActive` with the current window focus.  Tested in osu! by switching focus during gameplay to see the pause menu appear.


peppy:
I've added test coverage. Not sure how you tested.

peppy (approved):

peppy (approved):

PR #3926, opened 2020-10-10
Fix rare memory leak resulting from unsafe DelayedLoadUnloadWrapper event logic

peppy:
As reported by @smoogipoo via memory dump after a long play session.

The causality of this is quite simple to understand:

- `EndDelayedLoad` is run and schedules to `Game.Schedule`
- Before the next update loop, the instance is disposed.
- The dispose logic cleans up the instance's events.
- Next update loop runs the scheduled callback, scheduling `checkForUnload` permanently.

There was no check ensuring the event subscription would not happen after such a scenario.

To fix this, I've taken the absolutely safest path possible: removing all complex async disposal guarding logic and forced the unbind into the synchronous `UnbindAllBindables` call. I propose we do this for all similar cases (low overhead, performance/behaviour critical) rather than trying to protect against the async disposal flow.


smoogipoo (approved):
Looks ok

PR #3929, opened 2020-10-12
Fix LoadComponentAsync enumerating input enumerable multiple times

peppy:
This is a huge oversight: calling `LoadComponentAsync` with a live `IEnumerable` will result in the provided drawable(s) being constructed multiple times. The final returned values are also not the loaded once, but fresh instances.

I've confirmed the test I added behaves per expectations on master / this branch, but I've somehow broken exceptions and reference retention with this change. PRing for feedback.

### DIFF
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -154,6 +154,8 @@ namespace osu.Framework.Graphics.Containers
             loadingComponents ??= new WeakList<Drawable>();
 
             var loadables = components.ToArray();
+            if (loadables.Length == 0)
+                return Task.CompletedTask;
 
             foreach (var d in loadables)
             {
@@ -166,9 +168,6 @@ namespace osu.Framework.Graphics.Containers
             return Task.Factory.StartNew(() => loadComponents(loadables, deps, true, linkedSource.Token), linkedSource.Token, TaskCreationOptions.HideScheduler, taskScheduler).ContinueWith(loaded =>
             {
                 var exception = loaded.Exception?.AsSingular();
-
-                if (!loaded.Result.Any())
-                    return;
 
                 if (linkedSource.Token.IsCancellationRequested)
                 {
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -211,7 +211,6 @@ namespace osu.Framework.Graphics.Containers
             if (IsDisposed)
                 throw new ObjectDisposedException(ToString());
 
-            // ReSharper disable once IteratorMethodResultIsIgnored (we don't care about the results here).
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Containers/CompositeDrawable.cs
```diff
@@ -153,19 +153,21 @@ protected internal Task LoadComponentAsync<TLoadable>([NotNull] TLoadable compon
...
+            var taskScheduler = loadables.Any(c => c.IsLongRunning) ? long_load_scheduler : threaded_scheduler;
 
-            return Task.Factory.StartNew(() => loadComponents(ref components, deps, true, linkedSource.Token), linkedSource.Token, TaskCreationOptions.HideScheduler, taskScheduler).ContinueWith(t =>
+            return Task.Factory.StartNew(() => loadComponents(loadables, deps, true, linkedSource.Token), linkedSource.Token, TaskCreationOptions.HideScheduler, taskScheduler).ContinueWith(loaded =>
             {
-                var exception = t.Exception?.AsSingular();
+                var exception = loaded.Exception?.AsSingular();
 
-                if (!components.Any())
+                if (!loaded.Result.Any())
```
smoogipoo, line 170: At least one of the issues is going to be caused by this - accessing `Result` will throw if the task has faulted.

This could be moved out to just after `components.ToArray()` above - I don't believe there was a reason for this being here other than to prevent multiple evaluation.

The other failures are DLUW - for some reason references aren't being lost there, but I'm not yet sure on the reason.

peppy:
Should be good to go now

bdach (approved):
As a change this seems correct from what I can tell. Shame about the few extra allocs but I think I prefer the usage safety. If it's deemed a problem then we can scope down the param type to an array if need be.

peppy:
Yeah, not sure performance/allocs is a huge deal for this flow as loading drawables should be a *relatively* rare operation. Can be fixed if/when it becomes an issue for a specific usage.

osu.Framework/Graphics/Containers/CompositeDrawable.cs
```diff
@@ -209,19 +211,23 @@ protected internal Task LoadComponentAsync<TLoadable>([NotNull] TLoadable compon
             if (IsDisposed)
                 throw new ObjectDisposedException(ToString());
 
-            loadComponents(ref components, Dependencies, false);
+            // ReSharper disable once IteratorMethodResultIsIgnored (we don't care about the results here).
```
smoogipoo, line 214: I don't get it, what iterator? Is this an out of date comment? I don't get a warning here.

smoogipoo (changes requested):
Rest seems fine

osu.Framework/Graphics/Containers/CompositeDrawable.cs
peppy: out of date comment, yes.

smoogipoo (approved):

PR #3934, opened 2020-10-14
Implement precise scrolling for SDL

swoolcock:
Adds support for "precise" scrolling on macOS for the SDL backend.

Due to the way SDL handles scroll wheel events on macOS, the precise deltas in the `NSEvent` are ignored.  Unfortunately, the originating `NSEvent` is not available to us anywhere in the SDL event loop itself.

To solve this, we swizzle out `[SDLView scrollWheel:(NSEvent *)]` to use our own custom function that handles precision deltas.  If precision deltas are unavailable (due to using a physical notched scroll wheel rather than a trackpad), it calls the original implementation of the method.

The logic for converting and scaling precise deltas is the same as that in osuTK's `CocoaNativeWindow`.

Instantiation of `IWindowBackend` and `IGraphicsBackend` have been moved to abstract methods rather than being passed down through constructors.


peppy (approved):
works as expected, nice

PR #3935, opened 2020-10-14
Allow DelayedLoadWrapper to late-construct its content

peppy:
Not only does this tidy up the `DelayedLoadUnloadWrapper` flow, but it allows us to (in both cases) avoid the overhead of construction of the content drawable until a load actually occurs.

Tested against my beatmap carousel changes, this reduces finalizer disposal to zero (and greatly reduces allocs in fast scrolling) as expected.

- Make sure you are okay with the Relative/AutoSize propgation. I haven't changed anything for the old-style ctor for compatibility reasons, but now the transfer is done at load time, overwriting anything set locally. Maybe this needs to be mentioned somewhere in the xmldoc?
- Should we obsolete the old ctor?

### DIFF
--- a/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
+++ b/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
@@ -16,6 +16,10 @@ namespace osu.Framework.Graphics.Containers
     /// Has the ability to delay the loading until it has been visible on-screen for a specified duration.
     /// In order to benefit from delayed load, we must be inside a <see cref="ScrollContainer{T}"/>.
     /// </summary>
+    /// <remarks>
+    /// <see cref="LifetimeStart"/> and <see cref="LifetimeEnd"/> are propagated from the wrapper to the content on content load.
+    /// After load, the content's lifetime is preferred, meaning any changes to content's lifetime post-load will be respected.
+    /// </remarks>
     public class DelayedLoadWrapper : CompositeDrawable
     {
         [Resolved]
@@ -56,16 +60,30 @@ namespace osu.Framework.Graphics.Containers
             AddLayout(isIntersectingCache);
         }
 
+        private double lifetimeStart = double.MinValue;
+
         public override double LifetimeStart
         {
-            get => Content?.LifetimeStart ?? base.LifetimeStart;
-            set => Content.LifetimeStart = value;
+            get => Content?.LifetimeStart ?? lifetimeStart;
+            set
+            {
+                if (Content != null)
+                    Content.LifetimeStart = value;
+                lifetimeStart = value;
+            }
         }
+
+        private double lifetimeEnd = double.MaxValue;
 
         public override double LifetimeEnd
         {
-            get => Content?.LifetimeEnd ?? base.LifetimeEnd;
-            set => Content.LifetimeEnd = value;
+            get => Content?.LifetimeEnd ?? lifetimeEnd;
+            set
+            {
+                if (Content != null)
+                    Content.LifetimeEnd = value;
+                lifetimeEnd = value;
+            }
         }
 
         private Drawable content;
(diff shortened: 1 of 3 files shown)


bdach:
The change itself seems fine to me.

As for the obsoletion of the old ctor - we actually have game-side usages that would arguably become uglier if we forced late construction. See [`LeaderboardScore` for one](https://github.com/ppy/osu/blob/74a0886cb8f41018a95296f24e1b113af83307df/osu.Game/Online/Leaderboards/LeaderboardScore.cs#L117-L133) - the constructed-but-not-loaded drawable is then [used to register an `OnLoadComplete`](https://github.com/ppy/osu/blob/74a0886cb8f41018a95296f24e1b113af83307df/osu.Game/Online/Leaderboards/LeaderboardScore.cs#L230) callback; forcing late construction would probably make this way worse to read (would have to subscribe to `DelayedLoadComplete` and then nested-subscribe to `OnLoadComplete`).

Also, just so there's no blockage - this technically conflicts with #3936 but the resolution is relatively easy. Here's the combined diff:

<details>
<summary>combined diff</summary>

```diff
diff --cc osu.Framework/Graphics/Containers/DelayedLoadUnloadWrapper.cs
index 49d012b81a,f7c72c2e94..d383017ee8
--- a/osu.Framework/Graphics/Containers/DelayedLoadUnloadWrapper.cs
+++ b/osu.Framework/Graphics/Containers/DelayedLoadUnloadWrapper.cs
@@@ -56,8 -58,10 +56,8 @@@ namespace osu.Framework.Graphics.Contai
              }
          }
  
-         private bool contentLoaded;
 -        public override Drawable Content => base.Content ?? (Content = createContentFunction());
 -
          private ScheduledDelegate scheduledLifetimeUpdate;
+         private ScheduledDelegate scheduledUnloadCheckRegistration;
  
          protected override void EndDelayedLoad(Drawable content)
          {
diff --cc osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
index 7895f5e13c,904f4ec90c..d3922c3e20
--- a/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
+++ b/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
@@@ -121,14 -83,16 +121,18 @@@ namespace osu.Framework.Graphics.Contai
  
          protected void BeginDelayedLoad()
          {
-             if (loadTask != null) throw new InvalidOperationException("Load is already started!");
+             if (DelayedLoadTriggered || DelayedLoadCompleted)
+                 throw new InvalidOperationException("Load has already started!");
  
 +            Content ??= createFunc();
 +
+             DelayedLoadTriggered = true;
              DelayedLoadStarted?.Invoke(Content);
  
+             cancellationTokenSource = new CancellationTokenSource();
+ 
              // The callback is run on the game's scheduler since DLUW needs to unload when no updates are being received.
-             loadTask = LoadComponentAsync(Content, EndDelayedLoad, scheduler: Game.Scheduler);
+             LoadComponentAsync(Content, EndDelayedLoad, scheduler: Game.Scheduler, cancellation: cancellationTokenSource.Token);
          }
  
          protected virtual void EndDelayedLoad(Drawable content)
```

</details>

(in short, take both additions in DLW and both removals in DLUW).

bdach (approved):

osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
```diff
@@ -21,37 +21,75 @@ public class DelayedLoadWrapper : CompositeDrawable
...
+            this.timeBeforeLoad = timeBeforeLoad;
 
             AddLayout(optimisingContainerCache);
             AddLayout(isIntersectingCache);
         }
 
         public override double LifetimeStart
         {
-            get => Content.LifetimeStart;
+            get => Content?.LifetimeStart ?? base.LifetimeStart;
```
bdach, line 61: I sort of glossed over this before, but this is a bit of an API inconsistency in that setting a DLW's lifetime behaves differently to setting DLUW's lifetime before the content is constructed.

* To set a DLW's content's lifetime before it's constructed you have to make sure to specify the lifetime in the content-constructing delegate, as setting lifetime on DLW before construction will throw on a nullref.
* To set a DLUW's content's lifetime before it's constructed you can just set `Lifetime{Start,End}` on the DLUW itself and the logic therein will ensure it's transferred upon the content being constructed.

Maybe we want to move the lifetime logic, along with the lifetime-setting schedule, to DLW entirely?

peppy: Yeah, probably. The whole lifetime propagation is a bit haphazard in general though.

smoogipoo (approved):
lgtm

PR #3941, opened 2020-10-16
Fix DLUW sometimes crashing on disposal

smoogipoo:
Alternative to & closes #3936
Fixes the CI issues in https://github.com/ppy/osu/pull/10493 (at least, I can't make `TestScenePlaySongSelect.TestGroupedDifficultyIconSelecting` fail).

Reverts https://github.com/ppy/osu-framework/pull/3926, and fixes the memory leak - by locking and checking the disposal state prior to scheduling the unload delegate.

It wasn't always scheduled twice on the game like this - the disposal check already existed previously in `checkForUnload`, but it wasn't hoisted out of there when the game started being scheduled to a second time in https://github.com/ppy/osu-framework/pull/3825.

Credit to @bdach for the test case (taken from #3936).

### DIFF
--- a/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
+++ b/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
@@ -104,9 +104,9 @@ namespace osu.Framework.Graphics.Containers
             });
         }
 
-        internal override void UnbindAllBindables()
+        protected override void Dispose(bool isDisposing)
         {
-            base.UnbindAllBindables();
+            base.Dispose(isDisposing);
             CancelTasks();
         }
 
--- a/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
+++ b/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
@@ -104,6 +104,12 @@ namespace osu.Framework.Graphics.Containers
             });
         }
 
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
```diff
@@ -104,9 +104,9 @@ protected virtual void EndDelayedLoad(Drawable content)
             });
         }
 
-        internal override void UnbindAllBindables()
```
peppy, line 107: sure we don't want to leave this just to clean things up ASAP?

smoogipoo: Yeah sure, I think there should be no issue with that.

smoogipoo: Wait, there is an issue with that... Somehow causes the assertions in the callback to fail...

bdach: Removing this actually neutered my test case as it used `UnbindAllBindables` to simulate load task cancellation. Putting it back triggers the double load issue I pointed out in my original PR.

smoogipoo:
Looking into merging this into @bdach's PR...

PR #3945, opened 2020-10-19
Allow specifying a custom fixed width source character

peppy:


osu.Framework/Graphics/Sprites/SpriteText.cs
```diff
@@ -416,7 +416,11 @@ public MarginPadding Padding
         #region Characters
 
         private readonly LayoutValue charactersCache = new LayoutValue(Invalidation.DrawSize | Invalidation.Presence, InvalidationSource.Parent);
-        private readonly List<TextBuilderGlyph> charactersBacking = new List<TextBuilderGlyph>();
+
+        /// <summary>
+        /// Glyph list to be passed to <see cref="TextBuilder"/>.
+        /// </summary>
+        protected readonly List<TextBuilderGlyph> CharactersBacking = new List<TextBuilderGlyph>();
```
smoogipoo, line 423: Protected? What for?

osu.Framework/Text/TextBuilder.cs
```diff
@@ -50,8 +51,9 @@ public class TextBuilder
         /// <param name="characterList">That list to contain all resulting <see cref="TextBuilderGlyph"/>s.</param>
         /// <param name="neverFixedWidthCharacters">The characters for which fixed width should never be applied.</param>
         /// <param name="fallbackCharacter">The character to use if a glyph lookup fails.</param>
+        /// <param name="fixedWidthCalculationCharacter">The character to use to calculate the fixed width width. Defaults to 'm'.</param>
         public TextBuilder(ITexturedGlyphLookupStore store, FontUsage font, float maxWidth = float.MaxValue, bool useFontSizeAsHeight = true, Vector2 startOffset = default, Vector2 spacing = default,
-                           List<TextBuilderGlyph> characterList = null, char[] neverFixedWidthCharacters = null, char fallbackCharacter = '?')
+                           List<TextBuilderGlyph> characterList = null, char[] neverFixedWidthCharacters = null, char fallbackCharacter = '?', char fixedWidthCalculationCharacter = 'm')
```
smoogipoo, line 56: Did you intend to add a virtual property to `SpriteText` for this?

smoogipoo (changes requested):

osu.Framework/Graphics/Sprites/SpriteText.cs
peppy: See https://github.com/ppy/osu/pull/10556

osu.Framework/Text/TextBuilder.cs
peppy: I guess we can do it that way? I'm not sure I understand the purpose of `CreateTextBuilder` in that case.. should ew make this class `internal` if you don't expect it to be constructed externally?

osu.Framework/Graphics/Sprites/SpriteText.cs
smoogipoo: I think `SpriteText` should do `Characters = textBuilder.Characters`, which would be a no-op for the basic `SpriteText`.

But I think exposing the char as a virtual property is better.

osu.Framework/Text/TextBuilder.cs
smoogipoo: Not sure on `internal`. Aside from the issue that `SpriteText` should be taking on the TB's character list.

smoogipoo (approved):

PR #3947, opened 2020-10-23
Add DrawablePool statistics to GlobalStatistics

smoogipoo:
This can be/has been useful in visualising pool usage.

Preview: https://drive.google.com/file/d/16bMpWYfDMOecs67TBJMY2M7-7bEs7eEG/view?usp=sharing

The current format is `count_available / count_newed (count_in_use)`, but please provide an alternative if this doesn't sit right with you.

### DIFF
--- a/osu.Framework/Graphics/Pooling/DrawablePool.cs
+++ b/osu.Framework/Graphics/Pooling/DrawablePool.cs
@@ -160,8 +160,10 @@ namespace osu.Framework.Graphics.Pooling
             get => countInUse;
             private set
             {
-                if (statistic != null)
-                    statistic.Value.CountInUse += value - countInUse;
+                if (statistic == null)
+                    return;
+
+                statistic.Value.CountInUse += value - countInUse;
                 countInUse = value;
             }
         }
@@ -176,8 +178,10 @@ namespace osu.Framework.Graphics.Pooling
             get => countConstructed;
             private set
             {
-                if (statistic != null)
-                    statistic.Value.CountConstructed += value - countConstructed;
+                if (statistic == null)
+                    return;
+
+                statistic.Value.CountConstructed += value - countConstructed;
                 countConstructed = value;
             }
         }
@@ -192,8 +196,10 @@ namespace osu.Framework.Graphics.Pooling
             get => countAvailable;
             private set
             {
-                if (statistic != null)
-                    statistic.Value.CountAvailable += value - countAvailable;
+                if (statistic == null)
+                    return;
+
+                statistic.Value.CountAvailable += value - countAvailable;
                 countAvailable = value;
             }
         }
--- a/osu.Framework/Graphics/Pooling/DrawablePool.cs
+++ b/osu.Framework/Graphics/Pooling/DrawablePool.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using osu.Framework.Allocation;
 using osu.Framework.Extensions.TypeExtensions;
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Pooling/DrawablePool.cs
```diff
@@ -121,7 +138,80 @@ private bool push(T poolableDrawable)
...
+            get => localCountInUse;
+            set
+            {
+                if (statistic != null)
+                    statistic.Value.CountInUse += value - localCountInUse;
+                localCountInUse = value;
+            }
+        }
+
+        private int localCountNewed;
```
peppy, line 171: to avoid the "local" prefixes, how about exposing these as public get? i think it wouldn't do any harm (and having xmldoc would be nice)

osu.Framework/Graphics/Pooling/DrawablePool.cs
```diff
@@ -121,7 +138,80 @@ private bool push(T poolableDrawable)
...
+
+        private int countNewed
```
peppy, line 173: `CountConstructed` maybe?

osu.Framework/Graphics/Pooling/DrawablePool.cs
```diff
@@ -121,7 +138,80 @@ private bool push(T poolableDrawable)
...
+        }
+
+        private int localCountInUse;
+
+        private int countInUse
+        {
+            get => localCountInUse;
+            set
+            {
+                if (statistic != null)
```
peppy, line 165: when is this null? seems to be only after disposal?

peppy (changes requested):

peppy:
Display format seems fine.

osu.Framework/Graphics/Pooling/DrawablePool.cs
smoogipoo: Yeah, dispose is the reason. I'm not sure if it's being too cautious, but I was imagining a situation where e.g. the pool is disposed early and then the drawable somehow expires and wants to add itself back to the pool.

I wouldn't be able to say when that would occur though - could remove it and see if any exception ever pops up?

peppy: the null check is fine to remain but if it's only on disposal, i think it should just return if null and remove the weird delayed addition logic?

smoogipoo: What "weird delayed addition logic" are you referring to. Do you want to just invert this like:
```
if (null)
    return;
statistic++;
count++;
```

osu.Framework/Graphics/Pooling/DrawablePool.cs
smoogipoo: In the process of doing so, I've replaced the existing `CountAvailable` property (which used to forward directly to `pool.Count`).

osu.Framework/Graphics/Pooling/DrawablePool.cs
peppy: yeah that's fine. the previous logic read weirdly.

smoogipoo: One thing to note is that disposal isn't thread safe. I could move it into UnbindAllBindables, any opinions? Kinda makes these null-checks redundant unless the addition happens _after_ disposal completes...

peppy: I've just make them asserts and call it a day.

peppy (approved):

PR #3948, opened 2020-10-23
Add a way to retrieve unknown types from DrawablePool

smoogipoo:
Hitobject pooling has brought up a case where it's necessary to retrieve drawables from a set of untyped pools where constraints on the types retrieved are locally guaranteed.

This makes the following (example) scenario possible:

```csharp
private List<IDrawablePool> pools = new List<IDrawablePool>();

void Register<T>(DrawablePool<T> pool)
    where T : DrawableHitObject, new()
{
    pools.Add(pool);
}

DrawableHitObject Retrieve()
{
    // DrawableHitObject type guaranteed by the constraint on Register().
    return (DrawableHitObject)pools[2].Get(d => ((DrawableHitObject)d).SomeMethod());
}
```

### DIFF
--- a/osu.Framework/Graphics/Pooling/DrawablePool.cs
+++ b/osu.Framework/Graphics/Pooling/DrawablePool.cs
@@ -72,6 +72,8 @@ namespace osu.Framework.Graphics.Pooling
             push((T)pooledDrawable);
         }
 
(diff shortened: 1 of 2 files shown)


peppy (approved):
guess this is fine

PR #3949, opened 2020-10-23
Add a way to re-sort CompositeDrawable children

smoogipoo:
Re-sorting children on comparer change is really ugly and error prone. It's done in at least `HitObjectContainer` osu!-side, where the comparer is sorted by `HitObject.StartTime` (a mutable variable), and the solution is to remove-then-readd:
https://github.com/ppy/osu/blob/89797d7a57e6c86bb9c15c03fd67e4b60cedbfaa/osu.Game/Rulesets/UI/HitObjectContainer.cs#L67-L75

The upcoming hitobject pooling changes demonstrate the error-proneness of this - remove-then-add is not possible with `PoolableDrawable` because they're returned to the `DrawablePool` when the parent is changed:
https://github.com/ppy/osu-framework/blob/d42278299b1b3ce1b06781563f063b71f89238e3/osu.Framework/Graphics/Pooling/PoolableDrawable.cs#L115-L124

Based on the several workarounds I've tried to get around this, I've come to the conclusion that the only reasonable solution is to expose a method that can re-sort children in-place based on the `CompositeDrawable`'s existing comparer.

I've only exposed this as a protected method from `CompositeDrawable`, because CompositeDrawable defines the comparer itself so only it can know whether this method should be used.

### DIFF
--- a/osu.Framework.Tests/Lists/TestSortedList.cs
+++ b/osu.Framework.Tests/Lists/TestSortedList.cs
@@ -101,5 +101,36 @@ namespace osu.Framework.Tests.Lists
                     list.RemoveAt(list.Count - 1);
             });
         }
+
+        [Test]
+        public void TestSortAfterChangedComparer()
+        {
+            var list = new SortedList<TestObjectWithAdjustableComparer>
+            {
+                new TestObjectWithAdjustableComparer { Id = 0 },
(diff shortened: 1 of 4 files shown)


peppy (approved):
seems sane

PR #3950, opened 2020-10-24
Use scan codes instead of key codes in SDL mapping

bdach:
Aims to address [some](https://github.com/ppy/osuTK/pull/67#issuecomment-715893612) [complaints](https://github.com/ppy/osuTK/pull/67#issuecomment-715893732) with SDL keyboard handling.

# Summary

Users have reported that when using SDL on some regional keyboard layouts (German QWERTZ/French AZERTY, for one), some keys which are used to insert characters like ü, ä, ß, do not seem to trigger text box input.

As it turns out, not all possible values of the `SDL_Keycode` enumeration (https://wiki.libsdl.org/SDL_Keycode) have their symbolic names. In the case of those special keyboard layouts, pressing the special diacritics keys would emit a `Keycode` with a value correspondent to the Unicode code point of the character pressed by the user, which has no osuTK correspondent, therefore getting converted to `Key.Unknown` and as such discarded.

Contrarily to that, `SDL_Scancode` (https://wiki.libsdl.org/SDL_Scancode) values correspond to actual physical keys present on the keyboard, and are lifted from the USB spec. Therefore they are probably a better fit to produce key events than keycode was.

Note that this does mean however that the `Key` enumeration, currently residing in osuTK, does not (and never has, really) correspond to the actual text output pressing said key will produce. In more practical terms, this means that while this fix will make text box input work again, other places will likely still look wrong (see keybinding panels, for instance, that use the key type from the keypress event for display purposes). This should probably be fixed at a later date by providing a
`Key` -> `char` (or similar) mapping function in the framework, valid for the current keyboard layout - if possible, that is.

This does also not fix input using the <kbd>AltGr</kbd> key not committing to textboxes properly. That won't be fixed as long as this lives:

https://github.com/ppy/osu-framework/blob/82f49bb7c664bb982febb010356c6aabd7ff0416/osu.Framework/Graphics/UserInterface/TextBox.cs#L706-L707

# Remarks

I considered two alternative resolutions, both incorrect:

* removing [this guard for `Unknown`](https://github.com/ppy/osu-framework/blob/82f49bb7c664bb982febb010356c6aabd7ff0416/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs#L771-L772) - would clump up all the special keys into one, potentially firing wrong event handlers if multiple special keys were pressed,
* casting the `keysym` value directly to `Key` - could conflict with some media keys.

Conversion can be cross-checked [here](https://wiki.libsdl.org/SDL_Scancode). It's mostly a

```
%s/SDL_Keycode\.SDLK/SDL_Scancode\.SDL_SCANCODE/g
```

replace, except for:

* `QUOTE` -> `APOSTROPHE`
* `BACKQUOTE` -> `GRAVE`
* `[a-z]` changed to capitals `[A-Z]`
* `NONUSBACKSLASH` added (red key marked on screenshots in discussion linked above).

I tested on German QWERTZ and French AZERTY layouts on my ubuntu box, but cc'ing @Morilli @FamousPig @Corne2Plum3 - cross-checking would be appreciated if possible, because I do not physically have that red extra ISO layout key. To check you only have to run the visual tests in the framework and try input in `TestSceneTextBox` for example.

### DIFF
--- a/osu.Framework/Platform/Sdl/Sdl2Extensions.cs
+++ b/osu.Framework/Platform/Sdl/Sdl2Extensions.cs
@@ -11,427 +11,427 @@ namespace osu.Framework.Platform.Sdl
     {
         public static Key ToKey(this SDL.SDL_Keysym sdlKeysym)
         {
-            switch (sdlKeysym.sym)
+            switch (sdlKeysym.scancode)
             {
                 default:
-                case SDL.SDL_Keycode.SDLK_UNKNOWN:
+                case SDL.SDL_Scancode.SDL_SCANCODE_UNKNOWN:
                     return Key.Unknown;
 
-                case SDL.SDL_Keycode.SDLK_KP_COMMA:
+                case SDL.SDL_Scancode.SDL_SCANCODE_KP_COMMA:
                     return Key.Comma;
 
-                case SDL.SDL_Keycode.SDLK_KP_TAB:
+                case SDL.SDL_Scancode.SDL_SCANCODE_KP_TAB:
                     return Key.Tab;
 
-                case SDL.SDL_Keycode.SDLK_KP_BACKSPACE:
+                case SDL.SDL_Scancode.SDL_SCANCODE_KP_BACKSPACE:
                     return Key.BackSpace;
 
-                case SDL.SDL_Keycode.SDLK_KP_A:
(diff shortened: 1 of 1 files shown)


Morilli:
Work perfectly fine on my end from what I can see 👍

peppy (approved):

PR #3951, opened 2020-10-24
Fix unmanaged string being incorrectly parsed as ANSI in SDL

FamousPig:
[SDL_TEXTINPUTEVENT.text](https://wiki.libsdl.org/SDL_TextInputEvent) is incorrectly parsed as ANSI instead of UTF-8, leading to incorrect non-ascii text inputs.

Before:
before

After:
after

### DIFF
--- a/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
+++ b/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
@@ -754,7 +754,7 @@ namespace osu.Framework.Platform.Sdl
             if (ptr == IntPtr.Zero)
                 return;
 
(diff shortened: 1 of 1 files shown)


bdach:
"Funny" thing is, this will not (and seemingly did not, unless I had several strokes) manifest on unix as `Marshal.PtrToStringAnsi` [is really `PtrToStringInWhateverCodepageYouHappenToBeOn`, and on unix it's usually utf-8.](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.charset?view=netcore-3.1)

Anyway, seems correct as far as I can tell.

bdach (approved):

PR #3952, opened 2020-10-25
Fix waveform graph drifting on zoom changes

bdach:
Resolves https://github.com/ppy/osu/issues/9482.

# Summary

The waveform graph had a tendency of seemingly randomly drifting back and forth on zoom changes. This was due to an overlooked shortcoming in the point generation algorithm, which resulted in floating-point errors.

The root cause of that was a for-loop in the point-generating routine. The fractional index `i` was kept to determine the neighbourhood from which to sample the next generated point. However, it was being incremented in each iteration of said for-loop by `pointsPerGeneratedPoint`. At larger zoom levels this could result in accumulating floating-point precision errors, causing the visible jitters back and forth.

Resolve by keeping two indices; one fractional for the sampling neighbourhood, and one integral for the point which is being generated. Instead of incrementing by `pointsPerGeneratedPoint`, multiplication is used to avoid accumulating errors.

# Remarks

[Video demonstrating effect of change.](https://drive.google.com/file/d/1XZOwFeeyiO5GwAAZlXH95EpXw3YBOTqD/view?usp=sharing)

### DIFF
--- a/osu.Framework/Audio/Track/Waveform.cs
+++ b/osu.Framework/Audio/Track/Waveform.cs
@@ -207,12 +207,19 @@ namespace osu.Framework.Audio.Track
                     filter[i] = (float)Blur.EvalGaussian(i, pointsPerGeneratedPoint);
                 }
 
-                for (float i = 0; i < points.Count; i += pointsPerGeneratedPoint)
+                // we're keeping two indices: one for the original (fractional!) point we're generating based on,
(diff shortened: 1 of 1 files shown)


peppy (approved):
Looks to fix the issue.

peppy:
@smoogipoo can you make sure you're okay with the logic/comment changes? you're more familiar with this code than i am.

smoogipoo (approved):

PR #3953, opened 2020-10-26
Extract lifetime managing parts of LifetimeManagementContainer

smoogipoo:
This is best tested osu!-side using https://github.com/smoogipoo/osu/tree/hitobject-pooling (branch isn't ready for PR yet though)

## Summary

Hitobject pooling requires very special handling of drawables - namely - that there are few drawables and potentially thousands of models that can be attached to them. The current LTMC implementation isn't flexible enough to allow this.

The main purpose of this PR is to extract the lifetime managing parts of LTMC into a more basic component, `LifetimeEntryManager`, that operates on model data. LTMC and `HitObjectContainer` both use this in their own ways for their own purposes.

## Important changes

Much of the logic is shared between the previous LTMC implementation and the new LTEM implementation, save for some refactoring, but there are some important changes:

### LTMC no longer has a load state check in `CheckChildrenLife()`

https://github.com/ppy/osu-framework/blob/4cb1bdf694109a045cd78ad35067125f138ac7e1/osu.Framework/Graphics/Containers/LifetimeManagementContainer.cs#L121-L122

I don't believe this was correct, because `Drawable` is guaranteed to have a clock when it becomes `LoadState.Ready`:

https://github.com/ppy/osu-framework/blob/4cb1bdf694109a045cd78ad35067125f138ac7e1/osu.Framework/Graphics/Drawable.cs#L265-L267

The only case to watch out for is `CompositeDrawable.LoadAsyncComplete()` which calls `CheckChildrenLife()` _before_ it gets a ready state, but the call to `LoadAsyncComplete()` always happens after the clock is updated:
https://github.com/ppy/osu-framework/blob/4cb1bdf694109a045cd78ad35067125f138ac7e1/osu.Framework/Graphics/Drawable.cs#L272-L290

### LTMC no longer checks for `child.LoadState >= Ready` before making alive

In a similar vein to the above, LTMC had the following code before moving children from effectively the `newEntries` list:

https://github.com/ppy/osu-framework/blob/4cb1bdf694109a045cd78ad35067125f138ac7e1/osu.Framework/Graphics/Containers/LifetimeManagementContainer.cs#L126-L143

I believe the LoadState check here is also unnecessary, as for this code to run two things are required:
1. The CompositeDrawable is at least in `LoadState.Ready`.
2. The drawables are part of the composite's hierarchy.

Suppose that a very timely addition happened to the composite, juuuuuuuuuuust before `CheckChildrenLife()` is run. Such a case would cause the drawable to immediately load, as per the following block in `AddInternal()`:

https://github.com/ppy/osu-framework/blob/4cb1bdf694109a045cd78ad35067125f138ac7e1/osu.Framework/Graphics/Containers/CompositeDrawable.cs#L544-L552

So I think LTMC can guarantee by this point that the child is `>= LoadState.Ready` by the time this code is run.

### `AddInternal()` has an overload that turns off the lifetime management

For use in `HitObjectContainer`, I've added an overload to `AddInternal()` that turns off the lifetime management. In the case of unmanaged drawables, their aliveness update _may_ be delegated until the first `CheckChildrenLife()` (as per the `MakeChildAlive()` spec).

I'm open to opinions/alternatives here, but it's unfortunate that LTMC overrides `AddInternal()` to do its thing :(

### Lifetime changes are now always updated in LTMC's next frame

Previously, LTMC used to schedule drawable lifetime updates to its `UpdateAfterChildren()`:

https://github.com/ppy/osu-framework/blob/4cb1bdf694109a045cd78ad35067125f138ac7e1/osu.Framework/Graphics/Containers/LifetimeManagementContainer.cs#L213-L225

This differs from a normal `CompositeDrawable`, which always updates children lifetimes in their next update frame and no such schedule exists. The way this works in LTEM is the drawables get pushed back tonto the `newEntries` list to be re-processed in the next frame:

https://github.com/smoogipoo/osu-framework/blob/a190f6af696627467cdc3a230f975a9487bbdea2/osu.Framework/Graphics/Performance/LifetimeEntryManager.cs#L161-L172

I'm mentioning this one because it's a slight change that specifically breaks `TestSceneLifetimeManagementContainer.TestFuzz`, and is why that test has had a line added to it.

### All tests have been duplicated into `LifetimeEntryManagerTest`, and `LifetimeEntryManagerTest.TestFuzz` is always enabled

It doesn't say much about the raw performance of the component, but not being part of the draw hierarchy improves the test speed significantly, to the point where the test can now remain always-enabled. It's still ignored in `TestSceneLifetimeManagementContainer` though.

### Added a method to update based on a time range

A small addition that helps the osu!-side implementation. HOC needs to have a way to specify objects -2000ms - +2000ms from the current time should be alive for editor blueprints.

I can extract this into a separate PR if you'd prefer, though I think it's fairly straight forward.

### Re-namespaced `LifetimeBoundaryCrossingDirection` and `LifetimeBoundaryKind`

It's a slight breaking change, let me know if I should move them back (used in two places osu!-side). I'm generally not 100% on the namespacing on anything here, so I'm open to discussion.


osu.Framework/Graphics/Containers/LifetimeManagementContainer.cs
```diff
@@ -22,247 +20,91 @@ namespace osu.Framework.Graphics.Containers
...
+            unmanagedDrawablesToProcess.Clear();
 
-        protected internal override void AddInternal(Drawable drawable)
-        {
-            Trace.Assert(!drawable.RemoveWhenNotAlive, $"{nameof(RemoveWhenNotAlive)} is not supported for {nameof(LifetimeManagementContainer)}");
-            drawable.LifetimeChanged += childLifetimeChanged;
-
-            newChildren.Add(drawable);
-            base.AddInternal(drawable);
+            return aliveChanged | manager.Update(Time.Current);
```
bdach, line 94: not sure how to feel about this... it's clever, but i also sort of dislike it for that reason

osu.Framework.Tests/Graphics/LifetimeEntryManagerTest.cs
```diff
@@ -0,0 +1,342 @@
...
+            checkCrossing(manager.Entries[2], 1, LifetimeBoundaryKind.End, LifetimeBoundaryCrossingDirection.Forward);
+
+            manager.Update(1);
+            checkNoCrossing(manager.Entries[0]);
+            checkNoCrossing(manager.Entries[1]);
+            checkCrossing(manager.Entries[2], 0, LifetimeBoundaryKind.End, LifetimeBoundaryCrossingDirection.Backward);
+
+            manager.Update(-1);
+            checkCrossing(manager.Entries[0], 0, LifetimeBoundaryKind.End, LifetimeBoundaryCrossingDirection.Backward);
+            checkCrossing(manager.Entries[1], 0, LifetimeBoundaryKind.End, LifetimeBoundaryCrossingDirection.Backward);
```
bdach, line 115: For completeness, should also assert
```suggestion
            checkCrossing(manager.Entries[1], 0, LifetimeBoundaryKind.End, LifetimeBoundaryCrossingDirection.Backward);
            checkCrossing(manager.Entries[1], 1, LifetimeBoundaryKind.Start, LifetimeBoundaryCrossingDirection.Backward);
```

osu.Framework.Tests/Graphics/LifetimeEntryManagerTest.cs
```diff
@@ -0,0 +1,342 @@
...
+        public void TestLifetimeChangeOnCallback()
+        {
+            int updateTime = 0;
+
+            manager.AddEntry(new LifetimeEntry { LifetimeStart = 0, LifetimeEnd = 1 });
+            manager.EntryCrossedBoundary += (entry, kind, direction) =>
+            {
+                switch (kind)
+                {
+                    case LifetimeBoundaryKind.End when direction == LifetimeBoundaryCrossingDirection.Forward:
```
bdach, line 129: I think we generally tend to avoid `case..when` but I'm fine with it myself in this case.

osu.Framework.Tests/Graphics/LifetimeEntryManagerTest.cs
```diff
@@ -0,0 +1,342 @@
...
+            void randomLifetime(out double l, out double r)
+            {
+                l = rng.Next(5);
+                r = rng.Next(5);
+
+                if (l > r)
+                {
+                    var l1 = l;
+                    l = r;
+                    r = l1;
```
bdach, line 227: neat trick for this
```suggestion
                    (l, r) = (r, l);
```

bdach (commented):
A few notes from a read-through. The raw numbers looked scarier than the actual diff - it did take me an hour or two to get through still... Looks mostly sane as far as I can tell.

Haven't ran with the linked branch yet.

osu.Framework.Tests/Graphics/LifetimeEntryManagerTest.cs
smoogipoo: I changed it because I thought it looked cleaner than the previous:

https://github.com/ppy/osu-framework/blob/4cb1bdf694109a045cd78ad35067125f138ac7e1/osu.Framework.Tests/Visual/Containers/TestSceneLifetimeManagementContainer.cs#L176-L183

osu.Framework/Graphics/Performance/LifetimeEntryManager.cs
```diff
@@ -0,0 +1,418 @@
...
+            futureEntries.Clear();
+        }
+
+        /// <summary>
+        /// Invoked when the lifetime of an entry has changed, so that an appropriate sorting order is maintained.
+        /// </summary>
+        /// <param name="entry">The <see cref="LifetimeEntry"/> that changed.</param>
+        /// <param name="lifetimeStart">The new start time.</param>
+        /// <param name="lifetimeEnd">The new end time.</param>
+        private void requestLifetimeUpdate(LifetimeEntry entry, double lifetimeStart, double lifetimeEnd)
```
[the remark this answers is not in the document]
smoogipoo: I've added some comments and changed the name of the method. Hope it's more explanatory now?

osu.Framework/Graphics/Performance/LifetimeEntry.cs
```diff
@@ -0,0 +1,85 @@
...
+            get => lifetimeStart;
+            set
+            {
+                if (lifetimeStart == value)
+                    return;
+
+                if (RequestLifetimeUpdate != null)
+                    RequestLifetimeUpdate.Invoke(this, value, lifetimeEnd);
+                else
+                    UpdateLifetime(value, lifetimeEnd);
```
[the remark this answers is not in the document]
smoogipoo: Added remark to xmldoc on `RequestLifetimeUpdate`. Keep in mind this is internal stuff - I don't expect anything else other than LTEM to bind to it.

osu.Framework/Graphics/Containers/LifetimeManagementContainer.cs
smoogipoo: I've de-inlined it. Please request an alternative if that's not the sort of change you foresaw happening.

bdach:
All of the recent changes look good to me :+1:

osu.Framework.Tests/Graphics/LifetimeEntryManagerTest.cs
bdach: Yeah I think it's slightly cleaner too. Just pointed it out as I saw it.

peppy (approved):

PR #3955, opened 2020-10-27
Fix ChildrenOfType being thread-unsafe

peppy:
Could potentially fail when components are being asynchronously loaded.

```
osu.Game.Tests.Visual.Editing.TestSceneEditorClipboard.TestCutPaste(2000)

TearDown : System.InvalidOperationException : Collection was modified; enumeration operation may not execute.
--TearDown
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at osu.Framework.Testing.TestingExtensions.ChildrenOfType[T](Drawable drawable)+MoveNext()
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Boolean& found)
   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source)
   at osu.Game.Tests.Visual.EditorTestScene.get_EditorComponentsReady()
```

Not that this really solves the issue but it should be fine for this test-only method.

### DIFF
--- a/osu.Framework/Testing/TestingExtensions.cs
+++ b/osu.Framework/Testing/TestingExtensions.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Collections.Generic;
+using System.Linq;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 
@@ -19,7 +20,7 @@ namespace osu.Framework.Testing
 
             if (drawable is CompositeDrawable composite)
             {
-                foreach (var child in composite.InternalChildren)
+                foreach (var child in composite.InternalChildren.ToArray())
                 {
                     foreach (var found in child.ChildrenOfType<T>())
                         yield return found;


smoogipoo:
I'm not quite sure this changes much, ToArray() doesn't inherently make this code any safer. But I see no harm.

smoogipoo (approved):

smoogipoo:
Just wondering, is there any reason for the test not checking for the load state first? Something like:

```
AddStep("load editor", () => Child = editor = new Editor());
AddUntilStep("wait for load", () => editor.IsLoaded);
AddAssert("assert child", () => editor.ChildrenOfType<SelectionBlueprint>().Any()).
```
We do this elsewhere and have avoided this problem until now.

peppy:
Sure, it can. But the sub components are also async loaded. It would require more than just that, I think.

And yeah I get it's no more safe, but it will avoid the exception triggering.

smoogipoo:
It's a pain point for sure...

image

bdach:
As mentioned above, this can still fail if the mutation happens while `ToArray()` is performing the underlying array copy (see https://ci.appveyor.com/project/peppy/osu/builds/36065912/tests for one example of this). Not sure whether there's a better solution for that other than running the `ToArray()` in a loop with swallowing exceptions until it succeeds...

PR #3956, opened 2020-10-28
Fix OnResume being called twice when calling Exit from a blocked OnExiting implementation

peppy:
An issue that arose while looking into https://github.com/ppy/osu/issues/10418. Not required to fix that issue, but probably good to fix this while we're here.

### DIFF
--- a/osu.Framework/Screens/ScreenStack.cs
+++ b/osu.Framework/Screens/ScreenStack.cs
@@ -214,7 +214,6 @@ namespace osu.Framework.Screens
                 throw new ScreenNotInStackException(nameof(MakeCurrent));
 
             // while a parent still exists and exiting is not blocked, continue to iterate upwards.
-            IScreen firstScreen = CurrentScreen;
             IScreen exitSource = null;
 
             while (CurrentScreen != null)
--- a/osu.Framework/Screens/ScreenStack.cs
+++ b/osu.Framework/Screens/ScreenStack.cs
@@ -219,11 +219,20 @@ namespace osu.Framework.Screens
 
             while (CurrentScreen != null)
             {
-                if (exitFrom(exitSource, shouldFireResumeEvent: false) || CurrentScreen == target)
+                if (exitFrom(exitSource, shouldFireResumeEvent: false))
                 {
-                    // don't fire the resume event if the first screen blocked the exit.
-                    if (CurrentScreen != firstScreen)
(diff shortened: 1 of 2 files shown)


bdach (approved):
I *think* this is correct, but I'm somewhat trusting the test coverage, since I'm not too familiar with screen stack logic yet; might be wise to get a +1 from @smoogipoo as well.

osu.Framework/Screens/ScreenStack.cs
```diff
@@ -214,16 +214,24 @@ internal void MakeCurrent(IScreen target)
...
-                if (exitFrom(exitSource, shouldFireResumeEvent: false) || CurrentScreen == target)
+                if (exitFrom(exitSource, shouldFireResumeEvent: false))
                 {
-                    // don't fire the resume event if the first screen blocked the exit.
-                    if (CurrentScreen != firstScreen)
+                    // exit was blocked, but a nested exit operation may have succeeded (ie. a screen calling this.Exit() after blocking).
+                    // in such a case, the MakeCurrent / resumeFrom flow would have already been performed.
+                    // to avoid a duplicate resumeFrom event, only fire from here if it can be assured that the current screen is still the one which blocked the exit above.
+                    if (CurrentScreen == exitSource)
                         resumeFrom(exitSource);
```
smoogipoo, line 227: It's not related to this new logic since it happens on master, but this is still wrong as it means the resumed screen is always the current screen if blocked. Just something to note.

```csharp
[Test]
public void TestMakeCurrentIntermediateResumes()
{
    TestScreen screen1 = null;
    TestScreen screen2 = null;
    TestScreen screen3 = null;

    pushAndEnsureCurrent(() => screen1 = new TestScreen());
    pushAndEnsureCurrent(() => screen2 = new TestScreen
    {
        Exiting = () => true
    }, () => screen1);
    pushAndEnsureCurrent(() => screen3 = new TestScreen(), () => screen2);

    AddStep("make screen1 current", () => screen1.MakeCurrent());

    AddAssert("screen3 exited to screen2", () => screen3.ExitedTo == screen2);
    AddAssert("screen2 resumed from screen3", () => screen2.ResumedFrom == screen3); // fails (== screen2)
}
```

smoogipoo (approved):

osu.Framework/Screens/ScreenStack.cs
peppy: yeah, noticed that was a bit weird when going through the logic. may look at a fix.

Issue #2860, opened 2019-09-25
Raw input freezes both tablet and mouse cursor MacOs


CelestialKitty:
**Describe the bug:**  I turn on raw input and that stoped all cursor movement. can still use buttons.

**Screenshots or videos showing encountered issue:** 

**osu!lazer version:** all versions

**Logs:**

Kyrosai:
I'm currently also experiencing this issue, I enabled raw input & it froze my mouse/tablet cursor. Currently I cannot do anything about it, the game isn't letting me disable raw input because the cursor is frozen and it doesn't register clicks whatsoever.  I've attempted to reinstall the game and restart my iMac and so far theres no results.

Issue #2247, opened 2019-03-13
Raw Input "ignores" mouse confinement


Gabixel:
Basically the raw input seems to "ignore" the `confine mouse cursor` feature in some situations. (previous issue: ppy/osu/issues/4463)

[Here's a demonstration](https://drive.google.com/file/d/1B6Dqd7PvHriqn1qfcQpbtnUTVwUuno-y/view?usp=sharing).

In summary, I noticed that (when raw input is enabled):
- The cursor can go outside the screen when in fullscreen/borderless mode if `confine cursor` is set to `never`
- The cursor can go outside the window in windowed mode when the left/right/scroll mouse button is held when `confine cursor` is set to `never` or `fullscreen`

peppy:
See https://github.com/ppy/osu-framework/pull/1408 for the cause of this.

Gabixel:
The issue seems to be fixed with the new "high precision mouse" option. Can be closed.

Issue #2077, opened 2019-01-09
Cursor Unresponsive When "Raw Input" Enabled and Using Trackpad


Zoomreddin:
**Expected behaviour:** Enabling the "raw input" setting will allow the cursor sensitivity to be changed.

**Observed behaviour:** Enabling the "raw input" setting causes the cursor to freeze in the centre of the window, unresponsive to trackpad input. This persists across relaunches of the game until the "raw input" setting is disabled.

**Why this is an issue:** Starting in MacOS Sierra, Apple removed functionality that applications relied on to disable mouse acceleration system-wide. It is now not possible to disable mouse acceleration without keeping the default mouse sensitivity.

**Steps to reproduce:**

1. Launch the game.
2. Open settings.
3. Enable the "raw input" setting.
4. Attempt to move the cursor with the trackpad.

**Notes:** I haven't tried reproducing this with anything other than the system trackpad. If someone could let me know if this can be reproduced with a wireless one that would be great.

**System information:**
Version: MacOS Mojave [10.14.2 (18C54)]
Model: MacBook Air (13-inch, Mid 2012)

peppy:
Can be reproduced with all cursor input devices on macOS. Unsure if the reason is the same one you specify, but can confirm raw input has never worked on lazer on macOS (and has not yet been investigated or addressed).

This is an issue at osu-framework level, so I've moved this issue across to there.

Zoomreddin:
Oops, I should have titled that heading in a clearer way (I meant that there isn't a system workaround to change the sensitivity). Not sure if the problem is related to the changes to the way that the cursor is handled in newer versions, but I'll see if I can reproduce on El Capitan to confirm.

Issue #1378, opened 2018-02-07
Raw input confines mouse incorrectly and moves mouse upon moving the window


MrTheMake:
There are currently two issues with raw input:

1. If you try to move the mouse beyond the resolution's borders, the mouse is confined at the border. If you want to move it back, it jumps back back to the border until all the movement done outside the resolution is reversed by mouse movement
2. Moving the game window moves the mouse to incorrect positions.

Here is a video that shows both behaviours: https://streamable.com/e7hk3 (The first issue may be hard to catch)

peppy:
Does this only happen around the title bar? Also, what version of windows are you on?

MrTheMake:
The same issue happens when resizing the game window. The mouse is incorrectly positioned after resizing it.
I am running Windows 10.

ekrctb:
Does this still happens now (after my PR)?

PR #3960, opened 2020-11-01
SDL raw input

swoolcock:
Implements raw input with the SDL backend.  Should work fine on Windows (tested) and Linux (untested), but macOS is non-trivial and thus a future task.

There were some gotchas that made this take longer than it should have, specifically that events were being invoked in batches rather than sequentially.  The call to `SDL_SetRelativeMouseMode` purges all pending SDL mouse motion events, but it's possible that they have already been polled and added to `eventScheduler`.  I have thus changed the event loop to only handle a single SDL event at a time.  This means that any event invocations can be sure to immediately execute SDL commands before the next invocation.  Note that it is still using the scheduler, so that subclasses can inject events if required (such as with `MacOSWindowBackend`).

While the "raw" part of macOS is not implemented by SDL, the "relative" part is.  This means that the sensitivity slider and map absolute options can be used, but for now the mouse still includes OS acceleration.

"Map Absolute" scales the relative movement based on a window with a long edge of 1024 pixels.  This is somewhat of a guess, since it was hard to match against osuTK's implementation.

Needs testing on Linux, and verification on the implementation's "correctness" based on the user's expectation of the functionality.

Should hopefully resolve all or most of #2860, #2814, #2247, #2077, #1378

Requires a change to osu! to correctly select the raw input mouse handler when running SDL.

### DIFF
--- a/osu.Framework/Platform/DesktopWindow.cs
+++ b/osu.Framework/Platform/DesktopWindow.cs
@@ -71,14 +71,13 @@ namespace osu.Framework.Platform
         protected override IGraphicsBackend CreateGraphicsBackend() => new Sdl2GraphicsBackend();
 
         /// <summary>
-        /// Sets or gets the mouse position in unscaled screen coordinates, relative to the top left corner of the window.
-        /// Setting the mouse position will also disable <see cref="Window.RelativeMouseMode"/> if enabled.
+        /// Enables or disables <see cref="Window.RelativeMouseMode"/> based on the given <paramref name="position"/>.
+        /// If the position is within the window and relative mode is disabled, relative mode will be enabled.
+        /// If the position is outside the window and relative mode is enabled, relative mode will be disabled
+        /// and the mouse cursor will be warped to <paramref name="position"/>.
+        /// <param name="position">The given screen location to check, relative to the top left corner of the window, in logical coordinates. If null, defaults to current position.</param>
         /// </summary>
-        public Vector2 MousePosition
-        {
-            get => WindowBackend.MousePosition;
-            set => WindowBackend.MousePosition = value;
-        }
+        public void UpdateRelativeMode(Vector2? position = null) => WindowBackend.UpdateRelativeMode(position);
 
         public override void SetupWindow(FrameworkConfigManager config)
         {
@@ -112,6 +111,8 @@ namespace osu.Framework.Platform
                 WindowBackend.Size = evt.NewValue;
                 Size.Value = evt.NewValue;
             };
+
+            IsActive.ValueChanged += _ => UpdateRelativeMode();
 
             config.BindWith(FrameworkSetting.SizeFullscreen, sizeFullscreen);
             config.BindWith(FrameworkSetting.WindowedSize, sizeWindowed);
--- a/osu.Framework/Platform/DesktopWindow.cs
+++ b/osu.Framework/Platform/DesktopWindow.cs
@@ -70,6 +70,16 @@ namespace osu.Framework.Platform
         protected override IWindowBackend CreateWindowBackend() => new Sdl2WindowBackend();
         protected override IGraphicsBackend CreateGraphicsBackend() => new Sdl2GraphicsBackend();
 
+        /// <summary>
+        /// Sets or gets the mouse position in unscaled screen coordinates, relative to the top left corner of the window.
+        /// Setting the mouse position will also disable <see cref="Window.RelativeMouseMode"/> if enabled.
+        /// </summary>
+        public Vector2 MousePosition
+        {
+            get => WindowBackend.MousePosition;
+            set => WindowBackend.MousePosition = value;
+        }
+
         public override void SetupWindow(FrameworkConfigManager config)
         {
             base.SetupWindow(config);
--- a/osu.Framework/Platform/DesktopWindow.cs
+++ b/osu.Framework/Platform/DesktopWindow.cs
@@ -75,7 +75,7 @@ namespace osu.Framework.Platform
         /// If the position is within the window and relative mode is disabled, relative mode will be enabled.
         /// If the position is outside the window and relative mode is enabled, relative mode will be disabled
         /// and the mouse cursor will be warped to <paramref name="position"/>.
-        /// <param name="position">The given screen location to check, relative to the top left corner of the window, in logical coordinates. If null, defaults to current position.</param>
+        /// <param name="position">The given screen location to check, relative to the top left corner of the window, in scaled coordinates. If null, defaults to current position.</param>
         /// </summary>
         public void UpdateRelativeMode(Vector2? position = null) => WindowBackend.UpdateRelativeMode(position);
 
--- a/osu.Framework/Platform/DesktopWindow.cs
+++ b/osu.Framework/Platform/DesktopWindow.cs
@@ -32,6 +32,7 @@ namespace osu.Framework.Platform
         private readonly Bindable<DisplayIndex> windowDisplayIndex = new Bindable<DisplayIndex>();
 
         public readonly Bindable<ConfineMouseMode> ConfineMouseMode = new Bindable<ConfineMouseMode>();
+        public readonly BindableBool MapAbsoluteInputToWindow = new BindableBool();
 
         /// <summary>
         /// Gets or sets the window's position on the current screen given a relative value between 0 and 1.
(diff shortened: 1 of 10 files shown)


osu.Framework/Platform/IWindow.cs
```diff
@@ -49,6 +49,11 @@ public interface IWindow : IGameWindow
         /// </summary>
         bool CursorInWindow { get; }
 
+        /// <summary>
+        /// Enables or disables relative mouse mode.
```
peppy: Why is this a bool though? The flow of all this is very confusing to me.

On windows, raw input comes in as either absolute or relative depending on the device. Are you saying that this gets switched depending on the source? The way it's implemented it looks like something we/a consumer is switching

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -482,6 +498,43 @@ public override unsafe void SetIcon(Image<Rgba32> image)
...
+                    SDL.SDL_GetMouseState(out x, out y);
+                else
+                {
+                    x = (int)(position.Value.X / scale);
+                    y = (int)(position.Value.Y / scale);
+                }
+
+                // we use a padding around the window to ensure that a relative mode transition cannot occur while the mouse is over the window decorations
+                // this fixes an issue where mouse clicks could potentially be sent to the window decoration and thus block SDL_PollEvent
+                var relativeMargin = relativeEnabled ? -10 : 10;
```
swoolcock: They're arbitrary values I found worked well after testing.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
```diff
@@ -745,8 +801,13 @@ private void handleMouseButtonEvent(SDL.SDL_MouseButtonEvent evtButton)
...
         }
 
-        private void handleMouseMotionEvent(SDL.SDL_MouseMotionEvent evtMotion) =>
-            ScheduleEvent(() => OnMouseMove(new Vector2(evtMotion.x * scale, evtMotion.y * scale)));
+        private void handleMouseMotionEvent(SDL.SDL_MouseMotionEvent evtMotion)
+        {
+            if (SDL.SDL_GetRelativeMouseMode() == SDL.SDL_bool.SDL_FALSE)
+                ScheduleEvent(() => OnMouseMove(new Vector2(evtMotion.x * scale, evtMotion.y * scale)));
+            else
+                ScheduleEvent(() => OnMouseMoveRelative(new Vector2(evtMotion.xrel, evtMotion.yrel)));
```
swoolcock: Not necessary, since the delta is coming directly from the device and not in screen space.

osu.Framework/Platform/DesktopWindow.cs
```diff
@@ -117,6 +129,8 @@ public override void SetupWindow(FrameworkConfigManager config)
             config.BindWith(FrameworkSetting.ConfineMouseMode, ConfineMouseMode);
             ConfineMouseMode.BindValueChanged(confineMouseModeChanged, true);
 
+            config.BindWith(FrameworkSetting.MapAbsoluteInputToWindow, MapAbsoluteInputToWindow);
```
swoolcock: It's exposed to `RawMouseHandler` in a similar fashion to how `DesktopGameWindow.MapAbsoluteInputToWindow` is exposed to `OsuTKRawMouseHandler`.  I suppose the config could be resolved in the handler instead.

osu.Framework/Platform/Window.cs
```diff
@@ -188,10 +199,19 @@ public bool VerticalSync
...
-        /// Invoked when the user moves the mouse cursor within the window.
+        /// Invoked when the user moves the mouse cursor, if relative mouse mode is disabled.
+        /// The <see cref="Vector2"/> provided is the position relative to the top left corner of the window,
+        /// with DPI scaling applied.
         /// </summary>
         public event Action<Vector2> MouseMove;
 
+        /// <summary>
+        /// Invoked when the user moves the mouse cursor, if relative mouse mode is enabled.
+        /// The <see cref="Vector2"/> provided is the number of pixels the cursor has moved since the previous
```
swoolcock: If you think it's worth using integers instead, we should probably change both `MouseMove` and `MouseMoveRelative` to use `Point`, then.

peppy:
Initial issues noticed on windows (can only test mouse at the moment, need to find cables for tablet):

- When window if not focused, input events aren't handled fast enough. This happens with raw input on and off https://youtu.be/szf3riM-wSE
- Map absolute to window seems to be incorrectly applying to mouse movement. It should only apply to absolute pointer devices.
- Toggling raw input while absolute to window is enabled causes jumps on toggle of 5-10 pixels

osu.Framework/Platform/IWindow.cs
swoolcock: The only way to access "raw input" through SDL is by enabling relative mode, which reports unaccelerated deltas from the mouse rather than absolute positions.  If this is not the way we want to implement raw input, then we cannot do it with SDL alone.

peppy: i'll have to test with a tablet, but this definitely does not sound what we want. and if that's the case, why not rename "relative" bools to "rawInput", since apparently that's what they are doing?

peppy:
mapAbsoluteToScreen is fine to work on mice if it actually does what it says btw. It should make the mouse move from one side of the window to the other over the same physical distance regardless of window size. We'll need this to make cursor sensitivity portable.

osu.Framework/Platform/IWindow.cs
peppy, line 53: what *is* relative mouse mode?

osu.Framework/Platform/IWindowBackend.cs
```diff
@@ -55,6 +55,13 @@ public interface IWindowBackend
         /// </summary>
         bool CursorConfined { get; set; }
 
+        /// <summary>
+        /// Enables or disables relative mouse mode.
+        /// While relative mouse mode is disabled, <see cref="MouseMove"/> events will be fired.
```
peppy, line 60: this doesn't explain when/why i'd want either.

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
peppy, line 526: where do these numbers come from?

osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs
peppy, line 809: what's the reason for not multiplying in scale here? is it actually not necessary?

osu.Framework/Platform/Window.cs
peppy, line 210: pixels in what space? also pixels are not floating point, so may want to rewrite this to refer to the "distance the mouse has moved" or something if it's actually floating point data.

osu.Framework/Platform/DesktopWindow.cs
peppy, line 132: no value changed action? what's the point of this bindable?

peppy (changes requested):

bdach:
It does indeed seem that the SDL mouse API is insufficient for some of the requirements listed above. After some digging I have some good news and some bad news. SDL's tip branch does seem to have a touch input API, which is supposed to also support tablets - see [this readme document](https://hg.libsdl.org/SDL/file/2b1cce20e59b/docs/README-touch.md).

Bad news is (provided the README is even up to date):

> Linux:
> (...) Currently only Wacom tablets are supported.
> (...)
> Windows:
> Unfortunately there is no windows support as of yet. Support for Windows 7 is planned, but we currently have no way to test. (...)

We're kind of going into device driver territory here. Windows can provide raw input via WM_INPUT, and it seems that linux should also be able to via the input subsystem, as long as drivers are written properly ([link](https://www.kernel.org/doc/html/v4.12/input/event-codes.html#guidelines)). I can't see a .NET wrapper over the latter, though...

bdach:
I've tested the current code on linux.

* Can also reproduce the warp on raw input toggle - seems like the raw input handler is not marshaling back its last cursor position to the OS correctly for it to be able to resume where raw input left off.
* I'm also getting very weird behaviour if a framework window with raw input off is behind a normal window. Within the confines of the normal window the cursor works fine, but if you move to the framework window, the OS mouse pointer will show and be locked in place at the center of the framework window, while the game pointer will very choppily attempt to move as requested by the user.

swoolcock:
Looking more at how SDL and various other windowing toolkits handle raw input of pointing devices, it seems that honestly OpenTK's implementation is closest to what we actually want.  It exposes both absolute and relative values from attached pointing devices via `Mouse.GetStates()`, while also calling `MouseMove` based on the OS cursor position.

Both SDL and GLFW can only handle absolute OR relative motion for a single device via the event loop, and the developer needs to select which type of input they want at the time.  GLFW has a similar implementation.

It may be worth just copy-pasting a lot of the platform-specific mouse implementation from osuTK and repurposing it for our requirements.  We know it currently works, and it shouldn't be too much effort.

PR #3962, opened 2020-11-01
Fix ChildrenOfType thread safety when querying while items are being added

peppy:
In addition to #3955, which didn't completely solve the issue as `.ToArray()` can still cause failures (see https://ci.appveyor.com/project/peppy/osu/builds/36065912/tests)


bdach (approved):
This also works, I suppose :+1:

PR #3963, opened 2020-11-01
Revert "Replace SDL_GL_SwapWindow with flushBuffers call on macOS"

peppy:
Poorly tested breaking changes (on window resize game content does not update).


bdach (approved):
No macOS here, but given this is a straight revert I'll throw an approve to speed things up along.

PR #3978, opened 2020-11-07
Allow SDL to use existing raw input implementation

peppy:
Simplest possible implementation. Note that this ends up using osuTK's fallback when mouse is outside the window / non-focused. I don't think this is a huge deal.

- [x] Depends on https://github.com/ppy/osu-framework/pull/3981 (for simplicity of merge, due to renames)

### DIFF
--- a/osu.Framework.Tests/Visual/Input/TestSceneInputManager.cs
+++ b/osu.Framework.Tests/Visual/Input/TestSceneInputManager.cs
@@ -190,7 +190,7 @@ namespace osu.Framework.Tests.Visual.Input
 
         private void setRawInputConfig(bool x)
         {
-            config.Set(FrameworkSetting.IgnoredInputHandlers, x ? nameof(OsuTKMouseHandler) : nameof(OsuTKRawMouseHandler));
+            config.Set(FrameworkSetting.IgnoredInputHandlers, x ? $"{nameof(OsuTKMouseHandler)} {nameof(MouseHandler)}" : nameof(OsuTKRawMouseHandler));
         }
 
         private void setConfineMouseModeConfig(bool x)
--- a/osu.Framework/Platform/GameWindow.cs
+++ b/osu.Framework/Platform/GameWindow.cs
@@ -57,9 +57,6 @@ namespace osu.Framework.Platform
 
         private readonly BindableBool cursorInWindow = new BindableBool();
 
-        /// <summary>
-        /// Provides a read-only bindable that monitors the whether the cursor is in the window.
-        /// </summary>
         public IBindable<bool> CursorInWindow => cursorInWindow;
 
         /// <summary>
--- a/osu.Framework/Platform/GameWindow.cs
+++ b/osu.Framework/Platform/GameWindow.cs
@@ -55,10 +55,12 @@ namespace osu.Framework.Platform
 
         protected readonly Scheduler UpdateFrameScheduler = new Scheduler();
 
+        private readonly BindableBool cursorInWindow = new BindableBool();
+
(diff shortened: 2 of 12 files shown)


osu.Framework/Platform/GameWindow.cs
```diff
@@ -55,10 +55,12 @@ public abstract class GameWindow : IWindow
 
         protected readonly Scheduler UpdateFrameScheduler = new Scheduler();
 
+        private readonly BindableBool cursorInWindow = new BindableBool();
+
         /// <summary>
-        /// Whether the OS cursor is currently contained within the game window.
+        /// Provides a read-only bindable that monitors the whether the cursor is in the window.
```
swoolcock, line 61: Extra "the" in xmldoc.

osu.Framework/Platform/IWindow.cs
```diff
@@ -47,7 +47,7 @@ public interface IWindow : IGameWindow
         /// <summary>
         /// Whether the OS cursor is currently contained within the game window.
```
swoolcock, line 48: Should probably copy the xmldoc from `GameWindow` here.

swoolcock (commented):
Looks good.

osu.Framework/Platform/GameWindow.cs
peppy: Was copied straight from `Window.cs`. All of these classes are a nightmare to navigate and have multiple issues.

bdach:
There's an actual compilation failure on the iOS side with these changes as this assignment is no longer valid:

https://github.com/ppy/osu-framework/blob/c765605199c2cb23223cb38d5d809781337d02c8/osu.Framework.iOS/IOSGameWindow.cs#L27-L28

osu.Framework.Tests/Visual/Input/TestSceneInputManager.cs
```diff
@@ -190,7 +190,7 @@ private void setCursorSensivityConfig(double x)
 
         private void setRawInputConfig(bool x)
         {
-            config.Set(FrameworkSetting.IgnoredInputHandlers, x ? nameof(OsuTKMouseHandler) : nameof(OsuTKRawMouseHandler));
+            config.Set(FrameworkSetting.IgnoredInputHandlers, x ? $"{nameof(OsuTKMouseHandler)} {nameof(MouseHandler)}" : nameof(OsuTKRawMouseHandler));
```
bdach, line 193: Is this logic going to be duplicated game-side for now to make the toggle there work? Right now I'd say the raw input toggle setting handling code is [a little *raw*, too](https://github.com/ppy/osu/blob/8dddd8aff56ac819ee8165e42155a9400b5eb9d1/osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs#L74-L91), and with this change it's probably gonna get a bit worse...

peppy: It's all a bit haphazard right now. Eventually the input subsystem will need a rewrite to better support various different devices. Probably something to consider at a later stage (and just make-this-work for now)

bdach (approved):

smoogipoo (approved):
Seems reasonable to me. Is the eventual goal to use SDL? As it is the raw input implementation is completely broken on linux. Or is the goal to move tk's implementation locally and implement linux/macOS ourselves alongside?

bdach:
We were discussing on discord that eventually we'll probably want per-platform impls (WM_INPUT on windows and something else - sysfs/evdev directly? - on linux).

PR #3979, opened 2020-11-07
Output global statistics to log file when displayed

peppy:
Should help when debugging issues with users and wanting to see a snapshot.

image

### DIFF
--- a/osu.Framework/Statistics/GlobalStatistics.cs
+++ b/osu.Framework/Statistics/GlobalStatistics.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Linq;
 using osu.Framework.Bindables;
+using osu.Framework.Logging;
 
 namespace osu.Framework.Statistics
 {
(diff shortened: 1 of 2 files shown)


bdach (approved):

osu.Framework/Statistics/GlobalStatistics.cs
```diff
@@ -64,5 +65,20 @@ private static void register(IGlobalStatistic stat)
...
+        {
+            Logger.Log("----- Global Statistics -----", LoggingTarget.Performance);
+
+            foreach (var group in GlobalStatistics.Statistics.GroupBy(s => s.Group))
+            {
+                Logger.Log($"# {group.Key}", LoggingTarget.Performance);
+
+                foreach (var i in group)
+                    Logger.Log($"{i.Name.PadRight(30)}: {i.DisplayValue}", LoggingTarget.Performance);
+            }
```
bdach, line 79: actually now that I look at this again, this should probably take the lock to avoid dying on a modification while this is being enumerated - I'll get it

bdach (changes requested):

osu.Framework/Statistics/GlobalStatistics.cs
peppy: yeah, actually noticed this but originally had the code written in a different class (in a safer way..)

bdach (approved):

PR #3980, opened 2020-11-08
Change audio component invalidation logic to use flag instead of delegate

peppy:
Fixes https://github.com/ppy/osu/issues/10727 (the focus-related portion).

### DIFF
--- a/osu.Framework/Audio/AudioComponent.cs
+++ b/osu.Framework/Audio/AudioComponent.cs
@@ -23,6 +23,12 @@ namespace osu.Framework.Audio
         private bool acceptingActions = true;
 
         /// <summary>
+        /// Whether an audio thread specific action can be performed inline.
+        /// </summary>
+        protected bool CanPerformInline =>
+            ThreadSafety.IsAudioThread || (ThreadSafety.ExecutionMode == ExecutionMode.SingleThread && ThreadSafety.IsUpdateThread);
+
+        /// <summary>
         /// Enqueues an action to be performed on the audio thread.
         /// </summary>
         /// <param name="action">The action to perform.</param>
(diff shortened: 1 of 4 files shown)


osu.Framework/Audio/AudioComponent.cs
```diff
@@ -101,8 +105,13 @@ public void Update()
                 task.RunSynchronously();
 
             if (!IsDisposed)
+            {
                 UpdateState();
 
+                if (Looping && HasCompleted)
```
peppy, line 111: Had to move this out to allow calling `UpdateState` in start calls.

osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -263,6 +263,9 @@ public override void Start()
 
         private bool startInternal()
         {
+            // ensure state is correct before starting.
+            UpdateState();
```
peppy, line 267: without this the frequency handling test will fail. it's probably best we call this so there's never any delay between a track starting and its attributes being updated, though.

bdach:
Tested this for a few minutes, seems to resolve the issue at least. The changes look *mostly* safe to me - operation order is changed somewhat (i.e. in `TrackVirtual`) but I can't see it having a major impact on execution/semantics. Definitely not merging on my own, though.

bdach (approved):

smoogipoo:
What part of this fixes the focus issue? Somehow delegates not being cleared out?

bdach:
Yep, tasks being enqueued but never dequeued.

smoogipoo:
And why do the tasks never get dequeued? Isn't the audio thread running much faster than the update thread when unfocused (1000fps compared to the 30fps or whatever the update thread is limited to)?

peppy:
Well there's two parts: the actual action overhead (up to millions of actions being created per second, several megabytes). That is what I'm aiming to solve here.

Anything else needs potential further investigation, like why they are not being processed (somehow not in the alive audio hierarchy? not sure).

smoogipoo:
Ok. I'll merge this for now then as I don't disagree with the underlying changes.

smoogipoo (approved):

PR #3983, opened 2020-11-09
Add SDL support to SampleGame

peppy:
Caught me off-guard that this isn't a thing.

### DIFF
--- a/SampleGame.Desktop/Program.cs
+++ b/SampleGame.Desktop/Program.cs
@@ -2,17 +2,20 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using osu.Framework.Platform;
+using System.Linq;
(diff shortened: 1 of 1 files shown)


bdach:
Think it's worth doing the same for the flappy template while we're at it?

smoogipoo:
I really don't think we should be promoting this argument so far as to add it to the templates/samples... The eventual goal _is_ to remove osuTK, right?

peppy:
Yeah, would be against adding it to the template. But samplegame is kinda required because i use it for benchmarking and almost got numbers incorrect because it wasn't being applied.

bdach:
Oh, the empty template and sample game are two disparate things. My bad, I just associated template with "spinning box" without realising both have it.

smoogipoo (approved):

PR #3987, opened 2020-11-09
Add missing default value specification in OsuTKWindow

bdach:
Presumably closes https://github.com/ppy/osu/issues/10749, although I haven't tested.


bdach:
#3988

PR #3988, opened 2020-11-09
Fix CursorInWindow default value being incorrect in tk window classes

peppy:
Caused android/iOS input to stop working completely


bdach:
I've PR'd this already, but I guess I'll just close mine :smile:

bdach (approved):

PR #3989, opened 2020-11-09
Avoid resetting drawable lifetime on PooledDrawables that aren't assigned a pool

smoogipoo:
Tried a few things osu-side, but I think this is the simplest/least hacky solution.

Will fix the test failures in https://github.com/ppy/osu/pull/10717

### DIFF
--- a/osu.Framework/Graphics/Pooling/PoolableDrawable.cs
+++ b/osu.Framework/Graphics/Pooling/PoolableDrawable.cs
@@ -28,6 +28,12 @@ namespace osu.Framework.Graphics.Pooling
         /// </summary>
         private bool waitingForPrepare;
 
+        /// <summary>
+        /// Whether <see cref="Drawable.LifetimeStart"/> and <see cref="Drawable.LifetimeEnd"/> should be reset to their default values
+        /// every time this <see cref="PoolableDrawable"/> is assigned to a new consumer.
+        /// </summary>
+        public virtual bool ResetLifetimeWhenAssigned => true;
+
         public override bool IsPresent => waitingForPrepare || base.IsPresent;
 
         protected override void LoadComplete()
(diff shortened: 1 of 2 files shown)


bdach (dismissed):
seems fine to have there :+1:

peppy:
For the record, this means that pooled drawable's lifetimes could be changed at arbitrary usages and pollute future usages, yeah? Maybe the xmldoc should come with some kind of warning.

osu.Framework/Graphics/Pooling/PoolableDrawable.cs
```diff
@@ -98,8 +104,11 @@ internal void Assign()
 
             IsInUse = true;
 
-            LifetimeStart = double.MinValue;
-            LifetimeEnd = double.MaxValue;
+            if (ResetLifetimeWhenAssigned)
```
peppy, line 107: Would an `IsInPool` check be enough potentially?

smoogipoo: It's still a bit haphazard, but that'll work too. How would you propose telling the user then that they cannot set lifetimes during the `setupAction` because `Assign()` is called after it?

peppy: Ah, so the actual case here is in `setupAction`? Is there a way we can allow that to run after the lifetime resetting logic here?

smoogipoo: That's just a followup issue. The case failing in the referencing PR is from `LoadComplete()`, which would be fixed by `IsInPool` here. It's also not going to be an issue with hitobject pooling because the process is something like this:
```
var d = pool.Get(() => /* apply mods, apply hitobject */);

d.ApplyLifetime(lifetimeEntry);

return d;
```
So in my case the lifetime is set after the drawable is retrieved (post-`Assign()`). The case I pointed out above is if someone wants to move that `d.ApplyLifetime()` call into the `setupAction`, which would fail.

peppy (approved):

smoogipoo:
I've chosen to take both approaches: hoisting lifetime management out into `DrawablePool.Get()` and also inverting the order of calls to `setupAction` vs `Assign()` because it cleans up `DrawablePool.Get()`.

The first fixes the underlying issue with the osu-side PR. The second slightly changes the way schedules are handled - if `setupAction` schedules something to the drawable, the callback will happen _after_ `PrepareForUse`, which I think is the more expected flow of events and is very edge-case anyway.

peppy (approved):

PR #3991, opened 2020-11-10
Update visual studio code task arguments to use dash instead of slash

peppy:
As per https://github.com/ppy/osu/issues/10721.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -10,8 +10,8 @@
             "args": [
                 "build",
                 "--no-restore",
-                "/p:GenerateFullPaths=true",
-                "/m",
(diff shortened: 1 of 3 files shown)


.vscode/tasks.json
```diff
@@ -10,8 +10,8 @@
             "args": [
                 "build",
                 "--no-restore",
-                "/p:GenerateFullPaths=true",
-                "/m",
+                "-p:GenerateFullPaths=true",
+                "-m",
                 "/verbosity:m",
```
smoogipoo, line 15: And this (missing in entire PR).

smoogipoo (changes requested):

bdach (approved):

PR #3993, opened 2020-11-11
Fix initial application of state not occurring if invalidation not pending

peppy:
Turns out calling `UpdateState` is not enough since it still checks the `pendingInvalidation` flag.

Closes https://github.com/ppy/osu/issues/10778.

### DIFF
--- a/osu.Framework/Audio/Sample/SampleChannelBass.cs
+++ b/osu.Framework/Audio/Sample/SampleChannelBass.cs
@@ -82,12 +82,7 @@ namespace osu.Framework.Audio.Sample
 
                 relativeFrequencyHandler.SetChannel(channel);
                 bassAmplitudeProcessor?.SetChannel(channel);
-            });
 
-            InvalidateState();
-
-            EnqueueAction(() =>
-            {
                 // ensure state is correct before starting.
                 InvalidateState();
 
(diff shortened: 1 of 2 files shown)


osu.Framework/Audio/Sample/SampleChannelBass.cs
```diff
@@ -82,14 +82,9 @@ public override void Play(bool restart = true)
...
                 bassAmplitudeProcessor?.SetChannel(channel);
-            });
-
-            InvalidateState();
 
-            EnqueueAction(() =>
-            {
                 // ensure state is correct before starting.
-                UpdateState();
+                InvalidateState();
```
smoogipoo, line 87: Are you sure this doesn't need to be
```
InvalidateState()
UpdateState()
```
As far as I can see `InvalidateState` doesn't update itself.

peppy: Tested this and just `InvalidateState` was enough to pass tests and seemingly work without any regressions. Was tempted to leave it there as well but not sure?

smoogipoo: Looks like the `UpdateState()` addition is actually new from the past 3 days, so...

smoogipoo (approved):

PR #3994, opened 2020-11-11
Remove IWindow inheritance of IGameWindow

peppy:
First step towards sanity. This reduces the mental burden as we progressively simplify the `IWindow` portion (no need to worry about potentially hidden usage/implementation of underlying osuTK classes).

I'd recommend we test this for a bit (and make sure to get test coverage on all OSes) to ensure nothing was missed. No hurry to merge, I will continue building off this PR.

- [x] Test macOS (SDL/osuTK)
- [x] Test android
- [x] Test iOS
- [x] Test windows (special attention to raw input)
- [x] Test linux


osu.Framework/Platform/Window.cs
```diff
@@ -78,15 +75,33 @@ public bool VerticalSync
         /// </summary>
         public Bindable<Size> Size { get; } = new BindableSize();
 
+        public Size ClientSize
+        {
+            get => Size.Value;
+            set => Size.Value = value;
+        }
+
+        public WindowState WindowState
```
peppy, line 84: i've added these temporarily just to ease the transition. they would likely be removed in a next stage. not so simple as these are part of the quite-convoluted event flow (and deal with volatile variables that can either be changed by the framework or the OS).

osu.Framework/Platform/Window.cs
```diff
@@ -78,15 +75,33 @@ public bool VerticalSync
         /// </summary>
         public Bindable<Size> Size { get; } = new BindableSize();
 
+        public Size ClientSize
+        {
+            get => Size.Value;
+            set => Size.Value = value;
+        }
```
bdach, line 82: So this is unfortunately quite confusing... In other places (`Sdl2WindowBackend`), `Size` and `ClientSize` are two separate things. The former is the screen-space client window size (as returned by the [`SDL_GetWindowSize`](https://wiki.libsdl.org/SDL_GetWindowSize) function), and the latter is the actual drawable area for the screen buffer (as returned by [`SDL_GL_GetDrawableSize`](https://wiki.libsdl.org/SDL_GL_GetDrawableSize)). That *sounds* the same, but as documented there, it isn't in high-DPI/scaling scenarios. `ClientSize` used to delegate this to backend, so I think this does change behaviour.

I've looked at these back and forth a few times and I'm still not sure if this is ok... I *think* I managed to break it once (separate to unrelated stack overflow I mentioned on discord) by fiddling with scaling but can't reproduce again.

osu.Framework/Platform/Window.cs
```diff
@@ -483,225 +492,34 @@ private void size_ValueChanged(ValueChangedEvent<Size> evt)
...
-        public void ProcessEvents()
-        {
-        }
-
         public virtual Point PointToClient(Point point) => point;
 
         public virtual Point PointToScreen(Point point) => point;
 
-        public Icon Icon { get; set; }
+        public Rectangle Bounds { get; set; }
```
bdach, line 519: Seems unused now (was part of `IGameWindow` interface).

osu.Framework/Platform/IWindow.cs
```diff
@@ -3,18 +3,18 @@
...
 using JetBrains.Annotations;
 using osu.Framework.Bindables;
 using osu.Framework.Configuration;
-using osuTK.Platform;
 
 namespace osu.Framework.Platform
 {
     /// <summary>
     /// Interface representation of the game window, intended to hide any implementation-specific code.
     /// Currently inherits from osuTK; this will be removed as part of the <see cref="OsuTKWindow"/> refactor.
```
bdach, line 15: This line can be removed now.

osu.Framework/Platform/Window.cs
peppy: Yeah, I haven't looked at the backend implementation yet, so this is a best-effort based on my recollection that sdl code was only ever considering the client size.

bdach:
Didn't seem to be behaving worse than `master` in my linux testing yesterday. I think this is a good step forward - ignoring the client size thing for now as DPI scale is to be nuked later in the series & reimplemented from scratch anyway from what I gathered on discord.

peppy:
Tested on windows (via the last PR in this chain). Everything expected to be working is working.

bdach (approved):
now that we know all platforms should be working

smoogipoo (approved):

PR #3995, opened 2020-11-12
Tidy up some aspects on OsuTKWindow

peppy:
- Renames `Implementation` to `OsuTKGameWindow`.
- Removes static initialisation weirdness in mobile classes
- [x] Depends on https://github.com/ppy/osu-framework/pull/3994

Have tested this on both android and iOS.


osu.Framework.iOS/IOSGameWindow.cs
```diff
@@ -3,35 +3,39 @@
...
 
-        public IOSGameWindow()
-            : base(GameView)
+        public override void SetupWindow(FrameworkConfigManager config)
         {
+            Resize += onResize;
         }
 
-        public override void SetupWindow(FrameworkConfigManager config)
+        public IOSGameWindow([NotNull] IOSGameView gameView)
```
bdach, line 26: out of curiosity, why annotate just this one as `[NotNull]` and not the android one?

osu.Framework/Platform/OsuTKWindow.cs
```diff
@@ -169,11 +167,9 @@ protected OsuTKWindow(int width, int height)
         {
         }
 
-        /// <summary>
-        /// Not applicable to osuTK.
-        /// </summary>
         public void Create()
         {
+            Context.MakeCurrent(null);
         }
```
bdach, line 173: Side note, unrelated to the mergeability of this PR - when I went to check where this is being called I saw this:

https://github.com/ppy/osu-framework/blob/ac03ac835cda73303f89b6ba2fc9906152a38663/osu.Framework/Platform/GameHost.cs#L538-L549

(`CreateWindow()` then `window.Create()`). Should probably be rectified at a later stage as it reads badly.

osu.Framework.iOS/IOSGameWindow.cs
peppy: was probably automated

smoogipoo (approved):

Issue #3990, opened 2020-11-10
IWindow hierarchy needs serious refactoring


peppy:
Currently working with (trying to follow control flow) `IWindow` is nigh impossible.

I've tried twice now to refactor it but it's too complex to handle, so a rewrite from scratch may be the better approach. This *needs* to happen before progressing further with SDL2.

The issues I ran into while trying to do this:

- Knowing what actually works (and is implemented via SDL). A lot of properties, bindables etc. exist but aren't actually hooked up. But due to all the control flow there's 2-4 events/methods/bindables related to them to the point it can take over a minute of following the trail to understand if they are being used or actually have an implementation anywhere
- Keeping mental track of the flow and correctly moving implementation across from `WindowBackend` to `Window`. Especially difficult when implementation exist as bindables at one side and properties at the other.
- The urge to destroy things from dealing with the code (may be a personal issue).

My best effort is on [this branch](https://github.com/ppy/osu-framework/compare/master...peppy:remove-window-backend?expand=1) but likely won't be useful for anyone looking at this issue in the future (myself included).

Proposals:

- Imperative: `IWindowBackend` is removed. There's no need for this to exist and it is the main issue. The implementation should be in `Window`/`DesktopWindow`.
- Optional: Remove `OsuTK` completely. This can be done if it's seen to ease the process of refactoring the hierarchy in general. This would see the removal of `IGameWindow`.
- Optional: All non-bindables should be converted to bindables in `IWindow`/`Window`. Things like `CursorState` and `VerticalSync`
- Optional: Remove all properties/bindables which aren't consumed by the framework. Bring implementation down to the MVP to allow the refactor to be completed.

smoogipoo:
@swoolcock because it's been long forgotten - what was the reason for having `IWindowBackend` in the first place?

swoolcock:
> @swoolcock because it's been long forgotten - what was the reason for having `IWindowBackend` in the first place?
So that we can use a different windowing implementation on non-desktop platforms if required.

bdach:
There was this ole diagram that seems to have been eaten up in the edit history of #2448 that's perhaps demonstrative to the idea:

![obraz](https://user-images.githubusercontent.com/20418176/98649115-7f3fe780-2337-11eb-84e6-3276894bc3f9.png)

peppy:
[...]

PR #3996, opened 2020-11-12
Remove concept of window backend

peppy:
- Closes https://github.com/ppy/osu-framework/issues/3990
- [x] Depends on #3995
- [x] Need to figure why external resizes aren't updating the reported size immediately
- [x] Properly handle fullscreen support (doesn't work on master anyway soo)

This fixes too many issues to list, and makes too many refactors to explain. Not sure where to start, but recommend reviewing this using full file passes, not diff. Have tested that mostly everything works (equal to or more than previously) apart from the listed task list above.

This is by no means a final pass on this code. Just trying to get things into a roughly manageable state to figure out the remaining issues and continue to improve the readability.


bdach (commented):
Skimmed through this once. Seems saner although I will definitely have missed things as `DesktopWindow` is still pretty bad reading due to everything being in one class, sometimes intertwined between responsibilities. Extracting partials might be a future step for this code.

The code runs on linux, but I suspect it might not be working properly with the osuTK raw mouse handler in windowed mode (seems to offset mouse position by position of window top-left corner).

osu.Framework/Platform/DesktopWindow.cs
```diff
@@ -2,107 +2,1054 @@
...
 namespace osu.Framework.Platform
 {
     /// <summary>
-    /// Implementation of <see cref="Window"/> used for desktop platforms.
-    /// Uses <see cref="Sdl2WindowBackend"/> and <see cref="Sdl2GraphicsBackend"/> by default.
+    /// Implementation of <see cref="IWindow"/> that provides bindables and
+    /// delegates responsibility to window and graphics backends.
     /// </summary>
-    public class DesktopWindow : Window
+    public class DesktopWindow : IWindow
```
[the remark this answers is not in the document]
peppy: I'm open to changing it to this. Really depends on whether we consider sdl the "default" or just one of many options.

SampleGame.Desktop/Program.cs
```diff
@@ -13,9 +13,9 @@ public static class Program
         [STAThread]
         public static void Main(string[] args)
         {
-            bool useSdl = args.Contains(@"--sdl");
+            bool useOpenTK = args.Contains(@"--tk");
 
-            using (GameHost host = Host.GetSuitableHost(@"sample-game", useSdl: useSdl))
+            using (GameHost host = Host.GetSuitableHost(@"sample-game", useSdl: !useOpenTK))
```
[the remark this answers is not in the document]
peppy: hmm, could do. not sure.

osu.Framework/Platform/DesktopWindow.cs
```diff
@@ -2,107 +2,1054 @@
...
+        /// Gets the native window handle as provided by the operating system.
+        /// </summary>
+        public IntPtr WindowHandle
+        {
+            get
+            {
+                if (SdlWindowHandle == IntPtr.Zero)
+                    return IntPtr.Zero;
+
+                var wmInfo = getWindowWMInfo;
```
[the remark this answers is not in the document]
peppy: oops, was supposed to be a method.

peppy:
I've fixed up all the remaining issues I know of. Fullscreen now works correctly on windows (borderless not so much, yet). It doesn't work on macOS but that can be fixed later, needs further investigation.

osu.Framework/Input/Handlers/Keyboard/KeyboardHandler.cs
```diff
@@ -16,7 +16,7 @@ public class KeyboardHandler : InputHandler
 
         public override bool Initialize(GameHost host)
         {
-            if (!(host.Window is DesktopWindow window))
+            if (!(host.Window is SDL2DesktopWindow window))
```
bdach, line 19: This is something I didn't anticipate with the rename of `DesktopWindow` - now every seemingly generic keyboard/joystick handler refuses to initialise with anything but SDL. I wouldn't muddle this diff with changes in that regard but a next step would be probably to rename these handlers to match as they're windowing-toolkit-specific as they are anyway.

osu.Framework/Host.cs
```diff
@@ -12,7 +12,7 @@ namespace osu.Framework
 {
     public static class Host
     {
-        public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = false, bool portableInstallation = false, bool useSdl = false)
+        public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = false, bool portableInstallation = false, bool useOsuTK = true)
```
bdach, line 15: Hmm... so all existing framework consumers stay on osuTK, but get `(legacy osuTK)` tacked onto their window names... Not saying that's bad, just a thought that occurred in my head.

bdach (commented):
Looks mostly good to go, just a couple of notes in passing.

osu.Framework/Host.cs
peppy: oh this should be false, good catch.

osu.Framework/Input/Handlers/Keyboard/KeyboardHandler.cs
peppy: another good catch.

peppy: Touching on this again now that i've checked the code: I think the path forward here is replacing the conditions with `IWindow`, or removing it altogether.

The only reason this cannot be done is due to the existence of `OsuTK` legacy classes, which aren't hooked up correctly with the new event flow. Once that is removed all new `IWindow` implementations should be able to use the default input handlers in most cases.

bdach: Yep, definitely could definitely agree with either. Any way you cut it this is not a thing to be fixed by this PR.

smoogipoo (approved):
Let's do this

PR #3998, opened 2020-11-12
Update tools to allow InspectCode to run in non-Windows environment

ekrctb:
ReSharper command-line tool and `NVika` is updated to the latest version. The new versions publish itself as `dotnet tool` and it makes the build script simpler. The new versions also allow `InspectCode` target of the Cake build script to run in non-Windows machines.

There were new ReSharper errors due to the version difference. I used `!` (null-forgiving) operator to remove new "positive null reference" warnings. I also added a setting to prevented `*.h` shader header files to be analyzed as C++ header files and causing errors.


bdach:
The `!` operator will be a sticking point. I strongly dislike it myself and @peppy had similarly negative feelings about it as I recall. My rationale is that it spreads badness all over the code and encourages a mentality of "I'll just silence this" without thinking, therefore watering down meaning. Nullability annotations should be near the source, not at each usage.

Unfortunately you can't place a `[DisallowNull]` annotation inside of an anonymous tuple type declaration, so there's not many great alternative options available. The ones I'm thinking of would be to either define our own actual types (kinda over-engineered) or to outright `#nullable disable`, but honestly I think I'd prefer both to doing that `!` thing.

smoogipoo:
I'm not going to bother arguing in support for the operator, but the last time we disagreed I added the `AsNonNull()` extension method. https://github.com/ppy/osu-framework/blob/ac03ac835cda73303f89b6ba2fc9906152a38663/osu.Framework/Extensions/ObjectExtensions/ObjectExtensions.cs#L24-L29

bdach:
I guess I'll also leave a [link to the previous discord discussion about this](https://discord.com/channels/188630481301012481/589331078574112768/719874374655803482) for reference.

ekrctb:
I don't have a strong opinion on the null forgiving operator but I agree that abusing it should be discouraged.
But I think `AsNonNull` is not better than using `!` if `NullReferenceException` will be thrown immediately if the value is `null`.

In this case, however, the issue is with the tuple type. I made ad-hoc `structs. All those types are internal usage and it shouldn't be a breaking change. C#9 record types will make this kind of ad-hoc type a bit simpler.

peppy (approved):
looks good!

PR #3999, opened 2020-11-12
Add transform preview to draw visualiser

bdach:
This PR adds a way to preview transform application progress from the draw visualiser. See [video showing preview of behaviour](https://drive.google.com/file/d/1cDzC90larwbeEQEBmOLcJdVrZV_dmu7t/view?usp=sharing).

Display code partially shamelessly stolen from existing test scene and wedged into the draw visualiser. Allocating drawables every frame is admittedly wasteful, but seeing that it's for debugging purposes... Tracking transforms externally is not an easy task anyway (I fiddled some with `WeakReference`s but my attempts appeared crazily unsafe - could seemingly die on a nullref even if `TryGetTarget` returned `true`). For what it's worth I confirmed that the code doesn't run unless the tab is specifically opened.

Mostly usable for rewindable transforms. I made a proof-of-concept of also keeping non-rewindable transform preview for like half a second after removal, but it honestly wasn't very usable. It was also similarly unsafe because of either having to fiddle with `WeakReference`s and not being able to tell if they're still valid, or having to hold references to the transform from the drawable and make sure that they are eventually removed.

### DIFF
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
@@ -11,6 +11,7 @@ using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
 using osu.Framework.Graphics.Transforms;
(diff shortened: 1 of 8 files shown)


peppy (approved):
Seems fine for a first pass

PR #4006, opened 2020-11-18
Allow dynamic recompilation to access internal members

smoogipoo:
This is a bit of a proof-of-concept and I'm not sure on the viability of it going forward, ready for merge if you think it's fine. It fixes the `TestSceneDrawableJudgement -> LegacyJudgementPieceNew` case pointed out in Discord (accessing `SkinInfo.DEFAULT_SKIN`, etc).

See https://www.strathweb.com/2018/10/no-internalvisibleto-no-problem-bypassing-c-visibility-rules-with-roslyn/ for more info.

There is an internal compiler attribute `IgnoresAccessChecksTo(assemblyName)` which acts inversely to `InternalsVisibleTo(assemblyName)`. However it:
1. Requires setting internal compiler option `TopLevelBinderFlags`, and thus the API surface may change.
2. Requires importing all `internal` symbols, which can be an issue because it means there is the possibility of type conflicts in matching namespaces. This already happens with `JetBrains.Annotations`, which libraries define themselves and conflict with the nuget `JetBrains.Annotations` package.
    * To get around this, I'm excluding all assemblies that contain a `JetBrains.Annotations` namespace themselves, and only including the nuget package.
    * This is not an exhaustive search - if two libraries included the same namespace+type combo that is not `JetBrains.Annotations`, DCC would fail.
    * `Microsoft.EntityFrameworkCore` assemblies all get excluded so types referencing EF symbols can't be dynamically recompiled. This isn't a huge issue for us because all of our EF stuff is done in manager classes that can't be recompiled anyway (they're too far-reaching).

I haven't yet found a way to get past (2), and if it becomes a problem then we may have to reconsider...

### DIFF
--- a/osu.Framework/Testing/AssemblyReference.cs
+++ b/osu.Framework/Testing/AssemblyReference.cs
@@ -6,7 +6,7 @@ using Microsoft.CodeAnalysis;
 
 namespace osu.Framework.Testing
 {
-    public readonly struct AssemblyReference
+    internal readonly struct AssemblyReference
     {
         public readonly Assembly Assembly;
         public readonly bool IgnoreAccessChecks;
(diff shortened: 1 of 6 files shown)


peppy:
Is it fine if i gloss over the details in this one and exclaim "it works! (for now)" ?

peppy (approved):

smoogipoo:
I think there shouldn’t be any issues for us, at least.

PR #4009, opened 2020-11-19
Change default recursive parameter on BeginSequence methods to true

peppy:
True is the more common intended use case, but due to the parameter being optional regularly gets forgotten, resulting in broken behaviuor and a hard-to-debug situation at hand.

### BeginAbsoluteSequence and BeginDelayedSequence recurse to children by default

The optional `recursive` argument has changed default value from `false` to `true` to cover the more common usage scenario and avoid developer introduced bugs. Please check any usages you have of these functions to avoid behaviour changes.


bdach:
I fixed a few tiny nits in the docs.

Pretty scary breaking change, but I can get behind it. Haven't used transform sequences too much but already caught myself not specifying `true` when I intended to a few times.

osu.Framework/Graphics/Transforms/ITransformable.cs
```diff
@@ -14,15 +14,15 @@ public interface ITransformable
...
-        /// <returns>A <see cref="InvokeOnDisposal"/> to be used in a using() statement.</returns>
+        /// <returns>An <see cref="InvokeOnDisposal"/> to be used in a using() statement.</returns>
         IDisposable BeginDelayedSequence(double delay, bool recursive = true);
 
         /// <summary>
         /// Start a sequence of <see cref="Transform"/>s from an absolute time value (adjusts <see cref="TransformStartTime"/>).
         /// </summary>
         /// <param name="newTransformStartTime">The new value for <see cref="TransformStartTime"/>.</param>
-        /// <param name="recursive">Whether this should be applied to all children.</param>
-        /// <returns>A <see cref="InvokeOnDisposal"/> to be used in a using() statement. True by default.</returns>
```
peppy: oops.

smoogipoo (approved):

PR #4011, opened 2020-11-20
Fix crash in FlappyDon sample project

OliBomby:
FlappyDon crashes during the normal play and the Obstacles test. This PR fixes both.

### DIFF
--- a/osu.Framework.Templates/templates/template-flappy/FlappyDon.Game/Elements/Obstacles.cs
+++ b/osu.Framework.Templates/templates/template-flappy/FlappyDon.Game/Elements/Obstacles.cs
@@ -158,7 +158,7 @@ namespace FlappyDon.Game.Elements
                 crossedThresholdCount++;
 
                 // Alert the observer that the threshold was crossed in this update loop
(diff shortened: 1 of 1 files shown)


osu.Framework.Templates/templates/template-flappy/FlappyDon.Game/Elements/Obstacles.cs
```diff
@@ -127,14 +133,23 @@ protected override void Update()
...
-                    // Increase the obstacle count, which will reset threshold detection
-                    // for the pipe after this one.
-                    obstacleCount++;
+                    // Add the obstacle to the stack to remove later
+                    // because InternalChildren can't be updated while being enumerated.
+                    obstaclesToRemove.Push(obstacle);
                 }
             }
 
+            while (obstaclesToRemove.Count > 0)
```
peppy, line 142: to tidy this up you can use `while (obstaclesToRemove.TryPop(out var obstacle)) ...`

peppy (changes requested):

smoogipoo (approved):
Works as expected, thanks!

PR #4013, opened 2020-11-20
Rewrite GLWrapper's scheduled disposal logic to avoid delegate allocations

peppy:
Needs sanity testing (and benchmarks)


bdach:
I've briefly read the source (seemed sane) and resolved the CI inspection but done no further testing than that. Would need to run windows to benchmark this but might fire it up tomorrow.

peppy:
Centipede single run:

image

I tried to add a headless benchmark, but GLWrapper.Initialize is kind of not run in headless contexts and therefore it's not easy to make work.

osu.Framework/Graphics/OpenGL/GLWrapper.cs
```diff
@@ -112,26 +112,50 @@ internal static void Initialize(GameHost host)
...
-                else
-                    scheduleNextDisposal();
-            });
+                var item = pending_disposal_actions[i];
+
+                if (item.RemainingFrameDelay-- == 0)
+                {
+                    item.Action();
+                    lock (pending_disposal_actions)
+                        pending_disposal_actions.Remove(item);
```
smoogipoo, line 155: Use `RemoveAt(i)` to avoid an internal `IndexOf()`, though I don't think that would be much because: I think it would also be instructive to add a `Debug.Assert(i == 0)`.

osu.Framework/Graphics/OpenGL/GLWrapper.cs
```diff
@@ -112,26 +112,50 @@ internal static void Initialize(GameHost host)
             GL.Enable(EnableCap.Blend);
 
             IsInitialized = true;
+
+            host.UpdateThread.Scheduler.AddDelayed(checkPendingDisposals, 0, true);
```
smoogipoo, line 116: This is incorrect. The initial scheduling was done from the update thread (for whatever reason, I don't remember, need to blame), but the actual disposal (`item.Action()`) should be run on the `reset_scheduler` (draw thread).

smoogipoo (changes requested):

smoogipoo (approved):

PR #4015, opened 2020-11-20
Do not include bindable properties in UnbindAllBindables

### LINKED ISSUE #4014: Reconsider whether bindable unbinding should consider properties
Auto-properties will have their backing fields picked up by the field unbinds, so this will only affect properties that reference external bindables. I think that's pretty dodgy to do in general.

OliBomby:
closes #4014

As mentioned in the issue, unbinding external bindables is pretty dodgy. I updated the unbinding code and changed the test to test for the opposite behaviour.

### DIFF
--- a/osu.Framework.Tests/Bindables/BindableBindingTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableBindingTest.cs
@@ -377,7 +377,7 @@ namespace osu.Framework.Tests.Bindables
         }
 
         [Test]
-        public void TestUnbindOnDrawableDisposeProperty()
+        public void TestUnbindOnDrawableDoNotDisposeDelegatingProperty()
         {
             var bindable = new Bindable<int>();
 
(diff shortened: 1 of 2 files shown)


bdach:
I've expanded test coverage to ensure auto-properties (not delegating ones) are indeed auto-unbound and added a clarifying comment where the property unbound code used to be. Seems good to me otherwise.

bdach (approved):

peppy (dismissed):

osu.Framework.Tests/Bindables/BindableBindingTest.cs
```diff
@@ -392,7 +392,27 @@ public void TestUnbindOnDrawableDisposeProperty()
             drawable.Dispose();
             valueChanged = false;
 
-            drawable.SetValue(2);
+            bindable.Value = 2;
+            Assert.IsTrue(valueChanged, "bound correctly");
+
+            drawable.SetValue(3);
+            Assert.IsTrue(valueChanged, "bound correctly");
```
smoogipoo, line 399: valueChanged needs to be reset before the SetValue, otherwise assert can pass incorrectly.

smoogipoo (changes requested):

smoogipoo (approved):
lgtm

PR #4016, opened 2020-11-23
Add hexagonal container

harshavb:
Thanks to Poyo for all the help. Optimally uses 4 triangles per hexagon.

### DIFF
--- a/osu.Framework/Graphics/Containers/HexagonalContainer.cs
+++ b/osu.Framework/Graphics/Containers/HexagonalContainer.cs
@@ -1,4 +1,7 @@
-﻿using System.Collections.Generic;
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Collections.Generic;
 using System;
 using osu.Framework.Allocation;
 using osu.Framework.Graphics.Colour;
--- a/osu.Framework/Graphics/Containers/HexagonalContainer.cs
+++ b/osu.Framework/Graphics/Containers/HexagonalContainer.cs
@@ -1,0 +1,141 @@
+﻿using System.Collections.Generic;
+using System;
+using osu.Framework.Allocation;
+using osu.Framework.Graphics.Colour;
+using osu.Framework.Graphics.OpenGL.Buffers;
+using osu.Framework.Graphics.OpenGL;
+using osu.Framework.Graphics.Primitives;
+using osu.Framework.Graphics.Shaders;
+using osuTK.Graphics.ES30;
+using osuTK.Graphics;
+using osuTK;
+
+namespace osu.Framework.Graphics.Containers
+{
+    public class HexagonalContainer : HexagonalContainer<Drawable> { }
+
+    public class HexagonalContainer<T> : BufferedContainer<T>
+        where T : Drawable
+    {
+        private static readonly float sin_pi_over_3 = (float)Math.Sin(Math.PI / 3);
+        private static readonly float tan_pi_over_3 = (float)Math.Tan(Math.PI / 3);
+
+        public static readonly float HEXAGON_INRADIUS = sin_pi_over_3;
+
+        public IShader Shader { get; private set; }
+
+        private readonly HexagonalContainerDrawNodeSharedData sharedData;
+
+        public HexagonalContainer(RenderbufferInternalFormat[] formats = null, bool pixelSnapping = false) =>
+            sharedData = new HexagonalContainerDrawNodeSharedData(formats, pixelSnapping);
+
+        [BackgroundDependencyLoader]
+        private void load(ShaderManager shaders) => Shader = shaders.Load(VertexShaderDescriptor.TEXTURE_2, "TextureHexagon");
+
+        protected override DrawNode CreateDrawNode() => new HexagonalContainerDrawNode(this, sharedData);
+
+        public override bool ReceivePositionalInputAt(Vector2 screenSpacePos)
+        {
+            Vector2 norm = screenSpacePos - ScreenSpaceDrawQuad.TopLeft;
+            norm = new Vector2(norm.X / ScreenSpaceDrawQuad.Width, norm.Y / ScreenSpaceDrawQuad.Height); // apparently we can't divide Vector2s?
+            norm = (norm - new Vector2(0.5f)) * 2;
+
+            // hexagons are horizontally and vertically symmetrical, so we only have to test one quadrant :)
+            norm = new Vector2(Math.Abs(norm.X), Math.Abs(norm.Y));
+
+            if (norm.Y > sin_pi_over_3)
+            {
+                return false; // top bound
+            }
+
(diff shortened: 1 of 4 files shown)


bdach:
Do you have permission from @Poyo-SSB to PR this back upstream? Kind of dodgy that you're taking authorship credit on the commits, too.

For the record, original implementation was made in https://github.com/harshavb/piously/pull/13 and https://github.com/harshavb/piously/pull/21.

harshavb:
Fair enough. Will re-open if granted permission.

Poyo-SSB:
Permission granted. I can make adjustments and do rewrites as necessary.

bdach:
reopening for review then, I guess.

@Poyo-SSB would you like (co-)authorship credit on the existing commits? probably one of the rare cases we could allow for a force-push.

Poyo-SSB:
I don't mind either way. @harshavb can take care of that, probably.

osu.Framework/Graphics/Containers/HexagonalContainer.cs
```diff
@@ -0,0 +1,144 @@
...
+using osuTK.Graphics.ES30;
+using osuTK.Graphics;
+using osuTK;
+
+namespace osu.Framework.Graphics.Containers
+{
+    public class HexagonalContainer : HexagonalContainer<Drawable> { }
+
+    public class HexagonalContainer<T> : BufferedContainer<T>
+        where T : Drawable
```
bdach, line 21: is this generic all that necessary? why not just stick with `HexagonalContainer`?

osu.Framework/Graphics/Containers/HexagonalContainer.cs
```diff
@@ -0,0 +1,144 @@
...
+    {
+        private static readonly float sin_pi_over_3 = (float)Math.Sin(Math.PI / 3);
```
bdach, line 23: fancy way of writing `0.5f`, but sure

osu.Framework/Graphics/Containers/HexagonalContainer.cs
```diff
@@ -0,0 +1,144 @@
...
+        private static readonly float tan_pi_over_3 = (float)Math.Tan(Math.PI / 3);
+
+        public static readonly float HEXAGON_INRADIUS = sin_pi_over_3;
```
bdach, line 26: unused

osu.Framework/Graphics/Containers/HexagonalContainer.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        [BackgroundDependencyLoader]
+        private void load(ShaderManager shaders) => Shader = shaders.Load(VertexShaderDescriptor.TEXTURE_2, "TextureHexagon");
+
+        protected override DrawNode CreateDrawNode() => new HexagonalContainerDrawNode(this, sharedData);
+
+        public override bool ReceivePositionalInputAt(Vector2 screenSpacePos)
+        {
+            Vector2 norm = screenSpacePos - ScreenSpaceDrawQuad.TopLeft;
+            norm = new Vector2(norm.X / ScreenSpaceDrawQuad.Width, norm.Y / ScreenSpaceDrawQuad.Height); // apparently we can't divide Vector2s?
```
bdach, line 43: it is actually possible, just not via operator overload

```suggestion
            norm = Vector2.Divide(norm, ScreenSpaceDrawQuad.Size);
```

osu.Framework/Graphics/Containers/HexagonalContainer.cs
```diff
@@ -0,0 +1,144 @@
...
+
+                FrameBuffer current = SharedData.CurrentEffectBuffer;
+                FrameBuffer target = SharedData.GetNextEffectBuffer();
+
+                GLWrapper.SetBlend(BlendingParameters.None);
+
+                using (BindFrameBuffer(target))
+                {
+                    float resolution = Math.Max(Source.ScreenSpaceDrawQuad.Width, Source.ScreenSpaceDrawQuad.Height); // shh
+                    hexagonShader.GetUniform<float>("g_Resolution").UpdateValue(ref resolution);
```
bdach, line 104: what's with this comment?

osu.Framework/Graphics/Shapes/Hexagon.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                Primitives.Hexagon drawingHex = toHexagon(ScreenSpaceDrawQuad);
+                DrawTriangle(Texture, drawingHex.FarUpTriangle, DrawColourInfo.Colour, null, null,
+                    new Vector2(InflationAmount.X / DrawRectangle.Width, InflationAmount.Y / DrawRectangle.Height), TextureCoords);
+                DrawTriangle(Texture, drawingHex.NearUpTriangle, DrawColourInfo.Colour, null, null,
+                    new Vector2(InflationAmount.X / DrawRectangle.Width, InflationAmount.Y / DrawRectangle.Height), TextureCoords);
+                DrawTriangle(Texture, drawingHex.NearDownTriangle, DrawColourInfo.Colour, null, null,
+                    new Vector2(InflationAmount.X / DrawRectangle.Width, InflationAmount.Y / DrawRectangle.Height), TextureCoords);
+                DrawTriangle(Texture, drawingHex.FarDownTriangle, DrawColourInfo.Colour, null, null,
+                    new Vector2(InflationAmount.X / DrawRectangle.Width, InflationAmount.Y / DrawRectangle.Height), TextureCoords);
```
bdach, line 50: can these be a method at least or something? duplicated 8 times across 2 methods, only differing with the triangle involved.

harshavb:
Poyo should take full authorship credit, as all the code here is written by her.

bdach (changes requested):

bdach:
oh, and this PR should also include the test scene.

Poyo-SSB:
`HexagonalContainer` is generic for the same reason that other Containers are generic—for ease of access to single-type children, yes? I don't know if there are any performance optimizations that come with that or anything.

As for the comment, I'm unsure if passing in the resolution as a uniform is the canonical method of having screenspace-aware shaders. I wasn't able to find the way that CompositeDrawable does this.

bdach:
re: generic - i mean sure, i get that, it's just that it's generating quite a few CI inspections so i was wondering if it can be removed entirely. i guess it can stay for convenience though, the static fields can be moved to the non-generic base as protected.

not sure about the screenspace awareness. would have to poke `CompositeDrawable` myself.

peppy:
I'd remove generics for simplicity. `Container` is an exception since it is such a commonly used class. this is not.

bdach:
@harshavb did you happen to uncheck the "Allow edits from maintainers" checkbox on this PR? i have a batch of changes that are supposed to resolve my reviews, fix up style somewhat and reattribute previous commits, but i cannot push to this PR.

that said it might also be due to this being PRed from `master`, which is never a good idea.

harshavb:
@bdach It's likely because it's PRed from master, since it's not unchecked.

harshavb:
Moved to #4091

PR #4017, opened 2020-11-23
Fix PrepareForUse potentially getting called twice on the same pooled usage

peppy:
This turns out to be quite dangerous and the cause for https://github.com/ppy/osu/issues/10938. It also fixes hitobjects disappearing in the editor at random.

Closes https://github.com/ppy/osu/issues/10938.

### DIFF
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneDrawablePool.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneDrawablePool.cs
@@ -153,6 +153,30 @@ namespace osu.Framework.Tests.Visual.Drawables
         }
 
         [Test]
+        public void TestPrepareOnlyOnceOnMultipleUsages()
+        {
+            resetWithNewPool(() => new TestPool(TimePerAction, 1));
+
+            TestDrawable drawable = null;
+            TestDrawable drawable2 = null;
+
+            AddStep("consume item", () => drawable = consumeDrawable(false));
+
+            AddAssert("prepare was not run", () => drawable.PreparedCount == 0);
+            AddUntilStep("free was not run", () => drawable.FreedCount == 0);
+
+            AddStep("manually return drawable", () => pool.Return(drawable));
+            AddUntilStep("free was run", () => drawable.FreedCount == 1);
+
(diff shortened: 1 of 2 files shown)


osu.Framework.Tests/Visual/Drawables/TestSceneDrawablePool.cs
```diff
@@ -152,6 +152,30 @@ public void TestPrepareAndFreeMethods()
...
+            TestDrawable drawable = null;
+            TestDrawable drawable2 = null;
+
+            AddStep("consume item", () => drawable = consumeDrawable(false));
+
+            AddAssert("prepare was not run", () => drawable.PreparedCount == 0);
+            AddUntilStep("free was not run", () => drawable.FreedCount == 0);
+
+            AddStep("manually return drawable", () => pool.Return(drawable));
+            AddUntilStep("free was run", () => drawable.FreedCount == 1);
```
bdach, line 169: This is about my only hang-up with this - should free be allowed to run if prepare hasn't? I could live with that being the case but I imagine that consumers might be operating on the premise that free can't run on drawables that were never prepared.

I suppose you could say that the state of the poolable before prepare is undefined in which case running free is probably better than not...

peppy: It's actually *required* based on how we use these methods in some locations (take `FollowPointConnection` as an example):

```csharp
Get(() => { d.Entry = entry; } // run immediately in Assign

...

override FreeAfterUse()
{
    Entry = null;
}
```

smoogipoo: How about leaving a remark on the xmldoc for `FreeAfterUse`?

This looks fine to me - I'm also suspicious, but I don't see a way of handling it without reconsidering this class - giving `PrepareForUse` an argument or having a `teardownAction` to complement `setupAction` in `Get()`. Both of those sound equally as bad and have their own downsides that actually probably cause more harm than good.

peppy: Have added xmldoc for now.

smoogipoo (approved):

PR #4031, opened 2020-11-24
Update outdated dependencies

peppy:
Combines all feasible merges from dependabot.


smoogipoo (approved):

peppy:
Have tested on iOS and android.

Issue #3930, opened 2020-10-12
Running TestSceneRearrangeableListContainer.TestRemoveDuringLoadAndReAdd by itself fails


peppy:
![image](https://user-images.githubusercontent.com/191335/95769963-a634da80-0cf3-11eb-8002-78a56f34262d.png)

Running with another test works fine.

PR #4032, opened 2020-11-24
Fix RearrangeableList tests failing via headless run

peppy:
Can reproduce this 100% when running tests from rider (headless). Not sure if this covers the test scenario completely (seems like the drawables are loaded asynchronously), can add some wait and re-assert steps if that's deemed required.

Closes https://github.com/ppy/osu-framework/issues/3930.


bdach (approved):
I think it's a good enough resolution for now. Don't see this as worth finicking with unless it actually fails.

PR #4034, opened 2020-11-25
Make Scheduler.Add() return the scheduled delegate

smoogipoo:
As discussed in Discord, there's been very few times when a `bool` return has been useful here.

Additionally, in some places such as `PoolableDrawable`, it's unsafe to use `AddDelayed` unless you're fully conscious of potential rewindability because `AddDelayed` adds the clock's current time to the execution time. At this point, I'd say none of us are able to do this since such a regression went completely unnoticed in https://github.com/ppy/osu-framework/pull/4017.

Another point to consider is `Drawable.Schedule()` forwards to `Scheduler.AddDelayed` to apply the transform delay (even if 0), which is also "unsafe" as above if you're only expecting something to be executed in the next update. Such cases can be handled in two ways:
```
using (BeginAbsoluteSequence(double.MinValue))
    Schedule(() => { });

---

Scheduler.Add(() => { });
```
With the latter, as a result of this PR, now being cancellable by the delegate now being returned.

### DIFF
--- a/osu.Framework/Threading/Scheduler.cs
+++ b/osu.Framework/Threading/Scheduler.cs
@@ -198,6 +198,7 @@ namespace osu.Framework.Threading
         /// <summary>
         /// Add a task to be scheduled.
         /// </summary>
+        /// <remarks>If scheduled, the task will be run on the next <see cref="Update"/> independent of the current clock time.</remarks>
         /// <param name="task">The work to be done.</param>
         /// <param name="forceScheduled">If set to false, the task will be executed immediately if we are on the main thread.</param>
         /// <returns>The scheduled task, or <c>null</c> if the task was executed immediately.</returns>
@@ -221,6 +222,7 @@ namespace osu.Framework.Threading
         /// <summary>
         /// Add a task to be scheduled.
         /// </summary>
+        /// <remarks>The task will be run on the next <see cref="Update"/> independent of the current clock time.</remarks>
         /// <param name="task">The scheduled delegate to add.</param>
         /// <exception cref="InvalidOperationException">Thrown when attempting to add a scheduled delegate that has been already completed.</exception>
         public void Add(ScheduledDelegate task)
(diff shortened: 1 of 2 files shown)


osu.Framework/Threading/Scheduler.cs
```diff
@@ -200,20 +200,22 @@ public void CancelDelayedTasks()
         /// </summary>
         /// <param name="task">The work to be done.</param>
         /// <param name="forceScheduled">If set to false, the task will be executed immediately if we are on the main thread.</param>
-        /// <returns>Whether we could run without scheduling</returns>
-        public bool Add(Action task, bool forceScheduled = true)
+        /// <returns>The scheduled task, or <c>null</c> if the task was executed immediately.</returns>
+        public ScheduledDelegate Add(Action task, bool forceScheduled = true)
```
peppy, line 204: might be worth putting a `[CanBeNull]` just so we are 101% aware of the nullability.

```suggestion
        [CanBeNull]
        public ScheduledDelegate Add(Action task, bool forceScheduled = true)
```

osu.Framework/Threading/Scheduler.cs
```diff
@@ -200,20 +200,22 @@ public void CancelDelayedTasks()
         /// </summary>
```
peppy, line 200: Maybe add a mention that this doesn't consider clock/time at all in the xmldoc? Not sure if necessary but might give more visibility on the potential issues using AddDelayed

osu.Framework/Threading/Scheduler.cs
```diff
@@ -241,6 +243,7 @@ public void Add(ScheduledDelegate task)
         /// <param name="task">The work to be done.</param>
```
peppy, line 243: Same on this xmldoc? Not sure but maybe worth mentioning the fail case that we hit just for visibility.

peppy (changes requested):

peppy (approved):

PR #4036, opened 2020-11-25
Fix some debuggers not correctly showing exception stacks

smoogipoo:
From:
Screenshot_2020-11-25_18-38-53

To:
Screenshot_2020-11-25_18-39-37

Discovered this mostly by accident, but it seems like having the crashed thread exit before the debugger does its thing causes the stack trace to not display correctly.

Scenarios I've tested:
- Throw in `DrawableHitObject.load()` single/multi-threaded.
- Throw in `DrawableHitObject.Update()` single/multi-threaded.
- Throw in `DrawableHitObject.load()` with `TestSceneOsuPlayer.TestConstructor`.
    - This doesn't launch your debugger but that seems to just be a function of how nunit works since it runs all test code within exception-safe blocks.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -244,7 +244,7 @@ namespace osu.Framework.Platform
         {
             var exception = (Exception)args.ExceptionObject;
             exception.Data["unhandled"] = "unhandled";
-            handleException(exception);
+            handleException(sender, exception);
         }
 
         private void unobservedExceptionHandler(object sender, UnobservedTaskExceptionEventArgs args)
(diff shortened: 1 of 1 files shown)


bdach:
Weird, I've tested the three scenarios described above on lazer `master` without this PR checked out and all cases seemingly gave me stacks even without this change applied...

smoogipoo:
Me too. Perhaps this is only a problem with using a local o!f checkout...

smoogipoo:
I'll close this issue for now then, and revisit if any weird exception cases come up again.

peppy (approved):
Seems harmless, let's add it in and revert later if issues are discovered.

PR #4037, opened 2020-11-26
Close midi devices on disconnect

smoogipoo:
Should fix https://github.com/ppy/osu/issues/10969

I'm not currently able to test this, so would appreciate a confirmation. Should be replicable by adding a `GC.Collect(); GC.WaitForPendingFinalizers()` after the `openedDevices.Remove(key);` line.


peppy:
Works fine (tested on macOS)

peppy (approved):

PR #4039, opened 2020-11-28
Fix binding bindables potentially transfers values while it's disabled

### LINKED ISSUE #3218: Binding to a remote non-disabled bindable while locally disabled will fail
Due to the order of value propagation, a local disabled bindable will attempt to transfer Value before Disabled state, causing an exception to be thrown (see `BindTo` implementation).

Note that this same failure can occur in other bindable implementations like BindableNumber (see `Precision` / `MinValue` / `MaxValue` transfer). We may want to abstract out these properties into a BindableProperty implementation to simplify and centralise the disabled check / bypass logic.

frenzibyte:
Resolves #3218

This fixes the issue by transferring the values directly via `Set...` methods, which bypass the disabled check.

### DIFF
--- a/osu.Framework.Tests/Bindables/BindableBindingTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableBindingTest.cs
@@ -23,6 +23,32 @@ namespace osu.Framework.Tests.Bindables
         }
 
         [Test]
+        public void TestBindWhileDisabled()
+        {
+            Bindable<string> bindable1 = new Bindable<string> { Disabled = true };
+            Bindable<string> bindable2 = new Bindable<string>("different");
+
+            bindable1.BindTo(bindable2);
+
+            Assert.AreEqual(false, bindable1.Disabled);
+            Assert.AreEqual("different", bindable1.Value);
+            Assert.AreEqual(bindable2.Value, bindable1.Value);
+        }
+
+        [Test]
+        public void TestBindToDisabled()
+        {
+            Bindable<string> bindable1 = new Bindable<string>();
+            Bindable<string> bindable2 = new Bindable<string>("different") { Disabled = true };
+
+            bindable1.BindTo(bindable2);
+
+            Assert.AreEqual(true, bindable1.Disabled);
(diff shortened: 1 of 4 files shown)


peppy:
Did those failing tests somehow pass for you locally?

bdach:
The tests are failing because this is actually a breaking change in terms of semantics of `BindTo()`. Before, `BindTo()` would adhere to the general bindable rule of "if binding to the other bindable doesn't change the current value, don't trigger value change events", which is no longer the case when using `Set{Value,DefaultValue,Disabled}`, which propagates unconditionally. As seen by the test results breaks the interactions of `BindTo()` and bindable leasing, among other things.

bdach (changes requested):
see above

peppy:
Yep, that's precisely it. I'm just horribly confused how this made it to PR state without that being realised, since bindables are one of the areas we have pretty flawless test coverage.

frenzibyte:
> The tests are failing because this is actually a breaking change in terms of semantics of `BindTo()`. Before, `BindTo()` would adhere to the general bindable rule of "if binding to the other bindable doesn't change the current value, don't trigger value change events", which is no longer the case when using `Set{Value,DefaultValue,Disabled}`, which propagates unconditionally. As seen by the test results that breaks the interactions of `BindTo()` and bindable leasing, among other things.

Ah, my bad on this one, pushed things blindly.

Should be good now, one thing to note though is that the equality comparer inside `Set...()` methods uses the current value instead of `previousValue` for checking (eg: `this.value` instead of `previousValue`), having 2 things that should logically be the same but are coming from different members doesn't sound good to have.

Not quite sure what to do with it, I still don't get why `previousValue` exists in the set methods, bindables should never be desynchronised unless I overlooked something.

bdach:
Tests disagree, the newest version still apparently breaks leasing. Did you run them all before pushing?

frenzibyte:
Yeah... I haven't, looking into it.

frenzibyte:
Closing this as I think it's better to resolve this with the `BindableProperty` refactor rather than touching on the existing value propagation logic.

PR #4040, opened 2020-11-28
Fix native delegate callback getting garbage collected

peppy:
Didn't notice this as it only happens when startup occurs fast enough to GC to have run (not in my VM or with a debugger attached, but can happen on my gaming PC).

Fixes the following startup crash:
```
Process terminated. A callback was made on a garbage collected delegate of type
'SDL2-CS!SDL2.SDL+SDL_EventFilter::Invoke'.
   at SDL2.SDL.SDL_ShowWindow(IntPtr)
   at SDL2.SDL.SDL_ShowWindow(IntPtr)
   at osu.Framework.Platform.SDL2DesktopWindow+<>c__DisplayClass67_0.<set_Visible>b__0()
```

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -406,13 +406,16 @@ namespace osu.Framework.Platform
             WindowMode.TriggerChange();
         }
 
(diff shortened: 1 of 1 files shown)


bdach:
Could reproduce the described failure when building release and placing a strategic `GC.Collect()`; confirmed that it's resolved with this change :+1:

bdach (approved):

Issue #4044, opened 2020-11-30
Unexpected results with hover and click events with SDL on MacOS Retina displays


ben9583:
Following the switch from OpenTK to SDL, made default in `osu-framework` v2020.1127.1, hover events and click events are not triggered at the correct locations.

A few weeks ago, I tried to run an `osu-framework` project using the --sdl flag to run with SDL instead of OpenTK, but the result was this issue, so I removed the flag. With the new version of `osu-framework`, SDL was made the default, and so was the issue with the hover and click events. As a bypass, I am running with `useOsuTK: True`, which fixes the problem.

Someone else working on my project updated to use SDL, but he does not experience any issues. I am running on a 2019 16-inch Macbook Pro with MacOS Catalina 10.15.7, and he runs a Windows machine.

In the video below is our game ([harshavb/piously](https://github.com/harshavb/piously)) running with OpenTK. When the mouse hovers over the logo, it increases in scale, and a click is responded to properly. All of the menu buttons also work.
[Piously with OpenTK](https://youtu.be/BURlwTSomaI)

This video shows our game with SDL. The logo does not respond when the mouse is hovered over it, but when the mouse is near the bottom right corner of the screen. Furthermore, the bounding zones for the hover and click events are much larger than should be. I believe the size and position for the events was scaled by a factor of 2, which is a common issue with apps on retina Macbook displays.
[Piously with SDL](https://youtu.be/9oFVYjXxXTU)

PR #4045, opened 2020-11-30
Fix DPI scaling not working correctly when running under SDL2

peppy:
Tested on macOS

Closes https://github.com/ppy/osu-framework/issues/4044.

@ben9583 would appreciate confirmation that this fixes the issue for you too.

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -219,9 +219,9 @@ namespace osu.Framework.Platform
         /// <summary>
         /// Returns the drawable area, after scaling.
         /// </summary>
-        public Size ClientSize => new Size((int)(Size.Width * scale), (int)(Size.Height * scale));
+        public Size ClientSize => new Size(Size.Width, Size.Height);
 
-        private const float scale = 1; // TODO: figure if/when we need this.
+        public float Scale = 1;
 
         /// <summary>
         /// Queries the physical displays and their supported resolutions.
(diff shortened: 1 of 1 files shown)


ben9583:
I'm having some issues; the game only takes up the bottom left of the screen and the cursor in game is mapped to only be in the bottom corner.
[video](https://youtu.be/O1ESmXNmaHA)

peppy:
Are you sure you are testing on this PR? I've made no changes which could cause that.

ben9583:
Yes, I cloned from `peppy:fix-dpi-scale`; on master, the window is 1/2 the length and height, but works properly. It can be scaled up to the same size as on `peppy:fix-dpi-scale` and have the game take up the whole screen.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -219,9 +219,9 @@ public WindowState WindowState
         /// <summary>
         /// Returns the drawable area, after scaling.
         /// </summary>
-        public Size ClientSize => new Size((int)(Size.Width * scale), (int)(Size.Height * scale));
+        public Size ClientSize => new Size(Size.Width, Size.Height);
```
bdach, line 222: I suspect removing the multiply by `scale` here is what is causing the issue described above, but I don't have the means to check. It would certainly make sense, though, given that this is what `GLWrapper` reads in `Reset()`.

ben9583:
So I'm not really sure how this happened, but the issue seems to have fixed itself... After rebuilding a couple times on the `peppy:fix-dpi-scale` branch, the resolution is proper and the hover/click events on `osu.Framework.Tests` is working as expected. I'm not really sure what happened.

osu.Framework/Platform/SDL2DesktopWindow.cs
peppy: this was intentionally removed to make things function correctly for me. will check further again but this was a conscious change.

bdach: was a guess, apparently this is fine [judging from the comment below](https://github.com/ppy/osu-framework/pull/4045#issuecomment-735925412)

smoogipoo (approved):

PR #4048, opened 2020-12-01
Fix crash when AudioCollectionManager is modified on disposal

### LINKED ISSUE #3713: Collection modified exception when disposing audio components
See: https://ci.appveyor.com/project/peppy/osu/builds/34056896/tests

```
--TearDown
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at osu.Framework.Audio.AudioCollectionManager`1.Dispose()
   at osu.Framework.Audio.AudioCollectionManager`1.Dispose()
   at osu.Framework.Game.Dispose(Boolean isDisposing)
   at osu.Game.OsuGameBase.Dispose(Boolean isDisposing) in C:\projects\osu\osu.Game\OsuGameBase.cs:line 339
   at osu.Game.OsuGame.Dispose(Boolean isDisposing) in C:\projects\osu\osu.Game\OsuGame.cs:line 525
   at osu.Framework.Graphics.Drawable.dispose(Boolean isDisposing)
   at osu.Framework.Graphics.Drawable.Dispose()
```
Seems like there's nothing in `AudioCollectionManager.Dispose()` preventing the collection being enumerated over from being asynchronously modified by the audio thread. A lock is probably required.

miterosan:
* fixes #3713

### DIFF
--- a/osu.Framework.Tests/Audio/AudioCollectionManagerTest.cs
+++ b/osu.Framework.Tests/Audio/AudioCollectionManagerTest.cs
@@ -13,18 +13,39 @@ namespace osu.Framework.Tests.Audio
         [Test]
         public void TestDisposalWhileItemsAreAddedDoesNotThrowInvalidOperationException()
         {
-            var manager = new AudioCollectionManager<AdjustableAudioComponent>();
+            var manager = new TestAudioCollectionManager();
+
+            var threadExecutionFinished = new ManualResetEventSlim();
+            var updateLoopStarted = new ManualResetEventSlim();
 
             // add a huge amount of items to the queue
-            for (int i = 0; i < 10000; i++) manager.AddItem(new TestingAdjustableAudioComponent());
+            for (int i = 0; i < 10000; i++)
+                manager.AddItem(new TestingAdjustableAudioComponent());
 
-            // in a seperate thread start processing the queue
-            new Thread(() => manager.Update()).Start();
+            // in a separate thread start processing the queue
+            var thread = new Thread(() =>
+            {
+                while (!manager.IsDisposed)
+                {
+                    manager.Update();
+                    updateLoopStarted.Set();
+                }
 
-            // wait a little for beginning of the update to start
-            Thread.Sleep(4);
+                threadExecutionFinished.Set();
+            });
+
+            thread.Start();
+
+            Assert.IsTrue(updateLoopStarted.Wait(1000));
 
             Assert.DoesNotThrow(() => manager.Dispose());
+
+            Assert.IsTrue(threadExecutionFinished.Wait(1000));
+        }
+
+        private class TestAudioCollectionManager : AudioCollectionManager<AdjustableAudioComponent>
+        {
+            public new bool IsDisposed => base.IsDisposed;
         }
 
         private class TestingAdjustableAudioComponent : AdjustableAudioComponent
--- a/osu.Framework.Tests/Audio/AudioCollectionManagerTest.cs
+++ b/osu.Framework.Tests/Audio/AudioCollectionManagerTest.cs
@@ -36,11 +36,11 @@ namespace osu.Framework.Tests.Audio
 
             thread.Start();
 
-            Assert.IsTrue(updateLoopStarted.Wait(1000));
+            Assert.IsTrue(updateLoopStarted.Wait(10000));
 
             Assert.DoesNotThrow(() => manager.Dispose());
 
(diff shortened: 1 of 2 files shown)


bdach:
presumably also closes https://github.com/ppy/osu/issues/11029?

peppy:
For a fix like this, you really need to attach some documentation on what the cause / resolution is. Audio threading is very delicate so it's hard to make a call on this fix with nothing to go by but a test.

miterosan:
Problem:

The exception happens when AudioCollectionManager.Items gets updated and disposed at the same time. 

Access of AudioCollectionManager.Items only happens when the action queue gets processed. (That happens in AudioCollectionManager.Update)

But AudioCollectionManager.Dispose() does not queue the access of AudioCollectionManager.Items. 
This allows that the queue gets processed and modifies AudioCollectionManager.Items while AudioCollectionManager.Dispose is iterating over AudioCollectionManager.Items.

Fix: 

Move the disposal of the items into the queuedDisposal.

Dispose path before this PR:

1. AudioCollectionManager.Dispose()
2. **Iterating over AudioCollectionManager.Items and disposing**
3. AudioComponent.Dispose()
4. AudioComponent.EnqueueAction(Dispose(true))
5. AudioComponent.Update
6. AudioComponent.Dispose(true)
7. AdjustableAudioComponent.Dispose(true)
8. Unbind all bindings from AudioCollectionManager.Items

Dispose path after this PR

1. AudioComponent.Dispose()
2. AudioComponent.EnqueueAction(Dispose(true))
3. AudioComponent.Update
4. AudioComponent.Dispose(true)
5. AudioCollectionManager.Dispose(true)
6. **Iterating over AudioCollectionManager.Items and disposing** 
7. AdjustableAudioComponent.Dispose(true)
8. Unbind all bindings from AudioCollectionManager.Items

bdach:
The above means that this comment is now false, no?

https://github.com/ppy/osu-framework/blob/6df9a71818c8ca7a59b079f575a3081d2969e282/osu.Framework/Audio/AudioCollectionManager.cs#L54

At the very least, it needs to be trimmed now; but given that this comment was placed there for a reason, I'm inclined to believe there might be a reason as to why the operations were ordered this way, which means that there is a possibility that this fix will break something else.

bdach (changes requested):
see above

miterosan:
The reason for the comment:

The comment made it clear to

miterosan:
This is indeed complex. I updated the comment for clarification.

The comment hints that AudioCollectionManager.Items and children should queue their own disposal before the AudioCollectionManager (See AudioComponent.Dispose) queues its own Dispose and then disables its queuing. 

Moving the .Items disposal into the beginning of the "main" disposal only changes the thread where the disposal gets queued.
The processing of the queued disposal of the items does not change. It always happens at the end of  AudioCollectionManager.Update when UpdateChildren() gets called, which in turn updates the children, which process their queued disposal.

miterosan:
> The comment hints that AudioCollectionManager.Items and children should queue their own disposal before the AudioCollectionManager (See AudioComponent.Dispose) queues its own Dispose and then disables its queuing.

Due to the way AudioComponent.Update works, it does not matter if the queuing of items disposal happens in the beginning of the queued disposal or directly when .Dispose gets called.

The comment was introduced in https://github.com/ppy/osu-framework/commit/08f20b1ab73dabbb272bff5930cc3f9da1e8bf8d and then made unnessesary in https://github.com/ppy/osu-framework/commit/85eff66c00e809d0513d54a5b9f21daf1b076c1d

peppy:
Okay, I've checked blame history of the original code and this looks quite safe. Have also tested against osu! and it looks to work as expected (cannot reproduce the test failures we were seeing over many runs).

I've rewritten your test to avoid the arbitrary sleep and ensure the thread is exited before the test finshes. @smoogipoo requesting a quick code check.

peppy (approved):

peppy (dismissed):

miterosan:
@peppy Now the test is always green, even before my change

peppy:
Should be better now. One of the semaphores was fired after the update, which is actually too late to cause the issue (I incorrectly assumed it would still happen on further updates, but that's not the case).

miterosan:
works 👍

The test is now working correctly on my machine.

bdach:
I'm now reasonably convinced that this is OK with the provided explanations :+1:

bdach (approved):

PR #4057, opened 2020-12-02
Fix some exceptions potentially leading to deadlocked process

smoogipoo:
The first commit is just a cleanup of the condition, which has been bothering me since I wrote the code in the first place. When debugging this I found it challenging to immediately compute the result of the binary operations.

The second commit ensures that the handler can't be hit a second time from another `Task` failing. This is always going to be a no-op because the input thread has died.

The third commit makes sure that the async disposal queue signals its completion if it itself dies. It would only block for a maximum of 10 seconds, but it should still quit ASAP otherwise e.g. the test runner could be left running.

I can replicate both above issues with the osu!-side `TestSceneSpectator` (crashes in disposal).

### DIFF
--- a/osu.Framework/Allocation/AsyncDisposalQueue.cs
+++ b/osu.Framework/Allocation/AsyncDisposalQueue.cs
@@ -43,28 +43,33 @@ namespace osu.Framework.Allocation
 
             runTask = Task.Run(() =>
             {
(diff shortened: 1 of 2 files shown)


peppy (approved):
Sounds sane on paper.

PR #1968, opened 2018-11-01
Make Linux specific file names include version numbers

morguldir:
This fixes loading issues for linux flavours like Ubuntu, also "libSDL2-2.0" is not valid on any distros, because libSDL2-2.0 always has version numbers at the end.

---

Fix broken library name in linux clipboard, make osu specific library names work without development packages, which was required on some linux flavours.

PR #4059, opened 2020-12-02
Standardise naming of SDL2 references

### LINKED ISSUE #4000: Fix SDL2 naming for consistency
Once all the window backend stuff is fixed up, the following renaming should be applied (please comment if any issues are seen with this direction):

- [x] Depends on https://github.com/ppy/osu-framework/issues/3990
- [x] `DesktopWindow` -> `SDL2DesktopWindow` (done in #3994)
- [ ] `Sdl` -> `SDL2`
- [ ] `SDL` -> `SDL2`

peppy:
Closes #4000.

### DIFF
--- a/osu.Framework/Platform/Linux/SDL2/SDL2Clipboard.cs
+++ b/osu.Framework/Platform/Linux/SDL2/SDL2Clipboard.cs
@@ -4,9 +4,9 @@
 using System;
 using System.Runtime.InteropServices;
 
-namespace osu.Framework.Platform.Linux.Sdl
+namespace osu.Framework.Platform.Linux.SDL2
 {
-    public class SdlClipboard : Clipboard
+    public class SDL2Clipboard : Clipboard
     {
         private const string lib = "libSDL2-2.0.so.0";
 
--- a/osu.Framework/Platform/Linux/LinuxGameHost.cs
+++ b/osu.Framework/Platform/Linux/LinuxGameHost.cs
@@ -3,7 +3,7 @@
 
 using System;
 using System.IO;
-using osu.Framework.Platform.Linux.Sdl;
+using osu.Framework.Platform.Linux.SDL2;
 using osuTK;
 
 namespace osu.Framework.Platform.Linux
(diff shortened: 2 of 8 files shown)


osu.Framework/Platform/Linux/SDL2/SDL2Clipboard.cs
```diff
@@ -4,9 +4,9 @@
 using System;
 using System.Runtime.InteropServices;
 
-namespace osu.Framework.Platform.Linux.Sdl
+namespace osu.Framework.Platform.Linux.SDL2
 {
-    public class SdlClipboard : Clipboard
+    public class SDL2Clipboard : Clipboard
     {
         private const string lib = "libSDL2-2.0.so.0";
```
smoogipoo, line 11: Just realised, but this is outdated with the recent SDL2-CS change.

smoogipoo (approved):

osu.Framework/Platform/Linux/SDL2/SDL2Clipboard.cs
peppy: good catch, will update it in line with the package bump (once appveyor catches up).

smoogipoo: Also I'll post here for tracking purposes but here too:

https://github.com/ppy/SDL2-CS/blob/master/app.config#L5

Not sure this has any effect on .net core tho.

peppy: hmm.. and here i was thinking that renaming it would be more reliable. maybe that one should be updated too?

are you able to test the one here works fine with a system-wide sdl install (when only specifying `libSDL2.so`)?

smoogipoo: It works for me, but I'm not sure if it'll work on all systems (https://github.com/ppy/osu-framework/pull/1968)

Issue #4060, opened 2020-12-02
Fullscreen doesn't always result in correct rendering


peppy:
This is a regression from SDL2 changes, and likely an issue with handling at our end. Doesn't happen every time, but on some fullscreen resolution changes the render size will be that of the user's window size setting.

![Screen Shot 2020-12-02 at 18 34 45](https://user-images.githubusercontent.com/191335/100855063-1f1f0a00-34cd-11eb-9455-a4250b7857e2.png)

PR #4062, opened 2020-12-02
Fix switching fullscreen resolutions potentially resulting in incorrect render resolution

peppy:
This is really painful to need to do, but is required due to two other hacks/workarounds:

This addition for fullscreen, where we enter windowed mode briefly:
https://github.com/ppy/osu-framework/blob/6272037fdf3eb9d17afe40b3823e47f6340b9c1e/osu.Framework/Platform/SDL2DesktopWindow.cs#L963-L966

In conjunction with the SDL event filtering where we handle resize events (to stop blocking updates when resizing the window):
https://github.com/ppy/osu-framework/blob/6272037fdf3eb9d17afe40b3823e47f6340b9c1e/osu.Framework/Platform/SDL2DesktopWindow.cs#L420-L427

Closes https://github.com/ppy/osu-framework/issues/4060


bdach:
This one I'm also not sure how to repro but I guess the solution makes sense to me.

bdach (approved):

smoogipoo (approved):

Issue #4061, opened 2020-12-02
Entering borderless mode from fullscreen doesn't always result in correct display


peppy:
Having trouble capturing this but did reproduce multiple times so logging an issue (results in the game taking up a portion of the screen anchored bottom-left that doesn't match the desktop resolution, with the rest of the window being black). May be somewhat related to #4060.

PR #4063, opened 2020-12-02
Fix borderless size being based on a potentially outdated value

peppy:
Closes https://github.com/ppy/osu-framework/issues/4061

Reading from `CurrentDisplayMode` is awkward as it is only valid when in fullscreen (and only initialised below the usage here).


bdach:
I *think* this is ok, although I can't say that I can reproduce the linked issue...

bdach (approved):

smoogipoo (approved):

PR #4064, opened 2020-12-02
Update SDL2-CS with bundled linux binaries

peppy:


bdach:
Confirmed that the shared objects are getting copied to the output directory (both in the case of framework visual tests and the game), and that they are subsequently looked up from the runtimes dir using `LD_DEBUG` :+1:

bdach (approved):

PR #4065, opened 2020-12-03
Remove disposed pool statistics from global statistics

peppy:
Open to suggestion on how to improve this visually. We could potentially add an identifier string to better describe multiple instances of the same pool (allow them to override or provide a string based on their specifics).

image

### DIFF
--- a/osu.Framework/Graphics/Pooling/DrawablePool.cs
+++ b/osu.Framework/Graphics/Pooling/DrawablePool.cs
@@ -23,13 +23,12 @@ namespace osu.Framework.Graphics.Pooling
     /// <typeparam name="T">The type of drawable to be pooled.</typeparam>
     public class DrawablePool<T> : CompositeDrawable, IDrawablePool where T : PoolableDrawable, new()
     {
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Pooling/DrawablePool.cs
```diff
@@ -45,7 +44,12 @@ public DrawablePool(int initialSize, int? maximumSize = null)
             this.maximumSize = maximumSize;
             this.initialSize = initialSize;
 
-            statistic = usage_statistic;
+            int id = Interlocked.Increment(ref poolInstanceID);
+
+            usageStatistic = GlobalStatistics.Get<DrawablePoolUsageStatistic>(nameof(DrawablePool<T>), typeof(T).ReadableName() + $"`{id}");
+            usageStatistic.Value = new DrawablePoolUsageStatistic();
+
+            statistic = usageStatistic;
```
bdach, line 52: could probably have just one field instead of both `usageStatistic` and `statistic` at this point? not sure why there ever were two, really, to be honest

bdach (approved):

PR #4069, opened 2020-12-04
Handle value rejection on binding to another bindable

frenzibyte:
This fixes an edge-case where a bindable would reject the given value of another bindable on `BindTo()`, causing them to be desynchronised. 

This fixes it by changing the other bindable's value to the updated value of the current bindable, since they're bound to each other and therefore must be synchronised.


frenzibyte:
The above test failure should be irrelevant, it passes locally on my end:
image

osu.Framework/Bindables/Bindable.cs
```diff
@@ -196,6 +196,12 @@ public virtual void BindTo(Bindable<T> them)
 
             addWeakReference(them.weakReference);
             them.addWeakReference(weakReference);
+
+            if (!EqualityComparer<T>.Default.Equals(Value, them.Value))
+            {
+                // The value of the other bindable has been rejected from being set here, propagate the updated value to them instead, to keep synchronisation.
```
bdach, line 202: This reads awkwardly to me, would reword to something like:

```
// Our value is different to their value, presumably due to our value change callbacks changing it.
// To ensure consistency, propagate our value to them in the backwards direction.
```

This is generally quite dicey, as is the entire "feature" of "mutate own value on binding". With this, if you try binding two bindables that mutate their own values, you'll get a stack overflow exception. That can be "alleviated" by doing this special branch before the bidirectional weak references are established, but that in turn causes a value desynchronization (and will crash anyway on any new change to either bindable).

peppy: Agreed that the current comment reads *back to front*.

bdach:
On second thought, this fix (if it's deemed the proper direction, which I'm not yet 100% convinced of - maybe this edge case should just raise an exception?) should extend to every bindable property which raises change callbacks, which would make at least: `Default`, `Disabled`, `MinValue`, `MaxValue`, `Precision` and `CollectionChanged`.

frenzibyte:
Every event behaves regularly like that, and there are test cases for each `TestValueChangedWithUpstreamRejection`, `TestDisabledChangedWithUpstreamRejection`, etc..

osu.Framework/Bindables/Bindable.cs
frenzibyte: I guess we're fine with:
```
// Our value is different to their value, presumably due to our value change callbacks changing it.
// To ensure consistency, propagate our value to them in the backwards direction.
```
from @bdach's suggestion, yeah?

bdach:
Then why does this PR only handle `Value` by adding that special conditional you have in the diff there? Surely other aforementioned properties should also receive the same treatment?

frenzibyte:
Oh nevermind, I misread your comment, it would be pretty much easy to do with the PR'd `BindableProperty` class, but very messy when done as-is.

I guess this PR could wait for that class first before fixing.

bdach:
I would probably rather have the "messy" but simple checks as done here than that `BindableProperty` stuff (that is hinging on exposing the bindings of a bindable as public, which I maintain should not be done).

osu.Framework/Bindables/Bindable.cs
frenzibyte: I've pushed it, seems fine, but I modified it a little to not look weird in code: 
https://github.com/ppy/osu-framework/blob/a845e7e8d50c0956edaf6c55a1f412a897ef123f/osu.Framework/Bindables/Bindable.cs#L200-L201

frenzibyte:
Gonna close this one for now, there's nothing waiting on this PR so will return to it later once `BindableProperty` becomes a thing (one that doesn't require `Bindings` to be exposed, not sure why I even thought of doing that).

PR #4073, opened 2020-12-05
Fix mouse polling on maximised SDL window calculating incorrectly

frenzibyte:
Previously, only normal (un-maximised) window state would use the window position to determine the global cursor position in relation to the game's position when being outside of the window, which results in incorrect mouse positions when in maximised window state (due to it *incorrectly* using the location of the entire monitor instead) as can be seen below.

Before:

After:

---

One side effect from this is that now the position and size of a maximized window would be saved to config as well, which may not be a good thing to have, although it still saves in some cases on master, and I think saving whether the window was last in maximised state or not and create the window based on it would be a better choice IMO (in a separate PR).

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -880,7 +880,7 @@ namespace osu.Framework.Platform
                 case SDL.SDL_WindowEventID.SDL_WINDOWEVENT_MOVED:
                     var newPosition = new Point(evtWindow.data1, evtWindow.data2);
 
-                    if (windowState == WindowState.Normal && !newPosition.Equals(Position))
+                    if (WindowMode.Value == Configuration.WindowMode.Windowed && !newPosition.Equals(Position))
                     {
                         position = newPosition;
                         updateWindowPositionConfigFromCurrent();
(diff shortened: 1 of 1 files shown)


bdach:
Makes sense to me on a common-sense basis.

@swoolcock do you recall if these checks were against `WindowState.Normal` for a reason? On checking history, they were in your original code if I'm not mistaken.

swoolcock:
> Makes sense to me on a common-sense basis.
> 
> @swoolcock do you recall if these checks were against `WindowState.Normal` for a reason? On checking history, they were in your original code if I'm not mistaken.

The backend classes previously contained no such bindables.

bdach:
There were no bindables, but I was asking about i.e. this check in particular:

https://github.com/ppy/osu-framework/blob/3e9518ac534dc9ad423d0a80361c0771a08c0992/osu.Framework/Platform/SDL2DesktopWindow.cs#L567

which originated in 8d22711945c07e35b2e8ed2ef274fa5b44cc08b3, before the recent refactoring effort. As far as I can tell it should always been checking for maximised (minimised is doubtful, but I guess there's not much harm in including that too for simplicity).

swoolcock:
> There were no bindables, but I was asking about i.e. this check in particular:
> 
> https://github.com/ppy/osu-framework/blob/3e9518ac534dc9ad423d0a80361c0771a08c0992/osu.Framework/Platform/SDL2DesktopWindow.cs#L567
> 
> which originated in [8d22711](https://github.com/ppy/osu-framework/commit/8d22711945c07e35b2e8ed2ef274fa5b44cc08b3), before the recent refactoring effort. As far as I can tell it should always been checking for maximised (minimised is doubtful, but I guess there's not much harm in including that too for simplicity).

If the window is normal/windowed, the position should be taken relative to the window's current windowed position.  If it's fullscreen it should be taken relative to the the display's top left.  This is because `Position` previously maintained its windowed value when changing to fullscreen.

peppy (approved):
Looks good.

peppy:
Still seems to happen for me.

peppy:
Scratch that, I think my master was out-of-date

PR #4075, opened 2020-12-05
Support NumLock in SDL key mapping function

### LINKED ISSUE #2728: Numlock is not considered for keyboard input
**Describe the bug:** 
I use a seperate numpad for clicking as controls. However, it seems the numlock is completely ingored. As a result, I can only use the numbers as they are as input and I can't use the arrow keys on it for selecting maps. I usually have clicking set to "End" and "Arrow down", but in Lazer it can only be "1" and "2".

**Screenshots or videos showing encountered issue:** 
https://imgur.com/trLhs7H

**osu!lazer version:** 
2019.730.0

**Logs:** 
Runtime log (not really useful, but the only one that seemed somewhat logical):
https://pastebin.com/kgju6rGF

bdach:
Closes #2728 (in the SDL case).

- [x] Needs testing on Apple devices (macOS + iOS with external keyboard)

In order to properly support the behaviour of <kbd>NumLock</kbd> that is in line with user expectations, check the state of <kbd>NumLock</kbd> when unmapping SDL's scancodes and return the appropriate alternative keys for keypad numbers and the comma key when <kbd>NumLock</kbd> is off.

I can imagine this could be wrong on Apple platforms as they don't have the notion of a <kbd>NumLock</kbd> key, hence I'd be grateful if this could be checked. I don't really want to pepper in platform checks unless I absolutely have to.

Briefly considered not using `HasFlag` due to its memory concerns but I don't necessarily think it's the hottest path out there and there are way more worrying places that use `HasFlag` already performance-wise, so for now I opted for readability. Can be cleaned up at a later stage if deemed worthwhile, I think.

Tested on Ubuntu 18.04 & Windows 10.

### DIFF
--- a/osu.Framework/Platform/SDL2/SDL2Extensions.cs
+++ b/osu.Framework/Platform/SDL2/SDL2Extensions.cs
@@ -11,7 +11,9 @@ namespace osu.Framework.Platform.SDL2
     {
         public static Key ToKey(this SDL.SDL_Keysym sdlKeysym)
         {
-            bool numLockOn = sdlKeysym.mod.HasFlag(SDL.SDL_Keymod.KMOD_NUM);
+            // Apple devices don't have the notion of NumLock (they have a Clear key instead).
+            // treat them as if they always have NumLock on (the numpad always performs its primary actions).
+            bool numLockOn = sdlKeysym.mod.HasFlag(SDL.SDL_Keymod.KMOD_NUM) || RuntimeInfo.IsApple;
 
             switch (sdlKeysym.scancode)
             {
--- a/osu.Framework/Platform/SDL2/SDL2Extensions.cs
+++ b/osu.Framework/Platform/SDL2/SDL2Extensions.cs
@@ -11,6 +11,8 @@ namespace osu.Framework.Platform.SDL2
     {
         public static Key ToKey(this SDL.SDL_Keysym sdlKeysym)
         {
+            bool numLockOn = sdlKeysym.mod.HasFlag(SDL.SDL_Keymod.KMOD_NUM);
+
             switch (sdlKeysym.scancode)
             {
                 default:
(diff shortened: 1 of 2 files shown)


frenzibyte:
in macOS (not sure about iOS), yeah; this is bringing the numlock functionality even though the OS doesn't support it, SDL mistakes the numlock modifier key for the *clear* key in [apple's (+ numeric pads) keyboards.](https://en.wikipedia.org/wiki/Apple_Keyboard)

I would expect SDL to not count switching numlock modifier on and off on macOS, but unfortunately, they do. I guess fixing it locally here would be easier and faster.

bdach:
I was aware that SDL2 clumps up <kbd>NumLock</kbd> and <kbd>Clear</kbd> into one scancode already, but the fact that pressing it toggles the `KMOD_NUM` flag is... weird. I'm not sure what the <kbd>Clear</kbd> key even does at this point to be fair, and googling isn't helping. Some sources even suggest that <kbd>Clear</kbd> *is* apple's <kbd>NumLock</kbd>.

frenzibyte:
The <kbd>clear</kbd> key in macOS is just a regular key used in calculators to clear what's on screen, then what's on memory (the C/AC buttons):

(*apologies for the watermark*, the last two pressed keys are the <kbd>clear</kbd> keys)

> Some sources even suggest that Clear is apple's NumLock.

I believe these are about how you switch numlock on in windows OS with an apple keyboard (when using bootcamp, to run windows in mac systems)

peppy:
This looks to work correctly on macOS. Not really going to test on iOS as that is quite an outlier case already, but in theory it should work there.

The only weird thing I've noticed is that apple keyboard with numpads have a dedicated `=` key, which isn't in the osuTK `Keys` enum. Unfortunately I can't find anything sensible to map it to, so I think we should just leave it for the time being.

peppy (approved):

PR #4076, opened 2020-12-05
Fix window getting moved out-of-place while in maximised state

### LINKED ISSUE #4070: The taskbar limit is ignored if the app is maximized before (and after) switching between Windowed and Fullscreen/Borderless screen mode
**Describe your problem:** If the app is in windowed mode (but maximized), and you switch from windowed to fullscreen/borderless and switch back to windowed, for some reason the app forgets about the taskbar "limit" (on my Windows 10 setup) and it maximizes until it reaches the bottom edge of the screen.

**Screenshots or videos showing encountered issue:** [Video example with osu!lazer](https://streamable.com/u4q6js)

frenzibyte:
Closes #4070

The issue is coming from the window implementation setting `WindowState` to `Normal` regardless of whether it is maximised or not, which causes the window to potentially have its position changed according to the config while being in maximised state.

This also adds `SDL_RestoreWindow(handle)` to normal state for correct window state updating, as, previously, switching from maximised/minimised state to normal state via `WindowState` property would do absolutely nothing.

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -219,6 +219,12 @@ namespace osu.Framework.Platform
         }
 
         /// <summary>
+        /// Stores whether the window used to be in maximised state or not.
+        /// Used to properly decide what window state to pick when switching to windowed mode (see <see cref="WindowMode"/> change event)
+        /// </summary>
+        private bool windowMaximised;
+
+        /// <summary>
         /// Returns the drawable area, after scaling.
         /// </summary>
         public Size ClientSize => new Size(Size.Width, Size.Height);
@@ -921,18 +927,6 @@ namespace osu.Framework.Platform
         }
 
         /// <summary>
-        /// Stores whether the window used to be in maximised state or not.
-        /// Used to properly decide what window state to pick when switching to windowed mode (see <see cref="WindowMode"/> change event)
-        /// </summary>
-        private bool windowMaximised;
-
-        private void updateMaximisedState()
-        {
-            if (windowState == WindowState.Normal || windowState == WindowState.Maximised)
-                windowMaximised = windowState == WindowState.Maximised;
-        }
-
-        /// <summary>
         /// Should be run on a regular basis to check for external window state changes.
         /// </summary>
         private void updateWindowSpecifics()
@@ -950,9 +944,9 @@ namespace osu.Framework.Platform
             {
                 windowState = currentState;
                 ScheduleEvent(() => OnWindowStateChanged(currentState));
-            }
 
-            updateMaximisedState();
+                updateMaximisedState();
+            }
 
             int newDisplayIndex = SDL.SDL_GetWindowDisplayIndex(SDLWindowHandle);
 
@@ -1018,6 +1012,12 @@ namespace osu.Framework.Platform
                 currentDisplayMode = new DisplayMode(mode.format.ToString(), new Size(mode.w, mode.h), 32, mode.refresh_rate, displayIndex, displayIndex);
 
             isChangingWindowState = false;
+        }
+
+        private void updateMaximisedState()
+        {
+            if (windowState == WindowState.Normal || windowState == WindowState.Maximised)
+                windowMaximised = windowState == WindowState.Maximised;
         }
 
         protected void OnHidden() { }
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -926,7 +926,7 @@ namespace osu.Framework.Platform
         /// </summary>
         private bool windowMaximised;
 
-        private void updateMaximisedState(WindowState windowState)
+        private void updateMaximisedState()
         {
             if (windowState == WindowState.Normal || windowState == WindowState.Maximised)
                 windowMaximised = windowState == WindowState.Maximised;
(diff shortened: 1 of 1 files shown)


osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -1095,7 +1098,16 @@ public void SetupWindow(FrameworkConfigManager config)
                         break;
 
                     case Configuration.WindowMode.Windowed:
-                        WindowState = WindowState.Normal;
+                        // todo: this will cause WindowState value to stay at Fullscreen/Borderless for a whole frame
+                        // but is required for taking correct actions based on whether the window is maximised or not.
+                        ScheduleCommand(() =>
```
peppy, line 1103: How did you find this method? I'd like to think that someone else has run into this problem in the past and there is some documentation we can link to? Don't really want to invent our own solutions where possible..

frenzibyte: More like it was getting incorrectly handled previously, by setting `WindowState to `Normal`, and attempting to change its position according to the config, while it is actually in maximized mode.

bdach:
I think the "not change position if the window is in maximised state" variant sounds like a saner thing to do seeing whatever that fullscreen workaround is doing, but I'm not sure I'm understanding what the caveat about setting `WindowState` with that variant would be - would you mind to elaborate what the problem is there?

frenzibyte:
`WindowState` already has a `Maximised` enum state, and would expect `Normal` to be that the window should be in unmaximised state, in which the window can be moved around the display with no issues.

Choosing the other variant would mean breaking one case of `WindowState` switching via property setter, as I've already mentioned in the description:
> but would break switching from maximised state to normal state via `WindowState` property (as, with this PR, switching to normal state would be restoring the window from maximised/minimised state).

bdach:
Quoting the OP again does not answer my question. I was asking since I don't understand the reasoning provided there to begin with.

frenzibyte:
Ah, my bad, I got driven by my words without looking back at them. using the other variant would cause switching into `Windowed` mode always makes the window unmaximised regardless of what state it was in, as I've added `SDL_RestoreWindow(handle)` there for fixing switching from `Maximised`/`Minimised` into `Normal` not causing any effects.

bdach:
Just as a loose idea, maybe storing info about whether or not the window was maximised previously could be of help? Hesitant to suggest that as the windowing code is close to reaching critical density again, but maybe there is a way to store and use that info in a legible manner...

frenzibyte:
Hmm, I've tried doing that by window event IDs, and failed to work for me (couldn't tell if an `SDL_WINDOWEVENT_RESTORED` means from maximised state or from minimised state), but I think I can do this by checking the flags every window event (basically inside the `updateWindowSpecifics` method), it may not look that good but would 100% be better for getting rid of the "unsynced window state with mode" issue that is present in the pushed solution. Good suggestion, I'll look into doing it tomorrow.

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: @peppy I've updated the PR by using an alternative solution suggested by @bdach, please check if it looks more saner now.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -934,6 +940,9 @@ private void updateWindowSpecifics()
 
             var currentState = ((SDL.SDL_WindowFlags)SDL.SDL_GetWindowFlags(SDLWindowHandle)).ToWindowState();
 
+            if (currentState == WindowState.Normal || currentState == WindowState.Maximised)
+                windowMaximised = currentState == WindowState.Maximised;
```
peppy, line 944: can this just be made into a method (`updateMaximisedState`) and used in both locations? seems like this is basically doing the same thing as your code in the `switch` statement below, but this version is more concise and doesn't need to be split over different `case`s.

hopefully it could then just be called after the full `switch` execution.

peppy (changes requested):

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -920,6 +920,18 @@ private void handleWindowEvent(SDL.SDL_WindowEvent evtWindow)
             }
         }
 
+        /// <summary>
+        /// Stores whether the window used to be in maximised state or not.
+        /// Used to properly decide what window state to pick when switching to windowed mode (see <see cref="WindowMode"/> change event)
+        /// </summary>
+        private bool windowMaximised;
+
+        private void updateMaximisedState(WindowState windowState)
```
peppy, line 929: does this really need a parameter passed in..? it's a field in this class already.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -934,6 +946,8 @@ private void updateWindowSpecifics()
 
             var currentState = ((SDL.SDL_WindowFlags)SDL.SDL_GetWindowFlags(SDLWindowHandle)).ToWindowState();
 
+            updateMaximisedState(currentState);
```
peppy, line 949: just move this below the switch and remove the parameter right?

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: I've done it that way because `OnWindowStateChanged(...)` could potentially change window mode while maximised state not updated. but actually, nevermind, the event is scheduled anyway.

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: I'm not sure what you mean, but this should be placed both here (for catching external window state changes), and in `updateWindowStateAndSize()` (for catching internal window state changes, i.e. via changing `WindowState` property)

peppy (approved):

PR #4077, opened 2020-12-06
Improve support for disabled dropdowns

bdach:
Follow-up on https://github.com/ppy/osu/pull/11101 and the issues with disabling dropdowns I found there.

In general, the existing support for disabled dropdowns was lacking. They were fully interactable, but trying to click on dropdown items just did nothing. This pull aims to achieve the following effects:

- Disabled dropdowns are no longer interactable. Clicking on them does not expand the dropdown, and hovering on them does not change their background colour.
- Disabled dropdowns now have a visual indication of their state, by being partially faded out. This is done by adding a `DisabledColour` to `DropdownHeader`.
- Disabled dropdowns no longer handle changing value via keyboard. They would previously attempt to do so, and promptly crash due to `Current` being disabled.
- If an actively opened dropdown becomes disabled, it is closed immediately.

This includes a test case for the new behaviours, and some slight refactorings of the test scene (first three commits) as I didn't particularly like the way it was written (and needed to re-use some helpers that were already there).

### DIFF
--- a/osu.Framework/Graphics/UserInterface/DropdownHeader.cs
+++ b/osu.Framework/Graphics/UserInterface/DropdownHeader.cs
@@ -3,6 +3,7 @@
 
 using osuTK.Graphics;
 using System;
+using osu.Framework.Bindables;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Input;
@@ -16,6 +17,8 @@ namespace osu.Framework.Graphics.UserInterface
     {
         public event Action<DropdownSelectionAction> ChangeSelection;
 
+        public Bindable<bool> Disabled { get; } = new BindableBool();
+
         protected Container Background;
         protected Container Foreground;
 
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/UserInterface/DropdownHeader.cs
```diff
@@ -16,6 +17,8 @@ public abstract class DropdownHeader : ClickableContainer, IKeyBindingHandler<Pl
     {
         public event Action<DropdownSelectionAction> ChangeSelection;
 
+        public Bindable<bool> Disabled { get; } = new BindableBool();
```
frenzibyte, line 20: `DropdownHeader` already exposes an `Enabled` property (by inheriting `ClickableContainer`), which would make this component expose publicly both an `Enabled` property and `Disabled` property.

frenzibyte (changes requested):
Otherwise, this looks to be and feels very good 👍🏼

osu.Framework/Graphics/UserInterface/DropdownHeader.cs
bdach: good point

frenzibyte (approved):

peppy (approved):
Looks solid

Issue #4068, opened 2020-12-04
Borderless fullscreen runs exclusive on some graphics cards


Joehuu:
Should match stable behavior.
![byOakAurlh](https://user-images.githubusercontent.com/35318437/101105481-864ac600-3582-11eb-8644-2f70d1fb32e8.gif)

peppy:
Can you confirm that borderless is not running exclusive full screen for you under SDL? I believe that may be the cause.

Joehuu:
Seems like it is running exclusive fullscreen.

![chrome_BX6GIpenYa](https://user-images.githubusercontent.com/35318437/101112105-72f32700-3591-11eb-9407-b944783464d5.jpg)

peppy:
Aha, that's going to be the cause of what you're seeing then. Will repurpose this issue thread tracking that (SDL is missing the one-pixel-hackery required to force borderless on some graphics card drivers).

Issue #4074, opened 2020-12-05
Window preview is cut to Windowed mode resolution in Borderless/Fullscreen mode


Sinamer:
**Describe the bug:** Works correctly with --tk flag
Window preview always cuts to the size of the window in Windowed mode, when it should show the full window in Fullscreen or Borderless modes

**Screenshots or videos showing encountered issue:** https://streamable.com/171yp4

**osu!lazer version:** 2020.1204.0

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/5647420/runtime.log)

peppy:
Looks like we are switching to windowed mode on focus loss when we shouldn't be.

peppy:
Okay, so there's a bit more to this.

1. Borderless isn't working as it is on osuTK due to the missing one-pixel hack. This means that borderless *would* should correctly in the preview but it's actually just an alias of fullscreen right now.
2. Fullscreen is behaving more correctly in SDL than osuTK. In osuTK, alt-tabbing from osu! when it's running fullscreen will not restore the desktop to the normal desktop resolution, leaving osu! to run in a windowed (borderless) state at its own specified resolution.

If you check other games that run fullscreen, you'll probably find that they don't get a preview at all. The only reason we're getting one here is due to SDL seemingly switching back to the previous window size briefly on alt-tab. I've confirmed that this isn't something being done at our end. If we really cared enough about the preview being correctly sized, we would likely need to adjust the window size to match the proposed fullscreen size before switching to fullscreen.

Maybe something we do want to consider, but it's extra state to manage at our end.

Issue #3456, opened 2020-04-17
Borderless Mode extends 1 extra pixel right and 1 down.


CheeseManFuu:
**Describe the bug:** 

On my triple 1920x1080 setup, I noticed Borderless Window mode bled onto my other monitor by one pixel. I'm not sure if this is exclusive to 1080p or multiple monitors. It happens regardless of screen scaling. My taskbar is on the bottom of the rightmost monitor, and hides.

While writing this, after realizing I could use in-game screenshots to test if it also happened when there "wasn't" an extra pixel of workspace to go to (i.e. the rightmost monitor), I also found out that it was not only 1921 pixels wide, but also 1081 pixels tall.

**Screenshots or videos showing encountered issue:** 

![borderlosu](https://user-images.githubusercontent.com/7471405/79525580-d4797c00-8028-11ea-961a-8792a4dbcb82.png)
(hard to tell at a glance on a stitched image, but a rectangle select of osu! will show a 1921x1080 area)

![Annotation 2020-04-16 214009](https://user-images.githubusercontent.com/7471405/79526438-e78d4b80-802a-11ea-964f-d98bbff503b7.png)
(unedited screenshot and its properties)

**osu!lazer version:** 
Appeared in 2020.325.0 (I hadn't touched Lazer since, but also hadn't noticed at that point), still appears in 2020.413.0

Using Windows 10 Professional v1903.

peppy:
Bumping into milestone. Worth investigating if this is still required or not (would be nice to be able to remove it if it was a stable-specific thing).

CheeseManFuu:
Gladly; I'll check again tomorrow for Lazer, Stable, and Cutting Edge first chance I get.

CheeseManFuu:
Yeah, some 10 minutes. I was stopped for a while by my incompetence reinstalling Lazer.

Anyway, what I've found: Still happens in Lazer, and only in borderless. Stable/CE unaffected in both, as expected.

[...]

PR #4080, opened 2020-12-07
Fix borderless fullscreen unexpectedly running in exclusive mode on windows

peppy:
This employs the age-old hack that we've been relying on in stable/osuTK of misaligning the window with the screen. On a quick check, many other games use this same method, so it seems like the best we're going to do for the time being.

I've left the existing borderless operation mode for non-windows platforms, since it works better.

Closes https://github.com/ppy/osu-framework/issues/4068
Addresses part of https://github.com/ppy/osu-framework/issues/4074#issuecomment-739690107
Brings back issue mentioned in https://github.com/ppy/osu-framework/issues/3456 (i believe we may be able to solve this, but this is a separate task which requires further investigation into how other games/windowing systems get it working).

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -962,6 +962,7 @@ namespace osu.Framework.Platform
                 case WindowState.Normal:
                     Size = sizeWindowed.Value;
 
+                    SDL.SDL_SetWindowBordered(SDLWindowHandle, SDL.SDL_bool.SDL_TRUE);
                     SDL.SDL_SetWindowFullscreen(SDLWindowHandle, (uint)SDL.SDL_bool.SDL_FALSE);
                     SDL.SDL_SetWindowSize(SDLWindowHandle, Size.Width, Size.Height);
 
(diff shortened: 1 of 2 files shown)


osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -962,6 +962,7 @@ private void updateWindowStateAndSize()
                 case WindowState.Normal:
                     Size = sizeWindowed.Value;
 
+                    SDL.SDL_SetWindowBordered(SDLWindowHandle, SDL.SDL_bool.SDL_TRUE);
```
peppy, line 965: I realise that this should probably be in the windows-only implementation, but it seems to be a harmless noop so left it here for simplicity. Can abstract out if there's an issue with this.

Joehuu:
This seems to update `WindowedSize` to my screen resolution + 1x1 when switching between modes 99% of the time, making windowed like borderless and impossible to resize.

peppy:
Yeah, I've seen this happen more than once, but I can also seem to repro that on master at random. Maybe easier with this change due to there no longer being a huge lag period when changing to borderless?

Was planning on addressing it as a separate issue, but maybe it's best to fix here..

smoogipoo (approved):

frenzibyte:
This is giving me borderless unresizable windows once switching back to windowed:

peppy:
Sure it's not something to do with maximised state or otherwise? can you set a breakpoint and see why the border restoration call isn't being hit?

PR #4085, opened 2020-12-10
Fix maximised window position and size being saved to config

frenzibyte:
Fixes maximised window's size silently getting saved internally to framework config.

This is occurring due to us updating window sizes on an `SDL_WINDOWEVENT_RESIZED`, as that event seems to be sent before SDL finishes processing the external change (putting the window in maximised mode, in this case), unlike `SDL_WINDOWEVENT_SIZE_CHANGED`. Here are the window events I got, and what the value of `WindowState` was for each:
image

as you can see `SDL_WINDOWEVENT_RESIZED` is sent while the window state is still at `Normal`, while `SDL_WINDOWEVENT_SIZE_CHANGED` is sent while the window state has changed to `Maximised`.

Also, `SDL_WINDOWEVENT_RESIZED` would always have an `SDL_WINDOWEVENT_SIZE_CHANGED` event sent afterwards, [as stated in the docs](https://wiki.libsdl.org/SDL_WindowEventID), so technically nothing should ever be regressing here.

---

Also fixes maximised window's position getting saved as well, due to no checks preventing from that.

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -998,6 +998,9 @@ namespace osu.Framework.Platform
                 case WindowState.Maximised:
                     SDL.SDL_SetWindowFullscreen(SDLWindowHandle, (uint)SDL.SDL_bool.SDL_FALSE);
                     SDL.SDL_MaximizeWindow(SDLWindowHandle);
+
+                    SDL.SDL_GL_GetDrawableSize(SDLWindowHandle, out int w, out int h);
+                    Size = new Size(w, h);
                     break;
 
                 case WindowState.Minimised:
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -895,7 +895,6 @@ namespace osu.Framework.Platform
 
                     break;
 
-                case SDL.SDL_WindowEventID.SDL_WINDOWEVENT_RESIZED:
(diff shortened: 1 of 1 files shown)


bdach:
> Also, `SDL_WINDOWEVENT_RESIZED` would always have an `SDL_WINDOWEVENT_SIZE_CHANGED` event sent afterwards, [as stated in the docs](https://wiki.libsdl.org/SDL_WindowEventID), so technically nothing should ever be regressing here.

I don't think that's actually what the docs say. To quote directly:

> `SDL_WINDOWEVENT_RESIZED`
>		window has been resized to data1xdata2; **this event is always preceded by `SDL_WINDOWEVENT_SIZE_CHANGED`**
>
> `SDL_WINDOWEVENT_SIZE_CHANGED`
>		window size has changed, either as a result of an API call or through the system or user changing the window size; **this event is followed by `SDL_WINDOWEVENT_RESIZED` if the size was changed by an external event, i.e. the user or the window manager**

That's actually the other way around logically compared to what you stated. Restating the above in terms of your original order of formulation, `SDL_WINDOWEVENT_SIZE_CHANGED` *might*, but not necessarily always *is* succeeded by `SDL_WINDOWEVENT_RESIZED`.

That said it *does* seem to work on linux if I go to `TestSceneFullscreen` and change the window size to 640x640 via the test step provided, but have you checked whether programmatic changes of the window size work correctly at all, and if so, on which platforms?

peppy:
Seems to work as expected on macOS and windows. I think it's safe to go with this for now.

peppy (approved):

PR #4089, opened 2020-12-10
Save maximised window state for use in next startup

frenzibyte:
- [x] Depends on #4085 (fixing bugs encounterable here)

Saves maximised window state internally to framework config for use in next startups, this PR also has several changes required for "start-up in maximised state" to work correctly:
#### https://github.com/ppy/osu-framework/commit/5c16fa6464f8081358e373a931294e9ec64de64f
To make restoring the window from initial maximised state work properly, as now the window would switch to maximised state while its already positioned and sized to config values, unlike previously where it would switch to maximised state while its position is 0 and size is `default_width,default_height` (1366x768).

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -408,7 +408,6 @@ namespace osu.Framework.Platform
             graphicsBackend.Initialise(this);
 
             updateWindowSpecifics();
-            WindowMode.TriggerChange();
         }
 
         // reference must be kept to avoid GC, see https://stackoverflow.com/a/6193914
@@ -499,13 +498,18 @@ namespace osu.Framework.Platform
 
         public void SwapBuffers()
         {
-            graphicsBackend.SwapBuffers();
-
             if (firstDraw)
             {
+                // SDL actions like SDL_MaximizeWindow() **may** display the window regardless of whether it was hidden.
+                // This is the best place to perform such actions, as the window would have something drawn to it right on next frame.
+                WindowMode.TriggerChange();
+
                 Visible = true;
                 firstDraw = false;
+                return;
             }
+
+            graphicsBackend.SwapBuffers();
         }
 
         /// <summary>
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -501,7 +501,9 @@ namespace osu.Framework.Platform
             if (firstDraw)
             {
                 // SDL actions like SDL_MaximizeWindow() **may** display the window regardless of whether it was hidden.
-                // This is the best place to perform such actions, as the window would have something drawn to it right on next frame.
+                // This is the best place to perform such actions, as the window would have something drawn to it.
+                graphicsBackend.ResetBuffer(IWindow.MaxSize);
+                graphicsBackend.SwapBuffers();
                 WindowMode.TriggerChange();
 
                 Visible = true;
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -330,11 +330,8 @@ namespace osu.Framework.Platform
             }
         }
 
-        private void updateWindowPositionFromConfig()
+        private Point getWindowPositionFromConfig()
         {
-            if (WindowState != WindowState.Normal)
-                return;
-
             var configPosition = new Vector2((float)windowPositionX.Value, (float)windowPositionY.Value);
 
             var displayBounds = CurrentDisplay.Bounds;
@@ -342,7 +339,7 @@ namespace osu.Framework.Platform
             var windowX = (int)Math.Round((displayBounds.Width - windowSize.Width) * configPosition.X);
             var windowY = (int)Math.Round((displayBounds.Height - windowSize.Height) * configPosition.Y);
 
-            Position = new Point(windowX + displayBounds.X, windowY + displayBounds.Y);
+            return new Point(windowX + displayBounds.X, windowY + displayBounds.Y);
         }
 
         private void updateWindowPositionConfigFromCurrent()
@@ -397,16 +394,17 @@ namespace osu.Framework.Platform
         /// </summary>
         public virtual void Create()
         {
-            SDL.SDL_WindowFlags flags = SDL.SDL_WindowFlags.SDL_WINDOW_OPENGL |
-                                        SDL.SDL_WindowFlags.SDL_WINDOW_RESIZABLE |
-                                        SDL.SDL_WindowFlags.SDL_WINDOW_ALLOW_HIGHDPI |
-                                        SDL.SDL_WindowFlags.SDL_WINDOW_HIDDEN | // shown after first swap to avoid white flash on startup (windows)
-                                        WindowState.ToFlags();
+            const SDL.SDL_WindowFlags flags = SDL.SDL_WindowFlags.SDL_WINDOW_OPENGL |
+                                              SDL.SDL_WindowFlags.SDL_WINDOW_RESIZABLE |
+                                              SDL.SDL_WindowFlags.SDL_WINDOW_ALLOW_HIGHDPI |
+                                              SDL.SDL_WindowFlags.SDL_WINDOW_HIDDEN; // shown after first swap to avoid white flash on startup (windows)
 
             SDL.SDL_SetHint(SDL.SDL_HINT_WINDOWS_NO_CLOSE_ON_ALT_F4, "1");
 
-            SDLWindowHandle = SDL.SDL_CreateWindow(title, Position.X, Position.Y, Size.Width, Size.Height, flags);
+            position = getWindowPositionFromConfig();
+            Size = sizeWindowed.Value;
 
+            SDLWindowHandle = SDL.SDL_CreateWindow(title, Position.X, Position.Y, Size.Width, Size.Height, flags);
             Exists = true;
 
             MouseEntered += () => cursorInWindow.Value = true;
@@ -978,7 +976,7 @@ namespace osu.Framework.Platform
 
                     SDL.SDL_SetWindowSize(SDLWindowHandle, Size.Width, Size.Height);
 
-                    updateWindowPositionFromConfig();
+                    Position = getWindowPositionFromConfig();
                     break;
 
                 case WindowState.Fullscreen:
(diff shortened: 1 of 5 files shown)


osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -394,16 +383,17 @@ public SDL2DesktopWindow()
...
+            const SDL.SDL_WindowFlags flags = SDL.SDL_WindowFlags.SDL_WINDOW_OPENGL |
+                                              SDL.SDL_WindowFlags.SDL_WINDOW_RESIZABLE |
+                                              SDL.SDL_WindowFlags.SDL_WINDOW_ALLOW_HIGHDPI |
+                                              SDL.SDL_WindowFlags.SDL_WINDOW_HIDDEN; // shown after first swap to avoid white flash on startup (windows)
 
             SDL.SDL_SetHint(SDL.SDL_HINT_WINDOWS_NO_CLOSE_ON_ALT_F4, "1");
 
-            SDLWindowHandle = SDL.SDL_CreateWindow(title, Position.X, Position.Y, Size.Width, Size.Height, flags);
+            position = getWindowPositionFromConfig();
+            Size = sizeWindowed.Value;
```
bdach, line 394: Quoting the OP:

> the window would switch to maximised state while its already positioned and sized to config values, unlike previously where it would switch to maximised state while its position is 0 and size is `default_width,default_height` (1366x768)

What difference does it make if it's maximised anyway? Does it matter on multi-display setups? I'm really confused by these sets.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -1019,7 +1013,15 @@ private void updateWindowStateAndSize()
         private void updateMaximisedState()
         {
             if (windowState == WindowState.Normal || windowState == WindowState.Maximised)
```
bdach, line 1015: Not 100% related as this was here already, but I'm super confused by this guard. Does it need to exist?

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: Sorry if I wasn't clear, I meant that when the window is restored from initial maximised state by user, it'll restore back to the config values set.

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: Yes, the `maximised` field should not be overwritten on irrelevant window states, since that field is supposed to save whether the window was maximised in windowed mode or not.

osu.Framework/Platform/SDL2DesktopWindow.cs
bdach: But if the window is restored, `updateWindowStateAndSize` should surely fire? Why restore on window init if that code is also supposed to do that?

osu.Framework/Platform/SDL2DesktopWindow.cs
bdach: I kinda feel I'd prefer this to be written as

```csharp
switch (windowState)
{
	case WindowState.Normal:
		windowMaximised.Value = false;
		break;

	case WindowState.Maximised;
		windowMaximised.Value = true;
		break;

	// other window states should not be modifying the maximised flag state
}
```

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -1013,6 +1000,10 @@ private void updateWindowStateAndSize()
 
             updateMaximisedState();
 
+            // Functions like SDL_MaximizeWindow or SDL_RestoreWindow show the window
+            // even if they were hidden previously, which is not what we want.
+            updateWindowVisibility(Visible);
```
peppy, line 1005: does this not cause a flicker where the window is temporarily visible?

peppy (changes requested):

osu.Framework/Platform/SDL2DesktopWindow.cs
peppy: As expected, this does cause a flicker:

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: No, the restore is performed externally, not internally where we would have to rely on `updateWindowStateAndSize()`, I'll look through removing the actions invoked there if possible as looks like they misled you.

bdach: I'm not sure I follow. Are you saying that SDL does not currently listen to maximise/restore events? If so why isn't that done to fix this? That sounds like a shortcoming.

Also why do you want to remove anything from `updateWindowPositionAndSize`? Hopefully you're not suggesting that that function should no longer update position and size?

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: Looks like I got tricked into thinking that it doesn't flicker (*thanks windows.*)

Unfortunately, I couldn't get to completely get rid of the "initial white flash" issue, but I've limited its duration of occurrence to a number of milliseconds, that's the best I could do. (doesn't look noticeable on a booted windows, and shows for an average of 90ms on VM).

Not sure about the naming of `IGraphicsBackend.ResetBuffer` and the existence of `IWindow.MaxSize` though, only added for helping achieve proper results. Thoughts?

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: > Also why do you want to remove anything from updateWindowPositionAndSize? Hopefully you're not suggesting that that function should no longer update position and size?

Yeah, as `SDL_RestoreWindow` should take care of that, but scratch that since we have borderless mode changing the window position and size for doing the 1px hack.

> I'm not sure I follow. Are you saying that SDL does not currently listen to maximise/restore events? If so why isn't that done to fix this? That sounds like a shortcoming.
>
> If the issue is that there is no event to respond to on launch, then I'd be fine with calling the aforementioned function manually at window creation, too, so long as that logic is in one place and not two.

OK so, in the current implementation of SDL window handling, we handle two types of changes:
1. External changes: which could be done by the user resizing the window manually, or maximizing the window by the "maximize" button, or minimizing it, for example. 
 These are handled through windows events by updating our current states with them (check `updateWindowSpecifics()` and the switch cases), as we're not aware of them, and would allow us to handle internal changes properly without being desynced from the window's current state.
2. Internal/programmatic changes: which are done by the consumer/framework changing `WindowState` for example.
 These are handled by performing the required SDL actions for them (`updateWindowStateAndSize()` for the above case).

In the highlighted lines above, the window must be positioned and sized according to the config, such that when the user restores the window (**externally**, not programmatic via `updateWindowStateAndSize()`), they would be restored to the position and size they should be.

I've gone with the way we currently handle SDL window changes here, listening to restore events and changing the window's position and size according to the config by them doesn't sound correct (and would not be easy to handle as an `SDL_WINDOWEVENT_SIZE_CHANGED` event would be sent before the `SDL_WINDOWEVENT_RESTORED` event, also no event on launch)

Hope that clears up everything. (and I hope I haven't misunderstood anything)

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: Feel free to open a PR on that.

osu.Framework/Platform/SDL2DesktopWindow.cs
bdach: is it not possible to check whether the window is visible *before* attempting to restore maximised state from config? or otherwise delay the restoration until the window has been made visible?

frenzibyte: The issue here is that `SDL_MaximizeWindow()` makes the window visible regardless of whether it was hidden, therefore bringing back the "initial white content flash" issue.

For some reason even after performing a `glClear()` to the black colour and swapping buffers both before and after the `SDL_MaximizeWindow()` call (doing it after made no changes at all), there's a slight duration (around 100ms) where the window would show all white, not sure if it's because I'm running this in VM though, but I would consider this issue fixed on our side, as both https://github.com/ppy/osu-framework/pull/4089/commits/56b6b729b8a64b67b1bc144b0bebe74b3a3c64f0 and https://github.com/ppy/osu-framework/pull/4089/commits/515f4e2325f5331458cd3cd1563a2b567dbdbcad is all that can be done here.

bdach: i got that part. what i tried to say is, we can check whether the window is hidden, and not try to maximise the window if it is, to prevent the flicker from ever happening in the first place, instead of applying weird hacks to work around that.

frenzibyte: when would the window be maximised then?

bdach: there are events fired on the window showing/hiding (`SDL_WINDOWEVENT_SHOWN`/`SDL_WINDOWEVENT_HIDDEN`). those could maybe work, no?

i don't think using a magic 9999x9999 size (sounds potentially dangerous from a memory standpoint) and exposing a new method to be able to jam in said magic size to "clear to black" does not "a proper way" make.

frenzibyte: > also from the gif attached above, it doesn't even look like it's the actual window contents flickering, but the titlebar of the windiw.

that is from https://github.com/ppy/osu-framework/pull/4089/commits/c5d916e5f0e3b32e60d76f173f06ed778c16b7d5 hiding the window after becoming visible, therefore causing a flicker on startup. I've fixed that by reverting it and looking for an alternative way to do it.

> there are events fired on the window showing/hiding (`SDL_WINDOWEVENT_SHOWN`/`SDL_WINDOWEVENT_HIDDEN`). those could maybe work, no?

this will make the window switch into maximised state, after it has became visible from the first draw, not sure if that's something fine to be staying as-is.

> I think using a magic 9999x9999 size (sounds potentially dangerous from a memory standpoint) and exposing a new method to be able to jam in said magic size to "clear to black" does not "a proper way" make.

oh right, didn't realize that dangerous downside of the 9999x9999 magic..., my bad. I'll experiment more on alternative ways to revert https://github.com/ppy/osu-framework/pull/4089/commits/515f4e2325f5331458cd3cd1563a2b567dbdbcad with.

peppy: It's the whole window appearing for a frame. the gif doesn't capture because it's too low fps wise.

frenzibyte:
Closing this PR for now as it needs more work and apparently is broken on macOS.

PR #4092, opened 2020-12-12
Fix window size doubling up every startup on macOS retina displays

### LINKED ISSUE #4088: Window size doubles everytime program is relaunched in SDL
Specs:
16-inch MacBook Pro with higher-DPI retina display
Intel core i9-9980HK, AMD Radeon Pro 5500m
MacOS Catalina 10.15.7

On my display, there's a bug in which, if the window does not take up the maximum space it is allowed (screen display minus the dock at the bottom), the next time I launch the program, the window doubles in size, but the content of the program takes up only the bottom left quadrant of the screen as pictured below:

About to close the program with this resolution:
Screen Shot 2020-12-10 at 11 04 35 AM

Relaunching the program looks like this (changed background to yellow is normal):
Screen Shot 2020-12-10 at 11 04 49 AM

This is an issue with all `osu-framework` games, including osu!:
Screen Shot 2020-12-10 at 11 25 36 AM
 
My guess is that, although the program content correctly accounts for my retina display's layout of 2x2 physical pixels per logical pixel, the window assigns the window size in physical pixels to be the size in _logical pixels_, doubling its size.

When the window is resized at all, the program realizes its mistake and corrects the resolution so the program takes up the entire window. When the window reaches the maximum space it is allowed, it no longer tries to double in size and therefore doesn't display this bug, which I why I had never caught it before when I tested the solution for issue #4044 since I work at that resolution by default.

frenzibyte:
Closes #4088 

- [x] Depends on #4093 (for merge ease)

The reason is as mentioned here https://github.com/ppy/osu-framework/issues/4088#issuecomment-742722081.

Fixed by removing the scale on `sizeWindowed` bindable as it should not have it on. Also adds `updateWindowSize()` before `WindowMode.TriggerChange()` to get the correct display scale for setting up normal state properly.

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -412,6 +412,7 @@ namespace osu.Framework.Platform
             graphicsBackend.Initialise(this);
 
             updateWindowSpecifics();
+            updateWindowSize();
             WindowMode.TriggerChange();
         }
 
@@ -479,7 +480,7 @@ namespace osu.Framework.Platform
                 if (windowState == WindowState.Normal)
                 {
                     windowStateChanging = true;
-                    sizeWindowed.Value = newSize;
+                    sizeWindowed.Value = scaleSize(newSize, 1 / Scale);
                     windowStateChanging = false;
                 }
 
@@ -968,13 +969,13 @@ namespace osu.Framework.Platform
             switch (windowState)
             {
                 case WindowState.Normal:
-                    Size = sizeWindowed.Value;
+                    Size = scaleSize(sizeWindowed.Value, Scale);
 
                     SDL.SDL_SetWindowBordered(SDLWindowHandle, SDL.SDL_bool.SDL_TRUE);
                     SDL.SDL_SetWindowFullscreen(SDLWindowHandle, (uint)SDL.SDL_bool.SDL_FALSE);
                     SDL.SDL_RestoreWindow(SDLWindowHandle);
 
-                    SDL.SDL_SetWindowSize(SDLWindowHandle, Size.Width, Size.Height);
+                    SDL.SDL_SetWindowSize(SDLWindowHandle, sizeWindowed.Value.Width, sizeWindowed.Value.Height);
 
                     updateWindowPositionFromConfig();
                     break;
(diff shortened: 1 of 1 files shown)


osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -1263,6 +1264,8 @@ private static DisplayMode displayModeFromSDL(SDL.SDL_DisplayMode mode, int disp
             return new DisplayMode(SDL.SDL_GetPixelFormatName(mode.format), new Size(mode.w, mode.h), bpp, mode.refresh_rate, modeIndex, displayIndex);
         }
 
+        private static Size scaleSize(Size size, float scale) => new Size((int)(size.Width * scale), (int)(size.Height * scale));
```
bdach, line 1267: can be written as

```suggestion
        private static Size scaleSize(Size size, float scale) => Size.Round(size * scale);
```

or

```suggestion
        private static Size scaleSize(Size size, float scale) => (size * scale).ToSize(); // truncates
```

at which point the helper is somewhat questionable

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -415,6 +415,7 @@ public virtual void Create()
             graphicsBackend.Initialise(this);
 
             updateWindowSpecifics();
+            updateWindowSize();
```
bdach, line 418: the only other call of this method is behind a `WindowMode == Windowed` guard. are you sure this is safe to call in borderless/fullscreen?

frenzibyte: Yeah, your suggestion of moving config saving out of `updateWindowSize()` would make things better, as that guard in the other call existed due to it.

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: Oh, I didn't realize you could do that..., will remove.

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: Resolved with merging #4093 into here, as it has that suggestion pushed there. This will now depend on that PR being merged.

peppy (approved):

PR #4093, opened 2020-12-12
Fix continuous resize events no longer handled

frenzibyte:
Regressed in https://github.com/ppy/osu-framework/pull/4085

The aforementioned PR (by me..) removed the case handling the `RESIZED` event thinking that it gets fired before the SDL has processed it, while it actually gets sent after the `SIZE_CHANGED` event, [as noted there](https://github.com/ppy/osu-framework/pull/4085#issuecomment-742707285).

Turns out that there was an event filter set to listen for it to handle continuous resize events (as `SDL_PollEvent` blocks those), and since the `SDL_WINDOWEVENT_RESIZED` case was removed, these events are no longer handled, causing the window to not update the client size while resizing:

This fixes it by updating the client size on those continuous resize events, **but** without saving the new size to config, as we don't know what state the window would end up in once finishing from the external change (i.e. switching to maximised mode):

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -330,43 +330,10 @@ namespace osu.Framework.Platform
             }
         }
 
-        private void readWindowPositionFromConfig()
-        {
-            if (WindowState != WindowState.Normal)
-                return;
-
-            var configPosition = new Vector2((float)windowPositionX.Value, (float)windowPositionY.Value);
-
-            var displayBounds = CurrentDisplay.Bounds;
-            var windowSize = sizeWindowed.Value;
-            var windowX = (int)Math.Round((displayBounds.Width - windowSize.Width) * configPosition.X);
-            var windowY = (int)Math.Round((displayBounds.Height - windowSize.Height) * configPosition.Y);
-
-            Position = new Point(windowX + displayBounds.X, windowY + displayBounds.Y);
-        }
-
-        private void storeWindowPositionToConfig()
-        {
-            if (WindowState != WindowState.Normal)
-                return;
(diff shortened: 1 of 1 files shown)


osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -467,7 +469,11 @@ public void Run()
             SDL.SDL_Quit();
         }
 
-        private void updateWindowSize()
+        /// <summary>
+        /// Updates the client size and the scale according to the window.
+        /// </summary>
+        /// <param name="saveToConfig">Whether the new window size should be saved to the config.</param>
+        private void updateWindowSize(bool saveToConfig = true)
```
bdach, line 476: Given that this method has only two usages I'm opposed to having this default param. Would propose to split saving to config to a separate method, which the normal event loop will call but the event filter won't.

bdach (changes requested):

peppy (approved):

peppy (approved):

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -899,8 +872,10 @@ private void handleWindowEvent(SDL.SDL_WindowEvent evtWindow)
                     break;
 
                 case SDL.SDL_WindowEventID.SDL_WINDOWEVENT_SIZE_CHANGED:
-                    if (WindowMode.Value == Configuration.WindowMode.Windowed)
-                        updateWindowSize();
+                    updateWindowSize();
+                    if (WindowState == WindowState.Normal)
+                        storeWindowSizeToConfig();
```
peppy, line 877: this breaks switching from borderless to window.

PR #4094, opened 2020-12-13
Rename a few members in SDL2DesktopWindow for readability

bdach:
Noticed while looking through #4093.

The whole window state management really needs refactoring anyway, but this is not intended to be that - this PR has just a few simple renames to keep the WTF factor down for now.

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -935,7 +935,7 @@ namespace osu.Framework.Platform
         {
             // this method is potentially called from another thread (see event filter usage).
             // this flag ensures such calls don't interfere with a user-requested screen mode change.
-            if (isChangingWindowState)
+            if (windowStateAndSizeUpdateRunning)
                 return;
 
             Debug.Assert(SDLWindowHandle != IntPtr.Zero);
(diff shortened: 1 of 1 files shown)


osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -1078,11 +1078,14 @@ private MouseButton mouseButtonFromEvent(byte button)
...
+        /// Set to <c>true</c> while the window size is being stored to config to avoid bindable feedback.
         /// </summary>
-        private bool windowStateChanging;
+        private bool storingSizeToConfig;
 
-        private bool isChangingWindowState;
+        /// <summary>
+        /// Set to <c>true</c> when a call to <see cref="updateWindowStateAndSize"/> is in progress.
+        /// </summary>
+        private bool windowStateAndSizeUpdateRunning;
```
frenzibyte, line 1088: I would have renamed it to `processingInternalChange` or something similar to that, but I guess there's no harm with the current name here.

frenzibyte (approved):

osu.Framework/Platform/SDL2DesktopWindow.cs
bdach: i don't know what an "internal change" is. this name directly references the method the flag relates to.

peppy (approved):

PR #4095, opened 2020-12-14
Fix zero-length repeat delegates ignoring initial delay

peppy:
A bit of an oversight that we somehow hadn't encountered until now.

The conditional was far too early in the flow, meaning that `ExecutionTime` is completely ignored in the case that `RepeatInterval` is zero.

### DIFF
--- a/osu.Framework/Threading/Scheduler.cs
+++ b/osu.Framework/Threading/Scheduler.cs
@@ -30,7 +30,12 @@ namespace osu.Framework.Threading
         /// <summary>
         /// Whether there are any tasks queued to run (including delayed tasks in the future).
         /// </summary>
-        public bool HasPendingTasks => runQueue.Count > 0 || timedTasks.Count > 0 || perUpdateTasks.Count > 0;
+        public bool HasPendingTasks => TotalPendingTasks > 0;
+
+        /// <summary>
+        /// The total number of <see cref="ScheduledDelegate"/>s tracked by this instance for future execution.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


osu.Framework/Threading/Scheduler.cs
```diff
@@ -133,7 +133,9 @@ private void queueTimedTasks()
                             tasksToSchedule.Add(sd);
                         }
 
-                        if (!sd.Completed)
+                        if (sd.RepeatInterval == 0)
+                            perUpdateTasks.Add(sd);
```
smoogipoo, line 137: Isn't this going to be scheduled twice, because of the above block?

smoogipoo: Also, does this actually work the first time around? Shouldn't the `runQueue.Enqueue()` still happen?

peppy: The first run is handled by `Update`
image

I'll check / add test coverage of the double case, if that's actually a thing.

peppy: Regarding your first comment, I don't believe so? That code is adding the next iteration of the repeat.

smoogipoo: image

This is what I'm referring to.

peppy: aha, i see.

smoogipoo (approved):
lgtm

smoogipoo (approved):

PR #4096, opened 2020-12-14
Explicitly request compatibility profile

smoogipoo:
Fixes RenderDoc not attaching to o!f projects under SDL.


peppy (approved):
Tested working on macOS and windows.

PR #4099, opened 2020-12-15
Fix multiple remaining issues with window state changes

### LINKED ISSUE #4097: Changing from fullscreen to borderless doesn't work
Tested on current windows 10 in osu!lazer; seems to just stay on fullscreen mode.

### LINKED ISSUE #4098: Changing screen modes in a certain order causes the window border to disappear
Changing screen modes in the following order causes the window border to disappear (on Windows 10):

`Windowed`→`Borderless`→`Fullscreen`→`Windowed`.
It's very easy to reproduce for people who use <kbd>Alt</kbd>+<kbd>Enter</kbd> to switch between those modes.

(**Can be fixed by switching between borderless and windowed mode**)

example SDL

Works fine with `--tk` argument:

example tk

Tested on the latest osu!lazer release ([2020.1212.0](https://github.com/ppy/osu/releases/tag/2020.1212.0)).

peppy:
Changes explained in individual commit comments.

Closes #4098 
Closes #4097 

Tested on macOS and windows.


osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -1021,6 +1028,10 @@ private void storeWindowPositionToConfig()
             var windowX = Position.X - displayBounds.X;
             var windowY = Position.Y - displayBounds.Y;
 
+            // mainly here to handle garbage values (SDL_WINDOWEVENT_SIZE_CHANGED are delayed in borderless fullscreen on windows).
+            if (!displayBounds.Contains(windowX, windowY))
+                return;
```
bdach, line 1033: Not sure if this is the correct place for this guard. It should probably be in `handleWindowEvent()` itself - having it here means that the window will look correct, but the garbage `(-1, -1)` location is getting stored to fields and passed to event subscribers higher up:

https://github.com/ppy/osu-framework/blob/183cd5ae162c183283c37d76ed96d7a9d5d8839b/osu.Framework/Platform/SDL2DesktopWindow.cs#L866-L871

peppy: Yeah, I did consider moving it due to that, but wasn't sure if we wanted to fake the position internally. I mean, the window *is* technically at that location so I think it may be best to still say it is?

The rationale is that I'm just looking to avoid saving this to config.

bdach: I'll double check but when I was checking yesterday it seemed that it wasn't the actual window position, but it was almost as if SDL was submitting the wrong/stale position event (from borderless mode instead of windowed). Looked like an upstream bug to me...

peppy: Ah, right. I completely forgot that is the case here. We may be best off moving this for now then, yeah.

bdach (approved):

bdach:
never mind, re-checked windows but did not check actual windowed (not tiled) linux properly before approving - there's some sort of issue with linux in which setting window size via framework does not take effect until the mouse leaves window bounds. 

repro scenario:

* open `TestSceneFullscreen` in windowed mode
* click the `change window size step`
* nothing happens
* leave window bounds w/ mouse
* new window size takes effect

does not happen on `master`, seemingly caused by 183cd5ae16. will debug tomorrow if i'm not beaten to it.

bdach (changes requested):
as above

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -1113,15 +1115,15 @@ public void SetupWindow(FrameworkConfigManager config)
...
-                    ScheduleCommand(updateWindowStateAndSize);
+                    pendingWindowState = windowState;
             };
 
             sizeWindowed.ValueChanged += evt =>
             {
                 if (storingSizeToConfig) return;
 
                 if (windowState == WindowState.Normal)
-                    ScheduleCommand(updateWindowStateAndSize);
```
bdach, line 1124: The aforementioned regression is caused by no longer running this command in case of a size change - do you recall whether there was a reason not to fire it? Wouldn't actually be surprised if the scenario I described wasn't linux-specific at all, but I haven't checked yet.

peppy: The theory was that it would be run via the `pendingWindowState` being set to a new value (https://github.com/ppy/osu-framework/pull/4099/files/35be2f82191792a664fedf3f6b435ada404b0db1#diff-802f1281748362cec8a6ce40039ebf6ae50bd82897f7d4c0699df24234b7a5e3R924). Possibly an order of execution change?

bdach: Right, but that relies on `updateWindowSpecifics()` to be called, which it won't on this code path (at least not without some user interaction), because the scenario above is an internal size change coming from the framework. Think it's fine to just revert?

peppy: I think it should probably be updated to 
`ScheduleCommand(updateWindowSpecifics);` for conformity? That seems to be what we are doing elsewhere.

bdach: That does sound better, except for the fact that it trips this assert on window load due to binding to config happening before window construction...

https://github.com/ppy/osu-framework/blob/1fe981748d620ddd033fe5de7c406664727a07e2/osu.Framework/Platform/SDL2DesktopWindow.cs#L912

Probably would have to early-return instead of asserting non-null...?

bdach: Actually never mind, the bindable subscriptions are just ordered wrong (`BindTo()` after a `BindValueChanged(..., false)`). Reordering also works, will push that version for Windows/macOS testing.

bdach: I've changed the assert to an early-return and added a comment with rationale. It's not pretty at all but I suppose it's somewhat defensible and I mostly wanted to return this to a state of relative sanity at least.

bdach:
Issue mentioned above is fixed for me with the latest commit, but this needs re-testing on Windows and macOS. I can grab Windows but will need help with macOS.

bdach:
Tested windows. Found #4105 along the way and [resizing via window border is broken as heck](https://discordapp.com/channels/188630481301012481/589331078574112768/790320148343029771), but this PR seems to have achieved what it intended to wrt the two issues linked in the OP.

bdach (approved):

peppy:
Hmmmmmm... the end goal of my previous changes was to reduce the number of places that method was called/scheduled in weird ways. I've pushed another commit which should give the same behaviour as your change, while *hopefully* being a little less ugly. Yes, it will run some of the `get` portion twice if there is a window event queued but I'm not sure that really matters.

I've tested to some level on windows and it seems to check out. Can you double confirm (and check whether you think the code reads better or not)?

bdach:
Reads better to have one call site for sure, and theoretically it should work. Haven't actually tested on a machine yet, though.

bdach:
Tested linux and windows, seems fine; additionally profiled, overhead of calling `updateWindowSpecifics()` every update is negligible (<0.01% of total cpu time in framework tests). I say we get this in.

bdach (approved):

PR #4101, opened 2020-12-16
Add more platform actions for tab management

bdach:
Intended for (eventual) usage with https://github.com/ppy/osu/pull/11182; tested therein. Did not add any handling for `TabControl`, as framework tab controls are not necessarily meant to have mutable item sets.

Unsure about the naming. <kbd>Ctrl</kbd>+<kbd>T</kbd> and <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>T</kbd> feel wrong to name as `DocumentNew` or similar, since it's a strictly tab-related action, while <kbd>Ctrl</kbd>+<kbd>W</kbd> / <kbd>Ctrl</kbd>+<kbd>F4</kbd> will work in some multi-document applications that are not necessarily tabbed.


osu.Framework/Platform/GameHost.cs
```diff
@@ -981,6 +981,10 @@ public void Dispose()
             new KeyBinding(new KeyCombination(InputKey.Control, InputKey.PageDown), new PlatformAction(PlatformActionType.DocumentNext)),
             new KeyBinding(new KeyCombination(InputKey.Control, InputKey.Tab), new PlatformAction(PlatformActionType.DocumentNext)),
             new KeyBinding(new KeyCombination(InputKey.Control, InputKey.Shift, InputKey.Tab), new PlatformAction(PlatformActionType.DocumentPrevious)),
+            new KeyBinding(new KeyCombination(InputKey.Control, InputKey.W), new PlatformAction(PlatformActionType.DocumentClose)),
+            new KeyBinding(new KeyCombination(InputKey.Control, InputKey.F4), new PlatformAction(PlatformActionType.DocumentClose)),
```
peppy, line 985: Not sure if we want this to be global. On macOS this should be cmd+w only.

osu.Framework/Platform/GameHost.cs
```diff
@@ -981,6 +981,10 @@ public void Dispose()
...
+            new KeyBinding(new KeyCombination(InputKey.Control, InputKey.T), new PlatformAction(PlatformActionType.TabNew)),
+            new KeyBinding(new KeyCombination(InputKey.Control, InputKey.Shift, InputKey.T), new PlatformAction(PlatformActionType.TabRestore)),
```
peppy, line 987: Likewise (arguably) should be cmd+ on macOS. Although I believe most apps also allow ctrl+ presses as well, so maybe we support both if anything?

bdach: <kbd>Cmd</kbd> is `InputKey.Super`, right? Will do.

peppy: Yep, that's the one.

osu.Framework/Platform/GameHost.cs
bdach: Upon checking macOS actually has [an entirely disparate set of platform action bindings](https://github.com/ppy/osu-framework/blob/99475763d16a0fa07319990612f8f23a156fe6f0/osu.Framework/Platform/MacOS/MacOSGameHost.cs#L58-L93) and does not use these base ones at all, so as far as I can see there's nothing that would need changing.

Maybe Linux is arguable but I have <kbd>Ctrl</kbd>+<kbd>F4</kbd> working both in Firefox and in Rider on Linux so /shrug

peppy (approved):

PR #4103, opened 2020-12-17
Add filtering mode support to FontStore

### LINKED ISSUE #4102: Can't change FontStore filtering mode
I'm trying to import a pixel font, VT323, but FontStore defaults to linear filtering with no apparent way to change it. The result is not great when I start scaling text up.
I can open up a PR to pass filtering mode through to the underlying TextureStore, is there any reason why this isn't an option already?

boswelja:
The only changes needed were to pass filteringMode into the underlying TextureStore.
FontStoreTest needed changes to clear up constructor ambiguity, which shouldn't be a problem for projects that depend on the framework seeing as there's only 1 public constructor.
Resolves #4102


bdach:
@boswelja are you sure this works correctly? do you have a branch or something where this can be tested? because I'm not 100% sure this is working ok. I ~~stole~~ borrowed your project over at https://github.com/Giga-Pixel-Rumble/GigaPixelRumble, and after I blow up the font size large enough I get this:

2020-12-17-224123_1261x665_scrot

which means one of two things: either I'm not doing this correctly or this won't have the desired effect in all cases...

boswelja:
Oh interesting, I was having a hard time loading up these changes in the same project before opening the PR but have got it now and can reproduce what you're seeing. I'm unsure why this could be happening, any ideas?

osu.Framework.Tests/IO/FontStoreTest.cs
```diff
@@ -24,8 +25,8 @@ public void OneTimeSetUp()
         [Test]
         public void TestNestedScaleAdjust()
         {
-            using (var fontStore = new FontStore(new RawCachingGlyphStore(fontResourceStore, "OpenSans") { CacheStorage = storage }, scaleAdjust: 100))
-            using (var nestedFontStore = new FontStore(new RawCachingGlyphStore(fontResourceStore, "OpenSans-Bold") { CacheStorage = storage }, 10))
+            using (var fontStore = new FontStore(new RawCachingGlyphStore(fontResourceStore, "OpenSans") { CacheStorage = storage }, scaleAdjust: 100, All.Linear))
```
peppy, line 28: linear is the default and not required to be specified here. if you were looking to add coverage of the new scaling mode, please add a new test method to this class rather than using an existing, and explicitly test `NearestNeighbour`.

peppy (changes requested):

peppy:
@bdach were you testing using `NearestNeighbour`? If you happened to go based off the changes made in this diff you may have inadvertently chosen the existing default of `Linear`.

osu.Framework.Tests/IO/FontStoreTest.cs
boswelja: I was getting ambiguous constructor errors without it unfortunately, it shouldn't affect production as the other constructor is internal.
I'll look into adding a filtering option test.

boswelja:
@peppy I was able to reproduce it on my end with filteringMode set to All.Nearest. Changing filteringMode doesn't seem to change the resulting font at all unfortunately

peppy:
Looks like this wasn't ever handled correctly for the case of automatic mipmapping (only the min filter case is):

https://github.com/ppy/osu-framework/blob/99475763d16a0fa07319990612f8f23a156fe6f0/osu.Framework/Graphics/OpenGL/Textures/TextureGLSingle.cs#L473

Could you try copying the logic from the min filter to the max filter and see if it helps?

bdach:
Yeah I did specify linear filtering, and also was aware of the ctor ambiguity.

I think there might be filtering happening on `SpriteText` level or something like that, that's preventing this from working properly.

bdach:
Upon taking a look at this the mag filter isn't the issue here (although its handling may well be wrong as pointed out above, I haven't checked spec or anything). There are two issues at play here:

First, the actual assets as exported by bmfont are antialiased:

  2020-12-20-143420_695x339_scrot

That's a tooling shortcoming and probably unfortunately out of scope of this PR.

Secondly, the root framework font store uses the default filtering mode of `Linear` and can't be overridden. The naive approach of caching a new one over it leads to the following crash:

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at osu.Framework.IO.Stores.RawCachingGlyphStore.createCachedPageInfo(Int32 page) in /home/dachb/Dokumenty/opensource/osu-framework/osu.Framework/IO/Stores/RawCachingGlyphStore.cs:line 54
   at osu.Framework.IO.Stores.RawCachingGlyphStore.LoadCharacter(Character character) in /home/dachb/Dokumenty/opensource/osu-framework/osu.Framework/IO/Stores/RawCachingGlyphStore.cs:line 38
```

This is due to a lack of a cache storage provided to the cached font store, which framework consumers can't actually provide themselves, as the ctor allowing to do so is internal (and we probably shouldn't be exposing it). Allowing a method to specify the default filtering mode should probably be addressed in this PR, but I'm unsure yet of the method to do so.

boswelja:
Oh interesting, so it seems like we can't really test these changes anyways unless we can get the fonts exported without antialiasing / smoothing? If so, theres not really much point in continuing work on this PR until that's fixed. From my understanding simply passing through filteringMode won't do what I intended either so should I close this PR for now too?

bdach:
Partially yeah, as long as the assets are antialiased there's not much point fiddling with the filtering mode. Don't know about closing the PR but I guess it's not too much work to recreate if need be in the future...?

boswelja:
It's my understanding that passing in filteringMode is doing nothing anyways, is that right? If not I can load up the changes and test, as I've worked out how to disable antialiasing in BMFont.
Might be worth noting which settings are optional in the custom fonts section of the framework wiki though.

peppy:
I wouldn't say this needs to be closed. Since we've identified multiple underlying issues they should also be fixed here. For testing purposes, can you attach the non-aa'd font in a comment?

boswelja:
[PressStartP2.zip](https://github.com/ppy/osu-framework/files/5727999/PressStartP2.zip)
It's a different font from the original anti-aliased font as the last one wasn't quite square which was causing some pretty major issues with BMFont. The same idea applies to this one though, and it's a lot sharper to start with after disabling anti-aliasing.

boswelja:
Apologies for the delay, have been busy recently. I'm working on tracking down the issues identified here now, should I include fixing the max filter logic in TextureGLSingle with this PR? I cannot verify whether it helps this case directly until I come up with a solution for setting the default font store filter mode.

bdach:
I checked blame, and it seems that the min and mag filters *were* handled identically at one point before, but https://github.com/ppy/osu-framework/commit/38248e846bf6fceb41ed593d7e16050caa30e796 explicitly changed the handling to how it is now. I'm not sure I feel confident enough in stating as to why that was the case.

@Tom94 long shot, but do you recall by any chance why the two are handled differently?

boswelja:
While we wait for Tom to respond, it looks like new fonts are wrapped on a RawCachingGlyphStore, but I can't see a way to adjust how the textures are filtered in any of the loading mechanisms.Any ideas? Or am I barking up the wrong tree entirely, so to speak.

peppy:
`RawCachingGlyphStore` is just loading files ready for upload. It has no input on filtering modes. That would be the target `TextureStore`'s responsibility.

As for the change in https://github.com/ppy/osu-framework/commit/38248e846bf6fceb41ed593d7e16050caa30e796, I don't see anything too weird here? It looks like the first time filtering modes were added. I actually think this is correct, as you *can't* use mipmaps when filtering for magnification (mipmaps by definition are for smaller than original sizes, not larger). So I think the code on master is correct here.

bdach:
> I can't see a way to adjust how the textures are filtered in any of the loading mechanisms. Any ideas? Or am I barking up the wrong tree entirely, so to speak.

As I intended to convey above, the reason the fonts are anti-aliased is that the following store in `Game`:

https://github.com/ppy/osu-framework/blob/ac960afc8f154d3091fcc3b9d7ba2a9b9d4a54c6/osu.Framework/Game.cs#L153-L154

is using the default filtering mode of `All.Linear`. You can test this on framework `master` by passing `filteringMode: All.Nearest` in the internal ctor manually. Normally you could work around this by caching another in your game (to hide the main one), but that doesn't work because the game internally relies on `RawCachingGlyphStore`, which requires to create a `FontStore` with cache storage, which isn't possible because the constructor allowing to do so is not public.

The best way I can see out of this is to add a `CreateRootFontStore()` (or similar, not sure on the name) virtual method to `Game`, that by default uses the same way to construct the font store as is the case now, but which can be overridden by consumers as they see fit. To make that work properly you would probably have to ensure that the cache storage is installed to whatever the user created by doing something like

```csharp
Fonts = CreateRootFontStore()
{
	CacheStorage = cacheStorage;
}
```

where `CacheStorage` would be a property internal to the framework. @peppy do you think that would be an OK direction to take?

Tom94:
The reason min and mag filters are treated differently is because mipmaps don't make sense for magnification. Thus, the only valid mag filters are GL_NEAREST and GL_LINEAR, whereas for minification, there are also mipmap flavors.

Even though the program may run when you treat both filters identically, it'll likely not behave correctly. I'm fairly confident it won't make a different w.r.t. the problem this thread is about.
(more info at https://www.khronos.org/opengl/wiki/Sampler_Object )

All that said, it may be desirable to allow the user to set different mag- and min filters. Sometimes, you may want a mip-mapped min filter (smooth downscaling) with a GL_NEAREST mag filter (pixelated instead of blurred when zooming in).

boswelja:
@bdach Thanks for your help. I've come up with a solution and it seems to work fine in testing. I'll be pushing changes for review soon.
As for the mag filters, should I open another PR and force it to be either nearest or linear?

bdach:
Mag filters should definitely be separated out.

osu.Framework/Game.cs
```diff
@@ -51,6 +52,14 @@ public abstract class Game : Container, IKeyBindingHandler<FrameworkAction>, IKe
         /// </remarks>
         public FontStore Fonts { get; private set; }
 
+        /// <summary>
+        /// Sets the game-wide font store filtering mode.
+        /// </summary>
+        /// <remarks>
+        /// Currently child font store filtering will be overridden by this.
+        /// </remarks>
+        protected virtual All BaseFontStoreFilteringMode => All.Linear;
```
peppy, line 61:
```suggestion
        /// <summary>
        /// Sets the game-wide font store minification filtering mode. Defaults to Linear.
        /// </summary>
        /// <remarks>
        /// This property should be used instead of (and will override) setting directly on child font stores added to Fonts.
        /// </remarks>
        protected virtual All FontFilteringMode => All.Linear;
```

peppy (changes requested):

peppy (dismissed):
I think this is okay for now. If there are more similar settings in the future we may want to re-think the flow here, though.

bdach:
Truth be told I'm not 100% sold on this. `FontFilteringMode` being at game-level doesn't feel too right to me. Probably would have preferred my original proposal (being able to configure the root store completely) but not going to insist on it. May be good to get a third opinion.

peppy:
If we go with configuring the root store, it will need to be non-internal. I'm fine to wait on this for further opinion, though.

peppy:
Is there a reason we shouldn't add the filter mode as a parameter to `AddFont`? Seems like that could be a better place to handle this.

bdach:
Definitely would agree with configuring on a per-font basis. Not sure about whether the current framework structure can support it but it'd be 100% the best option.

peppy:
Pretty sure it can. I'll give it a try.

peppy:
The problem lies with how efficient we want this to be. The current flow of font creation guaranteed that multiple fonts will share atlas textures (the whole point of making everything as rigid as it is). We will need to decide for this change whether we:

- Want to spend the implementation overhead to make this implementation efficient (probably internally managed dictionaries per filter mode, adding the user fonts to the correct one so they can still share atlases).
- Leaving at is currently is, but having non-user fonts gain a different filter mode (will affect framework dialogs, for instance).
- Just externalising the `internal` ctor of `FontStore`, losing some rigidity but allowing the user to add a nested font store which will manage its atlas+minmode separate from the main one.

bdach:
Going off effort vs gain I'd probably advocate for going for something simple (so probably just publicising the ctor); I don't think there's that much interest for this feature. However I do recall trying to add a nested font store with nearest filtering but if I remember correctly it didn't work, but the exact reason as to why evades me right now. I think it might have been that the atlas textures themselves needed to also be created with a different filtering mode?

peppy:
Did you set `useAtlas` to true? if not, i believe it will inherit the parent `FontStore`'s atlas.

bdach:
Yeah that would make sense. The time I was checking at was when both the public and internal ctors had filtering mode params, and I couldn't do that from the consumer project because the ctor that allowed specifying `useAtlas` was `internal`.

peppy:
I've made a new constructor available specifically for creating a custom filtering font store.

Usage would be something as follows:

```csharp
            var nearestFonts = new FontStore(null, 100, minFilterMode: All.Nearest);
            Fonts.AddStore(nearestFonts);
            
            AddFont(Resources, "nearestNeighbourFont1", nearestFonts);
            AddFont(Resources, "nearestNeighbourFont2", nearestFonts);
```

I've tested this by adjusting the game fonts, but turns out it's not that easy to make actual test coverage for this (without making something like a `TestGameTestScene`) so have foregone that for the time being.

peppy:
@boswelja can you confirm this works fine for you?

boswelja:
Yep, seems to work fine for me. The difference that made this work via FontStore constructor is setting useAtlas to true right? I'll need to find out what an atlas is for and why that fixed it when I get the chance.
Thanks for helping me through this guys.

peppy:
Yep. If you have more than one font loading in your game, you should be able to see how the atlas is being used via the Ctrl+F3 view. Characters from multiple fonts should share the same backing texture.

If the atlas was to not be shared between `FontStore`s, each font would take up a *minimum* of 4mb memory (1024x1024x4), even if it only had a few characters.

bdach:
I've removed/reverted some no-longer-necessary bits and bobs, and fixed up the xmldocs (were referring to the wrong method to add a nested store I'm pretty sure). Other than that I'm good with this version.

bdach (approved):

smoogipoo (approved):

PR #4107, opened 2020-12-22
Store full namespace for last accessed test in test browser

peppy:
Without this, two test scenes with the same name but different namespace didn't behave as expected (the first instance would always be loaded on starting the test browser up).


bdach (approved):
tested, does what it says :+1:

PR #4109, opened 2020-12-26
Add support for horizontal mouse scroll wheels

bdach:
Closes https://github.com/ppy/osu/issues/11326.

First commit makes the game not-crash if the wheel is used in a binding, second adds actual support.

Tested with the `xdotool` as suggested in the issue thread and it seems to work fine, however - @bob8677 - it'd be nice if you were able to test this game-side with a framework checkout (steps how to do this are described [here](https://github.com/ppy/osu-framework/wiki/Testing-local-framework-checkout-with-other-projects)).

### DIFF
--- a/osu.Framework.Tests/Visual/Input/TestSceneKeyBindingsGrid.cs
+++ b/osu.Framework.Tests/Visual/Input/TestSceneKeyBindingsGrid.cs
@@ -78,9 +78,10 @@ namespace osu.Framework.Tests.Visual.Input
             }
         }
 
-        private void scrollMouseWheel(int dy)
+        private void scrollMouseWheel(int dx, int dy)
         {
-            AddStep($"scroll wheel {dy}", () => InputManager.ScrollVerticalBy(dy));
+            if (dx != 0) AddStep($"scroll wheel horizontally {dx}", () => InputManager.ScrollHorizontalBy(dx));
+            if (dy != 0) AddStep($"scroll wheel {dy}", () => InputManager.ScrollVerticalBy(dy));
         }
 
         private void check(TestAction action, params CheckConditions[] entries)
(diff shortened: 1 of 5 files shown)


bob8677:
I can confirm that this change works, however I think the buttons are reversed (moving the wheel right says wheel left and vice versa).

bdach:
That's pretty perplexing... I'd surely expect the positive X axis on the scroll to be right?

frenzibyte:
It's correct for me on my trackpad, scrolling left (by moving my fingers to the left) presses "scroll left".

bdach:
On my linux box, compared to firefox it does seem to be backwards, but [SDL docs seem to agree with my choice of coordinates](https://wiki.libsdl.org/SDL_MouseWheelEvent#Remarks) - but also helpfully state:

> SDL does not abstract the mouse wheel scroll directions to be consistent across all platforms (`SDL_MOUSEWHEEL_NORMAL`). If direction is `SDL_MOUSEWHEEL_FLIPPED` the values in `x` and `y` will be opposite. Multiply by -1 to change them back.

Quite helpful... Inclined to leave as-is as not in our scope.

bob8677:
I don't think its a big deal for it to be backwards in some situations, as its not commonly used and doesn't affect anything that much anyway, as it still works.

bob8677:
@frenzibyte could it be that way because of natural scrolling? Moving your fingers to the left is often set to moving things right. It seems wrong on my trackpad without natural scrolling turned on.

peppy:
Tested on windows (directions seem correct).

- Added basic test coverage
- Fixed the test browser display cursor being back-to-front for x scroll

peppy (dismissed):

bdach:
Thanks a bunch. I pushed one more tiny cosmetic change as the test scene was looking a little busted with the new buttons on it:

2020-12-27-124403_1920x1080_scrot

Afterwards:

2020-12-27-124523_1920x1080_scrot

peppy (approved):

peppy (approved):

PR #4110, opened 2020-12-26
Allow parsing same-type bindables

frenzibyte:
Fixes a case osu!-side in testing where invoking `ToMod()` on a non-deserialized (manually constructed) `APIMod` instance would attempt to parse a `Bindable`, and would result in an "Type must be IConvertible" exception because of falling to the `default` case in the parse method. Upon discussing, parsing `Bindable`s sounds OK to do.

### DIFF
--- a/osu.Framework.Tests/Bindables/BindableTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableTest.cs
@@ -57,7 +57,7 @@ namespace osu.Framework.Tests.Bindables
         }
 
         [Test]
-        public void TestParseSameTypeBindable()
+        public void TestParseBindableOfExactSameType()
         {
             var bindable1 = new BindableInt();
             var bindable2 = new BindableDouble();
@@ -79,13 +79,14 @@ namespace osu.Framework.Tests.Bindables
         }
 
         [Test]
-        public void TestParseSameInheritanceBindable()
+        public void TestParseBindableOfMatchingInterfaceType()
         {
+            // both of these implement IBindable<int>
             var bindable1 = new BindableInt(10) { MaxValue = 15 };
             var bindable2 = new Bindable<int>(20);
 
             bindable1.Parse(bindable2);
-            // ensure bindable allowed range is still respected.
+            // ensure MaxValue is still respected.
             Assert.That(bindable1.Value, Is.EqualTo(15));
 
             bindable2.Parse(bindable1);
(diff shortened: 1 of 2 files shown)


bdach:
Needs test coverage.

frenzibyte:
Added, that actually made me realize there's no direct check for different-type bindables. 😄

osu.Framework/Bindables/Bindable.cs
```diff
@@ -245,6 +245,13 @@ public virtual void Parse(object input)
                     Value = t;
                     break;
 
+                case IBindable _:
+                    if (!(input is IBindable<T> bindable))
+                        throw new ArgumentException($"Attempting to parse {input.GetType()}, which has differnet value type from {typeof(T)}", nameof(input));
```
bdach, line 250: has typo, but in general i'd recommend writing this as

```suggestion
                        throw new ArgumentException($"Expected bindable of type {nameof(IBindable)}<{typeof(T)}>, got {input.GetType()}", nameof(input));
```

osu.Framework.Tests/Bindables/BindableTest.cs
```diff
@@ -56,6 +56,25 @@ public void TestBindViaBindTarget()
...
+        [Test]
+        public void TestParseSameTypeBindable()
+        {
+            var bindable1 = new BindableInt();
+            var bindable2 = new BindableDouble();
+            var bindable3 = new BindableBool();
+
+            bindable1.Parse(new BindableInt(3));
+            bindable2.Parse(new BindableDouble(2.5));
+            bindable3.Parse(new BindableBool(true));
```
bdach, line 68: should also cover not-exactly-same-but-convertible types (`Bindable<int>` and `BindableInt`, etc.), and also things like `Bindable<string>`

bdach (changes requested):

osu.Framework.Tests/Bindables/BindableTest.cs
frenzibyte: Does this fulfill what you're suggesting? https://github.com/ppy/osu-framework/pull/4110/commits/1314d13a749de26db437288dec60fe6f96ada6f5

bdach (approved):
looks fine to me

peppy (approved):
logical

PR #4113, opened 2020-12-27
Fix hex-parsing #RGBA colours returning #RGRR based colours

ColdVolcano:
As an example, previously, attempting to parse `#1ABF` would return the value coresponding to `#1A11`

Not sure if this test-worthy


bdach:
Colour parsing already has incomplete test coverage, actually (see `ColourTest`). I will add appropriate test cases myself.

bdach (approved):

peppy (approved):

Issue #4112, opened 2020-12-27
Null Reference Exception in OsuJsonWebRequest.GetResponseString()


Neuheit:
From the latest commit in the osu! dev build, clicking on the multiplayer tab from startup results in the following NRE: 

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=osu.Framework
  StackTrace:
   at osu.Framework.IO.Network.WebRequest.GetResponseString()
```

This throws in line 157 in the `osu.Game.API.APIRequest` class, but I am creating the issue here as it is framework related.

bdach:
I cannot reproduce this neither on debug on release, and I'm not sure I understand why this would be a framework bug - and just saying there's a NRE does not help matter much. Please provide more details about your execution environment.

Neuheit:
OS: Windows 10 Build 19042
IDE: Visual Studio 2019 Build 16.8.3
.NET: Core 3.1

I figured this is a framework bug because a null ref was thrown from a framework method, which didn't seem intended to me.

peppy:
We already discussed this in discord right? Or are you a different user?

Either way, none of us can reproduce this so we need you to provide more details. I would recommend using a local framework reference if required (see https://github.com/ppy/osu-framework/wiki/Testing-local-framework-checkout-with-other-projects) and seeing what state the web request is in. If it has a null response, it also likely had an error along the way.

Neuheit:
> We already discussed this in discord right? Or are you a different user?

Yes I'm the same user. 

> Either way, none of us can reproduce this so we need you to provide more details. I would recommend using a local framework reference if required (see https://github.com/ppy/osu-framework/wiki/Testing-local-framework-checkout-with-other-projects) and seeing what state the web request is in. If it has a null response, it also likely had an error along the way.

Alright I'll take a look at that, thanks!

Firmatorenio:
For the sake of clearing up some confusion, there was another user whom this was discussed with, and _that_ was me.

peppy:
Closing this, looks to be behaving as intended.

PR #4115, opened 2020-12-29
Improve exception handling in WebRequest

peppy:
- No longer throw AggregateExceptions where we really don't need to.
- No longer attempt to deserialise a potentially null json response.

Helps to solve https://github.com/ppy/osu/issues/11347. I guess touches on the issue reported in https://github.com/ppy/osu-framework/issues/4112 too.

### DIFF
--- a/osu.Framework/IO/Network/WebRequest.cs
+++ b/osu.Framework/IO/Network/WebRequest.cs
@@ -10,6 +10,7 @@ using System.Net.Http.Headers;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using JetBrains.Annotations;
 using osu.Framework.Bindables;
 using osu.Framework.Extensions.ExceptionExtensions;
 using osu.Framework.Logging;
@@ -162,6 +163,7 @@ namespace osu.Framework.IO.Network
         /// Retrieve the full response body as a UTF8 encoded string.
         /// </summary>
         /// <returns>The response body.</returns>
+        [CanBeNull]
         public string GetResponseString()
         {
             try
--- a/osu.Framework/IO/Network/WebRequest.cs
+++ b/osu.Framework/IO/Network/WebRequest.cs
@@ -486,6 +486,8 @@ namespace osu.Framework.IO.Network
             else
                 logger.Add($@"Request to {Url} successfully completed!");
 
+            // if a failure happened on performing the request, there are still situations where we want to process the response.
+            // consider the case of a server returned error code which triggers a WebException, but the server is also returning details on the error in the response.
             try
             {
                 if (!wasTimeout)
@@ -493,8 +495,12 @@ namespace osu.Framework.IO.Network
             }
             catch (Exception se)
             {
-                logger.Add($"Processing response from {Url} failed with {se}.");
-                e = e == null ? se : new AggregateException(e, se);
+                // that said, we don't really care about an error when processing the response if there is already a higher level exception.
+                if (e != null)
+                {
+                    logger.Add($"Processing response from {Url} failed with {se}.");
+                    e = se;
+                }
             }
 
             if (e == null)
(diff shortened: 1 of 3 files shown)


osu.Framework/IO/Network/WebRequest.cs
```diff
@@ -486,15 +488,21 @@ protected virtual void Complete(Exception e = null)
...
             }
             catch (Exception se)
             {
-                logger.Add($"Processing response from {Url} failed with {se}.");
-                e = e == null ? se : new AggregateException(e, se);
+                // that said, we don't really care about an error when processing the response if there is already a higher level exception.
+                if (e != null)
+                {
+                    logger.Add($"Processing response from {Url} failed with {se}.");
+                    e = se;
```
bdach, line 504: Upon reflection, not sure about this one... Lower down this does `throw e;`, which means that the stack trace of `se` will get discarded by doing this (previously the aggregate was wrapping it; with this change, the exception will always appear to originate from this line). Maybe this should just be a

```suggestion
                    throw;
```

if we don't care what gets thrown anyway?

peppy: If you check back in history, we are only processing the response for a sepcific error handling case in osu-web. I really don't think we want aggregate exceptions to be fired for this because it just confuses the situation with no gain.

I'm not sure what you mean by the stack trace loss. Are you saying that by assigning and exception it is changing the stack trace somehow?

If this is really important then maybe changing the completion conditions to a `finally` block and using `throw` could work.

bdach: As far as I'm aware `throw` and `throw ex` are semantically different in a catch block ([reference SO post](https://stackoverflow.com/a/730255), [reference sharplab](https://sharplab.io/#v2:C4LglgNgPgAgTARgLACh4AZXwMwAIDequxu8cBRJVMCAnABQBEgPBuAg+4wJQDclxAvr1yCADgCcwANwCGwAKakEANlIAWXABVZAZ2ABBejXJSAxsDAB7AHYdBhFFSrBRAT0FU7Dz7hNmrAOgBJSwlzAGtZem43EgF7L2MZYwALXHoAUQAPY1lhX0tcWQybOM8PL0ck0XMAdwKMnhKHWKpmkhFxaTkFZRg1TR0AIQMEI1MLa1to4idXRvcphx9xwOCwiKi5mIWE4GTUzOzc8bri8rLy6cqahvLW/lQ+IA===)). From what I understand `throw` without the `ex` does some magic to capture the previous exception's stack trace and forward it, while `throw ex` just acts like a throw from its immediate position in code (potentially obfuscating where `ex` was originally coming from).

peppy: The exception we'd care about a location of is more likely the `e` not the `se`. I've change to a throw anyways.

osu.Framework/IO/Network/WebRequest.cs
```diff
@@ -486,15 +488,24 @@ protected virtual void Complete(Exception e = null)
...
             }
             catch (Exception se)
             {
-                logger.Add($"Processing response from {Url} failed with {se}.");
-                e = e == null ? se : new AggregateException(e, se);
+                // that said, we don't really care about an error when processing the response if there is already a higher level exception.
+                if (e != null)
+                {
+                    logger.Add($"Processing response from {Url} failed with {se}.");
+                    Failed?.Invoke(e);
```
bdach, line 504: I'm not sure this is right either.

* Previous code handled possibility of `e` being null but `se` not being so (with this code, if `e` was null, then `se` is silenced, and request is never completed).
* Logger logs `se`, API callback gets `e`, and `se` gets rethrown. Should probably all be one and the same?

peppy: should be `se` here

bdach (approved):
seems OK now

PR #4118, opened 2021-01-01
Fix slider bar crashing upon binding to disabled current

frenzibyte:
Another issue of https://github.com/ppy/osu-framework/issues/3218, also handles value changes from disabled current while at it, added test step to ensure `SliderBar` never crashes on disabled current again.

Also updates that outdated `instantaneous.Default = current.Default` line into a `current.DefaultChanged += e => instantaneous.Default = e.NewValue`.

- [ ] Depends on #4130

### DIFF
--- a/osu.Framework/Bindables/BindableNumberWithCurrent.cs
+++ b/osu.Framework/Bindables/BindableNumberWithCurrent.cs
@@ -23,7 +23,12 @@ namespace osu.Framework.Bindables
                 if (value == null)
                     throw new ArgumentNullException(nameof(value));
 
-                if (currentBound != null) UnbindFrom(currentBound);
+                if (currentBound != null)
+                    UnbindFrom(currentBound);
+
+                // must be re-enabled before binding below, see https://github.com/ppy/osu-framework/issues/3218.
+                Disabled = false;
+
                 BindTo(currentBound = (BindableNumber<T>)value);
             }
         }
--- a/osu.Framework/Graphics/UserInterface/SliderBar.cs
+++ b/osu.Framework/Graphics/UserInterface/SliderBar.cs
@@ -43,20 +43,10 @@ namespace osu.Framework.Graphics.UserInterface
 
         private readonly BindableNumberWithCurrent<T> current = new BindableNumberWithCurrent<T>();
 
-        protected BindableNumber<T> CurrentNumber => current;
-
         public Bindable<T> Current
         {
             get => current;
-            set
-            {
-                if (value == null)
-                    throw new ArgumentNullException(nameof(value));
-
-                current.Current = value;
-
-                currentNumberInstantaneous.Default = current.Default;
-            }
+            set => current.Current = value;
         }
 
         protected SliderBar()
@@ -64,6 +54,7 @@ namespace osu.Framework.Graphics.UserInterface
             currentNumberInstantaneous = new BindableNumber<T>();
 
             current.ValueChanged += e => currentNumberInstantaneous.Value = e.NewValue;
+            current.DefaultChanged += e => currentNumberInstantaneous.Default = e.NewValue;
             current.MinValueChanged += v => currentNumberInstantaneous.MinValue = v;
             current.MaxValueChanged += v => currentNumberInstantaneous.MaxValue = v;
             current.PrecisionChanged += v => currentNumberInstantaneous.Precision = v;
--- a/osu.Framework/Graphics/UserInterface/SliderBar.cs
+++ b/osu.Framework/Graphics/UserInterface/SliderBar.cs
@@ -53,7 +53,7 @@ namespace osu.Framework.Graphics.UserInterface
 
         protected SliderBar()
         {
-            currentNumberInstantaneous = new BindableNumber<T>();
+            currentNumberInstantaneous = new LeasedBindableNumber<T>();
 
             current.ValueChanged += e => currentNumberInstantaneous.Value = e.NewValue;
             current.DefaultChanged += e => currentNumberInstantaneous.Default = e.NewValue;
--- a/osu.Framework/Graphics/UserInterface/SliderBar.cs
+++ b/osu.Framework/Graphics/UserInterface/SliderBar.cs
@@ -43,6 +43,8 @@ namespace osu.Framework.Graphics.UserInterface
 
         private readonly BindableNumberWithCurrent<T> current = new BindableNumberWithCurrent<T>();
 
+        protected BindableNumber<T> CurrentNumber => current;
+
         public Bindable<T> Current
         {
             get => current;
--- a/osu.Framework/Graphics/UserInterface/SliderBar.cs
+++ b/osu.Framework/Graphics/UserInterface/SliderBar.cs
@@ -60,11 +60,11 @@ namespace osu.Framework.Graphics.UserInterface
             current.MinValueChanged += v => currentNumberInstantaneous.MinValue = v;
             current.MaxValueChanged += v => currentNumberInstantaneous.MaxValue = v;
             current.PrecisionChanged += v => currentNumberInstantaneous.Precision = v;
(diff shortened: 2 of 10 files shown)


bdach:
In the future, if you're submitting a fix to an issue without explaining how the fix works in the description but providing a test case instead, it would be nice if the commit that has the test case was first in the series, so I can check it out myself and step through the error without having to cherrypick the test onto `master`.

osu.Framework/Graphics/UserInterface/SliderBar.cs
```diff
@@ -160,7 +153,7 @@ protected override void OnDragEnd(DragEndEvent e)
 
         protected override bool OnKeyDown(KeyDownEvent e)
         {
-            if (currentNumberInstantaneous.Disabled)
+            if (current.Disabled)
```
bdach, line 156: Why are these changed? Again, `current` and `currentNumberInstantaneous` should either both be disabled or both enabled, but the former was not used once in this method, and now reads confusingly (`current` is checked for disable but `currentNumberInstantaneous` is actually the bindable that's being changed).

osu.Framework/Bindables/BindableNumberWithCurrent.cs
```diff
@@ -23,7 +23,12 @@ public Bindable<T> Current
                 if (value == null)
                     throw new ArgumentNullException(nameof(value));
 
-                if (currentBound != null) UnbindFrom(currentBound);
+                if (currentBound != null)
+                    UnbindFrom(currentBound);
+
+                // must be re-enabled before binding below, see https://github.com/ppy/osu-framework/issues/3218.
+                Disabled = false;
```
bdach, line 30: If we're doing this (which I'm not sure if we should be), `BindableWithCurrent` should probably get a similar thing.

bdach (changes requested):

osu.Framework/Bindables/BindableNumberWithCurrent.cs
frenzibyte: good catch, didn't know that existed.

osu.Framework/Graphics/UserInterface/SliderBar.cs
frenzibyte: `currentNumberInstantaneous` has to always be enabled to allow receiving value changes from disabled current without actually falling to a disabled exception, maybe providing an inline comment would help

frenzibyte:
> In the future, if you're submitting a fix to an issue without explaining how the fix works in the description but providing a test case instead, it would be nice if the commit that has the test case was first in the series, so I can check it out myself and step through the error without having to cherrypick the test onto `master`.

Gotcha.

osu.Framework/Graphics/UserInterface/SliderBar.cs
bdach: sounds like a lease might be what's desired here.

frenzibyte: thought about doing so but it seems that `LeasedBindableNumber<T>` is not a thing yet, guess I have to make one?

bdach: i guess, but this entire chain of changes is quickly spiraling out of control in regards to the number of actual changes required. this might take weeks to merge.

frenzibyte: yeah..., if there's any way to help in that, then I'll do

osu.Framework/Bindables/BindableNumberWithCurrent.cs
```diff
@@ -23,7 +23,12 @@ public Bindable<T> Current
                 if (value == null)
                     throw new ArgumentNullException(nameof(value));
 
-                if (currentBound != null) UnbindFrom(currentBound);
+                if (currentBound != null)
+                    UnbindFrom(currentBound);
+
+                // must be re-enabled before binding below, see https://github.com/ppy/osu-framework/issues/3218.
```
peppy, line 29: what happens after binding? is it re-disabled?

osu.Framework/Graphics/UserInterface/SliderBar.cs
```diff
@@ -48,30 +48,23 @@ public abstract class SliderBar<T> : Container, IHasCurrentValue<T>
...
         {
             get => current;
-            set
-            {
-                if (value == null)
-                    throw new ArgumentNullException(nameof(value));
-
-                current.Current = value;
-
-                currentNumberInstantaneous.Default = current.Default;
```
peppy, line 58: is there any test coverage checking this change works as expected?

osu.Framework/Graphics/UserInterface/SliderBar.cs
```diff
@@ -48,30 +48,23 @@ public abstract class SliderBar<T> : Container, IHasCurrentValue<T>
...
         {
             currentNumberInstantaneous = new BindableNumber<T>();
 
             current.ValueChanged += e => currentNumberInstantaneous.Value = e.NewValue;
+            current.DefaultChanged += e => currentNumberInstantaneous.Default = e.NewValue;
             current.MinValueChanged += v => currentNumberInstantaneous.MinValue = v;
             current.MaxValueChanged += v => currentNumberInstantaneous.MaxValue = v;
             current.PrecisionChanged += v => currentNumberInstantaneous.Precision = v;
-            current.DisabledChanged += v => currentNumberInstantaneous.Disabled = v;
+            // intentionally not updating disabled state to "currentNumberInstantaneous" to handle value changes from disabled current.
```
peppy, line 63: this explodes my head. needs rewriting

osu.Framework/Graphics/UserInterface/SliderBar.cs
```diff
@@ -202,7 +202,7 @@ private void handleMouseInput(UIEvent e)
         {
             var xPosition = ToLocalSpace(e.ScreenSpaceMousePosition).X - RangePadding;
 
-            if (currentNumberInstantaneous.Disabled)
+            if (current.Disabled)
```
peppy, line 205: so what's the reasoning behind this? `currentNumberInstantaneous` is no longer disabled ever?

peppy (changes requested):

peppy:
What crash does this fix? Where was the crash experienced? I'm having trouble following the path of documentation / reasoning here.

bdach:
Test case added in df011b2 crashes if cherry-picked onto `master`.

osu.Framework/Graphics/UserInterface/SliderBar.cs
frenzibyte: well, i've made it do like the others do `DefaultChanged += instantaneous.Default = value`, so thought no test coverage is really needed.

I believe that line was as-is because of no such thing as `DefaultChanged` back when this was implemented.

osu.Framework/Bindables/BindableNumberWithCurrent.cs
frenzibyte: yeah this is basically working around that https://github.com/ppy/osu-framework/issues/3218 issue, will see if I can properly fix the underlying issue or not (https://github.com/ppy/osu-framework/pull/4039 PR'd but tests broken)

osu.Framework/Graphics/UserInterface/SliderBar.cs
frenzibyte: will just remove altogether with a leased `currentNumberInstantaneous` instead.

osu.Framework/Graphics/UserInterface/SliderBar.cs
frenzibyte: will revert as well.

frenzibyte:
Closing this PR as the history got pretty dirty and there seems to be multiple different changes fixing separate things without clear explaination...

I'll look back into this and explain it more sanely.

PR #4130, opened 2021-01-01
Add proper support for leasing `BindableNumber<T>`

frenzibyte:
- [x] Depends on #3353 

I've tried my best to make this as shared as possible, but only a few lines did so. I was initially going with making each leased bindable class delegate their own bindable implementation through a fittable interface (`IBindable` for regular and `IBindableNumber` for number version).

But it appears that quite a lot of usages cast `LeasedBindable<T>` to `Bindable<T>` implicitly, and some of them are unavoidable and would need introducing a new `IMutableBindable` interface or otherwise.

Therefore I've gone with this way which introduces a new `LeasedBindableNumber<T>` inheriting `BindableNumber<T>` and overrides each property to set in leased context.


bdach:
At this point the "dependency" might as well not be listed. The one file that's in the dependency has not even remained intact here.

frenzibyte:
Yeah, you have a point, will take that into consideration for later.

osu.Framework.Tests/Bindables/BindableLeasingTest.cs
```diff
@@ -89,7 +89,7 @@ public void TestIncorrectEndLease()
 
             // end a lease with an incorrect bindable
             original.BeginLease(true);
-            Assert.Throws<InvalidOperationException>(() => original.EndLease(original));
+            Assert.Throws<InvalidOperationException>(() => original.EndLease(new Bindable<int>().BeginLease(true)));
```
bdach, line 92: very first line and i'm already confused... this is what happens if you mix changes like this. i had to think as to why this assertion was changed, and it was because the introduction of `ILeasedBindable` changed the signature of `EndLease()` to only accept `ILeasedBindable`s, so that improved compile-time safety, but muddled this diff. this is never explained anywhere.

please backport this change (and all changes to `ILeasedBindable` that are only in this PR) to the dependency, so we can merge that separately and deal with the rest of this diff separately. not reviewing further until that is done.

bdach (changes requested):

osu.Framework.Tests/Bindables/BindableLeasingTest.cs
frenzibyte: Sure.

peppy:
What is this change aiming to achieve? Is it being used somewhere or is it intended to be used in the future?

frenzibyte:
Yes, check https://github.com/ppy/osu-framework/pull/4118

peppy:
I don't get it. What does creating a `LeasedBindableNumber` mean in the first place? The concept of `LeasedBindable` is that it's something which is returned by another bindable, not instantiated directly.

frenzibyte:
> I don't get it. What does creating a `LeasedBindableNumber` mean in the first place? The concept of `LeasedBindable` is that it's something which is returned by another bindable, not instantiated directly.

Well, it gives the ability to change `MaxValue`, `MinValue`, `Precision` without them unsafely changing `Value` and cause an exception thrown. If using leased bindables for the sake of not caring about `Disabled` state a bad thing, then I'll look for another fix to the slider bar issue above.

frenzibyte:
Closing this as invalid, as per https://github.com/ppy/osu-framework/pull/4130#issuecomment-775666508.

frenzibyte:
Actually not closing as invalid, but rather, as useless, and can be returned back to if deemed required.

PR #4132, opened 2021-01-02
Improve reliability and performance of GL disposals

### LINKED ISSUE #4108: Random GL disposal-related crashes
Got booted out of game with this trace today:

```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at osu.Framework.Graphics.OpenGL.GLWrapper.checkPendingDisposals()
   at osu.Framework.Threading.ScheduledDelegate.RunTaskInternal()
   at osu.Framework.Threading.Scheduler.Update()
```

[Other people corroborate seeing similar](https://discord.com/channels/188630481301012481/188630652340404224/792065902375469086).

Seemingly most obvious candidate as to where this regressed is https://github.com/ppy/osu-framework/pull/4013. It's gotta be one of the accesses to `item`, I think - maybe the lock-free access to the item is racing with internal list operations and reading a `null`. Not even honestly sure how to confirm this, though.

bdach:
Resolves #4108.

To quote my comment there:

> Commit messages and inline comments should hopefully explain the main idea. From testing:
>
> * It passes the 100k drawable stress test given above.
> * I've benchmarked it on the test using dotTrace. Also tried adding a Benchmark.NET, but it's not trivial as it requires an actual game host and measuring that would probably be useless. It doesn't seem to be worse with the reworked concept, and it certainly seems to be better with the optimisation (times given are total times of the appropriate operations, measurement was once each time, over 2 minutes of running the stress test):
>
> | | current `master` | with `GLDisposalQueue` | with optimised `GLDisposalQueue` |
> | :- | -: | -: | -: |
> | scheduling disposals | 342ms | 311ms | 301ms |
> | checking disposals | 2267ms | 2179ms | 622ms |

I've also benchmarked against the game using dotTrace. Testing procedure was: release build, one-time run, from intro, through one map, to game exit - around 4 mins of operation overall each time. The results there are less impressive, but maybe that was to be expected, as the stress test case was a pretty degenerate one. Table for reference:

| | current `master` | with `GLDisposalQueue` | with optimised `GLDisposalQueue` |
| :- | -: | -: | -: |
| scheduling disposals | 19ms | 22ms | 9.3ms |
| checking disposals | 473ms | 744ms | 394ms |

That middle result might be an artifact of thermal throttling on my box, but I don't think it's all that relevant of a result as the optimisation in 5971ae2 seems like a safe optimisation to make (previous code was operating off of a similar assumption anyway - see the `Debug.Assert(i == 0)` before removal), and it's not like it's a super jarring difference - might just be normal-ish fluctuation.

### DIFF
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneStressGLDisposal.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneStressGLDisposal.cs
@@ -1,0 +1,47 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Linq;
+using NUnit.Framework;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Shapes;
+using osuTK;
+
+namespace osu.Framework.Tests.Visual.Drawables
+{
+    [Ignore("This test needs a game host to be useful.")]
+    public class TestSceneStressGLDisposal : FrameworkTestScene
+    {
+        private FillFlowContainer fillFlow;
+
+        protected override double TimePerAction => 0.001;
+
+        [SetUp]
+        public void SetUp() => Schedule(() =>
+        {
+            Add(fillFlow = new FillFlowContainer
+            {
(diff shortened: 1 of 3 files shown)


peppy (approved):
nice one

PR #4135, opened 2021-01-04
Enforce thread safety on common transform operations

peppy:
Until now, thread safety has been enforced only on child mutation (in `CompositeDrawable`). This extends the same checks to `Transformable`, covering common operations which can lead to game crashes from cross-thread enumeration and other weirdness.

I implemented this out of necessity after crashing twice from cross-thread enumeration in the transform stack and wanting to find the bad actor. It helped locate this and several other cases, which I will be fixing as follow-up pull requests (~2 osu! side, one harder one framework side in `VisibilityContainer`, only affecting custom implementations).

# Breaking changes

## Games will now throw (and crash) immediately on performing cross-thread transform operations

This may mean that as a framework consumer you need to re-think how you are performing operations between drawables. The easiest way to avoid issue is to use `Schedule` to force the target code to be on the correct thread.

### DIFF
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -2538,7 +2538,9 @@ namespace osu.Framework.Graphics
             return true;
         }
 
-        internal sealed override void EnsureMutationAllowed()
+        internal sealed override void EnsureTransformMutationAllowed() => EnsureMutationAllowed(nameof(Transforms));
+
+        internal void EnsureMutationAllowed(string member)
         {
             switch (LoadState)
             {
@@ -2547,20 +2549,20 @@ namespace osu.Framework.Graphics
 
                 case LoadState.Loading:
                     if (Thread.CurrentThread != LoadThread)
-                        throw new InvalidThreadForMutationException(LoadState, "not on the load thread");
+                        throw new InvalidThreadForMutationException(LoadState, member, "not on the load thread");
 
                     break;
 
                 case LoadState.Ready:
                     // Allow mutating from the load thread since parenting containers may still be in the loading state
                     if (Thread.CurrentThread != LoadThread && !ThreadSafety.IsUpdateThread)
-                        throw new InvalidThreadForMutationException(LoadState, "not on the load or update threads");
+                        throw new InvalidThreadForMutationException(LoadState, member, "not on the load or update threads");
 
                     break;
 
                 case LoadState.Loaded:
                     if (!ThreadSafety.IsUpdateThread)
-                        throw new InvalidThreadForMutationException(LoadState, "not on the update thread");
+                        throw new InvalidThreadForMutationException(LoadState, member, "not on the update thread");
 
                     break;
             }
@@ -2657,8 +2659,8 @@ namespace osu.Framework.Graphics
 
         public class InvalidThreadForMutationException : InvalidOperationException
         {
-            public InvalidThreadForMutationException(LoadState loadState, string description)
-                : base($"Cannot mutate the state of a {loadState} {nameof(Drawable)} while {description}. "
+            public InvalidThreadForMutationException(LoadState loadState, string member, string invalidThreadContextDescription)
+                : base($"Cannot mutate the {member} of a {loadState} {nameof(Drawable)} while {invalidThreadContextDescription}. "
                        + $"Consider using {nameof(Schedule)} to schedule the mutation operation.")
             {
             }
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -2540,6 +2540,11 @@ namespace osu.Framework.Graphics
 
         internal sealed override void EnsureTransformMutationAllowed() => EnsureMutationAllowed(nameof(Transforms));
 
+        /// <summary>
+        /// Check whether the current thread is valid for operating on thread-safe properties.
+        /// </summary>
+        /// <param name="member">The member to be operated on, used only for describing failures in exception messages.</param>
+        /// <exception cref="InvalidThreadForMutationException">If the current thread is not valid.</exception>
         internal void EnsureMutationAllowed(string member)
         {
             switch (LoadState)
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -2658,7 +2658,7 @@ namespace osu.Framework.Graphics
         public class InvalidThreadForMutationException : InvalidOperationException
         {
             public InvalidThreadForMutationException(LoadState loadState, string description)
-                : base($"Cannot mutate the children of a {loadState} {nameof(CompositeDrawable)} while {description}. "
+                : base($"Cannot mutate the state of a {loadState} {nameof(Drawable)} while {description}. "
                        + $"Consider using {nameof(Schedule)} to schedule the mutation operation.")
             {
             }
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -2538,6 +2538,34 @@ namespace osu.Framework.Graphics
             return true;
         }
 
+        internal sealed override void EnsureMutationAllowed()
+        {
+            switch (LoadState)
+            {
+                case LoadState.NotLoaded:
+                    break;
+
+                case LoadState.Loading:
+                    if (Thread.CurrentThread != LoadThread)
+                        throw new InvalidThreadForMutationException(LoadState, "not on the load thread");
+
+                    break;
+
+                case LoadState.Ready:
+                    // Allow mutating from the load thread since parenting containers may still be in the loading state
+                    if (Thread.CurrentThread != LoadThread && !ThreadSafety.IsUpdateThread)
+                        throw new InvalidThreadForMutationException(LoadState, "not on the load or update threads");
+
+                    break;
+
+                case LoadState.Loaded:
+                    if (!ThreadSafety.IsUpdateThread)
+                        throw new InvalidThreadForMutationException(LoadState, "not on the update thread");
+
+                    break;
+            }
+        }
+
         #endregion
 
         #region Transforms
@@ -2625,6 +2653,15 @@ namespace osu.Framework.Graphics
 
         private class EmptyDrawable : Drawable
         {
+        }
+
+        public class InvalidThreadForMutationException : InvalidOperationException
+        {
+            public InvalidThreadForMutationException(LoadState loadState, string description)
+                : base($"Cannot mutate the children of a {loadState} {nameof(CompositeDrawable)} while {description}. "
+                       + $"Consider using {nameof(Schedule)} to schedule the mutation operation.")
+            {
+            }
         }
     }
 
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -447,7 +447,7 @@ namespace osu.Framework.Graphics.Containers
         /// <returns>False if <paramref name="drawable"/> was not a child of this <see cref="CompositeDrawable"/> and true otherwise.</returns>
         protected internal virtual bool RemoveInternal(Drawable drawable)
         {
-            EnsureMutationAllowed();
+            EnsureChildMutationAllowed();
 
             if (drawable == null)
                 throw new ArgumentNullException(nameof(drawable));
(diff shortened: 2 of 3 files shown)


osu.Framework/Graphics/Drawable.cs
```diff
@@ -2626,6 +2654,15 @@ public override string ToString()
         private class EmptyDrawable : Drawable
         {
         }
+
+        public class InvalidThreadForMutationException : InvalidOperationException
+        {
+            public InvalidThreadForMutationException(LoadState loadState, string description)
+                : base($"Cannot mutate the children of a {loadState} {nameof(CompositeDrawable)} while {description}. "
+                       + $"Consider using {nameof(Schedule)} to schedule the mutation operation.")
```
bdach, line 2662: This message is not super great from a usability standpoint - it's too tightly coupled to the previous use case. We'll probably get folks asking why they're getting exceptions thrown in transform code even though they're not changing anything in the children.

bdach (changes requested):

bdach:
Overall not really sure of merging the `CompositeDrawable` helper into this use case. The following might be a micro-optimisation concern, but: I do know the JIT knows how to de-branch "throw helpers" (i.e. methods that always throw and never return), but I'm not sure how that will behave behind a virtual call. The helper being virtual might incur some overhead that on a large scale might turn out to not be all that negligible, and if it does, having separate throw helpers for transform and child mutation will resolve most of the current awkwardness pointed out in the above review.

peppy:
I considered but wrote off optimisation considerations since all the operations this was added to should be rare usages. If you think this is necessary to consider, then we either need to:

- Move `LoadThread` to `Transformable`
- Cast inside the helper method

If it's about optimisation, the second one gets ruled out. I was intentionally trying to avoid moving `LoadThread`, but what do you think? If it's `internal` maybe it doesn't matter?

osu.Framework/Graphics/Drawable.cs
peppy: I've reworded it for now, but as you mention, we may want to extract this logic out further.

bdach:
Hmm, right. It makes zero sense to have `LoadThread` in `Transformable`... probably fine to have the method virtual, then. Seeing that the override is `sealed`, only direct calls from `Transformable` should incur the penalty - any call from `Drawable` and inheritors should not use a `callvirt` due to the sealing.

For added style points - thoughts on the following patch?

```diff
diff --git a/osu.Framework/Graphics/Containers/CompositeDrawable.cs b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
index 4639b8a1c2..efc62af463 100644
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -447,7 +447,7 @@ protected internal int IndexOfInternal(Drawable drawable)
         /// <returns>False if <paramref name="drawable"/> was not a child of this <see cref="CompositeDrawable"/> and true otherwise.</returns>
         protected internal virtual bool RemoveInternal(Drawable drawable)
         {
-            EnsureMutationAllowed();
+            EnsureChildMutationAllowed();
 
             if (drawable == null)
                 throw new ArgumentNullException(nameof(drawable));
@@ -485,7 +485,7 @@ protected internal virtual bool RemoveInternal(Drawable drawable)
         /// </param>
         protected internal virtual void ClearInternal(bool disposeChildren = true)
         {
-            EnsureMutationAllowed();
+            EnsureChildMutationAllowed();
 
             if (internalChildren.Count == 0) return;
 
@@ -523,7 +523,7 @@ protected internal virtual void ClearInternal(bool disposeChildren = true)
         /// </summary>
         protected internal virtual void AddInternal(Drawable drawable)
         {
-            EnsureMutationAllowed();
+            EnsureChildMutationAllowed();
 
             if (IsDisposed)
                 throw new ObjectDisposedException(ToString(), "Disposed Drawables may not have children added.");
@@ -574,7 +574,7 @@ protected internal void AddRangeInternal(IEnumerable<Drawable> range)
         /// <param name="newDepth">The new depth value to be set.</param>
         protected internal void ChangeInternalChildDepth(Drawable child, float newDepth)
         {
-            EnsureMutationAllowed();
+            EnsureChildMutationAllowed();
 
             if (child.Depth == newDepth) return;
 
@@ -607,7 +607,7 @@ protected internal void ChangeInternalChildDepth(Drawable child, float newDepth)
         /// </remarks>
         protected internal void SortInternal()
         {
-            EnsureMutationAllowed();
+            EnsureChildMutationAllowed();
 
             internalChildren.Sort();
             aliveInternalChildren.Sort();
@@ -1181,7 +1181,7 @@ internal set
 
         public override void ApplyTransformsAt(double time, bool propagateChildren = false)
         {
-            EnsureMutationAllowed();
+            EnsureTransformMutationAllowed();
 
             base.ApplyTransformsAt(time, propagateChildren);
 
@@ -1194,7 +1194,7 @@ public override void ApplyTransformsAt(double time, bool propagateChildren = fal
 
         public override void ClearTransformsAfter(double time, bool propagateChildren = false, string targetMember = null)
         {
-            EnsureMutationAllowed();
+            EnsureTransformMutationAllowed();
 
             base.ClearTransformsAfter(time, propagateChildren, targetMember);
 
@@ -1223,7 +1223,7 @@ internal override void AddDelay(double duration, bool propagateChildren = false)
 
         public override IDisposable BeginAbsoluteSequence(double newTransformStartTime, bool recursive = true)
         {
-            EnsureMutationAllowed();
+            EnsureTransformMutationAllowed();
 
             var baseDisposalAction = base.BeginAbsoluteSequence(newTransformStartTime, recursive);
             if (!recursive)
@@ -1242,7 +1242,7 @@ public override IDisposable BeginAbsoluteSequence(double newTransformStartTime,
 
         public override void FinishTransforms(bool propagateChildren = false, string targetMember = null)
         {
-            EnsureMutationAllowed();
+            EnsureTransformMutationAllowed();
 
             base.FinishTransforms(propagateChildren, targetMember);
 
@@ -1279,6 +1279,8 @@ protected TransformSequence<CompositeDrawable> FadeEdgeEffectTo(Color4 newColour
         protected TransformSequence<CompositeDrawable> TweenEdgeEffectTo(EdgeEffectParameters newParams, double duration = 0, Easing easing = Easing.None) =>
             this.TransformTo(nameof(EdgeEffect), newParams, duration, easing);
 
+        internal void EnsureChildMutationAllowed() => EnsureMutationAllowed(nameof(InternalChildren));
+
         #endregion
 
         #region Interaction / Input
diff --git a/osu.Framework/Graphics/Drawable.cs b/osu.Framework/Graphics/Drawable.cs
index a00989da09..965af4431a 100644
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -2538,7 +2538,9 @@ internal virtual bool BuildPositionalInputQueue(Vector2 screenSpacePos, List<Dra
             return true;
         }
 
-        internal sealed override void EnsureMutationAllowed()
+        internal sealed override void EnsureTransformMutationAllowed() => EnsureMutationAllowed(nameof(Transforms));
+
+        internal void EnsureMutationAllowed(string member)
         {
             switch (LoadState)
             {
@@ -2547,20 +2549,20 @@ internal sealed override void EnsureMutationAllowed()
 
                 case LoadState.Loading:
                     if (Thread.CurrentThread != LoadThread)
-                        throw new InvalidThreadForMutationException(LoadState, "not on the load thread");
+                        throw new InvalidThreadForMutationException(LoadState, member, "not on the load thread");
 
                     break;
 
                 case LoadState.Ready:
                     // Allow mutating from the load thread since parenting containers may still be in the loading state
                     if (Thread.CurrentThread != LoadThread && !ThreadSafety.IsUpdateThread)
-                        throw new InvalidThreadForMutationException(LoadState, "not on the load or update threads");
+                        throw new InvalidThreadForMutationException(LoadState, member, "not on the load or update threads");
 
                     break;
 
                 case LoadState.Loaded:
                     if (!ThreadSafety.IsUpdateThread)
-                        throw new InvalidThreadForMutationException(LoadState, "not on the update thread");
+                        throw new InvalidThreadForMutationException(LoadState, member, "not on the update thread");
 
                     break;
             }
@@ -2657,8 +2659,8 @@ private class EmptyDrawable : Drawable
 
         public class InvalidThreadForMutationException : InvalidOperationException
         {
-            public InvalidThreadForMutationException(LoadState loadState, string description)
-                : base($"Cannot mutate the state of a {loadState} {nameof(Drawable)} while {description}. "
+            public InvalidThreadForMutationException(LoadState loadState, string member, string invalidThreadContextDescription)
+                : base($"Cannot mutate the {member} of a {loadState} {nameof(Drawable)} while {invalidThreadContextDescription}. "
                        + $"Consider using {nameof(Schedule)} to schedule the mutation operation.")
             {
             }
diff --git a/osu.Framework/Graphics/Transforms/Transformable.cs b/osu.Framework/Graphics/Transforms/Transformable.cs
index e14d0491fd..7c5deae5cf 100644
--- a/osu.Framework/Graphics/Transforms/Transformable.cs
+++ b/osu.Framework/Graphics/Transforms/Transformable.cs
@@ -138,7 +138,7 @@ private void updateTransforms(double time, bool forceRewindReprocess = false)
         /// <param name="toRemove">The <see cref="Transform"/> to remove.</param>
         public void RemoveTransform(Transform toRemove)
         {
-            EnsureMutationAllowed();
+            EnsureTransformMutationAllowed();
 
             getTrackerForGrouping(toRemove.TargetGrouping, false)?.RemoveTransform(toRemove);
 
@@ -155,7 +155,7 @@ public void RemoveTransform(Transform toRemove)
         /// </param>
         public virtual void ClearTransforms(bool propagateChildren = false, string targetMember = null)
         {
-            EnsureMutationAllowed();
+            EnsureTransformMutationAllowed();
 
             ClearTransformsAfter(double.NegativeInfinity, propagateChildren, targetMember);
         }
@@ -171,7 +171,7 @@ public virtual void ClearTransforms(bool propagateChildren = false, string targe
         /// </param>
         public virtual void ClearTransformsAfter(double time, bool propagateChildren = false, string targetMember = null)
         {
-            EnsureMutationAllowed();
+            EnsureTransformMutationAllowed();
 
             if (targetMember != null)
             {
@@ -195,7 +195,7 @@ public virtual void ClearTransformsAfter(double time, bool propagateChildren = f
         /// <param name="propagateChildren">Whether to also apply children's <see cref="Transform"/>s at <paramref name="time"/>.</param>
         public virtual void ApplyTransformsAt(double time, bool propagateChildren = false)
         {
-            EnsureMutationAllowed();
+            EnsureTransformMutationAllowed();
 
             if (RemoveCompletedTransforms) throw new InvalidOperationException($"Cannot arbitrarily apply transforms with {nameof(RemoveCompletedTransforms)} active.");
 
@@ -212,7 +212,7 @@ public virtual void ApplyTransformsAt(double time, bool propagateChildren = fals
         /// </param>
         public virtual void FinishTransforms(bool propagateChildren = false, string targetMember = null)
         {
-            EnsureMutationAllowed();
+            EnsureTransformMutationAllowed();
 
             if (targetMember != null)
             {
@@ -242,7 +242,7 @@ public virtual void FinishTransforms(bool propagateChildren = false, string targ
         /// <returns>An <see cref="InvokeOnDisposal"/> to be used in a using() statement.</returns>
         public IDisposable BeginDelayedSequence(double delay, bool recursive = true)
         {
-            EnsureMutationAllowed();
+            EnsureTransformMutationAllowed();
 
             if (delay == 0)
                 return null;
@@ -289,7 +289,7 @@ public DelayedSequenceSender(Transformable transformable, double delay, bool rec
         /// <exception cref="InvalidOperationException">Absolute sequences should never be nested inside another existing sequence.</exception>
         public virtual IDisposable BeginAbsoluteSequence(double newTransformStartTime, bool recursive = true)
         {
-            EnsureMutationAllowed();
+            EnsureTransformMutationAllowed();
 
             double oldTransformDelay = TransformDelay;
             double newTransformDelay = TransformDelay = newTransformStartTime - (Clock?.CurrentTime ?? 0);
@@ -333,7 +333,7 @@ public AbsoluteSequenceSender(Transformable transformable, double oldTransformDe
         /// <param name="customTransformID">When not null, the <see cref="Transform.TransformID"/> to assign for ordering.</param>
         public void AddTransform(Transform transform, ulong? customTransformID = null)
         {
-            EnsureMutationAllowed();
+            EnsureTransformMutationAllowed();
 
             if (transform == null)
                 throw new ArgumentNullException(nameof(transform));
@@ -372,6 +372,6 @@ public void AddTransform(Transform transform, ulong? customTransformID = null)
         /// Check whether this transformable is in a state where mutation of thread-safe properties is allowed.
         /// Will throw on failure.
         /// </summary>
-        internal abstract void EnsureMutationAllowed();
+        internal abstract void EnsureTransformMutationAllowed();
     }
 }
```

peppy:
Looks good.

I've made some minor changes to the xmldoc to better describe what these methods are actually doing (not sure why it mentioned state previously; it has nothing to do with the drawable's state).

bdach (approved):

PR #4136, opened 2021-01-06
Fix potential crash on shutdown due to thread race condition

peppy:
Null checks were missing from `GameThread.Thread` access. It can actually become null on a different thread due to execution completing normally.

I repro'd this on osu!-side, but unfortunately didn't save the log. Should be pretty straight-forward though.


bdach (approved):
seems correct

PR #4139, opened 2021-01-07
Fix playing back SampleChannels ending previous playbacks

peppy:
Reverts changes made in #3792. As mentioned in my comment there, cleanup of previous samples is a non-issue since that is handled by bass. The case that PR was fixing (hard to even know without reading the conversation thoroughly) looks to be looping samples, where `Play` is called twice and the previous channel ends up in a permanently playing state. I've added a case to fix that specifically, while allowing standard flow otherwise.

There is an expectation everywhere in the framework that sample playback is limited to provided `SAMPLE_CONCURRENCY`. I believe this expectation should carry through for a single instance of `SampleChannel`, where a new playback should be initialised for each call to play.

This has come up multiple times, but the class `SampleChannel` is named quite badly, since it tries to represent what BASS refers to as a channel, but internally fetches a new channel each playback. We'll probably have to address this at some point. Open to feedback (maybe we can obsolete the `Play` method and add a `bool` specific to whether a new channel should be used or not).

Apart from fixing concurrency, this also fixes weird clipping artifacts when a previous channel is stopped and immediately restarted.


osu.Framework/Audio/Sample/ISampleChannel.cs
```diff
@@ -9,13 +11,14 @@ namespace osu.Framework.Audio.Sample
     public interface ISampleChannel : IHasAmplitudes
     {
         /// <summary>
-        /// Start playback.
+        /// Start a new playback of this sample.
+        /// Note that this will not stop previous playbacks unless <see cref="Looping"/> is true (but concurrency will be limited by the source <see cref="ISampleStore.PlaybackConcurrency"/>.
```
bdach, line 15: This has me confused. If I'm reading correctly, `SampleChannelBass`'s implementation does not have these semantics. If the sample is already playing, it will not start a new playback of the sample at all, but preserve the old playback, correct? (see early-return if `currentChannelPlaying && !restart`)

Is that an oversight with the implementation, or is this xmldoc inaccurate?

bdach (changes requested):

bdach:
If I recall correctly what the original PR intended to fix was looping samples being allowed to play out until the end of the sample if they were stopped mid-loop. This was apparent in the pause overlay; having checked this PR with a framework checkout it looks that case has not regressed with this change (as I expected from reading the diff).

osu.Framework/Audio/Sample/ISampleChannel.cs
peppy: The coded behaviour is correct. I'm open to changes to the xmldoc if you think you can make it read better.

For the record, we use the `restart` parameter nowhere. Removing would be another option. It's actually pretty useless because for it to work, we also need to expose `Bass_ChannelPause`. The `Stop()` method we currently have *always* resets position.

osu.Framework/Audio/Sample/ISampleChannel.cs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using osu.Framework.Audio.Track;
```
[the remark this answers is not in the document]
peppy: Seems to be `ISampleStore`. Will move it, but I guess it's a minor breaking change for consumers.

osu.Framework/Audio/Sample/ISampleChannel.cs
bdach: Yeah I don't know if there actually is a way to describe the current behaviour elegantly. It's all weird, because a `Play()` will play the sample, unless it's already playing, in which case it won't do anything and ignore `restart`, unless it's looping, in which case it'll *continue* playing but take `restart` into consideration. Super wonky.

I'm actually not sure if that special weird looping case is even relevant anymore now that we have `ISamplePlaybackDisabler` game-side. In fact I'm going to check that right now. Maybe we can just nuke it and do the sane thing of "play if not playing, restart if restart requested". Maybe `Restart()` should be a separate method even.

peppy: I don't think your description here is totally correct: in the case `Play()` is called and the sample is already playing, it *will* play a second time (and doesn't ignore `restart` - it takes it into account and avoids restarting the sample if set to `false`).

bdach: Oh never mind, looks like I misread that conditional in there, you're right.

osu.Framework/Audio/Sample/SampleChannelBass.cs
```diff
@@ -70,10 +70,18 @@ public override void Play(bool restart = true)
...
-                // Free previous channels as we're creating a new channel for every playback, since old channels
-                // may be overriden when too many other channels are created from the same sample.
-                if (Bass.ChannelIsActive(channel) != PlaybackState.Stopped)
-                    Bass.ChannelStop(channel);
+                bool currentChannelPlaying = Bass.ChannelIsActive(channel) == PlaybackState.Playing;
+
+                if (currentChannelPlaying)
+                {
+                    // if restart is not requested and the sample is currently playing, nothing needs to be done.
+                    if (!restart)
```
smoogipoo, line 78: So if I understand this code correctly, if restarting and not looping then samples will be layered? I don't think a single `SampleChannel` should allow layering itself - a hitcircle playing the same hit `SampleChannel` twice in a row should not play two samples.

peppy: Okay, so how would you expect the flow to work for the 99% case, where you expect concurrency as exposed by the sample manager / game?

smoogipoo: So, with hover effects as the main example? I'd say you need a new sample channel at that point.

peppy: Which means a complete restructure and not using bass concurrency management.

smoogipoo (approved):
Lgtm, requesting a review from @bdach to see if he agrees with the direction.

bdach:
Just as a heads up: I don't have an issue with this diff but before I approve I want to suggest improvements to #4148, just to make sure all game-side use cases are taken care of properly. I have queued that up for this weekend.

peppy:
@bdach do your changes involve changing this PR? if not i think it's fine to get this in on its own (it fixes a separate issue).

bdach:
I'm not sure what it is but this diff confuses me greatly every time I read it - but I would agree that this is independent from the pull I mentioned and it *seems* like it should be correct. It would be good to have this tested, but that seems infeasible, so I guess I'll just approve and merge for now...

bdach (approved):

PR #4140, opened 2021-01-08
Fix track not considered complete when reaching beyond length

frenzibyte:
This fixes a lazer-side issue where tracks sometimes do not loop when they end even though they should.

### Reason

When a track is played to finish, the position could end up **beyond** the track length, and [with the current `isRunning` update conditions](https://github.com/ppy/osu-framework/blob/9e0a66b17cef545102c2672361ed3fa75319fcf5/osu.Framework/Audio/Track/TrackBass.cs#L194), the track loop logic below will never be executed:
https://github.com/ppy/osu-framework/blob/0a14a592d25399ffd4ac69768227761f11ce055f/osu.Framework/Audio/AudioComponent.cs#L111-L112 https://github.com/ppy/osu-framework/blob/d394011eeaddec8aa523424a5bb2c4a10d56c535/osu.Framework/Audio/Track/Track.cs#L24

Upon blaming, this commit https://github.com/ppy/osu-framework/pull/3025/commits/1c7df503908d623fbcc119fce154e7134385fd91 was the one introducing the conditions, and there didn't seem to be any reason on why it's exactly `position != length`.

But I've removed the whole condition and make `hasCompleted` set to true whenever track is ended.

This is not testable currently as I can't find any path for getting the track to play beyond length yet...


peppy:
Is this referenced in BASS documentation? Or are you guessing here?

frenzibyte:
Good point, I actually haven't checked which value is at wrong here, but upon doing so, in the documentation of `BASS_ChannelGetLength(...)`, it seems that the byte length [may not be 100% accurate](http://www.bass.radio42.com/help/html/fef81c45-ca57-1396-0f87-52b777107e28.htm), but it should be exact on MP3 files when using `BASS_STREAM_PRESCAN`, which we are using...

Will investigate further in a bit.

frenzibyte:
Seems the length we're getting from `BASS.ChannelGetLength(...)` is actually near correct checking with audacity:
image
image

But, on checking with `ffprobe`, the mp3 length is shown to be higher than the above:
```
salman@Salmans-MacBook-Pro files % ffprobe -v error -show_entries format=duration -of default=noprint_wrappers=1:nokey=1 ./2/22/2231a6d635ada1c165499e66a17f9305951697a9ef6e41917968c0bcaf26518d
41.377959
```

And leaving the track not restart on loop and keeping it until it stops by itself stops at a position between the known length, which is 43.343sec, and the length returned by `ffprobe` (actual data length?), which is 43.377sec.

Not sure whether that should be reported to BASS as a bug for continuing past 43.343sec (the length returned by BASS and audacity as well), what do you think?

bdach:
Is this a theoretical "issue", or can this be hit in some way from the game? I'm asking so that I can reproduce and investigate myself if it's the latter.

frenzibyte:
This is something you can hit in game, as in the description:
> This fixes a lazer-side issue where tracks sometimes do not loop when they end even though they should.

So basically keep playing track at different positions and at some point the track will stop and not loop at the end.

peppy:
If you're comparing to some external source, why are you rounding arbitrarily to milliseconds?

Also I'm not sure this fix it going to work; what happens if the length is wrong in the other direction?

frenzibyte:
> Sounds like we shouldn't be using this method of telling if the tracked got to the end in the first place?

Yeah, I thought about using the track end callback instead but wasn't sure if that's how we should be treating things (`CurrentTime >= Length` doesn't necessarily mean track completed), but seems more correct anyway as we have a `HasCompleted` boolean to use, will clarify in the `Length` property that it doesn't exactly mean the track's length, I'll have a go on it now.

osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -293,6 +298,8 @@ public async Task<bool> SeekAsync(double seek)
 
                 if (pos != Bass.ChannelGetPosition(activeStream))
                     Bass.ChannelSetPosition(activeStream, pos);
+
+                hasCompleted = false;
```
frenzibyte, line 302: Not sure about this, there isn't any "sync" indicating that a stream has been rewound back to be played again, so this is the only way possible for now.

bdach:
Test failure seems to be directly caused by these changes. I can reproduce hangs/deadlocks on `TestLoopingTrackDoesntSetCompleted`. Whatever that test checks, that seems to be terminally broken.

@frenzibyte Did you even *attempt* to run audio tests before pushing this?

frenzibyte:
One small change amended ate my life apparently, the deadlock was happening because of the track's end sync callback invoking `Restart()` synchronously (enqueue to audio thread and wait), but the audio thread only gets updated manually by the test itself, causing the deadlock.

Not sure what the best course of action is, but I guess making the `Restart()` invocation happen asynchronously from itself (enqueue to audio thread and continue) is the way to go for now:
```diff
                    stopCallback = new SyncCallback((a, b, c, d) => RaiseFailed());
                    endCallback = new SyncCallback((a, b, c, d) =>
                    {
                        if (Looping)
-                            Restart();
+                            EnqueueAction(() => Restart());
                        else
                        {
                            hasCompleted = true;
                            RaiseCompleted();
                        }
                    });
```

bdach:
I'm pretty sure sync callbacks don't run on the audio thread. They run on a native BASS thread. So doing operations with that usually should be done on the audio thread on the callback seems fundamentally unsafe, and yes, enqueueing that seems like the logical thing to do.

Although given that the restart now won't proceed immediately it should be ensured that there is no audio glitching caused by this change (i.e. the track looping on its own between the sync callback executing and the enqueued restart executing).

frenzibyte:
Track looping working properly lazer-side, so things should be fine if I understand you correctly.

peppy:
On an initial test, it looks like the bass loop flag can be applied to any channel, including track playback (http://www.bass.radio42.com/help/html/9781ee5b-e613-8409-0c73-8dcc6505083a.htm). This logic is already implemented in `SampleChannelBass` and should work much better than manually triggering a restart - as @bdach mentioned doing it manually could cause a gap due to thread synchronisation.

I've applied this change, please test and confirm everything looks as you'd expect.

frenzibyte:
Right, seems fine.

osu.Framework/Audio/Track/TrackBass.cs
bdach: `ChannelGetPosition` should be pretty accurate, right? Maybe this set to `false` should be under a `newPosition < previousPosition` check or similar? Although that may be wrong when negative rate comes into play... Maybe it should be `Math.Sign(newPosition - previousPosition) != direction` or similar.

Generally speaking I don't like this very much. Worried about misuse and consumers seeking way past end and accidentally clearing this flag in the process (and therefore potentially indirectly influencing `IsRunning` as well, as that property does some local hackery with bypassing what BASS returns).

frenzibyte:
@peppy Actually, your changes regressed looping behaviour, now it always loops back to `0`, instead of `RestartPoint`, can't find any way to make BASS's built-in looping actually loop back to a specific point.

Enqueuing the restart to audio thread at least will not change behaviour of how track looping used to be, and on testing with osu!lazer I see no regressions.

peppy:
Adding back custom logic when `RestartPoint` is non-zero sounds like a good solution. I'd still rather have/use the bass-level looping when we can.

Note that the track `endCallback` is still fired even with the bass looping flag, so adding a seek operation in there may be enough to fix said behaviour.

frenzibyte:
Yeah actually `BASS` already has a sync type (`BASS_SYNC_MIXTIME`) that is invoked immediately when it occurs, not when the sync is heard, and it can be OR'd with `BASS_SYNC_END` to perform the seek:
http://www.un4seen.com/doc/#bass/BASS_ChannelSetSync.html
image

Not sure about making `Completed` event raised on "mixtime" as well, will separate this type of sync to a separate callback.

peppy (approved):
This looks pretty solid.

Issue #4141, opened 2021-01-07
Crashes caused by MIDI callbacks on garbage collected delegates


sirloinsteak11:
After trying to open osu!lazer, nothing happens at all. 

This is what I got from the event viewer: https://imgur.com/c3rvu8n 

osu!lazer version: 2020.1229.0

Logs:
[session.log](https://github.com/ppy/osu/files/5784253/session.log)
[update_success.log](https://github.com/ppy/osu/files/5784254/update_success.log)
[runtime.log](https://github.com/ppy/osu/files/5784255/runtime.log)

Computer Specifications: 
CPU: Intel Core i3-6100T CPU @3.20GHz
GPU: AMD Radeon R9 M470X
8 GB RAM
2024 MB VRAM
Windows 10

frenzibyte:
After investigating a bit, this may be a result of not closing input on disabling / disposing input handler, which can be the case if you have "raw input" on, as that modifies the `ignoredInputHandlers` bindable list and lead to disposing previous handlers (including midi handler) and re-creating new ones, I'll PR a fix for that now.

peppy:
Tentatively closing as I believe the merged PR should have fixed this. Will be live in the osu! release, @sirloinsteak11 would be great if you can confirm that that point.

PR #4142, opened 2021-01-08
Close MIDI devices when disposing of / disabling MIDI input handler

frenzibyte:
May (resolve) #4141 

Same to what https://github.com/ppy/osu-framework/pull/4037 does, but this is in the case `MidiInputHandler` is disabled / [disposed of](https://github.com/ppy/osu-framework/blob/9901b9648ef316d62afaeb05cff5b129ea1922d5/osu.Framework/Input/Handlers/InputHandler.cs#L64).

Also, I don't have a MIDI device to test this, so I would appreciate someone testing this for me.

### DIFF
--- a/osu.Framework/Input/Handlers/Midi/MidiInputHandler.cs
+++ b/osu.Framework/Input/Handlers/Midi/MidiInputHandler.cs
@@ -44,6 +44,7 @@ namespace osu.Framework.Input.Handlers.Midi
 
                     foreach (var value in openedDevices.Values)
                     {
+                        value.CloseAsync().Wait();
                         value.MessageReceived -= onMidiMessageReceived;
                     }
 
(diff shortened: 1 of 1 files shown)


osu.Framework/Input/Handlers/Midi/MidiInputHandler.cs
```diff
@@ -44,6 +44,7 @@ public override bool Initialize(GameHost host)
 
                     foreach (var value in openedDevices.Values)
                     {
+                        value.CloseAsync().Wait();
                         value.MessageReceived -= onMidiMessageReceived;
```
peppy, line 48: move to a `closeDevice` method to avoid the same place being in two places.

peppy (changes requested):

peppy (approved):

peppy (approved):

PR #4143, opened 2021-01-11
Fix `GetLoadableTypes()` extension potentially including null `Type` instances

frenzibyte:
As [stated in the docs](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.reflectiontypeloadexception.types?view=net-5.0#property-value), `ReflectionTypeLoadException.`could potentially include null types instances in the array, causing `TestBrowser` to eat a null reference exception as it uses it:
https://github.com/ppy/osu-framework/blob/99a640861a45a4e9f5c2e32c8dffb483c5ecf457/osu.Framework/Testing/TestBrowser.cs#L72 https://github.com/ppy/osu-framework/blob/99a640861a45a4e9f5c2e32c8dffb483c5ecf457/osu.Framework/Testing/TestBrowser.cs#L87

This fixes test browser no longer working for me due to the above.

---

Also not entirely sure how to handle that `ConditionIsAlwaysTrueOrFalse` resharper warning, the array is declared as `Type[]?`, which, if I understand correctly, means it's either a null array, or an array with non-null type elements. But the array indeed had a null `Type` element sticking inside...

### DIFF
--- a/osu.Framework/Extensions/ExtensionMethods.cs
+++ b/osu.Framework/Extensions/ExtensionMethods.cs
@@ -162,7 +162,9 @@ namespace osu.Framework.Extensions
             }
             catch (ReflectionTypeLoadException e)
             {
-                return e.Types ?? Enumerable.Empty<Type>();
+                // ReSharper disable once ConditionIsAlwaysTrueOrFalse (this may contain null types, as stated in docs.)
+                // https://docs.microsoft.com/en-us/dotnet/api/system.reflection.reflectiontypeloadexception.types?view=net-5.0#property-value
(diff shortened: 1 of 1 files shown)


peppy:
For reference, how did you trigger this in the first place?

frenzibyte:
Had tau ruleset installed, while the test browser was fetching types through assemblies, it attempted to do so on that ruleset's assembly, and for some reason, the `ReflectionTypeLoadException` passed an array list with a `null` type inserted.

I'm guessing it's because I had an old unloadable version of that ruleset.

peppy (approved):
This is really weird (why wouldn't they mark it as `Type?[]`, but the documentation checks out.

PR #4144, opened 2021-01-12
Add IKeyBinding interface type for flexibility in key bindings

peppy:
This is a prerequisite for my realm `KeyBindingStore` implementation.

When working with realm, it is beneficial to have interfacing defining the underlying model types due to the restrictions on inheritance. I think in general having interfaces for model types like this is a good step forward, regardless.

Probably resulting in a breaking change for rulesets, but not sure that is avoidable (or an issue at this point in development).

### DIFF
--- a/osu.Framework/Input/Bindings/KeyBindingContainer.cs
+++ b/osu.Framework/Input/Bindings/KeyBindingContainer.cs
@@ -37,7 +37,7 @@ namespace osu.Framework.Input.Bindings
             this.matchingMode = matchingMode;
         }
 
-        private readonly List<KeyBinding> pressedBindings = new List<KeyBinding>();
+        private readonly List<IKeyBinding> pressedBindings = new List<IKeyBinding>();
 
         private readonly List<T> pressedActions = new List<T>();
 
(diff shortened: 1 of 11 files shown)


frenzibyte:
Could make `PlatformKeyBindings` also use `IKeyBinding` type as well while at it.

peppy:
Not sure I can see a reason for doing so. Can you imagine a platform which is specifying key bindings which aren't of a known type? Not really sure such a change should be made without reason.

frenzibyte:
Was thinking about it from a consistency point of view, but I guess that's a valid reason not to do it, nevermind.

osu.Framework/Input/Bindings/KeyBindingContainer.cs
```diff
@@ -168,7 +168,7 @@ private bool handleNewPressed(InputState state, InputKey newKey, bool repeat, Ve
             {
                 // if the current key pressed was a modifier, only handle modifier-only bindings.
                 // lambda expression is used so that the delegate is cached (see: https://github.com/dotnet/roslyn/issues/5835)
-                newlyPressed = newlyPressed.Where(b => b.KeyCombination.Keys.All(key => KeyCombination.IsModifierKey(key)));
+                newlyPressed = newlyPressed.Where(b => b.KeyCombination.Keys.All(KeyCombination.IsModifierKey));
```
bdach, line 171: won't this change to a method group cause per-call allocations, as per the preceding comment? may be the case that we don't care, but if we don't, then the comment should be trimmed.

bdach (changes requested):

osu.Framework/Input/Bindings/KeyBindingContainer.cs
peppy: Good catch. Not sure why this got changed, I think somewhere along the way it became a warning somehow..

bdach (approved):

smoogipoo (approved):

PR #4145, opened 2021-01-12
Run initial ReloadMappings call during async load

peppy:
Fixes a case of potentially querying a data source in LoadComplete causing an unknown overhead. No real reason not to run this initial call at the end of the async load process. Note that this can't go into the `load()` method directly as overridden implementations of this may depend on dependencies that are not yet available.


bdach (approved):
seems fair

peppy:
This actually.. might work against me with my latest round of changed. Closing tentatively.

PR #4148, opened 2021-01-14
Add a method of retrieving sample channels which can layer multiple playbacks

peppy:
Currently retrieving a sample channel and attempting to play it back will stop previous playback. This isn't always what we want. This adds a new flow for retrieving samples which will automatically layer, up to playback concurrency limits.

- [x] Depends on #4139

### DIFF
--- a/osu.Framework/Audio/Sample/SampleBass.cs
+++ b/osu.Framework/Audio/Sample/SampleBass.cs
@@ -12,7 +12,7 @@ namespace osu.Framework.Audio.Sample
 {
     internal sealed class SampleBass : Sample, IBassAudio
     {
-        internal volatile int SampleId;
+        internal int SampleId { get; private set; }
 
         public override bool IsLoaded => SampleId != 0;
 
@@ -39,6 +39,8 @@ namespace osu.Framework.Audio.Sample
             if (IsLoaded)
             {
                 Bass.SampleFree(SampleId);
+                SampleId = 0;
+
                 memoryLease?.Dispose();
             }
 
--- a/osu.Framework/Audio/Sample/SampleBass.cs
+++ b/osu.Framework/Audio/Sample/SampleBass.cs
@@ -12,9 +12,9 @@ namespace osu.Framework.Audio.Sample
 {
     internal sealed class SampleBass : Sample, IBassAudio
     {
-        private volatile int sampleId;
+        internal volatile int SampleId;
 
-        public override bool IsLoaded => sampleId != 0;
+        public override bool IsLoaded => SampleId != 0;
 
         private NativeMemoryTracker.NativeMemoryLease memoryLease;
 
@@ -28,7 +28,7 @@ namespace osu.Framework.Audio.Sample
             {
                 EnqueueAction(() =>
                 {
-                    sampleId = loadSample(data);
+                    SampleId = loadSample(data);
                     memoryLease = NativeMemoryTracker.AddMemory(this, data.Length);
                 });
             }
@@ -38,7 +38,7 @@ namespace osu.Framework.Audio.Sample
         {
             if (IsLoaded)
             {
-                Bass.SampleFree(sampleId);
+                Bass.SampleFree(SampleId);
                 memoryLease?.Dispose();
             }
 
@@ -51,11 +51,11 @@ namespace osu.Framework.Audio.Sample
                 return;
 
             // counter-intuitively, this is the correct API to use to migrate a sample to a new device.
-            Bass.ChannelSetDevice(sampleId, deviceIndex);
+            Bass.ChannelSetDevice(SampleId, deviceIndex);
             BassUtils.CheckFaulted(true);
         }
 
-        public int CreateChannel() => Bass.SampleGetChannel(sampleId);
+        public int CreateChannel() => Bass.SampleGetChannel(SampleId);
 
         private int loadSample(byte[] data)
         {
--- a/osu.Framework/Audio/Sample/SampleChannel.cs
+++ b/osu.Framework/Audio/Sample/SampleChannel.cs
@@ -11,6 +11,11 @@ namespace osu.Framework.Audio.Sample
     {
         protected bool WasStarted;
 
+        /// <summary>
+        /// Whether calling <see cref="Play"/> should forcefully stop an existing playback of this channel.
+        /// </summary>
+        internal bool PlayStopsPreviousPlayback;
+
         protected Sample Sample { get; set; }
 
         private readonly Action<SampleChannel> onPlay;
@@ -23,6 +28,9 @@ namespace osu.Framework.Audio.Sample
 
         public virtual void Play(bool restart = true)
         {
+            if (PlayStopsPreviousPlayback && Looping)
+                throw new InvalidOperationException($"Cannot play a layered sample playback if {nameof(Looping)} is enabled.");
+
             if (IsDisposed)
                 throw new ObjectDisposedException(ToString(), "Can not play disposed samples.");
 
@@ -32,6 +40,9 @@ namespace osu.Framework.Audio.Sample
 
         public virtual void Stop()
         {
+            if (PlayStopsPreviousPlayback)
+                throw new InvalidOperationException($"Cannot call {nameof(Stop)} on a layered sample playback.");
+
             if (IsDisposed)
                 throw new ObjectDisposedException(ToString(), "Can not stop disposed samples.");
         }
--- a/osu.Framework/Audio/Sample/SampleChannel.cs
+++ b/osu.Framework/Audio/Sample/SampleChannel.cs
@@ -13,6 +13,7 @@ namespace osu.Framework.Audio.Sample
 
         /// <summary>
         /// Whether calling <see cref="Play"/> should forcefully stop an existing playback of this channel.
+        /// Setting this to false will result in "layered" playback, with each call to Play triggering a new overlapping playback.
         /// </summary>
         internal bool PlayStopsPreviousPlayback = true;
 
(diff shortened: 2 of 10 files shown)


osu.Framework/Audio/Sample/SampleChannelBass.cs
```diff
@@ -78,7 +78,8 @@ public override void Play(bool restart = true)
                     if (!restart)
                         return;
 
-                    Stop();
+                    if (PlayStopsPreviousPlayback)
```
smoogipoo, line 81: I think this should wrap the entire thing, such that if `playing && !restart` then it also plays a new channel.

osu.Framework/Audio/Sample/SampleChannel.cs
```diff
@@ -11,6 +11,11 @@ public abstract class SampleChannel : AdjustableAudioComponent, ISampleChannel
     {
         protected bool WasStarted;
 
+        /// <summary>
+        /// Whether calling <see cref="Play"/> should forcefully stop an existing playback of this channel.
+        /// </summary>
+        internal bool PlayStopsPreviousPlayback;
```
smoogipoo, line 17: Name/value is inversed, which results in the conditions also appearing inversed.

osu.Framework/Audio/Sample/SampleChannel.cs
```diff
@@ -23,6 +28,9 @@ protected SampleChannel(Sample sample, Action<SampleChannel> onPlay)
 
         public virtual void Play(bool restart = true)
         {
+            if (PlayStopsPreviousPlayback && Looping)
```
smoogipoo, line 31: I'd add this check to `Looping.set` instead and seal the override.

smoogipoo (changes requested):

osu.Framework/Audio/Sample/SampleChannelBass.cs
peppy: throw instead maybe?

osu.Framework/Audio/Sample/SampleChannel.cs
peppy: I'll leave it here as well, just as a safety measure.

osu.Framework/Audio/Sample/SampleChannelBass.cs
smoogipoo: if `restart = false`? Yeah I think that's probably fine since it's not the default parameter.

osu.Framework/Audio/Sample/SampleChannel.cs
```diff
@@ -23,6 +29,15 @@ protected SampleChannel(Sample sample, Action<SampleChannel> onPlay)
 
         public virtual void Play(bool restart = true)
         {
+            if (!PlayStopsPreviousPlayback)
```
smoogipoo, line 32: The inverted condition here in combination with the field name is particularly hard to read.

I suggest renaming the field to:
```csharp
/// <summary>
/// Whether each invocation of <see cref="Play"/> results in a new (potentially concurrent) playback.
/// </summary>
internal bool ConcurrentPlayback;
```
And inverting conditions like this.

osu.Framework/Audio/Sample/SampleChannel.cs
```diff
@@ -32,13 +47,16 @@ public virtual void Play(bool restart = true)
...
+                throw new InvalidOperationException($"Cannot call {nameof(Stop)} on a layered sample playback.");
+
             if (IsDisposed)
                 throw new ObjectDisposedException(ToString(), "Can not stop disposed samples.");
         }
 
         protected override void Dispose(bool disposing)
         {
-            if (!IsDisposed)
+            if (!IsDisposed && PlayStopsPreviousPlayback)
```
smoogipoo, line 59: This is very unfortunate...

osu.Framework/Audio/Sample/SampleBass.cs
```diff
@@ -12,9 +12,9 @@ namespace osu.Framework.Audio.Sample
 {
     internal sealed class SampleBass : Sample, IBassAudio
     {
-        private volatile int sampleId;
+        internal volatile int SampleId;
```
smoogipoo, line 15:
```suggestion
        internal int SampleId { get; private set; }
```

smoogipoo (changes requested):

osu.Framework/Audio/Sample/SampleChannel.cs
peppy: well it's not going to work as expected either way. you'd hope it stopped *all* playbacks in a layered context, right? not sure we want to go that deep though.

if we just want to avoid this conditional, could make a protected stop implementation and bypass the exceptions.

osu.Framework/Audio/Sample/SampleChannel.cs
peppy: Yep, got halfway through doing this, will change.

osu.Framework/Audio/Sample/SampleBass.cs
peppy: are we okay with dropping the volatile here?

smoogipoo: It shouldn't make a difference because the only external multi-threaded use-case is via `IsLoaded` which shouldn't care about a read being reordered before the first write, otherwise it would have other problems like trying to make sure the audio thread is running completely synchronously. Means at the worst case it'll read `IsLoaded = false` slightly before `IsLoaded = true`.

Something to note which I also just noticed, we're not setting this to 0 in `Dispose()`. Probably want to do that so `IsLoaded` becomes `false` if disposed...

peppy: Have applied both, seems logical.

smoogipoo (approved):

osu.Framework/Audio/Sample/SampleChannel.cs
smoogipoo: Correct, I'd expect it to stop all playbacks in a layered context, but as you said that's quite a large level of complexity. I think the use of layered playbacks is small enough that this shouldn't be much of an issue...

osu.Framework/Audio/Sample/SampleChannel.cs
```diff
@@ -23,6 +28,15 @@ protected SampleChannel(Sample sample, Action<SampleChannel> onPlay)
...
         public virtual void Play(bool restart = true)
         {
+            if (ConcurrentPlayback)
+            {
+                if (Looping)
+                    throw new InvalidOperationException($"Cannot play a layered sample playback if {nameof(Looping)} is enabled.");
+
+                if (!restart)
+                    throw new ArgumentException("Cannot resume playback of a layered sample playback.", nameof(restart));
+            }
```
bdach, line 38: So I'm not sure where to add comments with this entire pull chain going on because it's really one huge blob at this point and everything affects everything, but I guess this is the most apt place. I personally very much don't like this API. `ConcurrentPlayback` getting set to `true` pretty much forces these two other flags' values and callers have to be unreasonably careful to not upset this at runtime. Similarly the throw in `Stop()` is suspect to me, for the same reason.

If you ask me - I'm fine with having `Get()` and `GetLayerable()` in concept, number of usages be damned - as they really *do* have different semantics. The first is expecting a more "track-like" behaviour, wherein the sample has duration and you want to make sure you are able to loop it, not double-play it, etc. etc., and what this PR is aiming to add is a more "sample-like", fire-and-forget... sample? Which is totally fine by me and probably even correct I'd say.

But I very much don't like the way this is written right now, because now the two share common API without actually *having* a sane common API, and you have to be immediately very careful what you're calling on what instead of just *not being able to call it*, because the code won't compile otherwise. If we do go down this path, I really think there really ought to be a `LayeredSampleChannel` or something, that's fully separated out of this, and has only declared operations/flags that are actually valid for it.

bdach: To put some money where my mouth is, here's a commit that feels like a slight improvement to me in this regard: https://github.com/ppy/osu-framework/commit/0e23f8fd25e633732558a4018e03c508ceca5132

It's not perfect, unfortunately. A few throws are still left in `SampleChannelBass` - would have to split that into two too to get rid - but it's not intended for direct usage so maybe fine...?

smoogipoo:
It becomes a bit of a mess. I took a stab at it myself, and while I've gotten the osu!framework-side to look like the following:
```csharp
sample = store.Get(name); // SampleChannel (or an ISampleChannel)
sample.Play(); // Compiles
sample.Play(false); // Does not compile
sample.Stop(); // Does not compile
sample.Looping = value; // Does not compile

//

sample = store.GetNonConcurrent(name); // INonConcurrentSampleChannel
sample.Play(); // Compiles (=> Play(true))
sample.Play(false); // Compiles
sample.Stop(); // Compiles
sample.Looping = value; // Compiles

// Note: I took the non-concurrent route in the hopes to reduce the changes in https://github.com/ppy/osu/pull/11482. It _should_ be invertible.
```
... the osu!-side changes look very, very, exceedingly bad. To the point where the same "NonConcurrent" moniker has to be applied to SkinnableSound, PoolableSkinnableSample, ISkin/Skin (including transformers, all derived types, skin providing container, and skinmanager), and many other miscellaneous tests.

There are ways you can probably clean things up a bit - like for example exposing `GetNonConcurrent()` as an extension method that returns a wrapped `SampleChannel` providing the extra methods and could clean up the store implementations, but that doesn't help with any of the usages of `SampleChannel`s that I've mentioned above.

I can't think of a way to make the fully-separate route pretty, however nice that sounds from an OOP sense. I'm even starting to consider removing the `GetLayerable` method and having it be part of the `SampleChannel` in the first place. Furthermore, I'd probably accept `ConcurrentPlayback = true` by default to reduce the changes seen osu!-side.

PR #4149, opened 2021-01-14
Implement a MVP SynchronizationContext for GameThread

peppy:
This doesn't necessarily need to be merged, but I wanted to get it PR'd to start a discussion (was a 15 minute implementation and lets me close a tab i had open to remind me to do this). I haven't yet explored the full scope of this (and if it can be used to potentially replace / fix / allow async in `BackgroundDependencyLoader` usage.

What this does already allow is usage in `Update` bound methods where we wish to `await` on other services or components, without the manual schedule continuation logic.

It may be decided that we want a context to exist and be set per `CompositeDrawable`, to ensure the continuations are run in the correct execution order. The overhead of `SetSynchronizationContext` is basically nil, so this is feasible.


smoogipoo:
I think it needs to be done per-`Drawable`. Having things scheduled to effectively the game's scheduler has bitten us hard in the past (see DLW/DLUW).

Otherwise, looks nice.

bdach:
[The docs](https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/consuming-the-task-based-asynchronous-pattern#suspending-execution-with-await) state that the synchronisation context is captured at the time of suspension, so it seems that the plan of switching the synchronisation context during drawable tree traversal seems feasible. I haven't done any testing yet, though.

peppy:
Bumping this with latest master to see if the `ConfigureAwait` changes we made helped with the seeming deadlock occurring on CI.

osu.Framework/Threading/Scheduler.cs
```diff
@@ -16,6 +16,13 @@ namespace osu.Framework.Threading
     /// </summary>
     public class Scheduler
     {
+        /// <summary>
+        /// A synchronisation context which posts all continuatiuons to this scheduler instance.
+        /// </summary>
+        public SynchronizationContext SynchronizationContext => synchronizationContext ??= new SchedulerSynchronizationContext(this);
+
+        private SynchronizationContext synchronizationContext;
```
smoogipoo, line 24: Is there a reason for this? `SynchronizationContext` is a pretty small object so it's not saving all that much.

I'm more concerned about thread-safety - `Scheduler` construction is thread-safe in `Drawable`, but this isn't.

smoogipoo (changes requested):

osu.Framework/Threading/Scheduler.cs
peppy: Probably not required, especially since `Scheduler` itself is lazy initialised.

peppy:
I've started to revisit this PR. @smoogipoo are you still hoping to see this on a per-`Drawable` basis? If this is the goal, then we need to figure out how this should work, as `Drawable.scheduler` is lazy-initialised and not always available. There are a couple of options:

- Have this as an opt-in per drawable that requires it (something like `override bool ScheduleAsyncContinuations => true`)
- Change the `SynchronizationContext` implementation to take a reference to the drawable instead of the scheduler, and handle the case where the scheduler may be null (initialising it at the point of usage).

I think option 2 is better and am going to try and continue in that direction, just wanted to bring this up for discussion in the mean time.

smoogipoo:
I prefer option 2 as well.

peppy:
This should be ready for review. I'm still testing osu!-side to make sure it covers the usage I need, so temporarily blocking merge until that is sorted.

ekrctb:
Another option (as opposed to attaching the context to a `Drawable`) is attach the synchronization context to a `Scheduler`, and the synchronization context is set during a `Scheduler.Update` call, and requiring explicit `Scheduler.RunAsync()` or something to spawn a task in the synchronization context of the scheduler.
I think this way is less prone to accidentally using the wrong context.

ekrctb:
Also, I doubt whether `SynchronizationContext.SetSynchronizationContext` should be called so frequently (not performance, but semantically).
By reading documents, I find an impression that the intended usage of the synchronization context is attaching one context to a thread and normally never changed.
<https://docs.microsoft.com/en-us/archive/msdn-magazine/2011/february/msdn-magazine-parallel-computing-it-s-all-about-the-synchronizationcontext>
> Another aspect of SynchronizationContext is that every thread has a “current” context. A thread’s context isn’t necessarily unique; its context instance may be shared with other threads. It’s possible for a thread to change its current context, but this is quite rare.

Perhaps it is better to use a field for tracking the current scheduler (or drawable) in one synchronization context, rather than using multiple synchronization contexts?

peppy:
Yeah, that's definitely a valid point and I did consider going in that direction. It's just a matter of "how to track the current drawable" if we do that, though. I'd rather not be DI'ing the `ThreadedSynchronisationContext` in every drawable's `Update`.

Not sure about requiring an explicit `async` call, probably needs a bit more thinking. The hope would be that we didn't have to, since that's approaching the same level of convoluted structure as we have now (ie. always requiring a `Schedule()` for continuations).

peppy:
I've revisited this as it looks like a very amicable alternative to thread pinning (in order to allow realm to understand that the update thread may change underlying managed thread).

> Also, I doubt whether `SynchronizationContext.SetSynchronizationContext` should be called so frequently (not necessary performance, but semantically).
> By reading documents, I find an impression that the intended usage of the synchronization context is attaching one context to a thread and normally never changed.
> https://docs.microsoft.com/en-us/archive/msdn-magazine/2011/february/msdn-magazine-parallel-computing-it-s-all-about-the-synchronizationcontext

This seems correct. Changing it per drawable also doesn't work for realm, as it also makes the thread==synccontext assumption.

> Perhaps it is better to use a field for tracking the current scheduler (or drawable) in one synchronization context, rather than using multiple synchronization contexts?

I also thought this would be a good fit, but i can't seem to make it work. I'm not sure we can do this because we would need to "capture" the scheduler at the point of async deviance, rather than the point of `Post`ing work. Did you happen to have a solution in mind for this?

Regardless, I've removed the drawable level context for now. I think getting this in is a good first step which we can build on if we find a way to (more correctly) schedule work to the local drawable.

ekrctb:
> I'm not sure we can do this because we would need to "capture" the scheduler at the point of async deviance, rather than the point of Posting work.

As I wrote in my comment above <https://github.com/ppy/osu-framework/pull/4149#issuecomment-889659373>, my solution needs an explicit call of a method on the correct `Scheduler` when the execution is getting into the async context from a non-async context. Like:

```diff
- box.OnUpdate += _ => asyncAction();
+ box.OnUpdate += _ => Scheduler.Add(asyncAction);
```

Then, during a `Scheduler.Update`, the "current scheduler" is set to the scheduler (also if it allows an inline execution, the current scheduler needs to be set there too). In this way, the async action will always resume in the scheduler first added to.

peppy:
Right. I think we can support this, but would you agree that's fine as a second and additional step

Installing an initial "fallback" `SynchronizationContext` as in this PR can still exist and catch cases where a user doesn't care about ordering.

ekrctb:
Sure. if there is a use-case of async and that would work in the current state of the PR, I agree this PR can be merged.

bdach (approved):

smoogipoo (approved):
Looks okay.

PR #4157, opened 2021-01-18
Change paths to match dotnet 5.0 for Rider and VSCode

MiraiSubject:


bdach:
The same will have to be done for template projects, which haven't been bumped up to .NET 5 yet.

MiraiSubject:
Oh right, will do that right away.

bdach:
It'll be wrong now, though, as you didn't bump package references in the templates.

For what it's worth, we also have [leftover references to `netcoreapp3.1` in rider run configs](https://github.com/ppy/osu-framework/blob/master/.idea/.idea.osu-framework.Desktop/.idea/runConfigurations/VisualTests.xml).

MiraiSubject:
Oh yeah just noticed, might as well do it right for everything at once

peppy (approved):

PR #4158, opened 2021-01-19
Fix video looping failing when frames are in available post-seek

peppy:
Depending on the timing of the decode loop, it is possible for there to still be frames arriving (or already arrived) from the end of the last playback when a loop is requested. We weren't accounting for this, meaning that some videos would fail the initial loop and only begin playing again after the seek lenience is exceeded.

Tested against and closes https://github.com/ppy/osu/issues/11490.


bdach (approved):
Makes sense to me, and confirmed that the issue is fixed with this change. :+1:

smoogipoo (approved):
I think this is fine? Kinda hard to parse but I have no suggestion for improvement.

bdach:
Yeah I struggled with parsing it too at first but that extra set of parens helped a lot when I figured out what it was doing there.

PR #4160, opened 2021-01-19
Extend render buffer to use all GL internal formats

Flutterish:


bdach:
I've applied several changes that aim to bring this pull in line with our coding style and conventions (in particular: whitespace and not using switch expressions), and also improve maintainability (by reordering the enum to match GL ES spec, and also linking said spec in the comment).

@Flutterish if you should want to continue submitting code to this project, I'd ask that you keep in mind our styling conventions instead of using your own.

bdach (approved):

smoogipoo (approved):

PR #4161, opened 2021-01-19
Rework Sample and SampleChannel to better handle concurrent playbacks

smoogipoo:
Alternative to / closes https://github.com/ppy/osu-framework/pull/4148.

Currently `Sample` and `SampleChannel` are very weird components - a `Sample` is the internal member that stores the sample ID, and a `SampleChannel` is what the user receives via `sampleStore.Get()`. A `SampleChannel` then plays and handles multiple playbacks internally.

This is very orthogonal to the way BASS handles samples - a channel is a single unique playback from a sample, and multiple channels can be played back concurrently.

This PR aims to implement a similar system to that of BASS:
* `SampleStore` contains a set of `SampleBassFactory`s, a unique factory for each sample name.
* `SampleBassFactory` holds the sample ID and provides the `CreateSample()` method. This object is hidden from the user.
* `Sample` is created via `SampleBassFactory.CreateSample()` and returned to the user from `SampleStore.Get()`.
  * This is a pretty basic object now, containing only a `Play()` method.
 * `SampleChannel` is returned from `Sample.Play()` and represents a single unique playback of the sample in a dedicated BASS channel.
   * This still provides `Stop()` and `Looping`, but only controls its own playback.

```
  (public)        (internal)       (public)    (public)

                                 |> Sample |> SampleChannel
            |> SampleBassFactory |
            |                    |> Sample
            |
            |> SampleBassFactory |> Sample |> SampleChannel
            |                              |> SampleChannel
            |                              |> SampleChannel
SampleStore |                              |> SampleChannel
            |
            |> SampleBassFactory |> Sample
            |
            |                    |> Sample
            |> SampleBassFactory |> Sample |> SampleChannel
                                 |> Sample
```
Example:
```csharp
var sample = samples.Get("name");

// Play a looping channel.
var channel1 = sample.GetChannel();
channel1.Looping = true;
channel1.Play();

// Play a normal channel concurrently.
// A new channel is returned, but playback will only be heard concurrently up to SampleStore.PlaybackConcurrency times.
var channel2 = sample.Play();

// Apply a volume adjustment to all channels (current and future) from the sample.
sample.Volume.Value = 0.5;

// Apply an independent volume adjustment to channel 1 only.
// The final volume of this channel will be 0.5 (sample) * 0.5 (channel) = 0.25.
channel1.Volume.Value = 0.5;

// Stop channel 2, but keep channel 1 playing.
channel2.Stop();

// Resume playback of channel 2. This doesn't restart unless playback has finished.
channel2.Play();

// Disposes the sample and all channels, halting playback.
sample.Dispose();
```

A big tossup I had was whether each call to `SampleStore.Get()` should return a new `Sample` instance. This is the case in the end. The common member of all `Sample`s is the internal `SampleBassFactory` that owns the sample ID. There's a few reasons for this.

If `Sample` owns the sample ID, then it should follow that `Sample` should be the one to clean it up on disposal, but `Sample.Dispose()` is unsafe from a consumer as it will permanently break that `Sample` for any further and existing usages. A contributing factor to this is that `SampleStore` would have to cache the `Sample` instances in order to achieve this, and would need additional provisions to allow "refreshing" the cache on disposal, however other usages of the sample would still remain broken.

The second reason is that having `Sample` inherit from `AdjustableAudioComponent` is useful so that it can be used as a parameter to `DrawableAudioWrapper` (from `DrawableSample`). It's possible to work around this by adding `DrawableSampleChannels` internally to the `DrawableSample` every `Play()`, but it gets very messy to do so.  

Finally, I see `Sample` being adjustable itself potentially reducing the amount of state handling that needs to be done by the user, for example in cases where samples are layered and e.g. balance is updated every update frame.

# Breaking Changes

## `ISampleStore.Get()` now returns `Sample`s instead of `SampleChannel`s

The sample subsystem has been reworked to better encapsulate sample channels. A `SampleChannel` is now a single unique playback of a `Sample`.

Example migration code:
```csharp
// Before:

SampleChannel sample = sampleStore.Get("my_sample");

sample.Looping = true;
sample.Play();

// ...

sample.Stop();

// ---------------------------------------------------------

// After:

Sample sample = sampleStore.Get("my_sample");

SampleChannel channel = sample.GetChannel();
channel.Looping = true;
channel.Play();

// ...

channel.Stop();
```

For concurrent playback, play the `Sample` multiple times. The maximum number of concurrent playbacks can be controlled via `ISampleStore.PlaybackConcurrency` (default: 2):
```csharp
Sample sample = sampleStore.Get("my_sample");

sample.Play(); // Playback #1
sample.Play(); // Playback #2
```


bdach:
I couldn't find enough time in the day today to look at this properly, but I've scrolled through the description and some usage code, and I'm still definitely on board with this. Will have a proper look through in the coming days, hopefully, if the overall traffic is merciful.

smoogipoo:
This PR is now ready, I've updated the PR description.

osu.Framework/Audio/Sample/SampleBassFactory.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            EnqueueAction(() =>
+            {
+                // Todo: Broken (SEGV on SampleGetInfo())
+                // if (!IsLoaded)
+                //     return;
+                //
+                // var sampleInfo = Bass.SampleGetInfo(SampleId);
+                // sampleInfo.Max = concurrency.NewValue;
+                // Bass.SampleSetInfo(SampleId, sampleInfo);
```
frenzibyte, line 51: Have dug through this, and finally found the reason. :p

Firstly checked to ensure that it doesn't happen on a simple C project, and it indeed doesn't happen there and works perfectly fine.

Then I started fiddling around with an isolated version of this in C#, until I found out that `ManagedBass` is likely to be at fault, as using [radio42's `Bass.NET` wrapper](http://bass.radio42.com/) and calling `SampleGetInfo` works correctly in there.

After checking through `ManagedBass`'s way of doing so: 
image

and looking back at the native code's declaration:
image


According to my knowledge, and [the docs with their near-equivalent sample](https://docs.microsoft.com/en-us/dotnet/framework/interop/marshaling-classes-structures-and-unions#systime-sample), `ManagedBass`'s method prototype is incorrectly declared, reference types are already pointers by themselves, what it should've been is:
```csharp
public static extern bool SampleGetInfo(int Handle, [In, Out] SampleInfo info);
```

(out attribute for indicating that changes to sample info by unmanaged code are expected)

Tested it and it indeed works now, I'll open a PR for that on their side, not sure how active they are these days but we'll see.

smoogipoo: Good stuff. They're pretty active, but we're still yet to update to ManagedBass 3.0.0 due to lack of testing / it not working on mobile platforms due to the removal of dllmap.

peppy:
What happens if a sample finishes before `.Looping = true` is run? It's a pretty edge case, but seems like a potential oversight.

A more common one may be setting the volume on a playback, but the sample begins playing at full volume for a brief moment before the change has propagated.

smoogipoo:
Both are very good points... I'm not entirely sure how to address those issues at the moment, so I need to mull over it a bit.

I'd like to avoid having `.Play()` on `SampleChannel` or in any place where it can be called multiple times and be required to throw an exception the second time around (e.g., was also considering exposing `CreateChannel()` which you could `.Play()` at the right time).

I'm thinking something along the lines of a `PlayWith(bool? looping, ...)` method. Would like for the process to be `.Play().With(looping: true)` but I don't know if that's possible...

peppy:
What about the case where you want to play a channel a second time? Or when you stop a looping channel but want to start it again?

smoogipoo:
Are you referring to pausing and continuing of channels? That can't happen, it makes no sense in terms of concurrency. If you want that then we can't use BASS concurrency at all.

smoogipoo:
I suppose it can work... So you're imagining the usage being:

```
var sample = sampleStore.Get(...);
var channel = sample.GetChannel();
channel.Play();
channel.Stop();
channel.Play(); // Resumes the channel from the last position, or restarts if it's at the end.
```

peppy:
I am, yeah. This was previously supported by our API, and no longer is.

There's one usage of precisely this (stopping the spinner looping sound and resuming from the point it was stopped).

peppy:
Is there a reason for not making the restart optional, like we have on `master`?

smoogipoo:
No reason, I'll see how the changes feel.

smoogipoo:
Also noticed one more issue with this which is that it's likely very leaky as `Sample`s are added to the collection but never removed.

smoogipoo:
I think I've covered all issues mentioned above now, and updated the PR description.

`TestTogglePlaying` is audibly broken on master regardless of `restart = false` - it will always restart the sample since `Stop()` frees the sample. It doesn't fail the test, but making it fail the test is quite a lot of trouble. This is fixed in this PR.

However I need feedback on a particular edge-case. When the sample reaches the end, a call to `Play()` will restart the sample. There's no special handling for this - I'm just calling `Bass.ChannelPlay()` and it's just how BASS works internally. `TrackBass` has custom logic for stopping the restart in this case, so if that's the more desirable option, I can add similar logic here.

As for setting `Looping`,`Volume`, etc - now that the `GetChannel()` method exists I think it's reasonable for the user to expect that a sample could have stopped playing between a .`Play` and `.Looping` adjustment, or could have started playing between a `.Play()` and `.Volume` adjustment.
Such properties can be set via the `GetChannel()` method, prior to `.Play()`.

bdach:
I think not allowing a completed sample playback to play again is the better option, for two reasons:

* Consistency with `TrackBass`
* Given now we have proper sample channels, which are single-use "playback objects", then a single use of a non-looping sample implies it should not be possible to start it again

The various adjustments (looping/volume) seem fine to be potentially racey to me. The API should be clear enough that the user should be expecting to set these before firing a playback.

osu.Framework/Audio/Sample/SampleBassFactory.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            PlaybackConcurrency.BindValueChanged(updatePlaybackConcurrency);
+        }
+
+        private void updatePlaybackConcurrency(ValueChangedEvent<int> concurrency)
+        {
+            EnqueueAction(() =>
+            {
+                // Broken in ManagedBass (https://github.com/ManagedBass/ManagedBass/pull/85).
```
bdach, line 45: For what it's worth upstream has merged this now, but no release yet.

smoogipoo:
> I think not allowing a completed sample playback to play again is the better option

I can't disagree with any of the reasons provided, so I've done it. The only weirdness is it can't use the same process as `TrackBass` because `ChannelSetSync()` seems to not be applicable to sample channels (only streams, rather confusingly).

osu.Framework.Tests/Visual/Audio/TestSceneSamples.cs
```diff
@@ -196,8 +196,8 @@ public void Play()
                 Played = true;
                 circle.ScaleTo(1.8f).ScaleTo(1, 600, Easing.OutQuint);
 
-                sample.Frequency.Value = 1 + Y / notes;
-                sample.Play();
+                var channel = sample.Play();
+                channel.Frequency.Value = 1 + Y / notes;
```
peppy, line 200: isn't this wrong? the sample will potentially start with the wrong frequency?

osu.Framework/Audio/Sample/SampleBassFactory.cs
peppy: so while this is broken, what behaviour do we actually get? is it using some global value still? or is this going to regress playback?

osu.Framework.Tests/Visual/Audio/TestSceneSamples.cs
peppy: I think this can be resolved by changing it to

```csharp
                var channel = sample.GetChannel();
                channel.Frequency.Value = 1 + Y / notes;
                channel.Play();
```

osu.Framework/Audio/Sample/SampleChannelBass.cs
```diff
@@ -1,30 +1,58 @@
...
+        }
+
+        public override void Play()
+        {
+            userRequestedPlay = true;
+
+            // Enqueue the playback to start.
+            //
+            // The playing state set below combined with the call to base.Play() will make this channel visible to the audio thread if it's not already.
+            // If the the audio thread were to update and see the channel not playing, it would reset the playing state and remove the channel from the audio thread's visibility.
```
peppy, line 46:
```suggestion
            // If the audio thread were to update and see the channel not playing, it would reset the playing state and remove the channel from the audio thread's visibility.
```

osu.Framework/Audio/Sample/SampleChannelBass.cs
```diff
@@ -1,30 +1,58 @@
...
+            //
+            // In order to prevent this race, playback is enqueued first so that the audio thread is guaranteed to play the channel when it becomes visible to it.
+            playChannel();
+
+            // The playing state keeps the channel alive to receive updates from the audio thread. It will not receive updates until base.Play().
+            playing = true;
+
+            // Notify to Sample/SampleBassFactory that this channel has come alive.
```
peppy, line 54:
```suggestion
            // Notifies Sample/SampleBassFactory that this channel has come alive.
```

osu.Framework/Audio/Sample/SampleChannelBass.cs
```diff
@@ -58,80 +86,115 @@ public override bool Looping
...
-                    if (!restart)
-                        return;
-
-                    Stop();
+                    case PlaybackState.Playing:
+                    case PlaybackState.Stalled: // Stalled counts as playing, as playback will continue once more data has streamed in.
+                        playing = true;
+                        break;
+
+                    case PlaybackState.Paused when userRequestedPlay:
```
peppy, line 101: if you're going to use `when` here (would probably prefer not doing so) why not combine with the above case?

peppy (changes requested):

osu.Framework/Audio/Sample/SampleChannelBass.cs
smoogipoo: > if you're going to use `when` here

Personally I find the switch statement easier to read, and this is as simple as the condition is going to get without explicitly mentioning the stalled case:
image

peppy: yeah, that's what i was proposing. i think that sits better than splitting them out?

smoogipoo: I've already done that in the latest commit, the comment above was only directed towards the use of `when`.

peppy (approved):
There's a lot of threading logic changes here but it looks to be pretty good (at least the API is something I can get behind now). Let's get this out in a release for testing further.

osu.Framework/Audio/Sample/SampleBassFactory.cs
smoogipoo: it's just not going to support per-`Sample` playback concurrency. the `ISampleStore` one will always be used at the point when the sample is initialised, like on current master.

peppy: Think it's worth adding a note on the property until this is fixed? Maybe not considering it's a new API and likely used by noone.

smoogipoo: Will do

peppy: or just note expose it until it's working?

PR #4162, opened 2021-01-19
Add tablet support

InfinityGhost:
- Added OpenTabletDriver package with implementation for raw input

- [x] Depends on #4166


InfinityGhost:
Looks like its building for .NET Core 3.1 in CI, is this updated on master?

bdach:
Where are you drawing that conclusion from? inspectcode is not the compiler, fwiw.

osu.Framework/Input/States/TabletState.cs
```diff
@@ -0,0 +1,51 @@
...
+    public class TabletState
+    {
+        public readonly ButtonStates<TabletPenButton> PenButtons = new ButtonStates<TabletPenButton>();
+        public readonly ButtonStates<TabletAuxiliaryButton> AuxiliaryButtons = new ButtonStates<TabletAuxiliaryButton>();
+
+        public Vector2 Position { set; get; }
+
+        public double Pressure { set; get; }
+
+        public bool IsStateValid { set; get; } = true;
```
bdach, line 19: What does this mean exactly? Is it similar to `MouseState.IsPositionValid` in some capacity? Mind adding xmldoc for this?

osu.Framework/Input/States/InputState.cs
```diff
@@ -48,13 +53,15 @@ public InputState(InputState other)
         /// <param name="mouse">The mouse state.</param>
         /// <param name="keyboard">The keyboard state.</param>
         /// <param name="touch">The touch state.</param>
+        /// <param name="tablet">The tablet state</param>
         /// <param name="joystick">The joystick state.</param>
         /// <param name="midi">The midi state.</param>
-        public InputState(MouseState mouse = null, KeyboardState keyboard = null, TouchState touch = null, JoystickState joystick = null, MidiState midi = null)
+        public InputState(MouseState mouse = null, KeyboardState keyboard = null, TouchState touch = null, TabletState tablet = null, JoystickState joystick = null, MidiState midi = null)
```
bdach, line 59: Technically a breaking change to framework consumers due to adding an argument in the middle. Maybe should go last instead.

osu.Framework/Input/States/InputState.cs
```diff
@@ -48,13 +53,15 @@ public InputState(InputState other)
         /// <param name="mouse">The mouse state.</param>
         /// <param name="keyboard">The keyboard state.</param>
         /// <param name="touch">The touch state.</param>
+        /// <param name="tablet">The tablet state</param>
```
bdach, line 56:
```suggestion
        /// <param name="tablet">The tablet state.</param>
```

osu.Framework/Input/Handlers/Tablet/RelativeTabletMode.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+#if NET5_0
+using OpenTabletDriver.Plugin.Output;
+using OpenTabletDriver.Plugin.Platform.Pointer;
+
+namespace osu.Framework.Input.Handlers.Tablet
+{
+    public class RelativeTabletMode : RelativeOutputMode
```
bdach, line 10: I'm not sure what this is, or if it needs to be `public` - xmldocs, and making `internal`, would be good if possible.

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+
+namespace osu.Framework.Input.Handlers.Tablet
+{
+    public class OpenTabletDriverHandler : InputHandler, IAbsolutePointer, IVirtualTablet, IRelativePointer
+    {
+        public override bool IsActive => TabletDriver.EnableInput;
+
+        public override int Priority => 0;
+
+        protected FrameworkTabletDriver TabletDriver { set; get; }
```
bdach, line 23: Why's this `protected`?

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+        public override bool Initialize(GameHost host)
+        {
+            TabletDriver = new FrameworkTabletDriver
+            {
+                OutputMode = new AbsoluteTabletMode(this)
+            };
+
+            TabletDriver.TabletChanged += (sender, e) =>
+            {
+                if (TabletDriver.OutputMode is AbsoluteOutputMode absoluteOutputMode && TabletDriver.Tablet is TabletState tablet)
```
bdach, line 34: Isn't the latter part of this (the `is TabletState` pattern match) redundant? Or is this just a fancy null check?

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+                    {
+                        // Ideally would be the maximum window dimensions
+                        Width = outputWidth = host.Window.ClientSize.Width,
+                        Height = outputHeight = host.Window.ClientSize.Height,
+                        Position = new Vector2(outputWidth / 2, outputHeight / 2)
+                    };
+                }
+            };
+
+            TabletDriver.ReportRecieved += (sender, report) =>
```
bdach, line 58: seems like you have a typo in your library (should be `Received`) :stuck_out_tongue:

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+            }, true);
+
+            return true;
+        }
+
+        public void SetPosition(Vector2 pos) => enqueueInput(new MousePositionAbsoluteInput { Position = new osuTK.Vector2(pos.X, pos.Y) });
+
+        public void SetPressure(float percentage) => enqueueInput(new MouseButtonInput(osuTK.Input.MouseButton.Left, percentage > 0));
+
+        public void Translate(Vector2 delta) => enqueueInput(new MousePositionRelativeInput { Delta = new osuTK.Vector2(delta.X, delta.Y) });
```
bdach, line 87: I'm super confused about these - they don't seem to have any actual usages?

InfinityGhost:
> Where are you drawing that conclusion from? inspectcode is not the compiler, fwiw.

Nvm, made that guess based on this at first glance
`Running on AMD 64 in 64-bit mode, .NET runtime 3.1.11`

osu.Framework/Input/Handlers/Tablet/FrameworkTabletDriver.cs
```diff
@@ -0,0 +1,60 @@
...
+            foreach (var config in getConfigurations())
+            {
+                if (TryMatch(config))
+                    break;
+            }
+        }
+
+        private IEnumerable<TabletConfiguration> getConfigurations()
+        {
+            // Retreive all embedded configurations
```
bdach, line 42:
```suggestion
            // Retrieve all embedded configurations
```

osu.Framework/Input/Handlers/Tablet/FrameworkTabletDriver.cs
```diff
@@ -0,0 +1,60 @@
...
+            var asm = typeof(Driver).Assembly;
+            return from path in asm.GetManifestResourceNames()
+                   where path.Contains(".json")
+                   let stream = asm.GetManifestResourceStream(path)
+                   select deserialize(stream);
```
bdach, line 47: Not sure this is something that can't be achieved with the fluent LINQ syntax which we tend to prefer.

bdach (changes requested):
Have not done any testing on this yet, just read through the diff relatively quickly.

osu.Framework/Input/TabletPenButton.cs
```diff
@@ -0,0 +1,17 @@
...
+    public enum TabletPenButton
+    {
+        Primary,
+        Secondary,
+        Tertiary,
+        Button4,
+        Button5,
+        Button6,
+        Button7,
+        Button8
```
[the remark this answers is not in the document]
InfinityGhost: Could remove a few of the pen buttons, did this in the case we may end up with more. In the auxiliary case however, we've seen tablets with 10+ buttons. 4 total sound better?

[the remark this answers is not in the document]
bdach: Oh interesting, I would not have expected 10+. Fine to leave, I was just mostly curious about this point.

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
InfinityGhost: Fancy null check, will change if it matches better with the coding conventions.

bdach: Yeah I'd probably prefer to be explicit. This is a bit too clever/weird as written right now I'd say.

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
InfinityGhost: Whoops, will fix this with the 0.5.1 package later.

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
InfinityGhost: The handler itself implements the interfaces the driver uses for cursor positioning, so we pass the handler itself as the pointer handler. Will add docs to clear this up if needed.

- https://github.com/InfinityGhost/osu-framework/blob/766ecf65a7c8141c0616a1344f9b080943d2d50e/osu.Framework/Input/Handlers/Tablet/AbsoluteTabletMode.cs#L12-L15
- https://github.com/InfinityGhost/OpenTabletDriver/blob/master/OpenTabletDriver.Plugin/Output/AbsoluteOutputMode.cs#L123-L124

bdach: Oh, so these are interface implementations, right. I'd consider making them explicit implementations, as I don't really see much reason why these should be publically visible - and doing so will immediately clear up the confusion due to declaring the interface type.

osu.Framework/Input/Handlers/Tablet/RelativeTabletMode.cs
InfinityGhost: This would be used in the case a user wished for relative input from their tablet. Right now it goes unused until its supported, or can flat out just be removed if there's no plans to make relative input supported.

osu.Framework/Input/States/InputState.cs
InfinityGhost: Will move it to the end.

osu.Framework/Input/States/TabletState.cs
InfinityGhost: Similar to IsStateValid of MouseState, yes, however it goes unused so will remove.

osu.Framework/Input/Handlers/Tablet/FrameworkTabletDriver.cs
InfinityGhost: Changed to the function syntax.

osu.Framework/Input/TabletPenButton.cs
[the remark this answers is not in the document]
InfinityGhost: I do agree on the pen button count being a lot, however, maybe worth changing if anyone else thinks it should be reduced.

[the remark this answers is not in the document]
bdach: We have up to 64 joystick axes in `InputKey` as is so if this were to be a problem we have much worse problems already ;)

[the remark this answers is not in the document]
InfinityGhost: Will leave it as is then

InfinityGhost:
Current issue, window size change doesn't update the output area, and there doesn't appear to be any bindable or event to hook to update this.

bdach:
Some pieces are already in place to get one in for SDL2 at least:

https://github.com/ppy/osu-framework/blob/0a8294590f2acb898096668591577d76367720ac/osu.Framework/Platform/SDL2DesktopWindow.cs#L448-L466

but they're not exposed in a nice enough way to be usable just yet.

InfinityGhost:
> Some pieces are already in place to get one in for SDL2 at least:
> 
> https://github.com/ppy/osu-framework/blob/0a8294590f2acb898096668591577d76367720ac/osu.Framework/Platform/SDL2DesktopWindow.cs#L448-L466
> 
> but they're not exposed in a nice enough way to be usable just yet.

Should that change be in a separate PR or on this branch? I don't want to break too far out of scope of this pull request unless its fine to do so with a relatively smaller change like that.

frenzibyte:
In a separate PR, for the same reason you stated, smaller changes that can stand on their own like this and are unrelated to the scope of this PR go into different small PRs this would depend on, for review and merge ease.

peppy:
This whole API feels very haphazard to me. Setting the `OutputMode` to an instance which is then used to actually interface with the API isn't something I've seen done before. Usually you'd be changing the mode on the library using a `Configuration` or `Enum` object and then consuming it.

Is there a specific reason for doing things the way you are? Or is that just how it ended up?

InfinityGhost:
> This whole API feels very haphazard to me. Setting the `OutputMode` to an instance which is then used to actually interface with the API isn't something I've seen done before. Usually you'd be changing the mode on the library using a `Configuration` or `Enum` object and then consuming it.
> 
> Is there a specific reason for doing things the way you are? Or is that just how it ended up?

The API was originally built to support expansion with plugins is the main reason to just expose and directly set the OutputMode instead of defining it in a configuration object.

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
InfinityGhost: Package updated, this is now resolved.

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
```diff
@@ -0,0 +1,56 @@
...
+namespace osu.Framework.Input.Handlers.Tablet
+{
+    public class TabletDriver : Driver
+    {
+        public TabletDriver()
+        {
+            Log.Output += (sender, logMessage) => Logger.Log($"{logMessage.Group}: {logMessage.Message}");
+            DevicesChanged += (sender, args) =>
+            {
+                if (Tablet == null && args.Additions.Any())
```
peppy, line 23: Is the current device getting disconnected supported somewhere? I can see cases of `Tablet` going from `null` to something, but is the reverse not also required to handle this workflow?

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
```diff
@@ -0,0 +1,56 @@
...
+                    DetectTablet();
+            };
+        }
+
+        public void DetectTablet()
+        {
+            foreach (var config in getConfigurations())
```
peppy, line 30: This is parsing the configuration files every time a tablet connection change occurs. Approximately how many of these are there? Wondering if it's worth caching them.

Also we need to consider users which have no tablets, where this kind of overhead may be unnecessary/unwanted. Definitely needs benchmarking, as this will add a synchronous overhead to framework startup.

peppy (changes requested):

peppy:
I've done an initial clean-up pass, and things look pretty good apart from the pointed out issues.

I haven't tested this yet; need to dig out my test tablets (will do so this week).

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
InfinityGhost: We have 78 configurations and soon more to be added, and its only around 336KiB for all of them. As for the people without tablets, maybe we could just disable the tablet input handler by default and make it something you enable manually.

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
InfinityGhost: When `Driver.TryMatch()` is called, if it fails to match, Tablet is set to null.

https://github.com/InfinityGhost/OpenTabletDriver/blob/master/OpenTabletDriver/Driver.cs#L69-L92

InfinityGhost: Also here
https://github.com/InfinityGhost/OpenTabletDriver/blob/master/OpenTabletDriver/Driver.cs#L170-L175

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
bdach: Is the entire json required to match a device? Looking at a few configurations, the name and OEM strings should be hopefully enough? Or maybe we can pre-construct some sort of matching mapping, so that only the configs that need to actually be loaded are loaded?

Definitely against having this be enabled manually. This should "just work" out of the box.

InfinityGhost: > Is the entire json required to match a device?

No, but it makes it much simpler to do. We could manually initialize but it complicates things for very little benefit.

> Looking at a few configurations, the name and OEM strings should be hopefully enough?

We still need the VID and PID of the device, since the name is just manually defined for organizational purposes and has no effect on tablet detection.

bdach: It would be maybe helpful to have some sort of fingerprinting mechanism, i.e. take out a few important unique bits of information about the newly-connected device, hash it up and throw that at a look-up table to see if there's a matching config. I have no idea if that's viable, however. Any sort of heuristics would probably render that unviable.

peppy: 336kb is a lot of json to parse (i wouldn't say "only"). definitely going to require further thought. disabling the handler by default is not a valid solution.

InfinityGhost: > It would be maybe helpful to have some sort of fingerprinting mechanism, i.e. take out a few important unique bits of information about the newly-connected device, hash it up and throw that at a look-up table to see if there's a matching config. I have no idea if that's viable, however. Any sort of heuristics would probably render that unviable.

This would be fine for Wacom and some XP-Pen tablets, but the issue arises when you get to Huion, VEIKK, etc. where they put identifier information in strings on the device that can differ (some include the manufacture date), which would throw a fingerprint off entirely. At most the only identifiers that are always consistent are the VID and PID, but we'd still end up having to parse *some* configurations at some point or another to get the right tablet specs.

bdach: Somehow I'm not surprised.

I dunno what to suggest at this point but parsing *every single* supported configuration to find the device really feels incorrect. There's not *that* many of them, but it's still well in the tens of JSON files to open.

Heck even bundling into some sort of single binary file, allowing to only read one file and skip on the parsing, would probably be nicer. Which I realise is troublesome for maintaining the configuration database, but maybe it could be done as a build step on the library side.

InfinityGhost: Just out of curiosity I wrote up a little benchmark for reading configurations via assembly using (roughly) the same code we're using here.

[Results](https://termbin.com/44v7); [Code](https://gist.github.com/InfinityGhost/646eb01e47abd83965710406712b6fa5)

Is 0.18s too long for parsing? Most of that time is waiting on the initial `asm.GetManifestResourceNames()` call, which we could just cache that call's output and make that a one-time slowdown at initialization.

bdach: It's not *horrible* but I wouldn't exactly call it negligible either.

InfinityGhost: Added a second run, storing the assembly and resource names in fields.

[Results](https://termbin.com/x58b); [Code](https://gist.github.com/InfinityGhost/646eb01e47abd83965710406712b6fa5#file-benchmark-with-caching-cs)

```
All configurations parsed in 00:00:00.1570357 including initial load.
~
All configurations parsed in 00:00:00.0039036 after initial load.
```

peppy: Firstly, we don't really use homegrown benchmarks, please use `BenchmarkDotNet`.

And yes, the highest allowable overhead would be around 20ms (would prefer <1ms). Nothing should be parsing json on startup if we can avoid it.

InfinityGhost:
Added a benchmark for testing detection performance, here's my run.

## Results
``` ini

BenchmarkDotNet=v0.12.1, OS=nixos 21.03pre269343.758b29b5a28
Intel Core i7-4790 CPU 3.60GHz (Haswell), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=5.0.100
  [Host]     : .NET Core 5.0.0 (CoreCLR 5.0.20.51904, CoreFX 5.0.20.51904), X64 RyuJIT
  DefaultJob : .NET Core 5.0.0 (CoreCLR 5.0.20.51904, CoreFX 5.0.20.51904), X64 RyuJIT


```

#### Without a tablet plugged in
|          Method |     Mean |   Error |  StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------- |---------:|--------:|--------:|-------:|------:|------:|----------:|
| DetectBenchmark | 604.6 ns | 6.56 ns | 5.82 ns | 0.1297 |     - |     - |     544 B |


#### With a Wacom CTL-470 plugged in
|          Method |     Mean |    Error |   StdDev |   Median | Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------- |---------:|---------:|---------:|---------:|------:|------:|------:|----------:|
| DetectBenchmark | 340.6 ms | 21.90 ms | 64.58 ms | 299.2 ms |     - |     - |     - | 806.61 KB |

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
```diff
@@ -0,0 +1,60 @@
...
+            DevicesChanged += (sender, args) =>
+            {
+                if (Tablet == null && args.Additions.Any())
+                    DetectTablet();
+            };
+        }
+
+        public void DetectTablet()
+        {
+            if (CurrentDevices.Select(d => d.VendorID).Intersect(known_vendors).Any())
```
bdach, line 33: So as I understand this code and the benchmark results, if there is no tablet connected, tablet detection will always take less than a millisecond, and if there is a tablet connected, figures in the hundreds of milliseconds can be expected. Which still seems pretty high.

This, to me, begs the question - if `VendorID` checking is happening anyway, can this `VendorID` not be used to at least thin out some devices initially? You've stated that the VID and PID should be consistent, so it should be theoretically possible to skip parsing some absolutely useless configs that have no chances of being matched?

InfinityGhost: Some optimization is completely doable since vendor IDs are assigned by the USB-IF, but product IDs are set by the manufacturer themselves which can make them unreliable for detection. In the case of UCLogic tablets such as Huion, XP-Pen, and VEIKK, the product IDs haven't been very reliable at all and additional checks have to be done to verify the tablet.

Most of the benefits from grabbing specific vendors would be on Wacom since its the only graphics tablet vendor with their own digitizers and they have their own vendor ID, where the UCLogic tablets typically share the UCLogic vendor ID.

peppy: Are you saying there are tablets which specify the vendor ID of a different vendor?

InfinityGhost: > Are you saying there are tablets which specify the vendor ID of a different vendor?

Basically, yes due to rebranding of tablets.

peppy: So it's a correct assumption that vendor IDs are not used in the matching process at all, and only for the initial "is any tablet connected at all"?

InfinityGhost: It's still used in the detection process for tablets that actually conform to vendor and product IDs, so there would be some benefit of using vendor IDs, but the benefits wouldn't likely be anything groundbreaking, moreso on those exceptions to VIDs. Wacom for instance would benefit since it would only have to look through Wacom configurations, but Huion would still end up looking through both Huion and Gaomon configurations since they share a VID.

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
```diff
@@ -0,0 +1,149 @@
...
+            {
+                case AbsoluteOutputMode absoluteOutputMode:
+                {
+                    float outputWidth, outputHeight;
+
+                    // Set output area in pixels
+                    absoluteOutputMode.Output = new Area
+                    {
+                        Width = outputWidth = window.ClientSize.Width,
+                        Height = outputHeight = window.ClientSize.Height,
```
peppy, line 89: This may need to change with incoming ClientSize fixes (making it correctly be scaled by the DPI setting).

peppy:
The `TabletDriver` constructor is taking 500ms here. This is unacceptably high.

peppy:
Here's a 10 second profiling effort:

20210305 133231 (JetBrains Rider)

If you'd prefer I step in and try and fix the performance issues at OTD's end, let me know. This is a relatively high priority for us as we move away from the weirdly broken raw input implementation we've been relying on, so I don't mind helping get things in a working state if that works better for you.

InfinityGhost:
> The `TabletDriver` constructor is taking 500ms here (with no devices connected). This is unacceptably high.

This is easily fixable, it's leftovers from the desktop plugin system that uses reflection which was really slow but versatile. https://github.com/OpenTabletDriver/OpenTabletDriver/pull/825 should resolve this, try with these changes.

Would test myself, but I'm waiting on my OSS rider license lol

InfinityGhost:
Performance fixes are within the v0.5.2 update, this should be basically ready now.

osu.Framework/Input/Events/TabletEvent.cs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Collections.Generic;
+using JetBrains.Annotations;
+using osu.Framework.Input.States;
+
+namespace osu.Framework.Input.Events
+{
+    public abstract class TabletEvent : UIEvent
```
frenzibyte, line 10: these events should probably have an overridable event handlers in `Drawable` for, with xmldoc.
```csharp
protected virtual bool OnTabletPenButtonPress(TabletPenButtonPressEvent e) => Handle(e);
protected virtual void OnTabletPenButtonRelease(TabletPenButtonReleaseEvent e) => Handle(e);
...
```

frenzibyte (changes requested):
Tablet pen and auxiliary button events need to be stored into `InputKey` enum and handled in both `KeyBindingContainer.Handle()` and `KeyCombination.FromInputState()` for key binding support.

Make sure the input keys [readable](https://github.com/ppy/osu-framework/blob/54f470560381f7826c986e572ddf5d4ba21038c1/osu.Framework/Input/Bindings/KeyCombination.cs#L156), `"Tablet Pen X"` and `"Tablet Aux X"` should do (all starting from `1`).

Also, handling for tablet events in `PassThroughInputManager` is missing.

---

Aside from the above, currently even if the tablet is active inside the game, the cursor must enter the window for the input manager to allow clicking and hovering with tablet.

This is due to `CursorInWindow` being false, leading `UserInputManager` to block mouse events. Sounds like it should add an exception for mouse input that's coming from tablet source (similar to how it's done with touch, via `ISourcedFromTouch`).

---

Other than the above, tablet input flow looks good, testing with osu!lazer.

bdach:
On a multi-monitor linux setup I'm seeing the game receive global coordinates - as in across all desktops, so for example if I have my tablet mapped to the right display, I can only ever see the cursor if the tablet pen is at the left edge of the input (where the screens join). Can someone verify if this is only a linux thing, or will this be the case on all platforms?

bdach:
So... the tablet detection refused to work for me on linux, and I traced it back to [this FAQ entry](https://github.com/OpenTabletDriver/OpenTabletDriver/wiki/Linux-FAQ#the-driver-fails-to-open-the-tablet-deviceioexception). And I have *several* questions...

Is it required for me to have to clone a separate repo and do admin-level tasks to have this work? I mean sure, it's linux, I'm fine with tinkering, but I don't really think we want to be getting user complaints about this. It should at least present a reasonable error that links to this FAQ, I think? Not to mention that the entire notion of a "usermode driver" kinda goes out the window if I have to do this... I haven't even checked what the rules do yet.

InfinityGhost:
> 0666 seems egregious. rw-rw-rw-? Really? That sounds like an outrageous security hole. The impact of this should be explained. It should NOT just be a copy-paste snippet.

If you take into account this only applies to the tablet endpoints where no sensitive data would ever be passed through anyway, there's really no security hole by doing this. There's absolutely no other way to make these endpoints read/write from user mode either than changing the group as well, but in my experience its even more work to make sure that the user is in the right group too (we fought with this for like 3 releases).

> Is it required for me to have to clone a separate repo and do admin-level tasks to have this work? I mean sure, it's linux, I'm fine with tinkering, but I don't really think we want to be getting user complaints about this. It should at least present a reasonable error that links to this FAQ, I think?

In terms of making it so this is usable out of the box, udev rules can be included in AppImages as well (where you need root to install anyway) but there's no way around udev rules even with user mode drivers. Windows has the equivalent of this permission level by default, so I don't see the issue.

bdach:
> There's absolutely no other way to make these endpoints read/write from user mode either than changing the group as well, but in my experience its even more work to make sure that the user is in the right group too (we fought with this for like 3 releases).

Okay, well, that makes it slightly more palatable, but still, I think this should be explained better in that FAQ, rather than just leaving a copy-paste snippet that looks a bit dodgy. Yes, I know, some people won't care, but the average linux user is more likely to care on average (I think?)

> In terms of making it so this is usable out of the box, udev rules can be included in AppImages as well

That would be good, but...

> (where you need root to install anyway)

This is news to me. I thought appimages were download-and-run? I for sure have not given the appimage releases root when I ran them. All I ever did was download and `$ chmod +x`.

InfinityGhost:
> Okay, well, that makes it slightly more palatable, but still, I think this should be explained better in that FAQ, rather than just leaving a copy-paste snippet that looks a bit dodgy. Yes, I know, some people won't care, but the average linux user is more likely to care on average (I think?)

I probably should note that, I don't maintain that FAQ as frequently as I should.

> > (where you need root to install anyway)
> 
> This is news to me. I thought appimages were download-and-run? I for sure have not given the appimage releases root when I ran them. All I ever did was download and `$ chmod +x`.

To be fair I've never actually used the AppImage myself because I just compile from source. The lack of built in tablet support (without following that FAQ) would have to be a compromise for portability since there is no other way to get those endpoints r/w at user mode. I figured most people used the AppImages like Debian users install `.deb`s but in a distribution agnostic way.

bdach:
In the interest of helping this along, I've pushed in the low-hanging button handling batch of @frenzibyte's suggested changes. (It would have been more helpful to link a branch to cherrypick from rather than list off what's broken as done above, but who am I to judge.)

I haven't made the `ISourcedFromTablet` changes yet, because they're (a) more involved, and (b) I'm not convinced it needs to be a separate thing from `ISourcedFromTouch`. Might look at it at a later point, but probably not on a weekday evening.

smoogipoo:
SDL itself also lists a similar thing for joystick support: https://github.com/libsdl-org/SDL/blob/main/docs/README-linux.md#joystick-does-not-work

Quite shocking, to be honest.

smoogipoo:
Does this not handle devices being connected? It seems to respond to tablets being disconnected by printing `Device: Device disconnected`, but not to being connected.

Also can't get my Wacom CTH-490 tablet to work. It detects and loads it but doesn't work well beyond that:
```
[runtime] 2021-03-11 03:09:59 [verbose]: Detect: Using device 'Intuos PTS'.
[runtime] 2021-03-11 03:09:59 [verbose]: Detect: Using report parser type 'OpenTabletDriver.Vendors.Wacom.IntuosV2ReportParser'.
[runtime] 2021-03-11 03:09:59 [verbose]: Detect: Device path: /sys/devices/pci0000:00/0000:00:08.1/0000:0e:00.3/usb5/5-1/5-1:1.0/0003:056A:033C.0001/hidraw/hidraw0
[runtime] 2021-03-11 03:09:59 [verbose]: Device: Set tablet feature: 02-02
[runtime] 2021-03-11 03:09:59 [verbose]: Detect: Found tablet 'Wacom CTH-490'
[runtime] 2021-03-11 03:09:59 [verbose]: Detect: Using device 'Intuos PTS'.
[runtime] 2021-03-11 03:09:59 [verbose]: Detect: Using auxiliary report parser type 'AuxReportParser'.
[runtime] 2021-03-11 03:09:59 [verbose]: Detect: Device path: /sys/devices/pci0000:00/0000:00:08.1/0000:0e:00.3/usb5/5-1/5-1:1.1/0003:056A:033C.0002/hidraw/hidraw1
```
Firstly, I had to disable the `wacom` kernel module ([input-wacom](https://github.com/linuxwacom/input-wacom)) which is installed by default on Arch Linux kernels, as it was sending relative input events via the `MouseHandler`. Secondly only the aux buttons work - the pen or pen buttons don't.

Works with my XP-Pen G640 out-of-the-box, though I can't easily test pen buttons because I've taken the buttons out of my pen and attached a custom grip >_>

peppy:
FWIW i still have intentions to fix  (or at least hint the user in the right direction to fix) the feedback scenario, because that's going to be present across all operating systems. That will come post merge of this, though.

InfinityGhost:
> Does this not handle devices being connected? It seems to respond to tablets being disconnected by printing `Device: Device disconnected`, but not to being connected.
> 
> Also can't get my Wacom CTH-490 tablet to work. It detects and loads it but doesn't work well beyond that:
> 
> ```
> [runtime] 2021-03-11 03:09:59 [verbose]: Detect: Using device 'Intuos PTS'.
> [runtime] 2021-03-11 03:09:59 [verbose]: Detect: Using report parser type 'OpenTabletDriver.Vendors.Wacom.IntuosV2ReportParser'.
> [runtime] 2021-03-11 03:09:59 [verbose]: Detect: Device path: /sys/devices/pci0000:00/0000:00:08.1/0000:0e:00.3/usb5/5-1/5-1:1.0/0003:056A:033C.0001/hidraw/hidraw0
> [runtime] 2021-03-11 03:09:59 [verbose]: Device: Set tablet feature: 02-02
> [runtime] 2021-03-11 03:09:59 [verbose]: Detect: Found tablet 'Wacom CTH-490'
> [runtime] 2021-03-11 03:09:59 [verbose]: Detect: Using device 'Intuos PTS'.
> [runtime] 2021-03-11 03:09:59 [verbose]: Detect: Using auxiliary report parser type 'AuxReportParser'.
> [runtime] 2021-03-11 03:09:59 [verbose]: Detect: Device path: /sys/devices/pci0000:00/0000:00:08.1/0000:0e:00.3/usb5/5-1/5-1:1.1/0003:056A:033C.0002/hidraw/hidraw1
> ```
> 
> Firstly, I had to disable the `wacom` kernel module ([input-wacom](https://github.com/linuxwacom/input-wacom)) which is installed by default on Arch Linux kernels, as it was sending relative input events via the `MouseHandler`. Secondly only the aux buttons work - the pen or pen buttons don't.
> 
> Works with my XP-Pen G640 out-of-the-box, though I can't easily test pen buttons because I've taken the buttons out of my pen and attached a custom grip >_>

Ah right, I released a hotfix for this bug, I forgot to whitelist a report parser. Will push the update that fixes this.

smoogipoo:
I've added the new input events to our wiki: https://github.com/ppy/osu-framework/wiki/Handling-input-events

smoogipoo:
Is the wacom tablet supposed to be working after that recent change? Still isn't, doesn't seem like the console output changed if that was to be expected:
```
[runtime] 2021-03-11 05:26:35 [verbose]: Detect: Using device 'Intuos PTS'.
[runtime] 2021-03-11 05:26:35 [verbose]: Detect: Using report parser type 'OpenTabletDriver.Vendors.Wacom.IntuosV2ReportParser'.
[runtime] 2021-03-11 05:26:35 [verbose]: Detect: Device path: /sys/devices/pci0000:00/0000:00:08.1/0000:0e:00.3/usb5/5-1/5-1:1.0/0003:056A:033C.000D/hidraw/hidraw9
[runtime] 2021-03-11 05:26:35 [verbose]: Device: Set tablet feature: 02-02
[runtime] 2021-03-11 05:26:35 [verbose]: Detect: Found tablet 'Wacom CTH-490'
[runtime] 2021-03-11 05:26:35 [verbose]: Detect: Using device 'Intuos PTS'.
[runtime] 2021-03-11 05:26:35 [verbose]: Detect: Using auxiliary report parser type 'AuxReportParser'.
[runtime] 2021-03-11 05:26:35 [verbose]: Detect: Device path: /sys/devices/pci0000:00/0000:00:08.1/0000:0e:00.3/usb5/5-1/5-1:1.1/0003:056A:033C.000E/hidraw/hidraw10
```

InfinityGhost:
> Is the wacom tablet supposed to be working after that recent change? Still isn't, doesn't seem like the console output changed if that was to be expected:
> 
> ```
> [runtime] 2021-03-11 05:26:35 [verbose]: Detect: Using device 'Intuos PTS'.
> [runtime] 2021-03-11 05:26:35 [verbose]: Detect: Using report parser type 'OpenTabletDriver.Vendors.Wacom.IntuosV2ReportParser'.
> [runtime] 2021-03-11 05:26:35 [verbose]: Detect: Device path: /sys/devices/pci0000:00/0000:00:08.1/0000:0e:00.3/usb5/5-1/5-1:1.0/0003:056A:033C.000D/hidraw/hidraw9
> [runtime] 2021-03-11 05:26:35 [verbose]: Device: Set tablet feature: 02-02
> [runtime] 2021-03-11 05:26:35 [verbose]: Detect: Found tablet 'Wacom CTH-490'
> [runtime] 2021-03-11 05:26:35 [verbose]: Detect: Using device 'Intuos PTS'.
> [runtime] 2021-03-11 05:26:35 [verbose]: Detect: Using auxiliary report parser type 'AuxReportParser'.
> [runtime] 2021-03-11 05:26:35 [verbose]: Detect: Device path: /sys/devices/pci0000:00/0000:00:08.1/0000:0e:00.3/usb5/5-1/5-1:1.1/0003:056A:033C.000E/hidraw/hidraw10
> ```

After doing testing on my own setup, pen buttons and aux buttons don't work in this current state either. Does positioning at least work? I think something is awry with the report parser, but I'm not 100% sure. Is data being passed through at the `tabletDriver.ReportReceived`event handler?

smoogipoo:
No, only the aux buttons work for me. Positioning and pen buttons don't work. Only `AuxReport` events are coming through `ReportReceived`.

X9VoiD:
There was a problem with the new IntuosV2 parser where tablets send an unexpected byte. A hotfix was done to fix it but it seems even on the same model that first byte changes too. If needed be we may add a LegacyIntuosV2 which is known to work but don't support the extra functionalities of the new parser (eraser state, and tilt, etc.) until the issue is fixed. On osu!'s use case the new parser is not required so it'll be fine to do that I guess.

smoogipoo (dismissed):
I think this is fine to go in as-is. The code looks very solid, though no doubt I've missed some little thing somewhere as is usually the case with new input handlers >_>

Only remaining issues:
1. CTH-490 not working (see comments).
2. Newly-connected tablets don't detect. Perhaps this is due to performance? vv
3. Startup performance (300ms with tablet plugged in, 30ms without). I personally think this is fine but...

... but all of these issues can be worked on in the future / subsequent PRs.

@peppy requesting another pass from you.

smoogipoo (dismissed):

peppy:
@InfinityGhost can you confirm intended behaviour of detecting tablets that are connected during runtime? is this not supported yet, or a bug?

bdach:
Pen buttons not working might also be an issue with the `ISourcedFromTablet` thing @frenzibyte mentioned? By which I mean button presses might need translating to synthetic mouse events, similarly to the way touchscreen touches are?

smoogipoo:
@bdach no, it's just that any pen events (movement, pressure, button presses) are simply not arriving. We're already synthesising correctly as far as I can tell (and it works with my other tablet).

In-so-far as the difference in synthesising between tablet and touch, I think that's also fine for the time being. It's probably something that can be done if there's a use for it, but probably also needs knowledge of tablet capabilities (like whether a screen is in use?), otherwise tablet and mouse are typically treated the same way by most applications.

AbstractQbit:
@smoogipoo cth-490 not working is a regression on our end, will get fixed when OpenTabletDriver/OpenTabletDriver#850 is merged, sorry for the confusion.

Regarding the startup performance, it's probably possible to narrow down the list of configs to be parsed via an index on VID and PID. And in case of Huion/Gaomon we can also cache the 201st string descriptor, so that it's not requested from device each time another config is probed.

InfinityGhost:
> @InfinityGhost can you confirm intended behaviour of detecting tablets that are connected during runtime? is this not supported yet, or a bug?

It's supported and in all of my testing fully functional. Seems like its potentially a bug within [my fork of HidSharp](https://github.com/InfinityGhost/HIDSharpCore).

peppy (dismissed):

smoogipoo:
@InfinityGhost is the "not detecting newly-connected tablets" issue not due to the condition here?
https://github.com/ppy/osu-framework/blob/fca3bf21d54e67b0d88a47e52b2bfbf933d43be0/osu.Framework/Input/Handlers/Tablet/TabletDriver.cs#L26-L30

peppy:
Looks about right. Tablet will be non-null at that point in all cases I believe? Unless disconnecting causes it to null?

Also important to note that the handler currently only looks to handle one device (the first it encounters) from what I can tell. Something to improve upon for sure.

peppy:
I changed the threading logic to handle a potential bug OTD's side (reported https://github.com/OpenTabletDriver/OpenTabletDriver/issues/857). @smoogipoo probably best you check for correctness before merging.

peppy (approved):

peppy:
@InfinityGhost How does OTD handle multiple tablets connected at once? I'm just getting around to implementing configuration of table/screen mappings and noticed that there is an event `TabletChanged` which seems to imply only one tablet can be tracked at once.

I know for a fact that there are scenarios where a user will have more than one tablets connected to their PC and expect to be able to configure and use each one simultaneously. Is that currently possible?

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
```diff
@@ -0,0 +1,145 @@
...
+                        Position = new Vector2(outputWidth / 2, outputHeight / 2)
```
peppy, line 86: Is this position value ever actually being used? I'm a bit confused why it's being specified as the centre point rather than top-left..

unless this is how your `Area` data type works..? if so it DEFINITELY needs documentation specifying this, as it's petty non-standard.. as it's pretty non-standard

adryzz: The position is needed in OTD with multi-monitor setups where you assign your tablet to a secondary monitor since the offset is relative to the "virtual screen".
(or for rotation/scaling too)

peppy: Confirm it is the "center" of the area? Because that seems really weird and needs xmldoc or something...

InfinityGhost: (I forgot to add xmldocs for a lot of stuff) but yes it is the center of the area. It makes things much more simple when it comes to rotations, since we just use that point as the origin of rotation as well so there is no confusion there.

InfinityGhost:
> @InfinityGhost How does OTD handle multiple tablets connected at once? I'm just getting around to implementing configuration of table/screen mappings and noticed that there is an event `TabletChanged` which seems to imply only one tablet can be tracked at once.
> 
> I know for a fact that there are scenarios where a user will have more than one tablets connected to their PC and expect to be able to configure and use each one simultaneously. Is that currently possible?

As it currently stands, handling of multiple concurrent tablets has been debated for a while so we haven't added it in. Since different tablets can have different dimensions and bounds, not all input areas can be a "one size fits all" thing. `AbsoluteOutputMode` is also limited by its single input and output area. We plan on rewriting the entire output pipeline anyway to resolve issues like this while also making it more flexible.

tl;dr the first one in the configuration list will be the only one detected and configured, its certainly not ideal but at the very least will always be consistent until we figure something there.

peppy:
So if you have two tablets and disconnect one, the other will immediately take over? Is the order consistent, or does it depend on the order connected to the system?

adryzz:
All the tablet configurations' json files are loaded with `DirectoryInfo.EnumerateFiles("*.json")`, so the order is alphabetical

InfinityGhost:
> So if you have two tablets and disconnect one, the other will immediately take over?

It would, but only if the logic triggers redetection on removal as well. We only trigger detection for device additions currently.

> Is the order consistent, or does it depend on the order connected to the system?

It's in the order of the configurations in which the `Assembly.GetManifestResourceNames()` request reports, which is always in alphabetical order.

> All the tablet configurations' json files are loaded with `DirectoryInfo.EnumerateFiles("*.json")`, so the order is alphabetical

This is only in the case of the desktop OpenTabletDriver implementation, but it is basically the same.

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
```diff
@@ -0,0 +1,85 @@
...
+            };
+        }
+
+        private readonly object detectLock = new object();
+
+        private CancellationTokenSource cancellationSource;
+
+        public void DetectTablet()
+        {
+            lock (detectLock)
```
smoogipoo, line 41: I'm very confused by this lock. What's this intending to lock? Surely it's not intending to lock anything inside the `Task.Run()`, so it's used for the cancellation token? In which case how about rewriting this as:
```
cancellationSource?.Cancel();
var localCancellation = cancellationSource = new CancellationTokenSource();

Task.Run(async () =>
{
    // wait a small delay as multiple devices may appear over a very short interval.
    await Task.Delay(50, localCancellation.Token).ConfigureAwait(false);
}, localCancellation.Token);
```
Or am I missing something?

peppy: Yep, specifically there for the cancellation.

Couldn't two threads arrive at the second line of your code at the same time and one task never get cancelled? Or am I missing something?

smoogipoo: That's true, I didn't know whether this was possible at OTD's end. I'll take this as is then.

PR #4163, opened 2021-01-21
Fix dependency injection potentially resolving dependency to the declaring instance

frenzibyte:
Found while doing some work lazer-side.

The issue is that on certain component patterns like the one below, `Foo` may be cached before its own dependencies are resolved, in that case, the `parent` dependency will be resolved to the same declaring instance (`Foo`).
```csharp
[Cached]
public FooType Foo = new FooType();

public class FooType
{
    // expected: some parent instance from ancestor composites.
    // actual: this class's instance.
    [Resolved(CanBeNull = true)]
    private FooType parent { get; set; }
}
```

Therefore I've fixed the issue by adding a parameter that accepts a list of instances to exclude while searching for the dependency in the container, and let both `ResolvedAttribute` and `BackgroundDependencyLoaderAttribute` exclude the declaring instance from the searching process.

### DIFF
--- a/osu.Framework/Allocation/IReadOnlyDependencyContainer.cs
+++ b/osu.Framework/Allocation/IReadOnlyDependencyContainer.cs
@@ -23,8 +23,9 @@ namespace osu.Framework.Allocation
         /// </summary>
         /// <param name="type">The dependency type to query for.</param>
         /// <param name="info">Extra information that identifies the cached dependency.</param>
+        /// <param name="exclusion">An optional array of instances to exclude while searching.</param>
         /// <returns>The requested dependency, or null if not found.</returns>
-        object Get(Type type, CacheInfo info);
+        object Get(Type type, CacheInfo info, object[] exclusion = null);
 
         /// <summary>
         /// Injects dependencies into the given instance.
--- a/osu.Framework.Tests/Dependencies/CachedModelDependenciesTest.cs
+++ b/osu.Framework.Tests/Dependencies/CachedModelDependenciesTest.cs
@@ -265,6 +265,26 @@ namespace osu.Framework.Tests.Dependencies
             Assert.AreEqual(null, resolver.BindableString.Value);
         }
 
+        [Test]
+        public void TestResolveFromInsideCachedContainer()
+        {
+            var parent = new SameTypeResolver();
+            var resolver = new SameTypeResolver();
+
+            var parentDependencies = new DependencyContainer();
+            parentDependencies.Cache(parent);
+
+            var modelDependencies = new CachedModelDependencyContainer<FieldModel>(parentDependencies);
+            modelDependencies.Model.Value = new FieldModel();
+
+            var dependencies = new DependencyContainer(modelDependencies);
+            dependencies.Cache(resolver);
+
+            dependencies.Inject(resolver);
+
+            Assert.AreSame(parent, resolver.Parent);
+        }
+
         private class NonBindablePublicFieldModel
         {
 #pragma warning disable 649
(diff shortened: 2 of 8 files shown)


osu.Framework/Allocation/IReadOnlyDependencyContainer.cs
```diff
@@ -23,8 +23,9 @@ public interface IReadOnlyDependencyContainer
         /// </summary>
         /// <param name="type">The dependency type to query for.</param>
         /// <param name="info">Extra information that identifies the cached dependency.</param>
+        /// <param name="exclusion">An optional array of instances to exclude while searching.</param>
         /// <returns>The requested dependency, or null if not found.</returns>
-        object Get(Type type, CacheInfo info);
+        object Get(Type type, CacheInfo info, object[] exclusion = null);
```
bdach, line 28: Not sure about why this is an array (lots of added allocs on everyone that's calling this - at least they're sort of one-time and not every update frame), and not sure why this only exists on this one method when others also potentially care about this.

Not to mention that this possibly shouldn't be outwardly exposed on a public interface, but I'm less sure on that one. Could be useful in some scenarios, but maybe it *shouldn't* be useful as that indicates misuse of the framework...

smoogipoo:
Is this actually an issue with `Drawable`? iirc `Drawable` always injects the parent dependencies into itself, not its own dependencies into itself. So that is `TestResolveFromInsideCachedContainer` is not replicating the draw hierarchy because it's doing `dependencies.Inject(receiver);` rather than `parentDependencies.Inject(receiver);`.

Can you give a more concrete example of where you found this in lazer?

frenzibyte:
Yeah and parent dependencies is the container that contains the cached `Foo` instance, perhaps the tests were misleading, will fix them

smoogipoo:
So it would be the following scenario?
```csharp
public class Bar : CompositeDrawable
{
    [Cached]
    private readonly Foo foo = new Foo();
}

public class Foo : Drawable
{
    [Resolved]
    private Foo foo { get;set; } // == Bar.foo
}
```
I'm not sure this is that unexpected...

frenzibyte:
Well, at least you don't except a class's dependency gets resolved to itself, in any way possible

bdach:
> Well, at least you don't except a class's dependency gets resolved to itself, in any way possible

This comment doesn't parse as English to me, but I also do think it's a bit pointless to allow a drawable to resolve itself from its parent. I imagine this might be a valid-ish use case in the case of a nested comment tree, for example - say you want to resolve the parent comment which is in a container, so you do the normal setup, and end up getting the same instance that is doing the resolving.

But to reiterate I'm not sure this current proposal is the way forward here, the API seems haphazard.

frenzibyte:
I'm gonna leave this here for now as I don't need it currently, I'll come back to it and think of a proper API later, if no one beats me to it.

smoogipoo:
I'm not discussing whether it's pointless or not but whether it's a strictly wrong behaviour, which I don't think it is.

DI is a top-down process - a drawable receives all dependencies cached by its parents recursively, and none of its own. This change adds another condition to that, which is that if its parent caches the child then the child doesn't receive itself as a dependency.  
How far does this go? What if a parent two levels up caches the child, should the child still receive itself as a dependency? This doesn't handle that case but I disagree that it should in the first place.

bdach:
Think you've turned my opinion upside down with that comment, and I'm no longer that convinced this is a good change. Seems like a bad precedent that complicates a simple mechanism when you put it that way.

frenzibyte:
Oh, I actually forgot to propagate the exclusion array all the way up, but either way yeah actually this looks like a bad idea, better throw an exception such that consumers don't think this is a bug, I'll close this for now as I don't need it for my work anyways.

PR #4165, opened 2021-01-22
Fix key combination display ordering

### LINKED ISSUE #4153: Shift is before ctrl in KeyCombination's ReadableString
osu!_Si6KfpAKHO

vinaqq:
Fixes #4153
 
- Introduce OrderAttribute from osu.Game
- To avoid adding an order to all `InputKeys`,  an `AllowPartialOrdering` flag was added on the `HasOrderedElements` to allow for only certain member to define a custom order as the current implementation of `GetValuesInOrder` would throw an exception. For now when a member does not define the OrderAttribute it will default back to the Enum value to keep ordering.
- Added simple tests to verify correct ordering of KeyCombination 

Unsure what to do with the current osu.Game `OrderAttribute`, removing it would introduce a dependency on this change ? Please let me know if I have to do anything else.

### DIFF
--- a/osu.Framework/Extensions/EnumExtensions/OrderedEnumExtensions.cs
+++ b/osu.Framework/Extensions/EnumExtensions/OrderedEnumExtensions.cs
@@ -1,0 +1,57 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using osu.Framework.Utils;
+
+namespace osu.Framework.Extensions.EnumExtensions
+{
+    public static class OrderedEnumExtensions
+    {
+        /// <summary>
+        /// Get values of an enum in order. Supports custom ordering via <see cref="OrderAttribute"/>.
+        /// </summary>
+        public static IEnumerable<T> GetValuesInOrder<T>()
+            where T : struct, Enum
+        {
+            var type = typeof(T);
+
+            if (!type.IsEnum)
+                throw new InvalidOperationException($"{typeof(T)} must be an enum");
+
+            IEnumerable<T> items = (T[])Enum.GetValues(type);
+
+            return GetValuesInOrder(items);
+        }
+
+        /// <summary>
+        /// Get values of a collection of enum values in order. Supports custom ordering via <see cref="OrderAttribute"/>.
+        /// </summary>
+        public static IEnumerable<T> GetValuesInOrder<T>(this IEnumerable<T> items)
+            where T : struct, Enum
+        {
+            var type = typeof(T);
+
+            if (!type.IsEnum)
+                throw new InvalidOperationException($"{typeof(T)} must be an enum");
+
+            if (!(Attribute.GetCustomAttribute(type, typeof(HasOrderedElementsAttribute)) is HasOrderedElementsAttribute orderedAttr))
+                return items;
+
+            return items.OrderBy(i =>
+            {
+                var fieldInfo = type.GetField(i.ToString());
+
+                if (fieldInfo?.GetCustomAttributes(typeof(OrderAttribute), false).FirstOrDefault() is OrderAttribute attr)
+                    return attr.Order;
+
+                if (orderedAttr.AllowPartialOrdering)
+                    return Convert.ToInt32(i);
+
+                throw new ArgumentException($"Not all values of {typeof(T)} have {nameof(OrderAttribute)} specified.");
+            });
+        }
+    }
+}
--- a/osu.Framework/Input/Bindings/KeyCombination.cs
+++ b/osu.Framework/Input/Bindings/KeyCombination.cs
@@ -142,11 +142,7 @@ namespace osu.Framework.Input.Bindings
         /// Get a string representation can be used with <see cref="KeyCombination(string)"/>.
         /// </summary>
         /// <returns>The string representation.</returns>
-        public override string ToString()
-        {
-            var sortedKeys = OrderAttributeUtils.GetValuesInOrder(Keys);
-            return string.Join(',', sortedKeys.Select(k => (int)k));
-        }
+        public override string ToString() => string.Join(',', Keys.Select(k => (int)k));
 
         public string ReadableString()
         {
--- a/osu.Framework/Input/Bindings/KeyCombination.cs
+++ b/osu.Framework/Input/Bindings/KeyCombination.cs
@@ -7,8 +7,8 @@ using System.Collections.Immutable;
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Input.States;
-using osu.Framework.Utils;
 using osuTK;
 using osuTK.Input;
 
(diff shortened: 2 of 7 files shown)


osu.Framework/Input/Bindings/KeyCombination.cs
```diff
@@ -32,7 +33,7 @@ namespace osu.Framework.Input.Bindings
         /// <remarks>This constructor is not optimized. Hot paths are assumed to use <see cref="FromInputState(InputState, Vector2?)"/>.</remarks>
         public KeyCombination(IEnumerable<InputKey> keys)
         {
-            Keys = keys?.Any() == true ? keys.Distinct().OrderBy(k => (int)k).ToImmutableArray() : none;
+            Keys = keys?.Any() == true ? OrderAttributeUtils.GetValuesInOrder(keys.Distinct()).ToImmutableArray() : none;
```
bdach, line 36: I don't think this is a good spot to put this sort in, for several reasons:

* The `OrderAttribute` sorting logic heavily uses reflection, so I don't know that it is a good idea from a performance standpoint to be calling that logic every time a key combination is constructed.
* Moreover, `KeyCombination.FromInputState()`, which is mentioned in the xmldoc to this method (and is the optimised path), now sorts in a different way to this. Seeing that `KeyCombination.Equals()` uses sequence equality, I imagine this breaks things.

If you ask me this sort should *only* be done locally for display purposes in `ToString()`/`ReadableString()`.

osu.Framework/Utils/OrderAttribute.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        /// <summary>
+        /// Get values of an enum in order. Supports custom ordering via <see cref="OrderAttribute"/>.
+        /// </summary>
+        public static IEnumerable<T> GetValuesInOrder<T>()
+        {
+            var type = typeof(T);
+
+            if (!type.IsEnum)
+                throw new InvalidOperationException("T must be an enum");
```
bdach, line 20: I know it was like this game-side, but it was arguably unhelpful/wrong - comments to this file are just fixes that we might as well go ahead with while we're moving things around

```suggestion
                throw new InvalidOperationException($"{typeof(T)} must be an enum");
```

osu.Framework/Utils/OrderAttribute.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        /// <summary>
+        /// Get values of a collection of enum values in order. Supports custom ordering via <see cref="OrderAttribute"/>.
+        /// </summary>
+        public static IEnumerable<T> GetValuesInOrder<T>(IEnumerable<T> items)
+        {
+            var type = typeof(T);
+
+            if (!type.IsEnum)
+                throw new InvalidOperationException("T must be an enum");
```
bdach, line 35:
```suggestion
            if (!type.IsEnum)
                throw new InvalidOperationException($"{typeof(T)} must be an enum");
```

osu.Framework/Utils/OrderAttribute.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            return items.OrderBy(i =>
+            {
+                if (type.GetField(i.ToString())?.GetCustomAttributes(typeof(OrderAttribute), false).FirstOrDefault() is OrderAttribute attr)
+                    return attr.Order;
+
+                if (orderedAttr.AllowPartialOrdering)
+                    return (int)Enum.Parse(type, i.ToString());
+
+                throw new ArgumentException($"Not all values of {nameof(T)} have {nameof(OrderAttribute)} specified.");
```
bdach, line 48:
```suggestion
                throw new ArgumentException($"Not all values of {typeof(T)} have {nameof(OrderAttribute)} specified.");
```

`nameof` is compile-time and does not work with generics

osu.Framework/Utils/OrderAttribute.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+    }
+
+    [AttributeUsage(AttributeTargets.Enum)]
+    public class HasOrderedElementsAttribute : Attribute
+    {
+        /// <summary>
+        /// Allow for partially ordering Enum members.
+        /// Members without an Order Attribute will default to their value as ordering key.
+        /// </summary>
```
bdach, line 70:
```suggestion
        /// <summary>
        /// Allow for partially ordering <see cref="Enum"/> members.
        /// Members without an <see cref="OrderAttribute"/> will default to their value as ordering key.
        /// </summary>
```

bdach (changes requested):

bdach:
Oh and needless to say, if this goes in, the game-side class should be removed.

osu.Framework/Input/Bindings/KeyCombination.cs
vinaqq: Thank you for the feedback @bdach. 
As you suggested I moved the sorting to ToString() and ReadableString().

bdach:
I made a mistake in suggesting that the sort should be applied to `ToString()` - that would be a breaking change. It should only be in `ReadableString()`; I'll fix that now.

(I'll also fix the tests since they aren't really asserting that this change works, more like asserting that the keys were sorted. Therefore I'm pretty sure that test as written would also pass on `master`.)

bdach (dismissed):

osu.Framework/Extensions/EnumExtensions/OrderedEnumExtensions.cs
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using osu.Framework.Utils;
+
+namespace osu.Framework.Extensions.EnumExtensions
```
peppy, line 9: Is there a reason we put these in their own namespace (this isn't the only case)? Seems pointless when most are single file implementations.

osu.Framework/Extensions/EnumExtensions/OrderedEnumExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    public static class OrderedEnumExtensions
```
peppy, line 11: Shouldn't this just be `EnumExtensions`? I think that's generally the standard naming schema for extension methods.

osu.Framework/Utils/OrderAttribute.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+
+namespace osu.Framework.Utils
+{
+    [AttributeUsage(AttributeTargets.Field)]
+    public class OrderAttribute : Attribute
```
peppy, line 9: probably deserves xmldoc. i know that `Order` attributes exist in at *least* one other place in database abstractions (or dotnet itself?) so it would be good to know why we have our own (and when it should be used).

peppy (changes requested):

osu.Framework/Extensions/EnumExtensions/OrderedEnumExtensions.cs
bdach: I don't know, I mostly did this to follow precedent.

peppy (approved):

PR #4166, opened 2021-01-23
Expose IWindow.Resized event

InfinityGhost:
Needed in #4162

### DIFF
--- a/osu.Framework/Platform/OsuTKWindow.cs
+++ b/osu.Framework/Platform/OsuTKWindow.cs
@@ -41,6 +41,11 @@ namespace osu.Framework.Platform
         public event Action Exited;
 
         /// <summary>
+        /// Invoked when the <see cref="OsuTKWindow"/> has resized.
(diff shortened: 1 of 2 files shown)


osu.Framework/Platform/OsuTKWindow.cs
```diff
@@ -106,7 +111,11 @@ protected OsuTKWindow([NotNull] IGameWindow osuTKGameWindow)
             // Moving or resizing the window needs to check to see if we've moved to a different display.
             // This will update the CurrentDisplay bindable.
             Move += (sender, e) => checkCurrentDisplay();
-            Resize += (sender, e) => checkCurrentDisplay();
+            Resize += (sender, e) =>
```
peppy, line 114: We don't actually use `OsuTKWindow` any more so I'm not sure how useful implementing this is here, but I guess it's good to have.

peppy (approved):

PR #2467, opened 2019-05-31
Remove placeholders and simplify ModelBackedDrawable

smoogipoo:
As evidenced by my previous attempts, getting placeholders to work in all edge cases is extremely difficult.
As a departure from those attempts I've removed the placeholder, in exchange for simplicity and what I think is a fully defined set of rules.

1. If `TransformImmediately` is `true`, it will always run `ApplyHideTransforms` on the currently-displayed model when a new model is set.
2. it displays the result of `CreateDrawable(Model)`.
3. `CreateDrawable(null)` restores the functionality of `UpdateableBeatmapBackgroundSprite`, which displays the default background in this case.
4. `CreateDrawable()` can also return null if nothing is to be displayed, in which case it'll act as if `TransformImmediately` is `true`.

Along with breaking changes from https://github.com/ppy/osu-framework/pull/2466 :

1. `ReplaceDrawable()` has been split into `ApplyHideTransforms()` and `ApplyShowTransforms()`.
2. `CreateDelayedLoadWrapper()` now handles recreation of content internally, removing the need for code such as https://github.com/ppy/osu/blob/master/osu.Game/Beatmaps/Drawables/UpdateableBeatmapBackgroundSprite.cs#L39 in the case of using an unload wrapper.

I've also added a new test scene `TestSceneModelBackedDrawableWithLoading` to show use-cases where you may want to display a loading animation. This is achieved through the new `OnLoadStarted` and `OnLoadFinished` methods.

PR #4167, opened 2021-01-23
Fix ModelBackedDrawable not applying transforms on reloaded wrapper

### LINKED ISSUE #3318: ModelBackedDrawable doesn't play fade in transform after unload
- Let modelbacked load
- Unload via offscreen / masked away
- Load a second time
- Fade in will not play

Currently no test coverage.

image

frenzibyte:
Closes #3318

Previously if `finishLoad(wrapper)` is called in the case of a DLUW reload, the method will early-return without applying fade in transforms to the wrapper (as the wrapper would not change on reload). 

This PR changes it to do so, now it'll always hide and apply show transforms when a drawable is loaded/reloaded, and if there is a previous displayed drawable different from the new one, then hide and expire it.

Also added proper test coverage for unloading MBDs, along with a [failing test case which this PR fixes](https://github.com/ppy/osu-framework/blob/7ec51b3486d306f362e5011d8b1eb6108fe0819d/osu.Framework.Tests/Visual/Drawables/TestSceneUnloadingModelBackedDrawable.cs#L72-L85).

---

I was initially going to inject test coverage into `TestSceneModelBackedDrawable` but realized it's actually completely unfriendly with drawable unloading and reloading (each created `TestModel` provides a drawable, instead of the other direction).

Was gonna attempt to refactor that test scene into being friendly but I realized it's just a waste of time instead of writing a new fresh test scene with DLUW in mind.


osu.Framework/Graphics/Containers/ModelBackedDrawable.cs
```diff
@@ -138,27 +138,25 @@ private void loadDrawable(Func<Drawable> createDrawableFunc)
         /// <param name="wrapper">The <see cref="DelayedLoadWrapper"/>.</param>
         private void finishLoad(DelayedLoadWrapper wrapper)
         {
-            var lastWrapper = displayedWrapper;
-
-            // If the wrapper hasn't changed then this invocation must be a result of a reload (e.g. DelayedLoadUnloadWrapper)
-            // In this case, we do not want to transform/expire the wrapper
```
bdach, line 144: By the comment it looks as if this was explicitly stated as intended. This PR has no explanation as to why you would need to change this and whether removing this special case doesn't break some existing use cases.

frenzibyte: Thought it was unnecessary given that really this PR is just applying a show transform to a drawable, but that's fair I suppose.

The second line of that comment indicated to me that the wrapper which is currently displayed shouldn't get hidden and expire in the case of a DLUW reload, but seems it's bad of me not to discuss it. 

@smoogipoo as you're the writer of this comment (in https://github.com/ppy/osu-framework/pull/2467), did you intend that drawables reloaded via DLUW must not have show transforms applied to them?

As for existing use cases, there is really one use case that is suffering from this issue `UpateableBeatmapBackgroundSprite`, there's another MBD+DLUW usage but for some reason it's applying the transform from inside the content drawable, ignoring MBD's feature of doing so.

smoogipoo: I don't remember the case that led to my comment there. It's possible that the internals of DLUW have changed so much over time that the comment can be invalidated.

smoogipoo: Well the reason for this is MBD + DLUW inside scrolling containers. But I think this makes sense - all usages of raw DLUW already have a `d.OnLoadComplete += d => d.FadeInFromZero(X)`.

osu.Framework/Graphics/Containers/ModelBackedDrawable.cs
```diff
@@ -138,27 +138,25 @@ private void loadDrawable(Func<Drawable> createDrawableFunc)
...
             wrapper?.FinishTransforms();
 
             var showTransforms = ApplyShowTransforms(wrapper);
 
-            // If we have a non-null new wrapper, we need to wait for the show transformation to complete before hiding the old wrapper,
-            // otherwise, we can hide the old wrapper instantaneously and leave a blank display
-            var hideTransforms = wrapper == null
-                ? ApplyHideTransforms(lastWrapper)
-                : ((Drawable)lastWrapper)?.Delay(TransformDuration)?.Append(ApplyHideTransforms);
+            if (displayedWrapper != null && displayedWrapper != wrapper)
```
smoogipoo, line 147: Please add the comment back here, regarding expiry.

smoogipoo (changes requested):

osu.Framework.Tests/Visual/Drawables/TestSceneUnloadingModelBackedDrawable.cs
```diff
@@ -0,0 +1,151 @@
...
+        }
+
+        [Test]
+        public void TestChangeWhileMaskedAway()
+        {
+            AddStep("mask away", () => backedDrawable.Position = new Vector2(-1));
+            AddWaitStep("wait for unload", 1);
+
+            AddStep("change model", () => backedDrawable.Model = 5);
+            AddWaitStep("wait for potential load", 5);
```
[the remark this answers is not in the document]
bdach: On its own maybe it wouldn't be so weird, but mixed with the wait steps it sort of implies that there's a link or dependency when reading the test, while in fact there is none.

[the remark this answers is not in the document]
frenzibyte: Aha, I see.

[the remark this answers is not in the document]
frenzibyte: Should be better now, check it out.

smoogipoo (dismissed):
This looks fine imo. @peppy will await your review.

osu.Framework.Tests/Visual/Drawables/TestSceneUnloadingModelBackedDrawable.cs
```diff
@@ -0,0 +1,145 @@
...
+using osu.Framework.Graphics.Shapes;
+using osu.Framework.Graphics.Sprites;
+using osu.Framework.Graphics.Transforms;
+using osu.Framework.Testing;
+using osuTK;
+using osuTK.Graphics;
+
+namespace osu.Framework.Tests.Visual.Drawables
+{
+    [System.ComponentModel.Description("MBD behaviour with unloading wrapper")]
```
peppy, line 17: i really dislike using these arbitrary acronyms in code/comments.

smoogipoo (approved):

peppy (approved):

osu.Framework.Tests/Visual/Drawables/TestSceneUnloadingModelBackedDrawable.cs
bdach: I'd go a step further and say that these sorts of descriptions say about as much as the class names, if not less - so at that point you might as well make do without.

peppy: already gone :)

frenzibyte: We may want to do the same everywhere across the codebase, to avoid this from happening again later on. (as I wasn't going to use acronyms until I noticed they're already used in the project)

peppy: Go for it. The only other I could find immediately was DLUW, which i also disagree with.

PR #4168, opened 2021-01-23
Add ability to seed RNG class number generator

Game4all:
# Summary
This adds the ability to seed `RNG`'s  underlying random number generator through the `Seed()` static method (I was surprised to see it wasn't implemented at first when I looked up IntelliSense on `RNG` class to seed it).

The `Random` class from .NET Standard lib doesn't support changing the seed at runtime so the instance gets recreated on each manual seeding.

### DIFF
--- a/osu.Framework/Utils/RNG.cs
+++ b/osu.Framework/Utils/RNG.cs
@@ -10,8 +10,13 @@ namespace osu.Framework.Utils
     /// </summary>
     public static class RNG
     {
-        // Base RNG. Maybe expose methods for re-seeding in the future?
-        private static readonly Random random = new Random();
+        private static Random random = new Random();
+
+        /// <summary>
+        /// Seeds the random number generator with the specified number.
+        /// </summary>
+        /// <param name="seed">The number to use to seed the random number generator.</param>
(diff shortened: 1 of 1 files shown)


bdach:
What do you need this for?

I'm not sure I agree with doing things this way. Setting this seed will change random outputs of a whole lot of other things. If multiple components call this they might up fighting for the seed among themselves.

I'd just instantiate `Random` with a seed, if you ask me.

Game4all:
I was using the RNG class extensively in a project and needed to be able to set the generator seed to some value before the game boots up.

The intent was to be able to set the seed to a sane and known value before the game boots so I could've argued that the method could be guarded to allow calling it once during the whole execution of the program. 

Anyways the first point you raised is definitely a problem, after searching by refs the osu! codebase, turns out `RNG` is used in a lot  of places which even include gameplay components (so I can see the problem I guess)

As such, I'm closing this and going in a way of implementing a fix locally instead.

peppy:
You should be making your own instance of `Random` instead. If you need easy access (and don't care about the implicications) make it static at your end, as is done inside of `RNG`.

PR #4171, opened 2021-01-27
Stop handling unobserved exceptions

peppy:
We have generally agreed that these should be left unobserved, as it is quite common to fire a `Task.Run` without expecting it to cause the whole application to crash.

I've left logging in as that part will definitely be useful for tracking down edge cases where we actually do care about the exceptions.

Aims to fix issue observed in https://github.com/ppy/osu/issues/11608#issuecomment-767839973.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -244,7 +244,7 @@ namespace osu.Framework.Platform
         {
             var exception = (Exception)args.ExceptionObject;
 
-            logException(exception, "unobserved");
+            logException(exception, "unhandled");
             abortExecutionFromException(sender, exception);
         }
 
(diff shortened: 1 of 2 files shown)


osu.Framework/Platform/GameHost.cs
```diff
@@ -243,61 +243,72 @@ public void Collect()
         private void unhandledExceptionHandler(object sender, UnhandledExceptionEventArgs args)
         {
             var exception = (Exception)args.ExceptionObject;
-            exception.Data["unhandled"] = "unhandled";
-            handleException(sender, exception);
+
+            logException(exception, "unobserved");
```
frenzibyte, line 247: think you meant `unhandled` here?
```suggestion
            logException(exception, "unhandled");
```

smoogipoo (approved):

PR #4183, opened 2021-02-01
Make window render to the cutout areas of android devices

Joehuu:
As said in https://github.com/ppy/osu/issues/11630#issuecomment-770571982, this makes it in line with iOS, but introduces the same issue of not being able to click stuff near the cutout area until the UI-side stuff is implemented. Can still use the scaling workaround in osu!-side, or just rotate device when viewing settings, etc. Won't cause any issues in gameplay.

Tested with Pixel 3a using the display cutout toggle in developer options (punch hole cutout):

Before:
Screenshot_20210131-222043

After:
Screenshot_20210131-222357

### DIFF
--- a/osu.Framework.Android/AndroidGameActivity.cs
+++ b/osu.Framework.Android/AndroidGameActivity.cs
@@ -55,6 +55,9 @@ namespace osu.Framework.Android
                 }
             };
 
+            if (Build.VERSION.SdkInt >= BuildVersionCodes.P)
+                Window.Attributes.LayoutInDisplayCutoutMode = LayoutInDisplayCutoutMode.ShortEdges;
+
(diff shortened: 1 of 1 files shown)


peppy:
@smoogipoo can you confirm you're okay with this direction?

peppy (approved):

smoogipoo:
I think so? I'm not knowledgeable on android stuff so I don't know what more feedback I could provide here.

smoogipoo (approved):

peppy:
The change here makes the lazer screen overlap "punch hole" cameras and "notch" areas of the screen, where it used to letterbox. Basically this brings things in line with what we see on iOS (without a safe area implementation).

PR #4186, opened 2021-02-04
Add cancellation support to WebRequest

smoogipoo:
This will come in handly when adding cancellation support to `OnlineStore`.

I overloaded `PerformAsync` instead of adding an optional parameter to not break o!f consumers.


peppy (approved):
Seems ok.

PR #4191, opened 2021-02-10
Add IPv4 fallback support on IPv6 failures

peppy:
[Not yet supported](https://github.com/dotnet/runtime/issues/26177) by .NET itself, so we need to do this ourselves.The case where this can happen is where DNSv6 succeeds, but the actual connection cannot be established. Without this added logic, a fallback to IPv4 (dns and connection) does not occur.

See inline comments for further rationale.

Closes https://github.com/ppy/osu/issues/3507.


smoogipoo (approved):
Looks solid to me

PR #4192, opened 2021-02-12
Fix input propagating to drawables which are not yet in a loaded state

peppy:
Closes https://github.com/ppy/osu/issues/10988.


osu.Framework/Graphics/Containers/CompositeDrawable.cs
```diff
@@ -1,4 +1,5 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+
```
frenzibyte, line 2:
```suggestion
```

smoogipoo (approved):

smoogipoo (approved):

PR #4197, opened 2021-02-16
Fix TextureStore and ResourceStore locks being held globally across all lookups

peppy:
- [x] Add test coverage

PRing this for feedback. I believe the code could be further simplified by using `Task.Run` instead of `TaskCompletionSource`, but in an effort to not add more new unknowns I went with this to avoid TPL usage.

Closes ppy/osu#11141.

### DIFF
--- a/osu.Framework/Graphics/Textures/TextureStore.cs
+++ b/osu.Framework/Graphics/Textures/TextureStore.cs
@@ -126,10 +126,10 @@ namespace osu.Framework.Graphics.Textures
 
             lock (retrievalCompletionSources)
             {
-                // check if an existing lookup was already started for this name.
-                if (!retrievalCompletionSources.TryGetValue(name, out task))
+                // check if an existing lookup was already started for this key.
+                if (!retrievalCompletionSources.TryGetValue(key, out task))
                     // if not, take responsibility for the lookup.
-                    retrievalCompletionSources[name] = (tcs = new TaskCompletionSource<Texture>()).Task;
+                    retrievalCompletionSources[key] = (tcs = new TaskCompletionSource<Texture>()).Task;
             }
 
             if (task != null)
--- a/osu.Framework/Graphics/Textures/TextureStore.cs
+++ b/osu.Framework/Graphics/Textures/TextureStore.cs
@@ -101,7 +101,7 @@ namespace osu.Framework.Graphics.Textures
         /// <returns>The texture.</returns>
         public new Texture Get(string name) => Get(name, default, default);
 
-        private readonly Dictionary<string, Task<Texture>> retrievalCompletionSources = new Dictionary<string, Task<Texture>>();
+        private readonly Dictionary<string, Task> retrievalCompletionSources = new Dictionary<string, Task>();
 
         /// <summary>
         /// Retrieves a texture from the store and adds it to the atlas.
@@ -121,7 +121,7 @@ namespace osu.Framework.Graphics.Textures
                 return cached;
 
             TaskCompletionSource<Texture> tcs = null;
-            Task<Texture> task;
+            Task task;
 
             lock (retrievalCompletionSources)
             {
@@ -131,8 +131,17 @@ namespace osu.Framework.Graphics.Textures
                     retrievalCompletionSources[key] = (tcs = new TaskCompletionSource<Texture>()).Task;
             }
 
+            // handle the case where a lookup is already in progress.
             if (task != null)
-                return task.Result;
+            {
+                task.Wait();
+
+                // always perform re-lookups through TryGetCached (see LargeTextureStore which has a custom implementation of this where it matters).
+                if (TryGetCached(key, out cached))
+                    return cached;
+
+                return null;
+            }
 
             this.LogIfNonBackgroundThread(key);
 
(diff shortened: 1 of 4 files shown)


osu.Framework/Graphics/Textures/TextureStore.cs
```diff
@@ -118,26 +121,45 @@ public virtual Texture Get(string name, WrapMode wrapModeS, WrapMode wrapModeT)
...
+
+            lock (retrievalCompletionSources)
             {
-                // If another retrieval of the texture happened before us, we should check if the texture exists in the cache again.
-                if (TryGetCached(key, out cached))
-                    return cached;
+                // check if an existing lookup was already started for this name.
+                if (!retrievalCompletionSources.TryGetValue(name, out task))
+                    // if not, take responsibility for the lookup.
+                    retrievalCompletionSources[name] = (tcs = new TaskCompletionSource<Texture>()).Task;
```
bdach, line 132: I don't believe using `name` (as opposed to `key`) is entirely correct. The same asset with a different wrap mode results in a slightly different texture on the atlas, after all.

Using `name` could lead to erroneous reads from task completion sources that request the same asset with a different wrap mode.

osu.Framework/IO/Stores/ResourceStore.cs
```diff
@@ -204,6 +189,11 @@ public virtual IEnumerable<string> GetAvailableResources()
             lock (stores) return stores.SelectMany(s => s.GetAvailableResources()).ExcludeSystemFileNames();
         }
 
+        private IResourceStore<T>[] getStores()
+        {
+            lock (stores) return stores.ToArray();
```
bdach, line 194: A loose thought about these allocs - I'm not sure it'll matter much, but a way to reduce these might be to have `{Add,Remove}Store` take the allocs by creating a copy of `stores` inside, having them mutate the copy to their hearts' contents, and then storing back the list reference atomically.

Most likely not a huge deal, but adding/removing stores should be an even rarer operation than querying for resources.

bdach (changes requested):
Other than the one comment (the other is more of an observation) I don't believe I see much wrong with the code. Tests would definitely be a plus, however.

osu.Framework/IO/Stores/ResourceStore.cs
peppy: Yeah, that might be a good direction, although you would hope that retrievals from stores are seldom enough that the alloc overhead is non-important. Given it only happens right now in cache failures (ie. there's always a store in front of a resource store to cache things) I'm on the fence whether we should change this now. @smoogipoo thoughts?

smoogipoo: Hmm... I was going to say it's fine given the rarity of resourcestore ever holding more than 1-2 nested stores, but then I realised that FontStore is a thing, which in the case of osu! has 13 nested stores. It's still at worst only ~100kb per second at 1000FPS but it's probably worthwhile to optimise the retrieval path as suggested.

peppy: when doing what though? retrieving hundreds of new glyphs constantly every frame without any caching? is this actually a valid use case?

peppy:
Turns out this new flow actually breaks `LargeTextureStore`, due to the texture being returns via the task result rather than being fetched via `TryGetCached` each time. Will need to revert the `Task`s to non-generic versions.

osu.Framework/IO/Stores/ResourceStore.cs
smoogipoo: True, I missed that it's backed by a TextureStore. I think this is fine as is then.

bdach (dismissed):

osu.Framework/Graphics/Textures/TextureStore.cs
```diff
@@ -118,26 +120,54 @@ public virtual Texture Get(string name, WrapMode wrapModeS, WrapMode wrapModeT)
             if (TryGetCached(key, out var cached))
                 return cached;
 
-            // Take an exclusive lock on retrieval of the texture.
-            lock (retrievalLock)
+            TaskCompletionSource<Texture> tcs = null;
+            Task task;
+
+            lock (retrievalCompletionSources)
```
smoogipoo, line 126: I can imagine a scenario where two threads reach this, one passes through, completes the entire rest of this method, completes the `finally`, removes the key from `retrievalCompletionSources`, and only then does the second one take the exclusive lock, which proceeds to fall through to the uncached path.

I believe this needs a `TryGetCached()` call inside it? Perhaps by moving the one above this block inside it.

smoogipoo (changes requested):

osu.Framework/Graphics/Textures/TextureStore.cs
peppy: Sounds fair.

The original code had the initial lookup outside the lock, do you think it's worth maintaining that? Or just presuming this lock is going to be pretty low contention and moving it inside.

smoogipoo: It should be low enough contention  that it doesn’t matter.

smoogipoo (approved):
lgtm now

PR #4201, opened 2021-02-17
Change IAdjustableAudioComponent/AudioAdjustments to use IBindable instead of BindableNumber

peppy:
I don't see any reason for it being a Bindable, or a Number.


peppy:
As it turns out I likely won't need this change for the purpose I was going to use it for, so merge or close at your discretion.

smoogipoo (approved):

PR #4203, opened 2021-02-18
Use ISample in DrawableSample

peppy:
Just PRing for visibility; don't merge without further discussion.


bdach:
This change is also sort of breaking, but I see no reason to not do it. Seems like a naturally flowing conclusion that a `DrawableSample` should be an `ISample`.

smoogipoo:
The only weird scenario is you can have `DrawableSample` be an ctor param? I suppose that's not a big deal...

smoogipoo (approved):

PR #4207, opened 2021-02-20
Fix incorrect entry assembly selection on android

bdach:
Aims to resolve https://github.com/ppy/osu/issues/11839. PRing for discussion.

# Root cause

On Android, `Assembly.GetEntryAssembly()` returns `null`. That is not at all surprising, as per the docs, there *is no* entry assembly - the game is being ran through mono.

`DebugUtils`'s entry assembly detection code would wrongly assume that if the entry assembly as given by the CLR is null, then the game was being ran in a test context (`IsNUnitRunning` is true). Therefore, on Android, all `DebugUtils` members that utilised the assembly detection, would use the properties of the `nunit.framework` assembly, instead of the executing assembly.

The reason why the `nunit.framework` assembly was chosen was that if an NUnit test context is requested outside of an actual test, NUnit will make up its own context (with the class name of `TestExecutionContext.AdhocContext`). That class resides in `nunit.framework`. ([Refer to the `nunit_test_assembly` accessor.](https://github.com/ppy/osu-framework/blob/105a17bc99cad251fa730b54c615d2b0d9a409d3/osu.Framework/Development/DebugUtils.cs#L31-L32))

Recent [upstream NUnit changes](https://github.com/nunit/nunit/pull/3748/commits/41021cc61823db44b39656e94cbe618ee2c4f439), that made it to the 3.13.1 release (enabling SourceLink support and shipping portable debug symbols) have caused this to trip over at the game side, and game builds in release configuration would show as debug builds. 

To resolve, leverage the `AdhocContext` in order to determine if we are running in the context of a particular, actual test. This makes it so that on Android, `DebugUtils.GetEntryAssembly()` will now defer to the executing assembly, and fix the wrong values of `IsDebugBuild` and similar.

# Remarks

This is a conservative resolution, because I'm not confident around this particular area of code, and from history it seems to have been fiddled with a lot - therefore I applied a local, targeted fix. If it were known for certain that NUnit never actually returns a null entry assembly, then this could be simplified to just checking the well-known assembly entry location names.

I have [previously stated on discord](https://discordapp.com/channels/188630481301012481/188630652340404224/812695469293305926) that this did not reproduce on the framework side; that was not correct. I was testing on an older commit; after pulling latest `master` I have reproduced the same issue in `SampleGame`. Any framework game after [the NUnit bump to 3.13.1](https://github.com/ppy/osu-framework/pull/4188), built normally (in release configuration) will hit this.

I have tested this with both `SampleGame` and lazer, and the change seems to have yielded the expected results. I also had a brief look at whether I haven't broken `IsNUnitRunning` by running tests from inside rider, and it doesn't seem that way, but skepticism/testing is welcome and warranted. My level of confidence in this fix is low - I am mostly submitting it for the discussion.

### DIFF
--- a/osu.Framework/Development/DebugUtils.cs
+++ b/osu.Framework/Development/DebugUtils.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using NUnit.Framework;
+using NUnit.Framework.Internal;
 
 namespace osu.Framework.Development
 {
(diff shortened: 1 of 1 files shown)


peppy (approved):
I think this reads fine. I've tested on desktop platforms and it looks to (as expected) work as previously. Let's get this in and get android releases happening again.

👍🏻

PR #4209, opened 2021-02-20
Make `IAdjustableAudioComponent` inherit `IAggregateAudioAdjustment`

### LINKED ISSUE #4208: Expose the aggregate track adjustment values through IAdjustableAudioComponent
Currently, `IAdjustableAudioComponent` exposes each of the adjustments as `BindableNumber` already.
https://github.com/ppy/osu-framework/blob/72eef659d4374052e04021361ba1549bafa7c442/osu.Framework/Audio/IAdjustableAudioComponent.cs#L16-L31
These however don't represent the updated values after adjustments are applied and will always return 1. (whether that is intended or not).

Therefore there is no way to actually fetch the true Tempo/Frequency/etc.. values from ITrack.

Use case,
https://github.com/H2n9/osu/blob/e3e06addb45c0b0ebd75a728c366130d4b261948/osu.Game/Rulesets/Mods/ModPitchShift.cs#L62

H2n9:
Exposes the Aggregate[Adjustment] bindables in ITrack.

Closes #4208

### DIFF
--- a/osu.Framework/Audio/AdjustableAudioComponent.cs
+++ b/osu.Framework/Audio/AdjustableAudioComponent.cs
@@ -8,7 +8,7 @@ namespace osu.Framework.Audio
     /// <summary>
     /// An audio component which allows for basic bindable adjustments to be applied.
     /// </summary>
-    public class AdjustableAudioComponent : AudioComponent, IAggregateAudioAdjustment, IAdjustableAudioComponent
+    public class AdjustableAudioComponent : AudioComponent, IAdjustableAudioComponent
     {
         private readonly AudioAdjustments adjustments = new AudioAdjustments();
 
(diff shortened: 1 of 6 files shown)


frenzibyte (approved):
I think this is fine, honestly it all feels like `IAdjustableAudioComponent` should inherit `IAggregateAudioAdjustment` itself, but I guess that's out of scope from the goal of this.

bdach:
It might be fine to have `IAdjustableAudioComponent : IAggregateAudioAdjustment`. I can't see much reason to not do that instead, except interface separation, maybe. Would be nice to get more opinions on this.

peppy:
Was thinking the same thing. If we're going to start exposing these on each component that implements them there's no reason not to.

bdach:
I've made the change. There are no actual code changes, just shuffling interfaces around, so this seems a harmless enough change.

smoogipoo (approved):

PR #4212, opened 2021-02-22
Fix AudioContainer not binding child adjustments to itself

smoogipoo:
Should be pretty self-explanatory. Fixes spinner frequency modulation not working.

### DIFF
--- a/osu.Framework/Graphics/Containers/AudioContainer.cs
+++ b/osu.Framework/Graphics/Containers/AudioContainer.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using osu.Framework.Audio;
 using osu.Framework.Graphics.Audio;
 using osu.Framework.Graphics.Effects;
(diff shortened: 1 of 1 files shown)


peppy:
I'm not sure I follow this change. `DrawableAudioWrapper` is already doing an outwards fetch of the nearest `IAggregateAudioAdjustment`, which should be handling this bind flow. Am I missing something here?

smoogipoo:
I see, good catch. This looks to be related to the osu-side pooling changes, which is why this works.

I don't know if the upwards lookup of `DrawableAudioWrapper` is what we want. Moving samples between containers is another case that is completely broken.

peppy:
The idea of the way it was done is that it would handle children nested quite deep (and generally follows our principles closer?). Not sure how that would be handled otherwise.

smoogipoo:
That's fair enough... Do you have any ideas for another fix then? My secondary suggestion is binding to `Invalidation.Parent` and re-looking up the parent, but I feel like you're inevitably going to respond with that BDL is completely broken for that case and it's not a supported operation. Would you be fine with that fix?

peppy:
I think I'd be fine with the proposed flow via parent invalidation. We already do a similar thing via `GetContainingInputManager()` so as long as the naming follows that one it shouldn't be too far out of line. I'd definitely make this `internal` framework implementation though... should never be of concern to a user at this point.

@smoogipoo did you want to make that change or shall i?

PR #4213, opened 2021-02-22
Rename built-in font OpenSans to OpenSans-Regular

MiraiSubject:
Right now to achieve similar functionality as `OsuFont` with font weights in my own project I import OpenSans as OpenSans-Regular from my own resources, but figured it would be worth PR'ing the font name change upstream for other consumers to use. 

https://github.com/MiraiSubject/KanojoWorks/blob/ab5bf25d3c3fe7bc1b2f301c554ec6cf0fdeffc1/KanojoWorks/KanojoWorksGameBase.cs#L53

I've tested it visually against the `SpriteText` test in framework tests and against https://github.com/MiraiSubject/KanojoWorks/commit/5552d33246febbf5f48a219c17d2aebc91358be0 with a local framework checkout

image

# Breaking Changes

## The regular weight of OpenSans has been renamed to OpenSans-Regular

Until now this font wasn't following the convention we use everywhere else. If you are referencing it directly, please update your strings to point to the new name.


bdach:
I think this might be technically a subtly breaking change - consider a consumer with an entirely different default font, that was using Open Sans without the regular weight specification; I'm reasonably certain after this change their fonts would switch.

That said I'm not necessarily against it myself. Seems like it tidies things up a bit.

bdach:
Actually, upon looking at `TestSceneSpriteText`, the breakage I mentioned in the above comment is visible there (and even visible in the screenshot). Note that every fourth sprite text is using Roboto and not not Open Sans. This change should be applied to fix this:

```diff
diff --git a/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteText.cs b/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteText.cs
index 874f739bd7..5b61d52161 100644
--- a/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteText.cs
+++ b/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteText.cs
@@ -49,7 +49,7 @@ public TestSceneSpriteText()
                 SpriteText text = new SpriteText
                 {
                     Text = $@"Font testy at size {i}",
-                    Font = new FontUsage("OpenSans", i, i % 4 > 1 ? "Bold" : null, i % 2 == 1),
+                    Font = new FontUsage("OpenSans", i, i % 4 > 1 ? "Bold" : "Regular", i % 2 == 1),
                     AllowMultiline = true,
                     RelativeSizeAxes = Axes.X,
                 };
```

which in turns reveals yet another breakage - for consistency, `OpenSans-Italic` should be renamed to `OpenSans-RegularItalic`.

bdach (changes requested):
as above

peppy:
I think this is fine to be a breaking change. It should be a pretty rare use case that a user is using this font, and even if they are they will notice it immediately and should find the changelog notes.

peppy (approved):

PR #4216, opened 2021-02-24
Fix leak of pinned sample handles

smoogipoo:
Internally, `ManagedBass` does a similar thing to our own `ObjectHandle`: https://github.com/ManagedBass/ManagedBass/blob/0cdc873a3b2eeb33bb338d254a76d20fc4f36c1b/src/Bass/Shared/GCPin.cs#L27-L38, with the difference that it only frees the handle when Bass provides `SyncFlags.Free`.

In the context of samples, this leaks the pinned handle, as bass will not invoke sync callbacks for `HSAMPLE`s. See [`Bass_ChannelSetSync`](http://www.un4seen.com/doc/#bass/BASS_ChannelSetSync.html) which is documented as accepting a handle of type `HMUSIC`, `HSTREAM` or `HRECORD`, and not `HSAMPLE`.

Don't really know how to benchmark this, but I've tested that the sync callback is not invoked, and have seen the reduction in pinned handles via dotMemory.

### DIFF
--- a/osu.Framework/Audio/Sample/SampleBassFactory.cs
+++ b/osu.Framework/Audio/Sample/SampleBassFactory.cs
@@ -76,9 +76,6 @@ namespace osu.Framework.Audio.Sample
         {
             const BassFlags flags = BassFlags.Default | BassFlags.SampleOverrideLongestPlaying;
 
-            if (RuntimeInfo.SupportsJIT)
-                return Bass.SampleLoad(data, 0, data.Length, PlaybackConcurrency.Value, flags);
-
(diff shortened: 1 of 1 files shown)


smoogipoo:
Reported at https://github.com/ManagedBass/ManagedBass/issues/88

bdach:
I haven't 100% checked the memory numbers but looked at everything else (`Bass.LastError` changing from `OK` to `Handle` after the attempt at a sync, and the lack of hits on a breakpoint to the free sync) and it seems to check out :+1:

bdach (approved):

PR #4217, opened 2021-02-25
Switch order of parameters on RomanisableString to make the original version come first

peppy:
I think it makes more sense to put the unicode (or as it is now named, original) string first, as it represents the "main" display version of the string. Or so you would think, but `ShowUnicode` is set to `false` by default sooo...

I believe this feels better anyway. Also added xmldoc.

### DIFF
--- a/osu.Framework/Localisation/RomanisableString.cs
+++ b/osu.Framework/Localisation/RomanisableString.cs
@@ -13,7 +13,7 @@ namespace osu.Framework.Localisation
         public readonly string Romanised;
         public readonly string Unicode;
 
-        public RomanisableString(string romanised, string unicode)
+        public RomanisableString(string unicode, string romanised)
         {
             Romanised = romanised;
             Unicode = unicode;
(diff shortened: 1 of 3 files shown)


osu.Framework/Localisation/RomanisableString.cs
```diff
@@ -1,30 +1,50 @@
...
 {
     /// <summary>
-    /// A string that has variations in Unicode and Romanised form.
+    /// A string that has a romanised fallback to allow a better experience for users that potentially can't read the original script.
+    /// See <see cref="FrameworkSetting.ShowUnicode"/>, which can toggle the display of romanised variants.
     /// </summary>
     public class RomanisableString
     {
+        /// <summary>
+        /// The string in its original script. May be null.
```
smoogipoo, line 17: If it can be null then it needs to be a `string?` Also, this handle both being null all that gracefully, though I don't know if that'll cause issues later on with `SpriteText`.

smoogipoo (changes requested):

osu.Framework/Localisation/RomanisableString.cs
peppy: Yeah, I noticed this while updating the class. Optimally, we would *require* the original to be non-null? But I can see that being annoying to handle in osu! usages..

smoogipoo: `string.Empty` as you've done is fine imo.

smoogipoo (approved):

PR #4218, opened 2021-02-25
Add optimised HasFlagFast<T> extension to Enum

smoogipoo:
On a RELEASE (this is important, will be explained later) startup of osu!, I noticed that a few enum values would thrash about in Gen0. This doesn't appear in dotMemory, but they're revealed via dotnet-sos:

```
> dumpheap -type Axes
         
         Address               MT     Size
...
00007f938cd7f680 00007f93b9b12b88       24

Statistics:
              MT    Count    TotalSize Class Name
00007f93b9b12b88     4222       101328 osu.Framework.Graphics.Axes
Total 4222 objects

> gcwhere 00007f938cd7f680

Address            Gen   Heap   segment            begin              allocated           size
00007F938CD7F680   0      0     00007F9383FFE000   00007F9383FFF000   00007F938CFD9FE8    0x18(24)
```

I originally noticed this issue with `CompositeDrawable.ChildLifeStateChange`, however I was originally testing with the DEBUG configuration.

In trying to isolate this, I constructed the following test:
```csharp
class Program
{
    private static DotNetRuntimeListener listener;

    static void Main(string[] args)
    {
        listener = new DotNetRuntimeListener();

        bool b = false;
        for (long i = 0; i < long.MaxValue - 1; i++)
            b = getFlag(Environment.TickCount).HasFlag(FlagsEnum.Flag2 | FlagsEnum.Flag3);

        Console.WriteLine(b);
    }

    [MethodImpl(MethodImplOptions.NoInlining)]
    private static FlagsEnum getFlag(int i) => (FlagsEnum)i;

    [Flags]
    private enum FlagsEnum
    {
        Flag1 = 1,
        Flag2 = 2,
        Flag3 = 4,
        Flag4 = 8
    }

    // https://medium.com/criteo-labs/c-in-process-clr-event-listeners-with-net-core-2-2-ef4075c14e87
    internal sealed class DotNetRuntimeListener : EventListener
    {
        private const int gc_keyword = 0x0000001;

        private const string statistics_grouping = "GC";

        protected override void OnEventSourceCreated(EventSource eventSource)
        {
            if (eventSource.Name == "Microsoft-Windows-DotNETRuntime")
                EnableEvents(eventSource, EventLevel.Verbose, (EventKeywords)gc_keyword);
        }

        protected override void OnEventWritten(EventWrittenEventArgs data)
        {
            switch ((EventType)data.EventId)
            {
                case EventType.GCStart_V1 when data.Payload != null:
                    // https://docs.microsoft.com/en-us/dotnet/framework/performance/garbage-collection-etw-events#gcstart_v1_event
                    Console.WriteLine($"GC | Gen: {data.Payload[1]} Reason: {data.Payload[2]} Type: {data.Payload[3]}");
                    break;
            }
        }

        private enum EventType
        {
            GCStart_V1 = 1,
            GCHeapStats_V1 = 4,
        }
    }
}

public static class EnumExtensions
{
    public static unsafe bool HasFlagFast<T>(this T enumValue, T flag) where T : unmanaged, Enum
    {
        int intVal1 = Unsafe.As<T, int>(ref enumValue);
        int intVal2 = Unsafe.As<T, int>(ref flag);

        return (intVal1 & intVal2) > 0;
    }
}
```
When run in DEBUG this prints many lines of `GC | Gen: 0 Reason: 0 Type: 0` to the console. When run in RELEASE, this does not print at all. It looks like the de-boxing of `Enum.HasFlag` is dependent on:
1) The JIT.
2) The build config, which probably relates to (1).

As you can see in the above isolation, I also implemented a custom `HasFlagFast` method which doesn't exhibit this behaviour in any build config. Further benchmarking (via BenchmarkDotNet) leads to the following results:
```
DEBUG:
|      Method |      Mean |     Error |    StdDev |     Gen 0 | Gen 1 | Gen 2 |  Allocated |
|------------ |----------:|----------:|----------:|----------:|------:|------:|-----------:|
|     HasFlag | 29.483 ms | 0.3228 ms | 0.2862 ms | 1406.2500 |     - |     - | 48000009 B |
|  BitwiseAnd |  4.416 ms | 0.0017 ms | 0.0016 ms |         - |     - |     - |        2 B |
| HasFlagFast | 10.017 ms | 0.0106 ms | 0.0088 ms |         - |     - |     - |        4 B |

RELEASE:
|      Method |     Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------ |---------:|----------:|----------:|------:|------:|------:|----------:|
|     HasFlag | 1.894 ms | 0.0010 ms | 0.0009 ms |     - |     - |     - |       1 B |
|  BitwiseAnd | 1.894 ms | 0.0004 ms | 0.0003 ms |     - |     - |     - |       1 B |
| HasFlagFast | 1.894 ms | 0.0010 ms | 0.0009 ms |     - |     - |     - |       1 B |
```
Keep in mind, however, that this isn't always a DEBUG thing - I was seeing `Axes` in Gen0 in RELEASE-mode osu!.

In this PR, I've built upon the above method to support all enum types (byte/short/int/long), and replaced all existing usages of `Enum.HasFlag` with the new `HasFlagFast` method. I've also banned `Enum.HasFlag` from use.

Now, the same steps as above shows 0 `Axes` objects in Gen0:
```
> dumpheap -type Axes

         Address               MT     Size

Statistics:
              MT    Count    TotalSize Class Name
Total 0 objects
```

### DIFF
--- a/CodeAnalysis/BannedSymbols.txt
+++ b/CodeAnalysis/BannedSymbols.txt
@@ -2,3 +2,4 @@ M:System.Object.Equals(System.Object,System.Object)~System.Boolean;Don't use obj
 M:System.Object.Equals(System.Object)~System.Boolean;Don't use object.Equals. Use IEquatable<T> or EqualityComparer<T>.Default instead.
 M:System.ValueType.Equals(System.Object)~System.Boolean;Don't use object.Equals(Fallbacks to ValueType). Use IEquatable<T> or EqualityComparer<T>.Default instead.
 T:System.IComparable;Don't use non-generic IComparable. Use generic version instead.
+M:System.Enum.HasFlag(System.Enum);Use osu.Framework.Extensions.EnumExtensions.HasFlagFast<T>() instead.
--- a/.editorconfig
+++ b/.editorconfig
@@ -209,3 +209,6 @@ dotnet_diagnostic.IDE0073.severity = warning
 
 #Disable operator overloads requiring alternate named methods
 dotnet_diagnostic.CA2225.severity = none
+
+# Banned APIs
+dotnet_diagnostic.RS0030.severity = error
--- a/osu.Framework.Benchmarks/BenchmarkEnum.cs
+++ b/osu.Framework.Benchmarks/BenchmarkEnum.cs
@@ -1,0 +1,61 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Runtime.CompilerServices;
+using BenchmarkDotNet.Attributes;
+using osu.Framework.Extensions.EnumExtensions;
+
+namespace osu.Framework.Benchmarks
+{
+    [MemoryDiagnoser]
+    public class BenchmarkEnum
+    {
+        [Benchmark]
+        public bool HasFlag()
+        {
+            bool result = false;
+
+#pragma warning disable RS0030 // (banned API)
+            for (int i = 0; i < 1000000; i++)
+                result |= getFlags(i).HasFlag((FlagsEnum)i);
+#pragma warning restore RS0030
+
+            return result;
+        }
+
+        [Benchmark]
+        public bool BitwiseAnd()
+        {
+            bool result = false;
+
+            for (int i = 0; i < 1000000; i++)
+                result |= (getFlags(i) & (FlagsEnum)i) > 0;
+
+            return result;
+        }
+
+        [Benchmark]
+        public bool HasFlagFast()
+        {
+            bool result = false;
+
+            for (int i = 0; i < 1000000; i++)
+                result |= getFlags(i).HasFlagFast((FlagsEnum)i);
+
+            return result;
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private FlagsEnum getFlags(int i) => (FlagsEnum)i;
+
+        [Flags]
+        private enum FlagsEnum
+        {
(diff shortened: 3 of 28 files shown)


peppy:
Are you planning on filing a bug upstream for this one?

smoogipoo:
I'm not sure how to create a viable example for them yet. Giving them a debug example won't help much since I assume the runtime disables a bunch of optimisations there anyway.

CodeAnalysis/BannedSymbols.txt
```diff
@@ -2,3 +2,4 @@ M:System.Object.Equals(System.Object,System.Object)~System.Boolean;Don't use obj
 M:System.Object.Equals(System.Object)~System.Boolean;Don't use object.Equals. Use IEquatable<T> or EqualityComparer<T>.Default instead.
 M:System.ValueType.Equals(System.Object)~System.Boolean;Don't use object.Equals(Fallbacks to ValueType). Use IEquatable<T> or EqualityComparer<T>.Default instead.
 T:System.IComparable;Don't use non-generic IComparable. Use generic version instead.
+M:System.Enum.HasFlag(System.Enum);Use osu.Framework.Extensions.EnumExtensions.HasFlagFast<T>() instead.
```
peppy, line 5: have to remember to add this to the osu-side configuration too.

peppy:
Is it worth adding the comparison benchmark to the benchmarks project?

peppy (approved):

smoogipoo:
Ah yep, will do.

peppy (approved):

PR #4224, opened 2021-02-26
Fix enumerating HoveredDrawables causing excessive allocations

peppy:
Screenshots should only be compared using relative % (run times were different):

Before:
20210226 163606 (Safari app)

After:
20210226 163358 (Parallels Desktop app)

### DIFF
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -107,7 +107,7 @@ namespace osu.Framework.Input
         /// Top-down in this case means reverse draw order, i.e. the front-most visible
         /// <see cref="Drawable"/> first, and <see cref="Container"/>s after their children.
         /// </summary>
-        public IReadOnlyList<Drawable> HoveredDrawables => hoveredDrawables;
+        public IReadOnlyList<Drawable> HoveredDrawables => hoveredDrawables.AsSlimReadOnly();
 
         /// <summary>
         /// Contains all <see cref="Drawable"/>s in top-down order which are considered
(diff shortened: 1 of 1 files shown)


smoogipoo:
Where are the rest of the allocations coming from (the backtraces)? I'd expect `SROLW.GetEnumerator()` to be 0 o_0.

peppy:
I don't think you can drill down into the enumerator itself..

smoogipoo:
Yeah but what's calling GetEnumerator() there?

peppy:
It's the `MenuCursorContainer`, same scenario I was fixing this for.

osu.Framework/Input/InputManager.cs
```diff
@@ -107,7 +107,7 @@ public Drawable DraggedDrawable
         /// Top-down in this case means reverse draw order, i.e. the front-most visible
         /// <see cref="Drawable"/> first, and <see cref="Container"/>s after their children.
         /// </summary>
-        public IReadOnlyList<Drawable> HoveredDrawables => hoveredDrawables;
+        public IReadOnlyList<Drawable> HoveredDrawables => hoveredDrawables.AsSlimReadOnly();
```
smoogipoo, line 110:
```suggestion
        public SlimReadOnlyListWrapper<Drawable> HoveredDrawables => hoveredDrawables.AsSlimReadOnly();
```
That should fix the remaining allocation.

smoogipoo (changes requested):

osu.Framework/Input/InputManager.cs
peppy: Hmm, right. Didn't realise we were happy with exposing these lists with the wrapper, but that *is* how it's done everywhere else.

smoogipoo (approved):

peppy:
20210226 165004 (Parallels Desktop app)

PR #4226, opened 2021-02-26
Reuse TextBuilder when validating SpriteText

smoogipoo:
Before:
Screenshot_2021-02-26_17-24-25

After:
Screenshot_2021-02-26_17-35-45

Provides the `InvalidateTextBuilder()` method for any consumers that have a custom `TextBuilder` and provide properties to control it other than truncate/ellipsisstring/multiline.


peppy:
Okay with changing this to

20210226 174414 (JetBrains Rider)?

smoogipoo:
Sure, I originally had it similar to that before.

peppy (approved):

PR #4227, opened 2021-02-26
Fix number formatter allocating many arrays

smoogipoo:
Before:
Screenshot_2021-02-26_18-19-47

After:
Screenshot_2021-02-26_18-41-54

By holding down ctrl to show text in the frame statistics.


bdach (approved):
Seems like a no-brainer :+1:

PR #4239, opened 2021-03-02
Remove finalizer from video texture uploads

peppy:
This is one rare case outside of `FrameStatisticsDisplay` that we are performing many texture uploads, where the finalizer of the uploads can, as we have seen, add considerable overhead.

I've confirmed that all loss-of-reference paths result in disposal of these instances (but failing that, one would hope that ffmpeg would clean up when the owning video is freed).

### DIFF
--- a/osu.Framework/Graphics/Video/VideoTextureUpload.cs
+++ b/osu.Framework/Graphics/Video/VideoTextureUpload.cs
@@ -29,7 +29,7 @@ namespace osu.Framework.Graphics.Video
         /// </summary>
         /// <param name="frame">The frame to upload.</param>
         /// <param name="freeFrameDelegate">A function to free the frame on disposal.</param>
-        public VideoTextureUpload(AVFrame* frame, FFmpegFuncs.AvFrameFreeDelegate freeFrameDelegate)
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Video/VideoTextureUpload.cs
```diff
@@ -1,40 +1,46 @@
...
 using osu.Framework.Graphics.Textures;
 using osuTK.Graphics.ES30;
 using FFmpeg.AutoGen;
+using osu.Framework.Graphics.Primitives;
+using SixLabors.ImageSharp.PixelFormats;
 
 namespace osu.Framework.Graphics.Video
 {
-    public unsafe class VideoTextureUpload : ArrayPoolTextureUpload
+    public unsafe class VideoTextureUpload : ITextureUpload
```
smoogipoo, line 13:
```suggestion
    public sealed unsafe class VideoTextureUpload : ITextureUpload
```
If not implementing the proper disposal pattern with a virtual method.

smoogipoo (changes requested):

smoogipoo (approved):

Issue #4229, opened 2021-03-01
Remove usage of finalizers wherever possible


peppy:
While investigating https://github.com/ppy/osu/issues/11800, we have basically identified the root cause to be the presence of objects with finalizers (even if they have been suppressed) existing in the finalizer queue.

Using the following model classes:

```csharp
public class NonFinalizingThing
{
    public NonFinalizingThing()
    {
    }
}

public class FinalizingThing
{
    public FinalizingThing()
    {
    }

    ~FinalizingThing()
    {
    }
}
```

Constant allocations without finalizer:

```csharp
for (int i = 0; i < 100000000; i++)
{
    var thing = new NonFinalizingThing();
}

```

![20210301 194407 (Parallels Desktop app)](https://user-images.githubusercontent.com/191335/109486652-8184a800-7ac6-11eb-95da-71e7f06c0c02.png)

With:

```csharp
for (int i = 0; i < 100000000; i++)
{
    var thing = new FinalizingThing();
    GC.SuppressFinalize(thing);
}

```

![20210301 194500 (Parallels Desktop app)](https://user-images.githubusercontent.com/191335/109486734-9c571c80-7ac6-11eb-91fb-33485be2eb71.png)

Note the pause time increasing from 0.1ms to around 7ms.

peppy:
Remaining classes which look to require the finalizer:

- `VideoDecoder`
- `BassCallback` (arguable, may want to make internal and avoid it, depending on alloc rate)
- `FrameBuffer` (GL resources)
- `RenderBuffer` (GL resources)
- `VertexBuffer` (GL resources)
- `Icon` (Native resources)
- `TimePeriod` (Native resources)
- `BackgroundStackTraceCollector` (thread cleanup; can probably be internalised / avoided but low impact)

Remaining classes to check / remove if possible, in order of impact:

- [ ] `AudioComponent` (high impact, needs checking)
- [ ] `TextureGL` (GL resources; can likely be moved to `TextureGLSingle`)
- [ ] `TextureWithRefCount` (i believe we rely on it here, but maybe could be internalised)
- [ ] `AsyncBufferStream` (probably safe to remove)

peppy:
Unassigning self in case anyone wants to attempt the remaining ones.

peppy:
I don't see any immediate way around the finalizer in `TextureWithRefCount` so I've marked that as complete. It's relatively low usage, at least (on a texture object which itself has pretty high memory/alloc overhead).

PR #4240, opened 2021-03-02
Remove finalizers from TextureUpload classes

peppy:
I've checked every usage and this looks safe.

- [x] Depends on https://github.com/ppy/osu-framework/pull/4239.

Partly addresses #4229.

### DIFF
--- a/osu.Framework/Graphics/Textures/TextureUpload.cs
+++ b/osu.Framework/Graphics/Textures/TextureUpload.cs
@@ -113,10 +113,23 @@ namespace osu.Framework.Graphics.Textures
 
         #region IDisposable Support
 
+        private bool disposed;
+
         public void Dispose()
         {
+            Dispose(true);
+            GC.SuppressFinalize(this);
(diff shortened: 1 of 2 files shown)


smoogipoo:
Implemented disposal patterns. Good to go.

smoogipoo (approved):

osu.Framework/Graphics/Textures/TextureUpload.cs
```diff
@@ -114,26 +115,21 @@ internal TextureUpload()
...
-        }
+            if (disposed)
+                return;
 
-        public void Dispose()
-        {
-            Dispose(true);
-            GC.SuppressFinalize(this);
+            image?.Dispose();
+            pixelMemory.Dispose();
```
smoogipoo, line 130: Also something to note, technically everything managed should be inside `if (isDisposing)` but it's not a huge issue. We can do a secondary pass over that project-wide.

PR #4242, opened 2021-03-02
Remove finalizer from Drawable and DrawNode

peppy:
These are the bigger ones. Will address the remaining finalizers (on lesser used classes) in a separate PR.

I haven't done intensive testing on this yet, only checked at a code-level, so please consider when reviewing.

Partly addresses #4229.

### DIFF
--- a/osu.Framework/Graphics/BufferedDrawNodeSharedData.cs
+++ b/osu.Framework/Graphics/BufferedDrawNodeSharedData.cs
@@ -96,15 +96,9 @@ namespace osu.Framework.Graphics
         /// </summary>
         internal void ResetCurrentEffectBuffer() => currentEffectBuffer = -1;
 
-        ~BufferedDrawNodeSharedData()
-        {
-            GLWrapper.ScheduleDisposal(() => Dispose(false));
-        }
-
         public void Dispose()
         {
             GLWrapper.ScheduleDisposal(() => Dispose(true));
-            GC.SuppressFinalize(this);
         }
 
         protected virtual void Dispose(bool isDisposing)
--- a/osu.Framework/Graphics/BufferedDrawNodeSharedData.cs
+++ b/osu.Framework/Graphics/BufferedDrawNodeSharedData.cs
@@ -99,6 +99,7 @@ namespace osu.Framework.Graphics
         public void Dispose()
         {
             GLWrapper.ScheduleDisposal(() => Dispose(true));
+            GC.SuppressFinalize(this);
         }
 
         protected virtual void Dispose(bool isDisposing)
--- a/osu.Framework/Graphics/DrawNode.cs
+++ b/osu.Framework/Graphics/DrawNode.cs
@@ -288,14 +288,17 @@ namespace osu.Framework.Graphics
 
         public void Dispose()
         {
-            if (referenceCount.Decrement() != 0)
-                return;
-
-            GLWrapper.ScheduleDisposal(() => Dispose(true));
+            Dispose(true);
+            GC.SuppressFinalize(this);
         }
 
         protected virtual void Dispose(bool isDisposing)
         {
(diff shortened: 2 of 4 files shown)


osu.Framework/Graphics/Drawable.cs
```diff
@@ -66,46 +66,21 @@ protected Drawable()
...
-        /// Disposes this drawable.
-        /// </summary>
-        protected virtual void Dispose(bool isDisposing)
-        {
-        }
-
-        private void dispose(bool isDisposing)
         {
             //we can't dispose if we are mid-load, else our children may get in a bad state.
             lock (LoadLock)
```
smoogipoo, line 79: Do you have anything against moving this to `Dispose()`, and moving the rest of this logic into `Dispose(bool)`? As it is it's kinda dodgy if someone were to implement a finalizer.

smoogipoo (changes requested):
Also please add back the suppressions.

osu.Framework/Graphics/Drawable.cs
peppy: it already is, unless i'm missing something? misread the diff maybe?

peppy: oh, you actually want the disposal logic moved? are you okay with the fact that base.Dispose could be omitted/forgotten?

smoogipoo: Yes, well this is the disposal pattern so...

osu.Framework/Graphics/DrawNode.cs
```diff
@@ -284,24 +284,21 @@ protected virtual void DrawOpaqueInterior(Action<TexturedVertex2D> vertexAction)
...
+            Dispose(true);
             GC.SuppressFinalize(this);
         }
 
         protected virtual void Dispose(bool isDisposing)
         {
+            if (referenceCount.Decrement() != 0)
+                return;
+
+            GLWrapper.ScheduleDisposal(() => Dispose(true));
```
smoogipoo, line 300: These should be moved back to `Dispose()`.

osu.Framework/Graphics/BufferedDrawNodeSharedData.cs
```diff
@@ -96,15 +96,9 @@ public FrameBuffer GetNextEffectBuffer()
...
 
-        ~BufferedDrawNodeSharedData()
-        {
-            GLWrapper.ScheduleDisposal(() => Dispose(false));
-        }
-
         public void Dispose()
         {
             GLWrapper.ScheduleDisposal(() => Dispose(true));
-            GC.SuppressFinalize(this);
```
smoogipoo, line 107: Add back

smoogipoo (changes requested):

osu.Framework/Graphics/DrawNode.cs
```diff
@@ -284,24 +284,19 @@ protected virtual void DrawOpaqueInterior(Action<TexturedVertex2D> vertexAction)
...
-                return;
-
             GLWrapper.ScheduleDisposal(() => Dispose(true));
             GC.SuppressFinalize(this);
         }
 
         protected virtual void Dispose(bool isDisposing)
         {
+            if (referenceCount.Decrement() != 0)
+                return;
```
smoogipoo, line 298: Still need to move this back. Why was it moved?

If a finalizer comes in (if ever implemented), it should get disposed no matter what the reference count is.

A case for something like this could be `DrawFrameRecordingContainer` (the other thing incrementing the refcount) holding the recorded frames and then the entire tree getting finalized.

smoogipoo (changes requested):

smoogipoo (approved):
Seems fine and looks to work :+1:

Issue #4248, opened 2021-03-02
SpriteText truncate does not work properly


Joehuu:
Regressed in https://github.com/ppy/osu-framework/pull/4226.

| Before | After |
| --- | --- |
| ![chrome_DZcuSVlekG](https://user-images.githubusercontent.com/35318437/109698670-80925b80-7b44-11eb-9b8b-37e593968f46.jpg) | ![chrome_tHoDETy9j6](https://user-images.githubusercontent.com/35318437/109698942-ccdd9b80-7b44-11eb-8156-0c27d0da7fbe.jpg) |
| ![st7vIeja94](https://user-images.githubusercontent.com/35318437/109698768-9d2e9380-7b44-11eb-8503-198ce61178ee.gif) | ![QHZV0PWAJH](https://user-images.githubusercontent.com/35318437/109699126-057d7500-7b45-11eb-9b18-bebb32e0cd75.gif) |

Note that the chat item resolves after hovering out.

PR #4249, opened 2021-03-02
Fix text builder not invalidating on indirect relative size changes 

bdach:
Resolves https://github.com/ppy/osu-framework/issues/4248.

As it turns out this was covered by framework tests, but only visually. I've hacked up something not-overly-pretty that will fail a test if this breaks again.

Not entirely sure I did this 100% correctly. It seems to work at least, but `DrawSize` might not be enough? I'm always confused by those flags.

Also this will technically invalidate `DrawSize` twice when setting `Width`/`Height` (once via base, once via the manual `invalidate(true, true);` call), but I wasn't sure if I should trim it and explain with a comment, and it shouldn't have any performance impact anyways.

Figured I'd open this for the test, at the very least.

### DIFF
--- a/osu.Framework/Graphics/Sprites/SpriteText.cs
+++ b/osu.Framework/Graphics/Sprites/SpriteText.cs
@@ -575,7 +575,7 @@ namespace osu.Framework.Graphics.Sprites
         /// </summary>
         protected virtual char FallbackCharacter => '?';
 
-        private readonly LayoutValue<TextBuilder> textBuilderCache = new LayoutValue<TextBuilder>(Invalidation.DrawSize);
+        private readonly LayoutValue<TextBuilder> textBuilderCache = new LayoutValue<TextBuilder>(Invalidation.DrawSize, InvalidationSource.Parent);
 
         /// <summary>
         /// Invalidates the current <see cref="TextBuilder"/>, causing a new one to be created next time it's required via <see cref="CreateTextBuilder"/>.
--- a/osu.Framework/Graphics/Sprites/SpriteText.cs
+++ b/osu.Framework/Graphics/Sprites/SpriteText.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
-using osu.Framework.Caching;
 using osu.Framework.Development;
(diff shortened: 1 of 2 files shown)


peppy (dismissed):

osu.Framework/Graphics/Sprites/SpriteText.cs
```diff
@@ -575,12 +575,12 @@ private void invalidate(bool characters = false, bool textBuilder = false)
         /// </summary>
         protected virtual char FallbackCharacter => '?';
 
-        private readonly Cached<TextBuilder> textBuilderBacking = new Cached<TextBuilder>();
+        private readonly LayoutValue<TextBuilder> textBuilderCache = new LayoutValue<TextBuilder>(Invalidation.DrawSize);
```
smoogipoo, line 578: I think this will trigger a second time for autosized texts? Probably needs `InvalidationSource.Parent`

smoogipoo (changes requested):

osu.Framework/Graphics/Sprites/SpriteText.cs
bdach: Good point. I'll add test coverage for it and check, but I'm pretty sure you're right.

bdach: I wasn't able to come up with a test that would indeed trigger this twice, but I did see several reasons for making this change, namely:

* the obvious - less invalidations = better
* it passes the new test
* `Width` and `Height` invalidate manually anyway
* other `LayoutValues` in this class also combine `DrawSize` with `Parent`
* judging by `Drawable.invalidateParentSizeDependencies()` it does seem to make sense to ignore invalidations from self, if I'm understanding it correctly

So I've just made the change. If it plays up in some other edge case I guess we'll revisit.

smoogipoo (approved):
Makes sense to me

osu.Framework/Graphics/Sprites/SpriteText.cs
smoogipoo: Yeah it's pretty hard to test layout, and your reasoning is sound. `Drawable.invalidateParentSizeDependencies()` is only an optimisation for `BypassAutoSizeAxes` though.

PR #4255, opened 2021-03-03
Improve code quality surrounding confine mode changes

### LINKED ISSUE #4134: The mouse is permanently confined in Borderless even if the confinement option is changed
The mouse confinement seems to have some issues with the `Borderless` screen mode:
- If the confine is set to "during gameplay" and you finish a song or exit the player loader, you can't move your cursor to a second screen anyway (even if you set the confinement to "never")
- ^ after replicating the above problem, if you set the screen mode to `Windowed`, the cursor still remains confined to the current screen (ignoring the size of the game window)
- the issue seems to be fixed when you resize/drag/unfocus the window (so after performing the second point), or alt-tabbing when in `Borderless` screen mode.

peppy:
Adds testability and fixes event flow a touch. Changes made while attempting to help with #4134, but don't yet have any idea what is going on so going to leave it here.

### DIFF
--- a/osu.Framework.Tests/Visual/Platform/TestSceneFullscreen.cs
+++ b/osu.Framework.Tests/Visual/Platform/TestSceneFullscreen.cs
@@ -77,6 +77,12 @@ namespace osu.Framework.Tests.Visual.Platform
         [Test]
         public void TestScreenModeSwitch()
         {
+            if (window == null)
+            {
+                Assert.Ignore("This test cannot run in headless mode (a window instance is required).");
+                return;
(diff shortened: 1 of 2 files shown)


smoogipoo (dismissed):

bdach:
`TestScreenModeSwitch` was failing when ran headless, because... well, it needs a window to be useful. So I went ahead and just ignored it.

Other than that this looks all right. :+1:

bdach (approved):

frenzibyte (approved):

PR #4257, opened 2021-03-04
Use SDL2 relative mouse mode by default

peppy:
This is a second attempt which supersedes and closes #3960. I moved much of the logic that was added to the `DesktopWindow` implementation locally to the handler, as I believe it makes more sense.

As it turns out, rewriting the pass-off logic from scratch was all that was required to fix the [issues i mentioned](https://github.com/ppy/osu-framework/pull/3960#issuecomment-720057361).

Tested on Windows and macOS. Linux testing would be appreciated.

Should close some or all of #2860, #2814, #2247, #2077, #1378, https://github.com/ppy/osu/issues/11929 (confirmed), https://github.com/ppy/osu/issues/3364 (confirmed), https://github.com/ppy/osu/issues/3489.


peppy:
Note that this *removes absolute mapping support*.

Need to consider that this means any raw input absolute devices that may have been working until now no longer will. I'm not sure if this is important. Needs testing with tablets + raw input for sure as it may be a regression.

If it is actually required, we can potentially support via the osuTK handler (for only the absolute case).

gagahpangeran:
Tested on linux with i3wm and X11.

- Raw input and change mouse sensitivity works (fullscreen, borderless, windowed).
- Confine mouse works (fullscreen, borderless, windowed).
- Dragging window by title bar works.
- Resizing window works.

Not sure about wayland issue (e.g. #3452), since I dont have wayland installed on my machine.

frenzibyte:
@H2n9 Since you've stated that you can reproduce [this raw input issue](https://github.com/ppy/osu/issues/3364) almost reliably, can you double-check it is completely fixed with this PR just in case?

maotovisk:
tested on arch linux (linux-zen 5.11-1) x11 + gnome:
 - Raw Input indeed works
 - Just as @gagahpangeran said, confining/dragging/resizing works aswell
 
 gonna test using wayland and report back

peppy:
Most interested in testing with *tablets*, if anyone is able to do that.

maotovisk:
Testing my Wacom CTL 472 with xinput, it seems once raw input is enabled, the tablet stops working:
 - raw input on > no data reported 
 - raw input off > data reported 

[xinput_output.txt](https://github.com/ppy/osu-framework/files/6082553/xinput_output.txt)

H2n9:
Tested this framework PR on lazer. Raw input no longer completely fails but some interesting new behaviours come up. (using same method as from [#3364)](https://github.com/ppy/osu/issues/3364#issuecomment-788910240)

After releasing the grabbed window from the previous reproduction steps, the cursor is locked to the lazer window despite whatever window is in focus. The cursor cannot be moved until osu recieves focus at which point it correctly works and the cursor is no longer locked. Continuing from this point, the only different behaviour is that the osu-lazer cursor does not follow the mouse off the screen(windowed) unlike in a normal startup scenario. (though i only manage to reproduce this 50% of the time that raw input failed)

Overall everything was a lot harder to reproduce however. I could really only manage to reproduce the above around 30% of the time with the intention of trying to cause it. If the osu window at all grabs cursor focus during startup, raw input seems to work fine.

Video example:
https://streamable.com/h2b07l

peppy:
That sounds like the osu! window getting an incorrect focused state (something we've struggled with over the years.. turns out window managers are just not good at this).

It sounds like this is true, but can you confirm that everything returns to normal after cycling focus to the osu! window?

H2n9:
Yeah, nothing significantly game-breaking happens after cycling focus. Only small difference is at 0:26 in the video. The osu-lazer cursor is confined instead of following the cursor outside the window which doesnt happen in a normal startup scenario (no cursor lock). Not sure which behaviour is intended but that's the only small thing I noticed.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -810,8 +825,13 @@ private void handleMouseButtonEvent(SDL.SDL_MouseButtonEvent evtButton)
...
         }
 
-        private void handleMouseMotionEvent(SDL.SDL_MouseMotionEvent evtMotion) =>
-            ScheduleEvent(() => OnMouseMove(new Vector2(evtMotion.x * Scale, evtMotion.y * Scale)));
+        private void handleMouseMotionEvent(SDL.SDL_MouseMotionEvent evtMotion)
+        {
+            if (SDL.SDL_GetRelativeMouseMode() == SDL.SDL_bool.SDL_FALSE)
+                ScheduleEvent(() => OnMouseMove(new Vector2(evtMotion.x * Scale, evtMotion.y * Scale)));
+            else
+                ScheduleEvent(() => OnMouseMoveRelative(new Vector2(evtMotion.xrel, evtMotion.yrel)));
```
frenzibyte, line 833: Cannot tell exactly which way is correct [as mouse acceleration still persists on relative mode for macOS](https://github.com/libsdl-org/SDL/issues/67), but doesn't this sound like it should be multiplied by `Scale`, like in the absolute case?

frenzibyte (changes requested):
On Windows, this is working perfectly aside from the mentioned issue.

On macOS, there is only one issue that's bugging me, occurs on a non-confined raw input enabled window, which is a weird position desync between the OS cursor and the game's mouse state, causing both cursors to show up, or causing the OS cursor to never show outside window: https://streamable.com/c5jryr

It may likely be an SDL deep issue, but haven't investigated through it yet to assure that; pointing it out here in case you find something about it.

osu.Framework/Platform/SDL2DesktopWindow.cs
peppy: probably. does it fix the issue for you?

peppy: I've made the change, please confirm it's fixed your reported issue.

frenzibyte: Oh right, I haven't realized that this causes a mismatch between where the cursor is on screen and where the game thinks the cursor is in...

Yeah this fixes that issue, except that the cursor hiding when leaving out the window still occurs:

https://user-images.githubusercontent.com/22781491/110136700-0c281880-7de1-11eb-8045-dba26c8ffae4.mov

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -1187,6 +1207,13 @@ public void CycleMode()
             WindowMode.Value = currentValue;
         }
 
+        /// <summary>
+        /// Update the host window manager's cursor position based on a location relative to window coordinates.
+        /// </summary>
+        /// <param name="position">A position inside the window.</param>
+        public void UpdateMousePosition(Vector2 position) => ScheduleCommand(() =>
+            SDL.SDL_WarpMouseInWindow(SDLWindowHandle, (int)position.X, (int)position.Y));
```
frenzibyte, line 1215: Coming from https://github.com/ppy/osu-framework/pull/4257#discussion_r588388959.

It seems that for macOS (doesn't happen for windows, weirdly), the OS cursor will not show up if `SDL_WrapMouseInWindow` is invoked with a position that is very near to the game window, needs to be at least 5 pixels away from it for the cursor to show up.

I don't think it's much of a deal though, I'll leave it up to you whether it should be fixed now or just call it an SDL bug and leave it.

osu.Framework/Input/Handlers/Mouse/SDL2RelativeMouseHandler.cs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                // store the last mouse position to propagate back to the host window manager when exiting relative mode.
+                lastPosition = position;
+            }
+
+            // handle the case where relative / raw input is active, but the cursor may have exited the window
+            // bounds and is not intended to be confined.
+            if (window.RelativeMouseMode && !window.CursorConfined)
+            {
+                if (position.X < 0 || position.Y < 0 || position.X > window.Size.Width || position.Y > window.Size.Height)
```
frenzibyte, line 86: sounds like it should be:
```suggestion
                if (position.X < 0 || position.Y < 0 || position.X >= window.Size.Width || position.Y >= window.Size.Height)
```

osu.Framework/Platform/SDL2DesktopWindow.cs
peppy: Where does this 5 pixel number come from?

frenzibyte: Just from trial and error, no idea why it has to be 5 pixels away.

peppy: Based on this and another issue with the cursor warping position randomly, I've disabled macOS support for now.

peppy:
I ended up merging `MouseHandler` and `RelativeMouseHandler` in this PR as doing it separately feels like unnecessary overhead. I've also disabled relative mode on macOS temporarily while I look further into the issues we are seeing.

This should still be in a good state to move forward with, for windows and linux. Do note that the "raw input" toggle in the test scene will now (intentionally) do nothing. There is now only one method of operation which is using the relative mouse path.

One thing that still needs testing is high-DPI screen configurations.

peppy:
If you're not interested in investigating I am fine to turn it off for linux as well, and we'll be back where we were. Which is fine really; just have to wait for someone to actually care enough to make sensitivity work properly.

bdach:
On an empirical look on my linux box (not looked at the code yet), I cannot reproduce the "no movement" thing. I am seeing some vertical/horizontal angle snapping, but the only mice/pointing devices I have currently available to me are: a laptop touchpad, a generic mouse, and a tablet (which is broken with this, as noted previously). So I'm not sure whether that's anything to go off of.

*However*, I notice that SDL2's mouse event fields are all integers. Say, theoretically, if a mouse had a high enough polling rate to report subpixel motions, then if SDL was doing rounding wrong somewhere along the way (maybe not even on its own volition, maybe because xinput was interfering somewhere), then that would explain both the "slow movement doesn't move" issue, *and* the angle snapping (wayward deltas would be getting discarded due to being below zero). Looking at the `InputManager` test scene, it's quite clear that if sensitivity is 1, then any movement is only along full pixels.

I peeked a look at source, and it looks like there [is an SDL function that looks to be doing this properly](https://github.com/libsdl-org/SDL/blob/bd06538778102f72bad8393ef07da5a1ec444217/src/events/SDL_mouse.c#L310-L323), at least in the generic code, but... it only ever works if sensitivity is set via their `SDL_HINT_MOUSE_RELATIVE_SPEED_SCALE` hint. If it's at the default 1, then whatever integers get in through the platform-specific code will fall out. I imagine if the raw underlying data is delta in pixels, then even that won't save us here, and will presumably behave funky. Generally I don't really believe we're seeing the full resolution of the mouse inputs that we should be.

Leaving mostly wild conjecture aside, on my linux box the relative mode looks to work relatively indistinguishably from normal operation with crappy mice, and inoperable with a Wacom tablet (mouse positions jitter on every positional update from the device wildly, to the tune of several thousands of pixels). So personally I'd like for tablet to be salvaged first if this is to go in.

peppy:
Firstly, in every OS I know mouse input is always in pixels. You should check the xinput or otherwise lower level code and confirm that is also the case (ie. the input to the function you linked).

As already mentioned, this can't be dropped as it *adds the ability to change sensitivity*, which is why I stay keyed on this in the first place.

smoogipoo:
What caused this was an xorg evdev config setting `Option "Resolution" "1200"` which is documented as "The resolution is used to scale relative motion events from mouse devices to 1000 DPI resolution.".

I don't actually need this, so I'll go ahead and review this as normal.

smoogipoo (approved):

PR #4260, opened 2021-03-05
Disable macOS cursor location assistance if not visible

frenzibyte:
Closes https://github.com/ppy/osu/issues/11772

Just realized such code for disabling this existed in the osuTK window implementation, added it here in a pretty simple way, and it seems to work perfectly.

Also made the shake feature enable itself back when cursor is not in window, wasn't feeling alright leaving it disabled there.

### DIFF
--- a/osu.Framework/Platform/MacOS/MacOSWindow.cs
+++ b/osu.Framework/Platform/MacOS/MacOSWindow.cs
@@ -22,6 +22,16 @@ namespace osu.Framework.Platform.MacOS
         private IntPtr originalScrollWheel;
         private ScrollWheelDelegate scrollWheelHandler;
 
+        public override bool CursorVisible
+        {
+            get => base.CursorVisible;
+            set
+            {
+                base.CursorVisible = value;
+                updateCursorAssistanceState();
+            }
+        }
+
         public override void Create()
         {
             base.Create();
@@ -30,6 +40,16 @@ namespace osu.Framework.Platform.MacOS
             var viewClass = Class.Get("SDLView");
             scrollWheelHandler = scrollWheel;
             originalScrollWheel = Class.SwizzleMethod(viewClass, "scrollWheel:", "v@:@", scrollWheelHandler);
+
+            CursorInWindow.BindValueChanged(_ => updateCursorAssistanceState(), true);
+        }
+
+        private void updateCursorAssistanceState()
+        {
+            if (CursorInWindow.Value && !CursorVisible)
+                NSApplication.PresentationOptions |= NSApplicationPresentationOptions.DisableCursorLocationAssistance;
+            else
+                NSApplication.PresentationOptions &= ~NSApplicationPresentationOptions.DisableCursorLocationAssistance;
         }
 
         /// <summary>
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -177,7 +177,7 @@ namespace osu.Framework.Platform
         /// <summary>
         /// Returns or sets the cursor's visibility within the window.
         /// </summary>
-        public bool CursorVisible
+        public virtual bool CursorVisible
         {
             get => cursorVisible;
             set


frenzibyte:
It appears this gets broken by borderless mode, blocking until I find a better fix for this.

frenzibyte:
Fixed, was happening because of SDL [overwriting the presentation options](https://github.com/libsdl-org/SDL/blob/a4ddb175f1f1d832960c830191daaab7eb25638f/src/video/cocoa/SDL_cocoawindow.m#L899-L906) when transitioning to fullscreen desktop/borderless mode, handled on our side instead.

osu.Framework/Platform/MacOS/MacOSWindow.cs
```diff
@@ -41,15 +48,30 @@ public override void Create()
             scrollWheelHandler = scrollWheel;
             originalScrollWheel = Class.SwizzleMethod(viewClass, "scrollWheel:", "v@:@", scrollWheelHandler);
 
+            // handle invisible cursor when providing presentation options for "fullscreen desktop" mode.
```
peppy, line 51: This isn't descriptive enough. Can you link to the reasoning for this being necessary? Is there some code in SDL2 which makes this required? Some stack overflow thread explaining why this works?

osu.Framework/Platform/MacOS/MacOSWindow.cs
```diff
@@ -17,8 +17,15 @@ public class MacOSWindow : SDL2DesktopWindow
         private static readonly IntPtr sel_scrollingdeltay = Selector.Get("scrollingDeltaY");
         private static readonly IntPtr sel_respondstoselector_ = Selector.Get("respondsToSelector:");
 
+        private const NSApplicationPresentationOptions default_fullscreen_presentation_options =
```
peppy, line 20: Where does this come from?

osu.Framework/Platform/MacOS/MacOSWindow.cs
```diff
@@ -41,15 +48,30 @@ public override void Create()
...
 
+        private uint windowWillUseFullScreen(IntPtr self, IntPtr cmd, IntPtr window, uint options) =>
+            (uint)applyCursorAssistanceState(default_fullscreen_presentation_options);
+
         private void updateCursorAssistanceState()
+        {
+            NSApplication.PresentationOptions = applyCursorAssistanceState(NSApplication.PresentationOptions);
+        }
+
+        private NSApplicationPresentationOptions applyCursorAssistanceState(NSApplicationPresentationOptions options)
```
peppy, line 67: I find the flow through this function really weird. Preferred how you had it previously.

peppy (changes requested):

osu.Framework/Platform/MacOS/MacOSWindow.cs
frenzibyte: Well, there is SDL2 code showing that it overwrites the presentation options when transitioning:
https://github.com/libsdl-org/SDL/blob/a4ddb175f1f1d832960c830191daaab7eb25638f/src/video/cocoa/SDL_cocoawindow.m#L899-L906

What I'm doing is simply registering to that method to pass the `DisableCursorLocationAssistance` presentation option when transitioning, as the SDL2 code above overwrites that option. `OsuTKMacOSWindow` does that as well.

osu.Framework/Platform/MacOS/MacOSWindow.cs
frenzibyte: From the SDL code, I think I can use method swizzling and read what SDL returns there and apply the cursor assistance disable option to it, if that's better.

osu.Framework/Platform/MacOS/MacOSWindow.cs
peppy: This needs to be mentioned then. Because you are overwriting all options, which could have implications with other states we may be changing on SDL2's end?

The comment you have here explains *what* you are doing but not *why*. There's no mention that without this the values get overwritten but SDL2.

And if you are to be replacing with local `const` defaults, you need to show via a link to SDL2 source or otherwise that this is safe to do.

osu.Framework/Platform/MacOS/MacOSWindow.cs
frenzibyte: Had to do it this way to not duplicate code somewhat, are you fine with:
image

frenzibyte: Pushed change

peppy: I think that looks better, yes.

peppy:
Have you checked whether this is required with relative mouse mode set to on? My hypothesis is that it wouldn't be, but worth checking.

osu.Framework/Platform/MacOS/MacOSWindow.cs
peppy: What is the "invisible cursor" you mention in the inline comment?

osu.Framework/Platform/MacOS/MacOSWindow.cs
```diff
@@ -30,6 +50,34 @@ public override void Create()
...
+            Class.RegisterMethod(windowClass, windowWillUseFullScreenHandler, "window:willUseFullScreenPresentationOptions:", "I@:@I");
+
+            CursorInWindow.BindValueChanged(_ => updateCursorAssistanceState(), true);
+        }
+
+        private bool shouldDisableCursorAssistance => CursorInWindow.Value && !CursorVisible;
+
+        private uint windowWillUseFullScreen(IntPtr self, IntPtr cmd, IntPtr window, uint options)
+        {
+            var fullscreenOptions = default_fullscreen_presentation_options;
```
peppy, line 67: this logic is different from the SDL implementation. won't it be incorrect in non-fullscreen scenarios?

peppy (changes requested):

osu.Framework/Platform/MacOS/MacOSWindow.cs
frenzibyte: `CursorVisible = false`, not sure how to reword it properly

peppy: How is that related to the code following it?

frenzibyte: I'm not sure what you mean, it handles disabling cursor assistance when switching to borderless/fulllscreen desktop mode, as SDL [overwrites](https://github.com/libsdl-org/SDL/blob/a4ddb175f1f1d832960c830191daaab7eb25638f/src/video/cocoa/SDL_cocoawindow.m#L899-L906) the presentation options on switching to that mode.

It does that by registering to the `willUseFullScreenPresentationOptions` method and providing its own fullscreen options + disabling cursor assistance conditionally

peppy: If i understand the flow correctly, you are applying the defaults without considering the conditional in SDL's code: https://github.com/libsdl-org/SDL/blob/a4ddb175f1f1d832960c830191daaab7eb25638f/src/video/cocoa/SDL_cocoawindow.m#L901

peppy:
Did you actually test this? Surely it was incorrect before the latest change you made?

Either way, this feels super-flaky. Have you checked whether this is required with relative mode on? If it isn't, I'd consider not even bothering and instead focusing on making relative mode work as the default.

frenzibyte:
I have tested this thoroughly, iirc `willUseFullScreenPresentationOptions` is only called when switching to borderless or fullscreen, that's why I missed to realize that.

I have checked with relative mode on, it does seem to disable the assistance, but I believe we have relative mode disabled on macOS currently? unless you think it's not worth adding this logic in then, in which case I don't mind.

peppy:
Let's leave this open but not merge until issues with relative mouse are resolved. If/when we can set the default mode of operation to relative mouse, this should not be required. After all, a user can disable this themselves if they *really* want to use non-relative mode.

PR #4262, opened 2021-03-07
Usage of unused variable

BlauFx:
While debugging my game I noticed that [ScrollDistance](https://github.com/ppy/osu-framework/blob/74a3308d505ffd45b646b246ca598d07e7419473/osu.Framework/Graphics/Containers/ScrollContainer.cs#L78) has no effect. It turns out the value is hardcoded and the variable is not being used.

### DIFF
--- a/osu.Framework/Graphics/Containers/ScrollContainer.cs
+++ b/osu.Framework/Graphics/Containers/ScrollContainer.cs
@@ -371,7 +371,7 @@ namespace osu.Framework.Graphics.Containers
             if (ScrollDirection == Direction.Horizontal && scrollDelta.X != 0)
                 scrollDeltaFloat = scrollDelta.X;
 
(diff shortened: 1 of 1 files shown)


bdach:
This actually regressed waaaaaaaaay back, at https://github.com/ppy/osu-framework/commit/530a4a6a2f3d226cd542483e9328f1f897ee04e3.

While this change does appear to make some sense, I'm not sure that the precise case should remain hardcoded? Both should probably be controllable somehow, even if by having the precise scroll be a fraction of the non-precise distance (would be one-eighth with current numbers).

smoogipoo:
I think that can be touched upon in a further PR and get this in as is. Not sure whether best is to expose it as controllable or divide or to keep it as hardcoded (because it is supposed to track precisely).

smoogipoo (approved):

Issue #4214, opened 2021-02-24
Thread pool completely broken on iOS


peppy:
Opening this issue since it isn't tracked anywhere else. This is the reason for lack of iOS deployments in recent times. I finally have a reproduction going locally in a debug environment, which means we can actually begin to make some forward progress on the issue.

Both of the following reproductions can be performed on `master` in a debug/device configuration. In my testing, I intentionally reduced the size of our `CompositeDrawable` `TaskScheduler`s to 1 thread just to avoid any potential issues due to total active thread count (probably irrelevant though).

Reproduction 1:

- Run `TestSceneTextures` to completion.
- Should freeze on "get texture" step in `RefCountTextureAvailability`.

![20210224 170935 (Visual Studio)](https://user-images.githubusercontent.com/191335/108967600-12bcde80-76c3-11eb-8eea-1cf27db4e30d.png)

Reproduction 2:

- Run `TestSceneVideo`
- Should freeze on getting the video from network

![20210224 170455 (Visual Studio)](https://user-images.githubusercontent.com/191335/108967108-6bd84280-76c2-11eb-800f-de0052530696.png)

At first sight, this looks to be standard .NET ThreadPool saturation, although I am having difficulty asserting this as the Visual Studio debugger seems to not show any running thread pool threads (I have seen some show up, but I guess it's hiding or doing something weird to paused/sleep/waiting threads?)

On the osu-side, this can be completely resolved by changing the startup sequence to anything that isn't triangles. This is likely due to the video decoding loop also using the threadpool (it's actually surprising how few things we have using the threadpool..).

Checking `ThreadPool.GetAvailableThreads(out int wrk, out int compl);` shows results of `200` for each count. Doesn't really mean much, but that's about all the visibility I can get without going deeper.

I've confirmed that thread pool tasks aren't hitting any exceptions, so am at a bit of a loss as to where the threads are going.

I intend on digging deeper into this one, but worthwhile to have a thread open in case anyone has a bright idea.

sr229:
From what I can tell, iOS only has a maximum of 64 threads, really worth investigating at the moment, I assume only threading is affected since from what I can deduce from the initial findings,  any intro but triangles work.

peppy:
Please, re-read my opening post. The reason it fails is due to TPL threadpool usage. This is already a known.

sr229:
One of the recommendations I got from asking through the .NET community was run the debugging using LLDB, quoting: 

[...]

PR #4263, opened 2021-03-08
Fix crash on startup on iOS due to host being null?

peppy:
I'm not sure I understand how this is happening, but this does fix the hard crash on osu! startup when running under iOS/Release (with debugger mode disabled). 

This is the only change required to make iOS work again. That said, I still think we want to get the `ConfigureAwait` changes in (#4261) because it's only a matter of time before we cause a deadlock like that.

This would explain why the await specific cases in `osu.Framework.Tests` where reproducible in the simulator while the osu! side crash was not.

A few things:

- The exception is silently eaten, causing the game to startup (and the draw thread to be waiting to draw). Not sure if this is an iOS specific issue, or if we are handling exceptions incorrectly across all platforms in some code path. We definitely need to investigate and fix this either way.
- This looks to somehow be caused by a race condition (or specific timing). Turning on Xamarin debugging support causes it to no longer fail. Also, testing on desktop platforms shows that the dependency is always available.

20210308 161954 (Xcode)

If any further debug information is required, I can run anything `lldb` from this point.

Note that this is probably not to be considered a final fix for the issue, without understanding how/why/what is going wrong.

### DIFF
--- a/osu.Framework/Input/Bindings/KeyBindingContainer.cs
+++ b/osu.Framework/Input/Bindings/KeyBindingContainer.cs
@@ -23,9 +23,6 @@ namespace osu.Framework.Input.Bindings
     public abstract class KeyBindingContainer<T> : KeyBindingContainer
         where T : struct
     {
-        [Resolved]
-        protected GameHost Host { get; private set; }
-
         private readonly SimultaneousBindingMode simultaneousMode;
         private readonly KeyCombinationMatchingMode matchingMode;
 
@@ -330,6 +327,12 @@ namespace osu.Framework.Input.Bindings
     /// </summary>
     public abstract class KeyBindingContainer : Container
     {
+        // This is only specified here (rather than in PlatformContainer, where it is consumed) to workaround
+        // a critical iOS / Xamarin bug. It should not need to be here.
+        // See https://github.com/ppy/osu-framework/pull/4263 for discussion.
+        [Resolved]
+        protected GameHost Host { get; private set; }
+
         protected IEnumerable<IKeyBinding> KeyBindings;
 
         public abstract IEnumerable<IKeyBinding> DefaultKeyBindings { get; }
--- a/osu.Framework/Input/Bindings/KeyBindingContainer.cs
+++ b/osu.Framework/Input/Bindings/KeyBindingContainer.cs
@@ -328,7 +328,8 @@ namespace osu.Framework.Input.Bindings
     public abstract class KeyBindingContainer : Container
     {
         // This is only specified here (rather than in PlatformContainer, where it is consumed) to workaround
-        // a critical iOS / Xamarin bug. It should not need to be here.
+        // a critical iOS / Xamarin bug, where consumer applications may crash during startup at an unmanaged level.
+        // It should eventually be removed when the issue is identified and fixed upstream.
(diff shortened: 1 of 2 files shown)


smoogipoo:
There's no way this fixes anything. I can't believe it. It would mean that the Drawable load sequence is _completely_ broken and would be a failure of catastrophic proportions.
1. The failing path is coming from `LoadComplete`, well after dependencies have been injected. There's 0% change that host is null.
2. The "exception" is EXC_BAD_ACCESS. This is a native exception.
3. This hasn't failed for us in the past year of iOS releases.

This is 101% a mono/xamarin.ios issue and needs isolation. At the absolute, bare minimum, this needs in-code documentation, even as a temporary fix, because this is so silly.

Does it always repro?

peppy:
It's bad access on address `0x0`, which would imply null. But yeah, it's most likely some kind of monotouch/mono optimisation kicking in. I can't understand how it would be something our end and only fail in this one scenario.

As already mentioned in the opening post, I don't consider this an actual fix.

It's 100% repro. I've gone back and forth to `master` and confirmed multiple times.

smoogipoo:
It's a mono-level native exception. Managed NREs don't terminate with kernel exceptions.

Make a Xamarin issue in that case (and are you updated to the latest Xamarin.iOS SDK?)

peppy:
This is the same issue that has stopped iOS from running for several months. It's persisted across every SDK version, including going back to releases where it was working correctly. So it's likely something else in our codebase in conjunction that has triggered the issue. The code around this failure hasn't changed in recent times.

smoogipoo:
Are you trying to say it's an issue with our codebase/our problem to fix? Because it's not. This is a GC or AOT issue (likely the latter) which our code hit a snag in. The reason it doesn't fail with debug is likely just due to lack of optimisations in the AOT.

peppy:
No. I am not. I'm not sure where you are getting that idea.

Also to clarify, it fails on Release builds with optimisations, just with the debugger startup overhead (attaching a remote debugging via the xamarin tcp endpoint). Which is why I think it may be timing related.

peppy:
To clarify: it started failing not on an SDK version change, but something our end (as far as I recall). So something we are doing broke something mono side, but changing SDK versions (as you suggested) doesn't have any effect.

smoogipoo:
The op mentions:

> but this does fix the hard crash on osu! startup when running under iOS/Release (with debugger mode disabled).

Your comment mentions:

> it fails on Release builds with optimisations, just with the debugger startup overhead (attaching a remote debugging via the xamarin tcp endpoint)

Which one is correct?

peppy:
Setting this to enabled:

20210308 173435 (Visual Studio)

and possibly turning on full debug information, can confirm this if it's important:

20210308 173506 (Visual Studio)

Other than that, the configuration is set to release.

smoogipoo:
Can you provide a native stack trace? Like the one you get in console.log.

peppy:
As mentioned in the opening post, there is none. It's hanging up via watchdog (10 seconds of frozen UI thread).

Here's the crash log.. doesn't show much helpful:

https://gist.github.com/peppy/9fe8f1d2b21528aaf8b5655191de53b6

It's also same as the original one posted in the issue thread (#4214).

smoogipoo (approved):

PR #4264, opened 2021-03-08
Implement windows raw input handling locally

peppy:
This adds local handling of raw input, bypassing SDL2. While this is usually not required, there are some quirks that we want to work around with specific devices which is not feasible with SDL2's limited API exposure. This also allows us to handle absolute packets, which are not supported by SDL2.

This is designed in a way that it can be disabled, but I believe there is no reason to. As such, by default it is enabled.

- [ ] Depends on #4257


smoogipoo:
This is quite broken with any sensitivity other than 1.0: https://drive.google.com/file/d/1fXVaqViQytFfAAUiMi2x3ddSYgiHRbfY/view

Can you repro any of this or is it something I should look at too?

peppy:
Just to reaffirm: I don't think there's anything wrong with this implementation itself. The issues you found are related to the already-merged relative mouse switch.

smoogipoo (approved):

peppy:
Have tested with multiple displays and it looks to behave as expected (at least with the devices i have at hand).

smoogipoo (approved):

PR #4265, opened 2021-03-08
Fix LocalisableString operator== failing for non-interned strings

smoogipoo:
Fixes https://github.com/ppy/osu/issues/11974

Preferring `.Equals()` implementations to `operator==` for internal comparison of LocalisableString data. Note that `operator==` forwards to `string.Equals(a, b, StringComparison.Ordinal)` internally (though it does this via a by-byte-span helper so it's not immediately obvious).

I've left in `operator==` and `operator!=` because removal would be a breaking change. I can obsolete them if people think that's the right path forward, but I'm not convinced on that.

In short:
```csharp
LocalisableString localisable = "a";
string str = "a";

// Ordinal comparisons
localisable.Equals(str);
str.Equals(localisable);

// Ordinal comparisons
localisable == str;
str == localisable;

// Culture-specific comparisons
localisable.ToString().Equals(str, StringComparison.OrdinalIgnoreCase);
str.Equals(localisable.ToString(), StringComparison.OrdinalIgnoreCase);
string.Equals(localisable.ToString(), str, StringComparison.OrdinalIgnoreCase);
```

### DIFF
--- a/osu.Framework/Graphics/Sprites/SpriteText.cs
+++ b/osu.Framework/Graphics/Sprites/SpriteText.cs
@@ -97,7 +97,7 @@ namespace osu.Framework.Graphics.Sprites
             get => text;
             set
             {
-                if (text == value)
+                if (text.Equals(value))
                     return;
 
                 text = value;
--- a/osu.Framework.Benchmarks/BenchmarkLocalisableString.cs
+++ b/osu.Framework.Benchmarks/BenchmarkLocalisableString.cs
@@ -1,0 +1,28 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using BenchmarkDotNet.Attributes;
+using osu.Framework.Localisation;
+
+namespace osu.Framework.Benchmarks
+{
+    [MemoryDiagnoser]
+    public class BenchmarkLocalisableString
+    {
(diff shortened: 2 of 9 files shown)


peppy (approved):

osu.Framework/Graphics/Sprites/SpriteText.cs
```diff
@@ -97,7 +97,7 @@ public LocalisableString Text
             get => text;
             set
             {
-                if (text == value)
+                if (text.Equals(value))
```
bdach, line 100: Just to double-confirm - this change has no behavioural effect aside from saving on the allocation from the implicit conversion, right?

bdach (approved):
Yeah I'm OK with this. I briefly considered having the equality operators have the semantics given here (so memberwise if type matches).

osu.Framework/Graphics/Sprites/SpriteText.cs
smoogipoo: Yeah this is semantically equivalent to `==`. It's left-over from a previous change where I had removed the `==` operator, but kept these changes in-case we wanted to obsolete it or something.

peppy: I guess it's preferred to use `.Equals` at osu! side as well?

smoogipoo: I'm not really fussed either way.

PR #4268, opened 2021-03-09
Add missed `ConfigureAwait(false)` calls for netstandard

bdach:
Resolves some build warnings I just noticed today. Missed by CI because *grumbles about dual-targeting and conditional compilation*.

.NET 6 can't come soon enough...


peppy (approved):

peppy:
Dangerous..

PR #4272, opened 2021-03-11
Prepare InputHandlers for configuration via bindables and reflection

peppy:
This is some initial tidy-up to get things into a working state. There will likely be more changes required going forward, but wanted to get this out to make sure the direction feels good.


peppy:
Note that I've gone in a bit of a different direction that doesn't necessarily require these changes to go in, but I think they all count as code quality improvements. You be the judge.

osu.Framework/Input/Handlers/Midi/MidiHandler.cs
```diff
@@ -15,7 +15,7 @@
 
 namespace osu.Framework.Input.Handlers.Midi
 {
-    public class MidiInputHandler : InputHandler
+    public class MidiHandler : InputHandler
```
bdach, line 18: Bit of a nitpick, but might want to override `Description` here (from what I gather it'll read `Midi` via the automatic implementation in `InputHandler`, instead of `MIDI`).

osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
```diff
@@ -17,30 +17,19 @@ namespace osu.Framework.Input.Handlers.Mouse
...
-
-        public override int Priority => 0;
-
-        private bool useRelativeMode = true;
-
         /// <summary>
         /// Whether relative mode should be preferred when the window has focus and the cursor is contained.
         /// </summary>
-        public bool UseRelativeMode
+        public BindableBool UseRelativeMode { get; set; } = new BindableBool(true)
```
bdach, line 23:
```suggestion
        public BindableBool UseRelativeMode { get; } = new BindableBool(true)
```

osu.Framework/Platform/MacOS/MacOSGameHost.cs
```diff
@@ -65,7 +65,7 @@ protected override IEnumerable<InputHandler> CreateAvailableInputHandlers()
             foreach (var h in handlers.OfType<MouseHandler>())
             {
                 // There are several bugs we need to fix with macOS / SDL2 cursor handling before switching this on.
-                h.UseRelativeMode = false;
+                h.UseRelativeMode.Value = false;
```
bdach, line 68: Would also probably want to disable this bindable, too.

bdach (changes requested):

bdach:
Generally I don't see this diff as necessarily hugely worse or better, just a bit different and maybe more consistent. So aside from the above I have no objections.

osu.Framework/Platform/MacOS/MacOSGameHost.cs
peppy: Intentionally didn't want to do this, so the user can decide for themselves.

bdach (approved):

bdach (dismissed):

smoogipoo (approved):

PR #4275, opened 2021-03-12
Fix potential null reference when iterating open midi devices

peppy:
Reproduced once while testing, you'll have to take my word on this one. The fail case was where there were two items in the array, first being `null`, second being the single device I actually have connected.

### DIFF
--- a/osu.Framework/Input/Handlers/Midi/MidiInputHandler.cs
+++ b/osu.Framework/Input/Handlers/Midi/MidiInputHandler.cs
@@ -42,10 +42,13 @@ namespace osu.Framework.Input.Handlers.Midi
                 {
                     scheduledRefreshDevices?.Cancel();
 
-                    foreach (var device in openedDevices.Values)
-                        closeDevice(device);
+                    lock (openedDevices)
+                    {
+                        foreach (var device in openedDevices.Values)
+                            closeDevice(device);
 
-                    openedDevices.Clear();
+                        openedDevices.Clear();
+                    }
                 }
             }, true);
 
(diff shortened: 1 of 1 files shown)


bdach:
I kind of can't resist asking the question where did that null come from there in the first place. Dictionaries can't contain null keys. That, to me, suggests that the dictionary was in an inconsistent state and this might indicate a deeper issue in that input handler's logic.

How did you hit this? Have you had a chance to look at the state of the `openedDevices` dictionary at the point of the crash? I think the more likely fail case is that a `refreshDevices()` call from this repeating scheduled delegate:

https://github.com/ppy/osu-framework/blob/ece2c882fb516a83b084c653e88a42959276e210/osu.Framework/Input/Handlers/Midi/MidiInputHandler.cs#L39

happened to race against a `Clear()` call of the `openedDevices` dictionary (mutated extensively by the device refresh) after disabling the handler:

https://github.com/ppy/osu-framework/blob/ece2c882fb516a83b084c653e88a42959276e210/osu.Framework/Input/Handlers/Midi/MidiInputHandler.cs#L48

or this synchronous call at the end of the method:

https://github.com/ppy/osu-framework/blob/ece2c882fb516a83b084c653e88a42959276e210/osu.Framework/Input/Handlers/Midi/MidiInputHandler.cs#L52

therefore crashing due to the lack of thread safety guarantees from the underlying dictionary.

peppy:
You have a good point, will reassess this.

peppy:
Have changed this to a simple lock. I think it should 100% cover the failure case I was seeing.

smoogipoo (approved):

PR #4276, opened 2021-03-12
Combine WindowsRawInputHandler and WindowsMouseHandler into a single implementation

peppy:
Having it as a separate handler made it a bit of a nightmare to configure things like sensitivity. Since it's intended to be an internal workaround/override, combining with `MouseHandler` itself works better than expected.

- [x] Depends on #4271 mostly for test scene usage.


osu.Framework/Platform/Windows/WindowsMouseHandler.cs
```diff
@@ -1,30 +1,123 @@
...
+using SDL2;
 
 namespace osu.Framework.Platform.Windows
 {
-    public class WindowsMouseHandler : MouseHandler
+    /// <summary>
+    /// A windows specific mouse input handler which overrides the SDL2 implementation of raw input.
+    /// This is done to better handle quirks of some devices.
+    /// </summary>
+    internal unsafe class WindowsMouseHandler : MouseHandler
```
bdach, line 19: If #4272 goes in, this will have `Description == "WindowsMouse"`.

bdach:
Am I reading this right, that windows won't be able to turn *off* raw input now? From an osu!-centric viewpoint I guess that might be fine, as I would hope that nobody was seriously playing osu! with mouse accel on, but it may be more questionable when viewing this issue from a game framework standpoint.

peppy:
turn off relative mode

osu.Framework/Platform/Windows/WindowsMouseHandler.cs
peppy: Fixed in #4272.

smoogipoo (approved):

PR #4277, opened 2021-03-12
Fix CursorInWindow not updating as expected

peppy:
It turns out with the introduction of relative mouse support in `MouseHandler`, we were toggling the state far more often than necessary. It was feasible that feedback loops when at the edge of the window could cause the state to flip-flop multiple times, causing what seems to be race conditions inside SDL or winapi. This is likely also the reason for https://github.com/libsdl-org/SDL/issues/4165 occurring (this change does greatly reduce the chance of this triggering, to the point I don't think users will ever encounter the issue).

FWIW, the underlying issue here is that the `SDL_WINDOWEVENT_ENTER` / `LEAVE` events stop arriving altogether.

@smoogipoo this resolves the issue you pointed out where the framework test cursor gets stuck at the edge of the window.

- [x] Depends on https://github.com/ppy/osu-framework/pull/4276 (because toggling raw input before this is a bit... weird).

### DIFF
--- a/osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
+++ b/osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
@@ -47,6 +47,7 @@ namespace osu.Framework.Input.Handlers.Mouse
         private Vector2? lastPosition;
 
         private IBindable<bool> isActive;
+        private IBindable<bool> cursorInWindow;
 
         /// <summary>
         /// Whether a non-relative mouse event has ever been received.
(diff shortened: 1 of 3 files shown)


bdach (approved):
off the source alone, seems like a reasonable resolution

osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
```diff
@@ -67,6 +68,9 @@ public override bool Initialize(GameHost host)
             isActive = window.IsActive.GetBoundCopy();
             isActive.BindValueChanged(_ => updateRelativeMode());
 
+            cursorInWindow = host.Window.CursorInWindow.GetBoundCopy();
+            cursorInWindow.BindValueChanged(_ => updateRelativeMode());
```
smoogipoo, line 72: Can this bindable be removed, or is an implementation missing here?

smoogipoo (changes requested):

osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
smoogipoo: Nevermind, I missed the line I was commenting on...

smoogipoo (dismissed):

smoogipoo (approved):

Issue #4206, opened 2021-02-20
Expose a "Resizable" Property on DesktopWindows


LeNitrous:
Our project Vignette is catered towards specifically for live streamers which means they will need a consistent and expected window size thus a resizable window would not be ideal. Forcing the game to always be in borderless full screen or exclusive full screen may also not be ideal in terms of user preference and experience.

SDL2 has `SDL.SetWindowResizable(IntPtr, SDL_bool)` but I'm not sure about OpenTK's counterpart.

peppy:
If we were to decide this is out of scope, are you able to call it directly yourself?

LeNitrous:
If by calling it directly myself without making changes in this repository, then no. The SDL window handle is marked as internal under `SDL2DesktopWindow` and I'm still looking for OpenTK's counterpart.

peppy:
Well you shouldn't be using OpenTK/osuTK anyway.

We might want to consider exposing the handle, not sure. Supporting fixed size windows may be something we want but definitely not an immediate priority.

LeNitrous:
Exposing the handle is definitely more than enough. Although using it inappropriately may cause certain internals to behave inappropriately which is the risk when done so. I could make a PR making the change in the accessor but it would still be the team's call whether it should be done or not.

PR #4279, opened 2021-03-13
Add Resizable property in SDLWindow

LeNitrous:
Closes https://github.com/ppy/osu-framework/issues/4206. Adds `Resizable` property in `SDLWindow`. Makes use of `SDL.SDL_SetWindowResizable(IntPtr, bool)`. This simply removes the ability of the window to be dragged in its frames to be resized so that consumers can programmatically control the window dimensions.


osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -90,6 +90,21 @@ public Point Position
             }
         }
 
+        private bool resizable;
```
peppy, line 93: isn't this going to be the wrong value at startup? probably needs to be read along with other window attributes.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -90,6 +90,21 @@ public Point Position
...
+
+        /// <summary>
+        /// Returns or sets whether the window is resizable or not.
+        /// </summary>
+        public bool Resizable
+        {
+            get => resizable;
+            set
+            {
+                resizable = value;
```
peppy, line 103: missing early return, see implementation of other properties in this class for reference.

peppy (dismissed):

bdach:
This should have test coverage, at least manual, given we're not likely to be using this much on the game side.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -90,6 +90,24 @@ public Point Position
             }
         }
 
+        private bool resizable;
+
+        /// <summary>
+        /// Returns or sets whether the window is resizable or not.
+        /// </summary>
+        public bool Resizable
```
bdach, line 98: [As stated in the SDL docs](https://wiki.libsdl.org/SDL_SetWindowResizable), setting this is invalid if fullscreen:

> Note: You can't change the resizable state of a fullscreen window.

This will therefore silently fail (sets the field successfully, but then doesn't actually do anything on the SDL side and reverts to the actual value on the next `updateWindowSpecifics()`). I'm not sure it should be allowed to call this without some kind of error in that case. Either that, or it should be ensured that any potential sets of this property aren't lost (and the appropriate commands are executed the next time the window changes state to windowed).

bdach (changes requested):

osu.Framework/Platform/SDL2DesktopWindow.cs
LeNitrous: I can just set the default value for this as `true` since its guaranteed that we will always have a resizable window on startup. Unless we choose to expose this to the user as a framework config variable.

https://github.com/ppy/osu-framework/blob/4b4dc843aff09dc71b92dbb4cc6606a74eafb94f/osu.Framework/Platform/SDL2DesktopWindow.cs#L411-L417

To ensure that `Resizable` is still the correct value, I can just call `SDL.SDL_SetWindowResizable` again after the window state has been changed to `Normal`.

LeNitrous:
Should this be a config variable in `framework.ini`? Or should we just leave it hidden to the end user?

peppy:
should not be exposed

LeNitrous:
I'm rather unsure of what you mean by "manual". Would a test scene that has steps on toggling the resizable state be enough?

bdach:
A test scene with a manual toggle is good enough for me, yes.

LeNitrous:
Added the test scene. Something to note on Windows, The Minimize/Maximize button gets greyed out once `Resizable = false`.

osu.Framework/Platform/SDL2DesktopWindow.cs
bdach: > I can just set the default value for this as true since its guaranteed that we will always have a resizable window on startup.

That's kind of unrelated to what I was asking about...?

> To ensure that `Resizable` is still the correct value, I can just call `SDL.SDL_SetWindowResizable` again after the window state has been changed to `Normal`.

That would be one of the ways I proposed to fix this, yes. Note that this would mean that the semantics of `Resizable` we expose would now differ from SDL's (in fullscreen, SDL says that the window is *not* resizable - so it's a byproduct of `SetWindowResizable` calls and the window mode, while ours would be an internal state variable, that doesn't change value with window mode changes, but is only valid in windowed). It should be documented as such.

osu.Framework.Tests/Visual/Platform/TestSceneResizableWindow.cs
```diff
@@ -0,0 +1,17 @@
...
+using osu.Framework.Platform;
+
+namespace osu.Framework.Tests.Visual.Platform
+{
+    public class TestSceneResizableWindow : FrameworkTestScene
+    {
+        [BackgroundDependencyLoader]
+        private void load(GameHost host)
+        {
+            AddToggleStep("toggle resizable", (state) => ((SDL2DesktopWindow)host.Window).Resizable = state);
```
bdach, line 14: see CI failure
```suggestion
            AddToggleStep("toggle resizable", state => ((SDL2DesktopWindow)host.Window).Resizable = state);
```

osu.Framework/Platform/SDL2DesktopWindow.cs
LeNitrous: > That's kind of unrelated to what I was asking about...?

Pardon me. I was referring to removing this:

https://github.com/ppy/osu-framework/blob/4b4dc843aff09dc71b92dbb4cc6606a74eafb94f/osu.Framework/Platform/SDL2DesktopWindow.cs#L993

Because inside `SDLDesktopWindow.Create()`, one of the flags set when creating the window is `SDL_WindowFlags.SDL_WINDOW_RESIZABLE`. From that, I can naively assume that we have a resizable window from the start.

In other things, I have updated the xmldocs for the property to note this behavior.

osu.Framework.Tests/Visual/Platform/TestSceneResizableWindow.cs
```diff
@@ -0,0 +1,24 @@
...
+using osu.Framework.Platform;
+
+namespace osu.Framework.Tests.Visual.Platform
+{
+    public class TestSceneResizableWindow : FrameworkTestScene
+    {
+        [BackgroundDependencyLoader]
+        private void load(GameHost host)
+        {
+            if (host.Window == null)
```
peppy, line 15: You're looking for `[HeadlessTest]` as an attribute on this class.

smoogipoo: It's the other way around, this class requires a window.

smoogipoo (approved):
lgtm

smoogipoo (dismissed):

bdach (approved):
seems ok

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -90,6 +90,24 @@ public Point Position
...
+        {
+            get => resizable;
+            set
+            {
+                if (resizable == value)
+                    return;
+
+                resizable = value;
+                ScheduleCommand(() => SDL.SDL_SetWindowResizable(SDLWindowHandle, value ? SDL.SDL_bool.SDL_TRUE : SDL.SDL_bool.SDL_FALSE));
```
smoogipoo, line 107: What happens if the window is already fullscreen? Is this just a no-op? Perhaps this should check the window state?

peppy: Should be safe enough:

20210316 120844 (iTerm2)

smoogipoo (approved):

PR #4280, opened 2021-03-13
Move SpriteText's bind logic to LoadComplete to alleviate potential threading issues

peppy:
This comes as the result of @smoogipoo discovering a deadlock osu!-side, which can be considered a regression of the localisation changes we made recently. The case is where a `Bindable` is assigned to `text.Current` during BDL (in a `LoadComponentAsync` context) while the `SpriteText` is also changing the value of `current` via the localised implementation. This could cause a deadlock in `GetEnumerator` as `Bindable`s are not intended to be thread safe for this kind of usage.

Example case:

thread 1:

20210313 235040 (Parallels Desktop app)

thread 2:

20210313 235136 (Parallels Desktop app)
...
20210313 235155 (Parallels Desktop app)

### DIFF
--- a/osu.Framework/Graphics/Sprites/SpriteText.cs
+++ b/osu.Framework/Graphics/Sprites/SpriteText.cs
@@ -61,6 +61,21 @@ namespace osu.Framework.Graphics.Sprites
         private void load(ShaderManager shaders)
         {
             localisedText = localisation.GetLocalisedString(text);
+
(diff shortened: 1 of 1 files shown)


bdach:
Registering value change callbacks in BDL seems like a programming error in the first place, so it looks like I shouldn't need to think too much about merging this...

Just to confirm because it's not 100% clear from just the stack screenshots, what was executing on the `SpriteText` BDL side? Was it the immediate execution of the `localisedText` value change callback?

bdach (approved):

PR #4282, opened 2021-03-16
Make ConfigManager.Set protected to avoid accidentally setting default values

peppy:
### ConfigManager.Set is now protected

Previously this method was `public` for convenience, but as it implicitly set the `Default` value of bindables it touched, could cause unexpected behaviour.

- For initialising defaults, continue to use `Set` from `InitialiseDefaults`
- For changing configuration values externally, first retrieve the `Bindable` and then change its value, ie.

```csharp
// previously
config.Set(ConfigType.Name, true);

// now
config.GetBindable<bool>(ConfigType.Name).Value = true;
```

### DIFF
--- a/osu.Framework/Configuration/ConfigManager.cs
+++ b/osu.Framework/Configuration/ConfigManager.cs
@@ -40,7 +40,7 @@ namespace osu.Framework.Configuration
         {
         }
 
-        public BindableDouble Set(TLookup lookup, double value, double? min = null, double? max = null, double? precision = null)
+        protected BindableDouble Set(TLookup lookup, double value, double? min = null, double? max = null, double? precision = null)
         {
             value = getDefault(lookup, value);
 
@@ -62,7 +62,7 @@ namespace osu.Framework.Configuration
             return bindable;
         }
 
-        public BindableFloat Set(TLookup lookup, float value, float? min = null, float? max = null, float? precision = null)
+        protected BindableFloat Set(TLookup lookup, float value, float? min = null, float? max = null, float? precision = null)
         {
             value = getDefault(lookup, value);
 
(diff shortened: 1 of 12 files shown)


osu.Framework/Configuration/ConfigManager.cs
```diff
@@ -40,7 +40,7 @@ protected virtual void InitialiseDefaults()
         {
         }
 
-        public BindableDouble Set(TLookup lookup, double value, double? min = null, double? max = null, double? precision = null)
+        protected BindableDouble Set(TLookup lookup, double value, double? min = null, double? max = null, double? precision = null)
```
bdach, line 43: Making these protected is definitely way better (i wasn't even at all aware these were touching defaults at all.............) but the name is still a little dicey. I still probably wouldn't be expecting this to touch defaults.

I think it was proposed somewhere in discord to name these `SetDefault()` or something already? I think I'd advocate for that. Well, either that, or xmldoc these. But a rename would be better.

peppy: The idea was to avoid breaking every implementation, but it is a simple fix so maybe we should just do that? I'm fine either way.

bdach: I mean it looks to be a pretty huge breaking change already, I think? Probably best to confirm with @smoogipoo first before jumping headlong into the rename on top of that.

peppy: Currently it only breaks on external usage (which depending on how that project was designed, may be large or small), but renaming the method would also break every `InitialiseDefaults` implementation.

If we're going to rename the method, may be worth considering adding a `public void SetValue` back as well, not sure.

smoogipoo: `Set()` is seldom used outside of tests, so this is a pretty minor breaking change. Renaming `Set()` altogether is a much more breaking one because of the aforementioned reason.

peppy: So we'll just go with how this is currently then?

smoogipoo: Yes, this is fine to keep breaking changes minimal.

smoogipoo (approved):

PR #4283, opened 2021-03-16
Fix SDL2 mouse position warping not respecting DPI scaling

frenzibyte:
Noticed it while I was trying the relative mouse mode, was causing mouse to jump far away from the window when exiting in retina screens:


peppy (approved):

smoogipoo (approved):
Looks reasonable

PR #4284, opened 2021-03-17
Expose required bindables via ITabletHandler interface for end-user configurability

peppy:
This exposes the bare minimum required for osu!-side configuration. Intended to be an initial MVP. Depending on how quickly this is merged, I may also add rotation.

Note that the `TabletInfo` class is required to ferry information away from the .netcore OTD library and avoid conditional compilation nightmares. It may eventually disappear.

- [x] Depends on #4285.

### DIFF
--- a/osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
+++ b/osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
@@ -29,6 +29,8 @@ namespace osu.Framework.Input.Handlers.Tablet
 
         public IBindable<Size> TabletSize { get; } = new BindableSize(new Size(160, 100));
 
+        public string DeviceName => tabletDriver.Tablet?.TabletProperties.Name ?? string.Empty;
+
         public override bool Initialize(GameHost host)
         {
             tabletDriver = new TabletDriver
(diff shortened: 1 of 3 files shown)


osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
```diff
@@ -93,21 +104,39 @@ private void updateOutputArea(IWindow window)
...
+                AreaSize.SetDefault();
+
+            AreaOffset.Default = new Vector2(inputWidth / 2, inputHeight / 2);
+
+            // likewise with the position, use the centre point if it has not been configured.
+            // it's safe to assume no user would set their centre point to 0,0 for now.
+            if (AreaOffset.Value == Vector2.Zero)
+                AreaOffset.SetDefault();
+
+            tablet.Value = new TabletInfo(tabletDriver.Tablet.TabletProperties.Name, AreaSize.Value);
```
smoogipoo, line 128: Shouldn't this be `AreaSize.Default`?

smoogipoo (changes requested):

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
peppy: good catch.

would appreciate if you can test this with a tablet, i still don't have a test device handy. should be usable with my osu-side configuration branch still.

smoogipoo (approved):

PR #4286, opened 2021-03-17
Show tooltip with full text when sprite text is truncated

Joehuu:
Resolves https://github.com/ppy/osu/issues/6120.

xACKTQO9I5

### DIFF
--- a/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteTextTruncate.cs
+++ b/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteTextTruncate.cs
@@ -3,6 +3,7 @@
 
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Cursor;
 using osu.Framework.Graphics.Shapes;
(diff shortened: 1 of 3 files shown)


peppy:
Not 100% on this being at the framework side. If it is, it will need to be a toggle (likely defaulting to off). @smoogipoo thoughts?

smoogipoo:
My biggest issue with this is the performance impact. Every `SpriteText` now counts for positional input, which negates the `RequestsPositionalInput` optimisation.

peppy:
@Joehuu make sure if you visit this to note the performance concerns raised by @smoogipoo as it is a very serious issue (we use a lot of `SpriteText` and having each and every one in the input queue is going to add a game-wide overhead we definitely want to avoid).

I think there should be a way to avoid the overhead, but you'll need to actively understand why it occurs, and find a way around it. See:

https://github.com/ppy/osu-framework/blob/9bbc6a81a2b8f035e2b043d3718d2db6e4f1d868/osu.Framework/Graphics/Drawable.cs#L2413-L2419

Joehuu:
Superseded by https://github.com/ppy/osu-framework/pull/5636.

PR #4287, opened 2021-03-17
Add public `SetValue()` method to `ConfigManager`

smoogipoo:
1. Public `Set()` methods added back as obsolete.
2. Protected `Set()` methods renamed to `SetDefault()`.
3. Public `SetValue()` method added as a shortcut to `GetBindable<>().Value = value`.  This also sets the default if one hasn't already been set, as a conscious decision in order to not throw exceptions.


peppy:
Might have to get this one too https://github.com/ppy/osu-framework/pull/4282/commits/7ab4926473bc4eb3c63eb1fcbf49c63a4bbdf67d

peppy (approved):

PR #4291, opened 2021-03-19
Switch audio subsystem to use BassMix

nekodex:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/4627

Can be tested osu!-side using https://github.com/ppy/osu/pull/14047

---

Some details on structure:

- **`IAudioMixer`** - Common interface between `DrawableAudioMixer` and `AudioMixer`.
- **`IAudioChannel`** - A mostly internal interface between `SampleChannel` and `Track`, which serves as the type that can be added to an `AudioMixer`.
- **`AudioMixer`** - The base audio mixer type. I could've combined this with `BassAudioMixer`, but that doesn't sit right with me in the context that we are not exposing "Bass" things to the user. This way if we ever switch to another audio library, we can derive this type again.
- **`IBassAudioChannel`** - A mostly internal interface between `SampleChannelBass` and `TrackBass`. It provides a data between these types and `IBassAudioMixer`.
- **`BassAudioMixer`** - The internal type which contains the actual BASSmix implementation.
---

## Mixing audio

In osu!framework, all `SampleChannel` and `Track` audio (hereby referred to as a "channel") is routed through an `AudioMixer`. DSP effects can be applied that can change the audio of channels mixed by it independent of other `AudioMixers` throughout the game.

### Global audio mixer

The global audio mixer affects all channels by default and can be accessed through `AudioManager.Mixer`. The following is a simple example of how to apply a DSP effect globally:
```csharp
[BackgroundDependencyLoader]
private void load(AudioManager audio, ISampleStore samples)
{
    // Add an effect to the global mixer.
    audio.Mixer.Effects.Add(new ...);

    // The sample has the above effect applied to it by default.
    samples.Get(...).Play();

    // The same is true for DrawableSample/DrawableTrack.
    DrawableSample drawableSample;
    Add(drawableSample = new DrawableSample(samples.Get(...)));
    drawableSample.Play();
}
```

### Local (custom) audio mixers

The global mixer provides too wide of an effect for use in many cases. There are two ways to create a mixer for only the specific audio which needs to be affected.

One option is to create a locally managed `AudioMixer` through `AudioManager.CreateAudioMixer()`:
```csharp
private AudioMixer mixer;

[BackgroundDependencyLoader]
private void load(AudioManager audio, ISampleStore samples)
{
    // Create the custom mixer.
    mixer = audio.CreateAudioMixer();

    // Add an effect to the mixer.
    mixer.Effects.Add(new ...);

    Sample sample = samples.Get(...);

    // Add channels to the custom mixer in order to apply the effect to them.
    SampleChannel channel = sample.GetChannel();
    mixer.Add(channel);
    channel.Play();

    // Channels that are not added to the custom mixer will only be affected by the global mixer.
    sample.Play();
}

{
    base.Dispose(isDisposing);

    // Be sure to dispose the mixer, otherwise memory will be leaked!
    mixer?.Dispose();
}
```

The other option is to use `DrawableAudioMixer`, which takes care of the hard work behind the scenes:
```csharp
[BackgroundDependencyLoader]
private void load(ISampleStore samples)
{
    DrawableAudioMixer drawableAudioMixer;
    DrawableSample drawableSample;

    // Add a mixer to the hierarchy, this will affect DrawableSample/DrawableSample children added to it, recursing until another DrawableAudioMixer is reached.
    // Note: it will not affect Sample/SampleChannel/Tracks!
    Add(drawableAudioMixer = new DrawableAudioMixer
    {
        Child = drawableSample = new DrawableSample(samples.Get(...)),
        Effects =
        {
            ...
        }
    });

    // This sample has the above effect applied to it.
    drawableSample.Play();

    // As above, you can still add channels to the mixer manually to apply the effect to them.
    Sample sample = samples.Get(...);
    SampleChannel channel = sample.GetChannel();
    drawableAudioMixer.Add(channel);
    channel.Play();
}
```

### Effect prioritisation

Effects are prioritised by the order that they appear in the `AudioMixer.Effects` or `DrawableAudioMixer.Effects` list. Effects can be added, removed, or moved around in order to change their priority.

### Changing the audio mixer

Channels can be moved to another audio mixer by adding or removing them from the `AudioMixer`:
```csharp
[BackgroundDependencyLoader]
private void load(ISampleStore samples)
{
    DrawableAudioMixer mixer1;
    DrawableAudioMixer mixer2;

    Children = new Drawable[]
    {
        mixer1 = new DrawableAudioMixer(),
        mixer2 = new DrawableAudioMixer()
    };

    // Add the channel to the first mixer.
    SampleChannel channel;
    mixer1.Add(channel = samples.Get(...).Play());

    // Move the channel to the second mixer.
    mixer2.Add(channel);

    // Remove the channel from the second mixer.
    // Note: The channel will be moved to the global mixer, not mixer1 from above!
    mixer2.Remove(channel);

    // This has no effect. All channels must be played by one mixer.
    audio.Mixer.Remove(channel);



    // Similarly, for DrawableSample:
    DrawableSample drawableSample = new DrawableSample(samples.Get(.));
    SampleChannel channel1 = drawableSample.Play();
    SampleChannel channel2 = drawableSample.Play();

    // Add the entire sample to the first mixer, this affects both played channels.
    mixer1.Add(drawableSample);

    // Move the entire sample to the second mixer, this affects both played channels.
    mixer1.Remove(drawableSample);
    mixer2.Add(drawableSample);

    // Move the first channel to the first mixer.
    mixer1.Add(channel1);

    // Move the second channel to the global mixer.
    mixer2.Remove(channel2);

    // Move the entire sample to the first mixer, this still affects both played channels.
    mixer2.Remove(drawableSample);
    mixer1.Add(drawableSample);
}
```

---

Original by @nekodex (changes by me where relevant):

This is a mixer-based audio output system (using BASSmix) that came about from my trying to alleviate some pain points I encountered while doing sound design for lazer - namely: audio levels being all over the place, no way to visualize audio output, no way to apply effects globally, etc.

The core idea is that all audio (i.e. tracks and samples) gets routed via a global 'mixer' instead of being output directly, allowing for effects and volume monitoring/limiting/etc to be done centrally.

Some of the main benefits of this approach are:
- The ability to utilize WASAPI exclusive mode output and all the benefits that entails (this fulfils the requirement of pre-mixing audio into a buffer, which can then be sent to an audio device)
- Being able to apply a master compressor on the post-mix output to limit overall output volume (this was the main reason I started working on this, not sure if that's the direction we want to go... needs more experimentation, but the option is there)
- The ability to apply effects (low-pass, band-pass, reverbs, etc) at a global level (e.g. to add flair to screen transitions, etc)
- Being able to monitor the audio levels of all playing tracks and samples
  - **smoogi: Overlay has been removed for the time being.**

~~While this proof-of-concept works in its current form (both in framework tests, [and in lazer with this branch](https://github.com/nekodex/osu/tree/bassmix)), I'm not particularly happy with how I've gone about implementing it... so I'm creating this PR more as a starting point for discussion about the future of audio in the framework - and also to gather feedback on potential issues that I may have overlooked.~~

General thoughts:
- ~~Probably better to do this all via depedency injection instead of how I've passed `AudioMixer` around everywhere?~~
  - **smoogi: Done.**
- ~~Maybe we don't want to be throwing around channel handles when interacting with the `AudioMixer`?~~
  - **smoogi: Done.**
- ~~Perhaps we should structure things using sub-mixers instead of a single main mixer (e.g. a submixer for samples that allows effects/compressors to only affect sample playback... or maybe even go deeper and support nested mixers similar to the drawable hierarchy?)~~
  - **smoogi: Done.**

Things to note:
- Relies on unreleased features in the [beta build of BASS](http://www.un4seen.com/forum/?topic=19224.msg134549#msg134549) ([which adds a hybrid mode for samples](http://www.un4seen.com/forum/?topic=19160.msg134251#msg134251), a requirement for routing samples into mixers) -- I've temporarily created a [ManagedBass fork](https://github.com/nekodex/ManagedBass/tree/beta) with the updated function signature for `BASS_SampleGetChannel` and the new flags required to use stream/hybrid mode
- ~~A 'master compressor' is enabled by default with parameters that didn't sound too bad for lazer, but this will behave differently depending on context and could sound horrible depending on what music and/or samples are playing (this is toggle-able in the mixer overlay, CTRL+F9)~~
  - **smoogi: Master compressor no longer applied by default, overlay removed for the time being.**

Known issues:
- ~~(Probably) doesn't handle device switching yet, but wouldn't be hard to add~~
  - **smoogi: Fixed.**
- ~~Some of the BASS-wrapping functions in `AudioMixer` aren't protected from being run on non-audio threads, which needs to be fixed~~
  - **smoogi: Fixed.**
- AFAIK sample/channel lifecycles are handled internally by BASS itself, but I'm currently tracking them to attach/detach to the mixer and show in the mixer overlay... not sure on the best way to handle cleanup here as there doesn't seem to be obvious relevant callbacks from BASS when this happens? This means currently the osu!direct preview tracks leave references around which prevent them from being disposed properly
  - **smoogi: Overlay has been removed for the time being.**
- The mixer gets created using the Float flag, [which a past PR suggests was problematic when used with streams](https://github.com/ppy/osu-framework/pull/2128) - I'm not sure if this also applies to mixers, as I don't seem to be affected by the described problem

Probably goes without saying that this will require extensive testing across all platforms to see if playback and audio latency is negatively affected.
  - ~~**smoogi: Audio latency seems to be ~100ms (linux) and affected by the `PlaybackBufferLength` (adjustable in `AudioManager`). Needs testing on other platforms.**~~
  - **smoogi: Latency issues seem resolved.**


osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -208,7 +215,7 @@ protected override void UpdateState()
             // because device validity check isn't done frequently, when switching to "No sound" device,
             // there will be a brief time where this track will be stopped, before we resume it manually (see comments in UpdateDevice(int).)
             // this makes us appear to be playing, even if we may not be.
-            isRunning = running || (isPlayed && !hasCompleted);
+            isRunning = /*running ||*/ (isPlayed && !hasCompleted);
```
peppy, line 218: how does this work with the commented out variable?

peppy:
On a quick pass, this looks about how I would have expected to implement it. I'm not sure about using DI, because as you may have noticed the audio hierarchy doesn't really support DI yet.

Will go through in more detail over the coming week.

osu.Framework/Audio/Sample/SampleChannelBass.cs
```diff
@@ -149,13 +151,13 @@ public override void Stop()
...
 
-            Bass.ChannelPlay(channel);
+            mixer.PlayChannel(channel);
         });
 
         private void stopChannel() => EnqueueAction(() =>
         {
             if (hasChannel)
-                Bass.ChannelPause(channel);
+                Dispose();
```
smoogipoo, line 160: Is this correct? The intention is to allow pausing and continuing playback, but this completely stops and removes the channel, which I imagine removes this ability?

Seems like `mixer.PauseChannel` exists for this purpose?

smoogipoo (commented):
Haven't done a full review, just a quick one that I noticed.

osu.Framework/Audio/Sample/SampleChannelBass.cs
nekodex: This is a hack to prevent samplechannels leaking, so yeah it does break pausing/resuming of samples... (I don't think I encountered pausing/resuming of samples anywhere in lazer though FWIW)

As you suggest, the intention is to use `mixer.PauseChannel`, but the lifecycle stuff kinda needs to be fixed first.

osu.Framework/Audio/Track/TrackBass.cs
nekodex: Oops, I meant to delete that, not leave it commented out.

IIRC because the channel is now a decoding channel instead of a playback one, the channel state is always "Playing" even when playback is actually paused... so the removal of the `running` check means it falls back to simply whether `isPlayed` is set (which occurs when play/pause happens)

peppy: Right, if that's the case the actual call above can also be removed.

peppy:
Weirdly, on this branch `libbass` and `libbassfx` (along with other libraries) aren't getting correctly placed in build folders (left: `osu.Framework.Tests` right: `NativeLibs`).. Something to do with permissions or the framework target maybe?

20210409 154831 (Finder)

Have confirmed this doesn't match expectations on `master`.

osu.Framework/Audio/AudioManager.cs
```diff
@@ -266,6 +273,14 @@ private bool setAudioDevice(int deviceIndex)
             //we have successfully initialised a new device.
             UpdateDevice(deviceIndex);
 
+            Mixer.Init();
```
peppy, line 276: Is there any reason why this is done here and not in the constructor of the mixer itself? That would allow us to remove the weird "log but not fail" errors when attempting to operate on a mixer that hasn't been initialised (see https://github.com/ppy/osu-framework/pull/4291/files#diff-4b61a46f418cd72df6a5763d0e40061689a77181125bc3d788a2a0ce5e8c5910R40)

osu.Framework/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,279 @@
...
+                {
+                    Logger.Log($"[AudioMixer] Attempted to add channel ({channelHandle}) when mixer not yet initialized!");
+                    return;
+                }
+
+                if (MixChannels.Contains(channelHandle))
+                {
+                    Logger.Log($"[AudioMixer] Channel ({channelHandle}) already added!");
+                    return;
+                }
```
peppy, line 48: Should be done outside the `EnqueueAction` and `throw` instead.

osu.Framework/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,279 @@
...
+                Logger.Log($"[AudioMixer] ChannelPlay(mixer): {Bass.LastError}");
+
+                EnableCompressor();
+                // EnableLimiter();
+            });
+        }
+
+        protected override void UpdateState()
+        {
+            EnqueueAction(() =>
```
peppy, line 255: Pretty sure `UpdateState` is guaranteed to be called on the audio thread and this is unnecessary.

osu.Framework/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,279 @@
...
+            {
+                var channels = MixChannels.ToArray();
```
peppy, line 257: We definitely don't want to be doing this kind of allocation on every update, that's for sure.

osu.Framework/Audio/Sample/SampleChannelBass.cs
```diff
@@ -207,7 +211,8 @@ protected override void Dispose(bool disposing)
 
             if (hasChannel)
             {
-                Bass.ChannelStop(channel);
+                mixer.StopChannel(channel);
+                mixer.RemoveChannel(channel);
```
peppy, line 215: Would hope that just calling `Remove` is enough here.

osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -54,17 +54,22 @@ public sealed class TrackBass : Track, IBassAudio
...
         /// <param name="data">The sample data stream.</param>
         /// <param name="quick">If true, the track will not be fully loaded, and should only be used for preview purposes.  Defaults to false.</param>
-        public TrackBass(Stream data, bool quick = false)
+        /// <param name="mixer">The <see cref="AudioMixer"/> to plug this Track into.</param>
+        public TrackBass(Stream data, bool quick = false, AudioMixer mixer = null)
         {
             if (data == null)
                 throw new ArgumentNullException(nameof(data));
 
+            if (mixer != null)
```
peppy, line 70: Null check unnecessary. Also, what's the case for the mixer coming in null? Are tracks usable in this state? It doesn't look like they would be.

osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -257,6 +265,8 @@ protected override void Dispose(bool disposing)
         public override void Stop()
         {
             base.Stop();
+
+            mixer.PauseChannel(activeStream);
```
peppy, line 269: Was this intentional? Looks like a misfire.

osu.Framework/osu.Framework.csproj
```diff
@@ -42,9 +43,9 @@
     <PackageReference Include="ppy.SDL2-CS" Version="1.0.225-alpha" />
   </ItemGroup>
   <ItemGroup Condition="$(TargetFrameworkIdentifier) == '.NETCoreApp'">
+    <ProjectReference Include="..\osu.Framework.NativeLibs\osu.Framework.NativeLibs.csproj" />
```
peppy, line 46: The main thing holding back this PR right now is that testing is a huge pain due to this not actually working. Probably best to just revert this for now (to the package reference) and potentially release a package with the pre-release bass version.

peppy (changes requested):

peppy:
@nekodex it doesn't look like you're using your `ManagedBass` fork any more. Is that intentional? Is it someone we still need to consider before merging this?

osu.Framework/osu.Framework.csproj
peppy: I've pushed a [prerelease version](https://www.nuget.org/packages/ppy.osu.Framework.NativeLibs/2021.504.0-alpha) of `NativeLibs` with up-to-date libraries (pre-release where available, otherwise latest on release page) and targeted it.

peppy:
I've fixed all the remaining tests (except for one related to zero frequency stuff, which I'm not even looking into because the logic is shoddy), but in doing so noticed that commands which we are used to having applied immediately are not not applied until the next `Bass.Update`. This means that the latency of immediate sample playback could be higher than previously unless we correctly manage `Bass.Update` calls.

We will need to do some further investigation on this, but it sounds like we may want to either have invalidation (and force an early call at the end of the audio thread's update loop), or just always invoke manually.

osu.Framework/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,272 @@
...
+                var channel = MixChannels[i];
+
+                if (Bass.ChannelIsActive(channel) == PlaybackState.Stopped)
+                {
+                    // HACK: avoid auto-cleanup of TrackBass channels - they are "Reverse" thanks to the tempo and reverse fx chain they have
+                    if (Bass.ChannelGetInfo(channel).ChannelType == ChannelType.Reverse) return;
+
+                    Logger.Log($"[AudioMixer] Channel gone, auto-removing ({channel})");
+                    RemoveChannel(channel);
+                    i--;
```
peppy, line 263: This is completely wrong and randomly removes looping samples. @nekodex do you have a reference for where this behaviour is documented / why you did it this way?

osu.Framework/Audio/Sample/SampleChannelBass.cs
peppy: This is causing the remaining looping test failure (zero frequency handler calls stop and disposes the channel).

smoogipoo: What do you mean by "lifecycle stuff", before we get too deep in this?

osu.Framework/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,272 @@
...
+                DisableLimiter();
+            else
+                EnableLimiter();
+        }
+
+        #endregion Limiter
+
+        #region Filter
+
+        public void EnableFilter()
```
smoogipoo, line 196: Are these (limiter/compressor/filter) just some arbitrary defaults? Or are the parameters some well-defined constants in the audio world? If they aren't we should probably just move these to where they're required - tests and OsuGame.

Or to put it another way, are the parameters supposed to be adjusted by the users in a similar way to how the test scene does?  
As I said previously, this stuff is pretty raw. I think that it would be better to expose a method to attach an `IEffectParameter`, that way all the `Bass` calls don't have to be done by the user.

osu.Framework/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,272 @@
...
+        private int filterHandle;
+
+        // the order in which FX are applied - higher number = earlier in chain, i.e. filter -> compressor -> limiter
+        private const int limiter_priority = 1;
+        private const int compressor_priority = 2;
+        private const int filter_priority = 3;
+
+        private const int frequency = 44100;
+
+        public void AddChannel(int channelHandle, bool addPaused = false)
```
smoogipoo, line 33: I also don't like all of these `int channelHandle`s either. If these were `internal` methods then I could get past this.

osu.Framework/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,272 @@
...
+        private int compressorHandle;
+        private int limiterHandle;
+        private int filterHandle;
+
+        // the order in which FX are applied - higher number = earlier in chain, i.e. filter -> compressor -> limiter
+        private const int limiter_priority = 1;
+        private const int compressor_priority = 2;
+        private const int filter_priority = 3;
+
+        private const int frequency = 44100;
```
smoogipoo, line 31: Will have to see if this causes any problem for those that use non-44KHz devices.

osu.Framework/Audio/AudioMixer.cs
smoogipoo: Agree - this should really not be required at all, seems like it's maybe just a hack for the test scene?

osu.Framework/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,272 @@
...
+                    return;
+                }
+
+                if (MixChannels.Contains(channelHandle))
+                {
+                    Logger.Log($"[AudioMixer] Channel ({channelHandle}) already added!");
+                    return;
+                }
+
+                BassFlags flags = addPaused ? BassFlags.MixerChanPause | BassFlags.MixerChanBuffer : BassFlags.MixerChanBuffer;
```
smoogipoo, line 49: Is my understanding correct that setting `BassFlags.MixerChanPause` causes the channel to be paused by default?
If so, in that case, don't `SampleChannelBass`'s play by default (via `sample.GetChannel()` without playing manually)? I think all channels should always be paused by default, unless I'm misunderstanding this.

osu.Framework/Audio/AudioMixer.cs
```diff
@@ -0,0 +1,272 @@
...
+        {
+            BassMix.ChannelFlags(channelHandle, BassFlags.MixerChanPause, BassFlags.MixerChanPause);
+
+            return Bass.LastError == Errors.OK;
+        }
+
+        public void StopChannel(int channelHandle)
+        {
+            BassMix.ChannelFlags(channelHandle, BassFlags.Default, BassFlags.MixerChanPause);
+            Bass.ChannelSetPosition(channelHandle, 0); // resets position and also flushes buffer
```
smoogipoo, line 89: How does this work? Isn't this equivalent to `PlayChannel` + seek to start? It would have to set flags to `BassFlags.MixerChanPause`, no?

osu.Framework/Audio/AudioManager.cs
smoogipoo: Absolutely, maybe do something similar to `SampleBassFactory()` where it enqueues an action in its ctor. Also need to consider the ability to change devices - that is, to use the above `UpdateDevice()` call by implementing the `IBassAudio` interface in `AudioMixer`.

smoogipoo (commented):
These are mostly notes and a few questions for now as I'm trying to understand the whole process and building a picture for how it should look.

smoogipoo:
Just thinking out loud again here, but what is the point of the `AudioMixer` class? Is it intended to be used in a way that you can attach an `AudioMixer` at some point local in the hierarchy? Because if not, maybe it's more worthwhile combining it into `AudioManager`?

osu.Framework.Tests/Visual/Audio/TestSceneBassMix.cs
```diff
@@ -0,0 +1,277 @@
...
+                };
+            }
+
+            // Create Mixer
+            mixerHandle = BassMix.CreateMixerStream(44100, 2, BassFlags.MixerNonStop);
+            Logger.Log($"[BASSDLL] CreateMixerStream: {Bass.LastError}");
+            // Make Mixer Go
+            Bass.ChannelPlay(mixerHandle);
+            Logger.Log($"[BASSDLL] ChannelPlay: {Bass.LastError}");
+            channelHandles[num_mix_channels - 1] = mixerHandle;
```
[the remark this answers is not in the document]
peppy: If you're not happy with this existing just as a proof-of-concept (not intended consumer usage), I think this should be moved to another PR. I agree it's too raw and I wasn't even considering it as part of the review, apart from showing how stuff can be setup in a future effort.

osu.Framework/Audio/AudioMixer.cs
peppy: I think it's fine to initially make these `internal`.

osu.Framework/Audio/AudioMixer.cs
peppy: This is just a test method and should be removed, along with the `FilterEnabled` flag.

peppy:
That is the intention. The idea is that you'd have an `AudioMixer` used for gameplay, to be able to apply effects to only subcomponents. ie. if you want a low-pass filter that is only applied to the audio track but not gameplay samples, or vice-versa.

I *believe* @nekodex was already making use of this to apply limiters to only sound effects?

smoogipoo:
This structure as it is doesn't really lend itself to that. Note how `GetSampleStore()`/`GetTrackStore()` pass in the audio manager's mixer, so there's only one that can ever really be as it currently is.

So the next question is... We don't have a hierarchical model for audio, only within the draw hierarchy. I can imagine two potential APIs:

### Add channels manually to the mixer
```
private ISample sample;
private AudioMixer audioMixer;

private void load(AudioManager audioManager, SampleStore samples)
{
    audioMixer = audioManager.CreateAudioMixer();
    sample = samples.GetSample(...);
}

public void Click() // Arbitrary demonstrative method
{
    SampleChannel channel;
    audioMixer.Add(channel = sample.GetChannel());
    // At some point later
    channel.Play();

    // Or...
    audioMixer.Add(sample.Play()); // Because Play() returns the channel
}
```

### Use the draw hierarchy for the parent mixer

private DrawableSample sample;

private void load(SampleStore samples)
{
    InternalChild = new AudioMixer
    {
        Child = sample = new DrawableSample(samples.Get(...))
    }
}

public void Click() // Arbitrary demonstrative method
{
    SampleChannel channel = sample.GetChannel();
    // At some point later
    channel.Play();

    // Or...
    sample.Play();

    // Both instances automatically add the created channel to the parenting audio mixer.
}

---

Perhaps of course, a combination of the two to allow both scenarios.

osu.Framework.Tests/Visual/Audio/TestSceneBassMix.cs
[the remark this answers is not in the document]
nekodex: I believe this was the initial proof-of-concept to get BassMix working with ManagedBass inside o!f, (i.e. before it was integrated into AudioManager) -- it's probably best to just get rid of this test case altogether to prevent people from using this as a reference example for their own projects...

osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -203,12 +211,10 @@ protected override void UpdateState()
         {
             base.UpdateState();
 
-            var running = isRunningState(Bass.ChannelIsActive(activeStream));
-
             // because device validity check isn't done frequently, when switching to "No sound" device,
             // there will be a brief time where this track will be stopped, before we resume it manually (see comments in UpdateDevice(int).)
             // this makes us appear to be playing, even if we may not be.
-            isRunning = running || (isPlayed && !hasCompleted);
+            isRunning = isPlayed && !hasCompleted;
```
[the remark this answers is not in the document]
nekodex: Which part? The removal of the `running` check? IIRC that's because a decode stream is always considered active/"running", so that check was longer valid.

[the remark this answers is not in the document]
smoogipoo: Oh, interesting. Is that not the case for samples? `SampleChannelBass` makes even heavier use of `ChannelIsActive` even though it's listed as `BassFlags.SampleChannelStream | BassFlags.Decode`.

osu.Framework/Audio/Sample/SampleChannelBass.cs
nekodex: I may be mis-remembering here, but I believe references were never being released when pause was used instead of stop, which resulted in leaks (within `AudioMixer`)...

Just the whole tracking of playback channels within `AudioMixer` is... not great.

osu.Framework/Audio/AudioMixer.cs
nekodex: I'm thinking we probably want this to either match whatever the playback device is using, or alternatively have it be configurable (with the latter probably being a better choice if we're heading towards WASAPI exclusive mode)?

osu.Framework/Audio/AudioMixer.cs
nekodex: I agree that throwing around `channelHandles` isn't great, but I couldn't think of a better way to handle this. The intention was that these methods only be used internally by o!f, though.

osu.Framework/Audio/AudioMixer.cs
nekodex: Yeah... I think beyond possibly always having a limiter at the end of the chain (to avoid clipping), we probably want to remove the rest in favour of exposing a method to allow users to define their own processing/effect chain(s).

osu.Framework/Audio/AudioMixer.cs
nekodex: This entire method is pretty hacky and probably should be rewritten TBH.

osu.Framework/Audio/AudioMixer.cs
nekodex: It's kinda confusing, but the play/pause flag applies to the _mixer_ channel, not the sample/track channel(s)... or put another way, `BassFlags.MixerChanPause` is a flag for `BassMix.MixerAddChannel()` and not `Bass.SampleGetChannel()`

Having said that, looking at `SampleChannelBass.cs` shows sample channels being added without the pause flag:
https://github.com/ppy/osu-framework/blob/6e4a6d5df667515b92ea1321d199efaf5a392da1/osu.Framework/Audio/Sample/SampleChannelBass.cs#L197

...but in a quick test they don't seem to auto-playback, so possibly samples don't auto-play? 🤷

smoogipoo: In terms of audible playback, is my initial understanding correct?

smoogipoo:
@peppy I don't need an immediate response but before I start architecting this further, do you have any qualms/questions/issues/thoughts/or confirm that the above structure I posted is something you imagine as the intended API?

peppy:
I'd go with the second option to start with, and maybe try to derive `AudioContainer` for the mixer container (or not, if that presents issues).

osu.Framework/Audio/AudioMixer.cs
nekodex: Uh yeah, that should be
```c#
BassMix.ChannelFlags(channelHandle, BassFlags.MixerChanPause, BassFlags.MixerChanPause);
```

nekodex:
> That is the intention. The idea is that you'd have an `AudioMixer` used for gameplay, to be able to apply effects to only subcomponents. ie. if you want a low-pass filter that is only applied to the audio track but not gameplay samples, or vice-versa.
> 
> I _believe_ @nekodex was already making use of this to apply limiters to only sound effects?

Yup, the goal was for mixers to eventually be hierarchical, but as this was a proof-of-concept, I limited it to a single top-level AudioMixer for the initial proposal... I did have working nested mixers to only apply effects to samples, but I left that out of the PR as it was already pretty big.

Also re: API, the first option feels kinda clunky to me (with manually having to add channels) compared to just using the draw hierarchy, for whatever my opinion is worth 😛

smoogipoo:
I think this should be ready to go. I've updated the description with the new structures and usage patterns enabled by my changes, which should be added to a wiki article somewhere.

@nekodex Sorry for completely hijacking this branch >.> See if you like the structure I've cooked up!

@nekodex I've removed the overlay for the time being. My reasoning is that this PR is quite large and for the overlay to work nicely it needs to name things (samples, tracks) as well as have some connection to the draw hierarchy. For the latter point, when you toggle the inspector in the draw visualiser (Ctrl+F1), there's two tabs there, and I imagine a third tab for when a `DrawableAudioMixer` is inspected.

In general, this still needs testing on other OSs. I've discovered ~100ms latency as a result of this, which seems to be 98% incurred by the `PlaybackBufferLength` of 100 defined in `AudioManager`. I don't know if this latency is the same on other OSs, because I've applied pulseaudio hax to reduce my latency globally that may not be taking effect here.

smoogipoo:
Test failures are related.

frenzibyte:
I can confirm I also get ~100ms latency on macOS.

smoogipoo:
We may be able to get away with the following:
```diff
diff --git a/osu.Framework/Audio/AudioManager.cs b/osu.Framework/Audio/AudioManager.cs
index 3417f5067c..7e4577961c 100644
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -323,10 +323,15 @@ protected virtual bool InitBass(int device)
                 didInit = Bass.Init(device);
             }
 
-            if (didInit)
-                thread.RegisterInitialisedDevice(device);
+            if (!didInit)
+                return false;
+
+            thread.RegisterInitialisedDevice(device);
 
-            return didInit;
+            Bass.GetInfo(out var info);
+            Bass.Configure(ManagedBass.Configuration.PlaybackBufferLength, info.MinBufferLength);
+
+            return true;
         }
 
         private void syncAudioDevices()
```
But I really want these settings to be adjustable before or shortly after doing so, as although it's recommended to use that value the BASS docs state that it's "still possible that the sound could break up on some systems".

osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
```diff
@@ -0,0 +1,293 @@
...
+            if (!(channel is IBassAudioChannel bassChannel))
+                return;
+
+            Debug.Assert(!mixedChannels.Contains(bassChannel));
+            mixedChannels.Add(bassChannel);
+
+            if (Handle == 0 || bassChannel.Handle == 0)
+                return;
+
+            ((IBassAudioMixer)this).RegisterHandle(bassChannel);
```
bdach, line 59: Pretty confused by this call and lack of its symmetry in `AddInternal()`. So a channel can be registered multiple times (when switching mixers, for instance), but it'll only be unregistered if a mixer it's in gets disposed? (`RemoveInternal()` just does `MixerRemoveChannel()` manually.)

But then _also_ unregistering calls `Remove()` which in turn calls `RemoveInternal()`...? Pretty convoluted...

osu.Framework/Audio/Sample/SampleBassFactory.cs
```diff
@@ -59,13 +63,6 @@ internal override void UpdateDevice(int deviceIndex)
             // The sample may not have already loaded if a device wasn't present in a previous load attempt.
             if (!IsLoaded)
                 loadSample();
-
-            if (!IsLoaded)
-                return;
-
-            // counter-intuitively, this is the correct API to use to migrate a sample to a new device.
-            Bass.ChannelSetDevice(SampleId, deviceIndex);
-            BassUtils.CheckFaulted(true);
```
bdach, line 68: Is this no longer an issue with the mixer present or something? Looks like a thing that Did Something™ and I'm not immediately sure if it's safe to remove.

osu.Framework/Graphics/Audio/DrawableSample.cs
```diff
@@ -26,12 +30,40 @@ public DrawableSample(ISample sample, bool disposeSampleOnDisposal = true)
...
+            mixer?.Add(channel);
+
+            return channel;
+        }
 
         public double Length => sample.Length;
 
         public Bindable<int> PlaybackConcurrency { get; } = new Bindable<int>(Sample.DEFAULT_CONCURRENCY);
+
+        private IAudioMixer mixer;
```
bdach, line 54: Should probably be marked nullable like other places do.

bdach (changes requested):

osu.Framework/Audio/Mixing/AudioMixer.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// </summary>
+    public abstract class AudioMixer : AdjustableAudioComponent, IAudioMixer
+    {
+        private readonly AudioMixer? defaultMixer;
+
+        /// <summary>
+        /// Creates a new <see cref="AudioMixer"/>.
+        /// </summary>
+        /// <param name="defaultMixer">The default <see cref="AudioMixer"/>, which <see cref="IAudioChannel"/>s will be moved to if removed from this one.
+        /// A <c>null</c> value indicates the default <see cref="AudioMixer"/>.</param>
```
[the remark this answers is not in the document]
smoogipoo: I'll rename this to `globalMixer`?

The idea is that removed audio channels are moved into the global mixer, and that a null value indicates that this audio mixer is the global one (i.e. channels can't be removed from it).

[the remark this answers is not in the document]
bdach: Yeah, I think that works.

osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
```diff
@@ -0,0 +1,293 @@
...
+        internal class EffectWithHandle
+        {
+            public int Handle { get; set; }
+            public int Priority { get; set; }
+
+            public readonly IEffectParameter Effect;
+
+            public EffectWithHandle(IEffectParameter? effect)
+            {
+                Effect = effect ?? throw new ArgumentNullException(nameof(effect));
```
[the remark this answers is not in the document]
smoogipoo: I agree. Did it because of this inspection, but didn't intend to keep it around like this:
image

osu.Framework/Audio/Sample/SampleBassFactory.cs
smoogipoo: I'll double check, but all of these channels are now just decoding streams. They don't play audio and don't really need a device. It's only the mixer which needs to be moved between devices.

Changing devices actually failed with this left as-is, so I'll double check.

osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
smoogipoo: Check if you understand the recent changes better, because I moved away from this whole Registration/Unregistration nonsense.

smoogipoo:
@frenzibyte Check if you can still repro audio breaking. As I mentioned in Discord, it's very likely that relying on weak references to remove channels resulted in the issue, and so I've removed weak references altogether.

frenzibyte:
I still can't reproduce the breakups anymore, but one thing I noticed before is that the BASS CPU% usage increases gradually as more and more samples play and finish and doesn't fall back to 1%~2% at all (assuming that's what eventually causes the breakups), while now with the weak list changes it doesn't reach beyond 2%.

peppy:
> We may be able to get away with the following:
> 
> ```diff
> diff --git a/osu.Framework/Audio/AudioManager.cs b/osu.Framework/Audio/AudioManager.cs
> index 3417f5067c..7e4577961c 100644
> --- a/osu.Framework/Audio/AudioManager.cs
> +++ b/osu.Framework/Audio/AudioManager.cs
> @@ -323,10 +323,15 @@ protected virtual bool InitBass(int device)
>                  didInit = Bass.Init(device);
>              }
>  
> -            if (didInit)
> -                thread.RegisterInitialisedDevice(device);
> +            if (!didInit)
> +                return false;
> +
> +            thread.RegisterInitialisedDevice(device);
>  
> -            return didInit;
> +            Bass.GetInfo(out var info);
> +            Bass.Configure(ManagedBass.Configuration.PlaybackBufferLength, info.MinBufferLength);
> +
> +            return true;
>          }
>  
>          private void syncAudioDevices()
> ```
> 
> But I really want these settings to be adjustable before or shortly after doing so, as although it's recommended to use that value the BASS docs state that it's "still possible that the sound could break up on some systems".

I'd definitely want to see this in the initial PR. And also recommend we initially *don't* add a setting (at least not in the UI) to better gauge how this default works for people. Rationale being I'd hope in 99% of cases this would not need to be adjusted, and should most definitely not cause (underrun) issues.

osu.Framework.Tests/Audio/BassAudioPipeline.cs
```diff
@@ -0,0 +1,101 @@
...
+using osu.Framework.Development;
+using osu.Framework.IO.Stores;
+using osu.Framework.Threading;
+
+namespace osu.Framework.Tests.Audio
+{
+    /// <summary>
+    /// Provides a BASS audio pipeline to be used for testing audio components.
+    /// </summary>
+    public class BassAudioPipeline : IDisposable
```
peppy, line 21: Found the naming of this class a bit weird (at first it seemed like "Pipeline" was referring to some bass-side api term, but seems not).

osu.Framework/Audio/BassUtils.cs
```diff
@@ -27,5 +27,19 @@ internal static bool CheckFaulted(bool throwException)
...
+            if (gain == 0) return 0;
+
+            return Math.Pow(10, gain / 20);
+        }
+
+        internal static double LevelToDb(double level)
+        {
+            if (level == 0) return 0;
+
+            return Math.Log(Math.Sqrt(level), 10) * 20f;
```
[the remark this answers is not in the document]
nekodex: Wow... uh yeah, nice catch. No clue where that square root came from... should just be:
```c#
return Math.Log(level, 10) * 20f;
```

osu.Framework.Tests/Audio/BassAudioPipeline.cs
smoogipoo: Any suggestions? I struggled with this one.

peppy (changes requested):
Just some initial thoughts, still working through this.

osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
```diff
@@ -0,0 +1,312 @@
...
+    /// Mixes together multiple <see cref="IAudioChannel"/> into one output via BASSmix.
+    /// </summary>
+    internal class BassAudioMixer : AudioMixer, IBassAudioChannelInterface, IBassAudio
+    {
+        /// <summary>
+        /// The handle for this mixer.
+        /// </summary>
+        public int Handle { get; private set; }
+
+        internal readonly List<EffectWithHandle> MixedEffects = new List<EffectWithHandle>();
```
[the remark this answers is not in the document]
smoogipoo: `Effects` conflicts with the base class' `Effects`, is mostly the reason this is named as it is. It's also only `internal` for testing purposes.

osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
```diff
@@ -0,0 +1,312 @@
...
+        private readonly List<IBassAudioChannel> mixedChannels = new List<IBassAudioChannel>();
```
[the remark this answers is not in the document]
smoogipoo: These fields as channels/effects can be thought of as having been applied to the native mixer, or are active.

I think I'll rename to `activeEffects` / `activeChannels`?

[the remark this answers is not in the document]
peppy: Right, because the base names are already taken. So this is basically separating the channels/effects which have actually been applied (on the audio thread) vs the ones the user has queued potentially from a different thread?

osu.Framework.Tests/Audio/BassAudioPipeline.cs
peppy: i.. don't have any ideas. `BassTestComponents`..?

leaving it as is may also be fine, but i'd rename all the local variables/fields to reference it as `bass` rather than `pipeline`, i think..

osu.Framework.Tests/Audio/TrackBassTest.cs
```diff
@@ -333,25 +322,16 @@ public void TestZeroFrequencyHandling()
...
 
             // assert time is frozen after 50ms sleep and didn't change with full precision, but "IsRunning" is still true.
-            Thread.Sleep(50);
-            updateTrack();
-
-            Assert.IsTrue(track.IsRunning);
-            Assert.AreEqual(currentTime, track.CurrentTime);
-
-            // set back to one and update track.
-            track.Frequency.Value = 1;
```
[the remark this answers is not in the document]
smoogipoo: This was done before I took over the PR. I've reverted the changes and applied my own to fix the test failures. Quite interesting differences in timing between master and this branch - restart doesn't take effect immediately which may be a BASS issue.
I'm running tests hundreds of times locally and also via CI after my changes.

smoogipoo:
The only thing I haven't fixed is `mixer.Add(sample.Play())` playing on the wrong mixer for a bit. It only affects `Sample` and not `DrawableSample`, so I don't think it's too bad for the time being. It matches the brokenness of setting properties after `.Play()` at least, and the path forward is likely to do something like `.PlayWith(c => mixer.Add(c))` or something like that.

peppy:
Is that code actually used anywhere though, apart from the example you give? If not, can't the example just be updated to add the channel to the mixer before calling `Play`?

smoogipoo:
I've updated the example.

osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
bdach: Yeah it looks better now. Some asymmetry still remains and threw me in a loop (eg. `AddChannelToBassMix` adding things to `activeChannels`, but `removeChannelFromBassMix` not removing from it), but I suppose I can understand the reasoning for that based on the usage sites.

osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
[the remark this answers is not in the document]
bdach: I see that the usage site depicted on the above screenshot has an `.AsNonNull()` now, but this throw still remains. I'll remove it.

osu.Framework/Audio/Sample/SampleChannelBass.cs
```diff
@@ -101,7 +100,7 @@ protected override void UpdateState()
         {
             if (hasChannel)
             {
-                switch (Bass.ChannelIsActive(channel))
+                switch (bassMixer.AsNonNull().ChannelIsActive(this))
```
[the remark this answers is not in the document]
smoogipoo: It can be null in two cases - in the constructor (e.g. if a `TrackBass` is instantiated on the audio thread), and when the component is removed from the mixer for the final time in dispose. This is the reason why my previous changes had an `PassThroughAudioChannelInterface`, which maintained never-null.

We can cast as `=> (BassAudioMixer)Mixer.AsNonNull()` instead, however some areas like `TrackBass.Mixer_set()` will still need to consider the null case and will not be able to use that method as a result. It also hides such errors in the first place.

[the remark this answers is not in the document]
bdach: Hmm, I see. The previous variant *was* nicer but I presume the reason it's gone is fallout from the removal of registration.

Maybe fine to leave in this case - although not sure about the "hiding errors" reasoning against the straight cast, as nullrefs will only ever manifest at runtime either way if any of these accesses happens at a wrong time.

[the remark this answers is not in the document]
smoogipoo: > I presume the reason it's gone is fallout from the removal of registration.

Rather, it was because it seemed confusing for others to understand the purpose of it existing/the "Interface" class.

I've inverted the `.AsNonNull()` usages, see if it's more amicable now :)

bdach:
Code-wise I don't have any issues with this anymore. Test failure on latest CI run looks related, though - even if it's to be fixed by tuning up timeouts or something.

peppy:
I'm still doing a final pass of this code-wide, but wanted to make a checklist of potential concerns/non-concerns regarding libraries and platform testing. Anyone is welcome to check off items and elaborate further, just thoughts which arose along the way:

- [ ] Testing on android
- [ ] Testing on iOS
- [ ] Whether we still need to use the beta bass version
- [ ] Whether we still need to use the ppy fork of `ManagedBass`
- [ ] Whether we can release a non-alpha version of `NativeLibs` before merge of this

smoogipoo:
Would appreciate re-testing with the latest commit, since I've updated the native BASS libraries (2.4.16.3). Note that you'll need a local nativelibs package to test.

bdach:
I briefly checked out how android looks at 2e8b7dd. Framework test project launches, the framework mixer test scene with the low pass filter looks to work as expected, and a game checkout (with the game-side changes) also launches and seems to play tracks and samples just fine.

I have not tested more complex bassmix usages or absolutely every single bass call game-side, though, so buyer beware.

peppy:
Submitted PR for upstream change https://github.com/ManagedBass/ManagedBass/pull/97. Have also deployed the slightly modified changes to nuget, have updated this branch to point at that for now.

In attempting to test iOS I ran into an AOT issue (https://github.com/xamarin/xamarin-macios/issues/12353), so I think our best bet is just to move forward presuming we can somehow get it working when we decide to spend a week or month on iOS.

Currently trying to deploy `NativeLibs` from this branch ahead of time so we can consume it here.

peppy (approved):

PR #4293, opened 2021-03-20
Avoid changing the enabled state of MidiHandler from the input thread

peppy:
There is a general expectation that the enabled state of input handlers is only changed from the update thread. `MidiHandler` was an exception to this rule.

Closes ppy/osu#12102.


bdach:
Seems fair. I also believe it is more correct anyway given that input handler enable/disable is exposed from the UI, since the user could try to re-enable the handler after it failing, while the intention of the code was for it to bail out and never try again until the next startup.

bdach (approved):

smoogipoo (approved):

PR #4303, opened 2021-03-22
Avoid using DI for audio component adjustment lookup to better support pooling

peppy:
This is a second attempt at https://github.com/ppy/osu-framework/pull/4212, minus the breakage.

The only part of this I'm not 100% sure about is the logic added to ensure no self-references. An alternative to this is to fix each occurrence that uses interface delegation (see https://github.com/ppy/osu/compare/master...peppy:audio-interafce-delegation-fixes?expand=1). In theory this looks like a better approach, until you realise that to make it works, `PoolableSkinnableSample` needs to have similar code to this framework-side change in order to get the parent adjustments (since it is pretending to be a wrapper without actually being one).

Tested to work well osu! side. Not sure if it's worth adding test coverage for this.

### DIFF
--- a/osu.Framework/Graphics/Audio/DrawableAudioWrapper.cs
+++ b/osu.Framework/Graphics/Audio/DrawableAudioWrapper.cs
@@ -48,6 +48,10 @@ namespace osu.Framework.Graphics.Audio
 
         private readonly AudioAdjustments adjustments = new AudioAdjustments();
 
+        private IAggregateAudioAdjustment parentAdjustment;
+
+        private readonly LayoutValue parentAdjustmentLayout = new LayoutValue(Invalidation.Parent);
+
         /// <summary>
         /// Creates a <see cref="DrawableAudioWrapper"/> that will contain a drawable child.
         /// Generally used to add adjustments to a hierarchy without adding an audio component.
@@ -56,6 +60,7 @@ namespace osu.Framework.Graphics.Audio
         protected DrawableAudioWrapper(Drawable content)
         {
             AddInternal(content);
+            AddLayout(parentAdjustmentLayout);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Audio/DrawableAudioWrapper.cs
```diff
@@ -69,11 +71,36 @@ protected DrawableAudioWrapper([NotNull] IAdjustableAudioComponent component, bo
...
-        [BackgroundDependencyLoader(true)]
-        private void load(IAggregateAudioAdjustment parentAdjustment)
+        private IAggregateAudioAdjustment parentAdjustment;
+
+        protected override bool OnInvalidate(Invalidation invalidation, InvalidationSource source)
         {
-            if (parentAdjustment != null)
-                adjustments.BindAdjustments(parentAdjustment);
+            // because these components may be pooled, relying on DI is not feasible.
+            // in the majority of cases the traversal should be quite short. may require later attention if a use case comes up which this is not true for.
```
smoogipoo, line 79: Sure, there's just two things that concern me.

1. `Drawable` receives a `Parent` invalidation basically immediately as its added to the parenting composite, for better or worse.
2. `CompositeDrawable` sets a parent after the child's parent is set (`loadChild(c1) -> loadChild(c2) -> loadChild(c3) -> c3.SetParent() -> c2.SetParent() -> c1.SetParent()`).

Since `Parent` is one of the special invalidation types that always propagates to children, it means a simple load where no parents change causes 3 lookups here which can get pretty rough. There's no xmldoc for this (https://github.com/ppy/osu-framework/pull/4304).

As for fixing, I think this needs a debounce (`Update` + backed by `LayoutValue`). However before going down that path, what do you think about just referencing `Parent.Dependencies` directly?

smoogipoo (changes requested):

osu.Framework/Graphics/Audio/DrawableAudioWrapper.cs
peppy: i'm fine with trying `Parent.Dependencies`. At worst it will give us reason to not do it in other scenarios if we run into failure cases.

peppy: As discussed in discord, using `Parent.Dependencies` doesn't work as expected due to the feedback-loop avoidance part, where we potentially skip a portion of traversal. I've gone with the debounce-via-`LayoutValue` part though.

osu.Framework/Graphics/Audio/DrawableAudioWrapper.cs
```diff
@@ -69,11 +75,40 @@ protected DrawableAudioWrapper([NotNull] IAdjustableAudioComponent component, bo
...
         }
 
-        [BackgroundDependencyLoader(true)]
-        private void load(IAggregateAudioAdjustment parentAdjustment)
+        protected override void Update()
+        {
+            base.Update();
+
+            if (!parentAdjustmentLayout.IsValid)
+                refreshAdjustments();
```
smoogipoo, line 83: needs a `parentAdjustmentLayout.Validate()`

smoogipoo (changes requested):

osu.Framework/Graphics/Audio/DrawableAudioWrapper.cs
peppy: Added and re-tested with osu! to make sure it still works.

smoogipoo (approved):
Let's see how this goes. I'm still concerned about performance, especially as it's going to manifest as a single-time event rather than a global reduction in frametime, but I don't have a better solution immediately.

PR #4304, opened 2021-03-22
Add xmldoc stating parent invalidations always propagate

smoogipoo:
Easy to miss.

### DIFF
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -2751,6 +2751,7 @@ namespace osu.Framework.Graphics
 
         /// <summary>
         /// A <see cref="Drawable.Parent"/> has changed.
+        /// This <see cref="Invalidation"/> flag always propagates to children.
         /// </summary>
         Parent = 1 << 6,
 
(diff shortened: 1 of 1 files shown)


peppy:
What does "always" mean in this context? Might be good to elaborate a bit (i earnestly don't know). Is it "always" as in send in every `Invalidation`, or "always" even for non-alive children or something else? Hard to tell without investigating further, I think.

smoogipoo:
Invalidations in general only apply to alive children, I just made the xmldoc be in inverse of `Invalidation.DrawNode`. Should I list it as `always propagates to alive children`?

osu.Framework/Graphics/Drawable.cs
```diff
@@ -2751,6 +2751,7 @@ public enum Invalidation
 
         /// <summary>
         /// A <see cref="Drawable.Parent"/> has changed.
+        /// This <see cref="Invalidation"/> flag always propagates to children.
```
peppy, line 2754:
```suggestion
        /// Unlike other <see cref="Invalidation"/> flag, this propagates to all children regardless of their <see cref="IsAlive"/> state.
```

might read better

peppy (approved):

Issue #4307, opened 2021-03-23
BindableSize does not propagate `MinValue`/`MaxValue` changes to bindings


frenzibyte:
I've attempted in addressing that until I realized that I don't really need it at the moment. I've done it by abstracting the `MinValue` and `MaxValue` propagation to a base `ConstraintedBindable` class, to avoid a bad amount of duplicated code.

Here's my changes: https://github.com/ppy/osu-framework/compare/master..frenzibyte:constrained-bindable

Can open a PR for it but would like to hear whether it's in the right direction, as I don't really want to open yet another bindable PR here to stay dead for a couple of months.

bdach:
What is "min/max value" even supposed to be defined as for size? Component-wise min/max? Seems arbitrary. You seem to have an even weirder implementation there (does width then height first for some reason?)

What do you need this for?

peppy:
FWIW I hit this same limitation before switching to use `Vector2` instead of `Size`. The `BindableSize` class is about halfway where it should be in terms of implementation, but I'm not sure we're using it in enough places to warrant the maintenance cost.

Also interested to know your use case for it.

frenzibyte:
Was gonna need it since I was initially gonna make use of the framework config's window size bindable for specifying the minimum and maximum window size value, but I've found adding the two properties at `IWindow` is a better choice.

At this point `BindableSize` can actually be removed altogether, it's only used now with https://github.com/ppy/osu-framework/pull/4308 to clamp value instead of attempt to set size beyond the specified range, but that's not a reason to keep it around for now.

bdach:
PR that and we can review it properly.

PR #4308, opened 2021-03-23
Add support for constraining game window size

### LINKED ISSUE #3501: Resizing window below minimum size is being corrected manually, window doesn't have a native minimum size
Video demonstration: https://www.youtube.com/watch?v=zEg1ePfI0Ns
This has, in some cases, crashed osu!lazer, but I unfortunately don't have a recording of that. A similar issue has occurred when running the sample game on gnome 3.34 which almost randomly resizes the window every few frames when you try to normally resize it.

frenzibyte:
Closes #3501 

I've changed how this works by adding two properties at `IWindow` for configuring the minimum and maximum size of the window, instead of specifying them at `FrameworkConfigManager` where it doesn't make much sense and would actually force itself to depend on https://github.com/ppy/osu-framework/issues/4307.

Few notes:

- The early guards on `MinSize` and `MaxSize` may have been kinda too much, but wasn't sure about allowing to set any value there.
- Have confirmed `SDL_SetWindowMinimumSize`/`SDL_SetWindowMaximumSize` values stay still regardless of switching between window modes.


osu.Framework.Tests/Visual/Platform/TestSceneWindowed.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        [Test]
+        public void TestMinimumSize()
+        {
+            AddStep("set window size 640x480", () => config.SetValue(FrameworkSetting.WindowedSize, new Size(640, 480)));
+            AddStep("set minimum size to 1024x768", () => sdlWindow.MinSize = new Size(1024, 768));
+            AddAssert("window size = 1024x768", () => config.Get<Size>(FrameworkSetting.WindowedSize) == new Size(1024, 768));
+
+            // todo: propagating min value changes with BindableSize not supported. (https://github.com/ppy/osu-framework/issues/4307)
+            //AddAssert("minimum size passed to config", () => ((BindableSize)config.GetBindable<Size>(FrameworkSetting.WindowedSize)).MinValue == new Size(1024, 768));
```
bdach, line 55: I'm not sure this should be a todo item at all ever. I think that min/max size is more of a constraint set by an application, than something that should be user-facing/user-modifiable. Therefore I'd agree with not persisting those values to config at all.

osu.Framework/Platform/IWindow.cs
```diff
@@ -166,6 +166,16 @@ public interface IWindow : IDisposable
...
 
+        /// <summary>
+        /// The minimum size of the window.
+        /// </summary>
+        Size MinSize { get; set; }
+
+        /// <summary>
+        /// The maximum size of the window.
+        /// </summary>
+        Size MaxSize { get; set; }
```
bdach, line 177: I'm not sure I wholly agree with putting this on `IWindow`, given that on a framework level it's describing a window on a platform-agnostic level, and these properties make little sense for mobile platforms. I'd just stick it on `SDL2DesktopWindow` and call it a day. [We already do casting to do desktop-specific stuff with the window game-side.](https://github.com/ppy/osu/blob/bb6e50eb27f8e8164e2c97b2a9b96bc62a45a4fd/osu.Desktop/OsuGameDesktop.cs#L139-L144)

This would also get rid of the misleading osuTK implementation that silently does nothing.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -145,6 +145,60 @@ private set
...
+                    return;
+
+                if (value.Width < 0 || value.Height < 0)
+                    throw new InvalidOperationException($"Expected positive size, got {value}.");
+
+                if (value.Width > maxSize.Width || value.Height > maxSize.Height)
+                    throw new InvalidOperationException($"Expected a size less than {nameof(MaxSize)} ({MaxSize}), got {value}");
+
+                minSize = value;
+                sizeWindowed.MinValue = value;
```
bdach, line 169: Would remove these sets, as per the other review comment.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -432,6 +486,9 @@ public virtual void Create()
 
             Exists = true;
 
+            MaxSize = new Size(int.MaxValue, int.MaxValue);
+            MinSize = new Size(640, 480);
```
bdach, line 490: I would either remove this new default, or add a breaking change note about it (to tell people to unset this to `(0,0)` if they want absolute control over the size). It potentially breaks projects that use windows in an unorthodox manner.

frenzibyte: Well, previously the window was supposed to be limited to `640x480` minimum size here in osu!framework, but was silently broken post SDL migration, so wasn't sure at first.

osu.Framework/Platform/IWindow.cs
frenzibyte: Yeah I was just not sure if casting to specific classes is the way here, I had it like you said initially.

bdach: The cleaner way to do this would be to adhere to interface segregation and have a `IDesktopWindow : IWindow` that extends the multiplatform capabilities with the specifics of desktop windows. But given that we have exactly one implementation right now that we're probably unlikely to switch out anytime soon, it's probably not worth the hassle.

osu.Framework.Tests/Visual/Platform/TestSceneWindowed.cs
frenzibyte: As a property in `Bindable` that is related in a way with `Value`, I would expect it to propagate across bindings. But either way, best to settle that in the linked issue thread, then remove this todo item here if the issue is deemed to be closed.

osu.Framework/Configuration/FrameworkConfigManager.cs
```diff
@@ -22,7 +22,7 @@ protected override void InitialiseDefaults()
         {
             SetDefault(FrameworkSetting.ShowLogOverlay, false);
 
-            SetDefault(FrameworkSetting.WindowedSize, new Size(1366, 768), new Size(640, 480));
```
peppy, line 25: Why was this specification removed? Is it an assumption that specifying the minimum size here no longer has any effect? If so, this needs a mention of it as a breaking change, as there may be consumers which are relying on this.

peppy (changes requested):

peppy:
Going to need to consider what `Min`/`MaxSize` mean - is this limiting the pre or post DPI scale application? I think it makes more sense to be limiting post application since you'd want the same amount of content to be on screen at the constraining minimum/maximum size.

osu.Framework/Configuration/FrameworkConfigManager.cs
frenzibyte: `SDL2DesktopWindow` already tries to propagate the `MinValue`/`MaxValue` to the config bindable, but won't work with the propagation issue.

And either way, SDL would resize the window back to the minimum size overwriting whatever has been set to `FrameworkSetting.WindowedSize`, so there should be no issues for now.

frenzibyte:
Fair point, I'll change it to be post DPI scale.

frenzibyte:
Hold up on this, visual tests got broken on that last commit. will fix in a bit.

peppy:
@frenzibyte Are we still waiting on a fix for this or is it good for review?

frenzibyte:
Visual tests got broken due to https://github.com/ppy/osu-framework/pull/4308/commits/7f1c93db57940a9b222f95b56b0da70f76a51fc0, will fix it and apply the above suggestions now.

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: I think we don't need a breaking change for this yeah? not much passed since the migration SDL.

frenzibyte:
Aight, good for review now.

osu.Framework/Configuration/FrameworkConfigManager.cs
```diff
@@ -22,7 +22,7 @@ protected override void InitialiseDefaults()
...
+            SetDefault(FrameworkSetting.WindowedSize, new Size(1366, 768));
```
Susko3, line 25: With this change, the following gets logged when starting visual-tests:

```
[runtime] 2021-04-16 19:03:01 [important]: Unable to parse config key WindowedSize: System.InvalidCastException: Invalid cast from 'System.String' to 'System.Drawing.Size'.
[runtime] 2021-04-16 19:03:01 [important]: at System.Convert.DefaultToType(IConvertible value, Type targetType, IFormatProvider provider)
[runtime] 2021-04-16 19:03:01 [important]: at System.String.System.IConvertible.ToType(Type type, IFormatProvider provider)
[runtime] 2021-04-16 19:03:01 [important]: at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
[runtime] 2021-04-16 19:03:01 [important]: at osu.Framework.Bindables.Bindable`1.Parse(Object input) in C:\...\osu-framework\osu.Framework\Bindables\Bindable.cs:line 260
[runtime] 2021-04-16 19:03:01 [important]: at osu.Framework.Configuration.IniConfigManager`1.PerformLoad() in C:\...\osu-framework\osu.Framework\Configuration\IniConfigManager.cs:line 66
```

In `framework.ini` file, the following gets written:
```
WindowedSize = {Width=1340, Height=796}
```

Without the change, it's:
```
WindowedSize = 1358x782
```

frenzibyte: basically using `Bindable<Size>` instead of `BindableSize`, will fix.

osu.Framework/Platform/SDL2DesktopWindow.cs
bdach: I haven't changed my mind on this honestly. Both this and the `WindowedSize` minimum removal/ineffectiveness need to be mentioned in some capacity.

bdach:
FWIW, to me this PR feels half-baked if `BindableSize.{Min,Max}Value` do not work properly and we're hacking around that with separate bindables for min/max. I didn't notice it before, but `BindTo()` already does component-wise clamping, so I'd be fine with making those values propagate in the same capacity to resolve https://github.com/ppy/osu-framework/issues/4307.

smoogipoo:
I'd definitely like to see `BindableSize` fixed first, since this isn't such an important change that needs to be pushed out immediately.

frenzibyte:
Have looked back into this, using `BindableSize` works but there's actually no proper way for a framework consumer to alter it, since `ConfigManager`s have no way of returning specific bindable implementations like `BindableSize`.

I've attempted at defining a `Game.GetFrameworkConfigRanges()` method in which the game can override the framework's setting ranges (to alter the window size range), but I feel that's giving the consumer a bit more access than it should have ([diff](https://github.com/ppy/osu-framework/commit/9ab77342761d125c770d3d58708b7fa5bb5272c3)).

Therefore I kept the `MinSize`/`MaxSize` properties as-is and bound them with the ranges specified in the config bindable. Would recommend making sure this works this as I couldn't properly test it on my machine ([SDL woes with M1](https://github.com/libsdl-org/SDL/issues/5203)), ~~but I was briefly able to test it using [a method I found](https://user-images.githubusercontent.com/22781491/165029445-bc5a47fd-cb67-4108-b23d-7510990283e8.mp4)...~~

osu.Framework/Platform/IWindow.cs
```diff
@@ -186,6 +186,18 @@ public interface IWindow : IDisposable
         /// </summary>
         Size ClientSize { get; }
 
+        /// <summary>
+        /// The minimum size of the window, before scaling.
```
peppy, line 190: Feels weird that this is the only xmldoc in this file that mentions "before scaling". Probably not required?

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -155,6 +155,58 @@ protected set
             }
         }
 
+        private Size minSize = Size.Empty;
```
peppy, line 158: Is this required, or can the bindable be used directly?

frenzibyte: Hmm, can replace with the bindable.

peppy:
Everything looked to be behaving fine, but I did get a crash when manually testing this on macOS:

```csharp

[runtime] 2022-04-25 09:42:42 [error]: An unhandled error has occurred.
[runtime] 2022-04-25 09:42:42 [error]: System.ArgumentException: '640' cannot be greater than 639.
[runtime] 2022-04-25 09:42:42 [error]: at System.Math.ThrowMinMaxException[T](T min, T max)
[runtime] 2022-04-25 09:42:42 [error]: at osu.Framework.Bindables.BindableSize.ClampValue(Size value, Size minValue, Size maxValue) in /Users/dean/Projects/osu-framework/osu.Framework/Bindables/BindableSize.cs:line 47
[runtime] 2022-04-25 09:42:42 [error]: at osu.Framework.Bindables.RangeConstrainedBindable`1.setValue(T value) in /Users/dean/Projects/osu-framework/osu.Framework/Bindables/RangeConstrainedBindable.cs:line 201
[runtime] 2022-04-25 09:42:42 [error]: at osu.Framework.Bindables.RangeConstrainedBindable`1.set_Value(T value) in /Users/dean/Projects/osu-framework/osu.Framework/Bindables/RangeConstrainedBindable.cs:line 46
[runtime] 2022-04-25 09:42:42 [error]: at osu.Framework.Platform.SDL2DesktopWindow.storeWindowSizeToConfig() in /Users/dean/Projects/osu-framework/osu.Framework/Platform/SDL2DesktopWindow.cs:line 1200
[runtime] 2022-04-25 09:42:42 [error]: at osu.Framework.Threading.ScheduledDelegate.InvokeTask() in /Users/dean/Projects/osu-framework/osu.Framework/Threading/ScheduledDelegate.cs:line 103
[runtime] 2022-04-25 09:42:42 [error]: at osu.Framework.Threading.ScheduledDelegate.RunTaskInternal() in /Users/dean/Projects/osu-framework/osu.Framework/Threading/ScheduledDelegate.cs:line 92
[runtime] 2022-04-25 09:42:42 [error]: at osu.Framework.Threading.Scheduler.Update() in /Users/dean/Projects/osu-framework/osu.Framework/Threading/Scheduler.cs:line 115
[runtime] 2022-04-25 09:42:42 [error]: at osu.Framework.Platform.SDL2DesktopWindow.Run() in /Users/dean/Projects/osu-framework/osu.Framework/Platform/SDL2DesktopWindow.cs:line 531
[runtime] 2022-04-25 09:42:42 [error]: at osu.Framework.Platform.GameHost.Run(Game game) in /Users/dean/Projects/osu-framework/osu.Framework/Platform/GameHost.cs:line 769
[runtime] 2022-04-25 09:42:42 [error]: at osu.Framework.Tests.Program.Main(String[] args) in /Users/dean/Projects/osu-framework/osu.Framework.Tests/Program.cs:line 23
```

frenzibyte:
Ah, was due to an early-guard in the range properties got removed while it's actually required...

osu.Framework/Platform/IWindow.cs
```diff
@@ -186,6 +186,18 @@ public interface IWindow : IDisposable
...
+        /// The minimum size of the window.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Thrown when setting a negative size, or size greater than <see cref="MaxSize"/>.</exception>
+        Size MinSize { get; set; }
+
+        /// <summary>
+        /// The maximum size of the window.
+        /// </summary>
+        /// <exception cref="InvalidOperationException">Thrown when setting a negative or zero size, or size less than or equal to <see cref="MinSize"/>.</exception>
+        Size MaxSize { get; set; }
```
bdach, line 199: i'll just note that as written, you can't unapply a maximum window size after you've set one before. as in, in the ideal world this would be a `Size?` and you'd set `null` to reset the maximum size limit. for `MinSize` this is less of an issue because there you can just set zero size.

but this is actually an [SDL shortcoming as well](https://github.com/libsdl-org/SDL/blob/d9009388c8924b15f74ec38a26cc62ea42b40110/src/video/SDL_video.c#L2316-L2327) so i'm going to let this slide. maybe worth xmldoc'ing at least.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -445,6 +485,11 @@ public virtual void Create()
 
             updateWindowSpecifics();
             updateWindowSize();
+
+            sizeWindowed.MinValueChanged += min => MinSize = min;
+            sizeWindowed.MaxValueChanged += max => MaxSize = max;
```
bdach, line 490: this is really baffling. `sizeWindow.{Min,Max}Value` are already used as backing storage for `{Min,Max}Size`. so what this is doing is basically setting `{Min,Max}Size` *to the same value it already is*, just to issue the side effects of the setter, namely the `SDL_SetWindow{Min,Max}Size` call. would it not be simpler to move _those_ SDL calls to the min/max value changed callbacks?

osu.Framework.Tests/Visual/Platform/TestSceneWindowed.cs
```diff
@@ -0,0 +1,96 @@
...
+        [Resolved]
+        private FrameworkConfigManager config { get; set; }
+
+        private SDL2DesktopWindow sdlWindow;
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            if (!(host.Window is SDL2DesktopWindow window))
+                return;
```
bdach, line 28: this soft cast is arguably universally worse than a hard cast that was there in `TestSceneResizableWindow`, because if the hard cast fails, it will throw immediately at the point of cast, rather than throw much later inside the tests themselves because `sdlWindow` hasn't been initialised to anything in BDL due to being of the wrong type

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: The problem is how to let consumers access these properties, that’s why all this song and dance of `MinSize` and `sizeWindowed.MinValue` backing it is for. See https://github.com/ppy/osu-framework/pull/4308#issuecomment-1108113447.

osu.Framework.Tests/Visual/Platform/TestSceneWindowed.cs
frenzibyte: Actually, there may be no reason to have this anymore, and I can agree it’s worse. Well, old code from this PR.

osu.Framework/Platform/IWindow.cs
frenzibyte: Seems supportable though, we can make null fall back to int.MaxValue in implementation.

osu.Framework/Platform/SDL2DesktopWindow.cs
bdach: i don't follow. what is the "problem" with accessing the properties? do you mean that you want to run the bound validation logic too or something?

there's no reason why the validation couldn't be run in the value changed callbacks too, is there? you'd have a pure delegated setter, and all side effects in the callback.

frenzibyte: It’s about how the users can possibly access the windowed size bindable, they would have to retrieve the bindable using `GetBindable` and cast it to `BindableSize` to access said properties.

bdach: sorry but i don't get it still. feels like i'm receiving half-answers that i need to connect the dots myself on and i can't.

frenzibyte: Put it this way: If we only provide these ranges via the windowed size config bindable (`BindableSize.MinValue`/`BindableSize.MaxValue`), then the only way for consumers to modify these is by querying the bindable and casting it to `BindableSize`, since `ConfigManager`s always return `Bindable<T>`.

In short, this is how a framework consumer can modify the range:
```diff
         [BackgroundDependencyLoader]
-        private void load()
+        private void load(FrameworkConfigManager config)
         {
+            var windowedSize = (BindableSize)config.GetBindable<Size>(FrameworkSetting.WindowedSize);
+            windowedSize.MinValue = new Size(200, 400);
+            windowedSize.MaxValue = new Size(600, 800);
```

While with exposing this at `IWindow`, the consumers can modify the range via:
```diff
+        public override void SetHost(GameHost host)
+        {
+            base.SetHost(host);
+            host.Window.MinSize = new Size(200, 400);
+            host.Window.MaxSize = new Size(600, 800);
+        }
```

The main reason that made me not continue with the config bindable way is for the fact that the consumers would have to directly cast it to `BindableSize`, unless you think that's fine.

bdach: i'm either not communicating clearly enough or i don't know what else. i'm fine with the properties existing, i'm less fine with the logic flow.

so let me try to counter with a question: why can't anything that is currently in `MinSize.set` other than the line:

```csharp
sizeWindowed.MinValue = value;
```

be moved to the `sizeWindowed.MinValueChanged` callback?

frenzibyte: Ah, so you meant moving everything in the callback and keeping the `IWindow` properties to redirect to the bindable properties. Thought you meant moving everything in the callback and removing the properties as well. Fair enough.

osu.Framework/Platform/IWindow.cs
frenzibyte: Now that I look at it, not entirely sure how to support nullable if the bindable properties are non-nullable themselves. Just gonna add xmldic to use maximum value to remove constraint.

osu.Framework/Platform/IWindow.cs
```diff
@@ -186,6 +186,24 @@ public interface IWindow : IDisposable
...
+        /// <exception cref="InvalidOperationException">Thrown when setting a negative size, or size greater than <see cref="MaxSize"/>.</exception>
+        Size MinSize { get; set; }
+
+        /// <summary>
+        /// The maximum size of the window.
+        /// </summary>
+        /// <remarks>
+        /// To remove this constraint, set a <see cref="Size"/> of maximum integer value: <c>new Size(int.MaxValue, int.MaxValue)</c>.
+        /// </remarks>
+        /// <exception cref="InvalidOperationException">Thrown when setting a negative or zero size, or size less than or equal to <see cref="MinSize"/>.</exception>
```
smoogipoo, line 204:
```suggestion
        /// <exception cref="InvalidOperationException">Thrown when setting a negative or zero size, or a size less than <see cref="MinSize"/>.</exception>
```
It's just less than, right?

osu.Framework/Platform/IWindow.cs
```diff
@@ -186,6 +186,24 @@ public interface IWindow : IDisposable
         /// </summary>
         Size ClientSize { get; }
 
+        /// <summary>
+        /// The minimum size of the window.
+        /// </summary>
+        /// <remarks>
+        /// To remove this constraint, set an empty <see cref="Size"/>: <c>Size.Empty</c>.
+        /// </remarks>
+        /// <exception cref="InvalidOperationException">Thrown when setting a negative size, or size greater than <see cref="MaxSize"/>.</exception>
```
smoogipoo, line 195:
```suggestion
        /// <exception cref="InvalidOperationException">Thrown when setting a negative size, or a size greater than <see cref="MaxSize"/>.</exception>
```

smoogipoo (changes requested):

osu.Framework/Platform/IWindow.cs
frenzibyte: Well, it can be less than or equal to, at which point it makes the window basically unresizable. Though that defeats the purpose of `Resizable`, do you think it should just throw?

smoogipoo: Imo it doesn't need to throw, and the code currently doesn't throw if it's equal to.

frenzibyte: Ah, nevermind, you're right.

smoogipoo (approved):

peppy (changes requested):
Needs attention.

bdach:
I think the last commit is fine? The min/max sizes should arguably be restored to sane defaults after every test anyway to ensure repeatable execution. Could do it via `[TearDownSteps]` even but the way pushed here works too so I'm not going to nitpick.

If anything I'd say the issue here is that `Set{Min,Max}Value` in `RangeConstrainedBindable` are not checking whether the range is valid in any capacity themselves.

frenzibyte:
Can imagine that being intentionally not done to avoid requiring a specific execution order when changing the ranges (i.e. changing `MinValue` to a value higher than `MaxValue`, only to update `MaxValue` afterwards).

bdach:
Yeah I did think about that. Maybe could have a `SetRange(T min, T max)` value on `RangeConstrainedBindable` to bypass this.

I did just push one commit here, because the max size reset step was fat-fingered (was setting min size instead). Looks alright to me otherwise but I'm going to wait for confirmation from someone else...

bdach (approved):

peppy (approved):

PR #4310, opened 2021-03-24
Allow caps lock events to fire via KeyDown / KeyUp

peppy:
Have tested on windows and macOS. Seems to work as expected.

Also tested on windows with the osu! side caps lock warning and that works once again.

As such, closes https://github.com/ppy/osu/issues/12165.

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -882,7 +882,7 @@ namespace osu.Framework.Platform
         {
             Key key = evtKey.keysym.ToKey();
 
-            if (key == Key.Unknown || key == Key.CapsLock)
+            if (key == Key.Unknown)
                 return;
 
             switch (evtKey.type)


bdach:
This pull does not seem to resolve https://github.com/ppy/osu/issues/11731 on platforms other than windows, because `OsuPasswordTextBox` [uses `Host.CapsLockEnabled](https://github.com/ppy/osu/blob/6a37fb41cc02a6e33f0a0fdf23df6cdf4b60f9de/osu.Game/Graphics/UserInterface/OsuPasswordTextBox.cs#L57), which is only implemented for windows:

https://github.com/ppy/osu-framework/blob/e6230d68d0dafa064d9ebcd193c8693b97fc2ad5/osu.Framework/Platform/GameHost.cs#L151-L154

https://github.com/ppy/osu-framework/blob/e6230d68d0dafa064d9ebcd193c8693b97fc2ad5/osu.Framework/Platform/Windows/WindowsGameHost.cs#L26-L29

Incidentally, [there's an SDL function](https://wiki.libsdl.org/SDL_GetModState) that may be able to help with this.

All that notwithstanding, binding keys to caps seems to work fine, so proceeding with merge otherwise.

bdach (approved):

PR #4311, opened 2021-03-25
Add caps lock state querying support for the general desktop platform case

peppy:
Idea taken from https://github.com/ppy/osu/issues/11731#issuecomment-806177375. I've left the windows specific implementation in place because it doesn't seem to be causing any harm. Could be removed on request I guess.

Tested on macOS; should probably be checked on linux as well.

Closes https://github.com/ppy/osu/issues/11731.


smoogipoo (approved):
Tested to work on linux.

Issue #4309, opened 2021-03-24
Desktop platform's default CursorState and RelativeMouseMode state are not compatible


peppy:
When creating a new project, the default `CursorState` is `Default`, which means the OS cursor is intended to be visible. With the recent changes to mouse handling, SDL relative mouse mode is now enabled by default, hiding the cursor.

These two default are not compatible, so we likely need to decide which to change. In addition, mutating the `CursorState` directly will not work as long as the relative mouse mode is not also changed.

To fix default behaviour, out choices are:

- Disable relative mouse mode by default. Arguably a worse default state for games.
- Hide the mouse by default. Will change the expectations of osu!framework until now.

To fix runtime behaviour, we may need to hard couple these two settings. Needs consideration, because it has the possibility of overriding a user/consumers saved state if done via `MouseHandler.Enabled`. May need an extra internal state to handle this.

Opening this for feedback, because I don't feel confident in making a call on what to do here.

bdach:
I feel like whether relative mode should be on by default is not that clear cut and depends on what the framework intends to be for in general. If the framework is supposed to be dedicated to rhythm games specifically, then sure, it does make sense to enable it by default. But in general, with things like puzzle games or even applications running on o!f you'd probably prefer the system mouse, because that comes with less problems in general, and does not change the OS-wide input characteristics.

Definitely agree with coupling the two things somehow to prevent developers from putting themselves in an invalid state, though.

PR #4312, opened 2021-03-25
Ensure relative mouse mode is never set when the cursor has not been hidden

peppy:
It turns out this was quite simple to implement due to `MouseHandler` already preferring relative mode given circumstances of the game. The result now is that even if the `MouseHandler` is set to prefer relative mode, it will automatically be disabled when the OS cursor is visible. I think this makes sense.

Closes https://github.com/ppy/osu-framework/issues/4309.


bdach (approved):
Seems to work on linux and I think it the behaviour makes sense, but I don't feel entirely confident in merging on my own.

smoogipoo (approved):

PR #4313, opened 2021-03-26
Add bounding box method for circular arcs

Naxesss:
We only need the end-points and each side of the arc to compute the bounding box. This method should be less intensive and marginally more accurate compared to approximating the entire circular arc and then computing the bounding box from resulting points.

Math logic used is the same, hence `circularArcHelper`. Because the interesting output from that helper method is not just one value, it passes them on to a given function instead. An alternative to this would be using the `out` keyword, or a custom struct, not sure which is preferred.

---

Main use case is https://github.com/ppy/osu/pull/12122. Here the bounding box is used as a limit for how large a PerfectCurve path is allowed to be before defaulting to Bezier. Said bounding box is computed every time the control points update, for example while dragging nodes or initially placing sliders.

### DIFF
--- a/osu.Framework/Utils/PathApproximator.cs
+++ b/osu.Framework/Utils/PathApproximator.cs
@@ -213,6 +213,10 @@ namespace osu.Framework.Utils
             double step = right_angle * pr.Direction;
 
             double quotient = pr.ThetaStart / right_angle;
+            // choose an initial right angle, closest to ThetaStart, going in the direction of the arc.
+            // if the direction/step is positive - choose the smallest multiple of PI / 2 greater than ThetaStart.
+            // if the direction/step is negative - choose the greatest multiple of PI / 2 smaller than ThetaStart.
+            // thanks to this, when looping over quadrant points to check if they lie on the arc, we only need to check against ThetaEnd.
             double closestRightAngle = right_angle * (pr.Direction > 0 ? Math.Ceiling(quotient) : Math.Floor(quotient));
 
             // at most, four quadrant points must be considered.
@@ -220,7 +224,7 @@ namespace osu.Framework.Utils
             {
                 double angle = closestRightAngle + step * i;
 
-                // check whether angle exceeds the range [ThetaStart, ThetaEnd].
+                // check whether angle has exceeded ThetaEnd.
                 // multiplying by Direction eliminates branching caused by the fact that step can be either positive or negative.
                 if (Precision.DefinitelyBigger((angle - pr.ThetaEnd) * pr.Direction, 0))
                     break;
--- a/osu.Framework/Utils/PathApproximator.cs
+++ b/osu.Framework/Utils/PathApproximator.cs
@@ -164,67 +164,27 @@ namespace osu.Framework.Utils
         /// <returns>A list of vectors representing the piecewise-linear approximation.</returns>
         public static List<Vector2> ApproximateCircularArc(ReadOnlySpan<Vector2> controlPoints)
         {
-            Vector2 a = controlPoints[0];
-            Vector2 b = controlPoints[1];
-            Vector2 c = controlPoints[2];
-
-            // If we have a degenerate triangle where a side-length is almost zero, then give up and fallback to a more numerically stable method.
-            if (Precision.AlmostEquals(0, (b.Y - a.Y) * (c.X - a.X) - (b.X - a.X) * (c.Y - a.Y)))
-                return ApproximateBezier(controlPoints);
-
-            // See: https://en.wikipedia.org/wiki/Circumscribed_circle#Cartesian_coordinates_2
-            float d = 2 * (a.X * (b - c).Y + b.X * (c - a).Y + c.X * (a - b).Y);
-            float aSq = a.LengthSquared;
-            float bSq = b.LengthSquared;
-            float cSq = c.LengthSquared;
-
-            Vector2 centre = new Vector2(
-                aSq * (b - c).Y + bSq * (c - a).Y + cSq * (a - b).Y,
-                aSq * (c - b).X + bSq * (a - c).X + cSq * (b - a).X) / d;
-
-            Vector2 dA = a - centre;
-            Vector2 dC = c - centre;
-
-            float r = dA.Length;
-
-            double thetaStart = Math.Atan2(dA.Y, dA.X);
-            double thetaEnd = Math.Atan2(dC.Y, dC.X);
-
-            while (thetaEnd < thetaStart)
-                thetaEnd += 2 * Math.PI;
-
-            double dir = 1;
-            double thetaRange = thetaEnd - thetaStart;
-
-            // Decide in which direction to draw the circle, depending on which side of
-            // AC B lies.
-            Vector2 orthoAtoC = c - a;
-            orthoAtoC = new Vector2(orthoAtoC.Y, -orthoAtoC.X);
-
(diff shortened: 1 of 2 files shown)


osu.Framework/Utils/PathApproximator.cs
```diff
@@ -276,6 +238,105 @@ public static List<Vector2> ApproximateLagrangePolynomial(ReadOnlySpan<Vector2>
...
+
+        /// <summary>
+        /// Computes all arguments necessary to approximate a circular arc given its control points.
+        /// These arguments are then passed to the given function, the result of which is returned.
+        /// </summary>
+        /// <param name="controlPoints">Three distinct points on the arc.</param>
+        /// <param name="pointsFunc">A function whose resulting points are returned. It is given a
+        /// start angle, delta angle, direction, radius, and centre.</param>
+        /// <returns>A list of vectors returned from the function provided.</returns>
+        private static List<Vector2> circularArcHelper(ReadOnlySpan<Vector2> controlPoints, Func<double, double, double, float, Vector2, List<Vector2>> pointsFunc)
```
peppy, line 292: I haven't gone through this in detail to find the reasoning, but the function argument here probably can't exist in this way. It's very hard to follow what's going on..

bdach: The easiest way to remove this would probably be to invert the flow and declare a private struct with the arguments of the lambda, and just return that data here instead of having that lambda be called on the results of the approximation.

Naxesss: Switched over to using a `CircularArcProperties` class (class because that allows me to return null for falling back to bezier approximation, seems more intuitive than including a "valid" property). Should be easier to follow now I think.

peppy: Looks much better now.

bdach:
I've written a test case for this, and it seems to be a bit... broken in some edge cases, particularly ones involving cardinal directions.

2021-03-26-231811_570x500_scrot
2021-03-26-231935_699x485_scrot
2021-03-26-232144_527x477_scrot

I might have messed up writing the test, but stuff like this could certainly explain it:

2021-03-26-231637_906x149_scrot

Will push that in a sec.

bdach:
I've fiddled with this a bit and believe I have gotten to a relatively stable state:

https://user-images.githubusercontent.com/20418176/112729146-ff539c00-8f2a-11eb-878e-b9549878aa14.mp4

Seems good now, but my changes were quite extensive, so this needs a double check.

bdach (approved):

Naxesss:
Looks sane to me. Only perhaps confusing part would be
```cs
double quotient = pr.ThetaStart / right_angle;
double closestRightAngle = right_angle * (pr.Direction > 0 ? Math.Ceiling(quotient) : Math.Floor(quotient));
```
as this makes the assumption that `pr.ThetaStart` will be negative for negative directions (which is always true, but that it is always the case isn't very intuitive, and depends on implementation details of `circularArcProperties`).

Could do this, which makes the `closestRightAngle` more clearly in the direction we're going and is otherwise equivalent:
```cs
double quotient = Math.Abs(pr.ThetaStart / right_angle);
double closestRightAngle = right_angle * Math.Ceiling(quotient) * pr.Direction;
```

bdach:
I don't think the sign of `ThetaStart` matters there in that context at all? The goal there was to find the next multiple of `PI / 2` to advance to given `Direction`. Which is why the ceiling is taken if the step is positive, and the floor is taken if the step is negative.

Naxesss:
The statement in the for loop only checks whether the `angle` is greater than where it should stop:
```cs
if (Precision.DefinitelyBigger((angle - pr.ThetaEnd) * pr.Direction, 0))
    break;
```
If `angle` is before the starting point of the arc, it would simply be added as a point despite not being in the arc, which is why it's important that `closestRightAngle` starts inside of it. Where it starts is dependent on the sign of `pr.ThetaStart` because of flooring/ceiling. Therefore the sign does matter in this context.

If I were to put `quotient = pr.ThetaStart * -1f / right_angle`, for example, then I would get this result in the test:

image

bdach:
> If angle is before the starting point of the arc, it would simply be added as a point despite not being in the arc, which is why it's important that `closestRightAngle` starts inside of it.

Yep, with you up to this point.

> Where it starts is dependent on the sign of `pr.ThetaStart` because of flooring/ceiling.

But I'm not seeing this one. Let's for the sake of argument assume the following (in degrees, not radians, for simplicity):

```
pr.ThetaStart = 120, pr.ThetaRange = 240, pr.Direction = -1
```

We have a positive starting angle, but a negative direction. Because of this, at the point of picking the first right angle, we pick 90deg (`Math.Floor(120 / 90) * 90`), and then proceed with 0deg, -90deg and -180deg (the last of which fails that range test).

Conversely, let's assume:

```
pr.ThetaStart = -120, pr.ThetaRange = 240, pr.Direction = 1
```

The starting angle there is -90deg (`Math.Ceiling(-120 / 90) * 90`), and the next angles are 0deg, 90deg, and -180deg (which is again outside of the range).

All the floor/ceil choice does is pick the direction of iteration across multiplies of `PI / 2`. It should be agnostic toward the sign of the starting point, as far as I can tell.

bdach:
I've attempted to clarify the comments some more. Let me know whether that did anything for understanding.

Naxesss:
Just realized after testing a bit that my confusion isn't necessarily about signs but more about the range `ThetaStart` can assume. If it's anything other than [-π, π], e.g. [0, 2π], things seem to break down. I recall this being the reason for doing modulus operations in my initial solution.

Take for example
```
pr.ThetaStart = -60, pr.ThetaRange = 240, pr.Direction = -1
quotinent = 120 / 240 = -0.25
closestRightAngle = 90 * Math.Floor(-0.25) = 0
```
Compared to an equivalent angle (-60 = 300)
```
pr.ThetaStart = 300, pr.ThetaRange = 240, pr.Direction = -1
quotinent = 300 / 240 = 1.25
closestRightAngle = 90 * Math.Floor(1.25) = 1
```
Latter skips 0, meaning it'll `break` before reaching all the way around.

Not sure this is worth dealing with, though - wouldn't really be as simple of a fix as I thought. So I'd say keep it as-is.

---

As for new comments - they do clarify how the loop condition works, so that's good. I probably wouldn't keep the "if x - choose y" ones, though, as that can for the most part be inferred from the code (variable name + "in the direction of the arc" makes that clear enough already).

osu.Framework/Utils/PathApproximator.cs
```diff
@@ -276,6 +290,79 @@ public static List<Vector2> ApproximateLagrangePolynomial(ReadOnlySpan<Vector2>
             return result;
         }
 
+        private class CircularArcProperties
```
smoogipoo, line 293: I'd like to avoid this allocation, because this could be used quite often.

I suggest making this a struct and adding a property like `IsValid` that is set to `true` only in the constructed case. `default(CircularArcProperties)` will then have `IsValid = false`.

smoogipoo (changes requested):
Haven't looked over the implementation yet.

osu.Framework/Utils/PathApproximator.cs
Naxesss: Changed to readonly struct with an `IsValid` property. Didn't realize you could use `default` for this, that's good to know.

smoogipoo (approved):
Lgtm

PR #4314, opened 2021-03-27
Use BASSWasapi to set BASS Config on Windows Systems

shaneMenzies:
Proposed solution to [this issue on osu!lazer](https://github.com/ppy/osu/issues/12056). Adds in BASSWasapi in order to give better control of the BASS Device Buffer Size and Update Period, and on Windows Systems tries to initialize BASS with Wasapi first.


bdach:
The way I see it this should be the absolute last resort. The preferred solution would be to get an upstream fix to bass itself.

PR #4315, opened 2021-03-29
Avoid feedback from input devices reporting via multiple handlers

peppy:
With the introduction of tablet support, there have been a number of users reporting their mouse jumping or "rubber banding" between two locations. The cause of this is having tablet drivers or software running external to a game, reporting back as mouse cursor updates, in addition to our local handling of tablets via OTD.

Until now, the general advice for such cases is to "close the other drivers/software", but this is a pretty bad user experience. For instance, an illustration may use the wacom software for their graphical work while also using the tablet to play games.

This PR aims to ignore externally occuring mouse events while a tablet is reporting via OTD. It may also be used in the futrue for other input handler where similar patterns can occur. This is done via a simple debounce pattern with a few basic rules:

- `InputHandler`s which are initialised higher in the construction order via `CreateAvailableInputHandlers` have priority over positional input. If a lower priority device was previously updating the cursor, a higher (or "earlier") device can immediately take over.
- A lower priority (or "later") device can take over the positional input flow after a short debounce window. This is set to 50ms, but as mentioned in the inline comments can be shortened if required. It feels quite okay as is, though.

https://user-images.githubusercontent.com/191335/112793588-824a3300-90a0-11eb-88fc-e6e7372aa9ba.mp4

Closes https://github.com/ppy/osu/issues/12171.

# Breaking Changes

## InputHandler no longer has a Priority property

Priority is now decided by the construction order of `InputHandler`s in `CreateAvailableInputHandlers`.

### DIFF
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -532,7 +532,7 @@ namespace osu.Framework.Input
                     // report mouse position data at a time. Handlers are given priority based on their constructed order.
                     // Devices which higher priority are allowed to take over control immediately, after which a delay is enforced (on every subsequent positional report)
                     // before a lower priority device can obtain control.
-                    if (i is MousePositionAbsoluteInput)
+                    if (i is MousePositionAbsoluteInput || i is MousePositionRelativeInput)
                     {
                         if (mouseSource == null // a new device taking control when no existing preference is present.
                             || mouseSource == h // if this is the device which currently has control, renew the debounce delay.
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -500,7 +500,7 @@ namespace osu.Framework.Input
 
         private InputHandler mouseSource;
 
-        private double sourceDebounceTimeRemaining;
+        private double mouseSourceDebounceTimeRemaining;
 
         private double lastPendingInputRetrievalTime;
 
@@ -537,10 +537,10 @@ namespace osu.Framework.Input
                         if (mouseSource == null // a new device taking control when no existing preference is present.
                             || mouseSource == h // if this is the device which currently has control, renew the debounce delay.
                             || !reachedPreviousMouseSource // we have not reached the previous mouse source, so a higher priority device can take over control.
-                            || sourceDebounceTimeRemaining <= 0) // a lower priority device taking over control if the debounce delay has elapsed.
+                            || mouseSourceDebounceTimeRemaining <= 0) // a lower priority device taking over control if the debounce delay has elapsed.
                         {
                             mouseSource = h;
-                            sourceDebounceTimeRemaining = mouse_source_debounce_time;
+                            mouseSourceDebounceTimeRemaining = mouse_source_debounce_time;
                         }
                         else
                         {
@@ -557,7 +557,7 @@ namespace osu.Framework.Input
                     reachedPreviousMouseSource = true;
             }
 
-            sourceDebounceTimeRemaining -= elapsed;
+            mouseSourceDebounceTimeRemaining -= elapsed;
             return inputs;
         }
 
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -553,6 +553,7 @@ namespace osu.Framework.Input
                 }
 
                 // track whether we have passed the handler which is currently in control of positional handling.
+                // importantly, this is updated regardless of whether the handler has reported any new inputs.
                 if (mouseSource == h)
                     reachedPreviousMouseSource = true;
             }
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -496,36 +496,68 @@ namespace osu.Framework.Input
 
         private readonly List<IInput> inputs = new List<IInput>();
 
+        private readonly List<IInput> dequeuedInputs = new List<IInput>();
+
         private InputHandler mouseSource;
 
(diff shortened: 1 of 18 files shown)


osu.Framework/Input/InputManager.cs
```diff
@@ -496,13 +496,68 @@ private void updateKeyRepeat(InputState state)
...
+                dequeuedInputs.Clear();
+                h.CollectPendingInputs(dequeuedInputs);
+
+                foreach (var i in dequeuedInputs)
+                {
+                    // To avoid the same device reporting via two channels (and causing feedback), only one handler should be allowed to
+                    // report mouse position data at a time. Handlers are given priority based on their constructed order.
+                    // Devices which higher priority are allowed to take over control immediately, after which a delay is enforced (on every subsequent positional report)
+                    // before a lower priority device can obtain control.
+                    if (i is MousePositionAbsoluteInput)
```
smoogipoo, line 535: Why only absolute input? I mean I understand why but the comment above says "only one handler should be allowed to report mouse position data at a time", giving no indication as to why relative input is excluded.

osu.Framework/Input/InputManager.cs
```diff
@@ -496,13 +496,68 @@ private void updateKeyRepeat(InputState state)
 
         private readonly List<IInput> inputs = new List<IInput>();
 
+        private readonly List<IInput> dequeuedInputs = new List<IInput>();
+
+        private InputHandler mouseSource;
+
+        private double sourceDebounceTimeRemaining;
```
smoogipoo, line 503: `mouseSourceDebounceTimeRemaining`

osu.Framework/Platform/DesktopGameHost.cs
```diff
@@ -88,12 +88,13 @@ private void startIPC()
             new InputHandler[]
             {
                 new KeyboardHandler(),
-                new MouseHandler(),
-                new JoystickHandler(),
-                new MidiHandler(),
 #if NET5_0
+                // tablet should get priority over mouse to correctly handle cases where tablet drivers report as mice as well.
```
smoogipoo, line 92: `InputHandler.Priority` is a thing. Perhaps `InputManager.InputHandlers` should be sorted by it?

osu.Framework/Input/InputManager.cs
```diff
@@ -496,13 +496,68 @@ private void updateKeyRepeat(InputState state)
...
+                {
+                    // To avoid the same device reporting via two channels (and causing feedback), only one handler should be allowed to
+                    // report mouse position data at a time. Handlers are given priority based on their constructed order.
+                    // Devices which higher priority are allowed to take over control immediately, after which a delay is enforced (on every subsequent positional report)
+                    // before a lower priority device can obtain control.
+                    if (i is MousePositionAbsoluteInput)
+                    {
+                        if (mouseSource == null // a new device taking control when no existing preference is present.
+                            || mouseSource == h // if this is the device which currently has control, renew the debounce delay.
+                            || !reachedPreviousMouseSource // we have not reached the previous mouse source, so a higher priority device can take over control.
```
smoogipoo, line 539: Isn't this going to allow a _lower_ priority device to take over, not higher?

For example:
```
Frame 1:
OTD
OTD
MH
MH
```
Frame 2:
```
MH <-- MH now takes over since reachedPreviousMouseSource == false
MH
```
Frame 3:
```
OTD <-- OTD now takes over reachedPreviousMouseSource == false
OTD
MH
MH
```
Which brings two things into question:
1. OTD and MH must be operating at the same frequency of events otherwise it's going to ping-pong between the two handlers
2. Is the debounce time actually doing anything? Seems like maybe this should be `!reachedPreciousMouseSource && sourceDebounceTimeRemaining <= 0` or something.

smoogipoo (changes requested):
I'm not sure I fully agree with this PR, but I have no better suggestions yet other than to directly couple OTD and MH, with the obvious drawback that this doesn't handle the same case in other usages (like iOS touch + mouse handler).

osu.Framework/Platform/DesktopGameHost.cs
peppy: Either that or we should just remove `Priority` as unused. Which do you prefer?

osu.Framework/Input/InputManager.cs
peppy: You're missing the part where that flag is set regardless of whether a source is returning inputs or not. Maybe this needs to be marked more clearly? I tried to do so with the inline comment on that line already.

osu.Framework/Platform/DesktopGameHost.cs
smoogipoo: If it's removed, then I'd expect the xmldoc on `CreateAvailableInputHandlers` to also mention the ordering. Otherwise, up to you.

osu.Framework/Input/InputManager.cs
peppy: Can add relative, sure. Was just keeping it simple.

osu.Framework/Platform/DesktopGameHost.cs
peppy: Just going to remove for now. It's a remnant of stable that we haven't really needed until now after all.

osu.Framework/Input/InputManager.cs
smoogipoo: Right...

smoogipoo (approved):
Let's go with this for the time being, I think.

smoogipoo (approved):

smoogipoo (approved):

PR #4316, opened 2021-03-29
Update libraries

peppy:
Mainly to bring OTD up-to-date with upstream fixes.

### DIFF
--- a/osu.Framework/osu.Framework.csproj
+++ b/osu.Framework/osu.Framework.csproj
@@ -24,18 +24,22 @@
     <PackageReference Include="managed-midi" Version="1.9.14" />
     <PackageReference Include="Markdig" Version="0.22.1" />
     <PackageReference Include="FFmpeg.AutoGen" Version="4.3.0.1" />
-    <PackageReference Include="Microsoft.Extensions.ObjectPool" Version="5.0.3" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.9.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="3.9.0">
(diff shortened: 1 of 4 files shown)


osu.Framework/osu.Framework.csproj
```diff
@@ -24,18 +24,22 @@
     <PackageReference Include="managed-midi" Version="1.9.14" />
     <PackageReference Include="Markdig" Version="0.22.1" />
     <PackageReference Include="FFmpeg.AutoGen" Version="4.3.0.1" />
-    <PackageReference Include="Microsoft.Extensions.ObjectPool" Version="5.0.3" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.9.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="3.9.0">
+      <NoWarn>NU1701</NoWarn> <!-- Requires .NETFramework for MSBuild, but we use Microsoft.Build.Locator which allows this package to work in .NETCoreApp. -->
+    </PackageReference>
```
smoogipoo, line 30: Why did you move these?

peppy: Dunno, rider did. Is there an issue?

smoogipoo: At least `Microsoft.CodeAnalysis.Workspaces.MSBuild` is incompatible with netstandard so it feels wrong to keep it here, also with the comment being what it is. The only reason it works is because `Microsoft.Build.Locator` is used which is only compatible with netcore, and remains in the original location.

peppy: Right, it was inside some conditional. I guess rider ignores those?

smoogipoo (approved):

PR #4327, opened 2021-03-31
Remove unnecessary redirection methods when triggering SDL window events

peppy:
Just looking to make tracking flow a touch easier when working with this class. These were left over from a forgotten time.


smoogipoo (approved):
Looks fine. I think we'll eventually have to take another look, because this is potentially allocating a lot, especially with 1000Hz mice:

Screenshot_2021-03-31_16-47-28

(This PR doesn't change this)

PR #4336, opened 2021-04-02
Fix 100% sample volumes for 1 frame in single-threaded execution

smoogipoo:
Fixes https://github.com/ppy/osu/issues/11845

## Repro

It's hard to unit test this, so I've left it out. It can be tested by adding a breakpoint on a condition such as:
```csharp
if (Bass.ChannelGetAttribute(channel, ChannelAttribute.Volume, out var vol) && vol > 0.8f)
{
}
```
After `Bass.ChannelPlay()` in `playChannel`.

## Explanation

This issue can only be repro'd in single-threaded mode because inline execution is only allowed in that mode.

`Sample`s and `SampleChannel`s are added to the audio hierarchy (and thus become alive) only when a `SampleChannel` is `Play`'ed. If at any point the `SampleChannel` becomes `Playing = false`, it's removed from the hierarchy, which also causes the parenting `Sample` to be removed from the hierarchy if no other `SampleChannel`s are alive.

Thus, in the case of `SampleChannelBass`, `Play()` intends to set `Playing = true` immediately such that the audio thread sees `IsAlive = true`:
https://github.com/ppy/osu-framework/blob/36f90369065065777d07f19788a10c597e36360e/osu.Framework/Audio/Sample/SampleChannel.cs#L47
Whether this is the proper expectation of `Playing` is probably a consideration for another day - these changes intend to not break this flow.

## The issue

The previous implementation enqueued the action to play the channel before making the channel alive, such that in the case that the audio thread runs concurrently, the audio thread will play the sample before updating the `Playing` state:
https://github.com/ppy/osu-framework/blob/36f90369065065777d07f19788a10c597e36360e/osu.Framework/Audio/Sample/SampleChannelBass.cs#L89-L117

In single-threaded execution mode the enqueued action is run in-line, such that there is a period of time that the channel is playing at full volume without yet being part of the audio hierarchy. In this case, being part of the audio hierarchy determines whether adjustment bindables have been set:
https://github.com/ppy/osu-framework/blob/36f90369065065777d07f19788a10c597e36360e/osu.Framework/Audio/AudioCollectionManager.cs#L17-L26

## The fix

A new local variable `enqueuedPlaybackStart` is set when the `Play()` action is enqueued and reset when it has completed. The `Playing` override ors this value with the local `playing` state to determine the final value of `Playing` independent of audio thread execution ordering.

### DIFF
--- a/osu.Framework/Audio/Sample/SampleChannelBass.cs
+++ b/osu.Framework/Audio/Sample/SampleChannelBass.cs
@@ -10,10 +10,27 @@ namespace osu.Framework.Audio.Sample
     {
         private readonly SampleBass sample;
         private volatile int channel;
+
+        /// <summary>
+        /// Whether the channel is currently playing.
+        /// </summary>
+        /// <remarks>
+        /// This is set to <c>true</c> immediately upon <see cref="Play"/>, but the channel may not be audibly playing yet.
+        /// </remarks>
+        public override bool Playing => playing || enqueuedPlaybackStart;
+
+        private volatile bool playing;
+
+        /// <summary>
+        /// <c>true</c> if the user last called <see cref="Play"/>.
+        /// <c>false</c> if the user last called <see cref="Stop"/>.
+        /// </summary>
         private volatile bool userRequestedPlay;
 
-        public override bool Playing => playing;
-        private volatile bool playing;
+        /// <summary>
+        /// Whether the playback start has been enqueued.
+        /// </summary>
+        private volatile bool enqueuedPlaybackStart;
 
         private readonly BassRelativeFrequencyHandler relativeFrequencyHandler;
         private BassAmplitudeProcessor bassAmplitudeProcessor;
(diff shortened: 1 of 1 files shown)


peppy:
I've read through your explanation four times but am still having difficulty understanding where the `playing` / `IsAlive` state becomes important (ie. part of the issue).

As far as I can tell, the main fix here was reordering `base.Play()` and `playChannel()` calls, no?

smoogipoo:
I hoped to explain why the previous code was done as is, why `playing = true` is set immediately, and what the failure would be if `base.Play()` happened before `playChannel`. This is with the previous code.

To try again:
* `playing =  true` is set so that `IsAlive` becomes true, so that the component is not immediately removed from the audio hierarchy.
* `base.Play()` adds the component to the hierarchy, allowing it to receive updates, and _binds adjustments_.

The failure occurs if an update comes in with concurrent threading and the `playChannel()` action hasn't been enqueued yet, so the code in `UpdateState()` runs which sets `playing = false` which kills the component. So the previous code required `playChannel()` to be called first to enqueue the action to for sure be run the next update, but in single threaded mode this is run in-line and begins playback where it was intended to be delayed until after `base.Play()`.

The new state (`enqueuedPlaybackStart`) is completely managed by the `Play()`/`playChannel()` flow and doesn't interact with `UpdateState` like the `playing` state does, which allows `playChannel()` to be placed after `base.Play()` as intended.

peppy:
Aha, so the original code was done in that way specifically *for* the multithreaded scenario. Sounds fine, just wanted to confirm the actual fix was the ordering change, since that wasn't mentioned anywhere.

smoogipoo:
Yup, what you've said is all correct.

peppy (approved):

peppy:
Seems to work as expected and read about as good as I think we can hope for.

Issue #2136, opened 2019-02-04
Left and right modifier keys are not distinguishable


oSumAtrIX:
**Describe your problem:** 
When trying to map LSHIFT in settings it gets mapped as SHIFT in general. This also applies to CTRL and RCTRL. 

**osu!lazer version:** 
 2019.131.0

PR #4340, opened 2021-04-05
Add left/right modifier support to KeyBindingContainer

peppy:
Don't really have much to say about this one. Pretty hellish to get the implementation right, and likely not going to be a simple one to review/understand, but it works.

May need to assess whether the array allocation overhead is an issue (although I think the scenarios where it is called are already limited to when input changes, so should be okay).

Closes https://github.com/ppy/osu-framework/issues/2136.

### DIFF
--- a/osu.Framework/Input/Bindings/KeyBindingContainer.cs
+++ b/osu.Framework/Input/Bindings/KeyBindingContainer.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using osu.Framework.Allocation;
 using osu.Framework.Extensions.IEnumerableExtensions;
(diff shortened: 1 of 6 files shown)


osu.Framework/Input/Bindings/KeyBindingContainer.cs
```diff
@@ -284,12 +297,14 @@ private void releasePressedActions()
...
         {
+            Logger.Log($"new key released {releasedKey}");
+
             var pressedCombination = KeyCombination.FromInputState(state);
 
             // we don't want to consider exact matching here as we are dealing with bindings, not actions.
             var newlyReleased = pressedBindings.Where(b => !b.KeyCombination.IsPressed(pressedCombination, KeyCombinationMatchingMode.Any)).ToList();
 
-            Trace.Assert(newlyReleased.All(b => b.KeyCombination.Keys.Contains(releasedKey)));
+            // Trace.Assert(newlyReleased.All(b => KeyCombination.ContainsKey(b.KeyCombination.Keys, releasedKey)));
```
bdach, line 307: Was this purposefully commented out, or just for testing? The log statements in this file are for debugging, I assume, but I'm not sure about this one.

peppy: The debug logging and this needs to be reverted, completely forgot.

Hopefully this doesn't trigger...

peppy:
Method complexity CI failure is likely not fixable, safe to ignore.

smoogipoo (approved):
This looks pretty good to me. The only minor concern I have is the allocation on `FromKeys()`, but I figure both this and the O(N^2)'dness of `ContainsKey()` are rare enough to not cause issues.

PR #4343, opened 2021-04-07
Listen to midi messages on all channels

ColdVolcano:
- Should close ppy/osu#12198

Note On/Off messages include channel data on it's status byte's low nibble, which caused messages sent over channels other than 1 (hex 0) to be discarded.
As we only handle Note On/Off messages it should be safe to ignore this nibble.

Tested using a Novation Launchpad Mk2 set to channels 6-8 and 14-16.


smoogipoo:
Tested on channels 1-16 (linux), works as expected. :+1:

smoogipoo (approved):

Issue #4322, opened 2021-03-30
Switching/Minimizing the window does not work when using Fullscreen mode


Sinamer:
The window stays on top, the only change being that the game stops accepting input (clicking LMB returns the control of the mouse). Neither Alt+Tab (for task switcher) nor Win+D (for minimizing all windows) works. The fps doesn't get limited to 60 as well, so it probably doesn't go into unfocused state at all? This happens only since the last release.

Here is how it looks (obs couldn't capture the task switcher, sorry for using the camera):
https://streamable.com/mj8z1y

Here is DxDiag if that is of any help (note this is a Nvidia Optimus laptop, but I'm running the game using the integrated graphics):
[DxDiag.txt](https://github.com/ppy/osu-framework/files/6231610/DxDiag.txt)

Susko3:
Only happening if `High precision mouse` is turned on.

With it disabled, Alt+Tabbing works as expected, and the game renders at 60 fps in the background.
But the osu!lazer window still doesn't minimize. The window is also visible after Win+D

Alt-Tabbing to another windows and then pressing Win+D leaves osu! in a weird state where it'll be the topmost window, but will pass clicks trough it, Right click open the desktop right click menu.

Sinamer:
Thanks for the tip, disabling `High precision mouse` does help with switching the window via Alt+Tab. Also having the same behavior with Win+D still leaving the game window on top, although in unfocused state now.

peppy:
This will likely be improved in the next release. Please report back when that happens!

Susko3:
This issue isn't completely fixed on latest master. Alt+Tabbing now works, but Win+D still doesn't. The game stays on top of the desktop, regardless if relative mode is enabled or disabled. Also broken when just opening Start with the Win key. Start opens, but the game window is still visually in the foreground. Clicking in the Start menu will click on Start, as if Start was in foreground.

PR #4344, opened 2021-04-07
Fix alt-tab not working well when in fullscreen mode

peppy:
Closes https://github.com/ppy/osu-framework/issues/4322.


smoogipoo (approved):

Susko3 (commented):
This change breaks the expected behavior when the window looses focus without the cursor leaving the window. If that happens, the cursor position stops getting updated (as relative mode is still enabled).

Easiest way to cause the issue is to open the Start menu. Moving the mouse will not update the cursor until the cursor leaves the window. It's a problem in both full screen and windowed modes.

PR #4348, opened 2021-04-07
Set the SDL hint to minimize on focus loss in fullscreen

### LINKED ISSUE #4346: Pressing windows key while mouse in window behaves incorrectly
In windowed mode on windows, hit the windows key while the game is focused and the cursor in the window. The cursor will stop responding but the windows cursor will not appear.

Susko3:
The default value for `SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS` has changed in SDL 2.0.14
The change was made in https://github.com/libsdl-org/SDL/commit/1947ca7028ab165cc3e6cbdb0b4b7c4db68d1710

Setting it to `true` restores the expected behavior, which we had before the bump of SDL libraries.

This PR also reverts the change made in #4344 as it's not needed, and that change created more problems.
Closes #4346 because of the revert.

### DIFF
--- a/osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
+++ b/osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
@@ -62,13 +62,7 @@ namespace osu.Framework.Input.Handlers.Mouse
             window = desktopWindow;
 
             isActive = window.IsActive.GetBoundCopy();
-            isActive.BindValueChanged(active =>
-            {
-                // importantly, we shouldn't trigger an update when the window becomes inactive.
-                // this can cause windows to get stuck in fullscreen.
-                if (active.NewValue)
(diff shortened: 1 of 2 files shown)


Susko3:
Will probably fix https://github.com/ppy/osu/issues/12260

Joehuu:
Doesn't fix the mentioned issue.

peppy:
Nice catch. I actually saw this change and forgot to test whether it is something we wanted. Seems like a weird default to change, but I guess they were doing testing with it. We'd want to report our results back to upstream so they are aware of the new issues introduced by this change.

peppy (approved):

peppy:
Looks to work great on windows, but not fix the macOS issue. Probably makes sense since I think that one has always been an issue, even before the recent version bump.

Issue #4347, opened 2021-04-07
Interface sounds continue to play with the window minimized


K4ktus123:
**Describe the bug:** 
Interface sounds play even after the window is minimized.

**Screenshots or videos showing encountered issue:** 

https://user-images.githubusercontent.com/55000815/113934529-2a28d480-97f6-11eb-856b-6ecf0e67e5d5.mp4

**osu!lazer version:** 
2021.407.1

**Logs:**
[database.log](https://github.com/ppy/osu/files/6274525/database.log)
[network.log](https://github.com/ppy/osu/files/6274526/network.log)
[performance.log](https://github.com/ppy/osu/files/6274527/performance.log)
[runtime.log](https://github.com/ppy/osu/files/6274528/runtime.log)
[updater.log](https://github.com/ppy/osu/files/6274529/updater.log)

frenzibyte:
Can reproduce on macOS as well apparently.

Susko3:
Sounds are also playing when osu! window is obstructed by another window. (Didn't happen on 2021.331.0)

FamousPig:
I don't think this is a framework issue. 
![2021-04-08_01-07-36](https://user-images.githubusercontent.com/22643729/113945580-e7bbc380-9806-11eb-8ae8-2e7d362c69b1.gif)
This Framework test scene (NestedHover) appears to work correctly. On a quick check (as in checking out the previous merge) this appears to be another regression from https://github.com/ppy/osu/pull/12225.

Putting this line (copied from `UserInputManager`) into `GlobalInputManager` solves the issue:

```C#
 protected override bool HandleHoverEvents => Host.Window?.CursorInWindow.Value ?? true;
```

But I can still not really wrap my head around how that PR functions and can't tell whether this is the correct solution or just a bandaid, so I'm not gonna PR it.

frenzibyte:
Sounds like `PassThroughInputManager` should read `HandleHoverEvents` value from its parent when using it then, but not sure how correct that would be.

peppy:
I think that solution sounds amicable, since it is already directly referencing the parent for the input state as well when that bool is set.

PR #4350, opened 2021-04-08
Fix PassThroughInputManager ignoring parent HandleHoverEvents state

peppy:
Closes https://github.com/ppy/osu-framework/issues/4347 via proposed solution in that issue thread.

I tested this manually using `TestScenePassThroughInputManger` via the hover event count. Not sure it's worth adding automated test coverage of this, but can be done if required.


smoogipoo (approved):
Seems reasonable

PR #4359, opened 2021-04-12
Fix readable key combination string including main modifier along direction-specific

frenzibyte:
Was causing key combinations like `(Shift, LShift, Control, LControl, R)` to display as `Ctrl-LCtrl-Shift-LShift-R`, instead of `LCtrl-LShift-R`, as can be seen in osu!'s key binding configuration:
CleanShot 2021-04-12 at 09 06 27@2x

Now written to hide the main modifier if the direction-specific one exists in the combination, otherwise show it.

### DIFF
--- a/osu.Framework/Input/Bindings/KeyCombination.cs
+++ b/osu.Framework/Input/Bindings/KeyCombination.cs
@@ -232,7 +232,7 @@ namespace osu.Framework.Input.Bindings
 
         public string ReadableString()
         {
-            var sortedKeys = Keys.GetValuesInOrder();
+            var sortedKeys = Keys.GetValuesInOrder().ToArray();
 
             return string.Join('-', sortedKeys.Select(key =>
             {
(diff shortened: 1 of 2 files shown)


osu.Framework/Input/Bindings/KeyCombination.cs
```diff
@@ -233,7 +233,38 @@ public override int GetHashCode()
         public string ReadableString()
         {
             var sortedKeys = Keys.GetValuesInOrder();
```
peppy, line 235: This needs a `.ToArray` or similar, else it's going to be enumerated many many times.

peppy (changes requested):

osu.Framework/Input/Bindings/KeyCombination.cs
frenzibyte: Ah, didn't notice that.

peppy (approved):
This whole code is probably doing far too much LINQ, but pretty rarely called so guess it's okay for now.

PR #4360, opened 2021-04-12
Allow OverlayContainers to optionally not block scroll

peppy:
As proposed in https://github.com/ppy/osu/pull/12375#issuecomment-817562409.

Will add test coverage if this is seen as an amicable direction.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneFocusedOverlayContainer.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneFocusedOverlayContainer.cs
@@ -47,8 +47,9 @@ namespace osu.Framework.Tests.Visual.UserInterface
             AddAssert("not visible", () => overlayContainer.State.Value == Visibility.Hidden);
         }
 
-        [Test]
-        public void TestScrollBlocking()
+        [TestCase(true)]
+        [TestCase(false)]
+        public void TestScrollBlocking(bool isBlocking)
         {
             AddStep("create container", () =>
             {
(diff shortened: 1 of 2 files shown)


smoogipoo:
I suppose it's fine. This is all way above my head so I have no suggestions.

peppy:
It all revolves around one question: do you consider scroll wheel input to be positional or not?

- The previous code is considering it to be positional
- osu! wants to handle the case where it is not

Adding the extra variable allows us to adjust this expectation so osu! overlays can not block mouse wheel, so it can be handled to adjust volume when the overlay itself is not interested in it (for scrolling or otherwise).

smoogipoo:
I do consider scroll to be positional, and osu! is doing very weird stuff - something conventional in any input system I've seen. In some cases, I don't understand why overrides to `false` are done - for example, why is it done in `MatchSettingsOverlay`?

I'm not really happy with any of this, because it's just working around the fact that we don't have per-screen keybinding containers. Is that correct? Other than that, it doesn't allow rebinding Alt+Wheel anymore.

peppy:
> I do consider scroll to be positional, and osu! is doing very weird stuff - something unconventional in any input system I've seen. In some cases, I don't understand why overrides to `false` are done - for example, why is it done in `MatchSettingsOverlay`?

I applied it to every usage of `OverlayContainer`.

> I'm not really happy with any of this, because it's just working around the fact that we don't have per-screen keybinding containers. Is that correct? Other than that, it doesn't allow rebinding Alt+Wheel anymore.

I don't think that will help at all. This is the case where we have components explicitly handling `OnScroll`, *not* using `KeyBindingContainer`. Check `VolumeControlReceptor` for the use case here, and `Timeline` for the second:

https://github.com/ppy/osu/blob/8ff13845d1b911031fef877656337bba471a3b0f/osu.Game/Screens/Edit/Compose/Components/Timeline/Timeline.cs#L150-L157

Timeline wants to handle scroll when alt is pressed, *regardless* of any `GlobalAction`, but is being blocked because global actions are always handled first now (which is intentional - we want global actions to always have precedence).

smoogipoo (approved):

PR #4362, opened 2021-04-13
Update OpenTabletDriver package to 0.5.3

InfinityGhost:
## Short Changelog

[Full changelog](https://github.com/OpenTabletDriver/OpenTabletDriver/releases/tag/v0.5.3)

- Fixed Wacom touch reports being handled as auxiliary buttons (now reports touch data through `ITouchReport`)
  > Touch implementation is subject to change once we find more touch capable tablets and parse their data structures. 
- Added 26 new tablet configurations (Wacom, Huion, XP-Pen)
- Fixed and/or renamed 33 existing tablet configurations (Gaomon, Huion, Wacom, XP-Pen)
- Improved common exception messages to be more descriptive and link to OpenTabletDriver FAQ
- Workaround bad/partial reports resulting in device detection failing

### DIFF
--- a/osu.Framework/osu.Framework.csproj
+++ b/osu.Framework/osu.Framework.csproj
@@ -46,7 +46,7 @@
          See https://github.com/NuGet/Home/issues/4514 and https://github.com/dotnet/sdk/issues/765 . -->
     <PackageReference Include="ppy.osu.Framework.NativeLibs" Version="2021.115.0" />
     <PackageReference Include="Microsoft.Build.Locator" Version="1.4.1" />
-    <PackageReference Include="OpenTabletDriver" Version="0.5.2.3" />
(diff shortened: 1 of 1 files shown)


peppy:
Just going to mention that I still get a visible log on startup

```
[runtime] 2021-04-13 03:15:35 [debug]: Device: Set tablet feature: 02-02
[runtime] 2021-04-13 03:15:35 [verbose]: Detect: Found tablet 'Wacom CTH-470'
[runtime] 2021-04-13 03:15:35 [important]: Detect: Failed to find auxiliary device, express keys may be unavailable.
```

Not sure if there are actually important enough to appear to the end user, and I can already see the 470 wasn't updated in terms of support, but thought it may be worth mentioning off hand.

peppy (approved):
Seems to work as expected for my two devices at least.

PR #4363, opened 2021-04-13
Fix audio device freeing on Linux

smoogipoo:
Closes #3700, closes #3387

On Linux, freeing the 0 (no-sound) device can sometimes cause a deadlock internal to BASS. I previously tried to repro this in native code, but it seems to happen after quite a few `Init() -> Free()` cycles which I didn't realise was the case before.

I've reworked the whole init/freeing flow a bit here:
1. Init does not `Free()` if BASS gives an `Already` error. Instead, it just sets the current device.
2. When freeing, the target device is selected via `Bass.CurrentDevice`. This is required as per [BASS docs](http://www.un4seen.com/doc/#bass/BASS_Free.html), which was not done previously.
3. On disposal, `AudioManager` joins the sync thread, to prevent any `setDevice()` from potentially happening post-dispose/post-AudioThread-exit.
4. On Linux, the no-sound device is never freed.

This fixes endless test execution on Linux. Needs testing on Windows/macOS.

Test run on linux:
osu!: 
image
o!f:
image

### DIFF
--- a/osu.Framework.Tests/Audio/AudioComponentTest.cs
+++ b/osu.Framework.Tests/Audio/AudioComponentTest.cs
@@ -2,7 +2,6 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Reflection;
-using System.Threading;
 using System.Threading.Tasks;
 using NUnit.Framework;
 using osu.Framework.Audio;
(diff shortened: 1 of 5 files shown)


peppy:
No issues with tests on windows/mac, anything else I should be testing?

smoogipoo:
Only other thing is checking that you can re-select a previously initialised audio device.

peppy:
@smoogipoo are you testing that via osu!, or is there some way via framework that i'm missing?

smoogipoo:
Via osu!.

peppy:
Seems fine in Window and macOS.

peppy (approved):

PR #4367, opened 2021-04-14
Fix possible nullref when setting DecoupleableInterpolatingFramedClock.Rate

smoogipoo:
When the source clock is not an `IAdjustableClock`, setting `DecoupleableInterpolatingFramedClock.Rate` would nullref.

### DIFF
--- a/osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs
+++ b/osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs
@@ -54,7 +54,11 @@ namespace osu.Framework.Timing
         public override double Rate
         {
             get => Source?.Rate ?? 1;
-            set => adjustableSource.Rate = value;
+            set
+            {
+                if (adjustableSource != null)
+                    adjustableSource.Rate = value;
+            }
         }
 
         public void ResetSpeedAdjustments() => Rate = 1;
--- a/osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs
+++ b/osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs
@@ -1,6 +1,8 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+#nullable enable
+
 using System;
 
 namespace osu.Framework.Timing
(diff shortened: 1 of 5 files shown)


osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs
```diff
@@ -52,7 +54,11 @@ public class DecoupleableInterpolatingFramedClock : InterpolatingFramedClock, IA
         public override double Rate
         {
             get => Source?.Rate ?? 1;
-            set => adjustableSource.Rate = value;
+            set
+            {
+                if (adjustableSource != null)
```
peppy, line 59: Isn't this pretty bad though? Means the rate set is just lost.

peppy (changes requested):

osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs
smoogipoo: So you're thinking adjustableSource should be non-nullable then?

peppy: Either that or this should throw (or store the rate when the source is null). I guess it depends on how you hit the null case in the first place (since it seems like a new scenario) as to how we resolve this.

smoogipoo: As for why I hit it: I gave a decoupleable clock a `GameplayClock` as a non-adjustable source. But that code is all going to change from my side, so this change should be looked at in isolation.

I'll revert this change since it seems like a throw is more what was expected here... Though I'll use the throw of `InterpolatingFramedClock` instead otherwise it's currently a nullref:
https://github.com/ppy/osu-framework/blob/6b7de8e285e7b475d8ebb47ad3bb11730569c009/osu.Framework/Timing/InterpolatingFramedClock.cs#L54-L58

peppy: As in it should not support the rate set operation at all? Or changing the throw to an `ArgumentException` or something?

smoogipoo: Not support it at all for when the adjustable source is null. I think storing a local rate value would be weird considering it wouldn't do anything and be out of sync with the source's rate.

peppy: That's fine, but just not sure `NotSupportedException` is the correct exception to use (wasn't sure if it should be conditionally returned).  But seems the is within scope according to the MSDN documentation.

peppy (approved):

PR #4370, opened 2021-04-16
Remove IBindable interface inheritance

smoogipoo:
Without this, the following appears as a totally valid operation at compile time, but will completely fall over at run time:
```csharp
Bindable<string> bindable = new Bindable<string>();

MyClass c = new MyClass();
c.MyBindable.BindTo(bindable);

public class MyClass
{
    public readonly IBindable<int> MyBindable = new Bindable<int>();
}
```

Note that this is still possible if `MyBindable` is exposed as an `IBindable` instead.

I've also removed `IParseable` from `IBindable` as that allowed directly changing the value of the `IBindable*`, which is something that should only be allowed on `Bindable<T>`/`BindableList<T>`.

# Breaking Changes

## `IBindable`, `IBindable<T>`, and `IBindableList<>` are no longer `IParseable`

`IParseable` is only implemented on the `Bindable<>` and `BindableList<>` classes. Code that relied on this functionality should instead cast to `IParseable`.

### DIFF
--- a/osu.Framework.Tests/Bindables/BindableTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableTest.cs
@@ -96,10 +96,10 @@ namespace osu.Framework.Tests.Bindables
         [TestCaseSource(nameof(getParsingConversionTests))]
         public void TestParse(Type type, object input, object output)
         {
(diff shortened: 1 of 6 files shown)


peppy (approved):
Looks logical. I remember running into this before but didn't have a solution at hand.

PR #4372, opened 2021-04-17
Add descriptions to rotation directions

frenzibyte:
To make them look more user-friendly when used in dropdowns, like the existing rotation direction setting for barrel roll mod osu!-side.

I wasn't sure about lumping the framework with that at first, but since it's already done in other places and is just about adding description attributes, I think it's fine. 

---

# vNext

## `RotationDirection.CounterClockwise` has been renamed to `Counterclockwise`

As it appears to be often one word, the enum member `CounterClockwise` has been renamed to `Counterclockwise`.

### DIFF
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneTransformSequence.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneTransformSequence.cs
@@ -121,12 +121,12 @@ namespace osu.Framework.Tests.Visual.Drawables
         private void animate()
         {
             boxes[0].Delay(500).Then(500).Then(500).Then(
-                b => b.Delay(500).Spin(1000, RotationDirection.CounterClockwise)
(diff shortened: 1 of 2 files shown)


peppy:
Actually looks like "counterclockwise" is one word: https://www.merriam-webster.com/dictionary/counterclockwise

Not sure if we want to make this a breaking change (changing casing of the `enum`) but probably should if this isn't used in too many places.

frenzibyte:
> Actually looks like "counterclockwise" is one word: merriam-webster.com/dictionary/counterclockwise

TIL. 
> Not sure if we want to make this a breaking change (changing casing of the enum) but probably should if this isn't used in too many places.

Only has 3 usages in one test scene so I've applied the change.

peppy:
Looks good, add a breaking change to PR description please.

frenzibyte:
Done so.

peppy (approved):

Issue #4323, opened 2021-03-30
Tablet area offset when the built-in tablet driver is disabled


Thomzzz:
As discribed in the title I disabled the new built-in tablet driver, so I would expect it does not mess with the cursor.
I use the Hawku tablet driver and I set it to a smaller area and shifted it to the top left, so off center.
When I start osu the middle of the tablet is the middle of the screen (not correct).
If I reset the Hawku driver it is fine again and I can do every thing.
When I select a song it is still fine. Only after the song has been loaded and it starts playing the problem occurs again.
And it also remains after the song in the main menu.

(When the shifting problem is there I cant reach the bottom or the right border of the screen and on the left and top I can get the cursor out of view)

Update: If I disable the Hawku driver completely and enable the built-in osu driver things seem to work fine (except I have to match the settings from the Hawku driver).

frenzibyte:
Are you running in `Fullscreen` mode while encountering this issue?

bdach:
It sounds like you might need to disable "high precision mouse" if it's enabled, and you want to continue using the "hawku driver".

Thomzzz:
I tried it out just now. So with the hawku driver on and build-in driver disabled.
Setting the "confine mouse cursor to window" to "never" also fixed the problem.
I tested it with  both "high precision mouse" enabled and disabled, and there were no problems.

My Windows UI scaling is at 150%.
(I also run borderless instead of full screen and vertical and horizontal scale set to 64%)

peppy:
If you are willing to make the changes temporarily, can you try turning all scaling to 100% and see if this fixes your issue?

Thomzzz:
It was not entirely clear if you meant scaling Windows or the horizontal/vertical, so I tested both.
In all test these conditions were the same:
- Hawku driver was running
- Tablet: Disabled
- High precision mouse: Enabled
- Confine mouse cursor to window: During gameplay

I tried all 4 combinations of (vert/hor scaling at 100% and 64%) and (Windows at 100% and 150%)
In all situations the problem persisted. 
Very rarely at Windows 100% the problem did not occur, but I did not notice any reason why.
To be honest I'm not even certain if the problem rarely not occuring can be linked with Windows at 100%

I hope this was of any help.
I'm always open to any other testing with this issue

peppy:
[...]

PR #4377, opened 2021-04-19
Fix SDL2 window `Position` property not updated sometimes

frenzibyte:
Should resolve https://github.com/ppy/osu-framework/issues/4323, and resolve https://github.com/ppy/osu/issues/12156.

Not much to say, `WindowsMouseHandler` uses `Window.Position` when calculating the mouse position for absolute input devices, and since `Window.Position` may not be updated sometimes (, the handler miscalculates the cursor position and causes it to be locked in a certain area or potentially completely freeze and not able to move.

Here's a video of reproducing the bug above:

https://user-images.githubusercontent.com/22781491/115179660-7f10f880-a0dc-11eb-9bfb-e50b859f5c11.mp4

There were two I cases I've found where the position property doesn't get updated, and are the only ones apparently:
1. when switching to other window modes https://github.com/ppy/osu-framework/commit/7d59db8b3bbfb62aed25196ae34d3289d0c6885e
2. during the borderless switch in `WindowsWindow` https://github.com/ppy/osu-framework/commit/b54df51d1a95a24e8207638ef7c1d04000b972c3

I'm not entirely sure why the position was first only updated on windowed mode, but I don't see any absolute harm in updating it on borderless and fullscreen modes, and logically makes sense to do so.

### DIFF
--- a/osu.Framework.Tests/Visual/Platform/TestSceneBorderless.cs
+++ b/osu.Framework.Tests/Visual/Platform/TestSceneBorderless.cs
@@ -164,6 +164,7 @@ namespace osu.Framework.Tests.Visual.Platform
 
                 // borderless alignment tests
                 AddStep("switch to borderless", () => windowMode.Value = WindowMode.Borderless);
+                AddAssert("check window position", () => new Point(window.Position.X + 1, window.Position.Y + 1) == display.Bounds.Location);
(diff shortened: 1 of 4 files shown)


peppy:
Which platforms have you tested this on so far?

frenzibyte:
Windows and macOS.

peppy (approved):

Issue #4378, opened 2021-04-19
Cursor position is updated outside of unfocused window


InfinityGhost:
Cursor position is updated outside of window when it is out of focus, causing inputs to escape and alter the cursor position on the desktop.

Can be replicated with a tablet while the TabletHandler is enabled, causes the cursor to be mapped to where the window is positioned while out of focus.

https://user-images.githubusercontent.com/39218853/115184048-05ace200-a0ab-11eb-8cfc-d89eb1ff843c.mp4

PR #4379, opened 2021-04-19
Only forward feedback cursor updates when the framework window is active

peppy:
Resolves https://github.com/ppy/osu-framework/issues/4378.

Tested on windows with wacom tablet.


smoogipoo (dismissed):

frenzibyte (changes requested):
Is this intentional? once the cursor is outside the window the mouse will be completely separated from tablet input, whereas I would expect the mouse to be synced back when the tablet input is back in window.

https://user-images.githubusercontent.com/22781491/115210074-e93e9300-a106-11eb-89a7-1da6a8cbcff0.mp4

peppy:
It was intentional yes. I'm open to discussion on expected behaviour. If that isn't what we want then only the window active check is required.

peppy:
Have switched it to always gobble input when focused.

smoogipoo (approved):

PR #4382, opened 2021-04-21
Speed up dynamic compilation type resolution

smoogipoo:
1. Adding a HashSet to prevent calling `GetSymbolInfo()` on the same syntax node seen in multiple locations.
    * This is currently only per-file to not run into failed recompilations. Additional perf improvements are possible by using the hashset globally, but the implementation becomes complex so I've added a todo for now.
2. Caching various `.ToString()` invocations.
3. Ignoring variable declarators - the name of a variable.
4. Ignoring variable name of assignment expressions, which is particularly important with our use of object initialisers.

Tested osu!-side by changing `EffectRowAttribute` in `TestSceneTimingScreen`:
Before:
image
After:
image

~50% speedup

### DIFF
--- a/osu.Framework/Testing/RoslynTypeReferenceBuilder.cs
+++ b/osu.Framework/Testing/RoslynTypeReferenceBuilder.cs
@@ -189,8 +189,6 @@ namespace osu.Framework.Testing
         /// <param name="rootReference">The root, where the map should start being build from.</param>
         private async Task buildReferenceMapRecursiveAsync(TypeReference rootReference)
         {
-            var seenSyntaxes = new HashSet<string>();
-
             var searchQueue = new Queue<TypeReference>();
             searchQueue.Enqueue(rootReference);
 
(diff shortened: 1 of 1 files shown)


osu.Framework/Testing/RoslynTypeReferenceBuilder.cs
```diff
@@ -189,6 +189,8 @@ private async Task buildReferenceMapAsync(Type testType, string changedFile)
         /// <param name="rootReference">The root, where the map should start being build from.</param>
         private async Task buildReferenceMapRecursiveAsync(TypeReference rootReference)
         {
+            var seenSyntaxes = new HashSet<string>();
```
peppy, line 192: left over i guess?

smoogipoo: Yeah I'll grab it but I've got other changes in progress atm.

smoogipoo:
In the previous code, I was marking seen syntaxes a little too eagerly, but upon fixing that I lost around 50% of the gains. So I had to add a few more things in to get a relatively close value to what my screenshots show (9.4sec vs 8.7sec in screenshot).

I think these are simple enough while I can still ensure correctness - I also looked into member accesses which further reduces the time taken to about 8sec, but that's a little bit harder to ensure correctness with/not possible.

peppy:
Will test this over the course of the day and merge if I don't run into any regression scenarios.

peppy (approved):

PR #4391, opened 2021-04-22
Multithreaded dynamic compilation type resolver

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/4390


peppy:
Intentionally blocked here?

smoogipoo:
Wanted to go through and make sure there wasn't any thread races, since I already solved a few. I think it should be fine.

peppy (approved):
one way to test

PR #4395, opened 2021-04-23
Fix TrackBass not updating current time after seek

smoogipoo:
Fixes https://github.com/ppy/osu/issues/12541

When seeks are performed in-line, there's a bit of an expectation that the resultant time is updated immediately rather than in the next frame.

This is also an issue with Play() and Stop(), but I'll be making an issue for those to be tackled later - they're a bit involved. (https://github.com/ppy/osu-framework/issues/4396)


peppy:
I have this really strange hunch that we tried this before and inadvertently broken something which was relying on the current behaviour... but I can't seem to find a paper trail so it may have been a different scenario.

Going to get this in and wait for any potential regressions.

peppy (approved):

PR #4397, opened 2021-04-23
Fix assertion tripping when seeking from within the mixtime callback

bdach:
The recent changes to `TrackBass` (#4395) added an instantaneous current time value update on a seek. As it turns out, it is not necessarily safe to unconditionally run that update, as BASS callbacks (such as the mixtime callback) could run from a separate unmanaged thread rather than the audio thread, tripping the `CanPerformInline` assertion check if a track had `Looping` and `RestartPoint` set.

This can be reproduced game-side with a framework checkout by seeking to the end of a track in song select and waiting for it to complete.

2021-04-23-213212_1743x861_scrot

To fix, add an explicit check for whether the current time update can be safely performed via `seekInternal()` before trying to attempt it.

---

I wanted to write a test case for this, but as it turns out `TrackBassTest` magically sidesteps this issue, because it calls `track.Update()`, which seems to change semantics somehow in that the callback actually executes on the audio thread rather than a BASS-owned one. Not sure how that's possible, but [apparently this is something that we're tangentially aware of](https://discord.com/channels/188630481301012481/188630652340404224/835230809261998121).

Not sure about doubling up the condition, but I kind of wanted to keep the assertion as it was where it was in `updateCurrentTime()`, to let future travelers know when the method is safe to call. That assertion is dead on release builds anyway. I'll move the guard inside `updateCurrentTime()` in place of the assertion on request.

### DIFF
--- a/osu.Framework/Audio/Track/TrackBass.cs
+++ b/osu.Framework/Audio/Track/TrackBass.cs
@@ -342,7 +342,11 @@ namespace osu.Framework.Audio.Track
             if (pos != Bass.ChannelGetPosition(activeStream))
                 Bass.ChannelSetPosition(activeStream, pos);
 
-            updateCurrentTime();
+            // current time updates are safe to perform from enqueued actions,
+            // but not always safe to perform from BASS callbacks, since those can sometimes use a separate thread.
+            // if it's not safe to update immediately here, the next UpdateState() call is guaranteed to update the time safely anyway.
+            if (CanPerformInline)
+                updateCurrentTime();
         }
 
         private void updateCurrentTime()


frenzibyte (approved):
Sounds reasonable

peppy (approved):

osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -342,7 +342,11 @@ private void seekInternal(double seek)
             if (pos != Bass.ChannelGetPosition(activeStream))
                 Bass.ChannelSetPosition(activeStream, pos);
 
-            updateCurrentTime();
+            // current time updates are safe to perform from enqueued actions,
+            // but not always safe to perform from BASS callbacks, since those can sometimes use a separate thread.
+            // if it's not safe to update immediately here, the next UpdateState() call is guaranteed to update the time safely anyway.
+            if (CanPerformInline)
+                updateCurrentTime();
```
smoogipoo, line 349: Actually on thinking about this again, what about the rest of this code? This is all doing bass operations as the updateCurrentTime method. Feels weird that one is conditional running on the correct thread and the other isn’t, no?

frenzibyte: The way I think about this is that the above code is performing BASS-specific logic and mutate certain states (like `isPlayed` and `hasCompleted`) for the next `UpdateState()` call to update its own public ones on the game's audio thread.

But I'm not really sure anymore which is considered "safe" and "unsafe", `hasCompleted` is exposed publicly and gets mutated by BASS threads, while `currentTime` would cause an assertion hit if not on game's audio thread.

It all sounds like either:
1. BASS native threads should never touch any members in the track instance.
    - which *may* not be good for current BASS mixtime logic.
2. Allow BASS native threads to mutate any members and bypass such assertions.
    - and potentially remove them.

bdach: I guess maybe the more correct thing to do here would be to enqueue the seek onto the audio thread from the mixtime callback? That said, that could cause unexpected effects for looping tracks. Definitely feels like the actual BASS seek should occur as close as possible to the mixtime callback...

smoogipoo: Most correct would be to schedule the entire thing. Whether an assumption can be made that BASS is perfectly happy to run BASS invocations asynchronously as it has done until now, is the question.

I'm tending towards removing the assertion and the precondition altogether, and not scheduling back for the time being. After searching the forums, I found at least one comment that mentions BASS should be thread-safe: http://www.un4seen.com/forum/?topic=17233.msg120790#msg120790

bdach: While the time update uses `Interlocked` and therefore I'd hope serialising/thread safe, an argument for keeping the assertions/guards is that current time of the track shouldn't really be changing during a frame if audio processing is taking place. Not sure how good of an argument it is, though.

smoogipoo: Ah, you're right about that. I also tend to agree that current time shouldn't change without the audio thread updating. Let's keep as is and revisit if it comes up as an issue.

PR #4399, opened 2021-04-24
Remove Borderless 1px position offset hack

Rubydesic:
It makes the osu window bleed into other monitors in Borderless mode. On Windows, if the other monitor is a lower refresh rate than the main one, the entire osu window is displayed at the lower refresh rate.

I'll admit, I don't really understand the reason this apparently deliberate offset was added in the first place, so this may introduce some sort of regression.

See also: https://github.com/ppy/osu/issues/12535

### DIFF
--- a/osu.Framework/Platform/Windows/WindowsWindow.cs
+++ b/osu.Framework/Platform/Windows/WindowsWindow.cs
@@ -38,14 +38,10 @@ namespace osu.Framework.Platform.Windows
         {
             SDL.SDL_SetWindowBordered(SDLWindowHandle, SDL.SDL_bool.SDL_FALSE);
 
-            Size positionOffsetHack = new Size(1, 1);
+            var newSize = CurrentDisplay.Bounds.Size;
 
-            var newSize = CurrentDisplay.Bounds.Size + positionOffsetHack;
-            var newPosition = CurrentDisplay.Bounds.Location - positionOffsetHack;
-
-            // for now let's use the same 1px hack that we've always used to force borderless.
             SDL.SDL_SetWindowSize(SDLWindowHandle, newSize.Width, newSize.Height);
-            Position = newPosition;
+            Position = CurrentDisplay.Bounds.Location;
 
             return newSize;
         }


smoogipoo:
Yes, you're correct that this will regress. This was done because full-screen borderless on some graphics cards is actually exclusive mode. A rudimentary blame on the code would have given you the reason for this code existing: https://github.com/ppy/osu-framework/pull/4080

However, I believe that stable doesn't maintain the _positional_ offset, which is probably why it's giving a different result. It uses [MonitorFromWindow](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-monitorfromwindow) which, if the window is located at (0,0), results in the correct refresh rate being chosen.
To be more precise, I _believe_ that stable applies the positional offset to exit borderless, but then reverts it to (0,0) and applies the size offset to maintain borderless.

peppy:
@Rubydesic are you willing to actually do proper testing with this and find out what is required to make this change correct? If not, let's close this PR for now as it's definitely not the correct direction.

Rubydesic:
@smoogipoo Thanks for the link, a 'rudimentary blame' does not surface this PR because it affected a different file. (Specifically, it surfaces cb763fc16277714f2b4d8195aa411a33eb007cd7 which provides no explanation).

@peppy I can't reproduce the problem and the linked issue(s) make no mention of specifically what configurations this breaks, so close this PR, but I would argue that the game window having the correct dimensions is the right direction - what needs to be figured out is how to make the graphics driver behave, not the other way around.

As a workaround, perhaps there can be a startup flag or hidden config option for this behaviour as it makes Borderless pretty much unuseable for people with multiple monitors.

frenzibyte:
> a 'rudimentary blame' does not surface this PR because it affected a different file. (Specifically, it surfaces cb763fc which provides no explanation)

CleanShot 2021-04-24 at 19 34 39@2x

peppy:
Consider that merging this would make borderless unusable for *many more users*. To the point where it will give the same behaviour as fullscreen.

Also, it does not break for all multi-monitor users in the current state (as I mentioned in the issue thread, it depends on layout).

Will close this for now. It's a medium priority to get this working correctly, so I'm sure we will get to fixing your issue in the near future. You can compile yourself for the time being if it is important to you to have this working.

PR #4401, opened 2021-04-25
Change dragging states before triggering end event

frenzibyte:
Matches behaviour as hover lose and focus lose, where the property would have its state changed before triggering the event.

Have checked for any usages of this property, none of them needs any action.

# vNext

## Both `Drawable.IsDragged` and `InputManager.DraggedDrawable` now change their state before triggering event

Both `Drawable.IsDragged` and `InputManager.DraggedDrawable` now change their state before the `DragEndEvent` is triggered, rather than after, matching other similar events like `FocusLostEvent` and `HoverLostEvent`.

### DIFF
--- a/osu.Framework/Input/MouseButtonEventManager.cs
+++ b/osu.Framework/Input/MouseButtonEventManager.cs
@@ -206,12 +206,15 @@ namespace osu.Framework.Input
         {
             DragStarted = false;
 
-            if (DraggedDrawable == null) return;
-
-            DraggedDrawable.IsDragged = false;
-            PropagateButtonEvent(new[] { DraggedDrawable }, new DragEndEvent(state, Button, MouseDownPosition));
+            var previousDragged = DraggedDrawable;
 
             DraggedDrawable = null;
+
+            if (previousDragged != null)
+            {
+                previousDragged.IsDragged = false;
+                PropagateButtonEvent(new[] { previousDragged }, new DragEndEvent(state, Button, MouseDownPosition));
+            }
         }
     }
 }
--- a/osu.Framework/Input/MouseButtonEventManager.cs
+++ b/osu.Framework/Input/MouseButtonEventManager.cs
@@ -208,9 +208,9 @@ namespace osu.Framework.Input
 
             if (DraggedDrawable == null) return;
 
+            DraggedDrawable.IsDragged = false;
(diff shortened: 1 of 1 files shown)


osu.Framework/Input/MouseButtonEventManager.cs
```diff
@@ -208,9 +208,9 @@ private void handleDragEnd(InputState state)
 
             if (DraggedDrawable == null) return;
 
+            DraggedDrawable.IsDragged = false;
             PropagateButtonEvent(new[] { DraggedDrawable }, new DragEndEvent(state, Button, MouseDownPosition));
 
-            DraggedDrawable.IsDragged = false;
             DraggedDrawable = null;
```
bdach, line 214: This should also be unset before the event is propagated. Compare focus, which stores the previous `FocusedDrawable` to a local and only then triggers the event. Hover does similar.

This should also be included in the breaking changes note.

bdach (changes requested):

osu.Framework/Input/MouseButtonEventManager.cs
frenzibyte: Ah, missed to notice that.

osu.Framework/Input/MouseButtonEventManager.cs
```diff
@@ -206,12 +206,15 @@ private void handleDragEnd(InputState state)
         {
             DragStarted = false;
 
-            if (DraggedDrawable == null) return;
```
bdach, line 209: not sure why this guard went away. something like the following reads much better and has no nesting/null checks:

```csharp
if (DraggedDrawable == null) return;

var previousDragged = DraggedDrawable;
previousDragged.IsDragged = false;
DraggedDrawable = null;

PropagateButtonEvent(new[] { DraggedDrawable }, new DragEndEvent(state, Button, MouseDownPosition));
```

but i guess this variant also works...

frenzibyte: Went with the same way focus lose logic does it, felt good to me at first, but I think yours sound much better.

bdach (approved):

smoogipoo (approved):
Makes sense

peppy (approved):

PR #4402, opened 2021-04-25
Change default value of `CursorInWindow` to false

### LINKED ISSUE #4400: `cursorInWindow` is wrongfully assumed to be true on startup
https://github.com/ppy/osu-framework/blob/b8e905548bdf517af5bb9790c0bc0b1b8f0205b9/osu.Framework/Platform/SDL2DesktopWindow.cs#L170

This assumption causes `cursorInWindow` not to match the actual cursor state when the mouse is outside the window on startup.

Steps to repro:
1. Open `visual-tests` > `Platform` > `ActiveState`
2. Set window mode to windowed
3. Close visual-tests
4. Open visual-tests, but keep your cursor outside the window
5. `host.Window.CursorInWindow` doesn't match the cursor state (being outside of the window)

In this state, the framework (as seen in `visual-tests` > `Input` > `InputManager`) does not see mouse move events until the cursor enters the window. With the default being false, this works as expected.

I've tracked the change of default to true to https://github.com/ppy/osu-framework/commit/e9d59b6af012a1ff83f588b5d4f31afc6c7a35d0.

Susko3:
Closes #4400.

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -167,7 +167,7 @@ namespace osu.Framework.Platform
 
         public IBindable<bool> IsActive => isActive;
 
-        private readonly BindableBool cursorInWindow = new BindableBool(true);
+        private readonly BindableBool cursorInWindow = new BindableBool();
 
         public IBindable<bool> CursorInWindow => cursorInWindow;
 


frenzibyte (approved):
Looks fine on Windows and macOS.

bdach:
While this seems valid from a quick check on linux, one more concern: `OsuTKWindow` actually specifies the default to be `true`, too, but that one is actually valid because we only use osuTK on mobile now, and there is no way to enter or leave the game window on mobile. May be wise to leave a comment behind with this change to explain the discrepancy.

peppy:
@bdach i think this is okay actually. generally `OsuTK` is considered a legacy class, and I don't think this is something we'll miss when (eventually) reimplementing mobile support against a different API.

`false` as a default sounds more sane to me.

peppy (approved):

PR #4404, opened 2021-04-26
Abstractify bindable number constraints logic and use for `BindableSize`

### LINKED ISSUE #4307: BindableSize does not propagate `MinValue`/`MaxValue` changes to bindings

frenzibyte:
Resolves #4307 

Rather than copying logic from `BindableNumber` over to `BindableSize`, I've abstracted the constraints logic into a `ConstrainedBindable` and let `BindableSize` inherit it and override the required properties.

For test cases, I've moved the constraints test cases from `BindableNumberTest` to a `ConstrainedBindableTest` test class and currently instantiate `BindableNumber`s as a sample for testing constraints propagation, not sure if also instantiating and testing a `BindableSize` is worth it, if one breaks on that aspect the other should also break.

Also made sure to document that `BindableSize` constraints are component-wise, to avoid any confusion.


osu.Framework/Bindables/BindableSize.cs
```diff
@@ -51,10 +51,13 @@ protected override Size ClampValue(Size value, Size minValue, Size maxValue)
 
         protected override int Compare(Size x, Size y)
         {
-            if (x.Width == y.Width && x.Height == y.Height)
-                return 0;
+            if (x.Width < y.Width || x.Height < y.Height)
+                return -1;
```
ekrctb, line 55: I don't think the new code is right either. For example, it is both `(0, 1) < (1, 0)` and `(1, 0) < (0, 1)`.

Can't we just use the lexicographic ordering?
```csharp
if (x.Width != y.Width) return x.Width.CompareTo(y.Width);
return x.Height.CompareTo(y.Height);
```

frenzibyte: hmm, this would allow `MinValue` to be `(0, 1)` and `MaxValue` to be `(1, 0)`, which is not valid.. not sure how to handle these cases, maybe override and throw undefined exception on that case instead?

ekrctb: I missed that the comparison is for comparison between min/max values.
As the ordering required for the size clamping is not a total ordering, but only a partial ordering, `Compare` cannot always return one of the `-1, 0, 1` values for each comparison, but it has to return a separate "unordered" case.
Perhaps the return type of `Compare` should just be `bool`, and return `true` if LHS is less than or equals to RHS. If the ordering between LHS and RHS is undefined (unordered), it returns `false`.
For `Size`, it is `Compare(x, y) => x.Width <= y.Width && x.Height <= y.Height`.

osu.Framework/Bindables/ConstrainedBindable.cs
```diff
@@ -0,0 +1,193 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+
+namespace osu.Framework.Bindables
+{
+    public abstract class ConstrainedBindable<T> : Bindable<T>
```
ekrctb, line 9: As this class implement a special kind of constraint (clamping), more specific can be used, like `RangeConstrainedBindable`?

osu.Framework/Bindables/BindableSize.cs
bdach: `BindableSize` should not implement `IComparable`, period. There is no sane way to declare a full ordering on it as stated above. Where is the requirement of implementing it coming from?

frenzibyte: doesn't implement `IComparable`, simply using the method for comparing `MinValue` and `MaxValue` to avoid non-overlapping ranges.

bdach: then this abstraction is wrong. declare a bound-checking method that returns a `bool` in `ConstrainedBindable` instead of that `CompareTo()`.

frenzibyte: hmm, you have a point. as it isn't used anywhere else.

bdach:
@frenzibyte Tests are failing and there are unresolved reviews. Is this going to get fixed?

frenzibyte:
Yeah, I'll get to it when I have the chance to.

osu.Framework.Tests/Bindables/ConstrainedBindableTest.cs
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using NUnit.Framework;
+using osu.Framework.Bindables;
+
+namespace osu.Framework.Tests.Bindables
+{
+    [TestFixture]
+    public class ConstrainedBindableTest
```
bdach, line 10: test class name does not match tested class's name anymore (should be `RangeConstrainedBindableTest`)

osu.Framework/Bindables/RangeConstrainedBindable.cs
```diff
@@ -0,0 +1,192 @@
...
+        /// Clamps the given <paramref name="value"/>.
+        /// </summary>
+        protected abstract T ClampValue(T value, T minValue, T maxValue);
+
+        /// <summary>
+        /// Whether a given range has a minimum value less than or equal to its maximum value (i.e. overlapping range).
+        /// </summary>
+        /// <param name="min">The range's minimum value.</param>
+        /// <param name="max">The range's maximum value.</param>
+        protected abstract bool IsOverlappingRange(T min, T max);
```
bdach, line 190: would rename to `IsValidRange` or something along these lines ("overlapping range" sounds like it's *not* a correct range).

osu.Framework/Bindables/BindableNumber.cs
```diff
@@ -446,6 +313,10 @@ public override bool IsDefault
             }
         }
 
+        protected override T ClampValue(T value, T minValue, T maxValue) => max(minValue, min(maxValue, value));
+
+        protected override bool IsOverlappingRange(T min, T max) => min.CompareTo(max) <= 0;
```
bdach, line 318: would probably seal these overrides

osu.Framework/Bindables/RangeConstrainedBindable.cs
```diff
@@ -0,0 +1,192 @@
...
+        /// <param name="source">The bindable that triggered this. A null value represents the current bindable instance.</param>
+        internal void SetMinValue(T minValue, bool updateCurrentValue, RangeConstrainedBindable<T> source)
+        {
+            this.minValue = minValue;
+            TriggerMinValueChange(source);
+
+            if (updateCurrentValue)
+            {
+                // Reapply the current value to respect the new minimum value.
+                Value = base.Value;
```
bdach, line 89: stuff like `Value = base.Value;` doing business logic and having silent side effects makes me extremely scared - preferred the previous `SetValue(Value)` thing `BindableNumber` had

osu.Framework/Bindables/RangeConstrainedBindable.cs
```diff
@@ -0,0 +1,192 @@
...
+            if (them is RangeConstrainedBindable<T> other)
+            {
+                MinValue = other.MinValue;
+                MaxValue = other.MaxValue;
+
+                if (!IsOverlappingRange(MinValue, MaxValue))
+                {
+                    throw new ArgumentOutOfRangeException(
+                        nameof(them), $"Can not weld bindable longs with non-overlapping min/max-ranges. The ranges were [{MinValue} - {MaxValue}] and [{other.MinValue} - {other.MaxValue}].");
+                }
```
bdach, line 170: this exception message direly needs updating ("weld" and not "bind"? bindable "longs"?).

as an aside it's interesting that this is only checked *after* already having transferred the value, not before - would probably check before, to avoid leaving this bindable in a potentially corrupted state.

bdach (changes requested):

osu.Framework/Bindables/RangeConstrainedBindable.cs
frenzibyte: yeah, thought it was just me not understanding the term `overlapping` properly.

osu.Framework/Bindables/BindableNumber.cs
frenzibyte: and at `BindableSize` too? or is there a special reason into this one?

osu.Framework/Bindables/RangeConstrainedBindable.cs
frenzibyte: right, the way I have it looks really bad and I haven't even noticed.

osu.Framework/Bindables/RangeConstrainedBindable.cs
frenzibyte: right, the exception message itself is broken since both `MinValue` and `other.MinValue` are gonna be the same if thrown after, will update.

osu.Framework/Bindables/BindableNumber.cs
bdach: I suppose you could do both. This one is a *bit* special because of the generics, but neither should hurt really.

osu.Framework/Bindables/RangeConstrainedBindable.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        public override void BindTo(Bindable<T> them)
+        {
+            if (them is RangeConstrainedBindable<T> other)
+            {
+                if (!IsValidRange(other.MinValue, other.MaxValue))
+                {
+                    throw new ArgumentOutOfRangeException(
+                        nameof(them), $"The target bindable has specified an invalid range of [{other.MinValue} - {other.MaxValue}].");
+                }
```
bdach, line 167: note that sets to `{Min,Max}Value` should really be checking range validity as well, but that is not a regression as the old code also didn't do that, so i'm purposefully looking away from that issue to not enlarge the diff any further

bdach (approved):
i can live with this now

peppy (approved):

PR #4405, opened 2021-04-26
Expose ThreadSafety

smoogipoo:
Would come in useful with components like https://github.com/ppy/osu/pull/11461, which have components that must be run on a specific game thread.

`ExecutionMode` is not exposed as:
1. It's `GameHost`-specific. It should probably be moved out of `ThreadSafety` or be made thread-static.
2. There isn't much reason for the execution mode to be known.

`EnsureInputThread()` + `EnsureAudioThread()` added for conformity.

### DIFF
--- a/osu.Framework/Development/ThreadSafety.cs
+++ b/osu.Framework/Development/ThreadSafety.cs
@@ -13,6 +13,12 @@ namespace osu.Framework.Development
     public static class ThreadSafety
     {
         /// <summary>
+        /// Asserts that the current code is executing on the input thread. Debug only.
+        /// </summary>
+        [Conditional("DEBUG")]
+        public static void EnsureInputThread() => Debug.Assert(IsInputThread);
+
+        /// <summary>
         /// Asserts that the current code is executing on the update thread. Debug only.
         /// </summary>
         [Conditional("DEBUG")]
--- a/osu.Framework/Development/ThreadSafety.cs
+++ b/osu.Framework/Development/ThreadSafety.cs
@@ -7,17 +7,67 @@ using osu.Framework.Platform;
 
 namespace osu.Framework.Development
 {
-    internal static class ThreadSafety
+    /// <summary>
+    /// Utilities to ensure correct thread usage throughout a game.
+    /// </summary>
+    public static class ThreadSafety
     {
+        /// <summary>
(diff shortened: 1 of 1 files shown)


osu.Framework/Development/ThreadSafety.cs
```diff
@@ -7,37 +7,73 @@
...
+    /// Utilities to ensure correct thread usage throughout a game.
+    /// </summary>
+    public static class ThreadSafety
     {
+        /// <summary>
+        /// Asserts that the current code is executing on the update thread. Debug only.
+        /// </summary>
         [Conditional("DEBUG")]
-        internal static void EnsureUpdateThread() => Debug.Assert(IsUpdateThread);
+        public static void EnsureUpdateThread() => Debug.Assert(IsUpdateThread);
```
peppy, line 25: Does this actually work with the `Conditional` attribute? Or require a local/debug framework build?

If it works, I'd probably reword the xmldoc to mention explicitly "Only asserts in debug builds due to performance concerns" or something along those lines.

smoogipoo: Good point, these actually don't work unless the framework is debug... Will `internal` them for now.

bdach: Do they really not? [Docs say](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.conditionalattribute?redirectedfrom=MSDN&view=net-5.0):

> Applying ConditionalAttribute to a method indicates to compilers that a call to the method should not be compiled into Microsoft intermediate language (MSIL) unless the conditional compilation symbol that is associated with ConditionalAttribute is defined.

That would imply that the attribute is checked at the call site rather than that the method is NOP'd out, but I haven't actually checked if it works like that across nuget packages...

smoogipoo: MSIL implies project compilation. Doesn't work across nuget packages or even referencing a release-mode project from a debug one.

bdach: Yeah, but what I'm getting at is that if I understand correctly, the body of the method will still make it to IL, but none of the calls to it will...

Ran a quick sharplab and that seemed to check out.

peppy (approved):

PR #4408, opened 2021-04-27
Add headless test runner for templates tests projects

frenzibyte:
Must be implemented for providing the correct dependencies for the test scene to run properly, 

I also have another PR almost ready for making `TestScene.CreateTestSceneTestRunner()` abstract as it's very recommended to be implemented per game tests project, rather than using the default `TestSceneTestRunner`.

### DIFF
--- a/osu.Framework.Templates/templates/template-empty/TemplateGame.Game.Tests/Visual/TemplateGameTestScene.cs
+++ b/osu.Framework.Templates/templates/template-empty/TemplateGame.Game.Tests/Visual/TemplateGameTestScene.cs
@@ -1,0 +1,22 @@
+using osu.Framework.Testing;
+
+namespace TemplateGame.Game.Tests.Visual
+{
+    public class TemplateGameTestScene : TestScene
+    {
+        protected override ITestSceneTestRunner CreateRunner() => new TemplateGameTestSceneTestRunner();
+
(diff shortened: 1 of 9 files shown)


osu.Framework.Templates/templates/template-empty/TemplateGame.Game.Tests/Visual/TemplateGameTestScene.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        private class TemplateGameTestSceneTestRunner : TemplateGameGameBase, ITestSceneTestRunner
+        {
+            private TestSceneTestRunner.TestRunner runner;
+
+            protected override void LoadAsyncComplete()
+            {
+                base.LoadAsyncComplete();
+                Add(runner = new TestSceneTestRunner.TestRunner());
+            }
```
frenzibyte, line 17: Currently doing this as it's done osu!-side:
CleanShot 2021-04-27 at 04 24 52

But I don't really understand why it's done in `LoadComplete` and `LoadAsyncComplete`, rather than at BDL or even constructor, since the test scene won't be loaded and added to `TestRunner` until the game is fully loaded:
https://github.com/ppy/osu-framework/blob/40162a1832bdaff32c853e03cbdb403ca885eff9/osu.Framework/Testing/TestScene.cs#L42-L58

smoogipoo (approved):
lgtm

PR #4409, opened 2021-04-27
Add ConfigureAwait(false) to DCC

smoogipoo:
Not sure how I missed these, guess rider didn't show them for some reason.


peppy:
99% sure it won't trigger on private methods, since the whole `ConfigureAwait` pattern is there to guard against consumers of an API that may not necessarily be following expectations. Not sure it's required here yet, haven't reviewed, but something to consider.

smoogipoo:
It does trigger rider _now_ for some reason, underlining the entire massive block >_>. It triggered as a compiler warning if you use local-framework, but it was too late by that point.

peppy:
Is the rule setup differently/incorrectly on the osu-framework project then?

smoogipoo:
Doesn't seem to be. Looks like the compiler warning is there in o!f, so discard what I said above:
image

peppy (approved):

PR #4410, opened 2021-04-27
Update tooling + fix inspections

smoogipoo:
Blocking for now pending inspection fixes.

It's come to my attention that dotnet-format hasn't been working since we switched to .NET 5. It's quite a useful tool because inspections happen very quickly (unlike inspectcode). In the process, this also requires using imports to be sorted, so long as `dotnet_sort_system_directives_first = true` is turned on.
It seems like Rider will sort usings when added, or via the sledgehammer option in the menu (Code -> Optimize Imports), but if this is undesired then we can just remove the .editorconfig option.

* Updated `dotnet-format` and `jetbrains.resharper.globaltools` except cake (we're far behind but it's about to be deprecated anyway).
* Sorted using directives to appease `dotnet-format`.
* Fixed inspectcode not running on .NET 5 TFM.
* Fixed nullref inspections that have been missed.
* Changed some new inspections to hints - I think for now we still prefer using `.Slice(start)` instead of `[start..]`, and no complex pattern matching.

I suggest going through the PR per-commit.


smoogipoo:
Holding back until the migration to gh-actions.

PR #4411, opened 2021-04-27
Simplify LifetimeEntry logic, and make SetLifetime overridable

ekrctb:
`RequestLifetimeUpdate` logic is changed so `SetLifetime` can be safely called from non-internal user and overridden.
For motivation, see osu-side PR: ppy/osu#12592.

Note that I'm not just making `LifetimeStart` and `LifetimeEnd` virtual. This is because those properties are critical to the correctness of the `LifetimeManager`.

### DIFF
--- a/osu.Framework/Graphics/Performance/LifetimeEntry.cs
+++ b/osu.Framework/Graphics/Performance/LifetimeEntry.cs
@@ -21,13 +21,8 @@ namespace osu.Framework.Graphics.Performance
         public double LifetimeStart
         {
             get => lifetimeStart;
-            set
-            {
-                if (lifetimeStart == value)
-                    return;
-
-                SetLifetime(value, lifetimeEnd);
-            }
+            // A method is used as C# doesn't allow the combination of a non-virtual getter and a virtual setter.
+            set => SetLifetimeStart(value);
         }
 
         private double lifetimeEnd = double.MaxValue;
@@ -38,13 +33,7 @@ namespace osu.Framework.Graphics.Performance
         public double LifetimeEnd
         {
             get => lifetimeEnd;
-            set
-            {
-                if (lifetimeEnd == value)
-                    return;
-
-                SetLifetime(lifetimeStart, value);
-            }
+            set => SetLifetimeEnd(value);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


smoogipoo (dismissed):
Lgtm, just awaiting response on osu!-side review.

osu.Framework/Graphics/Performance/LifetimeEntry.cs
```diff
@@ -38,26 +33,38 @@ public double LifetimeStart
...
         /// </summary>
         internal event Action<LifetimeEntry> RequestLifetimeUpdate;
 
+        /// <summary>
+        /// Update <see cref="LifetimeStart"/> of this <see cref="LifetimeEntry"/>.
+        /// </summary>
+        protected virtual void SetLifetimeStart(double start)
+        {
+            if (start != lifetimeStart)
+                SetLifetime(start, lifetimeEnd);
```
smoogipoo, line 50: It has to be the other way around to not introduce side effects. This way doesn't have any difference compared to before.
`SetLifetime()` can call these methods, but these can only handle the single endpoint.

It does mean that `RequestLifetimeUpdate` is invoked multiple times, but that's an `O(log n )` tradeoff that I think is fine.

smoogipoo (changes requested):

osu.Framework/Graphics/Performance/LifetimeEntry.cs
ekrctb: I don't understand. My change (with my osu side change) does certainly fixed the previous issue (I added a test).
`SetLifetime` is now protected, and that is critical to tracking real lifetime.

smoogipoo: Ah, I see. In my mind I also had `SetLifetime` sealed, which is where the difference is...

```csharp
public class LifetimeEntry
{
    private double lifetimeStart = double.MinValue;

    public double LifetimeStart
    {
        get => lifetimeStart;
        set => SetLifetimeStart(value);
    }

    private double lifetimeEnd = double.MaxValue;

    public double LifetimeEnd
    {
        get => lifetimeEnd;
        set => SetLifetimeEnd(value);
    }

    internal event Action<LifetimeEntry> RequestLifetimeUpdate;

    protected void SetLifetime(double start, double end)
    {
        SetLifetimeStart(start);
        SetLifetimeEnd(end);
    }

    protected virtual void SetLifetimeStart(double start)
    {
        if (start != lifetimeStart)
        {
            RequestLifetimeUpdate?.Invoke(this);
            lifetimeStart = start;
        }
    }

    protected virtual void SetLifetimeEnd(double end)
    {
        if (end != lifetimeEnd)
        {
            RequestLifetimeUpdate?.Invoke(this);
            lifetimeEnd = Math.Max(lifetimeStart, end); // Negative intervals are undesired.
        }
    }
}

public class HitObjectLifetimeEntry : LifetimeEntry
{
    private double realLifetimeStart = double.MinValue;
    private double realLifetimeEnd = double.MaxValue;

    protected override void SetLifetimeStart(double start)
    {
        realLifetimeStart = start;
        base.SetLifetimeStart(keepAlive ? double.MinValue : start);
    }

    protected override void SetLifetimeEnd(double end)
    {
        realLifetimeEnd = end;
        base.SetLifetimeEnd(keepAlive ? double.MaxValue : end);
    }
}
```
But I suppose this works fine as well.

smoogipoo (dismissed):
Seems fine to me. @peppy requesting your take too

osu.Framework/Graphics/Performance/LifetimeEntry.cs
peppy: I'm having a hard time reading the osu! side code changes with both `SetLifetimeEnd` *and* `SetLifetime` overridden. Would this still work with just the `start`/`end` variants marked `virtual` (as per @smoogipoo's suggestion above), and how would it look on the osu! side as a result? I'm thinking it may be easier to mentally parse.

ekrctb: I made `SetLifetime` non-virtual, and I think this <https://github.com/ppy/osu/pull/12592/commits/b87446a5778ee76c96f02cd163beebdc44476e24> is nicer.

peppy: That's much better, yep.

peppy (approved):

peppy:
@smoogipoo good to get this in?

smoogipoo (approved):
lgtm

Issue #4288, opened 2021-03-18
Move osu-framework to GitHub Actions


sr229:
As part of making the entire workflow stick to Actions (and to keep us always up to date with .NET upstream since it's Microsoft-supported), moving to Actions might be a best idea considering most of our workflow is supported by Actions.

bdach:
I don't really find the arguments here overly convincing. I see no particular reason to switch CI. There are way more important things to do.

smoogipoo:
It's low priority for sure, but definitely something we're looking to do in the long term.

peppy:
Would be great if someone would do this for us. I believe there were previous attempts but they weren't completed / merged. That may have been osu! side though.

sr229:
The reason I opened this because I asked in-stream if there were any stopgaps of moving to Actions, so I'm happily volunteering myself to do the migration (which may take time considering the Build Cake script is AppVeyor only)

bdach:
Cake should probably be done away with as part of this migration, actually. This was the plan a while ago, and I don't think we have much of a reason to keep it around.

sr229:
Got it, I'll keep that in mind while I'm writing it.

sr229:
Almost done porting the entire pipeline to Actions, I won't be able to clean out the AppVeyor script to prevent unnecessary failures. A follow-up PR is imperative once we transition completely to Actions.

![image](https://user-images.githubusercontent.com/14976516/116080263-6ea7e180-a6cb-11eb-8a8d-691dca7ef8a2.png)

PR #4407, opened 2021-04-26
Port CI to Actions

sr229:
Resolves #4288. This ports the entire CI to Actions. This is fairly incomplete on it's initial revisions at the moment. Currently it needs:

- [ ] CodeFileSanity
- [ ] Fix .NET Test
- [ ] Pack Library Steps

PR #4412, opened 2021-04-27
Migrate osu!framework to GitHub Actions

sr229:
Similar vein to GH-4407, but in a feature-complete state. Resolves GH-4288.

### Changes

- We're now using a standardized reporting with NUnit using the NUnit Logger with the NUnit Reporter Action. 

image

- CFS, MSBuild, Vika and Roslyn warnings will show up as warning and error annotations natively with no additional change needed.

image

- Secrets will be managed natively in GitHub, and can be called using a reference, making it much safer to use secrets.

### Caveats

- Currently there is no distinction between artifacts, you have to download the entire archive containing the nuget packages regardless if you have made a change in this specific package.
- Legacy scripts and AppVeyor definitions have not been cleaned out of this PR to allow AppVeyor checks to still pass. This can be tackled on a future PR once transition has started.

### Deprecations

- The cake script will be unused post-merge, so that means it has to be deprecated alongside AppVeyor if you plan to transition fully.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - "*"
+    tags:
+      - "*"
 
 name: Continuous Integration
 
@@ -94,7 +96,7 @@ jobs:
         id: deployment
         shell: bash
         run: |
-          if [ ${{ github.event == 'release' && github.actor == 'peppy' }} == true ]; then
+          if [ $(echo ${{github.ref}} | grep -q "refs/tags/"; echo $?) == 0 ]; then
             echo "::set-output name=VERSION::${GITHUB_REF#/refs\/tags\//}"
           else
             echo "::set-output name=VERSION::0.0.0+${{github.run_id}}"
@@ -121,7 +123,6 @@ jobs:
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v2
-        if: ${{ github.event != 'release' }}
         with:
           name: osu-framework
           path: ${{steps.artifactsPath.outputs.nuget_artifacts}}\*.nupkg
@@ -141,7 +142,7 @@ jobs:
         id: deployment
         shell: bash
         run: |
-          if [ ${{ github.event == 'release' && github.actor == 'peppy' }} == true ]; then
+          if [ $(echo ${{github.ref}} | grep -q "refs/tags/"; echo $?) == 0 ]; then
             echo "::set-output name=VERSION::${GITHUB_REF#/refs\/tags\//}"
           else
             echo "::set-output name=VERSION::0.0.0+${{github.run_id}}"
@@ -161,7 +162,6 @@ jobs:
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v2
-        if: ${{ github.event != 'release' }}
         with:
           name: osu-framework-templates
           path: ${{steps.artifactsPath.outputs.nuget_artifacts}}\*.nupkg
@@ -181,7 +181,7 @@ jobs:
         id: deployment
         shell: bash
         run: |
-          if [ ${{ github.event == 'release' && github.actor == 'peppy' }} == true ]; then
+          if [ $(echo ${{github.ref}} | grep -q "refs/tags/"; echo $?) == 0 ]; then
             echo "::set-output name=VERSION::${GITHUB_REF#/refs\/tags\//}"
           else
             echo "::set-output name=VERSION::0.0.0+${{github.run_id}}"
@@ -199,7 +199,6 @@ jobs:
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v2
-        if: ${{ github.event != 'release' }}
         with:
           name: osu-framework-android
           path: ${{steps.artifactsPath.outputs.nuget_artifacts}}\*.nupkg
@@ -218,7 +217,7 @@ jobs:
       - name: Set Variables
         id: deployment
         run: |
-          if [ ${{ github.event == 'release' && github.actor == 'peppy' }} == true ]; then
+          if [ $(echo ${{github.ref}} | grep -q "refs/tags/"; echo $?) == 0 ]; then
             echo "::set-output name=VERSION::${GITHUB_REF#/refs\/tags\//}"
           else
             echo "::set-output name=VERSION::0.0.0+${{github.run_id}}"
@@ -237,3 +236,23 @@ jobs:
         with:
           name: osu-framework-ios
           path: ${{steps.artifactsPath.outputs.nuget_artifacts}}/*.nupkg
+
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    needs: [pack-android, pack-framework, pack-template, pack-ios]
+    continue-on-error: true
+    steps:
+      - name: Get Git tag
+        id: tag
+        uses: olegtarasov/get-tag@v2.1
+
+      - name: Push to NuGet
+        run: |
+          if [ -z ${{steps.tag.outputs.tag}} ]; 
+            echo "::warn::Skipping publish, no tag detected."
+            exit 0;
+          else
+            echo "TBD! This is just to evaluate if this release pipeline works."
+            exit 0;
+          fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,15 +159,15 @@ jobs:
         with:
           name: osu-framework-templates
           path: ${{steps.artifactsPath.outputs.nuget_artifacts}}\*.nupkg
-  
+
   pack-android:
-     name: Pack (Android)
-     runs-on: windows-latest
-     needs: inspect-code
-     defaults:
+    name: Pack (Android)
+    runs-on: windows-latest
+    needs: inspect-code
+    defaults:
       run:
         shell: powershell
-     steps:
+    steps:
       - name: Checkout
         uses: actions/checkout@v2
 
@@ -184,6 +184,7 @@ jobs:
       - name: Set Artifacts Directory
         id: artifactsPath
         run: echo "::set-output name=NUGET_ARTIFACTS::${{github.workspace}}\artifacts"
+
       - name: Add msbuild to PATH
         uses: microsoft/setup-msbuild@v1.0.2
 
@@ -198,13 +199,13 @@ jobs:
           path: ${{steps.artifactsPath.outputs.nuget_artifacts}}\*.nupkg
 
   pack-ios:
-     name: Pack (iOS)
-     runs-on: macos-latest
-     needs: inspect-code
-     defaults:
-       run:
-         shell: bash
-     steps:
+    name: Pack (iOS)
+    runs-on: macos-latest
+    needs: inspect-code
+    defaults:
+      run:
+        shell: bash
+    steps:
       - name: Checkout
         uses: actions/checkout@v2
 
@@ -220,8 +221,6 @@ jobs:
       - name: Set Artifacts Directory
         id: artifactsPath
         run: echo "::set-output name=NUGET_ARTIFACTS::${{github.workspace}}/artifacts"
-      - name: Add msbuild to PATH
-        uses: microsoft/setup-msbuild@v1.0.2
 
       - name: Pack (iOS Framework)
         run: msbuild -bl:msbuildlog.binlog -v:m -target:Pack -r osu.Framework.iOS/osu.Framework.iOS.csproj -p:Configuration=Release -p:Version=${{steps.deployment.outputs.version}} -p:PackageOutputPath=${{steps.artifactsPath.outputs.nuget_artifacts}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -231,7 +231,7 @@ jobs:
 
       - name: Deploy
         run: |
-          if [ $(echo {{needs.check-if-tag.outputs.version}}| grep -q "0.0.0+${{github.run_id}}"; echo $?) == 0 ]; then
+          if [ $(echo ${{needs.check-if-tag.outputs.version}} | grep -q "0.0.0+${{github.run_id}}"; echo $?) == 0 ]; then
             echo "Skipping publish, no tag detected."
             exit 0;
           else
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,22 @@
 on: [push, pull_request]
-
 name: Continuous Integration
 
 jobs:
+  check-if-tag:
+    name: Get Git Tag
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{steps.deployment.outputs.version}}
+    steps:
+      - name: Set Variables
+        id: deployment
+        shell: bash
+        run: |
+          if [ $(echo ${{github.ref}} | grep -q "refs/tags/"; echo $?) == 0 ]; then
+            echo "::set-output name=VERSION::${GITHUB_REF#/refs\/tags\//}"
+          else
+            echo "::set-output name=VERSION::0.0.0+${{github.run_id}}"
+          fi
   test:
     name: Test
     runs-on: windows-latest
@@ -79,23 +93,13 @@ jobs:
   pack-framework:
     name: Pack (Framework)
     runs-on: windows-latest
-    needs: inspect-code
+    needs: [inspect-code, check-if-tag]
     defaults:
       run:
         shell: powershell
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-
-      - name: Set Variables
-        id: deployment
-        shell: bash
-        run: |
-          if [ $(echo ${{github.ref}} | grep -q "refs/tags/"; echo $?) == 0 ]; then
-            echo "::set-output name=VERSION::${GITHUB_REF#/refs\/tags\//}"
-          else
-            echo "::set-output name=VERSION::0.0.0+${{github.run_id}}"
-          fi
 
       - name: Set Artifacts Directory
         id: artifactsPath
@@ -114,7 +118,7 @@ jobs:
           dotnet-version: "5.0.x"
 
       - name: Pack (Framework)
-        run: dotnet pack -c Release osu.Framework /p:Version=${{steps.deployment.outputs.version}} /p:GenerateDocumentationFile=true  -o ${{steps.artifactsPath.outputs.nuget_artifacts}}
+        run: dotnet pack -c Release osu.Framework /p:Version=${{needs.check-if-tag.outputs.version}} /p:GenerateDocumentationFile=true  -o ${{steps.artifactsPath.outputs.nuget_artifacts}}
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v2
@@ -125,23 +129,13 @@ jobs:
   pack-template:
     name: Pack (Templates)
     runs-on: windows-latest
-    needs: inspect-code
+    needs: [inspect-code, check-if-tag]
     defaults:
       run:
         shell: powershell
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-
-      - name: Set Variables
-        id: deployment
-        shell: bash
-        run: |
-          if [ $(echo ${{github.ref}} | grep -q "refs/tags/"; echo $?) == 0 ]; then
-            echo "::set-output name=VERSION::${GITHUB_REF#/refs\/tags\//}"
-          else
-            echo "::set-output name=VERSION::0.0.0+${{github.run_id}}"
-          fi
 
       - name: Set Artifacts Directory
         id: artifactsPath
@@ -153,7 +147,7 @@ jobs:
           dotnet-version: "5.0.x"
 
       - name: Pack (Template)
-        run: dotnet pack -c Release osu.Framework.Templates /p:Configuration=Release /p:Version=${{steps.deployment.outputs.version}} /p:GenerateDocumentationFile=true /p:NoDefaultExcludes=true -o ${{steps.artifactsPath.outputs.nuget_artifacts}}
+        run: dotnet pack -c Release osu.Framework.Templates /p:Configuration=Release /p:Version=${{needs.check-if-tag.outputs.version}} /p:GenerateDocumentationFile=true /p:NoDefaultExcludes=true -o ${{steps.artifactsPath.outputs.nuget_artifacts}}
 
       - name: Upload Artifacts
         uses: actions/upload-artifact@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,146 @@
+on: [push, pull_request, release]
```
smoogipoo, line 1: I don't think we need `release`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,146 @@
...
+
+        - name: Report NUnit Logs
+          uses: MirrorNG/nunit-reporter@v1.0.9
+          with:
+            path: './TestResults/TestResults.xml'
+            access-token: ${{ secrets.GITHUB_TOKEN }}
+
+  inspect-code:
+     name: Code Quality
+     needs: test
```
smoogipoo: Any reason for this? Can run concurrently.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,146 @@
...
+        # https://itnext.io/how-to-support-multiple-net-sdks-in-github-actions-workflows-b988daa884e
+        - name: Install .NET 2.1.x LTS
+          uses: actions/setup-dotnet@v1
+          with:
+            dotnet-version: '2.1.x'
+
+        - name: Install .NET 3.1.x LTS
+          uses: actions/setup-dotnet@v1
+          with:
+             dotnet-version: '3.1.x'
```
smoogipoo: Are you absolutely sure you need this on `ubuntu-latest`? I'm able to run inspectcode locally on arch perfectly fine, but I do get errors on my windows VM.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,146 @@
...
+
+        - name: CodeFileSanity
+          run: |
+            # Suppress warnings from templates project
+            # https://unix.stackexchange.com/questions/104641/how-can-i-suppress-output-from-grep-so-that-it-only-returns-the-exit-status
+            dotnet codefilesanity | while read -r line; do
+               if [ $(echo $line| grep -q "./osu.Framework.Templates/"; echo $?) == 1 ]; then
+                 echo "::warning::$line"
+               fi
+            done
```
smoogipoo: I suppose this is fine for now, but please add a todo that this functionality should be added to CFS itself.

Also, have you tested that this actually fails the CI process?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,146 @@
...
+
+        - name: .NET Format (Dry Run)
+          run: dotnet format --dry-run --check
+
+        - name: InspectCode
+          run: |
+            dotnet jb inspectcode $(pwd)/osu-framework.Desktop.slnf --output=inspectcodereport.xml --cachesDir=$(pwd)/inspectcode --verbosity=WARN
+            dotnet nvika parsereport $(pwd)/inspectcodereport.xml  --treatwarningsaserrors
```
smoogipoo: Have you tested this also fails the CI process? Nvika's quite old (came before gh-actions) so I'd be surprised if it does (also: https://github.com/laedit/vika/issues/224).

```suggestion
            dotnet nvika parsereport $(pwd)/inspectcodereport.xml --treatwarningsaserrors
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,146 @@
...
+
+  pack-n-publish:
```
smoogipoo, line 79: `pack-and-publish`

Also, this shouldn't be publishing to nuget immediately. That's done manually on a pushed tag, and needs to be in another workflow.
All that this should be doing is uploading the artifacts. Though really, I'm not sure that this is required for the time being, since I don't believe any of us have been used the appveyor feed the last year (@peppy?), so it's probably fine to skip this altogether.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,146 @@
...
+            - name: Set Artifacts Directory
+              id: artifactsPath
+              run: echo "::set-output name=NUGET_ARTIFACTS::${{github.workspace}}\artifacts"
+
+            # FIXME: 3.1 LTS is required here because iOS builds still uses 3.1.
+            # https://itnext.io/how-to-support-multiple-net-sdks-in-github-actions-workflows-b988daa884e
+            - name: Install .NET 3.1.x LTS
+              uses: actions/setup-dotnet@v1
+              with:
+                  dotnet-version: '3.1.x'
```
smoogipoo: I don't understand - they use the Xamarin SDK which is neither .NET 3.1 nor 5. It's xamarin+msbuild.

They probably shouldn't even be compiled on windows in the first place, see: https://github.com/saamerm/Xamarin-GitHubActions-DevOps/blob/master/.github/workflows/main.yml

.github/workflows/ci.yml
```diff
@@ -0,0 +1,146 @@
...
+              run: msbuild -bl:msbuildlog.binlog -v:m -target:Pack -r osu.Framework.iOS/osu.Framework.iOS.csproj -p:Configuration=Release -p:Version=${{steps.deployment.outputs.version}} -p:PackageOutputPath=${{steps.artifactsPath.outputs.nuget_artifacts}}
+
+            - name: Pack (Android Framework)
+              run: msbuild -bl:msbuildlog.binlog -v:m -target:Pack -r osu.Framework.Android/osu.Framework.Android.csproj -p:Configuration=Release -p:Version=${{steps.deployment.outputs.version}} -p:PackageOutputPath=${{steps.artifactsPath.outputs.nuget_artifacts}}
+
+            - name: Pack (Native Libraries)
+              run: dotnet pack -c Release osu.Framework.NativeLibs /p:Configuration=Release /p:Version=${{steps.deployment.outputs.version}} /p:GenerateDocumentationFile=true -o ${{steps.artifactsPath.outputs.nuget_artifacts}}
+
+            - name: Pack (Template)
+              run: dotnet pack -c Release osu.Framework.NativeLibs /p:Configuration=Release /p:Version=${{steps.deployment.outputs.version}} /p:GenerateDocumentationFile=true /p:NoDefaultExcludes=true -o ${{steps.artifactsPath.outputs.nuget_artifacts}}
```
smoogipoo: Wrong project

.github/workflows/ci.yml
```diff
@@ -0,0 +1,146 @@
...
+              run: dotnet pack -c Release osu.Framework /p:Version=${{steps.deployment.outputs.version}} /p:GenerateDocumentationFile=true  -o ${{steps.artifactsPath.outputs.nuget_artifacts}}
+
+            - name: Pack (iOS Framework)
+              run: msbuild -bl:msbuildlog.binlog -v:m -target:Pack -r osu.Framework.iOS/osu.Framework.iOS.csproj -p:Configuration=Release -p:Version=${{steps.deployment.outputs.version}} -p:PackageOutputPath=${{steps.artifactsPath.outputs.nuget_artifacts}}
+
+            - name: Pack (Android Framework)
+              run: msbuild -bl:msbuildlog.binlog -v:m -target:Pack -r osu.Framework.Android/osu.Framework.Android.csproj -p:Configuration=Release -p:Version=${{steps.deployment.outputs.version}} -p:PackageOutputPath=${{steps.artifactsPath.outputs.nuget_artifacts}}
+
+            - name: Pack (Native Libraries)
+              run: dotnet pack -c Release osu.Framework.NativeLibs /p:Configuration=Release /p:Version=${{steps.deployment.outputs.version}} /p:GenerateDocumentationFile=true -o ${{steps.artifactsPath.outputs.nuget_artifacts}}
```
smoogipoo: Shouldn't be in this workflow.

smoogipoo (changes requested):

.github/workflows/ci.yml
peppy: I think there's not much harm in upload artifacts (assuming github has a sane retention policy - i'd be against having them stored permanently for every build as that is a waste of resources).

smoogipoo: Seems like the retention policy can be adjusted: https://docs.github.com/en/github/administering-a-repository/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-repository

.github/workflows/ci.yml
sr229: Using `setup-dotnet` overrides the active .NET installation on `PATH` so I had to install 3.1 LTS libraries for the builds to actually compile.

.github/workflows/ci.yml
peppy: 90 days sounds fine, or lower. That said, if it's adding a performance overhead switching off is also a feasible path.

.github/workflows/ci.yml
sr229: Any reasoning why this shouldn't be in this workflow? I closely followed the structure from Cake so I wasn't aware how you want the CI done.

smoogipoo: There is a completely separate appveyor workspace for publishing the native libs: https://github.com/ppy/osu-framework/blob/master/appveyor_native_deploy.yml

This is done manually. Otherwise we're uploading hundreds of MBs of packages with every o!f release.

.github/workflows/ci.yml
sr229: There should be no overhead in uploading the artifacts since they're there for testing purposes. The workflow only executes publish on a specific GitHub Event (which I assume is `tag`?) so this gets skipped altogether anyways.

.github/workflows/ci.yml
sr229: Haven't introduced a bad change to test things out, but Actions is designed to error if the said workflow sends a non-zero exit code (which is fairly standard for most analyzers).

.github/workflows/ci.yml
sr229: It only sends a warning, but it doesn't fail it, which is something we have to introduce in CFS either way.

.github/workflows/ci.yml
sr229: Inspect Code Workflow steps uses Bash syntax, and since I already have this structured it's no harm to use the Linux environment for Code Analysis.

.github/workflows/ci.yml
sr229: Got it, we'll make this non-linear in the next rev

.github/workflows/ci.yml
sr229: > They probably shouldn't even be compiled on windows in the first place, see: https://github.com/saamerm/Xamarin-GitHubActions-DevOps/blob/master/.github/workflows/main.yml

Regarding this, we could use `macos-latest`, but I wasn't certain if I should do that so I left it in a state that was close to what was on Cake. Sorry about that.

.github/workflows/ci.yml
smoogipoo: Yeah I know that about analyzers exiting on non-zero, wasn't sure whether nvika did that. Seems like it does:
```
[smgi@smgi osu-framework]$ dotnet jb inspectcode osu-framework.Desktop.slnf --output="build/inspectcodereport.xml" --verbosity=WARN && dotnet nvika parsereport build/inspectcodereport.xml --treatwarningsaserrors

...

[Fatal] Issues with severity error was found: the build will fail

[smgi@smgi osu-framework]$ $?
-bash: 5: command not found
```

We also can't just have it error and have the user fish for it in plain text. The inspectcode command already dumps thousands of lines of output (one for each file).

How does it look when it fails, can you provide a screenshot?

.github/workflows/ci.yml
sr229: Hmm, there isn't any alternative to listening to Git tags, so `release` was the closest thing.

.github/workflows/ci.yml
smoogipoo: Just to clarify, you could also just make a separate job for msbuild for the iOS/Android projects (windows-latest). There's no need to keep everything in one like this.

.github/workflows/ci.yml
peppy: It doesn't need to run on releases or tags. Tags are based on a commit on `master` which has inherently already passed CI.

.github/workflows/ci.yml
smoogipoo: What? I mean you shouldn't need to install LTS tools to run the dotnet tools.

I think the only one that's broken is dotnet-format, but that needs a lot of fixing up anyway, and it's fine to leave broken for now.

.github/workflows/ci.yml
sr229: Ah, I'll make it a seperate workflow soon.

.github/workflows/ci.yml
sr229: > Just to clarify, you could also just make a separate job just for the iOS/Android projects (`windows-latest`). There's no need to keep everything in one like this.

Got it, sorry if this was too monolithic, I mostly followed Cake with no afterthought of redesigning it again. Will do that in the next rev.

.github/workflows/ci.yml
sr229: > We also can't just have it error and have the user fish for it in plain text. The inspectcode command already dumps thousands of lines of output (one for each file).

Hmm, since InspectCode is also a GitHub Action, I think it'll be much safer to use that instead? [I think this is pretty close but I doubt this is what we want.](https://github.com/glassechidna/resharper-action)

.github/workflows/ci.yml
sr229: CFS Relies on 2.1 LTS which I thought should *just work* but it threw an error during development which led to this stopgap.

image

.github/workflows/ci.yml
sr229: Got it, I'll just run it against PRs and Pushes. As for releases I assume [Repository Dispatch](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#repository_dispatch) should be enough to cover the expected use case?

.github/workflows/ci.yml
smoogipoo: It's slightly out of date but you may be able to use the xml option (rather than solution):
```
with:
    report: $(pwd)/inspectcodereport.xml
```

.github/workflows/ci.yml
smoogipoo: See: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestbranchestags

You can run on tags.

.github/workflows/ci.yml
smoogipoo: Uh, well firstly CFS 15.0.0 is wrong, but I understand that it's because that's how it was in the cake file. It should be 0.0.36 which is targetted towards 3.1, and may be runnable by the NET5 runtime.

sr229:
As per review, tasks are more parallelized now. they've also been given unique artifact identifiers during upload which is close to what we have at appveyor.

.github/workflows/ci.yml
sr229: This should be resolved now by using the ReSharper Action.

.github/workflows/ci.yml
sr229: TODO added here, I'll see what I can do on the CFS side post-merge.

peppy:
Please consider fixing your editor up: get a yaml linter to show indentation issues, turn on whitespace cleaning on save. maybe turn on auto-format on save.

sr229:
> Please consider fixing your editor up: get a yaml linter to show indentation issues, turn on whitespace cleaning on save. maybe turn on auto-format on save.

I have the YAML linter installed on my editor but it seems not to catch stray whitespaces, I'll work on that later.

.github/workflows/ci.yml
sr229: > Ah, I'll make it a seperate workflow soon.

Just a follow-up on this: Would this workflow listen to pushes like the rest, or it will be triggered manually? (we can use `workflow_dispatch` if the latter.)

.github/workflows/ci.yml
sr229: This should be fixed but Actions is not firing for some reason, I'll get to fix that to see if moving versions worked.

.github/workflows/ci.yml
sr229: Added this as a seperate Dispatch workflow. Feel free to review that if there's something that doesn't seem to fit the bill.

sr229:
The workflows are pretty much done at this point as it's pretty close to the AppVeyor CI, however, I'm requesting another review before this gets merged (just to prevent some show-stoppers).

image

As for NativeLibs, I won't be able to test that since it needs to be in `master` to be tested, but this uses the same workflow code used in the main CI so I'm confident it'll work here as well.

peppy:
Re-reviews are implicit if you commit anything since last review, no need to ask.

Curious why you can't test NativeLibs on your fork's master though?

sr229:
> Curious why you can't test NativeLibs on your fork's master though?

Was planning to keep `master` clean as much as possible since it's used for syncing to ppy tree, though this PR might be an exception.

sr229:
Decided to merge this on the fork master, after some changes, NativeLibs is ready to go as well.

image

peppy:
> > Curious why you can't test NativeLibs on your fork's master though?
> 
> Was planning to keep `master` clean as much as possible since it's used for syncing to ppy tree, though this PR might be an exception.

You can force push your own master as much as you want. Or you can just create a second fork for testing if you see that as an issue.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,240 @@
...
+      - name: Restore Packages
+        run: dotnet restore
+
+      - name: CodeFileSanity
+        run: |
+          # TODO: Add ignore filters and GitHub Workflow Command Reporting in CFS. That way we don't have to do this workaround.
+          # FIXME: Suppress warnings from templates project
+          # https://unix.stackexchange.com/questions/104641/how-can-i-suppress-output-from-grep-so-that-it-only-returns-the-exit-status
+          dotnet codefilesanity | while read -r line; do
+             if [ $(echo $line| grep -q "./osu.Framework.Templates/"; echo $?) == 1 ]; then
```
ekrctb, line 77:
```suggestion
             if [[ "$line" != *"/osu.Framework.Templates/"* ]]; then
```

No need to create a process each line.

sr229: Not really a fan of Bash-conditionals, as much as possible, I use POSIX.

peppy: Personally I think the proposal reads better.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,240 @@
...
+
+      - name: InspectCode
+        run: dotnet jb inspectcode $(pwd)/osu-framework.Desktop.slnf --output=$(pwd)/inspectcodereport.xml --cachesDir=$(pwd)/inspectcode --verbosity=WARN
+
+      - name: ReSharper
+        uses: glassechidna/resharper-action@master
+        with:
+          report: ${{github.workspace}}/inspectcodereport.xml
+
+  pack-framework:
```
peppy, line 93: why is this whole thing in the `ci.yml`? should it not be its own workflow completely?

osu.Framework.Tests/osu.Framework.Tests.csproj
```diff
@@ -13,6 +13,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.9.4" />
     <PackageReference Include="NUnit" Version="3.13.1" />
     <PackageReference Include="NUnit3TestAdapter" Version="3.17.0" />
+    <PackageReference Include="NunitXml.TestLogger" Version="3.0.97" />
```
peppy, line 16: What is this used for?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,240 @@
+on: [push, pull_request]
+name: Continuous Integration
+
+jobs:
+  check-if-tag:
```
peppy, line 5: likewise, this tag getting stuff feels super out of place here. it should be inside a deploy yaml which is only run on release/tag in the first place.

peppy (changes requested):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,240 @@
...
+      - name: Compile
+        run: dotnet build -c Debug build/Desktop.proj
+
+      - name: Test
+        run: dotnet test $pwd/*.Tests/bin/Debug/*/*.Tests.dll --settings $pwd/build/vstestconfig.runsettings --logger:nunit
+        shell: powershell
+
+      - name: Report NUnit Logs
+        uses: MirrorNG/nunit-reporter@v1.0.9
+        if: ${{ always() }}
```
peppy, line 41: is this even required?

osu.Framework.Tests/osu.Framework.Tests.csproj
sr229: This replaces the AppVeyor Test Logger, it reports all the NUnit errors in Test Summary, it looks like this.

image

.github/workflows/ci.yml
sr229: Refer to my response [here](https://github.com/ppy/osu-framework/pull/4412#discussion_r623015171).

.github/workflows/ci.yml
sr229: I consolidated the release pipeline on the CI workflow since it'll be redundant anyways. The main idea here is it only makes a release for you if `github.ref` matches `refs/tags/x.y.z`, otherwise, it just uploads the artifacts.

.github/workflows/ci.yml
sr229: Different workflows cannot share jobs IIRC, so if you want a workflow to require task A but it's on a different workflow, you can't do so. Besides this will show up as a different CI check so it's a non issue if you're worried that it'll show up as one big workflow.

.github/workflows/ci.yml
sr229: tested it and it works just as well, thanks for the suggestion. Though I'd prefer this would be using POSIX conditionals for consistency but this will do for now.

.github/workflows/ci.yml
sr229: Also the reason I made the tag step was because of duplication, I decided to share the same logic for all workflows to be more consistent, it keeps the workflow cleaner as well.

sr229: Opted to change the terminology instead from workflow that it's main intent is to set Package version. Should resolve this properly.

osu.Framework.Tests/osu.Framework.Tests.csproj
peppy: Just curious where/how this interacts with the tests project itself. There are no usages of this within the project so it's going to be listed as a dependency which can be removed.

May be worthwhile adding a comment inline explaining why it's there so we don't do that by accident.

peppy (approved):

peppy:
Rather than dwelling on this for too long, I think this is in a fine state to get live so we can test how things work in our workflow (while also leaving appveyor running as the primary point of truth into the future, until we are confident with the change).

smoogipoo (approved):
looks fine now

PR #4425, opened 2021-04-30
Fix Test Coverage Reporting

sr229:
Fixes Test coverage not reporting as it should due to GitHub Limitations. This also uses the TRX format instead of requiring a seperate logger for coverage reporting.

**REQUIRES `report-nunit` WORKFLOW TO BE PRESENT IN THE BASE REPO TO ACTUALLY WORK!** 

image

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,8 @@ jobs:
         run: dotnet test $pwd/*.Tests/bin/Debug/*/*.Tests.dll --settings $pwd/build/vstestconfig.runsettings --logger "trx;LogFileName=TestResults.trx"
         shell: powershell
 
+      # Attempt to upload results even if test fails.
+      # https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#always
       - name: Upload Test Results
         uses: actions/upload-artifact@v2
         if: ${{ always() }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
         if: ${{ always() }}
         with:
           name: osu-framework-test-results
-          path: ${{github.workspace}}\TestResults.trx
+          path: ${{github.workspace}}\TestResults\TestResults.trx
 
   inspect-code:
     name: Code Quality
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,15 +33,15 @@ jobs:
         run: dotnet build -c Debug build/Desktop.proj
 
       - name: Test
-        run: dotnet test $pwd/*.Tests/bin/Debug/*/*.Tests.dll --settings $pwd/build/vstestconfig.runsettings --logger:nunit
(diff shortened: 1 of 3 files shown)


.github/workflows/report-nunit.yml
```diff
@@ -0,0 +1,23 @@
+# HACK: This is a workaround to allow PRs to report their coverage. This will run inside the base repository.
```
bdach, line 1: how is this a "hack" if this is the official github-sanctioned way to do this?

can leave the links, though.

.github/workflows/ci.yml
```diff
@@ -33,15 +33,15 @@ jobs:
...
       - name: Test
-        run: dotnet test $pwd/*.Tests/bin/Debug/*/*.Tests.dll --settings $pwd/build/vstestconfig.runsettings --logger:nunit
+        run: dotnet test $pwd/*.Tests/bin/Debug/*/*.Tests.dll --settings $pwd/build/vstestconfig.runsettings --logger "trx;LogFileName=TestResults.trx"
         shell: powershell
 
-      - name: Report NUnit Logs
-        uses: MirrorNG/nunit-reporter@v1.0.9
+      - name: Upload Test Results
+        uses: actions/upload-artifact@v2
         if: ${{ always() }}
```
bdach, line 41: i don't think this was ever properly explained in the original PR, but the reason this `always()` check is required is that [the artifact upload should be attempted even if the tests fail](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#always). a comment explaining this would be useful here, i think.

bdach:
also what are these `Signed-off-by: GitHub <noreply@github.com>` tags in the commit messages? were you using codespaces to edit these files or something?

sr229:
> also what are these `Signed-off-by: GitHub <noreply@github.com>` tags in the commit messages? were you using codespaces to edit these files or something?

Yep, using Codespaces atm since this was a quick fix. Sorry about that, I raised this bug regarding incorrect `user.name` and `user.email` being reported when using `git commit -s`. I'll fix this manually later.

bdach:
there is zero reason you should be using `-s` at all as a non-maintainer. if the original patch is yours, *you're the committer*. again, that's not what sign-off tags are used for in general (and we don't ever use them here, either).

also i would abstain from "fixing", because by "fixing" i am guessing you mean "reword all commits and force-push". that is not necessary.

.github/workflows/ci.yml
sr229: Should be commented now.

sr229:
> there is zero reason you should be using `-s` at all as a non-maintainer. if the original patch is yours, _you're the committer_. again, that's not what sign-off tags are used for in general (and we don't ever use them here, either).
> 
> also i would abstain from "fixing", because by "fixing" i am guessing you mean "reword all commits and force-push". that is not necessary.

Sorry about that, it became a habit of mine to sign commits since it was required to do so back on the projects I previously committed at. By I meant "fixing" is usually means "I'll have it fixed on the next PRs I submit".

bdach (approved):
seems fine

peppy (approved):

Issue #4349, opened 2021-04-08
Include ffmpeg libs in linux builds


smoogipoo:
The `osu.Framework.NativeLibs` package does not contain ffmpeg libs. Need to investigate the reason why and consider including them ourselves like is done for other platforms.

peppy:
FWIW the original reasoning was the unavailability of official (or semi official) static binaries we could be assured would be cross distribution compatible.

bdach:
This was attempted in #3846. Still unsure on the licensing kinks of that one.

bdach:
I'll just note that I've tried to test the new libraries with the most recent framework bump (https://github.com/ppy/osu/pull/18029) and they're not actually getting used for me, because:

```
      9322:	/lib/x86_64-linux-gnu/libm.so.6: error: version lookup error: version `GLIBC_2.29' not found (required by /home/dachb/Dokumenty/opensource/osu/osu.Desktop/bin/Debug/net6.0/runtimes/linux-x64/native/libavutil.so) (fatal)
      9322:	find library=avformat.so [0]; searching
      9322:	 search cache=/etc/ld.so.cache
      9322:	 search path=/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/lib:/usr/lib		(system search path)
      9322:	  trying file=/lib/x86_64-linux-gnu/avformat.so
      9322:	  trying file=/usr/lib/x86_64-linux-gnu/avformat.so
      9322:	  trying file=/lib/avformat.so
      9322:	  trying file=/usr/lib/avformat.so
      9322:	
      9322:	/lib/x86_64-linux-gnu/libc.so.6: error: version lookup error: version `GLIBC_2.28' not found (required by /home/dachb/Dokumenty/opensource/osu/osu.Desktop/bin/Debug/net6.0/runtimes/linux-x64/native/libavformat.so) (fatal)
      9322:	find library=avcodec.so [0]; searching
      9322:	 search cache=/etc/ld.so.cache
      9322:	 search path=/lib/x86_64-linux-gnu:/usr/lib/x86_64-linux-gnu:/lib:/usr/lib		(system search path)
      9322:	  trying file=/lib/x86_64-linux-gnu/avcodec.so
      9322:	  trying file=/usr/lib/x86_64-linux-gnu/avcodec.so
      9322:	  trying file=/lib/avcodec.so
      9322:	  trying file=/usr/lib/avcodec.so
``

[...]

PR #4430, opened 2021-05-04
Run Test Job as Matrix

### LINKED ISSUE #4426: Run tests in matrix with different platforms / threading modes
One thing we're missing at the moment in Actions is OS matrix for test suites. Since this is possible, we can write the test suite to run in multiple platforms, however there's also a requirement raised that we have to do these test suites with different threading modes, which I doubt there's a setting/property exposed to run these tests with a different threading mode. This issue will track the needed changes to get the test suites running with:

- [x] OS Matrix (GH-4430)
- [x] Threading Mode (Merged via Prerequisite PR GH-4427)

sr229:
Closes #4426. This will run the `test` job against all supported platform runners in GitHub Actions on SingleThread and MultiThread mode.

Currently tests are flaky so I'm not sure how to fix this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
 
       - name: Test
         run: dotnet test $pwd/*.Tests/bin/Debug/*/*.Tests.dll --settings $pwd/build/vstestconfig.runsettings --logger "trx;LogFileName=TestResults-${{matrix.os.prettyname}}.trx"
-        shell: powershell
+        shell: pwsh
 
       # Attempt to upload results even if test fails.
       # https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#always
@@ -50,7 +50,7 @@ jobs:
         if: ${{ always() }}
         with:
           name: osu-framework-test-results
-          path: ${{github.workspace}}\TestResults\TestResults-${{matrix.os.prettyname}}.trx
+          path: ${{github.workspace}}/TestResults/TestResults-${{matrix.os.prettyname}}.trx
 
   inspect-code:
     name: Code Quality
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,13 +20,16 @@ jobs:
   test:
     name: Test
     runs-on: ${{matrix.os.fullname}}
+    continue-on-error: true
+    env:
+      OSU_EXECUTION_MODE: ${{matrix.threadingMode}}
     strategy:
        matrix:
           os:
             - { prettyname: Windows, fullname: windows-latest }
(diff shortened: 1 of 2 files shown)


sr229:
I would like to remark Test Reporting will be broken since this PR includes the proper fixes.

.github/workflows/ci.yml
```diff
@@ -29,21 +39,28 @@ jobs:
         with:
           dotnet-version: "5.0.x"
 
+      # FIXME: libavformat is not included in Ubuntu. Let's fix that.
```
bdach, line 42: not sure what this comment is trying to convey by "let's fix that", but if you mean as a todo in the future, then i'd link the actual issue instead (https://github.com/ppy/osu-framework/issues/4349)

.github/workflows/ci.yml
```diff
@@ -29,21 +39,28 @@ jobs:
...
+         sudo apt-get update && \
+         sudo apt-get -y install libavformat-dev
+
       - name: Compile
         run: dotnet build -c Debug build/Desktop.proj
 
       - name: Test
-        run: dotnet test $pwd/*.Tests/bin/Debug/*/*.Tests.dll --settings $pwd/build/vstestconfig.runsettings --logger "trx;LogFileName=TestResults.trx"
-        shell: powershell
+        run: dotnet test $pwd/*.Tests/bin/Debug/*/*.Tests.dll --settings $pwd/build/vstestconfig.runsettings --logger "trx;LogFileName=TestResults-${{matrix.os.prettyname}}.trx"
```
LittleEndu, line 53: Isn't this missing the threading side of the matrix? Only 3 out of 6 test results are uploaded

bdach: yeah, seems like artifacts from one OS will overwrite each other in non-deterministic fashion because of how they're named.

bdach (changes requested):

.github/workflows/ci.yml
sr229: I should fix that later on, good catch.

.github/workflows/ci.yml
sr229: Thanks for the issue link, but this will be removed anyways since I opened a [issue](https://github.com/actions/virtual-environments/issues/3306) to include the dependency inside ubuntu-latest.

bdach:
Aside from the culture test failures, which are to be addressed by the PR above, I am also locally seeing failures at:

- [ ] `GameHostSuspendTest.TestPauseResume`
- [ ] `TestSceneSampleLooping.TestZeroFrequencyOnStart`

Probably best addressed before proceeding with this merge.

smoogipoo (approved):

bdach (approved):

PR #4433, opened 2021-05-05
Fix SampleChannels with zero frequency not playing in single-threaded execution

smoogipoo:
Fixes the sample channel test failure mentioned in https://github.com/ppy/osu-framework/pull/4430#issuecomment-832154842.

This was caused by the playChannel() callback temporarily setting `Playing` to false by virtue of setting `enqueuedPlaybackStart` to false.

In multi-threaded execution, this is not a problem because `UpdateState` will then run and set `Playing` back to true by setting `playing` to true.

In single-threaded execution, this _is_ a problem because `UpdateState` only runs on the next audio frame, at which point the `SampleChannel` is now dead (due to `Playing = false`) and won't receive the `UpdateState`.

### DIFF
--- a/osu.Framework/Audio/Sample/SampleChannelBass.cs
+++ b/osu.Framework/Audio/Sample/SampleChannelBass.cs
@@ -159,11 +159,10 @@ namespace osu.Framework.Audio.Sample
                 if (Played && Bass.ChannelIsActive(channel) == PlaybackState.Stopped)
                     return;
 
-                if (relativeFrequencyHandler.IsFrequencyZero)
-                    return;
-
-                Bass.ChannelPlay(channel);
                 playing = true;
+
+                if (!relativeFrequencyHandler.IsFrequencyZero)
+                    Bass.ChannelPlay(channel);
             }
             finally
             {


peppy (approved):

peppy:
This seems like a valid fix, but the flags surrounding this logic (`enqueuedPlaybackStart` and `playing`) feel very flaky. Hoping we can refactor this logic at a point in the future.

smoogipoo:
I'm open to suggestions, because I'd improve the surrounding logic if I could but I don't see a way to - I wrote this entire code from the ground up.

The only way I see is to change single threaded execution to always call `UpdateState()` after each action.

peppy:
All I'll say is that things get even more interesting with `BassMix`, where internal bass states we expect to be immediate are no longer such (and require a `Bass.Update` call before they update). You can see the changes I had to make to get things working again:

https://github.com/ppy/osu-framework/pull/4291/commits/f8d97e467cca16f49ab08d7ea14ac44fca6df544
https://github.com/ppy/osu-framework/pull/4291/commits/605173ada553644993a9ffdf39fbea235341e0c7

PR #4437, opened 2021-05-05
Fix sample channel test failing intermittently

smoogipoo:
As seen in https://github.com/ppy/osu-framework/runs/2511077249

Can be repro'd by adding a `Thread.Sleep(3000);` at the start of the enqueued action in `SampleChannelBass.playChannel()`.

This test relied on some logic which has changed since https://github.com/ppy/osu-framework/pull/4336 - `Play()` now immediately sets `Playing = true`. So this was failing because the audio thread was not having enough time to run in-between update frames to set it back to false.


peppy (approved):

smoogipoo:
Test test failure in CI is unrelated.

PR #4438, opened 2021-05-05
Split Pack Tasks Entirely

sr229:
This removes the Pack Tasks away from the main CI, and calls them via a dispatch flow instead during a tagged commit. ~~While it's possible to call them manually [you're limited to branches in UI.](https://github.community/t/workflow-dispatch-from-a-tag-in-actions-tab/130561)~~ The dispatch workflows has now been designed to derive a version from HEAD if it has a tag.

image

image


smoogipoo:
I was reading about this earlier - doesn't this do what's necessary?: https://stackoverflow.com/a/59349906 (read the comment too)

So something like:
```yml
on:
  push:
    tags:
      - '*'
```

sr229:
> I was reading about this earlier - doesn't this do what's necessary?: https://stackoverflow.com/a/59349906 (read the comment too)
> 
> So something like:
> 
> ```yaml
> on:
>   push:
>     tags:
>       - '*'
> ```

IIRC this needed to be done manually, so `workflow_dispatch` was used. I can switch it out right now if so.

smoogipoo:
Ah right, my bad. Automatic wouldn't make much sense anyway given nativelibs...

.github/workflows/ci.yml
```diff
@@ -110,150 +95,20 @@ jobs:
...
-      - name: Deploy
-        run: |
-          if [ $(echo ${{needs.check-if-tag.outputs.version}} | grep -q "0.0.0+${{github.run_id}}"; echo $?) == 0 ]; then
-            echo "Skipping publish, no tag detected."
-            exit 0;
-          else
-            dotnet nuget add source https://api.nuget.org/v3/index.json -n authed-nuget -u ${{secrets.NUGET_USER_NAME}} -p ${{secrets.NUGET_AUTH_TOKEN}}
-            dotnet nuget push ${{github.workspace}}/artifacts/*.nupkg --skip-duplicate --source authed-nuget
-          fi
+  dispatch-release:
```
smoogipoo, line 98: So I'm a little confused how this entire process works.

The nativelibs and pack workflows have a separate button to run their workflow, via `workflow_dispatch`: 
image

But the workflow is _also_ being dispatched from the main CI here. Am I right in understanding that this is intending to dispatch on a tag? Because if so, and this should all be manual, then this should not be here.

smoogipoo: Ah, this is because you can't select a tag? With the comments in the other .yml files?

sr229: My main worry is that the UI limitations would make this dispatch impossible, since running this by tags is API-facing only. I can remove it now.

.github/workflows/ci.yml
```diff
@@ -110,150 +95,20 @@ jobs:
...
+     name: Dispatch Release
+     needs: [test, inspect-code]
+     if: ${{ ${{ github.event != 'pull_request' && contains(github.ref, 'refs/heads/')  != true }} || github.event == 'pull_request' }}
```
bdach: What is happening here?

What's with the `(a && b) || a` check? What's `refs/heads/`? And shouldn't this dispatch on new tags only? As in, run on [`create`](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#create) events?

.github/workflows/ci.yml
smoogipoo: Have you considered just calling `git describe --tags --exact-match`?

.github/workflows/ci.yml
```diff
@@ -110,150 +95,20 @@ jobs:
...
-            echo "Skipping publish, no tag detected."
-            exit 0;
-          else
-            dotnet nuget add source https://api.nuget.org/v3/index.json -n authed-nuget -u ${{secrets.NUGET_USER_NAME}} -p ${{secrets.NUGET_AUTH_TOKEN}}
-            dotnet nuget push ${{github.workspace}}/artifacts/*.nupkg --skip-duplicate --source authed-nuget
-          fi
+  dispatch-release:
+     name: Dispatch Release
+     needs: [test, inspect-code]
+     if: ${{ github.event != 'pull_request' && contains(github.ref, 'refs/heads/')  != true || github.event == 'pull_request' }}
```
bdach, line 101: What is happening here?

What's with the `(a && b) || !a` check? Surely you mean `!a || b`? What's `refs/heads/`? And shouldn't this dispatch on new tags only? As in, run on [`create`](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#create) events, not on everything?

.github/workflows/ci.yml
```diff
@@ -110,150 +95,20 @@ jobs:
...
+         uses: benc-uk/workflow-dispatch@v1
+         with:
+           workflow: 'Continuous Integration - Pack'
+           token: ${{ secrets.GITHUB_TOKEN }}
+
+       - name: Dispatch Release (Native Libraries)
+         uses: benc-uk/workflow-dispatch@v1
+         with:
+           workflow: 'Continuous Integration - NativeLibs'
+           token: ${{ secrets.GITHUB_TOKEN }}
```
bdach, line 114: So am I to understand that each framework release will be accompanied by a nativelibs release? Bit extravagant, and not how things were usually done up to now. How do you bump nativelibs and then use that bumped version in an actual framework release with this flow?

.github/workflows/deploy-pack.yml
```diff
@@ -0,0 +1,181 @@
+# There is no manual way to call this out to run this on tags via UI.
+# See: https://github.community/t/workflow-dispatch-from-a-tag-in-actions-tab/130561
+on: workflow_dispatch
+name: Continuous Integration - Pack
```
bdach, line 4: this still says CI; should be
```suggestion
name: Deployment - Pack
```
probably?

.github/workflows/deploy-nativelibs.yml
```diff
@@ -8,18 +10,33 @@ jobs:
...
+            echo "::set-output name=VERSION::$(git describe --exact-match --tags HEAD)"
           else
-            echo "::set-output name=VERSION::0.0.0+${{github.run_id}}"
+            echo "::set-output name=VERSION::"
           fi
+
   deploy:
     name: Deploy
     runs-on: ubuntu-latest
+    if: ${{ contains(needs.check-if-tag.outputs.version, '')  == false }}
```
bdach, line 39: this seems bogus? any string contains an empty string, does it not? should this not be just

```suggestion
    if: ${{ needs.check-if-tag.outputs.version != '' }}
```

also see same conditional in `deploy-pack.yml`

bdach (changes requested):

.github/workflows/deploy-nativelibs.yml
sr229: Tried the following before and it'll still execute the step either way, which was the main reason I took too long on finalizing the logic because the evaluation is not working as it should.

bdach: ...and what you have *doesn't* always execute? something is *very* broken on github's side if this is really the case, then

sr229: Decided to make Package Version job fail entirely if HEAD has no tag which is probably what we want anyways 

image

peppy:
Please refrain from mindlessly merging master into your PRs. We will do this when necessary.

.github/workflows/deploy-nativelibs.yml
```diff
@@ -8,18 +10,34 @@ jobs:
...
           else
-            echo "::set-output name=VERSION::0.0.0+${{github.run_id}}"
+            echo "fatal: no tag detected for HEAD. Workflow will now stop."
+            exit 128;
           fi
+
   deploy:
     name: Deploy
     runs-on: ubuntu-latest
+    if: ${{ contains(needs.check-if-tag.outputs.version, '')  == false }}
```
bdach, line 40: this check was not deleted properly in 1f2e59f596b0b1f15d0bbc2d7568ca40953e6920. please fix it, because when i tried to do so myself, i ended up pushing a branch to this repository rather than pushing a commit to your fork.

and i don't know how you've managed to break maintainer pushing to PRs, but please, for the love of anything holy sort out this remote situation so that we can fix things for you and speed up this already gruesomely long process. because in fact [i believe this is the *second* time in two days that the same fiasco happens](https://discord.com/channels/188630481301012481/589331078574112768/839492356704436272), caused by us assuming that things work normally. it might be that "codespaces" thing again. if it is, i would actually recommend *not* using it as it looks to be terminally broken and not fit for actual development.

bdach (changes requested):

.github/workflows/deploy-nativelibs.yml
sr229: I'll remove the check now, seems a small oversight from me, whoops. As for broken pushes, I have no idea what happened, but that's something I may have to report to GitHub. I'll refrain using this while making PRs from now on until I clear this up.

peppy (approved):

PR #4439, opened 2021-05-05
Fix DLW test failing intermittently

smoogipoo:
As seen in https://github.com/ppy/osu-framework/runs/2511281078.

Can be tested by adding a `Thread.Sleep(1000);` in `TestBox.load()`.

There's technically three problems here, and this (using `LifetimeStart` instead of `LifetimeEnd`) is the easiest way to work around all issues without knowing how to fix them yet. The issues are:
1. `LifetimeEnd` causes the DLW to be expired before the contents finish loading. This immediately causes the DLW to be removed from its parent without overriding `RemoveWhenNotAlive` to `false`.
2. Regardless of the above, an expired DLW stops receiving updates and thus the scheduled addition does not run (scheduled to the DLW itself). One would think that setting `LifetimeEnd` to an absurdly high value would resolve this but also:
3. DLW briefly takes on its content's lifetime during the async load process.

### DIFF
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneDelayedLoadWrapper.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneDelayedLoadWrapper.cs
@@ -134,7 +134,7 @@ namespace osu.Framework.Tests.Visual.Drawables
         {
             Container wrapped = null;
             DelayedLoadWrapper wrapper = null;
-            double lifetimeEnd = 0;
+            double lifetimeStart = 0;
 
             AddStep("create child", () =>
             {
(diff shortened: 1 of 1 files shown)


bdach:
Test fix looks good.

As for a discussion of the three issues, the first two points seem unfortunate yet somewhat expected, and about the only potentially WTF part to me seems to be the third (DLW taking on lifetime from loading content).

Thing is it's a bit hard to define at which point the drawable's lifetime should already be "valid" if it's a plain property. It would be easy enough to do something like

```csharp
public override double LifetimeStart
{
	get => Content?.IsLoaded == true ? Content.LifetimeStart : lifetimeStart;
	set { ... }
}
```

but I'm not 100% that's what the consumers *want* to happen, per se... Food for thought for later, I suppose.

bdach (approved):

peppy:
Superseded by #4441.

Issue #4440, opened 2021-05-05
DelayedLoadWrapper briefly takes on its content's lifetime during async


smoogipoo:
Right now, DLW briefly takes on its contents' lifetime during the async load process. This is caused by the overrides of `LifetimeStart_get()` and `LifetimeEnd_get()` which don't take into account the load completion:
https://github.com/ppy/osu-framework/blob/635e4199dd3e3b4ae8964f057f482aa94862c7ce/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs#L63-L87

This brings up a few questions regarding the proper way to use lifetime with DLW.

### 1. Should the mentioned getters consider the value of `DelayedLoadComplete`?

Presumably the reason why these exist is for such code:
```csharp
new DelayedLoadWrapper(new Drawable
{
    LifetimeStart = A,
    LifetimeEnd = B
});
```
To make the DLW alive only during [A, B].

More recently, however, DLWs can also be instantiated with `Func<Drawable>` content:
```csharp
new DelayedLoadWrapper(() => new Drawable
{
    LifetimeStart = A,
    LifetimeEnd = B
});
```
Which causes DLWs to exist with their own lifetime for some time until the content begins loading, at which point they take on the [A, B] lifetime as above. This is pretty weird because DLWs could be alive before the load, but then die immediately after starting the load and block the content from being added:
https://github.com/ppy/osu-framework/blob/635e4199dd3e3b4ae8964f057f482aa94862c7ce/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs#L156-L171
(Note the `Schedule`).

If the getters were to consider the value of `DelayedLoadComplete`, then this wouldn't be the case, however the first example above would break. Leading to:

### 2. What is the proper way to use lifetimes with DLW?

The above examples are probably not representative of the way DLW should actually be used. The most common usages are:
```csharp
Drawable d;
new DelayedLoadWrapper(d = new Drawable());
d.OnLoadComplete += _ => d.FadeIn()'
```
(No `.Expire()` at all)

```csharp
Drawable d;
new DelayedLoadWrapper(d = new Drawable());
d.OnLoadComplete += _ =>
{
    d.FadeIn();
    someOtherWrapper.FadeOut().Expire();
}
```
(`Expire()` the DLW itself. See: `ModelBackedDrawable`)

There are no cases outside of tests where the DLW's content is `Expire()`d. Given this, it may not be a bad idea to remove the getter overrides altogether?

[...]

PR #4441, opened 2021-05-05
Remove DLW's special lifetime handling

smoogipoo:
Based on the findings found in https://github.com/ppy/osu-framework/issues/4440 - there are no cases where this passing of lifetime is used in any meaningful fashion.


bdach:
I guess this is an alternative to #4439, as in if this one is merged, might as well not bother with that one.

This would be a pretty major breaking change but I'm not necessarily opposed to it in principle. I guess if the lifetimes are decoupled entirely, the worst case scenario is having a DLW with a content that becomes immediately dead. But that sorta makes sense too. Neither way is 100% good in all use cases, lifetime is one of the trickier parts of framework to grasp anyways.

All things aside I would definitely want to at least run this past game-side tests with a framework checkout before approving, though, if you haven't already - just to be extra sure I guess.

smoogipoo:
> All things aside I would definitely want to at least run this past game-side tests with a framework checkout before approving, though, if you haven't already - just to be extra sure I guess.

Haven't done this yet.

> the worst case scenario is having a DLW with a content that becomes immediately dead

When do you foresee this happening? Is this a case where you'd `.Expire()` the content somehow, which we haven't done yet?

bdach:
> Is this the case where you'd `.Expire()` the content somehow, which we haven't done yet?

That one might not be as much of a problem, because usually an expire usually means that the caller is done with the drawable anyways. I'm more worried about cases like pooling/drawable reuse and somebody forgetting to reset the lifetime somehow. But then again there's no real sane way out of that, either, other than just the programmer making sure everything is set correctly...

smoogipoo:
The thing is, the current code means that the DLW itself could take on that bad lifetime and prevent the pooled drawable/whatnot from finishing the load process. The way around that is to consider `DelayedLoadComplete` in the getters as I mentioned in the issue, but then you start to introduce more special cases which become even more confusing for a user of the framework.

Imo there's something to be said for keeping things simple and consistent - in this case it's generally understood that when you touch lifetime, you only affect the drawable itself and don't cause side-effects in parents.

peppy (approved):
I think this is fine. Have run against osu!-side tests and it looks to behave okay.

Issue #3369, opened 2020-03-13
Single-frame layout issue with nested child dependencies


smoogipoo:
## Description

A single-frame layout discrepancy may be observed when nesting containers that are invalidated by their children.

Case 1 - autosize in autosize: https://drive.google.com/open?id=1T_MGjtFTowFU6OAD1BhpN-_fRG29U8Bi
In this case the parenting auto-size is invalidated when the inner autosize is updated (at the end of the frame).

Case 2 autosize in fill flow: https://drive.google.com/open?id=1ePqGK-hArak5mIbDv0Khi-OsQ1LrsVmw
In this case, the parenting fill flow layout is invalidated when the inner autosize is updated (at the end of the frame).

Both of these test cases can be found below.

This isn't a regression from the recent layout changes, but has always been the case for performance reasons.

## Path forward

There are two changes required for this:

In `CompositeDrawable.cs`:
```diff
- // Child invalidations should not propagate to other children.
- if (source == InvalidationSource.Child)
-     return anyInvalidated;
```

In `Drawable.cs`:
```diff
- if (source == InvalidationSource.Self)
+ if (source == InvalidationSource.Self || source == InvalidationSource.Child)
    Parent?.Invalidate(invalidation, InvalidationSource.Child);
```

However we can't do this without first considering the performance implications, and I'm not really sure why the `CompositeDrawable` change is required in the first place.

We likely can't fix this without first removing `OnInvalidate()` completely, and turning it into something delegate-based. A likely path forward while maintaining high performance is to store a reverse enum of child invalidation type dependencies - that is, which invalidation types any of a drawable's supertree cares about.

## Test scene
```cs
public class TestSceneLayoutIssue : FrameworkTestScene
{
    [Test]
    public void TestCase1()
    {
        TestContainer t = null;

        AddStep("create test", () =>
        {
            Child = new Container
            {
                Anchor = Anchor.Centre,
                Origin = Anchor.Centre,
                AutoSizeAxes = Axes.Both,
                Children = new Drawable[]
                {
                    new Box
                    {
                        RelativeSizeAxes = Axes.Both,
                        Colour = Color4.SlateGray
                    },
                    t = new TestContainer()
                }
            };
        });

        AddStep("set some text", () => t.Text = "whoops!");
    }

    [Test]
    public void TestCase2()
    {
        TestContainer t = null;

        AddStep("create test", () =>
        {
            Child = new Container
            {
                Anchor = Anchor.Centre,
[...]

PR #4445, opened 2021-05-10
Keep anchor and origin when wrapping drawable via `ContainerExtensions.Wrap`

frenzibyte:
Closes https://github.com/ppy/osu/issues/12733 as per https://github.com/ppy/osu/issues/12733#issuecomment-835938318

Seems arbitrary to revert the anchor/origin of the wrapped drawable back to `TopLeft` as the container should have the same exact size as the wrapped drawable, rendering that pointless.

And reverting the `Origin` could cause issues to drawables that, for example, rescale themselves from inside the implementation, rather than externally on the wrapper container.

### DIFF
--- a/osu.Framework/Graphics/Containers/ContainerExtensions.cs
+++ b/osu.Framework/Graphics/Containers/ContainerExtensions.cs
@@ -39,8 +39,6 @@ namespace osu.Framework.Graphics.Containers
 
             drawable.Position = Vector2.Zero;
             drawable.Rotation = 0;
-            drawable.Anchor = Anchor.TopLeft;
-            drawable.Origin = Anchor.TopLeft;
 
             // For anchor/origin positioning to be preserved correctly,
             // relatively sized axes must be lifted to the wrapping container.


smoogipoo (approved):
I think this makes sense. Can't really think of a reason it's done this way either, and blame doesn't give any clues.

peppy:
If we are removing this why are we leaving the two lines above? I'd be hesitant to merge this without cross checking all usages.

smoogipoo:
Because position and rotation have effects, and they're moved into the parent here.

smoogipoo:
I believe the issue here is nested autosize. See case 1 from https://github.com/ppy/osu-framework/issues/3369.

The nested size is updated first but it doesn't reposition due to `TopLeft` anchor/centre - its parent is used for repositioning. But its parent's size is updated in the next frame due to autosize, causing the linked issue.
By leaving the nested drawable as it is, it will immediately reposition to the centre of its parent even though its parent's size hasn't updated it.

peppy:
Right, that makes more sense. I guess this is fine if it fixes the one-frame issue in more cases without causing any problems.

In a quick pass everything using this wrapping looks to function correctly.

PR #4446, opened 2021-05-10
Use OpenSans in markdown container test

smoogipoo:
This is the most complete font available in the framework containing Bold, Italic, and BoldItalic styles.


gagahpangeran:
Looks like 'italic with underscore' and 'italic with asterisk' don't work 🤔

Screenshot from 2021-05-10 19-41-35

peppy (approved):

PR #4450, opened 2021-05-12
Fix incorrect comparison in `Dropdown` resulting in deselection of valid selection

peppy:
The `KeyCollection` class does not use the default `EqualityComparer` as one would expect, falling back to a reference comparison. This was causing the skin dropdown to behave incorrectly osu!-side.

I've added a new test here, but the existing `TestBasic` will already fail with the move away from `string`s as the model type in all dropdown tests.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/Dropdown.cs
+++ b/osu.Framework/Graphics/UserInterface/Dropdown.cs
@@ -63,7 +63,7 @@ namespace osu.Framework.Graphics.UserInterface
             foreach (var entry in items)
                 addDropdownItem(GenerateItemText(entry), entry);
 
-            if (Current.Value == null || !itemMap.Any(i => EqualityComparer<T>.Default.Equals(i.Key, Current.Value)))
+            if (Current.Value == null || !itemMap.Keys.Contains(Current.Value, EqualityComparer<T>.Default))
                 Current.Value = itemMap.Keys.FirstOrDefault();
             else
                 Current.TriggerChange();
--- a/osu.Framework/Graphics/UserInterface/Dropdown.cs
+++ b/osu.Framework/Graphics/UserInterface/Dropdown.cs
@@ -63,7 +63,7 @@ namespace osu.Framework.Graphics.UserInterface
             foreach (var entry in items)
                 addDropdownItem(GenerateItemText(entry), entry);
 
-            if (Current.Value == null || !itemMap.Keys.Contains(Current.Value))
+            if (Current.Value == null || !itemMap.Any(i => EqualityComparer<T>.Default.Equals(i.Key, Current.Value)))
                 Current.Value = itemMap.Keys.FirstOrDefault();
             else
                 Current.TriggerChange();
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/UserInterface/Dropdown.cs
```diff
@@ -63,7 +63,7 @@ private void setItems(IEnumerable<T> items)
             foreach (var entry in items)
                 addDropdownItem(GenerateItemText(entry), entry);
 
-            if (Current.Value == null || !itemMap.Keys.Contains(Current.Value))
+            if (Current.Value == null || !itemMap.Any(i => EqualityComparer<T>.Default.Equals(i.Key, Current.Value)))
```
bdach, line 66: You could also write

```suggestion
            if (Current.Value == null || !itemMap.Keys.Contains(Current.Value, EqualityComparer<T>.Default))
```

It seems to read a little better to me but I'm not fussed.

bdach:
Bit of a tangent, but here goes:

I was a bit surprised by this. I thought that dictionaries were supposed to work with equality comparers, but there's actually seemingly quite a deep cut on the dotnet side that makes `Dictionary.ContainsKey()` fail here.

First, I tried to go the path of least resistance and use `.ContainsKey()` directly. Turns out, it is exactly as you say - if there is no equality comparer given, then [`object.GetHashCode()` is used to find the appropriate hash bucket](https://github.com/dotnet/runtime/blob/720ee21578961961600f36cc8ca611a6d37206ca/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs#L366-L370). Ok, alright.

Then I remember `Dictionary<T>` instances can be constructed with an `IEqualityComparer<T>` instance. So I can pass in `IEqualityComparer<T>.Default`, right?... [Oh. No. No, I cannot.](https://github.com/dotnet/runtime/blob/720ee21578961961600f36cc8ca611a6d37206ca/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs#L57-L60)

For added WTF, after fetching the hash bucket index via `object.GetHashCode()` (even though `IEqualityComparer<T>` has a `GetHashCode()` method), [the inner code, that checks for hash collisions, uses `IEqualityComparer<T>.Equals(T)`](https://github.com/dotnet/runtime/blob/720ee21578961961600f36cc8ca611a6d37206ca/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs#L420-L423). So this creates this inconsistency that if you implement `IEquatable<T>` on a class and use it via an `IEqualityComparer<T>`, but don't override `object.GetHashCode()` to match, you'll get inconsistent behaviour like here.

This is not really dotnet's fault, to be fair. `IEquatable<T>` does not have a `GetHashCode()` member, and an `IEqualityComparer<T>` cannot infer how to make one off of a `IEquatable<T>` in any conceivable manner.

peppy:
Yep, it's one of those things you just need to be super-aware of when using `Dictionary` types. Interestingly, there is an inspection to ensure you implement `GetHashCode` when overriding `Equals`, but for whatever reason this only applies to the non-generic equality:

C04D7733-F39E-47E1-B1E0-D1C53AAACC0A

Not sure if there's a reason this isn't included for `IEquatable<T>`.

smoogipoo (approved):

PR #4457, opened 2021-05-21
Add support for `TestCaseSourceAttribute` in test scenes

frenzibyte:
Done mostly as in the [documentation](https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcasesource.html), might have a small difference between but don't want to stuff more things into this any further.

### DIFF
--- a/osu.Framework.Tests/Visual/Testing/TestSceneTestWithSource.cs
+++ b/osu.Framework.Tests/Visual/Testing/TestSceneTestWithSource.cs
@@ -1,0 +1,113 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Collections;
+using NUnit.Framework;
+using osu.Framework.Graphics.Containers;
+
+namespace osu.Framework.Tests.Visual.Testing
+{
+    public class TestSceneTestWithSource : FrameworkTestScene
+    {
+        protected static object[][] SourceField =
+        {
+            new object[] { 1, 2 },
+            new object[] { 3, 4 },
+            new object[] { 5, 6 },
+            new object[] { 7, 8 },
+        };
+
+        protected static object[][] SourceProperty => new[]
+        {
+            new object[] { 1, 2 },
+            new object[] { 3, 4 },
+            new object[] { 5, 6 },
+            new object[] { 7, 8 },
+        };
+
+        protected static object[][] SourceMethod() => new[]
+        {
+            new object[] { 1, 2 },
+            new object[] { 3, 4 },
+            new object[] { 5, 6 },
+            new object[] { 7, 8 },
+        };
+
+        protected static object[][] SourceMethodWithParameters(int x, int y) => new[]
+        {
+            new object[] { x + 1, y + 2 },
+            new object[] { x + 3, y + 4 },
+            new object[] { x + 5, y + 6 },
+            new object[] { x + 7, y + 8 },
+        };
+
+        protected static object[] SingleParameterSource = { 1, 2, 3, 4 };
+
+        protected static object[][] DifferentTypesSource =
+        {
+            new object[] { Visibility.Visible, FillDirection.Horizontal, false },
+            new object[] { Visibility.Hidden, FillDirection.Vertical, true },
+            new object[] { Visibility.Hidden, FillDirection.Full, false },
+        };
+
+        public static object[][] ExposedSourceField = SourceField;
+
+        public static object[][] ExposedSourceProperty => SourceProperty;
+
+        public static object[][] ExposedSourceMethod => SourceMethod();
+
+        [TestCaseSource(nameof(SourceField))]
+        public void TestSourceField(int a, int b)
+        {
+        }
+
+        [TestCaseSource(nameof(SourceProperty))]
+        public void TestSourceProperty(int a, int b)
+        {
+        }
+
+        [TestCaseSource(nameof(SourceMethod))]
+        public void TestSourceMethod(int a, int b)
+        {
+        }
+
+        [TestCaseSource(nameof(SourceMethodWithParameters), new object[] { 10, 20 })]
+        public void TestSourceParamsMethod(int a, int b)
+        {
+        }
+
+        [TestCaseSource(nameof(SourceField))]
+        [TestCaseSource(nameof(SourceProperty))]
+        [TestCaseSource(nameof(SourceMethod))]
+        [TestCaseSource(nameof(SourceMethodWithParameters), new object[] { 10, 20 })]
+        public void TestMultipleSources(int a, int b)
+        {
+        }
+
+        [TestCaseSource(nameof(SingleParameterSource))]
+        public void TestSingleParameterSource(int x)
+        {
+        }
+
+        [TestCaseSource(nameof(DifferentTypesSource))]
+        public void TestDifferentTypesSource(Visibility a, FillDirection b, bool c)
+        {
+        }
+
+        [TestCaseSource(typeof(TestEnumerable))]
+        public void TestCustomEnumerableSource(int a, int b)
+        {
+        }
+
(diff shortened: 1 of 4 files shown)


osu.Framework/Testing/TestBrowser.cs
```diff
@@ -590,14 +622,63 @@ void addSetUpSteps()
...
+
+            if (tcs.SourceName == null)
+                return (IEnumerable)Activator.CreateInstance(sourceDeclaringType);
+
+            var sourceMembers = sourceDeclaringType.AsNonNull().GetMember(tcs.SourceName, BindingFlags.Static | BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.FlattenHierarchy);
+            if (sourceMembers.Length == 0)
+                throw new InvalidOperationException($"No such source member exists with the name {tcs.SourceName} in {sourceDeclaringType} or its base types.");
+
+            if (sourceMembers.Length > 1)
+                throw new NotSupportedException($"There are multiple members with the same source name ({tcs.SourceName}) (e.g. method overloads).");
```
frenzibyte, line 646: FWIW NUnit already chokes on this scenario, so decided to just not support it for the time being.

osu.Framework/Testing/TestBrowser.cs
```diff
@@ -590,14 +622,63 @@ void addSetUpSteps()
...
+        {
+            var sourceDeclaringType = tcs.SourceType ?? testMethod.DeclaringType;
+            Debug.Assert(sourceDeclaringType != null);
+
+            if (tcs.SourceName == null)
+                return (IEnumerable)Activator.CreateInstance(sourceDeclaringType);
+
+            var sourceMembers = sourceDeclaringType.AsNonNull().GetMember(tcs.SourceName, BindingFlags.Static | BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.FlattenHierarchy);
+            if (sourceMembers.Length == 0)
+                throw new InvalidOperationException($"No such source member exists with the name {tcs.SourceName} in {sourceDeclaringType} or its base types.");
```
bdach, line 643:
```suggestion
                throw new InvalidOperationException($"No static member with the name of {tcs.SourceName} exists in {sourceDeclaringType} or its base types.");
```

osu.Framework/Testing/TestBrowser.cs
```diff
@@ -590,14 +622,63 @@ void addSetUpSteps()
...
             }
         }
 
+        private static IEnumerable getTestCaseSourceValue(MethodInfo testMethod, TestCaseSourceAttribute tcs)
+        {
+            var sourceDeclaringType = tcs.SourceType ?? testMethod.DeclaringType;
+            Debug.Assert(sourceDeclaringType != null);
+
+            if (tcs.SourceName == null)
+                return (IEnumerable)Activator.CreateInstance(sourceDeclaringType);
```
bdach, line 639: This is a detail, but as per spec, this should only be done if `tcs.SourceType` is set.

osu.Framework/Testing/TestBrowser.cs
```diff
@@ -590,14 +622,63 @@ void addSetUpSteps()
...
+            if (sourceMembers.Length > 1)
+                throw new NotSupportedException($"There are multiple members with the same source name ({tcs.SourceName}) (e.g. method overloads).");
+
+            var sourceMember = sourceMembers.Single();
+
+            switch (sourceMember)
+            {
+                case FieldInfo sf:
+                    if (!sf.IsStatic)
+                        throw new InvalidOperationException($"The source field {sf.Name} in {sf.DeclaringType.ReadableName()} must be static.");
```
bdach, line 654: Shouldn't be possible. Binding flags already specified static.

Goes everywhere lower down as well.

osu.Framework/Testing/TestBrowser.cs
```diff
@@ -590,14 +622,63 @@ void addSetUpSteps()
...
+
+                    return (IEnumerable)sp.GetValue(null);
+
+                case MethodInfo sm:
+                    if (!sm.IsStatic)
+                        throw new InvalidOperationException($"The source method {sm.Name} in {sm.DeclaringType.ReadableName()} must be static.");
+
+                    var methodParamsLength = sm.GetParameters().Length;
+                    if (methodParamsLength != (tcs.MethodParams?.Length ?? 0))
+                        throw new InvalidOperationException($"The given source method parameters count doesn't match the method. ({tcs.MethodParams?.Length ?? 0} != {methodParamsLength})");
```
bdach, line 673:
```suggestion
                        throw new InvalidOperationException($"The given source method parameters count doesn't match the method. (attribute has {tcs.MethodParams?.Length ?? 0}, method has {methodParamsLength})");
```

osu.Framework/Testing/TestBrowser.cs
```diff
@@ -590,14 +622,63 @@ void addSetUpSteps()
...
+
+                    return (IEnumerable)sm.Invoke(null, tcs.MethodParams);
+
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(sourceMember));
```
bdach, line 678: `NotSupportedException` seems more apt. Something like the following, maybe:

```csharp
                default:
                    throw new NotSupportedException($"{sourceMember.MemberType} is not a supported member type for {nameof(TestCaseSourceAttribute)} (must be static field, property or method)");
```

osu.Framework.Tests/Visual/Testing/TestSceneTestWithSource.cs
```diff
@@ -0,0 +1,113 @@
...
+    {
+        protected static object[][] SourceField =
+        {
+            new object[] { 1, 2 },
+            new object[] { 3, 4 },
+            new object[] { 5, 6 },
+            new object[] { 7, 8 },
+        };
+
+        protected static object[][] SourceProperty => new[]
```
bdach, line 20: This is a bit weird. I'd either use different data for field/property/method, or just proxy to the field instead of redeclaring the same literal array three times.

osu.Framework.Tests/Visual/Testing/TestSceneTestWithSource.cs
```diff
@@ -0,0 +1,113 @@
...
+            new object[] { Visibility.Visible, FillDirection.Horizontal, false },
+            new object[] { Visibility.Hidden, FillDirection.Vertical, true },
+            new object[] { Visibility.Hidden, FillDirection.Full, false },
+        };
+
+        public static object[][] ExposedSourceField = SourceField;
+
+        public static object[][] ExposedSourceProperty => SourceProperty;
+
+        public static object[][] ExposedSourceMethod => SourceMethod();
```
bdach, line 57: Is a property, not a method.

bdach (dismissed):

osu.Framework.Tests/Visual/Testing/TestSceneTestWithSource.cs
frenzibyte: Sure.

osu.Framework.Tests/Visual/Testing/TestSceneTestWithSource.cs
frenzibyte: That's a typo..

osu.Framework/Testing/TestBrowser.cs
frenzibyte: This case should only be reached if the member type is a weird one that's not a field/property/method, when the member is not found, it'll throw on here:
https://github.com/ppy/osu-framework/blob/8ad79257873da9d3f86d18d7eb0d0ab31a11a9d9/osu.Framework/Testing/TestBrowser.cs#L642-L643

Maybe should reword it a little to include members that aren't static, `Make sure the source member is static.`?

frenzibyte: Actually nevermind, you're already suggesting it above.

smoogipoo (approved):
lgtm

PR #4461, opened 2021-05-24
Populate connected controllers after SDL init.

miterosan:
Allows usage of joysticks and controllers which are already connected on startup.

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -385,6 +385,8 @@ namespace osu.Framework.Platform
                 updateCursorVisibility(!evt.NewValue.HasFlagFast(CursorState.Hidden));
                 updateCursorConfined(evt.NewValue.HasFlagFast(CursorState.Confined));
             };
+
+            populateJoysticks();
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -745,6 +747,24 @@ private void handleControllerButtonEvent(SDL.SDL_ControllerButtonEvent evtCbutto
         private void handleControllerAxisEvent(SDL.SDL_ControllerAxisEvent evtCaxis) =>
             enqueueJoystickAxisInput(((SDL.SDL_GameControllerAxis)evtCaxis.axis).ToJoystickAxisSource(), evtCaxis.axisValue);
 
+        /// <summary>
+        /// Populates <see cref="controllers"/> with joysticks that are already connected.
+        /// </summary>
+        private void populateJoysticks()
+        {
+            for (int i = 0; i < SDL.SDL_NumJoysticks(); i++)
```
peppy, line 755: Why is this logic added with zero consideration to the fact it's duplicating existing logic?

https://github.com/ppy/osu-framework/blob/569503aa312c19ed8d1584a06f431206af74a1b2/osu.Framework/Platform/SDL2DesktopWindow.cs#L753-L760

peppy (changes requested):

osu.Framework/Platform/SDL2DesktopWindow.cs
miterosan: Please explain how you mean this.
Do you expect the logic above to be separated into its own method? (So that the method for adding a controller/joystick is used at 3 locations?)

peppy: Yes, I'd hope the same logic is combined into a `private` method as we do everywhere else.

miterosan: unified

smoogipoo (approved):
Looks to work as expected

PR #4462, opened 2021-05-25
Flush framebuffer after FBO bind on macOS

smoogipoo:
Maybe it's fine to get this in for further testing? Seems like Chromium has experienced similar crashes/full system lockups as a result of FBOs and integrated this fix themselves:
http://crbug.com/1181068
http://crbug.com/783979
https://github.com/google/angle/commit/ce89d99fdeda61d9bc88fc7abd2aa3b4666d770e


peppy:
Can you add an inline comment mentioning the issue links?

peppy (approved):

PR #4463, opened 2021-05-25
Fix default audio device re-initting when devices change

smoogipoo:
This probably _fixes_ https://github.com/ppy/osu/issues/12638 and https://github.com/ppy/osu/issues/12482

The default value here is `string.Empty`, so this check would always pass whenever the audio devices changed.

### DIFF
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -353,7 +353,7 @@ namespace osu.Framework.Audio
         protected virtual bool IsCurrentDeviceValid()
         {
             var device = audioDevices.ElementAtOrDefault(Bass.CurrentDevice);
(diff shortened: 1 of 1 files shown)


peppy:
Was planning on trying to repro this one today. Before I look into it, did you try reproducing inside a VM and if so, was it successful?

smoogipoo:
I didn't try reproing in a VM, I reproed on my desktop because joystick devices don't seem to work correctly - my PS4 controller keeps disconnecting and reconnecting and adds an audio device of its own.

peppy (approved):

peppy:
Doesn't look to resolve ppy/osu#12638, but also doesn't seem to break anything and looks sane so will merge this as-is. Unsure about https://github.com/ppy/osu/issues/12482 so best follow up with the user in that thread once we get a release out.

peppy (approved):

PR #4469, opened 2021-05-26
Fix audio devices not working after being disconnected then reconnected

peppy:
This is a speculative fix that should be mergeable if we want to get things back on track. Needs further investigation into whether this is expected behaviour on bass's end or not.

This fixes the case where:
- The user disconnects a device, causing the windows default to change to another existing device. They then reconnect their preferred device and windows switches back.
- The user disconnects a device, and it is the last available device (causing bass to enter virtual device mode). They then reconnect the device and it becomes available for use again.

Closes ppy/osu#12638.

### DIFF
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -266,6 +266,11 @@ namespace osu.Framework.Audio
         }
 
         /// <summary>
+        /// Lock globally across all usages ensure no funny-business occurs during device free/init.
+        /// </summary>
+        private static object bass_init_lock = new object();
+
+        /// <summary>
         /// This method calls <see cref="Bass.Init(int, int, DeviceInitFlags, IntPtr, IntPtr)"/>.
         /// It can be overridden for unit testing.
         /// </summary>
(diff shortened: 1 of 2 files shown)


osu.Framework/Audio/AudioManager.cs
```diff
@@ -293,13 +298,23 @@ protected virtual bool InitBass(int device)
...
+                didInit = Bass.Init(device);
+
+                // If the device was already initialised, the device can be used without much fuss.
+                if (Bass.LastError == Errors.Already)
+                {
+                    Bass.CurrentDevice = device;
+
+                    // Without this call, on windows, a device which is disconnected then reconnected will look initialised
+                    // but not work correctly in practice.
+                    Bass.Free();
```
smoogipoo, line 314: As I said, you can't do this as is. This needs to be properly linked with the cleanup in `AudioThread` because there's a lot of potential for this to go wrong:
- This doesn't always work on linux.
- `AudioThread` may crash on exit if it then tries to switch to the device. For example if this frees but the subsequent init fails, because the device is still in the `AudioThread`'s initialised devices list.

smoogipoo (changes requested):

osu.Framework/Audio/AudioManager.cs
peppy: Can you explain the "doesn't always work" part? Is that documented somewhere?

I can add a platform check for now if it only affects linux.

smoogipoo: It's documented in `AudioThread`

peppy: How does the latest commit look in that case? We may also need to have the lock cover `Bass.Init` calls still (maybe moving the init method to `AudioThread` as well?)

smoogipoo: I would remove the lock completely actually - all of this code is running on the single audio thread. The only case that it might break is if multiple gamehosts are run, but I'm not even going to pretend that won't break anywhere else (like say, `ThreadSafety` for instance).

peppy: Fair enough. I was aiming to handle the scenario you mentioned of multiple test runners, but you'd know better than me if that is not a requirement.

osu.Framework/Threading/AudioThread.cs
```diff
@@ -110,12 +112,14 @@ protected override void PerformExit()
             // Safety net to ensure we have freed all devices before exiting.
             // This is mainly required for device-lost scenarios.
             // See https://github.com/ppy/osu-framework/pull/3378 for further discussion.
-            foreach (var d in initialisedDevices)
-                freeDevice(d);
+            foreach (var d in initialised_devices)
+                FreeDevice(d);
```
smoogipoo, line 116: This probably needs to be `ToArray()`d because the hashset is modified inside the call?

smoogipoo (changes requested):
Seems fine otherwise

osu.Framework/Threading/AudioThread.cs
smoogipoo: Looks like this doesn't throw, but I'd still like a `.ToArray()` here as that's a pretty wild .NET inconsistency >_>

smoogipoo (approved):

PR #4471, opened 2021-05-27
Fix KeyBindingContainer using InputState to build combinations

smoogipoo:
Fixes https://github.com/ppy/osu/issues/12936

`KeyBindingContainer` previously built the key combination directly from the `InputState`, not caring whether the event which changed the state ever reached the KBC. I identified two issues as a result of this:
1. Triggers the wrong binding when the KBC becomes alive.
    1. The KBC has bindings `{ A, B }, { B }`.
    1. `A` is held down.
    2. The KBC becomes alive.
    3. The user presses `B`.
    4. The binding for combination `{ A, B }` is triggered when only the one for `{ B }` is expected.
2. Triggers when other `Drawable`s handled input.
    1. The KBC has a binding for `{ A }`.
    2. `A` is held down.
    3. Another `Drawable` in the hierarchy handling input before the KBC handles `A` (never reaches the KBC).
    4. The user presses `B`, which reaches the KBC.
    5. The KBC triggers the binding for `{ A }`.

Also, in (2) above, the trace assertion from `handleNewReleased` would trigger if `B` is released as `B` is not a part of the combination forming the binding. I believe the assertion is just incorrect as I previously mentioned in https://github.com/ppy/osu-framework/pull/4465, so just removed it.

### DIFF
--- a/osu.Framework/Input/Bindings/KeyBindingContainer.cs
+++ b/osu.Framework/Input/Bindings/KeyBindingContainer.cs
@@ -2,14 +2,12 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using osu.Framework.Allocation;
 using osu.Framework.Extensions.IEnumerableExtensions;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Input.Events;
-using osu.Framework.Input.States;
 using osu.Framework.Logging;
 using osu.Framework.Platform;
 using osuTK;
@@ -100,17 +98,20 @@ namespace osu.Framework.Input.Bindings
             return true;
         }
 
+        /// <summary>
+        /// All input keys which are currently pressed and have reached this <see cref="KeyBindingContainer"/>.
+        /// </summary>
+        private readonly HashSet<InputKey> pressedInputKeys = new HashSet<InputKey>();
+
         protected override bool Handle(UIEvent e)
         {
-            var state = e.CurrentState;
-
             switch (e)
             {
                 case MouseDownEvent mouseDown:
-                    return handleNewPressed(state, KeyCombination.FromMouseButton(mouseDown.Button), false);
+                    return handleNewPressed(KeyCombination.FromMouseButton(mouseDown.Button), false);
 
                 case MouseUpEvent mouseUp:
-                    handleNewReleased(state, KeyCombination.FromMouseButton(mouseUp.Button));
+                    handleNewReleased(KeyCombination.FromMouseButton(mouseUp.Button));
                     return false;
 
                 case KeyDownEvent keyDown:
@@ -119,7 +120,7 @@ namespace osu.Framework.Input.Bindings
 
                     foreach (var key in KeyCombination.FromKey(keyDown.Key))
                     {
-                        if (handleNewPressed(state, key, keyDown.Repeat))
+                        if (handleNewPressed(key, keyDown.Repeat))
                             return true;
                     }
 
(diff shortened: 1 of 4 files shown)


osu.Framework/Input/Bindings/KeyBindingContainer.cs
```diff
@@ -119,7 +120,7 @@ protected override bool Handle(UIEvent e)
 
                     foreach (var key in KeyCombination.FromKey(keyDown.Key))
                     {
-                        if (handleNewPressed(state, key, keyDown.Repeat))
+                        if (handleNewPressed(key, keyDown.Repeat))
```
bdach, line 123: Creating the set of pressed keys from incoming events rather than relying on the full state breaks this in the case of modifiers, because they actually fire two keys and not one, so if the first one is handled, the second one won't get handled. Tests are failing because of this, but the simplest repro using `TestSceneKeyBindingsGrid` is:

1. Press <kbd>LShift</kbd> and then <kbd>LAlt</kbd> - the `Alt and LShift` action won't fire
2. Press them in the opposite order - the action will fire

This is because <kbd>Shift</kbd> precedes <kbd>LShift</kbd> and <kbd>Alt</kbd> precedes <kbd>LAlt</kbd>. In case (1), <kbd>Shift</kbd> firing first causes that particular `Alt and LShift` binding to not receive the <kbd>LShift</kbd> (as there are other bindings handling plain <kbd>Shift</kbd>), while in case (2) the <kbd>Alt</kbd> handling fires and succeeds first, and on the <kbd>LShift</kbd> press, the <klbd>Shift</kbd> case also fires first but fails to match the binding.

bdach (changes requested):

osu.Framework/Input/Bindings/KeyBindingContainer.cs
smoogipoo: I don't think that's incorrect from a real-world scenario POV. If shift is handled, then I wouldn't expect pressing alt to then fire the alt+lshift binding. Am I wrong in that?

bdach: I think the original behaviour made more sense to me, but it is a bit confusing to figure out which semantics we're after, as we have both keybinding matching modes and "meta"-keys (as in, a <kbd>Alt</kbd> key that is supposed to encompass <kbd>LAlt</kbd> and <kbd>RAlt</kbd>) and the intersection of those gets a little fuzzy.

smoogipoo: @peppy can you explain to me what the following code is doing?
https://github.com/ppy/osu-framework/blob/bf3b398128562782bd822a1d47fd468eb323d683/osu.Framework/Input/Bindings/KeyCombination.cs#L94-L101

And specifically, why it's not using `ContainsKeyPermissive()`? I'm struggling really hard with these parameter names and everything, and the comment isn't helping.

peppy: So first of all, I think the comment there is outdated and completely incorrect. This code is no longer an optimisation but actual required logic (since it is being used in every case below). To ensure "equality" we are doing a `Contains` check in both directions; this is one of those two directions.

As for why it's not using `ContainsKeyPermissive`:

The fail case is where the candidate is `LShift` and the user pressed `RShift`. Firstly, when a user pressed `RShift`, `Shift` is also included automatically:

https://github.com/ppy/osu-framework/blob/bfb60d778c86ead8499e66c2d28a4d525173ded7/osu.Framework/Input/Bindings/KeyCombination.cs#L552-L554

So for the pre-check (the code you are linking above), when comparing the candidate (binding) vs the user pressed keys, if we were to use the `Permissive` version of the function the user's `Shift` would incorrect match against `LShift` in the binding/candidate. This is something we never want, regardless of the matching mode.

Thinking about this another way: the non-direction modifier specifications are there for the benefit of creating bindings that support both left and right. The "permissive"ness of the match should only be considered when checking user pressed keys vs the bindings, not the other way around.

Let me know if this still doesn't make sense; I can try rephrasing in another way. Safe to say the comment should be removed and probably replaced with an example of why it's done this way (similar to what I've tried to do in this comment) for future visitors.

peppy:
I'm surprised your changes work, but they seem to. To save me going through this 10 times can you explain why this works even without the extra (non-directional) specifications?

peppy:
Never mind, I've gone through the logic and all seems to check out.

Those `ContainsKeys` methods are still really hard to mentally parse. Might be better if one didn't call the other... not sure (just inline the cases?).

I've added some inline commenting, please check it makes sense.

peppy (approved):

smoogipoo:
Sorry, I wanted to see if tests passed on CI first, because they passed locally and usages osu!-side seemed to pass too.

I noticed that if the candidate binding is `{ Ctrl + A }`, `ContainsKey()` matches `LCtrl`, `RCtrl`. I'm still trying to justify why this was required before.

smoogipoo:
There's another thing I want to check/test, I believe https://github.com/ppy/osu-framework/blob/2238e69ab20569015ff009b8494f775d1e1fa7ca/osu.Framework/Input/Bindings/KeyBindingContainer.cs#L127-L128 may not be a thing anymore with these changes.

peppy:
Not sure what you're saying there, is that case broken somehow? Are you saying it matches even without holding `A`?

smoogipoo:
No, I've changed my comment.

peppy:
That's intentional behaviour no? The whole reason for adding the non-direction modifiers is to allow for bindings (candidates) which are direction agnostic, which is the exact behaviuor you describe. It allows `PlatformAction` default bindings to not be duplicated for each side modifier, for instance.

smoogipoo:
From a high level yes.

I'm speaking towards the fact that prior to this change, while `FromKey()` was still an array, the pressed combination would _always_ contain `{ Ctrl, LCtrl, X }` for example, which would always match a binding of `{ Ctrl, X }` regardless of how `ContainsKey()` is implemented.

With this change, the pressed combination only contains `{ LCtrl, X }`, so the only thing that makes this all work is the way `ContainsKey()` is implemented.

smoogipoo:
https://github.com/ppy/osu-framework/pull/4471#issuecomment-854474998

I don't know what this comment was all about, any test I write doesn't fail on master nor this branch so I assume it was fixed and not relevant anymore. Added a test and removed.

peppy (approved):

peppy:
@smoogipoo just to confirm (even though it seems like it has been done) you have changed the behaviour to match what @bdach commented on in https://github.com/ppy/osu-framework/pull/4471#discussion_r644746789 right?

smoogipoo:
Yes, behaviour is as it was originally. It was fixed by only handling the key-specific modifiers rather than the common ones as mentioned in the discussions above.

bdach (approved):
looks to be all good now :)

PR #4342, opened 2021-04-05
Fix relative mouse mode sometimes not activating when cursor enters window

Susko3:
Closes #4341 (or at least works around the described problem)

With this PR, the weirdness around the cursor entering the window isn't fixed, But at least `RelativeMouseMode` will be correctly set on the next `FeedbackMousePositionChange`.

Issue #4341, opened 2021-04-05
Relative mouse mode sometimes doesn't get enabled when cursor enters window


Susko3:
When the cursor enters the window, `RelativeMouseMode` sometimes won't get enabled even if all conditions for it are met. This happens because `updateRelativeMode()` stops being called.

**To reproduce:**
Set your mouse report rate to 1000 Hz (or something higher that 125 Hz). Enable Relative mode in the InputManager visual test, increase (or decrease) sensitivity. Move your cursor in and out of the window. At some point, it'll stop applying the set sensitivity. This is because relative mouse mode is disabled.

Relevant lines of code:
https://github.com/ppy/osu-framework/blob/07cef87ebbb592e97e1b965b3f4147da03570352/osu.Framework/Input/Handlers/Mouse/MouseHandler.cs#L104-L136

Here is what happens when the bug occurs:
1. mouse enters window bounds -> `cursorInWindow` changed to `true` -> `updateRelativeMode()` gets called -> `window.RelativeMouseMode` is set to `true`
2. the next `FeedbackMousePositionChange` that gets called has negative `position.X`
   - mouse move events processed by `FeedbackMousePositionChange` lag ~3 events behind the latest mouse move event (more info in Remarks below), so it's calculating as if the cursor is still outside the window according to `position`, but inside the window according to `cursorInWindow`
3. `positionOutsideWindow` is true -> `window.RelativeMouseMode` set to `false`
4. After that `updateRelativeMode()` never gets called again, and `RelativeMouseMode` stays disabled when it should be enabled (`cursorInWindow` == `true`)

**Remarks:**
Happens more frequently if the mouse report rate (I set mine trough Logitech G Hub) is set to higher value. Setting to 1000 Hz makes it very frequent. Happens at 500 and 250 Hz, but less frequently. Couldn't get a repro with 125 Hz.

While debugging and testing this, I discovered that the mouse move events handled by `FeedbackMousePositionChange` can lag up to 5 events behind the latest `SDL_MOUSEMOTION` with 1000 Hz report rate. With 125 Hz, the event processed by `FeedbackMousePositionChange` is the latest `SDL_MOUSEMOTION`.

Sample log, with mouse at 1000 Hz report rate, and logging every 200th `FeedbackMousePositionChange`:
```
INFO: SDL EVENT: SDL_MOUSEMOTION (timestamp=479367 windowid=1 which=0 state=0 x=572 y=232 xrel=-7 yrel=2)
[...]

PR #4472, opened 2021-05-28
Disable relative mode only when leaving the window

Susko3:
Currently, the flow that disables relative mode (based on whether the cursor is in the window) sometimes disables relative mode when the cursor is entering the window.

This PR adds a check so that the disable relative mode flow is only triggered if the cursor is leaving the window.

Closes https://github.com/ppy/osu-framework/issues/4341
Closes PR https://github.com/ppy/osu-framework/pull/4342


peppy (approved):
Seems to work well in my testing. Much nicer solution 👍🏻 .

PR #4474, opened 2021-06-01
Fix samples not reloading if loaded too early

smoogipoo:
Fixes https://github.com/ppy/osu/issues/13266

If samples are loaded before `Bass.Init()`, `SampleLoad` fails with `Errors.Init`. In this case, and only this case, I want to reload the sample when a device is received afterwards.

I've added some safeguards in the hopes that it won't, but if the test fails intermittently then I can't see a good way to test this other than to ignore the test.


peppy (dismissed):

peppy:
Code itself looks fine, not sure about the test (see CI).

peppy (approved):

PR #4490, opened 2021-06-08
Avoid changing windows time period when focus changes

peppy:
In an attempt to figure out what is causing https://github.com/ppy/osu/issues/12702 we looked at all the operations we are performing on window active state change. This is the one that stood out the most.

Since this is a hard one to reproduce (does not affect all users), just going to get this out for user testing. Thinking about it, this could only be affecting users that have no other applications system-wide requesting a time period elevation. In general this is pretty rare as even Chrome will request it.

For reference, we don't do this in stable, and there is little reason to reset it on active state change (although it may be something we would consider if the game is ever minimising to system tray and expected to use zero resources).

### DIFF
--- a/osu.Framework/Platform/Windows/WindowsGameHost.cs
+++ b/osu.Framework/Platform/Windows/WindowsGameHost.cs
@@ -59,7 +59,7 @@ namespace osu.Framework.Platform.Windows
             // OnActivate / OnDeactivate may not fire, so the initial activity state may be unknown here.
             // In order to be certain we have the correct activity state we are querying the Windows API here.
 
-            timePeriod = new TimePeriod(1) { Active = true };
+            timePeriod = new TimePeriod(1);
         }
 
         protected override IWindow CreateWindow() => new WindowsWindow();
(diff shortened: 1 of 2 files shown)


osu.Framework/Platform/Windows/WindowsGameHost.cs
```diff
@@ -59,7 +59,7 @@ protected override void SetupForRun()
             // OnActivate / OnDeactivate may not fire, so the initial activity state may be unknown here.
             // In order to be certain we have the correct activity state we are querying the Windows API here.
 
-            timePeriod = new TimePeriod(1) { Active = true };
+            timePeriod = new TimePeriod(1);
```
smoogipoo, line 62: As it is, I don't think there's anything that enables it anymore. The static ctor only gets the capabilities.

smoogipoo (changes requested):

peppy:
Tested on windows to work as expected.

smoogipoo (approved):

PR #4492, opened 2021-06-08
Fix Markdown container crashing on absolute URIs with no authority part

bdach:
...Which means URIs with a colon after the scheme name but no following double-slashes, and is basically an euphemism for `mailto:` URIs. I can't really imagine ever hitting a case of this with another URL scheme.

Fixed by applying [the first workaround listed in mono docs](https://www.mono-project.com/docs/faq/known-issues/urikind-relativeorabsolute/) more directly. I have tested on Android to confirm that doing so does not regress #4483, and it seems that it doesn't, but may be safer to double-check iOS, too.

Unsure of the placement of the new method, but `Validation` as an existing static class seemed fitting enough.

Closes https://github.com/ppy/osu/issues/13386.

### DIFF
--- a/osu.Framework/Utils/Validation.cs
+++ b/osu.Framework/Utils/Validation.cs
@@ -44,7 +44,7 @@ namespace osu.Framework.Utils
         public static bool TryParseUri(string uriString, [NotNullWhen(true)] out Uri? result)
         {
 #pragma warning disable RS0030 // Bypassing banned API check, as it'll actually be used properly here
-            UriKind kind = uriString.StartsWith("/", StringComparison.Ordinal) ? UriKind.Relative : UriKind.RelativeOrAbsolute;
+            UriKind kind = uriString.StartsWith('/') ? UriKind.Relative : UriKind.RelativeOrAbsolute;
 #pragma warning restore RS0030
             return Uri.TryCreate(uriString, kind, out result);
         }
--- a/osu.Framework/Utils/Validation.cs
+++ b/osu.Framework/Utils/Validation.cs
@@ -1,8 +1,12 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
+using System.Diagnostics.CodeAnalysis;
 using osuTK;
 using osu.Framework.Graphics;
+
+#nullable enable
 
 namespace osu.Framework.Utils
 {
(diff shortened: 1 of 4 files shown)


osu.Framework/Utils/Validation.cs
```diff
@@ -23,5 +27,26 @@ public static class Validation
...
+        /// This method uses the .NET behaviour.
+        /// For more info, see <a href="https://www.mono-project.com/docs/faq/known-issues/urikind-relativeorabsolute/">Mono documentation</a>.
+        /// </remarks>
+        /// <param name="uriString">The string representation of the URI to parse.</param>
+        /// <param name="result">The resultant parsed URI, if parsing succeeded.</param>
+        /// <returns><see langword="true"/> if parsing succeeded; <see langword="false"/> otherwise.</returns>
+        public static bool TryParseUri(string uriString, [NotNullWhen(true)] out Uri? result)
+        {
+#pragma warning disable RS0030 // Bypassing banned API check, as it'll actually be used properly here
+            UriKind kind = uriString.StartsWith("/", StringComparison.Ordinal) ? UriKind.Relative : UriKind.RelativeOrAbsolute;
```
peppy, line 47: I don't believe this will work as we have cases where the url doesn't start with a `/` where we expect the lookups to be relative to the current wiki path (that's why I didn't use this method in the first place).

Needs to be checked against said cases locally at very least.

bdach: As long as it doesn't have a scheme at the start, it should still parse as relative I think? Not sure which cases would this encompass really. If you can list them I'll expand test coverage because that's apparently untested behaviour.

bdach: FWIW I only went this way because the only viable alternative that I can see that would be more in line with how the previous version was written is just checking all possible schemes, which starts sounding like reimplementing URI parsing which I'd rather not get into in fear of missing even more edge cases.

peppy: I think the case is linking images in the same path as the current wiki page.

20210609 131311 (JetBrains Rider-EAP)

It's this specific line here; not sure if we already have test coverage of its usage (where not just `RootUrl` but also `DocumentUrl` are available).

bdach: I just checked and there is a pre-existing test case that I *think* should be covering this already:

https://github.com/ppy/osu-framework/blob/81121e80e60166aa2f06fd7e3d97fd05c93c4dcd/osu.Framework.Tests/Visual/UserInterface/TestSceneMarkdownContainer.cs#L210-L223

I haven't ran it on mobile but it passes on desktop at least with these changes.

turbedi, line 47:
```suggestion
            UriKind kind = uriString.StartsWith('/') ? UriKind.Relative : UriKind.RelativeOrAbsolute;
```

peppy: Ah yep, that should cover this fine. I guess the refactoring I did around that logic was enough to preserve this behaviour.

peppy (approved):

PR #4494, opened 2021-06-09
Display border and edge effect for drawables with negative size

ekrctb:
Negative size means the rectangle has either `left > right` or `top > bottom` in coordinates.
It is displayed as if the rectangle is flipped to the congruent positive size rectangle.

- Closes ppy/osu#13406

(About commit history) I first tried to modify shader code but then I ended up with a simpler way by always normalizing the masking rectangle.


osu.Framework/Graphics/Primitives/RectangleF.cs
```diff
@@ -375,6 +375,11 @@ public RectangleF RelativeIn(RectangleF other)
             return new RectangleF((X - other.X) / other.Width, (Y - other.Y) / other.Height, scaleX, scaleY);
         }
 
+        /// <summary>
+        /// Create a new <see cref="RectangleF"/> congruent to this rectangle but with non-negative <see cref="Width"/> and <see cref="Height"/>.
+        /// </summary>
+        public RectangleF Normalize() => new RectangleF(Math.Min(Left, Right), Math.Min(Top, Bottom), Math.Abs(Width), Math.Abs(Height));
```
bdach, line 381: I do wonder if it might be worth dropping a `[MethodImpl(MethodImplOptions.AggressiveInlining)]` here, since it seems likely that this may end up being one of the hotter paths out there (although hopefully not that hot, thanks to invalidation). Might be worth benchmarking the performance hit (if any) associated with this change.

peppy (approved):

osu.Framework/Graphics/Primitives/RectangleF.cs
smoogipoo: This isn't going to be hot enough to matter, it'll be a difference of a few nanoseconds at most in the grand scheme of things and shouldn't block this PR.
I'd rather leave the JIT to determine whether to inline itself, because there can be disadvantages to inlining (from a jit compilation standpoint) and you can get trapped down a micro-benchmarking hole.

smoogipoo:
image

This blackness when flipped looks to be the wrapping mode. The following change fixes it:
```diff
diff --git a/osu.Framework.Tests/Visual/Containers/TestSceneSizing.cs b/osu.Framework.Tests/Visual/Containers/TestSceneSizing.cs
index 529af6b21f..2b033cc722 100644
--- a/osu.Framework.Tests/Visual/Containers/TestSceneSizing.cs
+++ b/osu.Framework.Tests/Visual/Containers/TestSceneSizing.cs
@@ -5,8 +5,11 @@
 using osu.Framework.Extensions.Color4Extensions;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.OpenGL.Textures;
+using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
+using osu.Framework.Graphics.Textures;
 using osu.Framework.Input.Events;
 using osu.Framework.Utils;
 using osuTK;
@@ -1166,10 +1169,11 @@ protected override void OnDrag(DragEvent e)
 
         public InfofulBox()
         {
-            Add(new Box
+            Add(new Sprite
             {
                 RelativeSizeAxes = Axes.Both,
                 Depth = float.MaxValue,
+                Texture = Texture.WhitePixel.Crop(new RectangleF(0, 0, 1, 1), Axes.X, WrapMode.ClampToEdge, WrapMode.ClampToEdge)
             });
         }
     }
```
Is this known?

ekrctb:
Actually, the blackness issue is also reproducible in master.
I cannot figure out what is causing the issue: the blackness is also gone when a texture ("sample-texture") is used instead of the `WhitePixel`, or even using `Texture.WhitePixel.Crop(new RectangleF(0, 0, 1, 1))` with the `None` wrapping modes.

The issue doesn't affect the current osu-side use-case of border-only drawables (`box.Colour = Color4.Transparent`).
Therefore, I propose this PR be merged as is, potentially adding a comment to the test scene, unless a proper fix can be figured out quickly.

smoogipoo:
Ooh fun. The blackness is caused by the front to back pass. I've got a fix but will merge as is.

smoogipoo (approved):

PR #4500, opened 2021-06-10
Improve safety of `GameThread.Running` (and make `GameThread.Start` blocking)

peppy:
Noticed in passing that there's a potential for failure with the way this flag is handled: the thread could be `!IsAlive` for a brief moment before it is nulled. Calls to `Start` are checking for null, which may result in an invalid operation if the previous thread had not been nulled out yet.

https://github.com/ppy/osu-framework/blob/7d10a151ae4230e595fa912f64bb16761e9b48f1/osu.Framework/Threading/GameThread.cs#L250-L254

I have not reproduced this one, nor was it reproduced anywhere. Just seemed a bit dodgy..

This also makes `GameThread.Start` blocking, to give it symmetry with `GameThread.Pause` (which blocks in a similar fashion).

### DIFF
--- a/osu.Framework/Threading/GameThread.cs
+++ b/osu.Framework/Threading/GameThread.cs
@@ -111,8 +111,6 @@ namespace osu.Framework.Threading
             Debug.Assert(Thread == null);
             Debug.Assert(!Running);
 
-            Running = true;
-
             Thread = new Thread(runWork)
             {
                 Name = PrefixedThreadNameFor(Name),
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):

peppy:
I've also fixed the issue seen in https://github.com/ppy/osu-framework/pull/4489/checks?check_run_id=2790895851 via https://github.com/ppy/osu-framework/pull/4500/commits/817d7f5493a86eb1a3f79b59b112b614de234afa. Was going to make it a separate pull request but it turns out that it ties in nicely with the `Running` property change.

smoogipoo (approved):

PR #4505, opened 2021-06-10
Add extra safety over thread start/pause logic

peppy:
Hopefully covers remaining test failure issues. The main resolution here is ensuring that `Thread` is `null` before `Running` is updated, but the lock has been added for extra safety, as we should never be starting/pausing a `GameThread` from different threads concurrently.

### DIFF
--- a/osu.Framework/Threading/GameThread.cs
+++ b/osu.Framework/Threading/GameThread.cs
@@ -74,6 +74,11 @@ namespace osu.Framework.Threading
 
         private readonly object startStopLock = new object();
 
+        /// <summary>
+        /// Whether a pause has been requested.
+        /// </summary>
+        private bool pauseRequested;
+
         internal void Initialize(bool withThrottling)
         {
             MakeCurrent();
@@ -137,7 +142,7 @@ namespace osu.Framework.Threading
             {
                 Initialize(true);
 
-                while (!exitCompleted && !paused)
+                while (!exitCompleted && !pauseRequested)
                 {
                     ProcessFrame();
                 }
@@ -221,14 +226,16 @@ namespace osu.Framework.Threading
             Thread.CurrentUICulture = culture;
         }
 
-        private bool paused;
-
         public void Pause()
         {
-            if (Thread == null)
-                return;
+            lock (startStopLock)
+            {
+                if (Thread == null)
+                    return;
 
-            paused = true;
+                pauseRequested = true;
+            }
+
             while (Running)
                 Thread.Sleep(1);
         }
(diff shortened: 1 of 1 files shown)


osu.Framework/Threading/GameThread.cs
```diff
@@ -223,40 +225,47 @@ private void updateCulture()
...
+                    Cleanup();
+                }
             }
         }
 
         protected virtual void Cleanup()
         {
-            Running = false;
             Thread = null;
+            Running = false;
```
bdach, line 246: I think this should also take the lock? `runWork()` calls this, which I think is off-thread when compared to `Start()`/`Stop()` which are called by the main thread?

peppy: Yep, sounds correct.

osu.Framework/Threading/GameThread.cs
```diff
@@ -223,40 +225,50 @@ private void updateCulture()
 
         public void Pause()
         {
-            if (Thread != null)
+            lock (startStopLock)
```
bdach, line 228: As can be seen on CI (I killed the github actions runs manually, appveyor timed out on its own) this deadlocks now because `Pause()` is blocking...

In order to achieve full correctness, a reader-writer lock is needed I think? At the least the monitor should be exited in the blocking case.

peppy: I'd sooner just remove the lock than use a read write mess.

peppy: I've gone with removing the lock on the `Pause` portion, which seems the safest of the options to me.

bdach: I think the worst scenario that can plausibly happen is

```
Thread A							Thread B

Pause()
	if (Thread != null) // fails
									Start()
										paused = false;
										createThread();
										Thread.Start();
	Cleanup();
```

which would leave an orphaned thread running. May be safest to unconditionally set `paused = true` regardless of whether the thread exists to make sure it cleans itself up via the check in `runWork()`, but I have done 0 testing of this.

peppy: Nice catch. I've also implemented the result of yesterday's conversation, which is that we shouldn't need to be calling `Cleanup` from the `Pause` method.

bdach (approved):
I *think* this is ok now. But it's early in the day and threading is hard, so going to see if I can get a double-check on that.

smoogipoo (approved):
This looks fine to me

Issue #3134, opened 2019-12-22
Implement color picker


andy840119:
Is there any design for color picker?

smoogipoo:
There isn't a framework design for this. For now you could implement it in a similar way to https://user-images.githubusercontent.com/1329837/42362645-3ba29dda-812f-11e8-9cd3-77c8b3bae66b.png

The implementation should be split into two parts: the abstract implementation that has no design, and the basic implementation that looks similar to the above but uses the osu!framework design language (no rounded corners, `FrameworkColour`, `FrameworkFont`).

andy840119:
![image](https://user-images.githubusercontent.com/9100368/71334003-49571080-2577-11ea-90e8-95d1592e42c6.png)

Should this be label or textbox?

smoogipoo:
probably a textbox

andy840119:
![image](https://user-images.githubusercontent.com/9100368/71346189-3c9be200-25a2-11ea-9600-0d26ab5a3aea.png)
Working progress.
It might take a while because too many bindables.

Flutterish:
I also made a color picker (https://github.com/Flutterish/Hitokori/releases/tag/0.18.4-2021.03.15). If you want to use something similar feel free to copy my code but I do not have the time necessary to make it framework-ready.

PR #4506, opened 2021-06-11
Add hex colour picker

bdach:
Another part of #3134. The last part will be combining the two, which is basically a glorified way of saying "two `.BindTo()` calls with extra steps".

https://user-images.githubusercontent.com/20418176/121752419-60454600-cb10-11eb-9b87-dfba104a5b07.mp4

A note on the UX of this: Signaling input errors better than just discarding the value input by the user - by flashing the text box or otherwise - would definitely make this feel better, but the issue is that the structure of `TextBox` does not support that very well. What I mean by that is that both the things that control the appearance of it, and the things that control its behaviour wrt input validity (`NotifyInputError()`, `CanAddCharacter(char)`) are driven by virtual members.

So I cannot really define just the latter without the former, because that would require either multiple inheritance (impossible in this language, at least right now), or reimplementation of either the appearance or the validation logic by consumers.

If it is deemed necessary, I'll take the detour and try to fix the state of `TextBox` to decouple inherited validation logic from inherited appearance, but it'll end up being a fair few breaking changes I imagine.

However, if it's some consolation, from my experience image editing software (photoshop, GIMP) hex colour entry controls seem to work in an extremely similar manner to this.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/BasicHexColourPicker.cs
+++ b/osu.Framework/Graphics/UserInterface/BasicHexColourPicker.cs
@@ -1,0 +1,50 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Graphics.Shapes;
+
+namespace osu.Framework.Graphics.UserInterface
+{
+    public class BasicHexColourPicker : HexColourPicker
+    {
+        public BasicHexColourPicker()
+        {
+            Background.Colour = FrameworkColour.GreenDarker;
+
+            Padding = new MarginPadding(20);
+            Spacing = 10;
+        }
+
+        protected override TextBox CreateHexCodeTextBox() => new BasicTextBox
+        {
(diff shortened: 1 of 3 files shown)


frenzibyte:
Yeah I think the current UX of the textbox is fair enough.

osu.Framework/Graphics/UserInterface/BasicHexColourPicker.cs
```diff
@@ -0,0 +1,50 @@
...
+                {
+                    RelativeSizeAxes = Axes.Both
+                };
+            }
+
+            protected override void LoadComplete()
+            {
+                base.LoadComplete();
+
+                Current.BindValueChanged(_ => updatePreview());
```
peppy, line 41: I think this needs a `true` argument here? Looks like there a potential the initial colour display could be forgotten based on operation order. I think with the way you have the decoupled bindable setup right now it's likely fine, but also no harm in making sure.

peppy (changes requested):
Looks good otherwise.

peppy (approved):

Issue #4502, opened 2021-06-10
GitHub actions CI doesn't output inspectcode errors


smoogipoo:
See for example https://github.com/ppy/osu-framework/runs/2790302770

The ReSharper job outputs an error to the console but the workflow did not fail.

bdach:
The [current action](https://github.com/glassechidna/resharper-action) used for reporting inspectcode results does not seem to exit with a non-zero exit code or in any way try to set the status of the check to indicate failure, it just parses the xml report and prints it in a particular format that is picked up by github to show the inline annotations.

There is [another action](https://github.com/muno92/resharper_inspectcode), recently created, that may be closer to what is required, but I have... limited trust towards it. It may just be better to roll our own, if we do decide to continue with actions, which I'm less and less convinced about as time goes because there are important features missing.

bdach:
Another thing that occurred to me is extending/forking nvika to provide this functionality. The project [already seems to be structured in a way amenable to it](https://github.com/laedit/vika/blob/master/src/NVika/BuildServers/AppVeyor.cs). May even be able to upstream such a solution if we're lucky.

PR #4514, opened 2021-06-15
Make Resharper inspections fail the CI job

smoogipoo:
Fixes https://github.com/ppy/osu-framework/issues/4502

For the time being, I've published my own nuget package (smoogipoo.nvika) which has a GitHub build server built in. I've PR'd the change (https://github.com/laedit/vika/pull/290) and intend to use the official package once that's merged in. And if it isn't merged, I'll adopt it into the ppy org.

It can be seen working here: https://github.com/smoogipoo/osu-framework/pull/28/checks
- CI job fails
- Annotations in the `Files changed` view.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -20,8 +20,8 @@
         "jb"
       ]
     },
-    "nvika": {
-      "version": "2.0.0",
+    "smoogipoo.nvika": {
+      "version": "1.0.0",
       "commands": [
         "nvika"
       ]
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -21,7 +21,7 @@
       ]
     },
     "smoogipoo.nvika": {
-      "version": "1.0.0",
+      "version": "1.0.1",
       "commands": [
         "nvika"
       ]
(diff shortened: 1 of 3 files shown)


smoogipoo:
I'm going to be force-pushing this one twice just for display purposes (draft PR). Ignore.

bdach:
So I was doing some recon on doing the same thing as this PR does, and I found [this community answer](https://github.community/t/maximum-number-of-annotations-that-can-be-created-using-github-actions-logging-commands/16991/2) that outlines yet another baffling actions limitation, which is why I haven't PR'd something this yet:

> Currently, GitHub Actions does not support so much annotations in a workflow run, the below is a summary of the limitation:
>
> * 10 warning annotations and 10 error annotations per step
> * 50 annotations per job (sum of annotations from all the steps)
> * 50 annotations per run (separate from the job annotations, these annotations aren’t created by users

We may get 50 instead of 10 if the nvika fork is adjusted to use API instead of those workflow commands (and it'll probably need to be in the other workflow to get tokens, same as the test report thing). But the best interim solution may be to upload the inspectcode XML report as a build artifact...

smoogipoo:
Oh god why...

bdach:
I guess if we go down the API/separate workflow run route then we can do the same thing that the test plugin does and have a full summary in the check description even if we run out of the annotation limit (see [here](https://github.com/ppy/osu-framework/runs/2829477019) for example)...

bdach:
Also going to note that github have their own strongly-typed http client for their API at https://github.com/octokit/octokit.net. May be easier to use than raw requests if you decide on revisiting that path.

smoogipoo:
I think this can be merged as is in any case - at least it will fail CI even if it doesn't give the full scope of issues. Code inspection can always be run locally via `build.sh` or `InspectCode.sh` (after https://github.com/ppy/osu-framework/pull/4517).

peppy (approved):

Issue #4511, opened 2021-06-14
Dyanmic compilation will hang forever when no `.sln` file is present


jai-x:
From this line in the `DyanmicClassCompiler`:

https://github.com/ppy/osu-framework/blob/3d076ff54df4fb4a40c87e559e319350cc1caa56/osu.Framework/Testing/DynamicClassCompiler.cs#L73

It may be desirable behaviour to log if this file is not present and indicate dynamic compilation is not available without it.

PR #4515, opened 2021-06-15
Add logging if dynamic compilation is disabled due to no solution file

jai-x:
Closes https://github.com/ppy/osu-framework/issues/4511


bdach (approved):
seems reasonable :+1:

peppy (approved):

Issue #3520, opened 2020-05-05
Controlling font for FrameworkAction dialogs


kotx:
The first font that is added is automatically used for the whole game. Some fonts do not work well with FrameworkActions (the global statistics with Ctrl+F2, etc). I'd like to be able to preserve or set a separate font for said dialogs.

smoogipoo:
Imo best is to preserve the fonts, but exposing a `ToolFont` property from `Game` may be a reasonable alternative.

kotsh:
I would keep it too, but the font I'm using is a bit too
 big so text is either truncated or wrapped. AFAIK there's no way of customizing or resizing the tool window either..?
![image](https://user-images.githubusercontent.com/33439542/81036780-37567a00-8e55-11ea-88aa-da47a5eb01c6.png)

peppy:
Exposing a ToolFont sounds like a good path forward. If anyone's looking for a low hanging fruit, this may be your chance.

harshavb:
Rather than exposing a ToolFont, it should be preferable to just make it so class ToolWindow and all implementations of ToolWindow only use a FrameworkFont, since that is already the case everywhere except for two instances in TextureVisualizer.cs. The issue actually seems to exist because FrameworkFont does not work as intended in any case since Roboto and RobotoCondensed aren't implemented by default in Game.cs. However, I cannot get FrameworkFont to work while still allowing custom fonts to be defined for all non ToolWindow text unless the custom font is added _after_ Roboto and RobotoCondensed are added, which is an issue since that would mean developers would have to add in Roboto and RobotoCondensed themselves for every project. I am unsure how to continue from here, but it would likely require some sort of reimplementation of either AddFont(), addFont(), or FontUsage().

PR #4516, opened 2021-06-15
Use `Roboto` font as framework default and set explicitly in system interfaces

jai-x:
Removes the `OpenSans` font in preference of the `Roboto` font as the framework default, and adds the remaining italic variants of the `Roboto` font for completeness.

Explicitly specifies usage of `FrameworkFont` within framework provided system interfaces.

Implemented in this PR:
* Log Overlay
* Draw Visualiser
* Texture Visualiser
* Default Tooltip text
* Performance Overlay

Already implemented:
* Global Statistics Display

Closes https://github.com/ppy/osu-framework/issues/4524
Implements part of https://github.com/ppy/osu-framework/issues/3520 (Fixes the system interfaces to not use the first font added by the framework consumer but does not expose a configurable `ToolFont`)


frenzibyte:
Has it actually been decided whether this should constantly use `FrameworkFont` or allow the `Game` to define the font for such components? kinda unsure about the former since osu!lazer was unintentionally using the torus font for the tools and it looked quite cooler that way since it's matching the UI.

jai-x:
I'm not sure it it's been discussed anywhere besides the linked issue.

Changes to `Tooltip` wont make a difference to the main game since it uses a derived `OsuToolTop` class, but yes it will affect DrawViz and the log overlay.

The  Visual Test UI, DrawViz window titles, and the `Basic*` group of UI elements were already using `FrameworkFont` which is why I decided to implement it this way.

peppy:
this is the correct direction (tools do not need to "look cool")￼.

needs explanation why it only "partially" closes (and change to the OP if that's the case, as it's going to close the issue as it stands).

jai-x:
I added "partly closes" as this solves the issue of the default font being used for system interfaces, but does not implement the preferred solution of exposing a `ToolFont` to make the system interface configurable as mentioned in the original issue.

peppy:
I'm okay with getting this in as-is and adding the customisable tool font if there's actually a demand for it.

smoogipoo:
I'm not sure I agree with elements like the tooltip container and BasicTabControl using `FrameworkFont`, but I looked around at other components and they seem to do the same thing.

peppy:
I think the reasoning there is that we are using those components in the overlays themselves.

Have you seen any other projects relying on the `Basic` components where this would actually be important? I think most people who are making something that is intended to look good would already be using their own implementation.

Although I guess you could argue that the `Basic` components aren't `sealed` and therefore some may be extending them (i'd really hope they were `sealed`..).

peppy:
@smoogipoo can you confirm you're okay with this initial direction? leaving the customisation of the framework font (or reconsidering `Basic` elements) until someone reports having an issue?

smoogipoo:
Yup.

peppy:
On a quick check, this doesn't actually change anything when used with osu!. @jai-x how are you testing these changes, and have you cross-tested with another framework based application that is specifying custom fonts?

jai-x:
I tested these changes while working on my own project that added the `PetMe64` font by default and seeing which elements were using the default font while running the test scenes.

After double checking, it seems this PR only seems to make a difference when running under visual tests and not when running the desktop project, which is weird since the OpenSans font is loaded when running the desktop project according to the logs.

I've made a minimal reproduction project here which is just the sample game template with the same font I was using added to the `BaseGame` class: https://github.com/jai-x/font-test

| x | Framework `2021.601.0` | Framework `this PR` |
|---|---|---|
| Using `dotnet run -p FontTest.Game.Tests` | font_test_test_scene_framework | font_test_test_scene_pr |
| Using `dotnet run -p FontTest.Desktop` | font_test_desktop_framework | font_test_desktop_pr |

Additionally, I seemed to have missed out in fixing the FPS counter.

smoogipoo:
Are you gonna grab the fix for FPS counter in this PR? I think it's fine to go as is and leave fixing that up to a separate PR.

jai-x:
Yes, I will fix the FPS counter too. I was unsure if #4524 is a blocker, but I guess not.

peppy:
It's definitely a blocker - need to understand why that is the case before going ahead with this (ie. whether there is a reason, like we are not loading the font intentionally).

jai-x:
As suggested in the discord, I've moved the loading of the `Roboto` font into the main `Game` class to replace `OpenSans` as the default font and to be available for use as `FrameworkFont`.

I've tested this locally using lazer `master` and my own project and it seems to work fine.

By not loading `OpenSans` by default, I think this is technically a breaking change for any framework consumers that use `new FontUsage("OpenSans") directly, but it think it should just gracefully fallback to using `Roboto` as it's the new default.

osu.Framework/Game.cs
```diff
@@ -157,10 +157,13 @@ private void load(FrameworkConfigManager config)
...
-            addFont(localFonts, Resources, @"Fonts/OpenSans/OpenSans-BoldItalic");
+            // Roboto (FrameworkFont.Regular)
+            addFont(localFonts, Resources, @"Fonts/Roboto/Roboto-Regular");
+            addFont(localFonts, Resources, @"Fonts/Roboto/Roboto-Bold");
+
+            // RobotoCondensed (FrameworkFont.Condensed)
+            addFont(localFonts, Resources, @"Fonts/RobotoCondensed/RobotoCondensed-Regular");
+            addFont(localFonts, Resources, @"Fonts/RobotoCondensed/RobotoCondensed-Bold");
 
             addFont(Fonts, Resources, @"Fonts/FontAwesome5/FontAwesome-Solid");
```
jai-x, line 168: Should `FontAwsome` also be added as a local font, given it is used on parts of the framework itself?

bdach: Probably best to check whether fontawesome will continue working in consumer projects (mainly the game) after doing so without any other consumer-side changes. And also a touch out of scope here I think.

peppy:
Seems to work / look good. @smoogipoo can you confirm you're okay with removing `OpenSans` and completely replacing with `Roboto`? I also plan to remove the resources in this PR if so.

smoogipoo:
Don't we currently use OpenSans for some monospace stuff?

peppy:
Probably. But (a) it's not a monospaced font and (b) roboto is as close to monospaced as OpenSans is, so it should gracefully replace it.

smoogipoo:
Oh sorry, not monospace. We use it for bold/italic variants. I dunno, because without it testing MarkdownContainer visually is not possible.

peppy:
roboto still has bold but not italic. not sure how important that is, but we can probably add the italic version if that's seen as a blocker?

smoogipoo:
I'd see that as a blocker yeah. Other than that I'm fine with removing OpenSans.

gagahpangeran:
The markdown test scene is using Open Sans because currently it's a complete set font for now (regular, regular italic, bold, bold italic).

https://github.com/ppy/osu-framework/blob/7e069f7f6f9f72f95c4865f95b83876decbe99ff/osu.Framework.Tests/Visual/UserInterface/TestSceneMarkdownContainer.cs#L314

If we want to remove OpenSans, I think we should add italic font of Roboto to test italic sprite text.

peppy:
@jai-x do you have a pipeline setup to create the italic resources for roboto? if so please add this in this PR (and remove the `OpenSans` resources completely).

osu.Framework/Game.cs
jai-x: You're right, this does seem out of scope.

jai-x:
As requested, I've added the Italic and BoldItalic variants of the Roboto font and removed OpenSans, so I think this PR is ready for review.

peppy:
`osu.Framework/Resources/Fonts/Roboto/Roboto-BoldItalic_1.png` this file is unfortunate. but i guess it's only going to be loaded into memory should there be a usage of bold+italic+framework font, so let's just go with it for now.

peppy (approved):

gagahpangeran:
@peppy maybe add this to breaking changes because open sans font is already removed.

peppy:
Done

PR #4520, opened 2021-06-16
Lock before unbinding subtree in DisposeChildAsync

smoogipoo:
This may be a simple and safe solution to the `UnbindAllBindablesSubTree()` issue that's been causing test failures.

I've been able to run both osu! + osu!framework tests without failures, and have done brief testing in-game with scenarios like entering player and exiting immediately.

### DIFF
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -821,7 +821,9 @@ namespace osu.Framework.Graphics.Containers
         /// <param name="drawable">The child to dispose.</param>
         internal void DisposeChildAsync(Drawable drawable)
         {
-            drawable.UnbindAllBindablesSubTree();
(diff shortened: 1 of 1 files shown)


smoogipoo:
Actually this is one level too high up (should be locked on the drawable's LoadLock), will need to retest.

smoogipoo:
Tested, still seems to work.

peppy:
Seems pretty no brainer if it works, but do we just ignore all the test failures? Are they unrelated?

smoogipoo:
They definitely have to be related, somehow

bdach:
I recall trying to fix this once - in a similar manner - and hitting deadlocks related to screens, which definitely seems to be what's happened on CI there. Which is a bummer since screens are exactly the context in which the bindable unbind stuff causes failures...

smoogipoo:
Gonna close for the time being. Needs more thought.

PR #4521, opened 2021-06-16
Add combined colour picker

### LINKED ISSUE #3134: Implement color picker

bdach:
* Closes #3134.
* Supersedes / closes #3138.

The full thing, finally.

https://user-images.githubusercontent.com/20418176/122276354-fd272b00-cee4-11eb-8303-0842efd75c1b.mp4

I initially said that this final step was going to be inconsequential. That was what I thought too, until yesterday, when during preparing to PR this I arrived at the _very_ late conclusion that the HSV part was firing too many value change callbacks and therefore made the hex part flicker way too much than acceptable. Therefore, this pull also adds a `TransferValueOnCommit` property (analogous to `Slider`) that resolves that, and test coverage of that.

I can split that on request (first two commits are that), but it's not *that* huge, so I figured I'd just leave it in one piece to cut the noise. It makes *some* sense being in here, because there was no real reason for it to be a thing until now.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/HSVColourPicker.cs
+++ b/osu.Framework/Graphics/UserInterface/HSVColourPicker.cs
@@ -21,13 +21,6 @@ namespace osu.Framework.Graphics.UserInterface
         }
 
         /// <summary>
-        /// When set, colour changes based on user input are only transferred to any bound <see cref="Current"/> on commit.
-        /// This is useful if the UI interaction could be adversely affected by the value changing rapidly, such as in the case of
-        /// the <see cref="SaturationValueSelector"/>'s or <see cref="HueSelector"/>'s interactive elements being moved by the user.
-        /// </summary>
-        public bool TransferValueOnCommit { get; set; }
-
-        /// <summary>
         /// The background of the control.
         /// </summary>
         protected Box Background { get; }
@@ -36,8 +29,6 @@ namespace osu.Framework.Graphics.UserInterface
         /// Contains the elements of the colour picker.
         /// </summary>
         protected FillFlowContainer Content { get; }
-
-        private readonly Bindable<Colour4> currentInstantaneous = new Bindable<Colour4>();
 
         private readonly SaturationValueSelector saturationValueSelector;
         private readonly HueSelector hueSelector;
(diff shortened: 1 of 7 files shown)


peppy:
A couple of visual issues:

- The preview really needs to update in real-time (potentially also the hex display?). Right now it feels weird having it only update on release. I haven't looked into the code side of things yet, so I'm not sure if this is a simple change or not.

- I'm sure this is probably an issue at the textbox end, but there's currently a weird one frame flash and two frame delay until the text is updated, making using this control feel laggy:

20210617 142244 (dotnet)

Note that this is after removing the fade in at

https://github.com/ppy/osu-framework/blob/68479228f1a07d8efe9b60fe498f93d97ae0e242/osu.Framework/Graphics/UserInterface/BasicTextBox.cs#L113-L117

which also needs to go (in the case of replacing a full string in a textbox it should be instant).

osu.Framework/Graphics/UserInterface/HSVColourPicker.cs
```diff
@@ -20,6 +20,13 @@ public Bindable<Colour4> Current
             set => current.Current = value;
         }
 
+        /// <summary>
+        /// When set, colour changes based on user input are only transferred to any bound <see cref="Current"/> on commit.
+        /// This is useful if the UI interaction could be adversely affected by the value changing rapidly, such as in the case of
+        /// the <see cref="SaturationValueSelector"/>'s or <see cref="HueSelector"/>'s interactive elements being moved by the user.
+        /// </summary>
+        public bool TransferValueOnCommit { get; set; }
```
peppy, line 28: In line with my feedback, maybe this needs to be moved to the `ColourPicker` control itself, rather than here (so the two sub-components can transfer values in real-time instead).

peppy (changes requested):
Seems fine apart from the visual issues mentioned. The first of which should be fixed in this PR (second is probably a second `TextBox` related change).

bdach:
> The preview really needs to update in real-time (potentially also the hex display?)

I had things working this way before. As stated above, the reason why I changed is that the text changes would start affecting framerate negatively. Even with the text box transform gone the FPS drop may remain, but I haven't checked.

As for the suggestion to only update the hex preview in real time, as written the components are architected completely wrong for that. The hex picker would have to expose two bindables (instant and committed) and presumably also become non-public due to that.

peppy:
One thing which may help is debouncing the UI updates, as if you don't do this you will be performing many updates per update frame (due to the input thread running at 1000hz). May be enough to fix the issue you mention, and potentially even better after fixes are applied to textbox.

bdach:
Does this behaviour look closer to what you would expect?

https://user-images.githubusercontent.com/20418176/122440913-50f54b00-cf9d-11eb-8f2c-a89da0f6854b.mp4

The above is after reverting the `TransferValueOnCommit` changes, adding some `Scheduler.AddOnce()` calls, locally finishing the textbox transforms and locally fixing the one-frame delay via the method described in https://github.com/ppy/osu-framework/pull/4522#issuecomment-863396432.

If it is desired for the temporary hex value to also transfer to the HSV part then that's trickier to achieve. I tried once and the code got quite confusing wrt binding directions quite quickly.

peppy:
I think this is fine for now, yep. Entering the hex would optimally also update immediately, but it's going to be one of the rarer cases and definitely shouldn't be blocking this initial implementation from being merged.

bdach:
Have pushed the changes that seemed relevant to this pull. The text box related ones (the fading animation and the 1-frame delay) seem better suited to #4522 or an alternative thereof.

peppy (approved):

Issue #3919, opened 2020-10-07
TextBox crashes if Current is disabled after a Text set


bdach:
Found while looking at ppy/osu#10412.

Given the following test case (place in `TestSceneTextBox`):

```csharp
[Test]
public void TestTextAndCurrent()
{
    InsertableTextBox textBox = null;

    AddStep("add textbox", () =>
    {
        textBoxes.Add(textBox = new InsertableTextBox
        {
            Size = new Vector2(200, 40),
            Text = "hello"
        });
    });

    AddStep("set text and disable current", () =>
    {
        textBox.Text = "goodbye";
        textBox.Current.Disabled = true;
    });
}
```

the result is a crash to desktop with the following stack trace:

```
Unhandled exception. System.InvalidOperationException: Can not set value to "" as bindable is disabled.
   at osu.Framework.Bindables.Bindable`1.set_Value(T value) in /home/dachb/Dokumenty/opensource/osu-framework/osu.Framework/Bindables/Bindable.cs:line 95
   at osu.Framework.Graphics.UserInterface.TextBox.updateCursorAndLayout() in /home/dachb/Dokumenty/opensource/osu-framework/osu.Framework/Graphics/UserInterface/TextBox.cs:line 333
   at osu.Framework.Graphics.UserInterface.TextBox.UpdateAfterChildren() in /home/dachb/Dokumenty/opensource/osu-framework/osu.Framework/Graphics/UserInterface/TextBox.cs:line 353
   at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree() in /home/dachb/Dokumenty/opensource/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 875
```

It's not a blocker for the issue, since there was another problem game-side, but this really ought to get fixed, as there's still a possibility for the game to hit this and I can't get to it today, so opening for tracking.

Yet another reason to dive into `TextBox`, I guess?

peppy:
Had an attempt at fixing this but it's not gonna be a simple one. Stopping the crash is easy enough (add a `Current.Disabled` check at one of two valid points) but it will leave the textbox's current text in a non-synced state.

I think the valid path forward is reassessing why we have this schedule logic in the first place, and doing a rewrite/refactor of the class to fix it up. Probably not a low-effort fix, so if we decide the stopping the crash short-term is for the best I can commit the fix, but it will still fail the final assert:

[...]

PR #4522, opened 2021-06-17
Fix single frame delay and fade in when replacing full text of a textbox

peppy:
- Closes https://github.com/ppy/osu/issues/13919

Just PRing as a proof of concept fix for the issues pointed out in https://github.com/ppy/osu-framework/pull/4521#issuecomment-862936220. I'm not sure if there are any issues with removing the schedule call here, but from a logical perspective it does make ense (the scheduling logic was there mostly to handle the keyboard input correctly, I believe).

Should not be merged without due consideration (which I haven't given it yet). Will revisit at a later date, if no one beats me.

Before:

https://user-images.githubusercontent.com/191335/122338144-41b0d600-cf7a-11eb-92ae-2dc862f6fd9e.mp4

After:

https://user-images.githubusercontent.com/191335/122338063-22b24400-cf7a-11eb-87ba-a079323ea233.mp4


bdach:
FWIW I have attempted to touch these schedules before to fix #3919, as that problem is also related to them, and came to the conclusion that that text update scheduler stuff seems to be untouchable as removing any of it starts breaking things.

peppy:
I think in this case the tests may be incorrect, but will need further investigation.

bdach:
So I checked this branch out just to have a quick look at the test failures and it seems that this change somehow breaks layouting and the glyphs of the initial text are really tiny...

2021-06-17-164312_1280x707_scrot

bdach:
An alternative way of fixing that I found after some poking at this would be to permit the following logic which runs in `UpdateAfterChildren()`:

https://github.com/ppy/osu-framework/blob/0b4a9c6834041119e25cab1d00e79acfdfceb980/osu.Framework/Graphics/UserInterface/TextBox.cs#L351-L356

to also run in `Update()`. The usage of `UpdateAfterChildren()` seems to be cause of the 1-frame gap, as a `Text` set from a parent can run before the textbox's `Update()`, but I'm not yet sure why nothing is shown in the box in that case. May be a smidge expensive of a change, too...

peppy:
> Edit: Never mind. Doing that breaks caret. Just letting the `textUpdateScheduler` run seems fine, though:

You mean running a second time in update before the after children call? Seems a bit suspicious, no? Given that the `Text_Set` flow is already invalidating the `Cached` which is responsible for making it run at that point.

Maybe I'm missing something..

bdach:
Yeah I dunno. Just spent about 45 minutes trying to figure out which piece of the schedule is causing the 1-frame hide and came out with nothing concrete. I also am not sure why the text sprites go away even though nothing is invalidated or even mutated

peppy:
Probably best to close this and just wait for someone to find the time to do a full refactor of the component then?

bdach:
I dunno, adding the `.FinishTransforms()` fix makes the one-frame issue more apparent to the point where I'm not even sure it's an improvement over the current state of things...

Probably best to close then. I may look at it later today again after I do the more pressing things. I don't really believe that textbox refactor is ever coming though, the class _is_ pretty bad to read but touching anything in it seems to break things, so it may just be that it needs to be written this way...

peppy:
At very least the class need documentation as to why it is done in the current way, as it's not apparent. I believe it may have something to do with the consumption of text from the SDL/other source, but it was written so long ago that's no longer clear.

bdach:
I've stared at this branch some more and I think I found a way to fix the layout issue. The cause of the layouting failure was that `updateCursorAndLayout` was reading `CalculatedTextSize` which is not valid to read before load complete. This is a diff that fixes:

```diff
diff --git a/osu.Framework/Graphics/UserInterface/TextBox.cs b/osu.Framework/Graphics/UserInterface/TextBox.cs
index 22e0d16498..00eca1ce6a 100644
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -147,6 +147,12 @@ private void load(GameHost host)
             }
         }
 
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+            setText(Text);
+        }
+
         public virtual bool OnPressed(PlatformAction action)
         {
             int? amount = null;
@@ -640,21 +646,31 @@ public virtual string Text
                     Placeholder.Hide();
 
                 if (!IsLoaded)
+                {
                     Current.Value = text = value;
+                    return;
+                }
+
+                setText(value);
+            }
+        }
 
-                int startBefore = selectionStart;
-                selectionStart = selectionEnd = 0;
+        private void setText(string value)
+        {
+            textUpdateScheduler.CancelDelayedTasks();
 
-                TextFlow?.Clear();
-                text = string.Empty;
+            int startBefore = selectionStart;
+            selectionStart = selectionEnd = 0;
 
-                // insert string and fast forward any transforms (generally when replacing the full content of a textbox we don't want any kind of fade etc.).
-                insertString(value, d => d.FinishTransforms());
+            TextFlow?.Clear();
+            text = string.Empty;
 
-                selectionStart = Math.Clamp(startBefore, 0, text.Length);
+            // insert string and fast forward any transforms (generally when replacing the full content of a textbox we don't want any kind of fade etc.).
+            insertString(value, d => d.FinishTransforms());
 
-                cursorAndLayout.Invalidate();
-            }
+            selectionStart = Math.Clamp(startBefore, 0, text.Length);
+
+            cursorAndLayout.Invalidate();
         }
 
         public string SelectedText => selectionLength > 0 ? Text.Substring(selectionLeft, selectionLength) : string.Empty;
```

The above also contains a change that calls `textUpdateScheduler.CancelDelayedTasks();` because none of those are valid any longer when a `Text` set comes in, as I previously once mentioned in discord.

Unsure whether you find this type of change acceptable, though. If that's too much I'll try exploring the other direction (calling the text scheduler in `Update()`) further.

bdach:
Have pushed the aforementioned patch, as well as a removal of the  `textUpdateScheduler` thing. Rationale for that can be found in commit message of 49790e3.

There is one issue that arises in specific scenarios when `Current.Value` change callbacks are installed (the caret can get displaced) but I have changes on top of this that fix that issue, as well as another perennial `TextBox` pain point (`Current` and `Text` desynchronisation).

bdach (approved):

bdach:
Actually now that I think about it, it's probably wise to run this past game-side tests first to ensure nothing was quietly depending on `Text` sets being marshaled to update thread via the scheduler... Dismissing approve pending that verification.

bdach:
Have ran this past game-side tests and got no failures back, so at least in theory looks good in that respect.

smoogipoo (approved):

PR #4527, opened 2021-06-19
Fix draw thread not releasing GL context when toggling threading mode

bdach:
Closes https://github.com/ppy/osu/issues/13580.

Accidentally regressed in #4505. `DrawThread` was silently relying on its `Cleanup()` method being called regardless of whether `Thread` was null or not, which also released the GL context via the override.

This cannot be tested on CI as it requires a game host with a GL context, but I have attempted to take several steps to make sure this does not happen again, and if it does, is caught much faster:

* Split the GL context release to a new override and documented it properly.
* Documented `Thread` so that readers recall what it is better and what `null` means.
* Added error checking to the SDL implementation of `MakeCurrent()`.
* Added a visual-only test scene that toggles the execution mode twice (if you run it at cf3bc2b, it'll hard-crash to desktop but immediately indicate that the error is related to not being able to acquire the GL context).


osu.Framework/Platform/SDL2/SDL2GraphicsBackend.cs
```diff
@@ -30,7 +30,12 @@ protected override IntPtr CreateContext()
             return context;
         }
 
-        protected override void MakeCurrent(IntPtr context) => SDL.SDL_GL_MakeCurrent(sdlWindowHandle, context);
+        protected override void MakeCurrent(IntPtr context)
+        {
+            int result = SDL.SDL_GL_MakeCurrent(sdlWindowHandle, context);
+            if (result < 0)
+                throw new InvalidOperationException($"Failed to acquire GL context ({SDL.SDL_GetError()})");
```
bdach, line 37: See [documentation](https://wiki.libsdl.org/SDL_GL_MakeCurrent) for the return value description.

osu.Framework/Threading/GameThread.cs
```diff
@@ -228,6 +235,8 @@ private void updateCulture()
 
         public void Pause()
         {
+            OnPause();
```
bdach, line 238: Unsure if this should be under the lock or not. For the only current usage it's fine to leave outside as the draw thread will re-acquire it again via the next `MakeCurrent()` call anyway.

osu.Framework.Tests/Visual/Platform/TestSceneExecutionModes.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using NUnit.Framework;
+using osu.Framework.Allocation;
+using osu.Framework.Bindables;
+using osu.Framework.Configuration;
+using osu.Framework.Platform;
+
+namespace osu.Framework.Tests.Visual.Platform
+{
+    [Ignore("This test does not cover correct GL context acquire/release when ran headless.")]
```
peppy:
```suggestion
    [Ignore("This test does not cover correct GL context acquire/release when run headless.")]
```

peppy (approved):

osu.Framework/Threading/GameThread.cs
peppy: I think it's fine where it is until it is used in a case where we need to reconsider this.

smoogipoo: Shouldn't code be running on the target thread here?

bdach: That's a good point, I got confused myself in the thread object vs OS thread distinction there... Weird that it works, though, I checked.

I'll fix that up. Probably need to call twice, once here in the non-null case and once in `runWork`.

peppy (approved):

peppy:
@smoogipoo give this one a check when you can; would like to get it in an imminent framework release.

smoogipoo (approved):
Seems fine

PR #4528, opened 2021-06-20
Add popover component & container

bdach:
This is another piece on the path to a colour picker on the game-side song setup screen. To quote [Apple UI guidelines](https://developer.apple.com/design/human-interface-guidelines/ios/views/popovers/):

> A popover is a transient view that appears above other content onscreen when you tap a control or in an area. Typically, a popover includes an arrow pointing to the location from which it emerged.

I won't bother to type more words and leave this video demo instead (all included in visual tests):

https://user-images.githubusercontent.com/20418176/122685059-8900c600-d209-11eb-8797-da36dcbe89f5.mp4

I'm PRing this in one big chunk, because splitting seemed a stretch. Half of the diff is basically tests which are a little finicky to set up to include enough configurations to be good enough visual coverage. Technically the first commit _can_ be split off if you think that is best; let me know.

Skepticism is definitely warranted on this one, I'm half-expecting this to go through a few rounds of review. I was half-following `TooltipContainer` and `ContextMenuContainer`, as this type of control is somewhere halfway between the two, but I might have transgressed somewhere along the way.


peppy:
Probably not of huge concern, but the layout will flip on a centre aligned popover when the horizontal window size changes:


https://user-images.githubusercontent.com/191335/124075618-194bd000-da80-11eb-9f55-73b4dce6ac19.mp4

bdach:
That's pretty ugly... I may throw some degree of tolerance there to bias the layout algorithm towards keeping the previous orientation for cases like this. Will see what good that will do.

osu.Framework/Graphics/Cursor/PopoverContainer.cs
```diff
@@ -0,0 +1,183 @@
...
+
+namespace osu.Framework.Graphics.Cursor
+{
+    public class PopoverContainer : CursorEffectContainer<PopoverContainer, IHasPopover>
+    {
+        private readonly Container content;
+        private readonly Container<Popover> popoverContainer;
+
+        private IHasPopover? target;
+        private Popover? currentPopover => popoverContainer.LastOrDefault();
```
peppy, line 26: This feels like a hidden detail which makes readability bad. I'd rather this was a field and explicitly set/updated.

osu.Framework/Graphics/Cursor/PopoverContainer.cs
```diff
@@ -0,0 +1,183 @@
...
+                return;
+
+            currentPopover?.Hide();
+
+            var newPopover = target.GetPopover();
+            if (newPopover == null)
+                return;
+
+            popoverContainer.Add(newPopover);
+            Debug.Assert(currentPopover != null);
```
peppy, line 71: Touching on the same issue with `currentPopover` mentioned above: this reads pretty bad. For  moment I had no idea how `currentPopover` was getting updated and how any of this code would work.

Is there a reason for not setting it explicitly?

osu.Framework/Graphics/Cursor/PopoverContainer.cs
```diff
@@ -0,0 +1,183 @@
...
+            currentPopover.Show();
+            currentPopover.State.BindValueChanged(_ => cleanUpPopover(currentPopover));
+        }
+
+        private void cleanUpPopover(Popover popover)
+        {
+            if (popover.State.Value == Visibility.Hidden)
+                popover.Expire();
```
peppy, line 79: Wondering if this logic would fit better inside `Popover` itself..

peppy (changes requested):

osu.Framework/Graphics/Cursor/PopoverContainer.cs
bdach: I had the current popover as a field initially, but changed to this unorthodox way after hitting a bug wherein I was holding onto references of expired drawables. Will revisit and see if I can make the obvious variant work.

osu.Framework/Graphics/Cursor/PopoverContainer.cs
bdach: In two minds about this one. The original intent of keeping the expire in the container and not the component itself was that the drawable is sort of independent, and the container is responsible for managing popover lifetimes.

Like, say, if you wanted to use a static popover somewhere, it may be a bit awkward if the user expects it to be reusable, but it just expires itself on close instead.

peppy: Yeah, that's definitely one perspective. Better hope that the implementation of `Popover` always runs all its animations on itself though, and in its `StateChanged` event before this expire call (else they are going to get lost).

bdach:
I've pushed in changes that aim to fix the weirdness with managing popover instances in the containers, as well as the jittery behaviour.

I've left the expiry logic as-is for now as it was convenient for doing clean-up of the container state itself - let me know if you still think that popovers should expire themselves on hide.

osu.Framework.Tests/Visual/UserInterface/TestScenePopoverContainer.cs
```diff
@@ -0,0 +1,217 @@
...
+                        Anchor popoverAnchor = 0;
+                        popoverAnchor |= (Anchor)((int)Anchor.x0 << i);
+                        popoverAnchor |= (Anchor)((int)Anchor.y0 << j);
+
+                        cells[j, i].Child = new ButtonWithPopover
+                        {
+                            Width = 200,
+                            Height = 30,
+                            Text = $"open {popoverAnchor}",
+                            Action = () => { },
```
smoogipoo, line 201: I'm fairly certain this component won't work on non-buttons. See this remark on `IsHovered` (which is what `FindTargets()` uses):
https://github.com/ppy/osu-framework/blob/5c95405bac5eafcb54f477cf2ce18f81921ac74f/osu.Framework/Graphics/Drawable.cs#L2507-L2508

Is this intended to be used in non-button scenarios? Perhaps this should be added to the positional input interfaces:
https://github.com/ppy/osu-framework/blob/5c95405bac5eafcb54f477cf2ce18f81921ac74f/osu.Framework/Graphics/Drawable.cs#L2408-L2413

smoogipoo: As a side note: we should probably remove those manual interface listings and replace with `IHandles(Non)PositionalInput` or something like that. It's not really extensible as it is.

bdach: Good point. I was not aware of these interface specs, and in my wip integration branch I was overriding `HandlePositionalInput`. Will add to the positional interfaces.

peppy:
On a conceptual level, is there any specific reason for structuring this the way it is (with the `PopoverContainer`)? ie. couldn't this whole thing exist as just the popover drawable? I guess the idea is to limit the total number of concurrent popovers to one?

With the point @smoogipoo touched on: it feels *really* weird that a popover can exist on an element which is non-interactable in the first place. Feels back to front.

Like that `positional_input_interfaces` thing, I can understand for tooltips but I'm not so sure for context menus and popovers..

bdach:
> I guess the idea is to limit the total number of concurrent popovers to one?

Right. I'm no UX designer but from what I gather the idea is that it's a semi-modal component and therefore should hold mental focus of the user. Having many of these flying around isn't all that great.

> it feels really weird that a popover can exist on an element which is non-interactable in the first place

I thought of it in such a way that declaring a popover is declaring interactability. But that may just be me, I dunno.

peppy:
More valid reasoning for the container seems to be that it's required to figure out the layout / bounds for where the popover can be displayed. I think it's fine for now, but still not 100% on either of those last two being in the interactive interfaces list.

peppy:
I've fixed a case where dragging from inside a popover to outside causing it to be incorrectly hidden. Looks good otherwise.

peppy (approved):

PR #4529, opened 2021-06-20
Apply few changes to the `ShaderManager` structure

frenzibyte:
Few simple changes to `ShaderManager` to become similar to other resource stores. Required for supporting https://github.com/ppy/osu/issues/13586.

- Make `LoadRaw` public and virtual, for the ability of overriding it and making a `FallbackShaderManager` osu!-side in `DrawableRulesetDependencies`.
- Add disposal logic that disposes of the underlying store, matches others like `TextureStore` which inherits `ResourceStore` that disposes of underlying stores.
- Make `ShaderManager` accept null underlying stores and make it an optional parameter, similar to `TextureStore` as well.

Also reorganizes the methods and xmldocs `Load` methods.

As a side note, I was also thinking about making `ShaderManager` inherit from `ResourceStore<byte[]>` and remove `LoadRaw` in favour of `ResourceStore.Get` instead (or even make it inherit `ResourceStore<ShaderPart>` and make a `ShaderLoaderStore`), but I realized that's changing too much than what's needed currently, so I left that part as-is for now. Can look into doing that in a separate PR if it sounds good, but it's out-of-scope here.

### DIFF
--- a/osu.Framework/Graphics/Shaders/ShaderManager.cs
+++ b/osu.Framework/Graphics/Shaders/ShaderManager.cs
@@ -48,7 +48,7 @@ namespace osu.Framework.Graphics.Shaders
             return name + ending;
         }
 
-        internal byte[] LoadRaw(string name) => store.Get(name);
+        public virtual byte[] LoadRaw(string name) => store.Get(name);
 
         private ShaderPart createShaderPart(string name, ShaderType type, bool bypassCache = false)
         {
--- a/osu.Framework/Graphics/Shaders/ShaderManager.cs
+++ b/osu.Framework/Graphics/Shaders/ShaderManager.cs
@@ -29,14 +29,6 @@ namespace osu.Framework.Graphics.Shaders
         }
 
         /// <summary>
-        /// Constructs a new <see cref="ShaderManager"/> with no backing store.
-        /// </summary>
-        protected ShaderManager()
-        {
-            store = new ResourceStore<byte[]>();
-        }
-
-        /// <summary>
         /// Retrieves raw shader data from the store.
         /// Use <see cref="Load"/> to retrieve a usable <see cref="IShader"/> instead.
         /// </summary>
--- a/osu.Framework/Graphics/Shaders/ShaderManager.cs
+++ b/osu.Framework/Graphics/Shaders/ShaderManager.cs
@@ -25,32 +25,34 @@ namespace osu.Framework.Graphics.Shaders
             this.store = store;
         }
 
-        private string getFileEnding(ShaderType type)
-        {
-            switch (type)
-            {
-                case ShaderType.FragmentShader:
-                    return @".fs";
-
-                case ShaderType.VertexShader:
-                    return @".vs";
-            }
-
-            return string.Empty;
-        }
-
-        private string ensureValidName(string name, ShaderType type)
-        {
-            string ending = getFileEnding(type);
-            if (!name.StartsWith(shader_prefix, StringComparison.Ordinal))
-                name = shader_prefix + name;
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Shaders/ShaderManager.cs
```diff
@@ -16,9 +17,10 @@ public class ShaderManager : IDisposable
         private readonly ConcurrentDictionary<string, ShaderPart> partCache = new ConcurrentDictionary<string, ShaderPart>();
         private readonly ConcurrentDictionary<(string, string), Shader> shaderCache = new ConcurrentDictionary<(string, string), Shader>();
 
-        private readonly ResourceStore<byte[]> store;
+        [CanBeNull]
+        private readonly IResourceStore<byte[]> store;
 
-        public ShaderManager(ResourceStore<byte[]> store)
+        public ShaderManager(IResourceStore<byte[]> store = null)
```
peppy, line 23: What's the use case where this is null?

frenzibyte: `FallbackShaderManager` implementation osu!-side, it doesn't require an underlying source, since it overrides the `LoadRaw` method to go through both the primary `ShaderManager` and the fallback `ShaderManager`.

I've made it optional as that's what `TextureStore` does as well, but I could remove it if it's deemed wrong. (and the same should likely apply to `TextureStore`)

peppy: So to answer my question: it is never used?

frenzibyte: Yes, along with the reasoning why.

peppy: Let's make it non-nullable. Also might as well decorate the class with `#nullable enable` while you're there.

frenzibyte: Done, I have also made the no-store case possible with a `protected` constructor only, rather than one `public` constructor with optional store.

osu.Framework/Graphics/Shaders/ShaderManager.cs
```diff
@@ -17,20 +18,30 @@ public class ShaderManager : IDisposable
...
+        /// </summary>
+        public ShaderManager(IResourceStore<byte[]> store)
         {
             this.store = store;
         }
 
+        /// <summary>
+        /// Constructs a new <see cref="ShaderManager"/> with no backing store.
+        /// </summary>
+        protected ShaderManager()
```
peppy, line 34: Do you think this is necessary? I don't understand how a `ShaderManager` is of any use without a backing store. If it's not, then this should be left up to the consumer to do.

peppy (changes requested):

osu.Framework/Graphics/Shaders/ShaderManager.cs
frenzibyte: Not really, just saves up from having to do it in [`FallbackShaderManager`](https://github.com/ppy/osu/blob/c933cbe89d0aedf6ddbef2c0f320595973a1aec5/osu.Game/Rulesets/UI/DrawableRulesetDependencies.cs#L186-L204). Can remove it.

peppy: Why would it even need to be done there? Surely if there's no source you just make the manager... null? Not make an empty manager and query for nothing.

peppy (approved):

smoogipoo (approved):

PR #4531, opened 2021-06-21
Fix `LoadRaw` return value not marked as nullable

frenzibyte:
`LoadRaw` returns `null` if the requested file could not be found. Didn't notice this as `ResourceStore`s don't have `#nullable enable` yet, which would make such mistakes easier to spot.

As an aside, I've double-checked ppy/osu#13594 with framework master + this PR, to ensure nothing requires changing any further.


peppy (approved):

peppy:
Probably could have also added hinting at the source but I guess this is something.

PR #4532, opened 2021-06-22
Change markdown heading text size to use actual font size

### LINKED ISSUE #4525: MarkdownHeading shouldn't use scale for text size
https://github.com/ppy/osu-framework/blob/0b4a9c6834041119e25cab1d00e79acfdfceb980/osu.Framework/Graphics/Containers/Markdown/MarkdownHeading.cs#L39

Or else it'll overflow.

Kapture 2021-06-17 at 18 34 30

osu!-side example:
osu_2021-06-17_18-37-45

gagahpangeran:
Closes #4525

As I mentioned in https://github.com/ppy/osu-framework/issues/4525#issuecomment-864344644

Currently, `GetFontSizeByLevel` returns scale number value instead of actual font size value. Then that value is used to change heading size using scale.

This changes is make `GetFontSizeByLevel` to return actual font size value and then change the font size directly instead of using scale.

This is a bit breaking changes, so I need more advise for this PR.

### DIFF
--- a/osu.Framework/Graphics/Containers/Markdown/MarkdownHeading.cs
+++ b/osu.Framework/Graphics/Containers/Markdown/MarkdownHeading.cs
@@ -3,7 +3,7 @@
 
 using Markdig.Syntax;
 using osu.Framework.Allocation;
-using osuTK;
+using osu.Framework.Graphics.Sprites;
 
 namespace osu.Framework.Graphics.Containers.Markdown
 {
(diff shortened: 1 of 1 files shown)


peppy:
I'm not sure if we need obsolescence support for this change, but if it's determined we do, you will need to leave the old method in place and mark obsolete, then make a new method (which an arbitrarily different name) call the old one (which will return 0 in the `virtual` implementation) and apply a ratio to get the correct font size from the old scale in a non-zero case.

@smoogipoo do you think it's worth doing so for this? I'd argue the usage of it is probably pretty rare, but it has been around since 2019 so we may want to for consistency.

smoogipoo:
To my knowledge, the only other project that uses the markdown container is https://github.com/karaoke-dev/karaoke, and as far as I'm aware they don't override this method.

I think it's pretty safe to break.

peppy:
Cool, will go ahead with merging this and adding a breaking change to the listing.

peppy (approved):

peppy:
@gagahpangeran can i presume you have an osu!-side partner PR for this pending?

gagahpangeran:
Yes, already did that.

PR #4533, opened 2021-06-22
Expose an event from `GameHost` before update thread changes native threads

peppy:
PRing for comment. Intended to be used as a way for realm to clean up after itself.

Not sure if we want to instead expose somewhere in `GameThread` itself.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -94,9 +94,9 @@ namespace osu.Framework.Platform
 
         /// <summary>
         /// Called from the current update thread before it is potentially moved to a different native thread.
-        /// This could occur from a change in <see cref="ExecutionMode"/>.
+        /// This could occur from a change in <see cref="ExecutionMode"/> or end of host execution.
         /// </summary>
-        public event Action UpdateThreadChanging;
+        public event Action UpdateThreadPausing;
 
         /// <summary>
         /// Whether the on screen keyboard covers a portion of the game window when presented to the user.
@@ -567,6 +567,8 @@ namespace osu.Framework.Platform
                     Monitor = { HandleGC = true },
                 });
 
+                UpdateThread.ThreadPausing += () => UpdateThreadPausing?.Invoke();
+
                 RegisterThread(DrawThread = new DrawThread(DrawFrame, this));
 
                 Trace.Listeners.Clear();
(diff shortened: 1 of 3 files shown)


osu.Framework/Platform/GameHost.cs
```diff
@@ -802,7 +808,16 @@ protected virtual void SetupConfig(IDictionary<FrameworkSetting, object> default
             }, true);
 
             executionMode = Config.GetBindable<ExecutionMode>(FrameworkSetting.ExecutionMode);
-            executionMode.BindValueChanged(e => threadRunner.ExecutionMode = e.NewValue, true);
+            executionMode.BindValueChanged(e =>
+            {
+                if (ExecutionState >= ExecutionState.Running)
```
smoogipoo, line 813: Nit: Perhaps `> ExecutionState.Stopped` instead? Should throw if it's stopping but not yet stopped.

smoogipoo (changes requested):
Seems fine otherwise.

peppy:
I've added a bit of discussion on discord that means this function will likely have to change. Basically, it turns out that `Update` frames can run on the old update thread even after this event is called. So the API will either need to be a set of two events exposed (`UpdateThreadChanging` -> `UpdateThreadChanged`) or some other equivalent that guaranteed states...

smoogipoo:
I was originally going to ask why this wasn't done in `ThreadRunner` itself. An event that's run at the end of `ensureCorrectExecutionMode` - doesn't matter if it's run on update or whatever (have the binder decide where to schedule it).

It could be re-exposed through `GameHost` but the idea is the original event comes from `ThreadRunner`.

peppy:
Yeah, that sounds better. I'll take a look at updating to do so.

osu.Framework/Threading/GameThread.cs
```diff
@@ -259,6 +264,7 @@ public void Pause()
         /// </summary>
         protected virtual void OnPause()
         {
+            ThreadPausing?.Invoke();
```
bdach, line 267: While this works, how about making this method private non-virtual (and making the draw thread subscribe itself to the event to release the GL context)? Would remove a bit of complexity in this setup and ensure there can be no funny business wrt something overriding `OnPause()` and accidentally breaking contract of `ThreadPausing` by suppressing a base call or something.

Maybe a bit of a paranoid one but the GL context release regression made me a bit paranoid about this sort of pattern of side effects via overrides as `GameThread`s have a lot of those and it feels a bit like walking a minefield at times...

peppy: i figured it was seldom modified code so should not be an issue, but can change that i guess. self subscription is a weird pattern to me so i tend to avoid it when writing code.

bdach: Fair call. I'm not insisting on this change, it can stay as is. Hopefully it shouldn't ever matter.

smoogipoo (approved):

PR #4539, opened 2021-06-23
Add shader disposal support

### LINKED ISSUE #4534: Shaders aren't disposed
Now that we're at the point that we're allowing 3rd party shaders to pop in and out of existence (see: https://github.com/ppy/osu/pull/13594), we need to take another look into shader disposal.

Disposal was removed in https://github.com/ppy/osu-framework/pull/2580.

* `ShaderPart` has remnants of a now-unused `delete` method.
* `Shader` is lacking any cleanup whatsoever.

Of importance:
* Make sure that shaders aren't disposed by the user, since their lifetime is controlled by a `ShaderManager` at all times. This means `IDisposable` cannot be added to the `IShader` interface.
* Disposal needs to be delayed for exactly 3 frames (go through `GLWrapper.ScheduleDisposal()`).
* Make sure to check that global uniform bindings are unregistered.

frenzibyte:
Closes #4534 

Follows the existing pattern for GL disposal, if a `Shader`/`ShaderPart` gets finalised, the 3-frame delay will be applied, but if it's explicitly disposed, then there won't be any delay from the class itself, it would come from whatever's disposing of it internally (the `ShaderManager` in this case).

Double-checked with osu!lazer master and ppy/osu#13594. Looks to be working as intended. 

Also added a simple test case ensuring `Shader` references aren't leaked after disposing of `ShaderManager` (which heavily depended on doing https://github.com/ppy/osu-framework/commit/cd36ea3b8608aaa305603c0763c142a2ffd063f9 for testing without GL context, kinda ugly but needed).

### DIFF
--- a/osu.Framework/Graphics/Shaders/Shader.cs
+++ b/osu.Framework/Graphics/Shaders/Shader.cs
@@ -19,7 +19,6 @@ namespace osu.Framework.Graphics.Shaders
         private int programID = -1;
 
         internal readonly Dictionary<string, IUniform> Uniforms = new Dictionary<string, IUniform>();
-        private IUniform[] uniformsArray;
         private readonly List<ShaderPart> parts;
 
         internal Shader(string name, List<ShaderPart> parts)
@@ -34,98 +33,18 @@ namespace osu.Framework.Graphics.Shaders
         {
             parts.RemoveAll(p => p == null);
             Uniforms.Clear();
-            uniformsArray = null;
 
             if (parts.Count == 0)
                 return;
 
-            programID = GL.CreateProgram();
+            programID = CreateProgram();
 
-            foreach (ShaderPart p in parts)
-            {
-                if (!p.Compiled) p.Compile();
-                GL.AttachShader(this, p);
+            if (!CompileInternal())
+                throw new ProgramLinkingFailedException(name, GetProgramLog());
 
-                foreach (ShaderInputInfo input in p.ShaderInputs)
-                    GL.BindAttribLocation(this, input.Location, input.Name);
-            }
+            IsLoaded = true;
 
-            GL.LinkProgram(this);
-            GL.GetProgram(this, GetProgramParameterName.LinkStatus, out int linkResult);
-
-            foreach (var part in parts)
-                GL.DetachShader(this, part);
-
-            IsLoaded = linkResult == 1;
-
-            if (!IsLoaded)
-                throw new ProgramLinkingFailedException(name, GL.GetProgramInfoLog(this));
-
-            // Obtain all the shader uniforms
-            GL.GetProgram(this, GetProgramParameterName.ActiveUniforms, out int uniformCount);
-            uniformsArray = new IUniform[uniformCount];
-
-            for (int i = 0; i < uniformCount; i++)
-            {
-                GL.GetActiveUniform(this, i, 100, out _, out _, out ActiveUniformType type, out string uniformName);
-
-                IUniform createUniform<T>(string name)
-                    where T : struct, IEquatable<T>
-                {
-                    int location = GL.GetUniformLocation(this, name);
-
-                    if (GlobalPropertyManager.CheckGlobalExists(name)) return new GlobalUniform<T>(this, name, location);
-
-                    return new Uniform<T>(this, name, location);
-                }
-
-                IUniform uniform;
-
-                switch (type)
-                {
-                    case ActiveUniformType.Bool:
-                        uniform = createUniform<bool>(uniformName);
-                        break;
-
-                    case ActiveUniformType.Float:
-                        uniform = createUniform<float>(uniformName);
-                        break;
-
-                    case ActiveUniformType.Int:
-                        uniform = createUniform<int>(uniformName);
-                        break;
-
-                    case ActiveUniformType.FloatMat3:
-                        uniform = createUniform<Matrix3>(uniformName);
-                        break;
-
-                    case ActiveUniformType.FloatMat4:
-                        uniform = createUniform<Matrix4>(uniformName);
-                        break;
-
-                    case ActiveUniformType.FloatVec2:
-                        uniform = createUniform<Vector2>(uniformName);
-                        break;
-
-                    case ActiveUniformType.FloatVec3:
-                        uniform = createUniform<Vector3>(uniformName);
-                        break;
-
-                    case ActiveUniformType.FloatVec4:
-                        uniform = createUniform<Vector4>(uniformName);
-                        break;
-
-                    case ActiveUniformType.Sampler2D:
-                        uniform = createUniform<int>(uniformName);
-                        break;
-
-                    default:
-                        continue;
-                }
-
-                uniformsArray[i] = uniform;
-                Uniforms.Add(uniformName, uniformsArray[i]);
-            }
+            SetupUniforms();
 
             GlobalPropertyManager.Register(this);
         }
@@ -145,7 +64,7 @@ namespace osu.Framework.Graphics.Shaders
 
             GLWrapper.UseProgram(this);
 
-            foreach (var uniform in uniformsArray)
+            foreach (var uniform in Uniforms.Values)
                 uniform?.Update();
 
             IsBound = true;
@@ -161,8 +80,6 @@ namespace osu.Framework.Graphics.Shaders
             IsBound = false;
         }
 
-        public override string ToString() => $@"{name} Shader (Compiled: {programID != -1})";
-
         /// <summary>
         /// Returns a uniform from the shader.
         /// </summary>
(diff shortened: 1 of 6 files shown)


osu.Framework/Graphics/Shaders/Shader.cs
```diff
@@ -56,127 +116,110 @@ internal void Compile()
...
+                return new Uniform<T>(this, name, location);
+            }
         }
 
-        internal void EnsureLoaded()
+        protected virtual string GetProgramLog() => GL.GetProgramInfoLog(this);
+
+        protected virtual int CreateProgram() => GL.CreateProgram();
+
+        protected virtual void DeleteProgram(int id)
```
bdach, line 185: Not too sure about exposing these for the tests so widely. Maybe these should be `private protected`. Nothing outside the framework assembly/a subclass should be overriding these.

osu.Framework/Graphics/Shaders/Shader.cs
```diff
@@ -56,127 +116,110 @@ internal void Compile()
...
         {
-            EnsureLoaded();
+            if (!IsDisposed)
+            {
+                IsDisposed = true;
 
-            return (Uniform<T>)Uniforms[name];
+                GlobalPropertyManager.Unregister(this);
+
+                DeleteProgram(this);
```
bdach, line 218: I think these are missing a `id != 1` check, which is now a case that has to be potentially handled as this is possibly being called from the finaliser and therefore there are no more guarantees about object state anymore. Not to mention the program might not even have been compiled yet.

osu.Framework/Graphics/Shaders/Shader.cs
```diff
@@ -19,7 +19,6 @@ public class Shader : IShader
         private int programID = -1;
 
         internal readonly Dictionary<string, IUniform> Uniforms = new Dictionary<string, IUniform>();
-        private IUniform[] uniformsArray;
```
bdach, line 22: I was skeptical about this removal (clearly this had some purpose), so I checked blame and found https://github.com/ppy/osu-framework/commit/9a3f310f75cbf8d4a6a0723cd908752f0a9515b9. So it looks like this should be benchmarked to check whether using the value collection isn't causing a performance regression, if not outright reverted.

I'm not sure what this change has to do with the rest of this pull anyway.

osu.Framework/Graphics/Shaders/Shader.cs
```diff
@@ -32,15 +31,76 @@ internal Shader(string name, List<ShaderPart> parts)
...
+            SetupUniforms();
+
+            GlobalPropertyManager.Register(this);
+        }
+
+        internal void EnsureLoaded()
+        {
+            if (!IsLoaded)
+                Compile();
+        }
```
bdach, line 59: Because `Compile()` bails if the shader is disposed, `EnsureLoaded()` does not in fact ensure that the shader is loaded. I imagine this may be a problem in all call sites that expect that to be the case.

bdach (changes requested):
Not overly familiar with this area but I'll have a go...

osu.Framework/Graphics/Shaders/Shader.cs
frenzibyte: It's actually checked inside of the method but that looks weird reading it again, will change.

osu.Framework/Graphics/Shaders/Shader.cs
frenzibyte: That's the one I was looking for.

osu.Framework/Graphics/Shaders/Shader.cs
frenzibyte: Good find, thought it isn't necessary compared to just `Uniforms.Values` but it appears I'm wrong. Wondering if enumerating on `Uniforms` then `kvp.Value.Update()` would also not be good?

osu.Framework/Graphics/Shaders/Shader.cs
frenzibyte: Seems like it should throw in here as well, `EnsureLoaded` must not be called if the shader has already been disposed of.

osu.Framework/Graphics/Shaders/Shader.cs
frenzibyte:
```
|                            Method | ItemCount |       Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------------------------- |---------- |-----------:|----------:|----------:|------:|------:|------:|----------:|
|         EnumerateDictionaryValues |      1000 | 3,769.5 ns | 151.16 ns | 443.33 ns |     - |     - |     - |         - |
| EnumerateDictionaryAndAccessValue |      1000 | 4,721.2 ns | 133.93 ns | 386.43 ns |     - |     - |     - |         - |
|                    EnumerateArray |      1000 |   551.7 ns |  11.08 ns |  31.25 ns |     - |     - |     - |         - |
```

Ok that shocked me... TIL.

osu.Framework/Graphics/Shaders/Shader.cs
frenzibyte: I've refactored that area a bit as I didn't actually intend `Compile()` to silently bail when the shader is disposed of, should be in a good state now.

frenzibyte:
Blocking momentarily while I figure out what's causing `Shader` references to still remain even though the manager is disposed of, scary enough that it happens on a real `Shader`, not a `TestShader` (even though the abstraction was meant in a way that doesn't result in differentiation between both classes).

frenzibyte:
Turns out it was due to the test scene stupidly forces a silent `Shader` compile before the `GLWrapper` scheduled one, resulting in compiling twice and registering to `GlobalPropertyManager` twice.

I have solved that in multiple ways, by:
 - Guarding in `Shader.Compile` to never recompile again, 
 - Guarding in `GlobalPropertyManager` to not add shader twice (not sure if `all_shaders` should actually exist in the first place, GC purposes?), 
 - And privatizing `Shader.Compile`, to be replaced with using `EnsureShaderCompiled` instead (all internally).

osu.Framework/Graphics/Shaders/Shader.cs
```diff
@@ -4,43 +4,119 @@
...
+        private void compile()
         {
-            parts.RemoveAll(p => p == null);
-            Uniforms.Clear();
-            uniformsArray = null;
+            if (IsDisposed)
+                throw new ObjectDisposedException(ToString(), "Can not compile a disposed shader.");
 
+            if (IsLoaded)
+                throw new InvalidOperationException("Attempting to compile already compiled shader.");
```
bdach, line 47:
```suggestion
                throw new InvalidOperationException("Attempting to compile an already-compiled shader.");
```

osu.Framework/Graphics/Shaders/Shader.cs
```diff
@@ -4,43 +4,119 @@
...
+
+            parts.RemoveAll(p => p == null);
```
bdach, line 49: What is this?

osu.Framework/Graphics/Shaders/Shader.cs
```diff
@@ -4,43 +4,119 @@
...
+            GlobalPropertyManager.Register(this);
+        }
+
+        internal void EnsureShaderCompiled()
+        {
+            if (IsDisposed)
+                throw new ObjectDisposedException(ToString(), "Can not compile a disposed shader.");
+
+            if (!shaderCompileDelegate.Completed)
+                shaderCompileDelegate.RunTask();
```
bdach, line 71: This looks dodgy/racey at first glance (even if the internal logic of `ScheduledDelegate` ensures that this won't do the wrong thing on accident). Should probably use `State` of the delegate directly.

osu.Framework/Graphics/Shaders/Shader.cs
frenzibyte: Dunno, another pre-existing thing, can remove if it's deemed redundant (I believe it to be so).

bdach: I didn't notice that it was there before, fair enough. Probably not going to insist on having it removed since it's been there since the first commit of framework...

osu.Framework/Graphics/Shaders/Shader.cs
frenzibyte: `State == Waiting` you mean? otherwise, I don't follow.

bdach: Correct.

smoogipoo (approved):
I'm not a super big fan of the `private protected virtual` stuff, but let's go with this for now.

frenzibyte:
> I'm not a super big fan of the `private protected virtual` stuff, but let's go with this for now.

Yeah without doubts, I'm not really sure what's the reason behind making all of them internal, but sounds like they should be separated in some better way, something like `Texture` and `TextureGL` maybe, needs investigation at some point.

PR #4540, opened 2021-06-23
Make tooltips text localisable

Game4all:
Pushing this to make localisation of the tooltips text possible in osu! by changing the `IHasTooltip.TooltipText` property to a `LocalisableString`. 

I guess I could have kept the string version of this property around to prevent breakage for other framework users and have a defaut implementation in the interface point to the string one, but I'm not sure default interface methods are being used in framework.

## Breaking change

`IHasTooltip.Text` now takes a `LocalisableString` instead of  regular string. Users of custom tooltip containers may also have to change checks in the `SetContent()` method of their tooltip type to check if content is a `LocalisableString` instead of a regular string.

### DIFF
--- a/osu.Framework/Graphics/Cursor/TooltipContainer.cs
+++ b/osu.Framework/Graphics/Cursor/TooltipContainer.cs
@@ -9,6 +9,7 @@ using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
 using osu.Framework.Input;
+using osu.Framework.Localisation;
 using osuTK;
 using osuTK.Graphics;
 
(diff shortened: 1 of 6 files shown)


osu.Framework/Graphics/Cursor/TooltipContainer.cs
```diff
@@ -178,8 +179,8 @@ private bool hasValidTooltip(ITooltipContentProvider target)
         {
             var targetContent = getTargetContent(target);
 
-            if (targetContent is string strContent)
-                return !string.IsNullOrEmpty(strContent);
+            if (targetContent is LocalisableString localisableString)
+                return !string.IsNullOrEmpty(localisableString.Data?.ToString());
```
Game4all, line 183: This one is a bit of a shortcut to check if the localisable string data is not null and empty and relies on the `ToString()` implementation of the different underlying string types, but since this check may only have to be done framework-side, I guess it is fine to inline it instead of adding a method to `LocalisableString`.

smoogipoo: Fine as is, though I'd probably just call `localisableString.ToString()`? It's the same thing in the end afaik.

smoogipoo (approved):
This is fine to me. Will wait until the next framework release before merging.

peppy:
> This is fine to me. Will wait until the next framework release before merging.

That time is now.

Issue #4541, opened 2021-06-24
GameThread is unsafely exposed


peppy:
Documenting a few issues which we should probably eventually get around to fixing:

- A consumer can call `host.UpdateThread.Pause()`. In single threaded mode this will fire a `ThreadPaused` event from the wrong thread. Also `Pause()` can be called after a thread is stopped and it will fire this event an extra time.
- A consumer can call `host.UpdateThread.Stop()`. They probably shouldn't be able to see this method on the internally managed threads.
- The `bool` flags for pause and exit don't feel safe (hard to confirm because there are three of them intertwining the logic). I think this needs to be rewritten with a single state enum that is always locked and asserted to be in a valid preceding state.

bdach:
> In single threaded mode this will fire a `ThreadPaused` event from the wrong thread.

To clarify, you would be expecting this event to be called by the main OS thread always? Because it is true that as is, it is called on the same OS thread that is calling for pause, so it can end up being incorrect if the pause isn't requested on the main OS thread (so any of the `ThreadRunner`-managed ones, since they run on main OS thread in single-threaded). But the intention was to call it on the main OS thread, because there is no other OS `Thread` to call the event on.

peppy:
for realm, at least, it should be fired from the thread that `GameThread` is running on. in the case of multi-threading, the `UpdateThread`; in the case of single-thread `InputThread`.

PR #4542, opened 2021-06-24
Rewrite `GameThread` and expose state as `IBindable<GameThreadState>`

peppy:
Not sure where to start or end on this one. Found many issues along the way. I'm sure there's more but this is running in a stable state and seems to be safer than what we had.

Would recommend taking the diff with a grain of salt and just parsing the whole classes involved. I guarantee we will find more areas of improvement during review (although try and keep scope in check).

Fixes issues listed in https://github.com/ppy/osu-framework/issues/4541 except for exposure of methods to consumers. Not addressing that yet because it's kind of their fault if they try to mess with threads directly.

I have not yet tested this with realm requirements, so blocking until I get to doing that.

### DIFF
--- a/osu.Framework/Platform/ThreadRunner.cs
+++ b/osu.Framework/Platform/ThreadRunner.cs
@@ -57,6 +57,8 @@ namespace osu.Framework.Platform
             }
         }
 
+        private readonly object startStopLock = new object();
+
         /// <summary>
         /// Construct a new ThreadRunner instance.
         /// </summary>
@@ -124,10 +126,13 @@ namespace osu.Framework.Platform
 
         public void Suspend()
         {
-            pauseAllThreads();
+            lock (startStopLock)
+            {
+                pauseAllThreads();
 
-            // set the active execution mode back to null to set the state checking back to when it can be resumed.
-            activeExecutionMode = null;
+                // set the active execution mode back to null to set the state checking back to when it can be resumed.
+                activeExecutionMode = null;
+            }
         }
 
         public void Stop()
@@ -158,12 +163,16 @@ namespace osu.Framework.Platform
 
         private void ensureCorrectExecutionMode()
         {
-            if (ExecutionMode == activeExecutionMode)
-                return;
+            // locking is required as this method may be called from two different threads.
+            lock (startStopLock)
+            {
+                if (ExecutionMode == activeExecutionMode)
+                    return;
 
-            // if null, we have not yet got an execution mode, so set this early to allow usage in GameThread.Initialize overrides.
-            activeExecutionMode ??= ThreadSafety.ExecutionMode = ExecutionMode;
-            Logger.Log($"Execution mode changed to {activeExecutionMode}");
+                // if null, we have not yet got an execution mode, so set this early to allow usage in GameThread.Initialize overrides.
+                activeExecutionMode ??= ThreadSafety.ExecutionMode = ExecutionMode;
+                Logger.Log($"Execution mode changed to {activeExecutionMode}");
+            }
 
             pauseAllThreads();
 
--- a/osu.Framework/Platform/ThreadRunner.cs
+++ b/osu.Framework/Platform/ThreadRunner.cs
@@ -129,8 +129,6 @@ namespace osu.Framework.Platform
             lock (startStopLock)
             {
                 pauseAllThreads();
-
-                // set the active execution mode back to null to set the state checking back to when it can be resumed.
                 activeExecutionMode = null;
             }
         }
--- a/osu.Framework/Platform/ThreadRunner.cs
+++ b/osu.Framework/Platform/ThreadRunner.cs
@@ -175,7 +175,6 @@ namespace osu.Framework.Platform
                     foreach (var t in Threads)
                     {
                         t.Start();
-                        t.Clock.Throttling = true;
                     }
 
                     break;
--- a/osu.Framework/Platform/ThreadRunner.cs
+++ b/osu.Framework/Platform/ThreadRunner.cs
@@ -155,8 +155,7 @@ namespace osu.Framework.Platform
             });
 
             // as the input thread isn't actually handled by a thread, the above join does not necessarily mean it has been completed to an exiting state.
-            while (!mainThread.Exited)
-                mainThread.ProcessFrame();
+            mainThread.WaitForState(GameThreadState.Exited);
 
             ThreadSafety.ResetAllForCurrentThread();
         }
--- a/osu.Framework/Platform/ThreadRunner.cs
+++ b/osu.Framework/Platform/ThreadRunner.cs
@@ -168,8 +168,7 @@ namespace osu.Framework.Platform
                 if (ExecutionMode == activeExecutionMode)
                     return;
 
-                // if null, we have not yet got an execution mode, so set this early to allow usage in GameThread.Initialize overrides.
-                activeExecutionMode ??= ThreadSafety.ExecutionMode = ExecutionMode;
+                activeExecutionMode = ThreadSafety.ExecutionMode = ExecutionMode;
                 Logger.Log($"Execution mode changed to {activeExecutionMode}");
             }
 
@@ -201,8 +200,6 @@ namespace osu.Framework.Platform
                     break;
                 }
             }
-
-            activeExecutionMode = ThreadSafety.ExecutionMode = ExecutionMode;
 
             updateMainThreadRates();
         }
--- a/osu.Framework/Platform/ThreadRunner.cs
+++ b/osu.Framework/Platform/ThreadRunner.cs
@@ -182,9 +182,7 @@ namespace osu.Framework.Platform
                 {
                     // switch to multi-threaded
                     foreach (var t in Threads)
-                    {
                         t.Start();
-                    }
 
                     break;
                 }
--- a/osu.Framework/Platform/ThreadRunner.cs
+++ b/osu.Framework/Platform/ThreadRunner.cs
@@ -109,7 +109,7 @@ namespace osu.Framework.Platform
                     lock (threads)
                     {
                         foreach (var t in threads)
-                            t.ProcessFrame();
+                            t.RunSingleFrame();
                     }
 
                     break;
@@ -117,7 +117,7 @@ namespace osu.Framework.Platform
 
                 case ExecutionMode.MultiThreaded:
                     // still need to run the main/input thread on the window loop
-                    mainThread.ProcessFrame();
+                    mainThread.RunSingleFrame();
                     break;
             }
         }
--- a/osu.Framework/Threading/GameThread.cs
+++ b/osu.Framework/Threading/GameThread.cs
@@ -192,7 +192,8 @@ namespace osu.Framework.Threading
         internal bool ProcessFrame()
         {
             if (state.Value != GameThreadState.Running)
-                throw new InvalidOperationException("Cannot process frames when thread is not running");
+                // host could be in a suspended state. the input thread will still make calls to ProcessFrame so we can't throw.
+                return false;
 
             try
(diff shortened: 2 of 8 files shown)


osu.Framework/Threading/GameThread.cs
```diff
@@ -177,20 +184,42 @@ internal virtual void MakeCurrent()
...
-                if (exitRequested)
+                if (exitRequested || pauseRequested)
                 {
-                    PerformExit();
-                    exitCompleted = true;
-                    return;
+                    lock (startStopLock)
+                    {
+                        if (state.Value != GameThreadState.Running)
+                            throw new InvalidOperationException($"Attempted to process frame when state is {state.Value}");
```
bdach, line 206: Not sure about this. I *think* I see why the outer guard returns `false` but this throws (because nobody should be calling `ProcessFrame()` from multiple threads, so there should be no possibility of a read/write race between that outer guard and this inner one) but it still feels dicey a bit...

osu.Framework/Threading/GameThread.cs
```diff
@@ -92,11 +93,24 @@ public double InactiveHz
...
 
         internal void Initialize(bool withThrottling)
         {
-            MakeCurrent();
+            lock (startStopLock)
+            {
+                Debug.Assert(state.Value != GameThreadState.Running);
+                Debug.Assert(state.Value != GameThreadState.Exited);
+
+                MakeCurrent();
```
bdach, line 110: Is there a particular reason why this is inside the lock, but `ProcessFrame()`'s usage isn't?

osu.Framework/Threading/GameThread.cs
```diff
@@ -142,31 +156,24 @@ private void createThread()
...
+            while (true)
             {
-                Initialize(true);
+                bool shouldContinue = ProcessFrame();
 
-                while (!exitCompleted && !pauseRequested)
+                if (!shouldContinue)
                 {
-                    ProcessFrame();
+                    Thread = null;
```
bdach, line 173: This nulling out of the thread does not take `startStopLock` like it did before via `Cleanup`, but the other set does take that lock.

bdach:
I've pointed out some things that seemed like inconsistencies. I don't know if I'm just tired out of my mind right now, or this is just plain hard, but I'm having a really hard time keeping a mental model of all of the operations and locking on `GameThread`, so I don't really consider this a proper/final review.

osu.Framework/Platform/ThreadRunner.cs
```diff
@@ -175,7 +184,6 @@ private void ensureCorrectExecutionMode()
                     foreach (var t in Threads)
                     {
                         t.Start();
-                        t.Clock.Throttling = true;
```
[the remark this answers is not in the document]
peppy: it can be left here if you want. nothing was really moved; it was being set until now twice, redundantly.

osu.Framework/Threading/GameThread.cs
```diff
@@ -20,9 +20,18 @@ public class GameThread
...
+        /// The current state of this thread.
+        /// </summary>
+        public IBindable<GameThreadState> State => state;
+
+        private readonly Bindable<GameThreadState> state = new Bindable<GameThreadState>();
+
         internal PerformanceMonitor Monitor { get; }
         public ThrottledFrameClock Clock { get; }
 
+        protected virtual bool UsesNativeThread => true;
```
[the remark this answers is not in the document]
peppy: it was required, as previously the input thread was overriding and thus bypassing state change logic. it should be about-as-readable as before, hopefully.

osu.Framework/Threading/GameThread.cs
peppy: MakeCurrent sets a required flag (ThreadSafety I believe). If you move it back you'll notice it throw in the tests.

peppy: As for why `ProcessFrame()` isn't: we don't want to be locking on anything which runs every frame. While the overhead is small, it does exit.

peppy:
@bdach I've kind of rewritten stuff again (and triggered the `XL` size unfortunately). The goal was to increase the amount of locking, but decrease the locations at which state changes can occur. I'm willing to give it another shot if this still doesn't feel good.

osu.Framework/Threading/GameThread.cs
peppy: Now locked.

osu.Framework/Threading/GameThread.cs
peppy: This whole method is now locked.

osu.Framework/Threading/GameThread.cs
peppy: Has been refactored and hopefully reads better now.

osu.Framework/Platform/ThreadRunner.cs
```diff
@@ -158,12 +163,16 @@ public void Stop()
...
+            lock (startStopLock)
+            {
+                if (ExecutionMode == activeExecutionMode)
+                    return;
 
-            // if null, we have not yet got an execution mode, so set this early to allow usage in GameThread.Initialize overrides.
-            activeExecutionMode ??= ThreadSafety.ExecutionMode = ExecutionMode;
-            Logger.Log($"Execution mode changed to {activeExecutionMode}");
+                // if null, we have not yet got an execution mode, so set this early to allow usage in GameThread.Initialize overrides.
+                activeExecutionMode ??= ThreadSafety.ExecutionMode = ExecutionMode;
```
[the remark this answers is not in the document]
peppy: The log line directly below... also looks like it would read incorrectly for all but the first change? It looks like changing this early in all cases doesn't have any side effects, so I've switched to doing that instead.

osu.Framework/Threading/GameThread.cs
```diff
@@ -128,183 +162,327 @@ internal GameThread(Action onNewFrame = null, string name = "unknown", bool moni
...
+
+                state.Value = GameThreadState.Starting;
             }
-            finally
-            {
-                OnPause();
-                Cleanup();
-            }
+
+            PrepareForWork();
```
bdach, line 203: This call creates and kicks off the thread so I believe the thread-creation part should be under `startStopLock`, too. The operation that unsets `Thread` back to `null` (`setExitState()`) takes that lock already.

osu.Framework/Threading/GameThread.cs
```diff
@@ -128,183 +162,327 @@ internal GameThread(Action onNewFrame = null, string name = "unknown", bool moni
...
         {
-            if (Thread == null || culture == null) return;
+            if (state.Value == targetState)
+                return;
 
-            Thread.CurrentCulture = culture;
-            Thread.CurrentUICulture = culture;
+            if (Thread == null)
+            {
+                // if the thread is null at this point, presume the call was made from the thread context.
```
bdach, line 315: Not sure what "thread context" is supposed to mean in this case. Is it "single-thread context"?

osu.Framework/Threading/GameThread.cs
```diff
@@ -128,183 +162,327 @@ internal GameThread(Action onNewFrame = null, string name = "unknown", bool moni
...
+
+            Thread = new Thread(runWork)
             {
-                Thread = null;
-                Running = false;
+                Name = PrefixedThreadNameFor(Name),
+                IsBackground = true,
+            };
+
+            updateCulture();
```
bdach, line 391: Seems redundant. `Initialize()` calls this too.

osu.Framework/Threading/GameThread.cs
```diff
@@ -128,183 +162,327 @@ internal GameThread(Action onNewFrame = null, string name = "unknown", bool moni
...
         }
 
-        public void Exit() => exitRequested = true;
-
-        public virtual void Start()
+        /// <summary>
+        /// Process a single frame of this thread's work.
+        /// </summary>
+        /// <returns>A potential execution state change.</returns>
+        private GameThreadState? processFrame()
```
bdach, line 406: I think this could just as well have a non-nullable return value and keep returning `Running` wherever it returns `null` now. Not a big deal, though.

bdach:
Dunno if I was just dead when I was reading the last version, but I gotta say overall this one reads much better/nicer than the previous revision. I haven't read it too long, just had a loose scroll, but I have more confidence in the locking this time round.

osu.Framework/Platform/ThreadRunner.cs
```diff
@@ -124,10 +126,13 @@ public virtual void RunMainLoop()
...
         public void Suspend()
         {
-            pauseAllThreads();
+            lock (startStopLock)
+            {
+                pauseAllThreads();
 
-            // set the active execution mode back to null to set the state checking back to when it can be resumed.
-            activeExecutionMode = null;
+                // set the active execution mode back to null to set the state checking back to when it can be resumed.
```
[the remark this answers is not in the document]
peppy: I honestly have no idea. Maybe just remove the comment altogether?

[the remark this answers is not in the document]
bdach: Works for me. I think the intent of the set is relatively clear-cut as is.

osu.Framework/Threading/GameThread.cs
peppy: I don't believe it can be, because  in multi-threaded execution this call also has a wait-for-state operation which would deadlock with the state change to running (in `Initialize`).

peppy: Actually the wait call may be able to be moved out, will try that.

bdach: Yeah I saw that, hence the mention of "the thread-creation part".

osu.Framework/Threading/GameThread.cs
peppy: It's mean to read that the execution context / caller is on the "thread" itself. Maybe best to just say "single thread mode" explicitly?

bdach: Yeah I think so. It's just that's the only usage of the phrase "thread context" in the entire file so it's a bit cryptic in that respect.

osu.Framework/Threading/GameThread.cs
peppy: This is a really weird one. It's copied from `master` but I'm unsure if it's intentionally doing this or not. Either way, it's not actually redundant. This is going to be setting the *caller* thread to the culture specified here. Which means the culture is going to leak onto the main thread (that is calling the `Start()` operation). Maybe intentional to make exceptions read in the specified culture, or maybe a slip-up?

Will remove this but also cc @smoogipoo to make sure it gets one more set of eyes.

osu.Framework/Threading/GameThread.cs
peppy: Neutral on this. Not sure which reads better.

osu.Framework/Threading/GameThread.cs
peppy: I've reworded this slightly, including hopefully more context as to why a comment is required in the first place.

osu.Framework/Threading/GameThread.cs
bdach: Oh that's a good call on being called on a different thread, but I'd like to think even if it should be done on the caller thread, it's already done on a higher level [in `ThreadRunner`](https://github.com/ppy/osu-framework/blob/f07e42be5345c4c1ef5d547c8abb541890afe9d3/osu.Framework/Platform/ThreadRunner.cs#L226-L241) as well...? Really weird but all in all shouldn't matter as long as all threads receive it at some point I suppose...

bdach (approved):

smoogipoo (approved):
lgtm

Issue #4543, opened 2021-06-24
Borderless window can be restored if it was maximized when windowed


Susko3:
Only happens on Windows, because of the borderless hack.
Discovered when I was testing https://github.com/ppy/osu/issues/13638

To repro:
1. Set to windowed
2. Maximize window
3. Press F11 to enter borderless
4. Press <kbd>Win</kbd>+<kbd>&darr;</kdb>

Example of restored borderless window:

![image](https://user-images.githubusercontent.com/16479013/123266220-1c811200-d4fc-11eb-8e7c-47f91ee6e40e.png)

Susko3:
This is a problem with SDL (or maybe with Windows) as a non-resizable maximized SDL window can be restored by dragging on the title bar, or with <kbd>Win</kbd>+<kbd>&darr;</kbd>.

One can test this with `visual-tests` &rarr; `Platform` &rarr; `ResizableWindow`

We may want to force restore the window before entering borderless mode.

peppy:
Not sure if this is something we need to be fixing. I don't think this is specific to osu! / osu!framework and can be repro'd in any application which supports borderless display.

If you need another way to test this kind of thing, I have a little utility app which can move/resize/break any application (https://github.com/peppy/pWindowJax) it's useful for testing these kinds of things.

PR #4544, opened 2021-06-25
Restore window before maximising to ensure the correct size is used

peppy:
Resolves https://github.com/ppy/osu/issues/13638.

https://github.com/ppy/osu-framework/issues/4543 may touch more on the underlying issues, which is likely that the maximised state we are managing internally is flimsy during non-windowed display modes.

I'm just looking to fix this in the most resilient way possible, and this change does that. Appreciate testing on linux to make sure it doesn't regress anything (if there's even a concept of "maximised" conveyed there..)


frenzibyte (approved):
Looks to be fine on my Debian box, macOS behaves strangely with maximised and borderless already so that can be ignored for now.

bdach (approved):
looks to work as expected on Windows and not regress Linux (the issue doesn't seem to be there on my ubuntu install on current `master`)

PR #4545, opened 2021-06-26
Differentiate between maximised and fullscreen borderless state in macOS

frenzibyte:
Considering them the same looks to be wrong and have been causing state management problems (e.g. switching to borderless in a maximised window applies no effect).

For confirmation, I've added a "current window state" to `TestSceneFullscreen`, and started playing with the window states while checking what the `WindowState` value is, after applying this change. Looks to be correct in all scenarios.

https://user-images.githubusercontent.com/22781491/123504990-3427da00-d665-11eb-84b6-93dcd1807508.mp4

### DIFF
--- a/osu.Framework/Platform/SDL2/SDL2Extensions.cs
+++ b/osu.Framework/Platform/SDL2/SDL2Extensions.cs
@@ -448,8 +448,7 @@ namespace osu.Framework.Platform.SDL2
         {
             // NOTE: on macOS, SDL2 does not differentiate between "maximised" and "fullscreen desktop"
             if (windowFlags.HasFlagFast(SDL.SDL_WindowFlags.SDL_WINDOW_FULLSCREEN_DESKTOP) ||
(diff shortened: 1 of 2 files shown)


osu.Framework/Platform/SDL2/SDL2Extensions.cs
```diff
@@ -448,8 +448,7 @@ public static WindowState ToWindowState(this SDL.SDL_WindowFlags windowFlags)
         {
             // NOTE: on macOS, SDL2 does not differentiate between "maximised" and "fullscreen desktop"
```
peppy, line 449: This looks outdated now?

frenzibyte: Whoops, forgot about that one.

peppy:
Seems to work more correctly with this change.

Bonus points if you can find why that comment was there, and when the underlying behaviour changed.

peppy (approved):

PR #4546, opened 2021-06-28
Fix front-to-back incorrectly shrinking negative-sized draw rectangles

smoogipoo:
Fixes the issue found in https://github.com/ppy/osu-framework/pull/4494#issuecomment-858338705

Confirmation via renderdoc in a modified testscene from #4494 (notice the backbuffer depth areas match):
Screenshot_2021-06-28_12-48-05
(left: positive size, right: negative size)


osu.Framework/Graphics/Primitives/RectangleF.cs
```diff
@@ -228,39 +228,152 @@ public RectangleF WithPositiveExtent
...
             Y - amount.Top,
             Width + amount.TotalHorizontal,
             Height + amount.TotalVertical);
 
+        /// <summary>
+        /// Gets a <see cref="RectangleF"/> with both <see cref="Width"/> and <see cref="Height"/> shrunk in the negative direction.
+        /// The centre of the resultant <see cref="RectangleF"/> remains unchanged.
+        /// </summary>
+        /// <remarks>
+        /// If either <see cref="Width"/> or <see cref="Height"/> of this <see cref="RectangleF"/> is negative, the resultant <see cref="RectangleF"/> is inflated in the corresponding direction.
```
peppy, line 272: Does this mean "for each of the directions"? By wording it "If either" it sounds like one direction being negative has an affect on both (I get that it doesn't from reading the code and the "corresponding" wording at the end, but it reads pretty awkwardly to the point if I was to use this I would need to confirm via code what it's doing).

ekrctb:
I'm still getting the blackness (although it is smaller than before) in the `Masking` test scene when the box is thin (small size).

smoogipoo:
> I'm still getting the blackness

Turns out that that was caused by two more failures - the shrinkage being too big causing sizes to go negative, and after fixing that, the polygon clipper returning practically infinite sizes for points/lines.

Both have been fixed.

osu.Framework/Utils/ConvexPolygonClipper.cs
```diff
@@ -94,6 +94,9 @@ public Span<Vector2> Clip(in Span<Vector2> buffer)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private int processClipEdge(in Line clipEdge, in Span<Vector2> buffer, in int inputCount)
         {
+            if (clipEdge.EndPoint == clipEdge.StartPoint)
+                return 0;
```
ekrctb, line 98: I don't think returning an empty polygon is correct.
```csharp
            assertPolygonEquals(new SimpleConvexPolygon(new Vector2[] { Vector2.Zero, new Vector2(0, 1), new Vector2(1, 0) }),
                new SimpleConvexPolygon(clip(new Quad(Vector2.Zero, Vector2.Zero, new Vector2(1, 0), new Vector2(0, 1)), new Quad(0, 0, 10, 10)).ToArray()),
                false);
```
Though it is enough "approximation" for the purpose of front-to-back pass I guess?

smoogipoo: Good catch.

ekrctb (approved):

PR #4547, opened 2021-06-28
Expose virtual method in `MarkdownImage`

gagahpangeran:
In order to resolve (https://github.com/ppy/osu/issues/13418), need these method so I can override some parts in `MarkdownImage`

I don't have the solution for the actual issue yet, still figuring out about it.

### DIFF
--- a/osu.Framework/Graphics/Containers/Markdown/MarkdownImage.cs
+++ b/osu.Framework/Graphics/Containers/Markdown/MarkdownImage.cs
@@ -27,8 +27,10 @@ namespace osu.Framework.Graphics.Containers.Markdown
         [BackgroundDependencyLoader]
         private void load()
         {
-            InternalChild = new DelayedLoadWrapper(CreateImageContainer(url));
+            InternalChild = CreateContent(url);
         }
+
+        protected virtual Drawable CreateContent(string url) => new DelayedLoadWrapper(CreateImageContainer(url));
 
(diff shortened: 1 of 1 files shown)


peppy:
Will wait to see how these are used before merging. You are aware of how to use a local framework checkout yeah?

gagahpangeran:
Yeah, PR-ing this for visibility now.

frenzibyte:
Probably worth getting a working implementation first, then start PR'ing the required parts for discussing it. (since either way, the usage of these changes is probably required for considering this and whether it should be abstracted this way etc.)

osu.Framework/Graphics/Containers/Markdown/MarkdownImage.cs
```diff
@@ -48,20 +50,26 @@ public ImageContainer(string url)
...
+                InternalChild = image = CreateSpriteImage();
             }
 
             [BackgroundDependencyLoader]
             private void load(TextureStore textures)
+            {
+                image.Texture = GetImageTexture(textures, url);
+            }
+
+            protected virtual Sprite CreateSpriteImage() => new Sprite();
```
smoogipoo, line 62: I think `CreateImageSprite()` makes more sense.

smoogipoo (changes requested):

smoogipoo (approved):

PR #4549, opened 2021-06-29
Avoid running exit sequence when it has already been run once

peppy:
I'm not sure if this had any adverse side effects (seems like it *should* have), but should make things saner in general regardless. Logging can be removed if preferred, but it seems like something good to have around.

Previously:

```csharp
[runtime] 2021-06-29 10:07:10 [verbose]: Host execution state changed to Running
[runtime] 2021-06-29 10:07:10 [verbose]: Execution mode changed to MultiThreaded
[runtime] 2021-06-29 10:07:10 [verbose]: BASS Initialized
[runtime] 2021-06-29 10:07:10 [verbose]: BASS Version:               2.4.15.26
[runtime] 2021-06-29 10:07:10 [verbose]: BASS FX Version:            2.4.12.3
[runtime] 2021-06-29 10:07:10 [verbose]: Device:                     No sound
[runtime] 2021-06-29 10:07:10 [verbose]: Drive:
[runtime] 2021-06-29 10:07:10 [verbose]: Host execution state changed to Stopping
[runtime] 2021-06-29 10:07:10 [verbose]: Host execution state changed to Stopping
[runtime] 2021-06-29 10:07:10 [verbose]: Host execution state changed to Stopped
[runtime] 2021-06-29 10:07:10 [verbose]: Host execution state changed to Stopping
[runtime] 2021-06-29 10:07:10 [verbose]: Host execution state changed to Stopped
```

Now:

```csharp
[runtime] 2021-06-29 10:08:18 [verbose]: Host execution state changed to Running
[runtime] 2021-06-29 10:08:18 [verbose]: Execution mode changed to MultiThreaded
[runtime] 2021-06-29 10:08:18 [verbose]: BASS Initialized
[runtime] 2021-06-29 10:08:18 [verbose]: BASS Version:               2.4.15.26
[runtime] 2021-06-29 10:08:18 [verbose]: BASS FX Version:            2.4.12.3
[runtime] 2021-06-29 10:08:18 [verbose]: Device:                     No sound
[runtime] 2021-06-29 10:08:18 [verbose]: Drive:
[runtime] 2021-06-29 10:08:18 [verbose]: Host execution state changed to Stopping
[runtime] 2021-06-29 10:08:18 [verbose]: Host execution state changed to Stopped
```

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -497,7 +497,15 @@ namespace osu.Framework.Platform
             }
         }
 
-        public ExecutionState ExecutionState { get; private set; }
+        public ExecutionState ExecutionState
+        {
+            get => executionState;
+            private set
+            {
+                executionState = value;
+                Logger.Log($"Host execution state changed to {value}");
+            }
+        }
 
         /// <summary>
         /// Schedules the game to exit in the next frame.
@@ -935,6 +943,7 @@ namespace osu.Framework.Platform
         private bool isDisposed;
 
         private readonly ManualResetEventSlim stoppedEvent = new ManualResetEventSlim(false);
+        private ExecutionState executionState;
 
         protected virtual void Dispose(bool disposing)
         {
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -518,7 +518,7 @@ namespace osu.Framework.Platform
         /// Schedules the game to exit in the next frame (or immediately if <paramref name="immediately"/> is true).
         /// </summary>
         /// <param name="immediately">If true, exits the game immediately.  If false (default), schedules the game to exit in the next frame.</param>
-        protected void PerformExit(bool immediately)
+        protected virtual void PerformExit(bool immediately)
         {
             switch (ExecutionState)
             {
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -508,13 +508,21 @@ namespace osu.Framework.Platform
         /// Schedules the game to exit in the next frame (or immediately if <paramref name="immediately"/> is true).
         /// </summary>
         /// <param name="immediately">If true, exits the game immediately.  If false (default), schedules the game to exit in the next frame.</param>
-        protected virtual void PerformExit(bool immediately)
+        protected void PerformExit(bool immediately)
         {
+            switch (ExecutionState)
+            {
+                case ExecutionState.Stopping:
+                case ExecutionState.Stopped:
+                    return;
+            }
+
+            ExecutionState = ExecutionState.Stopping;
+
             if (immediately)
                 exit();
             else
             {
-                ExecutionState = ExecutionState.Stopping;
                 InputThread.Scheduler.Add(exit, false);
             }
         }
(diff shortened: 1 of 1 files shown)


smoogipoo (dismissed):

smoogipoo:
CI is stuck.

peppy:
Fine to cancel these runs for now, should be a simple cause.

peppy:
@smoogipoo requesting you take a quick 5-10 minute look into this one and provide some insight if you can. seems to be fixed by removing some weird exception related logic, but I think it may break stuff, just can't figure out how to repro the breakage.

20210629 205617 (iTerm app)

https://github.com/ppy/osu-framework/commit/d309ea4f6663c9b5b788d90fd5727527b4602db5

smoogipoo:
Maybe this explains it? https://github.com/ppy/osu-framework/pull/3416

peppy:
Right. That's a very dodgy fix for a very dodgy issue. May need to re-address that in a better way.

peppy:
I've fixed the remaining case by changing how the exit process is done for exception funneling. I think this should be clearer now (the exit is always done in the main `Run()` method now, rather than in two places).

To test this, you can replace the `SampleGameGame` with this:

```csharp

using System;
using osu.Framework;
using osu.Framework.Configuration;
using osu.Framework.Graphics;
using osu.Framework.Platform;
using osuTK;
using osuTK.Graphics;

namespace SampleGame
{
    public class SampleGameGame : Game
    {
        private Box box;

        [BackgroundDependencyLoader]
        private void load(FrameworkConfigManager config)
        {
            Add(box = new Box
            {
                Anchor = Anchor.Centre,
                Origin = Anchor.Centre,
                Size = new Vector2(150, 150),
                Colour = Color4.Tomato
            });

            config.SetValue(FrameworkSetting.ExecutionMode, ExecutionMode.SingleThread);
        }

        {
            base.LoadComplete();

            Scheduler.AddDelayed(() => throw new InvalidOperationException(), 5000);
        }

        {
            base.Update();
            box.Rotation += (float)Time.Elapsed / 10;
        }
    }
}
```

peppy:
Re-reading the surrounding comments (https://github.com/ppy/osu-framework/blob/c59dca9c73f1448b5c7b3557c86c61fd34a53ba2/osu.Framework/Platform/GameHost.cs#L298-L299) I'm not sure this behaves correctly in all cases, but I can't seem to reproduce the issue with rider.

Comments need updating/removal at very least if this is actually correct.

bdach (dismissed):
looks logical to me :+1:

bdach:
I initially approved but noticed very late that the linux multi-threaded github actions job has been running for 2 hours and counting now, which is worrying as I can't really see much fault with this diff as is, to tell the truth...

bdach:
I think I repro'd at least one failure. `TestLogging.TestUnhandledIgnoredOnceException` can hang on a race. I'm a bit fuzzy on the details, but the short and short of it is that it is possible for the `GameHost` to enter a permanently-hanged state when a non-immediate and an immediate exit are requested one after another. The scenario is:

1. `GameHost.Exit()` is called, which calls `PerformExit(immediately: false)`, which sets `ExecutionState` to `Stopping` and schedules `exit()` to be ran on the input thread.
2. Following that, something somewhere throws an unallowable exception, which is rethrown on the input thread. The exception re-throw causes the game loop to be interrupted, and `PerformExit(immediately: true)` is then called.
3. The second `PerformExit()`, despite being immediate, is a no-op, because the first call has already set `ExecutionState`. But that `exit()` scheduled by the first call will never run, because the game loop is now dead.

The simplest fix would probably be to move the set of `ExecutionState` into `exit()`.

peppy:
I don't think we want to set `ExecutionState` later in the process, as it feels more correct for `Stopping` to be set *before* the scheduled `exit` completes. I've gone with the first strategy, of allowing immediate exit to still run even after a scheduled exit was previously performed.

I think this still reads fine.

smoogipoo (approved):
Seems fine

PR #4550, opened 2021-06-29
Add support for localising `IFormattable` objects

frenzibyte:
Introduces a `LocalisableFormattable` class, containing the `IFormattable` object and the format string, and a method for formatting the `IFormattable` with the provided format provider.

Allowing for localising any `IFormattable` object (integers, floats, `DateTime`s, etc...) with the currently selected locale, like in [here](https://github.com/ppy/osu/commit/feb67bee58f071c0cfddba75dccdaf648275a0e5) for example.

---

I've initially written this for resolving https://github.com/ppy/osu/issues/13559, but I think it's worth actually using this for localising all `IFormattable`-formatted strings osu!-side (i.e. `double.ToString("N0")` and otherwise) while at it, will look into that now.

### DIFF
--- a/osu.Framework/Localisation/ILocalisableStringData.cs
+++ b/osu.Framework/Localisation/ILocalisableStringData.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using osu.Framework.Configuration;
 
 #nullable enable
 
@@ -16,7 +17,7 @@ namespace osu.Framework.Localisation
         /// Gets a localised <see cref="string"/> using the given localisation store and other required data.
         /// </summary>
         /// <param name="store">The localisation store.</param>
-        /// <param name="preferUnicode">Whether unicode is preferred if available.</param>
-        string GetLocalised(ILocalisationStore? store, bool preferUnicode);
+        /// <param name="config">The config manager, used for retrieving current value of certain settings (i.e. <see cref="FrameworkSetting.ShowUnicode"/>.</param>
+        string GetLocalised(ILocalisationStore? store, FrameworkConfigManager config);
     }
 }
--- a/osu.Framework/Localisation/ILocalisableStringData.cs
+++ b/osu.Framework/Localisation/ILocalisableStringData.cs
@@ -1,0 +1,22 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+
+#nullable enable
+
+namespace osu.Framework.Localisation
+{
+    /// <summary>
+    /// An interface for <see cref="LocalisableString"/>'s data.
+    /// </summary>
+    public interface ILocalisableStringData : IEquatable<ILocalisableStringData>
+    {
+        /// <summary>
+        /// Gets a localised <see cref="string"/> using the given localisation store and other required data.
+        /// </summary>
+        /// <param name="store">The localisation store.</param>
+        /// <param name="preferUnicode">Whether unicode is preferred if available.</param>
+        string GetLocalised(ILocalisationStore? store, bool preferUnicode);
+    }
+}
--- a/osu.Framework/Localisation/ILocalisableStringData.cs
+++ b/osu.Framework/Localisation/ILocalisableStringData.cs
@@ -2,7 +2,6 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using osu.Framework.Configuration;
 
 #nullable enable
 
@@ -16,8 +15,7 @@ namespace osu.Framework.Localisation
         /// <summary>
         /// Gets a localised <see cref="string"/> using the given localisation store and other required data.
         /// </summary>
-        /// <param name="store">The localisation store.</param>
-        /// <param name="config">The config manager, used for retrieving current value of certain settings (i.e. <see cref="FrameworkSetting.ShowUnicode"/>.</param>
-        string GetLocalised(ILocalisationStore? store, FrameworkConfigManager config);
+        /// <param name="parameters">Any parameters that control the localisation method.</param>
+        string GetLocalised(LocalisationParameters parameters);
     }
 }
--- a/osu.Framework/Localisation/LocalisableFormattable.cs
+++ b/osu.Framework/Localisation/LocalisableFormattable.cs
@@ -47,22 +47,8 @@ namespace osu.Framework.Localisation
                    Format == other.Format;
         }
 
-        public bool Equals(ILocalisableStringData? other)
-        {
-            if (ReferenceEquals(null, other)) return false;
-            if (other.GetType() != GetType()) return false;
-
-            return Equals((LocalisableFormattable)other);
-        }
-
-        public override bool Equals(object? obj)
-        {
-            if (ReferenceEquals(null, obj)) return false;
-            if (ReferenceEquals(this, obj)) return true;
-            if (obj.GetType() != GetType()) return false;
-
-            return Equals((LocalisableFormattable)obj);
-        }
+        public bool Equals(ILocalisableStringData? other) => other is LocalisableFormattable formattable && Equals(formattable);
+        public override bool Equals(object? obj) => obj is LocalisableFormattable formattable && Equals(formattable);
 
         public override int GetHashCode() => HashCode.Combine(Value, Format);
     }
--- a/osu.Framework/Localisation/LocalisableFormattable.cs
+++ b/osu.Framework/Localisation/LocalisableFormattable.cs
@@ -49,7 +49,7 @@ namespace osu.Framework.Localisation
         {
             if (ReferenceEquals(null, obj)) return false;
             if (ReferenceEquals(this, obj)) return true;
-            if (obj.GetType() != this.GetType()) return false;
+            if (obj.GetType() != GetType()) return false;
 
             return Equals((LocalisableFormattable)obj);
         }
--- a/osu.Framework/Localisation/LocalisableFormattable.cs
+++ b/osu.Framework/Localisation/LocalisableFormattable.cs
@@ -1,0 +1,59 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+
+namespace osu.Framework.Localisation
+{
+    /// <summary>
+    /// A string allowing for formatting <see cref="IFormattable"/>s using the current locale.
+    /// </summary>
+    public class LocalisableFormattable : IEquatable<LocalisableFormattable>
+    {
+        public readonly IFormattable Value;
+        public readonly string Format;
+
+        /// <summary>
+        /// Creates a <see cref="LocalisableFormattable"/> with an <see cref="IFormattable"/> value and a format string.
+        /// </summary>
+        /// <param name="value">The <see cref="IFormattable"/> value.</param>
+        /// <param name="format">The format string.</param>
+        public LocalisableFormattable(IFormattable value, string format)
+        {
+            Value = value;
+            Format = format;
+        }
+
+        public string ToString(IFormatProvider formatProvider)
+        {
+            if (formatProvider == null)
+                return ToString();
+
+            return Value.ToString(Format, formatProvider);
+        }
+
+        public override string ToString() => Value.ToString(Format, CultureInfo.InvariantCulture);
+
+        public bool Equals(LocalisableFormattable other)
+        {
+            if (ReferenceEquals(null, other)) return false;
+            if (ReferenceEquals(this, other)) return true;
+
+            return EqualityComparer<object>.Default.Equals(Value, other.Value) &&
+                   Format == other.Format;
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            if (obj.GetType() != this.GetType()) return false;
+
+            return Equals((LocalisableFormattable)obj);
+        }
+
+        public override int GetHashCode() => HashCode.Combine(Value, Format);
+    }
+}
--- a/osu.Framework/Localisation/LocalisableFormattable.cs
+++ b/osu.Framework/Localisation/LocalisableFormattable.cs
@@ -38,8 +38,6 @@ namespace osu.Framework.Localisation
 
         public override string ToString() => Value.ToString(Format, CultureInfo.InvariantCulture);
 
-        public static implicit operator LocalisableString(LocalisableFormattable formattable) => new LocalisableString(formattable);
-
         public bool Equals(LocalisableFormattable? other)
         {
             if (ReferenceEquals(null, other)) return false;
--- a/osu.Framework/Localisation/LocalisableFormattable.cs
+++ b/osu.Framework/Localisation/LocalisableFormattable.cs
@@ -6,6 +6,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using osu.Framework.Configuration;
 
 namespace osu.Framework.Localisation
 {
@@ -28,7 +29,7 @@ namespace osu.Framework.Localisation
             Format = format;
         }
 
-        public string GetLocalised(ILocalisationStore? store, bool preferUnicode)
+        public string GetLocalised(ILocalisationStore? store, FrameworkConfigManager config)
         {
             if (store == null)
                 return ToString();
--- a/osu.Framework/Localisation/LocalisableFormattable.cs
+++ b/osu.Framework/Localisation/LocalisableFormattable.cs
@@ -1,6 +1,8 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+#nullable enable
+
 using System;
 using System.Collections.Generic;
 using System.Globalization;
@@ -10,7 +12,7 @@ namespace osu.Framework.Localisation
     /// <summary>
     /// A string allowing for formatting <see cref="IFormattable"/>s using the current locale.
     /// </summary>
-    public class LocalisableFormattable : IEquatable<LocalisableFormattable>
+    public class LocalisableFormattable : IEquatable<LocalisableFormattable>, ILocalisableStringData
     {
         public readonly IFormattable Value;
         public readonly string Format;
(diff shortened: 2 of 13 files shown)


osu.Framework/Localisation/LocalisationManager_LocalisedBindableString.cs
```diff
@@ -43,6 +43,10 @@ private void updateValue()
                         Value = translatable.Format(storage.Value);
                         break;
 
+                    case LocalisableFormattable formattable:
+                        Value = formattable.ToString(storage.Value?.EffectiveCulture);
+                        break;
```
smoogipoo, line 48: I don't want to see another case being added here, or anywhere in the other places it's been added.

- There needs to be an `ILocalisableStringData` interface which these "data" classes implement.
- `LocalisableString` needs to have a ctor or static method to convert from an `ILocalisableStringData` type.
- The data types need to have implicit operators to convert to `LocalisableString` via said ctor/static method.
- The interface needs to have methods for doing the localisation. No more `.GetPreferred()`/`.Format()`/`.ToString()` - just one method common to the data type interface.

osu.Framework/Localisation/LocalisableStringEqualityComparer.cs
```diff
@@ -64,6 +67,10 @@ public int GetHashCode(LocalisableString obj)
                 case RomanisableString romanisable:
                     hashCode.Add(romanisable.GetHashCode());
                     break;
+
+                case LocalisableFormattable formattable:
+                    hashCode.Add(formattable.GetHashCode());
+                    break;
```
smoogipoo, line 73: I believe all of this code is actually completely pointless, and could just be `hashCode.Add(obj.GetHashCode())`.

Need to run through `BenchmarkLocalisableString`, and add more cases to the benchmark for different types.

osu.Framework/Localisation/LocalisableStringEqualityComparer.cs
```diff
@@ -38,6 +38,9 @@ public bool Equals(LocalisableString x, LocalisableString y)
 
                 case RomanisableString romanisableX:
                     return romanisableX.Equals((RomanisableString)yData);
+
+                case LocalisableFormattable formattableX:
+                    return formattableX.Equals((LocalisableFormattable)yData);
             }
 
             return false;
```
smoogipoo, line 46: This should really fall back to `object.Equals(...)` with a `#pragma warning disable` around it.

smoogipoo (changes requested):

smoogipoo:
Honestly, I quite liked the original version using delegates because it could also be humanised. Which may have made the changes I proposed less important but still favourable to have.

frenzibyte:
Where's the original version you're talking about, for reference?

smoogipoo:
The one in https://github.com/ppy/osu/issues/13559#issuecomment-869936958

bdach:
For what it's worth I proposed in that issue thread that the `ToString()` method here could be made virtual and then overridden to call `.Humanize()` instead. But I still really want this structure to be data-backed in some way, and not basing it's equality operator on frivolously-allocated delegates all over.

osu.Framework/Localisation/LocalisationManager_LocalisedBindableString.cs
frenzibyte: Yeah the current implementation is quite inextensible with all the hardcoded cases in place. Didn't want to do it in this PR, but I see no reason not to do it right now and clean up the hardcoded mess.

smoogipoo:
I see, that's fair enough - I agree that a virtual `.ToString()` works better than the delegate. The rest of the comments I made above regarding the interface are still applicable for that to happen.

osu.Framework/Localisation/ILocalisableStringData.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// An interface for <see cref="LocalisableString"/>'s data.
+    /// </summary>
+    public interface ILocalisableStringData : IEquatable<ILocalisableStringData>
+    {
+        /// <summary>
+        /// Gets a localised <see cref="string"/> using the given localisation store and other required data.
+        /// </summary>
+        /// <param name="store">The localisation store.</param>
+        /// <param name="preferUnicode">Whether unicode is preferred if available.</param>
+        string GetLocalised(ILocalisationStore? store, bool preferUnicode);
```
smoogipoo, line 20: I suggest exposing the bool from the store and leaving one parameter.

osu.Framework/Localisation/LocalisableString.cs
```diff
@@ -27,9 +41,6 @@ namespace osu.Framework.Localisation
         public override int GetHashCode() => LocalisableStringEqualityComparer.Default.GetHashCode(this);
 
         public static implicit operator LocalisableString(string text) => new LocalisableString(text);
-        public static implicit operator LocalisableString(TranslatableString translatable) => new LocalisableString(translatable);
-        public static implicit operator LocalisableString(RomanisableString romanisable) => new LocalisableString(romanisable);
-        public static implicit operator LocalisableString(LocalisableFormattable formattable) => new LocalisableString(formattable);
```
smoogipoo: Keep these ones in if they can be, to prevent breaking changes.

frenzibyte: I've moved them to each class, unless that still results in a breaking change I'm failing to see?

smoogipoo:
```
C c = new D();

----------------

public class C
{
    public void M()
    {
    }

    public static implicit operator C(D d) => new C();
}

public class D
{
}

IL_00cb: newobj       instance void D::.ctor()
IL_00d0: call         class C C::op_Implicit(class D)
IL_00d5: stloc.2      // c

----------------

public class C
{
    public void M()
    {
    }
}

public class D
{
    public static implicit operator C(D d) => new C();
}

IL_00cb: newobj       instance void D::.ctor()
IL_00d0: call         class C D::op_Implicit(class D)
IL_00d5: stloc.2      // c
```
So yes, at an IL level implicit operators are just `op_Implicit` method calls on one of the classes.

It seems like you can't define implicit conversions in both classes, so perhaps we leave the ones that are currently where they are, where they are.

osu.Framework/Localisation/ILocalisableStringData.cs
frenzibyte: Exposing the bool from the store doesn't seem straightforward to do, unless it's fine to pass around the `FrameworkConfigManager` / bindable to each localisation store at construction.

And in the end, it doesn't really make sense to have that there... if you're proposing this to avoid adding more parameters here and causing breaking changes, maybe an args class would help?

Alternatively, I could pass the `LocalisationManager` here (maybe as `ILocalisationManager` for staying read-only) and avoid the args class.

osu.Framework/Localisation/LocalisableStringEqualityComparer.cs
frenzibyte: Is this still valid after adding the `ILocalisableStringData`? that already allows for implementing equality comparison at each data class implementing that interface.

osu.Framework/Localisation/LocalisableStringEqualityComparer.cs
frenzibyte: I guess the purpose is for benchmarking equality and mainly hash code calculation of the classes themselves, yeah? i.e.:
```
|                           Method |     Mean |    Error |    StdDev |   Median | Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------------- |---------:|---------:|----------:|---------:|------:|------:|------:|----------:|
|       BenchmarkRomanisableEquals | 16.85 ns | 0.040 ns |  0.034 ns | 16.85 ns |     - |     - |     - |         - |
|      BenchmarkTranslatableEquals | 16.78 ns | 0.040 ns |  0.037 ns | 16.78 ns |     - |     - |     - |         - |
|       BenchmarkFormattableEquals | 40.95 ns | 7.437 ns | 21.928 ns | 32.95 ns |     - |     - |     - |         - |
|  BenchmarkRomanisableGetHashCode | 85.81 ns | 4.947 ns | 14.353 ns | 84.85 ns |     - |     - |     - |         - |
| BenchmarkTranslatableGetHashCode | 85.34 ns | 4.975 ns | 14.592 ns | 86.99 ns |     - |     - |     - |         - |
|  BenchmarkFormattableGetHashCode | 46.82 ns | 2.662 ns |  7.765 ns | 44.52 ns |     - |     - |     - |         - |
```
is that enough benchmarks?

smoogipoo: Yep that's fine.

osu.Framework/Localisation/ILocalisableStringData.cs
bdach: Only one implementation seemingly cares about `preferUnicode`, and it's `RomanisableString`, so this definitely looks bad slapped on this interface.

My gut feeling is that this interface is maybe too unified now, and the romanisable case maybe should still remain independent. But I'd be fine with passing in the config manager through this interface, too.

Not sure what passing in the localisation manager would even do in this context; its bindables are private and should arguably remain so.

frenzibyte: If we can pass the localisation manager (as readonly), we can expose the `PreferUnicode` bindable from there instead of passing the framework config manager, I think of it as a cleaner way, but the other way also works and I don't mind it.

bdach: I don't like the idea of re-exposing config settings via public members on non-config classes too much. Dunno if there is a precedent for that already that I don't know about, but it messes with separation of concerns a bit too much if you ask me.

osu.Framework/Localisation/LocalisableFormattable.cs
```diff
@@ -0,0 +1,69 @@
...
+        public bool Equals(LocalisableFormattable? other)
+        {
+            if (ReferenceEquals(null, other)) return false;
+            if (ReferenceEquals(this, other)) return true;
+
+            return EqualityComparer<object>.Default.Equals(Value, other.Value) &&
+                   Format == other.Format;
+        }
+
+        public bool Equals(ILocalisableStringData? other)
```
bdach, line 50: Is there a reason why `Equals(LocalisableStringData?)` and `Equals(object?)` aren't just a:

```csharp
other is LocalisableFormattable otherFormattable && Equals(otherFormattable);
```

frenzibyte: I guess, I wasn't sure which pattern the equality logic should follow given that `TranslatableString` and `RomanisableString` used one that's quite different from what we usually use here and osu!-side (`ReferenceEquals` rather than pattern-matching).

osu.Framework/Localisation/LocalisationManager_LocalisedBindableString.cs
```diff
@@ -12,19 +13,26 @@ public partial class LocalisationManager
         private class LocalisedBindableString : Bindable<string>, ILocalisedBindableString
         {
             private readonly IBindable<ILocalisationStore?> storage = new Bindable<ILocalisationStore?>();
-            private readonly IBindable<bool> preferUnicode = new Bindable<bool>();
+
+            // ReSharper disable once PrivateFieldCanBeConvertedToLocalVariable (reference must be kept for bindable to not get GC'd)
+            private readonly IBindable<bool> preferUnicode;
```
bdach, line 18: Ok, well, perhaps partially my bad for giving a bad suggestion, but this doesn't make much sense now, does it? Now both this bindable string and `RomanisableString` are reading the same config setting but separately, and it causes this warning disable and a whole mess of other awkwardnesses (like semi-pointlessly holding onto a reference to `config`...)

May be indeed best to cut losses at this point and expose the setting (as internal!) in `LocalisationManager`...

osu.Framework/Localisation/LocalisableStringEqualityComparer.cs
```diff
@@ -49,23 +46,8 @@ public int GetHashCode(LocalisableString obj)
...
-                    hashCode.Add(translatable.GetHashCode());
-                    break;
-
-                case RomanisableString romanisable:
-                    hashCode.Add(romanisable.GetHashCode());
-                    break;
-            }
-
+            hashCode.Add(obj.Data.GetType());
+            hashCode.Add(obj.Data);
```
bdach, line 50: Can replace with a `HashCode.Combine()`

bdach (changes requested):

osu.Framework/Localisation/LocalisationManager_LocalisedBindableString.cs
frenzibyte: > May be indeed best to cut losses at this point and expose the setting (as internal!) in LocalisationManager...

Yep, also keeps things in scope and allows for exposing further game-specific settings if wanted/needed.

frenzibyte: Actually, my bad on the reference holding part, could have just instantiated a local bindable and did `Config.BindWith` (which appears to not spark that inspection).

But either way, here's how it looks with the `LocalisationManager` (passed as `ILocalisationManager`) instead https://github.com/ppy/osu-framework/commit/f388d7138545d3cb413464ec07657024fedc2571, will leave it to you if you think it's an improvement.

bdach: Not sure, feels half-way. Maybe the localisation store should also be exposed via the interface to keep 1 param instead of 2. Other opinions welcome.

smoogipoo: I've implemented my original vision of how this method should look (a72b44d6ec8491a68c6eed594d64ef70376777fc), please check it out and see if you agree.

frenzibyte: I've applied a simple change to avoid the ReSharper inspection suppression, but looks pretty good overall, and allows for extensibility better than the other ways above.

smoogipoo:
Benchmark results (making sure there's no regression between `(string)"a"` and `(LocalisableString)(string)"a"`):
```
|                                Method |      Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------------- |----------:|----------:|----------:|------:|------:|------:|----------:|
|                 BenchmarkStringEquals |  4.075 ns | 0.0377 ns | 0.0314 ns |     - |     - |     - |         - |
|      BenchmarkLocalisableStringEquals |  7.042 ns | 0.0423 ns | 0.0396 ns |     - |     - |     - |         - |
|            BenchmarkRomanisableEquals | 13.177 ns | 0.2306 ns | 0.2157 ns |     - |     - |     - |         - |
|           BenchmarkTranslatableEquals | 10.599 ns | 0.0247 ns | 0.0231 ns |     - |     - |     - |         - |
|            BenchmarkStringGetHashCode |  2.165 ns | 0.0019 ns | 0.0015 ns |     - |     - |     - |         - |
| BenchmarkLocalisableStringGetHashCode | 18.292 ns | 0.1584 ns | 0.1323 ns |     - |     - |     - |         - |
|       BenchmarkRomanisableGetHashCode | 31.252 ns | 0.5610 ns | 0.5247 ns |     - |     - |     - |         - |
|      BenchmarkTranslatableGetHashCode | 39.200 ns | 0.3968 ns | 0.3518 ns |     - |     - |     - |         - |

After:

|                                Method |      Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------------------- |----------:|----------:|----------:|------:|------:|------:|----------:|
|                 BenchmarkStringEquals |  4.367 ns | 0.0135 ns | 0.0120 ns |     - |     - |     - |         - |
|      BenchmarkLocalisableStringEquals |  7.026 ns | 0.0256 ns | 0.0239 ns |     - |     - |     - |         - |
|            BenchmarkRomanisableEquals | 14.885 ns | 0.2912 ns | 0.2582 ns |     - |     - |     - |         - |
|           BenchmarkTranslatableEquals | 15.184 ns | 0.2584 ns | 0.6039 ns |     - |     - |     - |         - |
|            BenchmarkFormattableEquals | 15.978 ns | 0.0047 ns | 0.0041 ns |     - |     - |     - |         - |
|            BenchmarkStringGetHashCode |  2.165 ns | 0.0009 ns | 0.0007 ns |     - |     - |     - |         - |
| BenchmarkLocalisableStringGetHashCode | 11.798 ns | 0.2674 ns | 0.4003 ns |     - |     - |     - |         - |
|       BenchmarkRomanisableGetHashCode | 19.493 ns | 0.3032 ns | 0.2688 ns |     - |     - |     - |         - |
|      BenchmarkTranslatableGetHashCode | 26.869 ns | 0.0941 ns | 0.0834 ns |     - |     - |     - |         - |
|       BenchmarkFormattableGetHashCode | 20.660 ns | 0.2058 ns | 0.1925 ns |     - |     - |     - |         - |
```

smoogipoo (dismissed):
Lgtm

osu.Framework/Localisation/LocalisableFormattable.cs
```diff
@@ -0,0 +1,55 @@
...
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+
+namespace osu.Framework.Localisation
+{
+    /// <summary>
+    /// A string allowing for formatting <see cref="IFormattable"/>s using the current locale.
+    /// </summary>
+    public class LocalisableFormattable : IEquatable<LocalisableFormattable>, ILocalisableStringData
```
smoogipoo, line 15: Maybe `LocalisableFormattableString` is a better fit?

bdach: No strong feelings either way here.

smoogipoo: For comparison, the .NET type (for interpolated strings) is `FormattableString`.

frenzibyte: been wondering how humanisable strings will be named, `LocalisableHumanisable`? `LocalisableHumanisableString`?

smoogipoo: I was thinking `HumanisableString`, actually. I was also thinking `FormattableString` here, but that would conflict with the .NET name, which _may_ be fine since `FormattableString` is a very rarely used thing anyway.

frenzibyte: I wasn't sure because `RomanisableString` and `TranslateableString` are both names that fit correctly and imply being from the localisation system, but names like `HumanisableString` and `FormattableString` feel like they're not conveying enough they're localisable or in the localisation system... that's why I agreed with the original proposal of naming it `LocaliasbleFormattable`.

Not a big deal either way, maybe it makes sense to drop the `Localisable` part altogether since the data classes themselves just allow themselves to be "localised" (by implementing the `ILocalisableStringData`).

bdach (approved):

osu.Framework/Localisation/LocalisableFormattable.cs
peppy: I definitely thing this should be renamed to either `FormattableString` or `LocalisableFormattableString`. Indifferent as to which (second fits the existing classes better, but is long, but probably also rare usage so i don't know).

peppy:
I'm honestly not sure about the `parameters` flow. At this point it's all a bit hard to understand, but rather than hold things up let's go ahead with it. Just need to sort out the rename issue.

osu.Framework/Localisation/LocalisableFormattable.cs
frenzibyte: I believe all usages that do stuff like `number.ToString("N0")` would use this string to display properly per-locale, so the usage may not be too rare... but still better to just consider `LocalisableFormattableString` at this point.

peppy: Choose one and go with it. I guess the longer one would be preferable for consistency if no one else want to make a call.

bdach: Agree with longer one.

bdach:
Have made the name change.

smoogipoo (approved):

smoogipoo (approved):

PR #4551, opened 2021-06-29
Fix TrackVirtual seeks failing in unexpected ways

smoogipoo:
- `Seek(CurrentTime)` failed even though the time is a valid value.
- `Seek(EndTime+1000)` succeeded even though the time is an invalid value.
  - Following `TrackBass`, this should fail but seek to the end.


bdach (approved):
seems right

Issue #4351, opened 2021-04-08
Cannot leave window if it's obstructed by always-on-top window and relative mode is enabled


Susko3:
The cursor can actually leave the window, it just takes ~2 seconds for it to realize "oh, I can actually move". The 2 seconds appear to start counting when the cursor enters the always-on-top window.

Example: Move cursor over Task Manager, wait 2 seconds..., you can now exit the window as expected. No need the ever touch the "edge".

Video:
![recording](https://user-images.githubusercontent.com/16479013/114030953-2a1ee800-987b-11eb-87b0-8cbf7cfa4a33.gif)
(Not shown in the video, but while the cursor is over Task Manager, clicking will focus Task Manager)

Examples of always on top windows:
- Task Manager (Options -> Always on top)
- Taskbar
- Taskbar previews (those that pop-up when you hover a taskbar icon)
- Notifications (from the Action Center) (when they first appear and slide in)
- Win+Space language switcher
- Volume and brightness popups (when you press the volume keys on your keyboard)
- and probably many more native Windows things, and many 3rd party programs

Also, if you Alt-Tab into the game window while the cursor is over the intersection of the always-on-top and game window, then `relativeMouseMode` and `cursorInWindow` will rapidly toggle. This will stop when the cursor enters the game window, or when you hit the edge of the game window.

PR #4552, opened 2021-06-29
Quick fix for ppy/osu#13684

yyny:
Unfortunate we could not have a conversation about this on ppy/osu.

The code in question appears to be used to stay in [Relative Mouse Mode (MSDN)](https://docs.microsoft.com/en-us/windows/uwp/gaming/relative-mouse-movement) until the user tries to move out of the window. As mentioned in the issue, on Linux this has the unintentional side effect of preventing the window manager from receiving mouse events while the cursor is inside the game window.

I'm not sure if this is an issue in Wayland, and I don't have access to Windows right now, however, the MSDN documentation seems to suggest this might be a problem on there also, so I believe a more fundamental change in the handling of `RelativeMouseMode` might be needed (Are #4351 and ppy/osu#12188 also the same issue?).

A better fix as also mentioned in the Microsoft documentation would be to only apply relative mouse mode during gameplay (and possibly also when the mouse is confined or the window is fullscreen), or to provide a key like 'Esc' to escape relative mouse mode as I suggested in the issue.

A dropdown could be added for "High precision mouse" as with "Confine mouse cursor to window" to allow the user to _always_ enable relative mouse mode, though I seriously doubt anyone cares to have higher precision mouse input in windowed mode and outside of gameplay. People who care about latency will already use fullscreen/confine mouse.

It would also be nice to have tooltips on certain settings to describe their impact in more detail, "High precision mouse" isn't very descriptive, and doesn't indicate that it grabs the mouse at all.

I can personally provide a better fix once we have a direction to take this in.

### DIFF
--- a/osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
+++ b/osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
@@ -122,8 +122,6 @@ namespace osu.Framework.Input.Handlers.Mouse
                 // to mouse control at any point.
                 window.UpdateMousePosition(position);
 
-            bool positionOutsideWindow = position.X < 0 || position.Y < 0 || position.X >= window.Size.Width || position.Y >= window.Size.Height;
-
             if (window.RelativeMouseMode)
             {
                 updateRelativeMode();
@@ -133,7 +131,7 @@ namespace osu.Framework.Input.Handlers.Mouse
 
                 // handle the case where relative / raw input is active, but the cursor may have exited the window
                 // bounds and is not intended to be confined.
-                if (!window.CursorState.HasFlagFast(CursorState.Confined) && positionOutsideWindow && !previousPositionOutsideWindow)
+                if (!window.CursorState.HasFlagFast(CursorState.Confined) && !mayStayConfined(position))
                 {
                     // setting relative mode to false will allow the window manager to take control until the next
                     // updateRelativeMode() call succeeds (likely from the cursor returning inside the window).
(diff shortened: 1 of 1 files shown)


bdach:
> "High precision mouse" isn't very descriptive, and doesn't indicate that it grabs the mouse at all

Where is this coming from, again? You keep stating this, but as far as I am aware high precision mouse/relative mouse mode does *not* grab the mouse. At most, what it does, is that it warps/recenters mouse position to keep the mouse inside the bounds of the screen. See [SDL](https://wiki.libsdl.org/SDL_SetRelativeMouseMode) [documentation](https://wiki.libsdl.org/SDL_HINT_MOUSE_RELATIVE_MODE_WARP).

Again, to restate, this time using code, grab = confine mode.

https://github.com/ppy/osu-framework/blob/e9beedbddc802c1f257a971a7ee82eeed093342f/osu.Framework/Platform/SDL2DesktopWindow.cs#L231-L232

Also as a side note this whole ordeal would probably not be anywhere near this painful if there was a video showing the behaviour you find undesirable.

peppy:
Ignoring some glaring issues with the code, this change seems completely misguided.

If there's an issue with the existing edge-of-window logic, that should be fixed. There should not be any OS-specific hard disables for functionality.

As far as I can tell this is going to *regress* behaviour.

peppy:
Closing as this completely breaks relative mouse mode and is not the correct direction.

bdach:
To expound on the above, here's video proof of that statement:

https://user-images.githubusercontent.com/20418176/123836246-c6500c80-d909-11eb-8f75-dc57ce5e91eb.mp4

Please pay attention to the log overlay in the bottom left, wherein I have added log statements to denote when `mayStayConfined` is causing the relative mouse mode to be turned off (which is on every mouse movement, and every time the mouse returns to the screen).

Even if (and that's a big if) dropping relative mouse mode was to be considered for linux it shouldn't be done in this roundabout way. It's still unclear to me what the *actual* problem is here.

gagahpangeran:
If you still wanna discuss about your issue, you can post video about your issue at https://github.com/ppy/osu/discussions as already mentioned several times at https://github.com/ppy/osu/issues/13684

PR #4553, opened 2021-06-29
Set more appropriate time limits for GH Actions workflows

bdach:
https://github.com/ppy/osu/pull/13639 but for framework.

Time limit for the main job matches appveyor. Briefly thought about going lower but /shrug. Not touching old deploy workflows either because those don't really run yet.


peppy (approved):
Could probably go lower than this it seems fine for now.

PR #4554, opened 2021-06-29
Add factory method creating `BindableWithCurrent`s depending on the value type

frenzibyte:
Allows for having a `BindableWithCurrent` or `BindableNumberWithCurrent` instance depending on the value type itself, [as proposed here](https://github.com/ppy/osu/issues/13649#issuecomment-870006577).

Initially, I was going to privatize the classes' constructors and force all usages to call `IBindableWithCurrent<T>.Create()` to avoid potentially constructing `BindableWithCurrent<float>` instead of the correct `BindableNumberWithCurrent<float>`, but since that results in severe breaking changes, I've let that as-is for now.

### DIFF
--- a/osu.Framework/Bindables/BindableNumber.cs
+++ b/osu.Framework/Bindables/BindableNumber.cs
@@ -16,9 +16,6 @@ namespace osu.Framework.Bindables
         public BindableNumber(T defaultValue = default)
             : base(defaultValue)
         {
-            // Directly comparing typeof(T) to type literal is recognized pattern of JIT and very fast.
-            // Just a pointer comparison for reference types, or constant for value types.
-            // The check will become NOP after optimization.
             if (!Validation.IsSupportedBindableNumberType<T>())
             {
                 throw new NotSupportedException(
--- a/osu.Framework/Bindables/BindableNumber.cs
+++ b/osu.Framework/Bindables/BindableNumber.cs
@@ -4,7 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.Globalization;
-using System.Runtime.CompilerServices;
+using osu.Framework.Utils;
 
 namespace osu.Framework.Bindables
 {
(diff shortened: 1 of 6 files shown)


osu.Framework/Bindables/IBindableWithCurrent.cs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using osu.Framework.Utils;
+
+namespace osu.Framework.Bindables
+{
+    public interface IBindableWithCurrent<T> : IBindable<T>
```
bdach, line 9: Any reason not to derive `IHasCurrentValue<T>` as well, therefore automatically inheriting the `Current` member without having to declare it?

frenzibyte: Yeah can do it, just copied it from there since I felt `IHasCurrentValue<T>` should be for the UI elements specifically (which provide the `BindableWithCurrent`), but honestly no reason not to just inherit it here as well and save up code lines.

bdach: It's not even entirely about the line count, it just makes sense to do semantically anyways.

osu.Framework/Utils/Validation.cs
```diff
@@ -48,5 +50,23 @@ public static bool TryParseUri(string uriString, [NotNullWhen(true)] out Uri? re
...
             return Uri.TryCreate(uriString, kind, out result);
         }
+
+        /// <summary>
+        /// Whether the specified type <typeparamref name="T"/> is a number type supported by <see cref="BindableNumber{T}"/>.
+        /// </summary>
+        /// <typeparam name="T">The type to check for.</typeparam>
+        /// <returns><see langword="true"/> if the type is supported; <see langword="false"/> otherwise.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static bool IsSupportedBindableNumberType<T>() =>
```
smoogipoo, line 60:
```suggestion
        public static bool IsSupportedBindableNumberType<T>() =>
```

osu.Framework/Bindables/BindableNumber.cs
```diff
@@ -19,7 +19,7 @@ public BindableNumber(T defaultValue = default)
             // Directly comparing typeof(T) to type literal is recognized pattern of JIT and very fast.
             // Just a pointer comparison for reference types, or constant for value types.
             // The check will become NOP after optimization.
```
smoogipoo, line 21: Probably delete this, or move it to the `Validation` method.

smoogipoo (changes requested):
Seems good otherwise

smoogipoo (approved):

PR #4557, opened 2021-07-01
Fix deadlock state if an exception is thrown during `GameHost.Run()`

peppy:
Without the check for `Idle` state, the `threadRunner` would attempt to stop threads which are not in a running state, causing a subsequent exception in the exit process, leaving the game in a `Stopping` state and waiting on the `stoppedEvent` in disposal logic.


ekrctb (approved):
I confirm the osu-side throwing `ImportTest` now time-outs instead of deadlocks when this framework is used.

smoogipoo (approved):

PR #4568, opened 2021-07-02
Fix `GetBoundCopy()` not working on "bindable with current" classes

frenzibyte:
Both missing constructors with `T defaultValue` parameter for `GetBoundCopy()` to successfully create an instance of the bindable type. Not sure if that can be enforced in some way.

Attached test case before change:
CleanShot 2021-07-02 at 23 54 50

### DIFF
--- a/osu.Framework/Bindables/IBindableWithCurrent.cs
+++ b/osu.Framework/Bindables/IBindableWithCurrent.cs
@@ -17,7 +17,7 @@ namespace osu.Framework.Bindables
         public static IBindableWithCurrent<T> Create()
         {
             if (Validation.IsSupportedBindableNumberType<T>())
-                return (IBindableWithCurrent<T>)Activator.CreateInstance(typeof(BindableNumberWithCurrent<>).MakeGenericType(typeof(T)));
+                return (IBindableWithCurrent<T>)Activator.CreateInstance(typeof(BindableNumberWithCurrent<>).MakeGenericType(typeof(T)), default(T));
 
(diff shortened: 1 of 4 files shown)


osu.Framework/Bindables/IBindableWithCurrent.cs
```diff
@@ -17,7 +17,7 @@ public interface IBindableWithCurrent<T> : IBindable<T>, IHasCurrentValue<T>
         public static IBindableWithCurrent<T> Create()
         {
             if (Validation.IsSupportedBindableNumberType<T>())
-                return (IBindableWithCurrent<T>)Activator.CreateInstance(typeof(BindableNumberWithCurrent<>).MakeGenericType(typeof(T)));
+                return (IBindableWithCurrent<T>)Activator.CreateInstance(typeof(BindableNumberWithCurrent<>).MakeGenericType(typeof(T)), default(T));
```
bdach, line 20: I assume this was added because this activator call crashes otherwise after adding the ctors to `BindableWithCurrent` and friends, correct?

frenzibyte: Yeah, because a consturctor has been added, `Activator.CreateInstance` no longer sees a parameterless constructor, and doesn't handle constructors with optional parameters, so had to add that `default(T)` in there.

bdach (approved):
seems correct

smoogipoo (approved):

PR #4569, opened 2021-07-03
Make `VertexUtils<T>` public

LeNitrous:
`VertexUtils<T>` is already documented but to reiterate, it is a helper class that keeps track of how many vertex attributes are enabled and enables and/or disables them as needed.

It will be needed for our project [vignette-project/vignette-live2d](https://github.com/vignette-project/vignette-application-live2d) which allows the framework to render Live2D models as the framework's current way, specifically making use of `VertexBatch<T>` is not suitable for our needs as we directly provide vertex and indices data to OpenGL.

image

One can argue that we can directly use `GL.EnableVertexAttribArray` directly but doing so will cause the class, or specifically `VertexUtils<T>.amountEnabledAttributes` to be in an invalid state.

Example code of how we use the class is this:
```csharp
public void Draw()
{
    if (vaoID == -1 && iboID == -1)
        Initialise();

    if (GLWrapper.BindBuffer(BufferTarget.ArrayBuffer, vaoID))
        VertexUtils<MeshVertex>.Bind();

    var vertices = new MeshVertex[Positions.Length];
    for (int i = 0; i < Positions.Length; i++)
    {
        vertices[i] = new MeshVertex
        {
            Position = Positions[i],
            TexturePosition = TexturePositions[i],
        };
    }

    GL.BufferData(BufferTarget.ArrayBuffer, (IntPtr)(Positions.Length * VertexUtils<MeshVertex>.STRIDE), vertices, BufferUsageHint.DynamicDraw);

    GLWrapper.BindBuffer(BufferTarget.ElementArrayBuffer, iboID);
    GL.DrawElements(PrimitiveType.Triangles, Indices.Length, DrawElementsType.UnsignedShort, IntPtr.Zero);
}
```

### DIFF
--- a/osu.Framework/Graphics/OpenGL/Vertices/VertexUtils.cs
+++ b/osu.Framework/Graphics/OpenGL/Vertices/VertexUtils.cs
@@ -15,7 +15,7 @@ namespace osu.Framework.Graphics.OpenGL.Vertices
     /// <summary>
     /// Helper method that provides functionality to enable and bind vertex attributes.
     /// </summary>
-    internal static class VertexUtils<T>
+    public static class VertexUtils<T>
         where T : struct, IVertex
     {
         /// <summary>


smoogipoo:
My only worry with this is `VertexUtils` is going to change as soon as VAOs are added/as soon as we adopt the opengl core profile (https://github.com/smoogipoo/osu-framework/tree/opengl-core-profile). But I suppose we can keep this around and obsolete it at that time...

smoogipoo (approved):

PR #4572, opened 2021-07-05
Remove bindable `State` and `Scheduler` requirements of `VideoDecoder`

peppy:
This was suppsoed to just be adding a test to confirm that the decoding loop is not running while a `Video` is non-present, but turned out to be a bit of a refactor. Initially I tried to write the test using the `State` bindable, but it would never change from `Running` state when the video becomes hidden, because the scheduler was not being run. Our usual workaround of overriding `IsPresent` and adding a `Scheduler.HasPendingTasks` check would just cause the decoding to start again.

It was weird that `VideoDecoder` - a non-drawable component - was requiring a scheduler at construction time just to run a bindable update on the update thread. As the `State` bindable was not being used anywhere, I have removed it, along with the scheduler requirement.

### DIFF
--- a/osu.Framework/Graphics/Video/VideoDecoder.cs
+++ b/osu.Framework/Graphics/Video/VideoDecoder.cs
@@ -13,10 +13,8 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using osu.Framework.Allocation;
-using osu.Framework.Bindables;
 using osu.Framework.Logging;
 using osu.Framework.Platform;
-using osu.Framework.Threading;
 using AGffmpeg = FFmpeg.AutoGen.ffmpeg;
 
 namespace osu.Framework.Graphics.Video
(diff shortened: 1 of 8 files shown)


osu.Framework/Graphics/Video/VideoDecoder.cs
```diff
@@ -57,27 +55,9 @@ public unsafe class VideoDecoder : IDisposable
...
 
         /// <summary>
-        /// The current state of the <see cref="VideoDecoder"/>, as a bindable.
+        /// The current state of the decoding process.
         /// </summary>
-        public IBindable<DecoderState> State => bindableState;
-
-        private readonly Bindable<DecoderState> bindableState = new Bindable<DecoderState>();
-
-        private volatile DecoderState volatileState;
```
bdach, line 66: While I'm fine with removing the bindable if it's really unused, I am not super sure about this volatile backing field removal. The state *is* changed from multiple threads (both update and a decoding thread), which I have confirmed experimentally in the debugger. And I *think* auto-property backing fields are non-volatile by default ([see sharplab](https://sharplab.io/#v2:C4LglgNgPgAgTABgLACh4EZWpugnACgCJAeDcBB9wgSgG4s0BmAAngYGEGBvVB7pxnBBgDEwAJwDOwDgwDmAU2BUGABxFgAbgENgshmPmKAvlx7HuK9Vp1qA9hC2Qd/XbIDG1gHYATGih68m6AIAyq4enhymfjLyDAC8AHzObl4+UTx6kglJYXEMmhAArrKpPEYoBkA=)).

That said, the decoder state *is* printed every frame in the test scene, and I couldn't see it break on a quick look after this in either debug or release, so...

peppy: I can't see how the missing volatile flag would ever cause this to break, so until it does break I'm not concerned.

smoogipoo (approved):

PR #4579, opened 2021-07-07
Obsolete ability for slider bar to handle keyboard input when not hovered

peppy:
This was a weird overload added for a very specific use case in osu!, which has since been deemed to add very hidden input handling functionality that we'd rather avoid.

It doesn't make sense for a slider bar to handle input when it's not hovered under any circumstances.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneSliderBar.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneSliderBar.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using NUnit.Framework;
 using osu.Framework.Bindables;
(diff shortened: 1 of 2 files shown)


smoogipoo (dismissed):

osu.Framework.Tests/Visual/UserInterface/TestSceneSliderBar.cs
```diff
@@ -220,6 +221,7 @@ public class TestSliderBar : BasicSliderBar<double>
         {
             public bool KeyboardInput;
 
+            [Obsolete("Implement this kind of behaviour separately instead.")]
```
frenzibyte, line 224:
```suggestion
            [Obsolete("Implement this kind of behaviour separately instead.")] // Can be removed 20220107
```

frenzibyte: ah nevermind this is an override

frenzibyte (approved):

smoogipoo (approved):

Issue #3974, opened 2020-11-02
Interpolation functions should be documented/refactored to better define behaviour beyond extents


peppy:
As noticed in https://github.com/ppy/osu/pull/10659, it is quite easy to assume (incorrectly) that interpolation functions clamp at their `start`/`endTime` extents (give the names of parameters it sounds like clamping may be enforced).

I'd propose that, unless we use these functions in a way which requires it somewhere, clamping is enforced. If that's not agreed upon, at very least documenting this behaviour in xmldoc may help avoid future similar cases.

bdach:
As a data point, this is already done locally in many places in the framework:

https://github.com/ppy/osu-framework/blob/c765605199c2cb23223cb38d5d809781337d02c8/osu.Framework/Graphics/Transforms/TransformBindable.cs#L24-L30

https://github.com/ppy/osu-framework/blob/c765605199c2cb23223cb38d5d809781337d02c8/osu.Framework/Graphics/Transforms/TransformCustom.cs#L171-L177

https://github.com/ppy/osu-framework/blob/c765605199c2cb23223cb38d5d809781337d02c8/osu.Framework/Graphics/TransformableExtensions.cs#L766-L772

On the other side I can see how not clamping can be sometimes useful, but at that point it's not really *inter*polation anymore rather than *extra*polation. And it doesn't really work for some easing types, either (something like `OutElastic` will not extrapolate well, I imagine).

As for my opinion I would probably make it a default-but-opt-out behaviour by having a `clamp` param defaulting to `true`. That change might hurt though, as it's at the level of "too many usages to actually check if every single one behaves correctly".

PR #4582, opened 2021-07-08
Implement `Colour4` interpolation function

### LINKED ISSUE #4581: Interpolation throws a `TypeInitializationException` when using with `Colour4`
I'm trying to do a little animation on a `SpriteText` in the `Update` method, what I want is to make the text color change every 500ms, from golden to lighter golden.

I saw in the documentation, in the topic [Transformations and Tweening](https://github.com/ppy/osu-framework/wiki/Transforms-and-Tweening), that in order for the transformation to occur at each frame, it is recommended to use interpolation.

Here is the code:
```cs
{
    textLogo2.Colour = Interpolation.ValueAt(Clock.ElapsedFrameTime, Colour4.Gold, Colour4.Gold.Lighten(1), 0, 200, Easing.OutQuint);

    base.Update();
}
```

The result is an Exception:
```
[runtime] 2021-07-08 00:01:02 [error]: An unhandled error has occurred.
[runtime] 2021-07-08 00:01:02 [error]: System.TypeInitializationException: The type initializer for 'GenericInterpolation`2' threw an exception.
[runtime] 2021-07-08 00:01:02 [error]: ---> System.NotSupportedException: Type osu.Framework.Graphics.Colour4 has no interpolation function. Implement the interface osu.Framework.Utils.IInterpolable`1[osu.Framework.Graphics.Colour4] interface on the object.
[runtime] 2021-07-08 00:01:02 [error]: at osu.Framework.Utils.Interpolation.GenericInterpolation`2..cctor()
[runtime] 2021-07-08 00:01:02 [error]: --- End of inner exception stack trace ---
[runtime] 2021-07-08 00:01:02 [error]: at osu.Framework.Utils.Interpolation.ValueAt[TValue,TEasing](Double time, TValue startValue, TValue endValue, Double startTime, Double endTime, TEasing& easing)
[runtime] 2021-07-08 00:01:02 [error]: at osu.Framework.Utils.Interpolation.ValueAt[TValue](Double time, TValue startValue, TValue endValue, Double startTime, Double endTime, Easing easing)
```

devclyde:
Closes #4581

I implemented the `Colour4` interpolation function, this does the same thing as `Color4`

### DIFF
--- a/osu.Framework/Utils/Interpolation.cs
+++ b/osu.Framework/Utils/Interpolation.cs
@@ -242,6 +242,26 @@ namespace osu.Framework.Utils
                     startColour.A + t * (endColour.A - startColour.A));
             }
 
+            public static Colour4 ValueAt(double time, Colour4 startColour, Colour4 endColour, double startTime, double endTime, in TEasing easing)
+            {
+                if (startColour == endColour)
+                    return startColour;
+
+                double current = time - startTime;
+                double duration = endTime - startTime;
+
+                if (duration == 0 || current == 0)
+                    return startColour;
+
+                float t = Math.Max(0, Math.Min(1, (float)easing.ApplyEasing(current / duration)));
+
+                return new Colour4(
+                    startColour.R + t * (endColour.R - startColour.R),
+                    startColour.G + t * (endColour.G - startColour.G),
+                    startColour.B + t * (endColour.B - startColour.B),
+                    startColour.A + t * (endColour.A - startColour.A));
+            }
+
             public static byte ValueAt(double time, byte val1, byte val2, double startTime, double endTime, in TEasing easing)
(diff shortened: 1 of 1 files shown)


frenzibyte (approved):
Seems fine, I would've expected `Colour4` to implicitly convert to `Color4` but due to the [generic variant of this](https://github.com/ppy/osu-framework/blob/d44d62688fb9e6cabb0e036c2f938657a926e3eb/osu.Framework/Utils/Interpolation.cs#L180-L181) existing, it takes priority over implicit conversion.

bdach:
I guess I don't see huge issues with this aside from the fact that contrary to everything else, `ValueAt` implementations for `Colo(u)?r4` clamp to the interpolation range rather than everything else which accepts arbitrary `t`, but that's probably something to deal with/unify in https://github.com/ppy/osu-framework/issues/3974...

bdach (approved):

PR #4586, opened 2021-07-09
Add Touchpad support for Windows 8+ and MacOS

SleepySwords:
### Considerations
- **This has only gone through Limited testing**
    - This has only be tested on a MacBook Air (MacOS Big Sur) 
    - This also has been tested on a Macbook Air with Bootcamp and high precision drivers (Windows 10)
    - So no 32 bit testing, other touchpads or if Windows 7 would crash/error

- [ ] There is currently no Linux support
    - Not sure if this would be the right way to implement this, but if someone wants to write an impl they are free to:  https://osu.ppy.sh/community/forums/topics/1238167

- [ ] Windows 7 doesn't support High Precision Touchpads
    - Should there be support to something like the Synaptics API?

- [ ] Clicking may become harder for some laptops
    - should absolute trackpad even be enabled in osu! menus?
    - should there be a keyboard bind to a click?

- [ ] Should there be multitouch support?
    - Is there a way to render multiple cursors currently?

- [ ] On Windows, an issue is that the touchpad thinks that there has been a release and then two new fingers, which muddles the TouchSource, I'm not sure if this issue is exclusive to my laptop, or the way it's implemented

- [ ] Should there be touchpad detection?
    - One implementation would be waiting for the first touch on the touchpad
    - Or some Native Implementation
        - For MacOS: https://eternalstorms.wordpress.com/2015/11/16/how-to-detect-force-touch-capable-devices-on-the-mac/ but applied to general touchpads
        - For Windows: https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getrawinputdevicelist


osu.Framework/Platform/Windows/Native/HidpUtils.cs
```diff
@@ -0,0 +1,252 @@
...
+            if (Input.GetRawInputDeviceInfoW(device, 0x20000005, preparsedData, ref payloadSize) == -1)
+                Input.ThrowLastError("Unable to get Raw Input Data");
+            return preparsedData;
+        }
+
+        private static unsafe IEnumerable<HidpValueCaps> getValueCaps(byte[] preparsedData)
+        {
+            var status = Input.HidP_GetCaps(preparsedData, out var caps);
+            if (status != NSStatus.HIDP_STATUS_SUCCESS)
+                throw new NativeException($"Error while getting Value Caps: {status}");
```
SleepySwords, line 30: Is it alright to throw an exception like this? Using the logger didn't produce a stacktrace.

osu.Framework/Platform/MacOS/MacOSTouchpadHandler.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        public override bool Initialize(GameHost host)
+        {
+            if (!base.Initialize(host))
+                return false;
+
+            if (!(host.Window is SDL2DesktopWindow))
+                return false;
+
+            var viewClass = Class.Get("SDLView");
```
SleepySwords, line 40: Not sure if it was the best idea to have this code here, or in the MacOSWindow

osu.Framework.Tests/Visual/Input/TestSceneTouchpadInput.cs
```diff
@@ -0,0 +1,208 @@
...
+            base.Dispose(isDisposing);
+
+            if (!isDisposing) return;
+
+            // Trying to ensure that this InputHandler isn't still enabled when browsing other tests.
+            var touchpadHandler = host.AvailableInputHandlers.OfType<TouchpadHandler>().FirstOrDefault();
+
+            if (touchpadHandler != null)
+            {
+                touchpadHandler.Enabled.Value = false;
```
SleepySwords, line 46: Is this really the only way to disable things when the Test isn't focused?

osu.Framework/Platform/MacOS/Native/Cocoa.cs
```diff
@@ -65,6 +69,40 @@ internal static class Cocoa
 
         public static float SendFloat(IntPtr receiver, IntPtr selector) => IntPtr.Size == 4 ? SendFloat_i386(receiver, selector) : (float)SendFloat_x64(receiver, selector);
 
+        [DllImport(LIB_OBJ_C, EntryPoint = "objc_msgSend_fpret")]
+        private static extern Vector2 SendVector2d_i386(IntPtr receiver, IntPtr selector);
```
SleepySwords, line 73: This is just an assumption, no idea if this would work on a 32-bit Mac

osu.Framework/Platform/Windows/Native/HidpUtils.cs
peppy: As long as every usage of this is guarded against so it doesn't crash the host.

osu.Framework/Platform/MacOS/Native/Cocoa.cs
peppy: 32 bit macs do not exist.

osu.Framework.Tests/Visual/Input/TestSceneTouchpadInput.cs
peppy: I think this is quite dangerous as the disposal could be run async. Best just leaving it in a toggle step like other similar tests (see raw input as an example).

peppy:
This looks like potentially the wrong direction to implement this. I think it would be best done in `OpenTabletHandler` and consumed via the same means. Otherwise we are adding yet another interface and required method of configuration for all framework consumers.

Is there any reason you can see which would make this unfeasible?

SleepySwords:
I'm not sure what you mean by that? Is there a way to make OpenTabletDriver support Touchpads via some sort of API or does OpenTabletDriver already have touchpad support?

peppy:
Would need some investigation, but if the goal is to add *absolute* input support for touchpads, that seems like the potential place to implement it (since the whole flow there is geared to such a pattern).

osu.Framework/Input/Handlers/Touchpad/TouchpadHandler.cs
```diff
@@ -0,0 +1,76 @@
...
+    public class TouchpadHandler : InputHandler
+    {
+        public override bool IsActive => false;
+
+        public override string Description => "Touchpad";
+
+        public Bindable<Vector2> AreaOffset { get; } = new Bindable<Vector2>(new Vector2(0.5f, 0.5f));
+
+        // Usually an AreaSize of (1, 1) leads to being unable to inability to reach the edges.
+        public Bindable<Vector2> AreaSize { get; } = new Bindable<Vector2>(new Vector2(0.9f, 0.9f));
```
peppy, line 24: In case it wasn't clear, this is the area of concern and why I am suggesting implementing as part of OTD. We would need to create several levels of abstraction to make this configurable via the same in-game options as the tablet. Or it would need to use the same control but via a common interface.

I'm not sure this is the correct way to go forward for such a niche feature. At the point a touchpad is being used as an absolute input device, it's basically a "tablet", so if it is possible to handle it using the same inteface rather than creating more local implementation I think that should be preferred.

SleepySwords: Ah I see, that makes more sense. My only concern in that case would be dependent on if the Touchpads have some sort of common ground (either in the Vendor ID, Device String or some other thing), or if you'd have to add a config per trackpad in OpenTabletDriver.

peppy: I believe they were looking to refactor things such that it wasn't depending on vendor ID alone, so may be worth looking into still.

SleepySwords: Looking more into it, it's def possible if the OpenTabletDriver refactor involves being able to choose the ReportParser by a UsagePage and Usage.

Both MacOS and Windows have UsagePage/Usage
https://developer.apple.com/documentation/hiddriverkit/handling_stylus_input_from_a_human_interface_device
https://docs.microsoft.com/en-us/windows-hardware/design/component-guidelines/windows-precision-touchpad-sample-report-descriptors

I'm not entirely sure if MacOS trackpad positions and transmitted via HID, it should be though.

peppy: @InfinityGhost Can i get your quick opinion on this one? I think you'd have a better idea if you think this is best suited as something added to OTD or our end.

InfinityGhost: Absolutely could be done on OpenTabletDriver's end. We've planned on adding detection via usage page and usage for a while but it wasn't implemented in our USB library.

On the note of configs per trackpad, not all trackpads have the same dimensions and not all of them report digitizer dimensions correctly in their HID descriptor, so may still be worth doing it per touchpad.

We've been slowly adding support for miscellaneous tablet tools/features such as touch, tablet mice, wheels, etc. so adding touchpad support into OpenTabletDriver wouldn't be far fetched either.

SleepySwords: Just a heads up, I don't think that MacOS reports trackpads events via HID (when searching for HIDs I did find usages of a touchpad, but the reports were just a relative generic mouse. Perhaps I'm just not looking in the right places).

The way it currently works in this PR is that it swizzles the methods in the SDLView. https://developer.apple.com/documentation/appkit/nsgesturerecognizer/2544863-touchesmovedwithevent?language=objc. Though since OpenTabletDriver doesn't have a view, you may need to use macOS private multitouch framework.

PR #4588, opened 2021-07-10
Throw when attempting to set container to `Children` and similar cases

frenzibyte:
Gets rid of the usual mistake of setting a `Container` to `Children`, or passing a `Container` to `AddRange`, which causes weird behaviour such as "May not add a drawable to multiple containers" without any trace of being related to this.

Tests seem to pass, double-checked and removed any usages to this quirk.

### DIFF
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -567,6 +567,12 @@ namespace osu.Framework.Graphics.Containers
         /// </summary>
         protected internal void AddRangeInternal(IEnumerable<Drawable> range)
         {
+            if (range is IContainerEnumerable<Drawable> container)
+            {
+                throw new InvalidOperationException($"Attempting to add a {container} as a range of children to {this}."
+                                                    + $"If intentional, consider using the {nameof(container.Children)} property instead.");
+            }
+
             foreach (Drawable d in range)
                 AddInternal(d);
         }
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -567,10 +567,10 @@ namespace osu.Framework.Graphics.Containers
         /// </summary>
         protected internal void AddRangeInternal(IEnumerable<Drawable> range)
         {
-            if (range is IContainerEnumerable<Drawable> container)
+            if (range is IContainerEnumerable<Drawable>)
             {
-                throw new InvalidOperationException($"Attempting to add a {container} as a range of children to {this}."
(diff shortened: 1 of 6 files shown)


osu.Framework/Graphics/UserInterface/Menu.cs
```diff
@@ -55,7 +55,7 @@ public abstract class Menu : CompositeDrawable, IStateful<MenuState>
         /// <summary>
         /// Gets the item representations contained by this <see cref="Menu"/>.
         /// </summary>
-        protected internal IReadOnlyList<DrawableMenuItem> Children => ItemsContainer;
+        protected internal IReadOnlyList<DrawableMenuItem> Children => ItemsContainer.Children;
```
bdach, line 58: What do these changes (here and in `MenuTestScene`) have to do with this pull? It's a get-only property...? Why would that make any diifference?

frenzibyte: Eh, not much point rather than just not faking out a container as a list, don't think it'll actually be used so can revert.

osu.Framework/Graphics/Containers/CompositeDrawable.cs
```diff
@@ -567,6 +567,12 @@ protected internal virtual void AddInternal(Drawable drawable)
         /// </summary>
         protected internal void AddRangeInternal(IEnumerable<Drawable> range)
         {
+            if (range is IContainerEnumerable<Drawable> container)
```
peppy, line 570:
```suggestion
            if (range is IContainerEnumerable<Drawable>)
```

No need to grab the variable for these calls.

frenzibyte: Would change to:
```csharp
            if (range is IContainerEnumerable<Drawable>)
            {
                throw new InvalidOperationException($"Attempting to add a {typeof(IContainer)} as a range of children to {this}."
                                                    + $"If intentional, consider using the {nameof(IContainerEnumerable<Drawable>.Children)} property instead.");
            }
```

is that fine?

osu.Framework/Graphics/UserInterface/Menu.cs
bdach: I mean if it was undesirable to be able to iterate containers to get their children then they shouldn't be implementing the enumerable interface in the first place...

frenzibyte: That would be my obvious first thought, but it seems it's heavily used to shorten `Container.Children.Linq()` to `Container.Linq()`, so just throwing on this specific case and leaving it working for now is the best choice.

If it breaks again somewhere else for some odd reason, then we can reassess the whole `IEnumerable` thing on containers.

smoogipoo (approved):
I think this is okay

PR #4590, opened 2021-07-10
Improve usability of popover & popover container

bdach:
Some assorted fixes and improvements which resulted from @peppy attempting to use the control for multiplayer room password input. I recommend reading commit-by-commit; I didn't want to split this into 3 or 4 pulls because noise.

* 58863c6 removes the weird `OnMouse{Down,Up}` handling that seemed to be unnecessary in the end. Shaves a fair bit of complexity off as well.
* 7fed4ba fixes not being able to use the popover container in hierarchies that had auto-size axes on them. Fix mirrors what `ContextMenuContainer` and `TooltipContainer` already do.
* 26cde79 is intended to guide towards the proper usage of popovers (which is putting what should go inside the popover into `Children` rather than `InternalChildren`).
* f32bebb adds support for sizing the popovers directly rather than them breaking when sizes are set on the popover itself, which leads to having to set the desired size on their contents instead.

### DIFF
--- a/osu.Framework/Graphics/Cursor/PopoverContainer.cs
+++ b/osu.Framework/Graphics/Cursor/PopoverContainer.cs
@@ -11,7 +11,6 @@ using osu.Framework.Graphics.UserInterface;
 using osu.Framework.Input.Events;
 using osu.Framework.Utils;
 using osuTK;
-using osuTK.Input;
 
 #nullable enable
 
@@ -44,9 +43,6 @@ namespace osu.Framework.Graphics.Cursor
 
         protected override bool OnClick(ClickEvent e)
         {
-            if (e.Button != MouseButton.Left)
-                return false;
-
             target = FindTargets().FirstOrDefault();
             if (target == null)
                 return false;
--- a/osu.Framework/Graphics/Cursor/PopoverContainer.cs
+++ b/osu.Framework/Graphics/Cursor/PopoverContainer.cs
@@ -42,34 +42,25 @@ namespace osu.Framework.Graphics.Cursor
             };
         }
 
-        protected override bool OnMouseDown(MouseDownEvent e)
+        protected override bool OnClick(ClickEvent e)
         {
-            switch (e.Button)
-            {
-                case MouseButton.Left:
-                    target = FindTargets().FirstOrDefault();
-                    break;
-            }
+            if (e.Button != MouseButton.Left)
+                return false;
 
-            return false;
-        }
-
-        protected override void OnMouseUp(MouseUpEvent e)
-        {
-            base.OnMouseUp(e);
-
+            target = FindTargets().FirstOrDefault();
             if (target == null)
-                return;
+                return false;
 
             currentPopover?.Hide();
 
             var newPopover = target.GetPopover();
             if (newPopover == null)
-                return;
+                return false;
 
             popoverContainer.Add(currentPopover = newPopover);
             currentPopover.Show();
             currentPopover.State.BindValueChanged(_ => cleanUpPopover(currentPopover));
+            return true;
         }
 
         private void cleanUpPopover(Popover popover)
--- a/osu.Framework/Graphics/Cursor/PopoverContainer.cs
+++ b/osu.Framework/Graphics/Cursor/PopoverContainer.cs
@@ -80,6 +80,19 @@ namespace osu.Framework.Graphics.Cursor
             updatePopoverPositioning();
         }
 
+        protected override void OnSizingChanged()
+        {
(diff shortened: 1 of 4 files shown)


osu.Framework/Graphics/Cursor/PopoverContainer.cs
```diff
@@ -42,34 +42,25 @@ public PopoverContainer()
...
-            }
-
-            return false;
-        }
-
-        protected override void OnMouseUp(MouseUpEvent e)
-        {
-            base.OnMouseUp(e);
+            if (e.Button != MouseButton.Left)
+                return false;
```
frenzibyte, line 48: I think it makes sense to make any button which is considered a "click" by the button event manager to activate the popup? (+ can't find any existing override that does it like this, but might have failed)

osu.Framework/Graphics/UserInterface/Popover.cs
```diff
@@ -138,6 +140,50 @@ private float getRotationFor(Anchor anchor)
...
+
+        public new Vector2 Size
+        {
+            get => Body.DrawSize;
+            set
+            {
+                Width = value.X;
+                Height = value.Y;
+            }
+        }
```
frenzibyte, line 183: Is it intentional to return `Draw{Width/Height/Size}` rather than `{Width/Height/Size}`? can't really figure out why.

frenzibyte (changes requested):

osu.Framework/Graphics/Cursor/PopoverContainer.cs
bdach: I think left click should be the primary mode of activation for popovers. I guess I could make this an extension point via a property or interface and users could choose which buttons they want, but I'm generally hesitant because it can clash with context menus, for one example.

osu.Framework/Graphics/UserInterface/Popover.cs
bdach: It was intentional, for when autosize is active on the body.

osu.Framework/Graphics/Cursor/PopoverContainer.cs
bdach: Also I guess right mouse clicks don't even fire `OnClick` right now so I might as well get rid of this guard and call it a day.

frenzibyte: I've also raised this because other UI controls don't look to early-return on non-left buttons in `OnClick` events, so popover would be kind of inconsistent there.

osu.Framework/Graphics/UserInterface/Popover.cs
frenzibyte: Delegating `Size` to an internal drawable's `DrawSize` rather than its `Size` sounds slightly arbitrary if you ask me. Are there gonna be future usages for this? otherwise I think it makes more sense to delegate to `Size` instead?

frenzibyte (dismissed):

bdach:
I've piled one more fix here - as it turns out, when trying to use `PopOut` game-side it turned out to be entirely broken, because `PopoverContainer` (due to order of registering for the `State` value change event) was always expiring the popover before any transforms from `PopOut` could even be added, and therefore potentially interrupting the animation.

Fixed by only expiring the popover when it's being replaced by another one (or nothing). I also considered hacking it together to make the order correct (moving the `BindValueChanged` call to `OnLoadComplete`) but I hated it because nested callbacks are bad.

frenzibyte (approved):

smoogipoo:
I don't know how I feel about the sizing stuff - it looks flimsy. Like if you wanted to put a text flow container in the content you just... may not? Unless you duplicate the size definition into an internal content container. It seems like you're hacking things together either way. Have you considered making it relative size and enforcing the _consumer_ to decide a fixed size or set the content to autosize instead?

smoogipoo:
Side note - can `private Container content` be made `readonly` in `Popover`? I don't see why it couldn't and am confused why that's not showing a CI inspection.

peppy:
Agree with sizing being flaky. I think it would be better long term to refactor the container to not rely on the `0x0` size to perform layout.

peppy (approved):
Going to go with this for now and see how things perform.

bdach:
I think there is a plausible path where the 0x0 sizing does not exist, but it would add significant complexity in order to preserve existing behaviour (would probably need to manually handle masking and use relative anchor).

I don't necessarily follow how relative axes would work here either, but I haven't yet tried to apply them, so I may be simply missing something.

PR #4591, opened 2021-07-10
Allow drawables to manually show and hide popovers

bdach:
- [x] Depends on #4590 (for mergeability).

Another thing that fell out of @peppy's attempt of integrating the popover into the multi screen is that sometimes the "automagic" way of showing the popover, via handling the mouse events and target finding, doesn't work for various reasons in some scenarios (for instance, if a click never reaches the popover, like when the popover target is inside a scroll, or it might be desired to show a popover as a result of another user interaction, such as focusing a text box).

This pull adds a method to manually drive the nearest `PopoverContainer` via extension methods that traverse the drawable hierarchy. The test scene shows what can be done with this (showing a popover when a textbox is focused).

A side thing that also happens in this pull to allow the above test scene to happen, is relaxing the constraint on the body of the popover (it no longer needs to be a `FocusedOverlayContainer` and take keyboard focus). This opens the popover to yet more use cases that it might not have been applicable to without it.

### DIFF
--- a/osu.Framework/Extensions/PopoverExtensions.cs
+++ b/osu.Framework/Extensions/PopoverExtensions.cs
@@ -1,0 +1,39 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Cursor;
+
+#nullable enable
+
+namespace osu.Framework.Extensions
+{
+    public static class PopoverExtensions
+    {
+        /// <summary>
+        /// Shows the popover for <paramref name="hasPopover"/> on its nearest <see cref="PopoverContainer"/> ancestor.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


osu.Framework/Extensions/PopoverExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                if (node is PopoverContainer popoverContainer)
+                {
+                    popoverContainer.SetTarget(target);
+                    return;
+                }
+
+                node = node.Parent;
+            }
+        }
```
frenzibyte, line 37: Nit: can benefit from the internal `FindClosestParent` method.

```suggestion
        private static void setTargetOnNearestPopover(Drawable origin, IHasPopover? target) => origin.FindClosestParent<PopoverContainer>()?.SetTarget(target);
```

frenzibyte (commented):
Looks pretty correct otherwise

osu.Framework/Extensions/PopoverExtensions.cs
bdach: fair enough

frenzibyte (approved):

peppy (approved):
Code seems sound enough, and works for my usage.

PR #4593, opened 2021-07-11
Reset a StepSlider to its default value when right-clicked

HoLLy-HaCKeR:
Allows the user to right-click a SliderStep in the test browser to reset it to its default value.

[Streamable](https://streamable.com/xihrgd)

### DIFF
--- a/osu.Framework/Testing/Drawables/Steps/StepSlider.cs
+++ b/osu.Framework/Testing/Drawables/Steps/StepSlider.cs
@@ -9,7 +9,6 @@ using osu.Framework.Graphics.UserInterface;
 using System;
 using osu.Framework.Bindables;
 using osu.Framework.Extensions.Color4Extensions;
-using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Input.Events;
 using osuTK.Input;
 
@@ -68,7 +67,7 @@ namespace osu.Framework.Testing.Drawables.Steps
 
         protected override bool OnMouseDown(MouseDownEvent e)
         {
-            if (e.Button.HasFlagFast(MouseButton.Right))
+            if (e.Button == MouseButton.Right)
             {
                 Current.SetDefault();
                 Flash();
--- a/osu.Framework/Testing/Drawables/Steps/StepSlider.cs
+++ b/osu.Framework/Testing/Drawables/Steps/StepSlider.cs
@@ -67,6 +67,7 @@ namespace osu.Framework.Testing.Drawables.Steps
 
         protected override bool OnMouseDown(MouseDownEvent e)
         {
+            // Reset value via right click. This shouldn't happen if a drag (via left button) is in progress.
             if (!IsDragged && e.Button == MouseButton.Right)
             {
                 Current.SetDefault();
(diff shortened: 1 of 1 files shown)


osu.Framework/Testing/Drawables/Steps/StepSlider.cs
```diff
@@ -64,19 +66,21 @@ public StepSlider(string description, T min, T max, T start)
             currentNumber.SetDefault();
         }
 
-        protected override void OnDragEnd(DragEndEvent e)
+        protected override bool OnMouseDown(MouseDownEvent e)
         {
-            var flash = new Box
+            if (e.Button.HasFlagFast(MouseButton.Right))
```
bdach, line 71: `MouseButton` isn't a flag enum, so using `HasFlagFast` in this context is actively misleading.

```suggestion
            if (e.Button == MouseButton.Right)
```

This should also probably not fire when a drag is in progress, because it doesn't really work that well in that case.

bdach (changes requested):

HoLLy-HaCKeR:
Applied feedback. I'm willing to change the right-click to something else if it's better UX-wise, it just what seemed most natural to me.

bdach (dismissed):
Okay with this. As was said on discord, right click seems a bit of a weird interaction to choose for reset but given it's a test scene usage I'm not really willing to belabor that further.

osu.Framework/Testing/Drawables/Steps/StepSlider.cs
```diff
@@ -64,19 +65,21 @@ public StepSlider(string description, T min, T max, T start)
             currentNumber.SetDefault();
         }
 
-        protected override void OnDragEnd(DragEndEvent e)
+        protected override bool OnMouseDown(MouseDownEvent e)
         {
-            var flash = new Box
+            if (!IsDragged && e.Button == MouseButton.Right)
```
peppy, line 70: This `IsDragged` check isn't going to be doing anything. A drawable doesn't become dragged until *after* the initial mousedown.

peppy (changes requested):

osu.Framework/Testing/Drawables/Steps/StepSlider.cs
bdach: I'm pretty sure I tested and this worked as expected...

Seems to check out against code, too. Drag handling seems to actually run on position changes (see mouse button event manager's `HandlePositionChange`)...?

peppy: How did you test? I can't imagine a scenario where `IsDragged` becomes `true` before `OnMouseDown`. It would require two mouse events to be handled at once, before the mouse down, with a default movement of 10 screen pixels.

bdach: I did the obvious thing - start dragging with left mouse, press right mouse, check that the reset didn't fire.

I suppose that does make two mouse down events. Unsure in which scenario this could fail (is dragging with right mouse even a thing?)

peppy: Aha, dragging is a shared state in this case. Will add an inline comment mentioning this, as I obviously found it a bit confusing as-is.

peppy (approved):

PR #4594, opened 2021-07-12
Remove extremely unsafe Bindable.Unbind() method

smoogipoo:
This is the most unsafe thing I've seen all year. Potentially leaving one side of the bindables completely bound.

Seems to only be used in this one method.


peppy:
Just to clarify, the unsafe part is that `Unbind` exists at all and there's no logic change here, correct?

peppy (approved):

smoogipoo:
Correct.

PR #4596, opened 2021-07-12
Avoid creating a second popover when one is already displayed for the current target

peppy:
As dicussed.

- [x] Depends on #4595 to avoid merge conflicts in tests

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestScenePopoverContainer.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestScenePopoverContainer.cs
@@ -104,6 +104,34 @@ namespace osu.Framework.Tests.Visual.UserInterface
         }
 
         [Test]
+        public void TestUseExistingWhenSameProviderClicked()
+        {
+            Popover first = null;
+
+            createContent(button => new BasicPopover
+            {
+                Child = new SpriteText
+                {
+                    Text = $"{button.Anchor} popover"
+                }
+            });
+
+            AddStep("click button", () =>
+            {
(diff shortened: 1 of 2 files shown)


bdach:
There was a possible issue with the guard, namely that `SetTarget()` could return `true` when going from null target to null target, and because the return value of that is used directly in `OnClick()`, that would lead to clicks being unnecessarily handled by the popover container.

I've fixed that; the rest (sans dependency concerns) looks good :+1:

bdach (approved):

peppy:
> There was a possible issue with the guard, namely that `SetTarget()` could return `true` when going from null target to null target, and because the return value of that is used directly in `OnClick()`, that would lead to clicks being unnecessarily handled by the popover container.
> 
> I've fixed that; the rest (sans dependency concerns) looks good 👍

Oh, that was actually intentional (I changed it to ensure this was the case). Feels weird that a popover container should let input fall through in some but not all cases. Up for discussion I guess. Probably need a use case to work with.

bdach:
If it was intentional, then `OnClick` should probably just hard-return `true`. Because the popover -> null case has `SetTarget` return false, it felt inconsistent that it wasn't blocking input but the null -> null case did.

peppy:
I think what you've changed it to is fine until we come up with a case where it doesn't meet expectations. As you say, it's not consistent at least.

peppy:
Actual test failure here, I'll investigate.

peppy:
As discussed (in discord) this behaviour is no longer feasible as the default `Popover` handling dismisses a previous popover on mouse down, rather than on click. Will close this for now - we can revisit if it's ever required.

PR #4599, opened 2021-07-12
Add ability to `SpriteText` to change displayed text case

Game4all:
osu-web uses in a few places the [`text-transform`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform) CSS property which allows overriding the case in which text is displayed. This is replicated osu!-side by calling `.ToUpper()` / `.ToLower()` on strings. However, with localisation, this becomes a bit more difficult to do so and I don't think adding yet another type of localisable string is the way to go hence this PR.

# Introducing the `TextTransform` property
This adds the ability to `SpriteText` to change the case in which the text is displayed through the `TextTransform` property. The used text transform will use the effective culture of the localisation manager for applying case transforms. I'm actually hesistant about the property name since it may create confusion with the `Transforms` property (maybe `Casing` fits better ?) so I'm open to suggestions for naming.

### DIFF
--- a/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteTextTextTransforms.cs
+++ b/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteTextTextTransforms.cs
@@ -1,0 +1,75 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Sprites;
+using osu.Framework.Graphics.UserInterface;
+using osu.Framework.Testing;
+
+namespace osu.Framework.Tests.Visual.Sprites
+{
+    public class TestSceneSpriteTextTextTransforms : TestScene
+    {
+        private readonly BasicDropdown<TextTransform> textTransformDropdown;
+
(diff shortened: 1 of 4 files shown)


peppy:
I haven't gone through this in detail, but one thing to consider is that this could potentially be implemented as part of the `LocalisableString` class hierarchy. I guess before I get into reviewing this, is there any reason you chose to implement at `SpriteText` specifically? You potentially touch on this in the opening post so if you could describe the perceived difficulties that would be helpful.

bdach:
Hard agree with the above. This should be supported at the localisable string level to avoid reimplementation in every place that will want to operate on the raw data (and there will be such places - text flow container probably will be one, to split the text across word boundaries).

Game4all:
I think that the main decision for adding this at a `SpriteText`-level is mainly after seing that text is uppercased / lowercased mainly for visual style and is applied mainly on `SpriteText`. 

I touched a very few words in the OP, saying that applying the procedure used previously but with localisation was a bit more difficult, but I guess that actually the word is a "bit more ugly": from an earlier attempt at adding case transformation to `LocalisableString`, trying to uppercase a formatted string would look like `((LocalisableString)myFormattedLocalisableString).ToUpper()` .

Also @bdach 's point on support on the need to reimplement display text transformation for virtually every UI primitive that is localisable nails that this approach is partly to not say completely wrong.

Game4all:
I'll be closing this as the direction is quite wrong imo and will explore a second time but further this time, storing case information along in the `LocalisableString`

PR #4602, opened 2021-07-13
Add default value ctor to BindableLong

miterosan:
This makes the ctor of BindableNumber accessible


frenzibyte (approved):
Not sure whether this should remain existing at all, but the constructor should be added.

peppy:
@miterosan are you using this in a project? if so i'm interested to know the use case.

miterosan:
Currently writing a game with osu!framework. Working on it now and then. It is not public anywhere, because it did not make the MVP stage yet.

I used BindableLong for storing the health of players.

PR #4603, opened 2021-07-13
Add ability to change case of a `LocalisableString`

Game4all:
Here is an initial implementation of case transformations for  `LocalisableString` as per the direction that was proposed in discussion of #4599. `LocalisableString` stores along its string data the case to apply on the final computed string in the `LocalisedBindableString`. 

Note that the case is intentionally set to `Default` for strings constructed from a plain string to allow casing the plain string using the BCL methods. Also, methods to change case of `LocalisableString` are implemented as extensions methods on `ILocalisableStringData` to construct a `LocalisableString` wrapping the data with the correct case instead of on the `LocalisableString`, since it's immutable anyway. This results in a syntax similar to string's one when dealing with `TranslatableString`.

### DIFF
--- a/osu.Framework.Tests/Localisation/LocalisationTest.cs
+++ b/osu.Framework.Tests/Localisation/LocalisationTest.cs
@@ -244,7 +244,7 @@ namespace osu.Framework.Tests.Localisation
         }
 
         [Test]
-        public void TestLocalisableStringCasing()
+        public void TestTransformableString()
         {
             const string localisable_string_en_uppercase = "LOCALISED EN";
             const string localisable_string_en_title_case = "Localised EN";
--- a/osu.Framework.Tests/Localisation/LocalisationTest.cs
+++ b/osu.Framework.Tests/Localisation/LocalisationTest.cs
@@ -235,7 +235,7 @@ namespace osu.Framework.Tests.Localisation
             var dateTime = new DateTime(1);
             const string format = "MMM yyyy";
 
-            var text = manager.GetLocalisedString(new LocalisableFormattableString(dateTime, format));
+            var text = manager.GetLocalisedString(dateTime.ToLocalisableString(format));
 
             Assert.AreEqual("Jan 0001", text.Value);
 
@@ -276,6 +276,81 @@ namespace osu.Framework.Tests.Localisation
             Assert.AreEqual(lowercasedText.Value, "localised tr (i/ı)");
         }
 
+        [Test]
+        public void TestTranslatableEvaluatingLocalisableFormattableString()
+        {
+            const string key = FakeStorage.LOCALISABLE_NUMBER_FORMAT_STRING_EN;
+
+            manager.AddLanguage("fr", new FakeStorage("fr"));
+
+            var text = manager.GetLocalisedString(new TranslatableString(key, key, new LocalisableFormattableString(0.1234, "0.00%")));
+
+            Assert.AreEqual("number 12.34% EN", text.Value);
+
+            config.SetValue(FrameworkSetting.Locale, "fr");
+
+            Assert.AreEqual("number 12,34% FR", text.Value);
+        }
+
+        [Test]
+        public void TestTranslatableEvaluatingRomanisableString()
+        {
+            const string key = FakeStorage.LOCALISABLE_FORMAT_STRING_EN;
+
+            var text = manager.GetLocalisedString(new TranslatableString(key, key, new RomanisableString("unicode", "romanised")));
+
+            Assert.AreEqual("unicode localised EN", text.Value);
+
+            config.SetValue(FrameworkSetting.ShowUnicode, false);
+
+            Assert.AreEqual("romanised localised EN", text.Value);
+        }
+
+        [Test]
+        public void TestTranslatableEvaluatingTranslatableString()
+        {
+            const string key = FakeStorage.LOCALISABLE_FORMAT_STRING_EN;
+            const string nested_key = FakeStorage.LOCALISABLE_STRING_EN;
+
+            manager.AddLanguage("ja", new FakeStorage("ja"));
+
+            var text = manager.GetLocalisedString(new TranslatableString(key, key, new TranslatableString(nested_key, nested_key)));
+
+            Assert.AreEqual("localised EN localised EN", text.Value);
+
+            config.SetValue(FrameworkSetting.Locale, "ja");
+
+            Assert.AreEqual("localised JA localised JA", text.Value);
+        }
+
+        [Test]
+        public void TestTranslatableEvaluatingComplexString()
+        {
+            const string key = FakeStorage.LOCALISABLE_COMPLEX_FORMAT_STRING_EN;
+            const string nested_key = FakeStorage.LOCALISABLE_NUMBER_FORMAT_STRING_EN;
+
+            manager.AddLanguage("fr", new FakeStorage("fr"));
+
+            var text = manager.GetLocalisedString(new TranslatableString(key, key,
+                new LocalisableFormattableString(12.34, "0.00"),
+                new TranslatableString(nested_key, nested_key, new LocalisableFormattableString(0.9876, "0.00%")),
+                new TranslatableString(nested_key, nested_key, new RomanisableString("unicode", "romanised"))));
+
+            Assert.AreEqual("number 12.34 with number 98.76% EN and number unicode EN EN", text.Value);
+
+            config.SetValue(FrameworkSetting.Locale, "fr");
+
+            Assert.AreEqual("number 12,34 with number 98,76% FR and number unicode FR FR", text.Value);
+
+            config.SetValue(FrameworkSetting.ShowUnicode, false);
+
+            Assert.AreEqual("number 12,34 with number 98,76% FR and number romanised FR FR", text.Value);
+
+            config.SetValue(FrameworkSetting.Locale, "en");
+
+            Assert.AreEqual("number 12.34 with number 98.76% EN and number romanised EN EN", text.Value);
+        }
+
         private class FakeFrameworkConfigManager : FrameworkConfigManager
         {
             protected override string Filename => null;
@@ -302,6 +377,8 @@ namespace osu.Framework.Tests.Localisation
             public const string LOCALISABLE_FORMAT_STRING_JA = "{0} localised JA";
             public const string LOCALISABLE_NUMBER_FORMAT_STRING_EN = "number {0} EN";
             public const string LOCALISABLE_NUMBER_FORMAT_STRING_FR = "number {0} FR";
+            public const string LOCALISABLE_COMPLEX_FORMAT_STRING_EN = "number {0} with {1} and {2} EN";
+            public const string LOCALISABLE_COMPLEX_FORMAT_STRING_FR = "number {0} with {1} and {2} FR";
 
             public CultureInfo EffectiveCulture { get; }
 
@@ -353,6 +430,16 @@ namespace osu.Framework.Tests.Localisation
 
                             case "fr":
                                 return LOCALISABLE_NUMBER_FORMAT_STRING_FR;
+                        }
+
+                    case LOCALISABLE_COMPLEX_FORMAT_STRING_EN:
+                        switch (locale)
+                        {
+                            default:
+                                return LOCALISABLE_COMPLEX_FORMAT_STRING_EN;
+
+                            case "fr":
+                                return LOCALISABLE_COMPLEX_FORMAT_STRING_FR;
                         }
 
                     default:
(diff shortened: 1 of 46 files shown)


osu.Framework/Localisation/LocalisableString.cs
```diff
@@ -17,20 +17,27 @@ namespace osu.Framework.Localisation
         /// </summary>
         internal readonly object? Data;
 
+        /// <summary>
+        /// The case to apply to the underlying string data.
+        /// </summary>
+        public readonly Casing Casing;
```
smoogipoo, line 23: I don't like this existing in `LocalisableString`. It makes more sense in `TranslatableString` because those strings are coming from a different location.

osu.Framework/Localisation/LocalisationManager_LocalisedBindableString.cs
```diff
@@ -34,7 +35,7 @@ private void updateValue()
                         break;
 
                     case ILocalisableStringData data:
-                        Value = data.GetLocalised(parameters.Value);
+                        Value = applyCase(data.GetLocalised(parameters.Value));
```
smoogipoo, line 38: Likewise, this would only exist internal to `TranslatableString`.

smoogipoo (changes requested):

osu.Framework/Localisation/LocalisableString.cs
frenzibyte: There's only one case I can think of where having this as-is would help (or even bringing back the proposed `TextTransform` property), and that is localisable formatted strings.

Moving this to `TranslatableString` will limit the possibility of having uppercase'd date formatted string (for example).

smoogipoo: For niche use cases like that, I would expect an override of `LocalisableFormattableString`. The main point is that I don't want to increase the size of `LocalisableString` itself.

peppy: Agree with moving it. As per original discussion I was implying the implementation wouldn't necessarily be in the `LocalisableString` base class. I think `TranslateableString` is a good home for it for now.

smoogipoo: From another perspective, I don't know how this is going to work with the localisation analyser... Is this the reason it was done this way?

Game4all: I guess I may throw in the the rationnale for the proposed changes: basically the goal was to be able to allow changing the case of the final localised text whatever the underlying string type may be while providing an API mimicking somewhat how it is done with `string`. 

Moving this case stuff to a common class crossed my mind for a time but I decided still to store case info in `LocalisableString` itself to provide a `string` like API for any localisable string type (as long as it implements `ILocalisableStringData`) through extension methods. I understand the concern about keeping the struct size small but I think it doesn't add a too big overhead (4 bytes + the 8 previous bytes from the nullable string data) but I can certainly give a shot at moving case transform to a common class ...

bdach: Another seemingly possible avenue that may be worth exploring is composition via the decorator pattern, eg. have a `LocalisableStringWithCasing` which accepts another `LocalisableString` and applies the casing convention to it.

Game4all: I like this one altough declarations will get a bit long in some cases but adding extensions methods to `ILocalisableString` to wrap the string with the specified case should cover it.

Game4all: Just pushed what may be a potential implementation of the direction described above.

osu.Framework/Localisation/TransformableString.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <summary>
+        /// The underlying localisable string data of this transformable string.
+        /// </summary>
+        public readonly ILocalisableStringData? Data;
+
+        /// <summary>
+        /// Constructs a new transformable string with specified underlying string data and casing.
+        /// </summary>
+        /// <param name="data"></param>
+        /// <param name="casing"></param>
```
frenzibyte, line 30: Incomplete param docs

osu.Framework/Localisation/TransformableString.cs
```diff
@@ -0,0 +1,95 @@
...
+        public TransformableString(ILocalisableStringData data, Casing casing)
```
frenzibyte, line 31: Should accept a `LocalisableString` instead, all translatable strings defined in the `Strings` cases game-side are already implicitly converted to type `LocalisableString`:

CleanShot 2021-07-14 at 19 39 51

osu.Framework/Localisation/TransformableString.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            Data = data;
+            Casing = casing;
+        }
+
+        public string GetLocalised(LocalisationParameters parameters)
```
frenzibyte, line 37: Should match the rest of the string data implementations, for consistency purposes:
 - `ToString` returns cased strings with invariant culture.
 - `GetLocalised` returns cased strings with given culture, and falls back to `ToString` if the store is null.

frenzibyte (changes requested):
Few *small* details with the implementation, but sounds to be a good direction.

osu.Framework/Localisation/TransformableString.cs
bdach: Disagree with this. Doing this would force the extension methods to also accept `LocalisableString`s, which doesn't work with usages in which the type on which the extension is being called is not `LocalisableString` (i.e. the test added here does not compile).

osu.Framework/Localisation/TransformableString.cs
```diff
@@ -0,0 +1,95 @@
...
+            if (ReferenceEquals(this, other)) return true;
+
+            return Casing == other.Casing && Data == other.Data;
+        }
+    }
+
+    /// <summary>
+    /// Case applicable to the underlying localisable string data of a <see cref="TransformableString"/>.
+    /// </summary>
+    public enum Casing
```
bdach, line 73: Bit thrown off by the varying capitalisation of these enum members. Would stick to `UpperCase`, `LowerCase`, `TitleCase`.

osu.Framework/Localisation/TransformableString.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Globalization;
+
+#nullable enable
+
+namespace osu.Framework.Localisation
+{
+    /// <summary>
+    /// A string which can apply case transformations to underlying localisable string data.
+    /// </summary>
+    public class TransformableString : IEquatable<TransformableString>, ILocalisableStringData
```
bdach, line 14: Not super sold on this name. It doesn't mention casing in any way.

osu.Framework/Localisation/TransformableString.cs
frenzibyte: OK, this shouldn't actually force to accept a `LocalisableString`, but instead have an overload for each extension to accept `LocalisableString` and delegate that to `string.Data.ToXYZ();`.

Otherwise this won't work on usages, which I believe would be:
```csharp
Text = ButtonSystemStrings.Solo.ToTitle();
```

osu.Framework.Tests/Localisation/LocalisationTest.cs
```diff
@@ -242,6 +243,22 @@ public void TestLocalisableFormattableString()
             Assert.AreEqual("janv. 0001", text.Value);
         }
 
+        [Test]
+        public void TestTransformableString()
+        {
+            const string localisable_string_en_uppercase = "LOCALISED EN";
+            const string localisable_string_en_title_case = "Localised EN";
+
+            config.SetValue(FrameworkSetting.Locale, "en");
```
bdach, line 252: Test coverage is needed for non-English locales/implementations of `To{Lower,Upper}`, ie.

```
> new System.Globalization.CultureInfo("en").TextInfo.ToUpper("this text has an i in it")
"THIS TEXT HAS AN I IN IT"

> new System.Globalization.CultureInfo("tr").TextInfo.ToUpper("this text has an i in it")
"THİS TEXT HAS AN İ İN İT"

> new System.Globalization.CultureInfo("en").TextInfo.ToLower("THIS TEXT HAS AN I IN IT")
"this text has an i in it"

> new System.Globalization.CultureInfo("tr").TextInfo.ToLower("THIS TEXT HAS AN I IN IT")
"thıs text has an ı ın ıt"
```

bdach (changes requested):

osu.Framework/Localisation/TransformableString.cs
Game4all: Actually we may want to take a `LocalisableString` instead of the localisable string interface since `LocalisableString` can be constructed with a plain string. We may want to have overloads for both  `LocalisableString` and the interface for extensions methods.

bdach: Both is fine, I think.

osu.Framework/Localisation/TransformableString.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        public bool Equals(ILocalisableStringData? other) => other is TransformableString transformable && Equals(transformable);
+
+        public bool Equals(TransformableString? other)
+        {
+            if (ReferenceEquals(null, other)) return false;
+            if (ReferenceEquals(this, other)) return true;
+
+            return Casing == other.Casing && Data == other.Data;
```
smoogipoo, line 66: Use `Data.Equals(other.Data)` instead.

smoogipoo (changes requested):
I think the path taken now is good though. As you said, it should definitely take a `LocalisableString` as ctor param instead, and probably change the extension methods to `LocalisableString`.

osu.Framework.Tests/Localisation/LocalisationTest.cs
Game4all: Hopefully this should be covered by added test in 6307dbd

osu.Framework/Localisation/TransformableString.cs
```diff
@@ -80,7 +95,7 @@ public enum Casing
         /// <summary>
         /// Transform the string data to uppercase.
         /// </summary>
-        Uppercase,
+        UpperCase,
```
peppy, line 98: Was fine how you had it. "uppercase" is a valid word.

frenzibyte: It was done because of https://github.com/ppy/osu-framework/pull/4603#discussion_r669802447, which I agree with.

osu.Framework/Localisation/TransformableString.cs
```diff
@@ -0,0 +1,110 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
```
smoogipoo, line 3: Please annotate with `#nullable enable`.

smoogipoo (changes requested):

osu.Framework/Localisation/TransformableString.cs
smoogipoo: @peppy can you confirm whether you're fine with it as is, or do you require this to be changed?

peppy: It's fine either way, I was just going for conformity with other cases that have come up until now.

osu.Framework/Localisation/TransformableString.cs
frenzibyte: Gonna agree with this one. Since it's usually not constructed explicitly and is provided via extension methods, could rename to something more explicit but long, like `CaseTransformableString`.

Game4all: Sounds good to me.

frenzibyte (changes requested):

osu.Framework.Tests/Localisation/LocalisationTest.cs
```diff
@@ -242,6 +243,44 @@ public void TestLocalisableFormattableString()
...
+            Assert.AreEqual(uppercasedText.Value, FakeStorage.LOCALISABLE_STRING_EN_UPPERCASED);
+
+            config.SetValue(FrameworkSetting.Locale, "tr");
+            Assert.AreEqual(uppercasedText.Value, localisedUppercasedTrCultureString);
+
+            var lowercasedText = manager.GetLocalisedString(new TranslatableString(FakeStorage.LOCALISABLE_STRING_EN, FakeStorage.LOCALISABLE_STRING_EN).ToLower());
+            Assert.AreEqual(lowercasedText.Value, localisedLowercasedTrCultureString);
+
+            config.SetValue(FrameworkSetting.Locale, "en");
+            Assert.AreEqual(lowercasedText.Value, localisable_string_en_lowercase);
```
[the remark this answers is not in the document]
bdach: Agree with the above, so I've just went ahead and applied it more or less verbatim.

bdach (dismissed):

frenzibyte (approved):

smoogipoo (approved):
lgtm

PR #4604, opened 2021-07-13
Fix popovers fading out in wrong location

### LINKED ISSUE #4597: Previous popovers can appear in the wrong location during fade out
This is due to all popovers being in the position handling container, even if they are dismissed.

I will look at this later unless beaten by anyone.

https://user-images.githubusercontent.com/191335/125256130-192cb980-e337-11eb-8016-29e129152af6.mov

bdach:
Closes #4597.

Popovers that have been dismissed by the user could be displayed in the wrong location, due to sharing the `popoverContainer` with the currently displayed popover, which is attached to the current (new) target.

To resolve, use the popover itself as a replacement for the popover container. This works because both the removed container and the popover were/are 0x0 absolute-sized drawables, so the removed container's position can be combined with the move-back-in-bounds adjustment.

This isn't the general positioning refactor which was asked for, I'm more trying to just make things mostly work for now and move forward with other things.

### DIFF
--- a/osu.Framework/Graphics/Cursor/PopoverContainer.cs
+++ b/osu.Framework/Graphics/Cursor/PopoverContainer.cs
@@ -17,7 +17,6 @@ namespace osu.Framework.Graphics.Cursor
     public class PopoverContainer : Container
     {
         private readonly Container content;
-        private readonly Container<Popover> popoverContainer;
 
         private IHasPopover? target;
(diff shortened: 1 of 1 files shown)


frenzibyte (approved):

peppy (approved):
Looks to work as expected. Curiously I tried this quickly but managed to bork it up pretty fast.

PR #4605, opened 2021-07-13
Dismiss popovers on pressing Escape

### LINKED ISSUE #4600: Popover should be dismissed on pressing `Escape` key
In line with other UI elements (dropdown / menu / textbox).

bdach:
Closes #4600.

Aside from the automated testing here, I tested with the game-side room password PR. It half-works, because currently two escape presses are needed (the first one is handled by the password textbox to yield focus, and only the second one is seen by the popover itself). I *think* that's something to fix at the usage site...?

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestScenePopoverContainer.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestScenePopoverContainer.cs
@@ -129,6 +129,28 @@ namespace osu.Framework.Tests.Visual.UserInterface
         }
 
         [Test]
+        public void TestHideViaKeyboard()
+        {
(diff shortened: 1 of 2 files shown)


frenzibyte (approved):
Sounds correct. The textbox focus taking priority issue isn't really fixable on popover's end, so that can be left for now.

peppy (approved):

peppy:
> I tested with the game-side room password PR. It half-works, because currently two escape presses are needed (the first one is handled by the password textbox to yield focus, and only the second one is seen by the popover itself). I _think_ that's something to fix at the usage site...?

Matches my expectation, yep.

PR #4606, opened 2021-07-15
Remove speculative macOS crash fix

smoogipoo:
This neither fixed https://github.com/ppy/osu/issues/12944 nor peppy's crashes, so pretty safe to disable. I don't want to be causing too many unnecessary flushes.


peppy:
I'm just hoping that my *reduction* in crashes isn't related to this being present 😅 .

peppy (approved):

PR #4607, opened 2021-07-15
Fix refresh rate potentially being incorrect on startup

peppy:
Due to initialisation order of the window, there's no guarantee that the `DisplayMode` will be in a good state when the `frameLimiter.TriggerChange()` call is performed.

This converts that property to a bindable to allow for late initialisation.

Closes https://github.com/ppy/osu/issues/13888.

Has been tested by @PandaDriver156 to fix the issue (one of those that reported it). I can't test locally because parallels decided to gimp their shit even more.


smoogipoo (approved):
lgtm!

PR #4609, opened 2021-07-15
Minor optimization and cleanup for extension methods

huoyaoyuan:
nit: I know how to maximize performance, but it doesn't worth for code complexity.

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkHashing.cs
+++ b/osu.Framework.Benchmarks/BenchmarkHashing.cs
@@ -1,0 +1,38 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.IO;
+using BenchmarkDotNet.Attributes;
+using osu.Framework.Extensions;
+
+namespace osu.Framework.Benchmarks
+{
+    [MemoryDiagnoser]
+    public class BenchmarkHashing
+    {
+        private const string test_string = @"A string with reasonable length";
+        private MemoryStream memoryStream;
+
+        [Benchmark]
+        public string StringMD5() => test_string.ComputeMD5Hash();
+
+        [Benchmark]
+        public string StringSHA() => test_string.ComputeSHA2Hash();
+
+        [GlobalSetup]
+        public void GlobalSetup()
+        {
+            byte[] array = new byte[1024];
+            var random = new Random(42);
+            random.NextBytes(array);
+            memoryStream = new MemoryStream(array);
+        }
+
+        [Benchmark]
+        public string StreamMD5() => memoryStream.ComputeMD5Hash();
+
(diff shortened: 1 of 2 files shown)


osu.Framework/Extensions/EnumExtensions/EnumExtensions.cs
```diff
@@ -29,13 +26,10 @@ public static IEnumerable<T> GetValuesInOrder<T>()
         /// <summary>
         /// Get values of a collection of enum values in order. Supports custom ordering via <see cref="OrderAttribute"/>.
         /// </summary>
-        public static IEnumerable<T> GetValuesInOrder<T>(this IEnumerable<T> items)
+        public static IEnumerable<T> GetValuesInOrder<T>(this IEnumerable<T> items) where T : struct, Enum
```
bdach, line 29: This was unconstrained on purpose. See 74f1ed2effaa0d4c3731fdfaaf99b75968f8e5e5. If the constraint is to be brought back that will be a breaking change game-side. And from what I recall it wasn't that easy to get rid of that non-constrained usage.

huoyaoyuan: OK I'll revert this

peppy:
Generally we ask for a simple benchmark comparison when pure optimisations are applied. Can you add one to the benchmarks project and show a before-after comparison?

huoyaoyuan:
Benchmark added.

master:

|    Method |       Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------- |-----------:|---------:|---------:|-------:|------:|------:|----------:|
| StringMD5 |   900.8 ns |  5.64 ns |  5.28 ns | 0.1392 |     - |     - |   1,168 B |
| StringSHA | 1,200.7 ns | 11.01 ns | 10.30 ns | 0.1087 |     - |     - |     912 B |
| StreamMD5 | 2,435.1 ns | 20.10 ns | 17.82 ns | 0.0572 |     - |     - |     504 B |
| StreamSHA | 5,045.0 ns | 25.58 ns | 23.93 ns | 0.0839 |     - |     - |     760 B |

PR:

|    Method |       Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------- |-----------:|---------:|---------:|-------:|------:|------:|----------:|
| StringMD5 |   749.5 ns | 14.21 ns | 13.96 ns | 0.0420 |     - |     - |     352 B |
| StringSHA | 1,180.3 ns |  9.00 ns |  8.42 ns | 0.0534 |     - |     - |     448 B |
| StreamMD5 | 2,365.3 ns |  7.56 ns |  7.07 ns | 0.0343 |     - |     - |     296 B |
| StreamSHA | 5,052.7 ns | 24.95 ns | 22.12 ns | 0.0458 |     - |     - |     392 B |

peppy (approved):
Seems fine

PR #4612, opened 2021-07-17
Fix textbox readonly flag not working in all cases

### LINKED ISSUE #4611: Readonly textbox is not readonly
To repro:

1. Open `visual-tests` to `UserInterface` &rarr; `TextBox`
3. Select the readonly textbox by dragging on it (if you just click on it, it won't work)
4. Press `Ctrl+X`, `Ctrl+V` and/or `Backspace`
   - Those actions will do what you expect them to do on a normal, non readonly textbox

The textbox is not readonly, as those actions work.
Typing doesn't work, as intended.

bdach:
Closes #4611.

Found at least two cases in which focus was incorrectly grabbed or not yielded by a readonly textbox:

* When dragging - as reported in the aforementioned issue; fixed by 9ee8c9a.
* When setting flag to `false` after construction and addition to hierarchy; fixed by d440c50.

The thing that is tested most is the reception of focus by the textbox. I skimped on doing paste/delete thoroughly except for one step + assert for brevity, but those should be handled correctly as long as this guard is there:

https://github.com/ppy/osu-framework/blob/08ffacbb0733efeb8973842321aed7dab7f1d61e/osu.Framework/Graphics/UserInterface/TextBox.cs#L150-L155

and there is one deletion step in there so if that is ever removed then the test will fail anyhow (alongside many others, I would hope?)

The extra condition in the `OnMouseDown` guard has no visible/easily testable effect, it's just an optimisation.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
@@ -459,19 +459,19 @@ namespace osu.Framework.Tests.Visual.UserInterface
         [Test]
         public void TestReadOnly()
         {
-            BasicTextBox readOnlyTextBox = null;
-            BasicTextBox nonReadOnlyTextBox = null;
+            BasicTextBox firstTextBox = null;
+            BasicTextBox secondTextBox = null;
 
             AddStep("add textboxes", () => textBoxes.AddRange(new[]
             {
-                readOnlyTextBox = new BasicTextBox
+                firstTextBox = new BasicTextBox
                 {
                     Text = "Readonly textbox",
                     Size = new Vector2(500, 30),
                     ReadOnly = true,
                     TabbableContentContainer = textBoxes
                 },
-                nonReadOnlyTextBox = new BasicTextBox
+                secondTextBox = new BasicTextBox
                 {
                     Text = "Standard textbox",
                     Size = new Vector2(500, 30),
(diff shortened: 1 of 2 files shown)


frenzibyte:
While this does indeed fix the issue, this makes `ReadOnly` completely disable the interaction with the textbox, while usually a read-only textbox can be highlighted by users and can copy text from (in Windows and HTML at least).

This will have to depend on whether we do want that behaviour, however, as it seems read-only textboxes currently (in master) already cannot be tabbed to.

bdach:
I dunno. What you say could be achieved but at a cost of associated complexity. There are multiple varying notions of what "readonly" means on various OSes, in various apps, even HTML has readonly and disabled, which differ on whether the textbox can hold focus (and if so, whether it shows a caret), whether it's tabbable, whether selection of the containing text is supported. There are so many that I don't know which is right.

frenzibyte:
I would've thought it to be interactable but not editable given my experience with Windows, but yeah it seems different OSes have different behaviours for this, and separate between "readonly" and "disabled" which is even more weirder.

I actually can't really think of a case osu!-side where having "readonly" textboxes (interactable but not editable) has any point, so your approach sounds good for now, and continues off what `ReadOnly` does already (disables tabbing, disables clicking).

frenzibyte (approved):

peppy:
I wouldn't be against renaming to `Enabled` or `Disabled` to avoid any potential misunderstandings.

bdach:
I could rename `ReadOnly` to `Disabled`... if `Current.Disabled` wasn't entirely uncoupled from `ReadOnly`. And coupling them will break some workarounds for #3919.

Not to mention that osu!-side `TextBox` applies an alpha of 0.3 on `Disabled` but not `ReadOnly`...

peppy:
Let's just fix this one here for now. Doesn't seem like the correct PR to attempt to change the behaviour and API.

peppy (approved):

PR #4613, opened 2021-07-17
Show soft keyboard on Android on click of an already focused TextBox

Susko3:
Closes https://github.com/ppy/osu/issues/5284

Right now, if you manually close the keyboard by pressing the system back button, the keyboard will not open again if you tap on the currently focused TextBox. This PR fixes that.

`ToggleSoftInputFromWindow` is depricated in Android API level 31 (see [here](https://developer.android.com/reference/android/view/inputmethod/InputMethodManager#toggleSoftInputFromWindow(android.os.IBinder,%20int,%20int))) so `ShowSoftInput` is used instead.

**For iOS:**
The way `UpdateFirstResponder` works means that just calling `view.KeyboardTextField.UpdateFirstResponder(true)` in `ShowSoftKeyboard()` won't work as expected, as the keyboard might not close when the textbox is unfocused.

I don't have an iOS device to test with, so it's on somebody else to implement this at a later time. (And to test if hiding the keyboard manually even causes problems in the first place.)

### DIFF
--- a/osu.Framework.iOS/Input/IOSTextInput.cs
+++ b/osu.Framework.iOS/Input/IOSTextInput.cs
@@ -50,7 +50,7 @@ namespace osu.Framework.iOS.Input
             view.KeyboardTextField.UpdateFirstResponder(true);
         }
 
-        public void ShowSoftKeyboard()
+        public void EnsureActivated()
         {
             /// If the user has manually closed the keyboard, it will not be shown until another <see cref="Framework.Graphics.UserInterface.TextBox"/>
             /// is focused. Calling <see cref="IOSGameView.HiddenTextField.UpdateFirstResponder"/> over and over again won't work, due to how
--- a/osu.Framework.iOS/Input/IOSTextInput.cs
+++ b/osu.Framework.iOS/Input/IOSTextInput.cs
@@ -52,6 +52,11 @@ namespace osu.Framework.iOS.Input
 
         public void ShowSoftKeyboard()
         {
+            /// If the user has manually closed the keyboard, it will not be shown until another <see cref="Framework.Graphics.UserInterface.TextBox"/>
+            /// is focused. Calling <see cref="IOSGameView.HiddenTextField.UpdateFirstResponder"/> over and over again won't work, due to how
+            /// `responderSemaphore` currently works.
+
+            // TODO: add iOS implementation
         }
 
         public event Action<string> OnNewImeComposition
--- a/osu.Framework.iOS/Input/IOSTextInput.cs
+++ b/osu.Framework.iOS/Input/IOSTextInput.cs
@@ -50,6 +50,10 @@ namespace osu.Framework.iOS.Input
             view.KeyboardTextField.UpdateFirstResponder(true);
         }
 
+        public void ShowSoftKeyboard()
+        {
+        }
+
         public event Action<string> OnNewImeComposition
         {
             add { }
--- a/osu.Framework/Input/ITextInputSource.cs
+++ b/osu.Framework/Input/ITextInputSource.cs
@@ -20,13 +20,10 @@ namespace osu.Framework.Input
         void Activate(object sender);
 
         /// <summary>
-        /// Requests the OS to show the on-screen/software keyboard.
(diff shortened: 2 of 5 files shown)


osu.Framework/Input/ITextInputSource.cs
```diff
@@ -19,6 +19,15 @@ public interface ITextInputSource
...
         void Activate(object sender);
 
+        /// <summary>
+        /// Requests the OS to show the on-screen/software keyboard.
+        /// </summary>
+        /// <remarks>
+        /// The OS is hopefully smart enough not to show the software keyboard if a hardware one is present.
+        /// Should be reguraly called when doing text editing operations, as the user might have manually closed the software keyboard.
+        /// </remarks>
+        void ShowSoftKeyboard();
```
bdach, line 29: Can this be named differently? `Reactivate()`, maybe? `EnsureActivated()`? Having a mention of a "soft keyboard" just sticks out like a sore thumb in this very abstract interface.

osu.Framework.iOS/Input/IOSTextInput.cs
```diff
@@ -50,6 +50,15 @@ public void Activate(object sender)
             view.KeyboardTextField.UpdateFirstResponder(true);
         }
 
+        public void ShowSoftKeyboard()
+        {
+            /// If the user has manually closed the keyboard, it will not be shown until another <see cref="Framework.Graphics.UserInterface.TextBox"/>
+            /// is focused. Calling <see cref="IOSGameView.HiddenTextField.UpdateFirstResponder"/> over and over again won't work, due to how
+            /// `responderSemaphore` currently works.
```
bdach, line 57: Quite confused by this. This is theoretical, right? You did not actually test this?

bdach (changes requested):

osu.Framework.iOS/Input/IOSTextInput.cs
Susko3: Everything iOS is theoretical, I haven't tested anything.

But looking at this code, it seems that calling `UpdateFirstResponder(true)` multiple times is a bad idea. `responderSemaphore ` value will increase and the keyboard will never close on it's own.

https://github.com/ppy/osu-framework/blob/08ffacbb0733efeb8973842321aed7dab7f1d61e/osu.Framework.iOS/IOSGameView.cs#L214-L230

For the manually closing keyboard part:
Changing the `FirstResponder` (what presumably hides and shows the keyboard on iOS) is only done in `Activate` and `Deactivate` code in [IOSTextInput.cs](https://github.com/ppy/osu-framework/blob/master/osu.Framework.iOS/Input/IOSTextInput.cs), so it seems that nothing will open the keyboard if it's manually closed.

osu.Framework/Input/ITextInputSource.cs
Susko3: I agree that this should be renamed. But if it were, then what would stick out even more is that the newly renamed `ShowSoftKeyboard()` doesn't have `(object sender)` as an argument, in contrast to `Activate` and `Deactivate`. Looking at git blame, it seems this was added for `lock`ing purposes a long time ago in https://github.com/ppy/osu-framework/commit/a6e152cb2931b54f772b730825e6b85edcc4c432.

The only thing that uses `Activate` and `Deactivate` is `TextBox`, and no locking seems to be done here. Removing `(object sender)` from  `Activate` and `Deactivate` doesn't seem to cause issues. Should I remove that, or is this out of scope for this PR?

frenzibyte: That's irrelevant, you could just add `object sender` in the method and pass the same value that's passed in `Activate` and `Deactivate`, removing it can come in a separate PR if deemed right.

bdach: I'm not necessarily opposed to removing? No strong opinions here. It does seem to be completely unused right now.

smoogipoo:
I think it's probably fine to get this in with minimal testing since it only affects Android for the time being?

smoogipoo (approved):

PR #4614, opened 2021-07-17
Remove unused `object sender` in ITextInputSource

Susko3:
As discussed in https://github.com/ppy/osu-framework/pull/4613#discussion_r671718379.


frenzibyte (approved):

bdach (approved):

peppy:
Can't guarantee this won't be required in the future but I think if it is to be added back it won't be an `object`.

peppy (approved):

PR #4615, opened 2021-07-18
Add extension for localisable formattable strings

frenzibyte:
Similar to the C# built-in `ToString`, this adds a `ToLocalisableString` extension creating a `LocalisableFormattableString` for the `IFormattable` with the specified format string.

Simplifies usages from:
```csharp
Text = new LocalisableFormattableString(value, "#,##0");
```
to
```csharp
Text = value.ToLocalisableString("#,##0");
```


bdach (approved):
sure, why not

smoogipoo (approved):

PR #4621, opened 2021-07-19
Fix colour picker current not updating on HSV changes when not present

### LINKED ISSUE #4620: HSV colour picker should account for `Scheduler.HasPendingTasks` while not present
I was trying to implement a basic chroma/color cycling mode to my UI by changing the value of `Hue` in [HSVColourPicker_HueSelector](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Graphics/UserInterface/HSVColourPicker_HueSelector.cs) periodically.

As soon as the colour picker is not visible on the UI, the `Hue`'s `BindValueChanged` event stops triggering and the color of the remaining UI elements is not changed anymore.

If there is not already a property to override this behavior, presuming it is intentional, it'd be good in scenarios like this one to have it. 

However, as noted by @peppy earlier today, this is (most likely) just a misuse of `Bindable`, but I was still advised to post this here. That is, I'd be curious to know what the intended way of adding a rainbow cycle to my UI would be.

bdach:
Closes #4620.

This is incidentally the perfect counterexample for why blanket-including `this.Scheduler.HasPendingTasks` in the base `Drawable.IsPresent` is not a be-all and end-all of these types of issues.

Doing "the obvious" does not work with this component's structure. The child components are not the ones accessible outwardly, the picker itself is. And if someone is messing with the inner components, they are creating a problem entirely by themselves by misusing them.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneColourPicker.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneColourPicker.cs
@@ -1,9 +1,11 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.UserInterface;
+using osu.Framework.Testing;
 
 namespace osu.Framework.Tests.Visual.UserInterface
 {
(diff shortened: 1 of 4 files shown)


frenzibyte:
Did not take that into account... that drops the idea of "making scheduler run while not present" completely.

Though (and unfortunately) I think the `ColourPicker` should also override `IsPresent` as it is the one that would be hidden?

bdach:
Done.

smoogipoo (approved):

PR #4622, opened 2021-07-20
Flatten `PlatformAction` struct to an enum

ekrctb:
The `ActionType` and `ActionMethod` only make sense for text edit operations, and it was confusing and cumbersome to use.
`PlatformAction.Delete` is added as an additional "Delete" key binding separate from `DeleteForwardChar`, as the later should only be used for text edit.
I have osu-side change <https://github.com/ppy/osu/compare/master...ekrctb:platform-action?expand=1>.

--- 

# vNext

## `PlatformAction` type is changed to an enum type

If `PlatformAction.ActionType` was matched, change to a matching of `PlatformAction` itself like:

```diff
bool OnPressed(PlatformAction action) {
-    switch (action.ActionType)
+    switch (action)
    {
-        case PlatformActionType.Cut:
+        case PlatformAction.Cut:
```

For `PlatformActionMethod.Delete`, and if only the Delete key should be handled (not Backspace etc.), change to `PlatformAction.Delete`:

```diff
-    switch (action.ActionMethod)
+    switch (action)
    {
-        case PlatformActionMethod.Delete:
+        case PlatformAction.Delete:
```


ekrctb:
`CodeFactor` issues are false positive. It is not pointing to any changed code.

frenzibyte:
Or rather, they're just pointing issues in general, but only triggering when performing any change to said files.

peppy:
I think this looks like a good direction. Curious if anything specifically triggered the need for change though?

osu.Framework/Input/PlatformAction.cs
```diff
@@ -0,0 +1,47 @@
...
+        MoveToListEnd,
+        DocumentPrevious,
+        DocumentNext,
+        DocumentClose,
+        TabNew,
+        TabRestore,
+
+        // Text edit specific actions
+        MoveBackwardChar,
+        MoveForwardChar,
```
bdach, line 27: Not sure I entirely like the terminology change from `Previous/Next` to `Backward/Forward` but I guess it's not a huge deal...

osu.Framework.iOS/IOSGameHost.cs
```diff
@@ -130,34 +130,35 @@ public override IResourceStore<TextureUpload> CreateTextureLoaderStore(IResource
...
+            new KeyBinding(new KeyCombination(InputKey.Super, InputKey.C), PlatformAction.Copy),
+            new KeyBinding(new KeyCombination(InputKey.Super, InputKey.V), PlatformAction.Paste),
+            new KeyBinding(new KeyCombination(InputKey.Super, InputKey.A), PlatformAction.SelectAll),
+            new KeyBinding(InputKey.Left, PlatformAction.MoveBackwardChar),
+            new KeyBinding(InputKey.Right, PlatformAction.MoveForwardChar),
+            new KeyBinding(InputKey.BackSpace, PlatformAction.DeleteBackwardChar),
+            new KeyBinding(InputKey.Delete, PlatformAction.DeleteForwardChar),
+            new KeyBinding(new KeyCombination(InputKey.Shift, InputKey.Left), PlatformAction.SelectBackwardChar),
+            new KeyBinding(new KeyCombination(InputKey.Shift, InputKey.Right), PlatformAction.SelectForwardChar),
+            new KeyBinding(new KeyCombination(InputKey.Shift, InputKey.BackSpace), PlatformAction.DeleteForwardChar),
```
bdach, line 143:
```suggestion
            new KeyBinding(new KeyCombination(InputKey.Shift, InputKey.BackSpace), PlatformAction.DeleteBackwardChar),
```

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -161,34 +161,30 @@ private void load(GameHost host)
...
                     if (string.IsNullOrEmpty(SelectedText) || !AllowClipboardExport) return true;
 
                     clipboard?.SetText(SelectedText);
 
-                    if (action.ActionType == PlatformActionType.Cut)
+                    if (action == PlatformAction.Cut)
                     {
                         string removedText = removeSelection();
                         OnUserTextRemoved(removedText);
                     }
```
bdach, line 183: Opportunity to reuse `DeleteBy()` here.

osu.Framework/Platform/GameHost.cs
```diff
@@ -1001,40 +1001,41 @@ public void Dispose()
...
+            new KeyBinding(new KeyCombination(InputKey.Control, InputKey.C), PlatformAction.Copy),
+            new KeyBinding(new KeyCombination(InputKey.Control, InputKey.V), PlatformAction.Paste),
+            new KeyBinding(new KeyCombination(InputKey.Control, InputKey.A), PlatformAction.SelectAll),
+            new KeyBinding(InputKey.Left, PlatformAction.MoveBackwardChar),
+            new KeyBinding(InputKey.Right, PlatformAction.MoveForwardChar),
+            new KeyBinding(InputKey.BackSpace, PlatformAction.DeleteBackwardChar),
+            new KeyBinding(InputKey.Delete, PlatformAction.DeleteForwardChar),
+            new KeyBinding(new KeyCombination(InputKey.Shift, InputKey.Left), PlatformAction.SelectBackwardChar),
+            new KeyBinding(new KeyCombination(InputKey.Shift, InputKey.Right), PlatformAction.SelectForwardChar),
+            new KeyBinding(new KeyCombination(InputKey.Shift, InputKey.BackSpace), PlatformAction.DeleteForwardChar),
```
bdach, line 1014:
```suggestion
            new KeyBinding(new KeyCombination(InputKey.Shift, InputKey.BackSpace), PlatformAction.DeleteBackwardChar),
```

osu.Framework/Platform/MacOS/MacOSGameHost.cs
```diff
@@ -73,42 +73,43 @@ protected override IEnumerable<InputHandler> CreateAvailableInputHandlers()
...
+            new KeyBinding(new KeyCombination(InputKey.Super, InputKey.C), PlatformAction.Copy),
+            new KeyBinding(new KeyCombination(InputKey.Super, InputKey.V), PlatformAction.Paste),
+            new KeyBinding(new KeyCombination(InputKey.Super, InputKey.A), PlatformAction.SelectAll),
+            new KeyBinding(InputKey.Left, PlatformAction.MoveBackwardChar),
+            new KeyBinding(InputKey.Right, PlatformAction.MoveForwardChar),
+            new KeyBinding(InputKey.BackSpace, PlatformAction.DeleteBackwardChar),
+            new KeyBinding(InputKey.Delete, PlatformAction.DeleteForwardChar),
+            new KeyBinding(new KeyCombination(InputKey.Shift, InputKey.Left), PlatformAction.SelectBackwardChar),
+            new KeyBinding(new KeyCombination(InputKey.Shift, InputKey.Right), PlatformAction.SelectForwardChar),
+            new KeyBinding(new KeyCombination(InputKey.Shift, InputKey.BackSpace), PlatformAction.DeleteForwardChar),
```
bdach, line 86:
```suggestion
            new KeyBinding(new KeyCombination(InputKey.Shift, InputKey.BackSpace), PlatformAction.DeleteBackwardChar),
```

bdach (changes requested):

peppy:
@smoogipoo are you okay with this as a breaking change without obsolescence? It's a pretty quick fix for osu! (see https://github.com/ppy/osu/compare/master...ekrctb:platform-action?expand=1) and I don't think too many people would be using it in more complicated ways than we are osu! side.

smoogipoo:
I think without obsolescence is fine.

osu.Framework/Input/PlatformAction.cs
peppy: I think `Previous`/`Next` sounds better in all cases. Shall we switch to that?

osu.Framework/Input/PlatformAction.cs
```diff
@@ -0,0 +1,47 @@
...
+        DeleteBackwardChar,
+        DeleteForwardChar,
+        SelectBackwardChar,
+        SelectForwardChar,
+
+        MoveBackwardWord,
+        MoveForwardWord,
```
peppy, line 34: `MoveBackward` and `MoveForward` may be exceptions to the rule. They can likely remain as they are as they are describing motion rather than actions.

peppy (changes requested):
As propose.

osu.Framework/Input/PlatformAction.cs
ekrctb: A reason I changed the name is because I thought next/previous might be confusing when the current cursor is *on* the word/line:

```
               ↓ cursor
PreviousWord Wo|rd NextWord
```

ekrctb:
Also this "backward/forward word" terminology is in the emacs document <https://www.gnu.org/software/emacs/manual/html_node/emacs/Words.html>

peppy:
Do emacs actually use backward/forward anywhere for an operation that isn't movement? I can understand the potential confusion in the case you mention, but I think as it stands it's more confusing for the majority case, where we'd benefit from the clarity of using Previous/Next.

ekrctb:
In the linked doc, 

> M-DEL
>    Kill back to the beginning of a word (backward-kill-word).

peppy:
I guess it's fine as-is. The extra case where it may be deleting from within a current word makes me more neutral to which we should use, and what you have seems to work.

peppy (approved):

Issue #4164, opened 2021-01-22
Screen.OnResuming may be called on a still loading screen


peppy:
```csharp
stack.Push(root); 

root.Push(screen1); 
screen1.Push(screen2); 

/* both screen1 and screen2 are still loading */ 
stack.MakeCurrent(root);
 
/* OnResuming fired on screen1 */
```

This can be an issue if there are any transforms or drawable operations in the screen's `OnResuming` implementation, as seen in osu! side test failures:

```
  Failed TestPerformBlockedByDialog(True) [2 ms]
  Error Message:
   TearDown : System.AggregateException : One or more errors occurred. (Cannot mutate the Transforms of a Loading Drawable while not on the load thread. Consider using Schedule to schedule the mutation operation.)
  ----> osu.Framework.Graphics.Drawable+InvalidThreadForMutationException : Cannot mutate the Transforms of a Loading Drawable while not on the load thread. Consider using Schedule to schedule the mutation operation.
Data:
  unhandled: unhandled

  Stack Trace:
  --TearDown
   at osu.Framework.Testing.TestScene.checkForErrors()
   at osu.Framework.Testing.TestScene.RunTests()
--InvalidThreadForMutationException
   at osu.Framework.Graphics.Drawable.EnsureMutationAllowed(String member)
   at osu.Framework.Graphics.Transforms.Transformable.AddTransform(Transform transform, Nullable`1 customTransformID)
   at osu.Framework.Graphics.TransformableExtensions.TransformTo[TThis](TThis t, Transform transform)
   at osu.Framework.Graphics.TransformableExtensions.MoveToX[T,TEasing](T drawable, Single destination, Double duration, TEasing& easing)
   at osu.Framework.Graphics.TransformableExtensions.MoveToX[T](T drawable, Single destination, Double duration, Easing easing)
   at osu.Game.Screens.BackgroundScreen.OnResuming(IScreen last) in Y:\osu\osu.Game\Screens\BackgroundScreen.cs:line 79
   at osu.Framework.Screens.ScreenStack.resumeFrom(IScreen source)
   at osu.Framework.Screens.ScreenStack.exitFrom(IScreen source, Boolean shouldFireExitEvent, Boolean shouldFireResumeEvent)
   at osu.Framework.Screens.ScreenStack.Exit(IScreen source)
   at osu.Game.Screens.OsuScreen.OnExiting(IScreen next) in Y:\osu\osu.Game\Screens\OsuScreen.cs:line 212
[...]

PR #4623, opened 2021-07-20
Add more testing of `ScreenStack`

peppy:
- Makes every `ScreenStack` test also test mutation allowed in callbacks
- Add test coverage of reported breaking scenario (closes https://github.com/ppy/osu-framework/issues/4164 as not reproducible).

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneScreenStack.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneScreenStack.cs
@@ -1036,6 +1036,8 @@ namespace osu.Framework.Tests.Visual.UserInterface
 
             public override void OnEntering(IScreen last)
             {
+                attemptTransformMutation();
+
                 EnteredFrom = last;
                 Entered?.Invoke();
 
@@ -1060,6 +1062,8 @@ namespace osu.Framework.Tests.Visual.UserInterface
 
             public override bool OnExiting(IScreen next)
             {
+                attemptTransformMutation();
+
                 ExitedTo = next;
                 Exited?.Invoke();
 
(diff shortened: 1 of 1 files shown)


osu.Framework.Tests/Visual/UserInterface/TestSceneScreenStack.cs
```diff
@@ -888,6 +888,21 @@ public void TestGetChildScreenAndGetParentScreenReturnNullWhenNotInStack()
...
+        /// Ensure that an intermediary screen doesn't block and doesn't attempt to fire events when not loaded.
+        /// </summary>
+        [Test]
+        public void TestMakeCurrentWhileScreensStillLoading()
+        {
+            TestScreen root = null;
+
+            pushAndEnsureCurrent(() => root = new TestScreen(id: 1));
+            AddStep("push slow", () => stack.Push(new TestScreenSlow { Exiting = () => true }));
+            AddStep("push second slow", () => stack.Push(new TestScreenSlow()));
```
smoogipoo, line 901: Make sure to allow the load to complete for these screens, otherwise they're going to be holding a BDL thread for the timeout duration (10s).

Maybe it's worth binding to `ScreenPushed` from inside `[SetUp]` and automatically setting AllowLoad. I _think_ `ScreenPushed` is called immediately iirc.

smoogipoo (changes requested):

osu.Framework.Tests/Visual/UserInterface/TestSceneScreenStack.cs
peppy: Good catch.

smoogipoo (approved):

PR #4627, opened 2021-07-20
Fix DrawableTrack/DrawableSample not receiving new parent adjustments when parent changes

smoogipoo:
I don't know if this breaks anything (quick check says no) - hopefully we've built around the assumption that this is working correctly, but it's now working as it was initially intended.


bdach (approved):
makes sense to me

peppy (approved):
ouch

PR #4629, opened 2021-07-21
Allow adding query string params to non-GET `WebRequest`s

bdach:
- [x] Depends on #4628 for mergeability

PRing for discussion. The end goal is to use this change for https://github.com/ppy/osu/issues/13966 and fixing a TODO item therein in one fell swoop, by applying the following diff game-side:

```diff
diff --git a/osu.Game/Online/Rooms/JoinRoomRequest.cs b/osu.Game/Online/Rooms/JoinRoomRequest.cs
index b2d772cac7..d9d4f2eb5c 100644
--- a/osu.Game/Online/Rooms/JoinRoomRequest.cs
+++ b/osu.Game/Online/Rooms/JoinRoomRequest.cs
@@ -22,10 +22,10 @@ protected override WebRequest CreateWebRequest()
         {
             var req = base.CreateWebRequest();
             req.Method = HttpMethod.Put;
+            req.AddParameter(@"password", Password, RequestParameterType.Query);
             return req;
         }
 
-        // Todo: Password needs to be specified here rather than via AddParameter() because this is a PUT request. May be a framework bug.
-        protected override string Target => $"rooms/{Room.RoomID.Value}/users/{User.Id}?password={Password}";
+        protected override string Target => $@"rooms/{Room.RoomID.Value}/users/{User.Id}";
     }
 }
```

The reason why I am proposing to allow adding both query and form params simultaneously is that many existing tools and frameworks already support it. cURL, postman, ASP.NET Core all will happily support both query string and form parameters on any request that has a body. So from a framework perspective I feel it's better to be less "opinionated"/rigid than more.

The change was designed to be non-breaking in any reasonable scenario (if anyone is expecting things like `OPTIONS` requests to work with form params, they have another thing coming).

### DIFF
--- a/osu.Framework/IO/Network/WebRequest.cs
+++ b/osu.Framework/IO/Network/WebRequest.cs
@@ -626,7 +626,8 @@ namespace osu.Framework.IO.Network
 
         /// <summary>
         /// Adds a raw POST body to this request.
-        /// This may not be used in conjunction with <see cref="AddFile"/> and <see cref="AddParameter(string,string,RequestParameterType)"/>.
+        /// This may not be used in conjunction with <see cref="AddFile"/>
+        /// and <see cref="AddParameter(string,string,RequestParameterType)"/> with the request type of <see cref="RequestParameterType.Form"/>.
         /// </summary>
         /// <param name="stream">The stream containing the raw data. This stream will _not_ be finalized by this request.</param>
         public void AddRaw(Stream stream)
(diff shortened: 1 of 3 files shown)


osu.Framework/IO/Network/WebRequest.cs
```diff
@@ -288,7 +294,7 @@ private async Task internalPerform(CancellationToken cancellationToken = default
 
                         if (rawContent != null)
                         {
-                            if (parameters.Count > 0)
+                            if (formParameters.Count > 0)
                                 throw new InvalidOperationException($"Cannot use {nameof(AddRaw)} in conjunction with {nameof(AddParameter)}");
```
peppy, line 298: I think this exception text can also be updated to read similar to the one below (only incompatible with form/file parameters).

peppy (approved):
Seems fine otherwise.

peppy (approved):

PR #4631, opened 2021-07-24
Hide popovers on target disposal

bdach:
Aims to ensure that popovers are cleaned up properly when their target gets disposed.

The schedule of `SetTarget(null)` is required as disposal can happen on non-update thread - omitting it was causing invalid transform mutation exceptions due to indirectly invoking `PopOut` otherwise. The scheduled `SetTarget(null)` call intends to ensure that both the reference to the disposed drawable is released and that the `OnDispose` subscription is removed.

### DIFF
--- a/osu.Framework/Graphics/Cursor/PopoverContainer.cs
+++ b/osu.Framework/Graphics/Cursor/PopoverContainer.cs
@@ -44,19 +44,27 @@ namespace osu.Framework.Graphics.Cursor
         /// <returns><see langword="true"/> if a new popover was shown, <see langword="false"/> otherwise.</returns>
         internal bool SetTarget(IHasPopover? newTarget)
         {
-            target = newTarget;
-
             currentPopover?.Hide();
             currentPopover?.Expire();
+
+            if (target is Drawable oldDrawableTarget)
+                oldDrawableTarget.OnDispose -= onTargetDisposed;
+
+            target = newTarget;
 
             var newPopover = target?.GetPopover();
             if (newPopover == null)
                 return false;
 
+            if (target is Drawable newDrawableTarget)
+                newDrawableTarget.OnDispose += onTargetDisposed;
+
             AddInternal(currentPopover = newPopover);
             currentPopover.Show();
             return true;
         }
+
+        private void onTargetDisposed() => Schedule(() => SetTarget(null));
 
         protected override void UpdateAfterChildren()
         {
@@ -181,6 +189,14 @@ namespace osu.Framework.Graphics.Cursor
 
             // the final size is the intersection of the X/Y areas.
             return availableSize;
+        }
+
+        protected override void Dispose(bool isDisposing)
+        {
+            base.Dispose(isDisposing);
+
+            if (target is Drawable drawableTarget)
+                drawableTarget.OnDispose -= onTargetDisposed;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Cursor/PopoverContainer.cs
```diff
@@ -44,20 +44,28 @@ public PopoverContainer()
...
+
             var newPopover = target?.GetPopover();
             if (newPopover == null)
                 return false;
 
+            if (target is Drawable newDrawableTarget)
+                newDrawableTarget.OnDispose += onTargetDisposed;
+
             AddInternal(currentPopover = newPopover);
             currentPopover.Show();
```
frenzibyte, line 63: very nit, but I feel it would read better to unsubscribe and subscribe to the dispose event before expire and after creation:
```suggestion
            if (target is Drawable oldDrawableTarget)
                oldDrawableTarget.OnDispose -= onTargetDisposed;

            currentPopover?.Hide();
            currentPopover?.Expire();

            target = newTarget;

            var newPopover = target?.GetPopover();
            if (newPopover == null)
                return false;

            AddInternal(currentPopover = newPopover);
            currentPopover.Show();

            if (target is Drawable newDrawableTarget)
                newDrawableTarget.OnDispose += onTargetDisposed;
```

frenzibyte (commented):
All good other than that small nit

osu.Framework/Graphics/Cursor/PopoverContainer.cs
bdach: Why would that order matter?

frenzibyte: It does not matter in terms of execution or anything else obviously, just felt it would read better that way, nothing else.

bdach: I don't follow why one would read worse than the other.

frenzibyte: More along the lines of `onTargetDisposed` not being necessary if the current popover isn't created yet, so it would make sense to put that after the creation.

But now that I read it again, it doesn't really matter which way to go as the subscribe logic doesn't relate itself with popovers by the naming, so you can disregard this.

frenzibyte (approved):

peppy:
Curious if there's a reason you didn't use the same approach that `ContextMenuContainer` uses (parent traversal to find a valid/`this` `ContextMenuContainer` on `target`). It seems like it may work, and may be a more tested method of implementation.

Anything binding to `IsDisposed` feels a bit awkward.

bdach:
I guess not, aside from context menu's method being more expensive. I can also add the presence check that context menu has as well, although do note that it won't work in absolutely all scenarios as presence flags are not propagated down the tree (so if the target's *parent* gets hidden, it won't work).

peppy:
I think the saving factor here is that if a parent of the `ContextMenuContainer` was made non-present then it's safe to say its children are too, so still displaying the menu is safe in that case (although may affect any hide animations i guess).

I'm just most interested here in using what we know already works rather than using a new solution.

@smoogipoo thoughts on this?

bdach:
In the interest of moving forward with this pull (and its dependency, which are both blockers for colour picker integration in the editor) I have pushed changes to use the same method of hiding as context container does.

It looks to work slightly better in practice somehow anyways (seems to hide the popover slightly earlier when exiting out of a screen with one showing?)

peppy (approved):

smoogipoo:
I agree with the recent changes.

smoogipoo (approved):

PR #4632, opened 2021-07-24
Ensure that popover handles all mouse events within the bounds of its body

bdach:
- [x] Depends on #4631 for mergeability

PRing for discussion. The reason that this is written so awkwardly (and the primary issue I was intending to resolve) was that with the current implementation, hover events were leaking to background drawables.

Unfortunately, doing

```csharp
protected override bool OnHover(HoverEvent e) => Body.ReceivePositionalInputAt(e.ScreenSpaceMousePosition);
```

is not viable. The reason why is that as you may recall, popovers have `ReceivePositionalInputEvent()` overridden to return `true` for any screen-space point. This in turn makes it so that the popover is *always* hovered, so long as the mouse is in bounds of the screen.

The thing about hover events is that a drawable will (correctly) *not* have its `OnHover()` called again if it is hovered, until it is finally unhovered. And the popover will _never_ become unhovered, again, as long as the mouse is within the screen. This makes it so that unless the mouse is *inside* the popover when it returns to the window, it will never be able to handle a hover again. That's the reason for the container hack - it will only handle input within its bounds, and *not* be hovered when mouse is anywhere else on the screen.

The biggest hang-up about this diff is the click handling. Unfortunately as the parent popover is the focus-controlling drawable, the mouse-blocking container has to let clicks through so that the popover can handle them itself and receive and yield focus correctly.

I can think of a few ways of resolving this differently, but they would be reverting to some of the previous behaviour (either blocking mouse events over the entire screen, or moving the focused container lower down and doing a `State` bind from the `Popover` to it, etc.) Willing to listen to alternatives.

### DIFF
--- a/osu.Framework/Graphics/Cursor/PopoverContainer.cs
+++ b/osu.Framework/Graphics/Cursor/PopoverContainer.cs
@@ -7,6 +7,7 @@ using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.UserInterface;
+using osu.Framework.Input.Events;
 using osu.Framework.Utils;
 using osuTK;
 
@@ -17,6 +18,7 @@ namespace osu.Framework.Graphics.Cursor
     public class PopoverContainer : Container
     {
         private readonly Container content;
+        private readonly Container dismissOnMouseDownContainer;
 
         private IHasPopover? target;
         private Popover? currentPopover;
@@ -31,6 +33,10 @@ namespace osu.Framework.Graphics.Cursor
                 {
                     RelativeSizeAxes = Axes.Both,
                 },
+                dismissOnMouseDownContainer = new DismissOnMouseDownContainer
+                {
+                    RelativeSizeAxes = Axes.Both
+                }
             };
         }
 
@@ -59,7 +65,7 @@ namespace osu.Framework.Graphics.Cursor
             if (target is Drawable newDrawableTarget)
                 newDrawableTarget.OnDispose += onTargetDisposed;
 
-            AddInternal(currentPopover = newPopover);
+            dismissOnMouseDownContainer.Add(currentPopover = newPopover);
             currentPopover.Show();
             return true;
         }
@@ -197,6 +203,15 @@ namespace osu.Framework.Graphics.Cursor
 
             if (target is Drawable drawableTarget)
                 drawableTarget.OnDispose -= onTargetDisposed;
+        }
+
+        private class DismissOnMouseDownContainer : Container
+        {
+            protected override bool OnMouseDown(MouseDownEvent e)
+            {
+                this.HidePopover();
+                return false;
+            }
         }
     }
 }
--- a/osu.Framework.Tests/Visual/UserInterface/TestScenePopoverContainer.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestScenePopoverContainer.cs
@@ -599,24 +599,54 @@ namespace osu.Framework.Tests.Visual.UserInterface
 
         private class EventHandlingContainer : Container
         {
+            private readonly Box colourBox;
+
             public bool ClickReceived { get; private set; }
             public bool HoverReceived { get; private set; }
+
+            protected override Container<Drawable> Content { get; }
+
+            public EventHandlingContainer()
+            {
+                AddInternal(new Container
+                {
+                    RelativeSizeAxes = Axes.Both,
+                    Children = new Drawable[]
+                    {
+                        colourBox = new Box
+                        {
+                            Colour = Color4.Black,
+                            RelativeSizeAxes = Axes.Both,
+                        },
+                        Content = new Container { RelativeSizeAxes = Axes.Both },
+                    }
+                });
+            }
 
             public void Reset()
             {
                 ClickReceived = HoverReceived = false;
+                colourBox.FadeColour(Color4.Black);
             }
 
             protected override bool OnClick(ClickEvent e)
             {
                 ClickReceived = true;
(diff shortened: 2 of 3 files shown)


frenzibyte:
This is the one alternative I thought about, which makes `Popover` no longer having to care about blocking and not blocking on certain circumstances:
```diff
diff --git a/osu.Framework/Graphics/Cursor/PopoverContainer.cs b/osu.Framework/Graphics/Cursor/PopoverContainer.cs
index f6939c9fa..b7a09a3f5 100644
--- a/osu.Framework/Graphics/Cursor/PopoverContainer.cs
+++ b/osu.Framework/Graphics/Cursor/PopoverContainer.cs
@@ -7,6 +7,7 @@
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.UserInterface;
+using osu.Framework.Input.Events;
 using osu.Framework.Utils;
 using osuTK;

@@ -17,6 +18,7 @@ namespace osu.Framework.Graphics.Cursor
     public class PopoverContainer : Container
     {
         private readonly Container content;
+        private readonly Container popoverContainer;

         private IHasPopover? target;
         private Popover? currentPopover;
@@ -31,6 +33,10 @@ public PopoverContainer()
                 {
                     RelativeSizeAxes = Axes.Both,
                 },
+                popoverContainer = new DismissOnMouseDownContainer
+                {
+                    RelativeSizeAxes = Axes.Both,
+                }
             };
         }

@@ -59,7 +65,7 @@ internal bool SetTarget(IHasPopover? newTarget)
             if (target is Drawable newDrawableTarget)
                 newDrawableTarget.OnDispose += onTargetDisposed;

-            AddInternal(currentPopover = newPopover);
+            popoverContainer.Add(currentPopover = newPopover);
             currentPopover.Show();
             return true;
         }
@@ -198,5 +204,14 @@ protected override void Dispose(bool isDisposing)
             if (target is Drawable drawableTarget)
                 drawableTarget.OnDispose -= onTargetDisposed;
         }
+
+        private class DismissOnMouseDownContainer : Container
+        {
+            protected override bool OnMouseDown(MouseDownEvent e)
+            {
+                this.HidePopover();
+                return base.OnMouseDown(e);
+            }
+        }
     }
 }
diff --git a/osu.Framework/Graphics/UserInterface/Popover.cs b/osu.Framework/Graphics/UserInterface/Popover.cs
index 803ae920b..85b7cf59d 100644
--- a/osu.Framework/Graphics/UserInterface/Popover.cs
+++ b/osu.Framework/Graphics/UserInterface/Popover.cs
@@ -19,18 +19,11 @@ namespace osu.Framework.Graphics.UserInterface
     /// </summary>
     public abstract class Popover : FocusedOverlayContainer
     {
-        protected override bool BlockPositionalInput => false; // not required as we only care about intercepting mouse down in certain cases.
+        protected override bool BlockPositionalInput => true;

-        public override bool ReceivePositionalInputAt(Vector2 screenSpacePos) => State.Value == Visibility.Visible;
-
-        protected override bool OnMouseDown(MouseDownEvent e)
-        {
-            this.HidePopover();
-            return base.OnMouseDown(e);
-        }
-
-        // click is handled here instead of in MouseBlockingContainer so that the popover can be focused.
-        protected override bool OnClick(ClickEvent e) => Body.ReceivePositionalInputAt(e.ScreenSpaceMousePosition);
+        public override bool ReceivePositionalInputAt(Vector2 screenSpacePos)
+            => Body.ReceivePositionalInputAt(screenSpacePos) ||
+               Arrow.ReceivePositionalInputAt(screenSpacePos);

         public override bool HandleNonPositionalInput => State.Value == Visibility.Visible;

@@ -94,7 +87,7 @@ protected Popover()
                 Children = new[]
                 {
                     Arrow = CreateArrow(),
-                    Body = new MouseBlockingContainer
+                    Body = new Container
                     {
                         AutoSizeAxes = Axes.Both,
                         Children = new Drawable[]
@@ -200,24 +193,5 @@ public new Vector2 Size
         }

         #endregion
-
-        private class MouseBlockingContainer : Container
-        {
-            protected override bool Handle(UIEvent e)
-            {
-                switch (e)
-                {
-                    case ClickEvent _:
-                        // let clicks be handled by the popover itself, since it can receive focus.
-                        return false;
-
-                    case MouseEvent _:
-                        return true;
-
-                    default:
-                        return false;
-                }
-            }
-        }
     }
 }
 ```

...at the cost of another special container that overrides `OnMouseDown` to do popover stuff.

The way the code works is that when a mouse event isn't handled by the popover (which defer to its body's bounds), then the event will reach `DismissOnMouseDownContainer` which will dismiss the popover.

Then `Popover` can override `BlockPositionalInput` to true for mouse events to not pass through its body.

This change will also correct `IsHovered`'s state to be based on its content, rater than always true.

**Comment by bdach** [comment]
Some questionable variable naming aside the above looked slightly better to me, so I've applied that more or less verbatim. Seems to work, passes tests at least.

File: osu.Framework/Graphics/Cursor/PopoverContainer.cs [hunk 1]
```diff
@@ -188,5 +194,14 @@ private Vector2 availableSizeAroundTargetForAnchor(Quad targetLocalQuad, Anchor
             // the final size is the intersection of the X/Y areas.
             return availableSize;
         }
+
+        private class DismissOnMouseDownContainer : Container
+        {
+            protected override bool OnMouseDown(MouseDownEvent e)
+            {
+                this.HidePopover();
+                return false;
```
peppy, line 203: Arguable we may want to change this to return `true` in the future, but let's go with this for now until an actual usability issue comes up.

peppy (approved):

peppy:
I updated the tests a bit just to confirm that the input catch container is doing what it says it is (and also add visual feedback of what is going on).

Issue #2525, opened 2019-06-11
Switching the keyboard language makes you unable to type in any textbox


EVAST9919:
- Type some stuff in any textbox
- Switch language (no matter how many times) **(using Win + Space)**
- You can't type anymore (switching back doesn't help)

Tested on Win 10 1903 (buld 18362.145) with English/Russian languages.

huoyaoyuan:
Also hitting this.
Only reproduces if you switch using Win+Space. Click on task bar switches successfully.

huoyaoyuan:
Tested with #3013, SDL does not have this issue.

Joehuu:
Seems to have regressed in 2020.609.0, probably in https://github.com/ppy/osu-framework/pull/3494.

Susko3:
This is an issues with SDL: https://github.com/libsdl-org/SDL/issues/4369

Susko3:
This issue is fixed in SDL with commit https://github.com/libsdl-org/SDL/commit/3b85e3fdfce01c528a6ebf9ac537bdc106ba4963

Issue #2198, opened 2019-02-27
Text-box focus problems with Windows Gaming Bar


Gabixel:
**Describe your problem:** Every text-box in osu!lazer behave strangely with the Windows Gaming Bar (`Windows` + `G`) when you close it. The best way to get everything back to normal is to focus another application, and then re-focus osu!lazer.

Also, if you type something after doing this, it causes all previously pressed keys (after closing the gaming bar) to appear at that moment.

One weird thing is that some keys still seem to work in some situations with this issue (like left and right arrows still interact with carousel).

**Screenshots or videos showing encountered issue:** 

https://user-images.githubusercontent.com/43073074/105320502-881a2880-5bc6-11eb-8d40-735f5ae48954.mp4

In the video, I used the on-screen keyboard for a better demonstration, but the same happens without using it.
Also, you can see that I use applications that could interfere with shortcuts. I can confirm that by closing those programs, the problem persists.

**osu!lazer version:** `2019.227.0` (updated in: `2021.118.1`)

Susko3:
This is a problem with SDL, upstream issue: https://github.com/libsdl-org/SDL/issues/4369
Same underlaying problem as https://github.com/ppy/osu-framework/issues/2525

*From https://github.com/libsdl-org/SDL/issues/4369:*
>The popup doesn't take window focus from the SDL window (the window will appear to be in focus), but it takes keyboard focus (pressing keyboard keys doesn't produce SDL_KEYUP/DOWN events).

The game bar also does this, so it's the same underlying problem with SDL.

PR #4634, opened 2021-07-25
Update SDL libraries with native library bump

peppy:
Closes https://github.com/ppy/osu-framework/issues/2198.
Closes https://github.com/ppy/osu-framework/issues/2525.
Closes https://github.com/ppy/osu/issues/11489.


bdach:
Linux looks fine with this bump. Have not checked windows yet.

peppy:
I've only tested on macOS, and had no issues there. @Susko3 are you able to confirm this fixes (at least one of) the bugs linked to make sure we're on the right track?

Susko3:
I'm not able to test right now, I'll be available to test in 1-2 days.

bdach:
Have tested with windows and both the <kbd>Win</kbd>+<kbd>Space</kbd> case and the game bar case look to be fixed.

bdach (approved):

PR #4635, opened 2021-07-25
Split `TextChunk` from `TextFlowContainer`

bdach:
This is a mostly "code quality" sort of change, but one that is pretty conducive to adding localisation to `TextFlowContainer` on the path that I have explored on my WIP/MVP branch. The commits are mostly simple source transformations:

* 578d6d3 renames `TextLine` to `TextChunk`. It was never really a "line" of text anyways; adding consecutive "lines" via repeated `AddText()` calls would make the text flow as normal and *not* add lines. The actual "new line" logic is in `AddLine()`.
* dfd39a8 splits the class to a new file - I intend it to grow, and contain all of the drawable-constructing logic that currently resides in the text flow itself.
* ede05f7 folds the `newLineIsParagraph` flag into the chunk - aside from simplifying data flow, it ensures that all information required to construct the drawable representation of a chunk resides *in* the chunk.


peppy (approved):
👍🏻

PR #4638, opened 2021-07-27
Add win-arm64 native libs

boswelja:
I've read over the FFMPEG License checklist (https://www.ffmpeg.org/legal.html), but I'm not too sure how required all the items are since it looks like we've already got DLLs for x64/x86 that don't have compile instructions, source archives, and have been compiled with `--enable-gpl`.

I've provided a compile script with this PR, but I haven't included the source archives directly, nor have I used `--enable-gpl` (which appears to be a requirement for postproc). Compile script based on the one provided by MATRIX-feather for #3846.

All tests in VisualTests appear to run successfully on my device (Surface Pro X), but I was unable to test in-game due to other libraries used by osu! not supporting win-arm64.


peppy:
Which osu!-side libraries did you find as blocking?

boswelja:
SQLite threw DLLNotFound right off the bat. It seems to be fixed in newer versions, but since they don't seem to be backwards-compatible, we'll probably need to upgrade EFCore.SQLite too.

peppy:
For my own sanity, can you make sure realm works correctly? We are in the process of migrating to that, and while on an initial check it does look to support "UWP", that doesn't guarantee ARM support.

boswelja:
Sure thing, will running the Realm example project suffice?

peppy:
Yep, should be fine!

boswelja:
Looks like Realm loads fine on ARM64 already. I did have to create a new project to test though, since I couldn't get the examples to launch, but that's probably my fault since the examples use Xamarin Forms and I have no idea how to run those projects.

bdach:
> it looks like we've already got DLLs for x64/x86 that don't have compile instructions, source archives, and have been compiled with `--emable-gpl`

How did you determine that there are DLLs with `--enable-gpl` on?

boswelja:
> > it looks like we've already got DLLs for x64/x86 that don't have compile instructions, source archives, and have been compiled with `--emable-gpl`
> 
> How did you determine that there are DLLs with `--enable-gpl` on?

As far as I can tell, postproc requires --enable-gpl, and simply using --enable-postproc wasn't enough to get it to build.

bdach:
That should probably *not* be bundled, then. Is postproc actually required to launch and run? The mobile nativelibs don't seem to have it so I'd hope not.

boswelja:
Nope, from running the tests included, none failed due to missing libs so it looks like it's not required.
We should probably look at the other bundled binaries anyways. Set up compile scripts etc to make updating them easier and make an attempt to comply with the checklist.

osu.Framework.NativeLibs/runtimes/win-arm64/source/compile-ffmpeg.sh
```diff
@@ -0,0 +1,23 @@
...
+echo "Configuring FFMPEG build"
+./configure --arch=aarch64 --target-os=mingw32 --cross-prefix=aarch64-w64-mingw32- \
+--enable-shared --disable-static --prefix=build --disable-programs --disable-doc
+
+echo "Compiling FFMPEG binaries"
+make -j
+make install
+
+echo "Moving binaries to native folder"
+cp build/bin/*.dll ../../native
```
bdach, line 23: I think technically all these steps should be doing `|| exit 1` to avoid stumbling forward where there is no real point to?

boswelja: I think make exits on failure anyways, but I've added `|| exit 1` just in case

jai-x: I think you could just add `set -e` at the top of the script instead of appending `|| exit 1` to each line.

boswelja: Sorry, I'm not very familiar with bash scripts. Can you explain what `set -e` does?

boswelja:
Not really related to this PR, but an update on Realm on win-arm64
I managed to get past the SQLite issues in my local environment, and now I'm getting DLLNotFound for realm-wrappers. I'm not sure why osu can't resolve these DLLs, but the project I threw together earlier doesn't have any issues.

osu.Framework.NativeLibs/runtimes/win-arm64/source/compile-ffmpeg.sh
bdach: As per [bash manual](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html):

```
Exit immediately if a pipeline (see Pipelines), which may consist of a single simple command (see Simple Commands), a list (see Lists), or a compound command (see Compound Commands) returns a non-zero status.
```

boswelja: Oh nice, that sounds like a good change then. Thanks for helping me understand!

boswelja:
Test failures on Linux? Doesn't seem related

bdach:
It is not related, I've opened https://github.com/ppy/osu-framework/issues/4662 for that.

I mostly don't know what to do with this pull myself since I have no means to check whether any of this works.

smoogipoo:
Yeah, I have no way to test this either, and GitHub doesn't offer a windows-arm actions runner.

peppy:
I don't think testing is required. Just need someone to go through and verify each built binary (ie. copy `bass` ones on top, and probably update to 2.6, and rebuild ffmpeg or check it matches the compile script).

boswelja:
I've got hardware to test on, so if you need me to check anything in particular I can give it a go. I did run through all VisualTests when opening this PR, and I've rerun audio tests after updating BASS etc today, and all seem to have passed.
I'd still like to validate audio quality though, I haven't had a chance to connect some decent headphones and the Surface Pro X speakers aren't very good.

boswelja:
Alright, I tried it with some decent headphones and there's definitely some audio issues. I think they're related to BASS, in which case I'll report back to the team over there, but I'd like to get someone else's thoughts to be sure.
Here's a link to some short samples: https://1drv.ms/u/s!Aofd-xCrxwuMgq5nHtawpjcS3BeKRA?e=P31rcP

bdach:
What's the last commit ("FFmpeg binary rebuild") about? It's hard to gather any context from a commit with only blobs in it and a one-liner commit message, and no follow-up comment or anything. Is it just a version bump or something?

boswelja:
For some reason rebuilding resulted in the binaries being changed. I'm not sure what caused the change, but it seems to have the same output consistently now. I figured it would be best to upload them in case there was some issue with flags when I was working on the compile script, if that makes sense.
While I'm here, I've brought up the audio output issue on the BASS forums in the hopes of getting that resolved.

boswelja:
Just a quick update on the audio quality issues, it seems to be resolved *before applying effects*. As soon as any effects get added to tracks or samples, the audio ends up distorted again. Will drop a message here when we get that resolved.

bdach:
I've checked the bass binaries and they check out against the links provided in commit messages. Haven't got round to running the FFmpeg script due to setup involved.

(As a side note the build script attached isn't marked executable which is a bit of an annoyance. `git` stores granted `+x` permissions in the index, so this *will* show up as a committable change and work portably.)

boswelja:
Quick update on this, Realm now has proper ARM64 support. Are there any other blockers here?

bdach:
Personally my primary problems are: not being able to test this (due to not having the requisite hardware) and not being able to verify the FFmpeg blobs. I *could* run the script provided but I think there's a high likelihood that it is going to result in different output because I don't know that FFmpeg does reproducible builds.

What code *is* here looks mostly sane to me, but the above concerns stop me from going ahead and merging this myself.

boswelja:
I see. For testing, the cheapest platform to run Windows on ARM natively is probably a Raspberry Pi. I've also heard of some hacks to get WoA running on MY MacBooks which might be more helpful. If there's anything else I can provide, let me know.

As for FFMPEG binaries, I'm happy for you (or anyone else for that matter) to run the script and commit their generated version. Another option could be to set up a CI workflow that executes the scripts itself, though not sure if that's any more practical than just committing scripts + binaries.

smoogipoo:
Would prefer a CI script do the building for us. I was about to mention this when the macOS libs were merged recently.

boswelja:
Should CI scripts be a separate PR? Wondering if (after testing) we remove the compiled binaries from this PR before merging.

smoogipoo:
Separate PR, yes. Best would be to run the script on a schedule and automatically create a PR with the required binaries. Our SDL2-CS fork already does this, for a point of reference: https://github.com/ppy/SDL2-CS/blob/master/.github/workflows/build.yml

boswelja:
Have dropped FFMPEG binaries for now and created #5093 to track CI workflow for building

osu.Framework.NativeLibs/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Ignore FFMPEG source directories
+**/ffmpeg-*/
```
bdach, line 2: This change still needs removing I think.

peppy (approved):

osu.Framework.NativeLibs/runtimes/win-arm64/source/compile-ffmpeg.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
bdach, line 1: This script is extraneous at this point but maybe let's leave it in for the sake of the future (could be used as a blueprint for building on CI).

bdach (approved):

PR #4639, opened 2021-07-28
Propagate button events in a certain order

### LINKED ISSUE #3992: KeyBindingContainer (somehow) needs to intertwine OnPressed events with standard input events
Currently when using a `KeyBindingContainer`, `OnPressed` / `OnReleased` are fired after the `OnKeyDown` / `OnKeyUp` events have already been handled by the full hierarchy. This can lead to scenarios where components that use `OnKeyDown` may receive and block an input event that could have been handled by a higher (in draw depth) component which is expected to handle the key via a key binding mapped to `OnPressed`.

Not sure how simple and feasible this is but it's a recurring issue that we experience osu!-side, usually worked around by making non-`KeyBinding` components selectively ignore input when modifiers or other important keys are pressed.

UselessToucan:
resolves #3992 

The idea is to categorise input queue content and propagate input events in the following order:
1. Focused drawable
2. Key binding containers
3. Other drawables

Also, there is an `All` group for edge cases like building `ButtonDownInputQueue` when the event was not handled.

### DIFF
--- a/osu.Framework/Input/ButtonEventManager.cs
+++ b/osu.Framework/Input/ButtonEventManager.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Linq;
 using JetBrains.Annotations;
 using osu.Framework.Graphics;
+using osu.Framework.Input.Bindings;
 using osu.Framework.Input.Events;
 using osu.Framework.Input.StateChanges;
 using osu.Framework.Input.States;
@@ -34,12 +35,12 @@ namespace osu.Framework.Input
         /// The input queue.
         /// </summary>
         [NotNull]
-        protected IEnumerable<Drawable> InputQueue => GetInputQueue.Invoke() ?? Enumerable.Empty<Drawable>();
+        protected ReadOnlyInputQueue InputQueue => GetInputQueue.Invoke() ?? new InputQueue();
 
         /// <summary>
         /// A function to retrieve the input queue.
         /// </summary>
-        internal Func<IEnumerable<Drawable>> GetInputQueue;
+        internal Func<ReadOnlyInputQueue> GetInputQueue;
 
         protected ButtonEventManager(TButton button)
         {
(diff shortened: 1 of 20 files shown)


smoogipoo (changes requested):
What's the goal of this change? Why should the MBEM know about the keybinding containers/propagate to keybinding containers specifically? Why should it know about the focused drawable specifically rather than the focused drawable just being injected at the start of the queue?

Also, I don't think it's correct for KBCs to always handle input before the rest of the hierarchy - this is only true for prioritised ones like in o!f.

Note that the way input is done in relation to KBC and the whole "keybindings are first propagated through the entire hierarchy before normal events" is already inadequate (see: https://github.com/ppy/osu-framework/issues/3992). Edit: Since I see that you mentioned the same PR I mentioned in the description, I'm not seeing how this resolves it at all? It just further solidifies whatever structure we already have.

osu.Framework/Input/ButtonEventManager.cs
```diff
@@ -106,6 +113,40 @@ private void handleButtonUp(InputState state)
...
+        /// Triggers events on drawables in <paramref name="inputQueue"/> until it is handled.
+        /// </summary>
+        /// <param name="inputQueue">The input queue.</param>
+        /// <param name="e">The event.</param>
+        /// <returns>The drawable which handled the event or null if none.</returns>
+        protected Drawable PropagateButtonEvent(ReadOnlyInputQueue inputQueue, UIEvent e)
+        {
+            return PropagateButtonEvent(inputQueue.GetFocusedDrawable(), e)
+                   ?? PropagateButtonEvent(inputQueue.KeyBingingContainers.ToList(), e)
+                   ?? PropagateButtonEvent(inputQueue.Regular.Where(drawable => !(drawable is KeyBindingContainer) && !drawable.HasFocus).ToList(), e);
```
smoogipoo, line 126: Specifically... This isn't "intertwining" input as far as I can tell. It's still two separate passes over the draw hierarchy, which is not what #3992 wants.

PR #4643, opened 2021-07-30
Add audio mixer visualiser (Ctrl+F9)

smoogipoo:
Reimplements the visualiser which was removed in https://github.com/ppy/osu-framework/pull/4291

- [x] Depends on https://github.com/ppy/osu-framework/pull/4291

Very basic - I think it'd be useful to display the channel/mixer names in the future, but should be enough for now since we'll only have few channels/mixers.


osu.Framework/Input/FrameworkActionContainer.cs
```diff
@@ -18,7 +18,8 @@ internal class FrameworkActionContainer : KeyBindingContainer<FrameworkAction>
             new KeyBinding(new[] { InputKey.Control, InputKey.F11 }, FrameworkAction.CycleFrameStatistics),
             new KeyBinding(new[] { InputKey.Control, InputKey.F10 }, FrameworkAction.ToggleLogOverlay),
             new KeyBinding(new[] { InputKey.Alt, InputKey.Enter }, FrameworkAction.ToggleFullscreen),
-            new KeyBinding(new[] { InputKey.F11 }, FrameworkAction.ToggleFullscreen)
+            new KeyBinding(new[] { InputKey.F11 }, FrameworkAction.ToggleFullscreen),
+            new KeyBinding(new[] { InputKey.Control, InputKey.F9 }, FrameworkAction.ToggleAudioMixerVisualiser),
```
peppy, line 22: Any reason for avoiding `Ctrl+F4`?

smoogipoo: No reason, this is what it initially was so I assumed it was suggested to be this value somewhere else. F4 sounds good.

bdach: Ctrl-F4 is `DocumentClose` on windows, fwiw.

peppy: Surprised we have that, even though it's probably never going to be used. I guess we should avoid F4 in that case. And maybe F5 as well - if we're supporting document close then we probably will end up supporting refresh via F5?

peppy: Let's just keep it as F9.

bdach: F4 will be used game-side for closing chat tabs. It's not that rare of a binding, on a quick check Firefox/Chrome/VS Code support it on windows. I think rider also does.

peppy (dismissed):

peppy (approved):

Susko3:
This should be added to the wiki at https://github.com/ppy/osu-framework/wiki/Framework-Key-Bindings

peppy:
I've added it

PR #4646, opened 2021-07-31
Implement shader for SaturationValueSelector

EVAST9919:
In this pr I tried to fix the problem mentioned [here](https://github.com/ppy/osu/pull/13714#discussion_r661826143) by using shader instead of 3 boxes on top of each other. The final result is indistinguishable (as far as I can tell), however masking can be applied correctly now.
|master|pr|
|---|---|
|old|new|

osu-side (needs a pr, removing `MaskingSmoothness = 0`)
|master|master without fix|osu-side pr|
|---|---|---|
|picker-old|picker-old-aa|picker-new|


osu.Framework/Resources/Shaders/sh_SaturationSelectorBackground.fs
```diff
@@ -0,0 +1,13 @@
+#include "sh_Utils.h"
+
+varying mediump vec2 v_TexCoord;
+varying mediump vec4 v_TexRect;
+
+uniform mediump float hueValue;
+
+void main(void)
+{
+    vec2 resolution = vec2(v_TexRect[2] - v_TexRect[0], v_TexRect[3] - v_TexRect[0]);
```
bdach, line 10: Pretty sure this is wrong. It's not a problem because the only usage presumably has `v_TexRect[0] == v_TexRect[1]` but this should be

```suggestion
    vec2 resolution = vec2(v_TexRect[2] - v_TexRect[0], v_TexRect[3] - v_TexRect[1]);
```

The four items are left/top/right/bottom coordinates, in that order:

https://github.com/ppy/osu-framework/blob/5f59a64f6795d4dbdfbffe86977baefa6c18d9ea/osu.Framework/Graphics/OpenGL/Textures/TextureGLSingle.cs#L326

Also goes for the other shader, obviously.

osu.Framework/Resources/Shaders/sh_SaturationSelectorBackground.fs
```diff
@@ -0,0 +1,13 @@
...
+    vec2 pixelPos = vec2(v_TexCoord.x/resolution.x, v_TexCoord.y/resolution.y);
```
bdach, line 11:
```suggestion
    vec2 pixelPos = v_TexCoord / resolution;
```

osu.Framework/Graphics/UserInterface/HSVColourPicker_SaturationValueSelector.cs
```diff
@@ -221,6 +204,64 @@ protected abstract class Marker : CompositeDrawable
...
             }
+
+            private class SaturationBox : Box, ITexturedShaderDrawable
+            {
+                public new IShader TextureShader { get; private set; }
+                public new IShader RoundedTextureShader { get; private set; }
+
+                private float hueValue;
+
+                public float HueValue
```
bdach, line 215: probably could just name this `Hue` to avoid confusion with the HSV colour model value

bdach (changes requested):

bdach:
Going to request a follow-up review on the drawnode stuff as I'm not overly familiar but looks good to me, generally.

bdach (dismissed):

smoogipoo (approved):
lgtm

PR #4649, opened 2021-08-03
Rename `Drawable.Click` to `Drawable.TriggerClick` to avoid naming conflict

peppy:
This is primarily to avoid a naming conflict with `ManualInputManager.Click`, but also better aligns with the fact that it's triggering an event (via `TriggerEvent`) internally.

For additional safety, `ManualInputManager` will also throw on usage of this method (as it is a common gotcha mistake when trying to create tests). I haven't bothered with the obsoleted method here since the user should see an obsolescence notice in such cases.


frenzibyte (approved):
I don't really like the existence of `TriggerClick` on drawables themselves, it's a hacky way to perform the action of a click while there's no actual mouse button clicking a drawable, which may potentially cause unexpected issues due to that fact alone.

I guess it'll have to continue existing for the time being.

bdach (approved):

Issue #3349, opened 2020-03-09
Opening draw visualiser and global statistics and clicking randomly hits assert


peppy:
```
[runtime] 2020-03-09 10:10:45 [error]: NUnit.Framework.AssertionException: Cyclic invocation of computeCharacters()!:
[runtime] 2020-03-09 10:10:45 [error]: at osu.Framework.Logging.ThrowingTraceListener.Fail(String message1, String message2) in /Users/dean/Projects/osu-framework/osu.Framework/Logging/ThrowingTraceListener.cs:line 25
[runtime] 2020-03-09 10:10:45 [error]: at System.Diagnostics.TraceInternal.Fail(String message, String detailMessage)
[runtime] 2020-03-09 10:10:45 [error]: at System.Diagnostics.TraceInternal.TraceProvider.Fail(String message, String detailMessage)
[runtime] 2020-03-09 10:10:45 [error]: at osu.Framework.Graphics.Sprites.SpriteText.computeCharacters() in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Sprites/SpriteText.cs:line 451
[runtime] 2020-03-09 10:10:45 [error]: at osu.Framework.Graphics.Sprites.SpriteText.get_Size() in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Sprites/SpriteText.cs:line 364
[runtime] 2020-03-09 10:10:45 [error]: at osu.Framework.Graphics.Drawable.get_DrawSize() in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Drawable.cs:line 791
[runtime] 2020-03-09 10:10:45 [error]: at osu.Framework.Graphics.Drawable.get_LayoutSize() in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Drawable.cs:line 829
[runtime] 2020-03-09 10:10:45 [error]: at osu.Framework.Graphics.Drawable.get_LayoutRectangle() in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Drawable.cs:line 852
```

peppy:
As this requires very specific usage to reproduce (and only happens in debug) lowering the priority for now.

PR #4651, opened 2021-08-03
Disable SpriteText assertion on cyclic computation

smoogipoo:
Resolves https://github.com/ppy/osu-framework/issues/3349

Autosize disallows cyclic invocations itself, so we are safe to incur an extra computation until autosize is split into two passes.


bdach:
Inspections to resolve here (the variable also has to go away for now).

smoogipoo:
Removed the variable, but left the assert as is.

peppy (approved):

PR #4652, opened 2021-08-03
Add the ability to actuate `PlatformAction`s via `ManualInputManager`

peppy:
Not 100% on the naming of the method (should it instead be `Key` as an
overload?).

### DIFF
--- a/osu.Framework/Testing/Input/ManualInputManager.cs
+++ b/osu.Framework/Testing/Input/ManualInputManager.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Linq;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Input;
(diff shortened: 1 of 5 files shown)


osu.Framework/Testing/Input/ManualInputManager.cs
```diff
@@ -91,13 +92,28 @@ public void Input(IInput input)
...
         {
             PressKey(key);
             ReleaseKey(key);
         }
 
+        /// <summary>
+        /// Press and release the keys in the specified <see cref="PlatformAction"/>.
+        /// </summary>
+        /// <param name="action">The platform action to actuate.</param>
+        public void Keys(PlatformAction action)
```
smoogipoo, line 106: Any reason for not `Action()`

peppy: I'm not against that, but would want to wait for a third opinion (@bdach ?)

frenzibyte:
There are currently tests that actuate platform actions by manually calling `OnPressed` on them, should they be updated to use the new method here or?

peppy:
Yeah, for sure. I probably missed those. There's quite a few that use a local `PlatformActionContainer` but I decided not to touch them because they are isolating the component they are testing from many of the same components on the screen.

osu.Framework/Testing/Input/ManualInputManager.cs
bdach: I don't know about `Action()`, seems a bit generic imo. I'd be more fine with `PlatformAction()`.

peppy:
@frenzibyte are you gonna push a change for that? is not, please tell me which tests you found which do this.

osu.Framework/Testing/Input/ManualInputManager.cs
bdach: Is this going to be changing or stay as-is? I'm not too fussed either way, just not sure on what the conclusion here is.

peppy: If we're changing it I'd prefer `PlatformAction`. Just worried it may be a bit hidden from a discoverability perspective.

smoogipoo (dismissed):

osu.Framework/Testing/Input/ManualInputManager.cs
smoogipoo: Let's just go with it as is. My original point for `Action()` was to _be_ generic, because I imagine a case where we'd have an overload for global actions. But we'll see how that pans out - osu! only uses `OsuManualInputManagerTestScene` instead of deriving `ManualInputManager` for that purpose.

smoogipoo (approved):

peppy:
I did try putting this method in `ManualInputManagerTestScene` scene itself, but this ended up feeling like the more correct place.

smoogipoo:
That doesn't really work though, since osu! doesn't inherit `ManualInputManagerTestScene`. It's a bit disjoint...

peppy:
Makes sense. Good thing I bailed on that idea before attempting to use it with osu!.

Issue #4504, opened 2021-06-10
Crash when using PublishSingleFile


i3ym:
When publishing an application using PublishSingleFile=true it crashes when trying to load osu framework.

```
> dotnet publish -c Release -r linux-x64 /p:PublishSingleFile=true
> ./bin/Release/net5.0/linux-x64/publish/test
Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'path1')
   at System.IO.Path.Combine(String path1, String path2)
   at osu.Framework.Platform.DesktopGameHost.GetDefaultGameStorage()
   at osu.Framework.Platform.GameHost.Run(Game game)
```

Looks like a problem is in the DebugUtils.GetEntryPath method which gets Assembly.Location and it [does not work](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/il3000) in single-file apps, at least in net5

bdach:
While this failure is probably the most blocking one, I'm not super sure we're able to fully support single-file without significant development effort, which I don't think is likely to happen anytime soon. Aside from the failure above, the following probably need to be checked:

https://github.com/ppy/osu-framework/blob/07945da4c031a76e0fea87184ca721f4a3d1d146/osu.Framework/Graphics/Video/VideoDecoder.cs#L547
https://github.com/ppy/osu-framework/blob/07945da4c031a76e0fea87184ca721f4a3d1d146/osu.Framework/IO/Stores/DllResourceStore.cs#L20
https://github.com/ppy/osu-framework/blob/99786238a02e0d6b69da86dd52e5506ee8ec0566/osu.Framework/RuntimeInfo.cs#L19-L28
https://github.com/ppy/osu-framework/blob/41f8306e8788a79ab5e05c609bad2da66856e4dd/osu.Framework/Platform/GameHost.cs#L217-L219

peppy:
Wiling to say that if you're trying to use this feature it'd be appreciated if you can fix/submit any required changes to make it work, because it's out of scope for us in the short term.

jai-x:
Closed by https://github.com/ppy/osu-framework/pull/4653 I think?

frenzibyte:
Yep, any issues can be reported separately, the one mentioned in the OP should be resolved by #4653.

PR #4653, opened 2021-08-03
Fix startup crash when publishing game in SingleFile and Portable mode

jai-x:
Creating a gamehost in portable mode will use the startup path of the
executable as the storage basepath. This basepath is resolved in
`DebugUtils` via `Assembly.GetEntryAssembly()`. When publishing in
SingleFile mode `Assembly.GetEntryAssembly()` will always return an
empty string which causes a crash on startup.

It is recommended to use `AppContext.BaseDirectory` instead.

See: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/il3000


peppy:
This will require extensive testing on iOS/android/desktop against osu! deploys. @jai-x is this a high priority for your use case?

jai-x:
From searching the codebase framework side, it seems `RuntimeInfo.StartupDirectory` is only ever used in the `DesktopGameHost` class here so I'm not sure mobile builds will be affected at all.
https://github.com/ppy/osu-framework/blob/9f1ab8b672cd2eddb1a279b121ca956eb72a6135/osu.Framework/Platform/DesktopGameHost.cs#L34-L35

And it should already have test coverage by these tests:
https://github.com/ppy/osu-framework/blob/master/osu.Framework.Tests/Platform/PortableInstallationTest.cs 

Obviously these tests aren't the same as publishing a release in single file mode, so I can build lazer locally to test this on Linux and Windows, but I don't have a Mac to test this with.

As for priority, it's not really high priority since I'm not making any releases yet, but I do prefer releasing in single file where possible.

frenzibyte:
osu!lazer-side it's primarily required for locating built-in ruleset DLLs. I've checked `AppContext.BaseDirectory` and it's suprisingly working in both normal usage, and nUnit test context by both `dotnet test` and rider unit test pane (still not sure how that goes with nuget packages however, as I had struggled to get that working)

But, 
 1. There's still `GetFrameworkAssemblyPath()` which may need reassessing whether it's required at all at this point.
 2. `DebugUtils.GetEntryPath()` should now be obsoleted in favour of using `AppContext.BaseDirectory` for accurate entry path, I believe.

bdach:
Related: #4504

jai-x:
Is there any specific testing I can do as evidence for this small change?

Given the way this method is used lazer side, I'm assuming this mainly affects Dll Ruleset, but is there anything I should also test framework side?

peppy:
Should be fine for now, we'll just need to test from our end for verification.

frenzibyte:
osu!lazer will likely never be able to use single-file/portable mode primarily due to built-in ruleset loading.

I've managed to test this with a local framework nupkg and it's also working fine with `AppContext.BaseDirectory`, so that should confirm this change to not affect o!f or osu! on desktop platforms. (pending mobile testing, although iOS is broken beyond repair at compile time currently)

As for single file support, other than the [areas](https://github.com/ppy/osu-framework/issues/4504#issuecomment-858597371) mentioned in the linked issue thread which may require updating, it seems to work fine.

bdach:
> osu!lazer will likely never be able to use single-file/portable mode mainly due to built-in ruleset loading.

I'm not sure where this assertion is coming from? It's not like publishing in portable mode entirely prevents dynamic assembly loading, or is there something I'm not aware of?

But otherwise - yes, testing the places that I have already identified as potentially problematic in the issue thread would be good.

frenzibyte:
Never was a bit over the top, but I meant that in normal cases it is not possible to retrieve the ruleset DLLs, I could be wrong though, it may be possible with a set of changes.

frenzibyte:
Yeah I was suggesting to obsolete it, and link to `AppContext.BaseDirectory` as the correct path.

peppy (approved):

peppy:
Have tested with no issues here.

smoogipoo (approved):

PR #4657, opened 2021-08-05
Interpolate between colours in linear space

frenzibyte:
Originally proposed by @bdach in https://github.com/ppy/osu/pull/14119#discussion_r682129668 for a specific case osu!-side, but since it's correct for colours to be interpolated in the linear space in general, I've applied that here in `Interpolation.ValueAt(Color4)` instead.

| | `TestSceneColourInterpolation` |
|-|---------------------------------|
| master | CleanShot 2021-08-05 at 19 44 45 |
|  PR | CleanShot 2021-08-05 at 19 42 59 |


bdach:
May be a good opportunity to add testing for this so that it doesn't regress if colours are ever touched again (like if the sRGB/linear stuff is ever attempted to be cleaned up).

frenzibyte:
Yeah just thought about it and now writing one.

bdach:
Not sure about having *this* many boxes in the test scene (may give people wrong ideas on how to do gradients) but other than that this seems correct.

Going to leave [this link](https://blog.johnnovak.net/2016/09/21/what-every-coder-should-know-about-gamma/#gradients) again, for cross-referencing purposes. Note that:

* the figure therein visually matches the OP
* the post states that CSS is doing gradients *wrong* by default (interpolates in sRGB space)
* [this link](https://lists.w3.org/Archives/Public/www-style/2012Jan/0607.html) to a w3c mailing list message which states that the midpoint between `rgb(0,0,0)` and `rgb(255,255,255)` is roughly `rgb(190,190,190)` - and in the test scene above we have a gradient from `rgb(255,0,0)` and `rgb(0,0,255)`, which after interpolation in linear space should _and does_ produce roughly `rgb(190,0,190)` (and _not_ `rgb(127,0,127)`).

bdach (dismissed):

smoogipoo:
I don't know about this. `Color4`/`Colour4` are supposed to be "space-less" - neither SRGB _nor_ RGB, yet this PR is saying that Color4 is _always_ SRGB.

Imo it makes more sense if this was a separate interpolation method for `SRGBColour`.

peppy:
> I don't know about this. `Color4`/`Colour4` are supposed to be "space-less" - neither SRGB _nor_ RGB, yet this PR is saying that Color4 is _always_ SRGB.
> 
> 
> 
> Imo it makes more sense if this was a separate interpolation method for `SRGBColour`.

In the interest of not rewriting everything I'd advise against this direction for now (it seems like you're proposing every colour specification across osu! should be changed from `Color4` to `SRGBColour`).

While it may be more correct it's going to lead to days of work to change that.

I *do* think that this interpolation function should mention that it assumes the input colours are sRGB in its xmldoc.

smoogipoo:
Yes, we should be using `SRGBColour` where sRGB colours are involved, but we don't need to change everything immediately. I.e. we only really need to change @frenzibyte's ongoing PR to take `SRGBColour` parameters.

I'd go the opposite way to say that merging this in locks us into a rewrite. I want to point out, specifically, that interpolating in sRGB space is wrong but is the [industry standard](https://erikmcclure.com/blog/everyone-does-srgb-wrong-because/). There's no magic bullet for fixing all the issues once we start adding hacks that re-sRGB already sRGB'd colours for interpolation purposes in order to match CSS/SVG/Photoshop/whatever else.

At least with my proposed way, `Color4` remains space-less and this special case of correct sRGB interpolation remains as an explicit operation.

smoogipoo:
Following up from voice comms: let's go with making this change to the interpolation func and have gamma-correct interpolation in all cases. A few cases like colour-fades will have incorrect (compared to web) intermediate values, but the final values are the same so /shrug. Gradients are already incorrect compared to web, where we already do gamma-correct interpolation anyway.

We discussed removing SRGBColour code-wide, or at least making internal, but it's a bit of a breaking change. May be best to leave it for now.

smoogipoo (approved):

peppy:
I've added a sprite comparison from d3 output, just to give us some degree of "ground truth" rather than comparing against ourselves.

@smoogipoo please confirm it looks okay to you.

peppy (approved):

smoogipoo (dismissed):
lgtm

smoogipoo (approved):

PR #4658, opened 2021-08-05
Update bass packages and sync iOS Microsoft.Diagnostics.Runtime reference

peppy:
- Update `Microsoft.Diagnostics.Runtime` reference (based on advice from https://github.com/xamarin/xamarin-macios/issues/12353)
- Update bass packages (one android bump seemingly was missed)

Note that I still can't compile on iOS. Multiple issues still remain. Just bringing everything up to date so we have a fresh slate to work against.


peppy:
@bdach testing on android would be appreciated, as a seemingly important bump was missed previously (I guess it worked because of local framework reference - transitive deps only tend to fail when nuget packages are involved). My hypothesis would be that the osu! side master will currently *not* work on android as it stands, so if you could also test that it would be appreciated.

bdach:
Ok, so I suppose that to test this properly I would have to make a local framework package from this branch?

I'll see what I can do today.

peppy:
- test latest `master` on osu! side as it stands to confirm the fail case
- test standard local-framework osu! to confirm i haven't regressed anything (no need for local packaging)

bdach:
Yeah, okay. I only suggested doing a local package because if there is an issue that only manifests with real nuget packages then testing this branch with a local framework did not seem any better or more informative than testing the *original* branch. But I suppose if there is no version discrepancy anymore that doesn't really matter.

peppy:
Yeah, you're welcome to test with a local package too, that would confirm the fix as well.

bdach:
As proposed above I've tested osu!-side `master` on android as it is right now, as well as osu!-side `master` with a local framework checkout of this branch. Both seem to work fine (launch, are able to download tracks from direct, enter gameplay and complete a map in multi).

Because there was no issue osu!-side on `master` without any changes I abstained from trying to make a local framework package with this PR's changes as there didn't seem to be much reason left to try doing that if it looked to work anyway.

osu.Framework.Android.props
```diff
@@ -55,7 +55,7 @@
     <Reference Include="Java.Interop" />
   </ItemGroup>
   <ItemGroup Label="Package References">
-    <PackageReference Include="ManagedBass" Version="2.0.4" />
+    <PackageReference Include="ManagedBass" Version="3.1.3-alpha" />
     <PackageReference Include="ManagedBass.Fx" Version="2.0.1" />
```
bdach, line 59: Although now that I look at this diff, this package isn't bumped here while it's bumped to 3.0.0 everywhere else... Going to re-test with that bumped.

peppy: the point release bump in theory shouldn’t do much, but still appreciated.

i guess this means we at very least don’t need the local included an inadroid props

bdach: Changed version of ManagedBass.Fx to 3.0.0 in android props and still works fine. Is it OK to push that change, or should I try removing entirely and seeing what happens?

peppy: I feel like if possible we'd like to remove that whole section from the props file...

frenzibyte:
The issue with `Microsoft.Diagnostics.Runtime` still remains (after rewinding back to before bassmix changes). But I've managed to isolate the issue and will now report it in the upstream issue thread, in case it helps them pinpointing the main cause faster. (interestingly it doesn't require a .NET Standard library project or anything special)

smoogipoo:
Is this to be merged, or still awaiting testing?

peppy:
Best to wait for feedback from @bdach regarding requirements in props file. As already mentioned, this doesn't seem to impede building osu! so it's pretty low priority for now (is not actually fixing anything, iOS included).

bdach:
I've just tested with a local framework checkout and on a quick look removing the bass dependencies from the android props file seems to have had no negative effects. Not sure whether I should be checking with a local package to make extra sure.

peppy:
Should be good for merging now, in theory.

smoogipoo (approved):

PR #4659, opened 2021-08-05
mark public CanBeNull as deprecated

### LINKED ISSUE #3849: Resolved attribute can set "CanBeNull" as a param or field
In the `osu` repository, I see cases of `[Resolved(canBeNull: true)]` (35) and `[Resolved(CanBeNull = true)]` (50). The framework sets with only the field. Just a code quality issue and discussion to make the field private (which it can) or remove param.

YVbakker:
This will mark the currently public `CanBeNull` attribute of `ResolvedAttribute` as deprecated, as this should become private at some point in the future and this change is breaking.

closes #3849

### DIFF
--- a/osu.Framework.Tests/Dependencies/CachedAttributeTest.cs
+++ b/osu.Framework.Tests/Dependencies/CachedAttributeTest.cs
@@ -295,12 +295,12 @@ namespace osu.Framework.Tests.Dependencies
         {
         }
 
-        [Cached(Type = typeof(object))]
+        [Cached(type: typeof(object))]
         private class Provider2
         {
         }
 
-        [Cached(Type = typeof(Provider1))]
+        [Cached(type: typeof(Provider1))]
         private class Provider3 : Provider1
         {
         }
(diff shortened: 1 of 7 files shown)


frenzibyte:
Thanks for the contribution. Few points on the change:
- The way this is obsoleted does not seem 100% right, at this state you can only just mark it as `readonly` already.
- This should also apply for the other properties, all of them should only ever be set from the constructor and nowhere else.

frenzibyte (changes requested):
As above.

smoogipoo:
I dunno... This seems like a waste of time for a breaking change that achieves nothing of particular importance.

smoogipoo:
Thanks for the contribution, but I'm going to take the decision to close this for the time being.

I need to see more discussion on the topic before I'm ready to consider this sort of change, such as what the C# best practices are for attributes. You can Google around to find [resources](https://www.codemag.com/article/0307041/Defining-and-Using-Custom-Attribute-Classes-in-C) hinting at using positional arguments as always-required params and named arguments as optional params, which is the complete inverse of this PR, as an example.

PR #4660, opened 2021-08-06
Update ReSharper tools to 2021.2

frenzibyte:


.github/workflows/ci.yml
```diff
@@ -90,7 +90,7 @@ jobs:
       #   run: dotnet format --dry-run --check
 
       - name: InspectCode
-        run: dotnet jb inspectcode $(pwd)/osu-framework.Desktop.slnf --output=$(pwd)/inspectcodereport.xml --cachesDir=$(pwd)/inspectcode --verbosity=WARN
+        run: dotnet jb inspectcode $(pwd)/osu-framework.Desktop.slnf --no-build --output=$(pwd)/inspectcodereport.xml --cachesDir=$(pwd)/inspectcode --verbosity=WARN
```
peppy, line 93: Are you sure this doesn't reduce inspection coverage?

frenzibyte: Since the relevant warning was stating `--no-build` to match previous version behaviour, I've went with it, as currently [issues](https://github.com/ppy/osu-framework/runs/3258795448) arise from using the new `--build` behaviour.

smoogipoo (approved):

osu.Framework/Graphics/Textures/ArrayPoolTextureUpload.cs
```diff
@@ -23,7 +23,7 @@ public class ArrayPoolTextureUpload : ITextureUpload
         /// <param name="arrayPool">The source pool to retrieve memory from. Shared default is used if null.</param>
         public ArrayPoolTextureUpload(int width, int height, ArrayPool<Rgba32> arrayPool = null)
         {
-            data = (this.arrayPool ??= ArrayPool<Rgba32>.Shared).Rent(width * height);
+            data = (this.arrayPool = ArrayPool<Rgba32>.Shared).Rent(width * height);
```
peppy: wait WHAT?????????????????????????

peppy: Looks like we're hardly using the class, which is lucky because this is broken - both in this PR and master.

smoogipoo: Refresh the page, I fixed it in the commit just above.

peppy (approved):

smoogipoo:
This version of InspectCode` seems to be deadlocking:
- https://github.com/ppy/osu-framework/runs/3260098458?check_suite_focus=true
- https://github.com/ppy/osu-framework/runs/3260412829?check_suite_focus=true
- https://github.com/ppy/osu-framework/runs/3260728249?check_suite_focus=true

Etc... Will revert.

PR #4664, opened 2021-08-06
Fix glyph edge interpolation for non-transparent pixels

peppy:
Uploading glyphs, which are usually `rgba(255,255,255,x)` to a texture atlas initialised with `rgba(0,0,0,0)` would cause incorrect interpolation of edge pixels in the case where a glyph has a non-zero alpha value.

I attempted to fix this in multiple different ways, but the solution here is the best combination of being performant and future-proof. A faster method would be to change the initialisation colour of the texture atlas in the case of glyph stores and call it a day, but this would break any potential usages where a user has fonts with colour data.

Note that the new `initialisationColour` flow is (in the end) unused in this PR, but I've left it in place for potential future usage and clarity as to what is going on. If you want to test the performance of the alternative method (initialising with transparent white) you can do so by changing only the `initialisation_colour` constant in `TextureGLAtlas` (and commenting out the padding upload methods if you want to avoid the performance incurrence they carry).

In some quick testing, the overhead of this is negligible now that the span retrieval is outside of the loops. This turned out to be a huge performance concern (as you might expect) and is probably also the reason for https://github.com/ppy/osu-framework/blob/58eea16f51b401f4138986a80c19467d7e0cca55/osu.Framework/Graphics/OpenGL/Textures/TextureGL.cs#L153, but that's for another day.

Cross-tested with `TestSceneWrapModes` to ensure the expectations there have not been altered.

Before:

20210806 163431 (dotnet)

After:

20210806 163455 (dotnet)

Before:

20210806 163725 (dotnet)

After:

20210806 163848 (dotnet)

---

If unsure about what is going on, here's some further details to hopefully ease review:

You can visualise the edges that are being drawn/interpolated to by altering the transfer function:

```csharp
        private void transferBorderPixel(ref Rgba32 dest, Rgba32 source)
        {
            dest.R = 255;
            dest.G = 0;
            dest.B = 0;
            dest.A = 255;
        }
```

This branch:
20210806 170403 (dotnet)
20210806 170626 (dotnet)

With the `initialisation_colour` change mentioned above:
20210806 170456 (dotnet)
20210806 170545 (dotnet)

Notice that with the `initialisation_colour` change it causes less of the border/padding uploads, which may be something we should investigate in the future (potentially better performance due to the "normal" case of glyph uploads requiring less uploads? but it does also incur a small i# overhead to initialise the `Image`).

### DIFF
--- a/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteIcon.cs
+++ b/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteIcon.cs
@@ -22,6 +22,7 @@ namespace osu.Framework.Tests.Visual.Sprites
     {
         public TestSceneSpriteIcon()
         {
+            Box background;
             FillFlowContainer flow;
 
             Add(new TooltipContainer
@@ -29,7 +30,7 @@ namespace osu.Framework.Tests.Visual.Sprites
                 RelativeSizeAxes = Axes.Both,
                 Children = new Drawable[]
                 {
-                    new Box
+                    background = new Box
                     {
                         Colour = Color4.Teal,
                         RelativeSizeAxes = Axes.Both,
@@ -43,6 +44,7 @@ namespace osu.Framework.Tests.Visual.Sprites
                             Origin = Anchor.TopRight,
                             RelativeSizeAxes = Axes.X,
                             AutoSizeAxes = Axes.Y,
+                            Spacing = new Vector2(5),
                             Direction = FillDirection.Full,
                         },
                     }
@@ -72,6 +74,7 @@ namespace osu.Framework.Tests.Visual.Sprites
 
             AddStep("toggle shadows", () => flow.Children.OfType<Icon>().ForEach(i => i.SpriteIcon.Shadow = !i.SpriteIcon.Shadow));
             AddStep("change icons", () => flow.Children.OfType<Icon>().ForEach(i => i.SpriteIcon.Icon = new IconUsage((char)(i.SpriteIcon.Icon.Icon + 1))));
+            AddStep("white background", () => background.FadeColour(Color4.White, 200));
         }
 
         private class Icon : Container, IHasTooltip
--- a/osu.Framework/Graphics/OpenGL/Textures/TextureGLAtlas.cs
+++ b/osu.Framework/Graphics/OpenGL/Textures/TextureGLAtlas.cs
@@ -49,9 +49,9 @@ namespace osu.Framework.Graphics.OpenGL.Textures
 
             var data = upload.Data;
 
-            uploadCornerPadding(data, middleBounds, actualPadding);
-            uploadHorizontalPadding(data, middleBounds, actualPadding);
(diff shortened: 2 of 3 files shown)


smoogipoo:
Some text looks really bad with this.

| before | after |
| --- | --- |
| spritetext_before | spritetext_after |

For example:
- "Run all steps"
- Fat top lines on `T` characters (e.g. in the test listing)
- Pixels to the sides of characters (`e` from `testy` at size 29, `9` at size 29, `3` at size 23).

peppy:
I'll revert back the last commit, which will fix that case. It means that behaviour will change slightly, but I think it shouldn't be a huge concern (basically the alpha rule will be applied based on `wrapmode=none` rather than `alpha<255`). I had it this way initially and tried to change it as a "will fit better" rule, but seems like that can't be counted on.

smoogipoo:
Just keep in mind, perhaps you know this, but `wrapmode=none` indicates no wrapping whatsoever. It doesn't mean that the texture stops at the border, but that GL has unlimited access to the rest of the texture. Based on that, I find it a bit weird that these rules are being applied to `wrapmode=none` in the first place imo.

peppy:
That's the whole fix here - the padding was *not* applied with wrapmode none previously, which meant there were transparent black pixels there.

smoogipoo:
Doesn't that indicate that `SpriteText` simply shouldn't use `wrapmode=none`? I find this change being applied in the general case to be very weird, it means every texture added to an atlas has an extra pixel of padding everywhere, and I'm not sure if that's wrong in some cases where normal sprites are used.

@Tom94 If you have a few minutes... Do you remember if there's a reason why we didn't apply padding to `wrapmode=none`? Was it purely a performance thing, or was there more to it?

peppy:
It *is* using `wrapmode=none`. Please read through my opening post again. If still unclear let's voice since this should only take a few minutes of understanding.

smoogipoo:
And I'm saying that it shouldn't?

peppy:
If that's the case, the default would need to be changed for every sprite (I don't think we'd ever want the previous behaviour).

smoogipoo:
Yeah, I don't disagree. I mentioned that in the original implementation: https://github.com/ppy/osu-framework/pull/3511#issuecomment-625161963

peppy:
I see. Either way, check the latest commit - it should behave more in line with expectations. Changing the wrapping mode alone would not have been enough to fix the reported issue. There's still the choice of "using a=0 or using a=source.a" which is required in some way.

smoogipoo (approved):
Yep. This makes sense to me.

Tom94:
Heya, just to answer smoogi‘s question about WrapMode.None: yes, IIRC the reason it doesn’t get padding on top of the mipmap-reared padding is purely performance-related. In all cases where good border interpolation is needed, WrapMode.ClampToEdge should do the trick.

I very much like the solution of an initialization color for optimizing special cases like fonts. 👍

PR #4665, opened 2021-08-06
Enable inspections for unused method parameters

peppy:


osu.Framework/Host.cs
```diff
@@ -6,30 +6,23 @@
...
-            {
-                EnableHighResolution = true,
-                Backend = RuntimeInfo.OS == RuntimeInfo.Platform.Linux ? PlatformBackend.Default : PlatformBackend.PreferNative
-            };
-
             switch (RuntimeInfo.OS)
             {
                 case RuntimeInfo.Platform.macOS:
-                    return new MacOSGameHost(gameName, bindIPC, toolkitOptions, portableInstallation);
+                    return new MacOSGameHost(gameName, bindIPC);
```
smoogipoo, line 19: Will/should be caught by CI, but removing the portable parameter seems unintended?

osu.Framework/Host.cs
```diff
@@ -6,30 +6,23 @@
...
-            var toolkitOptions = new ToolkitOptions
-            {
-                EnableHighResolution = true,
-                Backend = RuntimeInfo.OS == RuntimeInfo.Platform.Linux ? PlatformBackend.Default : PlatformBackend.PreferNative
-            };
-
             switch (RuntimeInfo.OS)
             {
                 case RuntimeInfo.Platform.macOS:
-                    return new MacOSGameHost(gameName, bindIPC, toolkitOptions, portableInstallation);
```
ekrctb, line 26: `portableInstallation` parameter is unused?

smoogipoo:
It looks like resharper CLI really didn't like the latest change :/

smoogipoo (approved):

PR #4669, opened 2021-08-07
Fix joystick/controller axes potentially sticking at non-zero values

peppy:
The `Precision.AlmostEqual` check meant that a transition from `0.0000302` to `0` would never fire an event, and as such all components would believe there is still an input to the physical device. This is an edge case where we should actually be handling true zero and allowing the event to still fire.

Note importantly that this can only currently happen when the deadzone is set to 0, rather than its current value. As such this incorrect behaviour will not trigger in realistic usages. We may allow tweaking the deadzone in the future though, at which point this *will* be required.

Incorrect behaviour:

https://user-images.githubusercontent.com/191335/128610877-6b586d6f-9de8-45aa-9e42-2f7f1f465d34.mp4

### DIFF
--- a/osu.Framework/Input/StateChanges/JoystickAxisInput.cs
+++ b/osu.Framework/Input/StateChanges/JoystickAxisInput.cs
@@ -35,13 +35,14 @@ namespace osu.Framework.Input.StateChanges
         {
             foreach (var a in Axes)
             {
-                // Not enough movement, don't fire event
-                if (Precision.AlmostEquals(state.Joystick.AxesValues[(int)a.Source], a.Value))
+                float oldValue = state.Joystick.AxesValues[(int)a.Source];
+
+                // Not enough movement, don't fire event (unless returning to zero).
+                if (oldValue == a.Value || (a.Value != 0 && Precision.AlmostEquals(oldValue, a.Value)))
                     continue;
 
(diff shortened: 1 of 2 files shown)


osu.Framework/Input/StateChanges/JoystickAxisInput.cs
```diff
@@ -35,13 +35,14 @@ public void Apply(InputState state, IInputStateChangeHandler handler)
...
+                float oldValue = state.Joystick.AxesValues[(int)a.Source];
+
+                // Not enough movement, don't fire event (unless returning to zero).
+                if (oldValue == a.Value || (a.Value != 0 && Precision.AlmostEquals(oldValue, a.Value)))
                     continue;
 
-                var lastValue = state.Joystick.AxesValues[(int)a.Source];
                 state.Joystick.AxesValues[(int)a.Source] = a.Value;
-                handler.HandleInputStateChange(new JoystickAxisChangeEvent(state, this, a, lastValue));
+                handler.HandleInputStateChange(new JoystickAxisChangeEvent(state, this, a, oldValue));
```
frenzibyte, line 45: Could consider rounding to 7 digits or so alternatively, avoiding the need to define special cases.

frenzibyte (approved):
Something for later if deemed required I guess.

smoogipoo (approved):

PR #4672, opened 2021-08-08
Migrate track BASS sync usages to mixer infrastructure

### LINKED ISSUE #4670: Ogg Tracks no longer loop correctly
Broken on: `2021.805.0` and `2021.807.0`
Last working correctly `2021.804.0`

Creating a `Track` from a `.ogg` file and playing it with `Looping = true` has a noticeable skip when it reaches the loop point.

The `.ogg` file I noticed this issue with (warning somewhat loud):
[test_ogg.zip](https://github.com/ppy/osu-framework/files/6949724/test_ogg.zip)

### LINKED ISSUE #4671: Deadlock in new audio code
Looks like bass getting in a bad state, but likely due to something we're doing:

rider64_2021-08-08_13-55-42

bdach:
* Fixes #4670 (mostly)
* Fixes #4671

PRing for comment / CI verification. Tests passed locally.

As stated in the issues above, there is a dedicated mixer function for syncs and the above two issues were caused by not using it. Caveats:

* The syncs are manually unregistered and re-registered on each mixer change. The first is not necessarily required, but I decided to do it for safety, as [ManagedBass internally tracks callbacks](https://github.com/ManagedBass/ManagedBass/blob/0d46c12a8a9bdf8ceffdf60950cbb905cd5719b7/src/AddOns/BassMix/BassMix.cs#L773-L784), so I wasn't entirely comfortable with reusing the callbacks or leaving their unregistration to BASS.
* The mixtime stop sync is replaced by a standard sync. This is intentional, because due to mixer playback buffering the mixtime sync was causing loops to audibly loop back to a later time instant than intended (causing #4670). Not using a mixtime sync means that now there may be a gap in the loop, but this is mostly a matter of choosing your own poison at this point, if you will.

This *may* also do something for https://github.com/ppy/osu/issues/14162 but it is unclear as that issue does not have many details.

### DIFF
--- a/osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
+++ b/osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
@@ -202,23 +202,23 @@ namespace osu.Framework.Audio.Mixing.Bass
             => BassMix.ChannelGetData(channel.Handle, buffer, length);
 
         /// <summary>
-        /// Sets up a synchronizer on a mixer source channel.
+        /// Sets up a synchroniser on a mixer source channel.
         /// </summary>
         /// <remarks>See: <see cref="ManagedBass.Mix.BassMix.ChannelSetSync(int, SyncFlags, long, SyncProcedure, IntPtr)"/>.</remarks>
-        /// <param name="channel">The <see cref="IBassAudioChannel"/> to set up the synchronizer for.</param>
+        /// <param name="channel">The <see cref="IBassAudioChannel"/> to set up the synchroniser for.</param>
         /// <param name="type">The type of sync.</param>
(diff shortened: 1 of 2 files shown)


frenzibyte (commented):
The track looping remains to be correct on testing with osu!lazer, so it seems fine using standard sync for both cases.

Though I have had checked again and it appears https://github.com/ppy/osu/issues/14162 is still not resolved yet, I'll further investigate that as it appears to be mostly spectator-specific.

peppy:
In my testing this looks to work as expected. I know @smoogipoo will want to review this but due to the severity of this bug I'm going to go ahead with this as-is to get a hotfix release out.

peppy (approved):

peppy (approved):

PR #4673, opened 2021-08-08
Improve logic and reduce allocations of `FillFlowContainer`s

Flutterish:
This PR improves the direction handling logic of `FillFlowContainer`s by using "main" and "cross" axes instead of X and Y similarly to how flexboxes do it. This allows to easily implement a new `FillDirection` - `FullVertical` which acts like `Full`, except vertical first.  I believe this also makes the code more clear and readable.

Additionally, by using lists stored in the `FillFlowContainer` itself instead of dynamically allocated in `ComputeLayoutPositions` it reduces allocations to minimum. This is a comparison between the old way and the new where 500 children were created and layout was performed each time. It does not include the memory the lists themselves take up, which means that while this PR reduces allocations, it increases overall memory usage. If this is not desired, I can easily revert all new memory tradeoffs.
Before:
Before
After:
After

### DIFF
--- a/osu.Framework.Tests/Visual/Layout/TestSceneFillFlowContainer.cs
+++ b/osu.Framework.Tests/Visual/Layout/TestSceneFillFlowContainer.cs
@@ -375,6 +375,13 @@ namespace osu.Framework.Tests.Visual.Layout
             fillContainer.Spacing = new Vector2(5, 5);
         }
 
+        [FlowTestCase(FlowTestType.FullVertical)]
+        private void test4()
+        {
+            fillContainer.Direction = FillDirection.FullVertical;
+            fillContainer.Spacing = new Vector2(5, 5);
+        }
+
         private class TestSceneDropdownHeader : DropdownHeader
         {
             private readonly SpriteText label;
@@ -423,6 +430,7 @@ namespace osu.Framework.Tests.Visual.Layout
             Full,
             Horizontal,
             Vertical,
+            FullVertical
         }
     }
 }
--- a/osu.Framework/Graphics/Containers/FillFlowContainer.cs
+++ b/osu.Framework/Graphics/Containers/FillFlowContainer.cs
@@ -126,9 +126,10 @@ namespace osu.Framework.Graphics.Containers
             // The positions for each child we will return later on.
             layoutPositions.Clear();
 
-            // We need to keep track of row widths such that we can compute correct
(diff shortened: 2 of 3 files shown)


peppy:
Thanks for your contribution. Those are some promising results, for sure. I haven't checked the code yet, but with allocation/performance improvements, we tend to add benchmark coverage to the benchmarks project for reproducibility. Would be appreciated if you could do that and add a comparison of the Benchmark.net output.

frenzibyte:
While this does reduce allocations of the flow layout logic from what you've attached, in addition to the above, this is missing couple of parts to reach to a mergable state.

- Code style not matchcing with the osu!framework codebase, see [code quality workflow's report](https://github.com/ppy/osu-framework/pull/4673/checks?check_run_id=3274533610).
- Renaming to "main" and "cross" and adding new members without clear documentation on what they are and what their purpose is on the code. Use XMLDoc on each new field/property/method/struct that's not clear at first glance.

Also, and most importantly, I don't see why this whole addition and consumption of the `CrossAxes` structure and `FullVertical` direction in the flow layout code is necessary in this PR, if the purpose is to reduce allocations? I have a few suggestions regarding it to make it less complicated and easier to parse, but I realized that it is basically out of scope and should be split to its own PR.

If it holds a main part in the allocation reduction, I would say it should definitely be stated here or in the OP as I fail to understand how. This all looks like a couple of separate changes squashed together in one commit, making it hard to understand the true purpose of this PR.

bdach:
This pull should really be two. One for the optimisation + benchmarks, one for the new functionailty. They're orthogonal to each other.

smoogipoo:
Agree with splitting into two PRs. Additionally the test and CI failures are relevant.

Flutterish:
I will split this into 2 pulls then. However I am not sure how to do it "properly" as they both work on the same file. Should I pull one, wait for it to get merged and then pull the other one? If so, which one first?

Regarding bdach's comment, the "FullVertical", while yes, it is a new functionality, it is more of a byproduct of the new code being able to handle it rather than the focus of the "logic improvements".

frenzibyte:
The first should be about the allocation reduction since that's the important part of this PR. Then create another PR that adds the "logic imporvement" part, all with the points I've noted above addressed.

Flutterish:
Thank you. Closing this PR.

PR #4676, opened 2021-08-10
Flush audio mixer on critical BASS operations (seek/stop)

smoogipoo:
Fixes https://github.com/ppy/osu/issues/14184

This reverts all the changes to this test which were made in https://github.com/ppy/osu-framework/pull/4291.

It seems like the proper way is to flush the mixer, by setting its position to 0: http://www.un4seen.com/forum/?topic=6895.msg46426#msg46426
I've applied this to `ChannelSetPosition()` and to `ChannelStop()` (already had the flush) and `ChannelPause()`. The latter of which seems to be more required for stopping playback upon 0 frequency for `TestZeroFrequencyHandling`.

Tested over 100 times locally with no test failures, and can't repro the osu!-side issue after these changes.

### DIFF
--- a/osu.Framework.Tests/Audio/TrackBassTest.cs
+++ b/osu.Framework.Tests/Audio/TrackBassTest.cs
@@ -170,11 +170,11 @@ namespace osu.Framework.Tests.Audio
         public void TestRestart()
         {
             startPlaybackAt(1000);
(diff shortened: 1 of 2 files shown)


peppy (approved):

AriosJentu:
Will test it in next update, thanks

PR #4681, opened 2021-08-11
Fix localisation overhead in `SpriteText` / `LocalisedBindableString`

### LINKED ISSUE #4679: Minor memory leak with SpriteText
Each time a `SpriteText` is loaded, it calls `LocalisationManager.GetLocalisedString`, this method adds a binding to `LocalisationManager.currentParameters`. From what I can tell, this binding is never removed. This causes `LocalisationManager.currentParameters.Bindings` to grow each time a `SpriteText` is created. Even after the binding is garbage collected, the reference remains in the bindings list because of how `WeakList` works. Here is a simple TestScene that demonstrates this behavior:

```csharp
using System;
using osu.Framework.Graphics.Sprites;
using osu.Framework.Logging;

namespace osu.Framework.Tests.Visual
{
    public class TestSceneMemoryLeak : FrameworkTestScene
    {
        [Resolved] LocalisationManager localisationManager { get; set; }
        int j = 0;
        {
            for (var i = 0; i < 50; i++) Add(new SpriteText());
            Clear(); // this calls Dispose on all the SpriteText we just created
            if (j == 200)
            {
                GC.Collect();
                // Since we're adding 50 SpriteText each update and doing this every 200 updates, expect exactly 10000 growth per log
                // Note: I had to switch all these fields to internal instead of private in order to view this count
                //    you can also view this count with dotnet-dump, which is how I discoved this in the first place
                Logger.Log($"Binding count: {localisationManager.currentParameters.Bindings.list.list.Count}");
                j = 0;
            }
            j++;
            base.Update();
        }
    }
}
// Output:
//   [runtime] 2021-08-10 23:13:22 [verbose]: Binding count: 10271
//   [runtime] 2021-08-10 23:13:23 [verbose]: Binding count: 20271
//   [runtime] 2021-08-10 23:13:24 [verbose]: Binding count: 30271
//   [runtime] 2021-08-10 23:13:25 [verbose]: Binding count: 40271
//   [runtime] 2021-08-10 23:13:26 [verbose]: Binding count: 50271
//   [runtime] 2021-08-10 23:13:27 [verbose]: Binding count: 60271
//   [runtime] 2021-08-10 23:13:27 [verbose]: Binding count: 70271
```

The binding count starts at 10271 instead of 10000 presumably because there are 271 `SpriteText`s in the visual test interface.

The cost of this seems to be at least 40 bytes for each `SpriteText`. In my game, each beatmap can consist of 1000's of `SpriteText` due to the way the notes are drawn. This was enough for the memory usage to noticeably increase during testing.

I have a very limited understanding of how `Bindable`/`WeakList` work, but it seems like the solution would be to unbind this binding at some point. I'm not sure whether or not this is worth fixing. It is also possible that this is just an intended side effect of `WeakList`.

peppy:
master:

**benchmark will not complete** (performance consistently degrades enough from the `WeakList` overhead that benchmark.net will not allow it).

### with only `UnbindAll` commit (https://github.com/ppy/osu-framework/commit/a2ff29a183b4173de8229cd41521d2651a8f8fa2)

|                Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
| BenchmarkNonLocalised | 827.4 ns |  9.59 ns |  8.01 ns | 0.0591 |     - |     - |     680 B |
|    BenchmarkLocalised | 939.9 ns | 14.54 ns | 11.35 ns | 0.0668 |     - |     - |     776 B |

### also with lazy binding commit (https://github.com/ppy/osu-framework/commit/59a4b8920359143a226a457849f4c3921e635959):

|                Method |      Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------------- |----------:|----------:|----------:|-------:|------:|------:|----------:|
| BenchmarkNonLocalised |  68.50 ns |  1.397 ns |  1.307 ns | 0.0131 |     - |     - |     152 B |
|    BenchmarkLocalised | 919.32 ns | 18.094 ns | 16.925 ns | 0.0677 |     - |     - |     784 B |

the extra 8 bytes here is the reference to `LocalisationManager`.

Closes #4679.

I *think* the benchmark not completing on master is enough by way of testing, but we could add proper testing by exposing some internals of `WeakList` if thats required.

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkLocalisedBindableString.cs
+++ b/osu.Framework.Benchmarks/BenchmarkLocalisedBindableString.cs
@@ -1,0 +1,45 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using BenchmarkDotNet.Attributes;
+using osu.Framework.Configuration;
+using osu.Framework.Localisation;
+using osu.Framework.Tests;
+
+namespace osu.Framework.Benchmarks
+{
+    [MemoryDiagnoser]
+    public class BenchmarkLocalisedBindableString
+    {
+        private LocalisationManager manager;
+        private TemporaryNativeStorage storage;
+
+        [GlobalSetup]
+        public void GlobalSetup()
+        {
+            storage = new TemporaryNativeStorage(new Guid().ToString());
+            manager = new LocalisationManager(new FrameworkConfigManager(storage));
+        }
+
+        [GlobalCleanup]
+        public void GlobalCleanup()
+        {
+            storage.Dispose();
+        }
+
+        [Benchmark]
+        public void BenchmarkNonLocalised()
+        {
+            var bindable = manager.GetLocalisedString("test");
+            bindable.UnbindAll();
+        }
+
+        [Benchmark]
+        public void BenchmarkLocalised()
+        {
+            var bindable = manager.GetLocalisedString(new TranslatableString("test", "test"));
+            bindable.UnbindAll();
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


smoogipoo (approved):
Looks ok

PR #4683, opened 2021-08-11
Reduce `FillFlowContainer` list allocations

Flutterish:
## **Edit: Updated results in comment below.**

Benchmark file names: `BenchmarkSingleFillFlowContainerAllocations` and `BenchmarkMultipleFillFlowContainerAllocations`.
```ini
BenchmarkDotNet=v0.13.0, OS=Windows 10.0.19043.1110 (21H1/May2021Update)
Intel Core i5-7600 CPU 3.50GHz (Kaby Lake), 1 CPU, 4 logical and 4 physical cores
.NET SDK=5.0.400
  [Host]     : .NET 5.0.9 (5.0.921.35908), X64 RyuJIT
  DefaultJob : .NET 5.0.9 (5.0.921.35908), X64 RyuJIT
```

## Old
Single `FillFlowContainer` (10k children total):
|   Method |     Mean |    Error |   StdDev |    Gen 0 |    Gen 1 | Gen 2 | Allocated |
|--------- |---------:|---------:|---------:|---------:|---------:|------:|----------:|
| RunFrame | 12.60 ms | 0.246 ms | 0.263 ms | 734.3750 | 203.1250 |     - |   2.51 MB |

Multiple `FillFlowContainers` (100k children total):
|   Method |     Mean |    Error |   StdDev |     Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------- |---------:|---------:|---------:|----------:|------:|------:|----------:|
| RunFrame | 94.47 ms | 1.250 ms | 1.108 ms | 8000.0000 |     - |     - |  24.33 MB |

## New
Single `FillFlowContainer` (10k children total):
|   Method |     Mean |    Error |   StdDev |    Gen 0 |    Gen 1 | Gen 2 | Allocated |
|--------- |---------:|---------:|---------:|---------:|---------:|------:|----------:|
| RunFrame | 10.80 ms | 0.092 ms | 0.086 ms | 718.7500 | 171.8750 |     - |      2 MB |

Multiple `FillFlowContainers` (100k children total):
|   Method |     Mean |    Error |   StdDev |     Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------- |---------:|---------:|---------:|----------:|------:|------:|----------:|
| RunFrame | 92.88 ms | 1.696 ms | 1.416 ms | 7333.3333 |     - |     - |     22 MB |

## Summary
Lets do some math to understand the results better:
There are 100 containers with 1 000 children each in the `Multiple FillFlowContainers` test, for a total of 100 000 children (10 times less in the `Single FillFlowContainer` test).
That would be **1000B** for each byte needed per child, 100 times. To compute the layout positions we need:
* A `Vector2` array = **8B** each
* A `Vector2` enumerable in performLayout = **8B** each
* A `Drawable` array (children) = **8B** each on x64
* An `int` array (row indices) = **4B** each
* A row offset list, **4B** per row

For a grand total of **28B** per child totalling **28000B** or about **27.34kiB** per container and **2800000B** in total or about **2.67MiB**.

The first improvement this PR introduces is pooling the used arrays.
After a bit of testing I have also discovered that pooling the int or drawable arrays generally did nothing or increased memory usage, which probably means this one has already been optimized away in some way.
This leaves the 2 `Vector2` arrays. Pooling them would mean pooling **16B** per child. I was going to cache the first one which unfortunately cant be pooled as it is returned, but I managed to pool it by making the method an iterator rather than returning the list.

The other major change is that I replaced the int array with an int list, but one per row instead of per child.
It stores the amout of children per row instead of which row a child at index i belongs to.
This saves up to **4B** per child per method call, but also is pooled.

Another thing it does is it pools the row offset list, meaning now it doesnt need to be allocated each call. This does not reduce the usage by a lot, but reduces GC pressure.

This theoretically saves up to **20B** per child which is about **1.91MiB**.
Mind you, this is 20/28 used bytes reducing the memory usage of this method to **28.5%**.

Now, the report states that the memory went from **24.33MiB** to **22MiB** which is a **2.33MiB** delta, but I think the **22MiB** are rounded for some reason (no decimal places at all) bringing it closer to the theoretical value.

If you do the same math to the `Single FillFlowContainer` test, the theoretical value is **0.19MiB**, which given the result is rounded like I suspect, aligns perfectly.


osu.Framework/Allocation/ListPool.cs
```diff
@@ -0,0 +1,76 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace osu.Framework.Allocation
+{
+    /// <summary>
+    /// Provides a resource pool that enables reusing instances of type <see cref="PooledList"/>.
+    /// </summary>
+    /// <typeparam name="T">The type of the elements in the pooled lists.</typeparam>
+    public class ListPool<T>
```
peppy, line 13: Highly disagree with this. The only pooling I'd want to see is using the .NET built-in pooling, not more custom implementations. Especially for the marginal saving you're getting in this PR.

Flutterish: This PR shaves **71.5%** allocs of the affected method and **9.5%** in the whole layout process. I would not call that marginal, especially that this can easily be extended to other methods that use dynamically allocated lists as it uses an easy to use `using var ... = Rent()` syntax. The alternative to this is using a built in array pool and reserving a maximum possible size and a `try...finally` block every time which definitely makes the code look less readable. Another alternative is to store the lists that would otherwise be dynamically allocated in the object itself, but then youre wasting space if you have more than one or a child class overrides the method and doesnt call the base.

Flutterish: I will use the built in then. I'll try to figure out a way to make it look passable with lack of a defer statement in C#.

I dont understand how my benchmark is wrong though? It shows that the allocations went from **24.33MiB** to **22MiB** for the whole layout process (**-9.5%**) and I calculated that the 2 methods that I touched use a total of **2.67MiB**, which after the optimization changed by **-2.33MiB** (**-71.5%** after correcting for the result being rounded for some reason).

peppy: make a benchmark to show the 9x% reduction. move the irrelevant parts into `GlobalSetup`.

Flutterish:
Thank you for the review. I have updated the code to use a wrapper around an `ObjectPool`s default implementation for the list pool which enables using the much cleaner `using var list = pool.Rent()` syntax like previously. I hope that having a wrapper is fine.

I also updated the benchmarks to reflect the changes in `ComputeLayoutPositions`, `performLayout` and a whole update frame where the container is invalidated named (Single/Multiple) "ComputeLayoutPositions", "PerformLayout" and "WholeLayout" respectively.

Additionally, I made a few more changes to the code which tremendously reduced allocations.
The benchmark results are as follow:
## Old
|                         Method |        Mean |       Error |      StdDev |     Gen 0 | Gen 1 | Gen 2 |   Allocated |
|------------------------------- |------------:|------------:|------------:|----------:|------:|------:|------------:|
| MultipleComputeLayoutPositions | 39,900.7 μs |   607.53 μs |   568.28 μs | 1692.3077 |     - |     - |  5258.59 KB |
|   SingleComputeLayoutPositions |    293.1 μs |     3.55 μs |     3.15 μs |   17.0898 |     - |     - |    52.59 KB |
|            MultipleWholeLayout | 91,980.3 μs | 1,802.51 μs | 2,213.64 μs | 8000.0000 |     - |     - | 24908.75 KB |
|              SingleWholeLayout | 27,720.7 μs |   241.32 μs |   213.93 μs |   62.5000 |     - |     - |   249.14 KB |
|          MultiplePerformLayout | 67,940.1 μs | 1,334.17 μs | 1,310.33 μs | 8125.0000 |     - |     - | 24909.65 KB |
|            SinglePerformLayout |    544.1 μs |     4.90 μs |     4.59 μs |   81.0547 |     - |     - |   249.09 KB |

## New
|                         Method |             Mean |            Error |           StdDev |     Gen 0 | Gen 1 | Gen 2 |   Allocated |
|------------------------------- |-----------------:|-----------------:|-----------------:|----------:|------:|------:|------------:|
| MultipleComputeLayoutPositions |      1,981.35 ns |        25.241 ns |        23.610 ns |    3.3150 |     - |     - |    10,400 B |
|   SingleComputeLayoutPositions |         17.51 ns |         0.238 ns |         0.211 ns |    0.0331 |     - |     - |       104 B |
|            MultipleWholeLayout | 83,649,206.23 ns | 1,661,364.766 ns | 3,540,503.768 ns | 2285.7143 |     - |     - | 7,445,059 B |
|              SingleWholeLayout | 27,689,791.21 ns |   454,581.976 ns |   446,460.283 ns |         - |     - |     - |    74,509 B |
|          MultiplePerformLayout | 61,447,258.73 ns |   543,030.467 ns |   481,382.364 ns | 2333.3333 |     - |     - | 7,445,814 B |
|            SinglePerformLayout |    476,478.29 ns |     2,484.337 ns |     2,202.300 ns |   23.4375 |     - |     - |    74,449 B |

## Summary
`ComputeLayoutPositions` allocations were reduced from **5258.59kiB** to **10,400B** (**-99.8%**).
`performLayout` allocations were reduced from **24909.65kiB** to **7271.3kiB** (**-70.8%**), same results for the whole update frame.

2 major changes that allowed for this were: 
* A new extension method `T FirstOrDefaultOfType<T>(this IEnumerable<object> self)` in `EnumerableExtensions.cs` which combines `IEnumberable<T>.OfType<K>().FirstOrDefault()` into one optimized operation. I dont actually know why this is better because LINQ operations are supposed to be lazily evaluated but it demonstrably is (about **6000kiB** difference).
* Using `TransformsForTargetMember(nameof(FlowTransform))` instead of all `Transforms` when searching for a `FlowTransform` in `FlowContainer`s children.

peppy:
Is the wrapper just avoiding a single `try-finally`? If so I don't think it's helping with readability or code quality.

Flutterish:
It is indeed avoiding a `try...finally` block. However I think that having a whole method indented by one more level makes the code look way worse than a single `using` statement which has the same performance and implications as a `try...finally` block and additionally makes it clear that the rented list is returned on the same line as its rented rather than hidden somewhere at the end of a method.

peppy:
Disagree. Also we don't use non-scoped `using` so you would still need to nest it.

osu.Framework/Extensions/IEnumerableExtensions/EnumerableExtensions.cs
```diff
@@ -104,5 +104,20 @@ public static IEnumerable<IEnumerable<T>> CartesianProduct<T>(this IEnumerable<I
                     select accseq.Concat(new[] { item })
             );
         }
+
+        /// <summary>
+        /// Gets the first element of type <typeparamref name="T"/> or default from a sequence.
+        /// </summary>
+        public static T FirstOrDefaultOfType<T>(this IEnumerable<object> self)
+        {
+            using var enumerator = self.GetEnumerator();
```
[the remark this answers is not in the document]
frenzibyte: Yeah I don't like this as well, It seems to have been added on this commit https://github.com/ppy/osu-framework/commit/e55798f7c3fec2cb0ec42840bf622c8edf7bab55 which makes me wondering if it was intended for more performance to begin with.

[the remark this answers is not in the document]
bdach: it was already explained that this was added for performance above. [see comment](https://github.com/ppy/osu-framework/pull/4683#issuecomment-898190436).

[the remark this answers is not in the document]
frenzibyte: I see...

Flutterish:
Sorry this took this long, I honestly forgot about this. Is the current state acceptable? The alloc statistics did not change with the lambda expression I replaced the extension method with. (well, I think like 5 bytes were added)

smoogipoo (approved):
Good stuff! Lgtm!

peppy:
I'm still okay with things as they haven't changed substantially from the last time I worked on this.

peppy (approved):

Flutterish:
Hey there. I started working on the second part of my FFC plan (the cross axis thing), but I discovered a slight bug that was introduced on this PR. In the visual test for FFC it seems that when too many rows are added (not hundreads, it happens after like 6) and the anchor of the FFC and origin of children is set to `Anchor.Centre`, it starts to jitter the Y position and eventually act as if the child origin was `Anchor.TopCentre`.
image

I spent about 5 minutes after writing this thinking how this could be and I know - `The rented arrays are at least the requested size. This means you cant use .Length, ever.` and https://github.com/ppy/osu-framework/blob/8f08fffc8defcf46447b8fbd4abbdf43408672d5/osu.Framework/Graphics/Containers/FillFlowContainer.cs#L229
This means the height would sometimes be a value outside the actual range of the rented array. The fix is simple - `float height = layoutPositions[children.Length - 1].Y;`, I believe it will be easier for one of you to push this to the repo rather than me creating a next pr hah

peppy:
Please create a PR, and if feasible add a test covering it.

Flutterish:
I dont think this needs a test (I don't even know what it would do), its just that rented arrays have a size that is "at least the requested size", so you cant use `.Length` (since it can be more than the requested size) and the `^` indexer does that.

peppy:
If it can't be easily tested then without is fine.

Issue #3470, opened 2020-04-20
Hot paths should not contain `throw` to promote inlining


smoogipoo:
Attaching `throw new X` causes the JIT to disable inlining for the method. More information on optimisation techniques can be found [here](https://reubenbond.github.io/posts/dotnet-perf-tuning).

We use guards very liberally in `Drawable` properties such as Position, Size, Scale, Rotation, Margin, Padding.
Other hot paths such as `Cached<T>.Value`, `LayoutValue<T>.Value`, `Drawable.Invalidate()` and `InvalidationList.Invalidate()`, are also causes for concern.

The best way to do this is to create a helper static class to throw these exceptions for us, for example: https://source.dot.net/#Microsoft.AspNetCore.Server.Kestrel.Core/Internal/ThrowHelper.cs,0d4931613cfa3edb

It's quite a minor performance gain in the grand scheme of things but it all adds up, especially down the `Invalidation` paths. Over 10000 calls this is the sort of performance gain expected:
|       Method |     Mean |    Error |   StdDev |
|------------- |---------:|---------:|---------:|
| ViaDelegated | 12.56 us | 0.045 us | 0.037 us |
|    ViaDirect | 23.97 us | 0.052 us | 0.041 us |

Code:
```csharp
private readonly Responder resp = new Responder();

[Benchmark]
public int ViaDelegated()
{
    for (int i = 0; i < 10000; i++)
        resp.ContainerWithDelegatedThrow = i / 1000;

    return resp.ContainerWithDelegatedThrow;
}

[Benchmark]
public int ViaDirect()
{
    for (int i = 0; i < 10000; i++)
        resp.ContainerWithDirectThrow = i / 1000;

    return resp.ContainerWithDirectThrow;
}

private class Responder
{
    private int container;

    public int ContainerWithDelegatedThrow
    {
        get => container;
        set
        {
            if (value < 0 || value > 100)
                delegatedThrow();

            container = value;
        }
    }

    public int ContainerWithDirectThrow
    {
        get => container;
        set
        {
            if (value < 0 || value > 100)
                throw new InvalidOperationException();

            container = value;
        }
    }

    private static void delegatedThrow() => throw new InvalidOperationException();
}
```

PR #4689, opened 2021-08-15
Optimise `GetBoundCopy` to not rely on reflection to create bindable instances

### LINKED ISSUE #4685: Bindable.GetBoundCopy() is slow
Has come up twice this week as a performance smell. Needs further investigation, because it was done this way with the belief that the reflection overhead is mostly a once-off overhead. In practice it does not seem this way.

We could likely get away with a `virtual` implementation since we have a finite group of bindable types.

3C7932D4-925E-408C-BFE2-8CD47D84F491

frenzibyte:
# Breaking changes

## All custom implementations of `IBindable` must implement `CreateInstance()`

Bindables previously used `Activator.CreateInstance()` to implement the `GetBoundCopy()` call. In profiling this has turned out to be a bottleneck in some scenarios, so in order to reduce the associated runtime overhead to about half, all implementors of `IBindable` now must implement `CreateInstance()`.

This also applies to all inheritors of framework-provided bindable types, who must instead override `CreateInstance()` from their base types.

For a given leaf bindable type in the inheritance hierarchy, the method should return a new constructed instance of the leaf type. The recommended pattern to use is as follows:

```csharp
public class BaseBindable : IBindable
{
    IBindable IBindable.CreateInstance() => CreateInstance();
    protected virtual BaseBindable CreateInstance() => new BaseBindable();

    IBindable IBindable.GetBoundCopy() => GetBoundCopy();
    public BaseBindable GetBoundCopy()
        => IBindable.GetBoundCopyImplementation(this); // automatically calls CreateInstance()
}

public sealed class CustomBindable : BaseBindable
{
    protected override BaseBindable CreateInstance() => new CustomBindable();
}
```

---

Closes #4685 

|          Method |     Mean |     Error |    StdDev | Ratio | RatioSD |
|---------------- |---------:|----------:|----------:|------:|--------:|
| GetBoundCopyOld | 2.352 us | 0.0425 us | 0.0840 us |  1.00 |    0.00 |
|    GetBoundCopy | 1.071 us | 0.0212 us | 0.0343 us |  0.46 |    0.02 |

While this avoids reflection-based metods of creating a bindable instance, the new protected `CreateInstance` method has to check that the returned instance matches the same type as the original bindable which is to be bound against, and throw if it doesn't match.

Doing that causes JIT to disable inlining on the method (see https://github.com/ppy/osu-framework/issues/3470). I've defined a `ThrowHelper` class for that purpose, as otherwise `GetBoundCopy` would've been less performant.

---

The fact that all inheritors of `Bindable`s need to override `CreateInstance` is a bit disturbing, but since the chances of making a new bindable class is rare, I believe it's worth the optimisation.

I've left `LocalisedBindableString` from implementing `CreateInstance` on it as it doesn't seem possible to do so, and it's already broken before so I don't see any way of supporting that currently. Leaving it for later when deemed required.


peppy:
Why does the `throw` need to be there? Is this intended to protect third parties that may be implementing their own bindable incorrectly? Not sure we need that.

peppy:
In addition, I'm not sure why you've listed four types of invocation in the benchmark when there should only be two compared - what we were using and what the new code is providing. It's unclear immediately which two we should be looking at (guessing the first two, where the... second is the old?), especially since the `baseline` is set incorrectly to the new one, rather than the old.

Here's a better way (in my opinion) to create benchmark comparisons for PR descriptions:

- Copy the pre-changes version of `Bindable` to `BindableOld`
- Create two benchmark methods with the same content, one instantiating a `Bindable` and the other `BindableOld`. Set the old one as the baseline

frenzibyte:
> Why does the `throw` need to be there? Is this intended to protect third parties that may be implementing their own bindable incorrectly? Not sure we need that.

Added that for two reasons:
 1. Would act same as what happens when invoking `Activator.CreateInstance` on a bindable which doesn't implement a matching constructor.
 2. Avoids the chances of forgetting to override `CreateInstance`, which can happen silently and result in unexpected behaviour. I can imagine that potentially happening osu!-side, with bindables that override the `Value` setter for example.

That said, I'm wondering if there's a reason you're against having it?

frenzibyte:
> In addition, I'm not sure why you've listed four types of invocation in the benchmark when there should only be two compared - what we were using and what the new code is providing. It's unclear immediately which two we should be looking at (guessing the first two, where the... second is the old?), especially since the `baseline` is set incorrectly to the new one, rather than the old.
> 
> Here's a better way (in my opinion) to create benchmark comparisons for PR descriptions:
> 
> * Copy the pre-changes version of `Bindable` to `BindableOld`
> * Create two benchmark methods with the same content, except that one instantiates a `Bindable` and the other `BindableOld`. Set the old one as the baseline

I thought `Baseline` is supposed to be set on the pure way of creating a bindable, then the other benchmarks would scale based on it to show the ratio between the pure way and the other methods. But perhaps not.

Your way sounds good, I can't exactly copy the pre-changed version of `Bindable`, but I can at least make a bindable that overrides to create the instance via `Activator` to reflect the overhead.

frenzibyte:
Benchmarks updated, shows a ratio of 0.50 on the new way.

osu.Framework/Bindables/Bindable.cs
```diff
@@ -391,25 +390,15 @@ public Bindable<T> GetUnboundCopy()
...
-            return copy;
-        }
-
-        IBindable IBindable.GetBoundCopy() => GetBoundCopy();
+        protected internal virtual Bindable<T> CreateInstance() => new Bindable<T>();
 
         IBindable<T> IBindable<T>.GetBoundCopy() => GetBoundCopy();
 
+        /// <inheritdoc cref="IBindable.GetBoundCopy"/>
+        public Bindable<T> GetBoundCopy() => (Bindable<T>)((IBindable)this).GetBoundCopy();
```
bdach, line 400: pretty gnarly but i don't see how this can be avoided...

osu.Framework/Bindables/LeasedBindable.cs
```diff
@@ -37,8 +37,8 @@ internal LeasedBindable([NotNull] Bindable<T> source, bool revertValueOnReturn)
             Disabled = true;
         }
 
-        public LeasedBindable(T value)
-            : base(value)
+        private LeasedBindable(T defaultValue = default)
+            : base(defaultValue)
```
bdach, line 41: is this change related?

peppy:
I guess the throwing is required due to the `virtual` implementation in `Bindable`, seems fine to leave it.

The new benchmarks make me question whether this was required in the first place. At 2.3us this method likely does not need this change.

osu.Framework/Utils/ThrowHelper.cs
```diff
@@ -0,0 +1,14 @@
...
+#nullable enable
+
+using System;
+
+namespace osu.Framework.Utils
+{
+    public static class ThrowHelper
+    {
+        public static void ThrowInvalidOperationException(string? message) => throw new InvalidOperationException(message);
+    }
```
bdach, line 13: I would rather have this scoped to the concrete bindable use case than just be a generic throw helper for one exception type. Or lean into it fully and have a `ThrowHelper<T>` (does that still get the perf benefits? need to check if going for that)

osu.Framework/Bindables/LeasedBindable.cs
frenzibyte: Brings it to a consistent state so I wouldn't have to pass `default` while constructing `LeasedBindable`, nothing important.

osu.Framework/Utils/ThrowHelper.cs
frenzibyte: Added it to be a good start for https://github.com/ppy/osu-framework/issues/3470, I'm wondering how it's possible to make `ThrowHelper` generic though? (assuming the generic parameter would be the exception)

osu.Framework/Bindables/Bindable.cs
frenzibyte: Yeah...

smoogipoo: Can this be extracted to a helper method rather than a sealed interface implementation?

frenzibyte: Had thought about it but scratched that away for the fact that `CreateInstance` used to be protected, but now that I've made it `internal` as well, no reason not to.

frenzibyte: This actually can't be a helper method as `IBindable<T>` has to return a bound `IBindable<T>` copy, while `Bindable<T>` has to return a bound `Bindable<T>` copy.

osu.Framework/Utils/ThrowHelper.cs
peppy: At very least, this should have xmldoc attached to the class to mention why it's being done. And potentially linking to the issue as a TODO for future usage.

peppy (changes requested):

osu.Framework/Utils/ThrowHelper.cs
frenzibyte: Good note, I've added that.

peppy (dismissed):

smoogipoo:
I've refactored this to remove the back-and-forth casting.

* `IBindable` exposes a `GetBoundCopy()` method.
* Custom interfaces like `IBindable<T>` can expose their own typed variant of `GetBoundCopy()`
* The final implementation (e.g. `Bindable<T>`) implements `GetBoundCopy()` via `IBindable.GetBoundCopyImplementation(this)`.
* `IBindable.GetBoundCopyImplementation()` is only available within `IBindable` subclasses.
* `IBindable.CreateInstance()` is now only `protected` instead of `protected internal`.

|          Method |       Mean |   Error |  StdDev | Ratio |
|---------------- |-----------:|--------:|--------:|------:|
| GetBoundCopyOld | 1,729.3 ns | 2.92 ns | 2.59 ns |  1.00 |
|    GetBoundCopy |   797.9 ns | 7.01 ns | 6.22 ns |  0.46 |

Relative performance hasn't changed from the OP.

frenzibyte (commented):
Looks good.

peppy (approved):

smoogipoo (approved):

bdach (approved):

bdach:
Code seems good to me. I've added a breaking change note to the PR description.

@frenzibyte before I go merging this, do you have the corresponding game-side changes ready? because this *will* break custom bindable types and the game has a few of those.

frenzibyte:
Yeah I'll ready up one now.

PR #4690, opened 2021-08-15
Introduce HandleModifier properties

### LINKED ISSUE #3992: KeyBindingContainer (somehow) needs to intertwine OnPressed events with standard input events

UselessToucan:
resolves #3992

Since the problem is mainly about modifier keys(Ctrl, Alt, Shift, Win), I have added `HandleCtrl`/`HandleAlt`/`HandleShift`/`HandleSuper` properties to the `Drawable` class. They are used to determine whether a button input event should be propagated to a drawable if the corresponding modifier is pressed.

Known issues:
1. `ScrollEvent`. It may [depend on a modifier state](https://github.com/ppy/osu/blob/7654df94f6f37b8382be7dfcb4f674e03bd35427/osu.Game/Graphics/Containers/OsuScrollContainer.cs#L105), but it is propagated by an `InputManager`. It probably should support new properties, but I haven't figured how to do it yet.

### DIFF
--- a/osu.Framework.Tests/Visual/Audio/TestSceneTrackAdjustments.cs
+++ b/osu.Framework.Tests/Visual/Audio/TestSceneTrackAdjustments.cs
@@ -182,6 +182,8 @@ namespace osu.Framework.Tests.Visual.Audio
             private void updateLocal(ValueChangedEvent<double> obj) =>
                 textLocal.Text = $"local: vol {audio.Volume.Value:F1} freq {audio.Frequency.Value:F1} tempo {audio.Tempo.Value:F1} bal {audio.Balance.Value:F1}";
 
+            public override bool HandleControl => true;
+
             protected override void OnDrag(DragEvent e)
             {
                 Position += e.Delta;
--- a/osu.Framework.Tests/Visual/Input/TestSceneKeyBindingContainer.cs
+++ b/osu.Framework.Tests/Visual/Input/TestSceneKeyBindingContainer.cs
@@ -7,6 +7,7 @@ using NUnit.Framework;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.UserInterface;
 using osu.Framework.Input.Bindings;
+using osu.Framework.Input.Events;
 using osu.Framework.Testing;
 using osuTK;
 using osuTK.Input;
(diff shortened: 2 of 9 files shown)


bdach:
I don't really get the point of this. How does this help with the linked issue, exactly?

As far as I understand this is just splitting key event handlers into more overly-specific methods (why are they modifier-specific, rather than being a general predicate that accepts a `KeyCombination`, for instance?) that have to be overridden anyhow to make the key binding container work. So to say that this resolves the linked issue seems quite the stretch.

UselessToucan:
> How does this help with the linked issue, exactly?

The problem is that a keypress which is expected to be handled as a keybinding using `OnPressed`/`OnReleased` methods, may be intercepted by some `OnKeyDown` method. At the moment, these cases are resolved locally: [sample 1](https://github.com/ppy/osu/blob/473011070fb0a8d6d1752836657572b0abf59205/osu.Game/Rulesets/Edit/HitObjectComposer.cs#L224), [sample 2](https://github.com/ppy/osu/blob/master/osu.Game/Screens/Menu/Button.cs#L212).

Suggested changes allows us to drop these workarounds. On the other side, it requires us to add property overrides if we need to use modifiers in a standard input method like `OnKeyDown`. If you are interested in how often it is required, see these commits: [framework](https://github.com/UselessToucan/osu-framework/commit/83efaf7130aea6ddb7b4964ad2098d61f6f90959), [game](https://github.com/UselessToucan/osu/commit/441d3cd59914a4c7aa75441099eae279d63ca215#diff-53d75a5400d727a5ff69b3f1bfa3ce8b3b762ba42ae7b45646e8974a1b032f10L95).

frenzibyte:
It sounds more like a workaround (and an extension to what's being currently done in the samples you posted), than a correct approach to resolve https://github.com/ppy/osu-framework/issues/3992.

Since the issue is regarding drawables receiving pure key events regardless of whether keybinding events were handled, I believe a first step would be to refactor `KeyBindingContainer` to inherit from `InputManager` and take control of event propagation, so that it can propagate keybinding events first, then propagate regular keyboard events if no drawable handled the keybinding event.

smoogipoo:
I think you're misunderstanding what the issue is, and possibly @frenzibyte is too.

Suppose a hierarchy is as follows:
```
Drawable1  <-- Handles raw event
Drawable2  <-- Doesn't handle anything
Drawable3  <-- Doesn't handle anything
Drawable4  <-- Handles keybinding
```
Input is handled in reverse order (closest to screen first), which would be 4 -> 3 -> 2 -> 1. The problem is that we propagate raw events first and _then_ keybinding events. Unless a `KeyBindingContainer` with `Prioritised = true` is set, in which case the order reverses at that point.  

The issue goes both ways, but for the example above consider `Prioritised=false` (for the `true` case, swap around 1 and 4).

So you'd look at 4, but it doesn't handle the raw event, 3 doesn't either, and neither does 2. But 1 does, and so it would receive the _entire_ event. If 1 didn't handle, input would then go on to form a keybinding to be handled by 4.

With intertwined input, it would mean that keybinding and raw events would be handled in lockstep:
```
Drawable4 -> try handle raw event
Drawable4 -> try handle keybinding
Drawable3 -> try handle raw event
Drawable3 -> try handle keybinding
Drawable2 -> try handle raw event
Drawable2 -> try handle keybinding
Drawable1 -> try handle raw event
Drawable1 -> try handle keybinding
```
Which, in this case, would handle input at 4.

frenzibyte:
Aha, I have misunderstood this indeed. Wasn't quite clear from the issue thread what "intertwined input" means.

So that concludes this PR to not have resolved the underlying issue in any capacity, and that what's needed here is indeed a refactor of `KeyBindingContainer` to become an `InputManager` of its own and control the event propagation to propagate both keybinding event and regular key event in lockstep for each drawable.

I don't see any other way around that without going unnecessarily extreme, I believe.

smoogipoo:
Going to close this as not the correct path forward.

PR #4695, opened 2021-08-18
Use a temporary directory to store unit test headless run files

peppy:
Generally it's good practice to store temporary files to the OS defined directory. This avoids IDEs or other OS level operations (ie. spotlight) indexing the files as they are created.

This also has some fixes to avoid log entries being written after a test host has already been disposed. Not sure whether this will be accepted as a solution, but it seems like about the best we're going to be able to do given the `static`-ness of `Logger`. Works well in practice.

Some further changes are required osu!-side to make use of this new class (in non-standard scenarios).

### DIFF
--- a/osu.Framework/Testing/TestRunHeadlessGameHost.cs
+++ b/osu.Framework/Testing/TestRunHeadlessGameHost.cs
@@ -34,7 +34,15 @@ namespace osu.Framework.Testing
             base.Dispose(isDisposing);
 
             if (!bypassCleanup)
-                Storage?.DeleteDirectory(string.Empty);
+            {
+                try
+                {
+                    Storage.DeleteDirectory(string.Empty);
+                }
+                catch
+                {
+                }
+            }
         }
     }
 }
--- a/osu.Framework/Testing/TestRunHeadlessGameHost.cs
+++ b/osu.Framework/Testing/TestRunHeadlessGameHost.cs
@@ -3,17 +3,18 @@
 
 using System.IO;
 using osu.Framework.Logging;
+using osu.Framework.Platform;
 
-namespace osu.Framework.Platform
+namespace osu.Framework.Testing
 {
     /// <summary>
     /// A GameHost which writes to the system temporary directory, attempting to clean up after the test run completes.
     /// </summary>
-    public class TestHeadlessGameHost : HeadlessGameHost
+    public class TestRunHeadlessGameHost : HeadlessGameHost
     {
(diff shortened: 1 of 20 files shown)


osu.Framework/Testing/TestRunHeadlessGameHost.cs
```diff
@@ -0,0 +1,40 @@
...
+        protected override void Dispose(bool isDisposing)
+        {
+            // ensure no more log entries are written during cleanup.
+            // there is a flush call in base.Dispose which seals the deal.
+            Logger.Enabled = false;
+
+            base.Dispose(isDisposing);
+
+            if (!bypassCleanup)
+                Storage?.DeleteDirectory(string.Empty);
```
peppy, line 37: Turns out this is already done in `TestScene` itself, but I don't see harm in doing it locally as well.

smoogipoo: Should have a try-catch around this.

smoogipoo (changes requested):

osu.Framework/Testing/TestRunHeadlessGameHost.cs
peppy: Was somehow hoping this wouldn't be required any more, since I didn't repro any failures, but sounds fair.

smoogipoo:
`drwxr-xr-x  3 smgi    smgi       60 Aug 19 10:33 of-test-headless`

Is there a reason why this doesn't append a GUID in all cases? We can't have tests re-using the same dir.

peppy:
> 
> 
> `drwxr-xr-x 3 smgi smgi 60 Aug 19 10:33 of-test-headless`
> 
> Is there a reason why this doesn't append a GUID in all cases? We can't have tests re-using the same dir.

The top level directory is intended so it's easy to clean up the files in one go. It's never used directly for a test - each test will use a subdirectory in that base path. GUIDs are already added to tests where they need unique directories, and the ones which don't (due to requiring constant names between test runs) tend to use a `CleanRunGameHost` or similar.

Basically nothing has changed from `master` to this in this regard. That said, if there is a way to handle a global "tests started" and "tests ended" I wouldn't be against making a unique subdirectory in the `of-test-headless` path for the full run, and potentially nuke it afterwards.

smoogipoo:
Ah, I thought that was the top-level directory for hosts. It's fine.

smoogipoo (approved):

PR #4696, opened 2021-08-18
Optimise `BeginAbsoluteSequence` recursive case

peppy:
Tackling remaining super hot allocations during gameplay (based on number of objects allocated, not size of objects).

master:
|                     Method |       Mean |    Error |   StdDev |   Gen 0 |  Gen 1 | Gen 2 | Allocated |
|--------------------------- |-----------:|---------:|---------:|--------:|-------:|------:|----------:|
|               NonRecursive | 1,032.7 us | 19.79 us | 19.44 us |       - |      - |     - |     198 B |
|                  Recursive |   994.2 us | 17.76 us | 16.61 us |  3.9063 |      - |     - |  56,294 B |
| SlightlyNestedNonRecursive | 1,025.9 us | 18.20 us | 16.13 us |       - |      - |     - |     198 B |
|    SlightlyNestedRecursive | 1,185.3 us | 23.11 us | 20.49 us | 17.5781 | 5.8594 |     - | 208,294 B |
|     VeryNestedNonRecursive | 1,019.5 us | 20.12 us | 16.80 us |       - |      - |     - |     194 B |
|        VeryNestedRecursive | 1,108.5 us | 21.74 us | 22.32 us | 11.7188 | 1.9531 |     - | 152,294 B |

this pr:
|                     Method |       Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------------------- |-----------:|---------:|---------:|-------:|------:|------:|----------:|
|               NonRecursive | 1,019.7 us | 19.81 us | 18.53 us |      - |     - |     - |     194 B |
|                  Recursive |   975.4 us | 18.79 us | 25.09 us | 3.9063 |     - |     - |  56,291 B |
| SlightlyNestedNonRecursive | 1,023.6 us | 18.99 us | 17.77 us |      - |     - |     - |     198 B |
|    SlightlyNestedRecursive | 1,090.3 us | 17.23 us | 15.28 us | 9.7656 |     - |     - | 120,334 B |
|     VeryNestedNonRecursive | 1,034.7 us | 14.08 us | 13.17 us |      - |     - |     - |     194 B |
|        VeryNestedRecursive | 1,026.4 us | 19.24 us | 20.59 us | 3.9063 |     - |     - |  64,870 B |

Is this direction is seen as agreeable, the same refactor can be applied to `BeginDelayedSequence`. There is probably more that can (and should) be done with transform logic, but this is a quick win.

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkBeginAbsoluteSequence.cs
+++ b/osu.Framework.Benchmarks/BenchmarkBeginAbsoluteSequence.cs
@@ -146,11 +146,15 @@ namespace osu.Framework.Benchmarks
 
                 Add(VeryNested = new Container());
 
-                Container nestedContainer = null;
+                Container target = VeryNested;
 
                 for (int i = 0; i < 1000; i++)
                 {
-                    (nestedContainer ?? VeryNested).Add(nestedContainer = new Container { Size = new Vector2(100), Colour = Color4.Black });
+                    var container = new Container { Size = new Vector2(100), Colour = Color4.Black };
+
+                    target.Add(container);
+
+                    target = container;
                 }
             }
         }
--- a/osu.Framework.Benchmarks/BenchmarkBeginAbsoluteSequence.cs
+++ b/osu.Framework.Benchmarks/BenchmarkBeginAbsoluteSequence.cs
@@ -1,0 +1,158 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using BenchmarkDotNet.Attributes;
+using NUnit.Framework;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Shapes;
+using osuTK;
+using osuTK.Graphics;
+
+namespace osu.Framework.Benchmarks
+{
+    [MemoryDiagnoser]
+    public class BenchmarkBeginAbsoluteSequence : GameBenchmark
+    {
+        private TestGame game;
+
+        [Test]
+        [Benchmark]
+        public void NonRecursive()
+        {
+            game.Schedule(() =>
+            {
+                using (game.Flat.BeginAbsoluteSequence(1000, false))
+                {
+                }
+            });
+
+            RunSingleFrame();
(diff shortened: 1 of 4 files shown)


peppy:
Real world:

single play session (running through `TestSceneAutoplay` once) 

object alloc count reduction 15%
memory alloc bytes reduction 10%

image

osu.Framework.Benchmarks/BenchmarkBeginAbsoluteSequence.cs
```diff
@@ -0,0 +1,158 @@
...
+                }
+
+                Add(VeryNested = new Container());
+
+                Container nestedContainer = null;
+
+                for (int i = 0; i < 1000; i++)
+                {
+                    (nestedContainer ?? VeryNested).Add(nestedContainer = new Container { Size = new Vector2(100), Colour = Color4.Black });
+                }
```
bdach, line 154: Maybe not too relevant, but this is... pretty dense code. For starters, the null-coalesce is redundant; you can just do `nestedContainer = VeryNested` at the start. Secondly, this could be rewritten as:

```
nestedContainer = nestedContainer.Child = new Container { ... }
```

But that's still a bit confusing (you have to remember that assignment is right-associative), so I'd probably just split into two statements.

osu.Framework/Graphics/Transforms/Transformable.cs
```diff
@@ -291,6 +291,16 @@ public virtual IDisposable BeginAbsoluteSequence(double newTransformStartTime, b
         {
             EnsureTransformMutationAllowed();
 
+            return createAbsoluteSequenceAction(newTransformStartTime);
+        }
+
+        internal virtual void BeginAbsoluteSequenceRecursive(double newTransformStartTime, List<IDisposable> actions)
```
bdach, line 297: I think this makes sense to do, but I'm not super sold on the name. Maybe `CollectAbsoluteSequenceActionsFromSubTree` or similar? Because it's a bit weird how this method mentions "recursion" but there is none to be had at a `Transformable` level, the recursion only kicks in for composites.

bdach (commented):
Looks valid to me. I have only relatively minor points to make.

smoogipoo (approved):

PR #4698, opened 2021-08-20
Update SDL and other menial dependencies

peppy:
Closes https://github.com/ppy/osu/issues/14399.


bdach:
Have tested the basics on my machine and things seem to work.

bdach (approved):

peppy:
oh this was supposed to be in the last release..

PR #4700, opened 2021-08-21
Add fallback path if AppContext.BaseDirectory is null

TheShadowEevee:
Closes https://github.com/ppy/osu/issues/6507

### PR Description: 
This PR is a simple fix to the long standing "Could not load rulesets from directory" error that appears on some (if not all?) Android devices apon every boot of the game. This apparently was being caused by https://github.com/xamarin/xamarin-android/issues/3489: TL;DR `AppContext.BaseDirectory` currently returns null on Android, and as `LoadFromDisk()` in RulesetStore.cs was the only function I noticed that relied on that heavily, it was the only one mainly affected.

### Fix:
`RuntimeInfo.StartupDirectory` continues to return `AppContext.BaseDirectory` by default, meaning nothing current should change, while now returning `Environment.SpecialFolder.Personal` if BaseDirectory is null. This is equivalent to the path `/data/user/0/sh.ppy.osulazer/files` on Android.

### References:
Xamarin implementation of `Environment.SpecialFolder`:
https://docs.microsoft.com/en-us/xamarin/android/platform/files/

### Tested Platforms:

- [x] Windows - Win10 21H1
- [x] MacOS - No change, `AppContext.BaseDirectory` returns correct path
- [x] Linux - Linux Mint, fully updated
- [x] Android - Hardware: Android Oreo and Android 11 (S7 Active and S20 FE)
- [x] iOS - No change, `AppContext.BaseDirectory` returns correct path (tested on separate project because iOS is broken here)

### Additional Notes:
- Given the issue at hand, I believe https://github.com/ppy/osu/issues/6507 is a candidate for transfer to this repo
- I am open to suggestions, including "Why is this PR so long for a one line change?"

### DIFF
--- a/osu.Framework/RuntimeInfo.cs
+++ b/osu.Framework/RuntimeInfo.cs
@@ -13,7 +13,7 @@ namespace osu.Framework
         /// <summary>
         /// The absolute path to the startup directory of this game.
         /// </summary>
-        public static string StartupDirectory { get; } = AppContext.BaseDirectory;
+        public static string StartupDirectory => AppContext.BaseDirectory ?? Environment.GetFolderPath(Environment.SpecialFolder.Personal);
 
         /// <summary>
         /// Returns the absolute path of osu.Framework.dll.


bdach:
A similar sort of fix was already attempted locally game-side in https://github.com/ppy/osu/pull/6510. Do you perhaps have an explanation for why this would be different than the aforementioned attempt at a fix (which didn't work fully, which I can attest to)?

TheShadowEevee:
> A similar sort of fix was already attempted locally game-side in [ppy/osu#6510](https://github.com/ppy/osu/pull/6510). Do you perhaps have an explanation for why this would be different than the aforementioned attempt at a fix (which didn't work fully, which I can attest to)?

That PR attempted a fix based on the assumption that changing the current directory was enough to fix the path that lazer looks in, which at the time may have been plausible given the code has definitely changed since then. The difference is that currently, the function with issues calls `RuntimeInfo.BaseDirectory` directly, which would normally drop the path even with the executable. Given that `RuntimeInfo.BaseDirectory` essentially just ran `AppContext.BaseDirectory`, which was not implemented in Xamarin properly, Android platforms would attempt to access the folder "null" breaking the whole function.

frenzibyte:
It would be interesting if `Environment.GetSpecialFolder(Environment.SpecialFolder.Personal)` does indeed not return the correct app directory as specified in their documentations.

One of the possibilities is that `Environment.CurrentDirectory` (back when it was a thing) got mutated by a certain operation, one that's specific to Xamarin perhaps.

I'm not a fan of this being a thing as it reads completely incorrect, but I would be fine with it until Xamarin fixes `AppContext.BaseDirectory`.

frenzibyte:
Actually nevermind, I'm confused on what this is attempting to achieve.

The fallback you added is returning a user storage path (similar to what `AndroidGameHost.UserStoragePath`) rather than a path where the assemblies of the game are (`AppContext.BaseDirectory`), that doesn't sound correct at all?

You should be able to load custom rulesets regardless of this fix, no? the exception thrown from `loadFromDisk` is caught and should not affect the `RulesetStore` process in loading user rulesets by the `loadUserRulesets` method.

frenzibyte (changes requested):
Missing details

TheShadowEevee:
> The fallback you added is returning a user storage path (similar to what `AndroidGameHost.UserStoragePath`) rather than a path where the assemblies of the game are (`AppContext.BaseDirectory`), that doesn't sound correct at all?

I see your point there; though Android doesn't expose app installation folders that I know of. I guess that does expose an issue in my thought process, though I would consider this path to be the Android startup directory. The fallback path is the private internal storage directory for osu!lazer, not publicly accessible without using `runas` or `backup` via adb.

> You should be able to load custom rulesets regardless of this fix, no? the exception thrown from `loadFromDisk` is caught and should not affect the `RulesetStore` process in loading user rulesets by the `loadUserRulesets` method.

That's true, although it's much better to fix the issue and not have an error on every start of the game. I don't claim to fix ruleset loading itself in anyway, just this specific error.

frenzibyte:
This specific point caused my confusion about your intentions here:
> An oddity that needs investigation: The Oreo device can't run rulesets (no errors either), but the Android 11 device can. While not in the scope of this PR that is something that should be looked at. I tested Sentakki! and Rush ruleset-wise, placing them in Android/data/sh.ppy.osulazer/files/rulesets.

You're mentioning a completely separate issue that made me think is somewhat relevant here.

> I see your point there; Android doesn't expose app installation folders that I know of. I guess that does expose an issue in my thought process, though I would consider this path to be the Android startup directory. The fallback path is the private internal storage directory for osu!lazer, not publicly accessible without using `runas` or `backup` via adb.
 
Although you're right about this and the (usually public and user data) `Environment.SpecialFolder.Personal` does point to an "internal storage" that's private and not accessible by user, I don't see the point of pointing to a folder that wouldn't really have what it should (i.e. arbitrary folder).

Have you attempted handling this osu!-side by checking if `RuntimeInfo.StartupDirectory` is null and not attempt to load from disk? here specifically, where it's mentioned that android cannot load rulesets from startup directory:
https://github.com/ppy/osu/blob/2dadc9d686071ae4e1df0e5c6859f584c28174ce/osu.Game/Rulesets/RulesetStore.cs#L30-L33

TheShadowEevee:
> You're mentioning a completely separate issue >  that made me think is somewhat relevant here.

You're right, that's completely irrelevant. I edited it out, sorry.

> Have you attempted handling this osu!-side instead

I agree this would likely be the better way to go given the folder differences, and should be easy enough to do given I tried it during testing. 

The reason I didn't do this in the first place was because currently `RuntimeInfo.StartupDirectory` returning null is unexpected behavior which could cause issues in other games/programs using the framework as a base.

I'll reimplement and retest this osu! side in a few hours, but I believe it would be best if someone made a note about StartupDirectory returning bull on Android platforms.

PR #4702, opened 2021-08-21
Fix HSV colour picker shaders not working on GL ES

bdach:
Closes https://github.com/ppy/osu/issues/14383.

Seemingly the cause was insufficient precision of `v_TexCoord` which is rectified by cc626cc.

11c755a is sort of unrelated but after reading the shader code after a while it felt weird/unidiomatic to use array indices so I opted to swizzle instead. Likely to end up being faster in the saturation selector case at least (although I don't claim to have benchmarked).

### DIFF
--- a/osu.Framework/Resources/Shaders/sh_HueSelectorBackground.fs
+++ b/osu.Framework/Resources/Shaders/sh_HueSelectorBackground.fs
@@ -5,6 +5,6 @@ varying mediump vec4 v_TexRect;
 
 void main(void)
 {
-    float hueValue = v_TexCoord.x / (v_TexRect[2] - v_TexRect[0]);
+    float hueValue = v_TexCoord.x / (v_TexRect.z - v_TexRect.x);
     gl_FragColor = hsv2rgb(vec4(hueValue, 1, 1, 1));
 }
--- a/osu.Framework/Resources/Shaders/sh_HueSelectorBackground.fs
+++ b/osu.Framework/Resources/Shaders/sh_HueSelectorBackground.fs
@@ -1,6 +1,6 @@
 #include "sh_Utils.h"
 
-varying mediump vec2 v_TexCoord;
+varying highp vec2 v_TexCoord;
 varying mediump vec4 v_TexRect;
 
 void main(void)
--- a/osu.Framework/Resources/Shaders/sh_HueSelectorBackgroundRounded.fs
+++ b/osu.Framework/Resources/Shaders/sh_HueSelectorBackgroundRounded.fs
@@ -5,6 +5,6 @@ varying highp vec2 v_TexCoord;
 
 void main(void)
 {
-    float hueValue = v_TexCoord.x / (v_TexRect[2] - v_TexRect[0]);
+    float hueValue = v_TexCoord.x / (v_TexRect.z - v_TexRect.x);
     gl_FragColor = getRoundedColor(hsv2rgb(vec4(hueValue, 1, 1, 1)), v_TexCoord);
 }
--- a/osu.Framework/Resources/Shaders/sh_HueSelectorBackgroundRounded.fs
+++ b/osu.Framework/Resources/Shaders/sh_HueSelectorBackgroundRounded.fs
@@ -1,7 +1,7 @@
 #include "sh_Utils.h"
 #include "sh_Masking.h"
 
-varying mediump vec2 v_TexCoord;
+varying highp vec2 v_TexCoord;
 
 void main(void)
 {
--- a/osu.Framework/Resources/Shaders/sh_SaturationSelectorBackground.fs
+++ b/osu.Framework/Resources/Shaders/sh_SaturationSelectorBackground.fs
@@ -1,6 +1,6 @@
 #include "sh_Utils.h"
 
-varying mediump vec2 v_TexCoord;
+varying highp vec2 v_TexCoord;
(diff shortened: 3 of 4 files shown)


frenzibyte (approved):

frenzibyte:
Interestingly, iOS still has this broken, but after increasing the precision of `v_TexRect` of both shaders as well it works again. Although I'm wondering if changing to `highp` has an impact on performance, I guess not given we're talking about a colour picker shader.

bdach:
`v_TexRect` cannot be very easily changed because it's defined in the masking shader file which is included in the rounded variants, so declaring again causes a compilation failure.

While I could technically assign it to a higher precision local, at that point the precision has presumably already been lost. But that's not something I can check myself either.

frenzibyte:
Right, I don't have any bright suggestions on the matter, you could use `#if GL_ES` preprocessors but then I wonder how bad it would be to switch to using `highp` in all scenarios (also wondering how possible it is to benchmark such change)

bdach:
Before I go doing that you could try if the first and easiest way works, if you wish (copying it to a local `highp vec4` in the first line of the shaders in which it is included outwardly). I don't expect that to do anything from the light reading I did about precision modifiers but maybe I'll be surprised, because I won't claim to have precisely understood the docs as they are somewhat confusing.

bdach:
After @frenzibyte has excluded some other ideas I've thrown around I guess the only ways I can think of to fix this is either:

1. to have `v_TexRect` be `highp` everywhere, or
2. by preprocessor (ab)use in doing something like the following:

```diff
diff --git a/osu.Framework/Resources/Shaders/sh_Masking.h b/osu.Framework/Resources/Shaders/sh_Masking.h
index 172da4a7e5..7cabd4722f 100644
--- a/osu.Framework/Resources/Shaders/sh_Masking.h
+++ b/osu.Framework/Resources/Shaders/sh_Masking.h
@@ -1,6 +1,8 @@
 ﻿varying highp vec2 v_MaskingPosition;
 varying lowp vec4 v_Colour;
-varying mediump vec4 v_TexRect;
+#ifndef CUSTOM_TEX_RECT
+    varying mediump vec4 v_TexRect;
+#endif
 varying mediump vec2 v_BlendRange;
 
 uniform highp float g_CornerRadius;
diff --git a/osu.Framework/Resources/Shaders/sh_SaturationSelectorBackgroundRounded.fs b/osu.Framework/Resources/Shaders/sh_SaturationSelectorBackgroundRounded.fs
index 1213e78382..a25ede7dbd 100644
--- a/osu.Framework/Resources/Shaders/sh_SaturationSelectorBackgroundRounded.fs
+++ b/osu.Framework/Resources/Shaders/sh_SaturationSelectorBackgroundRounded.fs
@@ -1,8 +1,11 @@
+#define CUSTOM_TEX_RECT
+
+varying highp vec4 v_TexRect;
+varying highp vec2 v_TexCoord;
+
 #include "sh_Utils.h"
 #include "sh_Masking.h"
 
-varying mediump vec2 v_TexCoord;
-
 uniform mediump float hue;
 
 void main(void)
```

But that's super dicey (relies on `#include` order, etc...)

Not sure what else to do here. Maybe rewrite without a shader somehow, because none of the shader solutions seem worth the hassle for such a local usage. Hoping someone can come up with some idea that could salvage this without much fallout everywhere else.

smoogipoo:
Shader includes should not define vertex attributes. We could add some post processing to the shader compiler which checks for duplicate definitions and picks the highest precision one + dedupes?

bdach:
I guess I could see that working, yeah. I wasn't aware that we were already parsing `#include` directives.

That de-duplicating code might end up being a touch specific though (e.g. only working for a subset of glsl types, or a subset of variables), because I don't want to reimplement a glsl parser just for that, obviously.

frenzibyte:
I also agree with custom parsers yeah, would make it easier to increase precision without working around it in the shaders themselves.

bdach:
Upon looking at the current shader compiler code I don't really feel super comfortable adding new functionality there without first covering the existing functionality with unit tests. I'd be willing to do that, but probably as a lower-priority item, and I'm not sure if anybody would be willing to go through that process and spend time on reviewing that (probably looking at a series of 3 pulls minimum at this point, this one included).

peppy:
Are we okay with getting this in as-is and leaving the post-processor improvements for a later change? Seems beneficial in terms of effort required.

frenzibyte:
Sounds fair, I've only seen the bug remaining in iOS so this can be merged alone and the rest can come later. (it'll still be an issue given that GL ES precisions are mobile-specific as far as I've read, but at least part of it will be fixed here)

frenzibyte (approved):

peppy:
Oh, this PR doesn't fix iOS? I guess I missed a portion of this thread. In that case we should probably still keep a tracking issue open but get the fix in for android. Or just hold off for longer – i'm not sure there's that high usage of the editor on android just yet.

bdach:
I'm fine with either, for what it's worth. This could be closed, even. The issue is not exactly the highest priority item. I mainly PR'd hoping that a simple change would be enough.

bdach:
I'm going to close this for now I think, because I'm not sure this is worth spending time on yet. The issue and the path forward on how to fix this everywhere are clear, I'd just rather this not add up to the open PR numbers without much reason.

PR #4703, opened 2021-08-22
Restructure `TextFlowContainer` to be backed by collection of `ITextPart`s

bdach:
This is the major change that is supposed to facilitate #4636.

# Breaking changes

## `TextFlowContainer` no longer returns raw `SpriteText`s

In preparation for adding localisation support to `TextFlowContainer`s, the various `AddText()`/`AddLine()` overloads will no longer return raw `SpriteText`s. Instead, an `ITextPart` structure will be returned.

Via `ITextPart`, the consumer can both access all `Drawables` associated with a given piece of text, as well as react to any future changes in representation of the text by subscribing to `DrawablePartsRecreated`. In the future, with more localisation changes, this event will be invoked once an `LocalisableString`'s displayable content changes, which will trigger a recreation of all parts' drawables, upon which any manual adjustments applied to `Drawables` can be re-applied again.

For consumers wanting to implement their own `ITextPart`s to extend the functionality of `TextFlowContainer`, an abstract `TextPart` is also provided which implements the typical flow of handling `Drawables` and `DrawablePartsRecreated`. The only thing that a consumer has to do when inheriting that class is to implement `CreateDrawablesFor(TextFlowContainer)`.

---

I have been sitting on this diff for a few reasons - other than the usual constant deficit of time, I have written this twice (the first attempt was swiftly defeated by existing game-side usage patterns), and feel like it could still use more testing (I have checked the weird usage patterns, as well as ran framework-side and game-side test suites against this change, but this is a change so far-reaching that that still feels too little). Therefore this pull is mostly intended as a request for comments as it comes to the structure.

Due to the size of this change, `LocalisableString`s are not yet stored or returned by the text flow. But I hope that the breaking change notice explains the central idea behind these changes and the way that they will fit into the localisation structure is self-evident. I have successfully managed to get `RomanisableString`s to work with the first iteration of this structure in a test scene.

An early draft of game-side changes to adapt to changes made in this PR can be previewed [here](https://github.com/bdach/osu/tree/text-flow-parts).

### DIFF
--- a/osu.Framework.Tests/Visual/Containers/TestSceneTextFlowContainer.cs
+++ b/osu.Framework.Tests/Visual/Containers/TestSceneTextFlowContainer.cs
@@ -4,10 +4,13 @@
 using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Extensions.Color4Extensions;
+using osu.Framework.Extensions.IEnumerableExtensions;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
+using osu.Framework.Testing;
+using osuTK;
 using osuTK.Graphics;
 
 namespace osu.Framework.Tests.Visual.Containers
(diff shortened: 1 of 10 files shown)


peppy:
I haven't gone through the internals for this, but the exposed interface and osu!-side changes look very sane 👍 .

bdach:
so i was going to go and resolve conflicts here after the merge of #4729, but as it turns out my supposition that this would be easy to apply after that pull was completely wrong and the addition of the generic type arguments forces me to either add generic arguments to all `ITextPart`s (which i don't like because it doesn't really fit with things like `TextPartManual` - if it wraps an arbitrary drawable, then why should it need a generic type only to be able to add to a `TextFlowContainer<T>`?), or partially revert that pull by removing the generic-typed `TextFlowContainer` and instead put the generic typing directly on `Add{Text,Paragraph}()`, yielding a signature of:

```csharp
public ITextPart AddText<T>(string text, Action<T> creationParameters = null)
    where T : SpriteText
```

i doubt this is going to get many opinions seeing there hasn't been any reviews yet, but any opinions on which is better? especially interested in @LeNitrous's opinion since it was your use case in the first place.

LeNitrous:
> i doubt this is going to get many opinions seeing there hasn't been any reviews yet, but any opinions on which is better? especially interested in @LeNitrous's opinion since it was your use case in the first place. i think i personally prefer the latter.

That method signature is better. I suppose making the entire container generic was a bit overkill to begin with.

peppy:
I could also get behind that.

More-local generic specs should always be preferred as we have found time and time again that class level specs reduce flexibility and increase complexity.

bdach:
the aforementioned method signature did most of the work, although i did hit yet another snag (the existence of `CreateSpriteText()` at a `TextFlowContainer` level). i managed to get past that by creating a `DefaultTextChunk` that defers to that method of the text flow to create the sprite, rather than using `new TSpriteText()` as specified by the generic argument.

due to the size of the conflict this all landed in the merge commit. if this is deemed unacceptable then i can rebase this change on top of `master`  and force-push, or close this PR and reopen again based on current `master`.

i have not tested very much aside from running framework unit tests and smoke-screening game, but the latest changes do not impact the game-side changes in any way when it comes to compiling at least (no further API breakage is expected than already declared in the OP).

peppy (approved):

bdach:
so as it turns out my latest changes attempting to salvage this after #4729 got merged, broke markdown containers, in particular the wiki home page game-side. the reason is that in general text flow's API surface is uncontrollably large, and `TextFlowContainer.CreateSpriteText()` and `TextChunk.CreateSpriteText()` both try to control how the sprite text is created, and in this particular case the former takes priority incorrectly. this wasn't an issue before #4729 but is now that there are type arguments in play.

as a last ditch effort before either reverting #4729 or trying to reshape text flow more completely (probably with wider-ranging breaking changes), how unacceptable would you see deleting `DefaultTextChunk` and having something like this living in `TextChunk`?

```csharp
        private TSpriteText createSpriteText(TextFlowContainer textFlowContainer)
        {
            var spriteText = textFlowContainer.CreateSpriteText() as TSpriteText ?? new TSpriteText();
            textFlowContainer.ApplyDefaultCreationParamters(spriteText);
            creationParameters?.Invoke(spriteText);
            return spriteText;
        }
```

yes, i know, it's a little magic with regards to typing, but as i said, it's a last ditch effort to keep changes minimal. (an alternative would be to pass `Func<TSpriteText>` delegates to the chunk from the text flow, but that sucks more in my opinion due to the delegate allocations involved, which will get more frequent as the drawables get recreated when the locale changes.)

peppy:
@bdach ready to get this merged into a release, or still want more time for the osu-side changes?;

bdach:
this can't be merged as is unfortunately, as `CustomisableTextContainer` is broken with the latest changes after merging in #4729 here. still haven't figured out a way to fix that i'd be happy with.

bdach:
I've pushed 684eb43 as a fix for the customizable/markdown text container regression mentioned above. I still don't like it at all, due to introducing yet another delegate, but I don't have many other smart ideas how to simultaneously salvage most of the existing API surface, and apply this part-based direction for text flow.

If the aforementioned commit is deemed acceptable, I've also updated the [game-side branch](https://github.com/ppy/osu/compare/master...bdach:text-flow-parts?expand=1) which is as far as I can tell ready to go and which I can PR tomorrow to go along with this. I've tested it both manually and via a full test suite and didn't find any breakage.

peppy:
Pretty sure I already went through the new commits a while back, and am about as okay with them as I am with the rest of this PR. Maybe we should try and get this one in?

PR #4705, opened 2021-08-24
Reduce verbosity of STBI errors

smoogipoo:
Closes https://github.com/ppy/osu/issues/14417


peppy:
Just hoping we don't dig these so deep we regress loading for all images but don't find out until someone one day posts a mobile device log file or something...

peppy (approved):

Issue #2669, opened 2019-07-17
Android crash while idle


ticotaco72:
**Describe the crash:** opened osu, been idle for 2 minutes, and crash

**osu!lazer version:** 438a294

**Logs:** Android crash report
```
android.runtime.JavaProxyThrowable: System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
  at osuTK.FrameEventArgs.set_Time (System.Double value) [0x0000c] in <83afe94143fd4c7099c83433d9fcfcc4>:0 
  at osuTK.Android.AndroidGameView.RunIteration (System.Threading.CancellationToken token) [0x00045] in <7484b4753167456b82f11e78c0669d25>:0 
  at osuTK.Android.AndroidGameView+<>c__DisplayClass58_0.<StartThread>b__2 (System.Object _) [0x00000] in <7484b4753167456b82f11e78c0669d25>:0 
  at Android.App.SyncContext+<>c__DisplayClass3_0.<Send>b__0 () [0x00000] in <95d915721ad748108ff096dce8f8f32c>:0 
  at Java.Lang.Thread+RunnableImplementor.Run () [0x00008] in <95d915721ad748108ff096dce8f8f32c>:0 
  at Java.Lang.IRunnableInvoker.n_Run (System.IntPtr jnienv, System.IntPtr native__this) [0x00009] in <95d915721ad748108ff096dce8f8f32c>:0 
  at (wrapper dynamic-method) System.Object.19(intptr,intptr)
	at mono.java.lang.RunnableImplementor.n_run(Native Method)
	at mono.java.lang.RunnableImplementor.run(RunnableImplementor.java:30)
	at android.os.Handler.handleCallback(Handler.java:754)
	at android.os.Handler.dispatchMessage(Handler.java:95)
	at android.os.Looper.loop(Looper.java:163)
	at android.app.ActivityThread.main(ActivityThread.java:6238)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:794)
```

**Computer Specifications:** 
Xiaomi Redmi 5a
Android 7.1.2
Ram 2.0 GB
Processor 4x1.40ghz

ticotaco72:
Interesting

I reproduced it next time and I looked into logs and there was in performance.log this 2 sec before crash
[...]

PR #4706, opened 2021-08-24
Update osuTK again (with android build corrected)

peppy:
See discussion in https://github.com/ppy/osu-framework/issues/2669.

Not sure why android has even been working in the mean time, as the specified package seems to have not been available on nuget...

### DIFF
--- a/osu.Framework.iOS/osu.Framework.iOS.csproj
+++ b/osu.Framework.iOS/osu.Framework.iOS.csproj
@@ -24,7 +24,7 @@
     <ProjectReference Include="..\osu.Framework\osu.Framework.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="ppy.osuTK.iOS" Version="1.0.173" />
+    <PackageReference Include="ppy.osuTK.iOS" Version="1.0.178" />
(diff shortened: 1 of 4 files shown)


smoogipoo (dismissed):

peppy:
Hold off on this, I missed some packages. No wonder.

osu.Framework.iOS/osu.Framework.iOS.csproj
```diff
@@ -24,7 +24,7 @@
     <ProjectReference Include="..\osu.Framework\osu.Framework.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="ppy.osuTK.iOS" Version="1.0.173" />
+    <PackageReference Include="ppy.osuTK.iOS" Version="1.0.178" />
     <PackageReference Include="SixLabors.ImageSharp" Version="1.0.3" />
```
smoogipoo, line 28: Can this be merged into the `iOS.props` file?

peppy: Probably, but it's not in an easily testable state. Maybe best left for when we get iOS back on track?

smoogipoo (approved):

PR #4708, opened 2021-08-25
Call `SDL_StartTextInput` explicitly to receive `SDL_TEXTINPUT` events

ATiltedTree:
On most platforms `SDL_Init` [already calls `SDL_StartTextInput` for us](https://github.com/libsdl-org/SDL/blob/130b6bebaec4bc9f37c103a69595e8c261f2af49/src/video/SDL_video.c#L565-L574), but not on all.

For example the `wayland` backend of `SDL` does not do this and therefore this is required to get input in textboxes to work with `wayland`.

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -367,6 +367,9 @@ namespace osu.Framework.Platform
         public SDL2DesktopWindow()
         {
             SDL.SDL_Init(SDL.SDL_INIT_VIDEO | SDL.SDL_INIT_GAMECONTROLLER);
+            // On most platforms SDL_Init already calls this function for us, but not on all.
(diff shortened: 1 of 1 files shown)


peppy:
From the documentation: "On some platforms using this function activates the screen keyboard."

This probably needs to be done specifically within the context of [ITextInputSource](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Input/ITextInputSource.cs) if it is only required during textbox (and not general game) input.

bdach:
[From SDL docs](https://wiki.libsdl.org/SDL_StartTextInput):

> On some platforms using this function activates the screen keyboard.

So I'm not certain that calling this function once ever on startup is a good idea. This call should probably go into `GameWindowTextInput` (and be accompanied by corresponding `SDL_StopTextInput` calls, too).

PR #4709, opened 2021-08-26
Add SDL2 Text Input support

jai-x:
This is a proof of concept PR containing changes related to using the SDL2 Text Input event system intended to gather feedback and if this implementation should be continued.

### Implementation
This PR intends to support the SDL mechanism of text input via events that are not strictly bound to key inputs.
Ideally this should solve previous issues with special characters not appearing on key press and be more keyboard layout agnostic and eventually pave the way for SDL2 IME support.

Changes include:
* Appropriately handling the `SDL_TEXTINPUT` events in the `SDL2Window` class and invoking a new event for it.
* Making a breaking changes to `ITextInputSource` to be event based instead of polling to consume pending text
* Splitting `GameWindowTextInput` into specific classes to be used for SDL and OsuTK windows
* Refactoring parts of `TextBox` to make use of the next `ITextInputSource.OnTextInput` event instead of polling for text via `GetPendingText`

### Remarks
SDL2 does not differentiate between text that has been input via normal keyboard input vs text that is committed via an IME.
As such the changes to `ITextInputSource` reflect this by removing the concept of `ImeActive` to just `Active` and having a single `OnTextInput` event be the only way of getting text. This is actually more similar to the way text input is handled in the Android implementation in `AndroidTextInput` class.
See also: https://wiki.libsdl.org/Tutorials-TextInput

### Caveats
* Tested only on Linux
* While this should allow further support for IMEs via handling the `SDL_TEXTEDITING` events correctly, the API for that is less than ideal as it does not provided a way to see the candidate list. It would likely be a Linux only IME option to used for fallback in favour of better native API's


osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -179,101 +170,119 @@ public virtual bool OnPressed(PlatformAction action)
...
 
                 case PlatformAction.SelectForwardLine:
                     ExpandSelectionBy(GetForwardLineAmount());
-                    return true;
+                    handled = true;
+                    break;
             }
 
-            return false;
+            textInput?.Activate();
```
bdach, line 283: This cannot be right, surely? So any platform action activates text input, even if it isn't handled by the text box?

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -798,11 +767,9 @@ protected override bool OnKeyDown(KeyDownEvent e)
                     return false;
             }
 
-            return base.OnKeyDown(e) || consumingText;
+            return (textInput?.Active ?? false) || base.OnKeyDown(e);
```
bdach, line 770: would prefer to write
```suggestion
            return (textInput?.Active == true) || base.OnKeyDown(e);
```

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -990,12 +955,25 @@ protected override void OnFocus(FocusEvent e)
...
+            if (textInput == null)
+                return;
+
+            textInput.Activate();
+            textInput.OnTextInput += handleTextInput;
+        }
+
+        private void handleTextInput(string text)
+        {
+            Schedule(() => InsertString(text));
```
bdach, line 976: what is the reason for using a normal scheduler rather than `textUpdateScheduler` as the `Text` setter does?

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -469,7 +478,8 @@ private int getCharacterClosestTo(Vector2 pos)
 
         private void moveSelection(int offset, bool expand)
         {
-            if (textInput?.ImeActive == true) return;
+            if (textInput?.Active == true)
+                textInput.Deactivate();
```
bdach, line 482: can this be split to a helper at least if it's going to be called from every mouse event handler in this class?

osu.Framework/Input/SDL2DesktopWindowTextInput.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        public void Deactivate()
+        {
+            window.StopTextInput();
+            active = false;
+        }
+
+        public void EnsureActivated()
+        { }
```
bdach, line 44: why is `EnsureActivated()` a no-op here? surely it should call `Activate()` if not active? from what I recall this method was supposed to pull up a software keyboard on platforms where one is supported, which is incidentally what `SDL_StartTextInput()` also does.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -853,14 +857,14 @@ private unsafe void handleTextInputEvent(SDL.SDL_TextInputEvent evtText)
             if (ptr == IntPtr.Zero)
                 return;
 
-            string text = Marshal.PtrToStringUTF8(ptr) ?? "";
+            string text = Marshal.PtrToStringUTF8(ptr) ?? String.Empty;
```
bdach, line 860:
```suggestion
            string text = Marshal.PtrToStringUTF8(ptr) ?? string.Empty;
```

osu.Framework/Platform/MacOS/MacOSGameHost.cs
```diff
@@ -44,7 +44,19 @@ public override string UserStoragePath
...
+        public override ITextInputSource GetTextInput()
+        {
+            if (Window == null)
+                return null;
+
+            if (Window is SDL2DesktopWindow)
+                return new SDL2DesktopWindowTextInput(Window as SDL2DesktopWindow);
+
+            if (Window is OsuTKWindow)
+                return new MacOSOsuTKTextInput(Window as OsuTKWindow);
```
bdach, line 56: pretty sure no desktop platform uses osuTK anymore, so this is essentially dead/unneeded code

bdach (changes requested):
Aside from the code quality inspections as per CI:

osu.Framework/Graphics/UserInterface/TextBox.cs
jai-x: The implication here is that the textinput should always be active for the textbox unless it is readonly. The handling of a platform action should temporarily disable it so as not to interfere and then be reenabled once the action has been handled.

osu.Framework/Graphics/UserInterface/TextBox.cs
jai-x: Mistake. I will make sure to use that scheduler as well as add the `OnUserTextAdded` function I seemed to have missed.

osu.Framework/Graphics/UserInterface/TextBox.cs
jai-x: Makes sense

osu.Framework/Input/SDL2DesktopWindowTextInput.cs
jai-x: I suppose I could also get rid of this method from the `ITextInputSource` interface since it's no longer needed.

osu.Framework/Platform/MacOS/MacOSGameHost.cs
jai-x: I was under the impression that you could still run the desktop with osuTK using the `--tk` flag. If that is no longer the case I can remove the osuTK specific keyboard handling code.

osu.Framework/Graphics/UserInterface/TextBox.cs
bdach: That makes some sense, but (a) that should probably be accompanied by an inline comment, and (b) won't deactivating text input cause the IME box/software keyboard/whatever to flicker for a split second? Are non-printable characters/modifier key combinations at risk of being swallowed by the ongoing text input?

osu.Framework/Input/SDL2DesktopWindowTextInput.cs
Susko3: `EnsureActivated` is still needed. Currently, it's used to show the keyboard on mobile platforms when the user manually closed it (bypassing `Deactivate()`, so framework thinks the keyboard is shown, and doesn't show it until a new textbox is activated.)

`EnsureActivated` should call `window.StartTextInput()` here to ensure the software keyboard is shown in case the user has manually hidden/closed it.

osu.Framework/Platform/MacOS/MacOSGameHost.cs
bdach: hm, i suppose there is that option still, i forgot - my bad

osu.Framework/Input/SDL2DesktopWindowTextInput.cs
bdach: No, it cannot be removed, I'm pretty sure. The mobile implementations use it to call native iOS/Android functions to pull up the software keyboard if it has been manually dismissed by the user via the back button or what have you.

osu.Framework/Graphics/UserInterface/TextBox.cs
jai-x: Yes the current IME candidate text would be disabled on a platform action, but this is intentional and falls in line with expectations for how I have used IMEs before on desktop platforms. You could consider the IME candidate text buffer ephemeral and separate from the actual textbox, to be dismissed when edit actions are performed on the actual textbox. This is somewhat less ideal for mobile platforms or platforms using screen keyboards as it would temporarily dismiss the on screen keyboard and SDL does not have a separate API for this. I would guess the actual user facing impact of this would not be too visible as SDL is not used on mobile platforms currently.

bdach: the fact that IME and on-screen keyboards are conjoined together so seems like a... rather major misdesign, if you ask me. but if nothing better can be done...

I suppose for now most of the impact of this would be constrained to mobile users that use keyboards. which is relatively niche but not outright insignificant i feel?

jai-x:
Aside from the code quality review given, is there any general feedback for the way I have implemented this so far? (i.e breaking changes made to the `ITextInputSource`)
If this is good then I can work to fix the code problems and update the mobile textinput source for the new interface to put this PR in a more mergable state.

bdach:
The `ITextInputSource` changes look fine to me at least, although as stated `EnsureActivated()` has to remain.

osu.Framework/Graphics/UserInterface/TextBox.cs
jai-x: For reasons I don't fully understand, wrapping the lamda function in `textUpdateScheduler.Add` doesn't seem to work? I would assume that since that scheduler is updated in `UpdateClock` it would run on every frame but I guess not?
I also noticed that `textUpdateScheduler.Update` is manually called within the context of `updateCursorAndLayout` which made me think to call `cursorAndLayout.Invalidate()` after the schedule but that had weird side effects where the text would be updated but the cursor would remain in place :/

jai-x:
In addition to the suggestions made above, I have also updated the iOS and Android text input source classes to match the changes to `ITextInputSource`. At the moment I don't have a proper setup for doing Android builds (and iOS is broken) so I've made these changes somewhat blindly but hopefully in a sane way.

osu.Framework.Android/Input/AndroidTextInput.cs
```diff
@@ -25,44 +27,18 @@ public AndroidTextInput(AndroidGameView view)
...
+            if (Active)
+                OnTextInput?.Invoke(text);
         }
 
         private void keyDown(Keycode arg, KeyEvent e)
         {
-            if (e.UnicodeChar != 0)
-                pending += (char)e.UnicodeChar;
+            if (Active && e.UnicodeChar != 0)
+                OnTextInput?.Invoke((char)e.UnicodeChar.ToString());
```
Susko3, line 39:
```suggestion
                OnTextInput?.Invoke(((char)e.UnicodeChar).ToString());
```

Susko3 (changes requested):
I'm not too keen on deactivating and activating the `TextInput` when handling `PlatformAction`s. As it creates a really bad experience on mobile; the keyboard quickly hides and appears again. Since deleting characters is a `PlatformAction`, pressing backspace on the on-screen keyboard is a really bad experience.

bdach:
It seems to me that the differences in what is expected from `ITextInputSource` in the IME case and in the on-screen keyboard case are insurmountable to the point that short of asking upstream to split these cases, we will likely have to have a desktop implementation that does what the IME case expects, and a mobile implementation that does what the on-screen keyboard case expects...

jai-x:
I'm inclined to agree with you on this. Given the requirements gathers from this PoC PR I think you would need some system that:
* Enables text input on textbox focus and shows screen keyboard for mobile platforms
* Disables text input on textbox unfocus and hides screen keyboard on mobile platforms
* Halts text input events when handling non-text input events without hiding the screen keyboard
* Ensures that halting text input will also dismiss any active IME candidate window without dismissing the screen keyboard on mobile platforms

While I think mobile platforms can do this with their native API's, the current SDL API for desktop is woefully inadequate for this, which is surprising given it uses proper IME integration in it's implementation (Windows `imm.h`, Linux `iBus`) but only exposes the `Start/Stop` events and `TEXT_INPUT` and `TEXT_EDITING` events.

bdach:
It wouldn't be a first to bypass sdl and roll our own (see manual `WM_INPUT` handling for high precision/raw mouse on windows). Obviously it would be ideal to go through it but if that is not viable then I don't think it'd be entirely out of the question.

jai-x:
After a bunch of manual testing I'm reopening this PR with the many `ITextInputSource.Deactivate` calls removed as they weren't needed. This comes with the he caveat that this is *not* handling the `TEXT_EDITIING` events for IME support so this is just somewhat improved text handling that is input based.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -899,7 +859,7 @@ protected override bool OnDragStart(DragStartEvent e)
 
         protected override bool OnDoubleClick(DoubleClickEvent e)
         {
-            if (textInput?.ImeActive == true) return true;
+            deactivateInput();
```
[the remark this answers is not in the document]
jai-x: Removed as not needed

osu.Framework/Input/ITextInputSource.cs
```diff
@@ -11,21 +11,18 @@ namespace osu.Framework.Input
...
 
+        void Deactivate();
+
         /// <summary>
         /// Ensures that the native implementation that retrieves user text input is activated
         /// and that the user can start entering text.
         /// </summary>
         void EnsureActivated();
-
-        event Action<string> OnNewImeComposition;
```
bdach, line 28: This will most likely have to return with an IME implementation (was supposed to be fired when IME text is being "previewed" but not yet committed to the textbox). So I'd leave this be (maybe renamed, if anything), because this does pretty much nothing as is anyway, as far as I can see - all codepaths that invoke this are pretty much dead ends.

bdach:
The code looks generally sane and I don't expect it to cause issues but I'll have to run it through the wringer of testing on platforms that I can test on (windows, linux, android) before stating that with any certainty. Someone else would have to grab macOS/iOS.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -730,57 +713,16 @@ public virtual string Text
...
-                OnUserTextAdded(pendingText);
-            }
-
-            if (consumingText)
-                Schedule(consumePendingText);
-        }
-
         #region Input event handling
 
         protected override bool OnKeyDown(KeyDownEvent e)
```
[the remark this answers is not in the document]
jai-x: I felt that this approach was the most logical as any keypress that isn't explicitly handled framework side may be used as part of a text composition so it is treated as such. Given it's now correctly using the text input events from SDL, it may be prudent to remove the `(Control || Alt || Super)` key being pressed but I haven't tested to see if that behaviour is as expected yet,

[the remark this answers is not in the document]
bdach: Thing is that if a text input is accepted by a text box, the key input that caused it probably shouldn't be able to propagate to any other drawable, which seems to be the biggest issue here given that they're things that are handled on pretty much disparate execution paths which don't look designed to guarantee any ordering guarantees (i.e. even if SDL will emit a text input before a key input, will it continue to work that way in the future? what if some other windowing backend is used? etc. etc.)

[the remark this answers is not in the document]
jai-x: I'm confused? Doesn't the current behaviour already stop the keypress propagating to other drawables if the the `ITextInput` is active? Have I got the boolean return the wrong way round?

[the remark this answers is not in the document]
bdach: I'm not saying that, but that check comes dead last after these early-returns, doesn't it? As written a textbox won't handle *any* key event if ctrl/super/alt are pressed, won't it? Which is wrong because ctrl/alt being pressed is no indication of whether the keypress has produced a character. That strongly depends on the keyboard layout and other similar factors.

For example, Polish uses accented letters. One example is `ę` which uses <kbd>AltGr</kbd>+<kbd>E</kbd>, wherein SDL encodes <kbd>AltGr</kbd> as <kbd>Ctrl</kbd>+<kbd>Alt</kbd> because legacy reasons. That produces a letter. But a keypress such as <kbd>Alt</kbd>+<kbd>F4</kbd> does not produce text and should never be handled as text, otherwise it will stop all sorts of global keybindings from working.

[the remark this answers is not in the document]
jai-x: I very much agree, which is why I proposed the removal of the (ctrl/super/alt) check in the earlier comment. As far as I can tell, the only distinct keypresses the textbox should care about are the ones already handled in the switch case and everything else blocked from propagating when text input is active as being assumed handled (other events like Alt+F4 generating a close event are already handled at a higher level I think).

Do you think after removing the ctrl/super/alt check the behaviour would be fine as is?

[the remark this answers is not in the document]
bdach: I don't think it's as simple as removing the check. Although I haven't tested, I believe that because input events are processed from the frontmost drawable and propagated front-to-back, you will have things like `Page{Up,Down}` scrolling not working if you've focused a textbox (if not if one is present at all).

osu.Framework/Input/ITextInputSource.cs
jai-x: Re-added

jai-x:
I've pushed commits to address most of the suggestions by @Susko3.

Regarding the way that an `ITextInput` instance binds to the window events, I'm of the belief that a game host should only have a single instance of an `ITextInput` instead of an instance being created per textbox as it reflects better on how the events are handled and tied to the host (similar to how there's only one instance of InputHandlers per host). I felt that making this change now would make the PR too large and it can be made after this is merged.

As for the remaining issues:
* I currently don't understand the weirdness with the `textUpdateScheduler` within `TextBox`. Making use of it leads to unexpected results with how the text is inserted whereas simply using the regular `Schedule` leads to how I expect the text input events to functions.
* As discussed above, blocking of certain keys in `TextBox.OnKeyDown` may be problematic with it interrupting certain platform actions. I'm in favour of just removing the Ctrl/Super/Alt check as that seems "Fine:tm:" from some limited local testing but still not sure on a proper solution.

I'm unsure on how to proceed on the above two issues.

frenzibyte:
> * I currently don't understand the weirdness with the `textUpdateScheduler` within `TextBox`. Making use of it leads to unexpected results with how the text is inserted whereas simply using the regular `Schedule` leads to how I expect the text input events to functions.

Well that special scheduler has just been removed from `TextBox` in https://github.com/ppy/osu-framework/pull/4522, simplifying the flow and resolving a couple of issues caused by it specifically.

osu.Framework/Graphics/UserInterface/TextBox.cs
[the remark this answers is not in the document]
jai-x: The behaviour of textboxes blocking events such as PageUp/PageDown seems correct to me?
Using html `<input type="text">` as an example most users would be familar with, all key events aside from ones that change focus seem to be handled/blocked by the textbox so the behaviour seems expected to me or at least has precident.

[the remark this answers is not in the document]
frenzibyte: I may be misunderstanding the code, but judging from your comment, wouldn't that mean any platform or global action would be blocked as well? and that's the reason why all modifier keys return `false` and not be handled by the `TextBox`?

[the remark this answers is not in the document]
bdach: That is another worry of mine yes. Although I did not initially give that example because from what I recall key binding containers have a "priority" mode that inverts input event processing order in their favour and so that may still continue to work wherever that is specified.

peppy:
Is it feasible to conflict resolve this PR post-#4865?

smoogipoo:
Superseded by https://github.com/ppy/osu-framework/pull/4985

PR #4728, opened 2021-08-27
Fix `FrameTimeDisplay` delegate leakage

peppy:
Until now, opening the `FrameTimeDisplay` once and then hiding it would leave it in a counting state in the background. While this isn't immediately cause for concern, it was also scheduling one delegate per pending update. These could build up over time and would only be freed on the next display of the display:

20210827 171244 (Parallels Desktop)

I also improved the animation behaviour to avoid it feels like sometimes it could wouldn't restore to the correct location. This means querying `DrawSize` more often, but I think it is implicitly validated in the normal draw hierarchy process so should not add overhead.


osu.Framework/Graphics/Performance/FrameTimeDisplay.cs
```diff
@@ -58,53 +55,51 @@ public FrameTimeDisplay(ThrottledFrameClock clock, GameThread thread)
...
+                ClearTransforms();
 
-                double clockFps = clock.FramesPerSecond;
-                double updateHz = clock.MaximumUpdateHz;
+                if (aimWidth == 0)
+                    Size = counter.DrawSize;
+                else if (Precision.AlmostBigger(counter.DrawWidth, aimWidth))
+                    this.ResizeTo(counter.DrawSize, 200, Easing.OutQuint);
+                else
+                    this.Delay(500).ResizeTo(counter.DrawSize, 200, Easing.InOutSine);
```
bdach, line 76: huge nit but was changing only one of these easings intentional?

bdach (commented):
all for this change, especially because it gets rid of nested schedules

osu.Framework/Graphics/Performance/FrameTimeDisplay.cs
peppy: Yeah, just changed this to what felt right.

bdach (approved):

smoogipoo (approved):

PR #4729, opened 2021-08-27
Add type arguments to `CustomizableTextContainer` and `TextFlowContainer`

### LINKED ISSUE #4721: TextFlowContainer should have type of `SpriteText` as a generic parameter
Use Case: Allow derived `SpriteText` classes to be an argument in many methods that take in `creationParameters`.

Example Case
```csharp
class MySpriteText : SpriteText
{
    bool MyProperty { get; set; }
}

class MyTextFlowContainer : TextFlowContainer<MySpriteText>
{
    protected override MySpriteText CreateSpriteText() => new MySpriteText()
}

var flow = new MyTextFlowContainer();
flow.AddText("Hello World", s => s.MyProperty = true);
```

LeNitrous:
Resolves #4721.


smoogipoo (approved):
I was gonna say an alternative would be to have `T : SpriteText, new()`, but I suppose this is fine and nothing to really dwell on for now.

PR #4734, opened 2021-08-28
Exclude non-breaking space from fixed width in sprite texts by default

bdach:
Closes https://github.com/ppy/osu/issues/14523.

| before | after |
| :-: | :-: |
| 2021-08-28-174949_749x171_scrot | 2021-08-28-175118_746x168_scrot |
| 2021-08-28-175001_327x160_scrot | 2021-08-28-175128_289x154_scrot |

I would have included a test case for this but the roboto font actually does not have a glyph generated for `U+00A0` using bmfont it seems, so it shows up as `?` in text which looks broken. I can go and fix this but I'm not sure if worth.


peppy (approved):
Almost feels like this should be a framework side ch--- oh this is framework 😅 .

Loreos7:
Looks kinda readable now. Thx for the fix.

PR #4735, opened 2021-08-28
Refactor tooltip test scene to assert behaviour rather than visual only

frenzibyte:
For the ability to add further test scenarios on future changes (one in which is upcoming), and also to ensure nothing regresses along the way.


osu.Framework.Tests/Visual/UserInterface/TestSceneTooltip.cs
```diff
@@ -1,58 +1,150 @@
...
+            ITooltip originalInstance = null;
+
+            hoverTooltipProvider(() => tooltipText);
+
+            AddStep("get tooltip instance", () => originalInstance = tooltipContainer.CurrentTooltip);
+            assertTooltipText(() => tooltipText.TooltipText);
+
+            hoverTooltipProvider(() => tooltipText);
+
+            AddAssert("tooltip reused", () => tooltipContainer.CurrentTooltip == originalInstance);
```
bdach, line 50: The overall test coverage is good but I'm not super convinced that the implementation details of when the instances get or don't get reused should be tested? I guess from a performance testing standpoint that is handy but it also feels like testing too much implementation details rather than just the desired end-user behaviour.

bdach:
Seems okay other than the above (which I'd be interested in more opinions on). I've also made some tiny wording cleanups while I was browsing.

osu.Framework.Tests/Visual/UserInterface/TestSceneTooltip.cs
smoogipoo: I totally get what you're saying, but rather than testing performance, isn't this testing correctness? Like osu!-side we have tooltips transforming to their new position/size on reuse, which wouldn't work if this breaks.

How do you test the end-user behaviour otherwise in this case?

smoogipoo (approved):
This looks okay to me

peppy (approved):

osu.Framework.Tests/Visual/UserInterface/TestSceneTooltip.cs
bdach: I guess if this is considered a behaviour that the framework is more or less explicitly providing to the consumer (i.e. tooltip reuse is part of the "contract" of the container, so to speak), then this is fine to test. I haven't looked at the next pull in the series though because apparently there are some upcoming changes to that there.

PR #4736, opened 2021-08-28
Refactor tooltips reusability logic to be dependent on the type and simplify for use cases

frenzibyte:
- [x] Depends on #4735

Upon looking through resolving ppy/osu#14531, I've stumbled across this edge case on tooltips reusability, where if two tooltips of different types with different implementation accept the same content type, one might be reused in place of the other, causing wrong tooltips appearing.

I've pushed a workaround for that in the PR ppy/osu#14533, but I decided to look into this edge case framework-side as well.

Now I've refactored the tooltips reusability logic to instead check for the type of the tooltips themselves, and only reuse the current tooltip if it matches the type, removing the need for having a boolean return value in `SetContent`.

This does it at the cost of creating tooltips everytime the hovered target changes, should be pretty negligible given that nothing is loaded and only happens when moving between targets, but I admit I haven't benchmarked that. 

I've also added generic variants of `IHasCustomTooltip` and `ITooltip` specifying the content type, to further simplify usage cases which now don't need to return a value on non-matching content type.

Here are the changes applied osu!-side for reference: https://github.com/ppy/osu/compare/master...frenzibyte:cleanup-tooltips (along with the removal and revert of the `TargetContentType` workaround in ppy/osu#14533).

# vNext

## `ITooltip.SetContent` no longer require returning `bool` value

Until now, the method of reusing tooltip instances was problematic (two tooltips handling the same data type could not exist). Instance sharing is now based on the constructed tooltip's `Type`, rather than the data type.

An example of how you should update your code follows.

Before:

```csharp
public override bool SetContent(object content)
{
    if (!(content is CustomContent custom))
        return false;

    text.Text = content.ToString(); // whatever you need to do here.
    return true;
}
```

After:

```csharp
public override void SetContent(object content)
{
    text.Text = content.ToString(); // whatever you need to do here.
}
```


peppy:
I'm okay with this change, but this PR does feel a bit overloaded with generic method additions (which are not required at all by the change/fix here). I guess it's okay.

But you need to rewrite the breaking change method. It reads horribly. You need to be much more specific and to the point. Something like "Until now, the method of reusing tooltip instances was problematic (two tooltips handling the same data type could not exist). Instance sharing is now based on the tooltip type, rather than the data type."

Also probably show an example of the required code change (simply replacing `bool` with `void`).

@smoogipoo requesting your eyes to make sure you're fine with the two changes.

peppy (approved):

smoogipoo (approved):
Code-wise looks fine to me. Agree with rewording the breaking change.

peppy (approved):

PR #4739, opened 2021-08-29
Improve code quality of `.ToLocalisableString()` extension

frenzibyte:
- Marks `IFormattable` in the `.ToLocalisableString()` extension method as non-nullable (via `#nullable enable`), just so that usages don't potentially fall into a nullref situation due to implicit conversions from `Nullable<T>` types to `IFormattable` (e.g. ppy/osu#14541 and ppy/osu#14562).

  Now usages get a clear warning if that potentially happens:
  CleanShot 2021-08-29 at 13 56 53

- Allows using the `.ToLocalisableString()` extension method without specifying a format string, by marking that as an optional parameter with the default being `null`, just like how [number](https://github.com/dotnet/runtime/blob/cffaa78235ea93d5e3eeb56956579df503e11250/src/libraries/System.Private.CoreLib/src/System/Double.cs#L265-L273) [structures](https://github.com/dotnet/runtime/blob/cffaa78235ea93d5e3eeb56956579df503e11250/src/libraries/System.Private.CoreLib/src/System/Int32.cs#L88-L106) implement their `ToString` methods, which I've checked from curiosity.

- Removes the wrongly placed duplicate extension class and migrates the `.ToLocalisableString()` to the right class.


bdach:
The [`IFormattable` docs](https://docs.microsoft.com/en-us/dotnet/api/system.iformattable?view=net-5.0#notes-to-implementers) do state that implementors must be prepared to handle a `null` format, so that particular change seems safe to proceed with.

That said, do you have upcoming changes for updating the game-side usages of `ToLocalisableString()`? The namespace changed, making this a breaking change.

frenzibyte:
Yep, had to go for a bit. I'll PR it now.

bdach (approved):

peppy (approved):

PR #4743, opened 2021-09-01
Fix `GetBackwardWordAmount()` going back two words if the last word is only one character long

### LINKED ISSUE #4742: Ctrl+Backspace in TextBox deletes two words if last word is one character long
1. image
2. Press <kbd>Ctrl</kbd>+<kbd>Backspace</kbd>
3. image

I'd expect the text to be "this is a text "

<kbd>Ctrl</kbd>+<kbd>Del</kbd> is unaffected by this bug.

Susko3:
Fixes #4742 

I've tested this and it works fine.
Not sure if tests should be added for this.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
@@ -331,7 +331,68 @@ namespace osu.Framework.Tests.Visual.UserInterface
         }
 
         [Test]
+        public void TestPreviousWordDeletionWithShortWords()
+        {
+            InsertableTextBox textBox = null;
+
+            AddStep("add textbox", () =>
+            {
+                textBoxes.Add(textBox = new InsertableTextBox
+                {
+                    Size = new Vector2(200, 40),
+                });
+            });
+
+            AddStep("click on textbox", () =>
+            {
+                InputManager.MoveMouseTo(textBox);
+                InputManager.Click(MouseButton.Left);
+            });
+
+            AddStep("insert three words", () => textBox.InsertString("a b c"));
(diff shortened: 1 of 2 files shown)


bdach:
Please do add tests if it is viable. I believe doing so should be relatively straightforward, as I recall tests have been added in the past for other textbox platform action failures, so the road should be paved.

Susko3:
Closest TestScene is [TestSceneTextBoxEvents](https://github.com/ppy/osu-framework/blob/master/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBoxEvents.cs), but this tests events, not test text editing.
TestSceneTextBoxEvents doesn't offer a simple way to: put text in textbox, perform action, check if the text looks the way it should.

A new test scene would be in order, but I don't feel like adding that and testing all text editing events right now.

bdach:
That isn't really true. See `TestSceneTextBox.TestNextWordDeletion()` for one example.

Susko3:
Have added tests now. `TestSceneTextBox.TestPreviousWordDeletionWithShortWords()` fails on master.

bdach:
I've stared at this a bit and I'm reasonably certain that this change is correct, despite the old `- 2` code being quite old (pre-2016!). Index math is confusing as ever, especially that the selection position has fencepost error "built in", but it does make sense to start searching from index `i - 1` backwards and index `i` forwards, as I conceived by going off the following demonstrative example:

```
selection indexing    0 1 2 3 4 5 6 7 8
actual string          A B _ A|_ A B C
character indexing     0 1 2 3 4 5 6 7
```

bdach (approved):

smoogipoo (approved):

PR #4749, opened 2021-09-02
Reduce video decode overhead after reaching end of stream

peppy:
Without this change, on reaching the end of a video the decode loop would spin on the `av_read_frame` loop, with no `sleep`. While this in general is not super-expensive, it's also not free, and can result in around 5-10% of a single core being used when video playback is no longer relevant.

Note that this doesn't occur during actual playback/pause (ie. before reaching the end of stream) due to the frame render count being upheld.

Closes https://github.com/ppy/osu/issues/14548, which is an edge case that performance overhead *is* a concern due to the video file throwing an error each `av_read_frame` in this state. With this PR, the error is no longer thrown. This means I can now playback the video mentioned in that issue without 100% cpu usage, and also without rider crashing immediately on the video ending (likely due to console spam).

If you want to manually test this fix, update `TestVideo` to consume a local stream:

```csharp
        private static Stream consumeVideoStream()
        {
            return File.OpenRead("/Users/Dean/video.avi");
        }
```

And point to the extracted video from the beatmap mentioned in the linked issue.

Note that there is a separate issue where after reaching the end of stream, a video will likely never play again even after seeking into the past. This seems to be an issue on `master` as well, which I'll look at fixing separately. Safe to say it didn't regress in this PR.

### DIFF
--- a/osu.Framework/Graphics/Video/VideoDecoder.cs
+++ b/osu.Framework/Graphics/Video/VideoDecoder.cs
@@ -252,7 +252,7 @@ namespace osu.Framework.Graphics.Video
         }
 
         [MonoPInvokeCallback(typeof(avio_alloc_context_seek))]
-        private static long seek(void* opaque, long offset, int whence)
+        private static long streamSeekCallbacks(void* opaque, long offset, int whence)
         {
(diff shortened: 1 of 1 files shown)


bdach (approved):
looks fine aside from the above

bdach (approved):

PR #4750, opened 2021-09-02
Fix video stopping playback permanent when reaching end of stream

peppy:
- [x] Depends on #4749 (mostly for mergeability).

### DIFF
--- a/osu.Framework/Graphics/Video/Video.cs
+++ b/osu.Framework/Graphics/Video/Video.cs
@@ -117,6 +117,16 @@ namespace osu.Framework.Graphics.Video
         {
             base.Update();
 
+            if (decoder.State == VideoDecoder.DecoderState.EndOfStream)
+            {
+                // if at the end of the stream but our playback enters a valid time region again, a seek operation is required to get the decoder back on track.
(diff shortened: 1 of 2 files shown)


bdach (dismissed):

osu.Framework/Graphics/Video/Video.cs
```diff
@@ -117,6 +117,16 @@ protected override void Update()
         {
             base.Update();
 
+            if (decoder.State == VideoDecoder.DecoderState.EndOfStream)
+            {
+                // if at the end of the stream but our playback enters a valid time region again, a seek operation is required to get the decoder back on track.
+                if (PlaybackPosition < decoder.Duration)
+                {
+                    decoder.Seek(PlaybackPosition);
+                    availableFrames.Clear();
```
bdach, line 126: On a second look, isn't this leaking `availableFrames` if there are potentially any left? Shouldn't they get returned to the decoder again?

peppy: Yeah good catch. A bit scary this isn't guarded against.

bdach (approved):

PR #4751, opened 2021-09-02
Fix video decode running before start time

peppy:
Noticed that we were continuously decoding frames before a video started due to no checks in place.

- [x] Depends on https://github.com/ppy/osu-framework/pull/4750 (for mergeability)

Note that I did some code quality improvements around the `outOfSync` logic but the only part which matters is the early return.

### DIFF
--- a/osu.Framework/Graphics/Video/Video.cs
+++ b/osu.Framework/Graphics/Video/Video.cs
@@ -127,11 +127,6 @@ namespace osu.Framework.Graphics.Video
                 }
             }
 
-            // if not looping and not yet having reached the start of the video, don't attempt to consume frames yet.
-            // if we begin consuming frames the decoder will be performing unnecessary work.
-            if (!Loop && PlaybackPosition < 0)
-                return;
-
             var peekFrame = availableFrames.Count > 0 ? availableFrames.Peek() : null;
             bool outOfSync = false;
 
--- a/osu.Framework/Graphics/Video/Video.cs
+++ b/osu.Framework/Graphics/Video/Video.cs
@@ -127,24 +127,33 @@ namespace osu.Framework.Graphics.Video
                 }
             }
 
-            var nextFrame = availableFrames.Count > 0 ? availableFrames.Peek() : null;
+            // if not looping and not yet having reached the start of the video, don't attempt to consume frames yet.
+            // if we begin consuming frames the decoder will be performing unnecessary work.
+            if (!Loop && PlaybackPosition < 0)
+                return;
 
-            if (nextFrame != null)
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/Video/Video.cs
```diff
@@ -127,26 +127,35 @@ protected override void Update()
                 }
             }
 
-            var nextFrame = availableFrames.Count > 0 ? availableFrames.Peek() : null;
+            // if not looping and not yet having reached the start of the video, don't attempt to consume frames yet.
+            // if we begin consuming frames the decoder will be performing unnecessary work.
+            if (!Loop && PlaybackPosition < 0)
```
bdach, line 132: I assume the loop exception here is to make sure that even with negative time the video can loop. Which would be fine, aside from the fact that as far as I can see, looping in negative time is completely busted right now.

With this branch as-is, when I run the `TestDecodingContinuesBeforeStartTimeIfLooping` I get a bunch of `Video too far out of sync` log messages. Which is because of this code in `AnimationClockComposite`:

https://github.com/ppy/osu-framework/blob/f2319e8aa8fb0015403241c15a86a5fc5788e695/osu.Framework/Graphics/Animations/AnimationClockComposite.cs#L86-L87

The `%` is, unfortunately, the nice math modulo (which guarantees that the result is in the interval [0, $right_operand]). It can be a _signed_  value. In particular, e.g. `-1000 % 600 = -400`. Which cannot be seeked to as there are physically no frames with that time.

But even with that fixed the animation doesn't play properly for reasons that I'm not entirely sure of yet (still spews `Video too far out of sync`, albeit at a lesser rate). But, the bigger point is, if `PlaybackPosition` was properly fixed to never return anything outside of the [0, $duration] range, this guard wouldn't be needed anymore, I don't think?

peppy: weird, that didn't happen when i tested.

peppy: Actually based on how things are written, I don't think looping before zero time is even correct. I'll update the logic to align with this expectation.

peppy: I've updated `PlaybackPosition` to be clamped as proposed. Seems to work well, and avoid the negative effect of the early return (potentially not buffering the initial frames of the video as early as possible).

bdach (dismissed):
I can get behind not looping before time 0, but I'll wait for opinions from others as to whether that's a behaviour that makes sense.

bdach (approved):

smoogipoo (approved):

PR #4756, opened 2021-09-05
Let `TabControl` always sort initial selection if auto sort is on

LittleEndu:
`TabControl` was relying on `TabContainer` to have correct layout for the initial autosort, but the layout isn't correct yet and all children in `TabContainer` have y=0. 

`y==0` is used by `TabItem` for its `IsPresent` which is checked by `selectTab()` when considering whether to sort the new selection or not.

If the tab is present then it isn't sorted, but the tab might not be present if the initial selected tab is changed when constructing the `TabControl`. The result was the initial selection being hidden in the dropdown.

This PR calls `performTabSort()` directly in `LoadComplete` so the initial selection is always auto-sorted.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs
@@ -204,6 +204,9 @@ namespace osu.Framework.Graphics.UserInterface
                 Current.Value = Items.First();
 
             Current.BindValueChanged(v => selectTab(v.NewValue != null ? tabMap[v.NewValue] : null), true);
+            // TabContainer doesn't have valid layout yet, so TabItems all have y=0 and selectTab() didn't call performTabSort() so we call it here instead
+            if (AutoSort && Current.Value != null)
+                performTabSort(tabMap[Current.Value]);
         }
 
         /// <summary>
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs
@@ -204,6 +204,7 @@ namespace osu.Framework.Graphics.UserInterface
                 Current.Value = Items.First();
 
             Current.BindValueChanged(v => selectTab(v.NewValue != null ? tabMap[v.NewValue] : null), true);
+
             // TabContainer doesn't have valid layout yet, so TabItems all have y=0 and selectTab() didn't call performTabSort() so we call it here instead
             if (AutoSort && Current.Value != null)
                 performTabSort(tabMap[Current.Value]);
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneTabControl.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneTabControl.cs
@@ -378,6 +378,30 @@ namespace osu.Framework.Tests.Visual.UserInterface
             AddAssert("initial selection is correct", () => tabControl.Current.Value == expectedInitialSelection);
         }
 
+        [TestCase(true, TestEnum.Test1, true)]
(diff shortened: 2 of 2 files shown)


osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -204,6 +204,9 @@ protected override void LoadComplete()
                 Current.Value = Items.First();
 
             Current.BindValueChanged(v => selectTab(v.NewValue != null ? tabMap[v.NewValue] : null), true);
+            // TabContainer doesn't have valid layout yet, so TabItems all have y=0 and selectTab() didn't call performTabSort() so we call it here instead
+            if (AutoSort && Current.Value != null)
+                performTabSort(tabMap[Current.Value]);
```
bdach, line 209: I think it would be *slightly* nicer if this happened in `performTabSort`, via a `firstSortExecuted` field or otherwise.

That said all of this component relying on `Y == 0` to tell if a tab isn't hidden in the overflow menu isn't all that great and it would be better as an actual property, albeit that's maybe a little out of scope here...?

LittleEndu: You mean this should be moved to `selectTab()` ???

```diff
Index: osu.Framework/Graphics/UserInterface/TabControl.cs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/osu.Framework/Graphics/UserInterface/TabControl.cs b/osu.Framework/Graphics/UserInterface/TabControl.cs
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs	(revision 2d2289a08763b60aba2448606d359375c6a70d6b)
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs	(date 1630849629387)
@@ -152,6 +152,7 @@
         private readonly Dictionary<T, TabItem<T>> tabMap = new Dictionary<T, TabItem<T>>();
 
         private bool firstSelection = true;
+        private bool firstSort = true;
 
         /// <summary>
         /// Creates a new <see cref="TabControl{T}"/>.
@@ -204,9 +205,6 @@
                 Current.Value = Items.First();
 
             Current.BindValueChanged(v => selectTab(v.NewValue != null ? tabMap[v.NewValue] : null), true);
-            // TabContainer doesn't have valid layout yet, so TabItems all have y=0 and selectTab() didn't call performTabSort() so we call it here instead
-            if (AutoSort && Current.Value != null)
-                performTabSort(tabMap[Current.Value]);
         }
 
         /// <summary>
@@ -351,7 +349,8 @@
         private void selectTab(TabItem<T> tab)
         {
             // Only reorder if not pinned and not showing
-            if (AutoSort && tab != null && !tab.IsPresent && !tab.Pinned)
+            // but always show current selection in front if this is first sort
+            if (AutoSort && tab != null && ((!tab.IsPresent && !tab.Pinned) || firstSort))
                 performTabSort(tab);
 
             // Deactivate previously selected tab
@@ -398,6 +397,7 @@
 
         private void performTabSort(TabItem<T> tab)
         {
+            firstSort = false;
             TabContainer.SetLayoutPosition(tab, getTabDepth(tab));
 
             // IsPresent of TabItems is based on Y position.
```

bdach: Yeah alright, that doesn't look much better than the current version after all. I don't really like the complexity around this autosort logic but it seems about the best thing that can be done without a larger restructuring.

Going to approve but I'm reluctant to merge without a double check.

bdach (dismissed):

peppy (approved):

PR #4757, opened 2021-09-05
Fix user storage potentially losing true existing install directory

peppy:
It was pointed out recently by @p00ya (after encountering this precise scenario) that user data could be completely unlinked from an existing "install" of an osu!framework game via the following steps:

- Run via release and have data created (previously defaults on macOS to `~/osu`, which is super weird but ok)
- Decide to help develop the game, install `dotnet-sdk` which makes a `.local/share` folder
- Run via release again and have data create at `.local/share/osu`, losing existing install at less-preferred `~/osu`

This changes the resolution to be a two-part lookup, with the first half checking for existing data folders and the second deciding where to create a *new* data folder.

Still not sure about the macOS default paths, but that's for another PR. We probably want to use `ApplicationData` (which resolves to `.config`) if anything. Or figure out correct usage of `~/Library/Application Data`.

This change also (as of https://github.com/ppy/osu-framework/pull/4757/commits/96afb3c3e936f787f4bbb5df66044677f05901bc) also alters the order of precedence to be based on whether the full path can be *created*, rather than if it *exists*.

### DIFF
--- a/osu.Framework.Android/AndroidGameHost.cs
+++ b/osu.Framework.Android/AndroidGameHost.cs
@@ -59,7 +59,9 @@ namespace osu.Framework.Android
 
         public override Storage GetStorage(string path) => new AndroidStorage(path, this);
 
-        public override string UserStoragePath => Application.Context.GetExternalFilesDir(string.Empty).ToString();
+        public override IEnumerable<string> UserStoragePaths => new[] {
+            Application.Context.GetExternalFilesDir(string.Empty).ToString()
+        };
 
         public override void OpenFileExternally(string filename)
             => throw new NotImplementedException();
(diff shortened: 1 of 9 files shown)


osu.Framework.Android/AndroidGameHost.cs
```diff
@@ -59,7 +59,9 @@ protected override void SetupConfig(IDictionary<FrameworkSetting, object> defaul
 
         public override Storage GetStorage(string path) => new AndroidStorage(path, this);
 
-        public override string UserStoragePath => Application.Context.GetExternalFilesDir(string.Empty).ToString();
+        public override IEnumerable<string> UserStoragePaths => new[] {
+            Application.Context.GetExternalFilesDir(string.Empty).ToString()
```
smoogipoo, line 63:
```suggestion
        public override IEnumerable<string> UserStoragePaths => new[]
        {
            Application.Context.GetExternalFilesDir(string.Empty).ToString()
```

smoogipoo (changes requested):

smoogipoo (approved):

PR #4768, opened 2021-09-10
Implement new audio mixer concurrency limits

smoogipoo:
I debated for quite a long time whether to keep playback concurrency as part of `Sample` or not. Here's a few observations I made:
1. One problem is that `Sample`s are inherently distinct objects - they don't know about each other and especially don't know about each others' channels, so they would have to be linked with the `AudioMixer` in some way anyway because that's the only component which knows about everything.
2. When considering the case of global (shared between all `Sample`s with the same data backing) vs per-`Sample` concurrency limitations, I didn't feel like keeping playback concurrency as part of `Sample` to be the most extensible and intuitive way to go about this if it would ever be required.
3. After looking at how we're currently doing concurrency, I noticed that we have to explicitly set `SampleStore.PlaybackConcurrency` when creating a custom `SampleStore` (e.g. the `DrawableRuleset` `SampleStore`). This feels very flaky to me, easy for a consumer to miss, and quite hard to document.

With all of that in mind, I came to the conclusion that concurrency should be part of `AudioMixer`. It works on audio channels, which means that there's no distinction between a track or a sample - the user should be able to use this without caring about there being special cases.

By default, I've set unlimited concurrency to both the global track mixer and sample mixer.

---

## Playback concurrency

By default, all `AudioMixer`s play as many channels as given to them concurrently. For cases where this may be undesirable, such as when playing back samples on user interactions, the `AudioMixer.Concurrency` property can be used to limit the number of concurrent playback channels.
```csharp
[BackgroundDependencyLoader]
private void load(ISampleStore samples, AudioManager audioManager)
{
    // Limit all global samples to two concurrent playbacks.
    audioManager.SampleMixer.Concurrency = 2;

    Sample globalSample = samples.Get(...);

    // The sample attached to the global audio mixer can only be played back two times.
    globalSample.Play(); // 1
    globalSample.Play();
    globalSample.Play(); // Upon calling this, playback #1 above will be stopped.


    DrawableAudioMixer localAudioMixer;
    DrawableSample localSample;

    // A local audio mixer allowing six concurrent playbacks.
    Add(localAudioMixer = new DrawableAudioMixer
    {
        Concurrency = 6,
        Child = localSample = new DrawableSample(samples.Get(...)),
        Effects =
        {
            ...
        }
    });

    // A sample within the local mixer can only be played back six times.
    localSample.Play(); // 1
    localSample.Play();
    localSample.Play();
    localSample.Play();
    localSample.Play();
    localSample.Play();
    localSample.Play(); // Upon calling this, playback #1 above will be stopped.
}
```

### DIFF
--- a/osu.Framework/Audio/Sample/ISample.cs
+++ b/osu.Framework/Audio/Sample/ISample.cs
@@ -1,8 +1,6 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using osu.Framework.Bindables;
-
 namespace osu.Framework.Audio.Sample
 {
     /// <summary>
@@ -16,28 +14,14 @@ namespace osu.Framework.Audio.Sample
         double Length { get; }
 
         /// <summary>
-        /// The number of times this sample (as identified by name) can be played back concurrently.
-        /// </summary>
-        /// <remarks>
-        /// This affects all <see cref="ISample"/> instances identified by the same sample name.
-        /// </remarks>
-        Bindable<int> PlaybackConcurrency { get; }
-
-        /// <summary>
         /// Creates a new unique playback channel for this <see cref="ISample"/> and immediately plays it.
         /// </summary>
-        /// <remarks>
-        /// Multiple channels can be played simultaneously, but can only be heard up to <see cref="PlaybackConcurrency"/> times.
-        /// </remarks>
         /// <returns>The unique <see cref="SampleChannel"/> for the playback.</returns>
         SampleChannel Play();
 
         /// <summary>
         /// Retrieves a unique playback channel for this <see cref="ISample"/>.
         /// </summary>
-        /// <remarks>
-        /// Multiple channels can be retrieved and played simultaneously, but can only be heard up to <see cref="PlaybackConcurrency"/> times.
-        /// </remarks>
         /// <returns>The unique <see cref="SampleChannel"/> for the playback.</returns>
         SampleChannel GetChannel();
     }
--- a/osu.Framework/Audio/Sample/SampleBassFactory.cs
+++ b/osu.Framework/Audio/Sample/SampleBassFactory.cs
@@ -8,7 +8,6 @@ using System.Runtime.InteropServices;
 using ManagedBass;
 using osu.Framework.Allocation;
 using osu.Framework.Audio.Mixing.Bass;
-using osu.Framework.Bindables;
 using osu.Framework.Platform;
 
 namespace osu.Framework.Audio.Sample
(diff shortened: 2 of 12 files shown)


smoogipoo:
Note that with these changes, concurrency is shared between all samples in the mixer. To achieve the effect of the previous "global" concurrency (via the same data backing) you'd have to add samples to a separate audio mixer.

So currently the osu! limit of 6 may be too low but in my testing (https://osu.ppy.sh/beatmapsets/39804#osu/129891 + https://osu.ppy.sh/beatmapsets/177187#osu/426550) I didn't notice any abrupt cutting out.

peppy:
So just to confirm, there's no way to limit concurrency per sample now? Limiting at a mixer may be something we want for general safety purposes, but the goal here was to allow us to use the concurrency specification in place of manual debounce logic, no? Does that mean that you envisage a mixer being created for hover sounds, for instance, if we want to limit concurrency there?

Also with a global limit of 6, wouldn't a long slider sound cut out if only 5 ticks played since it started? Or am I missing something?

osu.Framework/Audio/Sample/SampleBassFactory.cs
```diff
@@ -76,8 +54,12 @@ private void loadSample()
             int dataLength = data.Length;
 
             const BassFlags flags = BassFlags.Default | BassFlags.SampleOverrideLongestPlaying;
+
             using (var handle = new ObjectHandle<byte[]>(data, GCHandleType.Pinned))
-                SampleId = Bass.SampleLoad(handle.Address, 0, dataLength, PlaybackConcurrency.Value, flags);
+            {
+                // Concurrency is handled by the audio mixer, so the number of playbacks is irrelevant.
```
bdach, line 60: As a side note, this is sorta correct but not 100%. I'm pretty sure that the reason why the concurrency setting on the sample does nothing is that `SampleGetChannel` is called with `BassFlags.SampleChannelStream`. And that ignores concurrency, because [you cannot get a `BASS_ERROR_NOCHAN` error from bass if `BASS_SAMCHAN_NEW` isn't specified](https://www.un4seen.com/doc/#bass/BASS_SampleGetChannel.html).

bdach:
On reading, the source looked sane but I'm partially in agreement with the preceding comment that I'm not sure that this solves the problems that the game has. I guess you could split a mixer for UI sounds, or other sounds, but that raises some questions for me - how many mixers will we end up having? how well will they scale? etc. etc.

One thing is for sure, the assumption of limiting at mixer level simplifies implementation by quite a lot. Doing per-sample tracking would add complexity to that (probably would have to keep a dictionary of concurrency settings or something).

smoogipoo:
Yeah I agree, kinda forgot about that aspect of it... Will close this PR for the time being then.

osu.Framework/Audio/Sample/ISample.cs
```diff
@@ -15,29 +13,15 @@ public interface ISample : IAdjustableAudioComponent
         /// </summary>
         double Length { get; }
 
-        /// <summary>
-        /// The number of times this sample (as identified by name) can be played back concurrently.
-        /// </summary>
-        /// <remarks>
-        /// This affects all <see cref="ISample"/> instances identified by the same sample name.
-        /// </remarks>
-        Bindable<int> PlaybackConcurrency { get; }
```
peppy, line 24: Not too sure about just closing this PR, I can see there potentially being a use case for limiting at a mixer level. But I dunno if that's something we'd want for osu!.

But just to make sure we're on the same page, the xmldoc here is currently not working as documented right? The "identified by name" part specifically.

smoogipoo: By currently, do you mean in this PR or in master? In master it's not working at all. In this PR it's shared across _all_ samples, not just identified by the same sample name.

peppy: Yep, meant master.

PR #4771, opened 2021-09-13
Add flow for testing a nested game instance inside a `TestScene`

peppy:
Turns out that the way we had this setup previously made it very dangerous if a nested game instance was to call `Game.Exit()` (exiting full host execution and potentially skipping test cases with a "Success" status.

# Breaking changes

## `Game` instances can no longer be `Add`ed to a test scene directly

To test a `Game` instance, use the new `AddGame` method. Following is an example of how to update an existing usage.

Before:

```csharp
AddStep("Add game", () => 
{
    var game = new TestGame();
    game.SetHost(host);
    Add(game)
});
```

After:

```csharp
// SetHost and Add are automatically called internally.
AddStep("Add game", () => AddGame(game = new TestGame()));
```

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -541,7 +541,7 @@ namespace osu.Framework.Platform
         /// <summary>
         /// Schedules the game to exit in the next frame.
         /// </summary>
-        public virtual void Exit() => PerformExit(false);
+        public void Exit() => PerformExit(false);
 
         /// <summary>
         /// Schedules the game to exit in the next frame (or immediately if <paramref name="immediately"/> is true).
--- a/osu.Framework/Testing/TestScene.cs
+++ b/osu.Framework/Testing/TestScene.cs
@@ -52,12 +52,12 @@ namespace osu.Framework.Testing
             host = new TestSceneHost($"{GetType().Name}-{Guid.NewGuid()}", exitNestedGame);
             runner = CreateRunner();
 
-            if (!(runner is Game runnerGame))
+            if (!(runner is Game game))
                 throw new InvalidCastException($"The test runner must be a {nameof(Game)}.");
 
-            runTask = Task.Factory.StartNew(() => host.Run(runnerGame), TaskCreationOptions.LongRunning);
+            runTask = Task.Factory.StartNew(() => host.Run(game), TaskCreationOptions.LongRunning);
 
-            while (!runnerGame.IsLoaded)
+            while (!game.IsLoaded)
             {
                 checkForErrors();
                 Thread.Sleep(10);
--- a/osu.Framework/Testing/TestScene.cs
+++ b/osu.Framework/Testing/TestScene.cs
@@ -46,24 +46,6 @@ namespace osu.Framework.Testing
 
         public object DynamicCompilationOriginal { get; internal set; }
 
-        [OneTimeSetUp]
-        public void SetupGameHost()
-        {
-            host = new TestSceneHost($"{GetType().Name}-{Guid.NewGuid()}", exitNestedGame);
-            runner = CreateRunner();
-
-            if (!(runner is Game game))
-                throw new InvalidCastException($"The test runner must be a {nameof(Game)}.");
-
-            runTask = Task.Factory.StartNew(() => host.Run(game), TaskCreationOptions.LongRunning);
-
-            while (!game.IsLoaded)
-            {
-                checkForErrors();
-                Thread.Sleep(10);
-            }
-        }
-
         [BackgroundDependencyLoader]
         private void load(GameHost host)
         {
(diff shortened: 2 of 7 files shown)


osu.Framework/Testing/TestScene.cs
```diff
@@ -42,23 +42,46 @@ public abstract class TestScene : Container, IDynamicallyCompile
...
                 Thread.Sleep(10);
             }
         }
 
-        protected internal override void AddInternal(Drawable drawable) =>
+        /// <summary>
+        /// Add a full game instance in a nested state for visual testing.
+        /// </summary>
+        /// <remarks>
+        /// Any previous game added via this method will be disposed if called multiple times.
```
peppy, line 64: Not 100% on this behaviour. Can be changed to throw or otherwise based on what people prefer.

peppy:
Needs a fix for visual executions, looking into it now.

osu.Framework/Platform/GameHost.cs
```diff
@@ -541,7 +541,7 @@ private set
         /// <summary>
         /// Schedules the game to exit in the next frame.
         /// </summary>
-        public void Exit() => PerformExit(false);
+        public virtual void Exit() => PerformExit(false);
```
bdach, line 544: This one is maybe a bit strange. `PerformExit()` is already virtual (and apparently has no overrides?) - so not sure whether this also needs to be virtual?

bdach:
In general this looks good to me, but I'm not super certain as I haven't seen this code too much. It is a bit of a bummer that these changes don't look to have any effect on the test browser - any `Exit()` from a child game will still exit the entire browser. But that's probably out of scope here..

peppy:
Fixing the test browser is definitely something on my radar, but as you can probably guess it's a bit more involved (may require a different fix altogether) as we can't change the game host class.

osu.Framework/Testing/TestScene.cs
```diff
@@ -37,28 +39,69 @@ public abstract class TestScene : Container, IDynamicallyCompile
...
 
+        /// <summary>
+        /// A nested game instance, if added via <see cref="AddGame"/>.
+        /// </summary>
+        private Game nestedGame;
+
         public object DynamicCompilationOriginal { get; internal set; }
 
         [OneTimeSetUp]
         public void SetupGameHost()
```
frenzibyte, line 50: Although slightly out of scope from this PR, I think renaming this method to something like `SetupHeadless` would be better as to not confuse it of being performed in all scenarios (which now, with `TestSceneHost` instead of `TestRunHeadlessGameHost`, makes it think more like so)

(also same would apply to `DestroyGameHost`)
```suggestion
        public void SetupHeadless()
```

osu.Framework/Testing/TestScene.cs
frenzibyte: The name `AddGame` kinda feels like it's supposed to allow more than one nested game, maybe going with `RunGame` or something so that it would make sense running another game exits the current one?

osu.Framework/Testing/TestScene.cs
```diff
@@ -69,7 +112,7 @@ public void SetupGameHost()
         [OneTimeTearDown]
         public void DestroyGameHost()
         {
-            host.Exit();
+            (host as TestSceneHost)?.ExitFromRunner();
```
frenzibyte, line 115: nit: would make sense to direct cast since `DestroyGameHost` should only be invoked in headless context as far as I'm aware.
```suggestion
            var headlessHost = (TestSceneHost)host;
            headlessHost.ExitFromRunner();
```

frenzibyte (dismissed):
Sounds good otherwise.

osu.Framework/Testing/TestScene.cs
peppy: I was doing it for the null check, which I think we want.

frenzibyte: `host` was previously accessed without a null conditional though?

peppy: I guess you could say that, yeah.

osu.Framework/Testing/TestScene.cs
peppy: I've renamed in line with the other nunit specific methods.

osu.Framework/Platform/GameHost.cs
peppy: Have changed to use `PerformExit` instead, looks to work just as well.

osu.Framework/Testing/TestScene.cs
peppy: If the current behaviour is decided as the correct direction to go, then renaming sounds like a good improvement.

peppy:
Looks like something regressed in https://github.com/ppy/osu-framework/pull/4771/commits/2ff3e3fce0e72188f3dfb28d7c932afb442bbdf2.

osu.Framework/Testing/TestScene.cs
bdach: I think being able to swap game instances is okay but I'm guessing my opinion might not necessarily be the one you're looking for here.

bdach (approved):
i'll just leave this approve here, the remaining comment (renaming `AddGame()`) is not a huge deal for me

smoogipoo (approved):
lgtm

PR #4772, opened 2021-09-13
Fix double disposal of `ThreadedTaskScheduler` throwing an exception

peppy:


osu.Framework.Tests/Threading/ThreadedTaskSchedulerTest.cs
```diff
@@ -22,18 +22,22 @@ public void EnsureThreadedTaskSchedulerProcessesBeforeDispose()
 
             const int target_count = 128;
 
-            using (var taskScheduler = new ThreadedTaskScheduler(4, "test"))
+            using var taskScheduler = new ThreadedTaskScheduler(4, "test");
```
smoogipoo, line 25: We've blocked merging code using in-line using like this before. You could just dispose after the using block.

smoogipoo (changes requested):

bdach:
The diff looks good, but I'm pretty sure the merge revert at the end of history here (a8ed603) is going to cause problems. Because it builds on the dependency commits and shares hashes with them, merging `master` back here will cause the dependency to be reverted.

The merge revert will either have to be reverted a second time, or force-pushed out entirely.

peppy:
Yeah, I'll have to deal with that at that point. Would usually force push for such a case, but alas. Will deal with after one of the two is merged.

smoogipoo (approved):

Issue #3271, opened 2020-02-16
CSS-compatible font sizing


bdach:
Recently osu!-side there have been multiple contributions that aimed to bring visuals as close to web as possible. An issue raised during that period was that the numerically-identical font size looked smaller in the game than in web.

Right now, `SpriteText` & co. work in a way such that a height of 16 will yield a text sprite of the size of precisely 16. While that's what seems reasonable, it's actually not what CSS does:

![image](https://user-images.githubusercontent.com/20418176/74613905-4873d280-5113-11ea-8c7d-d78385d56d3b.png)

At this point I'll quickly [defer to this article](https://iamvdo.me/en/blog/css-font-metrics-line-height-and-vertical-align#lets-talk-about-font-size-first) explaining how CSS `font-size` works. Especially relevant is this picture:

![](https://iamvdo.me/content/01-blog/30-css-avance-metriques-des-fontes-line-height-et-vertical-align/upm-px-equivalent.png)

which shows that for a font:

* with a 1000-unit em-square,
* 680-unit capital height,
* a 1100-unit ascender,
* and a 540-unit descender,

the effective text height in pixels of a 100px line is actually 164px.

To see how this relates to how framework handles text, I used the existing `SpriteText` test scene, Open Sans and FontForge. The metrics for Open Sans are:

* em-square height: 2048 units,
* ascender height: 2189 units,
* descender height: 600 units,
* capital height: 1462 units.

That means that:

* the em-square's height is 2048 / (2189 + 600) = ~73.4% effective height of the line,
* and a capital letter's height is 1462 / (2189 + 600) = ~52.4% effective height of the line.

The measurements in-game seem to indicate that currently the framework is sizing the entire line (ascender to descender) when size is supplied, instead of sizing the em-square:

![font-sizing](https://user-images.githubusercontent.com/20418176/74613914-5de8fc80-5113-11ea-8317-1e4b56701661.png)

126 * 52.4% comes down to just over 66 pixels. Measurement is imprecise due to antialiasing. I've only checked this one font, but seems too much of a coincidence if that turned out to be a fluke.

Now, the bad news is, as far as I can see font spacing info comes opaquely through bmfont, and all the dimensions above are font-specific, therefore we'd most likely have to modify bmfont to supply us this info, as it doesn't seem to be there. [There is a `base` attribute in the `common` tag](https://www.angelcode.com/products/bmfont/doc/file_format.html), but seems to be the ascender-to-baseline measurement, so not what we want here. 

[...]

PR #4774, opened 2021-09-15
Align `TextBuilder` characters by their baselines during addition/removal

frenzibyte:
Previously characters were *implicitly aligned* using their `YOffset`s, and that works when only one font is used, but not when more than one font gets used inside one `SpriteText`/`TextBuilder`.

This PR aims to resolve the "multiple fonts" case by offsetting their `DrawRectangle`s based on the difference between the `Baseline` of the new character and the previous characters.

I've worked on this initially so that font-mixed sprite texts with #3271 don't look badly misaligned (as I noticed after implementing #3271), but apparently, there are already alignment improvements osu!-side with this PR alone (although slightly unnoticable):

- Before:
image

- After:
image

(notice `(TV Size)` aligned correctly with the other glyphs)

I have added in-depth test coverage in `TextBuilderTest` which had the perfect setup for having so, with all possible scenarios I can think of.

Along with test coverage, I've also added a benchmark including the regular "no baseline adjustment" cases, and "one baseline adjustment every 10 characters" case, and the results seem reasonable.

|                                    Method |     Mean |    Error |   StdDev |
|------------------------------------------ |---------:|---------:|---------:|
|                             AddCharacters | 19.99 us | 0.280 us | 0.323 us |
|       AddCharactersWithDifferentBaselines | 31.02 us | 0.219 us | 0.183 us |
|                       RemoveLastCharacter | 24.90 us | 0.493 us | 0.506 us |
| RemoveLastCharacterWithDifferentBaselines | 38.14 us | 0.570 us | 0.505 us |

I can see this being further optimised by offseting new characters to negative `X/Y`s rather than offsetting all the other previous characters in the line, but that might be a bit complicated to deal with in the `SpriteText_DrawNode` (especially given that `TextBuilder` can have multiple lines), therefore I've went with the simplest way of for looping all characters in the line for now.

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkTextBuilder.cs
+++ b/osu.Framework.Benchmarks/BenchmarkTextBuilder.cs
@@ -1,0 +1,45 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Threading.Tasks;
+using BenchmarkDotNet.Attributes;
+using osu.Framework.Graphics.Sprites;
+using osu.Framework.Graphics.Textures;
+using osu.Framework.Text;
+
+namespace osu.Framework.Benchmarks
+{
+    public class BenchmarkTextBuilder
+    {
+        private readonly ITexturedGlyphLookupStore store = new TestStore();
+
+        private TextBuilder textBuilder;
+
+        [Benchmark]
+        public void AddCharacters() => initialiseBuilder(false);
+
+        [Benchmark]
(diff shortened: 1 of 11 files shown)


smoogipoo (approved):
Looks solid. As I mentioned in discord, we probably want to rework `TextBuilder` altogether to improve performance - having everything done in-line in `AddCharacter`/`RemoveCharacter` is not good, but that goes for everything else as well.

peppy:
this is missing a breaking change clause for the updated method signature, will add.

peppy:
Actually there's quite a few changes, just going to leave it. All mostly internal stuff so I don't think too many people will be hit by it.

PR #4775, opened 2021-09-15
Avoid GL crash from vertex buffer index upload overflow

peppy:
Now throws instead. This should be a very edge case that can only be hit with custom usage of `QuadVertexBuffer`, and is intended to avoid a runaway scenario where the game becomes unresponsive.

sample exception:

```csharp
[runtime] 2021-09-15 09:30:10 [error]: An unhandled error has occurred.
[runtime] 2021-09-15 09:30:10 [error]: System.OverflowException: Attempted to initialise a QuadVertexBuffer with more than QuadVertexBuffer.MAX_QUADS quads (10922).
[runtime] 2021-09-15 09:30:10 [error]: at osu.Framework.Graphics.Batches.QuadBatch`1..ctor(Int32 size, Int32 maxBuffers) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Batches/QuadBatch.cs:line 19

```

### DIFF
--- a/osu.Framework/Graphics/OpenGL/Buffers/QuadVertexBuffer.cs
+++ b/osu.Framework/Graphics/OpenGL/Buffers/QuadVertexBuffer.cs
@@ -22,25 +22,28 @@ namespace osu.Framework.Graphics.OpenGL.Buffers
     public class QuadVertexBuffer<T> : VertexBuffer<T>
         where T : struct, IEquatable<T>, IVertex
     {
-        private readonly int amountQuads;
+        private readonly int amountIndices;
+
+        private const int indices_per_quad = TextureGLSingle.VERTICES_PER_QUAD + 2;
 
         internal QuadVertexBuffer(int amountQuads, BufferUsageHint usage)
             : base(amountQuads * TextureGLSingle.VERTICES_PER_QUAD, usage)
         {
-            this.amountQuads = amountQuads;
+            amountIndices = amountQuads * indices_per_quad;
+
+            if (amountIndices > ushort.MaxValue)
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/OpenGL/Buffers/QuadVertexBuffer.cs
```diff
@@ -22,25 +23,31 @@ static QuadIndexData()
...
+        /// The maximum number of quads supported by this buffer.
+        /// </summary>
+        public const int MAX_QUADS = ushort.MaxValue / indices_per_quad;
 
         internal QuadVertexBuffer(int amountQuads, BufferUsageHint usage)
             : base(amountQuads * TextureGLSingle.VERTICES_PER_QUAD, usage)
         {
-            this.amountQuads = amountQuads;
+            amountIndices = amountQuads * indices_per_quad;
+            Debug.Assert(amountIndices <= ushort.MaxValue);
```
bdach, line 39: I find it a little weird how this is a local assert (and a debug-only one at that), and the actual fatal exception is thrown outwardly from `QuadBatch`. Was there a reason for not throwing locally here?

peppy: historically we've used `Debug.Assert` for "known internal truths" and throwing for publicly exposed code. Not 100% sure on this concept, but that's the reasoning for doing it this way. `QuadVertexBuffer` is only exposed publicly via the call with the exception, and this class is `internal`, so it's just a last ditch effort to save things if a framework change was made that bypassed the former.

smoogipoo (approved):
Sure. I'm not 100% on this, but it's an interim change until VBOs are reworked.

PR #4778, opened 2021-09-17
Add ability for `SearchContainer` to match even when letters are missing

peppy:
Allows matching `MultiplayerLoungeSubScreen` using `MLSScre` and the likes. Seems like behaviour we'd probably want in all search containers, but if that's a point of conflict it can be done via a property or derived class.

### DIFF
--- a/osu.Framework/Graphics/Containers/SearchContainer.cs
+++ b/osu.Framework/Graphics/Containers/SearchContainer.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using osu.Framework.Caching;
 using osu.Framework.Extensions.IEnumerableExtensions;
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Containers/SearchContainer.cs
```diff
@@ -83,5 +84,25 @@ private static bool match(IFilterable filterable, IEnumerable<string> terms, boo
...
+            int index = 0;
+
+            for (int i = 0; i < needle.Length; i++)
+            {
+                // string.IndexOf doesn't have an overload which takes both a `startIndex` and `StringComparison` mode.
+                int found = CultureInfo.InvariantCulture.CompareInfo.IndexOf(haystack, needle[i], index, CompareOptions.OrdinalIgnoreCase);
+                if (found < 0)
+                    return false;
+
+                index = found;
```
ekrctb, line 102: Isn't this off-by-one? `xx` will match `x` I think.

peppy: yep, you're right. i had it correct then broke it after switching to `CompareInfo`

bdach (approved):

PR #4780, opened 2021-09-19
Publicly expose `SlimReadOnlyListWrapper.Count`

peppy:
Seems like an oversight as far as I can tell.


bdach (approved):
don't really see a reason not to

smoogipoo (approved):

Issue #4779, opened 2021-09-18
`FlushingStream` disposal can fail


bdach:
See [following CI run](https://github.com/ppy/osu-framework/runs/3631010396?check_suite_focus=true):

```
 The active test run was aborted. Reason: Test host process crashed : Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at System.IO.FileStream.get_IsClosed()
   at System.IO.FileStream.Flush(Boolean flushToDisk)
   at osu.Framework.Platform.FlushingStream.Dispose(Boolean disposing) in D:\a\osu-framework\osu-framework\osu.Framework\Platform\FlushingStream.cs:line 24
   at System.IO.FileStream.Finalize()
```

The nullref is clearly the internal `_fileHandle`. Apparently this is [only possible if the `FileStream` ctor throws](https://github.com/dotnet/runtime/blob/cf258a14b70ad9069470a108f13765e0e5988f51/src/libraries/System.Private.CoreLib/src/System/IO/FileStream.cs#L21), if that inline comment is to be trusted.

There are also [other](https://github.com/ppy/osu-framework/pull/4778/checks?check_run_id=3640074332) [CI failures](https://github.com/ppy/osu-framework/runs/3640074332?check_suite_focus=true) related to files that _may_ potentially be caused by `FlushingStream` but I have no concrete evidence that they are.

Only ever seen windows failures so far, so marking as windows-specific for the time being.

peppy:
Sounds like we may just want to try-catch this? Not sure if there's a better alternative. We've had to do similar for windows in other locations.

bdach:
Yeah, there is no actual other way to handle that, the file handles are not exposed.

The other errors about files being held open by some unnamed process are weird though. Would need to look at that more but am not on windows right now.

PR #4781, opened 2021-09-20
Avoid throwing errors on failure to flush stream

peppy:
Also saw this occur on macOS in the form of:

```csharp

System.ObjectDisposedException : Cannot access a closed file.
   at System.IO.FileStream.Flush(Boolean flushToDisk)
   at osu.Framework.Platform.FlushingStream.Dispose(Boolean disposing)
   at System.IO.Stream.Close()
   at osu.Game.Tournament.Models.StableInfo.SaveChanges() in
/Users/dean/Projects/osu/osu.Game.Tournament/Models/StableInfo.cs:line
58
```

See https://github.com/ppy/osu/blob/879b1ab046eefd501e82a8d4d4ce4baa124872ac/osu.Game.Tournament/Models/StableInfo.cs#L49, which uses a `StreamWriter` without specifying `leaveOpen=true`. This may be a point of concern/investigation though (ie. does this leave a case where the resulting file could be NUL byte filled?). I seem to recall `StreamWriter` performing its own flush operations, but possible not with the `true` parameter we added.

Either way, for the other case reported in the issue this try-catch is likely still what we want.

Closes https://github.com/ppy/osu-framework/issues/4779.

### DIFF
--- a/osu.Framework/Configuration/IniConfigManager.cs
+++ b/osu.Framework/Configuration/IniConfigManager.cs
@@ -84,10 +84,10 @@ namespace osu.Framework.Configuration
             try
             {
                 using (var stream = storage.GetStream(Filename, FileAccess.Write, FileMode.Create))
-                using (var sw = new StreamWriter(stream, leaveOpen: true))
+                using (var w = new StreamWriter(stream))
                 {
                     foreach (var p in ConfigStore)
-                        sw.WriteLine(@"{0} = {1}", p.Key, p.Value.ToString().AsNonNull().Replace("\n", "").Replace("\r", ""));
+                        w.WriteLine(@"{0} = {1}", p.Key, p.Value.ToString().AsNonNull().Replace("\n", "").Replace("\r", ""));
                 }
             }
             catch
--- a/osu.Framework/Configuration/IniConfigManager.cs
+++ b/osu.Framework/Configuration/IniConfigManager.cs
@@ -84,10 +84,10 @@ namespace osu.Framework.Configuration
             try
             {
                 using (var stream = storage.GetStream(Filename, FileAccess.Write, FileMode.Create))
-                using (var w = new StreamWriter(stream))
+                using (var sw = new StreamWriter(stream, leaveOpen: true))
                 {
                     foreach (var p in ConfigStore)
-                        w.WriteLine(@"{0} = {1}", p.Key, p.Value.ToString().AsNonNull().Replace("\n", "").Replace("\r", ""));
+                        sw.WriteLine(@"{0} = {1}", p.Key, p.Value.ToString().AsNonNull().Replace("\n", "").Replace("\r", ""));
                 }
             }
             catch
--- a/osu.Framework/Configuration/InputConfigManager.cs
+++ b/osu.Framework/Configuration/InputConfigManager.cs
@@ -36,7 +36,7 @@ namespace osu.Framework.Configuration
             try
             {
                 using (var stream = storage.GetStream(FILENAME, FileAccess.Write, FileMode.Create))
-                using (var sw = new StreamWriter(stream, leaveOpen: true))
+                using (var sw = new StreamWriter(stream))
                 {
                     sw.Write(JsonConvert.SerializeObject(this));
                     return true;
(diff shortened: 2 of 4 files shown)


peppy:
Turns out rider lied to me about the method signature in the last commit (only available in `net50`). In addition, using this overload is a point of pain as the signature available in netstandard doesn't have sane defaults exposed (see https://github.com/microsoft/referencesource/blob/master/mscorlib/system/io/streamwriter.cs).

Maybe we should be internalising this in `Storage` and exposing a `GetStreamWriter` to save local implementation in every usage?

bdach:
Disposal of a `StreamWriter` does seem to call `Flush()` ([netstandard](https://github.com/microsoft/referencesource/blob/5697c29004a34d80acdaf5742d7e699022c64ecd/mscorlib/system/io/streamwriter.cs#L236-L258), [netcore](https://github.com/dotnet/runtime/blob/v5.0.0/src/libraries/System.Private.CoreLib/src/System/IO/StreamWriter.cs#L178-L197)), *but* it calls it without the `true` param, so it looks like it won't flush the OS buffers.

I could be okay with adding a `GetStreamWriter()` or similar. The worst part about it would be figuring how to make sure consumers don't use `StreamWriter`s raw and therefore risk not having them flush.

peppy:
if our goal is to fix OS bugs for all consumers (not sure that's a correct standpoint) we can either mention this in `GetStream` or remove the ability to get streams for write without using the new method.

But I don't think `StreamWriter` is going to be the only case here.

smoogipoo:
Regarding the signature, I think it's accepted to either inline the defaults or to use `#if NETSTANDARD`. We do this in other areas already.

peppy:
Also the default aren't really well exposed in the previous versions (had to look then up using reference source).

smoogipoo:
I'm actually a little bit confused by this PR in general. `leaveOpen = false` by default, so `StreamWriter.Dispose()` calls `Stream.Close()` which calls `Dispose(true)`, which will reach this code.

An issue with `StreamWriter` would only occur if `leaveOpen = true`, as you've changed usages to do in this PR, which indeed calls `Stream.Flush(false)`. This change looks backwards from what we're trying to achieve here?

peppy:
Without this change exceptions are thrown due to the stream being closed at the point of reaching our `Flush(true)`, all while not performing the `Flush(true)` (so if we are to catch the exception it may still result in corruption).

peppy:
Hmm, I see what you're saying. The `StreamWriter` should still be calling the flush call once during disposal?

So we really just want an `isDipsoed` flag?

smoogipoo:
Just so we're on the same page, these are the two scenarios I'm thinking of (perhaps I'm missing something?):
```csharp
using (var stream = ...)
{
    using (var sw = new StreamWriter(stream, leaveOpen: true)
    {
    ...
    } // stream.Flush(false)
} // stream.Dispose() -> custom.Dispose() -> Flush(true)

using (var stream = ...)
{
    using (var sw = new StreamWriter(stream, leaveOpen: false) // Default param
    {
    ...
    } // stream.Close() -> stream.Dispose() -> custom.Dispose() -> Flush(true)
} // stream.Dispose()
```

> So we really just want an isDipsoed flag?

It seems like that should work.

peppy:
> Just so we're on the same page, these are the two scenarios I'm thinking of (perhaps I'm missing something?):
> 
> ```cs
> using (var stream = ...)
> {
>     using (var sw = new StreamWriter(stream, leaveOpen: true)
>     {
>     ...
>     } // stream.Flush(false)
> } // stream.Dispose() -> custom.Dispose() -> Flush(true)
> 
> using (var stream = ...)
> {
>     using (var sw = new StreamWriter(stream, leaveOpen: false) // Default param
>     {
>     ...
>     } // stream.Close() -> stream.Dispose() -> custom.Dispose() -> Flush(true)
> } // stream.Dispose()
> ```
> 
>  > So we really just want an isDipsoed flag?
> 
> It seems like that should work.

Makes sense, yeah.

I've changed to use a flag instead. Will re-test against macOS and report back.

peppy:
Works fine.

smoogipoo (approved):

PR #4783, opened 2021-09-21
Update `IsPrimaryInstance` on first access rather than after `GameHost.Run`

peppy:
This is mainly to fix the only framework-side test scene relying on two `GameHost`s running at the same time.

### DIFF
--- a/osu.Framework.Tests/Platform/HeadlessGameHostTest.cs
+++ b/osu.Framework.Tests/Platform/HeadlessGameHostTest.cs
@@ -36,7 +36,7 @@ namespace osu.Framework.Tests.Platform
         public void TestIpc()
         {
             using (var server = new BackgroundGameHeadlessGameHost(@"server", true))
-            using (var client = new BackgroundGameHeadlessGameHost(@"client", true))
+            using (var client = new HeadlessGameHost(@"client", true))
             {
                 Assert.IsTrue(server.IsPrimaryInstance, @"Server wasn't able to bind");
                 Assert.IsFalse(client.IsPrimaryInstance, @"Client was able to bind when it shouldn't have been able to");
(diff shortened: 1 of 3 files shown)


osu.Framework/Platform/DesktopGameHost.cs
```diff
@@ -39,17 +39,31 @@ protected sealed override Storage GetDefaultGameStorage()
 
         public sealed override Storage GetStorage(string path) => new DesktopStorage(path, this);
 
+        public override bool IsPrimaryInstance
+        {
+            get
+            {
+                // make sure we have actually attempted to bind IPC as this call may occur before the host is run.
+                ensureIPCReady();
```
bdach, line 47: Not super sure about a getter such as this having I/O side effects, it looks pretty gnarly. I'm not too familiar with this IPC stuff or what it's intended for, but if it's supposed to potentially be available before `GameHost.Run()` is called, is initialising in ctor an option?

Or, conversely, maybe it *should* be required for game hosts to be ran first for IPC to be available, and the *test* should be starting them?

osu.Framework.Tests/Platform/HeadlessGameHostTest.cs
```diff
@@ -36,7 +36,7 @@ public void TestGameHostDisposalWhenNeverRun()
         public void TestIpc()
         {
             using (var server = new BackgroundGameHeadlessGameHost(@"server", true))
-            using (var client = new BackgroundGameHeadlessGameHost(@"client", true))
+            using (var client = new HeadlessGameHost(@"client", true))
```
bdach, line 39: Do you mind spelling out what this change is? I don't think I get it. Is this ensuring that two hosts are spawned but only one game runs (because the inner host never has `Run(Game)` called on it)?

peppy: Yep. The inner one is used to emulate a second instance running, checking for an already running (and bound) IPC instance to forward a file association import message to.

See `osu.Desktop/Program.cs` for a real world example.

smoogipoo (approved):

PR #4784, opened 2021-09-21
Add global mutex on `GameHost` running state

peppy:
This enforces the expectation that only one `GameHost` should be running at a time. I'm not yet sure if we should make this change as it means that parallel test runs are no longer possible (unless run under a different .NET host). Maybe okay, but also maybe not as it breaks default settings on some IDEs. And also makes running tests a lot slower.

PRing for discussion and to see if we trigger any failures on CI as a result (to help track the cause behind https://github.com/ppy/osu-framework/runs/3659452824?check_suite_focus=true).

- [x] Depends on https://github.com/ppy/osu-framework/pull/4783 (to fix one case of concurrent host runs)

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -574,6 +574,8 @@ namespace osu.Framework.Platform
             stoppedEvent.Set();
         }
 
+        private static readonly SemaphoreSlim host_running_mutex = new SemaphoreSlim(1);
+
         public void Run(Game game)
         {
             GCSettings.LatencyMode = GCLatencyMode.SustainedLowLatency;
@@ -589,6 +591,9 @@ namespace osu.Framework.Platform
 
             try
             {
+                if (!host_running_mutex.Wait(10000))
+                    throw new TimeoutException($"This {nameof(GameHost)} could not start {game} because another {nameof(GameHost)} was already running.");
+
                 threadRunner = CreateThreadRunner(InputThread = new InputThread());
 
                 AppDomain.CurrentDomain.UnhandledException += unhandledExceptionHandler;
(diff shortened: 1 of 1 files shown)


peppy:
Realised that I didn't fix the one case where this was actually an issue (https://github.com/ppy/osu-framework/pull/4783) so I've added that as a dependency. Still worrying that the tests are going to hit the 1 hour timeout, and should be investigated separately.

bdach:
I was really under the impression that this was already the case (as in, several concurrent hosts under single .NET host were not allowed) and that tests weren't running with any parallelism. I'm relatively certain that every time I ran whole-solution tests locally game-side they executed serially regardless of assembly of origin.

I don't remember changing anything in rider settings to make that happen either... So I'm not really seeing where the slowdown in test runtime would come from right now.

bdach:
On double-checking, rider's internal test runner seems to disregard the nunit default of assembly parallelisation, it only ever shows when using `dotnet test` from CLI, which means that slowdowns will primarily impact CI from my view (I never run tests from CLI myself).

If that helps us get more consistent test execution at the expense of going back to appveyor-era CI times than I think it's fine to proceed with this change anyway. I'd rather have consistency over speed.

peppy:
In rider for me, running all osu! tests:

16 threads: 11 minutes 59 seconds
1 thread: 16 minutes 50 seconds

Not sure how I feel.

bdach:
I think that's not horrible. I'm sort of used to leaving tests to run and going to do $something_else when I want to re-run the full suite locally anyway.

Probably salient to get a third opinion though.

peppy:
Yeah, I'm not going to complain. Although this PR is blocked by the test failure related to log writes. I'll take a look into that one.

peppy:
The CI issue is definitely not related to this change, so shouldn't be considered a blocker. I'm looking at a fix but it'll be a separate PR.

smoogipoo (approved):

PR #4787, opened 2021-09-22
Fix `windowPositionX/Y` being reset to `0.5` on startup

### LINKED ISSUE #4608: Opening game sets `WindowedPositionX/Y` to (0.5, 0.5)
I've been testing this on Windows only.

Every time osu!/visual tests are opened, `WindowedPositionX` and `WindowedPositionY` get set to `0.5`. This happens with all `WindowMode`s: windowed, borderless and fullscreen.

To repro:
1. Open game
2. Set to windowed
3. Move window to a position that is not the center of the screen
4. Close game
5. Open game
6. (Bonus) Check `framework.ini` (values will be 0.5)
7. Cycle `WindowMode` to windowed (3x<kbd>F11</kbd>)
8. The window will be in the center of the screen, and not in the position it opened in.

Susko3:
Closes #4608

This code seems to be doing nothing productive, the only thing it can do is reset the `windowPosition` in the config to unwanted values. Resulting in the windowed window appearing in the center if the screen when changing modes or opening the game.

Tested only on Windows.

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -1128,8 +1128,6 @@ namespace osu.Framework.Platform
             CurrentDisplayBindable.ValueChanged += evt =>
             {
                 windowDisplayIndexBindable.Value = (DisplayIndex)evt.NewValue.Index;
-                windowPositionX.Value = 0.5;
-                windowPositionY.Value = 0.5;
             };
 
             config.BindWith(FrameworkSetting.LastDisplayDevice, windowDisplayIndexBindable);


bdach:
As I was expecting a fix for this to be somehow platform-specific and the OP doesn't offer any explanations for that, I looked into this independently and it does seem that these particular sets are at fault. Because the initial display index is -1, loading the config value for `FrameworkSetting.LastDisplayDevice` causes this recenter logic to always run on startup, after the config values for `WindowedPosition{X,Y}` are read.

The platform specific part of this seems to be that on linux a `SDL_WINDOWEVENT_MOVED` is fired *after* that set, and it has the original values read from config in the event args. So that event firing sort of "self-corrects" the recenter removed in this pull, in a weird sense.

As for why that logic is there I believe the rationale might have been to center the window if the display is changed via the bindable, but not on startup. Not super sure it's a behaviour worth preserving, so I'm okay with removing.

bdach (approved):

peppy (approved):
Tested on windows and macOS, seems good

PR #4792, opened 2021-09-24
Fix `ExecutionMode` potentially changing between equality and set calls

peppy:
Hard one to write a test for, but should be obvious why this is needed. An alternative is to make `ExecutionMode` a field-backed computed property and lock on `set`. Either works for me if that is preferred.


peppy:
This manifested only after there was code added in `ensureCorrectExecutionMode` that relied on there being a state change (ie. relying on the previous state always being different from the new one). With the existing implementation I don't think there are adverse consequences without this fix. Still best it's fixed, though.

smoogipoo (approved):

PR #4793, opened 2021-09-26
Remove SQLite specific methods from framework `Storage` classes

peppy:
Not sure how these got here, but they shouldn't be in o!f, as it is considered database backend agnostic.


bdach:
While I agree with these removals, the methods in question are actually being used by the game, so what's the plan there - is a correspondent game-side PR coming, or is this one just supposed to hang around until EF is finally excised?

peppy:
Yeah they can just be inlined at the usage. Luckily neither are doing anything special.

bdach (approved):

PR #4794, opened 2021-09-27
Fix CI failures due to file access denied to log files

peppy:
Not really sure how useful this test scene is, but the fix is easy enough so may as well just make it work properly. 

Initially I was thinking there was some cross-talk from other tests, but it looks like the `performance.log` file that keeps causing this issue is generated (due to many atlas backings being created for the `SpriteText` tests in the same scene) so it's a very local issue.

Fixes CI failures as seen at https://github.com/ppy/osu-framework/pull/4784/checks?check_run_id=3712782332.


smoogipoo (approved):

smoogipoo:
Very timely CI failure :sweat_smile:

smoogipoo (approved):

smoogipoo (approved):

PR #4796, opened 2021-09-27
Fix TestSceneDelayedLoadUnloadWrapper test failures

smoogipoo:
Has been happening intermittently for a while now, only just now had an idea for why it could occur.

Can be tested by adding a `Thread.Sleep(1000)` in `TestBox.load()`.


peppy:
Nice catch.

peppy (approved):

PR #4800, opened 2021-09-28
Fix game hosts exiting immediately on mobile

bdach:
This is a proposal of a fix for framework apps booting to black screen on android, as reported on https://github.com/ppy/osu/discussions/14874 (also in changelog, and other places).

Regressed in #4786, as using the non-virtual private `performGameExit()` neutered this android-local override:

https://github.com/ppy/osu-framework/blob/4ba0b636d97bedb072079f987213900856b43696/osu.Framework.Android/AndroidGameHost.cs#L87-L90

To make matters worse, `IWindow`'s xmldoc straight up lies:

https://github.com/ppy/osu-framework/blob/4ba0b636d97bedb072079f987213900856b43696/osu.Framework/Platform/IWindow.cs#L135-L138

As evidenced by this regression, the above claim that `Run()` is blocking is false on mobile.

There's probably about a dozen ways you could fix this; I tried three of them and they all felt bad to me in different ways. The one I settled on here is using a pre-existing `CanExit` property and adding a soft invariant that `PerformExit()` should never be called if a game host implementation overrides `CanExit` to `false`. This has the added "benefit" (?) of being able to trim the android-local overrides.

The other things I've tried (and can work) is either putting that `CanExit` guard directly in `performExit()`, or removing the private `performExit()` and assuming that if the protected `PerformExit()` is called with `true`, then it should never be overridden to not exit unless a platform doesn't support exit at all. I don't know which is best at this point, further opinions more than welcome.

### DIFF
--- a/osu.Framework.Android/AndroidGameHost.cs
+++ b/osu.Framework.Android/AndroidGameHost.cs
@@ -83,10 +83,5 @@ namespace osu.Framework.Android
 
         public override VideoDecoder CreateVideoDecoder(Stream stream)
             => new AndroidVideoDecoder(stream);
-
-        protected override void PerformExit(bool immediately)
-        {
-            // Do not exit on Android, Window.Run() does not block
-        }
     }
 }
--- a/osu.Framework.iOS/IOSGameHost.cs
+++ b/osu.Framework.iOS/IOSGameHost.cs
@@ -82,11 +82,6 @@ namespace osu.Framework.iOS
             DebugConfig.SetValue(DebugSetting.BypassFrontToBackPass, true);
         }
 
-        protected override void PerformExit(bool immediately)
-        {
-            // we shouldn't exit on iOS, as Window.Run does not block
-        }
-
         public override bool OnScreenKeyboardOverlapsGameWindow => true;
 
         protected override bool LimitedMemoryEnvironment => true;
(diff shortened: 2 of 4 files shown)


smoogipoo:
I'm not sure how I feel that `Run()` doesn't block. Not just the `IWindow` one, but the `GameHost`'s own method. To me it makes more sense if these platforms had some sort of mutex, but I'm not sure how that works in the grand scheme of things.

Does this mean that the game isn't actually running on the main/application thread on mobile platforms? Seems like this is something we'd want to consider for vsync, but then again I have very little knowledge of osuTK internals around mobile platforms.

peppy:
> Does this mean that the game isn't actually running on the main/application thread on mobile platforms? Seems like this is something we'd want to consider for vsync, but then again I have very little knowledge of osuTK internals around mobile platforms.

I'm pretty sure if you block the main thread nothing will be able to draw. Mobile platforms work a bit differently to desktop in that way. For vsync considerations, we're already doing things correctly, which is to say hooking on the window update loop provided by the OS, at least for iOS (search for `CADisplayLink` in osuTK if interested). Android's osuTK implementation is really dodge (does its own thread logic which is not vsync bound) which is likely why our performance there is nothing special.

In the interest of getting android working again I think this PR is fine as it stands, but we definitely want to revisit mobile game loop handling. probably when we eventually are able to drop osuTK completely.

peppy (approved):

smoogipoo (approved):

PR #4801, opened 2021-09-30
Fix minor typo in xmldoc

Susko3:


frenzibyte:
That is not a typo, but an abbrevation of the statement "if and only if". Read https://en.wikipedia.org/wiki/If_and_only_if.

peppy:
This is the second time it's been attempted to be corrected. We should probably just expand it out and not use that term.

PR #4819, opened 2021-10-12
Fix `EnqueueAction` randomly throwing if called from (async) disposal thread

peppy:
This may not fix every incorrect case, but will fix the disposal path (as seen [here](https://github.com/ppy/osu/pull/15038/checks?check_run_id=3865955663)).

Probably "good enough" for the time being?

### DIFF
--- a/osu.Framework/Audio/AudioComponent.cs
+++ b/osu.Framework/Audio/AudioComponent.cs
@@ -45,15 +45,6 @@ namespace osu.Framework.Audio
                 // we don't want consumers to block on operations after we are disposed.
                 return Task.CompletedTask;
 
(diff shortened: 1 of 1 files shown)


peppy:
It's not. In single threaded move it's feasible for the disposal to run on the same managed thread as the other four threads. Because the thread static bools only reset at the start of the next execution, it's stuck in `IsDrawThread` at that point in execution.

Maybe we want to reset these variables immediately?

smoogipoo:
But they're thread static. Whatever state they're run stuck in is irrelevant, unless that same threadrunner thread is somehow running disposals at the end of the frame/in the next input frame?

peppy:
Precisely that, yeah. It's the input thread and we yield it away. I'm pretty sure in such a case the thread statics will be read on the disposal thread?

Issue #4508, opened 2021-06-13
Framework cursor can go outside of the window on the bottom and right edge when cursor is confined and relative mouse is enabled


Susko3:
Alternative title: Drawables that are just touching the bottom/right edges can't be interacted with if mouse is on that edge (when cursor is contained and relative mode is enabled)

To repro:
1. Open `visual-tests` &rarr; `Input` &rarr; `InputManager`
2. Enable `relative mode` and `ConfineMouseMode`
3. Notice that `PassThroughInputManager` stops getting input when the mouse is on the bottom or right edge

If relative mode is disabled, SDL will clamp the cursor 1 pixel away from the edges. But if it's enabled, we have full control of the cursor and we clamp it exactly to the edge:

https://github.com/ppy/osu-framework/blob/dd796a2a81aa65c12c02553043b28363799a3836/osu.Framework/Input/UserInputManager.cs#L41

The cursor being exactly on the edge causes many problems with drawables that expand to that edge (but no further).
Eg. in osu!: you can't change volume, the mouse back button doesn't work as expected, buttons on the edge don't work.

To fix this, we could clamp the cursor slightly away from the edge: if `clientSize` was `1920x1080`, we could clamp it to `1919.999x1079.999`

PR #4825, opened 2021-10-13
Clamp the mouse 1 pixel off the right and bottom edges of the window when confining cursor

Susko3:
Closes https://github.com/ppy/osu-framework/issues/4508


peppy:
Have you cross-tested this with https://github.com/libsdl-org/SDL/pull/4831 already?

peppy:
So this is fixing the way we are handling disabling relative mode, rather than avoiding a glitch caused by cursor clipping near the edge of the window? Because the linked PR changes that last portion.

Issue #4618, opened 2021-07-19
`KeyCombination.ReadableString()` should take user's keyboard layout into account


Susko3:
This is the framework-side issue for https://github.com/ppy/osu/issues/11143.

I think the underlying way keybinds work should remain as it is now, use scancodes to map keys to actions.
What needs changing is how those keybinds are displayed to the user. Right now, it maps scancodes (technically, `InputKey` is used, but it's a 1:1 mapping to scancodes) to keys on a English (US) keyboard.

Going forward, the way keybinds are stringified for display to the user should change. It should respect the curretly selected keyboard layout: if the user presses the key <kbd>Đ</kbd> for a keybind, it should be reflected as <kbd>Đ</kbd> in the UI. Since we will take the current keyboard layout into account, the implementation will become platform-dependant (as each platform will have a different way to get the current keyboard layout, and to convert scancodes/`InputKey`s to user-readable strings).

For platforms using SDL2, [SDL_GetKeyName](https://wiki.libsdl.org/SDL_GetKeyName) seems like the best and easiest to implement option. Trough a simple series of conversions, we can convert `InputKey`s to user-readable strings:

- `InputKey` &rarr; `SDL_Scancode` (reverse of [SDL2Extensions.ToKey](https://github.com/ppy/osu-framework/blob/192b0eeeb1af9457bac12658d36a24c657bbed04/osu.Framework/Platform/SDL2/SDL2Extensions.cs#L13))
- `SDL_Scancode` &rarr; `SDL_Keycode` (https://wiki.libsdl.org/SDL_GetKeyFromScancode) (this part of the conversion is dependant on the current keyboard layout)
- `SDL_Keycode` &rarr; key name (https://wiki.libsdl.org/SDL_GetKeyName)

These key names are specific to the current keyboard layout, and seem to work perfectly, I've tested multiple keyboard layouts on Windows, and the key names returned by `SDL_GetKeyName` are a perfect match to the keys on the keyboard layout. (There are some problems with specific keys on some keyboard layout, but nothing major.)

I would like to implement this, but I'm not sure about some things:
- Since this is platform-specific code, I'm not sure where to put it. We will probably have a unified implementation for SDL2 desktop platforms, and (in the future) separate implementations for Android and iOS (which should fall back to what we currently have for the time being).
- There seem to be inconsistencies in how some `InputKey`s are stringified:
[...]

PR #4834, opened 2021-10-23
Add `ReadableKeyCombinationProvider`

Susko3:
Closes https://github.com/ppy/osu-framework/issues/4618

Added `ReadableKeyCombinationProvider` as described in the issue, accessible via DI.
Replacement for `KeyCombination.ReadableString()`, which should never be used after this PR. (Could possibly remove/obsolete, and convert tests to use `ReadableKeyCombinationProvider`.)

---

I wasn't able to come up with a way to neatly integrate this into `SpriteText`s (as is done with `LocalisableString`).
`LocalisableString` is simply rooted too deep inside `SpriteText`, and adding a `LocalisableString`-esque `KeyCombinationString` is a lot of work, and out of scope for this already big PR.

A generic `UpdateableString` / `BindableUpdateableString` could pave the way to easily add more updateable strings to `SpriteText`s (like a `KeyCombinationString`) in the future.

So for now, consumers can subscribe to `KeymapChanged` and manually update `SpriteText`s.

---

Tested thoroughly on Windows. Some minor testing can be done on Linux and macOS to verify the `KeymapChanged` flow, and that the modifier keys (Ctrl/Win/Command/Super/Alt/Option) conform to user expectations.

---

I have the osu!-side PR ready and tested. Tested that all previous usages of `KeyCombination.ReadableString()` work as expected when changed to `ReadableKeyCombinationProvider.GetReadableKey`.


osu.Framework/Platform/SDL2/SDL2ReadableKeyCombinationProvider.cs
```diff
@@ -0,0 +1,194 @@
...
+            if (string.IsNullOrEmpty(keyname))
+                return base.GetReadableKey(key);
+
+            return keyname.ToUpper();
+        }
+
+        private bool changeableByKeyboardLayout(InputKey key)
+        {
+            return key >= InputKey.A && key < InputKey.LastKey;
+        }
```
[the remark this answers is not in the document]
bdach: That may be true for windows, but isn't generally true for linux. `setxkbmap` has the following options enabled, just as an example:

```
  ctrl                 Ctrl position
  ctrl:nocaps          Caps Lock as Ctrl
  ctrl:lctrl_meta      Left Ctrl as Meta
  ctrl:swapcaps        Swap Ctrl and Caps Lock
  ctrl:ac_ctrl         At left of 'A'
  ctrl:aa_ctrl         At bottom left
  ctrl:rctrl_ralt      Right Ctrl as Right Alt
  ctrl:menu_rctrl      Menu as Right Ctrl
  ctrl:swap_lalt_lctl  Swap Left Alt with Left Ctrl
  ctrl:swap_lwin_lctl  Swap Left Win with Left Ctrl
  ctrl:swap_rwin_rctl  Swap Right Win with Right Ctrl
  ctrl:swap_lalt_lctl_lwin Left Alt as Ctrl, Left Ctrl as Win, Left Win as Left Alt
```

So I'm not sure about constraining the remappability to only what is the windows standard.

In fact if I drop the constraints given here to allow any key to be remapped some of the options above work perfectly fine.

bdach:
Have tested on linux and it works, but [as stated above](https://github.com/ppy/osu-framework/pull/4834#discussion_r739655016) is a bit too overly constrained with the regard as to what can be remapped, when compared to what `setxkbmap` can do. So I'd probably be in favour of dropping that and allowing everything to be remapped.

I guess that creates a bit of an issue in that we will no longer be able to rename these keys to our liking. That could be preserved if we map SDL string key names to our alternatives, although is prone to breakage if SDL ever change their names.

peppy:
Is there any downside to removing the limitations of which keys can be remapped?

bdach:
The one downside would be that we would have limited control of how special keys would be named in the readable representation. `SDL_GetKeyName()` - which is used here to unmap scancodes to key names in the user's current layout - returns a string, and so our [custom overrides that we use to name some special keys](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Input/Bindings/KeyCombination.cs#L324-L569) would presumably stop working reliably (or at all). I'm not sure where `SDL_GetKeyName()` itself gets its strings from but I'm guessing it may be platform-specific.

Susko3:
I've looked into the source of `SDL_GetKeyName()`, and it uses a simple array lookup (`Keycode` &rarr; `Scancode` &rarr; scancode name), and no platform specific APIs. The only platform specific thing is [some overrides for scancode names](https://github.com/libsdl-org/SDL/search?q=SDL_SetScancodeName).

So we can safely make our own override that provide our key names for specific `SDL_Keycode`s.

The current code takes `InputKey`s (which correspond to physical `SDL_Scancode`s) and gives overrides, but the new code (I'm working on it right now, will push when it's ready) will take keyboard layout-adjusted `SDL_Keycode`s and then provide override key names. This will hopefully give correct key names when paired with `setxkbmap`.

Since we will support all keys, there is no need for `changeableByKeyboardLayout()`.

@bdach could you run the [test program](https://gist.github.com/Susko3/85580d02d2ff7ecb19de95a3a76c0b10) and check whether the correct keycodes and scancodes appear when you use `setxkbmap`.
Eg. if you use `  ctrl:lctrl_meta - Left Ctrl as Meta`,
it should give out:
```
INFO: scancode: SDL_SCANCODE_LCTRL
INFO: keycode: SDLK_LGUI
INFO: name: Left GUI
```

From this example, we could very easily provide an override for `SDLK_LGUI` &rarr; `"LSuper"`

bdach:
I've ran an equivalent of the given program but written in C because that was easier to compile quickly. The result is indeed as follows:

```
(starting off with normal layout)

INFO: scancode: SDL_SCANCODE_LCTRL
INFO: keycode: SDLK_LCTRL
INFO: name: Left Ctrl

(setxkbmap -option "ctrl:lctrl_meta" ran here)

INFO: scancode: SDL_SCANCODE_LCTRL
INFO: keycode: SDLK_LGUI
INFO: name: Left GUI
```

Susko3:
I've made some changes to the SDL2 provider. The most important change is that platform specific overrides for key names happen after the conversion to SDL_Keycodes. This way, they should be adjusted for the current keyboard layout / `setxkbmap` mappings.

Restructured to use `TryGetNameFrom...` as IMO it reads better.

osu.Framework/Platform/SDL2/SDL2ReadableKeyCombinationProvider.cs
```diff
@@ -0,0 +1,241 @@
...
+                    name = base.GetReadableKey(inputKey);
+                    return true;
+
+                default:
+                    name = string.Empty;
+                    return false;
+            }
+        }
+
+        protected virtual bool TryGetNameFromKeycode(SDL.SDL_Keycode keycode, out string name)
```
Susko3, line 63: Not so sure about the code/string duplication here. 

All of the key names are copied over from `ReadableKeyCombinationProvider.GetReadableKey()`.
It's fine, but the duplication is unnecessary.

Best way to fix that is to write `SDL_Keycode` &rarr; `InputKey` converter and use `name = base.GetReadableKey(inputKey)`.
Same as in `TryGetNameFromInputKey()`:

https://github.com/ppy/osu-framework/blob/c304df466fd69697ea1b611e874d06d2ec82dd5e/osu.Framework/Platform/SDL2/SDL2ReadableKeyCombinationProvider.cs#L53-L55

But I don't feel like writing yet another converter. Also, this converter would be incomplete, as not all `SDL_Keycode`s appear in `TryGetNameFromKeycode()`. Leading us to the same situation `ToScancode()` was in.

bdach: Alternatively you could declare a list/hashset of keys that should fall back to the base implementation and check against that set? Although I'm not sure that it's that important in the large scale of things.

osu.Framework/Platform/Windows/WindowsReadableKeyCombinationProvider.cs
```diff
@@ -0,0 +1,42 @@
...
+
+using osu.Framework.Input.Bindings;
+using osu.Framework.Platform.SDL2;
+using SDL2;
+
+namespace osu.Framework.Platform.Windows
+{
+    public class WindowsReadableKeyCombinationProvider : SDL2ReadableKeyCombinationProvider
+    {
+        protected override bool TryGetNameFromInputKey(InputKey inputKey, out string name)
```
bdach, line 12: Overriding these hyper-specific protected methods feels like overkill. I would override just the top-level `GetReadableKey()` and call it a day. I wouldn't even bother to make `TryGetNameFromInputKey()` or `TryGetNameFromKeycode()` virtual.

osu.Framework/Input/ReadableKeyCombinationProvider.cs
```diff
@@ -0,0 +1,340 @@
...
+        internal void OnKeymapChanged()
+        {
+            KeymapChanged?.Invoke();
+        }
+
+        /// <summary>
+        /// Returns a human-readable string for a given <see cref="KeyCombination"/>.
+        /// </summary>
+        /// <remarks>
+        /// Consumers should subscribe to <see cref="KeymapChanged"/> and re-generate the readable KeyCombination when the keyboard layout changes.
```
bdach, line 27: nit: can cref

```suggestion
        /// Consumers should subscribe to <see cref="KeymapChanged"/> and re-generate the readable <see cref="KeyCombination"/> when the keyboard layout changes.
```

bdach (commented):
Generally looks good to me, just minor things left. Tested on linux, works as expected.

osu.Framework/Platform/SDL2/SDL2ReadableKeyCombinationProvider.cs
Susko3: A list/hashset won't work, as we don't have an `InputKey` here to use in `base.GetReadableKey(inputKey)`. We're working with keyboard-layout adjusted `SDL_Keycode`s.

It may be important because we may want to localize some key names, eg. `Ctrl` is fine, but `Media Play` should be localized. Best to have only one source of raw strings for key names in that case.

bdach: Good point.

> It may be important because we may want to localize some key names, eg. Ctrl is fine, but Media Play should be localized

This isn't hugely relevant, though - if that happens, with the way we have localisation set up, a static `LocalisableString` member for a given key name will be exposed somewhere, and then it would only be necessary to reference that (i.e. there would be a `KeyNameStrings.MediaPlay`).

osu.Framework/Platform/Windows/WindowsReadableKeyCombinationProvider.cs
Susko3: If you override `GetReadableKey()` you won't get the keys keyboard-layout adjusted, so overriding TryGetNameFromKeycode()` is required.

One could get away with overriding `GetReadableKey()` instead of `TryGetNameFromInputKey()`.
But I think having all left-right agnostic modifier key overrides in the same method is better.

bdach: Okay, fair. The complexity is unfortunate but looks to be unavoidable. I would at least leave some indication why this is done this way (either by xmldoc'ing the protected methods and explaining the difference between them, or via an inline comment somewhere - the former is probably better).

Susko3: On second thought, after writing a part of the xmldoc:

```cs
/// <remarks>
/// Overriden <see cref="InputKey"/>s should be ones that don't map to specific
/// <see cref="SDL.SDL_Scancode"/>s, but do require a key name different from the base implementation.
/// </remarks>
```

Maybe it's better to override `GetReadableKey()`, remove `TryGetNameFromInputKey()` and let the [current overrides in `SDL2ReadableKeyCombinationProvider`](https://github.com/ppy/osu-framework/blob/8bea5b5b530442184b1400297aa98da2471f9bc3/osu.Framework/Platform/SDL2/SDL2ReadableKeyCombinationProvider.cs#L50-L55) get caught at:

https://github.com/ppy/osu-framework/blob/8bea5b5b530442184b1400297aa98da2471f9bc3/osu.Framework/Platform/SDL2/SDL2ReadableKeyCombinationProvider.cs#L22-L24

Since `ToScancode()` already returns `SDL_SCANCODE_UNKNOWN` for the aforementioned `InputKey`s.
(Of course, with the comment in the above code snippet expanded.)

bdach (approved):
I'm pretty okay with this diff. Would be good to get testing on other platforms other than linux prior to merge but tentatively approving.

peppy (approved):

PR #4849, opened 2021-10-31
Add `LocalisableString` support for `TextFlowContainer`

### LINKED ISSUE #4636: TextFlowContainer needs refactoring to allow localisation support

bdach:
I'm PRing this for discussion first and foremost, as I'm not 100% on this diff, but it does look to work, and would close #4636.

https://user-images.githubusercontent.com/20418176/139578675-09c0c0af-1ffc-4fb4-88fb-48806dca5f1a.mp4

I had the original version of this written almost 2 months ago now. Upon going back to it I didn't like some places of it, but then stared at the problem at hand again for a few hours and didn't find solutions which would feel better, so I'm PRing as is.

The one painful thing in this is how to pass ownership of the `ILocalisedBindableString`s to the `TextChunk`s - I opted to expose the localisation manager internally on the text flow and add private properties to handle, but it is definitely unorthodox.

I had also considered making the chunks components but it is a bit painful due to them both being generic (forced hand after other changes) and the fact that the text flow would need to add them to the drawable hierarchy somewhere which feels like overcomplicating things. Suggestions on how to improve very welcome.

A branch with osu!-side changes corresponding to this pull can be previewed [here](https://github.com/bdach/osu/tree/localisable-text-flow). It has an immediate effect, too, due to i.e. some settings panel items like checkboxes having crowdin localisations already but not supporting displaying them up until now due to string-based API.

---

# Breaking changes

## `TextFlowContainer` now accepts `LocalisableString`s rather than raw `string`s

In order to support localisation, all textual arguments passed to `TextFlowContainer` are now `LocalisableString`s. While there is an implicit conversion from `string` to `LocalisableString` that technically makes this change non-API-breaking, the change is breaking from a binary compatibility standpoint.

In case a consumer wants to consume the framework's localisation API for their own uses, care must be taken in use cases where there are custom adjustments made to the text flow's drawables. Any manual adjustments that were made via manual tweaks to `SpriteText`s returned by text flow prior to #4703 must be now reapplied by subscribing to `ITextPart.DrawablePartsRecreated`, which will fire when the text part's visible content has changed.

## `LocalisationManager.GetLocalisedString()` has been renamed to `GetLocalisedBindableString()`

This was done to better indicate the "bindableness" of the return value, as well to facilitate adding a non-bindable variant that returns a raw `string`. That variant is currently `internal`/not public API, but can be considered for exposition if a reasonable use case is presented.

### DIFF
--- a/osu.Framework/Graphics/Containers/TextChunk.cs
+++ b/osu.Framework/Graphics/Containers/TextChunk.cs
@@ -1,11 +1,14 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+#nullable enable
+
 using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 using osu.Framework.Graphics.Sprites;
+using osu.Framework.Localisation;
 
 namespace osu.Framework.Graphics.Containers
 {
@@ -15,12 +18,14 @@ namespace osu.Framework.Graphics.Containers
     public class TextChunk<TSpriteText> : TextPart
         where TSpriteText : SpriteText, new()
     {
-        private readonly string text;
+        private readonly LocalisableString text;
         private readonly bool newLineIsParagraph;
         private readonly Func<TSpriteText> creationFunc;
-        private readonly Action<TSpriteText> creationParameters;
+        private readonly Action<TSpriteText>? creationParameters;
 
-        public TextChunk(string text, bool newLineIsParagraph, Func<TSpriteText> creationFunc, Action<TSpriteText> creationParameters = null)
+        private ILocalisedBindableString? current;
+
+        public TextChunk(LocalisableString text, bool newLineIsParagraph, Func<TSpriteText> creationFunc, Action<TSpriteText>? creationParameters = null)
         {
             this.text = text;
             this.newLineIsParagraph = newLineIsParagraph;
@@ -30,15 +35,21 @@ namespace osu.Framework.Graphics.Containers
 
         protected override IEnumerable<Drawable> CreateDrawablesFor(TextFlowContainer textFlowContainer)
         {
+            current ??= textFlowContainer.Localisation?.GetLocalisedString(text);
+
+            var drawables = new List<Drawable>();
+
             // !newLineIsParagraph effectively means that we want to add just *one* paragraph, which means we need to make sure that any previous paragraphs
             // are terminated. Thus, we add a NewLineContainer that indicates the end of the paragraph before adding our current paragraph.
             if (!newLineIsParagraph)
             {
                 var newLine = new TextNewLine(true);
-                textFlowContainer.AddPart(newLine);
+                newLine.RecreateDrawablesFor(textFlowContainer);
+                drawables.AddRange(newLine.Drawables);
             }
 
-            return CreateDrawablesFor(text, textFlowContainer);
+            drawables.AddRange(CreateDrawablesFor(current?.Value ?? text.ToString(), textFlowContainer));
+            return drawables;
         }
 
         protected virtual IEnumerable<Drawable> CreateDrawablesFor(string text, TextFlowContainer textFlowContainer)
@@ -50,7 +61,7 @@ namespace osu.Framework.Graphics.Containers
             {
                 if (!first)
                 {
-                    Drawable lastChild = sprites.LastOrDefault() ?? textFlowContainer.Children.LastOrDefault();
+                    Drawable? lastChild = sprites.LastOrDefault() ?? textFlowContainer.Children.LastOrDefault();
 
                     if (lastChild != null)
                     {
--- a/osu.Framework/Graphics/Containers/TextChunk.cs
+++ b/osu.Framework/Graphics/Containers/TextChunk.cs
@@ -23,21 +23,6 @@ namespace osu.Framework.Graphics.Containers
         private readonly Func<TSpriteText> creationFunc;
         private readonly Action<TSpriteText>? creationParameters;
 
-        private ILocalisedBindableString? currentBacking;
-
-        private ILocalisedBindableString? current
-        {
-            get => currentBacking;
-            set
-            {
-                if (value == null)
-                    return;
-
-                currentBacking = value;
-                currentBacking.BindValueChanged(_ => RaiseContentChanged());
-            }
-        }
-
         public TextChunk(LocalisableString text, bool newLineIsParagraph, Func<TSpriteText> creationFunc, Action<TSpriteText>? creationParameters = null)
         {
             this.text = text;
@@ -48,7 +33,7 @@ namespace osu.Framework.Graphics.Containers
 
         protected override IEnumerable<Drawable> CreateDrawablesFor(TextFlowContainer textFlowContainer)
         {
-            current ??= textFlowContainer.Localisation?.GetLocalisedBindableString(text);
+            string currentContent = textFlowContainer.Localisation?.GetLocalisedString(text) ?? text.ToString();
 
             var drawables = new List<Drawable>();
 
@@ -61,7 +46,7 @@ namespace osu.Framework.Graphics.Containers
                 drawables.AddRange(newLine.Drawables);
             }
 
-            drawables.AddRange(CreateDrawablesFor(current?.Value ?? text.ToString(), textFlowContainer));
+            drawables.AddRange(CreateDrawablesFor(currentContent, textFlowContainer));
             return drawables;
         }
 
--- a/osu.Framework/Graphics/Containers/TextChunk.cs
+++ b/osu.Framework/Graphics/Containers/TextChunk.cs
@@ -23,7 +23,20 @@ namespace osu.Framework.Graphics.Containers
         private readonly Func<TSpriteText> creationFunc;
         private readonly Action<TSpriteText>? creationParameters;
 
-        private ILocalisedBindableString? current;
+        private ILocalisedBindableString? currentBacking;
+
+        private ILocalisedBindableString? current
+        {
+            get => currentBacking;
+            set
+            {
+                if (value == null)
+                    return;
+
+                currentBacking = value;
+                currentBacking.BindValueChanged(_ => RaiseContentChanged());
+            }
+        }
 
         public TextChunk(LocalisableString text, bool newLineIsParagraph, Func<TSpriteText> creationFunc, Action<TSpriteText>? creationParameters = null)
         {
--- a/osu.Framework/Graphics/Containers/TextChunk.cs
+++ b/osu.Framework/Graphics/Containers/TextChunk.cs
@@ -48,7 +48,7 @@ namespace osu.Framework.Graphics.Containers
 
         protected override IEnumerable<Drawable> CreateDrawablesFor(TextFlowContainer textFlowContainer)
         {
-            current ??= textFlowContainer.Localisation?.GetLocalisedString(text);
+            current ??= textFlowContainer.Localisation?.GetLocalisedBindableString(text);
 
             var drawables = new List<Drawable>();
 
--- a/osu.Framework/Graphics/Containers/TextFlowContainer.cs
+++ b/osu.Framework/Graphics/Containers/TextFlowContainer.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using osu.Framework.Allocation;
+using osu.Framework.Bindables;
 using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Extensions.IEnumerableExtensions;
 using osu.Framework.Localisation;
@@ -123,7 +124,7 @@ namespace osu.Framework.Graphics.Containers
             set
             {
                 Clear();
-                clearParts();
+                parts.Clear();
 
                 AddText(value);
             }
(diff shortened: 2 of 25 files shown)


peppy (dismissed):
Code-wise this looks very sane to me.

osu.Framework/Graphics/Containers/TextChunk.cs
```diff
@@ -15,12 +18,27 @@ namespace osu.Framework.Graphics.Containers
...
+        {
+            get => currentBacking;
+            set
+            {
+                if (value == null)
+                    return;
 
-        public TextChunk(string text, bool newLineIsParagraph, Func<TSpriteText> creationFunc, Action<TSpriteText> creationParameters = null)
+                currentBacking = value;
+                currentBacking.BindValueChanged(_ => RaiseContentChanged());
```
smoogipoo, line 37: I find this pretty weird - raising `ContentChanged` for a parent from every `ITextPart`. Is there a reason you couldn't expose the `LocalisationManager.currentParameters` bindable to bind from `TextFlowContainer`?

smoogipoo: The resulting diff would read something like:

```diff
diff --git a/osu.Framework/Graphics/Containers/ITextPart.cs b/osu.Framework/Graphics/Containers/ITextPart.cs
index 7adc71cce9..76014d2191 100644
--- a/osu.Framework/Graphics/Containers/ITextPart.cs
+++ b/osu.Framework/Graphics/Containers/ITextPart.cs
@@ -18,12 +18,6 @@ public interface ITextPart
         /// </summary>
         IEnumerable<Drawable> Drawables { get; }
 
-        /// <summary>
-        /// Raised when the <see cref="ITextPart"/>'s content changed due to changing language, etc.
-        /// Signals to the parent <see cref="TextFlowContainer"/> that it needs to reload its contents in the following frame.
-        /// </summary>
-        event Action ContentChanged;
-
         /// <summary>
         /// Raised when <see cref="Drawables"/> is reconstructed (e.g. when the user language was changed).
         /// Can be used by consumers to re-apply manual adjustments to the appearance of <see cref="Drawables"/>.
diff --git a/osu.Framework/Graphics/Containers/TextChunk.cs b/osu.Framework/Graphics/Containers/TextChunk.cs
index cb33d0ec1d..e39d310dd2 100644
--- a/osu.Framework/Graphics/Containers/TextChunk.cs
+++ b/osu.Framework/Graphics/Containers/TextChunk.cs
@@ -23,21 +23,6 @@ public class TextChunk<TSpriteText> : TextPart
         private readonly Func<TSpriteText> creationFunc;
         private readonly Action<TSpriteText>? creationParameters;
 
-        private ILocalisedBindableString? currentBacking;
-
-        private ILocalisedBindableString? current
-        {
-            get => currentBacking;
-            set
-            {
-                if (value == null)
-                    return;
-
-                currentBacking = value;
-                currentBacking.BindValueChanged(_ => RaiseContentChanged());
-            }
-        }
-
         public TextChunk(LocalisableString text, bool newLineIsParagraph, Func<TSpriteText> creationFunc, Action<TSpriteText>? creationParameters = null)
         {
             this.text = text;
@@ -48,7 +33,8 @@ public TextChunk(LocalisableString text, bool newLineIsParagraph, Func<TSpriteTe
 
         protected override IEnumerable<Drawable> CreateDrawablesFor(TextFlowContainer textFlowContainer)
         {
-            current ??= textFlowContainer.Localisation?.GetLocalisedString(text);
+            ILocalisedBindableString? current = textFlowContainer.Localisation?.GetLocalisedString(text);
+            current?.UnbindAll();
 
             var drawables = new List<Drawable>();
 
diff --git a/osu.Framework/Graphics/Containers/TextFlowContainer.cs b/osu.Framework/Graphics/Containers/TextFlowContainer.cs
index a592a21d77..47ee77b17a 100644
--- a/osu.Framework/Graphics/Containers/TextFlowContainer.cs
+++ b/osu.Framework/Graphics/Containers/TextFlowContainer.cs
@@ -7,6 +7,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using osu.Framework.Allocation;
+using osu.Framework.Bindables;
 using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Extensions.IEnumerableExtensions;
 using osu.Framework.Localisation;
@@ -132,6 +133,8 @@ public LocalisableString Text
         [Resolved]
         internal LocalisationManager Localisation { get; private set; }
 
+        private readonly IBindable<LocalisationParameters> localisationParameters = new Bindable<LocalisationParameters>();
+
         public TextFlowContainer(Action<SpriteText> defaultCreationParameters = null)
         {
             this.defaultCreationParameters = defaultCreationParameters;
@@ -140,7 +143,9 @@ public TextFlowContainer(Action<SpriteText> defaultCreationParameters = null)
         protected override void LoadComplete()
         {
             base.LoadComplete();
-            recreateAllParts();
+
+            localisationParameters.BindTo(Localisation.CurrentParameters);
+            localisationParameters.BindValueChanged(_ => recreateAllParts(), true);
         }
 
         private void recreateAllParts()
@@ -320,7 +325,6 @@ public override void Clear(bool disposeChildren)
         protected internal ITextPart AddPart(ITextPart part)
         {
             parts.Add(part);
-            part.ContentChanged += scheduleRecreateAllParts;
 
             if (IsLoaded)
                 recreatePart(part);
@@ -337,26 +341,13 @@ public bool RemovePart(ITextPart partToRemove)
             if (!parts.Remove(partToRemove))
                 return false;
 
-            partToRemove.ContentChanged -= scheduleRecreateAllParts;
-
             if (IsLoaded)
                 recreateAllParts();
 
             return true;
         }
 
-        protected override void Dispose(bool isDisposing)
-        {
-            base.Dispose(isDisposing);
-            clearParts();
-        }
-
-        private void clearParts()
-        {
-            foreach (var part in parts)
-                part.ContentChanged -= scheduleRecreateAllParts;
-            parts.Clear();
-        }
+        private void clearParts() => parts.Clear();
 
         private readonly Cached layout = new Cached();
 
diff --git a/osu.Framework/Graphics/Containers/TextPart.cs b/osu.Framework/Graphics/Containers/TextPart.cs
index 42659df150..ffd13edaa5 100644
--- a/osu.Framework/Graphics/Containers/TextPart.cs
+++ b/osu.Framework/Graphics/Containers/TextPart.cs
@@ -15,7 +15,6 @@ public abstract class TextPart : ITextPart
     {
         public IEnumerable<Drawable> Drawables { get; }
         public event Action<IEnumerable<Drawable>> DrawablePartsRecreated;
-        public event Action ContentChanged;
 
         private readonly List<Drawable> drawables = new List<Drawable>();
 
@@ -36,11 +35,5 @@ public void RecreateDrawablesFor(TextFlowContainer textFlowContainer)
         /// to be appended to the <paramref name="textFlowContainer"/>.
         /// </summary>
         protected abstract IEnumerable<Drawable> CreateDrawablesFor(TextFlowContainer textFlowContainer);
-
-        /// <summary>
-        /// Raises <see cref="ContentChanged"/>, signalling to the parent <see cref="TextFlowContainer"/>
-        /// that the contents of this part have changed.
-        /// </summary>
-        protected void RaiseContentChanged() => ContentChanged?.Invoke();
     }
 }
diff --git a/osu.Framework/Localisation/LocalisationManager.cs b/osu.Framework/Localisation/LocalisationManager.cs
index fddc99d6b1..6df96be879 100644
--- a/osu.Framework/Localisation/LocalisationManager.cs
+++ b/osu.Framework/Localisation/LocalisationManager.cs
@@ -12,11 +12,11 @@ namespace osu.Framework.Localisation
 {
     public partial class LocalisationManager
     {
-        private readonly List<LocaleMapping> locales = new List<LocaleMapping>();
+        public IBindable<LocalisationParameters> CurrentParameters => currentParameters;
 
+        private readonly List<LocaleMapping> locales = new List<LocaleMapping>();
         private readonly Bindable<string> configLocale = new Bindable<string>();
         private readonly Bindable<bool> configPreferUnicode = new BindableBool();
-
         private readonly Bindable<LocalisationParameters> currentParameters = new Bindable<LocalisationParameters>(new LocalisationParameters(null, false));
 
         public LocalisationManager(FrameworkConfigManager config)
```

Which imo is much easier to read. And also reveals that `GetLocalisedString()` value is probably a little bit too chunky being a `Bindable` at all times, but that's a separate issue.

bdach: I don't think that is going to work. Or at least, work reliably. The `LocalisedBindableString`s change their contents when the `currentParameters` bindable is changed, too. Therefore this will only work if the `TextFlowContainer`'s callback to recreate contents is ensured to fire after the callbacks that change the localised content in the text parts.

This, I suppose, could be ensured by scheduling, but at that point I'm not sure whether the proposal is much better that what's already here.

smoogipoo: > The `LocalisedBindableString`s change their contents when the `currentParameters` bindable is changed, too

Where is that done? The only thing I saw was the value being immediately pulled out to be passed into `CreateDrawablesFor(string text, ...)`. I thought the purposes of the change in TFC was to recreate all these parts anyway?
My diff above works even though I've done an `UnbindAll()` after the `GetLocalisableString()` call in `TextChunk`, which seemingly indicates what you're saying is not what's actually happening.

bdach: Oh I see, I missed the part where you're taking out a new bindable every time the parameters change, which ensures the good ordering. I thought it was still the one bindable.

I guess I can give it a go then and see if no other issues fall out, although it feels a bit weird to have a bindable and use it as a primitive like that.

smoogipoo: So thinking about this more, I think I understand what you mean. The process you're mentioning is:

1. `TextChunk` calls `GetLocalisableString()` to get the bindable.
2. `TextChunk` binds to value change of this bindable to raise content changed. It also uses the current text to create the drawables.
3. If the localisable content changes via anything other than localisation, such as `LocalisableString` itself changing, then the bound content changed even gets invoked.

But I don't think that (3) is possible in the context it's used here. In `SpriteText` it is because the same bindable is reused, but here bindable always has its default which it was constructed with. And given that `LocalisableString` is also immutable, I think it's a good assumption to say that it can't change.

Also this feeds back into what I mentioned at the end of my comment: that `GetLocalisableString()` is a bit too fat for this sort of use-case anyway - there's no need to use the bindable at all imo.

smoogipoo: Also it may be a good thing to move the localised _value_ retrieval to `LocalisationManager` anyway, just from the context of slimming down `LocalisedBindableString` :wink:

bdach: That would work, yeah. Fine to do that in this pull? It'd just be moving the `ILocalisedBindableString.updateValue()` method up to the localisation manager itself and making it available for text flow.

smoogipoo: In this PR is fine, yep.

bdach: Have pushed changes. One thing I wasn't sure of is to how to name the new method. I ended up renaming `GetLocalisedString()` to `GetLocalisedBindableString()` and exposing the non-bindable variant as `internal GetLocalisedString()`, but unsure if you'll agree on that. Will revert on request and shuffle names across if there are better ideas.

osu.Framework/Graphics/Containers/TextFlowContainer.cs
```diff
@@ -119,17 +117,58 @@ public Anchor TextAnchor
...
+        }
+
+        private void recreateAllParts()
+        {
+            // manual parts need to be manually removed before clearing contents,
+            // to avoid accidentally disposing of them in the process.
+            foreach (var manualPart in parts.OfType<TextPartManual>())
+                RemoveRange(manualPart.Drawables);
+
+            // make sure not to clear the list of parts on accident.
```
peppy, line 158:
```suggestion
            // make sure not to clear the list of parts by accident.
```

peppy (changes requested):
Still looks sane to me code wise.

osu.Framework/Graphics/Containers/TextFlowContainer.cs
```diff
@@ -119,17 +117,58 @@ public Anchor TextAnchor
...
+        {
+            this.defaultCreationParameters = defaultCreationParameters;
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            localisationParameters.BindTo(Localisation.CurrentParameters);
+            localisationParameters.BindValueChanged(_ => recreateAllParts(), true);
```
smoogipoo, line 148: We'll have to see how bad this (and other `IsLoaded` checks) is on performance. The first "big" load will occur on the update thread, so things like markdown containers which make many `SpriteText`s could suffer.

smoogipoo (dismissed):

osu.Framework/Graphics/Containers/TextFlowContainer.cs
peppy: You raise a good point. Would it be feasible to move it to `LoadComplete` as part of this PR? If we want to avoid the bindable thread safety concerns, an early value transfer could be done, followed by the final `BindValueChanged(..., false)` in the current loacation.

smoogipoo: Do you mean inside BDL? It's already in LoadComplete in the PR.

peppy: I meant `LoadAsyncComplete` or whatever that method we have is called.

peppy (approved):

PR #4856, opened 2021-11-02
Allow `TextFlowContainer` parts to be generated asynchronously where possible

peppy:
Adds testing of relevant scenarios too.

### DIFF
--- a/osu.Framework/Graphics/Containers/TextFlowContainer.cs
+++ b/osu.Framework/Graphics/Containers/TextFlowContainer.cs
@@ -133,19 +133,27 @@ namespace osu.Framework.Graphics.Containers
         [Resolved]
         internal LocalisationManager Localisation { get; private set; }
 
-        private readonly IBindable<LocalisationParameters> localisationParameters = new Bindable<LocalisationParameters>();
+        private readonly Bindable<LocalisationParameters> localisationParameters = new Bindable<LocalisationParameters>();
 
         public TextFlowContainer(Action<SpriteText> defaultCreationParameters = null)
         {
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/Containers/TextFlowContainer.cs
```diff
@@ -339,12 +311,59 @@ public bool RemovePart(ITextPart partToRemove)
             if (!parts.Remove(partToRemove))
                 return false;
 
-            if (IsLoaded)
-                recreateAllParts();
-
+            partsCache.Invalidate();
```
smoogipoo, line 314: Is there a reason why this `Remove` path uses invalidate, whereas the `Add` path above uses `if (partCache.IsValid) recreatePart(part);`?

peppy: I was just copying across the existing logic. On calling `AddPart` it does a recreate of *only* the added part. Calling `RemovePart` would trigger a full `recreate`, which I've kept in this PR.

I haven't looked into whether this can/should be changed.

bdach: You could recreate everything on every add, although that feels a bit wasteful. You could also not recreate on removal too and just snip that one part, I suppose; the recreation there was done mostly for simplicity.

peppy: I'd probably say further changes should be done if/when necessary unless it's going to reduce complexity.

bdach (approved):
Looks fine to me :+1:

PR #4859, opened 2021-11-03
Allow BufferedDrawNode's frame buffer to be clipped to the root node

khang06:
fixes ppy/osu#15434

caching the root node might cause problems in the future, but there doesn't seem to be any better way of getting the window resolution from here that i can think of

### DIFF
--- a/osu.Framework/Graphics/BufferedDrawNode.cs
+++ b/osu.Framework/Graphics/BufferedDrawNode.cs
@@ -38,6 +38,7 @@ namespace osu.Framework.Graphics
         private RectangleF screenSpaceDrawRectangle;
         private Vector2 frameBufferScale;
         private Vector2 frameBufferSize;
+        private IDrawable? rootNodeCached;
 
         public BufferedDrawNode(IBufferedDrawable source, DrawNode child, BufferedDrawNodeSharedData sharedData)
             : base(source)
@@ -54,6 +55,8 @@ namespace osu.Framework.Graphics
             screenSpaceDrawRectangle = Source.ScreenSpaceDrawQuad.AABBFloat;
             DrawColourInfo = Source.FrameBufferDrawColour ?? new DrawColourInfo(Color4.White, base.DrawColourInfo.Blending);
             frameBufferScale = Source.FrameBufferScale;
+
+            clipDrawRectangle();
 
             frameBufferSize = new Vector2(MathF.Ceiling(screenSpaceDrawRectangle.Width * frameBufferScale.X), MathF.Ceiling(screenSpaceDrawRectangle.Height * frameBufferScale.Y));
             DrawRectangle = SharedData.PixelSnapping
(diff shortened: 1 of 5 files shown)


peppy:
This is a pretty nice solution 👍

smoogipoo (changes requested):
I'm surprised this works, but it seems fine. It will need to be exposed from `BufferedContainer` as well, either as a property or a ctor param.

osu.Framework/Graphics/BufferedDrawNode.cs
```diff
@@ -178,6 +181,26 @@ private void returnViewport()
...
+            {
+                rootNode = Source;
+                while (rootNode.Parent != null)
+                    rootNode = rootNode.Parent;
+                rootNodeCached = rootNode;
+            }
+
+            // Clip the screen space draw rectangle to the bounds of the root node
+            RectangleF clipBounds = new RectangleF(rootNode.ScreenSpaceDrawQuad.TopLeft, rootNode.ScreenSpaceDrawQuad.Size);
+            screenSpaceDrawRectangle.Intersect(clipBounds);
```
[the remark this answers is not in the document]
khang06: not sure what you mean. it caches a reference to the drawable itself and not the size of it, so i don't think i need to check if its size changed since it updates the size based on that regardless on if its cached or not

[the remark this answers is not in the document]
smoogipoo: Ah, I misread, all good.

peppy:
> I'm surprised this works, but it seems fine. It will need to be exposed from `BufferedContainer` as well, either as a property or a ctor param.

Is that requirement to fix sliders specifically? Or for potential future usage? I was thinking about this but couldn't come up with an answer as to whether that should be exposed to the user, as it requires a bit of thinking as to whether you'd want it or not. I'd argue that in most usages you wouldn't want it.

smoogipoo:
> Or for potential future usage?

It would be for hold notes in mania, for example, which can extend far off screen.

peppy:
It could probably be argued that it should be enabled by default? Something like:

`bool RenderOutsideScreenBounds { get; set; } = true`

smoogipoo:
Yeah I'd agree to being enabled by default too.

khang06:
would it really be useful having it be able to be changed outside of the constructor? PixelSnapping also seems to be readonly

smoogipoo:
Having it only in the ctor is enough imo.

osu.Framework/Graphics/Containers/BufferedContainer.cs
```diff
@@ -26,8 +26,8 @@ namespace osu.Framework.Graphics.Containers
     public class BufferedContainer : BufferedContainer<Drawable>
     {
         /// <inheritdoc />
-        public BufferedContainer(RenderbufferInternalFormat[] formats = null, bool pixelSnapping = false)
-            : base(formats, pixelSnapping)
+        public BufferedContainer(RenderbufferInternalFormat[] formats = null, bool pixelSnapping = false, bool clipToRootNode = false)
```
peppy, line 29:
```suggestion
        public BufferedContainer(RenderbufferInternalFormat[] formats = null, bool pixelSnapping = false, bool clipToRootNode = true)
```

I think we agreed that this is best left enabled by default.

osu.Framework/Graphics/Containers/BufferedContainer.cs
```diff
@@ -246,9 +246,10 @@ public bool RedrawOnScale
         /// <param name="formats">The render buffer formats attached to the frame buffers of this <see cref="BufferedContainer"/>.</param>
         /// <param name="pixelSnapping">Whether the frame buffer position should be snapped to the nearest pixel when blitting.
         /// This amounts to setting the texture filtering mode to "nearest".</param>
-        public BufferedContainer(RenderbufferInternalFormat[] formats = null, bool pixelSnapping = false)
+        /// <param name="clipToRootNode">Whether the frame buffer should be clipped to be contained in the root node.</param>
```
peppy, line 249:
```suggestion
        /// <param name="clipToRootNode">Whether the frame buffer should be clipped to be contained in the root node. Should only be disabled if you plan to draw to a region outside (or larger than) the screen.</param>
```

Might be better to clarify when this should be used, as it's a pretty special circumstance.

osu.Framework/Graphics/Containers/BufferedContainer.cs
```diff
@@ -246,9 +246,10 @@ public bool RedrawOnScale
...
+        public BufferedContainer(RenderbufferInternalFormat[] formats = null, bool pixelSnapping = false, bool clipToRootNode = false)
```
peppy, line 250:
```suggestion
        public BufferedContainer(RenderbufferInternalFormat[] formats = null, bool pixelSnapping = false, bool clipToRootNode = true)
```

peppy (changes requested):

peppy (approved):

smoogipoo (approved):

PR #4863, opened 2021-11-05
Move `CacheDrawnFrameBuffer` property to constructor, replacing `clipToRootNode`

peppy:
As seen by the [regressions](https://github.com/ppy/osu/issues/15475#issuecomment-961382253), with the current invalidation structure we have, clipping to the root node does not mix well with `BufferedContainer` in the case it is caching its framebuffer.

To minimise user error, this PR hard couples the state of caching with the state of clipping. The rationale is that when not buffering, you would never want to draw to an area larger than the screen, while when you *are* caching you are usually aware of the size of the container, and may be explicitly using caching to draw to a larger area than is currently being displayed.

We can iterate on this if other use cases come up which don't obide by these rules.

Closes #15475.

---

# Breaking Changes

## `BufferedContainer.CacheDrawnFrameBuffer` has been moved to a constructor argument

```csharp
// old code:
var bufferedContainer = new BufferedContainer() { CacheDrawnFrameBuffer = true };

// new code:
var bufferedContainer = new BufferedContainer(cachedFrameBuffer: true);
```

## `IEffect.CacheDrawnEffect` has been removed

We had no usages of this. If you were using it, nest the effected content in a `BufferedContainer` with `cachedFrameBuffer` set to `true`.

### DIFF
--- a/osu.Framework.Tests/Visual/Containers/TestSceneCachedBufferedContainer.cs
+++ b/osu.Framework.Tests/Visual/Containers/TestSceneCachedBufferedContainer.cs
@@ -120,7 +120,7 @@ namespace osu.Framework.Tests.Visual.Containers
             private readonly SpriteText count;
 
             public CountingBox(bool rotating = false, bool moving = false, bool cached = false)
-                : base(useCachedFrameBuffer: cached)
+                : base(cachedFrameBuffer: cached)
             {
                 this.rotating = rotating;
                 this.moving = moving;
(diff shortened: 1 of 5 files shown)


smoogipoo (approved):
I'm not sure I fully agree with them being tied together under one variable like this - there are potential uses for drawing outside the window and not caching at the same time, but this is saved by the fact that the user can call `ForceRedraw()` every frame anyway...

PR #4865, opened 2021-11-05
Add `SDL2DesktopWindowTextInput` and polish `ITextInputSource`s

Susko3:
Spiritual successor to https://github.com/ppy/osu-framework/pull/4709.

That PR is trying to do too much and is sitting unmerged too long due to some fundamental problems.

This PR takes two ideas from it: splitting off `GameWindowTextInput` and using `SDL_Start/StopTextInput`.
Importantly, there a no breaking changes to `TextBox`, or to how `ITextInputSource` works.

This PR is mostly code quality improvements so it's easier to add changes to `SDL2DesktopWindowTextInput` in the future (such as IME support or event-based text input).

SDL still uses the exact same text input flow as before, the only difference is that `SDL_StopTextInput()` is called when text input isn't needed (and `SDL_StartTextInput` when it's needed again).

TextBox remains as broken as it is right now.

#### *Some closing words*

Event-based text input shouldn't be abandoned. If implemented properly, it should be much better than the current hacky pooling approach. I think it's important to iron out how text events interact with / block `OnKeyDown`.

IME support is coming, need to do a bit more polish. Wanted to get this SDL text input change in because I don't want it ending up as #4709, sitting unmerged.


osu.Framework/Platform/MacOS/MacOSGameHost.cs
```diff
@@ -38,7 +38,13 @@ public override IEnumerable<string> UserStoragePaths
             }
         }
 
-        public override ITextInputSource GetTextInput() => Window == null ? null : new MacOSTextInput(Window);
+        public override ITextInputSource GetTextInput()
+        {
+            if (Window is OsuTKWindow osuTKWindow)
+                return new MacOSTextInput(osuTKWindow);
```
bdach, line 44: The old code would return `MacOSTextInput` regardless of the underlying window implementation type. Needs testing on macOS to ensure this doesn't break anything (which I cannot do).

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -396,6 +396,10 @@ public virtual void Create()
             SDL.SDL_SetHint(SDL.SDL_HINT_WINDOWS_NO_CLOSE_ON_ALT_F4, "1");
             SDL.SDL_SetHint(SDL.SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS, "1");
 
+            // We want text input to only be active when SDL2DesktopWindowTextInput is active.
+            // SDL activates it by default, so we deactivate it on startup.
```
bdach, line 400: How was this determined? Is it documented somewhere?

osu.Framework.iOS/Input/IOSTextInput.cs
```diff
@@ -59,16 +53,13 @@ public void EnsureActivated()
...
+            view.KeyboardTextField.UpdateFirstResponder(false);
         }
 
-        public event Action<string> OnNewImeResult
-        {
-            add { }
-            remove { }
-        }
+        public event Action<string> OnNewImeComposition { add { } remove { } }
+        public event Action<string> OnNewImeResult { add { } remove { } }
```
bdach, line 63: Nitpick / personal opinion but I much preferred the old formatting (goes for all places that have these one-lined).

osu.Framework.Android/Input/AndroidTextInput.cs
```diff
@@ -59,28 +50,40 @@ private void keyDown(Keycode arg, KeyEvent e)
...
 
-        public bool ImeActive => false;
-
-        public event Action<string> OnNewImeComposition;
-        public event Action<string> OnNewImeResult;
-
         public void Activate()
         {
+            view.KeyDown += keyDown;
+            view.CommitText += commitText;
```
bdach, line 56: Is this safe to subscribe to when not on the UI thread? Asking because this pull takes these subscription out of the `RunOnUiThread` block.

osu.Framework/Platform/SDL2DesktopWindow.cs
Susko3: Testing methodology: enable SDL event logging and then check whether `SDL_TEXTINPUT` events get sent.

```cs
SDL.SDL_SetHint(SDL.SDL_HINT_EVENT_LOGGING, "2");
```

Documented here:
https://github.com/libsdl-org/SDL/blob/cd66c050fe276368c6e7317b60e9d5ca13770cad/src/video/SDL_video.c#L568-L577

`SDL_HasScreenKeyboardSupport()` returns `SDL_FALSE`, at least on Windows on a PC without touch.

osu.Framework.Android/Input/AndroidTextInput.cs
Susko3: `KeyDown` and `CommitText` are our own events, so no problem there.

https://github.com/ppy/osu-framework/blob/f6f7d058a8bdae751737e2186c295cff89348e9b/osu.Framework.Android/AndroidGameView.cs#L22-L25

osu.Framework/Platform/SDL2DesktopWindow.cs
bdach: I would probably link that source inline for posterity.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -865,14 +873,20 @@ private unsafe void handleTextInputEvent(SDL.SDL_TextInputEvent evtText)
             if (ptr == IntPtr.Zero)
                 return;
 
-            string text = Marshal.PtrToStringUTF8(ptr) ?? "";
+            string text = Marshal.PtrToStringUTF8(ptr) ?? string.Empty;
 
-            foreach (char c in text)
-                ScheduleEvent(() => KeyTyped?.Invoke(c));
+            TextInput?.Invoke(text);
```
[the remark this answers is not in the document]
Susko3: Forgot to schedule, my mistake.

osu.Framework/Platform/MacOS/MacOSGameHost.cs
Susko3: `MacOSTextInput` used the SDL2 part of `GameWindowTextInput` anyway, so no change here.

bdach: Are you sure? Because that doesn't quite check out with my reading of this line in `master`:

https://github.com/ppy/osu-framework/blob/f25033f804be6e369fbe6df81cb8725a327dd8e3/osu.Framework/Platform/MacOS/MacOSGameHost.cs#L41

If there is a `Window` at all, it would get `MacOSTextInput`, would it not? I'm not saying that that *needs* to exist - maybe the SDL implementation handles everything fine - but there were some workarounds in `MacOSTextInput`, so I'm hesitant to approve without prior testing as to they're really not needed on macOS.

Susko3: Let's look at the current `MacOSTextInput` on `master`. The only thing that it's doing different from `GameWindowTextInput` is `HandleKeyPress`: (all other fields are private and used only in the overriding `HandleKeyPress`)

https://github.com/ppy/osu-framework/blob/f25033f804be6e369fbe6df81cb8725a327dd8e3/osu.Framework/Platform/MacOS/MacOSTextInput.cs#L24

The overridden `HandleKeyPress` only gets subscribed to if the window is `OsuTKWindow`, which is never the case:
https://github.com/ppy/osu-framework/blob/f25033f804be6e369fbe6df81cb8725a327dd8e3/osu.Framework/Input/GameWindowTextInput.cs#L60-L61

So `MacOSTextInput` is just an empty class over `GameWindowTextInput`, which has been 1:1 replicated in `SDL2DesktopWindowTextInput`.

bdach: Ok, fair. Thanks for spelling it out.

osu.Framework.iOS/Input/IOSTextInput.cs
peppy: These can be replaced with raw `event`s:

```suggestion
        public event Action<string> OnNewImeComposition;
        public event Action<string> OnNewImeResult;
```

peppy:
Initial testing on macOS shows no worse behaviour than previous. IME composition still fails as it does on `master`.

bdach:
Linux also looks good here, didn't notice any regressions.

osu.Framework.iOS/Input/IOSTextInput.cs
bdach: Since this has been brought up twice I've applied the above suggestion in the relevant places.

bdach: Never mind, this trips inspections due to unused events. I even ran inspectcode locally but it looked like they weren't going to be reported...

peppy (approved):

PR #4866, opened 2021-11-07
Remove obsolete framework input settings

frenzibyte:


bdach:
This doesn't compile.

bdach (changes requested):
see above

osu.Framework.Tests/Configuration/InputConfigManagerTest.cs
```diff
@@ -28,9 +28,6 @@ public void TestOldConfigPersists()
                 host.Run(new TestGame((h, config) =>
                 {
                     storage = h.Storage;
-#pragma warning disable 618
-                    config.SetValue(FrameworkSetting.CursorSensitivity, 5.0);
-#pragma warning restore 618
```
bdach, line 33: this entire test needs to go, no? it was testing that the old config pathway was getting persisted in the new format, which is no longer relevant/applicable, is it not?

osu.Framework/Configuration/FrameworkConfigManager.cs
```diff
@@ -59,14 +52,6 @@ public FrameworkConfigManager(Storage storage, IDictionary<FrameworkSetting, obj
             new TrackedSetting<bool>(FrameworkSetting.ShowLogOverlay, v => new SettingDescription(v, "Debug Logs", v ? "visible" : "hidden", "Ctrl+F10")),
             new TrackedSetting<Size>(FrameworkSetting.WindowedSize, v => new SettingDescription(v, "Screen resolution", $"{v.Width}x{v.Height}")),
             new TrackedSetting<WindowMode>(FrameworkSetting.WindowMode, v => new SettingDescription(v, "Screen Mode", v.ToString(), "Alt+Enter")),
-#pragma warning disable 618
-            new TrackedSetting<double>(FrameworkSetting.CursorSensitivity, v => new SettingDescription(v, "Cursor Sensitivity", v.ToString(@"0.##x"), "Ctrl+Alt+R to reset")),
-            new TrackedSetting<string>(FrameworkSetting.IgnoredInputHandlers, v =>
-            {
-                bool raw = !v.Contains("Raw");
-                return new SettingDescription(raw, "Raw Input", raw ? "enabled" : "disabled", "Ctrl+Alt+R to reset");
-            }),
```
bdach, line 68: removing these causes the game-side `Ctrl`+`Alt`+`R` hotkey to reset sensitivity/raw input to not work anymore.

Susko3:
Shouldn't `IHasCursorSensitivity` also be removed, since it's no longer used.

frenzibyte:
Going to close this for now as I thought it to be a simple change but turns out it breaks behaviour rather than removing obsolete methods.

PR #4868, opened 2021-11-08
Exclude BASS default audio device from the public `AudioDeviceNames` list

frenzibyte:
Closes ppy/osu#15498, see https://github.com/ppy/osu/issues/15498#issuecomment-962596983 for explaination.

This comes in two commits: 
 1. https://github.com/ppy/osu-framework/commit/802a3404ad9000507ed364fcff36a907d9ba9586 excludes the BASS default audio device (which is now included in all platforms) from the `AudioDeviceNames` list as per the title.
 2. https://github.com/ppy/osu-framework/commit/8cece706288ac23f0d7ea6441047f2d48461e620 doesn't change behaviour but explicitly chooses the BASS default audio device in `setAudioDevice` as a fallback, rather than blindly looking for the first `IsDefault = true` device.


osu.Framework/Audio/AudioManager.cs
```diff
@@ -342,13 +350,13 @@ private void syncAudioDevices()
 
             audioDevices = updatedAudioDevices;
 
-            // Bass should always be providing "No sound" device
-            Trace.Assert(audioDevices.Count > 0, "Bass did not provide any audio devices.");
+            // Bass should always be providing "No sound" and "Default" device
+            Trace.Assert(audioDevices.Count >= 2, "Bass did not provide any audio devices.");
```
smoogipoo, line 354: Shouldn't this PR also set the `BASS_CONFIG_DEV_DEFAULT` config for this assert to pass?

smoogipoo:
I'm a little bit confused about your second point, why is looking for a `.IsDefault` device considered "blind"? Isn't looking for an explicitly named "Default" device more blind?

osu.Framework/Audio/AudioManager.cs
```diff
@@ -258,7 +266,7 @@ private bool setAudioDevice(string deviceName = null)
                 return true;
 
             // try using the system default device
-            if (setAudioDevice(audioDevices.FindIndex(d => d.Name != deviceName && d.IsDefault)))
+            if (setAudioDevice(audioDevices.FindIndex(d => d.Name == bass_default_device)))
```
bdach, line 269: Yeah this makes no sense to me either. Not only am I confused as to why using `IsDefault` is wrong, when it looks to be delegating to the correct BASS flag:

2021-11-08-091628_1380x136_scrot

But additionally this changes behaviour because if the user's selected device *was* the default, this change will make it so that default device initialisation will be attempted *twice* for what seems like no real reason, which wasn't the case before.

frenzibyte:
> I'm a little bit confused about your second point, why is looking for a `.IsDefault` device considered "blind"? Isn't looking for an explicitly named "Default" device more blind?

It's more about the OS default device having a constant index and a constant name, while `IsDefault` applies to the BASS default device and the real audio device it maps to, so I thought checking explicitly for the "Default" device may be better.

Maybe changing it to an integer constant would be better, given that the BASS default device is always the first in the list, meaning it has an index of `1`. (similar to `Bass.NoSoundDevice`, even though `Bass.DefaultDevice` exists but it's a different index used for different purposes)

bdach:
I didn't understand any of the comments above but received the following piece of information from twitch chat:

> On Windows and macOS, when the BASS_CONFIG_DEV_DEFAULT option is enabled, the BASS_DEVICE_DEFAULT flag will be set on both the "Default" device and the device that it currently maps to. The BASS_DEVICE_DEFAULTCOM flag is only used on Windows Vista and newer. 

(source). Which is why the `IsDefault` thing is misleading.

And as a response to the above I'm fine with hardcoding the 1 index or whichever works (although the docs mention [something about `-1`](https://www.un4seen.com/doc/#bass/BASS_CONFIG_DEV_DEFAULT.html)?)

frenzibyte:
Initializing with `-1` index also works but that can't be possible with the way `AudioManager` initializes a device, as it uses the index given to fetch the audio device from the `audioDevices` list, and `-1` can't be used.

osu.Framework/Audio/AudioManager.cs
```diff
@@ -258,7 +266,7 @@ private bool setAudioDevice(string deviceName = null)
                 return true;
 
             // try using the system default device
-            if (setAudioDevice(audioDevices.FindIndex(d => d.Name != deviceName && d.IsDefault)))
+            if (setAudioDevice(bass_default_device))
```
bdach, line 269: Unsure if that huge of a deal, but if the previous condition returns the device with index 1 and initialising that fails, this will still try to initialise it again which it didn't use to do.

frenzibyte: That can be the case if someone specifically set the audio device name to "Default", just like if someone were to set the audio device name to "No sound", there's the potential to initialize it twice if it somehow failed to do so on first try.

This could be resolved by using `audioDeviceNames.FindIndex` and add `2`, but that's arguably out of scope from this PR.

bdach: I don't know if it's out of scope if this pull touches that behaviour? And I'm not really sure what the "2" thing means, either. It would be enough to store the index of the previous tried device and compare that against `bass_default_device`, no?

peppy: Is this pending further action?

bdach: It's not a huge concern on my side. I wouldn't mind merging this without addressing this comment, I suppose, I'm leaving this hanging mostly to see if anyone else has an opinion about it.

peppy: I think it's fine to stay like this. Should be non-fatal at least.

peppy (approved):

smoogipoo:
Bass test failures are relevant here.

peppy:
@frenzibyte looks to be actually failing on macOS test runs. Can reproduce locally.

peppy (changes requested):
Tested to work locally with osu!, but tests are broken.

frenzibyte:
The fact that it only failed for macOS when it's supposed to have failed for all platforms makes me kinda nervous whether the audio tests are actually working. Can't test on other platforms right now though.

osu.Framework/Audio/AudioManager.cs
```diff
@@ -24,6 +24,22 @@ namespace osu.Framework.Audio
...
+        /// </remarks>
+        private const int bass_default_device = 1;
+
+        /// <summary>
+        /// The beginning index of the BASS real audio devices.
+        /// </summary>
+        /// <remarks>
+        /// The first two devices are <see cref="Bass.NoSoundDevice"/> and <see cref="bass_default_device"/>.
+        /// </remarks>
+        private const int bass_first_real_device = 2;
```
peppy, line 41: Expose as `internal` and use this to replace all usages of `2` (two in this file and one in `AudioManagerWithDeviceLoss`).

peppy (changes requested):

osu.Framework/Audio/AudioManager.cs
frenzibyte: Renamed it to `BASS_INTERNAL_DEVICES` for the usages, was gonna define another `bass_first_real_device` but thought it's pointless.

peppy: `BASS_INTERNAL_DEVICE_COUNT`?

frenzibyte: Was double-minded on that 😅

peppy (approved):

frenzibyte:
Resolved [yet another test failure](https://github.com/ppy/osu-framework/runs/4222683494?check_suite_focus=true) that is random this time due to schedulers getting involved, which made it a time-dependant one.

Will stress run tests locally in hopes nothing is still failing.

peppy:
I think this is fine to get in?

peppy (approved):

smoogipoo (approved):
lgtm

PR #4870, opened 2021-11-08
Set window title after creation

Susko3:
Built SDL with debug, and got this warning when opening visual-tests:

image

The title does get set as expected, SDL probably has a good reason to complain.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -680,9 +680,9 @@ namespace osu.Framework.Platform
                 if (Window != null)
                 {
                     Window.SetupWindow(Config);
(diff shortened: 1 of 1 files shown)


peppy:
I also got this in the past. I believe we set it before creating the window to ensure it is correct at the point of window creation, but that seems like the wrong way to go about it.

Susko3:
osu! sets the title much later anyway, in https://github.com/ppy/osu-framework/blob/28602c5cc1c992abe9f0d5075d29129ef49b9c8d/osu.Framework/Platform/GameHost.cs#L700

which calls

https://github.com/ppy/osu/blob/6944151486e677bfd11f2390163aca9161defbbf/osu.Desktop/OsuGameDesktop.cs#L148

peppy (approved):

PR #4875, opened 2021-11-09
Fix audio mixers potentially not showing in `AudioMixer` visualiser

peppy:
The underlying events would potentially be fired too early and never reach the hooking methods due to the fact the invocation call site is run on the audio thread, before the binding may have completed.

Also done some restructuring for a future PR which allows showing an identifying string on the visualiser.

Can be tested via the existing `TestSceneAudioMixerVisualiser` while running in multi-thread mode (previously the new mixers would usually not display).

### DIFF
--- a/osu.Framework/Audio/Mixing/AudioMixer.cs
+++ b/osu.Framework/Audio/Mixing/AudioMixer.cs
@@ -14,6 +14,11 @@ namespace osu.Framework.Audio.Mixing
     /// </summary>
     public abstract class AudioMixer : AdjustableAudioComponent, IAudioMixer
     {
+        /// <summary>
+        /// The handle for this mixer.
+        /// </summary>
+        public int Handle { get; protected set; }
+
         private readonly AudioMixer? globalMixer;
 
         /// <summary>
--- a/osu.Framework/Audio/Mixing/AudioMixer.cs
+++ b/osu.Framework/Audio/Mixing/AudioMixer.cs
@@ -14,11 +14,6 @@ namespace osu.Framework.Audio.Mixing
     /// </summary>
     public abstract class AudioMixer : AdjustableAudioComponent, IAudioMixer
     {
-        /// <summary>
-        /// The handle for this mixer.
-        /// </summary>
-        public int Handle { get; protected set; }
-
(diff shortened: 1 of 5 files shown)


peppy:
Regression tests could be added on the aforementioned test scene if required for this one.

osu.Framework/Audio/Mixing/AudioMixer.cs
```diff
@@ -14,6 +14,11 @@ namespace osu.Framework.Audio.Mixing
     /// </summary>
     public abstract class AudioMixer : AdjustableAudioComponent, IAudioMixer
     {
+        /// <summary>
+        /// The handle for this mixer.
+        /// </summary>
+        public int Handle { get; protected set; }
```
bdach, line 20: This one is a little unfortunate because I *think* the audio subsystem was designed to look mostly agnostic from BASS but this handle is an implementation detail of BASS. On the other side it's kinda neither here nor there as `MixerDisplay` unceremoniously calls `BassMix.MixerGetChannels()` anyway...

I dunno, could be wrapped up a little nicer (eg. by exposing an abstract method returning channel levels here) but I'm not going to go and argue about it or anything.

peppy: Yeah considering how dependent on bass everything around this is I didn't feel it was worth the time abstracting out.

smoogipoo: This class is supposed to be bass-agnostic. Is there a reason you couldn't `(mixer as BassAudioMixer)` instead?

smoogipoo (changes requested):
Rest looks fine.

osu.Framework/Audio/Mixing/AudioMixer.cs
peppy: Can but felt pretty pointless (the whole visualiser depends on this).

smoogipoo: The visualiser isn't the only class using `AudioMixer`.

bdach (approved):

smoogipoo (approved):

PR #4876, opened 2021-11-09
Show audio mixer identifier strings on the visualiser

peppy:
Found it hard to track what's what. This helps somewhat.

- [x] Depends on #4875 to avoid merge conflicts.

### DIFF
--- a/osu.Framework/Graphics/Audio/DrawableAudioMixer.cs
+++ b/osu.Framework/Graphics/Audio/DrawableAudioMixer.cs
@@ -18,7 +18,7 @@ namespace osu.Framework.Graphics.Audio
         [BackgroundDependencyLoader]
         private void load(AudioManager audio)
         {
-            mixer = audio.CreateAudioMixer();
+            mixer = audio.CreateAudioMixer(Name);
             mixer.Effects.BindTo(Effects);
         }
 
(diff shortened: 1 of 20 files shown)


osu.Framework/Graphics/Audio/DrawableAudioMixer.cs
```diff
@@ -18,7 +18,7 @@ public class DrawableAudioMixer : AudioContainer, IAudioMixer
         [BackgroundDependencyLoader]
         private void load(AudioManager audio)
         {
-            mixer = audio.CreateAudioMixer();
+            mixer = audio.CreateAudioMixer(Name);
```
bdach, line 21: Unfortunate that this is timing dependent (will only work if set prior to load). Not sure if it's worth spending the time to add support for arbitrary-timed changes. If we had initialisers then we could use that but alas...

bdach (commented):
Most changes that aren't dependencies look good, just one minor note.

osu.Framework/Graphics/Audio/DrawableAudioMixer.cs
peppy: Not sure it matters too much. Very rare you'd be setting the name of a drawable at any point after construction.

bdach (approved):

Issue #4677, opened 2021-08-10
iOS deployment issues (summary)


peppy:
I'm making this issue just to document the current issues we are facing that are inhibiting iOS deployment of osu!. We are looking to seek external help for this as there is no one with the time/expertise combo local to the project.

# Critical

## ~~AOT failure (https://github.com/ppy/osu/issues/13302)~~ [Resolved]

This has been reported upstream, along with a solution by @frenzibyte involving a package rollback. Needs a second confirmation on the xamarin-side issue (https://github.com/xamarin/xamarin-macios/issues/12353) to hopefully get the ball rolling (as this does not look like our issue).

![](https://cdn.discordapp.com/attachments/589331078574112768/872728621490536469/unknown.png)

## ~~Testing of new bass library version~~ [Resolved]

We recently updated to ManagedBass 3.0 to consume newer underlying un4seen bass libraries (https://github.com/ppy/osu-framework/pull/4291). Since this we are seeing a compile time version mismatch. This looks to have been attempted to be fixed upstream (https://github.com/ManagedBass/ManagedBass/pull/93) but this fix does not work for us. It also seems like an incorrect way to fix this issue.

```csharp
warning MSB3243: No way to resolve conflict between "ManagedBass, Version=3.0.0.0, Culture=neutral, PublicKeyToken=null" and "ManagedBass, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null". Choosing "ManagedBass, Version=3.0.0.0, Culture=neutral, PublicKeyToken=null" arbitrarily. 
```

Testing with a local nupkg of ManagedBass (and the two addons we use) will be required to get all three working together, at which point a pull request needs to be made upstream to fix the actual issue.

Audio also needs to be tested to still work (we have not been able to compile to even test this since the BassMix conversion).

## ~~Cross nupkg testing~~ [Fulfilled]

osu! needs to be re-tested with all the above fixed.

## Deadlock investigation

We already attempted to fix random deadlocks during the startup sequence but have not yet understood a solid fix for it. We are also unsure if it's something our end or not. This needs a deeper investigation, either via somehow getting a deadlock to happen in a thread dumpable state, or using another method of diagnosis.

Previous fixes were centered around iOS having a `SynchronizationContext` while other platforms don't, and as we were doing some incorrect `async/await` it was causing deadlocks. Not sure if this is still the issue (we have fixed a few more cases since, see https://github.com/ppy/osu/issues/13231).

[...]

PR #4879, opened 2021-11-10
Downgrade `Microsoft.Diagnostics.Runtime` back to avoid AOT issues with iOS

frenzibyte:
- Resolves https://github.com/ppy/osu/issues/13302
- Resolves the first point mentioned in https://github.com/ppy/osu-framework/issues/4677
- Works around https://github.com/mono/mono/issues/21188.

This is the simplest solution for the time being. Ideally we wouldn't want this package to be included in Xamarin projects to begin with, but not really sure if there's a simple way to do that without using `#if` directives (not sure we want that).


peppy:
Please add an inline comment mentioning this so we don't accidentally upgrade it in the future.

peppy (approved):

PR #4880, opened 2021-11-10
Add more user-friendly variant of the exponential damp interpolation

ekrctb:
This function can be used to to interpolate a value in a frame-rate independent manner without maintaining additional variables.
It should replace the current frame-rate dependent transition used in osu e.g. <https://github.com/ppy/osu/blob/f9d5abff8a8d3b6752343c185eb7767a3944a532/osu.Game/Screens/Play/SkipOverlay.cs#L132>.

### DIFF
--- a/osu.Framework/Utils/Interpolation.cs
+++ b/osu.Framework/Utils/Interpolation.cs
@@ -31,10 +31,27 @@ namespace osu.Framework.Utils
         {
             if (@base < 0 || @base > 1)
                 throw new ArgumentOutOfRangeException(nameof(@base), $"{nameof(@base)} has to lie in [0,1], but is {@base}.");
-            if (exponent < 0)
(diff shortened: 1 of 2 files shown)


osu.Framework/Utils/Interpolation.cs
```diff
@@ -31,12 +31,29 @@ public static double Damp(double start, double final, double @base, double expon
...
-            if (exponent < 0)
-                throw new ArgumentOutOfRangeException(nameof(exponent), $"{nameof(exponent)} has to be bigger than 0, but is {exponent}.");
 
             return Lerp(start, final, 1 - Math.Pow(@base, exponent));
         }
 
+        /// <summary>
+        /// Interpolate the current value towards the target value based on the elapsed time.
+        /// If this function is applied to the current value at every frame,
+        /// the approximately same value is calculated at the same time regardless of the frame rate.
```
bdach, line 41: I don't really understand what this is trying to say (especially the "same value is calculated at the same time" part). Maybe just leave the part about framerate independence and call it a day?

bdach (approved):

PR #4881, opened 2021-11-10
Use upstream `ManagedBass` packages again and update to 3.1.0

frenzibyte:
- Fulfills the second point in https://github.com/ppy/osu-framework/issues/4677 ("Testing of new bass library version")

Also resolves visible `ManagedBass` reference conflicts in iOS projects:

CleanShot 2021-11-10 at 13 42 25@2x

Tested this update to work on macOS and iOS, would appreciate testing on other platforms just in case.


bdach:
Tested linux briefly, looks to work, although again, I haven't tested *very* much.

bdach (approved):

peppy:
Tested on windows, looks to work correctly.

peppy (approved):

PR #4882, opened 2021-11-10
Update native BASSmix libraries to 2.4.11

frenzibyte:
- [ ] Depends on #4881 (for testability)

Finally resolves the iOS native linking failures noticed in https://github.com/ppy/osu-framework/issues/4677#issuecomment-903776830, as the offending symbols seem to have been removed altogether in this new version.

Tested this update to work on macOS and iOS, would appreciate testing on other platforms just in case.


bdach:
Can you link source for binaries? Is it just latest release, or something special? Asking in order to verify their SHA.

frenzibyte:
It's from the latest releases:
 - Windows: http://www.un4seen.com/download.php?bassmix24
 - macOS: http://www.un4seen.com/download.php?bassmix24-osx
 - Linux: http://www.un4seen.com/download.php?bassmix24-linux
 - Android: http://www.un4seen.com/forum/?topic=13225.0 (http://www.un4seen.com/files/bass24-android.zip)
 - iOS: http://www.un4seen.com/forum/?topic=10910.0 (http://www.un4seen.com/files/bass24-ios.zip)

peppy (approved):
files match latest releases

PR #4884, opened 2021-11-10
Work around `System.Memory` assembly reference conflicts in iOS

frenzibyte:
Resolves errors seen in osu!framework tests iOS project specifically (e.g. `TestInterpolation.cs:L50`):

CleanShot 2021-11-10 at 14 57 51@2x

See https://jaylee.org/archive/2019/03/31/using-span-of-t-in-xamarin-cross-targeted-projects.html for more information.

Hopefully that can be removed once all projects target one framework (in .NET 6 I guess?)


osu.Framework.iOS.props
```diff
@@ -85,4 +85,17 @@
          See https://github.com/mono/mono/issues/21188 -->
     <PackageReference Include="Microsoft.Diagnostics.Runtime" Version="2.0.161401"/>
   </ItemGroup>
+  <!-- Workaround for System.Memory conflicts on MSBuild v16
+       See https://jaylee.org/archive/2019/03/31/using-span-of-t-in-xamarin-cross-targeted-projects.html -->
+  <Target
+    Name="VS16_RemoveSystemMemory"
+    BeforeTargets="FindReferenceAssembliesForReferences"
+    Condition="'$(MSBuildVersion)' &gt;= '16.0'">
```
peppy, line 93: The escaping of that >= is waky shit.

frenzibyte: I was under the thought that that is necessary, but it apparently isn't.

peppy: As long as you're sure it still works. I was fine with it matching the referenced article.

peppy (approved):

osu.Framework.iOS.props
smoogipoo: Looks like it should be escaped (as per standard xml practice). https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-conditions?view=vs-2022

`Note: In XML, the characters < and > must be escaped`

smoogipoo (changes requested):
Please escape it

smoogipoo (approved):

PR #4886, opened 2021-11-10
Fix crashes in `AudioManager` due to no threading guarantee on `CreateAudioMixer`

peppy:
The "add" calls were coming from the update thread while the "remove" calls were happening on the audio thread. Dangerous.


frenzibyte (approved):

bdach (approved):
Seems reasonable

PR #4888, opened 2021-11-11
Add methods to retrieve the levels/data of a mixer's output

nekodex:
Previously there were only methods to retrieve the levels/data of a mixer channel, but not the mixer output itself.


smoogipoo:
Where's this used? This class is internal so it's only usable inside o!f as it is. You'd need to expose it via `IAudioMixer` if you wanted to use it in osu!, which is where I assume this is being used?

nekodex:
Uh yeah, this was a pre-emptive thing and wasn't being used yet. I'll just close for now and add the changes to the PR where it's going to be used instead.

PR #4890, opened 2021-11-12
Preserve BASS FX and BASSmix library symbols from getting stripped in iOS

frenzibyte:
- [x] Depends on #3044 (for mergeability)
- Works around https://github.com/xamarin/xamarin-macios/issues/13359.

Normally, any native library wrapper class under iOS [would have to import from `__Internal`](https://docs.microsoft.com/en-us/xamarin/ios/platform/native-interop#static-libraries) in order to p/invoke the methods successfully. 

But `ManagedBass` played it differently on the BASS addons (BASS FX, BASSmix, etc.) starting from 3.0.0, by importing from the native library name as-is, but instead using `dllmap`s to map the native library names to `__Internal` to be able to p/invoke on iOS (see https://github.com/ManagedBass/ManagedBass/issues/30).

And that appears to be working ([minus one worked around issue](https://github.com/ManagedBass/ManagedBass/issues/76)), but not quite with mtouch's strip symbols feature... (which is enabled on Release configuration)

It appears that mtouch isn't aware of the `dllmap` configuration and thinks that the BASS FX and BASSmix symbols aren't used anywhere (while they're, but imported from `bass_fx`/`bassmix` with `dllmap` rather than `__Internal` directly), and therefore strips the symbols out and causing `EntryPointNotFoundEexception` when attempting to call any of the library methods.

---

The only workaround to this as far as I've seen from trying out several other ways is to tell mtouch to specifically preserve each symbol of the library, and I've done that via [this quick script I've wrote](https://gist.github.com/frenzibyte/a09112e31cb28bc8becfa76d6a2bd23a), which takes defined and external symbols from the library, and prepend the `--nosymbolstrip=` flag to each.

Other ways like using the `Preserve` attribute against `ManagedBass.Fx.BassFx` can't work, because `BassFx` imports the methods from `bass_fx`, while mtouch would be looking for imports from `__Internal` to actually preserve the symbols.


osu.Framework.iOS.props
```diff
@@ -1,16 +1,25 @@
...
     <MtouchVerbosity></MtouchVerbosity>
   </PropertyGroup>
+  <PropertyGroup>
+    <!-- Normally, native library symbols don't get stripped when are used by the framework via p/invokes,
+         but that is not the case for BASS FX and BASSmix, as the ManagedBass wrappers import from "bass_fx/bassmix"
+         while the libraries themselves get included to the app as "__Internal" instead (with dllmaps from "bass_fx/bassmix" to "__Internal").
+         And therefore mtouch would consider the libraries symbols as unused and strip them away.
+
+         In order to work around it, specify each symbol of each library with the "nosymbolstrip={symbol}" flag to preserve it from getting stripped.
+         Use the following script to generate the "nosymbolstrip" flags: https://gist.github.com/frenzibyte/a09112e31cb28bc8becfa76d6a2bd23a -->
```
bdach, line 18: this script should not be in a random gist, it should be in this repository.

extra points if you can make it update this props file automatically in the correct manner.

osu.Framework.iOS/Properties/AssemblyInfo.cs
```diff
@@ -2,10 +2,24 @@
...
+[assembly: LinkWith("libavcodec.a", SmartLink = false, ForceLoad = true)]
+[assembly: LinkWith("libavdevice.a", SmartLink = false, ForceLoad = true)]
+[assembly: LinkWith("libavfilter.a", SmartLink = false, ForceLoad = true)]
+[assembly: LinkWith("libavformat.a", SmartLink = false, ForceLoad = true)]
+[assembly: LinkWith("libavutil.a", SmartLink = false, ForceLoad = true)]
+[assembly: LinkWith("libbass.a", SmartLink = false, ForceLoad = true)]
+[assembly: LinkWith("libbass_fx.a", SmartLink = false, ForceLoad = true)]
+[assembly: LinkWith("libbassmix.a", SmartLink = false, ForceLoad = true)]
+[assembly: LinkWith("libswresample.a", SmartLink = false, ForceLoad = true)]
+[assembly: LinkWith("libswscale.a", SmartLink = false, ForceLoad = true)]
```
bdach, line 25: what's this? nothing in the PR description seems to explain it, as far as i can tell. looks like something for another pull entirely.

bdach (changes requested):

osu.Framework.iOS.props
frenzibyte: Well there are linked gists existing across the framework already, but sure can turn it to a framework wiki page if that's fine?

As for updating the props file automatically, sounds feasible, will look at it.

bdach: not a wiki page. an executable script *in* source. that you can run without copying it over from wiki or a gist or wherever.

osu.Framework.iOS/Properties/AssemblyInfo.cs
frenzibyte: Comes from the dependency PR #3044, just wanted things to go in a straight path rather than causing conflicts between.

osu.Framework.iOS.props
frenzibyte: Ah I see, will look at that.

frenzibyte: Added script in source with automatic project updating support. But the generated symbols will still have to be copied across to osu! and templates, unfortunately. As they can't access the native libraries to extract the symbols from, unless it's fine to attempt finding the osu-framework folder, which I'm not really sure about doing.

I'm planning on reporting the original issue to Xamarin anyways, as I'm hoping we wouldn't require this long term.

bdach:
Needs conflict resolution.

bdach:
I can't verify the script because linux `nm` seems to be incompatible with macOS shared objects, so going to dismiss review and leave it to be verified by anyone else that has the hard/software to do so.

(Albeit I will say that a few things look weird - why no `#!` preprocessor directive? what's with the `builtin` near `cd`? but those may well be some macOS quirks.)

frenzibyte:
> I can't verify the script because linux `nm` seems to be incompatible with macOS shared objects, so going to dismiss review and leave it to be verified by anyone else that has the hard/software to do so.
> 
> (Albeit I will say that a few things look weird - why no `#!` preprocessor directive? what's with the `builtin` near `cd`? but those may well be some macOS quirks.)

Just to clarify things out, I'm not really competent with sh/bash scripting, but felt this is enough to be correct.

Regarding the `#!` preprocessor, to be frank I've looked at the `build.sh` script in osu!framework and noticed it didn't have that included, so thought it shouldn't be included in this script as well.

As for the `builtin cd`, there is actually no point in having that now that I checked what `builtin` is actually for, will remove that.

frenzibyte:
Reported upstream at https://github.com/xamarin/xamarin-macios/issues/13359.

osu.Framework.iOS.props
peppy: > I'm planning on reporting the original issue to Xamarin anyways, as I'm hoping we wouldn't require this long term.

About that (check upstream report). Safe to say this workaround is sticking with us until NET6.

peppy (approved):
Looks to work as expected in terms of generation. I haven't tested this actually runs on a device but it looks like it should.

PR #4892, opened 2021-11-13
Fix .webm videos reporting incorrect duration

Opelkuh:
Possibly affected some other formats as well.  

[Docs for `AVFormatContext.duration`](https://ffmpeg.org/doxygen/4.1/structAVFormatContext.html#ad0ea78ac48f5bb0a15a15c1c472744d9)

### DIFF
--- a/osu.Framework/Graphics/Video/VideoDecoder.cs
+++ b/osu.Framework/Graphics/Video/VideoDecoder.cs
@@ -31,7 +31,7 @@ namespace osu.Framework.Graphics.Video
         /// <summary>
         /// The duration of the video that is being decoded. Can only be queried after the decoder has started decoding has loaded. This value may be an estimate by FFmpeg, depending on the video loaded.
         /// </summary>
-        public double Duration => stream == null ? 0 : duration * timeBaseInSeconds * 1000;
+        public double Duration { get; private set; }
 
         /// <summary>
         /// True if the decoder currently does not decode any more frames, false otherwise.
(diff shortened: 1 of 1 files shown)


bdach:
FFmpeg docs are pretty arcane to me, but as far as I can tell, this is correct, if going off of the docs for [`AVStream`](https://ffmpeg.org/doxygen/4.1/structAVStream.html#a4e04af7a5a4d8298649850df798dd0bc):

> `int64_t AVStream::duration`
>
> Decoding: duration of the stream, in stream time base.

And after checking it looks like this is enough to make the tests from #4893 pass, too, so yeah.

bdach (dismissed):

bdach:
Actually now that I look again there is a [potentially-related test failure](https://github.com/ppy/osu-framework/runs/4199451160?check_suite_focus=true) in the original CI run...

Opelkuh:
Sorry, I forgot to clarify the issue. `AVStream` duration was used correctly before this PR but .webm files have the stream's duration set to AV_NOPTS_VALUE (-9223372036854775808) which made it fallback to `AVFormatContext::duration` but then when calculating the Duration it still used the stream's time base instead of `AV_TIME_BASE`.

The CI test failure is probably some race condition caused by the clock in tests running so fast as nothing should've change for the .mp4 file that it's using. I can't reproduce it locally.

bdach:
Upon revisiting this, I could not reproduce the test failures when locally repeatedly running `TestSceneVideo`. I've also realised that this pull doesn't change the behaviour of the test scene involved as the value of `Duration`, as well as the method of calculating it, stays the same.

So while there may well be a race condition, it won't be caused by this pull, and so I think it's fine to proceed, but I'm going to hold back from merging for now to make sure there is agreement on that.

bdach (approved):

peppy (approved):

PR #4894, opened 2021-11-15
Revert no longer necessary iOS `GameHost` DI changes

frenzibyte:
Reverts https://github.com/ppy/osu-framework/pull/4263 as it appears to be no longer required from testing with osu! on Release configuration multiple times.


peppy:
Going to wait until I can test on a physical device thoroughly before merging this one.

peppy:
Honestly, I say we just leave this in place until migrating to net6.0 (or 7.0) support for iOS/android as it's not hurting anyone how it currently is...

Issue #4773, opened 2021-09-14
`KeyBindingHandler` consumes repeats when original key is not handled


peppy:
https://github.com/ppy/osu-framework/blob/1f7f73735e1dae2add77f6e6ad417a322acb0b1b/osu.Framework/Input/Bindings/KeyBindingContainer.cs#L118-L119

Consumption should be based on whether the original (non-repeat) event was handled or not, rather than whether any binding matches it. As seen in https://github.com/ppy/osu/issues/14428#issuecomment-918865134.

Should probably be addressed after https://github.com/ppy/osu-framework/pull/2601 is merged, as a lot of the logic around this has changed.

peppy:
This might get a bit complex to handle, as it probably should be making the decision on a per-binding level (ie. rather than returning early in the `Handle` function, it would need to do a lookup for the currently pressed keys in the incoming event, match and ensure it should actually be blocked).

A better path forward may be to just remove/obsolete `SendRepeats` alongside the `UIEvent` change, as it shouldn't be required with the new event arguments.

Joehuu:
Closing as resolved.

PR #4896, opened 2021-11-16
Remove `KeyBindingContainer.SendRepeats`

peppy:
Reasoning mentioned in https://github.com/ppy/osu-framework/issues/4773.

See a pretty shocking oversight in db22119df which went unnoticed until now as tests were focused on the non-event-based repeat handling (which is removed in this PR). Existing tests cover the fail case - revert the commit to confirm.

# Breaking Changes

## `KeyBindingContainer.SendRepeats` no longer exists

Use the `KeyBindingPressEvent` provided in `OnPressed` delegates to identify (and ignore or handle) repeats instead.

### DIFF
--- a/osu.Framework/Input/Bindings/KeyBindingContainer.cs
+++ b/osu.Framework/Input/Bindings/KeyBindingContainer.cs
@@ -75,12 +75,6 @@ namespace osu.Framework.Input.Bindings
         }
 
         /// <summary>
-        /// Override to enable or disable sending of repeated actions (disabled by default).
-        /// Each repeated action will have its own pressed/released event pair.
-        /// </summary>
-        protected virtual bool SendRepeats => false;
-
-        /// <summary>
         /// Whether this <see cref="KeyBindingContainer"/> should attempt to handle input before any of its children.
         /// </summary>
         protected virtual bool Prioritised => false;
@@ -118,9 +112,6 @@ namespace osu.Framework.Input.Bindings
                     return false;
 
                 case KeyDownEvent keyDown:
-                    if (keyDown.Repeat && !SendRepeats)
-                        return pressedBindings.Count > 0;
-
                     if (handleNewPressed(state, KeyCombination.FromKey(keyDown.Key), keyDown.Repeat))
                         return true;
 
--- a/osu.Framework/Input/Bindings/KeyBindingContainer.cs
+++ b/osu.Framework/Input/Bindings/KeyBindingContainer.cs
@@ -49,6 +49,7 @@ namespace osu.Framework.Input.Bindings
 
         private readonly Dictionary<IKeyBinding, List<Drawable>> keyBindingQueues = new Dictionary<IKeyBinding, List<Drawable>>();
         private readonly List<Drawable> queue = new List<Drawable>();
+        private List<Drawable> keyRepeatInputQueue;
 
         /// <summary>
         /// The input queue to be used for processing key bindings. Based on the non-positional <see cref="InputManager.NonPositionalInputQueue"/>.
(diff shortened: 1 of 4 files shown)


peppy:
Blocking temporarily. A couple of oversights I'm working on still.

peppy:
Turns out that key repeat (via event flow) wasn't really handled correctly in `KeyBindingContainer` until now. I've restructured things to work roughly as expected.

The existing test coverage seems to be enough to cover this behaviour.

bdach:
I've fired off a nitpick commit to correct a slightly strange test step name.

In general this diff looks sane to me. One thing I'm not sure of, is that when `SimultaneousBindingMode` is not `None`, repeats are maybe a touch counterintuitive in that when I press and hold <kbd>Ctrl</kbd>+<kbd>A</kbd> in the key bindings grid test scene, the initial press activates `A`, `Ctrl`, and `Ctrl_A` actions, but then repeats only get fired for whichever single key was pressed last, which is maybe a little counter-intuitive. But I'm not really super sure what the behaviour *should* be there either, so maybe best to leave the simple case working as-is and tackle that edge case if it ever comes up...

bdach (approved):

peppy:
Yeah, it isn't a 1:1 with expectations but I'd rather address that as a follow up, I think. Keep in mind it's working with the last *action*, not *bindings*, so in a simple case where `Ctrl-A` is bound to an action that will still get correctly repeated.

On macOS (and windows from my recollection), the last non-modifier is the key to be repeated, but modifiers are not included in the equation. So the true behaviour if we're looking for parity would be for the modifier `Ctrl-A` to repeat even in your scenario. We can make this happen if/when the requirement comes up.

osu.Framework/Input/Bindings/KeyBindingContainer.cs
```diff
@@ -176,15 +168,29 @@ protected override bool Handle(UIEvent e)
...
+            if (pressedActions.Count == 0)
+                return false;
+
+            // A simplistic approach to key repeat (that mostly matches OS level implementations) is that the last binding - or action - to
+            // trigger is the one and only action to repeat.
+            T action = pressedActions.Last();
+
+            var pressEvent = new KeyBindingPressEvent<T>(state, action, true);
+
+            return keyRepeatInputQueue.FirstOrDefault(d => triggerKeyBindingEvent(d, pressEvent)) != null;
```
smoogipoo, line 182: This should have a few more conditions, as per https://github.com/ppy/osu-framework/blob/05920eb3cbef9c72005cdaf5e91a510bcfacbd57/osu.Framework/Input/KeyEventManager.cs#L25-L26

But I understand that's a bit complex to get done in this PR, so looks fine as it is.

smoogipoo (approved):

osu.Framework/Input/Bindings/KeyBindingContainer.cs
smoogipoo, line 182: Actually, this is just incorrect because the input queues are per-keybinding...

peppy: Which case would be incorrect? The idea is that the repeat is only sent to the last pressed key binding, where this queue is taken from directly.

peppy: I saw that. But those conditionals also weren't applied to the other path as far as I can see (I guess it's less important because it's an immediate firing?) so I didn't think too much of it.

smoogipoo:
Going to need osu!-side fixes for this as well, fwiw.

osu.Framework/Input/Bindings/KeyBindingContainer.cs
smoogipoo: Right, I see. Seems okay on a second think...

smoogipoo: The other path being what? In both cases it's the non-repeat path, and the expectation is that it doesn't contain `!IsAlive` and `!IsPresent` items in the first place.

peppy: Yeah, that part. I'll add the conditionals in a follow-up PR.

PR #4897, opened 2021-11-17
Add Orientation setting to framework config

honguyenminh:
This allows games to have an orientation setting, mainly for mobile devices.


peppy:
I don't get it. This needs implementation as well, right? To actually make it work?

honguyenminh:
The implementation will be done in the app that uses the framework. Apps might choose different flag for different options here.

In osu's case the setting will be shown in osu.Game setting overlay, implementation for Android will be in a class in osu.Android and same goes for iOS if needed later.
This is just to provide the setting for the game using the framework.
Or am I doing it wrong and this should be implemented in the framework?

peppy:
If the setting is at the framework side, the implementation should also be done here. We generally keep all low-level handling (like rotation) at the framework side, so this *is* the correct place to put it.

Susko3:
Do keep in mind how [osu.Android/GameplayScreenRotationLocker](https://github.com/ppy/osu/blob/master/osu.Android/GameplayScreenRotationLocker.cs) works, so you may need to add additional options to the framework setting, or a "locking" logic to facilitate this.

May be useful to you, if you haven't seen it:
https://github.com/ppy/osu-framework/wiki/Testing-local-framework-checkout-with-other-projects

honguyenminh:
Implemented the logic

Client app can inherit the class to change the behavior, and/or wrap it in another class to extend functionality

osu.Framework/Configuration/Orientation.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework.Configuration
+{
+    /// <summary>
+    /// Screen orientation
+    /// </summary>
+    public enum Orientation
```
bdach, line 9: The enum should be arguably called `ScreenOrientation` rather than an arbitrary `Orientation`, and the xmldoc is barely (if at all) usable. It should explain that this is primarily intended for use with mobile platforms.

osu.Framework/Platform/OrientationManager.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using osu.Framework.Bindables;
+using osu.Framework.Configuration;
+using osu.Framework.Graphics;
+
+namespace osu.Framework.Platform
+{
+    public abstract class OrientationManager : Component
+    {
+        protected Bindable<Orientation> OrientationBindable;
```
bdach, line 12: Should be `{ get; protected set; }` or similar.

osu.Framework/Platform/OrientationManager.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        protected OrientationManager() { }
```
bdach, line 14: What is the point of this constructor? It should not exist.

osu.Framework/Platform/OrientationManager.cs
```diff
@@ -0,0 +1,23 @@
...
+        public OrientationManager(FrameworkConfigManager config)
+        {
+            OrientationBindable = config.GetBindable<Orientation>(FrameworkSetting.Orientation);
+            OrientationBindable.BindValueChanged(OrientationSettingChangedHandler);
+        }
+
+        public abstract void OrientationSettingChangedHandler(ValueChangedEvent<Orientation> value);
```
bdach, line 21: This would fit existing naming conventions better if it was named `OnScreenOrientationChanged` or similar. Also, the xmldoc from the android implementation should be moved here to the base implementation.

osu.Framework.Android/Platform/AndroidOrientationManager.cs
```diff
@@ -0,0 +1,67 @@
...
+using osu.Framework.Platform;
+using osu.Framework.Configuration;
+using osu.Framework.Bindables;
+using Android.App;
+
+namespace osu.Framework.Android.Platform
+{
+    public class AndroidOrientationManager : OrientationManager
+    {
+        public ScreenOrientation CurrentScreenOrientation { get; internal set; }
```
bdach, line 14: I don't see why this would need to be a field at all, or be publically exposed like it is. Would it not be enough to map the framework `Orientation` to android `ScreenOrientation` in the current `OrientationSettingChangedHandler`, and just delete this field?

osu.Framework.Android/Platform/AndroidOrientationManager.cs
```diff
@@ -0,0 +1,67 @@
...
+        public virtual ScreenOrientation OrientationToScreenOrientation(Orientation orientation)
+        {
+            switch (orientation)
+            {
+                case Orientation.Landscape:
+                    return ScreenOrientation.UserLandscape;
+                case Orientation.Portrait:
+                    return ScreenOrientation.UserPortrait;
+                case Orientation.Auto:
+                default:
```
bdach, line 50: I prefer to handle unknown enum members by throwing `ArgumentOutOfRangeException` rather than assuming that `Auto` was wanted as done here.

osu.Framework.Android/Platform/AndroidOrientationManager.cs
```diff
@@ -0,0 +1,67 @@
...
+        public AndroidOrientationManager(FrameworkConfigManager config, Activity gameActivity)
+            : base(config)
+        {
+            GameActivity = gameActivity;
+        }
+
+        /// <summary>
+        /// Get/set the game activity that have its orientation controlled.
+        /// </summary>
+        protected Activity GameActivity { get; set; }
```
bdach, line 31: Does not need to be `protected` until there are actual subclasses of this.

osu.Framework.Android/AndroidGameWindow.cs
```diff
@@ -18,9 +18,9 @@ public class AndroidGameWindow : OsuTKWindow
 
         public override bool Focused => true;
 
-        public override Platform.WindowState WindowState
+        public override Framework.Platform.WindowState WindowState
```
bdach, line 21: Why was this changed?

bdach (changes requested):

osu.Framework/Configuration/Orientation.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        Landscape = 0,
+        Portrait = 1,
+        Auto = 2
```
Susko3, line 13: Since we'll need `Locked` for `GameplayScreenRotationLocker`, might be useful to have it here in framework so we can easily switch to it.

bdach: I find having "landscape", "portrait", "auto" and "locked" in one enum type turbo weird. It feels so ill-defined, the last one isn't even specific (locked to landscape or portrait?). Not to mention that this enum should also work just as well for iOS, eventually.

[Apple's take on this](https://developer.apple.com/documentation/uikit/uiinterfaceorientation) - although apparently deprecated in favour of [some inscrutable "trait" thing](https://developer.apple.com/documentation/uikit/uitraitcollection) - seems to be designed towards allowed orientation and I think that it's a saner choice. Just basically have a flag enum like so:

```csharp
public enum ScreenOrientation
{
    LandscapeLeft,
    LandscapeRight
    Portrait,
    PortraitUpsideDown,
    AnyLandscape = LandscapeLeft | LandscapeRight,
    AnyPortrait = Portrait | PortraitUpsideDown,
    Any = AnyLandscape | AnyPortrait
}
```

It seems like that should map back to android & iOS-specific flags rather nicely...?

Susko3: Your enum looks much better, should give the user much more options.

`AnyLandscape` seems like a good default, since we set the default resolution on desktop to 1366x768.

We could have `Locked` logic local to `GameplayScreenRotationLocker` then.

Since we give the user more options in-app, I feel like these options should *always* override the screen rotation quick setting on Android. And possibly have an additional enum entry for 'follow system', which would let the system's auto-rotate setting control the orientation. (This is unnecessary for osu!, most users will be fine with the current options.)

**Rationale for this:**
I think most users set auto-rotate once and never change it. And when the user explicitly sets the orientation to landscape, they expect it to rotate both ways, regardless of set auto-rotate. This is how the YouTube app (will rotate both ways in fullscreen) and many landscape-only games behave.

Because of `GameplayScreenRotationLocker`, we'll have no problems with the orientation accidentally changing during gameplay and users being angry because they "clearly set auto-rotate to off in settings".

osu.Framework.Android/AndroidGameWindow.cs
honguyenminh: Because `osu.Framework.Android.Platform` namespace is used by the Android orientation manager, and the current namespace is `osu.Framework.Android` so `Platform.` will be understood by the compiler as `osu.Framework.Android.Platform` instead of `osu.Framework.Platform`

Just there to make it compile basically.

Susko3: image

The newly-added `AndroidOrientationManager` is in a new folder.

Maybe it's cleaner to do
```cs
using WindowState = osu.Framework.Platform.WindowState;
```

Or just move `AndroidOrientationManager` to `osu.Framework.Android`. After all, all other platform things such as `AndroidGameWindow` or `AndroidGameHost` reside in `osu.Framework.Android`.

osu.Framework.Android/Platform/AndroidOrientationManager.cs
honguyenminh: Because we would need this exposed to implement the locking mechanism in osu.

Susko3: That is not necessarily required, as one can do `.TriggerChange` on the framework setting bindable to force framework to set the correct orientation again.

Or just run the current screen orientation trough `OrientationToScreenOrientation` when turning off locked orientation.

bdach: I'm a strong believer that the game side rotation locking flow should still be using the framework setting added here, rather than manually sidestepping it by directly accessing the activity.

osu.Framework.Android/Platform/AndroidOrientationManager.cs
honguyenminh: Will be `private` in next commits

osu.Framework.Android/AndroidGameWindow.cs
bdach: Yeah the platform namespace suffix is probably not necessary. Pretty much everything residing in the android project is platform specific anyway.

osu.Framework.Android/Platform/AndroidOrientationManager.cs
Susko3: `Window.Attributes.ScreenOrientation` seems like a promising place to get the current orientation (direction specific) and then apply that.

osu.Framework/Platform/OrientationManager.cs
honguyenminh: you mean `protected Bindable<Orientation> OrientationBindable { get; set; }`
or
`public Bindable<Orientation> OrientationBindable { get; protected set; }`?

I think the latter but not sure

osu.Framework/Platform/OrientationManager.cs
honguyenminh: So that people can't initialize the class without initializing the config to avoid bugs. But yeah, I'll make it private, protected doesn't make sense here.

honguyenminh: Will be deleted in next commits

osu.Framework/Configuration/Orientation.cs
honguyenminh: Should I change everything else's name to match too (`ScreenOrientationManager`, `AndroidScreenOrientationManager`), or just change the enum only?

osu.Framework.Android/Platform/AndroidOrientationManager.cs
honguyenminh: So we should implement and expose a lock and unlock method here too, right?

Honestly I don't think any game would actually use this apart from osu though, even other mobile rhythm games don't. But having a choice is still nice, I guess.

osu.Framework.Android/Platform/AndroidOrientationManager.cs
honguyenminh: Will be changed in next commits

osu.Framework.Android/AndroidGameWindow.cs
honguyenminh: Will move to outer namespace and revert this commit

Susko3:
This PR desperately needs tests, because currently `OrientationManager` isn't hooked up to anything and the Android one isn't either.

Also, why does this have to be a class? Potential consumers should only interact with `FrameworkSetting.Orientation`.

This should be a simple `.BindValueChanged()` on the setting here:

https://github.com/ppy/osu-framework/blob/1d5c093173a803673baef6ce227c91e2175fcaef/osu.Framework.Android/AndroidGameActivity.cs#L61-L73

(Create a new public bindable at `GameHost` and use that here? Similar to `AllowScreenSuspension`.)

osu.Framework/Platform/OrientationManager.cs
honguyenminh: Will change in next commits

bdach (changes requested):

osu.Framework.Android/AndroidOrientationManager.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <param name="orientation">Orientation</param>
+        /// <returns><see cref="AndroidPM.ScreenOrientation"/> enum to use with Android SDK</returns>
+        public virtual AndroidPM.ScreenOrientation SettingToNativeOrientation(ScreenOrientation orientation)
+        {
+            switch (orientation)
+            {
+                case ScreenOrientation.AnyLandscape:
+                    return AndroidPM.ScreenOrientation.UserLandscape;
+                case ScreenOrientation.AnyPortrait:
+                    return AndroidPM.ScreenOrientation.UserPortrait;
```
Susko3, line 44: I think having `SensorLandscape` for `AnyLandscape` and `SensorPortrait` for `AnyPortrait` is more useful to / expected from users. Rationale: https://github.com/ppy/osu-framework/pull/4897#discussion_r752579428

After all, `ScreenOrientation.Any` uses `FullSensor`, not `FullUser`.

osu.Framework.Android/AndroidOrientationManager.cs
```diff
@@ -0,0 +1,87 @@
...
+            {
+                gameActivity.RequestedOrientation = SettingToNativeOrientation(value.NewValue);
+            });
+        }
+
+        /// <summary>
+        /// Lock/unlock the orientation to current orientation
+        /// </summary>
+        /// <param name="isLocked"><see langword="true"/> to lock the orientation and <see langword="false"/> to unlock</param>
+        public void SetOrientationLock(bool isLocked)
```
Susko3, line 75: Should be moved to base class.

And probably move most of the `isLocked` logic there as well.

osu.Framework.Android/AndroidOrientationManager.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        /// <summary>
+        /// Convert <see cref="ScreenOrientation"/> framework setting enum to Android's <see cref="AndroidPM.ScreenOrientation"/> enum
+        /// </summary>
+        /// <remarks>
+        /// Override this to customize orientation flags
+        /// </remarks>
+        /// <param name="orientation">Orientation</param>
+        /// <returns><see cref="AndroidPM.ScreenOrientation"/> enum to use with Android SDK</returns>
+        public virtual AndroidPM.ScreenOrientation SettingToNativeOrientation(ScreenOrientation orientation)
```
Susko3, line 37: I don't see how this being public is useful.

> Override this to customize orientation flags

If consumers want to modify screen orientation, they should do it exclusively trough `FrameworkSetting.ScreenOrientation`, without having to worry about the native implementation.

osu.Framework.Tests.Android/Visual/Platform/TestSceneScreenOrientation.cs
```diff
@@ -0,0 +1,235 @@
...
+        private PM.ScreenOrientation originalOrientation;
+
+        [BackgroundDependencyLoader]
+        private void load(GameHost host)
+        {
+            window = host.Window;
+            if (window == null) return;
+
+            originalOrientation = gameActivity.RequestedOrientation;
+            manager = new AndroidOrientationManager(config, gameActivity);
```
Susko3, line 66: As seen in this test, consumers have to create an `OrientationManager` if they want to use `FrameworkSetting.ScreenOrientation`. This doesn't make any sense. Framework settings should be self-contained in the framework.

All other `FrameworkSetting`s are bound to natively in the framework, and those bindings automatically run the necessary native functions.

Susko3 (changes requested):

osu.Framework/Configuration/ScreenOrientation.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// </summary>
+        Auto = 0,
+        /// <summary>
+        /// Locked landscape, with top-down goes from left to right on the portrait screen
+        /// </summary>
+        LandscapeLeft = 1,
+        /// <summary>
+        /// Locked landscape, with top-down goes from right to left on the portrait screen
+        /// </summary>
+        LandscapeRight = 2,
```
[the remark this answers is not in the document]
honguyenminh: Fixed

osu.Framework/Configuration/ScreenOrientation.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// </summary>
+        Portrait = 4,
+        /// <summary>
+        /// Locked reverse portrait orientation
+        /// </summary>
+        ReversePortrait = 8,
+        /// <summary>
+        /// Landscape orientation, allows landscape screen rotation
+        /// </summary>
+        AnyLandscape = LandscapeLeft | LandscapeRight, // 3
```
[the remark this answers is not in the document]
honguyenminh: It's just old habit, removed in next commits

osu.Framework.Tests.Android/AndroidVisualTestGame.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using osu.Framework.Allocation;
+using osu.Framework.Android;
+
+namespace osu.Framework.Tests.Android
+{
+    internal class AndroidVisualTestGame : VisualTestGame
+    {
+        [Cached]
+        // Allows dependency injection
```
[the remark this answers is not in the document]
honguyenminh: Removed

osu.Framework.Android/AndroidOrientationManager.cs
```diff
@@ -0,0 +1,87 @@
...
+using osu.Framework.Platform;
+using osu.Framework.Configuration;
+using osu.Framework.Bindables;
+using Android.App;
+
+namespace osu.Framework.Android
+{
+    public class AndroidOrientationManager : ScreenOrientationManager
+    {
+        private bool isLocked = false;
```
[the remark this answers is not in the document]
honguyenminh: `Disabled` would disallow change to the setting while the game is being played (orientation locked) though. Wouldn't this break osu!'s hot setting change while playing feature?

osu.Framework.Tests.Android/Visual/Platform/TestSceneScreenOrientation.cs
```diff
@@ -0,0 +1,235 @@
...
+
+            foreach (var orientation in testOrientations)
+            {
+                AddStep("Change setting to " + orientation.ToString(), () =>
+                {
+                    orientationBindable.Value = orientation;
+                });
+                AddAssert("Test if current screen is " + orientation.ToString(), () =>
+                {
+                    var nativeOrientation = manager.SettingToNativeOrientation(orientation);
```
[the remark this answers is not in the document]
honguyenminh: Changed, i was weird that day

osu.Framework.Android/AndroidOrientationManager.cs
honguyenminh: `AnyLandscape` and `AnyPortrait` are better than the sensor ones, because they are basically Sensors but they respect user's auto-rotate setting. Sensor are annoying, because they bypass the setting altogether.

I chose `FullSensor` for `Any` because it's the only one that does not reset back to standing portrait when user lock auto-rotate. If `FullUser` doesn't revert back to Portrait on lock, I'll totally use it. But it's not, so I had no other choice.

osu.Framework.Android/AndroidOrientationManager.cs
honguyenminh: Was originally public to allow consumer to get the actual native flag of a setting, some app might need it, but yeah, if you deem it to be unnecessary I'll make it protected

osu.Framework.Android/AndroidOrientationManager.cs
honguyenminh: Moved

osu.Framework.Tests.Android/Visual/Platform/TestSceneScreenOrientation.cs
honguyenminh: Will implement the class in android game host instead.
Sorry, I'm not used to the framework codebase yet, so I just don't know how to actually use it in the framework.

Susko3: Could do something similar to this.  (Unless you found a better way.)

https://github.com/ppy/osu-framework/blob/16e9569a3fb613732accc1e701c9eac4d53cbb93/osu.Framework.Android/AndroidGameActivity.cs#L61-L73

osu.Framework.Android/AndroidOrientationManager.cs
Susko3: Do you need a native orientation in the test? The only way to confirm that the test passed is to look at the actual screen orientation of the actual device.

If the reported/actual screen orientation is required for something else, it should be reported as `Framework.Configuration.ScreenOrientation`. Consumer should never have to worry about converting native orientation.

osu.Framework.Android/AndroidOrientationManager.cs
Susko3: >Sensor are annoying, because they bypass the setting altogether.

That's why we differentiate `LandscapeLeft` and `LandscapeRight` from `AnyLandscape`. If the user wants to lock their orientation in a specific landscape, they can.

I can see `UserPortrait` being the expectation from users, but _every single landscape-supported app_ on my phone behaves like `SensorLandscape` when in landscape/fullscreen. By using `UserLandscape`, we are breaking user expectations.

osu.Framework.Android/AndroidOrientationManager.cs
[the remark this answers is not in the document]
Susko3: The orientation is unlocked when not playing / paused. So no problems hot swapping there.

osu.Framework.Android/AndroidOrientationManager.cs
honguyenminh: Nah, it's just simply needed to show the current setting in the text box, tests passing does not depend on this.

If removed, we would need another way to update that textbox when the native screen orientation flag changes.

osu.Framework.Tests.Android/Visual/Platform/TestSceneScreenOrientation.cs
honguyenminh: Now consumer only need to interact with the setting itself and use the lock bindable inside `GameHost` to lock the orientation.

osu.Framework.Android/AndroidOrientationManager.cs
Susko3: Do we need that textbox?

osu.Framework.Android/AndroidOrientationManager.cs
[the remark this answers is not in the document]
honguyenminh: I mean, we are locking the screen orientation, not the setting itself, while `Disabled` is used to disable the setting altogether. Currently this separate variable allows consumer to change the setting even while playing (locked) and new setting will take effect after unlocked. This allows more flexibility for consumer imo.

If you think this is overkill then I'll move to `Disabled` property instead

osu.Framework.Android/AndroidOrientationManager.cs
honguyenminh: I mean it does occasionally helps me while debugging by telling me why the test failed.

If there's a `RequestedOrientationChanged` event in the Xamarin framework, I would never make it public in the first place.

So, we remove the textbox, right? I'll try to find a way to keep it without the method public, delete otherwise.

osu.Framework.Android/AndroidOrientationManager.cs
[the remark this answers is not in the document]
Susko3: I think you should move to `Disabled`, as currently, changing the framework setting won't update orientation if it's locked (early return [here](https://github.com/ppy/osu-framework/blob/90b0a57c088e9efc55d1b8dd21474f20e5f991ad/osu.Framework.Android/AndroidOrientationManager.cs#L64)). The setting is logically "disabled", makes more sense if it's actually `Disabled` in that case.

osu.Framework.Android/AndroidOrientationManager.cs
Susko3: If it's only for local debugging, you should remove it / not include it in the PR.

The test should, ideally, be platform-agnostic since it's testing sometimes common to all platforms -- a `FrameworkSetting`.

The test won't be of much use on desktop (only test that it doesn't crash; changing orientation is a noop on desktop), but it should be useful for both Android and iOS (when implementation comes).

So no platform specific orientations. (If it's really required for testing/debugging -- exposed as a raw `string`).

honguyenminh: Is protected now. The event is triggered after the class has run, so textbox works fine without the method. I was based I guess.

osu.Framework.Android/AndroidOrientationManager.cs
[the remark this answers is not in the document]
honguyenminh: I'll leave this for user to discuss, tbh I still want to separate the locking mechanism (which change the actual screen orientation, not the setting) from disabling the setting at all.

osu.Framework.Tests.Android/Visual/Platform/TestSceneScreenOrientation.cs
honguyenminh: Can I do something like in latest commit instead?

honguyenminh: Kind of implemented it in the latest commit, can you review it?

honguyenminh: Can you review it again? I'm not sure about my approach on this.

osu.Framework.Android/AndroidOrientationManager.cs
Susko3: My concerns here haven't been addressed. This conversion shouldn't be marked as resolved.

osu.Framework.Android/AndroidOrientationManager.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        /// <summary>
+        /// Convert <see cref="ScreenOrientation"/> framework setting enum to Android's <see cref="AndroidPM.ScreenOrientation"/> enum
+        /// </summary>
+        /// <remarks>
+        /// Override this to customize orientation flags
+        /// </remarks>
+        /// <param name="orientation">Orientation</param>
+        /// <returns><see cref="AndroidPM.ScreenOrientation"/> enum to use with Android SDK</returns>
+        protected virtual AndroidPM.ScreenOrientation SettingToNativeOrientation(ScreenOrientation orientation)
```
Susko3, line 36: As a consumer, it's not possible to override this behaviour (can't override the orientation manager provided by `GameHost`).

And I don't see how overriding this behaviour is necessary.

osu.Framework/Platform/ScreenOrientationManager.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using osu.Framework.Bindables;
+using osu.Framework.Configuration;
+
+namespace osu.Framework.Platform
+{
+    public abstract class ScreenOrientationManager
```
Susko3, line 10: Should be `internal`, as it's managed entirely by framework/`GameHost`, and consumers should only interact with it indirectly trough the new setting and the `LockScreenOrientation` bindable.

Same goes for `AndroidOrientationManager`.

osu.Framework/Platform/GameHost.cs
```diff
@@ -931,6 +950,14 @@ protected virtual void SetupConfig(IDictionary<FrameworkSetting, object> default
                     windowMode.Value = Window.DefaultWindowMode;
             }, true);
 
+            ScreenOrientationBindable = Config.GetBindable<ScreenOrientation>(FrameworkSetting.ScreenOrientation);
+            screenOrientationManager = GetScreenOrientationManager();
```
Susko3, line 954: Shouldn't `ScreenOrientationBindable` here be a simple function parameter to `GetScreenOrientationManager()`?
Instead of being a field that's accessed in `AndroidGameHost` for the sole reason of using it in the overridden `GetScreenOrientationManager()`

osu.Framework.Android/AndroidOrientationManager.cs
honguyenminh: Fair enough.
Will change in next commit

osu.Framework/Platform/GameHost.cs
```diff
@@ -931,6 +950,14 @@ protected virtual void SetupConfig(IDictionary<FrameworkSetting, object> default
...
+
+            LockScreenOrientation.ValueChanged += value =>
+            {
+                screenOrientationManager?.SetOrientationLock(value.NewValue);
+            };
```
Susko3, line 959: Should probably be moved to `ScreenOrientationManager`. And then pass the bindable in the ctor.

osu.Framework/Platform/GameHost.cs
```diff
@@ -153,6 +164,14 @@ public abstract class GameHost : IIpcHost, IDisposable
 
         private ReadableKeyCombinationProvider readableKeyCombinationProvider;
 
+        protected Bindable<ScreenOrientation> ScreenOrientationBindable { get; private set; }
+
+        [CanBeNull]
+        protected virtual ScreenOrientationManager GetScreenOrientationManager() => null;
```
Susko3, line 170:
```suggestion
        protected virtual ScreenOrientationManager CreateScreenOrientationManager() => null;
```

Should be `Create...` for classes that are single-instanced. To match `CreateWindow()`, `CreateReadableKeyCombinationProvider()`, etc.

`GetClipboard()` is creating a new clipboard every time, so it's `Get...`.

osu.Framework.Android/AndroidOrientationManager.cs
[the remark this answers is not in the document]
honguyenminh: I'll keep the lock mechanism separate from `Disabled`, because using `Disabled` as lock control breaks consumer's expectation, and creates an edge case for this one setting, we would have to notice user in the docs about this too. Disabling a setting shouldn't create side effects like locking the orientation.

Susko3 (changes requested):
CI isn't picking this up, but there are multiple inspections in the Android side of things.

osu.Framework.Android/AndroidOrientationManager.cs
honguyenminh: Will be private in next commit

osu.Framework/Platform/ScreenOrientationManager.cs
honguyenminh: If `ScreenOrientationManager` is internal, we wouldn't be able to access (and derive from) it from `osu.Framework.Android` because it is in a different project (assembly) so I'll keep the base class public.
Android implementation can be internal though.

Susko3: `osu.Framework.Android` should probably have `osu.Framework` internals visible.

Add it here: [osu.Framework/Properties/AssemblyInfo.cs](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Properties/AssemblyInfo.cs)

honguyenminh: Changed in next commit

osu.Framework/Platform/GameHost.cs
honguyenminh: Changed in latest commit

osu.Framework.Android/AndroidOrientationManager.cs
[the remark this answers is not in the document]
Susko3: > breaks consumer's expectation

Don't think there is any consumer expectation about disabling settings/config bindables. Probably the only expectation is from the framework, and that is that those bindables aren't disabled by consumers. Maybe even don't do manual checks that the bindable is disabled, as it'll throw [here](https://github.com/ppy/osu-framework/blob/3a6512b948da679a6ded1b53ef27358d7eaf61e5/osu.Framework/Bindables/Bindable.cs#L90-L91) for you.

So agreed that a bindable at `GameHost` (like you have right now) is a better option.

osu.Framework/Configuration/ScreenOrientation.cs
```diff
@@ -0,0 +1,57 @@
...
+    [Flags]
+    public enum ScreenOrientation
+    {
+        /// <summary>
+        /// Let the device handle rotation
+        /// </summary>
+        Auto = 0,
+
+        /// <summary>
+        /// Locked landscape, with top-down goes from left to right on the portrait screen
```
bdach, line 23: I don't get this xmldoc. Shouldn't this just say "device rotated 90 degrees left from default portrait orientation" or similar?

osu.Framework/Platform/ScreenOrientationManager.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// </remarks>
+        protected abstract void OnScreenOrientationSettingChanged(ValueChangedEvent<ScreenOrientation> value);
+
+        /// <summary>
+        /// This event will be invoked when ScreenOrientation is locked
+        /// </summary>
+        /// <remarks>
+        /// Override this event to set native behavior
+        /// </remarks>
+        protected abstract void OnScreenOrientationLocked();
```
bdach, line 44: What's the point of these if the setting bindables are there already? Surely you could just `BindValueChanged()` onto the bindables rather than use this?

osu.Framework/Platform/ScreenOrientationManager.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        /// <summary>
+        /// Lock/unlock the orientation to current orientation
+        /// </summary>
+        /// <param name="isLocked"><see langword="true"/> to lock the orientation and <see langword="false"/> to unlock</param>
+        /// <exception cref="InvalidOperationException">Thrown if setting is disabled</exception>
+        public void SetOrientationLock(bool isLocked)
+        {
+            if (orientationBindable.Disabled)
+                throw new InvalidOperationException("Can't change screen orientation lock when setting is disabled");
```
bdach, line 54: This makes no sense to me. What does this mean? How can this happen?

bdach (changes requested):
Still of the opinion that the "orientation manager" has no reason to exist at all.

osu.Framework/Platform/ScreenOrientationManager.cs
honguyenminh: Done. I totally overthink this.

osu.Framework/Platform/ScreenOrientationManager.cs
honguyenminh: This will throw if consumer disabled the orientation config, but still try to change the lock state. Or am I getting something wrong?

osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -90,6 +93,32 @@ protected override void OnCreate(Bundle savedInstanceState)
...
+                        throw new InvalidOperationException("Can't change screen orientation lock when setting is disabled");
+
+                    if (e.NewValue) // If locked
+                    {
+                        RunOnUiThread(() =>
+                        {
+                            RequestedOrientation = NativeOrientation.Locked;
+                        });
+                    }
+                    else host.ScreenOrientation.TriggerChange();
```
peppy, line 120: This is really bad. Split out the required logic to its own method rather than calling `TriggerChange`.

Or even better, simplify this whole block to:

```suggestion
  RunOnUiThread(() =>
                         {
                             RequestedOrientation = e.NewValue ? NativeOrientation.Locked : configToNativeOrientationEnum(host.ScreenOrientation.Value);
                         });
                    else host.ScreenOrientation.TriggerChange();
```

osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -90,6 +93,32 @@ protected override void OnCreate(Bundle savedInstanceState)
...
+
+                    RunOnUiThread(() =>
+                    {
+                        RequestedOrientation = configToNativeOrientationEnum(e.NewValue);
+                    });
+                });
+                host.LockScreenOrientation.BindValueChanged(e =>
+                {
+                    // Throw if consumer change locked state while ScreenOrientation bindable is disabled
+                    if (host.ScreenOrientation.Disabled)
```
peppy, line 110: Why isn't this just a direct bind to avoid this nonsense?

osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -90,6 +93,32 @@ protected override void OnCreate(Bundle savedInstanceState)
                             Window.ClearFlags(WindowManagerFlags.KeepScreenOn);
                     });
                 }, true);
+
+                host.ScreenOrientation.BindValueChanged(e =>
```
peppy, line 97: Doesn't this need a `true` argument to ensure it's run once?

osu.Framework/Platform/GameHost.cs
```diff
@@ -70,6 +70,24 @@ public abstract class GameHost : IIpcHost, IDisposable
...
+        /// </summary>
+        /// <remarks>
+        /// If <see cref="FrameworkSetting.ScreenOrientation"/> is changed while this is locked,
+        /// nothing will take effect immediately, however the setting will still be acknowledged,
+        /// and will take effect when this is unlocked again.
+        /// </remarks>
+        /// <exception cref="InvalidOperationException">
+        /// Throw if value is changed while <see cref="FrameworkSetting.ScreenOrientation"/> setting is disabled
+        /// </exception>
+        public readonly Bindable<bool> LockScreenOrientation = new Bindable<bool>();
```
peppy, line 89: I'm not sure how well these settings will play with what iOS exposes as configurable in terms of locking/rotating. That should be visited before merging this.

peppy (changes requested):

osu.Framework.Android/AndroidGameWindow.cs
honguyenminh: This change has been manually reverted at 712c8668112ddb637b2890c38b24066ec2d53279

osu.Framework/Platform/GameHost.cs
honguyenminh: From what I can find, setting iOS orientation will lock the orientation to 1 of the 4 possible orientations, so locking is possible, but I'm not sure if we can do something similar to "SensorLandscape" of Android (allow both rotation).
I found this repo that can be helpful for implementing this [here](https://github.com/wcoder/Xamarin.Plugin.DeviceOrientation/blob/master/src/DeviceOrientation/Plugin.DeviceOrientation.iOS/DeviceOrientationImplementation.cs)

I don't own an iOS device so I'm not sure how iOS users would expect the orientation to behave, so some elaboration from an actual user would be helpful.

frenzibyte: I believe it is possible to have locking and `SensorLandscape`-like support for iOS. Will look into writing a simple implementation for it today.

osu.Framework.Android/AndroidGameActivity.cs
honguyenminh: You mean directly binding it to Disabled of the screen orientation setting?

IMO doing that makes it hard for consumer to know about this behavior, since `Disabled` doesn't imply Locking the setting too. But if you think that's better, I'll move this there and remove the `LockScreenOrientation` bindable.

osu.Framework.Android/AndroidGameActivity.cs
honguyenminh: Changed in next commits.

osu.Framework/Platform/GameHost.cs
frenzibyte: Sorry for the delay, I've got an implementation ready for this now (95deedfac4fd83e2661c69d638946734dfa72988), and have got a video showcasing behaviour for iOS:

https://user-images.githubusercontent.com/22781491/150105768-6e47be45-e60a-43c7-b88c-8f2660d9013d.mp4

(also tested locally on my iPhone)

Implementation was indeed mostly straightforward, as I only had to update the `UIDevice` when user changes orientation and block autorotation when new one doesn't respect what the user has set (or if `LockScreenOrientation` is set).

Few notes while implementing and testing on iOS:
 1. iPhones without home buttons do not support the `ReversePortrait` orientation as mentioned in [apple docs](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621435-supportedinterfaceorientations), therefore it should be completely hidden in game settings:
    CleanShot 2022-01-19 at 12 52 13
 2. Applications in iOS cannot fully bypass the phone's "Portrait Orientation Lock" (unlike Android, which does with its `sensorLandscape`/`fullSensor` orientations), but setting the orientation to `LandscapeLeft`/`LandscapeRight` will still work. I think the best choice would be to hide the "Auto" orientation in game settings and include a note when using `AnyLandscape`/`Any` that it won't autorotate if the user has enabled portrait lock.

osu.Framework.Android/AndroidGameActivity.cs
peppy: @honguyenminh are you going to test/apply this change? probably required to move forward with this PR.

osu.Framework/Platform/GameHost.cs
peppy: Well, more importantly, I think you'll find *none* of these options should be exposed to the user on iOS.

osu.Framework.Android/AndroidGameActivity.cs
peppy: I think I misunderstood this, seems alright how you have it. Probably doesn't need the comment above as it's explaining exactly what the code is going (same goes for the comment on line 99).

osu.Framework.Android/AndroidGameActivity.cs
honguyenminh: Will change in next commit.
Sorry lunar new year's been dragging me back for a while.

osu.Framework.Android/AndroidGameActivity.cs
honguyenminh: Alrighty.

osu.Framework/Platform/GameHost.cs
frenzibyte: I don't follow what you mean? are you saying that iOS applications should usually not have a setting for changing orientation programmatically?

peppy: Correct. I don't think I've ever seen it in an app, which means it's likely against the UX guidelines.

honguyenminh: So this should be an Android-only thing?

I can try to move this to Android host specifically. Can't guarantee if I can actually do it without a mess of code though.

peppy: No it's in the correct place. It just should not be exposed to the user.

honguyenminh: I do find a few interesting thing on android side too. The Any<Portrait/Landscape> settings will not actually rotate to the reverse orientation from the home button or the "main" orientation of current device.
For example: a tablet with default orientation of landscape 90deg *clockwise* will **NOT** auto rotate to landscape 90deg *counter-clockwise* on setting's `AnyLandscape` aka native's `SensorLandscape` or `UserLandscape`. Or for a normal portrait phone (the one I'm using to test this for example), will not rotate to reverse portrait on `AnyPortrait` or <Sensor/User>Portrait. Seems like what orientation will be blocked by this is defined by OEM, but I can't find docs about this yet.
I can't seem to find a way to bypass this from code though. Full<User/Sensor> will bypass this fine, but these allow all orientation so that only makes `Auto` and `Any` work.

osu.Framework.Tests.Android/Visual/Platform/TestSceneScreenOrientation.cs
```diff
@@ -0,0 +1,223 @@
...
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Sprites;
+using osu.Framework.Graphics.UserInterface;
+using osu.Framework.Platform;
+using osu.Framework.Testing;
+using osuTK;
+using PM = Android.Content.PM;
+
+namespace osu.Framework.Tests.Android.Visual.Platform
```
Susko3, line 19: Screenshot_2022-01-24-13-02-39-11_fca46a2731136e673b8638a972bc0192(1)

These Android-only tests do make the test browser a bit weird.

These tests also aren't really testing the behaviour (the actual rotation of the device), they are just testing whether the correct orientation flags are set. The usefulness of this is dubious for me. Only useful thing that it's doing is validating locking behavior.

peppy:
I've tidied this up, but in the process I'm still not making sense of how this is to be used by a consumer.

I don't think it should be in the config full stop. Android and iOS has rotation lock, so that doesn't need to be locked here. If anything, it should either be:

- A startup setting set by the game
- Something the game can change at runtime, but not the user (ie. it shouldn't be in config)

Also I don't think there needs to be a separate "locked" property. That's android specific, and can probably be merged into the main enum and throw if attempted to be used on a platform that isn't android.

Susko3:
The original issue, for reference: https://github.com/ppy/osu/issues/12764 (not sure how this was lost in the PR).

My idea for the framework-side setting is to make it easier for consumers to change the setting without having to worry about running on UI thread, resolving the Activity, etc. And also to allow users to change the setting how they see fit. (Someone might want to play osu! in portrait, you never know.)

Exposing screen orientation to the user is common UX on Android.

>A startup setting set by the game

This is already how Android works, see `ScreenOrientation = ScreenOrientation.FullUser` [here (`OsuGameActivity`)](https://github.com/ppy/osu/blob/2c0c44a9504c831686d49f52b574051ec03cc092/osu.Android/OsuGameActivity.cs#L20).

We should probably change the default to be landscape.

osu.Framework.Tests.Android/Visual/Platform/TestSceneScreenOrientation.cs
peppy: I'd argue that this test should be in the standard framework tests? Not sure why it needs to be in the android project.

osu.Framework/Configuration/FrameworkConfigManager.cs
```diff
@@ -37,6 +37,7 @@ protected override void InitialiseDefaults()
             SetDefault(FrameworkSetting.SizeFullscreen, new Size(9999, 9999), new Size(320, 240));
             SetDefault(FrameworkSetting.FrameSync, FrameSync.Limit2x);
             SetDefault(FrameworkSetting.WindowMode, WindowMode.Windowed);
+            SetDefault(FrameworkSetting.ScreenOrientation, ScreenOrientation.AnyLandscape);
```
peppy, line 40: This is changing the default for *all consumers*. This should not be the case.

peppy (changes requested):

peppy:
I can say that this will not be merged unless it is removed from the config. Keep it as a game host setting and document it with a note that it won't apply to anything but android.

Or just don't do any of this and leave it up to the consumer to implement android specific behaviour if they want it.

peppy:
Let's keep this simple and close this PR for now.

@honguyenminh sorry to do this after the work you've put in, but hopefully you can understand why. I am willing to bring this back if there is demand to have the setting in other consumer projects (if this was implemented for your own use in another project please say so - to my knowledge it was added specifically for osu!, where we have decided we likely won't need the functionality).

To summarise:

- On android users (sometimes) expect apps to allow them to set orientation settings within the app.
- On iOS, such settings are not ever present.
- On desktop platforms this is irrelevant.

Because we are targetting multiple platforms, but focusing on providing consolidated/consistent behaviour across them all (without platform-specific features wherever possible) this change doesn't align with our goals.

The recommended method is to set startup values via `Info.plist` (iOS) or `ActivityAttribute` (Android). If you decide that you want to expose settings to the user (I'd argue is a pretty weird/unexpected UX but I am an apple user where UX means everything so YMMV), it should be easy enough to copy the implementation in this pull request into your own project / `ConfigManager` and get the required behaviour.

PR #4898, opened 2021-11-17
Change from UTC time to local time

JeffersonQin:
UTC time in logging file may cause misinterpretation for users.

### DIFF
--- a/osu.Framework/Logging/Logger.cs
+++ b/osu.Framework/Logging/Logger.cs
@@ -305,7 +305,7 @@ namespace osu.Framework.Logging
             IEnumerable<string> lines = logOutput
                                         .Replace(@"\r\n", @"\n")
                                         .Split('\n')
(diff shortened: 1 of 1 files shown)


frenzibyte:
And changing it from UTC time to local time makes it hard to tell at what point something happened specifically, when logs are shared between developers across other timezones.

If anything, both local and UTC should be displayed, with indicators for that.

peppy:
Logs aren't made for users. Can you explain the use case here better? ie. why would a user be looking in the first place?

peppy:
Please open an issue if you think there's a valid use case. I don't think there is one. UTC is standard for log files.

JeffersonQin:
okay, got that.



---Original---
From: "Dean ***@***.***&gt;
Date: Thu, Nov 18, 2021 00:39 AM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [ppy/osu-framework] Change from UTC time to local time (PR #4898)




 
Please open an issue if you think there's a valid use case. I don't think there is one. UTC is standard for log files.
 
—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

PR #4900, opened 2021-11-18
Update all `OnPressed` methods to block repeats where applicable

peppy:


osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -164,36 +164,42 @@ public virtual bool OnPressed(KeyBindingPressEvent<PlatformAction> e)
...
+                        if (e.Action == PlatformAction.Cut)
+                            DeleteBy(0);
 
-                    return true;
+                        return true;
 
-                case PlatformAction.Paste:
-                    //the text may get pasted into the hidden textbox, so we don't need any direct clipboard interaction here.
-                    string pending = textInput?.GetPendingText();
+                    case PlatformAction.Paste:
```
smoogipoo, line 183: This one can should remain repeatable.

smoogipoo (changes requested):

osu.Framework/Graphics/UserInterface/TextBox.cs
peppy: I guess from that angle they all can really? And just leave the noop handling to the operation itself? Not sure.

smoogipoo: That's true, there's no bad side effects from them all being repeatable as far as I can see.

smoogipoo (approved):

PR #4905, opened 2021-11-19
Improve readability and handling of fill directions in `FillFlowContainer`

Flutterish:
This PR introduces a concept used by flexboxes (and probably some other UI design components) called main and cross axes. This is basically a `Vector2` except rather than having `X` and `Y`, it has `Main` and `Cross` which can be interpreted depending on orientation. For example for `FillDirection.{Full, Horizontal}` the main axis is `Axes.X` and the cross axis is `Axes.Y`, and the opposite is true for the other `FillDirection`s.

This allows us to ignore the actual direction we are using (only converting at the beginning and end) and operate as if we were handling just `Full` or `Horizontal` with `Main` and `Cross` rather than `X` and `Y` respectively. Due to this this PR also introduces a new `FillDirection` - `FullVertical`.

This PR makes some cases, especially new row handling, easier to read. Speaking of rows, (namingwise) I replaced them with "spans" as they, just like cross axes, lack orientation. `FillFlowContainer`s have a `ForceNewRow` method, which I believe originally forced a row, even for the `Vertical` orientation, which did effectively nothing. Now, `ForceNewRow` was made obsolete and replaced with `ForceNewSpan` which would instead create a new column in such case. This is only a rename, as FFCs now use spans rather than rows and columns. **[request: add when the obsolete member can be removed]**

You might notice that `ForceNewSpan` creates a new span even if the direction should not create new spans (the not "Full" fill directions). This is how it worked originally too (for horizontal orientations, since it did nothing for vertical), but I think it should be discussed whether this is correct behaviour.

This PR also refactored child validation into its own method, rather than `if`s scattered across the method.

### DIFF
--- a/osu.Framework/Graphics/Containers/FillFlowContainer.cs
+++ b/osu.Framework/Graphics/Containers/FillFlowContainer.cs
@@ -86,28 +86,30 @@ namespace osu.Framework.Graphics.Containers
             }
         }
 
-        private Vector2 spacingFactor(Drawable c)
+        private CrossVector spacingFactor(Drawable c)
         {
             Vector2 result = c.RelativeOriginPosition;
             if (c.Anchor.HasFlagFast(Anchor.x2))
                 result.X = 1 - result.X;
             if (c.Anchor.HasFlagFast(Anchor.y2))
                 result.Y = 1 - result.Y;
-            return result;
+            return ToCross(result);
         }
 
         protected override IEnumerable<Vector2> ComputeLayoutPositions()
         {
-            var max = MaximumSize;
+            var max = ToCross(MaximumSize);
 
-            if (max == Vector2.Zero)
+            if (MaximumSize == Vector2.Zero)
             {
                 var s = ChildSize;
 
                 // If we are autosize and haven't specified a maximum size, we should allow infinite expansion.
                 // If we are inheriting then we need to use the parent size (our ActualSize).
-                max.X = AutoSizeAxes.HasFlagFast(Axes.X) ? float.MaxValue : s.X;
-                max.Y = AutoSizeAxes.HasFlagFast(Axes.Y) ? float.MaxValue : s.Y;
+                max = ToCross(new Vector2(
+                    AutoSizeAxes.HasFlagFast(Axes.X) ? float.MaxValue : s.X,
+                    AutoSizeAxes.HasFlagFast(Axes.Y) ? float.MaxValue : s.Y
+                ));
             }
 
             var children = FlowingChildren.ToArray();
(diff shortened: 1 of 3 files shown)


Flutterish:
Idea: we could use "line" instead of "span". I think that would be more straightforward as it also implies that its a 1-dimensional line rather than a span which could imply more dimensions, even though in this context there is no space for such conflict.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
```diff
@@ -320,6 +401,42 @@ public enum FillDirection
         /// <summary>
         /// Fill only vertically.
         /// </summary>
-        Vertical
+        Vertical,
+
+        /// <summary>
+        /// Fill vertically first, then fill horizontally via multiple columns.
+        /// </summary>
+        FullVertical,
```
bdach, line 409: I believe I've already mentioned this the first time around when this was being reviewed but I do not really like the asymmetry of `Full` and `FullVertical`. I think `Full` should be renamed to `FullHorizontal`, and `Full` should be a deprecated transitional alias for `FullHorizontal` for 6 months before eventual removal.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
```diff
@@ -299,7 +289,98 @@ static Axes toAxes(FillDirection direction)
         /// </summary>
         /// <param name="child">The child to check.</param>
         /// <returns>True if the given child should be placed on a new row, false otherwise.</returns>
+        [Obsolete("Use ForceNewSpan instead")]
```
bdach, line 292: Should have an inline comment indicating date of removal (which would be 6 months in the future).

osu.Framework/Graphics/Containers/FillFlowContainer.cs
```diff
@@ -299,7 +289,98 @@ static Axes toAxes(FillDirection direction)
...
         protected virtual bool ForceNewRow(Drawable child) => false;
+
+        /// <summary>
+        /// Returns true if the given child should be placed on a new span, false otherwise. This will be called automatically for each child in this FillFlowContainers FlowingChildren-List.
+        /// A span can refer to either row or column, depending on <see cref="Direction"/>.
+        /// </summary>
+        /// <param name="child">The child to check.</param>
+        /// <returns>True if the given child should be placed on a new span, false otherwise.</returns>
+#pragma warning disable CS0618 // Type or member is obsolete
+        protected virtual bool ForceNewSpan(Drawable child) => ForceNewRow(child);
```
bdach, line 302: Would tend to agree with calling `ForceNewLine` as proposed.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
```diff
@@ -86,211 +86,201 @@ public Vector2 Spacing
...
-                        {
-                            case FillDirection.Full:
-                                return Axes.Both;
-
-                            case FillDirection.Horizontal:
-                                return Axes.X;
-
-                            case FillDirection.Vertical:
-                                return Axes.Y;
+                    var spacingFactor = this.spacingFactor(c);
```
bdach, line 145: Can the private method be renamed to `calculateSpacingFactor()` or something to avoid the `this` reference?

osu.Framework/Graphics/Containers/FillFlowContainer.cs
```diff
@@ -86,211 +86,201 @@ public Vector2 Spacing
...
                         // on both sides of the step to be taken.
-                        stride = (Vector2.One - spacingFactor(c)) * size;
+                        stride = (CrossVector.One - spacingFactor) * size;
 
                         c = children[i + 1];
-                        size = c.BoundingBox.Size;
+                        size = ToCross(c.BoundingBox.Size);
 
-                        stride += spacingFactor(c) * size;
+                        stride += this.spacingFactor(c) * size;
```
bdach, line 182: Why is the `spacingFactor` variable not used here?

bdach (changes requested):

bdach:
In general I disagree that this improves code quality or readability. It is a new useful feature, but abstracting away concrete axes makes this code less understandable, not more.

As a point of fact, I find the `Main`/`Cross` distinction confusing. Especially so that I gather that the content primarily flows in the "cross" direction, until there is no more space, and only then the "main" direction is used. So in a sense the "main" direction of flow is... not the primary one?

I would suggest `FlowingAxis` and `LineAxis` or something similar as an alternative. That makes the distinction of which axis is used to flow content and which one is used to break content into lines clearer.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
Flutterish: Ah, yes, great idea. I did not rename it because I knew that name was already used and didnt know how to go about changing it.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
Flutterish: I requested a date in the topmost comment since the wiki says the date is 3-6 months and because of this I didnt know what date to set. I'll set it to 6 months then

Flutterish:
> As a point of opinion, I find the `Main`/`Cross` distinction confusing. Especially so that I gather that the content primarily flows in the "cross" direction, until there is no more space, and only then the "main" direction is used. So in a sense the "main" direction of flow is... not the primary one?

The "Main" axis is used first, then new rows/columns are added along the "Cross" axis.

Flutterish:
What names would you suggest then? Perhaps `Primary` and `Secondary`?

Flutterish:
I think the name "Main" should stay, but changing the "Cross" to something else might indeed be useful. "Line" something, "Side" or "Orth"/"Perp" could do?

bdach:
The reason why I suggested what I did is that I would like to be able to tell at a glance which is the flowing direction and which is the line-breaking direction. For instance, the reason why I was confused with the current setup is that the outermost-scoped code starts off using `Cross` and then only uses `Main` in a super-nested block.

Those are the defining features of those axes: across one content flows, and across one content is broken up into lines. I feel that names like "cross", "main", "side", "orth" are all too generic to convey that clearly.

Flutterish:
Would `FlowVector` with `Main` or `Flow` as primary and `Line` as orthogonal do? I dont think they should have the word "Axis" in them. I would then rename the `ToCross` to `ToFlowVector`, I think that would be clear.

bdach:
I could work with that, yeah.

Flutterish:
Ah, one more thing. I am working on a container that implements the full flexbox spec and the concept of a flow vector is used there too. I was thinking I should perhaps move it out to its own file somewhere (since right now its a protected nested type) so it could use the same type?

bdach:
One type per file is the preferred style anyhow.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
Flutterish: I thought this was weird too, and even commited a change to use the variable, but now I noticed that the test with random origins broke. This should not use the variable because its a spacing factor of a different child, assigned a few line before.

Flutterish:
There is one more thing that I could address in this PR - it is currently impossible to both reverse and center a flow at the same time. When children are centered they prefer left-right and top-down.
I am not sure how to call the member that would indicate which direction if preferred, but I think it could be an enum that could either reverse the flow in all cases, or define a preferred direction for the center case alone, since the non-center cases depend on child anchors moving them to the appropriate side of the container.

bdach:
That sounds like a new feature and as such a separate pull. I would ask to keep pulls as small and self-contained as possible.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
```diff
@@ -86,211 +86,194 @@ public Vector2 Spacing
...
 
-                Vector2 ourRelativeAnchor = children[0].RelativeAnchorPosition;
-
-                // Second pass, adjusting the positions for anchors of children.
-                // Uses rowWidths and height for centre-anchors.
-                for (int i = 0; i < children.Length; i++)
+                void finalizeLineFlow(int from, int to)
                 {
-                    var c = children[i];
+                    float centreOffset = ToFlowVector(children[from].BoundingBox.Size).Flow * calculateSpacingFactor(children[from]).Flow;
```
bdach, line 195: why is it only sufficient to check the first item from the line? this logic wasn't there previously and is not explained.

in fact i think it may be causing this regression of centering that i noticed with this pull:

2021-11-21-141327_2560x1014_scrot

this does not happen on `master`.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
```diff
@@ -86,211 +86,194 @@ public Vector2 Spacing
...
 
-                    stride += Spacing;
+                        stride += spacingFactor * size;
 
-                    if (stride.Y > rowHeight)
-                        rowHeight = stride.Y;
-                    current.X += stride.X;
+                        if (stride.Line > lineLineSize)
+                            lineLineSize = stride.Line;
+                        current.Flow += stride.Flow + ToFlowVector(Spacing).Flow;
```
bdach, line 187: why is spacing not added to the last item? this didn't use to be the case.

https://github.com/ppy/osu-framework/blob/d7bb25be651d040054c533b63a35f78176e4081a/osu.Framework/Graphics/Containers/FillFlowContainer.cs#L210-L226

bdach (changes requested):
this diff is incredibly difficult to review due to unexplained refactorings being applied on top of already complicated logic. i need clarification on some changes made here i don't understand before proceeding.

also there's at least one regression related to centering.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
Flutterish: All children are required to have the same flow axis anchor for single fill directions and all anchors to be the same for full fill directions. I think this happened due to the the stride not being added to the last item?

osu.Framework/Graphics/Containers/FillFlowContainer.cs
```diff
@@ -86,211 +86,194 @@ public Vector2 Spacing
...
-            var current = Vector2.Zero;
+            var current = FlowVector.Zero;
+            var size = ToFlowVector(children[0].BoundingBox.Size);
+            var spacingFactor = calculateSpacingFactor(children[0]);
+            var ourRelativeAnchor = children[0].RelativeAnchorPosition;
+            var (reverseFlow, centerFlow) = calculateFlowAnchor(children[0]);
 
-            // First pass, computing initial flow positions
-            Vector2 size = Vector2.Zero;
+            float lineLineSize = 0;
```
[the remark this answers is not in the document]
Flutterish: The current lines line size. I guess I can change `lineSize` to `totalLineSize`, but I'm not sure what to do with this one. I could keep it as `lineCrossSize` if that's more readable?

osu.Framework/Graphics/Containers/FillFlowContainer.cs
Flutterish: This was the cause of the centering regression. I failed to notice that `current` was used as a measure for the lines flow size in the spacing function.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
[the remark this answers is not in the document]
bdach: I don't really understand still, the "line's line size" part is breaking my brain. Is it the "length" of the current line being laid out in the flowing direction? Or is it the "breadth" of the line in the line-breaking direction?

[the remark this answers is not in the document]
Flutterish: Line size as in `FlowVector.Line`. Its the height of a row / width of a column. Kind of an unfortunate that we lay out lines along the line axis and also need its line axis size.

Flutterish:
Due to the confusion caused by a line being able to have a "line size" due to how the flow vector names are, I propose this naming:
Frame 1 (1)
The change only for the names of sizes on a given axis. The axes are still "flow" and "line", spans placed on the line axis are still "lines", but the line axis size is now a "cross size" and the flow axis size is now a "main size". I would also be fine with the main size being called a "flow size" but I think that might be confused with it being the 2d size.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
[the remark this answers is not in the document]
Flutterish: with the new naming (a size on the line axis is now called a cross size rather than a line size) this should not be an issue anymore

osu.Framework/Graphics/Containers/FillFlowContainer.cs
Flutterish: I had to reverse engineer what the original `rowBeginOffset` was doing (since this is just a rename of it), and I added a comment explaining why it is needed.
What it was is (lets assume rows so its easier to talk about than the direction agnostic version), when centering a row, children would, because of their centre anchor, laid out so that the first child at (0,0) would be perfectly centered and the next ones would be to the right of it. Then, to properly center them all we would need to first align the left edge of the first drawable (offsetting all children) to the center of a row, and then move all children by half the row width to the left.
The reason this centering bug happened is because I used `current` as a measure of a rows width, but because I did not add the stride of the last child to it, the last row would think it has a different width than it actually has.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
```diff
@@ -86,211 +86,204 @@ public Vector2 Spacing
...
-            var current = Vector2.Zero;
+            var current = FlowVector.Zero;
+            var size = ToFlowVector(children[0].BoundingBox.Size);
+            var spacingFactor = calculateSpacingFactor(children[0]);
+            var ourRelativeAnchor = children[0].RelativeAnchorPosition;
+            var (reverseFlow, centerFlow) = calculateFlowAnchor(children[0]);
 
-            // First pass, computing initial flow positions
-            Vector2 size = Vector2.Zero;
+            float lineCrossSize = 0;
```
bdach, line 147: initially there were "cross" and "main" sizes. then there were "flow" and "line" sizes. now there are... both? how does this simplify anything?

here's what i propose:

* rename `lineCrossSize` to `currentLineSize`, because that's what it is and it doesn't introduce any new terms.
* rename `totalCrossSize` to `lastLinePosition`, because that's what it is and it doesn't introduce any new terms.

osu.Framework/Graphics/Containers/FillFlowContainer.cs
```diff
@@ -86,211 +86,204 @@ public Vector2 Spacing
...
-                // Second pass, adjusting the positions for anchors of children.
-                // Uses rowWidths and height for centre-anchors.
-                for (int i = 0; i < children.Length; i++)
+                void finalizeLineFlow(int from, int to)
                 {
-                    var c = children[i];
+                    // we are centering with the help of the centre anchor, which sets the first child
+                    // to our middle and the next ones after it, so we have to align their edge to our center
+                    // and then move them towards the center by half of the lines main size.
+                    // this calculates the offset needed to align the edge
```
bdach, line 200: i don't understand a word of this, what with the pronoun game. who's "we"? who's "they"?

osu.Framework/Graphics/Containers/FillFlowContainer.cs
Flutterish: Flow Axis/Line Axis and Main Size/Cross Size is pretty much the same as X Axis/Y Axis and Width/Height. I dont like how there is the word "size" in main and cross because width/height do not have it, but what can you do. Main Size/Cross Size refer exclusively to size, while Flow/Line to position or axis. While initially I wanted to use Flow Size/Line Size which would reflect X Size/Y Size it conflicted with there being a concept such as a line, which you saw in action with the "lines line size". This is why I decided it might be better to have a separate word for size in a given axis like Width/Height.

> rename `lineCrossSize` to `currentLineSize`, because that's what it is and it doesn't introduce any new terms.

The problem with naming it `currentLineSize` is that it strips direction information. If I saw that identifier, I would think it refers to the main size of the current line, while it actually refers to the cross size.

> rename `totalCrossSize` to `lastLinePosition`, because that's what it is and it doesn't introduce any new terms

This one on the other hand confuses size and position. In the code, for centering I want to offset the children by half our cross size (minus the last lines cross size because of anchor reasons), not by the position of some line and while it is true that these 2 have the same numerical value, it doesn't change that the intent should be shown with the variable naming. (mathematically speaking, adding 2 positions (points) is undefined. you would need to use at least 1 vector, which represents a change in position (a size is a vector) for it to make sense)

osu.Framework/Graphics/Containers/FillFlowContainer.cs
Flutterish: This was my best shot at describing it with words. Maybe this visualization will make it more clear so we could come up with a more comprehensive comment?

1. This is how they are laid out by default with a centre anchor. The centreOffset is the size of the first child that extends to the left of (0,0).
image

2. We offset children by it, then take the main size of the line...
image

3. And offset them by -1/2 of it.
image

Issue #2945, opened 2019-11-03
Getting wrong input (probably encoding) from IME


huoyaoyuan:
Please test this carefully when #2525 is open. Always use mouse to toggle IME.

My system information:
Windows 10 Pro 18362.449
System display language: Simlified Chinese
Installed language: Simlified Chinese(PRC), English(US), Japanese
System codepage: 65001 (type `chcp` in cmd to confirm)

#### Simplified Chinese:
System draws an IME window (in poor quality)
![image](https://user-images.githubusercontent.com/5644458/68089064-2cb82980-fea0-11e9-8816-43108ccffeec.png)
![image](https://user-images.githubusercontent.com/5644458/68089084-53766000-fea0-11e9-82c6-135f6ea07c5a.png)
#### Japanese:
No IME window drawn. After typing something and press 1, something get selected and inputed into window, but obvioursly what I typed.
![image](https://user-images.githubusercontent.com/5644458/68089120-b4059d00-fea0-11e9-9715-5d8b1baf3812.png)

Visible characters in both scenario are korean characters that looks uncommon.

smoogipoo:
Would be good if we can avoid drawing an IME window altogether, and completely handle it at osu!framework side. That said, IME is a whole mess which is not likely to be implemented before we move to Veldrid/SDL.

huoyaoyuan:
After investigation, `osuTK.KeyPressEventArgs.KeyChar` returns the erroneous value. Nothing to do now since we are going to remove it.

huoyaoyuan:
Tested with #3013, SDL does not have this issue. But is doesn't draw Chinese IME, like Japanese.

Susko3:
@huoyaoyuan could you test with https://github.com/ppy/osu-framework/pull/4907.
Framework doesn't include CJK fonts, so you may need to copy over from osu-resources, or build osu! with framework changes.

PR #4907, opened 2021-11-20
Add SDL2 IME support

Susko3:
Closes https://github.com/ppy/osu-framework/issues/2945.

Improves the current `TextBox` IME composition logic, and hooks up the necessary SDL2 events.
A full rewrite was required for `onImeComposition()`.

The SDL IME implementation has some problems (most importantly, limited string length) and I see it as a temporary and quick solution to get IME working (to some degree). It's unfortunate that this is the case, there is already [an existing issue](https://github.com/libsdl-org/SDL/issues/1879) and [a proposed fix](https://github.com/libsdl-org/SDL/pull/4799), but it doesn't include Windows fixes.

IME composition is finalized when losing focus, and when clicking inside the text box (eg. to move the cursor/selection). This is how web browsers and word processors behave on my machine, so I mimicked that in TextBox.

**Short demo:**

https://user-images.githubusercontent.com/16479013/142743301-3a8a15db-7a30-4e5f-ad86-09cb518f695f.mp4

Prerequisites
---

- [x] https://github.com/ppy/SDL2-CS/pull/85
- [x] include the following PR in our SDL2-CS release: https://github.com/libsdl-org/SDL/pull/4931 
  - cherry-picking is fine if you don't want to introduce (potential) SDL2 bugs from upstream
- [x] bump SDL2-CS
- [ ] find a way to allow the user to interact with the native IME window when relative mouse mode is on (not really a blocker, but a nice to have)
    - waiting on https://github.com/libsdl-org/SDL/issues/4339#issuecomment-968080201
- [ ] ~~add CJK font support to framework (opportunity to update to latest version of Noto if we're out of date?)~~
    - don't add more fonts as to keep the framework `.dll` as thin as possible (with Noto: 3.18 MB &rarr; 19.4 MB)
    - when testing, I just copied over Noto from [osu-resources](https://github.com/ppy/osu-resources/tree/master/osu.Game.Resources/Fonts/Noto) and loaded it in

Verify IME behaviour
---

- [x] Windows 10
  - [x] IME works
  - [x] IME window is properly positioned (even with screen scaling > 100%)
- [x] Windows 11 ([reported as not properly working](https://github.com/libsdl-org/SDL/pull/4707#issuecomment-941857093))
  - [x] IME works
  - [x] IME window is properly positioned (even with screen scaling > 100%)
- [x] Linux (different desktop environments?)
  - [x] IME works
  - [ ] IME window is properly positioned (even with screen scaling > 100%)
- [x] macOS
  - [x] IME works
  - [x] IME window is properly positioned (even with screen scaling > 100%)

Some caveats that come with SDL2's implementation of IME support
---

- **[!!] only allows up to 32 bytes (10 UTF-8 encoded CJK characters) per composition/result**
  - honestly the biggest problem with SDL IME support
  - look into native libraries that work better (https://github.com/ryancheung/ImeSharp on Windows?)
- on Windows:
  - `selectionLength` is always 0 (seems to be a bug, but not really worth looking into, as SDL IME has bigger problems)
  - IME window is not visible in fullscreen mode (but seems to be clickable)
    - check behaviour with macOS and Linux; we can set `SDL_HINT_IME_SHOW_UI` to `false` when in fullscreen, per-platform (or globally) as required

Tests
---

Not currently added, exploring some options.

`TestRunHeadlessGameHost` with a custom `ManualTextInputSource` to send text events. (So visual tests not possible?)
`ManualInputManager` to send key presses (to verify `ImeCompositionActive` properly blocks those, even with all the scheduling).

Tests should verify that the compositing logic inside TextBox is correct.
And verify the event flow when consuming multiple events (`OnKeyDown`, `OnPressed(PlatformAction)`, `OnNewImeComposition`, `OnNewImeResult`) that happen and (roughly) the same time. (`OnNewIme*` will arrive sooner than `OnKeyDown` and `OnPressed` as those don't have to go trough `InputManager`.)

---

<details>
  <summary>Patch to test on Windows with partially fixed IME window position (no longer needed)</summary>

```diff
diff --git a/osu.Framework/Input/SDL2DesktopWindowTextInput.cs b/osu.Framework/Input/SDL2DesktopWindowTextInput.cs
index 3f6416eab..da47707d6 100644
--- a/osu.Framework/Input/SDL2DesktopWindowTextInput.cs
+++ b/osu.Framework/Input/SDL2DesktopWindowTextInput.cs
@@ -89,6 +89,9 @@ public void Deactivate()

         public void SetImeRectangle(RectangleF rectangle)
         {
+            if (RuntimeInfo.OS == RuntimeInfo.Platform.Windows)
+                rectangle.Y += rectangle.Height / 2;
+
             window.SetTextInputRect(rectangle);
         }
```
</details>

<details>
  <summary>Patch to test with Noto font</summary>

```diff
diff --git a/osu.Framework/Game.cs b/osu.Framework/Game.cs
index 40f388b3d..58dcf250b 100644
--- a/osu.Framework/Game.cs
+++ b/osu.Framework/Game.cs
@@ -175,6 +175,12 @@ private void load(FrameworkConfigManager config)
             addFont(Fonts, Resources, @"Fonts/FontAwesome5/FontAwesome-Regular");
             addFont(Fonts, Resources, @"Fonts/FontAwesome5/FontAwesome-Brands");

+            addFont(localFonts, Resources, @"Fonts/Noto/Noto-Basic");
+            addFont(localFonts, Resources, @"Fonts/Noto/Noto-Hangul");
+            addFont(localFonts, Resources, @"Fonts/Noto/Noto-CJK-Basic");
+            addFont(localFonts, Resources, @"Fonts/Noto/Noto-CJK-Compatibility");
+            addFont(localFonts, Resources, @"Fonts/Noto/Noto-Thai");
+
             dependencies.Cache(Fonts);

             Localisation = new LocalisationManager(config);
```
</details>

<details>
  <summary>Patch to not close the visual tests on Esc (required because of test input manager)</summary>

```diff
diff --git a/osu.Framework/Testing/TestBrowser.cs b/osu.Framework/Testing/TestBrowser.cs
index 04e7f6c3f..a117a3ecc 100644
--- a/osu.Framework/Testing/TestBrowser.cs
+++ b/osu.Framework/Testing/TestBrowser.cs
@@ -356,7 +356,7 @@ protected override bool OnKeyDown(KeyDownEvent e)
                 switch (e.Key)
                 {
                     case Key.Escape:
-                        exit();
+                        // exit();
                         return true;
                 }
             }
```
</details>


Susko3:
https://github.com/ppy/osu-framework/blob/1b818a973b006616a1d900af9c160657dbd4f96f/osu.Framework/Graphics/UserInterface/TextBox.cs#L784

(Can't comment on this inline, so here goes.)

The composition logic can't handle changes to text from other sources. If text is changed mid-composition, things will probably break.

To ensure things don't break, this `setText()` should probably be scheduled and `FinalizeImeComposition()` or `CancelImeComposition()` should be called just before `setText()` to ensure a clean slate.

My question is: Is it safe to schedule this?
If that's not possible, we could maybe add a flow that prematurely stops the composition (eg. by clearing `imeCompositionDrawables`)

bdach:
Text box used to have schedules on set but they were a constant source of bugs due to delayed effect of normal keyboard operations. See: https://github.com/ppy/osu-framework/pull/4522, https://github.com/ppy/osu/issues/13919, https://github.com/ppy/osu/issues/13920. (https://github.com/ppy/osu-framework/pull/4850 is also tangentially relevant.)

If you can make it work without it devolving into a mess of local hacks around the schedules then I'm all ears. But I wasn't able to.

huoyaoyuan:
Tested with Simplified Chinese IME, version 18362 and 19041. Both works well.
However, `OnUserTextAdded` is not triggered when IME inputs.

Susko3:
New events (added in https://github.com/ppy/osu-framework/pull/4907/commits/6184ec1267ca98b366a3e33c89a2352048266a15) are sent instead of `OnUserTextAdded`. Those are a better fit for the use case in `OsuTextBox`.
`OnImeResult` can be easily changed to `OnUserTextAdded`, if that is deemed necessary. 

<details>
  <summary>Patch for OsuTextBox to test `OnImeComposition` and `OnImeResult`</summary>

```diff
diff --git a/osu.Game/Graphics/UserInterface/OsuTextBox.cs b/osu.Game/Graphics/UserInterface/OsuTextBox.cs
index 96319b9fdd..9940c24749 100644
--- a/osu.Game/Graphics/UserInterface/OsuTextBox.cs
+++ b/osu.Game/Graphics/UserInterface/OsuTextBox.cs
@@ -86,6 +86,8 @@ private void load(OverlayColourProvider? colourProvider, OsuColour colour, Audio
 
         protected override Color4 SelectionColour => selectionColour;
 
+        private void playTextAddedSample() => textAddedSamples[RNG.Next(0, textAddedSamples.Length - 1)]?.Play();
+
         protected override void OnUserTextAdded(string added)
         {
             base.OnUserTextAdded(added);
@@ -93,7 +95,7 @@ protected override void OnUserTextAdded(string added)
             if (added.Any(char.IsUpper) && AllowUniqueCharacterSamples)
                 capsTextAddedSample?.Play();
             else
-                textAddedSamples[RNG.Next(0, 3)]?.Play();
+                playTextAddedSample();
         }
 
         protected override void OnUserTextRemoved(string removed)
@@ -117,6 +119,56 @@ protected override void OnCaretMoved(bool selecting)
             caretMovedSample?.Play();
         }
 
+        protected override void OnImeComposition(string newComposition, int removedTextLength, int addedTextLength, bool caretMoved)
+        {
+            base.OnImeComposition(newComposition, removedTextLength, addedTextLength, caretMoved);
+
+            if (string.IsNullOrEmpty(newComposition))
+            {
+                switch (removedTextLength)
+                {
+                    case 0:
+                        // empty composition event, composition wasn't changed, don't play anything.
+                        return;
+
+                    case 1:
+                        // composition probably ended by pressing backspace, or was cancelled.
+                        textRemovedSample?.Play();
+                        return;
+
+                    default:
+                        // longer text removed, composition ended because it was cancelled.
+                        textRemovedSample?.Play();
+                        return;
+                }
+            }
+
+            if (addedTextLength > 0)
+            {
+                playTextAddedSample();
+                return;
+            }
+
+            if (removedTextLength > 0)
+            {
+                textRemovedSample?.Play();
+                return;
+            }
+
+            if (caretMoved)
+            {
+                caretMovedSample?.Play();
+                return;
+            }
+        }
+
+        protected override void OnImeResult()
+        {
+            base.OnImeResult();
+
+            textCommittedSample?.Play();
+        }
+
         protected override void OnFocus(FocusEvent e)
         {
             BorderThickness = 3;
```
</details>

bdach:
Have briefly tested japanese input via `mozc` on linux. Appears to work well (although I cannot read the language), besides for the length limitation.

Going to look at source now.

osu.Framework/Input/ITextInputSource.cs
```diff
@@ -11,6 +12,9 @@ namespace osu.Framework.Input
     /// </summary>
     public interface ITextInputSource
     {
+        /// <summary>
+        /// Whether the IME is actively providing text composition trough <see cref="OnNewImeComposition"/> and accepting input from the user.
```
bdach, line 16:
```suggestion
        /// Whether the IME is actively providing text composition through <see cref="OnNewImeComposition"/> and accepting input from the user.
```

osu.Framework/Input/ITextInputSource.cs
```diff
@@ -25,7 +29,31 @@ public interface ITextInputSource
...
+
+        /// <summary>
+        /// Resets the IME.
+        /// This clears the current composition string and prepares it for new input.
+        /// </summary>
+        void ResetIme();
+
+        /// <summary>
+        /// Parameters are, in order: composition text, selection start, selection length.
+        /// </summary>
```
bdach, line 46: Not sure this is needed, the delegate type definition already explains the params. xmldoc here should inform as to when the event is called.

osu.Framework/Input/SDL2DesktopWindowTextInput.cs
```diff
@@ -10,13 +11,14 @@ public class SDL2DesktopWindowTextInput : ITextInputSource
     {
         private readonly SDL2DesktopWindow window;
         private string pending = string.Empty;
+        private bool active;
```
bdach, line 14: Can this not use [`SDL_IsTextInputActive`](https://wiki.libsdl.org/SDL_IsTextInputActive) instead? Would prefer querying the source rather than doing local tracking susceptible to ordering & desync issues.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -164,6 +165,9 @@ public virtual bool OnPressed(KeyBindingPressEvent<PlatformAction> e)
             if (!HandleLeftRightArrows && (e.Action == PlatformAction.MoveBackwardChar || e.Action == PlatformAction.MoveForwardChar))
                 return false;
 
+            if (e.Action.IsCommonTextEditingAction() && ImeCompositionActive)
+                return true;
```
bdach, line 169: Just to clarify my understanding, is this done to ensure that framework doesn't interfere with IME operation by handling these actions itself?

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -800,11 +844,26 @@ private void consumePendingText()
...
+        /// <summary>
+        /// Whether there is a ongoing IME composition.
+        /// </summary>
+        /// <remarks>
+        /// The IME should take full input priority, as a lot of the common text editing keys/shortcuts
+        /// are used internally in the IME for compositing.
+        /// Full data about the composition events is processed by <see cref="handleImeComposition"/> "passively"
+        /// so we shouldn't take any action on key events we receive.
+        /// </remarks>
+        protected bool ImeCompositionActive => textInput?.ImeActive == true || imeCompositionDrawables.Count > 0;
```
bdach, line 856: Why does the number of IME drawables factor into this at all? I'd sooner expect to just check the `ImeActive` flag - checking the drawables seems secondary to that and as such unnecessary?

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -1022,72 +1098,218 @@ protected override void OnFocus(FocusEvent e)
...
+                    // or when the keyboard layout and language have changed to a one supported by an IME.
+                    // we can use this opportunity to update the IME window so it appears in
+                    // the correct place once the user starts compositing.
+                    updateImeWindowPosition();
+
+                    // early return as SDL might sometimes send empty text editing events.
+                    // we don't want the currently selected text to be removed in that case
+                    // (we only want it removed once the user has entered _some_ text).
+                    // the composition text hasn't changed anyway, so there is no need to go
+                    // trough the rest of the method.
```
bdach, line 1201:
```suggestion
                    // through the rest of the method.
```

osu.Framework/Graphics/UserInterface/TextBox.cs
Susko3: `textInput?.ImeActive` is `true` if the native underlying composition is active.
But since the IME composition events and key press events are scheduled, it's possible that the events arrive when the native implementation isn't active.

So we also need to block in regards to something that is scheduled.
And `imeCompositionDrawables.Count > 0` is just that. It will be `false` after a scheduled empty IME composition (or a result), and will be `true` when there is some composition happening.

So the time between `textInput?.ImeActive == false` and a scheduled IME action is just enough to block incoming key presses.

---

On second thought, this ordering may break on single-threaded, or different update rate.

osu.Framework/Input/ITextInputSource.cs
Susko3: Leftover from before 1b818a973b006616a1d900af9c160657dbd4f96f.

> xmldoc here should inform as to when the event is called.

Agreed, should also mention that empty composition text means end of composition.

osu.Framework/Input/SDL2DesktopWindowTextInput.cs
Susko3: `active` is used only in `ResetIme()`. `ResetIme()` is used as a corrective action from consumers, requesting the native implementation be reset to a sane state.

So it should take user preference (set by `Activate()` and `Deactivate()`) instead of the actual SDL state.

Could add an xmldoc explaining what `active` actually means (consumer preference).

osu.Framework/Graphics/UserInterface/TextBox.cs
Susko3: Exactly, if textbox text was modified during IME composition, things would break.

osu.Framework/Input/SDL2DesktopWindowTextInput.cs
bdach: Yeah, clarification here would be good.

osu.Framework/Graphics/UserInterface/TextBox.cs
bdach: Given that the schedules look to be constrained solely to the IME path, you may have luck with re-introducing a concept that was specifically intended for IME but removed due to being confusing/unused - [a manual `Scheduler` for IME only](https://github.com/peppy/osu-framework/blob/dde9eb08ede4aa94d21590f92b7c60f85cac3947/osu.Framework/Graphics/UserInterface/TextBox.cs#L106). Then you could force all events scheduled onto that scheduler to execute prior to dropping active IME.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -989,6 +1062,9 @@ protected override void OnMouseUp(MouseUpEvent e)
 
         protected override void OnFocusLost(FocusLostEvent e)
         {
+            if (ImeCompositionActive)
+                FinalizeImeComposition();
```
peppy, line 1066: I'd probably move this into a private helper method since it's present in a few places.

peppy (changes requested):

peppy:
Eventually we'll probably want to render the full IME interface in-game, as this method won't work great in fullscreen mode. Seems like a step in the right direction though.

Susko3:
Upgraded SDL2-CS, have checked that the IME changes have made it in. Didn't test anything else.

osu.Framework/Graphics/UserInterface/TextBox.cs
Susko3: So I took a shot at implementing `textUpdateScheduler` for IME only, but I quickly ran into some problems, such as the caret ending up in the wrong position.

Quickly realized that updating `textUpdateScheduler` in `updateCursorAndLayout()` / `UpdateAfterChildren()` will break it, and updating in `Update()` produces results as expected.

It makes sense that it was broken when updating after children. If you schedule some text operation that will modify children of TextFlow _after_ updating children, and then proceed to use properties of said children (without updating them), you can expect things to break.

My hypothesis is that `textUpdateScheduler` has always been broken because it was updated too late, and by updating in `Update()` it may actually work as intended, as `TextFlow` drawables will get updated only after all changes to are applied to text.

So I tested https://github.com/ppy/osu/issues/13919 with this hypothesis.

Checked out that tag of both game and framework and applied this diff:

```diff
diff --git a/osu.Framework/Graphics/UserInterface/TextBox.cs b/osu.Framework/Graphics/UserInterface/TextBox.cs
index 9fdb95261..87938fdbd 100644
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -299,12 +299,16 @@ protected override void Dispose(bool isDisposing)

         private string textAtLastLayout = string.Empty;

+        protected override void Update()
+        {
+            base.Update();
+            textUpdateScheduler.Update();
+        }
+
         private void updateCursorAndLayout()
         {
             Placeholder.Font = Placeholder.Font.With(size: CalculatedTextSize);

-            textUpdateScheduler.Update();
-
             float cursorPos = 0;
             if (text.Length > 0)
                 cursorPos = getPositionAt(selectionLeft);
```

And the issue was fixed! Also, all framework and game `TextBox` tests passed.

So maybe we can bring back `textUpdateScheduler` for all text events, but `Update()` it at a better time. (Not in this PR, of course.)

I'll test some more, and if all goes well, add it for IME events in this PR.

bdach: At the point where the scheduler is being updated every `Update()` call (and only then) you may as well not have it and [use `cursorAndLayout` instead as `UpdateAfterChildren()` currently does](https://github.com/ppy/osu-framework/blob/fe00cc23f6393fd8891fc1d94e3ee37041e242fc/osu.Framework/Graphics/UserInterface/TextBox.cs#L417).

Incidentally, in the past I recall trying a similar solution, but discarding it, because at the point where both `Update()` and `UpdateAfterChildren()` are re-validating the same thing it feels like they both really should not be necessary and just one of them should theoretically suffice? Not sure.

Susko3: > both Update() and UpdateAfterChildren() are re-validating the same thing it feels like they both really should not be necessary and just one of them should theoretically suffice?

I don't think they're validating the same thing. From my understading:

`textUpdateScheduler` collects all the text changes and applies them as soon as possible (in the next update). Modifying, deleting, and adding drawables to `TextFlow` with `AddCharacterToFlow()`, and also changing `selectionStart/End`.

After that, `TextFlow` and child text drawables [are updated](https://github.com/ppy/osu-framework/blob/7e6e48c4d3433b74f941ff321a7d02bce4ce2853/osu.Framework/Graphics/Containers/CompositeDrawable.cs#L898). Now those drawables have correct positions and sizes (this is what `UpdateSubTree()` does, right?) so we can use those in calculating where the Caret should be.

So we do exactly that at the earliest time possible, in `UpdateAfterChildren()`. We call `updateCursorAndLayout()` if invalid.

`updateCursorAndLayout()` [only changes `Caret` and `TextContainer` positions](https://github.com/ppy/osu-framework/blob/dba74d00d0d23010abd6e3d3c0a766860d2f1508/osu.Framework/Graphics/UserInterface/TextBox.cs#L441-L444), and doesn't change anything about the `TextFlow`.

So updating at both times are required.

Susko3:
I'm hoping we can get https://github.com/ppy/osu-framework/pull/4908 in before this one, so that IME behaviour can be tested.

bdach:
@Susko3 before I go to re-review this, do you have other changes you'd like to make prior to that?

Susko3:
Things will mostly stay the same, only things that I'm looking to change is the `OnImeComposition` (the "event" similar to `OnUserTextAdded`) in TextBox and some tests.

bdach:
I'll give it a look over this weekend then, even if the remaining changes won't be there yet.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -1078,66 +1210,295 @@ private void unbindInput()
...
+
+            // remove characters that can't be added.
+
+            for (int index = 0; index < composition.Length; index++)
+            {
+                if (!canAddCharacter(composition[index]))
                 {
-                    d.Colour = Color4.White;
-                    d.FadeTo(1, 200, Easing.Out);
+                    composition = composition.Remove(index, 1);
```
bdach, line 1277: This is going to allocate a fair few strings if characters are going to be removed one by one, but I suppose I could look away for now...

bdach (commented):
Skimmed through this. I will say that I am really spooked by the complexity of this pull at this stage. It would be really helpful to get as much test coverage on this as feasible.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -356,6 +383,41 @@ protected void DeleteBy(int amount)
             }
         }
 
+        /// <summary>
+        /// Finalize the current IME composition if one is active.
+        /// </summary>
+        /// <remarks>Must only be called from the update thread.</remarks>
+        protected void FinalizeImeComposition()
+        {
+            if (!ImeCompositionActive && !textUpdateScheduler.HasPendingTasks)
```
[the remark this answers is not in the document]
Susko3: > Should this method really do nothing if there are no pending IME tasks on the scheduler?

That is not necessarily true. If a composition is active, the method will normally complete.

Checking the scheduler was only added as a last resort, in case `ImeCompositionActive` returns `false`.
Added to ensure this:
https://github.com/ppy/osu-framework/blob/ec18b129f49392720503c0c4f42e52eeb0de7b3e/osu.Framework/Graphics/UserInterface/TextBox.cs#L400-L402

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -161,7 +175,14 @@ private void load(GameHost host)
         protected override void LoadComplete()
         {
             base.LoadComplete();
+
             isActive.BindValueChanged(_ => Scheduler.AddOnce(updateCaretVisibility));
+            current.BindDisabledChanged(disabled =>
+            {
+                if (disabled)
+                    Scheduler.AddOnce(FinalizeImeComposition);
```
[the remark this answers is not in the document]
Susko3: First of all, `onImeResult()` (and by extension, `FinalizeImeComposition`) does not mutate `Text`, only thing it does is change character drawable opacity and move the cursor/caret. Next, `onImeComposition()` [won't do anything if `Current` is disabled](https://github.com/ppy/osu-framework/blob/a507938e89c3e4e1b9e0283750433085c25ee0da/osu.Framework/Graphics/UserInterface/TextBox.cs#L1346-L1350).

So if `Current` is disabled, it and `Text` won't be changed as a result of IME composition.

For clarity, we could clear `textUpdateScheduler`, as the tasks there will be no-op anyways.

I can see this crashing if `Current` was disabled in between `beginTextChange()` and `endTextChange()` (should never happen under normal circumstances). But this can crash when adding text by any means, not just when adding trough `onImeComposition()`.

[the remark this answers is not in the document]
bdach: I would prefer clearing the scheduler, because even if `onImeComposition()` is a no-op, `NotifyInputError()` could still potentially be raised multiple times in a frame. But maybe it's not a huge deal, I dunno.

The other thing is that `FinalizeImeComposition()` also raises `NotifyInputError()` if the textbox is disabled, which is maybe slightly controversial.

osu.Framework/Graphics/UserInterface/TextBox.cs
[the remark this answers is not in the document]
bdach: Maybe an inline comment at least what the `HasPendingTasks` check is for? These sorts of checks really do tend to become inscrutable N months from now when nobody remembers what the code did anymore.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -822,9 +917,14 @@ protected void EndConsumingText()
...
         /// Consumes any pending characters and adds them to the textbox if not <see cref="ReadOnly"/>.
         /// </summary>
-        /// <returns>Whether any characters were consumed.</returns>
         private void consumePendingText()
         {
+            if (!inputBound)
+            {
+                EndConsumingText();
+                return;
+            }
```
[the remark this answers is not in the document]
bdach: That sounds like this is a separate regression/bug that would occur in `master` and not something related to IME itself, but sure...

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -188,8 +213,11 @@ public virtual bool OnPressed(KeyBindingPressEvent<PlatformAction> e)
                     return true;
 
                 case PlatformAction.Paste:
+                    string pending = null;
+
                     //the text may get pasted into the hidden textbox, so we don't need any direct clipboard interaction here.
-                    string pending = textInput.GetPendingText();
+                    if (inputBound)
+                        pending = textInput.GetPendingText();
```
peppy, line 220: Is there a reason this can't be a call to `consumePendingText()` instead?

Susko3: Not sure. I think this is used on iOS only, as that implementation uses a native textbox for text input. Maybe the check below is needed to prevent text getting pasted twice (reading from both the clipboard and the native textbox)?

Anyways, I don't have an iOS device to test this, and I think this is out of scope for this PR. Only reason this code is modified is to add a sanity check. In retrospect, https://github.com/ppy/osu-framework/pull/4907/commits/a507938e89c3e4e1b9e0283750433085c25ee0da should have been a separate PR.

peppy: Fair enough.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -782,6 +905,11 @@ private void setText(string value)
         {
             bool beganChange = beginTextChange();
 
+            // finalize and cleanup the IME composition (if one is active) so we get a clean slate for pending text changes and future IME composition.
+            // `IsLoaded` check is required because `Text` could be set in the initializer / before the drawable loaded.
+            // `FinalizeImeComposition()` crashes if textbox isn't fully loaded.
+            if (IsLoaded) FinalizeImeComposition(false);
```
Susko3, line 911: `setText()` being called before textbox is fully loaded seems a bit strange. Given that `setText(Text)` is called in `LoadComplete()` I think this is unwanted.

Out of scope for this PR, so a quick and dirty `IsLoaded` check is required for now. (`FinalizeImeComposition()` crashes when trying to access `textInput` with nullref.)

bdach: Right now `setText` is the one thing that upholds the invariant `text == Text == Current.Value`. So however peculiar it might be that it's getting called before load complete, that needs to be the case unless something else can uphold that invariant. `Text` and `Current.Value` are data and need to have correct values always.

peppy:
Tested on Windows 11 (100% / 200%) and macOS (normal and HiDPI) to work as expected. Feels good.

peppy (approved):

PR #4908, opened 2021-11-21
Fixes for `TextBox`, and a `TextInputSource` refactor

Susko3:
Closes https://github.com/ppy/osu/issues/15716.

Discussed in: https://github.com/ppy/osu/issues/15716#issuecomment-974819198.

`ITextInputSource` has been refactored into `TextInputSource` with no usage changes, and almost no behavioral changes to consumers.

The only slight difference is that `GetPendingText()` is first come, first served instead of every `TextBox` having it's own text input to consume from. Shouldn't be a problem as we expect only one textbox to be active (and be actively consuming) at a time.

If we switch to event based text input, could have two text boxes get text just fine (even if it doesn't make much sense).

Refactoring `ITextInputSource` into a class turned out great, removing a bunch of duplicate code in the process.

Android had a slightly different [`GetPendingText()`](https://github.com/ppy/osu-framework/blob/d7bb25be651d040054c533b63a35f78176e4081a/osu.Framework.Android/Input/AndroidTextInput.cs#L31-L39), but does work as expected with the new one. (The old lock never really made sense, as [appending to pending](https://github.com/ppy/osu-framework/blob/d7bb25be651d040054c533b63a35f78176e4081a/osu.Framework.Android/Input/AndroidTextInput.cs#L50) wasn't locking.)

If you think locking is better than `try..finally`, now is the perfect time to bring it up.

Check the commit messages for more detailed changes.


osu.Framework/Input/TextInputSource.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// Generally hides a native implementation from the game framework.
+    /// </summary>
+    public class TextInputSource
+    {
+        /// <summary>
+        /// Whether the IME is actively providing text composition through <see cref="OnImeComposition"/> and accepting input from the user.
+        /// </summary>
+        public bool ImeActive { get; protected set; }
+
+        protected string PendingText { get; set; }
```
peppy, line 20: Don't feel confident with this being `protected` (mostly the `set` portion).

I think it can be set to `private` and a new `protected void AddPendingText(string)` method can handle the set usages.

osu.Framework/Input/TextInputSource.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        /// <summary>
+        /// Gets all the text that was inputted by the user since the last <see cref="GetPendingText"/> call.
+        /// </summary>
+        /// <remarks>
+        /// Should be periodically called to collect user-inputted text.
+        /// </remarks>
+        public string GetPendingText()
+        {
+            try
```
peppy, line 30: Definitely needs to be a `lock`. The same lock should also be used in `AddPendingText`.

osu.Framework/Input/TextInputSource.cs
```diff
@@ -0,0 +1,122 @@
...
+                ActivateTextInput();
+        }
+
+        /// <summary>
+        /// Ensures that the native implementation that retrieves user text input is activated
+        /// and that the user can start entering text.
+        /// </summary>
+        public void EnsureActivated()
+        {
+            if (activationCounter >= 1)
```
peppy, line 64: Reading through the discussion, I'm wondering if this ever exceeds `1` in current usages. If it's guaranteed not to, maybe we should be guarding against it rather than blindly supporting it.

peppy (changes requested):

osu.Framework/Input/TextInputSource.cs
Susko3: It's possible that textbox `TextBox A` has focus and is pending for dispose, while another textbox `FocusedTextBox B` has just obtained focus. Then `activationCounter` is `2`, and if the user clicks on `B`, [EnsureActivated will fire](https://github.com/ppy/osu-framework/blob/f2d719a103a60a0cff28d1ab4f25c9eaeaf1424e/osu.Framework/Graphics/UserInterface/TextBox.cs#L1006) -- this is somewhat expected with async disposal. (Though I'm guessing most textboxes will be unfocused before disposal.)

`EnsureActivated()` is a call to the native implementation saying: reset to a sane state, and allow the user to enter text.
This should always succeed if that textinput is active. Doesn't matter that two consumers requested activation.

Susko3:
Since `TextInputSource` is now cached, we can easily override it for tests.

https://github.com/ppy/osu-framework/blob/3a511100d2f5d75cd0e9b4e1eba8c2fb282c7285/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBoxEvents.cs#L71-L78

I'm looking to get this test working. Just have a question:

Where do I cache the manual `TextInputSource`?
In the test class itself, or in a separate container (and make the test text box a child of said container)?

frenzibyte:
It can be done in either way, making it cached in a container results in no need to resetting it every `SetUp` call, while making it cached directly means more code.

I would've made it cached directly and reset it on `SetUp` for less code (unless that's not feasible), but either way works really.

peppy (approved):

PR #4910, opened 2021-11-22
Hide `TextBox` caret when `GameHost` is inactive / keyboard focus is lost

Susko3:
A blinking cursor signals to the user that keyboard input will appear at that cursor.
This PR hides the caret when we lose keyboard focus to conform with that expectation.

Only hides the caret if there is no selection. The current show/hide logic on focus/unfocus is kept.
Native applications will usually grey-out the selection when inactive, but that is unnecessary for us.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -155,14 +155,13 @@ namespace osu.Framework.Graphics.UserInterface
         {
             textInput = host.GetTextInput();
             clipboard = host.GetClipboard();
-
             isActive = host.IsActive.GetBoundCopy();
-            isActive.BindValueChanged(_ => Scheduler.AddOnce(updateCaretVisibility));
         }
 
         protected override void LoadComplete()
         {
             base.LoadComplete();
+            isActive.BindValueChanged(_ => Scheduler.AddOnce(updateCaretVisibility));
             setText(Text);
         }
 
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -726,13 +726,10 @@ namespace osu.Framework.Graphics.UserInterface
 
         private void updateCaretVisibility()
         {
-            bool newVisibility =
-                // only show if we're focused
-                HasFocus && (
-                    // and the host is active
-                    isActive.Value
-                    // or if text is selected
-                    || selectionLength != 0);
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -709,6 +719,34 @@ public LocalisableString PlaceholderText
...
+
+        private void updateCaretVisibility()
+        {
+            bool newVisibility =
+                // only show if we're focused
+                HasFocus && (
+                    // and the host is active
+                    isActive.Value
+                    // or if text is selected
+                    || selectionLength != 0);
```
frenzibyte, line 735: would probably be better grouping the code and comment into one-liners, feels off to read it like that.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -148,6 +155,9 @@ private void load(GameHost host)
         {
             textInput = host.GetTextInput();
             clipboard = host.GetClipboard();
+
+            isActive = host.IsActive.GetBoundCopy();
+            isActive.BindValueChanged(_ => Scheduler.AddOnce(updateCaretVisibility));
```
bdach, line 160: `BindValueChanged` callbacks are usually registered in `LoadComplete()`.

Also this should presumably run the first time to avoid showing the caret if the window is initially unfocused? (i.e. call `BindValueChanged(..., true)`).

osu.Framework/Graphics/UserInterface/TextBox.cs
bdach: I would even go as far as to say that the comments bring no significant value as they mostly restate the code. A more pertinent comment would explain *why* such a conditional was written, not what was in it.

bdach (changes requested):

osu.Framework/Graphics/UserInterface/TextBox.cs
Susko3: Caret is [initially hidden](https://github.com/ppy/osu-framework/pull/4910/files#diff-0931d11ac8154d00e8c67e85c8dd9e86cbacc227d9bbe2eab1a0232bd56b77adR149-R150), so no problem there.

bdach:
Implementation seems generally sane, and the UX looks to match standard textbox UI paradigms. Going to request a double check of this first, however.

bdach (approved):

peppy (approved):

PR #4913, opened 2021-11-24
Add x86-64 architectures back for BASS iOS native libraries

frenzibyte:
During the BASS native libraries bump in #4291, osu!framework has lost support for x86-64 iOS (simulator arch).

I fiddled around with replacing the current `.a` native libraries with the full `.xcframework` provided in the BASS website, but turns out Xamarin don't support the type of XCFramework BASS provides. (as theirs wrap up the native libraries directly, which is a valid type of XCFramework, while Xamarin expects any referenced XCFramework to always contain a `.framework` inside)

For the time being, I've combined the libraries of both architectures using the following commands (macOS only):
```sh
# arm64 exists in both the arm and i386 versions, it has to be removed in one of them to combine correctly, otherwise an error will be thrown.
lipo -remove arm64 ./libbass.xcframework/ios-arm64_i386_x86_64-simulator/libbass.a -output ./libbass.xcframework/ios-arm64_i386_x86_64-simulator/libbass.a

lipo -create ./libbass.xcframework/ios-arm64_armv7_armv7s/libbass.a ./libbass.xcframework/ios-arm64_i386_x86_64-simulator/libbass.a -output ./libbass.a
```

To confirm the native library contains all the architectures needed, typing `lipo -detailed_info {library}` should result in the following:
```
Fat header in: ./{library}
fat_magic 0xcafebabe
nfat_arch 5
architecture armv7
    cputype CPU_TYPE_ARM
    cpusubtype CPU_SUBTYPE_ARM_V7
    capabilities 0x0
    offset 108
    size 235864
    align 2^2 (4)
architecture armv7s
    cputype CPU_TYPE_ARM
    cpusubtype CPU_SUBTYPE_ARM_V7S
    capabilities 0x0
    offset 235972
    size 234272
    align 2^2 (4)
architecture i386
    cputype CPU_TYPE_I386
    cpusubtype CPU_SUBTYPE_I386_ALL
    capabilities 0x0
    offset 470244
    size 298640
    align 2^2 (4)
architecture x86_64
    cputype CPU_TYPE_X86_64
    cpusubtype CPU_SUBTYPE_X86_64_ALL
    capabilities 0x0
    offset 768888
    size 377032
    align 2^3 (8)
architecture arm64
    cputype CPU_TYPE_ARM64
    cpusubtype CPU_SUBTYPE_ARM64_ALL
    capabilities 0x0
    offset 1145920
    size 314888
    align 2^3 (8)
```

---

This should now bring back support for running osu! under iOS simulator, and also removes the last blocker for running iOS CI on osu! (although it's for `osu.iOS.csproj` only`, osu! tests have a separate issue)

I've made [a change in a separate branch](https://github.com/frenzibyte/osu-framework/tree/ios-ci-improvement) to build not only `osu.Framework.iOS.csproj`, but `SampleGame.iOS.csproj` and `osu.Framework.Tests.iOS.csproj` as well, to reproduce the issue that this PR resolves:
 - Before: https://github.com/frenzibyte/osu-framework/runs/4310539778?check_suite_focus=true
 - After: https://github.com/frenzibyte/osu-framework/runs/4310626424?check_suite_focus=true
 
Not sure if it's fine PR'ing this change to cover building on `SampleGame.iOS` and `osu.Framework.Tests.iOS`, given that it builds each project separately, which may presumably take longer than building a solution.


bdach:
please link the source of the binaries.

frenzibyte:
Well I have mentioned that I combined the libraries of both arm and x86-64, each residing in the XCFramework of the latest BASS, BASS FX, and BASSmix releases: 
 - http://www.un4seen.com/download.php?bass24-ios
 - http://www.un4seen.com/download.php?z/0/bass_fx24-ios
 - http://www.un4seen.com/download.php?bassmix24-ios

peppy:
Only concern is that we were using an early release version of bassmix in the past, but I'm guessing that's already in the mainstream releases if this is working for you.

peppy (approved):

smoogipoo:
The script in the description should be committed to the repository.

Also, for future reference (hi, fellow git-blamer!):
```
bass => 2.4.16.7 (unchanged)
bass_fx => 2.4.12.6 (unchanged)
bassmix => 2.4.11.1 (unchanged)
```

peppy:
Had a feeling it should have been a script, I just made a mental note to come looking for this next time we get to updating the libraries. A script sitting next to the files appropriately named would be a good reminder.

frenzibyte:
Wasn't sure if a script is worth it given the simplicity of the commands, but will see.

PR #4915, opened 2021-11-26
Add support for nested AudioMixers

nekodex:
This functionality is a pre-requisite for applying processing to groupings of audio channels in Lazer, and affords the ability to chain mixers with different processing/effects.

Utilising this capability, a `GlobalMixer` has been added (which `TrackMixer` and `SampleMixer` are now routed into). This will allow for monitoring of the final summed audio output and for any final audio processing to happen just prior to output.

### DIFF
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -54,6 +54,11 @@ namespace osu.Framework.Audio
         private readonly AudioThread thread;
 
         /// <summary>
+        /// The global mixer which all audio routes through before output.
+        /// </summary>
+        public readonly AudioMixer GlobalMixer;
+
+        /// <summary>
         /// The global mixer which all tracks are routed into by default.
         /// </summary>
         public readonly AudioMixer TrackMixer;
@@ -161,8 +166,9 @@ namespace osu.Framework.Audio
                 return store;
             });
 
-            AddItem(TrackMixer = createAudioMixer(null, nameof(TrackMixer)));
-            AddItem(SampleMixer = createAudioMixer(null, nameof(SampleMixer)));
+            AddItem(GlobalMixer = new BassAudioMixer(null, nameof(GlobalMixer)));
+            GlobalMixer.Add(TrackMixer = new BassAudioMixer(GlobalMixer, nameof(TrackMixer)));
+            GlobalMixer.Add(SampleMixer = new BassAudioMixer(GlobalMixer, nameof(SampleMixer)));
 
             CancellationToken token = cancelSource.Token;
 
@@ -221,15 +227,17 @@ namespace osu.Framework.Audio
         /// Creates a new <see cref="AudioMixer"/>.
         /// </summary>
         /// <remarks>
-        /// Channels removed from this <see cref="AudioMixer"/> fall back to the global <see cref="SampleMixer"/>.
+        /// Audio output from the <see cref="AudioMixer"/> will be routed to <see cref="GlobalMixer"/>.
         /// </remarks>
         /// <param name="identifier">An identifier displayed on the audio mixer visualiser.</param>
-        public AudioMixer CreateAudioMixer(string identifier = default) => createAudioMixer(SampleMixer, !string.IsNullOrEmpty(identifier) ? identifier : $"user #{Interlocked.Increment(ref userMixerID)}");
+        public AudioMixer CreateAudioMixer(string identifier = default) => createAudioMixer(GlobalMixer, !string.IsNullOrEmpty(identifier) ? identifier : $"user #{Interlocked.Increment(ref userMixerID)}");
 
-        private AudioMixer createAudioMixer(AudioMixer globalMixer, string identifier)
+        private AudioMixer createAudioMixer(AudioMixer parentMixer, string identifier)
         {
-            var mixer = new BassAudioMixer(globalMixer, identifier);
-            AddItem(mixer);
+            var mixer = new BassAudioMixer(parentMixer, identifier);
+
+            parentMixer.Add(mixer);
+
             return mixer;
         }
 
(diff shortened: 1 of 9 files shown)


osu.Framework/Audio/Mixing/IAudioMixer.cs
```diff
@@ -23,6 +23,11 @@ public interface IAudioMixer
         /// </summary>
         BindableList<IEffectParameter> Effects { get; }
 
+        /// <summary>
+        /// The currently active channels.
+        /// </summary>
+        BindableList<IAudioChannel> Channels { get; }
```
bdach, line 29: Unsure if this needs to be public. To what end is this exposed? If it is just for the mixer visualiser, then maybe `internal` or similar is better?

osu.Framework/Audio/Mixing/AudioMixer.cs
```diff
@@ -54,11 +61,11 @@ public void Add(IAudioChannel channel)
...
         /// </summary>
         /// <param name="channel">The <see cref="IAudioChannel"/> to remove.</param>
-        /// <param name="returnToDefault">Whether <paramref name="channel"/> should be returned to the default mixer.</param>
-        protected void Remove(IAudioChannel channel, bool returnToDefault)
+        /// <param name="moveToParent">Whether <paramref name="channel"/> should be re-routed to the parent mixer.</param>
+        protected void Remove(IAudioChannel channel, bool moveToParent)
         {
-            // If this is the default mixer, prevent removal.
-            if (returnToDefault && globalMixer == null)
+            // If this is a top-level mixer, prevent removal.
```
bdach, line 67: Not sure about this comment, it seems misleading (and it arguably already was on `master`). Removal is only prevented if `moveToParent` is true.

osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
```diff
@@ -255,17 +266,21 @@ public void UpdateDevice(int deviceIndex)
...
-                var channel = activeChannels[i];
+                var channel = ActiveChannels[i];
                 if (channel.IsActive)
                     continue;
 
-                activeChannels.RemoveAt(i--);
+                ActiveChannels.RemoveAt(i--);
+
+                if (channel is BassAudioMixer bassAudioMixer && bassAudioMixer.IsDisposed)
+                    return;
```
bdach, line 278: Is this right? So a disposed mixer in the middle of the list of active channels will just... cause all of the channels after that to not be processed at all?

osu.Framework/Audio/AudioManager.cs
```diff
@@ -221,15 +227,17 @@ private void onDevicesChanged()
         /// Creates a new <see cref="AudioMixer"/>.
         /// </summary>
         /// <remarks>
-        /// Channels removed from this <see cref="AudioMixer"/> fall back to the global <see cref="SampleMixer"/>.
+        /// Audio output from the <see cref="AudioMixer"/> will be routed to <see cref="GlobalMixer"/>.
         /// </remarks>
         /// <param name="identifier">An identifier displayed on the audio mixer visualiser.</param>
-        public AudioMixer CreateAudioMixer(string identifier = default) => createAudioMixer(SampleMixer, !string.IsNullOrEmpty(identifier) ? identifier : $"user #{Interlocked.Increment(ref userMixerID)}");
+        public AudioMixer CreateAudioMixer(string identifier = default) => createAudioMixer(GlobalMixer, !string.IsNullOrEmpty(identifier) ? identifier : $"user #{Interlocked.Increment(ref userMixerID)}");
```
bdach, line 233: So I guess it is not possible to create an audio mixer through this method that is not under the top-level mixer? But you could potentially switch it in post by `Add`ing the mixer to another child mixer? Not sure how to feel about that.

In general the logic of association mixer channels with sub-mixer channels feels... off to me somehow. I can't follow it well, I'm still not 100% sure what happens where and why.

bdach (commented):
Aside from the CI failure:

osu.Framework/Audio/Mixing/AudioMixer.cs
```diff
@@ -96,9 +79,13 @@ protected void Remove(IAudioChannel channel, bool moveToParent)
 
         #region IAudioChannel
 
-        public AudioMixer? Mixer { get; }
+        public virtual AudioMixer? Mixer { get; set; }
```
peppy, line 82: I can see that this naming is used to match the `IAudioChannel` interface, but `mixer.Mixer = parent` reads pretty weird.

osu.Framework/Audio/Mixing/AudioMixer.cs
```diff
@@ -96,9 +79,13 @@ protected void Remove(IAudioChannel channel, bool moveToParent)
...
 
-        AudioMixer? IAudioChannel.Mixer { get; set; }
+        AudioMixer? IAudioChannel.Mixer
+        {
+            get => Mixer;
+            set => Mixer = value;
+        }
```
peppy, line 88: This looks to be implemented the same just above, I don't think this explicit  implementation is required any more.

peppy (commented):
Just a few things I noticed from the diff of new changes.

osu.Framework/Audio/Mixing/AudioMixer.cs
nekodex: Do you mean the part where it's named `Mixer`? If so, I tend to agree but not sure of a better name... `OutputTarget` maybe? but then if it's of type `AudioMixer`, that's also weird, so...

peppy: Yeah. Not sure on a solution yet.

frenzibyte: How about calling it `Parent` and explicitly implement `IAudioChannel.Mixer` to it:
```csharp
public AudioMixer? Mixer { get; set; }

AudioMixer? IAudioChannel.Mixer => Parent;
```

peppy: I was going to suggest that but for some reason I thought it wouldn't work. I think it's an amicable solution.

PR #4916, opened 2021-11-26
Update AudioMixerVisualiser to support nested mixers

nekodex:
Also refactored it a bit along the way to make it a less coupled to BASS (for whatever that's worth?).

I'm also kinda passing around the levels in a `float[]` (as returned from BASS), I'm not sure if this is the best approach? Worth creating a data structure for it or nah?

- [ ] depends on #4915 (for `GlobalMixer` and its hierarchy)

(I broke this out into a separate PR as the other one got kinda big.)


nekodex:
Closing this as it's now part of #4915, which uses the functionality for the visual test.

PR #4917, opened 2021-11-26
Fix caret potentially going out of textbox bounds with sufficiently high LeftRightPadding

Susko3:
For example, latest `master` with `LeftRightPadding = 100`:

https://user-images.githubusercontent.com/16479013/143657321-d029b6e7-1c18-4e6d-8dd0-84001cb2448d.mp4

Happens because [this check](https://github.com/ppy/osu-framework/blob/fe00cc23f6393fd8891fc1d94e3ee37041e242fc/osu.Framework/Graphics/UserInterface/TextBox.cs#L389-L392) doesn't succeed because of wrongly calculated `cursorRelativePositionAxesInBox`.

This isn't easily noticeable with TextBox's default of 5, or OsuTextBox's default of 10, but the caret will go ever so slightly more to the right before scrolling the view.

**This PR:**

https://user-images.githubusercontent.com/16479013/143657575-e6629333-4108-46a1-9019-b465dbcc4297.mp4

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
 using osu.Framework.Graphics.UserInterface;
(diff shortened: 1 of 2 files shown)


bdach:
Would like to see unit testing coverage for this, as the property in question is rather obscure/niche.

bdach:
The fix looks good, but I found the magic values in the test cases a bit too magic, so I took the liberty of rewriting the test cases a touch to be less specific and more explicit as to which behaviours are being tested.

Looks good to me now, but going to wait for a double check.

bdach (approved):

peppy (approved):

PR #4918, opened 2021-11-27
Introduce a pluralisable string type

Game4all:
PRing this in a working state to discuss about integrating this in the localisation pipeline.

This implements a `PluralisableString` type deriving from `TranslatableString`, that can display a pluralised variant of a translation according to the plural context.

I initially had the 'separator character' delimiting two variants in localised text as a ctor parameter but I guess it should be fine to have it be hardcoded as a '|' for now, which corresponds to what's used on web-side to separate plural variants altough I'm opened to make it a parameter again.

The integration should be transparent on osu!-side with required changes being on `osu-localisation-analyzer` end.
Plans to do so sum up to adding heuristics to detect strings which support pluralisation (contains the `|` delimiter) and use `PluralisableString` in place of `TranslatableString` in that case. I'll mention that I haven't delved into it yet but according to my memory from a previous look at the source, that should be pretty straightforward.

The diffstat makes this PR looks big where in fact more than half of it comes from the locale -> plural count mapping.

## Implementation references

https://github.com/laravel/framework/blob/f60d3da8f9ad236d339562cd89eb5dfde4eb7dce/src/Illuminate/Translation/MessageSelector.php#L99 for code-based plural rules
Spec at : https://unicode-org.github.io/cldr-staging/charts/latest/supplemental/language_plural_rules.html

### DIFF
--- a/osu.Framework/Localisation/PluralisableString.cs
+++ b/osu.Framework/Localisation/PluralisableString.cs
@@ -8,6 +8,9 @@ using System.Linq;
 
 namespace osu.Framework.Localisation
 {
+    /// <summary>
+    /// A string which can display a plural variant of a localised string according to the plural context.
+    /// </summary>
     public class PluralisableString : TranslatableString
     {
         private const char variant_delimiter = '|';
--- a/osu.Framework/Localisation/PluralisableString.cs
+++ b/osu.Framework/Localisation/PluralisableString.cs
@@ -34,7 +34,7 @@ namespace osu.Framework.Localisation
         protected override string GetLocalisedFormat(LocalisationParameters parameters, string format)
         {
             string[] variants = format.Split(Separator);
-            return variants.ElementAtOrDefault(getPluralIndex(parameters)) ?? variants.ElementAt(variants.Length);
+            return variants.ElementAtOrDefault(getPluralIndex(parameters)) ?? variants.ElementAt(variants.Length - 1);
         }
 
         public bool Equals(PluralisableString? other)
--- a/osu.Framework/Localisation/PluralisableString.cs
+++ b/osu.Framework/Localisation/PluralisableString.cs
@@ -13,15 +13,12 @@ namespace osu.Framework.Localisation
     /// </summary>
     public class PluralisableString : TranslatableString
     {
+        private const char variant_separator = '|';
+
         /// <summary>
         /// The plural count to use when applying plural rules.
         /// </summary>
         public readonly int Count;
-
-        /// <summary>
-        /// The character to use as a plural variant separator in localised text.
-        /// </summary>
-        public readonly char Separator;
 
         /// <summary>
         /// Creates a <see cref="PluralisableString"/> using texts.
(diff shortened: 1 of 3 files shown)


osu.Framework/Localisation/TranslatableString.cs
```diff
@@ -71,14 +73,14 @@ public string GetLocalised(LocalisationParameters parameters)
             catch (FormatException e)
             {
                 // The formatting has failed
-                Logger.Log($"Localised format failed. Key: {Key}, culture: {parameters.Store.EffectiveCulture}, fallback format string: \"{Fallback}\", localised format string: \"{localisedFormat}\". Exception: {e}",
+                Logger.Log($"Localised format failed. Key: {Key}, culture: {parameters.Store.EffectiveCulture}, fallback format string: \"{GetLocalisedFormat(parameters, Fallback)}\", localised format string: \"{localisedFormat}\". Exception: {e}",
```
bdach, line 76: Why does this call `GetLocalisedFormat()` again rather than use the `localisedFormat` local as previously?

osu.Framework/Localisation/TranslatableString.cs
```diff
@@ -90,7 +92,7 @@ public bool Equals(TranslatableString? other)
                    && Args.SequenceEqual(other.Args);
         }
 
-        public bool Equals(ILocalisableStringData? other) => other is TranslatableString translatable && Equals(translatable);
+        public virtual bool Equals(ILocalisableStringData? other) => other is TranslatableString translatable && Equals(translatable);
```
bdach, line 95: If `other` is a `PluralisableString`, this implementation is potentially wrong. The class type checking here should be done in an invariant manner (i.e. using `GetType() == other.GetType()` rather than via `is` - because in this case, a `PluralisableString` theoretically *is* a `TranslatableString`).

osu.Framework/Localisation/PluralisableString.cs
```diff
@@ -0,0 +1,362 @@
...
+        {
+            Count = count;
+        }
+
+        protected override string GetLocalisedFormat(LocalisationParameters parameters, string format)
+        {
+            string[] variants = format.Split(variant_separator);
+
+            //use the last plural form available for the current locale if the requested form is missing from translated strings.
+            return variants.ElementAtOrDefault(getPluralIndex(parameters)) ?? variants.ElementAt(variants.Length - 1);
```
bdach, line 53: Not sure about using `.ElementAt()` on an array. As far as I can tell [it shouldn't incur allocs](https://github.com/dotnet/runtime/blob/v5.0.11/src/libraries/System.Linq/src/System/Linq/ElementAt.cs) but you may as well write

```suggestion
            // use the last plural form available for the current locale if the requested form is missing from translated strings.
            int pluralIndex = getPluralIndex(parameters);
            return variants[Math.Min(pluralIndex, variants.Length - 1)];
```

bdach (commented):
Not 100% on the `PluralisableString : TranslatableString` inheritance. Feels it complicates code and maybe it may be easier to duplicate the pieces that need to be duplicated for it to work, dunno.

Game4all:
> Not 100% on the `PluralisableString : TranslatableString` inheritance. Feels it complicates code and maybe it may be easier to duplicate the pieces that need to be duplicated for it to work, dunno.

On a second thought, reverting the inheritance and simply copying over the format logic sounds cleaner, as that would remove the need to override the equal checks which are becoming messy. Will do that instead.

NiceAesth:
would this allow for omitting variables (including cases with multiple ones)?

`o ora ({1}) | {0} ore ({1}) | {0} de ore ({1})` would be required for languages such as romanian

smoogipoo:
As far as I can tell this PR already supports that, yes. (see `getPluralIndex()`).

PR #4919, opened 2021-11-28
Improve IPC to allow sending responses

smoogipoo:
1. Allows sending responses back to incoming requests.
2. Allows sending an IPC request with a response.
3. Extracts the port out from a const.

The goal here is to be able to accept IPC messages from osu!stable (e.g. to calculate difficulty) and send a response back.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -94,7 +94,7 @@ namespace osu.Framework.Platform
         /// </summary>
         public event Func<Exception, bool> ExceptionThrown;
 
-        public event Action<IpcMessage> MessageReceived;
+        public event Func<IpcMessage, IpcMessage> MessageReceived;
 
         /// <summary>
         /// Whether the on screen keyboard covers a portion of the game window when presented to the user.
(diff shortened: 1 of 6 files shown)


osu.Framework/Platform/GameHost.cs
```diff
@@ -111,7 +111,11 @@ public abstract class GameHost : IIpcHost, IDisposable
         /// </summary>
         protected virtual bool LimitedMemoryEnvironment => false;
 
-        protected void OnMessageReceived(IpcMessage message) => MessageReceived?.Invoke(message);
+        protected IpcMessage OnMessageReceived(IpcMessage message)
+        {
+            MessageReceived?.Invoke(message);
+            return null;
```
peppy, line 117: Is the null return intended here? Shouldn't the `MessageReceived` become a `Func` and return a potential response here?

smoogipoo: Yeah, I didn't want to make that change yet because it ripples throughout everything, but will adjust.

peppy:
The deeper you look at this class, the more jank it looks. Pretty bad..

smoogipoo:
Agree, definitely in need of a refactor.

osu.Framework/Platform/GameHost.cs
```diff
@@ -94,7 +94,7 @@ public abstract class GameHost : IIpcHost, IDisposable
         /// </summary>
         public event Func<Exception, bool> ExceptionThrown;
 
-        public event Action<IpcMessage> MessageReceived;
+        public event Func<IpcMessage, IpcMessage> MessageReceived;
```
peppy, line 97: let's xmldoc this while we're here. with the new method signature it's a bit hard to parse now.

peppy (approved):

PR #4921, opened 2021-11-29
Reduce failures in audio device re-init sequence

smoogipoo:
I added some logging in this pathway and got the following output:
```
Failed to select already-initialised audio device 1 (Init)!!
This is not the currently-selected device (-1).
Trying to get device info...
Got device info:
Name: Default, Driver: , Type: 0, Enabled: True, Initialised: False, Loopback: False
Devices:
0 | Name: No sound, Driver: , Type: 0, Enabled: True, Initialised: False, Loopback: False
1 | Name: Default, Driver: , Type: 0, Enabled: True, Initialised: False, Loopback: False
2 | Name: Soundflower (2ch), Driver: SoundflowerEngine:0, Type: 0, Enabled: True, Initialised: False, Loopback: False
3 | Name: Soundflower (64ch), Driver: SoundflowerEngine:1, Type: 0, Enabled: True, Initialised: False, Loopback: False
Retrying to init device 1...
Failed to init device (Already)!!
```
tl;dr: We try to init device 1, but it gives us `Errors.Already`. The device itself has `Initialised = false`. The device is not the current device so it can't be freed at all - `Bass.Free()` only frees the selected device, but we can't select the device without bass giving `Errors.Init`.

So if this fix doesn't work then the above will have to be reported to BASS. I'll be re-running this on CI a few times (5-10) to ensure that it works before unblocking.

### DIFF
--- a/osu.Framework/Threading/AudioThread.cs
+++ b/osu.Framework/Threading/AudioThread.cs
@@ -109,29 +109,16 @@ namespace osu.Framework.Threading
         internal static bool InitDevice(int deviceId)
         {
             Debug.Assert(ThreadSafety.IsAudioThread);
+            Trace.Assert(deviceId != -1); // The real device ID should always be used, as the -1 device has special cases which are hard to work with.
 
-            bool didInit = Bass.Init(deviceId);
-
-            // If the device was already initialised, the device can be used without much fuss.
-            if (Bass.LastError == Errors.Already)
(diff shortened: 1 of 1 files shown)


peppy (approved):

smoogipoo:
Looks fine on CI at least, did a very rudimentary in-game test and it also works there (linux only).

PR #4926, opened 2021-12-03
Apply new rider/resharper method body refactors

peppy:
These are showing up in the latest Rider beta and seem harmless enough. Make sure during review that someone checks with a non-beta release to ensure they don't show inspections there.


smoogipoo:
When you say "show up", what does that mean? Are they showing up as warnings, or are they purely informational? If they're showing up as errors, I think we should turn them off, in the same way that it's turned off for normal method bodies.

smoogipoo (approved):

peppy:
It's set to `warning` level on both framework and osu! projects, and doesn't look to stop compilation:

20211203 191651 (JetBrains Rider-EAP)

smoogipoo:
That's unrelated to this PR, right?

peppy:
Ah, wrong PR. They are showing as warnings here as well. I'm okay with turning off as well, but considering how few cases there are where we weren't already using this style it felt like leaving be was okay.

smoogipoo:
Hmm, okay, let's see how it turns out.

Susko3:
> Make sure during review that someone checks with a non-beta release to ensure they don't show inspections there.

Checked the changes on stable Rider 2021.2.2 and it looks good, no inspections.

osu.Framework.Tests/Visual/Containers/TestSceneCircularContainerSizing.cs
```diff
@@ -37,11 +37,9 @@ public class TestSceneCircularContainerSizing : FrameworkTestScene
 
             AddAssert("has correct corner radius", () => hasCorrectCornerRadius);
 
-            static void onUpdate(Container parent)
-            {
+            static void onUpdate(Container parent) =>
                 // Suppose the parent has some arbitrary size prior to the child being updated...
                 parent.Size = Vector2.One;
```
frenzibyte, line 42: I might be the only one here but I liked normal method bodies when the one-liner code has an inline comment attached on top of it, but it's not bothering me much so I think it's fine.

smoogipoo:
I was a little bit partial to the other format as well, but I'm trying not to nitpick every little thing in hopes of reducing conversation on simple PRs.

PR #4927, opened 2021-12-03
Keep BASS audio output from pausing on device processing timeout

frenzibyte:
- Closes https://github.com/ppy/osu/issues/14729
- Resolves https://github.com/ppy/osu/discussions/15886

Starting in BASS 2.4.16, BASS pauses audio output completely when the selected audio device stops processing, which can happen when a computer switches to sleep mode, for example.

There is a flag defined for this behaviour (`BASS_CONFIG_DEV_TIMEOUT`) which has now been disabled in this PR, bringing back the old behaviour which is to keep audio output running regardless of whether the selected audio device has stopped processing, resolving both of the issues mentioned at the top.

See https://www.un4seen.com/forum/?topic=19601 for discussion.

---

Also in case anyone wondering, I've got the bits of the `BASS_CONFIG_DEV_TIMEOUT` flag via the `bass.h` header included in the BASS libraries, which is `70`:
```c
#define BASS_CONFIG_DEV_TIMEOUT		70
```

### DIFF
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -367,6 +367,7 @@ namespace osu.Framework.Audio
             Bass.Configure(ManagedBass.Configuration.IncludeDefaultDevice, true);
 
             // Disable BASS_CONFIG_DEV_TIMEOUT flag to keep BASS audio output from pausing on device processing timeout.
+            // See https://www.un4seen.com/forum/?topic=19601 for more information.
             Bass.Configure((ManagedBass.Configuration)70, false);
 
             return AudioThread.InitDevice(device);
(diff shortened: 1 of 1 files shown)


bdach:
while I guess I can agree with the change (although I have not tested yet), why not try to PR this new flag upstream to managedbass/our fork? always nicer to have it there than use a random integer value jammed into source.

even with this current diff at least putting the link to the thread as an inline comment would probably be a good idea.

frenzibyte:
We're currently using the upstream version of `ManagedBass`, and I thought it's not really necessary to open a PR at their side to add a flag so that we can use it here, as I'm not sure how long it would take for them to make a release of that (and was tempted to continue that way given that we already do that for the iOS session flag and in `AudioThread.InitBass` with the new reinit flag).

I don't think it's that bad to continue this for a while until ManagedBass update at their side (there's only 3 cases of doing this so far, iirc)

Also wasn't sure it's necessary to include a thread that's only just questioning about the behaviour change, but sure?

bdach:
> Also wasn't sure it's necessary to include a thread that's only just questioning about the behaviour change, but sure?

well as is that thread seems to be the closest thing to "documentation" of the flag out there.

smoogipoo (approved):

smoogipoo:
Yep we should PR this flag + the bass reinit one to ManagedBass at some point.

Issue #4920, opened 2021-11-28
Exceptions thrown from a `Game`'s BDL do not get logged to disk sometimes


bdach:
Opening as an issue as I'm at a loss as to why this is the case. Originally reproduced with some game-side realm migration failures, but a minimal repro with `SampleGame` is as follows:

```diff
diff --git a/SampleGame/SampleGameGame.cs b/SampleGame/SampleGameGame.cs
index 3b97ebaeef..6b8c5bf0b0 100644
--- a/SampleGame/SampleGameGame.cs
+++ b/SampleGame/SampleGameGame.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using osu.Framework;
 using osu.Framework.Graphics;
 using osuTK;
@@ -17,6 +18,8 @@ public class SampleGameGame : Game
         [BackgroundDependencyLoader]
         private void load()
         {
+            throw new InvalidOperationException("test");
+
             Add(box = new Box
             {
                 Anchor = Anchor.Centre,
```

When ran either in debug or release *without debugger attached*, the `InvalidOperationException` is not logged to `runtime.log`, despite being visible in stdout.

It appears that even despite the unhandled exception callback being installed before the above code executes here:

https://github.com/ppy/osu-framework/blob/fe00cc23f6393fd8891fc1d94e3ee37041e242fc/osu.Framework/Platform/GameHost.cs#L642-L643

it does not fire for this particular thrown exception and the process is just dropped without ever hitting the log.

A "solution" is to catch explicitly, log and rethrow in `GameHost.bootstrapSceneGraph()`, but it's ugly and generally the fact that the `UnhandledException` event seemingly cannot be relied upon is pretty spooky.

A piece of key information and a possible clue is that running with debugger causes this to no longer reproduce, i.e. the handler fires correctly, which indicates there may be a race of some kind.

smoogipoo:
Hmm, I thought we had a try-catch around `bootstrapSceneGraph` at some point... Situation sounds weird.

peppy:
@bdach I know you mentioned when working on this that you tested flushing the logger, but did that end up not working?

The following patch looks to resolve the issue for me:

```diff
diff --git a/osu.Framework/Platform/GameHost.cs b/osu.Framework/Platform/GameHost.cs
[...]

PR #4928, opened 2021-12-06
Force flushing logger during unhandled/unobserved exception handling

peppy:
Fixes exceptions potentially not appearing in logs even though they occurred and caused a halt in execution.

Closes https://github.com/ppy/osu-framework/issues/4920.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -332,6 +332,11 @@ namespace osu.Framework.Platform
             AppDomain.CurrentDomain.UnhandledException -= unhandledExceptionHandler;
             TaskScheduler.UnobservedTaskException -= unobservedExceptionHandler;
 
+            // In the case of an unhandled or unobserved exception, it's feasible that the disposal flow for `GameHost` doesn't run.
+            // This can result in the exception not being logged (or being partially logged) due to the logger running asynchronously.
+            // We force flushing the logger here to ensure logging completes.
+            Logger.Flush();
+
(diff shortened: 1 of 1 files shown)


osu.Framework/Platform/GameHost.cs
```diff
@@ -332,6 +332,11 @@ private void abortExecutionFromException(object sender, Exception exception, boo
             AppDomain.CurrentDomain.UnhandledException -= unhandledExceptionHandler;
             TaskScheduler.UnobservedTaskException -= unobservedExceptionHandler;
 
+            // In the case of an unhandled or unobserved exception, it's feasible that the disposal flow for `GameHost` doesn't run.
+            // This can result in the exception not being logged (or being partially logged) due to the logger running asynchronously.
+            // We force flushing the logger here to ensure logging completes.
+            Logger.Flush();
```
smoogipoo, line 338: Is there a reason why this is not in `logException` instead? It seems like the unobserved path doesn't enter here, so the comment reads wrong.

smoogipoo (changes requested):

osu.Framework/Platform/GameHost.cs
peppy: You have a point. I moved it out based on advice in https://github.com/ppy/osu-framework/issues/4920#issuecomment-986074869, but it does seem like we want this in both cases. Will revisit.

peppy: In the unobserved case, execution is not aborted, which means this is likely fine as is. I'll update the comment.

smoogipoo (approved):

PR #4930, opened 2021-12-09
Improve exception messaging when calling `CompositeDrawable.InternalChild` on disposed drawable

peppy:
Until now, this was falling through to an exception firing on the (framework-internal) logic of nulling the `Parent` (see https://github.com/ppy/osu-framework/blob/dde9eb08ede4aa94d21590f92b7c60f85cac3947/osu.Framework/Graphics/Drawable.cs#L1481-L1482) which doesn't read too well to a framework consumer.

20211209 183808 (Safari)

In the above example ([ci failure](https://github.com/ppy/osu/runs/4453854462?check_suite_focus=true)) it reads so badly the actual cause of failure is uncertain.

### DIFF
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -267,7 +267,7 @@ namespace osu.Framework.Graphics.Containers
             try
             {
                 if (IsDisposed)
-                    throw new ObjectDisposedException(ToString(), "Disposed Drawables may not have children added.");
+                    throw new ObjectDisposedException(ToString(), "Disposed drawables may not have children added.");
 
                 child.Load(Clock, Dependencies, false);
 
@@ -346,7 +346,7 @@ namespace osu.Framework.Graphics.Containers
             set
             {
                 if (IsDisposed)
-                    throw new ObjectDisposedException(ToString(), "Disposed Drawables may not have children set.");
+                    throw new ObjectDisposedException(ToString(), "Disposed drawables may not have children set.");
 
                 ClearInternal();
                 AddInternal(value);
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -345,6 +345,9 @@ namespace osu.Framework.Graphics.Containers
             }
             set
             {
+                if (IsDisposed)
+                    throw new ObjectDisposedException(ToString(), "Disposed Drawables may not have children set.");
+
                 ClearInternal();
                 AddInternal(value);
             }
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Containers/CompositeDrawable.cs
```diff
@@ -345,6 +345,9 @@ protected internal Drawable InternalChild
             }
             set
             {
+                if (IsDisposed)
+                    throw new ObjectDisposedException(ToString(), "Disposed Drawables may not have children set.");
```
smoogipoo, line 349: Why not move this to `ClearInternal()`? Also I don't remember the reason why `Parent == null` isn't allowed :(

peppy: It can be added there as well if you want. The reason I have it here is so the message can match the usage. If you check the linked example (especially looking at the screenshot to see which line it's attached to), it's very unclear if the exception message reads something like "can't null parent" or "can't clear children" when you're actually setting one.

I guess less so for the latter, but still.

smoogipoo: Sure, in that case this also needs to be in the setter of `Children` too.

peppy: It's already in the setter for `InternalChildren` (that's where I copied this from).

smoogipoo: Where? I don't see it.

https://github.com/ppy/osu-framework/blob/21a6128555cd2786af32029eae16686bc831e570/osu.Framework/Graphics/Containers/CompositeDrawable.cs#L404-L420

peppy: wrong method, but still applies:

https://github.com/ppy/osu-framework/blob/3f720cb6830cdabb81330cb6914372d1b8186677/osu.Framework/Graphics/Containers/CompositeDrawable.cs#L547-L548


peppy: Although I guess if we're adding to `ClearInternal` then we need to preceded all internal `ClearInternal` calls, which may have been your point.

smoogipoo: I am aware of that one, but it doesn't apply. By not putting it in `ClearInternal` as I suggested, it sounds like you're wanting to completely subvert that to only show the relevant call (`InternalChild =` or `Children =`).

peppy: Check the latest commit to see if it's along the lines you are expecting.

smoogipoo (approved):

PR #4934, opened 2021-12-16
Refactor `TcpIpcProvider` to bring up to date with code standards

peppy:
The hope here is that this will fix intermittent test failures I see specifically on my local macOS install, osu!-side. I believe the root cause was the short `ipcThread.Join()` length meaning the socket potentially wasn't unbound in time.

But at the same time, this was a dated class and needed some clean-up.

Note that I'm still using a thread to do the listening loop as `TcpListener` isn't a great `async` resident, without proper support for `CancellationTokens`. Sticking to what works for now.

20211216 164718 (JetBrains Rider)

### DIFF
--- a/osu.Framework/Platform/TcpIpcProvider.cs
+++ b/osu.Framework/Platform/TcpIpcProvider.cs
@@ -66,7 +66,6 @@ namespace osu.Framework.Platform
                 };
 
                 thread.Start();
-
                 cancellationSource = new CancellationTokenSource();
                 return true;
             }
@@ -210,6 +209,7 @@ namespace osu.Framework.Platform
             if (listener != null)
             {
                 cancellationSource.Cancel();
+                thread.Join();
             }
         }
     }
--- a/osu.Framework/Platform/TcpIpcProvider.cs
+++ b/osu.Framework/Platform/TcpIpcProvider.cs
@@ -2,7 +2,6 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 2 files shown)


osu.Framework/Platform/TcpIpcProvider.cs
```diff
@@ -173,11 +187,16 @@ private async Task<IpcMessage> receive(Stream stream, CancellationToken cancella
...
+
+            Debug.Assert(typeName != null);
+
+            var type = Type.GetType(typeName);
             var value = json["Value"];
 
-            Trace.Assert(type != null);
-            Trace.Assert(value != null);
+            Debug.Assert(type != null);
+            Debug.Assert(value != null);
```
smoogipoo, line 199: These should be trace asserts because they're possible if a consumer misuses this class. Or exceptions. Or it could be argued that the messages should just be dropped altogether and the error be logged.

smoogipoo (changes requested):

osu.Framework/Platform/TcpIpcProvider.cs
peppy: Yeah, figured you'd mention this. They will still throw if they are actually null, and the error thrown is just about as good as the original `Trace.Assert` IMO. Unfortunately rider doesn't recognise `Trace.Assert` and it results null-deference warnings.

Will change to exceptions if you feel strongly.

smoogipoo: True, I'd still prefer exceptions though.

smoogipoo (approved):

PR #4936, opened 2021-12-16
Fix `AddOnce` scheduled tasks not getting updates to data from subsequent calls

peppy:
Bit of an unfortunate scenario with the constructor requiring a static delegate. Open to suggestions on how to improve this.

### DIFF
--- a/osu.Framework/Threading/ScheduledDelegateWithData.cs
+++ b/osu.Framework/Threading/ScheduledDelegateWithData.cs
@@ -5,7 +5,7 @@ using System;
 
 namespace osu.Framework.Threading
 {
-    public class ScheduledDelegateWithData<T> : ScheduledDelegate
+    internal class ScheduledDelegateWithData<T> : ScheduledDelegate
     {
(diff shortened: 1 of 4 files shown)


osu.Framework/Threading/ScheduledDelegateWithData.cs
```diff
@@ -9,10 +9,15 @@ public class ScheduledDelegateWithData<T> : ScheduledDelegate
     {
         public new readonly Action<T> Task;
 
+        public T Data;
```
smoogipoo, line 12: `internal` maybe? Feels a little unsafe. This entire class should be internal tbh.

smoogipoo (changes requested):

osu.Framework/Threading/ScheduledDelegateWithData.cs
peppy: Oops, yeah.

smoogipoo (approved):

PR #4938, opened 2021-12-16
Log loading components on test failure or game exit

peppy:
May help with figuring out the cause of intermittent failing tests. Just maybe.

- [x] Depends on #4937.

```csharp
[runtime] 2021-12-16 10:42:53 [verbose]: 🔸 Step #4 push slow
[runtime] 2021-12-16 10:42:53 [verbose]: 🔸 Step #5 push slow
[runtime] 2021-12-16 10:42:53 [verbose]: 📺 ScreenStack transition (TestScreenSlow » TestScreenSlow)
[runtime] 2021-12-16 10:42:53 [verbose]: 🔸 Step #6 push slow
[runtime] 2021-12-16 10:43:03 [verbose]: 💥 Step #7 Until: wait for screen 2 to load (23962 tries)
[runtime] 2021-12-16 10:43:03 [verbose]: ⏳ Currently loading components
[runtime] 2021-12-16 10:43:03 [verbose]: - TestSceneScreenStack+TestScreenSlow LoadState:Loading Thread:ThreadedTaskScheduler (LoadComponentsAsync (standard))
[runtime] 2021-12-16 10:43:03 [verbose]: - TestSceneScreenStack+TestScreenSlow LoadState:Loading Thread:ThreadedTaskScheduler (LoadComponentsAsync (standard))
[runtime] 2021-12-16 10:43:03 [error]: An unhandled error has occurred.
```

### DIFF
--- a/osu.Framework/Logging/LoadingComponentsLogger.cs
+++ b/osu.Framework/Logging/LoadingComponentsLogger.cs
@@ -33,13 +33,8 @@ namespace osu.Framework.Logging
             {
                 Logger.Log("⏳ Currently loading components");
 
-                lock (loading_components)
-                {
-                    foreach (var c in loading_components)
-                    {
-                        Logger.Log($"- {c.GetType().ReadableName(),-16} LoadState:{c.LoadState,-5} Thread:{c.LoadThread.Name}");
-                    }
-                }
+                foreach (var c in loading_components)
+                    Logger.Log($"- {c.GetType().ReadableName(),-16} LoadState:{c.LoadState,-5} Thread:{c.LoadThread.Name}");
 
                 loading_components.Clear();
             }
--- a/osu.Framework/Logging/LoadingComponentsLogger.cs
+++ b/osu.Framework/Logging/LoadingComponentsLogger.cs
@@ -1,0 +1,48 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Diagnostics;
+using osu.Framework.Extensions.TypeExtensions;
+using osu.Framework.Graphics;
+using osu.Framework.Lists;
+
+namespace osu.Framework.Logging
+{
+    internal static class LoadingComponentsLogger
+    {
+        private static readonly WeakList<Drawable> loading_components = new WeakList<Drawable>();
+
+        [Conditional("DEBUG")]
+        public static void Add(Drawable component)
+        {
+            lock (loading_components)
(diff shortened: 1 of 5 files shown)


osu.Framework/Logging/LoadingComponentsLogger.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        [Conditional("DEBUG")]
+        public static void LogAndFlush()
+        {
+            lock (loading_components)
+            {
+                Logger.Log("⏳ Currently loading components");
+
+                lock (loading_components)
+                {
```
smoogipoo, line 37: This is unnecessary, right?

smoogipoo (changes requested):

bdach (approved):

osu.Framework/Logging/LoadingComponentsLogger.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        [Conditional("DEBUG")]
+        public static void LogAndFlush()
+        {
+            lock (loading_components)
+            {
+                Logger.Log("⏳ Currently loading components");
+
+                foreach (var c in loading_components)
+                    Logger.Log($"- {c.GetType().ReadableName(),-16} LoadState:{c.LoadState,-5} Thread:{c.LoadThread.Name}");
```
smoogipoo, line 37: Today I learnt about this alignment syntax. Though I'm not sure the -5 is doing much here - it would have to be -9 because that's the max length.

smoogipoo (approved):

osu.Framework/Logging/LoadingComponentsLogger.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        [Conditional("DEBUG")]
+        public static void Remove(Drawable component)
+        {
+            lock (loading_components)
+                loading_components.Remove(component);
+        }
+
+        [Conditional("DEBUG")]
+        public static void LogAndFlush()
```
smoogipoo, line 30: These will never run in osu!-side CI. This refers to the environment which o!f is compiled with. You'd need to make use of `DebugUtils` otherwise.

peppy: Hmm, that's unfortunate.

Does that also mean that this method should not be publicly exposed?

https://github.com/ppy/osu-framework/blob/dde9eb08ede4aa94d21590f92b7c60f85cac3947/osu.Framework/Logging/Logger.cs#L270-L274

smoogipoo: Yeah that will have the same issue.

PR #4939, opened 2021-12-16
Add Shadow Colour and Offset to SpriteIcon

H2n9:
Added properties for `ShadowOffset` and `ShadowColour`. 
Noticed these weren't exposed when working on my own framework project.

Example
image

Default values have been set to visually match previous behaviour. 
Test steps also added to `TestSceneSpriteIcon`.

### DIFF
--- a/osu.Framework/Graphics/Sprites/SpriteIcon.cs
+++ b/osu.Framework/Graphics/Sprites/SpriteIcon.cs
@@ -132,7 +132,8 @@ namespace osu.Framework.Graphics.Sprites
             {
                 shadowColour = value;
 
-                spriteShadow.Colour = shadowColour;
+                if (spriteShadow != null)
+                    spriteShadow.Colour = shadowColour;
             }
         }
 
@@ -148,7 +149,8 @@ namespace osu.Framework.Graphics.Sprites
             {
                 shadowOffset = value;
 
-                spriteShadow.Position = shadowOffset;
+                if (spriteShadow != null)
+                    spriteShadow.Position = shadowOffset;
             }
         }
 
--- a/osu.Framework/Graphics/Sprites/SpriteIcon.cs
+++ b/osu.Framework/Graphics/Sprites/SpriteIcon.cs
@@ -48,8 +48,8 @@ namespace osu.Framework.Graphics.Sprites
                         Origin = Anchor.Centre,
                         RelativeSizeAxes = Axes.Both,
                         FillMode = FillMode.Fit,
-                        Y = 2,
-                        Colour = new Color4(0f, 0f, 0f, 0.2f),
+                        Position = shadowOffset,
+                        Colour = shadowColour
                     },
                     Alpha = shadow ? 1 : 0,
                 },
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Sprites/SpriteIcon.cs
```diff
@@ -120,6 +120,38 @@ public bool Shadow
...
+        /// The colour of the shadow displayed around the icon. A shadow will only be displayed if the <see cref="Shadow"/> property is set to true.
+        /// </summary>
+        public Color4 ShadowColour
+        {
+            get => shadowColour;
+            set
+            {
+                shadowColour = value;
+
+                spriteShadow.Colour = shadowColour;
```
bdach, line 135: This setter is unsafe to call before BDL. See following patch:

```diff
diff --git a/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteIcon.cs b/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteIcon.cs
index 0a6050b959..0e0f234ff9 100644
--- a/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteIcon.cs
+++ b/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteIcon.cs
@@ -77,6 +77,13 @@ public TestSceneSpriteIcon()
             AddStep("white background", () => background.FadeColour(Color4.White, 200));
             AddStep("move shadow offset", () => flow.Children.OfType<Icon>().ForEach(i => i.SpriteIcon.ShadowOffset += Vector2.One));
             AddStep("change shadow color", () => flow.Children.OfType<Icon>().ForEach(i => i.SpriteIcon.ShadowColour = Color4.Pink));
+            AddStep("add new icon", () => flow.Add(new SpriteIcon
+            {
+                Size = new Vector2(16),
+                Shadow = true,
+                ShadowOffset = new Vector2(5),
+                ShadowColour = Color4.Red
+            }));
         }
 
         private class Icon : Container, IHasTooltip
```

This leads to a crash on a null reference.

bdach (changes requested):

osu.Framework/Graphics/Sprites/SpriteIcon.cs
H2n9: Should be fixed now. Test case added as well.

osu.Framework/Graphics/Sprites/SpriteIcon.cs
```diff
@@ -120,6 +120,40 @@ public bool Shadow
...
+        /// </summary>
+        public Color4 ShadowColour
+        {
+            get => shadowColour;
+            set
+            {
+                shadowColour = value;
+
+                if (spriteShadow != null)
+                    spriteShadow.Colour = shadowColour;
```
bdach, line 136: The idea here is right, but this is usually done [by checking `IsLoaded` instead](https://github.com/ppy/osu-framework/blob/fa15cbee7992b7be1803042cedee233485003ee2/osu.Framework/Graphics/Containers/ModelBackedDrawable.cs#L47-L48), and having `LoadComplete()` [apply any values that were potentially set beforehand](https://github.com/ppy/osu-framework/blob/fa15cbee7992b7be1803042cedee233485003ee2/osu.Framework/Graphics/Containers/ModelBackedDrawable.cs#L91). ([Another game-side example.](https://github.com/ppy/osu/blob/7654df94f6f37b8382be7dfcb4f674e03bd35427/osu.Game.Tournament/Screens/Gameplay/Components/MatchHeader.cs))

Doing it that way is theoretically safer, because BDL executes on a separate thread rather than on the update thread, so with this implementation you could potentially have a race between the read of `spriteShadow` here and its write in BDL, which would cause a value set to be potentially silently dropped..

H2n9: Thanks for the explanation. If I am understanding correctly, would that also mean that [`bool Shadow`](https://github.com/ppy/osu-framework/blob/a5852bd26f0b4a1b21d1d9b9854e871432d4302c/osu.Framework/Graphics/Sprites/SpriteIcon.cs#L118-L119) has the same issue. Should I include a change for that aswell?

bdach: Sure, I suppose. Probably never going to be a better time than now.

bdach (approved):
Seems harmless, but I'll wait for a second opinion that this API is wanted.

peppy (approved):
This change does make me question why the offset is different from `SpriteText`, but that's out of scope of what's being done here I guess.

PR #4941, opened 2021-12-17
Add native Windows IME support

Susko3:
SDL2 text input shows some problems, mainly the 10 character limit and selection not working properly.
Existing issues on SDL's side: https://github.com/libsdl-org/SDL/issues/1879 and https://github.com/libsdl-org/SDL/issues/3385.

So I decided to write my own implementation for Windows IME handling. This is completely transparent to `TextBox` and even to `SDL2DesktopWindowTextInput`. The composition text is no longer arbitrarily limited, and selection works as expected.

It's unfortunate that we can't use `SDL_SetWindowsMessageHook` as we do in [`WindowsMouseHandler`](https://github.com/ppy/osu-framework/blob/6a2250ee584f1e38601b5cdb4b7397d27f730304/osu.Framework/Platform/Windows/WindowsMouseHandler.cs#L40) (reasoning: eb44a5c44aa43a383ef671c8b82759c14a92900e). Seems a bit strange that this was the case, but the alternative way of receiving window messages works just fine (albeit with a [minor issue](https://github.com/ppy/osu-framework/blob/06d5682369514dd5522558157b93d72bae0e312e/osu.Framework/Platform/Windows/WindowsWindow.cs#L103-L107)).

Some native IMM/IME code adapted from [chromium](https://github.com/chromium/chromium/blob/master/ui/base/ime/win/imm32_manager.cc).

---

With this in, we're one step closer to providing an in-game candidate window for composition. I've already got the native part working, but adding a candidate window in-game is a bit daunting to me.

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -584,7 +584,7 @@ namespace osu.Framework.Platform
         /// Resets internal state of the platform-native IME.
         /// This will clear its composition text and prepare it for new input.
         /// </summary>
-        public void ResetIme() => ScheduleCommand(() =>
+        public virtual void ResetIme() => ScheduleCommand(() =>
         {
             SDL.SDL_StopTextInput();
             SDL.SDL_StartTextInput();
@@ -645,11 +645,11 @@ namespace osu.Framework.Platform
                     break;
 
                 case SDL.SDL_EventType.SDL_TEXTEDITING:
-                    handleTextEditingEvent(e.edit);
+                    HandleTextEditingEvent(e.edit);
                     break;
 
                 case SDL.SDL_EventType.SDL_TEXTINPUT:
-                    handleTextInputEvent(e.text);
+                    HandleTextInputEvent(e.text);
                     break;
 
                 case SDL.SDL_EventType.SDL_KEYMAPCHANGED:
@@ -895,15 +895,15 @@ namespace osu.Framework.Platform
                 ScheduleEvent(() => MouseMoveRelative?.Invoke(new Vector2(evtMotion.xrel * Scale, evtMotion.yrel * Scale)));
         }
 
-        private unsafe void handleTextInputEvent(SDL.SDL_TextInputEvent evtText)
+        protected virtual unsafe void HandleTextInputEvent(SDL.SDL_TextInputEvent evtText)
         {
             if (!SDL2Extensions.TryGetStringFromBytePointer(evtText.text, out string text))
                 return;
 
-            ScheduleEvent(() => TextInput?.Invoke(text));
+            ScheduleEvent(() => TriggerTextInput(text));
         }
 
-        private unsafe void handleTextEditingEvent(SDL.SDL_TextEditingEvent evtEdit)
+        protected virtual unsafe void HandleTextEditingEvent(SDL.SDL_TextEditingEvent evtEdit)
         {
             if (!SDL2Extensions.TryGetStringFromBytePointer(evtEdit.text, out string text))
                 return;
@@ -912,7 +912,7 @@ namespace osu.Framework.Platform
             int start = evtEdit.start;
             int length = evtEdit.length;
 
-            ScheduleEvent(() => TextEditing?.Invoke(text, start, length));
+            ScheduleEvent(() => TriggerTextEditing(text, start, length));
         }
 
         private void handleKeyboardEvent(SDL.SDL_KeyboardEvent evtKey)
@@ -1463,10 +1463,14 @@ namespace osu.Framework.Platform
         /// </summary>
         public event Action<string> TextInput;
 
+        protected void TriggerTextInput(string text) => TextInput?.Invoke(text);
+
         /// <summary>
         /// Invoked when an IME text editing event occurs.
         /// </summary>
         public event TextEditingDelegate TextEditing;
+
+        protected void TriggerTextEditing(string text, int start, int length) => TextEditing?.Invoke(text, start, length);
 
         /// <inheritdoc cref="IWindow.KeymapChanged"/>
         public event Action KeymapChanged;
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -427,19 +427,22 @@ namespace osu.Framework.Platform
         /// </summary>
         public void Run()
         {
-            // polling via SDL_PollEvent blocks on resizes (https://stackoverflow.com/a/50858339)
             SDL.SDL_SetEventFilter(eventFilterDelegate = (_, eventPtr) =>
             {
-                // ReSharper disable once PossibleNullReferenceException
-                var e = (SDL.SDL_Event)Marshal.PtrToStructure(eventPtr, typeof(SDL.SDL_Event));
+                var e = Marshal.PtrToStructure<SDL.SDL_Event>(eventPtr);
+                OnSDLEvent?.Invoke(e);
 
+                return 1;
+            }, IntPtr.Zero);
+
+            // polling via SDL_PollEvent blocks on resizes (https://stackoverflow.com/a/50858339)
+            OnSDLEvent += e =>
+            {
                 if (e.type == SDL.SDL_EventType.SDL_WINDOWEVENT && e.window.windowEvent == SDL.SDL_WindowEventID.SDL_WINDOWEVENT_RESIZED)
                 {
                     updateWindowSize();
                 }
-
-                return 1;
-            }, IntPtr.Zero);
+            };
 
             while (Exists)
             {
@@ -1494,6 +1497,11 @@ namespace osu.Framework.Platform
         /// Invoked when the user drops a file into the window.
         /// </summary>
         public event Action<string> DragDrop;
+
+        /// <summary>
+        /// Invoked on every SDL event before it's posted to the event queue.
+        /// </summary>
+        protected event Action<SDL.SDL_Event> OnSDLEvent;
 
         #endregion
 
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -718,6 +718,10 @@ namespace osu.Framework.Platform
                 case SDL.SDL_EventType.SDL_DROPCOMPLETE:
                     handleDropEvent(e.drop);
                     break;
+
+                case SDL.SDL_EventType.SDL_SYSWMEVENT:
+                    HandleSysWMEvent(e.syswm);
+                    break;
             }
         }
 
@@ -984,6 +988,19 @@ namespace osu.Framework.Platform
                 case SDL.SDL_WindowEventID.SDL_WINDOWEVENT_CLOSE:
                     break;
             }
+        }
+
+        /// <summary>
+        /// Override if platform requires special handling of native events.
+        /// </summary>
+        /// <remarks>
+        /// Make sure to enable the events with
+        ///   <code>
+        ///     SDL.SDL_EventState(SDL.SDL_EventType.SDL_SYSWMEVENT, SDL.SDL_ENABLE);
+        ///   </code>
+        /// </remarks>
+        protected virtual void HandleSysWMEvent(SDL.SDL_SysWMEvent sysWM)
+        {
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


osu.Framework/Platform/Windows/WindowsWindow.cs
```diff
@@ -34,6 +35,101 @@ public WindowsWindow()
...
+                    break;
+
+                case Imm.WM_IME_COMPOSITION:
+                    using (var inputContext = new Imm.InputContext(hWnd))
+                    {
+                        if (Imm.TryGetImeResult(inputContext, lParam, out string resultText))
+                        {
+                            if (string.IsNullOrEmpty(resultText) && !string.IsNullOrEmpty(lastComposition))
+                            {
+                                // These events are somewhat delayed as they go trough SDL's event queue,
```
bdach, line 103:
```suggestion
                                // These events are somewhat delayed as they go through SDL's event queue,
```

bdach (dismissed):

bdach:
Generally disappointed to see more platform-specific handling, especially that it seems that it's bundled with workarounds to disable some SDL behaviour.

osu.Framework/Platform/Windows/Native/Imm.cs
```diff
@@ -0,0 +1,444 @@
...
+                    break;
+            }
+
+            targetStart = start;
+            targetEnd = end;
+            return true;
+        }
+
+        [DllImport("imm32.dll", CharSet = CharSet.Unicode)]
+        private static extern int ImmGetCompositionString(IntPtr hImc, CompositionString dwIndex, byte[] lpBuf, uint dwBufLen);
```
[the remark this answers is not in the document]
Susko3: The [Microsoft Docs](https://docs.microsoft.com/en-us/windows/win32/api/imm/nf-imm-immgetcompositionstringw) make it clear that the buffer is in bytes:
>`[in] dwBufLen`
>
>Size, in bytes, of the output buffer, even if the output is a Unicode string. The application sets this parameter to 0 if the function is to return the size of the required output buffer.

Also, this function is used for more than just unicode text. `tryGetCompositionTargetRange` uses its output for attributes, which are bytes:

https://docs.microsoft.com/en-us/windows/win32/intl/composition-string
>Attribute information is rendered in an array of 8-bit values that specifies the status of characters in the composition string. 

---

>because as is that unicode annotation seems a bit useless

It's used for picking the correct entry point: `ImmGetCompositionStringA` vs `ImmGetCompositionStringW`.
The unicode one works best for us.

osu.Framework/Platform/Windows/WindowsWindow.cs
```diff
@@ -34,6 +35,101 @@ public WindowsWindow()
...
+        }
+
+        protected override unsafe void HandleTextInputEvent(SDL.SDL_TextInputEvent evtText)
+        {
+            if (!SDL2Extensions.TryGetStringFromBytePointer(evtText.text, out string text))
+                return;
+
+            // block SDL text input if we already committed that text.
+            // keep in mind that the text provided by SDL may be shorter than our text.
+            if (lastFinalizedComposition != null && lastFinalizedComposition.StartsWith(text, StringComparison.Ordinal))
```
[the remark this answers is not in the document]
Susko3: >Why is this checking the last finalised IMM composition, rather than checking if there is any ongoing IMM composition at all?

It's unlikely that we'll get a TextInput event during composition since those only get sent by

- normal text input (doesn't happen during a composition) and
- IME finalize (only gets sent after our IMM has finalized, as SDL_SysWMEvents
are enqueued before SDL_TextInputEvents)
    - TextInput result events arrive after IMM result events so we block based on the last result. IMM composition is inactive by the time the TextInput event arrives, so we can't rely on that only.

But I'll add the check anyway, makes sense to block input in case something goes wrong at SDL's side.

>As in, shouldn't the SDL text input paths be fully dead if IMM is doing its thing?

One can't easily disable only IME events in SDL, as they are same as normal text input events.

SDL does provide a compile-time define `SDL_DISABLE_WINDOWS_IME` to disable IME code. This only affects [`SDL_windowskeyboard.c`](https://github.com/libsdl-org/SDL/blob/main/src/video/windows/SDL_windowskeyboard.c) and removes ~1400 lines of code. Safe to say that SDL is doing a lot of heavy lifting for us, initializing IME and making sure it all works when `SDL_StartTextInput()` is called.

We don't really want SDL IME disabled, unless we re-implement everything (let's not do that). So we have to make do and block SDL input manually here.

By using IMM functions, we're not blocking SDL from doing the same, in fact, we're calling the same functions twice, albeit at a different time.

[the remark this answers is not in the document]
bdach: The thing that gives me the creeps the most in the conditional here is the `StartsWith` check. First and foremost I'd like that check to not be there, and SDL events to be dropped if IMM is being manually handled by the windows codepath. Which is why I asked whether the presence of *any* IMM composition could replace this conditional.

> We don't really want SDL IME disabled, unless we re-implement everything (let's not do that)

Me saying "disabling SDL IME" was a mental shortcut for "dropping all SDL events while a manually handled IMM composition is taking place to avoid double processing". Hopefully that's consistent with what's actually being done here?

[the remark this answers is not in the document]
Susko3: >Me saying "disabling SDL IME" was a mental shortcut for "dropping all SDL events while a manually handled IMM composition is taking place to avoid double processing". Hopefully that's consistent with what's actually being done here?

Yeah, SDL IME _result_ events are dropped here. Composition events are always dropped by the empty `HandleTextEditingEvent`.

>I'd like that check to not be there

Will move the check so it's just a `bool recentImeResult`.

osu.Framework/Platform/Windows/WindowsWindow.cs
```diff
@@ -34,6 +35,101 @@ public WindowsWindow()
...
+                    {
+                        if (Imm.TryGetImeResult(inputContext, lParam, out string resultText))
+                        {
+                            if (string.IsNullOrEmpty(resultText) && !string.IsNullOrEmpty(lastComposition))
+                            {
+                                // These events are somewhat delayed as they go trough SDL's event queue,
+                                // so it's possible that the internal IME state has changed by the time we get the event,
+                                // especially so if a ongoing composition is finished and a new one is started with a single keystroke.
+                                // In case the internal IME state has changed and we get empty result text,
+                                // use the last composition as fallback.
```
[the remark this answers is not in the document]
Susko3: Internal IME state means: the text returned by `ImmGetCompositionString` for `GCS_COMPSTR` and `GCS_RESULTSTR`.
This gets lost as described in the comment.

I'll look into hooking `WndProc` earlier so we can take action immediately. Hopefully with that, IME won't change state.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -986,6 +990,19 @@ private void handleWindowEvent(SDL.SDL_WindowEvent evtWindow)
...
         }
 
+        /// <summary>
+        /// Override if platform requires special handling of native events.
+        /// </summary>
+        /// <remarks>
+        /// Make sure to enable the events with
+        ///   <code>
+        ///     SDL.SDL_EventState(SDL.SDL_EventType.SDL_SYSWMEVENT, SDL.SDL_ENABLE);
+        ///   </code>
```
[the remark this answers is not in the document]
Susko3: No longer relevant since we're using `SDL_SetEventFilter` / `OnSDLEvent` now.

osu.Framework/Platform/Windows/WindowsWindow.cs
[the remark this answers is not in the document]
Susko3: Hooking with `SDL_SetEventFilter()` seems to work really well. No surprises, as `SDL_SYSWMEVENT` is sent at the start of SDL's main [`WndProc`](https://github.com/libsdl-org/SDL/blob/ca18bf11eead1ecd58e919bc3dd63abea3b409e6/src/video/windows/SDL_windowsevents.c#L620-L637).

osu.Framework/Platform/Windows/Native/Imm.cs
```diff
@@ -14,178 +16,250 @@ namespace osu.Framework.Platform.Windows.Native
...
-            length = 0;
-
-            if (tryGetCompositionSize(inputContext, lParam, CompositionString.GCS_CURSORPOS, out int cursorPosition))
+            /// <summary>
+            /// Gets the current composition text and selection.
+            /// </summary>
+            public bool TryGetImeComposition(out string compositionText, out int start, out int length)
             {
-                start = cursorPosition;
+                if (handle.IsClosed || handle.IsInvalid)
```
bdach, line 50: Not super sure about these checks...  Can this ever realistically happen in a scenario other than "this object has been disposed"? If so, then the usual way to proceed is to throw `ObjectDisposedException` or similar rather than just soft-fail like this.

osu.Framework/Platform/Windows/WindowsWindow.cs
```diff
@@ -39,92 +39,90 @@ public override void Create()
...
+        /// <remarks>Used for blocking SDL IME results since we handle those ourselves.</remarks>
+        private bool recentImeResult;
 
         private void handleImeMessage(IntPtr hWnd, uint uMsg, long lParam)
         {
             switch (uMsg)
             {
                 case Imm.WM_IME_STARTCOMPOSITION:
-                    lastComposition = null;
+                    imeCompositionActive = false;
```
bdach, line 101: Why does a start composition event set the "active composition" flag to false?

osu.Framework/Platform/Windows/WindowsWindow.cs
```diff
@@ -39,92 +39,90 @@ public override void Create()
...
-                                // especially so if a ongoing composition is finished and a new one is started with a single keystroke.
-                                // In case the internal IME state has changed and we get empty result text,
-                                // use the last composition as fallback.
-                                resultText = lastComposition;
-                            }
-
-                            lastComposition = null;
-                            lastFinalizedComposition = resultText;
-
+                            imeCompositionActive = false;
```
bdach, line 110: Similarly, is every composition not terminated with a `WM_IME_ENDCOMPOSITION` message? In general not understanding why `imeCompositionActive` and `recentImeResult` continue to both exist, as in my mental model "composition active" should encompass everything from when the windows IME candidate window displays to when it hides.

osu.Framework/Platform/Windows/Native/Imm.cs
Susko3: From my understanding of SafeHandle, `IsClosed` can only happen when it's disposed. Since this means improper use of the class, I'll change that to a throw.

`IsInvalid` can happen if a bad `hWnd` was provided, or when `ImmGetContext` doesn't provide a valid handle for some reason. In this case, we shouldn't throw as the consumer-facing methods are `TryGet...`. Returning false is fine. We already return false for other Imm errors, such as those in `tryGetCompositionSize/String`.

bdach: Bad HWND sounds pretty fatal and I'd say it would be fine to throw on that too. Not sure about the imm function itself returning a zero/negative handle, though.

osu.Framework/Platform/Windows/WindowsWindow.cs
Susko3: `recentImeResult` is necessary, as the SDL event will arrive some indeterminate time later, most likely when composition is no longer (according to Imm).

I'll investigate how windows orders and when it sends the IME messages, and change as necessary.

osu.Framework/Platform/Windows/Native/Imm.cs
Susko3: Don't think we can differentiate between the two, Imm only returns a handle, and the docs [don't mention any error values](https://docs.microsoft.com/en-us/windows/win32/api/imm/nf-imm-immgetcontext).

osu.Framework/Platform/Windows/WindowsWindow.cs
```diff
@@ -34,6 +35,98 @@ public WindowsWindow()
...
+        public override void Create()
+        {
+            base.Create();
+
+            // enable window message events to use with `OnSDLEvent` below.
+            SDL.SDL_EventState(SDL.SDL_EventType.SDL_SYSWMEVENT, SDL.SDL_ENABLE);
+
+            OnSDLEvent += e =>
+            {
+                if (e.type != SDL.SDL_EventType.SDL_SYSWMEVENT) return;
```
peppy, line 47: Maybe split this into a named private/static method and `region` it along with all the methods below as IME specific logic, to keep things organised.

peppy (commented):
Haven't tested, but looks sane on a code pass.

bdach:
Tested this briefly today, seems to work okay as far as I can tell (although not a frequent IME user soooo...)

bdach (approved):

peppy (approved):

PR #4942, opened 2021-12-17
Add option to disallow IME to `TextBox`

Susko3:
- [x] Depends on #4941

Adds a new option `AllowIme` to `TextBox` that can be used to disallow/disable the IME when that text box has focus. (Currently, the IME will always be active/allowed.)
This is useful for situations where IME input is not wanted, such as for passwords, numbers, romanised text, etc.

Currently only works on windows.

### DIFF
--- a/osu.Framework/Platform/Windows/Native/Imm.cs
+++ b/osu.Framework/Platform/Windows/Native/Imm.cs
@@ -67,6 +67,18 @@ namespace osu.Framework.Platform.Windows.Native
         }
 
         /// <summary>
+        /// Sets whether IME is allowed.
+        /// </summary>
+        /// <remarks>
+        /// If IME is disallowed, text input is active and a language that uses IME is selected,
+        /// then IME will be forced into alphanumeric mode, behaving as normal keyboard input (no compositing will happen).
+        /// </remarks>
+        internal static void SetImeAllowed(IntPtr hWnd, bool allowed)
+        {
+            ImmAssociateContextEx(hWnd, IntPtr.Zero, allowed ? AssociationFlags.IACE_DEFAULT : AssociationFlags.IACE_IGNORENOCONTEXT);
+        }
+
+        /// <summary>
         /// Returns true if <paramref name="lParam"/> has the specified <paramref name="flag"/>.
         /// </summary>
         private static bool hasFlag(this long lParam, CompositionString flag) => (lParam & (long)flag) != 0;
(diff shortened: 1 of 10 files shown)


osu.Framework/Platform/Windows/Native/Imm.cs
```diff
@@ -484,6 +500,28 @@ private enum CompositionStringAction : uint
             CPS_CANCEL = 0x0004,
         }
 
+        /// <summary>
+        /// Flags specifying the type of association between the window and the input method context.
+        /// dwFlags for <see cref="ImmAssociateContextEx"/>.
+        /// </summary>
+        private enum AssociationFlags : uint
```
bdach, line 507: looks like this should technically be annotated with `[Flags]`?

osu.Framework/Platform/Windows/Native/Imm.cs
```diff
@@ -27,6 +27,18 @@ internal static void CancelComposition(IntPtr hWnd)
...
+        /// <summary>
+        /// Sets whether IME is allowed.
+        /// </summary>
+        /// <remarks>
+        /// If IME is disallowed, text input is active and a language that uses IME is selected,
+        /// then IME will be forced into alphanumeric mode, behaving as normal keyboard input (no compositing will happen).
+        /// </remarks>
+        internal static void SetImeAllowed(IntPtr hWnd, bool allowed)
+        {
+            ImmAssociateContextEx(hWnd, IntPtr.Zero, allowed ? AssociationFlags.IACE_DEFAULT : AssociationFlags.IACE_IGNORENOCONTEXT);
```
bdach, line 39: do you have any reference for knowing to pass a zero `hInc` handle, or what that means? i tried to find some but docs of imm.h are shocking.

Susko3: Used in chromium here:

https://github.com/chromium/chromium/blob/fd8a8914ca0183f0add65ae55f04e287543c7d4a/ui/base/ime/win/imm32_manager.cc#L397-L405

Though I don't know what difference is between `0` and `IACE_IGNORENOCONTEXT`.

osu.Framework/Platform/Windows/Native/Imm.cs
Susko3: Although the values do look to be flags, the wording on the docs makes it clear that only one flag should be used:

https://docs.microsoft.com/en-us/windows/win32/api/imm/nf-imm-immassociatecontextex
>Flags specifying the type of association between the window and the input method context. This parameter can have one of the following values.

bdach: I don't really perceive that to be clear but whatever.

bdach (approved):

smoogipoo (approved):
Lgtm

PR #4951, opened 2021-12-21
Adjust Android lifecycle code in line with context preservation changes

bdach:
- [x] Depends on https://github.com/ppy/osuTK/pull/74

* Removes the "kill on pause" behaviour which was in place due to the context loss.
* Ensures that the game host is not created multiple times on resuming game.
* Suspends all threads on activity pause and resumes them again on activity resume.
* Additionally I've had to manually pause/start all audio output on pause/resume via BASS as unfortunately pausing the audio thread itself did not appear to be sufficient (the audio would continue playing even after the app was paused/backgrounded). Not sure how this isn't an issue on iOS.

Tested on two Android devices that I own:

* Samsung Galaxy J7 2016 / SM-J710F (Android 7),
* OnePlus 8T / KB2003 (Android 11)

### DIFF
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -108,7 +108,8 @@ namespace osu.Framework.Android
         {
             base.OnLoad(e);
 
-            RenderGame();
+            if (Host == null)
+                RenderGame();
         }
 
         [STAThread]
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -108,6 +108,8 @@ namespace osu.Framework.Android
         {
             base.OnLoad(e);
 
+            // osuTK calls `OnLoad()` every time the application surface is created, which will also happen upon a resume,
+            // at which point the host is already present and running, so there is no reason to create another one.
             if (Host == null)
                 RenderGame();
         }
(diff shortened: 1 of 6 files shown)


smoogipoo (dismissed):

peppy:
> * Additionally I've had to manually pause/start all audio output on pause/resume via BASS as unfortunately pausing the audio thread itself did not appear to be sufficient (the audio would continue playing even after the app was paused/backgrounded). Not sure how this isn't an issue on iOS.

iOS is quite robust in that an application can't break / interfere with the overall device state, so it will force audio to stop playing when an app goes into the background. An app needs to request background audio support (which shows the user that an app is using this) in order to play in the background.

peppy:
Tested to work on my test device. Let's get this in for further user testing.

osu.Framework.Android/AndroidGameView.cs
```diff
@@ -108,7 +108,8 @@ protected override void OnLoad(EventArgs e)
         {
             base.OnLoad(e);
 
-            RenderGame();
+            if (Host == null)
```
peppy, line 111: I think this probably deserves an inline comment explaining why it only happens when `null`.

peppy (changes requested):

osu.Framework.Android/AndroidGameView.cs
bdach: I've transplanted the relevant commit message into an inline comment here.

peppy (approved):

Issue #4950, opened 2021-12-21
`InputConfigManager` only saves on exit


Susko3:
Noticed this while trying to make a fix for https://github.com/ppy/osu/issues/8264.
Wanted to add a new setting to `AndroidKeyboardHandler` but then realized that `input.json` is never created on Android.

`InputConfigManager` saves only once, and that is when [`GameHost` is disposed](https://github.com/ppy/osu-framework/blob/05e313b78fc62fed75b494f7c7dd84fc8e734315/osu.Framework/Platform/GameHost.cs#L1094). On Android, the host never exits cleanly, so this never gets called.

It should probably save when any setting for an `InputHandler` is changed. Not really sure what's the best way to achieve this.

---

For the Android fix, I can just add a new `FrameworkSetting`. But for a platform-specific fix, a toggle in `AndroidKeyboardHandler` and saving it to `input.json` is a much cleaner solution.

bdach:
The host never exiting cleanly on android may change sooner than expected. In fact I was just queueing up changes that may address this.

Susko3:
Looking at the osu!-side settings panel, it's borderline impossible to get non-desktop input handlers to have their own panel since they live in different projects.

So for simplicity, I'll save the value to config for now.

peppy:
This is a bit of a hard one to resolve. I can agree that the configuration should be updated more often (even on desktop platforms), similar to the `FrameworkConfigManager` which will queue a background save on any change. Unfortunately for `InputConfigManager` we don't have any visibility into the bindables / changes which are happening.

I guess we could do reflection on each of the `InputHandler`s in the list and bind to their changes? Would need more opinions on this because it's a bit ugly.

bdach:
An alternative would be to force all modifications to go through a `PerformWrite`-esque method that takes a `Func<InputHandler>` or similar that does the changes, and after that func is done making the changes the config manager would have a chance to save. Not sure which one is better at this point.

peppy:
I think I'd probably prefer the reflection route at this point in time.

PR #4952, opened 2021-12-21
Add setting to apply fix for Android keyboard not working for some users

Susko3:
Decided the to go with a new setting instead of a bindable in `AndroidKeyboardHandler` because of https://github.com/ppy/osu-framework/issues/4950.

Should hopefully resolve https://github.com/ppy/osu/issues/8264.

I can't confirm that this will fix it. None of my tested phones exhibited the above issue.
On a device that doesn't have this issue, enabling the setting doesn't break input.

We're seeing the same keyboard problems as those mentioned in the comments [here (https://github.com/termux)](https://github.com/termux/termux-app/blob/af16e79bf8e422811885f435c3aaf74d619dd7e6/terminal-view/src/main/java/com/termux/view/TerminalView.java#L265-L283).
I've applied the same fix that is used there.

Snippet from https://wiki.termux.com/wiki/Terminal_Settings#Workaround_for_some_keyboards_issues:

> **Workaround for some keyboards issues**
> 
> Some keyboards, e.g. a default one on Samsung devices, have issues with text input: text appearing only after pressing "enter", wrong keyboard layout being opened, etc. They are not compatible with the Termux input method because enforcing a **word-based input.**
> 
> As workaround, you can set the following option: 
> 
> [...]

Word-based input is exactly what we're seeing the the video of the issue. Keyboards forcing suggestions and word input.

I'm optimistic about this fix, but we can't know until it reaches end-users.

### DIFF
--- a/osu.Framework.Android/AndroidGameHost.cs
+++ b/osu.Framework.Android/AndroidGameHost.cs
@@ -30,6 +30,11 @@ namespace osu.Framework.Android
             this.gameView = gameView;
         }
 
+        /// <summary>
+        /// Allows <see cref="AndroidGameView"/> to access the config.
+        /// </summary>
+        internal new FrameworkConfigManager Config => base.Config;
+
         protected override void SetupConfig(IDictionary<FrameworkSetting, object> defaultOverrides)
         {
             if (!defaultOverrides.ContainsKey(FrameworkSetting.ExecutionMode))
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -9,6 +9,8 @@ using Android.Util;
 using Android.Views;
 using Android.Views.InputMethods;
 using osu.Framework.Android.Input;
+using osu.Framework.Bindables;
+using osu.Framework.Configuration;
 using osuTK.Graphics;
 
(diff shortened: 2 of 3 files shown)


peppy:
I’d rather not have a setting if we can avoid it.

bdach:
Definitely agree with the above to not make this a setting if it can be avoided. I'll test on my two devices today with the setting unconditionally enabled.

Susko3:
According to https://github.com/termux/termux-app/issues/686#issuecomment-570175715 forcing this setting on may break some keyboards.

Termux is a mature Android app, and I trust that they've come to the best compromise.

But something we can do that Termux can't is provide support for word-based input. We have a fully fledged `TextBox`, and if we feed enough info back into `AndroidInputConnection` we can support rich text editing.

peppy:
Not sure. That thread has a lot of noise so it's hard to say how to move forward. But I do disagree with having game settings which are required to make the game work (something as simple as input), so let's push this out without the setting and see if people have regressing problems, then deal with that separately.

peppy:
Tested to work fine on a Samsung SM-N9208 (API 24).

Susko3:
Pushed the change as requested, we'll see how it goes.

peppy (approved):

PR #4953, opened 2021-12-22
Add missing xmldoc to `InputConfigManager`

peppy:
Just noticed while investigating https://github.com/ppy/osu-framework/issues/4950.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -991,8 +991,7 @@ namespace osu.Framework.Platform
                 threadRunner.SetCulture(culture);
             }, true);
 
-            // intentionally done after everything above to ensure the new configuration location has priority over obsoleted values.
-            Dependencies.Cache(InputConfig = new InputConfigManager(Storage, AvailableInputHandlers));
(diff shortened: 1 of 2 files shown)


osu.Framework/Platform/GameHost.cs
```diff
@@ -991,8 +991,7 @@ protected virtual void SetupConfig(IDictionary<FrameworkSetting, object> default
                 threadRunner.SetCulture(culture);
             }, true);
 
-            // intentionally done after everything above to ensure the new configuration location has priority over obsoleted values.
-            Dependencies.Cache(InputConfig = new InputConfigManager(Storage, AvailableInputHandlers));
```
bdach, line 995: Was this cache removal intentional? Doesn't seem to be judging from the title of the pull.

peppy: Yeah, see first commit. It's not used anywhere, nor could it be used in any useful way.

peppy: I've also made the class `internal` for now with the rationale that (in its current state, at least) it's not useful (and could only really be misused).

Open to rolling this back if we want to leave it there for future use cases that may happen? Dunno.

smoogipoo (approved):

PR #4954, opened 2021-12-22
Add `CancellationToken` support to `IResourceStore.GetAsync`

peppy:
Noticed that a lot of avatars get stuck in a loading state, even though they are likely not required for loading purposes (they have since been disposed or forgotten).

Having cancellation token support should allow for improving such cases, cancelling the underlying web request if it is already in progress.

Note that this will require usages to be updated to pass the `CancellationToken` from DI. And maybe `async` support for BDL. But it's a start and doesn't look to break anything.


smoogipoo (approved):
It's a start, not going to do much for the time being though.

PR #4956, opened 2021-12-23
Fix multiple test failures in screen stack tests

peppy:
See individual commits for more details. Standard async/schedule oversight failures, "unfortunately".

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneScreenStack.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneScreenStack.cs
@@ -234,7 +234,6 @@ namespace osu.Framework.Tests.Visual.UserInterface
             AddUntilStep("base became current again", () => baseScreen.IsCurrentScreen());
 
             AddAssert("screen1 not current", () => !screen1.IsCurrentScreen());
-            AddAssert("screen1 exited", () => screen1.ExitedTo == baseScreen);
             AddAssert("screen1 was not added to hierarchy", () => !screen1.IsLoaded);
 
             AddStep("push fast", () => baseScreen.Push(screen2 = new TestScreen()));
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneScreenStack.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneScreenStack.cs
@@ -226,9 +226,15 @@ namespace osu.Framework.Tests.Visual.UserInterface
             AddStep("push slow", () => baseScreen.Push(screen1 = new TestScreenSlow()));
             AddStep("exit slow", () => screen1.Exit());
             AddStep("allow load", () => screen1.AllowLoad.Set());
+            AddAssert("ensure base not current", () => baseScreen.IsCurrentScreen());
             AddUntilStep("wait for screen to load", () => screen1.LoadState >= LoadState.Ready);
-            AddAssert("ensure not current", () => !screen1.IsCurrentScreen());
-            AddAssert("ensure base still current", () => baseScreen.IsCurrentScreen());
+
+            // Intentionally before the until step as it may catch a very brief incorrect state.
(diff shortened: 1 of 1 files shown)


osu.Framework.Tests/Visual/UserInterface/TestSceneScreenStack.cs
```diff
@@ -226,9 +226,15 @@ public void TestExitBeforePush()
...
+            AddAssert("ensure base not current", () => baseScreen.IsCurrentScreen());
             AddUntilStep("wait for screen to load", () => screen1.LoadState >= LoadState.Ready);
-            AddAssert("ensure not current", () => !screen1.IsCurrentScreen());
-            AddAssert("ensure base still current", () => baseScreen.IsCurrentScreen());
+
+            // Intentionally before the until step as it may catch a very brief incorrect state.
+            AddAssert("ensure slow not current", () => !screen1.IsCurrentScreen());
+
+            AddUntilStep("base became current again", () => baseScreen.IsCurrentScreen());
+            AddAssert("ensure slow not current", () => !screen1.IsCurrentScreen());
```
smoogipoo, line 236: Maybe it's just me, but I can't read these assertions. Isn't the proper assertion here that `screen1` got exited and is not `LoadComplete`?

So what I mean is:
```
AddStep("exit slow", () => screen1.Exit());
AddAssert("screen1 is still current screen", () => screen1.IsCurrentScreen());

AddStep("allow load", () => screen1.AllowLoad.Set());
AddUntilStep("wait for screen to load", () => screen1.LoadState >= LoadState.Ready);
AddUntilStep("base became current again", () => baseScreen.IsCurrentScreen());

AddAssert("screen1 exited", () => screen1.ExitedTo == baseScreen);
AddAssert("screen1 was not added to hierarchy", () => !screen1.LoadComplete);
```

smoogipoo (changes requested):

osu.Framework.Tests/Visual/UserInterface/TestSceneScreenStack.cs
```diff
@@ -226,9 +226,15 @@ public void TestExitBeforePush()
             AddStep("push slow", () => baseScreen.Push(screen1 = new TestScreenSlow()));
             AddStep("exit slow", () => screen1.Exit());
             AddStep("allow load", () => screen1.AllowLoad.Set());
+            AddAssert("ensure base not current", () => baseScreen.IsCurrentScreen());
```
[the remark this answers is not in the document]
peppy: I actually meant to be testing the description part, but turns out that's not correct for this test scene.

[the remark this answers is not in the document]
peppy: was supposed to be one line higher than it is.

osu.Framework.Tests/Visual/UserInterface/TestSceneScreenStack.cs
peppy: Would you be okay with also leaving the `IsCurrent` checks? They are providing coverage for that.

smoogipoo: The `// Intentionally` one's weird to me because it's testing internal implementation details rather than the end-user expectation.

The other two are redundant with one another and I can only gather that they're also testing implementation details:
image
Like... Assuming that "maybe screen1 will become current one frame later"?

peppy: yeah, the intention is to test implementation details. can remove if you think it's not worthwhile.

peppy: I've adjusted this to hopefully read better.

smoogipoo (approved):

PR #4962, opened 2021-12-24
Fix `RearrangeableListContainer` not disposing removed items

peppy:


bdach:
Relevant test failures here.

peppy:
I guess we're using the queue instead..

osu.Framework/Graphics/Containers/RearrangeableListContainer.cs
```diff
@@ -110,7 +111,11 @@ private void removeItems(IList items)
                 if (currentlyDraggedItem != null && EqualityComparer<TModel>.Default.Equals(currentlyDraggedItem.Model, item))
                     currentlyDraggedItem = null;
 
-                ListContainer.Remove(itemMap[item]);
+                var drawableItem = itemMap[item];
+
+                ListContainer.Remove(drawableItem);
+                AsyncDisposalQueue.Enqueue(drawableItem);
```
bdach, line 117: Out of curiosity I've looked at who else calls this to check if there are any other existing scenarios in framework, and stumbled upon `CompositeDrawable.DisposeChildAsync()` which also enqueues for async disposal, but also calls `UnbindAllBindablesSubTree()` on the disposed drawable. Therefore I think that would be an even better thing to call here?

```suggestion
                DisposeChildAsync(drawableItem);
```

bdach (approved):

PR #4966, opened 2021-12-24
Add explicit `Exported = true` to `AndroidGameActivity`

Susko3:
As discussed in https://github.com/ppy/osu/pull/16240#discussion_r775046302.


bdach:
Well to be fair, none of _these_ activities really _need_ to be exported, they don't have any intent filters. The `AndroidGameActivity` one makes the most sense as it's supposed to be an illustrative example. But I suppose there's no harm in specifying `true` in all of them.

Susko3:
Required as the activities are `MainLauncher = true`. Meaning they have the intent-filter for `android.intent.category.LAUNCHER`.

Susko3:
CI isn't liking this. Maybe best to block until runners are updated.

bdach:
> Required as the activities are `MainLauncher = true`. Meaning they have the intent-filter for `android.intent.category.LAUNCHER`.

Ok, fair point. Wasn't aware of this.

> CI isn't liking this. Maybe best to block until runners are updated.

Right, I never bothered to downgrade xamarin framework-side because it wasn't failing. I still think it's the safest thing to do but I don't really care enough to be touching the workflow specially just for this, would rather wait for the fix to come from upstream.

bdach (approved):

bdach:
Have revived and merged this because CI runner updates have unblocked it.

PR #4967, opened 2021-12-25
Fix `LoadComponentAsync` running load continuation asynchronously

peppy:
There is no reason for this to be run asynchronously. It's a lean continuation and has no user code. This should improve performance of async load tasks as there is one less thread transition.

It also means if the TPL thread pool is saturated, drawables will still load. This may in turn cause unblocking of other threads in the thread pool (due to bad user implementations in drawable load code, for instance) and avoid potential deadlocks.

This should 100% resolve the `TestMultiplayer` failures where screens are stuck in a `Ready` but not `Loaded` state.

It does *not* fix the actual TPL saturation in said case, which is likely due to bad `async` implementation (ie. a `.Wait()` waiting on a blocking `Task.Run`). Haven't pinpointed yet but am going through all usages systematically. `TestMultiplayerClient` seems likely, and is written in a way that's very hard to track `async` usages (probably should not inherit `MultiplayerClient`) but it could be something else.

### DIFF
--- a/osu.Framework.Tests/Containers/TestSceneLoadComponentAsync.cs
+++ b/osu.Framework.Tests/Containers/TestSceneLoadComponentAsync.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using NUnit.Framework;
 using osu.Framework.Allocation;
 using osu.Framework.Graphics;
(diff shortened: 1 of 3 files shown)


bdach:
Seems fair. From reading alone this seemed sane but I've also done some _very_ crude measurements of how long the continuation takes just for confirmation's sake and never got a time that's higher than half a millisecond, and that was all (1) on debug and (2) with weird logging added so I wouldn't be surprised to see actual times being an order of magnitude faster than that still.

bdach (approved):

smoogipoo (approved):
I guess we can go with this but it really should be unnecessary.

PR #4969, opened 2021-12-26
Reduce delegate allocation overheads

peppy:
Tested at osu! song select while scrolling. I've only addressed the cases which are excessive for now.

Before:
20211226 154033 (Parallels Desktop app)

After:
20211226 152913 (Parallels Desktop app)

Main targets are `DelayedLoadWrapper` and mouse movement.

Note that `DelayedLoadWrapper` still allocates quite a few `ScheduledDelegateWithData`, but this is harder to avoid. Still an improvement.

### DIFF
--- a/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
+++ b/osu.Framework/Graphics/Containers/DelayedLoadWrapper.cs
@@ -202,7 +202,7 @@ namespace osu.Framework.Graphics.Containers
             // The scheduled delegate will be cancelled if this wrapper has its UpdateSubTreeMasking() invoked, as more accurate intersections can be computed there instead.
             if (isIntersectingResetDelegate == null)
             {
-                isIntersectingResetDelegate = Game?.Scheduler.AddDelayed(() => IsIntersecting = false, 0);
+                isIntersectingResetDelegate = Game?.Scheduler.AddDelayed(wrapper => wrapper.IsIntersecting = false, this, 0);
                 result = true;
             }
 
--- a/osu.Framework/Graphics/Pooling/PoolableDrawable.cs
+++ b/osu.Framework/Graphics/Pooling/PoolableDrawable.cs
@@ -109,11 +109,13 @@ namespace osu.Framework.Graphics.Pooling
             // prepare call is scheduled as it may contain user code dependent on the clock being updated.
             // must use Scheduler.Add, not Schedule as we may have the wrong clock at this point in load.
             scheduledPrepare?.Cancel();
-            scheduledPrepare = Scheduler.Add(() =>
+            scheduledPrepare = Scheduler.Add(prepare, this);
+
+            void prepare(PoolableDrawable drawable)
             {
-                PrepareForUse();
-                waitingForPrepare = false;
-            });
+                drawable.PrepareForUse();
+                drawable.waitingForPrepare = false;
+            }
         }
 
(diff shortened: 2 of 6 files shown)


Susko3:
Looking at `SDL2DesktopWindow`, is scheduling all events even necessary?

https://github.com/ppy/osu-framework/blob/f9ae8d775ba2d03196c3ec80228b009c3dc85ab0/osu.Framework/Platform/SDL2DesktopWindow.cs#L454-L459

SDL events are handled by `handleSDLEvent()` in `pollSDLEvents()`. All of this is run on the main thread, so it short-circuits and executes the "scheduled" events immediately.

The only place where the scheduler is actually used is [here](https://github.com/ppy/osu-framework/blob/f9ae8d775ba2d03196c3ec80228b009c3dc85ab0/osu.Framework/Platform/SDL2DesktopWindow.cs#L486), where the event is force scheduled.

In your `After` comparision, I'm still seeing some `handle...Event` allocations. Could those be removed?

peppy:
Yep they can probably all be removed, but they are quite rare and we're out of scope for this change.

Removing the scheduler altogether would be optimal if it doesn't increase code complexity. That could be done as a follow up.

bdach:
Well I mean there is one usage where it's apparently needed as already stated, so I don't know about removing entirely.

bdach:
Code seems fine, but just to confirm my understanding - the savings here are from potentially leaner closures, i.e. better control over which data is stored to the scheduler? It's not super explicit to me from the code/screenshots alone.

Like, to take the `SDL2DesktopWindow` as an example - calls like [this](https://github.com/ppy/osu-framework/blob/98ccf029471f7f7d5307be2988732ad532554216/osu.Framework/Platform/SDL2DesktopWindow.cs#L576) will still incur a delegate alloc, right? Because as far as I remember references to delegates via method groups still alloc every time. I could imagine the size decreasing because the closure is potentially smaller (just the reference to the window, rather than a few vars), but I'm still not sure that is the case.

bdach:
Never mind, I think the other PR makes this clearer. The savings are from eliminating the dependency of the `Action` on the data, which internally makes it so that an instance of the `Action<T>` can be cached and reused. And in the inline execution case there's no allocation of the "wrapper" structure that pairs the `Action<T>` with its data.

See following [sharplab](https://sharplab.io/#v2:D4AQTADAsAUOCMtYgKwB4CWA7ALgPgAIBDAExPgApsdiAnAcwEoCBeQi5tu+ggagIDMAbmTpqAGgLVCpEmA6sZDRdz6CRMZPACcFWZRSMAdAEksANwD2AawCmHIwBVLAZRy1s9Dow0gdesnljMys7CkMnV3dPbyEgA==) for a simpler example of this.

bdach (approved):

peppy:
Yep, it's basically making the closure functions "static".

Although interestingly, this allocates still:

```csharp
schedule(staticFunc, data);

static void staticFunc(obj data) {}
```

and i have no idea why it does. you *have* to use the inline no-capture delegate to avoid the allocation. maybe something to test in sharplab.

bdach:
I *think* that's just what I mentioned earlier about `Func<>`s created via method groups always allocating? I.e. see https://github.com/dotnet/roslyn/issues/39869.

That's actually required by the language spec as is.

smoogipoo (approved):

Issue #4373, opened 2021-04-17
Android built-in cursor displayed on top of the game's own cursor


ravCraft:
Describe the bug:
I'm playing osu Lazer with keyboard and mouse on android. But there's a problem at the cursor ( Android cursor won't dissapear, The cursor will overlap with the game default cursor)
Screenshots or videos showing encountered issue:
![Screenshot_20210417_123341_25a1a32208bbcdc1d450b7aa854bb161](https://user-images.githubusercontent.com/78950119/115103448-1ad72300-9f7c-11eb-9c3d-11fd3032a802.jpg)
![Screenshot_20210417_123423_25a1a32208bbcdc1d450b7aa854bb161](https://user-images.githubusercontent.com/78950119/115103451-24f92180-9f7c-11eb-96cc-a5e2b1ee699f.jpg)
And also, i can't drag with my cursor and game cursor, both cursor are useless.
Disable mouse button in gameplay aren't working
osu!Lazer version:
2021.416.0
Logs:
[runtime.log](https://github.com/ppy/osu/files/6328702/runtime.log)

peppy:
Can you confirm

- what device you are using
- what input method you are using
- whether the cursor disappears in other apps or games

ravCraft:
well it was too late, but just telling you.
Device I'm using : Vivo Y20
Input Method : Keyboard and mouse ( using otg, well you maybe know about that)
Cursor on other app : Minecraft ( my cursor not duplicates on minecraft, just like playing on pc)

Once agai sorry, whatever u reading this or not, and my english was bad

sr229:
AFAIK there's no proper way to hide the cursor in Android, there isn't any API for it. I guess this may be a `wontfix` due to the lack of APIs to hide the cursor. It doesn't work like generic Linux where it acknowledges that it needs to hide the cursor when the game tells it to, it'll draw the cursor no matter what.

frenzibyte:
The fact that the other game they tested hides the cursor means there *may* be some sort of way for this, but I'm not sure whether that statement is correct nor I can confirm it at the moment.

LeNitrous:
Tested on Redmi 5A. Games such as Minecraft and Unity games like Arknights and Azur Lane show the pointer is present.

![Screenshot_20210510-144423_Minecraft](https://user-images.githubusercontent.com/20495991/117617599-5a6de500-b19f-11eb-80b7-07a577901027.png)

[...]

PR #4971, opened 2021-12-27
Improve Android mouse input and restructure Android input handlers

Susko3:
This brings us one step closer to resolving https://github.com/ppy/osu-framework/issues/4373.

In addition to separating mouse from touch events, `AndroidMouseHandler` enables the following:
- support for all mouse buttons (including forward and back)
- support for scrolling
- starting point for implementing [pointer capture](https://github.com/ppy/osu-framework/issues/4373#issuecomment-836278499).

---

Tested on Realme 6, Android 11 (API 30) with a bluetooth mouse.
On-screen keyboard input works as expected.

Test works with:
- [ ] touchpad (not required, but would be nice to test)

Test existing behaviour isn't broken for:
- [ ] physical keyboard
- [x] stylus

### DIFF
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -79,107 +79,23 @@ namespace osu.Framework.Android
                 case Keycode.VolumeUp:
                 case Keycode.VolumeMute:
                     return false;
-            }
 
-            // some implementations might send Mouse1 and Mouse2 as keyboard keycodes, so forward those only to the mouse event.
-            switch (e.Source)
-            {
-                case InputSourceType.Keyboard:
+                default:
                     KeyDown?.Invoke(keyCode, e);
                     return true;
-
-                case InputSourceType.Mouse:
-                case InputSourceType.Touchpad:
-                    MouseKeyDown?.Invoke(keyCode, e);
-                    return true;
             }
-
-            return base.OnKeyDown(keyCode, e);
         }
 
         public override bool OnKeyLongPress([GeneratedEnum] Keycode keyCode, KeyEvent e)
         {
-            switch (e.Source)
-            {
-                case InputSourceType.Keyboard:
-                    KeyLongPress?.Invoke(keyCode, e);
-                    return true;
-
-                case InputSourceType.Mouse:
-                case InputSourceType.Touchpad:
-                    MouseKeyLongPress?.Invoke(keyCode, e);
-                    return true;
-            }
-
-            return base.OnKeyLongPress(keyCode, e);
+            KeyLongPress?.Invoke(keyCode, e);
+            return true;
         }
 
         public override bool OnKeyUp([GeneratedEnum] Keycode keyCode, KeyEvent e)
         {
-            switch (e.Source)
-            {
-                case InputSourceType.Keyboard:
-                    KeyUp?.Invoke(keyCode, e);
-                    return true;
-
-                case InputSourceType.Mouse:
-                case InputSourceType.Touchpad:
-                    MouseKeyUp?.Invoke(keyCode, e);
-                    return true;
-            }
-
-            return base.OnKeyUp(keyCode, e);
-        }
-
-        public override bool OnHoverEvent(MotionEvent e)
-        {
-            switch (e.Source)
-            {
-                case InputSourceType.BluetoothStylus:
-                case InputSourceType.Stylus:
-                case InputSourceType.Touchscreen:
-                    Hover?.Invoke(e);
-                    return true;
-
-                case InputSourceType.Mouse:
-                case InputSourceType.Touchpad:
-                    MouseHover?.Invoke(e);
-                    return true;
-            }
-
-            return base.OnHoverEvent(e);
-        }
-
-        public override bool OnTouchEvent(MotionEvent e)
-        {
-            switch (e.Source)
-            {
-                case InputSourceType.BluetoothStylus:
-                case InputSourceType.Stylus:
-                case InputSourceType.Touchscreen:
-                    Touch?.Invoke(e);
-                    return true;
-
-                case InputSourceType.Mouse:
-                case InputSourceType.Touchpad:
-                    MouseTouch?.Invoke(e);
-                    return true;
-            }
-
-            return base.OnTouchEvent(e);
-        }
-
-        public override bool OnGenericMotionEvent(MotionEvent e)
-        {
-            switch (e.Source)
-            {
-                case InputSourceType.Mouse:
-                case InputSourceType.Touchpad:
-                    MouseGenericMotion?.Invoke(e);
-                    return true;
-            }
-
-            return base.OnGenericMotionEvent(e);
+            KeyUp?.Invoke(keyCode, e);
+            return true;
         }
 
         protected override void OnLoad(EventArgs e)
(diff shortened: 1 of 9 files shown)


bdach (commented):
seemingly correct but i have no actual means to test

osu.Framework.Android/Input/AndroidMouseHandler.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            if (!base.Initialize(host))
+                return false;
+
+            Enabled.BindValueChanged(enabled =>
+            {
+                if (enabled.NewValue)
+                {
+                    view.MouseKeyDown += onKeyDown;
+                    view.MouseKeyLongPress += onKeyDown;
```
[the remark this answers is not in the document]
Susko3: It should be fine, as the button state won't change if a press is queued/applied and that button is already pressed:

https://github.com/ppy/osu-framework/blob/cf3687ed8bbeec6122110d71988f39dd07878038/osu.Framework/Input/States/ButtonStates.cs#L34-L43

But yeah, `MouseKeyLongPress` is unnecessary, as it will never change state, and probably will never get sent by Android. (From my testing, it's always a normal key down, never a long press.)

osu.Framework.Android/AndroidGameView.cs
```diff
@@ -141,5 +219,94 @@ public override IInputConnection OnCreateInputConnection(EditorInfo outAttrs)
...
+        /// Invoked if the source is <see cref="InputSourceType.BluetoothStylus"/>, <see cref="InputSourceType.Stylus"/>
+        /// or <see cref="InputSourceType.Touchscreen"/>.
+        /// </remarks>
+        public new event Action<MotionEvent> Touch;
+
+        /// <summary>
+        /// Invoked on a key down event sourced from a mouse-type device.
+        /// </summary>
+        /// <remarks>Invoked if the source is <see cref="InputSourceType.Mouse"/> or <see cref="InputSourceType.Touchpad"/>.</remarks>
+        public event Action<Keycode, KeyEvent> MouseKeyDown;
```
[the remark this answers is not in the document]
Susko3: How about having a input handler parent to all android input handlers. Skeleton of the relevant bits:

```cs
public abstract class AndroidInputHandler : InputHandler
{
    protected abstract IEnumerable<InputSourceType> AllowedInputSources { get; }

    // since InputSourceType is a bitmask, we can precompute it and quickly match it.

    protected bool IsEventSourceCorrect(MotionEvent motionEvent)
    {
        // return true if event source matches AllowedInputSources.
    }
}
```

And then have all handlers inherit `AndroidInputHandler` and check the event source:

```cs
public class AndroidTouchHandler : AndroidInputHandler
{
    protected override IEnumerable<InputSourceType> AllowedInputSources => new[] { InputSourceType.Stylus, InputSourceType.Touchscreen };

    // ...
    
    private void handleTouch(object sender, View.TouchEventArgs e)
    {
        if (!IsEventSourceCorrect(e.Event)) return;
        
        // handle the event...
    }  
}
```

With that, we can have handlers subscribe to generic motion events, and avoid doing the source check in AndroidGameView.

osu.Framework.Android/Input/AndroidKeyboardHandler.cs
```diff
@@ -2,31 +2,30 @@
...
+            View.KeyUp += keyUp;
         }
 
         public override bool IsActive => true;
 
-        public override bool Initialize(GameHost host) => true;
-
         private void keyDown(Keycode keycode, KeyEvent e)
         {
+            if (!ShouldHandleEvent(e)) return;
```
bdach, line 27: I can generally agree with this overall structure but the base class should be responsible for checking `ShouldHandleEvent()` rather than each individual handler, if possible. Too easy to forget about it otherwise.

osu.Framework.Android/Input/AndroidInputHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public override bool Initialize(GameHost host)
+        {
+            if (!base.Initialize(host))
+                return false;
+
+            // compute the bitmask for later use.
+            foreach (var inputSource in HandledInputSources)
+            {
+                // ReSharper disable once BitwiseOperatorOnEnumWithoutFlags (InputSourceType is a flags enum, but is not properly marked as such)
```
bdach, line 49: Please don't put textual comments about resharper disables in the same line as the disable as that can have unintended effects such as disabling all inspections file-wide. See https://github.com/ppy/osu-framework/pull/4847#issuecomment-953180147, https://github.com/ppy/osu-framework/pull/4847#issuecomment-953464505.

osu.Framework.Android/Input/AndroidTouchHandler.cs
```diff
@@ -2,47 +2,44 @@
...
-            this.view = view;
-            view.Touch += handleTouch;
-            view.Hover += handleHover;
         }
 
-        public override bool Initialize(GameHost host) => true;
-
-        private void handleTouch(object sender, View.TouchEventArgs e)
+        protected override void OnTouch(MotionEvent touchEvent)
         {
```
Susko3, line 29: Using `MotionEvent` because `e.Event` is nullable and I didn't want to have a null check.

osu.Framework.Android/Input/AndroidInputHandler.cs
```diff
@@ -0,0 +1,319 @@
...
+        /// <summary>
+        /// The <see cref="InputSourceType"/>s that this <see cref="AndroidInputHandler"/> will handle.
+        /// </summary>
+        protected abstract IEnumerable<InputSourceType> HandledEventSources { get; }
+
+        /// <summary>
+        /// The <see cref="InputEventType"/>s that this <see cref="AndroidInputHandler"/> will handle.
+        /// </summary>
+        /// <remarks>For each <see cref="InputEventType"/> specified here, this <see cref="AndroidInputHandler"/> should override the appropriate method.</remarks>
+        protected abstract IEnumerable<InputEventType> HandledEventTypes { get; }
```
bdach, line 31: Not super sure on this setup. Rather than having to declare event types the handler is interested in and then having to override the correct methods (lest you get `NotSupportedException` at runtime), you could just subscribe to all events but leave the handlers empty. That'd get rid of some of the boilerplate.

I don't *hate* this setup though, will wait for more opinions.

peppy: I can agree with this. Seems like too much boilerplate for an edge case usage as it's currently written.

peppy (changes requested):
PR in general seems pretty fine, but would like to see it with manual event binding at each usage as proposed.

peppy:
Code looks good now. I'll try and give it a test today against touch and stylus.

peppy:
Stylus and touch work as expected here 👍

osu.Framework.Android/Input/AndroidInputHandler.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        /// <summary>
+        /// Invoked on every <see cref="AndroidGameView.CapturedPointer"/> event that matches an <see cref="InputSourceType"/> from <see cref="HandledEventSources"/>.
+        /// </summary>
+        /// <remarks>
+        /// Subscribe <see cref="HandleCapturedPointer"/> to <see cref="View"/>.<see cref="AndroidGameView.CapturedPointer"/> to receive events here.
+        /// </remarks>
+        protected virtual void OnCapturedPointer(MotionEvent capturedPointerEvent)
+        {
+            throw new NotSupportedException($"{nameof(HandleCapturedPointer)} subscribed to {nameof(View.CapturedPointer)} but the relevant method was not overriden.");
```
bdach, line 66: This isn't quite what I had in mind with my [last review](https://github.com/ppy/osu-framework/pull/4971#discussion_r776229506) but I suppose it works...?

To clarify, I was proposing subscribing to *all* input events which are available on the view at the `AndroidInputHandler` level, and calling the `OnEvent` methods on them there - as long as they match `eventSourceBitmask` - but leaving the input handlers blank to be overridden by inheritors. That way the only thing that needs to be done to handle an event in an implementation of this is overriding the correct handler, while this solution also requires manually hooking the events up in `OnInitialize()`.

osu.Framework.Android/Input/AndroidKeyboardHandler.cs
```diff
@@ -157,12 +175,5 @@ public static Key GetKeyCodeAsKey(Keycode keyCode)
             // this is the worst case scenario. Please note that the osu-framework keyboard handling cannot cope with Key.Unknown.
             return Key.Unknown;
         }
-
-        protected override void Dispose(bool disposing)
-        {
-            view.KeyDown -= keyDown;
-            view.KeyUp -= keyUp;
-            base.Dispose(disposing);
-        }
```
bdach, line 166: Also I notice that these unsubscribes on disposal seem to have been trimmed. Is something else going to do this automatically? It's probably not a huge deal either way but I feel safer with an explicit unsubscribe on disposal.

Susko3: https://github.com/ppy/osu-framework/blob/53b222719e0875c77f21737022f2194d7bb13c73/osu.Framework/Input/Handlers/InputHandler.cs#L79-L86

Base input handler sets enabled to false, so we're safe here.

osu.Framework.Android/Input/AndroidInputHandler.cs
Susko3: I tried doing it how [peppy suggested](https://github.com/ppy/osu-framework/pull/4971#pullrequestreview-842470905), and figured it was less wasteful than your proposal. So I went with that.

bdach: Not sure any overhead from these events is ever going to matter but I don't feel like splitting hairs over this.

bdach (approved):
code seems fine, haven't tested, relying on @peppy's testing as i could only test the same part of this PR at most (don't have required hardware to test mouse)

peppy (approved):

PR #4972, opened 2021-12-27
Don't fake-schedule SDL events

Susko3:
As discussed in https://github.com/ppy/osu-framework/pull/4969#issuecomment-1001160919.

Remaining things that are scheduled to the `eventScheduler`:
- [`storeWindowSizeToConfig()`](https://github.com/Susko3/osu-framework/blob/54fa49710ee92bd17d91fdc7ea257ace5229c53f/osu.Framework/Platform/SDL2DesktopWindow.cs#L481-L483) - rationale in inline comment
- in [`RequestClose()`](https://github.com/Susko3/osu-framework/blob/54fa49710ee92bd17d91fdc7ea257ace5229c53f/osu.Framework/Platform/SDL2DesktopWindow.cs#L494) - could be a `ScheduleCommand()`, but probably scheduled as event to possibly remove one frame of delay when closing (untested)
- [`TriggerMouseWheel` in `MacOSWindow`](https://github.com/Susko3/osu-framework/blob/54fa49710ee92bd17d91fdc7ea257ace5229c53f/osu.Framework/Platform/MacOS/MacOSWindow.cs#L59) - can't test this, so didn't remove. feel free to remove if you test it.

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -149,7 +149,7 @@ namespace osu.Framework.Platform
                 if (value.Equals(size)) return;
 
                 size = value;
-                ScheduleEvent(() => Resized?.Invoke());
+                Resized?.Invoke();
             }
         }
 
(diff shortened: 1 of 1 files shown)


osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -408,9 +408,6 @@ public virtual void Create()
 
             Exists = true;
 
-            MouseEntered += () => cursorInWindow.Value = true;
-            MouseLeft += () => cursorInWindow.Value = false;
```
bdach, line 412: just to clarify since it's not stated clearly anywhere - these were removed because `handleWindowEvent` already sets these flags, correct?

Susko3: Correct, I noted it in the commit message here: https://github.com/ppy/osu-framework/pull/4972/commits/54fa49710ee92bd17d91fdc7ea257ace5229c53f. Should I restate things like that in the PR body in the future?

bdach: honestly not sure. i'm not in the habit of checking individual commit messages, i usually review full diffs unless it is prohibitively difficult to do so. may be a personal thing though.

frenzibyte: I think it would be ideal to state it in the PR description for reviewers, and in the commit description for blamers.

peppy (approved):
In testing this seems to work correctly. Let's push it out for further testing.

PR #4973, opened 2021-12-30
Fix multiple deadlock cases in `WebRequest`

peppy:
Would have liked to split into synchronous and asynchronous paths completely, but can't be done until https://github.com/dotnet/runtime/pull/45300 is available to use (.net 6) as we depend on it for IPv4 fallback.

### DIFF
--- a/osu.Framework.Tests/IO/TestWebRequest.cs
+++ b/osu.Framework.Tests/IO/TestWebRequest.cs
@@ -696,6 +696,46 @@ namespace osu.Framework.Tests.IO
             Assert.Throws<ArgumentException>(() => request.AddParameter("cannot", "work", RequestParameterType.Form));
         }
 
+        /// <summary>
+        /// Ensure the work load, and importantly, the continuations do not run on the TPL thread pool.
+        /// Since we have our own task schedulers handling these load tasks.
+        /// </summary>
+        [Test]
+        public void TestSynchronousFlowTplReliance()
+        {
+            int workerMin;
+            int completionMin;
+            int workerMax;
+            int completionMax;
+
+            ManualResetEventSlim resetEvent = new ManualResetEventSlim();
+
+            // set limited threadpool capacity
+            ThreadPool.GetMinThreads(out workerMin, out completionMin);
+            ThreadPool.GetMaxThreads(out workerMax, out completionMax);
+
+            ThreadPool.SetMinThreads(2, 2);
+            ThreadPool.SetMaxThreads(2, 2);
+
+            var request = new DelayedWebRequest
+            {
+                Method = HttpMethod.Get,
+                AllowInsecureRequests = true,
+                Timeout = 1000,
+                Delay = 2
+            };
+
+            request.CompleteInvoked = () => request.Delay = 0;
+
+            request.Perform();
+
+            // restore capacity
+            resetEvent.Set();
+
+            ThreadPool.SetMinThreads(workerMin, completionMin);
+            ThreadPool.SetMaxThreads(workerMax, completionMax);
+        }
+
         [Test, Retry(5)]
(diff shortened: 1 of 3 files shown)


osu.Framework/IO/Network/WebRequest.cs
```diff
@@ -414,7 +409,9 @@ public void Perform()
         {
             try
             {
-                PerformAsync().Wait();
+                // Start a long-running task to ensure we don't block on a TPL thread pool thread.
+                // Unfortunately we can't use a full synchronous flow due to IPv4 fallback logic *requiring* the async path for now.
+                Task.Factory.StartNew(() => PerformAsync().Wait(), TaskCreationOptions.LongRunning).Wait();
```
bdach, line 414: To be fair I was never 100% on how to actually understand `LongRunning`. The docs are cagey on what it means:

> It also provides a hint to the task scheduler that an additional thread _might_ be required for the task so that it does not block the forward progress of other threads or work items on the local thread-pool queue.

(emphasis mine). So if taken at face value, then that means that this _could_ still deadlock? Not sure. But I suppose it's no worse than what it's already been, and potentially better.

osu.Framework.Tests/IO/TestWebRequest.cs
```diff
@@ -696,6 +696,46 @@ public void TestFormParamsNotSupportedForGet()
...
+                Timeout = 1000,
+                Delay = 2
+            };
+
+            request.CompleteInvoked = () => request.Delay = 0;
+
+            request.Perform();
+
+            // restore capacity
+            resetEvent.Set();
```
bdach, line 733: What is this doing? As far as I can tell nothing is waiting on this reset event. I think there should be two TPL tasks enqueued to do that that just... aren't here?

osu.Framework/IO/Network/WebRequest.cs
peppy: `LongRunning` will (in its current implementation) always make a new thread. We use it in all other places with this assumption.

Another option is to use `JoinableTaskFactory` from https://github.com/Microsoft/vs-threading in place of these. Seems like a decent library/option, but I opted to not add more dependencies or explore unknown solutions and stick to what I know works.

osu.Framework.Tests/IO/TestWebRequest.cs
peppy: I don't believe it's required any more. Was taken from the other recent test i wrote but the usage was removed due to being unnecessary for this test.

bdach: Well if it isn't required any more, then what's messing about with the thread pool thread counts, then? In the other test it was to limit TPL concurrency to max 4 threads and then saturate them to test independence of `LoadCompleteAsync()` from TPL. Going by that logic, then either the thread pool thread count changing should also go, or the reset event was necessary to begin with?

peppy: Probably needs an inline comment. The choice of 2 threads in the pool was specifically chosen, as it is the scenario which would previously cause a deadlock. Setting to 1 thread would make web requests not work at all, and setting to 3 would cause the deadlock to (usually) be avoided.

peppy: Have added more explanation, please see how that reads now.

peppy: Just found out that the task being used by NUnit is related to `Timeout` being specified on the test. Not sure whether that should also be mentioned in the inline comment.

bdach: I would probably add that info to the comment. As is this test sort of depends on nunit internals and so mentioning would make fixing easier if something in nunit ever changes.

bdach (approved):
seems good to me

Issue #4974, opened 2021-12-30
Ability to tripple-click on `TextBox` to select all text


oSumAtrIX:
Currently, `TextBox` only supports double-clicking to select the current word or space, but lacks the ability to select all text with a triple-click on `TextBox`. 
This behaviour should behave in the same way, when using the shortcut `CTRL+A` on text inputs which may not be limited to `TextBox` but other text-input elements.

PR #4975, opened 2021-12-31
Add ability to triple-click with mouse

oSumAtrIX:
Resolving #4974 with the exception of the test `triple click drag` failing.

### DIFF
--- a/osu.Framework.Tests/Visual/Input/TestSceneMouseStates.cs
+++ b/osu.Framework.Tests/Visual/Input/TestSceneMouseStates.cs
@@ -114,6 +114,7 @@ namespace osu.Framework.Tests.Visual.Input
         private static readonly Type drag_end = typeof(DragEndEvent);
         private static readonly Type click = typeof(ClickEvent);
         private static readonly Type double_click = typeof(DoubleClickEvent);
+        private static readonly Type triple_click = typeof(TripleClickEvent);
 
         [Test]
         public void BasicScroll()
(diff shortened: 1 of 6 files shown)


osu.Framework.Tests/Visual/Input/TestSceneMouseStates.cs
```diff
@@ -304,47 +305,79 @@ public void DragAndClick()
         }
 
         [Test]
-        public void ClickAndDoubleClick()
+        public void ClickAndMultipleClick()
```
peppy, line 308: Split out into a new test method. There's no need to overload this one further.

osu.Framework.Tests/Visual/Input/TestSceneMouseStates.cs
```diff
@@ -371,7 +422,7 @@ public void SeparateMouseDown()
             checkEventCount(mouse_up, 1);
         }
 
-        private void waitDoubleClickTime()
+        private void waitMultipleClickTime()
```
peppy, line 425: I don't think this needs to change name.

osu.Framework/Graphics/Drawable.cs
```diff
@@ -2158,6 +2161,16 @@ public bool TriggerEvent(UIEvent e)
...
 
+        /// <summary>
+        /// An event that occurs when a <see cref="MouseButton"/> is triple-clicked on this <see cref="Drawable"/>.
+        /// </summary>
+        /// <remarks>
+        /// This will only be invoked on the <see cref="Drawable"/> that returned <code>true</code> from a previous <see cref="OnDoubleClick"/> invocation.
+        /// </remarks>
+        /// <param name="e">The <see cref="TripleClickEvent"/> containing information about the input event.</param>
+        /// <returns>Whether to block the next <see cref="OnClick"/> event from occurring.</returns>
+        protected virtual bool OnTripleClick(TripleClickEvent e) => Handle(e);
```
peppy, line 2172: This seems like such a rare use case that it should *not* be implemented at a `Drawable` level, but instead locally to `Textbox`.

peppy (changes requested):

peppy:
Well the number of code changes has been reduced, but the event is still at a level I'm not sure we want it at. You've added the event type but it has zero usage or testing outside of `TextBox`. My proposal was instead to do the actual triple click handling logic *completely within `TextBox`* (ie. removing all the implementation from `MouseButtonEventManager`.

smoogipoo:
I also agree that I don't think this should be in `MouseButtonEventManager`. No API I've ever used exposed triple clicks.

Another option, from looking at javascript (https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#Properties), is to similarly expose the number of consecutive clicks within a short period via `ClickEvent`.

Issue #4693, opened 2021-08-17
Option to disable osu!framework applications from blocking compositors


smoogipoo:
As discussed in https://github.com/ppy/osu-framework/issues/4642, we probably want an option to disable the compositor-blocking default behaviour of SDL, which can be controlled via `SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR`.

We may even want this to be the default, since I believe other applications like browsers/electron apps don't block the compositor. This is a low-priority issue for us.

turbedi:
Fixed by #4976

PR #4976, opened 2021-12-31
Improve host configuration

Speykious:
This PR originally aimed at fixing #4693. However, after much discussion, it turned into implementing a better way to configure game hosts. Thus #4693 will be fixed as a side-effect.

***

(Old post about bypassing the compositor on Linux:)

The change has been tested and successful so far on Arch Linux with the KDE desktop environment (basically: it works on my machine), specifically by opening the visual tests window which would have normally disabled the compositor.
Given that this change is just about the compositor, it shouldn't impact anything else in the framework, so testing-wise I think it's fine.
(also it's quite literally a single line change.)

Last thing: with that in mind, games which do want to bypass the compositor (such as osu!) would have to use the `SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR` environment variable, unless we directly code in a way to specify whether we want to do that. What would be the preferred way here in your view?

### DIFF
--- a/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
+++ b/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
@@ -7,6 +7,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using NUnit.Framework;
 using osu.Framework.Allocation;
+using osu.Framework.Extensions;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
@@ -261,7 +262,7 @@ namespace osu.Framework.Tests.Exceptions
             [BackgroundDependencyLoader]
             private void load()
             {
-                Task.Delay((int)(1000 / Clock.Rate)).Wait();
+                Task.Delay((int)(1000 / Clock.Rate)).WaitSafely();
                 if (throws)
                     throw new AsyncTestException();
             }
--- a/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
+++ b/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
@@ -199,7 +199,11 @@ namespace osu.Framework.Tests.Exceptions
 
             try
             {
-                using (var host = new TestRunHeadlessGameHost($"{GetType().Name}-{Guid.NewGuid()}", realtime: false))
+                using (var host = new TestRunHeadlessGameHost(new HostConfig
+                {
+                    Name = $"{GetType().Name}-{Guid.NewGuid()}",
+                    Realtime = false,
+                }))
                 {
                     using (var game = new TestGame())
                     {
--- a/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
+++ b/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
@@ -199,7 +199,7 @@ namespace osu.Framework.Tests.Exceptions
 
             try
             {
-                using (var host = new TestRunHeadlessGameHost(new HostOptions { Name = $"{GetType().Name}-{Guid.NewGuid()}" }))
+                using (var host = new TestRunHeadlessGameHost($"{GetType().Name}-{Guid.NewGuid()}", new HostOptions()))
                 {
                     using (var game = new TestGame())
                     {
--- a/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
+++ b/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
@@ -200,10 +200,10 @@ namespace osu.Framework.Tests.Exceptions
             try
             {
                 using (var host = new TestRunHeadlessGameHost(new HostOptions
-                {
-                    Name = $"{GetType().Name}-{Guid.NewGuid()}",
-                    Realtime = false,
-                }))
+                       {
+                           Name = $"{GetType().Name}-{Guid.NewGuid()}",
+                           Realtime = false,
+                       }))
                 {
                     using (var game = new TestGame())
                     {
--- a/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
+++ b/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
@@ -199,11 +199,7 @@ namespace osu.Framework.Tests.Exceptions
 
             try
             {
-                using (var host = new TestRunHeadlessGameHost(new HostOptions
-                       {
-                           Name = $"{GetType().Name}-{Guid.NewGuid()}",
-                           Realtime = false,
-                       }))
+                using (var host = new TestRunHeadlessGameHost(new HostOptions { Name = $"{GetType().Name}-{Guid.NewGuid()}" }))
                 {
                     using (var game = new TestGame())
                     {
--- a/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
+++ b/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
@@ -199,7 +199,7 @@ namespace osu.Framework.Tests.Exceptions
 
             try
             {
-                using (var host = new TestRunHeadlessGameHost(new HostConfig
+                using (var host = new TestRunHeadlessGameHost(new HostOptions
                 {
                     Name = $"{GetType().Name}-{Guid.NewGuid()}",
                     Realtime = false,
--- a/osu.Framework/Host.cs
+++ b/osu.Framework/Host.cs
@@ -11,7 +11,7 @@ namespace osu.Framework
 {
     public static class Host
     {
-        [Obsolete("Use GetSuitableDesktopHost(string, HostOptions) instead.")]
+        [Obsolete("Use GetSuitableDesktopHost(string, HostOptions) instead.")] // Can be removed 20220715
         public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = false, bool portableInstallation = false)
         {
             return GetSuitableDesktopHost(gameName, new HostOptions
--- a/osu.Framework/Host.cs
+++ b/osu.Framework/Host.cs
@@ -26,18 +26,25 @@ namespace osu.Framework
                     return new WindowsGameHost(gameName, bindIPC, portableInstallation);
 
                 default:
-                    throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({Enum.GetName(typeof(RuntimeInfo.Platform), RuntimeInfo.OS)}).");
+                    throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({RuntimeInfo.OS}).");
             }
         }
 
         public static DesktopGameHost GetSuitableHost(HostConfig hostConfig)
         {
-            return RuntimeInfo.OS switch
+            switch (RuntimeInfo.OS)
             {
-                RuntimeInfo.Platform.Windows => new WindowsGameHost(hostConfig),
-                RuntimeInfo.Platform.Linux => new LinuxGameHost(hostConfig),
-                RuntimeInfo.Platform.macOS => new MacOSGameHost(hostConfig),
-                _ => throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({Enum.GetName(typeof(RuntimeInfo.Platform), RuntimeInfo.OS)})."),
+                case RuntimeInfo.Platform.Windows:
+                    return new WindowsGameHost(hostConfig);
+
+                case RuntimeInfo.Platform.Linux:
+                    return new LinuxGameHost(hostConfig);
+
+                case RuntimeInfo.Platform.macOS:
+                    return new MacOSGameHost(hostConfig);
+
+                default:
+                    throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({RuntimeInfo.OS}).");
             };
         }
     }
--- a/osu.Framework/Host.cs
+++ b/osu.Framework/Host.cs
@@ -11,7 +11,7 @@ namespace osu.Framework
 {
     public static class Host
     {
-        public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = false, bool portableInstallation = false)
+        public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = false, bool portableInstallation = false, bool bypassCompositorOnLinux = true)
         {
             switch (RuntimeInfo.OS)
             {
@@ -19,7 +19,7 @@ namespace osu.Framework
                     return new MacOSGameHost(gameName, bindIPC, portableInstallation);
 
                 case RuntimeInfo.Platform.Linux:
-                    return new LinuxGameHost(gameName, bindIPC, portableInstallation);
+                    return new LinuxGameHost(gameName, bindIPC, portableInstallation, bypassCompositorOnLinux);
 
                 case RuntimeInfo.Platform.Windows:
                     return new WindowsGameHost(gameName, bindIPC, portableInstallation);
--- a/osu.Framework/Host.cs
+++ b/osu.Framework/Host.cs
@@ -11,17 +11,17 @@ namespace osu.Framework
 {
     public static class Host
     {
-        [Obsolete("Use GetSuitableHost(HostConfig) instead.")]
+        [Obsolete("Use GetSuitableDesktopHost(string, HostOptions) instead.")]
         public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = false, bool portableInstallation = false)
         {
-            return GetSuitableHost(gameName, new HostOptions
+            return GetSuitableDesktopHost(gameName, new HostOptions
             {
                 BindIPC = bindIPC,
                 PortableInstallation = portableInstallation,
             });
         }
 
-        public static DesktopGameHost GetSuitableHost(string gameName, HostOptions hostOptions)
+        public static DesktopGameHost GetSuitableDesktopHost(string gameName, HostOptions hostOptions = null)
         {
             switch (RuntimeInfo.OS)
             {
--- a/osu.Framework/Host.cs
+++ b/osu.Framework/Host.cs
@@ -11,7 +11,8 @@ namespace osu.Framework
 {
     public static class Host
     {
-        public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = false, bool portableInstallation = false, bool bypassCompositorOnLinux = true)
+        [Obsolete("Use GetSuitableHost(HostConfig) instead.")]
+        public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = false, bool portableInstallation = false)
         {
             switch (RuntimeInfo.OS)
             {
@@ -19,7 +20,7 @@ namespace osu.Framework
                     return new MacOSGameHost(gameName, bindIPC, portableInstallation);
 
                 case RuntimeInfo.Platform.Linux:
-                    return new LinuxGameHost(gameName, bindIPC, portableInstallation, bypassCompositorOnLinux);
+                    return new LinuxGameHost(gameName, bindIPC, portableInstallation);
 
                 case RuntimeInfo.Platform.Windows:
                     return new WindowsGameHost(gameName, bindIPC, portableInstallation);
@@ -27,6 +28,17 @@ namespace osu.Framework
                 default:
                     throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({Enum.GetName(typeof(RuntimeInfo.Platform), RuntimeInfo.OS)}).");
             }
+        }
+
+        public static DesktopGameHost GetSuitableHost(HostConfig hostConfig)
+        {
+            return RuntimeInfo.OS switch
+            {
+                RuntimeInfo.Platform.Windows => new WindowsGameHost(hostConfig),
+                RuntimeInfo.Platform.Linux => new LinuxGameHost(hostConfig),
+                RuntimeInfo.Platform.macOS => new MacOSGameHost(hostConfig),
+                _ => throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({Enum.GetName(typeof(RuntimeInfo.Platform), RuntimeInfo.OS)})."),
+            };
         }
     }
 }
--- a/osu.Framework/Host.cs
+++ b/osu.Framework/Host.cs
@@ -14,26 +14,25 @@ namespace osu.Framework
         [Obsolete("Use GetSuitableHost(HostConfig) instead.")]
         public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = false, bool portableInstallation = false)
         {
-            return GetSuitableHost(new HostOptions
+            return GetSuitableHost(gameName, new HostOptions
             {
-                Name = gameName,
                 BindIPC = bindIPC,
                 PortableInstallation = portableInstallation,
             });
         }
 
-        public static DesktopGameHost GetSuitableHost(HostOptions hostOptions)
+        public static DesktopGameHost GetSuitableHost(string gameName, HostOptions hostOptions)
         {
             switch (RuntimeInfo.OS)
             {
                 case RuntimeInfo.Platform.Windows:
-                    return new WindowsGameHost(hostOptions);
+                    return new WindowsGameHost(gameName, hostOptions);
 
                 case RuntimeInfo.Platform.Linux:
-                    return new LinuxGameHost(hostOptions);
+                    return new LinuxGameHost(gameName, hostOptions);
 
                 case RuntimeInfo.Platform.macOS:
-                    return new MacOSGameHost(hostOptions);
+                    return new MacOSGameHost(gameName, hostOptions);
 
                 default:
                     throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({RuntimeInfo.OS}).");
--- a/osu.Framework/Host.cs
+++ b/osu.Framework/Host.cs
@@ -14,20 +14,12 @@ namespace osu.Framework
         [Obsolete("Use GetSuitableHost(HostConfig) instead.")]
         public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = false, bool portableInstallation = false)
         {
-            switch (RuntimeInfo.OS)
+            return GetSuitableHost(new HostConfig
             {
-                case RuntimeInfo.Platform.macOS:
-                    return new MacOSGameHost(gameName, bindIPC, portableInstallation);
-
-                case RuntimeInfo.Platform.Linux:
-                    return new LinuxGameHost(gameName, bindIPC, portableInstallation);
-
-                case RuntimeInfo.Platform.Windows:
-                    return new WindowsGameHost(gameName, bindIPC, portableInstallation);
-
-                default:
-                    throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({RuntimeInfo.OS}).");
-            }
+                Name = gameName,
+                BindIPC = bindIPC,
+                PortableInstallation = portableInstallation,
+            });
         }
 
         public static DesktopGameHost GetSuitableHost(HostConfig hostConfig)
--- a/osu.Framework/Host.cs
+++ b/osu.Framework/Host.cs
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using osu.Framework.Platform;
 using osu.Framework.Platform.Linux;
 using osu.Framework.Platform.MacOS;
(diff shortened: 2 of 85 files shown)


peppy:
I'm not even sure it's worth keeping this open. In its current state it is useless. You can add this hint in your application and get the same effect.

By adding it in the framework without any way to mutate it, you are making a breaking change, and one that we likely do not want.

Speykious:
Hmmm... Looks a tiny bit tricky to do the exposing right. I'll try though.

Wouldn't adding this hint in other applications require adding the `ppy.SDL2-CS` package to dependencies as well? In my view letting the framework expose it is the best option.

Speykious:
I'm really unsure if the `LinuxGameHost` should be responsible for this, not to mention that I then had to add an additional argument to the method in the static `Host` class...

peppy:
To confirm, having compositor enabled means the window is being re-buffered and thus introducing addition latency, correct?

bdach:
[This SDL thread](https://discourse.libsdl.org/t/stop-sdl2-telling-linux-to-disable-desktop-compositor/23206/2) mentions something about keeping the compositor enabled "lowering framerate", so I'd wager that it does indeed incur latency. So definitely not something that should be blanket enabled like this pull originally did.

osu.Framework/Host.cs
```diff
@@ -11,15 +11,15 @@ namespace osu.Framework
 {
     public static class Host
     {
-        public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = false, bool portableInstallation = false)
+        public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = false, bool portableInstallation = false, bool bypassCompositorOnLinux = true)
```
bdach, line 14: This is most definitely not the right place to put the switch though.

bdach (changes requested):

osu.Framework/Host.cs
Speykious: Figured. I felt dirty putting it there lol. But now I really wonder what's a cleaner way that I can do it.

peppy: Well I was also thinking along these lines, but if it is here, it would need to be named something more generic (and we'd need to ascertain if the same should/could be applied to other platforms).

For instance, on windows this is similar to the concept of exclusive fullscreen, and on macOS there are multiple fullscreen modes, one which actually gives the application more direct control of the screen.

bdach: Not sure about clumping compositor switches with exclusive fullscreen. A closer windows analogue would be to disable DWM somehow. No idea about Macs.

I'd see this option best placed on a `LinuxDesktopWindow : SDL2DesktopWindow`. And probably configured in-source per framework user (as either the end application is a game and wants compositor off, or is an... application, and wants to keep it on).

peppy: we probably need a `Configuration` class to pass into `GetSuitableHost` so we can extend these kinds of settings without adding more parameters..

Speykious: SO some kind of `HostConfiguration` instead of several parameters then?

peppy: I think that's best for now yep. Just keep in mind the existing methods should also be kept in place for compatibility reasons.

Speykious: Ok, I'll keep that in mind.

osu.Framework/Host.cs
```diff
@@ -28,5 +29,16 @@ public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = fa
                     throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({Enum.GetName(typeof(RuntimeInfo.Platform), RuntimeInfo.OS)}).");
             }
         }
+
+        public static DesktopGameHost GetSuitableHost(HostConfig hostConfig)
+        {
+            return RuntimeInfo.OS switch
```
bdach, line 35: Please use a normal switch statement rather than a switch expression (as that is the accepted code convention in this repository).

osu.Framework/Host.cs
```diff
@@ -28,5 +29,16 @@ public static DesktopGameHost GetSuitableHost(string gameName, bool bindIPC = fa
...
+            {
+                RuntimeInfo.Platform.Windows => new WindowsGameHost(hostConfig),
+                RuntimeInfo.Platform.Linux => new LinuxGameHost(hostConfig),
+                RuntimeInfo.Platform.macOS => new MacOSGameHost(hostConfig),
+                _ => throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({Enum.GetName(typeof(RuntimeInfo.Platform), RuntimeInfo.OS)})."),
```
bdach, line 40: Seems overkill?

```suggestion
                _ => throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({RuntimeInfo.OS)})."),
```

osu.Framework/HostConfig.cs
```diff
@@ -0,0 +1,35 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework
+{
+    public struct HostConfig
+    {
+        public string GameName { get; set; }
+        public bool BindIPC { get; set; }
+        public bool PortableInstallation { get; set; }
+        public bool BypassCompositor { get; set; }
```
bdach, line 11: These should all have xmldoc.

osu.Framework/HostConfig.cs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework
+{
+    public struct HostConfig
```
bdach, line 6: Not sure why `struct`. If it's not going to be a performance killer (which it isn't) I prefer `class`. Less gotchas.

osu.Framework/HostConfig.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        public string GameName { get; set; }
+        public bool BindIPC { get; set; }
+        public bool PortableInstallation { get; set; }
+        public bool BypassCompositor { get; set; }
+
+        public static HostConfig GameConfig(string gameName, bool bindIPC = false, bool portableInstallation = false)
```
bdach, line 13: Unsure that these need to exist.

bdach (changes requested):

osu.Framework/Host.cs
Speykious: Ah, ok, didn't know.

osu.Framework/Host.cs
Speykious: I agree (all I did was copy the exception from the old method)

osu.Framework/HostConfig.cs
Speykious: Will try to come up with good ones

osu.Framework/HostConfig.cs
Speykious: I thought it would make the intent of the program more explicit.

osu.Framework/HostConfig.cs
Speykious: I guess

Speykious: I think that I did that not for performance but rather semantic reasons, since it has no instance methods.

bdach: I suppose? Would also be fine with `readonly struct` I guess, but that won't work well without `get; init;` which I'm not sure will be allowed given the .NET Standard dual-targeting which is required for mobile.

Mutable structs are the worst of both worlds IMO as you cannot just mutate an instance of a struct anyway because the member access operator returns a temporary.

osu.Framework/Platform/DesktopGameHost.cs
```diff
@@ -23,13 +24,21 @@ public abstract class DesktopGameHost : GameHost
         private TcpIpcProvider ipcProvider;
         private readonly bool bindIPCPort;
 
+        [Obsolete("Use DesktopGameHost(HostConfig) instead.")]
         protected DesktopGameHost(string gameName = @"", bool bindIPCPort = false, bool portableInstallation = false)
             : base(gameName)
         {
             this.bindIPCPort = bindIPCPort;
             IsPortableInstallation = portableInstallation;
         }
```
Susko3, line 33: [`HeadlessGameHost`](https://github.com/ppy/osu-framework/blob/44794a221458d9cbb9fc67050c2e5c949a0d23fe/osu.Framework/Platform/HeadlessGameHost.cs#L33-L37) is using this obsolete ctor, and so are some tests.

osu.Framework/HostConfig.cs
Speykious: Oh, I didn't know about `init;`, that's a nice feature.
The netstandard2.1 target doesn't use the latest C# version?

osu.Framework/Platform/DesktopGameHost.cs
Speykious: Ah, I overlooked the tests!

osu.Framework/HostConfig.cs
bdach: As I said already, I don't know. But going off docs [I would wager that the answer is no](https://docs.microsoft.com/en-us/dotnet/standard/net-standard#when-to-target-net50-or-net60-vs-netstandard).

netstandard cannot be dropped yet because that's what mobile platforms use. with some luck .NET 6 could allow dropping that but i have no idea if it actually will.

Speykious: I see. Well, good to know that .NET 6 might fix things like this.
I'll turn it into a class for now.

osu.Framework/HostConfig.cs
Speykious: I thought about it a bit more and I think we can just do without.

osu.Framework/Platform/DesktopGameHost.cs
Speykious: Alright, I think I took all tests into account now.

Speykious:
Time to fix failing tests (looks like it's only about obsolete ctors still in use).

Speykious:
Some of the tests only had a constructor with a single `string name` parameter, so I left those as-is and only modified the internal code to use the non-obsolete constructor.

osu.Framework/Platform/Windows/WindowsGameHost.cs
```diff
@@ -32,11 +32,17 @@ public class WindowsGameHost : DesktopGameHost
 #endif
         public override bool CapsLockEnabled => Console.CapsLock;
 
+        [Obsolete("Use WindowsGameHost(HostConfig) instead.")]
         internal WindowsGameHost(string gameName, bool bindIPC = false, bool portableInstallation = false)
             : base(gameName, bindIPC, portableInstallation)
         {
         }
```
Susko3, line 39: Since this is `internal`, this could be removed without an `[Obsolete]`.
The obsolete `GetSuitableHost()` could then just construct a `HostConfig` and pass it onto `GetSuitableHost(HostConfig)`.

Same goes for other desktop hosts.

osu.Framework/Platform/HeadlessGameHost.cs
```diff
@@ -30,10 +30,22 @@ public class HeadlessGameHost : DesktopGameHost
...
+                Name = gameName,
+                BindIPC = bindIPC,
+                Realtime = realtime,
+                PortableInstallation = portableInstallation,
+            })
+        {
+        }
+
+        public HeadlessGameHost(HostConfig hostConfig)
+            : base(hostConfig.With(name: hostConfig.Name ?? Guid.NewGuid().ToString()))
```
Susko3, line 46: Currently, `HeadlessGameHost` allows parameter-less construction. Same goes for inheritors of `HeadlessGameHost`.

`TestLogging` and some usages osu-side use parameter-less construction, with this, those usages can remain the same.

If this is something we want to keep, we can do this:

```suggestion
        public HeadlessGameHost(HostConfig hostConfig = null)
            : base((hostConfig ??= new HostConfig()).With(name: hostConfig.Name ?? Guid.NewGuid().ToString()))
```

with adequate changes to inheriting classes.

osu.Framework/HostConfig.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// On Linux, the compositor re-buffers the application to apply various window effects,
+        /// increasing latency in the process. Thus it is a good idea for games to bypass it,
+        /// though normal applications would generally benefit from letting the window effects untouched. <br/>
+        /// If the SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR environment variable is set, this property will have no effect.
+        /// </remarks>
+        public bool BypassCompositor { get; set; }
+
+        public HostConfig()
+        {
+            Name = @"";
```
Susko3, line 48: Currently, `HeadlessGameHost` expects a null name as a default (so it can fall back to generating GUID), while `DesktopGameHost` and `GameHost` expect an empty string.

Probably best to have this null, and have a null check in `DesktopGameHost` or `GameHost`. But I'm not sure what is the definitive meaning of `strting.Emtpy` for Name in the context of different `GameHost`s.

Currently, having an empty name will cause `Storage` to be rooted in `%appdata%` on Windows, something we probably don't want. Empty path should probably throw in `DesktopGameHost`.

Empty string might be "supported" when using a portable installation.

Empty string seems to be used for mobile platforms, though.

osu.Framework/Testing/TestRunHeadlessGameHost.cs
```diff
@@ -20,13 +21,21 @@ public class TestRunHeadlessGameHost : HeadlessGameHost
 
         public static string TemporaryTestDirectory = Path.Combine(Path.GetTempPath(), "of-test-headless");
 
+        [Obsolete("Use TestRunHeadlessGameHost(HostConfig, bool) instead.")]
         public TestRunHeadlessGameHost(string name = null, bool bindIPC = false, bool realtime = false, bool portableInstallation = false, bool bypassCleanup = false)
             : base(name, bindIPC, realtime, portableInstallation)
         {
             this.bypassCleanup = bypassCleanup;
             UserStoragePaths = TemporaryTestDirectory.Yield();
         }
```
Susko3, line 30: This should probably also use `this(HostConfig, bool)` ctor (same as `HeadlessGameHost`) instead of relying on the obsolete ctor in `HeadlessGameHost`. This will reduce code duplication.

Same goes for `BackgroundGameHeadlessGameHost`.

Susko3 (changes requested):

osu.Framework/Platform/HeadlessGameHost.cs
Speykious: I cannot make this because the deprecated constructor still exists and covers the parameterless construction.

Speykious: Actually, if I edit the deprecated constructor to have the first argument mandatory, I could make the new constructor optionally parameterless...

Speykious: Actually, if I edit the deprecated constructor to have the first argument mandatory, I could make the new constructor optionally parameterless...

osu.Framework/HostConfig.cs
bdach: If null and empty string are going to mean different things (which I don't know that they should), then `Name` better be annotated as `string?` and xmldoc'd properly to explain what the point of that is.

Speykious: `string?` would give me a code style warning. I think it would make more sense once osu!framework gets ported to .NET 6 if it enables nullables.

bdach: then jetbrains' `[CanBeNull]` at least

peppy (approved):

peppy:
Please don't self-resolve review comments. I can't see what I even said any more.

peppy:
This has turned into a bit of a mess..

osu.Framework/Host.cs
```diff
@@ -1,31 +1,42 @@
...
+                    return new WindowsGameHost(hostOptions);
 
                 case RuntimeInfo.Platform.Linux:
-                    return new LinuxGameHost(gameName, bindIPC, portableInstallation);
+                    return new LinuxGameHost(hostOptions);
 
-                case RuntimeInfo.Platform.Windows:
-                    return new WindowsGameHost(gameName, bindIPC, portableInstallation);
+                case RuntimeInfo.Platform.macOS:
+                    return new MacOSGameHost(hostOptions);
```
peppy, line 37: why was this order even changed?

osu.Framework/Host.cs
```diff
@@ -1,31 +1,42 @@
...
 
                 default:
-                    throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({Enum.GetName(typeof(RuntimeInfo.Platform), RuntimeInfo.OS)}).");
+                    throw new InvalidOperationException($"Could not find a suitable host for the selected operating system ({RuntimeInfo.OS}).");
```
peppy, line 39: Why was this changed?

osu.Framework/HostOptions.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </summary>
+    public class HostOptions
+    {
+        /// <summary>
+        /// Name of the game or application.
+        /// </summary>
+        /// <remarks>
+        /// This property may be null. Host types like <see cref="Platform.HeadlessGameHost"/> fallback to a custom GUID string when it occurs.
+        /// </remarks>
+        [CanBeNull]
```
peppy, line 19: Specify `#nullable enable` on the file instead.

osu.Framework/HostOptions.cs
```diff
@@ -0,0 +1,61 @@
...
+        public bool BindIPC { get; set; }
+
+        /// <summary>
+        /// Whether it runs in real time.
+        /// </summary>
+        /// <remarks>
+        /// This is particularly relevant for a <see cref="Platform.HeadlessGameHost"/>.
+        /// Any other host, such as <see cref="Platform.DesktopGameHost"/>, is assumed to run in real time.
+        /// </remarks>
+        public bool Realtime { get; set; }
```
peppy, line 34: Shouldn't be here. Put back in parameters for the local usage for now.

osu.Framework/HostOptions.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <summary>
+        /// Whether to bypass the compositor.
+        /// </summary>
+        /// <remarks>
+        /// On Linux, the compositor re-buffers the application to apply various window effects,
+        /// increasing latency in the process. Thus it is a good idea for games to bypass it,
+        /// though normal applications would generally benefit from letting the window effects untouched. <br/>
+        /// If the SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR environment variable is set, this property will have no effect.
+        /// </remarks>
+        public bool BypassCompositor { get; set; }
```
peppy, line 50: Unsure this should be directly in `HostOptions` if it's going to specifically mention `Linux`.

We may want to consider `LinuxHostOptions : HostOptions { bool BypassCompositor }` or `HostOptions { PlatformHostOptions PlatformOptions { get; set; } }`

Open to opinions on this.

osu.Framework/HostOptions.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <remarks>
+        /// This property may be null. Host types like <see cref="Platform.HeadlessGameHost"/> fallback to a custom GUID string when it occurs.
+        /// </remarks>
+        [CanBeNull]
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Whether to bind the IPC port.
+        /// </summary>
+        public bool BindIPC { get; set; }
```
peppy, line 25: I think this is also desktop specific? Not sure it should be sent the whole way down. Maybe okay.

peppy (changes requested):

peppy:
I've self-applied the review comments I left to try and breathe some sense into the diff for further review. Still need discussion on whether the compositor option should sit inside `HostOptions` or not. Maybe it's fine to leave it in there for simplicity, but thinking forward it may not be the best place if we are looking for a permanent home.

osu.Framework/HostOptions.cs
Susko3: I'm opposed to `LinuxHostOptions` and the like.

Consumers ares supposed to use `HostOptions` in `GetSuitableHost()`. Feels wrong having to pass `LinuxHostOptions` even if not running on linux. If in the future we have eg. `WindowsHostOptions`, trying to pass in both windows and linux options won't easily work.

osu.Framework/HostOptions.cs
Susko3: Same can be argued for `PortableInstallation`.

But the whole `GetSuitableHost()` is desktop specific, as-is.

osu.Framework/HostOptions.cs
peppy: Yeah, for sure. If anything I think it should be the second of the proposed (having a separate `Options` class for platform specific options).

osu.Framework/HostOptions.cs
peppy: Maybe it's fine to leave desktop-related settings here for simplicity as it's the majority use case.

Speykious:
> Please don't self-resolve review comments. I can't see what I even said any more.

Sorry about that, understood

Speykious:
> but thinking forward it may not be the best place if we are looking for a permanent home.

What do you mean by permanent home?

peppy:
Aka future proofing its location so we don't need to move it again in the future when we add even more options.

osu.Framework/Host.cs
Speykious: I... don't know. I guess it should stay the same

peppy: It's fine to leave it for now, just curious.

osu.Framework/Host.cs
Speykious: See https://github.com/ppy/osu-framework/pull/4976#discussion_r777601844

Speykious:
Ah, I see... I think it's fine keeping platform-specific options in `HostOptions`. It puts all relevant host options in one place, and the only thing that happens when it doesn't apply to the platform is that it gets ignored.
There may be a better way inside of `HostOptions` to organize fields based on which platform they apply in, but externally when that object is passed around it's fine

osu.Framework/HostOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Various configuration properties for a <see cref="Host"/>.
+    /// </summary>
+    public class HostOptions
+    {
+        /// <summary>
+        /// Name of the game or application.
+        /// </summary>
+        /// <remarks>
+        /// This property may be null. Host types like <see cref="Platform.HeadlessGameHost"/> fallback to a custom GUID string when it occurs.
+        /// </remarks>
```
bdach, line 18: This property is now doubly confusing in that (a) I'd expect `null` to be the default value if it's permissible and does a different thing than an empty string does, and (b) `HeadlessGameHost` is now *not* using the null special path (as opposed to manually specifying `Guid.NewGuid` via the ctor overload with no explicit name in it).

In fact as far as I can tell all paths now expect this to be non-null, so this xmldoc is just stale and can be removed safely?

```suggestion
```

osu.Framework/HostOptions.cs
bdach: Maybe if we ever get to a point wherein this starts becoming a mess then we could arrange the platform specific stuff in a hierarchical manner, e.g.

```
HostOptions
 |- // general options go here
 |- LinuxOptions
 |   |- BypassCompositor
 |- DesktopOptions
     |- PortableInstallation
     |- BindIPC
```

But that's probably a bit too baroque for now... Maybe just throw some `#region`s in, grouping the platform specific options, and call it a day?

osu.Framework/HostOptions.cs
bdach: Also, if the above understanding is correct and the null path is not required anymore, I'd probably consider forcing all users to specify a name for their game by adding a ctor that requires it (and therefore preventing construction of `HostOptions` without giving it a name).

osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
```diff
@@ -199,7 +199,7 @@ private void runGameWithLogic(Action<Game> logic, Func<Game, bool> exitCondition
 
             try
             {
-                using (var host = new TestRunHeadlessGameHost($"{GetType().Name}-{Guid.NewGuid()}", realtime: false))
+                using (var host = new TestRunHeadlessGameHost(new HostOptions { Name = $"{GetType().Name}-{Guid.NewGuid()}" }))
```
bdach, line 202: `realtime: false` spec looks to have been dropped here

osu.Framework/Platform/HeadlessGameHost.cs
```diff
@@ -30,8 +30,19 @@ public class HeadlessGameHost : DesktopGameHost
 
         public override IEnumerable<string> UserStoragePaths => new[] { "./headless/" };
 
-        public HeadlessGameHost(string gameName = null, bool bindIPC = false, bool realtime = true, bool portableInstallation = false)
-            : base(gameName ?? Guid.NewGuid().ToString(), bindIPC, portableInstallation: portableInstallation)
+        [Obsolete("Use HeadlessGameHost(HostConfig hostConfig) instead.")]
```
bdach, line 33: obsolete notices need updating too

```suggestion
        [Obsolete("Use HeadlessGameHost(HostOptions hostOptions) instead.")]
```

removal dates also missing (goes for more places than this one)

osu.Framework/Platform/HeadlessGameHost.cs
```diff
@@ -30,8 +30,19 @@ public class HeadlessGameHost : DesktopGameHost
...
+            {
+                Name = gameName,
+                BindIPC = bindIPC,
+                PortableInstallation = portableInstallation,
+            }, realtime)
+        {
+        }
+
+        public HeadlessGameHost(HostOptions options = null, bool realtime = true)
+            : base(options ?? new HostOptions { Name = Guid.NewGuid().ToString() })
```
Susko3, line 45: This is currently different from `master`:

```cs
new HeadlessGameHost(bindIPC: true);
```
will produce different results to 
```cs
new HeadlessGameHost(new HostOptions { BindIPC = true });
```

The former will have guid name, while the latter will have empty name.

This was correct previously:
```cs
: base(hostConfig.With(name: hostConfig.Name ?? Guid.NewGuid().ToString()))
```

osu.Framework/Platform/HeadlessGameHost.cs
```diff
@@ -30,8 +30,19 @@ public class HeadlessGameHost : DesktopGameHost
 
         public override IEnumerable<string> UserStoragePaths => new[] { "./headless/" };
 
-        public HeadlessGameHost(string gameName = null, bool bindIPC = false, bool realtime = true, bool portableInstallation = false)
-            : base(gameName ?? Guid.NewGuid().ToString(), bindIPC, portableInstallation: portableInstallation)
+        [Obsolete("Use HeadlessGameHost(HostConfig hostConfig) instead.")]
+        public HeadlessGameHost(string gameName, bool bindIPC = false, bool realtime = true, bool portableInstallation = false)
```
Susko3, line 34: Method signature was changed from `master` (`gameName` no longer defaults to null).

Susko3 (changes requested):

osu.Framework/Platform/HeadlessGameHost.cs
bdach: I believe this was intentional? You can't have two ctors with all parameters having default values, the compiler will correctly complain that it cannot choose between the two. The other ctor will take over the default `null` case (although, as you rightly mentioned in the other comment, it doesn't behave 100% correctly if `HostOptions` are specified with no name spec).

osu.Framework/Platform/HeadlessGameHost.cs
peppy: I've removed the `With` method for simplicity. We can't use that.

osu.Framework/Platform/HeadlessGameHost.cs
peppy: I intentionally didn't add/update these until we are ready to agree on this PR (don't think that time has come yet).

osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
peppy: I believe I did this intentionally as it's the default.

osu.Framework/HostOptions.cs
peppy: I'm all for moving the name back out.

peppy:
I've made some further changes.

Note that all the empty `new HostOptions()` presence is to avoid ambiguity in ctor calls. These can be removed once the obsoleted methods are removed in the future.

bdach:
I think this diff looks fine now with the latest changes.

bdach (approved):

Susko3 (changes requested):
Also missing `// Can be removed` comments.

peppy (approved):

PR #4977, opened 2022-01-01
Allow ToolWindow to update its content layout after toolbar height changes

dekrain:
This allows for tool windows to have more buttons, as presented by adding a button for Draw Visualiser that inspects the game drawable.


dekrain:
There's a fix needed because now the content overflows under the visible area.

dekrain:
Now it should be good, but I think it would be better to put all that into a separate container class.

osu.Framework/Graphics/Visualisation/DrawVisualiser.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
```
dekrain, line 1: Not sure what happened there

Susko3: Encoding changed UTF-8 <-> UTF-8 BOM.

dekrain: Ah, makes sense

peppy:
Is this PR titled wrong? It looks to be adding a new button / feature but mentions that nowhere.

dekrain:
It's a fix that enables to add more buttons or other widgets. It wasn't possible before, because the toolbar will overflow and the tool content will overlap with buttons on the second row.

bdach:
The point is that on this PR's branch there is a button that isn't there on master:

2022-01-03-165319_585x671_scrot

Nothing in this PR thread so much as mentions why this would be added. I'd go as far as to say that I consider the usefulness of this button dubious as it is, as in many *many* hours of using the visualiser I don't recall ever having to go back to the root drawable.

If the button was added just to show that the fix works then that is not the right way to go about it.

dekrain:
The button just showcases that the fix works, yes. I'll move it to a separate tool that'd be designed for inspecting objects. I also wanted to find a cleaner way to lay out the content properly. I might do something similar to what osu!'s `SectionsContainer` does.

frenzibyte:
I think it would be best grouping this flexibility change with the feature that needs it rather than separating it to its own PR. Right now there's no benefit from this change as it stands, and, if that were to be added, we may not want the "inspect root" button to be displayed in such place to begin with. (maybe a clickable "go to root" text at the bottom, for instance).

osu.Framework/Graphics/Visualisation/ToolWindow.cs
```diff
@@ -102,16 +106,22 @@ protected void AddButton(string text, Action action)
...
+            base.UpdateAfterChildren();
+
+            float headerH = header.LayoutSize.Y;
+            if (lastHeaderHeight != headerH)
+            {
+                lastHeaderHeight = headerH;
+                var newPadding = MainHorizontalContent.Padding;
+                newPadding.Top = headerH;
+                MainHorizontalContent.Padding = newPadding;
+            }
```
bdach, line 124: Personally I'm okay with this being a standalone PR but what I'm less okay with is how this logic reads. Half-manual layouting like this shouldn't be done unless strictly necessary. Fill flows and containers are there to be used after all.

Have you tried any alternative solutions that don't do this, and if yes, do they have issues?

dekrain: The problem is that I've searched for a container that would handle it itself (I used `FillFlowContainer` originally, but it couldn't autosize the bottom part) and haven't found what I needed, and I noticed that osu `SectionsContainer` does something similar akin to this. I wouldn't be against having a similar logic in a separate container and using it instead, though.

bdach: The only reason why this sort of manual workaround exists is only because the existing layouting looks pretty weird overall. As far as I can tell it should ideally be looking something like

```
TooltipContainer
  GridContainer
    GridRow (Mode = AutoSize)
      Header
    GridRow (Mode = Distributed)
      MainHorizontalContent
```

And yes I know that `master` has a comment about `GridContainer` being unviable there. But things have changed since (that comment is almost 3 years old, and since then a significant change in invalidation handling has been merged), so I'd like confirmation that it is still unviable before proceeding.

dekrain: Fixed it.

dekrain:
I also fixed a smaller issue with the fact that in `PropertyDisplay` and `TransformDisplay` the scrollbar overflows on the bottom due to padding being applied to the scroll container.

frenzibyte:
> I also fixed a smaller issue with the fact that in `PropertyDisplay` and `TransformDisplay` the scrollbar overflows on the bottom due to padding being applied to the scroll container.

It's always best to separate irrelevant changes to their own PRs, so that they can be easily reviewed rather than mixing them together in one. But now that you've pushed it here, I suppose there's no point anymore.

osu.Framework/Graphics/Containers/ScrollContainer.cs
```diff
@@ -128,7 +128,7 @@ public bool ScrollbarOverlapsContent
         /// <summary>
         /// The maximum distance that the scrollbar can move in the scroll direction.
         /// </summary>
-        public float ScrollbarMovementExtent => Math.Max(DrawSize[ScrollDim] - Scrollbar.DrawSize[ScrollDim], 0);
+        public float ScrollbarMovementExtent => Math.Max(DrawSize[ScrollDim] - Scrollbar.DrawSize[ScrollDim] - Padding.Total[ScrollDim], 0);
```
peppy, line 131: This change needs test coverage and/or rationale. We can't go changing a central class like `ScrollContainer` without testing.

peppy (changes requested):

osu.Framework/Graphics/Containers/ScrollContainer.cs
dekrain: I merged the implementation from upstream.

peppy (approved):

PR #4979, opened 2022-01-03
Don't throw when returning a leased bindable that has already been returned

peppy:
Leases are commonly returned automatically by the framework. One such example is `ScreenStack` (https://github.com/ppy/osu-framework/blob/ceef26fa2675e89937e01b87e2ea40be83d14c58/osu.Framework/Screens/ScreenStack.cs#L349).

We likely do want this automatic cleanup, but it turns out to be very hidden to the end user. If a user attempts to clean up themselves without understanding this detail, it can lead to unexpected failures.

The example that came up is an osu!-side test failure which took 2-3 hours to hunt down the cause. Said example is very well protected against to the point I thought the issue was a threading one, but it turns out it is not: https://github.com/ppy/osu/blob/c6ca0e58956c2d64f69c2d24eef2b87eb0c76e79/osu.Game/Screens/OnlinePlay/Lounge/LoungeSubScreen.cs#L247-L251.

I don't think we were gaining anything from the `throw` this method was doing previously.

This fixes the test failure seen at https://github.com/peppy/osu/runs/4678044996?check_suite_focus=true.

### DIFF
--- a/osu.Framework/Bindables/ILeasedBindable.cs
+++ b/osu.Framework/Bindables/ILeasedBindable.cs
@@ -11,7 +11,10 @@ namespace osu.Framework.Bindables
         /// <summary>
         /// End the lease on the source <see cref="Bindable{T}"/>.
         /// </summary>
-        void Return();
+        /// <returns>
+        /// Whether the lease was returned. Will be <c>false</c> if already returned.
+        /// </returns>
+        bool Return();
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


osu.Framework/Bindables/ILeasedBindable.cs
```diff
@@ -11,7 +11,10 @@ public interface ILeasedBindable : IBindable
         /// <summary>
         /// End the lease on the source <see cref="Bindable{T}"/>.
         /// </summary>
-        void Return();
+        /// <returns>
+        /// Whether the lease was returned. Will be <c>false</c> if already returned.
```
bdach, line 15: maybe slightly clearer?

```suggestion
        /// Whether the lease was returned by this call. Will be <c>false</c> if already returned.
```

bdach (approved):
I think this makes sense but seeing that I suggested this on stream to begin with I'm probably going to try and get a third opinion just in case...?

bdach (approved):

smoogipoo (approved):
Sounds sane

PR #4980, opened 2022-01-03
Replace all usages of `Task.Wait`/`Task.Result` with safer alternatives

peppy:
As seen recently, it is quite easy in our semi-`async` world to accidentally cause a deadlock in the TPL thread pool. The cause of this is a `.Wait()` or `.Result` from within an already `async` context.

Finding these cases by reading code is quite hard, as they can be nested multiple calls deep. Likewise, I couldn't find an existing analyser that finds such cases (and to some it is considered a non-issue as the thread pool should grow to add new threads when it gets in this state - but for us this can take too long).

This is a proposal to replace all calls with a custom method with added safety. Note that this PR will intentionally fail as it points out the issues that are fixed in https://github.com/ppy/osu-framework/pull/4973.

The rationale for this change was that I couldn't cover all cases in osu!  side `TestMultiplayerClient` by reading the code alone.

### DIFF
--- a/CodeAnalysis/BannedSymbols.txt
+++ b/CodeAnalysis/BannedSymbols.txt
@@ -5,4 +5,4 @@ T:System.IComparable;Don't use non-generic IComparable. Use generic version inst
 M:System.Enum.HasFlag(System.Enum);Use osu.Framework.Extensions.EnumExtensions.HasFlagFast<T>() instead.
 F:System.UriKind.RelativeOrAbsolute;Incompatible results when run on mono (see https://www.mono-project.com/docs/faq/known-issues/urikind-relativeorabsolute/). Use Validation.TryParseUri(string, out Uri?) instead.
 M:System.Threading.Tasks.Task.Wait();Don't use Task.Wait. Use Task.WaitSafely() to ensure we avoid deadlocks.
-P:System.Threading.Tasks.Task`1.Result;Don't use Task.Result. Use Task.WaitSafelyForResult() to ensure we avoid deadlocks.
+P:System.Threading.Tasks.Task`1.Result;Don't use Task.Result. Use Task.GetCompletedResult() to safely retrieve the result of a completed task, or Task.WaitSafelyForResult() to synchronously wait for and retrieve the result of a not-completed task in a way that ensures that it will not deadlock.
--- a/CodeAnalysis/BannedSymbols.txt
+++ b/CodeAnalysis/BannedSymbols.txt
@@ -5,4 +5,4 @@ T:System.IComparable;Don't use non-generic IComparable. Use generic version inst
 M:System.Enum.HasFlag(System.Enum);Use osu.Framework.Extensions.EnumExtensions.HasFlagFast<T>() instead.
 F:System.UriKind.RelativeOrAbsolute;Incompatible results when run on mono (see https://www.mono-project.com/docs/faq/known-issues/urikind-relativeorabsolute/). Use Validation.TryParseUri(string, out Uri?) instead.
 M:System.Threading.Tasks.Task.Wait();Don't use Task.Wait. Use Task.WaitSafely() to ensure we avoid deadlocks.
-P:System.Threading.Tasks.Task`1.Result;Don't use Task.Result. Use Task.GetCompletedResult() to safely retrieve the result of a completed task, or Task.WaitSafelyForResult() to synchronously wait for and retrieve the result of a not-completed task in a way that ensures that it will not deadlock.
+P:System.Threading.Tasks.Task`1.Result;Don't use Task.Result. Use Task.WaitSafelyForResult() to ensure we avoid deadlocks.
--- a/CodeAnalysis/BannedSymbols.txt
+++ b/CodeAnalysis/BannedSymbols.txt
@@ -4,3 +4,4 @@ M:System.ValueType.Equals(System.Object)~System.Boolean;Don't use object.Equals(
 T:System.IComparable;Don't use non-generic IComparable. Use generic version instead.
 M:System.Enum.HasFlag(System.Enum);Use osu.Framework.Extensions.EnumExtensions.HasFlagFast<T>() instead.
 F:System.UriKind.RelativeOrAbsolute;Incompatible results when run on mono (see https://www.mono-project.com/docs/faq/known-issues/urikind-relativeorabsolute/). Use Validation.TryParseUri(string, out Uri?) instead.
+M:System.Threading.Tasks.Task.Wait();Don't use Task.Wait. Use Task.WaitSafely() to ensure we avoid deadlocks.
--- a/CodeAnalysis/BannedSymbols.txt
+++ b/CodeAnalysis/BannedSymbols.txt
@@ -5,4 +5,4 @@ T:System.IComparable;Don't use non-generic IComparable. Use generic version inst
 M:System.Enum.HasFlag(System.Enum);Use osu.Framework.Extensions.EnumExtensions.HasFlagFast<T>() instead.
 F:System.UriKind.RelativeOrAbsolute;Incompatible results when run on mono (see https://www.mono-project.com/docs/faq/known-issues/urikind-relativeorabsolute/). Use Validation.TryParseUri(string, out Uri?) instead.
 M:System.Threading.Tasks.Task.Wait();Don't use Task.Wait. Use Task.WaitSafely() to ensure we avoid deadlocks.
-P:System.Threading.Tasks.Task`1.Result;Don't use Task.Result. Use Task.WaitSafelyForResult() to ensure we avoid deadlocks.
+P:System.Threading.Tasks.Task`1.Result;Don't use Task.Result. Use Task.GetResultSafely() to ensure we avoid deadlocks.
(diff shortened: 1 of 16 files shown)


osu.Framework/Extensions/TaskExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+
+#pragma warning disable RS0030
+            task.Wait();
+#pragma warning restore RS0030
+        }
+
+        /// <summary>
+        /// Safe alternative to Task.Result which ensures the calling thread is not a thread pool thread.
+        /// </summary>
+        public static T WaitSafelyForResult<T>(this Task<T> task)
```
peppy, line 28: Due to the caveat mentioned below (line 31) maybe this should be called `SafelyConsumeResult` or something else?

bdach:
Needs conflict resolution against master post-merge of #4973 now.

bdach (commented):
Seems generally sane but I'm not super sure whether a runtime crash is much better than a deadlock? Going to ask for a second opinion for sure on this one.

osu.Framework/Extensions/TaskExtensions.cs
bdach: After looking at the game-side PR I'd agree with renaming. Calling `WaitSafelyForResult()` from continuations wherein the result is known to be present and that there will be no waiting involved reads really strangely otherwise.

peppy: An alternative to renaming would be to add one more property, something like `.CompletedResult` or I don't know (naming is hard) which throws in the case the task is not completed. We would use this in continuations where it was guaranteed that the result is ready.

peppy:
> Seems generally sane but I'm not super sure whether a runtime crash is much better than a deadlock? Going to ask for a second opinion for sure on this one.

Have to consider that this both:
- Make failures consistent (deadlocks can sometimes only happen one in 10 or 20 CI runs), and in many cases the banned symbol portion will heed user caution when using the replacement methods.
- Give a call stack on CI showing the exact location of the issue, something that is very hard to attain when the run ends in a deadlock/timeout.

osu.Framework/Extensions/TaskExtensions.cs
bdach: That works too. No strong preference here.

peppy:
I've had to make a fix to `VideoDecoder` that makes this PR a bit larger. Let me know if you'd prefer it is split out for review purposes.

bdach:
I'm ok with the `VideoDecoder` change being here. Not like it's a huge one anyhow.

Not sure if you wanted to go ahead with any of the [renames suggested above](https://github.com/ppy/osu-framework/pull/4980#pullrequestreview-842541841); looks fine to me otherwise.

peppy:
I've added a second method for continuation purposes. I think this feels better. Not sure about naming.

CodeAnalysis/BannedSymbols.txt
```diff
@@ -4,3 +4,5 @@ M:System.ValueType.Equals(System.Object)~System.Boolean;Don't use object.Equals(
 T:System.IComparable;Don't use non-generic IComparable. Use generic version instead.
 M:System.Enum.HasFlag(System.Enum);Use osu.Framework.Extensions.EnumExtensions.HasFlagFast<T>() instead.
 F:System.UriKind.RelativeOrAbsolute;Incompatible results when run on mono (see https://www.mono-project.com/docs/faq/known-issues/urikind-relativeorabsolute/). Use Validation.TryParseUri(string, out Uri?) instead.
+M:System.Threading.Tasks.Task.Wait();Don't use Task.Wait. Use Task.WaitSafely() to ensure we avoid deadlocks.
+P:System.Threading.Tasks.Task`1.Result;Don't use Task.Result. Use Task.WaitSafelyForResult() to ensure we avoid deadlocks.
```
bdach, line 8: Should also suggest `GetCompletedResult()` as an alternative.

```suggestion
P:System.Threading.Tasks.Task`1.Result;Don't use Task.Result. Use Task.GetCompletedResult() to safely retrieve the result of a completed task, or Task.WaitSafelyForResult() to synchronously wait for and retrieve the result of a not-completed task in a way that ensures that it will not deadlock.
```

bdach (commented):
One more nit and then looks good to go as far as I can tell :+1: Probably still going to wait for @smoogipoo's opinion on this though.

osu.Framework/Extensions/TaskExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        /// <summary>
+        /// Safe alternative to Task.Result which ensures the calling thread is not a thread pool thread.
+        /// </summary>
+        /// <remarks>
+        /// If sure that the task has completed, use <see cref="GetCompletedResult{T}"/> instead.
+        /// </remarks>
+        public static T WaitSafelyForResult<T>(this Task<T> task)
+        {
+            // We commonly access `.Result` from within `ContinueWith`, which is a safe usage (the task is guaranteed to be completed).
```
smoogipoo, line 49: It's not guaranteed to be completed. It can be failed, in which case accessing `.Result` causes an exception. In that case, the code below is going to throw the incorrect exception.

osu.Framework/Extensions/TaskExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+            // Unfortunately, the only way to allow these usages is to check whether the task is completed or not here.
+            // This does mean that there could be edge cases where this safety is skipped (ie. if the majority of executions complete
+            // immediately).
+            if (Thread.CurrentThread.IsThreadPoolThread && !task.IsCompleted)
```
smoogipoo, line 53: As per above, I suggest adding additional cases to this which bypass to the `.Result` call (such as `&& !task.IsFaulted && !task.IsFailed`).

osu.Framework/Extensions/TaskExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+#pragma warning restore RS0030
+        }
+
+        /// <summary>
+        /// Safe alternative to Task.Result which asserts that the task has already been completed (and thus will not block).
+        /// </summary>
+        /// <remarks>
+        /// If unsure that the task has completed, use <see cref="WaitSafelyForResult{T}"/> instead.
+        /// </remarks>
+        public static T GetCompletedResult<T>(this Task<T> task)
```
smoogipoo, line 31: I really don't understand the purpose of this method, and why it should ever be used instead of `WaitSafelyForResult`.

smoogipoo (changes requested):

osu.Framework/Extensions/TaskExtensions.cs
bdach: I believe that comment is actually outdated, the intention was to use `GetCompletedResult()` from within continuations.

osu.Framework/Extensions/TaskExtensions.cs
peppy: Apart from the different naming which feels better, it describes intent. This one is only to be used when the caller is 100% sure the task is completed. You can see example usages in the [osu! side PR](https://github.com/ppy/osu/pull/16309).

osu.Framework/Extensions/TaskExtensions.cs
peppy: Are you confusing `IsCompleted` for `IsCompletedSuccessfully`?

```csharp
    /// <summary>Gets a value that indicates whether the task has completed.</summary>
    /// <returns>
    /// <see langword="true" /> if the task has completed (that is, the task is in one of the three final states: <see cref="F:System.Threading.Tasks.TaskStatus.RanToCompletion" />, <see cref="F:System.Threading.Tasks.TaskStatus.Faulted" />, or <see cref="F:System.Threading.Tasks.TaskStatus.Canceled" />); otherwise, <see langword="false" />.</returns>
    public bool IsCompleted { get; }
```

smoogipoo: Indeed, I missed that.

osu.Framework/Extensions/TaskExtensions.cs
smoogipoo: Is there a point to this? It looks like a usage "error" waiting to happen.

bdach: If you're okay with always calling `WaitSafelyForResult` even inside continuations (which looks weird to me personally, given there will never be actual any waiting there) then might as well revert the addition of this method.

peppy: If we're reverting, the other one needs a better name. I'm personally not happy with calling `WaitSafely` from a continuation - it *looks* like an error.

smoogipoo: It doesn't have to be called `WaitSafelyForResult` in the first place. Why not just `GetResultSafely`? The main point is I don't want to have a ton of special cases for the user - call `WaitSafely()` instead of `Wait()` and `GetResultSafely()` instead of `Result`, that's simple to me.

bdach: `GetResultSafely()` works for me.

peppy: Will revert. I found the two methods super helpful when searching for usages to check they were correct, but I guess that's an edge case.

smoogipoo (approved):

PR #4982, opened 2022-01-05
Enable audio mixer buffering and adjust default playback buffer length

peppy:
PRing for further testing. Needs to be tested on all platforms to make sure we don't regress audio latency.

I'll provide testing methodology (for high precision testing) when I get to testing myself but welcome people to test this against platforms to see how it works / doesn't work.

Mostly aiming to fix buffer underruns on iOS. Note that this doesn't yet consider nested mixers. It's attempting to fix underruns that are currently occuring on release builds.

- [ ] Depends on #4981.


peppy:
Initial testing shows that this doesn't help much at all on iOS. The audio cutting out issue may be unrelated, or may require more settings changed. I'll experiment further.

bdach:
For what it's worth I'm willing to test linux if you give me the testing methodology, as I'm not super sure where to even begin with that.

peppy:
Not worth testing just yet, as I think we can get away without enabling in the first place (problem is in another castle for iOS). Will get some more testing in this week.

PR #4983, opened 2022-01-05
Consume touch `IsActive` state rather than checking for "any" touch for mouse mapping

peppy:
While the previous method worked fine, it means that a consumer would not be able to alter the flow of events (ie. blocking the "click" operations by changing the incoming touch).

In osu!, we wish to use this to allow the user to block the "click" portion while still convey movement of the "cursor".

Tested against iOS.

### DIFF
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -713,9 +713,9 @@ namespace osu.Framework.Input
         }
 
         /// <summary>
-        /// The number of touches which are currently down, causing a single cumulative "mouse down" state.
+        /// The number of touches which are currently active, causing a single cumulative "mouse down" state.
         /// </summary>
-        private int mouseMappedTouchesDown;
+        private readonly HashSet<TouchSource> mouseMappedTouchesDown = new HashSet<TouchSource>();
 
         /// <summary>
         /// Handles latest activated touch state change event to produce mouse input from.
@@ -735,10 +735,18 @@ namespace osu.Framework.Input
                 }.Apply(CurrentState, this);
             }
 
-            if (e.IsActive != null)
-                mouseMappedTouchesDown += (e.IsActive.Value ? 1 : -1);
+            switch (e.IsActive)
+            {
+                case true:
+                    mouseMappedTouchesDown.Add(e.Touch.Source);
+                    break;
 
-            new MouseButtonInputFromTouch(MouseButton.Left, mouseMappedTouchesDown > 0, e).Apply(CurrentState, this);
+                case false:
+                    mouseMappedTouchesDown.Remove(e.Touch.Source);
+                    break;
+            }
+
+            new MouseButtonInputFromTouch(MouseButton.Left, mouseMappedTouchesDown.Count > 0, e).Apply(CurrentState, this);
             return true;
         }
 
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -730,7 +730,7 @@ namespace osu.Framework.Input
                 }.Apply(CurrentState, this);
             }
 
-            new MouseButtonInputFromTouch(MouseButton.Left, e.State.Touch.ActiveSources.HasAnyButtonPressed, e).Apply(CurrentState, this);
+            new MouseButtonInputFromTouch(MouseButton.Left, e.IsActive != false, e).Apply(CurrentState, this);
             return true;
         }
 
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -713,6 +713,11 @@ namespace osu.Framework.Input
         }
 
         /// <summary>
+        /// The number of touches which are currently down, causing a single cumulative "mouse down" state.
+        /// </summary>
+        private int mouseMappedTouchesDown;
+
+        /// <summary>
         /// Handles latest activated touch state change event to produce mouse input from.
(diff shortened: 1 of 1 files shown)


bdach:
Test failures look relevant.

frenzibyte:
While this paves the way for disabling "mouse (from touch)" down/up input, this also changes the behaviour of mapping touches to mouse as now releasing the *latest held* touch will send a mouse up regardless of whether there are still other touches held, which makes the name of the behaviour `MapMouseToLatestTouch` no longer maps to *latest* touch.

I'm not too fussed with the behaviour change, but for the dependent case (ppy/osu#16330), we can as well avoid that by, instead, separating the touch movement and touch down/up path in `PassThroughInputManager` so that `RulesetInputManager` can early-return on them.

Though now that I'm writing this here, this will have to depend on whether native touch input should be disabled as well, which I think it should, for the case described in the issue the dependent PR fixes (keyboard for press/release & touch for movement).

```diff
diff --git a/osu.Framework/Input/PassThroughInputManager.cs b/osu.Framework/Input/PassThroughInputManager.cs
index 5b67e6654..55b9e0304 100644
--- a/osu.Framework/Input/PassThroughInputManager.cs
+++ b/osu.Framework/Input/PassThroughInputManager.cs
@@ -110,8 +110,16 @@ protected override bool Handle(UIEvent e)
                     new MouseScrollRelativeInput { Delta = scroll.ScrollDelta, IsPrecise = scroll.IsPrecise }.Apply(CurrentState, this);
                     break;

-                case TouchEvent touch:
-                    new TouchInput(touch.ScreenSpaceTouch, touch.IsActive(touch.ScreenSpaceTouch)).Apply(CurrentState, this);
+                case TouchMoveEvent touchMove:
+                    new TouchInput(touchMove.ScreenSpaceTouch, CurrentState.Touch.IsActive(touchMove.Touch.Source)).Apply(CurrentState, this);
+                    break;
+
+                case TouchDownEvent touchDown:
+                    new TouchInput(new Touch(touchDown.Touch.Source, CurrentState.Touch.TouchPositions[(int)touchDown.Touch.Source]), true).Apply(CurrentState, this);
+                    break;
+
+                case TouchUpEvent touchUp:
+                    new TouchInput(new Touch(touchUp.Touch.Source, CurrentState.Touch.TouchPositions[(int)touchUp.Touch.Source]), false).Apply(CurrentState, this);
                     break;

                 case MidiEvent midi:
@@ -173,11 +181,11 @@ protected virtual void SyncInputState(InputState state)
             var mouseButtonDifference = (state?.Mouse?.Buttons ?? new ButtonStates<MouseButton>()).EnumerateDifferen
ce(CurrentState.Mouse.Buttons);
             new MouseButtonInput(mouseButtonDifference.Released.Select(button => new ButtonInputEntry<MouseButton>(b
utton, false))).Apply(CurrentState, this);

-            new KeyboardKeyInput(state?.Keyboard?.Keys, CurrentState.Keyboard.Keys).Apply(CurrentState, this);
-
+            // same applies for touch input.
             var touchStateDifference = (state?.Touch ?? new TouchState()).EnumerateDifference(CurrentState.Touch);
             new TouchInput(touchStateDifference.deactivated, false).Apply(CurrentState, this);
-            new TouchInput(touchStateDifference.activated, true).Apply(CurrentState, this);
+
+            new KeyboardKeyInput(state?.Keyboard?.Keys, CurrentState.Keyboard.Keys).Apply(CurrentState, this);

             new JoystickButtonInput(state?.Joystick?.Buttons, CurrentState.Joystick.Buttons).Apply(CurrentState, thi
s);
             new JoystickAxisInput(state?.Joystick?.GetAxes()).Apply(CurrentState, this);
```
(diff for breaking down `TouchEvent` in `Handle()`, and also syncing touch input in a similar way to mouse so that positional data doesn't get synced as well)

peppy:
Should be resolved.

osu.Framework/Input/InputManager.cs
```diff
@@ -730,7 +735,10 @@ protected virtual bool HandleMouseTouchStateChange(TouchStateChangeEvent e)
                 }.Apply(CurrentState, this);
             }
 
-            new MouseButtonInputFromTouch(MouseButton.Left, e.IsActive != false, e).Apply(CurrentState, this);
+            if (e.IsActive != null)
+                mouseMappedTouchesDown += (e.IsActive.Value ? 1 : -1);
```
frenzibyte, line 739: `mouseMappedTouchesDown` should be clamped to 0 minimum, otherwise it will reach a negative state with the dependent PR changes (`RulesetInputManager` sending `HandleMosueTouchStateChange` with `e.IsActive = false` in all types of touch events)

or alternatively this could be fixed at `RulesetInputManager` by passing `null` to `IsActive` instead of `false` depending on whether mouse is held, and assert here that it can't get beyond zero (or beyond maximum number of touches).

frenzibyte (commented):
This could work as well. One small note.

osu.Framework/Input/InputManager.cs
peppy: I think the PR should be fixed. I'm not sure clamping here is in our best interest, might be better to throw if it's going to lead to unstable behaviour.

peppy: I've changed this once more, now to use a `HashSet` based on the `TouchSource` which should guarantee that it's very hard to get this into a bad state.

bdach (approved):
makes sense to me

PR #4987, opened 2022-01-09
Fix nested menus overflowing past screen bounds

### LINKED ISSUE #3861: Sub-menus should not overflow outside the parenting bounds
Top-level menus clamp to the sides of the parenting container, but their sub-menus do not. Example:

image

Not sure if clamping makes sense as it does for the top-level menu as opposed to it appearing on the other side.

bdach:
- [x] Depends on #4986
- Fixes #3861

https://user-images.githubusercontent.com/20418176/148696343-f1198d58-0aaa-49be-910c-c227abacf608.mp4

PRing primarily for comment.

This implementation mirrors observed behaviour of most nested menus that I could find, namely that determining whether to flip a menu or not happens locally to the given menu level (i.e. even if the parent menu is right-aligned to its parent's left side, the child menu of that can be left-aligned again if there is enough space).

The contentious part of this may be that I'm using `GetContainingInputManager()` to perform this positioning. This is done because the menu's parent is not necessarily its clipping bounds. And it's not like the menu *has* to have a parenting `ContextMenuContainer` or other drawable to work correctly. So clipping to the input manager seemed like the closest thing to "clipping to window bounds".

The implementation is a little repetitive too, with explicit branches for both orientations, but I'm not sure how to make better without making it read even worse than what's already there.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneNestedMenus.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneNestedMenus.cs
@@ -42,13 +42,24 @@ namespace osu.Framework.Tests.Visual.UserInterface
 
         private class ClickOpenMenu : BasicMenu
         {
-            protected override Menu CreateSubMenu() => new ClickOpenMenu(Direction, HoverOpenDelay, false);
+            private readonly int depth;
 
-            public ClickOpenMenu(Direction direction, double timePerAction, bool topLevel = true)
-                : base(direction, topLevel)
+            protected override Menu CreateSubMenu() => new ClickOpenMenu(Direction, HoverOpenDelay, depth + 1);
+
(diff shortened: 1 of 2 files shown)


peppy:
I haven't gone through the logic changes in this PR yet, but on initial testing noticed the editor menu bar is no longer aligned correctly:

20220113 122346 (osu!)

bdach:
Believe the above should be fixed with the latest commit.

Note that both direction branches in the layout code have now become extremely similar; the differences are very subtle (note where `Draw{Width,Height}` are added, and where they aren't). This may be promising from the standpoint of reducing the amount of code, but I'm not sure if the result won't be too inscrutable. I don't even currently have the brainpower to figure out how I could unify the branches or what the actual difference is. I'll revisit with a fresher mind.

peppy:
Looks to be working better now.

peppy (approved):

PR #4988, opened 2022-01-10
Allow waiting on tasks which were created with `LongRunning` regardless of thread

peppy:
Closes https://github.com/ppy/osu/issues/16396#issuecomment-1008687513.

Not 100% sure on this since long running tasks shouldn't really be run within `Task.Run` in the first place, but let's make things work again for the time being.

### DIFF
--- a/osu.Framework.Tests/IO/TestWebRequest.cs
+++ b/osu.Framework.Tests/IO/TestWebRequest.cs
@@ -11,6 +11,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 using NUnit.Framework;
+using osu.Framework.Extensions;
 using osu.Framework.Graphics;
 using osu.Framework.IO.Network;
 using WebRequest = osu.Framework.IO.Network.WebRequest;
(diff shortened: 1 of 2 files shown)


bdach (approved):
I *think* this makes sense. Tested this with game at least and the exceptions appear to be gone.

smoogipoo:
I don't think this does anything?

```
Task.Factory.StartNew(() =>
{
    Console.WriteLine(Thread.CurrentThread.IsThreadPoolThread);
}, TaskCreationOptions.LongRunning).Wait();

> False
```

peppy:
That's correct? It should be false in that case. `LongRunning` forces the task to not run on the thread pool.

smoogipoo:
Right, but the checks in the previous code were `if (Thread.CurrentThread.IsThreadPoolThread) throw X;`, which wouldn't have occurred anyway. Am I missing something?

peppy:
The added check is checking against the outer *task* itself. So in your example, the `.Wait` call (which would be `.WaitSafely` in our case). This PR allows `.WaitSafely` on the *task* that is defined as `LongRunning` to not throw.

smoogipoo:
Aha, thanks.

smoogipoo (approved):

PR #4989, opened 2022-01-10
Add support for Android pointer capture

### LINKED ISSUE #4373: Android built-in cursor displayed on top of the game's own cursor

Susko3:
Closes #4373.

https://developer.android.com/training/gestures/movement#pointer-capture

>The system can take the capture away from the view without you explicitly calling `releasePointerCapture()`, most commonly because the view hierarchy that contains the view that requested capture has lost focus.

We don't have to handle disabling capture ourselves when we lose focus, Android does it for us.

Unfortunately, Android's focus changed events aren't very useful. We can lose focus and get focus without any events being sent.
Example: Pulling down the status bar. Since the status bar is a system overlay/dialog, Android doesn't consider it as changing focus for this event, but does forcefully disabled pointer capture,

So we update capture on every non-relative mouse move event.

**Changes:**

`CursorState` was made to always be confined on Android. This is needed so that the cursor can't leave window bounds. Leaving window bounds doesn't make sense in fullscreen mobile apps.

**Issues:**

The only problem here is that we currently can't have a settings section for `AndroidMouseHandler` in osu!, as the android project isn't referenced. Since `InputHandler`s are supposed to be platform specific, we should create a way to do that.

### DIFF
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Android.Content;
+using Android.OS;
 using Android.Runtime;
 using Android.Text;
 using Android.Util;
@@ -16,6 +17,31 @@ namespace osu.Framework.Android
     public class AndroidGameView : osuTK.Android.AndroidGameView
     {
         public AndroidGameHost Host { get; private set; }
+
+        /// <summary>
+        /// Represents whether the mouse pointer is captured, as reported by Android through <see cref="OnPointerCaptureChange"/>.
+        /// </summary>
+        private bool pointerCaptured;
+
+        /// <summary>
+        /// Set Android's pointer capture.
+        /// </summary>
+        public bool PointerCapture
+        {
+            get => pointerCaptured;
+            set
+            {
+                // Pointer capture is only available on Android 8.0 and up
+                if (Build.VERSION.SdkInt < BuildVersionCodes.O) return;
+
+                if (pointerCaptured == value) return;
+
+                if (value)
+                    RequestPointerCapture();
+                else
+                    ReleasePointerCapture();
+            }
+        }
 
         private readonly Game game;
 
@@ -96,6 +122,12 @@ namespace osu.Framework.Android
         {
             KeyUp?.Invoke(keyCode, e);
             return true;
+        }
+
+        public override void OnPointerCaptureChange(bool hasCapture)
+        {
+            base.OnPointerCaptureChange(hasCapture);
+            pointerCaptured = hasCapture;
         }
 
         protected override void OnLoad(EventArgs e)
--- a/osu.Framework.Android/AndroidGameWindow.cs
+++ b/osu.Framework.Android/AndroidGameWindow.cs
@@ -24,6 +24,21 @@ namespace osu.Framework.Android
             set { }
         }
 
+        public event Action CursorStateChanged;
+
+        private CursorState cursorState = CursorState.Confined;
+
+        public override CursorState CursorState
+        {
+            get => cursorState;
+            set
+            {
+                // cursor should always be confined on mobile platforms, to have UserInputManager confine the cursor to window bounds
+                cursorState = value | CursorState.Confined;
+                CursorStateChanged?.Invoke();
+            }
+        }
+
         public AndroidGameWindow(AndroidGameView view)
             : base(view)
         {
--- a/osu.Framework.Android/Input/AndroidMouseHandler.cs
+++ b/osu.Framework.Android/Input/AndroidMouseHandler.cs
@@ -2,7 +2,10 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Collections.Generic;
+using Android.OS;
 using Android.Views;
+using osu.Framework.Bindables;
+using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Input.StateChanges;
 using osu.Framework.Platform;
 using osu.Framework.Statistics;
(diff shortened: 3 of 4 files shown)


bdach:
> The only problem here is that we currently can't have a settings section for `AndroidMouseHandler` in osu!, as the android project isn't referenced.

Just to note: `osu.Game` should not reference the android project. The desired directionality would be the other way: `osu.Game` should expose an extension point for input handler configuration that platform projects may use to register their own config sections.

osu.Framework.Android/AndroidGameView.cs
```diff
@@ -17,6 +18,31 @@ public class AndroidGameView : osuTK.Android.AndroidGameView
...
+        /// <summary>
+        /// Set Android's pointer capture.
+        /// </summary>
+        public bool PointerCapture
+        {
+            get => pointerCaptured;
+            set
+            {
+                // Pointer capture is only available on Android 8.0 and up
+                if (Build.VERSION.SdkInt < BuildVersionCodes.O) return;
```
bdach, line 35: Maybe this should be logged as important at least, to reduce confusion for users that aren't on a high enough SDK version?

The fact that this support is >=8.0 should definitely be mentioned in xmldoc though.

osu.Framework.Android/AndroidGameWindow.cs
```diff
@@ -24,6 +24,21 @@ public override Platform.WindowState WindowState
...
+        public override CursorState CursorState
+        {
+            get => cursorState;
+            set
+            {
+                // cursor should always be confined on mobile platforms, to have UserInputManager confine the cursor to window bounds
+                cursorState = value | CursorState.Confined;
+                CursorStateChanged?.Invoke();
+            }
+        }
```
bdach, line 40: Why is there a second backing field here rather than delegating get/set to base? The setter in particular is bypassing a bunch of osuTK logic by not calling base.

osu.Framework.Android/Input/AndroidMouseHandler.cs
```diff
@@ -114,11 +202,27 @@ private void handleMouseMoveEvent(MotionEvent evt)
             for (int i = 0; i < evt.HistorySize; i++)
                 handleMouseMove(new Vector2(evt.GetHistoricalX(i), evt.GetHistoricalY(i)));
 
-            handleMouseMove(new Vector2(evt.RawX, evt.RawY));
+            handleMouseMove(new Vector2(evt.GetX(), evt.GetY()));
```
bdach, line 205: What's this change about? Was `Raw{X,Y}` just wrong?

bdach (commented):
Seems mostly fine on a skim but I can't test due to lack of hardware.

osu.Framework.Android/Input/AndroidMouseHandler.cs
Susko3: Raw values were wrong, the cursor would end up in the wrong location when using split-screen. `Get{X,Y}()` translates to local coordinates and is what we want. `GetHistoricalX()` behaves the same as `GetX()` as well.

Susko3:
> > The only problem here is that we currently can't have a settings section for `AndroidMouseHandler` in osu!, as the android project isn't referenced.
> 
> Just to note: `osu.Game` should not reference the android project. The desired directionality would be the other way: `osu.Game` should expose an extension point for input handler configuration that platform projects may use to register their own config sections.

Yeah, that's what I meant, should have said "project *can't* be referenced".

Is moving [this](https://github.com/ppy/osu/blob/e333831c453b881972e0d4c4ac988bba13c4d6e5/osu.Game/Overlays/Settings/Sections/InputSection.cs#L55-L81) to a virtual method in `OsuGameBase` an alright solution for you? We could have the platform-specific overrides in `OsuGame{Desktop/Android/etc.}`.
This does make it a bit problematic for tests, as `OsuTestBrowser` game is used for tests for all platforms.

bdach:
> Is moving this to a virtual method in `OsuGameBase` an alright solution for you? We could have the platform-specific overrides in `OsuGame{Desktop/Android/etc.}`.

Sounds workable in theory. Not sure if there are any other issues with that other than tests.

peppy:
Of note, iOS also has similar functionality in the last release (it bypasses acceleration as well). We probably would want to implement it there too.

Susko3:
While the handlers match 1:1, the settings do not.
Confine mode doesn't make sense on mobile, and we'd need to have a warning for relative mode (or hide the setting altogether) when the user has an unsupported android version.

bdach:
As far as I'm concerned this diff looks about ready, but I have no actual means to test and am still sort of interested in how game side integration of this would look. So not sure whether to wait for that part before approve/merge.

Susko3:
Also we may want to hold off until #4950.

Could do a hacky thing and save in `OnPause()`, just before suspending.

https://github.com/ppy/osu-framework/blob/5eae03a179b3337445d2885c9fd1937e35682db8/osu.Framework.Android/AndroidGameActivity.cs#L96-L101

peppy (approved):

PR #4991, opened 2022-01-13
Output performance log messages when schedulers have long queues

peppy:
This intends to expose cases of schedulers being enqueued to but not running, which is something that has come up in the past and can cause unforeseen overheads or memory leaks.

Would be nice if we could better identify the owner of the scheduler, but not sure it's worth passing a reference in to the drawable or not. Can be added if that's seen as a requirement.

### DIFF
--- a/osu.Framework.Tests/Threading/SchedulerTest.cs
+++ b/osu.Framework.Tests/Threading/SchedulerTest.cs
@@ -4,6 +4,7 @@
 using System;
 using NUnit.Framework;
 using osu.Framework.Logging;
+using osu.Framework.Testing;
 using osu.Framework.Threading;
 using osu.Framework.Timing;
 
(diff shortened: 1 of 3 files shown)


osu.Framework/Threading/Scheduler.cs
```diff
@@ -276,7 +277,11 @@ public void Add([NotNull] ScheduledDelegate task)
                 throw new InvalidOperationException($"Can not add a {nameof(ScheduledDelegate)} that has been already {nameof(ScheduledDelegate.Completed)}");
 
             lock (queueLock)
+            {
                 timedTasks.AddInPlace(task);
+                if (timedTasks.Count % LOG_EXCESSSIVE_QUEUE_LENGTH_INTERVAL == 0)
+                    Logger.Log($"{nameof(Scheduler)} has {timedTasks.Count} timed tasks pending", LoggingTarget.Performance);
```
bdach, line 283: Similar log statement is in `enqueue()`, but one prints `nameof(Scheduler)` and one prints `this.ToString()`. Is this intentional?

osu.Framework/Logging/Logger.cs
```diff
@@ -83,6 +83,8 @@ public static Storage Storage
         /// </summary>
         public string Filename => $@"{Name}.log";
 
+        public int TotalLogOperations => logCount.Value;
```
bdach, line 86: Looks unused.

bdach:
Tests are failing because they happen to run after `TestLogging`, and one of those tests is setting `Logger.Enabled = false`.

osu.Framework/Logging/Logger.cs
peppy: Yeah it ended up being. Left it here since it might be useful in the future, but can remove instead of we don't want it.

osu.Framework/Threading/Scheduler.cs
```diff
@@ -143,15 +146,15 @@ private void queueTimedTasks()
 
                         if (sd.RepeatInterval > 0)
                         {
-                            if (timedTasks.Count > 1000)
+                            if (timedTasks.Count > LOG_EXCESSSIVE_QUEUE_LENGTH_INTERVAL)
```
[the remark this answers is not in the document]
peppy: which part? the reduction to 100?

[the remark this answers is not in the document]
smoogipoo: Correct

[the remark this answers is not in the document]
peppy: So 1000 for both cases?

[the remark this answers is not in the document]
smoogipoo: Sure, or have this be its own thing.

osu.Framework/Logging/Logger.cs
bdach: Can stay, was just wondering whether that was intentional.

osu.Framework.Tests/Threading/SchedulerTest.cs
```diff
@@ -21,6 +22,94 @@ public void Setup()
...
+            for (int i = 0; i < Scheduler.LOG_EXCESSSIVE_QUEUE_LENGTH_INTERVAL; i++)
+            {
+                scheduler.Add(() => { });
+                if (withFlushing) scheduler.Update();
+            }
+
+            Assert.AreEqual(withFlushing ? 0 : 2, matchingLogCount);
+
+            Logger.NewEntry -= logTest;
+            Logger.Enabled = false;
```
bdach, line 60: Isn't this kinda dicey? It's going to be disabling all logging which is enabled by default.

If anything the logger test was in the wrong by polluting the tests' logging configuration with a non-default value in the end...?

peppy: I'm just doing what the other logger tests already do. Logging is enabled by `GameHost` which should handle most cases we'd expect it to be outputting:

https://github.com/ppy/osu-framework/blob/97a12382e7b68b571c13885fbb2d2fe7cbc8e0d0/osu.Framework/Platform/GameHost.cs#L842

bdach: But the game host is initialised once per full test run, isn't it? If it wasn't and it was being restored before every test, then there would be no issue with what the logger tests were doing, as in disabling the logging and causing other tests to fail?

peppy: Bit confused. Is it not once per test class? That's how they get their own directories and stuff.

bdach: The scheduler tests don't use game hosts at all. They're just plain nunit fixtures. So if those tests run after the logger tests, which (mistakenly) expect that they can set `Logger.Enabled` to false because every game host will re-enable logging, then every test that's not run under a game host and is just a plain nunit fixture will have logging disabled (until someone creates another game host, because the field is process-global).

peppy: Sure, but that's how it's been until now? `GameHost` also disabled logging on exit.

Also logging only works with a storage, so outside of a `GameHost` its usage seems undefined in the first place?

bdach: If logging outside of a host is to be considered undefined behaviour then all of the logic inside of `TestLogOutputFromManyQueuedTasks` and `TestLogOutputFromManyQueuedScheduledTasks` would need to run under a game host explicitly created as part of the test, correct? Which it isn't right now.

peppy: Except Logger.Enabled is being called here, making it work.

bdach: I suppose so. Note, however, that the scheduler test isn't setting `Logger.Storage`, which means it silently either doesn't use any storage (if the test is ran in isolation) or uses the last set storage. It has no effect on the outcome of the test due to how it is written but maybe a bit ehh. Not sure, I don't want to prolong this conversation.

peppy: Can set to `null` here (alongise the `Enabled=true` part) if that reads better

bdach: Sure, I guess I'd be fine with that.

osu.Framework.Tests/Threading/SchedulerTest.cs
```diff
@@ -21,6 +23,106 @@ public void Setup()
...
+                }
+            }
+        }
+
+        [Test]
+        public void TestLogOutputFromManyQueuedScheduledTasks([Values(false, true)] bool withFlushing)
+        {
+            int matchingLogCount = 0;
+
+            using (var storage = new TemporaryNativeStorage(nameof(TestLogOutputFromManyQueuedTasks)))
```
smoogipoo, line 81: Wrong method reference.

smoogipoo (changes requested):

smoogipoo (approved):
Looks fine to me

PR #4992, opened 2022-01-14
Change `[MeansImplicitUse]` attribute on BDL so that unused params are treated as such

Susko3:
Adds `ImplicitUseKindFlags.InstantiatedNoFixedConstructorSignature` to `[MeansImplicitUse]` attribute for BDL.
The flag does exactly what we'd want, the BDL method itself is treated as used implicitly, but the params are not.

image

"More" info on jetbrains' site:
https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#ImplicitUseKindFlags

Discussed shortly on [discord](https://discord.com/channels/188630481301012481/589331078574112768/931314763718340628).


peppy (dismissed):

peppy:
Seems to work well. I've increased the inspection level for this to warning because it seems like something we should always catch framework-side.

Open to opinion on whether we should make the same change on the osu! side.

peppy (approved):

peppy:
Scratch that, turns out CI sees a lot more than my IDE does. Let's leave it as a hint for now.

peppy (approved):

peppy (approved):

PR #4993, opened 2022-01-18
Update OpenTabletDriver NuGet package to v0.6.0.2

InfinityGhost:
Updated the OpenTabletDriver NuGet package to v0.6.0.2

Multiple tablets are supported in this version of the package, however this is only set up to use the first detected tablet to match the previous version's implementation.


osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
```diff
@@ -49,36 +50,20 @@ public void DetectTablet()
...
-        }
-
-        private TabletConfiguration deserialize(Stream stream)
-        {
-            using (var reader = new StreamReader(stream))
-            using (var jsonReader = new JsonTextReader(reader))
-                return new JsonSerializer().Deserialize<TabletConfiguration>(jsonReader);
+                // ideally this would return if a tablet was detected, however it is not required.
+                // it is only used a hint that one or tablets were detected to be used by inheritors.
+                return true;
```
bdach, line 65: If the return value of this is mostly to be used by inheritors then maybe best to seal this class for the time being (as the implication is that nothing should inherit it, so there's no need to use an actual value here)?

That said, I think the return value of `base.Detect()` a bit higher up could be used as well, and would be the proper one? Not sure it's that much effort to just use that.

osu.Framework.Benchmarks/BenchmarkTabletDriver.cs
```diff
@@ -13,13 +15,18 @@ public class BenchmarkTabletDriver : BenchmarkTest
 
         public override void SetUp()
         {
-            driver = new TabletDriver();
+            var collection = new DriverServiceCollection()
+                .AddTransient<TabletDriver>();
+
+            var serviceProvider = collection.BuildServiceProvider();
+
+            driver = serviceProvider.GetRequiredService<TabletDriver>();
```
bdach, line 23: This pattern shows up a lot, to kind of jerry-rig Microsoft-provided DI to work. Maybe this could be extracted to a `public static TabletDriver Create()` in `TabletDriver`, though? You'd then be able to just call `TabletDriver.Create()` everywhere.

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
```diff
@@ -32,59 +42,79 @@ public class OpenTabletDriverHandler : InputHandler, IAbsolutePointer, IVirtualT
...
 
-            Enabled.BindValueChanged(d =>
-            {
-                if (d.NewValue)
+                            updateInputArea(device);
+                            updateOutputArea(host.Window);
+
+                            Debug.Assert(device != null);
+                            foreach (var endpoint in device.InputDevices)
+                                endpoint.Report += handleDeviceReport;
```
bdach, line 77: Will these subscriptions be correctly removed when the input handler is disabled/re-enabled a second time? I don't see an explicit unsubscription path.

osu.Framework.Benchmarks/BenchmarkTabletDriver.cs
InfinityGhost: Agreed, will do.

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
InfinityGhost: It is not worth the effort to pull the actual value from the above task. Sealing the class is likely the better move here.

osu.Framework.Benchmarks/BenchmarkTabletDriver.cs
bdach: `BenchmarkTabletDriver` pointed out in this conversation is still manually doing the `ServiceProvider` song and dance rather than using `TabletDriver.Create()`.

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
```diff
@@ -4,39 +4,43 @@
...
 
 namespace osu.Framework.Input.Handlers.Tablet
 {
-    public class TabletDriver : Driver
+    public sealed class TabletDriver : Driver
     {
         private static readonly IEnumerable<int> known_vendors = Enum.GetValues<DeviceVendor>().Cast<int>();
 
-        public TabletDriver()
+        public TabletDriver([NotNull] ICompositeDeviceHub deviceHub, [NotNull] IReportParserProvider reportParserProvider, [NotNull] IDeviceConfigurationProvider configurationProvider)
```
bdach, line 25: Now that `TabletDriver.Create()` exists, thoughts on privatising this ctor to boot? I'm not sure there is need for it to be public, and having a single point of entry will avoid confusing any further users (if any).

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
```diff
@@ -57,6 +60,19 @@ public override bool Detect()
...
+                        foreach (var device in InputDevices)
+                        {
+                            foreach (var endpoint in device.InputDevices)
+                            {
+                                endpoint.Report += DeviceReport;
+                                endpoint.ConnectionStateChanged += (sender, connected) =>
+                                {
+                                    if (!connected)
+                                        endpoint.Report -= DeviceReport;
+                                };
```
bdach, line 73: What's the lifetime of the `endpoint` object here? I'm asking because I'm a bit worried about potentially subscribing multiple times to `ConnectionStateChanged`, as it can be called multiple times via toggling `OpenTabletDriverHandler.Enabled` off and on. The inner unsubscription of `DeviceReport` should probably also contain a self-cleanup of the `ConnectionStateChanged` callback itself?

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
peppy: DI constructors have to be public, so we can't change this one unfortunately.

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
peppy: Checking the underlying code, it does look safe (once a device enters a `Connected==false` state it will never return to `true`).

peppy: Which leaves me questioning the value of the unbind of `endpoint.Report` then. Seems like we could just leave it to die alongside reference loss?

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
```diff
@@ -4,39 +4,43 @@
...
         }
 
         private readonly object detectLock = new object();
 
         private CancellationTokenSource cancellationSource;
 
-        public void DetectTablet()
+        public event EventHandler<IDeviceReport> DeviceReported;
+
+        public override bool Detect()
```
peppy, line 43: This scares me beyond belief. This is a method exposed by OTD, which we are overriding and the delegating the `base` work *to another thread*. Was this your intention? I don't think this can be how we do things.

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
```diff
@@ -32,59 +40,71 @@ public class OpenTabletDriverHandler : InputHandler, IAbsolutePointer, IVirtualT
...
-                switch (report)
+                if (d.NewValue && tabletDriver == null)
                 {
-                    case ITabletReport tabletReport:
-                        handleTabletReport(tabletReport);
-                        break;
+                    tabletDriver = TabletDriver.Create();
+                    tabletDriver.TabletsChanged += (s, e) =>
+                    {
+                        device = e.Any() ? tabletDriver.InputDevices.First() : null;
```
peppy, line 58: What's a `TreeList` vs a `Tree`? By only handling the first "tree" in the "tree list" here are we potentially missing devices?

peppy (changes requested):

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
InfinityGhost: As bad as this looks being that we're overriding it, the returned value is just a hint that doesn't actually get used for any logic at any point except if desired by an implementer such as the OpenTabletDriver desktop program and osu! framework. We plan on removing it and integrating this optimization so this isn't necessary in v0.7. We could make this method sync then just call a Task that does this in the background instead if that's more preferable.

osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
InfinityGhost: This was some naming mess that we didn't end up cleaning up like we intended, however in future versions (v0.7 and onward) it'll be referred changed from  `InputDeviceTree` -> `InputDevice` for the actual device and `InputDevice` `InputDeviceEndpoint`. `InputDeviceTreeList` is just a `Collection<InputDevice>` that disposes devices between detects and it will also be renamed in the future.

InfinityGhost: Forgot to answer the rest of your question here, but no we are not missing devices this way.

osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
peppy: I'm still not too sure why this is required in the first place. Would the following patch not be an amicable solution:

```diff
diff --git a/osu.Framework/Input/Handlers/Tablet/TabletDriver.cs b/osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
index e7c634f15..f31281f45 100644
--- a/osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
+++ b/osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
@@ -22,63 +22,53 @@ public sealed class TabletDriver : Driver
     {
         private static readonly IEnumerable<int> known_vendors = Enum.GetValues<DeviceVendor>().Cast<int>();
 
+        private CancellationTokenSource cancellationSource;
+
+        public event EventHandler<IDeviceReport> DeviceReported;
+
         public TabletDriver([NotNull] ICompositeDeviceHub deviceHub, [NotNull] IReportParserProvider reportParserProvider, [NotNull] IDeviceConfigurationProvider configurationProvider)
             : base(deviceHub, reportParserProvider, configurationProvider)
         {
             Log.Output += (sender, logMessage) => Logger.Log($"{logMessage.Group}: {logMessage.Message}", level: (LogLevel)logMessage.Level);
+
             deviceHub.DevicesChanged += (sender, args) =>
             {
                 // it's worth noting that this event fires on *any* device change system-wide, including non-tablet devices.
                 if (!Tablets.Any() && args.Additions.Any())
-                    Detect();
+                {
+                    cancellationSource?.Cancel();
+                    cancellationSource = new CancellationTokenSource();
+
+                    Task.Run(() => detectAsync(cancellationSource.Token), cancellationSource.Token);
+                }
             };
         }
 
-        private readonly object detectLock = new object();
-
-        private CancellationTokenSource cancellationSource;
+        private async Task detectAsync(CancellationToken cancellationToken)
+        {
+            // wait a small delay as multiple devices may appear over a very short interval.
+            await Task.Delay(50, cancellationToken).ConfigureAwait(false);
 
-        public event EventHandler<IDeviceReport> DeviceReported;
+            int foundVendor = CompositeDeviceHub.GetDevices().Select(d => d.VendorID).Intersect(known_vendors).FirstOrDefault();
 
-        public override bool Detect()
-        {
-            lock (detectLock)
+            if (foundVendor > 0)
             {
-                cancellationSource?.Cancel();
+                Logger.Log($"Tablet detected (vid{foundVendor}), searching for usable configuration...");
 
-                var cancellationToken = (cancellationSource = new CancellationTokenSource()).Token;
+                Detect();
 
-                Task.Run(async () =>
+                foreach (var device in InputDevices)
                 {
-                    // wait a small delay as multiple devices may appear over a very short interval.
-                    await Task.Delay(50, cancellationToken).ConfigureAwait(false);
-
-                    int foundVendor = CompositeDeviceHub.GetDevices().Select(d => d.VendorID).Intersect(known_vendors).FirstOrDefault();
-
-                    if (foundVendor > 0)
+                    foreach (var endpoint in device.InputDevices)
                     {
-                        Logger.Log($"Tablet detected (vid{foundVendor}), searching for usable configuration...");
-
-                        base.Detect();
-
-                        foreach (var device in InputDevices)
+                        endpoint.Report += DeviceReported;
+                        endpoint.ConnectionStateChanged += (sender, connected) =>
                         {
-                            foreach (var endpoint in device.InputDevices)
-                            {
-                                endpoint.Report += DeviceReported;
-                                endpoint.ConnectionStateChanged += (sender, connected) =>
-                                {
-                                    if (!connected)
-                                        endpoint.Report -= DeviceReported;
-                                };
-                            }
-                        }
+                            if (!connected)
+                                endpoint.Report -= DeviceReported;
+                        };
                     }
-                }, cancellationToken);
-
-                // ideally this would return if a tablet was detected, however it is not required.
-                // it is only used a hint that one or tablets were detected to be used by inheritors.
-                return true;
+                }
             }
         }
```

InfinityGhost: Would work just fine, for some reason that didn't cross my head at the time of writing that code.

peppy (approved):
Seems to work fine with my two test devices.

PR #4997, opened 2022-01-21
Enable class sealing inspections

peppy:
Applied to `private` methods. Went through all non-private inspections and didn't find any immediate wins.

Quite a few cases I think we could change classes to `internal` or `sealed` but not interested in triggering discussion at this point in time. Let's wait for .NET6/7.


frenzibyte:
Seems CI is showing a heap of other private classes that have not been sealed yet.

peppy:
Probably pointless for now.

PR #4999, opened 2022-01-24
Add missing thread safety check in `LoadComponentsAsync`

peppy:
Have run against osu! tests suite, looks okay.

### DIFF
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -143,7 +143,6 @@ namespace osu.Framework.Graphics.Containers
             if (game == null)
                 throw new InvalidOperationException($"May not invoke {nameof(LoadComponentAsync)} prior to this {nameof(CompositeDrawable)} being loaded.");
 
-            ThreadSafety.IsUpdateThread = false;
             EnsureMutationAllowed($"load components via {nameof(LoadComponentsAsync)}");
 
             if (IsDisposed)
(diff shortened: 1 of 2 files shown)


smoogipoo (approved):

osu.Framework/Graphics/Containers/CompositeDrawable.cs
```diff
@@ -143,6 +143,8 @@ protected internal Task LoadComponentAsync<TLoadable>([NotNull] TLoadable compon
             if (game == null)
                 throw new InvalidOperationException($"May not invoke {nameof(LoadComponentAsync)} prior to this {nameof(CompositeDrawable)} being loaded.");
 
+            EnsureMutationAllowed(nameof(LoadComponentsAsync));
```
bdach, line 146: The exception message that results from this is a little funny:

> Cannot mutate the LoadComponentsAsync of a Loaded Drawable while not on the update thread. Consider using Schedule to schedule the mutation operation.

Probably could use a bit more polish.

peppy: Should be better now, well spotted.

osu.Framework/Graphics/Containers/CompositeDrawable.cs
```diff
@@ -143,6 +143,9 @@ protected internal Task LoadComponentAsync<TLoadable>([NotNull] TLoadable compon
             if (game == null)
                 throw new InvalidOperationException($"May not invoke {nameof(LoadComponentAsync)} prior to this {nameof(CompositeDrawable)} being loaded.");
 
+            ThreadSafety.IsUpdateThread = false;
```
smoogipoo, line 146: Unintended, yeah?

smoogipoo (changes requested):

osu.Framework/Graphics/Containers/CompositeDrawable.cs
peppy: 😳

smoogipoo (approved):

PR #5003, opened 2022-01-26
Add ability to control console output logging

dekrain:
It introduces a switch that allows to force enable or force disable logging output to console, regardless whether it's a debug build.

On osu! side it will introduce a control in debug settings section.

### DIFF
--- a/osu.Framework/Configuration/FrameworkConfigManager.cs
+++ b/osu.Framework/Configuration/FrameworkConfigManager.cs
@@ -22,6 +22,7 @@ namespace osu.Framework.Configuration
         protected override void InitialiseDefaults()
         {
             SetDefault(FrameworkSetting.ShowLogOverlay, false);
+            SetDefault(FrameworkSetting.LogConsoleOutput, Logging.LogConsoleOutputSetting.Default);
 
             SetDefault(FrameworkSetting.WindowedSize, new Size(1366, 768), new Size(640, 480));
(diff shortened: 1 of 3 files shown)


frenzibyte:
While this PR on its own doesn't look harmful, I'm curious to hear what you require this for? reading logs from the `logs` folder of the game is always better than console output anyways. Console output is usually for quick log checking during debug builds, and may eventually be limited when too much is outputted to console.

dekrain:
For example, when an error occurs, it's easier to just look at console output instead of opening the logs every time

peppy:
It this something you plan to use in one of your own projects?

dekrain:
In nothing specific, except for adding a corresponding control in osu settings. It's generally to aid help in debugging whenever something happens in Release versions.

peppy:
I'm not sure about adding this in osu!, which is why I am asking. Console output can add performance concerns (most consoles use GPU rendering, so if output is high it may lead to user issues which they then pin on us).

Will wait for more opinions if that is the goal.

dekrain:
> Console output can add performance concerns (most consoles use GPU rendering, so if output is high it may lead to user issues which they then pin on us)

That's the reason it can be toggled, with `Default` value being the default, which acts like the original implementation does.
If you don't want to add this to osu!, I'm fine with that, just asking for opinion.

peppy:
If anything, I'd probably see this as command line setting. No need for it to be in the config. Even if it was in the config, a UI option like this would strictly not be added because it is confusing to a user and not applicable to the 99.9999999%.

PR #5005, opened 2022-01-29
Update SDL

peppy:


peppy:
Blocking until we understand whether this will regress relative mouse support (see https://github.com/libsdl-org/SDL/issues/5271).

Will need to check github action build logs for which commit the last SDL2-CS package was built against.

peppy:
Was already discovered that this has regressed in live builds, so I think we may as well push this one out.

peppy:
@smoogipoo can you give this a quick run on linux and make sure nothing serious has exploded?

smoogipoo:
Seems fine to me.

smoogipoo (approved):

PR #5006, opened 2022-01-31
Use `for` loop instead of `foreach` for critically hot path in invalidation

peppy:
Usually we'd avoid this level of micro optimisation, but this was showing up when assessing remaining song select performance concerns and seems like a no-brainer fix (the list is safe as it cannot be used post-load).

If this requires any thought beyond a quick merge then closing is fine. Had this left over from a profiling session.

unknown

(removes the `MoveNext` overhead at 18% of invalidation flow)


smoogipoo (approved):
As discussed, I'm okay with super micro-optimising hot paths like this in o-f.

PR #5010, opened 2022-02-04
Improve `SafeAreaDefiningContainer` bindable callback thread safety

frenzibyte:
As mentioned in https://github.com/ppy/osu/pull/16772#discussion_r799393996. Changes from the `host.Window.SafeAreaPadding` bindable are scheduled to the update thread instead of unsafely propagating them down the drawable hierarchy (also moved to `LoadComplete` becuase we probably don't want to bind on BDL?).

### DIFF
--- a/osu.Framework/Graphics/Containers/SafeAreaDefiningContainer.cs
+++ b/osu.Framework/Graphics/Containers/SafeAreaDefiningContainer.cs
@@ -35,12 +35,25 @@ namespace osu.Framework.Graphics.Containers
             }
         }
 
-        [BackgroundDependencyLoader]
-        private void load(GameHost host)
+        [Resolved]
+        private GameHost host { get; set; }
+
+        private IBindable<MarginPadding> hostSafeArea;
+
+        protected override void LoadComplete()
         {
-            if (!usesCustomBinding && host.Window != null)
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Containers/SafeAreaDefiningContainer.cs
```diff
@@ -35,13 +35,26 @@ public SafeAreaDefiningContainer(BindableSafeArea safeArea = null)
             }
         }
 
-        [BackgroundDependencyLoader]
-        private void load(GameHost host)
+        [Resolved]
+        private GameHost host { get; set; }
+
+        private IBindable<MarginPadding> hostSafeArea;
```
bdach, line 41: Feels a bit risky letting this be `null` sometimes (see early guard in `LoadComplete()`). I'd probably always create this bindable and use `BindTo()` instead? Or at least annotate this as `[CanBeNull]` or similar.

frenzibyte: I think `[CanBeNull]` would suffice.

bdach (approved):

peppy (approved):

PR #5011, opened 2022-02-04
Remove no longer required SDL2 P/Invokes in Linux clipboard

frenzibyte:
The code existed in late 2018, way back before SDL2 was fully implemented to osu!framework. Seems it can now directly use the `SDL2` wrapper and simplify linux clipboard code.

Would appreciate testing this as I'm slightly hesitant about this given it's linux but I'm not sure about just leaving it behind.


bdach:
While I don't particularly see a reason not to do this, I do wonder whether windows and macOS need to have native implementations of these rather than using `SDL.SDL_{Get,Set}Clipboard()` too.

Have not tested yet.

bdach:
Tested, looks to work as expected.

The remaining native implementations could get cleaned up in a separate pull, I suppose? Dunno, going to tentatively approve as is.

bdach (approved):

peppy (approved):

PR #2144, opened 2019-02-09
 Add bindable for 'safe area' padding

swoolcock:
Newer phones and tablets often have notches and rounded corners.  This PR exposes these as a bindable, since they dynamically change when the device is rotated.  Framework consumers can use this to layout UI elements so that they are always visible.

Includes tests, but the results are only evident when running on a device.  Can't really add any asserts either, so it's mostly for visual verification.

Will need to implement this in Android too.

- Closes #2075

---

Add support for 'safe area' padding on mobile devices.

PR #5012, opened 2022-02-04
Fetch safe area padding values on Android

bdach:
The safe area padding feature was never actually implemented for Android (see #2144), therefore making ppy/osu#16772 not have the desired effect there. Seems about as good a time to make work as any.

API reference:

* [`View.OnApplyWindowInsets(WindowInsets)`](https://developer.android.com/reference/android/view/View#onApplyWindowInsets(android.view.WindowInsets)) (API level 20, so below the minimum agreed supported level 21)
* [`WindowInsets.getDisplayCutout()`](https://developer.android.com/reference/android/view/WindowInsets#getDisplayCutout()) (API level 28, requires a runtime version check to not-crash on older devices)

Tested on:

* Samsung Galaxy J7 2016 / SM-J710F (Android 7 / API level 24),
* OnePlus 8T / KB2003 (Android 11 / API level 30)

### DIFF
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -3,11 +3,13 @@
 
 using System;
 using Android.Content;
+using Android.OS;
 using Android.Runtime;
 using Android.Text;
 using Android.Util;
 using Android.Views;
 using Android.Views.InputMethods;
+using JetBrains.Annotations;
 using osu.Framework.Android.Input;
 using osuTK.Graphics;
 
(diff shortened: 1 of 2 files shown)


Susko3:
This doesn't work as expected in multi-window mode. The padding gets needlessly applied.

Multi-window is something we must support ([we already do](https://github.com/ppy/osu-framework/blob/a273d71342c1db43bae6a03793705f07eb1d7be4/osu.Framework.Android/Properties/AssemblyInfo.cs#L17)), as Android will force it onto any app running on a tablet with Android 12: https://developer.android.com/guide/topics/large-screens/multi-window-support.

I'm actually working implementing this properly, but it's taking some time (as the values can't be just blindly applied).

<details>
  <summary>Screenshot (warning: tall)</summary>

Screenshot_2022-02-04-20-16-39-35_572064f74bd5f9fa804b05334aa4f912

</details>

bdach:
Would have been good to know you were working on this but oh well, I suppose I did the same thing by opening this without asking. I'll leave it to you then.

PR #5014, opened 2022-02-05
Implement `GetImage`/`SetImage`

uzervlad:
ppy/osu#16775

Implements `GetImage`/`SetImage` in `WindowsClipboard`, hopefully with support for other platforms coming soon

### DIFF
--- a/osu.Framework.Tests/Visual/Platform/TestSceneClipboard.cs
+++ b/osu.Framework.Tests/Visual/Platform/TestSceneClipboard.cs
@@ -10,7 +10,7 @@ using SixLabors.ImageSharp.PixelFormats;
 
 namespace osu.Framework.Tests.Visual.Platform
 {
-    [Ignore("This test cannot be run in headless mode (a window instance is required).")]
+    [Ignore("This test should not be run in headless mode, as it mutates the clipboard.")]
     public class TestSceneClipboard : FrameworkTestScene
     {
         private GameHost host;
--- a/osu.Framework.Tests/Visual/Platform/TestSceneClipboard.cs
+++ b/osu.Framework.Tests/Visual/Platform/TestSceneClipboard.cs
@@ -1,0 +1,74 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using NUnit.Framework;
+using osu.Framework.Allocation;
+using osu.Framework.Extensions;
+using osu.Framework.Platform;
+using SixLabors.ImageSharp;
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace osu.Framework.Tests.Visual.Platform
+{
+    [Ignore("This test cannot be run in headless mode (a window instance is required).")]
+    public class TestSceneClipboard : FrameworkTestScene
+    {
+        private GameHost host;
+
+        private Image<Rgba32> originalImage;
+        private Image<Rgba32> clipboardImage;
+
+        private Clipboard clipboard => host.GetClipboard();
+
+        [BackgroundDependencyLoader]
+        private void load(GameHost host)
+        {
+            this.host = host;
+        }
+
+        [Test]
+        public void TestImageCopy()
+        {
+            AddStep("screenshot screen", () =>
+            {
+                host.TakeScreenshotAsync().ContinueWith(t =>
+                {
+                    var image = t.GetResultSafely();
+                    originalImage = image.Clone();
+                    image.Dispose();
+                });
+            });
+
+            AddUntilStep("screenshot taken", () => originalImage != null);
+
+            AddStep("copy image to clipboard", () =>
+            {
+                clipboard.SetImage(originalImage);
+            });
+
+            AddStep("retrieve image from clipboard", () =>
+            {
+                clipboardImage = clipboard.GetImage<Rgba32>();
+            });
+
+            AddAssert("image retrieved", () => clipboardImage != null);
+
+            AddAssert("compare images", () =>
+            {
+                if (originalImage.Width != clipboardImage.Width && originalImage.Height != clipboardImage.Height)
+                    return false;
+
+                for (int x = 0; x < originalImage.Width; x++)
+                {
+                    for (int y = 0; y < originalImage.Height; y++)
+                    {
+                        if (originalImage[x, y] != clipboardImage[x, y])
+                            return false;
+                    }
+                }
+
+                return true;
+            });
+        }
+    }
+}
--- a/osu.Framework.iOS/IOSClipboard.cs
+++ b/osu.Framework.iOS/IOSClipboard.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using osu.Framework.Platform;
+using SixLabors.ImageSharp;
 using UIKit;
 
 namespace osu.Framework.iOS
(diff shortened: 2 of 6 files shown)


osu.Framework.iOS/IOSClipboard.cs
```diff
@@ -23,5 +23,10 @@ public override string GetText()
         }
 
         public override void SetText(string selectedText) => gameView.InvokeOnMainThread(() => UIPasteboard.General.String = selectedText);
+
+        public override void SetImage(Image image)
+        {
+            throw new NotImplementedException();
```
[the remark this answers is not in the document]
uzervlad: What should I do instead of an exception then?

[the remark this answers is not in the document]
bdach: if you're not aiming to implement the other platforms here, then the only reasonable remaining option is to do nothing (and probably warn in the xmldoc of the base `Clipboard` that platform support is limited).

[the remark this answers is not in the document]
peppy: If we're throwing it needs to be documented in the xmldoc (and should be `PlatformNotSupportedException` I guess).

Since we generally avoid that, a `bool` return type with success status may work better?

Susko3:
This needs some tests so it can be verified this works, and to show consumers how to use this properly (checking the return value, etc.).

bdach:
the test would probably have to be fully manual though (i.e. without asserts), and limited to windows only for now. there's not really a reasonable way to test this using assertions i don't think.

osu.Framework.iOS/IOSClipboard.cs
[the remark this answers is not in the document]
uzervlad: > Since we generally avoid that, a `bool` return type with success status may work better?

Seems like a good solution to me

osu.Framework/Platform/Windows/WindowsClipboard.cs
```diff
@@ -148,5 +154,72 @@ public override void SetText(string selectedText)
...
+                finally
+                {
+                    if (hGlobal != IntPtr.Zero)
+                        GlobalFree(hGlobal);
+                }
+            }
+            finally
+            {
+                CloseClipboard();
+            }
```
frenzibyte, line 220: This can probably be shared with the existing logic in `SetText` via a private `setClipboard(hGlobal, format)` or something.

osu.Framework/Platform/Clipboard.cs
```diff
@@ -1,12 +1,24 @@
...
 {
     public abstract class Clipboard
     {
         public abstract string GetText();
 
         public abstract void SetText(string selectedText);
+
+        /// <summary>
+        /// Copy the image to the clipboard.
+        /// </summary>
```
frenzibyte, line 16: nit: Might as well fill the entire `Clipboard` class with xmldoc.

frenzibyte (changes requested):

frenzibyte (commented):
Code-wise seems fine (minus or plus some details), but I can't test this yet so I'll defer that to someone else.

bdach:
well there are no tests whatsoever here still, or any other functionality that calls `SetImage()`, so it's unclear how this can actually be tested right now

uzervlad:
So I made a visual test, but I'm not sure how to do the first step right. The last one looks absurd, though it runs pretty much instantly for me.

<details>
<summary>Code</summary>

```cs
[Test]
public void TestScreenshotCopy()
{
    AddStep("screenshot screen", () =>
    {
        using (var image = host.TakeScreenshotAsync().GetAwaiter().GetResult())
        {
            originalImage = image.Clone();
        }
    });

    AddStep("copy image to clipboard", () =>
    {
        clipboard.SetImage(originalImage);
    });

    AddStep("retrieve image from clipboard", () =>
    {
        clipboardImage = clipboard.GetImage<Rgba32>();
    });

    AddAssert("compare images", () =>
    {
        if (originalImage.Width != clipboardImage.Width && originalImage.Height != clipboardImage.Height)
            return false;

        for (int x = 0; x < originalImage.Width; x++)
        {
            for (int y = 0; y < originalImage.Height; y++)
            {
                if (originalImage[x, y] != clipboardImage[x, y])
                    return false;
            }
        }

        return true;
    });
}
```
</details>

bdach:
You shouldn't be using `.GetAwaiter().GetResult()` (which is basically bypassing the `.Result` property which is banned API in this project). We have a `.GetResultSafely()` overload for that.

However, looking at how `.TakeScreenshotAsync()` is implemented, it will deadlock when you try to run the test in single threaded mode. That is because the screenshot implementation works via a task completion source, and enqueues the actual work of taking the screenshot onto the draw thread. In multi-threaded mode that's fine, but in single-threaded mode the update and draw "threads" run in an interleaved manner on one OS thread, and so the update thread will block waiting for the screenshot, but the screenshot - which is supposed to be taken by the draw thread - will never actually have a chance to be taken (because the update thread will be blocking the draw thread).

I think what you'll want to do there is something like the following:

```csharp
AddStep("screenshot screen", () =>
{
    host.TakeScreenshotAsync().ContinueWith(t =>
    {
        var image = t.GetResultSafely();
        originalImage = image.Clone();
        image.Dispose();
    });
});
AddUntilStep("screenshot taken", () => originalImage != null);
```

This works because continuations are automatically scheduled/enqueued for execution on the next update frame (via the `SynchronizationContext` installed by the framework).

osu.Framework/Platform/Windows/WindowsClipboard.cs
```diff
@@ -46,95 +51,100 @@ public class WindowsClipboard : Clipboard
...
-                        GlobalUnlock(handle);
-                }
-            }
-            finally
+            using (var stream = new MemoryStream())
             {
-                CloseClipboard();
+                var encoder = image.GetConfiguration().ImageFormatsManager.FindEncoder(BmpFormat.Instance);
+                image.Save(stream, encoder);
+                // The bitmap file header should not be included in clipboard
```
peppy, line 97: Need a reference stating why this is the case.

osu.Framework.Tests/Visual/Platform/TestSceneClipboard.cs
```diff
@@ -0,0 +1,74 @@
...
+using NUnit.Framework;
+using osu.Framework.Allocation;
+using osu.Framework.Extensions;
+using osu.Framework.Platform;
+using SixLabors.ImageSharp;
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace osu.Framework.Tests.Visual.Platform
+{
+    [Ignore("This test cannot be run in headless mode (a window instance is required).")]
```
peppy, line 13: You could test using a constructed image rather than a screenshot and remove this requirement. That said, i think it should still be ignored due to the fact it will mutate the clipboard even when running in the background.

I'd probably update the ignore message to state this as the reason, though.

peppy (changes requested):
Seems to work on windows as expected. A couple of issues I'd probably fix first.

osu.Framework.Tests/Visual/Platform/TestSceneClipboard.cs
bdach: The bigger reason to ignore this test is that as written it will fail on platforms that don't implement `{Get,Set}Image`.

osu.Framework/Platform/Windows/WindowsClipboard.cs
uzervlad: Not sure why exactly, I just found [this answer on StackOverflow](https://stackoverflow.com/a/7052068) which had those 14-bytes trimmed

peppy: This seems to suggest that any non-bitmap content in the clipboard will fail with `GetImage`, right? and the function doesn't need a generic type at this point?

bdach: I believe so, yeah. It can only be a bitmap.

peppy (approved):

PR #5016, opened 2022-02-05
Implement `SafeAreaPadding` on Android

Susko3:
Superseded https://github.com/ppy/osu-framework/pull/5012.

Tested on:
- Realme 6 / RMX2001 (Android 11 / API 30)
- LG G7 Fit / LG-Q850EMW (Android 9 / API 28)
- LG G Pad 8.3 / LG-V500 (Android 8.1 / API 27)

Testing methodology involves using multi-window and trying out different screen orientations.

Some of the extra code (compared to the original PR) seems to be working around a platform issue with my main device (Realme 6), as the original PR works just fine on the LG G7 Fit 🙃

At least we got non-complaint devices covered.

Anyways, I feel much safer calculating all the insets on the screen area, and then intersecting with the view at the very end.

### DIFF
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -11,6 +11,7 @@ using Android.Util;
 using Android.Views;
 using Android.Views.InputMethods;
 using osu.Framework.Android.Input;
+using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Platform;
@@ -24,6 +25,8 @@ namespace osu.Framework.Android
         public AndroidGameHost Host { get; private set; }
 
         public AndroidGameActivity Activity { get; }
+
+        public BindableSafeArea SafeAreaPadding { get; } = new BindableSafeArea();
 
         private readonly Game game;
 
@@ -120,12 +123,12 @@ namespace osu.Framework.Android
         [STAThread]
         public void RenderGame()
         {
+            LayoutChange += (_, __) => updateSafeArea();
+
             Host = new AndroidGameHost(this);
             Host.ExceptionThrown += handleException;
             Host.Run(game);
             HostStarted?.Invoke(Host);
-
-            LayoutChange += (_, __) => updateSafeArea();
 
             // if this is run immediately, we'll have an invalid layout (Width == Height == 0).
             Host.InputThread.Scheduler.Add(updateSafeArea);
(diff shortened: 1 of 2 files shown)


osu.Framework.Android/AndroidGameView.cs
```diff
@@ -3,25 +3,34 @@
...
 using osuTK.Graphics;
+using Debug = System.Diagnostics.Debug;
 
 namespace osu.Framework.Android
 {
     public class AndroidGameView : osuTK.Android.AndroidGameView
     {
         public AndroidGameHost Host { get; private set; }
 
+        public readonly AndroidGameActivity Activity;
```
bdach, line 26: kind of weird to see a `get; private set;` property next to a `public readonly` field - would prefer to keep it consistent one way or the other

bdach:
> Some of the extra code (compared to the original PR) seems to be working around a platform issue with my main device (Realme 6), as the original PR works just fine on the LG G7 Fit

do you mean the multi-window thing by this? because if yes, then i just checked and we either both have non-compliant devices or something in android itself changed something between API 28 and 30, because multi-window is also broken on my API 30 device in the same way you mentioned previously.

osu.Framework.Android/AndroidGameView.cs
```diff
@@ -3,25 +3,34 @@
...
+        public AndroidGameActivity Activity { get; }
+
         private readonly Game game;
 
-        public AndroidGameView(Context context, Game game)
-            : base(context)
+        public AndroidGameView(AndroidGameActivity activity, Game game)
+            : base(activity)
         {
+            Activity = activity;
```
Susko3, line 33: We currently have `(AndroidGameActivity)view.Context` sprinkled around in our code. Having the activity directly accessible like this should improve code quality.

This change was taken from a cleanup branch, useful here because of `Activity.IsInMultiWindowMode` later in the PR.

Susko3:
> > Some of the extra code (compared to the original PR) seems to be working around a platform issue with my main device (Realme 6), as the original PR works just fine on the LG G7 Fit
> 
> do you mean the multi-window thing by this? because if yes, then i just checked my solution again and we either both have non-compliant devices or something in android itself changed something between API 28 and 30, because multi-window is also broken on my API 30 device in the same way you mentioned previously.

Hmm, looks like something changed on Android's part. The view would draw under the status bar previously, but it doesn't in the new version, the status bar is just black. Google Maps draws under the status bar in multi-window, so there must be some flag (or similar) to set. Anyways, it's not really important.

I presume this PR works fine on your devices?

bdach:
> I presume this PR works fine on your devices?

have tested only just now. appears to work correctly.

bdach (approved):

osu.Framework.Android/AndroidGameView.cs
```diff
@@ -115,6 +124,11 @@ public void RenderGame()
             Host.ExceptionThrown += handleException;
             Host.Run(game);
             HostStarted?.Invoke(Host);
+
+            LayoutChange += (_, __) => updateSafeArea();
+
+            // if this is run immediately, we'll have an invalid layout (Width == Height == 0).
+            Host.InputThread.Scheduler.Add(updateSafeArea);
```
peppy, line 131: I'd feel a lot safer if these could be placed above the `Host.Run` call (to ensure beyond doubt that the event handling at https://github.com/ppy/osu-framework/pull/5016/files#diff-1f9486dea90c3d24267f78ddcaaf9155ede2cfc612414bf25cfe5877d75fa9d4R35 is not bound too late to handle an initial callback), is that possible?

I think the above is already impossible due to order of execution, but when reading just this code it feels a bit flaky.

peppy (changes requested):

peppy (approved):

PR #5019, opened 2022-02-07
Save the `input.json` config when input settings are changed

Susko3:
Closes https://github.com/ppy/osu-framework/issues/4950.

The code is a bit ugly, but that's to be expected when doing reflection with generics.

Quickly realized that constructing a `Action<ValueChangedEvent<T>>`  and calling `BindValueChanged` directly was gonna be very hard and even uglier.

Works as you'd expect. Tested with the mouse settings in `TestSceneInputManager`.


bdach (approved):
seems fine for the time being, at least until the dispose kinks are sorted out - i can see that having some knock on effects in this reflectiony logic

osu.Framework/Configuration/InputConfigManager.cs
```diff
@@ -83,5 +86,36 @@ protected override void PerformLoad()
...
+        /// and calls <see cref="ConfigManager.QueueBackgroundSave"/> when their values change.
+        /// </summary>
+        private void bindToHandlersBindables()
+        {
+            foreach (var handler in InputHandlers)
+            {
+                foreach (var property in handler.GetType().GetProperties())
+                {
+                    // get the underlying IBindable<T> for this property.
+                    var bindableType = property.PropertyType.GetInterfaces().FirstOrDefault(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IBindable<>));
```
frenzibyte, line 101: By accepting `IBindable<>` properties, this can potentially bind to non-configurable settings (e.g. `OpenTabletDriverHandler.Tablet`):

CleanShot 2022-02-07 at 23 13 13
https://github.com/ppy/osu-framework/blob/b9e666caba764ddacda2303b85181fffe0b2890a/osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs#L37-L39

All configurable settings should be of type `Bindable<>` anyways so it should be safe limiting to that instead.

frenzibyte (changes requested):
Not sure how to feel about using reflection, but since it is the simplest way through I suppose it's fine for the time being.

Missing test coverage, should be simple to add one (would imagine it to test modifying handler-specific properties, and also shared ones present at an `InputHandler` level such as `Enabled`).

osu.Framework/Configuration/InputConfigManager.cs
bdach: even if this was a real issue (which it arguably isn't), how is scoping down to the abstract class an actual fix to this, rather than adding some facility to ignore a bindable?

Susko3: I was going off what is saved in `input.json`.
This does look like it saves too much.

<details>

<summary>input.json</summary>

```json
{
    "InputHandlers": [
        {
            "$type": "osu.Framework.Input.Handlers.Keyboard.KeyboardHandler, osu.Framework",
            "Description": "Keyboard",
            "IsActive": true,
            "Enabled": true
        },
        {
            "$type": "osu.Framework.Input.Handlers.Tablet.OpenTabletDriverHandler, osu.Framework",
            "IsActive": true,
            "AreaOffset": {
                "x": 0.0,
                "y": 0.0
            },
            "AreaSize": {
                "x": 0.0,
                "y": 0.0
            },
            "Rotation": 0.0,
            "Tablet": null,
            "Description": "OpenTabletDriver",
            "Enabled": true
        },
        {
            "$type": "osu.Framework.Input.Handlers.Joystick.JoystickHandler, osu.Framework",
            "Description": "Joystick / Gamepad",
            "IsActive": true,
            "Enabled": true
        },
        {
            "$type": "osu.Framework.Input.Handlers.Midi.MidiHandler, osu.Framework",
            "Description": "MIDI",
            "IsActive": true,
            "Enabled": true
        },
        {
            "$type": "osu.Framework.Platform.Windows.WindowsMouseHandler, osu.Framework",
            "IsActive": true,
            "UseRelativeMode": false,
            "Sensitivity": 1.0,
            "Description": "Mouse",
            "Enabled": true
        }
    ]
}
```

</details>

bdach: the only things that are possibly extraneous there are `IsActive` on all handlers and the `Tablet` property i guess. but again, i argue that the way to go there is to stick a `[JsonProperty]` on all settings that are supposed to go to the file, enable explicit opt-in serialisation for `input.json`, and check for the presence of `[JsonProperty]` when doing the reflection logic which is already here. not sure whether that's in scope of this pull.

frenzibyte: I would at least scope it down rather than allowing unnecessary properties regardless of whether it's a real issue, as I see no harm in that. Adding a way to ignore certain bindables sounds like another solution but feels slightly more overengineered to me.

frenzibyte:
The test doesn't have to ensure the full saving flow works, but only ensure that at least `Save` is called when any setting is changed, as that's what's being changed/achieved in this PR (i.e. via a `TestInputConfigManager` class overriding `PerformSave`/`PerformLoad` to not interact with a real storage, and not require a game host).

Would like to note that not providing test coverage for changes like these could silently regress with any underlying change to the `Bindable` structure or the `ConfigManager` structure, and that's why I believe it is warranted here (especially for the fact that it's relying solely on reflection).

osu.Framework.Tests/Configuration/InputConfigManagerTest.cs
```diff
@@ -72,6 +73,38 @@ public void TestNewConfigPersists()
...
+        /// Verifies that <see cref="InputConfigManager"/> saves when any public-facing bindables of <see cref="InputHandler"/>s are changed.
+        /// </summary>
+        [Test]
+        public void TestChangingSettingsSaves()
+        {
+            var handler = new MouseHandler();
+            var config = new TestInputConfigManager(new[] { handler });
+
+            handler.Sensitivity.Value = 5;
+            Thread.Sleep(150); // wait for QueueBackgroundSave() debounce.
```
peppy, line 86: As shown by the test failures, we probably can't rely on timings like this for CI (all the `Thread.Sleep`s used in these tests are going to have similar intermittent problems). I'd probably just wait for it with a timeout (I think nunit's `Is.True.After` constraint may work well if you can't use `ManualResetEventSlim` style signalling).

frenzibyte (approved):

osu.Framework.Tests/Configuration/InputConfigManagerTest.cs
```diff
@@ -83,26 +83,25 @@ public void TestChangingSettingsSaves()
             var config = new TestInputConfigManager(new[] { handler });
 
             handler.Sensitivity.Value = 5;
-            Thread.Sleep(150); // wait for QueueBackgroundSave() debounce.
-            Assert.AreEqual(config.TimesSaved, 1);
+            Assert.IsTrue(config.SaveEvent.WaitOne(200)); // wait for QueueBackgroundSave() debounce.
```
peppy, line 86: Any reason for using a value this low? I think we've seen CI potentially stall for longer than this, so I'd suggest 1 or 10 seconds.

peppy (approved):

PR #5020, opened 2022-02-08
Fix crash with empty markdown fenced code blocks

smoogipoo:
Fixes https://github.com/ppy/osu/issues/16808

It still draws the background, which looks to be expected for this sort of input (tested via both dillinger.io and github).

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneMarkdownContainer.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneMarkdownContainer.cs
@@ -301,6 +301,12 @@ soft break with '\'";
             AddAssert("has correct autolink", () => markdownContainer.AutoLinks[0].Url == "https://discord.gg/ppy");
         }
 
+        [Test]
+        public void TestEmptyFencedBlock()
+        {
+            AddStep("set empty fenced block", () => markdownContainer.Text = @"```");
+        }
+
         private class TestMarkdownContainer : MarkdownContainer
         {
             public new string DocumentUrl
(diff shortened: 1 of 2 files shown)


frenzibyte (approved):

osu.Framework.Tests/Visual/UserInterface/TestSceneMarkdownContainer.cs
```diff
@@ -301,6 +301,12 @@ public void TestAutoLinkInline()
             AddAssert("has correct autolink", () => markdownContainer.AutoLinks[0].Url == "https://discord.gg/ppy");
         }
 
+        [Test]
+        public void TestEmptyFencedBlock()
+        {
+            AddStep("set empty fenced block", () => markdownContainer.Text = @"```");
```
peppy, line 307: Is this intentionally unbalanced? Seems to still repro with a balanced markdown string (which sits better with me for this test):

```suggestion
            AddStep("set empty fenced block", () => markdownContainer.Text = @"``````");
```

peppy (changes requested):

osu.Framework.Tests/Visual/UserInterface/TestSceneMarkdownContainer.cs
smoogipoo: Intentionally unbalanced just to show that even this unbalancedness draws a background like in other services.

peppy: Hmm, I'd probably rename the method to `TestUnbalancedFencedBlock` and maybe add a second one with your original name and my proposal.

peppy (approved):

PR #5021, opened 2022-02-10
Allow `CursorContainer` to update cursor position while hidden

frenzibyte:
Closes https://github.com/ppy/osu/issues/16792


peppy (approved):

peppy:
Have tested against osu! manually (and run remote tests). Looks to work okay, and read better than what we had.

PR #5022, opened 2022-02-10
Update to net6.0

peppy:
I've tested iOS but can't seem to get android to work on this branch *or* `master` (likely due to an incredibly broken monodevelop install). Hoping someone else can test that for me.

- [x] Test android builds

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ This framework is intended to take steps beyond what you would normally expect f
 ## Requirements
 
 - A desktop platform with the [.NET 6.0 SDK](https://dotnet.microsoft.com/download).
-- When running on linux, please have a system-wide ffmpeg installation available to.
+- When running on linux, please have a system-wide ffmpeg installation available to support video decoding.
 - When running on Windows 7 or 8.1, *[additional prerequisites](https://docs.microsoft.com/en-us/dotnet/core/install/windows?tabs=net60&pivots=os-windows#dependencies)** may be required to correctly run .NET 6 applications if your operating system is not up-to-date with the latest service packs.
 - When working with the codebase, we recommend using an IDE with intellisense and syntax highlighting, such as [Visual Studio 2019+](https://visualstudio.microsoft.com/vs/), [Jetbrains Rider](https://www.jetbrains.com/rider/) or [Visual Studio Code](https://code.visualstudio.com/).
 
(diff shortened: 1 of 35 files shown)


smoogipoo:
I don't think the `NetAnalyzers` change you made is required. It should be enough to change the following line:
https://github.com/ppy/osu-framework/blob/3fc1b93dda32eb9324f247106c48e6f3dfe737d4/Directory.Build.props#L24

peppy:
Oops, that was rider's suggested change.

smoogipoo (approved):

smoogipoo:
@bdach mostly requesting a review to see if you're fine with the changes around `stream.Read()` calls.

peppy:
If we want to do this correctly, a helper method sounds best. I had a quick look for an already existing one when making this PR but nothing came up.

bdach:
Yeah I don't necessarily know what to do with the above information if I'm honest. The scary part of that thread of reasoning is that if those reads ever failed up until now, they would have done so silently and any failures could be essentially a few steps removed from the original read (something like deserialisation failing, or random failures to read resources, etc.)

Joehuu:
The readme needs to be updated with "or higher" also removed.

peppy:
I've got a proof of concept stream extension function, but would rather PR it separate to this to keep things isolated.

README.md
```diff
@@ -27,9 +27,9 @@ This framework is intended to take steps beyond what you would normally expect f
 
 ## Requirements
 
-- A desktop platform with the [.NET 5.0 SDK](https://dotnet.microsoft.com/download) or higher installed.
-- When running on linux, please have a system-wide ffmpeg installation available to support video decoding.
-- When running on Windows 7 or 8.1, *[additional prerequisites](https://docs.microsoft.com/en-us/dotnet/core/install/windows?tabs=net50&pivots=os-windows#dependencies)** may be required to correctly run .NET 5 applications if your operating system is not up-to-date with the latest service packs.
+- A desktop platform with the [.NET 6.0 SDK](https://dotnet.microsoft.com/download).
+- When running on linux, please have a system-wide ffmpeg installation available to.
```
frenzibyte, line 31: unintentional?
```suggestion
- When running on linux, please have a system-wide ffmpeg installation available to support video decoding.
```

smoogipoo (approved):

PR #5024, opened 2022-02-11
Create helper methods to ensure stream reads are read correctly to the expected length

peppy:
Based loosely around the solution proposed [here](https://stackoverflow.com/a/221941).

The only thing I'm not 100% on is whether we should use the `EndOfStreamException` here as there could be a super edge case where a stream reads past the end of the expected length? Not sure that will ever happen though.

- [x] Depends on #5022

### DIFF
--- a/osu.Framework/Extensions/StreamExtensions.cs
+++ b/osu.Framework/Extensions/StreamExtensions.cs
@@ -56,18 +56,19 @@ namespace osu.Framework.Extensions
         {
             byte[] buffer = new byte[16 * 1024];
 
+            int remainingRead = length;
+
             using (var ms = new MemoryStream(length))
             {
                 int read;
-                int totalRead = 0;
 
-                while ((read = stream.Read(buffer, 0, buffer.Length)) > 0)
+                while ((read = stream.Read(buffer, 0, Math.Min(remainingRead, buffer.Length))) > 0)
                 {
                     ms.Write(buffer, 0, read);
-                    totalRead += read;
+                    remainingRead -= read;
                 }
 
-                if (totalRead != length)
+                if (remainingRead != 0)
                     throw new EndOfStreamException();
 
                 return ms.GetBuffer();
@@ -85,18 +86,19 @@ namespace osu.Framework.Extensions
         {
             byte[] buffer = new byte[16 * 1024];
 
+            int remainingRead = length;
+
             using (var ms = new MemoryStream(length))
             {
                 int read;
-                int totalRead = 0;
 
-                while ((read = await stream.ReadAsync(buffer.AsMemory(), cancellationToken).ConfigureAwait(false)) > 0)
+                while ((read = await stream.ReadAsync(buffer, 0, Math.Min(buffer.Length, remainingRead), cancellationToken).ConfigureAwait(false)) > 0)
                 {
                     await ms.WriteAsync(buffer, 0, read, cancellationToken).ConfigureAwait(false);
-                    totalRead += read;
+                    remainingRead -= read;
                 }
 
-                if (totalRead != length)
+                if (remainingRead != 0)
                     throw new EndOfStreamException();
 
                 return ms.GetBuffer();
--- a/osu.Framework/Extensions/StreamExtensions.cs
+++ b/osu.Framework/Extensions/StreamExtensions.cs
@@ -146,15 +146,9 @@ namespace osu.Framework.Extensions
         /// <returns>The full byte content.</returns>
         public static async Task<byte[]> ReadAllRemainingBytesToArrayAsync(this Stream stream, CancellationToken cancellationToken = default)
         {
-            byte[] buffer = new byte[buffer_size];
-
             using (var ms = new MemoryStream())
             {
-                int read;
-
-                while ((read = await stream.ReadAsync(buffer, cancellationToken).ConfigureAwait(false)) > 0)
-                    await ms.WriteAsync(buffer, 0, read, cancellationToken).ConfigureAwait(false);
-
+                await stream.CopyToAsync(ms, cancellationToken).ConfigureAwait(false);
                 return ms.ToArray();
             }
         }
--- a/osu.Framework/Extensions/StreamExtensions.cs
+++ b/osu.Framework/Extensions/StreamExtensions.cs
@@ -20,6 +20,7 @@ namespace osu.Framework.Extensions
         {
             if (stream.CanSeek)
             {
+                stream.Seek(0, SeekOrigin.Begin);
                 Debug.Assert(stream.Length < int.MaxValue);
                 return stream.ReadBytesToArray((int)stream.Length);
             }
@@ -37,6 +38,7 @@ namespace osu.Framework.Extensions
         {
             if (stream.CanSeek)
             {
+                stream.Seek(0, SeekOrigin.Begin);
                 Debug.Assert(stream.Length < int.MaxValue);
                 return stream.ReadBytesToArrayAsync((int)stream.Length, cancellationToken);
             }
--- a/osu.Framework/Extensions/StreamExtensions.cs
+++ b/osu.Framework/Extensions/StreamExtensions.cs
@@ -18,9 +18,13 @@ namespace osu.Framework.Extensions
         /// <returns>The full byte content.</returns>
         public static byte[] ReadAllBytesToArray(this Stream stream)
         {
-            Debug.Assert(stream.Length < int.MaxValue);
+            if (stream.CanSeek)
+            {
+                Debug.Assert(stream.Length < int.MaxValue);
+                return stream.ReadBytesToArray((int)stream.Length);
+            }
 
-            return stream.ReadBytesToArray((int)stream.Length);
+            return stream.ReadArbitraryBytesToArray();
         }
 
         /// <summary>
@@ -31,9 +35,13 @@ namespace osu.Framework.Extensions
         /// <returns>The full byte content.</returns>
         public static Task<byte[]> ReadAllBytesToArrayAsync(this Stream stream, CancellationToken cancellationToken)
         {
-            Debug.Assert(stream.Length < int.MaxValue);
+            if (stream.CanSeek)
+            {
+                Debug.Assert(stream.Length < int.MaxValue);
+                return stream.ReadBytesToArrayAsync((int)stream.Length, cancellationToken);
+            }
 
-            return stream.ReadBytesToArrayAsync((int)stream.Length, cancellationToken);
+            return stream.ReadArbitraryBytesToArrayAsync(cancellationToken);
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


peppy:
Blocking temporarily to add test coverage.

osu.Framework/Platform/TcpIpcProvider.cs
```diff
@@ -182,21 +182,21 @@ private async Task send(Stream stream, IpcMessage message)
...
-            int readBytes = await stream.ReadAsync(header.AsMemory(), cancellationToken).ConfigureAwait(false);
+            const int header_length = sizeof(int);
 
-            if (readBytes < header.Length)
-                throw new EndOfStreamException();
+            byte[] header = new byte[header_length];
 
-            int len = BitConverter.ToInt32(header, 0);
-            if (len == 0)
+            int read = await stream.ReadAsync(header.AsMemory(), cancellationToken).ConfigureAwait(false);
```
bdach, line 189: Shouldn't this be a

```csharp
byte[] header = await stream.ReadBytesToArrayAsync(header_length, cancellationToken).ConfigureAwait(false);
```

call instead?

osu.Framework/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// Read the full content of a stream.
+        /// </summary>
+        /// <param name="stream">The stream to read.</param>
+        /// <returns>The full byte content.</returns>
+        public static byte[] ReadAllBytesToArray(this Stream stream)
+        {
+            if (stream.CanSeek)
+            {
+                stream.Seek(0, SeekOrigin.Begin);
+                Debug.Assert(stream.Length < int.MaxValue);
```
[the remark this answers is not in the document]
peppy: Due to the int cast in the next line. Can remove if you don't think it's required.

osu.Framework/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,149 @@
...
+                return stream.ReadBytesToArray((int)stream.Length);
+            }
+
+            return stream.ReadAllRemainingBytesToArray();
```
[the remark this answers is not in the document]
peppy: Can do, was a toss up for me.

osu.Framework/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        /// <summary>
+        /// Read specified length from current position in stream.
+        /// </summary>
+        /// <param name="stream">The stream to read.</param>
+        /// <param name="length">The length to read.</param>
+        /// <returns>The full byte content.</returns>
+        public static byte[] ReadBytesToArray(this Stream stream, int length)
+        {
+            byte[] buffer = new byte[16 * 1024];
```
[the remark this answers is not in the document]
peppy: Matches .net internals

osu.Framework/Extensions/StreamExtensions.cs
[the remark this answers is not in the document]
bdach: Oh, I see. I think I'd rather change `ReadBytesToArray(Async)` to accept length as `long` rather than `int`. Seems like it will generate less friction against the underlying stream API anyhow.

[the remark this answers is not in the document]
peppy: The base API only allows `int` as far as I could see :(.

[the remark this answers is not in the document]
bdach: Oh because `Read()` takes `int` and `Position`/`Length` takes `long`... Kinda icky.

You could arguably handle that mismatch internally - since `ReadBytesToArray()` will only ever read at most 16 * 1024 bytes at a time, you could change that method to accept `long` and safely cast down the result of the `Math.Min()` operation to `int`. But given that `int.MaxValue` bytes is about 2GB of data then maybe no real need to worry about it because hopefully nobody is reading that much...?

bdach:
For kicks I've also written a quick and terrible `MemoryStream`-like implementation that only ever reads one byte at a time, to see if it would pass the tests added here - and the tests do indeed pass. Not sure if it's worth including but I'll leave here for posterity at least?

<details>
<summary>the slowest stream implementation to ever exist</summary>

```csharp
        private class SingleByteReadingMemoryStream : Stream
        {
            private readonly byte[] buffer;

            public SingleByteReadingMemoryStream(byte[] buffer)
            {
                this.buffer = buffer;
            }

            #region Reading

            public override bool CanRead => true;

            public override int Read(byte[] destination, int offset, int count)
            {
                if (count < 1 || position >= Length)
                    return 0;

                destination[offset] = buffer[position];
                position += 1;
                return 1;
            }

            #endregion

            #region Writing

            public override bool CanWrite => false;

            public override void Write(byte[] destination, int offset, int count) => throw new NotSupportedException();

            public override void Flush() => throw new NotSupportedException();

            public override void SetLength(long value) => throw new NotSupportedException();

            #endregion

            #region Seek support

            public override long Length => buffer.Length;

            private long position;

            public override long Position
            {
                get => position;
                set => Seek(value, SeekOrigin.Begin);
            }

            public override bool CanSeek => true;

            public override long Seek(long offset, SeekOrigin origin)
            {
                switch (origin)
                {
                    case SeekOrigin.Begin:
                        position = offset;
                        break;

                    case SeekOrigin.Current:
                        position += offset;
                        break;

                    case SeekOrigin.End:
                        position = Length + offset;
                        break;
                }

                return position;
            }

            #endregion
        }
```

</details>

The async test cases take 11 seconds (!!!) to complete with that implementation. I don't even want to think about the TPL thrashing that is causing.

osu.Framework/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,149 @@
...
+                while ((read = stream.Read(buffer, 0, Math.Min(remainingRead, buffer.Length))) > 0)
+                {
+                    ms.Write(buffer, 0, read);
+                    remainingRead -= read;
+                }
+
+                if (remainingRead != 0)
+                    throw new EndOfStreamException();
+
+                return ms.GetBuffer();
```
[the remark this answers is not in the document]
peppy: that's why there's a length check, yep.

osu.Framework/Extensions/StreamExtensions.cs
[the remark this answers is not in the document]
peppy: yeah, i doubt it'll be an issue. we can handle if someone hits the exception.

osu.Framework/Extensions/StreamExtensions.cs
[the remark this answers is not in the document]
peppy: to clarify, this usage in fact doesn't require the specific length check as the initial capacity was specified (which sets the internal buffer size).

bdach (dismissed):

osu.Framework/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,158 @@
...
+            byte[] buffer = new byte[buffer_size];
+
+            using (var ms = new MemoryStream())
+            {
+                int read;
+
+                while ((read = stream.Read(buffer, 0, buffer.Length)) > 0)
+                    ms.Write(buffer, 0, read);
+
+                return ms.ToArray();
```
smoogipoo, line 133: Any reason for this to use `.ToArray()` when the others use `.GetBuffer()`?

osu.Framework/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,158 @@
...
+            byte[] buffer = new byte[buffer_size];
+
+            using (var ms = new MemoryStream())
+            {
+                int read;
+
+                while ((read = await stream.ReadAsync(buffer.AsMemory(), cancellationToken).ConfigureAwait(false)) > 0)
+                    await ms.WriteAsync(buffer, 0, read, cancellationToken).ConfigureAwait(false);
+
+                return ms.ToArray();
```
smoogipoo, line 154: Here too.

osu.Framework/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// <returns>The full byte content.</returns>
+        public static async Task<byte[]> ReadAllRemainingBytesToArrayAsync(this Stream stream, CancellationToken cancellationToken = default)
+        {
+            byte[] buffer = new byte[buffer_size];
+
+            using (var ms = new MemoryStream())
+            {
+                int read;
+
+                while ((read = await stream.ReadAsync(buffer.AsMemory(), cancellationToken).ConfigureAwait(false)) > 0)
```
smoogipoo, line 151: Any reason for using `.AsMemory()`?

osu.Framework/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        /// <summary>
+        /// Read the full content of a seekable stream.
+        /// </summary>
+        /// <remarks>
+        /// For a non-seekable stream, use <see cref="ReadAllRemainingBytesToArray"/> instead.
+        /// </remarks>
+        /// <param name="stream">The stream to read.</param>
+        /// <param name="cancellationToken">A cancellation token.</param>
+        /// <returns>The full byte content.</returns>
```
smoogipoo, line 42: I'd prefer that these methods had `<exception>` docs for their special cases, which show up in Rider:
image
As opposed to a non-explicit tidbit in the remarks.

I don't mean just this method too, but `ReadBytesToArray()` as well because, as far as I know, `Stream.Read(length)` defines a _maximum_ read length, whereas the implementations below throw `EndOfStreamException`s in the same scenario.

smoogipoo (changes requested):

osu.Framework/Extensions/StreamExtensions.cs
peppy: Already touched on [in this discussion](https://github.com/ppy/osu-framework/pull/5024#discussion_r805152756).

The internal buffer is not equal to the actual read length as it uses the standard .NET buffer expansion (doubles in size each time it needs to). Requesting `GetBuffer` would return an array with zero bytes at the end in most cases, which we don't want here.

osu.Framework/Extensions/StreamExtensions.cs
peppy: Already touched on [in this discussion](https://github.com/ppy/osu-framework/pull/5024#discussion_r805152756).

The internal buffer is not equal to the actual read length as it uses the standard .NET buffer expansion (doubles in size each time it needs to). Requesting `GetBuffer` would return an array with zero bytes at the end in most cases, which we don't want here.

osu.Framework/Extensions/StreamExtensions.cs
peppy: No need, was a remnant of a previous version.

osu.Framework/Extensions/StreamExtensions.cs
peppy: Have added.

osu.Framework/Extensions/StreamExtensions.cs
peppy: Similar case to the one explained above. The only special case where I'm using `GetBuffer` is where the `new MemoryStream(length)` ctor is used, along with a length check (and throw) before returning the buffer.

osu.Framework/Extensions/StreamExtensions.cs
```diff
@@ -0,0 +1,162 @@
...
+        public static async Task<byte[]> ReadAllRemainingBytesToArrayAsync(this Stream stream, CancellationToken cancellationToken = default)
+        {
+            byte[] buffer = new byte[buffer_size];
+
+            using (var ms = new MemoryStream())
+            {
+                int read;
+
+                while ((read = await stream.ReadAsync(buffer, cancellationToken).ConfigureAwait(false)) > 0)
+                    await ms.WriteAsync(buffer, 0, read, cancellationToken).ConfigureAwait(false);
```
smoogipoo, line 156: Is there a reason for not using `stream.CopyToAsync()`? Seems like it does everything here plus uses the global array pool.

Would prefer to not reinvent the wheel.

smoogipoo (changes requested):

smoogipoo:
I have my gripes with this but let's see how it goes with the current implementation... What I was going to post is as follows, but I don't want to start a discussion about it:

I really don't know about this... The SO thread is based on the presumption that you can't trust `stream.Length`, but this code is explicitly saying you _can_ trust it (at least in the case of `ReadAllBytesToArray()`, which event asserts as much).  
Whether you can trust it in the case of e.g. network streams, is an answer I can't provide, because I simply don't know every edge case that exists. While it's true that [`Stream.Length`](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.length?view=net-6.0) throws an exception if not seekable which is covered by the guard in `ReadAllBytesToArray()`, could data be streamed in and `.Length` not represent the true length of the stream? Sounds possible to me, or at least there's preventing this.

Personal preference would be for the EOS exception to not exist and the methods to read as much as they can until EOS.

peppy:
> Personal preference would be for the EOS exception to not exist and the methods to read as much as they can until EOS.

Main reason for not doing this is it results in a second copy of the array being required, unless you want to branch for that exception case and add more complexity (something I was trying to avoid here because I agree with not wanting to reinvent the wheel, the point I wish these methods existed somewhere else and we didn't need to write them).

smoogipoo:
Dunno. Take that as you may. I don't think copying is a huge deal - at the point you're reading so much data that that's a point of concern, you probably shouldn't be doing this in the first place. But you could also return sliced `Memory` instances instead if that's that big of a concern.

smoogipoo (approved):

peppy:
This is being used for loading of all resources (aka `DllResourceStore`), which can be hundreds of megabytes for a game. So I thought it was a point of concern already. I was consciously benchmarking memory usage when writing these methods.

PR #5025, opened 2022-02-11
Bring Android build to zero warnings and cleanup configuration

Susko3:
Sample of the warnings we currently get:

### [XA0119](https://github.com/xamarin/xamarin-android/blob/main/Documentation/guides/messages/xa0119.md)
- C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets(434,3): Warning XA0119 : Using fast deployment and the linker at the same time is not recommended. Use fast deployment for Debug configurations and the linker for Release configurations.

- The docs also suggest to `<UseInterpreter>true</UseInterpreter>` for hot reload support, but this makes the game really laggy.

### [XA0125](https://github.com/xamarin/xamarin-android/blob/main/Documentation/guides/messages/xa0125.md)
- C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Xamarin\Android\Xamarin.Android.Legacy.targets(400,5): Warning XA0125 : 'osu.Framework.Tests.Android.pdb' is using a deprecated debug information level.
Set the debugging information to Portable in the Visual Studio project property pages or edit the project file in a text editor and set the 'DebugType' MSBuild property to 'portable' to use the newer, cross-platform debug information level.

- changed `DebugType` use the same as [`osu.Android.props`](https://github.com/ppy/osu/blob/b06caf2bf7b3a8705d26101f5e1fba464542a701/osu.Android.props#L25-L34)

### [XA4301](https://docs.microsoft.com/en-us/xamarin/android/errors-and-warnings/xa4301)
- C:\...\osu-framework\osu.Framework.Tests.Android\obj\Debug\100\lp\1\nl\arm64-v8a\libavcodec.so:
Warning XA4301 : APK already contains the item lib/arm64-v8a/libavcodec.so; ignoring.

---

Also used this opportunity to cleanup the Android `.props` and `.csproj` files. Just some cleanup work, no changes to build.

Tested to build and install fine on Rider 2021.3.3 and VS 2022.

### DIFF
--- a/SampleGame.Android/SampleGame.Android.csproj
+++ b/SampleGame.Android/SampleGame.Android.csproj
@@ -10,7 +10,6 @@
     <AndroidApplication>True</AndroidApplication>
     <AndroidManifest>Properties\AndroidManifest.xml</AndroidManifest>
     <AndroidSupportedAbis>armeabi-v7a;x86;arm64-v8a</AndroidSupportedAbis>
-    <TargetFrameworkVersion>v10.0</TargetFrameworkVersion>
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="SampleGameActivity.cs" />
--- a/SampleGame.Android/SampleGame.Android.csproj
+++ b/SampleGame.Android/SampleGame.Android.csproj
@@ -16,18 +16,9 @@
     <None Include="Properties\AndroidManifest.xml" />
   </ItemGroup>
   <ItemGroup>
-    <ProjectReference Include="..\osu.Framework.Android\osu.Framework.Android.csproj">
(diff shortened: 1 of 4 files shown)


Susko3:
The VS 2019 warning can be fixed by using `windows-2022` instead of `windows-latest`:
https://github.com/Susko3/osu-framework/actions/runs/1832090731
https://github.com/actions/virtual-environments/issues/4856

No idea about the `'Commons.Music.Midi.pdb' is using a deprecated debug information level.`
Don't get that when building locally, so it doesn't bother me that much.

peppy (approved):
let's see how this goes

PR #5026, opened 2022-02-12
Cleanup Android (and use fullscreen by default)

Susko3:
5161624f036e68e82672a6232b2579964dce6bbb **Use correct lifecycle methods for suspending and resuming host**
- the host would get suspended (even it was visible) when using split-screen on older android version (in Android 9 and lower) https://source.android.com/devices/tech/display/multi_display/multi-resume
- more info: https://developer.android.com/guide/components/activities/activity-lifecycle#alc - the image explains it well

dac31eb47340ca40ee52ad6a5565fa09d6d60783 **Always hide the status bar**
- we don't really support showing the status bar right now, so it's best to always hide it
- this matches iOS system bars behaviour
- can be revisited in the future if there is demand for showing it

### DIFF
--- a/osu.Framework.Android/AndroidGameActivity.cs
+++ b/osu.Framework.Android/AndroidGameActivity.cs
@@ -63,7 +63,7 @@ namespace osu.Framework.Android
 
             SetContentView(gameView = new AndroidGameView(this, CreateGame()));
 
-            UIVisibilityFlags = SystemUiFlags.LayoutFlags | SystemUiFlags.ImmersiveSticky | SystemUiFlags.HideNavigation;
+            UIVisibilityFlags = SystemUiFlags.LayoutFlags | SystemUiFlags.ImmersiveSticky | SystemUiFlags.HideNavigation | SystemUiFlags.Fullscreen;
 
             // Firing up the on-screen keyboard (eg: interacting with textboxes) may cause the UI visibility flags to be altered thus showing the navigation bar and potentially the status bar
             // This sets back the UI flags to hidden once the interaction with the on-screen keyboard has finished.
--- a/osu.Framework.Android/AndroidGameActivity.cs
+++ b/osu.Framework.Android/AndroidGameActivity.cs
@@ -93,16 +93,16 @@ namespace osu.Framework.Android
             };
         }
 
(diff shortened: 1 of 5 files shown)


osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -34,9 +35,14 @@ public abstract class AndroidGameActivity : Activity
         /// </summary>
         public SystemUiFlags UIVisibilityFlags
         {
-            get => (SystemUiFlags)Window.DecorView.SystemUiVisibility;
+            get
+            {
+                Debug.Assert(Window != null);
```
bdach, line 40: what are these assertions about?

bdach:
in general having very mixed feelings about bunching up cleanups with two actual behaviour changes into one pull. would have personally preferred taking this one step at a time.

osu.Framework.Android/AndroidGameActivity.cs
Susko3: image

Fixes these, since `Window` can be null, but never is for us.

https://developer.android.com/reference/android/app/Activity#getWindow()
> Window The current window, or null if the activity is not visual.

Could do something like this, but feels very hacky.

```cs
/// <summary>
/// The current window.
/// </summary>
/// <remarks>
/// Not <c>null</c> as the activity is always visual.
/// </remarks>
public override Window Window => base.Window!;

```

bdach: so this is just a warning cleanup change, correct? this is why i don't like mixed diffs like this.

for what it's work framework also has [an extension that may work here](https://github.com/ppy/osu-framework/blob/99786238a02e0d6b69da86dd52e5506ee8ec0566/osu.Framework/Extensions/ObjectExtensions/ObjectExtensions.cs#L24) but i'm not hugely fussed.

Susko3: > so this is just a warning cleanup change, correct?

Correct, the null asserts in 81bdb06b9668e0ee37c60be3b452f30659dde0e9 are only for that.

peppy (approved):
Looks sane enough..

PR #5028, opened 2022-02-12
Use `View` properties instead of osuTK ones for calculating `SafeArea`

Susko3:
Width and Height default to the ones in `GameViewBase`, no wonder they were invalid.

Since the native View properties are valid on creation, we can update earlier.

### DIFF
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -218,7 +218,7 @@ namespace osu.Framework.Android
 
             int[] location = new int[2];
             GetLocationOnScreen(location);
-            var viewArea = new RectangleI(location[0], location[1], (this as View).Width, (this as View).Height);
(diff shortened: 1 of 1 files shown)


osu.Framework.Android/AndroidGameView.cs
```diff
@@ -220,7 +218,7 @@ private void updateSafeArea()
 
             int[] location = new int[2];
             GetLocationOnScreen(location);
-            var viewArea = new RectangleI(location[0], location[1], Width, Height);
+            var viewArea = new RectangleI(location[0], location[1], (this as View).Width, (this as View).Height);
```
bdach, line 221: Don't know if it's just me but it seems really weird to be using soft casts that can't fail. Would prefer direct casts in this instance.

```suggestion
            var viewArea = new RectangleI(location[0], location[1], ((View)this).Width, ((View)this).Height);
```

smoogipoo (approved):

peppy (approved):

PR #5032, opened 2022-02-13
Add support for gradiented border colour

bdach:
PRing this as a proposal.

Some of the [newer](https://www.figma.com/file/NkdREXr6wFkA99oLw8PnBv/Client%2FMod-Selection?node-id=0%3A17) [figma designs](https://www.figma.com/file/NkdREXr6wFkA99oLw8PnBv/Client%2FMod-Selection?node-id=0%3A17) feature subtle but very present gradient effects on some border elements:

<details>
<summary>see examples</summary>

| example from mod select design | example from chat design |
| :-: | :-: |
| 2022-02-13-170241_162x306_scrot | 2022-02-13-170301_189x184_scrot |

</details>

While this does not necessarily need framework support, as a plausible workaround would be to nest two `CompositeDrawable`s - one larger with the gradient effect, and one smaller inside it with the solid colour, it feels rather finicky to do. So this PR is a proposal of how this could be supported natively.

Test coverage included:

https://user-images.githubusercontent.com/20418176/153761948-570d7a3e-e4c2-451c-8028-9659b45f4fb9.mp4

- [ ] Needs testing on iOS to make sure there are no problems with the shader (my initial version had similar issues as #4702, which I worked around by using masking attributes which are `highp`, so should hopefully work - it does on Android, at least).

---

# Breaking changes

## `CompositeDrawable.BorderColour` has changed type from `SRGBColour` to `ColourInfo`

To facilitate gradiented border support, the type of `CompositeDrawable.BorderColour` has changed from `SRGBColour` to `ColourInfo`. While some implicit conversions from `SRGBColour` to `ColourInfo` exist, some properties of `SRGBColour` are not available on `ColourInfo`, as the latter does not always represent a single colour, and may require appropriate adjustments.

### DIFF
--- a/osu.Framework/Graphics/Visualisation/DrawVisualiser.cs
+++ b/osu.Framework/Graphics/Visualisation/DrawVisualiser.cs
@@ -253,7 +253,7 @@ namespace osu.Framework.Graphics.Visualisation
                     if (!composite.Masking)
                         return;
 
-                    if (composite.BorderThickness > 0 && composite.BorderColour.Linear.A > 0
+                    if (composite.BorderThickness > 0 && composite.BorderColour.MinAlpha > 0
                         || composite.EdgeEffect.Type != EdgeEffectType.None && composite.EdgeEffect.Radius > 0 && composite.EdgeEffect.Colour.Linear.A > 0)
                     {
                         drawableTarget = composite;
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneBorderColour.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneBorderColour.cs
@@ -1,0 +1,47 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using NUnit.Framework;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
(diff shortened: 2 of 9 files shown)


osu.Framework/Graphics/Visualisation/DrawVisualiser.cs
```diff
@@ -253,7 +253,7 @@ void findTarget(Drawable drawable)
                     if (!composite.Masking)
                         return;
 
-                    if (composite.BorderThickness > 0 && composite.BorderColour.Linear.A > 0
+                    if (composite.BorderThickness > 0 && composite.BorderColour.MinAlpha > 0
```
smoogipoo, line 256: Perhaps `MaxAlpha` is more correct?

smoogipoo (approved):
Other than a small nit, looks good to me. Pending someone else testing this on iOS to resolve the task list @peppy / @frenzibyte.

frenzibyte:
Can confirm gradiented border colours display correctly on iOS using the attached test cases.

peppy:
Looks good to me, apart from the min->max change.

osu.Framework/Graphics/Visualisation/DrawVisualiser.cs
bdach: You're right, it would be. Good spot.

peppy (approved):

PR #5036, opened 2022-02-16
Add NativeLibs for osx-arm64

nekodex:
FFmpeg built from [FFmpeg 4.3.3](https://ffmpeg.org/releases/ffmpeg-4.3.3.tar.gz) using flags:
```
--disable-programs --disable-doc --disable-static --enable-shared --prefix=build \
--libdir=build/lib --enable-gpl
```
`--enable-gpl` seems to be required for `libpostproc.55.dylib` (which is present for other platforms), but I'm not sure if we're actually using it at all? Library load paths were then fixed using `install_name_tool`.

BASS / BASS_fx / BASSmix are pre-built libraries taken from https://www.un4seen.com

Visual/unit tests and whatnot seem to all work fine on my M1 running Monterey 12.1 (using .NET 6.0 for Arm64)


bdach:
I am not a lawyer but I am pretty sure we cannot compile FFmpeg with `--enable-gpl`. It would force framework and all consumers by extension to adhere to GPLv3, which would mean relicensing every project that uses o!f to GPLv3.

Better to stick to the LGPL parts if possible.

nekodex:
Yeah I agree, but that flag is required to match what's already being used in the project.

Ideally we should ensure we're not using those features and then remove those libraries/parts from all platforms' NativeLibs.

nekodex:
superseded by #5043

Issue #5039, opened 2022-02-17
Dynamic recompilation fails when a class uses a system diagnostic attribute


dekrain:
OS: Kubuntu 21.10
.NET versions: Confirmed on both 5.0.404 and 6.0.200
osu-framework: 2022.214.0-ish

When I use an attribute from the `System.Diagnostics.CodeAnalysis` namespace, e.g. `NotNullIfNotNullAttribute`, an error similar to this occurs during dynamic recompilation (reformatted for sanity):
```
System.AggregateException: One or more errors occurred.
(/osu.Framework/Graphics/Visualisation/VisualisedDrawable.cs(1161,10): error CS0433: The type 'NotNullAttribute' exists in both
'Microsoft.Build.Framework, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' and
'Newtonsoft.Json, Version=13.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed')
(/osu.Framework/Graphics/Visualisation/VisualisedDrawable.cs(269,18): error CS0433: The type 'NotNullIfNotNullAttribute' exists in both
'Microsoft.Build.Framework, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' and
'nunit.framework, Version=3.13.2.0, Culture=neutral, PublicKeyToken=2638cd05610744eb')
```

dekrain:
I tried using a similar solution by banning assemblies other than `System.Private.CoreLib` that define `System.*` types, but then NUNit framework wasn't loaded and unit test failed recompilation

Issue #4699, opened 2021-08-20
Error on dynamic recompilation


metametamoon:
dotnet version: 6.0.100-preview.6.21355.2
OS: Windows 10
osu-framework: 2021.728.0

I created a default osu-framework project using a template with a spinning box and tried to stop the rotation (by commenting the corresponding line) while running visual tests in debug mode (no debugger attached). However, the dynamic recompilation failed (the console log is attached to this issue). Installing NuGet.ProjectModel and System.ComponentModel from the NuGet site doesn't seem to help.
[log.txt](https://github.com/ppy/osu-framework/files/7020718/log.txt)

jai-x:
Possible duplicate of https://github.com/ppy/osu-framework/issues/4512

If you're using an installed version of `dotnet` that's managed by a version manager like `asdf`, it does not install all the required dependencies and you should use the system provided `dotnet` or download and install manually from the dotnet website.

Kremator1007:
I do not use any version manager though. I reinstalled dotnet from the microsoft site just in case, and the problem is still there (same exception is thrown).

jai-x:
Would you be able to post your system specs and the version of dotnet you are running?

Kremator1007:
I edited the comment to include the mentioned details.

jai-x:
I would try downgrading to use dotnet 5.x instead of dotnet 6-preview. I don't believe the framework has been tested to work on dotnet 6 yet.

Kremator1007:
I downgraded; unfortunately, the error perseveres. Manual installing of packages doesn't help still.

bdach:
I've looked into this some more and it looks as if having the template game target `net5.0` alongside `netstandard2.1` fixes this, but I have no idea why. Upon adding some logging in `RoslynTypeReferenceBuilder`, something seems to trigger the game project into attempting to load `System.ComponentModel.Composition`, which somehow works with lazer (even though it doesn't target `net5.0` explicitly), but fails with the templates.

In the end the game seems to find the dll in `/usr/share/dotnet/shared/Microsoft.NETCore.App/5.0.8/System.ComponentModel.dll` which is a bit weird. Maybe the project has to be aware of `net5.0`'s existence to see some assembly redirects or something? No idea.

Kremator1007:
In my case targeting net5.0 alongiside netstandard2.1 also fixes the issue.

bdach:
[...]

Issue #1270, opened 2017-12-18
Disallow dynamic compilation for nested types


smoogipoo:
It is possible to include nested types in `RequiredFiles` for a TestCase. 

Unfortunately this won't work since `DynamicClassCompiler` will attempt to look up the class filename by the type name, and will fail to do so. Upon failing, it won't recompile that class but the whole process may still succeed, leading to possibly unexpected results.

This should be disallowed and instead throw an exception if a file matching the type can't be found.

smoogipoo:
Moving to low priority backlog since this is basically irrelevant in most of our use cases (just a consideration).

PR #5041, opened 2022-02-18
Replace manual dynamic compilation with .NET Hot Reload

smoogipoo:
Closes https://github.com/ppy/osu-framework/issues/5039
Closes https://github.com/ppy/osu-framework/issues/1270
Closes https://github.com/ppy/osu-framework/issues/4699
Closes https://github.com/ppy/osu-framework/issues/4387
Closes https://github.com/ppy/osu/issues/9225

Notes:
- With Rider, you'll get a popup that you have to click "Apply Changes" in every time a change is made. It also seems to break every now and then and the popup just doesn't show, requiring a restart.
    image
- As suggested by @peppy, I recommend adding a keybind to `Apply Hot Reload Changes` in Rider (defaults to Alt+F10).
- Using `dotnet watch` from CLI in `osu.Framework.Tests` directory will continuously hot-reload.
- I haven't removed the assembly dropdown from the test browser in this PR. Doing so seems to be a larger change than it is at face value.

Self-contained publish size:
- Before: 134M
- After: 116M

---

### Visual test projects now use native .NET 6 "Hot reload"

Over the years we have maintained our own version of hot reload, affectionately named "dynamic compilation". Even after multiple complete rewrites of the system, there are still edge cases where it will unexpectedly fall over due to being too greedy in including what it considers required for the recompile.

The startup cost when running in debug was considerable (around 10-30s initialisation) and the first-compile overhead could also be high (5-60s). In addition, the dependencies required to make it work increased the final assembly size of osu!framework, even for release deploys.

With the introduction of cross-platform support for hot reload in .NET 6 we have made the decision to switch to the natively supported version.

To use this new version:
- From Rider, you'll get a popup that you have to click "Apply Changes" when a change is made to code while running. You can bind a key to `Apply Hot Reload Changes` in Rider (defaults to Alt+F10) to make it quicker to apply changes.
- Running `dotnet watch` from CLI on your test project will automatically watch files and recompile when a hange is seen.

New limitations:
- The limitations of hot reload are [listed here](https://github.com/dotnet/roslyn/blob/main/docs/wiki/EnC-Supported-Edits.md).
- Notably, adding new `override` methods or `class`es are not supported. You can workaround this for the common scenario of prototyping new components by creating subclasses and adding the `override` methods before the initial hot reload.

We are interested in hearing feedback on this change, especially troubling cases where the previous behaviour worked better for you. Hope is that the limitations of the new method are outweighed by the leaner assembly, better performance, and (in general) up-front error when a change can't be applied, rather than an error that can be delayed longer than it would take to run a full recompile/restart.


peppy:
Honestly I think this is better than what we have, even with the limitations. When working on new components, you can even add override methods as long as they are done as part of the introduction of the new class (which is usually how we write them anyway).

I usually hit cmd-s to trigger a framework dynamic compile, so adjusting rider shortcuts to instead to a hot reload removes any manual overhead that may be incurred:

JetBrains Rider-EAP 2022-02-18 at 08 33 49

peppy (approved):

smoogipoo:
Eh, I disagree with the override sentiment. I don't usually do that when it comes to input for example.

But I don't think there is another way to resolve issues like https://github.com/ppy/osu-framework/issues/5039 without rethinking the current DCC and I just don't have time for that right now.

smoogipoo:
Merge at your discretion, I haven't removed the assembly dropdown in the test browser yet.

peppy:
I still run into cases with the current dynamic compilation falls over, I'd say around 10-20% of the time I use it. I'd much rather know up front that it's not going to work, as the overhead of waiting for it to initialise then fail can be 15s (on a good PC) or over 1 minute (as reported by @frenzibyte).

osu.Framework/Properties/AssemblyInfo.cs
```diff
@@ -2,6 +2,11 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Runtime.CompilerServices;
+#if NET6_0_OR_GREATER
+using System.Reflection.Metadata;
+using osu.Framework.Testing;
+
```
smoogipoo, line 8: Don't ask about this newline

peppy (approved):

frenzibyte (commented):
I've thoroughly tested this and it's working pretty amazingly so far for me.

To summarize the limits of it:
 - Can't add new classes.
 - Can't override methods.
 - Can't modify code in generic-typed classes.
 - Can't modify value of constant fields.

And I would say I'm fine with all these limits in favour of the performance of .NET Hot Reload.

Also added to the OP that this closes ppy/osu#9225 upon checking.

---

However, just to note, one thing I noticed while I was testing this is that apparently, applying this arbitrary change to `BeatmapCardNormal` while viewing `TestSceneBeatmapCard` results in `InvalidProgramException: Common Language Runtime detected an invalid program`:
```diff
diff --git a/osu.Game/Beatmaps/Drawables/Cards/BeatmapCardNormal.cs b/osu.Game/Beatmaps/Drawables/Cards/BeatmapCardNormal.cs
index 08befd5340..4d7792697b 100644
--- a/osu.Game/Beatmaps/Drawables/Cards/BeatmapCardNormal.cs
+++ b/osu.Game/Beatmaps/Drawables/Cards/BeatmapCardNormal.cs
@@ -168,8 +168,8 @@ private void load()
                                     Name = @"Bottom content",
                                     RelativeSizeAxes = Axes.X,
                                     AutoSizeAxes = Axes.Y,
-                                    Anchor = Anchor.BottomLeft,
-                                    Origin = Anchor.BottomLeft,
+                                    Anchor = Anchor.CentreLeft,
+                                    Origin = Anchor.CentreLeft,
                                     Children = new Drawable[]
                                     {
                                         idleBottomContent = new FillFlowContainer
```

I'm not yet sure why, nor do I have the time to look into it, so gonna leave it here for now.

osu.Framework/Testing/TestScene.cs
```diff
@@ -29,7 +29,7 @@ namespace osu.Framework.Testing
 {
     [ExcludeFromDynamicCompile]
```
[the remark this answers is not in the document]
smoogipoo: True, but I'm going to leave it as is for the time being (not even obsoleted), at least until one of us works on a new component and sees how it is in practice.

frenzibyte:
Yeah, perhaps it is a platform-specific issue? In fact, I can repro when changing anything in that class apparently. Here's the relevant logs with the exception stack trace from the patch I sent above, in case useful:
```
[runtime] 2022-02-18 11:32:09 [verbose]: 🔸 Step #2 Normal
[runtime] 2022-02-18 11:32:09 [verbose]: 💨 BeatmapCard(TestSceneBeatmapCard) Normal
[runtime] 2022-02-18 11:32:09 [verbose]: 🔸 Step #3 register request handling
[runtime] 2022-02-18 11:32:10 [verbose]: 🔸 Step #4 ensure manager loaded
[runtime] 2022-02-18 11:32:10 [verbose]: 🔸 Step #5 remove map
[runtime] 2022-02-18 11:32:10 [verbose]: 🔸 Step #6 set Red scheme
[runtime] 2022-02-18 11:32:10 [verbose]: 💥 Failed
[runtime] 2022-02-18 11:32:10 [verbose]: ⏳ Currently loading components (0)
[runtime] 2022-02-18 11:32:10 [verbose]: 🧵 Task schedulers
[runtime] 2022-02-18 11:32:10 [verbose]: LoadComponentsAsync (standard) concurrency:4 running:0 pending:0
[runtime] 2022-02-18 11:32:10 [verbose]: LoadComponentsAsync (long load) concurrency:4 running:0 pending:0
[runtime] 2022-02-18 11:32:10 [verbose]: 🎱 Thread pool
[runtime] 2022-02-18 11:32:10 [verbose]: worker:          min 8      max 32,767 available 32,766
[runtime] 2022-02-18 11:32:10 [verbose]: completion:      min 8      max 1,000  available 1,000
[runtime] 2022-02-18 11:32:10 [verbose]: Error on step: System.InvalidProgramException: Common Language Runtime detected an invalid program.
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Game.Beatmaps.Drawables.Cards.BeatmapCardNormal.<>c__DisplayClass16_0.<load>b__0(BeatmapCardContent c)
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Framework.Graphics.DrawableExtensions.With[T](T drawable, Action`1 adjustment) in /Users/salman/Desktop/osu-framework/osu.Framework/Graphics/DrawableExtensions.cs:line 23
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Game.Beatmaps.Drawables.Cards.BeatmapCardNormal.load()
[runtime] 2022-02-18 11:32:10 [verbose]: --- End of stack trace from previous location ---
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Framework.Allocation.BackgroundDependencyLoaderAttribute.<>c__DisplayClass6_1.<CreateActivator>b__4(Object target, IReadOnlyDependencyContainer dc) in /Users/salman/Desktop/osu-framework/osu.Framework/Allocation/BackgroundDependencyLoaderAttribute.cs:line 76
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Framework.Allocation.DependencyActivator.activate(Object obj, IReadOnlyDependencyContainer dependencies) in /Users/salman/Desktop/osu-framework/osu.Framework/Allocation/DependencyActivator.cs:line 75
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Framework.Allocation.DependencyActivator.Activate(Object obj, IReadOnlyDependencyContainer dependencies) in /Users/salman/Desktop/osu-framework/osu.Framework/Allocation/DependencyActivator.cs:line 50
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Framework.Allocation.DependencyContainer.Inject[T](T instance) in /Users/salman/Desktop/osu-framework/osu.Framework/Allocation/DependencyContainer.cs:line 186
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Framework.Graphics.Drawable.InjectDependencies(IReadOnlyDependencyContainer dependencies) in /Users/salman/Desktop/osu-framework/osu.Framework/Graphics/Drawable.cs:line 299
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Framework.Graphics.Containers.CompositeDrawable.InjectDependencies(IReadOnlyDependencyContainer dependencies) in /Users/salman/Desktop/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 82
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Framework.Graphics.Drawable.load(IFrameBasedClock clock, IReadOnlyDependencyContainer dependencies) in /Users/salman/Desktop/osu-framework/osu.Framework/Graphics/Drawable.cs:line 273
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Framework.Graphics.Drawable.Load(IFrameBasedClock clock, IReadOnlyDependencyContainer dependencies, Boolean isDirectAsyncContext) in /Users/salman/Desktop/osu-framework/osu.Framework/Graphics/Drawable.cs:line 252
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Framework.Graphics.Containers.CompositeDrawable.loadChild(Drawable child) in /Users/salman/Desktop/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 288
[runtime] 2022-02-18 11:32:10 [verbose]: at osu.Framework.Graphics.Containers.CompositeDrawable.load(ShaderManager shaders, Nullable`1 cancellation) in /Users/salman/Desktop/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 263
```

peppy:
> 
> To summarize the limits of it:
> 
>  - Can't add new classes.
> 

Notably, nested classes do work.

bdach:
Just going to note that this is going to require osu!-side changes too, because there is [one reference to the dynamic class compiler](https://github.com/ppy/osu/blob/65ffcefdf9ec2e136b426378d986dd629cbd28a6/osu.Game/Tests/Visual/SkinnableTestScene.cs#L46) that is no longer valid after these.

I was going to do some UI work this weekend anyway, so I'll give this a proper spin and see how it behaves. Generally mostly on board with this from the testimonials above, since I can corroborate that the existing setup was pretty slow and flakey for me too, so if this is both faster and has more consistent rules about what will and what won't reload cleanly then that's a big improvement in my eyes.

bdach:
Also, for other ideavim users that like to use ex mode, you can do:

```
:command R action RiderDebuggerApplyEncChagnes
```

(note - that's not a typo above, that's what it's actually called), and then you can do `:R` from ex mode to reload changes. I got used to doing `:w` to reload my changes, and that doesn't work even if I bind ctrl-s to hot reload.

bdach:
I changed key mappings to bind hot reload to ctrl-s, and then opened the keymap config in a text editor and looked at the action ID there 😅

bdach:
So I've spent these two past days of UI dev with a local checkout of this branch. To be quite honest my feelings have cooled off quite a bit over these two days, because it's less of an improvement than I'd hoped it would be. Warning, essay incoming.

In general, yes the failures are more consistent in general, but with hot reload you can get away with a lot less than the previous setup would allow. When it works it works incredibly well, but when it doesn't, it forces a full rebuild - and I don't know if it's just me, but I don't remember those being as slow as I've found them this weekend.

It may be some .NET 6 thing, it may be due to having to build framework every time, it may also be due to me having resharper build fully off (because it sometimes will fail to invalidate some cache and I will see stale code running) - but either way, my full build times were about a minute every time. Which is worse in some cases than the previous state because while yes, the first dynamic compilation was always very slow, the following ones would be much quicker. So in a way, quick failures are preferable to slow surefire compilations, in some situations.

Also, sometimes on things that we have in framework, failures *will* be silent. Examples include: adding a new `[Resolved]` property to an existing class at runtime, or adding a new `[Cached]` property to an existing class at runtime; I've had those take no effect. I can understand this for sure - there is a very loose and indirection-y connection there, so the runtime can't possibly know what else it has to recompile, but it's unfortunate as it doesn't result in a loud error like all of the other surefire hot reload failures.

All that said, the previous setup was somewhat of a maintenace nightmare, and I've started adapting to hot reload somehow by running through a mental checklist of what I may need in advance, and then adding stubs for the initial compilation that I will then use with hot reload. I think once I get a handle on being better at that, this has the potential to be better than the previous state of affairs, but I'm not personally there yet. Unfortunately that doesn't hold for general "protectedness" and inheritance usage that we have a lot of in framework and game - there a recompilation hit will need to happen almost every time.

In summary I still think it's better to merge this and test this over a longer period of time. At worst we can revert to the previous state of affairs; at best, after some adapting this may turn out to be better than what we had. And given that this solution has .NET platform buy-in, maybe hot reload can itself also improve over time as the .NET team works on it?

bdach (approved):

jai-x:
Just adding in that `dotnet watch` has some (poorly) documented config options for exuding files and reference from watch, which *might* result in compile speedup for certain cases: https://docs.microsoft.com/en-us/aspnet/core/tutorials/dotnet-watch?view=aspnetcore-6.0#opt-out-of-files-to-be-watched

Directory.Build.props
```diff
@@ -21,7 +21,6 @@
   <ItemGroup Label="Code Analysis">
     <PackageReference Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="3.3.3" PrivateAssets="All" />
     <AdditionalFiles Include="$(MSBuildThisFileDirectory)CodeAnalysis\BannedSymbols.txt" />
-    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="6.0.0" PrivateAssets="All" />
```
smoogipoo, line 24: I'm not sure about this one. Can someone without .NET 7 SDK installed try undoing this change and testing `dotnet watch`?

peppy: Works fine for me

```csharp

~/Projects
❯ dotnet --version
6.0.200


[performance] 2022-02-21 03:54:12 [verbose]: TextureAtlas initialised (1024x1024)
[runtime] 2022-02-21 03:54:13 [verbose]: 🔸 Step #2 HardwareDecode
[runtime] 2022-02-21 03:54:13 [verbose]: 💨 Video(TestSceneVideo) HardwareDecode
[runtime] 2022-02-21 03:54:13 [verbose]: 🔸 Step #3 Reset clock
watch : File changed: /Users/dean/Projects/osu-framework/osu.Framework.Tests/Visual/Sprites/TestSceneVideo.cs.
watch : Hot reload of changes succeeded.
[runtime] 2022-02-21 03:54:28 [verbose]: 🔸 Step #2 HardwareDecode
[runtime] 2022-02-21 03:54:28 [verbose]: 💨 Video(TestSceneVideo) HardwareDecode
[runtime] 2022-02-21 03:54:28 [verbose]: 🔸 Step #3 Reset clock
watch : File changed: /Users/dean/Projects/osu-framework/osu.Framework.Tests/Visual/Sprites/TestSceneVideo.cs.
watch : Hot reload of changes succeeded.
[runtime] 2022-02-21 03:54:35 [verbose]: 🔸 Step #2 HardwareDecode
[runtime] 2022-02-21 03:54:35 [verbose]: 💨 Video(TestSceneVideo) HardwareDecode
[runtime] 2022-02-21 03:54:35 [verbose]: 🔸 Step #3 Reset clock
```

peppy (approved):

Directory.Build.props
smoogipoo: Okay, will just have to revert this then, and I'll uninstall .NET 7 locally.

PR #5043, opened 2022-02-18
Update FFmpeg libraries with LGPL versions and move out macOS dylibs to `osx` for arm64 support

nekodex:
This PR replaces the currently included FFmpeg libraries (of dubious origins) with ones explicitly built _without_ `--enable-gpl` and `--enable-nonfree` flags to ensure LGPL compliance.

Build scripts are included for macOS, as building universal (x86_64+arm64) dylibs is kind of annoying.

Windows libraries were cross-compiled on linux using mingw-w64 (as this was easiest for me). I didn't bother creating scripts for that though, as I feel it wouldn't be super useful for Windows users (who would likely be the target for building dlls).

Testing:
- [x] macOS(x86_64)
- [x] macOS(arm64)
- [x] Windows(64-bit)
- [x] Windows(32-bit) (not tested but should be okay to leave until someone runs into a potential problem)

```
bass => 2.4.16.7 (unchanged)
bass_fx => 2.4.12.6 (unchanged)
bassmix => 2.4.11.2 (old: 2.4.11.1)
```


peppy:
@nekodex how were you testing this one? i tried using the method @smoogipoo [proposed](https://ptb.discord.com/channels/188630481301012481/589331078574112768/943448887186841621) and libbass seems to be missing:

```csharp

[runtime] 2022-02-21 09:15:11 [error]: System.DllNotFoundException: Unable to load shared library 'bassmix' or one of its dependencies. In order to help diagnose loading problems, consider setting the DYLD_PRINT_LIBRARIES environment variable: dlopen(libbassmix, 0x0001): tried: 'libbassmix' (no such file), '/usr/local/lib/libbassmix' (no such file), '/usr/lib/libbassmix' (no such file), '/Users/dean/Projects/osu-framework/osu.Framework.Tests/bin/Debug/net6.0/libbassmix' (no such file), '/usr/local/lib/libbassmix' (no such file), '/usr/lib/libbassmix' (no such file)
[runtime] 2022-02-21 09:15:11 [error]: at ManagedBass.Mix.BassMix.BASS_Mixer_GetVersion()
[runtime] 2022-02-21 09:15:11 [error]: at ManagedBass.Mix.BassMix.get_Version()
[runtime] 2022-02-21 09:15:11 [error]: at osu.Framework.Audio.AudioManager.setAudioDevice(Int32 deviceIndex) in /Users/dean/Projects/osu-framework/osu.Framework/Audio/AudioManager.cs:line 325
[runtime] 2022-02-21 09:15:11 [error]: at osu.Framework.Audio.AudioManager.setAudioDevice(String deviceName) in /Users/dean/Projects/osu-framework/osu.Framework/Audio/AudioManager.cs:line 295
[runtime] 2022-02-21 09:15:11 [error]: at osu.Framework.Audio.AudioManager.<onDevicesChanged>b__39_0() in /Users/dean/Projects/osu-framework/osu.Framework/Audio/AudioManager.cs:line 214

```

peppy:
Turns out my nuget was somehow resolving an ancient `1.0.0` release no matter how i changed my nuget lookup order. Even after deleting global caches.

So I've just pushed this as a new version on nuget.org so we can move forward (https://www.nuget.org/packages/ppy.osu.Framework.NativeLibs/2022.222.0). Saves us doing two releases to get CI to build anyway.

peppy:
Interestingly, with the new compiles nativelibs as a whole reduced from 63mb to 34mb, which is a pretty nice saving (likely due to the reduced footprint without the encoder portions being included).

I've pushed an update to the reference for the package that I've already published. I've tested on macOS and windows (x64 only). Should be enough to push forward with this change.

peppy (approved):

smoogipoo:
Just going to note that this upgrades `libbassmix` according to my addition in the description.

PR #5044, opened 2022-02-19
Add a way to exit/minimize the app on Android when back is pressed

Susko3:
Relevant: https://github.com/ppy/osu/issues/16899.

This should live at framework, as it's useful for any app that wants to follow Android best practices for the back button.

[`MoveTaskToBack()`](https://developer.android.com/reference/android/app/Activity#moveTaskToBack(boolean)) available since API 1. Calling this will put the activity in the background and return to the last app (eg. the home screen if the game was opened from there, or the file manager if osu! was opened from beatmap import).

Tested on:
- Realme 6
- LG G7 Fit

Breaking changes
---

Pressing the back key on Android will now exit/minimize the app and return to home screen.
To restore the old behaviour of treating the back key as `InputKey.Escape`, add your own bindable to `GameHost.AllowExitingAndroid` and set it to `false`.

To be in line with mobile UX, your app should allow exiting when it's on the initial screen (eg. when backing out is no longer possible / when backing out would close the app on desktop platforms).


osu.Framework/Platform/GameHost.cs
```diff
@@ -70,6 +70,15 @@ public abstract class GameHost : IIpcHost, IDisposable
...
         public readonly AggregateBindable<bool> AllowScreenSuspension = new AggregateBindable<bool>((a, b) => a & b, new Bindable<bool>(true));
 
+        /// <summary>
+        /// Allow exiting the app on Android when the back button is pressed.
+        /// </summary>
+        /// <remarks>
+        /// If <c>true</c>, pressing back will return to the home screen (putting the app in background and suspending it).
+        /// If <c>false</c>, the back button will be treated as <see cref="InputKey.Escape"/>.
+        /// </remarks>
+        public readonly AggregateBindable<bool> AllowExitingAndroid = new AggregateBindable<bool>((a, b) => a & b, new Bindable<bool>(true));
```
peppy, line 80: Might be better to rename to `AllowSuspendingAndroid` (`AllowSuspendToBackground` without the android suffix maybe?) as it isn't really an exit operation. Not 100% on having android specific variables in here, but I guess it'll have to do for now.

Susko3: Just checked and the Android docs seem to call it "stopping the activity" and "moving to background". So `AllowSuspendToBackground` could work. Maybe there's a better name that would encompass the "when back is pressed" part of this.

How about `AllowNativeBackNavigation`?

frenzibyte, line 80: I'm not really sure on this approach. [Looking at the changes osu!-side](https://github.com/ppy/osu/pull/16924/files), this means that on every game implementation, this bindable will have to be set to true if the next `Back` key press will not be handled by the game.

Couldn't this instead still follow the `InputKey.Escape` path as usual, but when the `Game` reaches the point of exiting (via calling `GameHost.Exit`), `GameHost` would instead define a path for suspending the game by moving it to the background (if `CanExit = false`)? 

I think that could work better without having this and/or any game-side implementation.

peppy: I haven't looked into this in super detail yet but agree that if it can work similar to the allow exit flow that would be preferable.

Susko3: Have updated to do as requested. The osu! side flow had to be adjusted not to close the game when suspending it to background.

osu.Framework/Platform/GameHost.cs
```diff
@@ -583,12 +592,26 @@ private set
...
+        /// Suspends and minimizes the game to background.
+        /// </summary>
+        /// <remarks>
+        /// This is provided as an alternative to <see cref="Exit"/> on hosts that can't exit (see <see cref="CanExit"/>).
+        /// Should only be called if <see cref="CanSuspendToBackground"/> is <c>true</c>.
+        /// </remarks>
+        public virtual void SuspendToBackground()
+        {
+            if (!CanSuspendToBackground)
+                throw new InvalidOperationException($"Tried to call {nameof(SuspendToBackground)}() on a host that doesn't support it (see {nameof(CanSuspendToBackground)}).");
```
bdach, line 612: This throw is a bit strange because you can bypass it entirely by overriding `SuspendToBackground()` and just not calling `base`. Note that `Exit()` does not have the same issue because that one's not virtual.

osu.Framework/Platform/GameHost.cs
```diff
@@ -104,8 +104,17 @@ public abstract class GameHost : IIpcHost, IDisposable
...
+        /// <remarks>Also see <see cref="CanSuspendToBackground"/>.</remarks>
         public virtual bool CanExit => true;
 
+        /// <summary>
+        /// Whether this host can suspend and minimize to background.
+        /// </summary>
+        /// <remarks>
+        /// This and <see cref="SuspendToBackground"/> are an alternative way to exit on hosts that have <see cref="CanExit"/> <c>false</c>.
+        /// </remarks>
+        public virtual bool CanSuspendToBackground => false;
```
bdach, line 116: Dunno how to feel about this - is there any chance that android could just use the exit flow rather than invent a new one?

Susko3: In the context of the [osu!-side changes](https://github.com/ppy/osu/pull/16924/files#diff-858b737fcbe83cc82e88df78852a4ad4c4298b44fc29a61b84946c0bad838049R172), having it be merged into `Exit()` seems wrong, as it's not really exiting. It also has different UX implications, and you can't treat Android's "exit" as a desktop exit. Eg. osu! exits all game screens, and only then exits the host, which absolutely does not work with Android.

peppy: One thing to note is that we may be able to use this behaviour on desktop for the "hide to tray" functionality (aka "boss key") when we get to adding that back. Although minimise may be a closer mapping for the suspend action on desktop.

osu.Framework/Platform/GameHost.cs
peppy: Agree that this shouldn't throw. If anything having a `success` return value would be best.

peppy (changes requested):
Seems much better structurally.

peppy (approved):

bdach (approved):
appears to work

frenzibyte (approved):

PR #5048, opened 2022-02-22
Allow Alt+Enter to still toggle fullscreen when a textbox is focused

peppy:
This looks to be pretty standard/expected behaviour.

Resolves issue mentioned at https://github.com/ppy/osu/discussions/16950.


osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -1002,7 +1002,11 @@ protected override bool OnKeyDown(KeyDownEvent e)
 
                 case Key.KeypadEnter:
                 case Key.Enter:
-                    // same rationale as comment above.
+                    // alt-enter is commonly used to toggle fullscreen.
+                    if (e.AltPressed)
+                        return false;
+
+                    // same rationale as comment in case statement comment above.
```
frenzibyte, line 1009:
```suggestion
                    // same rationale as comment in case statement above.
```

bdach (approved):

PR #5049, opened 2022-02-22
Fix sub-menu getting an invalid position when parent menu is transferred to another drawable

smoogipoo:
Fixes https://github.com/ppy/osu/issues/16953

I have other fixes/cleanups to this class coming up, but I want to test them out a little bit more while fixing the root cause of this issue.

The cause of this issue is that osu!-side menus fade out when closed: https://github.com/ppy/osu/blob/ebdffb3cd91e89466c0c6c00dcdbb8da814ee4e5/osu.Game/Graphics/UserInterface/OsuContextMenu.cs#L61-L69, and so when another Drawable is clicked and the parent menu is "transferred" to it (i.e. it adopts that new Drawable's elements) there is a period of time where the submenu is fading out and receiving updates.

During this time, if the submenu tries to reposition itself beside the `MenuItem` which opened it, it would be doing so with a Drawable that doesn't exist and would receive an undefined position. If that position ends up being outside of the screen, then the menu gets permanently masked away.

The easiest way I've found to repro this in-game is by right clicking on the beatmap set panel at the middle of the screen, and then right clicking on the first beatmap set panel above it. You can see a glimpse of what's happening in the following screenshot - notice the submenu in the top-left corner of the screen:
image

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneContextMenu.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneContextMenu.cs
@@ -32,6 +32,48 @@ namespace osu.Framework.Tests.Visual.UserInterface
         [SetUp]
         public void Setup() => Schedule(Clear);
 
+        /// <summary>
+        /// Tests an edge case where the submenu is visible and continues updating for a short period of time after right clicking another item.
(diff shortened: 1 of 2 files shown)


bdach (approved):
Makes sense to me

PR #5052, opened 2022-02-25
Add `IsTablet` to framework

Susko3:
Logic taken from [`OsuGameActivity`](https://github.com/ppy/osu/blob/84171962e56faa9b5310450d01d0faedeffd2146/osu.Android/OsuGameActivity.cs#L74-L77).

The value is updated once on startup and when the (screen) configuration changes (see `DEFAULT_CONFIG_CHANGES`).

Not tested on folding devices or other devices that can change screens, but `OnConfigurationChanged` should be the right entry point for that: https://developer.android.com/guide/topics/large-screens/make-apps-fold-aware#window_size_changes.

We may potentially need to use `gameView.Display` to get the actual current display, but this display is `null` on startup.

### DIFF
--- a/osu.Framework.Android/AndroidGameActivity.cs
+++ b/osu.Framework.Android/AndroidGameActivity.cs
@@ -1,9 +1,11 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using Android.App;
 using Android.Content;
 using Android.Content.PM;
+using Android.Graphics;
 using Android.OS;
 using Android.Runtime;
 using Android.Views;
@@ -35,6 +37,16 @@ namespace osu.Framework.Android
         /// Whether this <see cref="AndroidGameActivity"/> is active (in the foreground).
         /// </summary>
         public BindableBool IsActive { get; } = new BindableBool();
+
+        /// <summary>
+        /// Whether the current screen is tablet-sized.
+        /// </summary>
+        /// <remarks>
+        /// This can potentially change when switching displays (eg. on folding phones).
+        /// Screens with at least medium width (as defined in https://developer.android.com/guide/topics/large-screens/support-different-screen-sizes)
+        /// are considered tablet-sized.
+        /// </remarks>
+        public BindableBool IsTablet { get; } = new BindableBool();
 
         /// <summary>
         /// The visibility flags for the system UI (status and navigation bars)
@@ -76,6 +88,8 @@ namespace osu.Framework.Android
             SetContentView(gameView = new AndroidGameView(this, CreateGame()));
 
             UIVisibilityFlags = SystemUiFlags.LayoutFlags | SystemUiFlags.ImmersiveSticky | SystemUiFlags.HideNavigation | SystemUiFlags.Fullscreen;
+
+            updateIsTablet();
 
             Debug.Assert(Window != null);
 
@@ -152,6 +166,25 @@ namespace osu.Framework.Android
         public override bool OnKeyLongPress([GeneratedEnum] Keycode keyCode, KeyEvent e)
         {
             return gameView.OnKeyLongPress(keyCode, e);
+        }
+
+        public override void OnConfigurationChanged(global::Android.Content.Res.Configuration newConfig)
+        {
+            base.OnConfigurationChanged(newConfig);
+            updateIsTablet();
+        }
+
+        /// <summary>
+        /// Update <see cref="IsTablet"/> value.
+        /// </summary>
+        private void updateIsTablet()
+        {
+            if (WindowManager?.DefaultDisplay == null || Resources?.DisplayMetrics == null) return;
+
+            Point displaySize = new Point();
+            WindowManager.DefaultDisplay.GetSize(displaySize);
+            float smallestWidthDp = Math.Min(displaySize.X, displaySize.Y) / Resources.DisplayMetrics.Density;
+            IsTablet.Value = smallestWidthDp >= 600f;
         }
     }
 }


osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -153,5 +167,24 @@ public override bool OnKeyLongPress([GeneratedEnum] Keycode keyCode, KeyEvent e)
         {
             return gameView.OnKeyLongPress(keyCode, e);
         }
+
+        public override void OnConfigurationChanged(global::Android.Content.Res.Configuration newConfig)
```
bdach, line 171: Is the `global::` here necessary? It's the first usage of that language construct in the entire project and I think I'd prefer it gone if feasible, or replaced with an aliased `using` if it is required.

Susko3: image

Even with a `using Configuration = ...` alias, it's resolved to `osu.Framework.Configuration`. `using` renaming does work, if that's the path we want to take.

This is probably the cleanest we can get without renaming:

```diff
diff --git a/osu.Framework.Android/AndroidGameActivity.cs b/osu.Framework.Android/AndroidGameActivity.cs
index 2edfb84ab..daef3e083 100644
--- a/osu.Framework.Android/AndroidGameActivity.cs
+++ b/osu.Framework.Android/AndroidGameActivity.cs
@@ -12,6 +12,7 @@
 using ManagedBass;
 using osu.Framework.Bindables;
 using Debug = System.Diagnostics.Debug;
+using Res = Android.Content.Res;

 namespace osu.Framework.Android
 {
@@ -168,7 +169,7 @@ public override bool OnKeyLongPress([GeneratedEnum] Keycode keyCode, KeyEvent e)
             return gameView.OnKeyLongPress(keyCode, e);
         }

-        public override void OnConfigurationChanged(global::Android.Content.Res.Configuration newConfig)
+        public override void OnConfigurationChanged(Res.Configuration newConfig)
         {
             base.OnConfigurationChanged(newConfig);
             updateIsTablet();
```

bdach: I'd like to think `using static` could also help here?

Susko3: `using static` doesn't make a difference.

The FQN is already used for another class clashing with the `Configuration` namespace.

https://github.com/ppy/osu-framework/blob/8379393786d75bf4b39056409c8806a35723fe91/osu.Framework/IO/Stores/GlyphStore.cs#L147

With `Android`, it's clashing with `osu.Framework.Android` as well, so the `global::` is required.

Probably the only way to have it be just `Configuration` is to have the `using` nested (https://github.com/dotnet/csharplang/discussions/3669#discussioncomment-146007):

```diff
diff --git a/osu.Framework.Android/AndroidGameActivity.cs b/osu.Framework.Android/AndroidGameActivity.cs
index 2edfb84ab..6b4c6eaf2 100644
--- a/osu.Framework.Android/AndroidGameActivity.cs
+++ b/osu.Framework.Android/AndroidGameActivity.cs
@@ -12,9 +12,13 @@
 using ManagedBass;
 using osu.Framework.Bindables;
 using Debug = System.Diagnostics.Debug;
+using Res = Android.Content.Res;

 namespace osu.Framework.Android
 {
+    // using global::Android.Content.Res;
+    using Res;
+
     // since `ActivityAttribute` can't be inherited, the below is only provided as an illustrative example of how to setup an activity for best compatibility.
     [Activity(ConfigurationChanges = DEFAULT_CONFIG_CHANGES, LaunchMode = DEFAULT_LAUNCH_MODE, MainLauncher = true)]
     public abstract class AndroidGameActivity : Activity
@@ -168,7 +172,7 @@ public override bool OnKeyLongPress([GeneratedEnum] Keycode keyCode, KeyEvent e)
             return gameView.OnKeyLongPress(keyCode, e);
         }

-        public override void OnConfigurationChanged(global::Android.Content.Res.Configuration newConfig)
+        public override void OnConfigurationChanged(Configuration newConfig)
         {
             base.OnConfigurationChanged(newConfig);
             updateIsTablet();
```

bdach: Out of all these aliasing the namespace to `Res` (so the first diff) seems most tolerable. That last one... overcomplicated.

osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -153,5 +168,24 @@ public override bool OnKeyLongPress([GeneratedEnum] Keycode keyCode, KeyEvent e)
...
+        /// <summary>
+        /// Update <see cref="IsTablet"/> value.
+        /// </summary>
+        private void updateIsTablet()
+        {
+            if (WindowManager?.DefaultDisplay == null || Resources?.DisplayMetrics == null) return;
+
+            Point displaySize = new Point();
+            WindowManager.DefaultDisplay.GetSize(displaySize);
+            float smallestWidthDp = Math.Min(displaySize.X, displaySize.Y) / Resources.DisplayMetrics.Density;
```
peppy, line 187: I recall that you found this calculation method somewhere in the android docs. Would be handy to include an inline comment pointing to that page if possible.

osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -182,8 +182,11 @@ private void updateIsTablet()
         {
             if (WindowManager?.DefaultDisplay == null || Resources?.DisplayMetrics == null) return;
 
+            // Width computed according to https://developer.android.com/guide/topics/large-screens/support-different-screen-sizes#java.
+            // Except that we use Display.GetRealSize() instead of WindowMetrics to support older devices and to get actual display size.
+
             Point displaySize = new Point();
-            WindowManager.DefaultDisplay.GetSize(displaySize);
+            WindowManager.DefaultDisplay.GetRealSize(displaySize);
```
bdach, line 189: This method is [marked as deprecated in API 31](https://developer.android.com/reference/android/view/Display#getRealSize(android.graphics.Point)) in favour of `getCurrentWindowMetrics()`. I would think that we do want to use the new one where applicable, and only fall back to `GetRealSize()` on devices below API 30.

Susko3: I agree, we'd use `getMaximumWindowMetrics()` then.

This requires bumping `monoandroid` to `12.0`. No so simple to do, as some things we use are deprecated. At some point, I'll bump it and do a full deprecation pass on osu!(framework), which will include this change.

PR #5053, opened 2022-02-28
Disable running `mono-cil-strip` during iOS builds

frenzibyte:
Should fix startup crashes seen on iOS release builds done on actual devices (https://github.com/ppy/osu-framework/pull/4263, [discord #1](https://ptb.discord.com/channels/188630481301012481/188630652340404224/908751597822160947), [discord #2](https://ptb.discord.com/channels/188630481301012481/188630652340404224/947140450412802198)). 

The underlying reason behind `PlatformActionContainer.host` being null in that PR (and `InputManager.Host` in the discord ones) was because `mono-cil-strip` appeared to be stripping out the `ResolvedAttribute` from those properties, for one reason or another, causing it to no longer be populated with the actual `GameHost` instance. 

This appeared again in the latest (pulled) osu! release and for me locally, which was a great opportunity to look deep into the [monotouch source code](https://github.com/xamarin/xamarin-macios/tree/main/tools). 

And after a lot of trial-and-error attempts, I have figured out that disabling the `mono-cil-strip` step was the way to fix that:

CleanShot 2022-02-28 at 18 10 52
CleanShot 2022-02-28 at 18 11 32
CleanShot 2022-02-28 at 18 12 25
(https://github.com/xamarin/xamarin-macios/blob/f1899bc25dd4524957e9c743f084541722f0f9de/tools/mtouch/mtouch.cs#L413)
(https://github.com/xamarin/xamarin-macios/blob/a812ce4242e1ac188760286fa01dfc511dfc07db/tools/mtouch/Application.mtouch.cs#L1638-L1643)
(notice how it's automatically disabled on non-device/simulator builds and debug builds, which explains why it only happens on device and release builds)

I have also reverted the workaround from #4263 now that the underlying issue has been identified and resolved with a better workaround (I couldn't do a straight PR revert as the changes there conflict with master, so had to do it manually).

As for reporting this upstream, I'm not sure how I can begin on isolating this pretty random issue and we will evenutally switch to .NET 6 iOS anyways, so working around this more generally should be enough for now.


peppy (approved):
awesome investigatory work. let's give this a try.

Issue #5055, opened 2022-03-02
Changing displays via `CurrentDisplayBindable` doesn't work


peppy:
I was looking to add [display selection](https://github.com/ppy/osu/discussions/17039) to osu!, but it using this bindable wasn't working at all.

It looks like the intention of this bindable was to have it settable. I do remember this working in the past.

A quick attempt to make it work is this diff:

```diff
diff --git a/osu.Framework/Platform/SDL2DesktopWindow.cs b/osu.Framework/Platform/SDL2DesktopWindow.cs
index 73b5e953d..01436bee8 100644
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -1174,7 +1174,11 @@ public void SetupWindow(FrameworkConfigManager config)
             };
 
             config.BindWith(FrameworkSetting.LastDisplayDevice, windowDisplayIndexBindable);
-            windowDisplayIndexBindable.BindValueChanged(evt => CurrentDisplay = Displays.ElementAtOrDefault((int)evt.NewValue) ?? PrimaryDisplay, true);
+            windowDisplayIndexBindable.BindValueChanged(evt =>
+            {
+                CurrentDisplay = Displays.ElementAtOrDefault((int)evt.NewValue) ?? PrimaryDisplay;
+                pendingWindowState = windowState;
+            }, true);
 
             sizeFullscreen.ValueChanged += evt =>
             {

```

This should be enough to show why it isn't working, but still has issues with various OS/display modes. At very least this should work reliably on windows.

In my testing so far:
- Works with macOS windowed, but not borderless or fullscreen
- Works with seemingly all modes on windows, but may be triggering more mode changes than expected. I've only tested on a VM so it's hard to say without further testing.

Not sure whether I should PR the above to at least have it hooked up, or whether more effort should be spent to ensure it works properly.

bdach:
I'd start with PRing the above to move the dependent game-side PR along. Always better to have something in place that's at least partially working rather than nothing.

PR #5056, opened 2022-03-03
Handle changes to `CurrentDisplay` to allow switching active display

peppy:
This used to work and likely regressed with the SDL implementation at some point.  Note that this implementation is MVP and may not work in certain configurations or on certain platforms, but is better than nothing (in theory).

See https://github.com/ppy/osu-framework/issues/5055.

Worth noting that this is now adding an extra pending mode change request on startup due to the initial trigger flag. In my limited testing this didn't seem to adversely affect anything on windows, but I'd appreciate testing on other platforms (and even on windows to confirm this).


bdach:
On linux this only seems to work when windowed, although this is likely related to the fact that I have two displays with different resolutions. Either way this is better than nothing, platform kinks can get worked out later.

bdach (approved):

PR #5057, opened 2022-03-05
Video playback and decoding fixes

Pizzabelly:
This fixes some issues I noticed in a simple video playback program using osu-framework.

1. If `UploadComplete` was false on the first attempt to display a frame, it wouldn't retry even after `UploadComplete` became true.
2. The check in `Video` to see if there was a seek after the video had finished would fire before the video was actually done playing, causing the video to jump around at the end. This was 2 issues. First, while the decoder state was EOF there could still be available frames. Second, checking if `PlaybackPosition` is less than `VideoDecoder.Duration` will be true even after the last frame is shown.
3. `Video` would not show the last couple frames or only return them after a seek.  This was because the decoder was not flushed on EOF.

### DIFF
--- a/osu.Framework/Graphics/Video/VideoDecoder.cs
+++ b/osu.Framework/Graphics/Video/VideoDecoder.cs
@@ -151,6 +151,7 @@ namespace osu.Framework.Graphics.Video
 
             decoderCommands.Enqueue(() =>
             {
+                ffmpeg.avcodec_flush_buffers(codecContext);
                 ffmpeg.av_seek_frame(formatContext, stream->index, (long)(targetTimestamp / timeBaseInSeconds / 1000.0), AGffmpeg.AVSEEK_FLAG_BACKWARD);
                 skipOutputUntilTime = targetTimestamp;
                 State = DecoderState.Ready;
@@ -487,21 +488,13 @@ namespace osu.Framework.Graphics.Video
 
                 if (packet->stream_index == stream->index)
                 {
-                    // send the packet for decoding.
-                    int sendPacketResult = ffmpeg.avcodec_send_packet(codecContext, packet);
+                    int sendPacketResult = sendPacket(receiveFrame, packet);
 
-                    // Note: EAGAIN can be returned if there's too many pending frames, which we have to read,
-                    // otherwise we would get stuck in an infinite loop.
-                    if (sendPacketResult == 0 || sendPacketResult == -AGffmpeg.EAGAIN)
+                    // keep the packet data for next frame if we didn't send it successfully.
+                    if (sendPacketResult == -AGffmpeg.EAGAIN)
                     {
-                        readDecodedFrames(receiveFrame);
-
-                        // keep the packet data for next frame if we didn't send it successfully.
-                        if (sendPacketResult != 0)
-                            unrefPacket = false;
+                        unrefPacket = false;
                     }
-                    else
-                        Logger.Log($"Failed to send avcodec packet: {getErrorMessage(sendPacketResult)}");
                 }
 
                 if (unrefPacket)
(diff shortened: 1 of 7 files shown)


osu.Framework/Graphics/Video/VideoDecoder.cs
```diff
@@ -487,28 +488,23 @@ private void decodeNextFrame(AVPacket* packet, AVFrame* receiveFrame)
...
-                        Logger.Log($"Failed to send avcodec packet: {getErrorMessage(sendPacketResult)}");
                 }
 
                 if (unrefPacket)
                     ffmpeg.av_packet_unref(packet);
             }
             else if (readFrameResult == AGffmpeg.AVERROR_EOF)
             {
+                // Flush decoder.
+                sendPacket(receiveFrame, null);
```
bdach, line 506: For reference if anyone else attempts reviewing this: [ffmpeg docs](https://www.ffmpeg.org/doxygen/3.4/group__lavc__decoding.html#ga58bc4bf1e0ac59e27362597e467efff3)

> `avpkt`
> The input AVPacket. (...) It can be NULL (or an AVPacket with data set to NULL and size set to 0); in this case, it is considered a flush packet, which signals the end of the stream. Sending the first flush packet will return success. Subsequent ones are unnecessary and will return AVERROR_EOF. If the decoder still has frames buffered, it will return them after sending a flush packet.

osu.Framework/Graphics/Video/VideoTexture.cs
```diff
@@ -117,8 +115,19 @@ protected override void DoUpload(ITextureUpload upload, IntPtr dataPointer)
...
+                if (i == 0)
+                {
+                    width = videoUpload.Frame->width;
+                    height = videoUpload.Frame->height;
+                }
+                else
+                {
+                    width = (videoUpload.Frame->width + 1) / 2;
+                    height = (videoUpload.Frame->height + 1) / 2;
+                }
```
bdach, line 128: Can this be split to a method or something given that it shows up repeated verbatim in 2 places now?

Pizzabelly: Is what I just added good?

bdach: I think so? I had a bit of a slight teeth pain on the usage of `uint` because it's usually really awkward to work with in C# but here it's mostly not so I'm able to look away, personally.

bdach (approved):

osu.Framework/Graphics/Video/VideoTexture.cs
```diff
@@ -112,13 +101,14 @@ protected override void DoUpload(ITextureUpload upload, IntPtr dataPointer)
...
+            for (uint i = 0; i < textureIds.Length; i++)
             {
                 GLWrapper.BindTexture(textureIds[i]);
 
-                GL.PixelStore(PixelStoreParameter.UnpackRowLength, videoUpload.Frame->linesize[(uint)i]);
-                GL.TexSubImage2D(TextureTarget2d.Texture2D, 0, 0, 0, videoUpload.Frame->width / (i > 0 ? 2 : 1), videoUpload.Frame->height / (i > 0 ? 2 : 1),
-                    PixelFormat.Red, PixelType.UnsignedByte, (IntPtr)videoUpload.Frame->data[(uint)i]);
+                GL.PixelStore(PixelStoreParameter.UnpackRowLength, videoUpload.Frame->linesize[i]);
+
+                GL.TexSubImage2D(TextureTarget2d.Texture2D, 0, 0, 0, videoUpload.GetPlaneWidth(i), videoUpload.GetPlaneHeight(i),
```
smoogipoo, line 110: Note: Slight change in semantics here for `i > 0`.
Before: `videoUpload.Frame->width / 2`
After: `(videoUpload.Frame->width + 1) / 2`

smoogipoo (approved):

osu.Framework/Graphics/Video/VideoTexture.cs
Pizzabelly: I'm pretty sure this is correct. Assuming we are given a video with uneven width or height that those are the dimensions of the luma component, we would need to round up the chroma dimensions in order to include the values for the pixels on the edge.

PR #5059, opened 2022-03-05
Fix potential out-of-order execution with `Drawable` scheduling method

frenzibyte:
In master, performing `Drawable.Schedule` with no transform delay prior to a regular `Scheduler.Add` does not make it execute in order, as the first gets enqueued to `timedTasks` while the second gets enqueued to `runQueue` directly, causing the second to execute before the first.

At first I was going to resolve this by making `Scheduler.AddDelayed(..., 0)` enqueue to `runQueue` directly, but looking at it again, I felt that to be weird and unnecessary logic compared to updating usages itself to use `Add` if there's no delay.

I'm honestly not sure which to pick at this point. For reference, here's a branch of the alternative direction (https://github.com/frenzibyte/osu-framework/compare/drawable-schedule-no-delay).

- [x] Depends on #5058

### DIFF
--- a/osu.Framework.Tests/Graphics/TestSceneDrawableScheduling.cs
+++ b/osu.Framework.Tests/Graphics/TestSceneDrawableScheduling.cs
@@ -1,0 +1,45 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using NUnit.Framework;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Testing;
+using osu.Framework.Tests.Visual;
+using osu.Framework.Threading;
+
+namespace osu.Framework.Tests.Graphics
+{
+    [HeadlessTest]
+    public class TestSceneDrawableScheduling : FrameworkTestScene
+    {
+        /// <summary>
+        /// Ensures scheduled delegates between <see cref="Drawable.Schedule"/> and <see cref="Scheduler.Add(Action, bool)"/> with no delays execute in correct order.
+        /// </summary>
+        [Test]
+        public void TestExecutionOrder([Values] bool afterChildren)
+        {
+            Container container = null;
+            bool firstExecutedInOrder = false;
+            bool secondExecutedInOrder = false;
+
+            AddStep("load container", () => Child = container = new Container());
+            AddStep("schedule delegates", () =>
+            {
+                if (!afterChildren)
+                {
+                    container.Schedule(() => firstExecutedInOrder = true);
+                    container.Scheduler.Add(() => secondExecutedInOrder = firstExecutedInOrder);
+                }
+                else
+                {
+                    container.ScheduleAfterChildren(() => firstExecutedInOrder = true);
+                    container.SchedulerAfterChildren.Add(() => secondExecutedInOrder = firstExecutedInOrder);
+                }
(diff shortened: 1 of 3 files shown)


bdach:
* Can you explain how you ran into this?
* Can this have a test case attached?

frenzibyte:
> Can you explain how you ran into this?

Pretty much the same as I mentioned in the first paragraph minus unnecessary details, basically:
 - Action A was scheduled with `Drawable.Schedule` (which translated to `Scheduler.AddDelayed(..., 0)`).
 - Action B was scheduled with `Scheduler.Add`.

Action B got executed before Action A, when none of them had any delays and Action A should have been executed first.

---

I'm unsure if a test case is required as it would be testing performing a `Drawable.Schedule` followed by `Scheduler.Add` with ensuring execution order to be correct, which feels a bit weird... could add one though, I guess it still makes sense to have one.

bdach:
but what were you *doing* that triggered this? i'm not interested in a more abstract explanation with "action A" and "action B" of what the failure case is. i see what the failure case is. i was interested what *specifically* you were doing where this surfaced, because apparently this order matters in some respect in whatever it is that you're doing.

> I'm unsure if a test case is required as it would be testing performing a `Drawable.Schedule` followed by `Scheduler.Add` with ensuring execution order to be correct, which feels a bit weird...

that's literally what "adding a test case" means? you're arguing that you're not sure if you should add a test case because you'd have to test the behaviour you introduced here? i don't get it.

if this ordering is something we really want to uphold then this is an assumption that should be tested.

frenzibyte:
I didn't think such details will be necessary as I had thought scheduled delegates should always be executed in order... 

...but, on `DrawableChannel`, chat lines are populated on BDL by a method named `newMessagesArrived`, and that method schedules itself using `Drawable.Schedule`. 

I have wrote a method that queries all chat lines to find which chat line to highlight, and I made it so that it gets scheduled if `DrawableChannel` is not loaded yet, so that the chat lines are populated correctly via the initial `newMessagesArrived` call.

I have used `Scheduler.Add<T>(action, data)` to schedule the highlighting method, but then I have realized the highlighting method is actually being executed before the scheduled intial `newMessagesArrived` logic, which is contrary to what I had expected (since `newMessagesArrived` ran first).

After debugging through, I've found out `Drawable.Schedule` does not execute before a regular `Scheduler.Add<T>(action, data)` due to using `AddDelayed` regardless of whether `TransformDelay = 0`, and that's why I ended up PR'ing this.

bdach:
i was asking because while i agree that this ordering is probably more correct, relying on it is probably not the wisest. and i'd say that in your use case relying on scheduling order seems kind of sketchy. i'd ideally like a solution that does not rely on something like that as it's potentially really easy to break - and this change could end up being potentially breaking (although i'd hope we don't rely on the old ordering anywhere either).

frenzibyte:
I'm not 100% sure on that since the initial `newMessagesArrived` call happens on BDL, which should be invoked before any side methods like highlighting a message get called.

frenzibyte:
I have added test coverage as requested.

smoogipoo (approved):
I think this probably makes sense in isolation. Bit of a shame that there's no linking between timed tasks and others - intuitively I imagine something like an incoming queue of delegates waiting to be assigned, however I'm not sure if that would work with the way `Scheduler` is written.

peppy (approved):

peppy:
I'm fine with this, until we come across a case that requires things to behave differently (hopefully never).

PR #5062, opened 2022-03-07
Abort test execution on unobserved exception

peppy:
The aim here is to associate unobserved exceptions with the test that fires them.

Examples of failures we cannot currently see or locate:

- https://github.com/ppy/osu/runs/5440702517?check_suite_focus=true
- https://github.com/ppy/osu/runs/5440702526?check_suite_focus=true

Of note:

5f648de7c: Fixes logging test potentially leaving an unobserved exception in the mix. Without this other tests will randomly fail due to the per-test fire logic only being run for `TestScene`s, which the logging tests are not.

84e62c3a1: Exposes the inner exception when logging / reporting unobserved exceptions. The `UnobservedException.Exception` itself is always non-useful as it is an aggregate. I'm not 100% sure on whether we want to expose the aggregate or call `AsSingular` as I am.

2ff818d70: Forces finalizers to run at the end of each test execution to coerce unobserved exceptions to fire

606c2aa87: Aborts test runs precisely where an unobserved exception is fired, to avoid it firing on a completely unrelated test.


peppy:
Note that I'm seeing `osu.Framework.Tests.Visual.Testing.TestSceneDerivedTestWithDerivedMethods` occasionally fail. Looking into this but PRing in the mean time in case someone catches something obvious.

bdach:
`TestLogging` test failure looks relevant - the `IsNUnitRunning` check added here is presumably messing with the expectations of that test...

Weirdly enough `TestLoadExceptions.TestUnobservedException` is also failing, but that failure I cannot repro locally.

osu.Framework/Testing/TestScene.cs
```diff
@@ -454,6 +454,16 @@ private void checkForErrors()
 
             if (runTask.Exception != null)
                 throw runTask.Exception;
+
+            // Force any unobserved exceptions to fire against the current test run.
+            // Without this they could be delayed until a future test scene is running, making tracking down the cause difficult.
+            collectAndFireUnobserved();
```
smoogipoo, line 460: Just note that even if this exists it's still potentially going to fire on a separate test.

peppy: What's the case for that?

smoogipoo: If the task's still ongoing by the time this is called, which is completely possible - e.g. a background web request.

peppy: Aha. I don’t think that’s the case we’re dealing with in osu! side failures touch wood.

peppy:
I fixed local tests but knew there were likely more CI only failures. Will check again today.

peppy:
So there are two remaining failures. They both seem quite important and not failures that we should expect nor ignore.

`NUnit.Framework.AssertionException:   Game execution was not aborted` (https://github.com/ppy/osu-framework/runs/5975802274?check_suite_focus=true / https://github.com/ppy/osu-framework/runs/5975802446?check_suite_focus=true / https://github.com/ppy/osu-framework/runs/5975802618?check_suite_focus=true)

Failing on all platforms, and suggesting that an unobserved exception is not properly exiting the game loop. It can be seen from the logs that the abort path *was* triggered, though (so it's not that the unobserved exception was missed / arrived too late). Is the abort sequence happening too early?

`Debug.Assert(Handle != 0)` when adding channel to bass mixed (https://github.com/ppy/osu-framework/runs/5975802532?check_suite_focus=true)

Potentially macOS only. Is this something that we have to live with, due to intermittent audio failures we're already become accustomed to? Not sure, @smoogipoo might know better.

peppy:
Looking into the mixer assertion specifically, I'm going to require some explanation as to the intention of it. It's asserting in the `public` add method, but null-check (and silently returning) in the *proteted* `AddInternal` method. This seems potentially back-to-front?

It looks like it's accepted that the mixer handle creation may fail and remain zero. If that's the case, we should either support this in the public method (`bool` return as to whether the sample could be added, or silently failing) or be more noisy about the mixer creation failing in the first place (and potentially figuring out why/when that can happen).

smoogipoo:
The reason why those null checks exist (in `AddInternal`) is because there's samples initialised at game start that are added to mixers before the `AudioManager` ever initialises the first device. I believe the same could occur normally in the hierarchy (e.g. when using `DrawableAudioMixer` + `DrawableSample`) depending on whether the audio thread gets a chance to run before the sample is played.

The asserts, on the other hand, are in places where the mixer + channel should always have been initialised.

I've quickly verified that the only relevant pathway here into `AddChannelToBassMix()` is the `TrackBass` one that is failing, I'll leave it as a reader's exercise that the other pathways into the method would adhere to the assertion conditions.

Now, when an `IAudioChannel` (`TrackBass`/`SampleChannelBass`) has its `Mixer` value set, it's done so _enqueued to the channel itself`:
image
This is to prevent the channels from being added to mixers without having been initialised first (which happens in the constructor of each one). So you get the ordering:
```
1. Initialise channel.
2. Add channel to mixer (handle presumed to be non-null).
3. ChannelPlay()
```

I suspect that the issue here is channel initialisation failing.

smoogipoo:
If you're able to repro the issue locally, does changing `AddChannelToBassMix()` to use `EnqueueAction()` work? I would need to think more about the ordering of things because it gets a little weird - you'd have to also enqueue `removeChannelFromBassMix()` but then you also need to consider the ordering between audio mixers in the hierarchy.

We probably need to do the whole queueing of actions to a central source thing we've discussed soon to avoid these issues, just to have a single topological order of actions.

peppy:
I can't repro locally, unfortunately.

peppy:
I figured out the issue with the unobserved test failures (was silly). Assuming the audio mixer issue doesn't arise over a few tests, I'd like to get this in the next framework release.

osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -378,21 +378,16 @@ private void initializeSyncs()
...
         {
-            Debug.Assert(stopCallback != null
-                         && stopSync != null
-                         && endCallback != null
-                         && endSync != null);
-
-            bassMixer.ChannelRemoveSync(this, stopSync.Value);
-            bassMixer.ChannelRemoveSync(this, endSync.Value);
+            if (stopSync != null) bassMixer.ChannelRemoveSync(this, stopSync.Value);
+            if (endSync != null) bassMixer.ChannelRemoveSync(this, endSync.Value);
```
smoogipoo, line 382: Are these changes still required? I'd hope for them to not be.

peppy: I'll try removing them.

osu.Framework.Tests/IO/TestLogging.cs
```diff
@@ -130,13 +130,16 @@ protected override void Update()
         }
 
         [Test]
-        public void TestGameUnobservedExceptionDoesntCrashGame()
+        public void TestGameUnobservedThrowsUnderNUnit()
```
peppy, line 133: This test is still intermittent (obviously, due to `TaskCrashTestGame`'s implementation). I think we should just remove the test altogether? It's basically a duplicate now that the assertion is reversed from original...

osu.Framework/Audio/Track/TrackBass.cs
peppy: Seems to still occur.. https://github.com/peppy/osu-framework/runs/6035783243?check_suite_focus=true

smoogipoo: Concerning, also looks like the assertion in AddChannelToBassMix still happens too.

peppy: And the new test still fails in weird ways too.. https://github.com/ppy/osu-framework/runs/6034748640?check_suite_focus=true

This is cursed.

peppy:
I’ve fixed the main failures in the new test. I think we should decide on an amicable if temporary solution to the remaining mixer issue because the benefits of getting this merged are potentially larger (fixing remaining osu! side test failures).

smoogipoo (approved):
I still feel very uneasy about this, but I suppose we can give it a shot. Shouldn't have an effect outside of tests at worst. Don't be surprised if I suddenly put a few days' work into mixers though! :wink:

PR #5065, opened 2022-03-10
Rework CircularProgress to use shader

EVAST9919:
To make things easy to review antialiasing will come in a separate pr.
Also adds toggle for rounded caps.

- [x] Needs fix for mobile platforms


frenzibyte (commented):
Left a few nits, I'm not sure about the trigonometry behind this as I'm not knowledgable on that, but overall this looks pretty good. Nice work.

EVAST9919:
Applied suggestions and added rounded version

osu.Framework/Resources/Shaders/sh_CircularProgress.fs
```diff
@@ -0,0 +1,50 @@
+﻿#include "sh_Utils.h"
+#include "sh_TextureWrapping.h"
+
+#define PI 3.1415926538
```
bdach, line 4: Interesting value of pi there, should be

```suggestion
#define PI 3.1415926536
```

going off of the math books that I have. (Incidentally the first result of googling "glsl pi" contains this same wrong value.)

Not that this is going to ever realistically matter but let's attempt to get the basic math constants right.

osu.Framework/Resources/Shaders/sh_CircularProgress.fs
```diff
@@ -0,0 +1,50 @@
...
+uniform lowp sampler2D m_Sampler;
+uniform mediump float progress;
+uniform mediump float innerRadius;
+
+bool insideProgressSector(vec2 pixelPos)
+{
+    if (progress >= 1.0)
+        return true;
+
+    float angle = atan(0.5 - pixelPos.y, 0.5 - pixelPos.x) - PI / 2.0;
```
bdach, line 19: use swizzling

```suggestion
    float angle = atan(0.5 - pixelPos.yx) - PI / 2.0;
```

osu.Framework/Resources/Shaders/sh_CircularProgress.fs
```diff
@@ -0,0 +1,50 @@
...
+
+void main(void)
+{
+    if (progress == 0.0 || innerRadius == 0.0)
+    {
+        gl_FragColor = vec4(0.0);
+        return;
+    }
+
+    vec2 resolution = vec2(v_TexRect[2] - v_TexRect[0], v_TexRect[3] - v_TexRect[1]);
```
bdach, line 46: use swizzling

```suggestion
    vec2 resolution = v_TexRect.zw - v_TexRect.xy;
```

osu.Framework.Tests/Visual/UserInterface/TestSceneCircularProgress.cs
```diff
@@ -71,16 +73,23 @@ public TestSceneCircularProgress()
...
                     Anchor = Anchor.Centre,
                     Origin = Anchor.Centre,
-                },
+                    Size = new Vector2(250),
+                    CornerRadius = 20,
+                    Child = clock = new CircularProgress
+                    {
+                        Anchor = Anchor.Centre,
+                        Origin = Anchor.Centre,
+                        Size = new Vector2(350)
```
bdach, line 90: Kind of weird to always force 1:1 aspect ratio in this test when it wasn't enforced previously. Would have liked to keep a test or even a step showing that this also works for arbitrary aspect ratios.

osu.Framework/Resources/Shaders/sh_CircularProgress.fs
```diff
@@ -0,0 +1,50 @@
...
+
+    if (angle < 0.0)
+        angle += 2.0 * PI;
+
+    return angle < 2.0 * PI * progress;
+}
+
+bool insideProgress(vec2 pixelPos)
+{
+    float innerBorder = 0.5 - (0.5 * innerRadius);
```
bdach, line 29: weird way to write this, would prefer

```suggestion
    float innerBorder = 0.5 * (1 - innerRadius);
```

osu.Framework/Resources/Shaders/sh_CircularProgress.fs
```diff
@@ -0,0 +1,50 @@
...
+    float outerBorder = 0.5;
+
+    float dstFromCentre = distance(pixelPos, vec2(0.5));
+    bool insideRing = dstFromCentre > innerBorder && dstFromCentre < outerBorder;
+
+    return insideRing && insideProgressSector(pixelPos);
+}
```
frenzibyte, line 36: Can have this logic moved to a header file and included in both `sh_CircularProgress` and `sh_CircularProgressRounded` to avoid code duplication (has been done before with other shaders like `sh_Video`, see `sh_yuv2rgb.h`).

bdach (changes requested):

bdach:
Review comments above apply to both shader files, of course, I just didn't want to double post.

frenzibyte (changes requested):
In addition to the above.

osu.Framework/Resources/Shaders/sh_CircularProgress.fs
EVAST9919: Yeah, just went with what is used elsewhere

osu.Framework/Resources/Shaders/sh_CircularProgress.fs
frenzibyte: Swizzling won't work on `atan` since it always expects two `y` and `x` parameters (crashes with `Incompatible types in initialization (and no available implicit conversion)`)

EVAST9919: Will not work since I'm using `atan(value1, value2)`, not `atan(vec2)`

osu.Framework/Resources/Shaders/sh_CircularProgress.fs
EVAST9919: True, looks much cleaner now

bdach:
Code seems fine but before approval this probably needs testing on iOS and android to make sure this doesn't fall over there due to insufficient `v_TexRect` precision like the hue/saturation shader already does.

frenzibyte:
iOS is broken in the same way as the HSV shaders, which means this will have to stay blocked until that is resolved, otherwise mobile devices will no longer be able to render any `CircularProgress`.

EVAST9919:
I've added precision modifiers since I'm guessing that may be the case , but unfortunately I can't really test this.

osu.Framework/Resources/Shaders/sh_CircularProgress.fs
```diff
@@ -0,0 +1,25 @@
...
+void main(void)
+{
+    if (progress == 0.0 || innerRadius == 0.0)
+    {
+        gl_FragColor = vec4(0.0);
+        return;
+    }
+
+    mediump vec2 resolution = v_TexRect.zw - v_TexRect.xy;
+    highp vec2 pixelPos = v_TexCoord / resolution;
```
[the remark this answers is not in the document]
smoogipoo: The alternative is to have `#if` directives for high precision vertex input.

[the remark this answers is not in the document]
EVAST9919: I assume you are suggesting this?
```
#ifdef GL_ES
    precision highp float;
#endif
```

EVAST9919:
Added potential fix which is a combination of suggestions [here](https://github.com/ppy/osu-framework/pull/5065#discussion_r926182315) and [here](https://github.com/ppy/osu-framework/pull/4702#issuecomment-903304728).
Works fine on Windows but not tested on other platforms (mobile ones especially)

smoogipoo (approved):
Needs testing on mobile but otherwise lgtm

frenzibyte (approved):
Works well, tested on iOS.

PR #5066, opened 2022-03-10
Add a way to confine the mouse cursor to a specific window area

Susko3:
Related: https://github.com/ppy/osu/issues/16956.

[SDL_SetWindowMouseRect](https://wiki.libsdl.org/SDL_SetWindowMouseRect) works quite well on Windows. Using the mouse rect and relative mode at the same time is intentionally avoided in case of SDL2 bugs (there were some weird bugs with it when I checked some time ago).

To verify this works on macOS and Linux, I suggest the following testing methodology;
- enable `ConfineMouseMode` and `cursor confine rect`
- verify the cursor is confined to the expected area (check the OS cursor with `Toggle cursor visibility`)
- check that toggling relative mode doesn't warp the cursor in any way (keep in mind that relative mode will not work with the OS cursor shown)
- repeat for other window modes
- also test that it works with display scaling != 100%

### DIFF
--- a/osu.Framework/Platform/IWindow.cs
+++ b/osu.Framework/Platform/IWindow.cs
@@ -69,7 +69,10 @@ namespace osu.Framework.Platform
         /// <summary>
         /// Area to which the mouse cursor is confined to when <see cref="CursorState"/> is <see cref="Platform.CursorState.Confined"/>.
         /// </summary>
-        /// <remarks>Will confine to the whole window by default (or when set to <c>null</c>).</remarks>
+        /// <remarks>
+        /// Will confine to the whole window by default (or when set to <c>null</c>).
+        /// Supported fully on desktop platforms, and on Android when relative mode is enabled.
+        /// </remarks>
         RectangleF? CursorConfineRect { get; set; }
 
         /// <summary>
--- a/osu.Framework/Platform/IWindow.cs
+++ b/osu.Framework/Platform/IWindow.cs
@@ -7,6 +7,7 @@ using System.Drawing;
 using JetBrains.Annotations;
 using osu.Framework.Bindables;
 using osu.Framework.Configuration;
+using RectangleF = osu.Framework.Graphics.Primitives.RectangleF;
 
 namespace osu.Framework.Platform
 {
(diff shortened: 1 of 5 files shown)


osu.Framework/Platform/IWindow.cs
```diff
@@ -62,8 +63,15 @@ public interface IWindow : IDisposable
         /// <summary>
         /// Controls the state of the OS cursor.
         /// </summary>
+        /// <remarks>If the cursor is <see cref="Platform.CursorState.Confined"/>, <see cref="CursorConfineRect"/> will be used.</remarks>
         CursorState CursorState { get; set; }
 
+        /// <summary>
+        /// Area to which the mouse cursor is confined to when <see cref="CursorState"/> is <see cref="Platform.CursorState.Confined"/>.
```
bdach, line 70: xmldoc should probably mention that this rectangle is in window-relative coordinates (as in, actual absolute device pixels) rather than in any framework coordinate space. And maybe also that this only works on desktop as-is.

Susko3: It's in framework's OS display scaling-scaled coordinate system, since mouse events are [scaled with that](https://github.com/ppy/osu-framework/blob/eab3e5af1172388c585cb6b3c09916365c2ea2a4/osu.Framework/Platform/SDL2DesktopWindow.cs#L921) and the mouse rect is [scaled](https://github.com/ppy/osu-framework/blob/741d648136bad69ac71e530fa84129fc8a69fbbc/osu.Framework/Platform/SDL2DesktopWindow.cs#L262) when passed to SDL, same as the [text input rect](https://github.com/ppy/osu-framework/blob/eab3e5af1172388c585cb6b3c09916365c2ea2a4/osu.Framework/Platform/SDL2DesktopWindow.cs#L627).

Plus it works on Android when relative mode is enabled.

bdach: Point taken on the first - I guess that'd be in... device-independent pixels would be the term?

Second would also be true I suppose, but either way probably best to note when it doesn't work so framework consumers know in advance what to expect.

bdach:
Checked linux, seems to work without much fuss here.

frenzibyte:
macOS also working pretty well, passing the mentioned methodology.

smoogipoo (approved):
ltgm :+1:

peppy (approved):

PR #5068, opened 2022-03-11
Improve the Windows borderless hack

Susko3:
Closes https://github.com/ppy/osu-framework/issues/3456.

Goals for this change
---

1. have the hack not be noticeable by multi-monitor users
    - partially resolved, will now show black instead of bleeding the game over. would be great if the pixels could be transparent
2. render the game at display resolution
3. don't cause off-by-one errors (eg. in mouse input, when translating game coordinates to window/screen space)
4. don't break things reliant on window size, such as screenshots
5. still prevent windows or drivers from putting the game in exclusive fullscreen

Implementation/changes
---

Will now only increase the window width by 1px (as opposed to both width and height, in addition to offsetting the window position). This is done to satisfy 2. and 3. while keeping 5.

Modifying the height (while satisfying 2. and 3.) requires messing with the GL viewport, as the GL origin is the bottom left corner, while windows uses top left. So I decided against that.

Tricking the game into thinking the window is smaller seems to have no ill side effects. The only (minor) problem is the mouse being able to go one pixel too far when it's confined. This can be fixed with `SDL_SetWindowMouseRect`, but since most users will use relative mode (which works fine), I don't think it's worth it.

---

You can test out how the hack looks when the window is even wider, eg. set `WindowsBorderlessWidthHack` to `20` -- the game will neatly render on it's display, while showing a black bar on the display to the right.

On my machines, setting this to `0` will cause the game to enter quasi-exclusive fullscreen, eg. the screen will briefly flash black when alt-tabbing. So expanding only width seems to work just fine for tricking windows and drivers.

Tested on:
- Windows 10 21H1, AMD Radeon R7 370 (desktop)
- Windows 10 21H2, AMD Ryzen 7 5700U with Radeon graphics (laptop)

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -14,6 +14,7 @@ using osu.Framework.Extensions.ImageExtensions;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Input;
 using osu.Framework.Platform.SDL2;
+using osu.Framework.Platform.Windows;
 using osu.Framework.Platform.Windows.Native;
 using osu.Framework.Threading;
 using osuTK;
@@ -147,12 +148,25 @@ namespace osu.Framework.Platform
             get => size;
             private set
             {
+                // trick the game into thinking the borderless window has normal size so that it doesn't render into the extra space.
+                value.Width -= WindowsBorderlessWidthHack;
+
                 if (value.Equals(size)) return;
 
                 size = value;
                 Resized?.Invoke();
             }
         }
+
+        /// <summary>
+        /// Amount of extra width added to window size when in borderless mode on Windows.
+        /// Some drivers required this to avoid the window switching to exclusive fullscreen automatically.
+        /// </summary>
+        /// <remarks>
+        /// Either <c>0</c> or <c>1</c>, depending on if the Windows borderless hack is active.
+        /// See <see cref="WindowsWindow.SetBorderless"/>.
+        /// </remarks>
+        protected int WindowsBorderlessWidthHack;
 
         /// <summary>
         /// Provides a bindable that controls the window's <see cref="CursorStateBindable"/>.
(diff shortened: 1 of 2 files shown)


osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -147,13 +148,26 @@ public Size Size
             get => size;
             private set
             {
+                // trick the game into thinking the borderless window has normal size so that it doesn't render into the extra space.
+                value.Width -= WindowsBorderlessWidthHack;
```
Susko3, line 152: This is done here (instead of in `WindowsWindow`) in case some SDL window event causes the size to be recalculated.

frenzibyte: You can still keep this in `WindowsWindow` via overrides, no? `Size` can be virtualised, and `updateWindowStateAndSize` can be exposed as protected and virtualised as well. Don't think there's too much of an issue with that.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -147,13 +148,26 @@ public Size Size
...
 
+        /// <summary>
+        /// Amount of extra width added to window size when in borderless mode on Windows.
+        /// Some drivers required this to avoid the window switching to exclusive fullscreen automatically.
+        /// </summary>
+        /// <remarks>
+        /// Either <c>0</c> or <c>1</c>, depending on if the Windows borderless hack is active.
+        /// See <see cref="WindowsWindow.SetBorderless"/>.
+        /// </remarks>
+        protected int WindowsBorderlessWidthHack;
```
frenzibyte, line 169: I find the existence of this property unnecessary. The value itself can be moved to a constant and the application in `Size` can be conditioned based on whether `WindowState == WindowState.FullscreenBorderless`. 

That will make moving all this logic to `WindowsWindow` be done with only virtualising `Size`'s setter instead of having to expose `updateWindowStateAndSize`, and will read better in general.

frenzibyte (changes requested):
Seems fine in general. Haven't tested this, only read through code.

frenzibyte (approved):

peppy:
Interestingly I was recently working around similar code on stable to fix a long-time bug. I have a good set of manual tests to go through to make sure things are working as intended, so will do that today.

peppy (approved):
Works fine with multi-monitor multi-refresh

PR #5069, opened 2022-03-12
Fix `ClearTransforms` potentially discarding container autosizing

### LINKED ISSUE #5064: `ClearTransforms` can discard autosizing indefinitely
Performing `ClearTransforms` clears autosizing transform, and therefore never make the container autosize until `childrenSizeDependencies` becomes invalidated again by updating any size-relevant properties.

One way I can see to fix this would be to invalidate `childrenSizeDependencies` on `ClearTransforms` so that the autosize transform resumes next frame if nothing else has invalidated that layout member.

Failing test case to demonstrate the issue at hand:
```csharp
        // in TestSceneCompositeDrawable
        [Test]
        public void TestClearTransformsOnDelayedAutoSize()
        {
            Container container = null;

            AddStep("create hierarchy", () =>
            {
                Child = container = new Container
                {
                    AutoSizeAxes = Axes.Both,
                    AutoSizeDuration = 1000,
                    Child = new Box { Size = new Vector2(100) },
                };

                container.OnLoadComplete += __ =>
                {
                    // query size to validate children size dependencies and trigger autosize transforms...
                    _ = container.Size;
                    // ...then clear transforms to interrupt autosize transform.
                    container.ClearTransforms();
                };
            });

            // expectation: container eventually autosizes to 100 size.
            // actual: container never autosizes and remains at 0 size.
            AddUntilStep("container autosized", () => container.Size == new Vector2(100));
        }
```

frenzibyte:
Closes #5064 

A similar case can be seen with `FlowContainer`'s layout transform, but that's more involved as the transforms are applied to the children, which makes catching them in `FlowContainer.ClearTransformsAfter` not enough to avoid the issue there (since the transform can still be cleared by the child itself).

### DIFF
--- a/osu.Framework.Tests/Visual/Containers/TestSceneCompositeDrawable.cs
+++ b/osu.Framework.Tests/Visual/Containers/TestSceneCompositeDrawable.cs
@@ -52,6 +52,41 @@ namespace osu.Framework.Tests.Visual.Containers
             AddStep("bad change child depth call fails", () => Assert.Throws<InvalidOperationException>(() => parent.ChangeChildDepth(nestedChild, 10)));
         }
 
+        /// <summary>
+        /// Ensures that <see cref="CompositeDrawable.ClearTransforms"/> does not discard auto-sizing indefinitely.
+        /// </summary>
+        [Test]
+        public void TestClearTransformsOnDelayedAutoSize()
+        {
+            Container container = null;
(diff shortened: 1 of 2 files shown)


peppy (approved):
LGTM

smoogipoo (approved):

smoogipoo (approved):

PR #5072, opened 2022-03-16
Disable HW video decoding if an unexpected error occurs

### LINKED ISSUE #4857: Hardware video acceleration should fallback to a disabled state when not enough memory is available
As reported [by a user](https://github.com/ppy/osu/discussions/15442), there is a potential that video playback no longer works if the user's hardware doesn't have enough memory. It looks to be a shortage of GPU memory in this case, although I can't say I've investigated deep enough to be sure about this.

The log error appearing for this user is as follows:

```
2021-11-02 23:24:49 [verbose]: Couldn't create hardware video decoder context AV_HWDEVICE_TYPE_CUDA for codec h264_cuvid: Unknown error occurred (-1313558101)
2021-11-02 23:24:49 [verbose]: Couldn't create hardware video decoder context AV_HWDEVICE_TYPE_CUDA for codec h264: Unknown error occurred (-1313558101)
2021-11-02 23:24:49 [verbose]: Successfully opened hardware video decoder context AV_HWDEVICE_TYPE_QSV for codec h264_qsv
2021-11-02 23:24:49 [verbose]: Successfully initialized decoder: h264_qsv
2021-11-02 23:24:50 [verbose]: Failed to send avcodec packet: Cannot allocate memory (-12)
```

I'd say in a case where this error is triggered once it should be safe to switch video acceleration off at a framework configuration level automatically, along with a log message saying why this was done.

Opelkuh:
Closes #4857

If any FFmpeg function that works with HW decoding fails with an unexpected error, HW decoding is disabled for the current "decoding session". If the error is `ENOMEM` then HW decoding is also disabled at framework config level.

I only tested this manually with the following patch:
```csharp
diff --git a/osu.Framework/Graphics/Video/VideoDecoder.cs b/osu.Framework/Graphics/Video/VideoDecoder.cs
index 8dd912716..b563d754f 100644
--- a/osu.Framework/Graphics/Video/VideoDecoder.cs
+++ b/osu.Framework/Graphics/Video/VideoDecoder.cs
@@ -534,6 +534,12 @@ private int sendPacket(AVFrame* receiveFrame, AVPacket* packet)
             // send the packet for decoding.
             int sendPacketResult = ffmpeg.avcodec_send_packet(codecContext, packet);
 
+            if (codecContext->hw_device_ctx != null && packet->dts * timeBaseInSeconds > 5)
+            {
+                // sendPacketResult = AGffmpeg.AVERROR_BUG2;
+                sendPacketResult = -AGffmpeg.ENOMEM;
+            }
+
             // Note: EAGAIN can be returned if there's too many pending frames, which we have to read,
             // otherwise we would get stuck in an infinite loop.
             if (sendPacketResult == 0 || sendPacketResult == -AGffmpeg.EAGAIN)
```

### DIFF
--- a/osu.Framework/Graphics/Video/VideoDecoder.cs
+++ b/osu.Framework/Graphics/Video/VideoDecoder.cs
@@ -594,10 +594,7 @@ namespace osu.Framework.Graphics.Video
                     if (transferResult < 0)
                     {
                         Logger.Log($"Failed to transfer frame from HW decoder: {getErrorMessage(transferResult)}");
-
-                        // only try to continue with HW decoder in case the frame was just misconfigured
-                        if (transferResult != -AGffmpeg.EINVAL)
-                            tryDisableHwDecoding(transferResult);
+                        tryDisableHwDecoding(transferResult);
 
                         hwTransferFrame.Dispose();
                         continue;
--- a/osu.Framework/Graphics/Video/VideoDecoder.cs
+++ b/osu.Framework/Graphics/Video/VideoDecoder.cs
@@ -80,6 +80,7 @@ namespace osu.Framework.Graphics.Video
 
         private bool inputOpened;
         private bool isDisposed;
+        private bool hwDecodingAllowed = true;
         private Stream videoStream;
 
         private double timeBaseInSeconds;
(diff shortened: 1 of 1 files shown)


bdach (commented):
seems mostly sane

bdach (approved):

smoogipoo (approved):

PR #5074, opened 2022-03-20
Extend Draw Visualiser to support inspecting objects and track multiple views.

dekrain:
There are still some things left for consideration, like the ability to resize `ToolWindow`, as type names can get overwhelmingly long to be shown. Alternatively we could limit the maximum label size and add a tooltip instead.

There are also performance aspects to consider.

Preview:
preview

Changes made for `osu.Framework/Graphics/Visualisation/VisualisedDrawable.cs`
- Rename to `ElementNode` (previously `VisualisedElement`), add support for object visualisation
- Visualise lists
  - Support `Frozen`, `Live (refreshable)` and `Live (observed)` collections
  - Refreshing work in progress
- Visualise bindables
  - Show live view of the value
  - Show weak references to bound bindables
- Visualise weak references
- Display groups of elements
- Iterate and modify tree nodes
- `ObjectNode` color code:
  - Ordinary `object`s - blue
  - Ordinary `ValueType`s - dark violet
  - `Drawable`s in object view - white
  - `null` - gray
  - Lists/enumerables - orange red
  - `Bindable`s - lime green
  - `WeakReference`s - brown

Changes made for `osu.Framework/Graphics/Visualisation/PropertyDisplay.cs`
- Make it work for objects (transparently through `ElementNode`)
- Display target, base & interface types
- Allow grouping by type
- Allow resetting the view
- Disable showing properties by default
- Format different kinds of values
- Allow inspecting values

The whole `DrawVisualiser` and `TreeContainer` framework has been redesigned completely to allow for
displaying multiple targets in a single window.


bdach:
Sorry but I am not reviewing a pull request with 3.5k lines changed. A decent size to aim for for a single PR would be around a tenth of that at a time.

Issue #5073, opened 2022-03-18
`PresentFileExternally()` is unsafe to call on mobile platforms due to no implementation


bdach:
Stemmed from https://github.com/ppy/osu/discussions/17333.

Many paths in game indiscriminately call `PresentFileExternally()` and eat exceptions due to lack of implementations on mobile. These endpoints either need to be implemented properly at some point using the correct mobile APIs if ones are available, or the API of `PresentFileExternally()` should be changed to handle potential failure to open better (via `bool` return value perhaps?)

frenzibyte:
I believe making `PresentFileExternally()` returning success value rather than throwing exception sounds always better, similar to how we have handled image `Clipboard`s on non-Windows platforms.

peppy:
Can agree with `bool` return as we have done elsewhere.

PR #5075, opened 2022-03-21
Change signature of `OpenFileExternally` and `PresentFileExternally` to gracefuly fail

peppy:
Mobile platforms were throwing `NotImplementedException` for these methods, which is very easy to overlook. Providing a return code makes it very clear that it is not guaranteed that these method will result in success.

Closes https://github.com/ppy/osu-framework/issues/5073.

### DIFF
--- a/osu.Framework/Platform/Storage.cs
+++ b/osu.Framework/Platform/Storage.cs
@@ -117,7 +117,7 @@ namespace osu.Framework.Platform
         /// <summary>
         /// Opens a native file browser window to the root path of this storage.
         /// </summary>
-        /// <returns>Whether the file was successfully presented.</returns>
+        /// <returns>Whether the storage was successfully presented.</returns>
         public bool PresentExternally() => OpenFileExternally(string.Empty);
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


osu.Framework/Platform/Storage.cs
```diff
@@ -111,12 +111,14 @@ public virtual Storage GetStorageForDirectory(string path)
...
         /// <param name="filename">The relative path to the file which should be opened.</param>
-        public abstract void OpenFileExternally(string filename);
+        /// <returns>Whether the file was successfully opened.</returns>
+        public abstract bool OpenFileExternally(string filename);
 
         /// <summary>
         /// Opens a native file browser window to the root path of this storage.
         /// </summary>
-        public void PresentExternally() => OpenFileExternally(string.Empty);
+        /// <returns>Whether the file was successfully presented.</returns>
```
frenzibyte, line 120: `PresentExternally` presents the storage folder. Unsure if it should be called `storage` or `folder` or `path`.
```suggestion
        /// <returns>Whether the storage was successfully presented.</returns>
```

frenzibyte (approved):

PR #5076, opened 2022-03-21
Apply new resharper non-obvious precedence inspections

peppy:
Should hopefully be non-controversial.


frenzibyte (approved):
I personally think the conditionals read much better with parantheses surrounding them (seeing `A && B || C` always makes me feel like it's `A && (B || C)`).

Interested to hear the other's opinions.

bdach (approved):
no objection from me for sure

smoogipoo (approved):

PR #5077, opened 2022-03-22
Prevent FillFlowContainer elements jumping rows with Spacing transform

mcpower:
If `Spacing` was transformed with a `Direction` of `FillDirection.Full`, some elements might "jump" between rows (see ppy/osu#9155) during the transform.

This commit introduces an "intended horizontal spacing" to `FillFlowContainer`, which is used instead of `Spacing.X` when to place
elements on a new row. Specifying this before a `Spacing` transform will prevent items from jumping between rows during the transform.

This PR fixes (but shouldn't close) ppy/osu#9155 after adding `IntendedXSpacing = 50f` [in the `FillFlowContainer` constructor](https://github.com/ppy/osu/blob/6e55d7df02f4e7afadc51ea53c4872fd16724777/osu.Game/Overlays/Mods/ModSection.cs#L234) of `ModSection`. A corresponding osu! PR would need to be created once this gets downstreamed to osu!.

### DIFF
--- a/osu.Framework/Graphics/Containers/FillFlowContainer.cs
+++ b/osu.Framework/Graphics/Containers/FillFlowContainer.cs
@@ -86,6 +86,29 @@ namespace osu.Framework.Graphics.Containers
             }
         }
 
+        private float? intendedXSpacing = null;
+
+        /// <summary>
+        /// The intended final horizontal spacing between individual elements, used to calculate how
+        /// many elements are in a given row when <see cref="Direction"/> is set to
+        /// <see cref="FillDirection.Full"/>. This is intended to be used when a
+        /// <see cref="Spacing"/> transform would cause elements to jump between rows during the
+        /// transform, such as in https://github.com/ppy/osu/issues/9155.
+        /// If null or not set, <see cref="Spacing"/>'s X component is used.
+        /// </summary>
+        public float? IntendedXSpacing
+        {
+            get => intendedXSpacing;
+            set
+            {
+                if (intendedXSpacing == value)
+                    return;
+
+                intendedXSpacing = value;
+                InvalidateLayout();
+            }
+        }
+
         private Vector2 spacingFactor(Drawable c)
         {
             Vector2 result = c.RelativeOriginPosition;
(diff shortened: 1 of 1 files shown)


bdach:
Several concerns:

* Immediate concern is that for this to even be considered, there should be a visual test case in framework showing how this works. There is none right now.
* More overarching concern is that this solution is very ad hoc, targeted at this one specific failure case, and it does not make sense to me to address this specific type of failure in framework of all places. Framework properties are supposed to be ubiquitous, and the fact that elements jump rows generally *is correct*, it's just that it looks bad in this one particular case. As such I'd much rather see a local fix to the mod overlay game-side, than this hyperspecific property in an otherwise generlc container.
* Finally, this problem will likely disappear on its own because the current mods overlay design [is set to be replaced](https://github.com/ppy/osu/issues/16917) with another one that is not going to exhibit the same issue.

bdach (changes requested):
see above

mcpower:
Ack. Thank you so much for the context - it is much appreciated.

> Immediate concern is that for this to even be considered, there should be a visual test case in framework showing how this works. There is none right now.

I was wondering about this. As there doesn't seem to be any specific visual tests for FillFlowContainer, I assumed that it was not meant to be tested - but I guess my assumption was incorrect.

> More overarching concern is that this solution is very ad hoc, targeted at this one specific failure case, and it does not make sense to me to address this specific type of failure in framework of all places. [...]

I agree. The specific bug seems like a hard problem to solve, and this PR takes the simplest - albeit very ad hoc - way of solving it. There's a few ways of fixing it from what I understand:

- This PR fixes it with code in osu!framework to deal with that specific case. The corresponding osu! change is trivial.
- Alternatively, something more general (such as a "columns per row with `FillDirection.Full`") might work for more than just `Spacing` transforms, but requires non-trivial code in osu! to determine the number of columns given a container size.
- Fixing it without any additional osu!framework code seems non-trivial without replacing the whole design,

> Framework properties are supposed to be ubiquitous, and the fact that elements jump rows generally *is correct*, it's just that it looks bad in this one particular case.

I agree that "jumps between rows during transition" is WAI. One possible mitigation to make it a bit nicer-looking is to transition the elements' position when the "jump" occurs, but that is presumably very difficult to do while looking good.

> Finally, this problem will likely disappear on its own because the current mods overlay design [is set to be replaced](https://github.com/ppy/osu/issues/16917) with another one that is not going to exhibit the same issue.

Awesome - that means this change - and the aforementioned bug! - will be obsolete soon enough.

***

I think overall, this PR should be closed. The aforementioned bug will be obsolete soon enough, and the current approach of fixing this issue definitely doesn't scale (especially with #4905 being in-flight). Thank you again!

PR #5078, opened 2022-03-24
Fix zero-size sprites being drawn

smoogipoo:
The real issue I wanted to address here is the `InflationAmount / DrawRectangle` divisions leading to `NaN` values, but I think it makes more sense to simply not draw sprites with zero sizes than inlining conditions.

### DIFF
--- a/osu.Framework/Graphics/Shapes/Triangle.cs
+++ b/osu.Framework/Graphics/Shapes/Triangle.cs
@@ -44,12 +44,18 @@ namespace osu.Framework.Graphics.Shapes
 
             protected override void Blit(Action<TexturedVertex2D> vertexAction)
             {
+                if (DrawRectangle.Width == 0 || DrawRectangle.Height == 0)
+                    return;
+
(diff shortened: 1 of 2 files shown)


peppy:
Is this in response to an issue or call stack?

smoogipoo:
It's a callstack from an internal branch that ensures the last added vertex equals the one given to the method. Which, due to the implementation of `Vector2.Equals()` actually fails because of the `NaN` values...

peppy:
I see. There's a few other custom implementations of `Blit` (`TextureVisualiser` looks like the most likely to be offensive), are these confirmed to not require the same checks?

smoogipoo:
Hmm... I mean it's kind of a much of a muchness, I only did it this way to avoid doing inline ternaries on the inflation percentage, but I can add it to `TextureVisualiser` if you really want it. Or I can change it to use ternaries.

The one in `BlitOpaqueInterior` is also not strictly required for the same reason as above.

peppy:
Let's go with it as-is for now then.

peppy (approved):

PR #5079, opened 2022-03-24
Expose synchronisation context on `GameThread`

peppy:
Allows for a bit of added flexibility. I am able to use this for instance to give `RealmAccess` a way to subscribe for notifications without being required to be called from the update thread. Without exposing this context here, the earliest point of obtaining it is in `Game.LoadComplete` which can be too late (and less guaranteed) for such scenarios.


bdach (approved):
Seems fair to me - if the scheduler is allowed to be public (and there are usages of that game-side already), I don't see why the sync context couldn't be.

smoogipoo (approved):

PR #5081, opened 2022-03-26
Fix ForceLocalVertexBatch not working on flattened composites

smoogipoo:
Only noticed this when debugging a private branch. The `CompositeDrawableDrawNode` will create its own vertex batch, so the composite needs to not be flattened in this case.


peppy:
were there any side effects that could be seen from this? performance or otherwise

smoogipoo:
We don't actually use this, so no.

peppy (approved):

PR #5082, opened 2022-03-26
Fix tracks/samples not updating running state immediately

### LINKED ISSUE #4396: TrackBass and SampleChannelBass don't update running state immediately
When a Play()/Stop() is performed in-line on a `TrackBass` or a `SampleChannelBass`, the respective playing state isn't updated immediately.
```diff
diff --git a/osu.Framework.Tests/Audio/SampleBassTest.cs b/osu.Framework.Tests/Audio/SampleBassTest.cs
index 08d617691..d4379b56e 100644
--- a/osu.Framework.Tests/Audio/SampleBassTest.cs
+++ b/osu.Framework.Tests/Audio/SampleBassTest.cs
@@ -100,6 +100,23 @@ public void TestStopsWhenFactoryDisposed()
             Assert.IsFalse(channel.Playing);
         }
 
+        [Test]
+        public void TestIsRunningUpdatedAfterInlinePlay()
+        {
+            runOnAudioThread(() => channel = sample.Play());
+            Assert.That(channel.Playing, Is.True);
+        }
+
+        [Test]
+        public void TestIsRunningUpdatedAfterInlineStop()
+        {
+            channel = sample.Play();
+            updateSample();
+
+            runOnAudioThread(() => channel.Stop());
+            Assert.That(channel.Playing, Is.False);
+        }
+
         private void updateSample() => runOnAudioThread(() => sampleFactory.Update());
 
         /// <summary>
diff --git a/osu.Framework.Tests/Audio/TrackBassTest.cs b/osu.Framework.Tests/Audio/TrackBassTest.cs
index a83501fe1..afad662c0 100644
--- a/osu.Framework.Tests/Audio/TrackBassTest.cs
+++ b/osu.Framework.Tests/Audio/TrackBassTest.cs
@@ -400,6 +400,23 @@ public void TestCurrentTimeUpdatedAfterInlineSeek()
             Assert.That(track.CurrentTime, Is.EqualTo(20000).Within(100));
         }
 
+        [Test]
+        public void TestIsRunningUpdatedAfterInlinePlay()
+        {
+            runOnAudioThread(() => track.Start());
+            Assert.That(track.IsRunning, Is.True);
+        }
+
+        [Test]
+        public void TestIsRunningUpdatedAfterInlineStop()
+        {
+            track.StartAsync();
+            updateTrack();
+
+            runOnAudioThread(() => track.Stop());
+            Assert.That(track.IsRunning, Is.False);
+        }
+
         private void takeEffectsAndUpdateAfter(int after)
         {
             updateTrack();
```
Note that the `SampleChannelBass.Play()` test already works due to `enqueuedPlaybackStart`, which was a result of another fix for single-threaded (in-line) execution in https://github.com/ppy/osu-framework/pull/4336.

frenzibyte:
- Closes https://github.com/ppy/osu-framework/issues/4396 
- Resolves https://github.com/ppy/osu/discussions/17361

`SampleChannelBass` is refactored to work in a similar manner to `TrackBass`, for ensuring that the zero frequency handler doesn't affect the state of `playing` when manipulating that during a stop call.

It has been mentioned in https://github.com/ppy/osu-framework/pull/4395#issue-865952658 that this is a bit involved, so I'm slightly wondering if I'm missing something with the current solution (cc @smoogipoo).

### DIFF
--- a/osu.Framework.Tests/Audio/SampleBassTest.cs
+++ b/osu.Framework.Tests/Audio/SampleBassTest.cs
@@ -89,6 +89,10 @@ namespace osu.Framework.Tests.Audio
             Assert.IsFalse(channel.Playing);
         }
 
+        /// <summary>
+        /// Tests the case where a play call can be run inline due to already being on the audio thread.
+        /// Because it's immediately executed, a `Bass.Update()` call is not required before the channel's state is updated.
+        /// </summary>
         [Test]
         public void TestPlayingUpdatedAfterInlinePlay()
         {
@@ -96,6 +100,10 @@ namespace osu.Framework.Tests.Audio
             Assert.That(channel.Playing, Is.True);
         }
 
+        /// <summary>
+        /// Tests the case where a stop call can be run inline due to already being on the audio thread.
+        /// Because it's immediately executed, a `Bass.Update()` call is not required before the channel's state is updated.
+        /// </summary>
         [Test]
         public void TestPlayingUpdatedAfterInlineStop()
         {
--- a/osu.Framework.Tests/Audio/SampleBassTest.cs
+++ b/osu.Framework.Tests/Audio/SampleBassTest.cs
@@ -88,5 +88,22 @@ namespace osu.Framework.Tests.Audio
 
             Assert.IsFalse(channel.Playing);
         }
+
+        [Test]
+        public void TestPlayingUpdatedAfterInlinePlay()
+        {
+            bass.RunOnAudioThread(() => channel = sample.Play());
+            Assert.That(channel.Playing, Is.True);
+        }
+
+        [Test]
+        public void TestPlayingUpdatedAfterInlineStop()
+        {
+            channel = sample.Play();
+            bass.Update();
+
+            bass.RunOnAudioThread(() => channel.Stop());
+            Assert.That(channel.Playing, Is.False);
+        }
     }
 }
--- a/osu.Framework/Audio/Sample/SampleChannelBass.cs
+++ b/osu.Framework/Audio/Sample/SampleChannelBass.cs
@@ -36,6 +36,20 @@ namespace osu.Framework.Audio.Sample
         /// </summary>
         private volatile bool enqueuedPlaybackStart;
 
+        public override bool Looping
+        {
+            get => base.Looping;
+            set
+            {
+                base.Looping = value;
+                setLoopFlag(Looping);
+            }
+        }
+
+        private bool hasChannel => channel != 0;
+
+        public override ChannelAmplitudes CurrentAmplitudes => (bassAmplitudeProcessor ??= new BassAmplitudeProcessor(this)).CurrentAmplitudes;
+
         private readonly BassRelativeFrequencyHandler relativeFrequencyHandler;
         private BassAmplitudeProcessor? bassAmplitudeProcessor;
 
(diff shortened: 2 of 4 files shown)


osu.Framework/Audio/Sample/SampleChannelBass.cs
```diff
@@ -36,6 +36,20 @@ internal sealed class SampleChannelBass : SampleChannel, IBassAudioChannel
         /// </summary>
         private volatile bool enqueuedPlaybackStart;
 
+        public override bool Looping
```
bdach, line 39: not sure why methods are being reordered in this fix PR but ok (am complaining because it makes it more difficult to spot and review the actual fix)

osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -213,10 +212,16 @@ public override void Stop()
         public Task StopAsync() => EnqueueAction(() =>
         {
             stopInternal();
-            isPlayed = false;
+            isRunning = isPlayed = false;
         });
 
-        private bool stopInternal() => isRunningState(bassMixer.ChannelIsActive(this)) && bassMixer.ChannelPause(this, true);
+        private void stopInternal()
```
bdach, line 218: what is the point of changing return type from `bool` to `void` here? it shouldn't change behaviour because of how `&&` and short circuiting works, but more importantly - what if the pause fails, for instance?

osu.Framework.Tests/Audio/SampleBassTest.cs
```diff
@@ -88,5 +88,22 @@ public void TestStopsWhenFactoryDisposed()
 
             Assert.IsFalse(channel.Playing);
         }
+
+        [Test]
+        public void TestPlayingUpdatedAfterInlinePlay()
+        {
+            bass.RunOnAudioThread(() => channel = sample.Play());
+            Assert.That(channel.Playing, Is.True);
```
bdach, line 96: isn't this essentially `TestStart()`, but this spins up a thread rather than calls `bass.Update()` explicitly? like i don't follow why this couldn't have been as an assertion in that test

osu.Framework/Audio/Sample/SampleChannelBass.cs
frenzibyte: While it makes it more difficult, I would rather reorder than leave it forgotten as I struggled to read the logic of this at first. But perhaps better doing it in a separate commit rather than embedding it with the fix.

osu.Framework/Audio/Track/TrackBass.cs
frenzibyte: Well, the return value of `stopInternal` has never been used, so felt it's better to just convert it to if conditions to make it read better, and remove the return value while at it.

You do bring a good point however, I thought about not changing the state of `isRunning` if the pause fails, but wasn't too sure about doing it in this PR as it changes behaviour (would need testing, specifically with "no audio" BASS devices to ensure they still work).

osu.Framework/Audio/Sample/SampleChannelBass.cs
bdach: > But perhaps better doing it in a separate commit rather than embedding it with the fix.

that would have been at least some help, yes

bdach:
for what it's worth i don't see any issue with the *actual* primary fix ([f51863b](https://github.com/ppy/osu-framework/pull/5082/commits/f51863b26be49d8ea67eed810e9f9c7b9023cf0b)) but i may be missing something.

osu.Framework.Tests/Audio/SampleBassTest.cs
bdach: right, ok, because of the `IsAudioThread` check in `EnqueueAction()`. may be a good idea to explain this to future readers

osu.Framework/Audio/Sample/SampleChannelBass.cs
peppy: I also don't know how to read this, without ordering back to get a better diff.

peppy (approved):

peppy:
@smoogipoo still requesting your review on this to be triple-sure there's no issue.

This branch may help in reviewing the changes in `SampleChannelBass` (only contains reordering to minimise the diff): https://github.com/ppy/osu-framework/compare/master...peppy:play-stop-update-state-better-diff?expand=1

smoogipoo:
@peppy Thanks, that branch does help a lot. This looks fine to me.

smoogipoo (approved):

PR #5083, opened 2022-03-27
Allow drawables to specify custom drag distance tolerance

ekgame:
Related PR: https://github.com/ppy/osu/pull/17488

This PR implements a dragging tolerance parameter for drawables. This change allows an object to define how sensitive it should be to dragging.

Generally, most draggable elements should have some room for error where the mouse can move just a bit and the action should still be considered a click. However, in some cases it's desirable to handle the drag right away, ex.: for small adjustments in an editor.

### DIFF
--- a/osu.Framework/Graphics/Containers/ScrollContainer.cs
+++ b/osu.Framework/Graphics/Containers/ScrollContainer.cs
@@ -327,7 +327,7 @@ namespace osu.Framework.Graphics.Containers
             // similar calculation to what is already done in MouseButtonEventManager.HandlePositionChange
             // handles the case where a drag was triggered on an axis we are not interested in.
             // can be removed if/when drag events are split out per axis or contain direction information.
-            dragBlocksClick |= Math.Abs(e.MouseDownPosition[ScrollDim] - e.MousePosition[ScrollDim]) > dragButtonManager.ClickDragDistance;
+            dragBlocksClick |= Math.Abs(e.MouseDownPosition[ScrollDim] - e.MousePosition[ScrollDim]) > Content.DragTolerance;
 
             scrollByOffset(scrollOffset, false);
         }
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -2519,6 +2519,11 @@ namespace osu.Framework.Graphics
         public bool IsDragged { get; internal set; }
 
         /// <summary>
+        /// The distance a pointer needs to move to start dragging this drawable.
+        /// </summary>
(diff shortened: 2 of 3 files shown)


osu.Framework/Graphics/Containers/ScrollContainer.cs
```diff
@@ -327,7 +327,7 @@ protected override void OnDrag(DragEvent e)
             // similar calculation to what is already done in MouseButtonEventManager.HandlePositionChange
             // handles the case where a drag was triggered on an axis we are not interested in.
             // can be removed if/when drag events are split out per axis or contain direction information.
-            dragBlocksClick |= Math.Abs(e.MouseDownPosition[ScrollDim] - e.MousePosition[ScrollDim]) > dragButtonManager.ClickDragDistance;
+            dragBlocksClick |= Math.Abs(e.MouseDownPosition[ScrollDim] - e.MousePosition[ScrollDim]) > Content.DragTolerance;
```
bdach, line 330: Both usages of `MouseButtonEventManager.ClickDragDistance` were replaced with `Content.DragTolerance` but `ClickDragDistance` is not removed or marked as obsolete - as it should be if nothing reads it anymore, to avoid confusing framework consumers.

bdach (changes requested):
Not sure we want this. Tests also missing.

osu.Framework/Graphics/Containers/ScrollContainer.cs
ekgame: Good catch, removed.

ekgame:
I'll wait for more opinions before adding tests.

smoogipoo:
If this is to be considered, I would be interested to know how other frameworks (WinForms/WPF/JS) handle this.

The MDN [example](https://jsfiddle.net/radonirinamaminiaina/zfnj5rv4/) for drag has a similar tolerance, so we're not doing something outlandish compared to other frameworks already. Is there a way to customise this behaviour in JS?

jai-x:
I don't believe this is user configurable in Javascript and is defined by the browser.

For reference here are the minimum threshold values for a drag event in Chrome:
https://github.com/chromium/chromium/blob/7bcf5f930c9e6795b8d6539129b01fa5f43c928c/third_party/blink/renderer/core/input/mouse_event_manager.cc#L121-L130

peppy:
I'm open to reducing the leniences to roughly match what chrome uses for testing, but do consider that our use case is slightly different (with tablet users being more dominant in the user base).

I don't think this PR is something we want though.

If others agree, I'm open to reducing the leniences and see if people complain. Maybe half way between what we are using and what chrome has (something like 6 or 7px).

PR #5087, opened 2022-04-01
Bump `SixLabors.ImageSharp` to v2.1.0

frenzibyte:
This is required for supporting macOS image clipboards, requiring TIFF encoder/decoder which is supported in ImageSharp v2.1.0.

Breaking changes come from https://github.com/SixLabors/ImageSharp/pull/1730/https://github.com/SixLabors/ImageSharp/issues/1739.

The important one is that `ArrayPoolMemoryAllocator` has been completely replaced with the new and default [`UniformUnmanagedMemoryPoolMemoryAllocator`](https://github.com/SixLabors/ImageSharp/blob/main/src/ImageSharp/Memory/Allocators/UniformUnmanagedMemoryPoolMemoryAllocator.cs), which is still partially `ArrayPool`-based for smaller allocations, but uses a [different pool implementation](https://github.com/SixLabors/ImageSharp/blob/main/src/ImageSharp/Memory/Allocators/Internals/UniformUnmanagedMemoryPool.cs) for larger allocations.

I can't check on the performance side of this, so would appreciate that being done prior to merge.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -40,7 +40,6 @@ using osu.Framework.Graphics.Textures;
 using osu.Framework.Graphics.Video;
 using osu.Framework.IO.Serialization;
 using osu.Framework.IO.Stores;
-using SixLabors.ImageSharp.Memory;
 using Image = SixLabors.ImageSharp.Image;
 using PixelFormat = osuTK.Graphics.ES30.PixelFormat;
 using Size = System.Drawing.Size;
@@ -114,11 +113,6 @@ namespace osu.Framework.Platform
         /// This and <see cref="SuspendToBackground"/> are an alternative way to exit on hosts that have <see cref="CanExit"/> <c>false</c>.
         /// </remarks>
         public virtual bool CanSuspendToBackground => false;
-
-        /// <summary>
-        /// Whether memory constraints should be considered before performance concerns.
-        /// </summary>
-        protected virtual bool LimitedMemoryEnvironment => false;
 
         protected IpcMessage OnMessageReceived(IpcMessage message) => MessageReceived?.Invoke(message);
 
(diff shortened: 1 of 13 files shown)


osu.Framework/Platform/GameHost.cs
```diff
@@ -667,8 +667,8 @@ public void Run(Game game)
 
             if (LimitedMemoryEnvironment)
             {
-                // recommended middle-ground https://github.com/SixLabors/docs/blob/master/articles/ImageSharp/MemoryManagement.md#working-in-memory-constrained-environments
-                SixLabors.ImageSharp.Configuration.Default.MemoryAllocator = ArrayPoolMemoryAllocator.CreateWithModeratePooling();
+                // recommended middle-ground https://github.com/SixLabors/docs/blob/main/articles/imagesharp/memorymanagement.md#configuring-the-pool-size
+                SixLabors.ImageSharp.Configuration.Default.MemoryAllocator = MemoryAllocator.Create(new MemoryAllocatorOptions { MaximumPoolSizeMegabytes = 1 });
```
frenzibyte, line 671: Maximum pool size number comes [from the previously used method `CreateWithModeratePooling()`](
https://github.com/SixLabors/ImageSharp/blob/458764972654b6beddc6d3f6891bcf1d8c9793a9/src/ImageSharp/Memory/Allocators/ArrayPoolMemoryAllocator.CommonFactoryMethods.cs#L58-L65) (`1024 * 1024`). 

Note that the default memory allocator uses one eighth of the platform's available memory according to `GC.GetMemoryInfo()`, which might consider this logic no longer necessary, unsure.

peppy: After reading their new implementation, I believe we should remove this custom spec completely and see how things perform. The reason we had this was that the pool never cleaned up, but the documentation says it now does.

peppy (approved):

PR #5088, opened 2022-04-01
Add image clipboard support for macOS

frenzibyte:
- [x] Depends on #5087 

The implementation focuses on conversion between `NSImage` and `ImageSharp.Image<TPixel>` using [`TIFFRepresentation`](https://developer.apple.com/documentation/appkit/nsimage/1519841-tiffrepresentation?language=objc) (which requires ImageSharp v2.1.0 for TIFF decoding/encoding support).

This would've been much easier without implementing the interop logic for it. Waiting for the day where we can separate macOS platform to its own project targeting `net6.0-macos`, and not have to do all this interop stuff.

### DIFF
--- a/osu.Framework.Tests/Visual/Platform/TestSceneClipboard.cs
+++ b/osu.Framework.Tests/Visual/Platform/TestSceneClipboard.cs
@@ -29,12 +29,11 @@ namespace osu.Framework.Tests.Visual.Platform
             this.host = host;
         }
 
-        [SetUp]
-        public void SetUp() => Schedule(Clear);
-
         [Test]
         public void TestImage()
         {
+            AddStep("clear previous screenshots", Clear);
+
             AddStep("screenshot screen", () =>
             {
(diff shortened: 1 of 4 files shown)


osu.Framework.Tests/Visual/Platform/TestSceneClipboard.cs
```diff
@@ -26,8 +29,11 @@ private void load(GameHost host)
             this.host = host;
         }
 
+        [SetUp]
+        public void SetUp() => Schedule(Clear);
```
peppy, line 33: This feels very weird, is there a reason it needs to be written this way?

frenzibyte: Well, I'm clearing the test scene on `SetUp` so that a screenshot can be taken without any stale sprite from a last run rendered (in the case of running via visual browser). I'm just gonna move that to a test step.

PR #5094, opened 2022-04-07
Fix issues with `Animation.ClearFrames` behaviour

peppy:
Added test fails at two different places depending on which of the two follow-up commits are built. Felt it was easier to bunch these together due to how intertangled they are.

- efc56b9bb Fixes `Animation` not updating display on adding a single frame after clearing previous frames.
- 0ce46fb60 Fixes previous display not getting cleared on calling `Animation.ClearFrames`

### DIFF
--- a/osu.Framework.Tests/Visual/Sprites/TestSceneAnimation.cs
+++ b/osu.Framework.Tests/Visual/Sprites/TestSceneAnimation.cs
@@ -265,7 +265,7 @@ namespace osu.Framework.Tests.Visual.Sprites
             AddAssert("animation is not showing frame", () => animation.ChildrenOfType<Sprite>().First().Texture == null);
 
             AddStep("add one frame back", () => animation.AddFrame(lastFrame));
-            AddAssert("animation is showing frame", () => animation.ChildrenOfType<Sprite>().First().Texture != null);
+            AddAssert("animation is showing frame", () => animation.ChildrenOfType<Sprite>().First().Texture == lastFrame);
         }
 
         private void loadNewAnimation(bool startFromCurrent = true, Action<TestAnimation> postLoadAction = null)
--- a/osu.Framework.Tests/Visual/Sprites/TestSceneAnimation.cs
+++ b/osu.Framework.Tests/Visual/Sprites/TestSceneAnimation.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Allocation;
 using osu.Framework.Graphics;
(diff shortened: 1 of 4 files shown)


osu.Framework/Graphics/Animations/DrawableAnimation.cs
```diff
@@ -22,6 +22,8 @@ protected override void DisplayFrame(Drawable content)
             container.Child = content;
         }
 
+        protected override void ClearDisplay() => container.Clear(false);
```
bdach, line 25: Now that I read this override, it's occurred to me that `DrawableAnimation` isn't disposing old frame drawables when `ClearFrames()` is called on it, it just drops references and goes on its merry way. As far as I can see this PR doesn't break that, it was already like that, but yeah. Unsure if we want to deal with that and eagerly dispose after all, or just leave it to the finaliser?

osu.Framework.Tests/Visual/Sprites/TestSceneAnimation.cs
```diff
@@ -249,13 +250,22 @@ public void TestGotoFrameBeforeLoaded(int frame)
...
+            AddStep("store current frame", () => lastFrame = animation.CurrentFrame);
+
             AddStep("clear frames", () => animation.ClearFrames());
+
             AddAssert("animation duration is 0", () => animation.Duration == 0);
             AddAssert("animation is at start", () => animation.CurrentFrameIndex == 0);
+            AddAssert("animation is not showing frame", () => animation.ChildrenOfType<Sprite>().First().Texture == null);
+
+            AddStep("add one frame back", () => animation.AddFrame(lastFrame));
+            AddAssert("animation is showing frame", () => animation.ChildrenOfType<Sprite>().First().Texture != null);
```
bdach, line 268: A bit of a weak assertion this, to the point where it doesn't actually fail for me with the fixes reverted. Any reason not to do the following?

```suggestion
            AddAssert("animation is showing frame", () => animation.ChildrenOfType<Sprite>().First().Texture == lastFrame);
```

osu.Framework/Graphics/Animations/DrawableAnimation.cs
peppy: Also noticed this. It makes explicit mention of this behaviour:

https://github.com/ppy/osu-framework/blob/0ce46fb60f3f5847f3d30821c228f193345ba2e4/osu.Framework/Graphics/Animations/DrawableAnimation.cs#L19-L20

So I didn't touch it for now. I think it's fine to leave it to finalizers given that this is a method we use rarely (and probably rely on this behaviour given the comment?) until it causes issue.

smoogipoo (approved):

PR #5095, opened 2022-04-08
Implement screen event args and add destination screen event for `OnExiting`

Joehuu:
As said https://github.com/ppy/osu/issues/13613#issuecomment-869467388, we want to use event args to not break compatibility further.

This was meant to be only for `OnExiting`, but I updated all the screen flows to be consistent. Not sure if we need any other info on the other flows as `Next`/`Last` should be plenty.


osu.Framework.Tests/Visual/UserInterface/TestSceneScreenStack.cs
```diff
@@ -982,6 +985,7 @@ private class TestScreen : Screen
 
             public IScreen EnteredFrom;
             public IScreen ExitedTo;
+            public IScreen DestinationIs;
```
PercyDan54, line 988: This reads weird, I think this should be just `Destination`

peppy:
Before even getting into review: this is a *super* breaking change that will break 100% of framework projects. We either want backwards compatibility / obsolescence or a very good reason not to require that.

Can you start by stating the use case for this change?

bdach:
The use case is the linked game side PR, as far as I can tell.

peppy:
I saw that, but I'd prefer a PR that has this large an effect explains in the description the goals.

osu.Framework/Screens/ScreenResumeEvent.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework.Screens
+{
+    public class ScreenResumeEvent : ScreenEvent
+    {
+        /// <summary>
+        /// The next Screen.
```
peppy, line 9: Is this wrong?

osu.Framework/Screens/ScreenSuspendEvent.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework.Screens
+{
+    public class ScreenSuspendEvent : ScreenEvent
+    {
+        /// <summary>
+        /// The new Screen.
```
peppy, line 9: All of these xmldoc seem randomly chosen and don't make much sense.

osu.Framework/Screens/ScreenEvent.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework.Screens
+{
+    /// <summary>
+    /// Denotes a screen event.
+    /// </summary>
+    public class ScreenEvent
```
peppy, line 9: Is there any reason this can't have `Last` and `Next` and act as the event which is used for all of the flows? I'm not sure we need to complicate things but having different subclassed types when they all have the same available properties.

peppy (changes requested):

osu.Framework/Screens/ScreenEvent.cs
Joehuu: I don't think that will work out. What will `Next` be on `OnEntering` / `OnResuming` and `Last` be on `OnExiting`/`OnSuspending`? They will also have different xmldoc which I will improve.

`OnExiting` is also a special case where it is called multiple times (e.g. Editor -> Song Select -> Main Menu) then `Game` window exiting, which is why in this PR, I added the `Destination` property which cannot be used with the other flows.

osu.Framework/Screens/ScreenResumeEvent.cs
Joehuu: I just copied the existing xmldoc, will improve.

osu.Framework/Screens/ScreenExitEvent.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public class ScreenExitEvent : ScreenEvent
+    {
+        /// <summary>
+        /// The <see cref="IScreen"/> that will be resumed next.
+        /// </summary>
+        public IScreen Next;
+
+        /// <summary>
+        /// The final <see cref="IScreen"/> of an exit operation.
```
frenzibyte, line 14:
```suggestion
        /// The final <see cref="IScreen"/> of this exit operation.
```
I think `this` reads better? Unsure, reading this at first glance made me question the difference between `Next` and `Destination`.

Maybe should explicitly clarify in `Destination` that it's different from `Next` when this is a multiple exiting operations (i.e. from `MakeCurrent`).

osu.Framework/Screens/Screen.cs
```diff
@@ -32,17 +32,17 @@ internal override void UpdateClock(IFrameBasedClock clock)
...
 
-        public virtual void OnResuming(IScreen last)
+        public virtual void OnResuming(ScreenResumeEvent e)
         {
         }
 
-        public virtual void OnSuspending(IScreen next)
+        public virtual void OnSuspending(ScreenSuspendEvent e)
         {
         }
```
frenzibyte, line 47: In regards to obsoletion, the old method signatures can still remain but obsoleted, with the new event-based variants still redirecting to them if not overriden.
```csharp
#pragma warning disable CS0618
        public virtual void OnEntering(ScreenEnterEvent e) => OnEntering(e.Last);

        public virtual bool OnExiting(ScreenExitEvent e) => OnExiting(e.Next);

        public virtual void OnResuming(ScreenResumeEvent e) => OnResuming(e.Last);

        public virtual void OnSuspending(ScreenSuspendEvent e) => OnSuspending(e.Next);
#pragma warning restore CS0618

        [Obsolete("Override OnEntering(ScreenEnterEvent) instead.")] // Can be removed 20221013
        public virtual void OnEntering(IScreen last)
        {
        }

        [Obsolete("Override OnExiting(ScreenExitEvent) instead.")] // Can be removed 20221013
        public virtual bool OnExiting(IScreen next) => false;

        [Obsolete("Override OnResuming(ScreenResumeEvent) instead.")] // Can be removed 20221013
        public virtual void OnResuming(IScreen last)
        {
        }

        [Obsolete("Override OnSuspending(ScreenSuspendEvent) instead.")] // Can be removed 20221013
        public virtual void OnSuspending(IScreen next)
        {
        }
```

This will cover 99% of all use cases, with the only remaining exceptions being the special logic in `OnlinePlayScreen` which forwards screen event calls to its own screen stack. I think it should be fine breaking just that specific case, rather than adding this obsoletion logic in `IScreen` as well.

frenzibyte (commented):
> Can you start by stating the use case for this change?

@peppy I think the use case, as I understand it, can be shortened to the following: 

A screen wants to interrupt a multiple exiting operation (caused by `MakeCurrent` on a far behind screen) to display a "confirm exit" dialog. The blocking screen is unaware of the destination screen, therefore it can't continue the exiting chain properly without exposing the destination screen as done in this PR.

This seems fine to me as long as obsoletions are in place and xmldocs are clear enough (right now `Next` and `Destination` feel slightly confusing together).

osu.Framework/Screens/Screen.cs
peppy: Looks fine, feel free to commit this to the PR.

osu.Framework/Screens/ScreenEvent.cs
peppy: > I don't think that will work out. What will `Next` be on `OnEntering` / `OnResuming` and `Last` be on `OnExiting`/`OnSuspending`? They will also have different xmldoc which I will improve.

The answers to these should be obvious. I'm not sure how to answer. `OnEntering` is *entering* a screen. That screen that is entering is the next screen displayed.

When exiting, the last screen is exiting. So it should be `Last`.

If this doesn't make sense it's going to be faster for me to apply code changes, which I can do if you'd prefer.

I'd also rename this event to `ScreenTransitionEvent` to match the actual usage.

> `OnExiting` is also a special case where it can be called multiple times (e.g. Editor -> Song Select -> Main Menu) then `Game` window exiting, which is why in this PR, I added the `Destination` property to let the exit dialog know the final screen, which cannot be used with the other flows.

That's fine. You can make a separate event type for exit specifically.

peppy (changes requested):

osu.Framework/Screens/ScreenEvent.cs
peppy: @smoogipoo at a conceptual level can you confirm you agree with this? ie. all these events should use a shared

```chspar
event ScreenTransitionEvent
{
    IScreen Last;
    IScreen Next;
}```

Joehuu: So it's just equivalent to `this`, right? In that comment, I wanted to imply that adding those are redundant but wasn't clear enough.

You can apply the changes yourself if you think that structure is okay.

peppy: The main goal is that it greatly simplifies explaining the last and next properties, or from the flip-side, understanding them during implementation. Dealing with the same event for each of the overridden methods rather than having to double-check each time is a positive in my eyes.

smoogipoo: > at a conceptual level can you confirm you agree with this

Sounds good to me

frenzibyte (approved):

smoogipoo (approved):
Lgtm

peppy:
Hold off momentarily on this one. I was reviewing the osu!-side change today and I'm still not 100% it's the correct way to be handling it (and if it's the only reason to be adding these events then it may be a blocking factor).

smoogipoo:
Awaiting your review, all good.

peppy:
While I think that the `PerformFromScreen` solution is likely better for the original use case this change had, I think adding these `EventArgs` is also beneficial in terms of code quality. I can imaging adding further properties in the future. Let's get this in 👍 .

peppy (approved):

bdach:
Added a [breaking change note](https://github.com/ppy/osu-framework/wiki/Breaking-changes#iscreen-navigation-interface-methods-now-receive-an-event-args-structure) about this on the wiki.

PR #5097, opened 2022-04-14
Upgrade Resharper to 2022.1.0-eap10

smoogipoo:
I've run this 5 times in https://github.com/smoogipoo/osu-framework/runs/6017742926?check_suite_focus=true, seems to be working without deadlocks. Would like to merge into master for further testing.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -15,7 +15,7 @@
       ]
     },
     "jetbrains.resharper.globaltools": {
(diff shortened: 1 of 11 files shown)


peppy (approved):

smoogipoo:
Hold off a bit, unsure if this is actually working when running locally >.>

smoogipoo:
Works. Just took longer than I expected.

peppy (approved):

PR #5099, opened 2022-04-14
Nullability annotations for bindables

huoyaoyuan:
Not bothering the thin value type wrappers (`BindableInt` etc).
The `Value` and `DefaultValue` property is annotated the same as type parameter, allowing both `Bindable<Class>` and `Bindable<Class?>`.

#### Risks

The default constructor of `Bindable<T>` can introduce null to non-null annotated properties. This does not change existing behavior, but *provides a fake sense of safety*. In the other hand, annotating `Value` to be always nullable can introduce too much noise for consumers. The best option should be warning for such constructor call, but it's not possible yet.

The resharper warning is somehow too aggressive that assuming not-yet annotated code to be non-nullable. It warns [here](https://github.com/ppy/osu-framework/blob/7477218aea5a2b38f720d3a6907fbdb0ff863413/osu.Framework/Graphics/Audio/DrawableTrack.cs#L111-L117) where it should be `ValueChangedEvent<IAudioMixer?>`. I tend to think it's not wrong to be defensive for not-yet annotated code.

### DIFF
--- a/osu-framework.sln.DotSettings
+++ b/osu-framework.sln.DotSettings
@@ -58,6 +58,7 @@
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=CollectionNeverQueried_002ELocal/@EntryIndexedValue">HINT</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=CommentTypo/@EntryIndexedValue">HINT</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=CompareOfFloatsByEqualityOperator/@EntryIndexedValue">HINT</s:String>
+	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=ConditionalAccessQualifierIsNonNullableAccordingToAPIContract/@EntryIndexedValue">HINT</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=ConvertClosureToMethodGroup/@EntryIndexedValue">HINT</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=ConvertConditionalTernaryExpressionToSwitchExpression/@EntryIndexedValue">DO_NOT_SHOW</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=ConvertIfDoToWhile/@EntryIndexedValue">WARNING</s:String>
--- a/osu.Framework/Bindables/AggregateBindable.cs
+++ b/osu.Framework/Bindables/AggregateBindable.cs
@@ -5,6 +5,8 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 
+#nullable enable
+
 namespace osu.Framework.Bindables
 {
     /// <summary>
@@ -29,7 +31,7 @@ namespace osu.Framework.Bindables
         /// </summary>
         /// <param name="aggregateFunction">The function to be used for aggregation, taking two input <typeparamref name="T"/> values and returning one output.</param>
         /// <param name="resultBindable">An optional newly constructed bindable to use for <see cref="Result"/>. The initial value of this bindable is used as the initial value for the aggregate.</param>
-        public AggregateBindable(Func<T, T, T> aggregateFunction, Bindable<T> resultBindable = null)
+        public AggregateBindable(Func<T, T, T> aggregateFunction, Bindable<T>? resultBindable = null)
         {
             this.aggregateFunction = aggregateFunction;
             result = resultBindable ?? new Bindable<T>();
@@ -75,10 +77,10 @@ namespace osu.Framework.Bindables
             }
         }
 
-        private WeakRefPair findExistingPair(IBindable<T> bindable) =>
+        private WeakRefPair? findExistingPair(IBindable<T> bindable) =>
             sourceMapping.FirstOrDefault(p => p.WeakReference.TryGetTarget(out var target) && target == bindable);
 
-        private void recalculateAggregate(ValueChangedEvent<T> obj = null)
+        private void recalculateAggregate(ValueChangedEvent<T>? obj = null)
         {
             T calculated = initialValue;
 
--- a/osu.Framework/Bindables/Bindable.cs
+++ b/osu.Framework/Bindables/Bindable.cs
@@ -130,7 +130,7 @@ namespace osu.Framework.Bindables
             TriggerDefaultChange(previousValue, source ?? this, true, bypassChecks);
         }
 
-        private WeakReference<Bindable<T>> weakReferenceInstance;
+        private WeakReference<Bindable<T>>? weakReferenceInstance;
 
         private WeakReference<Bindable<T>> weakReference => weakReferenceInstance ??= new WeakReference<Bindable<T>>(this);
 
@@ -139,7 +139,7 @@ namespace osu.Framework.Bindables
         /// </summary>
         [UsedImplicitly]
         private Bindable()
-            : this(default)
+            : this(default!)
         {
         }
 
@@ -147,12 +147,15 @@ namespace osu.Framework.Bindables
         /// Creates a new bindable instance initialised with a default value.
         /// </summary>
         /// <param name="defaultValue">The initial and default value for this bindable.</param>
-        public Bindable(T defaultValue = default)
+        /// <remarks>Consider to pass a default value for non-nullable T.</remarks>
+        public Bindable(T defaultValue = default!)
         {
-            value = Default = defaultValue;
+            // It lacks a way to represent "warn if called with non-nullable T".
+            // Not verifying to avoid breaking tons of existing usages.
+            value = this.defaultValue = defaultValue;
         }
 
-        protected LockedWeakList<Bindable<T>> Bindings { get; private set; }
+        protected LockedWeakList<Bindable<T>>? Bindings { get; private set; }
 
         void IBindable.BindTo(IBindable them)
         {
@@ -254,7 +257,7 @@ namespace osu.Framework.Bindables
 
                 default:
                     if (underlyingType.IsEnum)
-                        Value = (T)Enum.Parse(underlyingType, input.ToString());
+                        Value = (T)Enum.Parse(underlyingType, input.ToString()!);
                     else
                         Value = (T)Convert.ChangeType(input, underlyingType, CultureInfo.InvariantCulture);
 
@@ -376,7 +379,7 @@ namespace osu.Framework.Bindables
             tThem.removeWeakReference(weakReference);
         }
 
-        public string Description { get; set; }
+        public string? Description { get; set; }
 
         public override string ToString() => value?.ToString() ?? string.Empty;
 
@@ -409,7 +412,8 @@ namespace osu.Framework.Bindables
 
         void ISerializableBindable.DeserializeFrom(JsonReader reader, JsonSerializer serializer)
         {
-            Value = serializer.Deserialize<T>(reader);
+            // Deserialize returns null for json literal "null".
+            Value = serializer.Deserialize<T>(reader)!;
         }
 
         private LeasedBindable<T>? leasedBindable;
--- a/osu.Framework/Bindables/Bindable.cs
+++ b/osu.Framework/Bindables/Bindable.cs
@@ -11,6 +11,8 @@ using osu.Framework.Extensions.TypeExtensions;
 using osu.Framework.IO.Serialization;
 using osu.Framework.Lists;
 
+#nullable enable
+
 namespace osu.Framework.Bindables
 {
     /// <summary>
@@ -22,17 +24,17 @@ namespace osu.Framework.Bindables
         /// <summary>
         /// An event which is raised when <see cref="Value"/> has changed (or manually via <see cref="TriggerValueChange"/>).
         /// </summary>
-        public event Action<ValueChangedEvent<T>> ValueChanged;
+        public event Action<ValueChangedEvent<T>>? ValueChanged;
 
         /// <summary>
         /// An event which is raised when <see cref="Disabled"/> has changed (or manually via <see cref="TriggerDisabledChange"/>).
         /// </summary>
-        public event Action<bool> DisabledChanged;
+        public event Action<bool>? DisabledChanged;
 
         /// <summary>
         /// An event which is raised when <see cref="Default"/> has changed (or manually via <see cref="TriggerDefaultChange"/>).
         /// </summary>
-        public event Action<ValueChangedEvent<T>> DefaultChanged;
+        public event Action<ValueChangedEvent<T>>? DefaultChanged;
 
         private T value;
 
@@ -57,7 +59,7 @@ namespace osu.Framework.Bindables
             }
         }
 
-        internal void SetDisabled(bool value, bool bypassChecks = false, Bindable<T> source = null)
+        internal void SetDisabled(bool value, bool bypassChecks = false, Bindable<T>? source = null)
         {
             if (!bypassChecks)
                 throwIfLeased();
@@ -88,7 +90,7 @@ namespace osu.Framework.Bindables
                 // if the leased bindable decides to disable exclusive access (by setting Disabled = false) then anything will be able to write to Value.
 
                 if (Disabled)
-                    throw new InvalidOperationException($"Can not set value to \"{value.ToString()}\" as bindable is disabled.");
+                    throw new InvalidOperationException($"Can not set value to \"{value}\" as bindable is disabled.");
 
                 if (EqualityComparer<T>.Default.Equals(this.value, value)) return;
 
@@ -96,7 +98,7 @@ namespace osu.Framework.Bindables
             }
         }
 
-        internal void SetValue(T previousValue, T value, bool bypassChecks = false, Bindable<T> source = null)
+        internal void SetValue(T previousValue, T value, bool bypassChecks = false, Bindable<T>? source = null)
         {
             this.value = value;
             TriggerValueChange(previousValue, source ?? this, true, bypassChecks);
@@ -114,7 +116,7 @@ namespace osu.Framework.Bindables
                 // if the leased bindable decides to disable exclusive access (by setting Disabled = false) then anything will be able to write to Default.
 
                 if (Disabled)
-                    throw new InvalidOperationException($"Can not set default value to \"{value.ToString()}\" as bindable is disabled.");
+                    throw new InvalidOperationException($"Can not set default value to \"{value}\" as bindable is disabled.");
 
                 if (EqualityComparer<T>.Default.Equals(defaultValue, value)) return;
 
@@ -122,7 +124,7 @@ namespace osu.Framework.Bindables
             }
         }
 
-        internal void SetDefaultValue(T previousValue, T value, bool bypassChecks = false, Bindable<T> source = null)
+        internal void SetDefaultValue(T previousValue, T value, bool bypassChecks = false, Bindable<T>? source = null)
         {
             defaultValue = value;
             TriggerDefaultChange(previousValue, source ?? this, true, bypassChecks);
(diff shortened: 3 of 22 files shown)


bdach:
This pull really ought to include a preview of what the game-side changes will be to eliminate new nullability warnings. I cannot imagine there are _no_ nullability-related warnings that are going to arise game-side if this is to be merged...?

osu.Framework/Bindables/Bindable.cs
```diff
@@ -137,20 +139,23 @@ internal void SetDefaultValue(T previousValue, T value, bool bypassChecks = fals
...
+            : this(default!)
         {
         }
 
         /// <summary>
         /// Creates a new bindable instance initialised with a default value.
         /// </summary>
         /// <param name="defaultValue">The initial and default value for this bindable.</param>
-        public Bindable(T defaultValue = default)
+        /// <remarks>Consider to pass a default value for non-nullable T.</remarks>
```
bdach, line 150:
```suggestion
        /// <remarks>Consider to pass a default value for non-nullable <typeparamref name="T"/>.</remarks>
```

osu.Framework/Bindables/Bindable.cs
```diff
@@ -137,20 +139,23 @@ internal void SetDefaultValue(T previousValue, T value, bool bypassChecks = fals
...
+        public Bindable(T defaultValue = default!)
         {
-            value = Default = defaultValue;
+            // It lacks a way to represent "warn if called with non-nullable T".
+            // Not verifying to avoid breaking tons of existing usages.
```
bdach, line 154: Not sure what this comment means - is it trying to say that it is impossible to force the user to pass a non-null default if they're creating a `Bindable<T>` with a non-nullable inner type?

osu.Framework/Bindables/Bindable.cs
```diff
@@ -359,8 +364,7 @@ public void UnbindBindings()
 
         internal virtual void UnbindAllInternal()
         {
-            if (isLeased)
-                leasedBindable.Return();
+            leasedBindable?.Return();
```
bdach, line 367: Not sure why this was touched in this pull given that this has very little to do with nullability.

osu.Framework/Bindables/Bindable.cs
huoyaoyuan: Yes. That's what I mentioned in risk of PR description.

osu.Framework/Bindables/Bindable.cs
huoyaoyuan: Because the `isLeased` property is not annotated with `MemberNotNull`, and we haven't removed netstandard yet, it's a bit complex to add the annotation. Since the property is private, I'm simply adjusting its usage.

huoyaoyuan:
There's no warning when building game repo, since most of them haven't enabled nullability yet.

InspectCode reports some warnings. There's the diff to fix:
```diff
diff --git a/osu.Game/Overlays/Chat/ChatTextBar.cs b/osu.Game/Overlays/Chat/ChatTextBar.cs
index ef20149dac..029c3a1b7c 100644
--- a/osu.Game/Overlays/Chat/ChatTextBar.cs
+++ b/osu.Game/Overlays/Chat/ChatTextBar.cs
@@ -26,7 +26,7 @@ public class ChatTextBar : Container
         public event Action<string>? OnSearchTermsChanged;

         [Resolved]
-        private Bindable<Channel> currentChannel { get; set; } = null!;
+        private Bindable<Channel?> currentChannel { get; set; } = null!;

         private OsuTextFlowContainer chattingTextContainer = null!;
         private Container searchIconContainer = null!;
@@ -126,7 +126,7 @@ protected override void LoadComplete()

             currentChannel.BindValueChanged(change =>
             {
-                Channel newChannel = change.NewValue;
+                var newChannel = change.NewValue;

                 switch (newChannel?.Type)
                 {
diff --git a/osu.Game/Screens/OnlinePlay/Lounge/LoungeBackgroundScreen.cs b/osu.Game/Screens/OnlinePlay/Lounge/LoungeBackgroundScreen.cs
index 52a902f5da..bef99f3737 100644
--- a/osu.Game/Screens/OnlinePlay/Lounge/LoungeBackgroundScreen.cs
+++ b/osu.Game/Screens/OnlinePlay/Lounge/LoungeBackgroundScreen.cs
@@ -13,7 +13,7 @@ namespace osu.Game.Screens.OnlinePlay.Lounge
 {
     public class LoungeBackgroundScreen : OnlinePlayBackgroundScreen
     {
-        public readonly Bindable<Room> SelectedRoom = new Bindable<Room>();
+        public readonly Bindable<Room?> SelectedRoom = new Bindable<Room?>();
         private readonly BindableList<PlaylistItem> playlist = new BindableList<PlaylistItem>();

         public LoungeBackgroundScreen()
@@ -22,7 +22,7 @@ public LoungeBackgroundScreen()
             playlist.BindCollectionChanged((_, __) => PlaylistItem = playlist.GetCurrentItem());
         }

-        private void onSelectedRoomChanged(ValueChangedEvent<Room> room)
+        private void onSelectedRoomChanged(ValueChangedEvent<Room?> room)
         {
             if (room.OldValue != null)
                 playlist.UnbindFrom(room.OldValue.Playlist);
diff --git a/osu.Game/Screens/Play/PlayerSettings/BeatmapOffsetControl.cs b/osu.Game/Screens/Play/PlayerSettings/BeatmapOffsetControl.cs
index 1662ca399f..bcb7266338 100644
--- a/osu.Game/Screens/Play/PlayerSettings/BeatmapOffsetControl.cs
+++ b/osu.Game/Screens/Play/PlayerSettings/BeatmapOffsetControl.cs
@@ -29,7 +29,7 @@ namespace osu.Game.Screens.Play.PlayerSettings
 {
     public class BeatmapOffsetControl : CompositeDrawable
     {
-        public Bindable<ScoreInfo> ReferenceScore { get; } = new Bindable<ScoreInfo>();
+        public Bindable<ScoreInfo?> ReferenceScore { get; } = new Bindable<ScoreInfo?>();

         public BindableDouble Current { get; } = new BindableDouble
         {
@@ -180,7 +180,7 @@ void updateOffset()
             }
         }

-        private void scoreChanged(ValueChangedEvent<ScoreInfo> score)
+        private void scoreChanged(ValueChangedEvent<ScoreInfo?> score)
         {
             referenceScoreContainer.Clear();
```

As I concerned, it's hard to identify whether `Bindable<T>` or `Bindable<T?>` is correct. I also searched for the references of `IBindable<T>.Value`, and see most of them not checking for nulls. This also makes it painful to annotate `Value` as always-nullable.

osu-framework.sln.DotSettings
```diff
@@ -58,6 +58,7 @@
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=CollectionNeverQueried_002ELocal/@EntryIndexedValue">HINT</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=CommentTypo/@EntryIndexedValue">HINT</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=CompareOfFloatsByEqualityOperator/@EntryIndexedValue">HINT</s:String>
+	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=ConditionalAccessQualifierIsNonNullableAccordingToAPIContract/@EntryIndexedValue">HINT</s:String>
```
smoogipoo, line 61: As I mentioned in https://github.com/ppy/osu-framework/pull/5207#issuecomment-1139453312, we definitely do not want this. Use the new extensions if needed.

osu.Framework/Bindables/Bindable.cs
```diff
@@ -137,20 +139,23 @@ internal void SetDefaultValue(T previousValue, T value, bool bypassChecks = fals
...
+            : this(default!)
         {
         }
 
         /// <summary>
         /// Creates a new bindable instance initialised with a default value.
         /// </summary>
         /// <param name="defaultValue">The initial and default value for this bindable.</param>
-        public Bindable(T defaultValue = default)
+        /// <remarks>Consider to pass a default value for non-nullable <typeparamref name="T"/>.</remarks>
```
smoogipoo, line 150:
```suggestion
        /// <remarks>Consider passing a default value for non-nullable <typeparamref name="T"/>.</remarks>
```

osu.Framework/Bindables/Bindable.cs
smoogipoo: Probably just remove this comment and the derived ctor (`BindableWithCurrent`, etc), as they read bad.

osu.Framework/Bindables/Bindable.cs
smoogipoo: I'd probably prefer:
```
if (isLeased)
{
    Debug.Assert(leasedBindable != null);
    leasedBindable.Return();
}
```
Or maybe `leasedBindable.AsNonNull().Return()` also works.

smoogipoo (changes requested):

osu.Framework/Bindables/Bindable.cs
huoyaoyuan: Then we should check the usage of this ctor, and don't forget DI constructed ones.
Let lazy initialised instances to use explicit `null!` should be fine, but it may require a lot of code change.

smoogipoo: Sorry, I meant just removing the comments, not the ctors altogether. I'm not sure whether anything can (ever?) be done here.

Could use the `IsNullable()` overloads, but they have too much overhead.

Are the .NET team aware of this? It doesn't seem like an esoteric use case.

huoyaoyuan: This is the same pain of `default(ImmutableArray<T>)`. In that case `ImmutableArray` was designed for heavily used in hot path, so the decision is to let NRE blows up.

This is also a delay initialisation problem like DI. There is also no guarantee that initialisation (load) must be invoked after creating the object.

AFAIK, tracking delay initialisation is too far for language design.

I think we should look at the usages to see if delay initilisation of bindable can be avoided, and convert to delay create the `Bindable` instance instead.

huoyaoyuan: Another consideration: we may remove the default parameter, and add factories for creating bindables. For example:

```csharp
// not providing default value will always provide a nullable bindable
Bindable<T?> Create<T>() => new Bindable<T?>(null);
```

smoogipoo: I wouldn't call this delayed initialisation. Surely it's not within the scope of NRT/the language to determine that `T x = default` is `null` in a context where nulls aren't allowed, so the param _must_ be provided? In addition, I'd hope that `T x = default!` bypasses this behaviour as it does for `T x = null!`.

A Roslyn analyser for this would be quite simple to craft, and I'll do it myself provided the time.

Just to make sure we're on the same page, I would _require_ cases like this to provide a parameter. The rules of constructing bindables should be simple:
```csharp
// <T> is <int>, param not required.
public readonly Bindable<int> MyBindable1 = new Bindable<int>();

// <T> is <int?>, param not required.
public readonly Bindable<int?> MyBindable2 = new Bindable<int?>();

// <T> is <object>, param required.
public readonly Bindable<object> MyBindable3 = new Bindable<object>(new object());

// <T> is <object?>, param not required.
public readonly Bindable<object?> MyBindable4 = new Bindable<object?>();

public MyObject()
{
    // Same rules as above.
    MyBindable1 = new Bindable<int>();
    MyBindable2 = new Bindable<int?>();
    MyBindable3 = new Bindable<object>(new object());
    MyBindable4 = new Bindable<object?>();
}
```
Anything else is incorrect use:
```csharp
// Incorrect
public readonly Bindable<object> MyBindable1 = new Bindable<object>();

public MyObject()
{
    MyBindable1.Value = new object();
}
```

osu.Framework/Bindables/AggregateBindable.cs
```diff
@@ -5,6 +5,8 @@
 using System.Collections.Generic;
 using System.Linq;
 
+#nullable enable
+
```
andy840119, line 9: Seems nullable enable can be removed after #5241 ?

huoyaoyuan: Yep, all the `enable`s should go away and `disable`s be introduced as needed.

andy840119:
I think @huoyaoyuan's commits are good enough.
I guess i will based in this PR if he's not working on this PR anymore.
And i maybe not working on this part recently(just notice that `Bindable<T>` might need to  re-write as `Bindable<T?>` in osu side for prevent warning. I need to make a think about this.)

osu.Framework/Bindables/Bindable.cs
huoyaoyuan: Yes, this is the desired behavior I want. A custom analyser is good and we don't need to do workarounds about language features.

huoyaoyuan:
@smoogipoo I'm still interested, but I didn't have time to think about the design question here.

> A Roslyn analyser for this would be quite simple to craft, and I'll do it myself provided the time.

> Just to make sure we're on the same page, I would require cases like this to provide a parameter.

So the conclusion is to depend on a custom analyser right? I'll update this PR.

smoogipoo:
> So the conclusion is to depend on a custom analyser right? I'll update this PR.

This PR can happen without an analyzer.

huoyaoyuan:
CodeFactor is complaining about file I don't touch.

huoyaoyuan:
Mention me if there is still any issue remaining. I'm a little busy to track a lot of things these days.

osu.Framework/Localisation/ILocalisedBindableString.cs
```diff
@@ -5,6 +5,8 @@
 
 using osu.Framework.Bindables;
 
+#nullable enable
```
peppy, line 8: Should be removed?

osu.Framework/Bindables/IHasDescription.cs
```diff
@@ -13,6 +11,6 @@ public interface IHasDescription
         /// <summary>
         /// The description for this object.
         /// </summary>
-        string Description { get; }
+        string? Description { get; }
```
peppy, line 14: I think we should use `string.Empty` rather than allow nullable, not sure though.

osu.Framework/Bindables/Bindable.cs
```diff
@@ -139,20 +138,22 @@ internal void SetDefaultValue(T previousValue, T value, bool bypassChecks = fals
...
+            : this(default!)
         {
         }
 
         /// <summary>
         /// Creates a new bindable instance initialised with a default value.
         /// </summary>
         /// <param name="defaultValue">The initial and default value for this bindable.</param>
-        public Bindable(T defaultValue = default)
+        /// <remarks>Consider passing a default value for non-nullable <typeparamref name="T"/>.</remarks>
```
peppy, line 149: Not sure this is required? If a remark is here it should explain the *why*, not just "consider using optional parameter".

osu.Framework/Bindables/BindableWithCurrent.cs
```diff
@@ -28,9 +26,10 @@ public Bindable<T> Current
             }
         }
 
-        public BindableWithCurrent(T defaultValue = default)
+        public BindableWithCurrent(T defaultValue = default!)
             : base(defaultValue)
         {
+            // Not verifying, as base Bindable constructor
```
peppy, line 32: Is this an incomplete comment?

smoogipoo: The base class comment was since removed, this one should be too.

osu.Framework/Bindables/Bindable.cs
```diff
@@ -410,10 +413,13 @@ void ISerializableBindable.SerializeTo(JsonWriter writer, JsonSerializer seriali
 
         void ISerializableBindable.DeserializeFrom(JsonReader reader, JsonSerializer serializer)
         {
-            Value = serializer.Deserialize<T>(reader);
+            // Deserialize returns null for json literal "null".
+            var result = serializer.Deserialize<T>(reader);
+            Debug.Assert(result != null);
+            Value = result;
```
smoogipoo, line 419: I assume this is for a case like `Bindable<int?>`, right? In which case, if it returns `null` then I don't think this should assert. Probably use `.AsNonNull()` instead (basically just a null-forgiving now, no longer asserts internally).

smoogipoo (changes requested):
Rest looks pretty good to me. Will have to see how this works with the osu! repo before merge.

osu.Framework/Bindables/Bindable.cs
huoyaoyuan: The last change was by @frenzibyte 
The assert should be only valid when `T` is nullable, which is not detectable at all in runtime.

frenzibyte: Also `AsNonNull` can't be used for non-constrained generic types, we might want to apply what was proposed in [discord](https://discord.com/channels/188630481301012481/589331078574112768/992773358146044037) to get it working correctly with no constraints.

huoyaoyuan: Deserialization is a well-known hole of nullable annotations.
Here null forgiving operator should be the only option. It means we are bypassing any check because no check can be done.

huoyaoyuan: Note for you: `[return: NotNull]` is no longer needed at all since C# 9. `?` can be used on unconstrained type parameter.
So is there any decision here? I can't find anything valid other than `!`.

smoogipoo: Wait for https://github.com/ppy/osu-framework/pull/5296 then use `.AsNonNull()`.

huoyaoyuan: @smoogipoo I've already updated this. Is there still any issue remaining?

huoyaoyuan:
@peppy What do you think about this PR? Should I revive this or redo a new one?

peppy:
> @peppy What do you think about this PR? Should I revive this or redo a new one?

If you can resolve the conflicts, I think it's still in a state we could take another look at it. I have forgotten what state of review this was in, but on a quick check it looks like it was basically waiting for tooling fixes for NRT, so maybe there's hope still.

bdach:
- Code quality is failing.
- I will reiterate [what I already said above](https://github.com/ppy/osu-framework/pull/5099#issuecomment-1100711076) that this pull should be bundled with a preview of game-side changes that will follow after this pull. Over the last year we have been progressively enabling nullability on more and more files, so I find it unlikely that there will be no changes required.

huoyaoyuan:
> * Code quality is failing.

In tests it's explicitly testing exception thrown with null input. That's the reason why .NET runtime projects don't turn on NRT in their test projects. We have `#nullable disable` there but Resharper is still analyzing them.

huoyaoyuan:
The game side change is at https://github.com/ppy/osu/pull/23680.

huoyaoyuan:
https://github.com/ppy/osu-framework/blob/c274d818c128ee4b9be114ee72f8ff8b0ffc9781/osu.Framework.Tests/Bindables/AggregateBindableTest.cs#L53

This line is really causing troubles:

The initial null value in constructor is currently suppressed. However, InspectCode is too strict about being conservative around nulls. Does there need an `AsNullable` extension method to explicitly state this?

osu.Framework/Bindables/RangeConstrainedBindable.cs
```diff
@@ -64,7 +62,7 @@ public override T Value
         public bool HasDefinedRange => !EqualityComparer<T>.Default.Equals(MinValue, DefaultMinValue) ||
                                        !EqualityComparer<T>.Default.Equals(MaxValue, DefaultMaxValue);
 
-        protected RangeConstrainedBindable(T defaultValue = default)
+        protected RangeConstrainedBindable(T defaultValue = default!)
```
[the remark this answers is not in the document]
huoyaoyuan: Actually I planned to add more generic math constraint in the following. I'm supporting to add `struct` constraint.

osu.Framework/Bindables/BindableList.cs
```diff
@@ -388,25 +385,25 @@ public void CopyTo(Array array, int index)
...
+            set => this[index] = (T)value.AsNonNull();
         }
 
-        int IList.Add(object value)
+        int IList.Add(object? value)
         {
-            Add((T)value);
+            Add((T)value.AsNonNull());
             return Count - 1;
         }
```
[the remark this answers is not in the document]
huoyaoyuan: Some of the code are really ancient from 3 years or even 5 years ago. I need to catch up about new usages.

PR #5101, opened 2022-04-14
Remove escape to exit test browser

peppy:
This has been a pain point while testing for the longest time, where you hit escape expecting to interact with a test scene but instead exit the test browser.

Use Alt+F4, Cmd-W/Q or similar to exit instead. Or hook escape back up in your own `TestBrowser` implementation if you so choose.

### DIFF
--- a/osu.Framework.Tests/TestGame.cs
+++ b/osu.Framework.Tests/TestGame.cs
@@ -18,6 +18,6 @@ namespace osu.Framework.Tests
             Resources.AddStore(new NamespacedResourceStore<byte[]>(new DllResourceStore(typeof(TestGame).Assembly), "Resources"));
         }
 
-        protected override bool OnExiting() => BlockExit.Value;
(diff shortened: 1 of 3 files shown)


frenzibyte (approved):

bdach:
I don't know about merging this so fast, I would have left it be and ensure there's consensus on this I think? Not that I personally mind, I even prefer it this way, it's just a pretty major workflow change that came relatively out of the blue...

bdach:
Hmm, I suppose there was. It hid away from me (not on the channel I'd expect to see it, and also surrounded by webhook notifications). Would appreciate a link to the relevant discussion in the future in such cases.

frenzibyte:
Yeah I have only merged it based on the [consensus in discord](https://discord.com/channels/188630481301012481/188630652340404224/964114179051126784), otherwise I would've definitely let it be with just an approval. Attaching a link sounds fair, will keep that in mind.

PR #5102, opened 2022-04-14
Always use "No sound" audio device on headless test runs

frenzibyte:
Over the past few days, it has been quite annoying running full tests suite in terms of audio, as initializing mixers could result in some soft audio ticks (due to mixers having 0ms buffer), and some tests may directly modify the audio volume levels for testing convenience (see https://github.com/ppy/osu/blob/0099af87ff7e43e75c30c2105cd39fc52cfc996a/osu.Game.Tests/Visual/Gameplay/TestScenePlayerLoader.cs#L239-L249).

Have gone with the simplest fix of limiting audio devices to only "No sound", to make sure absolutely nothing is outputted to any of my actual audio devices. This is done as a better replacement to the current `AudioManager.Volume.Value = 0` logic in `TestSceneTestRunner`.

Alternative solution would be muting BASS at a global level, but we might as well never output to an actual audio device in that case to begin with.

### DIFF
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -15,6 +15,7 @@ using osu.Framework.Audio.Mixing.Bass;
 using osu.Framework.Audio.Sample;
 using osu.Framework.Audio.Track;
 using osu.Framework.Bindables;
+using osu.Framework.Development;
 using osu.Framework.Extensions.TypeExtensions;
 using osu.Framework.IO.Stores;
 using osu.Framework.Logging;
@@ -309,6 +310,10 @@ namespace osu.Framework.Audio
 
             // device is invalid
             if (!device.IsEnabled)
+                return false;
+
+            // we don't want bass initializing with real audio device on headless test runs.
+            if (deviceIndex != Bass.NoSoundDevice && DebugUtils.IsNUnitRunning)
                 return false;
 
             // initialize new device
--- a/osu.Framework/Testing/TestSceneTestRunner.cs
+++ b/osu.Framework/Testing/TestSceneTestRunner.cs
@@ -6,8 +6,6 @@ using System.Runtime.ExceptionServices;
 using System.Threading;
 using NUnit.Framework;
 using osu.Framework.Allocation;
-using osu.Framework.Bindables;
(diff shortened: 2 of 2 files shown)


peppy:
I do wonder if this will help with CI issues we've faced (ie the one in #5062)

osu.Framework/Audio/AudioManager.cs
```diff
@@ -311,6 +312,10 @@ private bool setAudioDevice(int deviceIndex)
             if (!device.IsEnabled)
                 return false;
 
+            // we don't want bass initializing with real audio device on headless test runs.
+            if (deviceIndex != Bass.NoSoundDevice && DebugUtils.IsNUnitRunning)
+                return false;
```
bdach, line 317: In general I'm not very comfortable with test-only paths residing in general components like `AudioManager`. Does this need to be done this way? Could the nunit test runner not set the current audio device setting to "no sound" or something, kind of like what the old logic did with volume? I'm of the opinion that `AudioManager` itself should neither know or care whether tests are running.

frenzibyte: Hmm, `AudioManager` only ever uses `NoSound` when no other audio devices are present in the list. 

One way I can see this happening is by a subclass `AudioManager` which has this logic as an override to `setAudioDevice`, but that would mean a `CreateAudioManager` method in `Game`? Would you be fine with that, or do you have another idea in mind.

bdach: Not sure about inheriting. I'd honestly consider exposing "no sound" as an actual device and using that at config level as the simplest solution Probably best to see other views on this first.

peppy: I think it's fine to do as is done in this PR, as we seem to do this in one other place for audio related issues:

https://github.com/ppy/osu-framework/blob/8b1a3d6aee3f1a56d22dcd18e07fcb73458175e4/osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs#L409-L411

I'd rather not expose "no sound" until we need to (ie. until a framework user actually requests it with a valid purpose).

peppy (approved):

smoogipoo (approved):

PR #5106, opened 2022-04-16
Fix iOS keyboard handler not switching properly with physical keyboard connected

frenzibyte:
- Should (resolve) ppy/osu#17513

When the software keyboard hides, the end value can potentially have a height greater than 120 (at least from what I've seen in https://github.com/ppy/osu/issues/17513#issuecomment-1100763055). Therefore there's a chance for the raw keyboard handler to remain disabled even after hiding the software keyboard.

A better solution is to check the distance between the software keyboard and the bottom of the game view instead.

Note that I'm not 100% sure whether this solves ppy/osu#17513 as I don't own an iPad nor an Apple Pencil to test this. Have only been relying on the iOS simulator.


peppy:
I'm willing to merge this as it's probably less-broken-than-before but...

> A better solution is to check the distance between the software keyboard and the bottom of the game view instead.

Not sure about this, given the software keyboard my be floating. I'm also not sure how well the 120px height check holds out. There's got to be a more reliable way of doing this. I fail to believe the correct way of checking whether a keyboard is present is to check its frame height.

frenzibyte:
Starting at iOS 14+, we can check whether a physical keyboard is connected via [`GCKeyboard.CoalescedKeyboard`](https://developer.apple.com/documentation/gamecontroller/gckeyboard/3626173-coalesced/) and keep the keyboard frame detection as a fall back for early iOS versions.

One thing I'm not sure about which will break this is whether you can toggle the software keyboard even with a physical keyboard connected, because if that's possible, it'll not work properly as the physical keyboard is still connected in that case and the "non-raw" keyboard handler will remain disabled.

I am capable of doing that in iOS simulator via "Toggle Software Keyboard" in menu, but not sure if that's just a debug option available in simulator only.

peppy:
Is this what you mean?

https://user-images.githubusercontent.com/191335/163698087-f8a2f74d-634a-40ca-a136-fed34b51fce9.mp4

peppy:
Here's another edge case you may need to consider with height checks: this is a custom keyboard i just made which is definitely not 120px high. This can be the case on iOS and iPadOS:

IMG_0469

frenzibyte:
From testing on iOS simulator, it appears I can avoid this entirely and keep the `HiddenTextField`-based keyboard handler active even when a hardware keyboard is connected, as typing with the hardware keyboard sends correct events to the `HiddenTextField`, therefore not requiring the hardware keyboard handler to be active in that case.

I've also renamed the handlers as below, to reflect what they're actually doing:
 - `IOSKeyboardHandler` -> `IOSTextFieldKeyboardHandler`
 - `IOSRawKeyboardHandler` -> `IOSHardwareKeyboardHandler`

This will require testing on a real device as I can only confirm this is working as intended via simulator.

frenzibyte:
On second thought, the latest changes will actually cause all key commands such as cut/copy/paste/select to no longer work when performed via hardware keyboard, since they are usually not handled via `HiddenTextField` (hence the `CanPerformAction` override), to keep the commands propagating to the responder which forwards them to the `IOSRawKeyboardHandler`/`IOSHardwareKeyboardHandler`.

This is as far as I could go with attempting to bring these simple key commands back:
```diff
diff --git a/osu.Framework.iOS/IOSGameView.cs b/osu.Framework.iOS/IOSGameView.cs
index 22241e5a8..07873974c 100644
--- a/osu.Framework.iOS/IOSGameView.cs
+++ b/osu.Framework.iOS/IOSGameView.cs
@@ -2,9 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Collections.Generic;
 using System.Drawing;
-using System.Linq;
 using System.Threading.Tasks;
 using CoreAnimation;
 using Foundation;
@@ -133,7 +131,7 @@ public class HiddenTextField : UITextField
         {
             public event Action<NSRange, string> HandleShouldChangeCharacters;
             public event Action HandleShouldReturn;
-            public event Action<UIKeyCommand> HandleKeyCommand;
+            public event Action<NSString, UIKeyModifierFlags> HandleKeyCommand;
 
             /// <summary>
             /// Placeholder text that the <see cref="HiddenTextField"/> will be populated with after every keystroke.
@@ -147,23 +145,6 @@ public class HiddenTextField : UITextField
 
             private int responderSemaphore;
 
-            private readonly IEnumerable<Selector> softwareBlockedActions = new[]
-            {
-                new Selector("cut:"),
-                new Selector("copy:"),
-                new Selector("select:"),
-                new Selector("selectAll:"),
-            };
-
-            private readonly IEnumerable<Selector> rawBlockedActions = new[]
-            {
-                new Selector("cut:"),
-                new Selector("copy:"),
-                new Selector("paste:"),
-                new Selector("select:"),
-                new Selector("selectAll:"),
-            };
-
             public override UITextSmartDashesType SmartDashesType => UITextSmartDashesType.No;
             public override UITextSmartInsertDeleteType SmartInsertDeleteType => UITextSmartInsertDeleteType.No;
             public override UITextSmartQuotesType SmartQuotesType => UITextSmartQuotesType.No;
@@ -206,25 +187,38 @@ public HiddenTextField()
 
             public override UIKeyCommand[] KeyCommands => new[]
             {
-                UIKeyCommand.Create(UIKeyCommand.LeftArrow, 0, new Selector("keyPressed:")),
-                UIKeyCommand.Create(UIKeyCommand.RightArrow, 0, new Selector("keyPressed:")),
-                UIKeyCommand.Create(UIKeyCommand.UpArrow, 0, new Selector("keyPressed:")),
-                UIKeyCommand.Create(UIKeyCommand.DownArrow, 0, new Selector("keyPressed:"))
+                UIKeyCommand.Create(UIKeyCommand.LeftArrow, 0, new Selector("arrow:")),
+                UIKeyCommand.Create(UIKeyCommand.RightArrow, 0, new Selector("arrow:")),
+                UIKeyCommand.Create(UIKeyCommand.UpArrow, 0, new Selector("arrow:")),
+                UIKeyCommand.Create(UIKeyCommand.DownArrow, 0, new Selector("arrow:")),
             };
 
             public override bool CanPerform(Selector action, NSObject withSender)
             {
-                if ((!softwareKeyboard && rawBlockedActions.Contains(action)) || (softwareKeyboard && softwareBlockedActions.Contains(action)))
-                    return false;
+                switch (action.Name)
+                {
+                    case "cut:":
+                    case "copy:":
+                    case "paste:":
+                    case "select:":
+                    case "selectAll:":
+                        return true;
+                }
 
                 return base.CanPerform(action, withSender);
             }
 
-            [Export("keyPressed:")]
-            private void keyPressed(UIKeyCommand cmd) => HandleKeyCommand?.Invoke(cmd);
+            public override void Cut(NSObject _) => HandleKeyCommand?.Invoke(new NSString("x"), UIKeyModifierFlags.Command);
+            public override void Copy(NSObject _) => HandleKeyCommand?.Invoke(new NSString("c"), UIKeyModifierFlags.Command);
+            public override void Paste(NSObject _) => HandleKeyCommand?.Invoke(new NSString("v"), UIKeyModifierFlags.Command);
+            public override void SelectAll(NSObject _) => HandleKeyCommand?.Invoke(new NSString("a"), UIKeyModifierFlags.Command);
+
+            [Export("arrow:")]
+            private void arrow(UIKeyCommand cmd) => HandleKeyCommand?.Invoke(cmd.Input, cmd.ModifierFlags);
 
             private void resetText()
             {
+                IUITextInput
                 if (SoftwareKeyboard)
                 {
                     // we put in some dummy text and move the cursor to the middle so that backspace (and potentially delete or cursor keys) will be detected
diff --git a/osu.Framework.iOS/Input/IOSTextFieldKeyboardHandler.cs b/osu.Framework.iOS/Input/IOSTextFieldKeyboardHandler.cs
index e74f58518..fdc39bd2a 100644
--- a/osu.Framework.iOS/Input/IOSTextFieldKeyboardHandler.cs
+++ b/osu.Framework.iOS/Input/IOSTextFieldKeyboardHandler.cs
@@ -74,7 +74,7 @@ private void handleShouldReturn()
             PendingInputs.Enqueue(new KeyboardKeyInput(Key.Enter, false));
         }
 
-        private void handleKeyCommand(UIKeyCommand cmd)
+        private void handleKeyCommand(NSString input, UIKeyModifierFlags modifiers)
         {
             if (!IsActive)
                 return;
@@ -83,23 +83,23 @@ private void handleKeyCommand(UIKeyCommand cmd)
             bool upper = false;
 
             // UIKeyCommand constants are not actually constants, so we can't use a switch
-            if (cmd.Input == UIKeyCommand.LeftArrow)
+            if (input == UIKeyCommand.LeftArrow)
                 key = Key.Left;
-            else if (cmd.Input == UIKeyCommand.RightArrow)
+            else if (input == UIKeyCommand.RightArrow)
                 key = Key.Right;
-            else if (cmd.Input == UIKeyCommand.UpArrow)
+            else if (input == UIKeyCommand.UpArrow)
                 key = Key.Up;
-            else if (cmd.Input == UIKeyCommand.DownArrow)
+            else if (input == UIKeyCommand.DownArrow)
                 key = Key.Down;
             else
-                key = keyForString(cmd.Input, out upper);
+                key = keyForString(input, out upper);
 
             if (!key.HasValue) return;
 
-            bool shiftHeld = (cmd.ModifierFlags & UIKeyModifierFlags.Shift) > 0 || upper;
-            bool superHeld = (cmd.ModifierFlags & UIKeyModifierFlags.Command) > 0;
-            bool ctrlHeld = (cmd.ModifierFlags & UIKeyModifierFlags.Control) > 0;
-            bool optionHeld = (cmd.ModifierFlags & UIKeyModifierFlags.Alternate) > 0;
+            bool shiftHeld = (modifiers & UIKeyModifierFlags.Shift) > 0 || upper;
+            bool superHeld = (modifiers & UIKeyModifierFlags.Command) > 0;
+            bool ctrlHeld = (modifiers & UIKeyModifierFlags.Control) > 0;
+            bool optionHeld = (modifiers & UIKeyModifierFlags.Alternate) > 0;
 
             if (shiftHeld) PendingInputs.Enqueue(new KeyboardKeyInput(Key.LShift, true));
             if (superHeld) PendingInputs.Enqueue(new KeyboardKeyInput(Key.LWin, true));
```

I personally feel the best course of action here is to remain at the keyboard height/distance check, and reassess all of this logic once osu!framework runs under Metal/MoltenVK so that testing changes via simulator doesn't be a pain in the ass.

I can imagine `HiddenTextField` changing to a more proper `IUITextInput`-based view backed by a switchable `TextBox` or something along the lines of that.

Also, the mentioned small keyboard edge case should be minor, as it will only make the software keyboard unresponsive, not affect anything else.

peppy:
Looks to work well, tested two keyboards and switching between soft keyboard and back again with textbox active.

peppy (approved):

PR #5107, opened 2022-04-17
Refactor iOS raw keyboard handler to rely on `UIPress` events instead

frenzibyte:
Previously, iOS raw keyboard handling were based on listening to `GSEvent`s which come from a private framework. Decoding `GSEvent`s were purely based on reading the event data and looking for the correct values via byte offsets, which doesn't even work properly on iOS simulator (no key input registered from simulated hardware keyboard).

Starting at iOS 13.4+, `UIPressesEvent`s provide pure `UIKey`s for knowing which key on the hardware keyboard has been pressed/released. Simplifying the entire process of handling hardware keyboard input and actually making it work on iOS simulator.

Since we still support down to iOS 10, I've kept the old `GSEvent` logic for early iOS versions. Not sure if it can actually be kept once osu! gets published to the App Store, however.

Note that, while this is working from testing on iOS simulator, I'm not entirely sure if this works properly for other generic physical keyboards, and I don't own any to test with my own device on.

- [x] Pending test on generic physical keyboards


osu.Framework.iOS/Input/IOSRawKeyboardHandler.cs
```diff
@@ -12,128 +13,465 @@ namespace osu.Framework.iOS.Input
...
+                    case UIPressPhase.Ended:
+                    case UIPressPhase.Cancelled:
+                        PendingInputs.Enqueue(new KeyboardKeyInput(key, false));
+                        break;
+                }
+
+                hasArrowKeys |= press.Type == UIPressType.UpArrow ||
+                                press.Type == UIPressType.DownArrow ||
+                                press.Type == UIPressType.LeftArrow ||
+                                press.Type == UIPressType.RightArrow;
```
frenzibyte, line 71: I'm not really 100% sure why this is a requirement (came from the old `GSEvent` logic in `GameUIApplication`), and trying to test this with osu! was a complete pain given iOS simulator and its extremely slow OpenGL support (i.e. software renderer).

osu.Framework.iOS/Input/IOSRawKeyboardHandler.cs
```diff
@@ -12,128 +13,465 @@ namespace osu.Framework.iOS.Input
...
+                //     return Key.PrintScreen;
+                //
+                // case UIKeyboardHidUsage.KeyboardScrollLock:
+                //     return Key.ScrollLock;
+                //
+                // case UIKeyboardHidUsage.KeyboardPause:
+                //     return Key.Pause;
+                //
+                // case UIKeyboardHidUsage.KeyboardInsert:
+                //     return Key.Insert;
```
frenzibyte, line 313: I've left a few commented lines like this on this keycode switch here just to highlight whether we actually want it to be working like this. I still kept this mapped as was done in the previous `keyMap` dictionary.

peppy:
I've tested this isolated, but needs merge resolution with the already-merged PR.

frenzibyte:
> I've tested this isolated, but needs merge resolution with the already-merged PR.

Have resolved conflicts. For reference, did you test with a generic keyboard? mainly concerned about those.

peppy:
Needs conflict resolution again

peppy (approved):

PR #5109, opened 2022-04-18
Remove local vertex storage from VertexBuffer

smoogipoo:
**This is an RFC. I will rebase and clean up this PR so please don't attempt to go through by-commit**

@peppy @bdach @frenzibyte This PR is in a "final" state for me, but I want your opinions and approvals on the structure of things and my ideas proposed before I go ahead with rebasing and cleaning it up. I've attempted to explain things (minus making a breaking changes section) so let me know if something doesn't make sense.

Some things like the condition added in `PathDrawNode` can/will be split out into separate PRs.

---

# Abstract

`VertexBuffer` currently houses CPU-side storage for all added vertices in order to de-dupe vertex uploads. This is inefficient in multiple regards:
- Permanently increases memory footprint.
- Possibly causes stutters as a result of the ImageSharp array pool allocating finalisable objects.
- Increases processing time for vertices. Every vertex addition, new or not, compares equality.

But we have a mechanism to avoid this - `Drawable` invalidation. We know exactly when `Drawable`s change and when `DrawNode`s are expected to draw new vertices as a result, which can be used to relax our requirement for vertex equality and remove the CPU-side storage.

# Batching

Batching requires constant maintenance to improve, and our current batching mechanisms are starting to show their age. There is a lot that can be done to improve batching, such as having an allocator with more knowledge over the structure of the scene graph to decide which vertices to batch together.

`DrawNode` should eventually not be able to create their own batches. Any methods of creating custom batches should be seen as temporary pending further implementation of such an allocator, at which point they'll be replaced by a lower-level solution.

# `IRenderer`

To cover the above changes to batching and future work in eventually removing `GLWrapper` entirety, the new `IRenderer` becomes the single new parameter of `DrawNode.Draw()`:

```diff
+ public interface IRenderer
+ {
+     // Will eventually replace GLWrapper.
+ }
+ 
+ public class OpenGLRenderer : IRenderer
+ {
+ }
  
  public class DrawNode
  {
-     public virtual void Draw(Action<TexturedVertex2D> vertexAction)
+     public virtual void Draw(IRenderer renderer)
      {
          // Actual drawing commands.
      }
  }

```

**For the time being**, `IRenderer.BeginQuads()` is exposed to interact with the parenting `CompositeDrawable`'s/global quad batch.

# `VertexGroup<T>`/`VertexGroupUsage<T>`

`VertexGroup<T>` is the new object by which a `DrawNode` "batches" vertices together. It stores metadata about vertices such as invalidation state, but does not store the vertices themselves.

`VertexGroupUsage<T>` denotes a usage of a `VertexGroup<T>`, allowing the user to "add" vertices to the group. This is a stack-only object and thus cannot be stored. This object internally optimises for the case where new vertices don't need to be uploaded.

**Note:** As a result of not storing vertices anymore, `VertexBatch` does not reset back to the start on every `Draw()`. All vertices are added sequentially for the duration of 1 frame.

# Drawing

Combining all the above, a `DrawNode` undergoes the following process to draw to the screen:

```csharp
public class CustomDrawNode : DrawNode
{
    // Create one or more VertexGroups to draw.
    private readonly VertexGroup<TexturedVertex2D> vertices = new VertexGroup<TexturedVertex2D>();

    {
        // Set up blending.
        base.Draw(renderer);

        // Setup shaders, textures, masking, etc.
        shader.Bind();
        texture.Bind();

        // Start a new usage of the VertexGroup.
        using (VertexGroupUsage<TexturedVertex2D> usage = renderer.BeginQuads(this, vertices))
        {
            // When it is known how many vertices are to be drawn ahead of time,
            // preevnting vertices from being constructed at all will yield small performance gains.
            //
            // This can be applied on a case-by-case basis where it is easy to do,
            // Add() will internally prevent doing other more expensive operations.
            if (!usage.TrySkip(4))
                return;

            usage.Add(...);
            usage.Add(...);
            usage.Add(...);
            usage.Add(...);
        }
    }
}

```

- A `DrawNode` **must not** change its vertices between frames unless the `Drawable` has been invalidated.
  - The new vertex values will not be rendered until the `Drawable` is invalidated.
  - Compiling osu!framework in `DEBUG` will throw exceptions when this contract is violated.
- A `VertexGroup<T>` **can not** be used multiple times in a single frame.
- `VertexGroupUsage<T>`s **can not** be nested.

## Custom batches

The `IRenderer` only provides a method to draw quads. **For the time being**, `DrawNode`s can create their own vertex batches if they need to and use them much like `IRenderer.BeginQuads()`:

```csharp
public class CustomDrawNode : DrawNode
{
    private readonly LinearBatch<TexturedVertex2D> batch = new LinearBatch<TexturedVertex2D>(..., PrimitiveType.Triangles);
    private readonly VertexGroup<TexturedVertex2D> vertices = new VertexGroup<TexturedVertex2D>();

    {
        base.Draw(renderer);

        shader.Bind();
        texture.Bind();

        using (var usage = batch.BeginUsage(this, vertices))
        {
            // As above.
        }
    }
}
```

Eventually, vertex batches will be removed from `DrawNode`s altogether and instead go via `IRenderer`.

## Vertex redirection

`TextureGL` and the helper methods of `DrawNode` such as `DrawNode.DrawQuad()` will only output `TexturedVertex2D`-typed vertices. For custom shaders which only accept a subset of properties, `VertexGroup` provides a redirection mechanism:

```csharp
public class CustomDrawNode : DrawNode
{
    // A VertexGroup which converts from TexturedVertex2D to CustomVertex.
    private readonly VertexGroup<TexturedVertex2D, CustomVertex> vertices = new VertexGroup<TexturedVertex2D, CustomVertex>(v => new CustomVertex { ... });

    public override Draw(IRenderer renderer)
    {
        // ...

        // As per normal:
        using (var usage = renderer.BeginQuads(this, vertices))
        {
            // ...
        }
    }
}
```

# Testing

When compiled in DEBUG, osu!framework stores all vertices and throws exceptions upon incorrect usages of `VertexGroup`. These include:
- Vertices changing between frames without an invalidation. Fatal (should be avoided at all cost).
- Vertex equality comparer being broken. Non-fatal (only affects the results of the above check).

# Performance

- Consistently seen a >50% reduction in frametime in complex scenes (`SongSelect`, `SongSelect+BeatmapListing`, `SongSelect+BeatmapListing+Beatmap`).
- I've seen 70MB reduction in total allocs during a play session, as well a reduction in steady-state RAM usage. A VRAM usage counter has been added to the native section in global statistics.
- The only allocation remaining is a static 1024-vertex array used to buffer vertex uploads.


Flutterish:
Will it still be possible to cache data in a VBO to then be manipulated with a projection matrix instead of regenerating it when the ScreenSpaceQuad changes? For this to be possible I usually used a VertexBatch (which was kept alive and therefore its memory not reclaimed) but it sounds like they are going to go away at some point...

smoogipoo:
A `VertexGroupUsage.TryReuse(int count)` method could be exposed which forgoes uploads if only the invalidation state has changed but not if anything else has changed like the target batch/insertion index/etc. Semantics would be the same as the existing `TrySkip(int count)`.

Would that cover your use case?

osu.Framework/Graphics/Drawable.cs
```diff
@@ -2758,6 +2758,11 @@ public enum Invalidation
         /// </summary>
         Parent = 1 << 6,
 
+        /// <summary>
+        /// The <see cref="CompositeDrawable.Masking"/> state has changed.
+        /// </summary>
+        Masking = 1 << 7,
```
bdach, line 2764: Just to confirm intent on this - if I'm understanding right, `Masking` is being split off here as a separate type of invalidation from `DrawInfo` to avoid unnecessary uploads, because masking changes don't/shouldn't affect vertex contents, is that correct?

bdach (commented):
Read through this diff. Probably not the best person to review this, as I have not seen a lot of the draw node internals, so I'm kind of feeling my way around all of this as I'm reviewing. But at a high level this seems pretty good structure-wise, there are just a couple of loose ends that I don't entirely know where they are supposed to lead.

Note that there are several review comments I'm *not* making in order not to muddy the waters here with trivialities. I'm leaving that for follow-up passes once the clean-up effort starts as signposted in the OP.

osu.Framework/Graphics/Batches/Internal/IVertexGroup.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        /// <summary>
+        /// Transforms a given vertex according to this group.
+        /// </summary>
+        /// <param name="vertex">The vertex to transform.</param>
+        /// <typeparam name="TInput">The input vertex type.</typeparam>
+        /// <typeparam name="TOutput">The output vertex type.</typeparam>
+        TOutput Transform<TInput, TOutput>(TInput vertex)
+            where TOutput : struct, IEquatable<TOutput>, IVertex
+            where TInput : struct, IEquatable<TInput>, IVertex;
```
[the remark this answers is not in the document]
smoogipoo: The only current use case is in `ModFlashlight`: https://github.com/ppy/osu/blob/34a60651b2085eca895fad76691c9c4383d91b58/osu.Game/Rulesets/Mods/ModFlashlight.cs#L216-L220.

I'd say it's arguable whether such a function should exist in the first place - the implementation above may be able to use `TexturedVertex2D` instead however I'm not sure about GLSL culling unused attributes and all of that... So it's a sustainable amount of overengineering to avoid further headaches?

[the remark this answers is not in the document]
bdach: Oh, it didn't even occur to me to check game-side usages. Admittedly some of what you've mentioned is flying way over my head (stuff re: attributes getting culled) so right now I'm just going to nod along 😅

osu.Framework/Graphics/Drawable.cs
smoogipoo: It's actually the opposite - changes to masking can affect the vertices drawn by children in the front-to-back pass as a result of insetting the opaque area, so those vertices need to be reuploaded.

This is used in properties which affect the value of `shrunkDrawRectangle`: https://github.com/ppy/osu-framework/blob/85e2bbe557745f573cef6cc2ed9236474819df46/osu.Framework/Graphics/Containers/CompositeDrawable_DrawNode.cs#L81-L89

Which is passed to the masking info:
https://github.com/ppy/osu-framework/blob/85e2bbe557745f573cef6cc2ed9236474819df46/osu.Framework/Graphics/Containers/CompositeDrawable_DrawNode.cs#L91-L97

Which then contributes to the FtB pass in clipping:
https://github.com/ppy/osu-framework/blob/85e2bbe557745f573cef6cc2ed9236474819df46/osu.Framework/Graphics/DrawNode.cs#L220-L224

osu.Framework/Graphics/OpenGL/Buffers/VertexBuffer.cs
```diff
@@ -16,36 +14,52 @@ public abstract class VertexBuffer<T> : IVertexBuffer, IDisposable
...
-        /// <param name="vertexIndex">The index of the vertex.</param>
-        /// <param name="vertex">The vertex.</param>
-        /// <returns>Whether the vertex changed.</returns>
-        public bool SetVertex(int vertexIndex, T vertex)
+        public void EnqueueVertex(int index, T vertex)
         {
-            ref var currentVertex = ref getMemory().Span[vertexIndex];
+            // A new upload must be started if the queue can't hold any more vertices, or if the enqueued index is disjoint from the current to-be-uploaded set.
+            if (uploadLength == upload_queue.Length || (uploadLength > 0 && index > uploadStart + uploadLength))
+                upload();
```
bdach, line 51: This is a part that I wasn't quite getting on the original read-through of this - I had 2 primary question marks here:

* When would the second condition happen (index to enqueue disjoint from existing upload)?
* I assume the limited size of the queue is there in order to limit the number of uploaded verts at a time? Is there any significance to the 1024 number?

osu.Framework/Graphics/Batches/VertexBatch.cs
```diff
@@ -63,71 +65,133 @@ protected void Dispose(bool disposing)
...
         {
-            if (currentVertexIndex == 0)
-                return 0;
+            ulong frameIndex = GLWrapper.CurrentTreeResetId;
 
-            VertexBuffer<T> vertexBuffer = currentVertexBuffer;
-            if (changeBeginIndex >= 0)
-                vertexBuffer.UpdateRange(changeBeginIndex, changeEndIndex);
+            // Disallow reusing the same group multiple times in the same draw frame.
+            if (vertices.Batch == this && frameIndex > 0 && vertices.FrameIndex == frameIndex)
```
bdach, line 166: Here I'm not quite getting why the batch check is there, given the neighbouring comment - seems to me like only checking the frame index would suffice?

osu.Framework/Platform/GameHost.cs
```diff
@@ -56,6 +59,13 @@ public abstract class GameHost : IIpcHost, IDisposable
 
         private InputConfigManager inputConfig { get; set; }
 
+        private static readonly QuadBatch<TexturedVertex2D>[] default_quad_batch =
+        {
+            new QuadBatch<TexturedVertex2D>(100),
+            new QuadBatch<TexturedVertex2D>(100),
+            new QuadBatch<TexturedVertex2D>(100),
+        };
```
bdach, line 67: Was a bit curious of this too - are these per-buffer batches for the triple buffer a thing that was there previously, or a new thing? Do these do anything useful?

bdach (commented):
Upon encouragement on discord here's a few more loose comments. I'm still trying to stick to the "meat" if you will, not about to go pointing out unfilled xmldocs or something (there's some of that in here, but I can go fix that in post if necessary).

osu.Framework/Platform/GameHost.cs
smoogipoo: The triple buffering wasn't there, and it's probably safe to remove from the context that it shouldn't violate the [drawing contracts](#drawing-rules).

From another perspective, though, I think this is beneficial to have in any case as it helps reduce pipeline stalls by not updating the same VBO in sequential frames.

I say "reduce" because the above isn't exactly the purpose of our triple-buffered `DrawNode`s and it is still possible that the same VBO could be updated in sequential frames just by virtue of the draw thread processing much faster than the update thread. It's something we're eventually going to have to address.

Can't hurt to have imo, these batches are very rarely if ever used since a `CompositeDrawable` will almost always exist with a batch of its own.

osu.Framework/Graphics/OpenGL/Buffers/VertexBuffer.cs
smoogipoo: 1.

Consider 3 sprites drawn to the same batch, all sharing the same texture, and suppose the game is in a steady-state in which all vertices have been uploaded.

The batch isn't drawn immediately as soon as a `VertexGroupUsage` ends, but rather whenever an important state changes (e.g. texture/shader/masking, see `GLWrapper.FlushCurrentBatch()`). So these 3 sprites can be assumed to be drawn with a single draw call since they share the same texture.

If then the two outer sprites were to change their vertices in a single frame, then the upload queue will have two distinct sets of vertices to upload ([0...3], [8...11]).

2.

There's no significance to the 1024 number, it's a pretty arbitrary value that assumes there will usually be an intermediate upload as a result of either the "disjoint" condition being hit or a draw call happening.

It's likely that this will be too low when we get to the point of uploading 100000-vertex models, but it'll be a long time before we get to that point and we can adjust it and/or use persistently mapped buffers to avoid this storage altogether.

Flutterish:
> A `VertexGroupUsage.TryReuse(int count)` method could be exposed which forgoes uploads if only the invalidation state has changed but not if anything else has changed like the target batch/insertion index/etc. Semantics would be the same as the existing `TrySkip(int count)`.
> 
> Would that cover your use case?

I'm not sure. The exact use case is more or less this:
```
// tiles is a collection of thousands of tiles which are to be merged into a single batch, 1 quad per
if ( tilesInvalidated ) {
    // this function will create and/or reuse quad vertex buffers to effectively group tiles into batches 
    // that use the same texture (or other uniforms) and generate quads for them
    // these buffers are also shared between all draw nodes (shared data, to avoid duplication because of the triple buffer)
    regenerateBuffers(); 
    tilesInvalidated = false;
}
// this is a quirk with o!f that makes me need to call UpdateRange(0,0) on all vertex buffers 
// because otherwise IVertexBuffer.Free is called and deletes both the cpu and gpu buffer
keepBuffersAlive();
shader.Bind();
// the matrix is generated off of the drawable Matrix3
shader.GetUniform<Matrix4>( "projMatrix" ).UpdateValue( ref matrix );
// this loops through all the texture -> vertex buffer entries and draws their whole content with VertexBuffer.Draw
drawBatches();
shader.Unbind();
```
It's not important for the vertices to be stored on the cpu, but the references to GL buffers and their lengths need to, which also means no other drawable can override these GL buffers.

peppy:
Before I dig too deep, I was looking at the performance changes from this PR and couldn't find a test scene which shows any perceivable difference. Is there a specific visual tests framework-side which can be used (a worst case example I guess?) or would this be feasible to add if not currently present?

peppy:
Also opening Ctrl+F3 seems to crash:

```csharp
[runtime] 2022-04-19 02:28:36 [error]: An unhandled error has occurred.
[runtime] 2022-04-19 02:28:36 [error]: NUnit.Framework.AssertionException
[runtime] 2022-04-19 02:28:36 [error]: at osu.Framework.Logging.ThrowingTraceListener.Fail(String message) in /Users/dean/Projects/osu-framework/osu.Framework/Logging/ThrowingTraceListener.cs:line 23
[runtime] 2022-04-19 02:28:36 [error]: at System.Diagnostics.TraceInternal.Fail(String message)
[runtime] 2022-04-19 02:28:36 [error]: at System.Diagnostics.Trace.Assert(Boolean condition)
[runtime] 2022-04-19 02:28:36 [error]: at osu.Framework.Graphics.OpenGL.GLWrapper.Reset(Vector2 size, Int32 treeIndex) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/OpenGL/GLWrapper.cs:line 154
[runtime] 2022-04-19 02:28:36 [error]: at osu.Framework.Platform.GameHost.DrawFrame() in /Users/dean/Projects/osu-framework/osu.Framework/Platform/GameHost.cs:line 499
[runtime] 2022-04-19 02:28:36 [error]: at osu.Framework.Threading.GameThread.processFrame() in /Users/dean/Projects/osu-framework/osu.Framework/Threading/GameThread.cs:line 450
System.InvalidOperationException: Vertex addition was skipped, but the contained vertex differs.
   at osu.Framework.Graphics.Batches.VertexBatch`1.osu.Framework.Graphics.Batches.Internal.IVertexBatch.EnsureCurrentVertex[TVertex](IVertexGroup vertices, TVertex vertex, String failureMessage) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Batches/VertexBatch.cs:line 110
   at osu.Framework.Graphics.Batches.VertexGroupUsage`1.Add(TInput vertex) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Batches/VertexGroupUsage.cs:line 49
   at osu.Framework.Graphics.OpenGL.Textures.TextureGLSingle.DrawQuad(VertexGroupUsage`1& vertices, Quad vertexQuad, ColourInfo drawColour, Nullable`1 textureRect, Nullable`1 inflationPercentage, Nullable`1 blendRangeOverride, Nullable`1 textureCoords) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/OpenGL/Textures/TextureGLSingle.cs:line 322
   at osu.Framework.Graphics.OpenGL.Textures.TextureGLSub.DrawQuad(VertexGroupUsage`1& vertices, Quad vertexQuad, ColourInfo drawColour, Nullable`1 textureRect, Nullable`1 inflationPercentage, Nullable`1 blendRangeOverride, Nullable`1 textureCoords) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/OpenGL/Textures/TextureGLSub.cs:line 84
   at osu.Framework.Graphics.Textures.Texture.DrawQuad(VertexGroupUsage`1& vertices, Quad vertexQuad, ColourInfo drawColour, Nullable`1 textureRect, Nullable`1 inflationPercentage, Nullable`1 blendRangeOverride, Nullable`1 textureCoords) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Textures/Texture.cs:line 183
   at osu.Framework.Graphics.Visualisation.TextureVisualiser.UsageBackground.UsageBackgroundDrawNode.Blit(VertexGroupUsage`1& usage) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Visualisation/TextureVisualiser.cs:line 238
   at osu.Framework.Graphics.Sprites.SpriteDrawNode.Draw(IRenderer renderer) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Sprites/SpriteDrawNode.cs:line 92
```

smoogipoo:
> Is there a specific visual tests framework-side which can be used

I suppose you could use `TestSceneFrontToBack`, for me it goes from 350 -> 450 fps.

> Also opening Ctrl+F3 seems to crash

Fixed. That `DrawNode`'s pretty dodgy.

frenzibyte (commented):
Have sat through this for a couple of hours to wrap my head around the new structure (and the drawing process in general). And so far this new structure is looking pretty great and a nice first step towards getting rid of `GLWrapper`.

As an aside, I've gone through the visual test browser, and noticed `TestSceneEasingCurves`/`TestSceneKeyBindingsGrid` running quite slow compared to `master`:
 - `master`:

    https://user-images.githubusercontent.com/22781491/164146395-c7895356-4c27-45da-856e-66783979bed2.mp4

 - PR:
   
   https://user-images.githubusercontent.com/22781491/164146611-4a5d8daf-9f6f-4c1e-ab25-d0ca369ed691.mp4

smoogipoo:
> As an aside, I've gone through the visual test browser, and noticed `TestSceneEasingCurves`/`TestSceneKeyBindingsGrid` running quite slow compared to `master`:

Can you profile this one and paste the hottest paths? I can't repro - that test scene goes from 800 -> 1100 fps for me.

osu.Framework/Graphics/Batches/Internal/IVertexGroup.cs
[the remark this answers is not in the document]
smoogipoo: > Do you mean that in regards to it culling unused attributes and expecting a layout of the used ones only (e.g. 2 attributes), while the buffer is populated with the full layout (6 attributes)?

Exactly. I think this actually throws an exception at our side when we go through all vertex attributes and try to bind them, because the unused ones are culled by GLSL:
https://github.com/ppy/osu-framework/blob/85e2bbe557745f573cef6cc2ed9236474819df46/osu.Framework/Graphics/Shaders/Shader.cs#L121-L128

osu.Framework/Graphics/Batches/VertexBatch.cs
```diff
@@ -63,71 +65,133 @@ protected void Dispose(bool disposing)
...
+            GLWrapper.SetActiveBatch(this);
 
-            return count;
+            // Make sure to test in DEBUG when changing the following heuristics.
+            bool uploadRequired =
+                // If this is a new usage or has been moved to a new batch.
+                vertices.Batch != this
+                // Or the DrawNode was newly invalidated.
+                || vertices.InvalidationID != drawNode.InvalidationID
+                // Or another DrawNode was inserted (and drew vertices) before this one.
```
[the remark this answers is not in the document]
smoogipoo: It's referring to vertices being added to the batch. Have adjusted wording.

frenzibyte:
From a 10 seconds record as I was entering the regressed test scene:

CleanShot 2022-04-20 at 07 04 56@2x

smoogipoo:
Sticking to `TestSceneEasingCurves` for now, how many vertices does it upload per frame? GlobalStatistics shows me around 500 vertices per frame:
image

peppy:
Draw thread for me:

JetBrains Rider 2022-04-20 at 04 26 52

frenzibyte:
`VerticesDraw` stays at a constant ~4500, while `VerticesUpl` spikes up from 500 to ~1100 at most.

https://user-images.githubusercontent.com/22781491/164150608-c2dde3d0-3cd4-44b4-aa7e-240ca5447dcb.mp4

osu.Framework/Graphics/OpenGL/Buffers/VertexBuffer.cs
```diff
@@ -16,36 +14,52 @@ public abstract class VertexBuffer<T> : IVertexBuffer, IDisposable
...
+
+        // ReSharper disable once StaticMemberInGenericType
+        private static readonly GlobalStatistic<int> vertex_memory_statistic = GlobalStatistics.Get<int>("Native", $"{nameof(VertexBuffer<T>)}");
+
+        // ReSharper disable once StaticMemberInGenericType
+        private static int uploadStart = int.MaxValue;
 
-        private int vboId = -1;
+        // ReSharper disable once StaticMemberInGenericType
+        private static int uploadLength;
```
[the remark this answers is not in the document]
smoogipoo: I like it. Have applied + applied some refactorings because I didn't like the bouncing back and forth for the upload.

osu.Framework/Graphics/OpenGL/Buffers/VertexBuffer.cs
smoogipoo: I've hopefully commented for (1) a little bit better in 258423a831c8fd6f56c5653004603753e50c4c05.

osu.Framework/Graphics/OpenGL/Buffers/VertexUploadQueue.cs
```diff
@@ -30,38 +41,46 @@ internal static class VertexUploadQueue<T>
         /// <param name="vertex">The vertex to upload.</param>
         public static void Enqueue(VertexBuffer<T> buffer, int index, T vertex)
         {
-            // A new upload must be started if the queue can't hold any more vertices, or if the enqueued index is disjoint from the current to-be-uploaded set.
-            if (uploadLength == upload_queue.Length || (uploadLength > 0 && index > uploadStart + uploadLength))
-                Upload(buffer);
+            // Flush the existing queue if the buffer is filled up.
+            if (uploadLength == upload_queue.Length
+                // Or if two non-contiguous sets of vertices are to be uploaded.
+                || (uploadLength > 0 && index != uploadStart + uploadLength)
```
smoogipoo, line 48: I've changed this from `>` to `!=` just so that it handles vertices being added to the other end of the list gracefully. It still queues up a separate upload so it isn't as efficient as it could be, but I also can't think of a case where you'd willingly add vertices in reverse order.

smoogipoo:
> spikes up from 500 to ~1100 at most (not sure if considerable).

There's something more to this that I'll have to investigate. 1000 vertices is nothing at all since the game regularly uploads 10x more than that.

smoogipoo:
From an initial investigation it seems macOS, and especially Apple Silicon, does _really_ not like buffer orphaning after draw calls. It's something I didn't consider but should've - the standard case of you shouldn't modify a VBO that's already in flight.

I'll need to think more about the structure to figure out how to fix this.

frenzibyte:
FWIW, I've managed to get Veldrid running on this branch https://github.com/ppy/osu-framework/compare/master...frenzibyte:veldrid-plus-vbo-invalidation (requires a local veld rid checkout at [this branch](https://github.com/mellinoe/veldrid/compare/master...frenzibyte:features-and-fixes)), in case you are interested.

smoogipoo:
@frenzibyte Regarding https://github.com/ppy/osu-framework/pull/5109#pullrequestreview-946404024, can you check again and see if my most recent commit fixes it?

It's not perfect, lacking tests, and a huge edge case isn't handled correctly, but should be enough to check performance for the time being.

smoogipoo:
Need some more time to dwell on this one. The problem right now is states that are cause a flush changing during a group's usage (textures, shaders, etc), whereas most of this was coded with the slight hint that that shouldn't be allowed.

frenzibyte:
> @frenzibyte Regarding [#5109 (review)](https://github.com/ppy/osu-framework/pull/5109#pullrequestreview-946404024), can you check again and see if my most recent commit fixes it?
> 
> It's not perfect, lacking tests, and a huge edge case isn't handled correctly, but should be enough to check performance for the time being.

It has brought performance back on par with master for `TestSceneEasingCurves`, but is still degraded a bit in `TestSceneKeyBindingsGrid` (PR shows 150fps while it runs at 250fps on master), for whatever that's worth.

peppy:
I want to do another review pass on this PR, but first confirming that it's in a final state @smoogipoo ?

Do you have anything to add regarding the performance regression in `TestSceneKeyBindingsGrid` mentioned above? For me on release 234 (master) vs 280 (this branch), on debug 130 (master) vs 104 (this branch). Which seems fine. Was @frenzibyte's above comment in debug mode?

smoogipoo:
> Was @frenzibyte's above comment in debug mode?

Apparently wasn't according to https://discord.com/channels/188630481301012481/589331078574112768/987972289716965427, however the takeaway is that behaviour is different on M1 (I believe) vs desktop. M1 _really_ doesn't like VBOs being reused in one draw frame whereas Radeon/NV seems to not mind it _as much_.

> Do you have anything to add regarding the performance regression in TestSceneKeyBindingsGrid mentioned above?

The above being said, I don't know why the difference in performance still remains to some degree.

> but first confirming that it's in a final state

It's close enough, but I still have to refactor and re-PR because this one's kinda messy. The general idea is there, though.

peppy:
Best to block pending that so no one gets too deep in reviewing here then?

smoogipoo:
This PR is blocked and there's a message in the PR description about it, is that enough?

peppy:
Ah, missed that. Is fine.

PR #5110, opened 2022-04-18
Fix accidental API breakage of `Game.OnExiting()` return value

bdach:
As correctly pointed out by @Joehuu, all of the recent troubles with the exit flow breaking originated in #5097, namely the [change to `GameHost` therein](https://github.com/ppy/osu-framework/pull/5097/files#diff-c694d93cea53f76879738dce61918a64bc3110e7df817250864f1f4754607512). This led to *unintentional API breakage in the latest framework release* and as such I think it makes sense to just revert it. The explanation why it is a regression is very simple but also stems from some very confusing code:

| `response` | `response ?? false` | `response == false` |
| :-: | :-: | :-: |
| `true` | `true` | `false` |
| `false` | `false` | `true` |

I'm not including the `null` case, as it was never reachable due to how the surrounding code is structured - [it is guaranteed that `response` is not null](https://github.com/ppy/osu-framework/blob/439aeddd031c79fe96ae6fe428082d0c833d9146/osu.Framework/Platform/GameHost.cs#L407-L414) when the aforementioned guard is reached.

#5101 partially papered over this, by updating xmldoc to the post-regression state and inverting one `OnExiting()` return value. This continued in #5104 wherein a second `OnExiting()` return value was inverted and a second xmldoc using the post-regression state was added. Both changes are reverted here, and xmldoc is updated to match.

This change supersedes / closes https://github.com/ppy/osu/pull/17870.

Test coverage for return value correctness included.

### DIFF
--- a/osu.Framework.Tests/Platform/GameExitTest.cs
+++ b/osu.Framework.Tests/Platform/GameExitTest.cs
@@ -40,6 +40,8 @@ namespace osu.Framework.Tests.Platform
             game.BlockExit.Value = true;
             // `RequestExit()` should return true.
             Assert.That(host.RequestExit(), Is.True);
+            // game's last exit result should match.
+            Assert.That(game.LastExitResult, Is.True);
             // exit should be blocked.
             Assert.That(() => host.ExecutionState, Is.EqualTo(ExecutionState.Running).After(timeout));
             Assert.That(task.IsCompleted, Is.False);
@@ -48,6 +50,8 @@ namespace osu.Framework.Tests.Platform
             game.BlockExit.Value = false;
             // `RequestExit()` should not be blocked and return false.
             Assert.That(host.RequestExit(), Is.False);
+            // game's last exit result should match.
+            Assert.That(game.LastExitResult, Is.False);
             // finally, the game should exit.
             Assert.That(() => host.ExecutionState, Is.EqualTo(ExecutionState.Stopped).After(timeout));
             task.WaitSafely();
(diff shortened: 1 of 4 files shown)


osu.Framework.Tests/Platform/GameExitTest.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            var task = Task.Factory.StartNew(() =>
+            {
+                using (host = new ManualExitHeadlessGameHost())
+                {
+                    game = new TestTestGame();
+                    gameCreated.Set();
+                    host.Run(game);
+                }
+            }, TaskCreationOptions.LongRunning);
```
frenzibyte, line 34: nit: could use the pre-existing `BackgroundGameHeadlessGameHost`, but will require a bit more of changes that I think either-or is fine.

frenzibyte (approved):

frenzibyte:
Actually, the test does not look happy in https://github.com/ppy/osu-framework/pull/5110/checks?check_run_id=6066491542

osu.Framework.Tests/Platform/GameExitTest.cs
bdach: I prefer local implementations in tests to modifying shared classes to fit multiple purposes.

bdach:
I saw that failure but have no idea why it fails when there are several other tests written in very similar ways. I would be much surprised if it was by anything done here, and if it turns out that it's starting to fail, I'd rather delete the test than put time into debugging whatever eldritch thing did that.

I'll try stressing it but with that failing stack trace I'm not expecting to get much out of that if anything at all.

bdach:
Repeated the test 100 times locally and got nothing to show for it. Not even sure how the performance monitor is collecting anything given that nothing looks to be touching it, aside from `TestSceneGlobalStatisticsDisplay` which already has a teardown step to hide the statistics and therefore disables performance collection. We're not running tests concurrently are we? That's about the only remaining feasible theory I have.

smoogipoo:
I'm guessing that it failed because `GameHostSuspendTest` failed on an assertion and kept another `GameHost` running on a background thread.

smoogipoo (approved):
Not sure how I derped so hard :/

peppy (approved):

PR #5111, opened 2022-04-21
Fix `TestExitBlocking` taking 10 seconds to run

peppy:
Initially I was just looking to fix the weird conditional (3b8c1a36e), but noticed the test is quite slow to execute.

I think the original code was a bit too safe. Internally the `host.RequestExit()` call is blocking until the exit has been handled, but I can understand that we may be trying to black box test here to avoid any uncertainty.

If this solution is seen as unsatisfactory then we can count update frames or something. I'd just like to avoid a single test taking 10 seconds to run while spinning.


smoogipoo (approved):
Seems reasonable

PR #5118, opened 2022-04-22
Allow layered `ScrollContainer`s to coexist on input handling

### LINKED ISSUE #3576: ScrollContainers of different directions should be able to coexist
Two ScrollContainers (nested or overlapping) should be able to coexist and only manage drags in their respective direction. Currently the first to handle input will do so regardless of the drag direction.

frenzibyte:
- Closes #3576

This changes drag event handling and scroll event handling in `ScrollContainer`s to be based on their `ScrollDirection`, to allow for layered scroll containers to coexist.

This however becomes a nuisance on two parts:
 - https://github.com/ppy/osu-framework/commit/20c1d72502c4f85d5e23def9a28aa02a5725dbb6: In order for scroll containers to coexist, the innermost must not `OnMouseDown` to let the parenting scroll containers enter the mouse-down/drag input queue, for receiving drag events.

   I'm not entirely sure if there's a reason for that in the first place, checking with git blame threw me back to 4 years ago (https://github.com/ppy/osu-framework/pull/316).
 - https://github.com/ppy/osu-framework/commit/7bff9b96bebc982b5dfabf75cd714b41237bf25a: When a scroll container is horizontal, scrolling vertically will also be considered as horizontal, therefore this coexistence change cannot apply when the inner scroll container is horizontal ([as done in test](https://github.com/ppy/osu-framework/blob/e81e2b80acf9c2edfaf532620b36fa8160b17f18/osu.Framework.Tests/Visual/Containers/TestSceneScrollContainer.cs#L438-L445)).

    In addition, it also results in this kinda weird UX (inner = horizontal, outer = vertical):

    https://user-images.githubusercontent.com/22781491/164593522-314825a1-8388-47c6-b9c2-090fe519fd3e.mp4

### DIFF
--- a/osu.Framework/Graphics/Containers/ScrollContainer.cs
+++ b/osu.Framework/Graphics/Containers/ScrollContainer.cs
@@ -243,6 +243,9 @@ namespace osu.Framework.Graphics.Containers
             if (IsDragging || e.Button != MouseButton.Left || Content.AliveInternalChildren.Count == 0)
                 return false;
 
+            if (Math.Abs(e.Delta.X) > Math.Abs(e.Delta.Y) != (ScrollDirection == Direction.Horizontal))
+                return false;
+
             lastDragTime = Time.Current;
             averageDragDelta = averageDragTime = 0;
 
--- a/osu.Framework/Graphics/Containers/ScrollContainer.cs
+++ b/osu.Framework/Graphics/Containers/ScrollContainer.cs
@@ -277,6 +277,9 @@ namespace osu.Framework.Graphics.Containers
 
         protected override bool OnMouseDown(MouseDownEvent e)
         {
+            if (IsDragging || e.Button != MouseButton.Left)
+                return false;
+
             // Continue from where we currently are scrolled to.
             Target = Current;
             return false;
--- a/osu.Framework/Graphics/Containers/ScrollContainer.cs
+++ b/osu.Framework/Graphics/Containers/ScrollContainer.cs
@@ -243,7 +243,9 @@ namespace osu.Framework.Graphics.Containers
             if (IsDragging || e.Button != MouseButton.Left || Content.AliveInternalChildren.Count == 0)
                 return false;
 
-            if (Math.Abs(e.Delta.X) > Math.Abs(e.Delta.Y) != (ScrollDirection == Direction.Horizontal))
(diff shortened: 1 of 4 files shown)


osu.Framework/Graphics/Containers/ScrollContainer.cs
```diff
@@ -274,12 +277,9 @@ protected override bool OnKeyDown(KeyDownEvent e)
 
         protected override bool OnMouseDown(MouseDownEvent e)
         {
-            if (IsDragging || e.Button != MouseButton.Left) return false;
```
peppy, line 277: Are you sure about the removal of the early return before setting `Target` below? It seems like that should probably not be updated in these cases. I haven't tested yet, but starting a drag with left mouse then right clicking could cause weirdness from `Target` re-sync.

osu.Framework/Graphics/Containers/ScrollContainer.cs
```diff
@@ -274,12 +277,9 @@ protected override bool OnKeyDown(KeyDownEvent e)
...
-
             // Continue from where we currently are scrolled to.
             Target = Current;
-
-            return true;
```
peppy, line 282: Will need to check osu!-side to make sure nothing was relying on this behaviour, but I'm not strongly against this change if it works everywhere we expect it to.

peppy (commented):
Haven't tested yet, just reading code changes

frenzibyte:
Unsure how to deal with this test:

CleanShot 2022-04-22 at 07 24 20@2x

Feels like an arbitrary choice to drag perfectly diagonally and check whether the scroll container would block the click or not, since it's going to depend on which is checked first in the early-guard.

peppy:
The diagonal in that test does look arbitrary. Changing to a specific direction is probably fine.

osu.Framework/Graphics/Containers/ScrollContainer.cs
frenzibyte: I would be interested if there's any scroll container overlayed on top of another component at all (unless wrapped in an `OverlayContainer` of course).

peppy:
Made some minor refactors, but looks and feels good to me. Will wait for a second opinion for merge.

peppy (approved):

smoogipoo (approved):

PR #5121, opened 2022-04-22
Enforce code style analysis shipped with .NET SDK

huoyaoyuan:
The IDExxxx analyzers were only shipped with Visual Studio. `dotnet format` gives some access to them.

From .NET 5, most of them are shipped together with .NET SDK, and can be included with `EnforceCodeStyleInBuild` property. They will now be invoked in any `dotnet build` command, available to other IDE and CI. `dotnet format` can be replaced now.

In this PR, I'm focusing on pure stylish analyzers to simplify reviewing. If you prefer I can also include language feature adoption ones.

#### Performance

This should theoretically add overhead to every build, but I can't see a measurable difference with `Measure-Command { dotnet build /p:EnforceCodeStyleInBuild=true }`. Turning on them during normal build could inform non Visual Studio users.
There are several analyzers not included in the SDK, which are claimed with performance issues. I suppose the already shipped ones are good in performance.

#### CI

This will now be enforced during Build&Test with WarningAsError enabled. A violation would cause a red in test legs. This is the same of CAxxxx analyzers.

Another options is to disable analyzers in Test leg and enforce them in Code Quality leg to improve clarity. What do you think?

### DIFF
--- a/osu.Framework.Tests/Polygons/LineIntersections.cs
+++ b/osu.Framework.Tests/Polygons/LineIntersections.cs
@@ -14,7 +14,9 @@ namespace osu.Framework.Tests.Polygons
         private static readonly Vector2 up_1 = new Vector2(0, 1);
         private static readonly Vector2 up_2 = new Vector2(0, 2);
         private static readonly Vector2 down_1 = new Vector2(0, -1);
+
         // private static readonly Vector2 down_2 = new Vector2(0, -2); // unused
+
         private static readonly Vector2 left_1 = new Vector2(-1, 0);
         private static readonly Vector2 left_2 = new Vector2(-2, 0);
         private static readonly Vector2 right_1 = new Vector2(1, 0);
--- a/osu.Framework.Tests/Polygons/LineIntersections.cs
+++ b/osu.Framework.Tests/Polygons/LineIntersections.cs
@@ -14,7 +14,7 @@ namespace osu.Framework.Tests.Polygons
         private static readonly Vector2 up_1 = new Vector2(0, 1);
         private static readonly Vector2 up_2 = new Vector2(0, 2);
         private static readonly Vector2 down_1 = new Vector2(0, -1);
-        private static readonly Vector2 down_2 = new Vector2(0, -2);
+        // private static readonly Vector2 down_2 = new Vector2(0, -2); // unused
         private static readonly Vector2 left_1 = new Vector2(-1, 0);
         private static readonly Vector2 left_2 = new Vector2(-2, 0);
         private static readonly Vector2 right_1 = new Vector2(1, 0);
--- a/osu.Framework.Tests/Polygons/LineIntersections.cs
+++ b/osu.Framework.Tests/Polygons/LineIntersections.cs
@@ -14,9 +14,6 @@ namespace osu.Framework.Tests.Polygons
         private static readonly Vector2 up_1 = new Vector2(0, 1);
         private static readonly Vector2 up_2 = new Vector2(0, 2);
         private static readonly Vector2 down_1 = new Vector2(0, -1);
-
-        // private static readonly Vector2 down_2 = new Vector2(0, -2); // unused
-
         private static readonly Vector2 left_1 = new Vector2(-1, 0);
         private static readonly Vector2 left_2 = new Vector2(-2, 0);
         private static readonly Vector2 right_1 = new Vector2(1, 0);
--- a/osu.Framework/Audio/AudioManager.cs
+++ b/osu.Framework/Audio/AudioManager.cs
@@ -170,8 +170,6 @@ namespace osu.Framework.Audio
             scheduler.Add(() =>
             {
                 // sync audioDevices every 1000ms
-
-#pragma warning disable IDE0055 // VS and Rider formatting conflict
                 new Thread(() =>
                 {
                     while (!token.IsCancellationRequested)
@@ -185,8 +183,10 @@ namespace osu.Framework.Audio
                         {
                         }
                     }
-                }) { IsBackground = true }.Start();
-#pragma warning restore IDE0055
+                })
+                {
+                    IsBackground = true
+                }.Start();
             });
         }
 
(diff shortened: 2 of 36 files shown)


osu.Framework/Localisation/LocalisableStringEqualityComparer.cs
```diff
@@ -12,7 +12,9 @@ namespace osu.Framework.Localisation
     public class LocalisableStringEqualityComparer : IEqualityComparer<LocalisableString>
     {
         // ReSharper disable once InconsistentNaming (follows EqualityComparer<T>.Default)
+#pragma warning disable IDE1006 // Naming style
```
peppy, line 15: What's the deal with this one? It the formatting rule for public-static-readonly not available for customisation?

huoyaoyuan: For static readonly fields we are using ALL_UPPER. PascalCase is used only for these ones for consistency with .NET BCL. Currently there is no way to customise by particular name/member.

osu.Framework.Tests/Polygons/LineIntersections.cs
```diff
@@ -14,7 +14,9 @@ public class LineIntersections
         private static readonly Vector2 up_1 = new Vector2(0, 1);
         private static readonly Vector2 up_2 = new Vector2(0, 2);
         private static readonly Vector2 down_1 = new Vector2(0, -1);
-        private static readonly Vector2 down_2 = new Vector2(0, -2);
+
+        // private static readonly Vector2 down_2 = new Vector2(0, -2); // unused
```
bdach, line 18: Why not just remove?

osu.Framework/Audio/AudioManager.cs
```diff
@@ -171,19 +171,20 @@ public AudioManager(AudioThread audioThread, ResourceStore<byte[]> trackStore, R
...
+                                Thread.Sleep(1000);
+                            }
+                            catch
+                            {
+                            }
                         }
-                    }
-                }) { IsBackground = true }.Start();
+                    })
+                { IsBackground = true }.Start();
```
bdach, line 187: What's happened with the indentation here? Old version looks more correct to me. Looks like possibly conflicting inspections between dotnet analyzers and inspectcode.,.?

osu.Framework/Graphics/Video/FFmpegFuncs.cs
```diff
@@ -7,6 +7,8 @@
 namespace osu.Framework.Graphics.Video
 {
     // ReSharper disable InconsistentNaming
+#pragma warning disable IDE1006 // Naming style
```
bdach, line 10: Can both of these be pulled out before the namespace declaration or something? The mixed indentation between the resharper disable and the dotnet disable is really jarring to me, and I'm guessing you can't indent `#pragma` directives.

osu.Framework/Platform/MacOS/Native/NSDictionary.cs
```diff
@@ -9,7 +9,9 @@ namespace osu.Framework.Platform.MacOS.Native
     {
         internal IntPtr Handle { get; }
 
+#pragma warning disable IDE0052 // Unread private member
         private static IntPtr classPointer = Class.Get("NSDictionary");
+#pragma warning restore IDE0052 //
```
frenzibyte, line 14: Can just remove.

osu.Framework.Tests/Polygons/LineIntersections.cs
huoyaoyuan: To keep it symmetric. Well the names are very straightforward, it's OK to remove.

osu.Framework/Audio/AudioManager.cs
huoyaoyuan: Yes, IDE0059 forces a new line here, and inspectcode forces indentation when the new line is added.

huoyaoyuan: Well...inspectcode forces indentation for the property initializer, while IDE0059 forces new line and no indentation. I'd want to refactor this to use `Task`.

osu.Framework/Audio/AudioManager.cs
```diff
@@ -170,21 +171,20 @@ public AudioManager(AudioThread audioThread, ResourceStore<byte[]> trackStore, R
             scheduler.Add(() =>
             {
                 // sync audioDevices every 1000ms
-                new Thread(() =>
+                Task.Run(async () =>
```
bdach, line 174: Dunno about this change, I would consider using the `TaskCreationOptions.LongRunning` flag here.

huoyaoyuan: As my understanding, `LongRunning` means occupying a thread for a long time without `await`. This task would `await` very frequently and allow task scheduler to reschedule every time.

frenzibyte: Would've gone with factoring this out to a named method rather than bother changing the logic in this PR, but sure I guess? not sure if it still shouldn't be `LongRunning` though, since that task will continue running for the entire game session.

huoyaoyuan: I'm preparing for a similar refactor anyway. `LongRunning` is just a hint, not a strong requirement.

peppy: shouldn't be in this PR.

huoyaoyuan: Simply disable the warning for now, since it's going to be refactored anyway.

osu.Framework/Localisation/LocalisableStringEqualityComparer.cs
peppy: I'd say this should be disabled at a project-wide level until it can be fixed. We also get this inspection from r# and it works better.

peppy: Actually @frenzibyte pointed out we can use auto properties here to get equivalent functionality and pass naming checks. I'll commit the changes.

peppy:
I've resolved all my issues with this branch. Will needs others to check again.

peppy (approved):

bdach (approved):

osu.Framework/Graphics/Video/FFmpegFuncs.cs
```diff
@@ -4,9 +4,11 @@
 using System.Runtime.InteropServices;
 using FFmpeg.AutoGen;
 
+// ReSharper disable InconsistentNaming
+#pragma warning disable IDE1006 // Naming style
```
frenzibyte, line 8: Not really sure about having both of these inspections active at the same time. It's also weird how one functions differently from the other:

https://user-images.githubusercontent.com/22781491/165195821-51466d80-20a1-4d82-8026-8a8a56761bc0.mp4

But I guess that can be left as a follow-up.

frenzibyte (approved):

smoogipoo (approved):

PR #5122, opened 2022-04-23
Add support for configurable deadzone for joystick/gamepad

maromalo:
Makes `deadzone_threshold` bindable to allow it to be changed. Copied from [MouseHandler.Sensitivity](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Input/Handlers/Mouse/MouseHandler.cs#L29).

### DIFF
--- a/osu.Framework/Input/Handlers/Joystick/JoystickHandler.cs
+++ b/osu.Framework/Input/Handlers/Joystick/JoystickHandler.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using osu.Framework.Bindables;
 using osu.Framework.Input.StateChanges;
 using osu.Framework.Platform;
 using osu.Framework.Statistics;
@@ -10,7 +11,12 @@ namespace osu.Framework.Input.Handlers.Joystick
 {
     public class JoystickHandler : InputHandler
     {
-        private const float deadzone_threshold = 0.075f;
+        public BindableDouble deadzone_threshold { get; } = new BindableDouble(0.075)
+        {
+            MinValue = 0,
+            MaxValue = 1,
+            Precision = 0.005,
+        };
 
         private readonly JoystickButton[] axisDirectionButtons = new JoystickButton[(int)JoystickAxisSource.AxisCount];
 
@@ -84,15 +90,15 @@ namespace osu.Framework.Input.Handlers.Joystick
             enqueueJoystickEvent(new JoystickAxisInput(new JoystickAxis(axis.Source, value)));
         }
 
-        private static float rescaleByDeadzone(float axisValue)
+        private float rescaleByDeadzone(float axisValue)
         {
             float absoluteValue = Math.Abs(axisValue);
 
-            if (absoluteValue < deadzone_threshold)
+            if (absoluteValue < deadzone_threshold.Value)
                 return 0;
 
             // rescale the given axis value such that the edge of the deadzone is considered the "new zero".
-            float absoluteRescaled = (absoluteValue - deadzone_threshold) / (1f - deadzone_threshold);
+            float absoluteRescaled = (absoluteValue - (float)deadzone_threshold.Value) / (1f - (float)deadzone_threshold.Value);
             return Math.Sign(axisValue) * absoluteRescaled;
         }
 


osu.Framework/Input/Handlers/Joystick/JoystickHandler.cs
```diff
@@ -84,15 +90,15 @@ private void enqueueJoystickAxisChanged(JoystickAxis axis)
...
         {
             float absoluteValue = Math.Abs(axisValue);
 
-            if (absoluteValue < deadzone_threshold)
+            if (absoluteValue < deadzone_threshold.Value)
                 return 0;
 
             // rescale the given axis value such that the edge of the deadzone is considered the "new zero".
-            float absoluteRescaled = (absoluteValue - deadzone_threshold) / (1f - deadzone_threshold);
+            float absoluteRescaled = (absoluteValue - (float)deadzone_threshold.Value) / (1f - (float)deadzone_threshold.Value);
```
frenzibyte, line 101: Might as well change `deadzone_threshold` to `BindableFloat` and avoid this extra changes.

osu.Framework/Input/Handlers/Joystick/JoystickHandler.cs
```diff
@@ -10,7 +11,12 @@ namespace osu.Framework.Input.Handlers.Joystick
 {
     public class JoystickHandler : InputHandler
     {
-        private const float deadzone_threshold = 0.075f;
+        public BindableDouble deadzone_threshold { get; } = new BindableDouble(0.075)
```
frenzibyte, line 14: Should be `DeadzoneThreshold` to match naming convention.

frenzibyte (changes requested):
Thanks for the contribution.

Does this come with an osu!-side implementation for user access to this setting?

maromalo:
> Thanks for the contribution.
> 
> Does this come with an osu!-side implementation for user access to this setting?

Yes, figured I should pull the framework-side first.

osu.Framework/Input/Handlers/Joystick/JoystickHandler.cs
maromalo: Fixed.

osu.Framework/Input/Handlers/Joystick/JoystickHandler.cs
maromalo: Fixed.

osu.Framework/Input/Handlers/Joystick/JoystickHandler.cs
```diff
@@ -10,7 +11,12 @@ namespace osu.Framework.Input.Handlers.Joystick
 {
     public class JoystickHandler : InputHandler
     {
-        private const float deadzone_threshold = 0.075f;
+        public BindableFloat DeadzoneThreshold { get; } = new BindableFloat(0.075f)
+        {
+            MinValue = 0,
+            MaxValue = 1,
```
peppy, line 17: I'm not sure the max value should be 1 here (it would make the game pad inoperable).. curious what kind of range other games expose. Maybe 0-0.9 or something similar?

frenzibyte: I would've suggested 0.75 as a sane highest deadzone value. Don't have any games to compare with, though.

maromalo: I'm thinking about maxing it at 0.995, at least for there to be a choice, otherwise 0.95 or 0.9 feel reasonable.

peppy: Let's go with 0.95 then. We can revisit if users request higher, but I think it should be good enough.

osu.Framework/Input/Handlers/Joystick/JoystickHandler.cs
```diff
@@ -10,7 +11,12 @@ namespace osu.Framework.Input.Handlers.Joystick
 {
     public class JoystickHandler : InputHandler
     {
-        private const float deadzone_threshold = 0.075f;
+        public BindableFloat DeadzoneThreshold { get; } = new BindableFloat(0.075f)
```
peppy, line 14:
```suggestion
        public BindableFloat DeadzoneThreshold { get; } = new BindableFloat(0.1f)
```

Sorry to request another change, but while we're here I think the default value could do with a slight bump.

peppy (changes requested):

frenzibyte (approved):

PR #5123, opened 2022-04-23
Implement CSS-style font size algorithm

frenzibyte:
- Closes https://github.com/ppy/osu-framework/issues/3271 
- Resolves the font size discrepancy in https://github.com/ppy/osu/issues/17940

Adds support for CSS-style font sizing, allowing for our sprite text sizes to match 1:1 with CSS. 

As a starting point, the option is disabled by default to avoid breaking every existing sprite text in all games, but as we move forward (with migrating to new designs in osu!), we'll want to make the option enabled by default, and potentially remove the option (keeping it only based on whether the font's metrics are available).

## `FontMetrics`

Since the font binaries don't store metadata about the metrics (ascent, descent, em size), it is required for osu! and all consumers of the framework to manually extract them from the font and specify them in `AddFont`, [similar to how it's done here]():
```diff
@@ -161,19 +162,24 @@ private void load(FrameworkConfigManager config)
             // note that currently this means there could be two async font load operations.
             Fonts.AddStore(localFonts = new FontStore(useAtlas: false));

+            // Roboto and Font Awesome have different metrics for Windows and macOS.
+            // The ones used for Windows are used here for the time being.
+            var robotoMetrics = new FontMetrics(ascent: 1946, descent: 512, emSize: 2048);
+            var fontAwesomeMetrics = new FontMetrics(ascent: 460, descent: 84, emSize: 512);
+
             // Roboto (FrameworkFont.Regular)
-            addFont(localFonts, Resources, @"Fonts/Roboto/Roboto-Regular");
-            addFont(localFonts, Resources, @"Fonts/Roboto/Roboto-RegularItalic");
-            addFont(localFonts, Resources, @"Fonts/Roboto/Roboto-Bold");
-            addFont(localFonts, Resources, @"Fonts/Roboto/Roboto-BoldItalic");
+            addFont(localFonts, Resources, @"Fonts/Roboto/Roboto-Regular", robotoMetrics);
+            addFont(localFonts, Resources, @"Fonts/Roboto/Roboto-RegularItalic", robotoMetrics);
+            addFont(localFonts, Resources, @"Fonts/Roboto/Roboto-Bold", robotoMetrics);
+            addFont(localFonts, Resources, @"Fonts/Roboto/Roboto-BoldItalic", robotoMetrics);

             // RobotoCondensed (FrameworkFont.Condensed)
-            addFont(localFonts, Resources, @"Fonts/RobotoCondensed/RobotoCondensed-Regular");
-            addFont(localFonts, Resources, @"Fonts/RobotoCondensed/RobotoCondensed-Bold");
+            addFont(localFonts, Resources, @"Fonts/RobotoCondensed/RobotoCondensed-Regular", robotoMetrics);
+            addFont(localFonts, Resources, @"Fonts/RobotoCondensed/RobotoCondensed-Bold", robotoMetrics);

-            addFont(Fonts, Resources, @"Fonts/FontAwesome5/FontAwesome-Solid");
-            addFont(Fonts, Resources, @"Fonts/FontAwesome5/FontAwesome-Regular");
-            addFont(Fonts, Resources, @"Fonts/FontAwesome5/FontAwesome-Brands");
+            addFont(Fonts, Resources, @"Fonts/FontAwesome5/FontAwesome-Solid", fontAwesomeMetrics);
+            addFont(Fonts, Resources, @"Fonts/FontAwesome5/FontAwesome-Regular", fontAwesomeMetrics);
+            addFont(Fonts, Resources, @"Fonts/FontAwesome5/FontAwesome-Brands", fontAwesomeMetrics);

             dependencies.Cache(Fonts);
```
 - Note that different platforms use different metric tables, and I'm not entirely sure if we should use each metric based on the platform, as it turns out [there can be differences](https://github.com/googlefonts/roboto/issues/267). Therefore I continued with the one [used by Windows](https://docs.microsoft.com/en-us/typography/opentype/spec/os2#stypoascender).

   This is still completely up to the consumer whether they want it to be platform-specific or not, but just mentioning this in regards to osu!.

 - I've kept the font metrics argument in `AddFont` as optional, for special fonts which don't have the concept of "font metrics" ("8-bit" fonts and the like).

I will write up more about this in the fonts wiki if the `FontMetrics` concept is agreed on, mainly about how to extract them and which values should be used, etc.

## `FontUsage.CssScaling`

Going forward, osu! and all consumers should always append `css: true` to their font usages, potentially using a different static and prefixing the old non-CSS ones with `Old`:
```diff
diff --git a/osu.Game/Graphics/OsuFont.cs b/osu.Game/Graphics/OsuFont.cs
index edb484021c..62eb7492f8 100644
--- a/osu.Game/Graphics/OsuFont.cs
+++ b/osu.Game/Graphics/OsuFont.cs
@@ -19,7 +19,9 @@ public static class OsuFont

         public static FontUsage Numeric => GetFont(Typeface.Venera, weight: FontWeight.Bold);

-        public static FontUsage Torus => GetFont(Typeface.Torus, weight: FontWeight.Regular);
+        public static FontUsage OldTorus => GetFont(Typeface.Torus, weight: FontWeight.Regular);
+
+        public static FontUsage Torus => GetFont(Typeface.Torus, weight: FontWeight.Regular, css: true);

         public static FontUsage TorusAlternate => GetFont(Typeface.TorusAlternate, weight: FontWeight.Regular);
```

It could also go the other way around, by keeping the old one as-is, and defining a `New` font with `css: true` applied instead.

---

I've added behavioural test coverage, and also added a visual illustration in the sprite text test scene, by comparing against a picture of a CSS text element with an equal font size, in a properly aligned container:

CleanShot 2022-04-23 at 06 48 29@2x

Have also added sprite texts with CSS-style enabled, alongside the pre-existing non-CSS ones:

CleanShot 2022-04-23 at 06 50 43@2x


bdach:
> and addition of noto font for testing purposes

Why was this done? I think it would have sufficed to test Roboto.

frenzibyte:
Just for the sake of adding more into the visualisation test, and also to confirm that the framework’s handling of multiple fonts also matches CSS.

osu.Framework/Graphics/Sprites/FontUsage.cs
```diff
@@ -65,13 +75,18 @@ namespace osu.Framework.Graphics.Sprites
         /// <param name="weight">The font weight.</param>
         /// <param name="italics">Whether the font is italic.</param>
         /// <param name="fixedWidth">Whether all characters should be spaced the same distance apart.</param>
-        public FontUsage([CanBeNull] string family = null, float size = default_text_size, [CanBeNull] string weight = null, bool italics = false, bool fixedWidth = false)
+        /// <param name="css">
+        /// Whether the text glyphs should scale according to their respective <see cref="IGlyphStore.Metrics"/>, matching CSS.
+        /// It is recommended to enable this for better alignment with other fonts.
+        /// </param>
+        public FontUsage([CanBeNull] string family = null, float size = default_text_size, [CanBeNull] string weight = null, bool italics = false, bool fixedWidth = false, bool css = false)
```
bdach, line 82: Without reading into this diff too much, I would have much preferred a syntax of

```csharp
new FontUsage(cssSize: 16)
```

rather than

```csharp
new FontUsage(size: 16, css: true)
```

Is changing to the first feasible at all? I know that it may be a bit annoying to handle, but it would be much nicer to use imo. Interested to know if others feel the same.

frenzibyte: Hmm, that seems fine, but would also mean forcing every usage to replace `size` with `cssSize`, as I intended that boolean to be set at a higher level instead (as a static font member of its own, and prefixing the current one with `Old` and pretty much never use it again). Similar to the `FontUsage` diff in OP.

bdach: I don't really agree with adding anything with "old" prefixes, because that implies you'd have to update all game side references of `Torus` to `OldTorus`? Is there some reason that we'd want that?

frenzibyte: Just so that we don’t have to play around with the font sizes we currently use, and not start to confuse between components which use CSS and ones which don’t, and also not have to write ”cssSize” in every sprite text declaration.

Though I may be the only one with this opinion, curious to hear others.

osu.Framework/Text/TextBuilder.cs
```diff
@@ -62,24 +70,26 @@ public float LineBaseHeight
         /// </summary>
         /// <param name="store">The store from which glyphs are to be retrieved from.</param>
         /// <param name="font">The font to use for glyph lookups from <paramref name="store"/>.</param>
-        /// <param name="useFontSizeAsHeight">True to use the provided <see cref="font"/> size as the height for each line. False if the height of each individual glyph should be used.</param>
+        /// <param name="useFullGlyphHeight">True to use <see cref="TextBuilderGlyph.Size"/> (full glyph size) as the height for each line. False if the height of each individual glyph should be used.</param>
```
bdach, line 73: Not sure what the renames of this parameter are about but I don't understand what it does either from the old code or the new code.

> **(full glyph size)** as the height for each line. False if the height of **each individual glyph** should be used.

What is the difference between those "glyph sizes"? I don't get it.

osu.Framework/Text/TextBuilder.cs
```diff
@@ -332,11 +347,10 @@ protected virtual void OnWidthExceeded()
...
         /// <param name="glyph">The glyph to retrieve the height of.</param>
         /// <returns>The height of the glyph.</returns>
-        private float getGlyphHeight<T>(ref T glyph)
-            where T : ITexturedCharacterGlyph
+        private float getGlyphHeight(ref TextBuilderGlyph glyph)
         {
-            if (useFontSizeAsHeight)
-                return font.Size;
+            if (useFullGlyphHeight)
+                return glyph.Size;
```
bdach, line 353: Can you explain this change? Are `font.Size` and `glyph.Size` the same quantity, or a different one? Is `useFontSizeAsHeight` being set to `true` the same thing as `useFullGlyphHeight` being set to `true`? Because to me it seems at a read-through of this diff that it isn't the same thing.

osu.Framework/Text/ITexturedGlyphLookupStore.cs
```diff
@@ -22,5 +23,11 @@ public interface ITexturedGlyphLookupStore
         /// <param name="character">The character to retrieve.</param>
         /// <returns>The character glyph.</returns>
         Task<ITexturedCharacterGlyph> GetAsync(string fontName, char character);
+
+        /// <summary>
+        /// Searches for a <see cref="IGlyphStore"/> with the specified name.
+        /// </summary>
+        /// <param name="name">The font name.</param>
+        IGlyphStore GetFont(string name);
```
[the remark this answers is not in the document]
frenzibyte: Hmm, I was allowing it for extensibility across different font store types defined by consumers, so that `TextBuilder` still works properly with them. But perhaps there's not many different implementations for a font store, so will keep it internal for now.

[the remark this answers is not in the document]
bdach: Put it this way: I'm fine with consumers potentially adding their own `ITexturedGlyphLookupStore` implementations and implementing `GetFont()` to work with framework, but I'm less fine with consumers potentially *calling* `GetFont()` and messing with the underlying glyph store. So the simplest way to address this would be to just hide it from public API until it is actually necessary.

osu.Framework/Text/TextBuilder.cs
frenzibyte: "Full glyph size" refers to the size provided by the font usage (+ CSS scaling), while "height of each individual glyph" refers to the height of each glyph texture without any padding applied to it (using [YOffset + Height](https://www.angelcode.com/products/bmfont/doc/render_text.html)).

Would changing to:
```suggestion
        /// <param name="useFullGlyphHeight">True to use <see cref="TextBuilderGlyph.Size"/> (full glyph size) as the height for each line. False if the height of each individual glyph (<see cref="TextBuilderGlyph.YOffset"/> + <see cref="TextBuilderGlyph.Height"/>) should be used.</param>
```
explain it better?

Could also add another note that this is mainly used to allow centering text on a container (although it doesn't exactly help right now, since `YOffset` is still a problem).

osu.Framework/Text/TextBuilder.cs
frenzibyte: `glyph.Size` is equal to `font.Size` multiplied by the computed `FontMetrics.GlyphScale` (i.e. CSS scale). 

`useFullGlyphHeight` is slightly different from `useFontSizeAsHeight` in that not all glyphs have the same exact height now, and that's mostly intended for a text with multiple fonts to not look jarring (which is the point of the CSS scaling algorithm).

osu.Framework/Text/TextBuilder.cs
bdach: I don't know. Feels like this flag would be easier to explain inverted (as a "trim glyphs" sort of setting). But I wouldn't do that in this diff. Maybe fine then, I give in.

osu.Framework/Text/IHasGlyphStores.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.IO.Stores;
+
+namespace osu.Framework.Text
+{
+    internal interface IHasGlyphStores
```
peppy, line 8: I think I'd prefer that this was called `IFontStore`.

osu.Framework/Text/FontMetrics.cs
```diff
@@ -0,0 +1,56 @@
...
+        public float EmSize { get; }
+
+        /// <summary>
+        /// Creates a new <see cref="FontMetrics"/>.
+        /// </summary>
+        /// <param name="ascent">The distance from the baseline to the highest ascender, representing the font's ascent.</param>
+        /// <param name="descent">The distance from the baseline to the lowest descender, representing the font's descent.</param>
+        /// <param name="emSize">The em size (i.e. UPM, units per em).</param>
+        /// <param name="lineGap">The typographic line gap.</param>
+        public FontMetrics(float ascent, float descent, float emSize, float lineGap = 0f)
```
peppy, line 41: We probably want to test the memory overheads of this storage, as quite a few of these persist in memory for the full application lifetime (and now the content is storage twice, as far as I can tell).

osu.Framework/IO/Stores/FontStore.cs
```diff
@@ -15,8 +16,13 @@
 
 namespace osu.Framework.IO.Stores
 {
-    public class FontStore : TextureStore, ITexturedGlyphLookupStore
+    public class FontStore : TextureStore, ITexturedGlyphLookupStore, IHasGlyphStores
     {
+        /// <summary>
+        /// The font size of the raw glyph textures.
+        /// </summary>
+        internal const float BASE_FONT_SIZE = 100;
```
peppy, line 24: Might be fine to expose this publicly?

peppy, line 24: Might be fine to expose this publicly?

osu.Framework/Game.cs
```diff
@@ -167,19 +168,24 @@ private void load(FrameworkConfigManager config)
             // note that currently this means there could be two async font load operations.
             Fonts.AddStore(localFonts = new FontStore(useAtlas: false));
 
+            // Roboto and Font Awesome have different metrics for Windows and macOS.
+            // The ones used for Windows are used here for the time being.
+            var robotoMetrics = new FontMetrics(ascent: 1946, descent: 512, emSize: 2048);
```
peppy, line 173: Would be handy to link a guide or something to know how to obtain these. As a developer I'd probably end up here to figure how to add custom fonts, and have little idea how to make these.

peppy, line 173: Would be handy to link a guide or something to know how to obtain these. As a developer I'd probably end up here to figure how to add custom fonts, and have little idea how to make these.

osu.Framework.Tests/TestGame.cs
```diff
@@ -12,10 +13,17 @@ internal class TestGame : Game
...
+
         [BackgroundDependencyLoader]
         private void load()
         {
             Resources.AddStore(new NamespacedResourceStore<byte[]>(new DllResourceStore(typeof(TestGame).Assembly), "Resources"));
+
+            // nested store for framework test fonts to not interfere with framework's default (roboto) in test scenes.
+            Fonts.AddStore(testFonts = new FontStore(useAtlas: false));
+
+            AddFont(Resources, "Fonts/Noto/Noto-Basic", metrics: new FontMetrics(880, 120, 1000), target: testFonts);
```
peppy, line 26: Probably worth noting that only a few of the noto sprite sheets are included, not the whole font.

peppy, line 26: Probably worth noting that only a few of the noto sprite sheets are included, not the whole font.

osu.Framework/Text/IHasGlyphStores.cs
frenzibyte: Thought about calling it that at first, but it's an internal interface and I felt like it sounds wrong for it to be internal if it has the same name as the class type which is `public`, I dunno.

If you think it's fine naming it `IFontStore` despite it being internal (unlike `FontStore`), then sure.

osu.Framework/IO/Stores/FontStore.cs
frenzibyte: Hmm, yeah, you're right.

osu.Framework/Game.cs
frenzibyte: Have mentioned that in PR description, but I plan to write up a section in the "Fonts" wiki page about metrics and how to extract them from fonts etc.

osu.Framework/Text/IHasGlyphStores.cs
peppy: Is there a reason to not make it public?

frenzibyte: Has been discussed in https://github.com/ppy/osu-framework/pull/5123#discussion_r856929342, short of: consumers shouldn't retrieve `IGlyphStore`s and get rubbed off the wrong way with it.

peppy: I dunno, I think the naming change is fine even as `internal`.

osu.Framework/Text/FontMetrics.cs
peppy: This was supposed to be deleted after I realised that `FontMetrics` are at a font level, not glyph level :(.

osu.Framework/Graphics/Sprites/FontUsage.cs
peppy: I see it as much of a muchness as to which is used. Interested in @smoogipoo's opinion on this one.

osu.Framework/Text/TextBuilder.cs
smoogipoo: So in other words, when you specify a `FontUsage` with size 20 and set `SpriteText.UseFullGlyphHeight=true`, you'll actually get a `SpriteText` size _not_ 20, but some arbitrary value?

I believe the point of this was to have a constant size for alignment purposes, so I'm not sure about this change. Especially since the default value here is `true`.

### The following screenshots are from current master:

`UseFullGlyphHeight = false`:
image

`UseFullGlyphHeight = true`:
image

---

In other words, there are three modes of sizing:
1. Size to "glyph height" (CSS height).
2. Size to "`FontUsage` height".
3. Size to actual texture height.

Can someone confirm for me that we absolutely in no circumstance want (2), which this appears to be removing?

frenzibyte: For the aligning purposes, wouldn't you want to size the sprite text to the maximum height of each texture, rather than `FontUsage.Size` or `glyph.Size`?

Also, (2) should remain as-is when disabling CSS, but I've never imagined it to be useful for alignment?

smoogipoo: Imo that depends on the type of alignment. If you're just looking to make some text take up 20px of space, and add 5px of bounding around it to make a button, then you can't really use a `SpriteText` in there anymore, right?

You'd have to do something like:
```csharp
new Container
{
    Height = 20,
    AutoSizeAxes = Axes.X,
    Child = new SpriteText
    {
        Anchor = Anchor.Centre,
        Origin = Anchor.Centre,
        Text = "some text",
        Font = new FontUsage(size: 20)
    }
}
```
As a side note, I'm not sure whether that sort of anchoring is correct either - might have to be `BottomLeft`?

smoogipoo: Hmm... Actually `SpriteText` exposes `Height`, so I think it's fine on a second think...

smoogipoo:
The test doesn't seem to be working for me:
image

I don't seem to have any local changes, so hopefully it's not just me?

frenzibyte:
Indeed it regressed, will look into this immediately...

smoogipoo (commented):
Rest of the code looks fine at a glance, apart from the test not working.

frenzibyte:
Apparently the test scene regressed because of https://github.com/ppy/osu-framework/pull/5123/commits/78d32b636864d29903c976fb319fda4b2f3c80e9 and I swear it wasn't broken to me the moment I tested it...

That commit was changing the metrics values to use the ones which Windows respects (typo), but the text images provided were taken on macOS (hhea), therefore causing this discrepancy.

The metrics provided by Windows lead to CSS scale of 1x (`(880 + 120) / 1000 = 1`), so I'm going to stick with the macOS ones with an inline comment and call it a day.

osu.Framework/Text/TextBuilder.cs
```diff
@@ -360,11 +374,16 @@ private bool tryCreateGlyph(char character, out TextBuilderGlyph glyph)
                 return false;
             }
 
+            float size = font.Size;
+
+            if (font.CssScaling && fontStoreGlyph.Metrics is FontMetrics metrics)
+                size *= metrics.GlyphScale;
```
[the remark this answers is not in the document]
frenzibyte: Do you expect something like `getGlyphSize(FontUsage font, FontMetrics metrics)`?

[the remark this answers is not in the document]
smoogipoo: Just a `getGlyphHeight()`. I think everything you need is already part of the class-level variables.

osu.Framework/Text/TextBuilder.cs
```diff
@@ -62,24 +70,26 @@ public float LineBaseHeight
         /// </summary>
         /// <param name="store">The store from which glyphs are to be retrieved from.</param>
         /// <param name="font">The font to use for glyph lookups from <paramref name="store"/>.</param>
-        /// <param name="useFontSizeAsHeight">True to use the provided <see cref="font"/> size as the height for each line. False if the height of each individual glyph should be used.</param>
+        /// <param name="useFullGlyphHeight">True to use <see cref="TextBuilderGlyph.Size"/> (full glyph size) as the height for each line. False if the height of each individual glyph (<see cref="TextBuilderGlyph.YOffset"/> + <see cref="TextBuilderGlyph.Height"/>) should be used.</param>
```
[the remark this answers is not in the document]
frenzibyte: @bdach seeing that you struggled to read how this was worded initially (https://github.com/ppy/osu-framework/pull/5123#discussion_r856930213), what's your opinion on changing it as suggested above, i.e.:
```diff
-        /// <param name="useFullGlyphHeight">True to use <see cref="TextBuilderGlyph.Size"/> (full glyph size) as the height for each line. False if the height of each individual glyph (<see cref="TextBuilderGlyph.YOffset"/> + <see cref="TextBuilderGlyph.Height"/>) should be used.</param>
+        /// <param name="useFullGlyphHeight">
+        /// True if the <see cref="TextBuilder"/>'s vertical bounds should be equal to the maximum full height of the glyphs (constant per font).
+        /// False to consider the maximum texture height of each individual glyph instead.
+        /// </param>
```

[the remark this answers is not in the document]
bdach: It's marginally better I guess. It still confuses me that "full height of the glyphs" is mentioned in one place and "height of each individual glyph" is in the other. The difference feels extremely subtle.

If I were to propose something then it'd probably be something along the lines of

```
        /// <param name="useFullGlyphHeight">
        /// When true, all texts that use a particular font will have a constant height, regardless of the letters used.
        /// When false, the height of a given text will be cropped to only fit the letters used in that particular text,
        /// and extra space above and below letters will be trimmed.
        /// </param>
```

osu.Framework/Graphics/Sprites/FontUsage.cs
frenzibyte: > Assuming, hopefully, at some point in the future we'll turn css-sizing on by default and require users to set css: false where that isn't desired. That can be done in a further PR after this is tested with the game.

Yep, that's the general direction I had in mind. Renaming to `cssScaling` seems fine presuming we'll go with the way of defining a public static `FontUsage` copy with it enabled, somewhat similar to the diff in https://github.com/ppy/osu-framework/pull/5123#discussion_r856862551.

bdach: I still disagree with the "public static `FontUsage` copy" thing. It doesn't achieve anything other than look really weird (what is an `OldTorus`?), and you immediately have to migrate all usages across to not break stuff. I suggest doing it the opposite way - specifying `cssScaling` locally where needed incrementally, and only removing the param once it becomes `true` everywhere.

bdach: Also note that the `OldTorus` thingy could (and likely would) transitively visually break rulesets, too.

frenzibyte: Hmm... let's leave discussing this in the osu!-side PR instead. Since we all agree with the `cssScaling: true` parameter here, that's enough to mark this conversation as resolved.

bdach:
> That commit was changing the metrics values to use the ones which Windows respects (typo), but the text images provided were taken on macOS (hhea)

Well this is disconcerting, are you saying this can differ per-platform somehow, too? I would hope not...

bdach:
I think I'm just going to pretend I didn't ask and nod along. This is a rabbit hole the depths of which I do not dare to explore.

bdach (approved):

frenzibyte:
Another use case popped up just now which required this PR, a font was proposed to fit better with Torus on Cyrillic glyphs, but the line height set on the font was too large. Without this PR, when attempting to import this font for the game, the characters will look too small compared to other fonts on the same font size. See [discord](https://discord.com/channels/188630481301012481/1097318920991559880/1385418854515081246).

bdach:
Or you could, you know, mess with the font some to change the line height using fontforge or something. [We already run post-processing on other fonts to fix weird metrics issues.](https://github.com/ppy/osu-resources/blob/master/osu.Game.Resources/Fonts/Venera/venera-postprocess.py)

PR #5124, opened 2022-04-23
Add mapping for the windows context menu key (and log unknown SDL keys)

Susko3:
The log message renders as:
```cs
[runtime] 2022-04-23 16:09:22 [verbose]: Unknown SDL key: SDL_SCANCODE_APPLICATION, SDLK_APPLICATION
```

The log will help if more unknown keys crop up in the future (as there are many scancodes we're not handling).

The menu key is reported by windows/SDL as `SDL_SCANCODE_APPLICATION`, and is listed as such on the [SDL wiki](https://wiki.libsdl.org/SDL_Scancode):
- "Application" (the Application / Compose / Context Menu (Windows) key)

image


frenzibyte (approved):
Sounds good. Will leave merge for someone who can test on Windows.

bdach (approved):
tested to work as expected

PR #5127, opened 2022-04-24
Add ability to transform localisable strings to sentence case

Joehuu:
Wanted to add after I regressed casing here:

image

Context menu item text is usually sentence case. I could use `ToTitle()` to fix in this case (as it's one word), but there can potentially be a translation that is more than one word.


osu.Framework/osu.Framework.csproj
```diff
@@ -21,6 +21,7 @@
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
   </PropertyGroup>
   <ItemGroup Label="Package References">
+    <PackageReference Include="Humanizer" Version="2.14.1" />
```
bdach, line 24: I don't know how to feel about pulling humanizer into framework directly rather than only doing it in game. It's not a small library, when you consider all of the related satellite assemblies. Definitely puts this pull into "not comfortable enough to merge without consensus that we want this" territory for me.

frenzibyte: Sounds like the sentence casing support should happen osu!-side instead, if feasible.

peppy: I agree that if we've avoided it thus far, it's probably better to continue to do so. Less dependencies is better.

Joehuu: I don't think this should be in osu!-side. It can be a commonly used method, and I foresee others using this (and myself if I ever make a project with this framework).

Would using `Humanizer.Core` resolve this? We are only using English and basic string manipulation (e.g. this PR) in this repo.

If not, we could take the relevant logic instead:

```diff
diff --git a/osu.Framework/Localisation/CaseTransformableString.cs b/osu.Framework/Localisation/CaseTransformableString.cs
index 9845fd46b..18bbc90f8 100644
--- a/osu.Framework/Localisation/CaseTransformableString.cs
+++ b/osu.Framework/Localisation/CaseTransformableString.cs
@@ -5,7 +5,6 @@

 using System;
 using System.Globalization;
-using Humanizer;

 namespace osu.Framework.Localisation
 {
@@ -38,26 +37,37 @@ public CaseTransformableString(LocalisableString str, Casing casing)
         public string GetLocalised(LocalisationParameters parameters)
         {
             string stringData = getStringData(parameters);
-            var cultureInfo = parameters.Store?.EffectiveCulture ?? CultureInfo.InvariantCulture;
+            var cultureText = parameters.Store?.EffectiveCulture?.TextInfo ?? CultureInfo.InvariantCulture.TextInfo;

             switch (Casing)
             {
                 case Casing.UpperCase:
-                    return stringData.Transform(cultureInfo, To.UpperCase);
+                    return cultureText.ToUpper(stringData);

                 case Casing.TitleCase:
-                    return stringData.Transform(cultureInfo, To.TitleCase);
+                    return cultureText.ToTitleCase(stringData);

                 case Casing.LowerCase:
-                    return stringData.Transform(cultureInfo, To.LowerCase);
+                    return cultureText.ToLower(stringData);

                 case Casing.SentenceCase:
-                    return stringData.Transform(cultureInfo, To.SentenceCase);
+                    return toSentenceCase(stringData, cultureText);

                 case Casing.Default:
                 default:
                     return stringData;
             }
+
+            // taken from https://github.com/Humanizr/Humanizer/blob/606e958cb83afc9be5b36716ac40d4daa9fa73a7/src/Humanizer/Transformer/ToSentenceCase.cs#L12-L22
+            string toSentenceCase(string input, TextInfo textInfo)
+            {
+                if (input.Length >= 1)
+                {
+                    return textInfo.ToUpper(input[0]) + input.Substring(1);
+                }
+
+                return textInfo.ToUpper(input);
+            }
         }

         public override string ToString() => GetLocalised(new LocalisationParameters(null, false));
diff --git a/osu.Framework/osu.Framework.csproj b/osu.Framework/osu.Framework.csproj
index 346f935e5..9f8bb6179 100644
--- a/osu.Framework/osu.Framework.csproj
+++ b/osu.Framework/osu.Framework.csproj
@@ -21,7 +21,6 @@
     <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
   </PropertyGroup>
   <ItemGroup Label="Package References">
-    <PackageReference Include="Humanizer" Version="2.14.1" />
     <PackageReference Include="managed-midi" Version="1.9.14" />
     <PackageReference Include="FFmpeg.AutoGen" Version="4.3.0.1" />
     <PackageReference Include="Microsoft.Extensions.ObjectPool" Version="5.0.11" />
```

frenzibyte: Better go with inline logic than referencing any packages in framework, to avoid future headaches IMO.

smoogipoo: Agree with extracting the required logic, more so since it's only 3 LOC.

huoyaoyuan: Localisation can be really painful when it comes to languages we are not familiar with. For example, a letter or something may map to multiple `char`s. It's better to ask speakers of some common mentioned language, e.g. Turkish.

As a Chinese speaker this makes no sense to me since there's no casing in Chinese. Generally, I tend to not manipulate strings because they can be complex.

smoogipoo: Until necessary, I don't think we need to have multiple translations for the strings "title", "Title", and "TITLE", depending on where they're used. Which is why this PR exists, and why the other methods are already provided.

There's one case I'm aware of where things may not be expected, and that is `.ToLower()`ing German strings where the noun is always title-cased.

Any such bad cases can be fixed on a case-by-case basis, by adding more translation strings. It's not a big deal.

bdach (approved):

peppy (approved):

PR #5128, opened 2022-04-24
Fix `Menu` not handling mouse down events

frenzibyte:
- Closes ppy/osu#17961

Was previously relying on its `ScrollContainer` layer which used to block mouse down from drawables behind it, until https://github.com/ppy/osu-framework/pull/5118.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneClosableMenu.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneClosableMenu.cs
@@ -1,6 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using NUnit.Framework;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.UserInterface;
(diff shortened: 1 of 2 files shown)


osu.Framework.Tests/Visual/UserInterface/TestSceneClosableMenu.cs
```diff
@@ -72,6 +73,38 @@ public void TestMenuIgnoresEscapeWhenClosed()
...
+
+            AddStep("add mouse handler", () => Add(new MouseHandlingLayer
+            {
+                Action = () => actionReceived = true,
+                Depth = 1,
+            }));
+
+            AddStep("click item", () => ClickItem(0, 0));
+            AddStep("click item", () => ClickItem(1, 0));
+            AddAssert("mouse handler not activated", () => !actionReceived);
```
bdach, line 89: Would also like to see an assertion that the menu item's `Action` was fired. Checking that the background drawables didn't receive the press is important but secondary to making sure the menu items are actually menu iteming properly.

frenzibyte: Sounds fair.

bdach (approved):

bdach:
Seems sane but I'll wait for a second approval.

peppy (approved):

PR #5130, opened 2022-04-24
Improve `LocalisableString` formatting and interpolation API

frenzibyte:
- May resolve https://github.com/ppy/osu/issues/17953

Upon discussion in [discord](https://discord.com/channels/188630481301012481/589331078574112768/966564266343817266), it turns out the current way to format `LocalisableString`s or create interpolation is via `TranslatableString` with a random key (e.g. `_`), which feels counter-intuitive.

I've cleaned things up by moving the format and interpolation logic from `TranslatableString` to `LocalisableFormattableString`, and sharing it to `TranslatableString` via inheritance (rationale is that `TranslatableString` shouldn't really bother with the formatting logic, and `LocalisableFormattableString` currently exists but has been incomplete). 

Allowing for formatting and interpolation support via `LocalisableFormattableString` instead:
```diff
                 public override LocalisableString TooltipText =>
                     Current.Value == 0
-                        ? new TranslatableString("_", @"{0} ms", base.TooltipText)
-                        : new TranslatableString("_", @"{0} ms {1}", base.TooltipText, getEarlyLateText(Current.Value));
+                        ? new LocalisableFormattableString($@"{base.TooltipText} ms")
+                        : new LocalisableFormattableString($@"{base.TooltipText} ms {getEarlyLateText(Current.Value)}");
```

In addition, I've added a `LocalisableString.Format` method (analogue to `string.Format`), [as proposed](https://discord.com/channels/188630481301012481/589331078574112768/966572503604539423), which can be used to avoid specifying the full constructor name, or if format is preferred:
```diff
                 public override LocalisableString TooltipText =>
                     Current.Value == 0
-                        ? new LocalisableFormattableString($@"{base.TooltipText} ms")
-                        : new LocalisableFormattableString($@"{base.TooltipText} ms getEarlyLateText(Current.Value));
+                        ? LocalisableString.Format(@"{0} ms", base.TooltipText)
+                        : LocalisableString.Format(@"{0} ms {1}", base.TooltipText, getEarlyLateText(Current.Value));
```

I've also thought about making it full-blown with an implicit conversion in `LocalisableString` from an interpolation string (internally `FormattableString`) to a `LocalisableFormattableString`. But turns out that doesn't work well due to the pre-existing `string` implicit conversion, which takes priority over the `FormattableString` one, [as mentioned](https://discord.com/channels/188630481301012481/589331078574112768/966573544194596894).

This has been brought up before in [stack overflow](https://stackoverflow.com/questions/35770713/overloaded-string-methods-with-string-interpolation), and has been discussed long ago across VB and C# (https://github.com/dotnet/roslyn/issues/46), but it has since been decided for `string`s to remain at taking priority over `FormattableString`s.

The only improvement I can think of to benefit from the interpolation and avoid explicitly constructing the implementation class with its full name is by adding a `LocalisableString.Interpolate` that does the job:
```diff
                 public override LocalisableString TooltipText =>
                     Current.Value == 0
-                        ? new LocalisableFormattableString($@"{base.TooltipText} ms")
-                        : new LocalisableFormattableString($@"{base.TooltipText} ms getEarlyLateText(Current.Value));
+                        ? LocalisableString.Interpolate($@"{base.TooltipText} ms")
+                        : LocalisableString.Interpolate($@"{base.TooltipText} ms {getEarlyLateText(Current.Value)}");
```

I'm pretty happy with the state this is now in, but further suggestions and ideas are welcomed.

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkLocalisableString.cs
+++ b/osu.Framework.Benchmarks/BenchmarkLocalisableString.cs
@@ -32,8 +32,8 @@ namespace osu.Framework.Benchmarks
             romanisableString2 = new RomanisableString("c", "d");
             translatableString1 = new TranslatableString("e", "f");
             translatableString2 = new TranslatableString("g", "h");
-            formattableString1 = new LocalisableFormattableString("i", new object[] { "j" });
-            formattableString2 = new LocalisableFormattableString("k", new object[] { "l" });
+            formattableString1 = new LocalisableFormattableString("({0})", new object[] { "j" });
+            formattableString2 = new LocalisableFormattableString("[{0}]", new object[] { "l" });
         }
 
         [Benchmark]
--- a/osu.Framework.Benchmarks/BenchmarkLocalisableString.cs
+++ b/osu.Framework.Benchmarks/BenchmarkLocalisableString.cs
@@ -32,8 +32,8 @@ namespace osu.Framework.Benchmarks
             romanisableString2 = new RomanisableString("c", "d");
             translatableString1 = new TranslatableString("e", "f");
             translatableString2 = new TranslatableString("g", "h");
-            formattableString1 = new LocalisableFormattableString(1.23, "N0");
-            formattableString2 = new LocalisableFormattableString(4.56, "N1");
+            formattableString1 = new LocalisableFormattableString("i", new object[] { "j" });
+            formattableString2 = new LocalisableFormattableString("k", new object[] { "l" });
         }
 
         [Benchmark]
--- a/osu.Framework.Benchmarks/BenchmarkLocalisableString.cs
+++ b/osu.Framework.Benchmarks/BenchmarkLocalisableString.cs
@@ -32,8 +32,8 @@ namespace osu.Framework.Benchmarks
             romanisableString2 = new RomanisableString("c", "d");
             translatableString1 = new TranslatableString("e", "f");
             translatableString2 = new TranslatableString("g", "h");
-            formattableString1 = new LocalisableFormattableString("({0})", new object[] { "j" });
-            formattableString2 = new LocalisableFormattableString("[{0}]", new object[] { "l" });
+            formattableString1 = LocalisableString.Format("({0})", "j");
+            formattableString2 = LocalisableString.Format("[{0}]", "l");
         }
 
         [Benchmark]
--- a/osu.Framework.Tests/Localisation/LocalisableStringTest.cs
+++ b/osu.Framework.Tests/Localisation/LocalisableStringTest.cs
@@ -1,7 +1,6 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System;
 using System.Collections.Generic;
 using NUnit.Framework;
 using osu.Framework.Localisation;
@@ -13,9 +12,6 @@ namespace osu.Framework.Tests.Localisation
     {
         private string makeStringA => makeString('a');
         private string makeStringB => makeString('b');
-
-        private IFormattable makeFormattableA => new DateTime(1);
-        private IFormattable makeFormattableB => new DateTime(2);
 
         [Test]
         public void TestTranslatableStringEqualsTranslatableString()
@@ -42,11 +38,11 @@ namespace osu.Framework.Tests.Localisation
         [Test]
         public void TestLocalisableFormattableEqualsLocalisableFormattable()
         {
-            var str1 = new LocalisableFormattableString(makeFormattableA, makeStringB);
-            var str2 = new LocalisableFormattableString(makeFormattableB, makeStringA);
+            var str1 = new LocalisableFormattableString(makeStringA, new object[] { makeStringB });
+            var str2 = new LocalisableFormattableString(makeStringB, new object[] { makeStringA });
 
             testEquals(true, str1, str1);
-            testEquals(true, str1, new LocalisableFormattableString(makeFormattableA, makeStringB));
+            testEquals(true, str1, new LocalisableFormattableString(makeStringA, new object[] { makeStringB }));
             testEquals(false, str1, str2);
         }
 
@@ -84,10 +80,10 @@ namespace osu.Framework.Tests.Localisation
         [Test]
         public void TestLocalisableStringEqualsLocalisableFormattable()
         {
-            LocalisableString localisable = new LocalisableFormattableString(makeFormattableA, makeStringB);
+            LocalisableString localisable = new LocalisableFormattableString(makeStringA, new object[] { makeStringB });
 
-            testEquals(true, localisable, new LocalisableFormattableString(makeFormattableA, makeStringB));
-            testEquals(false, localisable, new LocalisableFormattableString(makeFormattableB, makeStringA));
+            testEquals(true, localisable, new LocalisableFormattableString(makeStringA, new object[] { makeStringB }));
+            testEquals(false, localisable, new LocalisableFormattableString(makeStringB, new object[] { makeStringA }));
         }
 
         [Test]
--- a/osu.Framework.Tests/Localisation/LocalisableStringTest.cs
+++ b/osu.Framework.Tests/Localisation/LocalisableStringTest.cs
@@ -38,11 +38,11 @@ namespace osu.Framework.Tests.Localisation
         [Test]
         public void TestLocalisableFormattableEqualsLocalisableFormattable()
         {
-            var str1 = new LocalisableFormattableString(makeStringA, new object[] { makeStringB });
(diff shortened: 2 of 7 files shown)


peppy:
I think the syntax looks fine. Would you be able to make a wiki page covering the various syntax available for localisation as part of this? I still find it hard to locate the correct class/method when working with these and it'd be handy to have all the invocations listed in one place (that isn't a unit test).

frenzibyte:
Sounds good, will look into that.

osu.Framework.Tests/Localisation/LocalisationTest.cs
```diff
@@ -140,6 +140,34 @@ public void TestFormattedAndLocalised()
             Assert.AreEqual(expectedResult, formattedText.Value);
         }
 
+        [Test]
+        public void TestFormattedAndLocalisedUsingInterpolate()
+        {
+            var translatable1 = new TranslatableString(FakeStorage.LOCALISABLE_STRING_EN, FakeStorage.LOCALISABLE_STRING_EN);
+            var translatable2 = new TranslatableString(FakeStorage.LOCALISABLE_FORMAT_STRING_EN, FakeStorage.LOCALISABLE_FORMAT_STRING_EN, 0.1234.ToLocalisableString("0.00%"));
```
peppy, line 147: I know that this is just a test, but I don't think we'd even call an extension method on a raw number like this. Maybe change this to use `LocalisableFormattableString` instead?

osu.Framework/Localisation/LocalisableString.cs
```diff
@@ -33,20 +33,33 @@ public LocalisableString(ILocalisableStringData data)
...
+
         // it's somehow common to call default(LocalisableString), and we should return empty string then.
         public override string ToString() => Data?.ToString() ?? string.Empty;
 
         public bool Equals(LocalisableString other) => LocalisableStringEqualityComparer.Default.Equals(this, other);
         public override bool Equals(object? obj) => obj is LocalisableString other && Equals(other);
         public override int GetHashCode() => LocalisableStringEqualityComparer.Default.GetHashCode(this);
 
+        public static bool operator ==(LocalisableString left, LocalisableString right) => left.Equals(right);
+        public static bool operator !=(LocalisableString left, LocalisableString right) => !left.Equals(right);
```
bdach, line 57: why are members being moved around unchanged again, for no discernible reason...

osu.Framework/Localisation/TranslatableString.cs
```diff
@@ -42,65 +36,25 @@ public TranslatableString(string key, string fallback, params object?[] args)
...
-                Logger.Log($"Localised format failed. Key: {Key}, culture: {parameters.Store.EffectiveCulture}, fallback format string: \"{Fallback}\", localised format string: \"{localisedFormat}\". Exception: {e}",
-                    LoggingTarget.Runtime, LogLevel.Verbose);
-            }
-
-            return ToString();
-        }
-
-        public override string ToString() => string.Format(CultureInfo.InvariantCulture, Fallback, Args);
+        protected override string FormatString(string format, object?[] args, LocalisationParameters parameters)
+            => base.FormatString(parameters.Store?.Get(Key) ?? format, args, parameters);
```
bdach, line 45: this is pretty convoluted, as you have to remember that `format` is the `Fallback` which was provided to the base ctor - otherwise it looks pretty weird to me at face value that `format` is being overridden like so. could be cleaned up by making the actual fetching of the format string virtual rather than making the full formatting operation virtual but probably not worth changing?

osu.Framework.Benchmarks/BenchmarkLocalisableString.cs
```diff
@@ -32,8 +32,8 @@ public void GlobalSetup()
             romanisableString2 = new RomanisableString("c", "d");
             translatableString1 = new TranslatableString("e", "f");
             translatableString2 = new TranslatableString("g", "h");
-            formattableString1 = new LocalisableFormattableString(1.23, "N0");
-            formattableString2 = new LocalisableFormattableString(4.56, "N1");
+            formattableString1 = new LocalisableFormattableString("i", new object[] { "j" });
+            formattableString2 = new LocalisableFormattableString("k", new object[] { "l" });
```
bdach, line 36: why is this being changed from actual data to garbage that doesn't reflect any sane usage?

bdach (commented):
i guess this mostly works

osu.Framework.Tests/Localisation/LocalisationTest.cs
bdach: still a few of these left in this file

https://github.com/ppy/osu-framework/blob/dd6ea246f85b86dc3b5e236c8214ffffe8647737/osu.Framework.Tests/Localisation/LocalisationTest.cs#L367-L368

https://github.com/ppy/osu-framework/blob/dd6ea246f85b86dc3b5e236c8214ffffe8647737/osu.Framework.Tests/Localisation/LocalisationTest.cs#L394-L395

osu.Framework.Benchmarks/BenchmarkLocalisableString.cs
frenzibyte: Well the benchmark focuses on equality and hashcode but sure, can change.

osu.Framework/Localisation/LocalisableString.cs
frenzibyte: Well it’s changed to be below the equality methods, as it should ever be. But if you’re completely against moving members around, I might as well just never do that and wait until we have analysers in place.

osu.Framework/Localisation/TranslatableString.cs
frenzibyte: Hmm, that sounds fair, will see.

osu.Framework/Localisation/LocalisableString.cs
bdach: > changed to be below the equality methods, as it should ever be

this is like extremely arbitrary? i could as well say that it was together with overloaded operators before.

> if you’re completely against moving members around

i'm not, i'm against moving members around when it barely makes any sense and just adds pure mental overhead in reviewing. i have to go and check that the implementation wasn't changed. the reorder wasn't split out to a separate commit even.

frenzibyte: I mean at that point, I don’t really see how splitting to another commit would help unless you’re reviewing commit-by-commit, but either way I’ll refrain from moving members around.

osu.Framework/Localisation/TranslatableString.cs
bdach: could also just rename the `format` param to `fallback` only in the override for a simpler change. a bit unorthodox but hopefully aids understanding.

osu.Framework.Tests/Localisation/LocalisableStringTest.cs
```diff
@@ -42,11 +38,11 @@ public void TestRomanisableStringEqualsRomanisableString()
         [Test]
         public void TestLocalisableFormattableEqualsLocalisableFormattable()
         {
-            var str1 = new LocalisableFormattableString(makeFormattableA, makeStringB);
-            var str2 = new LocalisableFormattableString(makeFormattableB, makeStringA);
+            var str1 = new LocalisableFormattableString(makeStringA, new object[] { makeStringB });
```
smoogipoo, line 41: Is there a reason for not adding `params object?[]` to the ctor?

smoogipoo (commented):
Seems okay apart from the test failure.

osu.Framework.Tests/Localisation/LocalisableStringTest.cs
frenzibyte: Have mentioned it in the ctor itself, basically breaks the `FormattableString` ctor:
https://github.com/ppy/osu-framework/blob/93b7da9b33ea0e5ef49885c47a2fc973cd3355d1/osu.Framework/Localisation/LocalisableFormattableString.cs#L33-L39

smoogipoo: Hmm... I question if we still need to expose `LocalisableFormattableString` since its functionality seems to be completely handled by the new helpers, which was the point of this PR.

Maybe the ctor could be `protected internal`? That is, its only use-case in external code would be via a subclass.

I'd like the localisation structure to very closely mimic that of `string`, so I'd hope there's no reason to use the raw `LocalisableFormattableString`:
```
x.ToString("fmt") -> x.ToLocalisableString("fmt");
string.Format(a, b) -> LocalisableString.Format(a, b);
```

frenzibyte: Sounds good, have pushed changes, replacing all constructor usages with `LocalisableString.Format` instead.

smoogipoo (approved):

peppy (approved):

PR #5135, opened 2022-04-26
Improve extensibility of `LocalisationParameters`

frenzibyte:
After writing up a [wiki page regarding localisation](https://github.com/ppy/osu-framework/wiki/Localisation), it turns out `LocalisationParameters` isn't actually extensible given the way it's currently created and updated in `LocalisationManager`, along with the fact that there's no `CreateLocalisationManager` to create the custom subclass with.

The end goal would be to extend `LocalisationParameters` from a consumer project with simplicity in mind:
```csharp
    protected override LocalisationManager CreateLocalisationManager(FrameworkConfigManager frameworkConfig) 
        => new CustomLocalisationManager(frameworkConfig, localConfig);

    public class CustomLocalisationManager : LocalisationManager
    {
        private readonly IBindable<int> custom = new BindableInt();

        public CustomLocalisationManager(FrameworkConfigManager frameworkConfig, CustomConfigManager config)
            : base(frameworkConfig)
        {
            config.BindWith(CustomSetting.Custom, custom);
            custom.BindValueChanged(_ => UpdateLocalisationParameters(), true);
        }

        protected override LocalisationParameters CreateLocalisationParameters() => new CustomLocalisationParameters(base.CreateLocalisationParameters(), custom.Value);

        private class CustomLocalisationParameters : LocalisationParameters
        {
            public readonly int Custom;

            public CustomLocalisationParameters(LocalisationParameters parameters, int custom)
                : base(parameters.Store, parameters.PreferOriginalScript)
            {
                Custom = custom;
            }
        }
    }

### DIFF
--- a/osu.Framework/Game.cs
+++ b/osu.Framework/Game.cs
@@ -69,6 +69,12 @@ namespace osu.Framework
 
         protected override Container<Drawable> Content => content;
 
+        /// <summary>
+        /// Creates a new <see cref="LocalisationManager"/>.
+        /// </summary>
+        /// <param name="frameworkConfig">The framework config manager.</param>
+        protected virtual LocalisationManager CreateLocalisationManager(FrameworkConfigManager frameworkConfig) => new LocalisationManager(frameworkConfig);
+
         protected internal virtual UserInputManager CreateUserInputManager() => new UserInputManager();
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


smoogipoo:
I think it's better to pass the parameters down to the base class via a protected ctor. Means that if we ever add new parameters to `LocalisationParameters`, subclasses won't have to be updated to support it.

smoogipoo (dismissed):

smoogipoo (approved):
Lgtm

Issue #4159, opened 2021-01-19
Mouse position is changed upon alt-tabbing in fullscreen mode


Morilli:
When using fullscreen mode, alt-tabbing seems to cause a brief switch to windowed mode or a different resolution, which in turn seems to cause a change in mouse position. This can get pretty annoying after a while, and I don't think it should happen in the first place. Probably somewhat related to #4074.

https://user-images.githubusercontent.com/35152647/105013273-5e82c500-5a3f-11eb-870b-c5efdad9dc0b.mp4

Tested on current Windows 10 x64 with osu!lazer release 2021.118.1

Issue #4050, opened 2020-12-01
Cursor stuck top left when leaving game when it's fullscreen & raw input


LittleEndu:
https://youtu.be/CX7_A6Ylyaw (first part is fullscreen, second is borderless)

[DxDiag.txt](https://github.com/ppy/osu-framework/files/5623650/DxDiag.txt)

bdach:
What pointing device is this? Mouse, tablet? dxdiag lists multiple.

LittleEndu:
Both are stuck. Tablet does move the cursor freely but any cycle where there is no input from the tablet it goes back to top left.

LittleEndu:
It doesn't seem to. Unplugging the mouse didn't fix, unplugging the tablet didn't fix. Worth noting (and I think this might even invalidate the issue) is that I'm not on vendor software, I'm using hawku drivers. But the issue persist when I uninstall the driver and let windows grab generic driver after plugging the tablet back in.

bdach:
I can't say I tested SDL on windows but it really ought to be working with *some* device...

nbvdkamp:
I encountered this too on Windows with only mouse (don't own a tablet), upon following an external URL so it seems to be unrelated to tablets. It can get stuck in the corner of either monitor in multi monitor setups.

peppy:
Just to confirm, this only occurs until you alt-tab cycle once, after which point it's fixed?

Joehuu:
Okay, first off the PR doesn't fix the issue.

From my findings on Windows 10, I can now reproduce on framework by a different method. It needs to be windowed, and raw input needs to be on. Then, you need to move the cursor really fast inside and press `WIN-D`. The cursor will then teleport to the top left, but it won't get stuck.
![SQqdwloMUS](https://user-images.githubusercontent.com/35318437/100843930-ef6df480-342f-11eb-9278-56154be1ca27.gif)

On lazer, I can also reproduce on windowed with raw input by pressing `WIN-D`. Cursor still gets stuck though.
![ETERgwPyW3](https://user-images.githubusercontent.com/35318437/100844139-4378d900-3430-11eb-98e1-094aab14a3db.gif)

> Just to confirm, this only occurs until you alt-tab cycle once, after which point it's fixed?

Nope, for the fullscreen reproduction, you just need to minimize fullscreen in any way (alt-tab, win-d, etc.) and it'll never recover.

peppy:
[...]

PR #5136, opened 2022-04-26
Fix cursor warping to center of screen when alt-tabbing in fullscreen

Susko3:
Depends on:
- [x] SDL bump (for https://github.com/libsdl-org/SDL/commit/6c962177272c6a2ce684e4298b40c25ebdd1ac74)

Closes https://github.com/ppy/osu-framework/issues/4159.
Closes https://github.com/ppy/osu/issues/12228.
Closes https://github.com/ppy/osu-framework/issues/4050.

Setting `SDL_HINT_MOUSE_RELATIVE_MODE_CENTER` to `0` does fix the cursor position when `Sensitivity = 1`. Manual warping is required for other cases.

Warping the cursor with the current code fails as the events leading to the warp are queued in the SDL event queue, leading to the warp happening when the window has already lost focus and is minimized. So we have to warp earlier, done here by handling the `WM_KILLFOCUS`  message.

[`SDL_WarpMouseInWindow`](https://github.com/libsdl-org/SDL/blob/b293888c7dcf11bb6198ade4990043c226edcdc1/src/events/SDL_mouse.c#L889-L894) doesn't warp the OS cursor if relative mode is enabled, so we have to use `SDL_WarpMouseGlobal` / [`SetCursorPos`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setcursorpos) directly. (I'd rather not turn off relative as it might mess with `Window.RelativeMouseMode` state.)

The cursor will sometimes fail to warp if the mouse is moving. This [appears](https://github.com/libsdl-org/SDL/blob/3167ba342d3b7cf5459405d9d79fe56afefab431/src/video/windows/SDL_windowsmouse.c#L255) to be a bug in windows, where the warped position is overwritten by the mouse moving.

Tested that the cursor warps to the expected position at display scaling of 100% and 150%.


peppy (approved):
Tested to work correctly on windows 11. Code seems okay.

PR #5138, opened 2022-04-28
Add CI workflow for automated FFmpeg builds

smoogipoo:
On manual workflow dispatch, builds all relevant libraries and creates a PR. Sample PR can be seen here: https://github.com/smoogipoo/osu-framework/pull/82 (respective CI run: https://github.com/smoogipoo/osu-framework/actions/runs/2238241951)

This PR doesn't add/change the libs, just adds the build process.

I've tested the resulting libs on linux-x64 inclusive of a manual workaround. The rest (especially macOS) will still need to be tested, however I'd like to do that as part of the resulting PR, which will also contain a linux-specific library loading workaround.


.github/workflows/build-ffmpeg.yml
```diff
@@ -0,0 +1,165 @@
+name: Build FFmpeg
+on: workflow_dispatch
+
+jobs:
+  build-osx:
+    name: Build OSX
```
peppy, line 6: Is there a reason for using `osx` instead of `macOS` in various places in this PR?

smoogipoo: No reason. I've changed it to `macOS` everywhere.

osu.Framework.NativeLibs/scripts/ffmpeg/common.sh
```diff
@@ -0,0 +1,39 @@
...
+FFMPEG_FLAGS=(
+    --disable-programs
+    --disable-doc
+    --disable-static
+    --disable-debug
+    --disable-ffplay
+    --disable-ffprobe
+    --disable-avdevice
+    --disable-swresample
+    --disable-librtmp
```
smoogipoo, line 14: Note that these 5 disables are new in this PR, and were taken from https://github.com/ppy/osu-framework/pull/3846. I've tested that avdevice and swresample aren't used, and have tested the resulting libs so I don't expect breakage as a result of this.

osu.Framework.NativeLibs/scripts/ffmpeg/macOS/fetch_and_build.sh
```diff
@@ -1,32 +1,21 @@
...
+pushd $(dirname $0) > /dev/null
+SCRIPT_PATH=$(pwd)
+popd > /dev/null
+source $SCRIPT_PATH/../common.sh
+
+FFMPEG_FLAGS+=(
+    --target-os=darwin
+    --arch=$arch
+    --enable-cross-compile
+    --extra-cflags='-arch $arch'
```
peppy, line 12: I'm not sure how much we need this to work outside of github actions, but I think something about the licencing means that we do generally want these to be user-runnable. Since the instructions were removed, can we default `$arch` to the local system's arch if not specified?

also I can't seem to run this on my system, not sure why though..

```csharp
scripts/ffmpeg/macOS on  ffmpeg-build took 1m25s
❯ arch=x64 ./fetch_and_build.sh
-> Downloading source...                                                                                                       % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 13.4M  100 13.4M    0     0  1901k      0  0:00:07  0:00:07 --:--:-- 2707k                                               -> Configuring...
gcc is unable to create an executable file.
C compiler test failed.
```

I do have `gcc` installed and it has worked for other building tasks.

smoogipoo: I didn't really intend for these to be user runnable, which is why they don't do things like checking if the `ffmpeg-x.x.x` directory exists, but I'll see what can be done.

Regarding macOS, does it work by removing these 3 lines?
image

peppy: Yeah I'm not sure it's required since a user could just run the github actions, but I know that was why @nekodex had build instructions previously (to meet licence requirements).

peppy: > Regarding macOS, does it work by removing these 3 lines? image

seems to work

peppy (approved):

osu.Framework.NativeLibs/scripts/ffmpeg/macOS/fetch_and_build.sh
smoogipoo: For others following along, the cause of this was `arch=x64` when it should've been `arch=x86_64`. I've added back the prompt to select the arch for manual builds when `$arch` isn't set.

PR #5140, opened 2022-04-29
Optimise operations of Streams

huoyaoyuan:
- Override `Span` based write methods for derived streams since they are preferred in many library code.
- Optimise the extension methods, which are the main way we access streams in osu. Reduce allocations and copies.

Not touching some cases like TcpIpc, because it should be a rare path. It's too bothersome to maximise performance everywhere.


osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -133,8 +133,21 @@ internal TrackBass(Stream data, bool quick = false)
...
+                    // Buffering memory stream is definitely unworthy.
+                    dataStream = data;
+                    break;
+
+                default:
+                    // It would be most likely a FileStream.
+                    // Is it benenitial to buffer over OS and .NET file buffering?
+                    dataStream = new AsyncBufferStream(data, quick ? 8 : -1);
+                    break;
+            }
```
huoyaoyuan, line 150: I have a total rewrite of AsyncBufferStream to favor the performance of FileStream, but it requires .NET 6. I tend to left it as is now.

osu.Framework/IO/Stores/RawCachingGlyphStore.cs
```diff
@@ -104,34 +104,47 @@ private TextureUpload createTextureUpload(Character character, PageInfo page)
...
-            int readBytes = source.Read(readBuffer, 0, characterByteRegion);
+                while (!buffer.IsEmpty)
+                {
+                    int bytesRead = source.Read(buffer);
+                    buffer = buffer[bytesRead..];
 
-            Debug.Assert(readBytes == characterByteRegion);
+                    if (bytesRead == 0)
+                        throw new EndOfStreamException();
+                }
```
huoyaoyuan, line 127: I originally want to add a ReadToFulFill extension method, but in osu we generally use the extension methods only. Since it's planned for .NET 7, and after investigating the access pattern in game, I think it's not worthy for new extension methods.

osu.Framework/IO/Stores/RawCachingGlyphStore.cs
```diff
@@ -104,34 +104,47 @@ private TextureUpload createTextureUpload(Character character, PageInfo page)
             int pageWidth = page.Size.Width;
 
             int characterByteRegion = pageWidth * character.Height;
+            byte[] readBuffer = ArrayPool<byte>.Shared.Rent(characterByteRegion);
```
huoyaoyuan, line 107: Texture is somehow hot, so I choose to use array pool here.

bdach:
No benchmarking results whatsoever here, profile snapshots, anything? I'm definitely not going to attempt judging whether these optimisations actually optimise anything without looking at a prepared benchmark. I'll review the code, but I still want that to be backed by actual performance numbers.

osu.Framework/Audio/Callbacks/DataStreamFileProcedures.cs
```diff
@@ -27,55 +26,44 @@ public void Close(IntPtr user)
 
         public long Length(IntPtr user)
         {
-            if (dataStream == null) return 0;
+            if (!dataStream.CanSeek) return 0;
```
bdach, line 29: Not sure what these sorts of changes are doing in an "optimisation" PR but ok.

osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -133,8 +133,21 @@ internal TrackBass(Stream data, bool quick = false)
...
+                case MemoryStream _:
+                case UnmanagedMemoryStream _:
+                case AsyncBufferStream _:
+                    // Buffering memory stream is definitely unworthy.
+                    dataStream = data;
+                    break;
+
+                default:
+                    // It would be most likely a FileStream.
+                    // Is it benenitial to buffer over OS and .NET file buffering?
```
bdach, line 147: I'm not sure what this comment-question is supposed to be achieving. Is this apparently going to get replaced with net6, or what's the deal here? If it's getting replaced then this should just be a TODO saying so rather than a cryptic question.

osu.Framework/Extensions/StreamExtensions.cs
```diff
@@ -61,32 +64,19 @@ public static Task<byte[]> ReadAllBytesToArrayAsync(this Stream stream, Cancella
...
-            using (var ms = new MemoryStream(length))
+            while (!remaining.IsEmpty)
             {
-                while (remainingRead > 0)
-                {
-                    int read = stream.Read(buffer, 0, Math.Min(remainingRead, buffer.Length));
-                    if (read == 0)
-                        break;
+                int bytesRead = stream.Read(remaining);
+                remaining = remaining[bytesRead..];
```
bdach, line 73: Not sure we have universal agreement on adopting range syntax. Maybe best to call `.Slice()` for the time being.

peppy:
We do have some benchmark coverage of streams already, but there are some changes in this PR which I'm not sure are covered (which should be if they aren't) like the texture retrieval.

As above, would appreciate benchmark run results as a starting point.

osu.Framework/Audio/Callbacks/DataStreamFileProcedures.cs
huoyaoyuan: This is actually "correction". Non seekable stream throws for Length.

osu.Framework/Extensions/StreamExtensions.cs
huoyaoyuan: It is already used in some places.

huoyaoyuan:
Benchmark added for extension methods. Since the code patterns are similar, I'm only testing for sync path.

PR:
|              Method |  Length |         Mean |        Error |       StdDev |    Gen 0 |    Gen 1 |    Gen 2 |   Allocated |
|-------------------- |-------- |-------------:|-------------:|-------------:|---------:|---------:|---------:|------------:|
| ReadAllBytesToArray |     100 |     20.43 ns |     0.471 ns |     0.628 ns |   0.0153 |        - |        - |       128 B |
| ReadAllBytesToArray |   10000 |    438.81 ns |     8.662 ns |    16.481 ns |   1.1973 |   0.0424 |        - |    10,024 B |
| ReadAllBytesToArray | 1000000 | 95,501.93 ns | 2,396.354 ns | 6,990.276 ns | 175.5371 | 175.4150 | 175.4150 | 1,000,079 B |

master:
|              Method |  Length |         Mean |       Error |       StdDev |    Gen 0 |    Gen 1 |    Gen 2 | Allocated |
|-------------------- |-------- |-------------:|------------:|-------------:|---------:|---------:|---------:|----------:|
| ReadAllBytesToArray |     100 |     488.5 ns |     9.46 ns |     10.13 ns |   1.9798 |   0.0010 |        - |     16 KB |
| ReadAllBytesToArray |   10000 |   1,333.7 ns |    25.77 ns |     25.31 ns |   3.1643 |   0.1888 |        - |     26 KB |
| ReadAllBytesToArray | 1000000 | 158,213.4 ns | 5,924.38 ns | 17,187.69 ns | 202.1484 | 199.7070 | 199.7070 |    993 KB |

Extremely improves small streams, since the 16KB buffer allocation is massive. Note that 16KB array belongs to small object heap, and will add significant GC pressure. 
For large streams, the performance is ~1.5x by saving a copy.

huoyaoyuan:
The texture retrieval one is more like a clean up & correction. Reading stream in a loop is correct pattern for Stream contract. Moving the buffer from field to array pool helps enabling concurrency.

osu.Framework/Audio/Track/TrackBass.cs
huoyaoyuan: I wrote this before I constructed the .NET 6 rewrite. Yes, it could be a TODO here.

huoyaoyuan:
The title of this PR would be "clean up & optimisation". I wrote this mainly for clean up. Performance is improved by the way since redundant copy is eliminated.

osu.Framework/Extensions/StreamExtensions.cs
peppy: With the naming of the variable as `remaining`, this code is absolutely unreadable to my eyes. I'm not sure on a better name yet, but just putting that out there.

osu.Framework/IO/Stores/RawCachingGlyphStore.cs
```diff
@@ -104,34 +104,47 @@ private TextureUpload createTextureUpload(Character character, PageInfo page)
...
+                // consider to use System.IO.RandomAccess in .NET 6
+                source.Seek(pageWidth * character.Y, SeekOrigin.Begin);
+                var buffer = readBuffer.AsSpan(0, characterByteRegion);
 
-            source.Seek(pageWidth * character.Y, SeekOrigin.Begin);
-            int readBytes = source.Read(readBuffer, 0, characterByteRegion);
+                while (!buffer.IsEmpty)
+                {
+                    int bytesRead = source.Read(buffer);
+                    buffer = buffer[bytesRead..];
```
peppy, line 123: As above, I don't know how to read this. Having this variable called `buffer` sourced from `readBuffer` makes this code read very badly IMHO.

peppy (changes requested):

bdach:
Also just to go back to this for a second:

> Moving the buffer from field to array pool helps enabling concurrency.

I'm not sure how this is helping to "enable concurrency". `RawCachingGlyphStore.createTextureUpload()` is being called in one place only - namely `RawCachingGlyphStore.LoadCharacter()`, which already takes a lock on `pageLookup`. How is concurrency relevant in this context?

peppy:
For what it's worth I'm currently running benchmarks on the font store as I am also not sure whether the array pool usage is an improvement without it.

osu.Framework/Audio/Callbacks/DataStreamFileProcedures.cs
huoyaoyuan: I think it's better to mark it `internal`. This class is quite tighten with BASS, and not useful elsewhere.

peppy:
Master:


|                   Method | FetchCount |           Mean |        Error |       StdDev | Ratio |     Gen 0 |   Gen 1 |   Gen 2 | Allocated |
|------------------------- |----------- |---------------:|-------------:|-------------:|------:|----------:|--------:|--------:|----------:|
| BenchmarkRawCachingReuse |          1 |       901.9 us |      7.68 us |      7.18 us |  0.08 |   43.9453 | 35.1563 | 28.3203 |    277 KB |
|      BenchmarkRawCaching |          1 |    11,641.7 us |    161.12 us |    142.83 us |  1.00 |   15.6250 |       - |       - |    371 KB |
|                          |            |                |              |              |       |           |         |         |           |
| BenchmarkRawCachingReuse |         10 |     2,947.5 us |     14.58 us |     12.17 us |  0.08 |   74.2188 | 58.5938 | 54.6875 |    417 KB |
|      BenchmarkRawCaching |         10 |    36,140.4 us |    573.19 us |    536.16 us |  1.00 |         - |       - |       - |    746 KB |
|                          |            |                |              |              |       |           |         |         |           |
| BenchmarkRawCachingReuse |        100 |    16,526.8 us |     75.99 us |     67.36 us |  0.24 |   62.5000 | 31.2500 | 31.2500 |    581 KB |
|      BenchmarkRawCaching |        100 |    68,041.1 us |  1,218.80 us |  1,496.80 us |  1.00 |         - |       - |       - |  1,089 KB |
|                          |            |                |              |              |       |           |         |         |           |
| BenchmarkRawCachingReuse |       1000 |   157,595.5 us |  1,955.15 us |  1,828.85 us |  0.72 |         - |       - |       - |  2,391 KB |
|      BenchmarkRawCaching |       1000 |   219,043.5 us |  1,633.69 us |  1,528.15 us |  1.00 |         - |       - |       - |  2,871 KB |
|                          |            |                |              |              |       |           |         |         |           |
| BenchmarkRawCachingReuse |      10000 | 1,543,063.2 us | 14,221.47 us | 13,302.77 us |  0.94 | 1000.0000 |       - |       - | 15,186 KB |
|      BenchmarkRawCaching |      10000 | 1,642,248.3 us |  7,982.10 us |  7,466.46 us |  1.00 | 1000.0000 |       - |       - | 15,664 KB |


This PR:


|                   Method | FetchCount |           Mean |        Error |      StdDev | Ratio |     Gen 0 |  Gen 1 | Allocated |
|------------------------- |----------- |---------------:|-------------:|------------:|------:|----------:|-------:|----------:|
| BenchmarkRawCachingReuse |          1 |       882.3 us |      4.16 us |     3.69 us |  0.08 |    9.7656 | 2.9297 |    114 KB |
|      BenchmarkRawCaching |          1 |    11,405.3 us |    118.77 us |   111.10 us |  1.00 |   15.6250 |      - |    205 KB |
|                          |            |                |              |             |       |           |        |           |
| BenchmarkRawCachingReuse |         10 |     2,867.9 us |     29.81 us |    26.42 us |  0.08 |   11.7188 | 3.9063 |    152 KB |
|      BenchmarkRawCaching |         10 |    34,729.9 us |    251.93 us |   223.33 us |  1.00 |         - |      - |    484 KB |
|                          |            |                |              |             |       |           |        |           |
| BenchmarkRawCachingReuse |        100 |    16,393.2 us |    129.55 us |   121.18 us |  0.25 |         - |      - |    305 KB |
|      BenchmarkRawCaching |        100 |    64,976.6 us |    317.63 us |   297.12 us |  1.00 |         - |      - |    828 KB |
|                          |            |                |              |             |       |           |        |           |
| BenchmarkRawCachingReuse |       1000 |   157,922.0 us |  1,578.43 us | 1,476.47 us |  0.75 |         - |      - |  2,130 KB |
|      BenchmarkRawCaching |       1000 |   210,225.8 us |  2,191.94 us | 1,830.37 us |  1.00 |         - |      - |  2,606 KB |
|                          |            |                |              |             |       |           |        |           |
| BenchmarkRawCachingReuse |      10000 | 1,543,722.5 us | 11,257.69 us | 9,979.65 us |  0.94 | 1000.0000 |      - | 14,925 KB |
|      BenchmarkRawCaching |      10000 | 1,647,745.3 us | 11,297.06 us | 9,433.56 us |  1.00 | 1000.0000 |      - | 15,403 KB |

osu.Framework/IO/Stores/RawCachingGlyphStore.cs
huoyaoyuan: I feel it's better to encapsulate the loop reading pattern into a helper method, but the use cases are too few now. Maybe an internal extension method?

peppy: Might help, give it a try.

peppy (dismissed):
Seems fine now

osu.Framework/IO/AsyncBufferStream.cs
```diff
@@ -192,7 +192,7 @@ public override int Read(byte[] buffer, int offset, int count)
                 }
             }
 
-            Array.Copy(data, position, buffer, offset, amountBytesToRead);
+            data.AsSpan(position).CopyTo(buffer[..amountBytesToRead]);
```
frenzibyte, line 195: `amountBytesToRead` can be smaller than `data.Length - position`, which means that the destination span may become smaller than the source span, which is an [invalid operation](https://docs.microsoft.com/en-us/dotnet/api/system.span-1.copyto?view=net-6.0#exceptions).
```suggestion
            data.AsSpan(position, amountBytesToRead).CopyTo(buffer);
```

frenzibyte (changes requested):
Had a read through this and so far it looks good (except for a seemingly looking oversight).

osu.Framework/IO/AsyncBufferStream.cs
huoyaoyuan: You're right. I did not test this.

bdach (approved):

osu.Framework/Audio/Callbacks/DataStreamFileProcedures.cs
```diff
@@ -3,17 +3,16 @@
...
+
+#nullable enable
 
 namespace osu.Framework.Audio.Callbacks
 {
     /// <summary>
     /// Implementation of <see cref="IFileProcedures"/> that supports reading from a <see cref="Stream"/>.
     /// </summary>
-    public class DataStreamFileProcedures : IFileProcedures
+    internal class DataStreamFileProcedures : IFileProcedures
```
frenzibyte, line 14: This is regressing osu!-side usage for checking too short audio files in editor:

CleanShot 2022-05-04 at 15 01 37@2x

```suggestion
    public class DataStreamFileProcedures : IFileProcedures
```

frenzibyte (changes requested):
Looks good except for the note below, which I found as I was checking game-side with these changes.

osu.Framework/Audio/Callbacks/DataStreamFileProcedures.cs
bdach: I'd just revert the modifier change honestly.

huoyaoyuan: Should we add a null check then, probably just in constructor?

frenzibyte: `?? throw new ArgumentNullException` makes sense given that the game is not fully NRT-enabled.

frenzibyte (approved):

peppy (approved):

smoogipoo (commented):
I've done a code-only pass through this and everything looks correct. I won't be able to merge this within the next couple of days as I'd like to test in-game with sound, but if others have tested then feel free to merge.

peppy:
I've tested this on macOS, no issues here.

smoogipoo (approved):

PR #5141, opened 2022-04-29
Properly block SDL IME results sent as multiple text events + upstream SDL fixes

Susko3:
Depends on:
- [x] SDL bump (for https://github.com/libsdl-org/SDL/pull/5398)

Closes https://github.com/ppy/osu/issues/17231.

SDL will now send multiple `SDL_TEXTINPUT` events when the result is over 32 bytes. Here that is handled by blocking all text events that come in the same event loop. But because valid non-IME text events can arrive in the same event loop, we must check the text matches.

Eg. on Korean IME, pressing `t`, `h`, then `1`: (only `"소"` is part of the composition, `"1"` is just regular text input)
```cs
// first event poll, pressing 't'
INFO: SDL EVENT: SDL_TEXTEDITING (timestamp=78142 windowid=1 text='ㅅ' start=1 length=0)

// second, pressing 'h'
INFO: SDL EVENT: SDL_TEXTEDITING (timestamp=78208 windowid=1 text='소' start=1 length=0)

// third, pressing '1'
// notice the two SDL_TEXTINPUTs
INFO: SDL EVENT: SDL_TEXTEDITING (timestamp=78736 windowid=1 text='' start=0 length=0)
INFO: SDL EVENT: SDL_TEXTEDITING (timestamp=78737 windowid=1 text='' start=0 length=0)
INFO: SDL EVENT: SDL_TEXTINPUT (timestamp=78737 windowid=1 text='소')
INFO: SDL EVENT: SDL_KEYDOWN (timestamp=78737 windowid=1 state=pressed repeat=false scancode=30 keycode=49 mod=4096)
INFO: SDL EVENT: SDL_TEXTINPUT (timestamp=78737 windowid=1 text='1')
```


peppy (approved):
Not tested, but code looks fine so let's get this merged for user testing.

PR #5144, opened 2022-04-30
Fix iOS text key shortcuts no longer working

### LINKED ISSUE #5143: Hardware keyboard no longer activates text shortcuts on iOS
This was a shortcoming from #5106 after going back and through on solving this, should've been blocked.

Currently (post-#5106), there are three different types of keyboard handlers:
 1. [`IOSTextInput.handleShouldChangeCharacters`](https://github.com/ppy/osu-framework/blob/ee4ae587ca9a94a2b5fc470d59419b970e070686/osu.Framework.iOS/Input/IOSTextInput.cs#L20-L24), activated when focusing textbox, receiving any writable characters from the "hidden text field".
 2. [`IOSTextFieldKeyboardHandler`](https://github.com/ppy/osu-framework/blob/ee4ae587ca9a94a2b5fc470d59419b970e070686/osu.Framework.iOS/Input/IOSTextFieldKeyboardHandler.cs), activated when focusing textbox, handling any keys which mutate the text or shortcuts (i.e. "key commands").
 3. [`IOSHardwareKeyboardHandler`](https://github.com/ppy/osu-framework/blob/364f5eba43cee412768030c1f7678386b28ba432/osu.Framework.iOS/Input/IOSHardwareKeyboardHandler.cs#L14), activated when not focusing any textbox, handling raw key presses.

Previously, `IOSHardwareKeyboardHandler` was automatically switched to when the software keyboard is less than a specific height, to allow for the hardware keyboard to still trigger raw key events. That worked, but the keyboard height check was quite an arbitrary choice and caused issues when using an Apple Pencil (see https://github.com/ppy/osu/issues/17513).

Now with #5106, `IOSHardwareKeyboardHandler` is no longer switched to and `IOSTextFieldKeyboardHandler` is used on all cases when the textbox is focused, and that works well on a hardware keyboard, but does not handle text shortcuts (or any key combination from the hardware keyboard).

I've tried looking for any workaround that doesn't involve reverting #5106, one of which overrides the [standard text shortcut events](https://developer.apple.com/documentation/uikit/uiresponderstandardeditactions?language=objc) and translating them to o!f keyboard input, but for some reason I can't get that to work at all. Possibly due to the way `HiddenTextField` works right now (placeholder text and stuff).

The only way I can see forward is to revert #5106 and go with the alternative solution of checking the position of the keyboard instead, which is still arbitrary, but about the only way possible to support both hardware and software keyboards together.

---

Going forward, I can see the entirety of `HiddenTextField` moving away in favour of a custom `UITextInput` which responds to text input activations and updates its text accordingly, in order to support software keyboard properly on iOS without having placeholder text around.

frenzibyte:
- Closes #5143 

Gladly this didn't require reverting #5106. Done by keeping the `IOSHardwareKeyboardHandler` active all the time, and avoid duplicating input if handled by `IOSTextFieldKeyboardHandler`.

This can be best reviewed commit-by-commit, to explain how this was approached and what it has changed.

I have tested this on an iOS simulator and confirmed text shortcuts work now, but as I was testing this, for some reason some keys get stuck in a pressed state, and I have no clue how to tell if it's due to the game running poorly due to OpenGL or otherwise.

@peppy please check that this is working alright for you, and that pasting works properly from both hardware keyboard and software keyboard, and that no keys actually get stuck in pressed state (if that occurs, pressing left/right or triggering any framework platform action will no longer work).

If for some reason something doesn't work then it's probably best to revert #5106 for now and call it a day.

### DIFF
--- a/osu.Framework.iOS/Input/IOSHardwareKeyboardHandler.cs
+++ b/osu.Framework.iOS/Input/IOSHardwareKeyboardHandler.cs
@@ -16,10 +16,10 @@ namespace osu.Framework.iOS.Input
 
         public override bool Initialize(GameHost host)
         {
-            if (!(UIApplication.SharedApplication is GameUIApplication game) || !(host is IOSGameHost iosHost))
-                return false;
+            var iosApp = (GameUIApplication)UIApplication.SharedApplication;
+            var iosHost = (IOSGameHost)host;
 
-            game.KeyEvent += (keyCode, isDown) =>
+            iosApp.KeyEvent += (keyCode, isDown) =>
             {
                 if (!IsActive)
                     return;
--- a/osu.Framework.iOS/Input/IOSHardwareKeyboardHandler.cs
+++ b/osu.Framework.iOS/Input/IOSHardwareKeyboardHandler.cs
@@ -12,8 +12,7 @@ namespace osu.Framework.iOS.Input
 {
     public class IOSHardwareKeyboardHandler : InputHandler
     {
-        internal bool KeyboardActive = true;
-        public override bool IsActive => KeyboardActive;
+        public override bool IsActive => true;
 
         public override bool Initialize(GameHost host)
         {
--- a/osu.Framework.iOS/Input/IOSHardwareKeyboardHandler.cs
+++ b/osu.Framework.iOS/Input/IOSHardwareKeyboardHandler.cs
@@ -16,13 +16,19 @@ namespace osu.Framework.iOS.Input
 
         public override bool Initialize(GameHost host)
         {
-            if (!(UIApplication.SharedApplication is GameUIApplication game))
+            if (!(UIApplication.SharedApplication is GameUIApplication game) || !(host is IOSGameHost iosHost))
                 return false;
 
             game.KeyEvent += (keyCode, isDown) =>
             {
-                if (IsActive && keyMap.ContainsKey(keyCode))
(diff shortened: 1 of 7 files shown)


peppy:
I've tested this and the behaviour is definitely fixed. Can't repro anything getting stuck.

osu.Framework.iOS/Input/IOSHardwareKeyboardHandler.cs
```diff
@@ -12,18 +12,23 @@ namespace osu.Framework.iOS.Input
...
     public class IOSHardwareKeyboardHandler : InputHandler
     {
-        internal bool KeyboardActive = true;
-        public override bool IsActive => KeyboardActive;
+        public override bool IsActive => true;
 
         public override bool Initialize(GameHost host)
         {
-            if (!(UIApplication.SharedApplication is GameUIApplication game))
+            if (!(UIApplication.SharedApplication is GameUIApplication game) || !(host is IOSGameHost iosHost))
```
peppy, line 19: What's the case where `IOSGameHost` is not the host here?

frenzibyte: Just matching with the `UIApplication.SharedApplication` one, gonna direct cast.

peppy (approved):

PR #5147, opened 2022-05-01
Improve iOS text input deactivation to not require long delay

frenzibyte:
Something I found which is outdated by now given the refactors of `TextInputSource`, and which bugged me with its delay on hide (almost feels like the game is slow enough to hide the keyboard). Also resolves the TODO item at the same time.

Before:

https://user-images.githubusercontent.com/22781491/166139753-88cde4d4-786d-4875-9cd2-fd88ff7e7cce.mov

After:

https://user-images.githubusercontent.com/22781491/166139751-2803a330-4200-4660-ac61-44fe7250f9dd.mov

Relying on the `InputThread` scheduler may feel a bit weird, but it is about the only managed way for letting something wait one frame before it is invoked, and able to be cancelled by other operations beforehand.

### DIFF
--- a/osu.Framework.iOS/Input/IOSTextInput.cs
+++ b/osu.Framework.iOS/Input/IOSTextInput.cs
@@ -3,6 +3,7 @@
 
 using Foundation;
 using osu.Framework.Input;
+using osu.Framework.Threading;
 
 namespace osu.Framework.iOS.Input
 {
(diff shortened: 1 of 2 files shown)


osu.Framework.iOS/Input/IOSTextInput.cs
```diff
@@ -23,27 +24,36 @@ private void handleShouldChangeCharacters(NSRange range, string text)
...
         protected override void DeactivateTextInput()
         {
             view.KeyboardTextField.HandleShouldChangeCharacters -= handleShouldChangeCharacters;
-            view.KeyboardTextField.UpdateFirstResponder(false);
             host.TextFieldHandler.KeyboardActive = false;
+
+            // text input may be deactivated and activated at the same frame, as a result of switching between textboxes.
+            // this could cause the software keyboard to flicker, scheduling the operation to the next frame will help.
+            // todo: that should probably be improved framework-side instead, but this should do for now.
+            resignDelegate = host.InputThread.Scheduler.Add(() => view.KeyboardTextField.ResignFirstResponder());
```
Susko3, line 56: This happens on Android as well.
Ideally, `TextInputSource` logic should only check `activationCounter` and call `DeactivateTextInput()` at the end of an update frame, once the textboxes have exchanged focus.

Not sure what's the best way to schedule something to run at the end of a frame.

frenzibyte: Yeah, I've had the thought about fixing this framework-side itself until I realized `TextInputSource` was cached all the way from `GameHost`, which makes things potentially tricky/messy.

A first step may be to change it to a `Component`, and include it as part of the root scene in some respect to allow it receiving a scheduler.

Susko3: Actually works really well. Needs to be added before the `Game` so that it's loaded before it.



```diff
diff --git a/osu.Framework/Platform/GameHost.cs b/osu.Framework/Platform/GameHost.cs
index b4a5837e7..1af15d156 100644
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -724,7 +724,6 @@ public void Run(Game game)
                 }

                 Dependencies.CacheAs(readableKeyCombinationProvider = CreateReadableKeyCombinationProvider());
-                Dependencies.CacheAs(CreateTextInput());

                 ExecutionState = ExecutionState.Running;
                 threadRunner.Start();
@@ -903,7 +902,12 @@ public void ResetInputHandlers()
         private void bootstrapSceneGraph(Game game)
         {
             var root = game.CreateUserInputManager();
-            root.Child = new PlatformActionContainer
+
+            var textInput = CreateTextInput();
+            root.Add(textInput);
+            Dependencies.CacheAs(textInput);
+
+            root.Add(new PlatformActionContainer
             {
                 Child = new FrameworkActionContainer
                 {
@@ -913,7 +917,7 @@ private void bootstrapSceneGraph(Game game)
                         Child = game
                     }
                 }
-            };
+            });

             Dependencies.Cache(root);
             Dependencies.CacheAs(game);
```

Could be really nice for debugging, check fields etc.

image

frenzibyte: Could cache it first then add it later, similar to how most components do it in `OsuGame`/`OsuGameBase`.

peppy:
Is this PR superseded by #5150?

frenzibyte:
For the most part, if #5150 gets merged then all that's left is just cleaning up iOS here to not have any delay and becoming/resigning first responder directly instead.

peppy:
So that PR should be merged first?

frenzibyte:
Yeah, this originally was just working around the issue which that PR intends to actually solve. So since it has been PR'd, I will wait for it to be merged and update this accordingly.

frenzibyte:
This is just straight up fixed in SDL.

PR #5149, opened 2022-05-01
Replace `NET6_0` with `NET6_0_OR_GREATER` instead

frenzibyte:
This has became an issue on Xamarin projects when using Rider, as it assumes `NET6_0` is defined in `AssemblyInfo.cs` for some reason, while works correctly with `NET6_0_OR_GREATER`.

https://user-images.githubusercontent.com/22781491/166141013-a84e2c83-14b5-48c6-95be-92e44ad6e9c7.mp4

Best to migrate to `NET6_0_OR_GREATER` anyways since it's the correct way of checking.

Note that regardless of this change, ReSharper on `AssemblyInfo.cs` in the desktop project will for some reason think neither `NET6_0` nor `NET6_0_OR_GREATER` is defined (only `NETSTANDARD2_1`). Probably a ReSharper bug.

### DIFF
--- a/osu.Framework.Tests/Visual/Input/TestSceneTabletInput.cs
+++ b/osu.Framework.Tests/Visual/Input/TestSceneTabletInput.cs
@@ -1,7 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-#if NET6_0
+#if NET6_0_OR_GREATER
 using System.Linq;
 using osu.Framework.Allocation;
 using osu.Framework.Graphics;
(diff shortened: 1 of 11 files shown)


huoyaoyuan:
I'd suggest to use version-less `#if !NETSTANDARD` if Rider supports.

It's very unlikely that we would multi-targeting different .NET Core versions in the future, so the versioned ones looks too verbose.

Version less `NET` looks a bit confusing. `NETSTANDARD` is OK.

They are defined by the SDK. I don't know if Rider supports it correctly.

frenzibyte:
Using `#if !NETSTANDARD` looks to work on all cases, except for this one test scene which I have no idea why it's failing...

CleanShot 2022-05-01 at 15 23 20@2x

Meanwhile `#if NET6_0_OR_GREATER` works there as well:

CleanShot 2022-05-01 at 15 25 13@2x

peppy (approved):

smoogipoo (approved):

PR #5150, opened 2022-05-01
Fix `TextInputSource` reactivating when changing focus between `TextBox`es

Susko3:
- Discussed in https://github.com/ppy/osu-framework/pull/5147#discussion_r862450084.

The broken scenario:
- textbox `A` has focus, and textbox `B` is clicked
- `UserInputManager` updates and propagates the focus event:
- in `A`: `OnFocusLoss` &rarr;
  - `textInput.Deactivate()`
  - `activationCounter: 1 -> 0`
  - native input deactivated
- in `B`: `OnFocus` &rarr;
  - `textInput.Activate()`
  - `activationCounter: 0 -> 1`
  - native input activated

~~In the fix, the deactivate is scheduled, so we must be careful about activating in `B.OnFocus` (calling `ActivateTextInput` twice is bad), hence the `nativeTextInputActive` bool.~~

~~Scheduling the deactivate on the `UpdateThread` doesn't really work, as that will create a one frame delay (and passing `GameHost` to `TextInputSource` is questionable). Instead `TextInputSource` was made a `Component` so the drawable scheduler is used, allowing the update to happen right after `UserInputManager` is updated.~~

If this goes well, we could make `ReadableKeyCombinationProvider`, `Clipboard`, etc. components, hence the `addComponent()` function. Also for the future, `SDL2DesktopWindowTextInput` and co. could resolve `SDL2DesktopWindow` via BDL.


bdach:
I don't like this solution due to schedules being used, and would like to propose the following alternative:

* `FocusLostEvent` would receive a property indicating what the next focus target is (whether focus is to be handed over to another drawable, or being dropped immediately due to there being no other viable target)
* Similarly, `FocusEvent` would receive a property indicating what the previous focus target was
* If `TextBox` receives a `FocusLostEvent` that has a non-null next focus target that is a textbox, it doesn't deactivate text input
* If `TextBox` receives a `FocusEvent` that has a non-null previous focus target that is a textbox, it doesn't activate text input

frenzibyte:
Another idea I had in mind is to make `TextInputSource` only update the activation counter, but schedule the operation to the input thread for actually interacting with the window/view. But that doesn't work well on multi-threaded since the input thread could be executing said operation before the update thread sends focus event to the new textbox.

I feel @bdach's proposal is amicable to avoid dealing with threading and delaying entirely. The only concern I had about it is when we could potentially use `ITextInputSource` for something other than a `TextBox`, but I think that's very unlikely to be the case. May deserve an inline comment at best.

Susko3:
@bdach's approach works really well. I've pushed the changes to this branch.
`TextInputSource` no longer needs to be a `Component`, I've left it as-is for now.

To note, this won't work with osu!-side `FocusedTextBox`, as the focus isn't exchanged directly, but with a `null` focus in between:

```cs
2022-05-01 21:05:24 [debug]: Pressed (ToggleSettings) handled by ToolbarSettingsButton.
2022-05-01 21:05:24 [debug]: KeyDownEvent(O, False) handled by GlobalActionContainer.
2022-05-01 21:05:24 [verbose]: Focus contention triggered by SettingsOverlay.
2022-05-01 21:05:24 [verbose]: scheduling deactivate... activationCounter == 0
2022-05-01 21:05:24 [debug]: Focus changed from SeekLimitedSearchTextBox to nothing.
2022-05-01 21:05:24 [debug]: Focus changed from nothing to SeekLimitedSearchTextBox.
2022-05-01 21:05:24 [verbose]: abort deactivate, activationCounter > 0
```

But this is not a problem as a `FocusedTextBox` [behaves as a regular text box on mobile](https://github.com/ppy/osu/blob/726b49fdf3e170d67a9d07e54de12e1df0c741f0/osu.Game/Graphics/UserInterface/FocusedTextBox.cs#L24-L32), not taking focus eagerly. 

> The only concern I had about it is when we could potentially use `ITextInputSource` for something other than a `TextBox`, but I think that's very unlikely to be the case. May deserve an inline comment at best.

The logic is contained entirely within `TextBox`. Changing focus from a `TextBox` to some other drawable will use the normal deactivate/activate flow.

osu.Framework/Input/Events/FocusEvent.cs
```diff
@@ -1,18 +1,27 @@
...
+    /// An event representing that a drawable gained the focus.
     /// </summary>
     public class FocusEvent : UIEvent
     {
-        public FocusEvent(InputState state)
+        /// <summary>
+        /// The <see cref="Drawable"/> that has lost focus, or <c>null</c> if nothing was previously focused.
+        /// </summary>
+        [CanBeNull]
+        public readonly Drawable PreviousFocus;
```
peppy, line 19:
```suggestion
        public readonly Drawable PreviouslyFocused;
```

Think this reads better.

peppy (commented):
Structure seems fine overall?

osu.Framework/Input/Events/FocusEvent.cs
bdach: I tend to agree, and would extend this to renaming `FocusLostEvent.NextFocus -> FocusLostEvent.NextFocused` for parity.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -1224,30 +1224,40 @@ protected override void OnFocus(FocusEvent e)
...
                 textInput.EnsureActivated(AllowIme);
                 return;
             }
 
-            textInput.Activate(AllowIme);
+            // TextBox has special handling of text input activation when focus is changed directly from one TextBox to another.
+            // We don't deactivate and activate, but instead keep text input active during the focus handoff, so that virtual keyboards on phones don't flicker.
+
+            if (previousFocusWasTextBox)
+                textInput.EnsureActivated(AllowIme);
```
bdach, line 1239: Is this `.EnsureActivated()` call ever needed or just insurance of sorts? Something IME-related maybe?

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -1224,30 +1224,40 @@ protected override void OnFocus(FocusEvent e)
         /// </summary>
         private bool textInputBound;
 
-        private void bindInput()
+        private void bindInput(bool previousFocusWasTextBox)
```
bdach, line 1227: I really don't like `bool` input arguments as they're a bad sign that this method may be doing a smidge too much. Maybe the text input activations here and in `unbindInput()` could be inlined into their calling methods instead? Just to avoid passing down the condition as a boolean. Not sure.

osu.Framework/Graphics/UserInterface/TextBox.cs
Susko3: This call is so that the up-to-date `AllowIme` value is correctly propagated to the native text input when switching from a textbox that disallows IME to a one that allows it (or vice versa). This is tested in the `false` [testcase](https://github.com/ppy/osu-framework/pull/5150/files#diff-640d8e3c190c5a9daa78fd193a6cf1ca8ed6e23671cf4406b9e5c65148248c5bR418).

osu.Framework/Graphics/UserInterface/TextBox.cs
frenzibyte: Can agree, the boolean is only making the method feel more complicated.

osu.Framework/Graphics/UserInterface/TextBox.cs
frenzibyte: It should still be redundant over using `Activate` though, no? looking at the implementation of `Activate`, it should do exactly as what this is doing, but at a lower level:

CleanShot 2022-05-04 at 16 22 41@2x

(at this point, the previously focused `TextBox` is still keeping the text input activated, so `activationCounter` must be 1 before the increment)

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -1224,30 +1224,40 @@ protected override void OnFocus(FocusEvent e)
...
         {
             if (textInputBound)
             {
                 textInput.EnsureActivated(AllowIme);
                 return;
             }
 
-            textInput.Activate(AllowIme);
+            // TextBox has special handling of text input activation when focus is changed directly from one TextBox to another.
+            // We don't deactivate and activate, but instead keep text input active during the focus handoff, so that virtual keyboards on phones don't flicker.
```
frenzibyte, line 1236: I'm wondering if this comment is necessary, feels obvious enough. Probably fine.

osu.Framework/Graphics/UserInterface/TextBox.cs
bdach: `Activate()` cannot be used here. You missed the part where that method always increments `activationCounter`. You'd break deactivation (upon every handoff of the focus, the counter would increment by 1 rather than staying at 1).

Susko3: Calling `Activate` will increment the counter. We can't increment the counter, as that will lead to the counter going to infinity since we never decrement it.

bdach:
For the record I'm mostly ok with this pull at this stage but wanted to run it past android just to see how it works, which is mostly why I haven't approved it yet.

bdach:
I've tested this today, and there is a marked improvement in behaviour on one of my test devices with this change applied, so this is good to go as far as I'm concerned. Probably going to wait for a second approval though.

bdach (approved):

smoogipoo (approved):

PR #5151, opened 2022-05-02
Move caching specifications of markdown interfaces to be attached to the interfaces

peppy:
As seen in https://github.com/ppy/osu/pull/18049, it can be easy to forget adding the caching of these interfaces. I believe all usages of the interfaces intend to also cache (as that is the main purpose of the interface existing).

Rendering still seems to work okay.

Closes https://github.com/ppy/osu/issues/18039.

### DIFF
--- a/osu.Framework/Graphics/Containers/Markdown/IMarkdownTextComponent.cs
+++ b/osu.Framework/Graphics/Containers/Markdown/IMarkdownTextComponent.cs
@@ -1,10 +1,12 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using osu.Framework.Allocation;
 using osu.Framework.Graphics.Sprites;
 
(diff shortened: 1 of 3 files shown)


frenzibyte (approved):

bdach:
This is a scarier change than the local game-side fix, but I poked about the wiki overlay with a framework checkout and I can't find anything more broken than what was already there...

Think it's maybe best to just discard the game-side fix, given that this supersedes it?

bdach (approved):

peppy:
Sure, if this is preferred. Wanted a second opinion on whether we want to cache everywhere, since I am not 100% familiar with the wiki structure. If you think it's okay then let's go with this alone, to save duplicating the fix.

bdach:
I think this change definitely makes sense from a standpoint of how markdown documents work.

Browsing about I did notice two xmldocs that were a bit stale with this change now, which I've gone ahead and updated. Please check that you're ok with the wording.

bdach (approved):

peppy:
Reads fine 👍 .

frenzibyte (approved):

PR #5152, opened 2022-05-02
Adjust `GameExitTest` to attempt to understand CI failures

peppy:
I've gone through the testing procedure and code and cannot understand nor replicate how this can happen. Might be able to understand better with these modifications?

Reference:

https://github.com/ppy/osu-framework/runs/6254153200?check_suite_focus=true
https://github.com/ppy/osu-framework/runs/6240530753?check_suite_focus=true
https://github.com/ppy/osu-framework/runs/6227330132?check_suite_focus=true

Interested if anyone else can see an issue with the testing methodology.

### DIFF
--- a/osu.Framework.Tests/Platform/GameExitTest.cs
+++ b/osu.Framework.Tests/Platform/GameExitTest.cs
@@ -35,6 +35,8 @@ namespace osu.Framework.Tests.Platform
             gameCreated.Wait(timeout);
             Assert.IsTrue(game.BecameAlive.Wait(timeout));
 
+            Assert.That(() => host.ExecutionState, Is.EqualTo(ExecutionState.Running));
+
             // block game from exiting.
             game.BlockExit.Value = true;
             // `RequestExit()` should return true.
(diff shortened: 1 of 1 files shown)


frenzibyte (approved):

bdach:
As the unfortunate original author of this test, I am also at a loss as to why these failures are occurring. I don't remember this test failing once locally, even though I did stress-test it quite a bit.

I haven't seen the outright execution state assertion failures, but I do recall seeing the `PerformanceMonitor` - which I was confounded by because the performance monitor isn't even supposed to be running during tests except maybe like one test...

Anyways, I've pushed one change that removes a few delegates in `Assert.That` calls, which were left behind in #5111 (if you `Assert.That` with a timeout, you need to pass a `Func`, but there is no timeout anymore, so no point).

bdach (approved):
no harm in more assertions, i suppose

bdach:
A plausible explanation of how this can fail [has been presented on discord](https://discord.com/channels/188630481301012481/589331078574112768/970630017476362250), but I'm merging this anyway because there's no real reason not to (and to avoid any possible conflicts if that fix lands, I guess).

PR #5154, opened 2022-05-03
Refactor `SearchContainer` filtering operation and fix potential flaws

frenzibyte:
- Closes https://github.com/ppy/osu/issues/18003

I've initially intended to look at just adding guards against non-present drawables, but the current logic for filtering out drawables was quite convoluted that I looked at first refactoring it with flexibility in mind, then adding the necessary guards to solve the issues pointed out in ppy/osu#18003.

### DIFF
--- a/osu.Framework/Graphics/Containers/IFilterable.cs
+++ b/osu.Framework/Graphics/Containers/IFilterable.cs
@@ -3,15 +3,18 @@
 
 namespace osu.Framework.Graphics.Containers
 {
+    /// <summary>
+    /// Represents a component which can be filtered out on non-matching search terms.
+    /// </summary>
+    /// <remarks>
+    /// Hiding <see cref="IFilterable"/>s for purposes other than filtering require being wrapped in a
+    /// <see cref="VisibilityContainer"/> with its visibility state set to <see cref="Visibility.Hidden"/>.
+    /// </remarks>
     public interface IFilterable : IHasFilterTerms
     {
         /// <summary>
         /// Whether the current object is matching (ie. visible) given the current filter criteria of a parent.
         /// </summary>
-        /// <remarks>
-        /// It is recommended to not let this property update the <see cref="Drawable.Alpha"/> state of this filterable, as to not conflict with hiding it via other means.
-        /// But instead, override <see cref="Drawable.IsPresent"/> to respect the filter state, and invalidate <see cref="Invalidation.Presence"/> accordingly.
-        /// </remarks>
         bool MatchingFilter { set; }
 
         /// <summary>
--- a/osu.Framework/Graphics/Containers/IFilterable.cs
+++ b/osu.Framework/Graphics/Containers/IFilterable.cs
@@ -6,10 +6,6 @@ namespace osu.Framework.Graphics.Containers
     /// <summary>
     /// Represents a component which can be filtered out on non-matching search terms.
     /// </summary>
-    /// <remarks>
-    /// Hiding <see cref="IFilterable"/>s for purposes other than filtering require being wrapped in a
-    /// <see cref="VisibilityContainer"/> with its visibility state set to <see cref="Visibility.Hidden"/>.
-    /// </remarks>
     public interface IFilterable : IHasFilterTerms
     {
         /// <summary>
--- a/osu.Framework/Graphics/Containers/IFilterable.cs
+++ b/osu.Framework/Graphics/Containers/IFilterable.cs
@@ -8,6 +8,10 @@ namespace osu.Framework.Graphics.Containers
         /// <summary>
         /// Whether the current object is matching (ie. visible) given the current filter criteria of a parent.
         /// </summary>
+        /// <remarks>
+        /// It is recommended to not let this property update the <see cref="Drawable.Alpha"/> state of this filterable, as to not conflict with hiding it via other means.
+        /// But instead, override <see cref="Drawable.IsPresent"/> to respect the filter state, and invalidate <see cref="Invalidation.Presence"/> accordingly.
+        /// </remarks>
         bool MatchingFilter { set; }
 
         /// <summary>
--- a/osu.Framework/Graphics/Containers/SearchContainer.cs
+++ b/osu.Framework/Graphics/Containers/SearchContainer.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using osu.Framework.Caching;
-using osu.Framework.Extensions.IEnumerableExtensions;
 
 namespace osu.Framework.Graphics.Containers
 {
@@ -87,28 +86,53 @@ namespace osu.Framework.Graphics.Containers
         private void performFilter()
         {
             string[] terms = (searchTerm ?? string.Empty).Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
-            Children.OfType<IFilterable>().ForEach(child => match(child, terms, terms.Length > 0, allowNonContiguousMatching));
+            matchSubTree(this, terms, terms.Length > 0, allowNonContiguousMatching);
         }
 
-        private static bool match(IFilterable filterable, IEnumerable<string> searchTerms, bool searchActive, bool nonContiguousMatching)
+        /// <summary>
+        /// Performs a tree traversal from the given node to filter out drawables which do not match the specified search terms.
+        /// </summary>
+        /// <param name="drawable">The root drawable of this subtree.</param>
+        /// <param name="searchTerms">The search terms to filter drawables against.</param>
+        /// <param name="searchActive">Whether there are search terms usable for filtering drawables.</param>
+        /// <param name="nonContiguousMatching">Whether the matching algorithm should be non-contiguous, allowing for potentially non-matching characters to exist between matches.</param>
+        private static bool matchSubTree(Drawable drawable, IReadOnlyList<string> searchTerms, bool searchActive, bool nonContiguousMatching)
         {
-            //Words matched by parent is not needed to match children
-            string[] childTerms = searchTerms.Where(term =>
-                !filterable.FilterTerms.Any(filterTerm =>
-                    checkTerm(filterTerm, term, nonContiguousMatching))).ToArray();
+            bool matching = match(drawable, searchTerms, nonContiguousMatching, out var nonMatchingTerms);
 
-            bool matching = childTerms.Length == 0;
-
-            //We need to check the children and should any child match this matches as well
-            if (filterable is IHasFilterableChildren hasFilterableChildren)
+            if (drawable is IContainerEnumerable<Drawable> container)
             {
-                foreach (IFilterable child in hasFilterableChildren.FilterableChildren)
-                    matching |= match(child, childTerms, searchActive, nonContiguousMatching);
+                foreach (var child in container.Children)
+                    matching |= matchSubTree(child, nonMatchingTerms, searchActive, nonContiguousMatching);
             }
 
-            filterable.FilteringActive = searchActive;
-            return filterable.MatchingFilter = matching;
+            if (drawable is IFilterable filterable)
+            {
+                filterable.FilteringActive = searchActive;
+                filterable.MatchingFilter = matching;
+            }
+
+            return matching;
         }
+
+        /// <summary>
+        /// Whether the specified drawable matches the specified search terms.
+        /// </summary>
+        /// <param name="drawable">The drawable to check for match.</param>
+        /// <param name="searchTerms">The search terms to check against.</param>
+        /// <param name="nonContiguousMatching">Whether the matching algorithm should be non-contiguous, allowing for potentially non-matching characters to exist between matches.</param>
+        /// <param name="nonMatchingTerms">The search terms which do not match the specified drawable.</param>
+        private static bool match(Drawable drawable, IReadOnlyList<string> searchTerms, bool nonContiguousMatching, out IReadOnlyList<string> nonMatchingTerms)
+        {
+            nonMatchingTerms = searchTerms;
+
+            if (drawable is IFilterable filterable)
+                nonMatchingTerms = nonMatchingTerms.Where(term => !checkTerms(filterable.FilterTerms, term, nonContiguousMatching)).ToArray();
+
+            return nonMatchingTerms.Count == 0;
+        }
+
+        private static bool checkTerms(IEnumerable<string> filterTerms, string searchTerm, bool nonContiguous) => filterTerms.Any(term => checkTerm(term, searchTerm, nonContiguous));
 
         /// <summary>
         /// Check whether a search term exists in a forward direction, allowing for potentially non-matching characters to exist between matches.
--- a/osu.Framework/Graphics/Containers/SearchContainer.cs
+++ b/osu.Framework/Graphics/Containers/SearchContainer.cs
@@ -21,9 +21,9 @@ namespace osu.Framework.Graphics.Containers
     /// </summary>
     /// <remarks>
     /// <list type="bullet">
-    /// <item>Children which are searchable should be marked with the <see cref="IFilterable"/> interface. They do not need to be direct children.</item>
-    /// <item>Containers above <see cref="IFilterable"/>s which should be hidden only when all children are filtered away should also be marked with the <see cref="IFilterable"/> interface.</item>
-    /// <item>Containers which have the potential to hide children via external means should be <see cref="VisibilityContainer"/>s, in order to correctly exclude hidden child <see cref="IFilterable"/>s from search.</item>
+    /// <item>Children which are searchable should be marked with the <see cref="IFilterable"/> interface. They do not need to be direct children to work (filtering will traverse the full drawable subtree).</item>
+    /// <item>Marking a container (ie. a "group" or "section" that contains nested <see cref="IFilterable"/>s) as <see cref="IFilterable"/> will automatically keep it non-filtered as long as at least one nested item is not filtered away.</item>
+    /// <item>Any <see cref="IFilterable"/>s which are contained in a <see cref="VisibilityContainer"/> that is hidden will be excluded from filtering. This can be used to exclude certain items from consideration (ie. items which are hidden from display), allowing group/sections to be correctly filtered away.</item>
     /// </list>
     /// </remarks>
     public class SearchContainer<T> : FillFlowContainer<T> where T : Drawable
--- a/osu.Framework/Graphics/Containers/SearchContainer.cs
+++ b/osu.Framework/Graphics/Containers/SearchContainer.cs
@@ -100,7 +100,7 @@ namespace osu.Framework.Graphics.Containers
         {
             bool matching = match(drawable, searchTerms, nonContiguousMatching, out var nonMatchingTerms);
 
-            if (drawable.IsPresent && drawable is IContainerEnumerable<Drawable> container)
+            if (drawable is IContainerEnumerable<Drawable> container && (container as VisibilityContainer)?.State.Value != Visibility.Hidden)
             {
                 foreach (var child in container.Children)
                     matching |= matchSubTree(child, nonMatchingTerms, searchActive, nonContiguousMatching);
@@ -133,9 +133,6 @@ namespace osu.Framework.Graphics.Containers
                 // reset filter to ensure presence is not influenced by previous filter state when checking.
                 filterable.MatchingFilter = true;
             }
-
-            if (!drawable.IsPresent)
-                return false;
 
             return nonMatchingTerms.Count == 0;
         }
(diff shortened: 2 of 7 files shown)


osu.Framework/Graphics/Containers/SearchContainer.cs
```diff
@@ -87,29 +86,62 @@ protected override void Update()
...
+        /// <param name="nonContiguousMatching">Whether the matching algorithm should be non-contiguous, allowing for potentially non-matching characters to exist between matches.</param>
+        private static bool matchSubTree(Drawable drawable, IReadOnlyList<string> searchTerms, bool searchActive, bool nonContiguousMatching)
         {
-            //Words matched by parent is not needed to match children
-            string[] childTerms = searchTerms.Where(term =>
-                !filterable.FilterTerms.Any(filterTerm =>
-                    checkTerm(filterTerm, term, nonContiguousMatching))).ToArray();
+            bool matching = match(drawable, searchTerms, nonContiguousMatching, out var nonMatchingTerms);
+
+            if (drawable.IsPresent && drawable is IContainerEnumerable<Drawable> container)
```
bdach, line 103: The `drawable.IsPresent` check here doesn't work correctly. If a child becomes present after the last `performFilter()` call, it will not receive the correct `FilteringActive`/`MatchingFilter` state.

See this diff, applicable on top of test scene added in this pull, that makes the test fail:

```diff
diff --git a/osu.Framework.Tests/Visual/UserInterface/TestSceneSearchContainer.cs b/osu.Framework.Tests/Visual/UserInterface/TestSceneSearchContainer.cs
index 60b6f0ab16..4e1ef7c369 100644
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneSearchContainer.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneSearchContainer.cs
@@ -132,6 +132,8 @@ public void TestRefilterAfterNewChild()
         [Test]
         public void TestFilterRespectsHiddenChild()
         {
+            SearchableText searchableText = null;
+
             AddStep("Add new hidden filtered item", () => search.Add(new HeaderContainer("Subsection 3")
             {
                 AutoSizeAxes = Axes.Both,
@@ -139,13 +141,17 @@ public void TestFilterRespectsHiddenChild()
                 {
                     Alpha = 0f,
                     AutoSizeAxes = Axes.Both,
-                    Child = new SearchableText { Text = "Hidden text" },
+                    Child = searchableText = new SearchableText { Text = "Hidden text" },
                 },
             }));
             setTerm("Hidden text");
 
             checkCount(0);
             AddAssert("No subsection displayed", () => search.ChildrenOfType<HeaderContainer>().All(h => !h.IsPresent));
+
+            AddStep("unhide text", () => searchableText.Alpha = 1);
+            checkCount(1);
+            AddAssert("Subsection displayed", () => search.ChildrenOfType<HeaderContainer>().Any(h => h.IsPresent));
         }
 
         private void checkCount(int count)
```

You'd probably want to mess with invalidation somehow here to fix this in the general case. This has potential to change the entire structure of this so I won't bother reviewing further for now.

bdach (dismissed):

osu.Framework/Graphics/Containers/SearchContainer.cs
frenzibyte: Though, the component which is hidden in this case should be the container above `searchableText`. I would imagine it to work because the draw size would change at that point, but unsure if that invalidates the container actually. Good note.

bdach: > the component which is hidden in this case should be the container above `searchableText`

What does this mean? I don't understand.

frenzibyte: There's a container which `searchableText` is wrapped inside, that container is hidden. Reason why is because `MatchingFilter` would update the alpha of `searchableText`, therefore I need to hide it at a higher level to test.

bdach: Wait so are you saying that it is not allowed to change alpha on an `IFilterable` and have it work correctly with your changes? That seems like a problem. Isn't that what the game is doing to hide e.g. the resolution dropdown in non-fullscreen configurations, and therefore this wouldn't fix any of the issues with it?

frenzibyte: Well, on the resolution dropdown case, it has changed to affect `IsPresent` directly. In this PR, `MatchingFilter` is reset for every filterable before `IsPresent` is checked, to ensure presence is not affected by a previous filter state.

What usages should do, however, is to not make filtering and hiding both affect the `Alpha` of the same layer. Either filtering affects `IsPresent` directly similar to how it's done in settings now, or the hiding is applied at a layer inside the filterable component.

I've pushed changes for your case above as it is still a valid concern (despite the test not exactly correct).

bdach: > What usages should do, however, is to not make filtering and hiding both affect the `Alpha` of the same layer

Bit of a shame. I don't think it's likely this can be enforced automatically without too much hassle, so I'd say that this information should be at least somewhere in xmldoc.

frenzibyte: It feels like the `IsPresent` way should've been done at a `SearchContainer` level somehow, but I've added an advisory xmldoc for now.

frenzibyte:
Not sure how valid this is, but I've measured the time spent in `SearchContainer.performLayout` (in release configuration) and this is the output:
```
# master
2022-05-05 04:25:48 [verbose]: Performing layout
2022-05-05 04:25:48 [verbose]: Finished layout: 0.37120799999956944ms, processed 43 containers, and 144 drawables in total.
2022-05-05 04:25:48 [verbose]: Performing layout
2022-05-05 04:25:48 [verbose]: Finished layout: 0.36929200000031415ms, processed 43 containers, and 144 drawables in total.
2022-05-05 04:25:48 [verbose]: Performing layout
2022-05-05 04:25:48 [verbose]: Finished layout: 0.39216600000054314ms, processed 43 containers, and 144 drawables in total.

# PR
2022-05-05 04:20:48 [verbose]: Performing layout
2022-05-05 04:20:48 [verbose]: Finished layout: 0.40475000000014916ms, processed 317 containers, and 792 drawables in total.
2022-05-05 04:20:48 [verbose]: Performing layout
2022-05-05 04:20:48 [verbose]: Finished layout: 0.38362499999993815ms, processed 317 containers, and 792 drawables in total.
2022-05-05 04:20:48 [verbose]: Performing layout
2022-05-05 04:20:48 [verbose]: Finished layout: 0.5557079999998678ms, processed 317 containers, and 792 drawables in total.
```

<details>
<summary>master diff</summary>

```diff
diff --git a/osu.Framework/Graphics/Containers/SearchContainer.cs b/osu.Framework/Graphics/Containers/SearchContainer.cs
index 789d92e88..ba5bbd877 100644
--- a/osu.Framework/Graphics/Containers/SearchContainer.cs
+++ b/osu.Framework/Graphics/Containers/SearchContainer.cs
@@ -7,6 +7,7 @@
 using System.Linq;
 using osu.Framework.Caching;
 using osu.Framework.Extensions.IEnumerableExtensions;
+using osu.Framework.Timing;
 
 namespace osu.Framework.Graphics.Containers
 {
@@ -84,10 +85,22 @@ protected override void Update()
             }
         }
 
+        private static readonly StopwatchClock perf_clock = new StopwatchClock(true);
+        private static int container_count = 0;
+        private static int count = 0;
+
         private void performFilter()
         {
+            Logging.Logger.Log("Performing layout");
+            double timeBefore = perf_clock.CurrentTime;
+
+            container_count = 0;
+            count = 0;
+
             string[] terms = (searchTerm ?? string.Empty).Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
             Children.OfType<IFilterable>().ForEach(child => match(child, terms, terms.Length > 0, allowNonContiguousMatching));
+
+            Logging.Logger.Log($"Finished layout: {perf_clock.CurrentTime - timeBefore}ms, processed {container_count} containers, and {count} drawables in total.");
         }
 
         private static bool match(IFilterable filterable, IEnumerable<string> searchTerms, bool searchActive, bool nonContiguousMatching)
@@ -97,11 +110,14 @@ private static bool match(IFilterable filterable, IEnumerable<string> searchTerm
                 !filterable.FilterTerms.Any(filterTerm =>
                     checkTerm(filterTerm, term, nonContiguousMatching))).ToArray();
 
+            count++;
             bool matching = childTerms.Length == 0;
 
             //We need to check the children and should any child match this matches as well
             if (filterable is IHasFilterableChildren hasFilterableChildren)
             {
+                container_count++;
+
                 foreach (IFilterable child in hasFilterableChildren.FilterableChildren)
                     matching |= match(child, childTerms, searchActive, nonContiguousMatching);
             }
```

</details>


<details>
<summary>PR diff</summary>

```diff

diff --git a/osu.Framework/Graphics/Containers/SearchContainer.cs b/osu.Framework/Graphics/Containers/SearchContainer.cs
index 74018aca2..8353e61b2 100644
--- a/osu.Framework/Graphics/Containers/SearchContainer.cs
+++ b/osu.Framework/Graphics/Containers/SearchContainer.cs
@@ -6,6 +6,7 @@
 using System.Globalization;
 using System.Linq;
 using osu.Framework.Caching;
+using osu.Framework.Timing;
 
 namespace osu.Framework.Graphics.Containers
 {
@@ -83,10 +84,22 @@ protected override void Update()
             }
         }
 
+        private static readonly StopwatchClock perf_clock = new StopwatchClock(true);
+        private static int container_count = 0;
+        private static int count = 0;
+
         private void performFilter()
         {
+            Logging.Logger.Log("Performing layout");
+            double timeBefore = perf_clock.CurrentTime;
+
+            container_count = 0;
+            count = 0;
+
             string[] terms = (searchTerm ?? string.Empty).Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
             matchSubTree(this, terms, terms.Length > 0, allowNonContiguousMatching);
+
+            Logging.Logger.Log($"Finished layout: {perf_clock.CurrentTime - timeBefore}ms, processed {container_count} containers, and {count} drawables in total.");
         }
 
         /// <summary>
@@ -102,6 +115,8 @@ private static bool matchSubTree(Drawable drawable, IReadOnlyList<string> search
 
             if (drawable.IsPresent && drawable is IContainerEnumerable<Drawable> container)
             {
+                container_count++;
+
                 foreach (var child in container.Children)
                     matching |= matchSubTree(child, nonMatchingTerms, searchActive, nonContiguousMatching);
             }
@@ -124,6 +139,7 @@ private static bool matchSubTree(Drawable drawable, IReadOnlyList<string> search
         /// <param name="nonMatchingTerms">The search terms which do not match the specified drawable.</param>
         private static bool match(Drawable drawable, IReadOnlyList<string> searchTerms, bool nonContiguousMatching, out IReadOnlyList<string> nonMatchingTerms)
         {
+            count++;
             nonMatchingTerms = searchTerms;
 
             if (drawable is IFilterable filterable)
```

</details>

Unsure how to write a proper benchmark that shows the speed of this new implementation though, since the prior relies on `IHasFilterableChildren`, and that's deprecated now in this PR.

peppy:
Just a thought: Would making `MatchingFilter` a `get;set;` property and relying on that instead of `IsPresent` not clean things up? I'm still not fully on board with this PR due to the `IsPresent` dependency..

ie. for any containers osu-side which are hiding groups of settings, they would need to implement `IFilterable` and return `false` to `MatchingFilter` if they were hidden, or something like that..

peppy:
Another alternative would be to *require* the use of `VisibilityContainer` for any "hiding" containers, and explicitly mention that in the `SearchContainer` usage instructions. Then we could rely on `VisibilityContainer.State` instead of `IsPresent` for the traversal checks.

frenzibyte:
> ie. for any containers osu-side which are hiding groups of settings, they would need to implement `IFilterable` and return `false` to `MatchingFilter` if they were hidden, or something like that..

Not sure how to feel about implementing `IFilterable` for a component that isn't even filterable. I personally think the `IsPresent` method makes sense since if a container is not present, we should automatically ignore it and filter it out. I don't know---

> Another alternative would be to _require_ the use of `VisibilityContainer` for any "hiding" containers, and explicitly mention that in the `SearchContainer` usage instructions. Then we could rely on `VisibilityContainer.State` instead of `IsPresent` for the traversal checks.

Hmm... maybe that could work, but I'm not entirely sure how valid that is, since it means we'll have to enforce using `VisibilityContainer` anywhere we want to hide something for one reason or another.

Note that the `IsPresent` check is also done for settings which hide themselves (i.e. resolution dropdown). Would you be fine with wrapping those cases in a visibility container? `VisibilityContainer` is abstract, so we'll have to locally implement that as well. Maybe make `SettingsItem<T>` inherit from `VisibilityContainer` instead?

peppy:
I'd be okay with wrapping `resolutionDropdown`, yeah.

frenzibyte:
Looks to be feasible after trying that. I've gone and refactored the "hide-able" components in game settings to use `VisibilityContainer` ([branch for visibility](https://github.com/ppy/osu/compare/master..frenzibyte:filterable-stuff?expand=1)), and so far I don't see any issues with the new pattern.

osu.Framework/Graphics/Containers/IFilterable.cs
```diff
@@ -3,6 +3,13 @@
 
 namespace osu.Framework.Graphics.Containers
 {
+    /// <summary>
+    /// Represents a component which can be filtered out on non-matching search terms.
+    /// </summary>
+    /// <remarks>
+    /// Hiding <see cref="IFilterable"/>s for purposes other than filtering require being wrapped in a
+    /// <see cref="VisibilityContainer"/> with its visibility state set to <see cref="Visibility.Hidden"/>.
```
peppy, line 11: I think this documentation should be on the `SearchContainer` class, mentioning something like:

```
Summary:
A container which handles filtering of the child hierarchy.

Remarks:
- Children which are searchable should be marked with the `IFilterable` interface. They do not need to be direct children.
- Containers above `IFilterable`s which should be hidden only when all children are filtered away should be marked with the `IHasFilterableChildren` interface.
- Containers which have the potential to hide children via external means should be `VisibilityContainer`s, in order to correctly exclude hidden child `IFilterable`s from search.

**Review by peppy — CHANGES_REQUESTED** [review]

File: osu.Framework/Graphics/Containers/IFilterable.cs (hunk 2 above)
Comment (frenzibyte) [reply osu.Framework/Graphics/Containers/IFilterable.cs]: As mentioned, containers above `IFilterable`s which should hide based on their children `IFilterable`s will still have to be marked with an `IFilterable` interface, to allow for cases where some child `IFilterable`s may be nested below others:
```
 - Root
 -   IFilterable Header
 -     IFilterable 1
 -     IFilterable 2
 -     IFilterable 3
 -     (VisibilityContainer)
 -         IFilterable 4
```

Which, with an `IHasFilterableChildren` interface, would require explicitly specifying that in the property (`FilterableChildren => filterables.Append(filterableInVisibility)`), and that can get worst if the `FilterableChildren` impleemntation is inherited by a base class (i.e. how we have the settings overlay currently structured).

In addition, with an `IHasFilterableChildren` interface, it won't be simple to tell whether a filterable is hidden by a parent `VisibilityContainer`, unless we mark said container with an `IFilterable` interface and append it to the children list.

Comment (peppy) [reply osu.Framework/Graphics/Containers/IFilterable.cs]: Do you not find it weird that you may have containers that define as `IFilterable` and specify zero filter terms? It feels very counter intuitive.

Comment (frenzibyte) [reply osu.Framework/Graphics/Containers/IFilterable.cs]: Is there an existing case of that? I’m going by the concept that no container hide on child filterables without having its own search terms that group them (i.e. a header container).

Comment (peppy) [reply osu.Framework/Graphics/Containers/IFilterable.cs]: I guess it's maybe quite uncommon?

File: osu.Framework/Graphics/Containers/SearchContainer.cs [hunk 3]
```diff
@@ -19,7 +19,13 @@ public class SearchContainer : SearchContainer<Drawable>
...
     /// new items are added as direct children of this container.
     /// </summary>
-    /// <typeparam name="T"></typeparam>
+    /// <remarks>
+    /// <list type="bullet">
+    /// <item>Children which are searchable should be marked with the <see cref="IFilterable"/> interface. They do not need to be direct children.</item>
+    /// <item>Containers above <see cref="IFilterable"/>s which should be hidden only when all children are filtered away should also be marked with the <see cref="IFilterable"/> interface.</item>
+    /// <item>Containers which have the potential to hide children via external means should be <see cref="VisibilityContainer"/>s, in order to correctly exclude hidden child <see cref="IFilterable"/>s from search.</item>
+    /// </list>
+    /// </remarks>
```
peppy, line 28:
```suggestion
    /// <remarks>
    /// <list type="bullet">
    /// <item>Children which are searchable should be marked with the <see cref="IFilterable"/> interface. They do not need to be direct children to work (filtering will traverse the full drawable subtree).</item>
    /// <item>Marking a container (ie. a "group" or "section" that contains nested <see cref="IFilterable"/>s) as <see cref="IFilterable"/> will automatically keep it not-filtered as long as at least one nested item is not filtered away.</item>
    /// <item>Any <see cref="IFilterable"/>s which are contained in a <see cref="VisibilityContainer"/> that is hidden will be excluded from filtering. This can be used to exclude certain items from consideration via an external means (ie. items which have been collapsed), allowing group/sections to be correctly filtered away.</item>
    /// </list>
    /// </remarks>
```

Reworded to hopefully suck less (yes I know I wrote most of the previous version). Still not 100% on the last line.

peppy (changes requested):

osu.Framework/Graphics/Containers/SearchContainer.cs
frenzibyte: "items which have been collapsed" may feel misleading (in test scene, searching for a test scene already expands collapsed groups/folders). I feel it's not necessary to mention example of "external means".

peppy: I've reworded slightly, see how that is now.

frenzibyte: Reads better :+1:

frenzibyte:
> I did notice one issue regarding the layout invalidation change I pushed, which is that the search container is actually be invalidated per-frame, destroying the point of a searchCache.
>
> I’m not entirely sure what’s going on, but unless someone beats me to it, I’ll try investigating tomorrow.

Turns out this is happening due to the `SearchContainer` in test scene having a margin applied, which causes this logic to invalidate the layout per-frame:
https://github.com/ppy/osu-framework/blob/a379f77079502a8ccf3cf80832350fa88f5e987e/osu.Framework/Graphics/Containers/CompositeDrawable.cs#L1844-L1852

The `SearchContainer` used for game settings also has a margin applied by the surrounding `SectionsContainer`, and I'm not entirely sure if this is considered a bug. But since it traverses the full container subtree, it may be better to work around it at least.

Have pushed one that ignores the `MiscGeometry` invalidation, as it should also be redundant in the case of filtering (only `DrawSize` matters to catch addition of children in the hierarchy).

frenzibyte:
Actually I'm just going to revert that because it only ended up breaking tests, and I'm not entirely sure what can be done here... going to block for now at least.

bdach:
What needs to happen for this to be unblocked again (besides resolving the merge conflicts)? I am interested in getting this in as I'm trying and failing to work around multiple `SearchContainer` idiosyncrasies in order to add a search box to the mod select overlay.

bdach:
OK, but are you planning to get back to this at some point or would I need to go try and fix myself? That's what I'm most interested in.

bdach:
After looking at this pull again after a while I think it has a fatal internal contradiction in the special handling of `VisibilityContainer`s. On one side, the most common implementation of `MatchingFilter` (the one in the visual tests included) will use `Show()` and `Hide()` for achieving the visual effects of the filtering. On the other, if you have a `Thing : VisibilityContainer, IFilterable`, then if you write `MatchingFilter` using `Show()` and `Hide()`, you will end up inadvertently changing `Thing.State` to `Visibility.Hidden` and therefore exclude the entire `Thing` from further search, which this PR is hacking around with this:

https://github.com/ppy/osu-framework/blob/26c722ee4a8e16a891864254c1a3ae2721432422/osu.Framework/Graphics/Containers/SearchContainer.cs#L139-L140

which I'm pretty sure is the ultimate cause of the issue described above wherein the search container will invalidate every frame if no item is matching (because it's flapping between the "reset everything to unfiltered" state and the "everything is filtered" state).

I don't know what the fix is yet, but I'd say there is too much in-band data passing to my liking here. Likely we need a separate API to exclude `IFilterable`s from search that does not rely on visual properties like alpha or `VisibilityContainer.State`. Something like that has already been suggested above.

peppy:
Superseded by https://github.com/ppy/osu-framework/pull/5530.

PR #5159, opened 2022-05-05
Add Ctrl+Y as an alternative "redo" platform binding

### LINKED ISSUE #5157: PlatformAction.Redo is bound to Ctrl+Shift+Z but not Ctrl+Y
`ctrl + y`, as an alternative to `ctrl + shift + z` is commonly used as the redo action, especially with art software (for example because you can bind control and z and y to 3 close button on a graphics tablet and that's where you usually hold your hand. Having ctrl bound instead of the undo/redo action to 2 buttons is better because you can essentially have 2 times more binds with and without the modifier) but it is not a default platform key binding. There are no conflicting bindings, I think it should be added. I found it confusing I could not redo until I found its bound to the keyboard-normal `ctrl + shift + z`.

bdach:
Closes #5157

I believe this shouldn't apply to macOS, because that has [an entirely disparate set of platform keybindings](https://github.com/ppy/osu-framework/blob/80cffac85458e307d02bfa804d7da4d47405095f/osu.Framework/Platform/MacOS/MacOSGameHost.cs#L67-L118), but would appreciate testing to make sure.

Diff includes simple manual test scene for testing convenience (press any platform binding and the boxes will flash):

image

### DIFF
--- a/osu.Framework.Tests/Visual/Input/TestScenePlatformActionContainer.cs
+++ b/osu.Framework.Tests/Visual/Input/TestScenePlatformActionContainer.cs
@@ -1,0 +1,90 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Linq;
+using osu.Framework.Allocation;
+using osu.Framework.Graphics;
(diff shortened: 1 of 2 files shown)


frenzibyte (approved):
Tested on Windows and macOS, behaves as expected.

PR #5161, opened 2022-05-07
Fix default maximum size being invalid on Windows

Susko3:
Not sure why, but Windows treats `int.MaxValue` as zero here.
This would result in the window being forced to resize to minimum size (640x480) when trying to move or resize it.

### DIFF
--- a/osu.Framework/Configuration/FrameworkConfigManager.cs
+++ b/osu.Framework/Configuration/FrameworkConfigManager.cs
@@ -23,7 +23,7 @@ namespace osu.Framework.Configuration
         {
             SetDefault(FrameworkSetting.ShowLogOverlay, false);
 
-            SetDefault(FrameworkSetting.WindowedSize, new Size(1366, 768), new Size(640, 480));
+            SetDefault(FrameworkSetting.WindowedSize, new Size(1366, 768), new Size(640, 480), new Size(65536, 65536));
             SetDefault(FrameworkSetting.ConfineMouseMode, ConfineMouseMode.Fullscreen);
             SetDefault(FrameworkSetting.ExecutionMode, ExecutionMode.MultiThreaded);
             SetDefault(FrameworkSetting.WindowedPositionX, 0.5, -0.5, 1.5);
--- a/osu.Framework/Configuration/FrameworkConfigManager.cs
+++ b/osu.Framework/Configuration/FrameworkConfigManager.cs
@@ -23,7 +23,7 @@ namespace osu.Framework.Configuration
         {
             SetDefault(FrameworkSetting.ShowLogOverlay, false);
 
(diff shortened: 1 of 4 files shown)


osu.Framework/Configuration/FrameworkConfigManager.cs
```diff
@@ -23,7 +23,7 @@ protected override void InitialiseDefaults()
         {
             SetDefault(FrameworkSetting.ShowLogOverlay, false);
 
-            SetDefault(FrameworkSetting.WindowedSize, new Size(1366, 768), new Size(640, 480));
+            SetDefault(FrameworkSetting.WindowedSize, new Size(1366, 768), new Size(640, 480), new Size(int.MaxValue / 2, int.MaxValue / 2));
```
frenzibyte, line 26: Would limit the maximum value internally at `SDL_SetMaximumWindowSize` rather than in here.

frenzibyte (dismissed):
I guess this is fine? would've hoped there's a better way to achieve this but SDL doesn't provide much there.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -477,7 +477,10 @@ public virtual void Create()
                 if (max.Width < sizeWindowed.MinValue.Width || max.Height < sizeWindowed.MinValue.Height)
                     throw new InvalidOperationException($"Expected a size greater than min window size ({sizeWindowed.MinValue}), got {max}");
 
-                ScheduleCommand(() => SDL.SDL_SetWindowMaximumSize(SDLWindowHandle, max.Width, max.Height));
+                // Windows treats very large values as zero.
+                const int max_size = int.MaxValue / 2;
```
peppy, line 481: Rather than doing this locally, can we set a sane upper bounds on the bindable instead? Maybe `32768`?

frenzibyte: [That's how it used to be](https://github.com/ppy/osu-framework/pull/5161#discussion_r867366778), but the consumer will still be able to update that maximum value to go beyond `32768`, therefore I think it's better to limit it internally instead.

peppy: I dunno. This is pretty ugly as a result. Not sure that safety is required.

peppy: I've reverted the last commit roughly to the state it was in before. I think this reads better. If the user is setting stupid max size values that's on them, IMO.

peppy (dismissed):

peppy (approved):

osu.Framework/Platform/SDL2DesktopWindow.cs
frenzibyte: Fair enough, it does seem like an unnecessary safety overall.

frenzibyte (approved):

PR #5162, opened 2022-05-07
Fix `PlatfomAction` test not highlighting actions that map to the same key combination

Susko3:
Eg. `Delete` and `DeleteForwardChar` are both mapped to the `Del` key, and only one would highlight previously.

`Exit` is still handled to allow testing without closing the window.


peppy (approved):
Also noticed this previously, is a welcome change.

PR #5163, opened 2022-05-08
Enable `SDL_HINT_MOUSE_AUTO_CAPTURE` when not using relative mode

### LINKED ISSUE #4055: Held mouse buttons aren't updated when the cursor is outside of the window
To reproduce:
(0. use windowed mode)
1. hold any mouse button
2. leave the game window

The mouse button state should be updated, but isn't until the cursor returns to the window. This can be seen in the visual tests already.

Susko3:
Depends on:
- [x] SDL bump (for https://github.com/libsdl-org/SDL/commit/5ff42438e36e98c9d72ac70f5f5ce4599b96d3d2)

Partially resolves #4055. Now the issue only happens when relative mode is enabled.

To get mouse capture for relative mode, we'd probably want to do it locally in `MouseHandler` by keeping track of pressed buttons.

### DIFF
--- a/osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
+++ b/osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
@@ -93,11 +93,11 @@ namespace osu.Framework.Input.Handlers.Mouse
             cursorState = desktopWindow.CursorStateBindable.GetBoundCopy();
             cursorState.BindValueChanged(_ => updateRelativeMode());
 
-            UseRelativeMode.BindValueChanged(_ =>
+            UseRelativeMode.BindValueChanged(e =>
             {
-                if (window != null)
-                    updateRelativeMode();
-            });
+                window.MouseAutoCapture = !e.NewValue;
+                updateRelativeMode();
+            }, true);
 
             Enabled.BindValueChanged(enabled =>
             {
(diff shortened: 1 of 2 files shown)


osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
```diff
@@ -93,11 +93,11 @@ public override bool Initialize(GameHost host)
             cursorState = desktopWindow.CursorStateBindable.GetBoundCopy();
             cursorState.BindValueChanged(_ => updateRelativeMode());
 
-            UseRelativeMode.BindValueChanged(_ =>
+            UseRelativeMode.BindValueChanged(e =>
             {
-                if (window != null)
-                    updateRelativeMode();
```
Susko3, line 99: `window` is never null here. seems to be a leftover.

osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
```diff
@@ -93,11 +93,11 @@ public override bool Initialize(GameHost host)
...
-            });
+                window.MouseAutoCapture = !e.NewValue;
```
frenzibyte, line 98: This is actually the first time a hint is dynamically enabled/disabled. Is there a reason not to keep this enabled when relative mode is on? if so, that should be mentioned in the xmldoc, for future readers.

Susko3: Setting in `RelativeMouseMode` is wrong, as relative mode is disabled when leaving the window. `MouseAutoCapture` should mirror `UseRelativeMode`.

peppy (approved):

Issue #5164, opened 2022-05-08
Introduce "reduce dropped frames" support


SonoSooS:
# Main information

tl;dr: the text of this issue is just describing the effects of glFinish, and how calling or not calling it after SwapBuffers affects certain drivers.

Considering the UHD 620/630 driver bug, I suggest that instead of just a binary "call glFinish" / "no need to call glFinish", also have a third option named "force rendering", which should do platform-specific window redrawing, like calling InvalidateRect and glFinish on Windows before SwapBuffers.
Although the problem is that this would require platform-specific calls, which seem to have only very minimum marginal improvements on some drivers on top of calling glFinish, so considering the amount of work it needs, it might not be desired just for a "very small percentage" of userbase. Just calling glFinish is usually enough for most broken drivers.

# Details

After gathering a large amount of intel (no pun intended), it seems like it's finally time to implement "reduce dropped frames" support into the framework.

Looking at [GameHost.Swap()](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Platform/GameHost.cs#L530-L541), the current way of this being implemented is being enforced with VSync ([except as a workaround for macOS](https://github.com/ppy/osu-framework/pull/3258), which should be no longer needed once this is implemented).
Using glFinish on VSync is unnecessary, as it's the driver's job to block on SwapBuffers, and handle VSync-relevant stuff accordingly.

**However**, drivers are weird, and they come with tiny little bugs which add up into everyone's pain.
Most notorious are Intel UHD 620/630 GPUs, as they have peculiar driver issues, which make them require very simple, yet necessary workarounds. Although it seems to be not an Intel-exclusive problem, they are harder hit by this issue than other manufacturers.

There are probably still a few issues which I can't find via search, but this is the most prominent one I could find: [ppy/osu#7447].
This is a known problem with UHD 620/630, although those affected in the issue are hit much harder with this bug than it normally happens (like on Windows for example).
A notable mention I found is [ppy/osu#9851], which looks similar to the UHD 620/630 bug, except it's probably simply caused by a driver overload, and not a buggy driver. Considering that VSync got rid of the stutters, it is very most likely that the driver is overloaded so much that it freezes the entire system, which the UHD 620/630 bug also does quite regularly.

[...]

Issue #5067, opened 2022-03-11
Consider deprecating HasFlagsFast() extension method


smoogipoo:
As of https://github.com/dotnet/runtime/pull/59514, the runtime implementation has been hoisted into the C# core library and no longer incurs allocation/performance overhead. It seems like this change is present in at least .NET 6.0.1

```
|      Method |      Mean |     Error |    StdDev | Allocated |
|------------ |----------:|----------:|----------:|----------:|
|     HasFlag | 0.0232 ns | 0.0024 ns | 0.0023 ns |         - |
|  BitwiseAnd | 0.0115 ns | 0.0057 ns | 0.0050 ns |         - |
| HasFlagFast | 0.0160 ns | 0.0023 ns | 0.0020 ns |         - |
```

PR #5165, opened 2022-05-09
[#5164] Add driver workarounds support

SonoSooS:
I tried my best to put things at logical places, but to me it still feels like a bit of a hack, especially where the driver metadata is propagated through out to be reachable by GameHost.

I'm not good at naming things, so things like GraphicsBackendMetadata and PlatformWorkaroundMode might have to be renamed as well.

Tested on Windows, macOS, and Linux (Solus), and it seems to work as intended. Tested with a different GPU as well on Windows, and it's also fine there.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -532,12 +532,46 @@ namespace osu.Framework.Platform
         /// </summary>
         protected virtual void Swap()
         {
-            Window.SwapBuffers();
+            switch (currentWorkaroundMode)
+            {
+                case Platform.PlatformWorkaroundMode.ForceFinish:
+                    Window.SwapBuffers();
+                    GL.Finish();
 
-            if (Window.VerticalSync)
-                // without glFinish, vsync is basically unplayable due to the extra latency introduced.
-                // we will likely want to give the user control over this in the future as an advanced setting.
-                GL.Finish();
+                    break;
+
+                case Platform.PlatformWorkaroundMode.ForceNoFinish:
+                    Window.SwapBuffers();
+                    break;
+
+                case Platform.PlatformWorkaroundMode.WorkaroundIntelUHD630_Windows:
+                    // InvalidateRect is absolutely necessary for this driver workaround,
+                    //  otherwise the latency is higher, and dwm will still crash eventually
+                    if (this.Window is SDL2DesktopWindow sdlWindow)
+                        Windows.Native.Window.InvalidateRect(sdlWindow.WindowHandle, IntPtr.Zero, false);
+
+                    // First glFinish is needed, unless we're able to call wglSwapLayerBuffers directly
+                    GL.Finish();
+                    // Calls SwapBuffers internally
+                    Window.SwapBuffers();
+                    // Second glFinish required for synchronizing the driver
+                    GL.Finish();
+
+                    break;
+
+                case Platform.PlatformWorkaroundMode.Default:
+                default:
+                    Window.SwapBuffers();
+
+                    if (Window.VerticalSync)
+                    {
+                        // without glFinish, vsync is basically unplayable due to the extra latency introduced.
+                        // we will likely want to give the user control over this in the future as an advanced setting.
+                        GL.Finish();
+                    }
+
+                    break;
+            }
         }
 
         /// <summary>
@@ -747,6 +781,8 @@ namespace osu.Framework.Platform
                 {
                     if (Window != null)
                     {
+                        PlatformWorkaroundMode.TriggerChange();
+
                         switch (Window)
                         {
                             case SDL2DesktopWindow window:
@@ -930,6 +966,9 @@ namespace osu.Framework.Platform
 
         private Bindable<bool> bypassFrontToBackPass;
 
+        public Bindable<PlatformWorkaroundMode> PlatformWorkaroundMode;
+        private PlatformWorkaroundMode currentWorkaroundMode = Platform.PlatformWorkaroundMode.Default;
+
         private Bindable<FrameSync> frameSyncMode;
 
         private IBindable<DisplayMode> currentDisplayMode;
@@ -1020,6 +1059,30 @@ namespace osu.Framework.Platform
             }, true);
 
             inputConfig = new InputConfigManager(Storage, AvailableInputHandlers);
+
+            PlatformWorkaroundMode = Config.GetBindable<PlatformWorkaroundMode>(FrameworkSetting.PlatformWorkaroundMode);
+            PlatformWorkaroundMode.BindValueChanged(workaroundMode =>
+            {
+                if (workaroundMode.NewValue == Platform.PlatformWorkaroundMode.Auto)
+                {
+                    GraphicsBackendMetadata? rendererMetadata = this.Window.RendererMetadata;
+                    if (rendererMetadata.HasValue)
+                    {
+                        currentWorkaroundMode = PlatformWorkaroundDetector.DetectWorkaround(rendererMetadata.Value);
+                    }
+                    else
+                    {
+                        // No way to check the workaround mode, so fall back to defaults instead
+                        currentWorkaroundMode = Platform.PlatformWorkaroundMode.Default;
+                    }
+                }
+                else
+                {
+                    currentWorkaroundMode = workaroundMode.NewValue;
+                }
+
+                Logger.Log("Changed workaround mode to " + currentWorkaroundMode);
+            });
         }
 
         private void updateFrameSyncMode()
--- a/osu.Framework/Platform/PlatformWorkaroundDetector.cs
+++ b/osu.Framework/Platform/PlatformWorkaroundDetector.cs
@@ -1,0 +1,76 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.ComponentModel;
+
+namespace osu.Framework.Platform
+{
+    public static class PlatformWorkaroundDetector
+    {
+        public static PlatformWorkaroundMode DetectWorkaround(GraphicsBackendMetadata backendMetadata)
+            => DetectWorkaround(backendMetadata, RuntimeInfo.OS);
+
+        public static PlatformWorkaroundMode DetectWorkaround(GraphicsBackendMetadata backendMetadata, RuntimeInfo.Platform platform)
+        {
+            //HACK: Force macOS to use glFinish for the time being, until it is further investigated (https://github.com/ppy/osu/issues/7447)
+            if (platform == RuntimeInfo.Platform.macOS)
+                return PlatformWorkaroundMode.ForceFinish;
+
+
+            if (backendMetadata.Vendor == "Intel")
+            {
+                if (backendMetadata.RendererName.Contains("UHD Graphics 620") ||
+                    backendMetadata.RendererName.Contains("UHD Graphics 630"))
+                {
+                    // UHD 620/630 needs custom workarounds for Windows
+                    //  due to the bug causing excess overload until
+                    //  dwm or the driver ends up crashing.
+                    if (platform == RuntimeInfo.Platform.Windows)
+                        return PlatformWorkaroundMode.WorkaroundIntelUHD630_Windows;
+
+                    // On macOS there is simply just a scheduling bug,
+                    //  which can be kept in sync by just a mere glFinish.
+                    if (platform == RuntimeInfo.Platform.macOS)
+                        return PlatformWorkaroundMode.ForceFinish;
+                }
+            }
+
+            return PlatformWorkaroundMode.Default;
+        }
+    }
+
+    public enum PlatformWorkaroundMode
+    {
+        /// <summary>
+        /// Default is to glFinish on VSync, but don't otherwise.
+        /// </summary>
+        [Description("No workaround")]
+        Default,
+
+        /// <summary>
+        /// Acts just like <see cref="Default"/>, but should indicate
+        ///  forced workaround detection.
+        /// </summary>
+        [Description("Automatic workaround detection")]
+        Auto,
+
+        /// <summary>
+        /// Always call glFinish after SwapBuffers.
+        /// </summary>
+        [Description("Force drawing")]
+        ForceFinish,
+
+        /// <summary>
+        /// Never call glFinish after SwapBuffers.
+        /// </summary>
+        [Description("No synchronization")]
+        ForceNoFinish,
+
+        /// <summary>
+        /// Use InvalidateRect + glFinish twice to
+        ///  synchronize the UHD 620/630 driver
+        /// </summary>
+        [Description("Intel UHD 620/630 stutter workaround (Windows)")]
+        WorkaroundIntelUHD630_Windows
+    }
+}
--- a/osu.Framework/Platform/PlatformWorkaroundDetector.cs
+++ b/osu.Framework/Platform/PlatformWorkaroundDetector.cs
@@ -32,6 +32,10 @@ namespace osu.Framework.Platform
                     //  which can be kept in sync by just a mere glFinish.
                     if (platform == RuntimeInfo.Platform.macOS)
                         return PlatformWorkaroundMode.ForceFinish;
+                    
+                    // UHD 630 is not broken on Linux due to the Mesa driver being mature
+                    //  due to being open-source, and supported by the community.
+                    // Perhaps a check should be here for Linux to see if it's not running on Mesa drivers?
                 }
             }
 
@@ -57,13 +61,13 @@ namespace osu.Framework.Platform
         /// <summary>
         /// Always call glFinish after SwapBuffers.
         /// </summary>
-        [Description("Force drawing")]
+        [Description("Force drawing synchronization")]
         ForceFinish,
 
         /// <summary>
         /// Never call glFinish after SwapBuffers.
         /// </summary>
-        [Description("No synchronization")]
+        [Description("Disable drawing synchronization")]
         ForceNoFinish,
 
         /// <summary>
--- a/osu.Framework/Platform/Windows/Native/Window.cs
+++ b/osu.Framework/Platform/Windows/Native/Window.cs
@@ -1,0 +1,14 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace osu.Framework.Platform.Windows.Native
+{
+    internal static class Window
+    {
+        [DllImport("user32.dll")]
+        internal static extern uint InvalidateRect(IntPtr hwnd, IntPtr lpRect, bool bErase);
+    }
+}
--- a/osu.Framework/Configuration/FrameworkConfigManager.cs
+++ b/osu.Framework/Configuration/FrameworkConfigManager.cs
@@ -45,6 +45,8 @@ namespace osu.Framework.Configuration
             SetDefault(FrameworkSetting.IgnoredInputHandlers, string.Empty);
             SetDefault(FrameworkSetting.CursorSensitivity, 1.0, 0.1, 6, 0.01);
 #pragma warning restore 618
+
+            SetDefault(FrameworkSetting.PlatformWorkaroundMode, PlatformWorkaroundMode.Auto);
         }
 
         public FrameworkConfigManager(Storage storage, IDictionary<FrameworkSetting, object> defaultOverrides = null)
@@ -104,5 +106,7 @@ namespace osu.Framework.Configuration
 
         [Obsolete("Input-related settings are now stored in InputConfigManager. Adjustments should be made via Host.AvailableInputHandlers bindables directly.")] // can be removed 20210911
         MapAbsoluteInputToWindow,
+
+        PlatformWorkaroundMode
     }
 }
--- a/osu.Framework/Platform/GraphicsBackendMetadata.cs
+++ b/osu.Framework/Platform/GraphicsBackendMetadata.cs
@@ -1,0 +1,34 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework.Platform
+{
+    /// <summary>
+    /// Contains information to identify a specific graphics backend vendor and renderer
+    /// </summary>
+    public readonly struct GraphicsBackendMetadata
+    {
+        /// <summary>
+        /// Product or software used to render the display
+        /// </summary>
+        public readonly string RendererName;
+
+        /// <summary>
+        /// Vendor of the renderer product or software
+        /// </summary>
+        public readonly string Vendor;
+
+        /// <summary>
+        /// OpenGL version string, including vendor-specific version info
+        /// </summary>
+        public readonly string VersionString;
+
+
+        public GraphicsBackendMetadata(string rendererName, string vendor, string versionString)
+        {
+            this.RendererName = rendererName;
+            this.Vendor = vendor;
+            this.VersionString = versionString;
+        }
+    }
+}
--- a/osu.Framework/Platform/IGraphicsBackend.cs
+++ b/osu.Framework/Platform/IGraphicsBackend.cs
@@ -14,6 +14,11 @@ namespace osu.Framework.Platform
         bool VerticalSync { get; set; }
 
         /// <summary>
+        /// Returns information about the renderer, if available.
+        /// </summary>
+        GraphicsBackendMetadata? RendererMetadata { get; }
+
+        /// <summary>
         /// Initialises the graphics backend, given the current window backend.
         /// It is assumed that the window backend has been initialised.
         /// </summary>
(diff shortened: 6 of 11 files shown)


osu.Framework/Platform/PlatformWorkaroundDetector.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.ComponentModel;
+
+namespace osu.Framework.Platform
+{
+    public static class PlatformWorkaroundDetector
+    {
+        public static PlatformWorkaroundMode DetectWorkaround(GraphicsBackendMetadata backendMetadata)
+            => DetectWorkaround(backendMetadata, RuntimeInfo.OS);
+
+        public static PlatformWorkaroundMode DetectWorkaround(GraphicsBackendMetadata backendMetadata, RuntimeInfo.Platform platform)
```
smoogipoo, line 13: Initially this is fine, though I believe we may want this to be a list or a dictionary in the future. Reasoning being is that browsers typically have many workarounds that are applied to situationally to different systems, such that we will probably not want workarounds for specific GPUs/platforms but generic workarounds such as `ForceInvalidateRect`/`ForceFinish`/`FinishBeforeSwap`/etc.

osu.Framework/Platform/GameHost.cs
```diff
@@ -532,12 +532,46 @@ protected virtual void DrawFrame()
...
+                    break;
+
+                case Platform.PlatformWorkaroundMode.WorkaroundIntelUHD630_Windows:
+                    // InvalidateRect is absolutely necessary for this driver workaround,
+                    //  otherwise the latency is higher, and dwm will still crash eventually
+                    if (this.Window is SDL2DesktopWindow sdlWindow)
+                        Windows.Native.Window.InvalidateRect(sdlWindow.WindowHandle, IntPtr.Zero, false);
+
+                    // First glFinish is needed, unless we're able to call wglSwapLayerBuffers directly
+                    GL.Finish();
```
smoogipoo, line 554: I'm a little bit confused by this, isn't this basically what `SDL2GraphicsBackend.SwapBuffers()` does? Internally it seems to call https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-swapbuffers but it sounds weird that it would do anything more than call `wglSwapLayerBuffers` internally / not call that function at all.

peppy:
I think this direction is okay, apart from it being a setting. I'd propose this is removed. At most, it should be adjustable via envvar or something like that, for developer use only. Users should not be able to accidentally set this setting and have their game in an unknown state.

osu.Framework/Platform/PlatformWorkaroundDetector.cs
SonoSooS: Would an embedded resource file work? Like a .yaml or .json or .csv of specific pattern-matching values (not necessarily Regex, but something simple as `*UHD Graphics 630*` would work as well to some extent). Unless it's requested, I don't think I'll do the resource file approach for this particular pull request yet, so we can start rolling this feature out to those affected ASAP.

As for generic workarounds, I presume we'd want to combine them as `[Flags]`. I can do that, that should "simplify" the code, and remove redundant copypaste.

SonoSooS:
> I think this direction is okay, apart from it being a setting. I'd propose this is removed. At most, it should be adjustable via envvar or something like that, for developer use only. Users should not be able to accidentally set this setting and have their game in an unknown state.

Should I move it to debug settings (reset every launch due to no config file backing), or should I completely axe any config, and have it exclusively as an Environment variable?

osu.Framework/Platform/PlatformWorkaroundDetector.cs
bdach: Embedded resource may be fine as long as it's not re-engineering a new format or anything. Should be as simple as can be and reuse existing formats where possible - then again, I don't see any problem in having the patterns be defined by actual code rather than an embedded config file.

Not sure about making the workaround enum a flag enum - some of the current `PlatformWorkaroundMode`s are mutually exclusive (force/disable drawing sync). If anything there could be an object with multiple applicable "workaround settings" - but all of this isn't really required for anything right now, I think. It's not even clear that another one of these "workarounds" is going to ever arrive.

bdach:
Going back to more down-to-earth manners, code style here is going to need some nudging, CI doesn't like it.

SonoSooS:
> I'd say no config.
> 
> And probably not even environment variable, as those who would be interested in changing these knobs would have to be able to compile for themselves. I don't want people shooting themselves in the foot by giving suggestions to others to "create a shortcut with this env variable to lower input latency" or whatever.

I'll implement the changes proposed in the review comments, axe the configurability, not implement environment variable configurability, and we can leave this PR open until we decide if we need any user-facing accessibility or not.

Although personally I would still leave a moderately easy way to make this user-accessible, so we can instruct users with GPU problems (like in [ppy/osu#9851]) to test a few settings to determine what causes it. In fact, in that exact case I linked, I think it's a problem unique to the user who reported it, so they would somehow need to apply that fix to themselves without applying the a fix unnecessarily to everyone else, degrading everyone else's performance just because it's isolatable to a single user. The UHD 630 bug affects everyone on Windows and macOS, so that's why it requires the fix to be applied to everyone on Windows and macOS.



> Going back to more down-to-earth matters, code style here is going to need some nudging, CI doesn't like it.

I know, I just read some rules somewhere which say to not push unnecessary commits which unnecessarily overload the CI, and since I had review comments asking for changes, I decided not to push that code style fix (I was already done with before sleep) until I was done with the new changes.

osu.Framework/Platform/GameHost.cs
```diff
@@ -532,12 +533,42 @@ protected virtual void DrawFrame()
...
-                // we will likely want to give the user control over this in the future as an advanced setting.
+            if(currentWorkarouns.HasFlagFast(PlatformWorkaround.FinishBeforeSwap))
+            {
+                //TODO: This glFinish is only needed because we're not able to call wglSwapLayerBuffers directly, as to not break any possible future SDL code changes
                 GL.Finish();
+            }
+
+            Window.SwapBuffers();
+
+            if(currentWorkarouns.HasAnyFlagFast(PlatformWorkaround.FinishAfterSwapVSync | PlatformWorkaround.FinishAfterSwapNoVSync))
```
SonoSooS, line 552: The only reason I used two explicit enums checks instead of FinishAfterSwapAlways is purely due to code readability, as it's less confusing, and more explicit.

osu.Framework/Platform/GameHost.cs
```diff
@@ -532,12 +533,42 @@ protected virtual void DrawFrame()
         /// </summary>
         protected virtual void Swap()
         {
-            Window.SwapBuffers();
+            if(currentWorkarouns.HasFlagFast(PlatformWorkaround.WindowsInvalidateRect)) // Should only be set on Windows platforms
```
peppy, line 536: Typo

osu.Framework/Extensions/EnumExtensions/EnumExtensions.cs
```diff
@@ -95,5 +95,47 @@ public static IEnumerable<T> GetValuesInOrder<T>(this IEnumerable<T> items)
...
         }
+        
+        /// <summary>
+        /// Similar to <see cref="EnumExtensions.HasFlagFast{T}(T, T)"/>, but returns <see langword="true"/> if any of the given flags are set.
+        /// </summary>
+        /// <param name="enumValue">The enum to check.</param>
+        /// <param name="flags">The flags to check for.</param>
+        /// <returns>Whether any of the <paramref name="flags"/> are set in <paramref name="enumValue"/> or not</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static unsafe bool HasAnyFlagFast<T>(this T enumValue, T flags) where T : unmanaged, Enum
```
peppy, line 106: Feels like overkill for this PR. I would have just used `HasFlagFast` twice. But okay I guess.

osu.Framework/Platform/GameHost.cs
```diff
@@ -532,12 +533,42 @@ protected virtual void DrawFrame()
...
+            {
+                //TODO: This glFinish is only needed because we're not able to call wglSwapLayerBuffers directly, as to not break any possible future SDL code changes
                 GL.Finish();
+            }
+
+            Window.SwapBuffers();
+
+            if(currentWorkarouns.HasAnyFlagFast(PlatformWorkaround.FinishAfterSwapVSync | PlatformWorkaround.FinishAfterSwapNoVSync))
+            {
+                if(Window.VerticalSync)
```
peppy, line 554: Please check code style.

osu.Framework/Platform/GameHost.cs
peppy: I don't get it. This doesn't need to exist. The two conditionals are checked separately inside this block.

peppy (changes requested):

osu.Framework/Extensions/EnumExtensions/EnumExtensions.cs
smoogipoo: I'd say remove this and use `(enumValue & flags) > 0` at the callsite. `HasFlagFast` is deprecated but not yet marked as such: https://github.com/ppy/osu-framework/issues/5067.

I can't guarantee the implementation of `HasFlagFast` won't break and won't exhibit weirdness on some other data types - it differs a bit from the internal implementation in the runtime.

osu.Framework/Platform/GameHost.cs
SonoSooS: This is embarrassing. It seems like the style checker is silently failing to run on my end.

osu.Framework/Platform/GameHost.cs
SonoSooS: In hindsight I don't see its point either, as they are checked always anyway.

SonoSooS: Ah, I'm sorry. This wasn't mentioned in the contributing guidelines, and I thought I could spare some effort by flagging minor issues "fixed".

peppy:
I've applied fixes for the remaining inspections so we can keep moving this forward.

osu.Framework/Platform/PlatformWorkaroundDetector.cs
```diff
@@ -0,0 +1,85 @@
...
+                    //  dwm or the driver ends up crashing.
+                    if (platform == RuntimeInfo.Platform.Windows)
+                        return PlatformWorkaround.WindowsInvalidateRect
+                             | PlatformWorkaround.FinishBeforeSwap //TODO: wglSwapLayerBuffers is preferred over an explicit pre-SwapBuffers glFinish
+                             | PlatformWorkaround.FinishAfterSwapAlways;
+
+                    // On macOS there is simply just a scheduling bug,
+                    //  which can be kept in sync by just a mere glFinish.
+                    if (platform == RuntimeInfo.Platform.macOS)
+                        return PlatformWorkaround.FinishAfterSwapAlways;
```
peppy, line 37: This can't be reached due to the initial conditional on line 16.

SonoSooS: Yeah, as the comment says, we still need to test if macOS *really* needs the forced swap, or if only the UHD 630 needs the forced swap.

I only have three devices running macOS, so I can do some limited testing to determine if it's really needed *right now*. Although the low sample-size would still leave a bit uneasy if it gets merged after concluding that it's not needed.

SonoSooS: Yes, it had to be UHD 630, as the 2018 13" MacBook Pro comes with an i7-8750H, which is the exact same CPU I have, so it had to be UHD 630. Sadly we can't determine this from the logs though, as the Google Drive link in the original issue is dead.

frenzibyte: From the specifications mentioned in https://github.com/ppy/osu/issues/7447, the 2018 13" MacBook Pro actually comes with an i7-8559U, which doesn't have a UHD 620/630, but an [Iris Plus Graphics 655](https://ark.intel.com/content/www/us/en/ark/products/137979/intel-core-i78559u-processor-8m-cache-up-to-4-50-ghz.html) instead. Confirmed with [this specific comment](https://github.com/ppy/osu/issues/7447#issuecomment-570942858) which shows the GPU used, and lines up with my statement.

Note that on the same comment, it can be noticed that the issue is also resolved by switching to VSync, which means UHD 620/630 aren't the only GPUs causing this "freezing" issue in the first place. May wanna update on that against SDL.

SonoSooS: OUCH, I didn't notice that they didn't write "Touchbar". Thanks for the correction, non-Touchbar modells indeed don't come with an i7-8750H.

So it seems like apparently almost (if not) all Coffee Lake (8th gen) iGPUs are affected, some newer 7th gen are also reported being affected (although that may be a different issue?), and I even heard of a single instance of an Intel HD 535 being affected by this bug (which is weird, since it doesn't seem to even exist??? yet glGetString reported 535).

I would love to bring up the sources for my claims, but they didn't survive two browser migrations, so I'd have to find them again.

I guess for now let's just detect 8th-gen graphics, and worry about newer 7th-gen issues if they get reported?

peppy: Also to reiterate, the `glFinish` obliterates performance for me on an AMD card on macOS, last I checked.

SonoSooS: Here are my results:
 - NVIDIA GeForce 320M (Core 2 Duo P8600 MacBookPro7,1)
    - Windows (10 20H2)
        - Default: ~30-63-73FPS, the image is complete carnage and flickerfest
        - FinishAfterSwapAlways: 63-93-121FPS, the image still flickers under heavy load, but it's actually really playable
    - macOS (10.13.6)
        - Default: unknown FPS, the image features frequent few-frame stutters
        - FinishAfterSwapAlways: unknown FPS, probably the same performance as on Windows, and the stutters are gone (except the ones caused by the extremely weak CPU)
- Intel HD 3000 (i7-2860QM MacBookPro8,2)
    - Windows: iGPU driver does not work at all on any Windows version, not even with patches
    - macOS (10.13.6)
        - 0 (no glFinish on VSync): absolutely low latency with VSync, even when the FPS seriously dips below 60
        - Default (glFinish on VSync): there is definitely harder work on VSync, and if the FPS dips below 60 then there is a noticable latency
        - FinishAfterSwapAlways: the FPS is higher and more stable, but there is a noticable latency

Conclusion:
- NVIDIA GeForce 320M: FinishAfterSwapAlways would needed, but I don't think we should add this to the list, as it's a miracle that the game even runs, and it doesn't help that much in performance anway, especially since the entire screen is flickering garbage rectangles all over the place, and on macOS textures get corrupted into dashes and equals signs
- Intel HD 3000: absolute lack of glFinish is actually somehow better (which should not be surprising, as this is how it should be for every driver in the first place!), but I won't add an exception, as the performance is abysmal anyway in general, and the gains are too minimal to warrant wasted space in the code

Based on our very small amount of data, it is indeed better to revert the "force glFinish on macOS" patch, and just add the offending GPUs to the workaround list.

> Also to reiterate, the `glFinish` obliterates performance for me on an AMD card on macOS, last I checked.

That's good to know, as one of my MacBook Pros has a dead AMD GPU (the dreaded 2011 17" one), so I couldn't test that one myself.

> Can confirm the `glFinish` is decreasing FPS for me somewhat as well (M1 GPU), probably best to remove it altogether and include `Iris Plus Graphics 655` in the list.
> 
> Skeptical about including a full generation set though, unless there's solid evidence of the issue affecting all of them, I would rather just add the workaround for the GPUs we know are affected at least, and include more as we go, rather than potentially breaking things and receive more reports than we intend to see.

Roger that, I'll add the ones that I'm 100% certain about are broken, and I'll exclude a few ones I know are broken, but don't have enough sample size of evidence to prove it that it's not just an one-off error for those users (like HD 530 and 535).

SonoSooS: Removed the hardcoded macOS VSync completely, so now what has worked in the past without it should work again without performance detrement, but Intel Gen9 should receive workarounds, so they get more stable performance now.

Also moved the hardcoded detection to a separate function, as a lot of Intel Gen9 GPUs seem to be affected, but we don't have enough sample size yet to confirm if it's a widespread issue, or if it's unique to some users.

osu.Framework/Platform/GameHost.cs
```diff
@@ -532,12 +532,39 @@ protected virtual void DrawFrame()
         /// </summary>
         protected virtual void Swap()
         {
+            if ((currentWorkarounds & PlatformWorkaround.WindowsInvalidateRect) != 0) // Should only be set on Windows
+            {
+                // InvalidateRect is absolutely necessary for some driver workarounds,
+                //  otherwise the latency will be higher, and dwm might crash sometimes
```
peppy, line 538: Which drivers? What dwm crash? Links or better comment please.

osu.Framework/Platform/GameHost.cs
```diff
@@ -532,12 +532,39 @@ protected virtual void DrawFrame()
...
+                if (Window is SDL2DesktopWindow sdlWindow)
+                    Windows.Native.Window.InvalidateRect(sdlWindow.WindowHandle, IntPtr.Zero, false);
+            }
+
+            if ((currentWorkarounds & PlatformWorkaround.FinishBeforeSwap) != 0)
+            {
+                //TODO: This glFinish is only needed because we're not able to call wglSwapLayerBuffers directly, as to not break any possible future SDL code changes
```
peppy, line 545: This doesn't really explain anything. Is it implying that an alternative to having this here would be to call `wglSwapLayerBuffers` instead of `Window.SwapBuffers` or something?

osu.Framework/Platform/GameHost.cs
```diff
@@ -532,12 +532,39 @@ protected virtual void DrawFrame()
...
 
             if (Window.VerticalSync)
-                // without glFinish, vsync is basically unplayable due to the extra latency introduced.
-                // we will likely want to give the user control over this in the future as an advanced setting.
-                GL.Finish();
+            {
+                if ((currentWorkarounds & PlatformWorkaround.FinishAfterSwapVSync) != 0)
+                {
+                    // without glFinish, vsync is basically unplayable due to the extra latency introduced.
+                    // we will likely want to give the user control over this in the future as an advanced setting.
```
peppy, line 556: This doesn't make sense either. This workaround is only being applied to certain scenarios, but the comment reads as if it's required for all vsync scenarios.

If the comments aren't actually giving more details on the devices or scenarios where these are required, then just nuke them.

osu.Framework/Platform/GameHost.cs
```diff
@@ -1020,6 +1053,22 @@ protected virtual void SetupConfig(IDictionary<FrameworkSetting, object> default
             }, true);
 
             inputConfig = new InputConfigManager(Storage, AvailableInputHandlers);
+
+            PlatformWorkarounds.BindValueChanged(workaroundMode =>
```
peppy, line 1057: Just remove this whole bindable if possible. Should only need to be set once.

peppy (changes requested):
Initial review

osu.Framework/Platform/GameHost.cs
SonoSooS: Yeah, that would be the most ideal, and then we could rename FinishBeforeSwap to WindowsUseWglSwapLayerBuffers, and have wglSwapLayerBuffers called instead of Window.SwapBuffers when that flag is set.

osu.Framework/Platform/GameHost.cs
SonoSooS: I left this comment in the code, as it was already there, and didn't want to remove it just because of the "we will likely want to give the user control over this in the future as an advanced setting", especially because tests have shown that for example Intel HD 3000 actually gets a performance hit from glFinish on VSync.

And also considering nobody yet had an obvious final say on configurability, I still had hope that it could still change before this PR would get merged. I still don't know if I should implement the envvar configurability or not, as nobody has said it that it *has* to be done a specific way, with a period.

peppy: https://github.com/ppy/osu-framework/pull/5165#issuecomment-1122442763

I agree with this, else I would have said something against it. No configuration please.

osu.Framework/Platform/GameHost.cs
SonoSooS: Feels like now I'm trying to convey *too much* information.

Perhaps I should move a big chunk of it to #5164 and just leave a reference to it in the comments?

image

osu.Framework/Platform/GameHost.cs
SonoSooS: Completely removed the bindable, and moved the ternary operator assignment line to where the .TriggerChange call used to be.

Also renamed currentWorkarounds to platformWorkarounds to better describe what it's doing, without having to search for its references in code to figure out what it does.

osu.Framework/Platform/GameHost.cs
SonoSooS: Slightly revamped the glFinish comments, hopefully it's not excessively verbose.

image

osu.Framework/Platform/GameHost.cs
peppy: I'd much rather a short description with a permalink to somewhere with more discussion. All this block does is make my eyes glaze over, and there's nothing to tell me that this is an actual confirmed issue documented elsewhere (it should be).

osu.Framework/Platform/GameHost.cs
peppy: I don't know, it's still weird. I think we need someone else to write this. For instance you still mention "seems to improve latency" – according to who? on all devices?

then you mention ancient GPUs. like okay, but are they the only ones that use this workaround? do they need to?

the text you write only leads to more questions.

SonoSooS:
As per discussion at multiple places, I have removed Windows workarounds, as it's an issue which has to be solved by the user by upgrading/downgrading their OS and driver.

On macOS however that is not possible, so macOS workarounds have to be kept for now.

osu.Framework/Platform/PlatformWorkaroundDetector.cs
```diff
@@ -0,0 +1,126 @@
...
+            => value >= min && value <= max;
+    }
+
+    [Flags]
+    public enum PlatformWorkaround
+    {
+        /// <summary>
+        /// Indicates that glFinish is not needed at all for a particular driver/product
+        /// </summary>
+        NoFinish = 0,
```
peppy, line 104: Shouldn't this be the default? I thought the whole point of this change is to stop doing `glFinish` in scenarios it's not required (like my hardware, where it reduces my performance).

osu.Framework/Platform/Windows/Native/Window.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace osu.Framework.Platform.Windows.Native
+{
+    internal static class Window
+    {
+        [DllImport("user32.dll")]
+        internal static extern uint InvalidateRect(IntPtr hwnd, IntPtr lpRect, bool bErase);
```
peppy, line 12: Unused

peppy (changes requested):
As proposed

osu.Framework/Platform/PlatformWorkaroundDetector.cs
SonoSooS: Actually, my intent was to *add* glFinish where it's needed, but not only did this entire PR turn 180degrees, it actually turned like 1080degrees at this point, so it would be probably a good experiment to remove VSync glFinish.

While I'm kind of afraid of the consequences, something tells me that
1) if removing VSync glFinish degrades performance, then a driver downgrade/upgrade will fix it
2) if there is no such driver, then the affected users will open a new issue

So if I get an another confirmation, then I will make the complete lack of glFinish the default then, from the current one, which is VSync glFinish to decrease latency on some drivers (we don't have enough data of).

osu.Framework/Platform/Windows/Native/Window.cs
SonoSooS: I thought I removed all remnants! Removed in next commit.

osu.Framework/Platform/PlatformWorkaroundDetector.cs
peppy: I'd say it's fine to remove and test against all our available hardware (between the core dev team we have a good variety of apple devices from different eras).

Also, in an enum the `0` item should always be `None` or `Default`, not some arbitrary value.

SonoSooS: Is it okay if I have both `None` and `Default`? With `Default` being set to `None`.

peppy: Choose one or the other. We use `None` generally so I'd go with that.

peppy:
To confirm intention: half of the enum doesn't ever get triggered, and this will currently only affect macOS users. What's the plan with the remaining workaround you've proposed here? Is it no longer required?

SonoSooS:
> To confirm intention: half of the enum doesn't ever get triggered, and this will currently only affect macOS users. What's the plan with the remaining workaround you've proposed here? Is it no longer required?

I actually use all four enum values in my local builds, as I made a build for each of my laptops which I use, as some of them experience great latency improvements and FPS stabilization instead of just minor FPS increase. Slightly inconvenient to have to rebuild each update, but I guess it is what it is, and I have to deal with it.

I can trim down the code to what is actually used right now, and if we end up ever needing the deleted code, then we could just fish it back out from the git history.

Please confirm if the code has to be trimmed, and then I'll push the changes.

PR #5168, opened 2022-05-11
Allow retrieving full global statistics on demand

peppy:
To be used for realm error reporting.


smoogipoo:
o_O those CI failures...

smoogipoo (approved):

PR #5172, opened 2022-05-11
Reflect actual PowerShell invocation necessary in README

SonoSooS:
It is necessary to pass `-ExecutionPolicy Bypass` before the script filename if the script is not signed by a code signing certificate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Build configurations for the recommended IDEs (listed above) are included. You s
 
 ### Code analysis
 
-Code analysis can be run with `powershell ./InspectCode.ps1` or `InspectCode.sh`.
+Code analysis can be run with `powershell -ExecutionPolicy Bypass .\InspectCode.ps1` or `InspectCode.sh`.
 
 ## Contributing
 


bdach:
Execution policy is like the first thing any single person running powershell scripts bashes their head against. I don't think we should be mentioning this, much less mindlessly suggesting to bypass it.

SonoSooS:
To be fair, you do have a point at that, that any developer able to contribute could probably find `-ExecutionPolicy Bypass` afrter 1-2mins of searching, and remember it for life.

Although looking into InspectCode.ps1, it could be simply renamed to .bat, after changing `$LASTEXITCODE` into `%ERRORLEVEL%`, and then I could just do `InspectCode` instead of `powershell -ExecutionPolicy Bypass .\InspectCode.ps1`.

If I added InspectCode.bat which just invocated the PowerShell script, or renamed the .ps1 to .bat, would it be rejected?

bdach:
Batch scripts are rather antiquated in this day and age, all things considered. I personally much prefer a powershell script even if it doesn't do much. Not sure about other opinions.

frenzibyte:
Would agree with sticking to powershell rather than downgrading to batch.

SonoSooS:
To be fair, considering the "feature set" of Batch, bash, and PowerShell, Batch is definitely not known for its ease of use or rich feature set, so me suggesting a downgrade to Batch was perhaps a bit foolhardy of me. Who knows when in the future we might need to find a file programatically in these scripts, but if we had to, it would be painful to do in Batch.

Although I'd still love to have a "launcher" .bat script which launches the PowerShell script with `-ExecutionPolicy Bypass`, so I wouldn't have to type it out all the time. It's fine typing it out a few times, but when the time you execute it manually becomes multiple digits, then it does get annoying.

Susko3:
You can set powershell execution policy to always allow running local scripts in windows dev settings.

image

peppy:
If this is to be added, it should be a line above the script linking to documentation on changing the system execution policy, not attached to the command.

Conflicted to whether we need this though. May be best to leave that for the dev to figure as a bar-of-entry since it's basically something you need to do to dev on windows?

peppy:
Let's just close this for now, I don't think it's going to go anywhere since we all agree this is not the correct direction.

PR #5174, opened 2022-05-15
Support matching by both base and localised text in `SearchContainer`

maromalo:
- Needed for https://github.com/ppy/osu/pull/18270, from feedback.
- [x] Fix the tests

Note this change will require each class which inherits `IHasFilterTerms` (about 20 in osu!lazer) to have their definition of `FilterItems` be a `IEnumerable<LocalizedString>`

Now SearchContainer should match based on base text **and** localised text.

### DIFF
--- a/osu.Framework/Graphics/Containers/SearchContainer.cs
+++ b/osu.Framework/Graphics/Containers/SearchContainer.cs
@@ -97,8 +97,8 @@ namespace osu.Framework.Graphics.Containers
 
         private bool match(IFilterable filterable, IEnumerable<string> searchTerms, bool searchActive, bool nonContiguousMatching)
         {
-            IEnumerable<string> filterTerms = filterable.FilterTerms.SelectMany(localizedStr =>
-                new[] { localizedStr.ToString(), localisation?.GetLocalisedString(localizedStr) });
+            IEnumerable<string> filterTerms = filterable.FilterTerms.SelectMany(localisedStr =>
+                new[] { localisedStr.ToString(), localisation.GetLocalisedString(localisedStr) });
 
             //Words matched by parent is not needed to match children
             string[] childTerms = searchTerms.Where(term =>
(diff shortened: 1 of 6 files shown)


osu.Framework/Graphics/Sprites/SpriteText.cs
```diff
@@ -636,6 +636,6 @@ public float LineBaseHeight
             }
         }
 
-        public IEnumerable<string> FilterTerms => displayedText.Yield();
+        public IEnumerable<LocalisableString> FilterTerms => new LocalisableString(displayedText).Yield();
```
bdach, line 639: Why use `displayedText` and not

```suggestion
        public IEnumerable<LocalisableString> FilterTerms => Text.Yield();
```

osu.Framework/Graphics/Containers/SearchContainer.cs
```diff
@@ -90,11 +95,14 @@ private void performFilter()
             Children.OfType<IFilterable>().ForEach(child => match(child, terms, terms.Length > 0, allowNonContiguousMatching));
         }
 
-        private static bool match(IFilterable filterable, IEnumerable<string> searchTerms, bool searchActive, bool nonContiguousMatching)
+        private bool match(IFilterable filterable, IEnumerable<string> searchTerms, bool searchActive, bool nonContiguousMatching)
         {
+            IEnumerable<string> filterTerms = filterable.FilterTerms.SelectMany(localizedStr =>
+                new[] { localizedStr.ToString(), localisation?.GetLocalisedString(localizedStr) });
```
bdach, line 101: * Please use `localisedStr` or similar rather than `localizedStr` for consistent spelling.
* Why is the `localisation?.GetLocalisedString()` call null-checked? `match()` cannot be called unless the drawable is updating, at which point `LocalisationManager` should have been injected already. Also `checkTerm()` (called a few lines down) cannot be called on a null string, so this partially looks like there's a potential bug here.

bdach:
Additionally this is missing new test coverage for the localisation support. There should be new tests here to show that the feature is working as advertised.

smoogipoo (approved):
lgtm

PR #5178, opened 2022-05-16
Fix unsafe file writes in multiple framework components

peppy:
As noticed in https://github.com/ppy/osu-framework/issues/5177, writes of these files have been unsafe until now if the file write is interrupted partway. The length of the written file could be shorter than expected.

This uses a newly added method to write to a temporary file before moving to the final location.

osu-side branch for testing: https://github.com/ppy/osu/compare/master...peppy:update-framework?expand=1

Closes https://github.com/ppy/osu-framework/issues/5177.


osu.Framework/Platform/Storage.cs
```diff
@@ -98,6 +98,23 @@ public virtual Storage GetStorageForDirectory(string path)
...
+        public abstract void Move(string from, string to);
+
+        /// <summary>
+        /// Create a new file on disk, using a temporary file to write to before moving to the final location to ensure a half-written file cannot exist at the specified location.
+        /// </summary>
+        /// <remarks>
+        /// If the target file path already exists, it will be deleted before attempting to write a new version.
+        /// </remarks>
+        /// <param name="path">The path of the file to create or overwrite.</param>
+        /// <returns></returns>
```
bdach, line 115:
```suggestion
        /// <returns>A <see cref="Stream"/> to which contents of the file to be created safely should be written.</returns>
```

or just remove, idunno - bit hard to explain that one

osu.Framework/Platform/NativeStorage.cs
```diff
@@ -105,5 +109,32 @@ public override Storage GetStorageForDirectory([NotNull] string path)
...
+        private class SafeWriteStream : FlushingStream
+        {
+            private readonly string temporaryPath;
+            private readonly string finalPath;
+            private readonly Storage storage;
+
+            public SafeWriteStream(string temporaryPath, string finalPath, Storage storage)
+                : base(storage.GetFullPath(temporaryPath, true), FileMode.Create, FileAccess.Write)
+            {
+                storage.Delete(finalPath);
```
bdach, line 125: Out of curiosity mostly, isn't this delete kind of useless? Like to avoid any plausible concurrency issues, I'd say that I'd imagine it should happen as close to the move operation as possible, i.e. in `Dispose()`? This is the scenario that I'm thinking of here:

```
Thread A                                                        Thread B

streamA = new SafeFileStream("tempA.txt", "file.txt", storage)
    storage.Delete("file.txt")

                                                                streamB = new SafeFileStream("tempB.txt", "file.txt", storage)
                                                                    storage.Delete("file.txt")

                                                                streamB.Dispose()
                                                                    storage.Move("tempB.txt", "file.txt")

streamA.Dispose()
    storage.Move("tempA.txt", "file.txt")
```

The file delete doesn't really do anything useful for this sort of scenario. Moving it to disposal wouldn't eliminate the race there, but it would make it less likely to happen.

bdach (commented):
Seems generally sane

osu.Framework/Platform/NativeStorage.cs
peppy: I was more thinking about the case where the stream write operation failed along the way, in which case doing an early delete will potentially clean up an old version of the file (and leave no file instead).

A bit of an undefined expectation, and I'm not sure how correct it is.

peppy: @smoogipoo interested in your opinion on this one

smoogipoo: If we're talking a power outage or force-quit of the application, then this method of doing things is very dangerous. Because in that scenario you have no way of recovery - the original file has been deleted _and_ the new file has potentially not finished writing yet.

How about enforcing a non-random temporary path such as `_{filename}`, and always delete that path instead?

smoogipoo: No, I'm suggesting writing to a fixed temporary path instead.

```
File.Delete("_{filename}");
// Write to _{filename}
File.Delete("{filename}");
File.Move("_{filename}", "{filename}");
```
Or, as I recently found out, `File.Replace()` is a thing which does a backup. So something like:
```
File.Delete("_{filename}");
// Write to _{filename}
File.Replace("_{filename}", "{filename}", "{filename}_backup");
File.Delete("{filename}_backup");
```

peppy: First of all, the `File.Delete` you're proposing isn't required because we are using `Create` which overwrites.

But it's safe to say you conceptually disagree with "early clean up" as mentioned in https://github.com/ppy/osu-framework/pull/5178#discussion_r874320190? ie. leaving the old file in place until the new write is 100% completed is what we want?

I think I agree that this is better behaviour. I originally had the clean-up early as it was removing broken cache files, but now that we are using this in more places it makes less sense.

peppy: Oh, and if `File.Replace` is doing a `Copy` operation to create the backup, it can't be used here. As that will add a performance overhead I was looking to avoid by only using `Move` operations.

smoogipoo: > But it's safe to say you conceptually disagree with "early clean up" as mentioned in [#5178 (comment)](https://github.com/ppy/osu-framework/pull/5178#discussion_r874320190)? ie. leaving the old file in place until the new write is 100% completed is what we want?

Correct.

> As that will add a performance overhead

I'm actually curious on whether it would copy (in particular on Windows). See this comment on `LinkOrCopyFile`:
image

peppy: Probably using hard links internally (what I also want to use eventually for beatmap importing). Maybe okay.

smoogipoo:
Looks to be failing tests.

smoogipoo (approved):
Seems ok

Issue #5167, opened 2022-05-10
Windows natural scroll setting not applied correctly to horizontal scrolling


peppy:
See https://github.com/ppy/osu/discussions/18175

peppy:
I... don't think this is a "natural scroll" related issue. We actually look to be handling the direction back-to-front, even on macOS. I believe we somewhat arbitrarily implemented it earlier on without too much consideration, and got it wrong.

We're likely going to need to fix this with a breaking change.

Susko3:
Tested on a windows precision touchpad, and with a mouse that has a tilting scrollwheel.
Changing natural scrolling on the touchpad properly inverts framework handling, so it's not related to that.

In both cases vertical scrolling works as expected, while horizontal scrolling is inverted (goes in the wrong direction).
"Forced" horizontal scrolling (when the `ScrollContainer` only scrolls horizontally, so vertical scrolls are translated to horizontal) works as expected.

Windows doesn't have default handling for 'hold shift for horizontal scroll' -- maybe something we want to implement locally?

peppy:
For reference: https://github.com/ocornut/imgui/issues/4019#issuecomment-852093067

Susko3:
`TestCursor` used in `ManualInputManager` inverts the vertical axis only. And this makes it move according to expectation: scrolling up moves it up, tilting left moves it left -- is this how it behaves on macOS?

https://github.com/ppy/osu-framework/blob/c945622b6e8c268e6729446a499a3597b8293bcd/osu.Framework/Testing/Input/TestCursorContainer.cs#L149

frenzibyte:
on macOS, only the vertical axis is inverted on "natural scroll", yes. And o!f works pretty well with it as-is, unlike Windows (seemingly).

frenzibyte:
Actually to clarify, I haven't tested a tilting scroll wheel, and answering this will have to depend on whether enabling/disabling "natural scroll" on macOS affects the horizontal direction as well. 

I had been thinking about this with trackpad in mind (where dragging towards any direction moves to the opposite side of it).

Susko3:
> I had been thinking about this in my previous comment with trackpad in mind (where dragging towards any direction moves to the opposite side of it).

Same happens on windows, the red dot moves opposite the scroll direction. 

[...]

PR #5179, opened 2022-05-16
Fix horizontal scrolling being inverted on non-Apple platforms

Susko3:
- Closes: #5167

Tested on:
- [x] Windows
- [x] Android
- [x] Linux (wayland/X11?)
- iPadOS/iOS (just to confirm it's the same as macOS)

Breaking changes
---

`ScrollEvent.ScrollDelta.X` (horizontal scrolling) is now consistent across platforms. Previously, Windows, Linux and Android were inverted. As a reminder:

> Delta is positive when mouse wheel scrolled to the up or left, in non-"natural" scroll mode (ie. the classic way).

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -959,7 +959,7 @@ namespace osu.Framework.Platform
         }
 
         private void handleMouseWheelEvent(SDL.SDL_MouseWheelEvent evtWheel) =>
-            // SDL reports horizontal scroll opposite of what framework expects.
+            // SDL reports horizontal scroll opposite of what framework expects (in non-"natural" mode, scrolling to the right gives negative deltas while we expect positive).
             TriggerMouseWheel(new Vector2(-evtWheel.x, evtWheel.y), false);
 
         private void handleMouseButtonEvent(SDL.SDL_MouseButtonEvent evtButton)
@@ -1540,6 +1540,9 @@ namespace osu.Framework.Platform
         /// <summary>
         /// Invoked when the user scrolls the mouse wheel over the window.
         /// </summary>
+        /// <remarks>
+        /// Delta is positive when mouse wheel scrolled to the up or left, in non-"natural" scroll mode (ie. the classic way).
+        /// </remarks>
         public event Action<Vector2, bool> MouseWheel;
 
         protected void TriggerMouseWheel(Vector2 delta, bool precise) => MouseWheel?.Invoke(delta, precise);
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -959,7 +959,8 @@ namespace osu.Framework.Platform
         }
 
         private void handleMouseWheelEvent(SDL.SDL_MouseWheelEvent evtWheel) =>
-            TriggerMouseWheel(new Vector2(evtWheel.x, evtWheel.y), false);
+            // SDL reports horizontal scroll opposite of what framework expects.
+            TriggerMouseWheel(new Vector2(-evtWheel.x, evtWheel.y), false);
 
         private void handleMouseButtonEvent(SDL.SDL_MouseButtonEvent evtButton)
         {
--- a/osu.Framework.Android/Input/AndroidMouseHandler.cs
+++ b/osu.Framework.Android/Input/AndroidMouseHandler.cs
@@ -209,7 +209,9 @@ namespace osu.Framework.Android.Input
             }
         }
 
-        private Vector2 getEventScroll(MotionEvent e) => new Vector2(e.GetAxisValue(Axis.Hscroll), e.GetAxisValue(Axis.Vscroll));
+        private Vector2 getEventScroll(MotionEvent e) =>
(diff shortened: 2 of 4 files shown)


osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -959,7 +959,8 @@ private void handleJoyAxisEvent(SDL.SDL_JoyAxisEvent evtJaxis)
         }
 
         private void handleMouseWheelEvent(SDL.SDL_MouseWheelEvent evtWheel) =>
-            TriggerMouseWheel(new Vector2(evtWheel.x, evtWheel.y), false);
+            // SDL reports horizontal scroll opposite of what framework expects.
+            TriggerMouseWheel(new Vector2(-evtWheel.x, evtWheel.y), false);
```
frenzibyte, line 963: Probably need to test for confirmation, but wouldn’t this invert for macOS now..? (trackpad specifically)

Susko3: macOS has it's own [scroll handler](https://github.com/ppy/osu-framework/blob/dde9eb08ede4aa94d21590f92b7c60f85cac3947/osu.Framework/Platform/MacOS/MacOSWindow.cs#L35-L60). Should have mentioned that in the PR.

peppy: Already tested this earlier today with a similar fix. Works as intended.

peppy (dismissed):

bdach:
Checked linux. On x11 I can't get horizontal scroll to work at all on touchpad, on wayland the direction is more correct with this change and follows the system-wide "natural scrolling" settings.

bdach (approved):

frenzibyte:
Would've appreciated adding a description explaining the forward direction with this change, as I've had a hard time understanding how this is more correct when it's flipping everything (but macOS) to match what o!f "wants".

Coming from a [questioning I had](https://discord.com/channels/188630481301012481/589331078574112768/975837331678441502), now with this PR, scrolling right on Windows (non-natural) sends a negative `ScrollDelta` value, while scrolling up sends a positive `ScrollDelta` value. Both directions are supposed to be considered for increasing UX-wise, and this means that every implementation of `OnScroll` with similar purposes will now have to consider flipping the X delta to achieve expectations.

I believe SDL and Android are rather more correct here, while macOS is broken for doing the opposite and `ScrollContainer` flipping the direction of the X delta as well to satisfy macOS.

Alternative solution that would avoid the above concern would be to do the complete opposite, by flipping the X value in macOS (and iOS as well?) to match all other platforms (both precision-scroll handler and SDL handler), and change `ScrollContainer` to not flip the X delta during `OnScroll` (only flip Y delta).

As an aside, I would generally recommend documenting the scroll directions and their respective `+`/`-` signs in `ScrollEvent.ScrollDelta`, as-is there's no place where it is explained what the consumer can expect when scrolling in a specific direction.

peppy:
This change works correctly on macOS precision/non-precision, windows non-precision. I'm pretty confident with the direction.

One thing to note is that we will still need to add shift-scroll support for windows scroll wheel direction override. It turns out this doesn't seems to be handled at an OS level as far as I can tell (someone please confirm as I don't have a native windows installation to 101% confirm at the moment).

peppy (approved):

SonoSooS:
> One thing to note is that we will still need to add shift-scroll support for windows scroll wheel direction override. It turns out this doesn't seems to be handled at an OS level as far as I can tell (someone please confirm as I don't have a native windows installation to 101% confirm at the moment).

I know that it's not implemented at OS level, but did attach a debugger to lazer just in case, and indeed SHIFT has no effect on the scroll messages sent to the window (besides bit2 being set when SHIFT is held).

frenzibyte:
> One thing to note is that we will still need to add shift-scroll support for windows scroll wheel direction override. It turns out this doesn't seems to be handled at an OS level as far as I can tell (has been confirmed). And if this holds correct, we need to decide whether to implement this before the events are sent to drawables, or just at `ScrollContainer`:

Would go with altering scroll before event is sent, mainly for consistency with macOS's behaviour, and to allow providing X delta values to controls which require it without handling `ShiftPressed` for each case.

bdach:
Just to double-confirm intent here: we're all aware that this is technically a consumer breaking change (because it's changing as to which half of the X axis is considered positive), yes? And we're fine with going through with it anyway? It probably needs a breaking change note anyway.

Incidentally, the only game-side usage of `ScrollDelta.X` is the editor, and the behaviour there appears to be *more* correct with this change applied, so maybe not really that huge of a deal.

Susko3:
I've updated the PR description with the breaking change notice.

As can be seen from the title of this PR, I think this is less of a change and more of a fix. If you look at it from the perspective of macOS, `ScrollContainer`, and the osu! editor, nothing has changed.

frenzibyte:
I still believe this will be a nuisance when writing "increase/decrease" scroll implementations for UI controls, but I sound like a broken record at this point, and I suppose this is not that big of a deal, for now at least.

frenzibyte (approved):

smoogipoo (approved):
Makes sense to me

smoogipoo:
Think it's probably fine to get this in without testing i(Pad)OS.

PR #5080, opened 2022-03-26
Bring Android framework to .NET 6

huoyaoyuan:
- [x] Requires https://github.com/ppy/osuTK/pull/75

### Development requirement

Run `dotnet workload install android` to install the necessary SDK components (may require `sudo`).
Can be changed to `dotnet workload restore` once there's no old projects (Xamarin) remaining.
The default Android SDK version changed from 29 to 31.
Xamarin toolchain in Visual Studio should be still required. I haven't tested if the managed components can be compiled without Xamarin installed.

The framework visual test works fine on my phone.

PR #5181, opened 2022-05-18
Fix Android project failing to build on CI

smoogipoo:
An interim fix until https://github.com/ppy/osu-framework/pull/5080 is merged, intended to fix the recent Android build failures such as https://github.com/ppy/osu-framework/runs/6464582633?check_suite_focus=true

I've changed the iOS project to match.

Haven't tested @peppy 's local build/deploy pipelines.

### DIFF
--- a/osu.Framework.Android/global.json
+++ b/osu.Framework.Android/global.json
@@ -3,9 +3,5 @@
     "allowPrerelease": false,
     "rollForward": "minor",
     "version": "3.1.100"
-  },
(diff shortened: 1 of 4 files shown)


bdach (approved):

peppy:
Should be okay in theory. For this specific one if it's going to fail it will be at appveyor's end, but easier to merge and figure that out if it happens.

peppy (approved):

PR #5186, opened 2022-05-19
Use precise delta values for SDL scroll events

frenzibyte:
Since SDL 2.0.18, there are new "precise" fields for returning floating-point scroll delta values, replacing the integral ones.

I intentionally still haven't set the `precise` bool in events as the game assumes the scroll amount should be smaller than a non-precise one, and it seems SDL isn't actually applying a scale ratio to the precise deltas at all ([not even in macOS/cocoa where `CGEventSourceGetPixelsPerLine` should be used](https://github.com/libsdl-org/SDL/blob/56665e1d9de9a5062b1933f4d51266b47f809f2a/src/video/cocoa/SDL_cocoamouse.m#L492-L525)).

macOS looks fine on non-precise scroll (i.e. mouse wheel), but pending testing on other platforms for double-confirmation:
 - [x] Windows
 - [x] Linux


smoogipoo:
For macOS we use or own precise scroll implementation: https://github.com/ppy/osu-framework/blob/1a7c8f22dc5054c9a03076a5ebb129a38ca290e8/osu.Framework/Platform/MacOS/MacOSWindow.cs#L35-L60

Curious if this can all be removed now.

I don't have a precise input device to test on Linux, but it does work with a non-precise device.

frenzibyte:
It still can't be removed as the macOS one applies a scale ratio to precise events for better scrolling experience, while SDL doesn't apply anything to its `preciseX`/`preciseY` fields.

We could consider moving the scale ratio to be applied globally rather than on macOS only, but would mean doing `preciseX != x` or `preciseX % 1 != 0` to ensure the scroll event is from precise device, and I'm unsure how that's going to behave for each platform.

Susko3:
I've tried this on a windows precision touchpad. Having precise scrolling definitely is an improvement, but it needs a little tweaking to feel right.

The scroll should be reported as `IsPrecise`. Else scrolling in a `ScrollContainer` is too "flowy" and not very responsive. The increased distance decay really helps here.

Setting only the `IsPrecise` flag is not correct, as the scrolling amount is scaled down by a factor of 8 (in `ScrollContainer`, `ScrollDistance` is 80 by default). Setting this back to `80` feels much closer to scrolling in native windows apps.

https://github.com/ppy/osu-framework/blob/0fdff74e4366536aa060c2a0cc3c073a12092ec3/osu.Framework/Graphics/Containers/ScrollContainer.cs#L385

This raises the question of why precision scrolling is even scaled down. `ScrollContainer` scales it down by 8, while osu!-side, [`VolumeMeter`](https://github.com/ppy/osu/blob/7b71fb860b39574c71cad2def3055750cec4bdad/osu.Game/Rulesets/Edit/DistancedHitObjectComposer.cs#L112) and [`Editor`](https://github.com/ppy/osu/blob/ab5c736da1a8d3ab99f85a4621bc97a254834dc2/osu.Game/Screens/Edit/Editor.cs#L540) scale it down by 10.

Having drawables locally scale the scroll seems wrong and error prone to me, as some usages are not scaling it. Best to apply this scale at `MacOSWindow` and have a consistent scroll delta for drawables.

To reiterate: not scaling produces correct results on Windows/SDL.

How to detect precise scrolling
---

No easy way to do it natively in windows. I've spent way too much time trying to find a win32 function for this. Tried [`GetPointerType`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getpointertype), [`GetCurrentInputMessageSource`](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getcurrentinputmessagesource), and all report precision touchpads as mice.

Just checking for each event if it's not a whole number may cause minor misbehaviour, as precision scrolling can still come out as a whole number (very rarely).

Have a flag `recentPreciseMouseScroll` that gets set on any non-integral scroll, and is only cleared after a short delay. This way, one-off integral inputs (in-between precise inputs) can be properly marked as `IsPrecise = true`.

peppy:
The flag exists because in general the feel of precise scrolls is a bit different. Also it may trigger a slightly different operation.

Easiest example I can think of is that in the editor while precise scrolling on the timeline, I'd expect it to scroll to-the-pixel accuracy, and potentially snap to the nearest divisor when released.

With a scroll wheel that has discrete notches, I'd expect each notch/increment to move to a divisor.

This kind of difference in behaviour is a pretty common and user expected IMO.

bdach:
That concrete example doesn't convince me too much, because the described UX seems more or less achievable even if the `IsPrecise` scroll was gone, as long as it would be replaced by mouse events with fractional scroll deltas - but I'll defer to your judgement since again I don't really have proper means to feel the difference (if any).

peppy:
The issue is that precision needs to be 1:1 with finger movement. It's hard to do that in fractional terms. It should be akin to dragging a finger on the screen, except you are dragging it on a surface that isn't the screen.

Example of precision scroll in safari on iOS:

https://user-images.githubusercontent.com/191335/169392005-89624872-c6e6-457e-992c-2281747142c9.mp4

(note that while not shown here, scroll will keep inertia if released during a flick. it's the same as touch so i'm sure you can understand what I mean)

Example of non-precision scroll:

https://user-images.githubusercontent.com/191335/169391899-32622cda-680e-4ee2-b8f5-eba4004d43d2.mov

(no inertia, only slight smoothing, custom acceleration applied)

frenzibyte:
Would say it's best to leave improving the general precise-scrolling experience on Windows/Linux to another PR, or close this one as incomplete if having the precise delta values is worse rather than better. 

Was only making this change after seeing that such field exists while not being used, and asked for platform testing to ensure those fields aren't potentially producing bad results.

peppy:
@Susko3 now that your adjustment PR is in, can you check this and confirm things feel good on windows?

peppy:
I've tested this myself on windows and all seems to be in shape. Roughly matches non-precision scroll rates.

peppy (approved):

Susko3:
Ideally precise scroll should use `IsPrecise = true`. But this is fine for now.

peppy:
Oh, that's weird. is there a reason we're not setting that to true? Because we don't have the information right now?

peppy:
I think it's worth at least adding a TODO regarding that fallacy.

Susko3:
> Oh, that's weird. is there a reason we're not setting that to true? Because we don't have the information right now?

I explained it in more detail here: https://github.com/ppy/osu-framework/pull/5186#issuecomment-1132064369

peppy (approved):

PR #5188, opened 2022-05-19
Add support for horizontal scrolling via Shift key 

### LINKED ISSUE #5183: Support horizontal scrolling via Shift key on non-Apple platforms
Continuing track of what was left in https://github.com/ppy/osu-framework/pull/5179#issuecomment-1128404268 + https://github.com/ppy/osu-framework/pull/5179#issuecomment-1128739985.

Susko3:
- Closes #5183

Brings other platforms in line with macOS, where horizontal scrolling is applied at OS level.
Doesn't apply to [precision touchpads](https://discord.com/channels/188630481301012481/589331078574112768/976734156576915516), as 2D scrolling is always supported there natively.

Shift+scroll is used in the osu! editor for faster scrolling, but this change doesn't affect that as the components are [added together](https://github.com/ppy/osu/blob/ab5c736da1a8d3ab99f85a4621bc97a254834dc2/osu.Game/Screens/Edit/Editor.cs#L530).

### DIFF
--- a/osu.Framework/Input/StateChanges/MouseScrollRelativeInput.cs
+++ b/osu.Framework/Input/StateChanges/MouseScrollRelativeInput.cs
@@ -32,6 +32,9 @@ namespace osu.Framework.Input.StateChanges
 
             if (Delta != Vector2.Zero)
             {
+                if (!IsPrecise && Delta.X == 0 && state.Keyboard.ShiftPressed)
+                    Delta = new Vector2(Delta.Y, 0);
+
                 var lastScroll = mouse.Scroll;
                 mouse.Scroll += Delta;
                 mouse.LastSource = this;


osu.Framework/Input/StateChanges/MouseScrollRelativeInput.cs
```diff
@@ -32,6 +32,9 @@ public void Apply(InputState state, IInputStateChangeHandler handler)
 
             if (Delta != Vector2.Zero)
             {
+                if (!IsPrecise && Delta.X == 0 && state.Keyboard.ShiftPressed)
+                    Delta = new Vector2(Delta.Y, 0);
```
bdach, line 36: This is going to be a rather minor point, but I'm not sure that this is going far enough - if shift is pressed, then all horizontal scrolls also may need to be blocked, too.

I'm going to try and explain why I think this with the following video:

https://user-images.githubusercontent.com/20418176/169400290-427c2636-05bd-4e5e-a66d-d3881264d37d.mp4

On this video, I am using a non-precise touchpad to scroll in diagonal directions. As it turns out, if I scroll across one diagonal, I get much less movement than if I scroll across the other diagonal. This is happening because in one diagonal direction the X axis scrolls and the Y-converted-into-X axis scrolls cancel each other out, and in the other diagonal directions, they match in sign and therefore magnify each other.

peppy: I think that's fine as is.

Susko3: > if shift is pressed, then all horizontal scrolls also may need to be blocked, too.

Don't think this is necessary, what you showed really seems like an edge case. Shift+horizontal scrolling may be desider, eg. in the editor with a mouse that has a [horizontal scroll wheel](https://m.media-amazon.com/images/I/613a-3jtieL._AC_SY450_.jpg).

peppy (approved):

osu.Framework/Input/StateChanges/MouseScrollRelativeInput.cs
frenzibyte: I mean, `IsPrecise = true` is supposed to handle this case by not allowing <kbd>Shift</kbd> to be pressed, no? it's currently not the case because Windows/Linux can't have it set to true right now.

PR #5192, opened 2022-05-23
Update dependencies

peppy:


peppy:
Blocking temporarily. osuTK should potentially not be bumped (see https://ptb.discord.com/channels/188630481301012481/589331078574112768/978201252280279080)

peppy:
Now only bumps SDL.

smoogipoo (approved):

PR #5195, opened 2022-05-23
Normalise precise scrolling values to match non-precise scrolling

Susko3:
As noticed in https://github.com/ppy/osu-framework/pull/5186#issuecomment-1132064369.

[Most](https://github.com/ppy/osu/blob/7b71fb860b39574c71cad2def3055750cec4bdad/osu.Game/Rulesets/Edit/DistancedHitObjectComposer.cs#L112) [consumers](https://github.com/ppy/osu/blob/ab5c736da1a8d3ab99f85a4621bc97a254834dc2/osu.Game/Screens/Edit/Editor.cs#L540) [scale](https://github.com/ppy/osu/blob/9416346c942ad20baa7b8000ebcb0a028a8c796e/osu.Game/Overlays/Volume/VolumeMeter.cs#L330-L332) down the scroll delta (by `1/10`, instead of the `ScrollContainer`'s `1/8`), so it makes sense to standardise this behaviour.

The goals of this PR are:
- normalise `ScrollDelta` so consumers don't have to check `IsPrecise` to correctly handle scroll events
- have consistent scaling across usages (always use `1/8` instead of some things using `1/10`)
  - osu! side usages should be checked (they will appear faster than before, but will match `ScrollContainer` scrolling speeds)
  - if this feels wrong, the scaling could be changed to `1/10`, but that will be across all usages
- allow for easier implementation of SDL precise scrolling, as those values are correct without scaling
- as a side effect, `ScrollDistance` is now respected for precise scrolling to allow more customization

Breaking changes
---

`ScrollDelta`/`ScrollAmount` is now normalised across precise and non-precise scrolling. Applying custom scale factors to `IsPrecise` scrolling is no longer necessary.


peppy:
This seems fine to go ahead.

@Susko3 do you already have osu!-side changes for this prepared?

Susko3:
osu!-side changes: https://github.com/ppy/osu/pull/18438.

peppy (approved):

PR #5196, opened 2022-05-23
Remove redundant installation of `libavformat` in Linux CI runs

jai-x:
Linux `libavformat` native libraries are now packaged with framework.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,15 +78,6 @@ jobs:
         with:
           dotnet-version: "6.0.x"
 
-      # FIXME: libavformat is not included in Ubuntu. Let's fix that.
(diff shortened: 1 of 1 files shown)


bdach:
CI failure here looks a bit unnerving: https://github.com/ppy/osu-framework/runs/6562894762?check_suite_focus=true

Tough to say if it's related, if I were to guess I'd say it's a 50/50, but I don't recall ever seeing that particular failure ever before...

bdach:
Re-ran the failing job, and now it looks to be fine, so I'm going to cautiously approve these but wait for a second call to get this in...

bdach (approved):

smoogipoo (approved):

PR #5204, opened 2022-05-26
Detect exclusive fullscreen on Windows

smoogipoo:
Using the API [`SHQueryUserNotificationState`](https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shqueryusernotificationstate).

The state enters `QUNS_RUNNING_D3D_FULL_SCREEN` even for OpenGL applications. Otherwise, like on my Optimus-enabled laptop, the state enters `QUNS_BUSY`.

You can't seem to be able to query this value right away - it takes some time after the mode change. I've set a 5s timer before the value is queried.

Testing:
- [x] Desktop + Radeon Linux
  - Message does not show.
- [x] Desktop + Windows (Radeon 5700XT)
  - Detects capable.
- [x] Desktop + Windows (GTX 1650)
  - Detects capable.
- [x] Optimus laptop + Windows (Intel HD Graphics 530 + GTX 960M) dGPU:
  - Detects incapable.
- [x] Laptop + Windows (Ryzen 7 5700U w/ Radeon Graphics, iGPU only)
  - Detects capable.

Not sure about these ones:

- Optimus laptop + Windows (Intel HD Graphics 530 + GTX 960M) iGPU:
  - Detects incapable.
- Laptop + Windows (Intel HD Graphics 4400, iGPU only)
  - Detects incapable.
- Desktop + Windows (Intel HD Graphics 2500, iGPU only)
  - Detects incapable.
- Parallels / VMWare
  - Detects incapable.

### DIFF
--- a/osu.Framework/Platform/Windows/WindowsWindow.cs
+++ b/osu.Framework/Platform/Windows/WindowsWindow.cs
@@ -65,27 +65,46 @@ namespace osu.Framework.Platform.Windows
 
             var cancellationSource = fullscreenCapabilityDetectionCancellationSource = new CancellationTokenSource();
 
-            Task.Delay(5000, cancellationSource.Token).ContinueWith(_ => ScheduleEvent(() =>
+            // 50 attempts, 100ms apart = run the detection for a total of 5 seconds before yielding an incapable state.
+            const int max_attempts = 50;
+            const int time_per_attempt = 100;
+            int attempts = 0;
+
+            runSingleAttempt();
+
+            void runSingleAttempt()
             {
-                if (cancellationSource.IsCancellationRequested || WindowState != WindowState.Fullscreen || !IsActive.Value)
-                    return;
-
-                try
+                Task.Delay(time_per_attempt, cancellationSource.Token).ContinueWith(_ => ScheduleEvent(() =>
                 {
-                    SHQueryUserNotificationState(out var notificationState);
+                    if (cancellationSource.IsCancellationRequested || WindowState != WindowState.Fullscreen || !IsActive.Value)
+                        return;
 
-                    fullscreenCapability.Value = notificationState == QueryUserNotificationState.QUNS_RUNNING_D3D_FULL_SCREEN
-                        ? Windows.FullscreenCapability.Capable
-                        : Windows.FullscreenCapability.Incapable;
+                    attempts++;
 
-                    Logger.Log($"Exclusive fullscreen capability: {fullscreenCapability.Value} ({notificationState})");
-                }
-                catch (Exception ex)
-                {
-                    Logger.Error(ex, "Failed to detect fullscreen capabilities.");
-                    fullscreenCapability.Value = Windows.FullscreenCapability.Capable;
-                }
-            }), cancellationSource.Token);
+                    try
+                    {
+                        SHQueryUserNotificationState(out var notificationState);
+
+                        var capability = notificationState == QueryUserNotificationState.QUNS_RUNNING_D3D_FULL_SCREEN
+                            ? Windows.FullscreenCapability.Capable
+                            : Windows.FullscreenCapability.Incapable;
+
+                        if (capability == Windows.FullscreenCapability.Incapable && attempts < max_attempts)
+                        {
+                            runSingleAttempt();
+                            return;
+                        }
+
+                        fullscreenCapability.Value = capability;
+                        Logger.Log($"Exclusive fullscreen capability: {fullscreenCapability.Value} ({notificationState})");
+                    }
+                    catch (Exception ex)
+                    {
+                        Logger.Error(ex, "Failed to detect fullscreen capabilities.");
+                        fullscreenCapability.Value = Windows.FullscreenCapability.Capable;
+                    }
+                }), cancellationSource.Token);
+            }
         }
 
         public override void Create()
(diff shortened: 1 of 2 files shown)


osu.Framework/Platform/Windows/WindowsWindow.cs
```diff
@@ -39,6 +46,46 @@ public WindowsWindow()
...
+                    fullscreenCapability.Value = notificationState == QueryUserNotificationState.QUNS_RUNNING_D3D_FULL_SCREEN
+                        ? Windows.FullscreenCapability.Capable
+                        : Windows.FullscreenCapability.Incapable;
+
+                    Logger.Log($"Exclusive fullscreen capability: {fullscreenCapability.Value}");
+                }
+                catch (Exception ex)
+                {
+                    Logger.Error(ex, "Failed to detect fullscreen capabilities.");
+                    fullscreenCapability.Value = Windows.FullscreenCapability.Capable;
```
jai-x, line 86:
```suggestion
                    fullscreenCapability.Value = Windows.FullscreenCapability.Unknown;
```

smoogipoo: I intended for this to fake a capable state to not display a message at the other side. It's an edge case that I don't expect will ever be hit.

smoogipoo:
I'm not sure about the desktop + iGPU case. Everything I can point to looks like it should be exclusive - FPS is higher, gamebar minimises, my remote desktop doesn't update visually... :/

smoogipoo:
Looking around at other places, I believe this is the best that can be done. It may or may not be correct on Intel iGPUs, but I have no idea anymore.

Here's one example (Chromium): https://chromium.googlesource.com/chromium/src/+/3710d00089e1b6d64a92a50a060b9581abd33fcd/chrome/browser/fullscreen_win.cc#13

peppy:
Fails in parallels (even though I am getting exclusive fullscreen):

```
[runtime] 2022-05-27 08:26:23 [verbose]: Exclusive fullscreen capability: Incapable (QUNS_BUSY)
```

Not a huge deal, just documenting.

peppy:
Is the five second delay something we could potentially make shorter by polling on a response? I'm a bit uneasy having a constant here rather than a retry-until-success.

smoogipoo:
What do you mean by polling on a response? The 5 second delay exists because the state isn't set right away upon changing the window mode.

peppy:
I guess there's no way to know when it's ready for consumption?

smoogipoo:
Correct, the assumption is that it’s correct after the window has been fullscreened in the foreground for 5 seconds.

peppy:
Just worried about edge cases where 5 seconds is either too short (who knows..?) or too long (maybe the user has since minimised the window or alt-tabbed away and changed the state).

smoogipoo:
For sure. As for what if the user had done XYZ, that’s why I scheduled it to the event scheduler and do all the checks a second time.

peppy:
I guess we could re-schedule the request in the case the window is not active?

Also in line with what I was proposing, would it be feasible to query the API every 200 or 500ms for up to 5 seconds, and requery if the return value doesn't imply `Capable` yet? To allow it to update faster if it can. As long as the winapi call isn't too expensive this seems like potentially a better option.

smoogipoo:
> I guess we could re-schedule the request in the case the window is not active?

Just to be clear, this is already happening. The task is already cancelled on window state and `IsActive` transitions, and is restarted when it enters the correct state once again.

> Also in line with what I was proposing, would it be feasible to query the API every 200 or 500ms for up to 5 seconds, and requery if the return value doesn't imply `Capable` yet?

You couldn't early return for `Incapable`. It's the `Capable` state that takes a while to be active. A typical chronological sequence for the notification state is something like:
```
[Windowed]
QUNS_ACCEPTS_NOTIFICATIONS
QUNS_ACCEPTS_NOTIFICATIONS
QUNS_ACCEPTS_NOTIFICATIONS
[Transition to fullscreen]
QUNS_ACCEPTS_NOTIFICATIONS
QUNS_BUSY
QUNS_BUSY
[Fullscreen]
QUNS_BUSY (incapable)
QUNS_RUNNING_D3D_FULL_SCREEN (capable)
```
But, we could early return if the `Capable` state is reached.

peppy:
> But, we could early return if the `Capable` state is reached.
That's precisely what I meant, yeah.

> Just to be clear, this is already happening. The task is already cancelled on window state and `IsActive` transitions, and is restarted when it enters the correct state once again.
Missed that.

osu.Framework/Platform/Windows/WindowsWindow.cs
```diff
@@ -65,27 +65,46 @@ private void detectFullscreenCapability(WindowState state)
...
+                            runSingleAttempt();
+                            return;
+                        }
+
+                        fullscreenCapability.Value = capability;
+                        Logger.Log($"Exclusive fullscreen capability: {fullscreenCapability.Value} ({notificationState})");
+                    }
+                    catch (Exception ex)
+                    {
+                        Logger.Error(ex, "Failed to detect fullscreen capabilities.");
```
peppy, line 103: This is going to log in a way that will display to users, how common is it?

smoogipoo: Should never happen.

peppy (approved):

PR #5207, opened 2022-05-27
Add IsNull()/IsNotNull() helper extensions

smoogipoo:
The following is a typical usage of binding to events of DI'd members:
```csharp
#nullable enable

{
    [Resolved]
    private SomeObject SomeDependency { get; set; } = null!;

    {
        base.LoadComplete();
        SomeDependency.SomeEvent += someMethod;
    }

    private void someMethod()
    {
    }

    {
        base.Dispose(disposing);

        if (SomeDependency != null) // WARNING: Expression is always true according to NRT rules.
            SomeDependency.SomeEvent -= someMethod;
    }
}
```
I propose the `IsNull()`/`IsNotNull()` extension methods in order to bypass the warning inside `Dispose()` and increase our coverage of NRT. The resultant code becomes:
```csharp
{
    base.Dispose(disposing);

    if (SomeDependency.IsNotNull())
        SomeDependency.SomeEvent -= someMethod;
}
```

### DIFF
--- a/osu.Framework/Extensions/ObjectExtensions/ObjectExtensions.cs
+++ b/osu.Framework/Extensions/ObjectExtensions/ObjectExtensions.cs
@@ -1,9 +1,9 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System.Diagnostics;
-
 #nullable enable
+
+using System.Diagnostics;
 
 namespace osu.Framework.Extensions.ObjectExtensions
 {
@@ -27,5 +27,15 @@ namespace osu.Framework.Extensions.ObjectExtensions
             Debug.Assert(obj != null);
             return obj;
         }
+
+        /// <summary>
+        /// If the given object is null.
+        /// </summary>
+        public static bool IsNull<T>(this T obj) => ReferenceEquals(obj, null);
+
+        /// <summary>
+        /// <c>true</c> if the given object is not null, <c>false</c> otherwise.
+        /// </summary>
+        public static bool IsNotNull<T>(this T obj) => !ReferenceEquals(obj, null);
     }
 }
--- a/osu.Framework/Extensions/ObjectExtensions/ObjectExtensions.cs
+++ b/osu.Framework/Extensions/ObjectExtensions/ObjectExtensions.cs
@@ -4,6 +4,7 @@
 #nullable enable
 
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 
 namespace osu.Framework.Extensions.ObjectExtensions
 {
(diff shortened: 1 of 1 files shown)


peppy (approved):

huoyaoyuan:
In #5099, I found such always non-null warning was too aggressive, especially when the nullability can't be fully expressed.
The practice of .NET Runtime is that extra null check won't harm except extremely hot path.
How do you think about turning off the Resharper warning instead?

smoogipoo:
Ah, I didn't notice this was only a resharper warning. That's a good point.

smoogipoo:
The only disable ReSharper offers is a global disable, which we definitely don't want and removes half the purpose of NRT (to simplify and remove null checks).

This is a very extreme edge case for us, because we can usually assume the dependency is present in practically all user cases _except_ `Dispose()`. Any other cases are very unlikely to be hit - e.g. we can't assume dependencies are present in `CreateChildDependencies()` either but we haven't hit this case yet across the entire osu! game.

The only alternative I can provide is to use `!` everywhere, but that doesn't seem to be getting much traction within the team/contributors.

frenzibyte:
> The only alternative I can provide is to use ! everywhere, but that doesn't seem to be getting much traction within the team/contributors.

Assuming this means marking dependencies as nullable and using `!` everywhere, and yeah I'm not sure about that approach, since a non-null dependency can only ever be null when accessed either before load or on disposal.

But I guess that can also become a major problem because it's easy to miss such cases during review... 

I don't have a strong opinion on this to be honest, and having `IsNull`/`IsNotNull` that should only ever be used when not able to do `!= null` feels off... I guess I could live with it if it's explicitly documented that the method is only meant to be used for such places, and never used anywhere else.

frenzibyte:
> The only alternative I can provide is to use ! everywhere, but that doesn't seem to be getting much traction within the team/contributors.

Assuming this means marking dependencies as nullable and using `!` everywhere, and yeah I'm not sure about that approach, since a non-null dependency can only ever be null when accessed either before load or on disposal.

But I guess that can also become a major problem because it's easy to miss such cases during review... 

I don't have a strong opinion on this to be honest, and having `IsNull`/`IsNotNull` that should only ever be used when not able to do `!= null` feels off... I guess I could live with it if it's explicitly documented that the method is only meant to be used for such places, and never used anywhere else.

bdach:
My two cents:

* Agree that disabling "redundant non-null" checks everywhere defeats most of the point of nullability annotations. Yes it is awkward in some cases, but it's better than the alternative of allowing anything and everything to be null-checked even if there is no conceivable way that it can ever be null. It's a code quality concern for me.
* I personally find the `!` operator to be sloppy - a "shut the compiler up" measure at best. Therefore, while it's unavoidable in some cases, I argue that its scope of use should be as narrow as possible. I would detest needing to sprinkle an exclamation point upon every single access to a DI'd component.

In light of both views I'm more than fine with these existing, except maybe I'd add a note in the xmldoc (probably in a remarks section) as to why they exist and what the intended usage is.

bdach (approved):

PR #5209, opened 2022-05-29
Avoid GC overhead during game load by only doing a full blocking collect at the end of a test run

peppy:
Previously this would be occurring many times during game load, causing crippling test performance in scenarios where the game takes considerable time to load.

### DIFF
--- a/osu.Framework/Testing/TestScene.cs
+++ b/osu.Framework/Testing/TestScene.cs
@@ -421,6 +421,10 @@ namespace osu.Framework.Testing
             checkForErrors();
             runner.RunTestBlocking(this);
             checkForErrors();
+
+            // Force any unobserved exceptions to fire against the current test run.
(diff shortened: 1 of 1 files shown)


peppy:
Blocking until I confirm this solves the issue fully.

peppy:
Tentatively unblocking. This improved performance greatly, but not fully to original levels (depending on the tests runnings.

Before the original `GC`:





Current `master`:



This change (local framework as `Release`):

peppy:
Can confirm the remaining overhead is garbage build-up over longer test runs. Likely something we can resolve osu!-side (still investigating).

bdach (approved):

bdach:
Have independently confirmed the speedups locally. Fingers crossed that this will translate to CI runtime.

peppy:
As an update, I've made some further progress towards restoring performance.

JetBrains Rider 2022-05-29 at 20 00 21@2x

I think there's still some further savings to be had, but getting there.

PR #5210, opened 2022-05-29
Implicitly confine the cursor when it can't leave to another display

Susko3:
- Closes https://github.com/ppy/osu/issues/18159

Previously, moving the mouse outside the window in fullscreen would warp it and give back control to the OS. The OS would then warp the cursor back into the game window. In this short period, the cursor would jitter between being inside and outside the window.

The solution provided is not perfect, the cursor can still jitter if `Sensitivity < 1`:
Eg. on a 1920x1080 display, it can go down to 1079.9999999999, but will warp to 1079 flat. Importantly, it'll never leave the window.

This can probably be fixed by checking if there are adjacent displays, and then warping if `> 1079` instead of `>= 1080`, but it's not really that important. (Always checking for `> 1079` would leave the cursor in an invalid position when it's in-between two displays—potentially preventing it from escaping the window.)


peppy:
Conceptually this makes sense. Going to get it in for battle testing.

peppy (approved):

PR #5211, opened 2022-05-29
Simplify exiting flow

### LINKED ISSUE #5193: `Host.Window.RequestClose()` is unsafe to call in single threaded mode
Calling `Window.RequestClose()` in single threaded mode will lead to `threadRunner` running a frame within a frame.
In multithreaded, [this ](https://github.com/ppy/osu-framework/blob/f932f8df053f0011d755c95ad9a2ed61b94d136b/osu.Framework/Platform/GameHost.cs#L413) will just block on the input thread until the update thread runs once and the outcome of `Exiting` is determined.

The following stack trace and log can be reproduced by pressing Alt+F4 (Cmd+W) while in single threaded mode.
Order of events:
- `UserInputManager` at `(1)` updates and does a foreach over `GetPendingInputs()`
- the keydown for Alt+F4 is handled:
  - `Game` handles it and calls `Window.RequestClose()`
  - `SDL2DesktopWindow` immediately calls `GameHost.OnExitRequested` (even though the call is [supposedly scheduled](https://github.com/ppy/osu-framework/blob/0b4522ebf8f47bdd9e19053e56ad08e94ac8912e/osu.Framework/Platform/SDL2DesktopWindow.cs#L591))
  - this [runs a new (update) frame](https://github.com/ppy/osu-framework/blob/f932f8df053f0011d755c95ad9a2ed61b94d136b/osu.Framework/Platform/GameHost.cs#L411) (this is already really questionable behaviour)
  - this frame runs just fine, `UserInputManager` at `(2)` updates (and clears the `GetPendingInputs()` list)
- the original Alt+F4 had finally been handled, so the foreach looks for the next pending input, and throws `InvalidOperationException: Collection was modified; enumeration operation may not execute.`

```cs
2022-05-23 15:59:47 [verbose]: Host execution state changed to Stopping
2022-05-23 15:59:47 [debug]: Pressed (Exit) handled by VisualTestGame.
2022-05-23 15:59:47 [debug]: KeyDownEvent(F4, False) handled by PlatformActionContainer.
2022-05-23 15:59:47 [error]: An unhandled error has occurred.
2022-05-23 15:59:47 [error]: System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
2022-05-23 15:59:47 [error]: at System.Collections.Generic.List`1.Enumerator.MoveNext()
2022-05-23 15:59:47 [error]: at osu.Framework.Input.InputManager.Update() in C:\Users\Mateo\code\git\osu-framework\osu.Framework\Input\InputManager.cs:line 443
2022-05-23 15:59:47 [error]: at osu.Framework.Input.PassThroughInputManager.Update() in C:\Users\Mateo\code\git\osu-framework\osu.Framework\Input\PassThroughInputManager.cs:line 143
2022-05-23 15:59:47 [error]: at osu.Framework.Graphics.Drawable.UpdateSubTree() in C:\Users\Mateo\code\git\osu-framework\osu.Framework\Graphics\Drawable.cs:line 483
2022-05-23 15:59:47 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree() in C:\Users\Mateo\code\git\osu-framework\osu.Framework\Graphics\Containers\CompositeDrawable.cs:line 899
2022-05-23 15:59:47 [error]: at osu.Framework.Platform.GameHost.UpdateFrame() in C:\Users\Mateo\code\git\osu-framework\osu.Framework\Platform\GameHost.cs:line 453
2022-05-23 15:59:47 [error]: at osu.Framework.Threading.GameThread.processFrame() in C:\Users\Mateo\code\git\osu-framework\osu.Framework\Threading\GameThread.cs:line 450
2022-05-23 15:59:47 [verbose]: Host execution state changed to Stopped
```

image

Properly scheduling the "event" fixes this. As scheduling simulates the window close button (called from the event loop, so no nested `threadRunner` frames).
```diff
diff --git a/osu.Framework/Platform/SDL2DesktopWindow.cs b/osu.Framework/Platform/SDL2DesktopWindow.cs
index b4da7aa52..dc7528d6b 100644
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -588,7 +588,7 @@ private void updateWindowSize()
         /// <summary>
         /// Attempts to close the window.
         /// </summary>
-        public void RequestClose() => ScheduleEvent(() =>
+        public void RequestClose() => EventScheduler.Add(() =>
         {
             if (ExitRequested?.Invoke() != true)
                 Close();
```

### LINKED ISSUE #5194: Simplify exiting flow
While debugging #5193, I was genuinely confused as to why the exiting process is so convoluted, requiring a lot of back and forth between `Game`, `GameHost` and `IWindow`. With each part seemingly forwarding events/return values, while `Host.Exit()` and `Game.OnExiting()` are the only things that are actually doing useful work.

This part of the current process is really worrisome. Currently, it fails with single threaded, and blocks on the input/main thread in multithreaded.

https://github.com/ppy/osu-framework/blob/f932f8df053f0011d755c95ad9a2ed61b94d136b/osu.Framework/Platform/GameHost.cs#L398-L421

I propose something like this in `Game`:

```cs
/// <summary>
/// Called when the window close (X) button or <see cref="PlatformAction.Exit"/> is pressed.
/// </summary>
/// <remarks>
/// Override to block the exit process as required.
/// Always run from the update thread.
/// </remarks>
public virtual void ExitRequested()
{
    Exit();
}
```

This somewhat reminds me of #5044, where I first had a complicated blocking/allowing process, but in the end it was simplified to something similar to the above.

Susko3:
- Resolves #5194 
- Closes #5193 
- Fixes [OSU-C](https://sentry.ppy.sh/organizations/ppy/issues/13/?project=2)

The 'exit requested' event is now processed like any other window event, leaving it to the `Game` to handle it and exit the host.

Since `Game.OnExitRequested()` is called the same as before, this PR has no breaking changes.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -86,9 +86,12 @@ namespace osu.Framework.Platform
         public event Action Deactivated;
 
         /// <summary>
-        /// Called when the host is requesting to exit. Return <c>true</c> to block the exit process.
+        /// Invoked when an exit was requested. Always invoked from the update thread.
         /// </summary>
-        public event Func<bool> Exiting;
+        /// <remarks>
+        /// Usually invoked when the window close (X) button or another platform-native exit action has been pressed.
+        /// </remarks>
+        public event Action ExitRequested;
 
         public event Action Exited;
 
@@ -401,30 +404,7 @@ namespace osu.Framework.Platform
 
         protected virtual void OnDeactivated() => UpdateThread.Scheduler.Add(() => Deactivated?.Invoke());
 
-        /// <returns>true to cancel</returns>
-        protected virtual bool OnExitRequested()
-        {
-            if (ExecutionState <= ExecutionState.Stopping) return false;
-
-            bool? response = null;
-
-            UpdateThread.Scheduler.Add(delegate { response = Exiting?.Invoke() == true; });
-
-            //wait for a potentially blocking response
-            while (!response.HasValue)
-            {
-                if (ThreadSafety.ExecutionMode == ExecutionMode.SingleThread)
-                    threadRunner.RunMainLoop();
-                else
-                    Thread.Sleep(1);
-            }
-
-            if (response.Value)
-                return true;
-
-            Exit();
-            return false;
-        }
+        protected void OnExitRequested() => UpdateThread.Scheduler.Add(() => ExitRequested?.Invoke());
 
         protected virtual void OnExited()
         {
--- a/osu.Framework/Platform/IWindow.cs
+++ b/osu.Framework/Platform/IWindow.cs
@@ -33,10 +33,10 @@ namespace osu.Framework.Platform
         void Create();
 
         /// <summary>
-        /// Return value decides whether we should intercept and cancel this exit (if possible).
+        /// Invoked when the window close (X) button or another platform-native exit action has been pressed.
         /// </summary>
         [CanBeNull]
-        event Func<bool> ExitRequested;
+        event Action ExitRequested;
 
         /// <summary>
         /// Invoked when the <see cref="IWindow"/> has closed.
@@ -151,6 +151,7 @@ namespace osu.Framework.Platform
         /// <summary>
         /// Attempts to close the window.
         /// </summary>
+        [Obsolete("Use Game.RequestExit() instead.")]
         void RequestClose();
 
         /// <summary>
--- a/osu.Framework/Platform/IWindow.cs
+++ b/osu.Framework/Platform/IWindow.cs
@@ -149,12 +149,6 @@ namespace osu.Framework.Platform
         void Close();
 
         /// <summary>
-        /// Attempts to close the window.
-        /// </summary>
-        [Obsolete("Use Game.RequestExit() instead.")]
-        void RequestClose();
-
-        /// <summary>
         /// Start the window's run loop.
         /// Is a blocking call on desktop platforms, and a non-blocking call on mobile platforms.
         /// </summary>
--- a/osu.Framework/Platform/OsuTKWindow.cs
+++ b/osu.Framework/Platform/OsuTKWindow.cs
@@ -120,6 +120,8 @@ namespace osu.Framework.Platform
 
             Closing += (sender, e) =>
             {
+                // always block a graceful exit as it's treated as a regular window event.
+                // the host will force-close the window if the game decides not to block the exit.
                 ExitRequested?.Invoke();
                 e.Cancel = true;
             };
--- a/osu.Framework/Platform/OsuTKWindow.cs
+++ b/osu.Framework/Platform/OsuTKWindow.cs
@@ -31,10 +31,9 @@ namespace osu.Framework.Platform
         public abstract IGraphicsContext Context { get; }
 
         /// <summary>
-        /// Return value decides whether we should intercept and cancel this exit (if possible).
+        /// Invoked when the window close (X) button or another platform-native exit action has been pressed.
         /// </summary>
-        [CanBeNull]
-        public event Func<bool> ExitRequested;
+        public event Action ExitRequested;
 
         /// <summary>
         /// Invoked when the <see cref="OsuTKWindow"/> has closed.
@@ -119,7 +118,11 @@ namespace osu.Framework.Platform
                 Resized?.Invoke();
             };
 
-            Closing += (sender, e) => e.Cancel = ExitRequested?.Invoke() ?? false;
+            Closing += (sender, e) =>
+            {
+                ExitRequested?.Invoke();
+                e.Cancel = true;
+            };
             Closed += (sender, e) => Exited?.Invoke();
 
             MouseEnter += (sender, args) => cursorInWindow.Value = true;
@@ -428,10 +431,10 @@ namespace osu.Framework.Platform
 
         public void Close() => OsuTKGameWindow.Close();
 
+        [Obsolete("Use Game.RequestExit() instead.")]
         public void RequestClose()
         {
-            if (ExitRequested?.Invoke() != true)
-                Close();
+            ExitRequested?.Invoke();
         }
 
         public void ProcessEvents() => OsuTKGameWindow.ProcessEvents();
--- a/osu.Framework/Platform/OsuTKWindow.cs
+++ b/osu.Framework/Platform/OsuTKWindow.cs
@@ -433,12 +433,6 @@ namespace osu.Framework.Platform
 
         public void Close() => OsuTKGameWindow.Close();
 
-        [Obsolete("Use Game.RequestExit() instead.")]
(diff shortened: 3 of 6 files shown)


osu.Framework/Platform/OsuTKWindow.cs
```diff
@@ -119,7 +118,11 @@ protected OsuTKWindow([NotNull] IGameWindow osuTKGameWindow)
                 Resized?.Invoke();
             };
 
-            Closing += (sender, e) => e.Cancel = ExitRequested?.Invoke() ?? false;
+            Closing += (sender, e) =>
+            {
+                ExitRequested?.Invoke();
+                e.Cancel = true;
```
bdach, line 124: Why is this unconditionally canceled now? I'm not sure if that's correct. Is there even a way to test this?

osu.Framework/Platform/GameHost.cs
```diff
@@ -401,30 +404,7 @@ void waitForThrow()
 
         protected virtual void OnDeactivated() => UpdateThread.Scheduler.Add(() => Deactivated?.Invoke());
 
-        /// <returns>true to cancel</returns>
-        protected virtual bool OnExitRequested()
```
bdach, line 405: Because `virtual` was dropped here, this is still technically a breaking change if anyone has overridden this. I suppose game doesn't appear to do so, so there's a decent change no project ever does.

bdach:
Also this does not play well with game-side tests inheriting from `OsuGameTestScene`. Scenario to check:

1. Set up a framework checkout game-side
2. Open visual test browser
3. Open any of the test scenes inheriting from `OsuGameTestScene`, for example `TestSceneOsuGame`
4. Attempt to quit
5.
```
[runtime] 2022-05-29 16:21:34 [error]: An unhandled error has occurred.
[runtime] 2022-05-29 16:21:34 [error]: System.ObjectDisposedException: Cannot access a disposed object.
[runtime] 2022-05-29 16:21:34 [error]: Object name: 'Cannot retrieve items for an already disposed SampleStore'.
[runtime] 2022-05-29 16:21:34 [error]: at osu.Framework.Audio.Sample.SampleStore.Get(String name) in /home/dachb/Dokumenty/opensource/osu-framework/osu.Framework/Audio/Sample/SampleStore.cs:line 38
[runtime] 2022-05-29 16:21:34 [error]: at osu.Game.Graphics.UserInterface.HoverSounds.load(AudioManager audio) in /home/dachb/Dokumenty/opensource/osu/osu.Game/Graphics/UserInterface/HoverSounds.cs:line 32
```

Without these changes, the inner game would take over the exit process and display an "are you sure you want to exit" popup.

Not sure if it's this diff's fault or the test's, but pointing it out as something potentially worth looking at.

osu.Framework/Platform/OsuTKWindow.cs
Susko3: The idea is that a graceful close is always blocked, as we delegate (forcefully) closing the window to `GameHost.Exit()`.

The window is [never closed](https://github.com/ppy/osu-framework/blob/41bdfc45ec19f992c24a3ff04a8d206b8c5b6c67/osu.Framework/Platform/GameHost.cs#L608-L629) on mobile platforms. Looking at osuTK, on both [iOS](https://github.com/ppy/osuTK/blob/f1299a502e5c15f1462bde6d00a0f285c811ccfe/src/osuTK.iOS/iOSGameView.cs#L952) and [Android](https://github.com/ppy/osuTK/blob/f1299a502e5c15f1462bde6d00a0f285c811ccfe/src/osuTK.Android/GameViewBase.cs#L504), the `Closing` event is never invoked.

peppy: I'd want an inline comment explaining this delegation at very least.

Susko3:
Only happens in multithreaded mode, exiting works fine is single-threaded.

This is not strictly related to testing nested games, with a local framework and the following osu! diff, the same stack trace can happen.

```diff
diff --git a/osu.Game/OsuGame.cs b/osu.Game/OsuGame.cs
index 785881d97a..68d6cdd2f8 100644
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -1169,7 +1169,7 @@ protected override bool OnExiting()
             if (introScreen?.DidLoadMenu == true && !(ScreenStack.CurrentScreen is IntroScreen))
             {
                 Scheduler.Add(introScreen.MakeCurrent);
-                return true;
+                return false;
             }

             return base.OnExiting();
```

The failure happens in [`ThreadRunner.Stop()`](https://github.com/ppy/osu-framework/blob/25f0945a29353a5464795d03ef16aed69e4be254/osu.Framework/Platform/GameHost.cs#L650). The update thread is doing work (in this case, loading drawables), and at the same time, the audio thread is [disposing all `AudioManager`s](https://github.com/ppy/osu-framework/blob/c8115c9be11b0fd76a2a224c67b4d57a9b02a598/osu.Framework/Threading/AudioThread.cs#L81-L107).

But this issue is not strictly related to this PR, it happens in prod as well. This PR and the above diff make it really easy to reproduce.

I've had a look at Sentry, and there where 10 issues about `Cannot access a disposed object. Object name: 'Cannot retrieve items for an already disposed SampleStore'.`, all happening after `Host execution state changed to Stopping` or even `Host execution state changed to Stopped`. Have merged them into OSU-G6.

osu.Framework/Platform/GameHost.cs
peppy: Let's hope no game has implemented their own `GameHost` yet... I'd almost argue we should seal the implementations...

osu.Framework/Platform/IWindow.cs
```diff
@@ -151,6 +151,7 @@ public interface IWindow : IDisposable
         /// <summary>
         /// Attempts to close the window.
         /// </summary>
+        [Obsolete("Use Game.RequestExit() instead.")]
```
peppy, line 154: I'm not sure this was ever used anywhere? I'd be in favour of removing it from the interface instead of obsoleting unless there's some hidden concern I'm missing.

peppy (commented):
Seems fine apart from concern about `RequestClose` requiring to still exist.

osu.Framework/Platform/IWindow.cs
Susko3: This was used as a way to [exit the game](https://github.com/ppy/osu-framework/blob/e1e9e8c95f316fb7f3890d8413867812dee4dd4f/osu.Framework/Game.cs#L384), this usage was back-to-front and part of the reason for the existence of this PR. Not used by any consumers, will remove.

Don't think this is worth mentioning as a breaking change.

peppy (approved):

PR #5212, opened 2022-05-30
Fix test performance overhead due to held references of previous `GameHost`s

peppy:
During test runs, realm could potentially hold a `SynchronizationContext` open beyond known `GameHost` lifetime. By extension, a previous game instance could be referenced via a long reference path.

I foresee this change may be hard to swallow in the current location (I'm not sure how correct it is to stop executing events that reached the `SynchronizationContext` when the `scheduler` goes away). Open to any other suggestions of where we should be resolving this. Potentially calling `UnregisterThread` on host shutdown? Using a `WeakReference` to the `GameThread`?

Note that I'm also looking into multiple other similar issues, including why realm is holding the reference in the first place (is probably our fault). If this direction or the others proposed above are not seen as correct, the follow-up PRs may inherently solve the same issue.

Parallels Desktop 2022-05-30 at 04 16 29
Parallels Desktop 2022-05-30 at 04 16 45
Parallels Desktop 2022-05-30 at 04 17 32

This reduces test runs of the osu! test suite considerably:

`master`:

this PR:

### DIFF
--- a/osu.Framework/Threading/GameThread.cs
+++ b/osu.Framework/Threading/GameThread.cs
@@ -481,7 +481,6 @@ namespace osu.Framework.Threading
                 Debug.Assert(state.Value == GameThreadState.Running);
                 Debug.Assert(exitState == GameThreadState.Exited || exitState == GameThreadState.Paused);
 
-                SynchronizationContext.SetSynchronizationContext(null);
                 synchronizationContext.Dispose();
 
                 Thread = null;
--- a/osu.Framework/Threading/GameThread.cs
+++ b/osu.Framework/Threading/GameThread.cs
@@ -481,7 +481,7 @@ namespace osu.Framework.Threading
                 Debug.Assert(state.Value == GameThreadState.Running);
                 Debug.Assert(exitState == GameThreadState.Exited || exitState == GameThreadState.Paused);
 
-                synchronizationContext.Dispose();
+                synchronizationContext.DisassociateGameThread();
 
                 Thread = null;
                 OnSuspended();
--- a/osu.Framework/Threading/GameThread.cs
+++ b/osu.Framework/Threading/GameThread.cs
@@ -481,6 +481,9 @@ namespace osu.Framework.Threading
                 Debug.Assert(state.Value == GameThreadState.Running);
                 Debug.Assert(exitState == GameThreadState.Exited || exitState == GameThreadState.Paused);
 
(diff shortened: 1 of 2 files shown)


peppy:
For a similar (but simpler) reference path, here's realm holding a reference to itself:

Parallels Desktop 2022-05-30 at 04 29 13

osu.Framework/Threading/GameThread.cs
```diff
@@ -481,6 +481,9 @@ private void setExitState(GameThreadState exitState)
                 Debug.Assert(state.Value == GameThreadState.Running);
                 Debug.Assert(exitState == GameThreadState.Exited || exitState == GameThreadState.Paused);
 
+                SynchronizationContext.SetSynchronizationContext(null);
```
smoogipoo, line 484: Doesn't this mean that events from Realm may potentially arrive on the non-update thread and trigger a whole different class of failures?

peppy: Do you mean for the paused state? I intended this to only be for shutdown, which may have in hindsight been a bad assumption given the dual purpose of this method.

Would your concerns still hold if this only happened on final exit?

smoogipoo: User triggers a shutdown of their game -> thread state changes to exiting -> game drawables are _maybe_ disposed, unknown what state they actually are in -> they bind to realm and receive a message from realm, but these messages arrive on a TPL thread -> the drawables want to do something like change text -> triggers `InvalidThreadForMutation` exception.

Something along that line.

peppy: It's a fair concern.

I think we can remove this line and the rest of this PR will still work, would that make it acceptable to you? Will re-test without it (I think I only added as a safety towards the end of investigation).

smoogipoo: That sounds like it would work.

peppy: Seems to still work as expected. Not sure about using `Dispose()` for this purpose, maybe a named method like `UnregisterGameThread` would be more appropriate?

smoogipoo: Tending to agree

smoogipoo (approved):
I think this looks fine, let's see how it works out.

peppy:
Tests run against the latest version of this (just to confirm everything still looks correct):

JetBrains Rider 2022-05-30 at 05 29 27

PR #5213, opened 2022-05-30
Statically cache font parsing results to reduce unit test overhead

peppy:
This doesn't improve test performance in local testing as much as I hoped for, but it does reduce alloc traffic and may convert to better savings on CI runs. Note that the benefit is not seen in o!f test runs due to far less fonts being loaded in the first place.

If this isn't seen as a quick merge for one reason or another I'm okay with closing instead.

Parallels Desktop 2022-05-30 at 04 35 59

Parallels Desktop 2022-05-30 at 04 34 49


smoogipoo (approved):
I _think_ this should be fine

PR #5214, opened 2022-05-30
Use pooled memory for texture uploads to avoid an extra memory allocation

peppy:
Running osu! navigation test namespace sees a reduction of around 6%

Parallels Desktop 2022-05-30 at 06 25 07

Parallels Desktop 2022-05-30 at 06 28 11

(Left is this PR)

### DIFF
--- a/osu.Framework/Graphics/Textures/TextureUpload.cs
+++ b/osu.Framework/Graphics/Textures/TextureUpload.cs
@@ -2,9 +2,9 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
+using osu.Framework.Extensions;
 using osu.Framework.Extensions.ImageExtensions;
 using osu.Framework.Graphics.OpenGL;
 using osu.Framework.Graphics.OpenGL.Buffers;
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Textures/TextureUpload.cs
```diff
@@ -86,12 +86,9 @@ public TextureUpload(Stream stream)
 
             try
             {
-                using (var m = new MemoryStream())
-                {
-                    stream.CopyTo(m);
-                    using (var stbiImage = Stbi.LoadFromMemory(m, 4))
-                        return Image.LoadPixelData(MemoryMarshal.Cast<byte, TPixel>(stbiImage.Data), stbiImage.Width, stbiImage.Height);
-                }
+                using (var buffer = SixLabors.ImageSharp.Configuration.Default.MemoryAllocator.Allocate<byte>((int)stream.Length))
```
smoogipoo, line 89: `stream` needs to be copied into this buffer, right?

smoogipoo (changes requested):

osu.Framework/Graphics/Textures/TextureUpload.cs
peppy: The funny part is I ran through visual tests to check this was working and... didn't see anything off.

smoogipoo: Because of the try-catch, for better or worse :sweat_smile:

osu.Framework/Graphics/Textures/TextureUpload.cs
```diff
@@ -86,10 +87,13 @@ public TextureUpload(Stream stream)
...
             try
             {
-                using (var m = new MemoryStream())
+                using (var buffer = SixLabors.ImageSharp.Configuration.Default.MemoryAllocator.Allocate<byte>((int)stream.Length))
                 {
-                    stream.CopyTo(m);
-                    using (var stbiImage = Stbi.LoadFromMemory(m, 4))
+                    int read = stream.Read(buffer.Memory.Span);
+
+                    Debug.Assert(read == stream.Length);
```
smoogipoo, line 94: Now that I think about it, you can use the new `ReadToFill()` extension method that takes care of this assertion for you.

smoogipoo (approved):

PR #5217, opened 2022-05-30
Synchronously exit `ThreadRunner` threads

### LINKED ISSUE #5216: `ThreadRunner.Stop()` is not safe in multithreaded mode
Alternate title: Audio thread can exit while update thread accesses its resources, causing an `ObjectDisposedException`
Sentry Issue: [OSU-G6](https://sentry.ppy.sh/organizations/ppy/issues/547/?referrer=github_integration)
Originally discovered in https://github.com/ppy/osu-framework/pull/5211#issuecomment-1141045506.

[`ThreadRunner.Stop()`](https://github.com/ppy/osu-framework/blob/f51ace62599d9e4c74bbfae9e9a8868d3609a3d0/osu.Framework/Platform/ThreadRunner.cs#L138-L161) will exit the threads "asynchronously", requesting each thread to exit, and will then block until all threads exit. This is a problem as `UpdateThread` is dependant on `AudioThread` (other combinations don't seem to throw, tested by manually exiting threads while the game is running). If `AudioThread` exits before `UpdateThread`, and a drawable requests a resource (in this case, `SampleStore.Get` in BDL) it'll fail with an `ObjectDisposedException`.

Interestingly, pausing threads is done [synchronously](https://github.com/ppy/osu-framework/blob/da15cdeef0f37685d4130e603c45c60fc96277a7/osu.Framework/Threading/GameThread.cs#L300-L318), in reverse order: `Draw`, `Update`, `Audio`, `Input`

https://github.com/ppy/osu-framework/blob/f51ace62599d9e4c74bbfae9e9a8868d3609a3d0/osu.Framework/Platform/ThreadRunner.cs#L210-L215

Changing `ThreadRunner.Stop()` / `GameThread.Exit()` to exit synchronously and in reverse order fixes the issue.

To repro:
- checkout #5211 (without it, the exit is scheduled, and the race condition is less likely to occur)
- apply this diff:
```diff
diff --git a/osu.Framework.Tests/Program.cs b/osu.Framework.Tests/Program.cs
index 8fb25e4db..5eddd0a62 100644
--- a/osu.Framework.Tests/Program.cs
+++ b/osu.Framework.Tests/Program.cs
@@ -3,6 +3,7 @@

 using System;
 using System.Linq;
+using System.Threading;
 using osu.Framework.Platform;

 namespace osu.Framework.Tests
@@ -20,7 +21,23 @@ public static void Main(string[] args)
                 if (benchmark)
                     host.Run(new AutomatedVisualTestGame());
                 else
-                    host.Run(new VisualTestGame());
+                    host.Run(new BuggyGame());
+            }
+        }
+
+        private class BuggyGame : TestGame
+        {
+            protected override bool OnExiting()
+            {
+                Scheduler.Add(() =>
+                {
+                    // pretend that a lot of work is done here, eg. adding and loading new drawables...
+                    // a.k.a. allow the audio thread to exit.
+                    Thread.Sleep(10);
+                    Audio.Samples.Get("long.mp3");
+                });
+
+                return false;
             }
         }
     }
```

Susko3:
- Closes #5216
- Fixes [OSU-G6](https://sentry.ppy.sh/organizations/ppy/issues/547/?referrer=github_integration)

I've omitted the timeout as it will break the synchronicity of the exiting process. If the timeout is something that is wanted, it can be added back, but `throw`ing instead of just logging (and should probably apply to the `while` loops as well).

### DIFF
--- a/osu.Framework/Threading/GameThread.cs
+++ b/osu.Framework/Threading/GameThread.cs
@@ -341,9 +341,22 @@ namespace osu.Framework.Threading
             }
             else
             {
-                while (state.Value != targetState)
-                    Thread.Sleep(1);
+                switch (targetState)
+                {
+                    case GameThreadState.Exited:
+                    case GameThreadState.Paused:
+                        Thread.Join();
+                        break;
+
+                    default:
+                        while (state.Value != targetState)
+                            Thread.Sleep(1);
+
+                        break;
+                }
             }
+
+            Debug.Assert(state.Value == targetState);
         }
 
         /// <summary>
--- a/osu.Framework/Threading/GameThread.cs
+++ b/osu.Framework/Threading/GameThread.cs
@@ -341,19 +341,8 @@ namespace osu.Framework.Threading
             }
             else
             {
-                switch (targetState)
-                {
-                    case GameThreadState.Exited:
-                    case GameThreadState.Paused:
-                        Thread.Join();
-                        break;
-
-                    default:
-                        while (state.Value != targetState)
-                            Thread.Sleep(1);
-
-                        break;
-                }
+                while (state.Value != targetState)
+                    Thread.Sleep(1);
             }
 
             Debug.Assert(state.Value == targetState);
--- a/osu.Framework/Platform/ThreadRunner.cs
+++ b/osu.Framework/Platform/ThreadRunner.cs
@@ -137,25 +137,12 @@ namespace osu.Framework.Platform
 
         public void Stop()
         {
-            const int thread_join_timeout = 30000;
-
-            Threads.ForEach(t => t.Exit());
-            Threads.Where(t => t.Running).ForEach(t =>
+            // exit in reverse order so AudioThread is exited last (UpdateThread depends on AudioThread)
+            Threads.Reverse().ForEach(t =>
             {
-                var thread = t.Thread;
-
-                if (thread == null)
-                {
-                    // has already been cleaned up (or never started)
-                    return;
-                }
-
-                if (!thread.Join(thread_join_timeout))
-                    Logger.Log($"Thread {t.Name} failed to exit in allocated time ({thread_join_timeout}ms).", LoggingTarget.Runtime, LogLevel.Important);
+                t.Exit();
+                t.WaitForState(GameThreadState.Exited);
             });
-
-            // as the input thread isn't actually handled by a thread, the above join does not necessarily mean it has been completed to an exiting state.
-            mainThread.WaitForState(GameThreadState.Exited);
 
             ThreadSafety.ResetAllForCurrentThread();
         }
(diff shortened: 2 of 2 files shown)


osu.Framework/Threading/GameThread.cs
```diff
@@ -341,9 +341,22 @@ internal void WaitForState(GameThreadState targetState)
             }
             else
             {
-                while (state.Value != targetState)
-                    Thread.Sleep(1);
+                switch (targetState)
+                {
+                    case GameThreadState.Exited:
+                    case GameThreadState.Paused:
+                        Thread.Join();
```
bdach, line 348: I kind of don't like that `Join()` is placed here... would prefer it to be back in `ThreadRunner` if feasible, for two reasons:

* `WaitForState()` doing anything else than waiting for a thread's state to change to the desired one seems wrong/weird
* `GameThread` joining... itself? is also reading really weird to me.

Susko3: > `WaitForState()` doing anything else than waiting for a thread's state to change to the desired one seems wrong/weird

It's just an optimization, taking advantage of the native thread [immediately exiting](https://github.com/ppy/osu-framework/blob/4b8f595630dd7281472a07619e0066bd1beeeceb/osu.Framework/Threading/GameThread.cs#L419-L425) once it reaches those states.

> `GameThread` joining... itself? is also reading really weird to me.

That's not really fair to say, as `WaitForState()` is never called from within a native thread.

peppy: I'm okay with the change in structure personally, but I'm not sure about dropping the timeout parameter. Was there a reason this couldn't remain as a safety measure? It was intended to avoid the game potentially getting in a state where the process is stuck and a user needs to forcefully quit from task manager.

peppy: I just saw your reasoning for doing so.

I'd much rather it returns as a `throw` if that's what is required.

Susko3:
Discovered a potential race condition in the current code, will push a fix.
Happens due to `Thread = null` and setting exit state being [far apart](https://github.com/ppy/osu-framework/blob/4b8f595630dd7281472a07619e0066bd1beeeceb/osu.Framework/Threading/GameThread.cs#L499-L512).

Not sure why `Thread = null` is set this early, I'd expect it to be done as a cleanup step when the native thread exits (eg. [here](https://github.com/ppy/osu-framework/blob/4b8f595630dd7281472a07619e0066bd1beeeceb/osu.Framework/Threading/GameThread.cs#L425)). Out of scope, I don't want to accidentally break anything.

peppy:
Looks correct now, but will wait for a second opinion.

bdach (approved):
makes sense to me

peppy (approved):

PR #5219, opened 2022-06-01
Fix convex polygon clipper edge cases

smoogipoo:
Should fix https://github.com/ppy/osu/issues/17215

Although I'm not entirely sure on what's causing the above issue, I believe it will be fixed based on running the new fuzzing test for 1hr with no failures. It generates many millions of polygons; concave, convex, lines, etc, and then performs clipping on them.

To be honest, this is a bit of a workaround because I can't really find that much documentation on these edge cases nor can I find documentation of others using a fixed buffer size as we're doing. I can only assume that using a fixed buffer size for this algorithm is incorrect.

There are two fixes applied here:
1. Increasing the total buffer size to assume every line can be intersected with every other.
2. Making collinear points not be counted as being "in the right half plane of the line".

# Concavity

Assuming that every line can be intersected with every other line implies concavity. This isn't something that this clipper can handle, but at least it will fail gracefully. This is the only fix required for https://github.com/ppy/osu/issues/17215.

# Sharing of vertices

One case where the buffer size is exceeded seems to be when a corner of the two polygons is shared. For example the clipping and subject polygons:
```
s: (-1, 2), (-0.1, -0.5), (0.1, 0)
c: (-10, 1), (0.1, 0), (-0.5, 2)
```
image

Generates the clipped vertices:
```
(-0.9726362, 1.9502487),
(0.1, 0),
(0.1, 0),
(0.100000024, 0),
(0.100000024, -2.1287374E-09),
(-0.29404524, 0.039014377),
(-0.98174477, 1.94929)
```
This happens because the following code can add duplicate vertices to the array in this case:
https://github.com/ppy/osu-framework/blob/7d5c372a591bb4275f667228f3ec2ecf25d19a1e/osu.Framework/Utils/ConvexPolygonClipper.cs#L127-L136
Handling this doesn't seem intuitive since the same issue can occur in the next iteration - the end point of one line becomes the start point of the next.

It should be noted at this point, that I'm using a pretty bog-standard implementation of the [Sutherland-Hodgman algorithm](https://en.wikipedia.org/wiki/Sutherland%E2%80%93Hodgman_algorithm) which has no mention of how to handle such cases.

# Collinear point edge case

I found a floating point precision edge case with lines that are _not_ collinear in one part of code that _are_ collinear in another part of code. This is demonstrated by the polygons and resultant clipping area:
```
p1: (-10,2),(-0.5,0.1),(0.5,0.5)
p2: (-0.1,2),(0,0),(0.5,-0.1)
clip: (0,0),(-0.028776987,0.5755397),(0.3212766,0.525532),(0.3461539,0.43846154),(-0.0147058815,0.29411763),(0,0),(0.5,-0.1)
```
image

Note how there's a section below the black-shaded part that is not in the intersection of the two areas. This is caused by the bottom-most lines of the two polygons:
```
1: (-0.5, 0.1) -> (-10, 2)
2: (0.5, -0.1) -> (0, 0)
```
image

Graphically, these appear as if they're collinear, however based on the calculations of `Vector2Extensions.InRightHalfPlaneOf()`, you'd find that the start of line 1 is to the right of line 2, and the end of line 1 is to the left of line 2. This is a pretty "basic" float precision issue, but has the catastrophic consequence of making it seem like we should be able to find a single point of intersection between these two lines.

As a bit of background to understand this, think about two line segments that you could extend in virtual space to infinity. The line segments themselves don't need to be intersecting, but we can say that there's a single point of intersection if one point extended to infinity from one line appears to _left_ of the second line, and another point from that same line appears to the _right_ of the second line.  
If this is true, then we can find a parameter `t` representing the point of intersection of the two line segments. This parameter can be outside the range `[0,1]` for if the two line segments aren't intersecting.

By this theory we expect to be able to find the parameter `t` for the case above. This is done via the `Line.TryIntersectWith()` method, but what you'll find is that this method returns a parameter of 0! This is hitting the `denominator = 0` case of this method, implying that the lines are collinear and that we don't have a single point of intersection:
https://github.com/ppy/osu-framework/blob/7d5c372a591bb4275f667228f3ec2ecf25d19a1e/osu.Framework/Graphics/Primitives/Line.cs#L106-L121

This has bad effects for the convex clipper leading to the above result, but collinear lines don't matter all that much in the first place. So the fix I've gone with is to simply stop treating collinear points as being in the right half-plane of lines, resulting in the fixed clipping area:
image

### DIFF
--- a/osu.Framework.Tests/Polygons/ConvexPolygonClipperFuzzingTest.cs
+++ b/osu.Framework.Tests/Polygons/ConvexPolygonClipperFuzzingTest.cs
@@ -56,9 +56,6 @@ namespace osu.Framework.Tests.Polygons
         {
             Task[] tasks = new Task[parallelism];
 
-            uint countValid = 0;
-            uint countInvalid = 0;
-
             for (int i = 0; i < tasks.Length; i++)
             {
                 tasks[i] = Task.Factory.StartNew(() =>
@@ -82,10 +79,7 @@ namespace osu.Framework.Tests.Polygons
 
                         try
                         {
-                            if (clip(poly1, poly2).Length > 0)
-                                Interlocked.Increment(ref countValid);
-                            else
-                                Interlocked.Increment(ref countInvalid);
+                            clip(poly1, poly2);
                         }
                         catch (Exception ex)
                         {
@@ -95,10 +89,7 @@ namespace osu.Framework.Tests.Polygons
 
                         try
                         {
-                            if (clip(poly2, poly1).Length > 0)
-                                Interlocked.Increment(ref countValid);
-                            else
-                                Interlocked.Increment(ref countInvalid);
+                            clip(poly2, poly1);
                         }
                         catch (Exception ex)
                         {
--- a/osu.Framework.Tests/Polygons/ConvexPolygonClipperFuzzingTest.cs
+++ b/osu.Framework.Tests/Polygons/ConvexPolygonClipperFuzzingTest.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 using NUnit.Framework;
 using osu.Framework.Graphics.Primitives;
--- a/osu.Framework.Tests/Polygons/ConvexPolygonClipperFuzzingTest.cs
+++ b/osu.Framework.Tests/Polygons/ConvexPolygonClipperFuzzingTest.cs
@@ -1,0 +1,126 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using osu.Framework.Graphics.Primitives;
+using osu.Framework.Utils;
+using osuTK;
+
+namespace osu.Framework.Tests.Polygons
+{
+    [TestFixture]
+    [Ignore("This test will never complete if working correctly. Test manually if required.")]
+    public class ConvexPolygonClipperFuzzingTest
+    {
+        private const int parallelism = 4;
+
+        private static readonly float[] possible_values =
+        {
+            float.NegativeInfinity,
+            float.MinValue,
+            float.MinValue / 2,
+            -10.0f,
+            -2.0f,
+            -1.0f,
+            -0.5f,
+            -0.1f,
+            0.0f,
+            float.Epsilon,
+            0.1f,
+            0.5f,
+            1.0f,
+            2.0f,
+            10.0f,
+            float.MaxValue / 2,
+            float.MaxValue,
+            float.PositiveInfinity,
+            float.NaN,
+        };
+
+        private static readonly int[] possible_sizes =
+        {
+            3,
+            4,
+            5,
+            8,
+            16
+        };
+
+        [Test]
+        public void RunTest()
+        {
+            Task[] tasks = new Task[parallelism];
+
+            uint countValid = 0;
+            uint countInvalid = 0;
+
+            for (int i = 0; i < tasks.Length; i++)
+            {
+                tasks[i] = Task.Factory.StartNew(() =>
+                {
+                    while (true)
+                    {
+                        int count1 = getRand(possible_sizes);
+                        int count2 = getRand(possible_sizes);
+
+                        HashSet<Vector2> vertices1 = new HashSet<Vector2>();
+                        HashSet<Vector2> vertices2 = new HashSet<Vector2>();
+
+                        while (vertices1.Count < count1)
+                            vertices1.Add(new Vector2(getRand(possible_values), getRand(possible_values)));
+
+                        while (vertices2.Count < count2)
+                            vertices2.Add(new Vector2(getRand(possible_values), getRand(possible_values)));
+
+                        SimpleConvexPolygon poly1 = new SimpleConvexPolygon(vertices1.ToArray());
+                        SimpleConvexPolygon poly2 = new SimpleConvexPolygon(vertices2.ToArray());
+
+                        try
+                        {
+                            if (clip(poly1, poly2).Length > 0)
+                                Interlocked.Increment(ref countValid);
+                            else
+                                Interlocked.Increment(ref countInvalid);
+                        }
+                        catch (Exception ex)
+                        {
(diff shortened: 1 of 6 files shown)


osu.Framework.Tests/Polygons/ConvexPolygonClipperFuzzingTest.cs
```diff
@@ -0,0 +1,126 @@
...
+                        while (vertices2.Count < count2)
+                            vertices2.Add(new Vector2(getRand(possible_values), getRand(possible_values)));
+
+                        SimpleConvexPolygon poly1 = new SimpleConvexPolygon(vertices1.ToArray());
+                        SimpleConvexPolygon poly2 = new SimpleConvexPolygon(vertices2.ToArray());
+
+                        try
+                        {
+                            if (clip(poly1, poly2).Length > 0)
+                                Interlocked.Increment(ref countValid);
```
peppy, line 86: Might as well make these `int` to fix android/mono weirdness.

smoogipoo: These aren't actually required, only existed for debugging purposes. Will remove instead.

peppy (approved):
Seems non-offensive and probably more correct than it was.

bdach:
Ok, so my geometry is rusty, but here goes...

## Collinear point edge case / `InRightHalfPlaneOf` fix

I have nothing to add here. Full agreement with that change.

## Shared vertex edge case fix

As far as I can tell, that change is somewhat conservative and can be improved upon. Short essay how and why follows below, beware:

If I'm not wrong, excluding the duplicate points is actually easier than you state in the OP. The devil is in the following check, in the code path you mention in the OP:

https://github.com/ppy/osu-framework/blob/7d5c372a591bb4275f667228f3ec2ecf25d19a1e/osu.Framework/Utils/ConvexPolygonClipper.cs#L129

Note that:

* With the changes in this PR, `InRightHalfPlaneOf()` returns true for any point that is on the right side of the half-plane given by `clipEdge`, but does not *lie* on `clipEdge`. (Not sure if that holds up to formal definitions of a half-plane but let's truck on based on that assumption.)
* This means that the `!InRightHalfPlaneOf()` check above returns true for any point on the left side of the half-plane given by `clipEdge`, *and* for any point that *lies* on `clipEdge`, which is a big factor in why the duplicate points are getting added. If the segments share a common endpoint, then the common endpoint will lie on both edges, triggering Sutherland-Hodgman to output two vertices in that given iteration for no real reason, thinking that the segment crosses the half-plane while in fact it does not.

  However, if the conditional is changed to `InLeftHalfPlaneOf()` (which does not currently exist, but is basically `InRightHalfPlaneOf()` but with `< 0` at the end), then that problem disappears and duplicate points no longer get added.

The diff for the above is:

```diff
diff --git a/osu.Framework/Graphics/Vector2Extensions.cs b/osu.Framework/Graphics/Vector2Extensions.cs
index 69bb00d0ce..b1eb862499 100644
--- a/osu.Framework/Graphics/Vector2Extensions.cs
+++ b/osu.Framework/Graphics/Vector2Extensions.cs
@@ -108,5 +108,10 @@ public static float GetOrientation(in ReadOnlySpan<Vector2> vertices)
         public static bool InRightHalfPlaneOf(this Vector2 point, in Line line)
             => (line.EndPoint.X - line.StartPoint.X) * (point.Y - line.StartPoint.Y)
                 - (line.EndPoint.Y - line.StartPoint.Y) * (point.X - line.StartPoint.X) < 0;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool InLeftHalfPlaneOf(this Vector2 point, in Line line)
+            => (line.EndPoint.X - line.StartPoint.X) * (point.Y - line.StartPoint.Y)
+                - (line.EndPoint.Y - line.StartPoint.Y) * (point.X - line.StartPoint.X) > 0;
     }
 }
diff --git a/osu.Framework/Utils/ConvexPolygonClipper.cs b/osu.Framework/Utils/ConvexPolygonClipper.cs
index a0de4e79e7..eaa7a0659a 100644
--- a/osu.Framework/Utils/ConvexPolygonClipper.cs
+++ b/osu.Framework/Utils/ConvexPolygonClipper.cs
@@ -127,7 +127,7 @@ private void outputVertices(ref Vector2 startVertex, ref Vector2 endVertex, in L
         {
             if (endVertex.InRightHalfPlaneOf(clipEdge))
             {
-                if (!startVertex.InRightHalfPlaneOf(clipEdge))
+                if (startVertex.InLeftHalfPlaneOf(clipEdge))
                 {
                     clipEdge.TryIntersectWith(ref startVertex, ref endVertex, out float t);
                     buffer[bufferIndex++] = clipEdge.At(t);
```

## On the upper bound of vertices output

Eliminating the duplicates theoretically allows to improve the upper bound of the number of vertices in the intersection. I *believe* that as long as both polygons are convex, then the sufficient number of vertices would be equal to `subjectPolygon.GetVertices().Length + clipPolygon.GetVertices().Length`.

(The proof would go by induction - each iteration of Sutherland-Hodgman, i.e. each intersection of the subject polygon with each of the clipping half-planes can add at most 1 new vertex to the result, which means that the most you can get is what you started with, plus the number of iterations, which is the same as the sum of the number of vertices in both polygons. I checked this upper bound with the above diff applied, and the test cases added here passed.)

However, given that the above only applies to convex polygons, then it's probably best to err on the side of caution and keep the conservative bound used here.

---

I'll leave it up to you as to whether you wish to apply any of the above. This diff is most likely fine as provided anyway.

peppy:
Going to get this in as-is due to the severity of this issue (causes sporadic game crashes ~80 times a day with current lazer usage) and leave refinements as proposed above for a second pass.

smoogipoo:
> which means that the most you can get is what you started with, plus the number of iterations

If we assert that both polygons are convex, then I believe you can only intersect each edge of the subject polygon at most twice? Because intersecting it 3 times would imply that the clipping polygon is concave.

That's why the original bound was `subject.Count * 2`.

Anyway, that aside, I like your other suggestion very much and will apply it in a followup PR.

bdach:
> I believe you can only intersect each edge of the subject polygon at most twice?

This makes sense, but consider that during Sutherland-Hodgman new edges are created during the iterative half-plane clipping. Also the fact that the bound is not symmetrical makes me think it's wrong - if you switch the subject and clip polygons, you should get the same upper bound (because the set intersection relation is symmetric), which in the case of the original bound is not the case.

A more correct bound based on this premise would likely be `2 * max(clipPolygon.Vertices.Count(), subjectPolygon.Vertices.Count())`, and the one I gave is always guaranteed to be smaller than that.

PR #5221, opened 2022-06-02
Throw `OperationCanceledException` when `Perform`ing a `WebRequest`after `Abort`

peppy:
This allows identifying this scenario in client usage. This can be used to fix the issue seen at https://github.com/ppy/osu/issues/18524#issuecomment-1144428852, where the flow is:

- Game makes request
- Game cancels request before `Perform` fires its `Task`'d `PerformAsync` 
- `WebRequest` reports invalid operation

I'm not sure about the exception types here. Should we also throw an `OperationCanceled` from the `Perform` flow, rather than `TaskCanceled` that comes from .NET itself?

osu-side relevant fix would be something like https://github.com/ppy/osu/compare/master...peppy:web-request-canceled-fix?expand=1

### DIFF
--- a/osu.Framework/IO/Network/WebRequest.cs
+++ b/osu.Framework/IO/Network/WebRequest.cs
@@ -238,7 +238,12 @@ namespace osu.Framework.IO.Network
         public async Task PerformAsync(CancellationToken cancellationToken = default)
         {
             if (Completed)
+            {
+                if (Aborted)
+                    throw new OperationCanceledException($"The {nameof(WebRequest)} has been aborted.");
+
                 throw new InvalidOperationException($"The {nameof(WebRequest)} has already been run.");
+            }
 
             try
             {
--- a/osu.Framework.Tests/IO/TestWebRequest.cs
+++ b/osu.Framework.Tests/IO/TestWebRequest.cs
@@ -295,9 +295,7 @@ namespace osu.Framework.Tests.IO
             bool hasThrown = false;
             request.Failed += exception => hasThrown = exception != null;
 
-#pragma warning disable 4014
-            request.PerformAsync();
-#pragma warning restore 4014
+            Task.Run(() => request.PerformAsync());
 
             Assert.DoesNotThrow(request.Abort);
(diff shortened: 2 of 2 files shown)


osu.Framework/IO/Network/WebRequest.cs
```diff
@@ -238,7 +238,12 @@ public byte[] GetResponseData()
         public async Task PerformAsync(CancellationToken cancellationToken = default)
         {
             if (Completed)
+            {
+                if (Aborted)
+                    throw new OperationCanceledException($"The {nameof(WebRequest)} has been aborted.");
```
bdach, line 243: This can pass the cancellation token:

```suggestion
                    throw new OperationCanceledException($"The {nameof(WebRequest)} has been aborted.", cancellationToken);
```

Not sure it does much except net style points, though. In theory the runtime is supposed to be checking whether the exception's token is the same as the task's and then decide whether the state of the task should be `Canceled` or `Faulted` (would be `Canceled` with the proposed change), but I can't actually observe this happening.

bdach (approved):
Seems alright.

bdach:
> Should we also throw an `OperationCanceled` from the `Perform` flow, rather than `TaskCanceled` that comes from .NET itself?

I don't think this is a huge issue if I'm honest. It's quite common to expect the former in async flows and the latter in sync flows. And `TaskCanceledException` derives from `OperationCanceledException` anyhow.

osu.Framework/IO/Network/WebRequest.cs
smoogipoo: You're probably not observing it happening because you'd want to use the cancellation token of the request itself (i.e. the ones inside `internalPerform()`).

It would be nice to have, but I think it's fine not having it for now, as you'd still get a faulted if it's aborted during the request... Notice how `internalPerform()` calls `Complete()` on abort with an exception, which is then rethrown outwards from `Complete()`.

smoogipoo (approved):

Issue #5155, opened 2022-05-03
It is currently not possible to declare custom interfaces as "positional" or "non-positional" event handlers


Flutterish:
Currently, there is a closed set of interfaces, which when implemented mark the drawable as able to handle (non)positional input (in `HandleInputCache`). It is not possible to add your own.
I propose, instead of this list, a root interface such as `I(Non)PositionalInputHandler`, which would simply be an empty marker interface which all the current interfaces would inherit. This will allow new, custom interfaces which rely upon positional or non-positional input queues to be handled without needing to override `Handles(Non)PositionalInput` of every drawable which uses them.
My particular use case is, I was trying to add a FileDrop handler interface, but found I had to override `HandlesPositionalInput` of every drawable which uses it.

peppy:
I think it sounds okay. But I'd also say that we need a proper item-drop pattern within o!f itself at some point (not sure if @Flutterish is interested in helping make that happen).

Flutterish:
If you both think this proposal is okay, should I make a pr or do you want to handle it yourself?

---

> But I'd also say that we need a proper item-drop pattern within o!f itself at some point (not sure if @Flutterish is interested in helping make that happen).

If I'm honest, I'm not exactly sure how to do it properly. If you look at how I did it in [o!f-designer](https://github.com/Flutterish/osu-framework-designer) and the [file drop input manager](https://github.com/Flutterish/osu-framework-designer/blob/main/OsuFrameworkDesigner/OsuFrameworkDesigner.Game/Containers/DesignerInputManager.cs) and the [base game](https://github.com/Flutterish/osu-framework-designer/blob/main/OsuFrameworkDesigner/OsuFrameworkDesigner.Game/Containers/DesignerInputManager.cs), its a bit hacky and improper. The `IWindow` class has no standard file drop event (and there is no event for when its "hovering" with the dragged file either), and the input manager binds into the base game event (which probably should be a game host event) and then it also needs to check if there aren't other input managers that handle that event or otherwise they will be duplicated and I'm not sure how to solve that particular problem. I'm not sure I'm familiar enough with the o!f input system to do this in a proper manner.

bdach:
I'd still want @smoogipoo's opinion at least on this sort of change. So may be fine to open a PR but I wouldn't expect it to be merged this week.

As an aside:

> and there is no event for when its "hovering" with the dragged file either

[...]

PR #5222, opened 2022-06-02
Allow interfaces to declare themselves as (non)positional input handlers

Flutterish:
See #5155

### DIFF
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -27,9 +27,7 @@ using JetBrains.Annotations;
 using osu.Framework.Bindables;
 using osu.Framework.Development;
 using osu.Framework.Extensions.EnumExtensions;
-using osu.Framework.Graphics.Cursor;
 using osu.Framework.Graphics.OpenGL;
-using osu.Framework.Input.Bindings;
 using osu.Framework.Input.Events;
 using osu.Framework.Input.States;
 using osu.Framework.Layout;
@@ -37,6 +35,7 @@ using osu.Framework.Testing;
 using osu.Framework.Utils;
 using osuTK.Input;
 using Container = osu.Framework.Graphics.Containers.Container;
+using osu.Framework.Input.Handlers;
 
 namespace osu.Framework.Graphics
 {
@@ -2389,18 +2388,8 @@ namespace osu.Framework.Graphics
                 nameof(OnMidiUp)
             };
 
-            private static readonly Type[] positional_input_interfaces =
-            {
-                typeof(IHasTooltip),
-                typeof(IHasCustomTooltip),
-                typeof(IHasContextMenu),
-                typeof(IHasPopover),
-            };
-
-            private static readonly Type[] non_positional_input_interfaces =
-            {
-                typeof(IKeyBindingHandler),
-            };
+            private static readonly Type positional_interface = typeof(IPositionalInputHandler);
+            private static readonly Type non_positional_interface = typeof(INonPositionalInputHandler);
 
             private static readonly string[] positional_input_properties =
             {
@@ -2445,14 +2434,10 @@ namespace osu.Framework.Graphics
                         return true;
                 }
 
-                var inputInterfaces = positional ? positional_input_interfaces : non_positional_input_interfaces;
-
-                foreach (var inputInterface in inputInterfaces)
-                {
-                    // check if this type implements any interface which requires a drawable to handle input.
-                    if (inputInterface.IsAssignableFrom(type))
-                        return true;
-                }
+                // check if this type implements any interface which requires a drawable to handle input.
+                var inputInterface = positional ? positional_interface : non_positional_interface;
+                if (inputInterface.IsAssignableFrom(type))
+                    return true;
 
                 string[] inputProperties = positional ? positional_input_properties : non_positional_input_properties;
 
--- a/osu.Framework/Input/Handlers/INonPositionalInputHandler.cs
+++ b/osu.Framework/Input/Handlers/INonPositionalInputHandler.cs
@@ -1,0 +1,13 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework.Input.Handlers
+{
+    /// <summary>
+    /// Marker interface for interfaces which depend on non-positional input.
+    /// All drawables which implement this interface will have <see cref="Graphics.Drawable.HandleNonPositionalInput"/> set to <see langword="true"/> by default.
+    /// </summary>
+    public interface INonPositionalInputHandler
+    {
+    }
+}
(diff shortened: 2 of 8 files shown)


osu.Framework/Input/Handlers/INonPositionalInputHandler.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework.Input.Handlers
+{
+    /// <summary>
+    /// Marker interface for interfaces which depend on non-positional input.
+    /// All drawables which implement this interface will have <see cref="Graphics.Drawable.HandleNonPositionalInput"/> set to <see langword="true"/> by default.
+    /// </summary>
+    public interface INonPositionalInputHandler
```
Susko3, line 10: This is not the right namespace for the interfaces, `osu.Framework.Input.Handlers` is for platform-dependent `InputHandler`s. `osu.Framework.Input` might be a better place for these.

Also the name doesn't fit in IMO. The functionality of this is similar to `IHandleGlobalKeyboardInput` and `IRequireHighFrequencyMousePosition`. While the likes of `-Handler` (eg. `IKeyBindingHandler`, `IScrollBindingHandler`) have methods.

I suggest `IHandleNonPositionalInput` and `IHandlePositionalInput`.

Susko3 (changes requested):

osu.Framework/Input/Handlers/INonPositionalInputHandler.cs
Flutterish: Understood, and I agree with the suggested names. I'll apply these changes

osu.Framework/Input/IHandleNonPositionalInput.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Framework.Input
+{
+    /// <summary>
+    /// Marker interface for interfaces which depend on non-positional input.
+    /// All drawables which implement this interface will have <see cref="Graphics.Drawable.HandleNonPositionalInput"/> set to <see langword="true"/> by default.
+    /// </summary>
+    public interface IHandleNonPositionalInput
```
frenzibyte, line 10:
```suggestion
    /// <summary>
    /// Marker interface for drawables which depend on non-positional input.
    /// All drawables which implement this interface will have <see cref="Graphics.Drawable.HandleNonPositionalInput"/> set to <see langword="true"/> by default.
    /// </summary>
    public interface IHandleNonPositionalInput : IDrawable
```
(same applies to other interface)

osu.Framework/Graphics/Drawable.cs
```diff
@@ -2389,18 +2387,8 @@ private static class HandleInputCache
...
-                typeof(IHasContextMenu),
-                typeof(IHasPopover),
-            };
-
-            private static readonly Type[] non_positional_input_interfaces =
-            {
-                typeof(IKeyBindingHandler),
-            };
+            private static readonly Type positional_interface = typeof(IHandlePositionalInput);
+            private static readonly Type non_positional_interface = typeof(IHandleNonPositionalInput);
```
frenzibyte, line 2391: nit: Would prefer naming this back to `positional_input_interface`/`non_positional_input_interface`, or just inline now that it's a single constant interface.

osu.Framework/Input/IHandleNonPositionalInput.cs
frenzibyte, line 10: The name doesn't really make sense at face value. Maybe `IRequiresNonPositionalInput` at least (same applies to other interface).

Flutterish: The interface is specifically meant to be inherited directly by interfaces and not drawables, which is why the docs say "interfaces". `IHandleGlobalKeyboardInput` does not force the implementer to be an `IDrawable`, so I dont think this one should either.

osu.Framework/Graphics/Drawable.cs
Flutterish: I  can change change the name, but I believe inline would make it resolve the type each time rather than being cached.

frenzibyte: Sure, renaming is enough.

osu.Framework/Input/IHandleNonPositionalInput.cs
Flutterish: Following the logic of `IHandleGlobalKeyboardInput`, which doesn't either, I think this is fine? If you look at Susko3's comment, they pointed out "handler" interfaces which have an implementation have "Handler" at the end rather than "IHandle" which doesn't.

frenzibyte: I'm not sure about that, whichever implements `IHandleNonPositionalInput`/`IHandlePositionalInput` must surely be a drawable type, otherwise the interface is pointless for that type. Also, I can imagine drawables implementing that interface directly, e.g. abstract classes with certain logic that requires input.

frenzibyte: `IHandleGlobalKeyboardInput` is used to make a drawable which handles input exempted from `OverlayContainer`'s blocking of input. I don't see the correlation between that and this interface, i.e. I don't see how this interface has anything to do with handling input, it only has to do with forcing a drawable to receive input, because it requires it.

smoogipoo: I think `IHandle` is fine.

osu.Framework/Graphics/Drawable.cs
smoogipoo: `typeof()` is compile-time. Inlining is fine.

osu.Framework/Input/IHandleNonPositionalInput.cs
Flutterish: If so, then the global keyboard interface should have it added too.

Such drawables surely could just override Handles(Non)PositionalInput themselves, or implement an interfce/override a handler method though, no? I imagine the whole class which assigns default values to whether a drawable type does or doesnt handle a given input type is more of a confusion avioder/optimisation rather than *the* way to define it

frenzibyte: > If so, then the global keyboard interface should have it added too.

I see no problem with doing so? `IRequireHighFrequencyMousePosition` already implements `IDrawable` for what it's worth.

Flutterish: > I see no problem with doing so?

Oh, that wasn't a counterpoint, I was just saying that should this interface have it added, I should add it to the global keyboard one too.

frenzibyte: To reiterate: I'm suggesting for this interface to only ever be inherited by drawables or interfaces that only drawables can inherit, because it's the only place where this interface can function.

If this is focusing on the xmldoc part of the suggestion, then it can be reworded in some form. Having it as `Marker interface for interfaces` definitely reads weird, would suggest `Marker interface for <see cref="IDrawable"/>-based interfaces` at the very least.

PR #5223, opened 2022-06-02
Clear textbox selection when setting via `Text` or `Current`

### LINKED ISSUE #5170: `TextBox` selection behaves weirdly when changing via `Current`
Easiest example is a control point popover in osu! editor (with https://github.com/ppy/osu/pull/18220), which has a textbox-slider component bound to one current bindable, and when the textbox has selection in it, results in the following issue:

https://user-images.githubusercontent.com/22781491/167806663-52dedab2-72f9-42f9-b346-cdfddf7322ca.mov

It's expected at that point to clear selection, rather than keeping it or halfing it weirdly as shown in the video.

Susko3:
- Closes #5170

Not really sure what the existing code was doing. `insertString()` will move the caret to the expected position (end of text).
Includes some very rudimentary tests.


smoogipoo (approved):

peppy:
I think the original code was attempting to maintain a selection on value change, but it doesn't make sense to attempt to do this.

peppy (approved):

PR #5224, opened 2022-06-03
Support nullable reference types for dependency injection

smoogipoo:
```csharp
#nullable enable

class MyDrawable : Drawable
{
    // Non-null dependency.
    [Resolved]
    private object NonNullDependency { get; set; } = null!;

    // Nullable dependency.
    // [Resolved(CanBeNull = true)] : Not required anymore!
    [Resolved]
    private object? NullableDependency { get; set; } 

    // [BackgroundDependencyLoader(true)] : Not required anymore!
    [BackgroundDependencyLoader]
    private void load(object param1, object? param2)
    {
    }
}
```
This is only supplementary behaviour - supplying the nullable parameters of `[Resolved]` and `[BackgroundDependencyLoader]` preserves the previous functionality of injecting nulls.

### DIFF
--- a/osu.Framework.Tests/Dependencies/DependencyContainerTest.cs
+++ b/osu.Framework.Tests/Dependencies/DependencyContainerTest.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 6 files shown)


peppy:
This is very cool addition.

peppy (approved):

PR #5225, opened 2022-06-03
Prevent ConvexPolygonClipper from outputting duplicate vertices

smoogipoo:
As suggested in https://github.com/ppy/osu-framework/pull/5219#issuecomment-1145356172

### DIFF
--- a/osu.Framework.Tests/Polygons/ConvexPolygonClippingTest.cs
+++ b/osu.Framework.Tests/Polygons/ConvexPolygonClippingTest.cs
@@ -241,6 +241,16 @@ namespace osu.Framework.Tests.Polygons
                 false);
         }
 
+        [Test]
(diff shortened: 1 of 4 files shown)


bdach:
Well the tests have decisively proved me wrong that this was a simple change... The failures are pretty much all cases wherein the clip and subject polygon share vertices which actually need to be output in the intersection, in which case it is *wanted* for Sutherland-Hodgman to output two points in an iteration.

I don't know off the top of my head what the fix would be here. I'm not even sure there is one without adding some sort of heuristic. Probably safest and easiest to close this after all, sorry for the detour - should have run tests before suggesting  :(

smoogipoo:
Hah, my bad too. I think I only tested the newest test. Haven't looked into it but let's close this one for now.

PR #5227, opened 2022-06-03
Improve scrolling/dragging experience on singular scroll containers

frenzibyte:
- Closes https://github.com/ppy/osu/issues/18550

In #5118, scroll containers have changed to not accept drags/scrolls when containing a greater horizontal component, but that produces a bad UX as seen in https://github.com/ppy/osu/issues/18550, especially on dragging where starting a horizontal drag then continuing vertically no longer works.

Now the new behaviour will only happen when the scroll container is nested inside of another scroll container that accepts the other direction.

### DIFF
--- a/osu.Framework/Graphics/Containers/ScrollContainer.cs
+++ b/osu.Framework/Graphics/Containers/ScrollContainer.cs
@@ -13,7 +13,7 @@ using osuTK.Input;
 
 namespace osu.Framework.Graphics.Containers
 {
-    public abstract class ScrollContainer<T> : Container<T>, DelayedLoadWrapper.IOnScreenOptimisingContainer, IKeyBindingHandler<PlatformAction>
+    public abstract class ScrollContainer<T> : Container<T>, IScrollContainer, DelayedLoadWrapper.IOnScreenOptimisingContainer, IKeyBindingHandler<PlatformAction>
         where T : Drawable
     {
         /// <summary>
@@ -164,10 +164,7 @@ namespace osu.Framework.Graphics.Containers
             }
         }
 
-        /// <summary>
-        /// The direction in which scrolling is supported.
-        /// </summary>
-        protected readonly Direction ScrollDirection;
+        public Direction ScrollDirection { get; }
 
         /// <summary>
         /// The direction in which scrolling is supported, converted to an int for array index lookups.
@@ -247,6 +244,7 @@ namespace osu.Framework.Graphics.Containers
 
             if (dragWasMostlyHorizontal != (ScrollDirection == Direction.Horizontal))
                 return false;
+            }
 
             lastDragTime = Time.Current;
             averageDragDelta = averageDragTime = 0;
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/Containers/ScrollContainer.cs
```diff
@@ -368,12 +369,17 @@ protected override bool OnScroll(ScrollEvent e)
             if (Content.AliveInternalChildren.Count == 0)
                 return false;
 
-            bool scrollWasMostlyHorizontal = Math.Abs(e.ScrollDelta.X) > Math.Abs(e.ScrollDelta.Y);
+            var parentScrollContainer = this.FindClosestParent<IScrollContainer>();
```
peppy, line 372: I think this traversal lookup should be done once in `load` or `LoadComplete` rather than on every scroll event. For precision scrolls this could be called 1,000 times per second.

peppy (changes requested):

osu.Framework/Graphics/Containers/ScrollContainer.cs
peppy:
```csharp
diff --git a/osu.Framework/Graphics/Containers/ScrollContainer.cs b/osu.Framework/Graphics/Containers/ScrollContainer.cs
index 0f850bf9a..7c1244e0f 100644
--- a/osu.Framework/Graphics/Containers/ScrollContainer.cs
+++ b/osu.Framework/Graphics/Containers/ScrollContainer.cs
@@ -235,13 +235,17 @@ private void updatePadding()
             }
         }
 
+        {
+            parentScrollContainer = this.FindClosestParent<IScrollContainer>();
+        }
+
         protected override bool OnDragStart(DragStartEvent e)
         {
             if (IsDragging || e.Button != MouseButton.Left || Content.AliveInternalChildren.Count == 0)
                 return false;
 
-            var parentScrollContainer = this.FindClosestParent<IScrollContainer>();
-
             if (parentScrollContainer != null && parentScrollContainer.ScrollDirection != ScrollDirection)
             {
                 bool dragWasMostlyHorizontal = Math.Abs(e.Delta.X) > Math.Abs(e.Delta.Y);
@@ -304,6 +308,8 @@ protected override bool OnMouseDown(MouseDownEvent e)
 
         private bool dragBlocksClick;
 
+        private IScrollContainer parentScrollContainer;
+
         public override bool DragBlocksClick => dragBlocksClick;
 
         protected override void OnDrag(DragEvent e)
@@ -369,8 +375,6 @@ protected override bool OnScroll(ScrollEvent e)
             if (Content.AliveInternalChildren.Count == 0)
                 return false;
 
-            var parentScrollContainer = this.FindClosestParent<IScrollContainer>();
-
             if (parentScrollContainer != null && parentScrollContainer.ScrollDirection != ScrollDirection)
             {
                 bool scrollWasMostlyHorizontal = Math.Abs(e.ScrollDelta.X) > Math.Abs(e.ScrollDelta.Y);

```

This looks to work okay.

frenzibyte: Right, was mostly calling it at the point of event to handle the case where a `ScrollContainer` changes from one parent to another, but precise scrolling is a good point here. It looks to be possible to handle the concern I had using a `LayoutValue(Invalidation.Parent)`.

peppy (approved):

PR #5228, opened 2022-06-04
Fix synchronisation context disassociating scheduler on execution mode changes

frenzibyte:
- Closes https://github.com/ppy/osu/issues/18569

[As discussed in discord](https://discord.com/channels/188630481301012481/188630652340404224/982717832649375764), the scheduler shouldn't be disassociated when changing execution modes, since the `GameThread` instance is preserved on that case, only the underlying thread is reinstantiated.


smoogipoo (approved):
Looks reasonable

peppy (approved):

PR #5235, opened 2022-06-06
Rename "unlimited" to "basically unlimited" and add sane frame limiter

peppy:
This was already discussed internally and agreed on.

I'd also like to change the remaining options to be more in line with stable (ie. have an "optimised" setting that keeps update fps around 500 regardless of refresh rate), but that's for another day.

---

Games using osu!framework can generally run at *very* high frame rates when not much is going on.

This can be counter-productive due to the induced allocation and GPU overhead.
 - Allocation overhead can lead to excess garbage collection
 - GPU overhead can lead to unexpected pipeline blocking (and stutters as a result).
   Also, in general graphics card manufacturers do not test their hardware at insane frame rates and therefore drivers are not optimised to handle this kind of throughput.
 - We only harvest input at 1000hz, so running any higher has zero benefits.

We limit things to the same rate we poll input at, to keep both gamers and their systems happy and (more) stutter-free.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -51,6 +51,12 @@ namespace osu.Framework.Platform
     {
         public IWindow Window { get; private set; }
 
+        /// <summary>
+        /// Whether "unlimited" frame limiter should be allowed to exceed sane limits.
+        /// Only use this for benchmarking purposes (see <see cref="maximum_sane_fps"/> for further reasoning).
+        /// </summary>
+        public bool AllowBenchmarkUnlimitedFrames { get; set; }
+
         protected FrameworkDebugConfigManager DebugConfig { get; private set; }
 
         protected FrameworkConfigManager Config { get; private set; }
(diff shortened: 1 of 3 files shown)


ekrctb:
> What does that have to do with anything?

I'm saying these are the same kind of the logic. We know that higher the FPS lower the draw latency (on average), and the argument ">60fps doesn't help" is not valid. Hence the PR comment's "We only harvest input at 1000hz, so running any higher has zero benefits" is invalid as well.

> Are you claiming that anyone is going to notice one millisecond of latency?

I'm pretty sure I can differentiate 1ms input latency difference. I think osu!catch players are especially sensitive to the input latency due to how it is played.

bdach:
To the first point, I argue that capping framerate at 60Hz and capping it at 1000Hz is not comparable because of the nearly two magnitudes of difference between those numbers.

To the second - aside from sincerely doubting that anyone can detect 1ms of latency (which I suppose is subjective), I argue that the difference of capping at 1000 and not capping is minuscule at best for 99% of users and there is more to be gained by introducing the cap and hopefully reducing the frequency of user reports spurred by misuse of no frame limiter.

osu.Framework/Platform/GameHost.cs
```diff
@@ -51,6 +51,12 @@ public abstract class GameHost : IIpcHost, IDisposable
     {
         public IWindow Window { get; private set; }
 
+        /// <summary>
+        /// Whether "unlimited" frame limiter should be allowed to exceed sane limits.
+        /// Only use this for benchmarking purposes (see <see cref="maximum_sane_fps"/> for further reasoning).
+        /// </summary>
+        public bool AllowBenchmarkUnlimitedFrames { get; set; }
```
bdach, line 58: I guess exposing this makes sense, but was this intended to be used somewhere, or mostly manually? Going by xmldoc, one place where setting this would plausibly make sense is [`GameBenchmark.ManualGameHost`](https://github.com/ppy/osu-framework/blob/d83dbaf029df124a59b974fdf18b67bd0da6e14b/osu.Framework.Benchmarks/GameBenchmark.cs#L49).

Finding the actual place that enforces the framerate is a bit of a trip, but as far as I can tell it would be `ThrottledFrameClock` that does it, so I think that the flag will have an actual effect in the benchmarks.

peppy:
I was waiting for someone to call that out 😅. I originally wrote "negligible" but it's so close to zero (due to what @bdach touched on) that I felt "no" was better.

osu.Framework/Platform/GameHost.cs
peppy: Yep, this is intended as a manual flag for developers to bypass only when it makes sense (very rarely – even in profiling using a limited will give more correct results).

peppy:
> Ideally we want to eliminate code that is allocating every frame (unbounded allocation per real time) so GC shouldn't be negatively affected by higher FPS.

The primary concern is not allocations. We already are in a great state on that front (next to or at zero when not moving the mouse).

GPU blocking and weirdness caused by high frame throughputs is the main issue we see, and the primary reason for this change. Overheating, coil whine, and bad drivers causing pipeline blocks that take longer when running high frame rates are all very common.

smoogipoo (approved):

szmarczak:
> We only harvest input at 1000hz, so running any higher has zero benefits.

While this might be true for rendering, running a 8kHz mouse does make a difference: https://youtu.be/gOQNRvJbpmk?t=420

JetMouseWasTaken:
The reason I need higher fps is that I'd like to reduce screen tearing. When I'm moving my mouse at top speed, it cuts my cursor completely in half. I bought a viper 8k so it would get split by at most 1/8 of the width of the cursor, but for it to work I also need at least 8000 fps. I was gonna upgrade my really old gpu, but now I know that won't do anything to the fps
I'm not gonna turn on vsync because it adds latency, and causes a latency gradient. A gradient of 2 ms wouldn't be very noticable, but it's something I'd like to avoid.
I'll try disabling it in the code, making it think it's not a dev build, and seeing if I can log in on it.

peppy:
8000 fps will mean 8000 tears.

I think what you're looking for is vertical sync.

PR #5236, opened 2022-06-07
Properly set `IsPrecise` for SDL scroll events

Susko3:
Continuation of https://github.com/ppy/osu-framework/pull/5186#issuecomment-1132064369.

Tested on a windows precision touchpad, and with a regular mouse.


Susko3:
> I guess this method of detection will mean the flag _could_ be incorrect if the first event to arrive is from a precision source but exactly `1.00`. Very very unlikely in practice, but just confirming.

Correct, could be `false` for the first scroll, but I expect the following deltas to be precise.

Also, due to how the debounce is implemented, any scroll within 100 ms of `SDL_Init` will be considered precise, not worth mentioning as the window is not created/visible in the first 100 ms.

peppy:
I've tested this on windows and it works as expected (did not test transitioning to mouse from precision, but I think the code is sound enough).

peppy (approved):

PR #5238, opened 2022-06-10
Let BASS read stream data ahead of time to reduce audio stuttering

### LINKED ISSUE #5237: iOS audio stuttering
Can likely be fixed by tweaking the bass buffer settings, needs further testing.

Any changes applied should probably only for iOS to begin with (to avoid any potential latency incurrence on other platforms)

Video contains three noticeable stutters, but it can get much worse or better, depending on luck.

https://user-images.githubusercontent.com/191335/172516183-4bcac448-25d0-48ba-beda-9ed34ed1b928.MP4

frenzibyte:
- Closes #5237 

After deep investigation, enabling mixer buffering avoids some of the stuttering at the cost of playback latency (see [explanation](https://discord.com/channels/188630481301012481/589331078574112768/984492141919240192)), therefore it is not desired to enable it for o!f and we can still continue without it.

The stuttering occurs when opening song select or switching between beatmaps in osu!, and looks to generally happen due to gen1/gen2 collections (meaning you can manually cause a stutter by pressing "Clear all caches" in game settings).

Instead, a flag is now used for streams (and samples with `SampleChannelStream`) for having data buffered internally ahead-of-time to read without getting stalled by CLR being stuck due to GC collection or otherwise.

It is also generally recommended for low latency output as well, but most useful for having data buffered on BASS's side:

See [flags documentation in `BASS_StreamCreateFileUser`](http://www.un4seen.com/doc/#bass/BASS_StreamCreateFileUser.html).

The buffer can also be configured using [`BASS_CONFIG_ASYNCFILE_BUFFER`](http://www.un4seen.com/doc/#bass/BASS_CONFIG_ASYNCFILE_BUFFER.html).

~~This brings up the question of whether we should still keep `AsyncBufferStream`, but I'm unsure about removing it immediately in this PR as it would require further testing.~~ (maybe not, see [discord](https://discord.com/channels/188630481301012481/589331078574112768/984973320887881728))

### DIFF
--- a/osu.Framework/Audio/Sample/SampleChannelBass.cs
+++ b/osu.Framework/Audio/Sample/SampleChannelBass.cs
@@ -189,7 +189,7 @@ namespace osu.Framework.Audio.Sample
             if (hasChannel)
                 return;
 
-            channel = Bass.SampleGetChannel(sample.SampleId, BassFlags.SampleChannelStream | BassFlags.Decode);
+            channel = Bass.SampleGetChannel(sample.SampleId, BassFlags.SampleChannelStream | BassFlags.Decode | BassFlags.AsyncFile);
 
             if (!hasChannel)
                 return;
--- a/osu.Framework/Audio/Track/TrackBass.cs
+++ b/osu.Framework/Audio/Track/TrackBass.cs
@@ -151,7 +151,7 @@ namespace osu.Framework.Audio.Track
 
             fileCallbacks = new FileCallbacks(new DataStreamFileProcedures(dataStream));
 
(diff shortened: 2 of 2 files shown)


peppy:
I remember seeing this flag introduced in recent years and wondering if it could replace our buffering. Definitely warrants more investigation in the future.

Are you able to link the bass documentation for the flag in the PR description to provide a paper trail?

frenzibyte:
There's multiple places where this flag is documented, but I've added the `BASS_StreamCreateFileUser` page which contains the important part of what it's for.

peppy (approved):

PR #5240, opened 2022-06-13
Refactor `TextureStore` to improve nesting support

### LINKED ISSUE #5220: Adding `TextureStore`s to the `Game.Textures` does not work as expected
Creating a new `TextureStore` (which should use its own atlas, filtering mode etc.) and adding it to the game textures, only adds it as a `TextureUpload` lookup source, rather than a texture lookup source.
I understand that the game `TextureStore` is a `ResourceStore<TextureUpload>` with the added ability to get textures from its resources, however I find it unintuitive that adding a `TextureStore` to it doesnt use it as a texture lookup source.

My idea how to fix this would be to make `TextureStore` an `IResourceStore<Texture>` instead of a texture upload store, with an ability to add texture upload stores to it using the same method (`AddResourceStore(IResourceStore<TextureUpload>)`). `Game.Textures` would be replaced with some class which would aggregate `IResourceStore<Texture>`s. There still remains the problem of it being a breaking change since many BDLs request a `TextureStore` which would be replaced

frenzibyte:
- Closes #5220 

This refactors `TextureStore` into explicitly having two types of stores:
 - Texture data lookup store, in the form of `IResourceStore<TextureUpload>`, used for looking up texture data and loading a `Texture` for.
 - Nested texture store, in the form of `ITextureStore`, used as a secondary store if the `Texture` could not be found in the main store.

Which, in turn, removes the `TextureStore : ResourceStore<TextureUpload>` inheritance and forces `TextureStore`s to not be used only for lookup when nested in another store.

I've kept the name `AddStore` for nesting texture stores inline with `ResourceStore`, but can consider renaming to `AddNested` if that reads better. 

PR'ing this for feedback on direction.

# vNext

## `TextureStore.AddStore/RemoveStore` have been split into "store" and "texture source" methods

In an effort to make the texture API easier to comprehend, the `AddStore`/`RemoveStore` methods have been split into two pairs:
 - `AddTextureSource`/`RemoveTextureSource`, for adding/removing texture data lookup sources (i.e. `TextureLoaderStore`s)
 - `AddStore`/`RemoveStore`, for adding/removing `TextureStore`s.

Any existing usages of `AddStore` for adding lookup sources must be changed to use `AddTextureSource` instead.


osu.Framework/Graphics/Textures/TextureStore.cs
```diff
@@ -34,24 +39,57 @@ public class TextureStore : ResourceStore<TextureUpload>
...
             }
         }
 
-        private Texture getTexture(string name, WrapMode wrapModeS = WrapMode.None, WrapMode wrapModeT = WrapMode.None) => loadRaw(base.Get(name), wrapModeS, wrapModeT);
+        /// <summary>
+        /// Adds a texture data lookup source to load <see cref="Texture"/>s with.
+        /// </summary>
+        /// <remarks>
+        /// Lookup sources must be wrapped with a <see cref="TextureLoaderStore"/> to provide <see cref="TextureUpload"/>s.
+        /// Other <see cref="TextureUpload"/> stores are also allowed to be used as lookup sources.
```
peppy, line 63: Aren't these two lines contradictory? If this needs to be said it should be worded more like "Lookup sources can be implemented easily using a `TextureLoaderStore` to provide the final `TextureUpload`."

osu.Framework/Graphics/Textures/TextureStore.cs
```diff
@@ -34,24 +39,57 @@ public class TextureStore : ResourceStore<TextureUpload>
...
+        /// </remarks>
+        /// <param name="store">The store to add.</param>
+        public virtual void AddLookup(IResourceStore<TextureUpload> store) => uploadStore.AddStore(store);
```
peppy, line 66: I don't like the `Lookup` naming here. This is more like a transformer flow, right?

frenzibyte: Yeah unsure about the namings I used, thinking about this for a bit it might make much more sense using `AddStore` here and `AddNested` for the other one.

I went with `AddLookup` only to have it make more sense to the end consumer, since this sort of store is only used to look up texture uploads to load textures for, not look up full `Texture`s.

Interested in more opinions.

osu.Framework/Graphics/Textures/TextureStore.cs
frenzibyte: I personally never understood the point of a `TextureLoaderStore` until I went to the framework source code and read through it, so I believe having at least one xmldoc explaining how to transform image file stores into `TextureUpload` stores would give a better idea to the consumer on how that side works. The xmldoc you proposed sounds better and still hints at it.

osu.Framework/Graphics/Textures/TextureStore.cs
bdach: Not sure what the "transformer flow" mention is supposed to mean in this context, but I'd name this something like `AddTextureSource()` or something.

bdach:
Would be interested to see the corresponding osu!-side changes for this. I doubt there are none.

frenzibyte:
It's only a two lines at the time I checked, will PR it for visibility.

osu.Framework/Graphics/Textures/TextureStore.cs
frenzibyte: That sounds good as well, would want to use `AddSource` to keep it short in usages if that's agreeable.

frenzibyte: Now that I look at it, `AddSource` would mix rather badly with `AddStore`, I'm fine with going for `AddTextureSource` unless one has a second opinion.

osu.Framework/Graphics/Textures/ITextureStore.cs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+#nullable enable
```
bdach, line 4:
```suggestion
```

no longer required after #5241

osu.Framework/Graphics/Textures/TextureStore.cs
peppy: Let's go with `AddTextureSource` for now.

peppy:
Thoughts on also renaming `AddStore` to `AddFallbackStore` while we're here? As it stands I have a lot of trouble understanding that one, especially with the presence of this new method.

frenzibyte:
That wouldn't work well with cases like texture stores with custom filtering mode added to main `TextureStore`, i.e.:
```csharp
var nearestTextureStore = new TextureStore(nearestTextures, filteringMode: All.Nearest);
Textures.AddFallbackStore(nearestTextureStore);
```

Would you be fine with calling it `AddNestedStore` at least?

peppy:
I think something larger is wrong, because it's clear I'm not understanding the purpose of the method.

Will revisit this when I have the time to go over the full flow. If I can't make sense of it there's no way a consumer will.

smoogipoo:
The problem at large is the distinction between a `TextureLoaderStore` and a `TextureStore`. Would be nice if we didn't have to do the `TextureLoaderStore` dance...

peppy:
I've added some further xmldoc to help me understand this. I think I'm on board with these changes.

I have one remaining question though:

> Any existing usages of `AddStore` for adding lookup sources must be changed to use `AddLookup` instead.

It looks like even with the changes on this branch, previous code using `AddStore` will still compile and run. Will it fail in some unknown way, or still work fine, just be sub-optimal?

frenzibyte:
> > Any existing usages of `AddStore` for adding lookup sources must be changed to use `AddLookup` instead.
> 
> It looks like even with the changes on this branch, previous code using `AddStore` will still compile and run. Will it fail in some unknown way, or still work fine, just be sub-optimal?

Previously `AddStore` would accept both `TextureStore`s and `TextureLoaderStore`s. But now, if a consumer uses `AddStore` for `TextureLoaderStore` (the likely case), they'll have to update to `AddLookup`.

peppy:
So in the case of this:

https://github.com/ppy/osu/blob/495d747da55eab3cbfcc8331ffe5f6e73c6e3618/osu.Game/OsuGameBase.cs#L243

There is no real change in behaviour, apart from things being structured internally a bit weirdly?

bdach:
CI is failing here now, since the template projects actually are being built with these changes after #5250.

frenzibyte:
The template change didn't turn out as expected by this PR (using `AddStore` on a nested texture store with nearest filtering mode), since the entire resources assembly is added to `Resources`, which the base `Textures` store use as a lookup source. Therefore the template game has to still create a `TextureStore` and cache it instead of the base `Textures` one.

peppy (changes requested):

osu.Framework.Templates/templates/template-flappy/FlappyDon.Game/FlappyDonGameBase.cs
```diff
@@ -23,7 +23,7 @@ private void load()
 
             // To preserve the 8-bit aesthetic, disable texture filtering
             // so they won't become blurry when upscaled
-            textures = new TextureStore(Textures, filteringMode: All.Nearest);
+            textures = new TextureStore(Host.CreateTextureLoaderStore(new NamespacedResourceStore<byte[]>(Resources, @"Textures")), filteringMode: All.Nearest);
             dependencies.Cache(textures);
```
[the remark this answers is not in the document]
bdach: I believe this won't work as written because of the implicit ordering between `AddLookup()` and `AddStore()`. `Textures` will try itself first before proceeding to use nested stores. Because both stores use the same resources, then `Textures` wouldn't even use the `nearestNeighbourStore`, it would fetch the resource itself from `Resources` and then return it with the unexpected filtering mode.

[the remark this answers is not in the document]
peppy: Right. But that also means this whole PR doesn't fix the issue it's proposing to close?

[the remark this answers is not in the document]
frenzibyte: I can see two ways towards resolving this:
 1. Prioritise nested stores over main store (which would make more sense if we rename it to a "child"/"sub" texture store, or something of the sorts)
 2. Expose lookup resource store from `TextureStore` and use in the above case by doing:
    ```csharp
    var textures = new TextureStore(filteringMode: All.Nearest);
    textures.AddLookup(Textures.Sources);
    dependencies.Cache(textures);
    ```
    (that still means caching, but at least not duplicating the main texture store lookup sources)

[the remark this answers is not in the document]
bdach: I suppose so. It substitutes one unexpected behaviour (`AddStore()` working counterintuitively, because it only helps the texture store *look up* resources from the `TextureStore` provided, rather than allow it to be used wholesale) for another (you can do the above, but you have to keep in mind the inner workings of how `AddStore()` and `AddLookup()` interact).

I don't really claim know what the answer is here because the behaviour is inherently complex when two layers of lookup are permitted.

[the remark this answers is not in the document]
peppy: Another proposal would be to presume that if a user wants to change the filtering mode, it would be something they want changed globally, and expose it in some way at a `Game` level via a `virtual` property or otherwise.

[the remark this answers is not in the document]
frenzibyte: Since PR'ing this, I've been under the thought that consumers would nest texture stores for cases where there is a specific store away from the main `Game.Resources` one, and customise the texture store to use different filtering mode / scale adjust / etc.

> A proposal would be to just add the ability to specify the filter mode when adding a lookup store and having it all managed internally.

That would be pretty much what `AddStore(new TextureStore(lookupStore, filteringMode: ...))` should achieve. Except that in the flappydon case, the lookup store is already added to the main `Game.Resources` store, making this approach not quite feasible unless we prioritise nested stores over main store as mentioned.

> Another proposal would be to presume that if a user wants to change the filtering mode, it would be something they want changed globally, and expose it in some way at a `Game` level via a `virtual` property or otherwise.

This sounds pretty good for the flappydon case, and probably something that consumers would find better than creating a nested texture store that wraps the main `Textures` one (i.e. what flappydon does currently). I'll go with this approach for now, and we can reiterate later I think.

osu.Framework/Game.cs
```diff
@@ -34,6 +35,12 @@ public abstract class Game : Container, IKeyBindingHandler<FrameworkAction>, IKe
 
         public TextureStore Textures { get; private set; }
 
+        /// <summary>
+        /// Whether linear filtering/smoothing should be enabled for all textures fetched from <see cref="Textures"/>.
+        /// If disabled, textures will be displayed in nearest neighbour mode.
+        /// </summary>
+        protected virtual bool LinearTextureFiltering => true;
```
peppy, line 42: I think we should make this `DefaultTextureFilteringMode` and allow the user to choose either `Linear` or `Nearest` (does a better job at explaining what the choice means.

frenzibyte: Initially had it that way but thought I could simplify it to a boolean. This would also mean defining a new `FilteringMode` enum since we use OpenGL's `All` which we don't want to rely on as part of the public API.

peppy:
I'm still a bit lost as to potential usages of the newly added flows, but assuming there is valid reason for them that I'm missing, I'm generally okay with this structure now so I'll shoot a tentative approval.

Would want to make sure both @smoogipoo and @bdach are on board with the changes as they currently stand.

peppy (approved):

smoogipoo (approved):
Lgtm

smoogipoo (changes requested):
Whoops I added a comment >_>

osu.Framework/Graphics/Textures/TextureStore.cs
```diff
@@ -115,6 +158,50 @@ private Texture loadRaw(TextureUpload upload, WrapMode wrapModeS = WrapMode.None
...
+                }
+            }
+
+            return stream;
+        }
+
+        public IEnumerable<string> GetAvailableResources()
+        {
+            lock (nestedStores)
+                return uploadStore.GetAvailableResources().Concat(nestedStores.SelectMany(s => s.GetAvailableResources()).ExcludeSystemFileNames());
```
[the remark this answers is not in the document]
frenzibyte: Right since it's lazy enumeration, making the lock unnecessary. Feel free to apply that.

peppy (approved):

PR #5241, opened 2022-06-13
Enable NRT by default

smoogipoo:
I'm sorry in advance. I want to see if iOS/Android compiles because I don't have local workflows.

```bash
#!/bin/bash

YELLOW='\033[0;33m'
NC='\033[0m'

for file in $( find "$@" -type f -name '*.cs' ); do
    if grep -q "#nullable disable" "$file"; then
        echo -e "${YELLOW}Manual changes required: $file${NC}"
        continue
    fi

    if grep -qE "^#nullable enable" "$file"; then
        sed -e "/#nullable enable/,+1d" -i "$file"
        continue
    fi

    gawk -i inplace 'NR==3 { print "\n#nullable disable" }1' "$file"
done
```
Applied the following manual changes:
```
$ ~/Desktop/test.sh \
  ~/Repos/osu-framework/osu.Framework \
  ~/Repos/osu-framework/osu.Framework.Tests* \
  ~/Repos/osu-framework/osu.Framework.iOS \
  ~/Repos/osu-framework/osu.Framework.Android \
  ~/Repos/osu-framework/osu.Framework.Benchmarks \
  ~/Repos/osu-framework/SampleGame*

Manual changes required: /home/smgi/Repos/osu-framework/osu.Framework/Bindables/BindableDictionary.cs
Manual changes required: /home/smgi/Repos/osu-framework/osu.Framework.Tests/Extensions/TestIsNullableTypeExtensions.cs
Manual changes required: /home/smgi/Repos/osu-framework/osu.Framework.Tests/Dependencies/ResolvedAttributeTest.cs
Manual changes required: /home/smgi/Repos/osu-framework/osu.Framework.Tests/Dependencies/DependencyContainerTest.cs
```


frenzibyte (approved):
Gave it a manual check given that it's either `- #nullable enable` or `+ #nullable disable` (done using rider's github integration), but so far everything looks correct to me.

iOS also compiles and runs properly testing on simulator, would imagine Android to be the same with the assumption that the concern focuses on Xamarin projects in general and whether they support NRT.

frenzibyte:
Code Quality CI appears to be failing, for what it's worth.

smoogipoo:
It looks like it's failing due to the cache (can repro locally). Which is a bit of a shame because it's definitely inspecting every file (given by the time taken for the run to complete).

Funny, because I did consider cache busting, but then saw the run length and force-pushed above to remove https://github.com/ppy/osu-framework/commit/d756859fa93308dda1bac1e09329496396a08811 which did some cache busting :smile:

I'll report this to JetBrains, but I think it's within scope for us to cache bust on all `.csproj`/`.props` file changes anyway. Even if it means a longer run length every time we update dependencies, it makes sense to reinspect everything in-case some dependency changed the nullability contract in their API.

peppy:
I'm find with adding `csproj`/`props`.

smoogipoo:
Can't seem to repro in an isolated case / running into a few weird InspectCode errors >_> Let's see how cache busting goes.

Issue #4994, opened 2022-01-18
Crashes on Android after app is resumed if running in multi-threaded mode


bdach:
As reported in https://github.com/ppy/osu/discussions/16512
Follow-up to #4200

Looks like the context preservation added with https://github.com/ppy/osuTK/pull/74 only works on some devices because android gonna android (which I half-expected because the android documentation [seemed to indicate that this may be the case](https://developer.android.com/reference/android/opengl/GLSurfaceView#setPreserveEGLContextOnPause(boolean))), so it looks like in the worst case scenario there is still a need for the game to be fully cycled to continue working after resume.

bdach:
So seeing that this looks to be pretty much the number 1 issue per user/event count on sentry I went back to this one today to try and see if I can fix, but I'm having two problems:

* First and foremost, despite having an idea of how to fix it, I can actually not reproduce the context loss on any of my test devices despite numerous attempts. Apparently the loss is most common when the device goes to sleep, but it's just... not happening here. I even tried `adb` commands to force sleep to no avail.
* To know when the GL context is disposed, I'll want to hook onto [`GLSurfaceView.Renderer.onSurfaceCreated()`](https://developer.android.com/reference/kotlin/android/opengl/GLSurfaceView.Renderer#onsurfacecreated), which means I'll need to make osuTK changes - but I won't be able to use them because osuTK is hard-stuck on android after changes to compile it for net6 (https://github.com/ppy/osuTK/pull/75). So the fix ends up hard-depending on @huoyaoyuan's android on net6 series.

bdach:
Upon further investigation reproduction requires multithreaded execution. May not be the context being lost after all, rather than it being accessed before the game is allowed to do so.

peppy:
That `MakeCurrent()` call definitely looks suspicious (and likely unnecessary).

Does changing that alone make crashing less likely?

bdach:
No, it only changes error 1 to error 2. It's still a 100% crash either way.

Going to poke some more but my hopes are low.

bdach:
Had that thought too, then saw these two pages:

https://wiki.libsdl.org/Android
https://github.com/libsdl-org/SDL/blob/main/docs/README-android.md

and slowly backed away. A lot of eldritch incantations and corralling jni and what have you. I have zero clue about any of it.

bdach:
[...]

PR #5242, opened 2022-06-13
Suppress `SwapBuffers()` failures on android

bdach:
Workaround for the other crash described in #4994.

For reasons that are currently unclear, sometimes `SwapBuffers()` calls on android fail with `BAD_SURFACE` just before the game activity is suspended, and just after the game activity is resumed - but if given ample time to recover from the initial failures after resume, the game will continue functioning as usual.

As a temporary solution, suppress `SwapBuffers()` failures. The ideal solution would be to hook onto one of the many callbacks of
`GLSurfaceView`, but in testing it does not appear that the obvious ones fit the bill.

### DIFF
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -242,18 +242,25 @@ namespace osu.Framework.Android
             return new AndroidInputConnection(this, true);
         }
 
+        private static readonly Logger logger = Logger.GetLogger(LoggingTarget.Runtime);
+
         public override void SwapBuffers()
         {
             try
             {
                 base.SwapBuffers();
             }
-            catch
+            catch (GraphicsContextException ex)
             {
-                // sometimes buffers will spontaneously fail to swap just before the activity is suspended to background
-                // or just after it has been resumed, but will continue operating correctly after that transitionary period.
+                // sometimes buffers will spontaneously fail to swap with BAD_SURFACE
+                // just before the activity is suspended to background or just after it has been resumed,
+                // but will continue operating correctly after that transitionary period.
                 // despite some testing it is unclear which view callback can be used to tell whether it is safe to swap buffers,
-                // so for now just catch and suppress errors.
+                // so for now just catch and suppress these errors.
+                if (ex.Message.Contains("BAD_SURFACE", StringComparison.Ordinal))
+                    logger.Add($"BAD_SURFACE failure in {nameof(SwapBuffers)} suppressed", LogLevel.Verbose, ex);
+                else
+                    throw;
             }
         }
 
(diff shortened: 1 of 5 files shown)


osu.Framework.Android/AndroidGameView.cs
```diff
@@ -242,6 +242,21 @@ public override IInputConnection OnCreateInputConnection(EditorInfo outAttrs)
...
+            try
+            {
+                base.SwapBuffers();
+            }
+            catch
+            {
+                // sometimes buffers will spontaneously fail to swap just before the activity is suspended to background
+                // or just after it has been resumed, but will continue operating correctly after that transitionary period.
+                // despite some testing it is unclear which view callback can be used to tell whether it is safe to swap buffers,
+                // so for now just catch and suppress errors.
```
frenzibyte, line 256: While suppressing, I think it may still be better logging the thrown exception in case we come across another android-related issue where a failing `SwapBuffers` call is relevant (especially if it isn't a `BAD_SURFACE` error)

peppy: Can probably type this to `GraphicsContextException` and check `message` to only handle the specific ones we care about.

Not sure about logging, depends how many times it fires. If it was logged it shouldn't be `Logger.Error`, but a message mentioning that x exception was suppressed.

frenzibyte: Yeah `Logger.Log` at verbose level is what I had in mind, for the main purpose of showing on GitHub/sentry reports without irritating the user.

peppy:
I've deployed an updated osuTK and committed the update to this PR since it seems like the best place to make it happen. Note that it was built from a temporary non-net6 branch (https://github.com/ppy/osutk/tree/master-before-net6).

osu.Framework.Android/AndroidGameView.cs
bdach: I've considered scoping the suppression more but decided to start simple and see where it goes at review time. I'll make the suggested changes.

bdach: I've made the suggested changes. Please see if they match your expectations.

osu.Framework.Android/AndroidGameView.cs
```diff
@@ -242,18 +242,25 @@ public override IInputConnection OnCreateInputConnection(EditorInfo outAttrs)
             return new AndroidInputConnection(this, true);
         }
 
+        private static readonly Logger logger = Logger.GetLogger(LoggingTarget.Runtime);
```
peppy, line 245: Is this just to avoid the retrieval overhead?

peppy: Honestly I think dropping the stack is going to be more manageable. It's not conveying anything we don't already know.

peppy (approved):

PR #5245, opened 2022-06-14
Clean up `FrameStatisticsDisplay` and add exception to track down `NaN`

peppy:
See https://sentry.ppy.sh/share/issue/8af58e085ca14af4a7bfd15beeef62b1/.

This is the only variable I can see which can cause a `NaN`. Throwing will mean the game still may crash (although potentially not, since it will only fire once rather than every frame), but also mean we can track down the faulty flow.

I've checked each call to `Add` and can't find anything immediately. Welcome others to double-check.


peppy:
I considered changing the parameter to `ulong` but it brings quite a bit of casting with it, so not sure.

smoogipoo (approved):

PR #5247, opened 2022-06-15
Indicate some `Storage` methods as must-use

smoogipoo:
`[Pure]` seems to be as useful as JetBrains' own `[MustUseReturnValue]` here, so I'm going with it.


smoogipoo:
Hang on a sec. It seems like `[Pure]` may be able to be used.

peppy (approved):

PR #5250, opened 2022-06-16
Add templates to compile + inspectcode, deploy templates with framework

smoogipoo:
- Runs template tests on CI (similar to https://github.com/ppy/osu/pull/18723).
- Adds templates to `.slnfs`, allowing inspectcode + builds to run on them. Locally and on CI.
- Deploys templates along with main framework (similar to https://github.com/ppy/osu/pull/11523).
  - Uses the appveyor project repository as per the above PR to resolve nuget references.
- `osu.Framework.NativeLibs` remains as a manual process for the time being.

@peppy Note that I've removed `appveyor_xamarin_deploy.yml`. The main `appveyor_deploy.yml` file will deploy everything - desktop, Xamarin, and templates, all in one go.

Deploy run: https://ci.appveyor.com/project/smoogipoo/osu-framework/builds/43879324 (minus nuget deploy failures for obvious reasons)


build/build.cake
```diff
@@ -1,4 +1,6 @@
 using System.Threading;
+using System.Text.RegularExpressions;
+#addin "nuget:?package=Cake.FileHelpers&version=3.2.1"
```
smoogipoo, line 3: Intentionally using the non-latest version of this since we're stuck on cake 0.35.0 for now and the latest versions support 2.0+.

This will will go away once we migrate to GH actions anyway.

osu-framework.Android.slnf
```diff
@@ -7,7 +7,10 @@
       "osu.Framework.Android\\osu.Framework.Android.csproj",
       "osu.Framework.NativeLibs\\osu.Framework.NativeLibs.csproj",
       "osu.Framework.Tests.Android\\osu.Framework.Tests.Android.csproj",
-      "osu.Framework\\osu.Framework.csproj"
+      "osu.Framework\\osu.Framework.csproj",
+
+      "osu.Framework.Templates\\templates\\template-empty\\TemplateGame.Game\\TemplateGame.Game.csproj",
+      "osu.Framework.Templates\\templates\\template-flappy\\FlappyDon.Game\\FlappyDon.Game.csproj"
```
bdach, line 13: Just going to note that adding these to the android slnf doesn't really do much because there's no android launch project. Maybe I'll look into adding one some day, should be simple enough.

.github/workflows/ci.yml
```diff
@@ -79,10 +79,10 @@ jobs:
           dotnet-version: "6.0.x"
 
       - name: Compile
-        run: dotnet build -c Debug -warnaserror build/Desktop.proj
+        run: dotnet build -c Debug -warnaserror osu-framework.Desktop.slnf
 
       - name: Test
-        run: dotnet test $pwd/*.Tests/bin/Debug/*/*.Tests.dll --settings $pwd/build/vstestconfig.runsettings --logger "trx;LogFileName=TestResults-${{matrix.os.prettyname}}-${{matrix.threadingMode}}.trx"
```
bdach, line 85: Interestingly enough, it looks like the template tests are *not* being ran after all, judging from CI output: https://github.com/ppy/osu-framework/runs/6916451817?check_suite_focus=true#step:5:43

Can repro locally via rider, the test runs show up as "inconclusive". If I add `[TestFixture]` to `TemplateGameTestScene`, for instance, then all tests start running correctly. Pretty weird.

bdach (commented):
Looks pretty good for the most part on the code side. Not sure about the deploy specifics but also looks sane.

.github/workflows/ci.yml
smoogipoo: Seems to be because there's no `[TestFixture]`s in files... I remember some quirks about `[TestFixture]`, but it seems it picks up everything as soon as it's added to _one_ file - probably some optimisation.

bdach (approved):

peppy (approved):

bdach:
I don't think that's a huge issue. The platform-specific parts of template projects should be very thin and basically never break.

PR #5251, opened 2022-06-16
Show hidden items in DirectorySelectors

ggliv:
Discussed in ppy/osu#15759, resolves ppy/osu#18573.

This change includes hidden items in `DirectorySelector` dialogues and provides a way to reduce their opacity to distinguish them from non-hidden items.

I elected to make opacity reduction a defaulted constructor argument for `DirectorySelectorDirectory` to avoid making any breaking changes, but this could be changed to a property to force implementations to state their opacity preference explicitly.

I've only tested this on Linux with dotfiles, would appreciate someone checking it on macOS and Windows (though I can't immediately think of anything that would be different).

### DIFF
--- a/osu.Framework/Graphics/UserInterface/BasicDirectorySelectorBreadcrumbDisplay.cs
+++ b/osu.Framework/Graphics/UserInterface/BasicDirectorySelectorBreadcrumbDisplay.cs
@@ -45,7 +45,11 @@ namespace osu.Framework.Graphics.UserInterface
             protected override IconUsage? Icon => Directory.Name.Contains(Path.DirectorySeparatorChar) ? base.Icon : null;
 
             public BreadcrumbDisplayDirectory(DirectoryInfo directory, string displayName = null)
-                : base(directory, displayName, false)
+                : base(directory, displayName)
+            {
+            }
+
+            protected override void ApplyHiddenState()
             {
             }
 
--- a/osu.Framework/Graphics/UserInterface/BasicDirectorySelectorBreadcrumbDisplay.cs
+++ b/osu.Framework/Graphics/UserInterface/BasicDirectorySelectorBreadcrumbDisplay.cs
@@ -45,7 +45,7 @@ namespace osu.Framework.Graphics.UserInterface
             protected override IconUsage? Icon => Directory.Name.Contains(Path.DirectorySeparatorChar) ? base.Icon : null;
 
             public BreadcrumbDisplayDirectory(DirectoryInfo directory, string displayName = null)
-                : base(directory, displayName)
+                : base(directory, displayName, false)
             {
             }
 
--- a/osu.Framework/Graphics/UserInterface/BasicDirectorySelectorBreadcrumbDisplay.cs
+++ b/osu.Framework/Graphics/UserInterface/BasicDirectorySelectorBreadcrumbDisplay.cs
@@ -49,7 +49,8 @@ namespace osu.Framework.Graphics.UserInterface
             {
             }
 
-            protected override void ApplyHiddenState()
+            // this method is suppressed to ensure the breadcrumbs of hidden directories are presented the same way as non-hidden directories
+            protected sealed override void ApplyHiddenState()
             {
             }
 
--- a/osu.Framework/Graphics/UserInterface/BasicDirectorySelectorParentDirectory.cs
+++ b/osu.Framework/Graphics/UserInterface/BasicDirectorySelectorParentDirectory.cs
@@ -13,7 +13,7 @@ namespace osu.Framework.Graphics.UserInterface
         protected override IconUsage? Icon => FontAwesome.Solid.Folder;
 
         public BasicDirectorySelectorParentDirectory(DirectoryInfo directory)
-            : base(directory, "..")
+            : base(directory, "..", false)
         {
         }
     }
--- a/osu.Framework/Graphics/UserInterface/BasicDirectorySelectorParentDirectory.cs
+++ b/osu.Framework/Graphics/UserInterface/BasicDirectorySelectorParentDirectory.cs
@@ -13,7 +13,11 @@ namespace osu.Framework.Graphics.UserInterface
         protected override IconUsage? Icon => FontAwesome.Solid.Folder;
 
         public BasicDirectorySelectorParentDirectory(DirectoryInfo directory)
-            : base(directory, "..", false)
+            : base(directory, "..")
+        {
+        }
+
+        protected override void ApplyHiddenState()
         {
         }
     }
--- a/osu.Framework/Graphics/UserInterface/BasicDirectorySelectorParentDirectory.cs
+++ b/osu.Framework/Graphics/UserInterface/BasicDirectorySelectorParentDirectory.cs
@@ -17,7 +17,8 @@ namespace osu.Framework.Graphics.UserInterface
         {
         }
 
-        protected override void ApplyHiddenState()
+        // this method is suppressed to ensure that parent directories that are also hidden directories are presented the same way as non-hidden parent directories
+        protected sealed override void ApplyHiddenState()
         {
         }
     }
--- a/osu.Framework/Graphics/UserInterface/DirectorySelectorDirectory.cs
+++ b/osu.Framework/Graphics/UserInterface/DirectorySelectorDirectory.cs
@@ -32,7 +32,7 @@ namespace osu.Framework.Graphics.UserInterface
             }
             catch (UnauthorizedAccessException)
             {
-                // checking attributes on access-controled directories will throw an error so we handle it here to prevent a crash
+                // checking attributes on access-controlled directories will throw an error so we handle it here to prevent a crash
             }
         }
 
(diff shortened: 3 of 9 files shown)


osu.Framework/Graphics/UserInterface/DirectorySelectorItem.cs
```diff
@@ -24,6 +24,11 @@ public abstract class DirectorySelectorItem : CompositeDrawable
         /// </summary>
         protected abstract string FallbackName { get; }
 
+        /// <summary>
+        /// Whether this <see cref="DirectorySelectorItem"/> should have reduced opacity on its <see cref="Drawable"/>.
+        /// </summary>
+        protected abstract bool? ReduceOpacity { get; }
```
bdach, line 30: I don't like this property. It's overly generic.

I'd honestly inline the "reduced opacity" logic in `DirectoryListingFile` and `DirectorySelectorDirectory`. There's potentially a case for making an `ApplyHiddenState()` method in this class (which defaults to applying 0.5 alpha) so that the display of hidden items can be customised, but I'm not sure it's necessary.

osu.Framework/Graphics/UserInterface/DirectorySelectorDirectory.cs
```diff
@@ -7,21 +7,25 @@
...
         protected readonly DirectoryInfo Directory;
+        protected override bool? ReduceOpacity => isHiddenDirectoryWithReducedOpacity;
+        private readonly bool isHiddenDirectoryWithReducedOpacity;
         protected override string FallbackName => Directory.Name;
 
         [Resolved]
         private Bindable<DirectoryInfo> currentDirectory { get; set; }
 
-        protected DirectorySelectorDirectory(DirectoryInfo directory, string displayName = null)
+        protected DirectorySelectorDirectory(DirectoryInfo directory, string displayName = null, bool reduceHiddenDirectoryOpacity = false)
```
bdach, line 24: The `reduceHiddenDirectoryOpacity` flag shouldn't exist IMO.

If there is a case for disabling the opacity thing, I'd suggest doing it via an `ApplyHiddenState()` method on `DirectorySelectorItem` that consumers can override to be a no-op.

osu.Framework/Graphics/UserInterface/DirectorySelectorItem.cs
ggliv: `ReduceOpacity` has been removed and the logic has been moved as requested.

osu.Framework/Graphics/UserInterface/DirectorySelectorDirectory.cs
ggliv: You'd probably want to disable reduced opacity for parent and breadcrumb directories. I've gone ahead and added `ApplyHiddenState()`.

bdach:
Visual test for the component hard-crashes for me:

```
System.UnauthorizedAccessException: Access to the path '/sys/kernel/debug/tracing' is denied.
 ---> System.IO.IOException: Permission denied
   --- End of inner exception stack trace ---
   at System.IO.FileStatus.ThrowOnCacheInitializationError(ReadOnlySpan`1 path)
   at System.IO.FileStatus.GetAttributes(ReadOnlySpan`1 path, ReadOnlySpan`1 fileName)
   at System.IO.FileSystemInfo.get_Attributes()
   at osu.Framework.Graphics.UserInterface.DirectorySelectorDirectory..ctor(DirectoryInfo directory, String displayName) in /home/dachb/Dokumenty/opensource/osu-framework/osu.Framework/Graphics/UserInterface/DirectorySelectorDirectory.cs:line 26
```

It's a bit weird because `/sys/kernel/debug/tracing` is apparently a "drive" to .NET, but this shouldn't bring down the game regardless.

osu.Framework/Graphics/UserInterface/FileSelector.cs
```diff
@@ -80,6 +79,12 @@ protected override bool OnClick(ClickEvent e)
             }
 
             protected override string FallbackName => File.Name;
+
+            protected DirectoryListingFile()
+            {
+                if (currentFile?.Value.Attributes.HasFlagFast(FileAttributes.Hidden) ?? false)
```
bdach, line 85: Please use `== true` rather than `?? false`. We've agreed to only use the former in the project because `?? false` patterns are quite hard to read.

osu.Framework/Graphics/UserInterface/BasicDirectorySelectorBreadcrumbDisplay.cs
```diff
@@ -49,6 +49,10 @@ public BreadcrumbDisplayDirectory(DirectoryInfo directory, string displayName =
             {
             }
 
+            protected override void ApplyHiddenState()
+            {
+            }
```
bdach, line 54: These suppressing overrides should have an inline comment explaining the reason for suppression, and likely be `sealed` as well.

osu.Framework/Graphics/UserInterface/FileSelector.cs
ggliv: Changed.

osu.Framework/Graphics/UserInterface/BasicDirectorySelectorParentDirectory.cs
```diff
@@ -16,5 +16,10 @@ public BasicDirectorySelectorParentDirectory(DirectoryInfo directory)
             : base(directory, "..")
         {
         }
+
+        // this method is suppressed to ensure that parent directories that are also hidden directories are presented the same way as non-hidden parent directories
```
bastianpedersen, line 20:
```suggestion
        // This method is suppressed to ensure that parent directories that are also hidden directories are presented the same way as non-hidden parent directories.
```

osu.Framework/Graphics/UserInterface/BasicDirectorySelectorBreadcrumbDisplay.cs
```diff
@@ -49,6 +49,11 @@ public BreadcrumbDisplayDirectory(DirectoryInfo directory, string displayName =
             {
             }
 
+            // this method is suppressed to ensure the breadcrumbs of hidden directories are presented the same way as non-hidden directories
```
bastianpedersen, line 52:
```suggestion
            // This method is suppressed to ensure the breadcrumbs of hidden directories are presented the same way as non-hidden directories.
```

bastianpedersen (commented):
Nitpick.

osu.Framework/Graphics/UserInterface/BasicDirectorySelectorBreadcrumbDisplay.cs
bdach: Fine to ignore this review comment, it's too much of a nitpick.

osu.Framework/Graphics/UserInterface/DirectorySelectorDirectory.cs
```diff
@@ -22,6 +23,13 @@ protected DirectorySelectorDirectory(DirectoryInfo directory, string displayName
             : base(displayName)
         {
             Directory = directory;
+
+            try
+            {
+                if (directory?.Attributes.HasFlagFast(FileAttributes.Hidden) == true)
+                    ApplyHiddenState();
+            }
+            catch { }
```
smoogipoo, line 32: The try-catches seem weird to me, is there a reason for these existing? (Here and in `FileSelector`).

bdach: Reasoning is https://github.com/ppy/osu-framework/pull/5251#issuecomment-1158044995

tl;dr: even checking if the file is hidden can fail if you don't have access rights to it.

smoogipoo: Right. Definitely needs to be commented then, maybe even catch `UnauthorizedAccessException`s explicitly too.

bdach:
To save everyone another grueling round of review I've pushed [7646eb6](https://github.com/ppy/osu-framework/pull/5251/commits/7646eb6febd79cc5e0041b070309a8fb7d4c131c) which fixes the implementation of the hidden state of files, as the previous implementation was grossly incorrect - not only was the old implementation placed in a constructor that nothing used, it was also checking the wrong file.

That said, I think this is fine now, but I'm not merging without a second approval.

bdach (dismissed):

frenzibyte (dismissed):
This seems good as-is, have left a comment but can be left for a follow-up if too much work, although I can't imagine it being so.

frenzibyte:
I've applied further changes mainly to tweak the display and placement of the button, now looks:

https://user-images.githubusercontent.com/22781491/174421289-b5e55e52-c826-4901-bc96-4eacedeb1663.mp4

Unsure if the size of the button makes sense, but looks fine for a basic implementation.

frenzibyte (approved):

peppy:
Seems okay. The basic toggle button should really have its state visible but I don't think anyone in their right mind would use that basic design so maybe it's not a huge deal for now.

peppy (approved):

ggliv:
https://github.com/ppy/osu-framework/commit/c1ea24e654c0e89e17f5f8216962a454d7ed1b5d made hiding hidden items the default behavior, this issue should probably be reopened until a toggle is implemented osu-side.

PR #5254, opened 2022-06-19
Implement search text box within the draw visualizer

naoei:
Addresses [#5125](https://github.com/ppy/osu-framework/discussions/5125)

This seems like a straightforward implementation, however, I'm not entirely sure about the whole `WrapScrollContent()` approach here but I can't seem to figure out another possible solution.

https://user-images.githubusercontent.com/35349837/174502141-2fc988ec-d94d-461e-8cd1-938085a951a6.mp4

### DIFF
--- a/osu.Framework/Graphics/Visualisation/ToolWindow.cs
+++ b/osu.Framework/Graphics/Visualisation/ToolWindow.cs
@@ -28,7 +28,7 @@ namespace osu.Framework.Graphics.Visualisation
 
         protected readonly SearchContainer SearchContainer;
 
-        protected ToolWindow(string title, string keyHelpText, bool showSearchTextBox = false)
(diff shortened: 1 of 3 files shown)


peppy:
I think rather than the `Wrap` method, you should just implement the full search functionality (including text box) in the base tool window implementation. Going forward, I'd hope most of the tool windows could be searchable, so I think this makes sense.

It could be temporarily hidden (by default) on tool windows which don't support it by hiding the search textbox via a parameter in the ctor:

```csharp
        protected ToolWindow(string title, string keyHelpText, bool showSearchTextBox)
```

or something similar.

naoei:
Applied the requested changes. Looks a lot cleaner now.

peppy (approved):

peppy:
Seems fine for now, nice one.

PR #5257, opened 2022-06-23
Fix potential exception when calling `FinishTransforms` results in child changes

peppy:
As seen at https://sentry.ppy.sh/share/issue/8d469f0003bd446eb4bda7865b2e9bc9/.

Can probably figure out a repro case if test coverage is seen as beneficial, but it's obvious how this could happen (and the inline comment should guard against regression).

### DIFF
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -360,7 +360,10 @@ namespace osu.Framework.Graphics.Containers
             get
             {
                 if (InternalChildren.Count != 1)
-                    throw new InvalidOperationException($"Cannot call {nameof(InternalChild)} unless there's exactly one {nameof(Drawable)} in {nameof(InternalChildren)} (currently {InternalChildren.Count})!");
+                {
+                    throw new InvalidOperationException(
+                        $"Cannot call {nameof(InternalChild)} unless there's exactly one {nameof(Drawable)} in {nameof(InternalChildren)} (currently {InternalChildren.Count})!");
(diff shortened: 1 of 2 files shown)


smoogipoo:
Am I right in that the "obvious" case for how this can happen is `x.Transform().OnComplete(t => t.Expire());`? Or am I missing a simpler case?

peppy:
Yep, something like that. Any `OnComplete` or `OnAbort` or `Schedule` that causes a removal of drawables in some way.

smoogipoo:
I find this sort of usage of transforms extremely dodgy if that is the case, and am also confused as to why it wouldn't happen every time. I think I'll turn a blind eye but I don't agree with allowing this...

smoogipoo (approved):

peppy:
If you take a look at the linked example on sentry, the specific case is a high level component calling `FinishTransforms` on the full fail overlay, which in turn does some clean-up:

https://github.com/ppy/osu/blob/f8830c6850128456266c82de83273204f8b74ac0/osu.Game/Screens/Play/FailAnimation.cs#L121-L122

smoogipoo:
Yeah, but you have to understand how this is dodgy right? We're just blanket skipping over `Drawables` so now you potentially have to call `FinishTransforms()` multiple times to ensure that transforms are actually finished.

peppy:
I see what you mean. Probably need a safer enumerator helper method for these cases which reverses or re-runs or something.

Probably worth opening an issue tracking, since as you say it'll probably cause weirdness that we will never know how to track.

smoogipoo (approved):

PR #5261, opened 2022-06-23
Disallow usage of `string.To{Upper,Lower}()` without explicit culture

bdach:
Due to it relying on ambient current culture. Vaguely related to https://github.com/ppy/osu/issues/18729. I intend to apply this over there too later.

No major breakage here (though I did see a `performance-ınput.log`, that's dotless i, in my testing).

### DIFF
--- a/osu.Framework.iOS/Input/IOSTextFieldKeyboardHandler.cs
+++ b/osu.Framework.iOS/Input/IOSTextFieldKeyboardHandler.cs
@@ -4,6 +4,7 @@
 #nullable disable
 
 using System;
+using System.Globalization;
 using System.Linq;
 using Foundation;
 using osu.Framework.Input.Handlers;
--- a/osu.Framework.iOS/Input/IOSTextFieldKeyboardHandler.cs
+++ b/osu.Framework.iOS/Input/IOSTextFieldKeyboardHandler.cs
@@ -204,7 +204,7 @@ namespace osu.Framework.iOS.Input
                 default:
                     if (char.IsLetter(c))
                     {
-                        string keyName = c.ToString().ToUpper();
+                        string keyName = c.ToString().ToUpper(CultureInfo.CurrentCulture);
                         if (Enum.TryParse(keyName, out Key result))
                             return result;
                     }
--- a/osu.Framework/Platform/SDL2/SDL2ReadableKeyCombinationProvider.cs
+++ b/osu.Framework/Platform/SDL2/SDL2ReadableKeyCombinationProvider.cs
@@ -39,6 +39,9 @@ namespace osu.Framework.Platform.SDL2
 
             // SDL_GetKeyName() returned a unicode character that would be produced if that key was pressed.
             // consumers expect an uppercase letter.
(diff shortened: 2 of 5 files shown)


osu.Framework/Platform/SDL2/SDL2ReadableKeyCombinationProvider.cs
```diff
@@ -38,7 +39,10 @@ protected override string GetReadableKey(InputKey key)
 
             // SDL_GetKeyName() returned a unicode character that would be produced if that key was pressed.
             // consumers expect an uppercase letter.
-            return name.ToUpper();
+            // `.ToUpper()` with current culture may be slightly inaccurate if the framework locale
+            // is different from the locale of the keyboard layout being used,
+            // but we have no means to detect this anyway, as SDL doesn't provide the name of the layout.
+            return name.ToUpper(CultureInfo.CurrentCulture);
```
bdach, line 45: As stated above, this is culture-variant as the string is actually somewhat user-facing, but not 100% correct.

I tested the Turkish layout and it's wrong:

image

One of these capital `I` letters is supposed to have a dot over it. But this is not actually our fault this time, as [SDL also does this wrong](https://github.com/libsdl-org/SDL/blob/e0f3da497f4ccf7ce8fedcd0a62ed07f8f443462/src/events/SDL_keyboard.c#L1063-L1069). Time and time again, it is proven that internalisation is hell.

osu.Framework.iOS/Input/IOSTextFieldKeyboardHandler.cs
```diff
@@ -204,7 +204,7 @@ private void handleShouldReturn()
                 default:
                     if (char.IsLetter(c))
                     {
-                        string keyName = c.ToString().ToUpper();
+                        string keyName = c.ToString().ToUpper(CultureInfo.CurrentCulture);
```
bdach, line 207: I missed this in the initial changes because iOS, and cannot test because iOS. In theory it should behave the same as it already did. It looks potentially slightly wrong to me, but if it is, then it was as wrong before.

frenzibyte: Looking at the `Enum.TryParse(out Key)` below where this string is used, it looks completely wrong indeed. Will have to double-check but it should probably use invariant.

bdach: Well if an accented character or some other diacritic that isn't ASCII arrives here it's not gonna parse to the enum either way, regardless of casing. Probably safe to handle as follow-up if it's an actual problem.

peppy (approved):

osu.Framework/Platform/SDL2/SDL2ReadableKeyCombinationProvider.cs
bdach: Opened an upstream issue about this (https://github.com/libsdl-org/SDL/issues/5836).

peppy (approved):

smoogipoo (approved):

PR #5263, opened 2022-06-23
Early-return on `GetDescription`/`GetLocalisableDescription` if already string type

frenzibyte:
This is implicitly what `GetDescription` does, but `GetLocalisableDescription` fails with returning a `string`-backed `LocalisableString` instead of the actual localised string data (since it's falling back to the `GetDescription` branch).

Instead, now `GetDescription` and `GetLocalisableDescription` explicitly check if the type is already string and return the value instead of going through reflection.

This came up in https://github.com/ppy/osu/pull/18803#discussion_r903224609 where the method is trying to do too much just to display a localisable string with upper case. We might end up not relying on `LocalisableString`-typed tab items, but leaving this PR out just to match behaviour with `GetDescription` at least.


peppy (approved):

smoogipoo (approved):
I'm not super sure on overloading this extension method so much, but it seems okay with the appropriate doc.

PR #5265, opened 2022-06-24
Update c# language version to 9

peppy:
Mainly to fix the regression in `IHasCustomTooltip` disallowing struct usage (d90d27acb).

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 ﻿<!-- Contains required properties for osu!framework projects. -->
 <Project>
   <PropertyGroup Label="C#">
-    <LangVersion>8.0</LangVersion>
+    <LangVersion>10.0</LangVersion>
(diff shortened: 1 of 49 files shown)


peppy:
Blocked from 10 by iOS, 9 works and covers the requirements we have for now.

frenzibyte (approved):
Looks all good for me, have confirmed iOS working with both simulator and deploy to physical device.

smoogipoo:
There's a lot of suggestions in the CI output (we need to upgrade the NVika package so these show up on GitHub). Probably worth adding `DotSettings` rules for them (it's the `new XType(...)` => `new(...)` stuff which I think we all don't like?).

peppy:
I'll fix those.

peppy:
NVika looks happy now. I'll apply similar changes to osu-side once this is merged.

bdach (approved):

PR #5267, opened 2022-06-24
Fix `TripleBuffer` sleeping too long at high frame rates

peppy:
Previous version was using `Thread.Sleep(1)` when waiting on data, but with two threads running at roughly the same frame rate this would result in highly inefficient synchronisation. Using `ManualResetEventSlim` allows spinning for quick resolution cases.

Hopefully the changes should be self-explanatory and the code should be pretty straightforwards. Went through a few iterations to get to this point, have left some commits along the way to show different approaches leading up to the final one.

I've only tested this on macOS so far, which shows an increase of draw FPS from 500 to 900 with "Basically Unlimited" setting.

Addresses concerns raised at https://github.com/ppy/osu/discussions/18814.

### DIFF
--- a/osu.Framework.Tests/Graphics/TripleBufferTest.cs
+++ b/osu.Framework.Tests/Graphics/TripleBufferTest.cs
@@ -1,7 +1,6 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-#nullable disable
 using System.Threading;
 using System.Threading.Tasks;
 using NUnit.Framework;
--- a/osu.Framework.Tests/Graphics/TripleBufferTest.cs
+++ b/osu.Framework.Tests/Graphics/TripleBufferTest.cs
@@ -65,21 +65,21 @@ namespace osu.Framework.Tests.Graphics
                 var obj = new TestObject(i);
                 ManualResetEventSlim resetEventSlim = new ManualResetEventSlim();
 
-                var readTask = Task.Run(() =>
+                var readTask = Task.Factory.StartNew(() =>
                 {
                     resetEventSlim.Set();
                     using (var buffer = tripleBuffer.GetForRead())
                         Assert.That(buffer?.Object, Is.EqualTo(obj));
-                });
+                }, TaskCreationOptions.LongRunning);
 
-                Task.Run(() =>
+                Task.Factory.StartNew(() =>
                 {
                     resetEventSlim.Wait(1000);
                     Thread.Sleep(10);
 
                     using (var buffer = tripleBuffer.GetForWrite())
                         buffer.Object = obj;
-                });
+                }, TaskCreationOptions.LongRunning);
 
                 readTask.WaitSafely();
             }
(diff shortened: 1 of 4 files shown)


peppy:
Can confirm this recovers most all the performance on windows too (950-1000 fps on draw with 1000fps on update).

bdach:
One of the tests added is failing (except on linux, strangely).

peppy:
Can't immediately see how, nor can I repro locally with 1000000000 runs.

peppy:
For more commentary on this, here's a view of threads before and after this change, running at "Basically Unlimited". Top is `master`, bottom is this PR, order of threads is `Input`,`Audio`,`Update`,`Draw`:

image

Zooming in, you can see that due to the update thread being roughly halfway out of sync (it eventually stabilises to this node), the draw thread is doing *two* sleep operations - one for the throttling of the clock, but a second when querying for the next `TripleBuffer`, adding at least 1ms extra time slept to the equation.

image

This inhibits the draw frame rate.

The change in this branch will cause marginally higher CPU usage (measured at around double previous numbers), but that overhead lessens at lower frame limiters and should be a non-issue considering that we should generally have dedicated cores for each game thread. Also keep in mind it's running double the draw frames (which are one of the more cpu intensive pieces of pie currently).

Before:

image

After:

image



---

Of note, and as I've mentioned in the past, at 1000hz the precision inaccuracy of `Thread.Sleep` becomes a real issue when aiming for stable frame delivery. This is one reason stable runs at 960hz max (to give some lenience as we know for a fact we can't `Thread.Sleep` <1ms). May be something to consider, but before doing so I already have some testing in progress of replacing these `Thread.Sleep` calls with the `SpinWait` mechanism.

peppy:
I did some further experimentation with `GameThread` implementation (https://github.com/peppy/osu-framework/commits/spin-more-sleep-less) but the consensus I arrived at is that how we are already doing things is the best middle ground between CPU usage and accuracy.

Also as it turns out, the way we are doing error accumulation and sleeping means that our threads stay quite well synchronised. In my branch implementations I attempted to lock synchronisation (as a baseline, since it is still per-thread and not global) but it ends up being counter productive. A better approach in the future will involve cross-thread `ResetEvent`s in a similar way to this branch, allowing for `Update` to wait on a potentially incoming input (and more importantly, attempting to guarantee that input finishes just before update).

But this is all *very* diminishing gains and I'm mostly looking into it for fun, not because we need it.

osu.Framework.Tests/Graphics/TripleBufferTest.cs
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+#nullable disable
```
smoogipoo, line 4: Nit: Remove

smoogipoo (dismissed):
Have only done a visual pass for the time being since I'm not at my PC, but it looks okay.

smoogipoo (approved):
Let's see how this goes, it looks fine to me. I had to convince myself that the `.Reset()` before the `lock{}` is potentially better for timing, so hopefully I'm not wrong in that.

peppy:
> Let's see how this goes, it looks fine to me. I had to convince myself that the `.Reset()` before the `lock{}` is potentially better for timing, so hopefully I'm not wrong in that.

Yeah, in the *very* edge case it is marginally better. Leads to earlier synchronisation (by x frames?) of draw to update than otherwise, if a draw catchers a just-started update.

PR #5268, opened 2022-06-25
Fix overlay containers handling click and double-click events

frenzibyte:
All mouse events are limited to the `ButtonDownInputQueue`, which is determined by which drawable handles `MouseDownEvent`. Therefore `OverlayContainer`s don't need to handle all mouse events, but only `MouseDownEvent` to block positional input from propagating outside its hierarchy.

The previous logic was causing clicks and double-clicks to be handled as well, which in turn result in next clicks targeted at UI controls being blocked by the handled double click.

It was also handling drag start events, which marks this as a breaking change for any `OverlayContainer` implementation which was handling drag events without `OnDragStart` (there's none osu!-side).

- Closes https://github.com/ppy/osu/issues/18838

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneFocusedOverlayContainer.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneFocusedOverlayContainer.cs
@@ -138,13 +138,7 @@ namespace osu.Framework.Tests.Visual.UserInterface
                     return true;
                 }
 
-                return true;
-            }
-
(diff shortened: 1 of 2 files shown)


peppy (approved):

peppy:
Looks to work as expected, can't find any regressions in some initial testing.

Going to leave this unmerged as I figure @smoogipoo may want to double-check this. If not we can go ahead with merge.

smoogipoo (approved):

PR #5269, opened 2022-06-26
Initialise OpenTabletDriver asynchronously

peppy:
Startup time is abysmal of this component. I thought we already had this running async but seems not.

Timeline64_2022-06-26_11-23-59

### DIFF
--- a/osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
+++ b/osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
@@ -60,7 +60,7 @@ namespace osu.Framework.Input.Handlers.Tablet
                 {
                     lastInitTask = Task.Run(() =>
                     {
-                        tabletDriver = tabletDriver = TabletDriver.Create();
+                        tabletDriver = TabletDriver.Create();
                         tabletDriver.TabletsChanged += (_, e) =>
                         {
                             device = e.Any() ? tabletDriver.InputDevices.First() : null;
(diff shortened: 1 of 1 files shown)


osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
```diff
@@ -50,31 +53,37 @@ public override bool Initialize(GameHost host)
...
-                else if (!d.NewValue && tabletDriver != null)
+                else
                 {
-                    tabletDriver.Dispose();
-                    tabletDriver = null;
+                    initTask?.ContinueWith(_ =>
+                    {
+                        tabletDriver.Dispose();
+                        tabletDriver = null;
+                    });
```
frenzibyte, line 86: If a tablet driver initialisation is in progress, and the user disables the driver then enables it back again, wouldn't the newly began initialisation (from enabling the driver again) overwrite `tabletDriver`, and thereby cause this continuation of the last initialisation task to dispose of it?

peppy: I guess

bdach:
Pushed one minor fix ([d14f1c7](https://github.com/ppy/osu-framework/pull/5269/commits/d14f1c704dfa8cf054fda018ddd746622def5698)). I *think* this is fine otherwise.

bdach (approved):

smoogipoo (approved):

PR #5274, opened 2022-06-27
Fix pressing keys on a hardware keyboard showing the software keyboard

Susko3:
- Closes https://github.com/ppy/osu/issues/17561
- Closes https://github.com/ppy/osu/issues/6282 
   - can't repro, but not returning `null` and forcing `OnCheckIsTextEditor() => true` is the probable cause

[By default](https://user-images.githubusercontent.com/16479013/176005749-ecb6a979-1772-4d1a-99ef-d5808fcfe2d8.png), Android will show the software keyboard while text input is active even if a hardware keyboard is connected. And since we were always returning a valid `IInputConnection`, Android would "conveniently" show the software keyboard when any hardware keys are pressed.

Pressing keys will (re)focus the current view and probe `OnCreateInputConnection()` to see if the input method is to be activated.

The [docs](https://developer.android.com/reference/android/view/View#onCreateInputConnection(android.view.inputmethod.EditorInfo)) specify to return `null` when input methods are not supported by a specific `View`. Since we have only one `View`, returning `null` here means to disable text input for the time being (until a textbox is focused and `AndroidTextInput` is activated).

### DIFF
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -267,6 +267,7 @@ namespace osu.Framework.Android
             textInputActive = true;
             Activity.RunOnUiThread(() =>
             {
+                inputMethodManager.RestartInput(this); // this syncs the Android input method state with `OnCreateInputConnection()`.
                 RequestFocus();
                 inputMethodManager?.ShowSoftInput(this, 0);
             });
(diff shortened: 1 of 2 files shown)


peppy (approved):

peppy:
Probably best to get a second check on this one, looks okay code-wise though.

bdach (approved):

PR #5277, opened 2022-06-29
Fix slider bar drags not being relative to mouse down position

peppy:
Pretty large oversight. As mentioned over at https://github.com/ppy/osu/discussions/18938.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/SliderBar.cs
+++ b/osu.Framework/Graphics/UserInterface/SliderBar.cs
@@ -115,11 +115,27 @@ namespace osu.Framework.Graphics.UserInterface
         private void updateValue() => UpdateValue(NormalizedValue);
 
         private bool handleClick;
-        private float? relativeValueAtDragStart;
+        private float? relativeValueAtMouseDown;
 
         protected override bool OnMouseDown(MouseDownEvent e)
         {
-            handleClick = true;
+            if (HandleAsRelativeMovement(e))
+            {
+                float min = currentNumberInstantaneous.MinValue.ToSingle(NumberFormatInfo.InvariantInfo);
+                float max = currentNumberInstantaneous.MaxValue.ToSingle(NumberFormatInfo.InvariantInfo);
+                float val = currentNumberInstantaneous.Value.ToSingle(NumberFormatInfo.InvariantInfo);
+
+                relativeValueAtMouseDown = (val - min) / (max - min);
+                handleClick = false;
+            }
+            else
+            {
+                // Click shouldn't be handled if relative movement is happening.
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/UserInterface/SliderBar.cs
```diff
@@ -203,14 +217,36 @@ private bool commit()
             return true;
         }
 
-        private void handleMouseInput(UIEvent e)
+        /// <summary>
+        /// Whether mouse drag handling should be relative to the original mouse down position, or absolute based on the cursor.
```
smoogipoo, line 221: Just a precursory review, but I've been trying for the past 5 minutes to wrap my head around what "relative" and "absolute" means, and I haven't been able to do it yet.

Is there a reason for this sort of overengineering of terms? For example why couldn't this be called `IsHandleHovered` or `IsNubHovered` and do whatever "relative" and "absolute" means internally?

peppy: Not sure about terminology. There's no concept of a "nub" framework-side, so i wanted to avoid using that definition. For example, the basic slider bar doesn't have one at all.

I was going for the discerning factor that "relative" means the movement of the bar is equal to the distance the mouse travels, as opposed to updating precisely to where the mouse is.

I know I've seen at least one application *always* do relative movement (I'm sure there are valid cases) and others always do absolute. If you have better terms I'm open to suggestion, just not sure about "nub".

smoogipoo: Saying "relative to distance traveled rather than absolute based on the current cursor position" would be a huge improvement already.

peppy: So the xmldoc change alone is enough for you?

smoogipoo: That and the method should probably be renamed to `ShouldHandleAsRelativeDrag` since the user isn't expected to actually handle the drag.

peppy: The reason I used `Movement` and not `Drag` is that this also affected `OnClick`. Will go with drag if you think it feels okay given that.

frenzibyte: I personally would've much preferred for `SliderBar` to add the concept of an optional "nub" component and remove both `RangePadding` and `ShouldHandleAsRelativeDrag` in favour of it.

But the new xmldoc reads better at least, so I think we can leave the above for another time if we come across a similar issue.

frenzibyte (dismissed):

frenzibyte (approved):

PR #5278, opened 2022-06-29
Move joystick-axis-to-button simulation out of `JoystickHandler`

Susko3:
This will allow `AndroidJoystickHandler` (coming soon) to reuse this logic without code duplication.
Inheritance is not an option as `AndroidInputHandler` is used on Android.

This code does look a bit dubious as `JoystickButtonInput.Apply` is called within `JoystickAxisInput.Apply`, but the current order of events is preserved.

`KeyCombination` does have a similar [scroll delta to button](https://github.com/ppy/osu-framework/blob/ba1385330cc501f34937e08257e586c84e35d772/osu.Framework/Input/Bindings/KeyCombination.cs#L619-L632) conversion, but putting the logic there will break existing `JoystickButton.JoystickAxisN` usages.

### DIFF
--- a/osu.Framework/Input/Handlers/Joystick/JoystickHandler.cs
+++ b/osu.Framework/Input/Handlers/Joystick/JoystickHandler.cs
@@ -64,11 +64,8 @@ namespace osu.Framework.Input.Handlers.Joystick
         /// <summary>
         /// Enqueues a <see cref="JoystickAxisInput"/> taking into account the axis deadzone.
         /// </summary>
-        private void enqueueJoystickAxisChanged(JoystickAxis axis)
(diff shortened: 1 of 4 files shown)


peppy:
Does this mean the android joystick handler will need to have its own implementation of the deadzone?

Susko3:
I plan to make `rescaleByDeadzone` `internal static` to reuse the logic.

peppy (approved):

frenzibyte (approved):
I'm not really sure about the existence of "joystick axis buttons" instead of `KeyBindingContainer`-level ones, but I suppose this is fine for now.

Issue #2863, opened 2019-09-26
USB input device support in Android


DanielSvindseth:
Android Runs on a lot of devices, also tablets. Mouse support would mean a better experience more similar to those on traditional desktops and laptops.

Currently you can use a mouse but have to hold down M1 to move the in-game cursor, while disabeling mouse buttons (F10). A captured mouse pointer would be awesome.

**Proposal designs of the feature:** 

Since android O, `requestPointerCapture()` is able to capture the mouse-pointer like traditional platforms so a feature implementation would likely use this function.

EDIT: link to android developer about the function described: https://developer.android.com/training/gestures/movement#request_pointer_capture

- [x] Mouse support – https://github.com/ppy/osu-framework/pull/4989
- [ ] Joystick support – https://github.com/ppy/osu/discussions/18856

bdach:
Expanding scope to cover other potential currently unsupported USB devices (although mice/tablets are the primary ones to handle)

swoolcock:
This is less Android-specific now that iOS supports third party controllers.

frenzibyte:
Mouse support for Android is already in place here (which should cover tablet as well using pointer capture feature). @bdach anything else left to track here?

PR #5281, opened 2022-07-01
Add support for joysticks/gamepads on Android

Susko3:
- Closes https://github.com/ppy/osu/discussions/18856
- Closes https://github.com/ppy/osu-framework/issues/2863
   - tablet support still missing https://github.com/ppy/osu/issues/17382

Initial implementation based on https://developer.android.com/training/game-controllers/controller-input.
Axis and button values taken from there and by manually inspecting the android Keycode enum and docs.

During initial osu!-side testing I noticed that D-pad events were being dropped (on song select, where the game lags the most). The cause was not properly handling `MotionEvent` [batching](https://developer.android.com/reference/android/view/MotionEvent#batching), something that has already come up in [`AndroidMouseHandler`](https://github.com/ppy/osu-framework/blob/193238a90215b0e97ac2aa244ed7d30edc0a9880/osu.Framework.Android/Input/AndroidMouseHandler.cs#L224-L228). In fixing this, I refactored joystick, mouse and touch to use proper batching for all `MotionEvent`s.

Mouse and touch refactor is coming in a separate PR, but the groundwork has been laid out in this one: `HandleHistoricallyPerPointer` and `TryGetPosition` have been added, but currently remain unused.

---

Tested on a USB Xbox 360 controller clone, everything except the guide button works (Android doesn't provide a mapping for that). For my specific controller, D-pad inputs where reported exclusively trough `OnGenericMotion` &rarr; `Axis.HatX`/`Axis.HatY` and not in `OnKeyDown/Up`.

The same controller reported different `Device.MotionRanges` on different devices:
- Realme 6, Android 11: the list was non-null, but was empty
- LG G7 Fit, Android 9:  the list was populated with the expected axes (including `Gas` and `Brake`)

`availableAxes` can adapt to both of these scenarios, and properly excludes `Gas` and `Brake`.

### DIFF
--- a/osu.Framework.Android/Input/AndroidJoystickHandler.cs
+++ b/osu.Framework.Android/Input/AndroidJoystickHandler.cs
@@ -116,9 +116,28 @@ namespace osu.Framework.Android.Input
             var motionRanges = device.MotionRanges;
 
             availableAxes = motionRanges != null && motionRanges.Count > 0
-                // skip Dpad axes as they're handled separately in `applyDpadInput`
-                ? motionRanges.Select(m => m.Axis).Where(a => a != Axis.HatX && a != Axis.HatY).Distinct().ToList()
+                ? motionRanges.Select(m => m.Axis).Where(isValid).Distinct().ToList()
                 : AndroidInputExtensions.ALL_AXES;
+
+            bool isValid(Axis axis)
+            {
+                switch (axis)
+                {
+                    // D-pad axes are handled separately in `applyDpadInput`
+                    case Axis.HatX:
+                    case Axis.HatY:
+                    // Brake and Gas axes mirror the left and right trigger and are therefore ignored
+                    case Axis.Gas:
+                    case Axis.Brake:
+                        return false;
+                }
+
+                if (axis.TryGetJoystickAxisSource(out _))
+                    return true;
+
+                Logger.Log($"Unknown joystick axis: {axis}");
+                return false;
+            }
         }
 
         protected override void OnGenericMotion(MotionEvent genericMotionEvent)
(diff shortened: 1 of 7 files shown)


peppy (approved):
Code looks sane enough. Have not tested, hopefully someone else with a setup handy can do so.

osu.Framework.Android/Input/AndroidJoystickHandler.cs
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        /// <remarks>See xmldoc <see cref="AndroidKeyboardHandler.ShouldHandleEvent"/></remarks>
+        protected override bool ShouldHandleEvent(InputEvent? inputEvent)
+            => base.ShouldHandleEvent(inputEvent)
+               && (inputEvent.Source != InputSourceType.Keyboard
+                   // explicitly handle gamepad keycodes from "keyboards" (every gamepad button except the D-pad ones are included).
+                   // D-pad keycodes are overloaded; covering arrow keys on a keyboard, synthesized trackball (Android hardware feature from 2009) events,
+                   // and joystick/gamepad D-pads (according to the docs, but not yet seen in practice)
+                   || (inputEvent is KeyEvent keyEvent && KeyEvent.IsGamepadButton(keyEvent.KeyCode)));
```
Susko3, line 75: As it turns out, some controllers (steering wheel in my case) are reported as `InputSourceType.Keyboard` exactly (when pressing buttons). This additional check is required for those to work correctly.

Source for `KeyEvent.IsGamepadButton` (no D-pad in sight):

https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/KeyEvent.java;l=1899-1936;drc=11e61ab1fd1f868ee8ddd6fc86662f4f09df1a6a

peppy:
Going to get this in. It's a very fringe use case and I don't really have the energy to get a setup at my end (lacking correct USB cables, and not sure how well it would be supported on the android test device I have) but the code looks fine.

peppy (approved):

PepePondo:
Weird question, but would this be useable on desktop as well? I want to set up to play on my ROG Ally so I can play mania, and being able to use controllers instead of the touchscreen would be a godsend

smoogipoo:
Desktop has had support for controllers for a long time, yes.

eclipsek20:
I am trying to use an xbox wireless controller on android 13 and no input seems to be coming no matter what i press in the app.

PR #5282, opened 2022-07-01
Remove nullable disable annotation in the utils namespace

andy840119:
Remove nullable disable annotate in the utils namespace.
Seems there's no test case for the utils.

### DIFF
--- a/osu.Framework/Utils/Interpolation.cs
+++ b/osu.Framework/Utils/Interpolation.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Linq;
-using System.Reflection;
 using System.Runtime.Serialization;
 using osu.Framework.Extensions.Color4Extensions;
 using osu.Framework.Graphics;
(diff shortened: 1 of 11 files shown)


osu.Framework/Utils/Interpolation.cs
```diff
@@ -31,7 +28,7 @@ public static class Interpolation
         /// <param name="exponent">The exponent of the exponential. An exponent of 0 results in the start values, whereas larger exponents make the result converge to the final value.</param>
         public static double Damp(double start, double final, double @base, double exponent)
         {
-            if (@base < 0 || @base > 1)
+            if (@base is < 0 or > 1)
```
Susko3, line 31: I think the general consensus is against some of those patterns. Rider was suggesting similar patterns to me as well, so the checked-in configuration might not be correct/what we want.

osu.Framework/Utils/Interpolation.cs
```diff
@@ -384,17 +381,16 @@ static GenericInterpolation()
...
+                var valueAtMethod = typeof(GenericInterpolation<TEasing>).GetMethod(interpolation_method, parameters);
 
                 if (valueAtMethod != null)
                     FUNCTION = (InterpolationFunc<TValue, TEasing>)valueAtMethod.CreateDelegate(typeof(InterpolationFunc<TValue, TEasing>));
                 else
                 {
-                    var typeRef = FormatterServices.GetSafeUninitializedObject(typeof(TValue)) as IInterpolable<TValue>;
-
-                    if (typeRef == null)
+                    if (FormatterServices.GetSafeUninitializedObject(typeof(TValue)) is not IInterpolable<TValue> typeRef)
```
frenzibyte, line 393: This should also use the regular `!(get(typeof(TValue)) is IInterpolatable<TValue> typeRef)` syntax for now.

peppy (approved):

PR #5284, opened 2022-07-01
Allow disabling forced GC during test runs

peppy:
I'm not saying we should do this, but I want the flexibility to experiment with it. Once we have things in a stable state, it may make sense to disable for Github Actions to improve performance of test runs (from recollection, about 40% faster) and enable only when an issue comes up, or as a background runner on teamcity or otherwise.

### DIFF
--- a/osu.Framework/Testing/TestScene.cs
+++ b/osu.Framework/Testing/TestScene.cs
@@ -424,9 +424,12 @@ namespace osu.Framework.Testing
             runner.RunTestBlocking(this);
             checkForErrors();
 
-            // Force any unobserved exceptions to fire against the current test run.
-            // Without this they could be delayed until a future test scene is running, making tracking down the cause difficult.
-            collectAndFireUnobserved();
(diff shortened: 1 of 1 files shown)


osu.Framework/Testing/TestScene.cs
```diff
@@ -424,9 +424,12 @@ protected virtual void RunTestsFromNUnit()
             runner.RunTestBlocking(this);
             checkForErrors();
 
-            // Force any unobserved exceptions to fire against the current test run.
-            // Without this they could be delayed until a future test scene is running, making tracking down the cause difficult.
-            collectAndFireUnobserved();
+            if (Environment.GetEnvironmentVariable("OSU_FRAMEWORK_TESTS_NO_FORCED_GC") != "1")
```
frenzibyte, line 427: nit: Would remove the `FRAMEWORK` part to match `OSU_EXECUTION_MODE` (also because this affects all tests running under o!f, not framework tests specifically).

frenzibyte (approved):

PR #5287, opened 2022-07-03
Fix `TextureLoaderStore` not always appending extensions to lookups

frenzibyte:
- Closes https://github.com/ppy/osu/issues/18989

Tournament texture storage [uses a `StorageBackedResourceStore`](https://github.com/ppy/osu/blob/1ccfd69690c45aaad0057877aa5642896646d3d7/osu.Game.Tournament/TournamentGameBase.cs#L65), which is not of type `ResourceStore`, therefore `TextureLoaderStore` can't append the image extensions and fail lookups:

https://github.com/ppy/osu-framework/blob/e918d33889a39c4895c01bcca384e88c89b52a57/osu.Framework/Graphics/Textures/TextureLoaderStore.cs#L24-L29

This is a regression from https://github.com/ppy/osu-framework/pull/5240, since prior to that PR, [`TextureStore` appends extensions as well](https://github.com/ppy/osu-framework/blob/84758b770872e8f332ae0e5503206c5a941c89bb/osu.Framework/Graphics/Textures/TextureStore.cs#L44-L45) (this is also the reason why previously texture store sometimes looks up textures with double extensions, i.e. `texture.jpg.png`).

Instead, I've changed `TextureLoaderStore` to explicitly construct a resource store and add the given store along with the right extensions there, rather than safely cast and silently fail.

Note that `SampleStore` and `TrackStore` suffer from the same issue, which would fail similarly if constructed with a non-`ResourceStore<T>` store (e.g. `StorageBackedResourceStore`).

### DIFF
--- a/osu.Framework.Tests/Visual/Sprites/TestSceneTextures.cs
+++ b/osu.Framework.Tests/Visual/Sprites/TestSceneTextures.cs
@@ -229,7 +229,7 @@ namespace osu.Framework.Tests.Visual.Sprites
                 }
 
                 TotalCompletedLookups++;
-                return getFunc("sample-texture");
+                return getFunc("sample-texture.png");
             }
 
             public void Reset()
--- a/osu.Framework/Graphics/Textures/TextureLoaderStore.cs
+++ b/osu.Framework/Graphics/Textures/TextureLoaderStore.cs
@@ -19,13 +19,13 @@ namespace osu.Framework.Graphics.Textures
     /// </summary>
     public class TextureLoaderStore : IResourceStore<TextureUpload>
     {
-        private IResourceStore<byte[]> store { get; }
+        private readonly ResourceStore<byte[]> store;
 
         public TextureLoaderStore(IResourceStore<byte[]> store)
         {
-            this.store = store;
-            (store as ResourceStore<byte[]>)?.AddExtension(@"png");
-            (store as ResourceStore<byte[]>)?.AddExtension(@"jpg");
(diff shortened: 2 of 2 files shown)


peppy:
Tests need attention here.

frenzibyte:
Turns out in master, `TextureLoaderStore` populates the texture extensions to `game.Resources`, due to...

https://github.com/ppy/osu-framework/blob/604724157b84250bdf6ea02a7f1d124fd58ed8b5/osu.Framework/Game.cs#L181

https://github.com/ppy/osu-framework/blob/604724157b84250bdf6ea02a7f1d124fd58ed8b5/osu.Framework/Game.cs#L234-L235

(notice `Resources` passed to `addFont`, which creates a texture loader store of it) 

that is also the reason why we see lookups like `file.mp3.jpg` all over the place game-side.

This is relevant in the test failure because the test was never supposed to work without specifying the extension, but it was actually working because of the png/jpg extensions being added to `game.Resources`:
https://github.com/ppy/osu-framework/blob/4fef626f03dbf53f8954aa896526563dd1f0e42a/osu.Framework.Tests/Visual/Sprites/TestSceneTextures.cs#L264-L266

(notice `game.Resources` there in the store flow)

frenzibyte:
I'll run tests game-side and confirm this doesn't regress skins and storyboards just to be sure.

frenzibyte:
Can't seem to find any issues so far, this is ready for review.

peppy (approved):
Seems correct

PR #5288, opened 2022-07-03
Use null-forgiving operator for `AsNonNull` rather than assertion

frenzibyte:
~~Turns out `AsNonNull` never works osu!-side, due to the asserts being made for debug configuration only. Haven't checked if this works as expected when packaging framework yet (blocking until I confirm).~~

~~Alternative would be to explicitly throw exception, but not sure what's best fit for this.~~

I've been confused into thinking that `AsNonNull` should throw when the instance is null, but the intention behind it is to silently return the instance as non-nullable regardless of whether it's null (i.e. act exactly like a null-forgiving operator).

PR'ing this just to clear that confusion up.

### DIFF
--- a/osu.Framework/Extensions/ObjectExtensions/ObjectExtensions.cs
+++ b/osu.Framework/Extensions/ObjectExtensions/ObjectExtensions.cs
@@ -3,6 +3,7 @@
 
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using NUnit.Framework;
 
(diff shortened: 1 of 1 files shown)


frenzibyte:
Actually, the null-forgiving operator is meant to silence inspections only, not throw when the given reference is null. So I suppose this is intended? We might as well just use the null-forgiving operator here in this method instead.

smoogipoo:
I'd prefer using null-forgiving here instead.

smoogipoo (approved):

PR #5293, opened 2022-07-06
Remove noisy MIDI handler log output

peppy:
MIDI devices generally send traffic such as clock sync data, which we don't care about. For me, if I have my midi controller connected this creates a flow of 2-3 logger messages per second.

Uploading JetBrains Rider-EAP 2022-07-06 at 05.12.44.mp4…


smoogipoo (dismissed):
I'd say the on/off messages are also pretty noisy, but sure.

peppy:
About as noisy as other input events. Can probably be removed though? Generally we only care when they trigger something.

smoogipoo (approved):

PR #5295, opened 2022-07-06
Reduce `Scheduler` overhead when no tasks are pending

peppy:
A common scenario is that a drawable will use a `Schedule` call once then become mostly inactive. The `Scheduler` that was used for this call stays remnant, and has quite a large overhead due to the locking in `Scheduler.Update`. This aims to reduce that overhead considerably.

## Benchmark

Before:

|                         Method |        Mean |     Error |    StdDev |  Gen 0 | Allocated |
|------------------------------- |------------:|----------:|----------:|-------:|----------:|
|           UpdateEmptyScheduler |    82.15 us |  0.662 us |  0.553 us |      - |         - |
| UpdateSchedulerWithManyDelayed | 1,389.42 us | 20.222 us | 17.926 us |      - |       3 B |
| UpdateSchedulerWithEveryUpdate |   147.62 us |  2.294 us |  2.034 us |      - |         - |
|   UpdateSchedulerWithManyAdded |   144.63 us |  2.354 us |  2.087 us | 6.8359 |  80,000 B |

After:

|                         Method |         Mean |      Error |     StdDev |  Gen 0 | Allocated |
|------------------------------- |-------------:|-----------:|-----------:|-------:|----------:|
|           UpdateEmptyScheduler |     2.831 us |  0.0498 us |  0.0441 us |      - |         - |
| UpdateSchedulerWithManyDelayed | 2,148.074 us | 30.7526 us | 28.7660 us |      - |       6 B |
| UpdateSchedulerWithEveryUpdate |   104.743 us |  1.6779 us |  1.5695 us |      - |         - |
|   UpdateSchedulerWithManyAdded |    81.094 us |  1.5121 us |  1.4144 us | 6.8359 |  80,000 B |

I have... no idea why `WithManyDelayed` is showing a performance regression. There's no reason it should be. Maybe @smoogipoo will have a better immediate idea? I've tried multiple runs and it's always showing the same numbers..

## Real world (song select, idle)

Before:

JetBrains Rider-EAP 2022-07-06 at 08 07 22

After:

JetBrains Rider-EAP 2022-07-06 at 08 06 10

### DIFF
--- a/osu.Framework/Threading/Scheduler.cs
+++ b/osu.Framework/Threading/Scheduler.cs
@@ -136,63 +136,71 @@ namespace osu.Framework.Threading
 
         private void queueTimedTasks()
         {
-            double currentTimeLocal = currentTime;
-
-            foreach (var sd in timedTasks)
+            // Already checked before this method is called, but helps with path prediction?
+            if (timedTasks.Count != 0)
             {
-                if (sd.ExecutionTime <= currentTimeLocal)
+                double currentTimeLocal = currentTime;
+
+                foreach (var sd in timedTasks)
                 {
-                    tasksToRemove.Add(sd);
-
-                    if (sd.Cancelled) continue;
-
-                    if (sd.RepeatInterval == 0)
+                    if (sd.ExecutionTime <= currentTimeLocal)
                     {
-                        // handling of every-frame tasks is slightly different to reduce overhead.
-                        perUpdateTasks.Add(sd);
-                        continue;
+                        tasksToRemove.Add(sd);
+
+                        if (sd.Cancelled) continue;
+
+                        if (sd.RepeatInterval == 0)
+                        {
+                            // handling of every-frame tasks is slightly different to reduce overhead.
+                            perUpdateTasks.Add(sd);
+                            continue;
+                        }
+
+                        if (sd.RepeatInterval > 0)
+                        {
+                            if (timedTasks.Count > LOG_EXCESSSIVE_QUEUE_LENGTH_INTERVAL)
+                                throw new ArgumentException("Too many timed tasks are in the queue!");
+
+                            // schedule the next repeat of the task.
+                            sd.SetNextExecution(currentTimeLocal);
+                            tasksToSchedule.Add(sd);
+                        }
+
+                        if (!sd.Completed) enqueue(sd);
                     }
-
-                    if (sd.RepeatInterval > 0)
-                    {
-                        if (timedTasks.Count > LOG_EXCESSSIVE_QUEUE_LENGTH_INTERVAL)
-                            throw new ArgumentException("Too many timed tasks are in the queue!");
-
-                        // schedule the next repeat of the task.
-                        sd.SetNextExecution(currentTimeLocal);
-                        tasksToSchedule.Add(sd);
-                    }
-
-                    if (!sd.Completed) enqueue(sd);
                 }
+
+                foreach (var t in tasksToRemove)
+                    timedTasks.Remove(t);
+
+                tasksToRemove.Clear();
+
+                foreach (var t in tasksToSchedule)
+                    timedTasks.AddInPlace(t);
+
+                tasksToSchedule.Clear();
             }
-
-            foreach (var t in tasksToRemove)
-                timedTasks.Remove(t);
-
-            tasksToRemove.Clear();
-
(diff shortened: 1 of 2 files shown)


osu.Framework/Threading/Scheduler.cs
```diff
@@ -100,20 +100,29 @@ public void UpdateClock(IClock newClock)
...
+            bool hasPerUpdateTasks = perUpdateTasks.Count > 0;
+
+            if (hasTimedTasks || hasPerUpdateTasks) // avoid taking out a lock if there are no items.
             {
-                queueTimedTasks();
-                queuePerUpdateTasks();
+                lock (queueLock)
+                {
+                    if (hasTimedTasks) queueTimedTasks();
+                    if (perUpdateTasks.Count > 0) queuePerUpdateTasks();
```
frenzibyte, line 113:
```suggestion
                    if (hasPerUpdateTasks) queuePerUpdateTasks();
```

osu.Framework/Threading/Scheduler.cs
```diff
@@ -100,20 +100,29 @@ public void UpdateClock(IClock newClock)
...
+            bool hasTimedTasks = timedTasks.Count > 0;
+            bool hasPerUpdateTasks = perUpdateTasks.Count > 0;
+
+            if (hasTimedTasks || hasPerUpdateTasks) // avoid taking out a lock if there are no items.
             {
-                queueTimedTasks();
-                queuePerUpdateTasks();
+                lock (queueLock)
+                {
+                    if (hasTimedTasks) queueTimedTasks();
```
frenzibyte, line 112: Would assume the same concern about per-update tasks applies to timed tasks as well?

```suggestion
                    if (timedTasks.Count > 0) queueTimedTasks();
```

Would also inline the local since it's not correct to store it in a local in the first place.

peppy: No, the concern is only for the second, since the first method potentially adds the second type.

peppy: I've removed the inner checks since they don't bring much value to the table.

peppy:
> Interestingly, I don't see the performance regression mentioned in OP upon running the benchmark on my machine:

I can still repro the difference... will try and figure it out since it's a bit alarming to me.

smoogipoo:
I can also repro a little bit of slowdown (300us):

Master:

|                         Method |        Mean |      Error |    StdDev |  Gen 0 | Allocated |
|------------------------------- |------------:|-----------:|----------:|-------:|----------:|
|           UpdateEmptyScheduler |    65.01 us |   0.171 us |  0.009 us |      - |         - |
| UpdateSchedulerWithManyDelayed | 1,452.63 us | 263.561 us | 14.447 us |      - |       3 B |
| UpdateSchedulerWithEveryUpdate |   113.26 us |   1.446 us |  0.079 us |      - |         - |
|   UpdateSchedulerWithManyAdded |   114.33 us |  23.513 us |  1.289 us | 0.8545 |  80,000 B |

This PR:
|                         Method |         Mean |       Error |     StdDev |  Gen 0 | Allocated |
|------------------------------- |-------------:|------------:|-----------:|-------:|----------:|
|           UpdateEmptyScheduler |     2.893 us |   0.1699 us |  0.0093 us |      - |         - |
| UpdateSchedulerWithManyDelayed | 1,707.047 us | 232.6123 us | 12.7503 us |      - |       3 B |
| UpdateSchedulerWithEveryUpdate |   115.545 us |  10.8061 us |  0.5923 us |      - |         - |
|   UpdateSchedulerWithManyAdded |    82.162 us |   3.2974 us |  0.1807 us | 0.8545 |  80,000 B |

---

Seems to be caused by the removal of the count check in `updateTimedTasks`.

peppy:
How can that be the case though? I explicitly added the count check (removed in the last commit) to ensure it wasn't contributing to anything. Can you test with that commit reverted and comment on why the location being different matters if you have an idea?

smoogipoo:
Reverting the last commit doesn't fix it. The only thing that seems to fix it is moving the check inside.

I have a feeling this is more about branch prediction hinting. I did some testing on this a while back to see how C# behaves in the context of early returns, but I'm trying to remember the specifics...

smoogipoo (approved):

Issue #5033, opened 2022-02-14
Windows touch input with raw input turned on does not map correctly


peppy:
Can reproduce on any surface device (along with other windows touch devices), where cursor position updates work correctly but are constricted to a very small area in the top left of the screen, rather than covering the full window. It looks like the coordinate mapping is going wrong, although I haven't had a chance to investigate further:

https://github.com/ppy/osu-framework/blob/6a2250ee584f1e38601b5cdb4b7397d27f730304/osu.Framework/Platform/Windows/WindowsMouseHandler.cs#L93

As discussed in https://github.com/ppy/osu/discussions/16840#discussioncomment-2155525.

Issue #1514, opened 2018-04-13
Improve touchscreen support on desktop platforms


peppy:
Currently clicks are being handled incorrectly, and generally support doesn't exist.

Check out https://blogs.msdn.microsoft.com/dotnet/2017/04/05/announcing-the-net-framework-4-7/ (search `WPF Touch/Stylus support for Windows 10`) for a potential good starting point for better windows touch handling.

FreezyLemon:
This looks really nice, but is probably not usable in osu-Framework since all the required classes are in `PresentationCore.dll` which is the WPF "core" module ([example 1](https://docs.microsoft.com/en-us/dotnet/api/system.windows.input.stylus?view=netframework-4.7.1), [2](https://docs.microsoft.com/en-us/dotnet/api/system.windows.input.touch?view=netframework-4.7.1)).
I'm not sure whether adding this as a dependency is wanted.. though just as a reference, I'll mention that the file is ~1.3MB in size.

We might be able to implement this at a lower level in Windows (look at example 2, especially the `FrameReported` event, there are strong hints at it just being a native win API function call) without needing those classes though.

SpyHoodle:
Peppy, so your saying you don't know how to fix a massive hole in the upcoming osu!?
I'm playing with tablet on lazer and it doesnt work at all.
Please find SOME way to fix

peppy:
That was never said, no. It is just pending behind hundreds of other tasks that also need to be completed. Please have some patience.

SpyHoodle:
;-; soz

VPeruS:
WPF will be available in 3.0 but only for windows

iiSaLMaN:
Seems to be a high priority for touchscreen users.

PR #5299, opened 2022-07-07
Touch support for desktop platforms

Susko3:
- Closes https://github.com/ppy/osu-framework/issues/1514
- Closes https://github.com/ppy/osu-framework/issues/5033
- Closes https://github.com/ppy/osu/issues/18889

Touch currently uses [mouse emulation](https://github.com/libsdl-org/SDL/blob/6a2e6c82a0764a00123447d93999ebe14d509aa8/include/SDL_hints.h#L1367-L1376) on desktop platforms. This PR sets the correct flag and implements [`SDL_TouchFingerEvent`](https://wiki.libsdl.org/SDL_TouchFingerEvent) handling. Windows also emulates mouse events for touch, and those events have also been properly ignored in `WindowsMouseHandler`.

Initial testing [by peppy on discord](https://discord.com/channels/188630481301012481/589331078574112768/993813779160760340).

Raw input touch event detection [inspired](https://github.com/libsdl-org/SDL/blob/85bbf8eec9f5e422447807b6cabec8c9ec0859d9/src/video/windows/SDL_windowsevents.c#L825) [by](https://github.com/libsdl-org/SDL/blob/85bbf8eec9f5e422447807b6cabec8c9ec0859d9/src/video/windows/SDL_windowsevents.c#L521-L537) [SDL](https://github.com/libsdl-org/SDL/blob/85bbf8eec9f5e422447807b6cabec8c9ec0859d9/src/video/windows/SDL_windowsevents.c#L509-L511) and confirmed by [Microsoft Docs](https://docs.microsoft.com/en-us/windows/win32/tablet/system-events-and-mouse-messages). (The `0x82` check seems to be unnecessary, as it's already covered by `0x80`.)

`getTouchSource` / `assignNextAvailableTouchSource` logic taken from iOS touch handler.

### DIFF
--- a/osu.Framework/Platform/Windows/WindowsMouseHandler.cs
+++ b/osu.Framework/Platform/Windows/WindowsMouseHandler.cs
@@ -64,7 +64,7 @@ namespace osu.Framework.Platform.Windows
             if (message != Native.Input.WM_INPUT)
                 return IntPtr.Zero;
 
-            if (Native.Input.IsMouseMessageSourceTouch())
+            if (Native.Input.IsTouchEvent(Native.Input.GetMessageExtraInfo()))
                 return IntPtr.Zero;
 
             int payloadSize = sizeof(RawInputData);
--- a/osu.Framework/Platform/Windows/WindowsMouseHandler.cs
+++ b/osu.Framework/Platform/Windows/WindowsMouseHandler.cs
@@ -65,6 +65,7 @@ namespace osu.Framework.Platform.Windows
                 return IntPtr.Zero;
 
             if (Native.Input.IsTouchEvent(Native.Input.GetMessageExtraInfo()))
+                // sometimes GetMessageExtraInfo returns 0, so additionally, mouse.ExtraInformation is checked below.
                 // touch events are handled by TouchHandler
                 return IntPtr.Zero;
 
@@ -76,6 +77,10 @@ namespace osu.Framework.Platform.Windows
                 return IntPtr.Zero;
 
             var mouse = data.Mouse;
+
+            // `ExtraInformation` doens't have the MI_WP_SIGNATURE set, so we have to rely solely on the touch flag.
+            if (Native.Input.HasTouchFlag(mouse.ExtraInformation))
+                return IntPtr.Zero;
 
             //TODO: this isn't correct.
             if (mouse.ExtraInformation > 0)
--- a/osu.Framework/Platform/Windows/WindowsMouseHandler.cs
+++ b/osu.Framework/Platform/Windows/WindowsMouseHandler.cs
@@ -64,6 +64,9 @@ namespace osu.Framework.Platform.Windows
             if (message != Native.Input.WM_INPUT)
                 return IntPtr.Zero;
 
+            if (Native.Input.IsMouseMessageSourceTouch())
+                return IntPtr.Zero;
+
             int payloadSize = sizeof(RawInputData);
 
             Native.Input.GetRawInputData((IntPtr)lParam, RawInputCommand.Input, out var data, ref payloadSize, sizeof(RawInputHeader));
--- a/osu.Framework/Platform/Windows/WindowsMouseHandler.cs
+++ b/osu.Framework/Platform/Windows/WindowsMouseHandler.cs
@@ -65,6 +65,7 @@ namespace osu.Framework.Platform.Windows
                 return IntPtr.Zero;
 
             if (Native.Input.IsTouchEvent(Native.Input.GetMessageExtraInfo()))
+                // touch events are handled by TouchHandler
                 return IntPtr.Zero;
 
             int payloadSize = sizeof(RawInputData);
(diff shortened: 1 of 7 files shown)


peppy:
All working well now, until raw input is turned on, where things still seem to fall apart (movement seems to work but drag operations don't work, and tap operations only sometimes work).

I'll investigate further tomorrow if the issue isn't something obvious.

osu.Framework/Platform/Windows/WindowsMouseHandler.cs
```diff
@@ -64,6 +64,10 @@ private IntPtr onWndProc(IntPtr userData, IntPtr hWnd, uint message, ulong wPara
             if (message != Native.Input.WM_INPUT)
                 return IntPtr.Zero;
 
+            if (Native.Input.IsTouchEvent(Native.Input.GetMessageExtraInfo()))
+                // touch events are handled by TouchHandler
+                return IntPtr.Zero;
```
peppy, line 69: So this never gets hit in testing.

I also tested

```csharp
if (Native.Input.IsTouchEvent(data.Mouse.ExtraInformation))
```

and it also returns false even for touch.

Using `mouse.ExtraInformation>0` (see incorrect commented out version below) *does* work.

For reference, `ExtraInformation` is `139` in the touch event I tested.

Susko3: What if you comment out our [custom handling](https://github.com/ppy/osu-framework/blob/ba1385330cc501f34937e08257e586c84e35d772/osu.Framework/Platform/Windows/WindowsGameHost.cs#L62-L68) and let SDL do everything? Curious if SDL will properly filter out the raw mouse touch events.

Maybe the `|| Native.Input.GetMessageExtraInfo() & 0x82 == 0x82` check really is necessary.

frenzibyte: Upon blaming through SDL, they seem to have been using the same `IsTouchEvent` method used here, until 2 years ago, in which they added an additional condition [with the intent of fixing touch input being sent as regular mouse messages](https://github.com/libsdl-org/SDL/commit/0721931f5b92c96ff19bf936186bc4cf3dac38a1), I've also found this [thread](https://answers.microsoft.com/en-us/windows/forum/all/getmessageextrainfo-signature-missing-for-touch/f7ad7289-fd15-4061-98c5-2d90a176fe7c) from a year ago as well, which touches on the fact that the touch signature (`0xFF515700`) being missing (along with the touch identifier bit, except that it does seem to be present with `ExtraInformation` here).

I think that concludes the necessity of the `GetMessageExtraInfo() & 0x82 == 0x82` condition.

Susko3: I can get behind only checking for `0x80` (and omitting the `MI_WP_SIGNATURE_MASK` check), but the `0x02` there confuses me.

> https://docs.microsoft.com/en-us/windows/win32/tablet/system-events-and-mouse-messages:
> The lower 8 bits returned from `GetMessageExtraInfo` are variable. Of those bits, 7 (the lower 7, masked by 0x7F) are used to represent the cursor ID, zero for the mouse or a variable value for the pen ID.

Checking for `0x02` is probably wrong. [This thread](https://docs.microsoft.com/en-us/answers/questions/285548/getmessageextrainfo-signature-missing-for-touch-in.html) confirms that; the received values were `0xC7/C8/F7/F8/F9` -- some of which don't have `0x2` set.

Why I think just `0x80` is fine:
> https://docs.microsoft.com/en-us/windows/win32/tablet/system-events-and-mouse-messages (cont.):
> Additionally, in Windows Vista, the eighth bit, masked by 0x80, is used to differentiate touch input from pen input (0 = pen, 1 = touch).

---

> https://discord.com/channels/188630481301012481/589331078574112768994938919701467226
> `GetMessageExtraInfo` always seemed to return 0

It seems `GetMessageExtraInfo` is not reliable, so `mouse.ExtraInformation` is what we want to use. It contains at least some useful information, as peppy said [above](https://github.com/ppy/osu-framework/pull/5299#discussion_r916723692).

peppy:
Works perfectly now!

osu.Framework/Platform/Windows/WindowsMouseHandler.cs
```diff
@@ -64,6 +64,11 @@ private IntPtr onWndProc(IntPtr userData, IntPtr hWnd, uint message, ulong wPara
             if (message != Native.Input.WM_INPUT)
                 return IntPtr.Zero;
 
+            if (Native.Input.IsTouchEvent(Native.Input.GetMessageExtraInfo()))
+                // sometimes GetMessageExtraInfo returns 0, so additionally, mouse.ExtraInformation is checked below.
+                // touch events are handled by TouchHandler
+                return IntPtr.Zero;
```
peppy, line 70: I'd probably argue this is unnecessary at this point, but probably no harm in leaving it here..

peppy (approved):

PR #5301, opened 2022-07-13
Refactor `DirectorySelector` to allow custom layouts

ggliv:
Necessary for the application of https://github.com/ppy/osu/pull/19048#discussion_r918566267.

Also changes basic implementations to use `BasicCheckbox` instead of `BasicButton` so the toggle's current state is observable.


peppy:
This seems like a bad way to go about it. We shouldn't have to change the framework class every time we want to change the dialog layout. The `abstract` implementation should likely be refactored.

bdach (changes requested):

PR #5302, opened 2022-07-13
Fix running Bass tests causing deadlocks on Linux

smoogipoo:
I'm still in talks with Ian since this shouldn't be happening anyway, but what's happening is that Bass will call `Bass_Free()` on application exit (i.e. after all user code has exited) and deadlock because we haven't removed the track syncs.

I don't think there's any reason for us to not remove the syncs on disposal. But not only were we not removing the track syncs - the tests weren't disposing the components at all.

The deadlocks can be reproed on Linux by starting `TrackBassTest.TestBitrate` with debugger attached.

### DIFF
--- a/osu.Framework.Tests/Audio/BassTestComponents.cs
+++ b/osu.Framework.Tests/Audio/BassTestComponents.cs
@@ -65,8 +65,6 @@ namespace osu.Framework.Tests.Audio
         {
             RunOnAudioThread(() =>
             {
-                Mixer.Update();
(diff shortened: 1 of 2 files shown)


peppy (approved):

smoogipoo:
Bass forum thread here: https://www.un4seen.com/forum/?topic=19827.0 Though most of the conversation (where it was determined the issue is with syncs) is happening via emails.

Issue #5231, opened 2022-06-06
`SafeWriteStream` may not need to `Flush` on disposal


peppy:
At some point I'd like to confirm whether this is required. the `Flush` call adds an overhead, but now that we are `File.Move`ing the output I'd want to know whether this overhead still exists on the `Move` operation, and if so, whether removing the inheritance is in our best interest (from either a simplicity or performance angle).

smoogipoo:
The flush is intended to prevent 0-byte files. Do we have another mechanism for this somewhere?

peppy:
I'm just curious if the `Move` operation will cause the flush to occur implicitly. Unsure how the windows file IO API works.

PR #5305, opened 2022-07-15
Remove flush-to-disk operation on non-Windows platforms for lower overhead

frenzibyte:
- Closes https://github.com/ppy/osu/issues/19116
- See discussion on [Discord](https://discord.com/channels/188630481301012481/188630652340404224/997304567526080583)

The operation was initially added to fix a Windows-specific issue regarding zeros being written to the file, and after switching to .NET 6, the behaviour of the operation changed on macOS to become much stronger / more expensive (see https://github.com/dotnet/runtime/pull/44443).

We do not want this overhead on non-Windows platforms as there has been no signs of such issues occurring there, therefore the flush-to-disk operation is now limited to Windows platforms (we may eventually not need flushing at all on `SafeWriteStream`, see https://github.com/ppy/osu-framework/issues/5231).

### DIFF
--- a/osu.Framework/Platform/Storage.cs
+++ b/osu.Framework/Platform/Storage.cs
@@ -183,7 +183,8 @@ namespace osu.Framework.Platform
                     // this was added to work around some hardware writing zeroes to a file
                     // before writing actual content, causing corrupt files to exist on disk.
                     // as of .NET 6, flushing is very expensive on macOS so this is limited to only Windows,
-                    // but it may be entirely unnecessary due to the temporary file copying performed on this class.
+                    // but it may also be entirely unnecessary due to the temporary file copying performed on this class.
+                    // see: https://github.com/ppy/osu-framework/issues/5231
                     if (RuntimeInfo.OS == RuntimeInfo.Platform.Windows)
                     {
                         try
--- a/osu.Framework/Platform/Storage.cs
+++ b/osu.Framework/Platform/Storage.cs
@@ -160,7 +160,7 @@ namespace osu.Framework.Platform
         /// <summary>
         /// Uses a temporary file to ensure a file is written to completion before existing at its specified location.
         /// </summary>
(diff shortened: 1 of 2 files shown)


osu.Framework/Platform/Storage.cs
```diff
@@ -178,6 +178,22 @@ public SafeWriteStream(string temporaryPath, string finalPath, Storage storage)
...
+                        {
+                            Flush(true);
+                        }
+                        catch
+                        {
+                            // this may fail due to a lower level file access issue.
+                            // we don't want to throw in disposal though.
+                        }
+                    }
+                }
```
frenzibyte, line 195: I don't like this being duplicated from `FlushingStream`, but that's the only way to have it still exist but under a runtime condition (there's no way to do `FileStream(FileStream)`).

I've also considered moving the condition inside `FlushingStream`, but not sure that makes any sense (especially when it's wanted elsewhere), unless we start renaming it to `PlatformSpecificFlushingStream` or otherwise.

osu.Framework/Platform/Storage.cs
```diff
@@ -178,6 +178,22 @@ public SafeWriteStream(string temporaryPath, string finalPath, Storage storage)
 
             protected override void Dispose(bool disposing)
             {
+                if (!isDisposed)
+                {
+                    if (RuntimeInfo.OS == RuntimeInfo.Platform.Windows)
```
peppy, line 183: I'd add an inline comment here, something like

"This was added to workaround some hardware writing zeroes to a file before writing actual content, causing corrupt files to exist on disk. As of .NET 6, flushing is very expensive on macOS so this is limited to only windows, but it may also be unecessary on windows due to the temporary file copying performed by this class (see https://github.com/ppy/osu-framework/issues/5231)."

peppy (approved):

PR #5307, opened 2022-07-17
Remove nullable disable annotation in the text namespace.

andy840119:
Remove nullable disable annotate in the text namespace.


peppy:
@andy840119 i'm not sure it's a good time to be making these changes to these files, as there is large ongoing refactoring to textures etc. which will likely conflict.

andy840119:
I guess I will create another PR after refactor done🤔

PR #5308, opened 2022-07-17
Remove nullable disable annotation in the `Development` namespace.

andy840119:
Remove nullable disable annotation in the `Development` namespace.

### DIFF
--- a/osu.Framework/Development/ThreadSafety.cs
+++ b/osu.Framework/Development/ThreadSafety.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-#nullable disable
-
 using System;
 using System.Diagnostics;
 using osu.Framework.Platform;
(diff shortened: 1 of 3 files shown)


peppy:
Not sure what the code quality failure is here, definitely a false positive.

osu.Framework/Development/ThreadSafety.cs
```diff
@@ -86,7 +84,7 @@ public static class ThreadSafety
         /// <summary>
         /// The current execution mode.
         /// </summary>
-        internal static ExecutionMode ExecutionMode;
+        internal static ExecutionMode ExecutionMode = default;
```
smoogipoo, line 87: What is this doing? It's assigning a default value to a non-nullable value type.

smoogipoo (changes requested):

osu.Framework/Development/ThreadSafety.cs
frenzibyte: Probably a failed attempt at trying to appease code quality CI, which somehow has a false-positive failure (see https://github.com/ppy/osu-framework/runs/7381387221?check_suite_focus=true).

andy840119: I'll make this change back.
Still got no idea why the CI check failed @frenzibyte mentioned.

PR #5316, opened 2022-07-19
Fix bad access to `Waveform.GetPoints()` if waveform changes during regeneration

peppy:
Was so focused on the `task.GetResultSafely()` line that I missed the issue being in the `Logger` call itself. The outer waveform could change during the async `ContinueWith` operation, which would cause undefined behaviour. Note that this was an issue before the logger call was added – it was only brought to light by the recent failures. Probably leading to weirdness in other ways.

Closes https://github.com/ppy/osu-framework/issues/5314

### DIFF
--- a/osu.Framework/Audio/Track/Waveform.cs
+++ b/osu.Framework/Audio/Track/Waveform.cs
@@ -292,11 +292,7 @@ namespace osu.Framework.Audio.Track
         /// </summary>
         public async Task<List<Point>> GetPointsAsync()
         {
-            if (readTask == null)
-                return points;
-
             await readTask.ConfigureAwait(false);
-
             return points;
         }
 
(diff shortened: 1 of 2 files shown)


peppy:
To test this, I've been using local checkout of framework and running osu-side `TestSceneEditorTestGameplay.TestSharedClockState` on "run until fail". Without this change, it will generally fail in one or two iterations.

osu.Framework/Audio/Track/Waveform.cs
```diff
@@ -292,11 +292,7 @@ public async Task<Waveform> GenerateResampledAsync(int pointCount, CancellationT
         /// </summary>
         public async Task<List<Point>> GetPointsAsync()
         {
-            if (readTask == null)
-                return points;
```
smoogipoo, line 296: Is this safe...? There's an early return in the ctor before `readTask` is set.

peppy: That's *super* dodge.

smoogipoo (approved):

smoogipoo (approved):
Looks much better

PR #5317, opened 2022-07-19
Android input handlers refactor

Susko3:
Changes include:
- Incrementing `FrameStatistics` when enqueuing input
- Full NRT coverage
- **Logging of unhandled events**
   - Should help when adding new input types (eg. proper stylus handling) and to provide more context to input-related bug reports.
- Added some missing keyboard keys
- Historic handling of non-relative mouse input
- **Historic handling of touch input**
- Proper `TouchSource` validation for touch input
- General cleanup

Historic handling of touch input
----

Explainer: https://developer.android.com/reference/android/view/MotionEvent#batching

This is especially important if the game/device is lagging. Previously we were dropping a lot of events. I expect the UI to feel marginally smoother with historic handling, and scrolling/dragging to be more precise.

Before | After (with batching, historic handling)
---|---
before | after


osu.Framework.Android/Input/AndroidInputHandler.cs
```diff
@@ -30,6 +31,8 @@ public abstract class AndroidInputHandler : InputHandler
         /// </summary>
         protected readonly AndroidGameView View;
 
+        public override string Description => base.Description.Replace("Android", null);
```
peppy, line 34: This reads a bit weird. Is this to make things display correctly in settings?

peppy: Can we just leave the `Android` in the description? If it's not displayed anywhere else, I think I'd prefer it being there in logs.

Susko3: In that case, the `Description` can stay as-is, but logs will use `GetType().ReadableName()`

peppy: Still a bit confused here. Can we revert the `Replace` part?

peppy (approved):

PR #5325, opened 2022-07-29
Internalise vertex batches and expose via `IVertexBatch`

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/5324

- Makes `VertexBatch`,`VertexBuffer`, et-al `internal` and moves them to the OpenGL namespace.
- Exposes methods to create `IVertexBatch<T>` via `IRenderer`.
- Adds `PrimitiveTopology`, replacing the OpenGL-specific `PrimitiveType`.

# vNext

## `QuadBatch<T>` and `LinearBatch<T>` are no longer exposed

They should instead be created via the `IRenderer` and stored as an `IVertexBatch<T>` :

```diff
class MyDrawNode : DrawNode
{
-   private readonly QuadBatch<TexturedVertex2D> myBatch = new QuadBatch<TexturedVertex2D>(1, 1);
+   private IVertexBatch<TexturedVertex2D> myBatch;

    public override void Draw(IRenderer renderer)
    {
        base.Draw(renderer);

+       myBatch ??= renderer.CreateQuadBatch<TexturedVertex2D>(1, 1);
    }

    protected override void Dispose(bool disposing)
    {
        base.Dispose(disposing);
        batch?.Dispose();
    }
}
```

## OpenGL-specific `PrimitiveType` no longer used to create vertex batches

The new `PrimitiveTopology` enumeration should be used instead.

### DIFF
--- a/osu.Framework/Graphics/OpenGL/Batches/QuadBatch.cs
+++ b/osu.Framework/Graphics/OpenGL/Batches/QuadBatch.cs
@@ -8,9 +8,9 @@ using osu.Framework.Graphics.OpenGL.Buffers;
 using osu.Framework.Graphics.OpenGL.Vertices;
 using osuTK.Graphics.ES30;
 
-namespace osu.Framework.Graphics.Batches
+namespace osu.Framework.Graphics.OpenGL.Batches
 {
-    public class QuadBatch<T> : VertexBatch<T>
+    internal class QuadBatch<T> : VertexBatch<T>
         where T : struct, IEquatable<T>, IVertex
     {
         public QuadBatch(int size, int maxBuffers)
--- a/osu.Framework/Graphics/Rendering/IRenderer.cs
+++ b/osu.Framework/Graphics/Rendering/IRenderer.cs
@@ -1,6 +1,10 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
+using osu.Framework.Graphics.OpenGL.Vertices;
+using osuTK.Graphics.ES30;
+
 namespace osu.Framework.Graphics.Rendering
 {
     /// <summary>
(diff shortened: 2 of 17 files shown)


osu.Framework/Graphics/OpenGL/Batches/QuadBatch.cs
```diff
@@ -8,9 +8,9 @@
 using osu.Framework.Graphics.OpenGL.Vertices;
 using osuTK.Graphics.ES30;
 
-namespace osu.Framework.Graphics.Batches
+namespace osu.Framework.Graphics.OpenGL.Batches
 {
-    public class QuadBatch<T> : VertexBatch<T>
+    internal class QuadBatch<T> : VertexBatch<T>
```
frenzibyte, line 13: I would imagine this to be a good opportunity for prefixing all these classes with `OpenGL`, since it would be conflicting with the future Veldrid variants (unless we're fine with having same class name on different namespaces, I personally find it confusing to read).

osu.Framework/Graphics/Rendering/IRenderer.cs
```diff
@@ -1,12 +1,34 @@
...
 namespace osu.Framework.Graphics.Rendering
 {
     /// <summary>
     /// Draws to the screen.
     /// </summary>
     public interface IRenderer
     {
+        public const int VERTICES_PER_QUAD = 4;
+
+        public const int VERTICES_PER_TRIANGLE = 4;
```
frenzibyte, line 16: nit: These look to be unused currently, were they supposed to replace `TextureGLSingle`?

frenzibyte (commented):
Looks good otherwise.

osu.Framework/Graphics/OpenGL/Batches/QuadBatch.cs
smoogipoo: I want to leave renamings for a larger change at the end, not interspersed between PRs.

osu.Framework/Graphics/Rendering/IRenderer.cs
smoogipoo: One of them is used osu-side. Yes, these will replace the TextureGL ones but there’s no changes to textures yet.

frenzibyte (approved):

peppy (approved):

PR #5326, opened 2022-07-29
Internalise `FrameBuffer` and expose via `IFrameBuffer`

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/5324

- Adds the `IFrameBuffer` interface.
- Internalises `FrameBuffer` and adds a method to create frame buffers via `IRenderer`.
- Changes parameters of frame buffer creation to avoid ES30 references.

# vNext

## `FrameBuffer` is no longer exposed

Frame buffers may be created via the `IRenderer` and stored as an `IFrameBuffer`. Be aware that the parameter types have also changed!

```diff
class MyDrawNode : DrawNode
{
-   private readonly FrameBuffer myFrameBuffer = new FrameBuffer(new[] { RenderbufferInternalFormat.DepthComponent16 }, All.Nearest);
+   private IFrameBuffer myFrameBuffer;

    public override void Draw(IRenderer renderer)
    {
        base.Draw(renderer);

+       myFrameBuffer ??= renderer.CreateFrameBuffer(new[] { RenderBufferFormat.D16 }, TextureFilteringMode.Nearest);
    }

    protected override void Dispose(bool disposing)
    {
        base.Dispose(disposing);
        myFrameBuffer?.Dispose();
    }
}
```
Only a 16-bit depth format is supported for the time being, but more formats can be provided upon request.


osu.Framework/Graphics/BufferedDrawNode.cs
```diff
@@ -85,6 +85,8 @@ public override void ApplyState()
 
         public sealed override void Draw(IRenderer renderer)
         {
+            SharedData.Initialise(renderer);
+
```
frenzibyte, line 88: Feels misleading that this is calling `Initialise` every draw pass as if it's reinitialising but internally it early-returns if it's initialised. Would at least expose an `IsInitialised` bool and check here, and replace the early-guard with an assertion.

osu.Framework/Graphics/Rendering/IFrameBuffer.cs
```diff
@@ -0,0 +1,33 @@
...
+using osuTK;
+
+namespace osu.Framework.Graphics.Rendering
+{
+    public interface IFrameBuffer : IDisposable
+    {
+        /// <summary>
+        /// The framebuffer's backing texture.
+        /// </summary>
+        Texture Texture { get; }
```
frenzibyte, line 15: Is it safe to expose this as non-null? `FrameBuffer` will only create it when bound for the first time, and CI is yelling on a use-case that's guarding against null `Texture`:

CleanShot 2022-07-30 at 17 48 24@2x

smoogipoo: Hmm... I'll rejig this to make it safe. There's no reason `FrameBuffer` couldn't initialise immediately.

osu.Framework/Graphics/BufferedDrawNode.cs
smoogipoo: Sure. I've added an `IsInitialised`.

osu.Framework/Graphics/DrawNode.cs
```diff
@@ -274,7 +274,7 @@ protected virtual void DrawOpaqueInterior(IRenderer renderer)
             RectangleF textureRect = new RectangleF(0, frameBuffer.Texture.Height, frameBuffer.Texture.Width, -frameBuffer.Texture.Height);
 
             if (frameBuffer.Texture.TextureGL.Bind())
-                DrawQuad(frameBuffer.Texture, vertexQuad, drawColour, textureRect, vertexAction, inflationPercentage, blendRangeOverride);
+                DrawQuad(frameBuffer.Texture.TextureGL, vertexQuad, drawColour, textureRect, vertexAction, inflationPercentage, blendRangeOverride);
```
frenzibyte, line 277: I'm a bit confused by this change, does the `TextureBounds` method mess up with the dimensions given to the `TextureGLSingle.DrawQuad` call? Otherwise why would using `Texture` flip the texture.

smoogipoo: It does, yes. I actually don't know why, but I'll investigate. It doesn't happen on the main `irenderer` branch in which you can't access `TextureGL` anymore.

smoogipoo: Right, so it happens because `Texture.DrawQuad()` overrides `textureCoords` before passing to `TextureGL`.

https://github.com/ppy/osu-framework/blob/ef44cab294ec8bde310f75fd6a7986b0e8a894ab/osu.Framework/Graphics/Textures/Texture.cs#L180-L186

Which just so happens to invert the invert. I could fix this by also providing the `textureRect` here into the `DrawQuad()` call, but this is all going to go away fairly soon anyway...

frenzibyte: I see, let's keep it matching master until it's all away, yeah.

frenzibyte (dismissed):

peppy (approved):

peppy:
Applied a bit of conflict resolution, but I don't *think* I broke anything. Changes look sane.

frenzibyte (approved):

PR #5330, opened 2022-08-02
Internalise `TextureGL` and rewrite `Texture` to take its role

smoogipoo:
This PR is very big and nigh impossible to split out due to the amount of dependencies between `TextureGL` and `Texture`. Hopefully the diff should be pretty easy to go through, and I've done my best to separate the process via commits.

- `TextureGL*` classes merged into one.
- Opacity calculation and subtexturing has been moved to `Texture`. 
- `Texture` no longer exposes the underlying native texture.
- `Texture`s must be created through the `IRenderer`.

The rest should be explainable via the breaking changes below:

# vNext

## `WrapMode` and `Opacity` enums have been re-namespaced

Their new location is `osu.Framework.Graphics.Textures`.

## `Texture.WhitePixel` has been moved to `IRenderer`

It is no longer provided as a static member. Instead, [resolve](/ppy/osu-framework/wiki/Dependency-Injection) an `IRenderer` and access `IRenderer.WhitePixel`.

```diff
public class MyDrawable : Drawable
{
+   [BackgroundDependencyLoader]
+   private void load(IRenderer renderer)
+   {
+       texture ??= renderer.WhitePixel;
+   }

    private Texture texture;

    public Texture Texture
    {
-       get => texture ?? Texture.WhitePixel;
+       get => texture;
        set => texture = value;
    }
}
```

## `TextureStore`, `FontStore` and `LargeTextureStore` construction parameters have changed

- Requires an `IRenderer` parameter (via dependency injection).
- Filter mode parameter changed from `All` to `TextureFilteringMode`.

One common use case is to provide a new texture store from inside a derived `Game`, for which the following change is required:

```diff
public class TestGame : osu.Framework.Game
{
    private DependencyContainer dependencies;

    protected override IReadOnlyDependencyContainer CreateChildDependencies(IReadOnlyDependencyContainer parent) =>
        dependencies = new DependencyContainer(base.CreateChildDependencies(parent));

    [BackgroundDependencyLoader]
    private void load()
    {
-       var largeStore = new LargeTextureStore(Host.CreateTextureLoaderStore(new NamespacedResourceStore<byte[]>(Resources, @"Textures")), All.Nearest);
+       var largeStore = new LargeTextureStore(Host.Renderer, Host.CreateTextureLoaderStore(new NamespacedResourceStore<byte[]>(Resources, @"Textures")), TextureFilteringMode.Nearest);
        largeStore.AddTextureSource(Host.CreateTextureLoaderStore(new OnlineStore()));
        dependencies.Cache(largeStore);
    }
}
```

## `Texture` cannot be used to create textures

`Texture` must now be created via `IRenderer`.

```diff
public class MyDrawable : Drawable
{
-   private readonly Texture texture;

-   public MyDrawable()
-   {
-       texture = new Texture(100, 100);
-       texture.SetData(...);
-   }
+   private Texture texture;

+   [BackgroundDependencyLoader]
+   private void load(IRenderer renderer)
+   {
+       texture = renderer.CreateTexture(100, 100);
+       texture.SetData(...);
+   }
}
```

`DummyRenderer` may be used for cases where textures need to be created and neither an `IRenderer` nor `GameHost` is accessible:

```diff
[TestFixture]
public class MyTestFixture
{
    public void CreateATexture()
    {
-       Texture texture = new Texture(1, 1);
+       Texture texture = new DummyRenderer().CreateTexture(1, 1);
    }
}
```

## Texture drawing methods moved from `DrawNode` to `IRenderer` extension methods

Appearing alongside `IRenderer` is the new `RendererExtensions` class providing extension helper methods for common drawing procedures.

An example of the type of change required `DrawNode`:
```diff
public class MyDrawNode : DrawNode
{
    public override void Draw(IRenderer renderer)
    {
        base.Draw(renderer);

-       DrawQuad(renderer.WhitePixel, ...);
+       renderer.DrawQuad(renderer.WhitePixel, ...);
    }
}
```

The following methods have been moved to this class:
```
DrawTriangle()
DrawQuad()
DrawClipped()
DrawFrameBuffer()
```

## `TextureGL` is no longer accessible

Properties such as `TextureGL.BypassTextureUploadQueueing` have been moved to `Texture` itself, and `Texture` can be used for all drawing procedures.

During rendering, textures may now be bound via an integer sampling unit.

```diff
public class MyDrawable : Drawable
{
    private Texture texture;

    [BackgroundDependencyLoader]
    private void load(IRenderer renderer)
    {
        texture = renderer.CreateTexture(100, 100);
        texture.BypassUploadQueueing = true;
        texture.SetData(...);
    }
}

public class MyDrawNode : DrawNode
{
    private Texture texture;

    public override void Draw(IRenderer renderer)
    {
        base.Draw(renderer);

-       texture.TextureGL.Bind();
-       texture.TextureGL.Bind(TextureUnit.Texture1);
+       texture.Bind();
+       texture.Bind(1);

        // Note: The texture binding above isn't required in either case for the call below.
-       DrawQuad(texture.TextureGL, ...);
+       renderer.DrawQuad(texture, ...);
    }
}
```


osu.Framework.Tests/Graphics/TextureAtlasTest.cs
```diff
@@ -81,13 +82,34 @@ public void TestAtlasSecondRowAddRespectsWhitePixelSize()
...
+        {
+            const int atlas_size = 1024;
+
+            var atlas = new TextureAtlas(new DummyRenderer(), atlas_size, atlas_size);
+
+            // Direct atlas texture.
+            var tex1 = atlas.Add(1, 1).AsNonNull();
+
+            // TextureRegion.
+            var tex2 = tex1.Crop(new RectangleF(0, 0, 1, 1));
```
peppy, line 105: Would it be worth adding test coverage of the resultant texture sizes being correct for these?

osu.Framework/Graphics/OpenGL/Buffers/FrameBuffer.cs
```diff
@@ -137,13 +140,13 @@ public FrameBufferTexture(All filteringMode = All.Linear)
             public override int Width
             {
                 get => base.Width;
-                set => base.Width = Math.Clamp(value, 1, GLWrapper.MaxTextureSize);
+                set => base.Width = renderer == null ? value : Math.Clamp(value, 1, GLWrapper.MaxTextureSize);
```
peppy, line 143: Can the renderer actually be passed in as null? Had to double-take on this because NRT isn't applied on this file (might be a good time to apply that, if only here).

osu.Framework/Graphics/Performance/FrameStatisticsDisplay.cs
```diff
@@ -257,7 +257,12 @@ private void load()
 
             addArea(null, null, HEIGHT, amount_count_steps, columnUpload);
 
-            counterBarBackground?.Texture.SetData(columnUpload);
```
peppy, line 260: Not an issue, but I'm really curious where this texture came from in the original code...

osu.Framework/Graphics/Rendering/INativeTexture.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// <summary>
+        /// Whether the latest data has been uploaded.
+        /// </summary>
+        bool UploadComplete { get; }
+
+        /// <summary>
+        /// Whether the texture is currently queued for upload.
+        /// </summary>
+        bool IsQueuedForUpload { get; set; }
```
peppy, line 57: Is this required to have a setter?

osu.Framework/Graphics/Rendering/INativeTexture.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <returns>Whether the texture was successfully bound.</returns>
+        bool Bind(int unit, WrapMode wrapModeS, WrapMode wrapModeT);
+
+        /// <summary>
+        /// The size of this texture in bytes.
+        /// </summary>
+        int GetByteSize();
+
+        /// <summary>
+        /// The total amount of times this <see cref="INativeTexture"/> was bound.
```
peppy, line 91: Might help to make it clear that this is not a per-frame counter.

```suggestion
        /// The total amount of times this <see cref="INativeTexture"/> was bound (ever).
```

osu.Framework/Graphics/Textures/DisposableTexture.cs
```diff
@@ -1,32 +1,28 @@
...
+            : base(parent)
         {
+            this.parent = parent;
         }
 
         protected override void Dispose(bool isDisposing)
         {
             base.Dispose(isDisposing);
-            TextureGL.Dispose();
+            NativeTexture.Dispose();
```
peppy, line 24: I found this a bit awkward to read. One would assume that a `Texture` would be responsible for cleaning up the `NativeTexture`, but that's not the case (which is fine but on fresh eyes it may look like a wtf).

I think leaving xmldoc or a comment on the `NativeTexture` property in `Texture` explaining why the lifetime/disposal of this is not managed by the `Texture` itself would help.

osu.Framework/Graphics/Textures/TextureStore.cs
```diff
@@ -302,8 +302,8 @@ protected void Purge(Texture texture)
                 {
                     // we are doing this locally as right now, Textures don't dispose the underlying texture (leaving it to GC finalizers).
                     // in the case of a purge operation we are pretty sure this is the intended behaviour.
-                    tex?.TextureGL?.Dispose();
-                    tex?.Dispose();
+                    if (tex != null)
+                        new DisposableTexture(tex).Dispose();
```
peppy, line 306: Was this change just to keep everything going through the one disposal path? I think this cold also be written as

```csharp
tex.NativeTexture.Dispose();
tex.Dispose();
```

similar to the old code and still work correctly.

osu.Framework/Graphics/Textures/TextureRegion.cs
```diff
@@ -0,0 +1,82 @@
...
+            {
+                var adjustedBounds = upload.Bounds;
+
+                adjustedBounds.X += bounds.X;
+                adjustedBounds.Y += bounds.Y;
+
+                upload.Bounds = adjustedBounds;
+            }
+
+            // Todo: Not sure if this is correct...
```
peppy, line 56: Are you still unsure? If so, of which part?

osu.Framework/Graphics/Textures/TextureAtlas_InternalAtlasTexture.cs
```diff
@@ -0,0 +1,233 @@
...
+using SixLabors.ImageSharp.PixelFormats;
+
+namespace osu.Framework.Graphics.Textures
+{
+    public partial class TextureAtlas
+    {
+        /// <summary>
+        /// A texture which is acting as the backing for an atlas.
+        /// </summary>
+        private class InternalAtlasTexture : Texture
```
peppy, line 17: `BackingTexture` or `AtlasBackingTexture`?

osu.Framework/Graphics/Textures/TextureAtlas.cs
```diff
@@ -87,17 +87,19 @@ public void Reset()
...
+                return null;
 
             lock (textureRetrievalLock)
             {
                 Vector2I position = findPosition(width, height);
                 RectangleI bounds = new RectangleI(position.X, position.Y, width, height);
                 subTextureBounds.Add(bounds);
 
-                return new TextureGLSub(bounds, AtlasTexture, wrapModeS, wrapModeT);
+                // atlasTexture can't be null after findPosition().
```
peppy, line 101: `Debug.Assert` instead in that case, after the `findPosition` method call preferrably.

osu.Framework/Graphics/Textures/TextureAtlas.cs
```diff
@@ -39,19 +38,20 @@ internal TextureWhitePixel WhitePixel
         {
             get
             {
-                if (AtlasTexture == null)
+                if (atlasTexture == null)
                     Reset();
 
-                return new TextureWhitePixel(new TextureGLAtlasWhite(AtlasTexture));
+                return new TextureWhitePixel(atlasTexture.AsNonNull());
```
peppy, line 44: I'd probably also use a `Debug.Assert` here too.

osu.Framework/Graphics/Textures/Texture.cs
```diff
@@ -1,66 +1,81 @@
...
-            new TextureAtlas(TextureAtlas.WHITE_PIXEL_SIZE + TextureAtlas.PADDING, TextureAtlas.WHITE_PIXEL_SIZE + TextureAtlas.PADDING, true).WhitePixel);
-
-        public static Texture WhitePixel => white_pixel.Value;
-
-        public virtual TextureGL TextureGL { get; }
+        internal virtual INativeTexture NativeTexture { get; }
 
-        public string Filename;
-        public string AssetName;
+        public string Filename = string.Empty;
```
peppy, line 18: `Filename` is not used in framework or osu!, and is pretty pointless considering `AssetName` below. Might as well remove it as part of this PR?

peppy (changes requested):

osu.Framework/Graphics/OpenGL/Buffers/FrameBuffer.cs
smoogipoo: It can only during construction time, since `Width`/`Height` are set from the base ctor. NRT won't help with this :/

osu.Framework/Graphics/Performance/FrameStatisticsDisplay.cs
smoogipoo: In the constructor https://github.com/ppy/osu-framework/pull/5330/files#diff-3a107db5547cd47903bd20ce803297d8ffe8e65b30c595f3ea1e66949c2e41f2L155 . Unless you mean original code as in the code that was written 6 years ago, then idk.

peppy: Right, I see it now.

osu.Framework/Graphics/OpenGL/Buffers/FrameBuffer.cs
peppy: And this is actually accessed in such a state? Feels a bit haphazard TBH.. if we can avoid it I would recommend doing so.

osu.Framework/Graphics/Textures/DisposableTexture.cs
smoogipoo: The doc is in `Texture.Dispose()` itself. Best would be to add a doc _here_ if anything.

osu.Framework/Graphics/Textures/TextureStore.cs
smoogipoo: Yes, it could be written that way, but I don't want to add accesses to `NativeTexture`.

There's only one reference to `NativeTexture` outside of `Texture` & co, and that is `OpenGLRenderer`, and I would like to remove that also.

osu.Framework/Graphics/Textures/TextureRegion.cs
smoogipoo: I wasn't sure about the opacity update, but I've added it back in here. The comment is from early days.

For reference, see: https://github.com/ppy/osu-framework/blob/60f5b5871786a3520f59f62157e0112ed730e37c/osu.Framework/Graphics/OpenGL/Textures/TextureGLSub.cs#L104-L127

osu.Framework/Graphics/Textures/DisposableTexture.cs
smoogipoo: I've added a remark on the xmldoc.

osu.Framework/Graphics/Rendering/INativeTexture.cs
smoogipoo: No, technically shouldn't exist at all (specific to GL implementation), but can be cleaned up later. Removed setter for now.

osu.Framework/Graphics/OpenGL/Buffers/FrameBuffer.cs
smoogipoo: Here's a few options:
1. Not have `TextureGL` take `width`/`height` ctor parameters, and instead do something like `new TextureGL(...) { Width = ... , Height = ... }`
2. Have `INativeTexture` expose `Width`/`Height` as get-only and `SetWidth()`/`SetHeight()` methods that `Texture` calls into.

Pick your poison, or.. any better suggestions?

osu.Framework/Graphics/Rendering/INativeTexture.cs
peppy: I did notice that it's not being accessed even via the getter, so it could potentially be removed right now?

smoogipoo: Sure, done.

osu.Framework.Tests/Graphics/TextureAtlasTest.cs
smoogipoo: The size of textures retrieved from atlases being correct is checked a few tests up above, but I've added a new test for `TextureRegion` specifically since that's a path that isn't tested and very important to drawing as a whole.

osu.Framework/Graphics/OpenGL/Buffers/FrameBuffer.cs
peppy: I'm fine with it with your change to remove the null check.

smoogipoo: .... I made that change and promptly forgot I made it. Yes the nullcheck isn't required anymore since `Renderer` is exposed as protected from `TextureGL` :sweat_smile:

peppy (approved):

osu.Framework/Graphics/Textures/Texture.cs
```diff
@@ -109,88 +130,203 @@ public static Texture FromStream(Stream stream, TextureAtlas atlas = null)
...
-        /// <param name="textureRect">The texture rectangle in texture space.</param>
-        /// <param name="vertexAction">An action that adds vertices to a <see cref="VertexBatch{T}"/>.</param>
-        /// <param name="inflationPercentage">The percentage amount that <paramref name="textureRect"/> should be inflated.</param>
-        /// <param name="textureCoords">The texture coordinates of the triangle's vertices (translated from the corresponding quad's rectangle).</param>
-        internal void DrawTriangle(Triangle vertexTriangle, ColourInfo drawColour, RectangleF? textureRect = null, Action<TexturedVertex2D> vertexAction = null,
-                                   Vector2? inflationPercentage = null, RectangleF? textureCoords = null)
+        /// <param name="area">A display-space area of this texture to compute the UV coordinates of.<br/>
+        /// The full area is used if not provided: (0, 0, <see cref="DisplayWidth"/>, <see cref="DisplayHeight"/>).</param>
+        /// <returns>The UV coordinates of <paramref name="area"/> in this texture.</returns>
+        public virtual RectangleF GetTextureRect(RectangleF? area = null)
```
smoogipoo, line 261: One thing I forgot to mention is that I had a lot of trouble getting this `GetTextureRect()` method to work due to working in two "coordinate spaces".

In this PR, I've tried to use the terminology "texture-space" and "display-space" to describe which coordinate space the user or reviewer of code should be thinking of.
- "texture-space" is in coordinates `(0, 0, Width, Height)`
- "display-space" is in coordinates `(0, 0, DisplayWidth, DisplayHeight)`

This method here is one example, as well as:

- The XMLDoc on `TextureRegion`: https://github.com/ppy/osu-framework/pull/5330/files#diff-296537cb3a9b525374da9be847ea69b5532c71516de14c76ab09a73e1b373df0R17-R25
- The implementation of `TextureRegion`: https://github.com/ppy/osu-framework/pull/5330/files#diff-296537cb3a9b525374da9be847ea69b5532c71516de14c76ab09a73e1b373df0R60-R78

frenzibyte (commented):
Looks good on a post-merge read.

PR #5334, opened 2022-08-05
Safely restore TPL capacity after test failures

smoogipoo:
Hoping this will prevent test timeouts such as https://github.com/ppy/osu-framework/runs/7686246007?check_suite_focus=true


smoogipoo:
Seems to have worked - no more execution timeouts. HttpBin is just down atm.

peppy (approved):

PR #5337, opened 2022-08-07
Remove nullable disable annotation in the clock namespace

andy840119:
Remove nullable disable annotate in the clock namespace.

### DIFF
--- a/osu.Framework/Timing/FramedClock.cs
+++ b/osu.Framework/Timing/FramedClock.cs
@@ -3,6 +3,7 @@
 
 using osu.Framework.Extensions.TypeExtensions;
 using System;
+using System.Diagnostics;
 
 namespace osu.Framework.Timing
 {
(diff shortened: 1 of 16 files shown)


osu.Framework/Timing/FramedClock.cs
```diff
@@ -51,8 +52,10 @@ public FramedClock(IClock? source = null, bool processSource = true)
 
         private const int fps_calculation_interval = 250;
 
-        public void ChangeSource(IClock source)
+        public void ChangeSource(IClock? source)
```
peppy, line 55: What's going on here?

andy840119: Because `ChangeSource()` should accept null.
See:
https://github.com/ppy/osu-framework/blob/ba1385330cc501f34937e08257e586c84e35d772/osu.Framework/Timing/InterpolatingFramedClock.cs#L38-L41

peppy: It can't accept null then assert otherwise though.

andy840119: OK.
I think it should be better following how `DecoupleableInterpolatingFramedClock` did.
https://github.com/ppy/osu-framework/blob/ba1385330cc501f34937e08257e586c84e35d772/osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs#L120-L128

peppy (approved):

PR #5338, opened 2022-08-07
Remove nullable disable annotation in the threading namespace.

andy840119:
Remove nullable disable annotate in the threading namespace.

### DIFF
--- a/osu.Framework/Graphics/Performance/FrameStatisticsDisplay.cs
+++ b/osu.Framework/Graphics/Performance/FrameStatisticsDisplay.cs
@@ -19,7 +19,6 @@ using System.Buffers;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-using osu.Framework.Extensions.ObjectExtensions;
 using osu.Framework.Graphics.Rendering;
 using osu.Framework.Input.Events;
 using osuTK;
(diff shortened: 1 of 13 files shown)


osu.Framework/Threading/GameThread.cs
```diff
@@ -492,7 +488,7 @@ private void setExitState(GameThreadState exitState)
                 {
                     case GameThreadState.Exited:
                         Monitor?.Dispose();
-                        initializedEvent?.Dispose();
+                        initializedEvent.Dispose();
```
peppy, line 491: This is not quite safe, but not sure if it matters in practice (if there's an issue during construction it will potentially be null).

osu.Framework/Graphics/Performance/FrameStatisticsDisplay.cs
```diff
@@ -110,7 +111,7 @@ public FrameStatisticsMode State
         public FrameStatisticsDisplay(GameThread thread, ArrayPool<Rgba32> uploadPool)
         {
             Name = thread.Name;
-            monitor = thread.Monitor;
+            monitor = thread.Monitor.AsNonNull();
```
peppy, line 114: Is there a reason for using `AsNonNull` instead of `Debug.Assert` here?

andy840119: Fixed.

osu.Framework/Threading/GameThread.cs
andy840119: Not really sure but technically it should be only run once?

peppy: For safety you should probably follow other examples, and use `IsNotNull()`.

andy840119: Fixed 🙆🏼‍♂️

smoogipoo (approved):

PR #5344, opened 2022-08-10
Fix crash when minimising while fullscreen on windows

peppy:
SDL behaviour looks to have changed when calling `SDL_GetWindowSize` while minimised from a fullscreen state. This fixes crashes due to div-by-zero propagating into the draw hierarchy.

Haven't investigated in detail why this happened, but the fix seems solid enough to get out as a hotfix.


smoogipoo (approved):
As mentioned, this is probably a bug in SDL, but fine to remain permanently.

PR #5348, opened 2022-08-11
Fix `TabControl` throwing if current value is not found in tabs

peppy:


peppy:
^ bit of an unfortunate one. Guarding against icon creation failure, only to crash moments later if the user had the failing custom ruleset as their selected ruleset.

smoogipoo (approved):
I _think_ this is fine

Issue #3510, opened 2020-05-03
Arbitrary shapes


abraker95:
Ability to create all sorts of irregular polygons and have them transform. Support for changing position of vertices and adding/removing vertices from a shape.

This would be used to create a dodge the beat kind of gamemode.

peppy:
I'm not sure I understand your goal here. Are you looking for collision logic?

abraker95:
Ability to make a shape, add vertices to a shape, set events for vertices to move, etc. I see support for rectangles, circles, and triangles of various widths and lengths, but I am not aware if there is anything that allows this. The gif below demonstrates the flexibility in shape manipulation I am trying to describe:

![](https://i.imgur.com/cQTqX0z.gif)

peppy:
Are you willing to implement this in the framework, or looking for someone to do it for you? While having such functionality looks nice to have for some use cases, as we won't be using it for osu! it's highly unlikely to be given any priority for the foreseeable future (ie. the next year at least).

abraker95:
I have been trying, but I am not familiar enough with the framework to do this properly yet. I've been looking at `Box` and `Triangle` sprites and `WaveformGraph` to give me an ideas of how to use the framework make something like this.

PR #5351, opened 2022-08-11
Add stencil support for renderers

Flutterish:
This adds `StencilInfo` and appropriate methods to renderers. It also adds an 8-bit stencil buffer to the screen buffers (please doublecheck if all window types have it, I only added it to the SDL window as I have no way to check if the mobile ones or OsuTkWindow work).
This is a part of my work towards #3510 - the even-odd fill algorithm is very easy to implement with a stencil buffer.

### DIFF
--- a/osu.Framework/Graphics/Rendering/StencilInfo.cs
+++ b/osu.Framework/Graphics/Rendering/StencilInfo.cs
@@ -1,0 +1,68 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using osuTK.Graphics.ES30;
+
+namespace osu.Framework.Graphics.Rendering
+{
+    public struct StencilInfo : IEquatable<StencilInfo>
+    {
+        /// <summary>
+        /// Whether stencil testing should occur.
+        /// If this is false, no <see cref="StencilOp"/> will occur (use <see cref="StencilFunction.Always"/> instead).
+        /// </summary>
+        public readonly bool StencilTest;
+
+        /// <summary>
+        /// The stencil test function.
+        /// </summary>
+        public readonly StencilFunction TestFunction;
+
+        /// <summary>
+        /// The stencil test value to compare against.
+        /// </summary>
+        public readonly int TestValue;
+
+        /// <summary>
+        /// The stencil mask.
+        /// </summary>
+        public readonly int Mask;
+
+        /// <summary>
+        /// The operation to perform on the stencil buffer in case the stencil test failed.
+        /// </summary>
+        public readonly StencilOp StencilTestFailOperation;
+
+        /// <summary>
+        /// The operation to perform on the stencil buffer in case the depth test failed.
+        /// </summary>
+        public readonly StencilOp DepthTestFailOperation;
+
+        /// <summary>
+        /// The operation to perform on the stencil buffer in case the stencil test passed.
+        /// </summary>
+        public readonly StencilOp TestPassedOperation;
+
+        public StencilInfo(bool stencilTest = false, StencilFunction testFunction = StencilFunction.Always, int testValue = 1, int mask = 0xff,
+            StencilOp stencilFailed = StencilOp.Keep, StencilOp depthFailed = StencilOp.Keep, StencilOp passed = StencilOp.Replace)
+        {
+            StencilTest = stencilTest;
+            TestFunction = testFunction;
+            TestValue = testValue;
+            Mask = mask;
+            StencilTestFailOperation = stencilFailed;
+            DepthTestFailOperation = depthFailed;
+            TestPassedOperation = passed;
+        }
+
+        public bool Equals (StencilInfo other) =>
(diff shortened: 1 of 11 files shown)


osu.Framework/Graphics/Rendering/StencilInfo.cs
```diff
@@ -0,0 +1,73 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using osuTK.Graphics.ES30;
```
smoogipoo, line 5: Can't reference ES30. You need to make appropriate o!f enums.

smoogipoo (changes requested):
Rest looks okay

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -454,6 +454,7 @@ public SDL2DesktopWindow()
                 throw new InvalidOperationException($"Failed to initialise SDL: {SDL.SDL_GetError()}");
             }
 
+            SDL.SDL_GL_SetAttribute(SDL.SDL_GLattr.SDL_GL_STENCIL_SIZE, 8);
```
smoogipoo, line 457: This can't be in the window. It should be somewhere inside the `SDL2GraphicsBackend` (temporarily, because that class is going to go away). Perhaps it could fit in the constructor, or in a method like `InitialiseBeforeWindow()`.

As an aside... That entire class is kinda weird - `SDL2DesktopWindow` -> `IGraphicsBackend.Initialise()` -> `SDL2GraphicsBackend.Initialise()` -> `PassthroughGraphicsBackend.Initialise()` -> `SDL2GraphicsBackend.CreateContext()`. Too much back and forth.

smoogipoo (changes requested):

osu.Framework/Graphics/Rendering/StencilInfo.cs
Flutterish: I copied this from DepthInfo which also references ES30. I will make custom enums, but in that case you should also look into the existing structs which do this.

smoogipoo: Ouch, yeah that needs to change.

osu.Framework/Graphics/Rendering/StencilInfo.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// </summary>
+        public readonly StencilOp StencilTestFailOperation;
+
+        /// <summary>
+        /// The operation to perform on the stencil buffer in case the depth test failed.
+        /// </summary>
+        public readonly StencilOp DepthTestFailOperation;
+
+        /// <summary>
+        /// The operation to perform on the stencil buffer in case the stencil test passed.
```
smoogipoo, line 48:
```suggestion
        /// The operation to perform on the stencil buffer when both the depth and stencil tests pass.
```

osu.Framework/Graphics/Rendering/StencilInfo.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// </summary>
+        public readonly int Mask;
+
+        /// <summary>
+        /// The operation to perform on the stencil buffer in case the stencil test failed.
+        /// </summary>
+        public readonly StencilOp StencilTestFailOperation;
+
+        /// <summary>
+        /// The operation to perform on the stencil buffer in case the depth test failed.
```
smoogipoo, line 43:
```suggestion
        /// The operation to perform on the stencil buffer when the depth test fails.
```

osu.Framework/Graphics/Rendering/StencilInfo.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// </summary>
+        public readonly int TestValue;
+
+        /// <summary>
+        /// The stencil mask.
+        /// </summary>
+        public readonly int Mask;
+
+        /// <summary>
+        /// The operation to perform on the stencil buffer in case the stencil test failed.
```
smoogipoo, line 38:
```suggestion
        /// The operation to perform on the stencil buffer when the stencil test fails.
```

osu.Framework/Graphics/Rendering/StencilInfo.cs
smoogipoo: In https://github.com/ppy/osu-framework/pull/5352 I've added `DepthStencilFunction`, which you should be able to reuse here.

Flutterish:
Waiting for #5352 to be merged before I can use the `DepthStencilFunction`.

smoogipoo:
Looks good. Added a test scene to demonstrate usage + test it.

Flutterish:
Do you have a way to test whether mobile/ios "windows" (`OsuTkWindow`) have the stencil buffer attached? I suspect they might not but I don't think I have a way to test that myself.

smoogipoo:
I don't. As far as I know we're requesting the default context provided by the platforms.

smoogipoo (approved):

peppy (approved):

PR #5353, opened 2022-08-11
Add some missing rendering enums' values

Flutterish:
Adds a few more depth-stencil render buffer formats, based on Veldrid/D3D10+ compatibility.

### DIFF
--- a/osu.Framework/Graphics/OpenGL/OpenGLRenderer.cs
+++ b/osu.Framework/Graphics/OpenGL/OpenGLRenderer.cs
@@ -722,7 +722,7 @@ namespace osu.Framework.Graphics.OpenGL
             if (depthInfo.DepthTest)
             {
                 GL.Enable(EnableCap.DepthTest);
-                GL.DepthFunc(depthInfo.Function);
+                GL.DepthFunc(OpenGLUtils.ToDepthFunction(depthInfo.Function));
             }
             else
                 GL.Disable(EnableCap.DepthTest);
--- a/osu.Framework/Graphics/OpenGL/OpenGLRenderer.cs
+++ b/osu.Framework/Graphics/OpenGL/OpenGLRenderer.cs
@@ -846,6 +846,26 @@ namespace osu.Framework.Graphics.OpenGL
                             glFormats[i] = RenderbufferInternalFormat.DepthComponent16;
                             break;
 
+                        case RenderBufferFormat.D24:
+                            glFormats[i] = RenderbufferInternalFormat.DepthComponent24;
+                            break;
+
+                        case RenderBufferFormat.D32:
+                            glFormats[i] = RenderbufferInternalFormat.DepthComponent32f;
+                            break;
+
+                        case RenderBufferFormat.S8:
+                            glFormats[i] = RenderbufferInternalFormat.StencilIndex8;
+                            break;
+
+                        case RenderBufferFormat.D24S8:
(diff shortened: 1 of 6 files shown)


osu.Framework/Graphics/Rendering/PrimitiveTopology.cs
```diff
@@ -9,6 +9,7 @@ public enum PrimitiveTopology
         Lines,
         LineStrip,
         Triangles,
-        TriangleStrip
+        TriangleStrip,
+        TriangleFan
```
smoogipoo, line 13: Are you actually using this?

- [Veldrid doesn't support triangle fans.](https://github.com/mellinoe/veldrid/blob/master/src/Veldrid/PrimitiveTopology.cs)
- [Vulkan supports triangle fans](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPrimitiveTopology.html).
- [D3D10+ doesn't support triangle fans](https://docs.microsoft.com/en-au/windows/win32/direct3d9/triangle-fans?redirectedfrom=MSDN)
- [Metal doesn't support triangle fans](https://developer.apple.com/documentation/metal/mtlprimitivetype?language=objc)

osu.Framework/Graphics/OpenGL/OpenGLRenderer.cs
```diff
@@ -846,6 +846,26 @@ public IFrameBuffer CreateFrameBuffer(RenderBufferFormat[]? renderBufferFormats
...
+                            glFormats[i] = RenderbufferInternalFormat.StencilIndex8;
+                            break;
+
+                        case RenderBufferFormat.D24S8:
+                            glFormats[i] = RenderbufferInternalFormat.Depth24Stencil8;
+                            break;
+
+                        case RenderBufferFormat.D32S8:
+                            glFormats[i] = RenderbufferInternalFormat.Depth32fStencil8;
+                            break;
```
smoogipoo, line 867: Out of these, only
```
D32
D24S8
D32S8
```
Seem to be supported by [Veldrid](https://github.com/mellinoe/veldrid/blob/master/src/Veldrid/PixelFormat.cs) or [D3D11](https://docs.microsoft.com/en-us/windows/win32/api/d3d11/ns-d3d11-d3d11_depth_stencil_view_desc#remarks).

osu.Framework/Graphics/Rendering/PrimitiveTopology.cs
Flutterish: Well, this one in specific is useful for drawing arbitrary shapes defined by a curve at 3x smaller vertex count.
I suppose it would be possible to create a fallback batch which translates fan vertices to regular triangles?

smoogipoo: Strips and fans have the same vertex requirements and strips are better for performance.

osu.Framework/Graphics/OpenGL/OpenGLRenderer.cs
Flutterish: I added them for completeness, can remove the unsupported ones.

smoogipoo: Let’s stick to supported formats only.

smoogipoo (approved):

PR #5354, opened 2022-08-12
Add ability to push arbitrary projection matrix to `IRenderer`

Flutterish:
The old `PushOrtho` was moved to an extension method.
This change allows drawables to draw their vertices in transformed spaces, more specifically it is intended to be used in combination with the local 3x3 matrix to allow to upload vertices in local space rather than transforming them to screen space manually.
I have also used this to optimise the draw code of `Path` and `CircularProgress` which both heavily used transforming their local space vertices to screen space on the cpu.
This is also part of my work towards #3510 where the implementation benefits from this.


osu.Framework/Graphics/OpenGL/OpenGLRenderer.cs
```diff
@@ -59,7 +59,9 @@ internal class OpenGLRenderer : IRenderer
         public ref readonly MaskingInfo CurrentMaskingInfo => ref currentMaskingInfo;
 
         public RectangleI Viewport { get; private set; }
-        public RectangleF Ortho { get; private set; }
+
+        // this replaces the old ortho - we need its size to properly set viewports.
+        private Vector2 matrixScale => new Vector2(ProjectionMatrix.Row0.X / 2, -ProjectionMatrix.Row1.Y / 2);
```
smoogipoo, line 64: Is this going to be used elsewhere in the future? Is there a reason for not inlining this if not? Seems it's only used in one place.

smoogipoo:
I've added back the `PopOrtho()` to save on breaking change + I think it looks better if push and pop methods are named the same.

In the future we should also consider using `using` for this. Maybe something like `using (renderer.BeginProjectionMatrix(...))`.

osu.Framework/Graphics/OpenGL/OpenGLRenderer.cs
Flutterish: I left it in while removing the `Ortho` property, it's fine to inline.

osu.Framework/Graphics/Rendering/RendererExtensions.cs
```diff
@@ -235,5 +235,70 @@ public static class RendererExtensions
...
+
+        /// <summary>
+        /// Applies a new projection matrix so that all drawn vertices are transformed by <paramref name="matrix"/>. This also affects masking.
+        /// </summary>
+        /// <remarks>
+        /// After usage, restore the last state with <see cref="PopLocalMatrix"/>.
+        /// </remarks>
+        /// <param name="renderer">The renderer.</param>
+        /// <param name="matrix">The matrix.</param>
+        public static void PushLocalMatrix(this IRenderer renderer, Matrix4 matrix)
```
smoogipoo, line 266: This is more akin to a "model" matrix in traditional MVP terminology. How are you doing cameras in osu!xr, if at all?

Flutterish: oxr has an entirely separate rendering pipeline custom fit for 3D (and customizable per scene if needed), and it currently only interacts with the 2D IRenderer for textures (which I would like change, but it's going to require a few more additions). Generally there is one global projection matrix and one model matrix, but it is allowed to vary based on the specific drawable and shaders used.

smoogipoo: Hmm... I think this is fine as it is. I'm coming at this from the perspective of eventually implementing 3D into o!f itself, not as a separate rendering pipeline. At that time, I think this will suffice for both the model and the view matrices.

smoogipoo (approved):

osu.Framework/Graphics/Rendering/RendererExtensions.cs
Flutterish: Not related to matrices, but the "custom pipeline" part - a thing to consider about 3D is that its drawables need to be grouped, sorted (example: transparency sorting), batched and various other things which greatly benefit from flattening its draw hierarchy at the scene level (not to mention iteration speed - I would like to say, oxr can do 10k rotating cubes in [8u/3ms](https://media.discordapp.net/attachments/820300293060165632/983000083295506453/unknown.png?width=1122&height=634) and stationary in [2u/3ms](https://media.discordapp.net/attachments/820300293060165632/983006964512616448/unknown.png?width=1120&height=639) which is [several times faster than unity](https://media.discordapp.net/attachments/820300293060165632/983008502194790420/unknown.png?width=229&height=38). dunno how good of a test that is though). I think if we are going to implement 3D into o!f (which I would like to help with as I already have some full-stack experience with it), the 3D drawables and draw nodes need to be fit to allow for this.

PR #5356, opened 2022-08-12
Use local httpbin instance on CI runs

smoogipoo:
We had a lot of timeouts using hosted httpbin back on appveyor so we eventually decided to run a local httpbin instance. Since moving to GitHub actions, it's been mostly solid using the hosted instance, but over the past couple of days it's started failing again.

This uses [go-httpbin](https://github.com/mccutchen/go-httpbin), which is not quite 1-1 in terms of responses, hence why there's conversion functions.


osu.Framework.Tests/IO/TestWebRequest.cs
```diff
@@ -346,7 +345,7 @@ public void TestStartAfterAbort([Values(true, false)] bool async)
         [Test, Retry(5)]
         public void TestRestartAfterAbort([Values(true, false)] bool async)
         {
-            var request = new JsonWebRequest<HttpBinGetResponse>($"{default_protocol}://{host}/get")
+            var request = new JsonWebRequest<HttpBinGetResponse>($"{default_protocol}://{host}/delay/10")
```
peppy, line 348: Is this 10ms delay?

smoogipoo: 10 seconds. The request is aborted immediately afterwards but I'll reduce to 5 seconds if that makes you feel better. Not willing to go much/any lower than that.

peppy: Is the reasoning correct as for why it passed until now?

smoogipoo: Correct, it was relying on the test running faster than the response.

peppy (approved):

PR #5358, opened 2022-08-12
Add an "arbitrary shape" drawable

### LINKED ISSUE #3510: Arbitrary shapes

Flutterish:
Resolves #3510
Allows you to draw a vector-graphics-ish shape defined by a closed curve (of vertices).
There are 3 fill rules to select from: [NonZero](https://en.wikipedia.org/wiki/Nonzero-rule) (default for vector graphics), [EvenOdd](https://en.wikipedia.org/wiki/Even–odd_rule) and Fan (for convex shapes)
image
image

The implementation can also check for "insidedness" of a point in the `ReceivePositionalInputAt` method.
Please let me know if the `FillRule` docs are a bit confusing - I don't really know how to word them better so we will need to figure something out together.

### DIFF
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneArbitraryShape.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneArbitraryShape.cs
@@ -24,7 +24,8 @@ namespace osu.Framework.Tests.Visual.Drawables
         private List<SpriteIcon> directionGuides = new List<SpriteIcon>();
         private bool isShapeValid;
         private Box boundngBox;
-        public TestSceneArbitraryShape ()
+
+        public TestSceneArbitraryShape()
         {
             Add(new CircularContainer()
             {
@@ -62,7 +63,7 @@ namespace osu.Framework.Tests.Visual.Drawables
             shape.Texture = store.Get(@"sample-texture");
         }
 
-        protected override bool OnClick (ClickEvent e)
+        protected override bool OnClick(ClickEvent e)
         {
             DragHandle handle = new DragHandle() { Position = Content.ToLocalSpace(e.ScreenSpaceMousePosition) };
             handle.Dragged += onHandleDragged;
@@ -72,13 +73,13 @@ namespace osu.Framework.Tests.Visual.Drawables
             return true;
         }
 
-        private void onHandleDragged (DragHandle handle, DragEvent e)
+        private void onHandleDragged(DragHandle handle, DragEvent e)
         {
             handle.Position = Content.ToLocalSpace(e.ScreenSpaceMousePosition);
             isShapeValid = false;
         }
 
-        protected override void Update ()
+        protected override void Update()
         {
             base.Update();
             if (!isShapeValid)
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/Shapes/ArbitraryShape.cs
```diff
@@ -0,0 +1,324 @@
...
+using osu.Framework.Graphics.Shaders;
+using osu.Framework.Graphics.Textures;
+using osuTK;
+
+namespace osu.Framework.Graphics.Shapes
+{
+    /// <summary>
+    /// An arbitrary shape defined by a set of vertices forming a closed curve.
+    /// </summary>
+    public partial class ArbitraryShape : Drawable
```
frenzibyte, line 22: I don't know about using the term "arbitrary shape" for a general drawable class like this. Why not just `Shape`?

Flutterish: Ah, yes, I forgot to ask for a name change in the pr, as I didn't know how to properly call it. I'm not sure just "shape" is enough though, I would like something to imply the fact that its made up from vertrices (as opposed to say, an actual curve)

frenzibyte: Personally I would imagine a class with the name `Shape` to require me to input the vertices for constructing it, but will wait for more opinions.

Flutterish: Hm, I suppose if we call the discrete (vertex) variant a `Shape` and a possible future continuous (curve) variant a `VectorShape` there would be no conflict.

osu.Framework.Tests/Visual/Drawables/TestSceneArbitraryShape.cs
```diff
@@ -0,0 +1,167 @@
...
+        public TestSceneArbitraryShape()
+        {
+            Add(new CircularContainer()
+            {
+                Masking = true,
+                RelativeSizeAxes = Axes.Both,
+                Children = new Drawable[]
+                {
+                    new Box() { Colour = Colour4.Gray, RelativeSizeAxes = Axes.Both },
+                    boundngBox = new Box()
```
peppy, line 37:
```suggestion
                    boundingBox = new Box()
```

osu.Framework/Graphics/Shapes/ArbitraryShape.cs
jai-x: I would personally go with `Polygon` for the class name, as it implies to me a class that draws a shape with `n` arbitrary number of vertices in contrast with the existing classes of `Circle`, `Triangle`, and `Box` which have a set amount.

frenzibyte: That's also another name I had but it conflicts with the underlying data used for drawing triangles/quads (see `IPolygon`/`IConvexPolygon`)

Flutterish: What about `DrawablePolygon` or `PolygonShape`, rather than just `Polygon`?

frenzibyte: No idea, I would rather rename `IPolygon` to `IPolygonData` or otherwise rather than naming the drawable variant differently while it reflects a polygon.

Flutterish: Looking at Graphics.Primitives and Graphics.Shapes, a similar conflict exist between `Primitives.Triangle` and `Shapes.Triangle`. If that namespace difference is acceptable, why would `Shapes.Polygon` and `Primitives.I(Convex)Polygon` not be?

PR #5359, opened 2022-08-15
Attempt to fix NRT on Mono

smoogipoo:
I haven't tested this.

Can someone set up with a mobile deployment pipeline check that:
- These tests run _with_ failures on mobile on master, and _without_ failures on this branch: https://github.com/ppy/osu-framework/blob/master/osu.Framework.Tests/Extensions/TestIsNullableTypeExtensions.cs You'll need to hoist this test up into a visual test and run on the actual device.
- osu! runs fine after reverting this PR and importing a beatmap: https://github.com/ppy/osu/pull/19677/files


smoogipoo:
Actually, this is not going to work, because this is exactly what `NullabilityInfoContext` does internally. Sigh.

PR #5362, opened 2022-08-16
Make position transfer more reliable when track is stopped

peppy:
This came up when attempting to use coupled mode in the editor (it was using decoupled until now).
When the track finished, the decoupled clock would be at a different `CurrentTime` to the source track (sometimes ahead, sometimes behind it) due to a final value transfer not occurring.

This would in turn cause the track to not loop correctly, as the editor was relying on checking `Current` vs `Length` to check whether playback has finished.

I know this is adding even more complexity to the existing clocks, but right now I'm just trying to get everything working. I'm plannning to probably do a full rewrite of interpolating/decoupling in the future.

### DIFF
--- a/osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs
+++ b/osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs
@@ -84,19 +84,26 @@ namespace osu.Framework.Timing
             // process the decoupled clock to update the current proposed time.
             decoupledClock.ProcessFrame();
 
-            // if the source clock is started as a result of becoming capable of handling the decoupled time, the proposed time may change to reflect the interpolated source time.
-            // however the interpolated source time that was calculated inside base.ProcessFrame() (above) did not consider the current (post-seek) time of the source.
-            // in all other cases the proposed time will match before and after clocks are started/stopped.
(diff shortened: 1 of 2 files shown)


osu.Framework/Timing/DecoupleableInterpolatingFramedClock.cs
```diff
@@ -84,19 +84,26 @@ public override void ProcessFrame()
...
+
+                        // if the source stops, ensure that we are immediately in sync with its time value.
+                        //
+                        // note that this *won't* apply when a Stop() call is made. in such a case, the interpolated value will
+                        // remain as current (as this is more expected behaviour – if we did a transfer there would be a jump, potentially
+                        // backwards.
+                        if (adjustableSource != null)
+                        {
+                            decoupledStopwatch.Seek(adjustableSource.CurrentTime);
+                            decoupledClock.ProcessFrame();
```
peppy, line 104: Alternative to this (and moving the `ProposedTime` fetches) is to replace this with a recursive call:

```suggestion
                            ProcessFrame();
                            return;
```

But I'm not sure if that reads better or not.

peppy:
I've confirmed that osu! passes all tests with this change, for whatever that's worth.

PR #5365, opened 2022-08-19
Sourcelink improvements

smoogipoo:
While checking to see why Sourcelink isn't working osu!-side, I came up with a few things.

1. Update the package from `1.0.0` to `1.1.1`.
2. Remove embedded PDBs. Embedding PDBs [is not recommended](https://github.com/dotnet/sourcelink#alternative-pdb-distribution) by Microsoft. The downside to this is that we lose local symbolicated debugging support (which is why it was working for me), but we don't ever really debug with a local package anyway so I don't think this will affect us.

Compare:
- https://www.nuget.org/packages/ppy.osu.Framework/2022.818.0
- https://www.nuget.org/packages/smoogipoo.osu.Framework/0.0.0

Notice how my package has a "Download symbols" link. I believe this is the source of the issue - either by embedding the PDBs, or these extra options, nuget wasn't getting the symbols correctly. Or it's possible that the AppVeyor config is busted, though the [last deploy](https://ci.appveyor.com/project/peppy/osu-framework-a4n7e/build/job/txmb0ix479robbd5) indicates that the symbols package was uploaded...

My package also works - you can step into and breakpoint inside osu!framework constructs. But for me it looks up the local files because the file paths are stored along with the PDBs in the source link data, so I'm awaiting further testing by @peppy before unblocking.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -37,8 +37,4 @@
     <PackageProjectUrl>https://github.com/ppy/osu-framework</PackageProjectUrl>
     <RepositoryUrl>https://github.com/ppy/osu-framework</RepositoryUrl>
   </PropertyGroup>
-  <PropertyGroup Label="Sourcelink3">
-    <EnableSourceLink Condition=" '$(OS)' != 'Windows_NT' AND '$(MSBuildRuntimeType)' != 'Core' ">false</EnableSourceLink>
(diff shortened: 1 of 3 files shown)


peppy:
Looks to work correctly, after downloading of pdbs completes 👍 .

peppy (approved):

PR #5366, opened 2022-08-19
Add callbacks to text selection events of TextBox

nekodex:
Will be used for adding audio feedback to text selection in lazer

### DIFF
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -1506,8 +1506,6 @@ namespace osu.Framework.Graphics.UserInterface
                 return;
             }
 
-            storeTextSelectionBeforeChange();
-
             // used for tracking the selection to report for `OnImeComposition()`
             int oldStart = selectionStart;
             int oldEnd = selectionEnd;
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -1588,11 +1588,6 @@ namespace osu.Framework.Graphics.UserInterface
 
             if (userEvent) OnImeComposition(newComposition, removeCount, addCount, oldStart != selectionStart || oldEnd != selectionEnd);
 
-            if (newSelectionLength > 0)
-                onTextSelectionChanged(TextSelectionType.Character);
-            else
-                onTextDeselected();
-
             endTextChange(beganChange);
             cursorAndLayout.Invalidate();
         }
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -1689,7 +1689,7 @@ namespace osu.Framework.Graphics.UserInterface
             Word,
 
             /// <summary>
-            /// All of the text was selected (i.e. via Ctrl+A or Cmd+A).
+            /// All of the text was selected (i.e. via <see cref="PlatformAction.SelectAll"/>).
             /// </summary>
             All
         };
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -263,6 +263,7 @@ namespace osu.Framework.Graphics.UserInterface
                     selectionStart = 0;
                     selectionEnd = text.Length;
                     cursorAndLayout.Invalidate();
+                    onTextSelectionChanged(TextSelectionType.All);
                     return true;
 
                 // Cursor Manipulation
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -832,6 +842,49 @@ protected virtual void OnCaretMoved(bool selecting)
...
+        protected virtual void OnTextSelectionChanged(TextSelectionType selectionType)
+        {
+        }
+
+        /// <summary>
+        /// Invoked whenever text is deselected.
+        /// </summary>
+        protected virtual void OnTextDeselected()
+        {
+        }
```
Susko3, line 858: Having both `TextSelectionType.Deslect` and `OnTextDeselected()` feels wrong to me. Either remove `OnTextDeselected()`, or add a note to `OnTextSelectionChanged` that it won't be called with `TextSelectionType.Deslect`.

frenzibyte: This sounds like `TextSelectionType.Deselect` shouldn't be a thing to begin with.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -1513,6 +1574,8 @@ private void onImeComposition(string newComposition, int newSelectionStart, int
 
             if (userEvent) OnImeComposition(newComposition, removeCount, addCount, oldStart != selectionStart || oldEnd != selectionEnd);
 
+            onTextSelectionChanged(newSelectionLength > 0 ? TextSelectionType.Character : TextSelectionType.Deselect);
```
Susko3, line 1577: Not sure this is valid, as IME input is quite different from regular text input. I'd rather add `newSelectionLength` (or `textSelected`) to `OnImeComposition`.

With the osu!-side PR, this currently produces sounds for both `OnImeComposition` and `OnTextSelectionChanged`.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -832,6 +842,49 @@ protected virtual void OnCaretMoved(bool selecting)
...
+        }
+
+        private void onTextSelectionChanged(TextSelectionType selectionType)
+        {
+            if (lastSelectionStart == selectionStart && lastSelectionEnd == selectionEnd) return;
+
+            switch (selectionType)
+            {
+                case TextSelectionType.Deselect:
+                    if (lastSelectionStart != lastSelectionEnd && (lastSelectionStart > 0 || lastSelectionEnd > 0))
```
Susko3, line 867:
```suggestion
                    if (lastSelectionStart != lastSelectionEnd)
```

| `lastSelectionStart` | `lastSelectionEnd` | `lastSelectionStart != lastSelectionEnd` | `lastSelectionStart > 0 \|\| lastSelectionEnd > 0` |
| -- | -- | -- | -- |
| 0 | 0 | false | false |
| 0 | 5 | true | true |
| 4 | 0 | true | true |
| 4 | 5 | true | true |
| 4 | 4 | false | true |

Susko3 (changes requested):

osu.Framework/Graphics/UserInterface/TextBox.cs
nekodex: Hrm, I'll re-check the behaviour with IME.

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -832,6 +850,47 @@ protected virtual void OnCaretMoved(bool selecting)
...
+        private void onTextDeselected()
+        {
+            if (lastSelectionStart == selectionStart && lastSelectionEnd == selectionEnd)
+                return;
+
+            if (lastSelectionStart != lastSelectionEnd)
+                OnTextDeselected();
+        }
+
+        private void storeTextSelectionBeforeChange()
```
peppy, line 888: Before I propose a change here, can you confirm that the `lastSelectionX` values are always intended to be used immediately in the method this is called from? ie. it doesn't really need to be stored to fields in `TextBox`?

nekodex: That method is used to set the `lastSelectionX` values (before a selection change occurs) which are then compared against later in `onTextSelectionChanged`/`onTextDeselected`, if that's what you mean.

peppy: I'm thinking the method should return a tuple with the two values for local use, which could then be passed into those two methods as a single parameter rather than storing to `TextBox` fields. Would that work?

nekodex: Aha yeah, I'll do that.

peppy (approved):

osu.Framework/Graphics/UserInterface/TextBox.cs
peppy: Has been removed.

PR #5370, opened 2022-08-22
Fix potential crash if file with incorrect filename format is in font cache directory

peppy:
Closes https://github.com/ppy/osu/issues/19846#issuecomment-1221406051.

Also applied NRT while here.


peppy:
Cherry-picked culture locking from https://github.com/ppy/osu-framework/pull/5369 (slight change due to use of `TryParse` now).

smoogipoo (approved):

PR #5371, opened 2022-08-22
Fix potential startup crash if resolution/display pair from configuration results in no valid resolution

peppy:
Closes https://github.com/ppy/osu/issues/19877#issuecomment-1221438234.

I wasn't able to reproduce what the user was seeing even when trying my best to break it, but have tested the flow works by artifically adding an exception:

```diff
diff --git a/osu.Framework/Platform/SDL2DesktopWindow.cs b/osu.Framework/Platform/SDL2DesktopWindow.cs
index 8ca11367f..d68d8368f 100644
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -1233,6 +1233,9 @@ private void updateWindowStateAndSize()
                     {
                         var closestMode = getClosestDisplayMode(sizeFullscreen.Value, currentDisplayMode.Value.RefreshRate, currentDisplay.Index);
 
+                        if (!sizeFullscreen.IsDefault)
+                            throw new InvalidOperationException("test");
+
                         Size = new Size(closestMode.w, closestMode.h);
 
                         SDL.SDL_SetWindowDisplayMode(SDLWindowHandle, ref closestMode);

```

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow.cs
@@ -1229,12 +1229,29 @@ namespace osu.Framework.Platform
                     break;
 
                 case WindowState.Fullscreen:
-                    var closestMode = getClosestDisplayMode(sizeFullscreen.Value, currentDisplayMode.Value.RefreshRate, currentDisplay.Index);
+                    try
+                    {
+                        var closestMode = getClosestDisplayMode(sizeFullscreen.Value, currentDisplayMode.Value.RefreshRate, currentDisplay.Index);
 
-                    Size = new Size(closestMode.w, closestMode.h);
+                        Size = new Size(closestMode.w, closestMode.h);
 
-                    SDL.SDL_SetWindowDisplayMode(SDLWindowHandle, ref closestMode);
-                    SDL.SDL_SetWindowFullscreen(SDLWindowHandle, (uint)SDL.SDL_WindowFlags.SDL_WINDOW_FULLSCREEN);
+                        SDL.SDL_SetWindowDisplayMode(SDLWindowHandle, ref closestMode);
+                        SDL.SDL_SetWindowFullscreen(SDLWindowHandle, (uint)SDL.SDL_WindowFlags.SDL_WINDOW_FULLSCREEN);
+                    }
+                    catch
+                    {
+                        Logger.Log($"Unable to set specified fullscreen resolution ({sizeFullscreen} on display {windowDisplayIndexBindable}), resetting to defaults and trying again");
+
+                        // If a fullscreen display mode could not be retrieved, fallback to defaults and try again.
+                        if (!sizeFullscreen.IsDefault)
+                        {
+                            sizeFullscreen.SetDefault();
+                            windowDisplayIndexBindable.SetDefault();
+                            updateWindowSpecifics();
+                            return;
+                        }
+                    }
+
                     break;
 
                 case WindowState.FullscreenBorderless:


smoogipoo (approved):

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -1229,12 +1229,29 @@ private void updateWindowStateAndSize()
...
+                    {
+                        Logger.Log($"Unable to set specified fullscreen resolution ({sizeFullscreen} on display {windowDisplayIndexBindable}), resetting to defaults and trying again");
+
+                        // If a fullscreen display mode could not be retrieved, fallback to defaults and try again.
+                        if (!sizeFullscreen.IsDefault)
+                        {
+                            sizeFullscreen.SetDefault();
+                            windowDisplayIndexBindable.SetDefault();
+                            updateWindowSpecifics();
+                            return;
```
Susko3, line 1251: Setting `windowDisplayIndexBindable` doesn't actually do anything. It just sets `CurrentDisplay`, and tells the window to call `updateWindowStateAndSize()` with the same state again.

https://github.com/ppy/osu-framework/blob/5235e4e1e18b95f6a2a4b5490f3d8af78949ecd1/osu.Framework/Platform/SDL2DesktopWindow.cs#L1409-L1414

But `CurrentDisplay` (not to be confused with `currentDisplay` or `CurrentDisplayBindable`) is only used when storing window position to config (a noop in fullscreen):

image

But.... `FrameworkSetting.LastDisplayDevice`/`windowDisplayIndexBindable` does actually work when starting the game in fullscreen. I guess the full startup flow just happens to propagate the correct values.

Susko3 (changes requested):
This doesn't actually fix the [original issue](https://github.com/ppy/osu/issues/19877#issuecomment-1221499986), as there the `FrameworkSetting.SizeFullscreen` was already set to default.

The windowing/event/bindable flow in SDL2DesktopWinows is really confusing and convoluted. Probably the safest way to handle this is to reset most of the framework config to default, save, and then crash the game by rethrowing.

This could happen in `getClosestDisplayMode()` just before the throw.

osu.Framework/Platform/SDL2DesktopWindow.cs
peppy: I was counting on this being the piece that would fix the issue, but yeah. Doesn't look like it will do anything.

I fail to see how resetting the framework config to default will help if this doesn't though. Can you explain?

Susko3: [Resetting to default](https://github.com/ppy/osu/issues/19877#issuecomment-1221499986) helped in the original issue. And deleting it also helped you with some macOS native SDL crashes.

`windowDisplayIndexBindable` works with fullscreen on startup only because the window is actually in the `Normal` state on first creation.

```
2022-08-22 08:57:50 [verbose]: setting windowState to Normal
2022-08-22 08:57:50 [verbose]: stack:    at osu.Framework.Platform.SDL2DesktopWindow.set_windowState(WindowState value)
2022-08-22 08:57:50 [verbose]: at osu.Framework.Platform.SDL2DesktopWindow.updateWindowSpecifics()
2022-08-22 08:57:50 [verbose]: at osu.Framework.Platform.SDL2DesktopWindow.Create()
2022-08-22 08:57:50 [verbose]: at osu.Framework.Platform.Windows.WindowsWindow.Create()
2022-08-22 08:57:50 [verbose]: at osu.Framework.Platform.GameHost.Run(Game game)
2022-08-22 08:57:50 [verbose]: at osu.Framework.Tests.Program.Main(String[] args)
2022-08-22 08:57:50 [verbose]: 
2022-08-22 08:57:50 [verbose]: readWindowPositionFromConfig
2022-08-22 08:57:50 [verbose]: Stored display index (-1) doesn't match current index (1)
2022-08-22 08:57:50 [verbose]: Updated display mode to desktop resolution: 1920x1080@59, SDL_PIXELFORMAT_RGB888
2022-08-22 08:57:50 [verbose]: Host execution state changed to Running
2022-08-22 08:57:50 [verbose]: Execution mode changed to MultiThreaded
2022-08-22 08:57:51 [verbose]: setting windowState to Fullscreen
2022-08-22 08:57:51 [verbose]: stack:    at osu.Framework.Platform.SDL2DesktopWindow.set_windowState(WindowState value)
2022-08-22 08:57:51 [verbose]: at osu.Framework.Platform.SDL2DesktopWindow.updateWindowSpecifics()
2022-08-22 08:57:51 [verbose]: at osu.Framework.Platform.SDL2DesktopWindow.Run()
2022-08-22 08:57:51 [verbose]: at osu.Framework.Platform.GameHost.Run(Game game)
2022-08-22 08:57:51 [verbose]: at osu.Framework.Tests.Program.Main(String[] args)
2022-08-22 08:57:51 [verbose]: 
2022-08-22 08:57:51 [verbose]: Updated display mode to fullscreen resolution: 1920x1080@59, SDL_PIXELFORMAT_RGB888
2022-08-22 08:57:51 [verbose]: setting windowState to Fullscreen
2022-08-22 08:57:51 [verbose]: stack:    at osu.Framework.Platform.SDL2DesktopWindow.set_windowState(WindowState value)
2022-08-22 08:57:51 [verbose]: at osu.Framework.Platform.SDL2DesktopWindow.updateWindowSpecifics()
2022-08-22 08:57:51 [verbose]: at osu.Framework.Platform.SDL2DesktopWindow.handleWindowEvent(SDL_WindowEvent evtWindow)
2022-08-22 08:57:51 [verbose]: at osu.Framework.Platform.SDL2DesktopWindow.handleSDLEvent(SDL_Event e)
2022-08-22 08:57:51 [verbose]: at osu.Framework.Platform.SDL2DesktopWindow.pollSDLEvents()
2022-08-22 08:57:51 [verbose]: at osu.Framework.Platform.SDL2DesktopWindow.Run()
2022-08-22 08:57:51 [verbose]: at osu.Framework.Platform.GameHost.Run(Game game)
2022-08-22 08:57:51 [verbose]: at osu.Framework.Tests.Program.Main(String[] args)
2022-08-22 08:57:51 [verbose]: 
```

peppy: I'm just not sure resetting configuration is going to help here without restarting the game too. I'd rather only reset the pieces that would fix this issue.

Susko3: Reset some configuration and restart the game (or just force close it and let the user open it again) is what I'm thinking.

Resetting these might be useful:

- `WindowedSize` - in case the window is too large for the display (this was the fix for the [original issue](https://github.com/ppy/osu/issues/19877#issuecomment-1222093203)) (could even set to 640x480 if the default 1366x768 also fails)
- `WindowedPositionX` and `WindowedPositionY` - in case the window being halfway between two monitors breaks it
- `LastDisplayDevice` - in case a monitor is disconnected
- `SizeFullscreen` - in case of invalid fullscreen size
- `WindowMode` - not really needed but might be useful

peppy: @Susko3 if you happen to beat me to fixing this please do. I think it deserves priority but haven't got to it yet.

Susko3: I'll give it a shot, but I think fixing this will require a substantial rewrite of the windowing logic.

peppy:
#5649

PR #5372, opened 2022-08-22
Add method to get configuration from a `ConfigManager` for logging purposes

peppy:
Implementation borrowed from [OsuConfigManager](https://github.com/ppy/osu/blob/3fb3a18e68e58bb016d3340a84c44517b3b520f7/osu.Game/Configuration/OsuConfigManager.cs#L175-L187).

Allows logging framework configuration to sentry, which can be quite useful in tracking down certain issues. The case I'm interested in which made me want to add this was knowing whether the game is running in single or multithreaded mode.

### DIFF
--- a/osu.Framework/Configuration/ConfigManager.cs
+++ b/osu.Framework/Configuration/ConfigManager.cs
@@ -39,7 +39,7 @@ namespace osu.Framework.Configuration
         /// Get the full configuration for logging purposes.
         /// </summary>
         /// <remarks>
-        /// Excludes any potentially sensitive information via </remarks>
+        /// Excludes any potentially sensitive information via <see cref="CheckLookupContainsPrivateInformation"/>.</remarks>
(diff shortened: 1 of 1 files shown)


osu.Framework/Configuration/ConfigManager.cs
```diff
@@ -34,6 +35,23 @@ protected ConfigManager(IDictionary<TLookup, object> defaultOverrides = null)
             this.defaultOverrides = defaultOverrides;
         }
 
+        /// <summary>
+        /// Get the full configuration for logging purposes.
+        /// </summary>
+        /// <remarks>
+        /// Excludes any potentially sensitive information via </remarks>
```
Susko3, line 42: Missing a `<see cref="CheckLookupContainsPrivateInformation">`.

bdach (approved):

PR #5373, opened 2022-08-22
Fix `SliderBar` causing a crash if disabled while the user is dragging

peppy:


osu.Framework/Graphics/UserInterface/SliderBar.cs
```diff
@@ -213,6 +223,9 @@ private bool commit()
             if (!uncommittedChanges)
                 return false;
 
+            if (currentNumberInstantaneous.Disabled)
+                return false;
```
bdach, line 227: Bit weirded out by this guard. `current.DisabledChanged` already sets `uncommittedChanges = false`, which is checked by the previous guard, which makes this feel a little redundant. I initially thought this was some sort of extra safety against data races, but if it were, you'd probably want to check `current.Disabled` instead, as that's what's about to be assigned to and what would be the cause of a crash. What was the intent here?

For whatever it's worth, tests pass with this guard removed.

peppy: I don't think this is required anymore. I originally had things implemented here but it was as expected, too late.

bdach (approved):

PR #5374, opened 2022-08-22
Do not apply hidden state to system drives in directory selector

mashiro-azure:
A small QoL commit to fix the DirectorySelectorDirectory applying hidden state to system drives, thus fixing a color discrepancy between local drives and network drives. Breadcrumbs are not affected by this issue.

**Seen on: osu-framework & osu-lazer**
**Tested on: Windows 10 21H2 (Build 19044.1826)**

Original behavior (master):
master-Test

Additional Information (`directory` Attributes):
Local Drive:
master-localDrive
Network Drive:
master-networkDrive

Fixed:
fix-Test

P.S This is my first time contributing to an open-source project, feel free to point out anything that needs to be improved. Thank you.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/DirectorySelectorDirectory.cs
+++ b/osu.Framework/Graphics/UserInterface/DirectorySelectorDirectory.cs
@@ -29,7 +29,8 @@ namespace osu.Framework.Graphics.UserInterface
             {
                 bool isHidden = directory?.Attributes.HasFlagFast(FileAttributes.Hidden) == true;
 
-                // System drives show up as `System | Hidden | Directory` but shouldn't be shown in a hidden state.
+                // On Windows, system drives are returned with `System | Hidden | Directory` file attributes,
+                // but the expectation is that they shouldn't be shown in a hidden state.
                 bool isSystemDrive = directory?.Parent == null;
 
                 if (isHidden && !isSystemDrive)
--- a/osu.Framework/Graphics/UserInterface/DirectorySelectorDirectory.cs
+++ b/osu.Framework/Graphics/UserInterface/DirectorySelectorDirectory.cs
@@ -27,7 +27,8 @@ namespace osu.Framework.Graphics.UserInterface
 
             try
             {
-                if (directory?.Attributes.HasFlagFast(FileAttributes.Hidden) == true)
+                if (directory?.Attributes.HasFlagFast(FileAttributes.System) == true) { }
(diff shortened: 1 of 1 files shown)


peppy:
I can reproduce what you're seeing (system drives showing dimmed). Kinda looks okay in my opinion, but probably incorrect.

osu.Framework/Graphics/UserInterface/DirectorySelectorDirectory.cs
```diff
@@ -27,7 +27,8 @@ protected DirectorySelectorDirectory(DirectoryInfo directory, string displayName
 
             try
             {
-                if (directory?.Attributes.HasFlagFast(FileAttributes.Hidden) == true)
+                if (directory?.Attributes.HasFlagFast(FileAttributes.System) == true) { }
+                else if (directory?.Attributes.HasFlagFast(FileAttributes.Hidden) == true)
```
peppy, line 31: This should read better:

```suggestion
                bool isHidden = directory?.Attributes.HasFlagFast(FileAttributes.Hidden) == true;

                // System drives show up as `System | Hidden | Directory` but shouldn't be shown in a hidden state.
                bool isSystem = directory?.Attributes.HasFlagFast(FileAttributes.System) == true;

                if (isHidden && !isSystem)
```

peppy (dismissed):

mashiro-azure:
Suggested changes committed to PR branch, ready for review and merge. Thank you.

(Sorry if I'm doing anything wrong, still new to GitHub.)

osu.Framework/Graphics/UserInterface/DirectorySelectorDirectory.cs
bdach: I dunno about using `IsSystem` here at all. From a semantic point of view there is no reason why a hidden system file should generally become unhidden.

I'd probably use `directoryInfo.Parent == null` or similar as it feels more targeted for catching the "this is actually a drive" case.

peppy: Can agree with this. Seems less prone to failing.

bdach (approved):

bdach:
@mashiro-azure For future reference, your git install looks misconfigured. The commits in this pull that haven't been made through github have a broken author (shows as `local-branch <>` to me). It seems like you didn't set `user.name` and `user.email` in global config - I'm looking away this time, but please fix it if you are going to continue to submit PRs.

PR #5376, opened 2022-08-23
Expose method to allow implementations to perform operations when drawable items change in `RearrangeableListContainer`

peppy:
Without this it's basically impossible to perform post-change filtering.

Will be used to fix remaining test failures noticed in `TestScenePlaylistOverlay`.

### DIFF
--- a/osu.Framework/Graphics/Containers/RearrangeableListContainer.cs
+++ b/osu.Framework/Graphics/Containers/RearrangeableListContainer.cs
@@ -75,6 +75,13 @@ namespace osu.Framework.Graphics.Containers
             Items.CollectionChanged += collectionChanged;
         }
 
+        /// <summary>
+        /// Fired whenever new drawable items are added or removed from <see cref="ListContainer"/>.
+        /// </summary>
+        protected virtual void OnItemsChanged()
+        {
+        }
+
         private void collectionChanged(object sender, NotifyCollectionChangedEventArgs e)
         {
             switch (e.Action)
@@ -120,7 +127,8 @@ namespace osu.Framework.Graphics.Containers
                 itemMap.Remove(item);
             }
 
-            reSort();
+            sortItems();
+            OnItemsChanged();
         }
 
         private void addItems(IList items)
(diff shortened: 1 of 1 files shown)


bdach:
A thing or two:

* Right now this the new method may not fire if the collection is reset, as that is using an entirely disjoint code path:

    https://github.com/ppy/osu-framework/blob/d20f5ac523b4b646c108a72a3d3cc26ef17f5324/osu.Framework/Graphics/Containers/RearrangeableListContainer.cs#L95-L99

* Given that a `Replace` collection change is implemented as removal + addition, then `OnItemsChanged()` will fire twice then. Maybe the easier way out is to call it at the very end of [`collectionChanged()`](https://github.com/ppy/osu-framework/blob/d20f5ac523b4b646c108a72a3d3cc26ef17f5324/osu.Framework/Graphics/Containers/RearrangeableListContainer.cs#L78) instead?

peppy:
> * Right now this the new method may not fire if the collection is reset, as that is using an entirely disjoint code path:

I've added a call there.

> * Given that a `Replace` collection change is implemented as removal + addition, then `OnItemsChanged()` will fire twice then. Maybe the easier way out is to call it at the very end of [`collectionChanged()`](https://github.com/ppy/osu-framework/blob/d20f5ac523b4b646c108a72a3d3cc26ef17f5324/osu.Framework/Graphics/Containers/RearrangeableListContainer.cs#L78) instead?

And this is why I had to add the event in the first place, at the *very specific* place it's added. It's also the cause of the bug I'm aiming to fix. If this was moved to the proposed location, it would fire before the new drawable items are async loaded, making it non-useful.

Not sure there's a better way to explain this in the code. Could rename `OnItemsChanged` to `OnDrawableItemsChanged` (what I had originally) but I'm not sure that reads better.

bdach:
> And this is why I had to add the event in the first place, at the _very specific_ place it's added

Duh, I misread the code and didn't notice the asyncness in the add case. Whoops.

bdach (approved):

PR #5380, opened 2022-08-25
Avoid performing `Seek` operations on a decoupled source when the source is not running

peppy:
With https://github.com/ppy/osu-framework/pull/5355, we are now always performing a source seek. On the surface this seemed like it would be fine, but it turns out there is a performance issue if `Seek` is called too often.

This reverts the non-`Coupled` case to not perform a source seek if not required.

~I'm still cross-testing this with my ongoing clock changes to make sure it works well.~ Seems fine.


smoogipoo:
Indeed fixes the issue.

peppy:
I'm not sure how correct this is, but honestly just want to avoid breaking things. I'm still planning to completely replace this decoupled stuff from scratch using what we've learnt over the last few years.

smoogipoo (approved):

PR #5381, opened 2022-08-25
Prefix all remaining OpenGL-specific classes with "GL" in name

frenzibyte:
First steps towards supporting Veldrid, need to have all OpenGL-specific classes prefixed with `GL` to not conflict in names with the upcoming Veldrid variants. Few notes:
 - `GLLinear{Buffer,Batch}` reads a bit weird given the double L in the name, but not sure if it should be renamed.
 - `VideoTexture` has been renamed to `VideoTextuerGL` and also been moved to the `OpenGL.Textures` namespace, instead of its original place (`Videos`).
 - `IVertexBuffer` has also been prefixed as it's currently internal and OpenGL-specific, but it could potentially be moved out if we decide to have a base `Renderer`/`VertexBatch` class which makes use of this interface.

### DIFF
--- a/osu.Framework/Graphics/OpenGL/Buffers/GLLinearBuffer.cs
+++ b/osu.Framework/Graphics/OpenGL/Buffers/GLLinearBuffer.cs
@@ -10,9 +10,9 @@ using osuTK.Graphics.ES30;
 
 namespace osu.Framework.Graphics.OpenGL.Buffers
 {
-    internal static class OpenGLLinearIndexData
+    internal static class GLLinearIndexData
     {
-        static OpenGLLinearIndexData()
+        static GLLinearIndexData()
         {
             GL.GenBuffers(1, out EBO_ID);
         }
@@ -42,17 +42,17 @@ namespace osu.Framework.Graphics.OpenGL.Buffers
         {
             base.Initialise();
 
-            if (amountVertices > OpenGLLinearIndexData.MaxAmountIndices)
+            if (amountVertices > GLLinearIndexData.MaxAmountIndices)
             {
                 ushort[] indices = new ushort[amountVertices];
 
                 for (int i = 0; i < amountVertices; i++)
                     indices[i] = (ushort)i;
 
-                Renderer.BindBuffer(BufferTarget.ElementArrayBuffer, OpenGLLinearIndexData.EBO_ID);
+                Renderer.BindBuffer(BufferTarget.ElementArrayBuffer, GLLinearIndexData.EBO_ID);
                 GL.BufferData(BufferTarget.ElementArrayBuffer, (IntPtr)(amountVertices * sizeof(ushort)), indices, BufferUsageHint.StaticDraw);
 
-                OpenGLLinearIndexData.MaxAmountIndices = amountVertices;
+                GLLinearIndexData.MaxAmountIndices = amountVertices;
             }
         }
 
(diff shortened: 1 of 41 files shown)


peppy:
I'd prefer `GL` rather than `OpenGL` for brevity. Also optionally, it could be a suffix in all cases similar to `VideoTextureGL`. Need more opinions on that.

frenzibyte:
I was under the thought that the “GL” suffix was used to indicate that the class isn’t directly an inheritor of an interface/class with the same name (i.e. `TextureGL` looks like it’s own class, while `OpenGLTexture` sounds like it inherits from `Texture`).

Bit of a weak argument though, and I’ve already opted to call the Veldrid one as `VeldridTexture` rather than `TextureVd` anyways.

smoogipoo:
> “GL” suffix was used to indicate that the class isn’t directly an inheritor

There's nothing special about the naming. I agree with using "GL".

> Also optionally, it could be a suffix

Hmm. We typically don't suffix things in o!f/lazer, as opposed to stable. Between `VertexBufferVeldrid`/`VertexBufferVd` and `VeldridVertexBuffer`/`VdVertexBuffer`, I think I prefer the latter.

peppy:
Prefix is fine. "Vd" feels weird so I'd prefer using "Veldrid".

peppy (dismissed):

osu.Framework/Graphics/OpenGL/Buffers/GLLinearBuffer.cs
```diff
@@ -10,9 +10,9 @@
 
 namespace osu.Framework.Graphics.OpenGL.Buffers
 {
-    internal static class LinearIndexData
+    internal static class OpenGLLinearIndexData
```
smoogipoo, line 13: Still using `OpenGL` here.

osu.Framework/Graphics/OpenGL/Buffers/GLQuadBuffer.cs
```diff
@@ -11,9 +11,9 @@
 
 namespace osu.Framework.Graphics.OpenGL.Buffers
 {
-    internal static class QuadIndexData
+    internal static class OpenGLQuadIndexData
```
smoogipoo, line 14: Still using `OpenGL` here.

osu.Framework/Graphics/OpenGL/Textures/VideoTextureGL.cs
```diff
@@ -1,23 +1,22 @@
...
 using osu.Framework.Graphics.Textures;
+using osu.Framework.Graphics.Video;
 using osu.Framework.Platform;
+using osuTK.Graphics.ES30;
 
-namespace osu.Framework.Graphics.Video
+namespace osu.Framework.Graphics.OpenGL.Textures
 {
-    internal unsafe class VideoTexture : TextureGL
+    internal unsafe class VideoTextureGL : TextureGL
```
smoogipoo, line 15: Should be prefix instead of suffix.

osu.Framework/Graphics/OpenGL/Textures/TextureGL.cs
```diff
@@ -19,7 +19,7 @@ namespace osu.Framework.Graphics.OpenGL.Textures
 {
     internal class TextureGL : INativeTexture
```
smoogipoo, line 20: Prefix instead of suffix

smoogipoo (changes requested):

smoogipoo (approved):

PR #5382, opened 2022-08-26
Update `IContainer` interface to add explicit disposal argument on `Remove` methods

### LINKED ISSUE #5378: `IContainer.RemoveAll` is dangerous
We have time and time again used it in cases where drawables are to be removed permanently, without understanding that it bypasses disposal (and leaves things like `Bindables` bound, potentially forever).

Proposal:

```diff
diff --git a/osu.Framework/Graphics/Containers/IContainer.cs b/osu.Framework/Graphics/Containers/IContainer.cs
index 415821f23..73469599b 100644
--- a/osu.Framework/Graphics/Containers/IContainer.cs
+++ b/osu.Framework/Graphics/Containers/IContainer.cs
@@ -24,7 +24,7 @@ public interface IContainerEnumerable<out T> : IContainer
     {
         IReadOnlyList<T> Children { get; }
 
-        int RemoveAll(Predicate<T> match);
+        int RemoveAll(Predicate<T> match, bool disposeImmediately);
     }
 
     public interface IContainerCollection<in T> : IContainer
@@ -39,7 +39,7 @@ public interface IContainerCollection<in T> : IContainer
         void Add(T drawable);
         void AddRange(IEnumerable<T> collection);
 
-        bool Remove(T drawable);
-        void RemoveRange(IEnumerable<T> range);
+        bool Remove(T drawable, bool disposeImmediately);
+        void RemoveRange(IEnumerable<T> range, bool disposeImmediately);
     }
 }

```

Intentionally a breaking change to allow consumers to reassess existing usages.

cc/ @smoogipoo @bdach @frenzibyte

peppy:
Closes #5378.

### `CompositeDrawable.RemoveInternal`, `IContainer.Remove`, `IContainer.RemoveRange` and `IContainer.RemoveAll` now require a `bool` parameter

A common mistake we've seen made osu!-side is where drawables are `Remove`d from the hierarchy with the intention of never using them again. In such cases, disposal is not guaranteed – nor is unbinding of `Bindable` fields/properties – which can lead to event and object leakage.

To prevent this from happening, a new `disposeImmediately` parameter has been added.

Generally this should be set to `true` unless you intend to reuse the removed drawable (ie. by adding back to the hierarchy at a different location).

### DIFF
--- a/osu.Framework/Graphics/Containers/AudioContainer.cs
+++ b/osu.Framework/Graphics/Containers/AudioContainer.cs
@@ -100,7 +100,8 @@ namespace osu.Framework.Graphics.Containers
             set => container.Children = value;
         }
 
-        public int RemoveAll(Predicate<T> match) => container.RemoveAll(match);
+        public int RemoveAll(Predicate<T> match, bool disposeImmediately) =>
+            container.RemoveAll(match, disposeImmediately);
 
         public T Child
         {
@@ -135,14 +136,17 @@ namespace osu.Framework.Graphics.Containers
             container.AddRange(collection);
         }
 
-        public bool Remove(T drawable) => container.Remove(drawable);
+        public bool Remove(T drawable, bool disposeImmediately) => container.Remove(drawable, disposeImmediately);
+
+        bool ICollection<T>.Remove(T item) => container.Remove(item, true);
+
         int ICollection<T>.Count => container.Count;
 
         public bool IsReadOnly => container.IsReadOnly;
 
-        public void RemoveRange(IEnumerable<T> range)
+        public void RemoveRange(IEnumerable<T> range, bool disposeImmediately)
         {
-            container.RemoveRange(range);
+            container.RemoveRange(range, disposeImmediately);
         }
 
         int IReadOnlyCollection<T>.Count => container.Count;
--- a/osu.Framework/Graphics/Containers/Container.cs
+++ b/osu.Framework/Graphics/Containers/Container.cs
@@ -3,15 +3,15 @@
 
 #nullable disable
 
-using osu.Framework.Lists;
-using System.Collections.Generic;
 using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
 using osu.Framework.Extensions.TypeExtensions;
 using osu.Framework.Graphics.Colour;
-using osuTK;
-using System.Collections;
-using System.Diagnostics;
 using osu.Framework.Graphics.Effects;
+using osu.Framework.Lists;
+using osuTK;
 
 namespace osu.Framework.Graphics.Containers
 {
@@ -206,6 +206,12 @@ namespace osu.Framework.Graphics.Containers
         {
             if (drawable == Content)
                 throw new InvalidOperationException("Content may not be added to itself.");
+
+            if (drawable == null)
+                throw new ArgumentNullException(nameof(drawable));
+
+            if (drawable.IsDisposed)
+                throw new ObjectDisposedException(nameof(drawable));
 
             if (Content == this)
                 AddInternal(drawable);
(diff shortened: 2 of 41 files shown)


peppy:
I'm making an osu!-side companion to this still, will block until done and confirmed all is okay.

peppy:
I've tested various gameplay scenarios and things seem to be working well enough.

JetBrains Rider 2022-08-26 at 06 48 51

osu.Framework/Graphics/Containers/Container.cs
```diff
@@ -230,26 +232,40 @@ public void AddRange(IEnumerable<T> range)
...
+            if (Content != this)
+                return Content.Remove(drawable, disposeImmediately);
+
+            bool wasRemoved = RemoveInternal(drawable);
+
+            if (disposeImmediately)
+                drawable.Dispose();
+
+            return wasRemoved;
+        }
```
frenzibyte, line 263: Is there a reason `RemoveInternal` isn't updated to contain this argument as well, similar to `ClearInternal`?

peppy: It likely should be, but it's a lot of extra work (many usages). I can look at that if deemed best to do it all in one swoop.

frenzibyte: The diff is already large as it stands and is a breaking change, so I think it's worth breaking `RemoveInternal` similarly while at it rather than leaving it for later.

peppy: Have added this. Also updated osu-side changes and confirmed tests pass.

osu.Framework/Graphics/Containers/AudioContainer.cs
```diff
@@ -135,14 +136,17 @@ public void AddRange(IEnumerable<T> collection)
             container.AddRange(collection);
         }
 
-        public bool Remove(T drawable) => container.Remove(drawable);
+        public bool Remove(T drawable, bool disposeImmediately) => container.Remove(drawable, disposeImmediately);
+
+        bool ICollection<T>.Remove(T item) => container.Remove(item, true);
```
frenzibyte, line 141: This comment is for another day, but the code here makes me wonder if `Container` should inherit from `ICollection<T>` in the first place. It already inherits `IReadOnlyList<T>` so LINQ operations will remain to be possible, but having it inherit `ICollection<T>` gets bad given the side effects of immediately disposing drawables on removal.

osu.Framework/Graphics/Visualisation/VisualisedDrawable.cs
```diff
@@ -298,7 +298,7 @@ void IContainVisualisedDrawables.AddVisualiser(VisualisedDrawable visualiser)
             flow.Add(visualiser);
         }
 
-        void IContainVisualisedDrawables.RemoveVisualiser(VisualisedDrawable visualiser) => flow.Remove(visualiser);
+        void IContainVisualisedDrawables.RemoveVisualiser(VisualisedDrawable visualiser) => flow.Remove(visualiser, true);
```
frenzibyte, line 301: This method is used for transferring parents or changing child ID as well:
https://github.com/ppy/osu-framework/blob/04c5dd65bfb42e48afaa3aab2f9dd0bab28db3b8/osu.Framework/Graphics/Visualisation/VisualisedDrawable.cs#L460-L466
https://github.com/ppy/osu-framework/blob/04c5dd65bfb42e48afaa3aab2f9dd0bab28db3b8/osu.Framework/Graphics/Visualisation/VisualisedDrawable.cs#L287-L288
```suggestion
        void IContainVisualisedDrawables.RemoveVisualiser(VisualisedDrawable visualiser) => flow.Remove(visualiser, false);
```

frenzibyte (changes requested):

frenzibyte (approved):

smoogipoo (approved):

PR #5383, opened 2022-08-29
Fix nested storage lookup check not accounting for case insensitive filesystems

peppy:
Resolves https://github.com/ppy/osu/issues/19996#issuecomment-1229428045.

Tests do not fail as it stands. I can't figure out the correct call to reproduce what the user was seeing. Help.

### DIFF
--- a/osu.Framework.Tests/IO/TestDesktopStorage.cs
+++ b/osu.Framework.Tests/IO/TestDesktopStorage.cs
@@ -70,5 +70,40 @@ namespace osu.Framework.Tests.IO
                 Assert.That(storage.GetStorageForDirectory(string.Empty).GetFullPath(string.Empty), Is.EqualTo(storage.GetFullPath(string.Empty)));
             }
         }
+
+        [Test]
+        public void TestCaseSensitivity()
+        {
+            // First test if this unit test is running on a filesystem which is case sensitive.
+            File.WriteAllText(Path.Combine(TestRunHeadlessGameHost.TemporaryTestDirectory, "case-sensitivity-test"), "test");
+            bool isCaseSensitive = !File.Exists(Path.Combine(TestRunHeadlessGameHost.TemporaryTestDirectory, "CASE-SENSITIVITY-TEST"));
+            File.Delete("case-sensitivity-test");
+
+            // Then check our methods match this.
+            using (var storage = new TemporaryNativeStorage("test"))
+            {
+                using (var stream = storage.CreateFileSafely("test-file"))
+                    stream.WriteByte(0);
+
+                Assert.That(storage.Exists("test-file"), Is.True);
+                Assert.That(storage.Exists("TEST-FILE"), Is.EqualTo(!isCaseSensitive));
+
+                using (var storage2 = new TemporaryNativeStorage("TEST"))
+                {
+                    Assert.That(storage2.Exists("test-file"), Is.EqualTo(!isCaseSensitive));
+                    Assert.That(storage2.Exists("TEST-FILE"), Is.EqualTo(!isCaseSensitive));
+
+                    // Test that base path lookup checks don't incorrectly fail.
+                    // This should pass regardless of case sensitivity of the underlying partition.
+                    var nested = storage2.GetStorageForDirectory("subdir");
+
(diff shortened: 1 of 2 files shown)


osu.Framework.Tests/IO/TestDesktopStorage.cs
```diff
@@ -70,5 +70,40 @@ public void TestGetEmptySubDirectoryStorage()
...
             }
         }
+
+        [Test]
+        public void TestCaseSensitivity()
+        {
+            // First test if this unit test is running on a filesystem which is case sensitive.
+            File.WriteAllText(Path.Combine(TestRunHeadlessGameHost.TemporaryTestDirectory, "case-sensitivity-test"), "test");
+            bool isCaseSensitive = !File.Exists(Path.Combine(TestRunHeadlessGameHost.TemporaryTestDirectory, "CASE-SENSITIVITY-TEST"));
+            File.Delete("case-sensitivity-test");
```
bdach, line 80: This `File.Delete()` call is a bit ineffective (implicitly uses process cwd and doesn't actually delete the file from the temp folder).

osu.Framework/Platform/NativeStorage.cs
```diff
@@ -59,7 +59,7 @@ private IEnumerable<string> getRelativePaths(IEnumerable<string> paths)
             string basePath = Path.GetFullPath(GetFullPath(string.Empty));
             return paths.Select(Path.GetFullPath).Select(path =>
             {
-                if (!path.StartsWith(basePath, StringComparison.Ordinal)) throw new ArgumentException($"\"{path}\" does not start with \"{basePath}\" and is probably malformed");
+                if (!path.StartsWith(basePath, StringComparison.OrdinalIgnoreCase)) throw new ArgumentException($"\"{path}\" does not start with \"{basePath}\" and is probably malformed");
```
bdach, line 62: The reason why I didn't propose changing this to `OrdinalIgnoreCase` in the first place is that this is *less safe* now for case-*sensitive* filesystems - as in, this could potentially miss a malformed path.

I am relatively willing to look away - this is a private method and all cases where it is being called are controlled, i.e. basically `GetFiles()` and `GetDirectories()`. This is not an actual issue as it stands, but I think it's a bad idea in general to roll own versions of path parsing methods, because filesystems are not consistent and if handling this case sensitivity nonsense can be offloaded onto the standard library, then it should be.

bdach:
As for the question why the test isn't failing before the fix - I have no answer to that yet. I will note that I tried to reproduce on both a case-sensitive filesystem (ext4) and a case-insensitive one (fat32), but in both cases I was testing from linux, and the sentry reports are exclusive to windows. So I'm thinking there's _possibly_ some windows specific nonsense that affects outputs of `Path.GetFullPath()` or similar? but I have no actual proof...

peppy:
I've tested on windows though, and cannot repro. I was thinking I wasn't covering the precise calls in the order that osu! is.

osu.Framework/Platform/NativeStorage.cs
peppy: For sure, I can look into whether there's a method to replace this completely, but that's relatively high effort as it would need to be re-tested in all platforms.

bdach:
I tried to reproduce using the scenario originally provided by the reporter in the issue this is supposing to fix, *without* the fix, just to see what's crashing, and I'm not getting the crash at all. Not sure what that's about, so I asked for more precise repro steps.

osu.Framework/Platform/NativeStorage.cs
peppy: To elaborate on this, the malformed path check is mainly present as a security consideration, to make sure a user lookup cannot traverse out of a known root. So while the proposed change here is arguably "incorrect" for case-sensitive filesystems, it is also not that much of a concern (as the risk of something going wrong is relatively low due to the whole path having to match minus case).

smoogipoo (approved):

smoogipoo:
Is this fine to merge then?

peppy:
I think so. None of us can repro the original issue, but it *should* fix it and hasn't seemed to break anything during testing.

Probably get this one in and revert is it ends up causing any issues.

PR #5385, opened 2022-09-01
Pull iOS / android platform detection methods locally to avoid potential usage before ready

peppy:


smoogipoo (approved):

bdach:
Can confirm (via running https://github.com/ppy/osu/pull/19971, which includes a framework bump with this change) that this appears to resolve the issue.

PR #5386, opened 2022-09-02
Remove "GL" prefix from performance collection type enum

frenzibyte:
Still kept the name `Reset` as it makes sense for what the underlying operation is doing, despite the fact that it's called `BeginFrame` now.

### DIFF
--- a/osu.Framework/Statistics/FrameStatistics.cs
+++ b/osu.Framework/Statistics/FrameStatistics.cs
@@ -48,7 +48,7 @@ namespace osu.Framework.Statistics
         Sleep,
         Scheduler,
         IPC,
-        Reset,
+        DrawReset,
     }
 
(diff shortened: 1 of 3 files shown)


osu.Framework/Statistics/FrameStatistics.cs
```diff
@@ -48,7 +48,7 @@ internal enum PerformanceCollectionType
         Sleep,
         Scheduler,
         IPC,
-        GLReset,
+        Reset,
```
frenzibyte, line 51: Hmm, now that I look at it again, maybe `DrawReset` would make more sense?

peppy: Probably best.

peppy (approved):

PR #5387, opened 2022-09-02
Update `GlobalPropertyManager` to support registering `IShader`s

frenzibyte:
Was initially going to define a generic-less `GetUniform` method and move the generic variant to an extensions or seal it, but `GlobalPropertyManager` isn't aware whether the global uniform is defined by the shader or not, and `GetUniform` throws if the uniform could not be found.

Therefore I've went with an internal uniforms dictionary and called it a day.

### DIFF
--- a/osu.Framework/Graphics/Shaders/IUniform.cs
+++ b/osu.Framework/Graphics/Shaders/IUniform.cs
@@ -8,12 +8,29 @@ namespace osu.Framework.Graphics.Shaders
     /// <summary>
     /// Represents an updateable shader uniform.
     /// </summary>
-    internal interface IUniform
+    public interface IUniform
     {
-        void Update();
-
+        /// <summary>
+        /// The shader which this uniform was declared in.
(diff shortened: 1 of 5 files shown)


peppy (approved):

osu.Framework/Graphics/Shaders/IUniform.cs
```diff
@@ -8,12 +8,29 @@ namespace osu.Framework.Graphics.Shaders
     /// <summary>
     /// Represents an updateable shader uniform.
     /// </summary>
-    internal interface IUniform
+    public interface IUniform
```
smoogipoo, line 11: Does this need to be public?

frenzibyte: I was initially going to make the dictionary `protected internal` to allow custom shader implementations, but custom renderers are already not supported properly so will revert this for now.

smoogipoo (approved):

PR #5388, opened 2022-09-02
Add `unmanaged` constraint to vertex/uniform generic types

frenzibyte:
Required for Veldrid as it disallows updating vertex/uniform buffers with values that do not use the `unmanaged` constraint (see [`UpdateBuffer`](https://github.com/mellinoe/veldrid/blob/b881f7dcb444c4856e29464435b10d4be0ff38e0/src/Veldrid/CommandList.cs#L799-L803)). I've also updated non-required GL variants for consistency.


peppy (approved):

smoogipoo:
Makes sense. This is a breaking change though so I'll add that in.

smoogipoo:
Nevermind, it doesn't seem to affect as much as I thought it did.

smoogipoo (approved):

PR #5389, opened 2022-09-03
Disable threaded optimization on NVIDIA GPUs for Windows users

khang06:
Fixes ppy/osu#19421.

This pull request adds a bunch of code to interface with NVAPI to disable threaded optimization. Initializing NVAPI before the OpenGL context also seems to force the dedicated GPU to be used according to some online threads, but I haven't tested this myself because I don't have the hardware.

I'm a bit concerned about how the metadata is being filled out, but I couldn't find a good way to pass the game name from the game host all the way down to the graphics backend.


peppy:
Thanks for the contribution.

Rather than reimplementing things from scratch, did you consider existing libraries? https://www.nuget.org/packages?q=nvapi

peppy:
Also I have no idea if it will be helpful in review of this change, but here's stable's code:

https://gist.github.com/1ad50c4d3d9e376977ebbd8c598279e8

khang06:
I did look into existing libraries, but I didn't really want to risk screwing up any license-related stuff from third-party packages. I'm only implementing a small subset of the API, so it wasn't much of a problem to just do it myself instead.

osu.Framework/Platform/Windows/WindowsSDL2GraphicsBackend.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// <summary>
+    /// A version of <see cref="SDL2GraphicsBackend"/> that configures NVIDIA drivers on Windows for better performance.
+    /// </summary>
+    public class WindowsSDL2GraphicsBackend : SDL2GraphicsBackend
+    {
+        public override void InitialiseBeforeWindowCreation()
+        {
+            // Try to initialize NVAPI
+            // It's important to do this before graphics API initialization to make sure that the dedicated GPU is being used
+            NvLibrary.TryInitialize();
```
smoogipoo, line 21: I've forgotten the specifics of NVAPI, does this mean that initting it forces the dGPU?

khang06: Apparently just loading nvapi.dll is enough to force it to be enabled according to [this](https://stackoverflow.com/a/46749111/7795411), but I can't test this myself since I don't have a laptop with a dedicated GPU.

smoogipoo: Need to think about this one a bit, because I'm not sure that's the best thing to do if so. For one it completely disables exclusive fullscreen at least until Veldrid + D3D11 is used. Also I believe there are performance concerns but I can't remember the specifics.

peppy: Many users are specifically using the integrated GPU for better performance.

In stable I believe we have pre-checks to not run nvapi code on dual GPU devices in certain scenarios.

Might be best to hold off on this altogether until we get Direct X online, which may make it far less important a change.

khang06: It'll probably work fine if I move it after window initialization, but I still wouldn't be able to test how well that works out.

smoogipoo: I could test that out, but it's probably worth waiting for Veldrid (+D3D11) anyway. I think @frenzibyte has some code which would conflict with this too.

khang06: Yeah, I was a bit worried about the OpenGL vendor reading part and how that would work out with other backends.

peppy:
https://github.com/ppy/osu/pull/26354

PR #5390, opened 2022-09-03
Rename `DepthStencilFunction` to `BufferTestFunction`

Flutterish:
Around the time #5351 was merged, `DepthTestFunction` was renamed to `DepthStencilFunction` (#5352) to allow the enum to be reused.
At the time I didn't pay too much attention to it, but I believe the name `BufferTestFunction` works better as a generalisation of the `{Depth,Stencil}TestFunction` concept.

### DIFF
--- a/osu.Framework/Graphics/Rendering/BufferTestFunction.cs
+++ b/osu.Framework/Graphics/Rendering/BufferTestFunction.cs
@@ -3,7 +3,7 @@
 
 namespace osu.Framework.Graphics.Rendering
 {
-    public enum DepthStencilFunction
+    public enum BufferTestFunction
     {
         /// <summary>
         /// The test always fails.
--- a/osu.Framework.Tests/Visual/Graphics/TestSceneStencil.cs
+++ b/osu.Framework.Tests/Visual/Graphics/TestSceneStencil.cs
@@ -105,7 +105,7 @@ namespace osu.Framework.Tests.Visual.Graphics
                 // Populate the stencil buffer with 255 in places defined by the draw node.
                 renderer.PushStencilInfo(new StencilInfo(
                     stencilTest: true,
-                    testFunction: DepthStencilFunction.Always,
+                    testFunction: BufferTestFunction.Always,
                     255,
                     passed: StencilOperation.Replace));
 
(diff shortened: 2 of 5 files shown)


osu.Framework/Graphics/Rendering/BufferTestFunction.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace osu.Framework.Graphics.Rendering
 {
-    public enum DepthStencilFunction
+    public enum BufferTestFunction
```
frenzibyte, line 6: Frankly, I would've preferred the name 
`BufferComparisonFunction` more if anything, since the point of this enum is to decide how two values should be compared to tell which takes over the other. The name "buffer test" doesn't really indicate that as much as "depth/stencil test" would do, which are known stages of rendering. Will wait for opinions from others first.

Flutterish: One counterpoint of "Comparison" I was considering while thinking of names is the fact that it contains constant Always/Never which aren't really comparisons. Another things is that a "Test" is used to determine if a fragment gets drawn to screen at all.

smoogipoo:
Is there another usage of these functions in the native APIs? GL/Veldrid/etc.

frenzibyte:
On Veldrid (and OpenGL 3.2+), the same enum is used in [texture sampling](https://veldrid.dev/api/Veldrid.SamplerDescription.html#Veldrid_SamplerDescription_ComparisonKind) ([Vulkan](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkSamplerCreateInfo.html), [Direct3D 11](https://docs.microsoft.com/en-us/windows/win32/api/d3d11/ns-d3d11-d3d11_sampler_desc), [Metal](https://developer.apple.com/documentation/metal/mtlsamplerdescriptor/1516001-comparefunction), [OpenGL 3.2+](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glSamplerParameter.xhtml)).

smoogipoo (approved):

frenzibyte (approved):
Since one is fine with the new name then I guess it's fine.

PR #5392, opened 2022-09-06
Improve invalidation/validation performance

smoogipoo:
Initial:

|     Method |      Mean |     Error |    StdDev |
|----------- |----------:|----------:|----------:|
| Invalidate | 2.4286 ns | 0.0092 ns | 0.0072 ns |
|   Validate | 0.8363 ns | 0.0022 ns | 0.0020 ns |

Removing exception:

|     Method |      Mean |     Error |    StdDev |
|----------- |----------:|----------:|----------:|
| Invalidate | 0.5200 ns | 0.0233 ns | 0.0218 ns |
|   Validate | 0.8370 ns | 0.0063 ns | 0.0056 ns |

Removing branching:

|     Method |      Mean |     Error |    StdDev |
|----------- |----------:|----------:|----------:|
| Invalidate | 0.5070 ns | 0.0161 ns | 0.0143 ns |
|   Validate | 0.5107 ns | 0.0126 ns | 0.0098 ns |

I suspect the branching is different on different CPUs and different workloads, but will now be consistent across the board.

I expect this can be further vectorised as future work, but probably won't lead to much improvement in this case.

### DIFF
--- a/osu.Framework/Graphics/InvalidationList.cs
+++ b/osu.Framework/Graphics/InvalidationList.cs
@@ -1,9 +1,8 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-#nullable disable
-
 using System;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
 using osu.Framework.Layout;
 
@@ -53,8 +52,11 @@ namespace osu.Framework.Graphics
                     return invalidate(ref childInvalidation, flags);
 
                 default:
-                    throw new ArgumentException("Unexpected invalidation source.", nameof(source));
+                    return throwInvalidSourceException();
             }
+
+            [DoesNotReturn]
+            static bool throwInvalidSourceException() => throw new ArgumentException("Unexpected invalidation source.", nameof(source));
         }
 
         /// <summary>
--- a/osu.Framework/Graphics/InvalidationList.cs
+++ b/osu.Framework/Graphics/InvalidationList.cs
@@ -51,11 +51,8 @@ namespace osu.Framework.Graphics
                 case InvalidationSource.Parent:
                     return invalidate(parentInvalidation, flags, out parentInvalidation);
 
-                case InvalidationSource.Child:
-                    return invalidate(childInvalidation, flags, out childInvalidation);
-
                 default:
-                    return false;
+                    return invalidate(childInvalidation, flags, out childInvalidation);
             }
         }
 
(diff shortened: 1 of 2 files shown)


peppy:
Can you give some context as to what led to this? Were you seeing overhead somewhere?

peppy (dismissed):
Seems sane enough

osu.Framework/Graphics/InvalidationList.cs
```diff
@@ -44,17 +43,20 @@ public bool Invalidate(InvalidationSource source, Invalidation flags)
...
-                    return invalidate(ref childInvalidation, flags);
+                    return invalidate(childInvalidation, flags, out childInvalidation);
 
                 default:
-                    throw new ArgumentException("Unexpected invalidation source.", nameof(source));
+                    return throwInvalidSourceException();
             }
+
+            [DoesNotReturn]
+            static bool throwInvalidSourceException() => throw new ArgumentException("Unexpected invalidation source.", nameof(source));
```
frenzibyte, line 59: Could move to `ThrowHelper` instead in attempt to continue using the class rather than leaving it buried with only one usage.

smoogipoo:
Just randomly profiled one of my storyboard test beatmaps and `Drawable.invalidate()` (i.e. the whole thing) being 6% of total time.

Before:
image

After:
image

osu.Framework/Graphics/InvalidationList.cs
smoogipoo: Ah, I didn't realise we had such a class these days. Awesome.

smoogipoo: I've actually changed it to an assert because it's proven at the (single) callsite. Saves a bit of ASM for the JIT to compile so it can only be a benefit.

|     Method |      Mean |     Error |    StdDev |
|----------- |----------:|----------:|----------:|
| Invalidate | 0.4686 ns | 0.0017 ns | 0.0014 ns |
|   Validate | 0.5156 ns | 0.0057 ns | 0.0053 ns |

osu.Framework/Graphics/InvalidationList.cs
```diff
@@ -51,11 +51,8 @@ public bool Invalidate(InvalidationSource source, Invalidation flags)
                 case InvalidationSource.Parent:
                     return invalidate(parentInvalidation, flags, out parentInvalidation);
 
-                case InvalidationSource.Child:
-                    return invalidate(childInvalidation, flags, out childInvalidation);
-
                 default:
-                    return false;
+                    return invalidate(childInvalidation, flags, out childInvalidation);
```
peppy, line 55: I guess this works fine because the every source is covered above? Seems very micro optimisations.. might want to comment so it isn't changed in the future.

peppy (approved):

PR #5394, opened 2022-09-07
Allow Drawable `Scale` to be exactly zero

smoogipoo:
I don't know the original issue with storyboards so @peppy please test if this works for you.

`DrawInfo` still needs to handle zero scale. As it turns out, this value cannot be 0 for either the scale or the inverse scale, as input breaks either way. Tests have been added for this case.

### DIFF
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -970,11 +970,6 @@ namespace osu.Framework.Graphics
             get => scale;
             set
             {
-                if (Math.Abs(value.X) < Precision.FLOAT_EPSILON)
-                    value.X = Precision.FLOAT_EPSILON;
-                if (Math.Abs(value.Y) < Precision.FLOAT_EPSILON)
-                    value.Y = Precision.FLOAT_EPSILON;
-
                 if (scale == value)
                     return;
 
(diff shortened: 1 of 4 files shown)


osu.Framework/Graphics/Drawable.cs
```diff
@@ -1321,7 +1316,7 @@ public float Alpha
         /// Determines whether this Drawable is present based on its <see cref="Alpha"/> value.
         /// Can be forced always on with <see cref="AlwaysPresent"/>.
         /// </summary>
-        public virtual bool IsPresent => AlwaysPresent || (Alpha > visibility_cutoff && Math.Abs(Scale.X) > Precision.FLOAT_EPSILON && Math.Abs(Scale.Y) > Precision.FLOAT_EPSILON);
+        public virtual bool IsPresent => AlwaysPresent || (Alpha > visibility_cutoff && Scale.X != 0 && Scale.Y != 0);
```
peppy, line 1319: Not 100% sure about this, but I think it's more correct? Also probably a bit more efficient.

peppy:
The storyboard case required local changes to resolve, and wouldn't have hit this in the first place. But I think this makes sense and fixes a potential issue that could have been seen elsewhere.

peppy (approved):

Issue #5379, opened 2022-08-25
`Container<T>` enumerator does not throw if the list is modified


smoogipoo:
`Container<T>` uses a custom enumerator:

https://github.com/ppy/osu-framework/blob/0894ea15a0b54c5f6fea478aa45f20356f68bbdb/osu.Framework/Graphics/Containers/Container.cs#L472-L495

This does not throw if the underlying list is modified, for example...
```
foreach (var child in container)
    container.Remove(child);
```

frenzibyte:
For reference, `List<T>` achieves this by keeping track of a version field that updates whenever the list is mutated, and throws if it no longer matches during iteration.

PR #5396, opened 2022-09-07
Throw exception when modifying contents of Container<T> while iterating through it

hwabis:
Closes https://github.com/ppy/osu-framework/issues/5379

I believe I've covered all the methods that can mutate `Children`?

Also not sure if I should also check the version in `Reset()` of the enumerator. `List<T>` does it but I can't think of the case where that would throw

### DIFF
--- a/osu.Framework/Graphics/Containers/Container.cs
+++ b/osu.Framework/Graphics/Containers/Container.cs
@@ -39,9 +39,9 @@ namespace osu.Framework.Graphics.Containers
         /// <summary>
         /// This is checked when enumerating through this <see cref="Container"/> to throw when
         /// <see cref="Children"/> was mutated while enumerating (in <see cref="Enumerator"/>).
-        /// This is incremented whenever <see cref="Children"/> is mutated (e.g. with <see cref="Add(T)"/>). 
+        /// This is incremented whenever <see cref="Children"/> is mutated (e.g. with <see cref="Add(T)"/>).
         /// </summary>
-        private int version = 0;
+        private int enumeratorVersion;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/Containers/Container.cs
```diff
@@ -36,6 +36,13 @@ public class Container : Container<Drawable>
     public class Container<T> : CompositeDrawable, IContainerEnumerable<T>, IContainerCollection<T>, ICollection<T>, IReadOnlyList<T>
         where T : Drawable
     {
+        /// <summary>
+        /// This is checked when enumerating through this <see cref="Container"/> to throw when
+        /// <see cref="Children"/> was mutated while enumerating (in <see cref="Enumerator"/>).
+        /// This is incremented whenever <see cref="Children"/> is mutated (e.g. with <see cref="Add(T)"/>). 
+        /// </summary>
+        private int version = 0;
```
smoogipoo, line 44: I think this should probably be named like `enumeratorVersion` instead, just to indicate its use.

smoogipoo (approved):

peppy:
@smoogipoo did you happen to test this again osu-side tests? shouldn't stop making the change, but curious if we have any work to do there.

peppy (approved):

smoogipoo:
Ah nope, I haven't checked

smoogipoo:
No test failures.

frenzibyte:
Will note that this doesn't cover the case where the contents of `Container<T>` is actually modified, it only handles addition/deletion. That being said, not 100% sure if it should cover that. Maybe best to do it in order to remain consistent with `List<T>`'s behaviour?

PR #5402, opened 2022-09-13
Add option to neglect non-space characters such as diacritics to SearchContainer

Wleter:
This pull request addresses issue https://github.com/ppy/osu/issues/19591.

With this implementation diacritics are neglected altogether if NonSpaceCharacterMatching is set to false.

I think almost always diacritics should be neglected but maybe there are cases when it matters.

### DIFF
--- a/osu.Framework/Graphics/Containers/SearchContainer.cs
+++ b/osu.Framework/Graphics/Containers/SearchContainer.cs
@@ -126,14 +126,14 @@ namespace osu.Framework.Graphics.Containers
         private static bool checkTerm(string haystack, string needle, bool nonContiguous)
         {
             if (!nonContiguous)
-                return haystack.Contains(needle, StringComparison.OrdinalIgnoreCase);
+                return CultureInfo.InvariantCulture.CompareInfo.IndexOf(haystack, needle, CompareOptions.IgnoreNonSpace | CompareOptions.IgnoreCase) >= 0;
 
             int index = 0;
 
             for (int i = 0; i < needle.Length; i++)
             {
                 // string.IndexOf doesn't have an overload which takes both a `startIndex` and `StringComparison` mode.
-                int found = CultureInfo.InvariantCulture.CompareInfo.IndexOf(haystack, needle[i], index, CompareOptions.OrdinalIgnoreCase);
+                int found = CultureInfo.InvariantCulture.CompareInfo.IndexOf(haystack, needle[i], index, CompareOptions.IgnoreNonSpace | CompareOptions.IgnoreCase);
                 if (found < 0)
                     return false;
 
--- a/osu.Framework/Graphics/Containers/SearchContainer.cs
+++ b/osu.Framework/Graphics/Containers/SearchContainer.cs
@@ -52,6 +52,10 @@ namespace osu.Framework.Graphics.Containers
             }
         }
 
+        /// <summary>
+        /// Whether the matching algorithm should consider non-space characters like Diacritics.
+        /// If <c>false</c>, searching for "Adios" will match "Adiós".
+        /// </summary>
         public bool NonSpaceCharactersMatching
         {
             get => nonSpaceCharactersMatching;
--- a/osu.Framework/Graphics/Containers/SearchContainer.cs
+++ b/osu.Framework/Graphics/Containers/SearchContainer.cs
@@ -33,6 +33,8 @@ namespace osu.Framework.Graphics.Containers
 
         private bool allowNonContiguousMatching;
 
+        private bool nonSpaceCharactersMatching;
+
         /// <summary>
         /// Whether the matching algorithm should consider cases where other characters exist between consecutive characters in the search term.
         /// If <c>true</c>, searching for "BSI" will match "BeatmapSetInfo".
@@ -46,6 +48,19 @@ namespace osu.Framework.Graphics.Containers
                     return;
 
                 allowNonContiguousMatching = value;
+                filterValid.Invalidate();
+            }
+        }
+
+        public bool NonSpaceCharactersMatching
+        {
+            get => nonSpaceCharactersMatching;
+            set
+            {
+                if (value == nonSpaceCharactersMatching)
+                    return;
+
+                nonSpaceCharactersMatching = value;
                 filterValid.Invalidate();
             }
         }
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Containers/SearchContainer.cs
```diff
@@ -50,6 +52,23 @@ public bool AllowNonContiguousMatching
             }
         }
 
+        /// <summary>
+        /// Whether the matching algorithm should consider non-space characters like Diacritics.
+        /// If <c>false</c>, searching for "Adios" will match "Adiós".
+        /// </summary>
+        public bool NonSpaceCharactersMatching
```
smoogipoo, line 59: I find the naming hard to understand/weird. Maybe it's better if it matched the .NET API 1-1? So `bool IgnoreNonSpace` or `bool IgnoreNonSpaceCharacters`, and link to `<see cref="CompareOptions.IgnoreNonSpace"/>`

smoogipoo (changes requested):

osu.Framework/Graphics/Containers/SearchContainer.cs
```diff
@@ -123,21 +142,37 @@ private bool match(IFilterable filterable, IEnumerable<string> searchTerms, bool
...
+            {
+                for (int i = 0; i < needle.Length; i++)
+                {
+                    // string.IndexOf doesn't have an overload which takes both a `startIndex` and `StringComparison` mode.
+                    int found = CultureInfo.InvariantCulture.CompareInfo.IndexOf(haystack, needle[i], index, CompareOptions.OrdinalIgnoreCase);
+                    if (found < 0)
+                        return false;
+
+                    index = found + 1;
+                }
```
Susko3, line 175: This could use:

```cs
var compareOptions = nonSpaceCharacterMatching
    ? CompareOptions.OrdinalIgnoreCase
    : CompareOptions.IgnoreNonSpace | CompareOptions.IgnoreCase;
```

Instead of duplicating code.

Since this is now using non-`Ordinal` comparison, using `CurrentCulture` might be more appropriate (could be provided as an option if needed).

Susko3 (changes requested):

osu.Framework/Graphics/Containers/SearchContainer.cs
Wleter: changed

osu.Framework/Graphics/Containers/SearchContainer.cs
Wleter: Removed code duplication but retained `StringComparison.OrdinalIgnoreCase` for contiguous matching since it is recommended over `CompareOptions.OrdinalIgnoreCase`.

Changing `InvariantCulture` to `CurrentCulture` works for Spanish but for some reason it doesn't work for Polish language. I included this case in tests.

smoogipoo (dismissed):

Susko3 (dismissed):
Have not tested. The code changes look fine now.

I do wonder what kind of a performance impact this (presumably) more expensive comparison has.

Wleter:
> I do wonder what kind of a performance impact this (presumably) more expensive comparison has.

As a safer alternative `IgnoreNonSpace` can be by default set to `false` and changed to `true` only if needed.

Wleter:
I have some difficulties making `SearchContainer` work in benchmark environment. 
So instead I copy pasted `checkTerm` function and worked with it for now.

Searching 1000 random 20 letter strings in 1000 letter string gave results:

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
Intel Core i7-9750H CPU 2.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET SDK=6.0.400
  [Host]     : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT DEBUG
  DefaultJob : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT


```
|                            Method |      Mean |     Error |    StdDev | Allocated |
|---------------------------------- |----------:|----------:|----------:|----------:|
|              SearchIgnoreNonSpace | 62.268 ms | 2.0146 ms | 5.8126 ms |      60 B |
|                SearchWithNonSpace |  5.590 ms | 0.1116 ms | 0.2736 ms |       4 B |
| SearchNonContiguousIgnoreNonSpace | 83.602 ms | 2.0619 ms | 5.9819 ms |      80 B |
|   SearchNonContiguousWithNonSpace |  5.254 ms | 0.1006 ms | 0.2100 ms |       4 B |

Difference is rather high but the question is, is it acceptable difference.

I can try making `SearchContainer` work but I don't know if it is worth.

Susko3:
I think it would be beneficial to have it benchmarked in context, using `SearchContainer`. Using `GameBenchmark` and benchmarking `RunSingleFrame()` should work.

You can use [Params](https://benchmarkdotnet.org/articles/overview.html#params) to set `IgnoreNonSpace`, etc.

If you're unable to make `SearchContainer` work, then make `checkTerm` `internal` and use that in the benchmark instead of copying code. (Commit the benchmark to this PR as well.)

Wleter:
Thanks for suggestion, I guess I was on wrong track last time.

Now I get these results:

``` ini

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22000
Intel Core i7-9750H CPU 2.60GHz, 1 CPU, 12 logical and 6 physical cores
.NET SDK=6.0.400
  [Host]     : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT DEBUG
  DefaultJob : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT


```
|              Method | AllowNonContiguous | IgnoreNonSpace |     Mean |    Error |    StdDev |   Median | Allocated |
|-------------------- |------------------- |--------------- |---------:|---------:|----------:|---------:|----------:|
| **SearchRandomStrings** |              **False** |          **False** | **84.42 μs** | **4.811 μs** | **13.648 μs** | **80.80 μs** |      **2 KB** |
| **SearchRandomStrings** |              **False** |           **True** | **92.62 μs** | **7.832 μs** | **22.470 μs** | **82.90 μs** |      **2 KB** |
| **SearchRandomStrings** |               **True** |          **False** | **81.29 μs** | **3.510 μs** |  **9.900 μs** | **77.50 μs** |      **2 KB** |
| **SearchRandomStrings** |               **True** |           **True** | **83.45 μs** | **4.696 μs** | **13.474 μs** | **78.30 μs** |      **2 KB** |

These results were strange to me but I checked whether `checkTerm()` is being used and it was.
I guess now it searches in batches and then it searched through long `string` and it made the difference.

osu.Framework/Graphics/Containers/SearchContainer.cs
```diff
@@ -94,26 +113,26 @@ protected override void Update()
         private void performFilter()
         {
             string[] terms = (searchTerm ?? string.Empty).Split(new[] { ' ' }, StringSplitOptions.RemoveEmptyEntries);
```
turbedi, line 115: Small suggestion while you are already working on this code:
```suggestion
            string[] terms = (searchTerm ?? string.Empty).Split(' ', StringSplitOptions.RemoveEmptyEntries);
```
To avoid the allocation.

Issue #5399, opened 2022-09-09
`TextureAtlas.Add` is inaccessible for external use.


naoei:
I have recently come across a particular use case where I would have to use the `Add` method within `TextureAtlas`, where I'm generating a ton of textures from a cropped sprite sheet (some sprites can be duplicated in the same texture area horizontally and vertically) for file optimization reasons.

I'm pretty much required to have an allocated texture atlas for this, as for one, the generated textures would definitely overflow the set 1024x1024 texture atlas limit, and secondly, without it, I would end up having thousands of draw calls each frame potentially slowing down the GPU bus.

I have a sample project that can be run over on [GitLab](https://gitlab.com/Altenhh/gd.renderer), specifically over at [SpritesheetTextureStore on line 46](https://gitlab.com/Altenhh/gd.renderer/-/blob/master/gd.render.two/Textures/SpritesheetTextureStore.cs#L46).

peppy:
On a quick check, it looks quite safe to expose this function?

smoogipoo:
I wasn't aware this class was public in the first place. Need to make sure it has all the correct checks and balances too, such as disposal.

PR #5403, opened 2022-09-14
Expose `TextureAtlas.Add()`

smoogipoo:
Resolves https://github.com/ppy/osu-framework/issues/5399

I ended up not making this thing an `IDisposable`, because `.Reset()` loses references to the previous texture. I'd have to at least make this into a `List<>`, but I'm also not sure such complexity is required since it's a rare usage. The underlying `TextureGL` will eventually dispose via the finaliser.

Can change this upon request.


peppy (approved):

peppy:
@frenzibyte you still planning to review this?

frenzibyte:
Was hardly able to find the right time to get to it, but can check it now.

frenzibyte (approved):
Looks good.

PR #5404, opened 2022-09-14
Introduce base implementation for `IRenderer`

frenzibyte:
RFC. After spending enough time to get a working implementation of Veldrid, I've went ahead and gathered all the pieces that are shared between both the GL renderer and the Veldrid renderer to move it into one base `IRenderer` implementation, to keep actual renderers only interact with the graphics API/device (this also helped in fixing bugs I encountered with Veldrid, once I migrated it to using this base implementation).

The new base implementation has been as organised as possible, using regions for each section of the renderer (viewport, scissor, shader, framebuffer, etc.).

I've had to rewrite/refactor a bit of the existing logic in `GLRenderer` to keep methods renderer-agnostic and consistent, mainly:
 - `UseProgram` was removed from `IRenderer` and was replaced by `BindShader`/`UnbindShader`, to be consistent with framebuffer/etc. and also removing the term "program" which is a GL-specific term.

 - https://github.com/ppy/osu-framework/commit/12b8cd205da813fc49402413368b34fe037fc9b3: the `BindTexture` flow was completely refactored and `INativeTexture.Bind` was removed, so that renderer implementations don't have to worry about setting up wrap mode or uploading before bind etc.

   Now, texture-binding comes in three stages:
     1. `public IRenderer.BindTexture(Texture)`: public side of binding texture, performs white pixel optimisation checks and uploads texture before bind etc.
     2. `public Renderer.BindTexture(INativeTexture)`: internal side of binding texture, sets up wrap mode and stores texture in "bound textures" array for optimisation, should always be used by internal components of renderer (such as `GLTexture`/`GLVideoTexture` binding before uploading data)
     3. `abstract Renderer.SetTextureInternal(INativeTexture?)`: texture-binding implementation, handles API-specific side of binding texture, should never be called by any method except for `BindTexture(INativeTexture)`.

 - Video texture binding was refactored slightly such that, during upload, all 3 textures of the video will be bound on the first 3 units, and then `glActiveTexture` will be called on each unit to perform the operations (unlike master, in which all textures are bound to the first unit subsequently, which is not possible now because `BindTexture` only accepts an `INativeTexture`).
 
 - Backbuffer framebuffer is no longer included in the stack, and is indicated by passing `null` to `SetFramebufferInternal` instead. This is because backbuffers don't have `IFrameBuffer` representation of them and doing so may not be ideal, so I've used `null` as an indicator instead.

Few notes in addition:
 - https://github.com/ppy/osu-framework/commit/ed890512b4b00ad80b6a4a480eaad76dca896d4e: `IRenderer.CreateShader`/`IRenderer.CreateShaderPart` has been marked as `protected internal`, to allow consumers to override and create shaders in their custom renderer implementations.
 - https://github.com/ppy/osu-framework/commit/4c241cba706bed42c0250240890e6099f9385d81: `INativeTexture` has been marked public, to allow consumers to override and create native textures in their custom renderer implementations.
 - https://github.com/ppy/osu-framework/commit/22569d6dd184b425eec7a121ec3ab33c47a913e5: `IUniform`/`IUniformWithValue<T>` has been marked public, to allow consumers to override and set uniforms in their custom renderer implementations.


osu.Framework/Graphics/Shaders/IUniform.cs
```diff
@@ -8,7 +8,7 @@ namespace osu.Framework.Graphics.Shaders
     /// <summary>
     /// Represents an updateable shader uniform.
     /// </summary>
-    internal interface IUniform
+    public interface IUniform
```
smoogipoo, line 11: Again not sure about exposing these interfaces just yet. Thinking we'll do a single pass at the very end before we're ready to expose the interfaces, in-case they're going to change slightly (e.g. as `INativeTexture` has).

Please revert (here, `IUniformWithValue`, `INativeTexture`, etc).

smoogipoo (changes requested):

osu.Framework/Graphics/Shaders/IUniform.cs
frenzibyte: Hmm fair point, I think that would also mean `Renderer` will have to be completely internal.

smoogipoo: Ah, `IRenderer` is `public` itself and some interfaces were already exposed publicly. Probably fine as is then...

osu.Framework/Graphics/OpenGL/GLRenderer.cs
```diff
@@ -445,281 +219,14 @@ public void SetBlend(BlendingParameters blendingParameters)
...
-            if (ScissorOffset == offset)
-                return;
+        protected override void SetViewportInternal(RectangleI viewport) => GL.Viewport(viewport.Left, viewport.Top, viewport.Width, viewport.Height);
 
-            ScissorOffset = offset;
-        }
+        protected override void SetScissorInternal(RectangleI scissor) => GL.Scissor(scissor.X, Viewport.Height - scissor.Bottom, scissor.Width, scissor.Height);
 
-        public void PopScissorOffset()
+        protected override void SetDepthInfoInternal(DepthInfo depthInfo)
```
peppy, line 228: The `Internal` suffix on these methods isn't really required (ie. there's no naming conflict). Maybe can be dropped? I personally prefer avoiding using the `internal` keyword for cases like this wherever we can avoid it.

osu.Framework/Graphics/Rendering/RendererDisposalQueue.cs
```diff
@@ -0,0 +1,105 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace osu.Framework.Graphics.Rendering
+{
+    /// <summary>
+    /// Helper class used to manage Veldrid disposals in a thread-safe manner.
+    /// </summary>
+    internal class RendererDisposalQueue
```
peppy, line 15: Can you give a bit of background on this class? It looks to be basically a copy-paste of `GLDisposalQueue` but that class still exists (albeit not used). Did you mean to remove it?

osu.Framework/Graphics/OpenGL/GLRenderer.cs
frenzibyte: I was initially going with your proposal (only adding `Internal` when required), but then I've ended up spending a good full hour on why renderers stopped working all of a sudden, until I realised I was calling a protected version of a method rather than the private one (see [code](https://github.com/ppy/osu-framework/blob/22569d6dd184b425eec7a121ec3ab33c47a913e5/osu.Framework/Graphics/Rendering/Renderer.cs#L189) + fourth bullet for explanation).

Once I realised that, I've went ahead and suffixed all virtual methods with "Internal" for the sole reason that they should never be called except from the main-side one (i.e. `SetFramebufferInternal` should only be called from `setFramebuffer`). Suffix is open for rename (maybe `Core`), but I'm all ears if you have a better approach to this.

osu.Framework/Graphics/Rendering/RendererDisposalQueue.cs
frenzibyte: Yeah, I guess I did too much rebase and cherry-pick to the point I forgot to include the deletion of `GLDisposalQueue` here 😅

osu.Framework/Graphics/OpenGL/GLRenderer.cs
smoogipoo: Maybe rather than `Internal`, `Implementation` is a good alternative?

peppy: I prefer `Implementation`.

smoogipoo (approved):

peppy (approved):

PR #5405, opened 2022-09-14
Update ReSharper CLI to 2022.2.3

smoogipoo:
Should contain a fix for this: https://youtrack.jetbrains.com/issue/RSRP-489541/InspectCode-caches-inconsistent-after-turning-on-NRT

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -9,7 +9,7 @@
       ]
     },
     "jetbrains.resharper.globaltools": {
-      "version": "2022.1.1",
+      "version": "2022.2.3",
       "commands": [
         "jb"
       ]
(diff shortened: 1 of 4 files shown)


peppy (dismissed):

peppy:
I guess there were also some missed cases framework-side.

Explains why I saw these when doing a local framework checkout.

smoogipoo:
Fixed the possible nullref exceptions. I haven't enabled NRT for `Menu` because that class is big - touches quite a few files.

peppy (approved):

smoogipoo:
Sorry for the force push. I've reverted the `GridContainer` NRT changes because they ripple through the rest of the codebase, affecting things like `MarkdownTable` etc...

peppy (approved):

PR #5409, opened 2022-09-14
Fix `PoolableDrawable` prepare call schedules potentially creating overhead if drawable never updated

peppy:
If a drawable is taken from a pool but never `IsPresent` before being returned, the schduled `PrepareForUse` operation may never run. This can cause a single pooled drawable to end up with thousands of schduled delegates.

In this version, I removed usage of `Scheduler` completely. Rationale is that there was already a flag tracking whether prepare has been run (`waitingForPrepare`) and it may avoid creating a `Scheduler` in the case it is not used elsewhere (as these are lazily created in `Drawable`).

If preferred, [see this branch](https://github.com/ppy/osu-framework/compare/master...peppy:osu-framework:fix-pooling-shit-v1?expand=1) for a version which switches to using `AddOnce`.

This fixes one portion of editor performance during slider rotation eventually degrading (https://github.com/ppy/osu/issues/20208).

### DIFF
--- a/osu.Framework/Graphics/Pooling/PoolableDrawable.cs
+++ b/osu.Framework/Graphics/Pooling/PoolableDrawable.cs
@@ -1,13 +1,10 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-#nullable disable
-
 using System;
 using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Layout;
-using osu.Framework.Threading;
 
 namespace osu.Framework.Graphics.Pooling
 {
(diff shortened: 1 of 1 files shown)


frenzibyte:
I personally think it looks more organised to use `Scheduler.AddOnce`, and don't see it behave any different than overriding `Update()` and performing the preparation there.

frenzibyte (approved):
That said, I'm not entirely against `Update()`, and would save method group allocation (which will remain an issue until .NET 7, [I think](https://github.com/dotnet/roslyn/issues/5835#issuecomment-1016824887)).

peppy:
You're direct that in the competent usages we're not saving much, but I believe pooling should be as lean as possible for potential use cases. We would only be using pooling if already looking for high performance.

smoogipoo:
It couldn't be thousands of schedules, right? Because it first `.Cancel()`s the previous schedule?

I'm not against this though.

peppy:
Yeah, cancels but doesn't remove. I hit 30,000 schedules when testing in editor (without the osu! side fixes that avoid the case in the first place).

smoogipoo (approved):

frenzibyte:
So if a non-present drawable keeps scheduling and cancelling previous schedules, it'll just pill up because `Scheduler.Update` is never called? That sounds really bad, will open an issue thread for it at least.

PR #5417, opened 2022-09-15
Update README.md

MMagnate:
Fixed a couple grammatical mistakes.


smoogipoo:
Neither of these are incorrect. The second change sounds unnatural.

LittleEndu:
Agree that the commas should stay, but the first line edit seemed to also include the closing of brackets. Right now the brackets don't close.

smoogipoo:
I didn’t notice that, feel free to open another PR.

PR #5421, opened 2022-09-16
Forward tablet driver output to custom logging target

frenzibyte:
One step towards better tablet notifications for osu!, see https://github.com/ppy/osu/issues/20324.

This stops them from being posted as notifications with raw messages ([notice `entry.Target == null`](https://github.com/ppy/osu/blob/a2492a4ff5e4b07bc24bca171e9c73efa25b3f8f/osu.Game/OsuGame.cs#L1003)). Instead, it allows the game to treat them differently by checking against `entry.LoggerName == ITabletHandler.LOGGER_NAME` when subscribed to logger events.

Also improves console log output in general:
```
[runtime] 2022-09-16 12:39:24 [verbose]: Starting legacy IPC provider...
[runtime] 2022-09-16 12:39:25 [verbose]: Attempting to use custom storage location /Users/salman/osu-lazer-development/
[runtime] 2022-09-16 12:39:25 [verbose]: Storage successfully changed to /Users/salman/osu-lazer-development/.
[tablet] 2022-09-16 12:39:25 [verbose]: Detect: Searching for tablets...
[tablet] 2022-09-16 12:39:25 [debug]: Detect: Searching for tablet '10moon 1060N'
[tablet] 2022-09-16 12:39:25 [debug]: Detect: Searching for tablet 'Acepen AP 1060'
[tablet] 2022-09-16 12:39:25 [debug]: Detect: Searching for tablet 'Artisul A1201'
[tablet] 2022-09-16 12:39:25 [debug]: Detect: Searching for tablet 'Artisul AP604'
...
Console output is being limited. Please check tablet.log for full logs.
[tablet] 2022-09-16 12:39:25 [verbose]: Detect: No tablets were detected.
[runtime] 2022-09-16 12:39:26 [verbose]: SDL error log [debug]: Window data not set
[runtime] 2022-09-16 12:39:26 [verbose]: SDL error log [debug]: Window data not set
[runtime] 2022-09-16 12:39:26 [verbose]: GL Initialized
[runtime] 2022-09-16 12:39:26 [verbose]: GL Version:                 2.1 Metal - 76.3
[runtime] 2022-09-16 12:39:26 [verbose]: GL Renderer:                Apple M1 Pro
[runtime] 2022-09-16 12:39:26 [verbose]: GL Shader Language version: 1.20
[runtime] 2022-09-16 12:39:26 [verbose]: GL Vendor:                  Apple
```

Not 100% sure about the addition of a new log file though, alternatives would be:
 - keep all tablet logs as "verbose" regardless of the level (similar to SDL) and prefix them to identify on logger event callback (+ some regex osu!-side to decode level and message from `entry.Message`, very ugly)
 - keep all tablet logs as "verbose" but add an event from `ITabletHandler` to listen to tablet logs (requires cloning classes because of net6.0-netstandard2.1 dual targetting shit)
 - add a whole other special path in `Logger` for notifications that "shouldn't be shown to user but remain logged in file and in console"
 
Need further thoughts on this, one upside about separate file may be allowing to easily attach tablet logs when directing users to report their issues over at OpenTabletDriver's repo (which is a good idea but may need an agreement with the folks there beforehand, so I'm not doing it just yet).

### DIFF
--- a/osu.Framework.Android/Input/AndroidInputHandler.cs
+++ b/osu.Framework.Android/Input/AndroidInputHandler.cs
@@ -227,7 +227,7 @@ namespace osu.Framework.Android.Input
 
         private void logUnhandledEvent(string methodName, InputEvent inputEvent)
         {
-            Logger.Log($"Unknown {GetType().ReadableName()}.{methodName} event: {inputEvent}");
+            Log($"Unknown {GetType().ReadableName()}.{methodName} event: {inputEvent}");
         }
     }
 }
--- a/osu.Framework.Android/Input/AndroidJoystickHandler.cs
+++ b/osu.Framework.Android/Input/AndroidJoystickHandler.cs
@@ -144,7 +144,7 @@ namespace osu.Framework.Android.Input
                 if (axis.TryGetJoystickAxisSource(out _))
                     return true;
 
-                Logger.Log($"Unknown joystick axis: {axis}");
+                Log($"Unknown joystick axis: {axis}");
                 return false;
             }
         }
--- a/osu.Framework/Input/Handlers/InputHandler.cs
+++ b/osu.Framework/Input/Handlers/InputHandler.cs
@@ -8,12 +8,15 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using osu.Framework.Bindables;
 using osu.Framework.Input.StateChanges;
+using osu.Framework.Logging;
 using osu.Framework.Platform;
 
 namespace osu.Framework.Input.Handlers
 {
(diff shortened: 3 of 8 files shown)


peppy:
Rather than "tablet", how about making it "input" and using it for other cases of input related logging?

frenzibyte:
Seems workable, with a prefix on tablet logs to differentiate between them and other input.

peppy:
Needs conflict resolution now.

peppy (approved):
probably okay?

smoogipoo (approved):

PR #5426, opened 2022-09-20
Add new rider metadata changes for preferring project settings

frenzibyte:
Without this change, Rider would use different codestyle settings and start complaining about it in code analysis. This only started happening on Rider 2022.2.3, but it may have been something that got reset on my machine. @peppy confirm you get that as well?


peppy:
Two questions:

- Does this only happen when using the non-slnf?
- Does this happen on `osu` or any other project, or only `osu-framework`?

frenzibyte:
- Happens on both the desktop solution filter, and all other solutions/filters.
- Came across it on both osu-framework and osu.

peppy:
I'm definitely not seeing this yet..

frenzibyte:
Never mind, the `Default` code style configuration was somehow modified, reverting changes fixed it.

PR #5427, opened 2022-09-23
Fix disabling `UseFullGlyphHeight` does not trim top part of sprite text glyphs

### LINKED ISSUE #5166: Disabling `UseFullGlyphHeight` does not trim top part of sprite text glyphs
Currently disabling the property would end up making the sprite text look unaligned instead:
|                | `UseFullGlyphHeight = true` | `UseFullGlyphHeight = false` |
|----------------|:---------------------------:|:----------------------------:|
| with margin    | CleanShot 2022-05-10 at 12 51 35@2x | CleanShot 2022-05-10 at 12 51 53@2x |
| without margin | CleanShot 2022-05-10 at 12 50 20@2x |  |

As can be seen on the above screenshots, `UseFullGlyphHeight = false` does not trim the top part of the sprite text glyphs, causing them to look unaligned on UI components.

This property already works well on osu!-side's `LegacySpriteText`s as they don't have extra padding on the top like normal fonts do.

---

Used code for reference:
```csharp
            AutoSizeAxes = Axes.Both;
            InternalChildren = new Drawable[]
            {
                new Box
                {
                    RelativeSizeAxes = Axes.Both,
                },
                new SpriteText
                {
                    Anchor = Anchor.Centre,
                    Origin = Anchor.Centre,
                    Font = FontUsage.Default.With(size: 25),
                    // Margin = new MarginPadding(0),
                    // Margin = new MarginPadding(5),
                    Colour = Color4.Black,
                    // UseFullGlyphHeight = false,
                    // UseFullGlyphHeight = true,
                    Text = someText,
                }
            };
```

HiddenNode:
- Closes #5166.

I've modified an existing test because it wasn't testing corectly, and added a test to check that bounds are set correctly when `UseFontSizeAsHeight` is true.

- [x] Add more tests

Edit: i've added more test coverage mainly for code that currently doesn't have any use cases.

|  | master | this pr |
| ------------- | ------------- | ------------- |
| With margin | before_with_margin | after_margin |
|Without margin | before_no_margin | after_without_margin |


osu.Framework/Text/TextBuilder.cs
```diff
@@ -298,12 +326,20 @@ public void RemoveLastCharacter()
...
+
                 // As above, the bounds are calculated through the character draw rectangles
                 Bounds = Vector2.ComponentMax(
                     Bounds,
                     new Vector2(
-                        Characters[i].DrawRectangle.Left - Characters[i].XOffset + Characters[i].XAdvance,
-                        Characters[i].DrawRectangle.Top - Characters[i].YOffset + currentLineHeight)
+                        characterRightBound,
+                        characterBottomBound)
                 );
```
frenzibyte: Really hoping this could be computed based on `currentPos` + last character, but that's for another day.

osu.Framework/Text/TextBuilder.cs
```diff
@@ -245,8 +260,21 @@ public void RemoveLastCharacter()
...
+                // this is used only when useFontSizeAsHeight is false.
+                float glyphBaseLine = (Characters[i].Baseline - Characters[i].YOffset);
+
+                float glyphHeightOffset = 0;
+                float previousCharactersHeightOffset = 0;
+
+                if ((currentLineBase ?? 0f) > glyphBaseLine)
+                    glyphHeightOffset = (useFontSizeAsHeight) ? 0 : (currentLineBase ?? 0f) - glyphBaseLine;
+                else if ((currentLineBase ?? 0f) < glyphBaseLine)
+                    previousCharactersHeightOffset = (useFontSizeAsHeight) ? 0 : glyphBaseLine - (currentLineBase ?? 0f);
```
frenzibyte: Combining null coalesces with conditional/ternary statements makes it almost impossible to parse, better to add an explicit conditional for when `currentLineBase == null` instead.

frenzibyte (commented):
Thanks for the contribution. I've put notes on several pieces of the code, be sure to have a read on them as notes for future contributions. I've already gone through the changes myself since the code requires deep understanding of how it works currently (save for the `glyphHeightOffset` logic which I still don't understand).

osu.Framework/Text/TextBuilder.cs
```diff
@@ -164,29 +164,44 @@ public bool AddCharacter(char character)
...
-            else if (glyph.Baseline < currentLineBase)
-                glyph.DrawRectangle = glyph.DrawRectangle.Offset(0, currentLineBase.Value - glyph.Baseline);
 
             Characters.Add(glyph);
 
             currentPos.X += glyph.XAdvance;
-            currentLineBase = currentLineBase == null ? glyph.Baseline : Math.Max(currentLineBase.Value, glyph.Baseline);
-            currentLineHeight = Math.Max(currentLineHeight, getGlyphHeight(ref glyph));
+            currentLineHeight = Math.Max(currentLineHeight + previousCharactersHeightOffset, getGlyphHeight(ref glyph) + glyphHeightOffset);
+            currentLineBase = currentLineBase == null ? glyphBaseLine : Math.Max(currentLineBase.Value, glyphBaseLine);
```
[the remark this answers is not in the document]
HiddenNode: Thank you for taking the time to review my code.

`previousCharactersHeightOffset` is the offset applied to the previous characters when the `currentLineBase` is adjusted.
consider this scenario:
`builder.AddText("q");`

Result:
q
`builder.Add("P");`

Result with/without `previousCharactersHeightOffset`:
|with|without|
|-----|--------|
| qP| qP_n |

`glyphHeightOffset` is similar to `previousCharactersHeightOffset` and it would come into play if "q" was added after "P".

osu.Framework/Text/TextBuilder.cs
HiddenNode: `currentPos` + last character is not enough to compute the bounds of multiple lines as that would also require the maximum width of previous lines ( consider what happens when you have multiple lines and the last character of the longest line gets removed).

osu.Framework/Text/TextBuilder.cs
```diff
@@ -178,15 +181,25 @@ public bool AddCharacter(char character)
...
+                    ? 0
+                    : currentLineBase.Value - glyph.Baseline;
+
                 glyph.DrawRectangle = glyph.DrawRectangle.Offset(0, currentLineBase.Value - glyph.Baseline);
+            }
 
             Characters.Add(glyph);
 
             currentPos.X += glyph.XAdvance;
+            currentLineHeight = Math.Max(currentLineHeight + previousCharactersHeightOffset, getGlyphHeight(ref glyph) + glyphHeightOffset);
```
smoogipoo, line 201: If we assume that:
- `glyph.Baseline >= glyph.Height` ([defined by the font](https://www.angelcode.com/products/bmfont/doc/render_text.html)), and:
- The line height and baseline are monotonically increasing as new characters are added, then:

Isn't this pre-addition of the "height" offset (actually appears to be more of a "baseline" offset) the same as a post-addition? For example, couldn't this be written as:
```csharp
currentLineHeight = Math.Max(currentLineHeight, getGlyphHeight(ref glyph))
                    // Of course, assuming currentLineBase has a value...
                    + Math.Max(0, glyph.Baseline - currentLineBase.Value); 
```

In other words, if `!useFontSizeAsHeight` (the concerned case), then I _believe_ `currentLineHeight == currentLineBase` if my understanding is correct?

smoogipoo: > `glyph.Baseline >= glyph.Height`

On a second read, this actually doesn't appear to be true.

HiddenNode: > > `glyph.Baseline >= glyph.Height`
> 
> On a second read, this actually doesn't appear to be true.

looking at [the full picture](https://freetype.org/freetype2/docs/glyphs/glyphs-3.html) i don't get the impression that  `glyph.Baseline >= glyph.Height` either, and looking at the values at run-time we can see that the `glyph.Height` can be bigger than `glyph.Baseline` (and when `useFontSizeAsHeight` it actually always is (and should be) for characters that go below the baseline for ex: "q" "p"):

g_n

> Isn't this pre-addition of the "height" offset (actually appears to be more of a "baseline" offset) the same as a post-addition? For example, couldn't this be written as:

It actually is a baseline offset, that's just bad naming on my part.

osu.Framework/Text/TextBuilderGlyph.cs
```diff
@@ -16,13 +16,33 @@ public struct TextBuilderGlyph : ITexturedCharacterGlyph
...
+        public readonly float Baseline
+        {
+            get
+            {
+                if (!useFontSizeAsHeight)
+                    return (Glyph.Baseline - Glyph.YOffset) * textSize;
+
+                return Glyph.Baseline * textSize;
+            }
+        }
```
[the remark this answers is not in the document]
HiddenNode: Done.

smoogipoo (approved):
I think this is fine. As far as I can tell there's very few usages osu!-side and they don't seem heavily/at all impacted by this. The Argon main circle piece uses `UseFullGlyphHeight = true` but does its own offsetting, but that could use this and appears better aligned with `UseFullGlyphHeight = false`

osu.Framework/Text/TextBuilder.cs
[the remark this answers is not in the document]
frenzibyte: Thanks for the clarification, I've added test coverage for it.

frenzibyte:
The line height adjustment logic in `AddCharacter`/`RemoveLastCharacter` can be vastly simplified using the draw rectangle of the glyph instead:

```diff
diff --git a/osu.Framework/Text/TextBuilder.cs b/osu.Framework/Text/TextBuilder.cs
index d95ea5448..a890d3486 100644
--- a/osu.Framework/Text/TextBuilder.cs
+++ b/osu.Framework/Text/TextBuilder.cs
@@ -164,9 +164,6 @@ public bool AddCharacter(char character)
             // The kerning is only added after it is guaranteed that the character will be added, to not leave the current position in a bad state
             currentPos.X += kerning;
 
-            float glyphBaselineOffset = 0;
-            float previousCharactersBaselineOffset = 0;
-
             glyph.DrawRectangle = new RectangleF(new Vector2(currentPos.X + glyph.XOffset, currentPos.Y + glyph.YOffset), new Vector2(glyph.Width, glyph.Height));
             glyph.OnNewLine = currentNewLine;
 
@@ -178,27 +175,22 @@ public bool AddCharacter(char character)
                     previous.DrawRectangle = previous.DrawRectangle.Offset(0, glyph.Baseline - currentLineBase.Value);
                     Characters[i] = previous;
 
+                    currentLineHeight = Math.Max(currentLineHeight, previous.DrawRectangle.Bottom);
+
                     if (previous.OnNewLine)
                         break;
                 }
-
-                previousCharactersBaselineOffset = (glyph.OnNewLine || useFontSizeAsHeight || char.IsWhiteSpace(character))
-                    ? 0
-                    : glyph.Baseline - currentLineBase.Value;
             }
             else if (glyph.Baseline < currentLineBase)
             {
-                glyphBaselineOffset = (useFontSizeAsHeight || char.IsWhiteSpace(character))
-                    ? 0
-                    : currentLineBase.Value - glyph.Baseline;
-
                 glyph.DrawRectangle = glyph.DrawRectangle.Offset(0, currentLineBase.Value - glyph.Baseline);
+                currentLineHeight = Math.Max(currentLineHeight, glyph.DrawRectangle.Bottom);
             }
 
             Characters.Add(glyph);
 
             currentPos.X += glyph.XAdvance;
-            currentLineHeight = Math.Max(currentLineHeight + previousCharactersBaselineOffset, getGlyphHeight(ref glyph) + glyphBaselineOffset);
+            currentLineHeight = Math.Max(currentLineHeight, getGlyphHeight(ref glyph));
             currentLineBase = currentLineBase == null ? glyph.Baseline : Math.Max(currentLineBase.Value, glyph.Baseline);
             currentNewLine = false;
 
@@ -258,19 +250,8 @@ public void RemoveLastCharacter()
             {
                 var character = Characters[i];
 
-                float glyphHeightOffset = 0;
-                float previousCharactersHeightOffset = 0;
-
-                if (currentLineBase != null)
-                {
-                    if (character.Baseline > currentLineBase)
-                        previousCharactersHeightOffset = useFontSizeAsHeight ? 0 : character.Baseline - currentLineBase.Value;
-                    else if (character.Baseline < currentLineBase)
-                        glyphHeightOffset = useFontSizeAsHeight ? 0 : currentLineBase.Value - character.Baseline;
-                }
-
                 currentLineBase = currentLineBase == null ? character.Baseline : Math.Max(currentLineBase.Value, character.Baseline);
-                currentLineHeight = Math.Max(currentLineHeight + previousCharactersHeightOffset, getGlyphHeight(ref character) + glyphHeightOffset);
+                currentLineHeight = Math.Max(currentLineHeight, Math.Max(getGlyphHeight(ref character), character.DrawRectangle.Bottom));
 
                 if (character.OnNewLine)
                     break;
```

But the test scene needs a decent amount of rework in order to work, as it uses invalid numbers for the glyph's rectangle which causes this kind of change to break it horribly.

CleanShot 2022-11-04 at 22 41 29

frenzibyte (approved):

peppy (approved):

PR #5431, opened 2022-09-26
Fix VBO incorrectly flushing on no-op shader bind

peppy:
Previously, we were maintaining a variable pointing to the current shader such that if a `Pop` then `Push` happened resulting in the same shader being bound, it would not trigger a buffer flush or any draw call overhead.

This check was removed incorrectly.

Regressed in https://github.com/ppy/osu-framework/pull/5404.

Closes ppy/osu#20421.

### DIFF
--- a/osu.Framework/Graphics/Rendering/Renderer.cs
+++ b/osu.Framework/Graphics/Rendering/Renderer.cs
@@ -932,15 +932,18 @@ namespace osu.Framework.Graphics.Rendering
         {
             ThreadSafety.EnsureDrawThread();
 
+            if (shader == Shader)
+                return;
+
             if (shader != null)
             {
                 FrameStatistics.Increment(StatisticsCounterType.ShaderBinds);
 
                 FlushCurrentBatch();
                 SetShaderImplementation(shader);
-            }
 
-            Shader = shader;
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):

osu.Framework/Graphics/Rendering/Renderer.cs
```diff
@@ -932,15 +932,18 @@ private void setShader(IShader? shader)
...
             if (shader != null)
             {
                 FrameStatistics.Increment(StatisticsCounterType.ShaderBinds);
 
                 FlushCurrentBatch();
                 SetShaderImplementation(shader);
-            }
 
-            Shader = shader;
+                Shader = shader;
```
frenzibyte, line 945: `Shader` should be reset back to null if `setShader(null)` is called, to prevent the shader cleanup logic in `BeginFrame` from kicking in (which is used for when a shader is bound but not unbound before next frame). That said, leaving `Shader` as-is won't affect anything as `GLShader.Unbind` already guards against calling unbind while not actually bound.

peppy: Want to PR this change before it's forgotten?

PR #5432, opened 2022-09-27
Always show exclusive fullscreen as enabled on Intel platforms

smoogipoo:
Resolves https://github.com/ppy/osu/issues/19961

I've moved the code to `WindowsGLRenderer` since it now queries the GL vendor to check if it's `Intel`. I've tested this to work as expected on a PC with only an Intel iGPU and a laptop with both iGPU and dGPU (Optimus).

Hopefully this won't interfere with @frenzibyte 's changes.

### DIFF
--- a/osu.Framework/Platform/Windows/WindowsGLRenderer.cs
+++ b/osu.Framework/Platform/Windows/WindowsGLRenderer.cs
@@ -1,0 +1,102 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+using osu.Framework.Bindables;
+using osu.Framework.Graphics.OpenGL;
+using osu.Framework.Logging;
+
+namespace osu.Framework.Platform.Windows
+{
+    internal class WindowsGLRenderer : GLRenderer, IWindowsRenderer
+    {
+        public IBindable<FullscreenCapability> FullscreenCapability => fullscreenCapability;
+        private readonly Bindable<FullscreenCapability> fullscreenCapability = new Bindable<FullscreenCapability>();
+
+        private readonly WindowsGameHost host;
+
+        public WindowsGLRenderer(WindowsGameHost host)
+        {
+            this.host = host;
+        }
+
+        protected override void Initialise()
+        {
+            base.Initialise();
+
+            WindowsWindow window = (WindowsWindow)host.Window;
+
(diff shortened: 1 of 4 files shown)


osu.Framework/Platform/Windows/WindowsGLRenderer.cs
```diff
@@ -0,0 +1,115 @@
...
+            var cancellationToken = cancellationSource.Token;
+
+            // 50 attempts, 100ms apart = run the detection for a total of 5 seconds before yielding an incapable state.
+            const int max_attempts = 50;
+            const int time_per_attempt = 100;
+            int attempts = 0;
+
+            queueNextAttempt();
+
+            void queueNextAttempt() => Task.Delay(time_per_attempt, cancellationToken).ContinueWith(_ =>
```
smoogipoo, line 70: Note that this code is no longer tied to the window thread. It doesn't look to be required and I've tested the code to work fine without it.

osu.Framework/Platform/Windows/WindowsGLRenderer.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using osu.Framework.Bindables;
+using osu.Framework.Graphics.OpenGL;
+using osu.Framework.Logging;
+using osuTK.Graphics.ES30;
+
+namespace osu.Framework.Platform.Windows
+{
+    internal class WindowsGLRenderer : GLRenderer, IWindowsRenderer
```
frenzibyte, line 15: Is the fullscreen capabilities logic here supposed to be for OpenGL only? There is still only one renderer currently, but otherwise I'm thinking about having this placed and exposed in `WindowsGameHost`, with the vendor being determined by a property exposed by `IRenderer`.

peppy: At least for direct x, it should be a given that the system can run exclusive fullscreen.

peppy (approved):
Probably okay

osu.Framework/Platform/Windows/WindowsGLRenderer.cs
smoogipoo: I think it'll be hard to move this into `WindowsGameHost` because the order of initialisation is kinda undefined. The renderer is initialised later than the window itself, which means it could be possible to bind to the activated state of the window and have a `string.Empty` vendor in the renderer.  
As far as I can tell there's no way to hook onto the point in time where the renderer is initialised since that happens internally to `DrawThread`.

Yes, D3D is should always be able to run exclusive fullscreen. About vulkan/veldrid-gl - those can either duplicate this code or the winapi part of this code can be extracted, but otherwise each of those will likely have their own internal method calls.

PR #5436, opened 2022-09-30
Split out windowing and input of `SDL2DesktopWindow` into partial classes

Susko3:
This mostly moves code around, the only actual code changes are enabling NRT and c760ade059cc852a86b84b05f3c853345fdda5cb.

Hopefully this will make future changes easier, as the relevant code is now closer together.


peppy (approved):
Seems like a good first step, we can build on things from here.

smoogipoo:
@frenzibyte how much does this conflict with what you're doing?

frenzibyte:
Should be fine as the separated code is unrelated to what I have changed in my branch.

smoogipoo (approved):

PR #5438, opened 2022-10-02
Avoid flushing VBO when the renderer state is not affected

frenzibyte:
Haven't profiled this yet, but I've noticed certain cases where VBO could be flushed while no actual changes occur. Most of these cases may come from a path where flushing VBO wouldn't add any overhead, but still worth having it to define the real purpose of this method for any future renderer states (we could also consider encapsulating these methods into a `RendererState` class, but that's for a rainy day).

Everything seems to still work fine, but double-checking would be appreciated.


smoogipoo (approved):
Nice one

PR #5439, opened 2022-10-03
Fix `Displays` querying on every access and add `DisplaysChanged`

Susko3:
peppy reported it [on discord](https://discord.com/channels/188630481301012481/589331078574112768/1019675813953683568): `Displays` were queried in a hot path (every frame).

Adds `DisplaysChanged` event to `IWindow` so consumers can watch for display changes. For now, this doesn't check if the displays actually changed, and will be invoked every time the displays are queried.

### DIFF
--- a/osu.Framework/Platform/IWindow.cs
+++ b/osu.Framework/Platform/IWindow.cs
@@ -121,6 +121,12 @@ namespace osu.Framework.Platform
         IEnumerable<Display> Displays { get; }
 
         /// <summary>
+        /// Invoked when <see cref="Displays"/> has changed.
+        /// </summary>
+        [CanBeNull]
+        event Action<IEnumerable<Display>> DisplaysChanged;
+
+        /// <summary>
         /// Gets the <see cref="Display"/> that has been set as "primary" or "default" in the operating system.
         /// </summary>
         Display PrimaryDisplay { get; }
--- a/osu.Framework.Tests/Visual/Platform/TestSceneBorderless.cs
+++ b/osu.Framework.Tests/Visual/Platform/TestSceneBorderless.cs
@@ -4,6 +4,7 @@
 #nullable disable
 
 using System;
+using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
 using osu.Framework.Allocation;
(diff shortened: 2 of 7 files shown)


peppy (approved):
I *think* this looks correct

osu.Framework/Platform/IWindow.cs
```diff
@@ -120,6 +120,12 @@ public interface IWindow : IDisposable
         /// </summary>
         IEnumerable<Display> Displays { get; }
 
+        /// <summary>
+        /// Invoked when <see cref="Displays"/> has changed.
+        /// </summary>
+        [CanBeNull]
+        event Action<IEnumerable<Display>> DisplaysChanged;
```
frenzibyte, line 127: Is there a reason a callback event is used instead of making the displays list into a bindable list instead?

Susko3: I had contemplated using a `BindableList`, but this was scrapped for multiple reasons:

- using it as `ItemSource` in dropdowns would make sense, but this crashes as it's not on the update thread
- updating all the elements in the list requires `Clear()`ing it first, and this would cause trouble with things that expect at least one available display
- `BindCollectionChanged` / `NotifyCollectionChangedEventArgs` doesn't provide the actual list (existing consumer code expects a list)
- the way displays can change means that rebuilding the entire list is almost always necessary, so an event callback is simpler and gets the job done

frenzibyte: > - using it as ItemSource in dropdowns would make sense, but this crashes as it's not on the update thread

`Dropdown` should be responsible for scheduling updates of `ItemSource`, similar to how it already does with `Current`.

> - BindCollectionChanged / NotifyCollectionChangedEventArgs doesn't provide the actual list (existing consumer code expects a list)

A `BindableList` can be iterated as a list, while `NotifyCollectionChangedEventArgs` would only inform addition/removal of displays.

> - updating all the elements in the list requires Clear()ing it first, and this would cause trouble with things that expect at least one available display
> - the way displays can change means that rebuilding the entire list is almost always necessary, so an event callback is simpler and gets the job done

Fair enough, `BindableList` would be a nice addition to simplify listing of displays, but I guess it could be left off as it is until it's necessary.

PR #5442, opened 2022-10-03
Fix some physical keyboards not working on Android

Susko3:
- Closes https://github.com/ppy/osu/issues/19762

Confirmed this doesn't regress on my devices.
Pending testing on an affected device (I'll be able to borrow one soon).

The actual fix is only adding `InputSourceType.Dpad` to `AndroidKeyboardHandler.HandledEventSources`.
The rest of the code is for supporting/suppressing logging.


Susko3:
Tested on Samsung Galaxy Tab S8 with the official keyboard cover.
The keyboard works as expected now (not so much for the touchpad, but that is out of scope for this PR).

osu.Framework.Android/Input/AndroidInputHandler.cs
```diff
@@ -228,5 +228,29 @@ private void logUnhandledEvent(string methodName, InputEvent inputEvent)
...
+            /// </remarks>
+            Unhandled,
+
+            /// <summary>
+            /// Same as <see cref="Unhandled"/>, but will not be logged.
+            /// </summary>
+            /// <remarks>
+            /// Used when an event might also be handled by another handler, but that cannot be determined purely on <see cref="AndroidInputHandler.HandledEventSources"/>.
+            /// </remarks>
+            UnhandledSuppressLogging,
```
peppy, line 253: Hmmm... do we really need the logging in the first place? Wondering if it's better to move that to a debug flag or something, then remove the overhead of having this return `enum` in the first place.

Susko3: > do we really need the logging in the first place? Wondering if it's better to move that to a debug flag or something

I think logging by default is helpful here, reduces back-and-forth in Android input-related issues. A debug flag that enables logging for all events might be useful when the current logic handles an event, but it produces wrong results.


> remove the overhead of having this return `enum` in the first place.

Isn't `enum` basically an `int`, and the same goes for `bool`?

`UnhandledSuppressLogging` is kind of an workaround, ideally `ShouldHandleEvent()` would have returned return `false` for those events.

Additionally, the problem with logging is that multiple handlers (`Keyboard`, `Joystick`) try to handle the same event, but only one can handle it -- resulting in the other logging the event as unhandled. Logging of unhandled event should probably only happen after all handlers have been exhausted (`AndroidGameView` is the place where this might make sense to put).

peppy: Okay, given that this is local in the android project i'm not going to dwell on either issue for now.

peppy (approved):

PR #5444, opened 2022-10-04
Copy DotSettings `CodeInspection`s from osu! sln

Susko3:
Most of the inspections were weaker in o!f, except `UnusedParameter.Local` which lowered from `warning` to `hint`.


osu-framework.sln.DotSettings
```diff
@@ -62,7 +63,7 @@
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=CollectionNeverQueried_002ELocal/@EntryIndexedValue">HINT</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=CommentTypo/@EntryIndexedValue">HINT</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=CompareOfFloatsByEqualityOperator/@EntryIndexedValue">HINT</s:String>
-	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=ConvertClosureToMethodGroup/@EntryIndexedValue">HINT</s:String>
+	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=ConvertClosureToMethodGroup/@EntryIndexedValue">WARNING</s:String>
```
smoogipoo, line 66: I think this one is intentional, because for high performance scenarios using method groups will allocate the delegate every time whereas a closure won't (as backwards as that sounds).

This was documented here: https://github.com/dotnet/roslyn/issues/5835 and will be fixed in .NET 7. I'd prefer this remains until that point.

smoogipoo (changes requested):

smoogipoo (approved):

PR #5450, opened 2022-10-07
Refactor borderless test for component reuse and to show more info

Susko3:
The new `WindowDisplaysPreview` might be useful for future tests, so this is a preemptive refactor.

Before:
image

After:
image

### DIFF
--- a/osu.Framework.Tests/Visual/Platform/TestSceneBorderless.cs
+++ b/osu.Framework.Tests/Visual/Platform/TestSceneBorderless.cs
@@ -154,6 +154,7 @@ namespace osu.Framework.Tests.Visual.Platform
                             BorderThickness = 20,
                             Masking = true,
                             Depth = -10,
+                            Alpha = 0.7f,
                             Children = new Drawable[]
                             {
                                 new Box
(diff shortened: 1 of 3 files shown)


osu.Framework.Tests/Visual/Platform/TestSceneBorderless.cs
```diff
@@ -104,59 +57,27 @@ public TestSceneBorderless()
...
             }
 
-            window.DisplaysChanged += onDisplaysChanged;
-            refreshScreens(window.Displays);
-
             const string desc2 = "Check whether the window size is one pixel wider than the screen in each direction";
 
             Point originalWindowPosition = Point.Empty;
 
+            AddStep("nothing", () => { });
```
frenzibyte, line 79: we probably want to change how test scenes should behave on first run, but at least add the comment so that this is not mistaken for leftover/debug code.
```suggestion
            // so the test case doesn't change fullscreen size just when you enter it
            AddStep("nothing", () => { });
```

frenzibyte (commented):
Only have one display currently but seems fine with manually written specifications.

peppy (approved):

Issue #5131, opened 2022-04-25
Initial window size is incorrect on windows with >100% dpi


peppy:
Not considering scaling at the moment, should be.

frenzibyte:
This might come back to https://github.com/libsdl-org/SDL/issues/2119, which fortunately has a [recent PR](https://github.com/libsdl-org/SDL/pull/5393) adding proper DPI support.

PR #5452, opened 2022-10-08
Fix window size properties not correct on Hi-DPI screens

frenzibyte:
Came across it while checking out https://github.com/ppy/osu-framework/pull/5450. Testing with the M1 Pro's retina screen.

Before:

CleanShot 2022-10-08 at 15 15 40@2x

After:

CleanShot 2022-10-08 at 15 13 56@2x


smoogipoo:
Missing some code here

osu.Framework/Platform/SDL2DesktopWindow_Windowing.cs
```diff
@@ -542,10 +542,10 @@ private void updateWindowStateAndSize()
             switch (windowState)
             {
                 case WindowState.Normal:
-                    Size = (sizeWindowed.Value * Scale).ToSize();
+                    Size = sizeWindowed.Value;
 
                     SDL.SDL_RestoreWindow(SDLWindowHandle);
-                    SDL.SDL_SetWindowSize(SDLWindowHandle, sizeWindowed.Value.Width, sizeWindowed.Value.Height);
+                    SDL.SDL_SetWindowSize(SDLWindowHandle, Size.Width, Size.Height);
```
Susko3, line 548: I would expect `FrameworkSetting.WindowedSize` / `sizeWindowed` to match `ClientSize` (as that's exposed in `IWindow` -- `Size` is specific to `SDL2DesktopWindow`).

Have not tested this, not sure how it behaves on startup.

Also https://github.com/ppy/osu-framework/issues/5131 is something to keep in mind.

Susko3 (changes requested):

osu.Framework/Platform/SDL2DesktopWindow_Windowing.cs
frenzibyte: This PR shouldn't change behaviour so it's out of scope here, but I believe it's correct to make `WindowedSize` based on window size rather than client size, specifically in a scenario where a window is displayed on a 200% dpi display, but the window was moved back to a 100% dpi display (in which case, if `WindowedSize == ClientSize`, then the window will display 2x its size on the normal display, unlike `WindowedSize == Size`).

`IWindow` already explicitly names the client size as `ClientSize`, but if required, it can expose the `Size` property as `WindowSize`.

Susko3 (approved):

peppy (approved):

PR #5458, opened 2022-10-08
Make `updateWindowStateAndSize()` and accompanying methods stateless in regards to `WindowState`, `CurrentDisplay` and `CurrentDisplayMode`

Susko3:
Depends on:
- #5456
- #5457

This PR changes the flow to pass around `WindowState`, `CurrentDisplay` and `CurrentDisplayMode` instead of querying the window properties.

In the future, this will allow us to request a specific state without modifying the state bindables (and potentially causing unwanted side-effects because of the bound actions).


osu.Framework/Platform/SDL2DesktopWindow_Windowing.cs
```diff
@@ -613,20 +613,20 @@ private void fetchDisplayMode()
             }
         }
 
-        private void fetchMaximisedState()
+        private void fetchMaximisedState(WindowState windowState)
```
peppy, line 616: I'm not sure I understand the point of this when the input parameter is at the class level. I guess it will make more sense in a follow-up PR, based on the description?

Susko3: The follow-up PR will introduce the following, where passing in `Display` is required:

```cs
        private void updateDisplay(int displayIndex)
        {
            if (tryGetDisplayFromSDL(displayIndex, out Display display))
                // important that we don't set CurrentDisplayBinable.Value
                // because we can't be sure that that display will be the window ends up on.
                // or if the display's resolution, bounds, etc. will change.
                updateWindowStateAndSize(windowState, display, currentDisplayMode.Value);

            fetchCurrentDisplay(); // this will set CurrentDisplayBinable.Value
        }
```

The others (`WindowState`, `DisplayMode`) are not needed, but will be useful in a similar way in the future:

- request a DisplayMode -> find closest `SDL_DisplayMode` -> try to set it -> `fetchDisplayMode()` to get the actual mode that was set.

peppy: I'll take your word for it for now. But let me add that I think things may read better if as many of these methods can be converted to `static` and the result can be stored externally.

ie. I would prefer `static Point? readWindowPositionFromConfig()`

peppy (approved):

PR #5459, opened 2022-10-08
Fix changing displays working only in windowed

### LINKED ISSUE #5055: Changing displays via `CurrentDisplayBindable` doesn't work

Susko3:
- Closes #5055

Depends on:
- #5450
- #5458


Susko3:
`TestSceneBorderless` passes on both `master` and this PR. Ditto for `TestSceneFullscreen` (although this one fails on `window position updated` if the window is on the second display, can be fixed by changing the assertion to `window.CurrentDisplayBindable.Value.Bounds.Location`).

peppy (approved):

PR #5460, opened 2022-10-09
Allow blocking certain actions in `KeyBindingContainer`

frenzibyte:
Simplest way to allow osu! to keep smoke with "Relax" mod enabled, see https://github.com/ppy/osu/pull/20665#issue-1402274358 for reasoning. Not sure if test coverage is necessary, but can add one on request.

### DIFF
--- a/osu.Framework/Input/Bindings/KeyBindingContainer.cs
+++ b/osu.Framework/Input/Bindings/KeyBindingContainer.cs
@@ -47,6 +47,14 @@ namespace osu.Framework.Input.Bindings
         /// </summary>
         public IEnumerable<T> PressedActions => pressedActions;
 
+        /// <summary>
+        /// Actions which are blocked from activating via keybindings.
+        /// </summary>
+        /// <remarks>
+        /// Triggering actions via <see cref="TriggerPressed"/> bypasses this list.
+        /// </remarks>
+        protected virtual IEnumerable<T> BlockedActions => Enumerable.Empty<T>();
+
         private readonly Dictionary<IKeyBinding, List<Drawable>> keyBindingQueues = new Dictionary<IKeyBinding, List<Drawable>>();
         private readonly List<Drawable> queue = new List<Drawable>();
         private List<Drawable> keyRepeatInputQueue;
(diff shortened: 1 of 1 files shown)


bdach:
Do you mind explaining the root cause of the issue? This feels like a weird change to make to fix a game-side issue without a sufficient explanation.

frenzibyte:
I've mentioned the summary of it in OP, but here are the details; osu! currently blocks all UI events from `KeyBindingContainer`. Having it block hit buttons but still allow smoke doesn't look to be feasible with the `KeyBindingContainer` structure as it stands. 

I've initially went with overriding `PropagatePressed` and blocking there, but that comes with its own issues (mainly `TriggerPressed` will no longer work, so relax mod won't be able to hit any object anymore).

This left me with the only other option, which is checking each and every UI event and block any which match up with the bindings assigned to `OsuAction.{Left,Right}Button`. And that's definitely not a valid direction and is very prone to error the moment a new UI event pops in.

Therefore I've went ahead and exposed this `BlockedActions` list which should correctly block any keys from activating an action without causing `TriggerPressed` to be ineffective (first approach) or writing up complicated logic locally.

bdach:
I'm sorry but the OP leaves _a lot_ of pieces of the picture to be filled out by the reader as you had it.

> osu! currently blocks all UI events from `KeyBindingContainer`

You mean [this thing](https://github.com/ppy/osu/blob/9bb4c70393e5a5c6ba92db46b37982d664816ba0/osu.Game.Rulesets.Osu/OsuInputManager.cs#L20-L23), correct?

> This left me with the only other option, which is checking each and every UI event and block any which match up with the bindings assigned to `OsuAction.{Left,Right}Button`

I dunno, this seems to me like a *better* direction if anything from a layering standpoint. The blocking is a game-local behaviour, so why add framework-level exemptions for it?

If you're worried about letting slip potential new gameplay actions, then why not flip the check and block everything that *isn't* a smoke keypress?

frenzibyte:
The osu!-side PR should connect the dots and show the difference between framework-side changes and local overrides.

frenzibyte:
Closing as not required anymore, see https://github.com/ppy/osu/pull/20665#issuecomment-1272544360.

PR #5461, opened 2022-10-10
Fix release configuration not building

Susko3:
https://discord.com/channels/188630481301012481/589331078574112768/1028988858345459752

### DIFF
--- a/osu.Framework/Platform/SDL2DesktopWindow_Windowing.cs
+++ b/osu.Framework/Platform/SDL2DesktopWindow_Windowing.cs
@@ -319,8 +319,8 @@ namespace osu.Framework.Platform
                 {
                     if (tryGetDisplayFromSDL(i, out Display? display))
                         yield return display;
-
-                    Debug.Fail($"Failed to retrieve display at index ({i})");
+                    else
+                        Debug.Fail($"Failed to retrieve display at index ({i})");
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


osu.Framework/Platform/SDL2DesktopWindow_Windowing.cs
```diff
@@ -313,8 +314,9 @@ private IEnumerable<Display> getSDLDisplays()
         {
             return Enumerable.Range(0, SDL.SDL_GetNumVideoDisplays()).Select(i =>
             {
-                Debug.Assert(tryGetDisplayFromSDL(i, out var display));
-                return display;
+                bool success = tryGetDisplayFromSDL(i, out var display);
+                Debug.Assert(success);
+                return display.AsNonNull();
```
frenzibyte, line 319: Now displays will be returned regardless of whether they pass or fail? wouldn't that result in a NRE later on after the failure occurred? Would make sense to just not include null displays rather than using `AsNonNull`:
```csharp
            return Enumerable.Range(0, SDL.SDL_GetNumVideoDisplays()).Select(i =>
            {
                if (!tryGetDisplayFromSDL(i, out Display? display))
                {
                    Debug.Fail($"Failed to retrieve display at index ({displayIndex})");
                    return null;
                }

                return display;
            }).Where(d => d != null).ToArray();
```

Susko3: The idea of the assert is that SDL will not return an error when passing in a valid index (guaranteed by `SDL_GetNumVideoDisplays`).

But I guess your suggestion is safer and reads better.

peppy (approved):
Guess this works..

PR #5463, opened 2022-10-13
Fix colour picker background not rendering correctly on mobile platforms

frenzibyte:
- Closes https://github.com/ppy/osu/issues/14383

Follows a similar solution to what https://github.com/ppy/osu-framework/pull/5065 did. See https://github.com/ppy/osu-framework/pull/5065#discussion_r825536715 for detailed explanation. Renders correctly on iOS now.

### DIFF
--- a/osu.Framework/Resources/Shaders/sh_SaturationSelectorBackground.fs
+++ b/osu.Framework/Resources/Shaders/sh_SaturationSelectorBackground.fs
@@ -1,13 +1,13 @@
 #include "sh_Utils.h"
 
-varying mediump vec2 v_TexCoord;
-varying mediump vec4 v_TexRect;
+varying highp vec2 v_TexCoord;
+varying highp vec4 v_TexRect;
 
 uniform mediump float hue;
 
 void main(void)
 {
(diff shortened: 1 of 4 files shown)


Susko3 (approved):
Works fine on Android (11) as well.

osu.Framework/Resources/Shaders/sh_SaturationSelectorBackground.fs
```diff
@@ -1,13 +1,13 @@
...
+varying highp vec2 v_TexCoord;
+varying highp vec4 v_TexRect;
 
 uniform mediump float hue;
 
 void main(void)
 {
-    vec2 resolution = vec2(v_TexRect[2] - v_TexRect[0], v_TexRect[3] - v_TexRect[1]);
-    vec2 pixelPos = v_TexCoord / resolution;
+    highp vec2 resolution = vec2(v_TexRect[2] - v_TexRect[0], v_TexRect[3] - v_TexRect[1]);
```
bdach, line 10: While we're here, opportunity for swizzle:

```suggestion
    highp vec2 resolution = v_TexRect.zw - v_TexRect.xy;
```

(same goes for rounded variant)

smoogipoo (approved):

PR #5465, opened 2022-10-13
Implement antialiasing for CircularProgress shader

EVAST9919:
Few caveats:
* **There's probably a better way to get texel size**
* **This aa works the best for 1:1 aspect ratio.** The higher it is - the less smooth progress gets. _(which can probably be fixed by adding some more math, but that's for another day, still better than nothing)_
* To make progress appear smooth when `radius < texelSize` I'm leaving thickness the lowest value possible (at which it still appears smooth) and making the whole progress more transparent. Looks good enough imo.
* You can check how smoothening works by increasing `1.5f` value in the `Update()`


smoogipoo:
@frenzibyte / @bdach / @peppy able test this on mobile?

bdach (approved):
Android looks fine.

smoogipoo:
I'm not sure of the correct maths here, but my original thinking was you could get the "texel size" by using the derivative functions. I had in mind something like:

```
texelSize = max(abs(dFdx(pixelPos.x)), abs(dFdy(pixelPos.y)));
```

I'm not sure if `pixelPos` is correct to use but seems to give similar results, though I don't have an appropriate setup to test right now.

EVAST9919:
> I'm not sure of the correct maths here, but my original thinking was you could get the "texel size" by using the derivative functions

Your proposal works great and solves the issue with non 1:1 aspect ratios!

PR #5466, opened 2022-10-14
Add more quick way to click a drawable using `ManualInputManager`

Feodor0090:
To emulate input in visual test, `TriggerClick` is used in most cases. It's okay if we test 2-3 buttons component, but may produce false-positive tests in more complex cases.
Examples:
- It bypasses `LoadingLayer`
- https://github.com/ppy/osu/pull/18761

`TriggerClick` is used everywhere because it's simple:
```
AddStep("Click", () => button.TriggerClick());
```
Example with manual input:
```
AddStep("Click", () =>
{
    InputManager.MoveMouseTo(button);
    InputManager.Click(MouseButton.Left);
});
```
Now:
```
AddStep("Click", () => InputManager.Click(button));
```
With this shortcut, using manual input is just several character longer than triggering event directly, and it reads better than before.

### DIFF
--- a/osu.Framework/Testing/Input/ManualInputManager.cs
+++ b/osu.Framework/Testing/Input/ManualInputManager.cs
@@ -140,10 +140,21 @@ namespace osu.Framework.Testing.Input
         }
 
         /// <summary>
+        /// Clicks the drawable with the specified button.
+        /// </summary>
+        /// <param name="drawable">Drawable to click.</param>
+        /// <param name="button">Mouse button to use.</param>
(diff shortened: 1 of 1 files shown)


frenzibyte:
Rather than working around this by inventing a separate path for clicking a drawable, `TriggerClick()` should be fixed instead by properly going through an input queue and propagating clicks until one handles it. See https://github.com/ppy/osu/issues/14132.

peppy:
I tend to agree. Fixing the existing one makes much more sense. Adding yet another click method will just cause confusion as to which should be called.

PR #5471, opened 2022-10-20
Support iOS "discrete" (mouse wheel) scroll input

frenzibyte:
Been hanging on my list for a good while now. Have been testing with iOS simulator as I don't have the required accessory to connect a mouse to my phone. @peppy would appreciate testing on a physical device to ensure the logic works as expected, I didn't expect "translation" points to be a thing in discrete scroll type, but iOS simulator says otherwise.


peppy (approved):

peppy:
Closing as dead, we've moved to SDL3 and this class doesn't exist.

PR #5472, opened 2022-10-21
Optimise `ColourInfo` conversion and add conversion helper method

smoogipoo:
The performance gains here are mostly theoretical and I can't benchmark them meaningfully in a real-world scenario. Consider this more of a refactoring that adds an optimal (and more user-friendly) method for checking whether a `ColourInfo` represents a single colour and then returning that colour.

This comes off the back of the osu!-side smoke implementation (where I initially noticed this), which potentially uses this method for tens of thousands of vertices (up to 30k is what I was able to achieve after [the recent changes](https://github.com/ppy/osu/pull/20827)). Similar is also done with slider paths (e.g. at some point https://osu.ppy.sh/beatmapsets/756794#osu/1605148 draws some 20k vertices). So _at most_ you'd see ~0.2ms frametime improvement.

Before
|              Method |     Mean |    Error |   StdDev | Allocated |
|-------------------- |---------:|---------:|---------:|----------:|
| ConvertToSRGBColour | 14.31 ns | 0.139 ns | 0.123 ns |         - |
|     ConvertToColor4 | 15.69 ns | 0.122 ns | 0.102 ns |         - |

After:
|                    Method |       Mean |     Error |    StdDev | Allocated |
|-------------------------- |-----------:|----------:|----------:|----------:|
|       ConvertToSRGBColour |  6.9628 ns | 0.0684 ns | 0.0534 ns |         - |
|           ConvertToColor4 | 11.1604 ns | 0.1658 ns | 0.1551 ns |         - |
| ExtractAndConvertToColor4 |  0.4039 ns | 0.0247 ns | 0.0231 ns |         - |

It uses concepts I've learnt over the years:
- Methods with `throw` are not inlined - extract these methods to promote inlining.
- Having strings in methods, even if the code is not hit, leads to a significant increase in ASM putting pressure on the JIT.

For comparisons sake, this is the best "real-world" example I could make, by looping `TestSceneSmoke` over 30 seconds:
| before | after |
| --- | --- |
| a | b |

Again, I don't think these are indicative of anything so don't put much weight to them. There's too much noise.

### DIFF
--- a/osu.Framework/Graphics/Colour/ColourInfo.cs
+++ b/osu.Framework/Graphics/Colour/ColourInfo.cs
@@ -86,6 +86,17 @@ namespace osu.Framework.Graphics.Colour
             }
         }
 
+        /// <summary>
+        /// Attempts to extract the single colour represented by this <see cref="ColourInfo"/>.
+        /// </summary>
+        /// <param name="colour">The extracted colour. If <c>false</c> is returned, this represents the top-left colour.</param>
+        /// <returns>Whether the extracted colour is the single colour represented by this <see cref="ColourInfo"/>.</returns>
+        public readonly bool TryExtractSingleColour(out SRGBColour colour)
+        {
+            colour = TopLeft;
+            return HasSingleColour;
+        }
+
         public readonly SRGBColour Interpolate(Vector2 interp) => SRGBColour.FromVector(
             (1 - interp.Y) * ((1 - interp.X) * TopLeft.ToVector() + interp.X * TopRight.ToVector()) +
             interp.Y * ((1 - interp.X) * BottomLeft.ToVector() + interp.X * BottomRight.ToVector()));
--- a/osu.Framework/Graphics/Colour/ColourInfo.cs
+++ b/osu.Framework/Graphics/Colour/ColourInfo.cs
@@ -71,13 +71,8 @@ namespace osu.Framework.Graphics.Colour
         {
             readonly get
             {
-                if (!HasSingleColour)
-                    throwConversionFromMultiColourToSingleColourException();
-
+                Debug.Assert(HasSingleColour);
                 return TopLeft;
-
-                [DoesNotReturn]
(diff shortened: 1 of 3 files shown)


peppy:
Tests not happy 😢

osu.Framework/Graphics/Colour/ColourInfo.cs
```diff
@@ -227,7 +236,17 @@ public void ApplyChild(ColourInfo childColour, Quad interp)
...
+
+        public static implicit operator SRGBColour(ColourInfo colour)
+        {
+            if (!colour.HasSingleColour)
+                throwConversionFromMultiColourToSingleColourException();
+
+            return colour.singleColour;
+
+            [DoesNotReturn]
+            static void throwConversionFromMultiColourToSingleColourException() => throw new InvalidOperationException("Attempted to read single colour from multi-colour ColourInfo.");
```
turbedi, line 248: There already exists a throw helper for InvalidOperationException [here](https://github.com/ppy/osu-framework/blob/8777b5736e18db8cf95ddd0e47beff99c147622e/osu.Framework/Utils/ThrowHelper.cs#L15).
Maybe we can use this one?

frenzibyte: While this is supposed to be a correct use case of that method, because it causes a [string construction](https://github.com/ppy/osu-framework/pull/5392#discussion_r963708994) in the conversion method, which is what's being avoided in this PR.

smoogipoo:
Fixed tests, the assertion was added post-fact.

osu.Framework/Graphics/Colour/ColourInfo.cs
smoogipoo: Technically it's not a string construction, but rather the code constructing the string results in a lot of dead ASM code for the JIT to compile that could also potentially cause the function to not be inlined.

peppy (approved):

PR #5473, opened 2022-10-21
Make `getClosestDisplayMode()` static

Susko3:
This also fixes a merge error in this function, where it would use the `displayIndex` and `currentDisplay` fields instead of the provided `Display`.


bdach:
"Merge error"? Does this mean that the latest framework release needs to be pulled?

Susko3:
It's a little incorrect, but the passed in values appear to match the values in the fields of `SDL2DesktopWindow`.

peppy:
The fact that I can incorrectly resolve merge conflicts and not notice should hopefully show that the logic is very painful to understand still.

peppy (approved):

PR #5476, opened 2022-10-23
Update SDL2

peppy:


bdach:
iOS needs some attention:

```
 /home/runner/work/osu-framework/osu-framework/osu.Framework.iOS/osu.Framework.iOS.csproj : error NU1605: Detected package downgrade: SixLabors.ImageSharp from 2.1.3 to 2.1.0. Reference the package directly from the project to select a different version.  [/home/runner/work/osu-framework/osu-framework/osu-framework.sln]
/home/runner/work/osu-framework/osu-framework/osu.Framework.iOS/osu.Framework.iOS.csproj : error NU1605:  ppy.osu.Framework.iOS -> ppy.osu.Framework -> SixLabors.ImageSharp (>= 2.1.3)  [/home/runner/work/osu-framework/osu-framework/osu-framework.sln]
/home/runner/work/osu-framework/osu-framework/osu.Framework.iOS/osu.Framework.iOS.csproj : error NU1605:  ppy.osu.Framework.iOS -> SixLabors.ImageSharp (>= 2.1.0) [/home/runner/work/osu-framework/osu-framework/osu-framework.sln]
```

Probably salient to run safety checks on a device as well. That I cannot do.

peppy:
Yeah this isn't going to happen, we have stuff setup in a very delicate way and this issue just tips things over the edge.

I'll document what I found along the way in case it's useful to anyone, but I recommend we just don't update dependencies until we move forward to non-xamarin toolchain.

Branch with props updated: https://github.com/ppy/osu-framework/compare/master...peppy:osu-framework:update-libraries-fucked?expand=1

Causes a conflict on `System.Collections.Immutable` which can only be fixed by adding the following line to the `csproj` file (not the `props`, weirdly):

JetBrains Rider-EAP 2022-10-24 at 03 52 13

Turns out we aren't including the iOS `.props` file in the main project (and are using it to downgrade versions of certain libraries at the final assemblies only, ie. non-transitive projects). Including it causes new conflicts as they are seen as incorrectly downgrading the library.

TL;DR you can make it work, but it will probably require changes to osu-side projects (and framework consuming projects) so let's keep things working as they are. I've rolled back all but the SDL changes, as I'm not really sure where the immutable dependency actually comes from.

smoogipoo (approved):

PR #5478, opened 2022-10-24
Fix directory selector crashing when attempting to display a bitlocker locked drive

### LINKED ISSUE #5477: Game crashing on file selectors due to secondary drive being encrypted by bitlocker
Found it when I tried running perfcalcgui on a laptop with one drive encrypted.

image
image

This seems to be related to #5306 and https://github.com/ppy/osu/discussions/19502

peppy:
Closes #5477. Untested.


smoogipoo (approved):

smoogipoo:
Should probably display this error somewhere/give the error somewhere for a consumer to handle and display it, but probably fine for now?

bdach (approved):

PR #5479, opened 2022-10-24
Fix texture uploads potentially disposed too early

smoogipoo:
Fixes https://github.com/ppy/osu/issues/7477

These texture uploads may be crucial for the few frames before the texture's usage has concluded.


peppy (approved):

smoogipoo:
Ping @frenzibyte to port this to your Veldrid branch.

PR #5482, opened 2022-10-24
Rewrite `IGraphicsBackend` to support multiple backends

frenzibyte:
Major step towards supporting Veldrid, mainly allowing for the renderer to interact with the graphics side of the window for backend-specific functions (such as GL swapping or creating metal-backend layer, which require SDL calls) without resorting to hardcoded calls.

This graphics side of the window has been defined using an interface named `IWindowGraphics` (name tbd), providing main functions shared across all backends (e.g. drawable size). With backend-specific functions residing each in its own interface, and explicitly implemented in the `IWindowGraphics` implementation, requiring each renderer to cast the `IWindowGraphics` to said interface for calling its functions.

This becomes used in `VeldridRenderer` (in addition to the `GLRenderer` usage in this PR) as such:
```csharp
                    var metalGraphics = graphics as IMetalWindowGraphics ?? throw new InvalidOperationException($"Window graphics API must implement {nameof(IMetalWindowGraphics)}.");
                    swapchain.Source = SwapchainSource.CreateNSView(metalGraphics.CreateMetalView());
                    break;
```
```csharp
                case GraphicsBackend.OpenGL:
                    var openGLGraphics = graphics as IOpenGLWindowGraphics ?? throw new InvalidOperationException($"Window graphics API must implement {nameof(IOpenGLWindowGraphics)}");

                    Device = GraphicsDevice.CreateOpenGL(options, new OpenGLPlatformInfo(
                        openGLContextHandle: openGLGraphics.WindowContext,
                        getProcAddress: openGLGraphics.GetProcAddress,
                        makeCurrent: openGLGraphics.MakeCurrent,
                        getCurrentContext: () => openGLGraphics.CurrentContext,
                        clearCurrentContext: openGLGraphics.ClearCurrent,
                        deleteContext: openGLGraphics.DeleteContext,
                        swapBuffers: openGLGraphics.SwapBuffers,
                        setSyncToVerticalBlank: v => openGLGraphics.VerticalSync = v), swapchain.Width, swapchain.Height);

                    Device.LogOpenGL(out maxTextureSize);
                    break;
```

In addition, the `IWindowGraphics` (SDL2 implementation specifically) requires context of the current graphics backend to initialise with (mainly for OpenGL), and that's done using an introduced `GraphicsBackend` enum, but is currently going through an awkward back and fourth between the window side and renderer side. To explain it in detail:
 1. The renderer is created via `CreateRenderer()`, and provides the backend type it supports via `IRenderer.BackendType`
 2. The window and the graphics backend are created with knowledge of the renderer's backend type, i.e. `Window = CreateWindow(Renderer.BackendType)`
 3. The renderer is then *initialised* with the instance of the window's graphics backend, i.e. `Renderer.Initialise(Window.Graphics)`

One way to get around this back-and-fourth awkwardness would be to move the graphics backend property to `GameHost`, and ensure both the created renderer supports it and the graphics backend works with it. I haven't went with that approach since now a consumer can either override `BackendType` to set the graphics backend, or override `CreateRenderer` to create the renderer, and I'm unsure on how we want that to be going forward, maybe best to have it in `GameHost`? or even a framework setting while at it? @smoogipoo interested in your opinion.

- [ ] Needs testing on osuTK (checking in a bit...)


smoogipoo:
Need more time to think about this/to experiment with this, as I'm finding the class coupling here quite hard to follow at a first glance. As I understand it:
- The window class creates the window graphics.
- The window graphics then (e.g. in the case of SDL2WindowGraphics) uses the `IRenderer` to check the renderer type (opengl/metal) for specific graphics calls.
- The `IRenderer` uses the window graphics for a few of its internal functions like `MakeCurrent`.

I understand why it's done this way, and it sort of makes sense, but also leads to some pretty weird code as I was hoping to not have that enum.

frenzibyte:
That's precisely what I touched on at the end of the PR's description, have you read that?

smoogipoo:
I skimmed through it and missed that :sweat_smile:, went straight into code to understand things. Good to see we're on the same page.

frenzibyte:
@smoogipoo what are your thoughts on the approach I suggested in OP (moving it to `GameHost` as an overridable property, or a `FrameworkSetting` with default values for each platform). I see the enum existing either way since we'll probably want it exposed to the user for changing the renderer used for the game they are running.

smoogipoo:
I'm still reviewing this and I don't agree with that right now. I'm instead looking for ways to remove that enum or make it internal to the Veldrid implementation.

smoogipoo:
I'm going to close this since it's easier to review with everything together in https://github.com/ppy/osu-framework/pull/5483

Issue #4909, opened 2021-11-22
Add Veldrid renderer support (aka "compatibility mode")


peppy:
We don't yet have good backwards compatibility support for older hardware, where the hardware (or drivers) don't support the OpenGL standards we rely on.

This covers adding the same level of compatibility that we have in osu-stable. It should preferably be an automatic fallback whenever opengl fails to initialise. 

---

Support progress:

- [x] Bare-minimum `IRenderer` implementation (https://github.com/ppy/osu-framework/pull/5483)
- [x] Vertex buffers support (https://github.com/ppy/osu-framework/pull/5595)
- [ ] Shaders support (https://github.com/ppy/osu-framework/pull/5647)
- [ ] Textures support (https://github.com/ppy/osu-framework/pull/5647)
- [ ] Framebuffers support (https://github.com/ppy/osu-framework/pull/5647)
- [ ] Video textures support (https://github.com/ppy/osu-framework/pull/5647)
- [ ] Handle backend-specific differences (https://github.com/ppy/osu-framework/pull/5647)
- [ ] Screenshot rendering support (https://github.com/ppy/osu-framework/commit/8ad7ed6761943519501e8179edbc8f68d4d6c1bb, wip, doesn't fully work)
- [ ] Clean-up remaining GL calls (e.g. translation functions in `BlendingParameters`)

huupoke12:
I doubt if this is even practical to do anymore, since those hardware probably can't even run Windows 8.1 (no driver support). And .NET 6 only supports Windows 7 ESU, which probably only be purchased in enterprise environments.

smoogipoo:
We won't be using ANGLE. We'll be using Veldrid and have native D3D/Vulkan support.

peppy:
Also to keep in mind is that this isn't about "hardware older than Windows 8.1", but hardware/drivers which support D3D better than OpenGL.

smoogipoo:
Closing in favour of https://github.com/ppy/osu/issues/18892

PR #5483, opened 2022-10-24
Add bare-minimum Veldrid renderer implementation

frenzibyte:
- [x] Depends on #5482 

Veldrid support, here we go! First piece out of yet to come in fulfilling https://github.com/ppy/osu-framework/issues/4909 and bringing much better performance to osu!

This PR brings the bare minimum of Veldrid-based `IRenderer` implementation, excluding any pieces specific to textures/framebuffers/shaders/etc., those will come each in its own PR for review ease.

For testing whether Veldrid works, switch the renderer in `GameHost.CreateRenderer` and you should see the window displaying a background cycling colours over time (temporary until shaders support is in place).

Few concerns to note here:
 - I was initially going with the idea of having one `VeldridRenderer` implementation handling different graphics backends, but later on I realised I could instead split the implementation to different renderer classes based on the backend (i.e. have `D3D11Renderer`, `VulkanRenderer`, and `MetalRenderer` all inheriting from `VeldridRenderer` with a virtual initialisation method to override). Switching to different renderer classes per backend won't be that hard, so can go with that direction if it sounds better (and avoid the "Log" extension methods).
 - Linux platform will not work right now, need to figure out how to choose the right swapchain source (use x11 or use wayland), and I'm not sure whether that should be done using `RuntimeInfo` or by exposing a window subsystem enum in `IWindow`/`IWindowGraphics`.
 - Mobile platforms won't work for the time being, SDL support is a requisite (might have a stab on iOS sometime later, simulator has been driving me nuts).
 - Renderer initialisation point has been moved to the window thread, as I've came to realise after spending many hours that Veldrid for some odd reason breaks horribly when initialising on a thread away from the window thread, and it appears to be macOS specific, and potentially M1 specific. OpenGL (`GLRenderer`) still works fine on my end with the initialisation happening on window thread, but testing on other platforms would be appreciated.


smoogipoo:
As far as my testing on this branch goes:

- The struct that's passed into SDL to determine the pointers is missing the SDL version.
- X11/Wayland isn't supported as commented in code, so I've forced it to X11 locally.
- On Vulkan, if I draw above a certain FPS, Veldrid throws an out of range exception internally. Forcing swapbuffers/finish doesn't seem to fix.

Those aren't big issues and I'm not interested in fixing them to make this branch ready to go, just things of note. I looked at this branch primarily to get an idea of how things fit together in a final product, not to see if it works perfectly or is even runnable on Vulkan.

Now I'm taking a step back and looking at the structure of things in relation to https://github.com/ppy/osu-framework/pull/5482. As both you and I mentioned, I think the cross-communication of the `GraphicsBackend` is too much.

My eventual "goal" (whether it's something we want to allow or not) is to have a `GraphicsSurfaceType` enum `{ OpenGL, Vulkan, D3D, Metal }` and a function to create the `IRenderer`, both in the `HostOptions` class that the `GameHost` is initialised with.

The window is created with a preferred `GraphicsSurfaceType` value. It can choose to handle this however it wants, whether by forcing a surface type or throwing an exception for unsupported values.

The window creates an `IGraphicsSurface` which is the same as your `IWindowGraphics` class.

The renderer can also throw an exception if it doesn't support the window's surface.

I've applied the changes above in https://github.com/smoogipoo/osu-framework/tree/graphics-surface-wip . Please check it and if you're fine with it I'll merge it back into this branch.

smoogipoo (approved):
I think this is in a good enough state that I'm happy to get it in as is and fix the rest later :partying_face: Nice work.

I'm not super sure on the placement of VerticalSync/MakeCurrent/ClearCurrent/SwapBuffers, and am tempted to say there could be an `SDL2VeldridGraphicsSurface` handling it, but it's internal and others think it makes more sense as is.

frenzibyte:
I've taken a look at the changes and the new structure/naming. It looks pretty solid now, I like it! I'll start off by PR'ing each of the pieces I already have locally, and see where we go from there.

PR #5485, opened 2022-10-27
Implement Blob component

EVAST9919:
DrawNode logic is very similar to `CircularProgress`.
In the shader `pointCount` and `search range` may be probably adjusted further to improve performance even more but visuals may suffer for blobs with thick borders due to how approximation works there.
Test scene has been taken from `CircularProgress` and probably needs to be abstracted later.


smoogipoo:
I like that this is implemented as a separate shader, was scared it would get bunched into circular progress and that shader would become a hydra of sorts :)

osu.Framework/Resources/Shaders/sh_BlobUtils.h
```diff
@@ -0,0 +1,59 @@
...
+}
+
+lowp float blobAlphaAt(highp vec2 pixelPos, mediump float innerRadius, highp float texelSize, mediump float frequency, mediump float amplitude, int seed)
+{    
+    // Compute angle of the current pixel in the (0, 2*PI) range
+    mediump float pixelAngle = atan(0.5 - pixelPos.y, 0.5 - pixelPos.x) - HALF_PI;
+    if (pixelAngle < 0.0)
+        pixelAngle += TWO_PI;
+
+    int pointCount = 50;
```
smoogipoo, line 33: This should probably be a `const int` instead. In older GLSL versions (< 4.00) there's no guarantee of these being "dynamically uniform" and from experience Intel drivers specifically do minimal optimisations.

osu.Framework/Graphics/UserInterface/Blob.cs
```diff
@@ -0,0 +1,132 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+#nullable disable
```
smoogipoo, line 4: Remove this + update any relevant code.

smoogipoo (dismissed):

peppy:
Seems to be a bit broken with certain settings, not sure how easy this is to fix but:

osu Framework Tests 2022-10-27 at 03 20 55


https://user-images.githubusercontent.com/191335/198183443-a2f9c1ad-7e7b-458f-94ad-f3a1752ebcf7.mp4

osu.Framework/Graphics/UserInterface/Blob.cs
```diff
@@ -0,0 +1,132 @@
...
+
+using System;
+using osu.Framework.Allocation;
+using osu.Framework.Graphics.Rendering;
+using osu.Framework.Graphics.Shaders;
+using osu.Framework.Graphics.Sprites;
+
+namespace osu.Framework.Graphics.UserInterface
+{
+    public class Blob : Sprite
```
peppy, line 14: I'd probably call this `CircularBlob` or something else. Just `Blob` is too generic and has a different meaning in development terms.

EVAST9919:
> Seems to be a bit broken with certain settings, not sure how easy this is to fix

Yep, this is due to how we are approximating the path: we are plotting `pointCount` on a path within the search range and if it becomes too steep - this is the result.

**Potential solution:** increasing `pointCount` with the frequency, but iterating over potentially 1k or more points hurts the performance a lot.
Another solution would be pr'ing this into the osu-catch project directly because it seems to look fine for the needed task (if it even good enough for you) and not exposing this component for "wide use"

peppy:
I think it's a pretty good example component to have in the framework, but would be okay with either direction. Interested in @smoogipoo's thoughts.

smoogipoo:
I'm fine with it being in o!f.

peppy:
Ignoring the issue mentioned in discord regarding the seed (`RNG.Next()` set to the seed causes all perfect circles; using `RNG.Next(0,200)` to work around this for now) I've updated the osu!catch argon implementation to use this branch and it's looking pretty perfect:


https://user-images.githubusercontent.com/191335/198189931-fc9f60e8-4828-41bb-9e9d-b714a88b49a5.mp4

EVAST9919:
Okay, I don't see the shape issue being fixed with the current implementation, needs a big alg rethinking.
The rest have been addressed so far.

peppy:
```csharp

[runtime] 2022-10-27 09:15:13 [error]: An unhandled error has occurred.
[runtime] 2022-10-27 09:15:13 [error]: osu.Framework.Graphics.OpenGL.Shaders.GLShader+PartCompilationFailedException: A osu.Framework.Graphics.OpenGL.Shaders.GLShaderPart failed to compile: sh_CircularBlob.fs:
[runtime] 2022-10-27 09:15:13 [error]: ERROR: 0:116: '*' does not operate on 'int' and 'float'
[runtime] 2022-10-27 09:15:13 [error]: ERROR: 0:116: '*' does not operate on 'int' and 'float'
[runtime] 2022-10-27 09:15:13 [error]: ERROR: 0:123: Use of undeclared identifier 'noisePosition'
[runtime] 2022-10-27 09:15:13 [error]: ERROR: 0:124: Use of undeclared identifier 'noiseValue'
[runtime] 2022-10-27 09:15:13 [error]: ERROR: 0:125: Use of undeclared identifier 'pos'
[runtime] 2022-10-27 09:15:13 [error]: at osu.Framework.Graphics.OpenGL.Shaders.GLShaderPart.Compile() in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/OpenGL/Shaders/GLShaderPart.cs:line 151
```

peppy (approved):

PR #5487, opened 2022-10-27
Make CircularArcProperties public

OliBomby:
for https://github.com/ppy/osu/pull/20952


peppy:
Bit of a weird API this class has, but I guess it's a very edge use case.

peppy (approved):

smoogipoo (approved):

PR #5488, opened 2022-10-27
Create Keyboard Height Utility

NullifiedJosh:
This utility allows osu! to detect the height of the on-screen keyboard. This is important because when the keyboard covers a text box, it has a bad UX experience.

### DIFF
--- a/osu.Framework.Android/AndroidGameHost.cs
+++ b/osu.Framework.Android/AndroidGameHost.cs
@@ -17,6 +17,7 @@ using osu.Framework.Input.Handlers;
 using osu.Framework.Input.Handlers.Midi;
 using osu.Framework.IO.Stores;
 using osu.Framework.Platform;
+using osu.Framework.Utils;
 using Uri = Android.Net.Uri;
 
 namespace osu.Framework.Android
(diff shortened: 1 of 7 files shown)


NullifiedJosh:
I can see that someone has viewed this PR, but an update would be nice. I just don't want this to get stale state. No rush! 😀

smoogipoo:
This, as well as any mobile PRs in general, are very time consuming for us to review and quite low priority as such.

bdach:
If you don't wish this PR to become stale, then you should probably make it *do something useful*. As far as I can tell from source nothing is using these "keyboard utils" for anything, making all of the code added in this PR essentially dead and untestable.

Then again I don't really agree with the general direction of this PR in the first place. In the ideal implementation to me `DesktopKeyboardUtils` does not exist, period.

NullifiedJosh:
@bdach Do you mean make a project with it? ok.

bdach:
No. I mean the following:

Make these "keyboard utils" automatically work with every single current or future framework project when the software keyboard is pulled up, by using the safe area feature (which is already in place in the framework and works automatically for devices with notches) in order to move the game content away form where the software keyboard is shown.

NullifiedJosh:
I think I would need a bit of help with that.

NullifiedJosh:
I have an idea of how this could work, but it would make a break and change.

PR #5490, opened 2022-10-28
Refactor drawable pool statistics output to make more sense

peppy:
I was confused by the previous output, as it didn't do a great job showing the current pool usage – specifically it was not showing the current pool size in the denominator portion of the display, but the total usages. Which didn't make much sense at all.

The new format is `currentUsage / poolSize (x excess)`. Hopefully this is self-explanatory, so check that your understanding matches the following:

- `currentUsage` includes all current usages that started with a `Get` from the pool, and is decremented on each `Return`. This includes excess usages (that will not be returned to the pool).
- `poolSize` is the current size of the pool, ranging from `initialSize` to `maximumSize`. It is increased at the point of construction, not `Return`
- `excess` is the total excess drawables constructed which will not be returned to the pool. Note that based on the order of `Return`, the drawable which caused this count to increment may not be the one that is not-returned-to-the-queue. That depends on the order that usages finish.

I have tried to not change the existing flow at all. This is intending to only change accounting so we can display the new statistics output.

### DIFF
--- a/osu.Framework/Graphics/Pooling/DrawablePool.cs
+++ b/osu.Framework/Graphics/Pooling/DrawablePool.cs
@@ -115,7 +115,7 @@ namespace osu.Framework.Graphics.Pooling
             {
                 drawable = create();
 
-                if (currentPoolSize < maximumSize)
+                if (maximumSize == null || currentPoolSize < maximumSize)
                 {
                     CurrentPoolSize++;
                     Debug.Assert(maximumSize == null || CurrentPoolSize <= maximumSize);
--- a/osu.Framework/Graphics/Pooling/DrawablePool.cs
+++ b/osu.Framework/Graphics/Pooling/DrawablePool.cs
@@ -204,7 +204,7 @@ namespace osu.Framework.Graphics.Pooling
         private int countExcessConstructed;
 
         /// <summary>
-        /// The total number of drawables constructed.
+        /// The total number of drawables constructed that were not pooled.
         /// </summary>
         public int CountExcessConstructed
         {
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Pooling/DrawablePool.cs
```diff
@@ -97,19 +115,26 @@ public T Get(Action<T> setupAction = null)
             {
                 drawable = create();
 
+                if (currentPoolSize < maximumSize)
+                {
+                    CurrentPoolSize++;
+                    Debug.Assert(maximumSize == null || CurrentPoolSize <= maximumSize);
```
smoogipoo, line 121: Have you actually tested this code with `maximumSize = null`? `x < null` or `x >= null` is guaranteed to be `false`, so as far as I can tell when `maximumSize` is `null` it will be `CountExcess` that is incremented even though the code in `Return()` will push it to the pool (via the same mechanism) and thus it should be `CurrentPoolSize`.

peppy: I'm 99% sure there there is a test that covers this which caused me to add that case in the first place.

smoogipoo: I haven't looked at the tests yet, but if that's the case then the test is missing a case, because this is behaving incorrectly:

image

smoogipoo (changes requested):
As mentioned

osu.Framework/Graphics/Pooling/DrawablePool.cs
peppy: Fixed this case and added test coverage.

smoogipoo (approved):

PR #5491, opened 2022-10-28
CircularBlob shader improvements

EVAST9919:
* `pointCount` is based on input parameters now (in ctb case should become around 10 instead of 50)
* Removed early loop exit as suggested

### DIFF
--- a/osu.Framework/Resources/Shaders/sh_CircularBlobUtils.h
+++ b/osu.Framework/Resources/Shaders/sh_CircularBlobUtils.h
@@ -24,7 +24,7 @@ highp float noise(highp vec2 st)
 }
 
 lowp float blobAlphaAt(highp vec2 pixelPos, mediump float innerRadius, highp float texelSize, mediump float frequency, mediump float amplitude, int seed)
-{    
+{
     // Compute angle of the current pixel in the (0, 2*PI) range
     mediump float pixelAngle = atan(0.5 - pixelPos.y, 0.5 - pixelPos.x) - HALF_PI;
     if (pixelAngle < 0.0)
@@ -32,7 +32,7 @@ lowp float blobAlphaAt(highp vec2 pixelPos, mediump float innerRadius, highp flo
 
     mediump float complexity = (frequency + amplitude) * 0.5 + 1.0;
 
-    mediump float pointCount = 5.0 * complexity;
+    int pointCount = int(ceil(5.0 * complexity));
     mediump float searchRange = 0.1 * complexity; // in radians
 
     mediump float pathRadius = innerRadius * 0.25;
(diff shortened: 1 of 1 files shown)


peppy:
This crashes for me again.

```csharp
[runtime] 2022-10-28 08:08:09 [error]: An unhandled error has occurred.
[runtime] 2022-10-28 08:08:09 [error]: osu.Framework.Graphics.OpenGL.Shaders.GLShader+PartCompilationFailedException: A osu.Framework.Graphics.OpenGL.Shaders.GLShaderPart failed to compile: sh_CircularBlob.fs:
[runtime] 2022-10-28 08:08:09 [error]: ERROR: 0:121: '<' does not operate on 'int' and 'float'
```

Here's a solution

```diff
diff --git a/osu.Framework/Resources/Shaders/sh_CircularBlobUtils.h b/osu.Framework/Resources/Shaders/sh_CircularBlobUtils.h
index dec834f0d..fc03d8972 100644
--- a/osu.Framework/Resources/Shaders/sh_CircularBlobUtils.h
+++ b/osu.Framework/Resources/Shaders/sh_CircularBlobUtils.h
@@ -24,7 +24,7 @@ highp float noise(highp vec2 st)
 }
 
 lowp float blobAlphaAt(highp vec2 pixelPos, mediump float innerRadius, highp float texelSize, mediump float frequency, mediump float amplitude, int seed)
-{    
+{
     // Compute angle of the current pixel in the (0, 2*PI) range
     mediump float pixelAngle = atan(0.5 - pixelPos.y, 0.5 - pixelPos.x) - HALF_PI;
     if (pixelAngle < 0.0)
@@ -32,7 +32,7 @@ lowp float blobAlphaAt(highp vec2 pixelPos, mediump float innerRadius, highp flo
 
     mediump float complexity = (frequency + amplitude) * 0.5 + 1.0;
 
-    mediump float pointCount = 5.0 * complexity;
+    int pointCount = int(5.0 * complexity);
     mediump float searchRange = 0.1 * complexity; // in radians
 
     mediump float pathRadius = innerRadius * 0.25;
@@ -48,7 +48,7 @@ lowp float blobAlphaAt(highp vec2 pixelPos, mediump float innerRadius, highp flo
     // Plot points within a search range and check which one is closest
     for (int i = 0; i < pointCount; i++)
     {
-        mediump float angle = startAngle + searchRange * float(i) / pointCount;
+        mediump float angle = startAngle + searchRange * float(i) / float(pointCount);
         highp vec2 cs = vec2(cos(angle - HALF_PI), sin(angle - HALF_PI));
 
         highp float noiseValue = noise(noisePosition + cs * vec2(frequency));
@@ -56,6 +56,6 @@ lowp float blobAlphaAt(highp vec2 pixelPos, mediump float innerRadius, highp flo
 
         shortestDistance = min(shortestDistance, distance(pixelPos, pos));
     }
-    
+
     return smoothstep(texelSize, 0.0, shortestDistance - pathRadius);
 }

```

peppy (changes requested):
As proposed

peppy:
In testing, it looks like the point count multiplier could potentially be dropped further, but if  so, it also needs to factor in the `innerRadius`, as dropping it too low with a thin radius will cause the line to completely disappear.

Might be worth experimenting further, but I think with the proposed crash fix this is good enough to get in for further performance testing.

peppy (approved):

PR #5493, opened 2022-10-28
Fix `Renderer` holding a permanent reference to every texture ever created

### LINKED ISSUE #5486: `Renderer` holds references to all textures indefinitely
Needs to be weak, or cleaned up at some point:

Parallels Desktop 2022-10-27 at 06 22 10

In most cases, the disposal is explicitly called so this does not add a huge overhead, but it's still an issue.

peppy:
Closes #5486.

### DIFF
--- a/osu.Framework/Graphics/Rendering/Dummy/DummyNativeTexture.cs
+++ b/osu.Framework/Graphics/Rendering/Dummy/DummyNativeTexture.cs
@@ -20,12 +20,7 @@ namespace osu.Framework.Graphics.Rendering.Dummy
         public bool BypassTextureUploadQueueing { get; set; }
         public bool UploadComplete => true;
         public bool IsQueuedForUpload { get; set; }
-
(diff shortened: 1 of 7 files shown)


osu.Framework/Graphics/Rendering/Dummy/DummyNativeTexture.cs
```diff
@@ -21,6 +21,12 @@ internal class DummyNativeTexture : INativeTexture
         public bool UploadComplete => true;
         public bool IsQueuedForUpload { get; set; }
 
+        ulong INativeTexture.TotalBindCount
+        {
+            get => 0;
+            set => throw new System.NotImplementedException();
```
smoogipoo, line 27: Probably shouldn't throw. Maybe `{ }` instead, if it works?

smoogipoo (changes requested):

smoogipoo (approved):

PR #5494, opened 2022-10-29
Implement DrawableShader component

EVAST9919:
More and more components are sharing similar logic so I thought a parent class for them would be a nice thing to have.
Not sure about the name though.


osu.Framework/Graphics/DrawableShader.cs
```diff
@@ -0,0 +1,50 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Allocation;
+using osu.Framework.Graphics.Rendering;
+using osu.Framework.Graphics.Shaders;
+using osu.Framework.Graphics.Sprites;
+
+namespace osu.Framework.Graphics
+{
+    public class DrawableShader : Sprite
```
frenzibyte, line 11: Would call this `ShaderBasedSprite`, but generally, I think having this as a protected constructor in `Sprite` would make sense, or even make this an abstract `TexturedShaderDrawable` class, use the existing `TexturedShaderDrawNode` as its draw node, and make `Sprite` derive from it. Interested in more opinions.

osu.Framework/Graphics/UserInterface/HSVColourPicker_SaturationValueSelector.cs
```diff
@@ -257,10 +244,9 @@ public override void ApplyState()
                         hue = Source.hue;
                     }
 
-                    protected override void Blit(IRenderer renderer)
+                    protected override void UpdateUniforms(IShader shader)
                     {
-                        GetAppropriateShader(renderer).GetUniform<float>("hue").UpdateValue(ref hue);
-                        base.Blit(renderer);
+                        shader.GetUniform<float>("hue").UpdateValue(ref hue);
```
smoogipoo, line 249: Technically all of these should call `base.UpdateUniforms()`. I think that's a good precedent to set for this shader.

smoogipoo (changes requested):

smoogipoo:
I think this is fine as an initial implementation. It's a bit deviating from another idea I had (shader passes), but still probably useful to have.

EVAST9919:
Would probably be cool to have something like caching via `IFrameBuffer` in this component, so once drawn - we will be drawing only the texture instead of calling the shader every frame. In theory elements like `CircularBlob` will greatly benefit from that.

EVAST9919:
Closing as outdated

PR #5495, opened 2022-10-31
Compute CircularBlob's noise position in the DrawNode

EVAST9919:
Should potentially fix iOS visual issue, but it's hard to tell beforehand.
Calculating the same value for each pixel was not clever anyway.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/CircularBlob.cs
+++ b/osu.Framework/Graphics/UserInterface/CircularBlob.cs
@@ -6,6 +6,7 @@ using osu.Framework.Allocation;
 using osu.Framework.Graphics.Rendering;
 using osu.Framework.Graphics.Shaders;
 using osu.Framework.Graphics.Sprites;
+using osuTK;
 
 namespace osu.Framework.Graphics.UserInterface
 {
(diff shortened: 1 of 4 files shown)


Susko3:
Relevant conversation on discord: https://discord.com/channels/188630481301012481/589331078574112768/1036618137744977931

smoogipoo (approved):

smoogipoo:
@frenzibyte / @peppy any chance this could be tested on device before merge?

bdach:
Doesn't seem much better to me...

osu.Framework/Graphics/UserInterface/CircularBlob.cs
```diff
@@ -105,7 +107,15 @@ public override void ApplyState()
...
                 frequency = Source.frequency;
                 amplitude = Source.amplitude;
-                seed = Source.seed;
+
+                int newSeed = Source.seed;
+
+                if (seed != newSeed)
+                {
+                    Random rand = new Random(newSeed);
+                    noisePosition = new Vector2((float)(rand.NextDouble() * 100000), (float)(rand.NextDouble() * 100000));
```
bdach, line 116: Do these have to be multiplied by 100000? Could something lower like 1000 work?

Asking because decreasing this seems to fix the issue on GL ES. Don't ask me why, I don't have any clue yet.

bdach:
Pushed a fix. Looks better on my device now.

bdach (approved):

peppy (approved):

PR #5500, opened 2022-11-03
Fix collecting junk tablet input data 

gepbird:
Closes https://github.com/ppy/osu/issues/21000.

### DIFF
--- a/osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
+++ b/osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
@@ -71,6 +71,10 @@ namespace osu.Framework.Input.Handlers.Tablet
 
                 Detect();
 
+                // wait a small delay for OTD to finish detecting the testing devices
+                // and avoid collecting junk reports caused by the tests
+                await Task.Delay(50, cancellationToken).ConfigureAwait(false);
+
                 foreach (var device in InputDevices)
                 {
                     foreach (var endpoint in device.InputDevices)
--- a/osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
+++ b/osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
@@ -4,7 +4,6 @@
 #if NET6_0_OR_GREATER
 using System;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using OpenTabletDriver;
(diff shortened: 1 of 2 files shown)


osu.Framework/Input/Handlers/Tablet/TabletDriver.cs
```diff
@@ -60,9 +60,6 @@ private void detectAsync()
 
         private async Task detectAsync(CancellationToken cancellationToken)
         {
-            // wait a small delay as multiple devices may appear over a very short interval.
-            await Task.Delay(50, cancellationToken).ConfigureAwait(false);
```
smoogipoo, line 64: I feel like the reason for this was to prevent `DevicesChanged()` getting called many times within a short interval and causing several calls to `GetDevices()`. I think it'd still be desired along with the new delay.

gutyina70: Yes, `DevicesChanged` does get called multiple times sometimes. For example when unplugging the tablet, it calls 2 times, when plugging it in, it calls it only 1 times. But the [condition in DevicesChanged](https://github.com/ppy/osu-framework/pull/5500/files#diff-d997df1770defe61f6a0a2a9775bda0be3cba66ce707a6065eac35cf5bf9df75L46) limits the calls to only 1 (at least on my machine). If I understand correctly, you would only get multiple calls to `detectAsync`, when the user plugs in multiple tablets in a very short time.

I still think this delay should be removed, but let me know if it should be kept just in case.

gutyina70 (commented):
If we end up removing the original debounce delay, I think the cancellation token can be removed as well.

X9VoiD:
OTD does not send any "test reports", but it do can receive reports from the tablet that it mistakes as a report with actual tablet data. An example would be an initialization acknowledgement from some xp pen v1 tablets. (this is a regression of 0.6.x)

I'd say this PR is unnecessary and best resolved upstream.

gutyina70:
> OTD does not send any "test reports", but it do can receive reports from the tablet that it mistakes as a report with actual tablet data. An example would be an initialization acknowledgement from some xp pen v1 tablets. (this is a regression of 0.6.x)

Can you take a look at https://github.com/OpenTabletDriver/OpenTabletDriver/issues/2472?

peppy:
Blocking pending upstream investigation.

gepbird:
This issue was fixed in an OTD update.

PR #5501, opened 2022-11-06
Use (high resolution) waitable timers on windows when available

peppy:
Intends to fix the remaining inaccuracy that can be seen when running at frame rates close to the maximum precision provided by `Thread.Sleep` on windows.

A few preliminary notes:

- This is windows only. Other (unix/linux) platforms already had higher precision thread sleep. Of note, this does now perform `Thread.Sleep` using floating point rather than hard `int` rounding, so it may marginally improve results on non-windows platforms
- Frame jitter is now displayed on the frame statistics display. This is standard deviation between `NewFrame` calls averaged over last 256 frames
- Don't trip over the pinvoke code. I considered using [Vanara](https://github.com/dahall/Vanara) for this but that had a few issues (and the DLL is a bit large so we'd want pruning support). See https://github.com/dahall/Vanara/issues/339, may be something we can switch to in the future rather than implementing pinvokes locally, as it's much better documented and tested.
- CPU usage is basically unaffected, which is good.
- This may become available in SDL in a future release (see https://github.com/libsdl-org/SDL/pull/6232) but it's so much of an improvement that implementing local seems worthwhile for now.

Results:

Before (`Thread.Sleep`):
mpv_2022-11-06_15-01-51

image

After (`WaitableTimer`):
mpv_2022-11-06_15-02-07

image

You can count the individual frames being run. In both profile runs above, the reported frame rate is 1,000, but in the first multiple frames were being coalesced into single back-to-back executions (which is basically pointless to us).

Note that even when not using `HIGH_PRECISION` mode, the results are basically the same. So this should improve the experience not only on recent window builds but most all.

- [x] Needs testing on machines which are limited in performance (ie. not capping the frame limiter out). Should we still invoke the timer wait when time-to-be-waited is zero?

### DIFF
--- a/osu.Framework/Timing/ThrottledFrameClock.cs
+++ b/osu.Framework/Timing/ThrottledFrameClock.cs
@@ -104,6 +104,7 @@ namespace osu.Framework.Timing
 
             if (waitableTimer != IntPtr.Zero)
             {
+                // Not sure if we want to fall back to Thread.Sleep on failure here, needs further investigation.
                 if (Execution.SetWaitableTimerEx(waitableTimer, Execution.CreateFileTime(timeSpan), 0, null, default, IntPtr.Zero, 0))
                     Execution.WaitForSingleObject(waitableTimer, Execution.INFINITE);
             }
--- a/osu.Framework/Timing/ThrottledFrameClock.cs
+++ b/osu.Framework/Timing/ThrottledFrameClock.cs
@@ -35,20 +35,23 @@ namespace osu.Framework.Timing
 
         internal ThrottledFrameClock()
         {
-            try
+            if (RuntimeInfo.OS == RuntimeInfo.Platform.Windows)
             {
-                // Attempt to use CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, only available since Windows 10, version 1803.
-                waitableTimer = Execution.CreateWaitableTimerEx(IntPtr.Zero, null, Execution.CreateWaitableTimerFlags.CREATE_WAITABLE_TIMER_MANUAL_RESET | Execution.CreateWaitableTimerFlags.CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, Execution.TIMER_ALL_ACCESS);
-
-                if (waitableTimer == IntPtr.Zero)
+                try
                 {
-                    // Fall back to a more supported version. This is still far more accurate than Thread.Sleep.
-                    waitableTimer = Execution.CreateWaitableTimerEx(IntPtr.Zero, null, Execution.CreateWaitableTimerFlags.CREATE_WAITABLE_TIMER_MANUAL_RESET, Execution.TIMER_ALL_ACCESS);
+                    // Attempt to use CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, only available since Windows 10, version 1803.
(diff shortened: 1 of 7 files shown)


osu.Framework/Timing/ThrottledFrameClock.cs
```diff
@@ -30,6 +31,28 @@ public class ThrottledFrameClock : FramedClock
...
+                if (waitableTimer == IntPtr.Zero)
+                {
+                    // Fall back to a more supported version. This is still far more accurate than Thread.Sleep.
+                    waitableTimer = Execution.CreateWaitableTimerEx(IntPtr.Zero, null, Execution.CreateWaitableTimerFlags.CREATE_WAITABLE_TIMER_MANUAL_RESET, Execution.TIMER_ALL_ACCESS);
+                }
+            }
+            catch
+            {
+                // Non-windows systems will not be able to use these timers.
+            }
```
Susko3, line 53: I'd replace this with an `RuntimeInfo.OS` check instead of trying to call winapi on other platforms.

osu.Framework/Timing/ThrottledFrameClock.cs
```diff
@@ -63,21 +86,37 @@ private void throttle()
...
             double before = CurrentTime;
 
-            Thread.Sleep(milliseconds);
+            TimeSpan timeSpan = TimeSpan.FromMilliseconds(milliseconds);
+
+            if (waitableTimer != IntPtr.Zero)
+            {
+                if (Execution.SetWaitableTimerEx(waitableTimer, Execution.CreateFileTime(timeSpan), 0, null, default, IntPtr.Zero, 0))
+                    Execution.WaitForSingleObject(waitableTimer, Execution.INFINITE);
+            }
```
Susko3, line 107: If `SetWaitableTimerEx` fails, `Thread.Sleep` is skipped.

Something like

```cs
if (waitableTimer != IntPtr.Zero
    && Execution.SetWaitableTimerEx(waitableTimer, Execution.CreateFileTime(timeSpan), 0, null, default, IntPtr.Zero, 0))
{
    Execution.WaitForSingleObject(waitableTimer, Execution.INFINITE);
}
```

Or combine `Set` and `Wait` into a single helper function in `Execution`.

osu.Framework/Timing/ThrottledFrameClock.cs
```diff
@@ -30,6 +31,28 @@ public class ThrottledFrameClock : FramedClock
...
         public double TimeSlept { get; private set; }
 
+        private readonly IntPtr waitableTimer;
+
+        internal ThrottledFrameClock()
+        {
+            try
+            {
+                // Attempt to use CREATE_WAITABLE_TIMER_HIGH_RESOLUTION, only available since Windows 10, version 1803.
+                // ReSharper disable once BitwiseOperatorOnEnumWithoutFlags
```
Susko3, line 41: Unnecessary with the local implementation.

```suggestion
```

Susko3 (changes requested):
Appears to work fine, the `waitableTimer` survives trough `ExecutionMode` changes.

osu.Framework/Timing/ThrottledFrameClock.cs
peppy: Hmm, this was kind of intentional though, pending testing the todo point. Also I'm not sure the `Create` can succeed but the `Set` fail in any normal circumstance.

peppy:
Tested at lower frame rates to work as expected.

osu.Framework/Timing/ThrottledFrameClock.cs
peppy: After further testing I've updated to do the fallback to `Thread.Sleep`, although I couldn't find a case it would actually occur.

smoogipoo (approved):

PR #5502, opened 2022-11-06
Change game thread prefixes to suffixes

peppy:
This is just for my own sanity due to limitations in dotTrace. Without this, the actual part we care about (the thread's purpose) is truncated in the dotTrace interface with no way to properly view it.

I'm open to any other proposals

| before | after |
| -- | -- |
|image|image|


frenzibyte (approved):
I've also found it troubling to figure out which thread is which on Rider's integrated profiling window.

smoogipoo (approved):

Issue #5113, opened 2022-04-21
The default user language should be taken from the system language


Theighlin:

peppy:
I think this is pretty important to get working. Bit of an oversight until now.

huoyaoyuan:
I did some work about this several days ago. Things to mention:

- The framework sets default locale config to empty string. This sets `CurrentCulture` to neutral and overrides any user preference.
- The game doesn't handle variations of culture. The system default culture is usually with country/region code (`en-US`, `zh-CN` etc), but in osu we use non country/region code version (`en`, `zh`). It should use culture parent chain instead of string StartsWith, to handle complex scenarions like `zh-CN` -> `zh-Hans` -> `zh`.

peppy:
If it helps I can provide the fallback logic stable is using for deciding which culture to set. I know it does something similar to what you are proposing here.

Susko3:
Fixed in https://github.com/ppy/osu-framework/pull/5505. But do note that the language is taken from [the date and time format, and not the system UI language](https://github.com/ppy/osu-framework/pull/5505#issuecomment-1307822990).

PR #5505, opened 2022-11-08
Default `CurrentCulture` to system culture

Susko3:
Changes the default culture (when `FrameworkSetting.Locale` is set to `string.Empty`) to use the system culture (instead of the invariant culture) for all usages (and not only in `LocalisationManager`).

First step towards resolving https://github.com/ppy/osu/issues/21146.

### DIFF
--- a/osu.Framework.Tests/Localisation/CultureInfoHelperTest.cs
+++ b/osu.Framework.Tests/Localisation/CultureInfoHelperTest.cs
@@ -10,20 +10,20 @@ namespace osu.Framework.Tests.Localisation
     [TestFixture]
     public class CultureInfoHelperTest
     {
-        private const string invariant_culture = "";
+        private const string system_culture = "";
 
         [TestCase("en-US", true, "en-US")]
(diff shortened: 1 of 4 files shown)


frenzibyte:
Seems this would also resolve https://github.com/ppy/osu-framework/issues/5113 in the process.

peppy:
Needs some test resolution.

Susko3:
> Seems this would also resolve #5113 in the process.

Tested this on osu! and it seems to work. The language in osu! settings does display the wrong name due to [faulty code](https://github.com/ppy/osu/blob/b27e70ca479e622b12046722d8832b8d0ee81b63/osu.Game/Overlays/Settings/Sections/General/LanguageSettings.cs#L49-L50):

image

The default language is taken from the Regional Format (should probably use the 'Windows Display' / `CultureInfo.CurrentUICulture`) ('Apps and websites' is used for UWP apps and not available trough `CultureInfo` static members).

image

peppy (approved):

PR #5506, opened 2022-11-08
Remove all usage of `Add`/`RemoveMemoryPressure`

peppy:
Especially in cases like song select, we are churning a lot of GPU memory. This will eventually be fixed, but until then it seems that making the GC aware of this usage is detrimental to performance. In some very non-scientific benchmarks, removing these calls reduced gen 2 GC run counts by around 40-50% while paginating at song select.

@smoogipoo mentioned that he noticed this in the past, which resulted in https://github.com/ppy/osu-framework/pull/4298.  This PR undoes the changes there so we can once again see VBO memory usage in the global statistics overlay, but removes the underlying calls to `AddMemoryPressure` and `RemoveMemoryPressure` to maintain the expected behaviour.

### DIFF
--- a/osu.Framework/Graphics/OpenGL/Buffers/GLVertexBuffer.cs
+++ b/osu.Framework/Graphics/OpenGL/Buffers/GLVertexBuffer.cs
@@ -10,6 +10,7 @@ using osu.Framework.Statistics;
 using osu.Framework.Development;
 using osu.Framework.Graphics.Rendering;
 using osu.Framework.Graphics.Rendering.Vertices;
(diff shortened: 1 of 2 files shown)


smoogipoo:
A bit of a concern I have with this one is the churn of calls to the `Dispose()`, especially through VBOs, because that path isn't inlineable due to the embedded exception. See how it goes though.

smoogipoo (approved):

peppy:
As per discussion, I've removed the exception.

smoogipoo (approved):

PR #5507, opened 2022-11-11
Fix `TextFlowContainer` layout issues

### LINKED ISSUE #4753: `TextFlowContainer` no longer aligns text with different sizes on a baseline
CleanShot 2021-09-04 at 19 37 16

Notice the "Multiple Text Sizes" line not having the words aligned on one baseline, while each `SpriteText` implements `IHasLineBaseHeight` and `TextFlowContainer` has the logic for aligning them:
https://github.com/ppy/osu-framework/blob/1a5fd5be8a947907056d0c4d3eda0bb8d56be4b3/osu.Framework/Graphics/Containers/TextFlowContainer.cs#L383-L410

After debugging through, it seems the logic behind retrieving `childrenByLine` isn't actually working, and is instead adding every single sprite text in the flow solely to its own line, causing the above baseline adjustment logic to never apply.

This is happening due to the logic relying on `c.X > 0` to consider the children to be on the same line, but this is dependent on the flow layout code in `FlowContainer` to be called before, yet the `TextFlowContainer` layout update always runs before it:

https://github.com/ppy/osu-framework/blob/1a5fd5be8a947907056d0c4d3eda0bb8d56be4b3/osu.Framework/Graphics/Containers/TextFlowContainer.cs#L193-L202

Moving the base method call to be performed first before the text layout update seems to fix it, but caution needs to be taken on whether it's 100% the correct way.

### LINKED ISSUE #5084: `TextFlowContainer` incorrectly performs text layout with `TextAnchor` = `TopRight` & `CenterRight`
Code:
```csharp
new Container
{
    Size = new Vector2(100, 300),
    Children = new Drawable[]
    {
        new Box
        {
            Name = "Background",
            RelativeSizeAxes = Axes.Both,
            Alpha = 0.1f
        },
        new TextFlowContainer
        {
            RelativeSizeAxes = Axes.Both,
            TextAnchor = Anchor.TopRight,
            Text = "This is using TextAnchor TopRight"
        },
        new TextFlowContainer
        {
            RelativeSizeAxes = Axes.Both,
            TextAnchor = Anchor.CentreRight,
            Text = "This is using TextAnchor CentreRight"
        },
        new TextFlowContainer
        {
            RelativeSizeAxes = Axes.Both,
            TextAnchor = Anchor.BottomRight,
            Text = "This is using TextAnchor BottomRight"
        },
    },
}
```

Result:
image

### LINKED ISSUE #5499: `TextFlowContainer.TextAnchor` set to `BottomRight` has unexpected flow layout
### Discussed in https://github.com/ppy/osu-framework/discussions/5498

<div type='discussions-op-text'>

<sup>Originally posted by **HiddenNode** November  2, 2022</sup>
I am trying to understand how `TextFlowContainer.TextAnchor` affects layouting.

This is what `TextFlowContainer` looks like (minus the violet background) with:
```
Text = "A paragraph that does not fit into one line inside this TextFlowContainer" 
TextAnchor = Anchor.BottomRight
```
text

Is the text here layed out correctly, to be more specific, the second line is longer than the first line, is this how it should be layed out?
</div>

HiddenNode:
- Closes #5084
- Closes #4753 
- Closes #5499 

The idea here is to put `TextFlowContainer`'s Children inside a container and set its origin and anchor to 'TextAnchor' while the `Children` are all anchored to the top left, this way when `FillFlowContainer.ComputeLayoutPositions()` is called the text will always flow from left to right and from top to bottom.


osu.Framework/Graphics/Containers/FlowContainer.cs
```diff
@@ -75,32 +75,35 @@ public Vector2 MaximumSize
 
         private readonly Dictionary<Drawable, float> layoutChildren = new Dictionary<Drawable, float>();
 
-        protected internal override void AddInternal(Drawable drawable)
+        public override void Add(T drawable)
```
bdach, line 78: I don't think we can allow this? Like if you move the `layoutChildren` logic out to `Add()` and friends, then doesn't that mean that every single direct usage of `FlowContainer.AddInternal()` and friends in inheriting classes is now potentially not going to do layout properly?

Why are these sorts of changes (not using the internal members) done at this level?

HiddenNode: > I don't think we can allow this? Like if you move the layoutChildren logic out to Add() and friends, then doesn't that mean that every single direct usage of FlowContainer.AddInternal() and friends in inheriting classes is now potentially not going to do layout properly?

Yes, users of inheriting classes need to be aware of this, i have an idea of how to fix this.

>Why are these sorts of changes (not using the internal members) done at this level?

i had tried keeping the changes local to `TextFlowContainer` and as a result the logic got a bit convoluted, that's why i made the changes inside `FlowContainer`, would it be okay if i keep `Add()` defined at this level as long as i address your first point?

bdach: > Yes, users of inheriting classes need to be aware of this, i have an idea of how to fix this.

It makes for a major breaking change so any way to transparently migrate would go a long way.

> would it be okay if i keep `Add()` defined at this level as long as i address your first point?

I would like to see the 'convoluted' variant to be able to tell for sure.

HiddenNode: First point is addressed, i've tested the changes manually, can add test coverage if recommended.

osu.Framework/Graphics/Containers/FlowContainer.cs
```diff
@@ -77,32 +77,85 @@ public Vector2 MaximumSize
...
+            }
+
             base.AddInternal(drawable);
         }
 
+        public override void Add(T drawable)
+        {
+            if (drawable != null)
+            {
+                if (Content != this)
```
bdach, line 95: If these conditionals are to stay, then please make the first one an early return.

osu.Framework/Graphics/Containers/FlowContainer.cs
```diff
@@ -77,32 +77,85 @@ public Vector2 MaximumSize
...
-            // we have to ensure that the layout gets invalidated since Adding or Removing a child will affect the layout. The base class will not invalidate
-            // if we are set to AutoSizeAxes.None, but even in that situation, the layout can and often does change when children are added/removed.
-            InvalidateLayout();
+            if (Content == this)
+            {
+                layoutChildren.Add(drawable, 0f);
+                // we have to ensure that the layout gets invalidated since Adding or Removing a child will affect the layout. The base class will not invalidate
+                // if we are set to AutoSizeAxes.None, but even in that situation, the layout can and often does change when children are added/removed.
+                InvalidateLayout();
+            }
```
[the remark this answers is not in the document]
HiddenNode: Here is the [branch ](https://github.com/HiddenNode/osu-framework/tree/fix_TextFlowContainer_locally)
and the diff between this pr and the variant branch:
```diff
diff --git a/osu.Framework/Graphics/Containers/FlowContainer.cs b/osu.Framework/Graphics/Containers/FlowContainer.cs
index 03108d806..bee442603 100644
--- a/osu.Framework/Graphics/Containers/FlowContainer.cs
+++ b/osu.Framework/Graphics/Containers/FlowContainer.cs
@@ -73,89 +73,36 @@ public Vector2 MaximumSize
         /// </summary>
         protected virtual void InvalidateLayout() => layout.Invalidate();

-        private readonly Dictionary<Drawable, float> layoutChildren = new Dictionary<Drawable, float>();
+        protected readonly Dictionary<Drawable, float> LayoutChildren = new Dictionary<Drawable, float>();

         protected internal override void AddInternal(Drawable drawable)
         {
-            if (Content == this)
-            {
-                layoutChildren.Add(drawable, 0f);
-                // we have to ensure that the layout gets invalidated since Adding or Removing a child will affect the layout. The base class will not invalidate
-                // if we are set to AutoSizeAxes.None, but even in that situation, the layout can and often does change when children are added/removed.
-                InvalidateLayout();
-            }
-
+            LayoutChildren.Add(drawable, 0f);
+            // we have to ensure that the layout gets invalidated since Adding or Removing a child will affect the layout. The base class will not invalidate
+            // if we are set to AutoSizeAxes.None, but even in that situation, the layout can and often does change when children are added/removed.
+            InvalidateLayout();
             base.AddInternal(drawable);
         }

-        public override void Add(T drawable)
-        {
-            if (drawable != null)
-            {
-                if (Content != this)
-                {
-                    layoutChildren.Add(drawable, 0f);
-                    // we have to ensure that the layout gets invalidated since Adding or Removing a child will affect the layout. The base class will not invalidate
-                    // if we are set to AutoSizeAxes.None, but even in that situation, the layout can and often does change when children are added/removed.
-                    InvalidateLayout();
-                }
-
-                base.Add(drawable);
-            }
-        }
-
         protected internal override bool RemoveInternal(Drawable drawable, bool disposeImmediately)
         {
-            if (Content == this)
-            {
-                layoutChildren.Remove(drawable);
-                // we have to ensure that the layout gets invalidated since Adding or Removing a child will affect the layout. The base class will not invalidate
-                // if we are set to AutoSizeAxes.None, but even in that situation, the layout can and often does change when children are added/removed.
-                InvalidateLayout();
-            }
+            LayoutChildren.Remove(drawable);
+            // we have to ensure that the layout gets invalidated since Adding or Removing a child will affect the layout. The base class will not invalidate
+            // if we are set to AutoSizeAxes.None, but even in that situation, the layout can and often does change when children are added/removed.
+            InvalidateLayout();

             return base.RemoveInternal(drawable, disposeImmediately);
         }

-        public override bool Remove(T drawable, bool disposeImmediately)
-        {
-            if (Content != this)
-            {
-                layoutChildren.Remove(drawable);
-                // we have to ensure that the layout gets invalidated since Adding or Removing a child will affect the layout. The base class will not invalidate
-                // if we are set to AutoSizeAxes.None, but even in that situation, the layout can and often does change when children are added/removed.
-                InvalidateLayout();
-            }
-
-            return base.Remove(drawable, disposeImmediately);
-        }
-
         protected internal override void ClearInternal(bool disposeChildren = true)
         {
-            if (Content == this)
-            {
-                layoutChildren.Clear();
-                // we have to ensure that the layout gets invalidated since Adding or Removing a child will affect the layout. The base class will not invalidate
-                // if we are set to AutoSizeAxes.None, but even in that situation, the layout can and often does change when children are added/removed.
-                InvalidateLayout();
-            }
-
+            LayoutChildren.Clear();
+            // we have to ensure that the layout gets invalidated since Adding or Removing a child will affect the layout. The base class will not invalidate
+            // if we are set to AutoSizeAxes.None, but even in that situation, the layout can and often does change when children are added/removed.
+            InvalidateLayout();
             base.ClearInternal(disposeChildren);
         }

-        public override void Clear(bool disposeChildren)
-        {
-            if (Content != this)
-            {
-                layoutChildren.Clear();
-                // we have to ensure that the layout gets invalidated since Adding or Removing a child will affect the layout. The base class will not invalidate
-                // if we are set to AutoSizeAxes.None, but even in that situation, the layout can and often does change when children are added/removed.
-                InvalidateLayout();
-            }
-
-            base.Clear(disposeChildren);
-        }
-
         /// <summary>
         /// Changes the position of the drawable in the layout. A higher position value means the drawable will be processed later (that is, the drawables with the lowest position appear first, and the drawable with the highest position appear last).
         /// For example, the drawable with the lowest position value will be the left-most drawable in a horizontal <see cref="FillFlowContainer{T}"/> and the drawable with the highest position value will be the right-most drawable in a horizontal <see cref="FillFlowContainer{T}"/>.
@@ -164,10 +111,10 @@ public override void Clear(bool disposeChildren)
         /// <param name="newPosition">The new position in the layout the drawable should have.</param>
         public void SetLayoutPosition(Drawable drawable, float newPosition)
         {
-            if (!layoutChildren.ContainsKey(drawable))
+            if (!LayoutChildren.ContainsKey(drawable))
                 throw new InvalidOperationException($"Cannot change layout position of drawable which is not contained within this {nameof(FlowContainer<T>)}.");

-            layoutChildren[drawable] = newPosition;
+            LayoutChildren[drawable] = newPosition;
             InvalidateLayout();
         }

@@ -190,10 +137,10 @@ public void Insert(int position, T drawable)
         /// <returns>The position of the drawable in the layout.</returns>
         public float GetLayoutPosition(Drawable drawable)
         {
-            if (!layoutChildren.ContainsKey(drawable))
+            if (!LayoutChildren.ContainsKey(drawable))
                 throw new InvalidOperationException($"Cannot get layout position of drawable which is not contained within this {nameof(FlowContainer<T>)}.");

-            return layoutChildren[drawable];
+            return LayoutChildren[drawable];
         }

         protected override bool UpdateChildrenLife()
@@ -209,7 +156,7 @@ protected override bool UpdateChildrenLife()
         /// <summary>
         /// Gets the children that appear in the flow of this <see cref="FlowContainer{T}"/> in the order in which they are processed within the flowing layout.
         /// </summary>
-        public virtual IEnumerable<Drawable> FlowingChildren => Content.AliveInternalChildren.Where(d => d.IsPresent).OrderBy(d => layoutChildren[d]).ThenBy(d => d.ChildID);
+        public virtual IEnumerable<Drawable> FlowingChildren => AliveInternalChildren.Where(d => d.IsPresent).OrderBy(d => LayoutChildren[d]).ThenBy(d => d.ChildID);

         protected abstract IEnumerable<Vector2> ComputeLayoutPositions();

diff --git a/osu.Framework/Graphics/Containers/TextFlowContainer.cs b/osu.Framework/Graphics/Containers/TextFlowContainer.cs
index d87fd7a8a..ba87e7c53 100644
--- a/osu.Framework/Graphics/Containers/TextFlowContainer.cs
+++ b/osu.Framework/Graphics/Containers/TextFlowContainer.cs
@@ -29,6 +29,8 @@ public class TextFlowContainer : FillFlowContainer

         private readonly List<ITextPart> parts = new List<ITextPart>();

+        public override IEnumerable<Drawable> FlowingChildren => AliveChildren.Where(d => d.IsPresent).OrderBy(d => LayoutChildren[d]).ThenBy(d => d.ChildID);
+
         private readonly Cached partsCache = new Cached();

         /// <summary>
@@ -286,8 +288,19 @@ public override void Add(Drawable drawable)
             throw new InvalidOperationException($"Use {nameof(AddText)} to add text to a {nameof(TextFlowContainer)}.");
         }

+        public override bool Remove(Drawable drawable, bool disposeImmediately)
+        {
+            LayoutChildren.Remove(drawable);
+            InvalidateLayout();
+
+            return base.Remove(drawable, disposeImmediately);
+        }
+
         public override void Clear(bool disposeChildren)
         {
+            LayoutChildren.Clear();
+            InvalidateLayout();
+
             base.Clear(disposeChildren);
             parts.Clear();
         }
@@ -339,7 +352,12 @@ private void recreatePart(ITextPart part)
         {
             part.RecreateDrawablesFor(this);
             foreach (var drawable in part.Drawables)
+            {
                 base.Add(drawable);
+                LayoutChildren.Add(drawable, 0f);
+            }
+
+            InvalidateLayout();
         }

         private readonly Cached layout = new Cached();
```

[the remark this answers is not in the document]
bdach: The local fix is much more palatable as far as I'm concerned. May be a good idea to get a second opinion (@smoogipoo up for taking a look?) but yeah.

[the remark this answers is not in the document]
HiddenNode: I've made the change, Thank you for the feedback.

osu.Framework/Graphics/Containers/TextFlowContainer.cs
```diff
@@ -174,33 +188,23 @@ protected override void Update()
...
         protected override void UpdateAfterChildren()
         {
+            base.UpdateAfterChildren();
+
             if (!layout.IsValid)
             {
                 computeLayout();
                 layout.Validate();
             }
+        }
```
bdach, line 200: I think this entire override can be removed now? `base.UpdateAfterChildren()` calls `PerformLayout()` which is overridden below to do what it needs to, so this extra `layout.IsValid` check looks like unreachable code.

bdach (commented):
Looks fine aside from one minor point.

Whenever that's fixed, I'll probably add a benchmark to double-check that the added virtual dispatch isn't killing performance too much. But I think it should be fine given that we have invalidation there.

osu.Framework/Graphics/Containers/TextFlowContainer.cs
HiddenNode: `FlowContainer.layout` and `TextFlowContainer.layout` don't get invalidated for the same reasons, so this is still needed to handle a change in some `TexFlowContainer` Properties (`LineSpacing`, `ContentIndent`, etc...). this override calls `computeLayout` and not `PerformLayout`.

bdach:
Benchmark pushed in [a7d0765](https://github.com/ppy/osu-framework/pull/5507/commits/a7d0765c303333e2cafe652dda92e716afdc5571).

Current master:

|  Method |     Mean |   Error |   StdDev | Allocated |
|-------- |---------:|--------:|---------:|----------:|
| SetText | 116.8 ns | 9.12 ns | 26.60 ns |     153 B |

This PR:

|  Method |     Mean |    Error |   StdDev |   Median | Allocated |
|-------- |---------:|---------:|---------:|---------:|----------:|
| SetText | 172.6 ns | 23.98 ns | 68.79 ns | 139.3 ns |     153 B |

Looks like there is a fair bit of overhead but again, likely not a huge deal given invalidations.

Gonna request a second review to ensure I haven't screwed the testing methodology, but looks fine as far as I can tell.

bdach (approved):

osu.Framework/Graphics/Containers/FlowContainer.cs
```diff
@@ -73,11 +73,11 @@ public Vector2 MaximumSize
         /// </summary>
         protected virtual void InvalidateLayout() => layout.Invalidate();
 
-        private readonly Dictionary<Drawable, float> layoutChildren = new Dictionary<Drawable, float>();
+        protected readonly Dictionary<Drawable, float> LayoutChildren = new Dictionary<Drawable, float>();
```
smoogipoo, line 76: Hmm, I'm not sure about this. As far as I understand this entire implementation cuts through the standard FillFlowContainer flow by redirecting children to a content container, and then manually adds those children to this `LayoutChildren` in order to restore normal functionality.

Can it not be done the other way around - have `TextFlowContainer : CompositeDrawable` with a single `FillFlowContainer` child?

HiddenNode: >Can it not be done the other way around - have TextFlowContainer : CompositeDrawable with a single FillFlowContainer child?

i will give this a try, i would also suggest that you to take a look at [this](https://github.com/ppy/osu-framework/compare/master...HiddenNode:osu-framework:fix_FillFlowContainer_old?expand=1) (this pr before i force pushed), its similar and doesn't expose anything from `FlowContainer`.

bdach: I was mostly fine with the above branch until I got to [this part](https://github.com/HiddenNode/osu-framework/blob/47bcdb8c359a25bc507a5eecd38d6b96905c29a9/osu.Framework/Graphics/Containers/TextFlowContainer.cs#L280-L298) with the manual `InvalidateLayout()` which are a touch ugly and the hope would be that this wouldn't be necessary... Maybe some of that stuff can be moved into that inner container too, to eliminate the manual invalidations?

HiddenNode: It turns out those calls are redundant and invalidation happens at `FlowContainer`'s level , i've removed them.

bdach: In that case I'd be ok with that version. @smoogipoo what's your view?

smoogipoo: I find that implementation weird too, it's the reverse of my suggestion of a `Container` (/`CompositeDrawable`) containing a `FillFlowContainer` as the content.

In reading that tree, I'm not entirely sure what to make of `FillFlowContainer.Content`. I suppose if we can all agree that a `FillFlowContainer` not only flows its own children but also the children of its content?

Furthermore the condition of, e.g.:
```
AddInternal()
{
    if (Content == this)
    {
        layoutChildren.Add();
    }
}
```
Looks as if you could redirect the content of the FFC and then `AddInternal()` in an overridden class and have that child not be flowed by the FFC?

To me it almost sounds like FFC shouldn't be a `Container` but a `CompositeDrawable` so it doesn't have `Content`, because I'm not sure how much of this makes sense.

bdach: > we can all agree that any class deriving from `FlowContainer` should not redirect `Content`

You can't assume that because that is counterintuitive to everything else in the framework. If you want to assume that, then `Content` should just not exist (i.e. `TextFlowContainer` should inherit `CompositeDrawable` and call it a day).

HiddenNode: To clarify the assumption is that a class deriving from `FlowContainer` should not redirect `Content` AND expect its `InternalChildren` to be layed out.

PR #2585, opened 2019-06-28
Reduce size of texture atlases and allow automatic fallback

peppy:

PR #5508, opened 2022-11-12
Implement custom mip map generation & raise maximum texture atlas size to 4096x4096

Tom94:
Undoes https://github.com/ppy/osu-framework/pull/2585

Larger texture atlases drastically reduce the number of draw calls by avoiding texture switching.

The previous limit of 1024x1024 is not even enough to hold the latin font (as used in the font tests), let alone cjk. In osu! song select, the number draw calls is reduced from ~900 to ~400.

Large texture atlases were previously _not_ used, because their mipmap generation (then based on `GL.GenerateMipmap`) was slow, causing stutters. This PR introduces a custom mipmap generation pipeline that only touches the regions of the texture atlas that are being updated, hence circumventing stutters.


peppy:
The only apprehension I have here is our usage of `glGenerateMipmap`, which could potentially be much more expensive for every texture upload to an atlas as a result of this change.

https://github.com/ppy/osu-framework/blob/401fb3e945d4c3358ffba1acf91c2d1596b94baa/osu.Framework/Graphics/OpenGL/Textures/GLTexture.cs#L204

Before merging, this should be tested on every platform we have access to, especially lower powered ones.

peppy:
Would be good to have a test scene to test this precise scenario (which uploads a new texture each test step).

peppy:
With a focus on the spikes above average:

1024 With `glGenerateMipmap`:

osu Framework Tests 2022-11-15 at 07 16 56


8192 With `glGenerateMipmap`:

osu Framework Tests 2022-11-15 at 07 14 23

8192 With `glGenerateMipmap` but no `Nicest` hinting:

osu Framework Tests 2022-11-15 at 07 18 40

8192 Without `glGenerateMipmap`:

osu Framework Tests 2022-11-15 at 07 12 07

Pretty safe to say there *is* overhead which we need to address, but a path forward is probably very simple.

I'd want to see similar testing done on an M1 GPU, as @smoogipoo has mentioned overheads with texture uploads.

Diff to disable mipmapping:

```diff
diff --git a/osu.Framework/Graphics/Textures/TextureAtlas_BackingAtlasTexture.cs b/osu.Framework/Graphics/Textures/TextureAtlas_BackingAtlasTexture.cs
index 6233a9950..abdd41544 100644
--- a/osu.Framework/Graphics/Textures/TextureAtlas_BackingAtlasTexture.cs
+++ b/osu.Framework/Graphics/Textures/TextureAtlas_BackingAtlasTexture.cs
@@ -32,7 +32,7 @@ private class BackingAtlasTexture : Texture
             private static readonly Rgba32 initialisation_colour = default;
 
             public BackingAtlasTexture(IRenderer renderer, int width, int height, bool manualMipmaps, TextureFilteringMode filteringMode = TextureFilteringMode.Linear, int padding = 0)
-                : this(renderer.CreateTexture(width, height, manualMipmaps, filteringMode, initialisationColour: initialisation_colour))
+                : this(renderer.CreateTexture(width, height, true, TextureFilteringMode.Linear, initialisationColour: initialisation_colour))
             {
                 this.padding = padding;
                 atlasBounds = new RectangleI(0, 0, Width, Height);

```

Testing can be done using https://github.com/ppy/osu-framework/pull/5517. Note that the screenshots above were running one upload per test step, while I've changed the test to run per 50ms, so the spikes will be *more frequent* on that branch. Can be adjusted as required.

peppy (commented):
Requires further consideration

frenzibyte:
1024 with `glGenerateMipmap`:
CleanShot 2022-11-17 at 00 40 40

8192 with `glGenerateMipmap`:
CleanShot 2022-11-17 at 00 42 56

8192 with `glGenerateMipmap` but no `Nicest` hinting:
CleanShot 2022-11-17 at 00 44 45

8192 without `glGenerateMipmap`:
CleanShot 2022-11-17 at 00 48 22

peppy:
Performance wise, the new changes bring this back in line with `master` levels. Have not yet gone over the code changes with a fine comb.

Tom94:
Should be good for review now -- I'm done making implementation changes.

Morilli:
Is memory usage not a concern here?
Testing with osu!(lazer), memory allocation appears to be around 1GB higher at startup (2 8192x8192 texture atlases used), and about 1.5GB higher when starting to play a beatmap (4 texture atlases used) compared to without this change, which seems like a considerable amount.
The texture atlas allocations themselves seem to also create a big lagspike at the start of the first beatmap and when starting the game itself.

peppy:
> Is memory usage not a concern here?
> Testing with osu!(lazer), memory allocation appears to be around 1GB higher at startup (2 8192x8192 texture atlases used), and about 1.5GB higher when starting to play a beatmap (4 texture atlases used) compared to without this change, which seems like a considerable amount.

If it's an issue we can drop back down to 4096x feasibly. But going forward we would be moving towards using a single atlas game-wide. I am currently working on combining the main atlas with the font store one, so that will reduce the startup overhead to just one atlas texture.

For gameplay, I'm not sure what is making the atlases but it will need further investigation. The eventual plan is to add support for evicting textures from atlases so we can potentially have one shared global atlas across the whole game.

> The texture atlas allocations themselves seem to also create a big lagspike at the start of the first beatmap and when starting the game itself.

Can you show frame graphs of this happening, and provide hardware specs?

Morilli:
> going forward we would be moving towards using a single atlas game-wide

That sounds good and would probably solve that issue completely once implemented.

> Can you show frame graphs of this happening, and provide hardware specs?

My CPU is an intel i5 7600K (4 core ~4GHz), and my GPU a nvidia gtx 1060 (6GB). Here's a video showing the first map start: 

https://user-images.githubusercontent.com/35152647/202416356-3d217c9c-ea47-4a18-b62d-19203e91da5e.mp4

peppy:
Thanks for the video! @Tom94 is working on an improvement that may fix the huge initial stutter, so stay tuned for that.

Morilli:
Indeed, this is much better! The tiny lag remaining is basically unnoticable now 👍

peppy:
Also see https://github.com/ppy/osu-framework/pull/5521 which will reduce the memory by some factory.

Gameplay requires a bit of further consideration, but on an initial check we can probably remove one of the two atlases there as well.

peppy:
From a memory overhead angle, we may want to consider bumping this up to 4096x4096 to start with (67mb per atlas texture vs  268mb for 8192x). At least until we have atlas eviction support and a better algorithm to allow for sharing a single atlas game-wide.

For example, right now during gameplay there are 4-5 `TextureStore`s which each create their own atlas. Even if they don't consume anywhere near the full size of the atlas texture they will currently incur the above overhead at a GPU/driver level.

osu.Framework/Graphics/OpenGL/Textures/GLTexture.cs
```diff
@@ -194,17 +198,142 @@ public unsafe bool Upload()
...
+                // Mipmap generation using the merged upload regions follows
+                int frameBuffer = GL.GenFramebuffer();
+
+                BlendingParameters previousBlendingParameters = Renderer.CurrentBlendingParameters;
+
+                // Use a simple render state (no blending, masking, scissoring, stenciling, etc.)
+                Renderer.SetBlend(BlendingParameters.None);
+                Renderer.PushDepthInfo(new DepthInfo(false, false));
+                Renderer.PushStencilInfo(new StencilInfo(false));
+                Renderer.PushScissorState(false);
```
frenzibyte, line 248: May be safer to reset blending mask to `All` temporarily as well (in case a texture is uploaded during a back pass, if that's ever the case). Makes me think there should be an `IRenderer.ResetState` method or something similar to simplify the code here (can see it useful in a screenshot operation on non-OpenGL backend, where reading framebuffer pixels is not possible), but could leave that one for later.

osu.Framework/Graphics/OpenGL/Textures/GLTexture.cs
```diff
@@ -278,57 +407,52 @@ protected virtual void DoUpload(ITextureUpload upload, IntPtr dataPointer)
...
+            // Bind a dummy frame buffer such that we can later restore the current framebuffer
+            // state via Renderer.UnbindFrameBuffer(null);
+            Renderer.BindFrameBuffer(null);
+
+            // Initialize texture to solid color
+            int frameBuffer = GL.GenFramebuffer();
+            GL.BindFramebuffer(FramebufferTarget.Framebuffer, frameBuffer);
+            GL.FramebufferTexture2D(FramebufferTarget.Framebuffer, FramebufferAttachment.ColorAttachment0, TextureTarget2d.Texture2D, TextureId, level);
+            Renderer.Clear(new ClearInfo(initialisationColour));
+            GL.DeleteFramebuffer(frameBuffer);
```
frenzibyte, line 453: Sounds like scissoring and blending mask (technically colour write mask) should be reset here as well, otherwise I believe they will affect the initialisation here.

frenzibyte (commented):
Flags look a bit darker than usual, and resizing the game down to 640x480 makes them even darker.

Normal size:

CleanShot 2022-11-20 at 17 44 18

640x480:

CleanShot 2022-11-20 at 17 44 08

(notice the colour of the flag in the leaderboard card)

---

That being said, I haven't dwelt into the ends of the mipmap generation logic but overall it looks pretty interesting, and can see it eventually turning into its own class instantiated in `Renderer` and exposed to textures via a `GenerateMipmap` method that accepts a list of quads. But that can definitely be left for a follow-up.

peppy:
@Tom94 since you know this best, can you give the flag issue above a quick check and provide your thoughts? I hope to get to reviewing this in the next few days.

Tom94:
Fixed. I had forgotten that the SRGB texture format does a sneaky to-linear conversion on load, which the mipmap shader needs to undo to get back into sRGB space.

osu.Framework/Graphics/OpenGL/Textures/GLTexture.cs
peppy: Of note, this should still be considered when reviewing this PR.

smoogipoo:
In my latest changes, I've moved the shaders to resources. This is because we have mechanisms in place that ensures compatibility across multiple platforms such as [`sh_Precision_Internal.h`](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Resources/Shaders/sh_Precision_Internal.h).

Unrelated to these changes, I've noticed that the clearing of the texture levels isn't always going as expected (initialisation colour changed to red for visualisation purposes):

https://user-images.githubusercontent.com/1329837/209537123-54a5375b-7783-409c-8765-fecea7f8b7be.mp4

I originally noticed this with text where some characters had a texture bleed, however I've always been able to reproduce it with `TestSceneTextureCropping`.  
It's temperamental, because the textures seem to be cached at a GPU/driver level (Linux, so maybe related) - after reverting that red initialisation colour it _still appears_ in my level 1/2 mipmaps and bleeds into the level 3 mipmap (though level 3 isn't fully coloured):

https://user-images.githubusercontent.com/1329837/209537931-bfe6e704-8ae9-4bc3-a920-c73e9e6d5ccd.mp4

Not sure what's going on here yet, because the `Renderer.Clear()` call seems to be doing _something_, just not the correct thing?

frenzibyte:
Wondering if that has to do with what I mentioned in my two comments above.

osu.Framework/Graphics/OpenGL/Textures/GLTexture.cs
smoogipoo: Scissoring, depth, and colour write mask are reset internally to `Renderer.Clear()`. Am I missing something here? Blending is not reset (probably bad), and depth still performs a depth test (probably bad) but those don't seem to have any effect in resolving the issue I pointed out in https://github.com/ppy/osu-framework/pull/5508#issuecomment-1365065210

frenzibyte: Scissoring and depth test sure, but colour write mask? I can't seem to find any line that resets that?

smoogipoo: You're right about that - blend mask isn't set. Doesn't seem to have any effect though (disabled FTB).

frenzibyte:
I had revisited this PR in efforts to improve performance on mobile devices (iOS specifically), and spent a good few hours now to get it to fully work on Veldrid. I've only tested this on Metal, checking other backends would be appreciated.

frenzibyte:
@smoogipoo I can't reproduce what you're seeing [above](https://github.com/ppy/osu-framework/pull/5508#issuecomment-1365065210), maybe because I'm testing on M1.

EVAST9919:
In D3D text is transparent (or barely visible in some places)
d3d

osu.Framework/Resources/Shaders/sh_mipmap.vs
```diff
@@ -0,0 +1,16 @@
...
+void main()
+{
+    v_TexCoord = m_Position;
+
+// todo: I have no idea why this is required.
+#ifdef OPENGL_LEGACY
+    vec2 position = vec2(m_Position.x, m_Position.y);
+#else
+    vec2 position = vec2(m_Position.x, 1.0 - m_Position.y);
+#endif
```
smoogipoo, line 13: This is because NDC coordinates in GL are inverted. It's not a legacy-GL thing.

https://github.com/ppy/osu-framework/blob/68eecb7656dd0dbef3d8a9e13850fdf7cc0de7f0/osu.Framework/Resources/Shaders/Internal/sh_Vertex_Output.h#L22-L23

Reading material: https://matthewwellings.com/blog/the-new-vulkan-coordinate-system/

smoogipoo (changes requested):

osu.Framework/Resources/Shaders/sh_mipmap.vs
frenzibyte: So, funny story, I thought the same and I tried flipping things around myself, until I realised that `1.0 - m_Position.y` renders correctly for Veldrid OpenGL and Metal, but fails on legacy OpenGL. That's when I thought I don't think I should spend any more time for now so I bailed out.

frenzibyte:
Turns out D3D11 doesn't like using the same texture as a sampling resource and a render target at the same time, so I've separated them for now. This makes me wonder whether we can replace all of this with a compute shader at some point, but I haven't looked into whether it can sample from and write to different mipmap levels of the texture (it may be feasible with read-only `TextureView` for the higher mip level and read-write `TextureView` for the target mip level).

@EVAST9919 would appreciate confirming whether it works on your end now.

frenzibyte:
@peppy need veldrid package update with https://github.com/ppy/veldrid/pull/12 included for iOS to work (got broken with my latest changes above).

EVAST9919:
> @EVAST9919 would appreciate confirming whether it works on your end now.

Yep, all good now.

osu.Framework/Resources/Shaders/sh_mipmap.vs
frenzibyte: I've came back to this one and it turns out the dummy `BindFrameBuffer` calls were causing the global uniform buffer to not update the `g_BackbufferDraw` state (due to an early equality check), therefore the vertices aren't flipped vertically. I've replaced them using a proper frame buffer instance.

smoogipoo:
Legacy-GL: This has broken some storyboards. Here's one, where random text will be black: https://osu.ppy.sh/beatmapsets/499488#osu/1073964.

Legacy GL: My texture corruption issue still hasn't been fixed, but I haven't been able to reproduce any real world issues.

Legacy GL (though may not be renderer specific), when scrolling at song select (not reproducible):
```
System.InvalidOperationException: Attempting to bind a frame buffer while another one is still bound.Binding the same frame buffer more than once per frame may result in unexpected behaviour.
  at osu.Framework.Graphics.Rendering.Renderer.BindFrameBuffer(IFrameBuffer frameBuffer) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Rendering/Renderer.cs:line 932
  at osu.Framework.Graphics.OpenGL.Buffers.GLFrameBuffer..ctor(GLRenderer renderer, GLTexture glTexture, Int32 level) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/OpenGL/Buffers/GLFrameBuffer.cs:line 60
  at osu.Framework.Graphics.OpenGL.Textures.GLTexture.Upload() in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/OpenGL/Textures/GLTexture.cs:line 253
  at osu.Framework.Graphics.Rendering.Renderer.BindTexture(Texture texture, Int32 unit, Nullable`1 wrapModeS, Nullable`1 wrapModeT) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Rendering/Renderer.cs:line 838
  at osu.Framework.Graphics.Rendering.RendererExtensions.DrawQuad(IRenderer renderer, Texture texture, Quad vertexQuad, ColourInfo drawColour, Nullable`1 textureRect, Action`1 vertexAction, Nullable`1 inflationPercentage, Nullable`1 blendRangeOverride, Nullable`1 textureCoords) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Rendering/RendererExtensions.cs:line 127
  at osu.Framework.Graphics.Sprites.SpriteText.SpriteTextDrawNode.Draw(IRenderer renderer) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Sprites/SpriteText_DrawNode.cs:line 69
  at osu.Framework.Graphics.Containers.CompositeDrawable.CompositeDrawableDrawNode.Draw(IRenderer renderer) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable_DrawNode.cs:line 205
  at osu.Framework.Graphics.BufferedDrawNode.PopulateFrameBuffers(IRenderer renderer, Func`2 bindFrameBuffer) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/BufferedDrawNode.cs:line 125
  at osu.Framework.Graphics.DrawNode.PopulateFrameBuffers(IRenderer renderer) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/DrawNode.cs:line 138
  at osu.Framework.Graphics.Containers.CompositeDrawable.CompositeDrawableDrawNode.PopulateFrameBuffers(IRenderer renderer, Func`2 bindFrameBuffer) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable_DrawNode.cs:line 267
```

Veldrid-VK:

image

Memresable:
uh... apparently this pull request (I think? since textures usually take up memory space if they are huge) seems to have caused massive memory usage spike in the latest release (2023.4.19, Linux) and now It's unplayable unless if i rollback to older versions (this is on OpenGL, not the legacy one)
image

peppy:
This should not have increased memory usage in any meaningful way (at most 50mb or so). I can't reproduce.

PR #5509, opened 2022-11-12
Log sources of VBO flushes (aka batching interruptions) to global statistics

peppy:
2022-11-12 15 24 40@2x

2022-11-12 15 24 11@2x


bdach (approved):
Looks good to me

PR #5510, opened 2022-11-12
Remove `internal` access modifier from `CompositeDrawable.AddInternal`

peppy:
Not actually required.

`RemoveInternal` and `ClearInternal` are a bit more involved to remove, so I've just hit this one for now. It bit me and I didn't like it.


bdach (approved):
Don't see why not. Worst case is we break ourselves with this, consumers shouldn't care.

PR #5512, opened 2022-11-12
Use rounded fragment shader exclusively

### LINKED ISSUE #4537: Colour handling is not consistent between shaders
As noticed in https://github.com/ppy/osu/issues/13607#issuecomment-865944769, there is an issue with colour consistency across the rounded and non-rounded shaders, likely due to incorrect colour space conversion order. This is causing incorrect/washed out colours in one of the two scenarios.

Tom94:
This PR removes the non-masking / non-rounded shaders and uses the rounded shader instead. This has the main benefit of reducing the number of draw calls by requiring fewer batch flushes (e.g. upon shader switch).

Performance should stay equivalent due to a new early-out flag in the shader.

While making changes to the shader, it was convenient to fix https://github.com/ppy/osu/issues/13607 along the way and fix a bug in `toLinear` on GLES systems.

- Closes #4537 
- Closes ppy/osu#13607


peppy:
Looks good to my eye. Would appreciate one or two other reviews given the scope of this change.

peppy (approved):

osu.Framework/Resources/Shaders/sh_Masking.h
```diff
@@ -75,6 +76,11 @@ lowp vec4 getBorderColour()
 
 lowp vec4 getRoundedColor(lowp vec4 texel, mediump vec2 texCoord)
 {
+	if (!g_IsMasking)
+	{
+		return toSRGB(v_Colour * texel);
+	}
```
frenzibyte, line 82: I haven't deeply checked whether this changes behaviour, but `SpriteDrawNode` applies [another condition](https://github.com/ppy/osu-framework/blob/16d1c2d3356b0f59271fb49cff5fe8373ae9cdf3/osu.Framework/Graphics/Sprites/SpriteDrawNode.cs#L95) for whether to use rounded shader that doesn't look to be accounted here (diff shows `g_IsMasking` is only ever set when masking is active).

Tom94: Good catch, thanks! Should be addressed now.

frenzibyte (approved):

bdach (approved):
Let's do it

peppy (approved):

PR #5514, opened 2022-11-12
Fix possible NRE in `GetAsync` between `DllResourceStore`  and `ResourceStore`

Flutterish:
`GetAsync` in `DllResourceStore` was returning `Task<byte[]>?` rather than `Task<byte[]?>`, upon which `ResourceStore` relies for checking which of the possible filenames is the requested resource, if any.
https://github.com/ppy/osu-framework/blob/7bac4227a52cd9fd96def0e3b4815b9902e3bf44/osu.Framework/IO/Stores/ResourceStore.cs#L94
The `.ConfigueAwait` above was acting on a possibly null `Task` rather than a task which can possibly return null.


osu.Framework/IO/Stores/DllResourceStore.cs
```diff
@@ -54,7 +54,7 @@ public virtual Task<byte[]> GetAsync(string name, CancellationToken cancellation
             this.LogIfNonBackgroundThread(name);
 
             using (Stream input = GetStream(name))
-                return input?.ReadAllBytesToArrayAsync(cancellationToken);
+                return input?.ReadAllBytesToArrayAsync(cancellationToken) ?? Task.FromResult<byte[]>(null);
```
bdach, line 57: The intent here generally looks valid, but the thing I'm wondering is - was this actually ever correct code? By which I'm going to say, I'm not sure it's safe to return a `Task` which was returned by the stream from within a `using` block? Doesn't that technically mean that consumers may end up attempting to read from a disposed stream? If so, then the actual proper thing would be to `await` the task here.

It may well be that this was never an issue because resource streams use something like `MemoryStream` that has no backing IO primitive and as such doesn't care about it being disposed.

Flutterish: Huh, I didn't even notice that. It seems to be getting a stream with `Assembly.GetManifestResourceStream` (which to me sounds like in-memory because I think assembly resources are loaded in-memory?). I ran a quick test which resolved a texture and it didn't seem to have any issues.

bdach: I'd probably still use `await` to assuage all doubts. It should still fix the issue too.

bdach:
Cleaned up style some, added test coverage for failing scenario because I didn't know how to test the fix otherwise. Seems ok.

bdach (approved):

PR #5515, opened 2022-11-13
Make `IBindable : IFormattable` to allow formatting its value with `InvariantCulture`

Susko3:
Noticed while looking into https://discord.com/channels/188630481301012481/589331078574112768/1040392278348398703.

This results in the following problem when decoding osu! config when using `Norsk` language:
```cs
2022-11-13 22:35:08 [important]: Unable to parse config key LastProcessedMetadataId: System.FormatException: Input string was not in a correct format.
2022-11-13 22:35:08 [important]: at System.Number.ThrowOverflowOrFormatException(ParsingStatus status, TypeCode type)
2022-11-13 22:35:08 [important]: at System.String.System.IConvertible.ToInt32(IFormatProvider provider)
2022-11-13 22:35:08 [important]: at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
2022-11-13 22:35:08 [important]: at osu.Framework.Bindables.Bindable`1.Parse(Object input)
2022-11-13 22:35:08 [important]: at osu.Framework.Configuration.IniConfigManager`1.PerformLoad()
```

Excerpt from `game.ini`:

```ini
ScalingSizeX = 0.8
ScalingSizeY = 0.8
ScalingPositionX = 0.5
ScalingPositionY = 0.5
UIScale = 1.0
UIHoldActivationDelay = 0.0
EditorWaveformOpacity = 0,25
LastProcessedMetadataId = −1

```

That's `U+2212 : MINUS SIGN` and not the regular `U+002D : HYPHEN-MINUS` (also the decimal comma).

### DIFF
--- a/osu.Framework/Bindables/Bindable.cs
+++ b/osu.Framework/Bindables/Bindable.cs
@@ -380,7 +380,7 @@ namespace osu.Framework.Bindables
 
         public string Description { get; set; }
 
-        public override string ToString() => value?.ToString() ?? string.Empty;
+        public override string ToString() => FormattableString.Invariant($"{Value}");
 
         /// <summary>
         /// Create an unbound clone of this bindable.
--- a/osu.Framework/Bindables/Bindable.cs
+++ b/osu.Framework/Bindables/Bindable.cs
@@ -380,7 +380,9 @@ namespace osu.Framework.Bindables
 
         public string Description { get; set; }
 
-        public override string ToString() => FormattableString.Invariant($"{Value}");
+        public sealed override string ToString() => ToString(null, CultureInfo.CurrentCulture);
+
+        public virtual string ToString(string format, IFormatProvider formatProvider) => string.Format(formatProvider, $"{{0:{format}}}", Value);
 
         /// <summary>
         /// Create an unbound clone of this bindable.
--- a/osu.Framework/Bindables/BindableBool.cs
+++ b/osu.Framework/Bindables/BindableBool.cs
@@ -12,8 +12,6 @@ namespace osu.Framework.Bindables
         {
         }
 
-        public override string ToString() => Value.ToString();
-
         public override void Parse(object input)
         {
             if (input is "1")
--- a/osu.Framework/Bindables/BindableDictionary.cs
+++ b/osu.Framework/Bindables/BindableDictionary.cs
@@ -9,6 +9,7 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using osu.Framework.Caching;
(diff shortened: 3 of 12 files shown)


osu.Framework/Bindables/Bindable.cs
```diff
@@ -380,7 +380,7 @@ public virtual void UnbindFrom(IUnbindable them)
 
         public string Description { get; set; }
 
-        public override string ToString() => value?.ToString() ?? string.Empty;
+        public override string ToString() => FormattableString.Invariant($"{Value}");
```
frenzibyte, line 383: Bit of a weird place to make invariant, since it seems .NET-side `ToString` is culture-specific. If it's a serialisation issue shouldn't the logic behind it be fixed to serialise with invariant culture instead?

smoogipoo: I agree it's weird. Maybe we can provide an `IFormatProvider` overload instead?

Susko3: Bindable implementations like `BindableNumber` use `InvariantCulture` by default:

https://github.com/ppy/osu-framework/blob/ba1385330cc501f34937e08257e586c84e35d772/osu.Framework/Bindables/BindableNumber.cs#L300

I wouldn't be opposed to having `IBindable : IFormattable`. Or if that is not wanted, then just having `ToString(IFormatProvider)` as smoogi suggested.

osu.Framework/Bindables/Bindable.cs
```diff
@@ -380,7 +380,9 @@ public virtual void UnbindFrom(IUnbindable them)
 
         public string Description { get; set; }
 
-        public override string ToString() => value?.ToString() ?? string.Empty;
+        public sealed override string ToString() => ToString(null, CultureInfo.CurrentCulture);
```
peppy, line 383: The `null` here feels a bit weird.. maybe `string.Empty` works better?

peppy (commented):
Seems mostly okay?

osu.Framework/Bindables/Bindable.cs
Susko3: `string.Format` and the `DefaultInterpolatedStringHandler` use `null` format when it's empty, so that's what I used here.

```cs
$"{val}" // format is null
$"{val:g}" // format is "g"
string.Format("{0:}", val) // format is null
string.Format("{0}", val) // format is null
string.Format("{0:g}", val) // format is "g"
```

peppy: Fair enough.

peppy (approved):

smoogipoo (approved):

PR #5516, opened 2022-11-14
Fix `Dropdown` calling `GenerateItemText` before it's loaded

Susko3:
Ran into this while trying to show system default language in osu! settings.
The following causes a NRE with `localisation`:

```cs
protected class LanguageDropdownControl : DropdownControl
{
    [Resolved]
    private LocalisationManager localisation { get; set; } = null!;

    protected override LocalisableString GenerateItemText(Language item)
    {
        if (item == Language.SystemDefault)
        {
            if (LanguageExtensions.TryParseCultureCode(localisation.SystemDefaultLocaleMapping?.Name, out var language))
                return GeneralSettingsStrings.DefaultLanguage(language.GetDescription());

            return CommonStrings.Default;
        }

        return base.GenerateItemText(item);
    }
}
 ```

### DIFF
--- a/osu.Framework/Graphics/UserInterface/Dropdown.cs
+++ b/osu.Framework/Graphics/UserInterface/Dropdown.cs
@@ -52,7 +52,7 @@ namespace osu.Framework.Graphics.UserInterface
                 if (boundItemSource != null)
                     throw new InvalidOperationException($"Cannot manually set {nameof(Items)} when an {nameof(ItemSource)} is bound.");
 
-                setItems(value);
+                Scheduler.AddOnce(setItems, value);
             }
         }
 
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
@@ -6,10 +6,14 @@
 using System;
 using System.Linq;
 using NUnit.Framework;
+using osu.Framework.Allocation;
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by smoogipoo — APPROVED** [review]

File: osu.Framework/Graphics/UserInterface/Dropdown.cs [hunk 1]
```diff
@@ -52,7 +52,7 @@ public IEnumerable<T> Items
                 if (boundItemSource != null)
                     throw new InvalidOperationException($"Cannot manually set {nameof(Items)} when an {nameof(ItemSource)} is bound.");
 
-                setItems(value);
+                Scheduler.AddOnce(setItems, value);
```
peppy, line 55: This caused a performance regression (shifting the `Load` overhead of individual items' drawables to `LoadComplete`).

peppy:
I think this flow needs to change. If `LocalisationManager` is actually needed in that method, it likely needs to arrive as a parameter. A good amount of overhead may come from glyph/texture retrieval for the string which is to be displayed in each of the dropdown items, so I think the actual resolution of the string to be displayed needs to be in the async load flow.

bdach:
This is also causing actual regressions (see 2 issues linked above). I think we need to consider a revert, or an alternate solution to the issue.

peppy:
I'll take a look into this. It's been on my personal TODO list, and given it's causing other issues I'll divert my attention.

PR #5519, opened 2022-11-16
Apply NRT to all zero-refactor cases

peppy:
- Applied to all files then used rider's warning output to discard changes on any files with warnings / errors.
- Tested with framework and osu! (requires a handful of changes: https://github.com/ppy/osu/compare/master...peppy:osu:nrt-low-hanging?expand=1)

Regex for applying to rider error output (copied from Problems > All Solution Files tool window):

```
%s/^<\(.*\)>\(.*\) (.*/git checkout \1\2;/g
%s/\\/\//g
```

There's around 800 files with warnings or errors that need further attention. Most are simple changes which we can probably automate in future passes.


osu.Framework.Tests/Containers/ContainerEnumerableTest.cs
```diff
@@ -46,15 +46,15 @@ public void TestAddingContainerAsEnumerableRangeThrows(Type containerType)
             {
                 var unused = new AudioContainer
                 {
-                    Children = (IReadOnlyList<Drawable>)Activator.CreateInstance(containerType)
+                    Children = (IReadOnlyList<Drawable>)Activator.CreateInstance(containerType)!
```
Susko3, line 49: Shouldn't this use `.AsNonNull()`?
(Same goes for the other change in this file.)

peppy: I think it's much of a muchness. the main purpose of `AsNonNull` is to force runtime execution (ie. for `Dispose` methods).

smoogipoo (approved):
Ignoring the android inspections...

PR #5520, opened 2022-11-16
Refactor and simplify text builder bounds calculation 

frenzibyte:
Spent some time to continue off what I initially came up with in https://github.com/ppy/osu-framework/pull/5427#issuecomment-1304086404, as I still found it quite hard to understand how correct the current logic was and whether it works. The main idea of the new logic is to ensure the current line height is always equal to the glyph with the highest distance from the top of the line, whenever a new glyph is added or a glyph's draw rectangle was adjusted.

This is mainly a code quality change, but also fixes a minor bug on `RemoveLastCharacter` where the bounds will not take glyphs with baseline offsets into account.

@HiddenNode would appreciate checking whether this works against all possible cases you came across in #5427.


osu.Framework/Text/TextBuilder.cs
```diff
@@ -356,25 +323,6 @@ protected virtual void OnWidthExceeded()
...
-        /// <param name="glyph">The glyph to retrieve the height of.</param>
-        /// <returns>The height of the glyph.</returns>
-        private float getGlyphHeight<T>(ref T glyph)
-            where T : ITexturedCharacterGlyph
-        {
-            if (useFontSizeAsHeight)
-                return font.Size;
-
-            // Space characters typically have heights that exceed the height of all other characters in the font
-            // Thus, the height is forced to 0 such that only non-whitespace character heights are considered
```
HiddenNode, line 371: As mentionned here, the height of whitespace characters should be forced to 0, that was the point in using this function instead of using `glyph.Height` directly.

HiddenNode (commented):
There is only one thing which this change doesn't account for, which is the height of whitespace characters.

a b

osu.Framework/Text/TextBuilder.cs
frenzibyte: That still exists but at a different location (see `TextBuilderGlyph.Height`). Seems the baseline of whitespace glyphs are causing this.

frenzibyte: Good find, works correctly now.

CleanShot 2022-11-18 at 08 22 16

peppy (approved):
This is a hard change to review, but I think this looks quite good.

PR #5521, opened 2022-11-17
Allow fonts to share the main global texture atlas

peppy:
Intended to be used in conjunction with #5508.


peppy:
Blocked as this should not be merged without first increasing the size of texture atlas textures. Rationale is that we do not want to mix other textures with fonts while the atlas size is small, as it will likely lead to more texture binds due than current.

This change should be considered in conjunction with https://github.com/ppy/osu-framework/pull/5508.

Issue #5451, opened 2022-10-08
Changing window fullscreen size on macOS fails assertion


frenzibyte:
```
[runtime] 2022-10-08 12:04:32 [error]: An unhandled error has occurred.
[runtime] 2022-10-08 12:04:32 [error]: NUnit.Framework.AssertionException: Stored Displays don't match actual displays: Stored displays:
[runtime] 2022-10-08 12:04:32 [error]: Name: 0, Bounds: {X=0,Y=0,Width=3456,Height=2234}, DisplayModes: 88, Index: 0
[runtime] 2022-10-08 12:04:32 [error]: 
[runtime] 2022-10-08 12:04:32 [error]: Actual displays:
[runtime] 2022-10-08 12:04:32 [error]: Name: 0, Bounds: {X=0,Y=0,Width=3456,Height=2160}, DisplayModes: 88, Index: 0
[runtime] 2022-10-08 12:04:32 [error]: at osu.Framework.Logging.ThrowingTraceListener.Fail(String message1, String message2) in /Users/salman/Projects/osu-framework/osu.Framework/Logging/ThrowingTraceListener.cs:line 27
[runtime] 2022-10-08 12:04:32 [error]: at System.Diagnostics.TraceInternal.Fail(String message, String detailMessage)
[runtime] 2022-10-08 12:04:32 [error]: at System.Diagnostics.TraceInternal.TraceProvider.Fail(String message, String detailMessage)
[runtime] 2022-10-08 12:04:32 [error]: at System.Diagnostics.Debug.Fail(String message, String detailMessage)
[runtime] 2022-10-08 12:04:32 [error]: at System.Diagnostics.Debug.Assert(Boolean condition, String message, String detailMessage)
[runtime] 2022-10-08 12:04:32 [error]: at osu.Framework.Platform.SDL2DesktopWindow.assertDisplaysMatchSDL() in /Users/salman/Projects/osu-framework/osu.Framework/Platform/SDL2DesktopWindow_Windowing.cs:line 304
[runtime] 2022-10-08 12:04:32 [error]: at osu.Framework.Platform.SDL2DesktopWindow.handleWindowEvent(SDL_WindowEvent evtWindow) in /Users/salman/Projects/osu-framework/osu.Framework/Platform/SDL2DesktopWindow_Windowing.cs:line 444
```

Apparently changing display mode would change the display bounds altogether.

PR #5523, opened 2022-11-18
Fix displays sometimes not getting updated and failing assertion

Susko3:
- Closes https://github.com/ppy/osu-framework/issues/5451

First idea for a fix was to cover all the cases of `SDL_WindowEvent` that might be sent when displays update, but that quickly spiraled out of control. SDL might send the events in a different order, so the first one would fail the assertion, even though a later event would trigger `fetchDisplays()` as expected. Moving the assert to after the event loop fixed some of those cases.

Fetching the displays in `updateWindowStateAndSize()` also seems amicable as changing the window state will almost always change the displays (if using a non-native resolution).


peppy:
Do you see value in keeping the `if DEBUG` portions around? Makes me a bit uneasy.

peppy (approved):

PR #5526, opened 2022-11-18
Send masking parameters through via texture buffer

smoogipoo:


smoogipoo:
Superseded by https://github.com/ppy/osu-framework/pull/5952

Issue #5449, opened 2022-10-06
Moving window between displays unexpectedly moves the window again


Susko3:
To repro:
1. Change to windowed
2. Move window from one display to another (by dragging on the title bar, or with a keyboard shortcut eg. win+left arrow)
3. Notice that the window will move after finishing the drag move

Caused by a feedback loop resulting in `readWindowPositionFromConfig()` being called when it shouldn't.

Issue #5448, opened 2022-10-06
Rewrite SDL2 windowing logic to better handle changes from outside and to avoid feedback loops


Susko3:
RFC

My general idea of what the finished flow would look like:

```cs
public partial class SDL2DesktopWindow : IWindow
{
    private Display currentDisplay; // kinda sucks that there need to be this many "current displays", but alas.

    protected Display CurrentDisplay // protected so the name doesn't clash with the above
    {
        get => currentDisplay;
        // this could maybe be a function `setCurrentDisplay()`?
        // this avoids having a redundant property, one that may sometimes clash with required `IWindow` properties.
        set => CurrentDisplayBindable.Value = currentDisplay = value;
    }

    public Bindable<Display> CurrentDisplayBindable = new Bindable<Display>();

    public void SetupWindow(FrameworkConfigManager config)
    {
        // these BindValueChanged should be simple, the bulk of the logic should be in the likes of `updateDisplay`
        CurrentDisplayBindable.BindValueChanged(display =>
        {
            if (display.NewValue == CurrentDisplay)
                // if the values match, that means that the this set operation originates from within SDL2DesktopWindow
                // updating the display would lead to a feedback loop.
                return;

            Debug.Assert(display.OldValue == CurrentDisplay) // should probably pass, but this might not always be true

            ScheduleCommand(updateDisplay, display.NewValue.Index);
        });
    }

    private void updateDisplay(int displayIndex)
    {
        SDL.SDL_SetWindowPosition(...);

        // importantly, we always query SDL when setting things from foreign requests, so that the values are up-to-date.
        // notice that changing CurrentDisplayBindable.Value will _never_ set CurrentDisplay directly.
        // always assume that the foreign value may be invalid and that SDL might not be able to set the state as requested.
        CurrentDisplay = SDL.SDL_GetCurrentDisplay(SDLWindowHandle);

        // possibly update other things that may change when the display has changed,
        // or let the appropriate SDL_Events arrive (preferred)
    }

    private void handleDisplayEvent(SDL.SDL_DisplayEvent e)
    {
        // it's important that the code that interacts with SDL doesn't have to deal with bindables/flow/whatever
        // and can simply just set the properties.
        CurrentDisplay = SDL.SDL_GetCurrentDisplay(SDLWindowHandle);
    }
}
```

# Some thoughts on public window properties

# From `IWindow`:

#### `WindowState WindowState { get; set; }`
[...]

PR #5528, opened 2022-11-18
Refactor flow for setting and fetching current SDL display

Susko3:
- Closes https://github.com/ppy/osu-framework/issues/5449

This incorporates some ideas from https://github.com/ppy/osu-framework/issues/5448.
`updateDisplay()` from that proposal was scrapped in favour of `pendingDisplayIndex` to accommodate the startup case where both the `WindowMode` and `currentDisplay` are unknown and need to be updated. More detail about that will be posted in that issue thread.

The old flow assumed that the display can change iff the index changes.
Examples where this doesn't hold:
- changing display resolution/arrangement in windows settings
- setting a different resolution via `SizeFullscreen` in config

The new flow will update the display in more places and avoids feedback loops from the bindables (that was causing the linked issue).

Having `pendingDisplayIndex` and `consumePendingDisplayIndex()` turned out quite well.

### DIFF
--- a/osu.Framework/Platform/SDL2/SDL2Extensions.cs
+++ b/osu.Framework/Platform/SDL2/SDL2Extensions.cs
@@ -10,6 +10,7 @@ using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Input;
 using osu.Framework.Input.Bindings;
+using osuTK;
 using osuTK.Input;
 using SDL2;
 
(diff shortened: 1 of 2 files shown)


bdach:
Dunno why this showed up in my inbox again. Are there measurable gains from this or end goals to this, or is this just a refactor for the sake of a refactor?

Susko3:
This refactor fixes the issue mentioned in the OP. This PR will probably be closed as it relies too heavily on display indexes (indexes have changed to `SDL_DisplayID` in SDL3).

I just wanted to make sure the code still works with SDL3 -- and it appears to work. This code could a useful reference for future PRs that want to fix the issue.

PR #5530, opened 2022-11-20
Adjust search container logic to address game-side issues with hiding its children

bdach:
This is an RFC.

- Alternative to / supersedes https://github.com/ppy/osu-framework/pull/5154
- Will help address https://github.com/ppy/osu/issues/18003 if merged
- Required for further work on mod select search box game-side

# Contents

## `IHasFilterableChildren` is removed; hierarchical lookups now leverage `IContainerEnumerable<T>`

This is the same as [in the original PR](https://github.com/ppy/osu-framework/pull/5154). Rather than rely on `IHasFilterableChildren` to descend down complex hierarchies for filtering purposes, `IContainerEnumerable<T>.Children` is used, which results in less boilerplate as you don't need to remember to implement it whenever needed.

## `IConditionalFilterable` (?) is introduced

At first I tried to resurrect the aforementioned PR rather than rewrite, but the invalidation issues pointed out therein ended up being caused by a fatal contradiction in its logic [that I describe in more detail here](https://github.com/ppy/osu-framework/pull/5154#issuecomment-1312665847). As a result, this PR takes a different, explicit approach of introducing a new `IConditionalFilterable` interface.

The name is bad, I know (suggestions welcome), but the gist of it is that upon implementing `IConditionalFilterable`, you get one extra bindable to play with. If the value of said bindable is `true`, then the item is included in textual search as usual. If it is `false`, however, it - and its entire subtree - is excluded from further search on the premise that the item does not meet external criteria. In practical terms, you would set it to `false` when you want to hide some items inside a `SearchContainer` because they're unavailable due to other settings, which is [precisely the use case that needs fixing game-side](https://github.com/ppy/osu/issues/18003).

---

A branch for game-side adjustments for this framework change, that end up fixing https://github.com/ppy/osu/issues/18003, is available for preview at https://github.com/ppy/osu/compare/master...bdach:filtering-framework-change-adjustments?expand=1.

I also recommend examining the commit messages of d0d90e9ce403c71f44299ca8593b5d7a66451ef2 and b790cd353adc634e9a055c6f5bd50d130ba943ae.


smoogipoo (approved):

osu.Framework/Graphics/Containers/SearchContainer.cs
```diff
@@ -94,30 +103,57 @@ protected override void Update()
...
-        private bool match(IFilterable filterable, IEnumerable<string> searchTerms, bool searchActive, bool nonContiguousMatching)
+        private bool matchSubTree(Drawable drawable, IReadOnlyList<string> searchTerms, bool searchActive, bool nonContiguousMatching)
         {
-            IEnumerable<string> filterTerms = filterable.FilterTerms.SelectMany(localisedStr =>
-                new[] { localisedStr.ToString(), localisation.GetLocalisedString(localisedStr) });
+            bool matching = match(drawable, searchTerms, nonContiguousMatching, out var nonMatchingTerms);
+
+            if (drawable is IConditionalFilterable conditionalFilterable)
+            {
+                var canBeShownBindable = conditionalFilterable.CanBeShown.GetBoundCopy();
```
peppy, line 115: I'm a touch worried about the bindable overhead here if many filter operations happen over a short period (there will be a lot of bindable churn). Probably okay, but here's a quick comparison of loading `SettingsOverlay` and performing 10 filters + unfilters:

Before:

JetBrains Rider-EAP 2022-11-29 at 05 03 58

After:

JetBrains Rider-EAP 2022-11-29 at 05 00 53

The overhead is definitely visible, but maybe fine to eat? We can probably further optimise this at a later point (ie. avoiding clearing the list altogether and reusing existing bindings) if required.

peppy:
Let's go with this and see how things play out.

peppy (approved):

PR #5534, opened 2022-11-22
Silently return rather than throwing when performing mutations on disposed drawables

peppy:
A common pain point in test and production failures, which isn't really protecting us from incorrect usage. I think it makes more sense *not* to throw in these scenarios.

Example of a failure: https://github.com/ppy/osu/actions/runs/3511604121/jobs/5882531098

We would usually fix these by adding a `Schedule` inline, but this can potentially cause other unforeseen issues in normal usage (frame delay; events not running due to alive state etc). We've added such schedules in countless places.

Interested in thoughts of those marked for review (and anyone else that cares enough). Also please check through each exception to make sure you agree with all the removed ones. I've left some which make sense to keep around.


peppy:
There's a couple of test failures I'm unsure about, needs further investigation.

osu.Framework/Graphics/Containers/Container.cs
```diff
@@ -218,7 +218,7 @@ public virtual void Add(T drawable)
                 throw new ArgumentNullException(nameof(drawable));
 
             if (drawable.IsDisposed)
-                throw new ObjectDisposedException(nameof(drawable));
+                return;
```
bdach, line 221: This one looks suspect to me because in this case it's the drawable being *added* to the container that's disposed, rather than the container itself. As far as I can tell reverting this also makes the tests not spin forever (see CI - there are two tests in `TestLoadExceptions` that never terminate with this change).

peppy: Good catch, that's probably something we still want to error on.

bdach (approved):
I don't see a huge reason not to.

smoogipoo (approved):
I find this a little bit scary, but I guess we can try it and see what/how things blow up.

frenzibyte (commented):
Looking at the specific case mentioned in OP, I'm interested to know what's trigging the value change, since a drawable usually unbinds all of its bindables before disposal. But that being said, I can't think of that being potentially bad to cause issues hiding behind us, so I'm not completely against this change.

PR #5535, opened 2022-11-22
Extract common parts of localisation tests into base class

Susko3:


Susko3:
> Is this being used in a follow-up PR?

Yes, it will be helpful to have everything in one place for future changes.
I'm not currently planning to add more tests that use `LocalisationTestScene`.

I've just noticed that [`TestSceneSpriteTextScenarios`](https://github.com/ppy/osu-framework/blob/008ab005c621acc5915a911331429c0c93a0a37e/osu.Framework.Tests/Visual/Sprites/TestSceneSpriteTextScenarios.cs#L247) also has custom localisation, will migrate that to `LocalisationTestScene` as well.

peppy (approved):

PR #5537, opened 2022-11-22
Refactor `LocaleMapping` lookup to require an exact match to `FrameworkSetting.Locale`

Susko3:
- [x] Depends on #5536

This PR adds `SystemDefaultLocaleMapping` as a way to get which mapping is the default (for `Locale = ""`).

The `LocaleMapping.Name` now needs to match `FrameworkSetting.Locale` exactly (case-insensitive) for it to be used.
Loose matching doesn't make sense as `FrameworkSetting.Locale` is controlled by the game. The game knows what locales it has added and can set the `Locale` appropriately.

Loose matching is used for the `SystemDefaultLocaleMapping`, where it's needed.

### DIFF
--- a/osu.Framework.Tests/Localisation/LocalisationTest.cs
+++ b/osu.Framework.Tests/Localisation/LocalisationTest.cs
@@ -54,6 +54,8 @@ namespace osu.Framework.Tests.Localisation
         [Test]
         public void TestConfigSettingRetainedWhenAddingNewLanguage()
         {
+            manager.AddLanguage("ja-JP", new FakeStorage("ja-JP"));
+
             config.SetValue(FrameworkSetting.Locale, "ja-JP");
 
             // ensure that adding a new language which doesn't match the user's choice doesn't cause the configuration value to get reset.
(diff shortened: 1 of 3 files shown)


bdach:
Is there an end goal to this change? Difficult to tell by the very wordy (if not inscrutable) description.

Susko3:
The goal is to simplify the lookup as the fallback logic is entirely unnecessary with [the way locale is handled osu! side](https://github.com/ppy/osu/blob/54c2becda7b621a9837e6fba1bd006d6682ce629/osu.Game/Overlays/Settings/Sections/General/LanguageSettings.cs#L50).
`SystemDefaultLocaleMapping` also allows the default there to match the system/user preference.

I still have more changes with localisation coming.

bdach:
But is this just a cleanup series? Is there some new functionality at the end of this tunnel? I just don't know how much to prioritise this series right now.

Susko3:
The new functionality will allow customizing the `CultureInfo` and `IFormatProvider` that is passed in to `LocalisableString` formatting functions.

This will address https://github.com/ppy/osu/issues/21157 and will allow to have the 12/24 hour time apply globally with anything that is using `LocalisableString`.

For the far future, `SystemDefaultLocaleMapping` would also walk the entire language list set in system settings, not just the first language.

peppy (approved):
Looks solid enough.

PR #5539, opened 2022-11-23
Correctly return `AVERROR_EOF` in `VideoDecoder`

Igoorx:
See: https://github.com/FFmpeg/FFmpeg/blob/98f6c59f4f614f7409b225930a16a2172512c432/libavformat/avio.h#L405

### DIFF
--- a/osu.Framework/Graphics/Video/VideoDecoder.cs
+++ b/osu.Framework/Graphics/Video/VideoDecoder.cs
@@ -292,8 +292,9 @@ namespace osu.Framework.Graphics.Video
                 return 0;
 
             var span = new Span<byte>(bufferPtr, bufferSize);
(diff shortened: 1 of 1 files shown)


peppy (approved):

peppy (approved):

Igoorx:
> Thanks for the contribution! This looks correct and should be fine to get in, but I'm curious where/how you were seeing this (ie. did you experience this at runtime manifesting in some way?)

Indeed, I'm using a slightly old ffmpeg version and I saw `Invalid return value 0 for stream protocol` warning messages in the console window, after a quick search I found the comment I linked in the pull request.
In recent versions of ffmpeg this is just a debug assertion, so I guess it would only be noticeable by using a debug build of ffmpeg.

PR #5543, opened 2022-11-25
Make all Drawable classes partial

smoogipoo:
This is the big one :sweat_smile: 

I suggest reviewing changes per-commit.


peppy:
Test failures and CQ need attention here 🤔

bdach:
Test failure no. 1 (`CachedAttributeTest.TestCachedViaInterface`) _seems_ like it may be a simple generator fix. The suspect part is as follows:

https://github.com/ppy/osu-framework/blob/6f46341b1b2280a22e9db70505b2d509482abadf/osu.Framework.SourceGeneration/SyntaxContextReceiver.cs#L50-L56

Base interface types are excluded above, even though `[Cached]`'s xmldoc says this:

https://github.com/ppy/osu-framework/blob/6f46341b1b2280a22e9db70505b2d509482abadf/osu.Framework/Allocation/CachedAttribute.cs#L44-L48

Seems like changing to `.AllInterfaces` there should just fix, but given that this was explicitly commented, I'm hesitant to call it for sure at I may be missing something.

---

Test failures no. 2 and 3 (`TestResolveNullableInternal`) are more annoying. In those cases the source generator emits this:

```csharp
            public virtual void RegisterForDependencyActivation(osu.Framework.Allocation.IDependencyActivatorRegistry registry)
            {
                if (registry.IsRegistered(typeof(osu.Framework.Tests.Dependencies.SourceGeneration.DependencyContainerTest.Receiver11)))
                    return;
                registry.Register(typeof(osu.Framework.Tests.Dependencies.SourceGeneration.DependencyContainerTest.Receiver11), (t, d) =>
                {
                    ((osu.Framework.Tests.Dependencies.SourceGeneration.DependencyContainerTest.Receiver11)t).load(osu.Framework.Utils.SourceGeneratorUtils.GetDependency<int>(d, typeof(osu.Framework.Tests.Dependencies.SourceGeneration.DependencyContainerTest.Receiver11), null, null, true, false));
                }, null);
            }
```

The problem there is the call to the generic `GetDependency<T>` util. Note that `T = int` here (not `int?`, i.e. `Nullable<int>`), and as such this method:

https://github.com/ppy/osu-framework/blob/6f46341b1b2280a22e9db70505b2d509482abadf/osu.Framework/Utils/SourceGeneratorUtils.cs#L63-L76

is not *allowed* to return an `(int?)null` as the test expects it to.

Now the extra problem there is that a parent may be caching a `TValueType`, which should resolve lower down in a child that resolves `TValueType?`, i.e. `Nullable<TValueType>`.

I don't see a good way of solving this without having an extra helper method for value types specifically, that first attempts to query for cached `TValueType` (and maybe even `TValueType?`? although I'd hope not to do that second query because this introduces problems if *both* types end up being cached somehow...) and then falls back to `(TValueType?)null` upon failing to find one. I don't think this logic can be crammed into one generic `GetDependency<T>` that also works with reference types.

smoogipoo:
Thankfully, these are both issues that were caused by my refactoring and so I had plans for how to resolve them. Fixed in https://github.com/ppy/osu-framework/pull/5547.

osu.Framework.Tests/Dependencies/SourceGeneration/CachedModelDependenciesTest.cs
```diff
@@ -267,27 +267,31 @@ public void TestResolveIndividualProperties()
...
+        [Cached]
         private partial class NonReadOnlyFieldModel
         {
 #pragma warning disable 649
             public Bindable<int> Bindable;
 #pragma warning restore 649
         }
 
+        [Cached]
         private partial class PropertyModel
```
smoogipoo, line 295: These are unfortunate. The analyser is a bit more thorough in detecting where classes need to be partial than the source generator detects where it needs to implement. In this case the analyser detects it because it's used as an argument to `CachedModelDependencyContainer<T>` but the source gen only cares about how the class itself is structured.

It's a bit weird that this disparity exists, but I'll probably look into this in the future, perhaps by disallowing arbitrary objects and instead requiring use of an `IDependencyInjectionCandidate`, since doing so would reduce a lot of complexity.

osu.Framework.Tests/Dependencies/SourceGeneration/CachedAttributeTest.cs
```diff
@@ -472,10 +472,12 @@ private partial class Provider25 : IProviderInterface3
         {
         }
 
+        [Cached]
         private partial class Provider26 : Provider1
```
smoogipoo, line 476: The source gen also misses this one, where the `[Cached]` is on a base type. This only comes up with standard non-`Drawable` classes, and is potentially ordering dependent.

Added test for this here: https://github.com/ppy/osu-framework/pull/5547/commits/d646d4c3b73ae96b336ad0c943497d751fdc8593 but it's a bit more involved to fix.

For now I've fixed the test by adding `[Cached]` here also.

peppy:
New failures here.

smoogipoo:
It’s just the nupkg needs updating.

peppy:
Aha!

osu.Framework.Tests/Dependencies/SourceGeneration/CachedAttributeTest.cs
```diff
@@ -477,7 +477,6 @@ private partial class Provider26 : Provider1
         {
         }
 
-        [Cached]
```
bdach: Tests are failing due to this removal now, I think? Should be enough to revert the corresponding change in the test [here](https://github.com/ppy/osu-framework/pull/5543/commits/772c13abdb19fc8e5f9fd6a180f564f19fde3fa3#diff-d9237b657f351892884fef4669d29e4f5d2f3c73f653e5da5334c6abb025e82eR300).

bdach: I've just pushed the assertion revert myself in https://github.com/ppy/osu-framework/pull/5543/commits/470439d736c342bf5e4e5dc46dcf333652685fce because I can't possibly see a world where it would be incorrect.

Of note, the [reflection variant](https://github.com/smoogipoo/osu-framework/blob/470439d736c342bf5e4e5dc46dcf333652685fce/osu.Framework.Tests/Dependencies/Reflection/CachedAttributeTest.cs#L293-L303) and the [source generation variant](https://github.com/smoogipoo/osu-framework/blob/470439d736c342bf5e4e5dc46dcf333652685fce/osu.Framework.Tests/Dependencies/SourceGeneration/CachedAttributeTest.cs#L292-L302) of the test in question are now identical again, so there is no change in behaviour between reflection-based and SG-based DI in this particular case.

peppy (approved):

PR #5545, opened 2022-11-25
Use fully qualified name to ensure uniqueness

smoogipoo:
Using fully qualified names seems to stop rider from reprocessing files. @peppy please re-test this. Performance may not be as good as with incremental generator, but still seems acceptable to me.

Using `IIncrementalGenerator` will take a bit more work.


peppy:
When would lower performance be noticed? After making code changes?

smoogipoo:
Yeah, the same scenario as when processing files would trigger prior to this. It _seems_ a little bit slower to me but I want your second opinion.

peppy:
I've tested on two different PCs and cannot notice the overhead in a meaningful way. I think the CPU usage may be a bit higher but this is definitely usable 👍 .

peppy (approved):

PR #5546, opened 2022-11-26
Add support for macOS ctrl click (right click emulation)

Joehuu:
- [ ] Resolves https://github.com/ppy/osu/issues/6612
- [X] Depends on https://github.com/flibitijibibo/SDL2-CS/pull/240
- [ ] Depends on https://github.com/ppy/SDL2-CS bump

Note that this conflicts with `ControlPressed` cases like control-enter in song select for example. I was hoping to redirect it with cmd but there is also a `SuperPressed`. Not sure how to handle that.

### DIFF
--- a/osu.Framework/Platform/MacOS/MacOSWindow.cs
+++ b/osu.Framework/Platform/MacOS/MacOSWindow.cs
@@ -6,6 +6,7 @@
 using System;
 using osu.Framework.Platform.MacOS.Native;
 using osuTK;
+using SDL2;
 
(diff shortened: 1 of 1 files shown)


peppy:
Without checking the code, it sounds very weird that this should affect ctrl+enter.

Joehuu:
Never mind, I mistook ctrl-clicking carousel item/logo with ctrl-enter because it's in the same `ControlPressed` flow in song select.

Not sure if there are other conflicts then (editor?), haven't checked.

PR #5547, opened 2022-11-26
Fix incorrect source gen in edge cases of dependency injection

smoogipoo:
Fixes the test errors in https://github.com/ppy/osu-framework/pull/5543

Tested with o-f and osu.


osu.Framework.SourceGeneration/SyntaxHelpers.cs
```diff
@@ -72,7 +72,7 @@ public static InvocationExpressionSyntax GetDependencyInvocation(ITypeSymbol cal
                                                          SyntaxFactory.TypeArgumentList(
                                                              SyntaxFactory.SeparatedList(new[]
                                                              {
-                                                                 SyntaxFactory.ParseTypeName(GetUnderlyingType(requestedType))
+                                                                 SyntaxFactory.ParseTypeName(requestedType.ToDisplayString())
```
smoogipoo, line 75: The reason this was `GetUnderlyingType` before is because the source gen didn't emit NRT-enabled files at one point in time, and in such contexts `object?` is an invalid expression. But now this is fine since it does emit NRT-enabled files.

osu.Framework.SourceGeneration/SyntaxContextReceiver.cs
```diff
@@ -48,9 +48,9 @@ public void OnVisitSyntaxNode(GeneratorSyntaxContext context)
             }
 
             // Process any [Cached] attributes on any interface on the class excluding base types.
-            foreach (var iFace in symbol.Interfaces)
+            foreach (var iFace in SyntaxHelpers.GetDeclaredInterfacesOnType(symbol))
```
smoogipoo, line 51: The distinction here of "declared interfaces" is actually very important. One particular case osu! side looks like the following (simplified):

```csharp
// In the osu.Game project:

[Cached]
internal interface IOverlayManager { } // Note the use if internal.
public abstract class PlayerTestScene : IOverlayManager { }

// In the ruleset project:

public class TestSceneOsuPlayer : PlayerTestScene { }
```

"Declared" is important because `TestSceneOsuPlayer` cannot cache itself as an `IOverlayManager` due to not having access to that interface. It has to rely on the declarer of that interface (`PlayerTestScene`) to do that work.

smoogipoo (commented):
Just explaining a few things, in-case they're not self-explanatory.

peppy (approved):

bdach (approved):
Tested to fix the failures via a local nuget package.

PR #5548, opened 2022-11-26
Simplify resolution of dependency injection candidates

smoogipoo:
A common theme I observed is that finding candidate classes for dependency injection has considerable overheads, both in terms of correctness and likely also in performance (since more symbols have to be analysed otherwise).

The latter is the reason why the source generator didn't do as aggressive of a job as the analyser in generating the dependency injection classes, but this resulted in a disparity between the two that couldn't exist. Either both of them should miss inspections, or both of them should not.

This PR takes the second route of not missing inspections (greatest correctness) by requiring dependency injection classes to implement the `IDependencyInjectionCandidate` interface. For the most part this is a negligible overhead - the only such class outside of tests that I'm aware of is the osu!-side `Room`.

As a result, not only has the analyser been simplified to a quarter of its original size, but the SG perfectly follows the inspections provided by the analyser. The future incremental source generator implementation will also benefit from this.

The downside is that this is a breaking change, but as I mentioned above it's not a big one.

# vNext

## Classes used in dependency injection need to implement the `IDependencyInjectionCandidate` interface

This does not affect `Drawable` subclasses.

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkDependencyInjection.cs
+++ b/osu.Framework.Benchmarks/BenchmarkDependencyInjection.cs
@@ -40,7 +40,7 @@ namespace osu.Framework.Benchmarks
         }
     }
 
-    public class ClassInjectedWithReflection
+    public class ClassInjectedWithReflection : IDependencyInjectionCandidate
     {
         // ReSharper disable once UnusedAutoPropertyAccessor.Local
         [Resolved]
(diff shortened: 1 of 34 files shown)


smoogipoo:
Will need a source generator nupkg update with the changes in this PR, for this PR >_>

peppy (approved):

PR #5549, opened 2022-11-26
Add a way to use different sensitivities for horizontal and vertical mouse movements.

Cai1Hsu:
This allows you to use different sensitivities for horizontal and vertical mouse movements. With this PR and [this commit](https://github.com/Cai1Hsu/osu/commit/0ad06c9ee8aaae0a2499236160640ee3e654c3fc), it's possible to use different sensitivities.

https://user-images.githubusercontent.com/47081443/204096543-62cbb78c-c085-4ade-a0fc-a627a3dafbe7.mp4


osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
```diff
@@ -28,12 +28,24 @@ public class MouseHandler : InputHandler, IHasCursorSensitivity, INeedsMousePosi
...
             MaxValue = 10,
             Precision = 0.01
         };
+        
+        public BindableDouble SensitivityY { get; } = new BindableDouble(1)
+        {
+            MinValue = 0.1,
+            MaxValue = 10,
+            Precision = 0.01
+        };
```
bdach, line 48: This is going to be subjective I guess, but I'm not a huge fan of the design of these properties.

- Does `UseSeparateSensitivity` even need to exist?
- `Sensitivity` is only the actual sensitivity if the above is off, else it *really* is `SensitivityX`.

This is the cheap way of making this work but I don't believe it to be the best one when it comes to future-proofing and general data structure design. I would much prefer that this went the longer route and made `Sensitivity` a `Bindable<Vector2>` or something. `UseSeparateSensitivity` does not need to live in framework, it can only be a UI control on the game side.

Would be looking on more opinions on this from other reviewers.

Cai1Hsu: That's true. If we made `Sensitivity` a vector2 then there's no need for `UseSeparateSensitivity` to exist.

osu.Framework/Bindables/BindableVector2d.cs
```diff
@@ -0,0 +1,132 @@
...
+
+using System;
+using osuTK;
+
+namespace osu.Framework.Bindables
+{
+    /// <summary>
+    /// Represents a <see cref="Vector2d"/> bindable with defined component-wise constraints applied to it.
+    /// </summary>
+    public class BindableVector2d : RangeConstrainedBindable<Vector2d>
```
peppy, line 12: we never use lower case `2d`. Please check other code.

peppy: in addition, if this is really being added it needs test coverage like other bindables. i'm not sure how worthwhile this is though.

osu.Framework/Input/Handlers/Mouse/MouseHandler.cs
```diff
@@ -28,10 +28,10 @@ public class MouseHandler : InputHandler, IHasCursorSensitivity, INeedsMousePosi
             Description = "Allows for sensitivity adjustment and tighter control of input",
         };
 
-        public BindableDouble Sensitivity { get; } = new BindableDouble(1)
+        public BindableVector2d Sensitivity { get; } = new BindableVector2d(new Vector2d(1, 1))
```
peppy, line 31: This is going to break all configuration for all projects, and thus requires migration.

peppy (changes requested):

Cai1Hsu:
@peppy I think everything is ready though I haven't tested the changed code in [WindowsMouseHandler.cs](https://github.com/ppy/osu-framework/pull/5549/files#diff-115e2e1b939421143608b3cfa405282e586ae0f13905d445bc98294a28d0cae4) because I don't have a PC running Windows. But it should work properly.

PR #5551, opened 2022-11-28
Additive blend without drawcall

Tom94:
Moves the distinction between "Additive" and "Mixture" blend modes into the shader. This is controlled by the sign bit of each vertex's alpha channel, which feels slightly hacky, but is the least-intrusive solution I could think of. (And it's basically free, perf-wise.)

A necessary component of this change is the use of premultiplied alpha in the shader outputs, which means that buffered containers need special treatment: their framebuffers become textures with premultiplied alpha, which needs to be unmultiplied further down the line (for backwards compatibility). In this PR, this is handled by an additional shader uniform that denotes whether the input texture has premultiplied alpha or not.

Another thing of note: in the interest of API compatibility, additive blending on the `Drawable` side is controlled by an extra boolean within `BlendingParameters`, which feels hacky. I'd prefer removing the distinction between additive and mixture blend modes as far as the graphics API is concerned... but that'd require changing every single place where additive blending is currently set.

Lastly: some `osu-resources`-side shaders also need to have their output multiplied by alpha. I can make another PR for those -- it's a minor change. Until then, isolated places like osu!'s intro shader look broken.

### DIFF
--- a/osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
+++ b/osu.Framework/Extensions/Color4Extensions/Color4Extensions.cs
@@ -67,13 +67,13 @@ namespace osu.Framework.Extensions.Color4Extensions
         }
 
         /// <summary>
-        /// Returns a version of the color with negated alpha if the parameter is true.
+        /// Returns a version of the color with negated components depending on arguments.
         /// Used for the shader-level additive blend mode.
         /// </summary>
         /// <param name="colour">Original colour</param>
-        /// <param name="value">Negates alpha if true</param>
-        public static Color4 NegateAlphaIfTrue(this Color4 colour, bool value) =>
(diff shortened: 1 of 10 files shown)


peppy:
I think this looks like a pretty good direction to take things, but will note that `TestSceneComplexBlending` no longer work at my end:

https://user-images.githubusercontent.com/191335/204737607-d23526b3-6cf6-47f3-ac54-e72ec5556148.mp4

Tom94:
Fixed!

This problem shed light on a few problems osu! has with its blending in general. A few of those, I could fix and should make the game look slightly more correct looking forward. A few others of those are too late to be adjusted (or shouldn't be adjusted for aesthetic reasons, such as the design decision to do most, but not all, blending in sRGB space), so I added workarounds for these, along with verbose comments.

PR #5556, opened 2022-11-30
Add multi target file to source generator package

smoogipoo:
Following the template: https://github.com/dotnet/runtime/blob/main/eng/MultiTargetRoslynComponent.targets.template

And examples:
- https://github.com/ngocnicholas/airtable.net/blob/master/packages/System.Text.Json.6.0.5/build/System.Text.Json.targets
- https://github.com/reactiveui/refit/blob/main/Refit/targets/refit.targets

The intention is to make older SDKs - those used by Xamarin for our mobile compiles, select the correct analyser.


smoogipoo:
Looks good - iOS and Android builds passing.

peppy (approved):

Issue #5156, opened 2022-05-04
Disposed drawable can get a `FocusLostEvent` if removed while focused


Susko3:
Discovered when working on https://github.com/ppy/osu-framework/pull/5150#discussion_r863843183, when moving things from `unbindInput()` to `OnFocusLost()` and `Dispose()`.

Another way to put it: `InputManager.FocusedDrawable` can hold a disposed drawable.

Since focus is only updated when input manger updates, the disposed and focused drawable gets unfocused in `unfocusIfNoLongerValid()` and receives the focus loss event.

Lowering the update rate in the test will increase the failure rate.

Failure from the test below:

```
2022-05-04 19:51:23 [verbose]: 🔸 Step #20 click
2022-05-04 19:51:23 [debug]: ClickEvent(Left) handled by textbox 3(BasicTextBox).
2022-05-04 19:51:23 [debug]: Focus changed from nothing to textbox 3(BasicTextBox).
2022-05-04 19:51:23 [debug]: MouseClick handled by textbox 3(BasicTextBox).
2022-05-04 19:51:23 [verbose]: 🔸 Step #21 textbox focused
2022-05-04 19:51:23 [verbose]: 🔸 Step #22 clear and dispose children
2022-05-04 19:51:23 [debug]: Focus on "textbox 3(BasicTextBox)" no longer valid as a result of unfocusIfNoLongerValid.
2022-05-04 19:52:45 [error]: An unhandled error has occurred.
2022-05-04 19:52:45 [error]: NUnit.Framework.AssertionException: textbox "textbox 3" disposed:
2022-05-04 19:52:45 [error]: at osu.Framework.Logging.ThrowingTraceListener.Fail(String message1, String message2) in C:\Users\Mateo\code\git\osu-framework\osu.Framework\Logging\ThrowingTraceListener.cs:line 25
2022-05-04 19:52:45 [error]: at System.Diagnostics.TraceInternal.Fail(String message, String detailMessage)
2022-05-04 19:52:45 [error]: at System.Diagnostics.TraceInternal.TraceProvider.Fail(String message, String detailMessage)
2022-05-04 19:52:45 [error]: at System.Diagnostics.Debug.Fail(String message, String detailMessage)
2022-05-04 19:52:45 [error]: at System.Diagnostics.Debug.Assert(Boolean condition, String message)
[...]

PR #5557, opened 2022-12-01
Stop `OnDrag`, `OnDragEnd`, `OnMouseUp` etc. from firing on removed / disposed drawables

peppy:
RFC

Closes ppy/osu#21294 and other similar cases.

Interested in whether people agree with this direction or not. I think there's valid arguments in both directions, but I'm leaning towards this feeling more developer / implementation friendly.

A similar approach could/should be applied to focus events to address issues like #5156 if this direction is decided on. Would likely be as simple as adding an `IsRooted` check around https://github.com/ppy/osu-framework/blob/2e3e66598a5ef14c8810cf498452fb3552975218/osu.Framework/Input/InputManager.cs#L409


osu.Framework/Input/ButtonEventManager.cs
```diff
@@ -89,15 +88,17 @@ private bool handleButtonDown(InputState state)
...
         /// <summary>
         /// Handles the button being released.
         /// </summary>
         /// <param name="state">The current <see cref="InputState"/>.</param>
         private void handleButtonUp(InputState state)
         {
-            HandleButtonUp(state, ButtonDownInputQueue);
+            Debug.Assert(ButtonDownInputQueue != null);
+
+            HandleButtonUp(state, ButtonDownInputQueue.Where(d => d.IsRootedAt(InputManager)).ToList());
```
smoogipoo, line 101: This is... expensive... Potentially an O(n^2) operation in case no clicks are ever handled (i.e. click in dead space, so like the background in gameplay).

osu.Framework/Input/ButtonEventManager.cs
```diff
@@ -1,12 +1,11 @@
+#nullable enable
```
smoogipoo, line 1:
```suggestion
```

osu.Framework/Input/MouseButtonEventManager.cs
```diff
@@ -75,7 +64,7 @@ protected MouseButtonEventManager(MouseButton button)
         /// <summary>
         /// The drawable which is clicked by the last click.
         /// </summary>
-        protected WeakReference<Drawable> ClickedDrawable = new WeakReference<Drawable>(null);
+        protected WeakReference<Drawable> ClickedDrawable = new WeakReference<Drawable>(null!);
```
smoogipoo, line 67: `TryGetTarget()` seems [annotated accordingly](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/WeakReference.T.cs,59), so maybe this can just be nullable?

osu.Framework/Input/MouseButtonEventManager.cs
```diff
@@ -200,22 +193,38 @@ private void handleDragStart(InputState state)
...
+                handleDragDrawableBegin(draggable);
         }
 
-        private void handleDragEnd(InputState state)
+        private void handleDragDrawableBegin(Drawable drawable)
         {
-            DragStarted = false;
+            DraggedDrawable = drawable;
+            DraggedDrawable.IsDragged = true;
+            DraggedDrawable.Invalidated += draggedDrawableInvalidated;
```
smoogipoo, line 205: Is this ever unbound?

smoogipoo (changes requested):
Primary concern is performance. Need to mull over this one more

osu.Framework/Input/MouseButtonEventManager.cs
peppy: It gets weird if doing this (you'd have to `Debug.Assert` or conditionally check the `Drawable` is non-null after `TryGetTarget`), so I'd prefer to leave it like this.

osu.Framework/Input/MouseButtonEventManager.cs
peppy: Oops maybe not.

osu.Framework/Input/ButtonEventManager.cs
peppy: Yeah. Needs consideration.

frenzibyte: An alternative way to approach this would be propagating to the intersection of the button down input queue and the current input queue (similar to how it's done in `handleClick`).

smoogipoo: Yeah, I'd thought there was an existing case that did that. Sounds reasonable.

peppy: This doesn't work because we want to send `MouseUp` even when the cursor has moved away from the drawable. Using the `Intersect` method will break this.

```diff
diff --git a/osu.Framework/Input/ButtonEventManager.cs b/osu.Framework/Input/ButtonEventManager.cs
index fe63172e2..37962586e 100644
--- a/osu.Framework/Input/ButtonEventManager.cs
+++ b/osu.Framework/Input/ButtonEventManager.cs
@@ -97,7 +97,7 @@ private void handleButtonUp(InputState state)
         {
             Debug.Assert(ButtonDownInputQueue != null);
 
-            HandleButtonUp(state, ButtonDownInputQueue.Where(d => d.IsRootedAt(InputManager)).ToList());
+            HandleButtonUp(state, ButtonDownInputQueue.Intersect(InputQueue).ToList());
             ButtonDownInputQueue = null;
         }
 
```

Breaks this test:

https://github.com/ppy/osu-framework/blob/edb86a430aef51459af8e72f93e8cf6905293dd8/osu.Framework.Tests/Visual/Input/TestSceneMouseStates.cs#L409-L422

bdach: So first of all...

> Potentially an O(n^2) operation in case no clicks are ever handled (i.e. click in dead space, so like the background in gameplay).

It does look expensive yes, but I'm not sure the performance characterisation above is correct? `ButtonDownInputQueue` is being enumerated eagerly due to the `.ToList()` at the end, so it means that the _entire_ button down queue has to be checked for rootedness, _all_ of the time. So doing this really just punishes deep drawable hierarchies, I think?

That said, I did some profiling, and I'm not sure whether I messed anything up, but it... doesn't look that tragic? I tested both on a [test scene I wrote quickly to exercise deep hierarchies](https://gist.github.com/bdach/715b618a1cd0ce8ad62518f08d5164ef), and game-side, and got this back:

| test | master | this PR |
| :- | :-: | :-: |
| test scene (N=20 boxes) |  |  |
| test scene (N=40 boxes) |  |  |
| game side |  |  |

which... doesn't look tragic?

That said, if the above is still a worry, then here's an idea: We already have `Drawable.GetContainingInputManager()`, and use it relatively often for stuff. How about if we made a property out of that? Would be

```csharp
public InputManager ContainingInputManager { get; }
```

and it would be backed by a `LayoutValue` which re-validates after `Invalidation.Parent`. It's probably going to be a bit involved to make happen, but maybe worthwhile?

bdach:
Looks fine to me, except for the performance discussion, but [I can't exactly see it being disastrous in profiling](https://github.com/ppy/osu-framework/pull/5557#discussion_r1271288786) - unless I'm looking wrong I guess.

osu.Framework/Input/ButtonEventManager.cs
peppy: I think the suggestion sounds good, but likely adds complexity that we don't need until we hit a real-world performance issue from this. The profiling makes it looks maybe unnecessary. The only reason I'd consider it is that it's the kind of performance overhead which could go unnoticed, as it is only incurrent on a mouse click.

bdach: There's also the part where I recall you saying a while back that you wanted input handling to be zero-alloc. I don't believe we ever fully got there, but there was a decent push to eliminate those a while back, and doing this `.ToList()` dump adds one again.

peppy:
Passes osu-side tests FWIW

2023-07-22 22 28 30@2x

bdach (dismissed):

smoogipoo:
I'm seemingly unable to profile this effectively, but putting a stopwatch around it shows it usually spends anywhere from 0.2 to 0.5ms on average inside that method, sometimes spiking to several milliseconds.

I'll tentatively okay this for a merge, though I'm weary that due to the nature of the hierarchy any issues arising from this won't be discoverable.

osu.Framework/Graphics/DrawableExtensions.cs
```diff
@@ -45,6 +45,23 @@ public static T With<T>(this T drawable, Action<T> adjustment)
...
+        /// <param name="drawable">The <see cref="Drawable"/> to be checked.</param>
+        /// <param name="root">The root to be checked against.</param>
+        /// <returns>Whether the drawable was rooted.</returns>
+        internal static bool IsRootedAt(this Drawable drawable, Drawable root)
+        {
+            while ((drawable = drawable.Parent) != null)
+            {
+                if (drawable == root)
+                    return true;
+            }
```
smoogipoo, line 60: Shouldn't a drawable be "rooted" at itself?

smoogipoo:
As discovered in Discord convo, my stopwatch was profiling the whole `HandleButtonUp()` method call. It looks like this is well under 0.1ms in general.

osu.Framework/Graphics/DrawableExtensions.cs
peppy: as in this check should be done before the first parent navigation?

bdach: Sounds like it. Either that, or `do..while` I guess (although I'm never super sure if I'm correctly converting `while` to `do..while` in such cases...)

peppy: I prefer just having an explicit line here. Reads better.

bdach (approved):

PR #5559, opened 2022-12-01
Rework source generator to promote memoisation

smoogipoo:
I have only 60% confidence that this works and won't break in some spectacular way. Caching is one of the hardest things in programming, and combined with Roslyn-level bugs means this is very temperamental.

The general idea here is to use the `ClassDeclarationSyntax` as the equality comparer for `GeneratorClassCandidate`, because this is an immutable object that persists through compilations. Thus, `GeneratorClassCandidate` has to be split into two passes, the first of which is initialising the object (for caching purposes), and the second is populating the semantic members.  
I'm doing something extremely dodgy here in storing the `SemanticModel` at the first stage and clearing it in the second stage, but I'm doing this to avoid a `Compilation`. It seems to work but add it to the list of things that could break and contributes to my 60% confidence level.

My test where I noticed temperamental behaviour is to add a new `public partial class Game : IDependencyInjectionCandidate { }` at the top of `CustomInputManager.cs`. For some reason, adding an `IsValid` check to the pipeline breaks the whole thing with duplicate file errors. I don't know why this is, so I've left the `IsValid` check up to the emit stage.

Before:
image

After:
image

- Tested in osu-framework because my profiling session doesn't capture it in osu.
- Testing process is to add ~70 newlines in `Game.cs`, waiting for the rider inspection to underline the current line before inserting a new one each time.

### DIFF
--- a/osu.Framework.SourceGeneration/Common/Emitters/DependenciesFileEmitter.cs
+++ b/osu.Framework.SourceGeneration/Common/Emitters/DependenciesFileEmitter.cs
@@ -39,7 +39,7 @@ namespace osu.Framework.SourceGeneration.Emitters
             Candidate = candidate;
         }
 
-        public string Emit()
+        public void Emit(AddSourceDelegate addSource)
         {
             StringBuilder result = new StringBuilder();
             result.Append(headers);
@@ -60,7 +60,11 @@ namespace osu.Framework.SourceGeneration.Emitters
                                  .NormalizeWhitespace());
             }
 
-            return result.ToString();
+            // Fully qualified name, with generics replaced with friendly characters.
+            string typeName = Candidate.FullyQualifiedTypeName.Replace('<', '{').Replace('>', '}');
+            string filename = $"g_{typeName}_Dependencies.cs";
+
+            addSource(filename, result.ToString());
         }
 
         private MemberDeclarationSyntax emitDependenciesClass()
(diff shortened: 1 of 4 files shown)


X9VoiD (commented):
From [#osu-framework](https://discord.com/channels/188630481301012481/589331078574112768/1047932915889213480):

> this would have caused issues with reusing a potentially outdated SemanticModel, but if everything needed is only inside the class declaration then that'll be fine (not sure about multi partial)

In detail, SemanticModel allows one to query information that is syntactically "outside" of a syntax. Take this code for example:

```cs
partial class Class1
{
    // some DI stuff inside
}

partial class Class1
{
    // some other DI stuff inside
}
```

When querying using the `SemanticModel`, it merges the partial definitions into one. The behavior of the incremental generator would then change depending on which class declaration "won" the `Distinct()` call.

If it's the first one, relevant code changes (that affects code generation) on the second class declaration will incorrectly reuse the cached `GeneratorClassCandidate` (which is by now, contains outdated information). And the same is true vice-versa.

(I honestly can't think of a minimal example right now of other caching issues that doesn't include multi-partial)

Is there any situation where using multi partials are beneficial/needed? If not, it might be worthwhile to add an analyzer diagnostic that specifically prevents/warns on using multi partial for candidate classes.

smoogipoo:
I’m going to block this for now following on from the above comment and further discussion in Discord where it was discovered that this has other theoretic edge cases such as cached definition on interface types.

There’s still things to try so this isn’t the end of the line, but it only gets more complicated.

smoogipoo:
Superseded by https://github.com/ppy/osu-framework/pull/5560

PR #5562, opened 2022-12-03
Add multi phase source generator test

X9VoiD:
Allows testing of incremental code changes similar to what was manually done with #5560

This PR introduces a new verifier that supports all test cases of the old one, plus test cases that have filenames of the format:

```
<filename>.<phase>.cs
```

where:
 - all sources of the same phase will be compiled and be passed to the generator together,
 - and the generator's generated sources will be verified against generated sources of the same phase.

It is hard to determine whether the new verifier actually behaves the way we expect it, so I added a known failing case "MultiPartialResolvedMember" as shown in #5560, then also prepared two branches which runs the same test before and after #5560 is merged.

- [multi-phase-test-pre-fix](https://github.com/X9VoiD/osu-framework/tree/multi-phase-test-pre-fix)
- [multi-phase-test-post-fix](https://github.com/X9VoiD/osu-framework/tree/multi-phase-test-post-fix) 

The results are as we expected:
- `multi-phase-test-pre-fix` fails "MultiPartialResolvedMember" test.
- `multi-phase-test-post-fix` passes all.

Also added a way to listen to what the generator is doing and inspect its state, which this PR takes advantage of to determine if caching works as we expect it.


osu.Framework.SourceGeneration.Tests/osu.Framework.SourceGeneration.Tests.csproj
```diff
@@ -9,6 +9,10 @@
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.SourceGenerators.Testing.XUnit" Version="1.1.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.5.0-preview-20221003-04" />
     <PackageReference Include="xunit" Version="2.4.2" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.5">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+    </PackageReference>
```
bdach, line 15: Is this new dependency required? Not that it looks suspicious, just wondering why it's here at all.

osu.Framework.SourceGeneration.Tests/Verifiers/CSharpMultiPhaseSourceGeneratorVerifier.cs
```diff
@@ -0,0 +1,85 @@
...
+                            multiPhaseList.Add(new());
+
+                        string filename = match.Groups["filename"].Value + ".cs";
+                        multiPhaseList[phase].Add((filename, s.content));
+                    }
+                    // Not multi-phase, at this point all the files should be single-phase
+                    else
+                    {
+                        if (multiPhaseList.Count == 0)
+                            multiPhaseList.Add(new());
```
bdach, line 62: Not sure we like usages of target-typed `new()` round these parts...

osu.Framework.SourceGeneration.Tests/osu.Framework.SourceGeneration.Tests.csproj
X9VoiD: Needed it for test discovery in `dotnet test` and vscode.

osu.Framework.SourceGeneration.Tests/GeneratorHelper.cs
```diff
@@ -0,0 +1,116 @@
...
+            int matches = 0;
+            foreach (var (filename, content) in files)
+            {
+                if (!generatedSources.TryGetValue(filename, out var source))
+                    throw new Xunit.Sdk.XunitException($"Phase {phase}: Expected generated source {filename}, but it was not found");
+
+                string actual = source.SourceText.ToString();
+
+                if (actual != content)
+                    throw new Xunit.Sdk.XunitException($"Phase {phase}: Generated source {filename} did not match expected content");
```
smoogipoo, line 58: Is there no way to generate the diff here? Trying to merge master in this branch and, although I know what to change, it's kind of unfortunate not seeing exactly what has changed.

smoogipoo: Seems like the following works:
```
new XUnitVerifier().EqualOrDiff(content, actual, $"Phase {phase}: Generated source {filename} did not match expected content");
```

smoogipoo (approved):
I've done some very rudimentary review on this. The verifiers/helper classes are kinda unnecessary to focus on for review as long as they work, imo.

Issue #5563, opened 2022-12-03
DI source generator does not handle member names that conflict with namespaces


frenzibyte:
Originally reported in [discord chat](https://discord.com/channels/188630481301012481/589331078574112768/1048711155209752606) with the following error:

> 'OsuGameBase' does not contain a definition for 'Framework' and no accessible extension method 'Framework' accepting a first argument of type 'OsuGameBase' could be found (are you missing a using directive or an assembly reference?)

Happens due to the original component declaring a member with the name `osu`, which gets prioritised over the `osu` namespace, thus showing the error seen above. A general fix would be to prefix all class references with `global::` to avoid such conflict (originally proposed [here](https://discord.com/channels/188630481301012481/589331078574112768/1048712345033117876)).

bdach:
I need to sleep, so here's an update for now. This gets worse when considering nested types. The nice and easy case is

```csharp

public partial class ConflictGenerator : osu.Framework.Graphics.Drawable
{
    private bool osu;

    [osu.Framework.Allocation.Cached]
    public object CachedProperty { get; set; }

    [osu.Framework.Allocation.Resolved]
    public object ResolvedProperty { get; set; }

    [osu.Framework.Allocation.BackgroundDependencyLoader]
    private void load(object dependency)
    {
    }
}
```

as in the OP, which can only affect instance method bodies. The horrible, terrible, no-good case is

```csharp
namespace osu;

public partial class NestedConflictGenerator : osu.Framework.Graphics.Drawable
{
    [global::osu.Framework.Allocation.BackgroundDependencyLoader]
    private void load()
    {
    }

    [global::osu.Framework.Allocation.Resolved]
    private object resolvedMember { get; set; }

    public class osu
    {
    }
}
```

wherein the inner `osu` class can type with *any and every* typename spec emitted inside the partial class in the generated files.

String-appending `global::` is incorrect (except for places where the type names are already full strings) because `string::object` is not correct. [`SymbolDisplayFormat.FullyQualifiedFormat`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.symboldisplayformat.fullyqualifiedformat?view=roslyn-dotnet-4.3.0) looks promising as a remedy.

I'm willing to take a stab at this tomorrow but will yield if told to.

PR #5565, opened 2022-12-04
Emit `global::` prefix before type names in DI source generator

bdach:
I'm not 100% certain of this diff, but it does look to be working (generator tests, framework tests and game tests all passing), and it should close https://github.com/ppy/osu-framework/issues/5563 at its current state.

The goal of this change is to protect source-generated DI implementations from breaking down due to different namespacing confusions, including (but probably not limited to):

- Local members in classes whose names coincide with a top-level namespace, i.e.:

  ```csharp
  public partial class ConflictGenerator : osu.Framework.Graphics.Drawable
  {
      private bool osu;
  
      [osu.Framework.Allocation.Cached]
      public object CachedProperty { get; set; }
  
      [osu.Framework.Allocation.Resolved]
      public object ResolvedProperty { get; set; }
  
      [osu.Framework.Allocation.BackgroundDependencyLoader]
      private void load(object dependency)
      {
      }
  }
  ```

  In this case, attempting to call any method via bare FQN which begins with the top-level `osu` namespace (i.e. the public injection helpers) leads to a compile error as those calls are redirected to the field member.

- Nested classes whose names coincide with a top-level namespace, i.e.:

  ```csharp
  namespace osu;
  
  public partial class NestedConflictGenerator : osu.Framework.Graphics.Drawable
  {
      [global::osu.Framework.Allocation.BackgroundDependencyLoader]
      private void load()
      {
      }
  
      [global::osu.Framework.Allocation.Resolved]
      private object resolvedMember { get; set; }
  
      public class osu
      {
      }
  }
  ```

  In this case, attempting to reference *any type* via bare FQN which begins with the top-level `osu` namespace leads to a compile error as those types are presumed to live inside of the `osu` inner class.

In both cases, it is sufficient to prefix the type with a `global::` to escape the name-resolution-confusing local members, which this diff attempts to do for all possible type and member references that could be emitted by the SG and that could fall victim to this sort of type confusion.

This diff looks huge, but most of it is spent on updating tests. The actual change volume is not *that* high.

There are some caveats in this diff. I will be annotating them in more details with review comments shortly after opening.


osu.Framework.SourceGeneration/Common/GeneratorClassCandidate.cs
```diff
@@ -15,7 +15,7 @@ public class GeneratorClassCandidate
         public readonly ClassDeclarationSyntax ClassSyntax;
 
         public readonly string FullyQualifiedTypeName = string.Empty;
-        public readonly string TypeName = string.Empty;
+        public readonly string GlobalPrefixedTypeName = string.Empty;
```
bdach, line 18: In most places, I was able to just replace the standard `TypeName` with the `GlobalPrefixedTypeName`. This is one place where this was not possible, as `FullyQualifiedTypeName` is already being used to generate names for the generated source files. I could *attempt* to remove one of these, but the main complication to doing this is that while standard type names can be prefixed with `global::`, type keywords such as `object` or `int` cannot.

The viable directions are:
- Change `FullyQualifiedTypeName` here to ensure it never uses type keywords (i.e. instead of `object` use `System.Object`), and then remove `GlobalPrefixedTypeName`, resorting to string-prepending `global::` instead.
- Remove `FullyQualifiedTypeName`, and snip the leading `global::` instead when generating the source filenames.

osu.Framework.SourceGeneration/Common/Emitters/DependenciesFileEmitter.cs
```diff
@@ -78,7 +78,7 @@ private MemberDeclarationSyntax emitDependenciesClass()
                            SyntaxFactory.BaseList(
                                SyntaxFactory.SingletonSeparatedList<BaseTypeSyntax>(
                                    SyntaxFactory.SimpleBaseType(
-                                       SyntaxFactory.ParseTypeName("osu.Framework.Allocation.ISourceGeneratedDependencyActivator")))))
+                                       SyntaxFactory.ParseTypeName("global::osu.Framework.Allocation.ISourceGeneratedDependencyActivator")))))
```
bdach, line 81: I don't have a test case ready to prove that type names in the implementation list are also susceptible to namespacing confusions, but I think one could be constructed. It'd probably require some sort of screwed up namespace setup so that you have

```csharp
namespace Test
{
    public class DependencyParticipant  : Drawable
    {
    }

    namespace osu
    {
    }
}
```

therefore inadvertently making the supposedly-FQN type reference relative to the `Test` namespace because there is a child `osu` namespace that takes priority.

Every other type reference in the generated source is 100% susceptible to some sort of namespacing skullduggery, and there is new test coverage in this pull to cover those.

osu.Framework.SourceGeneration/Common/SyntaxHelpers.cs
```diff
@@ -115,6 +115,25 @@ public static bool IsIDependencyInjectionCandidateInterface(ITypeSymbol? type)
...
+        /// This method does not return <see langword="null"/> if <paramref name="type"/> is not <see langword="null"/>.
+        /// This cannot be expressed via <c>NotNullWhenNotNullAttribute</c> due to targeting .NET Standard 2.0.
+        /// </para>
+        /// </remarks>
+        public static string? GetGlobalPrefixedTypeName(ITypeSymbol? type)
+            => type?.ToDisplayString(fullyQualifiedFormatWithNullableAnnotations);
+
+        private static SymbolDisplayFormat fullyQualifiedFormatWithNullableAnnotations { get; } =
+            SymbolDisplayFormat.FullyQualifiedFormat
+                               .AddMiscellaneousOptions(SymbolDisplayMiscellaneousOptions.IncludeNullableReferenceTypeModifier);
```
bdach, line 135: - This method resolves the issue of being able to tell whether it is safe to prepend `global::` to the type name. `SymbolDisplayFormat.FullyQualifiedFormat` has `SymbolDisplayMiscellaneousOptions.UseSpecialTypes` on. Making it never use type keywords should be as simple as switching that option on, in theory.
- I couldn't however use `SymbolDisplayFormat.FullyQualifiedFormat` directly because it doesn't print NRT annotations, hence the static member with them explicitly enabled.

osu.Framework.SourceGeneration/Common/GeneratorClassCandidate.cs
smoogipoo: I think this is fine as is. `FullyQualifiedTypeName` could probably be named something like `UniqueFileName` or something like that. It really doesn't matter what it is as long as it's unique across all files. Could even be a GUID really though I'm not sure on the uniqueness - technically should be okay since this class is unique for a `ClassDeclarationSyntax`.

smoogipoo (approved):
lgtm

PR #5568, opened 2022-12-05
Disable source generation on iOS

frenzibyte:
Turns out source generation breaks badly on a solution that includes at least one Xamarin.iOS project, and blocks me completely from building iOS on framework or osu!-side with local checkout:

CleanShot_2022-12-05_at_18 45 57

Here's a video showcasing what's going on in complete isolation:

https://user-images.githubusercontent.com/22781491/205741906-ae0217c7-1bec-4b23-b491-a0602858631d.mp4

Once `Xamarin.iOS.CSharp.targets` is imported to the iOS project, the SG analyzer is immediately intercepted from the project that references it and is brought to the iOS project that doesn't even reference it. I have not the slightest clue on how this is even possible, but that appears to be the case here.

Until we migrate iOS projects to .NET 6, disabling source generation is the only way forward as far as I can see.

### DIFF
--- a/osu.Framework.SourceGeneration/targets/ppy.osu.Framework.SourceGeneration.targets
+++ b/osu.Framework.SourceGeneration/targets/ppy.osu.Framework.SourceGeneration.targets
@@ -19,9 +19,12 @@
   </Target>
 
   <Target Name="_ppy_osu_Framework_SourceGenerationRemoveAnalyzers"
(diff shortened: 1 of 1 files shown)


peppy:
Going to close this for now. I think pushing towards getting net60 working is a better end goal. And I can still build and deploy iOS fine so this shouldn't be a requirement short term.

PR #5569, opened 2022-12-05
Temporarily disable NRT annotations detection of nullable dependencies

frenzibyte:
As this has backfired twice before (https://github.com/ppy/osu/pull/19677, https://github.com/ppy/osu/pull/21512), this feature is unreliable as long as mobile projects are not updated to .NET 6. Disabling it would avoid those catastrophes from happening again for the time being.

# vNext
## NRT-based detection of nullable dependencies has been temporarily disabled

This feature never worked on mobile projects since they're not targeting .NET 6 yet. We cannot continue to have this feature enabled for the time being.

### DIFF
--- a/osu.Framework.SourceGeneration/Common/Data/BackgroundDependencyLoaderParameterData.cs
+++ b/osu.Framework.SourceGeneration/Common/Data/BackgroundDependencyLoaderParameterData.cs
@@ -19,9 +19,13 @@ namespace osu.Framework.SourceGeneration.Data
         public static BackgroundDependencyLoaderParameterData FromParameter(IParameterSymbol parameter)
         {
             string globalPrefixedTypeName = SyntaxHelpers.GetGlobalPrefixedTypeName(parameter.Type)!;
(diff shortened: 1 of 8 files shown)


frenzibyte:
Missed resource-based source generation tests. I'll get to them later.

peppy:
How did it wor until now?

smoogipoo:
Are you actually sure that the source generator version is broken?

frenzibyte:
> How did it work until now?

It never did for mobile, the `IsNullable` extensions were always returning false there (notice the target framework conditional preprocessor there). This was only ever noticeable on NRT-annotated dependencies (`Type?`) which don't set `CanBeNull = true` in the attribute and don't actually have a value cached for it (i.e. the osu!-side PRs mentioned in OP).

> Are you actually sure that the source generator version is broken? Edit: I guess it's fine if the intent is to match behaviour...

Correct, it is to match behaviour so we don't rely on this feature at all for the time being (and source generation is busted on iOS).

peppy:
@frenzibyte shall we also close this out given that we are pushing for `net60`?

PR #5570, opened 2022-12-06
Fix broken iOS build configurations of template projects in framework solution

frenzibyte:
Build configuration copied from `SampleGame.iOS` project. Builds correctly on `osu-farmework.iOS` filter now.


peppy (approved):
Not quite ready to test iOS but I'll trust your changes here, looks sane enough.

PR #5574, opened 2022-12-07
Guard against attempting to call `OpenUrlExternally` with non-http content

peppy:


bdach:
Not 100% sure this solution will hold up without game-side adjustments. Game-side we have message parsing logic that will also attempt to [externally open some `osu://` protocol URLs](https://github.com/ppy/osu/blob/2fba6cbe769187c4cd473216fe39e918e4d56406/osu.Game/Online/Chat/MessageFormatter.cs#L236-L237). These usages will hard crash with this change.

I'm also side-eyeing the android change since android likes to use custom URI schemes to launch other intents. Not saying there's anything wrong there - I don't *think* we're using that for anything, but I'm not sure.

peppy:
I think we'll want to fix the cases osu!-side and still think this is the correct direction for this change.

smoogipoo:
Part of my worry is that something is going to be handling links of the form `google.com` without the scheme, and this is going to start crashing. One component I can think of immediately is `MarkdownContainer`, which does in-fact crash now.

I don't know whether `MessageFormatter` also causes issues, and there are seemingly no tests for whether it can handle non-schemed links, but best make sure before a link in the in-game chat causes crashes...

peppy:
Stable gets around this by prefixing `https://` in cases it's not already there. I think we can do that here as well? Or osu-side, either works for me.

smoogipoo:
Sure, I'm okay with that but have to go through those cases in this PR.

bdach (approved):
Attempted to break this in a few ways via chat (using the linked game-side branch) and failed to do so. Users will _probably_ find a way, but at least it shouldn't be immediately obvious...?

smoogipoo (approved):

PR #5575, opened 2022-12-07
Implement DrawQuad method which doesn't need a texture

EVAST9919:
While working on https://github.com/ppy/osu/pull/21527 I realized that some shaders are "standalone" (don't need a texture to be displayed). Also I really wanted to have a reliable control over coordinates which are being used in the shader, which is now available using this method.

Few things to discuss:
* xmldoc probably need some refinements
* whether `blendRangeOverride` needed in this method
* whether `inflationPercentage` needs to be added back (both prameters not used in all my cases so far and I'm not sure if they will be helpful at all)


EVAST9919:
Superseded by https://github.com/ppy/osu-framework/pull/5577

PR #5577, opened 2022-12-08
Fix WhitePixel texture rectangle doesn't respect provided area

EVAST9919:
Supersedes https://github.com/ppy/osu-framework/pull/5575 as an actual fix.

### DIFF
--- a/osu.Framework/Graphics/Textures/TextureWhitePixel.cs
+++ b/osu.Framework/Graphics/Textures/TextureWhitePixel.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-#nullable disable
-
 using osu.Framework.Graphics.Primitives;
 
 namespace osu.Framework.Graphics.Textures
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Textures/TextureWhitePixel.cs
```diff
@@ -21,7 +19,7 @@ public override RectangleF GetTextureRect(RectangleF? area = null)
             // Let's be very conservative and use a tenth of the size of a pixel in the
             // largest possible texture.
             float smallestPixelTenth = 0.1f / NativeTexture.MaxSize;
-            return new RectangleF(0, 0, smallestPixelTenth, smallestPixelTenth);
+            return base.GetTextureRect(area) * smallestPixelTenth;
```
smoogipoo, line 22: Does `base.GetTextureRect(area * smallestPixelTenth)` work for you? Not necessarily saying that's how it should be yet, but from experience the ordering of this method is a little bit weird in that all of the areas are applied first and then the resulting value is passed down the chain where it gets converted into UV coords.

EVAST9919: It still works for my use case, however other shaders such as `CircularProgress` and both ones from colour picker becoming broken. I'm prety sure they can be fixed but from the first glance not sure where the problem is.
With my original change they still work as expected.

smoogipoo (approved):

PR #5578, opened 2022-12-11
Fix search container filtering too much

bdach:
As it turns out, 99915f9286d10b12b00479346c30005832ee8823 introduced a performance regression in `SearchContainer`. Due to borrowing from #5154 too much (see [relevant discord conversation](https://discord.com/channels/188630481301012481/188630652340404224/1051580223709266011)), the search logic was changed in order to hook into `InvalidateLayout()` to perform search, rather than just do it when a new internal child was added, as was the case previously.

For the sake of completeness, let's review the possible causes for `InvalidateLayout()` firing with respect to filtering:

- `AddInternal()`: Valid, as the new child needs to be examined against the current filter, and preserved by reverting to the old code.
- `RemoveInternal()`: Mostly irrelevant, as long as the presence of the removed drawable in the container doesn't impact the filter state of any of its siblings. Let's not support that case for now.
- `ClearInternal()`: Totally irrelevant, as anything that could be filtered is now gone.
- `SetLayoutPosition()`: Hopefully irrelevant.
- `UpdateChildrenLife()`: Not relevant. Nothing in `SearchContainer` considers lifetime explicitly.
- `FlowContainer.childLayout` becoming invalid: The invalidation flags there are `RequiredParentSizeToFit | Presence`. Neither should be relevant for filtering, given the direction taken with `IConditionalFilterable`.
- `Direction` and `Spacing`: 100% irrelevant.

---

@peppy Would appreciate it if you ran this through your perf wringer and see if it helps any with the [blocking load issue pointed out on discord](https://discord.com/channels/188630481301012481/188630652340404224/1049974688836751360). This does nothing for the JIT overhead pointed out therein, but this pull should cull some hundreds of extraneous filters that could be triggered on _some_ (but not all, from my testing) game-side settings sidebar loads, and those extra calls mostly stemmed from the `childLayout` invalidations - so I before I dig in any further I want to see if this at least restores baseline performance.


peppy:
Change seems good, and may help more in other cases, but doesn't seem to do much here (as expected, the JIT is still the killer):

Before:

4,444 calls to `matchSubTree()`

JIT: 104ms
User: 22ms

After:

2,229 calls to `matchSubTree()`

JIT: 104ms
User: 20ms

peppy (approved):

PR #5582, opened 2022-12-15
Bring iOS framework to .NET 6

frenzibyte:
- [x] Depends on https://github.com/ManagedBass/ManagedBass/pull/118 (potentially fork, cc @peppy)

And we're finally here, moving away from Xamarin once and for all. I highly recommend reviewing this PR commit-by-commit, as the diff got unintentionally too large to review all at once.

Moving away from .NET Standard will come in a follow-up PR, as this one is already too large.

There's a lot to unfold in here, but I'll point out the main portions of the changes:

### (https://github.com/ppy/osu-framework/commit/601d7082c471e0015479ff7bbc6392f660379806) Replace static libraries with XCFrameworks 

In .NET 5+, dllmaps are no longer a thing. Instead, .NET offers `NativeLibrary` which provides a method for mapping dll names, i.e. "dllmaps" but at runtime. However, that doesn't work with static libraries, as they're accessible under the library name `__Internal`, which `NativeLibrary` appears to not have access to or able to load from.

Therefore, I've replaced all of our static libraries with XCFramework bundles of themselves, which `NativeLibrary` can work with:
```csharp
            NativeLibrary.SetDllImportResolver(typeof(Bass).Assembly, (_, assembly, path) => NativeLibrary.Load("@rpath/bass.framework/bass", assembly, path));
            NativeLibrary.SetDllImportResolver(typeof(BassFx).Assembly, (_, assembly, path) => NativeLibrary.Load("@rpath/bass_fx.framework/bass_fx", assembly, path));
            NativeLibrary.SetDllImportResolver(typeof(BassMix).Assembly, (_, assembly, path) => NativeLibrary.Load("@rpath/bassmix.framework/bassmix", assembly, path));
```

On the bright side, the mono symbol strip exclusion nonsense is no longer required.

Directly wrapping the static libraries in a `.xcframework` folder isn't feasible as it's supposed to store different versions of the library for iOS and iOS simulator, while our libraries are fat/universal. It might be possible to wrap each library in a `.framework` folder instead, but I saved myself the time with downloading proper bundles instead. 

BASS bundles were retrieved from [un4seen](https://www.un4seen.com), while FFmpeg bundles were downloaded off of [arthenica/ffmpeg-kit](https://github.com/arthenica/ffmpeg-kit/releases/tag/v4.4) (macos-catalyst is included in that bundle but I removed it to save space).

### (https://github.com/ppy/osu-framework/commit/327c7775b06133388b0e5d6ab9a96402ecd2ac4a) Target iOS 13.4+ and remove no longer used logic

Originally discussed in [discord](https://discord.com/channels/188630481301012481/589331078574112768/1051181970538905710), it turns out the legacy side of `IOSHardwareKeyboardHandler` (for old devices) was actually relying on newer API which it shouldn't have access to (`UIKeyboardHidUsage` enum, iOS 13.4+). This only popped up in my build output once I switched to .NET 6, where it's aware of the minimum specified OS version.

Rather than going through the hassle of replacing enum usage with int or otherwise, I've went with targeting iOS 13.4+ instead and call it a day.

Other changes are either described obvious or described in the commit's description or in the changes of the commit itself.

### (https://github.com/ppy/osu-framework/commit/2bc6f8a54a3acec78d228b9d49aab6b2e125227f) Remove pinning from BASS callback handles on iOS

In .NET 5+ (not sure about older versions), it's no longer possible to allocate pinned GC handles for delegates. ~~It might be wise to switch to [function pointers](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/unsafe-code#function-pointers)~~ (nevermind they don't work on instance delegates) but switching to "normal" (no pinning) mode doesn't cause GC to randomly collect callback handles as far as I can see (cc @smoogipoo, you might know about this better than I do).

### `--nostrip` and `--nolinkaway` have been removed

Those flags are no longer necessary as .NET iOS does not seem to rely on the `mtouch` linker anymore. And the size of the application (testing with osu!, physical device, and release configuration) on .NET iOS already reached half the size of a Xamarin build!

CleanShot 2022-12-11 at 23 22 23
(left is xamarin, right is .net)


osu.Framework.iOS.props
```diff
@@ -1,100 +1,40 @@
...
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|iPhoneSimulator' ">
-    <MtouchNoSymbolStrip>true</MtouchNoSymbolStrip>
+    <RuntimeIdentifier>ios-arm64</RuntimeIdentifier>
   </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|iPhone' ">
-    <DeviceSpecificBuild>true</DeviceSpecificBuild>
-    <IOSDebuggerPort>28126</IOSDebuggerPort>
+  <PropertyGroup Condition="'$(Platform)' == 'iPhoneSimulator'">
+    <RuntimeIdentifiers>iossimulator-x64</RuntimeIdentifiers>
   </PropertyGroup>
```
frenzibyte, line 15: I was hoping to remove the "iPhone" and "iPhoneSimulator" platform configurations, but as it turns out, selecting a device in an IDE alone doesn't seem to note the compiler/linker whether to build for RID `ios-arm64` or `iossimulator-x64`.

Specifying both RIDs in one `RuntimeIdentifiers` key doesn't work, as the RIDs of that key must belong to the same "os" name. And specifying `ios-x64` doesn't work either because it doesn't exist.

Not sure if there's some sort of condition that specifies whether the selected device is "simulator".

smoogipoo (approved):

peppy (approved):

peppy:
Awesome job on this 💯

PR #5588, opened 2022-12-16
Rework circular blob to use Bezier curves

EVAST9919:
- [ ] Needs mobile testing

**WHY**
Current blob implementation has a bunch of problems such as: no fine control over the shape, visual inconsistencies in some extreme cases and fill setting which is not really a fill but a stroke width. This pr solves these issues.

**PERFORMANCE**
Since there's no way to practically benchmark fps (or not that I'm aware of) I just used fps counter to determine the difference between implementations under similar circumstances. TL;DR: it's on-pair with the old implementation.

| master | pr |
| ----------- | ----------- |
| fr-old | fr-new |
| old | new |

**BETTER CONTROL OVER SHAPE**
There are a lot of cases in which old implementation just can't even remotely reproduce shapes, possible by the new one. IMO this the behaviour I'd expect from this component.
In extreme cases performance is bad, howewer we are talking about cases which are not even possible with the old implementation. Probably limiting `PointCount` to 50 points max is good idea, not sure.
impossible

**ADDITIONAL NOTES**
* Due to a lot of new stuff introduced in the shader - it really needs testing on mobile _(and I will be shocked if it works right away, but there's no way)_
* Algorithm used to create curve control points can be explained if needed.
* This pr is really low priority since current implementation already works as expected, I just wanted this to get off my chest.


peppy:
From the screenshot it looks like the argon catch fruit are more circular / less varied. Could that be the case?

EVAST9919:
> From the screenshot it looks like the argon catch fruit are more circular / less varied. Could that be the case?

Yes, osu-side settings should be adjusted, can report on performance after I do that.

EVAST9919:
With smilar settings
new

EVAST9919:
@peppy This is probably the best I can do in terms of performance/visuals balance

| master | pr |
| ----------- | ----------- |
| master | pr |

PR #5589, opened 2022-12-16
Add footnote support to Markdown container

bdach:
Step one for https://github.com/ppy/osu/issues/21666.

The framework-side implementation is purposefully ~~barebones~~ basic. I didn't want to spend too much time on it and instead focused on the game-side implementation. But if you think there were too many shortcuts taken here, then do let me know and I'll revisit.

There is a test scene to preview this in available:

### DIFF
--- a/osu.Framework/Graphics/Containers/Markdown/Footnotes/MarkdownFootnoteLink.cs
+++ b/osu.Framework/Graphics/Containers/Markdown/Footnotes/MarkdownFootnoteLink.cs
@@ -1,0 +1,40 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using Markdig.Extensions.Footnotes;
+using osu.Framework.Allocation;
+using osu.Framework.Extensions.LocalisationExtensions;
+
+namespace osu.Framework.Graphics.Containers.Markdown.Footnotes
+{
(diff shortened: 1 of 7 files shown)


osu.Framework/Graphics/Containers/Markdown/Footnotes/MarkdownFootnoteLink.cs
```diff
@@ -0,0 +1,40 @@
...
+        private void load()
+        {
+            var spriteText = parentTextComponent.CreateSpriteText();
+
+            AutoSizeAxes = Axes.Both;
+            InternalChild = spriteText.With(t =>
+            {
+                float baseSize = t.Font.Size;
+                t.Font = t.Font.With(size: baseSize * 0.58f);
+                t.Margin = new MarginPadding { Bottom = 0.33f * baseSize };
```
bdach, line 35: Constants were lifted from https://en.wikipedia.org/wiki/Subscript_and_superscript#Desktop_publishing.

bdach:
Game-side implementation can be previewed at https://github.com/bdach/osu/tree/markdown-footnote-support. It's baked a little better (includes preview of footnote content on hover and bidirectional navigation between footnote links and footnote groups at the bottom of the document).

I'm not PRing yet because I branched that off of https://github.com/ppy/osu/pull/21629, so until that's in the diff will just be unreadable.

peppy (approved):

PR #5590, opened 2022-12-16
Fix Veldrid blocking iOS from building

frenzibyte:
Veldrid relies on a certain library for providing Direct3D 11 API, which is powered by a binding generator that fails to get AOT'd on iOS.
```
0>Mono Ahead of Time compiler - compiling assembly /Users/salman/Projects/osu-framework/SampleGame.iOS/obj/iPhone/Debug/net6.0-ios/ios-arm64/linked/SharpGen.Runtime.dll
AOTID 7CA136D1-DC5C-CDE1-6269-4B4A034E7F8B
* Assertion at /Users/runner/work/1/s/src/mono/mono/metadata/custom-attrs.c:599, condition `out_obj' not met


0>Tool /usr/local/share/dotnet/packs/Microsoft.NETCore.App.Runtime.AOT.osx-x64.Cross.ios-arm64/6.0.11/Sdk/../tools/mono-aot-cross execution started with arguments: --aot=mtriple=arm64-ios,data-outfile=obj/iPhone/Debug/net6.0-ios/ios-arm64/nativelibraries/aot-output/arm64/System.Drawing.Common.aotdata,static,asmonly,direct-icalls,full,readonly-value=ObjCRuntime.Runtime.Arch=i4/0,soft-debug,dwarfdebug,no-direct-calls,outfile=obj/iPhone/Debug/net6.0-ios/ios-arm64/nativelibraries/aot-output/arm64/System.Drawing.Common.dll.s --aot= --debug -O=gsharedvt /Users/salman/Projects/osu-framework/SampleGame.iOS/obj/iPhone/Debug/net6.0-ios/ios-arm64/linked/System.Drawing.Common.dll
        
0>/usr/local/share/dotnet/packs/Microsoft.iOS.Sdk/16.0.527/targets/Xamarin.Shared.Sdk.targets(1003,3): Error  : Failed to AOT compile SharpGen.Runtime.dll, the AOT compiler exited with code 134
[...]

PR #5592, opened 2022-12-17
Work around DirectX libraries blocking iOS from building

frenzibyte:
Attempt number 2, this time with not too extreme changes (see https://github.com/ppy/osu-framework/pull/5590 for previous approach).

`Vortice.Direct3D11` and other DirectX libraries rely on `SharpGen.Runtime` for bindings generation, which makes it referenced during framework building stage, but fail on iOS during AOT compilation for unknown reasons.

I've went ahead and compiled empty assemblies for the faulty libraries and set up a target to copy the assemblies to the linking folder, while expecting it to fail due to the assemblies being... well, empty. But luckily enough, it passed AOT stage and finished building successfully.


peppy (approved):
Seems much more palatable.

smoogipoo (approved):
I can't really test this, but it looks fine

PR #5595, opened 2022-12-19
Implement Veldrid vertex buffers

frenzibyte:
Implementations mostly copied from `GLVertexBuffer` and siblings. Not too sure about `VeldridIndexData`, I've went with it just to have something working (in a similar manner to `GLLinearIndexData`), but it's completely up for refactor.


osu.Framework/Graphics/Rendering/Vertices/VertexMemberAttribute.cs
```diff
@@ -19,6 +19,7 @@ public class VertexMemberAttribute : Attribute
         /// The type of each component of this vertex attribute member.
         /// E.g. a <see cref="osuTK.Vector2"/> is represented by 2 **<see cref="VertexAttribPointerType.Float"/>** components.
         /// </summary>
+        // todo: this should be replaced by an enum defined in o!f.
```
frenzibyte, line 22: As per above, we shouldn't rely on using osuTK enums for o!f rendering code anymore, but seeing that changing the enum would be a breaking change, I've left a TODO for now.

osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
```diff
@@ -245,6 +266,43 @@ protected override void SetFrameBufferImplementation(IFrameBuffer? frameBuffer)
...
+
+        private readonly Dictionary<GraphicsPipelineDescription, Pipeline> pipelineCache = new Dictionary<GraphicsPipelineDescription, Pipeline>();
+
+        private Pipeline getPipelineInstance()
+        {
+            if (!pipelineCache.TryGetValue(pipeline, out var instance))
+            {
+                pipelineCache[pipeline] = instance = Factory.CreateGraphicsPipeline(ref pipeline);
+                stat_graphics_pipeline_created.Value++;
+            }
```
smoogipoo, line 301: This is kinda scary. So we need a new graphics pipeline for every set of shaders/textures/framebuffers/etc? The construction of these pipelines [does not seem trivial](https://github.com/mellinoe/veldrid/blob/a23dc39a1b9e103f19c2f86c6c8ddcdc15823e44/src/Veldrid/Vk/VkPipeline.cs#L33-L344).

smoogipoo (approved):

osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
frenzibyte: I'm not 100% sure about my statistics, but checking the "total pipelines created" one on o!f visual tests it interestingly doesn't get beyond 3 different pipeline states. So I felt it's sufficient now just for the time being.

PR #5596, opened 2022-12-19
Implement Veldrid shaders

frenzibyte:
- [x] Depends on #5595

Shaders in Veldrid (i.e. other backends) work slightly different w.r.t. uniforms. In OpenGL, you can retrieve the uniforms by their names and supply the correct values as such. However, in Veldrid you must define a uniform buffer object to access the uniform wanted. 

This is impossible to work with on o!f since we may use the same vertex shader with multiple fragment shaders, and the UBO structure has to be consistent across both parts of the shader. For that, I've made a structure wherein all uniforms referenced by both parts of the shader are parsed and combined into one UBO with the structure copied into both parts during compilation.

We can also add validation logic to ensure the vertex layout structure defined by the shader and the vertex layout structure defined by the `IVertex` type match 1:1, and warn the consumer on any noticed differences, but I think it's best to leave that in the final steps.


osu.Framework/Resources/Shaders/sh_yuv2rgb-veldrid.h
```diff
@@ -0,0 +1,23 @@
+#include "sh_TextureWrapping.h"
+
+layout(set = 1, binding = 0) uniform texture2D m_TextureY;
+layout(set = 1, binding = 1) uniform texture2D m_TextureU;
+layout(set = 1, binding = 2) uniform texture2D m_TextureV;
+layout(set = 1, binding = 3) uniform sampler m_Sampler;
+
+uniform mediump mat3 yuvCoeff;
```
smoogipoo, line 8: Is there a reason for this to not have an explicit binding like all the others? As far as I can tell, unbound uniforms like this are considered to be part of the global uniform list which this isn't.

Full generated code for reference: https://gist.github.com/smoogipoo/8d7add0755e7c5fa767315a3e8381ed2

osu.Framework/Resources/Shaders/sh_yuv2rgb-veldrid.h
```diff
@@ -0,0 +1,23 @@
+#include "sh_TextureWrapping.h"
+
+layout(set = 1, binding = 0) uniform texture2D m_TextureY;
+layout(set = 1, binding = 1) uniform texture2D m_TextureU;
+layout(set = 1, binding = 2) uniform texture2D m_TextureV;
+layout(set = 1, binding = 3) uniform sampler m_Sampler;
```
smoogipoo, line 6: Additionally, I'm concerned that `shader_uniform_regex` is going to miss these - they're still output to the file however they aren't added to the `Uniforms` list due to the layout specification and the resulting `shader.GetUniform<int>("m_SamplerY")` call should fail if I'm reading everything correctly.

smoogipoo (changes requested):
I'm very much not a fan of the juggling of layouts...

But it might just be a matter of timing - could we instead take a step back first and expose includes for global uniforms and vertex layouts? Can you think of any issues we'd have in this path?

For example...

`sh_GlobalUniforms.h`:
```glsl
layout(set = 0, binding = 0) uniform g_GlobalUniforms
{
    bool g_GammaCorrection;
    ...
};
```

`sh_Texture2D_VertexAttributes.h`:
```glsl

#ifdef OSU_VERTEX_SHADER
layout(location = 0) in m_VertexIn
{
    highp vec2 m_Position;
    ...
};
#endif

#ifdef OSU_VERTEX_SHADER
layout(location = 0) out v_VertexOut
#else
layout(location = 0) in v_VertexOut
#endif
{
    highp vec2 v_MaskingPosition;
    ...
};
```

Where we inject the macros ourselves similar to how `#version` is injected via, i.e.:
```
#version 450
#define OSU_VERTEX_SHADER
```

Furthermore, as I alluded to in the past, we really shouldn't have included files defining anything. It might become a little bit less of a mess if the files were split as above, but it might also be a good idea to pass variables into these included files instead (e.g. `sh_TextureWrapping.h`).

smoogipoo:
Furthermore, if the above split works, then it might also be possible to use `#if __VERSION__ == 110` or something like that to combine the files [(or `120`, `130`, `140`, whatever's our minimal target)](https://en.wikipedia.org/wiki/OpenGL_Shading_Language). One branch with the `layout()` specifiers and the other without, for both renderers to use.

osu.Framework/Resources/Shaders/sh_yuv2rgb-veldrid.h
smoogipoo: As as side note, I'm fairly certain this file is completely incorrect.

- `texture2D` is not a type as far as I know. If it is, can you provide documentation?
- It changed naming from `m_Sampler{X}` to `m_Texture{X}`, breaking the DrawNode.
- It calls `texture(sampler2D(tex, samp), ...)` and I'm fairly sure `sampler2D` isn't a function? If it is, can you provide documentation?

frenzibyte: Veldrid shaders use a Vulkan-flavoured GLSL format, which is different than the usual GLSL as per the following:
 - `sampler2D` type for texture objects changed into `texture2D`
 - `texture2D(sampler, coord)` function for sampling textures changed into `texture(sampler2D(texture, sampler), coord)`

I also plan to move the uniform population logic in `VideoSpriteDrawNode.Draw` to a better place specific to GL since it won't work for Veldrid anyway.

osu.Framework/Resources/Shaders/sh_yuv2rgb-veldrid.h
frenzibyte: The generated buffer is more of a list of uniforms that are declared by the shader and the included headers, not specifically meant for "global uniforms".

frenzibyte:
I've spent some time figuring things out and for the most part, things weren't going nicely (apart from the vertex layout headers proposal).

- Separating global uniforms to a separate buffer required extensive refactor of `GlobalPropertyManager` (https://github.com/ppy/osu-framework/commit/3b738f2374330644253ecc410e8f08d04b27f6a3), but I've managed to get it working as far as I can see.
- Replacing the pre-processing logic for shader uniforms with explicitly defined UBOs, i.e.:
  ```glsl
  // on vertex shader
  layout(set = 1, binding = 0) uniform m_VertexUniforms
  {
      ...
  }

  // on fragment shader
  layout(set = 1, binding = 1) uniform m_FragmentUniforms
  {
      ...
  }
  ```

  Was simple to do in code, but ended up triggering a bug with SPIR-V when targeting Metal (short of the output MSL not respecting the binding number given to the UBO, see https://github.com/KhronosGroup/SPIRV-Cross/issues/1248).

- Attempting to merge OpenGL and Veldrid shaders together cannot be done without more regex, simply because Veldrid uses [Vulkan-flavoured GLSL 450](https://github.com/KhronosGroup/GLSL/blob/master/extensions/khr/GL_KHR_vulkan_glsl.txt) which is not compatible with the master shaders on multiple levels (see https://github.com/ppy/osu-framework/pull/5596#discussion_r1052657412).

As [mentioned](https://discord.com/channels/188630481301012481/589331078574112768/1055352839536386058) today, a better overall solution that avoids preprocessing would be for each `DrawNode` to provide resource sets defining the shader UBOs and texture resources, rather than providing arbitrary layouts by the renderer itself. I'll look into that in the coming weeks.

PR #5597, opened 2022-12-19
Implement Veldrid textures

frenzibyte:
- [ ] Depends on #5596 

The initialisation logic of the texture works completely different from `GLTexture`, since this one includes the concept of a "sampler" alongside the texture, which gets invalidated whenever the number of mipmap levels increases.

My implementation follows a simplified binding model in which each resource set that's bound to the GPU on draw consists of a single texture (or multiple textures in the case of video) and a sampler. I didn't think too much about it, and I can imagine us vastly improving the model to our benefit (if the current model isn't quite good).

In addition, basic drawing has now enabled and most of the o!f components can now be rendered with Veldrid (minus framebuffers, up next).


frenzibyte:
Closing in favour of https://github.com/ppy/osu-framework/pull/5647

PR #5598, opened 2022-12-19
Implement Veldrid framebuffers

frenzibyte:
- [ ] Depends on #5597 

Not any special compared to `GLFramebuffer`. But that being said, for some odd reason nothing is drawn for me while running on Metal (not sure about other backends), I haven't investigated this yet but I'm opening this PR any way rather than wasting more time on it with a single pair of eyes.

- [ ] Fix framebuffers rendering nothing


frenzibyte:
Closing in favour of https://github.com/ppy/osu-framework/pull/5647

PR #5600, opened 2022-12-19
Avoid performing GL occlusion queries on Veldrid

frenzibyte:
Was an annoying one to me, as the moment I accidentally switch to this test scene everything crashes because of the GL occlusion queries being performed in the draw node, and there's no way to escape this crash without commenting the entire test scene.


smoogipoo:
Yikes. I missed that this test scene does raw GL calls. But it is very internal so ./shrug.

smoogipoo (approved):

PR #5604, opened 2022-12-20
`CircularProgress` visual improvements

EVAST9919:
* Increased smoothening factor (should look a bit better on high-dpi screens). Not increasing further since it's already on the edge of feeling blurry on my 1080p screen.
* Fixed antialiasing for non 1:1 drawable aspect ratios

### DIFF
--- a/osu.Framework/Graphics/UserInterface/CircularProgress.cs
+++ b/osu.Framework/Graphics/UserInterface/CircularProgress.cs
@@ -110,7 +110,7 @@ namespace osu.Framework.Graphics.UserInterface
                 roundedCaps = Source.roundedCaps;
 
                 // Looks too sharp with 1px, let's give it a bit more
-                texelSize = 2f / ScreenSpaceDrawQuad.Size.X;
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/UserInterface/CircularProgress.cs
```diff
@@ -109,8 +109,8 @@ public override void ApplyState()
                 progress = Math.Abs((float)Source.current.Value);
                 roundedCaps = Source.roundedCaps;
 
-                // smoothstep looks too sharp with 1px, let's give it a bit more
-                texelSize = 1.5f / ScreenSpaceDrawQuad.Size.X;
+                // Looks too sharp with 1px, let's give it a bit more
```
frenzibyte, line 112: This comment makes no sense now, why was "smoothstep" removed from the comment?

EVAST9919: No particular reason, will revert that

EVAST9919:
Superseded by https://github.com/ppy/osu-framework/pull/5612

PR #5605, opened 2022-12-21
Use new ArgumentNullException.ThrowIfNull throw-helper API

turbedi:
This API was introduced in .NET 6.
In some cases, the custom exception message has been removed because the throw helper doesn't allow it. I can revert these changes if we want to keep the custom message.

### DIFF
--- a/osu.Framework/Allocation/InvokeOnDisposal.cs
+++ b/osu.Framework/Allocation/InvokeOnDisposal.cs
@@ -23,7 +23,9 @@ namespace osu.Framework.Allocation
         /// <param name="action">The action to invoke during disposal.</param>
         public InvokeOnDisposal(Action action)
         {
-            this.action = action ?? throw new ArgumentNullException(nameof(action));
+            ArgumentNullException.ThrowIfNull(action);
+
+            this.action = action;
         }
 
         #region IDisposable Support
--- a/osu.Framework/Allocation/InvokeOnDisposal.cs
+++ b/osu.Framework/Allocation/InvokeOnDisposal.cs
@@ -23,9 +23,7 @@ namespace osu.Framework.Allocation
         /// <param name="action">The action to invoke during disposal.</param>
         public InvokeOnDisposal(Action action)
         {
-            ArgumentNullException.ThrowIfNull(action);
-
-            this.action = action;
+            this.action = action ?? throw new ArgumentNullException(nameof(action));
         }
 
         #region IDisposable Support
--- a/osu.Framework/Bindables/NonNullableBindable.cs
+++ b/osu.Framework/Bindables/NonNullableBindable.cs
@@ -10,8 +10,7 @@ namespace osu.Framework.Bindables
     {
         public NonNullableBindable(T defaultValue)
         {
-            if (defaultValue == null)
-                throw new ArgumentNullException(nameof(defaultValue));
+            ArgumentNullException.ThrowIfNull(defaultValue);
 
             Value = Default = defaultValue;
         }
@@ -25,8 +24,7 @@ namespace osu.Framework.Bindables
             get => base.Value;
             set
             {
-                if (value == null)
-                    throw new ArgumentNullException(nameof(value), $"Cannot set {nameof(Value)} of a {nameof(NonNullableBindable<T>)} to null.");
+                ArgumentNullException.ThrowIfNull(value);
 
                 base.Value = value;
             }
--- a/osu.Framework/Bindables/NonNullableBindable.cs
+++ b/osu.Framework/Bindables/NonNullableBindable.cs
@@ -24,7 +24,8 @@ namespace osu.Framework.Bindables
             get => base.Value;
             set
             {
-                ArgumentNullException.ThrowIfNull(value);
+                if (value == null)
+                    throw new ArgumentNullException(nameof(value), $"Cannot set {nameof(Value)} of a {nameof(NonNullableBindable<T>)} to null.");
 
                 base.Value = value;
             }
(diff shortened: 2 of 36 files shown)


osu.Framework/Allocation/InvokeOnDisposal.cs
```diff
@@ -23,7 +23,9 @@ public class InvokeOnDisposal : IDisposable
         /// <param name="action">The action to invoke during disposal.</param>
         public InvokeOnDisposal(Action action)
         {
-            this.action = action ?? throw new ArgumentNullException(nameof(action));
+            ArgumentNullException.ThrowIfNull(action);
+
+            this.action = action;
```
bdach, line 28: Very lukewarm feelings on converting these onelines. Using these helpers instead of the previous inline expressions is ugly, and at the point where exceptions are being thrown I don't think performance matters enough to live with the ugliness.

osu.Framework/Bindables/NonNullableBindable.cs
```diff
@@ -25,8 +24,7 @@ public override T Value
             get => base.Value;
             set
             {
-                if (value == null)
-                    throw new ArgumentNullException(nameof(value), $"Cannot set {nameof(Value)} of a {nameof(NonNullableBindable<T>)} to null.");
+                ArgumentNullException.ThrowIfNull(value);
```
bdach, line 27: Would revert. Old exception message was useful. You could imagine trying to interact with an instance of this through `Bindable<T>` and being surprised that you can't assign a `null` `Value`. The message explained that it's due to the bindable's specific type (which is arguably... *imperfect* design, but it was better than a generic nullref).

osu.Framework/Graphics/Transforms/TransformSequence.cs
```diff
@@ -47,8 +47,7 @@ public class TransformSequence<T> : ITransformSequence where T : class, ITransfo
         /// <param name="origin">The <typeparamref name="T"/> to attach the new <see cref="TransformSequence{T}"/> to.</param>
         public TransformSequence(T origin)
         {
-            if (origin == null)
-                throw new ArgumentNullException(nameof(origin), $"May not create a {nameof(TransformSequence<T>)} with a null {nameof(origin)}.");
+            ArgumentNullException.ThrowIfNull(origin);
```
bdach, line 50: Would revert. Transforms tend to be wedged in quite deep and so a random nullref without a precise message will lead to too much digging around and gnashing of teeth.

turbedi:
Will revert the exceptions with custom messages and cases where the code is more cluttered than before later today.

turbedi:
I have reverted all cases where there was a custom exception message, or where the new pattern made the code uglier.

bdach (approved):
Seems okay. I'll probably wait and see if I can get a second approval going to make sure everyone is on board with this though.

osu.Framework/Allocation/InvokeOnDisposal.cs
smoogipoo: The difference isn't big, but for performance critical places - I'd argue this (and `ValueInvokeOnDisposal`) are - I am fully in support of using these helper methods if only to reduce ASM code size. It's a pretty common optimisation in the .NET world, which is why these methods were added in the first place: https://github.com/dotnet/runtime/issues/48573

See this gist: https://gist.github.com/smoogipoo/612287ba15e37c0ef1950520fe8b2b96

smoogipoo (approved):
We can get this in in the interim since the changes are simple, however I'm still of the opinion that it should be used in more places, as per https://github.com/ppy/osu-framework/pull/5605#discussion_r1055153165. Potentially even a full revert of https://github.com/ppy/osu-framework/pull/5605/commits/17b7b14d90c09064409ecada14ef3cbdaf0d99f1

bdach:
> Potentially even a full revert of [17b7b14](https://github.com/ppy/osu-framework/commit/17b7b14d90c09064409ecada14ef3cbdaf0d99f1)

Even the instances where there are custom exception messages? I can yield on the performance critical but ugly instances for sure, but the ones with custom exception messages are to me a major downgrade in developer usability.

smoogipoo:
I think they're pretty self explanatory. Between the following messages:
```
Unhandled exception. System.ArgumentNullException: May not create a TransformSequence with a null origin. (Parameter 'origin')
   at osu.Framework.Graphics.Transforms.TransformSequence`1..ctor(T origin) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Transforms/TransformSequence.cs:line 51
   at osu.Desktop.Program.Main(String[] args) in /home/smgi/Repos/osu/osu.Desktop/Program.cs:line 35
```
```
Unhandled exception. System.ArgumentNullException: Value cannot be null. (Parameter 'origin')
   at System.ArgumentNullException.Throw(String paramName)
   at System.ArgumentNullException.ThrowIfNull(Object argument, String paramName)
   at osu.Framework.Graphics.Transforms.TransformSequence`1..ctor(T origin) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Transforms/TransformSequence.cs:line 50
   at osu.Desktop.Program.Main(String[] args) in /home/smgi/Repos/osu/osu.Desktop/Program.cs:line 35
```
Given the appropriate stacktrace, I don't think any information is being lost. The second one may even be more readable since it's a consistent message across everything (e.g. for Sentry views).

bdach:
I'd still argue that the `NonNullableBindable` case should stay but I don't care enough to argue this further.

smoogipoo:
IOS not happy with this one, for some reason :(

frenzibyte:
That one's a known failure, still no clue why it's happening but re-running fixes it.

bdach:
That just looks like the iOS workflow broke for whatever reason. There are similar failures [on other pulls](https://github.com/ppy/osu-framework/actions/runs/3745104212/jobs/6359171576).

PR #5609, opened 2022-12-24
Move `CircularProgress` alpha hack to the draw node

EVAST9919:
Depends on:
 - [ ] https://github.com/ppy/osu-framework/pull/5604 (well, not really, that pr is fairly small and probably can be skipped since this one contains these changes anyway)

**The hack in question**
To make progress appear smooth if it's thickness < texelSize we are making it more transparent while leaving thickness the same. Previously calculated in the shader and one of potential breaking points on some platforms due to `mediump / highp`. (latter is my guess, not proven atm). Also was calculated a bit wrong, so the minimum thickness was thicker than it could be.
To closely inspect the result just increase the `texelSize` value to something big (like 40 instead of 2) and play around with the fill slider. You will see at which point the thickness will no longer be thinner and alpha multiplier starts to have effect.

Also I upped the precision of `pathRadius` to `highp` just in case.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneCircularProgress.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneCircularProgress.cs
@@ -132,7 +132,7 @@ namespace osu.Framework.Tests.Visual.UserInterface
 
             AddToggleStep("Toggle masking", m => maskingContainer.Masking = m);
             AddToggleStep("Toggle rounded caps", r => clock.RoundedCaps = r);
-            AddToggleStep("Toggle aspect ratio", r => clock.Size = r ? new Vector2(600, 400) : new Vector2(400));
+            AddToggleStep("Toggle aspect ratio", r => clock.Size = r ? new Vector2(600, 200) : new Vector2(400));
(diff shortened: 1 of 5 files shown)


frenzibyte:
Note that multiplication/division between `highp` and `mediump` will always result in a `highp` operation (according to the [docs](https://www.khronos.org/registry/OpenGL/specs/es/3.2/GLSL_ES_Specification_3.20.html#precision-and-precision-qualifiers)), it's addition/subtraction where the result is stored in `mediump` instead.

frenzibyte:
Since the issue turned out to be something completely different (see https://github.com/ppy/osu-framework/pull/5610), I'm not sure if this PR has any point. I also find it better to do the sub-pixel hack in the shader itself rather than modifying (i.e. copying) `DrawColourInfo` per-frame, but interested to hear others' opinions.

EVAST9919:
Fair enough. Although I'll make a new pr with some adjustments since in master as I mentioned minimum thickness is calculated a bit wrong and thicker than in could be. Or I can rework this one.

EVAST9919:
Closing as not solving original issue (solved by https://github.com/ppy/osu-framework/pull/5610)

Issue #5608, opened 2022-12-23
Add a way to create "raw GPU buffers"


Flutterish:
Discussion started on [discord](https://discord.com/channels/188630481301012481/589331078574112768/1055912719028584618)

This is a requirement for me to upgrade o!f-xr to work with arbitrary graphics surfaces (along with some minimal changes to shader managers, but I'll leave that for later). I would like to implement this myself if possible, but I would like to first know your opinion on the details.

The basic idea is that currently (from the public API) we can only create vertex batches, which are essentially a list of raw vertex buffers and a raw element buffer. This is somewhat limiting in terms of how we can render things. o!f-xr currently bypasses `IRenderer` for this step and creates raw buffers directly with the GL api.

I would like for the public API to add 3 new methods: a way to create a minimally-encapsulated vertex buffer, element buffer and vertex array respectively. This is pretty much a requirement for 3D rendering as we do not wish to re-upload or re-link big meshes every frame using batches. It might also prove useful for 2D rendering given complex shapes (paths, cached batches).

The ideal implementation of this for me would be something like this and nothing more (sample implementation with GL):
```cs
{I, GL}Renderer {
    public CreateRaw{Vertex, Element}Buffer(), CreateRawVertexArray()
}
GL{Vertex, Element}Buffer<T> {
    internal Handle = GL.GenBuffer();
    public Bind() => GL.BindBuffer(...);
    public Unbind() => GL.BindBuffer(...);
    // no book keeping - the cpu side will not store the uploaded data or its size. This can be done with a wrapper class
    public Buffer(ReadOnlySpan<T> data[, offset, usageHint]) =>GL.BufferData(...);
    public Dispose() => GL.DeleteBuffer(...);
    # VertexBuffer only
    public Draw( topology[, offset, count] ) => GL.DrawArrays(...);
    # ElementBuffer only
    // we do not store the topology in the element buffer. This can be done with a wrapper class
    public Draw( topology[, offset, count] ) =>GL.DrawElements(...);
}
GLVertexArray {
    internal Handle = GL.GenVertexArray();
    public Bind() => GL.BindVertexArray(...);
    public Unbind() => GL.BindVertexArray(...);
    public Dispose() => GL.DeleteVertexArray(...);
    public LinkElementBuffer( IRawElementBuffer ebo ) => ...; // just binds both since that links an ebo to a vao
    // binds the vao and vbo, then does stuff with VertexAttribPointer and EnableVertexAttribArray
[...]

PR #5614, opened 2022-12-28
Expose renderer buffers publically

Flutterish:
Related discussion (and motivation for this PR): https://github.com/ppy/osu-framework/issues/5608

This PR adds a way for users to create:
* A raw vertex buffer
* A raw index buffer
* A "state array". This is an open to backend specific optimizations array which holds render state. This should allow to reduce draw calls in some use cases - the one I implemented for the OpenGL renderer can use a VAO to cache vertex layout data and the bound index buffer, together or separately if needed. State arrays are constructed with flags which tell them which state they should cache.

This PR is not quite in a ready state yet, but I would still like to know your opinion on the direction of it.

There are still some issues this has:
* Batches grind with this change. The current implementation does not yet ensure that an active batch is flushed before the user can use raw buffers or the state array. Some solutions for this are:
    * Expose a "flush active batch" method and require the user to call it before using these.
    * Flush the batch automatically, though if batches themselves decide to use the buffers they might need to somehow signal to disable this while in use.
    * A `using` region where raw buffers can be used. The batches would be flushed on begin of such region.
* `DrawOpaqueInterior` works incorrectly if `BackbufferDrawDepth` of `DepthWrappingVertex` is cached in a raw vertex buffer. A workaround for this I found (and I have no clue why it works) is setting `CanDrawOpaqueInterior` to true.
    * A way to fix this could be to add a uniform backbuffer depth offset to the shaders and keep the vertex data as 0 for such caches.
    * You can see this in the test I added if you remove this hack and move your cursor outside the window on any test setting other than "batch".
* To draw anything with the provided shaders we *need* to use `DepthWrappingVertex` because that's what the shaders require
    * Not fixable by using a custom shader as this attribute is auto-generated
    * Also can be worked around by skipping the `BackbufferDrawDepth` when setting the vertex layout and using the uniform trick
* Not implemented on veldrid. Oustide of my expertise.


frenzibyte:
I'm not entirely fond by the direction of this PR. To start off, can we base the buffer implementations used by vertex batches off of these new "raw" buffer classes? At which point, the term "raw" can be removed (interfaces would be named `IRendererBuffer` instead).

That being said, having an "index buffer" class supporting a "draw" method reads really weird. I mistook it for an "indexed vertex buffer" but it's not the case because it doesn't store any vertices, it only draws what's currently in the pipeline with the indices it has. Is there a reason for defining such method in the first place? Can't we just define an "indexed vertex buffer" class instead, or does that not work well with the "renderer state array"?

Flutterish:
> can we base the buffer implementations used by vertex batches off of these new "raw" buffer classes

I would love for that to happen, though please look at the issue no. 1. I still need to figure what the best way to integrate bash flushing with users using this api is. Though, perhaps a bit too much for one PR, this one already introduces quite a bit.

> At which point, the term "raw" can be removed (interfaces would be named IRendererBuffer instead).

Absolutely, that was my intention from the very beginning. The only reason I opted for the term "raw" is because there was already an interface named `IVertexBuffer` and I didn't want to introduce chaos by renaming it. I think it might be reasonable to do so when integrating these buffers into batches.

> That being said, having an "index buffer" class supporting a "draw" method reads really weird. I mistook it for an "indexed vertex buffer" but it's not the case because it doesn't store any vertices, it only draws what's currently in the pipeline with the indices it has. Is there a reason for defining such method in the first place? Can't we just define an "indexed vertex buffer" class instead, or does that not work well with the "renderer state array"?

An "indexed vertex buffer" is quite limited in comparison, as it can index only one vertex buffer, while an actual index buffer can be reused across many (or used on many at once), as well as many index buffers (or slices of one) can index one vertex buffer set. While I'm sure you can get around this by always using one vertex buffer and one index buffer, this is not ideal to work with imo. Also, an indexed vertex buffer is quite easy to make as a wrapper class, if one is desired.

That said though, the "draw" method is quite weird to even be on a vertex buffer now that I think about it - that is, many vertex buffers can have their layout set at once which makes it weird that one specific would be requested to draw them all. Perhaps both draw methods should be moved to `IRenderer`?

frenzibyte:
> I would love for that to happen, though please look at the issue no. 1. I still need to figure what the best way to integrate batch flushing with users using this api is. Though, perhaps a bit too much for one PR, this one already introduces quite a bit.

Exposing the `FlushCurrentBatch` method is definitely a no-no, we never want consumers to freely use such operation wherever they want. I'm thinking of something like the following:
```csharp
public class Renderer
{
    public void BindVertexBuffer(IVertexBuffer buffer)
    {
        if (buffer is not IBatchedVertexBuffer batchBuffer || batchBuffer.Batch == currentActiveBatch)
            FlushCurrentBatch(FlushBatchSource.BindVertexBuffer);

        BindBufferImplementation(buffer);
    }
}
```

> That said though, the "draw" method is quite weird to even be on a vertex buffer now that I think about it - that is, many vertex buffers can have their layout set at once which makes it weird that one specific would be requested to draw them all. Perhaps both draw methods should be moved to IRenderer?

Sounds like it might work better (defining `Draw` and `DrawIndexed` methods), but will definitely need more feedback before pursuing through such direction (@smoogipoo interested in your opinion).

PR #5617, opened 2022-12-31
Fix `DebugUtils.IsDebugBuild` not working on Android

### LINKED ISSUE #5616: `DebugUtils.IsDebugBuild` doesn't work on Android (always returns `false`)
https://discord.com/channels/188630481301012481/188630652340404224/1057592023382765631

Happens only when using a `Release` framework build (eg. from nuget).

This is because `Assembly.GetEntryAssembly()` always returns `null` on Android (and `Assembly.GetCallingAssembly()` is not useful as that will always return the `osu.Framework` assembly in this case).

https://github.com/ppy/osu-framework/blob/5012f8572eb9d28f333908a5c2deef0404abddd3/osu.Framework/Development/DebugUtils.cs#L49-L73

With the below diff, the box should be red in debug mode, but instead it's always green.

```diff
diff --git a/SampleGame.Android/SampleGame.Android.csproj b/SampleGame.Android/SampleGame.Android.csproj
index 2562ffd9d..106f8f5ff 100644
--- a/SampleGame.Android/SampleGame.Android.csproj
+++ b/SampleGame.Android/SampleGame.Android.csproj
@@ -7,7 +7,7 @@
     <AssemblyName>SampleGame.Android</AssemblyName>
   </PropertyGroup>
   <ItemGroup>
-    <ProjectReference Include="..\osu.Framework.Android\osu.Framework.Android.csproj" />
+    <PackageReference Include="ppy.osu.Framework.Android" Version="2022.1226.0 "/>
     <ProjectReference Include="..\SampleGame\SampleGame.csproj" />
   </ItemGroup>
 </Project>
diff --git a/SampleGame/SampleGame.csproj b/SampleGame/SampleGame.csproj
index 423a124a7..d75f6488d 100644
--- a/SampleGame/SampleGame.csproj
+++ b/SampleGame/SampleGame.csproj
@@ -3,6 +3,6 @@
     <TargetFramework>net6.0</TargetFramework>
   </PropertyGroup>
   <ItemGroup Label="Project References">
-    <ProjectReference Include="..\osu.Framework\osu.Framework.csproj" />
+    <PackageReference Include="ppy.osu.Framework" Version="2022.1226.0" />
   </ItemGroup>
 </Project>
diff --git a/SampleGame/SampleGameGame.cs b/SampleGame/SampleGameGame.cs
index ae72ffa24..abbc18847 100644
--- a/SampleGame/SampleGameGame.cs
+++ b/SampleGame/SampleGameGame.cs
@@ -7,6 +7,7 @@
 using osuTK.Graphics;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Allocation;
+using osu.Framework.Development;
 
 namespace SampleGame
 {
@@ -22,7 +23,7 @@ private void load()
                 Anchor = Anchor.Centre,
                 Origin = Anchor.Centre,
                 Size = new Vector2(150, 150),
-                Colour = Color4.Tomato
+                Colour = DebugUtils.IsDebugBuild ? Color4.Tomato : Color4.Green
             });
         }
```

Susko3:
- Closes #5616 

As explained in the issue, the current way doesn't work on Android, as it basically tests whether `osu.Framework` is built in `Debug`.
So instead store and check the assembly of the inheritor of `AndroidGameActivity` as the consumer entry-point assembly.

`OnCreate` is the entry point into an Android app, so there is no fear that something will call `IsDebugBuild` before `AndroidEntryAssembly` is set.

The following diff can be used to verify this change:
(The `osu-framework.sln` changes force `osu.Framework` and `osu.Framework.Android` to build in `Release` even when `Debug` is selected in the dropdown.)
```diff
diff --git a/SampleGame/SampleGameGame.cs b/SampleGame/SampleGameGame.cs
index ae72ffa24..abbc18847 100644
--- a/SampleGame/SampleGameGame.cs
+++ b/SampleGame/SampleGameGame.cs
@@ -7,6 +7,7 @@
 using osuTK.Graphics;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Allocation;
+using osu.Framework.Development;
 
 namespace SampleGame
 {
@@ -22,7 +23,7 @@ private void load()
                 Anchor = Anchor.Centre,
                 Origin = Anchor.Centre,
                 Size = new Vector2(150, 150),
-                Colour = Color4.Tomato
+                Colour = DebugUtils.IsDebugBuild ? Color4.Tomato : Color4.Green
             });
         }
 
diff --git a/osu-framework.sln b/osu-framework.sln
index 84ab8d9c6..db27cf8db 100644
--- a/osu-framework.sln
+++ b/osu-framework.sln
@@ -81,8 +81,8 @@ Global
 		Release|iPhoneSimulator = Release|iPhoneSimulator
 	EndGlobalSection
 	GlobalSection(ProjectConfigurationPlatforms) = postSolution
-		{C76BF5B3-985E-4D39-95FE-97C9C879B83A}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{C76BF5B3-985E-4D39-95FE-97C9C879B83A}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{C76BF5B3-985E-4D39-95FE-97C9C879B83A}.Debug|Any CPU.ActiveCfg = Release|Any CPU
+		{C76BF5B3-985E-4D39-95FE-97C9C879B83A}.Debug|Any CPU.Build.0 = Release|Any CPU
 		{C76BF5B3-985E-4D39-95FE-97C9C879B83A}.Debug|iPhone.ActiveCfg = Debug|Any CPU
 		{C76BF5B3-985E-4D39-95FE-97C9C879B83A}.Debug|iPhone.Build.0 = Debug|Any CPU
 		{C76BF5B3-985E-4D39-95FE-97C9C879B83A}.Debug|iPhoneSimulator.ActiveCfg = Debug|Any CPU
@@ -173,8 +173,8 @@ Global
 		{D972753E-45FC-4B82-B017-34BDE485F1BB}.Release|iPhone.Build.0 = Release|iPhone
 		{D972753E-45FC-4B82-B017-34BDE485F1BB}.Release|iPhoneSimulator.ActiveCfg = Release|iPhoneSimulator
 		{D972753E-45FC-4B82-B017-34BDE485F1BB}.Release|iPhoneSimulator.Build.0 = Release|iPhoneSimulator
-		{4D112E30-462B-4264-B44D-53B61ABB185E}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{4D112E30-462B-4264-B44D-53B61ABB185E}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{4D112E30-462B-4264-B44D-53B61ABB185E}.Debug|Any CPU.ActiveCfg = Release|Any CPU
+		{4D112E30-462B-4264-B44D-53B61ABB185E}.Debug|Any CPU.Build.0 = Release|Any CPU
 		{4D112E30-462B-4264-B44D-53B61ABB185E}.Debug|iPhone.ActiveCfg = Debug|Any CPU
 		{4D112E30-462B-4264-B44D-53B61ABB185E}.Debug|iPhone.Build.0 = Debug|Any CPU
 		{4D112E30-462B-4264-B44D-53B61ABB185E}.Debug|iPhoneSimulator.ActiveCfg = Debug|Any CPU
```

# Breaking changes

## `DebugUtils.GetEntryAssembly()` has been moved to `RuntimeInfo.EntryAssembly`

The property and function are identical in their behaviour.


osu.Framework/Development/DebugUtils.cs
```diff
@@ -44,10 +42,15 @@ public static class DebugUtils
             }
         );
 
+        /// <summary>
+        /// Alternative to <see cref="GetEntryAssembly"/> on Android, as <see cref="Assembly.GetEntryAssembly"/> returns <c>null</c> there.
+        /// </summary>
+        internal static Assembly? AndroidEntryAssembly { private get; set; }
```
bdach, line 48: Can we not have android-specific stuff in the main framework project? Even if it's only for `DebugUtils` stuff?

I'd rather have an internal static `EntryAssembly` field here, that gets static-initialised to what `GetEntryAssembly()` currently returns, and just overwrite that on android.

frenzibyte: Probably want to ban `GetEntryAssembly()` as well in favour of the new property.

Susko3:
I've added tests to check behaviour. Checked to work as expected on Android and Windows.
For Windows, I checked with regular/non-headless "double clicking `osu.Framework.Tests.exe`", running unit tests in Rider, and `dotnet test` (the `testhost, Version=15.0.0.0` comment stands true).

Needs double-check on iOS/NativeAOT, but I would expect it to work fine, as iOS uses a normal `Main` function like desktop.

```cs
PS> dotnet test --filter TestSceneDebugUtils
...
[performance] 2023-07-23 19:38:34 [verbose]: TextureAtlas initialised (1024x1024)
[runtime] 2023-07-23 19:38:34 [verbose]: ?? Class: TestSceneDebugUtils
[runtime] 2023-07-23 19:38:34 [verbose]: ?? Test:  TestConstructor
[runtime] 2023-07-23 19:38:34 [verbose]: ? TestSceneDebugUtils completed

[runtime] 2023-07-23 19:38:34 [verbose]: ?? Class: TestSceneDebugUtils
[runtime] 2023-07-23 19:38:34 [verbose]: ?? Test:  LogStatics
[runtime] 2023-07-23 19:38:34 [verbose]: ?? Step #1 log DebugUtils statics
[runtime] 2023-07-23 19:38:34 [verbose]: DebugUtils.IsNUnitRunning: True
[runtime] 2023-07-23 19:38:34 [verbose]: DebugUtils.IsDebugBuild: True
[runtime] 2023-07-23 19:38:34 [verbose]: DebugUtils.GetEntryAssembly(): osu.Framework.Tests, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
[runtime] 2023-07-23 19:38:34 [verbose]: Assembly.GetEntryAssembly(): testhost, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
[runtime] 2023-07-23 19:38:34 [verbose]: ? TestSceneDebugUtils completed

[runtime] 2023-07-23 19:38:34 [verbose]: ?? Class: TestSceneDebugUtils
[runtime] 2023-07-23 19:38:34 [verbose]: ?? Test:  TestEntryAssembly
[runtime] 2023-07-23 19:38:34 [verbose]: ?? Step #1 check GetEntryAssembly()
[runtime] 2023-07-23 19:38:34 [verbose]: ? TestSceneDebugUtils completed

[runtime] 2023-07-23 19:38:34 [verbose]: ?? Class: TestSceneDebugUtils
[runtime] 2023-07-23 19:38:34 [verbose]: ?? Test:  TestIsDebugBuild
[runtime] 2023-07-23 19:38:34 [verbose]: ?? Step #1 check IsDebugBuild
[runtime] 2023-07-23 19:38:34 [verbose]: ? TestSceneDebugUtils completed

[runtime] 2023-07-23 19:38:34 [verbose]: ?? Class: TestSceneDebugUtils
[runtime] 2023-07-23 19:38:34 [verbose]: ?? Test:  TestIsNUnitRunning
[runtime] 2023-07-23 19:38:34 [verbose]: ?? Step #1 check IsNUnitRunning
[runtime] 2023-07-23 19:38:34 [verbose]: ? TestSceneDebugUtils completed
```

CodeAnalysis/BannedSymbols.txt
```diff
@@ -12,3 +12,4 @@ M:System.Char.ToLower(System.Char);char.ToLower() changes behaviour depending on
 M:System.Char.ToUpper(System.Char);char.ToUpper() changes behaviour depending on CultureInfo.CurrentCulture. Use char.ToUpperInvariant() instead. If wanting culture-sensitive behaviour, explicitly provide CultureInfo.CurrentCulture.
 M:System.String.ToLower();string.ToLower() changes behaviour depending on CultureInfo.CurrentCulture. Use string.ToLowerInvariant() instead. If wanting culture-sensitive behaviour, explicitly provide CultureInfo.CurrentCulture or use LocalisableString.
 M:System.String.ToUpper();string.ToUpper() changes behaviour depending on CultureInfo.CurrentCulture. Use string.ToUpperInvariant() instead. If wanting culture-sensitive behaviour, explicitly provide CultureInfo.CurrentCulture or use LocalisableString.
+M:System.Reflection.Assembly.GetEntryAssembly();Use osu.Framework.Development.DebugUtils.GetEntryAssembly() instead
```
peppy, line 15: The weird part for me here is that this is obsoleting a `System` method for a `DebugUtils` method. Maybe this shouldn't be in `DebugUtils` if it's required to be used everywhere?

From another perspective, if the dotnet function if broken then it feels like it should be reported upstream rather than working around like this.

Susko3: I don't think it's possible to fix this upstream. Android is quite flexible with activities and it allows them to be defined in any assembly and all of them can be the main activity. This means it's entirely possible for a single Android app to have multiple entry assemblies.

osu!framework has a special case, as it only expects to have a single activity in the game app.

Do you have a better idea on where to put it? Keep in mind it uses `private nunit_test_assembly` from `DebugUtils`.

peppy: Is it safe to say that it should only be used by tests then? If so, I'm not sure about obsoleting the `Assembly.GetEntryAssembly()` in the first place then.

To reiterate, my issue is the global obsoleting alongside the location of the replacement method. Removing the former will resolve the issue.

Susko3: Since `DebugUtils.GetEntryAssembly()` was used in places that are not related to debug/testing, I decided to move it to `RuntimeInfo.EntryAssembly`.

Please check whether the new initialisation logic used there is too confusing.

smoogipoo (approved):

PR #5620, opened 2023-01-08
Fix `MarkdownFencedCodeBlock` potentially adding empty space at the end of the block

Stedoss:
Noticed this one when looking at the `wiki/Skinning/skin.ini/Blank` wiki page in the `WikiOverlay` in osu!. That wiki page in particular has some good examples of this, which are shown below.

Behaviour should be the same, changed `AddParagraph()` to `AddText()` to allow for paragraphs to be created on new lines as before. Unsure we need tests for this one?

(Sorry for the horrible formatting of the table below...)
| Before | After |
| ------- | ------ |
| image | image |
| image | image |
| image | image |
| `Client/File_formats/Osu_(file_format)` image | image |

### DIFF
--- a/osu.Framework/Graphics/Containers/Markdown/MarkdownFencedCodeBlock.cs
+++ b/osu.Framework/Graphics/Containers/Markdown/MarkdownFencedCodeBlock.cs
@@ -43,6 +43,7 @@ namespace osu.Framework.Graphics.Containers.Markdown
                 textFlowContainer = CreateTextFlow(),
             };
 
+            // Markdig sometimes appends empty lines to the processed block, only add original lines to the container
             for (int i = 0; i < fencedCodeBlock.Lines.Count; i++)
                 textFlowContainer.AddParagraph(fencedCodeBlock.Lines.Lines[i].ToString());
         }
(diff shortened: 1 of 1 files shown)


bdach:
At least a visual-only test for this would be prudent yes. I'd like to double-check why this is happening precisely without having to set up a framework checkout.

bdach:
Upon checking out the test scene I'm confused. How is this not an upstream bug?

image

Kinda weird to be working around it like that without even a mention that it's a workaround...

Stedoss:
This is what was confusing me too, really. I found very little documentation around this from markdig, and the first(?) commit for this in o!f also had a funky workaround for it: https://github.com/ppy/osu-framework/blob/494601b72dacde8094a12bbc79442c33d8384192/osu.Framework/Graphics/Containers/Markdown/MarkdownFencedCodeBlock.cs#L41

It's pretty unclear to me what markdig wants us to use to get the content of the block, so I figured it would be fine to use their own `ToString()` implementation.

If we want to use the existing line-iterating route then it should be reported upstream (probably should be anyways), and we need to be careful of potential issues with updating markdig on mobile platforms (maybe AOT has been fixed for newer versions?).

bdach:
If `fencedCodeBlock.Lines.Count` works correctly then I'd prefer the old loop over the lines back, but with the bounds specified manually (i.e. `for (int i = ...)` instead of `foreach`) rather than having to do `.ToString()` to join the lines once, to only split them again internally in the framework for proper display.

bdach (approved):

PR #5622, opened 2023-01-11
Disallow using `AddInternal` in game class

frenzibyte:
- Companion to https://github.com/ppy/osu/pull/22141 

`Game` already wraps a container for `Content`, so there have been two ways to add components to a game, which lead to broken drawable order as seen in ppy/osu#22141 due to using both ways (`AddInternal` and `Add`). 

To limit this from happening, `AddInternal` has been completely disabled in favour of using `Add` instead.

### DIFF
--- a/osu.Framework/Game.cs
+++ b/osu.Framework/Game.cs
@@ -120,7 +120,7 @@ namespace osu.Framework
             });
         }
 
-        protected sealed override void AddInternal(Drawable drawable) => throw new InvalidOperationException($"Use {nameof(Content)} instead.");
+        protected sealed override void AddInternal(Drawable drawable) => throw new InvalidOperationException($"Use {nameof(Add)} or {nameof(Content)} instead.");
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


osu.Framework/Game.cs
```diff
@@ -106,22 +106,22 @@ protected Game()
...
+            });
+
+            base.AddInternal(overlayContent = new DrawSizePreservingFillContainer
+            {
+                TargetDrawSize = new Vector2(1280, 960),
+                RelativeSizeAxes = Axes.Both,
             });
         }
 
+        protected sealed override void AddInternal(Drawable drawable) => throw new InvalidOperationException($"Use {nameof(Content)} instead.");
```
peppy, line 123: Not sure about the clarity of this message. Maybe mention using `Add` *or* `Content` instead?

peppy (approved):

peppy:
@ppy/area-client requesting one more review on this to make sure this is an agreed upon direction.

smoogipoo (approved):

PR #5623, opened 2023-01-12
Disable 1px width hack when using Direct3D 11

Susko3:
This hack is not needed on Direct3D 11 as exclusive fullscreen is [disabled by default in DirectX](https://devblogs.microsoft.com/directx/demystifying-full-screen-optimizations/) (as evident by overlays showing on top of the game).

The hack appears to also be required on Vulkan (tested on AMD GPU):

> Amount of extra width added to window size when in borderless mode on Windows.
> Some drivers require this to avoid the window switching to exclusive fullscreen automatically.

I don't have this exact exclusive fullscreen issue, but I do get something similar: disabling the hack (set value to 0) results in a short black flicker when alt-tabbing in and out of the game (both OpenGL and Vulkan). There is no flicker when using Direct3D 11, in both fullscreen and borderless modes.

---

I tried out returning `base.SetBorderless(Display display)` instead of the custom window moving logic when the hack is disabled, but this didn't work as expected. This would use `SDL_WindowFlags.SDL_WINDOW_FULLSCREEN_DESKTOP` causing alt-tab to minimise the window. The current (and expected) behaviour is that the windows stays visible.

### DIFF
--- a/osu.Framework/Platform/Windows/WindowsWindow.cs
+++ b/osu.Framework/Platform/Windows/WindowsWindow.cs
@@ -33,6 +33,18 @@ namespace osu.Framework.Platform.Windows
         public WindowsWindow(GraphicsSurfaceType surfaceType)
             : base(surfaceType)
         {
+            switch (surfaceType)
+            {
+                case GraphicsSurfaceType.OpenGL:
+                case GraphicsSurfaceType.Vulkan:
+                    windowsBorderlessWidthHack = 1;
+                    break;
+
+                case GraphicsSurfaceType.Direct3D11:
+                    windowsBorderlessWidthHack = 0;
+                    break;
+            }
+
             try
             {
                 // SDL doesn't handle DPI correctly on windows, but this brings things mostly in-line with expectations. (https://bugzilla.libsdl.org/show_bug.cgi?id=3281)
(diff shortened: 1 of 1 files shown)


peppy:
The black flicker means exclusive is actually working, which suggests the hack is breaking it (I would not be at all surprised). From that perspective, it should probably also be disabled for DX10?

Susko3:
Unless I'm missing something obvious, there's no DX10?

https://github.com/ppy/osu-framework/blob/c5b00495ba9dc655107c40157342915e47a59530/osu.Framework/Platform/GraphicsSurfaceType.cs#L13-L35

peppy:
Right, I was just going off your comment without checking the code. But just to confirm, the black flash is both expected and a good sign.

Susko3:
The black flash is a good sign in fullscreen mode. And this does happen for both Vulkan and OpenGL with Veldrid.

But it's a bad sign and should not happen in borderless fullscreen. With the 1px hack enabled, there is no flash in Vulkan nor OpenGL. With it disabled there is a black flash. So I've kept it enabled.

peppy:
Right, fair call.

osu.Framework/Platform/Windows/WindowsWindow.cs
```diff
@@ -33,6 +33,18 @@ public class WindowsWindow : SDL2DesktopWindow
         public WindowsWindow(GraphicsSurfaceType surfaceType)
             : base(surfaceType)
         {
+            switch (surfaceType)
+            {
+                case GraphicsSurfaceType.OpenGL:
+                case GraphicsSurfaceType.Vulkan:
+                    windowsBorderlessWidthHack = 1;
```
peppy, line 40: Probably not a huge deal because this is such internal logic, but I would have preferred this as a `bool applyBorderlessWindowHack` and leaving the original `const` as-is.

peppy (approved):

PR #5625, opened 2023-01-14
Move `IconGroup` out of `osu.Framework.Platform.Windows.Native` namespace and annotate with `SupportedOSPlatformAttribute`

Susko3:
It was really wrong that `SDL2DesktopWindow` was using `osu.Framework.Platform.Windows.Native`. This was caused by `IconGroup` being in that namespace, even though most of `IconGroup` is platform-agnostic.

This PR moves `IconGroup` to the more general `osu.Framework.Platform` namespace and then appropriately annotates the windows-specific `Icon`/`CreateIcon` API usage with [`SupportedOSPlatform`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.versioning.supportedosplatformattribute?view=net-7.0).

### DIFF
--- a/osu.Framework/Host.cs
+++ b/osu.Framework/Host.cs
@@ -4,6 +4,7 @@
 #nullable disable
 
 using System;
+using System.Diagnostics;
 using osu.Framework.Platform;
 using osu.Framework.Platform.Linux;
 using osu.Framework.Platform.MacOS;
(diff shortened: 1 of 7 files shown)


osu.Framework/Host.cs
```diff
@@ -17,7 +17,7 @@ public static DesktopGameHost GetSuitableDesktopHost(string gameName, HostOption
         {
             switch (RuntimeInfo.OS)
             {
-                case RuntimeInfo.Platform.Windows:
+                case RuntimeInfo.Platform.Windows when OperatingSystem.IsWindows():
                     return new WindowsGameHost(gameName, hostOptions);
```
frenzibyte, line 21: We could've avoided this only if [`SupportedOSPlatformGuard` had support for enums](https://github.com/dotnet/runtime/issues/51541#issue-862245608), but [alas](https://github.com/dotnet/runtime/issues/51541#issuecomment-827825898). Maybe the time to move from `RuntimeInfo.OS` switches to `OperatingSystem.IsXYZ` guards is coming near us. But at least move this to a debug assertion for now.

smoogipoo (approved):

PR #5626, opened 2023-01-15
Add `Pad{Left,Right}` extensions for `LocalisableString` and `ILocalisableStringData`

Susko3:
For use in https://github.com/ppy/osu/pull/22208#discussion_r1070462373. Mirrors `string.Pad{Left,Right}` methods.

### DIFF
--- a/osu.Framework.Tests/Localisation/LocalisationTest.cs
+++ b/osu.Framework.Tests/Localisation/LocalisationTest.cs
@@ -458,6 +458,40 @@ namespace osu.Framework.Tests.Localisation
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


frenzibyte:
Not sure whether these methods should still exist, given that the only use case for them was deemed bad in the first place. I understand that `LocalisableString` should have similar methods to `string`, but probably not worth adding it if it won't be used anyway? @ppy/area-client thoughts?

bdach:
Tend to agree that this should not be a thing.

Susko3:
Closing as it's not needed. Can be revisited if it ever comes up as a valid use case,

PR #5627, opened 2023-01-16
Retrieve tags from Bass tracks when available

naoey:
For use in ppy/osu#21189. Only retrieves a subset of all the tags read by Bass that seemed relevant to osu!.


osu.Framework/Audio/Track/ITrack.cs
```diff
@@ -46,6 +46,11 @@ public interface ITrack : IAdjustableClock, IHasAmplitudes, IAdjustableAudioComp
         /// </summary>
         int? Bitrate { get; }
 
+        /// <summary>
+        /// The ID3 tags contained in this track's metadata.
+        /// </summary>
+        Tags? Tags { get; }
```
bdach, line 52: I'm really not sure if we want this on `Track`. Especially so if it's going to not be implemented for anything but `TrackBass`.

Alternative would be something like

```csharp
public interface IHasID3Tags
{
    Tags Tags { get; }
}
```

which could only be applied to `TrackBass`, but that likely causes trouble with `DrawableTrack` (the inner track is private, so you wouldn't be able to access the tags otherwise).

osu.Framework/Audio/Track/Track.cs
```diff
@@ -23,6 +23,8 @@ public abstract class Track : AdjustableAudioComponent, ITrack, IAudioChannel
 
         public virtual bool Looping { get; set; }
 
+        public Tags? Tags { get => GetTags(); }
```
bdach, line 26: What's the point of this property-method split? Is it for encapsulation? I'm not super sure it's worth it as opposed to

```csharp
        public virtual Tags? Tags => null;
```

If really required though:

```suggestion
        public Tags? Tags => GetTags();
```

osu.Framework/Audio/Track/TrackBass.cs
```diff
@@ -303,6 +303,33 @@ public override async Task<bool> SeekAsync(double seek)
...
+                Artist = tags.Artist?.Trim()!,
+                Title = tags.Title,
+                Album = tags.Album,
+                Genre = tags.Genre,
+            };
+
+            if (string.IsNullOrEmpty(parsed.Artist))
+                parsed.Artist = tags.AlbumArtist;
+
+            if (Int32.TryParse(tags.BPM, out int bpm))
```
bdach, line 324:
```suggestion
            if (int.TryParse(tags.BPM, out int bpm))
```

Also what's the guarantee that this is gonna parse as `int`? What about fractional BPM?

bdach:
Also please check code style.

osu.Framework/Audio/Track/ITrack.cs
naoey: Made sense in my head as tags can be present on the track regardless of implementation (Bass or otherwise).

`DrawableTrack`'s inner track could probably be public since it's already readonly?

bdach: Sure, but layering wise I'm not sure it fits smack dab in the middle of the rest of the audio subsystem.

I'm just mostly pondering. Need more opinions on this for sure before proceeding in any particular direction.

osu.Framework/Audio/Track/TrackBass.cs
naoey: Hadn't considered that, on second thought not even sure this is worth having because it doesn't seem to be populated very commonly.

osu.Framework/Audio/Track/ITrack.cs
peppy: I remember there being an overhead with tag retrieval, in stable it’s only requested in the editor to avoid this.

Will need to check if this is still the case.

osu.Framework/Audio/Track/Track.cs
```diff
@@ -23,6 +23,8 @@ public abstract class Track : AdjustableAudioComponent, ITrack, IAudioChannel
 
         public virtual bool Looping { get; set; }
 
+        public Tags? Tags => GetTags();
```
peppy, line 26: If we're going to go as far as having a property, it should probably cache so they are only retrieved once.

PR #5631, opened 2023-01-20
Fix emphasised markdown text not flowing correctly

Joehuu:
Before:
chrome_6FlAF9cz2f

After:
chrome_hqjKkCM4A7

### DIFF
--- a/osu.Framework/Graphics/Containers/Markdown/MarkdownTextFlowContainer.cs
+++ b/osu.Framework/Graphics/Containers/Markdown/MarkdownTextFlowContainer.cs
@@ -186,28 +186,19 @@ namespace osu.Framework.Graphics.Containers.Markdown
                 }
             }
 
-            var textDrawable = CreateEmphasisedSpriteText(hasBold, hasItalic);
-            textDrawable.Text = text;
-
-            AddDrawable(textDrawable);
+            AddText(text, t => ApplyEmphasisedCreationParameters(t, hasBold, hasItalic));
         }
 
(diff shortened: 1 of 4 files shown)


osu.Framework/Graphics/Containers/Markdown/MarkdownTextFlowContainer.cs
```diff
@@ -186,28 +186,19 @@ private void addEmphasis(string text, List<string> emphases)
...
         /// <param name="italic">Whether the text should be italicised.</param>
-        /// <returns>The <see cref="SpriteText"/> with emphases applied.</returns>
-        protected virtual SpriteText CreateEmphasisedSpriteText(bool bold, bool italic)
-        {
-            var textDrawable = CreateSpriteText();
-
-            textDrawable.Font = textDrawable.Font.With(weight: bold ? "Bold" : null, italics: italic);
-
-            return textDrawable;
-        }
```
frenzibyte, line 210: I would've kept this method and used `AddPart(CreateChunkFor(text, true, CreateEmphasisedSpriteText))` to avoid breaking changes, but it's not that bad to consider doing so.

frenzibyte (approved):

osu.Framework/Graphics/Containers/Markdown/MarkdownTextFlowContainer.cs
frenzibyte: gonna request one more opinion from @ppy/team-client on whether it's fine to break this before merge, we can avoid it but I feel the new method makes more sense.

smoogipoo (approved):

PR #5632, opened 2023-01-21
Add composite text chunks

Feodor0090:
Their main purpose is to allow localization of text flows which have parts with different formatting. 

- I moved code which parsed `[xyz]` things from text to base class to avoid dublicating.
- "`\n` = new paragraph" is forced here because children chunks add line breaks before themselves when not.
- Not sure about `drawable = (Drawable?)cb.DynamicInvoke(args);` `if (drawable == null)` thing in customizable chunk - when it may return null? Maybe it's better to throw?

### DIFF
--- a/osu.Framework.Tests/Visual/Localisation/TestSceneCompositeChunkLocalisation.cs
+++ b/osu.Framework.Tests/Visual/Localisation/TestSceneCompositeChunkLocalisation.cs
@@ -1,0 +1,135 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using NUnit.Framework;
+using osu.Framework.Allocation;
+using osu.Framework.Extensions.LocalisationExtensions;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Shapes;
+using osu.Framework.Graphics.Sprites;
+using osu.Framework.Localisation;
+using osuTK.Graphics;
+
+namespace osu.Framework.Tests.Visual.Localisation
+{
+    public partial class TestSceneCompositeChunkLocalisation : LocalisationTestScene
+    {
+        private const string rank = "rank";
+        private const string rank_default = "[0] achieved rank #[1] on [2] ([3])";
+        private const string simple = "simple";
+
+        private CompositeTextChunk<SpriteText>? chunk;
+        private Color4 color = Color4.Cyan;
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            // strings sourced from osu-web crowdin (https://crowdin.com/translate/osu-web/)
+            Manager.AddLanguage("en", new TestLocalisationStore("en", new Dictionary<string, string>
+            {
+                [rank] = rank_default,
+                [simple] = "simple english",
+            }));
+
+            Manager.AddLanguage("fr", new TestLocalisationStore("fr", new Dictionary<string, string>
+            {
+                [rank] = "[0] a atteint le rang #[1] sur [2] ([3])",
+                [simple] = "simple french",
+            }));
+
+            Manager.AddLanguage("tr", new TestLocalisationStore("tr", new Dictionary<string, string>
+            {
+                [rank] = "[0] [2] ([3]) beatmapinde #[1] sıralamaya ulaştı",
+                [simple] = "simple turkish",
+            }));
+        }
+
+        [Test]
+        public void TestTextFlowLocalisation()
+        {
+            Container? textFlowParent = null;
+            TextFlowContainer? textFlowContainer = null;
+
+            AddStep("create text flow", () => Child = textFlowParent = new Container
+            {
+                RelativeSizeAxes = Axes.X,
+                AutoSizeAxes = Axes.Y,
+                Origin = Anchor.Centre,
+                Anchor = Anchor.Centre,
+                Children = new Drawable[]
+                {
+                    new Box
+                    {
+                        RelativeSizeAxes = Axes.Both,
+                        Colour = Colour4.FromHex("#333")
+                    },
+                    textFlowContainer = new TextFlowContainer(text => text.Font = FrameworkFont.Condensed)
+                    {
+                        RelativeSizeAxes = Axes.X,
+                        AutoSizeAxes = Axes.Y,
+                        Origin = Anchor.Centre,
+                        Anchor = Anchor.Centre,
+                    },
+                }
+            });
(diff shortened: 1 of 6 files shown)


osu.Framework/Graphics/Containers/FormattableTextChunk.cs
```diff
@@ -0,0 +1,75 @@
...
+    /// <typeparam name="TSpriteText"></typeparam>
+    public abstract class FormattableTextChunk<TSpriteText> : TextChunk<TSpriteText>
+        where TSpriteText : SpriteText, new()
+    {
+        protected FormattableTextChunk(LocalisableString text, bool newLineIsParagraph, Func<TSpriteText> creationFunc, Action<TSpriteText>? creationParameters = null)
+            : base(text, newLineIsParagraph, creationFunc, creationParameters)
+        {
+        }
+
+        protected abstract IEnumerable<Drawable>? GetDrawablesForSubstitution(string placeholder, TextFlowContainer textFlowContainer);
```
bdach, line 23: No xmldoc?

osu.Framework/Graphics/Containers/FormattableTextChunk.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        protected override IEnumerable<Drawable> CreateDrawablesFor(string text, TextFlowContainer textFlowContainer)
```
bdach, line 25: Should probably be `sealed`.

osu.Framework/Graphics/Containers/FormattableTextChunk.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            var sprites = new List<Drawable>();
+            int index = 0;
+            string str = text;
+
+            while (index < str.Length)
+            {
+                IEnumerable<Drawable>? placeholderDrawables = null;
+                int nextPlaceholderIndex = str.IndexOf(CustomizableTextContainer.UNESCAPED_LEFT, index, StringComparison.Ordinal);
```
bdach, line 34: This constant referencing is getting real backwards now. `CustomizableTextContainer` is a subtype of `TextFlowContainer`, but `FormattableTextChunk` is a base class of both `CustomizableTextChunk` and `CompositeTextChunk`, the latter of which is now usable in a standard `TextFlowContainer`, but is still using escape constants from `CustomizableTextContainer` again?

Constants should be moved here if this is to be the direction (which I'm unsure of - more on this later).

osu.Framework/Graphics/Containers/CustomizableTextContainer.cs
```diff
@@ -88,9 +87,10 @@ public int AddPlaceholder(Drawable drawable)
         /// <param name="name">The name of the placeholder.</param>
         /// <param name="iconFactory">The icon factory matching <paramref name="name"/>, if the method returned <see langword="true"/>.</param>
         /// <returns>Whether an icon factory was found for the given <paramref name="name"/>.</returns>
-        internal bool TryGetIconFactory(string name, out Delegate iconFactory) => iconFactories.TryGetValue(name, out iconFactory);
+        internal bool TryGetIconFactory(string name, [MaybeNullWhen(false)] out Delegate iconFactory) => iconFactories.TryGetValue(name, out iconFactory);
```
bdach, line 90: Prefer the other way around if I'm honest, i.e.

```suggestion
        internal bool TryGetIconFactory(string name, [NotNullWhen(true)] out Delegate? iconFactory) => iconFactories.TryGetValue(name, out iconFactory);
```

osu.Framework.Tests/Visual/Localisation/TestSceneCompositeChunkLocalisation.cs
```diff
@@ -0,0 +1,135 @@
...
+using osu.Framework.Graphics.Sprites;
+using osu.Framework.Localisation;
+using osuTK.Graphics;
+
+namespace osu.Framework.Tests.Visual.Localisation
+{
+    public partial class TestSceneCompositeChunkLocalisation : LocalisationTestScene
+    {
+        private const string rank = "rank";
+        private const string rank_default = "[0] achieved rank #[1] on [2] ([3])";
```
bdach, line 21: The thing that interests me most in this pull is that this usage code shows that the changes here are *not* immediately usable with localisation strings from `osu-resources`. From where, and how, is a string like this going to be retrieved?

First of all, `osu-resources` assets use [curly braces and not square ones](https://github.com/ppy/osu-resources/blob/master/osu.Game.Resources/Localisation/Web/Beatmaps.resx#L74). Secondly, the template string is not even *retrievable* via the C# classes because [you can't currently get them without filling the placeholders](https://github.com/ppy/osu-resources/blob/6619773926b3c1c0347a2ecfdd06fbe5b7f1d87d/osu.Game.Resources/Localisation/Web/BeatmapsStrings.cs#L35).

So what's the plan here of making this work, then? Is every single usage going to use the method linked above but just hardcode square-bracketed placeholders back in again?

bdach (changes requested):
Just from a brief source read-through

osu.Framework.Tests/Visual/Localisation/TestSceneCompositeChunkLocalisation.cs
Feodor0090: Giving the fact that this thing is going to be used only in few places (right now i can only remember registration screen and skin editor title), i hoped that as exception the strings will be kept in crowdin/resources already with `[]` to bypass `TranslatableString`'s formatting. If this won't happen, formatting `{}` to `[]` sounds better to me than creating a way to get unformatted string, considering how they arrive to a chunk and how they are exposed (you still need to pass something to substitute just to get string's object).

Introducing translatable string class without formatting at all won't work IMO - if usage is guaranteed to be single, it can be kept with `[]` as is, if it doesn't, there will be more mess in localization classes (2 props for the same string).

bdach: > i hoped that as exception the strings will be kept in crowdin/resources already with `[]` to bypass `TranslatableString`'s formatting.

No can do. Not without extra work. The web strings to resources pipeline is fully automated. How is the automated tool supposed to know in which strings to leave placeholders in, and for which to generate the function helpers?

> Introducing translatable string class without formatting at all won't work IMO - if usage is guaranteed to be single, it can be kept with `[]` as is, if it doesn't, there will be more mess in localization classes (2 props for the same string).

I think that is completely fine, personally.

As an alternative however, I will offer this as food for thought: what if `TextFlowContainer` could look at a `LocalisableString`, recognise that it is a `TranslatableString` with params, and internally handle this so that you can call the resources-provided method for the full string, but then *also* provide a set of formatting definitions that would apply on top of that? This seems functionally feasible, and eliminates this entire discussion with no extra changes required.

PR #5633, opened 2023-01-21
Fix bindable list/dictionary not correctly checking for already bound instances

frenzibyte:
I've also changed `Bindable` to match the message (and removed the instance string output part because it's already bound and will share the same value, therefore pointless). 

This may be the ultimate key as to why we're seeing these `IndexOutOfRangeException`s osu!-side with multiplayer tests (cc @smoogipoo, might interest you). It's significantly hinted in the error output:

CleanShot 2023-01-21 at 21 39 19

As it can be seen, `MultiplayerClient` removes the playlist item via index, and that first propagates the change from the first bindable to the second bindable, which then propagates to another bindable that since fails with the `IndexOutOfRangeException`. What I'm thinking is that the third bindable is actually the original bindable, and that couldn't be caught both in `removeAt` because that only checks one level above, and in `BindTo` because the check was broken.

I'll run the tests locally with local framework checkout in hopes I still get to reproduce the failure, but failing that, time will tell on osu! CI where the unnecessary binding is coming from.

### DIFF
--- a/osu.Framework/Bindables/Bindable.cs
+++ b/osu.Framework/Bindables/Bindable.cs
@@ -199,8 +199,8 @@ namespace osu.Framework.Bindables
         /// <exception cref="InvalidOperationException">Thrown when attempting to bind to an already bound object.</exception>
         public virtual void BindTo(Bindable<T> them)
         {
-            if (Bindings?.Contains(them) == true)
-                throw new InvalidOperationException($"This bindable is already bound to the requested bindable ({them}).");
+            if (Bindings?.Contains(them.weakReference) == true)
+                throw new ArgumentException("An already bound bindable cannot be bound again.");
 
             them.CopyTo(this);
 
--- a/osu.Framework/Bindables/BindableDictionary.cs
+++ b/osu.Framework/Bindables/BindableDictionary.cs
@@ -454,10 +454,10 @@ namespace osu.Framework.Bindables
         {
             ArgumentNullException.ThrowIfNull(them);
 
-            if (bindings?.Contains(weakReference) == true)
-                throw new ArgumentException("An already bound collection can not be bound again.");
+            if (bindings?.Contains(them.weakReference) == true)
+                throw new ArgumentException("An already bound dictionary can not be bound again.");
             if (them == this)
-                throw new ArgumentException("A collection can not be bound to itself");
+                throw new ArgumentException("A dictionary can not be bound to itself");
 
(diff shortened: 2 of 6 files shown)


frenzibyte:
Upon investigating the test failures I was targeting in OP, it turns out this PR doesn't actually help in fixing them at all (see [discord](https://discord.com/channels/188630481301012481/589331078574112768/1066829250608500817)). It's still valid to keep open and merge though.

osu.Framework/Bindables/Bindable.cs
```diff
@@ -199,8 +199,8 @@ public virtual void CopyTo(Bindable<T> them)
         /// <exception cref="InvalidOperationException">Thrown when attempting to bind to an already bound object.</exception>
         public virtual void BindTo(Bindable<T> them)
         {
-            if (Bindings?.Contains(them) == true)
-                throw new InvalidOperationException($"This bindable is already bound to the requested bindable ({them}).");
```
peppy, line 203: I think this message was intended to expose the *type* of the target bindable, rather than the value. Not sure if it's worth considering that (and using `GetType` instead of implicit `ToString`) or not. Maybe?

osu.Framework/Bindables/Bindable.cs
```diff
@@ -199,8 +199,8 @@ public virtual void CopyTo(Bindable<T> them)
...
+            if (Bindings?.Contains(them.weakReference) == true)
```
peppy, line 202: All tests still look to pass without this change. Am I missing something?

frenzibyte: This is just sort of an optimisation. As checking via instance goes through a lot more checks than just the weak reference.

osu.Framework/Bindables/BindableDictionary.cs
```diff
@@ -454,10 +454,10 @@ public void BindTo(BindableDictionary<TKey, TValue> them)
         {
             ArgumentNullException.ThrowIfNull(them);
 
-            if (bindings?.Contains(weakReference) == true)
-                throw new ArgumentException("An already bound collection can not be bound again.");
+            if (bindings?.Contains(them.weakReference) == true)
```
peppy, line 457: Oh, I missed THIS. Didn't realise this was the actual fix.

peppy (approved):
I think this is okay

smoogipoo (approved):

PR #5636, opened 2023-01-25
Expose truncated state of `SpriteText`

Joehuu:
- Supersedes and closes https://github.com/ppy/osu-framework/pull/4286

Use case is in https://github.com/ppy/osu/compare/master...Joehuu:osu:truncating-text-tooltips.

The way around the overhead is to limit it to `Truncate = true` cases + overriding `HandlePositionalInput` to check when an ellipsis is added. Can't really override that on base `SpriteText` as it'll affect other inherited classes like `DrawableDate`.


peppy (dismissed):

peppy:
Looks fine.

@Joehuu for the osu! side implementation, i'd not call `base.HandlePositionalInput` at all (kinda redundant) and i'd also make `TruncatingSpriteText` a `sealed` class because abstracting further may lead to unexpected issues with input handling.

smoogipoo:
I worry that this is going to become a case of "change every text in the game to an `OsuTruncatingSpriteText`". Are you expecting this to be the case? Note that the optimisation of `HandlePositionalInput` is based on whether any of the methods have been overridden and not whether they've returned true/false.

peppy:
> Note that the optimisation of `HandlePositionalInput` is based on whether any of the methods have been overridden and not whether they've returned true/false.

That's precisely why I proposed sealing the class and not calling the base method – it would maintain the correct optimisation in all usages if done that way.

Joehuu:
> I worry that this is going to become a case of "change every text in the game to an OsuTruncatingSpriteText". Are you expecting this to be the case?

This won't be applied to every text. The osu!-side PR is only changing a subset of them, mostly text that we have no control over (usernames, beatmap metadata, and skin names).

peppy:
I don't see any harm in getting this in, regardless of whether we are using it or not. Could probably do with test coverage but honestly truncation doesn't have any yet and it's worked okay, so it can come later if ever.

peppy (approved):

PR #5637, opened 2023-01-26
Improve bindable binding thread safety

### LINKED ISSUE #5635: Resolving bindables via dependency injection is unsafe
Dependency injection happens on the load thread, and resolving bindables require creating a bound copy of the cached bindable. When a drawable is asynchronously loaded, there's a chance for the resolved bindable to be desynchronised from the remote bindable, if the remote bindable changes as the local bindable is being bound.

This is the direct reason for the `ArgumentOutOfRangeException` test failures that've been recurring in osu!. Here's brief logging of what's going on:
```
[runtime] 2023-01-22 21:51:14 [verbose]: [ThreadedTaskScheduler (LoadComponentsAsync (standard)), 39] Creating new playlist item bindable list with number 6720
[runtime] 2023-01-22 21:51:14 [verbose]: [Update (GameThread), 19] BindableList<PlaylistItem>(6693): adding 1 item(s) to list
[runtime] 2023-01-22 21:51:14 [verbose]: [ThreadedTaskScheduler (LoadComponentsAsync (standard)), 39] Binding bindable instance 6720 to 6693
[runtime] 2023-01-22 21:51:14 [verbose]: [ThreadedTaskScheduler (LoadComponentsAsync (standard)), 39] Parsing content of bindable instance 6693 to 6720
[runtime] 2023-01-22 21:51:14 [verbose]: [ThreadedTaskScheduler (LoadComponentsAsync (standard)), 39] Bindables 6720 and 6693 have equal content (0, 0), nothing changed
...
[runtime] 2023-01-22 21:51:14 [verbose]: [Update (GameThread), 19] BindableList<PlaylistItem>(6718): removing item with index 0, item: osu.Game.Online.Rooms.PlaylistItem, caller is 6693
[runtime] 2023-01-22 21:51:14 [verbose]: [Update (GameThread), 19] BindableList<PlaylistItem>(6719): removing item with index 0, item: osu.Game.Online.Rooms.PlaylistItem, caller is 6693
[runtime] 2023-01-22 21:51:14 [verbose]: [Update (GameThread), 19] BindableList<PlaylistItem>(6720): removing item with index 0, item: doesn't exist, caller is 6693, stack trace
[runtime] 2023-01-22 21:51:14 [error]: An unhandled error has occurred.
[runtime] 2023-01-22 21:51:14 [error]: ---> System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
[runtime] 2023-01-22 21:51:14 [error]: at osu.Framework.Bindables.BindableList`1.removeAt(Int32 index, BindableList`1 caller)
[runtime] 2023-01-22 21:51:14 [error]: at osu.Framework.Bindables.BindableList`1.removeAt(Int32 index, BindableList`1 caller)
[runtime] 2023-01-22 21:51:14 [error]: at osu.Framework.Bindables.BindableList`1.removeAt(Int32 index, BindableList`1 caller)
```

As can be seen above, one of the bindables (6720) failed to copy from a remote bindable (6693), since the remote bindable was being changed in a separate thread, and the local bindable failed to see the change.

I'm seeing multiple ways through fixing this, and I need more opinions on the matter to come to an amicable solution here:
 - Delay binding to remote bindables until `LoadComplete` (this follows a rule of thumb we use on bindables in general, which is binding and listening to value changes in `LoadComplete` / update thread)
 - Add thread safety to binding bindables via mutex lock (skeptical of this way as it tends to increase the chances of deadlocks, but need more voices)

cc/ @ppy/team-client

frenzibyte:
- Closes #5635 

Until now, `BindTo` was always unsafe against cross-thread operations. If a bindable was to be changed on one thread, and another bindable was trying to bind to it on another thread, then there's a chance for the other bindable to be in a desynchronised state from its remote, in the following way:
 - Thread 1: Bindable 2 copies from bindable 1
 - Thread 2: Bindable 1 changes value and propagates to bindings (none currently)
 - Thread 1: Bindable 2 adds bindable 1 to bindings list and vice versa
 - Result: Bindable 2 has stale value and is now desynchronised from bindable 1.

The above is happening in #5635 when a drawable is loaded in an asynchronous thread, since in that thread DI creates a bound copy of the cached bindables, thus using `BindTo` on an async thread.

I'm not all that familiar into thread synchronisation, so I would appreciate thoroughly checking this PR in case I failed to notice an edge case here.


osu.Framework/Bindables/Bindable.cs
```diff
@@ -202,10 +213,13 @@ public virtual void BindTo(Bindable<T> them)
...
-            them.CopyTo(this);
+            lock (them.bindableChangeLock)
+            {
+                them.CopyTo(this);
 
-            addWeakReference(them.weakReference);
-            them.addWeakReference(weakReference);
+                addWeakReference(them.weakReference);
+                them.addWeakReference(weakReference);
+            }
```
bdach, line 222: This doesn't look much safer than the old code. This sort of reciprocal two-way sync code taking only *one* of the two relevant locks rings alarm bells for me immediately.

Here's a theoretical race scenario that exploits this (I haven't experimentally confirmed, but I'm fairly certain this is a real issue):

```
thread 1                              | thread 2
--------------------------------------+--------------------------------------
this.Value = 5;                       |
                                      |
this.BindTo(them);                    |
                                      |
    lock (them.BindableChangeLock)    |
    {                                 |
        them.CopyTo(this);            |
                                      |
                                      | this.Value = 7;
                                      |
                                      |     // never taken by the bind!
                                      |     lock (BindableChangeLocK) 
                                      |     {
                                      |          // value change callbacks
                                      |          // won't fire yet as weakrefs
                                      |          // are not set up yet!
                                      |     }
                                      |
        // bindings are set up here   |
        // but Value has desynced!    |
    }                                 |
```

but really the same breakage applies for this entire critical section (if you don't take both locks when weak references are being added, you can end up not propagating changes from either bindable to the other).

The same goes for unbind, and for the other bindable types too, obviously.

frenzibyte: Good catch, I initially was locking on both the local and the remote bindable lock objects, but I couldn't figure out a way to add a reliable test case for it, which incorrectly led me to think it's impossible to happen. I'll add the lock.

PR #5638, opened 2023-01-28
Fix `AudioManager.GetSampleStore()` mutating provided store 

bdach:
This PR intends to resolve what I perceive to be the root cause of https://github.com/ppy/osu/issues/22084. Namely, `AudioManager.GetSampleStore()` is not a pure method, as it mutates the store provided to it in the constructor, which is a footgun if the same store is then provided to some other component.

In the case of the aforementioned issue, [the same store is used to create a sample store and a texture store](https://github.com/ppy/osu/blob/d38316bf4f552629cf86dc5ccfe78c3f2fd95f27/osu.Game/Skinning/Skin.cs#L71-L80), which then caused the texture store to receive audio streams (as its underlying storage would be biased towards returning audio), which it then attempted to read as images and invariably failed in doing so.

Initially I fixed this game-side, but then upon noticing that `TextureLoaderStore` does this properly:

https://github.com/ppy/osu-framework/blob/9a50bc4b11def8aecff3f6eb0aa49cfd5c1278dd/osu.Framework/Graphics/Textures/TextureLoaderStore.cs#L26-L28

I decided this is the generally more correct approach.

Note that isolating the passed-in storage alone would lead to a bit of a decrease in flexibility, as you may want to be able to add automatic lookups for other extensions (as osu! already did for `.ogg` files in the snippet linked above). This is why `ISampleStore` also gains an `AddExtension(string)` method, to compensate for this.

The osu!-side changes that need to be applied combined with this pull to fix the game-side issue can be previewed at https://github.com/ppy/osu/compare/master...bdach:osu:conflicting-filenames-in-skin-alt?expand=1.

### DIFF
--- a/osu.Framework.Tests/Visual/Audio/TestSceneAudioManager.cs
+++ b/osu.Framework.Tests/Visual/Audio/TestSceneAudioManager.cs
@@ -3,9 +3,12 @@
 
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using NUnit.Framework;
 using osu.Framework.Allocation;
 using osu.Framework.Audio;
+using osu.Framework.Audio.Sample;
 using osu.Framework.IO.Stores;
 using osu.Framework.Testing;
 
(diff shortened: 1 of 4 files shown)


peppy (approved):
Looks very sane.

PR #5639, opened 2023-02-03
Remove improper usage causing double-disposing of `Game` 

Susko3:
While a `Game` is `IDisposable`, it's the responsibility of the `GameHost` to dispose it when exiting. So disposing with `using` or manually by calling `Dispose()` is not necessary. Luckily, this double-dispose doesn't have any side-effects, as `CompositeDrawable` would early return and stop any potential disposal chain.

I've checked osu!-side and there are no improper usages there (except an [intended one](https://github.com/ppy/osu/blob/845bbf55fe5d48293afb159c603a177d00bffb65/osu.Game.Tests/Visual/Navigation/TestSceneScreenNavigation.cs#L692)).

This was checked by adding a disposed check and throw in `Game`/`OsuGameBase` and running all tests locally:
```cs
        {
            if (IsDisposed)
                throw new ObjectDisposedException("Game is disposed twice");

            // ...
        }
```

### DIFF
--- a/SampleGame.Desktop/Program.cs
+++ b/SampleGame.Desktop/Program.cs
@@ -13,8 +13,7 @@ namespace SampleGame.Desktop
         public static void Main(string[] args)
         {
             using (GameHost host = Host.GetSuitableDesktopHost(@"sample-game"))
-            using (Game game = new SampleGameGame())
-                host.Run(game);
+                host.Run(new SampleGameGame());
         }
     }
 }
--- a/osu.Framework.Templates/templates/template-empty/TemplateGame.Desktop/Program.cs
+++ b/osu.Framework.Templates/templates/template-empty/TemplateGame.Desktop/Program.cs
@@ -9,8 +9,7 @@ namespace TemplateGame.Desktop
         public static void Main()
         {
             using (GameHost host = Host.GetSuitableDesktopHost(@"TemplateGame"))
-            using (osu.Framework.Game game = new TemplateGameGame())
-                host.Run(game);
+                host.Run(new TemplateGameGame());
         }
     }
 }
--- a/osu.Framework.Templates/templates/template-empty/TemplateGame.Game.Tests/Program.cs
+++ b/osu.Framework.Templates/templates/template-empty/TemplateGame.Game.Tests/Program.cs
@@ -8,8 +8,7 @@ namespace TemplateGame.Game.Tests
         public static void Main()
         {
             using (GameHost host = Host.GetSuitableDesktopHost("visual-tests"))
-            using (var game = new TemplateGameTestBrowser())
-                host.Run(game);
+                host.Run(new TemplateGameTestBrowser());
         }
     }
 }
--- a/osu.Framework.Templates/templates/template-flappy/FlappyDon.Desktop/Program.cs
+++ b/osu.Framework.Templates/templates/template-flappy/FlappyDon.Desktop/Program.cs
@@ -9,10 +9,7 @@ namespace FlappyDon.Desktop
         public static void Main()
         {
             using (GameHost host = Host.GetSuitableDesktopHost(@"FlappyDon"))
-            using (osu.Framework.Game game = new FlappyDonGame())
-            {
-                host.Run(game);
-            }
+                host.Run(new FlappyDonGame());
         }
     }
 }
--- a/osu.Framework.Templates/templates/template-flappy/FlappyDon.Game.Tests/Program.cs
+++ b/osu.Framework.Templates/templates/template-flappy/FlappyDon.Game.Tests/Program.cs
@@ -9,8 +9,7 @@ namespace FlappyDon.Game.Tests
         public static void Main()
         {
             using (GameHost host = Host.GetSuitableDesktopHost("visual-tests"))
-            using (var game = new FlappyDonTestBrowser())
-                host.Run(game);
+                host.Run(new FlappyDonTestBrowser());
         }
     }
 }
--- a/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
+++ b/osu.Framework.Tests/Exceptions/TestLoadExceptions.cs
@@ -247,22 +247,21 @@ namespace osu.Framework.Tests.Exceptions
             {
                 using (var host = new TestRunHeadlessGameHost($"{GetType().Name}-{Guid.NewGuid()}", new HostOptions()))
                 {
-                    using (var game = new TestGame())
+                    var game = new TestGame();
+
+                    game.Schedule(() =>
                     {
-                        game.Schedule(() =>
+                        storage = host.Storage;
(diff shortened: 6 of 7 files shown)


peppy:
Not sure about this change. I feel that as a consumer, you should be actively disposing (usually via `using`) anything which is `IDisposable`, even if it's a noop. Removing these brings a degree of uncertainty, that if I was a user of the framework I would have to check into the source code and probably get lost for a while figuring out that it's safe to not dispose (as the templates are proposing with this change).

peppy (commented):
As commented.

bdach:
I agree with not removing, and will add the following excerpt from the [official guidelines on how to implement `Dispose()`](https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose):

> To help ensure that resources are always cleaned up appropriately, a [Dispose](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose) method should be idempotent, such that it is callable multiple times without throwing an exception. Furthermore, subsequent invocations of [Dispose](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose) should do nothing.

In fact, if I'm honest, I find the fact that the host is disposing the game _more_ disputable than the explicit disposals in the sorts of usages the template projects are doing, because in terms of ownership, it feels that the outward code "owns" the game and ownership is not transferred when the game is passed to a host - but it's probably okay, just to ensure proper cleanup for new framework users.

Susko3:
```cs
using (GameHost host = Host.GetSuitableDesktopHost(@"sample-game"))
{
    using (Game game = new SampleGameGame())
    {
        host.Run(game);
    } // <-- game is disposed and recursively disposes the entire drawable hierarchy below it
} // <-- host is disposed and disposes the ``Root`` (UserInputManager)
```

The first dispose is pulling the rug from under the `GameHost` and disposing its hierarchy while ignoring the [safety](https://github.com/ppy/osu-framework/blob/210c040de95d3a6a88d2b0fab07528d16a1330ff/osu.Framework/Platform/GameHost.cs#L1131-L1143) put in place by the host.

Guess I'll look away until it becomes a problem.

peppy:
Hmm, that does seem like a valid concern that I'd want to at least confirm and assess before closing this, so let's leave it open momentarily.

bdach:
Dunno why this valid concern was clearly described only now (OP is very nebulous in comparison), but this means that the disposal order in the usages mentioned is inverted if anything (i.e. the `GameHost` should be the inner `using`).

peppy:
Yeah, the new explanation changes things drastically. I'll need a bit of time to think how we can better resolve this.

peppy:
On revisiting this, I think that if we are going to fix this, we should begin by noting that calling `GameHost.Run` twice, even with two different game instances will fail at:

https://github.com/ppy/osu-framework/blob/210c040de95d3a6a88d2b0fab07528d16a1330ff/osu.Framework/Platform/GameHost.cs#L664-L665

Which brings up the question: should `GameHost` require disposal in the first place? Should cleanup just be done at the end of `Run`, before returning?

Also, the safety mentioned by @Susko3 above also doesn't look too bad in its current state. The main goal is to guard against `GameHost.Dispose` being called while it's still running, which won't be the case in (current) incorrect disposal order.

Wondering if the most valid answer would be:

```csharp
using (Game game = new SampleGameGame())
{
    Host
        .GetSuitableDesktopHost(@"sample-game")
        .Run(game);
}
```

where `GameHost` is no longer `IDisposable`. This would be a breaking change, but it'd also be an easy one for users to fix in < 1 minute, so I don't see that as a huge issue.

Susko3:
> Should GameHost require disposal in the first place?

Because the host created a lot of resources that are `IDisposable` I think it should also be `IDisposable`. This is reinforced in the article @bdach mentioned earlier:

> If your class owns a field or property, and its type implements IDisposable, the containing class itself should also implement IDisposable.

> Should cleanup just be done at the end of Run, before returning?

Per the above, I don't think disposal shouldn't happen there. This also raises concerns on mobile where `Run` is a non-blocking call, but cleanup doesn't happen there anyways.

>Wondering if the most valid answer would be:
>
> ```cs
>using (Game game = new SampleGameGame())
>{
>    Host
>        .GetSuitableDesktopHost(@"sample-game")
>        .Run(game);
>}

While this doesn't fix the original issue of the game being disposed twice (which is not actually an issue, per IDisposable guidelines). Having the game in the outermost scope does make it better, as the `using` disposal happens after host cleanup.

Something that I find unusual in this order of operation is that the game name is explicitly specified to the host, even though we already know what game we're (going to be) hosting.

---

Since `GameHost` takes ownership of the game (as it'll dispose it) would it make sense to pass in the game in the constructor? That would clearly signify transfer of ownership, and also the game name can be deduced from the passed in `Game`, requiring no additional parameters. (could be taken from `Drawable.Name`, a new abstract property, or a ctor param: `protected Game(string gameName)`)

This would also bring it closer to how mobile hosts work right now, where the game is created before the host. And would also fix the game name always being [`string.Empty`](https://github.com/ppy/osu-framework/blob/ba1385330cc501f34937e08257e586c84e35d772/osu.Framework/Platform/OsuTKGameHost.cs#L15).

So something like:
```cs
var game = new SampleGameGame();
using (GameHost host = Host.GetSuitableDesktopHost(game))
    host.Run();
```

bdach:
> Wondering if the most valid answer would be:
> 
> ```cs
> using (Game game = new SampleGameGame())
> {
>     Host
>         .GetSuitableDesktopHost(@"sample-game")
>         .Run(game);
> }
> ```
> 
> where `GameHost` is no longer `IDisposable`.

I kinda like this one in appearance at least. It reminds me of what [ASP.NET `Program.cs` looked like](https://github.com/dotnet/aspnetcore/blob/v3.1.32/src/ProjectTemplates/Web.ProjectTemplates/content/WebApi-CSharp/Program.cs) prior to 5.0.

---

> This also raises concerns on mobile where `Run` is a non-blocking call, but cleanup doesn't happen there anyways.

Not sure why this would be an issue so long as `Run()` cleans up after itself.

> Since `GameHost` takes ownership of the game (as it'll dispose it) would it make sense to pass in the game in the constructor? That would clearly signify transfer of ownership, and also the game name can be deduced from the passed in `Game`, requiring no additional parameters. (could be taken from `Drawable.Name`, a new abstract property, or a ctor param: `protected Game(string gameName)`)

Careful with this; usages like [`CleanRunHeadlessGameHost`](https://github.com/ppy/osu/blob/master/osu.Game/Tests/CleanRunHeadlessGameHost.cs#L30) still need to be supported.

> ```cs
> var game = new SampleGameGame();
> using (GameHost host = Host.GetSuitableDesktopHost(game))
>     host.Run();
> ```

Don't hate this one either, barring that one point above about sometimes needing to request a host with a specific name for storage preparation.

PR #5640, opened 2023-02-04
Add a way to raise the window

Susko3:
Use case is raising the primary window when opening files/links via IPC (non-primary instance passing arguments to the primary), see https://github.com/ppy/osu/issues/22519.

Only implemented on desktop platforms. Raising a window is not a thing on mobile platforms, the OS has full control there (on Android we set `LaunchMode.SingleInstance` so every intent is passed to the same instance).

The tests are finicky on windows, as windows doesn't allow a window to raise/focus itself and will instead flash the taskbar button orange. If the window is minimised beforehand, there is a much better chance of the raise working.
For the use case described above, the raise always works. Windows probably does some heuristics to figure out if this is a "valid" raise request, and doing IPC like this is a common pattern.


peppy:
On macOS, this works unless the window is hidden. That's probably fine / the best we can do though. Can agree with leaving this mostly up to SDL's implementation.

peppy (approved):

PR #5641, opened 2023-02-04
Fix `TestSceneWindowed` crashing on mobile platforms

Susko3:


bdach:
I'm not sure what purpose doing this brings if doing literally *anything else* in that test scene will fail to work anyway. That test is *entirely* predicated on the existence of an SDL window. It will never make sense in any shape or form on mobile.

To me, the way to fix this is thus:

```diff
diff --git a/osu.Framework.Tests.Android/osu.Framework.Tests.Android.csproj b/osu.Framework.Tests.Android/osu.Framework.Tests.Android.csproj
index c888c1392..bfd2d9ef3 100644
--- a/osu.Framework.Tests.Android/osu.Framework.Tests.Android.csproj
+++ b/osu.Framework.Tests.Android/osu.Framework.Tests.Android.csproj
@@ -7,7 +7,7 @@
     <AssemblyName>osu.Framework.Tests.Android</AssemblyName>
   </PropertyGroup>
   <ItemGroup>
-    <Compile Include="..\osu.Framework.Tests\**\*.cs" Exclude="..\osu.Framework.Tests\Program.cs;..\osu.Framework.Tests\obj\**\*;..\osu.Framework.Tests\bin\**\*">
+    <Compile Include="..\osu.Framework.Tests\**\*.cs" Exclude="..\osu.Framework.Tests\Program.cs;..\osu.Framework.Tests\Visual\Platform\TestSceneWindowed.cs;..\osu.Framework.Tests\obj\**\*;..\osu.Framework.Tests\bin\**\*">
       <Link>%(RecursiveDir)%(Filename)%(Extension)</Link>
     </Compile>
     <EmbeddedResource Include="..\osu.Framework.Tests\Resources\**\*">
```

(applies to iOS project too, which goes without saying).

Susko3:
If we're to not show this test on mobile, I think using `(Un)SupportedOSPlatformAttribute` in some form is better than hiding it in msbuild properties.

Putting something like this on the class, and then doing a runtime check when [resolving valid tests](https://github.com/ppy/osu-framework/blob/9bbc6a81a2b8f035e2b043d3718d2db6e4f1d868/osu.Framework/Testing/TestBrowser.cs#L89).
```cs
    [SupportedOSPlatform("windows")]
    [SupportedOSPlatform("linux")]
    [SupportedOSPlatform("macos")]
    public partial class TestSceneWindowed : FrameworkTestScene { ... }
```

bdach:
Sure, that could also work, but also *is* more work. Not really fussed as to which way is used really. `SupportedOSPlatform` is also clunky to a degree in my eye as there is no way to tell it that it should apply to "desktop".

bdach:
Closing as superseded by https://github.com/ppy/osu-framework/pull/5643.

PR #5644, opened 2023-02-06
Improve path draw node's "cap" calculations and remove gaps

goodtrailer:
While using RenderDoc I noticed that slider caps have excess vertices/triangles:
before
In the screenshot above, the triangles from the `quadBatch` and `halfCircleBatch` (which misleadingly handles more than just half circles) are overlaid, with the `quadBatch` at lower alpha so that it's easier to see what I mean. The circular "caps" have more triangles than actually required as a result of:
* incorrect angle calculations
* duplicate path points

The "incorrect angle calculations" are demonstrated by the cap at the top. It's actually the opposite of the desired angle (so estimating, the desired angle is roughly ~60 degrees, but it's calculated as ~300 degrees). As a result, much of the cap is unnecessary and buried underneath the quads, illustrated in red:

image

The "duplicate path points" are illustrated by the start and end caps. osu! generates excess "duplicate" path points for sliders (i.e. the first/last several path points are identical in position), leading to 0 length segments. So there are extra caps produced with angle 0, leading to extra "vertical" caps, illustrated by the blue line (this is an issue with every slider; the start/end caps always have this vertical cap):

image

While the "duplicate path points" is equally as much an osu!-side issue, this isn't really desirable behavior either. This PR fixes cap generation by improving the angle calculations and ignoring segment lengths below some really small number (10^-5). Here is the result:

after

According to RenderDoc, the number of vertices drops from 1293 to 1095.

---

One other issue I happened to notice is in the way cap vertices and quad vertices are calculated. Because they are calculated in different methods via different techniques (cumulative angle addition vs. orthogonal lines), vertices that are intended to overlap may occasionally differ slightly. This leaves small spaces between edges that can cause fragment/pixel gaps during rasterization, which is *probably* the cause of ppy/osu#21896 (though this is going to need some testing, since the original discussion only pointed out a handful of examples). This gap is absolutely tiny though, which is probably why the issue appears so infrequently and depends on window size. For example, using the first RenderDoc capture from [this discussion reply](https://github.com/ppy/osu/discussions/21896#discussioncomment-4624675):

before-render

To illustrate the issue, here is an excerpt of the raw vertex position data for the vertices around the problematic area, a triangle cap between two quads. The left edge of the triangle cap should match the edge of one of the quads, while the right edge of the triangle cap should match the edge of the other quad. However, note how vertices 143 and 296 do not precisely match:
<details>
<summary>Vertex position data comparison</summary>

```
Vertex columns:
VTX, IDX, m_Position.x, m_Position.y, m_Position.z

Vertices from Colour Pass #2, glDrawElements(477) (aka halfCircleBatch):
141, 141, 169.26622,    45.24561,     1.00,
142, 142, 193.70415,    26.00381,     0.00,
143, 143, 193.21559,    25.39903,     0.00,

Vertices from Colour Pass #2, glDrawElements(696) (aka quadBatch):
294, 196, 169.26622,    45.24561,     1.00,
295, 197, 165.34532,    40.51416,     1.00,
296, 199, 193.21558,    25.39902,     0.00,

285, 190, 193.70415,    26.00381,     0.00,
286, 191, 197.44341,    30.75284,     0.00,
287, 189, 169.26622,    45.24561,     1.00,

---

Comparing the vertices of the left edge of the problematic triangle:

left edge: vertices 141->143 from halfCircleBatch
left edge: vertices 294->296 from quadBatch

141, 141, 169.26622,    45.24561,     1.00,
143, 143, 193.21559,    25.39903,     0.00,                <-- 193.21559

294, 196, 169.26622,    45.24561,     1.00,
296, 199, 193.21558,    25.39902,     0.00,                <-- 193.21558

vertices 141 and 294 do match
vertices 143 and 296 do NOT match                          <-- important bit !
thus the left edge does NOT overlap perfectly

---

Comparing the vertices of the right edge of the problematic triangle:

right edge: 141->142 from halfCircleBatch
right edge: 287->285 from quadBatch

141, 141, 169.26622,    45.24561,     1.00,
142, 142, 193.70415,    26.00381,     0.00,

287, 189, 169.26622,    45.24561,     1.00,
285, 190, 193.70415,    26.00381,     0.00,

vertices 141 and 287 do match
vertices 142 and 285 do match
thus the right edge overlaps perfectly
```

</details>

This PR fixes this by "snapping" the first and last vertices of every cap to ensure that cap and quad vertex positions match perfectly. Here is the result (note the lack of a white pixel):

after-render

However, since this "snapping" relies on alternating between cap and quad calculation for each path segment, only one vertex batch can be used. This slightly complicates the quad calculations, as they must be split into triangle calculations.


ojh050118:
fuck you

osu.Framework/Graphics/Lines/Path_DrawNode.cs
```diff
@@ -21,18 +20,18 @@ public partial class Path
         private class PathDrawNode : DrawNode
         {
             public const int MAX_RES = 24;
+            public const float MIN_SEGMENT_LENGTH = 1e-5f;
```
peppy, line 23: Wondering if these need to be publicly exposed. If they are, they probably should have xmldoc, or if there's no real reason we can make both `private`. There's no usages in other classes or in the osu! project.

peppy, line 23: Wondering if these need to be publicly exposed. If they are, they probably should have xmldoc, or if there's no real reason we can make both `private`. There's no usages in other classes or in the osu! project.

osu.Framework/Graphics/Lines/Path_DrawNode.cs
```diff
@@ -52,54 +51,161 @@ public override void ApplyState()
...
 
-            private void addLineCap(Vector2 origin, float theta, float thetaDiff, RectangleF texRect)
+            private void addSegmentQuads(Line segment, Line segmentLeft, Line segmentRight, RectangleF texRect)
             {
-                const float step = MathF.PI / MAX_RES;
+                Debug.Assert(triangleBatch != null);
 
-                float dir = Math.Sign(thetaDiff);
-                thetaDiff = dir * thetaDiff;
+                // Each segment of the slider is actually rendered as 2 quads, being split in half along the approximating line.
```
peppy, line 66:
```suggestion
                // Each segment of the path is actually rendered as 2 quads, being split in half along the approximating line.
```

osu.Framework/Graphics/Lines/Path_DrawNode.cs
```diff
@@ -52,54 +51,161 @@ public override void ApplyState()
...
+                // On this line the depth is 1 instead of 0, which is done properly handle self-overlap using the depth buffer.
```
peppy, line 67:
```suggestion
                // On this line the depth is 1 instead of 0, which is done in order to properly handle self-overlap using the depth buffer.
```

osu.Framework/Graphics/Lines/Path_DrawNode.cs
```diff
@@ -108,100 +214,72 @@ private void addLineCap(Vector2 origin, float theta, float thetaDiff, RectangleF
...
-                    addLineQuads(segment, texRect);
+                        addSegmentCaps(flippedLeft, flippedRight, segmentLeft, segmentRight, texRect);
+                    }
+
+                    prevSegmentLeft = segmentLeft;
+                    prevSegmentRight = segmentRight;
+                }
             }
 
             public override void Draw(IRenderer renderer)
```
peppy, line 272: We generally keep `private` methods beneath `public` ones, so please move this above your methods above (but below the `ApplyState`).

peppy (changes requested):
Just code quality changes. The overall change seems quite sound.

peppy:
I'm seeing a crash when using this branch:

```csharp

[runtime] 2023-02-07 05:59:52 [error]: System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
[runtime] 2023-02-07 05:59:52 [error]: at System.Collections.Generic.List`1.get_Item(Int32 index)
[runtime] 2023-02-07 05:59:52 [error]: at osu.Framework.Graphics.Lines.Path.PathDrawNode.updateVertexBuffer() in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Lines/Path_DrawNode.cs:line 232
[runtime] 2023-02-07 05:59:52 [error]: at osu.Framework.Graphics.Lines.Path.PathDrawNode.Draw(IRenderer renderer) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Lines/Path_DrawNode.cs:line 294
[runtime] 2023-02-07 05:59:52 [error]: at osu.Framework.Graphics.BufferedDrawNode.Draw(IRenderer renderer) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/BufferedDrawNode.cs:line 105
[runtime] 2023-02-07 05:59:52 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.CompositeDrawableDrawNode.Draw(IRenderer renderer) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable_DrawNode.cs:line 204
```

A few seconds into https://osu.ppy.sh/beatmapsets/1676686#osu/3465105.

goodtrailer:
I was confused for a bit, because none of the diffs of that mapset were crashing for me. Then I realized I only tested the code the snaking slider options disabled. That was quite the oversight, since snaking sliders causes there to (initially) be a single very small segment, small enough that it's skipped and there are no segments to draw, leading to out of bounds indices. Maybe this deserves a test case?

osu.Framework/Graphics/Lines/Path_DrawNode.cs
goodtrailer: Actually, the draw node is itself a `private` nested class, so it doesn't even make sense to have these `public`, since they'd be inaccessible anyway.

bdach:
The more tests the better, that is for sure.

peppy (approved):
Looks and works well for me!

bdach:
I've read through this. There's good work done in here, but a few gripes.

While I appreciate the detailed description, to me it was not detailed enough. I would have personally expanded it with details about:

- why the original code was wrong,
- added info about the angle coordinate system pretty much completely changing between old and new versions,
- there are a few clever tricks used that it would have been nice to get a straight-up explanation of (like the way start/end caps are drawn by essentially going back-and-forth over the first and last segment).

As for the code itself, I wish there were less method moves, as the full diff is pretty much unreadable. Commit by commit was better but still subpar.

All of the above notwithstanding, I also have an actual issue. If I go to `TestScenePathWithZeroLength` and crank up the path radius, like so:

```diff
diff --git a/osu.Framework.Tests/Visual/Drawables/TestScenePathWithZeroLength.cs b/osu.Framework.Tests/Visual/Drawables/TestScenePathWithZeroLength.cs
index 032c817e8..f6c4e87b4 100644
--- a/osu.Framework.Tests/Visual/Drawables/TestScenePathWithZeroLength.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestScenePathWithZeroLength.cs
@@ -77,7 +77,7 @@ public TestScenePathWithZeroLength()
 
         private static Path createPath(Vector2[] points) => new Path
         {
-            PathRadius = 2,
+            PathRadius = 50,
             Vertices = points,
         };
     }
```

I get this:

image

On comparing with `master` by cherry-picking the test out onto it, the old code would fare much better with this particular edge case, so it is a regression.

bdach (changes requested):
see above issue with the start caps

bdach:
Upon closer inspection, it is the new cap drawing method combined with floating point trickery that is causing the aforementioned issue.

This is a crude debuggy view of how the tris are drawn in the failing case:

image

In the debugger, `theta0` is wrongly -pi/2:

image

This in turn is because floats are floats and not all floating points were created equal:

image

Only the `Y` coordinate behaves as specified in the test; the `X` coordinate doesn't because `135.355347f + 0.000001f == 135.355347f`. Previously the code was resilient to this because cap vertices were calculated entirely separately from the segments themselves; with this change this is no longer the case.

A crude way to "fix" this is to increase `min_segment_length` by an order of magnitude. Unsure if that's considered too hacky, however. Maybe it's better to draw the start/end caps separately as a special case, rather than resort to the zigzag method?

goodtrailer:
> As for the code itself, I wish there were less method moves, as the full diff is pretty much unreadable. Commit by commit was better but still subpar.

Yeah, this was a mistake. I also think I might've included an erroneous commit. I tested the changes with a single commit on a fresh branch (and moved some methods around to match the original ordering), and the diff is much more readable, so I'll have to close this PR and re-open it with a fresh one later, after addressing the issues at hand.

> * added info about the angle coordinate system pretty much completely changing between old and new versions in the cap-drawing method,
> * there are a few clever tricks used that it would have been nice to get a straight-up explanation of (like the way start/end caps are drawn by essentially going back-and-forth over the first and last segment).

Good point. I'll have to add some comments about what "left" and "right" mean in the code, and also how the start/end cap semi-circles work.

> A crude way to "fix" this is to increase min_segment_length by an order of magnitude. Unsure if that's considered too hacky, however. Maybe it's better to draw the start/end caps separately as a special case, rather than resort to the zigzag method?

Unfortunately, it seems this bug isn't exclusive to start/end caps. I created the following test case based off the buggy one:
```csharp
Cell(1).AddRange(new[]
{
    createLabel("Middle few tiny segments"),
    createPath(new[]
    {
        new Vector2(50f, 100f),
        new Vector2(100f),
        new Vector2(100f + 1e-6f),
        new Vector2(100f + 2e-6f),
        new Vector2(100f + 3e-6f),
        new Vector2(100f + 4e-6f),
        new Vector2(200f),
    }),
});
```
It suffers the same issue, except at a middle "joint" cap. Bumping up the threshold by an order of magnitude to 1e-4 works, but only really up to vertex positions hovering around ≈±2000, since relative epsilon of single-precision floats is 2^-24 (and 2^-24 * 2048 = 2^-13 ≈ 1.22e-4). Assuming vertex positions never really go beyond ≈2000 (which is true for osu! at least), then 1e-4 is fine. But this feels a bit backwards to me, since the logic backing osu!framework would be based off the conventions of osu!.

Perhaps the PR should simply do away with the minimum threshold segment (and only contain the obtuse angle and fragment gap stuff)? Having the extra caps doesn't affect anything visually. While generating excess geometry is still undesirable behavioiur, the responsibility would be on dependent code (i.e. osu!) to not provide excess vertices in the first place.

bdach:
> Another solution is to use a relative minimum length, where segment lengths must be greater than max(1e-5, 1e-6 * vertex position), though 1e-6 here is kinda arbitrary (I tried 1e-7 and it didn't work, despite 2^-24 ≈ 6e-8, so idk). Not sure how I feel about this one.

This was my immediate thought. If it works then maybe worth a go, but then it still is a fair bit of arbitrary thresholding

> Perhaps the PR should simply do away with the minimum threshold segment (and only contain the obtuse angle and fragment gap stuff)? Having the extra caps doesn't affect anything visually

I think I weakly prefer this, this PR is a net gain in number of verts rendered anyway regardless of that - but probably wise to get a second opinion on this (@peppy?)

peppy:
I'd also prefer the second proposal

goodtrailer:
Removed the minimum length threshold, and the vertex count for the slider mentioned in the original post went up to 1200. So yeah, still an improvement, but less so.

I've decided to not close this PR yet, so that the commit-by-commit diff is there. If the changes are good, then I'll reopen this PR on a fresh branch as a single condensed commit (since as mentioned earlier, the total diff is pretty mangled by method re-positioning and erroneous commits).

bdach:
> If the changes are good, then I'll reopen this PR on a fresh branch as a single condensed commit (since as mentioned earlier, the total diff is pretty mangled by method re-positioning and erroneous commits).

At this point that's not necessary. We don't much care about sparkly clean history, and at this point we have enough reviewer eyes that a third should not be required, so I think we'll manage now that me and @peppy have seen the code previously before.

bdach (approved):
Pretty happy with this now. Good work :+1:

peppy (approved):

PR #5646, opened 2023-02-07
Add proper async lookup flow to `TextureStore`

peppy:
Addresses some of the remaining concerns brought to light in https://github.com/ppy/osu/issues/22446 (in conjunction with consumer `Drawable`s being updated to use async BDL flow, see [branch](https://github.com/ppy/osu/compare/master...peppy:osu:online-texture-load-use-getasync?expand=1)).

One shortcoming of this current implementation is that underlying lookups will never be cancelled, even if all dependant lookups are. I personally don't see this as an issue since that was never possible previously, but it may be something we want to consider in the future. May be helpful for scenarios like fast scrolling on beatmap listing where we may be able to avoid some web requests by early cancelling?

### DIFF
--- a/osu.Framework.Tests/Visual/Sprites/TestSceneTextures.cs
+++ b/osu.Framework.Tests/Visual/Sprites/TestSceneTextures.cs
@@ -117,6 +117,33 @@ namespace osu.Framework.Tests.Visual.Sprites
         }
 
         /// <summary>
+        /// Tests the case where multiple lookups occur which overlap each other, for the same texture.
+        /// Then one of the lookups is cancelled.
+        /// The underlying lookup should continue to completion.
+        /// </summary>
+        [Test]
+        public void TestFetchContentionSameLookupCancellation()
+        {
+            Avatar avatar1 = null;
+            Avatar avatar2 = null;
+
+            AddStep("begin blocking load", () => spriteContainer.BlockingOnlineStore.StartBlocking());
+            AddStep("get first", () => avatar1 = addSprite("1"));
+            AddStep("get second", () => avatar2 = addSprite("1"));
+
+            AddAssert("neither are loaded", () => avatar1.Texture == null && avatar2.Texture == null);
+
+            AddStep("Cancel first load", () => spriteContainer.OfType<DelayedLoadWrapper>().First().RemoveAndDisposeImmediately());
+            AddUntilStep("wait for disposed", () => avatar1.IsDisposed);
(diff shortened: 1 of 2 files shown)


peppy:
Blocking for now as this doesn't seem to work as expected (async lookups do not block the BDL load process correctly). Example when testing against results screen on osu! on [this branch](https://github.com/ppy/osu/compare/master...peppy:osu:online-texture-load-use-getasync?expand=1):

```csharp

[runtime] 2023-02-07 05:42:53 [error]: An unobserved error has occurred.
[runtime] 2023-02-07 05:42:53 [error]: osu.Framework.Graphics.Drawable+InvalidThreadForMutationException: Cannot mutate the InternalChildren on a Loaded Drawable while not on the update thread. Consider using Schedule to schedule the mutation operation.
[runtime] 2023-02-07 05:42:53 [error]: at osu.Framework.Graphics.Drawable.EnsureMutationAllowed(String action) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Drawable.cs:line 2618
[runtime] 2023-02-07 05:42:53 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.EnsureChildMutationAllowed() in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 1378
[runtime] 2023-02-07 05:42:53 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.ClearInternal(Boolean disposeChildren) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 536
[runtime] 2023-02-07 05:42:53 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.set_InternalChild(Drawable value) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 372
[runtime] 2023-02-07 05:42:53 [error]: at osu.Game.Users.UserCoverBackground.Cover.load(LargeTextureStore textures, CancellationToken cancellationToken) in /Users/dean/Projects/osu/osu.Game/Users/UserCoverBackground.cs:line 65

```

smoogipoo:
Wanna test if this behaves better with https://github.com/smoogipoo/osu-framework/tree/bdl-async ?

smoogipoo:
Oh right, I forgot to mention that you'll need a local source generator package.

PR #5648, opened 2023-02-08
Implement `ColourInfo.Interpolate(Quad)` function

EVAST9919:
Used in 2 places osu-side already and seems like it belongs in the framework.
Naming can most likely be improved.


osu.Framework/Graphics/Colour/ColourInfo.cs
```diff
@@ -132,6 +132,21 @@ public void ApplyChild(ColourInfo childColour, Quad interp)
             BottomRight = newBottomRight;
         }
 
+        public readonly ColourInfo GetPartialQuad(Quad quad)
```
frenzibyte, line 135: `Quad` is a misleading term in the name since this is returning a `ColourInfo`. I would say at the very least use `GetPartial`.

bdach: I would just overload `Interpolate()` here and have this be the `Interpolate(Quad)` overload. It is what this method is doing; it's doing bilinear interpolation but over a quad (sub? super?)region.

bdach (approved):

peppy (approved):

PR #5650, opened 2023-02-11
Show game name in reason of sleep inhibition

easyteacher:
With https://github.com/libsdl-org/SDL/pull/4704 it's possible to provide the game name to DBus. This allows the battery widget in Plasma show the game name instead of "My SDL application".

|Before|After|
|-|-|
|图片|图片|

### DIFF
--- a/osu.Framework/Platform/Linux/LinuxGameHost.cs
+++ b/osu.Framework/Platform/Linux/LinuxGameHost.cs
@@ -32,6 +32,7 @@ namespace osu.Framework.Platform.Linux
         protected override void SetupForRun()
         {
             SDL.SDL_SetHint(SDL.SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, BypassCompositor ? "1" : "0");
+            SDL.SDL_SetHint(SDL.SDL_APP_NAME, Name);
             base.SetupForRun();
         }
 


osu.Framework/Platform/Linux/LinuxGameHost.cs
```diff
@@ -32,6 +32,7 @@ internal LinuxGameHost(string gameName, HostOptions options)
         protected override void SetupForRun()
         {
             SDL.SDL_SetHint(SDL.SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, BypassCompositor ? "1" : "0");
+            SDL.SDL_SetHint(SDL.SDL_APP_NAME, Name);
```
bdach, line 35: First of all, this doesn't even resolve to a valid member. You want

```suggestion
            SDL.SDL_SetHint(SDL.SDL_HINT_APP_NAME, Name);
```

Secondly, I'm not overly keen on the placement of this. While yes, it appears that right now this hint is only ever used in linux pathways on the SDL side, the name and the [docs](https://github.com/libsdl-org/SDL/blob/7285f0d441330386758f926aeb8b5be0f4486e4e/include/SDL_hints.h#L148-L166) of it do not carry any linuxy connotations. There's nothing there about dbus, or x11, or wayland, or whatever else. So to safeguard against upstream using this in other platform-agnostic contexts I'd see this set in a more central place, so that all platforms set this hint.

bdach (changes requested):

osu.Framework/Platform/Linux/LinuxGameHost.cs
easyteacher: Are you sure SDL_HINT_APP_NAME works? SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR also doesn't include "HINT".

easyteacher: Oh never mind, it's defined [here](https://github.com/flibitijibibo/SDL2-CS/blob/f8c6fc407fbb22072fdafcda918aec52b2102519/src/SDL2.cs#L764).

easyteacher: I don't see SDL being used in WindowsGameHost or MacOSGameHost, as I am not familiar with the code, only changing LinuxGameHost seems safer to me.

bdach: It would have to be somewhere in `SDL2DesktopWindow` itself, probably.

It's actually funny that `LinuxGameHost`'s SDL hint sets even work, because as far as I can tell they happen before `SDL_Init()` is ever actually called.

easyteacher: > they happen before SDL_Init() is ever actually called

I do not understand how it influences hints, as far as I can tell SDL_APP_NAME is not set by default, and SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR works as expected.

easyteacher: Name is defined in GameHost, is it possible to get the game name in SDL2DesktopWindow ? Or should I import SDL2 and set hints in GameHost?

bdach: I'm honestly not sure what would be the best way. Thinking that maybe the correct thing to do there is set this hint whenever the window `Title` is being set. Would probably want a few opinions on that (@Susko3? @peppy?)

Susko3: I think that using the title of the window makes more sense. The name passed to the host is something internal to the game, not really meant for users to see. The window title is already user-visible, and will include the exclamation mark in `osu!`.

peppy: Title also makes sense to me.

bdach: Hoping to knock this one off the list, I tried setting this hint along with the title. Unfortunately it doesn't work. It seems that if the hint is set too late, SDL won't pass it to dbus/pulseaudio/pipewire/whatever else, ergo, it's timing-dependent. It must be set before `SDL2DesktopWindow` is created. Love it.

I'm not sure I care enough to think about this problem too much more.

PR #5647, opened 2023-02-07
Implement remaining parts for Veldrid support

smoogipoo:
This is a WIP branch that supersedes https://github.com/ppy/osu-framework/pull/5596 https://github.com/ppy/osu-framework/pull/5597 https://github.com/ppy/osu-framework/pull/5598 https://github.com/ppy/osu-framework/pull/5599

PR #5653, opened 2023-02-16
Fix veldrid pipeline caching not working properly

frenzibyte:
Noticed in #5647. Pipeline descriptions contain array fields, which are stored by reference. Modifying those arrays affect stored lookup keys in the cache dictionary, therefore cloning for the arrays are required before storing the description as a lookup key.

This resolves the first issue mentioned in https://github.com/ppy/osu-framework/pull/5647#issuecomment-1421562057.

### DIFF
--- a/osu.Framework/Graphics/Veldrid/VeldridExtensions.cs
+++ b/osu.Framework/Graphics/Veldrid/VeldridExtensions.cs
@@ -282,6 +282,22 @@ namespace osu.Framework.Graphics.Veldrid
             }
         }
 
+        public static GraphicsPipelineDescription Clone(this GraphicsPipelineDescription pipeline)
+        {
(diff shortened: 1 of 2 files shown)


peppy:
Is this called often? Looks quite expensive..

smoogipoo (approved):

smoogipoo:
The hope is that most times we'll get a cache hit, since the pipeline only changes when some state changes like blending/shaders/uniform layouts, and we'd expect the total number of combinations of those to be exhausted at some point.

That being said, I've done no profiling of `VeldridRenderer` and we'll need to do a separate pass on that in general.

peppy:
Right, didn't realise this was used for caching and not just equality debouncing.

PR #5654, opened 2023-02-21
Prefer the SDL2 Wayland backend over X11 on Linux

timschumi:
Using the SDL2 backend for X11 with XWayland results in some [issues with handling external tablet input](https://opentabletdriver.net/Wiki/FAQ/Linux).

SDL [reverted their "Prefer Wayland over X11" change due to certain issues](https://github.com/libsdl-org/SDL/commit/254fcc90eb22bb159ab365ad956222a9c5632841) and scheduled it for SDL3 instead, but considering that osu-framework has a much smaller scope, let's try to change the preferred order for ourselves.

Fixes ppy/osu#21939

### DIFF
--- a/osu.Framework/Platform/Linux/LinuxGameHost.cs
+++ b/osu.Framework/Platform/Linux/LinuxGameHost.cs
@@ -32,6 +32,7 @@ namespace osu.Framework.Platform.Linux
         protected override void SetupForRun()
         {
             SDL.SDL_SetHint(SDL.SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, BypassCompositor ? "1" : "0");
+            SDL.SDL_SetHint("SDL_VIDEODRIVER", "wayland,x11");
             base.SetupForRun();
         }
 


osu.Framework/Platform/Linux/LinuxGameHost.cs
```diff
@@ -32,6 +32,7 @@ internal LinuxGameHost(string gameName, HostOptions options)
         protected override void SetupForRun()
         {
             SDL.SDL_SetHint(SDL.SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, BypassCompositor ? "1" : "0");
+            SDL.SDL_SetHint("SDL_VIDEODRIVER", "wayland,x11");
```
timschumi, line 35: This should probably use a string constant for the hint name, but that depends on flibitijibibo/SDL2-CS#243 being merged and pulling that into ppy/SDL2-CS (or we just cherry-pick that commit manually, independent of the upstream).

peppy:
I'm pretty sure this was proposed previously and rejected due to breaking stuff?

smoogipoo:
It runs for me using X11, but I definitely want a second opinion as I'm generally against pushing Wayland in its current state/know that it breaks certain workflows such as screen capture and potentially mouse acceleration.

bdach:
I've already responded to a discussion suggesting this recently: https://github.com/ppy/osu/discussions/22651#discussioncomment-4985552

tl;dr: I have no reason to believe that preferring Wayland is universally better than not doing so, as both having it off and on causes issues on various setups, and would prefer to defer to upstream when it comes to change of default.

smoogipoo:
I agree, sounds good.

timschumi:
> I'm generally against pushing Wayland to be the default in its current state because I know that it breaks certain workflows such as screen capture and potentially mouse acceleration.

Screen capture is, at least to my knowledge, a general Wayland issue that affects both native Wayland and XWayland similarly.

Mouse acceleration would indeed be something to investigate. I assume that you are talking about mouse position updates that are reported by the compositor?

> I have no reason to believe that preferring Wayland is universally better than not doing so, as both having it off and on causes issues on various setups, and would prefer to defer to upstream when it comes to change of default.

I don't plan to claim that it's universally better. I do however believe that, considering the circumstances and upstream progress, it is a reasonable idea to try and explore if we can make Wayland the default.

In any case, we should probably at least start tracking actual Wayland issues (with the Wayland backend, not the X11 one) in some central place (or start tagging them with something like `linux-x11`, `linux-xwayland`, and `linux-wayland`).

I don't expect an actual upstream defaults switch to ever happen unless osu-framework switches to SDL3, but considering that they have scheduled that change at all makes me believe that they are pretty confident in that most of the issues have been resolved. The only apparent leftover breakage from the upstream SDL issue is that the Steam overlay doesn't support Wayland yet. I'm currently flipping through the SDL3 commit history to figure out whether there have been any Wayland fixes that haven't yet landed in SDL2. If there aren't any, would that count as a "let's wait on upstream to switch the default"?

Most of the Wayland-related issues that I have found in osu or osu-framework unfortunately don't make any mention of whether they are using the actual Wayland backend or the X11 backend via XWayland. However, since x11 has been the default for most of the timespan where those issues were reported, those appear to be mostly issues with the x11 backend in combination with XWayland.

The few issues that actually seem to be (at least at some point in time) valid Wayland backend issues are the following ones:

- ppy/osu#15625
  - Claims that `SDL_VIDEODRIVER=x11` fixed the issue, although that should have been the default at this point in time
  - Might be related to an upstream issue that is marked as fixed, and a similar issue I have found somewhere else mentioned that Nvidia driver updates made it work
- ppy/osu#1685
 - This was filed during the short time where `SDL_VIDEODRIVER=wayland` was the default (do we build non-tagged releases?)

Of course, I'm happy to extend that list or move it somewhere else in case you know of any Wayland backend issues, the majority of search results for "Wayland" is unfortunately only things like "Are you using Wayland?".

> So I'm against this change personally, users should adjust themselves locally as required for now.

I'm aware that this wouldn't exactly be the most gentle change (breaking existing setups, etc.), but that option technically also exists for Wayland users that would like to force usage of XWayland.

bdach:
I'm personally at the point wherein I would rather spend time on migrating to SDL3, which we'll likely have to do anyway at some point, and which has potential to be a net gain on all platforms, rather than do a bunch of extra tracking work specifically for linux - which, like it or not, is a minority when it comes to user base.

If upstream decides that they have backported enough fixes from SDL3 to SDL2 to warrant switching backends again in SDL2 - fine, we can take that too. But I strongly believe they will have a far better view on the situation than we do.

timschumi:
> I'm personally at the point wherein I would rather spend time on migrating to SDL3, which we'll likely have to do anyway at some point

That - of course - is also an option, I just assumed that it would be even less accepted than additional tracking since SDL3 turns out to be a major overhaul API-wise (well, who would've guessed).

Well, I guess I'll start looking into it once SDL3 gets its first release, assuming that nobody else got to it by then.

PR #5658, opened 2023-02-22
Improve menu handling on disabled items (or items with no action assigned)

peppy:
Until now, clicking a disabled menu item (or item with no action) would have still closed the menu. Also, the hover state would make it look like the menu items are interactive. I noticed this when coming across our "menu separator" implementation osu! side (and wanted to change it to not show a hover background state).

This behaviour does not match common UX expected from operating systems. Both macOS and windows will not show a hover state and not dismiss menus on clicking disabled items.

I've updated our handling to match expectations.

Originally I was going to apply changes only osu!-side but it turns out the hover foreground/background state handling is quite low level, so fixing here made sense.


peppy:
See usage at https://github.com/ppy/osu/pull/22716. It may be that we want to combine the `IsActionable` and `HasSubmenu` into a single flag.

osu.Framework/Graphics/UserInterface/Menu.cs
```diff
@@ -817,20 +816,25 @@ public MenuItemState State
...
+            /// </summary>
+            protected bool IsActionable => !Item.Action.Disabled && Item.Action.Value != null;
+
             /// <summary>
             /// Called after the <see cref="BackgroundColour"/> is modified or the hover state changes.
             /// </summary>
             protected virtual void UpdateBackgroundColour()
             {
-                Background.FadeColour(IsHovered ? BackgroundColourHover : BackgroundColour);
+                Background.FadeColour(IsHovered && (IsActionable || HasSubmenu) ? BackgroundColourHover : BackgroundColour);
```
bdach, line 829: Given that `IsActionable` influences the visual state of the menu item now, any change of `Action.Value` and `Action.Disabled` should call `Update{Background,Foreground}Colour()`.

Test case that exhibits (mostly visually, unfortunately) what I mean:

```diff
diff --git a/osu.Framework.Tests/Visual/UserInterface/TestSceneClosableMenu.cs b/osu.Framework.Tests/Visual/UserInterface/TestSceneClosableMenu.cs
index 8fa3133e3..e89a14365 100644
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneClosableMenu.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneClosableMenu.cs
@@ -4,6 +4,7 @@
 #nullable disable
 
 using System;
+using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.UserInterface;
@@ -62,6 +63,25 @@ public void TestClickItemWithActionClosesMenus()
             });
         }
 
+        [Test]
+        public void TestClickItemWithActionAssignedDuringNavigationClosesMenus()
+        {
+            AddStep("click item", () => ClickItem(0, 0));
+            AddStep("hover item", () => InputManager.MoveMouseTo(Menus.GetSubStructure(1).GetMenuItems()[0]));
+            AddStep("assign action", () => Menus.GetSubStructure(1).GetMenuItems().Cast<Menu.DrawableMenuItem>().First().Item.Action.Value = () => { });
+            AddStep("click item", () => InputManager.Click(MouseButton.Left));
+            AddAssert("all menus closed", () =>
+            {
+                for (int i = 1; i >= 0; --i)
+                {
+                    if (Menus.GetSubMenu(i).State == MenuState.Open)
+                        return false;
+                }
+
+                return true;
+            });
+        }
+
         [Test]
         public void TestClickItemWithoutActionDoesNotCloseMenus()
         {
```

Menu item will appear as disabled, but close the menu when clicked (because the visual state doesn't update when `Action` is set between hover and click).

Submenus are immutable so thankfully that is not a problem.

bdach (changes requested):

peppy:
Should be good for re-review

PR #5659, opened 2023-02-23
Add support for uniform buffers and use for the global uniforms

smoogipoo:
Prereqs:
- [ ] https://github.com/ppy/osu-framework/pull/5655

## Abstract

Veldrid, likely due to other renderers like Metal, only supports uniform buffers. Plain old uniforms are not supported.

This adds first class support for uniform buffers.

Additionally, this also adds a small `GLESRenderer` to be used on mobile platforms, although only Android works for now.

## Usage

Shader definition:
```glsl
UNIFORM_BLOCK(0, g_MyBuffer)
{
    float value1;
    vec4 value2;
    float value3;
}
```

DrawNode usage:
```csharp
class MyDrawNode : DrawNode
{
    private IUniformBuffer<MyBufferData>? bufferData;

    {
        base.Draw(renderer);

        // Create the uniform buffer.
        bufferData ??= renderer.CreateUniformBuffer<MyBufferData>();

        // Update the uniform buffer.
        bufferData.Data = bufferData.Data with
        {
            Value1 = 1.0f,
            Value2 = Vector4.One,
            Value3 = 1.0f
        };

        // Bind the buffer to the shader.
        shader.BindUniformBuffer("g_MyBuffer", bufferData);

        shader.Bind();
        // Draw...
    }

    {
        base.Dispose(isDisposing);

        // Ensure that the buffer is disposed.
        bufferData?.Dispose();
    }
}

// Pack = 1 is required.
[StructLayout(LayoutKind.Sequential, Pack = 1)]
// Record structs automatically implement IEquatable<>
public record struct MyBufferData
{
    // Use types provided in the osu.Framework.Graphics.Shaders.Types namespace.
    public UniformFloat Value1;

    // Each type lists the boundary it needs to be aligned to.
    // In this case, the next value is a UniformVector4 which specifies that it needs to be aligned to a 16-byte boundary.
    // The UniformPadding4, UniformPadding8, and UniformPadding12 types are provided to assist with correctly aligning uniform data.
    private readonly UniformPadding12 pad1;

    public UniformVector4 Value2;
    public UniformFloat Value3;

    // The struct must be a multiple of 16 bytes.
    private readonly UniformPadding12 pad2;
}
```

## Future improvements

I'm not entirely sold on the packing method. It seems there's no built-in construct in C# that allows us to do the packing required by `layout(std140)` (the only sane layout), so it needs to be crafted by hand. I've added some simple runtime validation for this via reflection but perhaps it can be done better (analyser/source-gen?) in the future.


smoogipoo:
Combined into https://github.com/ppy/osu-framework/pull/5655

PR #5660, opened 2023-02-26
Add BindableColour4

Poyo-SSB:
For osu!-side config stuff.

Serialization uses hex codes to make it easier for users to edit. This sacrifices precision but I assume that's okay given that `BindableFloat`/`BindableDouble` serialize with a limited number of significant figures.

### DIFF
--- a/osu.Framework/Bindables/BindableColour4.cs
+++ b/osu.Framework/Bindables/BindableColour4.cs
@@ -18,10 +18,19 @@ namespace osu.Framework.Bindables
 
         public override void Parse(object input)
         {
-            if (input is string hex && Colour4.TryParseHex(hex, out Colour4 colour))
-                Value = colour;
-            else
-                base.Parse(input);
+            switch (input)
+            {
+                case string str:
+                    if (!Colour4.TryParseHex(str, out Colour4 colour))
+                        throw new ArgumentException($"Input string was in wrong format! (expected valid hex colour, actual: '{str}')");
+
+                    Value = colour;
+                    break;
+
+                default:
+                    base.Parse(input);
+                    break;
+            }
         }
 
         protected override Bindable<Colour4> CreateInstance() => new BindableColour4();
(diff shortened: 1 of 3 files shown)


peppy:
Could you please provide an example of where this is to be used?

peppy:
This has no test coverage. Please see every other bindable class which has corresponding tests (ie. https://github.com/ppy/osu-framework/blob/ba1385330cc501f34937e08257e586c84e35d772/osu.Framework.Tests/Bindables/BindableDoubleTest.cs#L12-L11)

osu.Framework/Bindables/BindableColour4.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    public class BindableColour4 : Bindable<Colour4>
+    {
+        public BindableColour4(Colour4 value = default)
+            : base(value)
+        {
+        }
+
+        // 8-bit precision should probably be enough for serialization.
+        public override string ToString(string format, IFormatProvider formatProvider) => Value.ToHex();
```
peppy, line 17: Pending more details of the use case, I'd want to reconsider rgb byte values since that is what osu! has historically used.

peppy (commented):
Needs tests

Poyo-SSB:
> Could you please provide an example of where this is to be used?

Primary use case is with `IniConfigManager`, where a `Bindable<Colour4>` will fail to be parsed. Very useful for non-osu! o!f projects.

> I'd want to reconsider rgb byte values since that is what osu! has historically used.

Are RGB byte values used anywhere outside of legacy cases (`skin.ini`/legacy beatmap format)? I'd argue that hex codes are a more widely recognized form of textual color representation. Alternatively I could implement parsing for the RGB byte triples while still serializing to hex by default. Or we could just go with the triples both in and out.

peppy:
> Are RGB byte values used anywhere outside of legacy cases (`skin.ini`/legacy beatmap format)? I'd argue that hex codes are a more widely recognized form of textual color representation. Alternatively I could implement parsing for the RGB byte triples while still serializing to hex by default. Or we could just go with the triples both in and out.

Completely agree. That's why I wanted to discern what the usage of this was. If it's a general change which is not intended to interact with legacy skins (as you've confirmed) then storing as hex is likely the best path.

peppy (approved):

PR #5664, opened 2023-03-03
Fix dropdown menu not scrolling correctly when hovering first/last visible item

Joehuu:
- Fixes https://github.com/ppy/osu/issues/9442

When an item that is out of view gets hovered, `ScrollIntoView` will override the user scroll.

Before:
h3BGy8jh0T

After:
JjkSE7Ltyc

### DIFF
--- a/osu.Framework/Graphics/UserInterface/Dropdown.cs
+++ b/osu.Framework/Graphics/UserInterface/Dropdown.cs
@@ -410,12 +410,13 @@ namespace osu.Framework.Graphics.UserInterface
             /// Preselects an item from this <see cref="DropdownMenu"/>.
             /// </summary>
             /// <param name="item">The item to select.</param>
-            protected void PreselectItem(MenuItem item)
+            /// <param name="scrollToItem">Whether to scroll to the preselected item.</param>
+            protected void PreselectItem(MenuItem item, bool scrollToItem = true)
             {
                 Children.OfType<DrawableDropdownMenuItem>().ForEach(c =>
                 {
                     c.IsPreSelected = compareItemEquality(item, c.Item);
-                    if (c.IsPreSelected)
+                    if (c.IsPreSelected && scrollToItem)
                         ContentContainer.ScrollIntoView(c);
                 });
             }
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/UserInterface/Dropdown.cs
```diff
@@ -521,7 +522,7 @@ protected override void UpdateForegroundColour()
 
                 protected override bool OnHover(HoverEvent e)
                 {
-                    PreselectionRequested?.Invoke(Item as DropdownMenuItem<T>);
+                    PreselectionRequested?.Invoke(Item as DropdownMenuItem<T>, false);
```
bdach, line 525: The thing with this is that it slightly alters behaviour in the case wherein a dropdown item is partially scrolled in.

Before:

https://user-images.githubusercontent.com/20418176/223222143-5b84f615-24db-4b75-958b-8f53a1e5f083.mp4

After:

https://user-images.githubusercontent.com/20418176/223222164-22ae33bf-e4f3-423f-ac12-d68ba6fed408.mp4

I'm honestly not sure if that's a good or a bad thing, just something I figure worth pointing out I guess? Looking at a few places (windows explorer, firefox `<select>` dropdowns) they do the same thing wherein they will scroll in fully a partially-visible dropdown item if it is hovered.

Joehuu: Looking at those places + other common apps (discord), I notice that they don't seem to preselect `OnHover()` but rather `OnMouseMove()`. Using that still fixes what this PR is aiming for while still keeping that behavior you described.

But that really changes the UX (responsiveness) that users expect from now (playing lazer / stable):
osu Framework Tests_kTFeM1htsN

Maybe that's for the best? This change also fixes what I was going to look after this PR below.

---

Using `OnMouseMove()` fixes another bug involving keyboard input:
Before:
osu Framework Tests_uRWL8wfdKv

After:
osu Framework Tests_isIayvz07g

bdach: Yeah I don't know about the responsiveness change. Feels like we'd still want to have preselection take effect on hover without mouse movement.

Can hover + mouse move be combined maybe?

peppy: java: on move
windows: on move
macOS: on hover

I'd personally be fine with changing the behaviour to on move. I think it feels less jank (and is what my brain expects, even after being primarily on macOS for so many years).

PR #5666, opened 2023-03-04
Fix external links not working on Android

Joehuu:
- Framework side of https://github.com/ppy/osu/pull/22790

### DIFF
--- a/SampleGame.Android/AndroidManifest.xml
+++ b/SampleGame.Android/AndroidManifest.xml
@@ -8,5 +8,10 @@
 			<category android:name="android.intent.category.BROWSABLE" />
 			<data android:scheme="https" />
 		</intent>
+		<intent>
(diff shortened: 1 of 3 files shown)


bdach:
Not sure how everybody else is feeling but for this one I tend to think we don't need to do the [manifest merging thing we did in the game project](https://github.com/ppy/osu/pull/22790#issuecomment-1454962981) and we can probably go with how things are right now. Rationale being that:
- there is much less duplication here,
- `SampleGame.Android` doesn't really *need* this change for anything as it's just a spinning box (so we may even want to revert that),
- this is a "framework" project, so adding an overlay does nothing for consumers anyway. And if we had android template projects they couldn't use the overlay anyways (as they have to be self-contained).

Susko3 (dismissed):

PR #5670, opened 2023-03-08
Fix `Quad.Contains` sometimes failing due to floating point precision issues

### LINKED ISSUE #5669: `Quad.Contains` doesn't work correctly at precise centre point
See commented test in https://github.com/ppy/osu-framework/pull/5668

Both of the following fail when they shouldn't:

```csharp
Assert.That(
    new Quad(
            new Vector2(1513.5333f, 695.51416f),
            new Vector2(1679.6332f, 695.51416f),
            new Vector2(1513.5333f, 861.614f),
            new Vector2(1679.6332f, 861.614f)
        )
        .Contains(
            new Vector2(1596.5833f, 778.5641f)
        ), Is.True);

// or alternatively
var quad = new Quad(
    new Vector2(1513.5333f, 695.51416f),
    new Vector2(1679.6332f, 695.51416f),
    new Vector2(1513.5333f, 861.614f),
    new Vector2(1679.6332f, 861.614f)
);

Assert.That(quad.Contains(quad.Centre), Is.True);
```
 Looks related to floating point precision issues in the double-triangle contains implementation:

https://github.com/ppy/osu-framework/blob/63405204eba19d054f24c8f349137129a1557936/osu.Framework/Graphics/Primitives/Quad.cs#L110-L112

Noticed while writing a test scene where

```
InputManager.MoveMouseTo(drawable);
Assert.That(drawable.IsHovered, Is.True);
```

was failing at certain window resolutions.

bdach:
Closes #5669.

I will pre-empt this by saying that I am only *mostly* confident of this fix, so please be on the look-out if you don't like anything here. It took me the better part of an hour to convince myself that this works due to reasons I will elaborate on below.

To the point: In order to avoid the issue demonstrated by the test in #5668, a different way of computing whether a given point is inside of the quad is used. Namely, this method uses the perpendicular dot product, which has the property of

$$ \textbf{a}^\perp \cdot \textbf{b} = |\textbf{a}| |\textbf{b}| \sin \theta $$

wherein $\theta$ is the angle between the vectors $\textbf{a}$ and $\textbf{b}$.[^1] In particular, this means that because the sine of an angle is positive in the range [0, 180deg], and negative in the range [-180deg, 0], it allows to distinguish which side of the half-plane defined by the direction of vector $\textbf{a}$ the vector $\textbf{b}$ lies. On the Cartesian plane, it will be positive if $\textbf{b}$ is "on the left of" or "counterclockwise" to $\textbf{a}$, zero if they are collinear, and negative otherwise.

Assuming `Quad` is convex (which can be done, as it inherits `IConvexPolygon` - and notably, the previous implementation also silently assumed convexity, because in the general case you can't just arbitrarily slice the quad into two tris if it's not convex), we can say that a point lies within the quad if it lies on the same side of four half-planes defined by the quad's sides (by which I mean, for every side, we can divide the plane using a line going through both ends of the segment, and take the half-plane where the quad's interior is). Due to this we can therefore check whether for every vertex of the quad, the sign of the perp dot product between the vector which starts at the current vertex and ends at the next vertex (going in counterclockwise order), and the vector which starts at the current vertex and ends at the point being tested by `Contains()`, is the same - if it is, then the tested point is inside of the quad.

Now this is the part where this gets confusing: On the Cartesian plane, the sign we want would be *positive* (or zero, as zero means collinearity). But because computer graphics are hell and somebody somewhere decided that when drawing to the screen Y should be *down* and not up and that is what `Quad` abides by, the signs need to flip. So for our test here we use zero or less.

Interestingly, I have managed to produce a benchmark (included in the diff) which happens to show that this method is faster:

### Before

|   Method | NumPoints |        Mean |     Error |    StdDev | Allocated |
|--------- |---------- |------------:|----------:|----------:|----------:|
| Contains |        10 |    274.5 ns |   1.10 ns |   1.03 ns |         - |
| Contains |       100 |  2,903.4 ns |   3.65 ns |   3.41 ns |         - |
| Contains |      1000 | 29,330.9 ns | 297.09 ns | 263.36 ns |         - |

### After

|   Method | NumPoints |        Mean |     Error |    StdDev | Allocated |
|--------- |---------- |------------:|----------:|----------:|----------:|
| Contains |        10 |    62.32 ns |  0.726 ns |  0.679 ns |         - |
| Contains |       100 |   573.51 ns |  3.283 ns |  2.742 ns |         - |
| Contains |      1000 | 5,937.07 ns | 62.791 ns | 52.433 ns |         - |

but this is very likely highly dependent on data. The new method will be likely faster for rejections (as checking half-planes may allow to reject a point faster) but slower for hits (as all four half-planes must be examined), which is probably the correct access pattern to optimise for anyways?

As for testing, I did some using the draw visualiser but admittedly not much. I have checked that this diff passes game-side tests which is something extra I suppose?

[^1]: https://mathworld.wolfram.com/PerpDotProduct.html

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkQuad.cs
+++ b/osu.Framework.Benchmarks/BenchmarkQuad.cs
@@ -1,0 +1,47 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using BenchmarkDotNet.Attributes;
+using osu.Framework.Graphics.Primitives;
+using osuTK;
+
+namespace osu.Framework.Benchmarks
+{
+    [MemoryDiagnoser]
+    public class BenchmarkQuad
+    {
+        private Quad quad;
+
+        [Params(10, 100, 1000)]
+        public int NumPoints;
+
+        private Vector2[] points;
+        private bool[] results;
+
+        [GlobalSetup]
+        public void GlobalSetup()
+        {
+            quad = new Quad(
+                new Vector2(3, 0),
+                new Vector2(5, 1),
+                new Vector2(0, 5),
+                new Vector2(7, 7)
+            );
+
+            points = new Vector2[NumPoints];
+            results = new bool[NumPoints];
+
+            var random = new Random(20230307);
+            for (int i = 0; i < NumPoints; ++i)
+                points[i] = new Vector2(random.Next(0, 10), random.Next(0, 10));
+        }
+
+        [Benchmark]
+        public void Contains()
+        {
+            for (int i = 0; i < NumPoints; ++i)
+                results[i] = quad.Contains(points[i]);
+        }
+    }
+}
--- a/osu.Framework.Benchmarks/BenchmarkQuad.cs
+++ b/osu.Framework.Benchmarks/BenchmarkQuad.cs
@@ -16,8 +16,8 @@ namespace osu.Framework.Benchmarks
         [Params(10, 100, 1000)]
         public int NumPoints;
 
-        private Vector2[] points;
-        private bool[] results;
+        private Vector2[] points = null!;
+        private bool[] results = null!;
 
         [GlobalSetup]
         public void GlobalSetup()
--- a/osu.Framework.Tests/Primitives/QuadTest.cs
+++ b/osu.Framework.Tests/Primitives/QuadTest.cs
@@ -23,6 +23,13 @@ namespace osu.Framework.Tests.Primitives
                 new Vector2(1679.6332f, 695.51416f),
                 new Vector2(1513.5333f, 861.614f),
                 new Vector2(1679.6332f, 861.614f)
+            ),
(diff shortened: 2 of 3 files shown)


peppy:
[Blaming back](https://github.com/ppy/osu-framework/blame/8c69bbf03e1ccbcee508a7ea709c3391ddaa8ab7/osu.Framework/Graphics/Primitives/Quad.cs#L114C25-L121
) to try and find the original code owner ends at the initial framework commit 😅.

@Tom94 I don't remember if you were behind the original implementation of this (or if it was taken from another project), but this might be something you're interested in taking a quick look at.

The math looks to check out as far as I can tell, and fixes the case I was encountering 👍 .

peppy (approved):

Tom94:
Ha, yes, I remember hacking the `Contains` function using two `Triangles`. It already felt hacky back then and the new implementation is indeed better in every respect as long as we don't create non-convex quads anywhere. (We shouldn't, if all we do is trainsform rectangles with matrices.)

PR #5671, opened 2023-03-09
Add `async Task` support to BDL

smoogipoo:
This adds support (reflection and source-gen), as well as an analyser to make sure that `async` BDL methods always return `Task`. There's no test for the new analyser, like there isn't a test for the other analyser, because that's pending more tooling that has yet to be ported over from `osu-localisation-analyser`.

It works by waiting on the BDL method via `WaitSafely()`.

To be tested alongside https://github.com/ppy/osu-framework/pull/5646


smoogipoo:
The new test will fail until a new SG package is released/consumed here.

peppy:
One thing I was going to ask is should we be using `WaitSafetly` here as opposed to `Wait` with a `CancellationToken` provided?

smoogipoo:
I... did not know `.Wait()` had a `CancellationToken` override. Sounds reasonable.

smoogipoo:
I don't know if we should be passing the cancellation token into the `.Wait()`. In a simple test, it looks like it cancels the wait even if the task is still running, which makes sense but I don't think that would be expected from o!f - to still have things potentially running in an async context while potentially a Dispose() is now firing off.

Which brings up another issue - if we don't already (I don't think we do), BDL needs a `SynchronizationContext` otherwise async continuation should fail due to:

https://github.com/ppy/osu-framework/blob/66d562fe1ab45641f665dcc6c08d081f54108aed/osu.Framework/Graphics/Drawable.cs#L2596-L2622

smoogipoo:
I've tested my newest changes with the above branches and it looks to be working correctly now.

osu.Framework/Utils/SourceGeneratorUtils.cs
```diff
@@ -73,5 +77,63 @@ public static T GetDependency<T>(IReadOnlyDependencyContainer container, Type ca
...
+                    return this;
+
+                TaskAwaiter awaiter = task.GetAwaiter();
+                awaiter.OnCompleted(() => ready.Set());
+
+                while (!awaiter.IsCompleted)
+                {
+                    ready.Wait();
+                    scheduler.Update();
+                }
```
smoogipoo, line 127: I'm not sure about this implementation.

One would think that simply
```csharp
while (!task.IsCompleted)
    scheduler.Update();
```
Would serve the same purpose, however this seems to not be behaving correctly - the Send/Post methods don't get called and my debugger deadlocks.

Please do suggest a better solution.

smoogipoo:
The changes to the test above fail on master and on this branch prior to the synchronisation context.

As discussed, in the future we could improve on the synchronisation context in a few ways:
- Right now it's allocating a new context every `load()`. We could probably make it `[ThreadStatic]` inside the BDL threadpool.
- It may be possible to rework it such that we don't need to block a BDL thread for the entire duration. I think this one in particular is going to require quite a few changes/rethinking.

osu.Framework/Utils/SourceGeneratorUtils.cs
peppy: I dunno, it looks to work? I'd say we go with it until we see an issue with it.

peppy: Just for visibility, discussed this further IRL and this change also works:

```diff
diff --git a/osu.Framework.Tests/Dependencies/TestSceneAsyncBackgroundDependencyLoader.cs b/osu.Framework.Tests/Dependencies/TestSceneAsyncBackgroundDependencyLoader.cs
index 2d9000344..44dd9b681 100644
--- a/osu.Framework.Tests/Dependencies/TestSceneAsyncBackgroundDependencyLoader.cs
+++ b/osu.Framework.Tests/Dependencies/TestSceneAsyncBackgroundDependencyLoader.cs
@@ -10,8 +10,9 @@
 
 namespace osu.Framework.Tests.Dependencies
 {
+#pragma warning disable OFSG001
     [HeadlessTest]
-    public partial class TestSceneAsyncBackgroundDependencyLoader : FrameworkTestScene
+    public class TestSceneAsyncBackgroundDependencyLoader : FrameworkTestScene
     {
         [Test]
         public void TestLoadOnUpdateThread()
@@ -79,4 +80,5 @@ public void Verify()
         }
 #pragma warning restore OFSG001
     }
+#pragma warning restore OFSG001
 }
diff --git a/osu.Framework/Utils/SourceGeneratorUtils.cs b/osu.Framework/Utils/SourceGeneratorUtils.cs
index b60ed5f8d..ae7159839 100644
--- a/osu.Framework/Utils/SourceGeneratorUtils.cs
+++ b/osu.Framework/Utils/SourceGeneratorUtils.cs
@@ -2,7 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Runtime.CompilerServices;
+using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 using osu.Framework.Allocation;
@@ -90,7 +90,6 @@ private class BackgroundDependencyLoaderSynchronizationContext : Synchronization
         {
             private readonly SynchronizationContext? lastContext;
             private readonly Scheduler scheduler;
-            private readonly ManualResetEventSlim ready = new ManualResetEventSlim();
 
             public BackgroundDependencyLoaderSynchronizationContext()
             {
@@ -108,7 +107,6 @@ public override void Send(SendOrPostCallback d, object? state)
             public override void Post(SendOrPostCallback d, object? state)
             {
                 scheduler.Add(() => d(state));
-                ready.Set();
             }
 
 #pragma warning disable RS0030
@@ -117,12 +115,9 @@ public IBackgroundDependencyLoaderContext WaitFor(Task? task)
                 if (task == null)
                     return this;
 
-                TaskAwaiter awaiter = task.GetAwaiter();
-                awaiter.OnCompleted(() => ready.Set());
-
-                while (!awaiter.IsCompleted)
+                while (!task.IsCompleted)
                 {
-                    ready.Wait();
+                    Debugger.NotifyOfCrossThreadDependency();
                     scheduler.Update();
                 }
 
```

*but* we probably shouldn't do this due to the method [being quite expensive](https://github.com/dotnet/runtime/issues/38736) (it's the reason async causes such huge overheads for us already).

I'm... not sure.

smoogipoo: As far as I understand, `Debugger.NotifyOfCrossThreadDependency()` causes other threads to run in order to be able to evaluate the function. That is, if the debugger is paused in this while loop, then the async thread is _also_ paused, but it needs to run to be able to `Post()` the continuation.

So it seems like it is the intended fix here.

smoogipoo:
As a note, `ConfigureAwait(true)` is the default. I'd be fine with either enforcing `ConfigureAwait(true)` or _not_ `ConfigureAwait(false)` (i.e. not specifying anything at all), which may make more sense from a consumer/code-standards perspective.

PR #5673, opened 2023-03-10
Update github actions to latest versions

peppy:
Aiming to stop warnings which are telling us we need to do this (see https://github.com/ppy/osu-framework/actions/runs/4383176901).


smoogipoo:
Looks like a few places still need to be updated.

peppy:
Hopefully got them all this time..

Issue #5674, opened 2023-03-10
HSV colour picker causes GPU-level crash on M1 with OpenGL Core


frenzibyte:
Simply running `TestSceneHSVColourPicker` once or twice results in the following:

```
[runtime] 2023-03-10 19:11:36 [verbose]: 💨 HSVColourPicker(TestSceneHSVColourPicker) HueSelectorInput
[runtime] 2023-03-10 19:11:36 [verbose]: 🔸 Step #3 [SetUp]
[runtime] 2023-03-10 19:11:36 [verbose]: 🔸 Step #4 create picker
[runtime] 2023-03-10 19:11:36 [verbose]: 🔸 Step #5 hue selector has 0
FALLBACK (log once): Fallback to SW vertex processing for VERT_BUF_REQ
FALLBACK (log once): Fallback to SW fragment processing for FRAG_BUF_REQ
FALLBACK (log once): Fallback to SW vertex processing, m_disable_code: 1
FALLBACK (log once): Fallback to SW fragment processing, m_disable_code: 10000
```

This started occurring since #5655, and according to the logs it seems to be specific to M1's OpenGL–Metal wrapper driver. I've bumped this all the way to p0 as it's going to result in unwanted crashes osu!-side and must be fixed before release.

Preliminary investigation shows that this occurs specifically when the `m_HueData` buffer is populated with `hue = 0`.

cc/ @smoogipoo

PR #5675, opened 2023-03-11
Fix UBOs not being initialised when data is default

smoogipoo:
Fixes https://github.com/ppy/osu-framework/issues/5674

When setting `Data`, if the value is `default(T)` then the equality comparer will exit early. But it turns out that macOS (maybe M1 specific) needs the call to `GL.BufferData()` to initialise the buffer's memory.

I don't know if this will also affect the Veldrid implementation, since we can't currently run that with OpenGL on macOS.


frenzibyte (approved):
Looks good. Veldrid's OpenGL implementation already [clears the buffer on initialisation](https://github.com/mellinoe/veldrid/blob/a121087cadf38755f28c397a5b3c42ff1c559a19/src/Veldrid/OpenGL/OpenGLBuffer.cs#L74-L85), so it's doesn't look to be required there (testing it also shows it to work).

For reference, I'm seeing a similar crash with Metal backend, but that's a bit too involved than this one and is lesser in priority, so I'm counting the original issue thread as resolved.

PR #5682, opened 2023-03-16
Allow renderer choice to be stored in user configuration (and add automatic fallbacks)

peppy:
RFC

The goal here is to provide both renderer fallback logic, but also expose the renderer to the user in a way they can interact with it from game settings (or via config file for recovery, should that ever be required).

Example with user set configuration that is invalid:

```csharp
[runtime] 2023-03-16 10:45:09 [verbose]: 🖼 Configuration renderer choice: Vulkan
[runtime] 2023-03-16 10:45:09 [verbose]: 🖼️ Renderer fallback order: [ Vulkan, Metal, OpenGLLegacy ]
[runtime] 2023-03-16 10:45:09 [verbose]: 🖼️ Attempting initialisation using renderer: veldrid surface: Vulkan
[runtime] 2023-03-16 10:45:09 [verbose]: Updated display mode to desktop resolution: 2560x1440@144, SDL_PIXELFORMAT_ARGB8888
[runtime] 2023-03-16 10:45:09 [verbose]: 🖼️ Renderer initialisation failed with:
[runtime] 2023-03-16 10:45:09 [verbose]: System.TypeInitializationException: The type initializer for 'Vulkan.VulkanNative' threw an exception.
[runtime] 2023-03-16 10:45:09 [verbose]: ---> System.InvalidOperationException: Could not load libvulkan.dylib
[runtime] 2023-03-16 10:45:09 [verbose]: at Vulkan.NativeLibrary..ctor(String libraryName)
[runtime] 2023-03-16 10:45:09 [verbose]: at Vulkan.NativeLibrary.UnixNativeLibrary..ctor(String libraryName)
[runtime] 2023-03-16 10:45:09 [verbose]: at Vulkan.NativeLibrary.Load(String libraryName)
[runtime] 2023-03-16 10:45:09 [verbose]: at Vulkan.VulkanNative.LoadNativeLibrary()
[runtime] 2023-03-16 10:45:09 [verbose]: at Vulkan.VulkanNative..cctor()
[runtime] 2023-03-16 10:45:09 [verbose]: --- End of inner exception stack trace ---
[runtime] 2023-03-16 10:45:09 [verbose]: at Veldrid.Vk.VulkanUtil.EnumerateInstanceLayers()
[runtime] 2023-03-16 10:45:09 [verbose]: at Veldrid.Vk.VkGraphicsDevice.CreateInstance(Boolean debug, VulkanDeviceOptions options)
[runtime] 2023-03-16 10:45:09 [verbose]: at Veldrid.Vk.VkGraphicsDevice..ctor(GraphicsDeviceOptions options, Nullable`1 scDesc, VulkanDeviceOptions vkOptions)
[runtime] 2023-03-16 10:45:09 [verbose]: at Veldrid.Vk.VkGraphicsDevice..ctor(GraphicsDeviceOptions options, Nullable`1 scDesc)
[runtime] 2023-03-16 10:45:09 [verbose]: at Veldrid.GraphicsDevice.CreateVulkan(GraphicsDeviceOptions options, SwapchainDescription swapchainDescription)
[runtime] 2023-03-16 10:45:09 [verbose]: at osu.Framework.Graphics.Veldrid.VeldridRenderer.Initialise(IGraphicsSurface graphicsSurface) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs:line 146
[runtime] 2023-03-16 10:45:09 [verbose]: at osu.Framework.Graphics.Rendering.Renderer.osu.Framework.Graphics.Rendering.IRenderer.Initialise(IGraphicsSurface graphicsSurface) in /Users/dean/Projects/osu-framework/osu.Framework/Graphics/Rendering/Renderer.cs:line 171
[runtime] 2023-03-16 10:45:09 [verbose]: at osu.Framework.Platform.GameHost.setupRendererAndWindow(String renderer, GraphicsSurfaceType surfaceType) in /Users/dean/Projects/osu-framework/osu.Framework/Platform/GameHost.cs:line 897
[runtime] 2023-03-16 10:45:09 [verbose]: 🖼️ Attempting initialisation using renderer: veldrid surface: Metal
[runtime] 2023-03-16 10:45:09 [verbose]: Updated display mode to desktop resolution: 2560x1440@144, SDL_PIXELFORMAT_ARGB8888
[runtime] 2023-03-16 10:45:09 [verbose]: Metal Initialized
[runtime] 2023-03-16 10:45:09 [verbose]: Metal Feature Set: macOS GPU family 2 (v1)
[runtime] 2023-03-16 10:45:09 [verbose]: 🖼️ Renderer initialised!
```

Legacy example:

```csharp
[runtime] 2023-03-16 10:45:45 [verbose]: 🖼 Configuration renderer choice: OpenGLLegacy
[runtime] 2023-03-16 10:45:45 [verbose]: 🖼️ Renderer fallback order: [ OpenGLLegacy ]
[runtime] 2023-03-16 10:45:45 [verbose]: 🖼️ Attempting initialisation using renderer: gl surface: OpenGL
[runtime] 2023-03-16 10:45:45 [verbose]: Updated display mode to desktop resolution: 2560x1440@144, SDL_PIXELFORMAT_ARGB8888
[runtime] 2023-03-16 10:45:45 [verbose]: GL Initialized
[runtime] 2023-03-16 10:45:45 [verbose]: GL Version:                 4.1 ATI-4.9.51
[runtime] 2023-03-16 10:45:45 [verbose]: GL Renderer:                AMD Radeon Pro 580X OpenGL Engine
[runtime] 2023-03-16 10:45:45 [verbose]: GL Shader Language version: 4.10
[runtime] 2023-03-16 10:45:45 [verbose]: GL Vendor:                  ATI Technologies Inc.
[runtime] 2023-03-16 10:45:45 [verbose]: GL Extensions:              GL_ARB_blend_func_extended GL_ARB_draw_buffers_blend GL_ARB_draw_indirect GL_ARB_ES2_compatibility GL_ARB_explicit_attrib_location GL_ARB_gpu_shader_fp64 GL_ARB_gpu_shader5 GL_ARB_instanced_arrays GL_ARB_internalformat_query GL_ARB_occlusion_query2 GL_ARB_sample_shading GL_ARB_sampler_objects GL_ARB_separate_shader_objects GL_ARB_shader_bit_encoding GL_ARB_shader_subroutine GL_ARB_shading_language_include GL_ARB_tessellation_shader GL_ARB_texture_buffer_object_rgb32 GL_ARB_texture_cube_map_array GL_ARB_texture_gather GL_ARB_texture_query_lod GL_ARB_texture_rgb10_a2ui GL_ARB_texture_storage GL_ARB_texture_swizzle GL_ARB_timer_query GL_ARB_transform_feedback2 GL_ARB_transform_feedback3 GL_ARB_vertex_attrib_64bit GL_ARB_vertex_type_2_10_10_10_rev GL_ARB_viewport_array GL_EXT_debug_label GL_EXT_debug_marker GL_EXT_depth_bounds_test GL_EXT_texture_compression_s3tc GL_EXT_texture_filter_anisotropic GL_EXT_texture_mirror_clamp GL_EXT_texture_sRGB_decode GL_APPLE_client_storage GL_APPLE_container_object_shareable GL_APPLE_flush_render GL_APPLE_object_purgeable GL_APPLE_rgb_422 GL_APPLE_row_bytes GL_APPLE_texture_range GL_ATI_texture_mirror_once GL_NV_texture_barrier
[runtime] 2023-03-16 10:45:45 [verbose]: 🖼️ Renderer initialised!

```

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -785,107 +785,130 @@ namespace osu.Framework.Platform
             // Always give preference to environment variables.
             if (FrameworkEnvironment.PreferredGraphicsSurface != null || FrameworkEnvironment.PreferredGraphicsRenderer != null)
             {
+                Logger.Log("🖼️ Using environment variables for renderer and surface selection.", level: LogLevel.Important);
+
                 // And allow this to hard fail with no fallbacks.
                 setupRendererAndWindow(
                     FrameworkEnvironment.PreferredGraphicsRenderer ?? "veldrid",
                     FrameworkEnvironment.PreferredGraphicsSurface ?? GraphicsSurfaceType.OpenGL);
+                return;
             }
-            else
+
+            List<GraphicsSurfaceType> attemptSurfaceTypes = new List<GraphicsSurfaceType>();
+            bool attemptOperatingSystemSpecificFallbacks = true;
+
+            var configRenderer = Config.Get<Configuration.Renderer>(FrameworkSetting.Renderer);
+            Logger.Log($"🖼️ Configuration renderer choice: {configRenderer}");
+
+            switch (configRenderer)
             {
-                List<GraphicsSurfaceType> attemptSurfaceTypes = new List<GraphicsSurfaceType>();
-                bool attemptOperatingSystemSpecificFallbacks = true;
+                case Configuration.Renderer.Metal:
+                    attemptSurfaceTypes.Add(GraphicsSurfaceType.Metal);
+                    break;
 
-                switch (Config.Get<Configuration.Renderer>(FrameworkSetting.Renderer))
-                {
-                    case Configuration.Renderer.Metal:
-                        attemptSurfaceTypes.Add(GraphicsSurfaceType.Metal);
-                        break;
+                case Configuration.Renderer.Vulkan:
+                    attemptSurfaceTypes.Add(GraphicsSurfaceType.Vulkan);
+                    break;
 
-                    case Configuration.Renderer.Vulkan:
-                        attemptSurfaceTypes.Add(GraphicsSurfaceType.Vulkan);
-                        break;
+                case Configuration.Renderer.Direct3D11:
+                    attemptSurfaceTypes.Add(GraphicsSurfaceType.Direct3D11);
+                    break;
 
-                    case Configuration.Renderer.Direct3D11:
-                        attemptSurfaceTypes.Add(GraphicsSurfaceType.Direct3D11);
-                        break;
+                case Configuration.Renderer.OpenGL:
+                    attemptSurfaceTypes.Add(GraphicsSurfaceType.OpenGL);
+                    break;
 
-                    case Configuration.Renderer.OpenGL:
-                        attemptSurfaceTypes.Add(GraphicsSurfaceType.OpenGL);
-                        break;
-
-                    case Configuration.Renderer.Legacy:
-                        attemptOperatingSystemSpecificFallbacks = false;
-                        break;
-                }
-
-                if (attemptOperatingSystemSpecificFallbacks)
-                {
-                    // Best case, we can make use of veldrid with a new graphics API.
-                    switch (RuntimeInfo.OS)
-                    {
-                        case RuntimeInfo.Platform.Windows:
-                            attemptSurfaceTypes.Add(GraphicsSurfaceType.Vulkan);
-                            attemptSurfaceTypes.Add(GraphicsSurfaceType.Direct3D11);
-                            break;
-
-                        case RuntimeInfo.Platform.Linux:
-                        case RuntimeInfo.Platform.Android:
-                            attemptSurfaceTypes.Add(GraphicsSurfaceType.Vulkan);
-                            break;
-
-                        case RuntimeInfo.Platform.macOS:
-                        case RuntimeInfo.Platform.iOS:
-                            attemptSurfaceTypes.Add(GraphicsSurfaceType.Metal);
-                            break;
-                    }
-                }
-
-                foreach (var attemptSurfaceType in attemptSurfaceTypes)
-                {
-                    try
-                    {
-                        setupRendererAndWindow("veldrid", attemptSurfaceType);
-                        return;
-                    }
-                    catch
-                    {
-                        // If we fail, assume the user may have had a custom setting and switch it back to automatic.
-                        Config.SetValue(FrameworkSetting.Renderer, Configuration.Renderer.Automatic);
-                    }
-                }
-
-                // fallback to legacy renderer. this is basically guaranteed to support all platforms.
-                setupRendererAndWindow("gl", GraphicsSurfaceType.OpenGL);
-                Config.SetValue(FrameworkSetting.Renderer, Configuration.Renderer.Legacy);
+                case Configuration.Renderer.Legacy:
+                    attemptOperatingSystemSpecificFallbacks = false;
+                    break;
             }
+
+            if (attemptOperatingSystemSpecificFallbacks)
+            {
+                // Best case, we can make use of veldrid with a new graphics API.
+                switch (RuntimeInfo.OS)
+                {
+                    case RuntimeInfo.Platform.Windows:
+                        addFallback(GraphicsSurfaceType.Vulkan);
+                        addFallback(GraphicsSurfaceType.Direct3D11);
+                        break;
+
+                    case RuntimeInfo.Platform.Linux:
+                    case RuntimeInfo.Platform.Android:
+                        addFallback(GraphicsSurfaceType.Vulkan);
+                        break;
+
+                    case RuntimeInfo.Platform.macOS:
+                    case RuntimeInfo.Platform.iOS:
+                        addFallback(GraphicsSurfaceType.Metal);
+                        break;
+                }
+
+                void addFallback(GraphicsSurfaceType surface)
+                {
+                    if (!attemptSurfaceTypes.Contains(surface))
+                        attemptSurfaceTypes.Add(surface);
+                }
+            }
+
+            Logger.Log($"🖼️ Renderer fallback order: [ {string.Join(", ", attemptSurfaceTypes.Select(e => e.ToString()).Append("Legacy"))} ]");
+
+            foreach (var attemptSurfaceType in attemptSurfaceTypes)
+            {
+                try
+                {
+                    setupRendererAndWindow("veldrid", attemptSurfaceType);
+                    return;
+                }
+                catch
+                {
+                    // If we fail, assume the user may have had a custom setting and switch it back to automatic.
+                    Config.SetValue(FrameworkSetting.Renderer, Configuration.Renderer.Automatic);
+                }
+            }
+
+            // fallback to legacy renderer. this is basically guaranteed to support all platforms.
+            setupRendererAndWindow("gl", GraphicsSurfaceType.OpenGL);
+            Config.SetValue(FrameworkSetting.Renderer, Configuration.Renderer.Legacy);
         }
 
         private void setupRendererAndWindow(string renderer, GraphicsSurfaceType surfaceType)
         {
-            Logger.Log($"Attempting initialisation using renderer: {renderer} surface: {surfaceType}");
+            Logger.Log($"🖼️ Attempting initialisation using renderer: {renderer} surface: {surfaceType}");
 
-            Renderer = renderer == "veldrid"
-                ? new VeldridRenderer()
-                : CreateGLRenderer();
-
-            // Prepare window
-            Window = CreateWindow(surfaceType);
-
-            initialiseInputHandlers();
-
-            if (Window != null)
+            try
             {
-                Window.SetupWindow(Config);
+                Renderer = renderer == "veldrid"
+                    ? new VeldridRenderer()
+                    : CreateGLRenderer();
 
-                Window.Create();
-                Window.Title = $@"osu!framework (running ""{Name}"")";
+                // Prepare window
+                Window = CreateWindow(surfaceType);
 
-                Renderer.Initialise(Window.GraphicsSurface);
+                initialiseInputHandlers();
 
-                currentDisplayMode = Window.CurrentDisplayMode.GetBoundCopy();
-                currentDisplayMode.BindValueChanged(_ => updateFrameSyncMode());
+                if (Window != null)
+                {
+                    Window.SetupWindow(Config);
 
-                IsActive.BindTo(Window.IsActive);
+                    Window.Create();
+                    Window.Title = $@"osu!framework (running ""{Name}"")";
+
+                    Renderer.Initialise(Window.GraphicsSurface);
+
+                    currentDisplayMode = Window.CurrentDisplayMode.GetBoundCopy();
+                    currentDisplayMode.BindValueChanged(_ => updateFrameSyncMode());
+
+                    IsActive.BindTo(Window.IsActive);
+                }
+
+                Logger.Log("🖼️ Renderer initialised!");
+            }
+            catch (Exception e)
+            {
+                Logger.Log("🖼️ Renderer initialisation failed with:");
+                Logger.Log(e.ToString());
+                throw;
             }
         }
 
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -888,13 +888,26 @@ namespace osu.Framework.Platform
 
         protected void SetupRendererAndWindow(string renderer, GraphicsSurfaceType surfaceType)
         {
+            switch (renderer)
+            {
+                case "veldrid":
+                    SetupRendererAndWindow(new VeldridRenderer(), surfaceType);
+                    break;
+
+                default:
+                case "gl":
+                    SetupRendererAndWindow(CreateGLRenderer(), surfaceType);
+                    break;
+            }
+        }
+
+        protected void SetupRendererAndWindow(IRenderer renderer, GraphicsSurfaceType surfaceType)
+        {
             Logger.Log($"🖼️ Attempting initialisation using renderer: {renderer} surface: {surfaceType}");
 
             try
             {
-                Renderer = renderer == "veldrid"
-                    ? new VeldridRenderer()
-                    : CreateGLRenderer();
+                Renderer = renderer;
 
                 // Prepare window
                 Window = CreateWindow(surfaceType);
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -668,8 +668,6 @@ namespace osu.Framework.Platform
             if (ExecutionState != ExecutionState.Idle)
                 throw new InvalidOperationException("A game that has already been run cannot be restarted.");
 
-            Renderer = CreateRenderer();
-
             try
             {
                 if (!host_running_mutex.Wait(10000))
@@ -689,13 +687,6 @@ namespace osu.Framework.Platform
                     Monitor = { HandleGC = true },
                 });
 
-                GraphicsSurfaceType surfaceType = FrameworkEnvironment.PreferredGraphicsSurface ?? GraphicsSurfaceType.OpenGL;
-
-                Logger.Log("Using renderer: " + Renderer.GetType().ReadableName());
-                Logger.Log("Using graphics surface: " + surfaceType);
-
-                RegisterThread(DrawThread = new DrawThread(DrawFrame, this, $"{Renderer.GetType().ReadableName().Replace("Renderer", "")} / {surfaceType}"));
-
                 Trace.Listeners.Clear();
                 Trace.Listeners.Add(new ThrowingTraceListener());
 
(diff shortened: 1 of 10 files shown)


frenzibyte (dismissed):
Looking pretty solid to me. I've pushed a number of minor changes to further solidify the logic and get this to a complete state. @peppy check and see whether you're on board with it.

@smoogipoo would appreciate another pair of eyes on this one.

peppy:
Would have preferred review rather than direct commits for changes.

osu.Framework/Platform/SDL2DesktopWindow.cs
```diff
@@ -303,11 +298,7 @@ public void OnDraw()
         /// <summary>
         /// Forcefully closes the window.
         /// </summary>
-        public void Close() => ScheduleCommand(() =>
```
peppy: What's the reasoning behind this one?

I'm very unsure about delaying `Quit` until `Dispose` as well. That seems super wrong. `Dispose` isn't supposed to be used for something like closing a window – it should be for cleanup of resources that were potentially forgotten about.

At very least I'd want `Close` to do the window destroy and have a `Close` call from `Dispose` as a safety.

frenzibyte: Well, `Dispose` to me sounds like the natural place for cleaning up unmanaged resources, which would include calling `SDL_Quit` and `SDL_DestroyWindow`.

Doing this at `Close` method should be fine, as long as the `Exited` event is still invoked before any other operation.

frenzibyte:
> Would have preferred review rather than direct commits for changes.

Fair enough, will keep that in mind.

osu.Framework/Platform/SDL2DesktopWindow.cs
peppy: I'll make changes

peppy: I've made some changes, see what you think. Seems to work well in testing standard and also multi-fallback flows.

frenzibyte (approved):

peppy:
I've added one more important log output to allow for this:

osu! 2023-03-17 at 08 55 05

peppy:
I've refactored this heavily to allow the game-side select to work... Will probably just push to this branch since it changes so much.

peppy:
@frenzibyte please re-review for readability. see usage at https://github.com/ppy/osu/compare/master...peppy:osu:add-renderer-selection?expand=1 if curious.

osu.Framework/Platform/GameHost.cs
```diff
@@ -814,6 +795,197 @@ public void Run(Game game)
...
+            {
+                case RuntimeInfo.Platform.Windows:
+                    yield return RendererType.Vulkan;
+                    yield return RendererType.Direct3D11;
+
+                    break;
+
+                case RuntimeInfo.Platform.Linux:
+                case RuntimeInfo.Platform.Android:
+                    yield return RendererType.Vulkan;
```
smoogipoo, line 830: I suggest we don't use Vulkan on Linux yet. Since I'm a Linux user I'll be debugging it, but it is broken at runtime.

osu.Framework/Platform/GameHost.cs
```diff
@@ -814,6 +795,197 @@ public void Run(Game game)
...
+            // Prepare window
+            Window = CreateWindow(surfaceType);
+
+            if (Window == null)
+            {
+                Logger.Log("🖼️ Renderer could not be initialised, no window exists.");
+                return;
+            }
+
+            Window.SetupWindow(Config);
```
smoogipoo, line 959: Any reason for this to not be in the `try { }`? Disposal being the point.

osu.Framework/Platform/GameHost.cs
```diff
@@ -814,6 +795,197 @@ public void Run(Game game)
...
+        /// </summary>
+        public RendererType ResolvedRenderer { get; private set; }
+
+        /// <summary>
+        /// All valid <see cref="RendererType"/>s for the current platform.
+        /// </summary>
+        public IEnumerable<RendererType> GetValidRenderersForCurrentPlatform() =>
+            GetPreferredRenderersForCurrentPlatform()
+                .Append(RendererType.OpenGL)
+                .Append(RendererType.OpenGLLegacy);
```
smoogipoo, line 809: I'd probably reverse the order of these for now since OpenGL is pretty broken until https://github.com/mellinoe/veldrid/pull/481 and I don't know how deep the effects of it go beyond the intro (e.g. I don't if there's some case where sliders also won't render as a result/etc).

osu.Framework/Platform/GameHost.cs
```diff
@@ -814,6 +795,197 @@ public void Run(Game game)
...
+        public IEnumerable<RendererType> GetPreferredRenderersForCurrentPlatform()
+        {
+            yield return RendererType.Automatic;
+
+            // Best case, we can make use of veldrid with a new graphics API.
+            switch (RuntimeInfo.OS)
+            {
+                case RuntimeInfo.Platform.Windows:
+                    yield return RendererType.Vulkan;
+                    yield return RendererType.Direct3D11;
```
smoogipoo, line 824: I don't think we should default to Vulkan on Windows yet. For me it looks to be deadlocking in an infinite allocate loop on a 3060ti, and I'm not sure where yet (somewhere inside Veldrid) because I haven't really been testing Vulkan on Windows.

D3D11, while still not yet perfect and will experience glitchyness, is at least a little bit more stable and I will be continuing my work there before Vulkan because I don't know how Vulkan will behave with Optimus for example.

smoogipoo (dismissed):
Bit of a preliminary review for now

osu.Framework/Platform/GameHost.cs
peppy: It used to be, I think..

osu.Framework/Platform/GameHost.cs
peppy: Yeah for sure. Also potentially d3d before vulkan on windows

osu.Framework/Platform/GameHost.cs
peppy: Yep, agree here too.

osu.Framework/Platform/GameHost.cs
frenzibyte: I've moved it outside the try-catch block since it pretty much focuses on binding window bindables with the framework config, as far as I've seen. Not sure what you mean by the disposal part, might be missing that piece.

osu.Framework/Platform/GameHost.cs
peppy: @smoogipoo the way i have things setup right now, if a renderer isn't in the preference list it won't be displayed at all. do you think we should still allow it to be displayed in settings (this can be done quite easily, just want to check your opinion first).

osu.Framework/Platform/GameHost.cs
peppy: These aren't in the preferred renderers, so order doesn't actually matter (it's basically for display order).

Veldrid GL isn't preferred at all for now.

osu.Framework/Platform/GameHost.cs
```diff
@@ -814,6 +795,221 @@ public void Run(Game game)
...
+
+            yield return RendererType.OpenGL;
+            yield return RendererType.OpenGLLegacy;
+        }
+
+        /// <summary>
+        /// All preferred <see cref="RendererType"/>s for the current platform.
+        /// These are used in order of appearance when <see cref="RendererType.Automatic"/> is used, or when the user preference fails to initialise.
+        /// </summary>
+        public IEnumerable<RendererType> GetPreferredRenderersForCurrentPlatform()
```
frenzibyte, line 839: Not really sure about having two methods for returning a list of renderer types. This method also doesn't return `OpenGLLegacy`, so it's not really returning all renderers that are actually attempted initialisation on with o!f.

I would stick to one method that returns a preferred list of renderers with all types present, and use `OrderBy(r => r)` to fix display order in the dropdown.

Something like:
```diff
diff --git a/osu.Framework.Tests/Visual/Platform/TestSceneRenderer.cs b/osu.Framework.Tests/Visual/Platform/TestSceneRenderer.cs
index 43cb47371..a2cc9f93e 100644
--- a/osu.Framework.Tests/Visual/Platform/TestSceneRenderer.cs
+++ b/osu.Framework.Tests/Visual/Platform/TestSceneRenderer.cs
@@ -34,7 +34,7 @@ public partial class TestSceneRenderer : FrameworkTestScene
                 Anchor = Anchor.Centre,
                 Origin = Anchor.Centre,
                 Current = config.GetBindable<RendererType>(FrameworkSetting.Renderer),
-                Items = host.GetValidRenderersForCurrentPlatform(),
+                Items = host.GetValidRenderers().OrderBy(r => r),
                 Width = 200f,
             });
         });
diff --git a/osu.Framework/Platform/GameHost.cs b/osu.Framework/Platform/GameHost.cs
index 4f3f40d53..e60276e4a 100644
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -806,7 +806,7 @@ public void Run(Game game)
         /// <summary>
         /// All valid <see cref="RendererType"/>s for the current platform.
         /// </summary>
-        public IEnumerable<RendererType> GetValidRenderersForCurrentPlatform()
+        public IEnumerable<RendererType> GetValidRenderers()
         {
             yield return RendererType.Automatic;
 
@@ -814,12 +814,14 @@ public IEnumerable<RendererType> GetValidRenderersForCurrentPlatform()
             {
                 case RuntimeInfo.Platform.Windows:
                     yield return RendererType.Direct3D11;
+                    yield return RendererType.OpenGLLegacy;
                     yield return RendererType.Vulkan;
 
                     break;
 
                 case RuntimeInfo.Platform.Linux:
                 case RuntimeInfo.Platform.Android:
+                    yield return RendererType.OpenGLLegacy;
                     yield return RendererType.Vulkan;
 
                     break;
@@ -827,43 +829,12 @@ public IEnumerable<RendererType> GetValidRenderersForCurrentPlatform()
                 case RuntimeInfo.Platform.macOS:
                 case RuntimeInfo.Platform.iOS:
                     yield return RendererType.Metal;
+                    yield return RendererType.OpenGLLegacy;
 
                     break;
             }
 
             yield return RendererType.OpenGL;
-            yield return RendererType.OpenGLLegacy;
-        }
-
-        /// <summary>
-        /// All preferred <see cref="RendererType"/>s for the current platform.
-        /// These are used in order of appearance when <see cref="RendererType.Automatic"/> is used, or when the user preference fails to initialise.
-        /// </summary>
-        public IEnumerable<RendererType> GetPreferredRenderersForCurrentPlatform()
-        {
-            yield return RendererType.Automatic;
-
-            // Best case, we can make use of veldrid with a new graphics API.
-            switch (RuntimeInfo.OS)
-            {
-                case RuntimeInfo.Platform.Windows:
-                    // yield return RendererType.Vulkan;
-                    yield return RendererType.Direct3D11;
-
-                    break;
-
-                case RuntimeInfo.Platform.Linux:
-                case RuntimeInfo.Platform.Android:
-                    // yield return RendererType.Vulkan;
-
-                    break;
-
-                case RuntimeInfo.Platform.macOS:
-                case RuntimeInfo.Platform.iOS:
-                    yield return RendererType.Metal;
-
-                    break;
-            }
         }
 
         protected virtual void ChooseAndSetupRenderer()
@@ -884,47 +855,40 @@ protected virtual void ChooseAndSetupRenderer()
             Logger.Log($"🖼️ Configuration renderer choice: {configRenderer}");
 
             // Attempt to initialise various veldrid surface types.
-            // If legacy GL was requested we can skip this and fallback to the final logic below.
-            if (configRenderer.Value != RendererType.OpenGLLegacy)
-            {
-                var rendererTypes = GetPreferredRenderersForCurrentPlatform().ToList();
+            var rendererTypes = GetValidRenderers().Where(r => r != RendererType.Automatic).ToList();
 
-                Logger.Log($"🖼️ Renderer fallback order: [ {string.Join(", ", rendererTypes.Select(e => e.GetDescription()).Append("OpenGL (Legacy)"))} ]");
+            Logger.Log($"🖼️ Renderer fallback order: [ {string.Join(", ", rendererTypes.Select(e => e.GetDescription()))} ]");
 
-                // Move user's preference to the start of the attempts.
-                rendererTypes.Remove(configRenderer.Value);
-                rendererTypes.Insert(0, configRenderer.Value);
+            // Move user's preference to the start of the attempts.
+            rendererTypes.Remove(configRenderer.Value);
+            rendererTypes.Insert(0, configRenderer.Value);
 
-                foreach (RendererType type in rendererTypes)
+            foreach (RendererType type in rendererTypes)
+            {
+                try
                 {
-                    if (type == RendererType.Automatic)
-                        continue;
+                    if (type == RendererType.OpenGLLegacy)
+                        SetupRendererAndWindow("gl", GraphicsSurfaceType.OpenGL);
+                    else
+                        SetupRendererAndWindow("veldrid", graphicsSurfaceForRenderer(type));
 
-                    try
-                    {
-                        SetupRendererAndWindow("veldrid", rendererToGraphicsSurfaceType(type));
-                        ResolvedRenderer = type;
-                        return;
-                    }
-                    catch
+                    ResolvedRenderer = type;
+                    return;
+                }
+                catch
+                {
+                    if (configRenderer.Value != RendererType.Automatic)
                     {
-                        if (configRenderer.Value != RendererType.Automatic)
-                        {
-                            // If we fail, assume the user may have had a custom setting and switch it back to automatic.
-                            Logger.Log($"The selected renderer ({configRenderer.Value.GetDescription()}) failed to initialise. Renderer selection has been reverted to automatic.",
-                                level: LogLevel.Important);
-                            configRenderer.Value = RendererType.Automatic;
-                        }
+                        // If we fail, assume the user may have had a custom setting and switch it back to automatic.
+                        Logger.Log($"The selected renderer ({configRenderer.Value.GetDescription()}) failed to initialise. Renderer selection has been reverted to automatic.",
+                            level: LogLevel.Important);
+                        configRenderer.Value = RendererType.Automatic;
                     }
                 }
             }
-
-            // fallback to legacy renderer. this is basically guaranteed to support all platforms.
-            SetupRendererAndWindow("gl", GraphicsSurfaceType.OpenGL);
-            ResolvedRenderer = RendererType.OpenGLLegacy;
         }
 
-        private static GraphicsSurfaceType rendererToGraphicsSurfaceType(RendererType renderer)
+        private static GraphicsSurfaceType graphicsSurfaceForRenderer(RendererType renderer)
         {
             GraphicsSurfaceType surface;
 
```

That way only one method is exposed and it lists exactly the renderer types that are attempted initialisation on by o!f in correct order.

osu.Framework/Platform/GameHost.cs
```diff
@@ -814,6 +795,221 @@ public void Run(Game game)
             }
         }
 
+        /// <summary>
+        /// The renderer which the game host is currently running with.
+        /// </summary>
+        public RendererType ResolvedRenderer { get; private set; }
```
frenzibyte, line 801: At first glance I thought this is just a duplicate of `IGraphicsSurface.Type`, until I realised it's of a different enum. Probably worth mentioning in remarks section:
```suggestion
        /// <remarks>
        /// This is similar to <see cref="IGraphicsSurface.Type"/> except that this is expressed as a <see cref="RendererType"/> rather than a <see cref="GraphicsSurfaceType"/>.
        /// </remarks>
        public RendererType ResolvedRenderer { get; private set; }
```

osu.Framework/Platform/GameHost.cs
```diff
@@ -1130,9 +1326,14 @@ private void setVSyncMode()
...
-        protected abstract IEnumerable<InputHandler> CreateAvailableInputHandlers();
+        protected abstract IEnumerable<InputHandler>
+            CreateAvailableInputHandlers();
 
-        public ImmutableArray<InputHandler> AvailableInputHandlers { get; private set; }
+        public ImmutableArray<InputHandler> AvailableInputHandlers
+        {
+            get;
+            private set;
+        }
```
frenzibyte, line 1336: No idea what happened here 😅

osu.Framework/Platform/GameHost.cs
peppy: Or maybe we can just use `Type.ToString()`? It's still giving equal information, just in a slightly different format.

osu.Framework/Platform/GameHost.cs
peppy: I've ended up going for something similar to your proposal, see how it reads.

frenzibyte: Reads pretty well 👍🏻

frenzibyte (approved):

osu.Framework/Platform/GameHost.cs
smoogipoo: What I mean by disposal is if `SetupWindow()` fails then the exception is caught at a higher level and the catch below doesn't call `Dispose()` on it. I've fixed it.

smoogipoo (approved):

PR #5684, opened 2023-03-17
Fix Veldrid framebuffers leaking GPU memory

smoogipoo:
I noticed this when entering and exiting song select, which would retain about 100MB VRAM every time.

Unlike with GL, resizing Veldrid FBOs requires them to be recreated. This PR is a three part fix:
- It was recreating them but not disposing the old ones. 
- It was not disposing the depth texture, which would likely have exacerbated this issue during gameplay (sliders).
- It was not rebinding the FBO upon such recreation. ~~I'm still evaluating whether this was the cause of other brokenness I was seeing.~~ (it is not)

The changes are hard to read since this is practically a total refactoring of the flow of this class.

### DIFF
--- a/osu.Framework/Graphics/Veldrid/Buffers/VeldridFrameBuffer.cs
+++ b/osu.Framework/Graphics/Veldrid/Buffers/VeldridFrameBuffer.cs
@@ -2,26 +2,29 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


frenzibyte (approved):

smoogipoo:
I explicitly put the initialisation of the colour target and texture inside that method so that everything's in the one place. But okay I suppose.

frenzibyte:
The member nullability attribute felt too long in the method so I thought moving it to constructor would be better.

PR #5687, opened 2023-03-19
Enforce presence of all fragment shader inputs in Veldrid

frenzibyte:
- Reverts https://github.com/ppy/osu-framework/commit/53272d8cafc35bb60c7e9ee40f17f41413e92435 in favour of this PR
- Can also revert https://github.com/ppy/osu-resources/pull/249 & https://github.com/ppy/osu/pull/22883

This works around D3D11 not working on shaders that don't make use of all members in the fragment input / vertex output. It seems D3D11 naturally expects all vertex output members are used in the fragment shader, and its optimisation of culling out unused fragment input members seems to affect the vertex output / fragment input structure.

One solution was to design localised vertex structures that match the needs of the fragment shader, but that turned out to be not pretty as time went and more shaders required this workaround.

This solution, on the other hand, prevents D3D11 from culling out unused fragment input by adding each of them via regex and passing them in the output structure.

Tested to work with Parallels, with no major differences performance-wise.


osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
```diff
@@ -109,25 +117,65 @@ private string loadFile(byte[]? bytes, bool mainFile)
...
+        /// </para>
+        /// </remarks>
+        /// <param name="vertexOutput">The list of vertex output members.</param>
+        public VeldridShaderPart WithFragmentOutput(IEnumerable<VeldridShaderAttribute> vertexOutput)
+        {
+            Debug.Assert(Type == ShaderPartType.Fragment);
+
+            string fragmentOutputCode = code;
+
+            int fragmentOutputLayoutIndex = shader_output_pattern.Matches(fragmentOutputCode).Count;
```
smoogipoo, line 157: Isn't this the same as the `Output` field? Also I don't think `Count` is generally correct though for our common cases it will be - any reason for not using `Max(Output.Location)`?

osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
```diff
@@ -109,25 +117,65 @@ private string loadFile(byte[]? bytes, bool mainFile)
...
+            var fragmentOutputAssignment = new StringBuilder();
+
+            var unusedFragmentInput = vertexOutput.Where(a => !fragmentOutputCode.Contains(a.Name)).ToList();
+
+            foreach (VeldridShaderAttribute attribute in unusedFragmentInput)
+            {
+                fragmentOutputLayout.AppendLine($"layout (location = {attribute.Location}) in {attribute.Type} {attribute.Name};");
+                fragmentOutputLayout.AppendLine($"layout (location = {fragmentOutputLayoutIndex++}) out {attribute.Type} o_{attribute.Name};");
+
+                fragmentOutputAssignment.AppendLine($"o_{attribute.Name} = {attribute.Name};");
```
smoogipoo, line 169: I would probably give this a GUID. Something like `unused_{Guid.NewGuid().ToString("N")}`.

smoogipoo (changes requested):
Seems okay, just a few nits.

osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
frenzibyte: Good catch, didn't think this one properly.

smoogipoo:
> Must revert https://github.com/ppy/osu-resources/pull/249

Can you explain why it "must" be reverted? As far as I understand this change should work regardless?

frenzibyte:
Since the PR depends on osu!-side changes, it's better to revert it and close the osu!-side PR rather than the other way around.

Now that I look at it though, I thought the changes required osu! to use a different vertex type, but it seems the only changes required are just using the new vertex shader, so it's not that big of a deal.

peppy:
What's the status of this PR? Might be worth revisiting and closing if deemed unnecessary at this point, or bringing up-to-date.

frenzibyte:
It's still valid for merge and would avoid all the vertex shaders that were created specifically to work around this (e.g. `sh_Blur.vs` and `sh_LogoAnimation.vs`)

peppy:
👍 

Needs some conflict resolution at very least then

frenzibyte:
Have updated this now, but do note that I haven't tested it on DirectX because compiling and running on a VM is kind of a pain for me. Testing and making sure blurred framebuffers renders correctly on DirectX would be appreciated.

osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
```diff
@@ -92,42 +102,82 @@ private string loadFile(byte[]? bytes, bool mainFile)
...
+            Debug.Assert(Type == ShaderPartType.Fragment);
+
+            string fragmentOutputCode = code;
+
+            int fragmentOutputLayoutIndex = Output.Max(m => m.Location) + 1;
+
+            var fragmentOutputLayout = new StringBuilder();
+            var fragmentOutputAssignment = new StringBuilder();
+
+            var unusedFragmentInput = vertexOutput.Where(a => !fragmentOutputCode.Contains(a.Name)).ToList();
```
smoogipoo, line 159: Is there a reason for using name rather than layout index? Something like `vertexOutput.Where(vo => !Input.Any(fi => fi.Location == vo.Location))`.

Also sticking with the above naming, this would be `unusedVertexOutputs`.

osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
```diff
@@ -92,42 +102,82 @@ private string loadFile(byte[]? bytes, bool mainFile)
...
+        /// <para>
+        /// In D3D11, unused fragment inputs cull their corresponding vertex output, which affects the vertex input/output structure leading to seemingly undefined behaviour.
+        /// To prevent this from happening, make sure all unused vertex output are used and sent in the fragment output so that D3D11 doesn't omit them.
+        /// </para>
+        /// <para>
+        /// This creates a new <see cref="VeldridShaderPart"/> rather than altering this existing instance since this is cached at a <see cref="IShaderStore"/> level and should remain immutable.
+        /// </para>
+        /// </remarks>
+        /// <param name="vertexOutput">The list of vertex output members.</param>
+        public VeldridShaderPart WithFragmentOutput(IEnumerable<VeldridShaderAttribute> vertexOutput)
```
smoogipoo, line 148: How about either `WithVertexOutput()` or  `WithPassthroughInput(IEnumerable<VeldridShaderAttribute> input)`?

osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
```diff
@@ -92,42 +102,82 @@ private string loadFile(byte[]? bytes, bool mainFile)
...
+
+            var unusedFragmentInput = vertexOutput.Where(a => !fragmentOutputCode.Contains(a.Name)).ToList();
+
+            foreach (VeldridShaderAttribute attribute in unusedFragmentInput)
+            {
+                string name = $"unused_input_{Guid.NewGuid():N}";
+
+                fragmentOutputLayout.AppendLine($"layout (location = {attribute.Location}) in {attribute.Type} {name};");
+                fragmentOutputLayout.AppendLine($"layout (location = {fragmentOutputLayoutIndex++}) out {attribute.Type} o_{name};");
+                fragmentOutputAssignment.Append($"o_{name} = {name};\n    ");
```
smoogipoo, line 167: Would probably change to `__unused_o_{name}` or something like that, just to prevent potential naming conflicts.

osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
```diff
@@ -13,45 +16,52 @@ namespace osu.Framework.Graphics.Veldrid.Shaders
...
         public readonly ShaderPartType Type;
 
-        private readonly List<string> shaderCodes = new List<string>();
+        private string header = string.Empty;
+
+        private readonly string code;
         private readonly IShaderStore store;
 
+        public IReadOnlyList<VeldridShaderAttribute> Input { get; private set; } = new List<VeldridShaderAttribute>();
+        public IReadOnlyList<VeldridShaderAttribute> Output { get; private set; } = new List<VeldridShaderAttribute>();
```
smoogipoo, line 32: I feel like `Inputs` / `Outputs` works better but maybe this is okay.

osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
```diff
@@ -92,42 +102,82 @@ private string loadFile(byte[]? bytes, bool mainFile)
...
+                fragmentOutputAssignment.Append($"o_{name} = {name};\n    ");
+            }
+
+            fragmentOutputCode = fragmentOutputCode.Replace("{{ fragment_output_layout }}", fragmentOutputLayout.ToString().Trim());
+            fragmentOutputCode = fragmentOutputCode.Replace("{{ fragment_output_assignment }}", fragmentOutputAssignment.ToString().Trim());
+
+            return new VeldridShaderPart(fragmentOutputCode, header, Type, store)
+            {
+                Input = Input,
+                Output = Output
```
smoogipoo, line 176: The assigned lists don't have the added inputs/outputs, probably best to clone the lists and add them?

smoogipoo (changes requested):

osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
frenzibyte: `WithPassthroughInput` sounds better, I'll go with that one.

osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
frenzibyte: Hmm, since it's countable let's go with `Inputs`/`Outputs`.

osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
frenzibyte: I don't quite follow what you mean.

smoogipoo: You're adding new inputs and outputs via `attributesLayout`. Those aren't included in the returned lists here.

smoogipoo (approved):

PR #5689, opened 2023-03-20
Fix draw thread unnecessarily marking GL context as current

### LINKED ISSUE #5688: Various `wglMakeCurrent()` errors are spammed to log when using Veldrid / OpenGL
I know that Veldrid+OpenGL is not a priority, so I'm reporting for visibility.
Tested on `visual-tests`, and it looks fine, so I'm not sure what the errors are about.

The only three that appear (the second and third mostly seem to appear when the window is obscured or minimized):

```cs
2023-03-19 15:00:05 [verbose]: SDL error log [debug]: wglMakeCurrent(): The requested resource is in use.
2023-03-19 15:00:05 [verbose]: SDL error log [debug]: wglMakeCurrent(): The handle is invalid.
2023-03-19 15:00:05 [verbose]: SDL error log [debug]: wglMakeCurrent(): The requested transformation operation is not supported.
```

Breakpointing the errors leads to:
https://github.com/ppy/osu-framework/blob/e90e308c9b8e34f37890ad40a1da646db5aa6086/osu.Framework/Platform/SDL2/SDL2GraphicsSurface.cs#L176

Full logs: [runtime.log](https://github.com/ppy/osu-framework/files/11011443/runtime.log)

frenzibyte:
- Closes #5688 

I'll preface this by saying this only affects Veldrid/OpenGL, so this doesn't need immediate attention at all and can be left for another release if preferred.

`MakeCurrent` calls are supposed to happen internally within the renderer whenever it's expecting to execute GL commands. In Veldrid, all GL commands are executed in a separate thread, therefore calling `MakeCurrent` here is unnecessary and will fail due to being acquired in the Veldrid thread.

I've kept the `ClearCurrent` call though since I'm not entirely sure of a better place for it that exists on `GLRenderer`, and I'm slightly skeptical of doing it in `FinishFrame`.

### DIFF
--- a/osu.Framework/Graphics/OpenGL/GLRenderer.cs
+++ b/osu.Framework/Graphics/OpenGL/GLRenderer.cs
@@ -103,6 +103,9 @@ namespace osu.Framework.Graphics.OpenGL
             lastBoundBuffers.AsSpan().Clear();
             lastBoundVertexArray = 0;
 
+            // Seems to be required on some drivers as the context is lost from the draw thread.
+            MakeCurrent();
+
             GL.UseProgram(0);
 
             base.BeginFrame(windowSize);
--- a/osu.Framework/Threading/DrawThread.cs
+++ b/osu.Framework/Threading/DrawThread.cs
@@ -28,8 +28,6 @@ namespace osu.Framework.Threading
 
             if (window != null)
             {
-                host.Renderer.MakeCurrent();
-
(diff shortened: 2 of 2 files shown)


Susko3 (approved):
Tested to work fine on `Veldrid/OpenGL` and `WindowsGL/OpenGL`, on both single- and multithreaded. And on `Android/OpenGL` for good measure.

smoogipoo (approved):

PR #5690, opened 2023-03-20
Fix D3D VBOs sometimes initialised with junk data

smoogipoo:
I noticed that every once in a while certain vertices would get incorrect data.

Furthermore this would always occur on vertices that were 0-initialised, which is _always_ the case with exactly one vertex in FBOs since the FtB pass doesn't do anything there (depth = 0). That explains why the issues I was seeing was always when `BufferedContainer` + blur were involved (e.g. backgrounds, in-gameplay break overlay arrows, song select leaderboard glows).

This makes it so that we always perform one buffer update with 0-initialised data when initialising a VBO. In `GLRenderer` we are relying on `glBufferData` to both initialise the storage for us _and_ clean out the driver/GPU buffer, so this was just missed in the port to Veldrid.

### DIFF
--- a/osu.Framework/Graphics/OpenGL/Buffers/GLVertexBuffer.cs
+++ b/osu.Framework/Graphics/OpenGL/Buffers/GLVertexBuffer.cs
@@ -69,7 +69,7 @@ namespace osu.Framework.Graphics.OpenGL.Buffers
 
             vboId = GL.GenBuffer();
             GL.BindBuffer(BufferTarget.ArrayBuffer, vboId);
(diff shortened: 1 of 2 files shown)


peppy (dismissed):

smoogipoo:
The documentation of [`glBufferData`](https://registry.khronos.org/OpenGL-Refpages/gl4/html/glBufferData.xhtml) lists a similar thing to D3D:
```
 If data is NULL, a data store of the specified size is still created, but its contents remain uninitialized and thus undefined. 
```
Which could explain some [issues](https://github.com/ppy/osu/issues/11136) we've had in the past. I've updated this PR to include a similar fix for the `GLRenderer` implementation.

peppy (approved):

PR #5693, opened 2023-03-21
Switch to using custom veldrid fork

peppy:
Contains:

- https://github.com/mellinoe/veldrid/pull/482
- https://github.com/mellinoe/veldrid/pull/481
- https://github.com/mellinoe/veldrid/pull/483


smoogipoo:
The other change I’ve got is https://github.com/mellinoe/veldrid/pull/483 . Still looking into exclusive fullscreen but I’m not sure on the status of it yet.

peppy:
> The other change I’ve got is [mellinoe/veldrid#483](https://github.com/mellinoe/veldrid/pull/483) . Still looking into exclusive fullscreen but I’m not sure on the status of it yet.

Added in 02d8ae3!

smoogipoo (approved):

PR #5698, opened 2023-03-22
Migrate iOS framework to SDL

frenzibyte:
- [x] Depends on https://github.com/ppy/osu-framework/pull/5697

This PR completely moves away from osuTK/OpenTK and starts using SDL, simplifying the iOS project codebase immensely and allowing to have Veldrid/Metal renderer support on iOS. I've organised the changes per commit since the diff is quite large, therefore reviewing per commit is highly recommended here.

There are changes applied for the implementations to remain AOT-compliant, despite the interpreter being enabled. These were added before I've enabled the interpreter, and can be reverted, but I've kept them as-is for better performance if any.

### iOS project structure has been revamped

In efforts to migrate the framework to SDL on iOS, the project structure required major changes. For consumers, you're required to remove the `AppDelegate` class attached in the game project as `GameAppDelegate` no longer exists, and change the `Application.Main` method to call `GameApplication.Main` as such:
```diff
-        public static void Main(string[] args) => UIApplication.Main(args, null, typeof(AppDelegate));
+        public static void Main(string[] args) => GameApplication.Main(new MyGameIOS());
```


osu.Framework/Platform/SDL2Window_Input.cs
```diff
@@ -234,8 +234,8 @@ private void handleTouchFingerEvent(SDL.SDL_TouchFingerEvent evtTfinger)
             if (existingSource == null)
                 return;
 
-            float x = evtTfinger.x * Size.Width;
-            float y = evtTfinger.y * Size.Height;
+            float x = evtTfinger.x * ClientSize.Width;
+            float y = evtTfinger.y * ClientSize.Height;
```
bdach, line 238: If anything worries me slightly it's this change. I'm trusting that it's correct but it's the sort of thing that can potentially break cross-platform and I can't test because I don't have requisite hardware to test (such as a touch-capable windows laptop or something).

bdach (approved):
Reviewing mostly on structure / code basis, can't test iOS, haven't tested much else (but I don't really expect breakage out of these changes).

osu.Framework/Platform/SDL2Window_Input.cs
Susko3: This is correct per https://discord.com/channels/188630481301012481/589331078574112768/1079101343937085540.

The size here should match the size of the `UserInputManager`.

bdach (approved):

osu.Framework/Platform/SDL2Window_Windowing.cs
```diff
@@ -130,12 +130,21 @@ private set
...
+        protected virtual IEnumerable<WindowMode> DefaultSupportedWindowModes
+        {
+            get
+            {
+                if (RuntimeInfo.IsMobile)
+                    return new[] { Configuration.WindowMode.Fullscreen };
+
+                return Enum.GetValues<WindowMode>();
+            }
+        }
```
Susko3, line 147: Nit: could be made `abstract` and defined for desktop and mobile separately.

Susko3 (approved):
Reviewing on code structure. Fine for initial support, but more of the desktop-specific code should be moved out of `SDL2Window` in a future PR.

osu.Framework/Platform/SDL2Window_Windowing.cs
bdach: I actually kinda prefer this to more inheritance. We'll see how we're going to get sdl running on android anyways because I'm pretty sure it's not going to be as simple as this PR.

Issue #5694, opened 2023-03-21
Buffered containers render with wrong colours on Android


bdach:
Found this when doing otherwise unrelated testing. I will probably be looking at this more tomorrow as well, but just opening for awareness/tracking/maybe someone else who's more up to date with this renderer stuff can take over.

To show what I mean, apply the following diff on current `master`:

```diff
diff --git a/osu.Framework.Tests/Visual/Containers/TestSceneFrontToBackBufferedContainer.cs b/osu.Framework.Tests/Visual/Containers/TestSceneFrontToBackBufferedContainer.cs
index 4388fd54d..67cce9d04 100644
--- a/osu.Framework.Tests/Visual/Containers/TestSceneFrontToBackBufferedContainer.cs
+++ b/osu.Framework.Tests/Visual/Containers/TestSceneFrontToBackBufferedContainer.cs
@@ -39,6 +39,31 @@ public void TestBufferedContainerBehindBox()
             });
         }
 
+        [Test]
+        public void TestBrokenColours()
+        {
+            AddStep("set children", () =>
+            {
+                Children = new Drawable[]
+                {
+                    new Box
+                    {
+                        Anchor = Anchor.Centre,
+                        Origin = Anchor.BottomCentre,
+                        Colour = Color4.Orange,
+                        Size = new Vector2(200)
+                    },
+                    new Box
+                    {
+                        Anchor = Anchor.Centre,
+                        Origin = Anchor.Centre,
+                        Colour = Color4.SlateGray,
+                        Size = new Vector2(300),
+                    },
+                };
+            });
+        }
+
         [Test]
         public void TestBufferedContainerAboveBox()
         {

```

then open `TestBufferedContainerBehindBox`, and compare the visual appearance of `BufferedContainerBehindBox` and `BrokenColours`:

| `BufferedContainerBehindBox` | `BrokenColours` |
| :-: | :-: |
| ![Screenshot_2023-03-21-22-47-37-34_fca46a2731136e673b8638a972bc0192](https://user-images.githubusercontent.com/20418176/226749577-d200d0c3-341e-41e1-9b76-27fcda3f8b39.jpg) | ![Screenshot_2023-03-21-22-47-41-39_fca46a2731136e673b8638a972bc0192](https://user-images.githubusercontent.com/20418176/226749629-376e7f80-4995-448c-9683-c50e1b23ac93.jpg) |

[...]

PR #5700, opened 2023-03-23
Add workaround for incorrect SRGB conversion inside `BufferedContainer`s on android

peppy:
See https://github.com/ppy/osu-framework/issues/5694.

Not closing the issue because I don't see this as a permanent fix. That said, maybe it won't be required if we switch to veldrid GL and it works there? Not even testing that at this point because I just want to ensure we can get the next release out on android without everything toppling over.

Tested on my only test device to work as expected.


bdach (approved):
works fine here

Susko3 (approved):

Issue #5701, opened 2023-03-23
Blur direction changes between d3d11 and opengl surfaces


bdach:
This can be seen already on `TestSceneEffects`, but can be made more visible by applying the following diff:

```diff
diff --git a/osu.Framework.Tests/Visual/Drawables/TestSceneEffects.cs b/osu.Framework.Tests/Visual/Drawables/TestSceneEffects.cs
index 59c0d6499..267e18640 100644
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneEffects.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneEffects.cs
@@ -1,14 +1,14 @@
 ∩╗┐// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-#nullable disable
-
+using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Colour;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Effects;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
+using osu.Framework.Graphics.Textures;
 using osuTK;
 using osuTK.Graphics;
 
@@ -17,9 +17,10 @@ namespace osu.Framework.Tests.Visual.Drawables
     [System.ComponentModel.Description("implementing the IEffect interface")]
     public partial class TestSceneEffects : FrameworkTestScene
     {
-        private readonly SpriteText changeColourText;
+        private SpriteText changeColourText = null!;
 
-        public TestSceneEffects()
+        [BackgroundDependencyLoader]
+        private void load(TextureStore textures)
         {
             var effect = new EdgeEffect
             {
@@ -234,6 +235,15 @@ public TestSceneEffects()
                             },
                         }
                     },
+                    new Sprite
+                    {
+                        Texture = textures.Get(@"sample-texture")
+                    }.WithEffect(new BlurEffect
+                    {
+                        Sigma = new Vector2(2f, 5f),
+                        Strength = 2f,
+                        Rotation = 45f,
+                    }),
                 }
             });
         }
```

| `OSU_GRAPHICS_SURFACE=OpenGL` | `OSU_GRAPHICS_SURFACE=Direct3D11` |
| :-: | :-: |
[...]

PR #5706, opened 2023-03-25
Fix blur direction changing between renderer surfaces

smoogipoo:
Fixes https://github.com/ppy/osu-framework/issues/5701

The core of the issue is that in OpenGL textures are inverted with the bottom-left texel being (0,0). Historically, we've always been rendering to FBOs such that their output is stored as if the top-left texel was (0,0), and then we had some logic to invert the FBO's texture when it was used for drawing.

So this is a two part fix:
1. Making FBOs rendered to as expected by the surface. For OpenGL if you view them in RenderDoc, they'll now be inverted. This makes all surfaces render the same way.
2. Removing the custom inversion logic for the blur direction. This only existed because the DrawNode assumed our incorrect rendering of FBOs. This makes the blur direction clockwise once again as expected.

@bdach Please double check that this fixes the issue.

### DIFF
--- a/osu.Framework/Graphics/Rendering/Renderer.cs
+++ b/osu.Framework/Graphics/Rendering/Renderer.cs
@@ -189,7 +189,8 @@ namespace osu.Framework.Graphics.Rendering
             globalUniformBuffer.Data = globalUniformBuffer.Data with
             {
                 IsDepthRangeZeroToOne = IsDepthRangeZeroToOne,
-                IsClipSpaceYInverted = IsClipSpaceYInverted
+                IsClipSpaceYInverted = IsClipSpaceYInverted,
+                IsUvOriginTopLeft = IsUvOriginTopLeft
             };
 
             Debug.Assert(defaultQuadBatch != null);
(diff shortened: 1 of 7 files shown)


bdach:
Can confirm that the issue looks fixed on this branch on my machine 👍 

I'll read through the code once again momentarily but I'm not sure I feel very comfortable with it as I'm seeing most of the stuff touched here for the first time, so I'll probably leave it up for a double-check anyways.

osu.Framework/Graphics/Rendering/Renderer.cs
```diff
@@ -1240,36 +1233,5 @@ static void checkAlignment(FieldInfo field, int offset, int expectedAlignment)
...
-            public ClipSpaceAdjustedTexture(Texture parent, WrapMode wrapModeS = WrapMode.None, WrapMode wrapModeT = WrapMode.None)
-                : base(parent, wrapModeS, wrapModeT)
-            {
-            }
-
-            public override RectangleF GetTextureRect(RectangleF? area = null)
-            {
-                return base.NativeTexture.Renderer.IsUvOriginTopLeft
-                    ? base.GetTextureRect(area)
-                    : base.GetTextureRect(area).Inflate(new Vector2(0, -1));
```
bdach, line 1269: This line is the only one from this diff that I cannot logically explain to myself - mostly because I don't understand what the previous code was doing. Why was one pixel being skimmed off the top and bottom of the texture rect here? Because - if I'm reading correctly - that's what `Inflate()` does, it doesn't flip the texture or anything - so I don't see the logical connection to the blur direction issue.

The fact that I don't know what this was doing means that I can't really begin to guess if this can be removed safely. You probably know better than I do, so it's more than likely correct, and I'm just mostly asking to educate myself I guess?

smoogipoo: In this case, the underlying texture rect (expressed as xywh) is `(0, 0, 1, 1)`, so this is actually inverting the texture. But it is a very roundabout way of doing so I agree.

It's implemented as
```
new RectangleF(
    X - amount.Left,
    Y - amount.Top,
    Width + amount.TotalHorizontal,
    Height + amount.TotalVertical);
```
So for the (0, 0, 1, 1) rect this becomes `(0 - 0, 0 - (-1), 1 + 0, 1 + (-1 + -1))` = `(0, 1, 1, -1)`

This is basically what was doing the final inversion when drawing the FBO texture to the screen, because the texture's contents were inverted from what OpenGL expected, we needed to invert it once again here.

bdach: Ohhh... right, makes sense 👍

bdach (approved):
Briefly tested d3d11+opengl+vulkan on windows, as well as legacy opengl on android - don't see anything going inverted unexpectedly, which I'd expect to see if anything was wrong. Double-check would be probably best though - I'm really not very confident around this rendering stuff as I've only just began actually paying attention to it.

PR #5710, opened 2023-03-27
Fix opening external links on Android

Susko3:
- Closes https://github.com/ppy/osu/issues/22993
- Alternative to https://github.com/ppy/osu/pull/23005
- Somewhat reverts #5666 

This PR follows the [guidelines](https://developer.android.com/training/package-visibility/use-cases#open-urls-browser-or-other-app) for apps targeting Android 11+.

The links added in the tests only fail if they are opened in their respective app (not the default web browser).

- https://github.com/ppy/osu/pull/22790 should be reverted after this is merged


Joehuu (approved):
This was my other solution minus the log (see https://github.com/ppy/osu/issues/22777#issuecomment-1454459145) before resulting to edit the manifest, so LGTM.

peppy (approved):

bdach (approved):
*shakes fist at android team for forcing us to rely on try-catching exceptions for application logic*

PR #5711, opened 2023-03-28
Make `ShaderManager.Load` `virtual` and add basic test coverage

peppy:


osu.Framework/Graphics/Shaders/ShaderManager.cs
```diff
@@ -39,8 +39,7 @@ public ShaderManager(IRenderer renderer, IResourceStore<byte[]> store)
         /// </summary>
         /// <param name="vertex">The vertex shader name.</param>
         /// <param name="fragment">The fragment shader name.</param>
-        /// <param name="continuousCompilation"></param>
-        public IShader Load(string vertex, string fragment, bool continuousCompilation = false)
```
bdach, line 43: Heh, this lived for a *while*. The param was there since initial framework commit but never did anything because it was only used in a commented `#if DEBUG` block that eventually got culled in https://github.com/ppy/osu-framework/commit/d6548a9c64234b4a6059cea79ccc2eb18b1a51ee (2017!). So looks pretty safe to remove and I don't think we even bother with a breaking change notice because nobody should ever care.

osu.Framework/Graphics/Veldrid/Shaders/VeldridShader.cs
```diff
@@ -108,6 +108,8 @@ public void BindUniformBlock(string blockName, IUniformBuffer buffer)
 
         private void compile()
         {
+            Logger.Log($"🖍️ Compiling shader {name}...");
```
bdach, line 111: Do we want this logged to the default target (runtime) and at the default level (verbose)? This means that it'll appear in the main `runtime.log` even in release builds, not sure if that's too noisy...

Could log at debug level, or to performance target. Or both. Dunno.

bdach (approved):

PR #5715, opened 2023-03-29
Fix srgb conversion

peppy:


peppy:
Not sure how this ended up in a PR. misclick?

PR #1409, opened 2018-02-22
WIP: iOS Compatibility (Framework-side)

swoolcock:
Work in progress getting osu!lazer to run on iOS devices.
Sound and touch input currently working, graphics are glitchy and occasionally freezes.

PR #5716, opened 2023-03-29
Remove iOS imagesharp AOT workaround

peppy:
I'm 99.9% sure we don't need this anymore. Originally added in https://github.com/ppy/osu-framework/pull/1409.


frenzibyte:
According to the original commit (https://github.com/ppy/osu-framework/commit/1c72acdc4b9c5fd90ae3333816e365fe99c91441), that logic was resolving runtime issues back then. Have thoroughly tested it and can confirm we don't need it anymore.

frenzibyte (approved):

PR #5717, opened 2023-03-29
Change `SupportedWindowModes` to non-bindable

peppy:
# Breaking changes

## `IWindow.SupportedWindowModes` type changed from `IBindableList<WindowMode>` to `IEnumerable<WindowMode>`

The property being bindable had mostly no use, as generally the list of supported modes will not change while the game is running, so there is no benefit to having it bindable.

Consumers that are broken by this change (due to using the property in bindable flow) can safely enumerate `SupportedWindowModes` at initialisation time without having to worry about the returned set of modes ever changing during game operation.

### DIFF
--- a/osu.Framework.Android/AndroidGameWindow.cs
+++ b/osu.Framework.Android/AndroidGameWindow.cs
@@ -53,7 +53,7 @@ namespace osu.Framework.Android
             SafeAreaPadding.BindTo(view.SafeAreaPadding);
         }
 
-        protected override IEnumerable<WindowMode> DefaultSupportedWindowModes => new[]
+        protected override IEnumerable<WindowMode> SupportedWindowModes => new[]
         {
             Configuration.WindowMode.Fullscreen,
         };
--- a/osu.Framework.Android/AndroidGameWindow.cs
+++ b/osu.Framework.Android/AndroidGameWindow.cs
@@ -53,7 +53,7 @@ namespace osu.Framework.Android
             SafeAreaPadding.BindTo(view.SafeAreaPadding);
         }
 
-        protected override IEnumerable<WindowMode> SupportedWindowModes => new[]
+        public override IEnumerable<WindowMode> SupportedWindowModes => new[]
         {
             Configuration.WindowMode.Fullscreen,
         };
--- a/osu.Framework/Platform/OsuTKWindow.cs
+++ b/osu.Framework/Platform/OsuTKWindow.cs
@@ -150,8 +150,6 @@ namespace osu.Framework.Platform
             MouseEnter += (_, _) => cursorInWindow.Value = true;
             MouseLeave += (_, _) => cursorInWindow.Value = false;
 
-            supportedWindowModes.AddRange(DefaultSupportedWindowModes);
-
             UpdateFrame += (_, _) => UpdateFrameScheduler.Update();
 
             graphicsSurface = new OsuTKGraphicsSurface(this);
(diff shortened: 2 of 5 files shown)


osu.Framework/Platform/OsuTKWindow.cs
```diff
@@ -257,9 +255,7 @@ private void checkCurrentDisplay()
         /// </summary>
         public virtual BindableSafeArea SafeAreaPadding { get; } = new BindableSafeArea();
 
-        private readonly BindableList<WindowMode> supportedWindowModes = new BindableList<WindowMode>();
-
-        public IBindableList<WindowMode> SupportedWindowModes => supportedWindowModes;
+        public IEnumerable<WindowMode> SupportedWindowModes => DefaultSupportedWindowModes;
```
bdach, line 258: If we're doing this (which I don't object to, for the record), do we even *need* `DefaultSupportedWindowModes` as separate a thing anymore? We could as well remove it and move the defaults to `SupportedWindowModes` instead.

osu.Framework/Platform/SDL2Window_Windowing.cs
```diff
@@ -223,7 +223,7 @@ public Size MaxSize
 
         public IBindable<bool> CursorInWindow => cursorInWindow;
 
-        public IBindableList<WindowMode> SupportedWindowModes { get; }
+        public IEnumerable<WindowMode> SupportedWindowModes { get; }
```
bdach, line 226: This is also a tad weird because the property is still backed by a `BindableList`:

https://github.com/ppy/osu-framework/blob/d586bb33ad4961d98db7c7f5a1acd7762977752b/osu.Framework/Platform/SDL2Window.cs#L192

osu.Framework/Platform/OsuTKWindow.cs
peppy: I think so, yeah.

bdach (approved):

osu.Framework.Android/AndroidGameWindow.cs
```diff
@@ -53,7 +53,7 @@ public override void SetupWindow(FrameworkConfigManager config)
             SafeAreaPadding.BindTo(view.SafeAreaPadding);
         }
 
-        protected override IEnumerable<WindowMode> DefaultSupportedWindowModes => new[]
+        protected override IEnumerable<WindowMode> SupportedWindowModes => new[]
```
Susko3, line 56:
```suggestion
        public override IEnumerable<WindowMode> SupportedWindowModes => new[]
```

Susko3 (changes requested):
Android needs attention. Fine otherwise.

osu.Framework.Android/AndroidGameWindow.cs
bdach: I'd seen the android change and presumed it was fine but missed the access modifiier >_<

bdach:
I've checked other projects via github search. [Some](https://github.com/MiraiSubject/KanojoWorks/blob/b17e8451558ef92e37d71796b5702ba81a6f95d1/KanojoWorks/Overlays/Settings/VideoSection.cs#L38) [consumers](https://github.com/pluckpawz/vignette/blob/cf6a286a8f45d4e05b07c8ca32a572a4b439b8ae/Vignette.Game/Settings/Sections/SystemSection.cs#L42) (and the game itself) will be broken by this, so I've included a breaking change notice in OP.

Still think this is OK to proceed with, though.

bdach (approved):

PR #5718, opened 2023-03-29
Fix Veldrid failing to initialise on Wayland

TheComputerGuy96:
wl_shell_surface is a part of wl_shell, which is a deprecated protocol and isn't supported by the majority of compositors today

I'm not sure why commit 37a3331473d37c9c895df746d94f379b28f90008 made this mistake (maybe SDL2 documentation wasn't clear about wl_shell?)

This mistake wasn't really relevant until recently, when Veldrid renderer started using this variable for creating a Wayland surface on Vulkan (which was always empty and so ended up crashing Mesa's Wayland Vulkan WSI code when creating a swapchain)

This change fixes Vulkan renderer crashes when SDL is using the Wayland video driver (I personally tested this on a debug build of osu! and it works)

### DIFF
--- a/osu.Framework/Platform/SDL2Window.cs
+++ b/osu.Framework/Platform/SDL2Window.cs
@@ -119,7 +119,7 @@ namespace osu.Framework.Platform
                         return wmInfo.info.uikit.window;
 
                     case SDL.SDL_SYSWM_TYPE.SDL_SYSWM_WAYLAND:
-                        return wmInfo.info.wl.shell_surface;
+                        return wmInfo.info.wl.surface;
 
                     case SDL.SDL_SYSWM_TYPE.SDL_SYSWM_ANDROID:
                         return wmInfo.info.android.window;


Susko3:
Could you provide relevant documentation for this change (deprecation notice, why `surface` is recommended, etc.) and the log files of the crash you're describing?

Does this crash on OpenGL or OpenGL (Legacy)?

TheComputerGuy96:
@Susko3
1) https://wayland.freedesktop.org/docs/html/apa.html#protocol-spec-wl_shell mentions the protocol deprecation

And here are all the major compositors that dropped it due to its deprecation:
GNOME: https://gitlab.gnome.org/GNOME/mutter/-/commit/a8bf9d8c6894d8e7618a80b010f5901346fcacd8
KDE: https://invent.kde.org/plasma/kwin/-/commit/4f4f4255381b21cc9c3d3c07c27e3bcb59c71609
Sway: https://github.com/swaywm/sway/commit/95a10dd4f35c6ffc14d23fa3a19a784e85a31724
Weston: https://gitlab.freedesktop.org/wayland/weston/-/commit/e6b8f5a5e40cd6c0b934e8ae079c86d5193efa96

2) I'm sure what the problem is, but here's a gdb snippet anyway: https://pastebin.com/srQ4j2pX

3) It obviously doesn't (OpenGL doesn't use `WindowHandle` at all and all renderers work in my personal testing with this change applied)

frenzibyte (dismissed):

frenzibyte (approved):

frenzibyte:
> I had to rebase the PR to not be out-of-date with recent changes, which means I lost the approval 🐸

As a side note, there's absolutely no need in force-pushing/rebasing PRs. Just merging master branch is fine.

PR #5721, opened 2023-03-30
Move gamma correction logic from renderer/shader to vertex input

frenzibyte:
- Closes https://github.com/ppy/osu/issues/22990
- Alternative to / closes https://github.com/ppy/osu-framework/pull/5713

We've generally been passing colours to the shaders in linear space, and convert the colours within the shader into sRGB space.

In Veldrid, this broke colour blending on buffered containers, as can be seen here:
image
(on the right is a box split into 50% and 100% opacity of the colour `#FF6633`, and on the left is the same but wrapped in a buffered container)

This breakage is essentially happening due to rendering happening in two stages and both the GPU driver and shader are incorrectly converting colour between sRGB and linear space until it ended up rendering the buffered container with a broken blend.

Rather than going through this fiasco, I've stopped both the renderer and the shader from ever applying any colour conversion in the shader, and I've instead let the framework pass colours to the shader in sRGB format.

The only difference this PR resulted in is `ColourInfo.Gradient{Horizontal,Vertical}` interpolating in sRGB space (gamma-incorrect), making it stand different from the other samples in the test scene:

image

But this matches both [browser](https://jsfiddle.net/uf621L3p/), and figma as well:

CleanShot 2023-03-30 at 03 20 23

So we're pretty much golden. I've updated the test scene to point as such. We can also consider replacing `Interpolation.ValueAt(Colour4)` with two different versions specifying in which space colours are being blended in, but I'll leave that for a follow-up PR since this one is quite stuffed.


osu.Framework/Graphics/Colour/SRGBColour.cs
```diff
@@ -59,8 +72,8 @@ public struct SRGBColour : IEquatable<SRGBColour>
                 first.Linear.A + second.Linear.A),
         };
 
-        public readonly Vector4 ToVector() => new Vector4(Linear.R, Linear.G, Linear.B, Linear.A);
-        public static SRGBColour FromVector(Vector4 v) => new SRGBColour { Linear = new Color4(v.X, v.Y, v.Z, v.W) };
+        public readonly Vector4 ToVector() => new Vector4(SRGB.R, SRGB.G, SRGB.B, SRGB.A);
```
peppy, line 75: The code quality failures here are likely worth considering. At very least we'd want to reduce the copy count to one by moving the property access to a local field, or at best avoid it being a computed property in the first place.

```suggestion
        public readonly Vector4 ToVector()
        {
            var linear = Linear;
            return new Vector4(linear.R, linear.G, linear.B, linear.A);
        }
```

Although I don't see this inspection locally so I'm not sure where it's coming from..

SampleGame/SampleGameGame.cs
bdach, line 1: I'm slightly weirded out that the display of `SampleGame` is being changed just for this pull. I'd have put the boxes in a test scene instead.

osu.Framework/Graphics/Colour/SRGBColour.cs
```diff
@@ -10,65 +10,89 @@ namespace osu.Framework.Graphics.Colour
 {
     /// <summary>
     /// A wrapper struct around Color4 that takes care of converting between sRGB and linear colour spaces.
-    /// Internally this struct stores the colour in linear space, which is exposed by the Linear member.
-    /// This struct implicitly converts to sRGB space Color4 values (i.e. it can be assigned and implicitly cast)
-    /// to sRGB Color4.
+    /// Internally this struct stores the colour in linear space, which is exposed by the <see cref="Linear"/> member.
+    /// This struct converts to sRGB space by using the <see cref="SRGB"/> member.
```
bdach, line 14: After https://github.com/ppy/osu-framework/pull/5721/commits/afa5ebc97f71a680ba0d5fa71e8972622e959d22 this is no longer correct.

```suggestion
    /// Internally this struct stores the colour in sRGB space, which is exposed by the <see cref="SRGB"/> member.
    /// This struct converts to linear space by using the <see cref="Linear"/> member.
```

osu.Framework/Graphics/Rendering/GlobalUniformData.cs
```diff
@@ -10,9 +10,8 @@ namespace osu.Framework.Graphics.Rendering
     [StructLayout(LayoutKind.Sequential, Pack = 1)]
     public record struct GlobalUniformData
     {
-        public UniformBool GammaCorrection;
         public UniformBool BackbufferDraw;
-        private readonly UniformPadding8 pad1;
+        private readonly UniformPadding12 pad1;
```
bdach, line 14: We have padding here and padding at the end of the global uniform block that comes out to 16 bytes total. Could/should we rearrange members here to consolidate the padding like [the other pull did](https://github.com/ppy/osu-framework/pull/5713/files#diff-1cd10e4e223070fbc73287aacb08684a05edd074e64f411ee595d2e5731fe4a2)?

SampleGame/SampleGameGame.cs
frenzibyte: Uh, this was definitely not intended.

osu.Framework/Graphics/Colour/SRGBColour.cs
frenzibyte: I guess it's more of which is prominent in the codebase now (in terms of direct usages). I didn't notice the operators were widely used though.

bdach: That said on second thought draw nodes are subject to invalidation, so the `ApplyChild()` calls should be 99% one-and-done init logic. So optimising for reads and avoiding `.ToLinear()` calls on the cpu *is* probably what we want.

osu.Framework.Tests/Visual/Drawables/TestSceneGammaCorrection.cs
```diff
@@ -0,0 +1,202 @@
...
+                        Size = new Vector2(750f, 50f),
+                        // Taken from https://observablehq.com/@d3/working-with-color
+                        Texture = textures.Get("d3-colour-interpolation"),
+                    },
+                    new SpriteText
+                    {
+                        Margin = new MarginPadding { Top = 20f },
+                        Anchor = Anchor.CentreLeft,
+                        Origin = Anchor.CentreLeft,
+                        Text = $"{nameof(Interpolation)}.{nameof(Interpolation.ValueAt)}(Red, Red) with 1px boxes",
```
Susko3, line 65:
```suggestion
                        Text = $"{nameof(Interpolation)}.{nameof(Interpolation.ValueAt)}(Red, Blue) with 1px boxes",
```

osu.Framework.Tests/Visual/Drawables/TestSceneGammaCorrection.cs
```diff
@@ -0,0 +1,202 @@
...
+                        Origin = Anchor.CentreLeft,
+                        Size = new Vector2(750f, 50f),
+                        Texture = textures.Get("figma-colour-interpolation"),
+                    },
+                    new SpriteText
+                    {
+                        Anchor = Anchor.CentreLeft,
+                        Origin = Anchor.CentreLeft,
+                        Margin = new MarginPadding { Top = 20f },
+                        Text = $"{nameof(ColourInfo)}.{nameof(ColourInfo.GradientHorizontal)}(Blue, Red)"
```
Susko3, line 106:
```suggestion
                        Text = $"{nameof(ColourInfo)}.{nameof(ColourInfo.GradientHorizontal)}(Red, Blue)"
```

Susko3 (changes requested):

peppy (approved):

smoogipoo (approved):

PR #5722, opened 2023-03-30
Fix buffered container children being clipped wrong on OpenGL

bdach:
- Closes https://github.com/ppy/osu/issues/22982
- Closes https://github.com/ppy/osu/issues/23009

Please treat with caution. While I tested that this does work, I only *kind of* know what I'm doing with this one.

Testing:

| Renderer/Platform | current master | this PR |
| :- | :-: | :-: |
| D3D11 @ Windows | d3d11_before | d3d11_after |
| OpenGL @ Windows | gl_before | gl_after |
| Vulkan @ Windows | vulkan_before | vulkan_after |
| OpenGL (legacy) @ Windows | gl_before | gl_after |
| OpenGL (legacy) @ Android | android_gl_before | android_gl_after |
| Metal @ macOS | CleanShot 2023-03-31 at 07 15 46 | CleanShot 2023-03-31 at 07 18 17 |

- [x] Probably could use a sanity check on Metal too (iOS/macOS).


osu.Framework/Graphics/Rendering/Renderer.cs
```diff
@@ -525,6 +525,12 @@ private void setScissor(RectangleI scissor)
                 scissor.Height = -scissor.Height;
             }
 
+            // compare: requiresFramebufferInvert flag in sh_Vertex_output.h.
+            if (!UsingBackbuffer && !IsUvOriginTopLeft)
+            {
+                scissor.Y = Viewport.Height - scissor.Bottom;
+            }
```
frenzibyte, line 532:
```suggestion
            // When rendering to a framebuffer on a backend with UV origin set to bottom-left (i.e. OpenGL), all vertex positions are flipped vertically in the shader (see sh_Vertex_Output.h).
            // In order for scissoring to still work with that behaviour in place, we need to flip the rectangle likewise.
            if (!UsingBackbuffer && !IsUvOriginTopLeft)
                scissor.Y = Viewport.Height - scissor.Bottom;
```

frenzibyte (commented):
This makes me uneasy about the general approach that the [regressing PR](https://github.com/ppy/osu-framework/pull/5706) took (which is flipping all vertices vertically when rendering to a framebuffer, as opposed to rendering the framebuffer texture upside-down). I know that it's to fix blur direction being different between graphics backends, but I haven't sat down to see why that's happening precisely, and the original PR's description doesn't seem to help in explaining that.

All that being said, I can't imagine another scenario being regressed here other than scissoring, so I guess we don't have to reconsider looking into the original change yet.

smoogipoo:
> which is flipping all vertices vertically when rendering to a framebuffer, as opposed to rendering the framebuffer texture upside-down

It's hard to explain without going into RenderDoc and seeing it for yourself, but the purpose of that PR was to standardise UV coordinates such their origin is always as expected by the renderer.

frenzibyte:
> If you flip UV coordinates, as we were doing previously, then you'd also need to flip any offsets on those coordinates (the blur direction). It's why we previously had the blur direction specified as a negative value for a clockwise direction, but I don't think that is a detail which the user should be concerned about.

Right, I've read through the blur logic and I understand now. Seems like it's a good direction then.

> That being said, I don't know if -gl_Position.y is the correct way of implementing it as opposed to something like viewport.Height - gl_Position.y.

Flipping the sign of the Y coordinate should essentially be the same as `viewport.Height - Y`, since the bounds of the NDC is mapped to the bounds of the viewport rectangle, as far as I understand.

osu.Framework/Graphics/Rendering/Renderer.cs
frenzibyte: In addition, this transformation should probably happen after `Scissor` is updated, since it's an internal detail that the consumer shouldn't notice when reading the `Scissor` property.

peppy (approved):
Fix does look solid and fix the issues mentioned.

frenzibyte (approved):

PR #5729, opened 2023-04-01
Add Veldrid screenshot support

frenzibyte:
- Closes https://github.com/ppy/osu/issues/22959
- [x] Depends on https://github.com/mellinoe/veldrid/pull/493 (rebased in https://github.com/ppy/veldrid/pull/7)
- [x] Depends on https://github.com/mellinoe/veldrid/pull/494

As per the issue thread, this follows a simple approach of reading backbuffer pixels by copying the swapchain texture to a staging buffer and mapping it for CPU read.

Tested to work on OpenGL, D3D11, and Metal. Vulkan may require further changes against Veldrid according to [an old PR that's achieving similar to what my PRs above achieve](https://github.com/mellinoe/veldrid/pull/363).

### DIFF
--- a/osu.Framework/Graphics/OpenGL/GLRenderer.cs
+++ b/osu.Framework/Graphics/OpenGL/GLRenderer.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Linq;
 using System.Text;
+using System.Runtime.InteropServices;
 using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Graphics.OpenGL.Buffers;
 using osu.Framework.Graphics.OpenGL.Textures;
@@ -18,7 +19,11 @@ using osu.Framework.Platform;
 using osu.Framework.Statistics;
 using osuTK;
 using osuTK.Graphics.ES30;
+using SixLabors.ImageSharp;
+using SixLabors.ImageSharp.Memory;
 using SixLabors.ImageSharp.PixelFormats;
+using SixLabors.ImageSharp.Processing;
+using Image = SixLabors.ImageSharp.Image;
 
 namespace osu.Framework.Graphics.OpenGL
 {
@@ -335,6 +340,18 @@ namespace osu.Framework.Graphics.OpenGL
             }
             else
                 GL.Disable(EnableCap.StencilTest);
+        }
+
+        public override Image<Rgba32> TakeScreenshot()
+        {
+            var size = ((IGraphicsSurface)openGLSurface).GetDrawableSize();
+            var data = MemoryAllocator.Default.Allocate<Rgba32>(size.Width * size.Height);
+
+            GL.ReadPixels(0, 0, size.Width, size.Height, PixelFormat.Rgba, PixelType.UnsignedByte, ref MemoryMarshal.GetReference(data.Memory.Span));
+
+            var image = Image.LoadPixelData<Rgba32>(data.Memory.Span, size.Width, size.Height);
+            image.Mutate(i => i.Flip(FlipMode.Vertical));
+            return image;
         }
 
         protected override IShaderPart CreateShaderPart(ShaderManager manager, string name, byte[]? rawData, ShaderPartType partType)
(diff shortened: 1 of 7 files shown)


bdach:
Both veldrid PRs you linked conflict with each other at `src/Veldrid/MTL/MTLSwapchain.cs`. Would probably be helpful to provide the correct merge conflict resolution somewhere.

bdach:
Upon testing d3d11 seems to have some sort of issue in windowed mode - looks like the stride is incorrect somewhere. Stale window dimensions perhaps.

| what it should look like (captured via snipping tool) | ingame screenshot |
| :-: | :-: |
| should_be | is |

bdach:
The following diff seems to fix the issue above:

```diff
diff --git a/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs b/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
index 6d94d072e..2952156dc 100644
--- a/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
+++ b/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
@@ -500,7 +500,10 @@ protected internal override unsafe Image<Rgba32> TakeScreenshot()
 
                 default:
                 {
-                    using var staging = Factory.CreateTexture(TextureDescription.Texture2D(texture.Width, texture.Height, 1, 1, texture.Format, TextureUsage.Staging));
+                    uint width = texture.Width;
+                    uint height = texture.Height;
+
+                    using var staging = Factory.CreateTexture(TextureDescription.Texture2D(width, height, 1, 1, texture.Format, TextureUsage.Staging));
                     using var commands = Factory.CreateCommandList();
                     using var fence = Factory.CreateFence(false);
 
@@ -513,11 +516,18 @@ protected internal override unsafe Image<Rgba32> TakeScreenshot()
                     var resource = Device.Map(staging, MapMode.Read);
                     var span = new Span<Bgra32>(resource.Data.ToPointer(), (int)(resource.SizeInBytes / Marshal.SizeOf<Bgra32>()));
 
-                    using var image = Image.LoadPixelData<Bgra32>(span, (int)staging.Width, (int)staging.Height);
+                    // on some backends (Direct3D11, in particular), the staging resource data may contain padding at the end of each row for alignment,
+                    // which means that for the image width, we cannot use the framebuffer's width raw.
+                    using var image = Image.LoadPixelData<Bgra32>(span, (int)(resource.RowPitch / Marshal.SizeOf<Bgra32>()), (int)height);
 
                     if (!Device.IsUvOriginTopLeft)
                         image.Mutate(i => i.Flip(FlipMode.Vertical));
 
+                    // if the image width doesn't match the framebuffer, it means that we still have padding at the end of each row mentioned above to get rid of.
+                    // snip it to get a clean image.
+                    if (image.Width != width)
+                        image.Mutate(i => i.Crop((int)width, (int)height));
+
                     Device.Unmap(staging);
 
                     return image.CloneAs<Rgba32>();
```

For rationale, see [remarks in documentation of `D3D11_SUBRESOURCE_DATA`](https://learn.microsoft.com/en-us/windows/win32/api/d3d11/ns-d3d11-d3d11_subresource_data#remarks), in particular this part:

> **Note**  An application must not rely on **SysMemPitch** being exactly equal to the number of texels in a line times the size of a texel. In some cases, **SysMemPitch** will include padding to skip past additional data in a line. This could be padding for alignment or the texture could be a subsection of a larger texture. For example, the **D3D11_SUBRESOURCE_DATA** structure could represent a 32 by 32 subsection of a 128 by 128 texture. The value for **SysMemSlicePitch** will reflect any padding included in **SysMemPitch**.

frenzibyte:
Good find, I've checked the change on Metal and also D3D11 via Parallels, and it's still working correctly.

peppy:
I've pushed an updated veldrid but haven't tested yet, any testing would be appreciated.

peppy:
Confirmed working on both cases here!

peppy (approved):

PR #5732, opened 2023-04-02
Fix non-legacy OpenGL sometimes flickering/freezing

frenzibyte:
Should resolve https://github.com/ppy/osu/issues/23088, can't test it right now.

When `setSwapchainFramebuffer` is non-null, Veldrid will always use it without falling back to `glBindFramebuffer(0)`. Therefore we should fall back to binding to framebuffer 0 ourselves rather than returning without binding anything.


Susko3 (approved):
Tested on osu! song select, fixes the flickering.

bdach (approved):

PR #5734, opened 2023-04-03
Improve performance of video texture uploads

smoogipoo:
Fixes https://github.com/ppy/osu/issues/22970

It's not the most that can be done, because Veldrid doesn't expose interfaces to do this sort of alignment, and although it seems possible in at least D3D, it's something I don't want to do unless absolutely necessary. This is already between 1-2 magnitudes faster for me, just in this one block nevermind GL execution time which is hard to profile.

I want confirmation that this fixes the issue before merge, though.

Also would be good to test this on mobile platforms.


bdach:
I can't test mobile (as android doesn't mix with veldrid yet), but desktop looks better to me.

| master | this PR |
| :-: | :-: |
| before | after |

bdach (approved):

peppy (approved):

Issue #5445, opened 2022-10-05
Non-endless transform loop performance needs improvement


peppy:
Right now, endless loops are very optimal, but cases of high loop counts (but not endless) result in huge amount of overhead.

Endless:

https://github.com/ppy/osu-framework/blob/ba1385330cc501f34937e08257e586c84e35d772/osu.Framework/Graphics/Transforms/TransformSequence.cs#L310-L354

Defined:

https://github.com/ppy/osu-framework/blob/ba1385330cc501f34937e08257e586c84e35d772/osu.Framework/Graphics/Transforms/TransformSequence.cs#L242-L272

In order to better support storyboards which have high loop counts, I would hope that we can optimise this to have a "remaining count" or similar rather than a full clone operation.

Would be used to fix https://github.com/ppy/osu/issues/12401#issuecomment-1261330336 (with a direct mapping of storyboard loop metadata as per WIP https://github.com/ppy/osu/compare/master...peppy:osu:storyboard-looping-loop?expand=1).

PR #5737, opened 2023-04-09
Change transform loop to use a counter

hwabis:
Fixes https://github.com/ppy/osu-framework/issues/5445

But even with these changes, the storyboard here https://github.com/ppy/osu/issues/12401#issue-857227725 still takes forever to load.


osu.Framework/Graphics/Transforms/Transform.cs
```diff
@@ -35,6 +35,10 @@ public abstract class Transform
 
         public bool IsLooping { get; internal set; }
```
smoogipoo, line 36: Is it possible to make this a `get`-only member, implemented like: `IsLooping => LoopCount > 0`, or is there a reason it's kept as a separate value?

smoogipoo (changes requested):
At a surface level, without having profiled to see the results, I think this looks fine. Will need a review from peppy because he's more in depth with transforms.

osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
```diff
@@ -164,7 +164,10 @@ public void UpdateTransforms(in double time, bool rewinding)
                             flushAppliedCache = true;
                         }
 
-                        if (t.IsLooping)
+                        if (t.LoopCount != -1)
+                            t.LoopCount--;
```
bdach, line 168: This PR breaks loops with finite counts. To reproduce, apply the following test:

```diff
diff --git a/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs b/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
index 023bd17f0..b2de05f53 100644
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneTransformRewinding.cs
@@ -343,6 +343,24 @@ public void LoopSequence()
             }
         }
 
+        [Test]
+        public void FiniteLoopSequence()
+        {
+            boxTest(box => { box.RotateTo(0).RotateTo(90, interval).Loop(0, 2); });
+
+            checkAtTime(0, box => box.Rotation == 0);
+            checkAtTime(interval - 1, box => Precision.AlmostEquals(box.Rotation, 90, 1));
+            checkAtTime(interval + 1, box => Precision.AlmostEquals(box.Rotation, 0, 1));
+            checkAtTime(2 * interval - 1, box => Precision.AlmostEquals(box.Rotation, 90, 1));
+
+            checkAtTime(3 * interval - 1, box => Precision.AlmostEquals(box.Rotation, 90, 1));
+
+            checkAtTime(2 * interval - 1, box => Precision.AlmostEquals(box.Rotation, 90, 1));
+            checkAtTime(interval + 1, box => Precision.AlmostEquals(box.Rotation, 0, 1));
+            checkAtTime(interval - 1, box => Precision.AlmostEquals(box.Rotation, 90, 1));
+            checkAtTime(0, box => box.Rotation == 0);
+        }
+
         [Test]
         public void StartInMiddleOfLoopSequence()
         {
```

Passes on `master`, fails on this branch. I strongly suspect (but haven't confirmed) this decrement to be incorrect without a corresponding increment when rewindable transforms are played back in the backward direction.

bdach (changes requested):

osu.Framework/Graphics/Transforms/TargetGroupingTransformTracker.cs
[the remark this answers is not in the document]
hwabis: Yes, applied

hwabis: Added the test and it's passing now.

I don't think it was about adding a `LoopCount` increment (although would it still make sense to have it somewhere?), but it was in `makeTransformsLooping`. (See the `pushForwardCount` part. Example: a transform list of `0-0ms 0->0deg, 0-250ms 0->90deg` makes `sortedTransforms` become `0-250ms 0->90deg, 250-250ms 0->0deg`, so we decrement the `LoopCount` for that shifted transform.)

osu.Framework/Graphics/Transforms/Transform.cs
hwabis: Applied

peppy:
I believe this looks quite correct now. Let's hope no one was using `LoopCount` and expecting it to be immutable from the user set value (probably unlikely).

peppy (approved):

peppy:
I'm gonna get this one in since I'm pretty confident. Will deal with any fallout after it's discovered.

peppy (approved):

PR #5739, opened 2023-04-14
Add ability to add additional stores to the shader manager

EVAST9919:
At first I was considering using `HashSet`, however `IResourceStore` isn't implementing it's own hash code, so the same store can't be recognised as already being added. Comparing strings provided by `GetavailableResources()` would be heavier, so `Dictionary` with a name string was my choice.
Not sure if I should use `ConcurrentDictionary` instead though.

### DIFF
--- a/osu.Framework/Graphics/Shaders/ShaderManager.cs
+++ b/osu.Framework/Graphics/Shaders/ShaderManager.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.IO;
 using osu.Framework.Graphics.Rendering;
 using osu.Framework.IO.Stores;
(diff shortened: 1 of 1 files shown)


smoogipoo:
I got some serious dejavu reading through this PR because I was working on a similar branch [here](https://github.com/smoogipoo/osu-framework/tree/shader-stores) and everything down to the title and first 5 words of this PR was exactly the same as I was about to post.

That aside, what prevented me from posting it is the issue of isolation. Because `ShaderManager` has a dual role of caching `IShaderPart`s and `IShader`s, there will be problems if, say, two rulesets define the lookup `flashlight/texture`, and either the two rulesets are played in sequence (there needs to be some cache busting logic) _or_ the two rulesets are present simultaneously such as in multiplayer (this method cannot handle that).

I'm continuing to work on this issue, but I don't think this change is a good path forward as it is.

EVAST9919:
@smoogipoo I'll leave it up to you then since I'm not sure I have anything valuable to propose here.

PR #5740, opened 2023-04-14
Refactor Veldrid VBO storages to speed up data upload commands on Metal

frenzibyte:
Initially, the VBO implementation was a straight-up copy pasta from `GLRenderer` and it was working well for the most part. Now, upon profiling on Metal, it turns out that every call to `Commands.UpdateBuffer` fires up a staging buffer (from a pool now) and calls `MTLBuffer.contents` which has some overhead when used in a hot path.

One way to avoid the frequent calls to `contents` would be to store a staging buffer and use `Commands.CopyBuffer` instead, which directly encodes a [`copy`](https://developer.apple.com/documentation/metal/mtlblitcommandencoder/1400767-copy) command without extra API calls.

Now, using `UpdateBuffer` on the staging buffer alone still won't avoid the overhead of `contents`, so I've mapped it to CPU using `Device.Map` (which is essentially just returning `contents`) and used the pointer directly for uploading vertices.

I've also removed the memory storage allocated with ImageSharp in favour of the staging buffer.

I had considered removing the staging buffer and directly mapping the vertex buffer itself since it's always created with "shared" storage mode, but that violates Veldrid's public API and even when bypassing that, it doesn't really add any more gains compared to the approach done here.

### DIFF
--- a/osu.Framework/Graphics/Veldrid/Buffers/VeldridLinearBuffer.cs
+++ b/osu.Framework/Graphics/Veldrid/Buffers/VeldridLinearBuffer.cs
@@ -15,13 +15,11 @@ namespace osu.Framework.Graphics.Veldrid.Buffers
         where T : unmanaged, IEquatable<T>, IVertex
     {
         private readonly VeldridRenderer renderer;
-        private readonly int amountVertices;
 
         internal VeldridLinearBuffer(VeldridRenderer renderer, int amountVertices, PrimitiveTopology type, BufferUsage usage)
             : base(renderer, amountVertices, usage)
         {
             this.renderer = renderer;
-            this.amountVertices = amountVertices;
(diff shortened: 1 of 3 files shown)


peppy:
Do you have any benchmark/profiling data for this one?

frenzibyte:
I'll admit I haven't profiled this enough, I've checked it again and I realised the overhead of `copy` got way too higher than usual, defeating the entire purpose of this PR...

| | |
|--------|-------|
| Before | CleanShot 2023-04-15 at 03 51 40 |
| After | CleanShot 2023-04-15 at 03 51 42 |

I'm not really sure what's going on there, I tried caching `MTLBuffer.contents()` but that also produced the same profiling data as above. Blocking this PR pending investigation.

PR #5744, opened 2023-04-15
Upload Veldrid VBOs directly rather than using a command list

frenzibyte:
- Alternative to / closes #5740 

On Metal, encoding a buffer copy command comes with an overhead, and that overhead becomes noticeable on frequently updated buffers (aka "dynamic").

Since we update our vertex buffers pretty much every single frame, it's wiser to update them directly from the CPU rather than encoding a copy command every time.

In addition, since we're updating vertex buffers directly from the CPU, we must ensure it shouldn't be read from the GPU. Simple triple-buffering logic should help with that, but note that it increases the memory usage to triple times its original size in return (maybe double-buffering is better?).

If the memory usage here turns out to have taken a bad turn, then this PR can be ignored for the time being (potentially wait for https://github.com/ppy/osu-framework/pull/5109).

Before:

CleanShot 2023-04-15 at 18 10 05

After:

CleanShot 2023-04-15 at 18 10 01

### DIFF
--- a/osu.Framework/Graphics/Veldrid/Batches/VeldridVertexBatch.cs
+++ b/osu.Framework/Graphics/Veldrid/Batches/VeldridVertexBatch.cs
@@ -13,7 +13,13 @@ namespace osu.Framework.Graphics.Veldrid.Batches
     internal abstract class VeldridVertexBatch<T> : IVertexBatch<T>
         where T : unmanaged, IEquatable<T>, IVertex
     {
-        public List<VeldridVertexBuffer<T>> VertexBuffers = new List<VeldridVertexBuffer<T>>();
+        private readonly List<VeldridVertexBuffer<T>>[] tripleBufferedVertexBuffers = new List<VeldridVertexBuffer<T>>[3];
+
+        public List<VeldridVertexBuffer<T>> CurrentVertexBuffers
+        {
+            get => tripleBufferedVertexBuffers[renderer.ResetId % 3];
+            set => tripleBufferedVertexBuffers[renderer.ResetId % 3] = value;
+        }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


peppy:
I think the memory usage is still within acceptable bounds. Also I'm 99.9% sure this one was discussed IRL with smoogi and seen as a probably-required direction. Let's get this in for now.

peppy (approved):

osu.Framework/Graphics/Veldrid/Batches/VeldridVertexBatch.cs
```diff
@@ -13,7 +13,13 @@ namespace osu.Framework.Graphics.Veldrid.Batches
...
         where T : unmanaged, IEquatable<T>, IVertex
     {
-        public List<VeldridVertexBuffer<T>> VertexBuffers = new List<VeldridVertexBuffer<T>>();
+        private readonly List<VeldridVertexBuffer<T>>[] tripleBufferedVertexBuffers = new List<VeldridVertexBuffer<T>>[3];
+
+        public List<VeldridVertexBuffer<T>> CurrentVertexBuffers
+        {
+            get => tripleBufferedVertexBuffers[renderer.ResetId % 3];
+            set => tripleBufferedVertexBuffers[renderer.ResetId % 3] = value;
+        }
```
smoogipoo, line 22: Just note that if we're going to do this then we absolutely need to give drawnodes a "shared data" or something similar, as this is effectively creating 9 VBOs per Drawable.

frenzibyte: I had left UBOs to still use the command list as uploading to UBOs won't be as frequent as uploading to VBOs are once masking data is moved to a separate buffer.

PR #5745, opened 2023-04-16
Move FBO content population to a separate pass and enable "memoryless" storage mode on Metal

frenzibyte:
I'll preface this by mentioning that I don't seem to notice that much of a gain on my phone, but I'm PR'ing this anyway as it saves GPU memory bandwidth according to Xcode.

Until now, we've always been switching between framebuffers in the middle of a render pass, which was pretty much fine with OpenGL. But with Metal, switching a framebuffer requires ending the current render pass and beginning a new one, therefore unnecessarily increasing the number of render passes in a frame, potentially along with the bandwidth of loading and storing framebuffer attachments from/to tile memory.

This PR moves away any logic that involves binding a framebuffer to a separate pass that occurs before the main render pass, and enables using ["memoryless" storage mode](https://developer.apple.com/documentation/metal/mtlstoragemode/memoryless) for depth/stencil attachments on Metal (only affects Apple GPUs), therefore saving the bandwidth of unnecessarily storing depth/stencil content into system memory (which we were forced to do previously because `BufferedDrawNode`s were interrupting the main render pass, making the backbuffer lose its depth content and break depth testing).

### `IFrameBuffer.Bind`/`BufferedDrawNode.BindFramebuffer` is no longer available

For optimal performance, `IFrameBuffer`s should not be bound anywhere freely. Instead, we've defined a new render pass specific for populating frame buffers with content.

The new render pass iterates through `DrawNode.PopulateFrameBuffers()`. Consumers should override the method for binding frame buffers and populating them.

PR #5746, opened 2023-04-17
Disable "memoryless" storage mode for now and fix potential mipmap generation crash

frenzibyte:
As per https://github.com/ppy/osu-framework/pull/5745#issuecomment-1511063628, this disables back "memoryless" storage mode for the time being.

In addition, this also handles a [noticed edge case](https://github.com/ppy/osu-framework/pull/5745#issuecomment-1511053877) of textures generating mipmap levels in the middle of a render pass. Since we're no longer using "memoryless" storage mode, we can safely unbind the previous framebuffer without its depth/stencil content being lost.

### DIFF
--- a/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
+++ b/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
@@ -187,7 +187,7 @@ namespace osu.Framework.Graphics.Veldrid
                     break;
 
                 case GraphicsSurfaceType.Metal:
-                    Device = GraphicsDevice.CreateMetal(options, swapchain, new MetalDeviceOptions { PreferMemorylessDepthTargets = true });
+                    Device = GraphicsDevice.CreateMetal(options, swapchain);
                     Device.LogMetal(out maxTextureSize);
                     break;
             }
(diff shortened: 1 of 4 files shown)


osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
```diff
@@ -187,7 +187,7 @@ protected override void Initialise(IGraphicsSurface graphicsSurface)
                     break;
 
                 case GraphicsSurfaceType.Metal:
-                    Device = GraphicsDevice.CreateMetal(options, swapchain, new MetalDeviceOptions { PreferMemorylessDepthTargets = true });
+                    Device = GraphicsDevice.CreateMetal(options, swapchain);
```
smoogipoo, line 190: I'm going to go back on what I said here, let's keep memoryless for now and see where issues come up.

smoogipoo (changes requested):

osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
frenzibyte: I'll need @peppy to confirm whether unbinding previous framebuffers before generating mipmaps doesn't break rendering on his iPad (i.e. test this branch with https://github.com/ppy/osu-framework/pull/5746/commits/b8ec3b5ae3c5f0911360aae83bbd3a580447a382 reverted), but I'm not sure myself when does a texture upload data in the middle of a render pass, so I think it's better to just keep this disabled for now.

frenzibyte:
Closing this as the original PR has been reverted, I'll open back as a draft.

PR #5751, opened 2023-04-18
Update `veldrid-spirv` native library with lower iOS deployment target

frenzibyte:
- Resolves ppy/osu#23198

Recompiled on same branch but with a lower deployment target (13.4) in order to resolve https://github.com/ppy/osu/issues/23198. I've tried updating the simulator version as well but that didn't work for odd reason, so I've kept it as-is.

### DIFF
--- a/osu.Framework.iOS/runtimes/ios/native/veldrid-spirv.xcframework/ios-arm64/veldrid-spirv.framework/Info.plist
+++ b/osu.Framework.iOS/runtimes/ios/native/veldrid-spirv.xcframework/ios-arm64/veldrid-spirv.framework/Info.plist
@@ -43,7 +43,7 @@
 	<key>DTXcodeBuild</key>
 	<string>14C18</string>
 	<key>MinimumOSVersion</key>
-	<string>16.2</string>
+	<string>13.4</string>
 	<key>UIDeviceFamily</key>
 	<array>
 		<integer>1</integer>


RoootTheFox:
I can confirm that this fixes ppy/osu#23198 on iOS 14.2

F04FA8E0-10B9-43FA-9D2D-9BEAFB18456D


Small side notes:
- Simulators always use the latest iOS version/sdk afaik, installing an older simulator in XCode would probably help but I doubt it makes a difference anyway
- [Info.plist still has iOS 16.2 set as minimum version](https://github.com/frenzibyte/osu-framework/blob/ec974a65e4d8df6aa0d148835ed42f5ef18bc039/osu.Framework.iOS/runtimes/ios/native/veldrid-spirv.xcframework/ios-arm64/veldrid-spirv.framework/Info.plist#L36-L46), not sure if this should be changed since it *did* work with 16.2 set (on iOS 14.2), but I'd probably change it to 13.4 for consistency and just to be safe

frenzibyte:
Good to hear it's working, also nice catch, I've updated it.

peppy (approved):

RoootTheFox:
> Thank you for your help.
I am not very used with github etc, do I have to compile the binaries by myself to be able to run the game?
Thank you.

Here's a small step-by-step tutorial for iOS
You'll need the following things:
- Filza (file manager, I'll assume you already have that one installed considering you're jailbroken)
- packages for zipping and unzipping: `sudo apt install zip unzip` (probably also already installed)

Then:
- Download the .ipa for the latest osu! release (don't use TestFlight to install) and [this file](https://github.com/frenzibyte/osu-framework/raw/e7a51ae8505d2e271197ab5e69f06a5fd1442d9d/osu.Framework.iOS/runtimes/ios/native/veldrid-spirv.xcframework/ios-arm64/veldrid-spirv.framework/veldrid-spirv),
- Share the IPA to Filza and open Filza
- long hold the .ipa, click rename and add `.zip` at the end
- click on it once, this will extract it
- go into `Payload/osu.iOS.app/Frameworks/veldrid-spirv.framework` and replace the existing `veldrid-spirv` file in there with the one you downloaded
- go back to the folder where the .ipa is and delete it (you won't need it anymore)
- long hold the `Payload` folder, click "Compress" and append `.ipa` to the name of the resulting file
- install the new .ipa directly through Filza, TrollStore or whatever else you can use to sideload IPAs (make sure to uninstall osu! if you still have it installed before doing this or it might fail)

frenzibyte:
I advise against sharing such tutorials here as it's off topic from the PR and the game itself as a whole, and you're better off waiting for next osu! release which is coming shortly rather than tainting game files that can lead you to being restricted in the game.

PR #5754, opened 2023-04-21
Fix broken `AudioMixerVisualiser` layout

nekodex:
Noticed this in passing, seems to have been caused by the changes to `ToolWindow` in #5254.

Before:
Screen Shot 2023-04-21 at 14 15 04

After:
Screen Shot 2023-04-21 at 14 15 19

.

### DIFF
--- a/osu.Framework/Graphics/Visualisation/Audio/AudioMixerVisualiser.cs
+++ b/osu.Framework/Graphics/Visualisation/Audio/AudioMixerVisualiser.cs
@@ -26,7 +26,7 @@ namespace osu.Framework.Graphics.Visualisation.Audio
         public AudioMixerVisualiser()
             : base("AudioMixer", "(Ctrl+F9 to toggle)")
         {
-            ScrollContent.Expire();
+            MainHorizontalContent.Clear();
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Visualisation/Audio/AudioMixerVisualiser.cs
```diff
@@ -26,7 +26,7 @@ internal partial class AudioMixerVisualiser : ToolWindow
         public AudioMixerVisualiser()
             : base("AudioMixer", "(Ctrl+F9 to toggle)")
         {
-            ScrollContent.Expire();
+            MainHorizontalContent.Clear();
             MainHorizontalContent.Add(new BasicScrollContainer(Direction.Horizontal)
```
Joehuu, line 30: Can use `Child` instead of `Clear()` and `Add()`.

peppy: I think I feel safer with `.Clear()` logically. We generally only use `Child` for the initial set of the child, not when replacing potential children with one child.

peppy (approved):

PR #5758, opened 2023-04-26
Simplify GL renderer uniform buffer binding

smoogipoo:


osu.Framework/Graphics/OpenGL/Shaders/GLShader.cs
```diff
@@ -122,9 +123,6 @@ public void Bind()
             for (int i = 0; i < textureUniforms.Count; i++)
                 textureUniforms[i].Update();
 
-            foreach (var block in uniformBlocksValues)
-                block?.Bind();
```
bdach, line 126: It appears that with this removal `uniformBlocksValues` is only ever added to and never queried - so I think it should be okay to delete it entirely?

osu.Framework/Graphics/OpenGL/Shaders/GLUniformBlock.cs
```diff
@@ -13,54 +10,20 @@ namespace osu.Framework.Graphics.OpenGL.Shaders
...
             // This creates a mapping in the shader program that binds the block at location `index` to location `binding` in the GL pipeline.
             GL.UniformBlockBinding(shader, index, binding);
         }
-
-        /// <summary>
-        /// Assigns an <see cref="IUniformBuffer{TData}"/> to this uniform block.
-        /// </summary>
-        /// <param name="buffer">The buffer to assign.</param>
-        /// <exception cref="ArgumentException">If the provided buffer is not a <see cref="GLUniformBuffer{TData}"/>.</exception>
-        public void Assign(IUniformBuffer buffer)
```
bdach, line 41: I'm finding it difficult to figure out what this was even doing. It looks like a measure aimed at reducing bind/flush count? Should I be checking for a performance hit if so, to see how many more flushes we get with this removed?

smoogipoo: At one point, yes, it was intended to have some sort of caching. But the way it worked meant it wasn't actually caching at all (see the previous `Bind` method).

I think it isn't useful as such, because UBOs are per-DrawNode (with the exclusion of the global UBO).

bdach (approved):
I *think* this is correct, although I'm not really well-versed in these renderer topics. `BindBuffer` batch flush count seems to go down with this change which seems like a good sign.

peppy (approved):

PR #5759, opened 2023-04-28
Disable exclusive fullscreen detection on veldrid renderers for now

peppy:
As far as I know, this doesn't work correctly (based on discussion with @smoogipoo), but people are taking the message seriously and choosing a potentially sub-optimal renderer as a result:

2023-04-28 17 17 21@2x

@smoogipoo please confirm before merging.

### DIFF
--- a/osu.Framework/Platform/Windows/WindowsGLRenderer.cs
+++ b/osu.Framework/Platform/Windows/WindowsGLRenderer.cs
@@ -57,6 +57,10 @@ namespace osu.Framework.Platform.Windows
             if (window.WindowState != WindowState.Fullscreen || !window.IsActive.Value || fullscreenCapability.Value != Windows.FullscreenCapability.Unknown)
                 return;
 
+            // This detection method was only ever tested / implemented with the legacy GL.
+            if (host.Renderer is not GLRenderer)
+                return;
+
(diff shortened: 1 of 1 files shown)


osu.Framework/Platform/Windows/WindowsGLRenderer.cs
```diff
@@ -57,6 +57,10 @@ private void detectFullscreenCapability(IWindow window)
             if (window.WindowState != WindowState.Fullscreen || !window.IsActive.Value || fullscreenCapability.Value != Windows.FullscreenCapability.Unknown)
                 return;
 
+            // This detection method was only ever tested / implemented with the legacy GL.
+            if (host.Renderer is not GLRenderer)
+                return;
```
smoogipoo, line 62: I don't think this is the correct fix because this is already the legacy GL renderer. I believe the correct fix is osu!-side, for the check to be against the renderer (`host.Renderer is IWindowsRenderer`) rather than the window (https://github.com/ppy/osu/blob/6308bef0c8de9af37061742049637ede263bff73/osu.Game/Overlays/Settings/Sections/Graphics/LayoutSettings.cs#L259)

smoogipoo (changes requested):

PR #5760, opened 2023-04-28
Refactor source generation for abstraction

smoogipoo:
This is a painful PR to look at because it moves everything around. I suggest not looking at the diff.

The only commits worth looking at are:
- [**Refactor multiphase tests to not use statics**](https://github.com/ppy/osu-framework/commit/74bef951b2cd8b2f69a283fb8dddc321d5caed6b)
- [**Remove unnecessary interface**](https://github.com/ppy/osu-framework/commit/0fc89f56a3a3ab2f4ab93b7792c458ab1f3a68a9)

The rest are moving of files around, splitting classes out, or slightly refactoring classes to expose some things as `abstract` where implementation is necessary.


osu.Framework.SourceGeneration/Generators/IncrementalSyntaxTarget.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            return Equals((IncrementalSyntaxTarget)obj);
+        }
+
+        public override int GetHashCode()
+        {
+            return Syntax.GetHashCode();
+        }
+
+        public class Comparer : IEqualityComparer<IncrementalSyntaxTarget>
```
bdach, line 61: I think this should be called something different, like `SyntaxNameComparer` (for a fully qualified class name of `IncrementalSyntaxTarget.SyntaxNameComparer`). I'm saying this because I find it a potential footgun that `IncrementalSyntaxTarget.Equals()` uses the `Syntax` member to determine equality, while this comparer uses `SyntaxName` (which, as far as I can see, won't even be initialised until `.WithName()` is called).

bdach (commented):
Most of this looks good, as far as I can tell. I've got one nitpick.

osu.Framework.SourceGeneration/Generators/IncrementalSyntaxTarget.cs
bdach: From the thumbs-up above I assume I'm not being unreasonable here, so I'll just make the change myself to speed up things along.

bdach (approved):

PR #5761, opened 2023-04-28
Add source generator for HandleInputCache

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/5760

| Description | Before | After |
| --- | --- | --- |
| Startup | GameStart1 | GameStart2 |
| Options | Options1 | Options2 |


osu.Framework.SourceGeneration/Generators/Input/HandleInputSemanticTarget.cs
```diff
@@ -0,0 +1,159 @@
...
+            "IHasTooltip",
+            "IHasCustomTooltip",
+            "IHasContextMenu",
+            "IHasPopover"
+        );
+
+        // HandleInputCache.non_positional_input_interfaces
+        private static readonly ImmutableHashSet<string> non_positional_input_interfaces = ImmutableHashSet.Create<string>(
+            "IKeyBindingHandler"
+        );
```
bdach, line 144: Hmm, these are not fully-qualified. Meaning that potentially if someone defines their own `FunnyProject.IKeyBindingHandler` interface, for instance, the source generator will presumably pick it up and start treating it as the framework interface. Not sure if we wanna guard against this, by specifying fully qualified names here and comparing those.

(This came to my mind when reading the tests actually, since the test cases kind of exploit these interfaces not having their name fully-qualified as the framework interface specifically.)

osu.Framework/Graphics/Drawable_HandleInputCache.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            private static readonly string[] non_positional_input_properties =
+            {
+                nameof(HandleNonPositionalInput),
+                nameof(AcceptsFocus),
+            };
+
+            public static bool RequestsPositionalInput(Drawable drawable)
+            {
+                if (drawable is ISourceGeneratedInputCache sgInput && sgInput.KnownType == drawable.GetType())
```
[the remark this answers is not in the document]
smoogipoo: They'll go away with a package bump :)

osu.Framework.SourceGeneration/Generators/Input/HandleInputSemanticTarget.cs
```diff
@@ -0,0 +1,159 @@
...
+        private bool isDrawableProperty(IPropertySymbol property)
+        {
+            while (property.OverriddenProperty != null)
+                property = property.OverriddenProperty;
+
+            return isDrawableType(property.ContainingType);
+        }
+
+        private bool isDrawableType(INamedTypeSymbol type)
+            => type.Name == "Drawable" && SyntaxHelpers.GetFullyQualifiedTypeName(type) == "osu.Framework.Graphics.Drawable";
```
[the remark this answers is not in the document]
smoogipoo: That's true... I think at one point this was a precondition, but looks to be no longer necessary.

osu.Framework/Input/ISourceGeneratedInputCache.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+
+namespace osu.Framework.Input
+{
+    public interface ISourceGeneratedInputCache
+    {
+        protected internal Type KnownType { get; }
```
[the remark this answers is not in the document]
smoogipoo: It is `GetType()` with one distinction: it lets us know which leaf of the type hierarchy actually implements this member.

The idea is that if a leaf isn't specified as `partial`, then this input cache can't be used for it. Now that I'm looking over this code again, I realise this may have an issue where the type hierarchy is specified as partial -> non-partial -> partial (i.e. the leaf node is partial but there's a branch that isn't).

osu.Framework.SourceGeneration/Generators/Input/HandleInputSemanticTarget.cs
smoogipoo: This is true. It's a bit of an issue with every SG we have (including BDL), I think it probably needs to be split out into two passes - first checking for the syntax name then checking the fully qualified semantic name.

Not sure if it's worth fixing here, but I'll take a shot at it.

osu.Framework/Input/ISourceGeneratedInputCache.cs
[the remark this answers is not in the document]
smoogipoo: I've changed the generator a bit so that it inspects the entire type hierarchy. Initial testing doesn't seem to indicate that this is a performance concern, but I can't tell precisely.

osu.Framework.SourceGeneration/Generators/Input/HandleInputSemanticTarget.cs
smoogipoo: Done, now uses fully qualified type names here.

bdach:
I've re-read this - all code review suggestions look applied, source looks sane. That said, I'd probably want to spend a bit of time setting up local unit test runs for both framework and game with this sourcegen enabled, to see if there are any regressions, before I'd be 100% comfortable with leaving an approval.

If you've already done this, or don't think that's necessary and that we can play it by ear whenever, then I can approve this and we can move forward. Otherwise, I'll set a few test runs up, but probably no sooner than this weekend.

smoogipoo:
Reviewing this over again myself for the 5th time now... I believe there's still an issue if a non-partial class is the leaf node.

./sigh

I think this is going to have to end up something like dependency injection after all.

bdach:
I'll hold off on the testing for now then until the code is a final state.

smoogipoo:
Added back `KnownType` as a method to handle non-partial leaf types.

This PR is now once again ready to go... I hope.

Just to be clear, this structure should be reviewed as a basis for all other such "simple" SGens, such as https://github.com/ppy/osu-framework/pull/5763

osu.Framework/Input/ISourceGeneratedHandleInputCache.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace osu.Framework.Input
+{
+    public interface ISourceGeneratedHandleInputCache
+    {
+        /// <summary>
+        /// The most-derived type that implements this interface.
+        /// This is used to handle non-partial leaf types that can't rely on the source-generated values and must fall back to reflection.
+        /// </summary>
+        protected internal Type KnownType => typeof(object);
```
bdach, line 14: It feels kind of awful to ask this after the PR has been through the wringer so many times but... the default implementation here returning `typeof(object)` feels really off? Like, to begin with, is this default impl actually necessary for anything? Can't it be left without one? And secondly, really any type returned by this should really be a subtype of `Drawable`, so encountering `object` here feels kinda weird...

smoogipoo: Yeah that's completely fair, I've reverted it. Only did it because I noticed errors that really... shouldn't have been there - I think Rider was trolling me :/

bdach:
So today I made a local sourcegen package, hooked it up to framework and a game checkout, ran full test suites on both (all passing), and confirmed the performance gains. I'm not sure it's time effective to be doing any further manual testing of this, so as far as I'm concerned this is probably as good to go as can get, and I'm gonna leave an approval.

Not sure how we want to approach releasing this, however. What I'd probably do is:

- get next game release out _without_ this change
- force merge this to bypass code quality
- immediately after, get a new sourcegen package out and bump it in a PR / on `master` even to make the code quality inspection go away
- after that, get a framework package out with the sourcegen active, and bump framework on game-side so that we can dogfood this change during the interval between next release and release-after-next and see if anything funny falls out

But not sure my opinion holds any water here. @ppy/team-client thoughts?

smoogipoo:
What we've done in the past is pushed a nuget package from the PR, and updated the package ref in the PR.

peppy:
I'll go ahead with making the nuget package.

@bdach fwiw you didn't approve

bdach (dismissed):
whoops fixed

peppy (approved):

PR #5763, opened 2023-04-29
Add source generator for IsLongRunning property

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/5760

Removes yet another reflection usage inside `Drawable`.


osu.Framework.SourceGeneration/SyntaxHelpers.cs
```diff
@@ -109,6 +112,9 @@ public static bool IsResolvedAttribute(ITypeSymbol? type)
         public static bool IsCachedAttribute(ITypeSymbol? type)
             => type?.Name == "CachedAttribute";
 
+        public static bool IsLongRunningLoadAttribute(ITypeSymbol? type)
+            => type?.Name == "LongRunningLoadAttribute";
```
bdach, line 116: This only checks the direct type name again rather than fully qualified, but given the surrounding code I'm probably more willing to look away.

bdach (changes requested):
Rest of dependency injection tests need updating (see CI). Other than that source looks okay, though I have not done testing.

bdach:
Have fixed merge conflicts and resolved blockers. @smoogipoo please check on my changes where able and take a look whether I haven't broken anything.

If everything looks okay I'll proceed with further testing with game prior to merge.

smoogipoo:
Only thing is, I want to see if this can be changed to use the same structure as https://github.com/ppy/osu-framework/pull/5761

PR #5766, opened 2023-04-29
Remove reflection usage from `Interpolation` class

smoogipoo:
This is only intended to remove the reflection from the ctor, both to reduce memory footprint but also to be more AOT-friendly. There appears to be no difference in performance.

Before:

|                   Method |      Mean |     Error |    StdDev | Allocated |
|------------------------- |----------:|----------:|----------:|----------:|
| InterpolateMarginPadding | 45.014 ns | 0.6000 ns | 0.5612 ns |         - |
|  InterpolateFloatGeneric |  8.282 ns | 0.1558 ns | 0.1382 ns |         - |
|         InterpolateFloat |  8.319 ns | 0.1909 ns | 0.2344 ns |         - |

After:

|                   Method |      Mean |     Error |    StdDev | Allocated |
|------------------------- |----------:|----------:|----------:|----------:|
| InterpolateMarginPadding | 45.721 ns | 0.3529 ns | 0.2947 ns |         - |
|  InterpolateFloatGeneric |  7.938 ns | 0.0656 ns | 0.0614 ns |         - |
|         InterpolateFloat |  6.279 ns | 0.1024 ns | 0.0855 ns |         - |

### DIFF
--- a/osu.Framework/Utils/Interpolation.cs
+++ b/osu.Framework/Utils/Interpolation.cs
@@ -2,7 +2,6 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Linq;
 using System.Runtime.Serialization;
 using osu.Framework.Extensions.Color4Extensions;
 using osu.Framework.Graphics;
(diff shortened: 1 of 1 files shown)


bdach:
> There appears to be no difference in performance.

I wouldn't expect any benchmark to show any difference in performance given that this PR is touching a static ctor. That said I don't see much a reason *not* to make this change, if only to be a bit friendlier to the JIT (in that it can optimise the unnecessary branches away for each combination of generic type args).

Ran this past game-side tests, no failures. Feeling somewhat confident about this one so I'll just get it in.

bdach (approved):

PR #5768, opened 2023-05-05
Move `ActiveCursor` initial population to ctor to allow safe non-null specification

peppy:
See discussion at https://github.com/ppy/osu/pull/23394#issuecomment-1534999637


bdach:
Making nullable triggered even more inspections on the framework side... This is a little cursed.

I think at this point the most amicable solution would be to just move the virtual `CreateCursor()` call to ctor in order to be able to safely have the member as non-null. Yes it'll trigger an inspection, but we're really just working around BDL execution ordering at this point...

peppy:
I'm always apprehensive about moving such methods to the `ctor` as it makes it impossible to async load the content (unless it's nested another level deep). But let's go with it.

bdach (approved):

PR #5771, opened 2023-05-06
Remove cast in `ResourceStore` constructor

nobbele:


bdach:
Good spot. Looking at history I can't see how this cast ever made sense.

- [Initial framework commit](https://github.com/ppy/osu-framework/commit/f3d00712fd6eceb0c999ccc5e1aecf69e91525b3#diff-acb68a89eb89ea0a142cdec123a58144f566800e8df25c2dfc63444d686bab70) has what mostly looks like a `foreach` autocasting iterator mistake (incoming param is `IResourceStore<T>`, which then is foreach-soft-cast to `ResourceStore<T>` *for some reason*).
- Then [this commit](https://github.com/ppy/osu-framework/commit/28e1db8f1a7eb08f2d07eaca43aa3e1a871a3b5b#diff-22691aab7f4673fe0df3bc7557324274454cd37c40f046352155b78ee04c9ca3) makes it an explicit cast to silence a warning, which okay, it _will_ silence the warning, but *why though*.
- And then there's [this one](https://github.com/ppy/osu-framework/commit/25810554446f73ec3b674786270294aed92b00dc#diff-22691aab7f4673fe0df3bc7557324274454cd37c40f046352155b78ee04c9ca3) which makes it linq to make better I guess?

And this is made more perplexing by the fact that `AddStore()` does not cast, and neither does the ctor with the single `IResourceStore<T>`. Really really strange...

Let's chalk this up to 2016/2017 code I suppose.

bdach (approved):

PR #5772, opened 2023-05-06
Allow `SearchContainer` implementations to manually trigger a `Filter` operation

Cootz:
As discussed [here](https://github.com/ppy/osu/discussions/22893#discussioncomment-5790182) this PR adds `SearchContainer.Filter`
#### Usage:
```csharp
public partial class MySearchContainer : SearchContainer
{
    public string MySearchTerm
    {
        get => SearchTerm;
        set
        {
            SearchTerm = value;

            if (!IsPresent)
                Filter();
        }
    }

    //OR

    public void ForceFilter() => Filter();
}
```


bdach:
Context which should be in the OP but isn't:

This is supposed to facilitate the implementation of search box in the mods overlay. The problem is thus: To use `SearchContainer`, you would want to add a `SearchContainer` to each mod column in the overlay. However, mod columns that have all of their mods filtered out will hide themselves to not take up space.

The other part of the problem is that `SearchContainer`'s filter logic needs the search container to be present to run, because it's based on invalidation and `Update()`. The change of the filter term only invalidates the filter, but the only thing that can re-validate the filter is `Update()`.

So if you have a case wherein the search container filters out all children because they don't match the filter, and then have logic that says "if all children of the search container are hidden, then hide an ancestor of the search container", then if you filter out all children and attempt to *un*filter them again, then the column will not unhide, because the filter doesn't run, because it can't as its `Update()` will never be reached.

This is supposed to be a somewhat primitive workaround for this that will allow force-executing the filter instantly on search term change if required. I considered doing this myself, but didn't really want to PR that without exploring better options which never materialised.

Also I'm not sure how accurate the sample in the OP is, because I'm pretty sure it only works if you directly hide the `SearchContainer`, unless you do some weird overriding. If you hide the `SearchContainer`'s parent, then the `IsPresent` flag on the `SearchContainer` *will not* change and so `SearchTerm changes will have no effect.

---

@ppy/team-client interested in thoughts on this

osu.Framework/Graphics/Containers/SearchContainer.cs
```diff
@@ -92,12 +92,15 @@ protected override void Update()
...
+                Filter();
+        }
+
+        protected void Filter()
+        {
+            canBeShownBindables.Clear();
+            performFilter();
+            filterValid.Validate();
+            FilterCompleted?.Invoke();
         }
```
peppy, line 104: I'm not against this change, but I'd prefer the `protected` call

- Had xmldoc
- Only did `filterValid.Invalidate()` and left the actual filtering to the `Update` call (better debouncing, guaranteed correct thread).

peppy (changes requested):

osu.Framework/Graphics/Containers/SearchContainer.cs
peppy: And reading @bdach's response, the second of these seems to not be enough to make the use case work. So i'll need to investigate this one further before giving a green light.

Cootz: Tbh I don't know what you're concerns are, but if you don't like the idea of invoking filtering outside of the update function we can try this instead:
```csharp
{
    base.Update();

    if (!filterValid.IsValid)
        filter();
}

protected void Filter()
{
    Scheduler.AddOnce(Update); //or Scheduler.Add(Update) ??
    Scheduler.Update();
}

private void filter()
{
    canBeShownBindables.Clear();
    performFilter();
    filterValid.Validate();
    FilterCompleted?.Invoke();
}
```

peppy:
I've reworded the xmldoc here, but have two follow-up questions:

- Can we early return if the filter is already in a valid state, or is there a reason we need to bypass this validity check?

```diff
diff --git a/osu.Framework/Graphics/Containers/SearchContainer.cs b/osu.Framework/Graphics/Containers/SearchContainer.cs
index 77c389b48..e9de9e7e4 100644
--- a/osu.Framework/Graphics/Containers/SearchContainer.cs
+++ b/osu.Framework/Graphics/Containers/SearchContainer.cs
@@ -104,9 +104,14 @@ protected override void Update()
         ///
         /// However, if <see cref="SearchContainer{T}"/> or any of its parents are hidden this will not be run.
         /// If an implementation relies on filtering to become present / visible, this method can be used to force a filter.
+        ///
+        /// Note that this will only run if the current filter is not in an already valid state.
         /// </remarks>
         protected void Filter()
         {
+            if (filterValid.IsValid)
+                return;
+
             canBeShownBindables.Clear();
             performFilter();
             filterValid.Validate();

```

- Would an alternative to any of this be an `override IsPresent => base.IsPresent || !filterValid.IsValid`? Not saying this is necessarily a great solution, but would it be an avenue for further exploration?

bdach:
> Would an alternative to any of this be an `override IsPresent => base.IsPresent || !filterValid.IsValid`? Not saying this is necessarily a great solution, but would it be an avenue for further exploration?

The mod search implementation hasn't changed very much in that particular respect from how I left it as a rough branch for @Cootz to look at, and that means that overriding presence does nothing to fix it, since it is the `SearchContainer`'s ancestor that gets hidden, meaning that an entire supertree containing the `SearchContainer` becomes non-present, meaning that fiddling with `IsPresent` in any way does nothing at all.

The `IsUpdating` solution [you mentioned on discord once](https://discord.com/channels/188630481301012481/589331078574112768/1105010337989799946) might work. I do not have the bandwidth to go and try it.

Cootz:
> Can we early return if the filter is already in a valid state...

Yes we can.

> Would an alternative to any of this be an `override IsPresent => base.IsPresent || !filterValid.IsValid`?

Already answered by @bdach, just add that I've spent a lot of time working on this issue as well, and that is the only thing I came up with.

osu.Framework/Graphics/Containers/SearchContainer.cs
```diff
@@ -91,8 +91,7 @@ protected override void Update()
         {
             base.Update();
 
-            if (!filterValid.IsValid)
-                Filter();
+            Filter();
```
peppy: I wanted to leave this here for visibility (and potentially better inlining). I'll add it back.

peppy (approved):

osu.Framework/Graphics/Containers/SearchContainer.cs
Cootz: ok

PR #5773, opened 2023-05-07
Remove `[ExcludeFromDynamicCompile]` attribute

bdach:
I think we've been living with .NET Hot Reload long enough to remove this.

# Breaking changes

## `[ExcludeFromDynamicCompile]` attribute has been removed

The attribute has had no effect since [migrating to Hot Reload](https://github.com/ppy/osu-framework/pull/5041).

All user classes annotated with this attribute can remove the attribute with no change to testing behaviour.


peppy:
Honestly I'd just nuke it.

peppy (approved):

PR #5776, opened 2023-05-08
Disable relative-mode mouse by default on Linux

smoogipoo:
There are still edge cases on Linux where relative-mode isn't working correctly. One example is my NoMachine setup, which grabs the cursor but pins it to `(windowWidth, windowHeight)`.

The change to `Default` is intentional (for macOS also), otherwise the in-game options will show a non-default setting for the default value on these platforms.


bdach (approved):
Seems reasonable

peppy (approved):

Issue #5770, opened 2023-05-05
Setting `VeldridUniformBuffer.Data` may call `renderer.FlushCurrentBatch` when not suppose to


EVAST9919:
Let's use `CircularBlob` as an example (but same happens with other drawables).
When populating `parametersBuffer.Data` comparison should occur, so if the same data already exists we are not calling `renderer.FlushCurrentBatch`. However data to compare is provided by the `storages[currentStorageIndex]`. At this point there are 2 storages: [0] with default data and [1] with the data from the previous frame. But `currentStorageIndex` is 0, so we are always comparing with the default data, thus making comparison fail.

_(Here's the point when I'm not sure what I'm talking about)_
**Potential point of failure:** [0] storage is **never** being populated with data. We are always increasing `currentStorageIndex` before that.

smoogipoo:
> At this point there are 2 storages: [0] with default data and [1] with the data from the previous frame. But `currentStorageIndex` is 0, so we are always comparing with the default data, thus making comparison fail.

This is covered by the second condition at the time of upload here:

PR #5777, opened 2023-05-08
Simplify Veldrid UBO uploads

smoogipoo:
Resolves https://github.com/ppy/osu-framework/issues/5770, and hopefully makes this code a lot more readable as to why it's doing things the way it is.

As a bonus, also removes the equality comparer from `Data_set()` which could improve performance where it's called multiple times (masking).


smoogipoo:
Closing temporarily, pending discussion.

smoogipoo:
Closing temporarily, pending discussion.

PR #5782, opened 2023-05-12
Fix source transfers on `Interpolating` and `Decoupled` clocks breaking

peppy:
Fixes https://github.com/ppy/osu/issues/23486

Please read individual commits and inline comments for further documentation. The important one (and the main fix) is https://github.com/ppy/osu-framework/commit/2102638056dfcf85d21b4d85266d53b5dd018767.

### DIFF
--- a/osu.Framework.Tests/Clocks/DecoupleableClockTest.cs
+++ b/osu.Framework.Tests/Clocks/DecoupleableClockTest.cs
@@ -136,6 +136,74 @@ namespace osu.Framework.Tests.Clocks
 
         #endregion
 
+        #region Source changes
+
+        [Test]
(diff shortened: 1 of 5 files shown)


bdach:
Usually when I see `DecoupleableInterpolatingFramedClock` mentioned I want to run away as far as humanly possible but this looks... almost tame? Like the changes make *a certain amount* of sense to me, I think, in that they look to be consistent with everything else around them - which is not necessarily a symptom of them being *correct*, mind - but framework tests are passing, game tests are passing, the issue looks fixed, and brief manual testing hasn't yielded breakage, so... onwards and upwards I guess?

bdach (approved):

smoogipoo (approved):
Looks reasonable

PR #5783, opened 2023-05-12
Cache value of `VerticalSync` to avoid overhead retrieving from SDL2 each time

peppy:
As seen by @smoogipoo 

I can't repro on macOS but this should be a pretty low-risk change, and especially important since this is called at least once per frame for opengl renderers:

https://cdn.discordapp.com/attachments/589331078574112768/1106511624686288946/image.png

### DIFF
--- a/osu.Framework/Platform/SDL2/SDL2GraphicsSurface.cs
+++ b/osu.Framework/Platform/SDL2/SDL2GraphicsSurface.cs
@@ -188,13 +188,26 @@ namespace osu.Framework.Platform.SDL2
             }
         }
 
+        // cache value locally as requesting from SDL is not free.
+        // it is assumed that we are the only thing changing vsync modes.
+        private bool? verticalSync = false;
+
         bool IOpenGLGraphicsSurface.VerticalSync
         {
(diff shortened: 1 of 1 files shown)


osu.Framework/Platform/SDL2/SDL2GraphicsSurface.cs
```diff
@@ -188,13 +188,26 @@ private IntPtr getProcAddress(string symbol)
             }
         }
 
+        // cache value locally as requesting from SDL is not free.
+        // it is assumed that we are the only thing changing vsync modes.
+        private bool? verticalSync = false;
```
bdach, line 193: `false` is a slightly strange choice of initialiser? I would have gone with `null`. Otherwise the non-null path in the getter is essentially dead code.

This is only saved by the fact that the first set of `VerticalSync` precedes its first set.

peppy: i don't know how that got there

bdach (approved):
The retrieval is nowhere near as bad on my machine as shown in the OP, but this is still faster

PR #5784, opened 2023-05-12
Reduce overhead induced by checking Bass CPU usage

peppy:
It's not the end of the world, but we're definitely querying too often.
JetBrains Rider 2023-05-12 at 12 27 01

### DIFF
--- a/osu.Framework/Threading/AudioThread.cs
+++ b/osu.Framework/Threading/AudioThread.cs
@@ -47,9 +47,12 @@ namespace osu.Framework.Threading
 
         private static readonly GlobalStatistic<double> cpu_usage = GlobalStatistics.Get<double>("Audio", "Bass CPU%");
 
+        private long frameCount;
+
         private void onNewFrame()
         {
-            cpu_usage.Value = Bass.CPUUsage;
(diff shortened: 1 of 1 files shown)


osu.Framework/Threading/AudioThread.cs
```diff
@@ -47,9 +47,12 @@ internal sealed override void MakeCurrent()
 
         private static readonly GlobalStatistic<double> cpu_usage = GlobalStatistics.Get<double>("Audio", "Bass CPU%");
 
+        private long frameCount;
+
         private void onNewFrame()
         {
-            cpu_usage.Value = Bass.CPUUsage;
+            if (frameCount++ % 1000 == 0)
```
OliBomby, line 54: If the goal is to optimize performance you could replace this with a bitwise operator to function as the modulo.
Something like `frameCount++ & 1023 == 0` could work.

peppy: That's probably going a bit over-the-top 😅

bdach (approved):
It's nowhere near as bad on my box, but I can definitely agree that we don't need to query that much

Issue #5779, opened 2023-05-10
(Veldrid) `DrawNode` changes in drawables with bound uniform block are causing draw performance loss


EVAST9919:
[Here's the test scene showing the issue](https://github.com/EVAST9919/osu-framework/blob/uniform-colour/osu.Framework.Tests/Visual/UserInterface/TestSceneUniformColour.cs)

On my machine (win10) In this scene when colour is static my frame times are around 1.9ms. Howewer with colour changing it becomes 5.4, which is kinda a big hit.
**What uniform blocks have to do with anything**: the issue is no longer present when changing `CircularProgress` to a simple `Box` for example and performance loss is insignificant.
**Why Veldrid:** with LegacyGL frametimes go from 1.5 (static) to 1.7 (changing), so the issue is no longer present in this case as well.

smoogipoo:
I can't seem to repro this :(

EVAST9919:
Here are my results:
| | Veldrid | LegacyGL |
|-----|-----|-----|
| Idle | ![veldrid-idle](https://github.com/ppy/osu-framework/assets/22874522/814ceb16-3c32-4548-b1ca-270f09fb92a3) | ![gl-idle](https://github.com/ppy/osu-framework/assets/22874522/8efe7864-5de4-484b-972f-f6db291c943c) |
| Animated | ![veldrid-anim](https://github.com/ppy/osu-framework/assets/22874522/b056e4d0-6317-4353-bbaf-5cd5c2b8a7ec) | ![gl-anim](https://github.com/ppy/osu-framework/assets/22874522/8ecab08a-f19f-4f3e-9c8f-423706a75542) |

EVAST9919:
Okay, turned out it's not only about colour, it's about any `DrawNode` change. Movement, scale, etc

smoogipoo:
Are you able to profile this and show the list of expensive methods? It sounds like you're hitting a case several others are, but that I haven't been able to repro.

EVAST9919:
@smoogipoo so I was using [your pr](https://github.com/ppy/osu-framework/pull/5778) as a reference and moticed that there are too many flushes still. However removing `FlushCurrentBatch` from `VeldridRenderer.BindUniformBuffer()` literally fixes everything. No sure if this is a correct fix though, but it does magic: my fps went from 500 in idle to 1k and ~950 when animated

https://github.com/ppy/osu-framework/assets/22874522/598630b7-bbc2-4c24-84b0-ecf5fd611660

[...]

PR #5785, opened 2023-05-12
Fix D3D11 losing performance by constantly mapping/unmapping buffers

smoogipoo:
Resolves https://github.com/ppy/osu-framework/issues/5779

I've encapsulated the relevant logic in an `IStagingBuffer<T>`.
- Persistently-mapped buffers appear to work out-of-the-box on D3D11 and Vulkan, but not on OpenGL.
- Persistently-mapped buffers _may_ also work on MTL, either by grabbing the contained `IntPtr` of the device buffer and using the more optimal `renderer.Device.UpdateBuffer()`, _or_ by the way the class already works. This needs more testing, and I've chosen to not change MTL's behaviour here.
- Therefore, managed buffers are used for MTL, OpenGL, and OpenGL ES. There should be no behavioural changes on those platforms.

Tested:
- [x] D3D11
- [x] Vulkan (Linux)
- [x] Vulkan (Windows)
- [x] OpenGL (Linux)
- [x] OpenGL (Windows)
- [x] Metal

### DIFF
--- a/osu.Framework/Graphics/Veldrid/Batches/VeldridLinearBatch.cs
+++ b/osu.Framework/Graphics/Veldrid/Batches/VeldridLinearBatch.cs
@@ -5,7 +5,6 @@ using System;
 using osu.Framework.Graphics.Rendering.Vertices;
 using osu.Framework.Graphics.Veldrid.Buffers;
 using Veldrid;
-using BufferUsage = Veldrid.BufferUsage;
 
 namespace osu.Framework.Graphics.Veldrid.Batches
 {
@@ -20,6 +19,6 @@ namespace osu.Framework.Graphics.Veldrid.Batches
             this.type = type;
         }
 
-        protected override VeldridVertexBuffer<T> CreateVertexBuffer(VeldridRenderer renderer) => new VeldridLinearBuffer<T>(renderer, Size, type, BufferUsage.Dynamic);
+        protected override VeldridVertexBuffer<T> CreateVertexBuffer(VeldridRenderer renderer) => new VeldridLinearBuffer<T>(renderer, Size, type);
     }
 }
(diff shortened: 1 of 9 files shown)


smoogipoo:
Hold off on this one for a sec, it changes Metal behaviour very slightly still.

smoogipoo:
Should be good to go now. Not really happy about it but also not really wanting to test uncharted waters right now.

bdach:
I'm not gonna pretend I know what is being done here too much but it does seem to make thing go faster™.

I did remove a double disposal of `stagingBuffer` in https://github.com/ppy/osu-framework/pull/5785/commits/09d6ed5fb36d2a493141d94513e67749c8b083b9 because I would hope it wasn't doing anything important. Other than that...

bdach (approved):

peppy (approved):

smoogipoo:
> I'm not gonna pretend I know what is being done here too much but it does seem to make thing go faster™.

Right, I forgot to actually mention that.

`Map()`/`Unmap()` are expensive because they cause flushes. Specifically, data is flushed on `Unmap()`, and `Map()` appears to do its own bit of synchronisation.

The obvious solution (to me) of keeping Veldrid's staging buffers persistently mapped doesn't work because the data is overwritten for multiple vertex batches. There would need to be some mechanism where buffers are returned back to the pool after the GPU is done with it, but I can't figure out the D3D incantation for that (something similar to [MTL's completed handler](https://github.com/ppy/veldrid/blob/f1f8dc043257420240196bf896da03e52842dbad/src/Veldrid/MTL/MTLGraphicsDevice.cs#LL206C36-L206C36)).

One thing we need to watch out for is if there's any weirdness/glitching. The assumption I'm making here is that _within a single frame_, the data in one VBO is completely isolated to it and its staging buffer.

bdach:
> One thing we need to watch out for is if there's any weirdness/glitching. The assumption I'm making here is that _within a single frame_, the data in one VBO is completely isolated to it and its staging buffer.

https://github.com/ppy/osu/issues/23538 looks eerily similar to that description. Mind you it's intel, so it may just be doing usual intel things.

Issue #4895, opened 2021-11-15
File/folder picker does not handle init folder and root folder access denied properly on Android 10 and lower


honguyenminh:
The file/folder picker on Android is a bit broken.
Everything is tested with osu!~lazer~ on real unrooted devices with no SD card plugged in, internal storage only.

On open, picker will drop user into `/sdcard` as init directory, which is denied on Android 10 and lower, even with Storage permission allowed. This works fine on Android 11 though, no problem is encountered.

The current code on directory change will try to traverse up the tree of the directory until it reaches a readable directory.
On Android 10 and lower, it will deny access to `/sdcard`, so the picker goes up one directory, which is root. Root is (partially) readable, we can see **some** content in here, but we can't go anywhere listed, access denied. So, after the picker load, user will be dropped into root folder and got stuck here, while the path shown is still kept at `/sdcard` (this is another bug, but this can be fixed quite easily).

On Android 11 (12 not tested because I don't have money) I can get into `/sdcard` and everything works fine. But if I went to root or "Computer" then same thing happens, I got stuck there, and `/sdcard` is not inside root somehow, so I'm forced to close the picker then open it again to get back to `/sdcard`. In this case a back button mechanism should kind of "fix" this. Somehow I can access this folder without even giving the Storage permission. Weird.

bdach:
known issue.

https://github.com/ppy/osu/pull/15248 has a flow for requesting permissions properly, but is broken because of android losing egl context on pause. i have a path forward how to resolve that but it may take a while more until that's ready.

honguyenminh:
Does the solution address user getting into inaccessible path though, because going up the tree until a readable folder is found is not a good idea, especially on *nix and Android.

Just stay in the current path and don't change path at all is better imo.

bdach:
that assumes that the original directory is still accessible which is not necessarily true either. example:

```
dir1
└─ dir 2
   └─ dir 3
```

now assume you're in `dir 3`, then delete `dir 2`, but then try to navigate to `dir 2` after that deletion. neither the target directory, or the original directory, exist anymore. this would previously lead to crashes.

just needs a better sane fallback directory for android, is all.

Susko3:
Yeah, we need a better `Computer` directory for Android.

The `Computer` directory is generated by this code:
[...]

PR #5786, opened 2023-05-13
Change initial file picker path on Android to `/storage/emulated/0`

Rystars:
changing the file path to /storage/emulated/0

### DIFF
--- a/osu.Framework.Android/AndroidGameHost.cs
+++ b/osu.Framework.Android/AndroidGameHost.cs
@@ -61,7 +61,7 @@ namespace osu.Framework.Android
                 new MidiHandler()
             };
 
-        public override string InitialFileSelectorPath => @"/sdcard";
+        public override string InitialFileSelectorPath => @"/storage/emulated/0";
 
         public override Storage GetStorage(string path) => new AndroidStorage(path, this);
 


bdach:
This was talked about in https://github.com/ppy/osu-framework/issues/4895. I'm not sure what the final conclusion is in that thread, though.

peppy:
`emulator` or `emulated`?

I won't begin to pretend I understand android so I'll leave this up to others.

Rystars:
> `emulator` or `emulated`?
> 
> I won't begin to pretend I understand android so I'll leave this up to others.



> `emulator` or `emulated`?
> 
> I won't begin to pretend I understand android so I'll leave this up to others.

emulated

Susko3:
I don't know if this is better than the current way, but I do know that that this will not always work.
The `0` in `/storage/emulated/0` denotes the current user and if the Android device has multiple users, this number will be `1`, `2`, etc. breaking this fix.

Rystars:
> I don't know if this is better than the current way, but I do know that that this will not always work. The `0` in `/storage/emulated/0` denotes the current user and if the Android device has multiple users, this number will be `1`, `2`, etc. breaking this fix.

is it true? but I see on the internet mostly 0

Susko3:
> but I see on the internet mostly 0

Most people have only one user on their device (or have multiple users without being aware of it). See [these docs](https://source.android.com/docs/devices/admin/multi-user-testing#device-paths) for more about multiple users.

PR #5789, opened 2023-05-13
Implement handling keyboard key repeat events

timiimit:
This PR adds a separates `KeyboardKeyInput` such that it no longer inherits from `ButtonInput`. Keyboard keys don't act exactly as other buttons because system sends repeated events when user is holding them down for a long time.

With this change users will be able to use their preferred system settings for repeating key events instead of our framework having a built-in way of emulating this behaviour. This issue has been raised here: 
- ppy/osu/issues/23494

### DIFF
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -1,4 +1,4 @@
-// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
 #nullable disable
@@ -42,9 +42,6 @@ namespace osu.Framework.Input
         public Drawable FocusedDrawable { get; internal set; }
 
         protected abstract ImmutableArray<InputHandler> InputHandlers { get; }
-
-        private double keyboardRepeatTime;
-        private Key? keyboardRepeatKey;
 
         /// <summary>
         /// The initial input state. <see cref="CurrentState"/> is always equal (as a reference) to the value returned from this.
@@ -440,13 +437,6 @@ namespace osu.Framework.Input
 
             foreach (var result in pendingInputs)
             {
-                if (result is KeyboardKeyInput k)
-                {
-                    if (k.Entries[0].IsRepeated)
-                    {
-                        int a = 0;
-                    }
-                }
                 result.Apply(CurrentState, this);
             }
 
@@ -663,13 +653,22 @@ namespace osu.Framework.Input
                               || k == Key.LShift || k == Key.RShift
                               || k == Key.LWin || k == Key.RWin;
 
-        protected virtual void HandleKeyboardKeyStateChange(ButtonStateChangeEvent<Key> keyboardKeyStateChange)
+        protected virtual void HandleKeyboardKeyStateChange(KeyboardKeyStateChangeEvent keyboardKeyStateChange)
         {
             var state = keyboardKeyStateChange.State;
             var key = keyboardKeyStateChange.Button;
             var kind = keyboardKeyStateChange.Kind;
+            var repeat = keyboardKeyStateChange.IsRepeated;
 
-            GetButtonEventManagerFor(key).HandleButtonStateChange(state, kind);
+            var keyManager = GetButtonEventManagerFor(key);
+            if (repeat)
+            {
+                keyManager.HandleRepeat(state);
+            }
+            else
+            {
+                keyManager.HandleButtonStateChange(state, kind);
+            }
         }
 
         protected virtual void HandleTouchStateChange(TouchStateChangeEvent e)
@@ -845,7 +844,7 @@ namespace osu.Framework.Input
                     HandleMouseButtonStateChange(mouseButtonStateChange);
                     return;
 
-                case ButtonStateChangeEvent<Key> keyboardKeyStateChange:
+                case KeyboardKeyStateChangeEvent keyboardKeyStateChange:
                     HandleKeyboardKeyStateChange(keyboardKeyStateChange);
                     return;
 
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -450,6 +450,13 @@ namespace osu.Framework.Input
 
             foreach (var result in pendingInputs)
             {
+                if (result is KeyboardKeyInput k)
+                {
+                    if (k.Entries[0].IsRepeated)
+                    {
+                        int a = 0;
+                    }
+                }
                 result.Apply(CurrentState, this);
             }
 
--- a/osu.Framework/Input/InputManager.cs
+++ b/osu.Framework/Input/InputManager.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
 #nullable disable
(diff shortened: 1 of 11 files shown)


bdach:
Uh... `KeyDownEvent` already has a `Repeat` flag.

https://github.com/ppy/osu-framework/blob/master/osu.Framework/Input/Events/KeyDownEvent.cs#L17

What purpose would it serve to have two implementations of the same thing?

timiimit:
@bdach its not two implementations of the same thing. This PR adds support for direct system handling of repeated events. Framework didn't previously know when system sends the repeat event. It generated one itself in `osu.Framework/Input/InputManager.cs` on a constant timer. User's system could have been sending repeat events whenever, but framework just ignored it.

bdach:
That would be a much better description of the PR than what you originally had. The OP reads as if we weren't doing key repeats in any way at all.

Nevertheless:

- This change 100% causes tests to fail. `TestSceneTextBoxKeyEvents` in particular.
- Android has [special handling](https://github.com/ppy/osu-framework/blob/master/osu.Framework.Android/Input/AndroidKeyboardHandler.cs) of keyboard for special reasons and needs to be updated.

osu.Framework/Input/StateChanges/Events/KeyboardKeyStateChangeEvent.cs
```diff
@@ -0,0 +1,24 @@
...
+using osuTK.Input;
+
+namespace osu.Framework.Input.StateChanges.Events
+{
+    public class KeyboardKeyStateChangeEvent : ButtonStateChangeEvent<Key>
+    {
+        /// <summary>
+        /// Whether this event was sent in the past and is now being repeated.
+        /// </summary>
+        public readonly bool IsRepeated;
```
bdach, line 16: Would prefer to see all instances of `IsRepeated` in this pull be renamed to `Repeat` to match the input event flag.

osu.Framework/Testing/Input/ManualInputManager.cs
```diff
@@ -88,13 +88,13 @@ public void Input(IInput input)
         /// To press and release a key immediately, use <see cref="Key"/>.
         /// </remarks>
         /// <param name="key">The key to press.</param>
-        public void PressKey(Key key) => Input(new KeyboardKeyInput(key, true));
+        public void PressKey(Key key) => Input(new KeyboardKeyInput(key, true, false));
```
bdach, line 91: This means that the manual input manager is going to fail to send any repeat events ever now. I imagine this may be a change that breaks several game-side tests. Or not cause them to fail, but cause them to test anything meaningful (because a manual input manager test will never send a repeat now).

This should be checked, and if that is the case, an alternative method should be supplied that can provide some sort of repeat mechanism for test cases so that test logic covering repeats may be exercised in the tests. Even if by moving the manual reimplementation of repeat removed from `InputManager` here.

osu.Framework/Input/StateChanges/Events/KeyboardKeyStateChangeEvent.cs
Gyoshi, line 16: Does it make sense for a a state change to be repeated? afai understand a state change would be initially pressing down or ultimately releasing the button.

bdach (commented):
Additionally:

osu.Framework/Input/StateChanges/Events/KeyboardKeyStateChangeEvent.cs
bdach: That is a bit of a moot point because you can't achieve what this pull wants to achieve without adding this flag where it is being added. Yes it maybe strictly doesn't match the structure around it (which is mostly based on state difference) but that data must flow somewhere and this is probably the only viable way - at least not without further refactoring which should *not* be done in this pull.

osu.Framework/Input/StateChanges/KeyboardKeyInput.cs
Susko3, line 1: I don't really like how this just copies code over from `ButtonInput<TButton>`. Would it be possible to not do that?

I'm starting to wonder if it would make sense to have repeat happen for every single button type (eg. joystick, midi). If joystick D-Pad down is pressed, I would expect the down action to be continuously executed (eg. moving down in osu! song select).

This special handling wouldn't be needed if all buttons had repeat.

Susko3 (changes requested):
Using system-provided repeat is a good thing, but I don't thing this is the right way to approach doing that.

The change brings in too much new logic for things that seem already implemented.

And I'm not sure how it behaves wrt using nested InputMangers:
If I'm pressing down a key, and a new input manager is focused/created, I would expect that the first key event it generates be `Repeat=false`.

As it currently stands, repeat doesn't work with `ManualInputManager`/`PassThroughInputManager` used in tests (see `TestSceneTextBox` and try to move the cursor).

Susko3:
Android is fairly easy to implement, just need to handle [`KeyEvent.getRepeatCount()`](https://developer.android.com/reference/android/view/KeyEvent#getRepeatCount()).

osu.Framework/Input/InputManager.cs
```diff
@@ -681,29 +653,21 @@ private void updateHoverEvents(InputState state)
...
             var kind = keyboardKeyStateChange.Kind;
+            var repeat = keyboardKeyStateChange.IsRepeated;
 
-            GetButtonEventManagerFor(key).HandleButtonStateChange(state, kind);
-
-            if (kind == ButtonStateChangeKind.Pressed)
+            var keyManager = GetButtonEventManagerFor(key);
+            if (repeat)
             {
-                if (!isModifierKey(key))
```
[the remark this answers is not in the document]
timiimit: I removed this by mistake, my bad

osu.Framework/Input/StateChanges/KeyboardKeyInput.cs
timiimit: I'll try to think of something to combine the code.

I tried that approach originally, but i thought it would mess up too many things. But i can do that if it is desired.

osu.Framework/Input/StateChanges/Events/KeyboardKeyStateChangeEvent.cs
timiimit: I was thinking about that too, I wasnt sure if it would affect any other part of the code. But everything i tested seemed to be working as expected. I do agree that naming might be a bit off but key events can be repeated on most (all?) platforms so there should be a way to handle it in the framework.

osu.Framework/Testing/Input/ManualInputManager.cs
timiimit: I reimplemented the logic that was removed from `InputManager` in `ManualInputManager`. all textbox tests in the framework work now.

osu.Framework/Input/StateChanges/KeyboardKeyInput.cs
bdach: I wouldn't do this. At least not in this pull. This one at least has a fairly small scope and shouldn't visibly change behaviour (at least in theory, I haven't gone and tested if that is the case).

Adding repeat to other input devices is a soft-breaking API change and needs to be handled appropriately. It's not in scope of this PR.

timiimit: > I don't really like how this just copies code over from ButtonInput<TButton>. Would it be possible to not do that?

I don't think there is a way to remove duplication without sacrificing naming. `ButtonInputEntry` calls button `Button` but  `KeyboardKeyInputEntry` calls what is actually a button `Key`. If it is okay to always use `Button` then there probably is a way to introduce inheritance and combine the two.

timiimit: I can't figure it out. I tried to make it work, but constraints really aren't my strong point.

osu.Framework/Input/InputManager.cs
```diff
@@ -653,13 +653,18 @@ private void updateHoverEvents(InputState state)
...
+        private bool isFirstKeyDownEvent = true;
+
         protected virtual void HandleKeyboardKeyStateChange(KeyboardKeyStateChangeEvent keyboardKeyStateChange)
         {
             var state = keyboardKeyStateChange.State;
             var key = keyboardKeyStateChange.Button;
             var kind = keyboardKeyStateChange.Kind;
             var repeat = keyboardKeyStateChange.Repeat;
 
+            if (isFirstKeyDownEvent && repeat)
```
timiimit, line 665: > If I'm pressing down a key, and a new input manager is focused/created, I would expect that the first key event it generates be Repeat=false.

I did handle this case in last commit, but after thinking some more about this it doesnt really make sense to me anymore.

Your expectation is the same as saying that the first keyboard event should be a keydown. That simply isn't true, it might be a keyup. User could be holding down a key from before any program has started and then released it as the first thing when it starts.

I think framework should be transparent with what events it is receiving.

PR #5790, opened 2023-05-14
Fix keyboard shortcuts not working as expected on non-QWERTY keyboard

### LINKED ISSUE #5767: InputKey does not use the correct keys on keyboards other than QWERTY

Susko3:
- Closes https://github.com/ppy/osu/issues/11143
- Closes #5767
- Alternative to / closes #5774

This PR does two things combined into one PR (to help see the full picture, as it's hopefully not too big):
1. Add support for propagating `SDL_Keycode`s trough the input hierarchy (implementation idea taken from MidiEvent/State `Velocity`)
   - added with the new `struct KeyboardKey`, and `char Character` properties on events / where needed
   - also supported on Android (but Android is werid and already kinda broken on `master`)
   - this is done in a semi-extensive way, to allow adding in shifted/altgr characters if needed
2. Add support for using characters from [1.] in `KeyCombination`/`InputKey`
   - this is done by adding new `InputKey`s: from `KeycodeA` trough `KeycodeZ`

While fixing the above issues, this PR also adds new things, mainly in [1.]:

The new `KeyboardEvent.Character` property would be useful for [access keys](https://learn.microsoft.com/en-us/windows/apps/design/input/access-keys) ([old docs, ref for underlining](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/bb246429(v=vs.85))) that are seen in newer [osu! figma designs](https://www.figma.com/file/DXKwqZhD5yyb1igc3mKo1P/Song-Select-Draft-2?type=design&node-id=0-1) (`...` underline on Mods, Random, Options, etc).
For access keys, I'd like to have some level of localisation support (and ways to check that the localisations don't have conflicts).

# Main considerations when making this PR

## Don't break existing behaviour

Consumers using `KeyboardEvent` will not notice any changes, as the new `Character` property is added in a way that only augments the existing information and no behaviour is changed.

Existing `KeyBinding`s are not affected, even in cases where the new keycode feature doesn't work.

## Simple to use for consumers

`KeyboardEvent`s have minimal API where it makes sense for the newly added `char Character` property.

Changing existing `KeyBinding`s to keycode ones is as simple as:

```diff
-new KeyBinding(new KeyCombination(InputKey.Control, InputKey.X), PlatformAction.Cut)
+new KeyBinding(new KeyCombination(InputKey.Control, InputKey.KeycodeX), PlatformAction.Cut)
```

## Only cover the most important use cases and don't have API footguns

Support has been added where it makes sense and where it would be used.
That's why the new keycodes don't work with `KeyCombinationMatchingMode.Exact`. While it would be possible to support this, there would need to be a way to correlate an `InputKey.Keycode{}` with an `InputKey.{}` (like the `Character` and the `Key` in a `KeyboardKey`). This is deemed unnecessary to implement as it's complex and `KeyCombinationMatchingMode.Exact` isn't used.

Even though `KeyboardKey.Character` has support for all Unicode BMP characters, `InputKey` only has support for `Keycode[A-Z]`. This is because only those keys can reliably be used on all (or most, I cannot check them all) keyboard layouts. Other characters, like punctuation, may not be available. Number keycode are not available as they would be functionally identical to `InputKey.Number#` + `InputKey.Keypad#`.

It would be nice to have support for plus and minus keys, but SDL doesn't work like that. Windows does have [virtual keys](https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes) for those (`VK_OEM_PLUS`, `VK_OEM_MINUS`), but SDL does not map them in this way (`VK_OEM_PLUS` would be `SDL_SCANCODE_EQUALS`, `SDLK_EQUALS` on en_us keyboard).

# Stuff to test and check

- [ ] Check that the keys on macOS / iOS map to the expected actions when using different keyboard layouts - that it matches other native programs

# Breaking changes

## `KeyCombination.FromKey()`

Signature changed from `FromKey(Key)` to `FromKey(KeyboardKey)`.
Return type changed from a single `InputKey` to `IEnumerable<InputKey>`. Consult the xmldoc for more information.

To keep existing behaviour, change:

```diff
 KeyDownEvent e;

-InputKey iKey = KeyCombination.FromKey(e.Key);
+InputKey iKey = KeyCombination.FromKey(e.KeyboardKey).First();
```


MatthieuHernandez (changes requested):
I tested your branch and found a bug.
When I press A or any other key it is counted as 2 times. (only happens with letters)
Example below:
image

Susko3:
> When I press A or any other key it is counted as 2 times. (only happens with letters)

That's to be expected, as both `InputKey.A` and `InputKey.KeycodeA` are pressed (they have the same display name as they are the same to the user).

I've updated the tests to show them separately now, hopefully this clears the confusion.

peppy:
Hoping we can get this in sometime soon... probably should be prioritised for review when we get around to reviewing more framework / QoL stuff.

osu.Framework/Input/KeyboardKey.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Should be matched against when the location of a key on the keyboard is more important than the character printed on it.
+        /// </remarks>
+        /// <seealso cref="Character"/>
+        public Key Key { get; }
+
+        /// <summary>
+        /// The character that this key would generate if pressed (roughly equivalent to the keycode - the character printed on the key).
+        /// Dependant on the system keyboard layout. Will be <c>default</c> if no character is available.
```
bdach, line 32: Why does this use `char` and fallback to `default` if there's no matching key, rather than, you know, use the explicit option, namely `char?`?

osu.Framework/Input/Bindings/InputKey.cs
```diff
@@ -2518,5 +2518,165 @@ public enum InputKey
...
 
         #endregion
+
+        #region Keyboard keycodes
+
+        /// <summary>
+        /// Key that produces the 'a' character when pressed. Depends on the current keyboard layout.
+        /// </summary>
+        /// <remarks>Different from <see cref="A"/>.</remarks>
+        KeycodeA = 10752 + 1,
```
bdach, line 2528: I feel like this is missing two details in this xmldoc:
- first, that there might *be* no physical key that produces this `InputKey` on the current keyboard layout,
- what `InputKey.A` _really_ means now (because this says that they're "different" now, but doesn't say how) - the best approximation of which would be "the key on the user's keyboard that is closest to the position of the A key on a QWERTY US keyboard", I guess?

osu.Framework/Input/Events/KeyboardEvent.cs
```diff
@@ -13,12 +13,19 @@ namespace osu.Framework.Input.Events
...
     {
+        /// <inheritdoc cref="Input.KeyboardKey.Key"/>
         public readonly Key Key;
 
-        /// <summary>
-        /// Whether a specific key is pressed.
-        /// </summary>
-        public bool IsPressed(Key key) => CurrentState.Keyboard.Keys.IsPressed(key);
+        /// <inheritdoc cref="Input.KeyboardKey.Character"/>
+        public readonly char Character;
```
bdach, line 20: Similarly here, I feel this should be nullable. Like what "character" does a `Key.LeftArrow` press produce?

osu.Framework/Input/States/KeyboardState.cs
```diff
@@ -1,6 +1,10 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+#nullable disable
```
bdach, line 4: Huh?

osu.Framework/Input/States/KeyboardState.cs
```diff
@@ -9,6 +13,26 @@ public class KeyboardState
     {
         public readonly ButtonStates<Key> Keys = new ButtonStates<Key>();
 
+        /// <summary>
+        /// Mapping of <see cref="Key"/>s to <see cref="KeyboardKey.Character"/>s printed on them.
+        /// </summary>
+        /// <seealso cref="KeyboardKey"/>
+        internal readonly Dictionary<Key, char> Characters = new Dictionary<Key, char>();
```
bdach, line 20: What is responsible for cleaning up potentially-stale key-to-character mappings after a key is released? Does anything even do that?

Or, to reword the question: what's the guarantee that this mapping is fully correct at any given moment?

osu.Framework/Platform/SDL2/SDL2Extensions.cs
```diff
@@ -452,6 +456,99 @@ public static Key ToKey(this SDL.SDL_Keysym sdlKeysym)
...
 
+        public static char GetCharacter(this SDL.SDL_Keycode keycode, bool numLockOn)
+        {
+            switch (keycode)
+            {
+                case SDL.SDL_Keycode.SDLK_UNKNOWN:
+                    return default;
+
+                case SDL.SDL_Keycode.SDLK_RETURN: // SDLK_RETURN is '\r'
+                    return '\n';
```
bdach, line 467: I'm not even sure how to react to this. So is it `\r` or `\n`? Does anything even _produce_ `\n` if not `SDLK_RETURN`? Does this mean there's a possibility that two `\n`s can be emitted? If yes then in what circumstances?

osu.Framework/Platform/SDL2/SDL2Extensions.cs
```diff
@@ -452,6 +456,99 @@ public static Key ToKey(this SDL.SDL_Keysym sdlKeysym)
...
+
+                case SDL.SDL_Keycode.SDLK_KP_9:
+                    return numLockOn ? '9' : default;
+
+                case SDL.SDL_Keycode.SDLK_KP_0:
+                    return numLockOn ? '0' : default;
+
+                case SDL.SDL_Keycode.SDLK_KP_PERIOD:
+                    // potentially return ',' as decimal separator based on keyboard layout.
+                    return numLockOn ? '.' : '\x7f'; // char for delete
```
bdach, line 526: Why would this method produce a control character _at all anywhere_? By that logic, where's `U+0008` (backspace)? `U+001B` (escape)?

osu.Framework/Input/Bindings/KeyCombination.cs
```diff
@@ -127,8 +127,8 @@ internal static bool ContainsAll(ImmutableArray<InputKey> candidateKey, Immutabl
                 case KeyCombinationMatchingMode.Exact:
                     foreach (var key in pressedKey)
                     {
-                        // in exact matching mode, every pressed key needs to be in the candidate.
-                        if (!ContainsKeyPermissive(candidateKey, key))
+                        // in exact matching mode, every pressed key except the "virtual" keycode keys needs to be in the candidate.
+                        if (!IsKeycode(key) && !ContainsKeyPermissive(candidateKey, key))
```
bdach, line 131: An explanation of the implications of this PR on the exact matching mode is required somewhere because there is none in the OP and I'm not sure what they are myself at this point.

osu.Framework/Extensions/ExtensionMethods.cs
```diff
@@ -335,6 +335,40 @@ public static string TrimDirectorySeparator(this string path)
         [Obsolete("Use char.IsAsciiDigit.")] // can be removed 20240901
         public static bool IsAsciiDigit(this char character) => char.IsAsciiDigit(character);
 
+        /// <summary>
+        /// Gets the string representation of a character.
+        /// </summary>
+        /// <param name="character"></param>
+        /// <returns><c>string</c> with form of <c>'\0'</c>, <c>'a'</c>, <c>'\x7f'</c>, etc.</returns>
+        public static string StringRepresentation(this char character)
```
bdach, line 343: What even is this method? Why does it exist?

osu.Framework/Input/States/KeyboardState.cs
Susko3: Probably leftover from a `master` merge. Will remove.

osu.Framework/Platform/SDL2/SDL2Extensions.cs
Susko3: `SDLK_RETURN` is defined in SDL as `\r`. It makes more sense to have it map to `\n` as that is the ubiquitous newline character (eg. `\n` makes line breaks in `TextFlowContainer`).

bdach: > It makes more sense to have it map to `\n`

I'd say whether it "makes more sense" depends on the answers to my _other_ questions, namely

> Does anything even _produce_ `\n` if not `SDLK_RETURN`? Does this mean there's a possibility that two `\n`s can be emitted? If yes then in what circumstances?

which I don't see the answer to here.

osu.Framework/Platform/SDL2/SDL2Extensions.cs
Susko3: > Why would this method produce a control character _at all anywhere_?

SDL already produces control characters, so I made this to match. I don't see the problem with providing control characters.

> By that logic, where's `U+0008` (backspace)? `U+001B` (escape)?

This is already handled by SDL. You can check by pressing those keys in `TestSceneInputManager`.

osu.Framework/Extensions/ExtensionMethods.cs
Susko3: Mostly for debugging, to have control characters be readable in text (eg. input event logging, `TestSceneInputManager`). The o!f font rendered would just displays them as `?` otherwise.

osu.Framework/Input/States/KeyboardState.cs
Susko3: This mapping may contain stale values and those are not cleaned up, but it's guaranteed to have valid values for currently pressed keys. This mapping is deliberately made `internal` as it's an implementation detail.

The public methods/properties of `KeyboardState` guarantee correctness as they access the mapping via `PressedKeys`.

Usage of the mapping outside `KeyboardState` requires more scrutiny. I'll make some tests for that, especially for `PassThroughInputManager`.

osu.Framework/Input/Bindings/InputKey.cs
Susko3: > first, that there might _be_ no physical key that produces this `InputKey` on the current keyboard layout

I thought that all keyboard layouts would have ascii alpha keys, and any non-Latin languages (eg. Japanese) would have a standard US keyboard and have an IME do all the work. Turns out some keyboard layouts don't have all the ascii, _but_ windows lies to SDL about what the characters are, so it works out fine. (Typing in a textbox is not affected by this lie.)

Take `Azerbaijani (Standard)`, which has some A-Z characters. Windows lies to SDL (in hopes of compatibility I guess) and overlays the A-Z of the US layout over the actual keyboard. This works in our favour and ensures all `InputKey.Keycode*` keys are accessible.

(Notice how the `Z` keycode is pressed twice, since the US A-Z doesn't overlap the original Z.)

image

> what `InputKey.A` _really_ means now (because this says that they're "different" now, but doesn't say how) - the best approximation of which would be "the key on the user's keyboard that is closest to the position of the A key on a QWERTY US keyboard", I guess?

Maybe better to have "Not to be confused with A"?

I'll expand the xmldoc of `InputKey.[A-Z]` to mention something like
```xml
/// <summary>
/// The physical key corresponding to the A key on a US QWERTY keyboard.
/// </summary>
/// <remarks>Not to be confused with <see cref="KeycodeA"/></remarks>
A = 83,
```

osu.Framework/Input/Bindings/KeyCombination.cs
Susko3: Thinking more about the relationship between physical keys and "virtual keys" (`char?`/Keycode) in our case. It seems that the relationship between a physical key and its keycode is similar to eg. `InputKey.{L,R}Control` and `InputKey.Control`. Same as joystick buttons - there's `Joystick1`, `Joystick2` and also the virtual `ABXY` face buttons (joystick buttons are out of scope).

If the key combination is `InputKey.Control, InputKey.Number9` and the user presses `InputKey.LControl, InputKey.RControl, InputKey.Number9` - that is an exact match.

Maybe it's wrong to have the virtual keys present in the "keys pressed by user" array, and it's the job of `ContainsKey` and `ContainsKeyPermissive` to unmap any `InputKey.Keycode*` found in the candidate combination.

I'll have to think more about this.

osu.Framework/Platform/SDL2/SDL2Extensions.cs
Susko3: It's possible for two physical keys to produce the same keycode. Eg. regular enter and numpad enter.

I've added a `.Distinct()` to cover that in `KeyboardState`, but it's not a silver bullet.

https://github.com/ppy/osu-framework/blob/54b49e1c77490243766e5e5e6ed56d413252c9c3/osu.Framework/Input/States/KeyboardState.cs#L32

Two keys having the same keycode means that `KeyBindingContainer.pressedInputKeys` can get desynchronised from `KeyboardState.PressedCharacters`, but that whole logic is [pending changes](https://github.com/ppy/osu-framework/pull/5790#discussion_r1540807220). I'll add a failing test for that.

osu.Framework/Input/Bindings/InputKey.cs
bdach: > I'll expand the xmldoc of InputKey.[A-Z] to mention something like

Doesn't look to have been applied yet.

osu.Framework/Extensions/InputKeyExtensions.cs
```diff
@@ -38,5 +43,63 @@ public static bool IsVirtual(this InputKey key)
                     return false;
             }
         }
+
+        /// <summary>
+        /// Gets the <see cref="Key"/> equavalent to this <see cref="InputKey"/>.
+        /// </summary>
+        /// <remarks>Inverse of <see cref="KeyCombination.FromKey"/>.</remarks>
+        public static bool IsKeyboardKey(this InputKey inputKey, out Key key)
```
bdach, line 51: This method is all sorts of confusing now.

- called `IsKeyboardKey` (water wet anyone?)
- has a weird `out` parameter that nobody really knows what it's doing

Does this method even _do_ anything at this point? Its only caller is `KeyCombination.GetVirtualKey()`, which - as far as I can tell -

- ensures that the incoming key is physical, so the first conditional is dead code
- handles modifiers before the switch in this method, so the entire switch is dead code
- depending on the range check this either returns the key but pushed through cast, or `Key.Unknown`

Can this just be inlined at its only usage?

osu.Framework/Input/Bindings/KeyCombination.cs
```diff
@@ -135,9 +141,11 @@ public bool IsPressed(KeyCombination pressedKeys, InputState inputState, KeyComb
         /// <param name="candidateKeyBinding">The candidate key binding to match against.</param>
         /// <param name="pressedPhysicalKeys">The keys which have been pressed by a user.</param>
         /// <param name="matchingMode">The matching mode to be used when checking.</param>
+        /// <param name="keyboardCharacters"><see cref="States.KeyboardState.Characters"/> of the current <see cref="InputState"/>.</param>
```
bdach, line 144: Param needs to be named better. Something like `keyToChararacterMapping` or whatever.

osu.Framework/Input/States/KeyboardState.cs
```diff
@@ -9,6 +11,26 @@ public class KeyboardState
     {
         public readonly ButtonStates<Key> Keys = new ButtonStates<Key>();
 
+        /// <summary>
+        /// Mapping of <see cref="Key"/>s to <see cref="KeyboardKey.Character"/>s printed on them.
```
bdach, line 15:
```suggestion
        /// Mapping of <see cref="Key"/>s to <see cref="KeyboardKey.Character"/>s which are produced when the key is pressed.
```

osu.Framework/Input/States/KeyboardState.cs
```diff
@@ -9,6 +11,26 @@ public class KeyboardState
...
+        /// <seealso cref="KeyboardKey"/>
+        internal readonly Dictionary<Key, char?> Characters = new Dictionary<Key, char?>();
+
+        /// <summary>
+        /// Whether a specific <see cref="Key"/> is pressed.
+        /// </summary>
+        public bool IsPressed(Key key) => Keys.IsPressed(key);
+
+        /// <summary>
+        /// Whether a specific <see cref="KeyboardKey.Character"/> is pressed.
```
bdach, line 26:
```suggestion
        /// Whether a key that produces the specific <paramref name="character"/> is pressed.
```

osu.Framework/Input/States/KeyboardState.cs
```diff
@@ -9,6 +11,26 @@ public class KeyboardState
...
         public readonly ButtonStates<Key> Keys = new ButtonStates<Key>();
 
+        /// <summary>
+        /// Mapping of <see cref="Key"/>s to <see cref="KeyboardKey.Character"/>s printed on them.
+        /// </summary>
+        /// <seealso cref="KeyboardKey"/>
+        internal readonly Dictionary<Key, char?> Characters = new Dictionary<Key, char?>();
+
+        /// <summary>
+        /// Whether a specific <see cref="Key"/> is pressed.
```
bdach, line 21:
```suggestion
        /// Whether a specific <paramref name="key"/> is pressed.
```

osu.Framework/Platform/SDL2/SDL2Extensions.cs
bdach: Can we just not handle control characters at all?

osu.Framework/Input/Bindings/KeyCombination.cs
bdach: What's the status on this?

bdach:
Re-reading this diff yet again I am yet again struck by the amount of complexity this introduces to everything.

I am wondering whether we can change direction here and sacrifice backwards compatibility for a simpler diff. I'd begin by not bothering to add all of that `Keycode{A-Z}` stuff and just changing semantics of `Key.{A-Z}` to mean keycode / letter produced rather than attempt to handle both keycode-like and scancode-like APIS and failing at both.

@Susko3 did you explore the possibility of doing something like the above? @smoogipoo would you be fine with such a hard break?

bdach:
> The implementation is complex, but I really tried to make the API understandable and simple to use

My primary issue at this stage wrt API is the `InputKey` enum. If you ask me it needs a sharp naming break. `InputKey.[A-Z]` should become `InputKey.Scancode[A-Z]` to _really_ convey the differentiation. Or [this](https://github.com/ppy/osu-framework/pull/5790/files#r1613126722). But I don't know if everyone else is willing to stomach that.

There's also the part where attempting to handle both keycode and scancode is an inherent increase in complexity. You might say it's unavoidable but it makes it a tough sell in general.

I dunno, I really need more of @ppy/team-client to sound off here. I just want to avoid another pooling situation where we go in a direction that seems logical one step at a time and then wake up a year later having no earthly idea how anything in pooling works anymore.

bdach:
the French guys were on my ass about this issue the entire COE so i'd like to see movement on this. yes there are merge conflicts but i wanna rack people's brains about the general direction first. bump @ppy/team-client

hwsmm:
Breaking default Z/X bindings doesn't sound too disastrous to me, but maybe we can use SDL_GetKeyFromScancode to convert Z/X scancodes to keycodes, and set them as defaults if an user doesn't have custom bindings.

I was just investigating https://github.com/ppy/osu/discussions/31889 and using keycode instead of scancode seems to be a fix there.

PR #5793, opened 2023-05-15
Change method group inspection to warning to match osu!

peppy:
Fixes warnings when using a local framework checkout.


bdach (approved):

smoogipoo:
This was done for a reason though, because method groups always allocate the delegate.

Just understand that, in particular the change to `KeyBindingContainer` will now cause more allocations.

bdach:
My bad, didn't look properly. I guess we can revert the non-test ones with local suppressions...?

smoogipoo:
It's likely to be minimal but I'd suggest reverting the KBC change.

The change happened here https://github.com/ppy/osu-framework/commit/143bc5b85008ab9f09ed65fc2921878195db43e1

With discussion in https://github.com/ppy/osu-framework/pull/4144#discussion_r556088692

But in general it's good to be aware of this quirk which is why the inspection wasn't applied framework-side.

peppy:
@smoogipoo are you okay with a local rule exception for the case mentioned? I'd still like to apply the inspection in general if we can.

PR #5794, opened 2023-05-15
Revert `Schedule` changes to `Dropdown`

peppy:
This reverts changes from https://github.com/ppy/osu-framework/pull/5516, but also removed one extra `Scheduler.AddOnce` ensuring that `Items_Get` returns the same result as `Items_Set` in all cases.

Resolves performance concerns cited at https://github.com/ppy/osu-framework/pull/5516/files#r1042047938
Closes https://github.com/ppy/osu/issues/22573
Closes https://github.com/ppy/osu/issues/23537

I've tried to come up with a better approach but have not been able to yet. Let's restore sanity first and foremost.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/Dropdown.cs
+++ b/osu.Framework/Graphics/UserInterface/Dropdown.cs
@@ -52,7 +52,7 @@ namespace osu.Framework.Graphics.UserInterface
                 if (boundItemSource != null)
                     throw new InvalidOperationException($"Cannot manually set {nameof(Items)} when an {nameof(ItemSource)} is bound.");
 
-                Scheduler.AddOnce(setItems, value);
+                setItems(value);
             }
         }
 
@@ -225,7 +225,7 @@ namespace osu.Framework.Graphics.UserInterface
                     Menu.State = MenuState.Closed;
             };
 
-            ItemSource.CollectionChanged += (_, _) => Scheduler.AddOnce(setItems, ItemSource);
+            ItemSource.CollectionChanged += (_, _) => setItems(itemSource);
         }
 
         private void preselectionConfirmed(int selectedIndex)
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
@@ -6,14 +6,10 @@
 using System;
 using System.Linq;
 using NUnit.Framework;
-using osu.Framework.Allocation;
 using osu.Framework.Bindables;
-using osu.Framework.Extensions.IEnumerableExtensions;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.UserInterface;
 using osu.Framework.Input;
-using osu.Framework.Localisation;
-using osu.Framework.Platform;
 using osu.Framework.Testing;
 using osuTK;
 using osuTK.Input;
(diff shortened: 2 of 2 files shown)


osu.Framework/Graphics/UserInterface/Dropdown.cs
```diff
@@ -225,7 +225,7 @@ protected Dropdown()
                     Menu.State = MenuState.Closed;
             };
 
-            ItemSource.CollectionChanged += (_, _) => Scheduler.AddOnce(setItems, ItemSource);
+            ItemSource.CollectionChanged += (_, _) => setItems(itemSource);
```
bdach, line 228: The annoying part about this particular schedule is that it was [fixing a game-side issue](https://github.com/ppy/osu-framework/pull/5475) that caused the fullscreen resolution dropdown to not ever work. The reasoning was that because `BindableList` has no way to do an atomic replace, `ItemSource` changes should be delayed a frame so that the consumers can do a `Clear()` and `AddRange()` from under the dropdown, and it won't notice the clear because it'll only update the items on the next frame. This is on the face of it reasonable in light of this particular issue, but also completely bonkers in 99% of other perfectly normal use cases.

But it does mean that if we revert this (and the corresponding test case) we'll have to deal with the broken resolution dropdown in some other way. I would honestly at this point opt for a game-side workaround which could look something like the following:

```diff
diff --git a/osu.Game/Overlays/Settings/Sections/Graphics/LayoutSettings.cs b/osu.Game/Overlays/Settings/Sections/Graphics/LayoutSettings.cs
index 2e26d15105..92b9ea3c93 100644
--- a/osu.Game/Overlays/Settings/Sections/Graphics/LayoutSettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Graphics/LayoutSettings.cs
@@ -193,6 +193,12 @@ protected override void LoadComplete()
 
             currentDisplay.BindValueChanged(display => Schedule(() =>
             {
+                // there is no easy way to perform an atomic swap on the `resolutions` list, which is bound to the dropdown via `ItemSource`.
+                // this means, if the old resolution isn't saved, the `RemoveRange()` call below will cause `resolutionDropdown.Current` to reset value,
+                // therefore making it impossible to select any dropdown value other than "Default".
+                // to circumvent this locally, store the old value here, so that we can attempt to restore it later.
+                var oldResolution = resolutionDropdown.Current.Value;
+
                 resolutions.RemoveRange(1, resolutions.Count - 1);
 
                 if (display.NewValue != null)
@@ -204,6 +210,9 @@ protected override void LoadComplete()
                                                 .Distinct());
                 }
 
+                if (resolutions.Contains(oldResolution))
+                    resolutionDropdown.Current.Value = oldResolution;
+
                 updateDisplaySettingsVisibility();
             }), true);
 ```

The display and resolution dropdowns game-side look to be pretty much unusable on my machine (completely broken on linux, behave weirdly on windows - if I change resolution on my ultrawide display, the game also gets moved to the other display every time, and I can only change resolution if I change it on my other display and only then change displays). But with the patch above, the behaviour seems no less broken than current `master`, so it seems it is a passable workaround.

If you're okay with applying that game-side, I think we can move forward with this pretty safely.

peppy: Well spotted potential regression.

I think that's an amicable workaround, yeah.

bdach: I may end up PRing a slightly different game-side workaround because the above has a bit of an issue wherein it will flicker two times when switching resolutions (because it'll go old resolution -> default -> old resolution).

But I'm merging this because that display code is half-broken anyway and `Dropdown` shouldn't have to accomodate for it.

bdach (approved):

PR #5805, opened 2023-05-21
Update SDL native libraries

peppy:
Also bumps minor package versions because we can.

Incorporates https://github.com/libsdl-org/SDL/issues/6835, of note (but also thousands of other changes soo testing is probably best).

testing:

- [x] macOS
- [x] windows
- [x] linux


bdach (approved):
Seems fine on windows and linux. Testing was pretty light though, not sure whether anything specific is expected of me to test.

peppy:
Let's go with it.

PR #5807, opened 2023-05-24
Avoid creating new `List<RectangleI>` when no texture uploads are required

peppy:
This was triggering on every texture *bind* operation. Oops.

Loading song select (from an already warmed up glyph cache):

| before | after |
| -- | -- |
|JetBrains Rider 2023-05-24 at 04 01 46 |JetBrains Rider 2023-05-24 at 04 03 47|


smoogipoo:
This should also be applied to `GLTexture`:

https://github.com/ppy/osu-framework/blob/9f689c541b7c58459a63caf6487e36c82b34fd2b/osu.Framework/Graphics/OpenGL/Textures/GLTexture.cs#L197-L216

peppy:
Good catch, applied.

smoogipoo (approved):

PR #5810, opened 2023-05-25
Use `DummyRenderer` in `HeadlessGameHost`

smoogipoo:
A significant amount of time is spent in tests on the `CompileGlslToSpirv()` method. It's quite hard to resolve because shaders are in stores, so it's not as simple as making things `static` unless by name-pairs which brings with it potential other side effects.

`HeadlessGameHost` does not draw anyway, so it doesn't sound like a stretch to use the `DummyRenderer`:

https://github.com/ppy/osu-framework/blob/5d72e48d64c98b483985f5b6a6888ad4c1dfbc33/osu.Framework/Platform/HeadlessGameHost.cs#L89-L92

(o!f UI tests) Before:

image

(o!f UI tests) After:

image

```
osu.Game.Tests `UserInterface` tests:

Before: 67s
After: 33.369s
```
```
osu.Game.Tests full:

Before: 14min
After: 8.5min
```

### DIFF
--- a/osu.Framework/Graphics/Rendering/Dummy/DummyRenderer.cs
+++ b/osu.Framework/Graphics/Rendering/Dummy/DummyRenderer.cs
@@ -50,7 +50,7 @@ namespace osu.Framework.Graphics.Rendering.Dummy
         public DummyRenderer()
         {
             maskingInfo = default;
-            WhitePixel = new Texture(new DummyNativeTexture(this), WrapMode.None, WrapMode.None);
+            WhitePixel = new TextureWhitePixel(new Texture(new DummyNativeTexture(this), WrapMode.None, WrapMode.None));
         }
 
         bool IRenderer.VerticalSync { get; set; } = true;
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Rendering/Dummy/DummyRenderer.cs
```diff
@@ -50,7 +50,7 @@ public sealed class DummyRenderer : IRenderer
         public DummyRenderer()
         {
             maskingInfo = default;
-            WhitePixel = new Texture(new DummyNativeTexture(this), WrapMode.None, WrapMode.None);
+            WhitePixel = new TextureWhitePixel(new Texture(new DummyNativeTexture(this), WrapMode.None, WrapMode.None));
```
smoogipoo, line 53: Required to expose the correct opacity for `TestSceneFrontToBackBox.TestOpaqueBoxWithMixedBlending`. This matches the implementation of other renderers (minus passing through an atlas).

peppy (approved):

Issue #5811, opened 2023-05-25
Android templates are missing from template projects


peppy:
For some reason they are just missing android. Very weird.

### Discussed in https://github.com/ppy/osu-framework/discussions/5808

<div type='discussions-op-text'>

<sup>Originally posted by **hwabis** May 25, 2023</sup>
I only see Desktop and iOS when I make a new template. Is there a reason that Android is not included? (SampleGame.Android ran fine for me so I don't see why not.)</div>

PR #5813, opened 2023-05-26
Add Android to project templates

hwabis:
Closes https://github.com/ppy/osu-framework/issues/5811. I have checked that the generated projects (Empty and FlappyDon) build fine (for Android and Desktop).

Note for the future: to generate projects using local templates, first run `./replace-references.sh 2023.521.0` (replace version if needed), then pack and install.

### DIFF
--- a/osu-framework.sln
+++ b/osu-framework.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29409.12
+# Visual Studio Version 17
+VisualStudioVersion = 17.3.32811.315
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "osu.Framework", "osu.Framework\osu.Framework.csproj", "{C76BF5B3-985E-4D39-95FE-97C9C879B83A}"
 EndProject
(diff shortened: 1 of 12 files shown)


osu-framework.sln
```diff
@@ -153,7 +157,7 @@ Global
 		{BBC0D18F-8595-43A6-AE61-5BF36A072CCE}.Release|iPhone.Build.0 = Release|Any CPU
 		{BBC0D18F-8595-43A6-AE61-5BF36A072CCE}.Release|iPhoneSimulator.ActiveCfg = Release|Any CPU
 		{BBC0D18F-8595-43A6-AE61-5BF36A072CCE}.Release|iPhoneSimulator.Build.0 = Release|Any CPU
-		{529D5E2E-774A-4831-9C9E-59E3E8DFF155}.Debug|Any CPU.ActiveCfg = Debug|iPhoneSimulator
+		{529D5E2E-774A-4831-9C9E-59E3E8DFF155}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
```
peppy, line 160: This looks a bit weird

hwabis: I'm not sure why it's like that or how to check whether it's correct...

frenzibyte:
> Note for the future: to generate projects using local templates, first run ./replace-references.sh 2023.521.0 (replace version if needed), then pack and install.

That's not the intended use of that script, it's just a helper tool for our template package deployment workflow.

hwabis:
> > Note for the future: to generate projects using local templates, first run ./replace-references.sh 2023.521.0 (replace version if needed), then pack and install.
> 
> That's not the intended use of that script, it's just a helper tool for our template deployment workflow.

The project reference paths in the generated project were all messed up if I didn't run it. I also tried generating using locally installed templates of `master` and it had the same issue. spaceman_atlas in Discord said he also got the same thing so I assumed the script was somehow necessary. We could doing something wrong?

frenzibyte:
I may have misread what you said. If you mean testing template packages locally then yeah you'll need to use the script. We should probably have another generalised script that does the whole process of packing the template (including replacing references), but that's for another day.

PR #5815, opened 2023-05-29
Fix not being able to parse empty string into nullable bindable

smoogipoo:
Previously:
image

My particular use case is I want to set a realm setting to `null`, but we currently store and retrieve them from the database as non-nullable strings. Not that having them stored as `null` would help (see `TestParseNullIntoNullableValueType` failure above).

Store: https://github.com/ppy/osu/blob/ba5f76fcc8b7e926310f22a6c5fc63bae19897fb/osu.Game/Rulesets/Configuration/RulesetConfigManager.cs#L72 (i.e. `Bindable<T>.ToString()` which returns `string.Empty` for a `null` value).

Retrieve: https://github.com/ppy/osu/blob/ba5f76fcc8b7e926310f22a6c5fc63bae19897fb/osu.Game/Configuration/RealmRulesetSetting.cs#L20-L21.

In summary:
```
Bindable<int>.Parse(null)               => ArgumentNullException()
Bindable<int>.Parse(string.Empty)       => FormatException()
Bindable<int?>.Parse(null)              => null
Bindable<int?>.Parse(string.Empty)      => null

Bindable<Class>.Parse(null)             => null (with and without NRT)
Bindable<Class>.Parse(string.Empty)     => null (with and without NRT)
Bindable<Class?>.Parse(null)            => null
Bindable<Class?>.Parse(string.Empty)    => null

Bindable<string>.Parse(null)            => null (with and without NRT)
Bindable<string>.Parse(string.Empty)    => string.Empty (with and without NRT)
Bindable<string?>.Parse(null)           => null
Bindable<string?>.Parse(string.Empty)   => string.Empty
```

### DIFF
--- a/osu.Framework.Tests/Bindables/BindableTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableTest.cs
@@ -195,7 +195,7 @@ namespace osu.Framework.Tests.Bindables
             bindable.Parse(string.Empty);
             Assert.That(bindable.Value, Is.Null);
         }
-#nullable restore
+#nullable disable
 
         [Test]
         public void TestParseNullIntoStringType()
@@ -245,7 +245,7 @@ namespace osu.Framework.Tests.Bindables
             bindable.Parse(string.Empty);
             Assert.That(bindable.Value, Is.Empty);
         }
-#nullable restore
+#nullable disable
 
         private static IEnumerable<object[]> getParsingConversionTests()
         {
--- a/osu.Framework.Tests/Bindables/BindableTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableTest.cs
@@ -215,7 +215,23 @@ namespace osu.Framework.Tests.Bindables
 
 #nullable enable
         [Test]
-        public void TestParseNullIntoNullableStringType()
+        public void TestParseNullIntoStringTypeWithNRT()
+        {
+            Bindable<string> bindable = new Bindable<string>();
+            bindable.Parse(null);
+            Assert.That(bindable.Value, Is.Null);
+        }
+
+        [Test]
+        public void TestParseEmptyStringIntoStringTypeWithNRT()
+        {
+            Bindable<string> bindable = new Bindable<string>();
+            bindable.Parse(string.Empty);
+            Assert.That(bindable.Value, Is.Empty);
+        }
+
+        [Test]
+        public void TestParseNullIntoNullableStringTypeWithNRT()
         {
             Bindable<string?> bindable = new Bindable<string?>();
             bindable.Parse(null);
@@ -223,7 +239,7 @@ namespace osu.Framework.Tests.Bindables
         }
 
         [Test]
-        public void TestParseEmptyStringIntoNullableStringType()
+        public void TestParseEmptyStringIntoNullableStringTypeWithNRT()
         {
             Bindable<string?> bindable = new Bindable<string?>();
             bindable.Parse(string.Empty);
--- a/osu.Framework.Tests/Bindables/BindableTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableTest.cs
@@ -107,6 +107,97 @@ namespace osu.Framework.Tests.Bindables
             Assert.That(value, Is.EqualTo(output));
         }
 
+        // Bindable<int>.Parse(null)
+        [Test]
+        public void TestParseNullIntoValueType()
+        {
+            Bindable<int> bindable = new Bindable<int>();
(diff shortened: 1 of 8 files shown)


osu.Framework/Bindables/Bindable.cs
```diff
@@ -262,7 +272,15 @@ public virtual void Parse(object input)
                     Value = bindable.Value;
                     break;
 
+                case string str when string.IsNullOrEmpty(str):
+                    // This is slightly different from the `null` case.
+                    // Consider a non-nullable default-initialised struct for which `ToString()` outputs an empty string.
+                    Value = default;
+                    break;
```
bdach, line 279: I find this case... particularly peculiar. It is operating on a lot of assumptions (e.g. that for every single non-nullable bindable there can ever be, an empty string is allowed to coerce to that bindable's inner type's `default`). Is this needed for anything?

I find the test cases that exercise this path kinda bizarre. `Bindable<int>` parsing empty strings into 0 feels just straight up weird. Built-in `int.Parse()` vomits on empty strings.

bdach: To compound the above, I wrote the following test:

```diff
diff --git a/osu.Framework.Tests/Bindables/BindableTest.cs b/osu.Framework.Tests/Bindables/BindableTest.cs
index 9fdb63118..8d52a5d1b 100644
--- a/osu.Framework.Tests/Bindables/BindableTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableTest.cs
@@ -196,6 +196,24 @@ public void TestParseEmptyStringIntoNullableReferenceTypeWithNRT()
             bindable.Parse(string.Empty);
             Assert.That(bindable.Value, Is.Null);
         }
+
+        // Bindable<string>.Parse(string.Empty)
+        [Test]
+        public void TestParseEmptyStringIntoNonNullableString()
+        {
+            Bindable<string> bindable = new Bindable<string>();
+            bindable.Parse(string.Empty);
+            Assert.That(bindable.Value, Is.Empty);
+        }
+
+        // Bindable<string>.Parse(string.Empty)
+        [Test]
+        public void TestParseEmptyStringIntoNullableString()
+        {
+            Bindable<string?> bindable = new Bindable<string?>();
+            bindable.Parse(string.Empty);
+            Assert.That(bindable.Value, Is.Empty);
+        }
 #nullable disable
 
         private static IEnumerable<object[]> getParsingConversionTests()
```

While they _pass_, they pass by what appears to either be sheer luck, or very careful placement. In the case of `Bindable<string>`, the `case T t` label is picked rather than the `case string str when string.IsNullOrEmpty(str)` label.

Maybe this was intentional, and maybe we want to keep this behaviour, but even if both are true, I'd say that maybe we want the above test cases to be in code too, just in case.

peppy: At a high level, I'd prefer matching the behaviour of .NET parse methods (throwing rather than graceful `default`). But need more context on whether this was done for a specific reason as per above.

smoogipoo: That's fair. So in your view all `string.Empty` cases should throw the same exception as `int.Parse()`, but they're fine to keep for nullables? In my particular case, as my OP mentions, I only need the `string.Empty` -> `null` coercion but maybe there's a better way of doing it that I haven't discovered yet.

bdach: For non-string nullable types, I can agree with parsing `string.Empty` to `null` for sure.

smoogipoo: I've adjusted the implementation a bit, see if you agree with the new results.

osu.Framework.Tests/Bindables/BindableTest.cs
```diff
@@ -196,7 +195,33 @@ public void TestParseEmptyStringIntoNullableReferenceTypeWithNRT()
...
+
+#nullable enable
+        [Test]
+        public void TestParseEmptyStringIntoNullableStringType()
+        {
+            Bindable<string?> bindable = new Bindable<string?>();
+            bindable.Parse(string.Empty);
+            Assert.That(bindable.Value, Is.Empty);
+        }
+#nullable restore
```
peppy, line 248: Was this change intended? Causing new inspections.

smoogipoo: Interesting, I missed that inspection, will revert. I thought this would also restore to file-level nullability context, but it looks to only restore to project-level (which this file overrides for... some reason).

peppy (dismissed):

PR #5820, opened 2023-05-31
Improve iOS/iPadOS frame timings by using `CADisplayLink`

peppy:
Alternative to and closes https://github.com/ppy/osu-framework/pull/5803 (the iOS portion – the macOS part is still
valid but only requires a veldrid version bump).

Testing at 2x limiter, 120hz display (iPad Pro 2018), release build local framework

| Header      | menu (contracted) | menu (expanded) | song select |
|-------------|-------------------|-----------------|-------------|
| `master`    | 120fps                 | 70fps               | 28fps           |
| this PR     | 120fps            | 80fps           | 28fps       |
| `master` gl | 120fps                 | 70fps               | 27fps           | 

Not much difference, but that's because we're still doing way too much work in draw. The main thing I want to show is that nothing regressed. Once we have more overhead this is only going to help more.

Of note, in a very simple framework visual test I *can* feel a reduction in input latency on this PR, similar to what I reported on macOS.

### DIFF
--- a/osu.Framework.iOS/IOSWindow.cs
+++ b/osu.Framework.iOS/IOSWindow.cs
@@ -51,6 +51,14 @@ namespace osu.Framework.iOS
 
         protected override void RunMainLoop()
         {
+            // Delegate running the main loop to CADisplayLink.
+            //
+            // Note that this is most effective in single thread mode.
+            // .. In multi-threaded mode it will time the *input* thread to the callbacks. This is kinda silly,
+            // but users shouldn't be using multi-threaded mode in the first place. Disabling it completely on
+            // iOS may be a good forward direction if this ever comes up, as a user may see a potentially higher
+            // frame rate with multi-threaded mode turned on, but it is going to give them worse input latency
+            // and higher power usage.
             SDL.SDL_iPhoneSetEventPump(SDL.SDL_bool.SDL_FALSE);
             SDL.SDL_iPhoneSetAnimationCallback(SDLWindowHandle, 1, _ => RunFrame(), IntPtr.Zero);
         }
--- a/osu.Framework.iOS/IOSWindow.cs
+++ b/osu.Framework.iOS/IOSWindow.cs
@@ -1,6 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using System.Diagnostics;
 using System.Drawing;
 using ObjCRuntime;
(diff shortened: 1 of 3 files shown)


osu.Framework.iOS/IOSWindow.cs
```diff
@@ -48,6 +49,20 @@ public override void Create()
...
+            //
+            // Note that this is most effective in single thread mode.
+            // .. In multi-threaded mode it will time the *input* thread to the callbacks. This is kinda silly,
+            // but users shouldn't be using multi-threaded mode in the first place. Disabling it completely on
+            // iOS may be a good forward direction if this ever comes up, as a user may see a potentially higher
+            // frame rate with multi-threaded mode turned on, but it is going to give them worse input latency
+            // and higher power usage.
+            SDL.SDL_iPhoneSetEventPump(SDL.SDL_bool.SDL_FALSE);
+            SDL.SDL_iPhoneSetAnimationCallback(SDLWindowHandle, 1, _ => RunFrame(), IntPtr.Zero);
+        }
```
smoogipoo, line 64: So I take it there's no need to block on anything here, akin to `while (Exists) { ... }`?

peppy: I originally had precisely that, but turns out that we already support non-blocking `Run` – it's used in android still.

Check out the `CanExit` flag which saves the day: https://github.com/search?q=repo%3Appy%2Fosu-framework%20canexit&type=code

osu.Framework.iOS/IOSWindow.cs
```diff
@@ -48,6 +49,20 @@ public override void Create()
...
+        protected override void RunMainLoop()
+        {
+            // Delegate running the main loop to CADisplayLink.
+            //
+            // Note that this is most effective in single thread mode.
+            // .. In multi-threaded mode it will time the *input* thread to the callbacks. This is kinda silly,
+            // but users shouldn't be using multi-threaded mode in the first place. Disabling it completely on
+            // iOS may be a good forward direction if this ever comes up, as a user may see a potentially higher
+            // frame rate with multi-threaded mode turned on, but it is going to give them worse input latency
+            // and higher power usage.
```
smoogipoo, line 61: Still not sure whether it's a good idea to do _everything_ in one frame. Out of curiosity, have you tried putting the DL on the draw thread, or know if that's even possible?

My thought process here is, combining update + draw in one cycle not only increases our chance to be put into a slower vsync profile, but is also not the lowest input-to-display latency we can achieve.

For example, suppose....

- The update thread keeps processing in the background.
- A display-link callback arrives, creating a draw frame.
- In the draw frame, we consume the latest update thread, sample input, post the sampled input back to the update frame, AND trigger a "last-chance" `IRequireHighFrequencyMousePosition` on that consumed update frame. This would be where we update e.g. cursor position to represent the most physically accurate position possible.

Supposing that optimistically a draw takes 8ms, the async update takes 3ms, and the high frequency update I'm proposing is instantaneous,
- Approach circles would be at most 11ms behind their true scale.
- The cursor would be at most 8ms ± touch panel latency behind its true position.

peppy: From a power perspective I'm not willing to consider anything else for now. We can revisit this in the future but I think it is in our best interest to reduce frame times as a priority instead.

Even at 45fps the input latency is absolutely amazing to play with.

osu.Framework.iOS/IOSWindow.cs
smoogipoo: Right, that makes sense. I suppose on IOS this is all wrapped inside a `UIApplicationMain` or whatever which likely does all that blocking internally.

smoogipoo (approved):

PR #5822, opened 2023-05-31
Change Tablet AreaSize and AreaOffset to Bindable<Vector2?>

Dimmitsaras:
For the edge case of `AreaOffset == 0,0` described here https://github.com/ppy/osu/pull/23628

There is one issue with this; somewhere along the way `.Value` gets initialized to `0,0` which means `if (AreaSize.Value == null)` and `if (AreaOffset.Value == null)` are never true.  I need some help figuring out what the cause could be (Are pull requests a good way to ask for help like this?)

For now, I'm forcing default values on restart regardless of user configuation. If you'd like I can commit the game side changes on that PR (or open a separate one), on which I have undone my changes and changed the bindable bindings to use callbacks instead.

BTW this is probably terrible, especially the null-coalescing spam as safeguards

### DIFF
--- a/osu.Framework/Input/Handlers/Tablet/ITabletHandler.cs
+++ b/osu.Framework/Input/Handlers/Tablet/ITabletHandler.cs
@@ -17,12 +17,12 @@ namespace osu.Framework.Input.Handlers.Tablet
         /// <summary>
         /// The offset of the area which should be mapped to the game window.
         /// </summary>
-        Bindable<Vector2> AreaOffset { get; }
+        Bindable<Vector2?> AreaOffset { get; }
 
         /// <summary>
         /// The size of the area which should be mapped to the game window.
         /// </summary>
-        Bindable<Vector2> AreaSize { get; }
+        Bindable<Vector2?> AreaSize { get; }
 
         /// <summary>
         /// Information on the currently connected tablet device. May be null if no tablet is detected.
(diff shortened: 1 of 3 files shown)


osu.Framework/Input/Handlers/Tablet/ITabletHandler.cs
```diff
@@ -34,6 +34,26 @@ public interface ITabletHandler
...
+
+        Vector2 GetAreaOffset()
+        {
+            return AreaOffset.Value ?? Vector2.One * 50f;
+        }
+
+        Vector2 GetAreaSize()
+        {
+            return AreaSize.Value ?? Vector2.One * 50f;
+        }
```
peppy, line 55: This isn't going to float. You've created a second way of accessing the same data which is already exposed via bindables. You could bypass this logic by still setting the `Bindable.Value` directly.

The only way this could be a thing is if the bindables are *removed* from the interface.

osu.Framework/Input/Handlers/Tablet/ITabletHandler.cs
```diff
@@ -34,6 +34,26 @@ public interface ITabletHandler
...
+
+        void UpdateAreaSize(Vector2 areaSize)
+        {
+            AreaSize.Value = areaSize;
+        }
+
+        Vector2 GetAreaOffset()
+        {
+            return AreaOffset.Value ?? Vector2.One * 50f;
```
peppy, line 49: What is 50f here? Needs to be a constant and fully documented as to what it's doing.

PR #5823, opened 2023-06-01
Allow selecting all text in a textbox programatically

peppy:
Arguably we could also add a `Focus` method since it's such a common operation for text boxes.


peppy:
I already added a test for it, could you explain what you're hoping for?

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -999,6 +996,29 @@ private void setText(string value)
...
+        /// This will force focus if not already focused.
+        /// </remarks>
+        public void SelectAll()
+        {
+            if (!HasFocus)
+            {
+                GetContainingInputManager().ChangeFocus(this);
+                Schedule(SelectAll);
+                return;
+            }
```
bdach, line 1012: I find this focus-hogging behaviour a little dodgy. To demonstrate why, here's an example test:

```diff
diff --git a/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs b/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
index 02d6387ca..edccb4c68 100644
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
@@ -825,6 +825,26 @@ public void TestSetTextSelection()
             AddAssert("nothing selected", () => textBox.SelectedText == string.Empty);
         }
 
+        [Test]
+        public void TestSelectAllWhileHiddenHasNoEffect()
+        {
+            TextBox textBox = null;
+
+            AddStep("add hidden textbox", () =>
+            {
+                textBoxes.Add(textBox = new BasicTextBox
+                {
+                    Size = new Vector2(300, 40),
+                    Text = "initial text",
+                    Alpha = 0
+                });
+            });
+
+            AddStep("select all", () => textBox.SelectAll());
+            AddStep("unhide textbox", () => textBox.Show());
+            AddAssert("no selection in textbox", () => textBox.SelectedText, () => Is.Empty);
+        }
+
         private void prependString(InsertableTextBox textBox, string text)
         {
             InputManager.Keys(PlatformAction.MoveBackwardLine);
```

The test case is inspired by the fact that focus will fail to change to non-present drawables. I'm not *entirely* sure what should happen in this scenario - but I would *probably* not expect `SelectAll()` to have this delayed effect wherein it takes away focus from somewhere else but only when the textbox is unhidden. It should probably either take away focus instantly if able, or abort the selection if unable. Is there a need for this particular greedy behaviour?

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -999,6 +996,29 @@ private void setText(string value)
...
+
+            var lastSelectionBounds = getTextSelectionBounds();
+
+            selectionStart = 0;
+            selectionEnd = text.Length;
+            cursorAndLayout.Invalidate();
+            onTextSelectionChanged(TextSelectionType.All, lastSelectionBounds);
```
Susko3, line 1019:
```suggestion
```
(Also requires rolling back the `PlatformAction.SelectAll` change, or having `private void selectAll(bool userEvent)` so it can still raise the event.)

Since this `SelectAll` is triggered programmatically, it shouldn't raise the user `OnTextSelectionChanged` event. This matches the behaviour of `OnUserText{Added,Removed}` when manually setting `Text`.

Raising the event already regresses the osu! side change, as it'll play the 'mouse click' and 'select all' samples at the same time, resulting in potentially unwanted clashing.

`SelectAll()` not raising the event should be tested in `TestSceneTextBoxEvents`.

Susko3 (changes requested):

osu.Framework/Graphics/UserInterface/TextBox.cs
frenzibyte: Removed in https://github.com/ppy/osu-framework/pull/5823/commits/768a37de14102d6ade5566de2814542d7477a818

osu.Framework/Graphics/UserInterface/TextBox.cs
frenzibyte: Resolved in https://github.com/ppy/osu-framework/pull/5823/commits/768a37de14102d6ade5566de2814542d7477a818

frenzibyte (approved):

PR #5829, opened 2023-06-09
Add disk cache for shader compilations

smoogipoo:
Shader compilation dominates our startup time, so they're now cached to disk (spirv + cross-compile + reflection) at `cache/shaders/`.

| Test | Before | After (cold) | After (hot) |
| --- | --- | --- | --- |
| framework | image | image | image
| osu | image | image | image |

* For framework, only the main thread is pictured. Snapshot lasts until the game is loaded.
* For osu, all threads are pictured because we use the async loader spinner to load shaders at startup. Snapshot lasts until the intro starts displaying (after the loading spinner).

On Vulkan, the refactoring means only one compilation occurs now:

| Before | After (cold) |
| --- | --- |
| image | image |

The results are also significant on the D3D and Metal surfaces which have to compile twice - once to GLSL for reflection and a second time to the destination platform. I haven't added profiling results for these but they can be imagined by doubling every duration from the first table.

### DIFF
--- a/osu.Framework/Graphics/OpenGL/Shaders/GLShaderPart.cs
+++ b/osu.Framework/Graphics/OpenGL/Shaders/GLShaderPart.cs
@@ -123,10 +123,10 @@ namespace osu.Framework.Graphics.OpenGL.Shaders
 
                         if (!string.IsNullOrEmpty(backbufferCode))
                         {
-                            string realMainName = "real_main_" + Guid.NewGuid().ToString("N");
+                            const string real_main_name = "__internal_real_main";
 
-                            backbufferCode = backbufferCode.Replace("{{ real_main }}", realMainName);
-                            code = Regex.Replace(code, @"void main\((.*)\)", $"void {realMainName}()") + backbufferCode + '\n';
+                            backbufferCode = backbufferCode.Replace("{{ real_main }}", real_main_name);
+                            code = Regex.Replace(code, @"void main\((.*)\)", $"void {real_main_name}()") + backbufferCode + '\n';
                         }
                     }
                 }
--- a/osu.Framework/Graphics/OpenGL/Shaders/GLShaderPart.cs
+++ b/osu.Framework/Graphics/OpenGL/Shaders/GLShaderPart.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-#nullable disable
-
 using System;
 using System.Collections.Generic;
 using System.Globalization;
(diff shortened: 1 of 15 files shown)


peppy:
Awesome stuff.

For the record, if a user was to change their hardware, could this cause an issue? Put another way, is there already a pathway to recovery if a cached shader file is corrupt in some way?

smoogipoo:
The compilations are hardware-agnostic. We could even distribute these files ourselves if it came to it.

Both data and a checksum of the data is written. If there's hardware-induced/OS-induced corruption, the file is abandoned and recompilation takes place, with the result of that then replacing the corrupt files.

osu.Framework/Graphics/Shaders/ShaderCompilationStore.cs
```diff
@@ -0,0 +1,133 @@
...
+                FragmentText = crossResult.FragmentShader,
+                Reflection = crossResult.Reflection
+            };
+
+            saveToCache(filename, compilation);
+
+            return compilation;
+        }
+
+        public ComputeProgramCompilation CompileCompute(string programText, CrossCompileTarget target)
```
bdach, line 47: Looks unused - is this intended for future consumption somewhere else?

osu.Framework/Graphics/OpenGL/Shaders/GLShaderPart.cs
```diff
@@ -123,10 +121,10 @@ private string loadFile(byte[] bytes, bool mainFile)
 
                         if (!string.IsNullOrEmpty(backbufferCode))
                         {
-                            string realMainName = "real_main_" + Guid.NewGuid().ToString("N");
+                            const string real_main_name = "__internal_real_main";
```
bdach, line 124: Is this main name change significant in any way? I'm not sure what the guid was doing, and blame doesn't help, since it was added in, of all things, https://github.com/ppy/osu-framework/commit/b9406dd3881cddbdf19fcd81a28f227deb96991a, which just barely predates me.

bdach (commented):
Source looks good except for a few parts I'm not getting. I'll give this a test spin on the platforms I have access to later.

osu.Framework/Graphics/Shaders/ShaderCompilationStore.cs
smoogipoo: Yeah, I was thinking it could be used for https://github.com/ppy/osu-framework/pull/5757 for example.

osu.Framework/Graphics/OpenGL/Shaders/GLShaderPart.cs
smoogipoo: It's significant by keeping the shader code deterministic. The guid was only present to avoid naming collisions.

bdach: Oh, right, good point. I'm assuming the naming collisions you mention are no longer relevant. Makes sense :+1:

smoogipoo: `__internal_` should be as effective as guid, and if not then we could suffix an incrementing value like `__internal_real_main_1`, but I doubt that'll happen.

Once case I can realistically think of is if we end up wanting multiple `main` trampolines, but there's no support for that right now.

bdach:
Tested on Windows and Android. Can confirm sizeable gains in startup time on both. I don't have profiles to show from Android because Android but it does look significantly snappier when launching.

This is pretty much good to go as far as I'm concerned, but I won't push the button yet pending potentially an iOS check and maybe a decision as to when we want to get this out, I guess.

bdach (dismissed):

peppy:
Tested to work well here.

I've updated the logging around shader compilation to read better. I think this is fine to be merged, but will leave it approved for someone to check my latest commit.

peppy (approved):

bdach (approved):

PR #5830, opened 2023-06-09
Expose `Path.BackgroundColour` as a settable property

smoogipoo:


osu.Framework/Graphics/Lines/Path.cs
```diff
@@ -280,7 +280,17 @@ protected Texture Texture
         // The path should not receive the true colour to avoid colour doubling when the frame-buffer is rendered to the back-buffer.
         public override DrawColourInfo DrawColourInfo => new DrawColourInfo(Color4.White, base.DrawColourInfo.Blending);
 
-        public Color4 BackgroundColour => new Color4(0, 0, 0, 0);
+        private Color4 backgroundColour = new Color4(0, 0, 0, 0);
+
+        public Color4 BackgroundColour
```
bdach, line 285: My main question here would be if we can make this work automagically. I see the note in the OP that premultiplied alpha seems to not work. That said, game-side sliders are actually `SmoothPath`s and not plain `Path`s. `SmoothPath` has `ColourAt(float position)`. So what I'm trying to get at is that maybe we can initialise this to `ColourAt(0)` for `SmoothPath` specifically, and therefore maybe avoid the need to expose this, and avoid the game-side change?

I do note that the above proposal has a caveat that the `ColourAt(float)` method is not something that is aware of possible data dependencies. Which is to say, if you define the return of `ColourAt(float)` to be dependent on a `BorderColour` property - like the game does - then the framework-side code that uses `ColourAt(float)` to clear the framebuffer to the proper colour to avoid the bad colour interpolation wouldn't be aware of that change. *However* I also note that game [already has to invalidate the texture for the border colour to apply](https://github.com/ppy/osu/blob/659a04206544eb3c16b7d0433b4794734221275f/osu.Game.Rulesets.Osu/Skinning/Default/DrawableSliderPath.cs#L29), so it looks like this potentially works?

peppy (approved):

peppy:
@smoogipoo please check my xmldoc makes sense.

smoogipoo:
Looks fine.

bdach (approved):

PR #5831, opened 2023-06-10
Remove `Dropdown.AddDropdownItem(text, value)` overload

Susko3:
This overload is hard to work with, as I would expect that calling `GenerateItemText()` on an existing dropdown item to be idempotent. (I've had this expectation in mind when trying to properly fix https://github.com/ppy/osu/pull/23858.)

Since this overload is `protected`, no one really used it (zero osu!-side usages). So I hope it would be fine to remove.

# Breaking changes

## `AddDropdownItem(LocalisableString text, T value)` has been removed

If you were using this, refactor your code to use `LocalisableString GenerateItemText(T value)` instead.

Easiest way is to add `LocalisableString? CustomDropdownText` to your `T`, or wrap `T` in a class that has one.
Then simply change `GenerateItemText()` to use that:

```cs
public partial class MyDropdown : Dropdown<T>
{
    protected override LocalisableString GenerateItemText(T value)
        => value.CustomDropdownText ?? base.GenerateItemText(value);
}
```

### DIFF
--- a/osu.Framework/Graphics/UserInterface/Dropdown.cs
+++ b/osu.Framework/Graphics/UserInterface/Dropdown.cs
@@ -95,19 +95,12 @@ namespace osu.Framework.Graphics.UserInterface
         /// Add a menu item directly while automatically generating a label.
         /// </summary>
         /// <param name="value">Value selected by the menu item.</param>
-        public void AddDropdownItem(T value) => AddDropdownItem(GenerateItemText(value), value);
-
-        /// <summary>
-        /// Add a menu item directly.
-        /// </summary>
-        /// <param name="text">Text to display on the menu item.</param>
-        /// <param name="value">Value selected by the menu item.</param>
-        protected void AddDropdownItem(LocalisableString text, T value)
+        public void AddDropdownItem(T value)
         {
             if (boundItemSource != null)
                 throw new InvalidOperationException($"Cannot manually add dropdown items when an {nameof(ItemSource)} is bound.");
 
-            addDropdownItem(text, value);
+            addDropdownItem(GenerateItemText(value), value);
         }
 
         private void addDropdownItem(LocalisableString text, T value)
--- a/osu.Framework/Graphics/UserInterface/Dropdown.cs
+++ b/osu.Framework/Graphics/UserInterface/Dropdown.cs
@@ -63,7 +63,7 @@ namespace osu.Framework.Graphics.UserInterface
                 return;
 
             foreach (var entry in items)
-                addDropdownItem(GenerateItemText(entry), entry);
+                addDropdownItem(entry);
 
             if (Current.Value == null || !itemMap.Keys.Contains(Current.Value, EqualityComparer<T>.Default))
                 Current.Value = itemMap.Keys.FirstOrDefault();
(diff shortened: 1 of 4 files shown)


osu.Framework/Testing/Drawables/Sections/ToolbarAssemblySection.cs
```diff
@@ -59,11 +60,11 @@ private void load(TestBrowser browser)
             };
         }
 
-        public void AddAssembly(string name, Assembly assembly) => assemblyDropdown.AddAssembly(name, assembly);
+        public void AddAssembly(Assembly assembly) => assemblyDropdown.AddAssembly(assembly);
 
         private partial class AssemblyDropdown : BasicDropdown<Assembly>
         {
-            public void AddAssembly(string name, Assembly assembly)
+            public void AddAssembly(Assembly assembly)
```
bdach, line 67: Borderline out of scope but this method seems like a bit of a vestige of the old dynamic test compilation setup. I'm not sure we need it anymore, we could probably get rid of this method and just replace its usage sites with `AddDropdownItem()`.

osu.Framework/Graphics/UserInterface/Dropdown.cs
```diff
@@ -95,19 +95,12 @@ public IBindableList<T> ItemSource
...
-        /// <param name="text">Text to display on the menu item.</param>
-        /// <param name="value">Value selected by the menu item.</param>
-        protected void AddDropdownItem(LocalisableString text, T value)
+        public void AddDropdownItem(T value)
         {
             if (boundItemSource != null)
                 throw new InvalidOperationException($"Cannot manually add dropdown items when an {nameof(ItemSource)} is bound.");
 
-            addDropdownItem(text, value);
+            addDropdownItem(GenerateItemText(value), value);
```
bdach, line 103: So this is probably fine and all, but I do wonder where this is going? As in, I'm not seeing how this change gets us any closer to fixing the issue at hand. The `GenerateItemText()` call is still a one-shot, as in if the item gets added before the method can return proper text, it will still get stuck with the bad text until the item is removed and inserted again like what the PRed game-side workaround is doing.

I don't necessarily mind this change, I'd probably even say that it's a good idea, I'm just not sure where this stands in light of the fact that the change is supposed to be a gateway to fixing the bug in question.

Susko3: Since `GenerateItemText()` expects a fully loaded state, I'd like to defer it to `LoadComplete` to fix the issue. The code below is the fix, but it will override any custom text. So I made this PR to remove the ability to set custom text.

```diff
@@ -261,6 +271,9 @@ protected override void LoadComplete()
         {
             base.LoadComplete();

+            foreach (var item in MenuItems)
+                item.Text.Value = GenerateItemText(item.Value);
+
             Header.Label = SelectedItem?.Text.Value ?? default;
         }
```

I've just pushed a commit that shows that `GenerateItemText()` is only called in one place, so the other part of this fix is to only call that `GenerateItemText()` if `IsLoaded == true`. It's not a trivial diff as it also requires adding new ctors to `DropdownMenuItem` and `MenuItem` <-- my reasoning for why the change will be in a separate PR.

bdach: Okay, well, I think that is a path I could agree to. Calling `GenerateItemText()` if `!IsLoaded` shouldn't bring back the performance issues cited at https://github.com/ppy/osu-framework/pull/5516#issuecomment-1340768668 with font glyph lookups and the like, so I think I'm on board with this direction.

Will want to run this past by @peppy though since we've all been fighting with this a bit and I would like to avoid further take-backsies in the future if it can be avoided.

bdach: Actually, no, wait - you want to be running that *unconditionally* in `LoadComplete()`? Then that just brings back the lookup overheads again...

Not sure.

Susko3: Oh, then I can just move it to `LoadAsyncComplete`.

bdach: That sounds better yeah.

bdach (approved):

osu.Framework/Testing/Drawables/Sections/ToolbarAssemblySection.cs
peppy: Applied this change since it's a simple one.

peppy (approved):

PR #5832, opened 2023-06-10
Add include guards to all shaders

smoogipoo:
This is [standard practice](https://en.wikipedia.org/wiki/Include_guard) in the C/C++ world.


bdach (approved):

bdach:
Sure, but mostly out of curiosity, is this headed somewhere special?

smoogipoo:
There’s no followup to this, no. On occasion I have tried to write shaders and encountered this, particularly with sh_Utils.

bdach:
Tested android out of overabundance of caution, looks okay. Seems harmless enough, so I'll just click the button to get one thing off everyone's lists...

Issue #5621, opened 2023-01-10
`FocusedOverlayContainer` makes `PopIn`/`PopOut` optional despite requiring implementation


frenzibyte:
After wasting an hour wondering why a focused overlay container doesn't receive any input despite actually having content displayed, it turns out the visibility state was actually set to "hidden" ([and it doesn't seem like I was the only one](https://discord.com/channels/188630481301012481/188630652340404224/1062427654680219778)).

`VisibilityContainer` by default requires inheritors to implement `PopIn`/`PopOut`, but `FocusedOverlayContainer` overrides those to perform focus logic. This results in `FocusedOverlayContainer` inheritors to not have to implement `PopIn`/`PopOut` despite actually needing to.

One way I can see to improve this would be for `FocusedOverlayContainer` and other classes of similar usage to override `UpdateState` instead, so that `PopIn` and `PopOut` remain abstract for non-abstract inheritors to implement, but I'll leave this as an issue thread for more feedback first.

jai-x:
Inclined to agree with this. I had the same moment when implementing the chat overlay.
Although adding a new overlay isn't that common that you run into this often, it is something that would be better if fixed.

peppy:
The proposal sounds good.

PR #5834, opened 2023-06-11
Remove overrides of PopIn/PopOut in FocusedOverlayContainer

jai-x:
Resolves https://github.com/ppy/osu-framework/issues/5621

Moves focus logic outside of `PopIn`/`PopOut` methods in `FocusedOverlayContainer`

Forces  non-abstract inheritors of `FocusedOverlayContainer` to correctly implement these methods.

# Breaking changes

## Subclasses of `FocusedOverlayContainer` must now implement `Pop{In,Out}()`

Previously, subclasses of `FocusedOverlayContainer` were not forced to implement `Pop{In,Out}` themselves, despite actually needing to do so for the overlay to actually play transitions out correctly. This has now been changed and inheritors of the class must always implement `Pop{In,Out}`.

In practice this should not break any reasonable consumers of the class, except for the need to remove any `base.Pop{In,Out}()` calls in classes that directly inherit `FocusedOverlayContainer` and call base in `Pop{In,Out}()`.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/Popover.cs
+++ b/osu.Framework/Graphics/UserInterface/Popover.cs
@@ -19,13 +19,17 @@ namespace osu.Framework.Graphics.UserInterface
     /// It typically is activated by another control and includes an arrow pointing to the location from which it emerged.
     /// (loosely paraphrasing: https://developer.apple.com/design/human-interface-guidelines/ios/views/popovers/)
     /// </summary>
-    public abstract partial class Popover : FocusedOverlayContainer
+    public abstract partial class Popover : OverlayContainer
     {
         protected override bool BlockPositionalInput => true;
 
         public override bool ReceivePositionalInputAt(Vector2 screenSpacePos) => Body.ReceivePositionalInputAt(screenSpacePos) || Arrow.ReceivePositionalInputAt(screenSpacePos);
 
         public override bool HandleNonPositionalInput => State.Value == Visibility.Visible;
+
+        public override bool RequestsFocus => State.Value == Visibility.Visible;
+
+        public override bool AcceptsFocus => State.Value == Visibility.Visible;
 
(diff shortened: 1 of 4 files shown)


osu.Framework/Graphics/UserInterface/Popover.cs
```diff
@@ -19,14 +19,18 @@ namespace osu.Framework.Graphics.UserInterface
...
         protected override bool BlockPositionalInput => true;
 
         public override bool ReceivePositionalInputAt(Vector2 screenSpacePos) => Body.ReceivePositionalInputAt(screenSpacePos) || Arrow.ReceivePositionalInputAt(screenSpacePos);
 
         public override bool HandleNonPositionalInput => State.Value == Visibility.Visible;
 
+        public override bool RequestsFocus => State.Value == Visibility.Visible;
+
+        public override bool AcceptsFocus => State.Value == Visibility.Visible;
+
```
jai-x, line 33: This is a bit of an odd change.

Since `PopOver` was overriding `PopIn/PopOut` and `FocusedOverlayContainer` used those methods for automatic getting focus, it meant that `PopOver` would never automatically gain focus when shown.

Since moving the focus gain to `UpdateState` in `FocusedOverlayContainer`, this changed the focus behaviour of `PopOver` and broke tests.

The solution here is to not inherit from `FocusedOverlayContainer` and allow `PopOver` to manage it's own focus state.

Alternatively, `FocusedOverlayContainer` could have a new overridable propertly that `PopOver` could override?
```cs
public virtual bool GainFocusOnShow => true
```

peppy (approved):
Looks logical

peppy:
osu! side changes: https://github.com/ppy/osu/pull/23882

Will wait for one more review on this one.

bdach (approved):
Seems sane

PR #5836, opened 2023-06-12
Fix `Dropdown` calling `GenerateItemText()` before it was fully loaded

Susko3:
- Alternative to / closes https://github.com/ppy/osu/pull/23858

Uses `LoadAsyncComplete` to avoid perf regressions mentioned here https://github.com/ppy/osu-framework/pull/5516#issuecomment-1340768668.

The last test was added to cover the `GenerateItemText()` call in `selectionChanged()` (this one is always scheduled, so there is no issue).


peppy:
Have confirmed this doesn't regress performance.

peppy (approved):

bdach (approved):

PR #5838, opened 2023-06-12
Add new `DeferredStagingBuffer` type

smoogipoo:
On occasion, I have noticed the new `PersistentStagingBuffer` causing graphics corruption. At least, I _think_ it's because of this type and its undefined nature (part of the reason why we now have an `Unmap()` inside the copy function).

As a result, I'm going back to buffers to figure out the most efficient update strategy. This implements a new `DeferredStagingBuffer`, which uses a managed CPU buffer and an intermediate driver-side `Staging` buffer, to do updates.
The new type is off by default, and is only accessible with an environment variable:

<table>
<tr>
<th>Name</th>
<th>Values</th>
</tr>
<tr>
<td>

`OSU_GRAPHICS_STAGING_BUFFER_TYPE`

</td>
<td>

`0` = `ManagedStagingBuffer`
`1` = `PersistentStagingBuffer`
`2` = `DeferredStagingBuffer`

</td>
<tr>
<table>

I'd like to test this new type over time, with the intent of it replacing `PersistentStagingBuffer`. I'm currently unable to reproduce the graphics corruption I was seeing previously, so I intend to be able to reproduce that once more first to see whether it was a fluke.

| Type | Fullscreen? | Performance |
| --- | --- | --- |
| `ManagedStagingBuffer` | No | image |
| `ManagedStagingBuffer` | Yes | image |
| `PersistentStagingBuffer` | No | image |
| `PersistentStagingBuffer` | Yes | image |
| `DeferredStagingBuffer` | No | image |
| `DeferredStagingBuffer` | Yes | image |

### DIFF
--- a/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
+++ b/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
@@ -7,7 +7,6 @@ using System.Diagnostics;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
-using osu.Framework.Configuration;
 using osu.Framework.Development;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.Rendering;
--- a/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
+++ b/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using osu.Framework.Configuration;
 using osu.Framework.Development;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.Rendering;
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
```diff
@@ -603,21 +603,32 @@ protected override INativeTexture CreateNativeVideoTexture(int width, int height
...
-                case GraphicsBackend.Metal:
-                // OpenGL backends need additional work to support coherency and persistently mapped buffers.
-                case GraphicsBackend.OpenGL:
-                case GraphicsBackend.OpenGLES:
-                    return new ManagedStagingBuffer<T>(this, count);
+                    switch (Device.BackendType)
+                    {
+                        case GraphicsBackend.Direct3D11:
+                        case GraphicsBackend.Vulkan:
+                            return new PersistentStagingBuffer<T>(this, count);
```
peppy, line 622: To confirm, this is changing d3d back to persistent in a hope that the corruption was a fluke / is fixed? And the new deferred buffer is a fallback for users to potentially test if corruption does come back?

smoogipoo: This is not changing D3D/Vulkan.

> And the new deferred buffer is a fallback for users to potentially test if corruption does come back?

Correct.

peppy: Ah, I misread.

peppy (approved):

PR #5839, opened 2023-06-13
Move midi initialisation to a thread

peppy:
MIDI is rarely used. On top of that, on macOS the overhead on first call to the `Devices` enum is very expensive (>1s). This shouldn't delay startup of the game.

JetBrains Rider 2023-06-13 at 06 18 12

Implementation is copied from `OpenTabletDriveHandler`, which should be quite reliable.

### DIFF
--- a/osu.Framework/Input/Handlers/Midi/MidiHandler.cs
+++ b/osu.Framework/Input/Handlers/Midi/MidiHandler.cs
@@ -77,7 +77,7 @@ namespace osu.Framework.Input.Handlers.Midi
                 }
             }, true);
 
-            return refreshDevices();
+            return true;
         }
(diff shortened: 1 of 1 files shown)


smoogipoo:
Is the `return refreshDevices()` not also relevant here?

peppy:
Hold off, this is really dodgy.

peppy:
> Is the `return refreshDevices()` not also relevant here?

yes.

peppy:
After 06d630ba1e052ac7bcff515e69b31d85679e41a4:

JetBrains Rider 2023-06-13 at 06 37 43

smoogipoo (approved):

PR #5843, opened 2023-06-16
Fix performance when viewing the tail end of a long / zoomed-in waveform

peppy:
Closes https://github.com/ppy/osu/issues/23905

| before | after |
| -- | -- |
|osu Framework Tests 2023-06-16 at 07 41 08|osu Framework Tests 2023-06-16 at 07 40 21|

### DIFF
--- a/osu.Framework/Graphics/Audio/WaveformGraph.cs
+++ b/osu.Framework/Graphics/Audio/WaveformGraph.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-#nullable disable
-
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -29,8 +27,8 @@ namespace osu.Framework.Graphics.Audio
     /// </summary>
     public partial class WaveformGraph : Drawable
     {
-        private IShader shader;
-        private Texture texture;
+        private IShader shader = null!;
+        private Texture texture = null!;
 
         [BackgroundDependencyLoader]
         private void load(ShaderManager shaders, IRenderer renderer)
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Audio/WaveformGraph.cs
```diff
@@ -342,17 +340,16 @@ public override void Draw(IRenderer renderer)
 
                 float separation = drawSize.X / (points.Count - 1);
 
-                for (int i = 0; i < points.Count - 1; i++)
+                // Equates to making sure that rightX >= localMaskingRectangle.Left (see assertions in loop, which enforce the same thing).
+                // Without this pre-check, very long waveform displays can get slow just from running the loop below (point counts in excess of 1mil).
+                int startIndex = (int)Math.Clamp(localMaskingRectangle.Left / separation, 0, points.Count - 1);
+                int endIndex = (int)Math.Clamp(localMaskingRectangle.Right / separation + 1, 0, points.Count - 1);
```
peppy, line 346: The index choice here seems correct (visual inspection). Translating these from the original conditions gets a bit dicey with the `<` `>`, but it's also fine if we err on the side of caution and make them include one more on each side.

So if this comes up in review, it's fine to decrement / increment the start / end if required. But visually this does seem to be what we want. Maybe adjusting `MaskingSmoothness` in a masking scenario might require more?

osu.Framework/Graphics/Audio/WaveformGraph.cs
```diff
@@ -342,17 +340,16 @@ public override void Draw(IRenderer renderer)
 
                 float separation = drawSize.X / (points.Count - 1);
 
-                for (int i = 0; i < points.Count - 1; i++)
+                // Equates to making sure that rightX >= localMaskingRectangle.Left (see assertions in loop, which enforce the same thing).
```
bdach, line 343: What assertions in the loop is this referring to? I don't see any there. Is this a vestige of some draft implementation which survived until now despite the corresponding asserts going away?

osu.Framework/Graphics/Audio/WaveformGraph.cs
bdach: Looking at the maths this does appear to be correct to me. Doesn't seem like any adjustments are needed.

osu.Framework/Graphics/Audio/WaveformGraph.cs
peppy: I removed the asserts in the end. It didn't seem useful.

bdach (approved):

PR #5844, opened 2023-06-16
Remove automatic conversion of vertical scroll to horizontal when holding shift

peppy:
This was breaking binding behaviour too often, and is a pretty niche use case. `ScrollContainer` already automatically converts vertical scroll to horizontal, so this really doesn't do much for us.

Also in 092f28c0a, I've refactored `ScrollContainer` scroll delta handling logic to be more understandable

It was previously not clear the `Y` delta would be used if no `X` delta was present.

Note that this also adds considerations for precision scroll devices.  Without it, dual-axis scrolls could feel quite weird if the user happens to scroll a perfectly vertical line (would be unexpectedly converted to horizontal).

### DIFF
--- a/osu.Framework/Graphics/Containers/ScrollContainer.cs
+++ b/osu.Framework/Graphics/Containers/ScrollContainer.cs
@@ -397,14 +397,18 @@ namespace osu.Framework.Graphics.Containers
                     return false;
             }
 
-            bool isPrecise = e.IsPrecise;
+            float scrollDeltaFloat = ScrollDirection == Direction.Horizontal
+                ? e.ScrollDelta.X
+                : e.ScrollDelta.Y;
 
-            Vector2 scrollDelta = e.ScrollDelta;
-            float scrollDeltaFloat = scrollDelta.Y;
-            if (ScrollDirection == Direction.Horizontal && scrollDelta.X != 0)
-                scrollDeltaFloat = scrollDelta.X;
+            // Let's assume precision scroll devices have both axes of scroll available.
+            //
+            // Then, for non-precision scroll devices, let's assuming the user's intention when scrolling
+            // is to *scroll*, regardless of the ScrollDirection.
+            if (!e.IsPrecise && scrollDeltaFloat == 0)
+                scrollDeltaFloat = e.ScrollDelta.Y + e.ScrollDelta.X;
 
-            scrollByOffset(ScrollDistance * -scrollDeltaFloat, true, isPrecise ? 0.05 : DistanceDecayScroll);
+            scrollByOffset(ScrollDistance * -scrollDeltaFloat, true, e.IsPrecise ? 0.05 : DistanceDecayScroll);
             return true;
         }
 
--- a/osu.Framework/Input/StateChanges/MouseScrollRelativeInput.cs
+++ b/osu.Framework/Input/StateChanges/MouseScrollRelativeInput.cs
@@ -34,9 +34,6 @@ namespace osu.Framework.Input.StateChanges
 
             if (Delta != Vector2.Zero)
             {
-                if (!IsPrecise && Delta.X == 0 && state.Keyboard.ShiftPressed)
-                    Delta = new Vector2(Delta.Y, 0);
-
                 var lastScroll = mouse.Scroll;
                 mouse.Scroll += Delta;
                 mouse.LastSource = this;


frenzibyte (changes requested):
The changes look solid. Unfortunately, on macOS, the shift-scroll feature is applied at an OS level, so we'll actually need to have logic that flips the X component of the scroll delta back to Y, and I'm not 100% sure whether we should do that. Going to wait for your call on this.

https://github.com/ppy/osu-framework/assets/22781491/9b65a7ce-9ac7-4bc9-8620-54ca10176745

osu.Framework/Input/StateChanges/MouseScrollRelativeInput.cs
```diff
@@ -34,9 +34,6 @@ public void Apply(InputState state, IInputStateChangeHandler handler)
 
             if (Delta != Vector2.Zero)
             {
-                if (!IsPrecise && Delta.X == 0 && state.Keyboard.ShiftPressed)
-                    Delta = new Vector2(Delta.Y, 0);
-
```
Susko3, line 39: This change breaks the following behaviour:

- `TestNestedScrolling(Horizontal, Vertical)`
- `IsPrecise = false`
- <kbd>Shift</kbd> held
- scrolling vertically (`Y`)
- mouse over inner (green) container

`master`: scrolls the outer (yellow) container horizontally
`this PR`:  scrolls the inner container vertically

This is due to the fact that `scrollWasMostlyHorizontal == false`, even if using vertical+shift, which should be considered horizontal.

I suggest replicating the exact behaviour of the removed code inside `ScrollContainer`:

```diff
diff --git a/osu.Framework/Graphics/Containers/ScrollContainer.cs b/osu.Framework/Graphics/Containers/ScrollContainer.cs
index a9609b2d1..9709c0137 100644
--- a/osu.Framework/Graphics/Containers/ScrollContainer.cs
+++ b/osu.Framework/Graphics/Containers/ScrollContainer.cs
@@ -387,9 +387,14 @@ protected override bool OnScroll(ScrollEvent e)
             if (Content.AliveInternalChildren.Count == 0)
                 return false;

+            var scrollDelta = e.ScrollDelta;
+
+            if (!e.IsPrecise && scrollDelta.X == 0 && e.CurrentState.Keyboard.ShiftPressed)
+                scrollDelta = new Vector2(scrollDelta.Y, 0);
+
             if (parentScrollContainer != null && parentScrollContainer.ScrollDirection != ScrollDirection)
             {
```

Then use `scrollDelta` instead of `e.ScrollDelta` in later calculations.

osu.Framework/Graphics/Containers/ScrollContainer.cs
```diff
@@ -397,14 +397,18 @@ protected override bool OnScroll(ScrollEvent e)
...
+                : e.ScrollDelta.Y;
+
+            // Let's assume precision scroll devices have both axes of scroll available.
+            //
+            // Then, for non-precision scroll devices, let's assuming the user's intention when scrolling
+            // is to *scroll*, regardless of the ScrollDirection.
+            if (!e.IsPrecise && scrollDeltaFloat == 0)
+                scrollDeltaFloat = e.ScrollDelta.Y + e.ScrollDelta.X;
+
+            scrollByOffset(ScrollDistance * -scrollDeltaFloat, true, e.IsPrecise ? 0.05 : DistanceDecayScroll);
```
Susko3, line 411: This "refactor" breaks the following behaviour:


- `TestNestedScrolling(Vertical, Horizontal)`
- `IsPrecise = true`
- scrolling vertically (`Y`)
- mouse over inner container

`master`: scrolls the inner container horizontally
`this PR`:  nothing scrolls

I think master is not perfect here either. Since this code assumes precision scroll has both axes, and that we follow user intent (scroll vertically) -- the outer container should scroll vertically.

Susko3 (changes requested):
Since we have multiple regression here and the tests are not failing I suggest that `TestSceneScrollContainer.TestNestedScrolling` should be updated. Needs some refactoring to separate dragging and scrolling tests and to expand the covered test matrix:

- both Horizontal+Vertical / Vertical+Horizontal nested containers (as it does currently)
 ×
- scrolling in the `X`/`Y` direction
 ×
- `IsPrecise`: `true`/`false`
 ×
- <kbd>Shift</kbd>: held/not held (only relevant for non-precise, but tests should show that it doesn't affect precise)

osu.Framework/Graphics/Containers/ScrollContainer.cs
peppy: I mention this in the PR description as an intentional change. I don't know of a precision scroll device that doesn't have two axes of freedom. Let me know if that assumption is wrong.

Susko3: I agree with the assumption, but the current code doesn't work as I would expect. I would expect precision scrolling to behave the same as drag scrolling in the video below:

https://github.com/ppy/osu-framework/assets/16479013/395c63c8-9b99-4cb6-8d32-6d3569167028

This diff fixes it for me. It allows the outer scroll container to handle the scroll if the inner one didn't.

```diff
diff --git a/osu.Framework/Graphics/Containers/ScrollContainer.cs b/osu.Framework/Graphics/Containers/ScrollContainer.cs
index a9609b2d1..e4e9ee9fb 100644
--- a/osu.Framework/Graphics/Containers/ScrollContainer.cs
+++ b/osu.Framework/Graphics/Containers/ScrollContainer.cs
@@ -401,6 +401,9 @@ protected override bool OnScroll(ScrollEvent e)
                 ? e.ScrollDelta.X
                 : e.ScrollDelta.Y;
 
+            if (e.IsPrecise && scrollDeltaFloat == 0)
+                return false;
+
             // Let's assume precision scroll devices have both axes of scroll available.
             //
             // Then, for non-precision scroll devices, let's assuming the user's intention when scrolling
```

Susko3:
I've improved the tests in https://github.com/ppy/osu-framework/pull/5845. Any future changes to this PR should use and modify those tests. The PR intentionally doesn't change behaviour so it's non-controversial to merge and avoids merge conflicts with this PR.

PR #5846, opened 2023-06-17
Ensure renderer-api-driven frame wait is only invoked when a frame is actually drawn

peppy:
I noticed very long pauses when loading song select on my windows PC. Tracking this down, the game was getting stuck on the D3D11Swapchain [wait handle](https://github.com/ppy/veldrid/blob/d4340776eba0541aa3598856dafccf25633072f2/src/Veldrid/D3D11/D3D11Swapchain.cs#L303) until the full one second timeout.

It turns out that the wait handle will not be set unless a frame finishes rendering. The fail case is where the update thread takes over 100ms to prepare a frame, causing `TripleBuffer.GetForRead` to timeout and return `null`, which early exits the draw frame method.

This would happen anywhere the game takes slightly too long to render an update frame. I could reproduce:

- Entering song select
- Opening settings
- Opening beatmap listing

To confirm / reproduce this issue, remove the 1000ms timeout from `WaitOne` in veldrid and add a single `Thread.Sleep(150)` in any update thread code, and watch the game deadlock indefinitely.

For good measure, here's a visual example of this happening:

image

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -468,6 +468,8 @@ namespace osu.Framework.Platform
 
         private readonly DepthValue depthValue = new DepthValue();
 
+        private bool didRenderFrame;
+
         protected virtual void DrawFrame()
         {
             if (Root == null)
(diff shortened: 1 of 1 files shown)


bdach (approved):

frenzibyte:
On Metal, the display link callback is always signalled regardless of whether a frame was drawn, so this doesn't quite have an effect on renderer. Testing confirms that anyways.

frenzibyte (approved):

PR #5847, opened 2023-06-18
Improve metal frame rendering when more than one display is connected

### LINKED ISSUE #5828: CVDisplayLink set to wrong display
Currently, CVDisplayLink on macOS will not update to the correct target refresh rate of the display the game window is located in, it will also start with the wrong refresh rate in some scenarios (see https://github.com/ppy/veldrid/pull/20).

A call needs to be made to [CVDisplayLinkSetCurrentCGDisplay](https://developer.apple.com/documentation/corevideo/1456768-cvdisplaylinksetcurrentcgdisplay) to correctly set the target refresh rate.

Calling [`DisplayLinkUpdateActiveMonitor`](https://github.com/ppy/veldrid/blob/542b3f255923610b8fb4e114f3f832ee3cbe0cbd/src/Veldrid/MTL/MTLGraphicsDevice.cs#L60) from the linked Veldrid's `MTLGraphicsDevice` should fix these issues. For instance,
```diff
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -977,6 +977,13 @@ protected void SetupRendererAndWindow(IRenderer renderer, GraphicsSurfaceType su
 
             currentDisplayMode = Window.CurrentDisplayMode.GetBoundCopy();
             currentDisplayMode.BindValueChanged(_ => updateFrameSyncMode());
+            Window.CurrentDisplayBindable.BindValueChanged(evt =>
+            {
+                if (Renderer is not VeldridRenderer veldridRenderer) return;
+
+                var bounds = evt.NewValue.Bounds;
+                veldridRenderer.Device.DisplayLinkUpdateActiveMonitor(bounds.X, bounds.Y, bounds.Width, bounds.Height);
+            }, true);
 
             IsActive.BindTo(Window.IsActive);
         }

```

peppy:
Incorporating @ILW8's work (https://github.com/ppy/veldrid/pull/20) to more correctly provide hinting to metal as to where the window is running on the host, which allows better selection of refresh rates and display timings.

Tested on single display macbook and dual display mac pro (with two external displays). In these cases, I'm not noticing anything working better than before, but it runs fine without any signs of regression.

Closes #5828.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -44,6 +44,7 @@ using osu.Framework.Graphics.Video;
 using osu.Framework.IO.Serialization;
 using osu.Framework.IO.Stores;
 using osu.Framework.Localisation;
+using Rectangle = System.Drawing.Rectangle;
 using Size = System.Drawing.Size;
 
 namespace osu.Framework.Platform
@@ -989,6 +990,16 @@ namespace osu.Framework.Platform
 
             currentDisplayMode = Window.CurrentDisplayMode.GetBoundCopy();
             currentDisplayMode.BindValueChanged(_ => updateFrameSyncMode());
+
+            Window.CurrentDisplayBindable.BindValueChanged(display =>
+            {
+                if (Renderer is VeldridRenderer veldridRenderer)
+                {
+                    Rectangle bounds = display.NewValue.Bounds;
+
+                    veldridRenderer.Device.UpdateActiveDisplay(bounds.X, bounds.Y, bounds.Width, bounds.Height);
+                }
+            }, true);
 
             IsActive.BindTo(Window.IsActive);
         }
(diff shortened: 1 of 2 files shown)


frenzibyte (approved):
Looks solid. Timing of calling the function is no longer needed now with https://github.com/ppy/veldrid/pull/20/commits/7beb1270bf1c083097c1bd22ba948da66c1a394a, since if the game is running on multithreaded and the function happened to be called while the draw thread is in the middle of a frame, then the only effect would be blocking the input thread until the draw frame finishes processing (unlike previously where it would pretty much deadlock).

We should eventually have better `IRenderer` API for this one way or another, but there's only one usage of this right now so let's get it in for the time being.

Tested to work with my setup, the rate of the callback (indicated by the draw FPS) changes correctly based on the display.

PR #5853, opened 2023-06-20
Ensure triple buffer never uses the same buffer twice in a row

peppy:
As discussed in #osu-framework, this is probably a safer way of implementing the draw node triple buffer. It ensures that the last read buffer will never be written to. This may be helpful in scenarios where an external resource (ie. the GPU) is still interacting with the last read object's content.

This comes with no changes to lockability overheads, and only updates the naive approach previously used to find the next usable buffer.

### DIFF
--- a/osu.Framework.Tests/Graphics/TripleBufferTest.cs
+++ b/osu.Framework.Tests/Graphics/TripleBufferTest.cs
@@ -329,7 +329,7 @@ namespace osu.Framework.Tests.Graphics
 
             public override string ToString()
             {
-                return $"{base.ToString()} buffer index: {ID}";
+                return $"{base.ToString()} ID: {ID}";
             }
         }
     }
--- a/osu.Framework.Tests/Graphics/TripleBufferTest.cs
+++ b/osu.Framework.Tests/Graphics/TripleBufferTest.cs
@@ -1,6 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using NUnit.Framework;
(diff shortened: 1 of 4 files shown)


bdach:
On my machine, this change almost instantly trips the following assertion:

https://github.com/ppy/osu-framework/blob/7cc5e42b1a7b61a4e3f6b70ef5bcf2337c85e324/osu.Framework/Allocation/TripleBuffer.cs#L98

The access pattern involved appears to be:

Untitled Diagram drawio(1)

The double write to buf#2 seems to be the issue. If I'm understanding [the relevant discord conversation](https://discord.com/channels/188630481301012481/589331078574112768/1120601398841720832) correctly, this doesn't appear to be necessarily _wrong_, as under the constraints laid down there as I understand them (as in, not only the currently-read buffer is unsafe to write to, but the last-completed-read buffer as well). So maybe the assertion in question needs to go, but I'm not super sure what it was doing there in the first place.

bdach (changes requested):
as above

peppy:
Good catch, that's definitely an issue (if the final read in your example finished earlier the double write to the same buffer could mean it misses an opportunity to read the earlier write).

peppy:
I've added your specific test case as `TestSameBufferIsNotWrittenTwiceInRowContestation2`, in addition to one other edge case I found.

I've also updated the test scenes to be zero-indexes and pre-initialised to improve legibility.

frenzibyte (approved):

osu.Framework.Tests/Graphics/TripleBufferTest.cs
```diff
@@ -34,6 +35,205 @@ public void TestReadOnly()
...
+                // buffer 1: old
+                // buffer 2: old
+
+                using (var write = tripleBuffer.GetForWrite())
+                {
+                    Assert.That(write.Object?.ID, Is.EqualTo(1));
+
+                    // buffer 0: reading
+                    // buffer 1: writing
+                    // buffer 2: old
```
bdach, line 223: Why is buffer 1 being picked for the write here? From a "GPU may still be holding onto the second-last frame data" standpoint, which is a point I gathered from the github discussion, wouldn't buffer 2 be the better option to try first? Does the fact that there is an ongoing read of buffer 0 change circumstances?

bdach (commented):
Assertions aren't being tripped anymore, but I'm not sure of my capability to review this at this point. I've skimmed the discord conversation but I'm not sure I can tell what the end goal is or what the needed change is or what we want to achieve here after all.

Expectations in one of the tests seem to be opposite of what was discussed on discord but maybe I'm just not understanding correctly...?

osu.Framework.Tests/Graphics/TripleBufferTest.cs
frenzibyte: I'm not 100% sure on this but from what I remember in the discord discussion, `TripleBuffer` was initially not round-robin at all, and with this PR, it is round-robin on only two buffers, so it's possible for it to write to a buffer that was "last last" read from (i.e. `0 -> 1 -> 0 -> 1 -> ...`).

While it is more safe to utilise the oldest buffer in the list, it's deemed as unnecessary work unless current behaviour is shown to result in corruption (at least that's what I've understood from the discord discussion).

That being said, I feel we should clarify that properly in tests.

peppy: > Why is buffer 1 being picked for the write here?

The only (valid) guarantee from a triple buffer is that the last read is not overwritten. Take a moment to confirm your understanding of that. Why? Because the next write is going to clobber the buffer anyway.

Could we choose a better buffer for the first write to give a fraction of a second more chance of the last-last-read not being clobbered? Sure, but I'd argue that's pointless because you're dancing with the devil.

I'm not against changing this, but if it is changed, it should *not* be in the documentation as there is no use case you could use this change to guarantee any solid behaviour.

smoogipoo (approved):

PR #5856, opened 2023-06-20
Fix stray inspection showing up in rider

peppy:
Not sure how this snuck its way in. Non-EAP rider.

### DIFF
--- a/osu.Framework.SourceGeneration/Generators/HandleInput/HandleInputSourceEmitter.cs
+++ b/osu.Framework.SourceGeneration/Generators/HandleInput/HandleInputSourceEmitter.cs
@@ -55,22 +55,20 @@ namespace osu.Framework.SourceGeneration.Generators.HandleInput
             if (Target.RequestsNonPositionalInput || isDrawable)
                 yield return createInputMember("RequestsNonPositionalInput", Target.RequestsNonPositionalInput);
 
-            MemberDeclarationSyntax createInputMember(string name, bool value)
-            {
-                return SyntaxFactory.PropertyDeclaration(
-                                        SyntaxFactory.PredefinedType(SyntaxFactory.Token(SyntaxKind.BoolKeyword)),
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):
I guess...

Though I don't know why you're seeing this inspection here (I don't) and apparently here only? Because there's other places that have similar:
https://github.com/ppy/osu-framework/blob/147424e287a409a6667e413aad0a822fb4aae886/osu.Framework.SourceGeneration/Generators/Dependencies/Emitters/DependenciesFileEmitter.cs#L105-L125
https://github.com/ppy/osu-framework/blob/147424e287a409a6667e413aad0a822fb4aae886/osu.Framework.SourceGeneration/Generators/Dependencies/Emitters/DependenciesFileEmitter.cs#L67-L82
https://github.com/ppy/osu-framework/blob/147424e287a409a6667e413aad0a822fb4aae886/osu.Framework.SourceGeneration/Generators/Dependencies/Emitters/DependenciesFileEmitter.cs#L34-L65
https://github.com/ppy/osu-framework/blob/147424e287a409a6667e413aad0a822fb4aae886/osu.Framework.SourceGeneration/Generators/Dependencies/Emitters/BackgroundDependencyLoaderEmitter.cs#L44-L52
https://github.com/ppy/osu-framework/blob/147424e287a409a6667e413aad0a822fb4aae886/osu.Framework.SourceGeneration/Generators/Dependencies/Emitters/CachedMemberEmitter.cs#L37-L46
https://github.com/ppy/osu-framework/blob/147424e287a409a6667e413aad0a822fb4aae886/osu.Framework.SourceGeneration/Generators/Dependencies/Emitters/ResolvedMemberEmitter.cs#L40-L49

etc etc etc etc...

As far as I can tell these should be hints:
image

Having braces in these multi-line statements is generally a code style I've been using with more recent SGen implementations/work.

Issue #5861, opened 2023-06-21
Stencil isn't working within `BufferedContainer`s


EVAST9919:
How to reproduce: in the `StencilTestScene` make `StencilDrawable` a child of a `BufferedContainer`. Result will look like Stencil was not applied at all.
Tested with LegacyGL and D3D, so it probably just isn't handled by the `BufferedContainer` implementation.

smoogipoo:
It's not handled by default, but it should work if you give it a stencil format:

`new BufferedContainer(new[] { RenderBufferFormat.D32S8 })`

EVAST9919:
It *kinda* does. Main buffer works fine, but effect buffers become empty.
![Снимок экрана (377)](https://github.com/ppy/osu-framework/assets/22874522/1d9bce5f-bb7e-4344-b83e-beece70d5c64)
That's because effect buffers are being created with the same format as the main buffer.
![Снимок экрана 2023-06-22 172858](https://github.com/ppy/osu-framework/assets/22874522/a8d8d63e-b69c-4ab1-8f31-6cd6b9d2936f)
Making formats null for them fixes the issue.
![Снимок экрана (376)](https://github.com/ppy/osu-framework/assets/22874522/6a1b4acb-4a91-4c44-9aa4-38e0e19c44ab)
So what's the way forward? Adding a separate array to use for effect buffers or always leaving them as null?

smoogipoo:
Stencil should not be applied to the effect buffers. I don't think there's ever a valid use for that since you're drawing an entire quad.

If you'd want to, for example, blur and _then_ apply the grid stencil from that test scene to create crisp gutters, then the way of doing that should be to invert the process - blur first in a sub-container, and then stencil in the parenting container.

EVAST9919:
I'm not applying stencil to the effect buffers. What I'm saying is that with your proposal Main and Effect buffers are being created with stencil format, which is probably incorrect (because applying stencil format *only* to the main buffer fixes my issues). Instead I'm proposing to split format specification for main and effect buffers to have more control.

smoogipoo:
Is there ever a use for applying stencil to the effect buffers? Even separately from the main buffer? Because it's drawing an entire quad.
So my proposal is that effect buffers shouldn't get any stencil (_or depth_) formats at all.

EVAST9919:
Looks like I wasn't clear enough: I'm not applying stencil to the buffered container. It's the drawable within the buffered container which contains stencil isn't being rendered.

[...]

PR #5863, opened 2023-06-22
Add early guards against misuse of stencil testing

frenzibyte:
Coming from https://github.com/ppy/osu-framework/issues/5861#issue-1768470004, this PR adds guards against enabling stencil tests when targeting a framebuffer with no stencil attachment.

### DIFF
--- a/osu.Framework/Graphics/Rendering/Renderer.cs
+++ b/osu.Framework/Graphics/Rendering/Renderer.cs
@@ -751,6 +751,9 @@ namespace osu.Framework.Graphics.Rendering
             if (CurrentStencilInfo.Equals(stencilInfo))
                 return;
 
+            if (stencilInfo.StencilTest && FrameBuffer != null && (FrameBuffer.Formats == null || !FrameBuffer.Formats[0].HasStencilAttachment()))
+                throw new InvalidOperationException("Cannot enable stencil tests on a frame buffer with no stencil attachment.");
+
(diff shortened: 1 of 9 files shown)


osu.Framework/Graphics/Rendering/Renderer.cs
```diff
@@ -751,6 +751,9 @@ private void setStencilInfo(StencilInfo stencilInfo)
             if (CurrentStencilInfo.Equals(stencilInfo))
                 return;
 
+            if (stencilInfo.StencilTest && FrameBuffer != null && (FrameBuffer.Formats == null || !FrameBuffer.Formats[0].HasStencilAttachment()))
```
smoogipoo, line 754: Why `Formats[0]` and not `Formats.Any()`?

Or... I think the enumerator's going to be an issue here since it's an `IList`, so probably separate it out into an extension that loops through all the formats.

smoogipoo (changes requested):

smoogipoo:
Have you actually tested this to work? I'm curious how it avoids the ordering case, if at all...

For example, you currently have an FBO bound that does _not_ have a stencil RBO, how do you now bind an FBO _with_ a stencil RBO? From reading the code, it looks like the only way you can do that is if you bind the backbuffer (`FrameBuffer = null`) and only then bind the new RBO, otherwise either one of those methods are going to trigger the exception.

Issue #5858, opened 2023-06-21
Stencil isn't working with D3D renderer


EVAST9919:
| OpenGL/LegacyGL | D3D |
|--------|-----|
| ![Снимок экрана (373)](https://github.com/ppy/osu-framework/assets/22874522/16883732-d8e0-4307-96d0-867cb097cec0) | ![Снимок экрана (371)](https://github.com/ppy/osu-framework/assets/22874522/ee1fe235-218a-4ccf-8eec-3ad647917f93) |

PR #5866, opened 2023-06-23
Define stencil attachment for swap-chain frame buffers in Veldrid

frenzibyte:
- [ ] Depends on #5863 (for the `ToPixelFormat` extension)
- One other and final step towards resolving https://github.com/ppy/osu/issues/23970
- May result in resolving https://github.com/ppy/osu-framework/issues/5858 (it doesn't on Metal)


smoogipoo:
I'm not sure about this. From my experience, 32-bit depth isn't super well supported.

https://feedback.wildfiregames.com/gl/constants/

image

https://gamedev.stackexchange.com/a/120648

PR #5867, opened 2023-06-23
Apply NRT to all files where it can be easily applied

peppy:
Something I've had sitting around. Tested against osu! (see https://github.com/ppy/osu/pull/24019).

Another pass will follow, but feel that getting this one out of the way is a good first step.

- 75ed421 is an automated pass. reviewing it separately is recommended as it should be a breeze
- 3064650 are manual fixes after the automated pass to fix non-compile-time warnings (aka r#)
- d002fda is stuff rider didn't pick up on (CI script unfortunately crashes locally for me)
- BOM in files is inconsistent in our repository. this arbitrarily changes BOM. please don't focus on this. we should normalise this at a later point.

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkTextBuilder.cs
+++ b/osu.Framework.Benchmarks/BenchmarkTextBuilder.cs
@@ -51,7 +51,7 @@ namespace osu.Framework.Benchmarks
                 new CharacterGlyph(character, character, character, character, character, null),
                 new DummyRenderer().CreateTexture(1, 1));
 
-            public Task<ITexturedCharacterGlyph> GetAsync(string fontName, char character) => Task.Run(() => Get(fontName, character));
+            public Task<ITexturedCharacterGlyph?> GetAsync(string fontName, char character) => Task.Run<ITexturedCharacterGlyph?>(() => Get(fontName, character));
         }
     }
 }
--- a/osu.Framework/IO/Stores/GlyphStore.cs
+++ b/osu.Framework/IO/Stores/GlyphStore.cs
@@ -1,4 +1,4 @@
-// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
 #nullable disable
@@ -118,7 +118,6 @@ namespace osu.Framework.IO.Stores
             return $@"{AssetName}_{page.ToString().PadLeft((Font.Pages.Count - 1).ToString().Length, '0')}.png";
         }
 
-        [CanBeNull]
         public CharacterGlyph Get(char character)
         {
             if (Font == null)
--- a/osu.Framework/Input/Events/FocusLostEvent.cs
+++ b/osu.Framework/Input/Events/FocusLostEvent.cs
@@ -16,7 +16,7 @@ namespace osu.Framework.Input.Events
         /// </summary>
         public readonly Drawable? NextFocused;
 
-        public FocusLostEvent(InputState state, Drawable nextFocused)
+        public FocusLostEvent(InputState state, Drawable? nextFocused)
             : base(state)
         {
             NextFocused = nextFocused;
--- a/osu.Framework/Input/Events/FocusLostEvent.cs
+++ b/osu.Framework/Input/Events/FocusLostEvent.cs
@@ -1,7 +1,5 @@
-// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
-
-#nullable disable
 
 using JetBrains.Annotations;
 using osu.Framework.Graphics;
--- a/osu.Framework/Input/Events/FocusLostEvent.cs
+++ b/osu.Framework/Input/Events/FocusLostEvent.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using JetBrains.Annotations;
 using osu.Framework.Graphics;
 using osu.Framework.Input.States;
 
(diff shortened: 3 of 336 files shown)


osu.Framework/IO/Stores/GlyphStore.cs
```diff
@@ -118,7 +118,6 @@ protected string GetFilenameForPage(int page)
             return $@"{AssetName}_{page.ToString().PadLeft((Font.Pages.Count - 1).ToString().Length, '0')}.png";
         }
 
-        [CanBeNull]
```
bdach, line 121: What's happened here? `#nullable disable` is not removed from the file, so `CharacterGlyph?` return type cannot be applied - so why is this removed?

osu.Framework/Input/Events/FocusLostEvent.cs
```diff
@@ -17,8 +14,7 @@ public class FocusLostEvent : UIEvent
         /// <summary>
         /// The <see cref="Drawable"/> that will gain focus, or <c>null</c> if nothing will gain focus.
         /// </summary>
-        [CanBeNull]
-        public readonly Drawable NextFocused;
+        public readonly Drawable? NextFocused;
```
bdach, line 17: Bit weird that this is nullable, but the ctor param isn't. `FocusEvent` has the ctor param nullable, too.

osu.Framework/Testing/DrawFrameRecordingContainer.cs
```diff
@@ -33,7 +30,7 @@ public DrawFrameRecordingContainer()
         }
 
         [BackgroundDependencyLoader(true)]
```
bdach, line 32:
```suggestion
        [BackgroundDependencyLoader]
```

osu.Framework.Benchmarks/BenchmarkTextBuilder.cs
```diff
@@ -51,7 +51,7 @@ private class TestStore : ITexturedGlyphLookupStore
                 new CharacterGlyph(character, character, character, character, character, null),
                 new DummyRenderer().CreateTexture(1, 1));
 
-            public Task<ITexturedCharacterGlyph> GetAsync(string fontName, char character) => Task.Run(() => Get(fontName, character));
+            public Task<ITexturedCharacterGlyph?> GetAsync(string fontName, char character) => Task.Run<ITexturedCharacterGlyph?>(() => Get(fontName, character));
```
bdach, line 54: Nitpick but I'd probably do

```diff
diff --git a/osu.Framework.Benchmarks/BenchmarkTextBuilder.cs b/osu.Framework.Benchmarks/BenchmarkTextBuilder.cs
index fb774e296..0c83350a2 100644
--- a/osu.Framework.Benchmarks/BenchmarkTextBuilder.cs
+++ b/osu.Framework.Benchmarks/BenchmarkTextBuilder.cs
@@ -47,11 +47,11 @@ private void initialiseBuilder(bool withDifferentBaselines)
 
         private class TestStore : ITexturedGlyphLookupStore
         {
-            public ITexturedCharacterGlyph Get(string fontName, char character) => new TexturedCharacterGlyph(
+            public ITexturedCharacterGlyph? Get(string fontName, char character) => new TexturedCharacterGlyph(
                 new CharacterGlyph(character, character, character, character, character, null),
                 new DummyRenderer().CreateTexture(1, 1));
 
-            public Task<ITexturedCharacterGlyph?> GetAsync(string fontName, char character) => Task.Run<ITexturedCharacterGlyph?>(() => Get(fontName, character));
+            public Task<ITexturedCharacterGlyph?> GetAsync(string fontName, char character) => Task.Run(() => Get(fontName, character));
         }
     }
 }

```

to match the nullability annotations on the interface instead?

osu.Framework/Testing/TestingExtensions.cs
```diff
@@ -14,7 +12,7 @@ public static class TestingExtensions
         /// <summary>
         /// Find all children recursively of a specific type. As this is expensive and dangerous, it should only be used for testing purposes.
         /// </summary>
-        public static IEnumerable<T> ChildrenOfType<T>(this Drawable drawable)
+        public static IEnumerable<T> ChildrenOfType<T>(this Drawable? drawable)
```
bdach, line 15: What's with this one? Is anyone really calling `ChildrenOfType()` on a null reference?

bdach:
Also there are some new [build](https://github.com/ppy/osu-framework/pull/5867/files#annotation_12197382190) [warnings](https://github.com/ppy/osu-framework/pull/5867/files#annotation_12197382192) in the iOS project.

peppy:
> Also there are some new [build](https://github.com/ppy/osu-framework/pull/5867/files#annotation_12197382190) [warnings](https://github.com/ppy/osu-framework/pull/5867/files#annotation_12197382192) in the iOS project.

Ah right. Mobile platforms were a TODO on these.

osu.Framework.Benchmarks/BenchmarkTextBuilder.cs
peppy: This gives a new warning, give it a try in code and see if you can come up with a better solution.

osu.Framework/Testing/TestingExtensions.cs
peppy: Might be okay. But also may have been a weird osu! test usage (I likely would not have made this change without a good reason). Will need to run tests against this before applying.

osu.Framework/Input/Events/FocusLostEvent.cs
peppy: Fixed this, good catch.

osu.Framework.Benchmarks/BenchmarkTextBuilder.cs
bdach: I even tried applying but then missed the incredibly subtle hinting that that's a warning >_>

osu.Framework/Testing/TestingExtensions.cs
bdach: If you don't recall from memory then I'm not sure I care that much for you to go check. I was mostly curious is all.

peppy: Turns out it's to ease cases like this:

2023-06-24 09 43 52@2x

There's around 30 usages osu! side which would need updating if we revert this. I think making the method nullable makes it a touch more flexible though, given its intended purpose.

bdach (approved):
Seems good. Going to expedite merge now to minimise chances of merge conflictery here. If it turns out something goes bad we can revert in a separate pull.

PR #5870, opened 2023-06-25
Add support for alternative clipboard keys (ctrl-insert / shift-delete / shift-insert)

peppy:
Generally considered windows first, but also seems to be supported on linux so let's just make them global.

From dicussion at https://github.com/ppy/osu/discussions/24017.


peppy:
FWIW this needs testing. On my macbook I'm having difficulty testing due to missing keys on keyboard.

bdach (approved):
Tested to work on windows and linux.

PR #5871, opened 2023-06-25
Add option to limit at 1x refresh rate

DuckI1ng:
Related discussions:
https://github.com/ppy/osu/discussions/18745
https://github.com/ppy/osu/discussions/23161

### DIFF
--- a/osu.Framework/Configuration/FrameSync.cs
+++ b/osu.Framework/Configuration/FrameSync.cs
@@ -9,6 +9,9 @@ namespace osu.Framework.Configuration
     {
         VSync,
 
+        [Description("1x refresh rate")]
+        Limit1x,
+
         [Description("2x refresh rate")]
         Limit2x,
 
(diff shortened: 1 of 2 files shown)


jai-x:
1x refresh rate is the same as VSync.

What are you trying to achieve?

DuckI1ng:
> 1x refresh rate is the same as VSync.
> 
> What are you trying to achieve?

It is not, using VSync comes with a latency penalty.

BlauFx:
> > 1x refresh rate is the same as VSync.
> > What are you trying to achieve?
> 
> It is not, using VSync comes with a latency penalty.

In which way is there a latency penalty compared to 1x refresh rate?

peppy:
Thanks for your contribution.

Sorry but this is not the correct forward direction. Frame limiters need a re-think and that will happen in the coming days, but adding this right now is very confusing to a user.

PR #5873, opened 2023-06-26
Update deployment workflow to automatically run on new tag

peppy:
Logic taken from https://github.com/ppy/osu-web/blob/master/.github/workflows/pack.yml.

Haven't tested, will do next release.

### DIFF
--- a/.github/workflows/deploy-pack.yml
+++ b/.github/workflows/deploy-pack.yml
@@ -1,43 +1,47 @@
-# There is no manual way to call this out to run this on tags via UI.
-# See: https://github.community/t/workflow-dispatch-from-a-tag-in-actions-tab/130561
-on: workflow_dispatch
-name: Deploy - Framework
+name: Pack and nuget
+
+on:
+  push:
+    tags:
+      - '*'
 
 jobs:
-  check-if-tag:
-    name: Set Package Version
+  notify_pending_production_deploy:
     runs-on: ubuntu-latest
-    outputs:
-      version: ${{steps.deployment.outputs.version}}
     steps:
-      - name: Checkout
+      -
+        name: Submit pending deployment notification
         run: |
-          REPOSITORY="https://${{ github.actor }}:${{ github.token }}@github.com/${{ github.repository }}.git"
-          BRANCH="${GITHUB_REF/#refs\/heads\//}"
+          export TITLE="Pending osu-framework Production Deployment: $GITHUB_REF_NAME"
+          export URL="https://github.com/ppy/osu-framework/actions/runs/$GITHUB_RUN_ID"
+          export DESCRIPTION="Docker image was built for tag $GITHUB_REF_NAME and awaiting approval for production deployment:
+          [View Workflow Run]($URL)"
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-pack.yml
```diff
@@ -1,51 +1,55 @@
...
         run: |
-          if [ $(git describe --exact-match --tags HEAD &> /dev/null; echo $?) == 0 ]; then
-            echo "::set-output name=VERSION::$(git describe --exact-match --tags HEAD)"
-          else
-            echo "fatal: no tag detected for HEAD. Workflow will now stop."
-            exit 128;
-          fi
+          export TITLE="Pending osu-framework Production Deployment: $GITHUB_REF_NAME"
+          export URL="https://github.com/ppy/osu-framework/actions/runs/$GITHUB_RUN_ID"
+          export DESCRIPTION="Docker image was built for tag $GITHUB_REF_NAME and awaiting approval for production deployment:
```
bdach, line 17:
```suggestion
          export DESCRIPTION="Awaiting approval for building NuGet packages for tag $GITHUB_REF_NAME:
```

I'm going to assume that the "approval" part is achieved thanks to the `environment: production` definitions on the remaining jobs, so that seems to check out. But if that's how that works, that would mean that nothing has been built yet at the point of submitting a message asking for approval, as per suggestion above.

.github/workflows/deploy-pack.yml
```diff
@@ -1,51 +1,55 @@
...
+    tags:
+      - '*'
 
 jobs:
-  check-if-tag:
-    name: Set Package Version
+  notify_pending_production_deploy:
     runs-on: ubuntu-latest
-    outputs:
-      version: ${{steps.deployment.outputs.version}}
```
bdach, line 11: This output is depended on in several places of the workflow definition:

https://github.com/ppy/osu-framework/blob/a085962853a611c50b59c4d03dc43558698a55b2/.github/workflows/deploy-pack.yml#L62
https://github.com/ppy/osu-framework/blob/a085962853a611c50b59c4d03dc43558698a55b2/.github/workflows/deploy-pack.yml#L93
https://github.com/ppy/osu-framework/blob/a085962853a611c50b59c4d03dc43558698a55b2/.github/workflows/deploy-pack.yml#L96
https://github.com/ppy/osu-framework/blob/a085962853a611c50b59c4d03dc43558698a55b2/.github/workflows/deploy-pack.yml#L131
https://github.com/ppy/osu-framework/blob/a085962853a611c50b59c4d03dc43558698a55b2/.github/workflows/deploy-pack.yml#L168

They all need to be replaced to use [`{{ github.ref_name }}`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) instead I guess.

bdach (changes requested):

.github/workflows/deploy-pack.yml
peppy: That's correct, yep.

.github/workflows/deploy-pack.yml
peppy: Good catch.

bdach (approved):
Works in my head

PR #5875, opened 2023-06-26
Caution.md

stevenkiethanete:
all sprites were modified not changing the osu!ui.dll's code


bdach:
what in the world is this

PR #5876, opened 2023-06-28
Update veldrid

peppy:
Pulls in:

- https://github.com/ppy/veldrid/pull/25


bdach (approved):
Tested to work on a Win8.1 VM (d3d11 initialises with this now) and on Win11.

PR #5877, opened 2023-06-30
Set PerMonitor V2 DPI awareness on Windows

Susko3:
osu! will use PerMonitor V2 in https://github.com/ppy/osu/pull/24092, so this PR changes framework to match.

The catch clauses could probably just catch [`EntryPointNotFoundException`](https://learn.microsoft.com/en-us/dotnet/api/system.entrypointnotfoundexception?view=net-7.0), but catching everything seems safer.

### DIFF
--- a/osu.Framework/Platform/Windows/WindowsWindow.cs
+++ b/osu.Framework/Platform/Windows/WindowsWindow.cs
@@ -49,14 +49,31 @@ namespace osu.Framework.Platform.Windows
                     break;
             }
 
+            if (!declareDpiAwareV2())
+                declareDpiAware();
+        }
+
+        private bool declareDpiAwareV2()
+        {
             try
             {
-                // SDL doesn't handle DPI correctly on windows, but this brings things mostly in-line with expectations. (https://bugzilla.libsdl.org/show_bug.cgi?id=3281)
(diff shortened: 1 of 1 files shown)


bdach:
Is it an option to just change framework and get rid of the manifest changes? It feels kinda weird to me that we're duplicating this stuff.

Susko3:
Yep, it works fine with framework only. I'll update the osu!-side PR to remove the dpi manifest entries.

bdach:
So upon [reading up on relevant docs](https://learn.microsoft.com/en-us/windows/win32/hidpi/setting-the-default-dpi-awareness-for-a-process) it appears Microsoft strongly advocates using the manifest over the API calls. Which is unfortunate given my previous question above. I thought doing the API calls was going to be better for us because it meant that all framework consumers don't have to duplicate manifests anymore, and maybe it's still fine until proven otherwise? The reasons for why they want users to use manifest over winapi are at best vague - something about it not working after the `HWND` has been created, but as far as I'm concerned, we only ever create one `HWND` and it happens after this API call, so maybe fine?

I can't however materially test the change itself. I can only test that it doesn't die on Windows 10 and 8.1 (via VM) and it does appear to not die, so that's good. Not sure what to do with this change now.

peppy:
Probably fine to go-with-it-until-it-breaks.

peppy (approved):

PR #5878, opened 2023-07-01
Fix tablet output area is not updated when switching fullscreen and borderless mode on MacOS with metal renderer.

AkiSakurai:
Log of fetchWindowSize
```
Window Size:{Width=1792, Height=1067} Drawable Size: {Width=3586, Height=2134}
Window Size:{Width=1792, Height=1067} Drawable Size: {Width=3584, Height=2134}
Window Size:{Width=1304, Height=537} Drawable Size: {Width=3584, Height=2134}
Window Size:{Width=1304, Height=537} Drawable Size: {Width=2608, Height=1074}
Window Size:{Width=1304, Height=537} Drawable Size: {Width=2608, Height=1074}
****Window Size:{Width=1792, Height=1120} Drawable Size: {Width=2608, Height=1074}
Window Size:{Width=1792, Height=1120} Drawable Size: {Width=3584, Height=2240}
Window Size:{Width=1792, Height=1120} Drawable Size: {Width=3584, Height=2240}
Window Size:{Width=1792, Height=1120} Drawable Size: {Width=3584, Height=2240}
```

The drawable size is updated at the beginning of the frame, so the client size and window size might not be synchronised.

https://github.com/ppy/osu-framework/blob/19cbb8f4a681b4e8cb7838127ddcf91b71870cea/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs#L206-L212

The tablet driver takes the wrong client size, as the event is only triggered when the window size changes.

https://github.com/ppy/osu-framework/blob/19cbb8f4a681b4e8cb7838127ddcf91b71870cea/osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs#L123-L144

### DIFF
--- a/osu.Framework/Platform/SDL2Window_Windowing.cs
+++ b/osu.Framework/Platform/SDL2Window_Windowing.cs
@@ -439,8 +439,14 @@ namespace osu.Framework.Platform
             if (w <= 0 || drawableW <= 0)
                 return;
 
+            Size lastClientSize = ClientSize;
             Scale = (float)drawableW / w;
-            Size = new Size(w, h);
+            size = new Size(w, h);
+
+            if (lastClientSize != ClientSize)
+            {
+                Resized?.Invoke();
+            }
 
             // This function may be invoked before the SDL internal states are all changed. (as documented here: https://wiki.libsdl.org/SDL_SetEventFilter)
             // Scheduling the store to config until after the event poll has run will ensure the window is in the correct state.


osu.Framework/Platform/SDL2Window_Windowing.cs
```diff
@@ -439,8 +439,14 @@ private void fetchWindowSize()
...
 
+            Size lastClientSize = ClientSize;
             Scale = (float)drawableW / w;
-            Size = new Size(w, h);
+            size = new Size(w, h);
+
+            if (lastClientSize != ClientSize)
+            {
+                Resized?.Invoke();
+            }
```
bdach, line 449: I don't understand. Previously this was setting `Size`. And notably:

https://github.com/ppy/osu-framework/blob/28a72ebb442015ff6bbe1186695b9ce27f68759f/osu.Framework/Platform/SDL2Window_Windowing.cs#L185-L195

So this looks like it's _manually reimplementing_ what setting `Size` _already was doing_.

Not to mention that what this is checking is `ClientSize` which _is not the same value_ as it is the value post-scaling:

https://github.com/ppy/osu-framework/blob/28a72ebb442015ff6bbe1186695b9ce27f68759f/osu.Framework/Platform/SDL2Window_Windowing.cs#L273

Can you explain? Because as is this change reads completely wrong. There _may be_ something to it (maybe scale is changing when going fullscreen? whatever else?) but right now I cannot make any sense of this.

Susko3:
Could you open an issue showing the bug you're seeing. Preferably with a screen recording and full logs attached. And also test (and show your logs) on non-metal rendered. If you are modifying the code to get additional logging, please post the `git diff` of that.

Also, you say you're switching between fullscreen and borderless, but it seems you're also switching displays:

image

AkiSakurai:
moved to issue #5879.

PR #5880, opened 2023-07-01
Fix `IWindow.Resized` being invoked with invalid data during `WindowMode` switch

### LINKED ISSUE #5879: The tablet output area is not properly configured when switching from fullscreen to borderless mode on MacOS with the metal renderer.
This issue is not observed when using the OpenGL renderer. I have not tested it on other operating systems.

The video below demonstrates the problem encountered when switching from borderless to fullscreen and back to borderless. 

OpenGL Renderer

https://github.com/ppy/osu-framework/assets/75532970/1af00e6e-d1f5-4815-80e0-ffe79e99f591

```
UpdateWindowStateAndSize Fullscreen
Invoke Resized Size: {Width=1920, Height=1080} ClientSize: {Width=1367, Height=480}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
Invoke Resized Size: {Width=1920, Height=1027} ClientSize: {Width=3840, Height=2054}
FetchWindowSize Size: {Width=1920, Height=1027} ClientSize: {Width=3840, Height=2054}
Invoke Resized Size: {Width=1921, Height=1027} ClientSize: {Width=1921, Height=1027}
FetchWindowSize Size: {Width=1921, Height=1027} ClientSize: {Width=1921, Height=1027}
FetchWindowSize Size: {Width=1921, Height=1027} ClientSize: {Width=1921, Height=1027}
Invoke Resized Size: {Width=1920, Height=1027} ClientSize: {Width=1920, Height=1027}
FetchWindowSize Size: {Width=1920, Height=1027} ClientSize: {Width=1920, Height=1027}
FetchWindowSize Size: {Width=1920, Height=1027} ClientSize: {Width=1920, Height=1027}
Invoke Resized Size: {Width=1367, Height=480} ClientSize: {Width=1367, Height=480}
FetchWindowSize Size: {Width=1367, Height=480} ClientSize: {Width=1367, Height=480}
FetchWindowSize Size: {Width=1367, Height=480} ClientSize: {Width=1367, Height=480}
FetchWindowSize Size: {Width=1367, Height=480} ClientSize: {Width=1367, Height=480}
UpdateWindowStateAndSize FullscreenBorderless
Invoke Resized Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
```
Metal renderer

https://github.com/ppy/osu-framework/assets/75532970/1a8ac233-5999-4826-af85-c82db14278ca

```
UpdateWindowStateAndSize Fullscreen
Invoke Resized Size: {Width=1920, Height=1080} ClientSize: {Width=1367, Height=480}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1367, Height=480}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
Invoke Resized Size: {Width=1920, Height=1027} ClientSize: {Width=3840, Height=2054}
FetchWindowSize Size: {Width=1920, Height=1027} ClientSize: {Width=3840, Height=2054}
Invoke Resized Size: {Width=1921, Height=1027} ClientSize: {Width=3840, Height=2054}
FetchWindowSize Size: {Width=1921, Height=1027} ClientSize: {Width=3840, Height=2054}
FetchWindowSize Size: {Width=1921, Height=1027} ClientSize: {Width=1921, Height=1027}
Invoke Resized Size: {Width=1920, Height=1027} ClientSize: {Width=1921, Height=1027}
FetchWindowSize Size: {Width=1920, Height=1027} ClientSize: {Width=1921, Height=1027}
FetchWindowSize Size: {Width=1920, Height=1027} ClientSize: {Width=1920, Height=1027}
Invoke Resized Size: {Width=1367, Height=480} ClientSize: {Width=1920, Height=1027}
FetchWindowSize Size: {Width=1367, Height=480} ClientSize: {Width=1920, Height=1027}
FetchWindowSize Size: {Width=1367, Height=480} ClientSize: {Width=1367, Height=480}
FetchWindowSize Size: {Width=1367, Height=480} ClientSize: {Width=1367, Height=480}
UpdateWindowStateAndSize FullscreenBorderless
Invoke Resized Size: {Width=1920, Height=1080} ClientSize: {Width=1367, Height=480}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1367, Height=480}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
FetchWindowSize Size: {Width=1920, Height=1080} ClientSize: {Width=1920, Height=1080}
```

The versions of osu! and osu!framework used were from the GitHub repositories located at https://github.com/ppy/osu/tree/2023.621.0 and https://github.com/ppy/osu-framework/tree/2023.620.0, respectively.
These modifications were used to generate the log.

``` diff
diff --git a/osu.Framework/Platform/SDL2Window_Windowing.cs b/osu.Framework/Platform/SDL2Window_Windowing.cs
index debf407df..1101b59a2 100644
--- a/osu.Framework/Platform/SDL2Window_Windowing.cs
+++ b/osu.Framework/Platform/SDL2Window_Windowing.cs
@@ -190,6 +190,8 @@ protected set
                 if (value.Equals(size)) return;
 
                 size = value;
+                Console.WriteLine($"Invoke Resized Size: {size} ClientSize: {graphicsSurface.GetDrawableSize()}");
+
                 Resized?.Invoke();
             }
         }
@@ -442,6 +444,8 @@ private void fetchWindowSize()
             Scale = (float)drawableW / w;
             Size = new Size(w, h);
 
+            Console.WriteLine($"FetchWindowSize Size: {Size} ClientSize: {graphicsSurface.GetDrawableSize()}");
+
             // This function may be invoked before the SDL internal states are all changed. (as documented here: https://wiki.libsdl.org/SDL_SetEventFilter)
             // Scheduling the store to config until after the event poll has run will ensure the window is in the correct state.
             EventScheduler.AddOnce(storeWindowSizeToConfig);
@@ -574,6 +578,9 @@ private void updateAndFetchWindowSpecifics()
         /// </summary>
         protected virtual void UpdateWindowStateAndSize(WindowState state, Display display, DisplayMode displayMode)
         {
+            
+            Console.WriteLine($"UpdateWindowStateAndSize {state}");
+

```
[runtime.log](https://github.com/ppy/osu-framework/files/11925623/runtime.log)

Susko3:
- Resolves #5879 

As seen in that issue, `Resized` would get invoked when the window was invalid / in the process of changing window modes. This is because we have a multi-step window mode change process _and_ we're fetching the window size with an event filter (SDL would usually coalesce those events to a single event with the last set, valid size).

Callstack from one `fetchWindowSize()` that would fetch the invalid size. Notice how `UpdateWindowStateAndSize()` is in progress.

image

### DIFF
--- a/osu.Framework.Tests/Visual/Platform/TestSceneWindowModeResizedEvent.cs
+++ b/osu.Framework.Tests/Visual/Platform/TestSceneWindowModeResizedEvent.cs
@@ -1,0 +1,90 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Collections.Generic;
+using System.Drawing;
+using NUnit.Framework;
+using osu.Framework.Allocation;
+using osu.Framework.Bindables;
+using osu.Framework.Configuration;
+using osu.Framework.Extensions.ObjectExtensions;
+using osu.Framework.Graphics;
+using osu.Framework.Platform;
+
+namespace osu.Framework.Tests.Visual.Platform
+{
+    [Ignore("This test cannot run in headless mode (a window instance is required).")]
+    [System.ComponentModel.Description($"Checks that {nameof(IWindow.Resized)} behaves correctly when {nameof(IWindow.WindowMode)} is changed.")]
+    public partial class TestSceneWindowModeResizedEvent : FrameworkTestScene
+    {
+        private static readonly Size windowed_size = new Size(1280, 720);
+
+        private readonly Bindable<WindowMode> windowMode = new Bindable<WindowMode>();
+        private IWindow window = null!;
+
+        private readonly Queue<Size> resizeInvokes = new Queue<Size>();
+
+        [Resolved]
+        private FrameworkConfigManager config { get; set; } = null!;
+
+        public TestSceneWindowModeResizedEvent()
+        {
+            Child = new WindowDisplaysPreview
+            {
+                RelativeSizeAxes = Axes.Both
+            };
+
+            AddLabel("set up defaults");
+            AddStep($"set windowed size to {windowed_size}", () => config.SetValue(FrameworkSetting.WindowedSize, windowed_size));
+            AddStep("set default fullscreen size", () => config.GetBindable<Size>(FrameworkSetting.SizeFullscreen).SetDefault());
+        }
+
+        [BackgroundDependencyLoader]
+        private void load(GameHost host)
+        {
+            window = host.Window!;
+            window.Resized += windowResized;
+            config.BindWith(FrameworkSetting.WindowMode, windowMode);
+        }
+
+        private void windowResized() => resizeInvokes.Enqueue(window.ClientSize);
+
+        private void setUp(WindowMode startingMode, WindowMode finalMode)
+        {
+            AddStep($"set mode to {startingMode}", () => windowMode.Value = startingMode);
+            AddStep("clear resize queue", () => resizeInvokes.Clear());
+            AddStep($"set mode to {finalMode}", () => windowMode.Value = finalMode);
+        }
+
+        [TestCase(WindowMode.Borderless, WindowMode.Fullscreen)]
+        [TestCase(WindowMode.Fullscreen, WindowMode.Borderless)]
+        public void TestFullscreenAndBorderless(WindowMode startingMode, WindowMode finalMode)
+        {
+            setUp(startingMode, finalMode);
+            // since the two window modes take up the entire display, the size of the window shouldn't change.
+            AddAssert("resize queue is empty", () => resizeInvokes, () => Is.Empty);
+        }
+
+        [TestCase(WindowMode.Windowed, WindowMode.Borderless)]
+        [TestCase(WindowMode.Windowed, WindowMode.Fullscreen)]
+        [TestCase(WindowMode.Fullscreen, WindowMode.Windowed)]
+        [TestCase(WindowMode.Borderless, WindowMode.Windowed)]
+        public void TestModeSwitchWindowed([Values] WindowMode startingMode, [Values] WindowMode finalMode)
+        {
+            setUp(startingMode, finalMode);
+            AddAssert("only one resize event", () => resizeInvokes, () => Has.Count.EqualTo(1));
+            if (finalMode == WindowMode.Windowed)
+                AddAssert("resized to windowed size", () => resizeInvokes.Dequeue(), () => Is.EqualTo(windowed_size));
+            else
+                AddAssert("resized to display size", () => resizeInvokes.Dequeue(), () => Is.EqualTo(window.CurrentDisplayBindable.Value.Bounds.Size));
+        }
+
+        protected override void Dispose(bool isDisposing)
+        {
+            if (window.IsNotNull())
+                window.Resized -= windowResized;
+
+            base.Dispose(isDisposing);
+        }
+    }
+}
--- a/osu.Framework.Tests/Visual/Platform/TestSceneWindowModeResizedEvent.cs
+++ b/osu.Framework.Tests/Visual/Platform/TestSceneWindowModeResizedEvent.cs
@@ -10,6 +10,7 @@ using osu.Framework.Configuration;
 using osu.Framework.Extensions.ObjectExtensions;
 using osu.Framework.Graphics;
 using osu.Framework.Platform;
+using osu.Framework.Testing;
 
 namespace osu.Framework.Tests.Visual.Platform
 {
@@ -33,10 +34,6 @@ namespace osu.Framework.Tests.Visual.Platform
             {
                 RelativeSizeAxes = Axes.Both
             };
-
-            AddLabel("set up defaults");
-            AddStep($"set windowed size to {windowed_size}", () => config.SetValue(FrameworkSetting.WindowedSize, windowed_size));
-            AddStep("set default fullscreen size", () => config.GetBindable<Size>(FrameworkSetting.SizeFullscreen).SetDefault());
         }
 
         [BackgroundDependencyLoader]
@@ -48,6 +45,13 @@ namespace osu.Framework.Tests.Visual.Platform
         }
 
         private void windowResized() => resizeInvokes.Enqueue(window.ClientSize);
+
+        [SetUpSteps]
+        public void SetUpSteps()
+        {
+            AddStep($"set windowed size to {windowed_size}", () => config.SetValue(FrameworkSetting.WindowedSize, windowed_size));
+            AddStep("set default fullscreen size", () => config.GetBindable<Size>(FrameworkSetting.SizeFullscreen).SetDefault());
+        }
 
         private void setUp(WindowMode startingMode, WindowMode finalMode)
         {
--- a/osu.Framework/Platform/SDL2Window_Windowing.cs
+++ b/osu.Framework/Platform/SDL2Window_Windowing.cs
@@ -534,7 +534,9 @@ namespace osu.Framework.Platform
                 windowState = pendingWindowState.Value;
                 pendingWindowState = null;
 
+                updatingWindowStateAndSize = true;
                 UpdateWindowStateAndSize(windowState, currentDisplay, currentDisplayMode.Value);
+                updatingWindowStateAndSize = false;
 
                 fetchWindowSize();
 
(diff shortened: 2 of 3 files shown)


AkiSakurai:
Confirmed it fixes #5879.

Just find that there is a close related problem where, when moving a window from an HDPI display to an LDPI display, the resize event is not triggered because the window size does not change
```
[runtime] 2023-07-01 15:07:33 [verbose]: UpdateWindowStateAndSize Normal
[runtime] 2023-07-01 15:07:33 [verbose]: fetchWindowSize {Width=1366, Height=768} {Width=1366, Height=768}
[runtime] 2023-07-01 15:07:33 [verbose]: Updated display mode to desktop resolution: 1920x1080@60, SDL_PIXELFORMAT_ARGB8888
[runtime] 2023-07-01 15:07:33 [verbose]: fetchWindowSize {Width=1366, Height=768} {Width=1366, Height=768}
[runtime] 2023-07-01 15:07:33 [verbose]: fetchWindowSize {Width=1366, Height=768} {Width=1366, Height=768}
[runtime] 2023-07-01 15:07:33 [verbose]: fetchWindowSize {Width=1366, Height=768} {Width=2732, Height=1536}
[runtime] 2023-07-01 15:07:33 [verbose]: UpdateWindowStateAndSize Normal
[runtime] 2023-07-01 15:07:33 [verbose]: fetchWindowSize {Width=1366, Height=768} {Width=2732, Height=1536}
[runtime] 2023-07-01 15:07:33 [verbose]: Updated display mode to desktop resolution: 1792x1120@60, SDL_PIXELFORMAT_ARGB8888
[runtime] 2023-07-01 15:07:33 [verbose]: fetchWindowSize {Width=1366, Height=768} {Width=2732, Height=1536}
[runtime] 2023-07-01 15:07:33 [verbose]: fetchWindowSize {Width=1366, Height=768} {Width=2732, Height=1536}
[runtime] 2023-07-01 15:07:34 [verbose]: fetchWindowSize {Width=1366, Height=768} {Width=1366, Height=768}
[runtime] 2023-07-01 15:07:34 [verbose]: UpdateWindowStateAndSize Normal
[runtime] 2023-07-01 15:07:34 [verbose]: fetchWindowSize {Width=1366, Height=768} {Width=1366, Height=768}
[runtime] 2023-07-01 15:07:34 [verbose]: Updated display mode to desktop resolution: 1920x1080@60, SDL_PIXELFORMAT_ARGB8888


```

osu.Framework/Platform/SDL2Window_Windowing.cs
```diff
@@ -534,7 +534,9 @@ private void updateAndFetchWindowSpecifics()
                 windowState = pendingWindowState.Value;
                 pendingWindowState = null;
 
+                updatingWindowStateAndSize = true;
                 UpdateWindowStateAndSize(windowState, currentDisplay, currentDisplayMode.Value);
+                updatingWindowStateAndSize = false;
```
bdach, line 539: Why is the `updatingWindowStateAndSize` flag not being updated _inside_ `UpdateWindowStateAndSize()`? Seems error-prone.

osu.Framework/Platform/SDL2Window_Windowing.cs
```diff
@@ -722,6 +724,15 @@ private void storeWindowPositionToConfig()
         /// </summary>
         private bool storingSizeToConfig;
 
+        /// <summary>
+        /// Set when <see cref="UpdateWindowStateAndSize"/> is in progress to avoid <see cref="fetchWindowSize"/> being called with invalid data.
+        /// </summary>
+        /// <remarks>
+        /// Since <see cref="UpdateWindowStateAndSize"/> is a multi-step process, intermediary windows size changes might be invalid.
+        /// This is usually not a problem, but since <see cref="HandleEventFromFilter"/> runs out-of-band, invalid data might appear in those events.
```
bdach, line 732: I don't really follow this. Sure the filter may be running during the switch, but shouldn't it eventually settle on the final, correct size of the window?

osu.Framework.Tests/Visual/Platform/TestSceneWindowModeResizedEvent.cs
```diff
@@ -0,0 +1,90 @@
...
+        public TestSceneWindowModeResizedEvent()
+        {
+            Child = new WindowDisplaysPreview
+            {
+                RelativeSizeAxes = Axes.Both
+            };
+
+            AddLabel("set up defaults");
+            AddStep($"set windowed size to {windowed_size}", () => config.SetValue(FrameworkSetting.WindowedSize, windowed_size));
+            AddStep("set default fullscreen size", () => config.GetBindable<Size>(FrameworkSetting.SizeFullscreen).SetDefault());
```
bdach, line 39: This is the weirdest. Literally no other test except one adds fake labels like that. Can we just move these steps to a `[SetUpSteps]` method or something instead?

osu.Framework/Platform/SDL2Window_Windowing.cs
Susko3: Having it inside `UpdateWindowStateAndSize()` is error prone, as it won't work properly when overriding it.

bdach: And what if `UpdateWindowStateAndSize()` is called in another place and the caller forgets to set the flag?

osu.Framework/Platform/SDL2Window_Windowing.cs
Susko3: It seem that running things in a filter runs even before SDL has a chance to update it's internal state. As seen in https://github.com/ppy/osu-framework/pull/5878, the call to `SDL_Metal_GetDrawableSize` returns unwanted/invalid data in the second line:
```
Window Size:{Width=1304, Height=537} Drawable Size: {Width=2608, Height=1074}
****Window Size:{Width=1792, Height=1120} Drawable Size: {Width=2608, Height=1074}
```

> shouldn't it eventually settle on the final, correct size of the window? What does the invalid data break, precisely? 

It does settle, but the issue is that it doesn't invoke `Resized` when it's settled (at least when using Metal).

Besides, calling into events when we're in the process of changing the window state is not that great of an idea to begin with. The code in the filter was intended for when the window being resized, limiting to only that scenario would be best (not what this PR currently does).

osu.Framework.Tests/Visual/Platform/TestSceneWindowModeResizedEvent.cs
Susko3: Sure, I just wanted to avoid having them run with each test.

Susko3:
@AkiSakurai if you can repro that on `master`, please open it as a separate issue so it doesn't get lost here!
I was wondering if something like that would crop up 🙂

osu.Framework/Platform/SDL2Window_Windowing.cs
bdach: > It does settle, but the issue is that it doesn't invoke `Resized` when it's settled (at least when using Metal).

Presumably due to the "equals previous value" guard in setter of `Size`. I see.

osu.Framework/Platform/SDL2Window_Windowing.cs
Susko3: > Better yet if we could make the method `protected internal` so it's only our internal problem.

I think the entire `SDL2Window` should be `internal`, games should only really be using what's exposed in `IWindow`.

bdach: Let's just leave it as out of scope for now.

bdach (approved):
I think this makes sense

PR #5883, opened 2023-07-03
Reduce complexity overhead (and avoid one copy) when obtaining character SSDQs in `SpriteTextDrawNode`

peppy:
Of note, this doesn't show a huge improvement in benchmarks, but in an edge case noticed in osu!, the `AddRange` operation can add a perceivable overhead:

JetBrains Rider 2023-07-03 at 08 17 17

| Before | After |
| :---: | :---: |
| JetBrains Rider 2023-07-03 at 08 16 07 | JetBrains Rider 2023-07-03 at 08 15 11 |

We believe this may be a dotnet runtime quirk.

Even though this change can't be shown in isolated benchmarks, I'd argue the code quality improvement is worth it.

Of note, the caching of the SSDQs at `SpriteText` was a bit redundant as it was being invalidated by basically everything. So it makes sense to just fetch it every time, regardless, in the draw node itself.

The only potential saving we could obtain with the previous logic would be to shift the load to the `Update` frame. But this wasn't being done. Rather than investigating whether that has any benefits, I'd rather focus on getting `SpriteText` rewritten to use an optimised shader.

master

|           Method |        Mean |     Error |    StdDev | Allocated |
|----------------- |------------:|----------:|----------:|----------:|
|   TestStaticText |    984.4 us |   1.30 us |   1.22 us |       6 B |
|   TestMovingText |  2,701.5 us |  47.68 us |  42.27 us |     117 B |
| TestChangingText | 11,381.6 us | 161.87 us | 143.49 us |   64182 B |

this pr (pooled at 2967626edae655123d64ac5a8ef10d683ddece38)

|           Method |        Mean |     Error |    StdDev | Allocated |
|----------------- |------------:|----------:|----------:|----------:|
|   TestStaticText |    985.3 us |   2.46 us |   2.30 us |       6 B |
|   TestMovingText |  2,928.0 us |  57.80 us |  75.16 us |     119 B |
| TestChangingText | 11,426.4 us | 186.17 us | 155.46 us |   64184 B |

this pr (list at 90a2e286bc84e1bacbd76d8f32606ce7b09685de)

|           Method |        Mean |     Error |    StdDev | Allocated |
|----------------- |------------:|----------:|----------:|----------:|
|   TestStaticText |    985.5 us |   1.80 us |   1.68 us |       6 B |
|   TestMovingText |  2,476.1 us |  47.54 us |  50.86 us |     110 B |
| TestChangingText | 10,793.3 us | 211.44 us | 217.14 us |   64180 B |

### DIFF
--- a/osu.Framework/Graphics/Sprites/SpriteText.cs
+++ b/osu.Framework/Graphics/Sprites/SpriteText.cs
@@ -27,8 +27,6 @@ namespace osu.Framework.Graphics.Sprites
     /// </summary>
     public partial class SpriteText : Drawable, IHasLineBaseHeight, ITexturedShaderDrawable, IHasText, IHasFilterTerms, IFillFlowContainer, IHasCurrentValue<string>
     {
-        private const float default_text_size = 20;
-
         /// <remarks>
         /// <c>U+00A0</c> is the Unicode NON-BREAKING SPACE character (distinct from the standard ASCII space).
         /// <c>U+202F</c> is the Unicode NARROW NO-BREAK SPACE character.
@@ -56,8 +54,6 @@ namespace osu.Framework.Graphics.Sprites
             });
 
             AddLayout(charactersCache);
-            AddLayout(parentScreenSpaceCache);
-            AddLayout(localScreenSpaceCache);
             AddLayout(shadowOffsetCache);
             AddLayout(textBuilderCache);
         }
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/Sprites/SpriteText_DrawNode.cs
```diff
@@ -78,6 +80,31 @@ public override void Draw(IRenderer renderer)
...
+            private void updateScreenSpaceCharacters()
+            {
+                int partCount = Source.characters.Count;
+
+                parts ??= new List<ScreenSpaceCharacterPart>(partCount);
+                parts.Clear();
+
+                Vector2 inflationAmount = DrawInfo.MatrixInverse.ExtractScale().Xy;
+
+                foreach (var character in Source.characters)
```
bdach, line 96: If I have the capacity thing right above, then you could call `parts.EnsureCapacity(Source.characters.Count);` in this loop. The xmldoc [does not seem promising](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.ensurecapacity?view=net-7.0#definition), but it doesn't seem to match the [actual implementation](https://github.com/dotnet/runtime/blob/2fbb7eb1ba014de7200e39f048d1ea74a39f0b61/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L426-L466). This has theoretical gains in the cases where the count of character changes (rather than incurring copies on-the-fly as list capacity is expanded, incurring struct copies, the list capacity will be expanded once upfront and no copies will happen after that).

bdach (commented):
Would like @smoogipoo's take on the invalidation stuff if possible.

osu.Framework/Graphics/Sprites/SpriteText_DrawNode.cs
```diff
@@ -78,6 +80,31 @@ public override void Draw(IRenderer renderer)
...
             }
+
+            /// <summary>
+            /// The characters in screen space. These are ready to be drawn.
+            /// </summary>
+            private void updateScreenSpaceCharacters()
+            {
+                int partCount = Source.characters.Count;
+
+                parts ??= new List<ScreenSpaceCharacterPart>(partCount);
```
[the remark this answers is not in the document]
peppy: The copy I meant was the fact that items were previously added once to a list in `SpriteText` then subsequently copied to a second list in `DrawNode`.

peppy:
> Would like @smoogipoo's take on the invalidation stuff if possible.

I double-checked with him IRL on the invalidation part when making this change, so he should be on board with it 😄

osu.Framework/Graphics/Sprites/SpriteText_DrawNode.cs
peppy: Interesting. But wouldn't you want to call it just before the loop?

bdach: Uh, right. Not sure how I got _that_ mixed up...

peppy:
With `EnsureCapacity` (wouldn't expect a change, so just confirming it doesn't get worse):

|           Method |        Mean |    Error |   StdDev | Allocated |
|----------------- |------------:|---------:|---------:|----------:|
|   TestStaticText |    985.8 us |  0.99 us |  0.93 us |       6 B |
|   TestMovingText |  2,529.1 us | 49.84 us | 48.95 us |     108 B |
| TestChangingText | 10,811.6 us | 79.74 us | 74.59 us |   64166 B |

osu.Framework/Graphics/Sprites/SpriteText_DrawNode.cs
peppy: Have applied.

bdach (approved):

osu.Framework/Graphics/Sprites/SpriteText.cs
```diff
@@ -501,53 +497,6 @@ private void computeCharacters()
             }
         }
 
-        private readonly LayoutValue parentScreenSpaceCache = new LayoutValue(Invalidation.DrawSize | Invalidation.Presence | Invalidation.DrawInfo, InvalidationSource.Parent);
-        private readonly LayoutValue localScreenSpaceCache = new LayoutValue(Invalidation.MiscGeometry, InvalidationSource.Self);
```
smoogipoo, line 505: I suppose one of the advantages of this is that the overhead only occurred once per invalidation, rather than 3 times (once for each DrawNode) per invalidation.

peppy: The lookup was already lazy, so I'm not sure if this would ever be the case, but maybe.

PR #5884, opened 2023-07-03
Fix game benchmarks being completely broken

peppy:
| Before | After |
| :---: | :---: |
| JetBrains Rider 2023-07-03 at 08 26 51 | JetBrains Rider 2023-07-03 at 08 25 00 |

### DIFF
--- a/osu.Framework.Benchmarks/GameBenchmark.cs
+++ b/osu.Framework.Benchmarks/GameBenchmark.cs
@@ -110,7 +110,9 @@ namespace osu.Framework.Benchmarks
 
             public override void RunMainLoop()
             {
-                if (!RunOnce.Wait(10000))
+                // This is set very high as we rely on it blocking frames from running from the background thread
+                // but need the look to continue to stay active until the benchmark finishes, to handle disposal.
+                if (!RunOnce.Wait(120000))
                     throw new TimeoutException("Run request didn't arrive for a long time");
 
                 RunSingleFrame();
--- a/osu.Framework.Benchmarks/GameBenchmark.cs
+++ b/osu.Framework.Benchmarks/GameBenchmark.cs
@@ -104,6 +104,9 @@ namespace osu.Framework.Benchmarks
             public void RunSingleFrame()
             {
                 ExecutionMode = ExecutionMode.SingleThread;
+
+                // Importantly, this calls the base method, bypassing the custom wait logic below
+                // (which is blocking execution by thread runner while the benchmark runs).
                 base.RunMainLoop();
             }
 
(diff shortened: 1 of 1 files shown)


osu.Framework.Benchmarks/GameBenchmark.cs
```diff
@@ -110,7 +110,9 @@ public void RunSingleFrame()
 
             public override void RunMainLoop()
             {
-                if (!RunOnce.Wait(10000))
+                // This is set very high as we rely on it blocking frames from running from the background thread
+                // but need the look to continue to stay active until the benchmark finishes, to handle disposal.
+                if (!RunOnce.Wait(120000))
```
bdach, line 115: Not hugely fond of this solution. What if someone writes a benchmark which has a total running time in excess of 2 minutes? It's not hugely unheard of with Benchmark.NET (because as I understand it, this needs to block for the entirety of the time that Benchmark.NET is doing its thing? the comment doesn't explain very well, I'm mostly gathering from my own experimentation).

I'd probably rather have

```diff
diff --git a/osu.Framework.Benchmarks/GameBenchmark.cs b/osu.Framework.Benchmarks/GameBenchmark.cs
index 395916ccc..6fedcc79d 100644
--- a/osu.Framework.Benchmarks/GameBenchmark.cs
+++ b/osu.Framework.Benchmarks/GameBenchmark.cs
@@ -110,10 +110,9 @@ public void RunSingleFrame()
 
             public override void RunMainLoop()
             {
-                // This is set very high as we rely on it blocking frames from running from the background thread
-                // but need the look to continue to stay active until the benchmark finishes, to handle disposal.
-                if (!RunOnce.Wait(120000))
-                    throw new TimeoutException("Run request didn't arrive for a long time");
+#pragma warning disable RS0030
+                RunOnce.Wait();
+#pragma warning restore RS0030
 
                 RunSingleFrame();
                 RunOnce.Reset();

```

(effectively locally reverting https://github.com/ppy/osu-framework/pull/5259) than this.

peppy: That sounds fine too.

A saving grace is that BDN should never hit 2 minutes unless the benchmark is flawed, but I agree that we don't want to skew results even in such a flawed cases.

smoogipoo:
Am I not understanding things, or is this 10 seconds _per iteration_ of the benchmark?

This needs some more commenting...

peppy:
The benchmark would run the workload for over 10 seconds (including warmup time etc.) and after 10 seconds the game would start processing frames, crash, and `RunSingleFrame` would take nanoseconds instead of actual time.

You can repro by running any game benchmark and watching the benchmark output go silly fast halfway through without this PR.

smoogipoo (approved):

PR #5888, opened 2023-07-04
Add basic "render scale" support for mobile devices

frenzibyte:
Opened as draft as it's been mentioned that this breaks front-to-back passes on Android. Probably requires further investigation before merge, but opening it anyway rather than leaving the change in my machine.

### DIFF
--- a/osu.Framework.Tests/Visual/Platform/TestSceneRenderer.cs
+++ b/osu.Framework.Tests/Visual/Platform/TestSceneRenderer.cs
@@ -6,9 +6,11 @@ using NUnit.Framework;
 using osu.Framework.Allocation;
 using osu.Framework.Configuration;
 using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Sprites;
 using osu.Framework.Graphics.UserInterface;
 using osu.Framework.Platform;
+using osuTK;
 
 namespace osu.Framework.Tests.Visual.Platform
 {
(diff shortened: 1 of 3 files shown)


peppy:
I think I'd want to see this avoid the `BufferedContainer` overhead when render scale is set to 1.0.

frenzibyte:
That's alrwady the case? When render scale is set to 1.0x, the game content is moved away from the buffered container.

peppy:
Missed that somehow.

peppy:
I'm not sure how much benefit android gets from having front-to-back enabled, so one option would be to disable it and get this in. @bdach don't suppose you've tested front to back recently with osu! scenarios to see if it's beneficial (with one or more overlays on screen)?

peppy (approved):

bdach:
I'll take a look tomorrow. Ran out of time today.

bdach:
So, on front-to-back on android: in my rather crude, oneshot, experimental testing on a device sample of N=1 it seems to either have no benefit at all, or to make things worse:

| bypass off | bypass on |
| :-: | :-: |
| Screenshot_2023-07-05-19-55-19-29_25a1a32208bbcdc1d450b7aa854bb161 | Screenshot_2023-07-05-19-55-11-11_25a1a32208bbcdc1d450b7aa854bb161 |

Note the 4fps difference. It was a pretty consistent one, too.

As for render scale, it appears to help some on android fps-wise:

| 1.0x renderscale | ~0.5x renderscale |
| :-: | :-: |
| Screenshot_2023-07-05-20-02-08-68_fca46a2731136e673b8638a972bc0192 | Screenshot_2023-07-05-20-04-43-64_fca46a2731136e673b8638a972bc0192 |

I also attempted to test game with renderscale via

```diff
diff --git a/osu.Game/OsuGame.cs b/osu.Game/OsuGame.cs
index 5b654e0c16..d4ae66e20c 100644
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -124,6 +124,9 @@ public partial class OsuGame : OsuGameBase, IKeyBindingHandler<GlobalAction>, IL
         [Resolved]
         private FrameworkConfigManager frameworkConfig { get; set; }
 
+        [Resolved]
+        private FrameworkDebugConfigManager frameworkDebugConfig { get; set; }
+
         [Cached]
         private readonly DifficultyRecommender difficultyRecommender = new DifficultyRecommender();
 
@@ -810,6 +813,9 @@ protected override void LoadComplete()
         {
             base.LoadComplete();
 
+            frameworkConfig.SetValue(FrameworkSetting.RenderScale, 0.6f);
+            frameworkDebugConfig.SetValue(DebugSetting.BypassFrontToBackPass, true);
+
             var languages = Enum.GetValues<Language>();
 
             var mappings = languages.Select(language =>

```

and got this:

| bypass off | bypass on |
| :-: | :-: |
| Screenshot_2023-07-05-20-23-57-42_25a1a32208bbcdc1d450b7aa854bb161 | Screenshot_2023-07-05-20-23-47-80_25a1a32208bbcdc1d450b7aa854bb161 |

so modest gains. However, this same patch above also revealed the following breakage of intro:

https://github.com/ppy/osu-framework/assets/20418176/32f804ca-1d52-4866-bedd-8a835df85990

Note the completely missing osu! logo animation. This is not an android issue, happens on desktop with the patch above as well.

bdach (changes requested):
above is probably a blocker

peppy:
Front-to-back also has an overhead on metal/intel (1ms draw time on the same test as @bdach did above in osu!).

@smoogipoo thoughts on whether it makes sense to leave it enabled at this point? I guess we should check direct x as well then make a call?

peppy:
Getting some flickering with this change (when viewing #5904). Weirdly, it doesn't happen when testing in `TestSceneRenderer`.

https://github.com/ppy/osu-framework/assets/191335/28d4aad2-dc5e-44a6-8f46-40b2b6a454c2


Blocking pending further investigation..

peppy:
This should use `pixelSnapping` after https://github.com/ppy/osu-framework/pull/5926 is merged. It shouldn't be a huge concern if the container is always at (0,0), but `Nearest` scaling will look better (prefer pixellated rather then badly blended).

PR #5894, opened 2023-07-05
Re-use vertex buffers after batch flushes until filled completely in Veldrid

### LINKED ISSUE #5892: Veldrid renderer should keep re-using a vertex buffer until it's filled with vertices
There's about 1ms encoding overhead from switching to a new vertex buffer on every draw call in a 500 draw calls scenario.

frenzibyte:
- Closes #5892 

The aggregate overhead of switching vertex buffers per draw call in a 500 draw calls scenario is non-negligible on iOS (around 1ms encoding time). Needs performance testing on other platforms & backends.

Before:

IMG_6081

After:

IMG_6080

Notice the draw time decrease from ~12ms to ~11ms.


osu.Framework/Graphics/Veldrid/Batches/VeldridVertexBatch.cs
```diff
@@ -82,21 +84,25 @@ void IVertexBatch.ResetCounters()
         /// Adds a vertex to this <see cref="VeldridVertexBatch{T}"/>.
         /// </summary>
         /// <param name="v">The vertex to add.</param>
+        // todo: this might break when the vertices that are about to be added cannot all fit into this buffer.
```
peppy, line 87: How are we supposed to consider this in review? Is this a proof of concept or RFC pull request?

peppy:
iOS may be file due to direct buffer copies, but this has the potential of blocking the pipeline elsewhere and will require either ensuring we are doing direct-to-memory non-contentious uploads, or testing across all platforms / devices.

At the same point, I'd want to revisit the triple buffer removal in this class. But I still haven't gotten any feedback from @smoogipoo on how to test it properly (I haven't been able to reproduce any contention even with the proposed test method)

https://discord.com/channels/188630481301012481/589331078574112768/1121652595572887612

osu.Framework/Graphics/Veldrid/Batches/VeldridVertexBatch.cs
frenzibyte: Uh I've checked this one already but forgot to remove the TODO comment. I was thinking this would break in a scenario where draw calls with different number of vertices (e.g. drawing a quad followed by a triangle), but turns out `DrawTriangle` actually draws 4 vertices rather than 3 as well, and the batch itself is supposed to be on quads only (i.e. always batching 4 vertices together), therefore this comment is pretty much invalid.

smoogipoo:
There's no way this is good. We already tried this once before, and I think it was macOS Apple Silicon or x86 that behaved poorly.

peppy:
On metal (desktop / intel) there's negligible change in performance and frame stability. Weird that it's more visible on iOS. Will re-test on M2 later today.

peppy:
> Notice the draw time decrease from ~12ms to ~11ms.

Notice the input time increase from ~4 to ~5.5ms in the other direction.

Can you understand why I asked for the graphs now?

peppy:
I've removed the nullability you added. It felt very hard to understand.

peppy:
Windows / d3d:

| Before | After |
|:------:|:-----:|
|image|image|

peppy (approved):

PR #5896, opened 2023-07-05
Reduce VBO buffer count to 2

peppy:
As discussed in far too much length. We can test reducing this further at a later point. I've added an envvar to aid in this (`OSU_GRAPHICS_VBO_COUNT`).


smoogipoo (approved):

peppy:
I wanted to tidy up this file a bit more, but will leave that until after https://github.com/ppy/osu-framework/pull/5894 is considered (made sure this branch doesn't conflict with that PR).

PR #5897, opened 2023-07-06
Make frame statistics display usable on mobile platforms

frenzibyte:
RFC. Only affects debug builds.

This pull request adds the following features:
- Double-click on any area on the bottom right corner of the screen to cycle between frame statistics display (and also make it visible from hidden state).
- Tap with a single finger to expand the frame statistics display when in full state (similar to holding <kbd>Ctrl</kbd>).
- Tap with a second finger to freeze the frame statistics display (similar to holding <kbd>Shift</kbd>).

Definitely not the best UX, but it would help to have anything like this to easily view the frame graphs and counters without modifying my builds and waiting for compilation every time.

It'll still be annoying to use this right now because it only works on debug builds and I always build in release to test performance, but at least it'll be a one-line change to utilise the display rather than re-implementing it or looking for a patch.


osu.Framework/Game.cs
```diff
@@ -482,5 +499,25 @@ protected override void Dispose(bool isDisposing)
...
+        private partial class FrameStatisticsTouchReceptor : Drawable
+        {
+            private readonly Game game;
+
+            public FrameStatisticsTouchReceptor(Game game)
+            {
+                this.game = game;
+            }
+
+            protected override bool OnClick(ClickEvent e) => true;
```
peppy, line 512: Have you tested this with osu..? It feels like blocking clicks in this area could have adverse effects?

Any reason it's not only handling touch as well?

osu.Framework/Game.cs
```diff
@@ -263,6 +265,18 @@ protected override void LoadComplete()
             }, overlayContent.Add);
 
             FrameStatistics.BindValueChanged(e => performanceOverlay.State = e.NewValue, true);
+
+            if (DebugUtils.IsDebugBuild)
+            {
+                Add(new FrameStatisticsTouchReceptor(this)
```
peppy, line 271: Should this be `AddInternal`? This is going to add inside user defined content, which will likely break in some cases.

peppy: Almost think it would be better off to have this be a visible element on the screen for touch devices so you know where to interact with it.

osu.Framework/Graphics/Performance/PerformanceOverlay.cs
```diff
@@ -71,6 +75,71 @@ protected override void LoadComplete()
...
+
+            base.OnKeyUp(e);
+        }
+
+        protected override bool OnTouchDown(TouchDownEvent e)
+        {
+            switch (e.Touch.Source)
+            {
+                case TouchSource.Touch1:
+                    if (State == FrameStatisticsMode.Full)
```
peppy, line 125: This would work better if it only expanded while holding, and reverted on `TouchUp`. Same goes for the two finger version below.

peppy (changes requested):

osu.Framework/Game.cs
frenzibyte: The component is placed below everything in the game, so as long as absolutely nothing handles click at the moment I touch the screen, the game will respond to that and handle the double click to cycle between frame statistics modes.

osu.Framework/Game.cs
frenzibyte: Sure, it can be `AddInternal`, but I don't know about making it visible, I would hope for it to be as hidden as possible so it doesn't potentially obscure anything in the area it handles.

osu.Framework/Graphics/Performance/PerformanceOverlay.cs
frenzibyte: For sure, that was how I had it initially, but I switched to this behaviour because originally I made the performance overlay change modes by holding in the bottom-right area, which was conflicting with this. But now that I changed it to double-click, we can make this work using hold instead.

osu.Framework/Game.cs
```diff
@@ -263,6 +265,18 @@ protected override void LoadComplete()
             }, overlayContent.Add);
 
             FrameStatistics.BindValueChanged(e => performanceOverlay.State = e.NewValue, true);
+
+            if (DebugUtils.IsDebugBuild)
```
frenzibyte, line 269: I've touched on this in OP, but this line has been a burden ever since I've first utilised this PR for my investigations. Because, in order for the numbers to have any meaning, I need to build in "release" mode with AOT enabled. And 99% the times I would forget to comment this line, and the end up waiting another 30 minutes for another build which has this line commented.

How opposed would you be towards making an envvar for this instead? That way I can have local run configurations which enable this, but mainly I wouldn't have to wait for another build to have the receptor added.

osu.Framework/Game.cs
```diff
@@ -263,6 +265,18 @@ protected override void LoadComplete()
             }, overlayContent.Add);
 
             FrameStatistics.BindValueChanged(e => performanceOverlay.State = e.NewValue, true);
+
+            if (Environment.GetEnvironmentVariable("OSU_FRAME_STATISTICS_VIA_TOUCH") == "1" || DebugUtils.IsDebugBuild)
```
smoogipoo, line 269: Move to `FrameworkEnvironment` please.

peppy (approved):

PR #5901, opened 2023-07-06
Avoid initialising textures when unnecessary

peppy:
It turns out we were initialising every mipmap level of every textures, while it is only actually required in scenarios we are not already replacing every pixel of the texture.

For framework usages, this is for atlases alone.

I've implemented this by changing the `initialisationColour` to be nullable. For native textures, it has a `null` default, which is perfect to cover all overrides (`VideoTexture` and `FrameBufferTexture`).

For the `IRenderer` exposed usage, the parameter was already required, so there is no default. This is good as it ensures a consume isn't going to see a change in behaviour out-of-the-blue.

---

At least on metal, this has *insane* performance effects across the whole game. Basically all my stutters are gone, including song select and player loading transition. These stutters were caused from buffered containers changing size and triggering this upload every frame.

Of note, the overhead is still the texture upload overhead, which @frenzibyte has touched on previously (I believe it's on his list of things to address). But without the insane number of textures uploads, the remaining ones generally don't exceed allocated frame time.

### DIFF
--- a/osu.Framework/Graphics/OpenGL/Textures/GLTexture.cs
+++ b/osu.Framework/Graphics/OpenGL/Textures/GLTexture.cs
@@ -69,7 +69,7 @@ namespace osu.Framework.Graphics.OpenGL.Textures
         private readonly List<RectangleI> uploadedRegions = new List<RectangleI>();
 
         private readonly All filteringMode;
-        private readonly Color4 initialisationColour;
+        private readonly Color4? initialisationColour;
 
         /// <summary>
         /// Creates a new <see cref="GLTexture"/>.
@@ -79,8 +79,8 @@ namespace osu.Framework.Graphics.OpenGL.Textures
         /// <param name="height">The height of the texture.</param>
         /// <param name="manualMipmaps">Whether manual mipmaps will be uploaded to the texture. If false, the texture will compute mipmaps automatically.</param>
         /// <param name="filteringMode">The filtering mode.</param>
-        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads).</param>
-        public GLTexture(GLRenderer renderer, int width, int height, bool manualMipmaps = false, All filteringMode = All.Linear, Color4 initialisationColour = default)
+        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads). If null, not initialisation is provided out-of-the-box.</param>
+        public GLTexture(GLRenderer renderer, int width, int height, bool manualMipmaps = false, All filteringMode = All.Linear, Color4? initialisationColour = null)
         {
             Renderer = renderer;
             Width = width;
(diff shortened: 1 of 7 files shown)


osu.Framework/Graphics/OpenGL/Textures/GLTexture.cs
```diff
@@ -79,8 +79,8 @@ public string Identifier
         /// <param name="height">The height of the texture.</param>
         /// <param name="manualMipmaps">Whether manual mipmaps will be uploaded to the texture. If false, the texture will compute mipmaps automatically.</param>
         /// <param name="filteringMode">The filtering mode.</param>
-        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads).</param>
-        public GLTexture(GLRenderer renderer, int width, int height, bool manualMipmaps = false, All filteringMode = All.Linear, Color4 initialisationColour = default)
+        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads). If null, not initialisation is provided out-of-the-box.</param>
```
smoogipoo, line 82:
```suggestion
        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads). If null, no initialisation is provided out-of-the-box.</param>
```

osu.Framework/Graphics/Rendering/IRenderer.cs
```diff
@@ -372,10 +372,18 @@ public interface IRenderer
...
 
         /// <summary>
-        /// Creates a new texture.
+        /// Creates a new <see cref="Texture"/>.
         /// </summary>
+        /// <param name="width">The width of the texture.</param>
+        /// <param name="height">The height of the texture.</param>
+        /// <param name="manualMipmaps">Whether manual mipmaps will be uploaded to the texture. If false, the texture will compute mipmaps automatically.</param>
+        /// <param name="filteringMode">The filtering mode.</param>
+        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads). If null, not initialisation is provided out-of-the-box.</param>
```
smoogipoo, line 381:
```suggestion
        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads). If null, no initialisation is provided out-of-the-box.</param>
```

osu.Framework/Graphics/Rendering/Renderer.cs
```diff
@@ -1084,10 +1084,10 @@ internal IShader GetMipmapShader()
         /// <param name="height">The height of the texture.</param>
         /// <param name="manualMipmaps">Whether manual mipmaps will be uploaded to the texture. If false, the texture will compute mipmaps automatically.</param>
         /// <param name="filteringMode">The filtering mode.</param>
-        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads).</param>
+        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads). If null, not initialisation is provided out-of-the-box.</param>
```
smoogipoo, line 1087:
```suggestion
        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads). If null, no initialisation is provided out-of-the-box.</param>
```

osu.Framework/Graphics/Veldrid/Textures/VeldridTexture.cs
```diff
@@ -83,9 +83,9 @@ protected virtual TextureUsage Usages
         /// <param name="height">The height of the texture.</param>
         /// <param name="manualMipmaps">Whether manual mipmaps will be uploaded to the texture. If false, the texture will compute mipmaps automatically.</param>
         /// <param name="filteringMode">The filtering mode.</param>
-        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads).</param>
+        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads). If null, not initialisation is provided out-of-the-box.</param>
```
smoogipoo, line 86:
```suggestion
        /// <param name="initialisationColour">The colour to initialise texture levels with (in the case of sub region initial uploads). If null, no initialisation is provided out-of-the-box.</param>
```

smoogipoo (changes requested):

smoogipoo (approved):

frenzibyte (commented):
Late review but this looks pretty good.

Issue #5902, opened 2023-07-06
Consider switching back to AOT compilation on iOS for potentially better performance


frenzibyte:
See [discord discussions](https://discord.com/channels/188630481301012481/1116207304681992212/1126392761189412994).

- [x] Switch o!f projects to use AOT compilation (https://github.com/ppy/osu-framework/pull/5903)
- [x] Switch osu! projects to use AOT compilation (https://github.com/ppy/osu/pull/24338)

PR #5903, opened 2023-07-07
Switch osu!framework iOS projects to use AOT compilation on release configurations

frenzibyte:
- [x] Depends on https://github.com/ppy/veldrid/pull/29
- [x] Depends on https://github.com/ppy/SDL2-CS/pull/169

First step towards #5902. After recent investigations, turns out Mono interpreter isn't as efficient as AOT is, and is the cause of many minor overheads that are affecting draw time negatively.

Switching osu! to use AOT compilation is going to be a considerable effort so I'll change this on o!f first and begin investigating iOS performance on a higher scale before moving into osu!.

Note that we're intentionally not using the `--interpreter=all` flag here (which allows the app to fall back to interpretation for dynamic code rather than throw exception), as we don't want to rely on interpretation at all until we come across a case where we can't help it.

### DIFF
--- a/osu.Framework.iOS.props
+++ b/osu.Framework.iOS.props
@@ -8,11 +8,11 @@
     <NoWarn>$(NoWarn);MT7091</NoWarn>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)' == 'Debug'">
-    <!-- On debug configurations, we need Mono interpreter for faster compilation. -->
+    <!-- On debug configurations, we use Mono interpreter for faster compilation. -->
     <UseInterpreter>true</UseInterpreter>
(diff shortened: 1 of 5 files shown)


peppy:
> Note that we're intentionally not using the --interpreter=all flag here (which allows the app to fall back to interpretation for dynamic code rather than throw exception), as we don't want to rely on interpretation at all until we come across a case where we can't help it.

I believe this will mean switching signalr back to crap mode, which I'd rather avoid if we can (at least osu! side).

But I guess this is a per-project setting so that's fine?

frenzibyte:
Yeah we can set this flag locally in osu! if we want to.

peppy (approved):

PR #5904, opened 2023-07-07
Add basic visual performance test scenes

frenzibyte:
Test scenes that were useful for me while profiling iOS performance. Tidied them up to rely on bindables, but I'm not a huge fan of the way I have it structured. Since they're test scenes in the end, I didn't want to focus too much on the code quality.

CleanShot 2023-07-07 at 08 47 19
CleanShot 2023-07-07 at 08 48 27
CleanShot 2023-07-07 at 08 49 23
CleanShot 2023-07-07 at 08 49 57
CleanShot 2023-07-07 at 08 50 12


peppy:
Have you considered making the render scale adjust local to these tests?

frenzibyte:
The history of this branch was quite messed up, so I've rewrote it again. Also since we're making render scaling local to the test scene (which is a good idea), I've removed the dependency on the global render scale PR.

osu.Framework/Game.cs
```diff
@@ -265,7 +265,7 @@ protected override void LoadComplete()
             FrameStatistics.BindValueChanged(e => performanceOverlay.State = e.NewValue, true);
         }
 
-        protected readonly Bindable<FrameStatisticsMode> FrameStatistics = new Bindable<FrameStatisticsMode>();
+        protected internal readonly Bindable<FrameStatisticsMode> FrameStatistics = new Bindable<FrameStatisticsMode>();
```
bdach, line 268: What's this for? None of the new tests seem to be using this?

```suggestion
        protected readonly Bindable<FrameStatisticsMode> FrameStatistics = new Bindable<FrameStatisticsMode>();
```

osu.Framework/Graphics/Rendering/IRenderer.cs
```diff
@@ -208,7 +208,7 @@ public interface IRenderer
         /// Flushes the currently active vertex batch.
         /// </summary>
         /// <param name="source">The source performing the flush, for profiling purposes.</param>
-        internal void FlushCurrentBatch(FlushBatchSource? source);
+        public void FlushCurrentBatch(FlushBatchSource? source);
```
bdach, line 211: This one also seems to be unused in the tests?

So either

```suggestion
        internal void FlushCurrentBatch(FlushBatchSource? source);
```

to revert, or

```suggestion
        void FlushCurrentBatch(FlushBatchSource? source);
```

to remove the redundant modifier.

osu.Framework.Tests/Visual/Performance/TestSceneTexturePerformance.cs
```diff
@@ -0,0 +1,150 @@
...
+
+                    store?.Dispose();
+                    store = new TextureStore(renderer, host.CreateTextureLoaderStore(new NamespacedResourceStore<byte[]>(game.Resources, @"Textures")), manualMipmaps: v.NewValue);
+                }, true);
+
+                UniqueTextures.BindValueChanged(v =>
+                {
+                    if (v.NewValue)
+                        Texture = store!.Get("sample-texture");
+                }, true);
```
bdach, line 118: This is weird because untoggling this toggle does... nothing? So once you go unique textures, you can never actually go back?

This combined with the weird passing of the textures into this component (which I'm not even 100% sure is not memory-leaky) just makes me think the stores should be managed locally by this component.

osu.Framework.Tests/Visual/Performance/PerformanceTestScene.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            base.Content.Child = content = new BufferedContainer
+            {
+                RelativeSizeAxes = Axes.Both,
+            };
+
+            AddLabel("General");
+            AddToggleStep("hide content", v => Content.Alpha = v ? 0 : 1);
+            AddToggleStep("bypass front to back", v => bypassFrontToBack.Value = v);
+            AddSliderStep("render scale", 0f, 1f, 1f, v => content.FrameBufferScale = new Vector2(v));
```
bdach, line 38: I'm not sure we want `0f` to be the minimum here. Surprised it doesn't just crash.

osu.Framework.Tests/Visual/Performance/TestSceneSpriteTextPerformance.cs
```diff
@@ -0,0 +1,72 @@
...
+            var text = new StringBuilder();
+
+            for (int p = 0; p < paragraphsCount; p++)
+            {
+                for (int w = 0; w < wordsCount; w++)
+                {
+                    for (int c = 0; c < wordLength; c++)
+                    {
+                        char character = (char)RNG.Next('a', 'z');
+                        character = c == 0 ? char.ToUpperInvariant(character) : char.ToLowerInvariant(character);
```
bdach, line 52: Kinda redundant. Character is known to be lowercase already.

```suggestion
                        character = c == 0 ? char.ToUpperInvariant(character) : character;
```

osu.Framework/Game.cs
frenzibyte: Drive-by change, whoops.

osu.Framework/Graphics/Rendering/IRenderer.cs
frenzibyte: Another drive-by change :eyes:

osu.Framework.Tests/Visual/Performance/TestSceneTexturePerformance.cs
frenzibyte: Good catch. Not sure how passing the textures could potentially leak memory though, they get finalised once they're not used anymore, so it's fine to pass them wherever as long as one of the usages don't continue to exist after the test scene is destructed (which I can't imagine happening).

I don't quite understand your conclusion though, you mean the mipmapped/non-mipmapped textures should be retrieved by this sprite rather than passing them by constructor? Because, well, it's better to retrieve the textures in test scene BDL, since it's loaded asynchronously.

bdach: > Not sure how passing the textures could potentially leak memory though

I don't particularly think they do but I'm not sure. The usage that made me pause is the `TextureStore` which is created to fetch a texture and then immediately thrown away in BDL. I *think* that as long as no uploads are created by the store it's safe?

And yeah on second thought it may not be feasible to manage the texture store from within the component because there's no good way to share the texture  from there. The point about untoggling still stands.

peppy:
Going to get this in. Feels like a great initial state.

peppy (approved):

PR #5906, opened 2023-07-07
Add source generator for IsLongRunning property

smoogipoo:
Supersedes https://github.com/ppy/osu-framework/pull/5763

Compared to the above PR, this uses the newer structure of `HandleInputCache` rather than polluting dependency injection.

Removes yet another usage of reflection inside `Drawable`. This is also queried every `Drawable` load, so could add up very easily.

I've moved the property to `Drawable_IsLongRunning.cs` just to isolate the potentially-reflection part of `Drawable` away from the main file. This way it makes it easier to remove the `using System.Reflection` directive and see what's left.

|                                 Method |         Mean |     Error |    StdDev | Allocated |
|--------------------------------------- |-------------:|----------:|----------:|----------:|
|       QueryNonLongRunningViaReflection |   654.947 ns | 9.9584 ns | 8.8279 ns |      48 B |
|          QueryLongRunningViaReflection | 1,246.236 ns | 9.0457 ns | 7.5536 ns |     128 B |
| QueryNonLongRunningViaSourceGeneration |     6.426 ns | 0.0713 ns | 0.0632 ns |         - |
|    QueryLongRunningViaSourceGeneration |     6.527 ns | 0.1540 ns | 0.1582 ns |         - |


bdach:
Seems good. Can reproduce benchmarked gains, framework and game tests pass.

@peppy Gonna need a sourcegen package here to proceed.

bdach (dismissed):

peppy (approved):

PR #5909, opened 2023-07-07
Apply code changes required for a future update to .NET 7/8

smoogipoo:
This just allows us to target .NET 7/8 without unexpected behaviour.

## [Fix Linux/Unix storage paths](https://github.com/ppy/osu-framework/commit/d7ff5d6478ada5017518aa19843f3fdca0d174cf)

`MyDocuments` and `Personal` are [documented](https://learn.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=net-7.0#fields) as referring to the same thing.

There was a bug with previous versions of .NET where `Personal` would incorrectly refer to `$HOME` on Unix/Linux. Since 2022-10 (so this change is likely also required for net7.0), this now correctly refers to `$HOME/Documents`

The runtime change that caused this is https://github.com/dotnet/runtime/pull/76250.


osu.Framework.Tests/Polygons/ConvexPolygonClippingTest.cs
```diff
@@ -300,12 +300,12 @@ public void TestFuzzedEdgeCases(Vector2[] clipVertices, Vector2[] subjectVertice
         }
 
         private Span<Vector2> clip(SimpleConvexPolygon clipPolygon, SimpleConvexPolygon subjectPolygon)
-            => new ConvexPolygonClipper<SimpleConvexPolygon, SimpleConvexPolygon>(ref clipPolygon, ref subjectPolygon).Clip();
+            => new ConvexPolygonClipper<SimpleConvexPolygon, SimpleConvexPolygon>(clipPolygon, subjectPolygon).Clip();
```
frenzibyte, line 303: For anyone wondering, the change appears to be described in details [here](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/low-level-struct-improvements). TL;DR: it's a C# 11.0 feature that improves safety of passing variables by reference, along with the definition of `scoped` keyword which "allows" this code to work as-is, but violates the rule anyway and the change in this PR is more correct). 

CleanShot 2023-07-08 at 06 32 22

osu.Framework/Utils/ConvexPolygonClipper.cs
```diff
@@ -46,7 +52,7 @@ public int GetClipBufferSize()
         /// </summary>
         /// <param name="buffer">The buffer to contain the clipped vertices. Must have a length of <see cref="GetClipBufferSize"/>.</param>
         /// <returns>A clockwise-ordered set of vertices representing the result of clipping <see cref="subjectPolygon"/> by <see cref="clipPolygon"/>.</returns>
-        public Span<Vector2> Clip(in Span<Vector2> buffer)
+        public Span<Vector2> Clip(Span<Vector2> buffer)
```
frenzibyte, line 55: This method can still pass the buffer span by reference since its usage is scoped to this method, but that will require prefixing the parameter with the `scoped` keyword which is a C# 11 feature, and we've also disabled front-to-back passes which is where this method is used frequently, so... I'll keep this PR open since it's not required anywhere currently.

peppy (approved):

PR #5911, opened 2023-07-08
Fix Veldrid texture uploads allocating new command buffers and resources

### LINKED ISSUE #5895: Veldrid renderer shouldn't allocate new command buffers during texture uploads (and re-use staging textures where possible)

frenzibyte:
- Closes #5895 

`GraphicsDevice.UpdateTexture` is very inefficient for dynamic textures, since it allocates both a local command buffer and a staging texture for perform the upload.

This PR comes with two changes (skipping the minor rename commit), the first one which changes texture uploads to be performed all on one command buffer (`BufferUpdateCommands`), and the second one caches staging textures to cut the allocation overhead.

The final numbers for Metal don't seem to be as good as they should be, still investigating that.

Profiling with [this test case](https://github.com/ppy/osu-framework/pull/5904/commits/bcadb6564fb2124bf14f9a9d11e2979a5c5d77e4) shows the following:
 - `master`:

   image

 - After using `BufferUpdateCommands`:

   image

 - After caching staging textures:

   image

Not totally sure why `CopyTexture` shows different numbers, but the number is consistent per commit anyway.


peppy:
m2:

| Before | After |
| :---: | :---: |
| 2023-07-09 10 54 48@2x | 2023-07-09 10 56 12@2x |

osu.Framework/Graphics/Veldrid/VeldridStagingTexturePool.cs
```diff
@@ -0,0 +1,85 @@
...
+                }
+            }
+
+            var texture = renderer.Factory.CreateTexture(TextureDescription.Texture2D((uint)width, (uint)height, 1, 1, format, TextureUsage.Staging));
+            used.Add(new StagingTextureCache(texture, renderer.FrameIndex));
+            stat_used.Value++;
+            return texture;
+        }
+
+        public void ReturnTextures(ulong latestFrame)
```
peppy, line 48: This flow has me a bit confused. Usually you return an object manually. Also this always seems to just be one frame behind, so why can't it be `NewFrame` and always return all textures?

frenzibyte: Returning all textures is unsafe, especially when the texture has been used last frame, since its content may have not been copied to the actual texture by the GPU yet.

The concept of returning an object manually doesn't really work out here because it's being written by the CPU and read by the GPU, therefore the only way to safely know when it can be returned to the pool is by checking whether the GPU is done with it every `BeginFrame` cycle.

I could rename this method to `SignalUsageFinish` or otherwise if that makes this any clearer (probably need xmldoc as well).

peppy: How about renaming to `NewFrame` or similar, as we already use that elsewhere? Definitely needs xmldoc stats assertions (staging buffers are always used "one frame maximum" etc.)

frenzibyte: I've changed the flow and added xmldoc now, see if it looks better now. It might read better to rename this to `Update`, but I've gone with `NewFrame` as proposed.

peppy:
windows:

| Before | After |
| :---: | :---: |
| osu Framework Tests_2023-07-09_14-25-30|image|

osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
```diff
@@ -205,6 +219,24 @@ protected override void Initialise(IGraphicsSurface graphicsSurface)
...
         {
+            for (int i = 0; i < pendingFramesFences.Count; i++)
+            {
+                var fence = pendingFramesFences[i];
+
+                // Frame usages are assumed to be sequential and linear.
+                //
+                // Therefore to cover any edge cases where signal may not return, or return too late, we
+                // check all and take on the frame index of the latest fence rather than breaking at the first.
+                if (!fence.Fence.Signaled) continue;
```
peppy, line 230: As discussed in discord, I'd rather be able to fix this to be a more correct algorithm and also ensure against leaks. So let's figure the veldrid changes first.

peppy: Fixed in 1cc1e28.

peppy (approved):

peppy (approved):

PR #5914, opened 2023-07-10
Replace most usages of `IsDynamicCodeCompiled` with `IsDynamicCodeSupported`

frenzibyte:
`IsDynamicCodeSupported` returns true in iOS when building with interpreter mode on, so it's probably better to use that instead (kind of as a way of testing dynamic code on iOS). Note that I have kept the `TransformCustom` cases as they generate CIL instructions, which I'm not sure would outperform reflection in an interpretation environment (I have tested this and I didn't see any noticeable differences, but felt safer not to change it).

### DIFF
--- a/osu.Framework/Audio/Callbacks/BassCallback.cs
+++ b/osu.Framework/Audio/Callbacks/BassCallback.cs
@@ -18,7 +18,7 @@ namespace osu.Framework.Audio.Callbacks
 
         protected BassCallback()
         {
-            if (!RuntimeFeature.IsDynamicCodeCompiled)
+            if (!RuntimeFeature.IsDynamicCodeSupported)
                 handle = new ObjectHandle<BassCallback>(this, GCHandleType.Normal);
         }
 
--- a/osu.Framework/Audio/Callbacks/FileCallbacks.cs
+++ b/osu.Framework/Audio/Callbacks/FileCallbacks.cs
@@ -41,14 +41,14 @@ namespace osu.Framework.Audio.Callbacks
 
         public FileCallbacks(IFileProcedures implementation)
         {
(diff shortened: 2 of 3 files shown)


smoogipoo:
Have you tested that this runs too?

peppy:
Let's save this for after the next framework release, as to not add any potential regressions.

frenzibyte:
> Have you tested that this runs too?

Yeah I've tested this just now on iOS, not sure about other platforms like Android though.

> Let's save this for after the next framework release, as to not add any potential regressions.

Yeah for sure.

bdach:
> not sure about other platforms like Android though

You know what, I tested and am not sure either. Especially so that the OP is so wonderfully vague about why this is being changed that it matches the [actual](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.runtimefeature.isdynamiccodecompiled?view=net-7.0) [xmldocs](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.runtimefeature.isdynamiccodesupported?view=net-7.0) of both methods in uselessness.

Anyways, in testing, this seems to change _something_ on android:

image

Is this good or bad? Who knows. Game runs, at least. Music plays, tracks loop, so maybe fine.

I'll also bring up that [this was already mentioned for whatever reason over half a year ago](https://discord.com/channels/188630481301012481/589331078574112768/1051803662315421727). Why? No idea.

frenzibyte:
See https://discord.com/channels/188630481301012481/589331078574112768/1127777353561542676, https://github.com/ManagedBass/ManagedBass/pull/118#discussion_r1257622779, and https://github.com/dotnet/runtime/issues/25959#issue-558370042.

frenzibyte:
What you have there shows that the game is running using interpretation.

smoogipoo (approved):

PR #5915, opened 2023-07-10
Remove most locking overhead in `TripleBuffer`

peppy:
One of those things I see in passing and can't un-see. Call this a developer indulgence rather than an optimisation if you want.

| Before | After |
|:------:|:-----:|
| JetBrains Rider 2021-09 at 07 04 10 | JetBrains Rider 2023-07-10 at 07 09 26 |
| JetBrains Rider 2021-09 at 07 03 40 | JetBrains Rider 2023-07-10 at 07 09 00 |

The time profiling results should be taken with a grain of salt. Of note is the reduction in lock %, mainly in `finishUsage`.

### DIFF
--- a/osu.Framework/Allocation/TripleBuffer.cs
+++ b/osu.Framework/Allocation/TripleBuffer.cs
@@ -85,22 +85,15 @@ namespace osu.Framework.Allocation
 
             lock (buffers)
             {
-                // re-fetch the original pendingCompletedWriteIndex inside the lock in case there's a newer index available.
-                if (pendingCompletedWriteIndex > -1)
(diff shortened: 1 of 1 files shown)


bdach:
Can't see any fault with this, theoretically speaking. The split critical section in `getPendingReadBuffer()` (the xchg first and the monitor block second) had me worried the most and even then I don't see how it can be abused to do anything transgressive.

Will probably ask for a second review though...

bdach (approved):

peppy:
(checked with smoogi and he was fine with getting this in)

PR #5917, opened 2023-07-10
Revert recent changes to UBO upload method for non-metal renderers for now

peppy:
See https://github.com/ppy/osu-framework/pull/5913#issuecomment-1628381235.

PR #5918, opened 2023-07-10
Use staging buffers for metal UBO uploads

peppy:
As proposed by @smoogipoo. In the vertex upload performance test scene, this performs ~50% better on macOS for me.

Making this PR to track a potential improvement.

- [ ] Depends on https://github.com/ppy/osu-framework/pull/5917

### DIFF
--- a/osu.Framework/Graphics/Veldrid/Buffers/VeldridUniformBufferStorage.cs
+++ b/osu.Framework/Graphics/Veldrid/Buffers/VeldridUniformBufferStorage.cs
@@ -33,6 +33,8 @@ namespace osu.Framework.Graphics.Veldrid.Buffers
             {
                 data = value;
 
+                // These pathways are faster on respective platforms.
+                // Test using TestSceneVertexUploadPerformance.
                 if (renderer.Device.BackendType == GraphicsBackend.Metal)
                 {
                     var staging = renderer.GetFreeStagingBuffer(buffer.SizeInBytes);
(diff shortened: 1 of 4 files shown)


osu.Framework/Graphics/Veldrid/Buffers/VeldridUniformBufferStorage.cs
```diff
@@ -18,28 +19,38 @@ internal class VeldridUniformBufferStorage<TData>
...
         {
             this.renderer = renderer;
 
-            buffer = renderer.Factory.CreateBuffer(new BufferDescription((uint)Marshal.SizeOf(default(TData)), BufferUsage.UniformBuffer));
+            // These pathways are faster on respective platforms.
+            // Test using TestSceneVertexUploadPerformance.
+            if (renderer.Device.BackendType == GraphicsBackend.Metal)
+            {
+                buffer = renderer.Factory.CreateBuffer(new BufferDescription((uint)Marshal.SizeOf(default(TData)), BufferUsage.UniformBuffer | BufferUsage.Dynamic));
+                stagingBuffer = renderer.CreateStagingBuffer<TData>(1);
```
smoogipoo, line 33: Invert these two and replace `| Dynamic` with `| stagingBuffer.CopyTargetUsageFlags`

smoogipoo (changes requested):

smoogipoo:
Personally, I'd prefer that the other pathway was fixed and the `CreateStagingBuffer` implementation was optimal for all backends, because it would have an effect on VBO uploads that don't have such a conditional.

peppy:
In theory, if this worked then yeah, it should be applied to all usages.

PR #5919, opened 2023-07-10
Improve `Clipboard` class for use in tests and on Android

Susko3:
As noted in https://github.com/ppy/osu/pull/24171, clipboard can't be used in headless tests. This PR adds a virtual clipboard that can be used in test, and improves old clipboard code along the way.

Changes:
- `Clipboard` is now a cached dependency (instead of acquiring it from `GameHost.GetClipboard()`) this brings it in line with osu-framework best practises(maybe?) and allows overriding it for tests if necessary
- `HeadlessGameHost` now uses a `HeadlessClipboard` that stores the data in-memory and doesn't modify the host OS clipboard
- `AndroidClipboard` has been added, text only for now to keep it simple. [docs](https://developer.android.com/develop/ui/views/touch-and-input/copy-paste)

### `GameHost.GetClipboard()` is obsolete

The new way to retrieve a `Clipboard` instance is to resolve it via dependency injection directly. This does not require resolving the `GameHost` anymore:

```diff
 [BackgroundDependencyLoader]
-private void load(GameHost host)
+private void load(Clipboard clipboard)
 {
-    host?.Clipboard.SetText("example");
+    clipboard.SetText("example");
 }
```

### DIFF
--- a/osu.Framework.Android/AndroidClipboard.cs
+++ b/osu.Framework.Android/AndroidClipboard.cs
@@ -1,0 +1,31 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using Android.Content;
+using osu.Framework.Platform;
+using SixLabors.ImageSharp;
+
+namespace osu.Framework.Android
+{
+    public class AndroidClipboard : Clipboard
+    {
+        private readonly ClipboardManager? clipboardManager;
+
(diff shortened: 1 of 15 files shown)


bdach:
Seems okay.

Android implementation tested to work. Mind you, right now to work 100% properly it requires an actual physical keyboard (without one, there's no way to access the typical android text edit menu with the copy option - paste may still work though if the software keyboard (ab)uses IME to implement paste). But it does work within those constraints.

peppy (approved):

PR #5921, opened 2023-07-11
Add a fence pool for frame tracking fences

peppy:
Ignoring the fact that these tracking fences are doing nothing in D3D (they are set immediately https://github.com/ppy/veldrid/blob/6a43198af1b069e56ac53770249a5424c1859500/src/Veldrid/D3D11/D3D11GraphicsDevice.cs#L229-L232):

This is a regression so let's fix it locally for the time being.

Time spent in `ManualResetEvent` construction:

Parallels Desktop 2023-07-11 at 05 01 20

Before this PR:

Parallels Desktop 2023-07-11 at 05 07 32

After this PR:

Parallels Desktop 2023-07-11 at 05 05 52

### DIFF
--- a/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
+++ b/osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
@@ -71,6 +71,11 @@ namespace osu.Framework.Graphics.Veldrid
         /// </summary>
         private readonly List<FrameCompletionFence> pendingFramesFences = new List<FrameCompletionFence>();
 
+        /// <summary>
+        /// We are using fences every frame. Construction can be expensive, so let's pool some.
+        /// </summary>
+        private readonly Queue<Fence> perFrameFencePool = new Queue<Fence>();
+
         public VeldridIndexData SharedLinearIndex { get; }
         public VeldridIndexData SharedQuadIndex { get; }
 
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
```diff
@@ -215,6 +220,9 @@ protected override void Initialise(IGraphicsSurface graphicsSurface)
             BufferUpdateCommands = Factory.CreateCommandList();
 
             pipeline.Outputs = Device.SwapchainFramebuffer.OutputDescription;
+
+            for (int i = 0; i < 16; i++)
```
smoogipoo, line 224: Is this 16 a magic number? What happens in the (unlikely) even that the pool runs dry, crash on `Dequeue()`?

smoogipoo (changes requested):

osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
peppy: Yep. On metal I've never seen usages increase above 1. If we exceed 3 then we're probably going to see texture corruption and stuff elsewhere though? I can use 3 instead of 16 if it feels more robust.

peppy: Or we can change it to add to the pool locally, should also be fine.

peppy:
I think we still want this in, at least for now. Can potentially be reverted if we decide we don't want end-of-frame fencing on all platforms.

smoogipoo (approved):

PR #5922, opened 2023-07-11
Split texture upload performance test into own `TestScene`

peppy:
This also makes the uploads happen every frame, as this is more useful for diagnosing contention edge cases.

Notes from testing:

- Metal can show signs of corruption with higher texture upload counts.
- D3D has very low performance compared to metal with similar texture upload counts.

### DIFF
--- a/osu.Framework.Tests/Visual/Performance/TestSceneTextureUploadPerformance.cs
+++ b/osu.Framework.Tests/Visual/Performance/TestSceneTextureUploadPerformance.cs
@@ -1,0 +1,116 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using osu.Framework.Allocation;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Primitives;
+using osu.Framework.Graphics.Rendering;
+using osu.Framework.Graphics.Sprites;
+using osu.Framework.Graphics.Textures;
+using osu.Framework.IO.Stores;
+using osu.Framework.Platform;
+using osuTK;
+using SixLabors.ImageSharp.PixelFormats;
+using PixelFormat = osuTK.Graphics.ES30.PixelFormat;
+
+namespace osu.Framework.Tests.Visual.Performance
+{
+    public partial class TestSceneTextureUploadPerformance : PerformanceTestScene
+    {
+        private int count;
+
+        private FillFlowContainer<Sprite>? fill;
+
+        [Resolved]
+        private IRenderer renderer { get; set; } = null!;
+
+        private PersistentTextureUpload sampleTextureUpload = null!;
+
+        [BackgroundDependencyLoader]
+        private void load(Game game, GameHost host)
+        {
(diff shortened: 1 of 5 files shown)


osu.Framework.Tests/Visual/Performance/TestSceneTextureUploadPerformance.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            AddSliderStep("count", 1, 100, 10, v =>
+            {
+                count = v;
+                recreate();
+            });
```
frenzibyte, line 47: Right now changing the value of the slider recreates the fill flow and all sprites, causing a flicker on every instant value when "sliding" the slider.

Creating the flow once and adding/removing sprites based on the count number (similar to `TestSceneBoxPerformance`) would feel better:
```diff
diff --git a/osu.Framework.Tests/Visual/Performance/TestSceneTextureUploadPerformance.cs b/osu.Framework.Tests/Visual/Performance/TestSceneTextureUploadPerformance.cs
index 6c077fc93..a0933b408 100644
--- a/osu.Framework.Tests/Visual/Performance/TestSceneTextureUploadPerformance.cs
+++ b/osu.Framework.Tests/Visual/Performance/TestSceneTextureUploadPerformance.cs
@@ -19,8 +19,6 @@ namespace osu.Framework.Tests.Visual.Performance
 {
     public partial class TestSceneTextureUploadPerformance : PerformanceTestScene
     {
-        private int count;
-
         private FillFlowContainer<Sprite>? fill;
 
         [Resolved]
@@ -40,15 +38,6 @@ protected override void LoadComplete()
         {
             base.LoadComplete();
 
-            AddSliderStep("count", 1, 100, 10, v =>
-            {
-                count = v;
-                recreate();
-            });
-        }
-
-        private void recreate()
-        {
             Child = fill = new FillFlowContainer<Sprite>
             {
                 RelativeSizeAxes = Axes.X,
@@ -58,16 +47,22 @@ private void recreate()
                 Origin = Anchor.Centre,
             };
 
-            for (int i = 0; i < count; i++)
+            AddSliderStep("count", 1, 100, 10, newCount =>
             {
-                fill.Add(new Sprite
-                {
-                    Size = new Vector2(128),
-                    Texture = renderer.CreateTexture(512, 512)
-                });
-            }
+                for (int i = fill.Count - 1; i >= newCount; i--)
+                    fill.Remove(fill.Children[i], true);
+
+                for (int i = fill.Count; i < newCount; i++)
+                    fill.Add(createSprite());
+            });
         }
 
+        private Sprite createSprite() => new Sprite
+        {
+            Size = new Vector2(64),
+            Texture = renderer.CreateTexture(512, 512)
+        };
+
         private ulong lastUploadedFrame;
 
         protected override void Update()

```

peppy: This is the same as every other performance test scene though, right?

The plan is to go through and fix all of these in one go.

peppy: Feel free to commit the change, just note that it needs to be fixed everywhere else.

frenzibyte (approved):

PR #5925, opened 2023-07-12
Fix screenshots not working on Metal

frenzibyte:
- Regressed in https://github.com/ppy/veldrid/pull/30

Initially I was looking into properly implementing a "wait for fence" logic over at Veldrid using [`notifyListener:atValue:block`](https://developer.apple.com/documentation/metal/mtlsharedevent/2966574-notifylistener?changes=_8&language=objc), but I've stumbled across issues with P/Invokes specific to that function, and tried going back and fourth on it for over an hour to no avail. For the time being, I'm working around this locally.


peppy (approved):
Maybe okay.

smoogipoo (approved):

PR #5929, opened 2023-07-13
Revert veldrid update

peppy:
- Reverts https://github.com/ppy/veldrid/pull/32
- Reverts https://github.com/ppy/veldrid/pull/31

This reverts commit 02626fd633a21e68cdae78c528b58a06b281ceae.


bdach:
OP is confusing, I'm pretty sure it should say that it's reverting https://github.com/ppy/veldrid/pull/30 and https://github.com/ppy/veldrid/pull/31? (it currently says it's reverting https://github.com/ppy/veldrid/pull/32, which is a revert, so :exploding_head:)

Other than that, looks good in theory. Someone with a Mac should probably confirm in practice, though.

peppy:
Yeah, misread the PR title. Fixed now.

bdach (dismissed):

frenzibyte (approved):

PR #5930, opened 2023-07-14
Revert vertex batch buffer changes

peppy:
Unsafe for reasons mentioned in inline comment


frenzibyte (approved):
Performance remains good for Metal on my machine.

osu.Framework/Graphics/Veldrid/Batches/VeldridVertexBatch.cs
```diff
@@ -17,13 +17,13 @@ internal abstract class VeldridVertexBatch<T> : IVertexBatch<T>
...
+        /// We already have a triple buffer (see <see cref="GameHost.DrawRoots"/>) governing draw nodes.
+        /// In theory we could set this to two, but there's also a global usage of a vertex batch in <see cref="VeldridRenderer"/> (see <see cref="VeldridRenderer.DefaultQuadBatch"/>).
         ///
-        /// Note that due to the higher level triple buffer, the actual number of buffers we are storing is three times as high as this constant.
-        /// Maintaining this many buffers is a cause of concern from an memory alloc / GPU upload perspective.
+        /// So this is for now an unfortunate memory overhead. Further work could be done to provide
+        /// these in a way they were not created per draw-node, reducing buffer overhead from 9 to 3.
         /// </summary>
-        private const int vertex_buffer_count = 2;
+        private const int vertex_buffer_count = 3;
```
smoogipoo, line 26: Do we want to change this back to 2?

peppy: This is actually required, i tested each change individually (and just re-checked now).

PR #5933, opened 2023-07-18
Add support for Unicode Surrogates (including emoji) in `SpriteText` and `TextBuilder`

Susko3:
`char` cannot represent [Unicode scalar values](https://unicode.org/glossary/#unicode_scalar_value) greater then `0xFFFF` (aka [surrogates](https://www.unicode.org/versions/Unicode15.0.0/ch03.pdf#G2630)) which are used by emoji and various scripts (eg. some CJK characters). Fixing this is just a matter of decoding the strings into [`System.Text.Rune`](https://learn.microsoft.com/en-us/dotnet/api/system.text.rune?view=net-7.0) (a wrapper around `int`) and passing the values to `SharpFNT`. `SharpFNT` and `BMFont` already support scalar values over `0xFFFF`.

The changes are mostly changing `char` to `Rune` and accessing `string`s with [`Rune.GetRuneAt()`](https://learn.microsoft.com/en-us/dotnet/api/system.text.rune.getruneat?view=net-7.0) and [`string.EnumerateRunes()`](https://learn.microsoft.com/en-us/dotnet/api/system.string.enumeraterunes?view=net-7.0) instead of by-character.

I've not changed every relevant `char` usage to `Rune`, as it would result in more verbose code. More details in self-review.

I've added the monochrome [Noto Emoji](https://fonts.google.com/noto/specimen/Noto+Emoji) font to test resources to see the change in action. I may have gone a bit overboard with selecting emoji, as the generated font uses 3.28 MB. I've tested Noto Color Emoji and BMFont doesn't seem to support that.
Note that I've included [Variation Selectors 15 and 16](https://en.wikipedia.org/wiki/Variation_Selectors_(Unicode_block)) in the font so they don't show as `?` (they render as nothing instead).

# What doesn't work

## `TextBox`

Requires more changes to support this, as it needs additional consideration for how text selection, deleting, etc. should work. Best done in a separate PR.

## Sequences of more than one character

Such as ZWJ emoji 😶‍🌫️ (including colour specifiers), flags 🇭🇷, tag sequences 🏴󠁧󠁢󠁥󠁮󠁧󠁿, and other characters that are actually multiple code points (combining characters and whatnot).

.NET has support for this with [`StringInfo`](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.stringinfo.gettextelementenumerator?view=net-7.0) (probably much worse performance than enumerating runes) but I don't think BMFont and SharpFNT support it.


osu.Framework/Graphics/Sprites/SpriteText.cs
```diff
@@ -554,7 +556,7 @@ private void invalidate(bool characters = false, bool textBuilder = false)
         /// <returns>The <see cref="TextBuilder"/>.</returns>
         protected virtual TextBuilder CreateTextBuilder(ITexturedGlyphLookupStore store)
         {
-            char[] excludeCharacters = FixedWidthExcludeCharacters ?? default_never_fixed_width_characters;
+            Rune[] excludeCharacters = (FixedWidthExcludeCharacters ?? default_never_fixed_width_characters).Select(c => new Rune(c)).ToArray();
```
Susko3, line 559: This is probably not that great for performance. The alternative is to make everything `Rune[]` and force consumers to cast to `Rune`. Same goes for `FixedWidthReferenceCharacter` and `FallbackCharacter`, which currently remain as `char`.

```diff
diff --git a/osu.Framework/Graphics/Sprites/SpriteText.cs b/osu.Framework/Graphics/Sprites/SpriteText.cs
index 04bebf0fc..863ec3c46 100644
--- a/osu.Framework/Graphics/Sprites/SpriteText.cs
+++ b/osu.Framework/Graphics/Sprites/SpriteText.cs
@@ -33,7 +33,7 @@ public partial class SpriteText : Drawable, IHasLineBaseHeight, ITexturedShaderD
         /// <c>U+00A0</c> is the Unicode NON-BREAKING SPACE character (distinct from the standard ASCII space).
         /// <c>U+202F</c> is the Unicode NARROW NO-BREAK SPACE character.
         /// </remarks>
-        private static readonly char[] default_never_fixed_width_characters = { '.', ',', ':', ' ', '\u00A0', '\u202F' };
+        private static readonly Rune[] default_never_fixed_width_characters = new[] { '.', ',', ':', ' ', '\u00A0', '\u202F' }.Select(c => new Rune(c)).ToArray();
 
         [Resolved]
         private FontStore store { get; set; }
```

osu.Framework/IO/Stores/GlyphStore.cs
```diff
@@ -130,31 +131,31 @@ public CharacterGlyph Get(char character)
...
         public TextureUpload Get(string name)
         {
             if (Font == null) return null;
 
-            if (name.Length > 1 && !name.StartsWith($@"{FontName}/", StringComparison.Ordinal))
+            if (name.EnumerateRunes().Count() > 1 && !name.StartsWith($@"{FontName}/", StringComparison.Ordinal))
                 return null;
 
-            return Font.Characters.TryGetValue(name.Last(), out Character c) ? LoadCharacter(c) : null;
+            return Font.Characters.TryGetValue(name.EnumerateRunes().Last().Value, out Character c) ? LoadCharacter(c) : null;
```
Susko3, line 148: Could be optimised to not call `EnumerateRunes()` twice. Not really sure.

But I also think this logic is really weird, mangling together the font name and character. See also https://github.com/ppy/osu-framework/blob/809808745f8c2b4bed745d280ecef172c9438a8d/osu.Framework/IO/Stores/FontStore.cs#L135-L158

osu.Framework.Tests/TestGame.cs
```diff
@@ -16,6 +16,7 @@ internal partial class TestGame : Game
         private void load()
         {
             Resources.AddStore(new NamespacedResourceStore<byte[]>(new DllResourceStore(typeof(TestGame).Assembly), "Resources"));
+            AddFont(Resources, @"Fonts/NotoEmoji/NotoEmoji");
```
peppy, line 19: Very weird that this is only added in the test game, but added at such a high level.

Susko3: > Very weird that this is only added in the test game

I've added the fonts only to `osu.Framework.Tests` project. It is only added as a PoC and to verify that my changes work visually. This is to keep the the framework dll size small, and matches the current font situation in framework and osu!: framework has basic fonts to support its functionality, while the game adds any fonts it needs, such as large file size CJK (and now emoji) fonts.

> but added at such a high level.

What would be a better place for it? I would want to have it available at least in `SpriteText` and `TextBox` test scenes.

Susko3: Actually, the framework does use emoji: when logging things in tests. Adding only those could be beneficial for viewing the logs in-game. Unfortunately the emoji are monochrome, so they don't stand out as much, but it's way better than seeing two "??".

I'll add only the emoji used in logging to keep the size small (and squash the commits to keep the history small).

Hopefully having two fonts with the same name won't cause issues (if we add `NotoEmoji` to osu!).

frenzibyte: And then start updating the framework-side emoji attachments every time we use a new emoji for logging? I don't know about that, I would say either it exists as a whole or not exist at all in the main framework project.

PR #5935, opened 2023-07-19
Add option to avoid automatically running the first step in `TestScene`s

Susko3:
Since it was quite common to have a "do nothing" step, this makes sense as an option.

### DIFF
--- a/osu.Framework/Testing/TestScene.cs
+++ b/osu.Framework/Testing/TestScene.cs
@@ -52,7 +52,7 @@ namespace osu.Framework.Testing
         /// <remarks>
         /// Defaults to <c>true</c>. Should be set to <c>false</c> if the first step in the first <see cref="TestAttribute">test</see> has unwanted-by-default behaviour.
         /// </remarks>
-        public virtual bool RunFirstActualStepOnLoad => true;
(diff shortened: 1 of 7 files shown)


osu.Framework/Testing/TestScene.cs
```diff
@@ -40,6 +40,20 @@ public abstract partial class TestScene : Container
...
+        protected virtual double TimePerAction => 200;
+
+        /// <summary>
+        /// Whether to automatically run the the first actual <see cref="StepButton"/> (one that is not part of <see cref="SetUpAttribute">[SetUp]</see> or <see cref="SetUpStepsAttribute">[SetUpSteps]</see>)
+        /// when the test is first loaded.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to <c>true</c>. Should be set to <c>false</c> if the first step in the first <see cref="TestAttribute">test</see> has unwanted-by-default behaviour.
+        /// </remarks>
+        public virtual bool RunFirstActualStepOnLoad => true;
```
frenzibyte, line 55: Name feels too mouthful, I would just call it `RunFirstStepAutomatically` and keep the technical details in the xmldoc.

peppy (approved):

PR #5937, opened 2023-07-19
Fix window size not being stored to config during resize

### LINKED ISSUE #5936: Window size not stored to config during resize
- Regressed with https://github.com/ppy/osu-framework/pull/4510

Reported [on discord](https://discord.com/channels/188630481301012481/188630652340404224/1131144941985345596). During a resize in multithreaded mode, the window size wouldn't get stored to the config. This causes regression in osu!, not showing the window size toast during resize. The size would get properly stored once the resize finished.

## Possible fix

Use [SDL_AddEventWatch](https://wiki.libsdl.org/SDL2/SDL_AddEventWatch) instead of [SDL_SetEventFilter](https://wiki.libsdl.org/SDL2/SDL_SetEventFilter) for querying resize events outside the event loop.

The event filter is called before the internal SDL state is updated, but the watch is called after, which should be perfect for our use case.

Susko3:
- Closes #5936

- [x] Depends on #5935
   (to make the test not disable resizing on startup)

Uses [`SDL_AddEventWatch`](https://wiki.libsdl.org/SDL2/SDL_AddEventWatch) as mentioned in the issue and removes the scheduling of `storeWindowSizeToConfig()`. We have recently had [another issue](https://github.com/ppy/osu-framework/issues/5879) caused by invalid window state when using the event filter, so moving it to the event watch is a welcome change.

I plan to use `FrameworkConfigVisualiser` in more windowing tests going forward.


osu.Framework/Platform/SDL2Window_Windowing.cs
```diff
@@ -442,9 +442,7 @@ private void fetchWindowSize()
             Scale = (float)drawableW / w;
             Size = new Size(w, h);
 
-            // This function may be invoked before the SDL internal states are all changed. (as documented here: https://wiki.libsdl.org/SDL_SetEventFilter)
-            // Scheduling the store to config until after the event poll has run will ensure the window is in the correct state.
-            EventScheduler.AddOnce(storeWindowSizeToConfig);
```
peppy, line 447: Are you sure about removing the `AddOnce` part of this?

frenzibyte:
Need to check against iOS with AOT build as this touches against unmanaged callbacks.

osu.Framework/Platform/SDL2Window_Windowing.cs
Susko3: Well if you want the config window size to change on every resize, scheduling it only once is counter-productive. Besides, the schedule is what caused the bug, as it would run only after the event loop finished -- and resizing blocks the event loop.

peppy: Hmm, I see. That scheduler is a bit misleading.

frenzibyte (approved):
Looks good. Can confirm it's working correctly on macOS, and also tested to confirm iOS works with the new "watch" filter in AOT mode.

peppy (approved):

PR #5938, opened 2023-07-20
Change source generators to only run for release builds

peppy:
### Source generators will now only run on release builds

As we continue to add more source generators, we've seen increases in compile-time overheads, with local testing showing over 2x compile times with source generators turned on.

osu!framework source generators are made to optimise builds at runtime (mostly by removing reflection overhead). As such, it doesn't make sense to run these for debug releases as they are basically custom release-targeting optimisations.

This should not result in a noticeable change in runtime performance during debug, but will reduce compilation times by over 50% in most cases. This is valuable during debug as the most common case is frequently building / hot reloading for quick iteration.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,9 +58,10 @@ jobs:
        fail-fast: false
        matrix:
           os:
-            - { prettyname: Windows, fullname: windows-latest }
-            - { prettyname: macOS, fullname: macos-latest }
-            - { prettyname: Linux, fullname: ubuntu-latest }
+            - { prettyname: Windows, fullname: windows-latest, configuration: Debug }
+            - { prettyname: macOS, fullname: macos-latest, configuration: Debug }
+            - { prettyname: Linux, fullname: ubuntu-latest, configuration: Debug }
+            - { prettyname: Linux (Release), fullname: ubuntu-latest, configuration: Release }
           threadingMode: ['SingleThread', 'MultiThreaded']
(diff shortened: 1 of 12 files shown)


.github/workflows/ci.yml
```diff
@@ -84,10 +85,10 @@ jobs:
         shell: bash
 
       - name: Compile
-        run: dotnet build -c Debug -warnaserror osu-framework.Desktop.slnf
+        run: dotnet build -c ${{matrix.os.configuration}} -warnaserror osu-framework.Desktop.slnf
 
       - name: Test
-        run: dotnet test $pwd/**/*.Tests/bin/Debug/*/*.Tests.dll --settings $pwd/build/vstestconfig.runsettings --logger "trx;LogFileName=TestResults-${{matrix.os.prettyname}}-${{matrix.threadingMode}}.trx"
+        run: dotnet test $pwd/**/*.Tests/bin/Debug/*/*.Tests.dll --no-build --settings $pwd/build/vstestconfig.runsettings --logger "trx;LogFileName=TestResults-${{matrix.os.prettyname}}-${{matrix.threadingMode}}-${{matrix.os.configuration}}.trx"
```
smoogipoo, line 91: $pwd/\*\*/\*.Tests/bin/**Debug**/\*/\*.Tests.dll

here is going to be an issue

peppy: hmm right

smoogipoo (approved):

PR #5941, opened 2023-07-23
Fix custom interpolation functions potentially causing SIGSEGV on iOS

frenzibyte:
- Works around https://github.com/dotnet/runtime/issues/89359
- Required for enabling AOT on osu!.

### DIFF
--- a/osu.Framework/Utils/Interpolation.cs
+++ b/osu.Framework/Utils/Interpolation.cs
@@ -2,7 +2,6 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Runtime.Serialization;
 using osu.Framework.Extensions.Color4Extensions;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Colour;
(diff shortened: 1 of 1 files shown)


peppy (approved):

smoogipoo:
As a note, this will allocate every call now. Testable with:

```
[MemoryDiagnoser]
public class StructInvokeThroughBox
{
    [Benchmark]
    public int ByIs()
    {
        int result = 0;

        for (int i = 0; i < 1_000_000; i++)
            result += getByIs(new MyStruct());

        return result;
    }

    [Benchmark]
    public int ByCast()
    {
        int result = 0;

        for (int i = 0; i < 1_000_000; i++)
            result += getByCast(new MyStruct());

        return result;
    }

    private int getByIs<T>(T val)
    {
        if (val is IMyInterface intfT)
            return intfT.Get();

        return 0;
    }

    private int getByCast<T>(T val)
    {
        if (typeof(T) == typeof(IMyInterface))
            return ((IMyInterface)val).Get();

        return 0;
    }

    private struct MyStruct : IMyInterface
    {
        public int Get() => 1;
    }

    private interface IMyInterface
    {
        int Get();
    }
}
```

frenzibyte:
Turns out the underlying issue has already been reported before, but it's only resolved in a future preview / final version of .NET 8

PR #5945, opened 2023-07-26
Add ability to `AddSetupStep`

peppy:
Useful in tests which define setup steps inline (where multiple test methods have different initialisation requirements)

### DIFF
--- a/osu.Framework/Testing/TestScene.cs
+++ b/osu.Framework/Testing/TestScene.cs
@@ -255,9 +255,15 @@ namespace osu.Framework.Testing
 
         private bool addStepsAsSetupSteps;
 
-        public StepButton AddStep(string description, Action action)
+        public StepButton AddSetupStep(string description, Action action) =>
+            addStep(description, action, true);
+
+        public StepButton AddStep(string description, Action action) =>
+            addStep(description, action, addStepsAsSetupSteps);
+
+        private StepButton addStep(string description, Action action, bool isSetupStep)
         {
-            var step = new SingleStepButton(addStepsAsSetupSteps)
+            var step = new SingleStepButton(isSetupStep)
             {
                 Text = description,
                 Action = action


Susko3:
Having `[SetUp]`, `[SetUpSteps]` and now `AddSetupStep()`, I think it's time to reconsider the default value (and existence of) [`AutomaticallyRunFirstStep`](https://github.com/ppy/osu-framework/blob/ad29059480e1dc7654074fd4913ad35ffef98259/osu.Framework/Testing/TestScene.cs#L48-L55).

It would be best if `AutomaticallyRunFirstStep` didn't even exist (only set up steps run automatically). But some older test scenes rely on it for ease of use. 

I think it would cause the least breakage if `AutomaticallyRunFirstStep` defaults to true only if no set up steps have been added (see diff). Any test scenes that break from this change could then be updated to properly use set up steps via the attribute or the newly-added method.

```diff
diff --git a/osu.Framework/Testing/TestScene.cs b/osu.Framework/Testing/TestScene.cs
index f4fd47f51..646442ae2 100644
--- a/osu.Framework/Testing/TestScene.cs
+++ b/osu.Framework/Testing/TestScene.cs
@@ -45,6 +45,8 @@ public abstract partial class TestScene : Container
         /// </summary>
         protected virtual double TimePerAction => 200;
 
+        private bool automaticallyRunFirstStep = true;
+
         /// <summary>
         /// Whether to automatically run the the first actual <see cref="StepButton"/> (one that is not part of <see cref="SetUpAttribute">[SetUp]</see> or <see cref="SetUpStepsAttribute">[SetUpSteps]</see>)
         /// when the test is first loaded.
@@ -52,7 +54,7 @@ public abstract partial class TestScene : Container
         /// <remarks>
         /// Defaults to <c>true</c>. Should be set to <c>false</c> if the first step in the first <see cref="TestAttribute">test</see> has unwanted-by-default behaviour.
         /// </remarks>
-        public virtual bool AutomaticallyRunFirstStep => true;
+        public virtual bool AutomaticallyRunFirstStep => automaticallyRunFirstStep;
 
         private GameHost host;
         private Task runTask;
@@ -263,6 +265,9 @@ private void runNextStep(Action onCompletion, Action<Exception> onError, Func<St
 
         private StepButton addStep(string description, Action action, bool isSetupStep)
         {
+            if (isSetupStep)
+                automaticallyRunFirstStep = false;
+
             var step = new SingleStepButton(isSetupStep)
             {
                 Text = description,
```

peppy:
Doesn't seem we need this after all. Haven't felt the necessity recently.

PR #5950, opened 2023-07-28
Add support for shader storage buffer objects (SSBOs)

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/veldrid/pull/33

SSBOs are a way to submit large amounts of data to GPUs.

While SSBOs are pretty well supported by modern GPUs[^1], this PR encapsulates two modes of operation under the "SSBO" umbrella - one which implements SSBOs directly, and a fallback mode which utilises UBOs. This encapsulation is provided by `IRenderer.CreateShaderStorageBufferObject<T>(uboSize, ssboSize)`.

UBOs are generally limited to at least 16KiB[^2][^3], though modern GL implementations frequently provide a larger size (usually up to 64KiB). SSBOs can generally be up to the size of available VRAM.

For the time being, this implementation provides no validation as to the size of the buffers, so it is left up to users to specify "safe" values as documented by the `CreateShaderStorageBufferObject` method. This may change with future work.

In general, however, users should instead use a secondary implementation provided in this PR - `ShaderStorageBufferObjectStack<T>`. This automates the heavy lifting process of binding multiple UBOs when the data exceeds the 16KiB limit, or when the total amount of data isn't known ahead of time.

Using either these implementations is somewhat complicated as the shader needs to implement both pathways, however a sample implementation of both is provided in `TestSceneShaderStorageBufferObject`, `sh_SSBOTest.vs`, and `sh_SSBOTest.fs`. This is deserving of its own wiki page.

The `OSU_GRAPHICS_NO_SSBO` environment variable may be set to `1` to globally disable SSBOs.

[^1]: [80%](https://feedback.wildfiregames.com/gl/extensions/) of [these GPUs](https://feedback.wildfiregames.com/gpu/) support `GL_ARB_shader_storage_buffer_object`.
[^2]: OpenGL: [`MAX_UNIFORM_BLOCK_SIZE`](https://registry.khronos.org/OpenGL/extensions/ARB/ARB_uniform_buffer_object.txt)
[^3]: D3D11: [`Each constant buffer can hold up to 4096 vectors; each vector contains up to four 32-bit values.`](https://learn.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-constants#remarks)


osu.Framework.Tests/Graphics/ShaderStorageBufferObjectStackTest.cs
```diff
@@ -0,0 +1,258 @@
...
+            Assert.That(stack.CurrentOffset, Is.EqualTo(4));
+            Assert.That(stack.CurrentBuffer[4], Is.EqualTo(size + 2));
+
+            // ... where this is the copied item as a result of the immediate push...
+            Assert.That(stack.CurrentBuffer[3], Is.EqualTo(copiedItem2));
+
+            // ... and these are the same items from the first pushes above.
+            Assert.That(stack.CurrentBuffer[2], Is.EqualTo(size + 1));
+            Assert.That(stack.CurrentBuffer[1], Is.EqualTo(size));
+            Assert.That(stack.CurrentBuffer[0], Is.EqualTo(copiedItem1));
```
bdach, line 220: Wait, I don't get this. These items were popped, so why are they still in the buffer in any way? I'd expect them to get overwritten?

bdach:
The veldrid prereq is merged, but it doesn't look like the veldrid bump made it here yet?

osu.Framework.Tests/Graphics/ShaderStorageBufferObjectStackTest.cs
smoogipoo: SSBOStack never overwrites data because it doesn't know if a draw call has been launched for the previous data. I'm also unsure whether it would be more expensive to overwrite said data while it's in use by the GPU, in this case.

bdach: Okay I guess I somehow didn't read xmldoc so that's my bad.

I do feel like the current xmldoc is leaving things half-stated, because if I'm understanding correctly, you would generally call `.Clear()` at the start of every draw call to reclaim all of the buffer space and prevent runaway VRAM usage, right? I'd spell that out explicitly there.

smoogipoo: Yeah that sounds about right... I do wonder if this is yet another one of those things that should be "automagically" handled for you like every other instance of `ResetCounters()` or whatever (VBOs/UBOs...).

osu.Framework/Graphics/Rendering/ShaderStorageBufferObjectStack.cs
```diff
@@ -0,0 +1,191 @@
...
+                //
+                // When transitioning to a new buffer, we want to minimise a certain "thrashing" effect that occurs with successive push/pops.
+                // For example, suppose the sequence: push -> draw -> pop -> push -> draw -> pop -> etc...
+                // Each push transitions to buffer X+1 and each pop transitions back to buffer X, resulting in many pipeline flushes.
+                //
+                // This is a very specific use case that arises when several consumers push items anonymously from one other.
+                //
+                // A little hack is employed to alleviate this issue for ONE push-pop sequence:
+                // When transitioning to a new buffer, we copy the last item from the last buffer into the new buffer,
+                // and adjust the stack so that we no longer refer to a position inside the last buffer upon a pop.
```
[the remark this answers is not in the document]
smoogipoo: I think it depends on the usage characteristics... I think you might've written this while still thinking that the individual buffers don't have to retain all the data (https://github.com/ppy/osu-framework/pull/5950#discussion_r1293502961), in which case I'd agree that splitting pages into half could lead to better usage characteristics in the case of e.g. double push/double pop, with effectively no extra cost.

But, since we have to retain all previous data in previous buffers, then I think there is very little advantage of doing this. It'll turn from a split into a simple copy of half the elements, reducing the effective size of each buffer by half when it could be expensive to do so - the masking buffer is only 60 elements as is.

osu.Framework/Graphics/OpenGL/Buffers/GLShaderStorageBufferObject.cs
```diff
@@ -0,0 +1,96 @@
...
+                GL.BindBuffer(BufferTarget.UniformBuffer, Id);
+                GL.BufferSubData(BufferTarget.UniformBuffer, (IntPtr)(changeBeginIndex * elementSize), (IntPtr)(elementSize * changeCount), ref data[changeBeginIndex]);
+                GL.BindBuffer(BufferTarget.UniformBuffer, 0);
+            }
+            else
+            {
+                GL.BindBuffer(BufferTarget.UniformBuffer, Id);
+                GL.BufferSubData(BufferTarget.UniformBuffer, (IntPtr)(changeBeginIndex * elementSize), (IntPtr)(elementSize * changeCount), ref data[changeBeginIndex]);
+                GL.BindBuffer(BufferTarget.UniformBuffer, 0);
+            }
```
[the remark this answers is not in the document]
smoogipoo: I think there might've been a period of time where these were different, or were intended to be. But that's unnecessary. Will remove.

osu.Framework.Tests/Visual/Graphics/TestSceneShaderStorageBufferObject.cs
```diff
@@ -0,0 +1,278 @@
...
+
+                // Create the SSBO.
+                colourBuffer ??= new ShaderStorageBufferObjectStack<ColourData>(renderer, ubo_size, ssbo_size);
+
+                var rng = new Random(1337);
+
+                // Bind the custom shader.
+                shader.Bind();
+
+                // Submit vertices, making sure that we don't submit an index which would overflow the SSBO.
```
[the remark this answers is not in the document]
smoogipoo: Nope, you're right it's irrelevant here.

osu.Framework.Tests/Graphics/ShaderStorageBufferObjectStackTest.cs
smoogipoo: I've also added some XMLDoc to `.Clear()`, but if you have any better suggestions then I'm all ears. Proper documentation without requiring developers to have white-box knowledge about everything, is still one of the most difficult things for me.

peppy:
> The veldrid prereq is merged, but it doesn't look like the veldrid bump made it here yet?

Probably need to cherry-pick https://github.com/ppy/osu-framework/pull/5952/commits/d71f9c7872b7e2b5ed93f9d4a1540dd81eb31f33 across to here?

osu.Framework/Graphics/Rendering/ShaderStorageBufferObjectStack.cs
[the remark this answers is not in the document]
bdach: Right, fair call. Was just mostly musing anyways, but you're correct that I wasn't understanding the nuances correctly at the point of writing the original comment.

bdach:
Uh... possible bad news. On my box, on windows, upon opening the SSBO test scene, I get this:

```
Exception Info: System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception.
   at Vortice.Direct3D11.ID3D11DeviceContext.DrawIndexed(Int32 indexCount, Int32 startIndexLocation, Int32 baseVertexLocation)
   at osu.Framework.Graphics.Veldrid.VeldridRenderer.DrawVertices(PrimitiveTopology type, Int32 vertexStart, Int32 verticesCount) in D:\Open Source\osu-framework\osu.Framework\Graphics\Veldrid\VeldridRenderer.cs:line 601
   at osu.Framework.Graphics.Veldrid.Batches.VeldridVertexBatch`1.Draw() in D:\Open Source\osu-framework\osu.Framework\Graphics\Veldrid\Batches\VeldridVertexBatch.cs:line 161
   at osu.Framework.Graphics.Veldrid.Batches.VeldridVertexBatch`1.Add(T v) in D:\Open Source\osu-framework\osu.Framework\Graphics\Veldrid\Batches\VeldridVertexBatch.cs:line 116
   at osu.Framework.Tests.Visual.Graphics.TestSceneShaderStorageBufferObject.RawStorageBufferDrawNode.Draw(IRenderer renderer) in D:\Open Source\osu-framework\osu.Framework.Tests\Visual\Graphics\TestSceneShaderStorageBufferObject.cs:line 129
   at osu.Framework.Graphics.Containers.CompositeDrawable.CompositeDrawableDrawNode.Draw(IRenderer renderer) in D:\Open Source\osu-framework\osu.Framework\Graphics\Containers\CompositeDrawable_DrawNode.cs:line 204
```

Wonderfully unspecific, of course, looks to be a generic access violation. Not sure how do I even find out what this is exactly.

bdach (approved):

Issue #5953, opened 2023-07-30
`BindableList` mutation operations are unsafe with certain binding patterns


bdach:
As described [on discord](https://discord.com/channels/188630481301012481/188630652340404224/1135289640136495174).

Reproduction test case:

```diff
diff --git a/osu.Framework.Tests/Bindables/BindableListTest.cs b/osu.Framework.Tests/Bindables/BindableListTest.cs
index 9df3106dd..202d8eef0 100644
--- a/osu.Framework.Tests/Bindables/BindableListTest.cs
+++ b/osu.Framework.Tests/Bindables/BindableListTest.cs
@@ -995,6 +995,21 @@ public void TestRemoveAtNotifiesBoundLists()
             Assert.That(triggeredArgs.OldStartingIndex, Is.EqualTo(0));
         }
 
+        [Test]
+        public void TestRemoveAtBranchingBinds()
+        {
+            var b1 = new BindableList<int> { 1, 2, 3, 4, 5 };
+
+            var b2 = b1.GetBoundCopy();
+            var b3 = b1.GetBoundCopy();
+
+            var b4 = new BindableList<int>();
+            b2.BindTo(b4);
+            b3.BindTo(b4);
+
+            b1.RemoveAt(b1.Count - 1);
+        }
+
         #endregion
 
         #region .RemoveAll(match)

```

will fail with

```
System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')
   at System.Collections.Generic.List`1.get_Item(Int32 index)
   at osu.Framework.Bindables.BindableList`1.removeAt(Int32 index, BindableList`1 caller) in /home/dachb/Documents/opensource/osu-framework/osu.Framework/Bindables/BindableList.cs:line 274
   at osu.Framework.Bindables.BindableList`1.RemoveAt(Int32 index) in /home/dachb/Documents/opensource/osu-framework/osu.Framework/Bindables/BindableList.cs:line 268
   at osu.Framework.Tests.Bindables.BindableListTest.TestRemoveAtBranchingBinds() in /home/dachb/Documents/opensource/osu-framework/osu.Framework.Tests/Bindables/BindableListTest.cs:line 1010
```

This can also happen for any remove/add operation, not only a `RemoveAt()` issue specifically (although that one is loudest, as it will lead to a crash, adds will lead to silent corruption). Core issue is that the following binding weak reference graph is allowed:

```mermaid
flowchart TD
[...]

PR #5954, opened 2023-07-31
Fix `BindableList` being unsafe for complex bind trees

peppy:
- Closes https://github.com/ppy/osu-framework/issues/5953 (using solution 1).
- *Probably* also covers https://github.com/ppy/osu-framework/issues/5635 (see call stack, matches original investigation). I'm not listing this as closing the issue, because the threading concern is likely still real, but the specific stack in that issue suggests it will cover that case.
- Closes https://github.com/ppy/osu/issues/16026
- Closes https://github.com/ppy/osu/issues/24085

Given how egregious this issue is, I would argue we use this solution as it's very simple to implement and review. Performance overhead should be negligible, and either way shouldn't matter as primary goal should be to get this in a robust state.

### DIFF
--- a/osu.Framework/Bindables/BindableList.cs
+++ b/osu.Framework/Bindables/BindableList.cs
@@ -54,11 +54,16 @@ namespace osu.Framework.Bindables
         public T this[int index]
         {
             get => collection[index];
-            set => setIndex(index, value, null);
+            set => setIndex(index, value, new HashSet<BindableList<T>>());
         }
 
-        private void setIndex(int index, T item, BindableList<T> caller)
+        private void setIndex(int index, T item, HashSet<BindableList<T>> appliedInstances)
         {
+            if (appliedInstances.Contains(this))
+                return;
+
+            appliedInstances.Add(this);
+
             ensureMutationAllowed();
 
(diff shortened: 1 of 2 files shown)


smoogipoo (approved):

osu.Framework/Bindables/BindableList.cs
```diff
@@ -350,14 +338,30 @@ private void replaceRange(int index, int count, IList newItems, BindableList<T>
...
+        /// </summary>
+        /// <param name="appliedInstances">A hash set to be passed down the recursive call tree tracking all applied instances.</param>
+        /// <returns>Whether the current instance has already been applied.</returns>
+        private bool checkAlreadyApplied(HashSet<BindableList<T>> appliedInstances)
+        {
+            if (appliedInstances.Contains(this))
+                return true;
+
+            appliedInstances.Add(this);
+            return false;
```
Susko3, line 362: Less readable but should be faster.

```suggestion
            return !appliedInstances.Add(this);
```

peppy: Nice one.

bdach (approved):

PR #5955, opened 2023-07-31
Fix `ModelBackedDrawable` potentially crashing on badly timed `null`

peppy:
The test added should hopefully explain this one, but here's a text explanation:

`CreateDrawable()` is allowed to return `null`s. The case where it returns `null` on the first call was handled by `loadDrawable()` in the early exit, but the nested case (which is called after unload operations where the first invocation of `CreateDrawable` has already been consumed) was not covered.

To fix this without requiring changes in usages, I've made `ModelBackedDrawable` use an empty drawable in cases it would otherwise have crashed due to `null` (see c341a6365326ae9305f28e047b6d8d1b3761e1df).

I've also applied NRT with existing null expectation to make things clearer.

This fixes flaky tests in `TestSceneBeatmapSetOverlay` (see https://teamcity.ppy.sh/buildConfiguration/Osu_Build/40918?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildTestsSection=true):

```csharp
TearDown : System.AggregateException : One or more errors occurred. (Value cannot be null. (Parameter 'component'))
  ----> System.ArgumentNullException : Value cannot be null. (Parameter 'component')
--TearDown
   at osu.Framework.Testing.TestScene.checkForErrors()
   at osu.Framework.Testing.TestScene.RunTestsFromNUnit()
--ArgumentNullException
   at System.ArgumentNullException.Throw(String paramName)
   at osu.Framework.Graphics.Containers.DelayedLoadWrapper.BeginDelayedLoad()
   at osu.Framework.Graphics.Containers.DelayedLoadWrapper.Update()
   at osu.Framework.Graphics.Drawable.UpdateSubTree()
   at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree()
   at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree()
   at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree()
   at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree()
```

It is also highly likely this could happen at runtime and lead to user crashes.

### DIFF
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneModelBackedDrawable.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneModelBackedDrawable.cs
@@ -32,6 +32,7 @@ namespace osu.Framework.Tests.Visual.Drawables
         public void TestEmptyDefaultState()
         {
             AddStep("setup", () => createModelBackedDrawable(false));
+            AddUntilStep("wait for load", () => backedDrawable.DelayedLoadFinished);
             AddAssert("nothing shown", () => backedDrawable.DisplayedDrawable == null);
         }
 
@@ -265,8 +266,8 @@ namespace osu.Framework.Tests.Visual.Drawables
         private partial class TestModelBackedDrawable : ModelBackedDrawable<TestModel>
         {
             public bool ShowNullModel;
-
             public bool HasIntermediate;
+            public bool DelayedLoadFinished;
 
             protected override Drawable? CreateDrawable(TestModel? model)
             {
@@ -284,6 +285,18 @@ namespace osu.Framework.Tests.Visual.Drawables
             }
 
             protected override bool TransformImmediately => HasIntermediate;
+
+            protected override void OnLoadStarted()
+            {
+                base.OnLoadStarted();
+                DelayedLoadFinished = false;
+            }
+
+            protected override void OnLoadFinished()
+            {
+                base.OnLoadFinished();
+                DelayedLoadFinished = true;
+            }
         }
     }
 }
--- a/osu.Framework.Tests/Visual/Drawables/TestSceneModelBackedDrawable.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestSceneModelBackedDrawable.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-#nullable disable
-
 using System;
 using System.Threading;
 using NUnit.Framework;
@@ -18,7 +16,7 @@ namespace osu.Framework.Tests.Visual.Drawables
 {
     public partial class TestSceneModelBackedDrawable : FrameworkTestScene
     {
-        private TestModelBackedDrawable backedDrawable;
+        private TestModelBackedDrawable backedDrawable = null!;
 
         private void createModelBackedDrawable(bool hasIntermediate, bool showNullModel = false) =>
             Child = backedDrawable = new TestModelBackedDrawable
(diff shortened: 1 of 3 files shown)


peppy:
Required diff osu! side to fix test failure:

```diff
diff --git a/osu.Game.Tests/Visual/UserInterface/TestSceneUpdateableBeatmapSetCover.cs b/osu.Game.Tests/Visual/UserInterface/TestSceneUpdateableBeatmapSetCover.cs
index a1a546d4a7..54532001a9 100644
--- a/osu.Game.Tests/Visual/UserInterface/TestSceneUpdateableBeatmapSetCover.cs
+++ b/osu.Game.Tests/Visual/UserInterface/TestSceneUpdateableBeatmapSetCover.cs
@@ -99,16 +99,18 @@ public void TestSetNullBeatmapWhileLoading()
         {
             TestUpdateableOnlineBeatmapSetCover updateableCover = null;
 
-            AddStep("setup cover", () => Child = updateableCover = new TestUpdateableOnlineBeatmapSetCover
+            AddStep("setup cover", () => Child = updateableCover = new TestUpdateableOnlineBeatmapSetCover(400)
             {
                 OnlineInfo = CreateAPIBeatmapSet(),
                 RelativeSizeAxes = Axes.Both,
                 Masking = true,
             });
 
-            AddStep("change model", () => updateableCover.OnlineInfo = null);
-            AddWaitStep("wait some", 5);
-            AddAssert("no cover added", () => !updateableCover.ChildrenOfType<DelayedLoadUnloadWrapper>().Any());
+            AddStep("change model to null", () => updateableCover.OnlineInfo = null);
+
+            AddUntilStep("wait for load", () => updateableCover.DelayedLoadFinished);
+
+            AddAssert("no cover added", () => !updateableCover.ChildrenOfType<TestOnlineBeatmapSetCover>().Any());
         }
 
         [Test]
@@ -143,11 +145,19 @@ private partial class TestUpdateableOnlineBeatmapSetCover : UpdateableOnlineBeat
         {
             private readonly int loadDelay;
 
+            public bool DelayedLoadFinished;
+
             public TestUpdateableOnlineBeatmapSetCover(int loadDelay = 10000)
             {
                 this.loadDelay = loadDelay;
             }
 
+            protected override void OnLoadFinished()
+            {
+                base.OnLoadFinished();
+                DelayedLoadFinished = true;
+            }
+
             protected override Drawable CreateDrawable(IBeatmapSetOnlineInfo model)
             {
                 if (model == null)

```

bdach:
Seems okay, but I needed to do some touch ups for code quality and a failing test framework-side (that wasn't failing locally until I applied something like the osu!-side test change). Please make sure you're OK with what transpired here.

bdach (approved):

osu.Framework.Tests/Visual/Drawables/TestSceneModelBackedDrawable.cs
```diff
@@ -33,7 +33,7 @@ public void TestEmptyDefaultState()
         {
             AddStep("setup", () => createModelBackedDrawable(false));
             AddUntilStep("wait for load", () => backedDrawable.DelayedLoadFinished);
-            AddAssert("nothing shown", () => backedDrawable.DisplayedDrawable == null);
+            AddAssert("nothing shown", () => backedDrawable.DisplayedDrawable, () => Is.Null.Or.InstanceOf(Empty().GetType()));
```
peppy, line 36: Kinda weird we have to test for both here? Haven't looked into why but I assume there's a good reason?

bdach: I mean I mostly made it "both" because both are valid for the initial state of the drawable and both mean nothing shown. It could just be the empty I guess.

PR #5956, opened 2023-07-31
Fix `DrawableAudioWrapper` deadlock by moving unbind to `UnbindAllBindables`

peppy:
Hit this a few times recently, and the fix actually looks quite simple?

Should close https://github.com/ppy/osu/issues/24013.


smoogipoo (approved):
I think this is okay

PR #5958, opened 2023-08-01
Disallow `SafeWriteStream` logic from running on finalisation

bdach:
RFC.

The core motivation for this change is several failing test runs on the osu! side such as:

- https://github.com/ppy/osu/actions/runs/5726098589/job/15515865792
- https://github.com/ppy/osu/actions/runs/5707825455/job/15464878786
- https://github.com/ppy/osu/actions/runs/5729725579/job/15527053558

in which `SafeWriteStream` appears to be in the process of being finalised and attempting to call `Dispose()`.

Finalisers are [notoriously](https://ericlippert.com/2015/05/18/when-everything-you-know-is-wrong-part-one/) [unreliable](https://ericlippert.com/2015/05/21/when-everything-you-know-is-wrong-part-two/), as they give no guarantees as to if they will be called, when they will be called, how many times they will be called, and so on, so I think having them doing anything in `SafeWriteStream` at all was wrong to begin with and the sane thing is to just not let the class run any of the "substitute the temporary file in-place" logic in the case of finalisation, because there are no sane assumptions to work off at that point anymore.

This change is made with full acknowledgement that this may lead to scenarios that could cause data loss that wouldn't cause data loss before, but the fact that they wouldn't was a pure fluke.

### DIFF
--- a/osu.Framework/Platform/Storage.cs
+++ b/osu.Framework/Platform/Storage.cs
@@ -179,7 +179,15 @@ namespace osu.Framework.Platform
 
             protected override void Dispose(bool disposing)
             {
-                if (disposing && !isDisposed)
+                // Don't perform any custom logic when arriving via the finaliser.
+                // We assume that all usages of `SafeWriteStream` correctly follow a local disposal pattern.
+                if (!disposing)
+                {
+                    base.Dispose(false);
+                    return;
+                }
+
+                if (!isDisposed)
                 {
                     // this was added to work around some hardware writing zeroes to a file
                     // before writing actual content, causing corrupt files to exist on disk.
@@ -202,7 +210,7 @@ namespace osu.Framework.Platform
 
                 base.Dispose(disposing);
 
-                if (disposing && !isDisposed)
+                if (!isDisposed)
                 {
                     storage.Delete(finalPath);
                     storage.Move(temporaryPath, finalPath);
--- a/osu.Framework/Platform/Storage.cs
+++ b/osu.Framework/Platform/Storage.cs
@@ -179,7 +179,7 @@ namespace osu.Framework.Platform
 
             protected override void Dispose(bool disposing)
             {
-                if (!isDisposed)
+                if (disposing && !isDisposed)
                 {
                     // this was added to work around some hardware writing zeroes to a file
                     // before writing actual content, causing corrupt files to exist on disk.
@@ -202,7 +202,7 @@ namespace osu.Framework.Platform
 
                 base.Dispose(disposing);
 
-                if (!isDisposed)
+                if (disposing && !isDisposed)
                 {
                     storage.Delete(finalPath);
                     storage.Move(temporaryPath, finalPath);
--- a/osu.Framework/Platform/Storage.cs
+++ b/osu.Framework/Platform/Storage.cs
@@ -208,7 +208,7 @@ namespace osu.Framework.Platform
                     }
                 }
 
-                base.Dispose(disposing);
+                base.Dispose(true);
 
                 if (!isDisposed)
                 {


osu.Framework/Platform/Storage.cs
```diff
@@ -202,7 +202,7 @@ protected override void Dispose(bool disposing)
 
                 base.Dispose(disposing);
 
-                if (!isDisposed)
```
peppy, line 205: Isn't this now incorrect because this logic isn't running? The file will remain at the temporary location as far as I can tell.

peppy (commented):
I think the more correct fix here may be to ensure all usages osu! side are being correctly disposed.. or handling the case where the directory goes away underneath the write and handle gracefully (like basic streams would probably do with single-file writes).

osu.Framework/Platform/Storage.cs
bdach: What would be incorrect precisely?

If the stream is disposed or closed, it will have `Dispose(true)` called on it. So in normal usage, when the stream is plainly disposed, the class will work precisely as it did before, won't it?

bdach:
> I think the more correct fix here may be to ensure all usages osu! side are being correctly disposed

In several hours spent on looking into this, I've seen no evidence that something isn't disposing usages correctly.

Mind you, what I _have_ noticed is the following:

- This only seems to occur in game-side tests which create their own `OsuConfigManager`s. I almost PRed a game side change related to that which added `IsDisposed` guards to `Dispose(bool)` overrides in those test cases, but then couldn't justify to myself why that fix would make sense or do anything, because `Drawable` still calls `GC.SuppressFinalize()`, so finalization shouldn't be calling those overrides. And that explanation also doesn't match the failing stack trace.
- `ConfigManager` save-on-dispose logic is super dodgy, because it receives a `Storage` from outside with no notion of lifetime. So in test scenarios it could happen that the test host cleans up the storage before the config manager is disposed. However - again - that doesn't fit the stack trace. The stack trace has finalisation in it specifically, and seems to be a result of the `SafeWriteStream` being finalised _directly_.

In other words, I'm not seeing an explanation for why this can happen other than "any usage that spawns a `SafeWriteStream` is fundamentally unsafe because finalisers are the worst ever and essentially get to run as the GC pleases". But maybe I'm missing something.

osu.Framework/Platform/Storage.cs
peppy: It will yep.

But if someone forgets to dispose, a previous case of the finalizer properly resolving the final file would now no longer work.

As in, at this point we may as well remove the finalizer altogether and rely on users correctly implementing the disposal pattern?

peppy: And I guess your thinking is "just do what `FileStream` does in this case and leave it behind, but avoid triggering any of our special logic"? It seems like a valid path since we can't just remove the base finalizer, but would want to confirm, and maybe add an inline comment mentioning this.

peppy: I'd propose this patch for readability:

```diff
diff --git a/osu.Framework/Platform/Storage.cs b/osu.Framework/Platform/Storage.cs
index 33ce7b1a5..86a609099 100644
--- a/osu.Framework/Platform/Storage.cs
+++ b/osu.Framework/Platform/Storage.cs
@@ -179,7 +179,15 @@ public SafeWriteStream(string temporaryPath, string finalPath, Storage storage)
 
             protected override void Dispose(bool disposing)
             {
-                if (disposing && !isDisposed)
+                // Don't perform any custom logic when arriving via the finaliser.
+                // We assume that all usages of `SafeWriteStream` correctly follow a local disposal pattern.
+                if (!disposing)
+                {
+                    base.Dispose(false);
+                    return;
+                }
+
+                if (!isDisposed)
                 {
                     // this was added to work around some hardware writing zeroes to a file
                     // before writing actual content, causing corrupt files to exist on disk.
@@ -202,7 +210,7 @@ protected override void Dispose(bool disposing)
 
                 base.Dispose(disposing);
 
-                if (disposing && !isDisposed)
+                if (!isDisposed)
                 {
                     storage.Delete(finalPath);
                     storage.Move(temporaryPath, finalPath);

```

bdach: > But if someone forgets to dispose, a previous case of the finalizer properly resolving the final file would now no longer work (will leave a temporary file in place and not move to the final location).

Yes it would. And that's what the last paragraph in the OP explicitly acknowledges.

> As in, at this point we may as well remove the finalizer altogether and rely on users correctly implementing the disposal pattern?

I'm not sure what this would entail specifically, because note that the finaliser is being called [from the base `FileStream` specifically](https://github.com/dotnet/runtime/blob/e2383d11c0ff656703b2889316a351b49cf7cfdf/src/libraries/System.Private.CoreLib/src/System/IO/FileStream.cs#L155). To my knowledge that means that we can't do anything about it anymore. Or, I guess you could call `GC.SuppressFinalise()` in ctor of the `SafeWriteStream`, but that's (a) mostly a weak suggestion to the GC (as per xmldoc, if the finaliser already has started to run on the object that suppresses finalisation, it's already too late, which is even more mind-blowing seeing that *finalisers can run concurrently with constructors*), and (b) messing with `FileStream` internals that I have no idea about.

peppy (approved):

peppy (approved):

Issue #3652, opened 2020-06-23
Add variable corner radius per corner to CompositeDrawables


Jeremiidesu:
It would be useful for cases like these, to avoid using sprites: (notice the top left corner that features a much higher corner radius than the bottom left corner)
![image](https://user-images.githubusercontent.com/20820425/85478552-beee7880-b5bc-11ea-96d7-d0a70651076d.png)

(In this specific situation, I want to avoid using sprites because the size of that component will vary in height)

Jeremiidesu:
Thought I'd leave my attempts at it from back in December/January here - am probably not going to touch it ever again without starting from scratch, so not opening a PR from this.

https://github.com/Jeremiidesu/osu-framework/tree/corner-radii

The test scene VariableCornerRadii I had created under Containers shows decently well what I had achieved, but it still occasionally crashed and some stuff was still broken (like hollow EdgeEffects). Mostly pushing it as possible reference should one want to check out what I had come up with.

PR #5959, opened 2023-08-02
Add support for setting corner radius per corner

Feodor0090:
... i.e. now you can have 3 corners with radius 100 and the 4th with 10, for example.

Half-broken and non-mergeable at this state. Game probably won't run due to transforms.

If the direction is ok, i need help with bounding box calculation (now it calculates using the biggest radius) and uniforms buffer - there are too many paddings and idk how to rearrange the fields so nothing breaks.
Also, non-rounded borders and rounded renders in different manner so their combination looks weird. Didn't understand what to do with this too.
image
 (bottom one is 0/50, top one is 5/50, notice the step where rounding begins)

xmldocs and interpolations will be done later.

### DIFF
--- a/osu.Framework.Tests/Visual/Containers/TestSceneCircularContainer.cs
+++ b/osu.Framework.Tests/Visual/Containers/TestSceneCircularContainer.cs
@@ -19,11 +19,11 @@ namespace osu.Framework.Tests.Visual.Containers
         public TestSceneCircularContainer()
         {
             AddStep("128x128 box", () => addContainer(new Vector2(128)));
-            AddAssert("Expect CornerRadius = 64", () => Precision.AlmostEquals(container.CornerRadius, 64));
+            AddAssert("Expect CornerRadius = 64", () => Precision.AlmostEquals(container.CornerRadius.Max(), 64));
             AddStep("128x64 box", () => addContainer(new Vector2(128, 64)));
-            AddAssert("Expect CornerRadius = 32", () => Precision.AlmostEquals(container.CornerRadius, 32));
+            AddAssert("Expect CornerRadius = 32", () => Precision.AlmostEquals(container.CornerRadius.Max(), 32));
             AddStep("64x128 box", () => addContainer(new Vector2(64, 128)));
-            AddAssert("Expect CornerRadius = 32", () => Precision.AlmostEquals(container.CornerRadius, 32));
+            AddAssert("Expect CornerRadius = 32", () => Precision.AlmostEquals(container.CornerRadius.Max(), 32));
         }
 
         private void addContainer(Vector2 size)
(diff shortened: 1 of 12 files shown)


bdach:
Previously: https://github.com/ppy/osu-framework/issues/3652#issuecomment-841837373

Feodor0090:
image

While messing with edge effect rendering code, all that i managed to learn is that bottom-right quarter size is left unchanged from actual container size, and top-left quarters are growing in size so that glow "overrun" is equal to effect radius. Removing inflation in `RectangleF effectRect = maskingInfo.Value.MaskingRect.Inflate(edgeEffect.Radius).Offset(edgeEffect.Offset);` makes this work as expected, adding `-Radius` (shrinking) makes edge overruns to top-left quarter.

There is obviously something with either UV or `g_MaskingRect`, so i randomly did

```diff
diff --git a/osu.Framework/Resources/Shaders/sh_Masking.h b/osu.Framework/Resources/Shaders/sh_Masking.h
index e9190362a..68fd25ce1 100644
--- a/osu.Framework/Resources/Shaders/sh_Masking.h
+++ b/osu.Framework/Resources/Shaders/sh_Masking.h
@@ -63,7 +63,7 @@ lowp vec4 getBorderColour()
 
 highp float getCornerRadius()
 {
-       highp vec2 relativeTexCoord = v_MaskingPosition / (g_MaskingRect.zw - g_MaskingRect.xy);
+       highp vec2 relativeTexCoord = v_MaskingPosition / ((g_MaskingRect.zw - g_MaskingRect.xy)-vec2(g_MaskingBlendRange*2, g_MaskingBlendRange*2));
        highp float top = mix(g_CornerRadius.x, g_CornerRadius.z, step(0.5, relativeTexCoord.x));
        highp float bottom = mix(g_CornerRadius.y, g_CornerRadius.w, step(0.5, relativeTexCoord.x));
        return mix(top, bottom, step(0.5, relativeTexCoord.y));
@@ -71,7 +71,7 @@ highp float getCornerRadius()
 
 highp float getInnerCornerRadius()
 {
-       highp vec2 relativeTexCoord = v_MaskingPosition / (g_MaskingRect.zw - g_MaskingRect.xy);
+       highp vec2 relativeTexCoord = v_MaskingPosition / ((g_MaskingRect.zw - g_MaskingRect.xy)-vec2(g_MaskingBlendRange*2, g_MaskingBlendRange*2));
        highp float top = mix(g_InnerCornerRadius.x, g_InnerCornerRadius.z, step(0.5, relativeTexCoord.x));
        highp float bottom = mix(g_InnerCornerRadius.y, g_InnerCornerRadius.w, step(0.5, relativeTexCoord.x));
        return mix(top, bottom, step(0.5, relativeTexCoord.y));
```
and...
image
... it works?

Was this an actual bug or i'm missing something and that's not a solution?

PR #5961, opened 2023-08-04
Fix potential user-facing error when same shader is compiled in parallel

peppy:
Noticed this when testing multiplayer spectator, and more than one user had flashlight enabled.

See inline comment for a better description.

Relevant crash:

```
2023-08-03 10:47:32 [verbose]: Requesting any changes since last known queue id 7370295
2023-08-03 10:47:33 [error]: Failed to save shader to cache.
2023-08-03 10:47:33 [error]: System.IO.IOException: The file '/Users/dean/Library/Application Support/osu-development/cache/shaders/2ad9c747fce2b1e65f237d2355197306#abd6023987102484f5046ab5bd157cca#1' already exists.
2023-08-03 10:47:33 [error]: at System.IO.FileSystem.LinkOrCopyFile(String sourceFullPath, String destFullPath)
2023-08-03 10:47:33 [error]: at System.IO.FileSystem.MoveFile(String sourceFullPath, String destFullPath, Boolean overwrite)
2023-08-03 10:47:33 [error]: at System.IO.File.Move(String sourceFileName, String destFileName, Boolean overwrite)
2023-08-03 10:47:33 [error]: at osu.Framework.Utils.General.AttemptWithRetryOnException[TException](Action action, Int32 attempts, Boolean throwOnFailure)
2023-08-03 10:47:33 [error]: at osu.Framework.Platform.NativeStorage.Move(String from, String to)
2023-08-03 10:47:33 [error]: at osu.Framework.Platform.Storage.SafeWriteStream.Dispose(Boolean disposing)
2023-08-03 10:47:33 [error]: at System.IO.Stream.Close()
2023-08-03 10:47:33 [error]: at System.IO.BinaryWriter.Dispose()
2023-08-03 10:47:33 [error]: at osu.Framework.Graphics.Shaders.ShaderCompilationStore.saveToCache(String filename, Object compilation)
2023-08-03 10:47:33 [verbose]: 🖍️ Shader Texture2D/Video compiled!
```

### DIFF
--- a/osu.Framework/Graphics/Shaders/ShaderCompilationStore.cs
+++ b/osu.Framework/Graphics/Shaders/ShaderCompilationStore.cs
@@ -84,17 +84,21 @@ namespace osu.Framework.Graphics.Shaders
                 if (CacheStorage == null)
                     return false;
 
+                string checksum;
+                string data;
+
                 lock (save_lock)
                 {
                     if (!CacheStorage.Exists(filename))
                         return false;
+
+                    using var stream = CacheStorage.GetStream(filename);
+
+                    using var br = new BinaryReader(stream);
+
+                    checksum = br.ReadString();
+                    data = br.ReadString();
                 }
-
-                using var stream = CacheStorage.GetStream(filename);
-                using var br = new BinaryReader(stream);
-
-                string checksum = br.ReadString();
-                string data = br.ReadString();
 
                 if (data.ComputeMD5Hash() != checksum)
                 {
--- a/osu.Framework/Graphics/Shaders/ShaderCompilationStore.cs
+++ b/osu.Framework/Graphics/Shaders/ShaderCompilationStore.cs
@@ -84,8 +84,11 @@ namespace osu.Framework.Graphics.Shaders
                 if (CacheStorage == null)
                     return false;
 
-                if (!CacheStorage.Exists(filename))
-                    return false;
+                lock (save_lock)
+                {
+                    if (!CacheStorage.Exists(filename))
+                        return false;
+                }
 
                 using var stream = CacheStorage.GetStream(filename);
                 using var br = new BinaryReader(stream);
--- a/osu.Framework/Graphics/Shaders/ShaderCompilationStore.cs
+++ b/osu.Framework/Graphics/Shaders/ShaderCompilationStore.cs
@@ -111,28 +111,39 @@ namespace osu.Framework.Graphics.Shaders
             return false;
         }
 
+        private static readonly object save_lock = new object();
+
         private void saveToCache(string filename, object compilation)
         {
-            if (CacheStorage == null)
-                return;
-
-            try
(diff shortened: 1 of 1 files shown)


bdach:
I mean sure, this fixes the race between writes, but what about races between reads and writes? Talking about this error in particular, which will be user-facing if it ever gets hit:

https://github.com/ppy/osu-framework/blob/e5f1aabc22f2f149f48c4fd9ca7c6f8381b00ec0/osu.Framework/Graphics/Shaders/ShaderCompilationStore.cs#L108

Possible scenario where this gets hit is something along the lines of:

```
Thread A				Thread B
------------------------------------------------------------------------------------------
tryGetCached<T>()
	finds shader cache file
	reads file
	cache miss due to corruption

CompileVertexFragment()
	compiles shader
	
saveToCache()
	takes save lock
					tryGetCached<T>()
						finds shader cache file
	deletes previous shader
						attempts to read file deleted by thread A
						"Failed to read cached shader compilation
						- recompiling." shown to user
```

peppy:
Yeah that's pretty dodgy. I'll look to fix that as well in this PR I guess.

osu.Framework/Graphics/Shaders/ShaderCompilationStore.cs
```diff
@@ -84,8 +84,11 @@ private bool tryGetCached<T>(string filename, [NotNullWhen(true)] out T? compila
...
-                if (!CacheStorage.Exists(filename))
-                    return false;
+                lock (save_lock)
+                {
+                    if (!CacheStorage.Exists(filename))
+                        return false;
+                }
 
                 using var stream = CacheStorage.GetStream(filename);
                 using var br = new BinaryReader(stream);
```
smoogipoo, line 94: Should this part, including the reading of data, not be included in the lock?

My thinking is similar to @bdach 's - thread A enters first, reads a corrupted file and recompiles, thread B starts doing the same thing and passes the above `Exists` check, but then thread A deletes the file before thread B has a chance to call `GetStream` and `ReadString()`.

Is that not important?

peppy: If it passes the exists check the full locked write has happened, so that is safe. As far as I understand it anyway.

bdach: > we should just lock the whole compile/read process

I'm pretty sure that's the closest thing to a fully correct solution.

As @smoogipoo says, the `GetStream()` call is essentially a second existence check; if that returns `null` (and it theoretically can), the lock as taken only on `Exists()` here basically did nothing. And, even _after_ getting a non-null stream, it's probably safest to hold onto the lock until the file read is completed, because otherwise there's a potential that a writing thread could try to start writing while a read is in progress on another thread - which would work pretty well using posix file operation semantics, but very much not so with windows file semantics which are very exclusive-access.

bdach (approved):

PR #5962, opened 2023-08-11
Update workarounds file to fix iOS AOT build issues via command line

frenzibyte:
Reported in https://discord.com/channels/188630481301012481/188630652340404224/1139427983132725351


peppy (approved):
worth a try

PR #5963, opened 2023-08-11
Add hint to hide iOS home gesture bar

NiceAesth:
Added to SDL2-CS in https://github.com/flibitijibibo/SDL2-CS/pull/161

https://valadoc.org/sdl2/SDL.Hint.IOS_HIDE_HOME_INDICATOR.html

This hides the annoying gesture bar which gets in the way quite a bit especially on iPhone devices.

Have tested it very briefly since Rider refuses to play nicely with Xcode 15.

### DIFF
--- a/osu.Framework.iOS/IOSWindow.cs
+++ b/osu.Framework.iOS/IOSWindow.cs
@@ -46,7 +46,7 @@ namespace osu.Framework.iOS
         {
             base.Create();
 
-            SDL.SDL_SetHint(SDL.SDL_HINT_IOS_HIDE_HOME_INDICATOR, "1");
+            SDL.SDL_SetHint(SDL.SDL_HINT_IOS_HIDE_HOME_INDICATOR, "2");
 
             window = Runtime.GetNSObject<UIWindow>(WindowHandle);
             updateSafeArea();
--- a/osu.Framework.iOS/IOSWindow.cs
+++ b/osu.Framework.iOS/IOSWindow.cs
@@ -46,6 +46,8 @@ namespace osu.Framework.iOS
         {
             base.Create();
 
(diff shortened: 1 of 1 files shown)


NiceAesth:
Nevermind, this has the same issue as the default. I had tested with a value of "1" which hid the bar until the screen was touched and assumed that "2" would behave the same.

peppy:
I'm confused. A setting of `2` is what we want. It means the first interaction with the bar is ignored. It's also what we already have live on builds. Are you sure you were testing this correctly?

Joehuu:
> Bit annoyed iOS does not provide a way to fully hide it.

You can use "guided access" to remove the bar/gesture temporarily (https://support.apple.com/en-us/HT202612):

> Guided Access limits your device to a single app and lets you control which features are available. You can turn on Guided Access when you let a child use your device, or when **accidental gestures might distract you**.

NiceAesth:
> I'm confused. A setting of `2` is what we want. It means the first interaction with the bar is ignored. It's also what we already have live on builds. Are you sure you were testing this correctly?

I had tested it with 1 which fully hid the bar, then changed to 2 assuming it would also hide the bar for that option and bring it up only on swipe instead of on tap. Apparently the behaviour that I wanted (fully hidden bar instead of dimmed) isn't possible which is a bit disappointing since the bar gets in the way quite often.



> > Bit annoyed iOS does not provide a way to fully hide it.
> 
> You can use "guided access" to remove the bar/gesture temporarily (https://support.apple.com/en-us/HT202612):
> 
> > Guided Access limits your device to a single app and lets you control which features are available. You can turn on Guided Access when you let a child use your device, or when **accidental gestures might distract you**.

Yeah, I did end up coming across that when searching for this issue. Seems like every fullscreen app has to deal with this. Very Apple-like to have a decision like this that requires annoying workarounds.

PR #5964, opened 2023-08-11
Fix crash on tablet stylus touch when stylus is over inactive area and mouse confine is off

bdach:
RFC. Closes https://github.com/ppy/osu/issues/24463.

This was always a problem, but got much louder after https://github.com/ppy/osu-framework/pull/5557. Reproduction scenario for issue:

1. Start with game in windowed state
2. Enable tablet handler
3. Change active area to be smaller than available tablet area
4. Disable mouse confine
5. Move tablet stylus to inactive area (which will cause the mouse pointer to leave the game window bounds)
6. Touch tablet

https://github.com/ppy/osu-framework/assets/20418176/e8a7f31c-f479-446a-a5cd-9472b2785329

Part of the cause here is this:

https://github.com/ppy/osu-framework/blob/75ed421f60228920abd819cebc5982cb7799bbbb/osu.Framework/Input/UserInputManager.cs#L68-L72

which allows the button release event to reach the `ButtonEventManager` without a prior press event if the stylus is pressed and released while outside window. Initially I figured I'd just remove the "mouse pressed" condition, but it turns out that doing so breaks correct handling of mouse capture when relative mouse mode is off. This has been explained with an inline comment in 05524832059a3e9c6481240e20b58d8390779d15.

I also considered implementing a "fake confine" in `OpenTabletDriverHandler`; if a positional input is sent outside of the active tablet bounds, or a stylus touch is detected when the mouse isn't inside the game window, discard the input. But this feels weird and I'm not even sure if this'll fix all cases, so I'm beelining for the simplest.

### DIFF
--- a/osu.Framework.Tests/Visual/Input/TestSceneTabletInput.cs
+++ b/osu.Framework.Tests/Visual/Input/TestSceneTabletInput.cs
@@ -3,6 +3,7 @@
 
 using System.Linq;
 using osu.Framework.Allocation;
+using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
@@ -18,6 +19,8 @@ namespace osu.Framework.Tests.Visual.Input
 {
     public partial class TestSceneTabletInput : FrameworkTestScene
     {
+        private readonly FillFlowContainer contentFlow;
+
         public TestSceneTabletInput()
         {
             var penButtonFlow = new FillFlowContainer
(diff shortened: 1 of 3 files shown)


bdach:
> I also considered implementing a \"fake confine\" in OpenTabletDriverHandler; if a positional input is sent outside of the active tablet bounds, or a stylus touch is detected when the mouse isn't inside the game window, discard the input.

Apparently there's [an opentabletdriver toggle for this](https://discord.com/channels/188630481301012481/589331078574112768/1139755424015454319). Haven't tried it yet, but maybe useful? Still not sure what the user expectations are in this case.

peppy (approved):
I think this is fine as a first step

PR #5966, opened 2023-08-16
Make `KeyBindingContainer.Prioritised` work for positional input too 

bdach:
RFC.

This change would allow to replace [this game-side workaround](https://github.com/ppy/osu/blob/34a519f80354d9de353b4feee85be15c221a2d97/osu.Game/Input/Bindings/GlobalActionContainer.cs#L166-L170) with usage of `KeyBindingContainer.IsPrioritised`. The workaround in question is a blocker in resolving a game-side p0 issue (https://github.com/ppy/osu/issues/24248#issuecomment-1652296315).

More specifically, this change ensures that the game-side workaround can be replaced with `IsPrioritised` without regressing issues like https://github.com/ppy/osu/issues/6026. The fail case is when a child of the `KeyBindingContainer` is handling a plain input handler method like `OnMouseDown()`, and handles the positional input before the `KeyBindingContainer` even has a chance to see it. This is covered in the added test cases.

### DIFF
--- a/osu.Framework.Tests/Visual/Input/TestSceneKeyBindingContainer.cs
+++ b/osu.Framework.Tests/Visual/Input/TestSceneKeyBindingContainer.cs
@@ -297,6 +297,58 @@ namespace osu.Framework.Tests.Visual.Input
             AddStep("release B", () => InputManager.ReleaseKey(Key.B));
         }
 
+        [Test]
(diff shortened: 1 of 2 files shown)


peppy:
I think this makes sense. The resultant code osu!-side also looks better (less-hacky) than what we have as a result.

peppy (approved):

smoogipoo (approved):

PR #5969, opened 2023-08-18
Add DeltaDash to the list of projects using the framework

minisbett:
I'm not quite sure whether open-source is a requirement for being added to this list, there are plans on open-sourcing sometime in the future but for the early state it was deemed closed-source is better, although certain components like parsers and algorithms for the game will be provided as NuGet packages.


peppy:
Doesn't need to be open source, all good 👍 .

peppy (approved):

PR #5970, opened 2023-08-19
Implement window flashing

ItsShamed:
The implementation is quite simple (thanks to SDL), but what worries me is _where_ it should be implemented and/or exposed.

Currently, the behaviour is implemented in `SDL2DesktopWindow` as it makes more sense due to window flashing being exclusive to desktops, and it is exposed in `DesktopGameHost`, which is a class that can be accessed by consumer games through DI and (safe-)casting and that can access internal `IWindow` implementations – in this case, it would be the bridge between the game and the framework's windowing internals.

It could be argued that the implementation in `SDL2DesktopWindow` is redundant and everything could be implemented in `DesktopGameHost` by gathering the window handle via `(Window as SDL2Window)?.SDLWindowHandle` for example – and I'd agree with that, it's just a matter of opinions and what feels best. 
Leaving like that for the time being unless instructed otherwise.


osu.Framework/Platform/DesktopGameHost.cs
```diff
@@ -26,6 +26,10 @@ protected DesktopGameHost(string gameName, HostOptions options = null)
             IsPortableInstallation = Options.PortableInstallation;
         }
 
+        public void FlashWindow(bool briefly = true) => (Window as SDL2DesktopWindow)?.Flash(briefly);
+
+        public void CancelWindowFlashing() => (Window as SDL2DesktopWindow)?.CancelFlash();
```
bdach, line 31: Disagree with proxying windowing methods to `DesktopGameHost`. Nothing else does this, even though we already have platform-exclusive windowing things.

Please add to `IWindow` instead, and throw `NotSupportedException` in the `OsuTKWindow` implementation. This doesn't matter for our uses since that's an only-android thing. Consumers can go through `host.Window` instead.

osu.Framework.Tests/Visual/Platform/TestSceneWindowFlash.cs
```diff
@@ -0,0 +1,71 @@
...
+using osu.Framework.Allocation;
+using osu.Framework.Bindables;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Sprites;
+using osu.Framework.Platform;
+
+namespace osu.Framework.Tests.Visual.Platform
+{
+    public partial class TestSceneWindowFlash : FrameworkTestScene
```
bdach, line 13: Should be ignored on CI

```csharp
    [Ignore("This test cannot run in headless mode (a window instance is required).")]
```

bdach (changes requested):

osu.Framework/Platform/OsuTKWindow.cs
```diff
@@ -94,6 +94,10 @@ public void Raise()
         {
         }
 
+        public void Flash(bool _) => throw new NotSupportedException("OsuTK windows can't flash.");
```
peppy, line 97: Should probably be `PlatformNotSupportedException`?

Also I question how we implement this as a consumer. Do we have to read the documentation, realise it only works with a desktop game window and do a pre-check for that? If so, the xmldoc should explicitly reference that class. Alternatively, we could make this a no-op instead if that's deemed as easier for usage.

ItsShamed: > Alternatively, we could make this a no-op instead if that's deemed as easier for usage.

Yes, that's probably better and that was what I initially did (idk why I did all that refactoring 😅)

ItsShamed:
Also, what should be the default behaviour? Should it flash briefly or until focused? (I'd tend to choose the former personally)

peppy (approved):

bdach (approved):

PR #5973, opened 2023-08-20
Bump Android target SDK version to 33

bdach:
[Google wants this done by August 31.](https://developer.android.com/google/play/requirements/target-sdk)

The implications of this for framework projects appear to be minimal. The biggest one is probably the change to media permissions which makes them more granular - but we're not handling the permission granting flow correctly yet anyways, so it's really a nothingburger.

That said, to test the visual tests file picker, the visual tests android project now requests the three media permissions (but you still have to enable them yourself via the system UI). Not declaring them means that you couldn't even do that.

This will introduce a new build warning:

```
Microsoft.Android.Sdk.Tooling.targets(68, 5): [XA1008] The TargetFrameworkVersion (Android API level 31) is lower than the targetSdkVersion (33). Please increase the `$(TargetFrameworkVersion)` or decrease the `android:targetSdkVersion` in the `AndroidManifest.xml` so that the API levels match.
```

but I cannot make it go away. Tried the following to no avail:

```diff
diff --git a/osu.Framework.Android.props b/osu.Framework.Android.props
index 1a479229b..ffd083b94 100644
--- a/osu.Framework.Android.props
+++ b/osu.Framework.Android.props
@@ -1,6 +1,7 @@
 ∩╗┐<Project>
   <PropertyGroup>
     <SupportedOSPlatformVersion>21.0</SupportedOSPlatformVersion>
+    <TargetFrameworkVersion>v13.0</TargetFrameworkVersion>
     <RuntimeIdentifiers>android-x86;android-arm;android-arm64</RuntimeIdentifiers>
     <AndroidPackageFormat>apk</AndroidPackageFormat>
     <AndroidHttpClientHandlerType>Xamarin.Android.Net.AndroidMessageHandler</AndroidHttpClientHandlerType>
```

Doesn't appear to be a props thing either, putting the above in the final .csproj also does nothing. I think this one will only go away when xamarin releases catch up. Or we migrate to net7.0. It's a weird one because it doesn't even seem to break anything (the apps themselves seem to work just fine, and the granular permissions thing appears to work flawlessly too).

### DIFF
--- a/osu.Framework.Tests.Android/AndroidManifest.xml
+++ b/osu.Framework.Tests.Android/AndroidManifest.xml
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" package="osu.Framework.Tests.Android" android:installLocation="auto">
-	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="31" />
+	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="33" />
 	<application android:label="osu!framework test" />
 </manifest>
--- a/osu.Framework.Tests.Android/AndroidManifest.xml
+++ b/osu.Framework.Tests.Android/AndroidManifest.xml
@@ -2,4 +2,8 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" package="osu.Framework.Tests.Android" android:installLocation="auto">
 	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="33" />
 	<application android:label="osu!framework test" />
+	<!-- permissions below are mostly requested for testing the file selector -->
+	<uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
+	<uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
+	<uses-permission android:name="android.permission.READ_MEDIA_AUDIO" />
 </manifest>
--- a/SampleGame.Android/AndroidManifest.xml
+++ b/SampleGame.Android/AndroidManifest.xml
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" package="SampleGame.Android" android:installLocation="auto">
-	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="31" />
+	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="33" />
 	<application />
 </manifest>
--- a/osu.Framework.Android/Properties/AssemblyInfo.cs
+++ b/osu.Framework.Android/Properties/AssemblyInfo.cs
@@ -17,8 +17,6 @@ using Android.App;
     ResizeableActivity = true,
     Theme = "@android:style/Theme.Black.NoTitleBar"
 )]
-[assembly: UsesPermission(Manifest.Permission.ReadExternalStorage)]
-[assembly: UsesPermission(Manifest.Permission.WriteExternalStorage)]
 [assembly: UsesPermission(Manifest.Permission.WakeLock)]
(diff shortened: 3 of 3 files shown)


smoogipoo (approved):

peppy (approved):

osu.Framework.Tests.Android/AndroidManifest.xml
```diff
@@ -1,5 +1,9 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" package="osu.Framework.Tests.Android" android:installLocation="auto">
-	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="31" />
+	<uses-sdk android:minSdkVersion="21" android:targetSdkVersion="33" />
 	<application android:label="osu!framework test" />
+	<!-- permissions below are mostly requested for testing the file selector -->
+	<uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
+	<uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
+	<uses-permission android:name="android.permission.READ_MEDIA_AUDIO" />
```
Susko3, line 8: Could be changed to assembly-level `UsesPermission` so consumers doesn't have to explicitly have `uses-permission` in the manifest.

```diff
diff --git a/osu.Framework.Android/Properties/AssemblyInfo.cs b/osu.Framework.Android/Properties/AssemblyInfo.cs
index 16db49e19..ab69a8998 100644
--- a/osu.Framework.Android/Properties/AssemblyInfo.cs
+++ b/osu.Framework.Android/Properties/AssemblyInfo.cs
@@ -19,6 +19,9 @@
 )]
 [assembly: UsesPermission(Manifest.Permission.ReadExternalStorage)]
 [assembly: UsesPermission(Manifest.Permission.WriteExternalStorage)]
+[assembly: UsesPermission("android.permission.READ_MEDIA_IMAGES")]
+[assembly: UsesPermission("android.permission.READ_MEDIA_VIDEO")]
+[assembly: UsesPermission("android.permission.READ_MEDIA_AUDIO")]
 [assembly: UsesPermission(Manifest.Permission.WakeLock)]
 [assembly: UsesPermission(Manifest.Permission.ReadFrameBuffer)]
 [assembly: UsesPermission(Manifest.Permission.Internet)]
```

This should allow consumers to make a [runtime request](https://developer.android.com/reference/android/app/Activity#requestPermissions(java.lang.String[],%20int)) for only the permissions they need (eg. [osu! might only request `IMAGES` and `AUDIO`](https://github.com/ppy/osu/pull/24599)). (But this is just a thought for the far future, as Android runtime permission requests not yet implemented in osu!framework).

Susko3 (commented):
Not really happy with the new warning, but it'll be fixed when we upgrade to net7.0 so it's fine.

osu.Framework.Tests.Android/AndroidManifest.xml
bdach: This would make this apply transitively to every framework project, correct? Going by principle of minimal privileges I don't think we want that.

bdach: If we're going to have to be moving stuff out then why bother unifying...

bdach: In fact I've removed two permissions from assembly manifest in 8e076b2967d133cbbeb3a76253448fc08e6e49e0 because they now [do](https://developer.android.com/reference/android/Manifest.permission#READ_EXTERNAL_STORAGE) [nothing](https://developer.android.com/reference/android/Manifest.permission#WRITE_EXTERNAL_STORAGE).

I really see no reason why something like `SampleGame` should be requesting any storage permissions anywhere. Even if it is some hidden app store view. It just looks shady if it doesn't need those permissions to function.

Everything else in assembly manifest (wake lock, read framebuffer, internet access) is 100% reasonable for any game, but storage access is where it gets dicey for me.

peppy (approved):

Issue #5974, opened 2023-08-20
Reduce FFmpeg binary size


FreezyLemon:
The current ffmpeg build scripts already remove some unnecessary components. The feature set can be reduced a lot further though.

<details>
<summary>Current features (./configure)</summary>

```
External libraries:
mediafoundation

External libraries providing hardware acceleration:
d3d11va                 dxva2

Libraries:
avcodec                 avfilter                avformat                avutil                  swscale

Programs:

Enabled decoders:
aac                     bfi                     hevc                    pcm_f24le               stl
aac_fixed               bink                    hnm4_video              pcm_f32be               subrip
aac_latm                binkaudio_dct           hq_hqa                  pcm_f32le               subviewer
aasc                    binkaudio_rdft          hqx                     pcm_f64be               subviewer1
ac3                     bintext                 huffyuv                 pcm_f64le               sunrast
ac3_fixed               bitpacked               hymt                    pcm_lxf                 svq1
acelp_kelvin            bmp                     iac                     pcm_mulaw               svq3
adpcm_4xm               bmv_audio               idcin                   pcm_s16be               tak
adpcm_adx               bmv_video               idf                     pcm_s16be_planar        targa
adpcm_afc               brender_pix             iff_ilbm                pcm_s16le               targa_y216
adpcm_agm               c93                     ilbc                    pcm_s16le_planar        text
adpcm_aica              cavs                    imc                     pcm_s24be               theora
adpcm_argo              ccaption                imm4                    pcm_s24daud             thp
adpcm_ct                cdgraphics              imm5                    pcm_s24le               tiertexseqvideo
adpcm_dtk               cdtoons                 indeo2                  pcm_s24le_planar        tiff
adpcm_ea                cdxl                    indeo3                  pcm_s32be               tmv
adpcm_ea_maxis_xa       cfhd                    indeo4                  pcm_s32le               truehd
adpcm_ea_r1             cinepak                 indeo5                  pcm_s32le_planar        truemotion1
adpcm_ea_r2             clearvideo              interplay_acm           pcm_s64be               truemotion2
adpcm_ea_r3             cljr                    interplay_dpcm          pcm_s64le               truemotion2rt
adpcm_ea_xas            cllc                    interplay_video         pcm_s8                  truespeech
adpcm_g722              comfortnoise            jacosub                 pcm_s8_planar           tscc2
adpcm_g726              cook                    jpeg2000                pcm_u16be               tta
[...]

PR #5980, opened 2023-08-23
Remove libavfilter where possible

FreezyLemon:
The framework never calls any of the filter functions, and FFmpeg itself doesn't load the av* libraries dynamically, so this should be fine. The decoding tests in TestSceneVideo also pass without the library present.

Android and iOS get pre-built ffmpeg libraries from an external package, so this won't do much for them except not loading the library into memory at runtime.


bdach:
Looks to work on windows, android, and linux. On linux I've done `LD_DEBUG=libs ./osu.Framework.Tests > test.log 2>&1` and got no hits when grepping that log for `libavfilter` so seems good.

Probably needs apple platform testing before merge. Reminder that for desktop testing a local `osu.Framework.NativeLibs` must be created (`dotnet pack -c Release /p:Version=2023.823.0`). And we'll need to deploy that to nuget and bump it in the root framework project after this is merged.

bdach (approved):

FreezyLemon:
> Reminder that for desktop testing a local `osu.Framework.NativeLibs` must be created (`dotnet pack -c Release /p:Version=2023.823.0`) and referenced in `osu.Framework.csproj`.

For testing, just deleting the avfilter file in `runtimes/<platform>/native` should be fine aswell, as long as you don't rebuild. You could also just leave it there and check if the library gets loaded (procexp on Windows, strace on Linux show that it's not loaded at all, even if the file exists).

>And we'll need to deploy that to nuget and bump it in the root framework project after this is merged.

Might be worth waiting, I'm currently working on #5974 and that would need another `osu.Framework.NativeLibs` NuGet release.

peppy:
Tested on macOS to work as expected.

peppy (approved):

PR #5984, opened 2023-08-24
Change behaviour of `Storage.Move` to overwrite existing by default

peppy:
Coming from https://github.com/ppy/osu/actions/runs/5948760882/job/16133189924, the only way I can see this happening is if the underlying `File.Move` logic is less atomic than we expect it to be.

In the process of looking into this, I noticed there's an `overwrite` flag available for `File.Move` which tends to prefer a `rename` operation internally on unix, rather than `link`.

On windows (the failcase in question) it likely won't help at the end of the day.

Anyway, personally I think having it overwrite by default is what we prefer from our API. If that isn't agreed on then closing this without further consideration is fine.

### DIFF
--- a/osu.Framework/Platform/NativeStorage.cs
+++ b/osu.Framework/Platform/NativeStorage.cs
@@ -48,7 +48,7 @@ namespace osu.Framework.Platform
         {
             // Retry move operations as it can fail on windows intermittently with IOExceptions:
             // The process cannot access the file because it is being used by another process.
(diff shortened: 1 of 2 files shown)


smoogipoo:
I weakly agree with this. The other option is to do as .NET does and add an overload.

bdach (approved):
No issue from me. Up to you if you wanna entertain making that a flag and what the default should be, but I think it's basically fine as is.

PR #5988, opened 2023-09-11
Disallow overriding `DrawScale`

peppy:
This was added in a time long forgotten specifically for storyboard usage. Since then, we have established much better ways of doing this (easiest is just to make a nested container).

The reasoning for sealing this is that draw scale adds a hidden scale which is not publicly exposed and can affect all children. It caused me much frustration in understanding how storyboard elements were a completely unexpected size.

Given that there was only a few osu-local usages of this, I think we can remove this as a breaking change.

Other candidates are `Drawable.Width` and `Drawable.Height` which should be `sealed`, but that's a bit higher effort due to at least on framework side override usage in a class deriving `CompositeDrawable`.

- [ ] Wait for osu! side changes to be finalised and accepted as working (coming later)


bdach:
Uh... I'm having a bit of a dissonance between the title and the contents of the PR, since this appears to be _removing_ `DrawScale` as a member. Which, I guess _does_ have the result of disallowing overriding it, but in general is not how I'd describe the change in a vacuum...

Not that I'm _opposed_ to the change, just slightly taken aback.

peppy:
Reworded 😃

PR #5989, opened 2023-09-12
Fix `OpenUrlExternally` potentially crashing applications due to misconfigurations (and log errors instead)

yesseruser:
Added a try/catch statement to all overrides of the osu.Framework.Platform.GameHost.OpenUrlExternally method (in Desktop, IOS and Android platforms) for the bodies of the methods (excluding the url valid check).
Fixes https://github.com/ppy/osu/issues/24696


osu.Framework.iOS/IOSGameHost.cs
```diff
@@ -14,9 +14,11 @@
 using osu.Framework.IO.Stores;
 using osu.Framework.iOS.Graphics.Textures;
 using osu.Framework.iOS.Graphics.Video;
+using osu.Framework.Logging;
 using osu.Framework.Platform;
 using osu.Framework.Platform.MacOS;
 using UIKit;
+using static System.Net.Mime.MediaTypeNames;
```
bdach, line 21:
```suggestion
```

osu.Framework/Platform/DesktopGameHost.cs
```diff
@@ -6,9 +6,11 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Linq.Expressions;
```
bdach, line 9:
```suggestion
```

bdach:
The added `Logger` was massively weird and I've reverted it in https://github.com/ppy/osu-framework/pull/5989/commits/9bc79c648a50deaf6d40b62a104c4163b0d4548c. The general advice is: if you don't see things like error messages extracted anywhere already, probably best to not do it.

Other than that, this is probably fine, as far as I can tell. I'm not sure the iOS variant makes sense, because I'm not sure if `InvokeOnMainThread()` marshals error correctly in a way that can even make the try-catch work. I can see a world wherein it does completely nothing and the try-catch should instead be moved *inside* the `InvokeOnMainThread()` callback. But I'll leave testing that to someone with an actual device.

bdach (approved):
see above

yesseruser:
Okay, thank you.

peppy:
>  I'm not sure the iOS variant makes sense, because I'm not sure if InvokeOnMainThread()

I don't think it does, but I also don't think it matters. This is mainly for desktop applications where it's been throwing.

PR #5991, opened 2023-09-14
Allow testflight/app store links to be opened externally

peppy:
This allows apps using osu!framework to redirect users to find a new version of the app from testflight.

Closes https://github.com/ppy/osu/issues/22832.


bdach (approved):
Can't test myself, but leaving an approval for the general approach taken and placement of the fix.

Pedantically speaking one may desire the protocol allowlist to be configurable but let's cross that bridge only when we know it absolutely must be crossed.

Issue #5986, opened 2023-09-08
`TextureAtlas` isn't using space efficiently


EVAST9919:
Suppose we have 2 textures to be added into the atlas: one is wide and another is tall (images 1 & 2). Now if we want to add another texture - it should be very small to fit with the current implementation (3), otherwise new atlas will be created (4).
The issue is that the huge empty atlas area is being unused.

| 1 | 2 | 3 | 4 |
| ----------- | ----------- | ----------- | ----------- |
| ![1](https://github.com/ppy/osu-framework/assets/22874522/9381f3b1-edea-4909-96e6-88e44ed4eaba) | ![2](https://github.com/ppy/osu-framework/assets/22874522/82be76dc-64d1-4856-b6cf-7bfb8389f585) | ![3](https://github.com/ppy/osu-framework/assets/22874522/f7a0c9c3-cda6-4c4a-86bc-fbfcc7407735) | ![4](https://github.com/ppy/osu-framework/assets/22874522/a046f0d9-9d41-4398-9329-9b3470ffa453) |

Implementing some sort of packing algorithm would probably be an overkill, but there should be at least some simple approach to make things better.

[Relevant code](https://github.com/ppy/osu-framework/blob/fc5cd7ec2a5d5f7ec6b3e258350073c2e093f9bc/osu.Framework/Graphics/Textures/TextureAtlas.cs#L144C9-L175C10)
[Test scene to play around with](https://github.com/EVAST9919/osu-framework/blob/texture-atlas-packing/osu.Framework.Tests/Visual/Layout/TestSceneTextureAtlasPacking.cs)

smoogipoo:
The algorithm that's chosen here should also consider being able to remove areas from the atlas. This is something that blocks us from using these atlases for larger more-ephemeral textures like avatars and beatmap/carousel backgrounds.

PR #5994, opened 2023-09-15
Implement `IRectanglePacker` interface and various packing algorithms

EVAST9919:
A step towards resolving https://github.com/ppy/osu-framework/issues/5986.
`TextureAtlas` has been reworked to support new `IRectanglePacker` interface, so switching the algorithm is a one line change. For now it's using the same algo as previously since I'm not sure about performance concerns (probably benchmarking needed).

Also adds a nice test scene to play around with:
Снимок экрана (13)


peppy:
If performance is a perceived concern, I'd recommend adding benchmarks. They should be very simple to create, since we already have all the tooling in place.

EVAST9919:
So I've added some benchmarks in which I'm creating and populating bins with 20 rectangles of various sizes. As you can see the difference is quite big. Basically we are trading performance for potential capacity (as expected, probably).
From these we can assume that Guillotine is a good middle ground, although it can peroform worse than Shelf (currently used in TextureAtlas on master) in some cases fitness-wise. However with the underlying structure we can easily track big chunks of previously wasted space [(as per issue)](https://github.com/ppy/osu-framework/issues/5986) and removing a rectangle is as simple as adding the new empty space to already existing list, which is [a desired behavior](https://github.com/ppy/osu-framework/issues/5986#issuecomment-1720734832).
_Also note that my implementation probably isn't perfect and can potentially be improved._
|                 Method |         Mean |       Error |      StdDev |   Gen0 | Allocated |
|----------------------- |-------------:|------------:|------------:|-------:|----------:|
|          PopulateShelf |     914.7 ns |     1.48 ns |     1.38 ns | 0.0191 |      40 B |
| PopulateShelfRemainder |   1,039.4 ns |     2.06 ns |     1.92 ns | 0.0381 |      80 B |
|     PopulateGuillotine |  11,709.0 ns |    29.93 ns |    28.00 ns | 0.5341 |    1128 B |
|        PopulateMaximal | 337,441.9 ns | 1,453.73 ns | 1,359.82 ns | 0.9766 |    2176 B |

PR #5997, opened 2023-09-20
Don't block cut/copy platform actions on empty textbox selection

peppy:
Closes https://github.com/ppy/osu/issues/24867.

If this is seen as a bad direction, I can fix this locally at osu!'s end (using the `AllowClipboardExport` property).

### DIFF
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -240,14 +240,15 @@ namespace osu.Framework.Graphics.UserInterface
                 // Clipboard
                 case PlatformAction.Cut:
                 case PlatformAction.Copy:
-                    // Blocking even when no text is selected could be deemed more correct, but this means shift-delete / ctrl-insert
-                    // cannot be handled elsewhere. We've seen usages of this, so let's not block unless we handle for now.
-                    if (string.IsNullOrEmpty(SelectedText) || !AllowClipboardExport) return false;
+                    if (!AllowClipboardExport) return false;
 
-                    clipboard.SetText(SelectedText);
+                    if (!string.IsNullOrEmpty(SelectedText))
+                    {
+                        clipboard.SetText(SelectedText);
 
-                    if (e.Action == PlatformAction.Cut)
-                        DeleteBy(0);
+                        if (e.Action == PlatformAction.Cut)
+                            DeleteBy(0);
+                    }
 
                     return true;
 
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -240,7 +240,9 @@ namespace osu.Framework.Graphics.UserInterface
                 // Clipboard
                 case PlatformAction.Cut:
                 case PlatformAction.Copy:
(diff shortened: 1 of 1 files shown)


smoogipoo:
I don't know about this. The song select issue occurs because of a very specific interaction of a textbox that holds onto focus, which is not something the framework provides.

bdach:
Yeah I'm not feeling this either. The line for me probably draws at

```diff
diff --git a/osu.Framework/Graphics/UserInterface/TextBox.cs b/osu.Framework/Graphics/UserInterface/TextBox.cs
index 5f5e70900..a7391434f 100644
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -240,9 +240,8 @@ public virtual bool OnPressed(KeyBindingPressEvent<PlatformAction> e)
                 // Clipboard
                 case PlatformAction.Cut:
                 case PlatformAction.Copy:
-                    // Blocking even when no text is selected could be deemed more correct, but this means shift-delete / ctrl-insert
-                    // cannot be handled elsewhere. We've seen usages of this, so let's not block unless we handle for now.
-                    if (string.IsNullOrEmpty(SelectedText) || !AllowClipboardExport) return false;
+                    if (!AllowClipboardExport) return false;
+                    if (string.IsNullOrEmpty(SelectedText)) return true;
 
                     clipboard.SetText(SelectedText);
 
```

because the empty string selection case being handled like that just feels really weird to me from a general standpoint. But even with the above, you'd still have to do

```diff
diff --git a/osu.Game/Graphics/UserInterface/FocusedTextBox.cs b/osu.Game/Graphics/UserInterface/FocusedTextBox.cs
index 338f32f321..ed4b24213a 100644
--- a/osu.Game/Graphics/UserInterface/FocusedTextBox.cs
+++ b/osu.Game/Graphics/UserInterface/FocusedTextBox.cs
@@ -26,6 +26,8 @@ public partial class FocusedTextBox : OsuTextBox, IKeyBindingHandler<GlobalActio
         /// </summary>
         protected virtual bool ClearTextOnBackKey => true;
 
+        protected override bool AllowClipboardExport => false;
+
         public void TakeFocus()
         {
             if (!allowImmediateFocus)

```

game-side, at which point I have several questions (why would focused textboxes be excluded from clipboard? what about being able to convey that to the user? etc. etc.), at which point it is _probably_ better to do something else, like [the proposed solution with `Tab` being used to manage focus on such views](https://github.com/ppy/osu/issues/24867#issuecomment-1726786526) - but that would _also_ in my view be a major UX regression of search just in order to bring back a hotkey I wasn't even _aware of the existence of_ until now.

Honestly I'd probably go and nuke the hotkey. Unless it's a stable holdover or something I see little point.

peppy:
It is a stable thing. And I think it's nice to have a delete hotkey (and can't think of a better one..)

bdach:
> Your proposed solution was my fallback method failing what is in this PR.

We can *probably* try it, but I'd wager that 95% of players are more accustomed to being able to just type in song select to search than having a delete hotkey and that search is used disproportionately more frequently.

peppy:
Oh definitely not the `Tab` one, just the disabling clipboard and non-blocking in this case. I'll change this PR to do that instead.

bdach (approved):

smoogipoo (approved):

PR #5998, opened 2023-09-20
Avoid throwing exception when attempting to access `WebRequest` response in fail path

peppy:
One less potential web request exception = maybe a touch better debug performance? Not sure, but this throws on every timeout of requests [osu! side](https://github.com/ppy/osu/blob/9f842ccdc057e6c6ec60def1ac607f1ea00f02e1/osu.Game/Online/API/APIRequest.cs#L199) and can easily be avoided.

### DIFF
--- a/osu.Framework/IO/Network/JsonWebRequest.cs
+++ b/osu.Framework/IO/Network/JsonWebRequest.cs
@@ -22,9 +22,6 @@ namespace osu.Framework.IO.Network
 
         protected override void ProcessResponse()
         {
-            if (ResponseStream == null)
-                return;
-
             string response = GetResponseString();
             if (response != null)
                 ResponseObject = JsonConvert.DeserializeObject<T>(response);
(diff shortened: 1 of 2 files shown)


bdach:
Can wholeheartedly agree with this, the existing `catch` branches already returning null make this a no-brainer for me.

In fact I've piled on a few more changes; https://github.com/ppy/osu-framework/pull/5998/commits/f04ae81032697de524393725557d1feeaec6dae3 and https://github.com/ppy/osu-framework/pull/5998/commits/b3995cca7196951bc8e2e609fc74998051ef9fb3 are mostly annotations / docs, but https://github.com/ppy/osu-framework/pull/5998/commits/0e1c9e62688a0a432b33362126e941c45cae823e is notable for removing a similar condition placed in `JsonWebRequest` (i.e. a subclass).

Please check that you agree with these before we go forward with this.

peppy:
Changes look good 👍 .

Issue #705, opened 2017-05-09
Add ASIO/WASAPI drivers to Reduce Audio Latency


posesix:
Opening an issue for discussion since this is something that should be included in the official distribution.

osu! uses default BASS procedures, which use DirectSound on Windows, causing large latencies due to high latency software mixing in Windows Vista and Windows 7 (as I've tested). I get **70ms latency** on average using 240FPS on both my Windows 7 64-bit PC and Windows Vista 32-bit laptop. This may be less of a problem on Linux since BASS defaults to the ALSA driver, which is a low-level & low-latency driver, but I have not tested it fully.

The BASS API has extensions for ASIO and WASAPI support under the same licenses.
We can effectively (hopefully) reduce the latency to less than 4ms on ASIO, and a worse case on WASAPI (Exclusive).

BASS does not unify their main API with ASIO and WASAPI APIs for some reason, so some changes and additions to audio code will be needed. These APIs are very easy to work with, and pretty much the same as using their main API once you've got them initialised. We have to set up a BASS MIXER and load sounds and music as BASS STREAMS, instead of SAMPLES. The STREAMS are connected to the MIXER and the MIXER feeds audio data to the ASIO/WASAPI event procedures. BASS CHANNEL procedures can be used with streams just like samples.

We will also need to add options for sound drivers in the options menu. It will simply list available drivers depending on your platform. A simple drop-down box can be created after detecting available drivers. Options could be "_Default_", "_ASIO_", "_WASAPI (Shared)_", and "_WASAPI (Exclusive)_". A note should be added to WASAPI (Exclusive) and ASIO since they take exclusive access of an audio device, so only osu! can be playing sound from that device.

Linux does not use ASIO, so this will not affect Linux. WASAPI is Windows-exclusive, but does not achieve lower latencies compared to ASIO.

This will also add some dynamic-link dependencies and the use of add-ons from ManagedBass. BASSMIX, BASSASIO, and BASSWASAPI will be used. 

I can work on the code for ASIO & WASAPI support, but some minor design changes may be required.

Tom94:
The latency you describe is not inherent to windows audio mixing. I've had many sound cards, and for most of them there was no perceptible delay, even through DirectSound and an active audio mixer. Therefore, the fault can't lie with windows' audio pipeline. Seems to me like the fault lies with the hardware / driver and not DirectSound.

[...]

PR #6002, opened 2023-09-23
Implement SDL3 Audio backend

hwsmm:
https://github.com/ppy/osu-framework/pull/6002#issuecomment-3226583712


smoogipoo:
This is super cool, though I don't see osu! using this in the near future if ever. I'd see this implemented as an additive nuget package, and exposed through some way that isn't the framework config. For example, it could be included in `HostOptions` or simply as a virtual method in `Game`.

peppy:
I'd actually like to experiment with this and see how good support is for the upcoming changes I want to make with WASAPI initialisation. So I don't want to throw this out. Moving away from bass would be a huge consideration, but I wouldn't throw it away.

I'm going to mark this as a draft as I don't see it getting reviewed or merged anytime soon, but I still useful to have around as a reference for what is involved in making this work, and potential performance / latency cross-checking in a future.

hwsmm:
I'll maintain this until some of you have anything to do with this, mostly because I am now so used to playing the game with these patches, I can't really go back to BASS...
FWIW, on Linux with Pipewire at very small buffer, hitsound latency was almost on par with patched wine which osu! linux players mostly use. It produces some artifacts, so shouldn't be default, though.

peppy:
@hwsmm thanks for keeping this up-to-date. curious on your thoughts on the move to sdl3 – does this help in any way on the audio side of things? i don't want to promise anything but we have internally discussed the potential of moving to sdl3 as an audio provider at some point.

curious on how wasapi is looking specifically, as i read that sdl3 was going to add better support, potentially replacing my attempt at initialising for lower latency (https://github.com/ppy/osu-framework/pull/6088) which ended up having compatibility issues and limited results.

hwsmm:
SDL3 Audio doesn't give any improvements over SDL2 for the usecase of osu!framework in my opinion.
From what I see, improvements in SDL3 audio is mainly to let developers reduce burden of writing audio subsystems by themselves, and my implementation already has mixers and everything, so they are kind of meaningless here and even add a bit of overhead in some cases.

SDL has a good WASAPI implementation, but they don't provide any means to adjust buffer sizes (https://github.com/libsdl-org/SDL/blob/main/src/audio/wasapi/SDL_wasapi.c#L635 -- only uses default '0' value).
Buffer length usually ends up being around 10ms (or more, should be plenty good?) on most systems I tested on. I am not sure how better it could be compared to BASS DirectSound.
We can make use of [this API](https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient3-getsharedmodeengineperiod) to determine lowest possible latency, but it's only available on Windows 10 or higher. I can probably write a PR (or a patch for our SDL builds?) to SDL for this if anyone wants, but I guess you can also just implement it within C# and pass the value to BASSWASAPI.
Though I didn't know such API existed until now. I don't really play osu! on Windows, anyway..

However, the problem that hinders _real_ low latency lies in .NET GC overhead. As I wrote in the PR description and the previous comment, audio stutters a lot if I force low latency on Linux, and it should be the same on Windows, too.
Any value below 10ms seems to cause underruns in audio, and it's very noticeable if you go back to main menu from song select in game.
I remember that [the same](https://discord.com/channels/188630481301012481/1189078742555951214/1189429052080132096) happened in your WASAPI testing discussions on Discord, and this is probably because [WASAPI callback](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Threading/AudioThread.cs#L225) exists in managed land.
I even wrote a simple C library out of boredom to handle SDL audio callback in unmanaged land, and it solved most of audio stutters issues on very small buffer on Linux (around 1ms), but you won't want this since it requires a couple of additional native libraries.

So what I am trying to say here is, that even with the Windows API above, it's going to be pretty hard to get the stable low latency audio experience unless audio is handled in completely unmanaged land.
The best solution might be making a native library for BASSWASAPI callback if you only consider Windows, but I would like to see any kind of support for Linux since there are some players (including me) not wanting to make a complete switch to lazer because patched WINE provides better latency.

smoogipoo:
> However, the problem that hinders real low latency lies in .NET GC overhead.

That's, honestly, the first time I'm hearing of this. I would imagine that code returns in a fraction of a microsecond so I wouldn't expect the GC to have that sort of impact on it, but I suppose it would make sense for super small buffer lengths.

If it is GC and not e.g. the cost of context switching, there's a few things I'd try:
- Wrapping in `GC.TryStartNoGCRegion()`/`GC.EndNoGCRegion()`? 
- Using `UnmanagedCallersOnly`.
- Using `SuppressGCTransition`.

Perhaps, all at the same time.

hwsmm:
I guess my comment was a bit ambiguous. I meant that the audio callback returns fast, but the problem is supposedly 'stop-the-world' effect of GC.
Whenever STW occurs, SDL fails to get audio data in time from the audio callback located in managed land.
Please correct me if I misunderstood your message, I have only used C# for less than an year at this point.

Here are my test results for things you mentioned:
1. `GC.TryStartNoGCRegion()/GC.EndNoGCRegion()` pair mostly fixes the problem, except for longer stutter from time to time.
2. `UnmanagedCallersOnly` is already applied to the audio callback.
3. `SuppressGCTransition` doesn't seem to help. Only one native method is used in the callback, anyway..

TheComputerGuy96:
It would be nice to combine all of this audio component work and make a new open-source BASS-compatible audio library
that everyone could use (not just C# applications) in a separate repository (I'm personally going to call it Treble until a better name can be found)

hwsmm:
First of all, this PR is not compatible with BASS. It just implements o!f audio components in C# and SDL, while the current o!f audio backend is written in BASS API.
Which means, this PR calculates current track position by converting data position to milliseconds in C#, but current o!f calls BASS_ChannelGetPosition (and some more) to do the same.
It would be very time consuming to re-expose internal APIs in a new project and to make it compatible with BASS and its addons (o!f uses BASSMIX, BASSFX and BASSWASAPI), and BASS is still maintained well.

I do love open source, and it's true that I wrote this PR to get rid of a closed source lib from osu!framework, but there is only little point to both audio library consumers and us in maintaining a new audio library when there are mature open-source audio libraries out there. Better contribute to them if they are missing some features. It could be another example for [xkcd new standard](https://xkcd.com/927/) in my opinion.

Just consider this as a random audio backend for a game engine. Unity, Unreal and Godot, they all have their own audio backend, but they are not standalone, right?

hwsmm:
Update: I did some research on WASAPI, and found that IAudioClient3 (available on Windows 10 and higher) makes it possible to use smaller buffer size than 10ms according to [this](https://learn.microsoft.com/en-us/windows-hardware/drivers/audio/low-latency-audio#faq), and BASSWASAPI is probably not using it since it only takes seconds as buffer/period units. IAudioClient3 takes sample counts as an initialization argument.

I submitted a patch to SDL to support IAudioClient3 for lower latency: https://github.com/libsdl-org/SDL/pull/9683 (it got merged a month ago, so it's included in the last SDL3 bump), and I think it might be what @peppy has wanted for a better WASAPI initialization?

I did some test with the forementioned SDL patch:
1. Place speaker, keyboard, and microphone closely to one another
2. Start recording, and play a very simple chart with beep hitsound
3. Measure the gap between hitsound and keyboard hit with Audacity

|         | Current SDL | SDL w/ the patch | Default BASS | BASSWASAPI |
|---------|-------------|------------------|--------------|-------------|
| Latency | 35ms        | 15ms             | 40ms         | 35ms        |

There were also noticeable audio stutters with small buffer on Windows, too (not during gameplay though). My native library didn't bring such stutters on Windows, just as on Linux, so it's probably a GC problem again.

Latency is configurable with [`SDL_HINT_AUDIO_DEVICE_SAMPLE_FRAMES`](https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_hints.h#L318), but I am not sure how we should go about it, as the experience with low latency is not exactly satisfying.

---

Also, the new mixer effect interface feels a bit weird to reimplement since `IEffectParameter` cannot have any additional members in it. I tried fixing this in my way, and [this](https://gist.github.com/hwsmm/4d46c40315c1c2e858197bdb8f37c080) is my current draft diff. It isn't applied to current PR tree since it requires an osu!-side patch.

... and I came up with [this conversation on Discord](https://discord.com/channels/188630481301012481/188630652340404224/1250777251465789470) after having done the work.
I'll just use this in my fork if it's not desirable. It doesn't require a dictionary in mixer, so (really a bit) less overhead, too. Not sure if we should keep an applied FX list in a mixer...

peppy:
@hwsmm very cool to hear. The work I did on WASAPI initialisation seems like it should have worked (and in my very limited isolation testing it did) but after appplying to osu! the results people got showed the buffer not actually respecting what it should have based on setting. So you are likely on to something here.

Would want @smoogipoo to take a look at the effect API to see if he's on board with the direction you specify.

I was discussing this branch recently with smoogi and touched on potentially getting it merged / replacing BASS and he mentioned effects not working yet. But on reading through this, it seems like BQF effects *are* working and that's all we use right now?

smoogipoo:
Last time I tried this branch, it still had a few bugs/unimplemented stuff. But now that SDL3's API is stabilised, how do you feel about getting this in at some point? Even if it's not complete, I think having something in the working set will help with development/get more eyes on it.

hwsmm:
To be honest, I'm a bit unsure about this PR at this point.

It can bring some goodies to the framework, such as:
- Free audio stack without introducing a new native library (FFmpeg needs to support audio decoding though)
- All of components except for decode/output are managed, so it should be easy to debug
- Better support for Linux audio servers, and potential support for niche architectures

However, from what I've seen for an year, most people who want this to get merged want low-latency audio, which this PR fails to deliver due to GC stutters. No improvements on Windows (only good until 10ms buffer), maybe even worse on Linux (good until 20ms) if you want 'good' output.
Gameplay is fine probably because of LowLatency GC option which we can't use for long time.

So, as a solution for this, I have been writing a C library in the meantime as I mentioned above, and it turned out to be quite nice for me so far. Using this can also remove raw audio processing code from this PR, effectively making framework consume an external package for alternative audio backends, but obviously, it's native, so potentially unsafe.

I could have tried improving an existing audio library like I said above, but I ended up making a new one, because decoding needs to happen asynchronously not to delay playback, and it needs to support tempo adjust/bq filters/resampling. I couldn't find any open-source libraries that can do all of them, so I did the 'new standard' thing... that was pretty fun, so I don't regret. It was not that hard because all I did was rewriting what I wrote into C, too.

I know that this isn't the best way, but audio stutter is **way more** annoying than visual stutters. You just get pop noises into your ears.

I want this PR to be good enough for what users expect, and a viable alternative to BASS for most usecases.
Note that it is not a replacement of BASS as it just re-implements my C# functions in C, so almost everything is one or two call away without converting units and things because it was pretty much made for osu!framework, but it's also standalone, so I guess it makes more sense to maintain and package it in my separate repository and periodically bump version here.

I will have to do some work (and probably a lot of work on mobile toolchains) to actually introduce it here, so I'd like to know how you think about this option.

I understand you even if you don't want this, because I also think this sounds like a step backwards in some ways. Current C# version can still live as a free alternative for the framework.

smoogipoo:
Beyond just latency, we still have ongoing reports of offset gradually falling behind (long trail starting here https://github.com/ppy/osu-stable-issues/issues/306, stable issue but the same happens on lazer - see referencing issues) and stutters at the start of maps (BASS startup time). The second we can resolve by changing the clock we use as reference but the first is kind of hard to do anything with - un4seen doesn't work closely enough with us to get anywhere.

As far as I've been told, both of those issues are fixed with SDL3.

I'm not against using native code, as long as bulk of the core logic is C#.

smoogipoo:
Am thinking you might even be able to make a C# NativeAOT project for it if people are uncomfortable with C. I have a bunch of experience setting that up on all relevant platforms including mobile if needed.

hwsmm:
Sorry for not responding quickly.

That's actually not surprising since I started writing this backend because I was consistently hitting early on the game no matter what offset I use.

Honestly, I don't really know what to do, so I couldn't reply to your comment on time. I'm fine with maintaining either C or C# version, but if we end up with C# one, I think we'd get another bug report about audio stutters.
The native library unfortunately contains most of core logic to avoid GC stutters as much as possible. It is just an audio library. You create a channel, add it to a mixer, and then it plays.

I guess NativeAOT might work if people dislike using C, but we are not compiling the entire game with it, right? If we are compiling the audio framework in a shared object with NativeAOT, should we make bindings for it, or is there a way to dynamically load a natively AOT-compiled library in C#?

smoogipoo:
> I'm fine with maintaining either C or C# version, but if we end up with C# one, I think we'd get another bug report about audio stutters

The hopeful idea is for the NativeAOT lib to do things in such a way that every allocation is accounted for, while also making it touchable by core developers (who are most experienced in C#).

> I guess NativeAOT might work if people dislike using C, but we are not compiling the entire game with it, right?

Nah. I don't think we ever will - the JIT is actually quite useful for us optimisation-wise. But that won't help in this case because the idea is to isolate the osu! GC from the NativeAOT GC.

> If we are compiling the audio framework in a shared object with NativeAOT, should we make bindings for it, or is there a way to dynamically load a natively AOT-compiled library in C#?

You need to expose functions using `[UnmanagedCallersOnly]` (`EntryPoint` is required) ([example](https://github.com/smoogipoo/osu-native/blob/8e8b9721bce0fb448bf8b610d800c68872884adb/Sources/osu.Native/Program.cs#L31-L54)). Then you use it just as a normal native lib ([example](https://github.com/smoogipoo/osu-native/blob/8e8b9721bce0fb448bf8b610d800c68872884adb/Bindings/CSharp/osu.Native.Bindings/Native.cs)).

hwsmm:
Yeah, I agree on the point, too.

But I am worrying about how we should pass objects between .NET and NativeAOT. Do we need to create `ObjectHandle`s for everything, as .NET objects are not blittable?

smoogipoo:
Yeah. You'd basically interface with it as if it's a normal C lib -  there's no passing of objects or hackily dereferencing into a matching managed signature (except for blittable/`unmanaged` types ofc).

hwsmm:
Well, it sounds like rewriting my C library into C# again if I am understanding it correctly.

I think it shouldn't take so long since I already have the .NET part from my native project (unless I run into weird issues). I'll start writing it sooner or later if you are fine with it.

smoogipoo:
Hold off on that one for a bit. Do you have a sample of the C lib to look at to see what needs to be done?

hwsmm:
https://gist.github.com/hwsmm/c2bb0e55694e14c83372b8bdcb73830b

Here are (hand-written) function bindings. Once all of them are implemented, it should work well.
I hope function names are clear enough to tell what they do.

hwsmm:
I've just sent a mail!

hwsmm:
Starting from https://discord.com/channels/188630481301012481/589331078574112768/1396539637861847141, I copied most part from this PR to the new library, and I got most things working... on my PC.

As a side note, I was actually a few hours before (pre-)releasing my native library here when smoogi sent me that message since it had been pretty stable lately, so I was not motivated to write it again in NativeAOT, but I ended up deciding to do because it required much less effort because I already had some in C#.

I honestly got surprised with Satori. When I finished the basic implementation and tested with plain NativeAOT, the result was not good on very small buffer (around 1ms).

However, with Satori, it still had some underruns, but was very acceptable than plain NativeAOT and this PR.

Anyway, here are some points to know:
- It is LGPL because SoundTouch.Net (LGPL v2.1) is statically compiled.
- BASS is the default decoder because it needs to work for now. It can also use FFmpeg 7 with libswresample and audio decoder if it is available and BASS init code is removed from `SDL3AudioManager.cs`.
- Decoder is integrated because I want to keep API stable for my another native library (not public yet).
- Android probably works (not tested yet), but no iOS because I don't have any Apple devices. Too cumbersome to figure out xcframework when I can only use CI for Apple toolchain.
- Satori is not available on macOS x86_64 (due to [build failure](https://github.com/hwsmm/Satori/actions/runs/17020964347) - [this upstream issue](https://www.github.com/dotnet/runtime/issues/104105) suggests that the issue has been fixed in new XCode 16 releases but no luck for me) and Android (don't know how)
- Repository structure is pretty rough right now. I may fix in the future, but it works for now ([repo](https://github.com/hwsmm/mysoundlib)).
- You can set an environment variable `SDL_AUDIO_DEVICE_SAMPLE_FRAMES` to adjust latency. 128 should work well on Windows with a good audio driver.

How to try:
1. `git clone -b sdl-audio https://github.com/hwsmm/osu-framework.git`
2. Apply [this commit](https://gist.github.com/hwsmm/8eb12a0e99deec4218576c8a942fe42e) on top of upstream osu
3. `cd osu && ./UseLocalFramework.sh`
4. `cd osu.Desktop && dotnet run`
5. Edit AudioDriver in framework.ini to `SDL3`

This thing is still very early, so it may be (completely) broken in some cases. I'd be grateful if you can try and let me know if it works well.

hwsmm:
@smoogipoo I am not expecting an immediate response, but here is a mention in case you didn't enable notification on this thread.

smoogipoo:
This is still in my emails but very low priority at the moment. I've been waiting on a conclusion to the discussion in https://github.com/ppy/osu-framework/issues/705 + https://github.com/ppy/osu-framework/pull/6628 because @peppy's already reviewing that.

Rudicito:
@hwsmm I tested your changes. Note that I'm using 2025.816.0-lazer, not the latest version, because your framework is a bit outdated.

I'm on Linux (Fedora 42).

I tried with this [beatmap](https://osu.ppy.sh/beatmapsets/1173192#osu/2446833). Note that I’ve set a 10 ms audio offset, since that’s what I use in my regular lazer client, that I deactivate beatmap hitsound, and that I use argon pro.

## With BASS:
20251022_18h01m46s_grim

## With SDL3 (without set SDL_AUDIO_DEVICE_SAMPLE_FRAMES):
20251022_18h02m10s_grim

## With SDL3 (SDL_AUDIO_DEVICE_SAMPLE_FRAMES = 128):
20251022_18h56m04s_grim

## With SDL3 (SDL_AUDIO_DEVICE_SAMPLE_FRAMES = 64)
20251022_18h43m10s_grim

For SDL3 and SDL_AUDIO_DEVICE_SAMPLE_FRAMES set to 64 and 128, the hitsound feels more responsive, feel like I was often clicking early. Didn't encounter audio issues with these values. In the editor, I feel like the hitsound are early.

With SDL_AUDIO_DEVICE_SAMPLE_FRAMES set to 32, sound start crackling, didn't try playing a beatmap.

Hope this help, you can ask me to test things if you want.

Maybe I should have test with no audio offset? I can do it if you want.

hwsmm:
@Rudicito Hey, thanks for testings! Reporting issues you encounter while using it is sufficient for now (you may use whatever offset you want). sdl-audio-alt branch is now updated, so you can probably update your game.

unknowwiiplayer:
The following was written with AI assistance based on my own testing and findings.

I am not an audio developer, so I cannot evaluate the architectural or maintenance concerns involved in merging this work. My contribution here is empirical: I am an experienced taiko player, and input-to-hitsound latency is the main reason I normally avoid osu! and prefer OpenTaiko with ASIO on Windows.

I tested the newer `sdl-audio-alt` work against osu! `2026.711.0-lazer` on CachyOS with PipeWire.

To get the current game compiling, I used:

* the `sdl-audio-alt` branch of `hwsmm/osu-framework`
* the existing `AudioFilter` compatibility patch
* a small additional adaptation for the newer effects in `AfToggleSection.cs`

On the osu! side, the additional adaptation was small: six newer `AddEffect()` / `RemoveEffect()` calls were migrated to `GetNewEffect()`, `Apply()`, and `Remove()`.

The result compiled successfully.

I was then able to play stably with PipeWire running at a 48-frame quantum at 44.1 kHz.

The gameplay difference was not subtle. Input-to-hitsound response felt dramatically tighter, and osu! felt better to me than it ever has before, including my experience playing on Windows. I only play taiko, so that is the mode I can confidently evaluate.

I understand that a working local build does not address all of the architectural, maintenance, mobile, or cross-platform concerns involved in upstreaming an alternative audio backend.

However, it does demonstrate that the newer SDL3 audio work can still run on the current game, and that its practical benefit for latency-sensitive Linux players can be enormous.

I would really like to see this work revisited and given a concrete path toward an officially supported low-latency audio option, even if it initially remains experimental or Linux-only.

I can provide the exact osu! diff and testing details if they are useful.

hwsmm:
I am sorry, but osu! project does not accept any kind of AI assistance (including documentation things). Please take a look at this section: https://github.com/ppy/osu-framework#contributing

I just updated my patches in the `sdl-audio-alt` comment. It's not a big deal anyway, since I still play the game with my backend and fix issues as I find them.

Also I moved the old branch to `sdl-audio-legacy`, since I have not been touching it for long time.

PR #6005, opened 2023-09-26
Fix keyboard handler getting disabled on window creation failure

### LINKED ISSUE #6007: `GameHost.ChooseAndSetupRenderer()` silences exceptions from `CreateWindow()`
Root cause of https://github.com/ppy/osu/issues/24260 (very likely).

The most notable exception in `CreateWindow()` is `SDL_Init` failing. This can be forced with:
```diff
diff --git a/osu.Framework/Platform/SDL2Window.cs b/osu.Framework/Platform/SDL2Window.cs
index 8ae1d7215..3cfc63f69 100644
--- a/osu.Framework/Platform/SDL2Window.cs
+++ b/osu.Framework/Platform/SDL2Window.cs
@@ -186,7 +186,7 @@ protected SDL2Window(GraphicsSurfaceType surfaceType)
         {
             ObjectHandle = new ObjectHandle<SDL2Window>(this, GCHandleType.Normal);
 
-            if (SDL.SDL_Init(SDL.SDL_INIT_VIDEO | SDL.SDL_INIT_GAMECONTROLLER) < 0)
+            if (SDL.SDL_Init(SDL.SDL_INIT_VIDEO | SDL.SDL_INIT_GAMECONTROLLER) < int.MaxValue)
             {
                 throw new InvalidOperationException($"Failed to initialise SDL: {SDL.SDL_GetError()}");
             }
```

```cs
2023-09-26 09:05:44 [verbose]: 🖼️ Configuration renderer choice: Automatic
2023-09-26 09:05:44 [verbose]: 🖼️ Renderer fallback order: [ Direct3D 11, OpenGL (Legacy), OpenGL, Vulkan ]
2023-09-26 09:05:44 [verbose]: 🖼️ Initialising "Veldrid" renderer with "Direct3D11" surface
2023-09-26 09:05:44 [verbose]: 🖼️ Initialising "WindowsGL" renderer with "OpenGL" surface
2023-09-26 09:05:44 [verbose]: 🖼️ Initialising "Veldrid" renderer with "OpenGL" surface
2023-09-26 09:05:44 [verbose]: 🖼️ Initialising "Veldrid" renderer with "Vulkan" surface
2023-09-26 09:05:44 [error]: No usable renderer was found!
2023-09-26 09:05:44 [error]: An unhandled error has occurred.
2023-09-26 09:05:44 [error]: System.NullReferenceException: Object reference not set to an instance of an object.
2023-09-26 09:05:44 [error]: at osu.Framework.Input.Handlers.Tablet.OpenTabletDriverHandler.Initialize(GameHost host) in C:\...\osu-framework\osu.Framework\Input\Handlers\Tablet\OpenTabletDriverHandler.cs:line 53
2023-09-26 09:05:44 [error]: at osu.Framework.Platform.GameHost.initialiseInputHandlers() in C:\...\osu-framework\osu.Framework\Platform\GameHost.cs:line 1098
2023-09-26 09:05:44 [error]: at osu.Framework.Platform.GameHost.Run(Game game) in C:\...\osu-framework\osu.Framework\Platform\GameHost.cs:line 727
2023-09-26 09:05:44 [error]: at osu.Framework.Tests.Program.Main(String[] args) in C:\...\osu-framework\osu.Framework.Tests\Program.cs:line 23
```

This fails with the tabled driver trying to dereference `host.Window`. And the `InvalidOperationException` from `SDL_Init` failing is nowhere to be seen.

peppy:
Has been reported by multiple users, probably due to a once-off startup failure caused by external factors.

Can be tested with something like:

```diff
diff --git a/osu.Framework/Platform/MacOS/MacOSGameHost.cs b/osu.Framework/Platform/MacOS/MacOSGameHost.cs
index 1600b9af0..2cb5e4432 100644
--- a/osu.Framework/Platform/MacOS/MacOSGameHost.cs
+++ b/osu.Framework/Platform/MacOS/MacOSGameHost.cs
@@ -21,7 +21,7 @@ internal MacOSGameHost(string gameName, HostOptions options)
         {
         }
 
-        protected override IWindow CreateWindow(GraphicsSurfaceType preferredSurface) => new MacOSWindow(preferredSurface);
+        protected override IWindow CreateWindow(GraphicsSurfaceType preferredSurface) => null;
 
         public override IEnumerable<string> UserStoragePaths
         {

```

JetBrains Rider 2023-09-26 at 06 42 55

---

Closes https://github.com/ppy/osu/issues/24260.
Closes #6007.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -53,6 +53,11 @@ namespace osu.Framework.Platform
     {
         public IWindow Window { get; private set; }
 
+        /// <summary>
+        /// Whether <see cref="Window"/> needs to be non-null for startup to succeed.
+        /// </summary>
+        protected virtual bool RequireWindowExists => true;
+
         public IRenderer Renderer { get; private set; }
 
         public string RendererInfo { get; private set; }
@@ -724,6 +729,12 @@ namespace osu.Framework.Platform
 
                 ChooseAndSetupRenderer();
 
+                if (RequireWindowExists && Window == null)
+                {
+                    Logger.Log("Aborting startup as no window could be created.");
+                    return;
+                }
+
                 initialiseInputHandlers();
 
                 // Prepare renderer (requires config).
@@ -964,10 +975,11 @@ namespace osu.Framework.Platform
             // Prepare window
             Window = CreateWindow(surfaceType);
 
-            if (Window == null)
+            if (RequireWindowExists && Window == null)
             {
-                Logger.Log("🖼️ Renderer could not be initialised, no window exists.");
-                return;
+                // Window creation may fail in the case of a catastrophic failure (ie. graphics driver or SDL2 level).
+                // In such cases, we want to throw here to immediately mark this renderer setup as failed.
+                throw new InvalidOperationException("🖼️ Renderer could not be initialised as window creation failed.");
             }
 
             try
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -729,6 +729,8 @@ namespace osu.Framework.Platform
 
                 ChooseAndSetupRenderer();
 
+                // Window creation may fail in the case of a catastrophic failure (ie. graphics driver or SDL2 level).
+                // In such cases, we want to throw here to immediately mark this renderer setup as failed.
                 if (RequireWindowExists && Window == null)
                 {
                     Logger.Log("Aborting startup as no window could be created.");
@@ -979,11 +981,10 @@ namespace osu.Framework.Platform
 
                 if (Window == null)
                 {
+                    // Can be null, usually via Headless execution.
                     if (!RequireWindowExists)
                         return;
 
-                    // Window creation may fail in the case of a catastrophic failure (ie. graphics driver or SDL2 level).
-                    // In such cases, we want to throw here to immediately mark this renderer setup as failed.
                     throw new InvalidOperationException("🖼️ Renderer could not be initialised as window creation failed.");
                 }
 
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -972,18 +972,21 @@ namespace osu.Framework.Platform
             Renderer = renderer;
             Renderer.CacheStorage = CacheStorage.GetStorageForDirectory("shaders");
 
-            // Prepare window
-            Window = CreateWindow(surfaceType);
-
-            if (RequireWindowExists && Window == null)
-            {
-                // Window creation may fail in the case of a catastrophic failure (ie. graphics driver or SDL2 level).
-                // In such cases, we want to throw here to immediately mark this renderer setup as failed.
-                throw new InvalidOperationException("🖼️ Renderer could not be initialised as window creation failed.");
-            }
-
             try
             {
+                // Prepare window
+                Window = CreateWindow(surfaceType);
+
+                if (Window == null)
+                {
+                    if (!RequireWindowExists)
+                        return;
+
+                    // Window creation may fail in the case of a catastrophic failure (ie. graphics driver or SDL2 level).
+                    // In such cases, we want to throw here to immediately mark this renderer setup as failed.
+                    throw new InvalidOperationException("🖼️ Renderer could not be initialised as window creation failed.");
+                }
+
                 Window.SetupWindow(Config);
                 Window.Create();
                 Window.Title = $@"osu!framework (running ""{Name}"")";
--- a/osu.Framework/Platform/HeadlessGameHost.cs
+++ b/osu.Framework/Platform/HeadlessGameHost.cs
@@ -47,6 +47,8 @@ namespace osu.Framework.Platform
             this.realtime = realtime;
         }
 
+        protected override bool RequireWindowExists => false;
+
         protected override IWindow CreateWindow(GraphicsSurfaceType preferredSurface) => null;
 
         protected override Clipboard CreateClipboard() => new HeadlessClipboard();


osu.Framework/Platform/GameHost.cs
```diff
@@ -964,10 +975,11 @@ protected void SetupRendererAndWindow(IRenderer renderer, GraphicsSurfaceType su
             // Prepare window
             Window = CreateWindow(surfaceType);
 
-            if (Window == null)
+            if (RequireWindowExists && Window == null)
```
Susko3, line 978: This change doesn't make much sense, because `CreateWindow()` never really returns `null`. The actual failure is `CreateWindow()` throwing, see #6007.

smoogipoo (approved):

Susko3 (changes requested):
Needs to consider the failure case of #6007.

osu.Framework/Platform/GameHost.cs
Susko3: I think this part needs to be reverted, `RequireWindowExists` seems to actually fix the problem.

If you want to fix #6007 in this pull as well, I think it's best to move `CreateWindow()` to inside the try/catch block.

peppy: I'm quite confused.

This was all intentional, as `CreateWindow` takes a `surfaceType` and could fail for some renderer options and not others. So the try-catch and the `throw` are both intentional here.

I'm not sure how to process #6007. I guess this PR can be marked as closing it?

Susko3: > This was all intentional, as `CreateWindow` takes a `surfaceType` and could fail for some renderer options and not others. So the try-catch and the throw are both intentional here.

Correct. But `CreateWindow` can fail by throwing an exception and not by returning `null`.

> I guess this PR can be marked as closing it?

Not in its current state. It may throw an exception, but the original exception is still silenced. I think it's important to have SDL_Init failures be logged.

peppy: Okay, so something like this?

```diff
diff --git a/osu.Framework/Platform/GameHost.cs b/osu.Framework/Platform/GameHost.cs
index 601e0df81..691a3aa17 100644
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -972,18 +972,18 @@ protected void SetupRendererAndWindow(IRenderer renderer, GraphicsSurfaceType su
             Renderer = renderer;
             Renderer.CacheStorage = CacheStorage.GetStorageForDirectory("shaders");
 
-            // Prepare window
-            Window = CreateWindow(surfaceType);
-
-            if (RequireWindowExists && Window == null)
-            {
-                // Window creation may fail in the case of a catastrophic failure (ie. graphics driver or SDL2 level).
-                // In such cases, we want to throw here to immediately mark this renderer setup as failed.
-                throw new InvalidOperationException("🖼️ Renderer could not be initialised as window creation failed.");
-            }
-
             try
             {
+                // Prepare window
+                Window = CreateWindow(surfaceType);
+
+                if (RequireWindowExists && Window == null)
+                {
+                    // Window creation may fail in the case of a catastrophic failure (ie. graphics driver or SDL2 level).
+                    // In such cases, we want to throw here to immediately mark this renderer setup as failed.
+                    throw new InvalidOperationException("🖼️ Renderer could not be initialised as window creation failed.");
+                }
+
                 Window.SetupWindow(Config);
                 Window.Create();
                 Window.Title = $@"osu!framework (running ""{Name}"")";

```

Susko3:
```diff
diff --git a/osu.Framework/Platform/GameHost.cs b/osu.Framework/Platform/GameHost.cs
index 601e0df81..d21829c35 100644
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -972,18 +972,17 @@ protected void SetupRendererAndWindow(IRenderer renderer, GraphicsSurfaceType su
             Renderer = renderer;
             Renderer.CacheStorage = CacheStorage.GetStorageForDirectory("shaders");
 
-            // Prepare window
-            Window = CreateWindow(surfaceType);
-
-            if (RequireWindowExists && Window == null)
-            {
-                // Window creation may fail in the case of a catastrophic failure (ie. graphics driver or SDL2 level).
-                // In such cases, we want to throw here to immediately mark this renderer setup as failed.
-                throw new InvalidOperationException("­ƒû╝´©Å Renderer could not be initialised as window creation failed.");
-            }
-
             try
             {
+                // Prepare window
+                Window = CreateWindow(surfaceType);
+
+                if (Window == null)
+                {
+                    // window can be null when running headless
+                    return;
+                }
+
                 Window.SetupWindow(Config);
                 Window.Create();
                 Window.Title = $@"osu!framework (running ""{Name}"")";
```

Something like this.

Maybe `Window == null` meaning 'headless run' is not that great. Time to look into having `DummyWindow`?

peppy: I still think my proposed diff may make more sense if we can expect window creation to potentially fail for a specific `surfaceType`. @smoogipoo do you know if this is possible?

Susko3: Your proposed diff and the current state of the PR make the tests not work due to nullref of `Window` in
```cs
Window.SetupWindow(Config);
```

peppy: Right, that failed along the refactoring path. Can you check the latest version?

osu.Framework/Platform/GameHost.cs
```diff
@@ -961,17 +972,21 @@ protected void SetupRendererAndWindow(IRenderer renderer, GraphicsSurfaceType su
...
+                Window = CreateWindow(surfaceType);
+
+                if (Window == null)
+                {
+                    if (!RequireWindowExists)
+                        return;
+
+                    // Window creation may fail in the case of a catastrophic failure (ie. graphics driver or SDL2 level).
+                    // In such cases, we want to throw here to immediately mark this renderer setup as failed.
+                    throw new InvalidOperationException("🖼️ Renderer could not be initialised as window creation failed.");
```
Susko3, line 987: I think it's impossible that `CreateWindow` returns `null` (except when running headless). So this is dead code with a scary comment.

I'm fine with the exception (it verifies that hosts with `RequireWindowExists == true` don't create `null` windows), but the comment should be moved to where it actually applies: to the `RequireWindowExists` check in `GameHost.Run()`.

Susko3 (changes requested):
Everything else seems fine to me and is more correct.
- Closes #6007 with the new changes

osu.Framework/Platform/GameHost.cs
peppy: Sure, I've moved it.

Susko3 (approved):

smoogipoo (approved):

PR #6006, opened 2023-09-26
Stop logging non-error output from OpenTabletDriver

peppy:
See https://github.com/ppy/osu/discussions/24915. Maybe wait for a response before going ahead with this, but it doesn't seem worthwhile to log non-errors to the user.


smoogipoo (approved):
Seems like @X9VoiD is on board with this change according to the discussion thread, so I'll go ahead and merge.

PR #6009, opened 2023-09-29
Fix `WaveformGraph` overhead when `DrawPosition` is changed

peppy:
The previous attempt to fix this (https://github.com/ppy/osu-framework/pull/5197) helped, but there were still some scenarios where invalidation would cause regeneration.

Let's just do our own invalidation for now.

One thing I'm not sure about is whether checking `DrawSize` here is enough (should I be using `ScreenSpaceDrawQuad.Size` instead?).

Closes https://github.com/ppy/osu/issues/24951

| Before | After |
| :---: | :---: |
| osu! 2023-09-29 at 10 03 06 | osu! 2023-09-29 at 10 02 11 |

### DIFF
--- a/osu.Framework/Graphics/Audio/WaveformGraph.cs
+++ b/osu.Framework/Graphics/Audio/WaveformGraph.cs
@@ -14,7 +14,6 @@ using osu.Framework.Graphics.Rendering;
 using osu.Framework.Graphics.Rendering.Vertices;
 using osu.Framework.Graphics.Shaders;
 using osu.Framework.Graphics.Textures;
-using osu.Framework.Layout;
 using osu.Framework.Logging;
 using osu.Framework.Utils;
 using osuTK;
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Audio/WaveformGraph.cs
```diff
@@ -158,18 +157,15 @@ public Color4 BaseColour
...
-            bool result = base.OnInvalidate(invalidation, source);
+            base.Update();
 
-            if ((invalidation & Invalidation.RequiredParentSizeToFit) > 0)
-            {
-                // We should regenerate when `Scale` changed, but not `Position`.
-                // Unfortunately both of these are grouped together in `MiscGeometry`.
+            // Can't use invalidation for this as RequiredParentSizeToFit is closes, but also triggers on DrawPosition changes.
+            if (lastGeneratedDrawSize != null && DrawSize != lastGeneratedDrawSize)
                 queueRegeneration();
```
smoogipoo, line 168: The comment previously mentioned scale, but that's not included in `DrawSize`. Perhaps it makes more sense this way, though? That is - when we think of scale it's usually in the context of stretching data and not generating new data for the stretched bounds?

smoogipoo (approved):
The screenshots don't really make much sense as an observer to the issue, but I think it's fine.

peppy:
> The screenshots don't really make much sense as an observer to the issue, but I think it's fine.

Yeah it's subtle. See the spikes on the update thread, which occur every best in the editor.

smoogipoo:
As a side note, `TestSceneWaveform` is absolute crap... What even is that "zoom" slider?

Can there maybe be a `TestSceneWaveformGraph` that demonstrates some of these cases - moving it around, changing resolution, scaling, resizing, along with a counter for number of regens for each? There's already a `OnWaveformRegenerated` virtual method as part of `WaveformGraph`, so all I'm imagining is that incrementing a counter similar to `TestSceneCachedBufferedContainer`.

osu.Framework/Graphics/Audio/WaveformGraph.cs
bdach: `RequiredParentSizeToFit = MiscGeometry | DrawSize`. `MiscGeometry` invalidates on `Scale`, so the comment made sense. I don't necessarily disagree with ignoring it though as you say.

bdach (approved):
Let's go with this for now

bdach:
> Can there maybe be a `TestSceneWaveformGraph` that demonstrates some of these cases - moving it around, changing resolution, scaling, resizing, along with a counter for number of regens for each? There's already a `OnWaveformRegenerated` virtual method as part of `WaveformGraph`, so all I'm imagining is incrementing a counter similar to `TestSceneCachedBufferedContainer` from there.

To speed things up I'm just gonna leave that for the next time we touch that I think. Or a separate PR.

PR #6014, opened 2023-10-04
Fix race condition in ThreadedTaskScheduler when adding tasks

smoogipoo:
See: https://github.com/ppy/osu/runs/17361055215#r0s1

### DIFF
--- a/osu.Framework.Tests/Threading/ThreadedTaskSchedulerTest.cs
+++ b/osu.Framework.Tests/Threading/ThreadedTaskSchedulerTest.cs
@@ -83,5 +83,28 @@ namespace osu.Framework.Tests.Threading
             Assert.That(run.Wait(30000));
             Assert.That(exited.Wait(30000));
         }
+
+        [Test]
+        [Repeat(1000)]
+        public void QueueAndDisposeStressTest()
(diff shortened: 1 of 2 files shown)


peppy:
Is there a reason the test coverage for this can't be committed?

smoogipoo:
How? It's a race condition... It's just fails at some random point.

bdach:
We've added such tests to source in the past. Compare:
 https://github.com/ppy/osu-framework/blob/ecc3ce5cfe4c8467c152cb46ee1c464e4d5f0932/osu.Framework.Tests/Polygons/ConvexPolygonClipperFuzzingTest.cs#L17
https://github.com/ppy/osu-framework/blob/ecc3ce5cfe4c8467c152cb46ee1c464e4d5f0932/osu.Framework.Tests/Visual/Drawables/TestSceneStressGLDisposal.cs#L16

peppy:
Even if it's just run a few hundred times (and only fails one out of several runs), I think having the test around as a reference is valuable for any future manual testing.

smoogipoo:
TIL. Added the test.

peppy (approved):

PR #6017, opened 2023-10-04
Fix source clock not correctly getting started after a decoupled transition

peppy:
Cause is pretty obvious...

### DIFF
--- a/osu.Framework.Tests/Clocks/DecouplingFramedClockTest.cs
+++ b/osu.Framework.Tests/Clocks/DecouplingFramedClockTest.cs
@@ -397,6 +397,17 @@ namespace osu.Framework.Tests.Clocks
 
             Assert.That(source.CurrentTime, Is.EqualTo(decouplingClock.CurrentTime).Within(30));
             Assert.That(source.IsRunning, Is.True);
+
+            // Subsequently test stop/start works correctly.
+            decouplingClock.Stop();
+            decouplingClock.ProcessFrame();
+            Assert.That(decouplingClock.IsRunning, Is.False);
+            Assert.That(source.IsRunning, Is.False);
+
+            decouplingClock.Start();
+            decouplingClock.ProcessFrame();
+            Assert.That(decouplingClock.IsRunning, Is.True);
+            Assert.That(source.IsRunning, Is.True);
         }
 
         [Test]
--- a/osu.Framework/Timing/DecouplingFramedClock.cs
+++ b/osu.Framework/Timing/DecouplingFramedClock.cs
@@ -125,6 +125,7 @@ namespace osu.Framework.Timing
                 {
                     adjustableSourceClock.Seek(currentTime);
                     adjustableSourceClock.Start();
+                    lastSeekFailed = false;
 
                     // Don't use the source's time until next frame, as our decoupled time is likely more accurate
                     // (starting a clock, especially a TrackBass may have slight discrepancies).


bdach:
Initially, this fix bothered me because I was worried that this `lastSeekFailed` flag could be wrong in some other scenario and prevent a start, too. But after some experimentation, I'm *probably* okay with it.

For the record, I tried to see if I can apply something like

```diff
diff --git a/osu.Framework/Timing/DecouplingFramedClock.cs b/osu.Framework/Timing/DecouplingFramedClock.cs
index e4d248f4b..e7d6304ed 100644
--- a/osu.Framework/Timing/DecouplingFramedClock.cs
+++ b/osu.Framework/Timing/DecouplingFramedClock.cs
@@ -62,10 +62,11 @@ public sealed class DecouplingFramedClock : ISourceChangeableClock, IAdjustableC
         private double? lastReferenceTime;
 
         /// <summary>
-        /// Whether the last <see cref="Seek"/> operation failed.
+        /// The last time instant given to <see cref="Seek"/> to which the source clock failed to seek.
         /// This denotes that we need to <see cref="Start"/> in decoupled mode (if possible).
         /// </summary>
-        private bool lastSeekFailed;
+        private double? lastFailedSeekPosition;
+        //private bool lastSeekFailed;
 
         /// <summary>
         /// This clock is used when we are decoupling from the source.
@@ -89,6 +90,12 @@ public void ProcessFrame()
 
             try
             {
+                if (lastFailedSeekPosition != null && lastFailedSeekPosition != Source.CurrentTime)
+                {
+                    // source moved, reset last failed seek flag.
+                    lastFailedSeekPosition = null;
+                }
+
                 // If the source is running, there is never a need for any decoupling logic.
                 if (Source.IsRunning)
                 {
@@ -125,7 +132,6 @@ public void ProcessFrame()
                 {
                     adjustableSourceClock.Seek(currentTime);
                     adjustableSourceClock.Start();
-                    lastSeekFailed = false;
 
                     // Don't use the source's time until next frame, as our decoupled time is likely more accurate
                     // (starting a clock, especially a TrackBass may have slight discrepancies).
@@ -164,7 +170,7 @@ public void Reset()
         {
             adjustableSourceClock.Reset();
             isRunning = false;
-            lastSeekFailed = false;
+            lastFailedSeekPosition = null;
             currentTime = 0;
         }
 
@@ -175,7 +181,7 @@ public void Start()
 
             // If the previous seek failed, avoid calling `Start` on the source clock.
             // Doing so would potentially cause it to start from an incorrect location (ie. 0 in the case where we are tracking negative time).
-            if (lastSeekFailed && AllowDecoupling)
+            if (lastFailedSeekPosition != null && AllowDecoupling)
             {
                 isRunning = true;
                 return;
@@ -193,7 +199,7 @@ public void Stop()
 
         public bool Seek(double position)
         {
-            lastSeekFailed = !adjustableSourceClock.Seek(position);
+            bool lastSeekFailed = !adjustableSourceClock.Seek(position);
 
             if (!lastSeekFailed)
             {
@@ -204,6 +210,8 @@ public bool Seek(double position)
             }
             else
             {
+                lastFailedSeekPosition = position;
+
                 if (!AllowDecoupling)
                     return false;
 

```


i.e. rather than track failed seeks as a binary, track the _instant_ to which the seek has failed and clear it if the source clock has since moved. But I dunno, I'm not sure it's that much clearer / better than what you have.

As a total aside, can we rename the internal `isRunning` flag to something like `shouldBeRunning` or `isRunningInternal` or something? I had a total :exploding_head: moment when attempting to track `isRunning` and `IsRunning` and (re-)discovering that one is not just a front / backing field for the other.

peppy:
> As a total aside, can we rename the internal isRunning flag to something like shouldBeRunning or

Done

PR #6019, opened 2023-10-05
Fix incorrect implementation of `TrackVirtual`

peppy:
Noticed when going through osu! side test failures.

Also adds safeties and test coverage to `DecouplingFramedClock` against a source which may still start after a failed seek. Better safe than sorry I guess.

Note that the test doesn't fail even without the `TrackVirtual` fix, as it's not using a `TrackVirtual`.

### DIFF
--- a/osu.Framework/Timing/DecouplingFramedClock.cs
+++ b/osu.Framework/Timing/DecouplingFramedClock.cs
@@ -124,11 +124,9 @@ namespace osu.Framework.Timing
                 if (lastTime < 0 && currentTime >= 0)
                 {
                     // We still need to check the seek was successful, else we might have already exceeded valid length of the source.
-                    if (adjustableSourceClock.Seek(currentTime))
-                    {
+                    lastSeekFailed = !adjustableSourceClock.Seek(currentTime);
+                    if (!lastSeekFailed)
                         adjustableSourceClock.Start();
-                        lastSeekFailed = false;
-                    }
 
                     // Don't use the source's time until next frame, as our decoupled time is likely more accurate
                     // (starting a clock, especially a TrackBass may have slight discrepancies).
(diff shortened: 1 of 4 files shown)


bdach:
The `TrackVirtual`  behaviour changes look sane. I've added test coverage of them in https://github.com/ppy/osu-framework/pull/6019/commits/6b4471cdee4bbd9011d467bb55b195b896baf5bf; please double-check that the tests look as you'd expect them to.

osu.Framework/Timing/DecouplingFramedClock.cs
```diff
@@ -123,9 +123,12 @@ public void ProcessFrame()
...
                 {
-                    adjustableSourceClock.Seek(currentTime);
-                    adjustableSourceClock.Start();
-                    lastSeekFailed = false;
+                    // We still need to check the seek was successful, else we might have already exceeded valid length of the source.
+                    if (adjustableSourceClock.Seek(currentTime))
+                    {
+                        adjustableSourceClock.Start();
+                        lastSeekFailed = false;
+                    }
```
bdach, line 131: I've been thinking that maybe something like this should be here but decided to let it go...

Now that we're actually doing it, I wonder what should happen to the `lastSeekFailed` flag if this seek _fails_? Maybe this should read closer to

```diff
diff --git a/osu.Framework/Timing/DecouplingFramedClock.cs b/osu.Framework/Timing/DecouplingFramedClock.cs
index 1ba4a1c78..9a96cc4b4 100644
--- a/osu.Framework/Timing/DecouplingFramedClock.cs
+++ b/osu.Framework/Timing/DecouplingFramedClock.cs
@@ -124,11 +124,9 @@ public void ProcessFrame()
                 if (lastTime < 0 && currentTime >= 0)
                 {
                     // We still need to check the seek was successful, else we might have already exceeded valid length of the source.
-                    if (adjustableSourceClock.Seek(currentTime))
-                    {
+                    lastSeekFailed = !adjustableSourceClock.Seek(currentTime);
+                    if (!lastSeekFailed)
                         adjustableSourceClock.Start();
-                        lastSeekFailed = false;
-                    }
 
                     // Don't use the source's time until next frame, as our decoupled time is likely more accurate
                     // (starting a clock, especially a TrackBass may have slight discrepancies).

```

Diff above appears to pass framework tests, for whatever that's worth.

peppy: Looking fine, have applied.

bdach (approved):

PR #6020, opened 2023-10-05
Fix `InterpolatingFramedClock` not correctly handling seeks when source clock is stopped

peppy:
I'm crossing my fingers that at face value this one is understandable. I'll add inline review notes as to why the changes were made.

### DIFF
--- a/osu.Framework/Timing/InterpolatingFramedClock.cs
+++ b/osu.Framework/Timing/InterpolatingFramedClock.cs
@@ -32,7 +32,7 @@ namespace osu.Framework.Timing
 
         public virtual double Rate => FramedSourceClock.Rate;
 
-        public virtual bool IsRunning => sourceIsRunning;
+        public virtual bool IsRunning { get; private set; }
 
         public virtual double ElapsedFrameTime { get; private set; }
 
@@ -40,13 +40,9 @@ namespace osu.Framework.Timing
 
         protected IFrameBasedClock FramedSourceClock;
 
-        public virtual double CurrentTime => currentTime;
+        public virtual double CurrentTime { get; private set; }
 
         private readonly FramedClock realtimeClock = new FramedClock(new StopwatchClock(true));
-
-        private double currentInterpolatedTime;
-
-        private bool sourceIsRunning;
 
         private double currentTime;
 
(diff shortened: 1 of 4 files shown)


peppy:
Hold off on this momentarily while I cross-check this against osu! side (I think while the fix here is correct, it doesn't actually fix the original issue I was attempting to resolve).

peppy:
@bdach when you get a chance, can you give this 10 minutes on the watch and tell me on a scale of 0 to 10 how difficult these changes are to parse? I'll provide ample documentation based on the answer.

Of note, [9a95f19](https://github.com/ppy/osu-framework/pull/6020/commits/9a95f19dc41d9fa24c4f8bf305917171c70c32bb) finally fixed the issue I was encountering (fixing https://github.com/ppy/osu/runs/17415814918#r0s0), but I don't know how "yeah that makes common sense" of a change it is.

bdach:
On a _very_ rapid check, https://github.com/ppy/osu-framework/pull/6020/commits/a4aca0fd2dfd38f3673f4af37059954a4f0a7189, https://github.com/ppy/osu-framework/pull/6020/commits/2a6b28dafd0119b1653e4afd785d91854839d076, and https://github.com/ppy/osu-framework/pull/6020/commits/9a95f19dc41d9fa24c4f8bf305917171c70c32bb intuitively do seem to make sense to me. The main qualm for me with those is that they appear to be using somewhat heuristic methods to detect seeks that I cannot immediately ascertain the complete correctness of (stuff like "if source stopped and elapsed then it must have seeked" - I don't immediately know this to be true).

https://github.com/ppy/osu-framework/pull/6020/commits/237c9f87bfc22ccd4062f46e566a7b76de9c1e2e mostly flies over my head though. I _vaguely_ see how it could change stuff but a failure case explanation there would probably help.

peppy:
> stuff like "if source stopped and elapsed then it must have seeked" 

With the direction we've taken framed clocks, it's guaranteed that on a `ProcessFrame`, `ElapsedFrameTime` will be the precise difference since the last call. Does that make things easier to parse?

peppy:
> https://github.com/ppy/osu-framework/commit/237c9f87bfc22ccd4062f46e566a7b76de9c1e2e mostly flies over my head though. I vaguely see how it could change stuff but a failure case explanation there would probably help.

Working through the failing test in the same commit, the scenario is this:

- source is started
- `InterpolatingClock` enters a state of interpolation with: `InterpolatingClockTest+TestInterpolatingFramedClock: curr 10 elap 10 running True`
- source clock is stopped
- source clock is seeked
- `InterpolatingClock.ProcessFrame` results in : `InterpolatingClockTest+TestInterpolatingFramedClock: curr -10000 elap 0 running False`

Expectation is that the difference in time is `-10010` (roughly `-10000` as per test lenience) but it is actual `0`.

This is due to the fact that `resetInterpolation` reset `lastInterpolatedTime` and `currentTime`, but `currentTime` was subsequently updated to the source time, causing an incorrect elapsed time.

I chose to implement `ElapsedFrameTime` similar to `DecouplingClock`, where it is a stand-alone final step, rather than weirdly embedded in the interpolation calculations.

osu.Framework/Timing/InterpolatingFramedClock.cs
```diff
@@ -101,18 +101,34 @@ public virtual void ProcessFrame()
...
-                    currentInterpolatedTime = Rate >= 0 ? Math.Max(lastInterpolatedTime, currentInterpolatedTime) : Math.Min(lastInterpolatedTime, currentInterpolatedTime);
+                    currentInterpolatedTime = Rate >= 0 ? Math.Max(timeBefore, currentInterpolatedTime) : Math.Min(timeBefore, currentInterpolatedTime);
                 }
+
+                currentTime = currentInterpolatedTime;
+
+                // Of importance, only start interpolating from the next frame.
+                // The first frame after a clock starts may give very incorrect results, ie. due to a seek in the frame before.
+                if (sourceHasElapsed)
+                    IsInterpolating = true;
```
bdach, line 112: It took me probably the better part of an hour and pen & paper, but I might have found an issue here.

Consider what happens if the following branch is taken above:

https://github.com/ppy/osu-framework/blob/237c9f87bfc22ccd4062f46e566a7b76de9c1e2e/osu.Framework/Timing/InterpolatingFramedClock.cs#L89

due to exceeding `AllowableErrorMilliseconds`. Therein, we have

https://github.com/ppy/osu-framework/blob/237c9f87bfc22ccd4062f46e566a7b76de9c1e2e/osu.Framework/Timing/InterpolatingFramedClock.cs#L95-L96

But it's not infeasible that after taking that branch and resetting `IsInterpolating` to `false`, this `sourceHasElapsed` branch will _also_ be taken, undoing the effects of the interpolation disable. After all, interpolation desyncs, when they happen, generally happen when the source _is_ elapsing.

This should likely not be running in the "allowance exceeded" case. I have no idea how you test this, however.

For what it's worth, I only found this because I wanted to holistically read the entirety of `ProcessFrame()` and simulate the clock's workings step by step, rather than considering the individual fixes in isolation as before. Other than this issue, generally, I'm now somewhat confident that this new logic (a) does interpolation and (b) is more resilient against doing interpolation across seeks / source changes.

peppy: Yeah, this is completely correct. A bad oversight from my end.

peppy: I've possibly rewritten the method to possibly read better, and also fix the issue you mentioned. Touched on in the commit message, but I took inspiration from `DecouplingClock`'s structure to try and standardise things a bit.

I also added a failing test for something around the scenario you mentioned.

osu.Framework/Timing/InterpolatingFramedClock.cs
```diff
@@ -61,58 +55,83 @@ public InterpolatingFramedClock(IFrameBasedClock? source = null)
...
+                    currentTime += (FramedSourceClock.CurrentTime - currentTime) / 8;
+
+                    bool withinAllowableError = Math.Abs(FramedSourceClock.CurrentTime - currentTime) <= AllowableErrorMilliseconds;
+
+                    if (!withinAllowableError)
+                    {
+                        // if we've exceeded the allowable error, we should use the source clock's time value.
+                        IsInterpolating = false;
+                        currentTime = FramedSourceClock.CurrentTime;
+                    }
```
bdach, line 112: I dunno...

This check for allowable error used to check the difference taking *only* the realtime clock into account, and disregarding the difference between source and interpolated time. Now it checks it _after_ compensating for the difference between source and interpolated. This makes this new implementation potentially more lenient.

osu.Framework/Timing/InterpolatingFramedClock.cs
```diff
@@ -61,58 +55,83 @@ public InterpolatingFramedClock(IFrameBasedClock? source = null)
...
 
-        private void resetInterpolation()
-        {
-            currentTime = 0;
-            lastInterpolatedTime = 0;
-            currentInterpolatedTime = 0;
+                // seeking backwards should only be allowed if the source is explicitly doing that.
+                bool elapsedInOpposingDirection = FramedSourceClock.ElapsedFrameTime != 0 && Math.Sign(FramedSourceClock.ElapsedFrameTime) != Math.Sign(Rate);
+                if (!elapsedInOpposingDirection)
+                    currentTime = Rate >= 0 ? Math.Max(lastTime, currentTime) : Math.Min(lastTime, currentTime);
```
bdach, line 127: The fact that this is now apparently applied unconditionally as long as the source is running, regardless of whether interpolation is happening or not, also weirds me out. I'm not even sure I can fully grasp the consequences without like, probably mapping out all possible code paths that can feasibly be taken before this change and after and making sure no execution branch is broken by doing this.

bdach (commented):
I dunno. I found 2 major implementation differences from a rough read-through. And that's not even without properly going through this with a fine-toothed comb.

I'll continue if the changes are intended, but I dunno.

osu.Framework/Timing/InterpolatingFramedClock.cs
peppy: Yep, intentional.

Can revert if you disagree.

osu.Framework/Timing/InterpolatingFramedClock.cs
peppy: Again, this will never be an issue if the source clock is implemented correctly. Because `ElaspsedTime` should always match difference between frames. But it feels safer to alwys run it.

bdach (approved):
Alright well I *think* this is okay, if the changes pointed out above are intentional.

~~Disclaimer that I have not run this against game-side locally, I'm assuming slash hoping that you did. Not sure if you want more pairs of eyes on this either.~~ As per on stream conversation game-side tests were run and one review is considered enough so I'm doing it.

PR #6021, opened 2023-10-08
Implement `EdgeEffectParameters` for `CircularProgress`

EVAST9919:
Okay, I've been sitting for too long on this one, but we need to start at least somewhere.

### **Overview**
All the properties of `EdgeEffectParameters` are implemented and work as expected, except `Roundness`, which we talk about later. Glowing part is drawn as a separate pass with a separate shader to allow offset and different blending modes depending on the type of the effect (glow/shadow).

I left 2 methods for glow in the shader which are `getGlow` and `getBlur` in case one is more preferable over the other. Although I personally like first more, and it's falloff curve can be adjusted further if needed.

| glow | blur |
| --- | --- |
| glow | blur |

### **Main issue**
The strength of the glow depends on the distance of the pixel from the shape, which results in unwanted behaviour near sharp edges or concave areas.
Снимок экрана (44)
This is a side effect of SDFs. To resolve that we would need to have a function to calculate a distance to a "smoother shape" in which `Roundness` property of `EdgeEffectParameters` will have effect. I have a working solution for a progress with rounded caps enabled, however math there is overly complex and for sharp edges it's even crazier. Let's leave it for another day/pr.


EVAST9919:
With this pr and added support of multiple `EdgeEffectParameters` (that'd be a separate pr) this is how spinner will look like:
Снимок экрана (54)

osu.Framework/Resources/Shaders/sh_CircularProgressEdgeEffect.fs
```diff
@@ -0,0 +1,78 @@
...
+lowp float getGlowDebug(highp float distance, highp float size)
+{
+    if (distance < 0.0)
+        return 1.0;
+
+    if (distance < size)
+        return 0.5;
+
+    return 0.0;
+}
```
smoogipoo, line 54: Did you intend to keep this function around?

EVAST9919: I thought someone would need it to ensure correct draw size or whatever. But can be safely removed if that's not the case.

smoogipoo:
I took a look at this with @Tom94 's guidance...

The first is a suggestion: have you tried lerping in the SDF for the full circle once the end points get close, to reduce the effect of the wedge between them?:

image

The second is and issue, with non-rounded caps and offset glow, this doesn't look correct:

image

EVAST9919:
> The first is a suggestion: have you tried lerping in the SDF for the full circle once the end points get close, to reduce the effect of the wedge between them?

It's hard to define what's considered "close". With `fill = 1` edges are intersecting at the centre, so they are already "close" with `progress = 0.5` and with fill approaching 0 they are close only with progress approaching 1. And this whole structure should depend on glow size as well: with bigger glow size this "closure" should be different. A mess really. But I'll try, may be it won't be as difficult to figure out.

> The second is and issue, with non-rounded caps and offset glow, this doesn't look correct:

Yeah, that's because current SDF implementation isn't correct by itself: it can't correctly calculate distance inside the shape near the edges, so it needs a rewrite (which I was trying to avoid in this pr tbh, but I guess we need that here)

EVAST9919:
Okay, so as cool as it sounds in theory and much as I'd like it to work - there are too much edge cases to cover to be considered as a viable replacement for a proper blur/glow. And even with potentially all of them covered - that's too much for a single use case. And I'm not even taking into account that spinner has multiple glows on top of each other which should be blended multiplicatively between each other but rendered additively at the end, so in perfect world would require additional frame buffer and at this point the overhead will be not far off from the buffered contaner.

IMO the easiest way to achieve spinner glow would be a prebaked texture with the glow and a small sampling shader osu-side, which I'll try to implement later.

However it seems like the circularProgress shader still needs to be reworked to support edge smoothing as required by [the new design of volume sliders](https://www.figma.com/file/Oqr4fuKKCGfBtzzUP8Qdis/Volume-Slider?type=design&node-id=0-1&mode=design&t=RgMOFPwn8ac5MSDF-0), which I have a proper version of and will be pr'd soon as well.

PR #6022, opened 2023-10-10
Add button to randomise test scene background colour

peppy:
I got sick of doing this manually.

In the future I'd like to expand this to probably be able to persist across hot reloads of the same test scene.

https://github.com/ppy/osu-framework/assets/191335/560b8924-f5d5-4cc9-b3f1-1bc5e8dbb7b0


smoogipoo (approved):
This is fine by me. @bdach gonna do a quick pass over you to too see if you have any thoughts :)

bdach (approved):
Don't really see a reason not to I guess

PR #6023, opened 2023-10-10
Add notes regarding using inline delegates with `Scheduler.AddOnce`

peppy:


osu.Framework.Tests/Threading/SchedulerTest.cs
```diff
@@ -429,6 +429,8 @@ public void TestAddOnceInlineDelegate()
         {
             classInvocations = 0;
 
+            // Note that while this works here (even with a capture), there's no guarantee that will always be the
+            // case. As such it's always best to use a local function or private method.
```
bdach, line 433: I'm... surprised that this works...

In this case the compiler has _somehow_ decided that it can correctly reduce the inline delegate to a method group:



For comparison, an _actual_ method group:



Wizardry I tell ya...

bdach (approved):

PR #6024, opened 2023-10-12
FFmpeg: Enable MPEG-4 parser and decoder

FreezyLemon:
Closes ppy/osu#25086.

[MPEG-4 Part 2](https://en.wikipedia.org/wiki/MPEG-4_Part_2) is a video coding standard that sits "between" H.262 and H.264. Xvid, DivX and some less known formats implement it.

FFmpeg supports these derivative codecs through the `mpeg4` decoder, so to support `xvid` and `divx` (which seems to have been used in some beatmaps), this PR enables it and the according parser.

Note that this does not depend on `libxvid`, which is GPL.

### DIFF
--- a/.github/workflows/build-ffmpeg.yml
+++ b/.github/workflows/build-ffmpeg.yml
@@ -69,8 +69,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          sudo apt-get update && sudo apt-get upgrade
-          sudo apt-get install -y make nasm gcc mingw-w64
+          sudo apt-get update
+          sudo apt-get install nasm mingw-w64
 
       - name: Build
         run: osu.Framework.NativeLibs/scripts/ffmpeg/build-win.sh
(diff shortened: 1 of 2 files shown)


FreezyLemon:
Sorry, yeah they are: https://github.com/FreezyLemon/osu-framework/actions/runs/6499895277

Had to remove `apt-get upgrade` in the CI script, apparently you're not really supposed to do that.

FreezyLemon:
Noticed another beatmap that doesn't work while testing: https://osu.ppy.sh/beatmapsets/10435#osu/42152

It uses "Sorenson Spark" video. I'll investigate and add another codec, I guess.

FreezyLemon:
New build artifacts, including the FLV/H.263/Sorenson Spark decoder: https://github.com/FreezyLemon/osu-framework/actions/runs/6500576741

peppy (approved):

PR #6027, opened 2023-10-13
Update native libs

peppy:
Pulls in new (restored) video codecs.


bdach (approved):
Tested to work on windows and linux.

PR #6031, opened 2023-10-17
Bring back shader workaround to fix blur on iOS

frenzibyte:
- Closes https://github.com/ppy/osu/issues/24960
- Reverts https://github.com/ppy/osu-framework/pull/5979

I could not investigate this as I failed to repro in local builds on neither my iPad nor my phone, but I have confirmed at least that the blur breakage only occurred on the next working release after the [workaround removal PR](https://github.com/ppy/osu-framework/pull/5979) was merged, so I'm going ahead with reverting that for now.

I've made sure at least that blurring still works on all of my devices regardless of its state previously.

### DIFF
--- a/osu.Framework/Resources/Shaders/sh_Blur.fs
+++ b/osu.Framework/Resources/Shaders/sh_Blur.fs
@@ -44,7 +44,12 @@ lowp vec4 blur(int radius, highp vec2 direction, mediump vec2 texCoord, mediump 
 			break;
 	}
 
-	return sum / totalFactor;
+	// this is supposed to be unnecessary with https://github.com/ppy/veldrid-spirv/pull/4,
+	// but blurring is still broken on some Apple devices when removing it (at least on an M2 iPad Pro and an iPhone 12).
+	// todo: investigate this.
+	float one = g_BackbufferDraw ? 0 : 1;
+
+	return sum / totalFactor * one;
 }
 
 void main(void)


osu.Framework/Resources/Shaders/sh_Blur.fs
```diff
@@ -44,7 +44,12 @@ lowp vec4 blur(int radius, highp vec2 direction, mediump vec2 texCoord, mediump
 			break;
 	}
 
-	return sum / totalFactor;
+	// this is supposed to be unnecessary with https://github.com/ppy/veldrid-spirv/pull/4,
+	// but blurring is still broken on some Apple devices when removing it (at least on an M2 iPad Pro and an iPhone 12).
```
frenzibyte, line 48: Went ahead with mentioning the devices that I see blurring broke with, as it was mentioned in [discord](https://discord.com/channels/188630481301012481/589331078574112768/1156929941355642900) that there was at least one Apple device that doesn't experience this.

peppy: You mention here the two devices you mentioned you failed to repro on, what's the deal there?

frenzibyte: I miswrote the description, I meant that I could not repro on both of those devices in local builds, but I witnessed the bug on both of them in release builds.

smoogipoo: Doesn't that sound like a packaging issue then? As if an old version of veldrid-spirv is being used that doesn't contain https://github.com/ppy/veldrid-spirv/commit/3f6bee281eb9d3e86ec6b83d6bd5a8ecb4953113

peppy: Yeah, sounds likely the case. I think it's fine to get this in with the existing comment though. We pretty much know the cause here.

peppy (approved):

smoogipoo (approved):

osu.Framework/Resources/Shaders/sh_Blur.fs
frenzibyte: I am not sure that is the case. As I mentioned above, blurring still works with a local build both in debug and release configuration (yes, with AOT).

peppy: I'm not sure how much it matters at this point, but did you check the checksums as above? I've run into this before and a full clean was required (potentially with manual `obj` cleanup) to fix.

PR #6032, opened 2023-10-17
Disable retrieval of SDL display modes on non-desktop platforms

frenzibyte:
Currently, SDL provides two display modes for an iPad, both of which resemble the same display's resolution, except that the second one is the portrait version of the first one.

IMG_9F4E6B2C62C4-1

For simplicity, let's not provide any display modes in such case.

### DIFF
--- a/osu.Framework/Platform/SDL2Window_Windowing.cs
+++ b/osu.Framework/Platform/SDL2Window_Windowing.cs
@@ -350,30 +350,30 @@ namespace osu.Framework.Platform
                 return false;
             }
 
-            int numModes = SDL.SDL_GetNumDisplayModes(displayIndex);
+            DisplayMode[] displayModes = Array.Empty<DisplayMode>();
 
-            if (numModes < 0)
+            if (RuntimeInfo.IsDesktop)
             {
-                Logger.Log($"Failed to get display modes for display at index ({displayIndex}) ({rect.w}x{rect.h}). SDL Error: {SDL.SDL_GetError()} ({numModes})");
-                display = null;
-                return false;
-            }
+                int numModes = SDL.SDL_GetNumDisplayModes(displayIndex);
 
-            if (numModes == 0) Logger.Log($"Display at index ({displayIndex}) ({rect.w}x{rect.h}) has no display modes. Fullscreen might not work.");
+                if (numModes < 0)
+                {
+                    Logger.Log($"Failed to get display modes for display at index ({displayIndex}) ({rect.w}x{rect.h}). SDL Error: {SDL.SDL_GetError()} ({numModes})");
+                    display = null;
+                    return false;
+                }
 
-            var displayModes = Enumerable.Range(0, numModes)
+                if (numModes == 0)
+                    Logger.Log($"Display at index ({displayIndex}) ({rect.w}x{rect.h}) has no display modes. Fullscreen might not work.");
+
+                displayModes = Enumerable.Range(0, numModes)
                                          .Select(modeIndex =>
                                          {
                                              SDL.SDL_GetDisplayMode(displayIndex, modeIndex, out var mode);
                                              return mode.ToDisplayMode(displayIndex);
                                          })
                                          .ToArray();
-
-            // on an iPad, SDL returns a display mode for both portrait and landscape orientations,
-            // but we don't support having both of them in the display modes array (since display modes are usually treated as different resolutions).
-            // for simplicity, do not provide any display modes on iOS regardless of SDL.
-            if (RuntimeInfo.OS == RuntimeInfo.Platform.iOS)
-                displayModes = Array.Empty<DisplayMode>();
+            }
 
             display = new Display(displayIndex, SDL.SDL_GetDisplayName(displayIndex), new Rectangle(rect.x, rect.y, rect.w, rect.h), displayModes);
             return true;
--- a/osu.Framework/Platform/SDL2Window_Windowing.cs
+++ b/osu.Framework/Platform/SDL2Window_Windowing.cs
@@ -369,6 +369,12 @@ namespace osu.Framework.Platform
                                          })
                                          .ToArray();
 
+            // on an iPad, SDL returns a display mode for both portrait and landscape orientations,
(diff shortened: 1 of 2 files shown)


osu.Framework/Platform/SDL2Window_Windowing.cs
```diff
@@ -369,6 +369,12 @@ private static bool tryGetDisplayFromSDL(int displayIndex, [NotNullWhen(true)] o
                                          })
                                          .ToArray();
 
+            // on an iPad, SDL returns a display mode for both portrait and landscape orientations,
+            // but we don't support having both of them in the display modes array (since display modes are usually treated as different resolutions).
+            // for simplicity, do not provide any display modes on iOS regardless of SDL.
+            if (RuntimeInfo.OS == RuntimeInfo.Platform.iOS)
+                displayModes = Array.Empty<DisplayMode>();
```
smoogipoo, line 376: It sounds like we don't want to display resolution settings on mobile platforms period. Is there a reason Android isn't included here?

frenzibyte: Fair point. Need someone with an Android device to check this one and see if they get multiple display modes similar to what's in the PR description above.

bdach: There is no dropdown on my Android test device on the latest release.



That said I don't suppose extending this  check to cover it would do any harm.

peppy: My only apprehension would be that android devices may not just be phones. And some may correctly handle multiple resolutions. Hard to say.

Susko3:
Does the portrait resolution work as expected, does it actually rotate the display to portrait and show the game fine?

If it does work fine, I don't see a reason to arbitrarily limit the resolutions/display rotation on mobile platforms. If osu! doesn't work that well on portrait on mobile, hide the dropdown in osu! only.

frenzibyte:
> Does the portrait resolution work as expected, does it actually rotate the display to portrait and show the game fine?
> 
> If it does work fine, I don't see a reason to arbitrarily limit the resolutions/display rotation on mobile platforms. If osu! doesn't work that well on portrait on mobile, hide the dropdown in osu! only.

Nope, it does not function at all.

Susko3:
> it does not function at all.

Probably because of the iOS specific code;

https://github.com/ppy/osu-framework/blob/90d9286f8b209ef9a58620f89875c258702d1732/osu.Framework.iOS/IOSWindow.cs#L37-L43

The requested display mode / resolution is completely ignored. I'd consider this a bug, one that seems simple to fix.

frenzibyte:
It's certainly not simple, since it's intentionally ignored for a reason (see https://github.com/ppy/osu-framework/pull/5698#issuecomment-1482580925, which you should find easily with blame). 

Calling the base method has no effect either (can confirm it calls `SDL_SetWindowDisplayMode` with the specified display mode), and honestly, I'm not sure what sense it makes to provide display modes as a method of allowing to change device orientation, so I would rather hide it from here rather than changing it osu!-side.

Susko3:
>Calling the base method has no effect either (can confirm it calls SDL_SetWindowDisplayMode with the specified display mode)

Weird. I guess SDL expects using something like [`SDL_HINT_ORIENTATIONS`](https://github.com/libsdl-org/SDL/blob/bf269571fc484afe5155f2f5e5f269dd328ce6f6/include/SDL3/SDL_hints.h#L1303-L1312) for that. The [SDL logic that selects the orientations](https://github.com/libsdl-org/SDL/blob/bf269571fc484afe5155f2f5e5f269dd328ce6f6/src/video/uikit/SDL_uikitwindow.m#L410) also seems really complex. Has it been tested that orientations work as specified in the [Info.plist](https://github.com/ppy/osu-framework/blob/90d9286f8b209ef9a58620f89875c258702d1732/osu.Framework.Tests.iOS/Info.plist#L29-L34) file — that SDL isn't interfering here?

>I'm not sure what sense it makes to provide display modes as a method of allowing to change device orientation, so I would rather hide it from here rather than changing it osu!-side.

Good point, supported orientations are specified in the manifest files on mobile. And the expected UX for changing to a supported orientation is to just rotate the device.

osu.Framework/Platform/SDL2Window_Windowing.cs
Susko3: I read trough a bunch of Android and ChromeOS docs. All they really mention is that the user can resize the window, with no mention of the application being able to programmatically set the size.

The only thing that comes close is being able to [set an initial size when the game is first run](https://chromeos.dev/en/android/window-management#initial-launch-size).

So 👍 for hiding display modes on all mobile platforms.

frenzibyte:
In response to https://github.com/ppy/osu-framework/pull/6032#discussion_r1364749906, I've disabled retrieval of SDL display modes completely on non-desktop platforms. @Susko3 or anyone else with an Android device, please check and confirm the game is still behaving correctly with this change.

Susko3:
Android doesn't use SDL, so there's nothing to test 🙂

osu.Framework/Platform/SDL2Window_Windowing.cs
```diff
@@ -350,24 +350,30 @@ private static bool tryGetDisplayFromSDL(int displayIndex, [NotNullWhen(true)] o
                 return false;
             }
 
-            int numModes = SDL.SDL_GetNumDisplayModes(displayIndex);
+            DisplayMode[] displayModes = Array.Empty<DisplayMode>();
 
-            if (numModes < 0)
+            if (RuntimeInfo.IsDesktop)
```
Susko3, line 355: Note: since the code is now relevant only on desktop, it could be moved to `SDL2DesktopWindow`.
This isn't the only usage of `RuntimeInfo.Is{Desktop,Mobile}` in this class, so it's fine as is. Better to keep the display fetching code all together.

Susko3 (approved):

peppy:
Not sure. I doubt this is correct but let's go with it and revert when we need to.

peppy (approved):

Issue #3687, opened 2020-07-05
Pushing a screen at OnExiting without blocking the exit freezes the UI


mcendu:
Per title. The following test if put in `TestSceneScreenStack` triggers it. I am not sure what is the intended behavior here.
```c#
        [Test]
        public void TestPushOnExitingWithoutBlocking()
        {
            TestScreen screen1 = null;

            pushAndEnsureCurrent(() =>
            {
                screen1 = new TestScreen(id: 1);
                screen1.Exiting = () =>
                {
                    screen1.Push(new TestScreen(id: 2));
                    return false;
                };
                return screen1;
            });

            AddStep("Exit screen 1", () => screen1.Exit());
        }
```

peppy:
It's undefined behaviour (you're pushing a new screen to a screen that you've implied to be exited).

So let me ask you then: what do you expect in this scenario? what are you trying to make happen here? the new screen pushed to the parent in the stack?

mcendu:
Ok

peppy:
This shouldn't be closed (it does need a resolution, even if that is just throwing an exception). Could you answer my question to help us decide how this behaves or offer an alternative?

mcendu:
I think it is good to throw an exception here; nobody is going to expect such a thing to happen IMO.

PR #6033, opened 2023-10-22
Throw exception when pushing a screen at OnExiting without blocking

hwabis:
Fixes https://github.com/ppy/osu-framework/issues/3687


smoogipoo (approved):
I think this is fine. Surrounding code is weird so requesting second review.

peppy (approved):
Ran against osu! tests just to make sure, looks good.

PR #6035, opened 2023-10-26
Correctly handle OTD "multi-reports"

Susko3:
Noticed when looking into https://github.com/ppy/osu/issues/5378#issuecomment-1780613477.

Since a single OTD `IDeviceReport` can implement [multiple report interfaces](https://github.com/OpenTabletDriver/OpenTabletDriver/blob/b996c8feb0b3da53ff0599940af550118f6422ca/OpenTabletDriver.Configurations/Parsers/Wacom/Bamboo/BambooTabletReport.cs#L7), a `switch` would handle only one of the report types.


smoogipoo:
Maybe this is why the bamboo tablet in the office wasn't working. @peppy can you test?

Susko3:
This should only fix aux buttons on the tablet not working.

smoogipoo (approved):

peppy (approved):

Issue #5788, opened 2023-05-13
Releasing mouse outside of window does not end drag state


Gyoshi:
Tested in both osu! and osu.Framework.Tests.exe, on Desktop (windows 11)

In windowed and borderless windowed, if you drag but release mouse button outside of window:
Mouse stays lit up (in osu!) and Drawables being interacted with remain active until clicking and releasing again *inside* the window.

bdach:
Details on which platform this was tested on would be helpful.

Gyoshi:
Windows 11

hwabis:
Would the correct behavior be for the cursor in o!f to be released as soon as the mouse leaves the window? (Or would the cursor in o!f release only when the mouse is released, even when outside the window? I'm not familiar with input handling so I don't even know if that's possible.)

peppy:
Doesn't happen on macOS, does on windows.

FreezyLemon:
This can be reproduced in TestSceneInputManager. There is a toggle for the MouseHandler's `UseRelativeMode`, and I can reproduce that with relative mode disabled, this bug does not occur. WIth relative mode enabled, it does.

It looks like it's been reported before: #4055. The PR closing it (#5163) states that it only partially fixes this behaviour, but the issue was closed anyways.

The reason for disabling this auto-capture behaviour when relative mode is enabled is explained in the code:
https://github.com/ppy/osu-framework/blob/102ae9498b215cc4eb2150a8e63c5563a58dafb3/osu.Framework/Platform/SDL2Window_Input.cs#L57-L71

FreezyLemon:
Note that I can also reproduce this on Linux.

gadielisawesome:
I can reproduce this sometimes on Android when using osu in splitscreen.

Susko3:
@gadielisawesome please open a new issue for android, with additional info and a screen recording. If you can, check how it behaves with relative mode turned on and off.

gadielisawesome:
It's been 2 months. I no longer own an Android device, so I cannot reproduce this issue.

PR #6036, opened 2023-10-26
Fix releasing mouse buttons outside of window not working in relative mode

Susko3:
- Closes https://github.com/ppy/osu-framework/issues/5788

I originally envisioned the fix to complicate the relative mouse logic further, capturing the cursor when at least one button is held down. That was too hard and not worth the effort. Instead, query the global mouse button state when it's outside the window and release any buttons that should not be held.

It's important that the buttons can only be _released_, not pressed when the cursor is outside the window. Also important (but not that much) is that the buttons are released after the move event is handled so the released events have the most up-to-date position.


peppy:
Not sure about this, but let's just get it in for testing. Proper review of these SDL nuances is too time consuming to warrant at the moment.

peppy (approved):

PR #6038, opened 2023-10-28
Allow selecting all text in a `TextBox` programmatically

frenzibyte:
- Supersedes https://github.com/ppy/osu-framework/pull/5823 (only the selection part, ignoring focus acquirement)
- To be used in https://github.com/ppy/osu/issues/24606

I've chose to be as strict as possible with the method, as to not potentially go out of control and backfire on the consumer. I went as far as to make the method throw when the text box is focused, but I lowered it down to returning `false`/`true` instead.

The remaining part of https://github.com/ppy/osu-framework/pull/5823 can be achieved by a `FocusedTextBox` class that overrides `OnFocus` and calls `SelectAll` from inside.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneTextBox.cs
@@ -3,6 +3,7 @@
 
 #nullable disable
 
+using System;
 using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Extensions;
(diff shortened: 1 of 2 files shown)


bdach:
> To be used in https://github.com/ppy/osu/issues/24606

How is this needed for that issue?

frenzibyte:
As per the title of that issue thread, this is required to add the behaviour of selecting search text box content whenever the overlay is displayed (or any search text box that gets focused).

frenzibyte:
> I guess this is going for the "select all" UX but it's not even clear that it should be the one that gets implemented?

There's some preference on that going by the description of that issue thread.

peppy:
Not really seeing why this required a new PR when it's using 99% of the code from the old one.

peppy:
I've pushed relevant changes to #5823.

frenzibyte:
> Not really seeing why this required a new PR when it's using 99% of the code from the old one.

Since the original PR was old and following a different approach, I wasn't sure what to do with it. I don't think I can take over it and push immediately, and I wasn't sure posting a comment in there would have it updated immediately and not be left over (maybe not in this particular case where you are interested in the dependant PR).

PR #6039, opened 2023-10-30
Fix wrong bytes per sample in TrackBass

hwsmm:
BASS uses 16-bit samples by default if neither BassFlags.Byte nor BassFlags.Float is specified. (https://www.un4seen.com/doc/#bass/BASS_StreamCreate.html)

You can find this out by applying the following diff:
```diff
--- a/osu.Framework/Audio/Track/TrackBass.cs
+++ b/osu.Framework/Audio/Track/TrackBass.cs
@@ -15,6 +15,7 @@
 using osu.Framework.Extensions;
 using osu.Framework.Extensions.ObjectExtensions;
 using osu.Framework.Utils;
+using osu.Framework.Logging;

 namespace osu.Framework.Audio.Track
 {
@@ -106,6 +107,9 @@ internal TrackBass(Stream data, string name, bool quick = false)

                 bool success = seconds >= 0;

+                if (Bass.ChannelGetInfo(activeStream, out var info))
+                    Logger.Log($"Track Resolution: {info.Resolution}");
+
                 if (success)
                 {
                     Length = seconds * 1000;
```

Waveform will no longer bring the constant from TrackBass since Waveform still uses float samples.

### DIFF
--- a/osu.Framework/Audio/Track/TrackBass.cs
+++ b/osu.Framework/Audio/Track/TrackBass.cs
@@ -20,8 +20,6 @@ namespace osu.Framework.Audio.Track
 {
     public sealed class TrackBass : Track, IBassAudio, IBassAudioChannel
     {
-        public const int BYTES_PER_SAMPLE = 4;
-
         private Stream? dataStream;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


bdach:
How did you notice this? Does this have any meaningful impact on anything?

hwsmm:
I could find because I have been working on #6002 which also uses BASS to decode, was just reading TrackBass and spotted this.

This PR probably has impact on nothing since the old value (4) was already correct enough since a sample frame has 2 channels and 2 bytes.

smoogipoo (approved):
Lgtm. It used to have `Float` specified, but that caused issues with macOS.

PR #6040, opened 2023-11-05
Fix `FrameStatisticsViaTouch` to work as described in the wiki

Susko3:
https://github.com/ppy/osu-framework/wiki/Environment-variables#osu_frame_statistics_via_touch

I think it makes more sense to have it as described in the wiki. A hybrid solution would be to default to enabled on debug, and disabled on release (the envvar would always override the default in the case).

Currently there's no way to disable touch frame statistics in debug builds (one can only enable them in release builds).

### DIFF
--- a/osu.Framework/FrameworkEnvironment.cs
+++ b/osu.Framework/FrameworkEnvironment.cs
@@ -23,7 +23,7 @@ namespace osu.Framework
             StartupExecutionMode = Enum.TryParse<ExecutionMode>(Environment.GetEnvironmentVariable("OSU_EXECUTION_MODE"), true, out var mode) ? mode : null;
             NoTestTimeout = Environment.GetEnvironmentVariable("OSU_TESTS_NO_TIMEOUT") == "1";
             ForceTestGC = Environment.GetEnvironmentVariable("OSU_TESTS_FORCED_GC") == "1";
-            FrameStatisticsViaTouch = Environment.GetEnvironmentVariable("OSU_FRAME_STATISTICS_VIA_TOUCH") == "1";
+            FrameStatisticsViaTouch = Environment.GetEnvironmentVariable("OSU_FRAME_STATISTICS_VIA_TOUCH") != "0";
             PreferredGraphicsSurface = Enum.TryParse<GraphicsSurfaceType>(Environment.GetEnvironmentVariable("OSU_GRAPHICS_SURFACE"), true, out var surface) ? surface : null;
             PreferredGraphicsRenderer = Environment.GetEnvironmentVariable("OSU_GRAPHICS_RENDERER")?.ToLowerInvariant();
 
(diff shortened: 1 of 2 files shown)


bdach:
Why would we want only *one of four* boolean envvars to work like this?

Susko3:
If you're commenting on the `!= "0"` vs `== "1"`, it's just flipping the default.
Do you think all boolean envvars should default to `false`?

Would having a helper function make the default more clear?
```cs
bool ParseBooleanEnvironmentVariable(string name, bool defaultValue)
```

bdach:
I guess I partially misunderstood what this was going for, but this implementation doesn't seem to match the wiki either, since this 100% should not be default-enabled in release builds, which it would be, because:

https://github.com/ppy/osu-framework/blob/0bf4656db58c5a8dc2f400cbf479dddc9bb3826e/osu.Framework/Game.cs#L269-L279

while the wiki says:

> Note that this only works in debug builds. Disable for debug builds by setting to 0.

If you're angling android for this, I'd rather just finally fix android debug builds not properly being marked as debug builds.

Susko3:
It seems you're looking at some other branch, I've changed the conditional to `&&` so it works as written on the wiki.
https://github.com/ppy/osu-framework/pull/6040/files#diff-c311d03d5213e6632087e2ddc2b3a2c8f2f2db806d33330312a387d719780445L269-R269

> If you're angling android for this, I'd rather just finally fix android debug builds not properly being marked as debug builds.

Completely forgot that `IsDebugBuild` still doesn't work on android 🙃
But I was actually annoyed by this on windows. I don't need the double tap for statistics as I can just press Ctrl+F11.

osu.Framework/FrameworkEnvironment.cs
```diff
@@ -23,7 +23,7 @@ static FrameworkEnvironment()
             StartupExecutionMode = Enum.TryParse<ExecutionMode>(Environment.GetEnvironmentVariable("OSU_EXECUTION_MODE"), true, out var mode) ? mode : null;
             NoTestTimeout = Environment.GetEnvironmentVariable("OSU_TESTS_NO_TIMEOUT") == "1";
             ForceTestGC = Environment.GetEnvironmentVariable("OSU_TESTS_FORCED_GC") == "1";
-            FrameStatisticsViaTouch = Environment.GetEnvironmentVariable("OSU_FRAME_STATISTICS_VIA_TOUCH") == "1";
+            FrameStatisticsViaTouch = Environment.GetEnvironmentVariable("OSU_FRAME_STATISTICS_VIA_TOUCH") != "0";
```
peppy, line 26: I get that this is flipping the default, but it reads really awkwardly. I'd prefer leaving the comparison as `=="1"` but adding a pre-check for empty string / null. Or something else that isn't this.

peppy (approved):
Probably okay

bdach (approved):

PR #6041, opened 2023-11-05
Fix mouse not working as expected on iOS

Susko3:
- (wrongly) reported in https://github.com/ppy/osu/issues/25368#issuecomment-1793771361

Untested.

https://github.com/libsdl-org/SDL/blob/9302d7732dfbb3c6364da142efd06ea9846206c1/include/SDL3/SDL_hints.h#L1226-L1233

### DIFF
--- a/osu.Framework/Platform/SDL2Window.cs
+++ b/osu.Framework/Platform/SDL2Window.cs
@@ -233,7 +233,8 @@ namespace osu.Framework.Platform
             SDL.SDL_SetHint(SDL.SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS, "1");
             SDL.SDL_SetHint(SDL.SDL_HINT_IME_SHOW_UI, "1");
             SDL.SDL_SetHint(SDL.SDL_HINT_MOUSE_RELATIVE_MODE_CENTER, "0");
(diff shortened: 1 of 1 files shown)


Susko3:
I am not sure why things are [not breaking](https://github.com/ppy/osu/issues/25368#issuecomment-1793794422) as much as they should without this flag, but it doesn't hurt to apply it.

frenzibyte:
I have checked on this and I can both confirm that touch events do arise from mouse input in master, and no longer do with this flag being set to zero. This looks logical to me, and makes the cursor no longer fade out abruptly when using a mouse (it's doing so on master, due to mouse input being treated as touch when clicking).

frenzibyte (approved):

Susko3:
@frenzibyte are you able to test with an Apple Pencil? A user has reported [that hover isn't working](https://github.com/ppy/osu/issues/25368#issuecomment-1793794422).

It'll also be useful to know whether pen input is treated as touch or mouse input on iOS.

frenzibyte:
Hovering has always been working for me with an Apple Pencil. Mine is second generation, not sure about the old one. 

As for event source, hovering reports mouse move events, but dragging reports touch down/move/up events. This gets slightly weird in osu! gameplay as the cursor becomes visible when hovering, which is the opposite of what one would expect. Probably needs an issue thread somewhere covering it up.

PR #6044, opened 2023-11-10
Add class for efficiently building B-Splines

default0:
This adds the IncrementalBSplineBuilder class which can be used to build up a B-Spline from a series of linear points efficiently. The class takes special care to keep the number of control points.

This is a preparatory framework side change to implement free-hand drawing of sliders in the osu!lazer editor.

# Breaking changes

## `PathApproximator` method signatures changed

The following `PathApproximator` methods have been renamed:

| old | new |
| :- | :- |
| `ApproximateBezier` | `BezierToPiecewiseLinear` |
| `ApproximateBSpline` | `BSplineToPiecewiseLinear` |
| `ApproximateCatmull` | `CatmullToPiecewiseLinear` |
| `ApproximateCircularArc` | `CircularArcToPiecewiseLinear` |
| `ApproximateLinear` | `LinearToPiecewiseLinear` |
| `ApproximateLagrangePolynomial` | `LagrangePolynomialToPiecewiseLinear` |

Additionally, `ApproximateBSpline`'s polynomial order argument `p` has been renamed to `degree` and lost its default value of 0.

### DIFF
--- a/osu.Framework.Tests/Visual/Drawables/TestScenePathApproximator.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestScenePathApproximator.cs
@@ -16,7 +16,7 @@ namespace osu.Framework.Tests.Visual.Drawables
     public partial class TestScenePathApproximator : GridTestScene
     {
         public TestScenePathApproximator()
-            : base(2, 3)
+            : base(2, 2)
         {
             Cell(0).AddRange(new[]
             {
--- a/osu.Framework.Tests/Visual/Drawables/TestScenePathApproximator.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestScenePathApproximator.cs
@@ -16,30 +16,30 @@ namespace osu.Framework.Tests.Visual.Drawables
     public partial class TestScenePathApproximator : GridTestScene
     {
         public TestScenePathApproximator()
-            : base(2, 2)
+            : base(2, 3)
         {
             Cell(0).AddRange(new[]
             {
-                createLabel("ApproximateBezier"),
-                new ApproximatedPathTest(PathApproximator.ApproximateBezier),
-            });
-
-            Cell(1).AddRange(new[]
-            {
-                createLabel("ApproximateCatmull"),
-                new ApproximatedPathTest(PathApproximator.ApproximateCatmull),
+                createLabel("BezierToPiecewiseLinear"),
+                new ApproximatedPathTest(PathApproximator.BezierToPiecewiseLinear),
             });
 
             Cell(2).AddRange(new[]
             {
-                createLabel("ApproximateCircularArc"),
-                new ApproximatedPathTest(PathApproximator.ApproximateCircularArc),
+                createLabel("CatmullToPiecewiseLinear"),
+                new ApproximatedPathTest(PathApproximator.CatmullToPiecewiseLinear),
             });
 
             Cell(3).AddRange(new[]
             {
-                createLabel("ApproximateLagrangePolynomial"),
-                new ApproximatedPathTest(PathApproximator.ApproximateLagrangePolynomial),
+                createLabel("CircularArcToPiecewiseLinear"),
+                new ApproximatedPathTest(PathApproximator.CircularArcToPiecewiseLinear),
+            });
+
+            Cell(4).AddRange(new[]
+            {
+                createLabel("LagrangePolynomialToPiecewiseLinear"),
+                new ApproximatedPathTest(PathApproximator.LagrangePolynomialToPiecewiseLinear),
             });
         }
 
@@ -50,10 +50,10 @@ namespace osu.Framework.Tests.Visual.Drawables
             Colour = Color4.White,
         };
 
+        public delegate List<Vector2> ApproximatorFunc(ReadOnlySpan<Vector2> controlPoints);
+
         private partial class ApproximatedPathTest : SmoothPath
         {
-            public delegate List<Vector2> ApproximatorFunc(ReadOnlySpan<Vector2> controlPoints);
-
             public ApproximatedPathTest(ApproximatorFunc approximator)
             {
                 Vector2[] points = new Vector2[5];
--- a/osu.Framework.Tests/Visual/Drawables/TestScenePathApproximator.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestScenePathApproximator.cs
@@ -24,19 +24,19 @@ namespace osu.Framework.Tests.Visual.Drawables
                 new ApproximatedPathTest(PathApproximator.BezierToPiecewiseLinear),
             });
 
-            Cell(2).AddRange(new[]
+            Cell(1).AddRange(new[]
             {
                 createLabel("CatmullToPiecewiseLinear"),
                 new ApproximatedPathTest(PathApproximator.CatmullToPiecewiseLinear),
             });
 
-            Cell(3).AddRange(new[]
+            Cell(2).AddRange(new[]
             {
                 createLabel("CircularArcToPiecewiseLinear"),
                 new ApproximatedPathTest(PathApproximator.CircularArcToPiecewiseLinear),
             });
 
-            Cell(4).AddRange(new[]
+            Cell(3).AddRange(new[]
             {
                 createLabel("LagrangePolynomialToPiecewiseLinear"),
                 new ApproximatedPathTest(PathApproximator.LagrangePolynomialToPiecewiseLinear),
--- a/osu.Framework.Tests/Visual/Drawables/TestScenePathApproximator.cs
+++ b/osu.Framework.Tests/Visual/Drawables/TestScenePathApproximator.cs
@@ -20,25 +20,25 @@ namespace osu.Framework.Tests.Visual.Drawables
         {
             Cell(0).AddRange(new[]
             {
-                createLabel("BezierToPiecewiseLinear"),
+                createLabel(nameof(PathApproximator.BezierToPiecewiseLinear)),
                 new ApproximatedPathTest(PathApproximator.BezierToPiecewiseLinear),
             });
 
             Cell(1).AddRange(new[]
             {
-                createLabel("CatmullToPiecewiseLinear"),
+                createLabel(nameof(PathApproximator.CatmullToPiecewiseLinear)),
                 new ApproximatedPathTest(PathApproximator.CatmullToPiecewiseLinear),
             });
 
             Cell(2).AddRange(new[]
             {
-                createLabel("CircularArcToPiecewiseLinear"),
+                createLabel(nameof(PathApproximator.CircularArcToPiecewiseLinear)),
                 new ApproximatedPathTest(PathApproximator.CircularArcToPiecewiseLinear),
             });
 
             Cell(3).AddRange(new[]
             {
-                createLabel("LagrangePolynomialToPiecewiseLinear"),
+                createLabel(nameof(PathApproximator.LagrangePolynomialToPiecewiseLinear)),
                 new ApproximatedPathTest(PathApproximator.LagrangePolynomialToPiecewiseLinear),
             });
         }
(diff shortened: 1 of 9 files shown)


OliBomby:
https://github.com/ppy/osu-framework/assets/17460441/a1f779df-d520-4461-9edb-c8f37fe5544c

It looks pretty cool, but i dont like how the B-splines dont approximate the drawn path accurately if you increase the tolerance so it doesn't spam too many anchors. It places the anchors on the drawn path so the spline path naturally drifts away from the drawn path. You're relying on spamming many anchors to get the spline to be close to the drawn path. I think you should use an algorithm to find optimal positions for the anchors so you can approximate the drawn path with fewer anchors.

default0:
> osu.Framework.Tests_eW6t9KFjIX.mp4
> 
> It looks pretty cool, but i dont like how the B-splines dont approximate the drawn path accurately if you increase the tolerance so it doesn't spam too many anchors. It places the anchors on the drawn path so the spline path naturally drifts away from the drawn path. You're relying on spamming many anchors to get the spline to be close to the drawn path. I think you should use an algorithm to find optimal positions for the anchors so you can approximate the drawn path with fewer anchors.
> 
> Also would be nice if this didn't use B-splines because I don't think peppy is planning to add B-splines to the editor. [ppy/osu#24457 (comment)](https://github.com/ppy/osu/pull/24457#issuecomment-1664697237)

Such an algorithm was already tried. It is both very complex, very hard to actually make computationally feasible, and is likely to cause bulging of the spline around corners - which is highly undesirable from a users POV.
The current solution always ensures that the path curves inwards from an approximate convex shape drawn by the user - this is usually very handy behavior for sliders because generally you would prefer something that is more compact than an ideal approximation - which this naturally provides.

If you have a concrete algorithm / approach in mind for determining the positions of control points, please share. I am open to ideas that are both runtime efficient enough to execute in real time even for fairly long hand-drawn paths as well as not causing undesirable side effects like bulging or instability of the approximation.

OliBomby:
> Such an algorithm was already tried. It is both very complex, very hard to actually make computationally feasible, and is likely to cause bulging of the spline around corners - which is highly undesirable from a users POV. The current solution always ensures that the path curves inwards from an approximate convex shape drawn by the user - this is usually very handy behavior for sliders because generally you would prefer something that is more compact than an ideal approximation - which this naturally provides.
> 
> If you have a concrete algorithm / approach in mind for determining the positions of control points, please share. I am open to ideas that are both runtime efficient enough to execute in real time even for fairly long hand-drawn paths as well as not causing undesirable side effects like bulging or instability of the approximation.

I don't think the path having a bias to curve inwards is preferred behaviour. The property which is preferred is the smoothing. You want a path that smoothes out bumps in the piecewise linear drawn path. The smoothing doesn't necessarily need to curve inwards, i think it's better if it tries to follow the drawn shape as closely as possible.

I have a concrete algorithm which can achieve this. It's an optimization based algorithm and it's implemented in Python. It can do simple shapes in very few iterations, as fast as 6 milliseconds for a full circle shape with 10 anchors. It scales pretty nicely higher orders too, up to around order 1000. It currently only does Beziers, but B-splines could be pretty easily adapted since thats in the same class of interpolations.
https://github.com/OliBomby/Bezier-Approximation/blob/master/shape_approximator.py

Tom94:
Hi OliBomby -- thanks for the elaborate answer! 

> I have a concrete algorithm which can achieve this.

That sounds really promising, actually, and I like the loss descent via Adam! How stable is the optimization in your experience? What I mean is: if a user were to interactively draw a path and run your algorithm every frame, could it be made (e.g. with a fixed RNG seed) such that the resulting sliders look like they're incrementally drawn without wobbling around too much?

> The smoothing doesn't necessarily need to curve inwards, i think it's better if it tries to follow the drawn shape as closely as possible.

The problem we encountered while implementing such a method was that those methods achieved a better fit by bulging around corners rather than becoming sharp. This was much more visually unpleasant than the inward-curving current solution with a larger curve distance... but at the same time we didn't do exactly the same thing that you did. Ours was more of a genetic algorithm and didn't make use of gradient information. Can your approach deal with this sort of situation? I suspect a careful regularizer / loss term might work and I'm hoping you've already ran into this and figured something out.

> It can do simple shapes in very few iterations, as fast as 6 milliseconds for a full circle shape with 10 anchors.

We might need 1-2 orders of magnitude more speed, depending on the context in which you benchmarked -- we need around 5-10ms (to leave headroom for 60 fps) for the most complicated sliders a mapper might draw on a fairly old machine (say a 5-10 years old laptop) to ensure this feature will work for everyone. But I feel like that's probably achievable with careful tuning and if the bulging situation can be kept in check.


It'd be tremendously helpful if you could implement your BSpline-adapted algorithm into the `PathApproximator` API. Then, we could compare many different situations interactively and, if it works better, easily transition interactive slider drawing to it in the future. (Probably out of scope for this particular PR, though.)

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -0,0 +1,190 @@
...
+
+        /// <summary>
+        /// Clears the input path and the B-Spline.
+        /// </summary>
+        public void Clear()
+        {
+            inputPath.Clear();
+            controlPoints.Clear();
+            if (outputCache.IsValid)
+                outputCache.Value.Clear();
```
bdach, line 128: Maybe this should set empty list instead? Fixes the original problem and while making the state of everything more consistent.

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -0,0 +1,190 @@
...
+        public void AddLinearPoint(Vector2 v)
+        {
+            if (!outputCache.IsValid)
+                redrawApproximatedPath();
+
+            if (inputPath.Count == 0)
+            {
+                inputPath.Add(v);
+                controlPoints.Add(inputPath[0]);
+                controlPoints.Add(inputPath[0]);
```
bdach, line 144: Probably worth an inline comment why (i.e. that the last point will be getting moved as the path is drawn).

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -0,0 +1,190 @@
...
+                return;
+            }
+
+            inputPath.Add(v);
+
+            var cps = controlPoints;
```
bdach, line 150: nit: This seems to just be an alias of `controlPoints`? Feels slightly pointless.

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        private Vector2 momentumDirection(IReadOnlyList<Vector2> vertices, int window)
+        {
+            if (vertices.Count < window + 1)
+                return Vector2.Zero;
+
+            var sum = Vector2.Zero;
+            for (int i = 0; i < window; i++)
+                sum += vertices[^(i + 1)] - vertices[^(i + 2)];
```
bdach, line 182: This seems... backwards? If you want a vector that points along a segment, then you'd take the coords of the end of that segment and subtract the coords of the start to get the vector. But this seems to be doing... the converse? Is that intentional for whatever reason that I'm not getting?

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -0,0 +1,190 @@
...
+
+            if (Precision.AlmostEquals(sum.X, 0, 1e-7f) && Precision.AlmostEquals(sum.Y, 0, 1e-7f))
+                return Vector2.Zero;
```
bdach, line 185: I assume this is done for numerical stability? How was the precision chosen here?

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -0,0 +1,190 @@
...
+            // Determine the alignment between the raw path and control path's momentums.
+            // It uses Vector2.Dot which calculates the cosine of the angle between two vectors.
+            // This alignment is used to adjust the control points based on the path's direction change.
+            float alignment = mraw == Vector2.Zero || mcp == Vector2.Zero ? 1.0f : MathF.Max(Vector2.Dot(mraw, mcp), 0.01f);
+
+            // Calculate the distance between the last two control points.
+            // This distance is then used, along with alignment, to decide if a new control point is needed.
+            // The threshold for adding a new control point is based on the alignment and a predefined accuracy factor.
+            float distance = Vector2.Distance(cps[^1], cps[^2]);
+            if (distance / MathF.Pow(alignment, 4) > Tolerance * 1000)
```
bdach, line 169: If my math is right (and it might not be), this is why the other comment about how the direction of the vector is backwards doesn't really matter.

Assuming the vector pointing in the direction of the raw path is called $\vec{a}$, and the vector pointing in the direction of the approximated path is called $\vec{b}$, then you can flip the sign of $\vec{a}$ without affecting the `distance`, because:

$$ \vec{a} \cdot \vec{b} = \left| \vec{a} \right| \left| \vec{b} \right| \cos \theta $$

$$ \vec{a} \cdot \left( -\vec{b} \right) = \left| \vec{a} \right| \left| \vec{b} \right| \cos (\pi - \theta) = -\left| \vec{a} \right| \left| \vec{b} \right| \cos \theta $$

so only the sign of `alignment` changes; but given that this takes `alignment` to the fourth power, the sign no longer matters...?

osu.Framework.Tests/Visual/Graphics/TestSceneShaderStorageBufferObject.cs
```diff
@@ -123,7 +123,7 @@ public override void Draw(IRenderer renderer)
                 shader.Bind();
                 shader.BindUniformBlock("g_ColourBuffer", colourBuffer);
 
-                // Submit vertices, making sure that we don't submit an index which would overflow the SSBO.
+                // Submit rawDrawnPath, making sure that we don't submit an index which would overflow the SSBO.
```
bdach, line 126: nit: mass rename gone wrong?

osu.Framework.Tests/Visual/Drawables/TestScenePathApproximator.cs
bdach, line 1: The changes here seem slightly strange because the count of cells increased to 6, but only 4 are used, and cell number 1 is skipped? Did you mean to add anything there?

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// The piecewise linear approximation of the B-spline created from the input path.
+        /// </summary>
+        public IReadOnlyList<Vector2> OutputPath
+        {
+            get
+            {
+                if (!outputCache.IsValid)
+                    redrawApproximatedPath();
+
+                return outputCache.Value;
```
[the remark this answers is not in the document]
bdach: Resolved by discussing IRL - leaking this reference doesn't matter because it's essentially throwaway.

osu.Framework/Utils/IncrementalBSplineBuilder.cs
bdach: Resolved IRL: done for brevity, I think that's okay.

osu.Framework/Utils/IncrementalBSplineBuilder.cs
bdach: Resolved IRL: this is just canonical momentum calculation, it's supposed to be like that (and may be changing with upcoming @Tom94 changes anyways?)

osu.Framework/Utils/IncrementalBSplineBuilder.cs
bdach: Discussed IRL: for numerical stability, default precision not enough. Asked for an inline comment explaining this.

bdach:
I'm pretty okay with this in general as is. I'd say I generally prefer this over @OliBomby's approach as it is just simpler and appears to work well enough.

Personally, to entertain considering the alternative, I'd have to see that it has to (a) perform at comparable speed or faster, and (b) be a measurable improvement in a significant number of reasonable cases. Otherwise I think there's no point complicating this with black-box optimisation.

OliBomby:
I agree my suggestion is a bit too complex to consider for this PR. I'll attempt implementing it in framework so we can do more comparisons and probably follow up with my own PR.

I'm optimistic that the mentioned concerns can be resolved and I know a few ways to make the algorithm even faster and more viable for real-time.

bdach (approved):

Issue #6046, opened 2023-11-13
IncrementalBSplineBuilder fails to add control points at start of input path


OliBomby:
Best explained using `TestSceneInteractivePathDrawing`. If you draw a path, the B-spline builder will not start adding control points until a certain distance from the starting point is reached, so any detail at the start of the input path is lost.

https://github.com/ppy/osu-framework/assets/17460441/2355a37a-4c34-47a8-af63-1754351334d9

I think the issue is on `IncrementalBSplineBuilder` line 159.
In this scenario, the control point count is 2 so momentum direction does not get calculated. Therefore only the distance is taken into account to determine whether to add a new control point.
A fix could be to take the momentum direction from the start of the input path as the momentum direction for the control points if the control point count is 2.

PR #6047, opened 2023-11-14
IncrementalBSplineBuilder: rewrite algorithm with many improvements 

Tom94:
- [x] Depends on https://github.com/ppy/osu-framework/pull/6045

- The BSpline builder now smoothens the input curve before processing it, making it more reliable.
- Corners are detected explicitly based on local maxima of winding rate. To be automatically converted to red control points on osu!'s end.
- The density of control points is also determined by winding rate.

Fixes https://github.com/ppy/osu-framework/issues/6046 and addresses comments in https://github.com/ppy/osu/pull/25409

https://github.com/ppy/osu-framework/assets/4923655/aa025169-8d57-4571-9d74-d6095d30f924

### DIFF
--- a/osu.Framework/Utils/IncrementalBSplineBuilder.cs
+++ b/osu.Framework/Utils/IncrementalBSplineBuilder.cs
@@ -48,16 +48,19 @@ namespace osu.Framework.Utils
 
         private float inputPathLength => cumulativeInputPathLength.Count == 0 ? 0 : cumulativeInputPathLength[^1];
 
-        public const float FD_EPSILON = PathApproximator.BEZIER_TOLERANCE * 8f;
+        /// <summary>
+        /// Spacing to use in spline-related finite difference (FD) calculations.
+        /// </summary>
+        internal const float FD_EPSILON = PathApproximator.BEZIER_TOLERANCE * 8f;
 
         /// <summary>
-        /// Get the amount of rotation (in radians) at a given point on the path.
+        /// Get the absolute amount of rotation (in radians) at a given point on the path.
         /// </summary>
         /// <param name="path">The path to get the rotation from.</param>
         /// <param name="cumulativeDistances">The cumulative distances of the path.</param>
         /// <param name="t">The point on the path to get the rotation from.</param>
-        /// <returns>The amount of rotation (in radians) at the given point on the path.</returns>
-        private static float getWindingAt(List<Vector2> path, List<float> cumulativeDistances, float t)
+        /// <returns>The absolute amount of rotation (in radians) at the given point on the path.</returns>
+        private static float getAbsWindingAt(List<Vector2> path, List<float> cumulativeDistances, float t)
         {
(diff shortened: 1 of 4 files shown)


osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -78,43 +151,209 @@ public IReadOnlyList<Vector2> OutputPath
...
+        private List<float> detectCorners(List<Vector2> vertices, List<float> distances)
+        {
+            var cornerT = new List<float> { 0f };
+
+            float threshold = cornerThreshold / FD_EPSILON;
+
+            const float step_size = FD_EPSILON;
+            int nSteps = (int)(distances[^1] / step_size);
+
+            // Empirically, averaging the winding rate over a neighborgood of 32 samples seems to be
```
bdach, line 229:
```suggestion
            // Empirically, averaging the winding rate over a neighborhood of 32 samples seems to be
```

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -78,43 +151,209 @@ public IReadOnlyList<Vector2> OutputPath
...
+            // a good representation of the neighborhood of the curve.
+            const int n_avg_samples = 32;
+            float avgCurvature = 0.0f;
+
+            for (int i = 0; i < nSteps; ++i)
+            {
+                // Update average curvature by adding the new winding rate and subtracting the old one from
+                // nAvgSamples steps ago.
+                float newt = i * step_size;
+                float newWinding = MathF.Abs(getWindingAt(vertices, distances, newt));
```
bdach, line 239: nit: slightly redundant abs (`getWindingAt()` already returns absolute)

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -78,43 +151,209 @@ public IReadOnlyList<Vector2> OutputPath
...
+                float oldt = (i - n_avg_samples) * step_size;
+                float oldWinding = oldt < 0 ? 0 : MathF.Abs(getWindingAt(vertices, distances, oldt));
+
+                avgCurvature += (newWinding - oldWinding) / n_avg_samples;
+
+                // Check whether the current winding rate is a local maximum and whether it exceeds the
+                // threshold as well as the surrounding average curvature. If so, we have found a corner.
+                // Also prohibit marking new corners that are too close to the previous one.
+                float midt = (i - n_avg_samples / 2f) * step_size;
+                float midWinding = midt < 0 ? 0 : MathF.Abs(getWindingAt(vertices, distances, midt));
```
bdach, line 250: could make this a local function maybe? (something like `windingAtSample(float i)`

bdach (commented):
source looks generally good, haven't tested yet in practice

bdach (approved):

PR #6048, opened 2023-11-16
Allow enabling computed mipmaps in `LargeTextureStore`

EVAST9919:
As [per discussion](https://discord.com/channels/188630481301012481/589331078574112768/1174532880408518838).

### DIFF
--- a/osu.Framework.Tests/Visual/Sprites/TestSceneLargeTextureStore.cs
+++ b/osu.Framework.Tests/Visual/Sprites/TestSceneLargeTextureStore.cs
@@ -5,6 +5,7 @@ using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Rendering;
+using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
 using osu.Framework.Graphics.Textures;
 using osu.Framework.IO.Stores;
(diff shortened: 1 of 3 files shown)


bdach:
No testing whatsoever?

EVAST9919:
Added 👌

osu.Framework.Tests/Visual/Sprites/TestSceneLargeTextureStore.cs
```diff
@@ -0,0 +1,97 @@
...
+namespace osu.Framework.Tests.Visual.Sprites
+{
+    public partial class TestSceneLargeTextureStore : FrameworkTestScene
+    {
+        private DependencyContainer dependencies = null!;
+
+        [BackgroundDependencyLoader]
+        private void load(IRenderer renderer, GameHost host, Game game)
+        {
+            dependencies.CacheAs(new LargeTextureStore(renderer, host.CreateTextureLoaderStore(new NamespacedResourceStore<byte[]>(game.Resources, "Textures")), manualMipmaps: false));
```
peppy, line 23: Setting `manualMipamaps: true` doesn't change anything in the test.

I'd hope the test shows the difference between the two.

peppy (changes requested):

osu.Framework.Tests/Visual/Sprites/TestSceneLargeTextureStore.cs
EVAST9919: >Setting manualMipamaps: true doesn't change anything in the test.

That's weird, for me the difference is very clear (I'm on D3D)

peppy: May have been hot reload failing. But I'd also hope the test can be run in both modes. I don't want to have to change code each time I want to check a test.

EVAST9919: Will do. Was just trying to ensure it works for you and not a renderer issue or something.

peppy (approved):

PR #6049, opened 2023-11-16
Use better `RunOnAudioThread()` logic in audio tests

Susko3:
Noticed the problem in https://discord.com/channels/188630481301012481/589331078574112768/1174627793221472276.

Updates all audio tests that start a new audio thread to use the `RunOnAudioThread()` logic from #5946.
Previously, some tests had custom logic that would silence exceptions.

### DIFF
--- a/osu.Framework.Tests/Audio/TrackVirtualTest.cs
+++ b/osu.Framework.Tests/Audio/TrackVirtualTest.cs
@@ -10,8 +10,6 @@ using NUnit.Framework;
 using osu.Framework.Audio;
 using osu.Framework.Audio.Track;
 using osu.Framework.Bindables;
-using osu.Framework.Development;
-using osu.Framework.Threading;
 
 namespace osu.Framework.Tests.Audio
 {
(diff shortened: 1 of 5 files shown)


peppy (approved):

peppy (approved):

osu.Framework.Tests/Audio/TrackVirtualTest.cs
```diff
@@ -383,40 +381,15 @@ private void startPlaybackAt(double time)
...
                 track.Update();
             });
         }
-
-        /// <summary>
-        /// Certain actions are invoked on the audio thread.
-        /// Here we simulate this process on a correctly named thread to avoid endless blocking.
-        /// </summary>
-        /// <param name="action">The action to perform.</param>
-        public static void RunOnAudioThread(Action action)
```
bdach, line 402: I am somewhat confounded by the fact that this test just removes this method and updates usages, while `BassTestComponents` keeps the analogous method and just proxies it to `AudioTestHelper`, but not a big deal I suppose?

bdach (approved):

osu.Framework.Tests/Audio/TrackVirtualTest.cs
peppy: I tried removing it but it’s quite ugly. The local one is used across multiple tests so I decided to just leave it in the end.

PR #6050, opened 2023-11-17
Update some inspections to (mostly) satisfy latest rider EAP

peppy:
- Rewrite veldrid extension logging code to avoid fixed-in-delegate inspection
- Fix multiple cases of using objects which may be disposed
- Add some missing nullability markers
- Add new migration flag

A couple of inspections are still flagging warning, but are r# bugs:

https://youtrack.jetbrains.com/issue/RSRP-494840/Incorrect-Conditional-access-qualifier-expression-is-known-to-be-not-null 
https://youtrack.jetbrains.com/issue/RSRP-494839/Resharper-asks-to-dispose-IEnumerator-which-does-not-implement-IDisposable

### DIFF
--- a/osu.Framework/Graphics/Containers/DrawSizePreservingFillContainer.cs
+++ b/osu.Framework/Graphics/Containers/DrawSizePreservingFillContainer.cs
@@ -44,7 +44,7 @@ namespace osu.Framework.Graphics.Containers
         {
             base.Update();
 
-            Vector2 drawSizeRatio = Vector2.Divide(Parent.ChildSize, TargetDrawSize);
+            Vector2 drawSizeRatio = Vector2.Divide(Parent!.ChildSize, TargetDrawSize);
 
             switch (Strategy)
             {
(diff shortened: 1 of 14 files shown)


osu.Framework/Graphics/Containers/DrawSizePreservingFillContainer.cs
```diff
@@ -44,7 +44,7 @@ protected override void Update()
         {
             base.Update();
 
-            Vector2 drawSizeRatio = Vector2.Divide(Parent.ChildSize, TargetDrawSize);
+            Vector2 drawSizeRatio = Vector2.Divide(Parent!.ChildSize, TargetDrawSize);
```
smoogipoo, line 47: One step closer for to finally stop using `.AsNonNull()` and `Debug.Assert()`...

smoogipoo (approved):

osu.Framework/Graphics/Containers/DrawSizePreservingFillContainer.cs
peppy: using `!` is pretty common in the code base, main issues i still have is chaining with `!` (not) operators and other weirdness.

PR #6052, opened 2023-11-19
Don't draw when the window is minimised

Susko3:
Original [discord thread](https://discord.com/channels/188630481301012481/589331078574112768/1175771698814013520) that sparked this change.

When the window is restored from minimised state, it'll show the old frame (with DX11) until a new frame is rendered.

This change visibly reduced the CPU and GPU usage (GPU goes to 0%) when the window is minimised, previously I'd only see a small reduction in GPU usage.


peppy (approved):

peppy:
There might be some implementation side effects from the draw thread not being run for long periods of time (we had the same thing with stable), but we can deal with them as they come up.

PR #6055, opened 2023-11-20
Improve incremental bspline builder

Tom94:
This PR makes the following improvements to the incremental BSpline builder:
1. the last vertex of the BSpline now follows the mouse perfectly without any jitter. Vertices are still inserted with `FD_EPSILON * 2` spacing to ensure derivatives are computed stably; only the currently last vertex is affected by this change.
2. the control point spacing algorithm has been tweaked to prevent global dependencies along the path. This gets rid of jittering that the previous batch of "improvements" introduced. See comparison videos below.

Before this PR:
https://github.com/ppy/osu-framework/assets/4923655/69f60092-f24f-4f62-a296-143a3058fe2c

After this PR:
https://github.com/ppy/osu-framework/assets/4923655/8d8c72d0-0822-4364-8929-1c2e9598c6d2

### DIFF
--- a/osu.Framework/Utils/IncrementalBSplineBuilder.cs
+++ b/osu.Framework/Utils/IncrementalBSplineBuilder.cs
@@ -320,22 +320,25 @@ namespace osu.Framework.Utils
                         totalWinding += getAbsWindingAt(vertices, distances, t);
                     }
 
-                    int nControlPoints = (int)(totalWinding / Tolerance);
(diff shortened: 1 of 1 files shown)


bdach:
I don't really trust myself to properly gauge what is the desired UX of the drawing process at this point so I'm deferring review of this one.

peppy (dismissed):

peppy:
This behaves much better than before, so let's go with it 👍 .

peppy (approved):

PR #6056, opened 2023-11-20
Add method to convert piecewise linear curves to spline control points

OliBomby:
## Summary
This PR introduces an algorithm for approximating arbitrary piecewise linear paths with B-splines or Bezier curves. The method `PiecewiseLinearToBSpline` is then used to improve the paths generated by `IncrementalBSplineBuilder`, making the curve accurately follow the input path using as few control points as possible.

- Adds methods `PiecewiseLinearToBezier` and `PiecewiseLinearToBSpline` to `PathApproximator`.
- Adds `TestScenePiecewiseLinearToBSpline` to interactively test the algorithm on various curves with controllable parameters.
- Adds `BenchmarkPiecewiseLinearToBezier` to benchmark the performance of the algorithm.

### Utility
This addresses two points of concern mentioned in https://github.com/ppy/osu/pull/25409:
> It's very hard to use the freehand to create a shape you're trying to make.

The generated path more closely follows your input, so it should be easier to create the shape you are actually trying to create because you dont have to manually account for the inward bias.

>The resultant points are too frequent to provide a good manual tweaking experience after creating

The path requires less control points to still follow your input, so there are less control points to require tweaking afterwards.

## TODOs
A later PR will introduce interop with `IncrementalBSplineBuilder`. I plan to improve performance by re-using results from previous frames so fewer iterations are necessary per frame to optimize the control points (this also would decrease wobbling).

Also weight matrices can be cached in between frames for even more performance gains. I also want to use the total winding to increase precision in the approximation algorithm when necessary (on very long, complex paths).

Usability can be improved by locking previous control points which have been sufficiently optimised, so there is no wobble in the entire path while drawing the path.

## Benchmarks
I added a nuget package to handle efficient large matrix multiplication, for this I added Tensor.NET. I chose this with performance in mind. I added a benchmark and compared performance against NumSharp, where Tensor.NET won by a large margin. Then I rewrote it to pure C# because Tensor.NET is not compatible and it got way more performant. Super large matrices might still be faster in Tensor.NET but that wont be the case most of the time.

Benchmark using NumSharp:
rider64_qFjXZbNaIq

Benchmark using Tensor.NET  (about 10x faster and comparable to native numpy in python):
rider64_rFjk6UnHx5

Benchmark using pure C# and optimized allocations:
rider64_x4RZsMLJsy

Benchmark using `System.Numerics.Tensors` for SIMD:
rider64_zh4BIyM8cI

@Tom94 Please give feedback, what do you think about the results?

https://github.com/ppy/osu-framework/assets/17460441/14c713e7-9573-446a-b23b-c95eb3dbb12e


bdach:
There's this and there's https://github.com/ppy/osu-framework/pull/6055. I'm not even sure what to do with these right now.

@OliBomby At the point where this sort of stuff is happening, I'd appreciate at least a minimum effort of communication before opening this.

osu.Framework/osu.Framework.csproj
```diff
@@ -52,5 +52,6 @@
     <!-- Any version ahead of this will cause runtime issues on iOS
          See https://github.com/xoofx/markdig/issues/564 -->
     <PackageReference Include="Markdig" Version="0.23.0" />
+    <PackageReference Include="Tensor.NET" Version="0.1.2" />
```
bdach, line 55: I'm not specifically sure why but this kills the build.

The library is also literally a student project, has a single author, and I'm not even sure it'll be maintained going forward. Red flag galore.

OliBomby: Yeah I'm gonna need to find some alternative for this. Not sure what the best way is to substitute this package. I could use NumSharp and deal with terrible performance, or try implementing the whole algorithm in C++ and including the dll, or try implementing the whole algorithm in pure C# (which might not perform better than NumSharp.

Tensor.NET used optimised C++ implementations behind the scenes so that might be what kills the build in CI. It builds locally though, so you can still try it out.

OliBomby:
> There's this and there's #6055. I'm not even sure what to do with these right now.
> 
> @OliBomby At the point where this sort of stuff is happening, I'd appreciate at least a minimum effort of communication before opening this.

I had mentioned in https://github.com/ppy/osu-framework/pull/6044#issuecomment-1807136073 that I'd open this PR. It's mostly meant for @Tom94 to compare the utility this algorithm provides.

osu.Framework/osu.Framework.csproj
bdach: We're not doing any custom cpp, period.

Tom94:
Hi, I find the path approximations from the video *very* impressive for the relatively small amount of code used in the reconstruction! Thanks a bunch for the follow-up.

I'd say this PR is orthogonal to the existing BSpline builder. We should probably keep the existing hard corner detection while being able to either drop-in replace the winding-based control point placement in between corners or use it as initialization for this algorithm. I feel like it'd be valuable to have this in framework in any case as something like a self-contained `List<Vector2> optimizeBSpline(List<Vector2> initialCps, List<Vector2> targetPath)` function that can be optionally enabled in `IncrementalBSplineBuilder` once it becomes production ready (perf and otherwise).

As you already point out, there are a few rough edges, but I think they can be ironed out. My main gripes would be:
1. Control points are very far from the curve at high `Tolerance` values... but I think that regime is not so important anyway. A lower default value, combined with a (potentially tweakable) hard constraint on the maximum distance from the curve could solve this.
2. The entire curve wobbles a lot when new points are added -- but you seem to already be on top of this in your TODO list. :)
3. For the test scene, I recommend adding lines that visually connect the control points so it's easier to see which part of the BSpline they correspond to when there are self-intersections. This wasn't needed when control points would always fall on the input path, but now I find myself confused with some of the reconstructions.

osu.Framework/osu.Framework.csproj
Tom94: Would it make sense to implement the tensor math in terms of vectorized 4x4 blocks that C# natively provides? https://learn.microsoft.com/en-us/dotnet/standard/simd

I suspect this'll get rid of sufficiently much inner-loop that a C# impl will be similarly fast as a C++ one.

peppy:
> Control points are very far from the curve at high Tolerance values

Not looking to deep into this, but just want to add that this is a blocker in terms of osu! side usability. Control points generated by all settings exposed to the user must look sane to the user so they can easily edit/tweak the curve after drawing it.

OliBomby:
Good news! I've rewritten the algorithm to not use Tensor.NET and be pure C# instead, and its actually 2x-10x faster and uses 100x times less heap allocations. I didn't expect a naive implementation without SIMD to still be this fast. Benchmark below.
rider64_x4RZsMLJsy
Though there is a lot more code now for all the matrix operations, so it might warrant being put in its own class.

> Not looking too deep into this, but just want to add that this is a blocker in terms of osu! side usability. Control points generated by all settings exposed to the user must look sane to the user so they can easily edit/tweak the curve after drawing it.

The distance between control points and the curve depends on the locality and number of control points. I think if we restrict the order to be sufficiently low and add enough control points, the result should be close enough to the curve to look sane.

> For the test scene, I recommend adding lines that visually connect the control points so it's easier to see which part of the BSpline they correspond to when there are self-intersections. This wasn't needed when control points would always fall on the input path, but now I find myself confused with some of the reconstructions.

I think this is a good addition, it lets us see more accurately if the control points are 'sane'.

osu.Framework/Utils/PathApproximator.cs
```diff
@@ -406,37 +410,175 @@ public static List<Vector2> LagrangePolynomialToPiecewiseLinear(ReadOnlySpan<Vec
...
+            for (int i = 0; i < m; i++)
+            {
+                for (int j = 0; j < n; j++)
+                {
+                    result[i, j] = mat1[i, j] - mat2[i, j];
+                }
+            }
+        }
+
+        private static void matmul(float[,] mat1, float[,] mat2, float[,] result)
```
smoogipoo: How girthy are these matrices expected to be?

OliBomby:
I've removed the changes to `IncrementalBSplineBuilder` to save that for another PR and not make this PR too huge. I've now marked this ready for review just for the `PiecewiseLinearToBezier` and `PiecewiseLinearToBSpline` methods.

osu.Framework/Utils/PathApproximator.cs
```diff
@@ -297,6 +298,409 @@ public static List<Vector2> LagrangePolynomialToPiecewiseLinear(ReadOnlySpan<Vec
...
+                for (int j = 0; j < numControlPoints; j++)
+                {
+                    result[i, j] = coefficients[j] * p[i, j] * p[numTestPoints - i - 1, numControlPoints - j - 1];
+                }
+            }
+
+            return result;
+        }
+
+        private static long[] binomialCoefficients(int n)
```
bdach, line 686: The fact that this says it is computing binomial coefficients, but it is computing $n$ _terms_ rather than ${n \choose 0}, {n \choose 1}, \dots, {n \choose n}$ (which is $n + 1$ terms) as I would generally expect it to by definition is very confusing and made confirming the iterative derivation's correctness more difficult. I feel this deserves a comment somewhere at least, if not a full re-do of the method.

osu.Framework/Utils/PathApproximator.cs
```diff
@@ -297,6 +298,409 @@ public static List<Vector2> LagrangePolynomialToPiecewiseLinear(ReadOnlySpan<Vec
...
+
+            for (int i = 0; i < count; i++)
+            {
+                result[i] = start + (end - start) * i / (count - 1);
+            }
+
+            return result;
+        }
+
+        private static void getDistanceDistribution(float[,] points, float[] result)
```
bdach, line 522: What is this supposed to be doing? What are the inputs?

osu.Framework/Utils/PathApproximator.cs
```diff
@@ -297,6 +298,409 @@ public static List<Vector2> LagrangePolynomialToPiecewiseLinear(ReadOnlySpan<Vec
...
+                for (int j = 0; j < m1; j++)
+                {
+                    float mCorr = m[i, j] * mMult;
+                    float vCorr = v[i, j] * vMult;
+                    parameters[i, j] -= learningRate * mCorr / (MathF.Sqrt(vCorr) + epsilon);
+                }
+            }
+        }
+
+        // mat1 can not be the same array as result, or it will not work correctly
```
bdach, line 443: This should be an assertion.

bdach (commented):
code is underdocumented and opaque

osu.Framework/Utils/PathApproximator.cs
OliBomby: Ah i wasn't aware that it is convention to return N+1 terms. I'll change the method.

osu.Framework/Utils/PathApproximator.cs
OliBomby: It calculates a normalized cumulative distribution for the Euclidean distance between points on a piecewise-linear path.
I'll add xmldoc

osu.Framework/Utils/PathApproximator.cs
OliBomby: so true

bdach (approved):
still think this is opaque in places but since i'm getting told to move this on -

PR #6058, opened 2023-11-22
Refactor dropdown test scene thoroughly

frenzibyte:
I have been looking towards https://github.com/ppy/osu/issues/24315, and the test scene was way dated and quite bothersome to use, so I figured it was time to refactor it.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
@@ -27,7 +27,7 @@ namespace osu.Framework.Tests.Visual.UserInterface
         [SetUp]
         public new void SetUp() => Schedule(() =>
         {
-            testDropdown = setupDropdowns<TestDropdown>(1)[0];
+            testDropdown = setupDropdowns(1)[0];
         });
 
         [Test]
@@ -60,7 +60,7 @@ namespace osu.Framework.Tests.Visual.UserInterface
         {
             TestDropdown[] dropdowns = null!;
 
-            AddStep("create two dropdowns", () => dropdowns = setupDropdowns<TestDropdown>(2));
+            AddStep("create two dropdowns", () => dropdowns = setupDropdowns(2));
 
             toggleDropdownViaClick(() => dropdowns[0], "dropdown 1");
             AddAssert("dropdown 1 is open", () => dropdowns[0].Menu.State == MenuState.Open);
@@ -227,6 +227,7 @@ namespace osu.Framework.Tests.Visual.UserInterface
                 testDropdown.ClearItems();
                 testDropdown.ItemSource = bindableList = new BindableList<TestModel>();
             });
+
             toggleDropdownViaClick(() => testDropdown);
 
             AddAssert("no elements in dropdown", () => !testDropdown.Items.Any());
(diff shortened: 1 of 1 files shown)


osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
```diff
@@ -23,106 +22,68 @@ namespace osu.Framework.Tests.Visual.UserInterface
...
-        private const float explicit_height = 100;
-        private float calculatedHeight;
-        private readonly TestDropdown testDropdown, testDropdownMenu, bindableDropdown, emptyDropdown, disabledDropdown;
-        private readonly PlatformActionContainer platformActionContainerKeyboardSelection, platformActionContainerKeyboardPreselection, platformActionContainerEmptyDropdown;
-        private readonly BindableList<TestModel> bindableList = new BindableList<TestModel>();
-
-        private int previousIndex;
-        private int lastVisibleIndexOnTheCurrentPage, lastVisibleIndexOnTheNextPage;
-        private int firstVisibleIndexOnTheCurrentPage, firstVisibleIndexOnThePreviousPage;
+        private TestDropdown testDropdown = null!;
```
bdach, line 25: If some tests are going to set this, and some not, this field should just not exist, and any dropdown fields should instead become locals in each test.

osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
```diff
@@ -447,15 +314,47 @@ public void TestSetNonExistentItem([Values] bool afterBdl)
             AddAssert("text is expected", () => dropdown.SelectedItem.Text.Value.ToString(), () => Is.EqualTo("loaded: non-existent item"));
         }
 
-        private void toggleDropdownViaClick(TestDropdown dropdown, string dropdownName = null) => AddStep($"click {dropdownName ?? "dropdown"}", () =>
+        private TDropdown[] setupDropdowns<TDropdown>(int count, Action<TDropdown> applyBeforeLoad = null!)
```
bdach, line 317: This `applyBeforeLoad` thing is pretty bad. Test code is an exemption from DRY, I'd be completely fine with deferring the child set or whatever to the individual tests. Doing that usually results in better composability in the future anyways.

bdach (commented):
Code quality failures here

osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
frenzibyte: Only 4 out of 13 tests replace `testDropdown`, not sure how to feel about making each test case go through the process of creating a dropdown.

osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
frenzibyte: At the point where a test case wants to create a dropdown without the entire logic of `setupDropdowns`, it might as well just construct one and do whatever it wants. I've removed the `applyBeforeLoad` parameter and manually created a dropdown for each test case that used it.

osu.Framework.Tests/Visual/UserInterface/TestSceneDropdown.cs
peppy: Probably best to make the tests that use the basic one call a method as their first step. Explicit is always better in tests.

peppy:
Tests and code quality failing.

frenzibyte:
> Tests and code quality failing.

Tests were not failing for whatever that's worth. Whenever `dotnet format` fails, that affects test runners as well. Probably shouldn't do so.

bdach (approved):

peppy (approved):

PR #6061, opened 2023-11-24
Add proper handling for `BindableList` collection changes in dropdown

frenzibyte:
- [x] Depends on #6058 

Dependency for https://github.com/ppy/osu/issues/24315. This was unexpectedly more complicated than I thought.

To start off, I rewrote the interactions between `Current.Value` and `SelectedItem` to make much more sense than it was before, so I could easily port them in the `BindableList.CollectionChanged` path.

I've also added support for reordering items in a `Menu`, to be able to handle replace/move events. The logic behind that is unfortunately awkward, as I've had to use `FillFlowConatiner`'s layout position logic to make it work. It would've been totally better if `CompositeDrawable` allowed inserting a child directly between others, but alas.


osu.Framework/Graphics/UserInterface/Menu.cs
```diff
@@ -286,7 +293,12 @@ public virtual void Add(MenuItem item)
 
             drawableItem.SetFlowDirection(Direction);
 
-            ItemsContainer.Add(drawableItem);
+            var items = Children.OrderBy(ItemsContainer.GetLayoutPosition).ToList();
+
+            for (int i = position; i < items.Count; i++)
+                ItemsContainer.SetLayoutPosition(items[i], i + 1);
```
bdach, line 299: I have my doubts about this. By skipping part of the list, this is depending on the fact that nothing else calls `SetLayoutPosition()` on any of these items externally. That is not guaranteed. `ItemsContainer` is protected and can be arbitrarily messed with by any subclass.

If this is to remain then `ItemsContainer` should probably be scoped down to a type that does not provide `SetLayoutPosition()` out of the box. `Container<DrawableMenuItem>` seems to work okay.

If that happens, it better be accompanied by an inline comment, too.

frenzibyte: I don't understand why there could be an implementation that would try to manipulate the layout position of each menu item, but I can change the public type of the container, sure.

peppy (dismissed):
Very hard to review this kind of logic, but it seems to be roughly correct? Anything which goes wrong is likely going to be found post-merge so I'm ready for that.

frenzibyte:
Do you have any suggestions on how I could've made it better to review (if any)? The UI component in itself is certainly complicated by nature, but I hoped I made the changes easier to understand by the organisation of the commits.

peppy:
> Do you have any suggestions on how I could've made it better to review (if any)? The UI component in itself is certainly complicated by nature, but I hoped I made the changes easier to understand by the organisation of the commits.

no. it's just annoying logic to review, huge potentials for incorrect indices in inserts / removals / etc.

osu.Framework/Graphics/UserInterface/Menu.cs
bdach: > I don't understand why there could be an implementation that would try to manipulate the layout position of each menu item

If you can call a method on something it should not break the internal state of the object. That is OOP 101. See: encapsulation.

Design your types in ways that make them impossible to be broken.

osu.Framework/Graphics/UserInterface/Dropdown.cs
```diff
@@ -321,6 +382,7 @@ private void clearItems()
         {
             itemMap.Clear();
             Menu.Clear();
+            selectedItem = null;
```
bdach, line 385: Is there a reason for this change? It changes behaviour when compared to `master`. In the test covering this, clearing the `ItemSource` makes the dropdown display the `null` string rather than empty string like it would previously.

If this change is to be made then I will need assurance that nothing will start displaying nulls game-side or breaking in other ways from this change (or I will need to go check myself).

osu.Framework/Graphics/UserInterface/Menu.cs
frenzibyte: Sure, but there must be a balance defined between the perfection of designing a type and the time spent on it (gained behaviour vs code quality). In addition, the correct approach would be to define a method in `CompositeDrawable` to allow inserting a drawable between others, but I'm wary of adding such logic as opposed to using a simple workaround that's unlikely to break any time soon.

Would you be fine if I add a TODO comment on the `SetLayoutPosition` calls stating the above?

bdach: What's in this branch currently is fine. I was just explaining to you the reasoning for caring about it in the first place.

peppy (approved):

bdach (approved):

PR #6063, opened 2023-11-27
Add a session timestamp prefix to log files (and retain 7 days of logs)

peppy:
Will be useful for diagnosing user issues where they already restarted the game and as a consequence, nuked previous logs.


osu.Framework/Logging/Logger.cs
```diff
@@ -65,8 +73,7 @@ public static Storage Storage
             {
                 storage = value ?? throw new ArgumentNullException(nameof(value));
 
-                // clear static loggers so they are correctly purged at the new storage location.
-                static_loggers.Clear();
```
Susko3, line 69: Just noting this: changing the logger storage when the game is running will no longer write the logger header to the new log files.

Changing the logger storage during runtime is dubious, and the old code is incorrect since it wasn't locking `flush_sync_lock` when accessing a thread-shared resource (`static_loggers`).

So I'm not against this change.

peppy: Unfortunately the setter is used osu! side:

CleanShot 2023-11-28 at 08 28 22


I actually removed this because it was writing the headers more than once to the same file location. I guess I'll add it back and fix that part instead..

peppy: On second thoughts, I guess this depends on what we consider the log header to be. Should it signal the start of a session and therefore *not* be present on a partial log? I could imagine this being one correct way of looking at it.

osu.Framework/Logging/Logger.cs
```diff
@@ -409,30 +414,27 @@ private void writePendingLines()
...
+        private bool headerAdded;
+
+        private static void cycleLogs()
         {
-            lock (flush_sync_lock)
+            try
             {
-                scheduler.Add(() =>
+                DateTime logCycleCutoff = DateTime.UtcNow.AddDays(-7);
+                var logFiles = new DirectoryInfo(storage.GetFullPath(string.Empty)).GetFiles();
```
smoogipoo, line 424: A bit worried about this, in-case storage is set incorrectly it could nuke unrelated files.

How about doing additional checks against the filename? Something like making sure it matches `*.{Name}.log`?

osu.Framework/Logging/Logger.cs
smoogipoo: I consider it as the start of a session, but I wouldn't mind it being output every time.

smoogipoo (changes requested):

osu.Framework/Logging/Logger.cs
peppy: I think I'd rather it didn't output every time so we know that some logs might be missing from the start.

smoogipoo (approved):

PR #6066, opened 2023-11-29
Improve incremental B-spline builder via stochastic optimisation

OliBomby:
## Summary
This PR proposes an improvement to `IncrementalBSplineBuilder` to make it optimize control point placement to more closely match the input path. It utilizes `PiecewiseLinearToBSpline` to iteratively optimize the control point placement while drawing the path, and when a segment is finished it will optimize the segment as a whole a bit further.
Furthermore `IncrementalBSplineBuilder.ControlPoints` is now a `IReadOnlyList<List<Vector2>>` so each segment is explicitly provided.

- `IncrementalBSplineBuilder.ControlPoints` is now a `IReadOnlyList<List<Vector2>>`.
- Added `IncrementalBSplineBuilder.Finish()` which should be called when you finish drawing.
- Updated recommended parameters in `TestSceneInteractivePathDrawing`.
- Added `learnableMask` to `PiecewiseLinearToBSpline` to control which control points to optimize.

### Requirements
- [x] https://github.com/ppy/osu-framework/pull/6056

### Considerations
- `IncrementalBSplineBuilder.ControlPoints` could potentially be modified from the outside which would disrupt the builder as it updates from the previous control points instead of completely remaking the control points every time. I don't know if this should be considered as an issue.
- Slider building could be further improved by adding circular arc type sections. These are generally better looking and easier to edit, also people can't draw perfect circles by hand. This could be achieved by detecting segments which could be reasonably approximated with a circular arc and replacing those segments, though this is probably outside the scope of a B-Spline builder.

https://github.com/ppy/osu-framework/assets/17460441/ed4abace-6d80-47d4-8b43-58304fbf8515


peppy:
Okay this looks pretty amazing.

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -273,6 +279,153 @@ private List<float> detectCorners(List<Vector2> vertices, List<float> distances)
...
+
+            // Make sure the last segment has the correct end-points
+            if (lastSegment.Count >= 1)
+                lastSegment[0] = cps[0];
+            else
+                lastSegment.Add(cps[0]);
+            if (lastSegment.Count >= 2)
+                lastSegment[^1] = cps[^1];
+            else
+                lastSegment.Add(cps[^1]);
```
bdach, line 350: What's this doing? I'd hope `initializeSegment()` would already enforce this?

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -273,6 +279,153 @@ private List<float> detectCorners(List<Vector2> vertices, List<float> distances)
...
+                }
+            }
+            else if (cps.Count < lastSegment.Count)
+            {
+                int toRemove = lastSegment.Count - cps.Count;
+                lastSegment.RemoveRange(lastSegment.Count - toRemove - 1, toRemove);
+            }
+
+            // Optimize the control point placement
+            if (lastSegment.Count is <= 2 or >= 100) return;
```
bdach, line 369: We don't really use this `is <= 2 or >= 100` pattern match syntax; please use the classic style.

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -273,6 +279,153 @@ private List<float> detectCorners(List<Vector2> vertices, List<float> distances)
...
+
+            // Make a mask to prevent modifying the control points which have already been optimized enough.
+            // Also the end-points can not move.
```
bdach, line 372: Where is the "have already been optimised enough" part done? I only see the end `2 * degree` control points being fixed in place.

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -421,5 +529,16 @@ public void AddLinearPoint(Vector2 v)
...
         }
+
+        /// <summary>
+        /// Call this when you are done building the path.
+        /// This method applies the final step of post-processing.
+        /// </summary>
+        public void Finish()
+        {
+            outputCache.Invalidate();
+            controlPointsPartiallyInvalid = true;
```
bdach, line 540: What does `controlPointsPartiallyInvalid` mean? And why is it `true` _when done drawing_?

osu.Framework/Utils/IncrementalBSplineBuilder.cs
OliBomby: `initializeSegment()` does not already enforce this because it does not update the `lastSegment`, it only returns what would be the initial control points for this segment. `lastSegment` is not equal to the initial control points because it already has been optimized for several iterations in the previous frames. 

This part of the code is responsible for updating the end-points of `lastSegment` to be equal to `cps` or adding the end-points if it does not already have them.

osu.Framework/Utils/IncrementalBSplineBuilder.cs
OliBomby: okay

osu.Framework/Utils/IncrementalBSplineBuilder.cs
OliBomby: `controlPointsPartiallyInvalid` means that `controlPoints` is valid for the current input path and settings minus some part of the path which was added onto the end. It will then re-use the previous `controlPoints` to prevent doing double work on the segments it has already computed control points for, basically only updating the last segment of the `controlPoints`.

I set it to true when done drawing because we want to optimize the control points a bit further once we know this is the entire input path. This leads to better results.

bdach: I've renamed this flag and a few other things in https://github.com/ppy/osu-framework/pull/6066/commits/e7c7397291c8cea889e1cd10e237a101e81808b8 to hopefully aid in understanding. Please check that you're ok with this.

OliBomby: I think these new names kinda misrepresent the meaning. 

The method `regenerateLastApproximatedSegment` doesn't regenerate the last segment, because it re-uses the existing control points and updates their position or adds/removes some. Also the effect is not limited to just the last segment. If enough path has been added it could be adding several segments to the path. Saying it 'updates' the control points would be more accurate there. Same thing for `shouldOptimiseLastSegment`.

bdach: Well what you had was pretty terrible readability-wise (`regenerateApproximatedPathControlPoints` vs `updateApproximatedPathControlPoints` - what is the difference there? what's the significance of "regenerate" vs "update"? the entirety of the context that the second method only touches the last part is lost).

I dunno. Maybe the new names aren't strictly correct but at least they attempt to explain themselves in some respect.

peppy (approved):

PR #6071, opened 2023-12-01
Request `READ_EXTERNAL_STORAGE` permission on older android versions

Susko3:
Reported in https://discord.com/channels/188630481301012481/1097318920991559880/1179849444032790589.

Documented (badly) in https://developer.android.com/reference/android/Manifest.permission#READ_MEDIA_IMAGES
> For apps with a targetSdkVersion of Build.VERSION_CODES.S_V2 or lower, the READ_EXTERNAL_STORAGE permission is required, instead, to read image files.

They actually meant to say "for apps that will run on Build.VERSION_CODES.S_V2 or lower [...]"


bdach:
Was any testing done on this? I want to say I'm pretty sure I tested the original change on an android 7 device and saw nothing like what's being reported there.

As much as I dislike android I would like to think they are still able to write accurate documentation.

bdach:
Can this come with an inline comment in the manifest at least then? We're likely never getting rid of that thing anyway but at least that would help explain things given that android docs cannot.

peppy (approved):

PR #6072, opened 2023-12-02
Implement dropdown searching

frenzibyte:
- Framework-side implementation for https://github.com/ppy/osu/issues/21784

RFC. The implementation is a bit heavy, commit-by-commit reviewing is advised.

There are a few hacks with this implementation, mainly the use of a local input manager to allow the search text box to be focused while the menu is focused as well. I tried not to pay too much attention to the details of this implementation to not waste time on it.

https://github.com/ppy/osu-framework/assets/22781491/4e7d7fec-0026-4a6c-83bd-f877e98df617


osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -770,7 +770,17 @@ protected virtual Drawable AddCharacterToFlow(char c)
...
+
+        /// <summary>
+        /// A fixed size for the text displayed in this <see cref="TextBox"/>. If left unset, text size will be computed based on the dimensions of the <see cref="TextBox"/>.
+        /// </summary>
+        public float TextSize
+        {
+            init => customTextSize = value;
+        }
+
+        protected float DisplayedTextSize => customTextSize ?? (TextFlow.DrawSize.Y - (TextFlow.Padding.Top + TextFlow.Padding.Bottom));
```
bdach, line 783: Is there any reason this is this weird? Why is the public member init-only when it looks like it would work when set at any time? Can the "displayed" text size just be plainly publically gettable? And can we name this `FontSize`?

osu.Framework/Graphics/UserInterface/Menu.cs
```diff
@@ -912,7 +942,7 @@ protected override bool OnClick(ClickEvent e)
 
         #endregion
 
-        private partial class ItemsFlow : FillFlowContainer<DrawableMenuItem>
+        private partial class ItemsFlow : SearchContainer<DrawableMenuItem>
```
bdach, line 945: This is somewhat far-reaching...? Every single menu out there is now searchable? Not sure what the performance implications of this are, have you checked?

osu.Framework/Graphics/UserInterface/Menu.cs
```diff
@@ -70,6 +71,27 @@ public abstract partial class Menu : CompositeDrawable, IStateful<MenuState>
...
+        {
+            get => itemsFlow.AllowNonContiguousMatching;
+            set => itemsFlow.AllowNonContiguousMatching = value;
+        }
+
+        public event Action FilterCompleted
+        {
+            add => itemsFlow.FilterCompleted += value;
+            remove => itemsFlow.FilterCompleted -= value;
+        }
```
bdach, line 93: I don't know about this either... Very far-reaching API surface for this singular feature.

osu.Framework/Graphics/UserInterface/Menu.cs
```diff
@@ -582,9 +604,11 @@ private void menuItemHovered(DrawableMenuItem item)
 
         public override bool HandleNonPositionalInput => State == MenuState.Open;
 
+        protected virtual bool CloseOnEscape => !TopLevelMenu;
+
         protected override bool OnKeyDown(KeyDownEvent e)
         {
-            if (e.Key == Key.Escape && !TopLevelMenu)
+            if (e.Key == Key.Escape && CloseOnEscape)
```
bdach, line 611: Can we not add this virtual? Can this be handled more locally to the dropdowns somehow?

osu.Framework/Graphics/UserInterface/Menu.cs
```diff
@@ -698,6 +722,12 @@ public abstract partial class DrawableMenuItem : CompositeDrawable, IStateful<Me
             /// </summary>
             public virtual bool CloseMenuOnClick => true;
 
+            public IEnumerable<LocalisableString> FilterTerms => Item.Text.Value.Yield();
+
+            public abstract bool MatchingFilter { get; set; }
+
+            public abstract bool FilteringActive { set; }
```
bdach, line 729: This is an API break. Consumer classes will have to implement this. Even if they won't / can't do anything with it.

I'd argue we should avoid that if possible.

osu.Framework/Graphics/UserInterface/DropdownHeader.cs
```diff
@@ -15,13 +16,23 @@
...
         public event Action<DropdownSelectionAction> ChangeSelection;
 
         protected Container Background;
         protected Container Foreground;
 
+        public bool AlwaysShowSearchBar
+        {
+            get => SearchBar.AlwaysDisplayOnFocus;
+            set => SearchBar.AlwaysDisplayOnFocus = value;
+        }
```
bdach, line 30: `AlwaysShowSearchBar`, `AlwaysDisplayOnFocus`... can we pick one naming convention and stick with it?

osu.Framework/Graphics/UserInterface/Dropdown.cs
```diff
@@ -175,6 +192,25 @@ protected virtual LocalisableString GenerateItemText(T item)
             }
         }
 
+        /// <summary>
+        /// Puts the state of this <see cref="Dropdown{T}"/> one level back:
+        ///  - If the dropdown search bar contains text, this method will reset it.
+        ///  - If the dropdown is open, this method wil close it.
+        /// </summary>
+        public bool Back()
```
bdach, line 200: This is a weird method and should probably just be inlined at its only usage.

osu.Framework/Graphics/UserInterface/Dropdown.cs
```diff
@@ -216,26 +252,43 @@ protected Dropdown()
...
+            Header.SearchTerm.ValueChanged += t => Menu.SearchTerm = t.NewValue;
+
+            Header.State.BindTo(State);
+
+            Menu.RelativeSizeAxes = Axes.X;
             Menu.PreselectionConfirmed += preselectionConfirmed;
+            Menu.FilterCompleted += menuFilterCompleted;
+            Menu.StateChanged += state => State.Value = state;
+
+            State.ValueChanged += state => Menu.State = state.NewValue;
```
bdach, line 267: Oh jesus christmas. My brain just shut down at this number of callbacks. I'm not sure how I'm ever to review that this all coordinates correctly.

Can this be _somehow_ simplified to not require this level of coordination?

osu.Framework/Graphics/UserInterface/Menu.cs
frenzibyte: Yes, but it felt odd to handle locally in a general component like `Dropdown`, as opposed to modify `Menu`. Here's the diff:
```diff
diff --git a/osu.Framework/Graphics/UserInterface/Dropdown.cs b/osu.Framework/Graphics/UserInterface/Dropdown.cs
index 585c97bbf..d93c1af9c 100644
--- a/osu.Framework/Graphics/UserInterface/Dropdown.cs
+++ b/osu.Framework/Graphics/UserInterface/Dropdown.cs
@@ -609,55 +609,56 @@ protected override bool OnHover(HoverEvent e)
 
             #endregion
 
-            // we'll handle closing the menu in Dropdown instead,
-            // since a search bar may be active and we want to reset it rather than closing the menu.
-            protected override bool CloseOnEscape => false;
-
             protected override bool OnKeyDown(KeyDownEvent e)
             {
                 var visibleMenuItemsList = VisibleMenuItems.ToList();
-                if (!visibleMenuItemsList.Any())
-                    return base.OnKeyDown(e);
-
-                var currentPreselected = PreselectedItem;
-                int targetPreselectionIndex = visibleMenuItemsList.IndexOf(currentPreselected);
 
-                switch (e.Key)
+                if (visibleMenuItemsList.Any())
                 {
-                    case Key.Up:
-                        PreselectItem(targetPreselectionIndex - 1);
-                        return true;
-
-                    case Key.Down:
-                        PreselectItem(targetPreselectionIndex + 1);
-                        return true;
-
-                    case Key.PageUp:
-                        var firstVisibleItem = VisibleMenuItems.First();
-
-                        if (currentPreselected == firstVisibleItem)
-                            PreselectItem(targetPreselectionIndex - VisibleMenuItems.Count());
-                        else
-                            PreselectItem(visibleMenuItemsList.IndexOf(firstVisibleItem));
-                        return true;
-
-                    case Key.PageDown:
-                        var lastVisibleItem = VisibleMenuItems.Last();
+                    var currentPreselected = PreselectedItem;
+                    int targetPreselectionIndex = visibleMenuItemsList.IndexOf(currentPreselected);
 
-                        if (currentPreselected == lastVisibleItem)
-                            PreselectItem(targetPreselectionIndex + VisibleMenuItems.Count());
-                        else
-                            PreselectItem(visibleMenuItemsList.IndexOf(lastVisibleItem));
-                        return true;
+                    switch (e.Key)
+                    {
+                        case Key.Up:
+                            PreselectItem(targetPreselectionIndex - 1);
+                            return true;
+
+                        case Key.Down:
+                            PreselectItem(targetPreselectionIndex + 1);
+                            return true;
+
+                        case Key.PageUp:
+                            var firstVisibleItem = VisibleMenuItems.First();
+
+                            if (currentPreselected == firstVisibleItem)
+                                PreselectItem(targetPreselectionIndex - VisibleMenuItems.Count());
+                            else
+                                PreselectItem(visibleMenuItemsList.IndexOf(firstVisibleItem));
+                            return true;
+
+                        case Key.PageDown:
+                            var lastVisibleItem = VisibleMenuItems.Last();
+
+                            if (currentPreselected == lastVisibleItem)
+                                PreselectItem(targetPreselectionIndex + VisibleMenuItems.Count());
+                            else
+                                PreselectItem(visibleMenuItemsList.IndexOf(lastVisibleItem));
+                            return true;
+
+                        case Key.Enter:
+                            var preselectedItem = VisibleMenuItems.ElementAt(targetPreselectionIndex);
+                            PreselectionConfirmed?.Invoke((DropdownMenuItem<T>)preselectedItem.Item);
+                            return true;
+                    }
+                }
 
-                    case Key.Enter:
-                        var preselectedItem = VisibleMenuItems.ElementAt(targetPreselectionIndex);
-                        PreselectionConfirmed?.Invoke((DropdownMenuItem<T>)preselectedItem.Item);
-                        return true;
+                // we'll handle closing the menu in Dropdown instead,
+                // since a search bar may be active and we want to reset it rather than closing the menu.
+                if (e.Key == Key.Escape)
+                    return false;
 
-                    default:
-                        return base.OnKeyDown(e);
-                }
+                return base.OnKeyDown(e);
             }
 
             public bool OnPressed(KeyBindingPressEvent<PlatformAction> e)

```

If you find it more acceptable this way, I'll apply accordingly.

osu.Framework/Graphics/UserInterface/Dropdown.cs
frenzibyte: This method is intentionally separated as it'll be called from `OsuDropdown` when it handles the `Back` action. Naming is open for suggestions, I thought of `TryClose` at first, maybe go fully explicit and name it `ClearSearchOrClose`.

osu.Framework/Graphics/UserInterface/Menu.cs
frenzibyte: For the sake of moving forward, I applied this.

osu.Framework/Graphics/UserInterface/TextBox.cs
frenzibyte: Making it non-init means adding extra logic to handle updating font size of existing characters, so better keep it init-only.

osu.Framework/Graphics/UserInterface/Menu.cs
frenzibyte: Moved to `DropdownMenu`.

frenzibyte:
Comparison before/after latest changes, to ensure visual correctness:

| | dropdown | text box | `TestBrowserTextBox` |
|-|----------|----------|----------------------|
| before | CleanShot 2023-12-11 at 06 46 23@2x | CleanShot 2023-12-11 at 06 45 49@2x | CleanShot 2023-12-11 at 15 47 51@2x |
| after | CleanShot 2023-12-11 at 10 29 14@2x | CleanShot 2023-12-11 at 06 48 03@2x | CleanShot 2023-12-11 at 10 32 27@2x |

Notice caret is placed and sized correctly on the dropdown search bar. Also notice caret got shorter in `TestBrowserTextBox` after changes, that is actually the correct size for it, since the text box was previously using a smaller font indirectly (by modifying the height of `TextFlow`), thus the caret was taller than the text height.

peppy (approved):

PR #6077, opened 2023-12-07
Implement ability to get `Texture`'s average colour

EVAST9919:
Some design elements require sampling colour from the texture. This implementation covers this functionality in the simplest way possible - getting average colour of non-transparent pixels.

However I'm marking this as a draft due to this issue https://github.com/ppy/osu/issues/9307. Basically looks like iterating over pixel data may cause performance issues.


peppy:
Where is this intended to be used? Also are you sure we want average colour? Generally things like this prefer a smarter algorithm, which picks the accent colours from an image.

EVAST9919:
Closing as outdated/not finished/not required.

PR #6078, opened 2023-12-10
Make `ToTriangle` in `Sharps.Triangle` overrideable

Frederisk:
As a practical and useful example. In some cases, the user needs to change the orientation of the triangle. But simple rotation does not meet the needs, as an example:

```cs
this.InternalChildren = [
  new Box {
    RelativeSizeAxes = Axes.Both,
    Width = 0.8f,
    Height = 0.25f,
    Anchor = Anchor.Centre,
    Origin = Anchor.Centre,
    Colour = Color4.Blue,
  },
  new Triangle {
    RelativeSizeAxes = Axes.Both,
    Rotation = 90,
    Width = 0.8f,
    Height = 0.25f,
    // Width = 0.25f, Height = 0.8f, // Also not working properly.
    Anchor = Anchor.Centre,
    Origin = Anchor.Centre,
    Colour = Color4.Yellow,
  },
];
```

| ` Width = 0.8f, Height = 0.25f,` | `Width = 0.25f, Height = 0.8f,` |
| - | - |
| image | image |

It is not possible to obtain a `Triangle` with the same length and width as the `Box`. Currently, modifying this behavior to make it work as expected would require writing a lot of code, and most of logic would be duplicated with what is already in the `Triangle` class.

The easiest way to override `Triangle` and make it behave as expected is to override the `ToTriangle` method. After modification you can easily change its orientation and it will behave as expected:

```cs
internal partial class RightTriangle : Triangle {
  protected override Primitives.Triangle ToTriangle(Quad q) => new Primitives.Triangle(
    (q.TopRight + q.BottomRight) / 2,
    q.TopLeft,
    q.BottomLeft);
}
// And then...
this.InternalChildren = [
  new Box {
    RelativeSizeAxes = Axes.Both,
    Width = 0.8f,
    Height = 0.25f,
    Anchor = Anchor.Centre,
    Origin = Anchor.Centre,
    Colour = Color4.Blue,
  },
  new RightTriangle {
    RelativeSizeAxes = Axes.Both,
    Width = 0.8f,
    Height = 0.25f,
    Anchor = Anchor.Centre,
    Origin = Anchor.Centre,
    Colour = Color4.Yellow,
  },
];
```

image

### DIFF
--- a/osu.Framework/Graphics/Shapes/Triangle.cs
+++ b/osu.Framework/Graphics/Shapes/Triangle.cs
@@ -30,19 +30,21 @@ namespace osu.Framework.Graphics.Shapes
             Texture ??= renderer.WhitePixel;
         }
 
-        public override RectangleF BoundingBox => toTriangle(ToParentSpace(LayoutRectangle)).AABBFloat;
+        public override RectangleF BoundingBox => ToTriangle(ToParentSpace(LayoutRectangle)).AABBFloat;
 
-        private static Primitives.Triangle toTriangle(Quad q) => new Primitives.Triangle(
+        protected Primitives.Triangle ToTriangle(Quad q) => new Primitives.Triangle(
             (q.TopLeft + q.TopRight) / 2,
             q.BottomLeft,
             q.BottomRight);
 
-        public override bool Contains(Vector2 screenSpacePos) => toTriangle(ScreenSpaceDrawQuad).Contains(screenSpacePos);
+        public override bool Contains(Vector2 screenSpacePos) => ToTriangle(ScreenSpaceDrawQuad).Contains(screenSpacePos);
 
         protected override DrawNode CreateDrawNode() => new TriangleDrawNode(this);
 
         private class TriangleDrawNode : SpriteDrawNode
         {
+            protected new Triangle Source => (Triangle)base.Source;
+
             public TriangleDrawNode(Triangle source)
                 : base(source)
             {
@@ -53,7 +55,7 @@ namespace osu.Framework.Graphics.Shapes
                 if (DrawRectangle.Width == 0 || DrawRectangle.Height == 0)
                     return;
 
-                renderer.DrawTriangle(Texture, toTriangle(ScreenSpaceDrawQuad), DrawColourInfo.Colour, null, null,
+                renderer.DrawTriangle(Texture, Source.ToTriangle(ScreenSpaceDrawQuad), DrawColourInfo.Colour, null, null,
                     new Vector2(InflationAmount.X / DrawRectangle.Width, InflationAmount.Y / DrawRectangle.Height), TextureCoords);
             }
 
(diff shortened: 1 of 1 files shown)


Susko3:
It seems you are misunderstanding how `RelativeSizeAxes` works when paired with `Rotation`. If you use apsolute sizes, you'll see that it works out fine.
Alternatively, add this line and the two will match perfectly:

```diff
this.InternalChildren = [
  new Box {
    RelativeSizeAxes = Axes.Both,
+   Rotation = 90,
    Width = 0.8f,
    Height = 0.25f,
    Anchor = Anchor.Centre,
    Origin = Anchor.Centre,
    Colour = Color4.Blue,
  },
  new Triangle {
    RelativeSizeAxes = Axes.Both,
    Rotation = 90,
    Width = 0.8f,
    Height = 0.25f,
    // Width = 0.25f, Height = 0.8f, // Also not working properly.
    Anchor = Anchor.Centre,
    Origin = Anchor.Centre,
    Colour = Color4.Yellow,
  },
];
```

If you need additional assistance in understanding how `RelativeSIzeAxis` works when used with `Rotation`, feel free to create a new discussion or reach out in [the discord](https://discord.gg/invite/ppy).

---

I don't think it's valuable to expose `toTriangle()` for overriding. We would need to check for vertex order, and it would incur additional virtual method invocation costs.

https://github.com/ppy/osu-framework/blob/63405204eba19d054f24c8f349137129a1557936/osu.Framework/Graphics/Primitives/Triangle.cs#L14-L18

bdach:
> It seems you are misunderstanding how `RelativeSizeAxes` works when paired with `Rotation`

The thing is that the current framework behaviour is not really intuitive or what a framework consumer would expect. Not to disagree with the fact that the solution here is incorrect in fixing the wrong thing, because it is, but yeah. And [we have trouble with this game-side too](https://github.com/ppy/osu/issues/24388).

Then again, I don't know whether there's anything that can feasibly be done to address this framework-side.

Frederisk:
Sorry I didn't notice the performance issues described in the structure. And apart from this, I can't think of any other solution. Maybe I'll just have to build my own implementations so that I can use them in some less performance-sensitive scenarios.🤔

> Alternatively, add this line and the two will match perfectly: ...

This does get one of the desired behaviors: matched size. However, the size was not what I expected. (0.8 times the width of the entire parent object.）

> If you use apsolute sizes, you'll see that it works out fine.

I've tried this approach, like binding the absolute size of other elements to update itself synchronously, etc. But seriously, the complicated bindings make the logic confusing and hard to follow, so I don't really like it.

> The thing is that the current framework behaviour is not really intuitive or what a framework consumer would expect.

In practice, I can also understand that it behaves like: the size is calculated first and then rotated. However, in some scenarios, consumers (such as me) will want to reverse the order of processing. I hope what I'm thinking isn't a weird thing.😕

Susko3:
This is probably the simplest way to achieve what you want. Rotates the triangle, and then scales to match the desired width and height.

```cs
public partial class RotatedTriangle : Triangle
{
    protected override Vector2 DrawScale => Vector2.Divide(DrawSize.Yx, DrawSize);

    public RotatedTriangle()
    {
        Anchor = Anchor.Centre;
        Origin = Anchor.Centre;
        Rotation = 90;
    }
}
```

PR #6079, opened 2023-12-13
Fix dropdown/menu components not providing items in correct order

frenzibyte:
The change in `Dropdown` is quite ugly, I would rather let the component always store a bindable list and get away with the extra list storage, but I'm not fully confident about pushing such change any time soon.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/Menu.cs
+++ b/osu.Framework/Graphics/UserInterface/Menu.cs
@@ -62,7 +62,7 @@ namespace osu.Framework.Graphics.UserInterface
         /// <summary>
         /// Gets the item representations contained by this <see cref="Menu"/>.
         /// </summary>
-        protected internal IReadOnlyList<DrawableMenuItem> Children => itemsFlow.Children;
+        protected internal IEnumerable<DrawableMenuItem> Children => itemsFlow.Children.OrderBy(c => itemsFlow.GetLayoutPosition(c));
 
         protected readonly Direction Direction;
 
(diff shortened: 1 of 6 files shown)


osu.Framework/Graphics/UserInterface/Menu.cs
```diff
@@ -62,7 +62,7 @@ public abstract partial class Menu : CompositeDrawable, IStateful<MenuState>
         /// <summary>
         /// Gets the item representations contained by this <see cref="Menu"/>.
         /// </summary>
-        protected internal IReadOnlyList<DrawableMenuItem> Children => itemsFlow.Children;
+        protected internal IEnumerable<DrawableMenuItem> Children => itemsFlow.Children.OrderBy(c => itemsFlow.GetLayoutPosition(c));
```
frenzibyte, line 65: In case anyone asks, we cannot use `FlowingChildren` here as it filters the items by extra properties that are not wanted (life state).

peppy:
For future reviewer, these can be reviewed when merged:

https://github.com/ppy/osu/blob/c3396a93a63262547ad0c3ead389f5d702109357/osu.Game.Tests/Visual/SongSelectV2/TestSceneCollectionDropdown.cs#L259-L261
https://github.com/ppy/osu/blob/d5fe7e20be19cc8431382ccd5b74e87d7091dfa9/osu.Game.Tests/Visual/SongSelect/TestSceneCollectionDropdown.cs#L268-L270

(Just removing from my local todo list)

PR #6080, opened 2023-12-18
Fix dropdowns showing popup keyboard on mobile even when search bar is initially hidden

peppy:


bdach:
As is, this diff does nothing for me (testing on android).

This helps, but also feels tremendously jank:

```diff
diff --git a/osu.Framework/Graphics/UserInterface/DropdownSearchBar.cs b/osu.Framework/Graphics/UserInterface/DropdownSearchBar.cs
index 31fdf4bb1..3f7c009a6 100644
--- a/osu.Framework/Graphics/UserInterface/DropdownSearchBar.cs
+++ b/osu.Framework/Graphics/UserInterface/DropdownSearchBar.cs
@@ -74,8 +74,12 @@ public void ObtainFocus()
         {
             // On mobile platforms, let's not make the keyboard popup unless the dropdown is intentionally searchable.
             bool willShowOverlappingKeyboard = host?.OnScreenKeyboardOverlapsGameWindow == true;
+
             if (willShowOverlappingKeyboard && !AlwaysDisplayOnFocus)
+            {
+                textBoxInputManager.UseParentInput = false;
                 return;
+            }
 
             textBoxInputManager.ChangeFocus(textBox);
             obtainedFocus = true;

```

For whatever reason that I don't really wish to get into the nested input manager doesn't really care if its parent handled or didn't handle the input, and just proceeds to give focus to the textbox anyway, which causes text input to activate. So the "fix" is to just tell it to sit down and not break stuff.

Think it's okay to go with that?

cc @frenzibyte (i had several bad feelings about using nested input managers for anything that isn't a test, and the number of bugs that stem from introducing it tells me that i should have been more forceful in saying so)

bdach:
Have applied with disclaimer.

bdach (approved):

PR #6083, opened 2023-12-21
Allow partial enumeration of directory entries

### LINKED ISSUE #6082: DirectorySelector fails to enumerate `~/Library/Application Support` directories
When attempting to click `~/Library/Application Support` using the in-game file browser, the file browser flashes red and refuses to go into the directory.

It looks like while `DirectorySelector.TryGetEntriesForPath` is iterating over/sorting directory entries, the below exception is thrown and and caught. It is likely due to [`DirectoryInfo.GetDirectories`](https://learn.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.getdirectories) attempting to pre-populate DirectoryInfo.
```
System.UnauthorizedAccessException: Access to the path '/Users/daohe/Library/Application Support/com.apple.LaunchServicesTemplateApp.dv' is denied.
```

https://github.com/ppy/osu-framework/blob/d267e0fde51f64192bbc1a2315283a5cc2e400be/osu.Framework/Graphics/UserInterface/DirectorySelector.cs#L201-L219

Inaccessible directories in `~/Library/Application Support/` has also been a problem with JetBrains IDEs (e.g. [this RIDER issue](https://youtrack.jetbrains.com/issue/RIDER-93883/Rider-2023.1-and-later-cannot-show-Unitys-declaration.#focus=Comments-27-7443030.0-0)). According to [this IDEA issue](https://youtrack.jetbrains.com/issue/IDEA-334412/com.apple.LaunchServicesTemplateApp.dv-Operation-not-permitted-when-using-IDEA-in-macOS-Sonoma), `~/Library/Application Support/com.apple.LaunchServicesTemplateApp.dv` is used by macOS Sonoma to support the new "Add to Dock" feature.

ILW8:
Closes #6082

Due to [`DirectoryInfo.GetDirectories`](https://learn.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.getdirectories) pre-populating some DirectoryInfo properties, `TryGetEntriesForPath` may fail and return `false` despite there being valid subdirectories for the user to safely pick.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/DirectorySelector.cs
+++ b/osu.Framework/Graphics/UserInterface/DirectorySelector.cs
@@ -11,6 +11,7 @@ using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Graphics.Containers;
+using osu.Framework.Logging;
 using osu.Framework.Platform;
 using osuTK;
 
@@ -204,10 +205,19 @@ namespace osu.Framework.Graphics.UserInterface
 
             try
             {
-                foreach (var dir in path.GetDirectories().OrderBy(d => d.Name))
+                foreach (string directoryName in Directory.GetDirectories(path.FullName).OrderBy(d => d))
                 {
-                    if (ShowHiddenItems.Value || !dir.Attributes.HasFlagFast(FileAttributes.Hidden))
-                        items.Add(CreateDirectoryItem(dir));
+                    try
+                    {
+                        DirectoryInfo di = new DirectoryInfo(directoryName);
+                        if (ShowHiddenItems.Value || !di.Attributes.HasFlagFast(FileAttributes.Hidden))
+                            items.Add(CreateDirectoryItem(di));
+                    }
+                    catch
+                    {
+                        // Don't fail enumeration if we fail getting attributes for a single entry
+                        Logger.Log($"Directory {directoryName} is inaccessible", LoggingTarget.Information, LogLevel.Debug);
+                    }
                 }
 
                 return true;


osu.Framework/Graphics/UserInterface/DirectorySelector.cs
```diff
@@ -204,10 +205,19 @@ protected virtual bool TryGetEntriesForPath(DirectoryInfo path, out ICollection<
...
+                            items.Add(CreateDirectoryItem(di));
+                    }
+                    catch
+                    {
+                        // Don't fail enumeration if we fail getting attributes for a single entry
+                        Logger.Log($"Directory {directoryName} is inaccessible", LoggingTarget.Information, LogLevel.Debug);
+                    }
                 }
 
                 return true;
```
Susko3, line 223: ~~To keep in line with the intentions of the method, I think it should return `false` (like it previously did) if enumerating _all_ directories failed. I.e. if the inner `catch` got triggered, and `items.Count == 0`, `false` should be returned.~~

~~Empty directories should return `true` as it currently works.~~

If you opt for `IgnoreInaccessible` I'm fine with returning `true` even if all the entries couldn't be opened.

osu.Framework/Graphics/UserInterface/DirectorySelector.cs
```diff
@@ -204,10 +205,19 @@ protected virtual bool TryGetEntriesForPath(DirectoryInfo path, out ICollection<
 
             try
             {
-                foreach (var dir in path.GetDirectories().OrderBy(d => d.Name))
```
Susko3, line 207: Looking into this, I found an [`EnumerateDirectories(String, EnumerationOptions)`](https://learn.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.enumeratedirectories?view=net-8.0#system-io-directoryinfo-enumeratedirectories(system-string-system-io-enumerationoptions)) overload that can take in `EnumerationOptions`, which has the option `IgnoreInaccessible` -- exactly what you're looking for (there's also `AttributesToSkip` that could be used for hidden items).

The same should also apply to `FileSelector`.

Susko3 (changes requested):
Found an interesting overload that could be used to improve code quality and (probably and very slightly) performance.

osu.Framework/Graphics/UserInterface/DirectorySelector.cs
ILW8: Unfortunately EnumerateDirectories, even with `IgnoreInaccessible`, throws an exception. I've tried many of the other overloads that allow ignoring certain failures with no success; enumerating using a string of the path really was a last-resort.

https://github.com/ppy/osu-framework/assets/1176059/fd9531be-577d-4199-98a4-3071fabd99fb

osu.Framework/Graphics/UserInterface/DirectorySelector.cs
ILW8: You're right, I overlooked the scenario you described.

PR #6085, opened 2023-12-23
add window hide/show methods (preliminary work for boss key)

adryzz:
Preliminary work for https://github.com/ppy/osu/issues/6790.

Tested on Windows, MacOS and Linux (X11/Wayland), and doesn't work on MacOS (tracked upstream https://github.com/libsdl-org/SDL/issues/8216)


smoogipoo:
Does this transfer focus? Any reason we should be using `Show()` over `Raise()`?

adryzz:
I tested `Raise()` and on the KDE wayland session at least, the window does not raise, as it is hidden.

smoogipoo (approved):

adryzz:
> Failed CanSeekForwardsOnInterpolationFail

i don't think this has anything to do with this PR

PR #6091, opened 2023-12-25
Rename concept of "global" mixer with "fallback"

peppy:
Hopefully this makes immediate sense.

Splitting this out from WASAPI changes to allow me to use the term "global mixer" more correctly.


smoogipoo (approved):
Sure, sounds fine

PR #6093, opened 2023-12-27
Disable WASAPI initialisation mode for now

peppy:
Need some more time to test this out. As I kinda expected, it doesn't work well across the board.


bdach (approved):
\>_>

PR #6094, opened 2023-12-27
Add setting to control automatic minimising on focus loss in fullscreen mode

Susko3:
- Resolves https://github.com/ppy/osu/discussions/17172

Uses `SDL_HINT_VIDEO_MINIMIZE_ON_FOCUS_LOSS` on desktop (and iOS, but I'm not sure if it does anything).
Mobile isn't really a consideration here, as the OS has full control over "minimising a window".

Added tests that require manually alt-tabbing. Tested on windows 11 with DirectX (seamless transition) and OpenGL legacy (black screen flicker).

### DIFF
--- a/osu.Framework.Tests/Visual/Platform/TestSceneFullscreen.cs
+++ b/osu.Framework.Tests/Visual/Platform/TestSceneFullscreen.cs
@@ -3,6 +3,7 @@
 
 #nullable disable
 
+using System;
 using System.Collections.Generic;
 using System.Drawing;
 using System.Linq;
@@ -19,6 +20,7 @@ using osu.Framework.Input;
 using osu.Framework.Logging;
 using osu.Framework.Platform;
 using osuTK;
+using WindowState = osu.Framework.Platform.WindowState;
 
 namespace osu.Framework.Tests.Visual.Platform
 {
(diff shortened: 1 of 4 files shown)


smoogipoo (approved):

THEGOLDENPRO:
Hi, I can't find this setting you are mentioning here.

It seems like ever since this feature was released my game has been launching like the image below despite it being in fullscreen mode.

Toggling fullscreen mode on and off seems to temporarily fix it but it occurs once again when I relaunch.

image

If you ever need my platform details here:
- OS: Arch Linux x86_64
- DE: KDE Plasma 5.27.10
- Kernel: 6.6.9-arch1-1

Let me know if you need anything else, thank you.

Susko3:
1. This feature is not yet available, as it was merged after the latest release. When framework is updated, you'll be able to change the setting in `framework.ini`
2. The issue you mentioned in unrelated to this PR. See https://github.com/ppy/osu/discussions/26241.

THEGOLDENPRO:
> 1. This feature is not yet available, as it was merged after the latest release. When framework is updated, you'll be able to change the setting in `framework.ini`
> 2. The issue you mentioned in unrelated to this PR. See [Window resizes incorrectly on startup (Linux/Wayland) osu#26241](https://github.com/ppy/osu/discussions/26241).

ah I see, thank you for pointing me in the right direction. 👍

peppy:
> 1. This feature is not yet available, as it was merged after the latest release. When framework is updated, you'll be able to change the setting in `framework.ini`

It is planned to be exposed via settings right? I don't think we should ever recommend editing INI files directly to users.

bdach:
> It is planned to be exposed via settings right? I don't think we should ever recommend editing INI files directly to users.

https://github.com/ppy/osu/pull/26166

PR #6095, opened 2023-12-28
Implement concept of multiple focus targets

frenzibyte:
Preliminary step for removing the necessity of an input manager from `DropdownSearchBar`.

The end goal of this implementation is to make it possible for the dropdown search bar and menu to be focused together at once when the dropdown is open (in addition to `Dropdown` itself), to promote non-positional input to all those components.

After countless attempts at multiple approaches for this end goal, I've settled on this implementation as one with the least resistance when integrated with existing usages of focus in o!f and osu!.

The approach taken here revolves around a property in the `Drawable` class returning a list of drawables that should gain/lose focus alongside the main drawable providing them (in my case, the main drawable class would be `Dropdown`, and it would override the property to return both `Menu` and `DropdownHeader`->`DropdownSearchBar`->`TextBox`).

Test coverage made here is pretty basic (showcasing how the feature can be utilised), but I aim to add more cases if the approach is agreed upon.


smoogipoo:
My immediate concern with this is that it's a very specific way of handling this one case. But as soon as you have multiple focus in a case that's slightly less contained - suppose some `FocusedOverlayContainer` scenario, it becomes annoying to handle.

peppy:
I'm not sure if I had a dangling review that I lost, or I mentioned my thoughts somewhere else like discord but I came to roughly the same conclusion (aka. i hope we can avoid this).

I'd definitely agree on the direction of attempting to fix the underlying issue without this change.

PR #6097, opened 2023-12-29
Refactor dropdown search bar and fix many issues with it

frenzibyte:
- Closes https://github.com/ppy/osu/issues/25769
- Closes https://github.com/ppy/osu/issues/25850
- Closes https://github.com/ppy/osu/issues/26079
- [ ] Depends on #6095 

Using the concept of "multiple focus targets", now `Dropdown` gains focus when in an open state, and in addition sends focus events to `Menu` and indirectly to the search bar text box using the chain "`DropdownHeader` -> `DropdownSearchBar` -> `TextBox`".

This eliminates the necessity of an input manager inside `DropdownSearchBar`, which in turn resolves many issues that arose from it (such as [blocking key events from propagating past the search bar text box](https://github.com/ppy/osu/issues/25769), or [the search bar text box receiving focus despite being blocked by another component in front of it](https://github.com/ppy/osu/issues/26079)).

A necessary component in this implementation is to stop both `Menu` and `TextBox` from managing the focus state completely, using an internal boolean property. I'm not quite fond of the approach, but I did not want to overthink this more than I had done already (one idea I had that I scraped off was defining an `IUIControl` interface, and let `Menu`/`TextBox` avoid handling focus if there's a parent with the `IUIControl` interface, via DI or parent traversal).

As a side detail in this change, positional input on the search bar is now possible (i.e. you can drag the search bar text box to change selection).


osu.Framework/Graphics/UserInterface/DropdownSearchBar.cs
```diff
@@ -1,30 +1,39 @@
...
 
-        public Bindable<string> SearchTerm { get; } = new Bindable<string>();
+        protected internal override IEnumerable<Drawable> AdditionalFocusTargets
+        {
+            get
+            {
+                // On mobile platforms, let's not make the keyboard popup unless the dropdown is intentionally searchable.
+                // todo: however, this causes such dropdowns to not be searchable at all when using hardware keyboard on mobile.
+                if (host.OnScreenKeyboardOverlapsGameWindow && !AlwaysDisplayOnFocus)
+                    return Array.Empty<Drawable>();
```
frenzibyte, line 30: The change in https://github.com/ppy/osu-framework/pull/6080 has been moved here. Once a dropdown is focused, if `AlwaysDisplayOnFocus` is false and the platform shows a software keyboard, then the `TextBox` is excluded from being focused, therefore not showing any software keyboard when the dropdown is open.

frenzibyte:
Sorry about that, I've included test coverage for https://github.com/ppy/osu/issues/25769 and  https://github.com/ppy/osu/issues/26079, which are the issues fixed by the "multiple focus target" approach. Should be good enough to explore alternative options with. I've also pushed a separate branch for the tests in case needed: https://github.com/ppy/osu-framework/compare/master...frenzibyte:osu-framework:dropdown-search-bar-issues

PR #6098, opened 2023-12-31
Ignore input while `KeyBindingContainer` is not in a loaded state

frenzibyte:
- Closes https://github.com/ppy/osu/issues/23555

This issue is occurring due to the presence of a `PassThroughInputManager` layer above the key binding container. All input managers process input on `Update`, so even if a drawable has not received its first update frame, the input manager may propagate input changes to such drawable.

`KeyBindingContainer` gets into a weird state when handling input while it's not loaded. Since it's not loaded, `KeyBindings` would be null since it only gets populated in `LoadComplete`, so no press/release events are propagated to children for the first key. However, the key gets stored in `pressedInputKeys`, and later on, when the next key is pressed, the old key incorrectly becomes processed as a new binding as well, causing the double-press issue.

For simplicity, I've made `handleNewPressed` return early if the key binding container is not loaded yet. 

That being said, it should be taken into consideration that a drawable receiving input events before it's loaded is highly unexpected, and may need to be iterated upon at a later point.

### DIFF
--- a/osu.Framework.Tests/Visual/Input/TestSceneKeyBindingContainer.cs
+++ b/osu.Framework.Tests/Visual/Input/TestSceneKeyBindingContainer.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using NUnit.Framework;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.UserInterface;
+using osu.Framework.Input;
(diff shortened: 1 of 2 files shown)


frenzibyte:
Closing in favour of https://github.com/ppy/osu-framework/pull/6101, see OP in #6101 for more information.

PR #6100, opened 2023-12-31
Fix android storage path on .NET 8

i3ym:
As per https://github.com/dotnet/runtime/pull/76250 and https://github.com/dotnet/docs/issues/31423, `System.Environment.SpecialFolder.Profile` since .NET 8 points to `$HOME/Documents` instead of `$HOME`, causing
```
android.runtime.JavaProxyThrowable: [System.IO.DirectoryNotFoundException]: IO_PathNotFound_Path, /data/user/0/com.test.app/files/Documents
```
on startup.
This PR changes it to always target `$HOME`.

Similar change was previously done in https://github.com/ppy/osu-framework/pull/5909/commits/d7ff5d6478ada5017518aa19843f3fdca0d174cf, but android was skipped.


Susko3 (approved):
Can confirm the crash when running/building `SampleGame.Android` with .NET 8. And this change fixes it. No regressions when running .NET 6.

smoogipoo (approved):

PR #6101, opened 2024-01-01
Fix `InputManager` including non-loaded direct children in input queue

frenzibyte:
- Closes https://github.com/ppy/osu/issues/23555

`KeyBindingContainer` gets into a weird state when handling input while it's not loaded. Since it's not loaded, `KeyBindings` would be null since it only gets populated in `LoadComplete`, so no press/release events are propagated to children for the first key. However, the key gets stored in `pressedInputKeys`, and later on, when the next key is pressed, the old key incorrectly becomes processed as a new binding as well, causing the double-press issue.

`KeyBindingContainer` was not supposed to enter the input queue if it's not fully loaded in the first place, but turns out `InputManager` was incorrectly adding it (by bypassing `ShouldBeConsideredForInput`).

### DIFF
--- a/osu.Framework.Tests/Visual/Input/TestSceneKeyBindingContainer.cs
+++ b/osu.Framework.Tests/Visual/Input/TestSceneKeyBindingContainer.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using NUnit.Framework;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.UserInterface;
+using osu.Framework.Input;
 using osu.Framework.Input.Bindings;
 using osu.Framework.Input.Events;
 using osu.Framework.Testing;
@@ -48,7 +49,7 @@ namespace osu.Framework.Tests.Visual.Input
         }
 
         [Test]
-        public void TestPressKeyBeforeKeyBindingContainerAdded()
+        public void TestPressHalfCombinationBeforeKeyBindingContainerAdded()
         {
             List<TestAction> pressedActions = new List<TestAction>();
             List<TestAction> releasedActions = new List<TestAction>();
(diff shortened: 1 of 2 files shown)


osu.Framework/Input/InputManager.cs
```diff
@@ -593,8 +593,12 @@ private SlimReadOnlyListWrapper<Drawable> buildNonPositionalInputQueue()
                 FrameStatistics.Increment(StatisticsCounterType.InputQueue);
 
             var children = AliveInternalChildren;
+
             for (int i = 0; i < children.Count; i++)
-                children[i].BuildNonPositionalInputQueue(inputQueue);
+            {
+                if (ShouldBeConsideredForInput(children[i]))
+                    children[i].BuildNonPositionalInputQueue(inputQueue);
```
bdach, line 600: Two things:

The silent direct implication of this is that:
- [non-loaded drawables](https://github.com/ppy/osu-framework/blob/3eb7a688a5ba9edf0e23d2e3187ccec36266a1c7/osu.Framework/Graphics/Containers/CompositeDrawable.cs#L1407) will not be included here (likely fine)
- [non-current screens](https://github.com/ppy/osu-framework/blob/3eb7a688a5ba9edf0e23d2e3187ccec36266a1c7/osu.Framework/Screens/ScreenStack.cs#L375) will not either (likely fine)

Given that `CompositeDrawable` already respects that method with the aforementioned implications, this seems likely fine, _but_ that doesn't answer the question why this is only applied to `InputManager.buildNonPositionalInputQueue()` but not to

https://github.com/ppy/osu-framework/blob/3eb7a688a5ba9edf0e23d2e3187ccec36266a1c7/osu.Framework/Input/InputManager.cs#L615-L628

I'm pretty sure it should be applied to both.

osu.Framework.Tests/Visual/Input/TestSceneKeyBindingContainer.cs
```diff
@@ -81,6 +82,77 @@ public void TestPressKeyBeforeKeyBindingContainerAdded()
...
+            AddAssert("no actions released", () => releasedActions, () => Is.Empty);
+
+            AddStep("release key A", () => InputManager.ReleaseKey(Key.A));
+            AddAssert("only one action triggered", () => pressedActions, () => Has.Count.EqualTo(1));
+            AddAssert("only one action released", () => releasedActions, () => Has.Count.EqualTo(1));
+            AddAssert("ActionA released", () => releasedActions[0], () => Is.EqualTo(TestAction.ActionA));
+        }
+
+        [Test]
+        public void TestPressKeyBeforeKeyBindingContainerAdded_WithPassThroughInputManager()
```
bdach, line 120: One thing that I wondered with this test is what happens once <kbd>Enter</kbd> is released after the key binding container is loaded / added. And the answer to that is nothing happens, i.e. the input manager does not receive the release of the <kbd>Enter</kbd> key. Which is probably a good thing, as it's better to send nothing than half a press, but it still has me worried about the implications on game usages.

So this is probably okay, but I'd appreciate assurances that nothing falls over game-side with this change applied, i.e. running game-side tests and making sure nothing fails.

bdach (changes requested):

osu.Framework/Input/InputManager.cs
frenzibyte: Good catch, totally forgot about that one.

osu.Framework.Tests/Visual/Input/TestSceneKeyBindingContainer.cs
frenzibyte: > One thing that I wondered with this test is what happens once Enter is released after the key binding container is loaded / added. And the answer to that is nothing happens, i.e. the input manager does not receive the release of the Enter key. Which is probably a good thing, as it's better to send nothing than half a press, but it still has me worried about the implications on game usages.

If by "input manager" you mean the key binding container, then that is correct. I had run game-side tests beforehand and ensured nothing broke, but I'll check again to confirm.

bdach (approved):

PR #6104, opened 2024-01-04
Set process priority of framework window to high when focused

peppy:
This is something we do on osu!stable and has large gains (for some users) on desktop systems, especially when shared with other heavy applications.

Tested on windows and macOS. Windows works as expected, macOS fails due to lack of permissions. Would appreciate linux testing.

If deemed more appropriate, can be moved to `WindowsGameHost`.

### DIFF
--- a/osu.Framework/Platform/DesktopGameHost.cs
+++ b/osu.Framework/Platform/DesktopGameHost.cs
@@ -112,6 +112,32 @@ namespace osu.Framework.Platform
             return ipcProvider.SendMessageAsync(message);
         }
 
+        protected override void OnActivated()
+        {
+            setGamePriority(true);
+            base.OnActivated();
+        }
+
+        protected override void OnDeactivated()
+        {
+            setGamePriority(false);
+            base.OnDeactivated();
+        }
+
+        private void setGamePriority(bool active)
+        {
+            try
+            {
+                // We set process priority after the window becomes active, because for whatever reason windows will
+                // reset this when the window becomes active after being inactive when game mode is enabled.
+                Process.GetCurrentProcess().PriorityClass = active ? ProcessPriorityClass.High : ProcessPriorityClass.Normal;
+            }
+            catch
+            {
+                // This doesn't work on all operating systems.
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


Susko3:
Have you considered how this may affect headless testing? `HeadlessGameHost` derives from `DesktopGameHost` and is considered to be [active by default](https://github.com/ppy/osu-framework/blob/1d37c4707a3a6b75a5215108b049a917264d3e8d/osu.Framework/Platform/GameHost.cs#L80). Tests probably shouldn't be using high priority.

bdach:
I'm not sure it realistically matters until proven otherwise. Do you have some specific concern in mind?

Susko3:
My concerns mostly come from the tone of the [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.processpriorityclass?view=net-8.0): "Use extreme care when specifying High for the process's priority class [...]"
And this affecting headless tests being a "hidden" change.

I've just run the full framework test suite and haven't noticed any slowdowns on my machine, so it's probably fine.

bdach:
Linux also fails on lack of permissions (only when attempting to set to high).

```
System.ComponentModel.Win32Exception (13): Permission denied
   at System.Diagnostics.Process.set_PriorityClassCore(ProcessPriorityClass value)
   at System.Diagnostics.Process.set_PriorityClass(ProcessPriorityClass value)
   at osu.Framework.Platform.DesktopGameHost.setGamePriority(Boolean active) in /home/dachb/Documents/opensource/osu-framework/osu.Framework/Platform/DesktopGameHost.cs:line 133
```

bdach (dismissed):
Seems probably fine to me. Let's see how this pans out.

@peppy your call on whether you want to address the tests concern brought up above.

peppy:
i think it might be best to, but hesitant because I'm not sure how to structure it without thinking for more than a few seconds.

bdach:
Probably make `setGamePriority()` virtual and suppress in the headless impl? Bit dirty but at this point...

bdach (approved):

PR #6105, opened 2024-01-04
Migrate Android framework to SDL

hwsmm:
This PR lets Android devices use SDL for windowing/input. osuTK usage got removed in the process of doing it.
It also removes Android input handling code and replaces them with existing SDL components.

I did test using Vulkan with it and it kind of works, but it was just full of glitches (a lot of flashes and tearing). Veldrid backends (OpenGL/Vulkan) crash if they are brought up from background. I don't know well about graphics, anyway.. the only graphics backend that works well is Legacy OpenGL.

Java is needed because SDL Android heavily relies on JNI, and its binding is made with [Xamarin.Android](https://learn.microsoft.com/en-us/xamarin/android/platform/binding-java-library/).

I didn't reuse the code, but brought an idea from https://github.com/0x0ade/SDL2Droid-CS, it helped me a lot!

Tested on my tablet with Android 14.

You can use [this osu! branch](https://github.com/hwsmm/osu/tree/sdl-cs-android) if you want to test.

Screenshot_20240103_162707


bdach:
Interesting that you got this to work (I tried what sounds to be the exact same approach and it didn't work - input was broken, and so was debugging). Interested to see if this actually works.

The packaging 200% needs work though. We cannot be consuming random binaries from random sources. The build toolchain for our SDL2 fork (https://github.com/ppy/SDL2-CS) must be used for this. Maybe the extra wrapper code must live there too, and if possible, the changes to the SDL Android project should be upstreamed.

hwsmm:
I understand that reproducibility and integrity matter a lot in packaging, so I even considered excluding my binaries from this PR, but I thought it would cause some big inconvenience for testing, so I just went with having binaries included. I should have mentioned it.

I will look into that repo and open another PR there if I come up with an okay solution.

Susko3:
> SDL Android looks for the SDL main function from dynamic library symbol. Let me know if there's any better way. I'd be happy to make it better.

You could try using [`UnmanagedCallersOnlyAttribute`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.unmanagedcallersonlyattribute?view=net-8.0) for this. I think this will expose the function from the managed `osu.Framework.Android.dll` or `SampleGame.Android.dll` so hopefully SDL might be able to see it. I think that this requires compiling with NativeAOT.

With this, `SDL2AndroidMainSetter` should be completely unnecessary.

[Example code from another project](https://github.com/kostya9/NetObsBindings/blob/d1c2200b8e7ef6b940c35c1f3818deaa05481569/samples/SimplePlugin/ObsPlugin.cs#L30-L41), here the native library expects a `obs_module_load` to be exposed on the DLL, this project is compiled with NativeAOT and the native library can see and call the function.

bdach:
> I think that this requires compiling with NativeAOT.

I'm not sure it will be relevant but the last time the words "AOT" and "android" were in close proximity [things fizzled out very quickly due to being terminally broken](https://github.com/ppy/osu/pull/24500). So I'm not holding my breath...

hwsmm:
I've run into some issues with Xamarin toolchain, mainly JAR bindings not working if I have .NET 8 installed, it took some good hours to point out the issue. I eventually used another PC to work on it, and I _think_ I got it. There are now no changes in Java code, too.

Above PR adds SDL2-CS-Android project to SDL2-CS, and I'll commit changes that are needed from framework-side once it gets merged, since there isn't a NuGet package to depend on yet.

You can test it by removing `osu.Framework.Android.SDL2` reference from `osu.Framework.Android` and add a reference to `SDL2-CS-Android` in `osu.Framework.Android`.

Susko3:
https://github.com/libsdl-org/SDL/commit/3a482ebae062cbc127e4e0faf9a07f69b54b7419 could be used to avoid the need for `SDL2AndroidMainSetter`. It's currently only available in SDL3, but we could backport it to SDL2.

hwsmm:
I will just push my local framework as local SDL2-CS is needed to test at the moment anyway, and tidy up once things look okay.

With @Susko3's suggestion, I got it working and successfully removed SDL2AndroidMainSetter! You can test it by using [SDL2-CS update-android-binaries branch](https://github.com/hwsmm/SDL2-CS/tree/update-android-binaries).

Joehuu:
On the framework apps (test and samplegame), I get a potential crash when rotating 90 degrees (osu! not affected as it can't rotate 90):
```
02-10 14:12:41.974 12820 12853 E AndroidRuntime: FATAL EXCEPTION: SDLThread
02-10 14:12:41.974 12820 12853 E AndroidRuntime: Process: osu.Framework.Tests.Android, PID: 12820
02-10 14:12:41.974 12820 12853 E AndroidRuntime: android.runtime.JavaProxyThrowable: [NUnit.Framework.AssertionException]: Stored Displays don't match actual displays: Stored displays:
02-10 14:12:41.974 12820 12853 E AndroidRuntime:   Name: 0, Bounds: {X=0,Y=0,Width=2400,Height=1080}, DisplayModes: 0, Index: 0
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 
02-10 14:12:41.974 12820 12853 E AndroidRuntime: Actual displays:
02-10 14:12:41.974 12820 12853 E AndroidRuntime:   Name: 0, Bounds: {X=0,Y=0,Width=1080,Height=2400}, DisplayModes: 0, Index: 0
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at osu.Framework.Logging.ThrowingTraceListener.Fail(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at System.Diagnostics.TraceInternal.Fail(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at System.Diagnostics.TraceInternal+TraceProvider.Fail(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at System.Diagnostics.Debug.Fail(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at System.Diagnostics.Debug.Assert(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at osu.Framework.Platform.SDL2Window.assertDisplaysMatchSDL(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at osu.Framework.Platform.SDL2Window.<handleWindowEvent>b__290_0(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at osu.Framework.Threading.ScheduledDelegate.InvokeTask(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at osu.Framework.Threading.ScheduledDelegate.RunTaskInternal(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at osu.Framework.Threading.Scheduler.Update(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at osu.Framework.Platform.SDL2Window.RunFrame(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at osu.Framework.Platform.SDL2Window.RunMainLoop(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at osu.Framework.Platform.SDL2Window.Run(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at osu.Framework.Platform.GameHost.Run(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at osu.Framework.Android.AndroidGameActivity.<CreateSDLMainRunnable>b__17_0(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at Java.Lang.Runnable.Run(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at Java.Lang.IRunnableInvoker.n_Run(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PP_V(Unknown Source:0)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at mono.java.lang.Runnable.n_run(Native Method)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at mono.java.lang.Runnable.run(Runnable.java:31)
02-10 14:12:41.974 12820 12853 E AndroidRuntime: 	at java.lang.Thread.run(Thread.java:1012)
```
---
This also resolves https://github.com/ppy/osu/issues/5236 for android, but there's a resizing bug which may be a blocker as it's really broken:

https://github.com/ppy/osu-framework/assets/35318437/de145a8e-8590-48f9-a084-7b9d87e70ddb

Joehuu (changes requested):
As commented above.

hwsmm:
I actually have run into that crash when I was writing this PR, but didn't fix it eventually since `assertDisplaysMatchSDL` is only present in Debug configuration. I fixed it in my last commit.
https://github.com/ppy/osu-framework/blob/0e887087dad9c7c774d6f03c61403bf7d9fd7302/osu.Framework/Platform/SDL2Window_Windowing.cs#L309-L317

I can't reproduce the resizing bug in the video you posted, In the latest release (osuTK), my game doesn't get moved/resized at all, so I can't see what I am typing because keyboard hides bottom half, and in my build (SDL2), the game goes up (only showing bottom half), so I can see my chat fine.

My phone is Galaxy A51 on Android 13, so something may be different with keyboard handling, but I'm not really sure..

Joehuu:
I can't seem to reproduce anymore. Probably incidentally fixed with the new commits.

It still has an off-by-one screen offset bug:
- click/focus chat text box (no screen offset)
- click chat text box again (puts it into view)
- click settings text box (still offsets the screen where the chat text box used to be)

bdach:
I've added the package, and reverted the `.sln` changes since they seemed superfluous and `.sln`s are a pain to review properly.

I did a smoke test, and while I would _loooove_ to get this in, I'm hesitant. Despite the fact that this generally works better than when I tried this - on a brief smoke test on device 1 (yet to test device 2) things generally seem to work, aside from small issues like back button not working like it used to - but the debuggability is **zero**. After a certain point in the game's launch sequence the activity just dies with debugger attached. I've never really loved the xamarin debugging experience - it's always sucked - but at least the debugger was _available_ and now it's not.

I dunno how to feel about that, especially so that I'm the most likely to be on the hook when it comes to maintaining this in the future.

I can't reproduce @Joehuu's reported issue either, for whatever that's worth.

bdach:
Aaaaand on test device number 2 this just dies on startup:

```
2024-04-19 12:21:51.912 13687-13706 AndroidRuntime          osu.Framework.Tests.Android          E  FATAL EXCEPTION: SDLThread
                                                                                                    Process: osu.Framework.Tests.Android, PID: 13687
                                                                                                    android.runtime.JavaProxyThrowable: [System.TypeLoadException]: Could not resolve type with token 010000a2 from typeref (expected class 'SDL.Utf8String' in assembly 'SDL3-CS, Version=2024.418.1.0, Culture=neutral, PublicKeyToken=null')
                                                                                                    	at Java.Lang.Runnable.Run(Unknown Source)
                                                                                                    	at Java.Lang.IRunnableInvoker.n_Run(Unknown Source)
                                                                                                    	at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PP_V(Unknown Source)
                                                                                                    	at mono.java.lang.Runnable.n_run(Native Method)
                                                                                                    	at mono.java.lang.Runnable.run(Runnable.java:31)
                                                                                                    	at java.lang.Thread.run(Thread.java:762)
```

I dunno. I was hoping for easy wins here but this and the lack of debugger is probably going to take weeks to months to diagnose and fix if one wanted to.

smoogipoo:
> Aaaaand on test device number 2 this just dies on startup:

That one sounds like it's possibly using the wrong lib. Try clean rebuild, and by that I mean `find ./osu -name 'bin' -type d -print0 | xargs -0 rm -rd`.

osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -31,37 +34,55 @@ public abstract class AndroidGameActivity : Activity
...
+        protected override SDLSurface CreateSDLSurface(Context? context) => new AndroidGameSurface(this, context);
 
-        public override void OnTrimMemory([GeneratedEnum] TrimMemory level)
+        protected override IRunnable CreateSDLMainRunnable() => new Runnable(() =>
         {
-            base.OnTrimMemory(level);
-            gameView.Host?.Collect();
-        }
+            // blocks back button
+            SDL.SDL3.SDL_SetHint(SDL.SDL3.SDL_HINT_ANDROID_TRAP_BACK_BUTTON, "1"u8);
```
bdach, line 63: Why? I expected back button not working being some SDL-shaped change in behaviour, but why do this explicitly?

smoogipoo:
> Not sure why you deleted comment

Hmm, I deleted it because the exception says `in assembly 'SDL3-CS, Version=2024.418.1.0`, which _is_ the latest version. Bizzare.

osu.Framework.Android/AndroidGameActivity.cs
hwsmm: https://github.com/ppy/osu-framework/blob/af8bbd2d56d615b2987f7ad0a08aaeed312662db/osu.Framework.Android/AndroidGameActivity.cs#L132-L136

I added it because the original implementation had above code, but it seems that back button is mapped as escape in [the current implementation](https://github.com/ppy/osu-framework/blob/master/osu.Framework.Android/Input/AndroidKeyboardHandler.cs#L128-L129). It's now fixed by adding the key (SDL_SCANCODE_AC_BACK) in SDL3Extension.

osu.Framework/Platform/SDL/SDL3GraphicsSurface.cs
```diff
@@ -216,5 +216,13 @@ bool IOpenGLGraphicsSurface.VerticalSync
         bool ILinuxGraphicsSurface.IsWayland => window.IsWayland;
 
         #endregion
+
+        #region Android-specific implementation
+
+#pragma warning disable CA1416 // This is only usable in Android
+        IntPtr IAndroidGraphicsSurface.JniEnvHandle => SDL3.SDL_AndroidGetJNIEnv();
+#pragma warning restore CA1416
```
Susko3, line 224:
```suggestion
        [SupportedOSPlatform("android")]
        IntPtr IAndroidGraphicsSurface.JniEnvHandle => SDL3.SDL_AndroidGetJNIEnv();
```

Strangely, this won't cause platform warnings elsewhere.

osu.Framework/Platform/SDL3Window_Windowing.cs
```diff
@@ -226,7 +226,9 @@ public Size MaxSize
 
         public IBindable<bool> IsActive => isActive;
 
-        private readonly BindableBool cursorInWindow = new BindableBool();
+        // This is default to true because Android never gets window enter event.
+        // shouldn't harm so much to have true as default
+        private readonly BindableBool cursorInWindow = new BindableBool(true);
```
Susko3, line 231: This needs to default to false. See https://github.com/ppy/osu-framework/pull/4402.

Opening the game with event logging enabled I'm seeing the mouse enter and focus gained events. Not sure why these aren't handled in the event loop.

```cs
SDL3.SDL_SetHint(SDL3.SDL_HINT_EVENT_LOGGING, "1"u8);
```

```cs
2024-04-19 18:11:12.625 24271-24299 DOTNET                  osu.Framework.Tests.Android          I  [runtime] 2024-04-19 16:11:12 [verbose]: SDL application log [info]: SDL EVENT: SDL_EVENT_WINDOW_MOUSE_ENTER (timestamp=32486847 windowid=2 event=SDL_EVENT_WINDOW_MOUSE_ENTER data1=0 data2=0)
2024-04-19 18:11:12.627 24271-24299 DOTNET                  osu.Framework.Tests.Android          I  [runtime] 2024-04-19 16:11:12 [verbose]: SDL application log [info]: SDL EVENT: SDL_EVENT_WINDOW_FOCUS_GAINED (timestamp=36083693 windowid=2 event=SDL_EVENT_WINDOW_FOCUS_GAINED data1=0 data2=0)
```

If you must manually set `CursorInWindow`, you can do it in `AndroidGameWindow.Create()` in a way that makes it obvious that this is a hack:

```cs
((BindableBool)CursorInWindow).Value = true;
```

osu.Framework/Platform/SDL3Window_Windowing.cs
```diff
@@ -526,6 +528,7 @@ private unsafe void handleWindowEvent(SDL_WindowEvent evtWindow)
                 case SDL_EventType.SDL_EVENT_WINDOW_FOCUS_LOST:
                 case SDL_EventType.SDL_EVENT_WINDOW_SHOWN:
                 case SDL_EventType.SDL_EVENT_WINDOW_HIDDEN:
+                case SDL_EventType.SDL_EVENT_WINDOW_RESIZED when RuntimeInfo.IsMobile:
```
Susko3, line 531: Please scope to android as mobile includes ios.

`SDL_EVENT_DISPLAY_ORIENTATION` should handle this failure case, but it seems SDL is sending that event before the displays are actually updated. Something to report upstream.

```cs
2024-04-19 18:21:31.093 25803-25803 DOTNET                  osu.Framework.Tests.Android          I  [runtime] 2024-04-19 16:21:31 [verbose]: SDL application log [info]: SDL EVENT: SDL_EVENT_DISPLAY_ORIENTATION (timestamp=3414693397 display=1 event=SDL_EVENT_DISPLAY_ORIENTATION data1=2)
2024-04-19 18:21:31.174 25803-25803 SDL                     osu.Framework.Tests.Android          V  onConfigurationChanged()
2024-04-19 18:21:31.178 25803-25803 OplusZoomW...ViewHelper osu.Framework.Tests.Android          V  setLastReportedMergedConfiguration mZoomDisplayHeight: 1080 getDecorView.DecorView@6173a96[TestGameActivity]
2024-04-19 18:21:31.216 25803-25803 SDL                     osu.Framework.Tests.Android          V  surfaceChanged()
2024-04-19 18:21:31.216 25803-25803 SDL                     osu.Framework.Tests.Android          V  Window size: 2400x1080
2024-04-19 18:21:31.216 25803-25803 SDL                     osu.Framework.Tests.Android          V  Device size: 2400x1080
2024-04-19 18:21:31.301 25803-25803 DOTNET                  osu.Framework.Tests.Android          I  [runtime] 2024-04-19 16:21:31 [verbose]: SDL application log [info]: SDL EVENT: SDL_EVENT_WINDOW_RESIZED (timestamp=3538189628 windowid=2 event=SDL_EVENT_WINDOW_RESIZED data1=2400 data2=1080)
2024-04-19 18:21:31.302 25803-25803 DOTNET                  osu.Framework.Tests.Android          I  [runtime] 2024-04-19 16:21:31 [verbose]: SDL application log [info]: SDL EVENT: SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED (timestamp=3623588397 windowid=2 event=SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED data1=2400 data2=1080)
2024-04-19 18:21:31.307 25803-25803 Surface                 osu.Framework.Tests.Android          D  lockCanvas
2024-04-19 18:21:31.333 25803-25803 Surface                 osu.Framework.Tests.Android          D  lockCanvas
2024-04-19 18:21:31.343 25803-25841 AndroidRuntime          osu.Framework.Tests.Android          E  FATAL EXCEPTION: SDLThread
                                                                                                    Process: osu.Framework.Tests.Android, PID: 25803
                                                                                                    android.runtime.JavaProxyThrowable: [NUnit.Framework.AssertionException]: Stored Displays don't match actual displays: Stored displays:
                                                                                                      Name: 1, Bounds: {X=0,Y=0,Width=1080,Height=2400}, DisplayModes: 0, Index: 0
                                                                                                    
                                                                                                    Actual displays:
                                                                                                      Name: 1, Bounds: {X=0,Y=0,Width=2400,Height=1080}, DisplayModes: 0, Index: 0
```

osu.Framework/Graphics/Veldrid/VeldridDevice.cs
```diff
@@ -166,6 +166,13 @@ public VeldridDevice(IGraphicsSurface graphicsSurface)
                         : SwapchainSource.CreateXlib(this.graphicsSurface.DisplayHandle, this.graphicsSurface.WindowHandle);
                     break;
                 }
+
+                case RuntimeInfo.Platform.Android:
+                {
+                    var androidGraphics = (IAndroidGraphicsSurface)graphicsSurface;
+                    swapchain.Source = SwapchainSource.CreateAndroidSurface(graphicsSurface.DisplayHandle, androidGraphics.JniEnvHandle);
```
Susko3, line 173:
```suggestion
                    var androidGraphics = (IAndroidGraphicsSurface)this.graphicsSurface;
                    swapchain.Source = SwapchainSource.CreateAndroidSurface(this.graphicsSurface.DisplayHandle, androidGraphics.JniEnvHandle);
```

osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -31,37 +34,55 @@ public abstract class AndroidGameActivity : Activity
...
         protected abstract Game CreateGame();
 
+        private Game? game;
+
+        private GameHost? host;
+
         /// <summary>
         /// Whether this <see cref="AndroidGameActivity"/> is active (in the foreground).
         /// </summary>
         public BindableBool IsActive { get; } = new BindableBool();
```
Susko3, line 48: This is now unused, you can remove it.

osu.Framework.Android/AndroidGameSurface.cs
```diff
@@ -0,0 +1,122 @@
...
+using osu.Framework.Bindables;
+using Android.Views;
+using AndroidX.Core.View;
+using AndroidX.Window.Layout;
+
+namespace osu.Framework.Android
+{
+    internal class AndroidGameSurface : SDLSurface
+    {
+        public AndroidGameActivity Activity { get; } = null!;
```
Susko3, line 20: Can be made `private`.

osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -31,37 +34,55 @@ public abstract class AndroidGameActivity : Activity
...
         {
-            base.OnTrimMemory(level);
-            gameView.Host?.Collect();
-        }
+            // blocks back button
+            SDL.SDL3.SDL_SetHint(SDL.SDL3.SDL_HINT_ANDROID_TRAP_BACK_BUTTON, "1"u8);
+
+            // hints are here because they don't apply well in another location such as SDL2Window
+
+            host = new AndroidGameHost(this);
```
Susko3, line 67: The `GameHost` should be in a `using` statement as the `host.Run()` call is blocking.

osu.Framework.Android/AndroidGameWindow.cs
Susko3, line 1: Needs to have custom `UpdateWindowStateAndSize()` logic, same as `iOSWindow`. SDL gives a bunch of errors when setting the window mode and doesn't always hide the status bar.

 Honestly, I'd say you should create a `SDL3MobileWindow` class that houses this common code.

https://github.com/ppy/osu-framework/blob/af8bbd2d56d615b2987f7ad0a08aaeed312662db/osu.Framework.iOS/IOSWindow.cs#L39-L45

osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -31,37 +34,55 @@ public abstract class AndroidGameActivity : Activity
 
         protected const LaunchMode DEFAULT_LAUNCH_MODE = LaunchMode.SingleInstance;
 
+        internal static AndroidGameSurface Surface => (AndroidGameSurface)MSurface!;
+
         protected abstract Game CreateGame();
 
+        private Game? game;
```
Susko3, line 41: Doesn't need to be stored. Might as well have the `CreateGame()` call directly in `host.Run(CreateGame())`.

Susko3 (changes requested):
I get this crash when tapping the screen in a certain way:

1. Tap and hold first finger on screen
2. Tap and hold second finger
3. Release first finger
4. Press first finger again

```cs
An unhandled error has occurred.
NUnit.Framework.AssertionException: :
at osu.Framework.Logging.ThrowingTraceListener.Fail(String message1, String message2)
at System.Diagnostics.TraceInternal.Fail(String message, String detailMessage)
at System.Diagnostics.TraceInternal.TraceProvider.Fail(String message, String detailMessage)
at System.Diagnostics.Debug.Fail(String message, String detailMessage)
at System.Diagnostics.Debug.Assert(Boolean condition)
at osu.Framework.Platform.SDL3Window.HandleTouchFingerEvent(SDL_TouchFingerEvent evtTfinger)
at osu.Framework.Platform.SDL3Window.HandleEvent(SDL_Event e)
at osu.Framework.Platform.SDL3Window.pollSDLEvents()
at osu.Framework.Platform.SDL3Window.RunFrame()
at osu.Framework.Platform.SDL3Window.RunMainLoop()
at osu.Framework.Platform.SDL3Window.Run()
at osu.Framework.Platform.GameHost.Run(Game game)
at osu.Framework.Android.AndroidGameActivity.<CreateSDLMainRunnable>b__13_0()
at Java.Lang.Runnable.Run()
at Java.Lang.IRunnableInvoker.n_Run(IntPtr jnienv, IntPtr native__this)
at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PP_V(_JniMarshal_PP_V callback, IntPtr jnienv, IntPtr klazz)
```

[logs from first touch down to crash.txt](https://github.com/ppy/osu-framework/files/15043393/logs.from.first.touch.down.to.crash.txt)

Joehuu:
> I can't reproduce @Joehuu's reported issue either, for whatever that's worth.

Tested on three devices of varying versions: Google Pixel 6a (Android 14), Moto G Play (2021, Android 11), and Moto G5 Plus (Android 8) and all exhibit the same problem. Here's a video of framework showing the same thing:

https://github.com/ppy/osu-framework/assets/35318437/a5810b41-edef-4b8b-a976-812287f67522

osu.Framework/Platform/SDL3Window_Windowing.cs
Susko3: - Reported the stale data issue upstream: https://github.com/libsdl-org/SDL/issues/9585

You should mention this issue in the code as a TODO.

hwsmm:
I applied most of the review suggestions, but there seem to be some remaining issues (keyboard screen offsets and multi-touch assert fail for now?), but I currently don't have much time to actually investigate into them.. I should be more free after a week, so I'll look into them then.

Though if anyone suggests fixes for the forementioned issues in the meantime, I'll try applying them.

Susko3:
The main problem with the keyboard is that we're calling `SDL_StartTextInput` before [`SDL_SetTextInputRect`](https://wiki.libsdl.org/SDL3/SDL_SetTextInputRect). This leads to the keyboard using the stale text input rect value.

From `SDL_SetTextInputRect` documentation:

> To start text input in a given location, this function is intended to be called before [SDL_StartTextInput](https://wiki.libsdl.org/SDL3/SDL_StartTextInput), although some platforms support moving the rectangle even while text input (and a composition) is active.

I'll try to figure something out.

osu.Framework.Android/AndroidGameActivity.cs
Susko3: Hmm, maybe it shouldn't be disposed if `AndroidGameHost.CanExit` is `false`.

```cs
[runtime] 2024-04-20 17:36:17 [error]: An unhandled error has occurred.
[runtime] 2024-04-20 17:36:17 [error]: System.InvalidOperationException: Exit must be called before the GameHost is disposed.
[runtime] 2024-04-20 17:36:17 [error]: at osu.Framework.Platform.GameHost.Dispose(Boolean disposing)
[runtime] 2024-04-20 17:36:17 [error]: at osu.Framework.Platform.GameHost.Dispose()
[runtime] 2024-04-20 17:36:17 [error]: at osu.Framework.Android.AndroidGameActivity.<CreateSDLMainRunnable>b__7_0()
[runtime] 2024-04-20 17:36:17 [error]: at Java.Lang.Runnable.Run()
[runtime] 2024-04-20 17:36:17 [error]: at Java.Lang.IRunnableInvoker.n_Run(IntPtr jnienv, IntPtr native__this)
[runtime] 2024-04-20 17:36:17 [error]: at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PP_V(_JniMarshal_PP_V callback, IntPtr jnienv, IntPtr klazz)
```

Susko3:
I've reported https://github.com/ppy/osu-framework/pull/6105#pullrequestreview-2011866789 upstream in https://github.com/libsdl-org/SDL/issues/9591. This issue is not unique to android and can fail on Windows too.

osu.Framework/Platform/SDL3Window_Windowing.cs
Susko3: It seems we're setting the event filter too late, and those events are getting dropped:
- https://github.com/libsdl-org/SDL/issues/9592

bdach:
@Susko3 is this fine to move forward after all of the other PRs?

Susko3:
Yep this is good to go, just needs merge master + removal of no longer necessary workarounds:
1. Revert `SDL3Window_Windowing.cs`
2. `AndroidGameWindow.Create()` workarounds

Also [the `AndroidGameHost` shouldn't be disposed](https://github.com/ppy/osu-framework/pull/6105#discussion_r1573351603).

bdach:
> Revert `SDL3Window_Windowing.cs`

All of it? What about [this thing](https://github.com/ppy/osu-framework/pull/6105/files#diff-29dfec3d1a572a8fa40cc6228f95219bda6ccc551190bee986d3e194d0ecd03aR530-R531)? I don't see it being moved anywhere?

Susko3:
> What about [this thing](https://github.com/ppy/osu-framework/pull/6105/files#diff-29dfec3d1a572a8fa40cc6228f95219bda6ccc551190bee986d3e194d0ecd03aR530-R531)? I don't see it being moved anywhere?

You're correct, that has to stay, I must have confused it with the dropped events that were fixed with https://github.com/ppy/osu-framework/pull/6259.

bdach (dismissed):

osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -44,11 +44,6 @@ public abstract class AndroidGameActivity : SDLActivity
 
         protected override IRunnable CreateSDLMainRunnable() => new Runnable(() =>
         {
-            // blocks back button
-            SDL.SDL3.SDL_SetHint(SDL.SDL3.SDL_HINT_ANDROID_TRAP_BACK_BUTTON, "1"u8);
-
-            // hints are here because they don't apply well in another location such as SDL3Window
-
```
hwsmm: The reason why I didn't remove them was because SDL docs had this comment on the hint:
```
 * This is necessary for the right mouse button to work on some Android
 * devices, or to be able to trap the back button for use in your code
 * reliably. If this hint is true, the back button will show up as an
 * SDL_EVENT_KEY_DOWN / SDL_EVENT_KEY_UP pair with a keycode of
 * SDL_SCANCODE_AC_BACK.
 *
 * The variable can be set to the following values:
 *
 * - "0": Back button will be handled as usual for system. (default)
 * - "1": Back button will be trapped, allowing you to handle the key press
 *   manually. (This will also let right mouse click work on systems where the
 *   right mouse button functions as back.)
```

https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_hints.h#L132-L135

I should have made it more clear..

Not sure how it works without the hint though, but I guess it may be fine if it works?

bdach: Seemed fine to me /shrug

osu.Framework.Android/AndroidGameSurface.cs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        public override WindowInsets? OnApplyWindowInsets(WindowInsets? insets)
+        {
+            updateSafeArea(insets);
+            return base.OnApplyWindowInsets(insets);
+        }
+
+        /// <summary>
+        /// Updates the <see cref="IWindow.SafeAreaPadding"/>, taking into account screen insets that may be obstructing this <see cref="AndroidGameView"/>.
```
Susko3, line 69: Should be updated to mention the proper class, and remove `using osuTK.Android;`

```suggestion
        /// Updates the <see cref="IWindow.SafeAreaPadding"/>, taking into account screen insets that may be obstructing this <see cref="AndroidGameSurface"/>.
```

osu.Framework.Android/osu.Framework.Android.csproj
```diff
@@ -19,6 +19,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="ppy.osuTK.Android" Version="1.0.211" />
```
Susko3, line 21:
```suggestion
```

Not needed if the xmldoc reference in `AndroidGameSurface` is removed.

Susko3 (changes requested):
Seems to work well, just need to remove the `osuTK.Android` reference as it's no longer needed.

Susko3:
> just need to remove the `osuTK.Android` reference as it's no longer needed.

Or you can merge right now, and we do this later in osuTK cleanup.

Susko3 (approved):

Joehuu (approved):

bdach (approved):

PR #6108, opened 2024-01-07
Remove per-call array allocations from `InputSampler`

peppy:
Benchmark is part of a larger operation and doesn't show the sheer scale of this as it's running in a test at lower input sampling rates. This gets very out of hand with high poll rate devices.

| Before | After |
| :---: | :---: |
| 2024-01-07 13 10 29@2x | 2024-01-07 13 26 33@2x |

Alternatives would be `ref List<>` as a parameter or just returning a single `Vector2?` always, but my proposal here offers future flexibility and should be safe based on the only true usage of this method being to immediately handle the returned value.

### DIFF
--- a/osu.Framework/Input/InputResampler.cs
+++ b/osu.Framework/Input/InputResampler.cs
@@ -27,21 +27,30 @@ namespace osu.Framework.Input
         /// </summary>
         public bool ResampleRawInput { get; set; }
 
+        private readonly List<Vector2> returnedPositions = new List<Vector2>();
+
         /// <summary>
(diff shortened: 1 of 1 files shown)


bdach:
> just returning a single `Vector2?` always

This would have been my choice, this current API screams "swing and a miss at future-proofing that never happened and probably never will at this stage". But oh well.

In fact the reason that this is basically unused except for cursor trail is the reason why I'll go with this for now, because otherwise I'd say that this is pretty dangerous design.

bdach (approved):

PR #6109, opened 2024-01-08
Reduce allocs when setting value on `BindableNumber` with precision

smoogipoo:
BenchmarkDotNet isn't outputting a summary table, so I'll leave it up to @peppy to provide before/after tales. But the optimisation is pretty standard (see other cases of the same being done in this file).

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkBindableNumber.cs
+++ b/osu.Framework.Benchmarks/BenchmarkBindableNumber.cs
@@ -1,0 +1,29 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using BenchmarkDotNet.Attributes;
+using osu.Framework.Bindables;
+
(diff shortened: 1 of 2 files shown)


bdach:
Before:

|                Method |     Mean |    Error |   StdDev |   Gen0 | Allocated |
|---------------------- |---------:|---------:|---------:|-------:|----------:|
|   SetValueNoPrecision | 13.08 us | 0.322 us | 0.949 us |      - |         - |
| SetValueWithPrecision | 47.66 us | 0.987 us | 2.911 us | 3.7842 |   48000 B |

After:

|                Method |     Mean |    Error |   StdDev | Allocated |
|---------------------- |---------:|---------:|---------:|----------:|
|   SetValueNoPrecision | 13.74 us | 0.273 us | 0.741 us |         - |
| SetValueWithPrecision | 25.52 us | 0.508 us | 1.285 us |         - |

This is gonna be a nitpick but it'd be super helpful for the future if the benchmark was in a separate commit to the fix, so that I don't have to manually revert the fix to check difference in benchmark results.

bdach (approved):

Issue #6110, opened 2024-01-08
`SpriteIcon`s are unnecessary expensive


EVAST9919:
Currently `SpriteIcon` is basically a container with 2 `Sprite`s - one for icon intelf and one for it's shadow, which is an overkill if you have many icons in the scene. This can be *relatively easy* improved by moving shadows into draw node instead.
Note that making `SpriteIcon : Sprite` is a no-go because current implementation implies that they are always centered, fit inside the drawable and texture shouldn't be allowed to be set outside. Thus a new custom drawable should be implemented.

I've made a basic implementation and performance gains are evident. (ignore icons aspect ratios for now, didn't want to spend much time in case this is not-needed/not an huge issue)
Master:
![icons-master](https://github.com/ppy/osu-framework/assets/22874522/47487bfc-4daf-4102-9d61-afccb11b4825)
My implementation:
![Снимок экрана (407)](https://github.com/ppy/osu-framework/assets/22874522/79ed8300-128b-4ef3-820d-4008c869f5f9)

bdach:
Can you show demonstrable performance gains in *real* usages, not in the "spam every font awesome icon that exists" test scene?

I'm not necessarily rejecting this but using the absolute worst case to argue that this gives "significant" perf gains seems like a stretch.

EVAST9919:
Probably bad wording from my side, that's for sure, but I've never said that this will be a game changer or anything.
I understand that since this won't have a huge impact on the game it's automatically low-priority, but if it can be improved - why not?

bdach:
Again, not rejecting this. Just wanting to find out how real the perf gains are to prioritise fixing this properly.

EVAST9919:
This is probably the best it's gonna get.
| master | proposal |
|---|---|
| ![song-select-old](https://github.com/ppy/osu-framework/assets/22874522/a05d2a55-dd37-4d89-bda3-c860b65adb10) | ![song-select-new](https://github.com/ppy/osu-framework/assets/22874522/3c56069c-2c69-49cc-9c9f-466693a5a049) |

bdach:
Performance gain is nice, but why aren't the icons sized the same on both? It's not just proportional upscaling, they're clearly stretched on the second screenshot.

EVAST9919:
> Performance gain is nice, but why aren't the icons sized the same on both? It's not just proportional upscaling, they're clearly stretched on the second screenshot.

As said in the OP
[...]

PR #6111, opened 2024-01-08
Simplify `SpriteIcon`

EVAST9919:
Closes https://github.com/ppy/osu-framework/issues/6110

As far as I can tell, there's no visual difference (which is desired behaviour).
| | master | pr |
|---|---|---|
|tests|test-scene-before|test-scene-after|
|test-rotated|rotation-before|rotation-after|
|settings|settings-before|settings-after|
|song-select|song-select-before|song-select-after|
|mods|mods-before|mods-after|


peppy:
Weirdly, I brought this up recently during a stream. Not sure if you noticed and took action or just a coincidence, but great to see 😅

peppy (approved):

PR #6115, opened 2024-01-09
Fix insane joystick allocations overhead

peppy:


peppy:
Turns out this only affects tests. And I'm not even going to try to fix the rest of the issues (basically `PassThroughInputManager` is broke beyond recovery).

peppy:
Turns out it doesn't only affect tests. `RulesetInputManager` is one of these.

I'm going to have to fix the rest of this mess...

Will do that in separate PRs.

osu.Framework/Input/States/JoystickState.cs
```diff
@@ -22,10 +21,17 @@ public class JoystickState
         /// </summary>
         internal readonly JoystickButton[] AxisDirectionButtons = new JoystickButton[MAX_AXES];
 
+        private readonly JoystickAxis[] axes = new JoystickAxis[MAX_AXES];
```
bdach, line 24: I do wonder if this is worth the hassle. Like if `GetAxes()` is just a thin veneer over this array then you might as well expose it as public anyway. Heck all of the other input state classes do that.

Probably fine either way but I do wonder about your opinion.

peppy: I'm fine with that, but the main reason I didn't is because when I've done so before, issue has been brought with the non-immutability nature of arrays. Hinted on this in the OP (the fact we're casting this to `IEnumerable` also causes enumerator overheads).

Dunno, I lean more towards "we're making a game engine, not a financial framework which needs to guard against users doing silly things at all costs". We've sacrificed performance (and time to fix that performance gap) due to being overly careful over the years and I'm not sure how worth it that is.

bdach: I'll go with this for now as not to slow this down over trivialities but other input state classes already do raw arrays so /shrug

bdach (approved):

PR #6116, opened 2024-01-09
Remove alloc overhead for `NormalizedValue`

peppy:
Was ~200KB/s during gameplay.

Not even going to provide a benchmark, should be obvious.

### DIFF
--- a/osu.Framework/Bindables/RangeConstrainedBindable.cs
+++ b/osu.Framework/Bindables/RangeConstrainedBindable.cs
@@ -74,6 +74,47 @@ namespace osu.Framework.Bindables
             setValue(defaultValue);
         }
 
+        public float NormalizedValue
+        {
+            get
+            {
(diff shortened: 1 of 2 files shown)


bdach:
Just for my peace of mind...

before

|                      Method |     Mean |    Error |   StdDev |   Gen0 | Allocated |
|---------------------------- |---------:|---------:|---------:|-------:|----------:|
| NormalizedValueDoesNotAlloc | 27.65 us | 0.472 us | 0.580 us | 5.7068 |  70.24 KB |

after

|                      Method |     Mean |     Error |    StdDev | Allocated |
|---------------------------- |---------:|----------:|----------:|----------:|
| NormalizedValueDoesNotAlloc | 5.264 us | 0.0489 us | 0.0409 us |         - |

Not committing the benchmark though, was very ad-hoc and I don't care enough to clean it up.

bdach (approved):

PR #6121, opened 2024-01-10
Reduce string allocations during shader creation

peppy:
| Before | After |
| :---: | :---: |
| 2024-01-11 05 19 21@2x | 2024-01-11 05 19 17@2x |

### DIFF
--- a/osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
+++ b/osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
@@ -123,9 +123,9 @@ namespace osu.Framework.Graphics.Veldrid.Shaders
                     return output;
 
                 Inputs.AddRange(
-                    shader_input_pattern.Matches(output).Select(m => new VeldridShaderAttribute(int.Parse(m.Groups[1].Value, CultureInfo.InvariantCulture), m.Groups[2].Value)).ToList());
+                    shader_input_pattern.Matches(output).Select(m => new VeldridShaderAttribute(int.Parse(m.Groups[1].Value, CultureInfo.InvariantCulture), m.Groups[2].Value)));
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Veldrid/Shaders/VeldridShaderPart.cs
```diff
@@ -105,35 +111,33 @@ private string loadFile(byte[]? bytes, bool mainFile)
...
-                    Inputs.AddRange(
-                        shader_input_pattern.Matches(result).Select(m => new VeldridShaderAttribute(int.Parse(m.Groups[1].Value, CultureInfo.InvariantCulture), m.Groups[2].Value)).ToList());
-                    Outputs.AddRange(
-                        shader_output_pattern.Matches(result).Select(m => new VeldridShaderAttribute(int.Parse(m.Groups[1].Value, CultureInfo.InvariantCulture), m.Groups[2].Value)).ToList());
+                if (!mainFile)
+                    return output;
 
-                    string outputCode = loadFile(store.GetRawData($"Internal/sh_{Type}_Output.h"), false);
+                Inputs.AddRange(
+                    shader_input_pattern.Matches(output).Select(m => new VeldridShaderAttribute(int.Parse(m.Groups[1].Value, CultureInfo.InvariantCulture), m.Groups[2].Value)).ToList());
```
turbedi, line 126: Unrelated to the changes in this PR, but: Is the ToList() call at the end actually needed?

peppy: They aren't required, no. I'll remove them.

Note that there's still a lot of remaining optimisations that can be done to this shader compilation class. I just addressed the easy/quick one.

bdach (approved):

PR #6122, opened 2024-01-10
Remove delegate allocation in jitter calculation in `FramedClock`

peppy:
| Before | After |
| :---: | :---: |
| 2024-01-11 05 33 07@2x | 2024-01-11 05 40 58@2x |

### DIFF
--- a/osu.Framework/Timing/FramedClock.cs
+++ b/osu.Framework/Timing/FramedClock.cs
@@ -1,10 +1,9 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using osu.Framework.Extensions.TypeExtensions;
 using System;
 using System.Diagnostics;
-using System.Linq;
+using osu.Framework.Extensions.TypeExtensions;
 
 namespace osu.Framework.Timing
 {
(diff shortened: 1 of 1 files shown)


osu.Framework/Timing/FramedClock.cs
```diff
@@ -85,9 +84,18 @@ public virtual void ProcessFrame()
...
+                    double sumOfSquares = 0;
+
+                    foreach (double v in betweenFrameTimes)
+                    {
+                        sum += v;
+                        sumOfSquares += v * v;
+                    }
+
+                    double avg = sum / betweenFrameTimes.Length;
+                    double variance = (sumOfSquares / betweenFrameTimes.Length) - (avg * avg);
```
bdach, line 97: I am _shook_ that this can be done and be actually correct.

Proof:

$$
\begin{align}
\sigma_x &= \frac{\sum_i (x_i - \bar{x})^2}{|x|} = \\
&= \frac{\sum_i (x_i^2 - 2x_i \bar{x} + \bar{x}^2)}{|x|} = \\
&= \frac{\sum_i x_i^2}{|x|} - 2 \bar{x} \frac{\sum_i x_i}{|x|} + \bar{x}^2 \cdot \frac{\sum_i 1}{|x|} = \\
&= \frac{\sum_i x_i^2}{|x|} - 2 \bar{x} \cdot \bar{x} + \bar{x}^2 \cdot \frac{|x|}{|x|} = \\
&= \frac{\sum_i x_i^2}{|x|} - \bar{x}^2
\end{align}
$$

wherein $x_i$ are elements of a set, $\bar{x}$ is the mean of the set, and $|x|$ is the cardinality of the set.

I thought there was no way this was gonna work but kept failing to find counterexamples...

bdach (approved):

PR #6124, opened 2024-01-11
Remove allocations during bass device sync polling operation

peppy:
Minor gains because this is just a one second poll, but is what it is.

| Before | After |
| :---: | :---: |
| CleanShot 2024-01-11 at 05 48 29 | CleanShot 2024-01-11 at 05 47 48 |

Also fixed a couple of incorrect calls on array building. Check underlying implementation for reasoning (one allocates, other doesn't).

### DIFF
--- a/osu.Framework/Input/Bindings/KeyCombination.cs
+++ b/osu.Framework/Input/Bindings/KeyCombination.cs
@@ -644,7 +644,9 @@ namespace osu.Framework.Input.Bindings
 
             Debug.Assert(!keys.Contains(InputKey.None)); // Having None in pressed keys will break IsPressed
             keys.Sort();
-            return new KeyCombination(keys.ToImmutable());
+
+            // Can't use `MoveToImmutable` here as we don't provide accurate capacity.
+            return new KeyCombination(keys.ToImmutableList());
         }
     }
 
--- a/osu.Framework/Platform/SDL2Window_Windowing.cs
+++ b/osu.Framework/Platform/SDL2Window_Windowing.cs
@@ -341,7 +341,7 @@ namespace osu.Framework.Platform
                     Logger.Log($"Failed to retrieve SDL display at index ({i})", level: LogLevel.Error);
             }
 
-            return builder.ToImmutable();
+            return builder.MoveToImmutable();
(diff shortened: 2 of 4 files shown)


osu.Framework/Input/Bindings/KeyCombination.cs
```diff
@@ -644,7 +644,9 @@ public static KeyCombination FromInputState(InputState state, Vector2? scrollDel
 
             Debug.Assert(!keys.Contains(InputKey.None)); // Having None in pressed keys will break IsPressed
             keys.Sort();
-            return new KeyCombination(keys.ToImmutable());
+
+            // Can't use `MoveToImmutable` here as we don't provide accurate capacity.
+            return new KeyCombination(keys.ToImmutableList());
```
bdach, line 649: Not sure what this change is doing in here...

osu.Framework/Platform/SDL2Window_Windowing.cs
```diff
@@ -341,7 +341,7 @@ private static IEnumerable<Display> getSDLDisplays()
                     Logger.Log($"Failed to retrieve SDL display at index ({i})", level: LogLevel.Error);
             }
 
-            return builder.ToImmutable();
+            return builder.MoveToImmutable();
```
bdach, line 344: ...or this one either, for that matter...

peppy: See above.

osu.Framework/Input/Bindings/KeyCombination.cs
bdach: In this particular case it looks to me that the difference should be zero and both will allocate, but /shrug. (The `.MoveToImmutable()` stuff seems valid.)

peppy: Oh yeah, I meant to revert this case. As per the comment, we can't use `MoveTo` due to capacity not being set (a hidden gotcha).

This will still allocate as the call I incorrectly "reverted" to is synonymous with the original.

bdach (approved):

PR #6125, opened 2024-01-11
Remove overheads of MIDI device polling

peppy:
Just one, though. Low priority. At this point I'm just trying to get a clean slate.

CleanShot 2024-01-11 at 06 17 36

### DIFF
--- a/osu.Framework/Input/Handlers/Midi/MidiHandler.cs
+++ b/osu.Framework/Input/Handlers/Midi/MidiHandler.cs
@@ -86,34 +86,40 @@ namespace osu.Framework.Input.Handlers.Midi
         {
             try
             {
-                var inputs = MidiAccessManager.Default.Inputs.ToList();
-
                 lock (openedDevices)
                 {
-                    // check removed devices
-                    foreach (string key in openedDevices.Keys.ToArray())
+                    int accountedDevices = 0;
+
+                    foreach (var device in MidiAccessManager.Default.Inputs)
                     {
-                        var device = openedDevices[key];
+                        accountedDevices++;
 
-                        if (inputs.All(i => i.Id != key))
-                        {
-                            closeDevice(device);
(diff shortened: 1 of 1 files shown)


osu.Framework/Input/Handlers/Midi/MidiHandler.cs
```diff
@@ -86,34 +86,40 @@ private bool refreshDevices()
...
+                        var newInput = MidiAccessManager.Default.OpenInputAsync(device.Id).GetResultSafely();
+                        newInput.MessageReceived += onMidiMessageReceived;
+                        openedDevices[device.Id] = newInput;
+
+                        Log($"Connected MIDI device: {newInput.Details.Name}");
                     }
 
-                    // check added devices
-                    foreach (IMidiPortDetails input in inputs)
+                    if (accountedDevices != openedDevices.Count)
```
bdach, line 107: I don't know if this is valid. What if between polls a device is removed *and* another is added? That's not an outrageous proposition, polls happen every second.

Like yes, if the count *does* meaningfully change from one second to the next, then yes this works, but if it doesn't but the *actual* devices do...

bdach: I don't have enough midi devices to test experimentally, so I guess I'll just cross my fingers and hope that's how it ends up working.

bdach (approved):

osu.Framework/Input/Handlers/Midi/MidiHandler.cs
peppy: If it makes you any more confident, the midi device count doesn't update for me ever, even when disconnecting/reconnecting a device and nuking from my system. only when i restart the app.

i tracked it back to the native coreaudio call and all seemed correct, so not sure what gives.

PR #6126, opened 2024-01-11
Reduce allocation overhead for keyboard / binding / button handling

peppy:
Overall gains are pretty substantial. All testing done with 13 keystrokes. Ignore overall allocations as the runtime of tests was significantly different.

| Before | After |
| :---: | :---: |
| CleanShot 2024-01-11 at 07 44 09 | CleanShot 2024-01-11 at 07 44 17 |

---

94b4b2435aa692d79274fae680b31ceb1dfc8ecb Remove allocation overhead in button propagation flow

| Before | After |
| :---: | :---: |
| CleanShot 2024-01-11 at 07 40 36 | CleanShot 2024-01-11 at 07 41 14 |

d16d899a67415dff193f5eb11a272cd99265916a Remove allocation overhead from `KeyBindingContainer` key handling

| Before | After |
| :---: | :---: |
| CleanShot 2024-01-11 at 07 35 00 | CleanShot 2024-01-11 at 07 35 54 |

7641f7021ef606f0e68ac28f5078b869648f8edd Reduce allocation overhead of `KeyCombination` constructor

| Before | After |
| :---: | :---: |
| CleanShot 2024-01-11 at 07 30 23 | CleanShot 2024-01-11 at 07 30 28 |

### DIFF
--- a/osu.Framework/Input/Bindings/KeyBindingContainer.cs
+++ b/osu.Framework/Input/Bindings/KeyBindingContainer.cs
@@ -202,6 +202,8 @@ namespace osu.Framework.Input.Bindings
             return drawables.FirstOrDefault(d => triggerKeyBindingEvent(d, pressEvent)) != null;
         }
 
+        private readonly List<IKeyBinding> newlyPressed = new List<IKeyBinding>();
+
         private bool handleNewPressed(InputState state, InputKey newKey, Vector2? scrollDelta = null, bool isPrecise = false)
         {
             pressedInputKeys.Add(newKey);
@@ -210,10 +212,20 @@ namespace osu.Framework.Input.Bindings
             var pressedCombination = new KeyCombination(pressedInputKeys);
 
             bool handled = false;
-            var bindings = KeyBindings?.Except(pressedBindings) ?? Enumerable.Empty<IKeyBinding>();
 
-            var newlyPressed = bindings.Where(m =>
-                m.KeyCombination.IsPressed(pressedCombination, matchingMode));
+            newlyPressed.Clear();
+
+            if (KeyBindings != null)
+            {
+                foreach (IKeyBinding binding in KeyBindings)
+                {
+                    if (pressedBindings.Contains(binding))
+                        continue;
+
+                    if (binding.KeyCombination.IsPressed(pressedCombination, matchingMode))
+                        newlyPressed.Add(binding);
+                }
+            }
 
             if (KeyCombination.IsModifierKey(newKey))
             {
@@ -221,11 +233,15 @@ namespace osu.Framework.Input.Bindings
                 // lambda expression is used so that the delegate is cached (see: https://github.com/dotnet/roslyn/issues/5835)
                 // TODO: remove when we switch to .NET 7.
                 // ReSharper disable once ConvertClosureToMethodGroup
-                newlyPressed = newlyPressed.Where(b => b.KeyCombination.Keys.All(key => KeyCombination.IsModifierKey(key)));
+                for (int i = 0; i < newlyPressed.Count; i++)
+                {
+                    if (!newlyPressed[i].KeyCombination.Keys.All(key => KeyCombination.IsModifierKey(key)))
+                        newlyPressed.RemoveAt(i--);
+                }
             }
 
             // we want to always handle bindings with more keys before bindings with less.
-            newlyPressed = newlyPressed.OrderByDescending(b => b.KeyCombination.Keys.Length).ToList();
+            newlyPressed.Sort(static (a, b) => b.KeyCombination.Keys.Length.CompareTo(a.KeyCombination.Keys.Length));
 
             pressedBindings.AddRange(newlyPressed);
 
@@ -343,14 +359,13 @@ namespace osu.Framework.Input.Bindings
             // we don't want to consider exact matching here as we are dealing with bindings, not actions.
             var pressedCombination = new KeyCombination(pressedInputKeys);
 
-            var newlyReleased = pressedInputKeys.Count == 0
-                ? pressedBindings.ToList()
-                : pressedBindings.Where(b => !b.KeyCombination.IsPressed(pressedCombination, KeyCombinationMatchingMode.Any)).ToList();
-
-            foreach (var binding in newlyReleased)
+            for (int i = 0; i < pressedBindings.Count; i++)
             {
-                pressedBindings.Remove(binding);
+                var binding = pressedBindings[i];
 
+                if (binding.KeyCombination.IsPressed(pressedCombination, KeyCombinationMatchingMode.Any)) continue;
+
+                pressedBindings.RemoveAt(i--);
                 PropagateReleased(getInputQueue(binding).Where(d => d.IsRootedAt(this)), state, binding.GetAction<T>());
                 keyBindingQueues[binding].Clear();
             }
--- a/osu.Framework/Input/Bindings/KeyBindingContainer.cs
+++ b/osu.Framework/Input/Bindings/KeyBindingContainer.cs
@@ -211,6 +211,7 @@ namespace osu.Framework.Input.Bindings
 
             bool handled = false;
             var bindings = KeyBindings?.Except(pressedBindings) ?? Enumerable.Empty<IKeyBinding>();
+
             var newlyPressed = bindings.Where(m =>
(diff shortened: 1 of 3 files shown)


osu.Framework/Input/Bindings/KeyCombination.cs
```diff
@@ -29,10 +29,20 @@ namespace osu.Framework.Input.Bindings
...
-            Keys = keys?.Any() == true ? keys.Distinct().OrderBy(k => (int)k).ToImmutableArray() : none;
+            if (keys == null || !keys.Any())
+            {
+                Keys = ImmutableArray<InputKey>.Empty;
+                return;
+            }
+
+            var keyBuilder = ImmutableArray.CreateBuilder<InputKey>(keys.Count());
+
+            keyBuilder.AddRange(keys);
```
bdach, line 42: As per inspection `keys` can be enumerated up to three times here. Probably shouldn't be. Maybe ctor argument should be `ICollection`.

osu.Framework/Input/Bindings/KeyBindingContainer.cs
```diff
@@ -210,21 +212,36 @@ private bool handleNewPressed(InputState state, InputKey newKey, Vector2? scroll
...
+                for (int i = 0; i < newlyPressed.Count; i++)
+                {
+                    if (!newlyPressed[i].KeyCombination.Keys.All(key => KeyCombination.IsModifierKey(key)))
+                        newlyPressed.RemoveAt(i--);
+                }
             }
 
             // we want to always handle bindings with more keys before bindings with less.
-            newlyPressed = newlyPressed.OrderByDescending(b => b.KeyCombination.Keys.Length).ToList();
+            newlyPressed.Sort(static (a, b) => b.KeyCombination.Keys.Length.CompareTo(a.KeyCombination.Keys.Length));
```
bdach, line 244: This is technically a change in behaviour as `.Sort()` uses an unstable algorithm while `.OrderByDescending()` uses a stable one. I don't realistically think it matters though.

osu.Framework/Input/Bindings/KeyCombination.cs
peppy: That's what I was looking for.

osu.Framework/Input/Bindings/KeyCombination.cs
```diff
@@ -29,10 +29,20 @@ namespace osu.Framework.Input.Bindings
...
         /// </summary>
         /// <param name="keys">The keys.</param>
-        /// <remarks>This constructor is not optimized. Hot paths are assumed to use <see cref="FromInputState(InputState, Vector2?)"/>.</remarks>
-        public KeyCombination(IEnumerable<InputKey>? keys)
+        public KeyCombination(ICollection<InputKey>? keys)
         {
-            Keys = keys?.Any() == true ? keys.Distinct().OrderBy(k => (int)k).ToImmutableArray() : none;
+            if (keys == null || !keys.Any())
+            {
+                Keys = ImmutableArray<InputKey>.Empty;
```
Susko3, line 36: The previous code would use `KeyCombiantion.none` instead of an empty array.

Susko3 (changes requested):

osu.Framework/Input/Bindings/KeyCombination.cs
bdach: to spell it out because the above kinda leaves it implied, they're unfortunately not the same thing (empty collection vs collection containing `InputKey.None`)

osu.Framework/Input/Bindings/KeyCombination.cs
```diff
@@ -29,10 +29,20 @@ namespace osu.Framework.Input.Bindings
         /// Construct a new instance.
         /// </summary>
         /// <param name="keys">The keys.</param>
-        /// <remarks>This constructor is not optimized. Hot paths are assumed to use <see cref="FromInputState(InputState, Vector2?)"/>.</remarks>
-        public KeyCombination(IEnumerable<InputKey>? keys)
+        public KeyCombination(ICollection<InputKey>? keys)
         {
-            Keys = keys?.Any() == true ? keys.Distinct().OrderBy(k => (int)k).ToImmutableArray() : none;
```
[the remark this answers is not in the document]
peppy: what's the case where it isn't already distinct?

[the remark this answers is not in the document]
Susko3: With #5790, different physical keys might generate the same virtual keycode (eg. number row 0 and numpad 0). But I see that the `FromInputState()` code already does it's own de-duplication for keyboard keys specifically:
 https://github.com/ppy/osu-framework/blob/master/osu.Framework%2FInput%2FBindings%2FKeyCombination.cs#L625-L626

So it's probably fine, but the optimized ctor shouldn't be `public`.

osu.Framework/Input/Bindings/KeyCombination.cs
peppy: Have fixed.

bdach (approved):

osu.Framework/Input/Bindings/KeyBindingContainer.cs
```diff
@@ -342,14 +359,13 @@ private void handleNewReleased(InputState state, InputKey releasedKey)
...
-                ? pressedBindings.ToList()
-                : pressedBindings.Where(b => !b.KeyCombination.IsPressed(pressedCombination, KeyCombinationMatchingMode.Any)).ToList();
-
-            foreach (var binding in newlyReleased)
+            for (int i = 0; i < pressedBindings.Count; i++)
             {
-                pressedBindings.Remove(binding);
+                var binding = pressedBindings[i];
+
+                if (binding.KeyCombination.IsPressed(pressedCombination, KeyCombinationMatchingMode.Any)) continue;
```
bdach, line 366: There's a behavioural change here that breaks tests...

The previous code had an explicit `pressedInputKeys.Count == 0` check. That would basically "short circuit" the `IsPressed()` checking by just assuming that in such a case all pressed combinations must be released. But now this transformation breaks things because due to the lack of that special check, `IsPressed` is given `KeyCombination.none` as the `pressedCombination` here and dies.

Something like this fixes:

```diff
diff --git a/osu.Framework/Input/Bindings/KeyBindingContainer.cs b/osu.Framework/Input/Bindings/KeyBindingContainer.cs
index 6fcb2ba06..4e7908212 100644
--- a/osu.Framework/Input/Bindings/KeyBindingContainer.cs
+++ b/osu.Framework/Input/Bindings/KeyBindingContainer.cs
@@ -363,11 +363,12 @@ private void handleNewReleased(InputState state, InputKey releasedKey)
             {
                 var binding = pressedBindings[i];
 
-                if (binding.KeyCombination.IsPressed(pressedCombination, KeyCombinationMatchingMode.Any)) continue;
-
-                pressedBindings.RemoveAt(i--);
-                PropagateReleased(getInputQueue(binding).Where(d => d.IsRootedAt(this)), state, binding.GetAction<T>());
-                keyBindingQueues[binding].Clear();
+                if (pressedInputKeys.Count == 0 || !binding.KeyCombination.IsPressed(pressedCombination, KeyCombinationMatchingMode.Any))
+                {
+                    pressedBindings.RemoveAt(i--);
+                    PropagateReleased(getInputQueue(binding).Where(d => d.IsRootedAt(this)), state, binding.GetAction<T>());
+                    keyBindingQueues[binding].Clear();
+                }
             }
         }
 

```

but also feels immensely dodgy... Not sure what *else* to do, though.

bdach (changes requested):

osu.Framework/Input/Bindings/KeyBindingContainer.cs
peppy: I don't get it.

My thinking is that if `pressedBindings` is empty, there's nothing to release. What am I missing?

bdach: The previous code was filtering `pressedInputBindings` _basing on the value of_ `pressedKeys`. Significant difference.

Which then, with the current code, dies at

https://github.com/ppy/osu-framework/blob/master/osu.Framework%2FInput%2FBindings%2FKeyCombination.cs#L75

peppy: Right. Got confused with variables to some degree. I think your solution is okay.

bdach (approved):

Issue #6127, opened 2024-01-11
Unable to open more than one osu!framework based project at the same time


anispwyn:
when you're opening something like osu!lazer, [fluXis](https://github.com/TeamFluXis/fluXis) would refuse to open and vise versa (it report STATUS_SUCCESS after closed)

for some reason, you can open [Vigette](https://github.com/vignetteapp/vignette) and osu!lazer (or fluXis) just fine, probably because it(Vigette) is using the old version of the framework

you can watch an example video here:
https://streamable.com/yxxxtb

tested using:
osu!lazer [2023.1231.0](https://github.com/ppy/osu/releases/tag/2023.1231.0)
fluXis [Hotfix v2024.1.1](https://github.com/TeamFluXis/fluXis/releases/tag/2024.1.1)

environment:
windows 11 enterprise insider preview build 26020
r5 6600h
rtx 3050ti

bdach:
> probably because it(Vigette) is using the old version of the framework

No. Root cause will be in here: https://github.com/TeamFluXis/fluXis/blob/5890fbe6487e3323c379a83454f935d5a0f4f346/fluXis.Desktop/Program.cs#L25-L45

fluxis instance thinks the previously opened osu! instance is the primary instance of itself rather than a completely different framework app.

The ipc port probably needs to be configurable per framework app:

https://github.com/ppy/osu-framework/blob/master/osu.Framework/Platform/DesktopGameHost.cs#L18

PR #6129, opened 2024-01-12
Make IPC port configurable via `HostOptions`

tsunyoku:
Resolves https://github.com/ppy/osu-framework/issues/6127

# Breaking changes

## `HostOptions.BindIPC` has been superseded by `HostOptions.IPCPort`

Previously, the IPC port used for multiple instances of a single osu!framework app was hardcoded in the IPC host itself, effectively making it so that _all osu!framework apps_ would share the same IPC port, which obviously cannot work.

To allow multiple osu!framework apps to utilise IPC concurrently, `BindIPC` has thus been replaced by `IPCPort`.

- Setting `IPCPort = null` is equivalent to `BindIPC = false`.
- Setting `IPCPort` to a non-null value is equivalent to `BindIPC = true` *and* will force the use the specific port provided.

Note that it is advised to use a "user port" (in the range of 1024-49151) as per RFC 6335.

### DIFF
--- a/osu.Framework/HostOptions.cs
+++ b/osu.Framework/HostOptions.cs
@@ -11,7 +11,9 @@ namespace osu.Framework
     public class HostOptions
     {
         /// <summary>
-        /// The IPC port to bind. See <see cref="IIpcHost"/> for more details on usage.
+        /// The IPC port to bind. This port should be between 1024 and 49151,
+        /// and should be unique to avoid conflicts with other osu!framework apps.
+        /// See <see cref="IIpcHost"/> for more details on usage.
         /// </summary>
         public int? IPCPort { get; set; }
 
--- a/osu.Framework/HostOptions.cs
+++ b/osu.Framework/HostOptions.cs
@@ -12,7 +12,8 @@ namespace osu.Framework
     {
         /// <summary>
         /// The IPC port to bind. This port should be between 1024 and 49151,
-        /// and should be unique to avoid conflicts with other osu!framework apps.
+        /// should be shared by all instances of a given osu!framework app,
+        /// but be distinct from IPC ports specified by other osu!framework apps.
         /// See <see cref="IIpcHost"/> for more details on usage.
         /// </summary>
         public int? IPCPort { get; set; }
--- a/osu.Framework/HostOptions.cs
+++ b/osu.Framework/HostOptions.cs
@@ -11,14 +11,9 @@ namespace osu.Framework
     public class HostOptions
     {
         /// <summary>
-        /// Whether to bind the IPC port. See <see cref="IIpcHost"/> for more details on usage.
-        /// </summary>
-        public bool BindIPC { get; set; }
-
-        /// <summary>
         /// The IPC port to bind. See <see cref="IIpcHost"/> for more details on usage.
         /// </summary>
-        public int IPCPort { get; set; } = 45356;
+        public int? IPCPort { get; set; }
 
         /// <summary>
         /// Whether this is a portable installation. Will cause all game files to be placed alongside the executable, rather than in the standard data directory.
--- a/osu.Framework/HostOptions.cs
+++ b/osu.Framework/HostOptions.cs
@@ -16,6 +16,11 @@ namespace osu.Framework
         public bool BindIPC { get; set; }
 
         /// <summary>
+        /// The IPC port to bind. See <see cref="IIpcHost"/> for more details on usage.
+        /// </summary>
+        public int IPCPort { get; set; } = 45356;
+
+        /// <summary>
         /// Whether this is a portable installation. Will cause all game files to be placed alongside the executable, rather than in the standard data directory.
         /// </summary>
         public bool PortableInstallation { get; set; }
--- a/osu.Framework.Tests/Platform/HeadlessGameHostTest.cs
+++ b/osu.Framework.Tests/Platform/HeadlessGameHostTest.cs
@@ -89,8 +89,8 @@ namespace osu.Framework.Tests.Platform
         [Test]
         public void TestIpc()
         {
-            using (var server = new BackgroundGameHeadlessGameHost(@"server", new HostOptions { BindIPC = true }))
-            using (var client = new HeadlessGameHost(@"client", new HostOptions { BindIPC = true }))
+            using (var server = new BackgroundGameHeadlessGameHost(@"server", new HostOptions { IPCPort = 45356 }))
+            using (var client = new HeadlessGameHost(@"client", new HostOptions { IPCPort = 45356 }))
             {
                 Assert.IsTrue(server.IsPrimaryInstance, @"Server wasn't able to bind");
                 Assert.IsFalse(client.IsPrimaryInstance, @"Client was able to bind when it shouldn't have been able to");
--- a/osu.Framework.Tests/Platform/UserInputManagerTest.cs
+++ b/osu.Framework.Tests/Platform/UserInputManagerTest.cs
@@ -13,7 +13,7 @@ namespace osu.Framework.Tests.Platform
         [Test]
         public void IsAliveTest()
         {
-            using (var client = new TestHeadlessGameHost(@"client", true))
+            using (var client = new TestHeadlessGameHost(@"client", 45356))
             {
                 var testGame = new TestTestGame();
                 client.Run(testGame);
@@ -25,8 +25,8 @@ namespace osu.Framework.Tests.Platform
         {
             public Drawable CurrentRoot => Root;
 
-            public TestHeadlessGameHost(string gameName, bool bindIPC)
-                : base(gameName, new HostOptions { BindIPC = bindIPC })
+            public TestHeadlessGameHost(string gameName, int? ipcPort)
+                : base(gameName, new HostOptions { IPCPort = ipcPort })
             {
             }
         }
--- a/osu.Framework/Platform/DesktopGameHost.cs
+++ b/osu.Framework/Platform/DesktopGameHost.cs
@@ -15,15 +15,15 @@ namespace osu.Framework.Platform
 {
     public abstract class DesktopGameHost : SDL2GameHost
     {
-        public const int IPC_PORT = 45356;
-
         private TcpIpcProvider ipcProvider;
         private readonly bool bindIPCPort;
+        private readonly int ipcPort;
 
         protected DesktopGameHost(string gameName, HostOptions options = null)
             : base(gameName, options)
         {
             bindIPCPort = Options.BindIPC;
+            ipcPort = Options.IPCPort;
             IsPortableInstallation = Options.PortableInstallation;
         }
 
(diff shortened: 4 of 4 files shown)


osu.Framework/HostOptions.cs
```diff
@@ -15,6 +15,11 @@ public class HostOptions
         /// </summary>
         public bool BindIPC { get; set; }
 
+        /// <summary>
+        /// The IPC port to bind. See <see cref="IIpcHost"/> for more details on usage.
+        /// </summary>
+        public int IPCPort { get; set; } = 45356;
```
bdach, line 21: I do wonder from a usability standpoint as to whether this should be a separate thing. It will be very easy for consumers to set `BindIPC = true` but still continue to be blissfully unaware of the fact that they're binding to a port that's potentially shared with other framework apps.

If you ask me this should be `int? IPCPort { get; set; }` and it should supersede `BindIPC`, i.e. if you set the port you _will_ automatically bind to it. If you don't, you won't. This immediately forces the consumer to consider what happens if they set a particular port.

@ppy/team-client opinions?

tsunyoku: Yeah, that sounds right. Unless you're someone actually who is aware of the issue existing and then read through o!f to find this you're probably not going to notice. I'll change it if others agree on the desired direction.

frenzibyte: Sounds good to me.

tsunyoku: I've applied the change.

osu.Framework/HostOptions.cs
```diff
@@ -11,9 +11,9 @@ namespace osu.Framework
     public class HostOptions
     {
         /// <summary>
-        /// Whether to bind the IPC port. See <see cref="IIpcHost"/> for more details on usage.
+        /// The IPC port to bind. See <see cref="IIpcHost"/> for more details on usage.
```
peppy, line 14: Probably worth mentioning a valid range of ports which should be considered. And mention that this should be unique to the application to avoid conflicts with other osu!framework apps.

tsunyoku: I've added some extra information detailing the allowed port range and unique expectation. For more context on the port range specified, ports between 0 and 49151 are the only assignable ports with the first 1024 being "system" ports on some systems i.e reserved and therefore require admin to be bound to (and sometimes would be entirely denied). This leaves us with the remaining assignable range: 1024-49151. More about this can be read here: http://www.faqs.org/rfcs/rfc6335.html

bdach:
Baseline changes to apply game-side after this:

```diff
diff --git a/Templates/Rulesets/ruleset-empty/osu.Game.Rulesets.EmptyFreeform.Tests/VisualTestRunner.cs b/Templates/Rulesets/ruleset-empty/osu.Game.Rulesets.EmptyFreeform.Tests/VisualTestRunner.cs
index 03ee7c9204..8b7a3e6045 100644
--- a/Templates/Rulesets/ruleset-empty/osu.Game.Rulesets.EmptyFreeform.Tests/VisualTestRunner.cs
+++ b/Templates/Rulesets/ruleset-empty/osu.Game.Rulesets.EmptyFreeform.Tests/VisualTestRunner.cs
@@ -13,7 +13,7 @@ public static class VisualTestRunner
         [STAThread]
         public static int Main(string[] args)
         {
-            using (DesktopGameHost host = Host.GetSuitableDesktopHost(@"osu", new HostOptions { BindIPC = true }))
+            using (DesktopGameHost host = Host.GetSuitableDesktopHost(@"osu", new HostOptions { IPCPort = 45356 }))
             {
                 host.Run(new OsuTestBrowser());
                 return 0;
diff --git a/Templates/Rulesets/ruleset-example/osu.Game.Rulesets.Pippidon.Tests/VisualTestRunner.cs b/Templates/Rulesets/ruleset-example/osu.Game.Rulesets.Pippidon.Tests/VisualTestRunner.cs
index 55c0cf6a3b..8a270bf5f0 100644
--- a/Templates/Rulesets/ruleset-example/osu.Game.Rulesets.Pippidon.Tests/VisualTestRunner.cs
+++ b/Templates/Rulesets/ruleset-example/osu.Game.Rulesets.Pippidon.Tests/VisualTestRunner.cs
@@ -13,7 +13,7 @@ public static class VisualTestRunner
         [STAThread]
         public static int Main(string[] args)
         {
-            using (DesktopGameHost host = Host.GetSuitableDesktopHost(@"osu", new HostOptions { BindIPC = true }))
+            using (DesktopGameHost host = Host.GetSuitableDesktopHost(@"osu", new HostOptions { IPCPort = 45356 }))
             {
                 host.Run(new OsuTestBrowser());
                 return 0;
diff --git a/Templates/Rulesets/ruleset-scrolling-empty/osu.Game.Rulesets.EmptyScrolling.Tests/VisualTestRunner.cs b/Templates/Rulesets/ruleset-scrolling-empty/osu.Game.Rulesets.EmptyScrolling.Tests/VisualTestRunner.cs
index b45505678c..f2fafee414 100644
--- a/Templates/Rulesets/ruleset-scrolling-empty/osu.Game.Rulesets.EmptyScrolling.Tests/VisualTestRunner.cs
+++ b/Templates/Rulesets/ruleset-scrolling-empty/osu.Game.Rulesets.EmptyScrolling.Tests/VisualTestRunner.cs
@@ -13,7 +13,7 @@ public static class VisualTestRunner
         [STAThread]
         public static int Main(string[] args)
         {
-            using (DesktopGameHost host = Host.GetSuitableDesktopHost(@"osu", new HostOptions { BindIPC = true }))
+            using (DesktopGameHost host = Host.GetSuitableDesktopHost(@"osu", new HostOptions { IPCPort = 45356 }))
             {
                 host.Run(new OsuTestBrowser());
                 return 0;
diff --git a/Templates/Rulesets/ruleset-scrolling-example/osu.Game.Rulesets.Pippidon.Tests/VisualTestRunner.cs b/Templates/Rulesets/ruleset-scrolling-example/osu.Game.Rulesets.Pippidon.Tests/VisualTestRunner.cs
index 55c0cf6a3b..8a270bf5f0 100644
--- a/Templates/Rulesets/ruleset-scrolling-example/osu.Game.Rulesets.Pippidon.Tests/VisualTestRunner.cs
+++ b/Templates/Rulesets/ruleset-scrolling-example/osu.Game.Rulesets.Pippidon.Tests/VisualTestRunner.cs
@@ -13,7 +13,7 @@ public static class VisualTestRunner
         [STAThread]
         public static int Main(string[] args)
         {
-            using (DesktopGameHost host = Host.GetSuitableDesktopHost(@"osu", new HostOptions { BindIPC = true }))
+            using (DesktopGameHost host = Host.GetSuitableDesktopHost(@"osu", new HostOptions { IPCPort = 45356 }))
             {
                 host.Run(new OsuTestBrowser());
                 return 0;
diff --git a/osu.Desktop/Program.cs b/osu.Desktop/Program.cs
index 6b95a82703..ef3aff6bdf 100644
--- a/osu.Desktop/Program.cs
+++ b/osu.Desktop/Program.cs
@@ -102,7 +102,7 @@ public static void Main(string[] args)
                 }
             }
 
-            using (DesktopGameHost host = Host.GetSuitableDesktopHost(gameName, new HostOptions { BindIPC = !tournamentClient }))
+            using (DesktopGameHost host = Host.GetSuitableDesktopHost(gameName, new HostOptions { IPCPort = tournamentClient ? null : 45356 }))
             {
                 if (!host.IsPrimaryInstance)
                 {
diff --git a/osu.Game.Tests/Visual/Navigation/TestSceneInterProcessCommunication.cs b/osu.Game.Tests/Visual/Navigation/TestSceneInterProcessCommunication.cs
index 1ecd38e1d3..85531b66a2 100644
--- a/osu.Game.Tests/Visual/Navigation/TestSceneInterProcessCommunication.cs
+++ b/osu.Game.Tests/Visual/Navigation/TestSceneInterProcessCommunication.cs
@@ -63,7 +63,7 @@ public override void SetUpSteps()
             });
             AddStep("create IPC sender channels", () =>
             {
-                ipcSenderHost = new HeadlessGameHost(gameHost.Name, new HostOptions { BindIPC = true });
+                ipcSenderHost = new HeadlessGameHost(gameHost.Name, new HostOptions { IPCPort = 45356 });
                 osuSchemeLinkIPCSender = new OsuSchemeLinkIPCChannel(ipcSenderHost);
                 archiveImportIPCSender = new ArchiveImportIPCChannel(ipcSenderHost);
             });
diff --git a/osu.Game.Tournament.Tests/TournamentTestRunner.cs b/osu.Game.Tournament.Tests/TournamentTestRunner.cs
index 5f642b14f5..1d38648243 100644
--- a/osu.Game.Tournament.Tests/TournamentTestRunner.cs
+++ b/osu.Game.Tournament.Tests/TournamentTestRunner.cs
@@ -12,7 +12,7 @@ public static class TournamentTestRunner
         [STAThread]
         public static int Main(string[] args)
         {
-            using (DesktopGameHost host = Host.GetSuitableDesktopHost(@"osu-development", new HostOptions { BindIPC = true }))
+            using (DesktopGameHost host = Host.GetSuitableDesktopHost(@"osu-development", new HostOptions { IPCPort = 45356 }))
             {
                 host.Run(new TournamentTestBrowser());
                 return 0;
diff --git a/osu.Game/Tests/CleanRunHeadlessGameHost.cs b/osu.Game/Tests/CleanRunHeadlessGameHost.cs
index f3c69201e2..6a5c756a9e 100644
--- a/osu.Game/Tests/CleanRunHeadlessGameHost.cs
+++ b/osu.Game/Tests/CleanRunHeadlessGameHost.cs
@@ -27,7 +27,7 @@ public class CleanRunHeadlessGameHost : TestRunHeadlessGameHost
                                         [CallerMemberName] string callingMethodName = @"")
             : base($"{callingMethodName}-{Guid.NewGuid()}", new HostOptions
             {
-                BindIPC = bindIPC,
+                IPCPort = 45356,
             }, bypassCleanup: bypassCleanupOnDispose, realtime: realtime)
         {
             this.bypassCleanupOnSetup = bypassCleanupOnSetup;
diff --git a/osu.Game/Tests/VisualTestRunner.cs b/osu.Game/Tests/VisualTestRunner.cs
index e04c71d193..c6a0eaf9bd 100644
--- a/osu.Game/Tests/VisualTestRunner.cs
+++ b/osu.Game/Tests/VisualTestRunner.cs
@@ -12,7 +12,7 @@ public static class VisualTestRunner
         [STAThread]
         public static int Main(string[] args)
         {
-            using (DesktopGameHost host = Host.GetSuitableDesktopHost(@"osu-development", new HostOptions { BindIPC = true, }))
+            using (DesktopGameHost host = Host.GetSuitableDesktopHost(@"osu-development", new HostOptions { IPCPort = 45356, }))
             {
                 host.Run(new OsuTestBrowser());
                 return 0;

```

May want to change the visual test runner port to be able to run that independently of game, not sure. The above is the minimal viable.

bdach (approved):

PR #6130, opened 2024-01-12
Fix `KeyCombination` throwing when duplicates are fed in

peppy:
This class may require some test coverage. Just pushing out for immediate visibility.


bdach:
Added test coverage, fixed bug in implementation (https://github.com/ppy/osu-framework/pull/6130/commits/e2f325f04e571ac07396d2f8237185d5f29188c3)

bdach (approved):

peppy:
> fixed bug in implementation

🤕

Issue #6131, opened 2024-01-13
Frame statistics graph causes the visuals to become unresponsive if window isn't focused for a long time


Sinamer:
The visuals will recover eventually if you either wait long enough, or turn off the graph (although next time it gets opened it will cause the hang right away).
The input stays responsive, only the visuals seem to be affected

The steps to repro:
1. Open frame statistics graph (tap Ctrl+F11 twice)
2. Switch away from the window
3. Wait for a couple minutes (three minutes was enough to cause a complete hang on my side)
4. Observe the framework tests process increase memory consumption indefinitely
5. Switch back to the window

This is what it looks like:
![image](https://github.com/ppy/osu-framework/assets/53346335/b3e6e3a7-8f2f-4fdf-9542-5451bf2c5e54)
![image](https://github.com/ppy/osu-framework/assets/53346335/d72a8452-80a2-42a6-8096-c91559975679)

PR #6133, opened 2024-01-14
Fix runaway texture uploads when window is minimised with frame statistics visible

peppy:
Closes https://github.com/ppy/osu-framework/issues/6131.

Tried to think of a way to avoid this at a higher level but not really sure we can.


osu.Framework/Graphics/OpenGL/Textures/GLTexture.cs
```diff
@@ -188,6 +189,9 @@ public void SetData(ITextureUpload upload)
         {
             lock (uploadQueue)
             {
+                if (uploadQueue.Count >= 100 && uploadQueue.Count % 100 == 0)
+                    Logger.Log($"Texture {Identifier}'s upload queue is large ({uploadQueue.Count})");
```
peppy, line 193: Note that "Identifier" probably doesn't work ever. But I'm not even going to attempt to fix this right now (we have like 10 texture classes all with their own `Name`, `AssetName`, `Identifier` etc.).

bdach (approved):

PR #6135, opened 2024-01-15
Fix DragStart and TouchMove input events possibly being handled by unrooted drawables

OliBomby:
Drawables in the `ButtonDownInputQueue` could become unrooted between the button down event and a drag start event or touch move event. 
If a parent of a drawable gets removed, the children will keep `IsAlive` true forever, so checking for `IsAlive` is not sufficient.

Failing tests included. The test for the key repeat input does not fail before my fixes, but I added it for good measure.

### DIFF
--- a/osu.Framework.Tests/Input/MouseInputTest.cs
+++ b/osu.Framework.Tests/Input/MouseInputTest.cs
@@ -108,6 +108,7 @@ namespace osu.Framework.Tests.Input
         {
             InputReceptor receptor = null;
             Container receptorParent = null;
+            Vector2 lastPosition = Vector2.Zero;
 
             AddStep("create hierarchy", () =>
             {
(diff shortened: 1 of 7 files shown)


osu.Framework.Tests/Input/MouseInputTest.cs
```diff
@@ -99,10 +100,43 @@ public void TestNoLongerValidDrawableDoesNotReceiveDoubleClick()
...
+                    }
+                };
+            });
+
+            AddStep("move mouse to receptor", () => InputManager.MoveMouseTo(receptor));
+            AddStep("press button", () => InputManager.PressButton(MouseButton.Left));
+
+            AddStep("remove receptor parent", () => Remove(receptorParent, true));
+
+            AddStep("drag mouse", () => InputManager.MoveMouseTo(receptor, new Vector2(10)));
```
Susko3, line 131: This is moving the mouse relative to a drawable that is disposed and no longer in the scene graph. Seems like undefined behaviour to me.

Susko3 (changes requested):
Same goes for the other tests.

osu.Framework/Input/KeyEventManager.cs
```diff
@@ -24,7 +24,7 @@ public KeyEventManager(Key key)
         public void HandleRepeat(InputState state)
         {
             // Only drawables that can still handle input should handle the repeat
-            var drawables = ButtonDownInputQueue.AsNonNull().Intersect(InputQueue).Where(t => t.IsAlive && t.IsPresent);
+            var drawables = ButtonDownInputQueue.AsNonNull().Intersect(InputQueue);
```
bdach, line 27: So the `Alive && Present` check was added in https://github.com/ppy/osu-framework/pull/3209, along with a test, but I went and checked and at least as far back as 2b034d42012119f32c7b3f23e648c3e023774e20 (didn't check before that because I can't be bothered installing net3.1) this was not necessary and the test added in that PR was passing with that check reverted...

So as long as this doesn't blow up game-side I'm going to call this _proooobably_ fine?

bdach (approved):
passed game-side tests, so I'm calling it fine

PR #6137, opened 2024-01-16
Add the ability to solo a test in visual testing using `[Solo]` attribute

peppy:
I've been wanting this for years. Discussed implementation details recently with smoogi and agreed that getting something very simple implemented would be beneficial to test development efficiency.

Basically, rather than commenting out 100 tests in a method where you want to test one specific test repeatedly (commonly using hot reload), you can now add `[Solo]` to the test in question and the runner will ignore the rest.

This has no effect when tests are run from nunit, as you may expect.

### DIFF
--- a/osu.Framework/Testing/TestBrowser.cs
+++ b/osu.Framework/Testing/TestBrowser.cs
@@ -400,7 +400,13 @@ namespace osu.Framework.Testing
 
             bool hadTestAttributeTest = false;
 
-            foreach (var m in newTest.GetType().GetMethods())
+            var methods = newTest.GetType().GetMethods();
+
+            var solo = methods.FirstOrDefault(m => m.GetCustomAttribute(typeof(SoloAttribute), false) != null);
+            if (solo != null)
+                methods = new[] { solo };
+
+            foreach (var m in methods)
             {
                 string name = m.Name;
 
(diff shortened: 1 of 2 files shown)


osu.Framework/Testing/TestBrowser.cs
```diff
@@ -400,7 +400,13 @@ private void finishLoad(TestScene newTest, Action onCompletion)
 
             bool hadTestAttributeTest = false;
 
-            foreach (var m in newTest.GetType().GetMethods())
+            var methods = newTest.GetType().GetMethods();
+
+            var solo = methods.FirstOrDefault(m => m.GetCustomAttribute(typeof(SoloAttribute), false) != null);
+            if (solo != null)
+                methods = new[] { solo };
```
Susko3, line 407: Why not allow multiple solo methods? Change the logic to a `.Select()` and overwrite `methods` only if the resulting array is non-empty.

Would be useful for testing two or more specific methods.

peppy: Sure, sounds fine I think.

peppy:
Hope this one can get in as a quick change. I was just trying to use it and realised it's still not merged 🥲 .

smoogipoo (approved):

PR #6138, opened 2024-01-17
Write log header later to ensure game identifiers are present

peppy:
The flow in `GameHost` ensures that the game/version identifiers are set before `Logger.Storage` is set. Unfortunately, until now any call to the static `Logger.Log` would immediately materialise the headers into their final string form. A stray log call before `GameHost.Run` could result in less-than-optimal headers.

This can be seen in osu!'s runtime log from an [early log call](https://github.com/ppy/osu/blob/1e3c332658a320fb2dbf49e4a97e2927cafc4dcc/osu.Desktop/Program.cs#L124).

Before:

```
2024-01-17 15:34:40 [verbose]: ----------------------------------------------------------
2024-01-17 15:34:40 [verbose]: runtime Log for dean (LogLevel: Debug)
2024-01-17 15:34:40 [verbose]: Running game unknown on .NET 6.0.25
2024-01-17 15:34:40 [verbose]: Environment: macOS (Unix 14.3.0), 8 cores
2024-01-17 15:34:40 [verbose]: ----------------------------------------------------------
2024-01-17 15:34:40 [verbose]: Starting legacy IPC provider...
```

After:

```
----------------------------------------------------------
runtime Log for dean (LogLevel: Debug)
Running osu-development 0.0.0.0 on .NET 6.0.25
Environment: macOS (Unix 14.3.0), 8 cores 
----------------------------------------------------------
2024-01-17 15:27:24 [verbose]: Starting legacy IPC provider...
```

Note that timestamps are not prefixed to the header now, but I think this is probably for the best.

### DIFF
--- a/osu.Framework/Logging/Logger.cs
+++ b/osu.Framework/Logging/Logger.cs
@@ -302,8 +302,6 @@ namespace osu.Framework.Logging
             if (!Enabled || level < Level)
                 return;
 
-            ensureHeader();
-
             logCount.Value++;
 
             message = ApplyFilters(message);
@@ -394,6 +392,17 @@ namespace osu.Framework.Logging
                 using (var stream = Storage.GetStream(Filename, FileAccess.Write, FileMode.Append))
                 using (var writer = new StreamWriter(stream))
                 {
+                    if (!headerAdded)
+                    {
+                        writer.WriteLine("----------------------------------------------------------");
+                        writer.WriteLine($"{Name} Log for {UserIdentifier} (LogLevel: {Level})");
+                        writer.WriteLine($"Running {GameIdentifier} {VersionIdentifier} on .NET {Environment.Version}");
+                        writer.WriteLine($"Environment: {RuntimeInfo.OS} ({Environment.OSVersion}), {Environment.ProcessorCount} cores ");
+                        writer.WriteLine("----------------------------------------------------------");
+
+                        headerAdded = true;
+                    }
+
                     foreach (string line in lines)
                         writer.WriteLine(line);
                 }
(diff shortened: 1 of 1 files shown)


bdach:
> Note that timestamps are not prefixed to the header now, but I think this is probably for the best.

Is there any particular reason for that? Not that I necessarily care very much, but it just reads like a non-sequitur compared to the rest of the OP. I don't see why that measure would need to be taken.

peppy:
> I don't see why that measure would need to be taken.

I can add it back, but if you check the code it should be apparently why they aren't there (different code path now).

It would look something like this:

```diff
diff --git a/osu.Framework/Logging/Logger.cs b/osu.Framework/Logging/Logger.cs
index 32731b801..63dddb504 100644
--- a/osu.Framework/Logging/Logger.cs
+++ b/osu.Framework/Logging/Logger.cs
@@ -315,7 +315,7 @@ private void add(string message = @"", LogLevel level = LogLevel.Verbose, Except
             IEnumerable<string> lines = logOutput
                                         .Replace(@"\r\n", @"\n")
                                         .Split('\n')
-                                        .Select(s => $@"{DateTime.UtcNow.ToString("yyyy-MM-dd HH:mm:ss", CultureInfo.InvariantCulture)} [{level.ToString().ToLowerInvariant()}]: {s.Trim()}");
+                                        .Select(l => addTimestamp(l, level));
 
             if (outputToListeners)
             {
@@ -394,11 +394,16 @@ private void writePendingLines()
                 {
                     if (!headerAdded)
                     {
-                        writer.WriteLine("----------------------------------------------------------");
-                        writer.WriteLine($"{Name} Log for {UserIdentifier} (LogLevel: {Level})");
-                        writer.WriteLine($"Running {GameIdentifier} {VersionIdentifier} on .NET {Environment.Version}");
-                        writer.WriteLine($"Environment: {RuntimeInfo.OS} ({Environment.OSVersion}), {Environment.ProcessorCount} cores ");
-                        writer.WriteLine("----------------------------------------------------------");
+                        lines = new[]
+                                {
+                                    "----------------------------------------------------------",
+                                    $"{Name} Log for {UserIdentifier} (LogLevel: {Level})",
+                                    $"Running {GameIdentifier} {VersionIdentifier} on .NET {Environment.Version}",
+                                    $"Environment: {RuntimeInfo.OS} ({Environment.OSVersion}), {Environment.ProcessorCount} cores ",
+                                    "----------------------------------------------------------"
+                                }
+                                .Select(l => addTimestamp(l, Level))
+                                .Concat(lines).ToArray();
 
                         headerAdded = true;
                     }
@@ -412,6 +417,11 @@ private void writePendingLines()
             }
         }
 
+        private string addTimestamp(string s, LogLevel level)
+        {
+            return $@"{DateTime.UtcNow.ToString("yyyy-MM-dd HH:mm:ss", CultureInfo.InvariantCulture)} [{level.ToString().ToLowerInvariant()}]: {s.Trim()}";
+        }
+
         /// <summary>
         /// Whether the output of this logger should be sent to listeners of <see cref="Debug"/> and <see cref="Console"/>.
         /// Defaults to true.

```

bdach (approved):
seems fine

PR #6140, opened 2024-01-18
Add flaky test attribute and mark `CanSeekForwardOnInterpolationFail` as flaky

frenzibyte:
Been failing for quite some time now.

### DIFF
--- a/osu.Framework.Tests/Clocks/InterpolatingFramedClockTest.cs
+++ b/osu.Framework.Tests/Clocks/InterpolatingFramedClockTest.cs
@@ -156,7 +156,13 @@ namespace osu.Framework.Tests.Clocks
             Assert.Greater(interpolatedCount, 10);
         }
 
+        // always failing on single-thread macOS CI, failing when test asserts interpolating.IsInterpolating = false
+        //
+        // CanSeekForwardsOnInterpolationFail
+        //     Expected: False
+        //     But was:  True
         [Test]
+        [FlakyTest]
         public void CanSeekForwardsOnInterpolationFail()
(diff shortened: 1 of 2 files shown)


osu.Framework.Tests/Clocks/InterpolatingFramedClockTest.cs
```diff
@@ -156,7 +156,13 @@ public void NeverInterpolatesBackwardsOnInterpolationFail()
             Assert.Greater(interpolatedCount, 10);
         }
 
+        // always failing on single-thread macOS CI, failing when test asserts interpolating.IsInterpolating = false
```
peppy, line 159: If it always fails, flaky probably won't work? It's only intended for intermittent failures.

frenzibyte: Well sometimes it passes (see current CI status), so I wasn't sure about fully ignoring it. But I could easily do so, and make the ignore macOS-only using `Assert.Ignore`. Will leave it to your call.

peppy: Dunno, I read your comment "always failing" and made the assumption that was the case. Change the comment to "Intermittently failing" i guess

peppy, line 159:
```suggestion
        // Regularly failing on single-thread macOS CI, failing when test asserts interpolating.IsInterpolating = false
```

peppy (approved):

PR #6142, opened 2024-01-18
Add adjustment step setting for `SliderBar`

peppy:
Picture this scenario: you have a bindable which you want to bind to a slider bar control to allow user adjustment. Generally, when adjusting via a slider bar, adjustments should be sane values, but a user might want to override with a textbox input also bound the to the same bindable.

Using precision becomes non-feasible because it will limit the bindable across all usages (foreshadowing https://github.com/ppy/osu/issues/25862). Like with other UI frameworks, we want to set a tick/step value which governs inputs made via the slider bar.

There's an argument for "fix this locally" but I guarantee we will run into this scenario more and more, especially after we make textbox input for slider config settings more common and users ask for higher precision.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/SliderBar.cs
+++ b/osu.Framework/Graphics/UserInterface/SliderBar.cs
@@ -8,6 +8,7 @@ using osu.Framework.Graphics.Containers;
 using osuTK.Input;
 using osuTK;
 using osu.Framework.Input.Events;
+using osu.Framework.Utils;
 
 namespace osu.Framework.Graphics.UserInterface
 {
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/UserInterface/SliderBar.cs
```diff
@@ -27,7 +28,12 @@ public abstract partial class SliderBar<T> : Container, IHasCurrentValue<T>
         /// <summary>
         /// A custom step value for each key press which actuates a change on this control.
         /// </summary>
-        public float KeyboardStep;
+        public double KeyboardStep { get; set; }
```
peppy, line 31: Not sure whether `double` or `float`, indifferent but they should match.

peppy:
This works, but doesn't completely solve issues:

https://github.com/ppy/osu-framework/assets/191335/c99b2631-8ed8-489c-9d9a-33f80c7020fa

Open to any general feedback on direction.

PR #6143, opened 2024-01-18
Include rounded corner radius in android safe area implementation

bdach:
Android apparently has a [separate API for "window cutouts"](https://developer.android.com/develop/ui/views/layout/display-cutout), i.e. for cameras and the like, and a [separate API for "rounded corners"](https://developer.android.com/develop/ui/views/layout/insets/rounded-corners), i.e. devices where the rounded corners of the display will cause some areas of the theoretically-rectangular display to be obscured. Because there is no engineering like overengineering, right? Bonus points for them being usable at different API levels, too.

Disclaimer that this does absolutely nothing on my one test device that has a rounded display and a high enough API level, but honestly it may be a device issue since it's not even returning `true` for `WindowInsets.IsRound`. Maybe it'll help someone, no idea.

### DIFF
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -245,6 +245,22 @@ namespace osu.Framework.Android
                     usableScreenArea = usableScreenArea.Shrink(cutout.SafeInsetLeft, cutout.SafeInsetRight, cutout.SafeInsetTop, cutout.SafeInsetBottom);
             }
 
+            if (OperatingSystem.IsAndroidVersionAtLeast(31) && RootWindowInsets != null)
+            {
+                var topLeftCorner = RootWindowInsets.GetRoundedCorner((int)RoundedCornerPosition.TopLeft);
+                var topRightCorner = RootWindowInsets.GetRoundedCorner((int)RoundedCornerPosition.TopRight);
+                var bottomLeftCorner = RootWindowInsets.GetRoundedCorner((int)RoundedCornerPosition.BottomLeft);
+                var bottomRightCorner = RootWindowInsets.GetRoundedCorner((int)RoundedCornerPosition.BottomRight);
+
+                int cornerInsetLeft = Math.Max(topLeftCorner?.Radius ?? 0, bottomLeftCorner?.Radius ?? 0);
+                int cornerInsetRight = Math.Max(topRightCorner?.Radius ?? 0, bottomRightCorner?.Radius ?? 0);
+                int cornerInsetTop = Math.Max(topLeftCorner?.Radius ?? 0, topRightCorner?.Radius ?? 0);
+                int cornerInsetBottom = Math.Max(bottomLeftCorner?.Radius ?? 0, bottomRightCorner?.Radius ?? 0);
+
+                var radiusInsetArea = screenArea.Shrink(cornerInsetLeft, cornerInsetRight, cornerInsetTop, cornerInsetBottom);
+                usableScreenArea = usableScreenArea.Intersect(radiusInsetArea);
+            }
+
             if (OperatingSystem.IsAndroidVersionAtLeast(24) && Activity.IsInMultiWindowMode)
             {
                 // if we are in multi-window mode, the status bar is always visible (even if we request to hide it) and could be obstructing our view.
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -273,8 +273,6 @@ namespace osu.Framework.Android
                 usableScreenArea = usableScreenArea.Intersect(screenArea.Shrink(0, 0, statusBarHeight, 0));
             }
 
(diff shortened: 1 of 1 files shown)


LittleEndu:
Reading the code, I'm guessing that the resulting safe area is less than optimal.

image

It would make way more sense if the code only cut away from the longest screen side (resulting in what I colored orange), or if optimizing for most play area only (1-(sqrt(2)/2) of the radius should be cut away (resulting in what I colored green). Hopefully someone can test what the resulting screen actually looks like.

Also there's a TODO still left in.

osu.Framework.Android/AndroidGameView.cs
```diff
@@ -245,6 +245,22 @@ private void updateSafeArea()
...
+                var bottomLeftCorner = RootWindowInsets.GetRoundedCorner((int)RoundedCornerPosition.BottomLeft);
+                var bottomRightCorner = RootWindowInsets.GetRoundedCorner((int)RoundedCornerPosition.BottomRight);
+
+                int cornerInsetLeft = Math.Max(topLeftCorner?.Radius ?? 0, bottomLeftCorner?.Radius ?? 0);
+                int cornerInsetRight = Math.Max(topRightCorner?.Radius ?? 0, bottomRightCorner?.Radius ?? 0);
+                int cornerInsetTop = Math.Max(topLeftCorner?.Radius ?? 0, topRightCorner?.Radius ?? 0);
+                int cornerInsetBottom = Math.Max(bottomLeftCorner?.Radius ?? 0, bottomRightCorner?.Radius ?? 0);
+
+                var radiusInsetArea = screenArea.Shrink(cornerInsetLeft, cornerInsetRight, cornerInsetTop, cornerInsetBottom);
+                usableScreenArea = usableScreenArea.Intersect(radiusInsetArea);
```
Susko3, line 268: I think the math here is too conservative. See the illustration (green would be considered the safe area):

image

It should either apply top & bottom or left & right padding, not both at the same time.

Susko3 (changes requested):
I've yet to test this on a real device, but the native android stuff looks correct.

~~You should remove this todo about rounded corners:~~

~~https://github.com/ppy/osu-framework/blob/f67b4bfa910f0d23970a8d6a7d0a9afa262e4706/osu.Framework.Android/AndroidGameView.cs#L276C1-L276C1~~

Susko3:
I'm more for the orange solution as it matched what other games do & doesn't complicate things with 4-side padding. The green solution is too simple and not a true optimal area, as it doesn't consider the width and height of the screen (if the screen is really tall, it would be optimal to push the corner point towards the orange solution).

bdach:
can anyone with an iOS device say what iOS does with rounded corners? @frenzibyte @peppy?

Susko3:
I think this is accurate, except the [bottom safe area edge is overriden locally](https://github.com/ppy/osu/blob/2be989b9f570f46bf95a11a162fcab9a3e0adc63/osu.iOS/OsuGameIOS.cs#L22-L25).


safe_area

frenzibyte:
iPhone 12:

image

bdach:
that looks weird because it cuts out mostly at the long edge *but* also appears to cut out a bit on the short edge?

i dunno. think i'm just gonna make it cut at the long edge and hope everyone here can be satisfied with that.

osu.Framework.Android/AndroidGameView.cs
```diff
@@ -245,6 +245,25 @@ private void updateSafeArea()
...
+                var topRightCorner = RootWindowInsets.GetRoundedCorner((int)RoundedCornerPosition.TopRight);
+                var bottomLeftCorner = RootWindowInsets.GetRoundedCorner((int)RoundedCornerPosition.BottomLeft);
+                var bottomRightCorner = RootWindowInsets.GetRoundedCorner((int)RoundedCornerPosition.BottomRight);
+
+                int cornerInsetLeft = Math.Max(topLeftCorner?.Radius ?? 0, bottomLeftCorner?.Radius ?? 0);
+                int cornerInsetRight = Math.Max(topRightCorner?.Radius ?? 0, bottomRightCorner?.Radius ?? 0);
+                int cornerInsetTop = Math.Max(topLeftCorner?.Radius ?? 0, topRightCorner?.Radius ?? 0);
+                int cornerInsetBottom = Math.Max(bottomLeftCorner?.Radius ?? 0, bottomRightCorner?.Radius ?? 0);
+
+                var radiusInsetArea = screenArea.Width >= screenArea.Height
```
bdach, line 260: i hope i can be forgiven for probably imperfectly handling square screens

peppy:
> that looks weird because it cuts out mostly at the long edge *but* also appears to cut out a bit on the short edge?
> 
that's just the bezel

osu.Framework.Android/AndroidGameView.cs
bdach: it's not a setting, and i have absolutely no idea, can we skip that question?

Susko3 (approved):
Tested to work as expected when hard-coding corner radii.

Joehuu (approved):
LGTM on Pixel 6a.

peppy (approved):
Looks sane

PR #6151, opened 2024-01-26
Implement Linux clipboard with support for images

Speykious:
This is a followup to my recent comment on this discussion about being able to copy a screenshot in osu!: https://github.com/ppy/osu/discussions/16775#discussioncomment-8248801

This is an implementation of a Linux clipboard by spawning an `xclip` process. It should work on all X11 systems.
I tested in the clipboard visual test scene and all tests pass, so there shouldn't be a problem!

However, I haven't implemented anything for Wayland yet. At least as it stands, if it fails to find `xclip`, it'll fallback on the SDL2 clipboard via inheritance. That means that images won't work on Wayland yet, but I'll look into a way I can potentially implement that.

### DIFF
--- a/osu.Framework/Platform/Linux/LinuxClipboard.cs
+++ b/osu.Framework/Platform/Linux/LinuxClipboard.cs
@@ -1,0 +1,132 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using NuGet.Packaging;
+using SixLabors.ImageSharp;
+using SixLabors.ImageSharp.Formats.Png;
+
+namespace osu.Framework.Platform.Linux
+{
+    public class LinuxClipboard : Clipboard
+    {
+        protected Process Xclip(IEnumerable<string> args)
+        {
+            Process xclip = new Process();
+            xclip.StartInfo.FileName = "xclip";
+            xclip.StartInfo.UseShellExecute = false;
+            xclip.StartInfo.RedirectStandardInput = true;
+            xclip.StartInfo.RedirectStandardError = true;
+            xclip.StartInfo.RedirectStandardOutput = true;
+            xclip.StartInfo.ArgumentList.AddRange(args);
+            return xclip;
+        }
+
+        protected bool HasTarget(string target)
+        {
+            using Process xclip = Xclip(new[] { "-selection", "clipboard", "-t", "TARGETS", "-o" });
+            xclip.Start();
+
+            bool hasTarget = false;
+
+            try
+            {
+                string? line;
+                while ((line = xclip.StandardOutput.ReadLine()) != null)
+                {
+                    if (line == target)
+                    {
+                        hasTarget = true;
+                        break;
+                    }
+                }
+            }
+            catch (Exception)
+            {
+            }
+
+            xclip.StandardInput.Close();
+            xclip.WaitForExit();
+
+            return hasTarget;
+        }
+
+        public override Image<TPixel>? GetImage<TPixel>()
+        {
+            if (!HasTarget("image/png"))
+                return null;
+
+            using Process xclip = Xclip(new[] { "-selection", "clipboard", "-t", "image/png", "-o" });
+            xclip.Start();
+
+            // reserve 5 MB because it's a bit above a typical png size
+            using MemoryStream xclipStdout = new MemoryStream(5 * 1024 * 1024);
+
+            // If we don't read stdout before waiting, the process might hang because some internal buffer got full.
+            // I know, weird...
+            xclip.StandardOutput.BaseStream.CopyTo(xclipStdout);
+            xclipStdout.Position = 0;
+
+            xclip.StandardInput.Close();
+            xclip.WaitForExit();
+
+            if (xclip.ExitCode == 0)
+                return Image.Load<TPixel>(xclipStdout);
+            else
+                return null;
+        }
+
+        public override string? GetText()
+        {
+            if (!HasTarget("text/plain"))
+                return null;
+
+            using Process xclip = Xclip(new[] { "-selection", "clipboard", "-t", "text/plain", "-o" });
+            xclip.Start();
+
+            string? stdout = null;
+            try
+            {
+                stdout = xclip.StandardOutput.ReadToEnd();
+            }
+            catch (Exception)
+            {
+            }
+
+            xclip.StandardInput.Close();
+            xclip.WaitForExit();
+
+            if (xclip.ExitCode == 0)
+                return stdout;
+            else
+                return null;
+        }
+
+        public override bool SetImage(Image image)
+        {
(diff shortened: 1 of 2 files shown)


bdach:
I don't agree with the approach taken here of calling out into `xclip`.

There is no guarantee that `xclip` is: present on the user's system, *is actually an* `xclip` rather than some other random executable, is of a correct enough version to accept the flags that it's being given, will work on a user's system (consider a dual-Wayland-and-X distro that has utilities for both, but the user is currently in a Wayland session), etc., etc. It also breaks the concept of the appimages we ship being self sufficient.

I would look upon something like this much more favourably if it either (a) used direct X calls, or better yet, (b), called out to an actual native *library* wrapping Linux clipboard * that we can control, version, and ship.

> However, I haven't implemented anything for Wayland yet

I wouldn't be surprised if there wasn't anything on Wayland that allows implementing this, or if there is, it's walled behind a protocol extension that only N% of compositors have agreed to implement. Surely this will be the year of the Linux desktop now.

Speykious:
I understand the compromise. Technically `xclip` is not guaranteed to be the actual `xclip` my code is looking for here.

To be completely fair, without considering the appimage factor, if any Linux user using X11 doesn't have `xclip` or has an `xclip` program that isn't the standard one every other program expects, it's pretty much guaranteed to be a user issue. If anything, this should at least tone down the lack of guarantee of `xclip` being the correct program, since it's just so ubiquitous in the Linux ecosystem.

That issue shouldn't be crashing the program though, and that's why I made sure to make my clipboard implementation fallback to the SDL2 clipboard in case of failing. That fallback also works in the case of someone having `xclip` while being on Wayland (since`xclip` will just exit with status 1) like me right now since I was just about to test some Wayland stuff for this PR.

I get why it defeats the point of the appimage though. I'll look around to see if there isn't already a native library (or even a C# one) that can provide this.

As for whether or not it's the correct version, I think it's not a problem. All the flags I am using are present today and were still present in [this commit of `xclip` from 2010](https://github.com/astrand/xclip/tree/23c15101a77a584b1ae6c996de8aac254135ae16):

<details>

<summary>Code block showing that every command still works with 2010 xclip</summary>

```sh
# getting targets: works
xclip (23c1510) [?] via C v13.2.1-gcc 
❯ xclip -selection clipboard -t TARGETS -o
TIMESTAMP
TARGETS
application/x-qt-image
image/png
image/avif
image/bmp
# ...

# pasting an image to stdout: works
xclip (23c1510) [?] via C v13.2.1-gcc
❯ xclip -selection clipboard -t image/png -o > image.png

# pasting text to stdout: works
xclip (23c1510) [?] via C v13.2.1-gcc
❯ xclip -selection clipboard -t text/plain -o

# copying an image from stdin: works
xclip (23c1510) [?] via C v13.2.1-gcc
❯ xclip -selection clipboard -t image/png -i < image.png

# copying text from stdin: works
xclip (23c1510) [?] via C v13.2.1-gcc
❯ echo "hello" | xclip -selection clipboard -t text/plain -i
```

</details>

Now for Wayland, one (evidently non-ideal) solution is actually quite similar to what I'm doing right now: using [`wl-clipboard`](https://github.com/bugaevc/wl-clipboard). Another is, maybe turn `wl-clipboard` into a dynamic library that can be shipped with the appimage. Ultimately though as I said I'll have to look around to evaluate all of this.

Speykious:
I should add: given that I made the clipboard fallback on SDL2, if I don't find any more viable solution right now, I think it wouldn't be a problem to leave that approach as it stands. It's in its own self-contained class, so it can be modified in the future very easily when a better solution comes up, and at the very least something like >95% of the Linux userbase would be able to copy images, which is better than 0%.

Speykious:
> For Wayland, I found [this](https://wayland.freedesktop.org/docs/html/ch04.html#sect-Protocol-data-sharing), looks like it's implemented in [KDE](https://invent.kde.org/plasma/kwin/-/blob/master/src/wayland/datasource.cpp), [GNOME](https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/wayland/meta-wayland-data-device.c) and [wlroots](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/data_device/wlr_data_source.c), which covers probably all Wayland users.

Ah, yeah pretty much. That's also what I found besides other programs.

As for X11, I've been investigating yesterday into what `xclip` has to do to interact with the clipboard, and it's wild. It has to spawn an invisible window and at one point fork itself to keep a reference to some buffer alive for you to be able to paste it. I guess we can thank X11 and its technical debt from 2 decades ago when the use case wasn't having a single user on a single machine.

Ideally I could make a native implementation myself using [this document on Selections](https://tronche.com/gui/x/icccm/sec-2.html) (which I already started, to experiment with how things work), but I'm worried about the forking bit since obviously one must not fork the parent program just to interact with the clipboard. If at some point I realize it's impossible to achieve without forking, then I'd be inclined to say that `xclip` or a program that does the same thing is at this point the only viable solution, and if we must not defeat the point of an appimage, osu!framework might as well contain and ship its own clipboard executable. I hope this won't be the case... :v

Speykious:
Update: I did a little bit of a test. Using [Xephyr](https://wiki.archlinux.org/title/Xephyr), I ran a bare no-config awesomewm to see how a minimal window manager with no additional tools would behave when it comes to clipboards. Specifically, I wanted to know if SDL's clipboard copying persisted after the application closes. As it turns out, it doesn't!
```sh
$ # open osu-lazer

$ # copy some text from within the game, for example in the editor

$ xclip -sel clip -o
just some text

$ # close osu-lazer

$ xclip -sel clip -o
Error: target STRING not available
```
At this point, I'm continuing the custom native clipboard library path, and it occurs to me that trying to bother with making the selection persist after the app closes is useless. If we're not making a standalone program who's sole purpose is copying (like `xclip`), we shouldn't try to fork anything ourselves.

Installing a decent clipboard manager on Linux can alleviate this problem. Decent here is a very low bar, since it just needs to keep the last copy alive for it to be viable, but usually it'll also have a history. Of course desktop environments don't have this problem in the first place, since they already have a clipboard manager built in. I'd say I can safely continue on this path without bothering with forking.

bdach:
Let me reiterate something from my last reply to this:

> or better yet, (b), called out to an actual native _library_ wrapping Linux clipboard that we can control, version, and ship.

If X clipboard is such an inscrutable mess that you're making it out to be in the replies above then using an external library is the _only_ viable path forward I can see here. I do not wish to have to maintain a manual implementation of this if it's anything over like 10 lines of obvious p/invokes.

Speykious:
Not gonna lie, I myself wouldn't want something like this to be something specific to osu!, it just wouldn't make sense. On top of the maintenance burden, it would be impossible to use anywhere else, which is stupid.
So yeah I guess I'm making a native lib now :v

bdach:
surely there is one out there that can be used already...

Speykious:
The ones I've looked at so far didn't support different mime types unfortunately, and that would defeat the point of this PR. To some extent I'm surprised SDL2 didn't have that in the first place.

Speykious:
Closed in favor of https://github.com/ppy/osu-framework/pull/6241

PR #6152, opened 2024-01-27
Fix Waveform not disposing Stream

hwsmm:
I found this issue a bit back while testing performance of Waveform implementation of #6002.
Having `data.Dispose()` in my code hindered the benchmarks from running, and it seemed that the benchmarks used a single stream through its entire session.

I assumed that the stream gets disposed somewhere else, but I couldn't find any code for that. This could be my oversight, but writing this PR didn't take much time anyway...


bdach (approved):
seems okay

PR #6153, opened 2024-01-28
FFmpeg Linux binaries: Include version in file name

FreezyLemon:
I explain the rationale behind this change in #6154.

The included patch is necessary for local builds on some distros, and it will become necessary on more distros (and CI) as binutils gets updated (ver 2.41 breaks the ffmpeg build).

Note that the new NativeLibs version will cause the current logic to not find the new libraries. See the linked PR for the required logic change.

### DIFF
--- a/osu.Framework.NativeLibs/scripts/ffmpeg/fix-binutils-2.41.patch
+++ b/osu.Framework.NativeLibs/scripts/ffmpeg/fix-binutils-2.41.patch
@@ -1,0 +1,76 @@
+From effadce6c756247ea8bae32dc13bb3e6f464f0eb Mon Sep 17 00:00:00 2001
+From: =?utf8?q?R=C3=A9mi=20Denis-Courmont?= <remi@remlab.net>
+Date: Sun, 16 Jul 2023 18:18:02 +0300
+Subject: [PATCH] avcodec/x86/mathops: clip constants used with shift
+ instructions within inline assembly
+
+Fixes assembling with binutil as >= 2.41
+
+Signed-off-by: James Almer <jamrial@gmail.com>
+---
+ libavcodec/x86/mathops.h | 26 +++++++++++++++++++++++---
+ 1 file changed, 23 insertions(+), 3 deletions(-)
+
+diff --git a/libavcodec/x86/mathops.h b/libavcodec/x86/mathops.h
+index 6298f5ed19..ca7e2dffc1 100644
+--- a/libavcodec/x86/mathops.h
++++ b/libavcodec/x86/mathops.h
+@@ -35,12 +35,20 @@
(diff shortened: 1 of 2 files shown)


osu.Framework.NativeLibs/scripts/ffmpeg/fix-binutils-2.41.patch
bdach, line 1: Ugh this stinks...

- Does ffmpeg not do backports at all?
- Is this patch going to work for `binutils` <2.41? The commit message doesn't say, but we build the images on github actions on ubuntu 20.04, and as per https://packages.ubuntu.com/search?keywords=binutils, it'll be 2.34 on there.

FreezyLemon: I know. I don't exactly *like* adding the responsibility of maintaining patchfiles to the project. But while the FFmpeg project does release patches to older versions, I couldn't find a 4.3 or even 4.X-compatible release with this fix.

And yes, this will work on older versions. The `if (__builtin_constant_p(shift))` is used to check whether the old code can be used or not. To make sure, [here](https://github.com/FreezyLemon/osu-framework/actions/runs/7756281886)'s a build from the PR branch. ([this](https://github.com/FreezyLemon/osu-framework/actions/runs/7756281886/job/21153274126#step:4:103) is the line confirming the patch)

FreezyLemon: Another possible solution would be updating FFmpeg to 6.1. That would require some working around an unfixed bug in the framework code, but it would also avoid this kind of patching.

FreezyLemon:
I've been playing around with using newer FFmpeg versions (>=5.0) in osu-framework and I noticed that FFmpeg.AutoGen had a major API change at some point:

VideoDecoder has some [custom logic](https://github.com/ppy/osu-framework/blob/e7ffb598853dc152c1424917aacdcc6f65eb11fd/osu.Framework/Graphics/Video/VideoDecoder.cs#L807) for loading the FFmpeg libraries which is not possible in newer versions of FFmpeg.AutoGen. Now there seems to only be a `RootPath` property that acts as a search directory for FFmpeg.AutoGen to try and find the libav* libraries.

Why I mention this: The code in FFmpeg.AutoGen that searches for these libav* libraries [expects versioned Linux libraries](https://github.com/Ruslan-B/FFmpeg.AutoGen/blob/28e5e18cd1041b5243e1d88e35225ee5e0a64d7a/FFmpeg.AutoGen/Native/LinuxFunctionResolver.cs#L12). I can't say for sure whether this new `RootPath` approach is enough for what the framework does atm, but this change could be another reason to use versioned file names.

smoogipoo (approved):

Issue #6025, opened 2023-10-12
Add hardware decoding support on Linux platforms (VA-API, VDPAU)


FreezyLemon:
The libraries necessary to enable hwaccel on Linux (libva, libvdpau) are required by FFmpeg, but probably shouldn't be bundled with the game.

One good idea by @bdach was to bundle FFmpeg libraries without hardware acceleration, and to modify the library loading logic so that FFmpeg libraries installed on the system are preferred over the bundled ones. 

That way, if a user installs FFmpeg with hwaccel support for their specific machine, it'll load those libraries (= hardware decoding) and falls back to the bundled libraries (= software decoding) when no installed FFmpeg can be found.

PR #6154, opened 2024-01-28
FFmpeg: Only load versioned libraries on Linux

FreezyLemon:
Prerequisite:
- [x]  #6153 (TODO: Update NativeLib in this PR after the the other PR is merged)

macOS and Windows already use versioned files, e.g. `libavutil.56.dylib`, `avutil-56.dll`. Linux uses unversioned binaries at the moment, presumably because the `NativeLibrary.Load` methods don't handle the Linux versioning conventions (`libavutil.so.56`) very well.

This change makes it so only compatible FFmpeg libraries are loaded[^1], which is the first step towards implementing #6025.

[^1]: As determined by the file name, which has some obvious caveats. Properly checking the library version seems like an overkill.


osu.Framework/Graphics/Video/VideoDecoder.cs
```diff
@@ -808,26 +814,17 @@ protected virtual FFmpegFuncs CreateFuncs()
             {
                 int version = FFmpeg.AutoGen.ffmpeg.LibraryVersionMap[name];
 
-                string libraryName;
-
                 // "lib" prefix and extensions are resolved by .net core
-                switch (RuntimeInfo.OS)
+                string libraryName = RuntimeInfo.OS switch
```
smoogipoo, line 818: Unfortunately, we don't use switch expressions in this project yet :(

smoogipoo (changes requested):

osu.Framework/Graphics/Video/VideoDecoder.cs
```diff
@@ -821,8 +827,11 @@ protected virtual FFmpegFuncs CreateFuncs()
                         libraryName = $"{name}-{version}";
                         break;
 
+                    // To handle versioning in Linux, we have to specify the entire file name
+                    // because Linux uses a version suffix after the file extension (e.g. libavutil.so.56)
+                    // More info: https://learn.microsoft.com/en-us/dotnet/standard/native-interop/native-library-loading?view=net-6.0
                     case RuntimeInfo.Platform.Linux:
-                        libraryName = name;
+                        libraryName = $"lib{name}.so.{version}";
```
sr229, line 834: You should be able to handle cases where there's a more updated of the library present, otherwise this will just break since system packages tend to update quite fast depending on the distro.

sr229 (commented):
Some small nits, I think you may need to make a new CI pipeline to test the switching.

bdach:
> "you may need to make a new CI pipeline to test the switching"

what switching?

osu.Framework/Graphics/Video/VideoDecoder.cs
bdach: what? we bundle these binaries in the appimage. they _must_ exist. we have not used the ambient ffmpeg install for a while now.

sr229:
> > "you may need to make a new CI pipeline to test the switching"
> 
> @sr229 what switching?

Apologies, it seems this code is also meaning to set the stone for #6025 so I may have misread the intent of the PR as asking for a precise version especially in the context of the aforementioned issue this would break on a minor version upgrade. Please disregard and dismiss.

osu.Framework/Graphics/Video/VideoDecoder.cs
FreezyLemon: Sorry, but was this a response to the review posted by sr229? I think something got deleted, but am not sure. I can answer this if it is directed at the PR, but I'm not sure

bdach: Yes my comment was directed at @sr229. Review said something about "should not be hardcoding version because user might have different version installed on system".

smoogipoo (approved):
Lgtm! Thanks!

Issue #6157, opened 2024-01-29
`CancellationToken` should probably always be present in BDL calls


peppy:
Currently it's nullable, which makes usage a bit annoying. Should always be cached as `default` if not present.

PR #6159, opened 2024-01-29
Rework dependencies to allow value types

smoogipoo:
Resolves https://github.com/ppy/osu-framework/issues/6157

Consider this an RFC.

The way structs are handled right now is a mess. If I (as the person who wrote all of this) can't understand how they're handled, then I don't think anyone can. And my confusion is the result of two conflicting thought processes:
- `CancellationToken` is somehow a special type. How is it so?
- The way I view structs is as always being initialised, whether in a "zeroed" state or otherwise (unless they're nullable).

Every value type is a special type that can only be cached by osu!framework. `CancellationToken` is doubly special because it _sometimes_ arrives in `load()`, and _sometimes_ doesn't, depending on whether you're in an async-load context. So the **only proper way** to use BDL with `CancellationToken` is to make it nullable.

In my opinion, if you want to cache and resolve a value-type dependency and accept whatever footguns that comes with (by-val copies, you can't cache multiple `int`s, etc), then you should be able to do so. But the framework should retain predictable semantics, which for me means matching the C# spec - `ValueType = default!` is a zero-initialised struct. This is a simplification of semantics to me.

New semantics:
```
Get<int>() => 0;               // Allowed
Get<int?>() => null;           // Allowed
Cache<int>(0);                 // Allowed
Cache<int>(null);              // Disallowed
Cache<int?>(0);                // Warning in NRT contexts
CacheAs<IInterface>(MyStruct); // Allowed - boxed.
Cache<object>(null);           // Exception
Cache<object?>(null);          // Exception + warning in NRT contexts
```

Is this something I foresee being used? No. It's just for the sake of clarifying semantics.

### DIFF
--- a/osu.Framework.Tests/Dependencies/Reflection/CachedAttributeTest.cs
+++ b/osu.Framework.Tests/Dependencies/Reflection/CachedAttributeTest.cs
@@ -51,11 +51,13 @@ namespace osu.Framework.Tests.Dependencies.Reflection
         }
 
         [Test]
-        public void TestAttemptToCacheStruct()
+        public void TestCacheStruct()
         {
             var provider = new Provider4();
 
-            Assert.Throws<ArgumentException>(() => DependencyActivator.MergeDependencies(provider, new DependencyContainer()));
+            var dependencies = DependencyActivator.MergeDependencies(provider, new DependencyContainer());
+
+            Assert.IsNotNull(dependencies.Get<int?>());
         }
 
         [Test]
@@ -136,7 +138,9 @@ namespace osu.Framework.Tests.Dependencies.Reflection
         {
             var provider = new Provider12();
 
-            Assert.Throws<ArgumentException>(() => DependencyActivator.MergeDependencies(provider, new DependencyContainer()));
+            var dependencies = DependencyActivator.MergeDependencies(provider, new DependencyContainer());
+
+            Assert.IsNotNull(dependencies.Get<IProvidedInterface1>());
         }
 
         /// <summary>
@@ -149,13 +153,13 @@ namespace osu.Framework.Tests.Dependencies.Reflection
 
             var dependencies = DependencyActivator.MergeDependencies(provider, new DependencyContainer());
 
-            Assert.AreEqual(provider.CachedObject.Value, dependencies.GetValue<CachedStructProvider.Struct>().Value);
+            Assert.AreEqual(provider.CachedObject.Value, dependencies.Get<CachedStructProvider.Struct>().Value);
         }
 
         [Test]
-        public void TestGetValueNullInternal()
+        public void TestGetNullInternal()
         {
-            Assert.AreEqual(default(int), new DependencyContainer().GetValue<int>());
+            Assert.AreEqual(default(int), new DependencyContainer().Get<int>());
         }
 
         /// <summary>
@@ -171,7 +175,7 @@ namespace osu.Framework.Tests.Dependencies.Reflection
 
             var dependencies = DependencyActivator.MergeDependencies(provider, new DependencyContainer());
 
-            Assert.AreEqual(testValue, dependencies.GetValue<int?>());
+            Assert.AreEqual(testValue, dependencies.Get<int?>());
         }
 
         [Test]
@@ -460,9 +464,7 @@ namespace osu.Framework.Tests.Dependencies.Reflection
             public object Provided1
             {
                 get => null;
-                set
-                {
-                }
+                set { }
             }
         }
 
@@ -471,9 +473,7 @@ namespace osu.Framework.Tests.Dependencies.Reflection
             [Cached]
             public object Provided1
             {
-                set
-                {
-                }
+                set { }
             }
         }
 
--- a/osu.Framework.Tests/Dependencies/Reflection/ResolvedAttributeTest.cs
+++ b/osu.Framework.Tests/Dependencies/Reflection/ResolvedAttributeTest.cs
@@ -4,6 +4,7 @@
 #nullable disable
 
 using System.Diagnostics.CodeAnalysis;
+using System.Threading;
 using NUnit.Framework;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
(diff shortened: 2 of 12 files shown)


osu.Framework/Allocation/DependencyContainer.cs
```diff
@@ -17,142 +14,85 @@ public class DependencyContainer : IReadOnlyDependencyContainer
...
-        /// Caches an instance of a type as its most derived type. This instance will be returned each time you <see cref="Get(Type)"/>.
+        /// Caches an instance of a type as its most derived type. This instance will be returned each time you <see cref="Get{T}()"/>.
         /// </summary>
         /// <param name="instance">The instance to cache.</param>
         /// <param name="info">Extra information to identify <paramref name="instance"/> in the cache.</param>
         public void Cache(object instance, CacheInfo info)
-            => CacheAs(instance.GetType(), info, instance, false);
+        {
+            if (instance == null)
+                throw new ArgumentNullException(nameof(instance));
```
turbedi, line 43:
```suggestion
            ArgumentNullException.ThrowIfNull(instance);
```

bdach:
Why are tests failing here? I'd expect failures in the sourcegen path until a package bump or something, but the reflection path is failing too?

bdach:
Reading the OP:

```csharp
Cache<int?>(0);                // Warning in NRT contexts
```

Why would this warn? Looks like perfectly legal code to me...

osu.Framework/Allocation/DependencyContainer.cs
```diff
@@ -17,142 +14,105 @@ public class DependencyContainer : IReadOnlyDependencyContainer
...
-        /// <param name="instance">The instance to cache. Must be or derive from <typeparamref name="T"/>.</param>
-        /// <param name="info">Extra information to identify <paramref name="instance"/> in the cache.</param>
-        public void CacheAs<T>(T instance, CacheInfo info) where T : class
-            => CacheAs(typeof(T), info, instance, false);
+            CacheAs(instance.GetType(), info, instance);
+        }
 
         /// <summary>
-        /// Caches an instance of a type as a type of <paramref name="type"/>. This instance will be returned each time you <see cref="Get(Type)"/>.
+        /// Caches an instance of a type as its most derived type. This instance will be returned each time you <see cref="Get{T}()"/>.
```
bdach, line 48: If we're editing these, I'd probably clarify further to say

```suggestion
        /// Caches an instance of a type as its most derived type as returned by <see cref="object.GetType"/>. This instance will be returned each time you <see cref="Get{T}()"/>.
```

(goes for all other overloads of this)

osu.Framework/Allocation/IReadOnlyDependencyContainer.cs
```diff
@@ -12,101 +10,48 @@ namespace osu.Framework.Allocation
...
-            where T : class
-        {
-            value = container.Get<T>(info);
-            return value != null;
-        }
+        bool TryGet<T>(out T value, CacheInfo info);
+
+        object? Get(Type type);
+
+        object? Get(Type type, CacheInfo info);
```
bdach, line 48: What are these for? It's not super clear from reading diff but seeing `object?` return type on public API doesn't feel great...

osu.Framework/Allocation/DependencyContainer.cs
```diff
@@ -164,14 +124,38 @@ internal void CacheAs(Type type, CacheInfo info, object instance, bool allowValu
             cache[info] = instance;
         }
 
-        public object Get(Type type)
-            => Get(type, default);
+        public T Get<T>() => Get<T>(default);
+
+        public T Get<T>(CacheInfo info)
+        {
+            TryGet(out T value, info);
```
bdach, line 131: A `TryGet()` call with no return value check? What's going on here? Isn't this gonna end up being wrong for reference types?

osu.Framework.Tests/Dependencies/Reflection/CachedAttributeTest.cs
```diff
@@ -149,29 +153,22 @@ public void TestCacheStructInternal()
 
             var dependencies = DependencyActivator.MergeDependencies(provider, new DependencyContainer());
 
-            Assert.AreEqual(provider.CachedObject.Value, dependencies.GetValue<CachedStructProvider.Struct>().Value);
+            Assert.AreEqual(provider.CachedObject.Value, dependencies.Get<CachedStructProvider.Struct>().Value);
```
bdach, line 156: This test is pretty much superseded by `TestCacheStruct` it looks like. Probably can be removed.

osu.Framework.Tests/Dependencies/Reflection/CachedAttributeTest.cs
```diff
@@ -149,29 +153,22 @@ public void TestCacheStructInternal()
...
         }
 
         [Test]
-        public void TestGetValueNullInternal()
+        public void TestGetNullInternal()
         {
-            Assert.AreEqual(default(int), new DependencyContainer().GetValue<int>());
+            Assert.AreEqual(default(int), new DependencyContainer().Get<int>());
```
bdach, line 162: This creates a duality in semantics in that a "required value-type dependency" does not practically exist. If you forget to cache one, anyone that tries to resolve it will just get `default` back.

If the argument here is that that is fine and matches the general type semantics of C# in general, that's fine, but should probably be documented better.

osu.Framework.Tests/Dependencies/Reflection/ResolvedAttributeTest.cs
```diff
@@ -141,26 +142,13 @@ public void TestResolveInternalStruct()
...
-        }
-
         [Test]
         public void TestResolveStructWithoutNullPermits()
         {
-            Assert.Throws<DependencyNotRegisteredException>(() => new DependencyContainer().Inject(new Receiver14()));
+            var receiver = new Receiver14();
+
+            Assert.DoesNotThrow(() => new DependencyContainer().Inject(receiver));
+            Assert.AreEqual(0, receiver.Obj);
```
bdach, line 151: Again, this test is basically `TestResolveDefaultStruct` isn't it?

osu.Framework.Tests/Dependencies/SourceGeneration/CachedModelDependenciesTest.cs
bdach, line 1: Why are there changes in this test but not the reflection variant of it?

bdach (commented):
rather difficult diff to get through

smoogipoo:
> Why would this warn? Looks like perfectly legal code to me...

Because caching `null` is not allowed [1](https://github.com/ppy/osu-framework/pull/6159/files#diff-8dfa47c5818e59a59dddd611022829a33487f3eddca0dcc005bd80215677eb85R42) [2](https://github.com/ppy/osu-framework/pull/6159/files#diff-8dfa47c5818e59a59dddd611022829a33487f3eddca0dcc005bd80215677eb85R108-L146), therefore attempting to cache a may-be-null value warns about that.

smoogipoo:
Almost thinking all of these tests need to be rewritten...

PR #6160, opened 2024-01-29
Fix nested absolute sequences

smoogipoo:
Whoops. I forgot to consider this possibility.

Fixes the errors in https://github.com/ppy/osu/pull/26772#issuecomment-1914287313


peppy:
Tested against osu!, looks fine and tests work now.

peppy (approved):

PR #6161, opened 2024-01-29
Fix `SpriteIcon` flickering for one frame

EVAST9919:


frenzibyte:
On a quick look, setting the icon property calls `updateTexture` only when the sprite is in a loaded state, but now the initial `updateTexture` call happens before that, leading to a potential case where the icon isn't updated anymore.

This should occur with a simple scenario such as:
```csharp
Child = icon = new SpriteIcon
{
    Icon = A,
}

// here, icon should have had BDL called with updateTexture displaying icon A.

icon.Icon = B;

// here, updateTexture shouldn't have been called in the set above because `icon.IsLoaded` is false (should be in "ready" state).

// expected: icon shows B, actual: icon shows A
```

I haven't checked this though, so correct me if I'm wrong, with a reason why.

bdach (approved):
Seems fine to me

PR #6162, opened 2024-01-30
Revert "Fix `IBindableList.GetEnumerator()` boxing and allocating"

bdach:
- Closes https://github.com/ppy/osu/issues/26809

Android [decided it didn't like this optimisation](https://sentry.ppy.sh/organizations/ppy/issues/?end=2024-01-30T17%3A11%3A59&groupStatsPeriod=auto&project=2&query=firstRelease%3Aosu%402024.130.2+os.name%3AAndroid&referrer=issue-list&sort=freq&start=2024-01-30T11%3A10%3A00). Not exactly surprised given the track record we have with these sorts of tricky changes (something inside mono _reeeeeally_ doesn't like interface level trickery).


Susko3 (approved):
Tested to work fine with playing a beatmap on osu! when building `Release`.

peppy (approved):

PR #6165, opened 2024-02-01
Fix rearrangeable lists don't react on collection `Move` events

Feodor0090:
See added test.

### DIFF
--- a/osu.Framework.Tests/Visual/UserInterface/TestSceneRearrangeableListContainer.cs
+++ b/osu.Framework.Tests/Visual/UserInterface/TestSceneRearrangeableListContainer.cs
@@ -290,6 +290,43 @@ namespace osu.Framework.Tests.Visual.UserInterface
             AddUntilStep("only one item", () => delayedList.ChildrenOfType<BasicRearrangeableListItem<int>>().Count() == 1);
         }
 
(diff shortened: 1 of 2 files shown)


bdach:
Just for general information, can you provide context where you encountered this / what this is for?

Feodor0090:
> Just for general information, can you provide context where you encountered this / what this is for?

I have two screens with lists bound to single `BindableList`, i noticed that after i drag an item in one screen, list in the other screen doesn't update. Test shows this case.

bdach:
So this is in some external project / un-PR'd code? It's not related to any issue in the game?

Feodor0090:
> So this is in some external project / un-PR'd code? It's not related to any issue in the game?
No, it is not.

Feodor0090:
I had feeling that i saw cross-bound rearrangeable usages somewhere in osu!, well, i managed to find one.

https://github.com/ppy/osu-framework/assets/53872073/98bee4d2-f52d-458e-9e01-823a0b6082af

Exactly my use case: two screens, two bound lists. Notice that when i move "easy" diff down, it's still on top when i return to playlist setup screen.

smoogipoo (approved):

PR #6167, opened 2024-02-01
Allow to customize popover blocker

Feodor0090:
In my project I need to skip all non-left mouse buttons through popover container. I also use custom logic that has conditions not to close popovers on outside clicks. To do this i had to copy-paste out the whole popover subsystem just to set my own `DismissOnMouseDownContainer`. I want to switch to built-in one again so proposing this.

This allows to override component that handles outside clicks in `PopoverContainer`. This is no-op change for existing usages.

### DIFF
--- a/osu.Framework/Graphics/Cursor/PopoverContainer.cs
+++ b/osu.Framework/Graphics/Cursor/PopoverContainer.cs
@@ -16,7 +16,7 @@ namespace osu.Framework.Graphics.Cursor
     public partial class PopoverContainer : Container
     {
         private readonly Container content;
-        private readonly Container dismissOnMouseDownContainer;
+        private readonly Container<Drawable> dismissOnMouseDownContainer;
 
         private IHasPopover? target;
         private Popover? currentPopover;
@@ -177,7 +177,7 @@ namespace osu.Framework.Graphics.Cursor
             return availableSize;
         }
 
-        private partial class DismissOnMouseDownContainer : Container
+        private partial class DismissOnMouseDownContainer : Container<Drawable>
         {
             protected override bool OnMouseDown(MouseDownEvent e)
             {
--- a/osu.Framework/Graphics/Cursor/PopoverContainer.cs
+++ b/osu.Framework/Graphics/Cursor/PopoverContainer.cs
@@ -31,10 +31,7 @@ namespace osu.Framework.Graphics.Cursor
                 {
                     RelativeSizeAxes = Axes.Both,
                 },
-                dismissOnMouseDownContainer = new DismissOnMouseDownContainer
-                {
-                    RelativeSizeAxes = Axes.Both
-                }
+                dismissOnMouseDownContainer = CreateInternalPopoverContainer().With(x => x.RelativeSizeAxes = Axes.Both)
             };
         }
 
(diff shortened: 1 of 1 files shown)


bdach:
I really do not like this api surface... It feels way too internal to expose.

Can you describe your use case? Video of your expected behaviour perhaps?

Feodor0090:
> Can you describe your use case?

<details>
<summary>Long explanation what's going on</summary>

I use popovers for all dropdowns and dropdown-like panels in toolbars. Quick example from osu! is music player (i.e. you click a button in toolbar and get a panel near it, osu! uses its own logic, i use popovers for that).
Sometimes they contain _a lot of_ checkboxes/scrolls/etc. so general popover logic "if you don't need it, just ignore it" stops working, you probably want to explicitly close such menu first and only then do something else. So, if there is a popover, i block left clicks so they close the popover but do nothing else (like context menus).

In some cases (they are not there yet, but I think they will be) such menu may have something that should not be lost on accidental outside click. I exposed `Func<bool>` from my popover for future usage.

Of course, such menus may contain components with context menus. Quick solution is to move context menu container inside a popover, but this won't work because while context menu is opened, i dim everything behind it and its target. I want this dim to applied to whole screen. Also, first click outside both popover and child context menu IMO should close menu bc it's more nested. So, "dismisser" must stop only left clicks (so right ones can reach context menu container).

To block context menus from opening on components behind popover, i use `IHasContextMenu` on "dismisser" container.

</details>

To do all described above, I used custom "dismisser":
```csharp
private partial class MLDismissOnMouseDownContainer : Container, IHasMLMenu
{
    {
        if (e.Button == MouseButton.Left && Count > 0)
        {
            var p = (BellePopover)this[0];
            if (p.CanClose())
                this.HideBellePopover();
            return true;
        }
        return false;
    }
    public MLMenuItem[]? PositionalContextMenuItems => Count > 0 ? Array.Empty<MLMenuItem>() : null;
    public MLMenuItem[]? NonPositionalContextMenuItems => PositionalContextMenuItems;
}
``` 
so my popover container is created like this
```csharp
public partial class MLPopoverContainer : Container<Drawable>
{
    public MLPopoverContainer()
    {
        InternalChildren = new Drawable[]
        {
            content = new Container
            {
                RelativeSizeAxes = Axes.Both,
            },
            dismissOnMouseDownContainer = new MLDismissOnMouseDownContainer
            {
                RelativeSizeAxes = Axes.Both
            }
        };
    }
```

As you can see, the only altered logic is internal popover container that this PR allows to do without copy-pasting out the whole sybsystem.

<details>
<summary>video</summary>

ad-hoc demo scene:

https://github.com/ppy/osu-framework/assets/53872073/353d9f1d-c846-4615-898e-244ebcf2d07c

</details>

bdach:
> I use popovers for all dropdowns and dropdown-like panels in toolbars.
> (...)
> Sometimes they contain a lot of checkboxes/scrolls/etc. so general popover logic "if you don't need it, just ignore it" stops working, you probably want to explicitly close such menu first and only then do something else

I would say that this is misusing what the component is supposed to be. As per https://developer.apple.com/design/human-interface-guidelines/popovers:

> **Use a popover to expose a small amount of information or functionality.** Because a popover disappears after people interact with it, limit the amount of functionality in the popover to a few related tasks. For example, a calendar event popover makes it easy for people to change the date or time of an event, or to move it to another calendar. The popover disappears after the change, letting people continue reviewing the events on their calendar.

I would say that if you want longer-lasting overlays like the osu! music player, then rolling your own logic is the best course forward for both the framework and your project. As such I am against merging this change.

PR #6168, opened 2024-02-02
Fix tablet events being enqueued when window is inactive

btmxh:
This is a simple fix for ppy/osu#26284 (at least on my system). Maybe adding this check should be considered for other input handlers.


bdach:
I would like to think this should be getting handled by

https://github.com/ppy/osu-framework/blob/eecf3b19533cfcfadd5adb515a19a6cf15e7d53e/osu.Framework/Input/UserInputManager.cs#L34-L84

so if it's not that direction should be investigated instead as it's likely to be a SDL failure (that `CursorInWindow` flag is based on mouse enter/leave events) or just incorrect usage and that code should be using `IsActive` too / instead.

Adding such checks to input handlers directly sounds 100% wrong.

PR #6169, opened 2024-02-02
Upgrade to .NET 8 and C# 12

### LINKED ISSUE #6030: .NET MAUI 6 has been unsupported for some time (preventing iOS 17 support)
We are currently targeting .NET MAUI 6 (i.e. `net6.0-ios`/`net6.0-android`), which has been "prematurely" out of support for some time now (see [here](https://dotnet.microsoft.com/en-us/platform/support/policy/maui#release-cadence) for more information). Now, with both iOS 17 and Xcode 15's release, updates to the .NET iOS toolchain have only been offered to .NET MAUI 7 (starting at .NET SDK 7.0.402), and thus osu! and o!f cannot deploy to iOS 17 devices for the time being.

Until we move to .NET 7/8, I'm opening this issue thread to keep track of this limitation. I have also set up https://github.com/ppy/osu-framework/compare/master...frenzibyte:osu-framework:ios-17-master to still be able to deploy to my devices in the interim.

smoogipoo:
- Closes #6030 

Don't merge this yet. osu! may be updated to .NET 8 _before_ osu!framework.

Upgrading to C# 12 is required because we use a few constructs that are no longer supported by newer SDKs on older C# versions.

I've tested this on iOS but not Android.


Susko3:
I've fixed some android inspections in https://github.com/ppy/osu-framework/pull/6171.

What remains is the obsoleted APIs in `AndroidGameView`. Those can be ignored (with `#pragma`) for now.

But the `targetSdkVersion` needs addressing, for every android project:
> The TargetFrameworkVersion (Android API level 34) is higher than the targetSdkVersion (33). Please increase the `android:targetSdkVersion` in the `AndroidManifest.xml` so that the API levels match.

bdach:
Android seems to launch fine, at least whatever's runnable framework-side (visual tests & sample game).

bdach (approved):

peppy (approved):

PR #6172, opened 2024-02-03
Return a single vector in `InputResampler`

EVAST9919:
Because it always does so anyway.


peppy:
Please check blame history for this. IRL discussion with smoogi it was decided we'd prefer returning the raw `List<>` because it may return more than one point in the future.

EVAST9919:
Alright then

peppy:
You're welcome to change this PR to do precisely that if you want!

PR #6174, opened 2024-02-04
Add `[RequireStaticDelegate]` annotations

Susko3:
Adds [`[RequireStaticDelegate]`](https://www.jetbrains.com/help/resharper/Reference__Code_Annotation_Attributes.html#RequireStaticDelegateAttribute) annotation to places where there is an `Action<T>` + `T data` overload and the intent is to avoid capture.

Adding this to `InvokeOnDisposal` is a no-brainer, but I'm not entirely sure about adding to `Scheduler` and other places. So this is an RFC of sorts.

There are valid use cases of non-static delegates, so I'm thinking it would be useful to apply the attribute, but change the inspection severity to a hint -- let the programmer know that context is captured.
For cases where we want to raise an error for capturing context (eg. in `InvokeOnDisposal`), `[RequireStaticDelegate(IsError = true)]` can be used. This will force an error regardless of the configured severity. Unless the severity is configured as `DO_NOT_SHOW`, in which case the error is silenced.

image

Example of setting `LambdaShouldNotCaptureContext` severity to hint. If this is considered too noisy, then we don't apply the attribute in `Scheduler`, etc.

# Inspections that can be configured

## LambdaExpressionMustBeStatic

Active when the lambda doesn't capture context, but isn't `static`.

image

I think this is best set to `hint` (defaults to `suggestion`).

## [LambdaShouldNotCaptureContext](https://www.jetbrains.com/help/resharper/LambdaShouldNotCaptureContext.html)

Active when the lambda captures context (eg. variables, `this`):

image

Including when a method is passed and `this` is implicitly captured:

image


smoogipoo:
I like this and I started going down this path with the recent osu-side issues.

I think we should use IsError liberally, including the ezisting cases in this PR.

bdach:
> I'm not entirely sure about adding to `Scheduler` and other places

Shouldn't be applied to scheduler at any level higher than a hint at most I think. Using `Scheduler.AddOnce()` for instance like that has semantics implications.

smoogipoo (approved):

PR #6175, opened 2024-02-05
Fix obsolete code in Android safe area handling by using AndroidX

Susko3:
This replaces obsolete calls with AndroidX compat classes that allow using new-style APIs on older android platforms.

Eg. [Display.getRealSize()](https://developer.android.com/reference/android/view/Display#getRealSize(android.graphics.Point)) is deprecated on API 31+ and the suggested alternative is [WindowManager.getCurrentWindowMetrics()](https://developer.android.com/reference/android/view/WindowManager#getCurrentWindowMetrics()) (API 30+), so instead of having to call the old API on old platforms and the new API on new platforms, we use AndroidX [WindowMetricsCalculator](https://developer.android.com/reference/androidx/window/layout/WindowMetricsCalculator) which does that for us.

`Microsoft.Maui.Essentials` is using `Xamarin.AndroidX` libraries ([see nuget](https://www.nuget.org/packages/Microsoft.Maui.Essentials#dependencies-body-tab)), so it's safe and supported to use them.

### DIFF
--- a/osu.Framework.Android/AndroidGameView.cs
+++ b/osu.Framework.Android/AndroidGameView.cs
@@ -187,24 +187,17 @@ namespace osu.Framework.Android
                 RenderGame();
         }
 
+        public override WindowInsets? OnApplyWindowInsets(WindowInsets? insets)
+        {
+            updateSafeArea();
+            return base.OnApplyWindowInsets(insets);
+        }
+
         [STAThread]
         public void RenderGame()
         {
             // request focus so that joystick input can immediately work.
             RequestFocus();
-
-            LayoutChange += (_, _) => updateSafeArea();
-
-            Activity.IsActive.BindValueChanged(active =>
-            {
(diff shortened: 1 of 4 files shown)


osu.Framework.Android/AndroidGameView.cs
```diff
@@ -188,25 +187,18 @@ protected override void OnLoad(EventArgs e)
...
 
-            LayoutChange += (_, _) => updateSafeArea();
-
-            Activity.IsActive.BindValueChanged(active =>
-            {
-                // When rotating device 180 degrees in the background,
-                // LayoutChange doesn't trigger after returning to game.
-                // So update safe area once active again.
-                if (active.NewValue)
-                    updateSafeArea();
```
Susko3, line 205: I've tested this scenario on two of my devices, and there are no issues when using `OnApplyWindowInsets()`.

bdach (approved):

PR #6176, opened 2024-02-06
Ignore missing feedback api

Feodor0090:
This exists only on win8+: https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowfeedbacksetting#requirements

Before:
[1707225741.runtime.log](https://github.com/ppy/osu-framework/files/14180973/1707225741.runtime.log)

After:
image_2024-02-06_16-40-43

### DIFF
--- a/osu.Framework/Platform/Windows/Native/Input.cs
+++ b/osu.Framework/Platform/Windows/Native/Input.cs
@@ -74,8 +74,16 @@ namespace osu.Framework.Platform.Windows.Native
 
         public static unsafe void SetWindowFeedbackSetting(IntPtr hwnd, FeedbackType feedback, bool configuration)
         {
-            int config = configuration ? 1 : 0; // mimics win32 BOOL type.
-            SetWindowFeedbackSetting(hwnd, feedback, 0, sizeof(int), &config);
+            try
+            {
+                int config = configuration ? 1 : 0; // mimics win32 BOOL type.
+                SetWindowFeedbackSetting(hwnd, feedback, 0, sizeof(int), &config);
+            }
+            catch
+            {
+                // https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowfeedbacksetting#requirements
+                // this API only exists in Win8+.
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


smoogipoo:
I think it'd be better if this was try-caught in the native function, for if anything else uses it they'd need to be aware that this is the case.

Also, can you test that calling another function like `GetSystemMetrics()` from that class doesn't crash? The way I remember these used to work is that they were `TypeLoadException`s or similar (basically as part of the static constructor), while that doesn't look to be the case here.

Feodor0090:
TL:DR; it works.

```diff
diff --git a/osu.Framework/Platform/Windows/WindowsWindow.cs b/osu.Framework/Platform/Windows/WindowsWindow.cs
index fa4c144a4..7be5cb008 100644
--- a/osu.Framework/Platform/Windows/WindowsWindow.cs
+++ b/osu.Framework/Platform/Windows/WindowsWindow.cs
@@ -6,6 +6,7 @@
 using System.Runtime.InteropServices;
 using System.Runtime.Versioning;
 using osu.Framework.Input.Handlers.Mouse;
+using osu.Framework.Logging;
 using osu.Framework.Platform.SDL2;
 using osu.Framework.Platform.Windows.Native;
 using osuTK;
@@ -87,6 +88,10 @@ public override void Create()
             foreach (var feedbackType in Enum.GetValues<FeedbackType>())
                 Native.Input.SetWindowFeedbackSetting(WindowHandle, feedbackType, false);
 
+            // another native call:
+            var rect = Native.Input.GetVirtualScreenRect();
+            Logger.Log($"Virtual screen rect: {rect.X}:{rect.Y} {rect.Width}x{rect.Height}");
+
             // enable window message events to use with `OnSDLEvent` below.
             SDL.SDL_EventState(SDL.SDL_EventType.SDL_SYSWMEVENT, SDL.SDL_ENABLE);
         }
```

[1707229158.runtime.log](https://github.com/ppy/osu-framework/files/14181486/1707229158.runtime.log)

smoogipoo (approved):
I'm pretty happy this in even if we don't support Windows 7 anymore. We should try support as far back as possible in general imo.

PR #6182, opened 2024-02-17
Generalise markdown code block implementation to support indented style

frenzibyte:
- Closes https://github.com/ppy/osu/issues/27194

We commonly use the fenced style for code blocks (the ` ``` ` symbol), but there exists another style for code blocks in markdown, known as [indented code blocks](https://www.jetbrains.com/help/hub/markdown-syntax.html#indented-code-blocks). 

Markdig already handles everything for us and outputs the above as `CodeBlock` objects, but our current implementation was handling `FencedCodeBlock` specifically. Both styles render similarly, so I generalised the implementation to read from `CodeBlock`s instead of `FencedCodeBlock`s.

`CodeBlock` is a superclass of `FencedCodeBlock`, so this PR doesn't have to do extra logic for `FencedCodeBlock` objects.

### DIFF
--- a/osu.Framework/Graphics/Containers/Markdown/MarkdownContainer.cs
+++ b/osu.Framework/Graphics/Containers/Markdown/MarkdownContainer.cs
@@ -241,8 +241,8 @@ namespace osu.Framework.Graphics.Containers.Markdown
                     container.Add(CreateQuoteBlock(quoteBlock));
                     break;
 
-                case FencedCodeBlock fencedCodeBlock:
-                    container.Add(CreateFencedCodeBlock(fencedCodeBlock));
+                case CodeBlock codeBlock:
+                    container.Add(CreateCodeBlock(codeBlock));
                     break;
 
                 case Table table:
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/Containers/Markdown/MarkdownContainer.cs
```diff
@@ -241,8 +241,8 @@ protected virtual void AddMarkdownComponent(IMarkdownObject markdownObject, Fill
                     container.Add(CreateQuoteBlock(quoteBlock));
                     break;
 
-                case FencedCodeBlock fencedCodeBlock:
-                    container.Add(CreateFencedCodeBlock(fencedCodeBlock));
+                case CodeBlock codeBlock:
```
bdach, line 244: I was skeptical of this because these changes are rarely so simple (except if an inheritance tree is properly constructed). I'd argue that markdig's isn't, in this case, because it is:

```
CodeBlock
    FencedCodeBlock
        MathBlock
    YamlFrontMatterCodeBlock
```

Like okay I can see this making sense _syntactically_ for a parser, but not _semantically_. I would wager it is rather rare that a `YamlFrontMatterCodeBlock` should be _visually_ handled as a plain `CodeBlock`.

That said...... framework doesn't enable yaml frontmatter in the pipeline by default, and [game does the correct things](https://github.com/ppy/osu/blob/398eaee5d0b98993b2e5a53e072d4ae7e88c38b4/osu.Game/Graphics/Containers/Markdown/OsuMarkdownContainer.cs#L33-L35) [to compensate for this already](https://github.com/ppy/osu/blob/398eaee5d0b98993b2e5a53e072d4ae7e88c38b4/osu.Game/Overlays/Wiki/Markdown/WikiMarkdownContainer.cs#L37-L39), so I think I'm looking away...

The one that is the other problem is the math fenced thing, since a

```
$$
math block
$$
```

will now strip the dollar signs both sides and the text will print as a code block, but if _that_ is to be fixed then the math markdig extension shouldn't be enabled by default.

Either way I think neither issue is actionable here.

bdach (approved):

PR #6184, opened 2024-02-18
Use ThrowHelper methods in more places

turbedi:
- Might help with https://github.com/ppy/osu-framework/issues/3470

ObjectDisposedException.ThrowIf(), ArgumentException.ThrowIfNullOrEmpty() and the ArgumentOutOfRangeException methods have been introduced in .NET 7/8

### DIFF
--- a/osu.Framework/Bindables/BindableNumber.cs
+++ b/osu.Framework/Bindables/BindableNumber.cs
@@ -42,8 +42,7 @@ namespace osu.Framework.Bindables
                 if (precision.Equals(value))
                     return;
 
-                if (value.CompareTo(default) <= 0)
-                    throw new ArgumentOutOfRangeException(nameof(Precision), value, "Must be greater than 0.");
+                ArgumentOutOfRangeException.ThrowIfNegativeOrZero(value.CompareTo(default));
 
                 SetPrecision(value, true, this);
             }
--- a/osu.Framework/Bindables/BindableNumber.cs
+++ b/osu.Framework/Bindables/BindableNumber.cs
@@ -42,7 +42,8 @@ namespace osu.Framework.Bindables
                 if (precision.Equals(value))
                     return;
 
-                ArgumentOutOfRangeException.ThrowIfNegativeOrZero(value.CompareTo(default));
+                if (value.CompareTo(default) <= 0)
+                    throw new ArgumentOutOfRangeException(nameof(Precision), value, "Must be greater than 0.");
 
                 SetPrecision(value, true, this);
             }
--- a/osu.Framework/Graphics/Containers/GridContainer.cs
+++ b/osu.Framework/Graphics/Containers/GridContainer.cs
@@ -420,11 +420,8 @@ namespace osu.Framework.Graphics.Containers
         /// <param name="maxSize">The maximum size of this row or column.</param>
         public Dimension(GridSizeMode mode = GridSizeMode.Distributed, float size = 0, float minSize = 0, float maxSize = float.MaxValue)
         {
-            if (minSize < 0)
-                throw new ArgumentOutOfRangeException(nameof(minSize), "Must be greater than 0.");
-
-            if (minSize > maxSize)
-                throw new ArgumentOutOfRangeException(nameof(minSize), $"Must be less than {nameof(maxSize)}.");
+            ArgumentOutOfRangeException.ThrowIfNegative(minSize);
+            ArgumentOutOfRangeException.ThrowIfGreaterThan(minSize, maxSize);
 
             Mode = mode;
             Size = size;
--- a/osu.Framework/Testing/Drawables/TestGroupButton.cs
+++ b/osu.Framework/Testing/Drawables/TestGroupButton.cs
@@ -45,7 +45,8 @@ namespace osu.Framework.Testing.Drawables
         {
             var tests = group.TestTypes;
 
-            ArgumentOutOfRangeException.ThrowIfZero(tests.Length);
+            if (tests.Length == 0)
+                throw new ArgumentOutOfRangeException(nameof(group), tests.Length, "Type array must not be empty!");
 
             Group = group;
 
--- a/osu.Framework/Testing/Drawables/TestGroupButton.cs
+++ b/osu.Framework/Testing/Drawables/TestGroupButton.cs
@@ -45,8 +45,7 @@ namespace osu.Framework.Testing.Drawables
         {
             var tests = group.TestTypes;
 
-            if (tests.Length == 0)
-                throw new ArgumentOutOfRangeException(nameof(group), tests.Length, "Type array must not be empty!");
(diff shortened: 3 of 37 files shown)


osu.Framework/Bindables/BindableNumber.cs
```diff
@@ -42,8 +42,7 @@ public T Precision
                 if (precision.Equals(value))
                     return;
 
-                if (value.CompareTo(default) <= 0)
-                    throw new ArgumentOutOfRangeException(nameof(Precision), value, "Must be greater than 0.");
+                ArgumentOutOfRangeException.ThrowIfNegativeOrZero(value.CompareTo(default));
```
smoogipoo, line 45: Can this be `ThrowIfNegativeOrZero(value)`?

turbedi: Yes should be fine. Will change it.

osu.Framework/Graphics/Containers/GridContainer.cs
```diff
@@ -420,11 +420,8 @@ public class Dimension
         /// <param name="maxSize">The maximum size of this row or column.</param>
         public Dimension(GridSizeMode mode = GridSizeMode.Distributed, float size = 0, float minSize = 0, float maxSize = float.MaxValue)
         {
-            if (minSize < 0)
-                throw new ArgumentOutOfRangeException(nameof(minSize), "Must be greater than 0.");
-
-            if (minSize > maxSize)
-                throw new ArgumentOutOfRangeException(nameof(minSize), $"Must be less than {nameof(maxSize)}.");
+            ArgumentOutOfRangeException.ThrowIfNegative(minSize);
+            ArgumentOutOfRangeException.ThrowIfGreaterThan(minSize, maxSize);
```
Susko3, line 424: Just noting that `maxSize` no will longer appear in the error message.

master:
```cs
System.ArgumentOutOfRangeException : Must be less than maxSize. (Parameter 'minSize')
```

This PR:
```cs
System.ArgumentOutOfRangeException : minSize ('20') must be less than or equal to '10'. (Parameter 'minSize')
Actual value was 20.
```

osu.Framework/Testing/Drawables/TestGroupButton.cs
```diff
@@ -45,8 +45,7 @@ public TestGroupButton(Action<Type> loadTest, TestGroup group)
         {
             var tests = group.TestTypes;
 
-            if (tests.Length == 0)
-                throw new ArgumentOutOfRangeException(nameof(group), tests.Length, "Type array must not be empty!");
+            ArgumentOutOfRangeException.ThrowIfZero(tests.Length);
```
Susko3, line 48: Should be reverted, `tests.Length` is not an argument of this method. The error message doesn't make much sense.
```cs
0 ('tests.Length') must be a non-zero value.
```

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -140,8 +138,7 @@ public float CornerThreshold
             get => cornerThreshold;
             set
             {
-                if (cornerThreshold < 0)
-                    throw new ArgumentOutOfRangeException(nameof(value), "CornerThreshold must not be negative.");
+                ArgumentOutOfRangeException.ThrowIfNegative(cornerThreshold);
```
Susko3, line 141: Wait this is actually wrong on master, it's checking the old `cornerThreshold` and not the newly-set `value`.
Should be changed to

```cs
ArgumentOutOfRangeException.ThrowIfNegative(value);
```

osu.Framework/Utils/RNG.cs
```diff
@@ -61,8 +60,7 @@ public static double NextDouble(double maxValue)
         /// <returns>A double-precision floating point number in the range [minValue,maxValue).</returns>
         public static double NextDouble(double minValue, double maxValue)
         {
-            if (minValue > maxValue)
-                throw new ArgumentOutOfRangeException(nameof(minValue), "The given minimum value must be less than or equal to the given maximum value.");
+            ArgumentOutOfRangeException.ThrowIfGreaterThan(minValue, maxValue);
```
Susko3, line 63: See other comment about `maxValue`/`maxSize` no longer appearing in the error message. Same applies for the other usage of `ThrowIfGreaterThan()` in this file.

osu.Framework/Utils/IncrementalBSplineBuilder.cs
```diff
@@ -117,8 +116,7 @@ public float Tolerance
             get => tolerance;
             set
             {
-                if (tolerance < 0)
-                    throw new ArgumentOutOfRangeException(nameof(value), "Tolerance must not be negative.");
+                ArgumentOutOfRangeException.ThrowIfNegative(tolerance);
```
Susko3, line 119: Should be:

```cs
ArgumentOutOfRangeException.ThrowIfNegative(value);
```

Susko3 (changes requested):
Seems ok but some error messages have changed for the worse.

osu.Framework/Bindables/BindableNumber.cs
Susko3: Having this done via `CompareTo()` results in the actual value being misrepresented. Eg. setting a `BindableFloat.Precision` to `-5f` will show `-1` as the actual value.

```cs
System.ArgumentOutOfRangeException : value.CompareTo(default) ('-1') must be a non-negative and non-zero value. (Parameter 'value.CompareTo(default)')
Actual value was -1.
```

This is better handled via generic math (could just call `ArgumentOutOfRangeException.ThrowIfNegativeOrZero(value)`). But that is a breaking change and should not be done in this PR, so best to revert for now.

turbedi:
I've addressed some of the comments. Should I also revert all usages of ThrowIfGreaterThan() to retain the max value in the error message, or is it fine like that?

smoogipoo:
The other path I can suggest is to make an `ExceptionUtils.ValidateRange()`. Otherwise, I'm _weakly_ okay with the change in messaging but I'd want other reviews.

frenzibyte:
I personally don't feel the exception message matters that much. Sure it might be a little bit cryptic, but the "must be greater than"/"must be lower than" is already giving it out I think.

bdach:
My view on the `ThrowIfGreaterThan()` thing is that I'd be fine with the max value not being printed *only* if it's a constant. So in cases where the max value isn't statically known just by looking at the direct call site it should print the max in the exception message.

In favour of revert.

osu.Framework/Bindables/BindableNumber.cs
turbedi: Done

turbedi:
All comments have been addressed.

peppy (approved):

bdach (approved):

PR #6186, opened 2024-02-19
Cleanup common Renderer methods

smoogipoo:
- Adds `Renderer.DrawVertices()`, which is the method that binds the global UBO.
- Makes `globalUniformBuffer` private.
- Adds `Renderer.BindUniformBuffer()`, which does the state tracking/de-duping of flushes around UBO bindings.
- A few other misc cleanups.

### DIFF
--- a/osu.Framework/Graphics/Rendering/Renderer.cs
+++ b/osu.Framework/Graphics/Rendering/Renderer.cs
@@ -966,6 +966,20 @@ namespace osu.Framework.Graphics.Rendering
         /// <param name="frameBuffer">The framebuffer to use, or null to use the backbuffer (i.e. main framebuffer).</param>
         protected abstract void SetFrameBufferImplementation(IFrameBuffer? frameBuffer);
 
+        /// <summary>
+        /// Deletes a frame buffer.
+        /// </summary>
+        /// <param name="frameBuffer">The frame buffer to delete.</param>
+        public void DeleteFrameBuffer(IFrameBuffer frameBuffer)
+        {
+            while (FrameBuffer == frameBuffer)
+                UnbindFrameBuffer(frameBuffer);
+
+            ScheduleDisposal(DeleteFrameBufferImplementation, frameBuffer);
+        }
+
+        protected abstract void DeleteFrameBufferImplementation(IFrameBuffer frameBuffer);
+
         #endregion
 
         public void DrawVertices(PrimitiveTopology topology, int vertexStart, int verticesCount)
--- a/osu.Framework/Graphics/Rendering/Renderer.cs
+++ b/osu.Framework/Graphics/Rendering/Renderer.cs
@@ -110,6 +110,7 @@ namespace osu.Framework.Graphics.Rendering
 
         private readonly DepthValue backBufferDepth = new DepthValue();
 
+        private readonly Dictionary<string, IUniformBuffer> boundUniformBuffers = new Dictionary<string, IUniformBuffer>();
         private readonly Stack<IVertexBatch<TexturedVertex2D>> quadBatches = new Stack<IVertexBatch<TexturedVertex2D>>();
         private readonly List<IVertexBuffer> vertexBuffersInUse = new List<IVertexBuffer>();
         private readonly List<IVertexBatch> batchResetList = new List<IVertexBatch>();
@@ -228,6 +229,7 @@ namespace osu.Framework.Graphics.Rendering
             Shader?.Unbind();
             Shader = null;
 
+            boundUniformBuffers.Clear();
             viewportStack.Clear();
             projectionMatrixStack.Clear();
             maskingStack.Clear();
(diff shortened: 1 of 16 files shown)


osu.Framework/Graphics/Rendering/Renderer.cs
```diff
@@ -1040,6 +1042,19 @@ internal void SetUniform<T>(IUniformWithValue<T> uniform)
...
+        public void BindUniformBuffer(string blockName, IUniformBuffer buffer)
+        {
+            if (boundUniformBuffers.TryGetValue(blockName, out IUniformBuffer? current) && current == buffer)
+                return;
+
+            FlushCurrentBatch(FlushBatchSource.BindBuffer);
+            SetUniformBufferImplementation(blockName, buffer);
+
+            boundUniformBuffers[blockName] = buffer;
+        }
```
bdach, line 1054: This change weirds me out. The buffer tracking is added to the base `Renderer`, but both renderer implementations still maintain their own `boundUniformBuffers` dictionaries? Is this intentional just for the type safety or something? It sure is going to cause memory overheads (although they're likely not relevant).

osu.Framework/Graphics/Veldrid/Buffers/VeldridFrameBuffer.cs
```diff
@@ -66,18 +63,6 @@ public VeldridFrameBuffer(VeldridRenderer renderer, PixelFormat[]? formats = nul
...
-        {
-            this.renderer = renderer;
-            this.colourTarget = colourTarget;
-            this.mipLevel = mipLevel;
-
-            Texture = renderer.CreateTexture(colourTarget);
-            externalColourTarget = true;
-
-            recreateResources();
-        }
```
bdach, line 79: This ctor blames back so https://github.com/ppy/osu-framework/pull/5508, so not super sure about removing it. If that's a change we'll eventually want to push again it could probably stay.

osu.Framework/Graphics/Veldrid/Buffers/VeldridFrameBuffer.cs
```diff
@@ -127,7 +127,11 @@ protected void Dispose(bool disposing)
             if (isDisposed)
                 return;
 
-            renderer.DeleteFrameBuffer(this);
+            while (renderer.IsFrameBufferBound(this))
+                Unbind();
+
+            deleteResources(true);
```
bdach, line 133: This is just inlining the old method right? Commit message mentions "avoiding redirection" so I just want to make sure I'm understanding correctly and this isn't some lower-level concern.

osu.Framework/Graphics/Rendering/Renderer.cs
```diff
@@ -966,6 +966,18 @@ private void setFrameBuffer(IFrameBuffer? frameBuffer, bool force = false)
 
         #endregion
 
+        public void DrawVertices(PrimitiveTopology topology, int vertexStart, int verticesCount)
+        {
+            if (Shader == null)
+                throw new InvalidOperationException("No shader bound.");
```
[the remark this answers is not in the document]
smoogipoo: Yes, it's for general usage safety. It's probably not strictly required.

osu.Framework/Graphics/Rendering/Renderer.cs
smoogipoo: It was just the simplest way for me to move the most common logic (dedupe/flush) to the base class.

Something to consider is that they can only be "bound" in `DrawVertices`. With the way things currently work, it's unsafe to do it earlier than that because there could be [misordered](https://github.com/ppy/osu-framework/blob/a8baecb020efd6a6aff78ea4fef82346e7b51995/osu.Framework/Graphics/Containers/BufferedContainer_DrawNode.cs#L120-L123) events, for example:
```
shader.BindUniformBuffer()
shader.Bind()
// Or
shader.Bind()
shader.BindUniformBuffer()
```
At least, we don't explicitly state that the bind order must be shader -> uniform buffer...

The situation can probably be improved in the future, perhaps by storing the bound UBOs to the shader instead.

osu.Framework/Graphics/Veldrid/Buffers/VeldridFrameBuffer.cs
smoogipoo: Yes, just inlining it.

osu.Framework/Graphics/Rendering/Renderer.cs
bdach: Is the misordering issue a blocker for e.g. exposing `Renderer`'s dictionary as protected and then using that in renderer implementations, rather than have the base renderer track _as well as_ the implementations?

smoogipoo: That sounds backwards to me because I'm tending towards making more of this class `private` like I did with the global UBO. The remaining `protected` ones are even too much for me.

For implementers I think it makes more sense to be given abstract methods  and store whatever state they need from, rather than to have to piece the parts of `Renderer` that they need at a given point...

osu.Framework/Graphics/Veldrid/Buffers/VeldridFrameBuffer.cs
smoogipoo: I've added it back

osu.Framework/Graphics/Veldrid/Buffers/VeldridFrameBuffer.cs
smoogipoo: Just looking through things again, `GLRenderer` does the same thing. It might be worth moving this to the base `Renderer` class instead...

bdach: Are you going to be trying for that? Not sure whether this PR is safe to re-review otherwise.

smoogipoo: Yep, I've moved it back & added a common method in `Renderer`. Actually, the implementation in `master` is a bit scary because it disposed the FBOs immediately rather than the 3-frame-delayed `ScheduleDisposal`...

peppy (approved):

PR #6187, opened 2024-02-19
Introduce abstraction of `VeldridDevice` and pipelines

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/6186

The goal is to move out most of the low-level device-oriented implementation of Veldrid into a new class called `VeldridDevice` and associated `BasicPipeline`/`GraphicsPipeline` classes.  
This will all be used for other implementations, such as my upcoming deferred renderer, which need support for everything here minus the very `VeldridRenderer`-specific implementations of multiple command lists/etc.

---

One thing I'll also mention, is that in the process of implementing the deferred renderer a while ago I got increasingly annoyed at the difference in code style between blocked methods, lambdas with generics, lambdas without generics, and properties:
```csharp
public void Method()
{
    /**/;
}

public void LambdaWithGenerics<T>()
    where T : X
    => /**/;

public void LambdaWithoutGenerics() => /**/;

public X Property => /**/;
```
When scrolling through code, I found it much easier to read the latter two if the expression was placed on a new line like the former two.

I'm not enforcing this, and I don't mean this to change our code style going forward, but it's something that I've applied to this PR and I hope it's agreeable, for my own sanity.


bdach:
This has merge conflicts around the `DeleteFramebufferImplementation()` method added during review of https://github.com/ppy/osu-framework/pull/6186. Unsure of resolution and impact on the rest of the series, so I'm not attempting to fix.

smoogipoo:
I think it's still fine to implement `DeleteFrameBufferImplementation` locally (both here and the upcoming `DeferredRenderer`. I've fixed the conflicts.

osu.Framework/Graphics/Veldrid/VeldridRenderer.cs
```diff
@@ -18,478 +13,148 @@
...
-            set => Device.SyncToVerticalBlank = value;
+            get => veldridDevice.VerticalSync;
+            set => veldridDevice.VerticalSync = value;
         }
 
         protected internal override bool AllowTearing
         {
-            get => Device.AllowTearing;
-            set => Device.AllowTearing = value;
+            get => veldridDevice.AllowTearing;
```
bdach, line 39: I worry a bit about the number of levels of abstraction being introduced here.

```
IRenderer.AllowTearing
VeldridRenderer.AllowTearing
VeidridDevice.AllowTearing
Veldrid.GraphicsDevice.AllowTearing
```

I suppose I get why this is happening, veldrid appears to embrace the individual renderers' quirks and does not appear to attempt to present a unified device interface, but it is a bit painful to look at...

osu.Framework/Graphics/Veldrid/VeldridDevice.cs
```diff
@@ -0,0 +1,466 @@
...
+
+            graphicsPipeline = new GraphicsPipeline(this);
+            StagingTexturePool = new VeldridStagingTexturePool(this);
+        }
+
+        /// <summary>
+        /// Notifies the device that a new frame has started.
+        /// </summary>
+        /// <param name="windowSize">The window size.</param>
+        public void NewFrame(Vector2I windowSize)
```
bdach, line 257: All of the methods here I have a weird time with.

- `VeldridRenderer` generally interacts with the `VeldridDevice`.
- If `VeldridRenderer` needs to call a `GraphicsPipeline`, it will do it via `VeldridDevice.Graphics`.
- But some operations like this one are pulled *out* to the device level. Rather than having them be `GraphicsPipeline` operations. But then stuff like `GraphicsPipeline.Begin()` will operate on the... device again?

Like to me, either the operation should be fully implemented in `Device` without the double-back, or `Device` should be just a dumb veldrid adapter with the ctor and `GraphicsPipeline` should be operating start-to-end on it. This weird half-way back-and-forth interaction between a device and a pipeline makes grasping what is even happening here rather difficult. The distinction seems fuzzy.

bdach (commented):
basic pass on interface structure, code not scrutinised closely nor tested

osu.Framework/Graphics/Veldrid/Pipelines/BasicPipeline.cs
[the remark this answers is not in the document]
smoogipoo: > Put this way: is there ever going to be any `BasicPipeline` that isn't `GraphicsPipeline`?

There already is - `VeldridRenderer` creates two extra `BasicPipeline`s whose entire purpose is to upload data in very specific scenarios. I intend(ed) to not change that.

We could make everything a `GraphicsPipeline` and remove `BasicPipeline`, if that's what you're suggesting. I think that `GraphicsPipeline` is a bit heavy-handed because it does quite a bit of state tracking, even though it's the bare minimum for rendering to work.

I'm definitely stretching it, but I imagine it'll be easier with these split into smaller components if for example at some point a `ComputePipeline` comes along that has to do its own state tracking separate from `GraphicsPipeline`.  
Take that idea with a grain of salt. My vision going into this was only to separate logic into smaller pieces that a consumer can do as they want with, without circling back to a component like `VeldridRenderer` managing everything in its own hacky way that a consumer may not want. I have no particular future goal in mind.

By the way, I originally intended for `GraphicsPipeline` to be totally independent and not included with the `VeldridDevice`. The difficult part in that is that the `Fence` (controlling `VeldridStagingTexturePool`) needs to be on the graphics pipeline - in other words we need to know when a graphics frame finishes and not a random pipeline execution.

> Which, sure, but the `IBasicPipeline` stuff is a bit mystical to me.

The vibe I'm getting is that the `IBasic`/`IGraphics`/`Basic`/`Graphics` structure is a bit too much. I agree. The only reason it exists is to hide `Begin()`/`End()` for `VeldridDevice.Graphics`...

> but if this is PR'd in isolation then I'll be reviewing it in isolation

This should be reviewed in isolation. Unlike `VeldridRenderer`, `DeferredRenderer` only uses the one `GraphicsPipeline`.

osu.Framework/Graphics/Veldrid/VeldridDevice.cs
```diff
@@ -0,0 +1,466 @@
...
+
+        /// <summary>
+        /// The current execution of this <see cref="BasicPipeline"/>.
+        /// </summary>
+        public ulong ExecutionIndex { get; private set; }
+
+        /// <summary>
+        /// The staging texture pool.
+        /// </summary>
+        public readonly VeldridStagingTexturePool StagingTexturePool;
```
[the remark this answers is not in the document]
smoogipoo: I agree. See https://github.com/ppy/osu-framework/pull/6187#discussion_r1505580264, and this is similar - this was originally a private member in `BasicPipeline`, but the only reason it's here is because it has to be attached to the graphics execution, and I wanted to keep `GraphicsPipeline` to bare essentials...

May see if this can be refactored/rethought. Accepting of suggestions.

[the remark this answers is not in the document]
bdach: I meant it in the far more mundane sense of just making a public property out of it for consistency.

osu.Framework/Graphics/Veldrid/Pipelines/BasicPipeline.cs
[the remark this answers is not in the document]
bdach: > There already is - `VeldridRenderer` creates two extra `BasicPipeline`s whose entire purpose is to upload data in very specific scenarios. I intend(ed) to not change that.

I might have not seen that scrolling past a 700-line diffstat file.. 😅 

> We could make everything a `GraphicsPipeline` and remove `BasicPipeline`, if that's what you're suggesting.

In light of the above that sounds wrong to do, so no. It's probably fine to keep the basic implementation.

> The only reason the interfaces exist is to hide `Begin()`/`End()` for `VeldridDevice.Graphics`...

Well the thing about `IBasicPipeline` the interface that specifically confuses me the most is how loose `IBasicPipeline` is basically. By which I mean:

- Rider says `Device` and `UpdateTexture` (both overloads) are unused.
- `Factory` is exposed, but you may also do `Device.ResourceFactory` as well.
- It's supposed to be "non-graphical", but `Commands` is right there, and you can fire draw commands from that which would normally be in the purview of a `GraphicsPipeline`.

It's just very weird to see that used in `VeldridIndexBuffer` in that state. I'd say that I'd probably either (a) use a different interface, _specific_ to the index buffer usage, which has only the operations that that needs and no more, and no way to perform the same operation in more ways than one, or (b), just not bother on the grounds that this is all internals and it's protecting interfaces for the sake of protecting interfaces and any flagrant layering violation should be caught at review time anyway.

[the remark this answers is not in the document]
smoogipoo: https://github.com/smoogipoo/osu-framework/commit/a8058f70cafb7e3b5172106925102c6a3f8a1546

How does this sit with you? I would remove the pipeline interfaces with this, but I haven't included that in the diff.

The weird part is `UpdateTexture` taking in a staging pool parameter, but it does otherwise fit the original mental model I had for this.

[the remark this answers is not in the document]
bdach: On a quick look, I like the decreased indirection in `VeldridRenderer`. Everything else seems like a sidegrade, so overall in favour - but do note that I don't really yet understand any of the intricacies as I only read the code from a high-level perspective. If there's something in that structure that you dislike (because seeing the commit message, I presume you already had things structured that way and changed your mind?), then do let me know because again, I may just be clueless.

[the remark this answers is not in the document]
smoogipoo: I've moved the changes here.

> I presume you already had things structured that way and changed your mind

Previously, the staging pool was part of the `BasicPipeline`. That led to texture corruption in `TestSceneTextureUploadPerformance`. 

I don't know why - likely some async trickery - but the staging resources have to be reset based on the graphics pipeline fence.

osu.Framework/Graphics/Veldrid/Pipelines/GraphicsPipeline.cs
```diff
@@ -0,0 +1,317 @@
...
+
+        /// <summary>
+        /// Sets the active vertex buffer.
+        /// </summary>
+        /// <param name="buffer">The vertex buffer.</param>
+        /// <param name="layout">The layout of vertices in the buffer.</param>
+        public void SetVertexBuffer(DeviceBuffer buffer, VertexLayoutDescription layout)
+        {
+            if (buffer == currentVertexBuffer && layout.Equals(currentVertexLayout))
+                return;
```
bdach, line 176: The `layout.Equals(currentVertexLayout)` check is new here as far as I can tell? What's it for?

Is it just because it was previously hardcoded to use `VeldridVertexUtils<T>.Layout` or something and now isn't?

osu.Framework/Graphics/Veldrid/Pipelines/GraphicsPipeline.cs
```diff
@@ -0,0 +1,317 @@
...
+            : base(device)
+        {
+            pipelineDesc.Outputs = Device.SwapchainFramebuffer.OutputDescription;
+        }
+
+        public override void Begin()
+        {
+            base.Begin();
+
+            attachedTextures.Clear();
```
bdach, line 51: The interesting thing here is that in the old `VeldridRenderer`, the analogous field (`boundTextureUnits`) was never cleared... Intentional change?

Goes for `attachedUniformBuffers` too.

osu.Framework/Graphics/Veldrid/Pipelines/GraphicsPipeline.cs
```diff
@@ -0,0 +1,317 @@
...
+        public void AttachUniformBuffer(string name, IVeldridUniformBuffer buffer)
+            => attachedUniformBuffers[name] = buffer;
+
+        /// <summary>
+        /// Sets the offset of a uniform buffer.
+        /// </summary>
+        /// <param name="buffer">The uniform buffer.</param>
+        /// <param name="bufferOffsetInBytes">The offset in the uniform buffer.</param>
+        public void SetUniformBufferOffset(IVeldridUniformBuffer buffer, uint bufferOffsetInBytes)
+            => uniformBufferOffsets[buffer] = bufferOffsetInBytes;
```
bdach, line 227: What's this for? Nothing uses it.

bdach (commented):
Structure looks simpler, i.e. better.

Pass on actual implementation.

bdach:
Platform testing looks fine, so after clarifications re: above I imagine this'll be good to go.

osu.Framework/Graphics/Veldrid/Pipelines/GraphicsPipeline.cs
smoogipoo: This isn't required for this branch in isolation, but is required for deferred renderer, which keeps one VBO bound with multiple vertex layouts.

I can split this out if necessary.

osu.Framework/Graphics/Veldrid/Pipelines/GraphicsPipeline.cs
smoogipoo: Same as above, this isn't required for this branch in isolation, but is used by the deferred renderer. I can split out if required.

bdach (approved):

osu.Framework/Graphics/Veldrid/Pipelines/GraphicsPipeline.cs
bdach: Nah let's just go forward with this.

PR #6188, opened 2024-02-19
Reduce frequency of global UBO changes

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/6186

Since the pre-req PR adds a common `DrawVertices()` method, we can now update the global UBO as late as possible - right before vertices need to be drawn.

The easiest way to see the difference is by using the GL renderer and noticing the reduction in the `Draw.UniformUpl` statistic. In `TestSceneTextFlow` it reduces uploads from 54 -> 28. This may lead to a minor performance increase.  
The Veldrid implementation does its own similar late-upload, which _could_ be removed but isn't done in this PR.

_In general_ though, this should slightly increase performance (not benchmarked) just from reducing struct copies.

### DIFF
--- a/osu.Framework/Graphics/Rendering/Renderer.cs
+++ b/osu.Framework/Graphics/Rendering/Renderer.cs
@@ -70,7 +70,7 @@ namespace osu.Framework.Graphics.Rendering
         public WrapMode CurrentWrapModeS { get; private set; }
         public WrapMode CurrentWrapModeT { get; private set; }
         public bool IsMaskingActive { get; private set; }
-        public bool UsingBackbuffer => FrameBuffer == null;
+        public bool UsingBackbuffer { get; private set; }
         public Texture WhitePixel => whitePixel.Value;
         DepthValue IRenderer.BackbufferDepth => backBufferDepth;
 
@@ -652,6 +652,7 @@ namespace osu.Framework.Graphics.Rendering
 
             currentMaskingInfo = maskingInfo;
 
+            // Masking is enabled for as long as any masking info is active that's not the default.
             IsMaskingActive = maskingStack.Count > 1;
             globalUniformsChanged = true;
         }
@@ -909,6 +910,8 @@ namespace osu.Framework.Graphics.Rendering
             SetFrameBufferImplementation(frameBuffer);
 
             FrameBuffer = frameBuffer;
+
+            UsingBackbuffer = frameBuffer == null;
             globalUniformsChanged = true;
         }
 
--- a/osu.Framework/Graphics/Rendering/Renderer.cs
+++ b/osu.Framework/Graphics/Rendering/Renderer.cs
@@ -26,6 +26,7 @@ using osuTK.Graphics;
 using SixLabors.ImageSharp;
 using SixLabors.ImageSharp.PixelFormats;
 using RectangleF = osu.Framework.Graphics.Primitives.RectangleF;
+using Texture = osu.Framework.Graphics.Textures.Texture;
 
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/Rendering/Renderer.cs
```diff
@@ -698,6 +651,9 @@ private void setMaskingInfo(MaskingInfo maskingInfo, bool isPushing, bool overwr
                 PopScissor();
 
             currentMaskingInfo = maskingInfo;
+
+            IsMaskingActive = maskingStack.Count > 1;
```
bdach, line 655: This change is making me somewhat nervous. I'm guessing it's necessary to set masking info as late as possible, and nothing seems to be _dropping_ updates of this, but it's a bit shoestring - the only _actual_ point of set is here, `{Pop,Push}MaskingInfo()` calls that, and `BeginFrame()` calls `PushMaskingInfo()` so it checks out, but...

osu.Framework/Graphics/Rendering/Renderer.cs
```diff
@@ -68,8 +69,8 @@ public abstract class Renderer : IRenderer
         public StencilInfo CurrentStencilInfo { get; private set; }
         public WrapMode CurrentWrapModeS { get; private set; }
         public WrapMode CurrentWrapModeT { get; private set; }
-        public bool IsMaskingActive => maskingStack.Count > 1;
-        public bool UsingBackbuffer => frameBufferStack.Count == 0;
+        public bool IsMaskingActive { get; private set; }
+        public bool UsingBackbuffer => FrameBuffer == null;
```
bdach, line 73: Any reason for this change? It doesn't appear to be wrong, but I can't see what this gains.

bdach:
Can't see any issues in platform testing so this is _probably_ ok to go through with although I would appreciate addressing the reviews above.

osu.Framework/Graphics/Rendering/Renderer.cs
smoogipoo: There is actually a reason for this and the masking change above. I'm going to try to explain it, but it takes a bit of a mental hurdle...

There are three important operations that are happening:
```
// IsMaskingActive state changes here 
stack.Push();

// At this point, the OLD IsMaskingActive state is used
FlushCurrentBatch();

// The new masking state is set here
setNewState();
```

In the new implementation, the order has changed:
```
stack.Push();
setNewState();
FlushCurrentBatch();
```

The reason is because the UBO `set()` has been moved to `DrawVertices()`. As far as I can tell, this only affects the `IsMasking` and `UsingBackbuffer` states.

smoogipoo: I could make `UsingBackbuffer` not implemented as a computed property and set in the same way `IsMakingActive` is set. Would that improve things?

bdach: I suspected this was reordering shenanigans at play.

> I'm trying to figure out if there's a better refactoring that makes this clearer

I'm not sure one exists, this is a subtle detail. I'd maybe inline comment at most.

> I could make `UsingBackbuffer` not implemented as a computed property and set in the same way `IsMakingActive` is set. Would that improve things?

Weakly in favour because that would at least unify handling of these two special cases.

smoogipoo: I've refactored it a little bit, but I'm really not sure how to comment it without... stating the obvious? I've added a little bit of whitespace to make it more obvious that it's related to "updating the global uniforms".

PR #6189, opened 2024-02-19
Increase interface surface of Veldrid components

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/6187

Some components are generic enough that they don't need to be reimplemented:
- `VeldridShader`
- `VeldridTexture`

These components use `IVeldridRenderer`.

Other components are a bit more specific to `VeldridRenderer` and either need to be reimplemented, or not implemented at all:
- `VeldridVertexBuffer`
- `VeldridFrameBuffer`

These components use `VeldridRenderer`.


bdach:
@smoogipoo I've resolved conflicts here to the best of my judgement but please double-check. Will not be reviewing for a while anyhow.

osu.Framework/Graphics/Veldrid/Buffers/IVeldridVertexBuffer.cs
```diff
@@ -8,8 +8,7 @@
 
 namespace osu.Framework.Graphics.Veldrid.Buffers
 {
-    internal interface IVeldridVertexBuffer<in T> : IVertexBuffer, IDisposable
-        where T : unmanaged, IEquatable<T>, IVertex
+    internal interface IVeldridVertexBuffer : IVertexBuffer, IDisposable
```
bdach, line 11: I assume this interface is really only used in future pulls? I'm not seeing what this gains in the context of this specific one.

I'm aware that this sort of thing is usually an escape hatch maneuver to be able to do limited operations on the interface without having to specify the `<T>`, but it's difficult to see where this is coming from here.

bdach (commented):
Source looks mostly sane

osu.Framework/Graphics/Veldrid/Buffers/IVeldridVertexBuffer.cs
smoogipoo: It actually doesn't seem to be used, for better or worse - `GraphicsPipeline.SetVertexBuffer()` takes in a `DeviceBuffer` which is pretty weird, and it's non-trivial to make it an `IVeldridVertexBuffer` with upcoming changes.

I've reverted this for now, and can redo in the future if needed.

bdach (approved):

PR #6195, opened 2024-02-23
Prefer using `UserProfile` over `Personal` to account for breaking changes in .NET 8

jvyden:
This PR aims to address the issues regarding recent breaking changes to `Environment.GetFolderPath` for Unix systems. The changes in question are noted by Microsoft [here](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/8.0/getfolderpath-unix), although the changes to Android do not affect o!f; only Linux and MacOS.

I hit the effects of this breaking change in osu!lazer due to admittedly strange circumstances which I'll explain with a timeline.

- I install osu!stable, causing `~/osu` to be created.
- I later install osu!lazer, which then uses this directory.
- I clone osu!lazer to `~/Documents/osu` at some point.
- **Years later, o!f migrates to .NET 8.**
- Lazer tries to check for `SpecialFolder.Personal`, which in .NET 6 would be `$HOME`, but in .NET 8 is now `$HOME/Documents`.
- Lazer sees that this folder exists and uses this directory as the user storage folder.
- The original lazer installation directory at `~/osu` is then ignored.

We can solve this by simply introducing the user's home folder to the list of available storage paths. It was discussed to be a final fallback, however I'm not sure this is a good idea as the problem would still trigger in my scenario.

### DIFF
--- a/osu.Framework/Platform/Linux/LinuxGameHost.cs
+++ b/osu.Framework/Platform/Linux/LinuxGameHost.cs
@@ -45,6 +45,7 @@ namespace osu.Framework.Platform.Linux
                     yield return xdg;
 
                 yield return Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".local", "share");
+                yield return Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
 
                 foreach (string path in base.UserStoragePaths)
                     yield return path;
--- a/osu.Framework/Platform/MacOS/MacOSGameHost.cs
+++ b/osu.Framework/Platform/MacOS/MacOSGameHost.cs
@@ -35,6 +35,7 @@ namespace osu.Framework.Platform.MacOS
                     yield return xdg;
 
                 yield return Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".local", "share");
+                yield return Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
 
                 foreach (string path in base.UserStoragePaths)
                     yield return path;


bdach:
I don't understand what this means:

> I later install osu!lazer, which then uses this directory.

and thus find it very difficult to understand what this is trying to fix.

Can you describe the folder structure in _full_ detail? How are stable and lazer in "one directory"? If they are, I would like to think that that was a _custom_ game location, and as such the path to the storage should be just written directly to `storage.ini`?

> I clone osu!lazer to ~/Documents/osu at some point.

This is ambiguous too. What does "clone" mean? "github clone"? "copy of all game files"?

smoogipoo:
I'm a bit confused how this entire thing is working... Here's my sample app + testing:

```csharp
Console.WriteLine($"UserProfile => {Environment.GetFolderPath(Environment.SpecialFolder.UserProfile)}");
Console.WriteLine($"Personal => {Environment.GetFolderPath(Environment.SpecialFolder.Personal)}");
Console.WriteLine($"LocalApplicationData => {Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData)}");
```

In particular, notice how the `Personal` directory is `string.Empty` on Linux, which could be the cause of issues like https://github.com/ppy/osu/issues/27294 . But this doesn't match the documentation which says this should be `$HOME`???

But also, the above would imply that
```
yield return Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".local", "share");
```
... is returning a non-existent directory, which doesn't seem to be the case because the user in that thread [states that `~/.local/share` exists](https://github.com/ppy/osu/issues/27294#issuecomment-1960900010).

I don't think that we should ever be writing to `$HOME`, and if osu-lazer used this directory it was likely a bug.

Doesn't it seem more appropriate that we use `LocalApplicationData` as the strict path? @bdach @peppy curious on your thoughts.

## Linux

```
$ dotnet run -f net6.0
UserProfile => /home/smgi
Personal => /home/smgi
LocalApplicationData => /home/smgi/.local/share

$ dotnet run -f net8.0
UserProfile => /home/smgi
Personal =>
LocalApplicationData => /home/smgi/.local/share
```

## macOS

```
$ dotnet run -f net6.0
UserProfile => /Users/smgi
Personal => /Users/smgi
LocalApplicationData => /Users/smgi/.local/share

$ dotnet run -f net8.0
UserProfile => /Users/smgi
Personal => /Users/smgi/Documents
LocalApplicationData => /Users/smgi/Library/Application Support
```

## Windows

```
$ dotnet run -f net6.0
UserProfile => C:\Users\smoog
Personal => C:\Users\smoog\OneDrive\Documents
LocalApplicationData => C:\Users\smoog\AppData\Local

$ dotnet run -f net8.0
UserProfile => C:\Users\smoog
Personal => C:\Users\smoog\OneDrive\Documents
LocalApplicationData => C:\Users\smoog\AppData\Local
PS C:\Users\smoog\OneDrive\Desktop\test>
```

## iOS

```
$ net6.0

(could not test)

$ net8.0

UserProfile => /private/var/mobile/Containers/Data/Application/39777EA0-B0DA-4F72-906C-823A16E8CF57
Personal => /var/mobile/Containers/Data/Application/39777EA0-B0DA-4F72-906C-823A16E8CF57/Documents
LocalApplicationData => /var/mobile/Containers/Data/Application/39777EA0-B0DA-4F72-906C-823A16E8CF57/Documents
```

## Android

```
$ net6.0

(could not test)

$ net8.0

UserProfile => /data/user/0/com.companyname.AndroidDirectoryTest/files
Personal => /data/user/0/com.companyname.AndroidDirectoryTest/files/Documents
LocalApplicationData => /data/user/0/com.companyname.AndroidDirectoryTest/files
```

bdach:
> ```
> $ dotnet run -f net6.0
> UserProfile => /home/smgi
> Personal => /home/smgi
> LocalApplicationData => /home/smgi/.local/share
> 
> $ dotnet run -f net8.0
> UserProfile => /home/smgi
> Personal =>
> LocalApplicationData => /home/smgi/.local/share
> ```

I'm not seeing the same.

```
$ dotnet run -f net6.0     
UserProfile => /home/dachb
Personal => /home/dachb
LocalApplicationData => /home/dachb/.local/share

$ dotnet run -f net8.0     
UserProfile => /home/dachb
Personal => /home/dachb/Documents
LocalApplicationData => /home/dachb/.local/share
```

This is probably some distro/xdg chicanery at work.

bdach:
I'm also ultra confused about this on macos:

```
(net6)
ApplicationData => /Users/smgi/.config
LocalApplicationData => /Users/smgi/.local/share

(net8)
ApplicationData => /Users/smgi/Library/Application Support
LocalApplicationData => /Users/smgi/Library/Application Support
```

I thought user files were living under `Application Support` for months now?

---

As for your question I'd probably use `LocalApplicationData` locally and then do something like windows does:

https://github.com/ppy/osu-framework/blob/781db0a67c84faae736579d0ef3e9722714880d4/osu.Framework/Platform/Windows/WindowsGameHost.cs#L32-L33

e.g. suppress the base call.

I'm worried about applying that on a higher level because the fact that none of us can apparently tell what is going on with these path is _terrifying_.

smoogipoo:
It's weird, because the breaking changes state `Personal | Uses XDG_DOCUMENTS_DIR if available; otherwise $HOME/Documents`. I don't have `XDG_DOCUMENTS_DIR` defined, nor do I have `XDG_DATA_HOME` which is what o!f checks against.

To me it really feels like we should be using:

| os | enum | ex. value |
| -- | --- | --- |
| Linux | `LocalApplicationData` | `~/.local/share` |
| macOS | `LocalApplicationData` | `~/Library/Application Support` |
| Windows | `ApplicationData` | `%APPDATA%` |
| iOS | `LocalApplicationData` | |
| Android | `GetExternalFilesDir` | |

smoogipoo:
> I thought user files were living under `Application Support` for months now?

On `macOS` we have an override for that:

https://github.com/ppy/osu-framework/blob/781db0a67c84faae736579d0ef3e9722714880d4/osu.Framework/Platform/MacOS/MacOSGameHost.cs#L30

Just to be clear, my intention would be to remove these weird overrides.

bdach:
> I don't have `XDG_DOCUMENTS_DIR` defined, nor do I have `XDG_DATA_HOME` which is what o!f checks against.

For whatever that's worth I don't think I do either:

```
$ env | grep XDG
XDG_SEAT=seat0
XDG_SESSION_TYPE=x11
XDG_SESSION_CLASS=user
XDG_SESSION_ID=2
XDG_RUNTIME_DIR=/run/user/1000
XDG_CURRENT_DESKTOP=i3
XDG_SESSION_DESKTOP=i3
XDG_VTNR=2
XDG_CONFIG_DIRS=/etc/xdg/xdg-i3:/etc/xdg
XDG_DATA_DIRS=/usr/share/i3:/usr/share/gnome:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop
```

> To me it really feels like we should be using:

Yeah I'd agree with that.

bdach:
Also here's a fun upstream issue thread... https://github.com/dotnet/runtime/issues/93094

Specifically, https://github.com/dotnet/runtime/issues/93094#issuecomment-1751826804:

> If I add `Environment.SpecialFolderOption.DoNotVerify` the method will return /home/user/Documents.

[docs:](https://learn.microsoft.com/en-us/dotnet/api/system.environment.specialfolderoption?view=net-8.0)

> `DoNotVerify`
>
> **The path to the folder is returned without verifying whether the path exists.** If the folder is located on a network, specifying this option can reduce lag time.

That is one of the stupidest API design choices I might have ever seen........

smoogipoo:
Aha - specifying `Environment.SpecialFolderOption.DoNotVerify` gives me the paths on Linux:
```
UserProfile => /home/smgi
Personal => /home/smgi/Documents
ApplicationData => /home/smgi/.config
LocalApplicationData => /home/smgi/.local/share
```
And yes, that `Personal` dir doesn't exist.

I'm also thinking we should use `Environment.SpecialFolderOption.Create`.

bdach:
Yeah everything seems to check out. Thanks for clarifying.

PR #6196, opened 2024-02-23
Adjust user storage paths for better cross-platform support

smoogipoo:
Supersedes / closes https://github.com/ppy/osu-framework/pull/6195

Discussed in https://github.com/ppy/osu-framework/pull/6195.
Discussed in https://github.com/ppy/osu/issues/27294#issuecomment-1961076113

## The problem

In [.NET 8](https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/8.0/getfolderpath-unix), the `Personal` special directory was changed to point to `$HOME/Documents`. On many Linux distros (mine included - a standard arch install), this doesn't exist.

Because we don't specify `Environment.SpecialFolderOption.DoNotVerify` as an argument to `Environment.GetFolderPath()`, the returned string is empty.

This causes a crash because `GameHost` first iterates through all values of `UserStoragePaths`, which now includes an empty string, checking if `./osu` exists in any one.

## The fix

To not deal with this headache ever again, I want to simplify the multiple levels fallbacks. It looks like it's possible to use `LocalApplicationData` on most platforms, which gives equivalent values to what we've expected so far:

| os | enum | ex. value |
| -- | --- | --- |
| Linux | `LocalApplicationData` | `~/.local/share` |
| macOS | `LocalApplicationData` | `~/Library/Application Support` |
| Windows | `ApplicationData` | `%APPDATA%` |
| iOS | `LocalApplicationData` | |
| Android | `GetExternalFilesDir` | |

Note, however, that `XDG_DATA_HOME` is no longer used as an override. This should be equivalent to `~/.local/share`[1](https://wiki.archlinux.org/title/XDG_Base_Directory)[2](https://man.freebsd.org/cgi/man.cgi?query=dbus-daemon&sektion=1&format=html), so I don't expect this to impact many users.


bdach:
> Note, however, that `XDG_DATA_HOME` is no longer used as an override

Not by us, but dotnet should be still reading it: https://github.com/dotnet/runtime/blob/3138206e748ea73329ebe26a8824cbba326d377b/src/libraries/System.Private.CoreLib/src/System/Environment.GetFolderPathCore.Unix.cs#L124-L132

smoogipoo:
Sorry for the force push - I accidentally forgot to commit literally the most important change in here.

> Not by us, but dotnet should be still reading it

Perfect :)

bdach:
Testing:

- Linux
  - [x] Confirmed existence of issue by overwriting `$HOME` to a directory that has no `Documents` subdir
  - [x] Confirmed that this PR fixes it
    - Observed behaviour: `.local/share/osu` is created under overwritten `$HOME`
  - [x] Tested that existing storages are still read correctly after change
  - [x] Tested that new storages are still created in the same place
  - [x] Tested that `$XDG_DATA_HOME` override still works with this change
- Windows
  - [ ] Preventative check that existing storages are still read correctly after change
  - [ ] Preventative check that new storages are still created in the same place
- Android
  - [ ] Preventative check that existing storages are still read correctly after change
  - [ ] Preventative check that new storages are still created in the same place

osu.Framework/Platform/MacOS/MacOSGameHost.cs
```diff
@@ -23,24 +21,6 @@ internal MacOSGameHost(string gameName, HostOptions options)
...
-            get
-            {
-                yield return Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), "Library", "Application Support");
-
-                string xdg = Environment.GetEnvironmentVariable("XDG_DATA_HOME");
-
-                if (!string.IsNullOrEmpty(xdg))
-                    yield return xdg;
-
-                yield return Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".local", "share");
```
peppy, line 37: This will mean that any users which had data at this local will end up with a new install, right? I know that my data was here for some period (it may have been an older default we used in lazer itself).

smoogipoo: Hmm, that true. Can add this back just for macOS.

bdach (approved):
Outside of apple land, looks good to me

peppy (approved):

PR #6199, opened 2024-02-26
Remove bindable from `CircularProgress`

EVAST9919:
There's zero to no reason to have it `Bindable` imo.
Performance impact is tiny, though in cases such as `TestSceneVertexUploadPerformance` it's quite noticeable (which is unrealistic, but still).
|master|pr|
|---|---|
|master|pr|


frenzibyte:
I would look more into whether we can change `ValueChangedEvent` to a `struct` rather than slowly turning away from using `Bindable` in UI components, but turns out it's been attempted before in https://github.com/ppy/osu-framework/pull/6114 and things aren't going good in there.

I think this particular instance makes more sense to not be a bindable though, since it's a continuously changing component, rather than being dependant on user manually changing it (i.e. most other UI components).

frenzibyte (approved):

peppy (approved):

PR #6203, opened 2024-03-04
Use new BCL APIs from .NET 7/8

huoyaoyuan:
The BCL versions are generally safer and faster.

### DIFF
--- a/osu.Framework/Extensions/ExtensionMethods.cs
+++ b/osu.Framework/Extensions/ExtensionMethods.cs
@@ -332,7 +332,8 @@ namespace osu.Framework.Extensions
         /// </remarks>
         /// <param name="character">The character to check.</param>
         /// <returns>True if the character is an ASCII digit.</returns>
-        public static bool IsAsciiDigit(this char character) => character >= '0' && character <= '9';
+        [Obsolete("Use char.IsAsciiDigit.")] // can be removed 20240901
+        public static bool IsAsciiDigit(this char character) => char.IsAsciiDigit(character);
 
         /// <summary>
         /// Converts an osuTK <see cref="DisplayDevice"/> to a <see cref="Display"/> structure.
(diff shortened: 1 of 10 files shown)


osu.Framework/Extensions/ExtensionMethods.cs
```diff
@@ -332,7 +332,8 @@ public static string TrimDirectorySeparator(this string path)
         /// </remarks>
         /// <param name="character">The character to check.</param>
         /// <returns>True if the character is an ASCII digit.</returns>
-        public static bool IsAsciiDigit(this char character) => character >= '0' && character <= '9';
+        [Obsolete("Use char.IsAsciiDigit.")] // can be removed 20240901
+        public static bool IsAsciiDigit(this char character) => char.IsAsciiDigit(character);
```
huoyaoyuan, line 336: CoreLib version is slightly faster for this.

osu.Framework/Utils/MathUtils.cs
```diff
@@ -12,40 +12,32 @@ public static class MathUtils
         /// </summary>
         /// <param name="degrees">An angle in degrees.</param>
         /// <returns>The angle expressed in radians.</returns>
-        public static float DegreesToRadians(float degrees)
-        {
-            return degrees * MathF.PI / 180.0f;
-        }
+        [Obsolete("Use float.DegressToRadians.")] // can be removed 20240901
+        public static float DegreesToRadians(float degrees) => float.DegreesToRadians(degrees);
```
smoogipoo, line 16: Putting these in `float`/`double` is a very _interesting_ API decision to me.

smoogipoo (approved):

osu.Framework/Utils/MathUtils.cs
bdach: [it's a whole shebang.](https://devblogs.microsoft.com/dotnet/dotnet-7-generic-math/) personally abstractions over numbers have been *loooooooooooong* overdue so i'm in hook line and sinker

PR #6210, opened 2024-03-11
Fix over time degradation of deferred renderer performance

smoogipoo:
I happened upon this by total accident. I don't have a reliable way to reproduce this, so I hope the changes make sense on their own.

Because the list isn't cleared, it would keep increasing in size over time and eventually take down the game. For me, this happened abruptly. Also disposal wasn't implemented...

As a side note, I wish to implement this entire flow better in the future but I haven't found a way yet.

### DIFF
--- a/osu.Framework/Graphics/Rendering/Deferred/DeferredUniformBuffer.cs
+++ b/osu.Framework/Graphics/Rendering/Deferred/DeferredUniformBuffer.cs
@@ -80,6 +80,11 @@ namespace osu.Framework.Graphics.Rendering.Deferred
             dead_chunks.Clear();
         }
 
+        ~DeferredUniformBuffer()
+        {
+            Dispose(false);
+        }
+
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Rendering/Deferred/DeferredUniformBuffer.cs
```diff
@@ -77,10 +77,34 @@ void IVeldridUniformBuffer.ResetCounters()
 
             data = default;
             currentChunk = default;
+            dead_chunks.Clear();
+        }
+
+        ~DeferredUniformBuffer()
```
bdach, line 83: What is the finaliser for? Safety? Does this class contain unmanaged resources? Should I be worried about the consequent gc overhead stemming from this having a finaliser?

bdach (approved):

PR #6211, opened 2024-03-11
Fix incorrect disposal pattern

smoogipoo:
This could cause problems if the object is resurrected. It's not the proper disposal pattern.


bdach (approved):
I'm not sure how a `VeldridTexture` could become "resurrected" but this does look closer to how a disposal should look to me.

PR #6212, opened 2024-03-11
Provides generic version of `LifetimeEntryManager`

Frederisk:
I noticed that the `LifetimeEntryManager` class is not generic. This will cause me to have to cast repeatedly if I overload the `LifetimeEntry` class:

```cs
public class MyEntry : LiftTimeEntry {}
// ...
private readonly LifetimeEntryManager manager = new();
// ...
this.manager.AddEntry(new MyEntry());
this.manager.EntryBecameAlive += (LifetimeEntry e) => {
  var my = (MyEntry)e;
}
this.manager.EntryBecameDead += (LifetimeEntry e) => {
  var my = (MyEntry)e;
}
```

I thought maybe we should provide a generic version (LifetimeEntryManager<MyEntry>) to avoid such excessive casts. In addition, I noticed that osu! also casts classes like [this](https://ptb.discord.com/channels/188630481301012481/589331078574112768/1216754528079319070).

Ultimately, I settled on using CRTP as a trick to accomplish this. However, due to limitations of C# features, some functions may not be implemented perfectly. And for the sake of compatibility, I still provide a non-generic version to avoid some break-changes.

Todo:

- [ ] Rewrite XML comments of these classes.

### DIFF
--- a/osu.Framework/Graphics/Performance/LifetimeEntry.cs
+++ b/osu.Framework/Graphics/Performance/LifetimeEntry.cs
@@ -7,13 +7,17 @@ using System;
 
 namespace osu.Framework.Graphics.Performance
 {
+    public class LifetimeEntry : LifetimeEntry<LifetimeEntry>
+    {
+    }
+
     /// <summary>
     /// An object for a <see cref="LifetimeEntryManager"/> to consume, which provides a <see cref="LifetimeStart"/> and <see cref="LifetimeEnd"/>.
     /// </summary>
     /// <remarks>
     /// Management of the object which the <see cref="LifetimeEntry"/> refers to is left up to the consumer.
     /// </remarks>
-    public class LifetimeEntry
+    public abstract class LifetimeEntry<T> where T : LifetimeEntry<T>
     {
         private double lifetimeStart = double.MinValue;
 
@@ -42,12 +46,12 @@ namespace osu.Framework.Graphics.Performance
         /// Invoked before <see cref="LifetimeStart"/> or <see cref="LifetimeEnd"/> changes.
         /// It is used because <see cref="LifetimeChanged"/> cannot be used to ensure comparator stability.
         /// </summary>
-        internal event Action<LifetimeEntry> RequestLifetimeUpdate;
+        internal event Action<T> RequestLifetimeUpdate;
 
         /// <summary>
         /// Invoked after <see cref="LifetimeStart"/> or <see cref="LifetimeEnd"/> changes.
         /// </summary>
-        public event Action<LifetimeEntry> LifetimeChanged;
+        public event Action<T> LifetimeChanged;
 
         /// <summary>
         /// Update <see cref="LifetimeStart"/> of this <see cref="LifetimeEntry"/>.
@@ -74,12 +78,12 @@ namespace osu.Framework.Graphics.Performance
         /// <param name="end">The new <see cref="LifetimeEnd"/> value.</param>
         protected void SetLifetime(double start, double end)
         {
-            RequestLifetimeUpdate?.Invoke(this);
+            RequestLifetimeUpdate?.Invoke((T)this);
 
             lifetimeStart = start;
             lifetimeEnd = Math.Max(start, end); // Negative intervals are undesired.
 
-            LifetimeChanged?.Invoke(this);
+            LifetimeChanged?.Invoke((T)this);
         }
 
         /// <summary>
--- a/osu.Framework/Graphics/Performance/LifetimeEntry.cs
+++ b/osu.Framework/Graphics/Performance/LifetimeEntry.cs
@@ -1,8 +1,6 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-#nullable disable
-
 using System;
 
 namespace osu.Framework.Graphics.Performance
@@ -46,12 +44,12 @@ namespace osu.Framework.Graphics.Performance
         /// Invoked before <see cref="LifetimeStart"/> or <see cref="LifetimeEnd"/> changes.
         /// It is used because <see cref="LifetimeChanged"/> cannot be used to ensure comparator stability.
         /// </summary>
-        internal event Action<T> RequestLifetimeUpdate;
+        internal event Action<T>? RequestLifetimeUpdate;
 
         /// <summary>
         /// Invoked after <see cref="LifetimeStart"/> or <see cref="LifetimeEnd"/> changes.
         /// </summary>
-        public event Action<T> LifetimeChanged;
+        public event Action<T>? LifetimeChanged;
 
         /// <summary>
         /// Update <see cref="LifetimeStart"/> of this <see cref="LifetimeEntry"/>.
--- a/osu.Framework/Graphics/Performance/LifetimeEntryManager.cs
+++ b/osu.Framework/Graphics/Performance/LifetimeEntryManager.cs
@@ -13,6 +13,10 @@ using osu.Framework.Statistics;
 
 namespace osu.Framework.Graphics.Performance
 {
+    public class LifetimeEntryManager : LifetimeEntryManager<LifetimeEntry>
+    {
+    }
+
     /// <summary>
     /// Provides time-optimised updates for lifetime change notifications.
     /// This is used in specialised <see cref="CompositeDrawable"/>s to optimise lifetime changes (see: <see cref="LifetimeManagementContainer"/>).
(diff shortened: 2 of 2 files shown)


Frederisk:
Another question is whether we need to replace `LifetimeEntryManager` in some existing classes with generic versions? For example, `LifetimeManagementContainer` or `PooledDrawableWithLifetimeContainer` in osu!.

bdach:
This PR should include a preview of what the required game-side changes are after merging this.

Frederisk:
> Performance is not a factor here. The "gains" above are negligible at best, if not straight up just measurement error.

Yeah, just making sure I don't run into new performance issues.

> This PR should include a preview of what the required game-side changes are after merging this.

In theory, I provide a non-generic version, so existing code doesn't need any modifications and still works as usual. In other words, it is feasible for other content to remain unchanged after the PR is merged.

And if modification is needed, for classes such as `LifetimeManagementContainer`, we need:

```diff
- private readonly LifetimeEntryManager manager = new();
+ private readonly LifetimeEntryManager<DrawableLifetimeEntry > manager = new();
...
- private class DrawableLifetimeEntry : LifetimeEntry, IDisposable {
+ private class DrawableLifetimeEntry : LifetimeEntry<DrawableLifetimeEntry>, IDisposable {
...
}
...
```

and remove all redundant type conversions.

The process involving PooledDrawableWithLifetimeContainer, another layer of generics, is generally similar, except:

```diff
public abstract partial class PooledDrawableWithLifetimeContainer<TEntry, TDrawable> : CompositeDrawable
- where TEntry : LifetimeEntry
+ where TEntry : LifetimeEntry<TEntry>
 where TDrawable : Drawable {
...
}
```

bdach:
I would expect to see a preview of how converting the game-side usages to use the generics would look like, to evaluate if this implementation is the code quality improvement it claims to be.

API surface changes must always be judged by the usage code.

Frederisk:
I don't quite understand how I should provide a preview, do I need to open another PR in the repo of osu!?

bdach:
A link to an un-PR'd branch or a patch would also do.

Frederisk:
I tried to modify `PooledDrawableWithLifetimeContainer` in osu! to the expected generic type as [an example](https://github.com/Frederisk/osu/commit/e39d923d025155f3e5c6d654e07153247dbe5023). Finding other classes that need to be modified (including those in o!f and osu!) and modifying them will take me a lot of time, so it may take a while before I propose a proper complete PR.

osu.Framework/Graphics/Performance/LifetimeEntry.cs
```diff
@@ -1,24 +1,30 @@
...
-    public class LifetimeEntry
+    public class LifetimeEntry : LifetimeEntry<LifetimeEntry>
+    {
+    }
+
+    /// <summary>
+    /// The required base type for the <see cref="LifetimeEntryManager{T}"/> to consume, which provides a <see cref="LifetimeEntry{T}.LifetimeStart"/> and <see cref="LifetimeEntry{T}.LifetimeEnd"/>.
+    /// </summary>
+    /// <typeparam name="TDerived">The implemented class itself. Used to provide derived category information for base categories using the Curiously Recurring Template Pattern(CRTP).</typeparam>
+    public abstract class LifetimeEntry<TDerived> where TDerived : LifetimeEntry<TDerived>
```
bdach, line 22: Nooooooooooooooooooooooooope. Nope nope nope nope. I'm not having anything that the cpp people invented. Especially not with a name like "curiously recurring". I don't even follow what this is supposed to be doing.

I will not accept this sort of arcane construction here. If this makes-or-breaks the entire series then it might as well be closed right now.

osu.Framework/Graphics/Performance/LifetimeEntry.cs
```diff
@@ -68,27 +74,44 @@ protected virtual void SetLifetimeEnd(double end)
...
+            // public class NewEntry<T> : LifetimeEntry<T> where T : LifetimeEntry<T> { }
+            // NewEntry<MyEntry> a = new();
+            // ```
+            // Happily, however, the compiler correctly prevents code like this from compiling. This is also enough to deter users from writing incorrect code:
+            // ```csharp
+            // public class ErrorEntry : NewEntry<MyEntry> { }
+            // LifetimeEntryManager<NewEntry<MyEntry>> error1 = new(); // Compiler error.
+            // LifetimeEntryManager<ErrorEntry> error2 = new(); // Compiler error.
+            // ```
+            RequestLifetimeUpdate?.Invoke((TDerived)this);
```
bdach, line 100: Case in point for the above diatribe. If API correctness of a method requires _this_ sort of preamble to convince readers that the API surface is safe, the whole thing is overdone.

bdach (changes requested):

osu.Framework/Graphics/Performance/LifetimeEntry.cs
Frederisk: > I will not accept this sort of arcane construction here.

Sadly, this is necessary. Manager needs to use this type argument to get the actual type of the field.

I even thought this was a fairly common way of writing [modern C#](https://blog.stephencleary.com/2022/09/modern-csharp-techniques-1-curiously-recurring-generic-pattern.html). I just want to clarify here what the real name of this mode is.😕

osu.Framework/Graphics/Performance/LifetimeEntry.cs
Frederisk: Basically, it's the same thing, I'm just trying to illustrate why we still can't get away from typecasting. I simply think that a more detailed explanation will help myself or other users understand the design logic clearly.

These comments are not actually necessary, since I don't really need any additional constraints or commitments. These are written within the design scope of the compiler and I'm not trying to break anything.

I tried to clarify my thinking through comments, but now it seems like this is overkill?

osu.Framework/Graphics/Performance/LifetimeEntry.cs
bdach: I'm honestly having a difficult time accepting that linked article if it's giving `IEquatable<T>` as an example of this. `IEquatable<T>` is nowhere near as difficult to parse! It's not doing this `IEquatable<T> where T : IEquatable<T>` brain twister thing in any class signature. Its methods accept `T`.

osu.Framework/Graphics/Performance/LifetimeEntry.cs
bdach: I'm not even sure what that is attempting to explain. Why is the first snippet "unwanted"? It looks like perfectly reasonable code. What is the second snippet supposed to be? What are `NewEntry` and `MyEntry`?

osu.Framework/Graphics/Performance/LifetimeEntry.cs
Frederisk: I just spent some time searching for something I myself learned about this many years ago. Luckily, I found [it](https://zpbappi.com/curiously-recurring-template-pattern-in-csharp/) and I thought this article might be a good explanation.

osu.Framework/Graphics/Performance/LifetimeEntry.cs
Frederisk: > I'm not even sure what that is attempting to explain. Why is the first snippet "unwanted"?

Because if the user writes code like that, the cast here will throw an exception.

```cs
LifetimeEntry<MyEntry>
// TDerived is "MyEntry" here.
// this.GetType() == LifetimeEntry
...
(MyEntry)this // throw Exception A base type cannot be cast to a derived type.
```

Frederisk: However, this subtle error will be exposed through the following compilation errors when writing. I just want to make this clear. After all, this recursive form of generics can be a bit confusing to understand how the compiler will handle it.

Frederisk: > What is the second snippet supposed to be? What are `NewEntry` and `MyEntry`?

Assumptions about an extreme situation. Although I declared this class as abstract class. But like all abstract classes there are problems that may be encountered. You can't prevent users from implementing an abstract class without taking any meaningful action. This is basically like:

```cs
abstract class A {
  // some thing here.
}
class B : A { } // class B is the same as A, but not it's no longer abstract!
```

This makes it possible for users to write illogical code.

Then in the third code block, I actually illustrate that even if this happens, the compiler will correctly prevent the user from doing this.

Frederisk:
I understand that CRTP has caused confusion for someone, so please allow me to provide some clarification here:

What is CRTP: CRTP is a trick but not a hack. It is used here to pass the `Entry` type information to the `Manager`. CRTP is not unique to C/C++, it can also be implemented correctly in many languages ​​​​that provide generics, such as Java, Kotlin, and C#. CRTP is just an idiom here. It's unquestionably legal in the C# language because obviously we don't need any compiler magic and it won't produce any compiler warnings or errors.

Why CRTP: When implementing a generic `Manager` in a normal way, you will encounter the following problems (I'll explain using a simplified model):

```cs
public class Manager<T> where T : Entry {
  protected T? item;
  public void Foo(T entry) {
    entry.Do += (sender) => {
      this.item = sender;
    }
  }
}

public class Entry {
  public event Action<Entry>? Do;
  public void Bar() {
    Do?.Invoke(this);
  }
}
```

Then obviously, `sender` cannot be safely converted to `T`. And on the other hand, even if I implemented `MyEntry`, it would still be of the wrong type for `Do` and would require us to use casts to maintain it unsafely.

```cs
public class MyEntry : Entry { }
...
MyEntry.Do.GetType == typeof(Action<Entry>); // But we prefer it to be Action<MyEntry>
```

Therefore, we need a type parameter to pass the type of the current class to the `Entry` class's own field and the `Manager` that uses this class:

```cs
public class Manager<T> where T : EntryBase<T> {
    protected T? item;
    public void Foo(T entry) {
        entry.Do += (sender) => {
            this.item = sender;
        };
    }
}

public abstract class EntryBase<T> where T : EntryBase<T> {
    public event Action<T>? Do;
    public void Bar() {
        Do?.Invoke((T)this);
    }
}

public class Entry : EntryBase<Entry> { }
```

Now, everything works fine, our `Do` can have the correct type, and the `sender` no longer needs any casts.

Of course, we can notice that there is a type cast in `Do?.Invoke`. But this is safe, and I stated in the comments that the compiler properly prevents users from breaking the safety of the cast.

Maybe you are uncomfortable with such lengthy comments, because it seems to imply that I am using comments to maintain the normal operation of a class. However, what I need to explain is that those comments are just for those who want to understand the design details. After all, I have to admit that many higher-level techniques are used internally in this class, and not everyone can easily understand them. The intent within the code.

For users of this class, they only need to follow [this rule](https://github.com/ppy/osu-framework/pull/6212/files):

```
public class MyEntry : EntryBase<MyEntry> {
  // something here
}
...
var manager = new Manager<MyEntry>;
```

And the compiler will correctly block incorrect code.

BTW, English is not my first language, so if there is any ambiguity in my explanation please point it out and I will rephrase it.

Frederisk:
But the comments above just now reminded me that maybe I should give `LifetimeEntry<TDerived>` a new name, such as `LifetimeEntryBase<TDerived>`, to make the code clearer.🤔

bdach:
Let me be very clear on this:

- This PR series wants to remove some casting when using `LifetimeEntryManager`. Generally, a cast takes five seconds to parse mentally.
- What is offered instead is a generic-based "trick" which *invariably* takes more than five seconds to parse mentally, and which requires implementors to know about it and understand how it works.

I do not find that an acceptable tradeoff at this time regardless of how many words of explanation are thrown at me. I am not interested in reviewing this series any further. If anyone else is they are free to do so but I will not participate in further discussions on this matter.

Frederisk:
Well, I respect your opinion. I regret that we did not reach a consensus.🥲 But in fact, users of these two classes don't even need to understand how they work. They only need to inherit from the generic class. So I don't think this invariably requires users to spend time mentally parsing it.

Frederisk:
As a fact, CRTP is a trick that is already used in .NET.

For example: [IAdditionOperators](https://learn.microsoft.com/en-us/dotnet/api/system.numerics.iadditionoperators-3?view=net-8.0):

```cs
public interface IAdditionOperators<TSelf,TOther,TResult> where TSelf : IAdditionOperators<TSelf,TOther,TResult>
```

Issue #6202, opened 2024-03-02
Periodic AccessViolation somewhere in BASS


smoogipoo:
Reported in https://github.com/ppy/osu/issues/27466

It doesn't happen all the time... I've also been able to reproduce this every once in a blue moon on linux, especially with a debugger attached. I vaguely remember resolving an issue related to this back when mixers were originally implemented, but can't remember the specifics (https://github.com/ppy/osu-framework/pull/4912).

```
Application: osu!.exe
CoreCLR Version: 8.0.224.6711
.NET Version: 8.0.2
Description: The process was terminated due to an unhandled exception.
Exception Info: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
Stack:
at ManagedBass.Bass.FXSetPriority(Int32, Int32)
at ManagedBass.Bass.FXSetPriority(Int32, Int32)
at osu.Framework.Audio.Mixing.Bass.BassAudioMixer+<>c__DisplayClass29_0.<onEffectsChanged>g__applyEffects|2(Int32, Int32)
at osu.Framework.Audio.Mixing.Bass.BassAudioMixer+<>c__DisplayClass29_0.<onEffectsChanged>b__0()
at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
at System.Threading.Tasks.Task.ExecuteWithThreadLocal(System.Threading.Tasks.Task ByRef, System.Threading.Thread)
at System.Threading.Tasks.ThreadPoolTaskScheduler.TryExecuteTaskInline(System.Threading.Tasks.Task, Boolean)
at System.Threading.Tasks.Task.InternalRunSynchronously(System.Threading.Tasks.TaskScheduler, Boolean)
at osu.Framework.Audio.AudioComponent.Update()
at osu.Framework.Audio.AudioCollectionManager`1[[System.__Canon, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].UpdateChildren()
at osu.Framework.Audio.AudioComponent.Update()
at osu.Framework.Threading.AudioThread.onNewFrame()
at osu.Framework.Threading.GameThread.processFrame()
at osu.Framework.Threading.GameThread.<createThread>g__runWork|66_0()
at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
```

PR #6214, opened 2024-03-13
Remove effect prioritisation to fix segfaults

smoogipoo:
- Closes https://github.com/ppy/osu-framework/issues/6202

I intend for this to be a somewhat temporary change because this entire thing needs to be rewritten... As far as I can tell we're not relying on effect priorities in osu! (cc @nekodex ).

We've had issues with `FXSetPriority` segfaulting the test runner in the past, but it has not been reported to be doing the same in-game: https://github.com/ppy/osu/discussions/27338

I haven't been able to repro this in osu!framework, but, under pipewire the game should segfault when entering `PlayerLoader` with the following diff applied:
```diff
diff --git a/osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs b/osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
index 2605a84008..1ff9a69fbc 100644
--- a/osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
+++ b/osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
@@ -14,6 +14,7 @@
 using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Extensions.ObjectExtensions;
 using osu.Framework.Statistics;
+using osu.Framework.Utils;

 namespace osu.Framework.Audio.Mixing.Bass
 {
@@ -272,6 +273,14 @@ protected override void UpdateState()
                 removeChannelFromBassMix(channel);
             }

+            foreach (var effect in ActiveEffects)
+            {
+                if (effect.Handle == 0)
+                    continue;
+
+                ManagedBass.Bass.FXSetPriority(effect.Handle, RNG.Next(0, 10));
+            }
+
             FrameStatistics.Add(StatisticsCounterType.MixChannels, activeChannels.Count);
             base.UpdateState();
         }
```

I considered the alternative of instead removing and re-adding the effects. This causes a short but noticeable glitch when player load finishes.

What's the cause of all of this? I don't know. Likely some sort of x86-register dependent behaviour in BASS (wouldn't be the first time). It's a bit too deep for the time being.

### DIFF
--- a/osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
+++ b/osu.Framework/Audio/Mixing/Bass/BassAudioMixer.cs
@@ -10,7 +10,6 @@ using System.Runtime.InteropServices;
 using ManagedBass;
 using ManagedBass.Mix;
 using osu.Framework.Bindables;
-using osu.Framework.Development;
 using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Extensions.ObjectExtensions;
 using osu.Framework.Statistics;
@@ -421,13 +420,7 @@ namespace osu.Framework.Audio.Mixing.Bass
                     // Effects with greatest priority are stored at the front of the list.
                     effect.Priority = -i;
 
-                    if (effect.Handle != 0)
-                    {
-                        // Todo: Temporary bypass to attempt to fix failing test runs.
-                        if (!DebugUtils.IsNUnitRunning)
-                            ManagedBass.Bass.FXSetPriority(effect.Handle, effect.Priority);
-                    }
-                    else
+                    if (effect.Handle == 0)
                         effect.Handle = ManagedBass.Bass.ChannelSetFX(Handle, effect.Effect.FXType, effect.Priority);
 
                     ManagedBass.Bass.FXSetParameters(effect.Handle, effect.Effect);


bdach:
Incidentally... https://github.com/ppy/osu/issues/27466#issuecomment-1974845078

That's windows too by the way.

bdach (approved):

nekodex:
Hmm, while I don't believe we're using effect priorities in osu! (yet), I should mention that this change may result in unexpected behaviour whenever the order of effects are changed on a mixer.

Not all audio effects are transitive (e.g. compressor->reverb does not result in the same output as reverb->compressor), so if someone using framework reorders effects expecting the output to change and it doesn't (due to this PR), it may lead to some confusion.

I almost think the alternative of a "short glitch" would be preferable, as at least the resulting behaviour (effects processing in the correct order) would be as expected from a user's perspective.

Ultimately this is probably an edge-case that likely no-one will hit, but I could see it potentially happening for a DAW/DJ application for instance, or as a stretch, a game with dynamic effects applying and mutating realtime in response to... stuff.

smoogipoo:
For now I'm banking on mixers being unused by anyone other than us, so the important part for me at this point is whether that will have an effect on our existing usages of o!f. Effect prioritisation is pretty awkward right now, along with the whole replace-effect-at-index-to-update thing...

(not to mention in going through the code I found potential out-of-range indexing 🙈, it's all so bad)

I'm also considering that we're still lacking nested mixers. With nested mixers, wouldn't effect prioritisation either become a non-issue or become an explicit detail for special cases like you mentioned (due to performance, if anything)?

nekodex:
Yeah I agree, the effect handling in framework is a bit... not great... at the moment.

For the time being, removing `FXSetPriority` is probably fine because:
- We're not even setting `effect.Priority` anywhere yet, let alone changing it.
- We're only using low-pass/high-pass filters in osu! currently, and filters are effectively transitive meaning we're not gonna notice any issues if they're run out of order anyway. That may change in the future though if/when we start using other effects.

I'm not sure I see how nested mixers alleviates anything though? With each mixer having its own list of effects, we're just nesting the problem at that point? 😅

We could probably workaround this in the future by setting priority at effect creation time and then never changing it. Likely that'll be enough for our needs anyway.

Regardless, surely the correct solution here is to get the segfault in `BASS_FXSetPriority` fixed rather than trying to work around it? ...funnily enough, I came across this surprisingly relevant thread while briefly digging into this 😜 https://www.un4seen.com/forum/?topic=19592.0

smoogipoo:
Hmm, indeed it it would be nesting the problem... For that particular case I was imagining a structure like this:

```
Mixer
{
    Effects = { Compressor },
    Children =
    {
        Mixer
        {
            Effects = { Reverb },
            Children =
            {
                /* Tracks, samples, etc */
            }
        }
    }
}
```

Do you imagine changing priority to be a more common operation, or for it to only be global like my example there?

nekodex:
Honestly? I'm not sure we'll need effect priority changing for osu! itself at all. I imagine we'd be defining effect priority once (e.g. in a `BackgroundDependencyLoader` method) and then leaving it at that.

I can't really think of any scenarios for osu! off the top of my head where we'd need to alter effect priorities after the fact.

PR #6218, opened 2024-03-17
Add event handler for user change of `TabControl` tab selection

frenzibyte:
- See https://github.com/ppy/osu/pull/27448 for previous discussion

The abstraction may look a little bit too much, but I've been trying to find a middle-ground between code readability and simplicity of structure. 

The previously existing `SelectTab` method is supposed to do all that's necessary for the tab selection to be changed, without necessarily being related to user action, so I've renamed it to `UpdateTabSelection`.

I've also introduced two methods, `SelectTab` and `SelectItem`, the former is for internal use by `TabItem` click event to update tab selection and trigger user event, meanwhile the latter is for external use by calling it with the target item to update tab selection and trigger user event as well (e.g. osu! calling the method when changing editor screen mode, so that an auxiliary sample is played).

I didn't want `SelectTab` to exist in the first place, but it's necessary so that the `TabItem` click event doesn't have to call `SelectItem` and perform an unnecessary lookup for an already known tab item, so that's why it's there.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs
@@ -376,11 +376,11 @@ namespace osu.Framework.Graphics.UserInterface
 
         private bool selectTab(TabItem<T> tab)
         {
-            var lastTab = SelectedTab;
+            if (tab == SelectedTab)
+                return false;
 
-            updateSelectedTab(tab);
-            Current.Value = SelectedTab != null ? SelectedTab.Value : default;
-            return SelectedTab != lastTab;
+            Current.Value = tab != null ? tab.Value : default;
+            return true;
         }
 
         private void updateSelectedTab(TabItem<T> tab)
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs
@@ -384,15 +384,6 @@ namespace osu.Framework.Graphics.UserInterface
                 OnUserTabSelectionChanged(tab);
         }
 
-        private bool selectTab(TabItem<T> tab)
-        {
-            if (tab == SelectedTab)
-                return false;
-
-            Current.Value = tab != null ? tab.Value : default;
-            return true;
-        }
-
         /// <summary>
         /// Switches the currently selected tab forward or backward one index, optionally wrapping.
         /// </summary>
@@ -421,6 +412,15 @@ namespace osu.Framework.Graphics.UserInterface
                 found = allTabs.FirstOrDefault(t => t != SelectedTab);
 
             return found != null && selectTab(found);
+        }
+
+        private bool selectTab(TabItem<T> tab)
+        {
+            if (tab == SelectedTab)
+                return false;
+
+            Current.Value = tab != null ? tab.Value : default;
+            return true;
         }
 
         private void activationRequested(TabItem<T> tab)
--- a/osu.Framework/Graphics/UserInterface/TabControl.cs
+++ b/osu.Framework/Graphics/UserInterface/TabControl.cs
@@ -208,10 +208,22 @@ namespace osu.Framework.Graphics.UserInterface
 
             Current.BindValueChanged(v =>
             {
-                if (v.NewValue != null && tabMap.TryGetValue(v.NewValue, out var found))
-                    updateSelectedTab(found);
-                else
-                    updateSelectedTab(null);
+                TabItem<T> tab = null;
+
+                if (v.NewValue != null)
+                    tabMap.TryGetValue(v.NewValue, out tab);
+
+                // Only reorder if not pinned and not showing
+                if (AutoSort && tab != null && !tab.IsPresent && !tab.Pinned)
+                    performTabSort(tab);
+
+                // Deactivate previously selected tab
+                if (SelectedTab != null && SelectedTab != tab) SelectedTab.Active.Value = false;
+
+                SelectedTab = tab;
+
+                if (SelectedTab != null)
+                    SelectedTab.Active.Value = true;
             }, true);
 
             // TabContainer doesn't have valid layout yet, so TabItems all have y=0 and selectTab() didn't call performTabSort() so we call it here instead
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -378,10 +403,13 @@ private void selectTab(TabItem<T> tab)
         /// </summary>
         /// <param name="direction">Pass 1 to move to the next tab, or -1 to move to the previous tab.</param>
         /// <param name="wrap">If <c>true</c>, moving past the start or the end of the tab list will wrap to the opposite end.</param>
-        public virtual void SwitchTab(int direction, bool wrap = true)
+        /// <returns>Whether tab selection has changed as a result of this call.</returns>
+        public virtual bool SwitchTab(int direction, bool wrap = true)
```
peppy, line 407: So this method doesn't signal a user selection but `SelectItem` does? How does that make sense naming / exposure wise?

frenzibyte: I did not realise this method is public, thought it was protected. We can signal the event from here and define a private version that's used by the "switch tab on remove" case.

The reason why I don't want the sound to play on the "switch tab on remove" scenario because there would already be a sound played when the user removes the tab (e.g. a click sound on the "close" button).

osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -348,17 +351,42 @@ private void updateDropdown(TabItem<T> tab, bool isVisible)
...
+                throw new InvalidOperationException($"Item {item} cannot be selected as it does not exist in this {nameof(TabControl<T>)}");
+
+            SelectTab(tab);
+        }
+
+        /// <summary>
+        /// Selects a <see cref="TabItem{T}"/> and signals an event that the user selected the given tab via <see cref="OnUserTabSelectionChanged"/>.
+        /// </summary>
+        /// <param name="tab">The tab to select.</param>
+        protected void SelectTab(TabItem<T> tab)
```
peppy, line 370: We now have

`SelectTab`
`UpdateTabSelection`
`updateSelectedTab`

which makes me feel uneasy.

in addition, `UpdateTabSelection` is `virtual` for a single usage osu-side which is intended to block actual value changes. Maybe we should look to remove that usage and make it non-virtual to simplify things at this end? It feels very flaky.

frenzibyte: Does this look fine as an alternative of the override?
```diff
diff --git a/osu.Game/Graphics/UserInterface/ScreenBreadcrumbControl.cs b/osu.Game/Graphics/UserInterface/ScreenBreadcrumbControl.cs
index 80fb4fd2e4..54d20ed23b 100644
--- a/osu.Game/Graphics/UserInterface/ScreenBreadcrumbControl.cs
+++ b/osu.Game/Graphics/UserInterface/ScreenBreadcrumbControl.cs
@@ -15,6 +15,8 @@ namespace osu.Game.Graphics.UserInterface
     /// </summary>
     public partial class ScreenBreadcrumbControl : BreadcrumbControl<IScreen>
     {
+        private bool allowUpdatingCurrent;
+
         public ScreenBreadcrumbControl(ScreenStack stack)
         {
             stack.ScreenPushed += onPushed;
@@ -24,27 +26,41 @@ public ScreenBreadcrumbControl(ScreenStack stack)
                 onPushed(null, stack.CurrentScreen);
         }
 
-        protected override bool UpdateTabSelection(TabItem<IScreen> tab)
+        protected override void LoadComplete()
         {
-            // override base method to prevent current item from being changed on click.
-            // depend on screen push/exit to change current item instead.
-            var lastTab = SelectedTab;
-            tab.Value.MakeCurrent();
-            return tab != lastTab;
+            base.LoadComplete();
+
+            Current.BindValueChanged(c =>
+            {
+                if (!allowUpdatingCurrent)
+                {
+                    // prevent current item from being changed on click.
+                    // depend on screen push/exit to change current item instead.
+                    updateCurrentValue(c.OldValue);
+                    c.NewValue.MakeCurrent();
+                }
+            });
         }
 
         private void onPushed(IScreen lastScreen, IScreen newScreen)
         {
             AddItem(newScreen);
-            Current.Value = newScreen;
+            updateCurrentValue(newScreen);
         }
 
         private void onExited(IScreen lastScreen, IScreen newScreen)
         {
             if (newScreen != null)
-                Current.Value = newScreen;
+                updateCurrentValue(newScreen);
 
             Items.ToList().SkipWhile(s => s != Current.Value).Skip(1).ForEach(RemoveItem);
         }
+
+        private void updateCurrentValue(IScreen screen)
+        {
+            allowUpdatingCurrent = true;
+            Current.Value = screen;
+            allowUpdatingCurrent = false;
+        }
     }
 }
```

It works in practice but want your confirmation before proceeding with it.

frenzibyte: As for the existence of multiple methods, we can rename `UpdateTabSelection` to `selectTab` once we get rid of the usage above. `updateSelectedTab` still has to exist as it is used by some places that intentionally don't want current bindable to be updated.

peppy: > Does this look fine as an alternative of the override?

No I really don't like this. How about [this](https://github.com/ppy/osu/pull/27675) instead?

frenzibyte: Might as well.

osu.Framework/Graphics/UserInterface/TabControl.cs
```diff
@@ -349,16 +352,37 @@ private void updateDropdown(TabItem<T> tab, bool isVisible)
         }
 
         /// <summary>
-        /// Selects a <see cref="TabItem{T}"/>.
+        /// Selects a <see cref="TabItem{T}"/> and signals an event that the user selected the given tab value via <see cref="OnUserTabSelectionChanged"/>.
```
peppy, line 355: I dunno about all these public events  triggering the `OnUserTabSelectionChanged` method, but let's see how it works.

peppy (approved):

peppy:
I've made further simplifications to convoluted logic. Requesting a second review.

bdach (approved):
seems fine

PR #6219, opened 2024-03-18
Reduce scope of experimental renderer options

smoogipoo:
On a second look, I think I want to reduce the scope of this.

Windows: I want to push people towards D3D11 or Vulkan, even if OpenGL Experimental behaves better than OpenGL.
macOS: OpenGL should practically never be used.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -847,7 +847,6 @@ namespace osu.Framework.Platform
                     yield return RendererType.Direct3D11;
                     yield return RendererType.Deferred_Direct3D11;
                     yield return RendererType.OpenGL;
-                    yield return RendererType.Deferred_OpenGL;
                     yield return RendererType.Deferred_Vulkan;
 
                     break;
@@ -863,7 +862,6 @@ namespace osu.Framework.Platform
                     yield return RendererType.Metal;
                     yield return RendererType.Deferred_Metal;
                     yield return RendererType.OpenGL;
-                    yield return RendererType.Deferred_OpenGL;
 
                     break;
 


peppy:
Hmm, my only concern would be windows, where there's a push from users to recommend people try opengl.

Would the next step be to move d3d to the deferred renderer for windows testing as a follow-up, rather than potentially moving focus to the opengl one?

smoogipoo:
> push from users to recommend people try opengl

Separate from our own recommendations? I'm pushing towards D3D11 because it's more widely supported than OpenGL. I drafted up a discussion topic (currently locked) showcasing the sort of questions I want to ask: https://github.com/ppy/osu/discussions/27659

In particular, I think the relevant one is:

> I'm using Windows and all options behave poorly for me.

To which I'd follow up for that user specifically to try out `deferred/opengl` via environment variable.

> Would the next step be to implement/move d3d to the deferred renderer for windows testing as a follow-up, rather than potentially moving focus to the opengl one?

I'm not sure what you're asking here...

peppy:
> Separate from our own recommendations?

Yes, there are youtube videos, changelog comments and discussion resolutions where people recommend switching from d3d to opengl as a solution to various performance issues on lazer, with many people getting positive results. I've been turning a blind eye to this because if it works for them then all the better.

Ignore the question regarding d3d, I didn't realise there's already an experimental option for that, which is what I was asking about (I am not up to scratch on the renderer PRs as there's too many other things happening).

peppy (approved):
fine i think.

bdach:
> Separate from our own recommendations?

Well... about that...

People are complaining about d3d11 borderless being vsync-locked regardless of their frame limiter. This happens because `WaitForNextFrameReady()` does do something like that. To which my response has been: either go exclusive fullscreen (thus enabling tearable swapchain), or switch to another renderer. And because vulkan has been considered broken (and still is, to a degree, see [here](https://discord.com/channels/188630481301012481/589331078574112768/1216692908128735242)). And people do want borderless because tabbing out is faster.

So I dunno. Make of that what you will.

smoogipoo:
> vulkan has been considered broken (and still is, to a degree, see [here](https://discord.com/channels/188630481301012481/589331078574112768/1216692908128735242))

See my response to that in the following conversion - it's not a Vulkan issue. The deferred renderer should have fixed Vulkan otherwise (it bricking the game when entering gameplay). Have I missed something else?

> Sometimes opengl also helps with the inscrutable "0xc000005 access violation exception" hard windows crashes that keep getting reported

OpenGL (i.e. legacy) is still available. Just not deferred/OpenGL.

> d3d11 borderless being vsync-locked regardless of their frame limiter

Which wouldn't be an issue except for:
- Bigger numbers = better (not true in all cases).
- If being capped is causing framerate to drop below vsync or be unstable in situations where the GPU can handle well above vsync with no problem. This is likely caused by inter-frame dependencies and is definitely one of the things I want feedback on - the deferred renderer may already improve this.

bdach:
> See my response to that in the conversation following - it's not a Vulkan issue. The deferred renderer should have fixed Vulkan otherwise (it bricking the game when entering gameplay). Have I missed something else?

I'll admit I haven't exactly debugged but in profiling it appeared there was a vulkan-specific issue with constant allocation increases on the... buffering? masking? performance test scene that neither opengl nor d3d11 did not exhibit.

Issue #1966, opened 2018-10-31
PassThroughInputManager syncs key presses even if they were handled


smoogipoo:
In `PassThroughInputManager.Update()` it syncs the state from the parenting input manager.

For mouse input, only released buttons are handled.
For keyboard input, pressed and released buttons are handled.

This has the effect that if a `Drawable` handles `KeyDown()` at a higher point in the hierarchy, the button press will _also_ be handled by every `PassThroughInputManager` in the draw hierarchy when they sync.

Test case:
```csharp
public class TestCaseScratch : TestCase
{
    public TestCaseScratch()
    {
        Children = new Drawable[]
        {
            new PassThroughInputManager
            {
                RelativeSizeAxes = Axes.Both,
                UseParentInput = true,
                Child = new InputBox { RelativeSizeAxes = Axes.Both }
            },
            new HiddenInputHandler { RelativeSizeAxes = Axes.Both }
        };
    }

    private class InputBox : Box
    {
        protected override bool OnKeyDown(KeyDownEvent e)
        {
            Colour = Color4.Red;
            return true;
        }

        protected override bool OnKeyUp(KeyUpEvent e)
        {
            Colour = Color4.White;
            return true;
        }
    }

    private class HiddenInputHandler : Box
    {
        public HiddenInputHandler()
        {
            Alpha = 0;
            AlwaysPresent = true;
        }

        protected override bool Handle(UIEvent e) => true;
    }
}
```

UselessToucan:
I have a couple of questions.
Lets say we have the scene graph.
![image](https://user-images.githubusercontent.com/22450664/48051763-0a210100-e1b7-11e8-9797-f5e6bbad4060.png)

1. Do I understand correctly that `PassThroughInputManager` should sync input state changes related to `Container B` and `Container C`(including their children)?
2. Do I understand correctly that `PassThroughInputManager` should ignore input state changes out of its scope? For example, input state changes related to `Container A`.
3. Do I understand correctly that the relationship between `PassThroughInputManager` and parent `InputManager` is many to one? I mean, we can have multiple `PassThroughInputManager` which will sync with one parent `InputManager`.

smoogipoo:
1. Not sure what you mean by this. Syncing input state changes related to its own children?
2. Container B/C should probably still see that key A is down even if Container A handled it.
3. Yes

Issue #1672, opened 2018-06-29
PassThroughInputManager needs to receive initial state from parent


smoogipoo:
In the case that `UseParentInput` is set to `true`, it may not apply the full parent state when handling input.

E.g.

1. Enter a mania beatmap without auto.
2. Hold ctrl while in `PlayerLoader` just before gameplay.
3. Attempt to press Ctrl+(-/+)

Expected: The scroll speed should change.
Actual: Because modifier keys specifically are not propagated on repeat, the mania input manager will handle a -/+ keypress without `state.Keyboard.ControlPressed` being `true`.

PR #6221, opened 2024-03-19
Change `PassThroughInputManager` to not sync new presses outside `Handle()`

frenzibyte:
- Discussed in https://github.com/ppy/osu/issues/9658#issuecomment-2005767323, resolving one part of the issue.
- Closes https://github.com/ppy/osu-framework/issues/1966

When the game is paused, the game input manager has `UseParentInput` set to false for input to be frozen. When resuming gameplay in osu!, we ask the user to press a key at the location of the cursor first, then toggle back `UseParentInput` and resume gameplay accordingly.

If the user had paused without holding down any key, then when toggling back `UseParentInput`, either at the setter of `UseParentInput`, or at any subsequent update frames of the input manager, the key that was pressed to resume gameplay will be synchronised with the game input manager, causing gameplay to receive an incorrect press event and pollute the score for the player.

This PR introduces a flag, `SyncNewPresses`, which can be disabled to prevent the above from happening, by making `Sync()` not apply press input from parent state, and only look at releasing instead.


bdach:
I dunno. I threw the flag proposal out as a potential way of avoiding complications. But I look at this diff and see more complications.

I'd like this to be simpler. Isn't it an option to (spitballing) make `Sync()` virtual and override it to do nothing in the resume overlay case or something?

At the least the flag needs to be named different. Something like `TriggerInputEventsOnSync` or something. Ideally it should mean that, too.

frenzibyte:
I think you're misunderstanding the issue. The point of this change is to make the input manager that contains the hit objects not sync new presses using `Sync()`, so it doesn't catch the press by the resume overlay if we intentionally block it from receiving it by the `PassThroughInputManager.Handle(UIEvent)` path. In other words, we want the following to happen:
 - User presses osu! action to resume.
 - `ResumeOverlay` receives the action press, but, using some method that I haven't decided on yet, it will not let the `DrawableRuleset.KeyBindingInputManager` (hit objects input manager) receive the action press via the `Handle(UIEvent)` path
 - Then, since I want `DrawableRuleset.KeyBindingInputManager` to have "sync new presses" disabled, once it's enabled to receive parent input (set `UseParentInput` to true), it will not apply the user's press of the osu! action.

I suggest we do away with the flag and conform to the new behaviour introduced by this PR, and reassess when something bad happens.

smoogipoo:
I don't like the increase in complexity here. I'd be _tentatively_ okay with using the same logic as the mouse buttons, but I also have long forgotten why mouse buttons specifically differed.

frenzibyte:
The mouse button behaviour looks to have just been added as-is since https://github.com/ppy/osu-framework/pull/1682 with no specific reason.

frenzibyte:
I touched on that already in https://github.com/ppy/osu/issues/9658#issuecomment-2005767323 (see final point in bullet 2, where I linked the issue). But that being said, the diff will not change by much actually:
```diff
diff --git a/osu.Framework/Input/PassThroughInputManager.cs b/osu.Framework/Input/PassThroughInputManager.cs
index 40855fa26..e82e66ab6 100644
--- a/osu.Framework/Input/PassThroughInputManager.cs
+++ b/osu.Framework/Input/PassThroughInputManager.cs
@@ -48,15 +48,6 @@ public virtual bool UseParentInput
 
         private bool useParentInput = true;
 
-        /// <summary>
-        /// Whether to synchronise newly pressed buttons on <see cref="Sync"/>.
-        /// Pressed buttons are still synchronised at the point of loading the input manager.
-        /// </summary>
-        /// <remarks>
-        /// This does not apply for mouse buttons.
-        /// </remarks>
-        public bool SyncNewPresses { get; init; } = true;
-
         public override bool HandleHoverEvents => UseParentInput ? parentInputManager.HandleHoverEvents : base.HandleHoverEvents;
 
         internal override bool BuildNonPositionalInputQueue(List<Drawable> queue, bool allowBlocking = true)
@@ -251,7 +242,7 @@ protected virtual void SyncInputState(InputState state, bool applyPresses)
                 }
             }
 
-            if (SyncNewPresses || applyPresses)
+            if (applyPresses)
             {
                 // since we have a flag for syncing new presses, it is expected that we apply mouse button presses as well.
                 // this wasn't the case before the existence of the flag for another reason that's irrelevant here,
```

The `applyPresses` property will have to exist so that the input manager can apply new presses on `LoadComplete` to satisfy https://github.com/ppy/osu-framework/issues/1672.

bdach:
Well I dunno what to say at this point if _that_ is the extent of the changes incurred by the flag. You're basically presenting two choices: a schedule hack or increased complexity. At this point I'd be even willing to reconsider the schedule hack because at least it's self-contained to a degree I guess?

frenzibyte:
Two things need to be clarified:
 1. The proposal of scheduling resume operation to be performed one frame later *does not* replace this PR, because as soon as `UseParentInput` is set to true and the ruleset input manager sees the pressed key it will still register it as pressed. The intention of this PR is to make the input manager completely disregard the pressed key regardless of how long it is held.
 2. Disabling synchronisation on ruleset input manager after initial load does not help either, because we still want to synchronise buttons that were previously held but are now released (i.e. a user holding a slider but paused midway then resumed but didn't continue holding).

I've juggled and reordered the class around in hopes it reduces complexity. Take another look at it and see what you think. I have updated test coverage to reflect the new expected behaviour of the class, and tested on osu! to ensure it's working correctly.

frenzibyte:
I have removed the initial state synchronisation logic. Removing it does not regress the case brought in #1672 because scroll speed actions have been migrated to global key bindings, therefore not reading from the ruleset input manager to trigger the action. Also, if we compare this behavior on a normal `Drawable`, a drawable does not receive an `OnKeyDown` if it's added to the hierarchy after a key has been pressed, and when the key is up, it does not receive an `OnKeyUp` either (see [test](https://github.com/ppy/osu-framework/blob/9c2a7d35fee9061987e174c8b71da94e998e2367/osu.Framework.Tests/Input/KeyboardInputTest.cs#L21-L62)). 

Therefore, for the sake of reducing complexity and matching general behaviour of drawable input handling, I have removed it, and also confirmed all tests both on o!f and osu! are passing, meanwhile the PR still maintains its original intent, which is to resolve https://github.com/ppy/osu/issues/9658 (only partly).

osu.Framework/Input/PassThroughInputManager.cs
```diff
@@ -94,137 +98,125 @@ protected override bool Handle(UIEvent e)
...
+                new MidiKeyInput(key, parentState?.Midi?.Velocities.GetValueOrDefault(key) ?? 0, false).Apply(CurrentState, this);
+            foreach (var button in tabletPenDiff.Released)
+                new TabletPenButtonInput(button, false).Apply(CurrentState, this);
+            foreach (var button in tabletAuxiliaryDiff.Released)
+                new TabletAuxiliaryButtonInput(button, false).Apply(CurrentState, this);
+
+            if (parentInputManager == null)
+                return;
+
+            var parentState1 = parentInputManager.CurrentState;
```
smoogipoo, line 208: What's the point of this? I'm not really grasping why the existing `parentState` above can't be used.

osu.Framework/Input/PassThroughInputManager.cs
```diff
@@ -94,137 +98,125 @@ protected override bool Handle(UIEvent e)
...
+            return false;
         }
 
         /// <summary>
-        /// Sync current state to a certain state.
+        /// Updates state of any buttons that have been released by parent or axes that have changed value while <see cref="UseParentInput"/> was disabled.
         /// </summary>
-        /// <param name="state">The state to synchronise current with. If this is null, it is regarded as an empty state.</param>
-        protected virtual void SyncInputState(InputState state)
+        private void syncIgnoredInput()
```
smoogipoo, line 176: "ignored input" is a bit weird. I understand why you've named it that but perhaps just calling it "syncReleasedInputs" is better?

osu.Framework/Input/PassThroughInputManager.cs
frenzibyte: Looks like a refactor error / me relying too much on Rider tools.

osu.Framework/Input/PassThroughInputManager.cs
frenzibyte: Yeah I wasn't 101% sure about explicitly saying "released" since it contains code for updating joystick axes as well. I'll split it into two methods just so the purpose of the methods are still completely clear.

smoogipoo (approved):
I think this is fine.

PR #6222, opened 2024-03-20
Simplify `CompositeDrawable.CheckChildrenLife`

EVAST9919:
Was trying to investigate why ccl is so expensive in complex scenes, but failed (my guess is that it's just the loop itself).
Now looks way simpler (imo) without `ChildLifeStateChange` enum.
No performance difference.
Kinda useless but eh


smoogipoo:
I'm fairly certain this was very intentionally done so as to keep the iteration order the same. https://github.com/ppy/osu-framework/pull/966

Given that you state there's no substantial gain in this, I'm not willing to consider this change at the moment. Changing this will require substantial scrutinising to make sure it doesn't regress anything.

PR #6223, opened 2024-03-24
Add support for custom clipboard formats

minetoblend:
Allows copying values with custom formats to the clipboard. For now only writes custom formats to the system clipboard on windows, other platforms will store the value in-memory for now. I only did windows for now to not blur up the PR in size.
For simplicity, I limited the custom formats to be text-only.

Adds support for [web custom formats](https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md) on windows, for interoperability with the browser clipboard API.

Web custom formats expects a clipboard entry containing a map of all custom formats in json. The name of the clipboard entries depends on the platform. For windows these entries are expected to be in the `Web Custom Format Map` entry:
```json
{
  "text/custom" : "Web Custom Format0",
  "image/custom" : "Web Custom Format1"
}
```

## Changes made:
- `osu.Framework/Platform/Clipboard.cs`
  - Allow writing custom formats
  - Allow writing multiple formats at the same time
- `osu.Framework/Platform/HeadlessClipboard.cs`
  - Added in-memory storage for custom formats
- `osu.Framework/Platform/Windows/WindowsClipboard.cs`
  - Write custom formats to win32 clipboard api
  - When copying, will generate clipboard entries for custom formats to support web custom formats
  - When pasting, clipboard will fallback to web custom formats clipboard entries and will try to retrieve the data from those instead (to allow copy paste from browser -> osu framework)
- `osu.Framework/Platform/MacOS/MacOSClipboard.cs`
  - Added in-memory fallback for custom formats
- `osu.Framework/Platform/SDL2/SDL2Clipboard.cs`
  - Added in-memory fallback for custom formats
- `osu.Framework.Android/AndroidClipboard.cs`
  - Added in-memory fallback for custom formats
- `osu.Framework/Platform/ClipboardEntry.cs`, `osu.Framework/Platform/ClipboardTextEntry.cs`, `osu.Framework/Platform/ClipboardImageEntry.cs`, `osu.Framework/Platform/ClipboardCustomEntry.cs`
  - Added classes to represent different types of clipboard entries

### DIFF
--- a/osu.Framework/Platform/Clipboard.cs
+++ b/osu.Framework/Platform/Clipboard.cs
@@ -48,13 +48,14 @@ namespace osu.Framework.Platform
         /// <summary>
         /// Get clipboard content with custom format
         /// </summary>
-        /// <returns></returns>
+        /// <returns>Name of the custom format</returns>
         public abstract string? GetCustom(string format);
 
         /// <summary>
         /// Get clipboard content with custom format
         /// </summary>
-        /// <returns></returns>
+        /// <param name="format">Name of the custom format</param>
+        /// <param name="text">Text to copy to the clipboard</param>
         public void SetCustom(string format, string text)
         {
             SetData(
@@ -65,7 +66,7 @@ namespace osu.Framework.Platform
         /// <summary>
         /// Copy multiple values to the clipboard
         /// </summary>
-        /// <param name="entries"></param>
+        /// <param name="entries">Entries to copy the clipboard</param>
         public abstract bool SetData(params ClipboardEntry[] entries);
     }
 }
--- a/osu.Framework/Platform/Clipboard.cs
+++ b/osu.Framework/Platform/Clipboard.cs
@@ -20,7 +20,12 @@ namespace osu.Framework.Platform
         /// Copy text to the clipboard.
         /// </summary>
         /// <param name="text">Text to copy to the clipboard</param>
-        public abstract void SetText(string text);
+        public void SetText(string text)
+        {
+            SetData(
+                new ClipboardTextEntry(text)
+            );
+        }
 
         /// <summary>
         /// Retrieve an image from the clipboard.
@@ -33,6 +38,34 @@ namespace osu.Framework.Platform
         /// </summary>
         /// <param name="image">The image to copy to the clipboard</param>
         /// <returns>Whether the image was successfully copied or not</returns>
-        public abstract bool SetImage(Image image);
+        public bool SetImage(Image image)
+        {
+            return SetData(
+                new ClipboardImageEntry(image)
+            );
+        }
+
+        /// <summary>
+        /// Get clipboard content with custom format
+        /// </summary>
+        /// <returns></returns>
+        public abstract string? GetCustom(string format);
+
+        /// <summary>
+        /// Get clipboard content with custom format
+        /// </summary>
+        /// <returns></returns>
+        public void SetCustom(string format, string text)
+        {
+            SetData(
+                new ClipboardCustomEntry(format, text)
+            );
+        }
+
+        /// <summary>
+        /// Copy multiple values to the clipboard
+        /// </summary>
+        /// <param name="entries"></param>
+        public abstract bool SetData(params ClipboardEntry[] entries);
     }
 }
--- a/osu.Framework/Platform/Windows/WindowsClipboard.cs
+++ b/osu.Framework/Platform/Windows/WindowsClipboard.cs
@@ -2,10 +2,12 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
 using System.Text;
+using Newtonsoft.Json;
 using SixLabors.ImageSharp;
 using SixLabors.ImageSharp.Formats.Bmp;
 
@@ -22,6 +24,9 @@ namespace osu.Framework.Platform.Windows
 
         [DllImport("user32.dll")]
         private static extern IntPtr SetClipboardData(uint uFormat, IntPtr hMem);
+
+        [DllImport("user32.dll")]
+        private static extern uint RegisterClipboardFormatW(IntPtr lpszFormat);
 
         [DllImport("User32.dll")]
         [return: MarshalAs(UnmanagedType.Bool)]
@@ -63,17 +68,11 @@ namespace osu.Framework.Platform.Windows
 
         private static readonly byte[] bmp_header_field = { 0x42, 0x4D };
 
+        private readonly Dictionary<string, uint> customFormats = new Dictionary<string, uint>();
+
         public override string? GetText()
         {
             return getClipboard(cf_unicodetext, bytes => Encoding.Unicode.GetString(bytes).TrimEnd('\0'));
-        }
-
-        public override void SetText(string text)
-        {
-            int bytes = (text.Length + 1) * 2;
-            IntPtr source = Marshal.StringToHGlobalUni(text);
-
-            setClipboard(source, bytes, cf_unicodetext);
         }
 
         public override Image<TPixel>? GetImage<TPixel>()
@@ -89,7 +88,118 @@ namespace osu.Framework.Platform.Windows
             });
         }
 
-        public override bool SetImage(Image image)
+        public override string? GetCustom(string formatName)
+        {
+            string? value = getClipboard(getFormat(formatName), bytes => Encoding.Unicode.GetString(bytes).TrimEnd('\0'));
+            if (value != null)
+                return value;
+
+            var webCustomFormats = getClipboard(getFormat("Web Custom Format Map"), bytes =>
+                JsonConvert.DeserializeObject<Dictionary<string, string>>(
+                    Encoding.ASCII.GetString(bytes))
+            );
+
+            if (webCustomFormats?[formatName] != null)
+            {
+                string? webValue = getClipboard(getFormat(webCustomFormats[formatName]), bytes => Encoding.ASCII.GetString(bytes).TrimEnd('\0'));
+                return webValue;
+            }
+
+            return null;
+        }
+
+        private uint getFormat(string formatName)
+        {
+            if (customFormats.TryGetValue(formatName, out uint format))
+            {
+                return format;
+            }
+
+            IntPtr source = Marshal.StringToHGlobalUni(formatName);
+
+            uint createdFormat = RegisterClipboardFormatW(source);
+
+            if (createdFormat != 0)
+            {
+                GlobalFree(source);
+            }
+
+            customFormats[formatName] = createdFormat;
+
+            return createdFormat;
+        }
+
+        public override bool SetData(params ClipboardEntry[] entries)
+        {
+            if (entries.Length == 0)
+            {
+                return false;
+            }
+
+            var rawEntries = new List<RawClipboardEntry>();
+
+            var webCustomFormats = new Dictionary<string, string>();
+
+            foreach (var entry in entries)
+            {
+                switch (entry)
+                {
+                    case ClipboardTextEntry textEntry:
+                        rawEntries.Add(createTextEntry(textEntry.Value, cf_unicodetext));
+                        break;
+
+                    case ClipboardImageEntry imageEntry:
+                        rawEntries.Add(createImageEntry(imageEntry.Value));
+                        break;
+
+                    case ClipboardCustomEntry customEntry:
+                        rawEntries.Add(
+                            createTextEntry(
+                                customEntry.Value,
+                                getFormat(customEntry.Format)
+                            )
+                        );
+
+                        string webCustomFormatName = $"Web Custom Format{webCustomFormats.Count}";
+
+                        rawEntries.Add(createAnsiTextEntry(customEntry.Value, getFormat(webCustomFormatName)));
+
+                        webCustomFormats[customEntry.Format] = webCustomFormatName;
+                        break;
+                }
+            }
+
+            if (webCustomFormats.Count > 0)
+            {
+                rawEntries.Add(
+                    createAnsiTextEntry(
+                        JsonConvert.SerializeObject(webCustomFormats),
+                        getFormat("Web Custom Format Map")
+                    )
+                );
+            }
+
+            return setClipboard(rawEntries);
+        }
+
+        private RawClipboardEntry createTextEntry(string text, uint format)
+        {
+            int bytes = (text.Length + 1) * 2;
+            IntPtr source = Marshal.StringToHGlobalUni(text);
+
+            return new RawClipboardEntry(source, bytes, format);
+        }
+
+        private RawClipboardEntry createAnsiTextEntry(string text, uint format)
+        {
+            // Deliberately dropping the last byte here because browsers refuse to parse it otherwise
+            int bytes = text.Length * Marshal.SystemMaxDBCSCharSize;
+            IntPtr source = Marshal.StringToHGlobalAnsi(text);
+
+            return new RawClipboardEntry(source, bytes, format);
+        }
+
+        private RawClipboardEntry createImageEntry(Image image)
         {
             byte[] array;
 
@@ -103,12 +213,33 @@ namespace osu.Framework.Platform.Windows
             IntPtr unmanagedPointer = Marshal.AllocHGlobal(array.Length);
             Marshal.Copy(array, 0, unmanagedPointer, array.Length);
 
-            return setClipboard(unmanagedPointer, array.Length, cf_dib);
+            return new RawClipboardEntry(unmanagedPointer, array.Length, cf_dib);
         }
 
-        private static bool setClipboard(IntPtr pointer, int bytes, uint format)
+        private readonly struct RawClipboardEntry
         {
-            bool success = false;
+            public readonly IntPtr Pointer;
+            public readonly int Bytes;
+            public readonly uint Format;
+
+            public RawClipboardEntry(IntPtr pointer, int bytes, uint format)
+            {
+                Pointer = pointer;
+                Bytes = bytes;
+                Format = format;
+            }
+        }
+
+        private static bool setClipboard(List<RawClipboardEntry> entries)
+        {
+            if (entries.Count == 0)
+            {
+                return false;
+            }
+
+            bool success = true;
+
+            List<IntPtr> toFree = new List<IntPtr>();
 
             try
             {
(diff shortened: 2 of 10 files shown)


osu.Framework/Platform/ClipboardData.cs
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Collections.Generic;
+using SixLabors.ImageSharp;
+
+namespace osu.Framework.Platform
+{
+    /// <summary>
+    /// Holds multiple values that can be copied to the clipboard
```
bdach, line 10: I'm not sure if I understand this correctly because none of this is very clear really but this doesn't seem like an appropriate description. What a `ClipboardData` is supposed to contain is multiple concurrent _representations_ of the same data. As in, you can paste a text representation of the copied thing somewhere else if that somewhere else only handles text, but if it handles a richer format, it can use that richer format.

This just seems to imply you can stuff 30 different pieces of data into the various nooks and crannies and be good, and I do not think that's what any of the clipboard APIs are for.

osu.Framework/Platform/ClipboardData.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        /// <summary>
+        /// Image to be stored in the default image format in the clipboard.
+        /// </summary>
+        public Image? Image;
+
+        /// <summary>
+        /// Values to be stored as custom formats on the clipboard.
+        /// </summary>
+        public readonly Dictionary<string, string> CustomFormatValues = new Dictionary<string, string>();
```
bdach, line 27: This is keyed by mime type correct? The xmldoc should say that.

osu.Framework/Platform/ClipboardData.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        /// <summary>
+        /// Adds a clipboard entry with a custom format
+        /// </summary>
+        /// <param name="format">format of the clipboard entry</param>
+        /// <param name="value">value of the clipboard entry</param>
+        public void AddCustom(string format, string value)
+        {
+            CustomFormatValues[format] = value;
+        }
```
bdach, line 43: Not sure this has any right to exist if `CustomFormatValues` is completely public anyway.

osu.Framework/Platform/Clipboard.cs
```diff
@@ -20,7 +20,15 @@ public abstract class Clipboard
...
         /// <param name="text">Text to copy to the clipboard</param>
-        public abstract void SetText(string text);
+        public void SetText(string text)
+        {
+            SetData(
+                new ClipboardData
+                {
+                    Text = text
+                }
+            );
```
bdach, line 30: What's with the spacing?

```suggestion
            SetData(new ClipboardData { Text = text });
```

osu.Framework/Platform/Clipboard.cs
```diff
@@ -33,6 +41,39 @@ public abstract class Clipboard
...
+                {
+                    Image = image
+                }
+            );
+        }
+
+        /// <summary>
+        /// Get clipboard content with custom format
+        /// </summary>
+        /// <returns>Name of the custom format</returns>
```
bdach, line 57: mime type*

osu.Framework/Platform/Clipboard.cs
```diff
@@ -33,6 +41,39 @@ public abstract class Clipboard
...
+            data.AddCustom(format, text);
+
+            SetData(data);
+        }
+
+        /// <summary>
+        /// Copy multiple values to the clipboard
+        /// </summary>
+        /// <param name="data">Data to copy the clipboard</param>
+        public abstract bool SetData(ClipboardData data);
```
bdach, line 77: please document return value

osu.Framework/Platform/Windows/WindowsClipboard.cs
```diff
@@ -89,7 +88,109 @@ public override void SetText(string text)
...
+                return format;
+            }
+
+            IntPtr source = Marshal.StringToHGlobalUni(formatName);
+
+            uint createdFormat = RegisterClipboardFormatW(source);
+
+            if (createdFormat != 0)
+            {
+                GlobalFree(source);
```
bdach, line 124: - Why is the buffer only freed if the format creation succeeds? It should be getting freed regardless no?
- Why is the function trudging along unimpeded even if the return value of the call is zero (which means that the registration of the format has failed)? No logging, no nothing, just pretend everything is fine?

osu.Framework/Platform/Windows/WindowsClipboard.cs
```diff
@@ -23,6 +25,9 @@ public class WindowsClipboard : Clipboard
         [DllImport("user32.dll")]
         private static extern IntPtr SetClipboardData(uint uFormat, IntPtr hMem);
 
+        [DllImport("user32.dll")]
+        private static extern uint RegisterClipboardFormatW(IntPtr lpszFormat);
```
bdach, line 29:
```suggestion
        [DllImport("user32.dll", SetLastError = true)]
        private static extern uint RegisterClipboardFormatW(IntPtr lpszFormat);
```

https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-registerclipboardformatw#return-value

osu.Framework/Platform/Windows/WindowsClipboard.cs
```diff
@@ -89,7 +88,109 @@ public override void SetText(string text)
...
+            var webCustomFormats = getClipboard(getFormat("Web Custom Format Map"), bytes =>
+                JsonConvert.DeserializeObject<Dictionary<string, string>>(
+                    Encoding.ASCII.GetString(bytes))
+            );
+
+            if (webCustomFormats?[formatName] != null)
+            {
+                string? webValue = getClipboard(getFormat(webCustomFormats[formatName]), bytes => Encoding.ASCII.GetString(bytes).TrimEnd('\0'));
+                return webValue;
+            }
```
bdach, line 106: Why does this try the mime format as given directly first, and only *then* goes to the "web custom format" thing next? Why is this not documented anywhere?

osu.Framework/Platform/Windows/WindowsClipboard.cs
```diff
@@ -89,7 +88,109 @@ public override void SetText(string text)
...
+                clipboardEntries.Add(createImageEntry(data.Image));
+
+            if (data.CustomFormatValues.Count > 0)
+            {
+                // Required for compatibility with browser clipboard https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md#on-windows
+                var webCustomFormats = new Dictionary<string, string>();
+
+                foreach (var entry in data.CustomFormatValues)
+                {
+                    string webCustomFormatName = $"Web Custom Format{webCustomFormats.Count}";
```
bdach, line 154: What is happening here. Why is it using the _count_ always? Isn't the point that every mime type is implicitly mapped to a web custom format id or something? So it should depend on the _format_ no? Not on the _count of all registered formats_?

Not to mention other browsers that might be digging in this in the background as well at the same time. I'm not even going to ask whether this was _tested_ as co-habiting correctly with another web browser.

osu.Framework/Platform/Windows/WindowsClipboard.cs
```diff
@@ -89,7 +88,109 @@ public override void SetText(string text)
...
+
+                clipboardEntries.Add(
+                    createAnsiTextEntry(
+                        JsonConvert.SerializeObject(webCustomFormats),
+                        getFormat("Web Custom Format Map")
+                    )
+                );
+            }
+
+            return setClipboard(clipboardEntries);
```
bdach, line 173: So the entries here are:

- plain text
- image
- the web custom format stuff

but the direct mime type does not have an entry? But when _reading_ clipboard it takes precedence over the web custom format stuff?

No idea what anything is here.

bdach (changes requested):
don't even know where to begin with this

osu.Framework/Platform/Windows/WindowsClipboard.cs
minetoblend: ## Number Format
The format using the count is required for supporting web custom formats, and differs based on the platform.

The reason they chose to do this is because Windows has a hard limit of about 16.000 clipboard formats that can be registered, which is shared between all applications. Linux has a limit for formats in place as well. The formats are using the above naming convention so clipboard formats get reused & websites can't register arbitrary amounts of new clipboard formats.
https://github.com/w3c/editing/blob/gh-pages/docs/clipboard-pickling/explainer.md#os-interaction-format-naming

On Windows the following clipboard entries are required to support this:
- `Web Custom Format Map` - contains a map of all clipboard formats, which maps the original formats to the reusable web custom format entires, and looks as follows:
```json
{
  "text/foo": "Web Custom Format0",
  "text/bar": "Web Custom Format1"
}
```
- `Web Custom Format<n>` - contains the actual clipboard data

MacOS does the same thing but instead of `Web Custom Format<n>` it uses `org.w3.web-custom-format.type-<n>`. Linux uses the format `application/web;type="custom/format<n>`

I will add documentation to the code that explains the reasoning behind all this.

## Browser Support
I have tested this on the following browsers:
- Chrome: supported
- Brave: supported
- Opera: supported
- Edge: supported
- Firefox: has agreed to the spec but has not finished implementing it yet
- Safari: does not support custom clipboard formats currently and has not announced any plans to do so. Since it doesn't run on windows anyways this should be a non-issue here

osu.Framework/Platform/Windows/WindowsClipboard.cs
minetoblend: It does make an entry for the original format. I can refactor this to separate the web custom format logic into it's own thing to make it more clear what this method does.

These are the clipboard entries created when copying a combination of a plain text entry as well as `osu/hitobjects`. Note that 3 of the 4 `CF_` entries are created implicitly by Windows.

image

osu.Framework/Platform/Windows/WindowsClipboard.cs
bdach: I would like to think I understood the concept of the mapping. I mostly got confused by the implementation, I failed to notice that the count changes as you iterate through this because `webCustomFormats` gains entries as this loop progresses.

That said why does this presume that the `Web Custom Format0` thingies are always zero-indexed and always use consecutive numbers? Is the point that every one someone puts something in the clipboard, they're also expected to touch the _registry_ to update the map thingy to their own demands? God almighty what a terrible setup.

osu.Framework/Platform/Windows/WindowsClipboard.cs
bdach: I probably got confused by the code because of inconsistent signatures. There's

```csharp
private ClipboardEntry createTextEntry(uint format, string text);
```

but then there's

```csharp
private ClipboardEntry createAnsiTextEntry(string text, uint format);
```

and I don't even know what I'm reading at that point. Please unify ordering.

osu.Framework/Platform/Windows/WindowsClipboard.cs
minetoblend: I noticed that the same applies to all other imported clipboard functions, should I be setting this flag on all of them?

bdach: Probably yes.

osu.Framework/Platform/Windows/WindowsClipboard.cs
minetoblend: They are essentially representing the same thing but in a different format. Since we can't exactly use both values if both are present, one of the two formats has to be prioritized. Prioritizing the original mime type instead of one only there for compatibility with browsers seems to be the best way to address this to me.

I added a comment explaining what I'm doing there, though I'm not quite sure how to explain why the actual format should go first.

osu.Framework/Platform/Windows/WindowsClipboard.cs
minetoblend: Alright, will do. What is the expected behaviour here if things fail (i.e. during copying or retrieving a value)? If I throw an exception this will need to be handled on the osu side of things.

osu.Framework/Platform/Windows/WindowsClipboard.cs
minetoblend: That was an oversight on my part, got some stuff mixed up with the pointer stuff. Yes this should always be freed. I am logging this as an error now, and have added documentation that 0 as return value indicates that registering the format was not successful. Further error handling should be done by the caller I think, since it might depend on situation what we want to do if this fails.

osu.Framework/Platform/Windows/WindowsClipboard.cs
bdach: Probably don't throw, log an error, return false to indicate clipboard wasn't set.

osu.Framework/Platform/Windows/WindowsClipboard.cs
minetoblend: I'm not sure if I'm following you correctly here, are you asking why I have chosen to use this format when writing the values? 

Browsers will actually read the value from any clipboard entry as long as it's in the map (I tested this). However, the preferred encoding for the clipboard on windows is UTF-16, while the browser treats these values as UTF-8, so a separate entry is needed with the value encoded as UTF-8. I chose to write to the `Web Custom Format0`... range because that's the same range that browsers are writing to. That way we can avoid registering more clipboard formats than needed, if a browser has already registered them.

I can make it write the custom mime type entries as UTF-8 instead, then we can have the map point at the actual entries instead. However, https://learn.microsoft.com/en-us/windows/win32/intl/unicode sounds to me like UTF-16 is what we should be using for those

osu.Framework/Platform/Windows/WindowsClipboard.cs
minetoblend: I skipped `IsClipboardFormatAvailable` and `GetClipboardData` since a zero return value is expected here and just means that the format was not unavailable.

osu.Framework/Platform/ClipboardData.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        public string? Text;
+
+        /// <summary>
+        /// Image to be stored in the default image format in the clipboard.
+        /// </summary>
+        public Image? Image;
+
+        /// <summary>
+        /// Contains Values to be stored as an entry custom mime type in the clipboard.
```
bdach, line 25: "entry custom mime type" is a bit nonsense, probably needs a reword

osu.Framework/Platform/Windows/WindowsClipboard.cs
bdach: No my question has nothing to do with utf. I have not mentioned "utf" once above.

My question is, and continues to be: is the `N` in the `Web Custom FormatN` entry mapped to a specific MIME type for a particular single copy operation? As in, are consumers expected to clear the entire mapping `WebCustomFormat` mapping and overwrite it with the mime types they're _specifically_ putting in the clipboard themselves?

If that is the case, then I'd say the `customFormats` dictionary needs to be cleared on every invocation of `SetData()`.

osu.Framework/Platform/Windows/WindowsClipboard.cs
bdach: > I skipped `IsClipboardFormatAvailable` and `GetClipboardData` since a zero return value is expected here

I mean both functions still expose error info via `GetLastError()` as far as I can tell? Even if we're not gonna check it they should be set as such.

osu.Framework/Platform/Windows/WindowsClipboard.cs
minetoblend: The mapping applies to the lifetime of a given clipboard content, they will remain valid and unchanged until the clipboard gets cleared again. The same `Web Custom FormatN` format can and will be reused for different mime types.

The `customFormats` dictionary contains a mapping from the format name (i.e. `Web Custom FormatN`) to the identifier that win32 has assigned to that format name. Once a format has been registered, `RegisterClipboardFormatW` will always return the same identifier for a clipboard format, so clearing the dictionary would have no effect besides having to register the format again. (Now that I think about it I should probably rename this field to something more descriptive)

Consumers aren't supposed to clear anything here, they just look at the `Web Custom Format Map` entry (if it's present) on each paste operation and then figure out the actual clipboard entry to use from there.

bdach:
I guess I have no major issues with the code itself but I'm not sure how I would test that this does what it says it does if I wanted to.

minetoblend:
> I guess I have no major issues with the code itself but I'm not sure how I would test that this does what it says it does if I wanted to. A test scene or something would go a long way.

Sure thing, I have a couple questions for testing this
- For testing the web custom formats, should I just check that the generated clipboard entries match an expected output, or should the test verify that the input is indeed parseable by a web browser?
- If I actually verify that the clipboard works with browsers, would spinning up a headless browser (i.e. puppeteer with https://github.com/hardkoded/puppeteer-sharp) during the test be fine?
- Since such a test would not make much sense on other platforms, can I somehow limit the test to run on windows only?

bdach:
>     * For testing the web custom formats, should I just check that the generated clipboard entries match an expected output, or should the test verify that the input is indeed parseable by a web browser?

The test can be semi-manual. I'd just like instructions how to perform it.

>     * If I actually verify that the clipboard works with browsers, would spinning up a headless browser (i.e. puppeteer with https://github.com/hardkoded/puppeteer-sharp) during the test be fine?

Don't do that.

>     * Since that particular test would not make much sense on other platforms, can I somehow limit the test to run on windows only?

[There is.](https://github.com/ppy/osu-framework/blob/61704343872cda8949d45c4779ad91d83b8f5d88/osu.Framework.Tests/Visual/Platform/TestSceneWindowFlash.cs#L16)

minetoblend:
> The test can be semi-manual. I'd just like instructions how to perform it.
> 
> >

I created an [html file](https://gist.github.com/minetoblend/7c9af3c8a0450bcf9f1819eac6b6e6e1) which can be used to test this. Note that the clipboard API can only be used from a secure context, so the page needs to be served via https. I hosted it over at https://clipboard-test.tiiny.site/ for convenience

If you want to test things manually, you can use these javascript snippets:
- Reading from clipboard
```js
// needs to be triggered by a user-caused event (i.e. click)
const [item] = await navigator.clipboard.read();

if(item && item.types.contains('mime type I need')) {
  const blob = await item.getType('mime type I need')
  const text = await blob.text()
}
```
- Copy co clipboard
```js
// needs to be triggered by a user-caused event (i.e. click)
const item = new ClipboardItem({
  'text/plain': new Blob(['text content'], { type: 'text/plain' }),
  'web custom/format': new Blob(['custom format content'], { type: 'web custom/format' })
})
await navigator.clipboard.write([item])
```

bdach:
> I also added some test cases in [ppy/osu/pull/27707](https://github.com/ppy/osu/pull/27707) that should hopefully be helpful for this

I would hope something can be included on the framework side too. It can be as simple as just printing the clipboard contents in all available formats out on the screen as a string, but it should live in framework.

bdach:
> By printing the clipboard contents to the screen, do you mean grabbing them from the system clipboard or just getting the list of entries that the `WindowsClipboad` wants to create?

Just add a text flow and put the key-value pairs with the clipboard contents in it, inside a standard visual test scene.

> Also, is a regular Test what I should be making to do this? Not sure since there's no assertions and sounds like it should be ran manually instead of being part of the test suite.

Yes it's fine to write a test without assertions for this. There are many like it, especially around keyboard layout handling and the like.

The point of this excursion is so that I can test that the functionality works without having to check out the game or something. The framework is a standalone project in principle and should be treated as such (even if in practice it turns out to be mostly subservient to the game's needs).

natsukagami:
>  linux would require an upgrade to SDL3

Looks like we're getting it with ppy/osu-framework#6234 (already in) and ppy/osu#27815 (soon)

PR #6225, opened 2024-03-24
Hide `PassThroughInputManager` from input queue if parent input is disabled

frenzibyte:
- Closes https://github.com/ppy/osu/issues/9658
- [x] Depends on #6221 

This resolves issue 1 pointed at https://github.com/ppy/osu/issues/9658#issuecomment-2005767323, and together with #6221, will resolve the issue entirely.

When the user presses the key to resume gameplay, `ResumeOverlay` enables the game input manager inside the key binding event handler. 

In master, `PassThroughInputManager` is still included in the input queue even if `UseParentInput` is disabled, therefore, the game input manager will be next in line to handle the resumption key press, and since it's enabled by the `ResumeOverlay` before its turn comes to handle the event, it will end up handling it and incorrectly counting it as a game press (`ResumeOverlay` cannot block the key by returning true because it is nested in a pass-through input manager by itself).

With this PR, `PassThroughInputManager` will be hidden from the input queue if `UseParentInput` is disabled, therefore, even if `ResumeOverlay` enables the game input manager, the input manager will not be in the input queue to handle the key press.

Alternative path would be to schedule the resume operation in osu! so that the resumption key press event in the current frame passes all over the drawables in the input queue without the game input manger handling it, then have game input manager become enabled next frame.

Checked and made sure tests pass in here and in osu! as well.


smoogipoo:
Wanna resolve conflicts here?

smoogipoo (approved):
Lgtm

PR #6228, opened 2024-03-27
Tweak behavior of cursor movement while text is selected to better match expectations

ggliv:
In most common text input fields that I've tested (Chrome, Firefox, Word, etc.), a left/right cursor movement while text is selected will place the cursor on the left/right border of the selection.

https://github.com/ppy/osu-framework/assets/47010459/9f6ee6fd-e0b5-44e8-9449-a089ee73d585

Currently, o!f's `TextBox` treats these cursor movements as... just regular movements. This can lead to some weird/unexpected cursor placements when coming out of selections.

https://github.com/ppy/osu-framework/assets/47010459/65ab1d54-c72d-49ce-a2a4-044a764664b2

These changes bring the behavior closer to what is usually seen in other programs.

https://github.com/ppy/osu-framework/assets/47010459/5a0f22a7-9011-45cc-bf59-149182c1d01a

Behavior when moving along word borders is also changed. Different programs I checked against had different behaviors, so I elected to copy what Chrome does (jump to the next/previous word from the selection borders).

### DIFF
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -302,23 +302,27 @@ namespace osu.Framework.Graphics.UserInterface
                     return true;
 
                 case PlatformAction.MoveBackwardWord:
-                    if (hasSelection && selectionEnd != selectionLeft)
+                    if (hasSelection)
                     {
-                        selectionEnd = selectionLeft;
-                        selectionStart = lastSelectionBounds.end;
+                        MoveCursorBy(selectionLeft - selectionEnd);
+                    }
+                    else
+                    {
+                        MoveCursorBy(GetBackwardWordAmount());
                     }
 
-                    MoveCursorBy(GetBackwardWordAmount());
                     return true;
 
                 case PlatformAction.MoveForwardWord:
-                    if (hasSelection && selectionEnd != selectionRight)
+                    if (hasSelection)
                     {
-                        selectionEnd = selectionRight;
-                        selectionStart = lastSelectionBounds.end;
+                        MoveCursorBy(selectionRight - selectionEnd);
+                    }
+                    else
+                    {
+                        MoveCursorBy(GetForwardWordAmount());
                     }
 
-                    MoveCursorBy(GetForwardWordAmount());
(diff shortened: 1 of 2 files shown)


osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -278,18 +278,50 @@ public virtual bool OnPressed(KeyBindingPressEvent<PlatformAction> e)
 
                 // Cursor Manipulation
                 case PlatformAction.MoveBackwardChar:
-                    MoveCursorBy(-1);
+                    if (hasSelection)
+                    {
+                        selectionStart = selectionEnd = selectionLeft;
+                        cursorAndLayout.Invalidate();
+                        onTextDeselected(lastSelectionBounds);
```
peppy, line 285: Makes me a bit uneasy that the implementation of `MoveCursorBy` was copied here. It's not the end of the world, but could this potentially be replace by a call to `MoveCursorBy` with a known offset value?

ggliv: Sure

osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -278,18 +278,46 @@ public virtual bool OnPressed(KeyBindingPressEvent<PlatformAction> e)
...
                     MoveCursorBy(GetBackwardWordAmount());
                     return true;
 
                 case PlatformAction.MoveForwardWord:
+                    if (hasSelection && selectionEnd != selectionRight)
+                    {
+                        selectionEnd = selectionRight;
+                        selectionStart = lastSelectionBounds.end;
+                    }
+
```
peppy, line 320: This change doesn't match my expectations. When there's a selection, this should be doing precisely the same thing as the other change you made (ie. it shouldn't skip a word while there's a selection – just move the cursor to the matching side of the selected word).

ggliv: Yeah this is the other behavior I saw. I don't have a strong preference on which is better, let's just go with your suggestion.

peppy (approved):

PR #6229, opened 2024-03-27
Refactor `KeyCombination.ContainsKey()` and `.ContainsKeyPermissive()` for better extensibility

Susko3:
- Prerequisite for fixing #5790
  - That PR is blocked because of [faulty logic](https://github.com/ppy/osu-framework/pull/5790/commits/b74b208dbccf60d6d25c7b56758996816067b4b6), this PR will make it easier to fix that logic

## Terminology

- Physical key: physical key on a [HID](https://en.wikipedia.org/wiki/Human_interface_device), such as a `osuTK.Input.Key`, `MouseButton`, `JoystickButton` and their direct mappings to `InputKeys`
- Virtual key: a logical key, an abstraction of some physical key(s). Example:
  - `(Input)Key.LControl` and `(Input)Key.RControl` map to the virtual key `InputKey.Control`
  - `SDL_Keycode` (the keyboard layout-dependant unicode character) that is used in #5790

### `KeyCombination.ContainsKey()` is used for two different purposes

These two purposes have very different semantics:

1. It's used to check that each InputKey from a candidate key combination is present in the _pressedKey_ [sic] array.
   - In this case, the key param can be a virtual key.
2. For `KeyCombinationMatchingMode.{Exact,Modifiers}`, it's used to check that there are no excess `InputKey`s from pressedKeys present in the combination.
   - In this case, the key param can only be a physical key and the switch is unreachable code.

This is evident from the xmldoc which explains 1. in `<summary>` and 2. in `<param name=.../>`

https://github.com/ppy/osu-framework/blob/51d8215266dc861ef2554d43a92d7b33237872f2/osu.Framework/Input/Bindings/KeyCombination.cs#L198-L205

This is fixed in e05083a750ce673b13d358db98edc17053495a31 and invalid tests are removed in 3e021ddd197138a97504d59d00fb217aa7868510 (check commit message as to why).

### Adding additional virtual keys is complicated

Since definitions of virtual keys are spread across `ContainsKeyPermissive()` and `ContainsKey()`, it's hard to add new ones, and impossible to do it programmatically.

This is fixed by 0ab0adc4c7e8206dc41e6e145feaf86c0e0594ff without changing semantics. The added `getVirtualKey()` static method is subject to change and will need to have the `InputState` propagated to it to make it work for #5790.

I'll be doing some follow-up cleanup after this PR is merged.

# Breaking changes

## `KeyCombination.IsPressed` now takes in an `InputState` parameter

```diff
-public bool IsPressed(KeyCombination pressedKeys, KeyCombinationMatchingMode matchingMode)
+public bool IsPressed(KeyCombination pressedKeys, InputState inputState, KeyCombinationMatchingMode matchingMode)
```


bdach:
This is so difficult to make sense of that I am trying to approach this PR a second time and still not understanding any of it.

- Why are the methods called `ContainsKey()` and `ContainsKeyPermissive()` if the "permissive" one doesn't look all that more or less permissive than the other? They both check the "physical" and the "virtual" key. The _actual_ difference seems to be in the directionality of the comparison? I think?

  The methods need renames to not be borderline obfuscatory.
- If I read in (2):

  > In this case, the key param can only be a physical key and the switch is unreachable code"

  Where are the _assertions_ for this? I want this to be asserted _in code_. So that determining that something breaks the invariant is so simple as seeing the game crash.
- Later on:

  > Since definitions of virtual keys are spread across ContainsKeyPermissive() and ContainsKey(), it's hard to add new ones, and impossible to do it programmatically

  Why do we want this again?

Susko3:
> > Since definitions of virtual keys are spread across ContainsKeyPermissive() and ContainsKey(), it's hard to add new ones, and impossible to do it programmatically
>       
> Why do we want this again?

This is needed to unblock #5790. That PR currently takes a naive approach to adding virtual keys: it just [adds them in `KeyBindingsContainer.pressedInputKeys`](https://github.com/ppy/osu-framework/pull/5790/files#diff-03aa999bf8c9aaa29c947145a03d49790fa05e45a3dd08dba2607d41ca5d1eb6R298-R304). This same mistake (but with virtual modifier keys) was fixed in 39032c9213012cb6c806c5a13232d518cdaa92b8, but there still exists a lot of code that expects both physical and virtual keys to be in the user pressed keys array. I believe this may add to the confusion.

osu.Framework/Input/Bindings/KeyCombination.cs
```diff
@@ -103,42 +104,71 @@ public bool IsPressed(KeyCombination pressedKeys, KeyCombinationMatchingMode mat
...
+        /// <param name="candidate">The candidate key binding to match against.</param>
+        /// <param name="pressedKeys">The keys which have been pressed by a user.</param>
         /// <param name="matchingMode">The matching mode to be used when checking.</param>
         /// <returns>Whether this is a match.</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal static bool ContainsAll(ImmutableArray<InputKey> candidateKey, ImmutableArray<InputKey> pressedKey, KeyCombinationMatchingMode matchingMode)
+        internal static bool ContainsAll(ImmutableArray<InputKey> candidate, ImmutableArray<InputKey> pressedKeys, KeyCombinationMatchingMode matchingMode)
         {
+            Debug.Assert(pressedKeys.All(k => k.IsPhysical()));
+            ImmutableArray<(InputKey Physical, InputKey? Virtual)> pressed = pressedKeys.Select(k => (k, getVirtualKey(k))).ToImmutableArray();
```
Susko3, line 142: This is really bad for perf. Should be bubbled up to `KeyBindingContainer.pressedInputKeys` (currently a `HashSet<InputKey>`, but should be changed to a `Dictionary<InputKey, InputKey?>` that maps [physical key -> virtual key]).

Leaving for later, when the code makes it past the first round of review.

Susko3:
Idk, it's really hard to explain and reason as the current code is broken in subtle ways and does things that don't make sense. But together it works out and the errors are not visible to the outside world.

osu.Framework/Input/Bindings/KeyCombination.cs
bdach: I don't get it, if it's a concern, why isn't there a static dictionary in this class then?

bdach (commented):
seems better than before but i still think it could be better yet

osu.Framework/Input/Bindings/KeyCombination.cs
```diff
@@ -103,42 +104,71 @@ public bool IsPressed(KeyCombination pressedKeys, KeyCombinationMatchingMode mat
...
         /// <param name="matchingMode">The matching mode to be used when checking.</param>
         /// <returns>Whether this is a match.</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal static bool ContainsAll(ImmutableArray<InputKey> candidateKey, ImmutableArray<InputKey> pressedKey, KeyCombinationMatchingMode matchingMode)
+        internal static bool ContainsAll(ImmutableArray<InputKey> candidateKeyBinding, ImmutableArray<InputKey> pressedPhysicalKeys, KeyCombinationMatchingMode matchingMode)
         {
+            Debug.Assert(pressedPhysicalKeys.All(k => k.IsPhysical()));
+            ImmutableArray<(InputKey Physical, InputKey? Virtual)> pressed = pressedPhysicalKeys.Select(k => (k, getVirtualKey(k))).ToImmutableArray();
+
+            Debug.Assert(pressed.All(k => k.Virtual == null || k.Virtual.Value.IsVirtual()));
```
smoogipoo, line 144: I suppose it's not a huge deal, but is there a reason this was structured to map physical/virtual keys in a separate array here rather than using `getVirtualKey()` at the points where it's required?

There are two minor concerns:
1. Obviously this will allocate an array every time. It's a small and infrequent allocation though, so I'm willing to let this pass.
2. The API of `KeyBindingContains` and `IsPressed` become a little bit weird now that the consumer is able to specify a separate physical and virtual key with what appears to be no validation. Maybe I'm misunderstanding something, but it sounds like [given your commit message about the invalid test cases](https://github.com/ppy/osu-framework/commit/3e021ddd197138a97504d59d00fb217aa7868510) that the user should never be dealing with virtual keys outside of `KeyCombination`. The APIs I'd expect are:
    - `KeyBindingContains(ImmutableArray<InputKey> candidateKeyBinding, InputKey physicalKey)`
    - `IsPressed(InputKey virtualOrPhysicalKey, InputKey candidateKey)`

Again please let me know if I've missed something.

smoogipoo: Here's my initial attempt at refactoring this to better explain what I mean: https://gist.github.com/smoogipoo/af35edc6492de8d5ea47ca4d64b7fc42

I'm wondering if there's some part of the extensibility argument that I'm missing with this.

smoogipoo: > The methods you're talking about are `internal`

Oh, I didn't notice that.

> I think the first one makes the most sense, as we can have the `getVirtualKey()` logic contained inside `KeyCombination`. Your refactor works well with that.

That one makes the most sense to me too for a public API. Maybe the breaking change should be in this PR because I don't know how it's going to be used - I'm only reviewing purely from a code-quality and isolation perspective because you definitely have a better vision for how this all works.

Unless the breaking change + implementation is significant :)

osu.Framework/Input/Bindings/KeyCombination.cs
```diff
@@ -91,54 +92,82 @@ private KeyCombination(ImmutableArray<InputKey> keys)
...
             Debug.Assert(!pressedKeys.Keys.Contains(InputKey.None)); // Having None in pressed keys will break IsPressed
 
             if (Keys == pressedKeys.Keys) // Fast test for reference equality of underlying array
                 return true;
 
-            return ContainsAll(Keys, pressedKeys.Keys, matchingMode);
+            return ContainsAll(Keys, pressedKeys.Keys, inputState, matchingMode);
+        }
+
+        private static InputKey? getVirtualKey(InputKey key, InputState inputState)
```
bdach, line 108: Why does this receive the full `InputState`? Is the goal here to eventually read the `Characters` thing from https://github.com/ppy/osu-framework/pull/5790, then I'd rather just pass that in specifically _when_ it exists.

The reason why I have an issue with this is that the `InputState` is propagated through methods multiple levels and when I look at the method signature of `ContainsAll()` I get _very_ uneasy because there's _three_ key collections there now - there's `candidateKeyBinding` and `pressedPhysicalKeys` which you are _supposed_ to be checking, but then there is also `InputState` which really just allows you to read everything from everywhere and that feels like terrible layering. `KeyBindingContains()` is similar but not as egregious imo.

I'm fine with the `IsPressed()` signature fwiw, it's everything below that that I have an issue with.

smoogipoo: Yeah that's what I was also hoping to figure out - how the `InputState` is actually going to be used here.

Susko3: The `InputState` will be used to read the `Characters` of the passed in `InputKey`. And if the character is `[a-z]`, it'll return the corresponding `InputKey.Keycode{A-Z}`.

> I'd rather just pass that in specifically _when_ it exists.

Alright, I'll do that.

bdach:
I think I can live with this diff now, but do note I brought some of the tests removed in https://github.com/ppy/osu-framework/pull/6229/commits/3e021ddd197138a97504d59d00fb217aa7868510 in an altered manner. See https://github.com/ppy/osu-framework/pull/6229/commits/8162563e786629b5aed1b360c8594a7c85dd6cfa for reasoning.

bdach (approved):

smoogipoo (approved):
lgtm

PR #6231, opened 2024-03-29
Split words on more characters in text flow container

bdach:
- Closes https://github.com/ppy/osu/issues/10085
- Closes https://github.com/ppy/osu/issues/14837

I'm not really considering the performance implications of this, just looking to get a baseline more-correct behaviour in place.

The change is basically a best-effort heuristic, although I did use a slightly smarter split method for CJK to attempt to ensure no hanging punctuation.


peppy (dismissed):
Feels like we should have applied this ages ago 👍

peppy:
Looks pretty good here:

https://github.com/ppy/osu-framework/assets/191335/4e01238b-656e-4eda-9eb6-b3374934a591

peppy (approved):

PR #6233, opened 2024-04-04
Update BASS libraries and add BASSLoud & BassEnc

smallketchup82:
Prerequisite for https://github.com/ppy/osu/pull/27793

This PR updates the BASS libraries to their latest version and also adds BASSLoud, and its dependency, BassEnc. In my testing using these, there does not seem to be any breaking changes or incompatibility. I still recommend testing this though.


smallketchup82:
Marking this  PR as a draft for now since I plan on rebasing it on top of https://github.com/ppy/osu-framework/pull/6352

Will likely force push the previous diffs away.

smallketchup82:
I've rebased the branch on top of master and cleaned up history a bit.

Here's a full changelog:
- Since master now has the latest bass libs, updating them here is no longer needed. I'll change the title and description to match this. It should probably make reviewing this PR a lot easier as now we're only adding new libs
- I've also taken to noting that osu-framework logs the version of bass and its addons on startup. I've added BASSloud to this
- As frenzibyte pointed out to me a while back, this PR doesn't actually add the libraries for mobile platforms. This is now fixed

For anyone that has this branch cloned locally, I'd recommend fetching and resetting to the remote branch head. Or delete and check out this branch again. I've confirmed with most that I know that have this branch cloned and made sure they do not have any local changes.

PR #6234, opened 2024-04-08
Upgrade to SDL3

Susko3:
osu! changes: https://github.com/ppy/osu/compare/master...Susko3:osu:wip-SDL3?expand=1

## Changes
- (SDL3-CS) from `namespace SDL2` & `class SDL` to `namespace SDL` & `class SDL3` (made more sense as the functions are imported from `SDL3.dll`)
- (SDL3-CS) structs and enums now live in the namespace instead of being nested in the `SDL3` class
- (SDL3-CS) pointer types are now proper pointers instead of `IntPtr`, which means more "unsafe" code
- (SDL3-CS) minor changes to return/struct types, reducing need for casts
- (SDL3-CS) simple function signature changes: from `ref`, `in` & `out` to pointers (unfortunate, but automatically generating these would be hell)
- (SDL3-CS) function that take strings now take `ReadOnlySpan<byte>`, requiring `Encoding.UTF8.GetBytes()` calls or [UTF-8 string literals](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/reference-types#utf-8-string-literals)
- (SDL3-CS) [function pointers](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/unsafe-code#function-pointers) instead of delegates in native callback functions
- simple renames done via SDL-provided [`build-scripits/rename_symbols.py`](https://github.com/libsdl-org/SDL/blob/main/build-scripts/rename_symbols.py)
- migrations from the [Migrating to SDL 3.0](https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md) guide
- minor update of `SDL3ControllerBindings` logic (basically per SDL3 migration guide)
- **removal of borderless fullscreen on macOS and Linux** (merged into fullscreen with SDL3)
- updated windowing logic to account for `SDL_DisplayID` instead of previously used display index
- osu!framework **`DisplayMode.RefreshRate` changed from `int` to `float`**
- SDL relative mouse mode instead of custom `WindowsMouseHandler` mouse logic
  - including removal of custom tablet handling logic. `SDL_pen.h` would be the replacement for that, but it's currently only supported on Linux
- updated `WindowsWindow` to account for `SDL_SYSWMEVENT` removal
  - has caused some minor regressions in IME handling (the new method isn't giving us all the expected messages)
- updated `JetBrains.Annotations` to latest version
- renamed SDL2 → SDL3

## Issues
- `IOpenGLGraphicsSurface.BackbufferFramebuffer` used to return useful information on iOS
   - waiting un upstream: https://github.com/libsdl-org/SDL/issues/9430
   - OpenGL probably won't work on iOS?
- minor IME regressions on windows
- releasing mouse button outside window with relative mouse mode causes next click to be ignored (SDL issue, it seems)
- (possible issue) non-OTD tablet input with relative mouse mode on windows (`WindowsMouseHandler` logic was removed)

## Future work
- add full support for async raw mouse + keyboard on windows #6235
- look into windows IME regressions (and maybe switch to SDL IME events)
- update my currently open PRs to SDL3 (when I feel like it)

## Testing
- [x] windows
- [x] linux (x11)
- [x] linux (wayland)
- [x] macos
- [x] ios
- [x] android

I've tested mouse, keyboard


smoogipoo:
Code quality wise this looks good to me.

smoogipoo:
I've tested Linux on Wayland and via XWayland, with the configs:

```
OSU_GRAPHICS_RENDERER=deferred OSU_GRAPHICS_SURFACE=vulkan SDL_VIDEO_DRIVER=wayland
OSU_GRAPHICS_RENDERER=deferred OSU_GRAPHICS_SURFACE=vulkan SDL_VIDEO_DRIVER=x11
SDL_VIDEO_DRIVER=wayland
SDL_VIDEO_DRIVER=x11
```

Haven't tested native X11, but that's probably okay. Resolved one issue in the process.

smoogipoo:
OpenGL is broken on iOS, but we've disabled that renderer. Metal works fine.

smoogipoo:
Android also tested to work fine.

smoogipoo (approved):

smoogipoo:
@peppy Maybe look to merge after the next release?

smoogipoo:
Can give it a shot. My concern would be the issues mentioned in the OP that I'm not fully sure the scope of.

peppy:
Yeah let's leave it one release.

osu.Framework/Platform/Linux/LinuxGameHost.cs
```diff
@@ -29,11 +29,11 @@ internal LinuxGameHost(string gameName, HostOptions? options)
 
         protected override void SetupForRun()
         {
-            SDL.SDL_SetHint(SDL.SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, BypassCompositor ? "1" : "0");
+            SDL3.SDL_SetHint(SDL3.SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, BypassCompositor ? "1"u8 : "0"u8);
```
peppy, line 32: Never seen this before 🤯

peppy (approved):
Code quality pass ✅ 
Tested on macOS ✅

peppy:
I think we can merge this at any point. We won't be doing another framework release / update before the osu! release this week.

That said, I'd probably want @bdach to sign off on this one at whatever depth he feels comfortable doing so. A bit of secondary windows testing is probably most important here, if anything.

bdach:
I'm just going to test that this launches in all environments that I have and call it a day. I don't think further review is productive.

bdach (approved):
doesn't die. have not read code. expecting 30 issues that were really upstream issues to be fixed and another 30 to regress

Issue #6181, opened 2024-02-17
Input thread blocking causes mouse handling stutters on windows


peppy:
Many reports of this recently, to the point I believe something may have regressed.

https://github.com/ppy/osu/issues/27208
https://github.com/ppy/osu/discussions/26998
https://github.com/ppy/osu/discussions/26953
https://github.com/ppy/osu/discussions/27211
https://github.com/ppy/osu/discussions/26875 (maybe)

owengaming001:
Hi, I made the https://github.com/ppy/osu/issues/27208 issue. I have since found out that this is related to my laptop keyboard, and isn't dependent on the touchpad. I've had typing lag-related issues in other programs, so it's possible that issue might not be entirely osu!lazer's fault. Though it's still worth noting that it doesn't seem to happen on stable. Also, it affected me for a while before I got around to reporting it.

Basically what I'm saying is: given the circumstances, my issue might be unrelated from the rest of those listed here. Hopefully this is helpful, sorry for any confusion or problems I may have caused. If there are any questions I can answer, then don't hesitate to let me know.

BenCheung0422:
I believe the four issues above are all related to tablet pen and the last one is related to mouse. I think it is generally cursor handling problem.

peppy:
Probably keep this open until a fix has been confirmed.

Sewlell:
Creator of [ppy/osu#27448](https://github.com/ppy/osu/issues/27488) here, the new April 12 Update seem to lessened the frequency of the stutter (included other rulesets, ie. sentakki) . Unfortunately it's still unresolved with several time it can be very noticeable and caused a FC choke.


Not sure about other people though ( need more confirmation ). But hey, at least I got 50/50 chances this time for me.

smoogipoo:
The update has not changed anything yet. This will likely be fixed or significantly improved in the next release.

IIIBlueberry:
The issue is still here, I did not notice mouse input(Logitech G102 Prodigy 1000hz polling) dropping on my old core 2 quad 9300 and r7 360 compared to my new better rig.

peppy:
> The issue is still here, I did not notice mouse input(Logitech G102 Prodigy 1000hz polling) dropping on my old core 2 quad 9300 and r7 360 compared to my new better rig.

To reiterate, this will only happen when something else on your system is handling input badly. For the time being, you can resolve this locally by trial-and-error shutting down apps (especially things in your system tray) until you find the culprit.

[...]

Issue #6141, opened 2024-01-18
Investigate receiving input without using window event flow


peppy:
Using window events can really suck if something on the OS decides to block the macOS main thread. Any app can do this and it will cause a delay in receiving events across all apps (example https://github.com/ppy/osu/issues/26604). I've hit this many times myself.

It may also be relevant on windows for key input, as we are not using raw input there (I've seen reports of some users hitting input delays on windows with similar blocking).

cc @Susko3 since you might have something to add here.

Susko3:
If I understand correctly, blocking the main thread only happens because we (or SDL) asks the OS for events (on Windows that would be via [`PeekMessage()`](https://github.com/libsdl-org/SDL/blob/f52a12a8fe9a6d52dfbed90310a894ce6be865c3/src/video/windows/SDL_windowsevents.c#L1855-L1858)) and that call blocks / takes a while for whatever reason. So the idea is to get (specifically) input events in another thread?

Would be useful to have a program that can generate this kind of lag. I played around in AutoHotkey and the best lag generator I could come up with is just sending a lot of mouse clicks. I'm unsure how representative this is of real-world lag, but it does spike the `InputThread` `WndProc` monitor and makes keyboard and mouse input lag.

Anyways, SDL3 recently added support for pooling Windows raw mouse input in a separate thread (https://github.com/libsdl-org/SDL/issues/8756).  I believe it should be possible to hook directly into these events by installing an SDL event filter/watch, like we currently do for other specific events.

It probably isn't too hard to add RAWKEYBOARD handling to SDL.

---

> It may also be relevant on windows for key input, as we are not using raw input there (I've seen reports of some users hitting input delays on windows with similar blocking).

[`onWndProc`](https://github.com/ppy/osu-framework/blob/3f4f735aca06ce25fef3ca39cf8d4119846f3363/osu.Framework/Platform/Windows/WindowsMouseHandler.cs#L64) in `WindowsMouseHandler` is only called by [`SDL_PumpEvents()`](https://github.com/libsdl-org/SDL/blob/f52a12a8fe9a6d52dfbed90310a894ce6be865c3/src/video/windows/SDL_windowsevents.c#L1855-L1858). So it is still very much tied to the event loop.

Susko3:
SDL3 has added raw keyboard on windows: https://github.com/libsdl-org/SDL/pull/9363. This is starting to look like a really interesting update.

Susko3:
[...]

PR #6235, opened 2024-04-08
Async mouse input on windows

Susko3:
- Related: https://github.com/ppy/osu-framework/issues/6141
- Should fix: https://github.com/ppy/osu-framework/issues/6181

Depends on:
- [x] #6234

This PR simply handles relative mouse and keyboard events in the `SDL_EventFilter` i.e. `HandleEventFromFilter`.

This works because SDL queries raw mouse and keyboard inputs on a separate thread. Our `SDL_EventFilter` will intercept those events before they are pushed to the internal SDL queue, handle them, and then request SDL to drop them.

## Testing

Apply the diff and toggle relative mouse mode in `TestSceneInputManger`.
This isn't a perfect simulation of `SDL_PumpEvents()` lag, but it shows the async input in action.

```diff
diff --git a/osu.Framework/Platform/SDL3Window.cs b/osu.Framework/Platform/SDL3Window.cs
index aac3baf44..737547203 100644
--- a/osu.Framework/Platform/SDL3Window.cs
+++ b/osu.Framework/Platform/SDL3Window.cs
@@ -460,6 +460,7 @@ private void setSDLIcon(Image<Rgba32> image)
         private void pollSDLEvents()
         {
             SDL3.SDL_PumpEvents();
+            Thread.Sleep(100);
 
             int eventsRead;
 ```

### CODE REVIEW

**Comment by peppy** [comment]
This seems to break input on my VM setup. Yes I have a weird setup but it did work, so something has potentially changed or broken. I haven't yet investigated why, but can if required (if the issue isn't obvious).

master:

https://github.com/ppy/osu-framework/assets/191335/4790d1b2-2fe9-490b-98f4-e2291c77040f

this PR:

https://github.com/ppy/osu-framework/assets/191335/44ab0dd0-4436-4f7f-8aca-4a291e68f1c5

**Review by peppy — COMMENTED** [review]
As commented

**Comment by Susko3** [comment]
Hmm, it's weird that the relative mouse movements align with the cursor in one scenario but not the other. But I wouldn't call it broken, the framework cursor is still moving proportionally to your mouse movements.

The weirder thing is that the mouse cursor isn't hidden when in relative mode.

I have no idea what could cause any of this. But if you'll be looking at this, don't forget to enable event logging with
```cs
SDL3.SDL_SetHint(SDL3.SDL_HINT_EVENT_LOGGING, "2"u8);
```

peppy:
> the framework cursor is still moving proportionally to your mouse movements

I tried to show this, but it's actually not proportionally correct either. I'll see what I can find.

peppy:
@Susko3 can you please check back on this? a lot of upstream changes meant that I basically had to reapply your changes here.

I'm also curious when we should *not* be using the filter method for handling things. I guess only when we want thread synchronisation on the input thread?

peppy:
@Susko3 sounds good. i think i'd like to make moving keyboard input there a goal (because otherwise it will run into the same issues as mouse). did you want to make any follow up changes?

Susko3:
> @Susko3 sounds good. i think i'd like to make moving keyboard input there a goal (because otherwise it will run into the same issues as mouse).

Yep, we can keep the [original](https://github.com/ppy/osu-framework/issues/6141) [issues](https://github.com/ppy/osu-framework/issues/6181) open. Having only mouse input async will make the issue less obvious, as the mouse is smooth, but the keyboard is delayed. Keyboard input needs additional consideration because [SDL_HINT_WINDOWS_RAW_KEYBOARD](https://wiki.libsdl.org/SDL3/SDL_HINT_WINDOWS_RAW_KEYBOARD) is off by default.

> did you want to make any follow up changes?

I wanted to make `handleMouseButtonEvent()` / `mousePressedButtons` thread-safe, but it will conflict with [#6488](https://github.com/ppy/osu-framework/pull/6488/files#diff-6cedbb22afe703f28f56daa971b858e2374575d5bd98db05b2678aca43894f9b). Changes in that PR also make it easier to implement thread safety, so it'll be easier if that one goes in first.

I will try to use [`System.Threading.Interlocked`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked?view=net-9.0) as it's a hot path and it's just setting integer flags.

Susko3:
Mouse input looks good now. Quickly tested on windows and android, I found no issues.

I think keyboard input can be handled in the filter iff text input is disabled, this avoids issues with ordering of key and text events while still reducing latency during gameplay.

peppy (approved):

peppy:
Getting this in for more widespread testing.

Let's not forget about keyboard as well. If you're able to follow up with the changes required for that at some point it would be hugely appreciated.

PR #6239, opened 2024-04-12
Fix tablet info not clearing when unplugged

Joehuu:
After a tablet is plugged in for the first time, "No tablet detected!" and the faq link would never show again when no tablet is connected.

`tablet.Value` is potentially null as said in the xmldoc so made it that way, probably regressed somewhere. `TabletInfo` and `ITabletHandler`'s xmldoc says:
https://github.com/ppy/osu-framework/blob/2ec0642943c077e8ebfb1aa58e4bb9971faee640/osu.Framework/Input/Handlers/Tablet/ITabletHandler.cs#L11
but have no idea what the removal refactor would look like. Just to note because the changes are relevant to those classes/interfaces.

The gifs below show me plugging and unplugging two different tablets:

| Before | After |
| --- | --- |
| C2GBCp91tT | Dp2Z7jZhby |
| SYhHVwcKdE | S5HzhN4zeE |

### DIFF
--- a/osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
+++ b/osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
@@ -54,9 +54,9 @@ namespace osu.Framework.Input.Handlers.Tablet
 
             host.Window.Resized += () => updateOutputArea(host.Window);
 
-            AreaOffset.BindValueChanged(_ => updateInputArea(device));
-            AreaSize.BindValueChanged(_ => updateInputArea(device), true);
-            Rotation.BindValueChanged(_ => updateInputArea(device), true);
+            AreaOffset.BindValueChanged(_ => updateTabletAndInputArea(device));
+            AreaSize.BindValueChanged(_ => updateTabletAndInputArea(device), true);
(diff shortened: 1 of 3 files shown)


osu.Framework/Input/Handlers/Tablet/OpenTabletDriverHandler.cs
```diff
@@ -54,9 +54,9 @@ public override bool Initialize(GameHost host)
 
             host.Window.Resized += () => updateOutputArea(host.Window);
 
-            AreaOffset.BindValueChanged(_ => updateInputArea(device));
-            AreaSize.BindValueChanged(_ => updateInputArea(device), true);
-            Rotation.BindValueChanged(_ => updateInputArea(device), true);
+            AreaOffset.BindValueChanged(_ => updateTabletAndInputArea(device));
+            AreaSize.BindValueChanged(_ => updateTabletAndInputArea(device), true);
```
peppy, line 58:
```suggestion
            AreaSize.BindValueChanged(_ => updateTabletAndInputArea(device));
```

peppy (dismissed):
Seems correct, but haven't tested with a tablet.

bdach (approved):
tested with tablet

PR #6241, opened 2024-04-13
Make SDL3Clipboard support images

Speykious:
This PR supersedes https://github.com/ppy/osu-framework/pull/6151 which aimed at adding support for copying and pasting images from the clipboard on Linux, but had the wrong approach of calling into an external executable (`xclip` for X11).

In the mean time I had actually started a native clipboard library in Rust for applications that would support any mime types and be given arbitrary data (the X11 implementation was somewhat finished, but not Wayland), but since osu!framework now uses SDL3, I can just use `SDL_GetClipboardData` and `SDL_SetClipboardData` [as suggested back then by @Susko3](https://github.com/ppy/osu/discussions/16775#discussioncomment-8382238).

This is a first attempt at implementing this, and on my Linux machine, the visual tests for copying images are all green! :D
~~Edit: actually doesn't work every time! D: I'll have to really refactor this code and make sure I understand what's going on.~~
Edit 2: bug identified, it seems to come from the given image itself...

~~That said, I really don't know what to think of this code, [...]~~ I think the code looks ok now :)


Speykious:
There's also the caveat that I only implemented a way to copy PNGs. Wouldn't it also need to be able to copy JPGs depending on some setting? I know osu! lets you choose the image format for screenshot savings.

smoogipoo:
What you have looks pretty standard for the new age of unmanaged->managed callbacks. That being said, this PR now competes with https://github.com/ppy/osu-framework/pull/6223 and we should probably make sure that this implementation is going to easily lend itself to other mime types in the future.

osu.Framework/Platform/SDL/SDL3Clipboard.cs
```diff
@@ -18,12 +21,75 @@ public class SDL3Clipboard : Clipboard
 
         public override Image<TPixel>? GetImage<TPixel>()
         {
-            return null;
+            byte[] buffer;
+
+            unsafe
+            {
+                nuint size = 0;
+                byte* data = (byte*)SDL3.SDL_GetClipboardData(Encoding.UTF8.GetBytes("image/png"), &size);
```
FreezyLemon, line 29: Maybe SDL3-CS handles this automagically somehow, but I think you're missing a null terminator here. The `byte[]` returned by `Encoding.GetBytes` does not include a null byte at the end.

Speykious: Would `"image/png\0"u8` work better then?

FreezyLemon: Yeah, that should work

Speykious:
image
I debugged a lot today (both with a debugger and with prints) and concluded that `pngData` was fine as a static due to how the API works. It only calls for a cleanup once if we're requesting for an image again.

Speykious:
image
It seems that sometimes the test gives a completely empty screenshot. I was worried that my implementation had some weird undefined behavior thing, but it seems like that bug is happening elsewhere...?

osu.Framework/Platform/SDL/SDL3Clipboard.cs
Speykious: done!

Speykious:
image
It seems I'm missing something related to the unmanaged call convention

Susko3:
I'm testing this on windows (by commenting out the `WindowsClipboard` override) and I can't get images from osu!framework into external programs (and vice versa). The clipboard tests pass fine, but applications complain about there being no image in the clipboard. And trying to paste images into framework doesn't work: either I get the stale image (last one copied from framework) or no image at all (even when copying a sample PNG from the internet).

image

Does copying images to/from framework into external programs work on Linux?

Speykious:
It does work on Linux (at least on my machine on X11), as long as there is a clipboard manager present. If there is no clipboard manager, the selection will disappear once the application closes, which is expected behavior.

I don't know why it doesn't work on Windows though. :/

Susko3:
Works perfectly fine on windows if the mime type is changed to `image/bmp` and the image is saved with `.SaveAsBmp()`. Do bitmaps work on Linux too?

I wonder if it works on macOS (which currently uses `MacOSClipboard`) and iOS (which uses `SDL3Clipboard`).

FreezyLemon:
> Works perfectly fine on windows if the mime type is changed to `image/bmp` and the image is saved with `.SaveAsBmp()`. Do bitmaps work on Linux too?
> 
> I wonder if it works on macOS (which currently uses `MacOSClipboard`) and iOS (which uses `SDL3Clipboard`).

At least on Wayland, the mime type also has to match. If I e.g. copy a PNG from my local storage, the current code works fine. If I use a Paint-like program to scribble around, Ctrl-A and Ctrl-C, then it's also saved as an `image/bmp` and needs that exact mime type to be retrieved from SDL.

The obvious solution would be to just have a list of image mimetypes that are tried one after another. FWIW, [SDL itself also does this](https://github.com/libsdl-org/SDL/blob/d0819bcc5cfe852ad60deab4e1ff7d5598f0b983/src/video/SDL_clipboard.c#L298-L304) for the built-in `GetClipboardText` function.

Speykious:
I'll try to implement a multi-mime-types solution.

Susko3:
We should probably ask ImageSharp about it's supported decoder formats, see https://github.com/SixLabors/ImageSharp/discussions/1353 (but with [`IImageFormat.MimeTypes`](https://docs.sixlabors.com/api/ImageSharp/SixLabors.ImageSharp.Formats.IImageFormat.html#SixLabors_ImageSharp_Formats_IImageFormat_MimeTypes)).

This is what I get on windows:
```
image/png
image/apng
image/jpeg
image/pjpeg
image/gif
image/bmp
image/x-windows-bmp
image/x-portable-pixmap
image/x-portable-anymap
image/x-tga
image/x-targa
image/tiff
image/tiff-fx
image/webp
image/qoi
image/x-qoi
image/vnd.qoi
```

osu.Framework/Platform/SDL/SDL3Clipboard.cs
```diff
@@ -18,12 +21,87 @@ public class SDL3Clipboard : Clipboard
...
+                    return null;
+
+                nuint size = 0;
+                byte* data = (byte*)SDL3.SDL_GetClipboardData("image/png\0"u8, &size);
+                if (data == null)
+                    return null;
+
+                buffer = new byte[size];
+                for (nuint i = 0; i < size; i++)
+                    buffer[i] = data[i];
```
Susko3, line 41: This should use [`Marshal.Copy`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.copy?view=net-8.0#system-runtime-interopservices-marshal-copy(system-intptr-system-byte()-system-int32-system-int32)).

osu.Framework/Platform/SDL/SDL3Clipboard.cs
```diff
@@ -18,12 +21,87 @@ public class SDL3Clipboard : Clipboard
...
+            }
+
+            return Image.Load<TPixel>(buffer);
+        }
+
+        private unsafe struct ClipboardImageData
+        {
+            public byte* RawBuffer;
+            public nuint Length;
         }
```
Susko3, line 53: Can probably be done much cleaner with a `GCHandleType.Pinned` `ObjectHandle<byte[]>`.

Similar to:
https://github.com/ppy/osu-framework/blob/32e7571aa858c4d055f45cc487c08a11aab64817/osu.Framework/Audio/Sample/SampleBassFactory.cs#L84-L85

osu.Framework/Platform/SDL/SDL3Clipboard.cs
```diff
@@ -18,12 +21,87 @@ public class SDL3Clipboard : Clipboard
...
+        }
+
+        [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvCdecl) })]
+        private static unsafe nint clipboardDataCallback(nint userdata, byte* mimeType, nuint* length)
+        {
+            ClipboardImageData* pngData = (ClipboardImageData*)userdata;
+
+            fixed (byte* pngMimetypeStr = "image/png\0"u8)
+            {
+                if (SDL3.SDL_strcmp(mimeType, pngMimetypeStr) == 0)
```
Susko3, line 88: There is no reason to delegate string comparison to SDL. Better to decode the pointer with `SDL3.PtrToStringUTF8` and use `==` on C# `string`s.

Susko3 (changes requested):
I'd rather rewrite this PR then to try and fix it.

osu.Framework/Platform/SDL/SDL3Clipboard.cs
Speykious: By using `SDL_strcmp`, I don't need to copy any data over before doing the comparison.

osu.Framework/Platform/SDL/SDL3Clipboard.cs
Speykious: Passing around a `byte[]` means the copying will happen in the callback and not before.
But I might end up using an `ObjectHandle` anyways to support multiple mime types. Only this time it'll need to be the `Image` itself.

Speykious: This is trickier than I expected.
In the `SampleBassFactory` code, the handle is created with `using`, so it gets freed immediately after `SampleLoad` gets called. This is fine because `SampleLoad` makes a copy of the data immediately - from its documentation:
> *There is no need to pin the memory buffer for this method, since after loading a sample from memory, the memory can safely be discarded, as a copy is made.*
But I can't just free the `ObjectHandle` immediately after `SDL_SetClipboardData` gets called. Instead I have to free it inside of the `cleanup` callback, which will be called with the old `userdata` before each new call to `SDL_SetClipboardData`.
And unfortunately, I can't just do it manually with `.Dispose()` either, because disposing an object handle obtained from a `nint` won't actually free the object handle, we can only free it on the original one.

I feel like now I need to have two swapping global variables (due to how the cleanup logic works), which would make an object handle unnecessary...

osu.Framework/Platform/SDL/SDL3Clipboard.cs
Speykious: I ended up doing it anyway to be able to use `ImageFormatManager.TryFindFormatByMimeType` :)

osu.Framework/Platform/SDL/SDL3Clipboard.cs
Speykious: > I feel like now I need to have two swapping global variables

Welp, this is what I have currently

Speykious:
> Works perfectly fine on windows if the mime type is changed to `image/bmp` and the image is saved with `.SaveAsBmp()`.

@Susko3 Can you tell me if this works on Windows now?

Speykious:
In any case, I can confirm that it works on MacOS.
image
image

Susko3:
Thanks for the PR,

Issue #6224, opened 2024-03-24
Masking sub-tree creation can be avoided with relative ease


EVAST9919:
Also adresses https://github.com/ppy/osu-framework/issues/1432

While `ComputeIsMaskedAway` isn't *that* expensive, iterating over children in complex scenes adds up. And we do that many times per composite. I tried to move all the masking computations into `UpdateSubTree` and it works fine (or I was lucky and didn't encounter any issues so far) and we get performance benefits.
I understand though that this is not as easy as it sounds since it affects the whole update system and benefits are not that big, but may be for the future it's something to consider. Or not.

[framework diff](https://github.com/ppy/osu-framework/compare/master...EVAST9919:masking-subtree?expand=1)
[osu-side diff](https://github.com/ppy/osu/compare/master...EVAST9919:masking-subtree?expand=1)
(I'll pr if that's a direction we want to go in)

||master|proposal|
|---|---|---|
|direct (600 items loaded)|![direct-master](https://github.com/ppy/osu-framework/assets/22874522/96721066-3e3f-4687-be92-8c2e3fabf63d)|![direct-pr](https://github.com/ppy/osu-framework/assets/22874522/5f666905-b209-46c9-93ff-63f33fa2db6a)|
|mods-overlay|![mods-master](https://github.com/ppy/osu-framework/assets/22874522/3c17961c-fbe7-471b-9a0c-d229cc7c2eb4)|![mods-pr](https://github.com/ppy/osu-framework/assets/22874522/9b8fb691-9739-4e61-b5d7-e9d1456a1e12)|
|song-select + direct + beatmap overlay|![overlays-master](https://github.com/ppy/osu-framework/assets/22874522/aef30d61-fe1f-4cfd-9217-889a314927b7)|![overlays-pr](https://github.com/ppy/osu-framework/assets/22874522/60633bd7-3004-48b7-afd9-fe508737aea5)|
|song-select|![ss-master](https://github.com/ppy/osu-framework/assets/22874522/21425c4d-ff64-4423-9e54-7ff1c0acc7f9)|![ss-pr](https://github.com/ppy/osu-framework/assets/22874522/c2606e8a-0420-4412-8aac-7abcdcea0f1f)|

smoogipoo:
I don't remember/think there's a particular reason why it's a separate pass.

EVAST9919:
Okay, there's a reason for that: that's to ensure all the updates/invalidations/whatever have taken place.
[...]

PR #6243, opened 2024-04-14
Avoid masking calculations on valid drawables

EVAST9919:
Not sure whether this selection of invalidation flags is sufficient or more than needed.

|master|pr|
|---|---|
|master|pr|

osu performance difference is similar to shown in https://github.com/ppy/osu-framework/issues/6224

### DIFF
--- a/osu.Framework/Graphics/Drawable.cs
+++ b/osu.Framework/Graphics/Drawable.cs
@@ -66,6 +66,7 @@ namespace osu.Framework.Graphics
             AddLayout(screenSpaceDrawQuadBacking);
             AddLayout(drawColourInfoBacking);
             AddLayout(requiredParentSizeToFitBacking);
+            AddLayout(maskingBacking);
         }
 
         private static readonly GlobalStatistic<int> total_count = GlobalStatistics.Get<int>(nameof(Drawable), "Total constructed");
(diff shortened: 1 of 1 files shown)


osu.Framework/Graphics/Drawable.cs
```diff
@@ -510,8 +511,11 @@ public virtual bool UpdateSubTreeMasking(Drawable source, RectangleF maskingBoun
             if (HasProxy && source != proxy)
                 return false;
 
-            if (!maskingBacking.IsValid)
-                IsMaskedAway = maskingBacking.Value = ComputeIsMaskedAway(maskingBounds);
+            if (!screenSpaceDrawQuadBacking.IsValid || lastMaskingBounds != maskingBounds)
```
smoogipoo, line 514: This is not correct. The SSDQ can be validated by something else, such as autosize.

smoogipoo (changes requested):

osu.Framework/Graphics/Drawable.cs
smoogipoo: Here's a test (`TestSceneDrawableLayout`):

```csharp
[Test]
public void TestIsMaskedAwayUpdatedIndependentOfScreenSpaceDrawQuad()
{
    TestBox1 box = null;

    AddStep("create test", () =>
    {
        Child = new Container
        {
            RelativeSizeAxes = Axes.Both,
            Masking = true,
            Child = box = new TestBox1
            {
                Size = new Vector2(100, 100),
                Position = new Vector2(10000, 10000)
            }
        };
    });

    AddUntilStep("masked away", () => box.IsMaskedAway);

    AddStep("move onto screen and validate immediately", () =>
    {
        box.Position = Vector2.Zero;
        _ = box.ScreenSpaceDrawQuad; // Validate everything.
    });

    AddUntilStep("not masked away", () => !box.IsMaskedAway);
}
```

Please add tests, especially when you break cases like this.

PR #6252, opened 2024-04-17
Fix unnecessary array copying in `WindowsClipboard.SetImage()`

Susko3:
The old code of `SetImage()` would:

1. Allocate the necessary memory with `MemoryStream`
2. Copy that to a temporary array
3. Skip some bytes with LINQ and copy that to a new array
4. Copy that array into unmanaged memory with `Marshal.AllocHGlobal()`
5. memcopy that memory in a new `GlobalAlloc()` array

The new code skips steps 2 and 3.

### DIFF
--- a/osu.Framework/Platform/Windows/WindowsClipboard.cs
+++ b/osu.Framework/Platform/Windows/WindowsClipboard.cs
@@ -94,16 +94,11 @@ namespace osu.Framework.Platform.Windows
             {
                 var encoder = image.Configuration.ImageFormatsManager.GetEncoder(BmpFormat.Instance);
                 image.Save(stream, encoder);
-                var span = new ReadOnlySpan<byte>(stream.GetBuffer(), bitmap_file_header_length, (int)stream.Length - bitmap_file_header_length);
-                IntPtr unmanagedPointer = Marshal.AllocHGlobal(span.Length);
 
-                unsafe
-                {
-                    fixed (void* source = span)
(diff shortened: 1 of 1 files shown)


osu.Framework/Platform/Windows/WindowsClipboard.cs
```diff
@@ -91,19 +90,21 @@ public override void SetText(string text)
...
-                array = stream.ToArray().Skip(bitmap_file_header_length).ToArray();
-            }
+                var span = new ReadOnlySpan<byte>(stream.GetBuffer(), bitmap_file_header_length, (int)stream.Length - bitmap_file_header_length);
+                IntPtr unmanagedPointer = Marshal.AllocHGlobal(span.Length);
 
-            IntPtr unmanagedPointer = Marshal.AllocHGlobal(array.Length);
-            Marshal.Copy(array, 0, unmanagedPointer, array.Length);
+                unsafe
+                {
+                    fixed (void* source = span)
```
Rekkonnect, line 102: I think that fixing the span is unnecessary here. Seeing how Microsoft does their own marshal copy implementation ([link](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs#L284)), both the unsafe and the fixed blocks can be removed.

smoogipoo (approved):

PR #6253, opened 2024-04-18
Fix flaky test attribute not working

smoogipoo:
There's a few issues here:

1. We were throwing arbitrary exceptions, not `AssertionException`s. The latter are handled specially by NUnit to allow retries.
2. If an assertion happens inside `[TearDown]`, it does not re-run the test at all.

Using `TestActionAttribute` allows us to do basically the same thing without running into the second issue. Note however that this does not do anything like trying to recover the `GameHost` after a non-assertion exception, so those will still fail immediately and not retry. I think that this is fine and "intentional".

Until steps now throw an `AssertionException` instead of a `TimeoutException`. I don't believe we rely on this anywhere. Messaging has also changed a bit:

```
Previous:

TearDown : System.TimeoutException : "aaaaa" timed out
--TearDown
   at osu.Framework.Testing.Drawables.Steps.UntilStepButton.<>c__DisplayClass11_0.<.ctor>b__0() in /Users/smgi/Repos/osu-framework/osu.Framework/Testing/Drawables/Steps/UntilStepButton.cs:line 65

Now:

"aaaaa" timed out
   at osu.Framework.Testing.Drawables.Steps.UntilStepButton.<>c__DisplayClass11_0.<.ctor>b__0() in /Users/smgi/Repos/osu-framework/osu.Framework/Testing/Drawables/Steps/UntilStepButton.cs:line 66
```

I've run a full osu! test run with this without issue.

### DIFF
--- a/osu.Framework.Tests/FlakyTestAttribute.cs
+++ b/osu.Framework.Tests/FlakyTestAttribute.cs
@@ -1,7 +1,6 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System;
 using NUnit.Framework;
 
 namespace osu.Framework.Tests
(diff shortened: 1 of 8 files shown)


peppy (approved):
This looks promising.

PR #6254, opened 2024-04-19
Add support for deferred SDL3 clipboard callbacks (and use it for images)

Susko3:
- Alternative to / closes https://github.com/ppy/osu-framework/pull/6241

## Improvements on previous PR

- made in a generic way to support future expansions
  - e.g. for https://github.com/ppy/osu-framework/pull/6223
- data is not copied when crossing the unmanaged boundary
- `SDL_HasClipboardData` is checked before getting the data

## How it works

All the required data for the clipboard callback is saved to a `ClipboardCallbackContext` object. An `ObjectHandle` is allocated for that object to keep it alive when it's owned by SDL (ownership is transferred upon successfully calling `SDL_SetClipboardData`).

When another application (or we) request clipboard data, the clipboard content is generated and the resulting `ReadOnlyMemory<byte>` is pinned (so the GC doesn't move it) and its pointer is passed to SDL.

When new clipboard data is set, our cleanup callback gets called so we unpin the `ReadOnlyMemory` and free the `ObjectHandle`.

## Tested on
- [x] Windows (with BMP format, still uses `WindowsClipbard`)
- [x] macOS (still uses `MacOSClipboard`)
- [x] X
- [x] Wayland

Please make sure that `TestSceneClipboard` passes and that copying and pasting images to/from external apps also works.

[iOS](https://github.com/libsdl-org/SDL/blob/fb4c15319ec61d963975b47dd1ec21fae184bc04/src/video/uikit/SDL_uikitclipboard.m) and [Android](https://github.com/libsdl-org/SDL/blob/fb4c15319ec61d963975b47dd1ec21fae184bc04/src/video/android/SDL_androidclipboard.c) are not supported (clipboard text only)


Speykious:
The `Debug.Assert` fails on my Linux machine.
image

This might not be due to not supporting the image format though. Here, `dataCallback` wrongly assumes that it'll only get called once with the requested mimetype, but on Linux it gets called multiple times with a different mimetype each time, even though we pass an array of mimetypes with only one element in it as an argument to `SDL_SetClipboardData`.

I'd recommend integrating the logic I wrote in [my latest commit](https://github.com/Speykious/osu-framework/blob/fe89a691e965558979373be1683830df71374a40/osu.Framework/Platform/SDL/SDL3Clipboard.cs#L102) to solve this issue.
Apparently the image could be dropped at any time since we don't own it, so the only solution I see for this is to copy the `Image` itself in `SetImage` and pass that to the callback context.

FreezyLemon:
Seems to fail for BMP files on Wayland:

<details>

    [runtime] 2024-04-19 20:01:01 [verbose]: SDL error log [debug]: Pipe timeout
    [runtime] 2024-04-19 20:01:01 [verbose]: Failed to get SDL clipboard data for image/png. SDL error: Pipe timeout
    [runtime] 2024-04-19 20:01:01 [verbose]: SDL error log [debug]: Pipe timeout
    [runtime] 2024-04-19 20:01:01 [verbose]: Failed to get SDL clipboard data for image/jpeg. SDL error: Pipe timeout
    [runtime] 2024-04-19 20:01:01 [verbose]: SDL error log [debug]: Pipe timeout
    [runtime] 2024-04-19 20:01:01 [verbose]: Failed to get SDL clipboard data for image/bmp. SDL error: Pipe timeout
    [runtime] 2024-04-19 20:01:01 [verbose]: SDL error log [debug]: Pipe timeout
    [runtime] 2024-04-19 20:01:01 [verbose]: Failed to get SDL clipboard data for image/tiff. SDL error: Pipe timeout
    [runtime] 2024-04-19 20:01:01 [verbose]: SDL error log [debug]: Pipe timeout
    [runtime] 2024-04-19 20:01:01 [verbose]: Failed to get SDL clipboard data for image/webp. SDL error: Pipe timeout
    [runtime] 2024-04-19 20:01:01 [verbose]: SDL error log [debug]: Pipe timeout
    [runtime] 2024-04-19 20:01:01 [verbose]: Failed to get SDL clipboard data for image/qoi. SDL error: Pipe timeout
    [runtime] 2024-04-19 20:01:01 [error]: Step "retrieve image from clipboard" SingleStepButton triggered an error
    [runtime] 2024-04-19 20:01:01 [error]: System.NullReferenceException: Object reference not set to an instance of an object.
    [runtime] 2024-04-19 20:01:01 [error]: at osu.Framework.Tests.Visual.Platform.TestSceneClipboard.<TestImage>b__15_5() in /home/jannik/Development/osu-framework/osu.Framework.Tests/Visual/Platform/TestSceneClipboard.cs:line 73
    [runtime] 2024-04-19 20:01:01 [error]: at osu.Framework.Testing.Drawables.Steps.SingleStepButton.<.ctor>b__1_0() in /home/jannik/Development/osu-framework/osu.Framework/Testing/Drawables/Steps/SingleStepButton.cs:line 19
    [runtime] 2024-04-19 20:01:01 [error]: at osu.Framework.Testing.Drawables.Steps.StepButton.PerformStep(Boolean userTriggered) in /home/jannik/Development/osu-framework/osu.Framework/Testing/Drawables/Steps/StepButton.cs:line 124
    [runtime] 2024-04-19 20:01:01 [error]: at osu.Framework.Testing.Drawables.Steps.StepButton.OnClick(ClickEvent e) in /home/jannik/Development/osu-framework/osu.Framework/Testing/Drawables/Steps/StepButton.cs:line 98
    [runtime] 2024-04-19 20:01:01 [debug]: ClickEvent(Left) handled by "retrieve image from clipboard" SingleStepButton.
    [runtime] 2024-04-19 20:01:01 [debug]: MouseClick handled by "retrieve image from clipboard" SingleStepButton.

</details>

Susko3:
> Apparently the image could be dropped at any time since we don't own it, so the only solution I see for this is to copy the `Image` itself in `SetImage` and pass that to the callback context.

If we're copying the image, we're going to allocate memory (and might do some processing to change the pixel format) -- might as well save it to the clipboard directly.

Speykious:
> If we're copying the image, we're going to allocate memory (and might do some processing to change the pixel format) -- might as well save it to the clipboard directly.

I think I was for whatever reason thinking about returning the image in multiple different formats at the same time, but I guess it's unnecessary...

Speykious:
That being said, it still fails for me. :(
image

Susko3:
@FreezyLemon your logs are a bit concerning. `SDL_HasClipboardData` is returning true, but then getting the data fails. Of course, it's not guaranteed that if `SDL_HasClipboardData == true` then `SDL_GetClipboardData` will always have some data (TOCTOU, etc.), but it failing in such a simple case warrants a bug report upstream.

Susko3:
> That being said, it still fails for me. :(

Right, the callback can get called multiple times. SDL internally copies the data on windows.

FreezyLemon:
> @FreezyLemon your logs are a bit concerning. `SDL_HasClipboardData` is returning true, but then getting the data fails. Of course, it's not guaranteed that if `SDL_HasClipboardData == true` then `SDL_GetClipboardData` will always have some data (TOCTOU, etc.), but it failing in such a simple case warrants a bug report upstream.

The pipe timeout does feel like an upstream bug, yeah. I'll try and create a repro and then report it.

osu.Framework/Allocation/ObjectHandle.cs
```diff
@@ -83,6 +83,12 @@ public bool GetTarget(out T target)
             return false;
         }
 
+        public void FreeUnsafe()
+        {
+            if (handle.IsAllocated)
+                handle.Free();
+        }
```
bdach, line 90: Why does this exist, and as public no less? Should be internal at the least. Or gone.

bdach:
> X11

Seems to work fine on X.

osu.Framework/Allocation/ObjectHandle.cs
Speykious: This PR is using an `ObjectHandle` and SDL3 needs to own it so that we can free it in the `cleanupCallback`, so I guess this was added to do that in an unsafe context with a handle obtained from a `nint`.
I wonder what you think of how I implemented it in the previous PR, which instead had two managed global variables that swap places from one call to the next.

bdach: > which instead had two managed global variables that swap places from one call to the next.

no no no no nope nope nope _nooooooope_.

Speykious: I mean, both solutions work. I wouldn't have done that if I wasn't sure that `SDL_SetClipboardData` worked by calling the cleanup callback once before calling the data callback multiple times. I don't really see the problem with my "hacky" implementation here. I do kinda see the problem with having a `FreeUnsafe` function available at all in `ObjectHandle` though.

bdach: if it's internal i don't have a problem with it anymore. globals are a much bigger problem than a single method, regardless of whether internal or public even. when i read "two alternating globals" i translate it to "this is gonna be buggy" in my head.

bdach: If you can figure something out then sure, but I'd be fine with just the access modifier change.

Susko3: I was thinking something like this:

```diff
diff --git a/osu.Framework/Allocation/ObjectHandle.cs b/osu.Framework/Allocation/ObjectHandle.cs
index 47d7548af..242677f01 100644
--- a/osu.Framework/Allocation/ObjectHandle.cs
+++ b/osu.Framework/Allocation/ObjectHandle.cs
@@ -29,7 +29,7 @@ public struct ObjectHandle<T> : IDisposable
 
         private GCHandle handle;
 
-        private readonly bool fromPointer;
+        private readonly bool canFree;
 
         /// <summary>
         /// Wraps the provided object with a <see cref="GCHandle" />, using the given <see cref="GCHandleType" />.
@@ -39,7 +39,7 @@ public struct ObjectHandle<T> : IDisposable
         public ObjectHandle(T target, GCHandleType handleType)
         {
             handle = GCHandle.Alloc(target, handleType);
-            fromPointer = false;
+            canFree = true;
         }
 
         /// <summary>
@@ -50,7 +50,19 @@ public ObjectHandle(T target, GCHandleType handleType)
         public ObjectHandle(IntPtr handle)
         {
             this.handle = GCHandle.FromIntPtr(handle);
-            fromPointer = true;
+            canFree = false;
+        }
+
+        /// <summary>
+        /// Recreates an <see cref="ObjectHandle{T}" /> based on the passed <see cref="IntPtr" />.
+        /// If <paramref name="ownsHandle"/> is <c>true</c>, disposing this object will free the handle.
+        /// </summary>
+        /// <param name="handle">Handle.</param>
+        /// <param name="ownsHandle">Whether this instance owns the underlying <see cref="GCHandle"/>. This should always be <c>true</c>.</param>
+        public ObjectHandle(IntPtr handle, bool ownsHandle)
+        {
+            this.handle = GCHandle.FromIntPtr(handle);
+            canFree = ownsHandle;
         }
 
         /// <summary>
@@ -83,17 +95,11 @@ public bool GetTarget(out T target)
             return false;
         }
 
-        public void FreeUnsafe()
-        {
-            if (handle.IsAllocated)
-                handle.Free();
-        }
-
         #region IDisposable Support
 
         public void Dispose()
         {
-            if (!fromPointer && handle.IsAllocated)
+            if (canFree && handle.IsAllocated)
                 handle.Free();
         }
 
diff --git a/osu.Framework/Platform/SDL/SDL3Clipboard.cs b/osu.Framework/Platform/SDL/SDL3Clipboard.cs
index 74ec8efac..5bc841684 100644
--- a/osu.Framework/Platform/SDL/SDL3Clipboard.cs
+++ b/osu.Framework/Platform/SDL/SDL3Clipboard.cs
@@ -142,7 +142,7 @@ private static unsafe bool trySetData(string mimeType, Func<ReadOnlyMemory<byte>
         [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvCdecl) })]
         private static unsafe IntPtr dataCallback(IntPtr userdata, byte* mimeType, UIntPtr* length)
         {
-            var objectHandle = new ObjectHandle<ClipboardCallbackContext>(userdata);
+            var objectHandle = new ObjectHandle<ClipboardCallbackContext>(userdata, true);
 
             if (!objectHandle.GetTarget(out var context) || context.MimeType != SDL3.PtrToStringUTF8(mimeType))
             {
@@ -158,12 +158,12 @@ private static unsafe IntPtr dataCallback(IntPtr userdata, byte* mimeType, UIntP
         [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvCdecl) })]
         private static void cleanupCallback(IntPtr userdata)
         {
-            var objectHandle = new ObjectHandle<ClipboardCallbackContext>(userdata);
+            var objectHandle = new ObjectHandle<ClipboardCallbackContext>(userdata, true);
 
             if (objectHandle.GetTarget(out var context))
             {
                 context.Dispose();
-                objectHandle.FreeUnsafe();
+                objectHandle.Dispose();
             }
         }
 
```

bdach: Sure I'd be okay with that I think.

Susko3: I'm against only changing the access modifier, as this just a different way of using `ObjectHandle`. Nothing is inherently "internal" to the framework about that.

osu.Framework/Allocation/ObjectHandle.cs
```diff
@@ -50,7 +50,19 @@ public ObjectHandle(T target, GCHandleType handleType)
...
-            fromPointer = true;
+            canFree = false;
+        }
+
+        /// <summary>
+        /// Recreates an <see cref="ObjectHandle{T}" /> based on the passed <see cref="IntPtr" />.
+        /// If <paramref name="ownsHandle"/> is <c>true</c>, disposing this object will free the handle.
+        /// </summary>
+        /// <param name="handle">Handle.</param>
+        /// <param name="ownsHandle">Whether this instance owns the underlying <see cref="GCHandle"/>. This should always be <c>true</c>.</param>
```
bdach, line 61: Upon reading more carefully (I only skimmed the gist of that patch before) I'm a bit baffled by this xmldoc; if "this should always be `true`", then why does the param exist? Like it's a bit weird...

Either let the consumer specify this always and remove the bool param, or make these like static methods (`ObjectHandle.Owned()`, `ObjectHandle.NonOwned()`) or something. This feels a bit like winapi-tier design ("you must put `0xDEADBEEF` in this field because we tell you so").

bdach (approved):
Not sure about whatever that wayland breakage but I don't really personally much care for wayland at this point. Broken wm be broken. Let upstream deal with it.

Probably needs that macOS check before merge though.

Susko3:
macOS should have [full support for all clipboard data](https://github.com/libsdl-org/SDL/blob/cfabb7a9acfdfe11aedabdba9acdcc893709f2cd/src/video/cocoa/SDL_cocoaclipboard.m), but I'm mostly wondering which image formats are supported and compatible with other apps. Looking online, I see tiff, png and jpg mentioned.

We can always put multiple formats on the clipboard 😉 

```diff
diff --git a/osu.Framework/Platform/MacOS/MacOSGameHost.cs b/osu.Framework/Platform/MacOS/MacOSGameHost.cs
index 0cb6f2362..45657c24a 100644
--- a/osu.Framework/Platform/MacOS/MacOSGameHost.cs
+++ b/osu.Framework/Platform/MacOS/MacOSGameHost.cs
@@ -11,6 +11,10 @@
 using osu.Framework.Input.Bindings;
 using osu.Framework.Input.Handlers;
 using osu.Framework.Input.Handlers.Mouse;
+using osu.Framework.Platform.SDL;
+using SixLabors.ImageSharp.Formats.Jpeg;
+using SixLabors.ImageSharp.Formats.Png;
+using SixLabors.ImageSharp.Formats.Tiff;
 
 namespace osu.Framework.Platform.MacOS
 {
@@ -35,7 +39,7 @@ public override IEnumerable<string> UserStoragePaths
             }
         }
 
-        protected override Clipboard CreateClipboard() => new MacOSClipboard();
+        protected override Clipboard CreateClipboard() => new SDL3Clipboard(TiffFormat.Instance /* or PngFormat, JpegFormat */);
 
         protected override ReadableKeyCombinationProvider CreateReadableKeyCombinationProvider() => new MacOSReadableKeyCombinationProvider();
 
```

osu.Framework/Platform/SDL/SDL3Clipboard.cs
```diff
@@ -17,12 +46,176 @@ public class SDL3Clipboard : Clipboard
...
+                }
+
+                return ret == 0;
+            }
+        }
+
+        [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvCdecl) })]
+        private static unsafe IntPtr dataCallback(IntPtr userdata, byte* mimeType, UIntPtr* length)
+        {
+            var objectHandle = new ObjectHandle<ClipboardCallbackContext>(userdata, true);
```
smoogipoo, line 145: We're passing `true` here which reads as if we should be cleaning this up, however that's not being done and appears incorrect to do. Should this be changed to `false` to prevent incorrect use?

osu.Framework/Platform/SDL/SDL3Clipboard.cs
```diff
@@ -17,12 +46,176 @@ public class SDL3Clipboard : Clipboard
...
+        [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvCdecl) })]
+        private static void cleanupCallback(IntPtr userdata)
+        {
+            var objectHandle = new ObjectHandle<ClipboardCallbackContext>(userdata, true);
+
+            if (objectHandle.GetTarget(out var context))
+            {
+                context.Dispose();
+                objectHandle.Dispose();
+            }
```
smoogipoo, line 167: Any reason to not do `using var objectHandle = ...` here and above (with the appropriate `false` parameter)?

I suppose there should be no case where `GetTarget()` fails to resolve but the `GCHandle` is still allocated, though it strikes me as a bit odd. Manually handling disposal like this.

osu.Framework/Platform/SDL/SDL3Clipboard.cs
```diff
@@ -17,12 +46,176 @@ public class SDL3Clipboard : Clipboard
...
+                    Debug.Assert(Address != IntPtr.Zero);
+                    Debug.Assert(DataLength != 0);
+                    return;
+                }
+
+                var data = dataProvider();
+                dataProvider = null!;
+                DataLength = (UIntPtr)data.Length;
+                memoryHandle = data.Pin();
+            }
```
smoogipoo, line 212: If this is pinned once and `null`ed afterwards, what's the purpose of this existing as a method as opposed to being done directly in the ctor and avoiding having to call this method/do the assertions?

osu.Framework/Platform/SDL/SDL3Clipboard.cs
```diff
@@ -17,12 +46,176 @@ public class SDL3Clipboard : Clipboard
...
+            /// <summary>
+            /// Address of the <see cref="ReadOnlyMemory{T}"/> returned by the <see cref="dataProvider"/>.
+            /// </summary>
+            /// <remarks>Pinned and suitable for passing to unmanaged code.</remarks>
+            public unsafe IntPtr Address => (IntPtr)memoryHandle.Pointer;
+
+            /// <summary>
+            /// Length of the <see cref="ReadOnlyMemory{T}"/> returned by the <see cref="dataProvider"/>.
+            /// </summary>
+            public UIntPtr DataLength;
```
smoogipoo, line 191: How about `{ private set; }`?

smoogipoo (commented):
Have tested on Wayland

smoogipoo:
On both Wayland and macOS, if I have something already in the clipboard, `SetClipboardText("")` doesn't clear it. The return code is `0` and there's no error in `SDL_GetError()`.

`SDL_ClearClipboardData()` also seems to do nothing in this case, I assume that's because it's clearing data and not text.

Looks like an SDL issue.

osu.Framework/Platform/SDL/SDL3Clipboard.cs
Susko3: The idea is to make use of deferred rendering of clipboard content. If no application requests the clipboard data, this code never runs and we don't need to serialise it to bytes.

This isn't of much use in this PR as the image is immediately serialised.

Susko3:
> On both Wayland and macOS, if I have something already in the clipboard, `SetClipboardText("")` doesn't clear it. The return code is `0` and there's no error in `SDL_GetError()`.
> 
> `SDL_ClearClipboardData()` also seems to do nothing in this case, I assume that's because it's clearing data and not text.
> 
> Looks like an SDL issue.

Definitely an SDL issue as `SetClipboardText()` is just shorthand for `SDL_SetClipboardData` with `text/plain` mime type. Plus, [calling `SetClipboardText(NULL)` or `SetClipboardText("")` actually calls `SDL_ClearClipboardData()`](
https://github.com/libsdl-org/SDL/blob/9245084c0677b605d59502aaa0a2c31c23d78ab0/src/video/SDL_clipboard.c#L267-L283)

smoogipoo (approved):
Let's give this a shot. Leaving MacOSGameHost as is right now, even though it appears to work.

PR #6255, opened 2024-04-19
FFmpeg build: Update to 7.0

FreezyLemon:
[Build artifacts on my fork](https://github.com/FreezyLemon/osu-framework/actions/runs/9085440627). For comparison: [Last FFmpeg build on main repo](https://github.com/ppy/osu-framework/actions/runs/8432434257)

Binary sizes increase a bit on all platforms, can't do much to avoid that in this case.

Changes:
- Update FFMPEG_VERSION to "7.0"
- Remove old workarounds (fixed upstream)
- Add Android and iOS builds
- Address a few misc. things

PR including the .NET-side migration: #6256

**Files that should be manually removed when updating binaries**:
- Old library versions (e.g. `libswscale.so.5` is replaced by `libswscale.so.8`)
- Unneeded Android files:
  - `libavdevice`
  - `libavfilter`
  - `libcpufeatures`
  - `libc++_shared`
  - `libmobileffmpeg*`
  - `libswresample`
- Unneeded iOS files:
  - `libavdevice`
  - `libswresample`

Note for later: There are some new hwaccels (vulkan, d3d12, metal?)


FreezyLemon:
Yes, I already noticed this when I realized (in the other PR) that I completely forgot about mobile platforms. Even if ffmpeg-kit did have 7.0 releases out now, I still think it would make sense to do the build in here. Partly for control over the binaries and partly for size optimizations

FreezyLemon:
I have some experimental Android builds running. I'll still have to look at how the libraries are loaded etc. to see if things need to be changed on that side.

What targets are supported on Android? FFmpeg allows building for x86, x86_64, and multiple ARM targets.

Also, should these mobile libraries remain in the `osu.Framework.Android`/`osu.Framework.iOS` projects? They could probably be moved to the `NativeLibs` project too.

FreezyLemon:
Added Android build. Looks good so far, but still needs testing (x86, x86_64 are probably not that important as they're most likely just used for emulators).

I enabled NEON in all cases since the minimum supported API level (21) [requires it](https://developer.android.com/ndk/guides/cpu-arm-neon):
> All ARMv8-based ("arm64") Android devices support Neon. Almost all ARMv7-based ("32-bit") Android devices support Neon, including all devices that shipped with API level 21 or later.

Raising the [API level](https://apilevels.com/) to 21 means only Android 5+ is supported. A build for API levels < 21 would need to be built without the NDK (or with a very old version) and would probably require more work to get running. But if this is a goal, I can try to do that.

smoogipoo:
21 is fine. It's what we specify in the props file anyway https://github.com/ppy/osu/blob/1f010c4bfd8236e817e1a13d9616827c97966db1/osu.Android.props#L3

FreezyLemon:
Added iOS build. Like the rest, needs testing. I tried to recreate the current `xcframework` files as closely as possible.

I looked at veldrid for the `CFBundleIdentifier` and chose `"sh.ppy.osu.Framework.iOS.$lib_name"`. So e.g. `sh.ppy.osu.Framework.iOS.libavcodec`.

frenzibyte (dismissed):
iOS build process needs to be revised, see [discord](https://discord.com/channels/188630481301012481/589331078574112768/1237055138041167915).

FreezyLemon:
After some coordination with @frenzibyte, I found [this line](https://github.com/arthenica/ffmpeg-kit/blob/e78b7956c7efd76e61930d9b3571773804c30565/scripts/apple/ffmpeg.sh#L602) in `ffmpeg-kit`'s build scripts. It is pretty non-obvious, but this modifies the `config.mak` file so that the install_name embedded in the binaries is compatible with framework bundles. The new patch file shows the exact change done by that `sed` invocation.

The binaries now seem to load, but the proper testing hasn't been done yet.

frenzibyte:
Have tested and reported in https://github.com/ppy/osu-framework/pull/6256#issuecomment-2111328342, this PR is good to go with regards to the iOS build scripts 👍

smoogipoo (approved):
Sorry for the wait, but good work on this. Let's see how it goes :)

PR #6257, opened 2024-04-20
Set IME rectangle before starting text input

Susko3:
Intended to fix on-screen keyboard overlapping text boxes on SDL3 Android (https://github.com/ppy/osu-framework/pull/6105#issuecomment-2066993789).

From [`SDL_SetTextInputRect`](https://wiki.libsdl.org/SDL3/SDL_SetTextInputRect) documentation:

> To start text input in a given location, this function is intended to be called before [SDL_StartTextInput](https://wiki.libsdl.org/SDL3/SDL_StartTextInput), although some platforms support moving the rectangle even while text input (and a composition) is active.

Both `TextInputSource.Activate()` and `EnsureActivated()` have the new IME rectangle parameter, but it's optional in `EnsureActivated()` as it's only intended to be set when changing focus to a new text box.

If a text box already has input focus, it's assumed that the last `SetImeRectangle()` call provided a more accurate rectangle, so `EnsureActivated()` is called without a rectangle. (SDL remembers the last used IME rectangle and re-uses it).

How it looks with SDL3 Android:

https://github.com/ppy/osu-framework/assets/16479013/d894c2dd-39b0-45a2-8043-ffb7d0aba8ac

### DIFF
--- a/osu.Framework/Graphics/UserInterface/TextBox.cs
+++ b/osu.Framework/Graphics/UserInterface/TextBox.cs
@@ -1362,9 +1362,9 @@ namespace osu.Framework.Graphics.UserInterface
             // We don't deactivate and activate, but instead keep text input active during the focus handoff, so that virtual keyboards on phones don't flicker.
 
             if (previousFocusWasTextBox)
-                textInput.EnsureActivated(AllowIme);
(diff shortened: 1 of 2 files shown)


Joehuu (approved):
Seems to work.

smoogipoo (approved):

PR #6260, opened 2024-04-20
Set SDL_HINT_APP_NAME in SDL3GameHost constructor

FreezyLemon:
Supersedes #5650. Closes ppy/osu#25783.

The [docs](https://wiki.libsdl.org/SDL3/SDL_HINT_APP_NAME) say that
> This hint should be set before SDL is initialized.

Since `SDL_Init` is called in the `SDL3Window` constructor, this means that the application name must be known at that point in time. So it's not possible to rely on the window title.

This appName value is then used to tell SDL the application name to show in OS windows, prompts, etc.:

image

The `appName` could arguably be used as the initial window title.

### DIFF
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -336,6 +336,11 @@ namespace osu.Framework.Platform
         {
             Options = options ?? new HostOptions();
 
+            if (string.IsNullOrEmpty(Options.FriendlyGameName))
+            {
+                Options.FriendlyGameName = $@"osu!framework (running ""{gameName}"")";
+            }
+
             Name = gameName;
 
             JsonConvert.DefaultSettings = () => new JsonSerializerSettings
--- a/osu.Framework/Platform/GameHost.cs
+++ b/osu.Framework/Platform/GameHost.cs
@@ -1031,7 +1031,7 @@ namespace osu.Framework.Platform
 
                 Window.SetupWindow(Config);
                 Window.Create();
-                Window.Title = $@"osu!framework (running ""{Name}"")";
+                Window.Title = Options.FriendlyGameName;
 
                 Renderer.Initialise(Window.GraphicsSurface);
 
--- a/osu.Framework.iOS/IOSGameHost.cs
+++ b/osu.Framework.iOS/IOSGameHost.cs
@@ -28,7 +28,7 @@ namespace osu.Framework.iOS
         {
         }
 
-        protected override IWindow CreateWindow(GraphicsSurfaceType preferredSurface) => new IOSWindow(preferredSurface);
+        protected override IWindow CreateWindow(GraphicsSurfaceType preferredSurface) => new IOSWindow(preferredSurface, Options.FriendlyGameName);
 
         protected override void SetupForRun()
         {
--- a/osu.Framework.iOS/IOSWindow.cs
+++ b/osu.Framework.iOS/IOSWindow.cs
@@ -31,8 +31,8 @@ namespace osu.Framework.iOS
             }
         }
 
-        public IOSWindow(GraphicsSurfaceType surfaceType)
-            : base(surfaceType)
+        public IOSWindow(GraphicsSurfaceType surfaceType, string appName)
+            : base(surfaceType, appName)
         {
         }
 
--- a/osu.Framework/HostOptions.cs
+++ b/osu.Framework/HostOptions.cs
@@ -33,5 +33,14 @@ namespace osu.Framework
         /// If the SDL_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR environment variable is set, this property will have no effect.
         /// </remarks>
         public bool BypassCompositor { get; set; } = true;
+
+        /// <summary>
+        /// The friendly name of the game to be hosted. This is used to display the name to the user,
+        /// for example in the window title bar or in OS windows and prompts.
+        /// </summary>
+        /// <remarks>
+        /// If empty, GameHost will choose a default name based on the gameName.
+        /// </remarks>
+        public string FriendlyGameName { get; set; } = string.Empty;
     }
 }
--- a/osu.Framework/Platform/Linux/LinuxGameHost.cs
+++ b/osu.Framework/Platform/Linux/LinuxGameHost.cs
@@ -33,7 +33,7 @@ namespace osu.Framework.Platform.Linux
             base.SetupForRun();
         }
 
-        protected override IWindow CreateWindow(GraphicsSurfaceType preferredSurface) => new SDL3DesktopWindow(preferredSurface);
+        protected override IWindow CreateWindow(GraphicsSurfaceType preferredSurface) => new SDL3DesktopWindow(preferredSurface, Options.FriendlyGameName);
 
         protected override ReadableKeyCombinationProvider CreateReadableKeyCombinationProvider() => new LinuxReadableKeyCombinationProvider();
 
--- a/osu.Framework/Platform/MacOS/MacOSGameHost.cs
+++ b/osu.Framework/Platform/MacOS/MacOSGameHost.cs
@@ -21,7 +21,7 @@ namespace osu.Framework.Platform.MacOS
         {
         }
 
-        protected override IWindow CreateWindow(GraphicsSurfaceType preferredSurface) => new MacOSWindow(preferredSurface);
+        protected override IWindow CreateWindow(GraphicsSurfaceType preferredSurface) => new MacOSWindow(preferredSurface, Options.FriendlyGameName);
 
         public override IEnumerable<string> UserStoragePaths
         {
--- a/osu.Framework/Platform/MacOS/MacOSWindow.cs
+++ b/osu.Framework/Platform/MacOS/MacOSWindow.cs
@@ -24,8 +24,8 @@ namespace osu.Framework.Platform.MacOS
         private IntPtr originalScrollWheel;
         private ScrollWheelDelegate scrollWheelHandler;
 
-        public MacOSWindow(GraphicsSurfaceType surfaceType)
-            : base(surfaceType)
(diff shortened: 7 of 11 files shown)


osu.Framework/Platform/GameHost.cs
```diff
@@ -1013,7 +1013,7 @@ protected void SetupRendererAndWindow(IRenderer renderer, GraphicsSurfaceType su
             try
             {
                 // Prepare window
-                Window = CreateWindow(surfaceType);
+                Window = CreateWindow(surfaceType, $@"osu!framework (running ""{Name}"")");
```
Susko3, line 1016: If you're not going to expose this publicly, might as well set the hint in the `SDL3GameHost` constructor and not do all this song and dance.

I guess the only problem with that approach is that `HeadlessGameHost` inherits from `SDL3GameHost`.

FreezyLemon: I always find it awkward to select a place to call these static functions that modify global state... I chose the `SDL3Window` because it seems to be a centralized place w.r.t. all the SDL functions.

OTOH, the application name isn't directly tied to the window, so putting it in GameHost does make more sense logically. Also shouldn't cause problems for headless AFAICT

Susko3: The only concern is that calling `SDL_SetHint` loads the SDL3 native library. Kinda weird to call the windowing library from headless tests, but the host inheritance is weirder.

This is what I'm thinking:

```diff
diff --git a/osu.Framework/Platform/SDL3GameHost.cs b/osu.Framework/Platform/SDL3GameHost.cs
index 345c06a13..337bf2861 100644
--- a/osu.Framework/Platform/SDL3GameHost.cs
+++ b/osu.Framework/Platform/SDL3GameHost.cs
@@ -11,6 +11,7 @@
 using osu.Framework.Input.Handlers.Tablet;
 using osu.Framework.Input.Handlers.Touch;
 using osu.Framework.Platform.SDL;
+using SDL;
 
 namespace osu.Framework.Platform
 {
@@ -21,6 +22,8 @@ public abstract class SDL3GameHost : GameHost
         protected SDL3GameHost(string gameName, HostOptions? options = null)
             : base(gameName, options)
         {
+            if (this is not HeadlessGameHost)
+                SDL3.SDL_SetHint(SDL3.SDL_HINT_APP_NAME, $@"osu!framework (running ""{Name}"")");
         }
 
         protected override TextInputSource CreateTextInput()
```

FreezyLemon: Seems sensible. I'll do that instead

osu.Framework/Platform/SDL3GameHost.cs
```diff
@@ -21,6 +22,8 @@ public abstract class SDL3GameHost : GameHost
         protected SDL3GameHost(string gameName, HostOptions? options = null)
             : base(gameName, options)
         {
+            if (this is not HeadlessGameHost)
+                SDL3.SDL_SetHint(SDL3.SDL_HINT_APP_NAME, $@"osu!framework (running ""{Name}"")");
```
Susko3, line 26: Hmm, we could easily allow consumers to change this in `HostOptions`. That way we can re-use this string for the initial window title. Plus, it's simple to have it default to `osu!framework (running "Name")`.

Not sure how'd you call this since `gameName` is already used. `FriendlyGameName`, `GameTitle`?

smoogipoo: I'd probably tend towards `FriendlyGameName` if anything. Though I question if there are too many names going around the place...
- `new GameHost(gameName, ...)`
- `Window.Title`
- `HostOptions.FriendlyGameName`

FreezyLemon: I like `FriendlyGameName`, it matches the existing `gameName` being more of an internal and/or technical value instead of the one to present to users.

How do you feel about this approach of passing it via `HostOptions`?

smoogipoo: I think it's fine. Given SDL's doc says "should be set before init", it's probably best that this is a once-off value like you have it.

osu.Framework/Platform/SDL3GameHost.cs
```diff
@@ -21,6 +22,8 @@ public abstract class SDL3GameHost : GameHost
         protected SDL3GameHost(string gameName, HostOptions? options = null)
             : base(gameName, options)
         {
+            if (this is not HeadlessGameHost)
+                SDL3.SDL_SetHint(SDL3.SDL_HINT_APP_NAME, Options.FriendlyGameName);
```
smoogipoo, line 26: How about moving this into `SDL3Window`? Perhaps by passing `appName` via ctor (or `HostOptions`...)

FreezyLemon: Interestingly, the PR originally did this: https://github.com/ppy/osu-framework/commit/3205a57ab58578a32e5a25afb0d1c2bb68884879

After feedback from Susko3, I changed it to move into the `SDL3GameHost` instead. See the first review for the related discussion

smoogipoo: My reasoning is primarily that if it needs to be set before `Init()` then it should probably be controlled by the thing that does the `Init()`.

@Susko3 how strongly do you feel about this being in `SDL3GameHost`?

Susko3: I'm fine with it being in `SDL3Window`. My main problem with the original approach was too much parameter passing in the `GameHost`s, but this is solved by `HostOptions`.

bdach:
I suppose this _works_, but my main question here is - what is the intent with respect to the game-side override of the window title (see [here](https://github.com/ppy/osu/blob/669b92a8e419ca06bbea4399d77d4a93bc1067fd/osu.Game/OsuGameBase.cs#L244-L248) and then [here](https://github.com/ppy/osu/blob/669b92a8e419ca06bbea4399d77d4a93bc1067fd/osu.Desktop/OsuGameDesktop.cs#L161))?

Like as far as I can tell, you can't actually _move_ that title logic out to use `FriendlyGameName` instead, because the logic is circular - the host needs to know the friendly game name _first_, but that name is not actually determined until the actual game instance to run on the host is constructed? Feels a bit circular imo.

Like it's either going to be this:

```diff
diff --git a/osu.Desktop/Program.cs b/osu.Desktop/Program.cs
index 29b05a402f..1dafe0d9ff 100644
--- a/osu.Desktop/Program.cs
+++ b/osu.Desktop/Program.cs
@@ -107,7 +107,11 @@ public static void Main(string[] args)
                 }
             }
 
-            using (DesktopGameHost host = Host.GetSuitableDesktopHost(gameName, new HostOptions { IPCPort = !tournamentClient ? OsuGame.IPC_PORT : null }))
+            using (DesktopGameHost host = Host.GetSuitableDesktopHost(gameName, new HostOptions
+                   {
+                       IPCPort = !tournamentClient ? OsuGame.IPC_PORT : null,
+                       FriendlyGameName = "osu!"
+                   }))
             {
                 if (!host.IsPrimaryInstance)
                 {

```

as a best-effort, or something like

```diff
diff --git a/osu.Desktop/Program.cs b/osu.Desktop/Program.cs
index 29b05a402f..d8364fc6e6 100644
--- a/osu.Desktop/Program.cs
+++ b/osu.Desktop/Program.cs
@@ -107,7 +107,11 @@ public static void Main(string[] args)
                 }
             }
 
-            using (DesktopGameHost host = Host.GetSuitableDesktopHost(gameName, new HostOptions { IPCPort = !tournamentClient ? OsuGame.IPC_PORT : null }))
+            using (DesktopGameHost host = Host.GetSuitableDesktopHost(gameName, new HostOptions
+                   {
+                       IPCPort = !tournamentClient ? OsuGame.IPC_PORT : null,
+                       FriendlyGameName = OsuGameBase.GAME_NAME,
+                   }))
             {
                 if (!host.IsPrimaryInstance)
                 {
diff --git a/osu.Game/OsuGameBase.cs b/osu.Game/OsuGameBase.cs
index fb7a238c46..6654b39137 100644
--- a/osu.Game/OsuGameBase.cs
+++ b/osu.Game/OsuGameBase.cs
@@ -75,6 +75,12 @@ public partial class OsuGameBase : Framework.Game, ICanAcceptFiles, IBeatSyncPro
     {
         public static readonly string[] VIDEO_EXTENSIONS = { ".mp4", ".mov", ".avi", ".flv", ".mpg", ".wmv", ".m4v" };
 
+#if DEBUG
+        public const string GAME_NAME = "osu! (development)";
+#else
+        public const string OSU_GAME_NAME = "osu!";
+#endif
+
         public const string OSU_PROTOCOL = "osu://";
 
         public const string CLIENT_STREAM_NAME = @"lazer";
@@ -241,11 +247,7 @@ public virtual string Version
 
         public OsuGameBase()
         {
-            Name = @"osu!";
-
-#if DEBUG
-            Name += " (development)";
-#endif
+            Name = GAME_NAME;
 
             allowableExceptions = UnhandledExceptionsBeforeCrash;
         }

```

I guess?

Food for thought as to whether things can be done better API-design-wise I guess. I'm probably fine with either resolution myself.

bdach (approved):

FreezyLemon:
Hm.. I see your point. Annoyingly, even if we ignore `Window.Title`, the SDL3GameHost needs to know the name at ctor-time, which is generally before `Framework.Game` gets created.

Unless SDL_Init is moved to a later point though, I don't think much can be done except working around it...

smoogipoo:
Both of those diffs look reasonable to me. As far I've convinced myself, its purpose is not the same as the window title so I wouldn't see us doing anything around that:
- `Window.Title`: Some dynamic text that contains contextual information.
- `FriendlyGameName`: Static text that identifies the app in system controls.
- `GetSuitableHost(gameName)`: Static text that identifies the app's runtime folder (`~/.local/share/{gameName}`).

smoogipoo (approved):

PR #6261, opened 2024-04-21
Update masking calculations without creating masking sub-tree

### LINKED ISSUE #1432: Add a method to bypass masking checks
Containers which have many (10k+) children and are guaranteed to never require masking checks can currently generate large overhead due to masking checks.

We should discuss a way to improve this, whether it means exposing a `protected bool RequiresMasking` to the user or otherwise.

EVAST9919:
Closes https://github.com/ppy/osu-framework/issues/1432 (we can't override `UpdateSubTreeMasking` to bypass masking checks since it no longer exists, so this became a requirement)
Improves upon proposal in https://github.com/ppy/osu-framework/issues/6224

Depends on:
- [ ] https://github.com/ppy/osu-framework/pull/6266

(Note that screenshot from `master` already contains https://github.com/ppy/osu-framework/pull/6243)
|master|pr|
|---|---|
|fr-master|fr-pr|

### DIFF
--- a/osu.Framework.Tests/Layout/TestSceneContainerLayout.cs
+++ b/osu.Framework.Tests/Layout/TestSceneContainerLayout.cs
@@ -7,6 +7,7 @@ using System;
 using NUnit.Framework;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Primitives;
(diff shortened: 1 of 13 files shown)


EVAST9919:
It will be easier to create a new pr rather than resolving conflicts.

PR #6263, opened 2024-04-22
Guard against unsupported window mode on startup

### LINKED ISSUE #6262: Platforms other than Windows crash when last windowed mode is borderless in latest master
Using macOS, but looking at the method, Windows is only implemented.

If you were using borderless before the SDL3 merge, it crashes after updating to latest master and you would have to manually set to `Windowed` / `Fullscreen`.

```
[runtime] 2024-04-22 17:42:58 [error]: An unhandled error has occurred.
[runtime] 2024-04-22 17:42:58 [error]: System.PlatformNotSupportedException: Operation is not supported on this platform.
[runtime] 2024-04-22 17:42:58 [error]: at osu.Framework.Platform.SDL3Window.SetBorderless(Display display)
[runtime] 2024-04-22 17:42:58 [error]: at osu.Framework.Platform.SDL3Window.UpdateWindowStateAndSize(WindowState state, Display display, DisplayMode displayMode)
[runtime] 2024-04-22 17:42:58 [error]: at osu.Framework.Platform.SDL3DesktopWindow.UpdateWindowStateAndSize(Wi
```

The following osu!-side lines can probably be removed if borderless is not a thing anymore for macOS:
https://github.com/ppy/osu/blob/308744348f8d7919d38b5508bb904a27b9909fad/osu.Game/OsuGame.cs#L855-L863
https://github.com/ppy/osu/blob/308744348f8d7919d38b5508bb904a27b9909fad/osu.Game/Overlays/Settings/Sections/Graphics/LayoutSettings.cs#L271-L279

Fullscreen on macOS somewhat doesn't take full control on SDL3 anymore, but cmd-tab has a horrible delay as opposed to borderless.

Susko3:
- Closes #6262 

I'd have hoped this GameHost code would guard against unsupported window modes, but this runs before the window is created so it has no effect.

https://github.com/ppy/osu-framework/blob/79d8e449cfb88a32b125ae3530ae53aee96b3f80/osu.Framework/Platform/GameHost.cs#L1229-L1236

### DIFF
--- a/osu.Framework/Platform/SDL3Window_Windowing.cs
+++ b/osu.Framework/Platform/SDL3Window_Windowing.cs
@@ -90,6 +90,9 @@ namespace osu.Framework.Platform
 
             config.BindWith(FrameworkSetting.WindowMode, WindowMode);
 
+            if (!SupportedWindowModes.Contains(WindowMode.Value))
+                WindowMode.Value = DefaultWindowMode;
+
             WindowMode.BindValueChanged(evt =>
             {
                 switch (evt.NewValue)


bdach (approved):

bdach:
@peppy this pull (or https://github.com/ppy/osu-framework/pull/6264, if it gets merged, I guess) must make it into a framework package before we do a game release.

PR #6264, opened 2024-04-22
Bring back borderless mode for macOS and Linux

Susko3:
Noticed when looking into #6262.

The [SDL3 migration readme](https://github.com/libsdl-org/SDL/blob/5ffb2f47d13f522743e3a064e642366caf4f215e/docs/README-migration.md#sdl_videoh) is cryptic about borderless being available:

> SDL_WINDOW_FULLSCREEN_DESKTOP has been removed, and you can call SDL_GetWindowFullscreenMode() to see whether an exclusive fullscreen mode will be used or the borderless fullscreen desktop mode will be used when the window is fullscreen.

Not only can you check if the used fullscreen mode will be borderless, you can also set it by passing `NULL` to [`SDL_SetWindowFullscreenMode`](https://wiki.libsdl.org/SDL3/SDL_SetWindowFullscreenMode)!


Joehuu:
Borderless itself seems to behave/work as expected (before SDL3) but windowed size get maximised after toggling between the modes.

https://github.com/ppy/osu-framework/assets/35318437/521207d0-d598-427b-96a2-397e6080a150

peppy:
I'm also somewhat hesitant to add back borderless mode on macOS when it goes against how basically any other app works.

bdach:
If borderless isn't coming back then we should remove [this game-side disclaimer](https://github.com/ppy/osu/blob/5e1f6964d19639b3a952ed6aa6aa1ef22dccabe1/osu.Game/Localisation/LayoutSettingsStrings.cs#L27-L30).

osu.Framework/Platform/SDL/SDL3Extensions.cs
```diff
@@ -859,16 +859,17 @@ public static SDL_Scancode ToScancode(this InputKey inputKey)
...
+            // for windows
             if (windowFlags.HasFlagFast(SDL_WindowFlags.SDL_WINDOW_BORDERLESS))
                 return WindowState.FullscreenBorderless;
 
             if (windowFlags.HasFlagFast(SDL_WindowFlags.SDL_WINDOW_MINIMIZED))
                 return WindowState.Minimised;
 
             if (windowFlags.HasFlagFast(SDL_WindowFlags.SDL_WINDOW_FULLSCREEN))
-                return WindowState.Fullscreen;
+                return isFullscreenBorderless ? WindowState.FullscreenBorderless : WindowState.Fullscreen;
```
frenzibyte, line 872: Regarding https://github.com/ppy/osu-framework/pull/6264#issuecomment-2070857205, it appears that once the window is in borderless mode, it does not have the `SDL_WINDOW_FULLSCREEN` flag set:

CleanShot 2024-05-20 at 15 19 01

Therefore, the `WindowState` is kept at normal even after the window has switched to borderless, and `sizeWindowed` becomes corrupted as a result, causing the window to become maximised when returning from borderless.

frenzibyte (changes requested):

osu.Framework/Platform/SDL/SDL3Extensions.cs
Susko3: @frenzibyte could you re-test this?

It's possible that this is because the window has not yet entered the fullscreen borderless mode due to animations still happening. Are the SDL window flags correct after the `SDL_EVENT_WINDOW_ENTER_FULLSCREEN` event arrives?

osu.Framework/Platform/SDL3/SDL3Window_Windowing.cs
```diff
@@ -805,7 +802,14 @@ private void storeWindowSizeToConfig()
         /// <returns>
         /// The size of the borderless window's draw area.
         /// </returns>
-        protected virtual Size SetBorderless(Display display) => throw new PlatformNotSupportedException();
+        protected virtual unsafe Size SetBorderless(Display display)
+        {
+            ensureWindowOnDisplay(display);
+
+            SDL_SetWindowFullscreen(SDLWindowHandle, true);
```
frenzibyte, line 809: This should be accompanied with `SDL_SetWindowFullscreenMode(SDLWindowHandle, null)` to actually use borderless mode.

frenzibyte (changes requested):

osu.Framework/Platform/SDL/SDL3Extensions.cs
frenzibyte: I can reproduce again, and the statement above seems correct. The following path is triggered before the `SDL_EVENT_WINDOW_ENTER_FULLSCREEN` event, causing the size bindable to incorrectly store the borderless size:
https://github.com/ppy/osu-framework/blob/316bbceb7c64727c0fb1cdec5cbfd8d498fd5b52/osu.Framework/Platform/SDL3/SDL3Window.cs#L316-L321

I recall there was something about `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED` being more reliable than `SDL_EVENT_WINDOW_RESIZED` to get the window size while the window is in its final state, I'm not sure how correct that is right now. 

All I can confirm is that `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED` gets called after `SDL_EVENT_WINDOW_ENTER_FULLSCREEN`, and we have only been handling `SDL_EVENT_WINDOW_RESIZED` so the game doesn't look weird when resizing it. Perhaps we can just not store the window size in the "resized" event.

Susko3: > The following path is triggered [...]

When this is triggered, what is the value of `SDL_GetWindowFlags(SDLWindowHandle)`? Framework has cached a stale state, but what does SDL say?

Also please still check this:
> Are the SDL window flags correct after the SDL_EVENT_WINDOW_ENTER_FULLSCREEN event arrives?

frenzibyte: It does have `SDL_WINDOW_FULLSCREEN` when checking the flags at `SDL_EVENT_WINDOW_RESIZED`, while `WindowState` is indeed stale.

The next event triggered after resize is `SDL_EVENT_WINDOW_ENTER_FULLSCREEN`, which is what updates the `WindowState` to `FullscreenBorderless` because of calling `updateAndFetchWindowSpecifics()` I suppose, and yes the SDL window flags remains correct thereafter.

osu.Framework/Platform/SDL3/SDL3Window_Windowing.cs
```diff
@@ -788,9 +786,9 @@ private void storeWindowPositionToConfig()
         /// </remarks>
         private bool updatingWindowStateAndSize;
 
-        private void storeWindowSizeToConfig()
+        private unsafe void storeWindowSizeToConfig()
         {
-            if (WindowState != WindowState.Normal)
+            if (WindowState != WindowState.Normal || SDL_GetWindowFlags(SDLWindowHandle).HasFlagFast(SDL_WindowFlags.SDL_WINDOW_FULLSCREEN))
```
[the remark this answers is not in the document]
Susko3: `WindowState` is probably invalid here as this is called from the event watch. Makes sense to avoid storing the size then.

Give it a test on `TestSceneWindowed`:

```diff
diff --git a/osu.Framework/Platform/SDL3/SDL3Window.cs b/osu.Framework/Platform/SDL3/SDL3Window.cs
index bb03b6093..33e509828 100644
--- a/osu.Framework/Platform/SDL3/SDL3Window.cs
+++ b/osu.Framework/Platform/SDL3/SDL3Window.cs
@@ -316,7 +316,7 @@ protected void HandleEventFromWatch(SDL_Event evt)
                 case SDL_EventType.SDL_EVENT_WINDOW_RESIZED:
                     // polling via SDL_PollEvent blocks on resizes (https://stackoverflow.com/a/50858339)
                     if (!updatingWindowStateAndSize)
-                        fetchWindowSize();
+                        fetchWindowSize(storeToConfig: false);
 
                     break;
             }
diff --git a/osu.Framework/Platform/SDL3/SDL3Window_Windowing.cs b/osu.Framework/Platform/SDL3/SDL3Window_Windowing.cs
index 8007a4a71..33f5a3512 100644
--- a/osu.Framework/Platform/SDL3/SDL3Window_Windowing.cs
+++ b/osu.Framework/Platform/SDL3/SDL3Window_Windowing.cs
@@ -445,7 +445,7 @@ private unsafe Rectangle windowDisplayBounds
         /// Updates <see cref="Size"/> and <see cref="Scale"/> according to SDL state.
         /// </summary>
         /// <returns>Whether the window size has been changed after updating.</returns>
-        private unsafe void fetchWindowSize()
+        private unsafe void fetchWindowSize(bool storeToConfig = true)
         {
             int w, h;
             SDL_GetWindowSize(SDLWindowHandle, &w, &h);
@@ -460,7 +460,8 @@ private unsafe void fetchWindowSize()
             Scale = (float)drawableW / w;
             Size = new Size(w, h);
 
-            storeWindowSizeToConfig();
+            if (storeToConfig)
+                storeWindowSizeToConfig();
         }
 
         #region SDL Event Handling
@@ -788,7 +789,7 @@ private void storeWindowPositionToConfig()
 
         private unsafe void storeWindowSizeToConfig()
         {
-            if (WindowState != WindowState.Normal || SDL_GetWindowFlags(SDLWindowHandle).HasFlagFast(SDL_WindowFlags.SDL_WINDOW_FULLSCREEN))
+            if (WindowState != WindowState.Normal)
                 return;
 
             storingSizeToConfig = true;
```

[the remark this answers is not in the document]
frenzibyte: Looks to work as expected.

frenzibyte (approved):
Leaving a tentative approval to say this is pretty much good from macOS's side. Leaving it to others as to whether this can get in or not.

peppy:
Gonna merge this to keep things moving.

peppy (approved):

PR #6266, opened 2024-04-23
Refactor child masking bounds handling

EVAST9919:
Split from https://github.com/ppy/osu-framework/pull/6261

Original pr has too much stuff changed at once, so a bit more lenient approach feels way better.
Performance shouldn't differ much (if at all).


osu.Framework/Graphics/Drawable_ProxyDrawable.cs
```diff
@@ -60,13 +60,7 @@ internal override DrawNode GenerateDrawNodeSubtree(ulong frame, int treeIndex, b
...
             public override bool IsPresent => false;
 
-            public override bool UpdateSubTreeMasking(Drawable source, RectangleF maskingBounds)
-            {
-                if (Original.IsDisposed)
-                    return false;
-
-                return Original.UpdateSubTreeMasking(this, maskingBounds);
-            }
+            protected override bool ComputeIsMaskedAway(RectangleF maskingBounds) => false;
```
EVAST9919, line 63: Should I be using `UpdateSubTreeMasking() => false` here may be? It's a bit hard to wrap my head around proxies, to be honest.

EVAST9919: Yes, I should.

osu.Framework/Graphics/Containers/CompositeDrawable.cs
```diff
@@ -1478,6 +1481,7 @@ protected set
                     return;
 
                 masking = value;
+                childMaskingBoundsBacking.Invalidate();
```
smoogipoo, line 1484: Since layout invalidation doesn't propagate, could there be a situation such as:
- Everything is validated.
- Parent has `Masking` set to `true`.
- The parent's masking bounds are invalidated, but the child isn't.
- The child still uses the old masking bounds.

It looks like your test scene asserts against `parent.ChildMaskingBounds`, but does not check an actual child's masking bounds.

smoogipoo: If this is the case, then we may want a masking invalidation bit instead.

osu.Framework.Tests/Layout/TestSceneContainerLayout.cs
```diff
@@ -498,6 +499,29 @@ public void TestNonAutoSizingParentDoesNotInvalidateSizeDependenciesFromChild()
...
+            {
+                Child = container = new Container
+                {
+                    Size = new Vector2(100)
+                };
+            });
+
+            AddAssert("Masking is off", () => container.Masking == false);
+            AddStep("Save child bounds", () => childMaskingBounds = container.ChildMaskingBounds);
+            AddStep("Disable masking", () => container.Masking = true);
```
smoogipoo, line 521:
```suggestion
            AddStep("Enable masking", () => container.Masking = true);
```

smoogipoo (changes requested):

EVAST9919:
> I believe that this assertion should pass.

The test you've provided failed because as you said children weren't aware of the masking changes in the sub-tree. Invalidating `DrawInfo` did the trick, and it shouldn't affect performance since there are little to no use cases of frequent masking toggling.

smoogipoo (approved):

smoogipoo:
Looks okay now, can't think of anything else that'll break. The `DrawNode` invalidation is probably not required but I can understand why you've added it. Not a big deal in any case.

PR #6268, opened 2024-04-29
Refactor `DisplayHandle` into seperate android and linux properties

Susko3:
- cleanup from https://github.com/ppy/osu-framework/pull/6105

The `DisplayHandle` is specific to linux window magnagers and is very different from the android surface handle.
I'm not sure if any of this code actually works / is useful, as android doesn't use veldrid yet.


bdach:
> The `DisplayHandle` is specific to linux window magnagers and is very different from the android surface handle.

In that case it should be annotated with `[SupportedOSPlatform("linux")]`.

osu.Framework/Platform/SDL3Window.cs
```diff
@@ -138,6 +139,9 @@ public virtual IntPtr DisplayHandle
             }
         }
 
+        [SupportedOSPlatform("android")]
+        public virtual IntPtr SurfaceHandle => throw new PlatformNotSupportedException();
```
bdach, line 143: Oh I hate that this exists but I see why. It's because the layering on this is kinda borked.

tl;dr: the graphics surface implementations live in base framework rather than the android project, so they can't directly reference something like `AndroidGameActivity` to bypass this song-and-dance entirely and just call what they want to call directly where they need it.

I guess with the `SupportedOSPlatform` attribute on this I could be okay with this but it feels a bit smelly nevertheless. There's probably a refactor in here that gets rid of it (likely by shipping the surface implementations off to the mobile projects - or at least the android one) but I'm not sure if worth.

@smoogipoo interested in your thoughts on layering here since this touches the veldrid parts

smoogipoo: Not worth overthinking this.

smoogipoo (approved):

smoogipoo (approved):

PR #6269, opened 2024-04-29
Allow customising the `Main()` function on Android

Susko3:
Since iOS and desktop have `Main()` functions by default, why not allow it on android. SDL expects us to provide a main function anyways.

If we ever add `HostOptions` that make sense on android, those can be set in the custom `Main()`.

Consumers should override either `Main()` or `CreateGame()`. This is a non-breaking change.

### DIFF
--- a/osu.Framework.Android/AndroidGameActivity.cs
+++ b/osu.Framework.Android/AndroidGameActivity.cs
@@ -33,7 +33,7 @@ namespace osu.Framework.Android
 
         internal static AndroidGameSurface Surface => (AndroidGameSurface)MSurface!;
 
-        protected abstract Game CreateGame();
+        protected virtual Game CreateGame() => throw new NotImplementedException();
 
         protected override string[] GetLibraries() => new string[] { "SDL3" };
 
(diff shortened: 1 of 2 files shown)


osu.Framework.Android/AndroidGameActivity.cs
```diff
@@ -34,34 +33,35 @@ public abstract class AndroidGameActivity : SDLActivity
...
         internal static AndroidGameSurface Surface => (AndroidGameSurface)MSurface!;
 
-        private GameHost? host;
-
-        protected abstract Game CreateGame();
+        /// <summary>
+        /// Create the game to be run.
+        /// </summary>
+        /// <remarks>Either override this or <see cref="Main"/>.</remarks>
+        protected virtual Game CreateGame() => throw new NotImplementedException();
```
bdach, line 40: I don't get it. Why would anyone want this? Where are "desktop and iOS" doing this sort of thing?

In general having this sort of thing throw at runtime just turns me off from this patch. It looks sloppy, it's bad OOP design. Having two overlapping overrides is something I actively hunt down in review and try to shut down as soon as possible. There should ideally be one way to do a thing.

osu.Framework.Android/AndroidGameHost.cs
```diff
@@ -25,12 +26,29 @@ public class AndroidGameHost : SDL3GameHost
...
+            : base(gameName, options)
         {
             this.activity = activity;
         }
 
+        protected override void SetupForRun()
+        {
+            base.SetupForRun();
+
+            activity.TrimMemory += Collect;
```
bdach, line 39: I like this change (calling into the existing `Collect()`, which also has an extra imagesharp allocator resource release call that isn't being called in curent master), but I don't get what it has to do with any of the rest of this diff. It's probably the only change from here I'd entertain at this point.

Susko3:
This PR was mostly meant to clean up `AndroidGameActivity` & how it interacts with the `AndroidGameHost`. Upon revisiting, I've realised most of the things I've moved around (`AllowScreenSuspension`, `TrimMemory`) are already implemented in SDL, so I've changed those to use the SDL implementation in #6271 and #6272.

Closing in favour of those. If we ever need `HostOptions` on android, we'll figure something out.

PR #6271, opened 2024-04-30
Use SDL3 screen saver functions to implement `AllowScreenSuspension`

Susko3:
This removes the custom Android and iOS implementations, and also adds support for desktop platforms.

[Android](https://github.com/Susko3/SDL/blob/dc61b4b2f81c72028d7114e8b914b044d46f94f0/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java#L903-L907) and [iOS](https://github.com/Susko3/SDL/blob/dc61b4b2f81c72028d7114e8b914b044d46f94f0/src/video/uikit/SDL_uikitvideo.m#L173-L176) implementation in SDL3 match our old custom code.


smoogipoo (approved):
I think trial by fire is best :)

PR #6272, opened 2024-04-30
Fix `GameHost.Collect()` being called twice on android

Susko3:
Depends on (to have identical behaviour):
- [x] https://github.com/libsdl-org/SDL/pull/9654
- [x] New SDL3-CS build.

SDL sends us a `SDL_EVENT_LOW_MEMORY` when Android's `onLowMemory()` handler fires. o!f handles this event and calls `GameHost.Collect()`. We also have our own handling in `onTrimMemory()`. Depending on how android handles those two functions, we may collect memory twice.

I think it makes more sense to just have `SDL_EVENT_LOW_MEMORY` handling, and not have custom logic just for android.

This PR was originally meant to pick out the useful parts from https://github.com/ppy/osu-framework/pull/6269/files#r1584189904, but I've since realised SDL already sends us a low memory event that we handle.

### DIFF
--- a/osu.Framework.Android/AndroidGameActivity.cs
+++ b/osu.Framework.Android/AndroidGameActivity.cs
@@ -11,7 +11,6 @@ using Java.Lang;
 using ManagedBass;
 using Org.Libsdl.App;
 using osu.Framework.Extensions.ObjectExtensions;
-using osu.Framework.Platform;
 using Debug = System.Diagnostics.Debug;
 
 namespace osu.Framework.Android
@@ -34,8 +33,6 @@ namespace osu.Framework.Android
 
         internal static AndroidGameSurface Surface => (AndroidGameSurface)MSurface!;
 
-        private GameHost? host;
-
         protected abstract Game CreateGame();
 
         protected override string[] GetLibraries() => new string[] { "SDL3" };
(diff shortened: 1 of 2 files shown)


hwsmm:
I actually noticed it before, so I removed it in [my cleanup commit](https://github.com/ppy/osu-framework/commit/b3314484f970d73c39bf5d94452afa7f6c761b1e#diff-4fee16ad66bef388a93c9fcdb9e7807d4792b23d33f8c5ed91612dc9022e52c9L50), but I guess some changes after mine added it back?

Though I think I should have added a comment in the commit to make sure it was intended.

hwsmm (approved):

peppy:
i guess we have a new sdl3-cs build now?

bdach:
Probably be best to get another sdl3-cs package out to get the fix for https://github.com/ppy/osu-framework/issues/6299.

Susko3:
I've bumped `ppy.SDL3-CS.Android` version to match `ppy.SDL3-CS` (they should always match, but it was missed). This includes the SDL3 change required for this PR.

Perhaps the `ppy.SDL3-CS.Android` package should depend on `ppy.SDL3-CS` of the same version. Bumping only one would cause package upgrade/downgrade build failures.

peppy (approved):

smoogipoo (approved):

PR #6276, opened 2024-05-02
Don't redraw `Path` on colour invalidation

EVAST9919:
Partially addresses https://github.com/ppy/osu/issues/21495

Helps in reducing lag on complex (aspire/generated) sliders during fadein/out by not redrawing the slider on colour/alpha change. Note that this won't improve things on position and other geometry changes.
This probably can be applied to all the buffered containers, but I could't think of a good way to generalize that without just copying the same code into `BufferedContainer`.

Map: https://osu.ppy.sh/beatmapsets/1746664#osu/3572698
master:
https://streamable.com/hhuou2
pr:
https://streamable.com/iv4k0o

### DIFF
--- a/osu.Framework/Graphics/Lines/Path.cs
+++ b/osu.Framework/Graphics/Lines/Path.cs
@@ -13,6 +13,7 @@ using System.Collections.Generic;
 using osu.Framework.Caching;
 using osu.Framework.Extensions.EnumExtensions;
 using osu.Framework.Graphics.Rendering;
+using osu.Framework.Layout;
 using osuTK.Graphics;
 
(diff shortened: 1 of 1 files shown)


bdach:
So _as far as I can tell_ this appears valid. I was sorta worried this could maybe break blending because of

https://github.com/ppy/osu-framework/blob/d73af6ccc9a197e63139711d2cb1df964f82fd0d/osu.Framework/Graphics/Lines/Path.cs#L290-L291

but it appears to be fine. That said this is not super obvious to me so I'd probably appreciate a second pair of eyes.

bdach (approved):

smoogipoo (approved):

PR #6280, opened 2024-05-11
Add parent container per row in `GridContainer`

EVAST9919:
Improves performance in cases when `GridContainer` is too big with Masking and DrawNode sub-trees having to traverse less elements when most of the rows are masked away.

This change is non-breaking, all the `GridContainer` usages can be left untouched (except https://github.com/ppy/osu/blob/6cfe1356cd9ac77f70d762fab573e58bfd61b2c2/osu.Game/Screens/Edit/EditorScreenWithTimeline.cs#L96, the only case I could found with broken depth, which is a downside of this pr).

# Example
Imagine a `GridContainer` with a size of 5 * 1000 cells with 20 rows visible. In master we are traversing each cell no matter what (5000 cells), now we are traversing all the rows + cells of non-masked rows (1000 + 20 * 5 = 1100 cells). And the longer the `GridContainer` the bigger the relative benefit.
Of course in case of a small grid there are more total containers, however I didn't notice any considerable performance regressions anywhere.

# Downside
Less control over Depth of the children.

# Test
Map with a big table for testing: https://osu.ppy.sh/beatmapsets/1238185#mania/2574372
|master|pr|
|---|---|
|master|pr|

### DIFF
--- a/osu.Framework/Graphics/Containers/GridContainer.cs
+++ b/osu.Framework/Graphics/Containers/GridContainer.cs
@@ -206,7 +206,6 @@ namespace osu.Framework.Graphics.Containers
                     rows[r].Add(cells[r, c]);
                 }
 
-                rows[r].Depth = r;
                 AddInternal(rows[r]);
             }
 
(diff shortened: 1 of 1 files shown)


smoogipoo:
I'm not sure about the depth issue. I'll have to look into the history of why it exists, but it may be a non-negotiable.

I believe the editor usage can probably be replaced by a proxied drawable? Do you have another alternative there?

EVAST9919:
> I believe the editor usage can probably be replaced by a proxied drawable? Do you have another alternative there?

Replacing depth with masking works since there are no elements in the timeline which will be drawn on top of the playfield.
Снимок экрана 2024-05-12 091815

smoogipoo:
That wasn't the intention of that code:

https://github.com/ppy/osu/commit/26110cd777e3c4880adc2a10c2bb49e1b64c7947

EVAST9919:
I see. Well, it looks like there are more caveats to this, so I'll think about other potential solutions.

PR #6281, opened 2024-05-12
Add TrackLoudness (BASSloud)

hwsmm:
- Requires https://github.com/peppy/ManagedBass/pull/1, https://github.com/ppy/osu-framework/pull/6233

This PR makes `TrackLoudness` available in framework for https://github.com/ppy/osu/pull/27793.

Diff from https://github.com/peppy/ManagedBass/pull/1#discussion_r1597561035 is also needed, but I can revert this if anyone doesn't want.


osu.Framework/Audio/Track/TrackLoudness.cs
```diff
@@ -0,0 +1,141 @@
...
+                    {
+                        Logger.Log($"Failed to start BassLoud: {Bass.LastError}");
+                        return;
+                    }
+
+                    while (Bass.ChannelGetData(decodeStream, buffer, buffer.Length) >= 0)
+                    {
+                    }
+
+                    float integratedLoudness = 1;
```
smallketchup82, line 80: After double checking my docs. The returned integratedLoudness will be -infinity if the track is complete silence. Meaning that it won't ever really exceed -1 under normal circumstances. I believe that if we're going to be associating success with less than 0, we might as well revert my changes earlier making the default value of integratedLoudness = 1. Since the reason I did that was that the if statements were checking greater than 0 and associating it with failure. whereas these if statements are checking less than.
```suggestion
                    float integratedLoudness = 0;
```

smallketchup82 (commented):
Despite the fact that I'm yet to test this, the diff alone looks fine.

osu.Framework/Audio/Track/TrackLoudness.cs
```diff
@@ -0,0 +1,141 @@
...
+                    {
+                        Logger.Log($"Failed to start BassLoud: {Bass.LastError}");
+                        return;
+                    }
+
+                    while (Bass.ChannelGetData(decodeStream, buffer, buffer.Length) >= 0)
+                    {
+                    }
+
+                    float integratedLoudness = 0;
```
smallketchup82, line 80: `Local variable 'integratedLoudness' hides field 'float? osu.Framework.Audio.Track.TrackLoudness.integratedLoudness'`

There is a private instance variable called `integratedLoudness`, and a local variable here called `integratedLoudness`. We should pick a different name for either `integratedLoudness`.

I would suggest renaming this to `bassIntegratedLoudness`

smallketchup82 (changes requested):

osu.Framework/Audio/Track/TrackLoudness.cs
```diff
@@ -0,0 +1,141 @@
...
+                byte[] buffer = ArrayPool<byte>.Shared.Rent(20000);
+
+                try
+                {
+                    int loudHandle = BassLoud.Start(decodeStream, BassFlags.BassLoudnessIntegrated | BassFlags.BassLoudnessAutofree, 0);
+
+                    if (loudHandle == 0)
+                    {
+                        Logger.Log($"Failed to start BassLoud: {Bass.LastError}");
+                        return;
```
smallketchup82, line 73: So the main reason we used `Logger.Log` and `return;` instead of throwing an error was because the old file didn't use a try catch, so throwing an error would result in the stream never getting freed. In this code, the finally would free the stream regardless of whether or whether not there is an error, so we can go back to throwing.

Should probably also add a catch to the try statement to grab the error messages, then log them to the errorfile

osu.Framework/Audio/Track/TrackLoudness.cs
```diff
@@ -0,0 +1,141 @@
...
+                    return;
+                }
+
+                FileCallbacks fileCallbacks = new FileCallbacks(new DataStreamFileProcedures(data));
+
+                int decodeStream = Bass.CreateStream(StreamSystem.NoBuffer, BassFlags.Decode | BassFlags.Float, fileCallbacks.Callbacks, fileCallbacks.Handle);
+
+                if (decodeStream == 0)
+                {
+                    Logger.Log($"Bass failed to create a stream while trying to measure loudness: {Bass.LastError}");
```
smallketchup82, line 59:
```suggestion
                    Logger.Log($"Bass failed to create a stream while trying to measure loudness: {Bass.LastError}", LoggingTarget.Runtime, LogLevel.Error);
```

osu.Framework/Audio/Track/TrackLoudness.cs
```diff
@@ -0,0 +1,141 @@
...
+            this.data = data;
+
+            readTask = Task.Run(() =>
+            {
+                if (data == null)
+                    return;
+
+                if (Bass.CurrentDevice < 0)
+                {
+                    Logger.Log("Failed to measure loudness as no bass device is available.");
```
smallketchup82, line 49: I can understand logging instead of throwing here, but at the same time we should probably be logging to the error logfile
```suggestion
                    Logger.Log("Failed to measure loudness as no bass device is available.", LoggingTarget.Runtime, LogLevel.Error);
```

smallketchup82 (changes requested):
Apart from minor work on error handling and the previous review, this should be good to go

smallketchup82 (approved):
lgtm

PR #6282, opened 2024-05-12
Improve input handling performance of `TableContainer`

EVAST9919:
Closes https://github.com/ppy/osu/issues/26886 (Or at very least should improve things considerably).

By default `CompositeDrawable` (in our case `GridContainer`) is trying to build positional input sub-tree with all it's internal children and then throwing away unused ones. By assuming all the content within cells isn't overlapping other cells and contained within it's boundaries (which is a case for the `TableContainer`) we can propagate positional input only to a set cell and ignore the others.

With that https://github.com/ppy/osu/pull/27958 could've been undone (performance on mouse move is way better now), however idle performance is still not that great since non-positional input sub-tree is [being created on each frame](https://github.com/ppy/osu-framework/blob/9c6dee5e854d396f9c223ab9f722995d3fa7a193/osu.Framework/Input/InputManager.cs#L497) if nothing is being focused.

This version of `GridContainer` can probably be made public with a better name.

Master:
https://streamable.com/t9tkah
Pr:
https://streamable.com/yy6jrq

### DIFF
--- a/osu.Framework/Graphics/Containers/CompositeDrawable.cs
+++ b/osu.Framework/Graphics/Containers/CompositeDrawable.cs
@@ -1453,13 +1453,18 @@ namespace osu.Framework.Graphics.Containers
             if (!ReceivePositionalInputAtSubTree(screenSpacePos))
                 return false;
 
+            PropagatePositionalInputQueue(screenSpacePos, queue);
+
+            return true;
+        }
+
+        internal virtual void PropagatePositionalInputQueue(Vector2 screenSpacePos, List<Drawable> queue)
+        {
             for (int i = 0; i < aliveInternalChildren.Count; ++i)
             {
                 if (ShouldBeConsideredForInput(aliveInternalChildren[i]))
                     aliveInternalChildren[i].BuildPositionalInputQueue(screenSpacePos, queue);
             }
-
-            return true;
         }
 
         #endregion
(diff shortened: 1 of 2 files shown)


smoogipoo:
We don't usually make the assumption that content isn't going to overlap, unless the content is masking. This is the case with `CompositeDrawable` too - it accepts full-screen input unless it's masking to potentially pass input to a child sizing outside of its bounds.

I'm concerned about changing that here - yes we're likely not relying on this anywhere right now, but it's one of those things that has very slightly different behaviour that's inevitably going to bite someone that maybe _does_ want that functionality.

It feels wrong to put this hack in o!f.

---

If you're interested in exploring other options, one thing I've wanted to do is experiment with using a quadtree, BVH, or similar data structure for input. Basically some way of subdividing the subtree by AABBs to reduce the number of intersections required to compute to something more like O(nlogn).

Here's a few good resources on BVH: 
- https://box2d.org/files/ErinCatto_DynamicBVH_GDC2019.pdf
- https://www.pbr-book.org/3ed-2018/Primitives_and_Intersection_Acceleration/Bounding_Volume_Hierarchies
- https://meistdan.github.io/publications/bvh_star/paper.pdf

I also started on a QT implementation for paths here (arguably more important than this PR): https://github.com/ppy/osu-framework/compare/master...smoogipoo:osu-framework:path-quadtree but it doesn't handle every case correctly, and it doesn't use AABBs.

smoogipoo:
I could be fine with this PR if it was locked behind a `bool` flag for the time being.

EVAST9919:
Closing since this local solution won't fix other similar issues which are just not reported yet and as mentioned more general approach should be implemented.

PR #6286, opened 2024-05-17
Implement focus traps

smoogipoo:
Fixes https://github.com/ppy/osu/issues/25769
Fixes https://github.com/ppy/osu/issues/26079 (likely)
Supersedes / closes https://github.com/ppy/osu-framework/pull/6095
Closes https://github.com/ppy/osu-framework/pull/6097 (note that https://github.com/ppy/osu/issues/25850 is likely not fixed by this and will need reimplementation)

By extracting focus-management logic out of `InputManager`, we can implement something akin to a focus trap. Components can trap the focus management (`ChangeFocus()`/`TriggerFocusContention()`) of their subcomponents to add supplementary logic.

This PR features a partial redesign of `Dropdown` making use of this. There are two components wanting to manage focus here - the `Menu` and the `TextBox`.
- The `Menu`'s focus is trapped by `Dropdown` and reduced to a no-op.
- The `TextBox`'s focus is trapped within `DropdownSearchBox` to perform additional functionality such as opening the dropdown on focus, selecting items on commit, clearing text on escape, etc.

This is not a breaking change, but I've obsoleted `InputManager.TriggerFocusContention()` and `InputManager.ChangeFocus()` partially to make this easy to test.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/DropdownSearchBar.cs
+++ b/osu.Framework/Graphics/UserInterface/DropdownSearchBar.cs
@@ -157,8 +157,6 @@ namespace osu.Framework.Graphics.UserInterface
         /// </summary>
         protected abstract TextBox CreateTextBox();
 
-        Drawable IFocusManager.FocusedDrawable => GetContainingFocusManager().FocusedDrawable;
-
         void IFocusManager.TriggerFocusContention(Drawable? triggerSource)
         {
             // Clear search text first without releasing focus.
--- a/osu.Framework/Graphics/UserInterface/DropdownSearchBar.cs
+++ b/osu.Framework/Graphics/UserInterface/DropdownSearchBar.cs
@@ -176,8 +176,24 @@ namespace osu.Framework.Graphics.UserInterface
 
         // Focus management is isolated to the IDropdown interface by Dropdown.
         Drawable IFocusManager.FocusedDrawable => GetContainingFocusManager().FocusedDrawable;
-        void IFocusManager.TriggerFocusContention(Drawable? triggerSource) => dropdown.TriggerFocusContention(triggerSource);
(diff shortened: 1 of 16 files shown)


osu.Framework/Graphics/UserInterface/DropdownSearchBar.cs
```diff
@@ -36,89 +30,159 @@ public bool AlwaysDisplayOnFocus
...
-            // as even despite that the text box will receive focus via the text box input manager;
-            // it is necessary to cut off the text box input manager from parent input entirely.
-            // TODO: preferably figure out a better way to do this.
-            bool willShowOverlappingKeyboard = host?.OnScreenKeyboardOverlapsGameWindow == true;
-
-            if (willShowOverlappingKeyboard && !AlwaysDisplayOnFocus)
-            {
-                textBoxInputManager.UseParentInput = false;
-                return;
-            }
```
smoogipoo, line 86: I haven't yet tested whether removing this is an issue. The logic here is quite convoluted.

smoogipoo: Fixed by https://github.com/ppy/osu-framework/pull/6286/commits/5724733a80f3226a9a71b7847410d4e8c93df18e. It's touching some scary code but I think it makes sense.

frenzibyte:
I've added tests in my other PR to ensure the issues linked above are specifically fixed (see https://github.com/ppy/osu-framework/pull/6097/commits/cf31661dda40e9f20e077dab5241c767d73de2b8), worth adding in this PR. 

For https://github.com/ppy/osu/issues/25850, I would just disable searching in `TabControl`s and call it a day. I can open a separate PR for that.

smoogipoo:
@frenzibyte Thanks. Added your tests almost verbatim in the last commit.

peppy (approved):

peppy:
Definitely better than the other approach, let's see how this pans out 😄 .

PR #6288, opened 2024-05-19
Build ffmpeg for linux-arm, linux-arm64, linux-x86

kvnp:
This PR adds the targets linux-arm, linux-arm64 as part of https://github.com/ppy/osu-deploy/pull/170. Will be updated once https://github.com/ppy/osu-framework/pull/6255 is merged.


FreezyLemon (changes requested):
Some notes:

- Please remove changes not related to the Linux build (macOS bump, merging the Windows build process). These are out of scope and maybe change the resulting binaries which would need testing
- The x86 build [doesn't work](https://github.com/FreezyLemon/osu-framework/actions/runs/9191208432). x86 support is not necessary so you don't need to include it at all, but if you do want to include it, please make sure it works.

I wonder if there's a way to achieve the same builds using cross-compilation similar to the Windows build scripts. That would make it a lot easier to compile these locally (installing QEMU and docker is fairly hefty for these simple builds). That said, it's not a high priority or blocking issue.

smoogipoo:
This one can be updated now that https://github.com/ppy/osu-framework/pull/6255 has been merged.

kvnp:
> Some notes:
> 
> * Please remove changes not related to the Linux build (macOS bump, merging the Windows build process). These are out of scope and maybe change the resulting binaries which would need testing
> * The x86 build [doesn't work](https://github.com/FreezyLemon/osu-framework/actions/runs/9191208432). x86 support is not necessary so you don't need to include it at all, but if you do want to include it, please make sure it works.

linux-x86 now builds after disabling asm. I'm not sure about the performance implications, but I'm glad to be one step ahead for now. According to this [issue](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=109137), GCC is having trouble building FFMPEG for x86 32-bit on a 64-bit system. I would open a new PR when this is fixed.

> I wonder if there's a way to achieve the same builds using cross-compilation similar to the Windows build scripts. That would make it a lot easier to compile these locally (installing QEMU and docker is fairly hefty for these simple builds). That said, it's not a high priority or blocking issue.

I find it much easier to set up Docker than to try to piece together a cross-compilation toolchain. If multiplatform images are needed, which is the case here, Docker can install the QEMU binaries [by itself with a single command](https://docs.docker.com/build/building/multi-platform/#qemu-without-docker-desktop). I would leave it as is for now, but would like to revisit this in the future when I have more experience with cross compiling on Linux.

.github/workflows/build-ffmpeg.yml
```diff
@@ -193,6 +193,44 @@ jobs:
...
+        uses: addnab/docker-run-action@v3
+        with:
+          image: ${{ matrix.container }}
+          options: -v ${{ github.workspace }}:/workspace -w /workspace -e DEBIAN_FRONTEND=noninteractive
+          run: |
+            apt-get update -y -qq
+            apt-get install -y build-essential curl
+            if [ "$(dpkg --print-architecture)" = "i386" ]; then
+              apt-get install -y nasm
+            fi
```
FreezyLemon, line 224: `--disable-asm` on x86 Linux should mean that this is not needed

osu.Framework.NativeLibs/scripts/ffmpeg/build-linux.sh
```diff
@@ -6,18 +6,31 @@ SCRIPT_PATH=$(pwd)
...
+elif [ "$(dpkg --print-architecture)" = "i386" ]; then
+    arch="x86"
+elif [ "$(dpkg --print-architecture)" = "arm64" ]; then
+    arch="arm64"
+elif [ "$(dpkg --print-architecture)" = "armhf" ]; then
+    arch="arm"
+else
+    echo "Unsupported architecture: $(dpkg --print-architecture)"
+    exit 1
+fi
```
FreezyLemon, line 20: Please don't use `dpkg` for this. `uname` is available almost everywhere while `dpkg` is Debian-specific.

osu.Framework.NativeLibs/scripts/ffmpeg/common.sh
```diff
@@ -56,6 +56,12 @@ function prep_ffmpeg() {
 
 function build_ffmpeg() {
     echo "-> Configuring..."
+    if [ "$arch" = "x86" ] && [ "$(uname -s)" = "Linux" ]; then
+        FFMPEG_FLAGS+=(
+            --disable-asm
+        )
+    fi
```
FreezyLemon, line 63: Could be moved to `build-linux.sh` (the other platforms also do arch-specific changes inside their own build scripts, see for example `build-win.sh` lines 23 - 38)

FreezyLemon (changes requested):
Thanks for addressing my review. I left some feedback, sadly the `dpkg` change breaks the scripts for me (I'm not on a Debian-based distro).

> linux-x86 now builds after disabling asm. I'm not sure about the performance implications, ...

It's cool that you got it to work, but without someone confirming that the binaries are usable in terms of performance, I'm not entirely sure it's a great idea to include them..

> I find it much easier to set up Docker than to try to piece together a cross-compilation toolchain.

FFmpeg really doesn't require much of a toolchain, basically gcc/clang plus the GNU assembler. It shouldn't be more complicated than the Windows script that already exists. To be fair, Docker is still simpler, but it's just not my preference.

sr229:
Looks like [ARM64 runners will be available in the end of the year](https://github.blog/news-insights/product-news/arm64-on-github-actions-powering-faster-more-efficient-build-systems/) and is available now for GitHub Enterprise Cloud/Team orgs. I suggest holding off this one until this becomes generally available.

Issue #6067, opened 2023-11-30
Circle doesnt obey Masking of CircularContainer


64ArthurAraujo:
Differently from other shapes circles doesn't follow the masking of the circular container and appears to be in a square container, basically what i was trying to do is something similar to the lazer's logo triangles but with circles.

```C#
new CircularContainer
{
	Anchor = Anchor.Centre,
        Origin = Anchor.Centre,
        Masking = true,
        Y = 0,
	X = 0,
        Size = Size,
        Children = new Drawable[]
        {
        	new Circle {
                        Anchor = Anchor.TopCentre,
                        Origin = Anchor.TopCentre,
                        Colour = Color4.DarkRed,
                        Y = -12.5f,
                	Size = new Vector2(64, 64)
		}
        }
}
```

![Captura de tela de 2023-11-30 10-53-14](https://github.com/ppy/osu-framework/assets/90941580/16c494f9-182c-4f19-a5dd-62bf98210128)   ![Captura de tela de 2023-11-30 10-53-35](https://github.com/ppy/osu-framework/assets/90941580/dd959683-670b-4673-a2ef-c30328d133ed) ![Captura de tela de 2023-11-30 10-53-50](https://github.com/ppy/osu-framework/assets/90941580/43108d73-5faa-4471-9e29-a67ccaa1dc22)

bdach:
Mostly duplicate of #2107. The recommended workaround is to use a `BufferedContainer`.

PR #6290, opened 2024-05-19
Implement `FastCircle` component

EVAST9919:
Adresses https://github.com/ppy/osu-framework/issues/6067, https://github.com/ppy/osu/issues/21495

Pros:
* No nested `Box` drawable (adds up in demanding scenarios)
* [Correct masking when used inside `CircularContainer`](https://github.com/ppy/osu-framework/assets/22874522/2a484318-582e-46f7-b60b-543fb4cbea45)
* No overhead from masking info and no batch breaking in cases when multiple circles are being drawn consecutively (though I think SSBO supposed to fix that?)

Cons (which are holding me back from replacing current `Circle` implementation with this one):
* `Circle` is no longer a container (so is the `Box` which I think is fine)
* No ability to set masking-related effects directly (again `Box` has the same properties)
* Points above might be considered a breaking change, but easily fixeable by using `CircularContainer` instead
* Texture coordinates abuse to pass draw size into the shader without uniform buffers
* Has a separate shader

Performance comparison in demanding scenarios:
|master|pr|
|---|---|
|Снимок экрана (524)|Снимок экрана (525)|
|master-mania|pr-mania|
|master-mania-2|pr-mania-2|
|master-slider|pr-slider|


bdach:
Just reading the title of this makes me want to go "no why is this being proposed even". I'd hope we can not resort to cpp style microoptimisations like this. But maybe it's just me 🤷 

I'd be much more okay with this replacing `Circle` as long as there is an actual perf gain and the presumed increase in draw calls from shader rebinding doesn't kill the perf gain. The tradeoffs sound fine to me.

EVAST9919:
> I'd be much more okay with this replacing `Circle` as long as there is an actual perf gain and the presumed increase in draw calls from shader rebinding doesn't kill the perf gain. The other tradeoffs mentioned sound fine to me.

My current proposals:
* We probably can do some hackery like providing some specific flag somewhere and use the same texture shader with reroute to the new one and thus shader switch will be eliminated, but I can't think of any sane and good-looking solutions
* We can leave this as a side component (probably with a better name and xmldoc) amd use it only in specific cases, which is what I'm doing in provided screenshots
* Wait for SSBO implementation and see performance difference then

osu.Framework/Resources/Shaders/sh_FastCircle.fs
```diff
@@ -0,0 +1,26 @@
...
+void main(void)
+{
+    highp vec2 pixelPos = v_TexRect.zw * 0.5 - abs(v_TexCoord - v_TexRect.zw * 0.5);
+    highp float radius = min(v_TexRect.z, v_TexRect.w) * 0.5;
+
+    highp float dst = max(pixelPos.x, pixelPos.y) > radius ? radius - min(pixelPos.x, pixelPos.y) : distance(pixelPos, vec2(radius));
+
+    highp float alpha = v_BlendRange.x == 0.0 ? float(dst < radius) : (clamp(radius - dst, 0.0, v_BlendRange.x) / v_BlendRange.x);
+
+    o_Colour = getRoundedColor(vec4(vec3(1.0), alpha), vec2(0.0));
```
smoogipoo, line 23: I'm not sure about the use of `v_TexRect` here.
- Have you tested textures at all? It looks like you're not actually sampling the texture.
- What about cropped textures? (`Texture.Crop()`)
- Does this work for rotated ellipses (textured and non-textured)? I assume so but it should be checked.

smoogipoo (changes requested):

osu.Framework/Resources/Shaders/sh_FastCircle.fs
EVAST9919: That's the thing: I'm abusing texture coordinates to pass screenspace size to the shader without uniforms (to compute antialiasing). This new circle isn't a sprite, so textures can't be set.
Regarding rotation: I'll add test showing it's working.

EVAST9919: To clarify: I'm not trying replicate `CircularContainer`, I'm simplifying one particular usage of it, which is `CircularContainer` with a plane `Box` with no texture inside (`Circle` class) to avoid masking overhead in cases where multiple circes are being drawn consecutively (GPU side) and reducing total drawable count (CPU side)

EVAST9919: > That's the thing: I'm abusing texture coordinates to pass screenspace size to the shader without uniforms (to compute antialiasing).

As a side note I can apply hacks explored in this pr to the `Triangles` shader since it's being widely used in the game to eliminate uniform buffer. Will pr after checking whether performance gain is worth doing so.

smoogipoo:
My current position on this is I'm not sure where it fits in right now but I'm not _totally_ opposed to it.

EVAST9919:
> I'd be much more okay with this replacing `Circle` as long as there is an actual perf gain and the presumed increase in draw calls from shader rebinding doesn't kill the perf gain. The other tradeoffs mentioned sound fine to me.
If this "fast" circle can't be applied universally then I'm not super interested in this imo.

So I've added the "worst case" performance test scene, in which drawables are alternating between circle and the box since I haven't noticed much difference in game when replacing circle with my implementation.
In this "worst case" scenario draw call count is the same and for my machine specifically my version runs faster. If you are fine with that then I can update this pr to replace original circle instead.
|`Circle`|`FastCircle`|
|---|---|
|Circle|Fast|

smoogipoo:
The part that stops me from replacing `Circle` is the lack of a texture. It's the reason I questioned it in my review.

Slight tidbit: draw calls aren't equal. It's very likely your system will handle many draw calls much better than, say, a mobile GPU.

peppy:
a separate shader is going to interrupt batching, so we need to be sure that's what we want (probably not?)

smoogipoo:
What you have to understand is that it already interrupts batch as a result of masking. It all depends on whether the gains of SSBO can be realised.

EVAST9919:
> The part that stops me from replacing `Circle` is the lack of a texture. It's the reason I questioned it in my review.

Which current `Circle` doesn't have either, so we are not losing anything.

> a separate shader is going to interrupt batching

There's a way to use the same shader, but I'm not sure if it's too ugly or some better approach can be implemented.

> I have a [working implementation](https://github.com/EVAST9919/osu-framework/compare/hacky-circle...EVAST9919:osu-framework:Circle-replace?expand=1#diff-432853374b221479db4515147fe4414d47b8b562813a52cb640f3cf533ae5256) which relies on unused `y` component of `BlendRange` being set to a unique value, however it's probably too ugly.

EVAST9919:
Hmm, found discrepancy in behaviour.
While `Circle` changes it's shape on `Scale`, `FastCircle` is always preserving the pill shape and I'm not sure which one is more expected to the end user.
Снимок экрана (528)

EVAST9919:
> Hmm, found discrepancy in behaviour

Okay, this one is easily fixable.
But also: while I can see how this component is too niche to include with the framework and potentially may cause performance issues upon misuse, I think it will be beneficial to at very least use it locally in osu! in the scenarios shown in OP, where we are sure that multiple circles will be drawn consecutively in a single draw call.
I have branches ready and can proceed further if necessary.

smoogipoo:
This PR's still in my emails fwiw, just very low priority while working on the many other tasks... I do eventually want to come back and re-read through the above comments + the capabilities offered by the existing `Circle`.

I may be okay blindly merging this in as `FastCircle` with documentation that it behaves slightly differently to `Circle` and is subject to change, for the specific scenarios that warrant it.

smoogipoo (approved):
I'm going to merge this one as it is for now to see where things go. Thanks!

peppy:
Hmm, was kinda hoping this would just replace `Circle`. I really don't want to have to go through code and decide when we want to use `FastCircle` vs `Circle`. Would prefer to just fix the few broken cases of `Circle` usage with the new implementation if that's possible.

smoogipoo:
My point is I don't know if this is the best that can be done. It still breaks the batch after all, and in a way that is much harder to prevent. 99% of cases would be fine with just using `Circle`, there are very specific use-cases that would _currently_ benefit majorly from this.

peppy:
Fair enough

PR #6298, opened 2024-05-22
Add back support for opening dropdown by setting internal states

smoogipoo:
See `TestSceneOsuDropdown` failures in https://github.com/ppy/osu/pull/28286


peppy (approved):
really don't like this, especially the schedule..

PR #6300, opened 2024-05-23
Fix IL3000 warning on Android

Susko3:
https://learn.microsoft.com/en-us/dotnet/core/deploying/single-file/warnings/il3000

> osu-framework\osu.Framework.Tests\IO\TestOnlineStore.cs(102,35): warning IL3000: 'System.Reflection.Assembly.Location' always returns an empty string for assemblies embedded in a single-file app. If the path to the app directory is needed, consider calling 'System.AppContext.BaseDirectory'.

Likely to show on embedded mobile platforms like iOS and Android.


peppy (approved):
Given this is only changed in test code I think it's fine without too much considerations.

Issue #5714, opened 2023-03-28
Implicit conversions in `SRGBColour` should not exist


frenzibyte:
As part of making sense of sRGB colours. It should be explicitly shown whether `SRGBColour` is going to give a linear or sRGB colour, by getting rid of the implicit conversions in `SRGBColour` and potentially replacing them in public API with just pure `Colour4`s.

This should make it clear in what space the colours passed to the shader are, and hopefully reduce confusion on sRGB colours in general.

PR #6301, opened 2024-05-25
Colour API proposal

FreezyLemon:
The changes in this PR are the additions needed to implement the proposal as a usable API. More things might need to be added later. Breaking changes are described below but not included in the PR so that reading the code is easier.

Keep in mind that this is a proposal and while the commits in this PR could be merged, they exist more to demonstrate the API as code. I don't mind changing stuff.

## API overview

- `ColourInfo`: Semantically unchanged. Describes the colours of the four vertices of a quad.
- `SRGBColour`: Main single-colour type that framework users interact with. Contains a gamma-corrected sRGB colour. Thin wrapper around `Colour4`. The type discourages invalid things like gamma-incorrect colour math[^1].
- `LinearColour` (new): Secondary colour type for framework users. Contains a linear-light sRGB colour[^2]. Thin wrapper around `Colour4`. Basically only exists to do colour math.
- `Colour4`: Should rarely be used directly. Low level, raw colour values, no inherent colour space. This is where the real math (and optimization) should happen so that all abstractions can benefit.
- `Color4`: Semantically the same as `Colour4`. `Colour4` (no `osuTK` dependency, uses `System.Numerics.Vector4` which should optimise better) should always be preferred over `Color4` unless interacting with an osuTK API that requires it.

## Implementation notes

- Turning `MultiplyAlpha` into a pure method is a hidden trap for existing usage (`[Pure]` should help with this, but not sure it's enough). Renaming it is an alternative, but making it pure is a requirement for a readonly `SRGBColour`.
- The colour constants `Color4.Red` etc. are copies of the CSS [`<named-color>`](https://developer.mozilla.org/en-US/docs/Web/CSS/named-color) values and are [defined](https://developer.mozilla.org/en-US/docs/Web/CSS/named-color#description) to be in the sRGB colour space, so they should live in `SRGBColour`. This is also necessary to remove the implicit conversions `Color4`/`Colour4` <-> `SRGBColour` later on.
- The change from `Color4` to `Colour4` inside `SRGBColour` is most likely going to have performance implications. I am not sure how to test this because colours are used everywhere, and I do not have enough insight into the renderer implementations to know how this impacts them.

## Next steps

This list is mostly unordered.

- [X] Update `Colour4` and `ColourInfo` docs
- [X] Make `SRGBColour` readonly and extend API to cover common usage
- [X] Implement LinearColour
- [ ] Remove `SRGBColour.Linear` (**breaking** with low impact)
  - Returns a `Colour4` which loses colour space information
  - `ToLinear()` should be used instead
- [ ] Remove `SRGBColour` operator overloads (`*`, `/`, `+`) (**breaking** with low impact)
  - Use `ToLinear()` and `LinearColour` instead
  - Make the expensive sRGB<->linear conversions explicit
  - Chaining the `SRGBColour` operators is slow and imprecise and turns the operators into a footgun[^3]
- [ ] Replace `Color4` and `Colour4` in public API with `SRGBColour` or maybe `Colour4` instead (**breaking** with very high impact)
  - can be done in steps
  - Notable: `OsuColour`, `Color4Extensions`
- [ ] Remove implicit conversions between `Color4`/`Colour4` and `SRGBColour` (see #5714) (**breaking** with high impact)
- [ ] Remove implicit conversions between `Color4`/`Colour4` and `ColourInfo` (**breaking** with very high impact)
  - This is going to be a huge change (`drawable.Colour = Color4.Red` must be replaced with `drawable.Colour = SRGBColour.Red`), but can often be done with search-and-replace.
  - Benefit: Removes `osuTK` from a lot of places.
- [ ] Figure out what to do with colour-related `Bindable` types
- [ ] Figure out what to do with colour pickers
- [ ] Maybe move hexcode parsing and related code to `SRGBColour` (these are also semantically sRGB)
- [ ] Maybe remove colour constants in `Colour4`

[^1]: Gamma-incorrect operations could be done with helper functions.
[^2]: [As described here](https://developer.mozilla.org/en-US/docs/Glossary/Color_space) (search for "srgb-linear"). [Can also be called](https://en.wikipedia.org/wiki/SRGB#From_sRGB_to_CIE_XYZ) "linear sRGB" or "gamma-expanded sRGB". This is usually called "linear colour"/"linear colour space" in and around o!f, which is why the struct is named `LinearColour`.
[^3]: For example, `a * b * c` will: Convert `a` and `b` to linear. Multiply those two. Convert the result back to sRGB. Convert the first result and `c` to linear. Multiply those two. Convert that result back to sRGB.

### DIFF
--- a/osu.Framework.Benchmarks/BenchmarkColourInfo.cs
+++ b/osu.Framework.Benchmarks/BenchmarkColourInfo.cs
@@ -3,8 +3,8 @@
 
 using System.Collections.Generic;
 using BenchmarkDotNet.Attributes;
+using osu.Framework.Graphics;
 using osu.Framework.Graphics.Colour;
-using osuTK.Graphics;
 
 namespace osu.Framework.Benchmarks
 {
@@ -18,9 +18,9 @@ namespace osu.Framework.Benchmarks
         {
             get
             {
-                yield return ColourInfo.SingleColour(Color4.Transparent);
-                yield return ColourInfo.SingleColour(Color4.Cyan);
-                yield return ColourInfo.SingleColour(Color4.DarkGray);
+                yield return ColourInfo.SingleColour(SRGBColour.Transparent);
+                yield return ColourInfo.SingleColour(SRGBColour.Cyan);
+                yield return ColourInfo.SingleColour(SRGBColour.DarkGray);
             }
         }
 
@@ -28,13 +28,13 @@ namespace osu.Framework.Benchmarks
         public SRGBColour ConvertToSRGBColour() => Colour;
 
         [Benchmark]
-        public Color4 ConvertToColor4() => ((SRGBColour)Colour).Linear;
+        public Colour4 ConvertToColour4() => ((SRGBColour)Colour).ToLinear().Linear;
 
         [Benchmark]
-        public Color4 ExtractAndConvertToColor4()
+        public Colour4 ExtractAndConvertToColour4()
         {
             Colour.TryExtractSingleColour(out SRGBColour colour);
-            return colour.Linear;
+            return colour.ToLinear().Linear;
         }
     }
 }
--- a/osu.Framework.Benchmarks/BenchmarkSRGBColourMultiplication.cs
+++ b/osu.Framework.Benchmarks/BenchmarkSRGBColourMultiplication.cs
@@ -3,31 +3,18 @@
 
 using BenchmarkDotNet.Attributes;
 using osu.Framework.Graphics.Colour;
-using osuTK.Graphics;
 
 namespace osu.Framework.Benchmarks
 {
     public class BenchmarkSRGBColourMultiplication : BenchmarkTest
     {
-        private static readonly SRGBColour white = new SRGBColour
-        {
-            SRGB = new Color4(1f, 1f, 1f, 1f)
-        };
+        private static readonly SRGBColour white = SRGBColour.White;
 
-        private static readonly SRGBColour white_with_opacity = new SRGBColour
-        {
-            SRGB = new Color4(1f, 1f, 1f, 0.5f)
-        };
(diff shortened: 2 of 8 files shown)


bdach:
> The type discourages invalid things like gamma-incorrect colour math
> (...)
> Gamma-incorrect operations could be done with helper functions.

I immediately disagree with these premises. They sound like ivory tower thinking to me.

As discussed multiple times to death in discord, like it or not, in 99% of use cases we *want* gamma "incorrect" to match every other graphics program that is gamma "incorrect". So we're going to have to sprinkle these "helpers" everywhere just to get what we want to match everything else, in 99% of use cases.

smoogipoo:
Without spending too much time on this for now, my preliminary thought is that SRGB doesn't make sense to think about and use directly. It's only really useful when interpolating/blending colours. For anything else, I'd prefer the user to not have to think about SRGB unless it needs to be resolved into a colour space for the purpose of doing some sort of maths.

It's unintuitive because we never think of colours in SRGB - when we translate an RGB value from figma or want "red", we're basically dealing with the "linear" colour space, though colour spaces are actually irrelevant at this instant in time.

Eventually, the question will arise (from a new o!f consumer): "which colour should I use, SRGBColour or LinearColour?" To, which the correct response will be "use either as they're identical, unless you're doing interpolation".  But really, the user shouldn't have to think about this or ask this at all.

So... I would actually be in favour of using `Color4` _more_, and `SRGBColour` less. The considerations, imo, only need to be around:
- `Interpolation.ValueAt()`: Perhaps this needs `Color4` and `SRGBColour` variants, with the user required to cast to perform gamma-correct interpolation.
- `Drawable.FadeColour()`: This hides the underlying interpolation and needs some sort of special support - perhaps a parameter `bool gammaCorrect` is fine but it will also need its own transform.

PR #6303, opened 2024-05-27
Mark `FindClosestParent<T>()` callers as possibly null-returning

bdach:
RFC.

In particular, the important callers in question here are `GetContainingInputManager()` and `GetContainingFocusManager()`. The hope is that with proper nullability annotations, https://github.com/ppy/osu-framework/pull/6302 would have been caught by static analysis.

When preparing this, I saw that about 20% of calls to `GetContainingFocusManager()` already handled null, and the rest did not. Make of that what you will.

Game-side branch for preview of impact: https://github.com/ppy/osu/compare/master...bdach:osu:find-closest-parent-null-adjustments?expand=1


osu.Framework/Graphics/UserInterface/TextBox.cs
```diff
@@ -1224,7 +1224,7 @@ protected override void OnDrag(DragEvent e)
 
                 selectionEnd = getCharacterClosestTo(e.MousePosition);
                 if (hasSelection)
-                    GetContainingFocusManager().ChangeFocus(this);
+                    GetContainingFocusManager()?.ChangeFocus(this);
```
smoogipoo, line 1227: Any reason to not use `AsNonNull()` here? Pretty much every other case makes the assumption that if you're getting into any of the input methods or `Update()` then you have at least a parenting input manager somewhere.

osu.Framework/Testing/TestBrowser.cs
```diff
@@ -324,7 +324,7 @@ public bool OnPressed(KeyBindingPressEvent<TestBrowserAction> e)
             {
                 case TestBrowserAction.Search:
                     if (leftContainer.Width == 0) toggleTestList();
-                    GetContainingFocusManager().ChangeFocus(searchTextBox);
+                    GetContainingFocusManager()?.ChangeFocus(searchTextBox);
```
smoogipoo, line 327: As above

smoogipoo (commented):
In principle I'm fine with this but I have to mention that the use of `?.` vs `.AsNonNull()` doesn't make much sense to me.

osu.Framework/Graphics/UserInterface/TextBox.cs
bdach: The thinking was minimal line of resistance. Wherever it was feasible to just null-propagate and ignore whether there is an input/focus manager, I did that. Wherever it generally isn't sane to assume the lack of one I did `.AsNonNull()` (e.g. update methods of popovers or menus that 200% rely on one existing, and if the input manager was to disappear from under it, there would be more breakage than a nullref).

osu.Framework/Graphics/Containers/TabbableContainer.cs
```diff
@@ -44,7 +44,7 @@ protected override bool OnKeyDown(KeyDownEvent e)
                 return false;
 
             var nextTab = nextTabStop(TabbableContentContainer, e.ShiftPressed);
-            if (nextTab != null) GetContainingFocusManager().ChangeFocus(nextTab);
+            if (nextTab != null) GetContainingFocusManager()?.ChangeFocus(nextTab);
```
smoogipoo, line 47:
```suggestion
            if (nextTab != null) GetContainingFocusManager().AsNonNull().ChangeFocus(nextTab);
```

osu.Framework/Graphics/UserInterface/Menu.cs
```diff
@@ -262,7 +263,7 @@ private void updateState()
                     {
                         Schedule(delegate
                         {
-                            if (State == MenuState.Open) GetContainingFocusManager().ChangeFocus(this);
+                            if (State == MenuState.Open) GetContainingFocusManager()?.ChangeFocus(this);
```
smoogipoo, line 266:
```suggestion
                            if (State == MenuState.Open) GetContainingFocusManager().AsNonNull().ChangeFocus(this);
```

osu.Framework/Graphics/UserInterface/Menu.cs
```diff
@@ -570,7 +571,7 @@ private void openSubmenuFor(DrawableMenuItem item)
             if (item.Item.Items.Count > 0)
             {
                 if (submenu.State == MenuState.Open)
-                    Schedule(delegate { GetContainingFocusManager().ChangeFocus(submenu); });
+                    Schedule(delegate { GetContainingFocusManager()?.ChangeFocus(submenu); });
```
smoogipoo, line 574:
```suggestion
                    Schedule(delegate { GetContainingFocusManager().AsNonNull().ChangeFocus(submenu); });
```

osu.Framework/Graphics/UserInterface/TextBox.cs
smoogipoo, line 1227:
```suggestion
                    GetContainingFocusManager().AsNonNull().ChangeFocus(this);
```

osu.Framework/Testing/TestBrowser.cs
smoogipoo, line 327:
```suggestion
                    GetContainingFocusManager().AsNonNull().ChangeFocus(searchTextBox);
```

osu.Framework/Graphics/UserInterface/TextBox.cs
smoogipoo: I marked the ones I'm expecting to be `.AsNonNull()` in the review below. Curious whether you agree?

It's really one of those scenarios where I would rather use `!` but, well that ship's sailed by this point...

bdach: Yeah okay, I see, you meant the call sites where it is *clearly and directly* visible that there *must* be an input manager present. Sure, I'll apply those :+1:

smoogipoo (dismissed):

smoogipoo (approved):

PR #6306, opened 2024-05-29
Remove OS username from log headers

sineplusx:
Discussed in [ppy/osu#28243](https://github.com/ppy/osu/discussions/28243#discussioncomment-9548742)

This PR removes the `UserIdentifier` from log headers, along with the field itself.

I also looked into the possibility of setting the `UserIdentifier` from game code. For osu!, this would be using the account username when logged in. I found this unpractical as it depends on the user being authenticated before the header (and subsequently, the first log line) is written.

To my knowledge, no consuming projects use this field to set a custom username.


smoogipoo (approved):
Requesting one more team approval.

peppy (approved):

PR #6307, opened 2024-06-02
Make shader setter in `SpriteText` protected

EVAST9919:
Needed for my personal project.

### DIFF
--- a/osu.Framework/Graphics/Sprites/Sprite.cs
+++ b/osu.Framework/Graphics/Sprites/Sprite.cs
@@ -30,6 +30,12 @@ namespace osu.Framework.Graphics.Sprites
             TextureShader = shaders.Load(VertexShaderDescriptor.TEXTURE_2, FragmentShaderDescriptor.TEXTURE);
         }
 
+        /// <summary>
+        /// The shader which should be used for rendering this sprite.
+        /// </summary>
+        /// <remarks>
+        /// This is automatically populated, but may be overridden if required for special cases.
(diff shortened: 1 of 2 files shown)


peppy:
Hmm, usually we'd made a `CreateTextureShader` method and expose that instead. I'm not against just exposing as `protected` because it's very hard to misuse (you'd have to set it in the `ctor` to have it overwritten by framework).

At very least, xmldoc mentioning this is automatically set but can be overwritten post-load?

Will wait for another opinion though.

peppy (commented):
as commented

bdach:
> Hmm, usually we'd made a `CreateTextureShader` method and expose that instead.

`Sprite` doesn't do this and it already has consumers of its protected setter:

https://github.com/ppy/osu-framework/blob/980b4cd0db2dd692a354bda26e3a8151ce5bd0a8/osu.Framework/Graphics/Sprites/Sprite.cs#L33

As to whether it should, I guess I have no big feelings about that, but arbitrary setter allows for doing more (aside from the pitfall of setting it in ctor).

> you'd have to set it in the `ctor` to have it overwritten by framework
> 
> At very least, xmldoc mentioning this is automatically set but can be overwritten post-load?

xmldoc can be added for sure. Though attempting to procure an `IShader` instance in a ctor would be *mighty* annoying I'd imagine so you could argue the API inertia here should already prevent that pitfall. But I agree that mentioning it (in both places maybe even) would not hurt.

peppy (approved):

bdach (approved):

PR #6312, opened 2024-06-24
Add virtualised list container

bdach:
RFC.

Game-side, we have a few lists/tables which perform dreadfully when stressed by a large number of rows/columns. Two of them reside in the editor, namely: the timing table and the verify issue list table.

The primary overheads in both are:
- All of the rows' content is constructed, loaded, and placed in the hierarchy eagerly, which causes a long freeze on first show.
- Even though 99.999999% of the content is not visible, it is still present in various places where it is causing trouble. While masking does cull *some* of it, the big killer is input handling - if any ancestor of a table row _might_ handle e.g. mouse input because they are all alive, all of them get shoved into the input queue and then all need to be checked on every mouse movement. (See https://github.com/ppy/osu-framework/pull/6282.) Mitigation of this has been applied via ~~dumb~~ hacks like having the _background_ of the table be a separate drawable under it and have it handle input. I was sorta hoping `ShouldBeConsideredForInput` could maybe fix this but it regretfully does not (it reduces *some* overhead but not enough to be acceptable yet).
- For mutable lists, 99% of usage sites will not handle the args carefully, but rather just clear everything and reconstruct from scratch, which is obviously rather expensive for long lists.

What *this* pull is intended to introduce is a sort of [`UITableView`](https://developer.apple.com/documentation/uikit/uitableview) but for osu!framework usage out-of-the-box.

- Exposes a "virtualised list" data-oriented interface. You derive from `VirtualisedListContainer`, and give a model and a drawable for it via the generic specs.
- The list item must be poolable and must have a bindable model on it (can't use `ModelBackedDrawable` because I can't inherit from both `PoolableDrawable` and that, so `IHasCurrentValue` it is)
- List items are assumed to have fixed heights so that layout can be ballparked in advance without items being physically present in the hierarchy.

For testing there is the test scene added here, but there's also [this branch](https://github.com/bdach/osu/compare/control-point-table-is-bad?expand=1) (needs to be cleaned up yet, very PoC). In my vague "performance testing" (read: looking at frame graph times), on a stress test like https://osu.ppy.sh/beatmapsets/1238185#mania/2574372 this can improve FPS by factors of 5 to even an order of magnitude:

- timing - 170fps -> 1000fps
- verify - 13fps -> 400fps

(both tested on release configuration, with static mouse).

The pooling takes care of the drawable construction overheads, and the lifetime games this container plays help with input handling (if a child isn't alive, [it doesn't get looked at with respect to input at all](https://github.com/ppy/osu-framework/blob/98617abd34280121f64967173cc42b915216dab7/osu.Framework/Graphics/Containers/CompositeDrawable.cs#L1428-L1432)). Collection changes are handled _somewhat_ efficiently (better than reconstructing everything from scratch, at least).

The one case that is still not as fast as I perhaps would have hoped for is active scrolling, but it's still better than it was.

### DIFF
--- a/osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
+++ b/osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
@@ -12,7 +12,6 @@ using osu.Framework.Bindables;
 using osu.Framework.Caching;
 using osu.Framework.Graphics.Pooling;
 using osu.Framework.Graphics.UserInterface;
-using osu.Framework.Logging;
 
 namespace osu.Framework.Graphics.Containers
 {
@@ -75,8 +74,6 @@ namespace osu.Framework.Graphics.Containers
         {
             visibilityCache.Invalidate();
             Items.Height = rowHeight * RowData.Count;
-
-            Logger.Log(string.Join(Environment.NewLine, RowData.Select(d => d.ToString())));
 
             var items = Items.FlowingChildren.ToArray();
 
--- a/osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
+++ b/osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
@@ -220,7 +220,7 @@ namespace osu.Framework.Graphics.Containers
                 }
             }
 
-            public bool Visible { get; set; }
+            public bool Visible { get; private set; }
 
             private readonly DrawablePool<TDrawable> pool;
 
--- a/osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
+++ b/osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
@@ -229,6 +229,12 @@ namespace osu.Framework.Graphics.Containers
                 Row = row;
                 this.pool = pool;
                 RelativeSizeAxes = Axes.X;
+
+                // to avoid overheads from input handling (or, to be more specific, from constructing input queues),
+                // the row manages its own lifetime.
+                // if the row is not alive, it is not in `AliveInternalChildren` of its parent,
+                // which means that it is omitted from consideration in `Build(Non)PositionalInputQueue()` et al.
+                LifetimeEnd = double.NegativeInfinity;
             }
 
             public void Load()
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
```diff
@@ -0,0 +1,250 @@
...
+                [MemberNotNull(nameof(row))]
+                set
+                {
+                    row = value;
+                    if (InternalChildren.SingleOrDefault() is TDrawable rowDrawable)
+                        rowDrawable.Current.Value = value;
+                }
+            }
+
+            public bool Visible { get; set; }
```
peppy, line 222:
```suggestion
            public bool Visible { get; private set; }
```

i'd hope

osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
```diff
@@ -0,0 +1,250 @@
...
+                Row = row;
+                this.pool = pool;
+                RelativeSizeAxes = Axes.X;
+            }
+
+            public void Load()
+            {
+                InternalChild = pool.Get(d => d.Current.Value = Row);
+                Visible = true;
+                LifetimeEnd = double.PositiveInfinity;
```
peppy, line 237: I'm guessing this is doing some wizardry to avoid processing of drawables when not loaded? It's not 100% obvious to me what this entails, so I'd hope for an inline comment explaining the reasoning.

Also a bit weird that at construction of an `ItemRow`, the default value is `double.PositiveInfinity` even though it's not yet loaded.

osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
```diff
@@ -0,0 +1,250 @@
...
+                {
+                    if (e.NewStartingIndex >= 0)
+                    {
+                        for (int i = e.NewStartingIndex; i < items.Length; ++i)
+                            Items.SetLayoutPosition(items[i], i + e.NewItems!.Count);
+                    }
+
+                    for (int i = 0; i < e.NewItems!.Count; ++i)
+                    {
+                        Items.Insert(Math.Max(e.NewStartingIndex, 0) + i, new ItemRow((TData)e.NewItems![i]!, pool)
```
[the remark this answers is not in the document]
bdach: Yes, I was aware that there was room for improvement still technically if we wanted to, but decided in the end that the tradeoff in implementation complexity did not seem worth it at this time.

peppy (approved):

PR #6314, opened 2024-06-25
Update SDL3-CS

hwsmm:
- Depends on https://github.com/ppy/SDL3-CS/pull/91

This only contains code changes for now as new SDL3-CS is not available yet.


peppy (approved):
Looks isolated enough to be fine.

PR #6315, opened 2024-06-26
Fix clicking on padded dropdown search bars not opening the menu

smoogipoo:
Fixes https://github.com/ppy/osu/issues/28594

In hindsight, it's probably not a good idea to handle click-to-open input in something like a "search bar".

In osu!, the search bar is padded to add space for the down-arrow icon: https://github.com/ppy/osu/blob/9039ec34bae03a9b20e43b2161cc5c9e767c3889/osu.Game/Graphics/UserInterface/OsuDropdown.cs#L405-L408

I also explored the option of moving the padding to the textbox proper, but textboxes don't seem to properly handle padding at all.

### DIFF
--- a/osu.Framework/Graphics/UserInterface/DropdownHeader.cs
+++ b/osu.Framework/Graphics/UserInterface/DropdownHeader.cs
@@ -98,6 +98,11 @@ namespace osu.Framework.Graphics.UserInterface
                     AutoSizeAxes = Axes.Y
                 },
                 SearchBar = CreateSearchBar(),
+                new ClickHandler
+                {
+                    RelativeSizeAxes = Axes.Both,
+                    Click = onClick
+                }
             };
         }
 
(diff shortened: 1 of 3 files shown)


osu.Framework/Graphics/UserInterface/DropdownHeader.cs
```diff
@@ -129,6 +134,22 @@ private void updateState()
...
+        {
+            // Allow input to fall through to the search bar (and its contained textbox) if it's visible.
+            if (SearchBar.State.Value == Visibility.Visible)
+                return false;
+
+            // Otherwise, the header acts as a button to show/hide the menu.
+            dropdown.ToggleMenu();
+
+            // And importantly, when the menu is closed as a result of the above toggle, block the search bar from receiving input.
+            return dropdown.MenuState == MenuState.Closed;
```
smoogipoo, line 150: I was _just about_ to remove this, and ended up adding it back. I want to explain this a little bit for future reference.

> "Why isn't this just `return true;`?"

Originally, this condition was required because of an interaction with the code changing the menu state when the textbox is focused/unfocused:
https://github.com/ppy/osu-framework/blob/94cd76a25f37896ab348569d0a619ffab2edf86b/osu.Framework/Graphics/UserInterface/DropdownSearchBar.cs#L114-L123

Basically, we need to make sure the textbox receives focus at some point, hence input needs to pass through when the dropdown is supposed to be visible, and _not_ pass through when it's supposed to be invisible.

By sheer luck, and as a result of a separate bugfix, this is _also_ handled by:

https://github.com/ppy/osu-framework/blob/94cd76a25f37896ab348569d0a619ffab2edf86b/osu.Framework/Graphics/UserInterface/DropdownSearchBar.cs#L158-L163

That specifically handles the case where the dropdown is opened externally, i.e. by calling `Dropdown.Menu.Open()`.

Given that the above is a very roundabout way of handling this, I prefer to be explicit about intentions here.

peppy (approved):

PR #6317, opened 2024-06-27
Remove faulty assertion in scroll container

frenzibyte:
- Closes https://github.com/ppy/osu-framework/issues/2697

Hit this multiple times while just wanting to scroll through a zoomed-in editor timeline in osu!. Discussed removing assertion in [discord](https://discord.com/channels/188630481301012481/589331078574112768/1255759444739096606).


bdach:
I'm not sure I see the point in commenting the assertion? Either remove it or fix it (presumably by having the check use relative tolerance rather than absolute).

bdach (approved):

PR #6321, opened 2024-06-29
Don't use `FillFlowContainer` in `VirtualisedListContainer`

EVAST9919:
[Relevant discussion](https://github.com/ppy/osu-framework/pull/6312#discussion_r1654031731)

This pr greatly improves scrolling performance (and allocations) of the `VirtualisedListContainer` by removing overhead from `FillFlowContainer` and avoiding drawables copying when unnecessary.
[osu-side change required](https://github.com/ppy/osu/compare/master...EVAST9919:virtual-table-update?expand=1)

master: https://streamable.com/fvwg2o
pr: https://streamable.com/boqmrs

osu-master: https://streamable.com/pub39y
osu-pr: https://streamable.com/sb4zbg

map used: https://osu.ppy.sh/beatmapsets/1238185#mania/2574372


osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
```diff
@@ -86,15 +85,21 @@ private void itemsChanged(object? sender, NotifyCollectionChangedEventArgs e)
                     if (e.NewStartingIndex >= 0)
                     {
                         for (int i = e.NewStartingIndex; i < items.Length; ++i)
-                            Items.SetLayoutPosition(items[i], i + e.NewItems.Count);
+                        {
+                            var itemToMove = Items[Items.IndexOf(items[i])];
+                            itemToMove.Y = (i + e.NewItems.Count) * rowHeight;
+                            Items.ChangeChildDepth(itemToMove, -(i + e.NewItems.Count));
```
EVAST9919, line 91: Hmm, may be it would be better to move this logic into `ItemFlow` and a method with a same `SetLayoutPosition`name?

smoogipoo:
I will once again say that I feel uncomfortable with changes like this where I would hope the overhead of FFC isn't this large.

Where is the overhead coming from? Have you profiled it?

EVAST9919:
> I will once again say that I feel uncomfortable with changes like this where I would hope the overhead of FFC isn't this large.
> 
> Where is the overhead coming from? Have you profiled it?

`row.Load()/Unload()` is triggering FFC's full layout recomputation. Probably could be avoided with some invalidation tricks, but approach in this pr looked simpler for me.

osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
bdach: Realistically I don't think it matters one way or another.

osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
```diff
@@ -86,15 +85,21 @@ private void itemsChanged(object? sender, NotifyCollectionChangedEventArgs e)
...
 
                     for (int i = 0; i < e.NewItems.Count; ++i)
                     {
-                        Items.Insert(Math.Max(e.NewStartingIndex, 0) + i, new ItemRow((TData)e.NewItems[i]!, pool)
+                        Items.Add(new ItemRow((TData)e.NewItems[i]!, pool)
                         {
                             Height = rowHeight,
                             LifetimeEnd = double.NegativeInfinity,
+                            Depth = -(Math.Max(e.NewStartingIndex, 0) + i),
+                            Y = (Math.Max(e.NewStartingIndex, 0) + i) * rowHeight
```
bdach, line 102: If `Y` is being explicitly managed here, what are the `Depth` hijinks for? Is this a safety against rows potentially overlapping and displaying in incorrect z-order or something?

Looking at the game-side change, I'm *hoping* it's not being set just in order for the index of the item to be accessible by consumers *via `Depth` of all things*?

osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
```diff
@@ -161,7 +174,7 @@ private void assertCorrectOrder()
             for (int i = 0; i < children.Length; ++i)
             {
                 float expected = i;
-                float actual = Items.GetLayoutPosition(children[i]);
+                float? actual = -children[i].Depth;
```
bdach, line 177: Dunno why this is checking the depth of all things, Y position is the thing that actually matters in this assertion.

osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
```diff
@@ -170,7 +183,7 @@ protected override void Update()
         {
             base.Update();
 
-            var currentVisibleRange = (Scroll.Current, Scroll.Current + Scroll.DrawHeight);
+            var currentVisibleRange = ((int)(Scroll.Current / rowHeight) + 1, (int)((Scroll.Current + Scroll.DrawHeight) / rowHeight));
```
bdach, line 186: Why is the lower bound plus one? I would expect that to cause items to disappear when partially scrolled out (except they actually don't appear to)?

bdach (commented):
The performance improvement is there and the code is not as offensive as I figured it would look when trying to get rid of fill flow. There are still some WTFs here (see review comments below) but if those can be buffed out I can see myself approving this.

osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
EVAST9919: Fixed in 7ca13ea126eec2491d1d604c52fb4bc47979b9bf

osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
EVAST9919: Math in the loop was a bit incorrect, which became apparent after changing var names, now fixed in d17082aa26e56480739fc7bf914863f36c155681

peppy (dismissed):
Updated code reads well to me 👍

osu.Framework/Graphics/Containers/VirtualisedListContainer.cs
smoogipoo: I'm a bit uncomfortable with the maths done here being duplicated in multiple places. I'd like this to be split out into a method only for the purpose of directing the programmer maintaining this code to using the correct method if they need to move items around.

It sort of feeds into why this was asked: https://github.com/ppy/osu-framework/pull/6321#discussion_r1661901919, and I separately had the same query.

bdach: I've split methods for this to hopefully make stuff nicer / more apparent: https://github.com/ppy/osu-framework/pull/6321/commits/6ad3c18fc75a8dc958a0b27098214b16fcfee306

bdach (approved):

